PR #22397, opened 2022-05-05
msw SpinCtrl: handle special keys (e.g. Ctrl+C)

DietmarSchwertberger:
There is a wxPython bug report about menu entries like "Copy\tCtrl+C" having precedence over special character handling in the text control of SpinCtrl:
https://github.com/wxWidgets/Phoenix/issues/2148 

This can be fixed by adding `MSWShouldPreProcessMessage(WXMSG* msg);`.
I have just copied the implementation from `wxTextEntry::MSWShouldPreProcessMessage`, except that `const` did not work for me.


vadz:
Thanks, this looks like the right thing to do, but could you please extract the existing function in some private header and reuse it from `wxSpinCtrl` rather than mostly duplicating the existing code?

DietmarSchwertberger:
Did not know that "Draft" can be seen already...
I was about to add the question "Is there a more generic way of handling this? Duplicating code is probably not the best way."

Is it OK to add to "wx/msw/private.h"?

vadz:
If this function is 100% identical to the existing one in `wxTextEntry`, I think it would be enough to just extract that function to some `wxMSWShouldPreProcessEditMessage()` in the same `src/msw/textentry.cpp` and declare it as `extern` in `src/msw/spinctrl.cpp`. If anything else is needed, I'd rather create a new `include/wx/msw/private/edit.h` as `wx/msw/private.h` itself already has too much stuff.

TIA!

DietmarSchwertberger:
Thanks. This seems to work and looks much nicer.

Issue #22349, opened 2022-04-23
3.1.6 png transparency shows black in wxTreeCtrl


msbroadf:
My app contains a wxTreeCtrl with png images used and these have transparency. 

In 3.1.4 this displayed correctly. Recompiling with 3.1.6 with no other changes to my code shows the same icons in the wxTreeCtrl but with black instead of the transparent area. 

Is this something related to the wxBitmapBundle not correctly supporting transparency at the moment?

vadz:
This is not supposed to happen, so it looks like a bug but, as always, it would be great to have a simple way of reproducing this, ideally in the treectrl sample. Could you please make a minimal patch to it showing the problem?

Also, which platform does this happen under?

vadz:
It would be great to at least have the PNG files showing the problem.

msbroadf:
Sorry i have been trying to compile the treectl example but havent yet succeeded (my build process is very complicated) but i have attached the png with the issue for the time being

![audio](https://user-images.githubusercontent.com/13748102/165187153-cdb80e36-347a-49eb-835b-42d4242a0504.png)

vadz:
Thanks, I'll test with this one, but normally you should be able to just use the provided make- or project file to build the sample without any changes.

PBfordev:
I cannot confirm this on MSW. I used the current master but I do not see any relevant changes since 3.1.6.

I downloaded the image provided by msbroadf, used bundled bin2c to embed it in the tree sample and use it instead of "file" icon there. I also set the background color to yellow, to make the transparency stand out.
![treectrl-sample-png](https://user-images.githubusercontent.com/12495521/165263276-73c0da37-ecf2-4986-ac6f-416e7d998f2c.png)

The patch is here: https://github.com/PBfordev/wxWidgets/commit/7b30db394238dfb2eff75ec945a6d8534284de7c

msbroadf:
I managed to compile with no changes to treectrl example and it looks like this:

![Untitled](https://user-images.githubusercontent.com/13748102/165270923-51b7deac-e218-4c18-9b82-ff054f7c98de.png)

My setup is as follows:

wxWidgets-3.1.6.zip downloaded from github
mingw64 cross compiler building exe on linux

PBfordev:
[...]

PR #22400, opened 2022-05-06
Make handling of masks wxImageList much simpler and more useful

vadz:
After looking at this code in my attempts to solve #22349 I came to the conclusion that we actually don't need to do anything half as complicated as what we're doing here and can drastically simplify the code while making it more useful.

Useful for me means "do the expected thing" and the expected thing here is clearly use whatever transparency is there in the images, be it alpha or mask. Using `mask == false` shouldn't prevent us from doing it but should only disable creating the default mask based on grey pixels, as this is the only thing which can be possibly unexpected (and hence harmful) here.

@a-wi I'm sorry to undo parts of your changes, but I really don't think that we should stick to the formal interpretation of the `mask` parameter here as, I believe, you tried to do. As I've tried to explain in the commit message, it was never supposed to mean "don't use alpha", for the simple reason that it predates alpha support in wxWidgets.

### DIFF
--- a/include/wx/generic/imaglist.h
+++ b/include/wx/generic/imaglist.h
@@ -58,6 +58,8 @@ public:
 private:
     const wxBitmap *DoGetPtr(int index) const;
 
+    wxBitmap GetImageListBitmap(const wxBitmap& bitmap) const;
+
     wxVector<wxBitmap> m_images;
     bool m_useMask;
 
--- a/include/wx/msw/bitmap.h
+++ b/include/wx/msw/bitmap.h
@@ -276,6 +276,10 @@ public:
     WXHBITMAP GetMaskBitmap() const { return m_maskBitmap; }
     void SetMaskBitmap(WXHBITMAP bmp) { m_maskBitmap = bmp; }
 
+#if wxUSE_IMAGE
+    bool MSWCreateFromImageMask(const wxImage& image);
+#endif // wxUSE_IMAGE
+
 protected:
     WXHBITMAP m_maskBitmap;
 
--- a/include/wx/msw/imaglist.h
+++ b/include/wx/msw/imaglist.h
@@ -200,6 +200,14 @@ protected:
   wxSize m_size;
 
 private:
+  // Private helper used by GetImageListBitmaps().
+  class wxMSWBitmaps;
+
+  // Fills the provided output "bitmaps" object with the actual bitmaps we need
+  // to use with the native control.
+  void GetImageListBitmaps(wxMSWBitmaps& bitmaps,
+                           const wxBitmap& bitmap, const wxBitmap& mask);
+
   bool m_useMask;
 
   void Init()
--- a/interface/wx/imaglist.h
+++ b/interface/wx/imaglist.h
@@ -28,8 +28,8 @@ enum
     @class wxImageList
 
     A wxImageList contains a list of images, which are stored in an unspecified
-    form. Images can have masks for transparent drawing, and can be made from a
-    variety of sources including bitmaps and icons.
+    form. Images can use alpha channel or masks for transparent drawing, and
+    can be made from a variety of sources including bitmaps and icons.
 
     wxImageList is used principally in conjunction with wxTreeCtrl and
     wxListCtrl classes.
@@ -62,7 +62,12 @@ public:
         @param height
             Height of the images in the list.
         @param mask
-            @true if masks should be created for all images.
+            If @true, all images will have masks, with the mask being created
+            from the light grey pixels if not specified otherwise, i.e. if the
+            image doesn't have neither alpha channel nor mask and no mask is
+            explicitly specified when adding it. Note that if an image does
+            have alpha channel or mask, it will always be used, whether this
+            parameter is @true or @false.
         @param initialCount
             The initial size of the list.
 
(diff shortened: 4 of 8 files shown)


vadz:
If there are no comments/objections, I'm going to merge this soon.

MaartenBent:
I found a regression after this change. When I draw a white bitmap and set a white mask after, it doesn't become transparent anymore. When I set the mask immediately after creating the bitmap, before drawing, it is fine. Was my code just wrong or is this a bug?

before / after:
<a href="https://user-images.githubusercontent.com/8088070/168796129-5b5812d1-5a0f-43bf-a837-f75e438cd4ef.png">before</a>  <a href="https://user-images.githubusercontent.com/8088070/168796158-361d567c-9635-41bf-b390-accd57b671a7.png">after</a>

patch:
```diff
diff --git "a/samples/treectrl/treetest.cpp" "b/samples/treectrl/treetest.cpp"
index 283909a898..d246981478 100644
--- "a/samples/treectrl/treetest.cpp"
+++ "b/samples/treectrl/treetest.cpp"
@@ -989,8 +989,19 @@ void MyTreeCtrl::CreateImages(int size)
     }
     else
     {
+        wxBitmap bitmap(iconSize);
+        {
+            wxMemoryDC dc;
+            dc.SelectObject(bitmap);
+            dc.SetBackground(*wxWHITE);
+            dc.Clear();
+        }
+        bitmap.SetMask(new wxMask(bitmap, *wxWHITE));
+        wxIcon icon;
+        icon.CopyFromBitmap(bitmap);
+
         icons[TreeCtrlIcon_File] =
-        icons[TreeCtrlIcon_FileSelected] = wxArtProvider::GetIcon(wxART_NORMAL_FILE, wxART_LIST, iconSize);
+        icons[TreeCtrlIcon_FileSelected] = icon;
         icons[TreeCtrlIcon_Folder] =
         icons[TreeCtrlIcon_FolderSelected] =
         icons[TreeCtrlIcon_FolderOpened] = wxArtProvider::GetIcon(wxART_FOLDER, wxART_LIST, iconSize);
```

MaartenBent:
It seems to be caused by this part, reverting it fixes the transparency:
```diff
@@ -81,12 +79,16 @@ bool wxImageList::Create(int width, int height, bool mask, int initial)
     // (e.g. ILC_COLOR16) shows completely broken bitmaps
     flags |= ILC_COLOR32;
 
+    m_useMask = mask;
+
     // For comctl32.dll < 6 always use masks as it doesn't support alpha.
-    if ( mask || wxApp::GetComCtl32Version() < 600 )
-    {
-        m_useMask = true;
+    //
+    // We also have to use masks when we don't have wxImage and wxDIB that are
+    // needed to handle alpha.
+#if wxUSE_WXDIB && wxUSE_IMAGE
+    if ( wxApp::GetComCtl32Version() < 600 )
+#endif
         flags |= ILC_MASK;
-    }
 
     // Grow by 1, I guess this is reasonable behaviour most of the time
     m_hImageList = (WXHIMAGELIST) ImageList_Create(width, height, flags,`
```

MaartenBent:
Is there a reason why `ILC_MASK` is not set anymore for recent comctl32? To me it seems this should always be set.

vadz:
Thanks for noticing this!

> Is there a reason why `ILC_MASK` is not set anymore for recent comctl32? To me it seems this should always be set.

The idea was that we'd just always use alpha if it's supported to make things simpler. But there is a snag: apparently fully transparent images are considered to be black by comctl32.dll :-(

I.e. just adding `dc.DrawPoint(0, 0);` to your patch above makes the problem go away, but as long as the DIB data is only zeroes, it appears completely black (BTW, I can't reproduce the "image becomes red" part: if I do `s/wxWHITE/wxRED/g` the image is still completely black for me, which is not surprising as the DIB still has exactly the same data consisting of only 0s).

I don't know if we can somehow convince Windows to use a fully transparent bitmap as an actually transparent one here, rather than as a solid black one. Using alpha of 1, rather than 0, for a single pixel is sufficient and practically invisible, but certainly feels like a horrible hack. So I guess we're going to switch back to using masks -- but then we should probably update the code to avoid converting the mask to alpha, as this seems useless if we rely on MSW doing it anyhow...

dconnet:
FYI, based on @MaartenBent's comment above, I did:
```
diff --git a/src/msw/imaglist.cpp b/src/msw/imaglist.cpp
index f2631a3acb..cecb053d6c 100644
--- a/src/msw/imaglist.cpp
+++ b/src/msw/imaglist.cpp
@@ -86,7 +86,7 @@ bool wxImageList::Create(int width, int height, bool mask, int initial)
     // We also have to use masks when we don't have wxImage and wxDIB that are
     // needed to handle alpha.
 #if wxUSE_WXDIB && wxUSE_IMAGE
-    if ( wxApp::GetComCtl32Version() < 600 )
+    if ( mask || wxApp::GetComCtl32Version() < 600 )
 #endif
         flags |= ILC_MASK;
```
and my black images went away. I'm generating a transparent image for use in an image list by doing:
```
		wxImage img(imageSize, imageSize, true);
		img.SetMaskColour(wxBLACK->Red(), wxBLACK->Green(), wxBLACK->Blue());
		bmp = wxBitmap(img);
```

PR #22403, opened 2022-05-06
grid: fix row/col drag markers

DietmarSchwertberger:
This PR fixes DC preparation and marker size when grid row/col drag markers are drawn.

From unifying row/col mouse handling, `m_rowLabelWidth` was left over.
`wxGridOperations::GetLabelSize` is now implemented/used instead.

When dragging a col/row, a marker is drawn/deleted.
Until now the grid's `DoPrepareDC` was used to scroll the `wxClientDC`. This does inialize both scroll directions, which is not correct. When e.g. dragging a row in a horizontally scrolled grid, there was an offset.
`wxGridOperations::PrepareDCForLabels` is now implemented instead, which resets one direction to the unscrolled value.

The DC handling could probably be unified with `wxGridRowLabelWindow::OnPaint`, but I'm not into this deep enough.

I'm not 100% happy with the names. Maybe, names `GetLineLabelSize` and `PrepareDCForLineLabels` would be better, but I don't like these either as no line numbers arguments are involved here.

### DIFF
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -4052,27 +4052,25 @@ void wxGrid::ProcessRowColLabelMouseEvent( const wxGridOperations &oper, wxMouse
             else if ( m_cursorMode == oper.GetCursorModeMove() )
             {
                 // determine the y or x position of the drop marker
-                int marker;
+                int markerPos;
                 if ( coord >= oper.GetLineStartPos(this, lineAt) + (oper.GetLineSize(this, lineAt) / 2) )
-                    marker = oper.GetLineEndPos(this, lineAt);
+                    markerPos = oper.GetLineEndPos(this, lineAt);
                 else
-                    marker = oper.GetLineStartPos(this, lineAt);
+                    markerPos = oper.GetLineStartPos(this, lineAt);
 
-                if ( marker != m_dragLastPos )
+                if ( markerPos != m_dragLastPos )
                 {
                     wxClientDC dc( headerWin );
                     oper.PrepareDCForLabels(this, dc);
 
-                    wxSize clientSize = headerWin->GetClientSize();
-
-                    marker++;
+                    int markerLength = oper.Select(headerWin->GetClientSize());
 
                     // Clean up the last indicator
                     if ( m_dragLastPos >= 0 )
                     {
                         wxPen pen(headerWin->GetBackgroundColour(), 2);
                         dc.SetPen(pen);
-                        oper.DrawParallelLine(dc, 0, oper.Select(clientSize), m_dragLastPos + 1);
+                        oper.DrawParallelLine(dc, 0, markerLength, m_dragLastPos + 1);
                         dc.SetPen(wxNullPen);
 
                         int lastLine = oper.PosToLine(this, m_dragLastPos, NULL, false);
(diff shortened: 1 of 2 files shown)


DietmarSchwertberger:
OK, I just found that `oper.Select` and `GetClientSize` are being used in another section.
I will re-work accordingly and drop `GetLabelSize` again.

DietmarSchwertberger:
I found the issues while experimenting with an improved visualisation of pressed row/col labels when starting a drag move operation.

This is the visualisation as of now:
grafik grafik


This is better, but would leave the right border:
grafik  grafik


This would also overdraw the right border:
grafik grafik
grafik


What do you think? The last version is my prefered one.
The ideal solution would probably be a renderer Method for a pressed label, but I doubt that this would be used a lot in custom implementations. Anyway, I would probably factor out the drawing in a separate method, as currently the "pressed" display is destroyed when the user moves over the row/col again while the drag is ongoing.


(The darker lines at bottom and right actually belong to the header cell. For visualisation, I used coloured pens in `wxGridRowHeaderRendererDefault::DrawBorder`:
grafik
The blue line is only drawn if `grid.GetBorder() == wxBORDER_NONE`.)

src/generic/grid.cpp
```diff
@@ -249,6 +249,20 @@ int wxGridColumnOperations::GetFirstLine(const wxGrid *grid, wxGridWindow *gridW
     return grid->GetNumberFrozenCols();
 }
 
+void wxGridOperations::PrepareDCForLabels(wxGrid *grid, wxDC &dc) const
+{
```
vadz, line 253: I think there should be a comment in this function explaining what it does and why, as it's not really obvious.

I'd also like to find some less ugly way of doing it, but I don't really see any right now... The only nice solution would be to factor out some `PrepareDCForScrollPos(wxDC& dc, int x, int y)` from `wxScrollHelperBase::DoPrepareDC()` and then call it with one of `x` or `y` set to the actual scroll position and the other one set to 0, but I'm not 100% sure it's worth it. OTOH it would definitely make this function short and simple enough to not need any comment!

vadz (commented):
Thanks, I agree that the last picture seems to be the best one, although, to be honest, I think it still doesn't look that great (personally I still think we should change the background instead of, or maybe in addition to, the border). But at least it's much easier to see the difference, so it's definitely an improvement.

This could be applied as is (in which case I could add a comment to the new function myself) or wait until you implement the proposed refactoring if you agree that it's worth doing it, please let me know.

Thanks again!

DietmarSchwertberger:
Thanks for reviewing. I have added a description to `PrepareDCForLabels`.

As of now, this PR contains just the fixes, which definitely should be in 3.1.7.

For the better visualisation of pressed headers, I can do a separate PR. It will probably take a few days. I would like to improve the overall behaviour to keep it pressed during the whole drag operation.
A change for the background colour would probably require this to be implemented in custom renderers. Just painting the borders is not very nice, but on the other hand does not do too much harm if it does not match what a custom renderer has drawn already.


I will probably look at your suggestion for `PrepareDCForScrollPos`. If it can be used also for `wxGridRowLabelWindow::OnPaint`, I will give it a try.

vadz:
Thanks, I'll (squash) merge this right now then.

PR #22404, opened 2022-05-06
MSW SpinCtrl: handle special keys (e.g. Ctrl+C)

DietmarSchwertberger:
Sorry, this is the same as PR22397, which was closed when I was re-creating my fork.

There is a wxPython bug report about menu entries like "Copy\tCtrl+C" having precedence over special character handling in the text control of SpinCtrl:
https://github.com/wxWidgets/Phoenix/issues/2148

The reason is that SpinCtrl does not have an implementation of `MSWShouldPreProcessMessage`.

This modification uses `MSWShouldPreProcessMessage(WXMSG* msg);` from `src/msw/textentry.c` also for `src/msw/spinctrl.cpp`.

### DIFF
--- a/include/wx/msw/spinctrl.h
+++ b/include/wx/msw/spinctrl.h
@@ -140,6 +140,10 @@ protected:
     void OnSetFocus(wxFocusEvent& event);
     void OnKillFocus(wxFocusEvent& event);
 
+    // returns true for special keys like "Ctrl+C" that should be handled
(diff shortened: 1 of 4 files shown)


vadz:
Thanks!

Will apply in a moment with minor changes (I prefer to keep this function out of the public header, but even so it's still tidier to use a `wx` prefix for it).

PR #22409, opened 2022-05-07
Fix (not) showing values for empty cells in wxDVC using wxDataViewVirtualListModel in wxGTK

### LINKED ISSUE #22359: Cells without values not correctly shown in wxDataViewCtrl with wxGTK 3.1.6
## Description
Cells without values are not correctly shown in wxDataViewCtrl with wxGTK 3.1.6.

This has initially been reported in Poedit: vslavik/poedit#746. Following comments from @vslavik and @vadz, I'm moving the ticket to wxWidgets.

In Poedit compiled against wxGTK 3.1.6, the icons indicating the presence of an issue or a comment on the translatable strings are incorrectly updated when scrolling up/down the list of strings. It looks like the strings without issue/comment "reuse" the icons of other visible strings with an issue or a comment.

Reverting commit [e067e42](https://github.com/wxWidgets/wxWidgets/commit/e067e4233caf691d52b52bd146568de697de8311) fixes the issue.

### Expected behaviour (wxGTK 3.1.5):

The warning/comment icons are correctly displayed.

https://user-images.githubusercontent.com/9394745/163816494-01653885-1d0a-442a-8507-6b89020ea8ab.mp4

### Actual behaviour (wxGTK 3.1.6):

The warning/comment icons are not correctly displayed - warning is displayed for string without warnings, the (wrong) warning icon for a given line changes to the comment icon when scrolling down.

https://user-images.githubusercontent.com/9394745/163816668-77f1c3c4-0860-482a-8370-5241dd539976.mp4

## Patch or snippet allowing to reproduce the problem

I'm not able to create such a snippet. I believe the relevant code in Poedit is located in [edlistctrl.cpp](https://github.com/vslavik/poedit/blob/v3.0.1-oss/src/edlistctrl.cpp); in particular at [line 336](https://github.com/vslavik/poedit/blob/v3.0.1-oss/src/edlistctrl.cpp#L336) in `PoeditListCtrl::Model::GetValueByRow` is the code that defines the icon to display. @vslavik: Please correct if I'm wrong.

## To Reproduce

Assuming Poedit 3.0.1 is installed under the environment mentioned below:

1. Decompress and open the following translation file: [zypper-master-fr.po.gz](https://github.com/vslavik/poedit/files/8505929/zypper-master-fr.po.gz)
2. (Optional) In order to get the same string ordering as in the screencasts above, configure View with the following:
    - Show warnings
    - Sort by file order
    - Group by context
    - Entries with error first
    - Untranslated entries first
3. Scroll down the string list

## Platform and version information

 - wxWidgets version: 3.1.6
 - wxWidgets port: wxGTK
 - OS and its version: openSUSE Tumbleweed (20220420)
 - For wxGTK only:
   + GTK version: 3.24.33
   + Which GDK backend is used: Wayland

vadz:
This should fix #22359 and, AFAICS, doesn't have any negative effects.

I couldn't find any real explanation about why was this check added in the first place, but there doesn't seem to be any performance reason to not make it even for the virtual list models: it is called for every item in the model, which makes it totally unsuitable for big numbers of items, but this was already the case before anyhow and the extra overhead of calling `HasValue()` for it shouldn't be noticeable compared to calling `GetValue()`. But please let me know if I'm missing anything here or, also, if there are any other situations in which this still doesn't fix the original problem.

### DIFF
--- a/src/gtk/dataview.cpp
+++ b/src/gtk/dataview.cpp
@@ -3199,14 +3199,6 @@ gtk_dataview_header_button_press_callback( GtkWidget *WXUNUSED(widget),
     return FALSE;
 }
 
-// Helper for wxGtkTreeCellDataFunc() below.
-static void wxGtkTreeSetVisibleProp(GtkCellRenderer *renderer, gboolean visible)
-{
-    wxGtkValue gvalue( G_TYPE_BOOLEAN );
-    g_value_set_boolean( gvalue, visible );
-    g_object_set_property( G_OBJECT(renderer), "visible", gvalue );
-}
-
 extern "C"
 {
 
@@ -3241,7 +3233,9 @@ static void wxGtkTreeCellDataFunc( GtkTreeViewColumn *WXUNUSED(column),
         visible = cell->PrepareForItem(wx_model, item, column);
     }
 
-    wxGtkTreeSetVisibleProp(renderer, visible);
+    wxGtkValue gvalue( G_TYPE_BOOLEAN );
+    g_value_set_boolean( gvalue, visible );
+    g_object_set_property( G_OBJECT(renderer), "visible", gvalue );
 }
 
 } // extern "C"
--- a/src/gtk/dataview.cpp
+++ b/src/gtk/dataview.cpp
@@ -3233,25 +3233,15 @@ static void wxGtkTreeCellDataFunc( GtkTreeViewColumn *WXUNUSED(column),
 
     wxDataViewModel *wx_model = tree_model->internal->GetDataViewModel();
 
-    if (!wx_model->IsVirtualListModel())
+    gboolean visible = wx_model->HasValue(item, column);
+    if ( visible )
     {
-        gboolean visible = wx_model->HasValue(item, column);
-        wxGtkTreeSetVisibleProp(renderer, visible);
+        cell->GtkSetCurrentItem(item);
 
-        if ( !visible )
-            return;
+        visible = cell->PrepareForItem(wx_model, item, column);
     }
 
-    cell->GtkSetCurrentItem(item);
-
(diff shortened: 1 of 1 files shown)


vslavik (approved):
Thanks! This looks good to me w.r.t. correctness.

Could it be further simplified by removing the `HasValue()` check and `if` statement, as `PrepareForItem()` already does that via`CheckedGetValue()`? I'm not saying it *should* be done, but checking if my understanding is correct and this condition is just an optimization.

I *think* that as it is now (and this logic is shared by some other ports), there's a very subtle, mostly academical difference between `!HasValue()` and not returning a value from `GetValue()`:

1. If `GetValue()` doesn't return a value, `PrepareForItem()` [still enables/disables cell state]( https://github.com/wxWidgets/wxWidgets/blob/8245ba96ee5ce3e19564eabc6976435ab9ae906f/src/common/datavcmn.cpp#L914) even for empty cells - apparently intentionally.
2. But if `HasValue()` returns false, `PrepareForItem()` is skipped altogether and enabled state is not adjusted.

vadz:
Yes, I think you're right, thanks, it looks better to simplify this even if I don't think this can make any difference in practice because `SetEnabled()` is only used by the custom renderers AFAICS and their `Render()` shouldn't be called anyhow if the cell is invisible. Done now.

BTW, concerning the "apparently intentionally" part, I think 361c6357b4 (Use wxDataViewRenderer::PrepareForItem() in all ports, 2015-08-29) just prevented the existing behaviour (but added an explicit comment about it), so I'm not totally certain if this is needed neither, but I'm definitely not going to change it without a good reason -- and this call is cheap, so it shouldn't result in any performance problems neither.

Issue #22420, opened 2022-05-11
wxGrid deleteRows() not working properly


roenck:
**Describe the bug**
Deletion of whole rows not working as expected for some build-combinations.

**Expected vs observed behaviour**
Delete a row removes it and the line is not shown anymore. Vs deleted lines appear 'darkgray' and not selectable.

**Patch or snippet allowing to reproduce the problem**
Hopefully should occur with a single wxGrid example.

**Platform and version information**
 - wxWidgets version you use: 3.1.5 and Git-Dev (from 2022-05-11)
 - wxWidgets port you use: wxMSW and wxGTK
 - OS and its version: Windows 10 and Ubuntu 20.04
 - For wxGTK only:
   + GTK version: 2.24.32
   + Which GDK backend is used: X11

Important, the issue occurs in combinations of platform and used libs and wx-releases:
---> Linux + wx3.1.5 + GTK2 ---> FAILED
---> Linux + wx3.1.5 + GTK3 ---> OK
---> Linux + wxGIT + GTK3 ---> FAILED
---> Windows + wxGIT + wxMSW --> FAILED

PBfordev:
I cannot reproduce using the grid sample, GIT master, and Windows 10.

I selected a row and used "Edit / Delete selected rows" menu command. The behaviour was as expected, i.e., the deleted row disappeared and the rows below it moved up. Tried with one row and with multiple rows with the same result. Tried also with the top row frozen, with row 2 hidden, with non-editable grid: Still the same result.

Is there anything special about your "non-deletable" rows?

roenck:
Tried to strip it to a minimal sample (Help -> About call the deleteRows-method).

[MinimalSampleGrid.zip](https://github.com/wxWidgets/wxWidgets/files/8672324/MinimalSampleGrid.zip)

![Grid_1](https://user-images.githubusercontent.com/100382302/167915591-99292f26-c00a-49f7-b91b-3c4301da2d45.png)
![Grid_2](https://user-images.githubusercontent.com/100382302/167915670-2ab49171-5ef2-404e-8ccf-192ee1409277.png)
![Grid_3](https://user-images.githubusercontent.com/100382302/167915686-79898511-419f-47cd-b2fa-3eff518a3983.png)

DietmarSchwertberger:
In grid.cpp `wxGrid::Redimension` there is a line:
` for ( rowPos = 0; rowPos < m_numCols; rowPos++ )`
This should be:
[...]

PR #22423, opened 2022-05-11
grid: fix row positions not updated when deleting rows (fix issue #22420)

DietmarSchwertberger:
Fixes issue #22420.


vadz:
Thanks Dietmar for noticing and fixing this so quickly!

Issue #22021, opened 2022-01-16
Revert or at least document regressions in wxImageList behaviour


vadz:
After the recent changes in 838e45fd9b58392df0936cfb2bfd19bf95d1645d and 8f08233a1322283958b3097099f03bbc09b796b4, images with alpha are not shown correctly when using `wxImageList` created with default value of `mask = true`, see [this thread](https://groups.google.com/g/wx-users/c/tv2diRGSxig/m/UHdI-eVdFQAJ). Ideal would be to restore the old behaviour, but if this is impossible we should at least assert or `wxLogDebug()` the explanation to prevent people from losing their time trying to understand why things don't work any longer.

Also see previous [discussion here](https://github.com/wxWidgets/wxWidgets/commit/838e45fd9b58392df0936cfb2bfd19bf95d1645d).

cc @a-wi

vadz:
@a-wi Just to avoid any duplication of work, I wanted to say that I'm going to fix this by converting mask to alpha in `wxImageList` rather than ignoring it entirely. Unfortunately I don't have a test case for the original bug (#19036), so I can only hope that this doesn't rebreak it again, but I'm relatively confident that it shouldn't.

Please let me know if you're, by chance, already working on this, but if not I'll try to do it myself soon as I'd really like to fix this before 3.1.6. Thanks!

a-wi:
At the moment I don't work on this so please proceed.

PR #22433, opened 2022-05-17
Revert to using mask in wxMSW wxImageList if this is what we have

vadz:
Always using alpha doesn't work in a special case of a fully transparent
bitmap: MSW renders a bitmap with all pixels having alpha value of 0 as
fully black for some reason and there doesn't seem anything that can be
done about this.

So revert to using mask, if we have it, instead of always converting it
to alpha as was done in 6feeed9fe9 (Handle transparency to the best of
our ability in wxImageList, 2022-05-05), see #22400.

Also rewrite the code in GetImageListBitmaps() to be hopefully more
clear by branching over HasAlpha() condition.

This commit is best viewed with git --color-moved
--color-moved-ws=ignore-all-space options.

----

@MaartenBent @dconnet This fixes the cases you mentioned in the comments for #22400 and doesn't break #22349 nor earlier #22021, so hopefully this should be the final version of this code.

### DIFF
--- a/src/msw/imaglist.cpp
+++ b/src/msw/imaglist.cpp
@@ -79,16 +79,13 @@ bool wxImageList::Create(int width, int height, bool mask, int initial)
     // (e.g. ILC_COLOR16) shows completely broken bitmaps
     flags |= ILC_COLOR32;
 
-    m_useMask = mask;
+    // we also need to always set ILC_MASK because, surprisingly, without it,
+    // completely transparent bitmaps, i.e. those with wxALPHA_TRANSPARENT
+    // value of alpha for all their pixels, don't work and appear completely
+    // black when drawn
+    flags |= ILC_MASK;
 
-    // For comctl32.dll < 6 always use masks as it doesn't support alpha.
-    //
-    // We also have to use masks when we don't have wxImage and wxDIB that are
-    // needed to handle alpha.
-#if wxUSE_WXDIB && wxUSE_IMAGE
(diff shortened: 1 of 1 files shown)


vadz:
Sorry, this still doesn't work, of course, because I completely forgot that we couldn't have both alpha and mask in the same image list. And as we don't know in advance whether we're going to have any completely transparent bitmaps or not, we really can't commit to using `ILC_MASK`.

So, finally, I am afraid I'm going to implement the horrible hack I mentioned before just because I see no other way to make things work in a reasonable way.

And reverting 6feeed9fe9 (Handle transparency to the best of our ability in wxImageList, 2022-05-05) is really not an option because we really, really want to use alpha and not the mask even for the image lists created with `mask = true` (in particular, although far from being only due to this, because using `mask = false` doesn't allow having fully transparent images neither).

PBfordev:
@vadz, I am not sure if all the transparency issues have been fixed.

When running the artprov sample and opening the resource browser dialog, in XPM-based arts the pixels that should be transparent are shown black (transparency is correct in the right panel):
wxartprov-black

The images are added to the list control via the image list with
https://github.com/wxWidgets/wxWidgets/blob/34fa234f48f30bd8fef0f442f55debceafa501bf/samples/artprov/artbrows.cpp#L32-L34

Using the latest master (as of 34fa234) on Windows 10.

vadz:
For the record, this should be fixed by #22487.

PR #22436, opened 2022-05-18
Fix for scaling of the bitmap size for wxBitmapComboBox

YuSanka:
Calculate m_usedImgSize from unscaled bitmap size instead of calculation from scaled control.

You can use **samples/widgets** > **BitmapCombobox** > **"Add widget icons"** for reproducing the issue.

If use this combobox on scaled Display, then space between bitmap and text increases.
bmpcb_bug

After fix bitmapcombobox works as expected
bmpcb_fix

### DIFF
--- a/src/common/bmpcboxcmn.cpp
+++ b/src/common/bmpcboxcmn.cpp
@@ -116,7 +116,7 @@ bool wxBitmapComboBoxBase::OnAddBitmap(const wxBitmapBundle& bitmap)
 {
     if ( bitmap.IsOk() )
     {
-        wxSize bmpDefaultSize = bitmap.GetPreferredLogicalSizeFor(GetControl());
+        wxSize bmpDefaultSize = bitmap.GetPreferredBitmapSizeAtScale(1.0);
         int width = bmpDefaultSize.GetWidth();
         int height = bmpDefaultSize.GetHeight();
 
@@ -219,7 +219,7 @@ void wxBitmapComboBoxBase::DrawItem(wxDC& dc,
 
         // Draw the image centered
         dc.DrawBitmap(bmp,
-                      rect.x + (m_usedImgSize.x-w)/2 + imgSpacingLeft,
+                      rect.x + (win->FromDIP(m_usedImgSize.x)-w)/2 + imgSpacingLeft,
                       rect.y + (rect.height-h)/2,
                       true);
     }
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, there is clearly a problem here, but I am not sure if this fix is correct. What are the units of `m_usedImgSize`? By default, it should use logical pixels, just as everything else, and in this case we shouldn't apply `FromDIP()` to it. I.e. I think that instead of adding more `FromDIP()`s around it, we should instead remove the one used in `DetermineIndent()`.

Could you please check if doing this would fix the problem?

YuSanka:
Unfortunately, no :(
Remove  FromDIP() around it in DetermineIndent() was my first idea. But if we move dialog to the display with different scale, than
DetermineIndent() will return value, which doesn't respect to the new scale.

vadz:
@YuSanka will you be able to look at this or should I try to do it? TIA!

YuSanka:
> This just means that we have to set things up so that `DetermineIndent()` is called when the DPI changes, i.e. add a `wxEVT_DPI_CHANGED` handler to this class and call it from there. Could you please try updating the PR to do it like this? TIA!

But it's already implemented in wxWidgets :)
If I understend everithing correct from the code, than the `DetermineIndent()` is called from 
`void wxBitmapComboBox::OnSize(wxSizeEvent& event)` after `wxEVT_DPI_CHANGED` handling

MaartenBent:
It doesn't work because once `m_usedImgSize` is assigned a value, it is never changed again. It should be updated when the DPI changes. The same for `m_fontHeight`.

I tried this patch (also simplified assigning `m`, and it seems to handle DPI changes correctly:

vadz:
Thanks Maarten, now I understand the problem and the solution looks very logical too.

@YuSanka Please check if the patch above works for you, TIA!

YuSanka:
Hello, guys.
Yes, this patch works for me too. Thanks, @MaartenBent .

vadz:
Sorry, I didn't test the previous patch and didn't realize it broke the unit tests. I've committed 212d59cd80 (Fix determining the size of wxBitmapComboBox bitmaps, 2022-05-25) now (without review, as I wanted to fix the tests a.s.a.p.), which should also improve the behaviour when all the bitmaps are not available in the same sizes.

Please let me know if you see any problems with this commit.

PR #22437, opened 2022-05-18
Fix showing fully fully transparent bitmaps in wxMSW wxImageList

vadz:
This was broken by 6feeed9fe9 (Handle transparency to the best of our
ability in wxImageList, 2022-05-05) as using alpha, rather than mask,
for these images resulted in alpha channel being just ignored.

Work around this by making at least one pixel not quite transparent in
this case.

This also makes things work for images using alpha channel with only 0
values, rather than mask covering the entire bitmap.

See #22400.

----

@MaartenBent @dconnet Sorry for all the notifications, but here is the other fix for the problem with fully transparent bitmaps. Please let me know if you see anything wrong with it (other than its intrinsic ugliness).

### DIFF
--- a/src/msw/imaglist.cpp
+++ b/src/msw/imaglist.cpp
@@ -213,6 +213,30 @@ wxImageList::GetImageListBitmaps(wxMSWBitmaps& bitmaps,
         if ( !img.HasAlpha() )
             img.InitAlpha();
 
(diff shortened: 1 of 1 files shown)


MaartenBent:
Thanks, this works good enough for me.

PR #22438, opened 2022-05-18
Fix wxGTK build with glib < 2.50

### LINKED ISSUE #22434: Compiling wxGTK on older systems (with glib < 2.50) is broken
Commit 2efa26cc1f (Merge branch 'gtk-log-suppress', 2022-05-17) breaks compiling wxGTK on older systems (with glib < 2.50).

This regression can be fixed by the following changes:
```patch
diff --git "a/src/gtk/notebook.cpp" "b/src/gtk/notebook.cpp"
index 1791b4b729..1bec3e3a83 100644
--- "a/src/gtk/notebook.cpp"
+++ "b/src/gtk/notebook.cpp"
@@ -427,12 +427,16 @@ wxNotebookPage *wxNotebook::DoRemovePage( size_t page )
     if ( !client )
         return NULL;
 
+#ifdef wxHAS_GLIB_LOG_WRITER
+
     // Suppress bogus assertion failures happening deep inside ATK (used by
     // GTK) that can't be avoided in any other way, see #22176.
     wxGTKImpl::LogFilterByMessage filterLog(
         "gtk_notebook_get_tab_label: assertion 'list != NULL' failed"
     );
 
+#endif // wxHAS_GLIB_LOG_WRITER
+
     // we don't need to unparent the client->m_widget; GTK+ will do
     // that for us (and will throw a warning if we do it!)
     gtk_notebook_remove_page( GTK_NOTEBOOK(m_widget), page );

```

vadz:
This was recently broken by bffcb88266 (Optionally detect not filtered
GTK log messages, 2022-05-12), see #22424.

Fix it by providing a trivial "do nothing" version of LogFilterByMessage
even if we can't implement it -- this seems to be preferable to using
preprocessor checks in the code using it.

Closes #22434.

---

@taler21 I couldn't actually test this with old glib as I don't have any sufficiently old systems at hand any more, please let me know if you can test it and confirm that it works (or if it doesn't, of course). TIA!

### DIFF
--- a/include/wx/gtk/private/log.h
+++ b/include/wx/gtk/private/log.h
@@ -128,7 +128,17 @@ private:
     wxDECLARE_NO_COPY_CLASS(LogFilterByMessage);
 };
 
-#endif // wxHAS_GLIB_LOG_WRITER
+#else // !wxHAS_GLIB_LOG_WRITER
+
+// Provide stubs to avoid having to use preprocessor checks in the code using
+// these classes.
+class LogFilterByMessage
+{
+public:
+    explicit LogFilterByMessage(const char* WXUNUSED(message)) { }
+};
+
+#endif // wxHAS_GLIB_LOG_WRITER/!wxHAS_GLIB_LOG_WRITER
 
 } // namespace wxGTKImpl
 


taler21:
Yes, it works. Thanks.

vadz:
Thanks for testing!

Issue #22439, opened 2022-05-19
Width of choice box for zoom in print preview too small


taler21:
Is there a good reason why the width of the `wxChoice` with the zoom levels is hardcoded?
In my opinion `wxDefaultSize` should be used, so that the `wxChoice` is sized according to its content.

```patch
diff --git "a/src/common/prntbase.cpp" "b/src/common/prntbase.cpp"
index 3089d00202..d44e99da15 100644
--- "a/src/common/prntbase.cpp"
+++ "b/src/common/prntbase.cpp"
@@ -1614,7 +1614,7 @@ void wxPreviewControlBar::CreateButtons()
         };
         int n = WXSIZEOF(choices);
 
-        m_zoomControl = new wxChoice( this, wxID_PREVIEW_ZOOM, wxDefaultPosition, wxSize(FromDIP(70),wxDefaultCoord), n, choices, 0 );
+        m_zoomControl = new wxChoice( this, wxID_PREVIEW_ZOOM, wxDefaultPosition, wxDefaultSize, n, choices, 0 );
         sizer.Add(m_zoomControl);
         SetZoomControl(m_printPreview->GetZoom());
 

```

Currently on some systems the box is too small.
For example under openSUSE 42.3 (KDE desktop with 'Adwaita' theme for GTK applications):
![print-preview-zoom-box-orig](https://user-images.githubusercontent.com/99262969/169275844-790a575e-db7c-4b73-80f4-d30e6854e4ba.png)

With my proposed changes it looks like this:
![print-preview-zoom-box-fixed](https://user-images.githubusercontent.com/99262969/169275882-049188c9-3e2b-471d-a8a8-44348b030fab.png)

By the way, what can also be seen is that the default width of the entire `wxPreviewControlBar` is too small, although sizers are used.
Any idea why?

vadz:
Yes, hard-coded `70` here is clearly wrong, I'll remove it, thanks! Amazingly, it seems to have survived since 9dff851556 (Use sizers, not constraints, 2003-04-17) which added this code...

As for the total size being too small, it's due to hard-coding it again, this time in the sample code. However just removing the hard-coded size is not enough, so I had to also make a couple of other changes to let the frame have a reasonable size by default.

All of those are part of #22441, please let me know if you see any problems with it. TIA!

taler21:
No problems, looks good now. Thanks!
There is only one minor thing, I am nitpicking now.
[...]

PR #22441, opened 2022-05-19
Print preview layout and appearance fixes

vadz:


vadz:
> I think the [example code in the documentation](https://github.com/wxWidgets/wxWidgets/blob/f43efa6b46caf091bb9f354080a12ee6aec3002e/docs/doxygen/overviews/printing.h#L60) should also be adapted to these changes, not only the printing sample.

Absolutely, thanks for noticing this, I've removed the size (and position) from the docs too.

taler21:
Since f0de5a93f8c, the size passed to the `wxPreviewFrame` constructor no longer has any effect.
I am wondering if this size should matter if specified explicitly. What do you think?

vadz:
> Since [f0de5a9](https://github.com/wxWidgets/wxWidgets/commit/f0de5a93f8cca76a0a073fc4900e7f37b0b6d624), the size passed to the `wxPreviewFrame` constructor no longer has any effect. I am wondering if this size should matter if specified explicitly. What do you think?

You're right once again, we should take the user-specified size into account. I've done this now, even though this was a bit cumbersome due to the unusual API of this class.

Please let me know if you see any other problems.

taler21:
Works fine now. Thanks.

Just a final note or question.
In commit 4925550 you only applied two changes of the three I [suggested](https://github.com/wxWidgets/wxWidgets/issues/22439#issuecomment-1133547699).
Was that intentional or is it a mistake?

PR #22442, opened 2022-05-19
Add wxWebViewEdge includes to monolithic builds too

MaartenBent:
Create a template with the additional webview includes and defines needed for Edge backend.
Use the template in the lib, dll, and monolithic build.

Fixes #19104

The bakefiles in the repo are generated with 0.2.13 but the latest version available is 0.2.12. So I had to manually modify the changed files in the bakefile installation directory. It seems to work, but maybe you can rerun bakefile before merging to make sure they are correct.

#19104 also suggests adding `--enable-webviewedge` to `configure.in`, I'll add that later and keep the PR a draft until then.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -989,8 +989,9 @@ MONODLL_CFLAGS = $(__monodll_PCH_INC) $(__INC_TIFF_BUILD_p) $(__INC_TIFF_p) \
 	-I$(top_srcdir)/src/stc/scintilla/include \
 	-I$(top_srcdir)/src/stc/scintilla/lexlib \
 	-I$(top_srcdir)/src/stc/scintilla/src -D__WX__ -DSCI_LEXER -DNO_CXX11_REGEX \
-	-DLINK_LEXERS -DwxUSE_BASE=1 -DWXMAKINGDLL $(PIC_FLAG) $(WX_CFLAGS) \
-	$(CPPFLAGS) $(CFLAGS)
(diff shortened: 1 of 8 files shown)


PBfordev:
Thanks for the quick fix, Maarten.

I tried building monolithic with both GCC and MSVC makefiles (in all combinations of shared/static and debug/release) and all the builds succeed.

I did not try with `wxUSE_WEBVIEW_EDGE_STATIC` set to `1`, as it seems this option is supported only when using CMake. If it is really so, perhaps that could be documented?

MaartenBent:
I haven't tried it either, but it is documented as:
https://github.com/wxWidgets/wxWidgets/blob/f43efa6b46caf091bb9f354080a12ee6aec3002e/interface/wx/webview.h#L365-L367
It uses `#pragma comment(lib, "WebView2LoaderStatic")`, so will only work with MSVC. So I think you can enable it when using the `.vcxproj` files and `makefile.vc`, but not when using GCC/Clang with `makefile.gcc`. I don't think it is limited to CMake.

PBfordev:
I see, thanks.

I knew it was MSVC-only but did not realize it was due to using its specific pragma. I just grepped bakefiles and makefiles for the define and found nothing, so I made a wrong assumption....

vadz:
Thanks for the fix! Do I understand correctly that this should be already merged or do you have any more planned changes here?

MaartenBent:
I'd like to add `--enable-webviewedge` to configure too before merging.

Issue #22432, opened 2022-05-17
wxGrid: assertion when dragging row/col 0 to corner label window


DietmarSchwertberger:
This can be reproduced in the grid sample:
 - enable row or col moving
 - drag the first row or col to the corner label window
 - an assertion is triggered

I think this is due to PR #22292 enabling mouse capture when dragging a row or col.
Previously drag moving was ended when the label window was left and therefore a LeftUp with row or col == -1 would not have occurerd.

I will work on this. It needs to be fixed for 3.1.7.

vadz:
Thanks for finding it and working on it!

DietmarSchwertberger:
PR is prepared.

I found the issue while looking for a better solution of the drop position check.
Currently a `DoEndMove` is called even for situations where finally no row is moved, e.g. when dragging a row to the first half of the next one or the second half of the previous one.
Still haven't found a nice implementation for this....

vadz:
FWIW I think we do need to call `DoEndMove()` in any case, as it resets `m_dragMoveRowOrCol`. But we probably shouldn't send the event, nor call `Set{Row,Col}Pos()` if the position didn't really change, should we?

DietmarSchwertberger:
Maybe.
Currently it's not called if `line == m_dragMoveRowOrCol`.
The check whether a move will happen is also required for deciding whether the insertion marker needs to be blue or grey. Currently it will often be drawn blue if the mouse is outside the original row/col.

PR #22292, opened 2022-04-09
grid usability: auto scrolling, cancel drag actions with ESCAPE

DietmarSchwertberger:
This is a first attempt.
For drag actions like resizing, selecting and moving, auto scrolling is implemented.
This will scroll by one scroll unit whenever a mouse event occurs and the mouse pointer is at the edge or outside the window.
Frozen rows/columns are handled.
I have seen that there is wxAutoScrollTimer. Maybe this could be used, but this is out of scope for me.
Maybe, the auto scroll code from `ProcessGridCellMouseEvent` could be refactored into a function that could also be used for rows and column. I will consider this, if the auto scroll functionality as such is accepted.

When a drag action is ongoing, the user can now cancel it with ESCAPE. Further dragging will be blocked until a mouse event occurs with left mouse button up. This is tracked by `m_canceledDragging`.

Please try out and provide feedback.

I have only tested on Windows and have not tested what happens if Right-To-Left is active.

PR #22443, opened 2022-05-20
wxGrid: fix row/col dragging to corner label (issue #22432)

DietmarSchwertberger:
Fix assertion when dragging row/col 0 to corner label (issue #22432).

(Regression from https://github.com/wxWidgets/wxWidgets/pull/22292 .)


vadz:
Thanks, makes sense to me and I'll merge it soon.

Thanks again for noticing and fixing this problem!

PR #22448, opened 2022-05-23
Make several changes in event-related docs

PBfordev:
Please verify that the "reverse order" change in event processing part (line 480 in `eventhandling.h`) is correct. It makes sense to me but it would be odd if I were the first to notice it after all these years....


vadz:
Thanks! I agree that "reverse" was missing there, even if the rest of the sentence hopefully made it clear how it worked. Perhaps we should use "LIFO" instead, or would this be too cryptic?

I'll merge this soon if there are no other comments.

PBfordev:
> Thanks! I agree that "reverse" was missing there, even if the rest of the sentence hopefully made it clear how it worked. Perhaps we should use "LIFO" instead, or would this be too cryptic?
> 
I'd rather not use the abbreviation here, even if the intended audience ought to be familiar with it.

PR #22449, opened 2022-05-23
wxBitmapBundle-related regressions

vslavik:
I'm creating this as draft PR because I realized I'm really not sure how to address some regressions in Poedit. I see two kind of issues so far:

- [x] Not showing bitmaps at all on macOS. I'm reasonably sure that 33db99a62f6941319e9d60e1f9e34889d37123c4 is the correct fix. For bitmaps loaded via `[NSImage imageNamed]`, they already supported multiple resolutions in `wxBitmap` and wxBitmap == wxBitmapBundle  for them. So just returning the NSImage regardless of expected size seems correct to me. Notice that this affects *all* bitmaps loaded via `wxArtProvider` with resources stored in app bundle on macOS.

- [ ] I'm seeing unexpected scaling of bitmaps with [now-horrible](https://github.com/wxWidgets/wxWidgets/commit/84cb293e712ba2db7a946a166a83e1d0fdec3297) (and previously ugly blurred, but recognizable) results in e.g. toolbar icons. I'm unsure how to best fix this and feel like I'm missing something important.

Generally speaking, resizing bitmaps or icons is *always* bad, regardless of whether it's e.g. bicubic (and so blurred) or nearest-neigbour (and so quite possibly outright damaged, e.g. removed vertical or horizontal edges in icons, sometimes making them unrecognizable). In sizes as small as this, the only safe resizing is scaling up by integer multiples. Even SVG rendering at different scale factors is going to produce questionable results.

Arguably, the entire point of `wxBitmapBundle`-like APIs is to avoid inevitably-ugly scaling of small bitmaps and it needs to take great care to avoid rescaling anything.

So if I look at the current code:

1. [Carefully picking the best size](https://github.com/wxWidgets/wxWidgets/blob/d90cb7511b07249d0cdd34278a2ab46e084817ab/src/msw/bmpbndl.cpp#L295) is great (this was the first thing I checked - this is pretty much how Poedit picks from its 1x, 2x and 1.5x manually-drawn icons on Windows). 

2. But that's not actually used if the bitmap was created through `wxArtProvider` (with a provider not overriding `CreateBitmapBundle()`).  And [this is not good](https://github.com/wxWidgets/wxWidgets/blob/d90cb7511b07249d0cdd34278a2ab46e084817ab/src/common/artprov.cpp#L169), see above. And it is coupled with [GetBitmap implementation](https://github.com/wxWidgets/wxWidgets/blob/d90cb7511b07249d0cdd34278a2ab46e084817ab/src/common/artprov.cpp#L174) which will happily resize to any size. 

3. Point 2 has consequences: bitmaps loaded this way (now represented as `wxBitmapBundle` internally) loose their intrinsic size. They become sponges willing to resize to any size. They don't signal their best size at all.

4. I think 84cb293e712ba2db7a946a166a83e1d0fdec3297 makes this worse, but don't care too deeply, because both situations were bad (and the commit's argument about consistency with other functions makes sense).

Here's a specific stacktrace showing where this goes wrong. I'm pretty sure it's not going to be the only one:

```
 	Poedit.exe!wxBitmapHelpers::Rescale(wxBitmap & bmp, const wxSize & sizeNeeded) Line 77	C++
>	Poedit.exe!wxArtProvider::RescaleOrResizeIfNeeded(wxBitmap & bmp, const wxSize & sizeNeeded) Line 338	C++
 	Poedit.exe!wxArtProvider::GetBitmap(const wxString & id, const wxString & client, const wxSize & size) Line 396	C++
 	Poedit.exe!`anonymous namespace'::wxBitmapBundleImplArt::GetBitmap(const wxSize & size) Line 174	C++
 	Poedit.exe!wxBitmapBundle::GetBitmap(const wxSize & size) Line 493	C++
 	Poedit.exe!wxToolBarToolBase::GetNormalBitmap(const wxSize & size) Line 153	C++
 	Poedit.exe!wxToolBar::Realize() Line 873	C++
 	Poedit.exe!wxToolBarXmlHandler::DoCreateResource() Line 246	C++
 	Poedit.exe!wxXmlResourceHandlerImpl::CreateResource(wxXmlNode * node, wxObject * parent, wxObject * instance) Line 1528	C++
 	Poedit.exe!wxXmlResourceHandler::CreateResource(wxXmlNode * node, wxObject * parent, wxObject * instance) Line 181	C++
 	Poedit.exe!wxXmlResource::DoCreateResFromNode(wxXmlNode & node, wxObject * parent, wxObject * instance, wxXmlResourceHandler * handlerToUse) Line 1095	C++
 	Poedit.exe!wxXmlResource::CreateResFromNode(wxXmlNode * node, wxObject * parent, wxObject * instance, wxXmlResourceHandler * handlerToUse) Line 372	C++
 	Poedit.exe!wxXmlResource::LoadToolBar(wxWindow * parent, const wxString & name) Line 511	C++
```

`wxToolBar::Realize()` asked for 24x24px bitmap, but mine was 34x34px or 26x26px (atypical, but who cares - some extra padding on top of 24/32 px icon).

I don't particularly care for Poedit, because I'll migrate it to `wxBitmapBundle` and sidestep this issue — at least I *think* that if `wxBitmapBundle` refuses to provide 24px version, `wxToolBar` will use what is given to it, as it used to do. But it's likely to affect at least some other code…

### DIFF
--- a/src/osx/core/bmpbndl.mm
+++ b/src/osx/core/bmpbndl.mm
@@ -148,7 +148,7 @@ wxSize wxOSXImageBundleImpl::GetPreferredBitmapSizeAtScale(double scale) const
 
 wxBitmap wxOSXImageBundleImpl::GetBitmap(const wxSize& WXUNUSED(size))
 {
-    return wxBitmap();
+    return wxBitmap(wxOSXGetImageFromBundleImpl(this));
 }
 
 wxBitmapBundle wxOSXMakeBundleFromImage( WXImage img)
--- a/src/common/artprov.cpp
+++ b/src/common/artprov.cpp
@@ -149,12 +149,13 @@ namespace
 class wxBitmapBundleImplArt : public wxBitmapBundleImpl
 {
 public:
-    wxBitmapBundleImplArt(const wxArtID& id,
+    wxBitmapBundleImplArt(const wxBitmap& bitmap,
+                          const wxArtID& id,
                           const wxArtClient& client,
                           const wxSize& size)
         : m_artId(id),
           m_artClient(client),
-          m_sizeDefault(GetValidSize(id, client, size))
+          m_sizeDefault(GetValidSize(bitmap, size))
     {
     }
 
@@ -176,24 +177,14 @@ public:
     }
 
 private:
-    static wxSize GetValidSize(const wxArtID& id,
-                               const wxArtClient& client,
-                               const wxSize& size)
+    static wxSize GetValidSize(const wxBitmap& bitmap, const wxSize& size)
     {
         // If valid size is provided, just use it.
         if ( size != wxDefaultSize )
             return size;
 
-        // Otherwise, try to get the size we'd use without creating a bitmap
-        // immediately.
-        const wxSize sizeHint = wxArtProvider::GetSizeHint(client);
-        if ( sizeHint != wxDefaultSize )
-            return sizeHint;
-
-        // If we really have to, do create a bitmap just to get its size. Note
-        // we need the size in logical pixels here, it will be scaled later if
-        // necessary, so use GetDIPSize() and not GetSize().
-        const wxBitmap bitmap = wxArtProvider::GetBitmap(id, client);
+        // Note we need the size in logical pixels here, it will be scaled
+        // later if necessary, so use GetDIPSize() and not GetSize().
         if ( bitmap.IsOk() )
             return bitmap.GetDIPSize();
 
@@ -439,9 +430,10 @@ wxBitmapBundle wxArtProvider::GetBitmapBundle(const wxArtID& id,
             // lower priority one: even if this means that the bitmap will be
             // scaled, at least we'll be using the expected bitmap rather than
             // potentially using a bitmap of a different style.
-            if ( provider->CreateBitmap(id, client, size).IsOk() )
+            const wxBitmap bitmap = provider->CreateBitmap(id, client, size);
+            if ( bitmap.IsOk() )
             {
-                bitmapbundle = wxBitmapBundle::FromImpl(new wxBitmapBundleImplArt(id, client, size));
+                bitmapbundle = wxBitmapBundle::FromImpl(new wxBitmapBundleImplArt(bitmap, id, client, size));
                 break;
             }
         }
--- a/src/common/artprov.cpp
+++ b/src/common/artprov.cpp
@@ -165,8 +165,9 @@ public:
 
     virtual wxSize GetPreferredBitmapSizeAtScale(double scale) const wxOVERRIDE
     {
-        // We have no preferred sizes.
-        return m_sizeDefault*scale;
+        // Any non-integer scaling or any downscaling will look ugly,
+        // only allow upscaling by a factor of 2.
+        return scale > 1.5 ? 2 * m_sizeDefault : m_sizeDefault;
     }
 
     virtual wxBitmap GetBitmap(const wxSize& size) wxOVERRIDE


vslavik:
> I'm unsure how to best fix this and feel like I'm missing something important.

In the best tradition of rubber duck debugging, I think I might actually know what would fix this: in `wxBitmapBundleImplArt::GetBitmap()`, call `wxArtProvider::GetBitmap()` in some (new) way that prevents the latter from calling `RescaleOrResizeIfNeeded()`…

I am not sure if it's worth doing, though - it's essentially for reasons of compatibility with  user-code `wxArtProvider` implementations...

src/osx/core/bmpbndl.mm
```diff
@@ -148,7 +148,7 @@ void WXDLLIMPEXP_CORE wxOSXBundleImplDestroyed(const wxBitmapBundleImpl* impl)
 
 wxBitmap wxOSXImageBundleImpl::GetBitmap(const wxSize& WXUNUSED(size))
 {
-    return wxBitmap();
+    return wxBitmap(wxOSXGetImageFromBundleImpl(this));
```
vadz, line 151: This looks correct to me, of course, thanks for fixing it, but I really wonder what happened here, i.e. clearly this could have never worked... Did `wxBitmapBundle` only work under Mac when used by the native control, due to Mac-specific code? @csomor please let me know if I'm missing something here.

BTW, comment before `wxOSXImageBundleImpl` declaration and for its ctor was copy-and-pasted from elsewhere and doesn't make any sense...

vadz:
(3) definitely doesn't seem right, so I think we need to change `wxBitmapBundleImplArt::GetPreferredBitmapSizeAtScale()` to return the closest (larger?) integer multiple of the requested size. And if we do this, shouldn't everything already work as expected, without further changes to `GetBitmap()`? As it will be called with the appropriate size, ` wxArtProvider::GetBitmap()` won't be doing any resizing, or am I missing something here?

And this is definitely important because I'm afraid that in practice we will have to live with user-defined `wxArtProvider` implementations for quite some time, as switching to `wxBitmapBundle` requires a non-trivial effort (I know as I made it for some -- but even in my case, not all -- of my projects). So compatibility is definitely important, the very last thing we want is for applications to look worse in high DPI with wx 3.2 than they looked before.

src/osx/core/bmpbndl.mm
csomor: @vadz no, you are right, when I copied over code from my old implementation, this got messed up, I guess since macOS code didn't need `wxBitmaps` anymore but used `wxBitmapBundles` for the native controls, thus retrieving the `NSImage` that was used for creating a `wxOSXImageBundleImpl`, I never came back to this omission. Thanks @vslavik

vslavik:
> (3) definitely doesn't seem right, so I think we need to change wxBitmapBundleImplArt::GetPreferredBitmapSizeAtScale() to return the closest (larger?) integer multiple of the requested size.

What if the art provider is capable of providing intermediate sizes like 1.25x or 1.5x on Windows? This would be needlessly restrictive then. But see below. 

 > As it will be called with the appropriate size, wxArtProvider::GetBitmap() won't be doing any resizing, or am I missing something here?

This won't help. My above suggestion was wrong too. I characterized the "sponginess" wrongly — the actual issue is that `wxBitmapBundleImplArt::GetValidSize()` uses `wxArtProvider::GetSizeHint()` to set a size it then enforces. The size hints were originally just *hints* but are now treated as *constraints*.  

8ea845af fixes that (buy you may not like it)

9730d550 limits scaling to 2x to prevent ugliness.

---

- [ ] Superscaling of icons in HiDPI-aware art providers

There's an entire class of regressions if user art provider is HiDPI aware to some extent. Poedit's one supports HiDPI, but only per-process. Its art provider returns properly scaled bitmaps (including choice logic similar to wxMSW's referenced above, so any scaling of the returned bitmaps would be bad). And they then get — sometimes — scaled up *again* by wx. Here, have a sampling:

Screen Shot 2022-05-24 at 13 05 28

That's unmodified Poedit running on Windows with 200% scaling. You can see that all the icons were loaded in 2x resolution as they should be. Some of them are OK (most of the toolbar, list, recent files). Some are upscaled and effectively show as 4x (the icon in *Translation* section, Sync toolbar icon, sidebar icons, welcome screen action icons, the close icon in welcome window). That one icon in the toolbar is set differently at runtime from the rest (loaded via XRC), that probably explains the difference. The rest are `wxStaticBitmap`, `wxBitmapButton` and owner-drawn `wxCommandLinkButton` (with the bitmap being drawn with and superscaled by `wxDC::DrawBitmap`).

I don't think 100% backward compatibility is possible here: if wx assumes user code was HiDPI-unaware (which seems to be what is prioritized now), it cannot avoid this superscaling without *some* user code updates. We could e.g. use PNG's PPI information (i.e. treat 72ppi as 1x, 144ppi as 2x etc.) but tools rarely set this correctly or even show, so that's a recipe for mystifying bugs down the road. Or the user code that is HiDPI-aware will need to at least set appropriate logical pixels size when loading bitmaps to work with wx-3.2.

vadz:
> This won't help. My above suggestion was wrong too. I characterized the "sponginess" wrongly — the actual issue is that `wxBitmapBundleImplArt::GetValidSize()` uses `wxArtProvider::GetSizeHint()` to set a size it then enforces. The size hints were originally just _hints_ but are now treated as _constraints_.

Yes, this does seem wrong now that you point it out.

> [8ea845a](https://github.com/wxWidgets/wxWidgets/commit/8ea845af881ba59739367362027f66c51fcd15a8) fixes that (buy you may not like it)

I think the idea here is correct, we should use the actual bitmap size. To be honest, I have no idea why did I want to postpone actually creating the bitmap here so much (the comment says that we "create a bitmap" only if "we really have to", but this makes no sense considering that we had already just created it -- but I suspect this must not have been the case in the original version of this code, as it already undergone several rounds of changes to ensure the correct priority of overridden `CreateBitmap()` and `CreateBitmapBundle()` of the stacked providers).

But I'd do this slightly differently and still avoid passing `bitmap` to the bundle ctor and instead just took care of setting the `size` correctly in the calling code: it already has the bitmap, so let it just pass its size. This would get rid of the ugly hardcoded `16,16` size which is actually never used now.

The only thing I'm not sure about here is whether we should use the caller-provided `size` at all (as your patch still does). I believe that we shouldn't and should always use the actual bitmap size as the base size, but I could be confused again here... Could you please double-check my reasoning?

> [9730d55](https://github.com/wxWidgets/wxWidgets/commit/9730d550215777dcf9a24261f480921d2344c26c) limits scaling to 2x to prevent ugliness.

This is a bit academic (but not completely so, at least one person on wx-dev actually used 3x scaling), but shouldn't we allow any integer scaling here?


>     * [ ]  Superscaling of icons in HiDPI-aware art providers
> 
> 
> There's an entire class of regressions if user art provider is HiDPI aware to some extent. Poedit's one supports HiDPI, but only per-process. Its art provider returns properly scaled bitmaps (including choice logic similar to wxMSW's referenced above, so any scaling of the returned bitmaps would be bad). And they then get — sometimes — scaled up _again_ by wx. Here, have a sampling:
> 
> Screen Shot 2022-05-24 at 13 05 28

Damn, this doesn't look good at all. I could live with all the bitmaps being scaled wrongly, but I just don't see how could scaling only some of them badly could be right :-(

> That's unmodified Poedit running on Windows with 200% scaling. You can see that all the icons were loaded in 2x resolution as they should be. Some of them are OK (most of the toolbar, list, recent files). Some are upscaled and effectively show as 4x (the icon in _Translation_ section, Sync toolbar icon, sidebar icons, welcome screen action icons, the close icon in welcome window). That one icon in the toolbar is set differently at runtime from the rest (loaded via XRC), that probably explains the difference. The rest are `wxStaticBitmap`, `wxBitmapButton` and owner-drawn `wxCommandLinkButton` (with the bitmap being drawn with and superscaled by `wxDC::DrawBitmap`).

If you can make any patches reproducing these particular problems, I'd like to have a look at them to see if we can do anything. I know that it would be simpler for you if I just built Poedit directly, but I suspect it might take a bit of time and I'm very short of it as I still have one major thing to finish before the scheduled 3.1.7 release date and there is really not much time left...

> I don't think 100% backward compatibility is possible here: if wx assumes user code was HiDPI-unaware (which seems to be what is prioritized now), it cannot avoid this superscaling without _some_ user code updates.

Yes, I agree that we can't always avoid it, the user code must set the scale factor correctly for the bitmaps that are already scaled because we do assume it's 1 by default (and I think it's the only reasonable assumption, even if it's clearly not perfect). But why are not _all_ bitmaps consistently upscaled in Poedit? If they were, it would be simple enough to understand what the problem is, but I'm pretty sure that if another application comes up like this after being recompiled with wx 3.2, people are just going to say that wx is broken instead of even trying to understand what exactly went wrong. It would be nice if we could somehow avoid or at least detect this.

If nothing else, we should clearly document it in the "incompatible changes" section, but I'm not even sure what should we say there. Would you have some idea of what would it have helped you to see there when upgrading wx? E.g. what advice should we give ("switch to overriding CreateBitmapBundle()"?) and to which situations exactly should it apply (any custom art provider using the current DPI?)?

Thanks again!

vslavik:
> If you can make any patches reproducing these particular problems

By the point I isolate the issues, I can as well fix them. I am unlikely to have much time for it *right now* - I have a deadline of my own. 

> But why are not all bitmaps consistently upscaled in Poedit?

Different code paths, I assume. Some come from icons (the recent files list) and are OK because of that. Some are generated, some loaded from wxArtProvider, some (the big toolbar icon) are a mystery. I'll investigate.

> Would you have some idea of what would it have helped you to see there when upgrading wx? E.g. what advice should we give ("switch to overriding CreateBitmapBundle()"?) and to which situations exactly should it apply (any custom art provider using the current DPI?)?

I think "any code that handles different-resolution bitmaps" and the advice is to set the scale factor correctly. Longer-term, to migrate to `wxBitmapBundle`, assuming that guarantees no algorithmic scaling will take place (I'm not sure about that, I was a mess and wasn't paying much attention to wx lately).

*However*, this is where I am currently mystified: Poedit's icons loading was mostly fixed by a single `SetScaleFactor()` call, but *it doesn't matter what value I pass to it*. I tried running Windows at 175% scaling, for which Poedit uses 1.5x bitmaps, and the output was *exactly the same* whether I called SetScaleFactor(1.5) or SetScaleFactor(1.75)...

vadz:
> > If you can make any patches reproducing these particular problems
> 
> By the point I isolate the issues, I can as well fix them. I am unlikely to have much time for it _right now_ - I have a deadline of my own.

I'd really like to keep to the schedule and release 3.1.7 in ~10 days. It would be very tempting to let it slip just a bit, of course, but then 3.2.0 won't happen in the beginning of July neither, as currently planned, and then it will probably get postponed to September...

So I guess we'll just have to live with this PR, possibly changed as I proposed above -- what do you think of doing it like this, especially the "always use the actual bitmap size" part?

> > But why are not all bitmaps consistently upscaled in Poedit?
> 
> Different code paths, I assume. Some come from icons (the recent files list) and are OK because of that. Some are generated, some loaded from wxArtProvider, some (the big toolbar icon) are a mystery. I'll investigate.

TIA!
 
> _However_, this is where I am currently mystified: Poedit's icons loading was mostly fixed by a single `SetScaleFactor()` call, but _it doesn't matter what value I pass to it_. I tried running Windows at 175% scaling, for which Poedit uses 1.5x bitmaps, and the output was _exactly the same_ whether I called SetScaleFactor(1.5) or SetScaleFactor(1.75)...

When using `wxBitmapBundle`, this would make sense -- it would use 1.5x bitmaps rather than scaling them to 1.75 anyhow. But if not using it, which is what you've been testing, if I understand correctly, then I don't know neither.

vslavik:
> So I guess we'll just have to live with this PR, possibly changed as I proposed above -- what do you think of doing it like this, especially the "always use the actual bitmap size" part?

I think it's correct (and my first instinct was to pass the size, not the bitmap, too), but let me complicate this further:

At this point, why even do the delay-loading at all? Why not just pass the actual *bitmap* to `wxBitmapBundleImplArt` and scale it there? Why re-load again through `wxArtProvider` later if we already obtained the bitmap? [^1]

And at that point, why even *have* `wxBitmapBundleImplArt`? If all it does now is integer upscaling, why not have that in `wxBitmapBundle` itself or in its MSW implementation (the only one that uses complicated scaling, I think?)? Where even non-wxArtProvider legacy code would benefit from integer upscaling?

[^1]: Notice that the way it is done now is subtly incorrect too, because art providers can be added or removed at runtime. Consider this situation: 1. retrieve a bitmap with `wxArtProvider::GetBitmapBundle()`, 2. modify provider chain with `wxArtProvider::Push()` and add a new provider that changes that specific bitmap, 3. draw the previously obtained bitmap. Previously, this would render the old bitmap, but will render the new one now due to lazy loading.

vadz:
Yes, I kind of lost the plot here, thanks for making me realize it. Having `wxBitmapBundleImplArt` only makes sense if we want to use `wxArtProvider` to get the bitmaps of different sizes and I originally wanted to do it because it seemed like we ought to be able to use both 16px and 32px bitmaps if a (custom) art provider has both of them. If we're always going to scale the first random bitmap we get from the art provider ourselves, there is no point for doing all this, we could just as well use `wxArtProvider::GetBitmap()` directly as this will get converted to `wxBitmapBundle` using this bitmap by any control using bitmaps and will end up being scaled appropriately anyhow.

Unfortunately Poedit example shows that sometimes asking the art provider for different sizes of the bitmap can be worse than scaling it. And the trouble is that we can't know it in advance.

So now I just don't know what to do any longer. Perhaps we should indeed just keep things simple and return a bundle consisting of a single bitmap from `GetBitmapBundle()` as a fallback and tell people that they need to override `CreateBitmapBundle()` in their custom art providers if they want to support bitmaps in multiple resolutions. OTOH if you already have code returning bitmaps of different sizes from your overridden `CreateBitmap()` it would be nice if you could somehow just make it work with wx 3.2 without rewriting it to override `CreateBitmapBundle()`. And I think that in many cases using the hint size as the normal size should work, but it's clearly going to break down badly in some cases too, as Poedit shows. But there just doesn't seem to be any way to distinguish between these cases with the existing API and adding new API wouldn't help the existing code.

To summarize, I guess my idea of supporting the existing custom art providers with multiple bitmap sizes just can't work and we should drop `wxBitmapBundleImplArt` entirely and simply return a bundle containing the single bitmap. Does anybody see anything better to do here? If not, I'll only take 33db99a62f6941319e9d60e1f9e34889d37123c4 from this PR and just remove `wxBitmapBundleImplArt` entirely.

vslavik:
> Does anybody see anything better to do here?

We could query the art provider for various sizes to try to enumerate what it supports, but that would break if the art provider did any resizing of its own - and even my past self wrote than in Poedit's. So I think having some sort of a migration guide is the best we can do.

There are other situations I've since identified, illustrated above[^1][^2] that don't really have a solution.

[^1]: That deformed X close button is rendered by creating a `wxBitmap` and drawing to it with `wxMemoryDC`. The bitmap has scale factor of 1 and is enlarged because of that. But changing `wxMemoryDC` scale factor to the current screen's one ("current" being tricky in itself), that would break HiDPI-unaware code.

[^2]: I have twice-enlarged icons in recent files menus in code where I do everything right: don't hardcode the size, use `wxSYS_SMALLICON_X` (returns 32px in 200% zoom case on win32). Use `wxIcon` with this explicit size to load. Yet, despite icons being multi-image, its conversion to `wxBitmapBundle` results in a single bitmap with 32px size and scale factor of 1 and not 2 as it should be. Again something trivial for me to fix in user code, but hard to handle correctly by default in all cases.

I now have a better understanding of `wxBitmapBundle` implementations and I while I understand why there are different implementations under the hood, I think they do too much, or differ in too much. In particular, the *logic* for picking the correct size should be predictable and platform-dependent and not dependent on implementation opaque to the user.

I've counted 3 different scaling behaviors so far:

1. `wxBitmapBundleImplArt` which I attempted to fix above. Removing it will actually make some of Poedit's code *worse*.
2. The default `wxBitmapBundleImplSet` implementation, which does scale to fit exact size and is prone to producing corrupted images
3. `wxBitmapBundleImplRC` which has the best logic for MSW and which I initially misread as applying to MSW platform and not only resources.

Contrary to my initial expectations, *this* is actually the cause of that one misbehaving toolbar button in the screenshot above:
1. Other bitmaps are loaded in XRC code and go through `wxBitmapBundle` code path. They're `wxBitmapBundleImplArt` under the hood.
2. The sync bitmap is read with `wxArtProvider::GetBitmap()`, so only the single bitmap is preserved and when the bitmap is passed from user code to `wxToolBar` and again converted to `wxBitmapBundle` internally, it uses `wxBitmapBundleImplSet`
3. The latter scales freely, while the former is heavily constrained with my patches.

Crucially, *this is completely opaque to user code*. It was hard to figure out what was going on -- and I *like* digging through library internals.

I think it's important to have a single (on a particular platform, it's OK to behave differently on MSW and OSX) `GetPreferredBitmapSizeAtScale()` function used by all bundle implementations - "preferred size at scale" is something that depends on OS conventions or toolkit capabilities, but is not a property of the mechanism for loading the bitmaps.  And [this logic](https://github.com/wxWidgets/wxWidgets/blob/d90cb7511b07249d0cdd34278a2ab46e084817ab/src/msw/bmpbndl.cpp#L192), perhaps extended with the possibility of doing 2x, 3x etc. upscaling instead of arbitrary, seems best.

vadz:
The logic in `wxBitmapBundleImplRC` should be the same as in `wxBitmapBundleImplSet`, are they really different? If so, they must have got out of sync after some change to the one or the other, but initially they were definitely intended to produce the same results. The only difference is that the latter has the bitmaps themselves, while the former has just their sizes, which is why they can't reuse the same code (well, they could, if we factored this logic out, and I might do that, but this is just an implementation details). Do you have a concrete example in which they return different results for the sets of bitmaps of the same sizes? Right now I don't see any, but I could well be missing something...

Other questions:

> wxBitmapBundleImplArt which I attempted to fix above. Removing it will actually make some of Poedit's code worse.

But using it, in the current master, clearly breaks Poedit too. Is having it really better? We must take some decision here before 3.1.7 and I'm still not sure what to do :-(

> I have twice-enlarged icons in recent files menus in code where I do everything right: don't hardcode the size, use `wxSYS_SMALLICON_X` (returns 32px in 200% zoom case on win32). Use `wxIcon` with this explicit size to load. Yet, despite icons being multi-image, its conversion to `wxBitmapBundle` results in a single bitmap with 32px size and scale factor of 1 and not 2 as it should be.

This looks like "just" a bug in wxMSW, the loaded icon should have the scale factor of 2. The problem is that `wxICOResourceHandler::LoadIcon()` already tries to set the correct scale factor for the icons loaded by `::LoadIcon()`, so I'm not sure why this doesn't happen. It would be really great if you could please check what goes on in this code (I suspect it would be faster for you to do it than to explain to me how exactly to reproduce it, but I could be wrong, in which case please do explain it, TIA!).

vadz:
> > The logic in wxBitmapBundleImplRC should be the same as in wxBitmapBundleImplSet, are they really different? If so, they must have got out of sync after some change to the one or the other, but initially they were definitely intended to produce the same results.
> 
> I was apparently confused by something else. There _is_ a difference, in the last line, but that wasn't what I (thought I) was seeing. The last line in `wxBitmapBundleImplSet` constrains upscaling and avoids it for scaling of less than 1.5x; the `wxBitmapBundleImplRC` implementation doesn't.

Yes, you're right, they did get out of sync. I've created #22481 to ensure that this doesn't happen again. If you don't have time to read anything below, please test this PR with Poedit and let me know what do you think!

Of course, if you do have time to read it (at least the tests part), it would be very much appreciated too.

> > Is having [wxBitmapBundleImplArt] really better? We must take some decision here before 3.1.7 and I'm still not sure what to do :-(
> 
> It isn't better to have it, wxBitmapBundleImplArt should be removed.

OK, I'll do this in another PR after a bit more testing.

> I hope I'm not confusing something again, but I think there are two issues with the `GetPreferredBitmapSizeAtScale()` algorithm used:
> 
>     1. It allows arbitrary upscaling. At 200% zoom setting, it will yield sharp icons, but at 225%, they'll be smudged, because they were upscaled to the requested size. But the extra 12% of size are barely noticeable, it would be better to use a perfectly scaled 200% version. That's caused by the last line above. I think running the loop again, multiplying all existing sizes by 2 (3, 4, …) would yield better results - no fractional upscaling or downscaling.

Thanks, this is clearly a good point and I've done this in the same PR too. This required changing more than `GetPreferredBitmapSizeAtScale()`, however, as in addition to returning the correct scale from it, we now must also choose the correct bitmap to scale: when the preferred scale is 3, we must scale the normal bitmap by 3, and not the double-sized bitmap by 1.5 (or, best of all, 1.5x bitmap by 2 if we have it, of course).

>     2. I think it's better to break the tie in favor of the smaller bitmap after all - that's [Poedit's logic](https://github.com/vslavik/poedit/blob/8905e774ad4d3f40d569efe299249f26f83ed7bd/src/hidpi.cpp#L71). It works well in practice: use 1x bitmaps for the 125% zoom factor, use 1.5x bitmaps for 125-175% factors, where the larger bitmap would be a bit too large. If you only have 1x and 2x, that would mean using 1x for 150%, which I think is better than oversized 200% icons. If you only have 1x, then combined with the point above, no upscaling would take place at 150%, which seems better to me too.

OK, I'm less sure about this, but let's try doing it like this -- this is 30fb836136 in the same PR.
 
> Note that the SVG implementation is [naive too](https://github.com/wxWidgets/wxWidgets/blob/6268b5c619ac7bd7479c0912275bce89d67d1f23/src/generic/bmpsvg.cpp#L155). You ideally want SVG icons snapped to a pixel grid too otherwise you get smudged strokes when drawn "in between pixels". It too would benefit from snapping to "nicer" scale factors.

Here I'm even less sure, IMO there should be some way to say that you want to have bitmaps of the correct size when using SVG. But OTOH sometimes you clearly still want to restrict it to (semi-?)integer factors too. At least with the new DoGetPreferredSize() you can easily define your own custom bundle implementation doing this, but it's still not very convenient, of course.
 
> > This looks like "just" a bug in wxMSW, the loaded icon should have the scale factor of 2. The problem is that wxICOResourceHandler::LoadIcon() already tries to set the correct scale factor for the icons loaded by ::LoadIcon(), so I'm not sure why this doesn't happen.
> 
> Ah, that would be because I don't use `wxICOResourceHandler` but `wxICOFileHandler` (via `wxIcon(filename, wxBITMAP_TYPE_ICO, desiredSize, desiredSize)` ctor).

I don't know what can we do about this one, we don't seem to have any way of knowing which scale factor should be used here because we can't distinguish loading large icon in normal DPI from loading small icon (with the same size) in high DPI, can we?

vadz:
Sorry, I've changed my mind about `wxBitmapBundleImplArt` again and added 16c2c95cdf (Use integer upscaling only in wxBitmapBundleImplArt, 2022-06-02) to #22481, explaining why I think it can still be useful in a comment.

I've also pushed the first commit here now and the commit above replaces the other ones (and even if we don't apply it, we're still not going to apply these ones neither as the only reasonable alternative seems to be removing `wxBitmapBundleImplArt` entirely), so I think this PR can be closed now -- but please reopen if you disagree, of course.

And I'd very much welcome any comments and testing in the other PR. TIA!

vslavik:
> Here I'm even less sure, IMO there should be some way to say that you want to have bitmaps of the correct size when using SVG. But OTOH sometimes you clearly still want to restrict it to (semi-?)integer factors too. At least with the new DoGetPreferredSize() you can easily define your own custom bundle implementation doing this, but it's still not very convenient, of course.

Yes, being able to customize the behavior is the best we can hope for. 

>  > Ah, that would be because I don't use wxICOResourceHandler but wxICOFileHandler (via wxIcon(filename, wxBITMAP_TYPE_ICO, desiredSize, desiredSize) ctor).
>
> I don't know what can we do about this one, we don't seem to have any way of knowing which scale factor should be used here because we can't distinguish loading large icon in normal DPI from loading small icon (with the same size) in high DPI, can we?

I don't think we can; I as much as said that above. A good migration guide that alerts to such issues and says what to do is IHMO the only realistic way to handle cases like this.

PR #22451, opened 2022-05-24
Update wxButtonXmlHandler

Randalphwa:
This updates the XRC **wxButtonXmlHandler**. It adds the same support for additional bitmaps that is in the **wxBitmapButtonXmlHandler**. It adds the missing wxBU_NOTEXT style. It adds a `margins` property that if specified will call `button->SetBitmapMargins()`.

No existing functionality was changed, so this should have zero impact on any existing XRC files.

For the schema, I used the newer names for the additional bitmaps -- there isn't any legacy XRC files to support in this case. However in the actual code, I do still check for the older names just in case a user simply changes the name of the class from wxBitmapButton to wxButton.

I compared using a C++ generated dialog versus loading the dialog from XRC and verified that they visually looked the same. I tested the `pressed` bitmap to verify the bitmap changed when clicked.

### DIFF
--- a/docs/doxygen/overviews/xrc_format.h
+++ b/docs/doxygen/overviews/xrc_format.h
@@ -907,6 +907,21 @@ Example:
     Label to display on the button (may be omitted if only the bitmap or stock ID is used).}
 @row3col{bitmap, @ref overview_xrcformat_type_bitmap,
     Bitmap to display in the button (optional).}
+@row3col{pressed, @ref overview_xrcformat_type_bitmap,
+     Bitmap to show when the button is pressed (default: none, same as @c bitmap).
+     This property exists since wxWidgets 3.1.6, but the equivalent (and still
+     supported) "selected" property can be used in the older versions.}
+@row3col{focus, @ref overview_xrcformat_type_bitmap,
+     Bitmap to show when the button has focus (default: none, same as @c bitmap).}
+@row3col{disabled, @ref overview_xrcformat_type_bitmap,
+     Bitmap to show when the button is disabled (default: none, same as @c bitmap).}
+@row3col{current, @ref overview_xrcformat_type_bitmap,
+     Bitmap to show when the mouse cursor hovers above the bitmap (default: none, same as @c bitmap).
+     This property exists since wxWidgets 3.1.6, but the equivalent (and still
+     supported) "hover" property can be used in the older versions.}
+@row3col{margins, @ref overview_xrcformat_type_size,
+    Set the margins between the bitmap and the text of the button.
+    This method is currently only implemented under MSW. If it is not called, a default margin is used around the bitmap.}
 @row3col{bitmapposition, @c wxLEFT|wxRIGHT|wxTOP|wxBOTTOM,
     Position of the bitmap in the button, see wxButton::SetBitmapPosition() (default: wxLEFT).}
 @row3col{default, @ref overview_xrcformat_type_bool,
--- a/docs/doxygen/overviews/xrc_format.h
+++ b/docs/doxygen/overviews/xrc_format.h
@@ -908,20 +908,16 @@ Example:
 @row3col{bitmap, @ref overview_xrcformat_type_bitmap,
     Bitmap to display in the button (optional).}
 @row3col{pressed, @ref overview_xrcformat_type_bitmap,
-     Bitmap to show when the button is pressed (default: none, same as @c bitmap).
-     This property exists since wxWidgets 3.1.6, but the equivalent (and still
-     supported) "selected" property can be used in the older versions.}
+     Bitmap to show when the button is pressed (default: none, same as @c bitmap). @since 3.1.7}
 @row3col{focus, @ref overview_xrcformat_type_bitmap,
-     Bitmap to show when the button has focus (default: none, same as @c bitmap).}
+     Bitmap to show when the button has focus (default: none, same as @c bitmap). @since 3.1.7}
 @row3col{disabled, @ref overview_xrcformat_type_bitmap,
-     Bitmap to show when the button is disabled (default: none, same as @c bitmap).}
+     Bitmap to show when the button is disabled (default: none, same as @c bitmap). @since 3.1.7}
 @row3col{current, @ref overview_xrcformat_type_bitmap,
-     Bitmap to show when the mouse cursor hovers above the bitmap (default: none, same as @c bitmap).
-     This property exists since wxWidgets 3.1.6, but the equivalent (and still
-     supported) "hover" property can be used in the older versions.}
+     Bitmap to show when the mouse cursor hovers above the bitmap (default: none, same as @c bitmap). @since 3.1.7}
 @row3col{margins, @ref overview_xrcformat_type_size,
     Set the margins between the bitmap and the text of the button.
-    This method is currently only implemented under MSW. If it is not called, a default margin is used around the bitmap.}
+    This method is currently only implemented under MSW. If it is not called, a default margin is used around the bitmap. @since 3.1.7}
 @row3col{bitmapposition, @c wxLEFT|wxRIGHT|wxTOP|wxBOTTOM,
     Position of the bitmap in the button, see wxButton::SetBitmapPosition() (default: wxLEFT).}
 @row3col{default, @ref overview_xrcformat_type_bool,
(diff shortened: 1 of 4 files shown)


docs/doxygen/overviews/xrc_format.h
```diff
@@ -907,6 +907,21 @@ pseudo-class. @c ownerdrawnitem objects have the following properties:
     Label to display on the button (may be omitted if only the bitmap or stock ID is used).}
 @row3col{bitmap, @ref overview_xrcformat_type_bitmap,
     Bitmap to display in the button (optional).}
+@row3col{pressed, @ref overview_xrcformat_type_bitmap,
+     Bitmap to show when the button is pressed (default: none, same as @c bitmap).
+     This property exists since wxWidgets 3.1.6, but the equivalent (and still
```
vadz, line 912: Is it correct to speak about the old property here? AFAICS it was never supported for `wxButton`, so this is misleading. IMHO we should just say that this property is new in 3.1.7.

misc/schema/xrc_schema.rnc
```diff
@@ -473,8 +473,8 @@ t_bitmap     = t_url?,
 t_font       = (
                    [xrc:p="o"] element size         {_, t_float }* &
                    [xrc:p="o"] element style        {_, ("normal" | "italic" | "slant") }* &
-                   [xrc:p="o"] element weight       {_, ("normal" | "thin" | "extralight" | "light" | 
-                                                         "medium" | "semibold" | "bold" | "extrabold" | 
+                   [xrc:p="o"] element weight       {_, ("normal" | "thin" | "extralight" | "light" |
```
vadz, line 476: This is very minor, but please avoid including whitespace-only changes in the commits with other, actually significant changes (this probably means configuring your editor to avoid stripping whitespace automatically from wx files).

src/xrc/xh_bttn.cpp
```diff
@@ -29,9 +29,31 @@ wxButtonXmlHandler::wxButtonXmlHandler()
...
+    XRC_ADD_STYLE(wxBU_NOTEXT);
     AddWindowStyles();
 }
 
+// Function calls the given setter with the contents of the node with the given
+// name, if present.
+//
+// If alternative parameter name is specified, it is used too.
+void
+wxButtonXmlHandler::SetBitmapIfSpecified(wxButton* button,
```
vadz, line 41: Could we avoid duplicating this function between here and `wxBitmapButtonXmlHandler`? It's not much code, but it still would be better to have it in a single place and it looks like it ought not be too difficult to just make it take `wxAnyButton`.

src/xrc/xh_bttn.cpp
```diff
@@ -55,6 +77,17 @@ wxObject *wxButtonXmlHandler::DoCreateResource()
...
+    SetBitmapIfSpecified(button, &wxBitmapButton::SetBitmapPressed,
+                         "pressed", "selected");
+    SetBitmapIfSpecified(button, &wxBitmapButton::SetBitmapFocus, "focus");
+    SetBitmapIfSpecified(button, &wxBitmapButton::SetBitmapDisabled, "disabled");
+    SetBitmapIfSpecified(button, &wxBitmapButton::SetBitmapCurrent,
+                         "current", "hover");
+
+    auto margins = GetSize("margins");
+    if (margins != wxDefaultSize)
+        button->SetBitmapMargins(margins.x, margins.y);
```
vadz, line 89: Why not directly use the overload taking `wxSize`?
```suggestion
        button->SetBitmapMargins(margins);
```

src/xrc/xh_bttn.cpp
```diff
@@ -55,6 +77,17 @@ wxObject *wxButtonXmlHandler::DoCreateResource()
 
     SetupWindow(button);
 
+    SetBitmapIfSpecified(button, &wxBitmapButton::SetBitmapPressed,
```
vadz, line 80: I think this merits a comment saying that we accept the old names here for kind of backwards cross-compatibility with `wxBitmapButton`.

vadz, line 80: This looks like too a copy-pasto (and explains at least some of the errors in the CI builds):
```suggestion
    SetBitmapIfSpecified(button, &wxButton::SetBitmapPressed,
```
(same thing below too, of course).

vadz (changes requested):
Thanks, this looks good, but it would be even better if we could avoid duplicating the code between here and `wxBitmapButtonXmlHandler`, it would be great if you could please try to do it.

But it could be merged as is too, however the CI build problems need to be fixed, of course.

src/xrc/xh_bttn.cpp
Randalphwa: Actually, if we don't support the older alternative names for wxButton, then the function could be eliminated entirely -- which would also eliminate the call to GetBitmap() for the older parameter name (the newer parameters call GetBitmapBundle()). So instead, the code would be:

```c++
    auto node = GetParamNode("pressed");
    if (node)
        button->SetBitmapPressed(GetBitmapBundle(node));
    node = GetParamNode("focus");
    if (node)
        button->SetBitmapFocus(GetBitmapBundle(node));
    node = GetParamNode("disabled");
    if (node)
        button->SetBitmapDisabled(GetBitmapBundle(node));
    node = GetParamNode("current");
    if (node)
        button->SetBitmapCurrent(GetBitmapBundle(node));

```

The other approach would be to make the function a static member of the wxBitmapButtonXmlHandler class, and call it directly. The one downside is that those alternative names (`selected` and `hover`) don't support bundles -- so calling the older wxBitmapButtonXmlHandler method could cause problems for people expecting it to support bundles.

docs/doxygen/overviews/xrc_format.h
```diff
@@ -907,6 +907,21 @@ pseudo-class. @c ownerdrawnitem objects have the following properties:
...
+     Bitmap to show when the button has focus (default: none, same as @c bitmap).}
+@row3col{disabled, @ref overview_xrcformat_type_bitmap,
+     Bitmap to show when the button is disabled (default: none, same as @c bitmap).}
+@row3col{current, @ref overview_xrcformat_type_bitmap,
+     Bitmap to show when the mouse cursor hovers above the bitmap (default: none, same as @c bitmap).
+     This property exists since wxWidgets 3.1.6, but the equivalent (and still
+     supported) "hover" property can be used in the older versions.}
+@row3col{margins, @ref overview_xrcformat_type_size,
+    Set the margins between the bitmap and the text of the button.
+    This method is currently only implemented under MSW. If it is not called, a default margin is used around the bitmap.}
```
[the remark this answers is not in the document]
Randalphwa: In looking through this file, I see that there are a variety of ways the version number has been specified ranging from the wordy `This property is only available since wxWidgets 3.1.1` to the succinct `@since 3.1.1`. In looking at the HTML file that doxygen produces, the text generated for `@since 3.1.1` takes up the most space, but is _very_ easy to spot -- a reader is unlikely to miss it.

My inclination is to go with `@since 3.1.7` for the new wxButton properties, and then if you agree, to do another PR after this one that replaces all the variants of how the version gets specified with a single consistent method, namely `@since n.n.n`.

misc/schema/xrc_schema.rnc
Randalphwa: Ack, didn't mean to do that. I switched my editor to both show me when whitespace changed, and specific to this repository to not trim trailing whitespace. Shouldn't happen again...

src/xrc/xh_bttn.cpp
Randalphwa: Just saw the comment about not using `auto` -- will change accordingly.

docs/doxygen/overviews/xrc_format.h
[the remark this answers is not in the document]
vadz: I thought `@since` only worked at paragraph level and not inside the tables like this one. If it does work (perhaps because we're using a slightly less ancient Doxygen version now), then it's definitely better to use it too and another PR changing this would be welcome, TIA!

src/xrc/xh_bttn.cpp
vadz: I'm fine with this, i.e. I think that while the rationale about supporting the deprecated synonyms for the existing XRC using `wxBitmapButton` changed to use `wxButton` makes sense, we don't really _have_ to do it like this and, thinking more about it, it might actually be better not to do it to avoid encouraging keeping using the old names.

So, to summarize, yes, I'd be perfectly fine with just this, thank you.

PR #22453, opened 2022-05-24
wxDropTarget fix [gtk port]: GdkDragContext must be set in target_dra…

valid-ptr:
…g_data_received, thus wxDataFormat will be correct when OnData callback is called

### DIFF
--- a/src/gtk/dnd.cpp
+++ b/src/gtk/dnd.cpp
@@ -373,8 +373,9 @@ static void target_drag_data_received( GtkWidget *WXUNUSED(widget),
 
     wxLogTrace(TRACE_DND, wxT( "Drop target: data received event") );
 
-    /* inform the wxDropTarget about the current GtkSelectionData.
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, I don't see anything wrong with it, but it's not totally clear what problem exactly does this fix, could you please explain? Looking at `wxDropTarget::GetData()`, it doesn't seem to be using `m_dragContext`, so why does setting it matter?

valid-ptr:
This comes from DVC's `wxEVT_DATAVIEW_ITEM_DROP` event.
Event's `GetDataFormat() `returns bad values (`fmt->GetType() `and `fmt->GetId() `as instance) without this PR

NOTE: D'n'D from different processes

vadz:
I still don't understand where is the drag context used exactly (do you?), but let's merge this if it fixes a real problem, thanks.

valid-ptr:
correct :)

Issue #22452, opened 2022-05-24
wxVersionInfo is missing a fourth part


rcane:
Currently `wxVersionInfo` saves a major, minor and micro part of a version. But there are cases when a complete version consists of four parts. I just realized that when I upgraded the WebView2 component from 101.0.1210.32 to 101.0.1210.53 and saw the result of `wxWebView::GetBackendVersionInfo()` not containing the last number. Not having the last part of the version makes it impossible to tell what actual version of the WebView2 component is currently in use.

Adding a fourth part (cmake calls it "tweak", others call it "build") to `wxVersionInfo` should be easy except for the change of the constructor which would break backward compatibility.

vadz:
Indeed, thanks.

PR #22455, opened 2022-05-25
Add fourth part to wxVersionInfo

rcane:
This implements the requested feature from #22452.

In addition to adding the fourth part to wxVersionInfo I also extended the Edge and IE webview backends to include the the new part into their versions.


interface/wx/versioninfo.h
```diff
@@ -74,6 +76,13 @@ class wxVersionInfo
     */
     int GetMicro() const;
 
+    /**
+        Get the tweak version, or build number.
+
+        @return Tweak version, or build number.
+    */
```
vadz, line 83:
```suggestion
        @since 3.1.7
    */
```

interface/wx/versioninfo.h
```diff
@@ -35,6 +35,7 @@ class wxVersionInfo
         @param major The major version component.
         @param minor The minor version component.
         @param micro The micro version component, 0 by default.
+        @param tweak The tweak version component, 0 by default.
```
vadz, line 38:
```suggestion
        @param tweak The tweak version component, 0 by default (this parameter is available since wxWidgets 3.1.7 only).
```

vadz (commented):
Looks good, thanks, just a couple of doc tweaks (sorry...).

BTW, I've tried to search for "tweak version" (and various variants thereof) but couldn't find any mentions of it. I've never seen this word used in this context myself and [Wikipedia](https://en.wikipedia.org/wiki/Software_versioning) doesn't mention it neither, so I wonder just how well-known it is. IMHO "build" is a much more common term, so maybe we should use it instead?

What do the others think, is anybody here familiar with the term "tweak" in the version context?

include/wx/versioninfo.h
```diff
@@ -23,15 +23,26 @@ class wxVersionInfo
...
+        , m_tweak(tweak)
+    {
+    }
+
+    // This constructor exists for backward compatibility (before the tweak
+    // part was added).
+    wxVersionInfo(const wxString& name, int major, int minor, int micro,
+                  const wxString& description,
+                  const wxString& copyright = wxString())
+        : wxVersionInfo(name, major, minor, micro, 0, description, copyright)
```
vadz, line 44: Oh, damn, sorry, I missed this during the review, but we still support C++98 and so can't use delegating constructors, so you'll have to duplicate the above lines here or add `Init()` function, as done in other places.

rcane:
As for the word `tweak`, I got that from [cmake](https://cmake.org/cmake/help/latest/variable/CMAKE_VERSION.html). 
`wxVersionInfo` calls the third part `micro` while cmake calls it `patch`. So there is obviously no consensus about that either. But I have no problem with renaming `tweak` to `build`. Or we could call it `nano` so that everybody is confused. :-)

vadz:
If you allow me to be not very nice just for a second, CMake is not exactly where I'd look for getting clear and unambiguous terminology from.

FWIW [Microsoft uses](https://docs.microsoft.com/en-us/dotnet/api/system.version) `major.minor[.build[.revision]]` and considering that we currently only use this with Microsoft products, it would, IMO, make sense to use the names compatible with their schema, i.e. either "build" or "revision".

rcane:
I was not suggesting that CMake should set any kind of standard here. We just use it as our build system and therefor I was familiar with `tweak` as the name for the fourth part of the version.
But lets just use the Microsoft scheme and call it `revision`.

vadz:
Thanks, merged now with a couple of cosmetic tweaks.

PR #22456, opened 2022-05-25
Use @since for wxWidgets version in XRC doc

Randalphwa:
Doxygen 1.8.17 that is being used to build the docs supports placing `@since` in tables. This PR changes the XRC file format overview to use this macro in order for version number to be both consistent and easy to spot.


vadz:
Thanks!

Just to be sure, are there any versions (>= 1.8.17) with which this doesn't work or does this part work correctly with all of them?

PR #22457, opened 2022-05-26
Grid: drag move improvements

DietmarSchwertberger:
Some improvements:
 - better visualization of the 'pressed' label of the row/col that will be dragged (see image below)
 - this is done by `DrawLineLabel` omitting the border
 - the 'pressed' state will also be maintained when re-painted, e.g. due to scrolling
 - correctly visualize with grey insertion markers if no move would happen (helper method `CheckIfAtDragSourceLine` checking the coordinate whether it's max. half a line away from the source line)
 - don't call `DoEndMove` in this case
 - correctly update marker color w. hidden cols/rows; e.g. row 2 is hidden and row 3 dragged to row 1 -> change from grey to blue
 - re-paint label window if drag move is canceled by lost mouse capture

I don't like the variable reset code at the end of the `event.LeftUp()` handler around line 4288 as this has grown a lot.
I'm not 100% sure, but I think this block, including the call to `ChangeCursorMode` could be replaced by a call to `EndDraggingIfNecessary` and the reset code for `m_dragMoveRowOrCol, m_dragLastPos,  m_dragLastColour` could then be added to `DoAfterDraggingEnd`.

grafik

### DIFF
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -2804,8 +2804,11 @@ protected:
     // operation in progress.
     int     m_dragMoveRowOrCol;
 
-    // Last horizontal mouse position while drag-moving a column.
+    // Last drag marker position while drag-moving a column.
     int     m_dragLastPos;
+
+    // Last drag marker colour while drag-moving a column.
+    const wxColour *m_dragLastColour;
 
     // Row or column (depending on m_cursorMode value) currently being resized
     // or -1 if there is no resize operation in progress.
@@ -2987,6 +2990,10 @@ private:
     // helper for Process...MouseEvent to scroll
     void CheckDoDragScroll(wxGridSubwindow *eventGridWindow, wxGridSubwindow *gridWindow,
                            wxPoint posEvent, int direction);
+
+    // helper for Process...LabelMouseEvent to check whether a drag operation
+    // would end at the source line, i.e. have no effect
+    bool CheckIfAtDragSourceLine(const wxGridOperations &oper, int coord);
 
     // return true if the grid should be refreshed right now
     bool ShouldRefresh() const
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -2804,10 +2804,10 @@ protected:
     // operation in progress.
     int     m_dragMoveRowOrCol;
 
-    // Last drag marker position while drag-moving a column.
+    // Last drag marker position while drag-moving a row or column.
     int     m_dragLastPos;
 
-    // Last drag marker colour while drag-moving a column.
+    // Last drag marker colour while drag-moving a row or column.
     const wxColour *m_dragLastColour;
 
     // Row or column (depending on m_cursorMode value) currently being resized
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -4283,13 +4283,7 @@ void wxGrid::ProcessRowColLabelMouseEvent( const wxGridOperations &oper, wxMouse
         {
             DoColHeaderClick(line);
         }
-
-        ChangeCursorMode(WXGRID_CURSOR_SELECT_CELL, labelWin);
-        m_dragMoveRowOrCol = -1;
-        m_dragLastPos = -1;
-        m_dragLastColour = NULL;
-        m_lastMousePos = wxDefaultPosition;
-        m_isDragging = false;
+        EndDraggingIfNecessary();
     }
 
     // ------------ Right button down
@@ -4535,6 +4529,10 @@ void wxGrid::DoAfterDraggingEnd()
     m_isDragging = false;
     m_startDragPos = wxDefaultPosition;
     m_lastMousePos = wxDefaultPosition;
+    // from drag moving row/col
+    m_dragMoveRowOrCol = -1;
+    m_dragLastPos = -1;
+    m_dragLastColour = NULL;
 
     m_cursorMode = WXGRID_CURSOR_SELECT_CELL;
     m_winCapture->SetCursor( *wxSTANDARD_CURSOR );
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -4515,6 +4515,9 @@ void wxGrid::CancelMouseCapture()
     // cancel operation currently in progress, whatever it is
     if ( m_winCapture )
     {
+        if ( m_cursorMode == WXGRID_CURSOR_MOVE_COL ||
+             m_cursorMode == WXGRID_CURSOR_MOVE_ROW )
+            m_winCapture->Refresh();
         DoAfterDraggingEnd();
     }
 }
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -4159,18 +4159,13 @@ void wxGrid::ProcessRowColLabelMouseEvent( const wxGridOperations &oper, wxMouse
             {
                 if ( oper.CanDragMove(this) )
                 {
+                    ChangeCursorMode(oper.GetCursorModeMove(), headerWin);
+
                     // Show button as pressed
+                    m_dragMoveRowOrCol = line;
                     wxClientDC dc( headerWin );
                     oper.PrepareDCForLabels(this, dc);
-                    dc.SetPen( wxPen( headerWin->GetBackgroundColour(), 1 ) );
-
-                    int lineStart = oper.GetLineStartPos(this, line);
-                    int lineEnd = oper.GetLineEndPos(this, line) - 1;
-                    int lineLength = oper.Select(headerWin->GetClientSize()) - 1;
-                    oper.DrawParallelLine(dc, 0, lineLength, lineStart);
-                    dual.DrawParallelLine(dc, lineStart, lineEnd, 1);
-
-                    ChangeCursorMode(oper.GetCursorModeMove(), headerWin);
+                    oper.DrawLineLabel(this, dc, line);
                 }
                 else
                 {
@@ -4290,6 +4285,7 @@ void wxGrid::ProcessRowColLabelMouseEvent( const wxGridOperations &oper, wxMouse
         }
 
         ChangeCursorMode(WXGRID_CURSOR_SELECT_CELL, labelWin);
+        m_dragMoveRowOrCol = -1;
         m_dragLastPos = -1;
         m_dragLastColour = NULL;
         m_lastMousePos = wxDefaultPosition;
@@ -7100,7 +7096,21 @@ void wxGrid::DrawRowLabel( wxDC& dc, int row )
                                 (gs_defaultHeaderRenderers.rowRenderer);
 
     wxRect rect(0, GetRowTop(row), m_rowLabelWidth, GetRowHeight(row));
-    rend.DrawBorder(*this, dc, rect);
+
+    if ( m_cursorMode == WXGRID_CURSOR_MOVE_ROW )
+    {
+        // clear the background:
+        // when called from ProcessRowColLabelMouseEvent the background is not
+        // cleared at this point
+        dc.SetPen(*wxTRANSPARENT_PEN);
+        dc.SetBrush(m_colLabelWin->GetBackgroundColour());
+        dc.DrawRectangle(rect);
+    }
+
+    // draw a border if the row is not being drag-moved
+    // (in that case it's omitted to have a 'pressed' appearance)
+    if (m_cursorMode != WXGRID_CURSOR_MOVE_ROW || row != m_dragMoveRowOrCol)
+        rend.DrawBorder(*this, dc, rect);
 
     int hAlign, vAlign;
     GetRowLabelAlignment(&hAlign, &vAlign);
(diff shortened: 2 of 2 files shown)


vadz:
Thanks, I didn't test it yet, but this looks reasonable, so I'll merge it if I don't see anything wrong with it when testing.

Concerning resetting the variables, using `wxON_BLOCK_EXIT_SET` might help a bit, although it would, of course, be even better to reuse the existing functions if possible.

vadz:
After testing, this does look much nicer, thank you! But there is a small problem: if you drag a row/column upwards/to the left and then drag it back to its original position, it correctly shows the divider as grey, but if you then drag it further down/to the right, the line between the original row/column and its next neighbour is not drawn at all any more. Could you please fix this, as this wasn't the case before and I'd like to avoid introducing regressions? TIA!

What was already the case before, and still is, is that none of this works with Wayland, as we can't use `wxClientDC` for drawing with it at all. But this is a different (even if, IMO, a more critical) issue, of course, so it's not an obstacle to merging this PR.

DietmarSchwertberger:
Thanks for reviewing.

I don't see a redraw issue. While the drag is ongoing, the original row/column is always drawn as 'pressed', i.e. without a border inside the label. The borders of the surrounding labels should be drawn, though. The one below/right is of course only very light.
Does it look different from the image above?

Regarding Wayland:
I was wondering already whether this all does work on Wayland, but I'm always only using it on Windows and have never built wx myself on Linux.
With this PR it's already easier to move things to the Paint handler: the 'pressed' state is drawn there already, so probably just the marker would have to be added and the `wxClientDC` code replaced with a suitable `Refresh`.
I could give it a try with this PR already.
What is the policy on this? Still use `wxClientDC` where it is working or use Refresh/Paint on all platforms?

vadz:
I don't know what could be the difference, I'm just running the sample on a pretty standard Debian Bookworm system with `GDK_BACKEND=x11`. And, in fact, you don't need to do anything complicated to see the problem: just pressing the mouse button over a column or row and not moving it at all makes the right/bottom border disappear. And while moving it to the left/up restores it, moving it to the right/down does not, resulting in this instead: 

grid_coldrag

(it's the column "C" being dragged here). It could be a Linux/wxGTK-only issue, but it's strange that it didn't appear before this change. Can you test the sample under Linux or not at all?

Re Wayland: `wxClientDC` is never going to work there (nor under Mac for that matter), so it would, of course, be best to avoid using it. But as we already use it anyhow, we could still merge this PR (once the problem above is fixed) and improve it later, i.e. using it is not a regression. But we definitely should avoid using it in any new code.

DietmarSchwertberger:
But that's the last variant as from
https://github.com/wxWidgets/wxWidgets/pull/22403#issuecomment-1120220008
https://github.com/wxWidgets/wxWidgets/pull/22403#pullrequestreview-965340804

I will look into a version without `wxClientDC`.
The alternative also could be platform dependent. E.g. https://github.com/matplotlib/matplotlib/pull/11944 seems to be failing on Arch and Fedora.

DietmarSchwertberger:
Sorry, I should have zoomed in.
It seems to me that there is an offset. Is it there also with a normal re-paint? I.e. start dragging, move outside the visible are to trigger a scroll, move back and trigger a scroll in the other direction?

DietmarSchwertberger:
Row label on Windows 10:
grafik

From your screenshot:
grafik
The light grey line at the top is offset to the right compared to the Windows screenshot.


Windows:
grafik

Your screenshot:
grafik

DietmarSchwertberger:
The `DrawBorder` methods have side effects to the `rect` argument.
Do these modifications to `wxGrid::DrawRowLabel` make a difference?
```
@@ -7114,6 +7136,13 @@ void wxGrid::DrawRowLabel( wxDC& dc, int row )
     // (in that case it's omitted to have a 'pressed' appearance)
     if (m_cursorMode != WXGRID_CURSOR_MOVE_ROW || row != m_dragMoveRowOrCol)
         rend.DrawBorder(*this, dc, rect);
+    else
+    {
+        if ( grid.GetBorder() == wxBORDER_NONE )
+            rect.Deflate(2);
+        else
+            rect.Deflate(1);
+    }

     int hAlign, vAlign;
     GetRowLabelAlignment(&hAlign, &vAlign);
@@ -7263,6 +7292,13 @@ void wxGrid::DrawColLabel(wxDC& dc, int col)
         // (in that case it's omitted to have a 'pressed' appearance)
         if (m_cursorMode != WXGRID_CURSOR_MOVE_COL || col != m_dragMoveRowOrCol)
             rend.DrawBorder(*this, dc, rect);
+        else
+        {
+            if ( grid.GetBorder() == wxBORDER_NONE )
+                rect.Deflate(2);
+            else
+                rect.Deflate(1);
+        }
     }

     int hAlign, vAlign;
```

DietmarSchwertberger:
This is Ubuntu X11 vs. Windows with coloured borders (drawing order is red, blue, green):
grafik grafik

vadz:
> The `DrawBorder` methods have side effects to the `rect` argument.

Yes, I think this was done in order to allow centering things correctly inside the bordered region.

> Do these modifications to `wxGrid::DrawRowLabel` / `DrawColLabel` make a difference?

No, sorry, this doesn't change anything visually. But then I don't understand how was it supposed to: the border is still not drawn, is it? And this is exactly the problem I'm seeing: when pressing the mouse (so the cursor mode is `WXGRID_CURSOR_MOVE_COL`) without moving it (so the current column is still `m_dragMoveRowOrCol`), `rend.DrawBorder()` is not called and hence there is no border between this cell and the next one. I'm still not sure why this doesn't happen under MSW, but it is definitely wrong not to not draw any border in this case. Where is something supposed to be drawn when the condition before the `DrawBorder()` call is false?

And yes, there is also the problem with one pixel gap that I didn't even notice. But this one was already present before, i.e. it's visible with the current master even without this PR.

DietmarSchwertberger:
Well, the one pixel offsets are not even the problem. It seems that the left border is just to light.
This is with the colored version where I see the green border:
grafik

DietmarSchwertberger:
On a close look at the previous images, you can see that on Ubuntu the header and it's left border are almost white while on Windows you can still differentiate the shades:
grafik
grafik

DietmarSchwertberger:
A workaround could be to paint a single pixel line at the right edge for the pressed button. Is there a system colour that could be used?

vadz:
How do you build the coloured version, do you have a patch I could apply to test it here please?

In any case, I still think there is more to the problem than just using wrong colours: this could well explain the weird initial appearance, i.e. that the border seems to disappear when the mouse is depressed. But it doesn't explain why it's not restored after moving the cursor to the next column, which becomes highlighted with the blue separator line -- surely the appearance of the original column should be restored to the initial one by then?

DietmarSchwertberger:
I see it the other way round: when pressing the mouse, I know where I am, so I don't need the visualisation. While the drag is ongoing, I want to see which one is the dragged row/col. Only when the drag is finished, the original, unpressed state should be restored again.

I'm using the code below for the coloured version:
```
@@ -336,7 +338,8 @@ void wxGridRowHeaderRendererDefault::DrawBorder(const wxGrid& grid,
                                                 wxDC& dc,
                                                 wxRect& rect) const
 {
-    dc.SetPen(wxPen(wxSystemSettings::GetColour(wxSYS_COLOUR_3DSHADOW)));
+    //dc.SetPen(wxPen(wxSystemSettings::GetColour(wxSYS_COLOUR_3DSHADOW)));
+    dc.SetPen(*wxRED_PEN);
     dc.DrawLine(rect.GetRight(), rect.GetTop(),
                 rect.GetRight(), rect.GetBottom());

@@ -349,13 +352,15 @@ void wxGridRowHeaderRendererDefault::DrawBorder(const wxGrid& grid,
     int ofs = 0;
     if ( grid.GetBorder() == wxBORDER_NONE )
     {
+        dc.SetPen(*wxBLUE_PEN);
         dc.DrawLine(rect.GetLeft(), rect.GetTop(),
                     rect.GetLeft(), rect.GetBottom());

         ofs = 1;
     }

-    dc.SetPen(wxPen(wxSystemSettings::GetColour(wxSYS_COLOUR_3DLIGHT)));
+    //dc.SetPen(wxPen(wxSystemSettings::GetColour(wxSYS_COLOUR_3DLIGHT)));
+    dc.SetPen(*wxGREEN_PEN);
     dc.DrawLine(rect.GetLeft() + ofs, rect.GetTop(),
                 rect.GetLeft() + ofs, rect.GetBottom());
     dc.DrawLine(rect.GetLeft() + ofs, rect.GetTop(),
@@ -368,7 +373,8 @@ void wxGridColumnHeaderRendererDefault::DrawBorder(const wxGrid& grid,
                                                    wxDC& dc,
                                                    wxRect& rect) const
 {
-    dc.SetPen(wxPen(wxSystemSettings::GetColour(wxSYS_COLOUR_3DSHADOW)));
+    //dc.SetPen(wxPen(wxSystemSettings::GetColour(wxSYS_COLOUR_3DSHADOW)));
+    dc.SetPen(*wxRED_PEN);
     dc.DrawLine(rect.GetRight(), rect.GetTop(),
                 rect.GetRight(), rect.GetBottom());
     dc.DrawLine(rect.GetLeft(), rect.GetBottom(),
@@ -378,13 +384,15 @@ void wxGridColumnHeaderRendererDefault::DrawBorder(const wxGrid& grid,
     int ofs = 0;
     if ( grid.GetBorder() == wxBORDER_NONE )
     {
+        dc.SetPen(*wxBLUE_PEN);
         dc.DrawLine(rect.GetLeft(), rect.GetTop(),
                     rect.GetRight(), rect.GetTop());

         ofs = 1;
     }

-    dc.SetPen(wxPen(wxSystemSettings::GetColour(wxSYS_COLOUR_3DLIGHT)));
+    //dc.SetPen(wxPen(wxSystemSettings::GetColour(wxSYS_COLOUR_3DLIGHT)));
+    dc.SetPen(*wxGREEN_PEN);
     dc.DrawLine(rect.GetLeft(), rect.GetTop() + ofs,
                 rect.GetLeft(), rect.GetBottom());
     dc.DrawLine(rect.GetLeft(), rect.GetTop() + ofs,
```

DietmarSchwertberger:
I'm wondering why on Ubuntu the pens seem to have some transparency. Could this also contribute in making the line lighter than intended?
(e.g. the green line has a red value of 32 where it did overdraw the red line)

vadz:
> I see it the other way round: when pressing the mouse, I know where I am, so I don't need the visualisation.

I am not sure if I agree, but the main point is that not having any separator between 2 different cells definitely looks like a bug, I don't believe anybody seeing it would think this was the intended behaviour. I.e. we may draw the same default border as when not dragging, or some different border (IMO a grey one would be fine), but we must draw something.

Also, just to really prove the point, the behaviour differs between left and right neighbours and there can really be no good reason to intentionally do something like this.

DietmarSchwertberger:
The behaviour between left and right differs as the 'lighting' is from top left. The contrast between `wxSYS_COLOUR_3DLIGHT` and the background colour is too low for any visible 3D effect.
I will try with a darker line in `wxSYS_COLOUR_3DSHADOW`.
Alternatively, e.g. on the Ubuntu standard theme an orange frame in `wxSYS_COLOUR_HIGHLIGHT` could be used.
On Windows, this would be a dark blue, but actually I would prefer the current behaviour.

DietmarSchwertberger:
With `wxSYS_COLOUR_HIGHLIGHT `:
Ubuntu:
grafik

Windows:
grafik

The code in `DrawColLabel` is a bit clumsy. Probably it should be moved to `DrawBorder`.
```
        if (m_cursorMode != WXGRID_CURSOR_MOVE_COL || col != m_dragMoveRowOrCol)
            rend.DrawBorder(*this, dc, rect);
        else
        {
            dc.SetPen(wxPen(wxSystemSettings::GetColour(wxSYS_COLOUR_HIGHLIGHT)));
            dc.DrawRectangle(rect);
            if ( GetBorder() == wxBORDER_NONE )
                rect.Deflate(2);
            else
                rect.Deflate(1);
```

vadz:
Sorry, I just don't have enough bandwidth for this right now, but, without looking at the details, I think that just about anything showing the border between the cells clearly should be acceptable.

DietmarSchwertberger:
No problem. I will push a modification to use `wxSYS_COLOUR_HIGHLIGHT` on GTK. Then I will start working on Wayland support. This will take some time and you can decide then which border variant to use.

vadz:
Just to be sure: did you already make all the planned changes, or are any more still upcoming? I.e. should I already retest this PR and merge it if no problems are found or not yet? TIA!

vadz:
Thanks, but let's leave this for another PR, as I think it will make a lot more changes to the code. If this is the final versions for X11, let's already merge it (provided there are no other problems). Please let me know if you disagree.

DietmarSchwertberger:
For me, merging would be fine.

On gtk a border is drawn. On Windows the 'pressed' state is drawn without border, as above in the first post.
For reference: 

grafik
grafik


I can't test on macOS, as my current VM is too old already.

DietmarSchwertberger:
From a technical point of view, the highlighted border would also work on MSW.
IMHO the resulting user interface of blue border plus grey and blue insertion markers is very ugly, though.
On Ubuntu the combination with the orange border is working better, but actually it would be better to have a colour scheme where the label background colour and `wxSYS_COLOUR_3DLIGHT` are not 'almost white'.

vadz:
The colour scheme can be more or less arbitrary (the only thing we can rely on is having contrasting text and background colours), which is exactly why I dislike doing this. Even under Linux different distributions use very different themes by default, without even speaking about the users selecting their own non-default themes.

vadz:
After testing removing `#ifdef __WXGTK__`, I think it looks not bad under MSW too and so I think we should just use this under all platforms. If you don't mind, I'm going to merge it like this -- but please let me know if you do.

vadz:
Sorry for the short notice, but I'm finally going to merge this to avoid doing it at the last possible moment later or postponing it until after 3.1.7. If you mind my last commit terribly, please let me know and we'll try to find some other solution, but for now this looks like an improvement that it would be nice to have in 3.1.7.

DietmarSchwertberger:
Thanks for mergin. It's fine for me. I was just offline a few days. The PR for Wayland support could take a few weeks.

PR #22458, opened 2022-05-26
AUI Floating Pane position fix

valid-ptr:
Sometimes pane's floating position is out of sync with it's owner (frame).
This PR solves this issue.

Example of the issue:
When you open some floating pane for the first time and i'ts floating_position is wxDefaultPosition, it's frame position will be in the center of managed window. After you close if and open it again, it's position will be in (0, 0) because OnMove event has been ignored.
NOTE: (0, 0) comes from OnSize event.

### DIFF
--- a/src/aui/floatpane.cpp
+++ b/src/aui/floatpane.cpp
@@ -203,6 +203,10 @@ void wxAuiFloatingFrame::OnClose(wxCloseEvent& evt)
 
 void wxAuiFloatingFrame::OnMoveEvent(wxMoveEvent& event)
 {
+    // Always sync pane's floating_pos with frame's position
+    if (m_ownerMgr)
+        m_ownerMgr->GetPane(m_paneWindow).floating_pos = event.GetPosition();
+
     if (!m_solidDrag)
     {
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, makes sense to me. If there are no comments from people more familiar with wxAUI, I'll merge it soon.

vadz:
After testing, this does look much nicer, thank you! But there is a small problem: if you drag a row/column upwards/to the left and then drag it back to its original position, it correctly shows the divider as grey, but if you then drag it further down/to the right, the line between the original row/column and its next neighbour is not drawn at all any more. Could you please fix this, as this wasn't the case before and I'd like to avoid introducing regressions? TIA!

What was already the case before, and still is, is that none of this works with Wayland, as we can't use `wxClientDC` for drawing with it at all. But this is a different (even if, IMO, a more critical) issue, of course, so it's not an obstacle to merging this PR.

vadz:
I'm going to merge this now and sorry for a wrong comment (well, correct comment, but on a wrong PR) if you've seen it before I deleted it.

PR #22460, opened 2022-05-26
BUG FIX - Comparison with wrong type string leads to a crash in wxDataViewBitmapRenderer::MacRender

hwiesmann:
A bitmap bundle type has to be compared with "wxBitmapBundle" and not "wxBitmap".


vadz:
Sorry, why did this get closed? Was it intentional? The change looks correct to me, so I'd like to merge it, but I'd like to check that I'm not missing something here.

hwiesmann:
I deleted the equivalent local path but accidentally also deleted the branch on GitHub. The fix is correct and should be applied.

PR #22461, opened 2022-05-27
Fix incorrect coding of "\"

Randalphwa:
This commit fixes a formatting problem with these two files
when built with Doxygen version 1.9.3.

The correct encoding is to escape the quotes and the backslash.
While the older style encoding used to work, it is no longer
supported due to a fix with how a single quote gets handled
by Doxygen. The escaped version works with Doxygen 1.8.17
so this will work fine with the current build version as well as
in more recent versions (tested with 1.9.3).


vadz:
Thanks, I don't understand anything in Doxygen quoting rules, but I'm glad that there is someone who does :-)

PR #22462, opened 2022-05-27
Fix MSVC compiler warnings

Randalphwa:
This adds a `static_cast<>` to fix warnings from the MSVC compiler (Version 19.29.30136 for x64) when compiling a 64-bit app.

`warning C4244: '=': conversion from 'int' to 'wxTextAttrDimensionFlags', possible loss of data`

`warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data`

I created a test header file that #includes roughly 700 header files from wxWidgets, and these were the only ones that generated warnings.

### DIFF
--- a/include/wx/choicebk.h
+++ b/include/wx/choicebk.h
@@ -86,7 +86,7 @@ protected:
 
     void UpdateSelectedPage(size_t newsel) wxOVERRIDE
     {
-        GetChoiceCtrl()->Select(newsel);
+        GetChoiceCtrl()->Select(static_cast<int>(newsel));
     }
 
     wxBookCtrlEvent* CreatePageChangingEvent() const wxOVERRIDE;
--- a/include/wx/htmllbox.h
+++ b/include/wx/htmllbox.h
@@ -265,7 +265,7 @@ public:
     // -----------------
 
     virtual unsigned int GetCount() const wxOVERRIDE
-        { return m_items.GetCount(); }
+        { return static_cast<unsigned int>(m_items.GetCount()); }
 
     virtual wxString GetString(unsigned int n) const wxOVERRIDE;
 
--- a/include/wx/richtext/richtextbuffer.h
+++ b/include/wx/richtext/richtextbuffer.h
@@ -368,7 +368,8 @@ public:
     /**
         Constructor taking value and units flag.
     */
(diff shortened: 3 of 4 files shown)


vadz:
Thanks, this should probably be merged, but I'm puzzled by something: both of these warnings seem to be disabled in `wx/defs.h` (whether or not it's a good idea to disable them is another question...), so why didn't this have any effect?

Also, we have `tests/allheaders.cpp` which exists to ensure that we don't have any warnings in our headers when using gcc/clang, but we're not building it in the GitHub CI workflows at all and while we do build it on Appveyor, it doesn't work too well and results in [many warnings](https://ci.appveyor.com/project/wxWidgets/wxwidgets/build/job/arcoyunesjgv93kk?fullLog=true#L2460) because we need to disable PCH support for this one. We really need to fix this to ensure there are no warnings in the headers with MSVC.

Randalphwa:
Hmm, hold off on the merge while I temporarily disable those checks in wx/defs.h to see what happens, and to see if I can figure out why these three files were generating warnings anyway. I'll take a look at tests/allheaders.cpp as well just to see if I missed any headers, and maybe see if I can make it effective in MSW builds.

Randalphwa:
I remember why now. I've made it a habit to always push pragma warnings, then include <wx/defs.h> and a few other wxWidgets files, then pop the pragma warnings. That's so that any warnings defs.h shuts off are only for wxWidgets code, and not for my own code. However, that also meant that I had to add these three files within that block because once I popped the pragma warnings, any where I included them had to re-push/pop the warnings around these files.

I commented out C4267 in defs.h and then recompiled with my massive header list, and found one additional file needing a static case.

allheaders.cpp includes "allheaders.h" which is supposed to list all the header files that can be included in user code, but it's missing several directories (xrc, ribbon, propgrid, generic, aui, html) and of course all the msw/ files for MSVC builds. I can open a separate PR and add them from my list if you like, though I'm not setup to run the test scripts to compile allheaders.cpp -- I just include the list in a test project I use.

vadz:
Thanks, I'll merge this, but to be ensure that there are no warnings in the headers with MSVC we really need to test for this in at least one of the CI builds. Maybe the best would be to just add `allheaders` to the MSVS test project, as then we could manually disable PCH support in it and set `/WX` flag.

PR #22464, opened 2022-05-27
Fix crash in Ribbon sample

MaartenBent:
`m_mutable_button_bar` was used on multiple pages. When one of those pages is deleted, the pointer becomes invalid. Use the ButtonBar associated with the event instead.


vadz:
Thanks as always!

Issue #22450, opened 2022-05-23
Performance degradation in 3.1.6 comparing to 3.1.5 when scrolling a big binary file loaded to wxStyledTextCtrl when the HiDPI awareness is set.


wadim-al:
**Describe the bug**
When a huge binary file is loaded to the wxStyledTextCtrl, the contents can be scrolled quickly when compiled against wxWidgets 3.1.5, but since the wxWidgets 3.1.6 if the HiDPI awareness is set in the manifest, the scrolling is considerably slower. OS scaling is set to 125%, monitor resolution is 2560x1440.

**To Reproduce**
1. Compile the attached snippet against 3.1.6 on Windows with high res monitor and set hidpi awareness in the manifest.
2. Run the example
3. Drag the scroll bar
4. See that redraw is visibly slow (under 1 fps)
5. Repeat all steps, but compile against 3.1.5
6. Compare scrolling performance between 3.1.5 and 3.1.6

**Platform and version information**
 - wxWidgets version you use: 3.1.6
 - wxWidgets port you use: wxMSW
 - OS and its version: Windows 10 x64 Pro 21H2 19044.1645
 - Compiler: VC 2019

**Expected vs observed behaviour**
I expected the same or better scrolling performance, but 3.1.6 is considerably slower.

[Video comparison side by side 3.1.5-3.1.6](https://youtu.be/urC5Hdlvn0Q)

**Image for loading**: 
[image](https://user-images.githubusercontent.com/29713984/169875472-1431e1da-f727-4c0a-926a-82882889a407.jpg)

**Min repro source**: [test1.cpp.txt](https://github.com/wxWidgets/wxWidgets/files/8760771/test1.txt)

**Patch or snippet allowing to reproduce the problem**:
<details>
  <summary>Code Example (click to expand)</summary>

```c
#include <wx/wx.h>
#include <wx/stc/stc.h>
#include <wx/file.h>

class MyApp : public wxApp {
public:
    virtual bool OnInit();
};

class MyFrame : public wxFrame {
public:
    MyFrame();
};

wxIMPLEMENT_APP(MyApp);

bool MyApp::OnInit() {
    MyFrame* frame = new MyFrame();
    frame->Show(true);
    return true;
}

MyFrame::MyFrame() : wxFrame(NULL, wxID_ANY, "STC scrolling issue") {
    wxBoxSizer* sizer = new wxBoxSizer(wxVERTICAL);
    this->SetSizer(sizer);
[...]

PR #22465, opened 2022-05-27
Use correct wxBitmap size for STC

MaartenBent:
This fixes rendering artifacts.

On Windows, the bitmap should have the width and height as provided to InitPixMap(), and initialized with the correct scale factor to prevent performance regressions. CreateWithDIPSize() in combination with ToDIP(width, height) cannot be used, because the scaled size will be rounded and can cause the final bitmap to become too large.

On macOS (retina) CreateWithDIPSize has to be used, so the internal bitmap context has the correct size and scale.

Remove obsolete mdc->GetImpl()->SetWindow(), this is not needed anymore because the DPI is now determined from the associated bitmap content scale factor, and not from the wxWindow.

See #22450

### DIFF
--- a/src/osx/core/bitmap.cpp
+++ b/src/osx/core/bitmap.cpp
@@ -76,7 +76,7 @@ public:
     bool HasAlpha() const;
     WXImage GetImage() const;
 
-    void SetScaleFactor(double scale) { m_scaleFactor = scale; }
+    void SetScaleFactor(double scale);
     double GetScaleFactor() const { return m_scaleFactor; }
 
     const void *GetRawAccess() const;
@@ -367,6 +367,18 @@ WXImage wxBitmapRefData::GetImage() const
     }
 
     return m_nsImage;
+}
+
+void wxBitmapRefData::SetScaleFactor( double scale )
+{
+    wxCHECK_RET( IsOk() , wxT("invalid bitmap") ) ;
+
+    if ( m_scaleFactor == scale )
+        return ;
+
+    CGContextScaleCTM( m_hBitmap, 1 / GetScaleFactor(), -1 / GetScaleFactor() );
+    m_scaleFactor = scale;
+    CGContextScaleCTM( m_hBitmap, GetScaleFactor(), -GetScaleFactor() );
 }
 
 void wxBitmapRefData::UseAlpha( bool use )
--- a/src/stc/PlatWX.cpp
+++ b/src/stc/PlatWX.cpp
@@ -293,13 +293,12 @@ void SurfaceImpl::InitPixMap(int width, int height, Surface *surface, WindowID w
     wxMemoryDC* mdc = surface
         ? new wxMemoryDC(static_cast<SurfaceImpl*>(surface)->hdc)
         : new wxMemoryDC();
-    mdc->GetImpl()->SetWindow(GETWIN(winid));
     hdc = mdc;
     hdcOwned = true;
     if (width < 1) width = 1;
     if (height < 1) height = 1;
-    bitmap = new wxBitmap();
-    bitmap->CreateWithDIPSize(width, height,(GETWIN(winid))->GetDPIScaleFactor());
+    bitmap = new wxBitmap(GETWIN(winid)->ToPhys(wxSize(width, height)));
+    bitmap->SetScaleFactor(GETWIN(winid)->GetDPIScaleFactor());
     mdc->SelectObject(*bitmap);
 }
 


vadz:
Thanks, I guess we should apply this because it fixes a real problem, but it looks like we're still missing some API if we need to use preprocessor checks for what looks like a perfectly reasonable operation. Should we add some other `wxBitmap::CreateXXX()` for this and what should "XXX" be if we do?

MaartenBent:
It would have to be a function something like:
```c++
bool wxBitmap::CreateWithScaleFactors(const wxSize& size, double contentScale, double dpiScale, int depth)
{
    if ( !Create(size * contentScale, depth) )
        return false;

    GetBitmapData()->m_scaleFactor = dpiScale;

    return true;
}
```

With the current API, something like this also works:
```c++
    bitmap = new wxBitmap();
    bitmap->CreateWithDIPSize(width, height, GETWIN(winid)->GetContentScaleFactor());
    bitmap->SetScaleFactor(GETWIN(winid)->GetDPIScaleFactor());
```

vadz:
But do we really need to pass the content scale to this function? AFAICS it's always either the same as DPI scale or 1, depending on the platform. So I think it should be enough to pass just the DIP scale, but only multiply the size by it on the platforms where `wxHAS_DPI_INDEPENDENT_PIXELS`. But this makes it very similar to the existing `CreateWithDIPSize()` which, in turn, makes me wonder if the existing function could be just implemented wrongly? What is so different about its use in wxSTC code and in `src/common/dcbufcmn.cpp`, for example? Is `wxBufferedDC` also broken right now?

BTW, on a completely unrelated note, why do we allocate `bitmap` on the heap here? We should be just using a bitmap object and check for `bitmap.IsOk()` instead...

MaartenBent:
`wxBufferedDC` seems to works fine. The problem is that Scintilla expects an actual [8x8 bitmap](https://github.com/wxWidgets/wxWidgets/blob/master/src/stc/scintilla/src/MarginView.cxx#L140). And it is not possible to create this at 175% scaling.

I'll commit another proposal, where the bitmap DIP size can be specified as `wxRealPoint`.

vadz:
Thanks, I think this is definitely better than the first version, but -- and please forgive me for being so picky -- it still feels like something is not quite right if we have to divide by the scale just to ensure that multiplying by it later gives the expected result.

If we're always going to use `wxRound(sz*scale)` why can't we just pass `win->ToPhys(sz)` as the bitmap size (and then set its scale factor)? This would make sense to me: the size we have is in logical pixels (isn't it?) and we need to get the bitmap size, which is always in physical pixels, from it. So why do we really need to add the new function and deal with rounding here?

If anything, we might want to add `wxBitmap::CreateWithPhysicalSizeAndScaleFactor()` (abbreviated to something more reasonable), but if we consider this to be something only rarely needed, using the simple ctor/`Create()` taking physical size followed by `SetScaleFactor()` call should be fine too, AFAICS.

Or am I still missing something here?

MaartenBent:
> If we're always going to use `wxRound(sz*scale)` why can't we just pass `win->ToPhys(sz)` as the bitmap size (and then set its scale factor)? This would make sense to me: the size we have is in logical pixels (isn't it?) and we need to get the bitmap size, which is always in physical pixels, from it. So why do we really need to add the new function and deal with rounding here?

If I understand correctly, you mean:
```c++
    wxSize size(wxMax(1, width), wxMax(1, height));
    bitmap = new wxBitmap(GETWIN(winid)->ToPhys(wxSize(width, height)));
    bitmap->SetScaleFactor(GETWIN(winid)->GetDPIScaleFactor());
```

On Windows this works, but not on macOS. You can't change the scale factor after the bitmap is created. Because `SetScaleFactor()` does not change `CGContextScaleCTM` in the internal `wxBitmapRefData`.
I'll try to fix this.

MaartenBent:
I based `wxBitmapRefData::SetScaleFactor` on `wxBitmapRefData::UseAlpha` (it also calls `CGContextScaleCTM`). Maybe a macOS experts can confirm this is a correct solution or if something simpler is possible.

vadz:
> If I understand correctly, you mean:
> 
> ```c++
>     wxSize size(wxMax(1, width), wxMax(1, height));
>     bitmap = new wxBitmap(GETWIN(winid)->ToPhys(wxSize(width, height)));
>     bitmap->SetScaleFactor(GETWIN(winid)->GetDPIScaleFactor());
> ```

Yes, exactly!
 
> On Windows this works, but not on macOS. You can't change the scale factor after the bitmap is created.

Ah, this is the part I was missing. We probably should add `wxBitmap::CreateWithPhysSize(wxSize, double)` to make it possible to set it directly when creating the bitmap as even with your changes changing it is not free, as it involves copying all the bitmap data.

> Because `SetScaleFactor()` does not change `CGContextScaleCTM` in the internal `wxBitmapRefData`. I'll try to fix this.

But we still should apply your fix, thank you! I'd just add a note to the documentation saying that calling `SetScaleFactor()` can be expensive under at least macOS for big bitmaps. And, if we're really nitpicking, maybe we should refactor this function and `UseAlpha()` to extract this bitmap recreating code in some common helper?

Thanks again!

MaartenBent:
I found a solution that doesn't require copying the data. Undo the current scale, and then apply the new scale:
```c++
    CGContextScaleCTM( hBitmap, 1 / GetScaleFactor(), -1 / GetScaleFactor() );
    m_scaleFactor = scale;
    CGContextScaleCTM( hBitmap, GetScaleFactor(), -GetScaleFactor() );
```

vadz:
Great, this looks perfect now. I'm going to merge it (probably on Monday) if there are no objections, thanks again!

PR #22467, opened 2022-05-28
Make doc/code changes to wxRichTextCtrlXmlHandler

Randalphwa:
## Doc fixes (xrc_format.h, xrc_schema.rnc)
Fixed the wrong handler name specified for wxRichTextCtrl.

Changed the wording to make it very clear that the handler must be explicitly loaded, and moved the minimum version requirement after the table.

Removed `maxlength` property. SetMaxLength() is for single-line edit controls, and does nothing on multi-line controls. Since wxRichTextCtrl is always a multi-line control, this parameter will not set the maximum length of the text as it indicated.

## Code changes (xh_richtext.cpp)

Added missing `wxRE_` styles for wxRichTextCtrlXmlHandler. While technically wxRE_READONLY has the same value as wxTE_READONLY, a user should be able to specify the wxRE_READONLY style to match the wxRichTextCtrl documentation.

Removed the wxTE_AUTO_URL style. This style is only supported in wxTextCtrl (MSW and GTK), not wxRichTextCtrl. Changing this will generate an assert about an unknown style in a legacy file that used it. Howevewr, the user may otherwise be confused because URLs still look like regular text, and they are not clickable as they would be with this style set in in a wxTextCtrl.

Removed the `maxlength` processing, since it ended up calling a function that simply returned (see description in doc section above about why it was removed from the docs).

### DIFF
--- a/docs/doxygen/overviews/xrc_format.h
+++ b/docs/doxygen/overviews/xrc_format.h
@@ -1935,19 +1935,17 @@ a single wxSizer child with non-ribbon windows in it.
 @hdr3col{property, type, description}
 @row3col{value, @ref overview_xrcformat_type_text,
     Initial value of the control (default: empty).}
-@row3col{maxlength, integer,
-    Maximum length of the text entered (default: unlimited).}
 @endTable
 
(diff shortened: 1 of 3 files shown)


vadz:

PR #22468, opened 2022-05-28
Add undetermined state to wxCheckBoxXmlHandler

Randalphwa:
A 3-state checkbox can be initialized as `checked` or `undetermined` -- this adds support for the `undetermined` initial state, but only if the checkbox style is set to wxCHK_3STATE.

### DIFF
--- a/docs/doxygen/overviews/xrc_format.h
+++ b/docs/doxygen/overviews/xrc_format.h
@@ -937,6 +937,8 @@ No additional properties.
      Label to use for the checkbox (default: empty).}
 @row3col{checked, @ref overview_xrcformat_type_bool,
      Should the checkbox be checked initially (default: 0)?}
+@row3col{undetermined, @ref overview_xrcformat_type_bool,
+     Should a 3-state checkbox be initialized as undetermined (default: 0)?}
 @endTable
 
 
--- a/docs/doxygen/overviews/xrc_format.h
+++ b/docs/doxygen/overviews/xrc_format.h
@@ -946,10 +946,10 @@ No additional properties.
 @hdr3col{property, type, description}
 @row3col{label, @ref overview_xrcformat_type_text,
      Label to use for the checkbox (default: empty).}
-@row3col{checked, @ref overview_xrcformat_type_bool,
-     Should the checkbox be checked initially (default: 0)?}
-@row3col{undetermined, @ref overview_xrcformat_type_bool,
-     Should a 3-state checkbox be initialized as undetermined (default: 0)?}
+@row3col{checked, integer,
+    Sets the initial state of the checkbox. 0 is unchecked (default),
+    1 is checked, and since wxWidgets 3.1.7, 2 sets the undetermined
+    state of a 3-state checkbox.}
 @endTable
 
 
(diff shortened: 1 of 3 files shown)


docs/doxygen/overviews/xrc_format.h
```diff
@@ -937,6 +937,8 @@ No additional properties.
      Label to use for the checkbox (default: empty).}
 @row3col{checked, @ref overview_xrcformat_type_bool,
      Should the checkbox be checked initially (default: 0)?}
+@row3col{undetermined, @ref overview_xrcformat_type_bool,
+     Should a 3-state checkbox be initialized as undetermined (default: 0)?}
```
vadz, line 941: Could you please add `@since 3.1.7`?

src/xrc/xh_chckb.cpp
```diff
@@ -43,7 +43,11 @@ wxObject *wxCheckBoxXmlHandler::DoCreateResource()
                     wxDefaultValidator,
                     GetName());
 
-    control->SetValue(GetBool( wxT("checked")));
+    if (GetBool("checked"))
+        control->SetValue(true);
+    else if (GetBool("undetermined") && (GetStyle() & wxCHK_3STATE))
+        control->Set3StateValue(wxCHK_UNDETERMINED);
```
vadz, line 49: I think we should report an error if `undetermined` is used for a checkbox without 3 state style, it would be IMO more helpful than silently ignoring it. There is `ReportError()` which could be used for this.

What do you think?

misc/schema/xrc_schema.rnc
```diff
@@ -760,7 +760,8 @@ wxCheckBox =
         stdObjectNodeAttributes &
         stdWindowProperties &
         [xrc:p="important"] element label {_, t_text }* &
-        [xrc:p="o"] element checked {_, t_bool }*
+        [xrc:p="o"] element checked {_, t_bool }* &
+        [xrc:p="o"] element undetermined {_, t_bool }*
```
vadz, line 764: I wonder if we should be really using `bool` for this one. This allows using `<undetermined>false</undetermined>` which is rather strange. Perhaps we should extend `checked` to allow it taking some other value? The trouble is that I don't know which value would fit with the currently allowed `0` and `1` values, neither `-1` nor `2` seem like a good fit.

vadz (commented):
Thanks, we definitely want to have this but I'm not totally sure if using a new attribute is the best way to do it. Does anybody else have any thoughts about this by chance?

And, in any case, I think we should report an error if an attempt to set the 3rd state is made for a 2-state checkbox.

misc/schema/xrc_schema.rnc
vadz: We need to take some decision here if we want to have this in 3.1.7, so let's use 0, 1 and 2 for the states if nobody objects too much.

TIA!

Randalphwa:
I rebased from upstream before amending the PR, but while I was running doxygen to verify the changes to xrc_format.h, another commit got added to master. That's showing up as a merge commit in my branch, and that's beyond by git knowledge to prevent. It's correct with the commit in the master branch, but messes with the history. I'm not sure what else to do on my end (other than being certain in the future to rebase just prior to pushing, rather than before testing).

vadz (commented):
Sorry, I'm not sure what is the problem with Git, but, in any case, I can merge/rebase myself as/if needed. I'd just like to ask you to make the change below (and test that it works as expected by trying to use a wrong parameter value) and then we can merge it.

TIA!

Randalphwa:
Good point. I made the changes and verified that -1 generates the expected warning in a debug build.

vadz:
Great, thanks, will merge soon.

PR #22470, opened 2022-05-29
Add digits property to wxSpinCtrlDoubleXmlHandler

Randalphwa:
This adds the ability to set the precision of the floating point number in a **wxSpinCtrlDoubleXmlHandler**. While it is possible to set the precision by specifying a precision value for the `inc` property, that has the side effect of changing the scroll buttons to use that increment. Note that without setting the precision, a user cannot enter in any non-integer value, as nothing beyond the decimal or comma is accepted (the default precision is zero).


vadz:
Looks good, thanks! Will merge it soon.

PR #22472, opened 2022-05-30
Generic wxDataViewCtrl: Focus event must be sent before selection eve…

valid-ptr:
…nt is sent

### DIFF
--- a/src/generic/datavgen.cpp
+++ b/src/generic/datavgen.cpp
@@ -3678,6 +3678,10 @@ void wxDataViewMainWindow::SendSelectionChangedEvent( const wxDataViewItem& item
     wxAccessible::NotifyEvent(wxACC_EVENT_OBJECT_SELECTIONWITHIN, m_owner, wxOBJID_CLIENT, wxACC_SELF);
 #endif // wxUSE_ACCESSIBILITY
 
+    // Focus event must be sent before selection event is sent
+    if (!m_hasFocus)
+        SetFocus();
+
     wxDataViewEvent le(wxEVT_DATAVIEW_SELECTION_CHANGED, m_owner, item);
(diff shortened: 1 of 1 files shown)


vadz:
Sorry, but I'd appreciate some more explanation. Why must it be sent?

Also, I could understand calling `SetFocus()` if it were done just in `OnMouse()`, but `SendSelectionChangedEvent()` is also called from other places, e.g. `Collapse()` and it's not obvious at all why calling `Collapse()` should give focus to the control. In fact, I'm almost certain that it would be rather unexpected.

valid-ptr:
> Also, I could understand calling `SetFocus()` if it were done just in `OnMouse()`

First of all, I've made it like that for the **left button down**, but then I've found that **right button down** change the selection also. So I've decided to do it inside `SendSelectionChangedEvent()` method.

I've missed that `SendSelectionChangedEvent()` method called from the `Collapse()`

NOTE:
On **_macos_**'s DVC pressing **right mouse button** does not change the selection (if any), it uses current selection (if any).
So it's a little bit different mechanics comparing to generic DVC.

How do you think, is it possible to check not only the **left button down** in the `CONDITION2` at the beginning of `OnMouse()`?:

```
    if (!m_hasFocus && CONDITION2)
        SetFocus();
```

I mean
CONDITION2 = `(event.LeftDown() || event.RightDown())`?
and not
CONDITION2 = `event.LeftDown()`?

vadz:
Sorry, but as it is, it's not going to be merged, so I'm closing it. I am still not sure what exactly the original problem was, but it would make sense to me to call `SetFocus()` on left click, before doing anything else, so if such a change can help you, let's do it. If not, please explain what the actual problem is before we try to solve it. Thanks!

PR #22474, opened 2022-05-30
Update wxToggleButtonXmlHandler

Randalphwa:
This PR adds the missing `pressed`, `focus`, `disabled` and `current` bitmaps to **wxToggleButtonXmlHandler**. It also adds the missing `wxBU_...` styles. Note that this handler will create either `wxToggleButton` or `wxBitmapToggleButton`, so it's necessary to support the styles that affect button text, even though only the `wxToggleButton` control can use them.

I also added the missing "margins" property, matching the wxButtonXmlHandler handler.

### DIFF
--- a/docs/doxygen/overviews/xrc_format.h
+++ b/docs/doxygen/overviews/xrc_format.h
@@ -893,6 +893,17 @@ Example:
 @hdr3col{property, type, description}
 @row3col{bitmap, @ref overview_xrcformat_type_bitmap,
      Label to display on the button (default: none).}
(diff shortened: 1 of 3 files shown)


Randalphwa:
It looks like this got submitted in the midst of four commits changing wxPropertyGridManager causing a build failure. As a contributor, I'm not able to rerun the test to verify the rest of my own changes.

vadz:
Yes, sorry about this, the CI failures have nothing to do with your changes. I'm going to merge this now, thanks (ideally I'd still like to have a common function for this stuff instead of copying the same code in multiple handlers but it could be done in a future PR...).

PR #22475, opened 2022-05-30
Update wxCommandLinkButtonXmlHandler

Randalphwa:
This PR adds the missing `pressed`, `focus`, `disabled` and `current` bitmaps to **wxCommandLinkButtonXmlHandler**.

This also removes the `wxBU_...` styles from this handler because **wxCommandLinkButton** doesn't support them (per the API documentation).

This is the fourth, and I believe final button type control that supports the 4 additional bitmaps besides the main one. For post 3.2, I'll look into moving all 4 of these into a single function call.


vadz:
Thanks!

Issue #14770, opened 2012-10-22
Provide wxFileDialog implementation using IFileDialog to be able to select initial directory under Windows 7


wxtrac:
*Issue migrated from trac ticket # 14770*

**component:** wxMSW | **priority:** normal

#### 2012-10-22 12:35:11: jondo (Robert Pollak) created the issue

___
I have set both the defaultFile and the matching defaultDir to an existing path, but instead of opening this path, my wxFileDialog instance reuses the path that was selected in a former wxFileDialog instance. This wrong path even persists when the application is restarted.

According to [this MSDN page](http://msdn.microsoft.com/en-us/library/windows/desktop/ms646839%28v=vs.85%29.aspx), `lpstrInitialDir` in the `OPENFILENAME` structure is interpreted differently in Windows 7. Maybe that is connected to my problem here?

The page also suggests that the "Common Item Dialog API" should be used instead. Is this already used in newer wxWidgets versions?

PBfordev:
Seeing the issue mentioned in the mailing list...

I tried to look into this and IMO the biggest challenge would be to keep compatibility, i.e., having two implementations: One for the old one and one for the new, limited one.

My idea was to introduce `wxFileDialogExtraControlsSimpl` class, which would have API with limited controls, mimicking `IFileDialogCustomize`. On Windows, the implementation would use `IFileDialogCustomize` implementation (`wxFileDialogExtraControlsSimpleImplIFD`) and on other platforms it would use normal wxWidgets controls (`wxFileDialogExtraControlsSimpleImplGeneric`). 

I wrote some test code but I abandoned the project before I thought out how to best plug it into `wxFileDialog`. Mostly because I would need to test the code on wxGTK and wxOSX which I could not do. Also, it was not that easy and perhaps the whole idea was wrong. IIRC, I also ran into the known limitation about retrieving a value from a text control which was probably the straw that broke my back here.

FWIW, I am attaching `wxFileDialogExtraControlsSimpl`  and some test code here anyway. Feel free to ignore.
[filedlg-ifd.zip](https://github.com/wxWidgets/wxWidgets/files/7873069/filedlg-ifd.zip)

vadz:
Just in case anybody finds this interesting, I finally have something working. It's still not ready to be merged (e.g. all the docs still remain to be written), but I think the API is reasonable by now. Here is the code in the dialogs sample which does the same thing as the current version but using the new API:
```cpp
[...]

PR #22476, opened 2022-05-30
Implement support for custom file dialog controls in new style MSW dialogs

vadz:
This implements the solution for #14770.

Things remaining to be done:
 - [x] Documentation.
 - [x] Fix bug with hiding/disabling radio buttons.
 - [x] More testing, notably under Mac.
 - [ ] Detect attempts to use custom controls when they're not valid any longer.

We could also wrap remaining `IFileDialogCustomize` methods:
 - [x] `AddComboBox()`: should be simple enough.
 - [ ] `AddSeparator()`: even simpler but not clear if really useful.
 - [ ] `{Start,End}VisualGroup()`: need to decide what to do with it in the generic implementation.
 - [ ] `AddMenu()`: doesn't map to any existing wx control, but could probably be emulated using something else.


vadz:
@imciner2 Please let me know if you can test this PR with KiCad, i.e. check migrating at least one of its file dialogs to the new API. TIA!

vadz:
I won't be able to implement all the tasks here before 3.1.7 but I'd still like to make this PR part of it to get more testing.

vadz:
I've added comboboxes support and will really stop here now. If anybody has any comments, please let me know, otherwise I'll merge this PR soon.

PR #22477, opened 2022-05-31
Fix an integer promotion caused test build error

stirwl:
[ 89%] Building CXX object tests/headers/CMakeFiles/test_headers.dir/__/__/__/__/tests/allheaders.cpp.obj
In file included from C:\data\wxWidgets\wxWidgets-src\tests\allheaders.h:283,
                 from C:\data\wxWidgets\wxWidgets-src\tests\allheaders.cpp:412:
C:/data/wxWidgets/wxWidgets-src/include/wx/rawbmp.h: In member function 'wxPixelDataOut<wxBitmap>::wxPixelDataIn<wxPixelFormat<void, 1, -1, -1, -1, -1, bool> >::Reference& wxPixelDataOut<wxBitmap>::wxPixelDataIn<wxPixelFormat<void, 1, -1, -1, -1, -1, bool> >::Reference::operator=(bool)':
C:/data/wxWidgets/wxWidgets-src/include/wx/rawbmp.h:720:29: error: conversion from 'int' to 'wxByte' {aka 'unsigned char'} may change value [-Werror=arith-conversion]
  720 |                     (*m_ptr &= ~mask) |= value;
      |                      ~~~~~~~^~~~~~~~
cc1plus.exe: some warnings being treated as errors

### DIFF
--- a/include/wx/rawbmp.h
+++ b/include/wx/rawbmp.h
@@ -717,7 +717,10 @@ struct wxPixelDataOut<wxBitmap>
                 {
                     wxByte mask = static_cast<wxByte>(1 << m_bit);
                     wxByte value = static_cast<wxByte>(b << m_bit);
-                    (*m_ptr &= ~mask) |= value;
+                    auto& val_m_ptr = *m_ptr;
+                    val_m_ptr = static_cast<wxByte>(val_m_ptr & ~mask);
+                    val_m_ptr |= value;
+
                     return *this;
                 }
                 operator bool() const
--- a/include/wx/rawbmp.h
+++ b/include/wx/rawbmp.h
@@ -717,7 +717,8 @@ struct wxPixelDataOut<wxBitmap>
                 {
                     wxByte mask = static_cast<wxByte>(1 << m_bit);
                     wxByte value = static_cast<wxByte>(b << m_bit);
-                    auto& val_m_ptr = *m_ptr;
+                    wxByte& val_m_ptr = *m_ptr;
+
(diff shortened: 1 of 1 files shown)


vadz:
You're using MSVC with clang, right?

Anyhow, we should fix it, but without using `auto` as we still support C++98. Could you please update the PR to use the explicit type instead? TIA!

stirwl:
> You're using MSVC with clang, right?
> 
> Anyhow, we should fix it, but without using `auto` as we still support C++98. Could you please update the PR to use the explicit type instead? TIA!

Hi vadz,
I'm using cmake + msys2 mingw64 under windows10.
Sure, I will replace 'auto' with its explicit type.

I would like to know more about wx, so that I'd like start from PR#22381 which is a UT issue. The error is reported when I tried to build wx with wxBUILD_TESTS=ALL

vadz:
> > You're using MSVC with clang, right?
> > Anyhow, we should fix it, but without using `auto` as we still support C++98. Could you please update the PR to use the explicit type instead? TIA!
> 
> Hi vadz, I'm using cmake + msys2 mingw64 under windows10.

Oops, sorry, I've somehow misread `cc1plus` as `cc1` (which is a MSVC thing) and got confused.

> Sure, I will replace 'auto' with its explicit type.

Thanks, will merge soon.
 
> I would like to know more about wx, so that I'd like start from PR#22381 which is a UT issue. The error is reported when I tried to build wx with wxBUILD_TESTS=ALL

I could never reproduce it locally, so if you can, it would be great if you could please debug it and try to understand what's going on there. TIA!

PR #22478, opened 2022-05-31
Fix wxCommandLinkButton alphabetical position

Randalphwa:
The alphabetical order for this handler was wrong, affecting not only it's placement in the main HTML page, but also the list of handlers on the right side.


vadz:
Thanks! There was a conflict with #22475 here, I think I've resolved it correctly, but please double check just in case.

PR #22479, opened 2022-05-31
Use translation context for standard OSX menu items

lanurmi:
The context primarily serves as a translator comment, informing that these
are indeed OSX menu item names, and not something else.

Also to match the established translations of OSX itself, for some
languages these may need to be translated using particular terminology
different than on other platforms.


vadz:
This is going to break the existing translations again, isn't it? I really don't know if it's a good idea to do it as I don't know if we can count on @taler21 fixing them up again...

vadz:
Great, thanks a lot @taler21!

I'm going to merge this, together with the two commits corresponding to your patches above.

PR #2582, opened 2021-11-05
Implement wxSecureZeroMemory() for clearing memory securely

lanurmi:
Wrap native function when possible, and fallback to a generic
implementation.

PR #22484, opened 2022-06-04
Always call wxSecureZeroMemory() to wipe memory in wxSecretValue

lanurmi:
Replaces a call to MSW native ::SecureZeroMemory(), and calls to plain
memset().

A sequel to PR #2582.


vadz:
Thanks, will merge soon.

PR #22488, opened 2022-06-05
Improve handling toolbar tools bitmap size

vadz:
To continue the discussion from [your comment in the other PR](https://github.com/wxWidgets/wxWidgets/pull/22481#issuecomment-1146644195):

> > i.e. sizeOrig*int(scale) instead of using ToDIP(). Do you see any reason not to do this?
> 
> I'm more than a little confused by the code involved:

After thinking about this all again, I've realized that I couldn't really explain what was this code doing myself neither, which seems like a good reason not to keep it -- so I didn't. As you can see, I've changed it significantly now and so none of these points:
 
>     1. ~The default size is 16x15 — how is that not causing problems, but other sizes are?~ (because default size is minimum)
> 
>     2. Why is the code explicitly allowing downscaling only in the very edge case of a) different tool sizes, b) with equal count of both, c) that doesn't happen to match the ideal size https://github.com/wxWidgets/wxWidgets/blob/731d29baceda04c5781c6e204cec3d9df8fe9d53/src/common/bmpbndl.cpp#L670-L677
>         I mean, why even bother?
> 
>     3. Why is it feeding the determined size back as the ideal size next time this function runs? Should the input to GetConsensusSizeFor be `m_requestedBitmapSize` only?

should be relevant any longer.

> I would make a different choice in the snippet above - in a tie, I would pick the size closer to `sizeIdeal`, instead of the larger one.

I couldn't even decide what `sizeIdeal` was supposed to be, so I've just dropped it, making this one moot as well.

More generally speaking, I've made `GetConsensusSizeFor()` testable now and while the existing tests are pretty trivial, it should be easy enough to add more test cases, in particular if we find any cases where it doesn't return the expected result. And if we restore `sizeIdeal` parameter in the future, we'd definitely need to add tests verifying that it works as expected (which would at least force us to define what is the expected behaviour).

> As for using `sizeOrig*int(scale)` instead of `ToDIP()`, unless I'm missing something it would be incorrect:
> 
>     1. `sizeOrig` (aka m_defaultWidth/m_defaultHeight) is expressed in physical pixels

Just to clarify a possible misconception: no, it's expressed in logical pixels. I've agonized a long time over this, but I'm relatively confident that this was the best choice and, at the very least, it's documented as being in LPs:

https://github.com/wxWidgets/wxWidgets/blob/34fa234f48f30bd8fef0f442f55debceafa501bf/interface/wx/toolbar.h#L904-L905

>     2. `ToDIP(sizeOrig)` converts it to DIP
> 
>     3. In `GetConsensusSizeFor` it is immediately converted back by multiplying it by `win->GetDPIScaleFactor()`
> 
>     4. (…and isn't actually used for much)

(and now not at all)

> You're right that calling `SetToolBitmapSize()` may be common.

Yes, see also [this post](https://groups.google.com/g/wx-dev/c/rPPODkXKv2k/m/Ii2gU48FBgAJ).

> Unless I misunderstand the problem, I would:
> 
>     1. Outright deprecate `SetToolBitmapSize()`

We could do this, but it would be better not to and I think we may keep it with this PR.
 
>     2. Either ignore the value provided or only use it as hint for `GetConsensusSizeFor`, but don't `IncTo` to it (possibly only if scaling factor > 1.0)

None of this would help with actually selecting the correct bitmap size. E.g. doing this would mean that toggling the bitmap size in the toolbar sample doesn't work at all.
 
>     3. When adjusting the size, use resize, not rescale — I'm not sure if it is feasible, if this is done in win32, probably not

It would be feasible, but I don't see what's the point of having big toolbar buttons with small bitmaps in the middle of them? The current behaviour seems better to me and more consistent with the other controls.

I know I'm asking for a lot, but it would be great if you could please check how does it affect Poedit. Thanks again!

### DIFF
--- a/docs/doxygen/overviews/high_dpi.md
+++ b/docs/doxygen/overviews/high_dpi.md
@@ -299,7 +299,8 @@ programs involves doing at least the following:
    wxArtProvider::CreateBitmapBundle() (and, of course, return multiple bitmaps
    from it) instead of wxArtProvider::CreateBitmap().
 4. Removing any calls to wxToolBar::SetToolBitmapSize() or, equivalently,
-   `<bitmapsize>` attributes from the XRC, as this forces unwanted scaling.
+   `<bitmapsize>` attributes from the XRC, as this overrides the best size
+   determination and may result in suboptimal appearance.
 
 
 
--- a/include/wx/bmpbndl.h
+++ b/include/wx/bmpbndl.h
@@ -154,13 +154,9 @@ public:
     // Get the bitmap size preferred by the majority of the elements of the
     // bundles at the given scale or the scale appropriate for the given window.
     static wxSize
-    GetConsensusSizeFor(double scale,
-                        const wxVector<wxBitmapBundle>& bundles,
-                        const wxSize& sizeDefault);
+    GetConsensusSizeFor(double scale, const wxVector<wxBitmapBundle>& bundles);
     static wxSize
-    GetConsensusSizeFor(wxWindow* win,
-                        const wxVector<wxBitmapBundle>& bundles,
-                        const wxSize& sizeDefault);
+    GetConsensusSizeFor(wxWindow* win, const wxVector<wxBitmapBundle>& bundles);
 
     // Create wxImageList and fill it with the images from the given bundles in
     // the sizes appropriate for the DPI scaling used for the specified window.
--- a/include/wx/bmpbndl.h
+++ b/include/wx/bmpbndl.h
@@ -152,7 +152,11 @@ public:
     // Implementation only from now on.
 
     // Get the bitmap size preferred by the majority of the elements of the
-    // bundles at the scale appropriate for the given scale.
+    // bundles at the given scale or the scale appropriate for the given window.
+    static wxSize
+    GetConsensusSizeFor(double scale,
+                        const wxVector<wxBitmapBundle>& bundles,
+                        const wxSize& sizeDefault);
     static wxSize
     GetConsensusSizeFor(wxWindow* win,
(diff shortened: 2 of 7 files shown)


vadz:
Sorry, I'm finally merging this to avoid doing it at the very last moment before the release (mid-Sunday on its eve is clearly much better), but we can still revert it before 3.1.7 if you (or anybody else) see/find any big problems with it, so please let me know if you do. TIA!

vslavik (approved):
Looks good to me, thanks!

> It would be feasible, but I don't see what's the point of having big toolbar buttons with small bitmaps in the middle of them? The current behaviour seems better to me and more consistent with the other controls.

To be honest, I guess I don't really understand the circumstances in which you'd have differently-sized bitmaps in the toolbar to begin with. I assumed it's all about gracefully handling incorrect behavior from user code (e.g. accidental 1px smaller bitmaps), in which case avoiding scaling would make sense.

> I know I'm asking for a lot, but it would be great if you could please check how does it affect Poedit. Thanks again!

I didn't test this yet (I was waiting with replying until I do — sorry, didn't realize that wasn't the right call today). But Poedit doesn't explicitly set toolbar size, so it shouldn't be affected.

vadz:
Thanks for looking at this!

I'm relatively sure that this improves things for the code which does call `SetToolBitmapSize()` so my main worry was to not break the code not using it somehow. I really don't think it should do that, but testing would be still welcome.

As for why we're doing all this: it's to take into account a scenario when we use 150% scaling and some bitmaps are available in x1.5 size, but others are available in x2 (only). It would seem to make sense to use the scaling factor preferred by the majority of the bitmaps in such a case, I think.

PR #22489, opened 2022-06-05
Fix memory leak in ListBox destructor

Kvaz1r:
Memory is allocate in wxGenericFileCtrl::SetWildcard when client data [append](https://github.com/wxWidgets/wxWidgets/blob/master/src/generic/filectrlg.cpp#L1202) to container but the memory is never released because [wxItemContainer::Clear](https://github.com/wxWidgets/wxWidgets/blob/master/src/common/ctrlsub.cpp#L92-L104) isn't called. DoClear() is called from Clear.


vadz:

PR #22491, opened 2022-06-06
Fix AUI Themed Sash

alilie:
The adornment doesn't get applied to dark theme. Fixed by using proper theme flag. With this flag the sash follows OS theme and the
adornment is drawn.

### DIFF
--- a/src/aui/dockart.cpp
+++ b/src/aui/dockart.cpp
@@ -424,25 +424,16 @@ void wxAuiDefaultDockArt::DrawSash(wxDC& dc, wxWindow *window, int orientation, 
     wxUnusedVar(window);
     wxUnusedVar(orientation);
 
-    if ( wxPlatformInfo::Get().CheckOSVersion(10, 14) && wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOW).Red() < 128 )
-    {
-        dc.SetPen(*wxTRANSPARENT_PEN);
-        dc.SetBrush(m_sashBrush);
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, from my point of view using `kHIThemeSplitterAdornmentMetal` definitely looks like a better idea than what c6523f016e (Dark Mode Patches, 2018-06-21) did here, but the latter was applied by @csomor who knows infinitely more than I do about Cocoa API, so I wonder if there could have been some reason to avoid it here. Notably, is this fine to use on pre 10.14 systems that we still support?

csomor:
Thanks, interesting idea, I wouldn't have thought that metal could have this meaning under Dark Mode, on pre 10.14 we'd have to test whether the window is metal as we do in renderer.cpp I guess... I'll have a look at this...

csomor:
Under 12.4 `kHIThemeSplitterAdornmentNone` gives the correct white divider when drawn horizontally, but a grey divider with an incorrectly oriented knob when drawn vertically. Under Dark Mode the horizontal divider still is white, while the vertical one is black, but still with the wrong knob. Actually I like the white divider better, it also conforms to the 'wide' divider from the GUI guidelines, but since it does not really work for anything than light mode and horizontally, I'd prefer `kHIThemeSplitterAdornmentMetal` to this, and it gives us the knob for free, with a simpler code. I'll check on earlier systems know to see whether the findings match.

csomor:
@alilie Thanks a lot, I've tested under 10.11 and 10.14 as well, it is the better solution everywhere, and since these are AUI windows, the metal grey version fits perfectly anyway.

Issue #22227, opened 2022-03-24
wxWidgets 3.1.5 does not build in xCode 8.3 for i386 arch


guzmanpaco:
**Describe the problem**

We tried to use wxWidgets 3.1.5 source code to build and link wxWidgets on xCode 8.3 on Mac OS 10.12 for our project. It does not compile using Release schema.

**To Reproduce**
1. Perform build steps described in wxWidgets Github (../configure --enable-debug, make)
https://github.com/wxWidgets/wxWidgets/blob/master/docs/osx/install.md

2. Create new project in xCode
3. Add wxcocoa xCode project of wxWidgets to project (build/osx/wxcocoa.xcodeproj)
4. Update project's settings to include the following frameworks:
a. libwx_osx_cocoa_static.a
b. QuartzCore.Framework
c. CoreFoundation.Framework
5. Create Release schema (Change debug setting to release)
6. Perform Build using Release schema and Debug Schema

**Patch or snippet allowing to reproduce the problem**
N/A. Default generated code by xCode can replicate the issue.

**Expected vs observed behaviour**
Expected:
Application builds correctly.

Actual:
Application encounters build error on Release schema (Static was building using 1778 files on Release, 889 files on Debug).
Build error on:
core/graphics.cpp
"Non-constant-expression cannot be narrowed from type 'double' to 'CGFloat' (aka 'float') in initializer list" on line 917 CGSize s = { static_cast<CGFloat>(f),f };

This is not encountered on Mac OS 11.4 using xCode 13. 3.1.6 on master branch does not encounter this issue (has a different issue with static not having volume.mm)

**A clear and concise description of what the bug is.**
Build error occurs using release schema on xCode 8.3 (xCode 8.3 and OS 10.12 are officially supported). Issue is not encountered on xCode 10.13.

**Platform and version information**

wxWidgets version you use: 3.1.5 (Release using source code tar)
wxWidgets port you use: wxOSX
OS and its version: macOS 10.12.6
Compiler being used: xCode 8.3, macos10.12

Attached zip containing:
1. Configure log
2. Make log
3. Build Log from xCode
[logs.zip](https://github.com/wxWidgets/wxWidgets/files/8338450/logs.zip)

lanurmi:
It appears that there are a bunch of other errors with 32-bit compilation in two other files, and at least the ones in `webview_webkit.mm`  do not seem very trivial to fix. 

Here are some examples (not a complete list):
```
[...]

PR #22495, opened 2022-06-06
Drop wxOSX/Cocoa 32 bit builds

vadz:
See #22227.

### DIFF
--- a/include/wx/osx/config_xcode.h
+++ b/include/wx/osx/config_xcode.h
@@ -71,15 +71,9 @@
 #define SIZEOF_WCHAR_T 4
 #define SIZEOF_SHORT 2
 #define SIZEOF_INT 4
-#ifdef __LP64__
 #define SIZEOF_VOID_P 8
 #define SIZEOF_LONG 8
 #define SIZEOF_SIZE_T 8
-#else
-#define SIZEOF_VOID_P 4
-#define SIZEOF_LONG 4
-#define SIZEOF_SIZE_T 4
-#endif
 #define SIZEOF_LONG_LONG 8
 #define wxSIZE_T_IS_ULONG 1
 #define wxWCHAR_T_IS_REAL_TYPE 1
--- a/docs/doxygen/mainpages/platdetails.h
+++ b/docs/doxygen/mainpages/platdetails.h
@@ -52,7 +52,7 @@ Use @c configure option @c \--with-gtk=2 to use GTK 2.
 
 wxOSX/Cocoa is the port of wxWidgets for the macOS platform. It requires
 a minimum SDK 10.11, Xcode 7.2.1 or greater (runs under 10.10.5 and higher),
-and fully supports amd64 as well ARM builds and deploying under 10.10.
+and supports x86_64 (but not i386) and ARM builds and deploying under 10.10.
 
 @subpage plat_osx_install "Build and Install Instructions"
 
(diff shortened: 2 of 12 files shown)


csomor:
@vadz Thanks for the effort, but while we get rid of 32 bit on macOS, I'd rather want to keep CGFloat exposed, there still is some iOS Hardware around that runs iOS 12 with arm7v and watchOS also still has a partially 32 bit world. And `__LP64__` should be there IMHO as well, but any macOS only code of course can get rid of` ! __LP64__`

vadz:
Sorry but why do you want to keep it exposed in public wx headers? Wouldn't any code using iOS hardware include the necessary (CoreGraphics?) headers anyhow? Typically we try to avoid polluting the global namespace with the platform-dependent symbols as much as possible to avoid bad surprises when recompiling the code on another platform, so it seems pretty strange to intentionally do it here.

csomor:
@vadz I see your point about the possible collision with another platform, but CGFloat wouldn't exist in the global space when not compiled on an Apple Platform anyway, or am I overlooking something ? being able to use native types in some public API helps writing cleaner code without having to always cast. But if having it in public hurts more, then so be it.
The non __LP64__ #defines should stay for every Apple platforms except macOS. This would lead to wrong code when compiled on one of the 32 bit platforms we still can run on.

vadz:
> @vadz I see your point about the possible collision with another platform, but CGFloat wouldn't exist in the global space when not compiled on an Apple Platform anyway,

But this is the danger -- suppose somebody uses `CGFloat` as an identifier in their own code (admittedly unlikely with this particular one, but in general we had plenty of problems like this). Their code would compile just fine under MSW and Linux but fail with tons of errors under Mac due to this identifier suddenly becoming a type.

> But if having it in public hurts more, then so be it.

I don't think it's a huge problem in this case, `CG` is a relatively unique prefix, but I just don't see why would we want to do it, and what do we gain from it? Just `CGFloat` on its own is not going to be enough to do anything useful, anybody using it should include CG headers anyhow.

Also, if we keep the public `CGFloat` redeclaration, do we keep `NS[U]Integer` one too? Because this one seems even more fragile and I'd like to get rid of it even more.

> The non **LP64** #defines should stay for every Apple platforms except macOS. This would lead to wrong code when compiled on one of the 32 bit platforms we still can run on.

Sorry, do you mean those in `include/wx/osx/config_xcode.h` or something else?

include/wx/osx/config_xcode.h
```diff
@@ -71,15 +71,9 @@
...
 #define SIZEOF_INT 4
-#ifdef __LP64__
 #define SIZEOF_VOID_P 8
 #define SIZEOF_LONG 8
 #define SIZEOF_SIZE_T 8
-#else
-#define SIZEOF_VOID_P 4
-#define SIZEOF_LONG 4
-#define SIZEOF_SIZE_T 4
-#endif
```
csomor, line 82: please leave this unchanged

csomor (changes requested):

include/wx/osx/config_xcode.h
vadz: Done, thanks.

barracuda156:
What was the point to deliberately break 32-bit builds on one system?

lanurmi:
> What was the point to deliberately break 32-bit builds on one system?

If you read the commit messages of this pull request, and discussions in #22227, you can deduce that the builds were not _deliberately_ broken; they got broken because of the lack of need and interest in testing 32-bit builds. As pointed out, the fixes needed wouldn't have been trivial, and nobody was interested enough to volunteer fixing the builds, or volunteer paying someone else to fix them.

PR #22497, opened 2022-06-07
Fix GetColumnCount for wxLC_LIST style

### LINKED ISSUE #22482: [wxUniv] wxFileDialog in listview mode doesn't allow change directory.
**Describe the bug**
wxFileDialog in listview mode doesn't allow change directory.

It was introduced in commit - https://github.com/wxWidgets/wxWidgets/commit/1e37f6e5ff316001e51377d0a5b1c60bcd730754

The reason - wxListMainWindow is returning 0 as column count because wxGenericFileCtrl [use](https://github.com/wxWidgets/wxWidgets/blob/6d6e5cde21969f91f9f4df85a72f0ae68e423cb1/src/generic/filectrlg.cpp#L942-L951) style wxLC_LIST for wxListCtrl so [m_columns](https://github.com/wxWidgets/wxWidgets/blob/0a8f7233cc6fd55535ae36d7d96324e9eb755f2b/include/wx/generic/private/listctrl.h#L797) isn't contains any column. 

Under MSW wxListCtrl  with wxLC_LIST style is [setting m_colCount](https://github.com/wxWidgets/wxWidgets/blob/d64ba81378a6b9c13db684ab94d852d1bfc5a4c8/src/msw/listctrl.cpp#L316-L317) to 1 in Create function so I guess that generic version should do something similar.

The bug is reproducible in dialog sample:
1. Show dialog (Dialogs -> File operations -> Open file)
2. Try to change directory in wxGenericFileCtrl with double mouse click.
3. Nothing happen.

 - wxWidgets version you use: master
 - wxWidgets port you use: wxUniv
 - OS and its version: Windows 10

Kvaz1r:
Fix: #22482 

Unlike wxMSW the generic version of GetColumnCount returns 0 for wxLC_LIST style that prevent correct handling SendNotify at least under wxUniv. 

I can't test other ports that use generic version so don't know does the bug exist there. If not and such changes may break something I'll update GetColumnCount in PR only for wxUniv with this:

```
int wxGenericListCtrl::GetColumnCount() const
{
#ifdef __WXUNIVERSAL__
    if ( HasFlag(wxLC_LIST) )
        return 1;
#endif
    return m_mainWin->GetColumnCount();
}
```

### DIFF
--- a/interface/wx/listctrl.h
+++ b/interface/wx/listctrl.h
@@ -566,6 +566,11 @@ public:
 
     /**
         Returns the number of columns.
+
+        The control can have multiple columns only in wxLC_REPORT mode. In
+        wxLC_LIST mode this function returns 1, as a list is still considered
+        to have a (single) column. In wxLC_SMALL_ICON and wxLC_ICON modes, it
+        returns 0 as there are no columns at all.
     */
     int GetColumnCount() const;
 
--- a/src/generic/listctrl.cpp
+++ b/src/generic/listctrl.cpp
@@ -2349,7 +2349,7 @@ void wxListMainWindow::SendNotify( size_t line,
 
     // provide information about the (first column of the) item in the event if
     // we have a valid item and any columns at all
-    if ( line != (size_t)-1 && GetColumnCount() )
+    if ( line != (size_t)-1 && GetListCtrl()->GetColumnCount() )
     {
         GetLine(line)->GetItem( 0, le.m_item );
     }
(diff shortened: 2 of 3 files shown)


Kvaz1r:
> Thanks, I think we should change this for all ports, but I'd like to add a new test to `tests/controls/listctrltest.cpp` verifying that they all behave consistently here. Please update it if you can, otherwise I'll try to do it myself later. TIA!


Updated

Kvaz1r:
> Thanks, I think we should change this for all ports, but I'd like to add a new test to `tests/controls/listctrltest.cpp` verifying that they all behave consistently here. Please update it if you can, otherwise I'll try to do it myself later. TIA!


Updated

vadz:
Thanks!

I've added a couple more tests for the other modes too. I'll merge it once the CI builds pass.

vadz:
Hmm, it looks like calling `GetListCtrl()` from `SendNotify()` is a bad idea because this fails when it's called when deleting all items from the dtor, so the apparently innocuous [3eee299](https://github.com/wxWidgets/wxWidgets/pull/22497/commits/3eee299cd317182fc0e460d4fcb1dd1e8e7c5558) broke everything, sorry. Will fix it soon.

PR #22502, opened 2022-06-08
Include macOS 12 name in wxGetOsDescription()

TcT2k:
macOS 13 is called _Ventura_
Additionally remove names for unsupported macOS versions

It seems to me Apple is backing up a little from using the code names for marketing materials, but we'll see what the future brings. For now it's unfortunately still useful to include the code name/marketing name in the OS description.


vadz:
Thanks, will merge soon.

PR #22503, opened 2022-06-08
Fix missing option description, wrong handler name

Randalphwa:
This PR fixes some problems with xrc_format.h

### wxDatePickerCtrl:

Added missing documentation for null-text option.

Note that xrc_schema.rnc already specified this parameter, but it did not exist in xrc_format.h

### wxStyledTextCtrl:

Fixed incorrect handler name


vadz:
Will merge soon, thanks!

PR #22508, opened 2022-06-09
Fix activation of macOS applications using LSUIElement==true

### LINKED ISSUE #16156: LSUIElement Applications stuck before wxApp::OnInit() until activated (e.g. by re-opening application)
*Issue migrated from trac ticket # 16156*

**component:** wxOSX | **priority:** normal | **keywords:** LSUIElement

#### 2014-04-03 21:22:08: sdutoit (Stefanus Du Toit) created the issue

___
A typical use case for an LSUIElement application (one that has LSUIElement set to 1 in its Info.plist) is one that uses just a menubar icon (in wxWidgets using a wxTaskBarIcon).

When building such an app using wxWidgets, the task bar icon won't appear when the application is launched, nor will anything running in wxApp::OnInit() get executed.

The reason for this is that wxWidget's OS X Cocoa implementation seems to rely on an initial activation event, as it calls

```
[NSApp run];
m_onInitResult = OnInit();
```

in its wxApp::CallOnInit() implementation, and then calls [NSApp stop:nil] to stop the run loop for the very first event it receives. When starting a LSUIElement application, no such initial event occurs, so the app just hangs in that call to NSApplication.run until something happens to create a Cocoa event.

I've worked around this in my app by adding the following member function to my wxApp-derived class:

```
    bool Initialize(int& argc, wxChar **argv) override {
        if (wxApp::Initialize(argc, argv)) {
            osxActivateApp();
            return true;
        } else {
            return false;
        }
    }
```

where osxActivateApp() is implemented using Objective-C as follows:

```
void osxActivateApp()
{
    [[NSApplication sharedApplication] activateIgnoringOtherApps: YES];
}
```

It's important to do this after wxWidgets instantiates the NSApplication with the proper release pool and such, which is why I do this in an overridden Initialize(), rather than, say, in my app's constructor.

At this time I'm not sure what the proper fix in wxWidgets is - one option is to add such a call to activateIgnoringOtherApps, but it seems like the proper fix would be to not rely on that initial activation event at all.

vadz:
Without doing this, such applications (e.g. wxTaskBarIcon-based ones)
don't get any events on startup and wxApp::OnInit() doesn't get called.

Closes #16156.

### DIFF
--- a/src/osx/cocoa/utils.mm
+++ b/src/osx/cocoa/utils.mm
@@ -81,14 +81,17 @@ void wxBell()
     [NSApp stop:nil];
     wxTheApp->OSXOnDidFinishLaunching();
 
-    // We need to activate the application manually if it's not part of a
-    // bundle, otherwise not only it won't come to the foreground, but under
-    // recent macOS versions (10.15+), its menus simply won't work at all.
+    // We may need to activate the application manually in a couple of cases.
     //
     // Note that we have not one but two methods to opt out from this behaviour
     // for compatibility.
     if ( !wxApp::sm_isEmbedded && wxTheApp && wxTheApp->OSXIsGUIApplication() )
     {
+        bool activate = false;
+
+        // If the application is not bundled, we need to do it as otherwise not
+        // only it won't come to the foreground, but under recent macOS
+        // versions (10.15+), its menus simply won't work at all.
         CFURLRef url = CFBundleCopyBundleURL(CFBundleGetMainBundle() ) ;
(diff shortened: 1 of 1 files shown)


csomor (approved):
Looks fine to me, there are so many options whether embedded wx situations want or don't want this, we probably will have to add another explicit way to influence things

vadz:
I've merged this now, thanks for the review.

And I think we might extract all this code into some new virtual `wxApp` function and tell people to override it if they need. But I don't know enough about what they might need/want to do, so I won't do it right now. It could be a good idea to add a virtual stub to `wxApp` to be able to do it later during 3.2 lifetime however.

PR #22514, opened 2022-06-10
Add wxAUI_TB_DEFAULT_STYLE as valid style

Randalphwa:
wxAUI_TB_DEFAULT_STYLE is defined as zero, so not supplying any style has the same effect, but it's a bit odd for this handler not to allow this value (most/all other handlers support the various ???_DEFAULT_STYLE variations).


vadz:
Thanks, will merge soon.

PR #22515, opened 2022-06-10
XRC: Add support for SetArtProvider in wxAuiNotebook handler

Randalphwa:
This adds the ability to change the art provider for wxAuiNotebook, which changes the appearance of the tabs. I used the same `art-provider` keyword that is used for wxRibbonBar, though the only common value between them is "default".

I only supplied options for "default" and "simple". Technically, there is a generic provider available for GTK and MSW (MAC automatically uses this), however since it is not documented (even though it exists in the tabart.h header file), and presumably breaks from standard UI on GTK/MSW, I did not add it. 

This could have been implemented as a boolean using "simple" but then it would have to be rewritten if other art providers are ever added, hence the if/else if clauses.

### DIFF
--- a/docs/doxygen/overviews/xrc_format.h
+++ b/docs/doxygen/overviews/xrc_format.h
@@ -741,6 +741,14 @@ wxAuiPaneInfo objects have the following properties:
 
 @subsubsection xrc_wxauinotebook wxAuiNotebook
 
+@beginTable
+@hdr3col{property, type, description}
+@row3col{art-provider, @ref overview_xrcformat_type_string,
+    One of @c default for wxAuiDefaultTabArt or @c simple for wxAuiSimpleTabArt
+    (default: @c default).
+    @since 3.2.0}
+@endTable
+
 A wxAuiNotebook can have one or more child objects of the @c notebookpage
 pseudo-class.
 @c notebookpage objects have the following properties:
--- a/misc/schema/xrc_schema.rnc
+++ b/misc/schema/xrc_schema.rnc
@@ -580,6 +580,7 @@ wxAuiNotebook =
         attribute class { "wxAuiNotebook" } &
         stdObjectNodeAttributes &
         stdWindowProperties &
+        [xrc:p="o"] element art-provider {_, ("default" | "simple") }* &
         (wxAuiNotebook_notebookpage | objectRef)*
     }
 
--- a/src/xrc/xh_aui.cpp
+++ b/src/xrc/xh_aui.cpp
@@ -275,6 +275,14 @@ wxObject *wxAuiXmlHandler::DoCreateResource()
                     GetSize(),
                     GetStyle(wxS("style")));
 
+        wxString provider = GetText("art-provider", false);
+        if (provider == "default" || provider.IsEmpty())
+            anb->SetArtProvider(new wxAuiDefaultTabArt);
+        else if (provider.CmpNoCase("simple") == 0)
+            anb->SetArtProvider(new wxAuiSimpleTabArt);
+        else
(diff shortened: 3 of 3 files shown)


vadz:
Thanks, I agree that this is useful and we should definitely use a name as there could well be other providers in the future.

The only question I have is whether we should be mapping the names to the C++ provider classes manually, as done now, or just interpret the provider name as the class name and try to create it using wxRTTI. The latter would have the advantage of allowing to use custom, i.e. application-defined, providers too, at the cost of requiring to specify `wxAuiDefaultTabArt` instead of just `default` in the XRC file. Maybe we could handle the known names as you do now, but interpret unknown ones as custom provider class names and use `wxClassInfo::FindClass()` for them, followed by `CreateObject()` if the class was found (but still giving an error if it isn't), what do you think?

Randalphwa:
It's a good idea, though I'm having problems with implementation. My initial thinking was to leave the "default", and otherwise let the user specify "wxAuiSimpleTabArt" instead of "simple". However, `wxClassInfo::FindClass("wxAuiSimpleTabArt")` returns a nullptr. Worse, it also returns a nullptr for
`wxClassInfo::FindClass("wxAuiTabArt")` which means I can't use `wxClassInfo::IsKindOf()` to verify that the user-supplied class actually derives from wxAuiTabArt.

Presumably one could add one of the DYNAMIC_CLASS macros to aui/tabart, but I have no idea what the ramifications of that would be for non-XRC calls to wxAuiNoteBook::SetArtProvider.

vadz:
Oh, I didn't even notice that `wxAuiTabArt` didn't inherit from `wxObject` and so neither it nor its subclasses used `wxDECLARE_XXX_CLASS()` neither. I think adding these macros to these classes should be backwards-compatible and AFAICS shouldn't create any problems, as they all have default ctors anyhow (this is a requirement for run-time creation), but this definitely would require far more changes than I had initially thought.

I'd almost certainly say to abandon the entire idea if I didn't think that it might be useful to specify a user-defined art provider in XRC. I'm not sure if this is actually true, mind you... Does anybody else have any thoughts about this, i.e. whether it's worth spending effort on supporting this or not?

Randalphwa:
Something else to consider is that wxRibbonBar does something similar:

```c++
    wxString provider = GetText("art-provider", false);

    if (provider == "default" || provider.IsEmpty())
        control->SetArtProvider(new wxRibbonDefaultArtProvider);
    else if (provider.CmpNoCase("aui") == 0)
        control->SetArtProvider(new wxRibbonAUIArtProvider);
    else if (provider.CmpNoCase("msw") == 0)
        control->SetArtProvider(new wxRibbonMSWArtProvider);
    else
        ReportError("invalid ribbon art provider");
```

And, like `wxAuiTabArt` , `wxRibbonArtProvider `doesn't inherit from `wxObject`. If we are going to change `wxAuiNotebook`, we should probably also change `wxRibbonBar` to be consistent. Of course in the case of `wxRibbonBar` support for the class names would need to be in addition to the shortcuts already provided to maintain backwards compatibility.

Randalphwa:
Perhaps another way to look at this is to take the XRC part out of the equation and ask why `wxArtProvider` inherits from `wxObject` but `wxRibbonArtProvider` and `wxAuiTabArt` do not. By my count, there are 208 wxWidgets classes that inherit from wxObject, so it certainly wouldn't be unusual to have `wxRibbonArtProvider` and `wxAuiTabArt` derive from `wxObject` as well. If adding the inheritance seems like a good idea, then I would add that as a separate PR, and hold off on this PR until that one gets merged.

PR #22517, opened 2022-06-11
Enable tab navigation in wxSimplebook.

adrianlopezroche:
Enable tab navigation in wxSimplebook, which is currently broken under wxGTK (and possibly other platforms).

### DIFF
--- a/include/wx/simplebook.h
+++ b/include/wx/simplebook.h
@@ -14,6 +14,7 @@
 
 #if wxUSE_BOOKCTRL
 
+#include "wx/containr.h"
 #include "wx/vector.h"
 
 // ----------------------------------------------------------------------------
@@ -22,7 +23,7 @@
 
 // NB: This class doesn't use DLL export declaration as it's fully inline.
 
-class wxSimplebook : public wxBookCtrlBase
+class wxSimplebook : public wxNavigationEnabled<wxBookCtrlBase>
 {
 public:
     wxSimplebook()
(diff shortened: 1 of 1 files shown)


adrianlopezroche:
I have written a small test program: [wxSimplebookTest.zip](https://github.com/wxWidgets/wxWidgets/files/8883472/wxSimplebookTest.zip)

oneeyeman1:
Hi,
Are you using wxPanel as children for the *Book?

vadz:
Thanks, I agree that this is the right thing to do and will merge this soon.

Also thanks for providing a test project, but, in the future, a small patch to some sample would be even more appreciated. E.g. this is what I used to test in the notebook sample:
```diff
diff --git a/samples/notebook/notebook.cpp b/samples/notebook/notebook.cpp
index e4eb4c9bb3..c93c3db620 100644
--- a/samples/notebook/notebook.cpp
+++ b/samples/notebook/notebook.cpp
@@ -102,6 +102,8 @@ wxPanel *CreateRadioButtonsPage(wxBookCtrlBase *parent)
     wxBoxSizer *sizerPanel = new wxBoxSizer(wxVERTICAL);
     sizerPanel->Add(radiobox1, 2, wxEXPAND);
     sizerPanel->Add(radiobox2, 1, wxEXPAND);
+    sizerPanel->Add(new wxButton(panel, wxID_OPEN));
+    sizerPanel->Add(new wxButton(panel, wxID_SAVE));
     panel->SetSizer(sizerPanel);
 
     return panel;
```
and it was enough to see the problem under wxGTK perfectly well.

adrianlopezroche:
> ... and it was enough to see the problem under wxGTK perfectly well.
Thanks for merging. I just tried out the sample and it turns out the problem is evident even without adding the wxID_OPEN and wxID_SAVE buttons. With the navigation patch the focus remains always on the textbox. With the patch it goes from "Fox" to "Amiga" to the textbox. So technically, it wasn't really necessary to submit a patch to see the problem.

PR #22522, opened 2022-06-11
Fix calling TransferDataFromCustomControls() with common dialogs

### LINKED ISSUE #22521: wxFileDialog customize hook not firing when using old fallback under MSW
wxFileDialogCustomize hooks currently do not fire `TransferDataFromCustomControls` on "Accept" of the dialog. 
They only fire once the wxFileDialog object goes out of scope / gets destructured.

Is this intentional behavior?

 - OS and its version: Windows 10.0.19044.1706

vadz:
When using common dialogs, because IFileDialog-based implementation is
not available either at compile- or run-time, this function needs to be
called while the extra controls still exist, i.e. before ShowModal()
returns, so do it from CDN_FILEOK handler.

Move the code that called it previously into the new MSWOnFileOK() for
consistency with the other callbacks.

Closes #22521.

### DIFF
--- a/include/wx/filedlg.h
+++ b/include/wx/filedlg.h
@@ -198,6 +198,8 @@ protected:
     wxWindow* CreateExtraControlWithParent(wxWindow* parent) const;
     // returns true if control is created, also sets m_extraControl
     bool CreateExtraControl();
+    // destroy m_extraControl and reset it to NULL
+    void DestroyExtraControl();
     // return true if SetExtraControlCreator() was called
     bool HasExtraControlCreator() const
         { return m_extraControlCreator != NULL; }
--- a/include/wx/msw/filedlg.h
+++ b/include/wx/msw/filedlg.h
@@ -60,6 +60,9 @@ private:
     // called when the currently selected type of files changes in the dialog
     void MSWOnTypeChange(int nFilterIndex);
 
+    // called when the dialog is accepted, i.e. a file is chosen in it
+    void MSWOnFileOK();
+
     // The real implementation of ShowModal() using traditional common dialog
     // functions.
     int ShowCommFileDialog(WXHWND owner);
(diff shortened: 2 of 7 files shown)


marekr:
Something that I will remark about in the case of "fallback" to the common dialog. Not sure it really amtters.

`SetExtraContentCreator` let you just append a wxPanel, using that you could basically pad the content away from the common dialog edge.

Now with the new customize api, there's no way to add the wxPanel with padding, so in the fallback case, the options a visually a bit up on the edge more than aesthically pleasing.

vadz:
I'm afraid there is nothing that allows us to insert a gap like this with `IFileDialogCustomizer`. There is a possibility to add a separator, but it's not very useful and still doesn't allow to control the layout.

Do things work for you in KiCad with this PR? I'd like to merge it soon, but it would be nice to get your confirmation, especially considering how spectacularly wrong I got it the first time.

marekr:
> I'm afraid there is nothing that allows us to insert a gap like this with `IFileDialogCustomizer`. There is a possibility to add a separator, but it's not very useful and still doesn't allow to control the layout.
> 
> Do things work for you in KiCad with this PR? I'd like to merge it soon, but it would be nice to get your confirmation, especially considering how spectacularly wrong I got it the first time.

The commit to fix the IFileDialog invoke for save dialogs works fine. 
The fallback I still need to instrument to test because the vcpkg workflow doesn't make it easy to just edit wxwidgets willynilly.

marekr:
The fallback works fine after instrumenting it, as well as save dialogs using IFileDialog correctly.

It is unfortunate in this case, without a wxPanel and sizer, there's a forth radio button that's cut off until the window is manually resized :/
image

But the fallback case for us will be very rare if not never happen now that IFileDialog works.

Issue #22512, opened 2022-06-10
[wxGTK] wxStyledTextCtrl - bug when using HiDPI


eranif:
**Describe the bug**
When enabling HiDPI or fractional text scaling (for example: 125%), there is an unwanted, unclickable margin which makes the
edit control unusable.
Beside the appearance disturbance, it also causes everything to "shift" to the right.
e.g. if I click with the mouse on column 1, the caret will appear in column 3 or 4

(image attached)
![stc-bug](https://user-images.githubusercontent.com/3677711/173081217-77004ceb-92e4-4ad1-9882-b70671e54c8f.png)

**Expected vs observed behaviour**
See above bug report

**Patch or snippet allowing to reproduce the problem**
To reproduce the problem:

- Open the stc sample
- Open the file `edit.cpp` and add somewhere in the function `Edit::InitializePrefs` this line `SetBufferedDraw(true);` (I added it at line 616)
- Re-compile the sample
- Start the sample, make sure that HiDPI is set

To enable HiDPI, you can either do that from the UI, or from the command line like this:

```bash
export GDK_SCALE=2
./stctest
```
**Platform and version information**
 - wxWidgets version you use: 3.16
 - wxWidgets port you use: wxGTK
 - Ubuntu 20.04

The simple fix (for me) was to simply remove the `SetBufferedDraw(true)` from the code...

This bug was reported by multiple users of mine:

https://github.com/eranif/codelite/issues/2992
https://github.com/eranif/codelite/issues/3010

MaartenBent:
Something inside Scintilla (or wxSTC) doesn't seem to handle the margin width correctly. When drawing each line from the buffer onto the screen, the wrong origin point is used.
For example, with the 16px margin in the STC sample, at 2x scaling it uses `x=16` instead of `x=32` as the origin point. This causing all the lines to shift 16px to the right. The void is filled with black pixels. You have line numbers enabled, so the gap is bigger.

I don't notice any issues on macOS. Maybe the problem could be in the GTK `Blit` implementation? 

An awful hack that seems to fix it for me is adjusting the origin point when blitting the image onto the screen:
```diff
diff --git "a/src/stc/PlatWX.cpp" "b/src/stc/PlatWX.cpp"
index 61d3551bdf..0b0a309007 100644
--- "a/src/stc/PlatWX.cpp"
+++ "b/src/stc/PlatWX.cpp"
[...]

PR #22525, opened 2022-06-12
Fix two issues with GTK3 at HiDPI

MaartenBent:
Fix `wxDC::Blit` and `wxBitmap::GetSubBitmap`, both need to scale the coordinates with the scale factor of the bitmap.

See #22512


vadz:
Great, thanks!

If @paulcor doesn't have any objections, I'll merge this soon.

paulcor:
I'd like to think about this, I'll look at it today.

PR #22527, opened 2022-06-13
Update version to 3.2

vadz:


vadz:
Would anybody know what needs to be done to fix the CMake/MSVC build? I'll try to debug it later, but just from looking at the files, I have no idea why `find_package(wxWidgets 3.2)` doesn't work any longer (in fact, I'd expect even `find_package(wxWidgets 3.1)` to still work too, but it also doesn't...), so if someone already sees it, please let me know to save time. TIA!

MaartenBent:
CMake needs to update their `FindwxWidgets` script. I has some hardcoded version numbers:
https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindwxWidgets.cmake#L294

vadz:
Thanks, but this file is very out of date and [an issue](https://gitlab.kitware.com/cmake/cmake/-/issues/21544) about not supporting any 3.1.x releases in it is still open, so I don't know if it's going to be fixed any time soon.

Can we do anything on our side to make this work? Provide our own `FindwxWidgets.cmake` version? Not use `find_package` at all (but what)? We really need to have some way of using wx 3.2 for CMake users under Windows before it's released.

cc @TcT2k

marekr:
You can provide a custom findwxwidgets but on windows each dependent project has to then include it in their local project tree to override the cmake distribution. 

However if windows users are using vcpkg for cmake (into they really should be to handle all the dependencies), this would get overrides in the vcpkg hooks.

vadz:
Well, we have 2 problems here:

1. The most urgent one, as it prevents me from merging this PR and upgrading the version to 3.2.0: our own minimal sample doesn't work.
2. The slightly less urgent but an even more important one: once we release 3.2.0, CMake users under Windows won't be able to use it.

The problem is that even if CMake maintainers update their `FindwxWidgets.cmake` file, this is not going to help people not using the latest version of CMake, i.e. on the day of 3.2.0 release using it still won't work. Which is why I wonder if there is some other way of consuming wx from CMake projects that we could (a) use in our own sample, solving the first problem right now and (b) recommend to our users, so that they would have a way to solve the second problem in their own projects.

Unfortunately I still don't know enough about CMake myself, but wasn't there a way to just consume the .targets file, which we already install, I believe, instead of using the hacks from `FindwxWidgets.cmake`?

MaartenBent:
Yes, by providing a `wxWidgetsConfig.cmake` file with it. I guess I have to take a look at it again...

TcT2k:
Just for general clarification a link to the documentation of [find_package()](https://cmake.org/cmake/help/latest/command/find_package.html). 
Basically `find_package()` has two modes _config mode_ using `...Config.cmake` and _module mode_ using one of the `FindXXX.cmake? modules.

> Yes, by providing a `wxWidgetsConfig.cmake` file with it. I guess I have to take a look at it again...

Hasn't there been a PR recently which created a `wxWidgetsConfig.cmake` or do I misremember that ?


Anyway it would certainly make sense to change the code in `FindwxWidgets.cmake` to not use fixed versions but maybe just extract the version from `wx/version.h`, but as quick fix a simple addition of 3.2 would probably be ok. But as @vadz already wrote this would require the use an updated cmake including such a change.

vadz:
Ideal would be to do both, I think, i.e. it would be great if somebody could please submit a MR to the CMake issue (its maintainer has just replied saying that it would be welcome, even if a better solution would be even better) so that it would work for the existing `find_package(wx)` users at least with the latest CMake, but using `wxWidgetsConfig.cmake` definitely would make more sense in the long run.

But I'd be grateful for any solution at all, provided we can have something working so that the version could be updated and we could make 3.2.0-rc1 a.s.a.n. TIA for your help!

MaartenBent:
I created a branch with updated `FindwxWidgets` for CMake [here](https://gitlab.kitware.com/MaartenBent/cmake/-/commits/find-wxwidgets/), which uses the version number from `version.h` for the library names, and just one list with major version numbers to find installations directories and wx-config names.
I'll test it some more and create a MR there soon.

> Hasn't there been a PR recently which created a wxWidgetsConfig.cmake or do I misremember that ?

Yes #2391, but it didn't use [CMakePackageConfigHelpers](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html) and had quite some errors so never got added.

I'm working on a minimal config file in https://github.com/MaartenBent/wxWidgets/commits/cmake-config-file which supports component and version checking. Besides a small bug on Linux, it seems to work well. I'll create a PR for this soon too.

vadz:
This works after Maarten's CMake fixes, so I'm going to merge it soon and split 3.2 branch soon afterwards.

PR #22528, opened 2022-06-13
Don't expand slider over best size in wxUniv

Kvaz1r:
Before/after:

un2

mcve:

```
#include <wx/wx.h>

{
public:
    MyFrame(const wxString& title) : wxFrame(NULL, wxID_ANY, title)
    {
        wxBoxSizer* mainSizer = new wxBoxSizer(wxVERTICAL);
        m_slider = new wxSlider(this, wxID_ANY, 50, 0, 100, wxDefaultPosition, wxDefaultSize, wxSL_VERTICAL);
        mainSizer->Add( m_slider, 1, wxALL|wxEXPAND, 5 );
        this->SetSizer(mainSizer);
        this->Layout();
        this->Centre(wxBOTH);
    }
    wxSlider* m_slider;
};

class MyApp : public wxApp
{
public:
    virtual bool OnInit()
    {
        if (!wxApp::OnInit())
            return false;

        MyFrame* frame = new MyFrame("MCVE");
        frame->Show(true);
        return true;
    }
};

wxIMPLEMENT_APP(MyApp);
```

### DIFF
--- a/src/univ/slider.cpp
+++ b/src/univ/slider.cpp
@@ -487,6 +487,15 @@ void wxSlider::CalcGeometry()
 
     // initialize to the full client rect
     wxRect rectTotal = GetClientRect();
+    wxSize bestClientSize = DoGetBestClientSize();
+    if ( !IsVert() && rectTotal.height > bestClientSize.y )
+    {
+        rectTotal.height = bestClientSize.y;
+    }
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, I think you could just hardcode the same `SLIDER_WIDTH` that is used in `DoGetBestClientSize()`, but I guess it's fine to do it like this too, so I'll merge this soon.

PR #22532, opened 2022-06-15
Menu bitmaps fixes

### LINKED ISSUE #22530: wxMenuItem bitmaps not working in specific use case since 3.1.6
**Describe the bug**
In KiCad, we have a scheme where we create a wxMenu with wxMenuItems "virtually" without ever displaying it. We then copy said virtual wxMenu with relevant wxMenuItems when we want to actually display the context menu on right click and pass the copy to `wxWindow::PopUpMenu`. It's mainly because we use dervived classes from wxMenu and wxMenuItem to track some additional metadata and we don't want it to get destructed by wxWidgets automatically.

During the copy process which is done in KiCad, we `GetBitmap` on the "source" `wxMenuItem` and `SetBitmap` with that provided bitmap on the new copy of the menu item. IsOk() validates on the returned wxBitmap and appears to have the 24x24 pixel icon visible in the gdi ref data struct.

In 3.0 up to 3.1.5, this worked fine. Something changed in 3.1.6 and 3.1.7 and it no longer works.

I have debugged deep into wxWidgets and it seems returning valid bitmaps and setting them via `SetBitmap`.
It even invokes `::SetMenuItemInfo` with a valid looking gdi handle to a bitmap when `PopUpMenu` is passed the menu to display. But no bitmap renders in the context menu.

Now, if I invoke `PopUpMenu` with the "virtual" `wxMenu` rather than the copy, then the bitmap appears just fine. 

Another test was I passed a new copy of the `wxBitmap` into the `SetBitmap` call of the copy, rather than using the `GetBitmap`, and it also did not display. 

So I'm a bit lost on what is going wrong and/or changed between wx 3.1.5 and 3.1.6. In both cases, original and "copy" wxMenuItem, we simply use `SetBitmap`, so its not like we are touching internal class state directly.

The bitmap displays just fine.

I am currently also unable to reproduce the bug in a test app where I made it do the same process, i.e. build a wxMenu and wxMenuItem with bitmap but never display it. Then on action, copy those items including bitmap and invoke `PopUpMenu` on the window. It works in this case even if I feed it the same image files we use.

vadz:
Make updating bitmaps from `wxEVT_MENU_OPEN` handler work again and also refactor the code to reduce duplication and allow adding `GetBitmapBundle()` easily -- and do this.

The switch to only calling `SetupBitmaps()` when opening the menu is a non-trivial change, but I don't see how to fix #22530 otherwise.

Any reviews and testing would be very welcome!

### DIFF
--- a/docs/doxygen/mainpages/introduction.h
+++ b/docs/doxygen/mainpages/introduction.h
@@ -91,8 +91,8 @@ NetBSD, Solaris, AIX, ...) and require GTK+ 2.6 or later or GTK+ 3.x. The
 primary supported compiler is GNU g++.
 
 @li wxOSX/Cocoa: This is the native port for Apple computers. wxOSX/Cocoa
-supports 32 or 64 bit Intel Macs running macOS 10.10 or later. The port can be
-built either with g++ or clang.
+supports 32 or 64 bit Intel Macs running macOS 10.10 or later. The port can
+currently be only built using Apple clang.
 
 Other platforms (e.g. iOS - with a minimum requirement of iOS 13), compilers
 (Sun CC, HP-UX aCC, IBM xlC or SGI mipsPro under Unix) and ports (wxQT, wxGTK1,
--- a/docs/doxygen/mainpages/platdetails.h
+++ b/docs/doxygen/mainpages/platdetails.h
@@ -54,6 +54,10 @@ wxOSX/Cocoa is the port of wxWidgets for the macOS platform. It requires
 a minimum SDK 10.11, Xcode 7.2.1 or greater (runs under 10.10.5 and higher),
 and supports x86_64 (but not i386) and ARM builds and deploying under 10.10.
 
+Note that Apple clang must be used to build wxOSX, due to the use of
+Apple-specific extensions ("blocks") in the macOS SDK headers, and hence the
+applications using it must be built using clang as well.
+
 @subpage plat_osx_install "Build and Install Instructions"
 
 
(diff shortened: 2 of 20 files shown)


marekr:
Testing, suddenly a test module I have that uses GetBitmap() on wxMenuItem without any args is now giving

```
        const wxBitmap& bitmap = referenceItem->GetBitmap();
```
E:\kicad\kicad-wx-test\src\wxtest1\wxtest1.cpp(93): error C2668: 'wxMenuItem::GetBitmap': ambiguous call to overloaded function
  E:\vcpkg2\installed\x64-windows\include\wx/msw/menuitem.h(99): note: could be 'wxBitmap wxMenuItem::GetBitmap(bool) const'
  E:\vcpkg2\installed\x64-windows\include\wx/msw/menuitem.h(86): note: or       'wxBitmap wxMenuItem::GetBitmap(void) const'
  E:\kicad\kicad-wx-test\src\wxtest1\wxtest1.cpp(93): note: while trying to match the argument list '()'
```

**Comment by vadz** [comment]
Damn, last minute typo, sorry :-( I'll (force-)push this fix in a moment:
```diff
diff --git a/include/wx/msw/menuitem.h b/include/wx/msw/menuitem.h
index 72c5499cfa..1e8bf02e5e 100644
--- a/include/wx/msw/menuitem.h
+++ b/include/wx/msw/menuitem.h
@@ -96,7 +96,7 @@ class WXDLLIMPEXP_CORE wxMenuItem : public wxMenuItemBase
         DoSetBitmap(bmp, bChecked);
     }
 
-    wxBitmap GetBitmap(bool bChecked = true) const;
+    wxBitmap GetBitmap(bool bChecked) const;
 
 #if wxUSE_OWNER_DRAWN
     void SetDisabledBitmap(const wxBitmapBundle& bmpDisabled)
```

marekr:
Looks good now. GetBitmapBundle() also works

PR #22534, opened 2022-06-15
Fix title bar miss height of floating pane

Kumazuma:
compare
it change fixed size as wxMiniFrame's title bar height to small font's pixel height. 
Right auidemo is before patch, left is after patch

### DIFF
--- a/src/gtk/minifram.cpp
+++ b/src/gtk/minifram.cpp
@@ -91,7 +91,8 @@ static gboolean expose_event(GtkWidget* widget, GdkEventExpose* gdk_event, wxMin
 
     if (win->m_miniTitle && !win->GetTitle().empty())
     {
-        dc.SetFont( *wxSMALL_FONT );
+        const wxFont* smallFont = wxSMALL_FONT;
+        dc.SetFont( *smallFont );
 
         wxBrush brush(wxSystemSettings::GetColour(wxSYS_COLOUR_HIGHLIGHT));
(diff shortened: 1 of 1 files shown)


PBfordev:
Just curious: I don't use Linux, but looking at the screenshots, it seems the AUI tool bar is broken as well?

I mean, the question mark icons in the horizontal toolbar are barely recognizable. In the vertical toolbar, the first two icons seem to be noticeably bigger than the rest?

The sample uses `wxArtProvider` but none of the icons look as if they come from the default Tango-based art provider: Is it not used on GTK or are these icons overwritten by GTK-specific ones which are not of uniform size nor do they scale well?

Kumazuma:
it's running on KDE neon (base on Ubuntu 20.04), so it's toolbar icons are broken. I don't know the solution

vadz:
Thanks, hardcoding the size as it's done now is definitely wrong, but using `FromDIP()` with the pixel size is not great neither, we should use something `dc.GetTextExtent("X").y` instead and add some constant for the margin size (and then use half of this margin instead of hardcoded `2` below). Also, I think we still have to make the height at least 16px so that the icon fits into it, even if the font is smaller.

Could you please check if updating the changes to use `wxMax(GetTextExtent().y, 16)`?

Kumazuma:
It's better, but I don't know it is fine that creating dc in initializing MiniFrame

PR #22536, opened 2022-06-15
CMake: Create a config file for use with find_package

MaartenBent:


MaartenBent:
Ready for review/merge.

vadz:
Thanks a lot!

Should we update `docs/doxygen/overviews/cmake.md` to recommend using the new method?

imciner2:
> Should we update docs/doxygen/overviews/cmake.md to recommend using the new method?

This method is only possible if the system wx library is compiled using its CMake build chain, since it requires wx to create the config files and install them. I believe many places still use the automake build system instead of the CMake-based build system (at least the Debian, Fedora and Homebrew build scripts do), which won't get the CMake config files created and installed. Adding the `CONFIG` flag to the `find_package` step on consumer applications will cause it to only look for the config files, which I think then would cause issues with locating the wx libraries for consumers using the automake-built library version (since it will skip using `FindwxWidgets.cmake` included with projects and CMake itself, and instead look for the non-existent config files).

The only way I can see this becoming the recommended method is if the recommended default build system for wx itself is also changed to be CMake and the packagers update to use that as well.

vadz:
After reading [the docs](https://cmake.org/cmake/help/latest/command/find_package.html) it seems that without either `MODULE` or `CONFIG` both modes will be used, but it seems that `MODULE` is used first and if `FindwxWidgets.cmake` is found, it's probably going to result in an error even if using `CONFIG` would have succeeded, is this right?

So it seems like you'd actually need to use `find_package(... CONFIG)` first and then, if it fails, `find_package(... MODULE)`. But this is, of course, too ugly to recommend and hence it would not be worth updating the documentation to mention it.

So, finally, let's merge this and, if there is some better way, update the docs later.

Thanks again, Maarten!

PR #22539, opened 2022-06-16
XRC: Support additional wxWizard properties

Randalphwa:
This PR adds additional properties to wxWizard. The user can now specify a border to place around the page area, a reserved space for a bitmap, as well as placement flags and background color for the bitmap. All of these use existing wxWizard API.

Example usage:

```xml
	<bitmap-placement>wxWIZARD_VALIGN_CENTRE|wxWIZARD_HALIGN_RIGHT</bitmap-placement>
	<bitmap-minwidth>150</bitmap-minwidth>
	<bitmap-bg>#F0F0F0</bitmap-bg>
```

This sets aside 150 pixels for the bitmap, and if the bitmap isn't large enough, it is aligned on the right side of the reserved area, and the background for the area not containing a bitmap is set to light grey.

### DIFF
--- a/docs/doxygen/overviews/xrc_format.h
+++ b/docs/doxygen/overviews/xrc_format.h
@@ -2517,6 +2517,22 @@ corresponds to the following tree of labels:
 @hdr3col{property, type, description}
 @row3col{bitmap, @ref overview_xrcformat_type_bitmap,
     Bitmap to display on the left side of the wizard (default: none).}
+@row3col{border, integer, Sets width of border around page area. (default: 0).
+    @since 3.2.0}
+@row3col{bmp_placement, @ref overview_xrcformat_type_style,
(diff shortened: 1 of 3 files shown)


vadz:
Thanks, this seems globally fine, but I'd like to nitpick on the names: we don't use neither `_clr` suffix nor `bmp_` prefix in any other XRC handlers, so they look out of place here. For the colour, I'd just use `_bg` because `bg` is already used in several handlers. And I'd use `bitmap` because, again, this is what is used in the existing handlers, resulting in `bitmap-bg` for the full name. And then `bitmap-placement` and `bitmap-min-width` for consistency.

Randalphwa:
I edited the PR example above to match the commit with the name change so that years from now when someone reads the description, they don't get totally confused... 😄

vadz:
Looks good to me, thanks for the update!

Will merge soon if there are no objections.

PR #22541, opened 2022-06-16
XRC: Support SetEmptyCellSize in wxGridBagSizer

Randalphwa:
This PR adds the ability for XRC to call wxGridBagSizer::SetEmptyCellSize() by supporting a `empty_cellsize` property.

### DIFF
--- a/docs/doxygen/overviews/xrc_format.h
+++ b/docs/doxygen/overviews/xrc_format.h
@@ -2733,6 +2733,9 @@ support the following properties:
     optionally the proportion can be appended after each number
     separated by a @c :
     (default: none).}
+@row3col{empty_cellsize, @ref overview_xrcformat_type_size,
+    Size used for cells in the grid with no item. (default: @c wxDefaultSize).
+    @since 3.2.0}
 @endTable
 
 @subsection overview_xrcformat_wxwrapsizer wxWrapSizer
--- a/src/xrc/xh_sizer.cpp
+++ b/src/xrc/xh_sizer.cpp
@@ -253,6 +253,13 @@ wxObject* wxSizerXmlHandler::Handle_sizer()
     m_isInside = true;
     m_isGBS = (m_class == wxT("wxGridBagSizer"));
 
+    if (m_isGBS)
+    {
+        wxSize cellsize = GetSize(wxT("empty_cellsize"));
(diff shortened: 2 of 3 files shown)


src/xrc/xh_sizer.cpp
```diff
@@ -253,6 +253,13 @@ wxObject* wxSizerXmlHandler::Handle_sizer()
     m_isInside = true;
     m_isGBS = (m_class == wxT("wxGridBagSizer"));
 
+    if (m_isGBS)
+    {
+        wxSize cellsize = GetSize(wxT("empty_cellsize"));
+        if (cellsize != wxDefaultSize)
+            wxDynamicCast(sizer, wxGridBagSizer)->SetEmptyCellSize(cellsize);
```
vadz, line 260: I'd use `static_cast<>` here because if the object dynamic type is not `wxGridBagSizer` we're going to crash anyhow.

docs/doxygen/overviews/xrc_format.h
```diff
@@ -2733,6 +2733,9 @@ support the following properties:
     optionally the proportion can be appended after each number
     separated by a @c :
     (default: none).}
+@row3col{empty_cellsize, @ref overview_xrcformat_type_size,
+    Size used for cells in the grid with no item. (default: @c wxDefaultSize).
```
vadz, line 2737: I wonder if it could be useful to link to `wxGridBagSizer::SetEmptyCellSize()` here.

vadz (commented):
This isn't really related to this PR, but setting `m_emptyCellSize` to `(10,20)` by default is wrong, we should use something like `(2*wxSizerFlags::GetDefaultBorderFractional(), 4*wxSizerFlags::GetDefaultBorderFractional())` instead, even if this is not ideal neither for the windows not on the main monitor.

docs/doxygen/overviews/xrc_format.h
vadz: Yes, I guess I mostly thought about doing it because I wasn't sure myself what this was, but it's true that I wouldn't have found much from our current docs... Let's leave it like this then and hopefully remember to add a link here if the documentation of the function itself is ever improved.

vadz:
Will also merge this one soon if nobody objects.

PR #2032, opened 2020-08-25
Monochrome bitmap support (bug #18888)

wsu-cb:
Here is a an attempt to provide basic support for monochrome bitmaps, in particular loading from and saving to files.  I also make wxBitmap(const wxImage&, depth 1) generate a monochrome bitmap.  The resulting bitmap is generated by wxQuantize.  I tried to preserve the colors selected by wxQuantize, but was unable to SetDIBColorTable() work.  Therefore, the resulting bitmap is currently always black-and-white.  Does anyone have any insight into what I am doing wrong with SetDIBColorTable()?

I have also included a basic unit test.

PR #22552, opened 2022-06-18
Fix creating wxBitmap of depth 32 but without real alpha in wxMSW

### LINKED ISSUE #22548: wxBitmap is not correct under wxMSW using a depth of 32
Since 3.1.6 or either 3.1.5 when creating a bitmap using:

`wxBitmap * bmp = new wxBitmap(someImageWithMask, 32, 1.0);`

The displayed bmp is showing in the incorrect size (too narrow) and color distorted like it is not supporting 32-bit. Looks like it is skipping 1 every 4 pixels.

Using:

`wxBitmap * bmp = new wxBitmap(someImageWithMask, wxBITMAP_SCREEN_DEPTH, 1.0);`

will show the bitmap correctly.

Also using 3.1.4 and before will show correctly. Also wxMac will always show correctly, so it's a wxMSW problem.

I have not tested it on 3.1.5.

 - 3.1.5 or 3.1.6
 - wxMSW
 - Windows 11

vadz:
This worked until 7e9afad53a (Add real support for monochrome bitmaps to
wxMSW, 2020-10-19) because we always created the DIB of the depth
matching the alpha presence, i.e. 32bpp only if alpha was used and 24bpp
otherwise, but since this change we can use different depths, and hence
different number of bytes per pixel, for the source and destination
pixel data and so need to account for it.

An alternative fix could be to still create 24bpp DIBs even when depth
of 32 is explicitly requested, as it was done before, but it's probably
better to respect the explicitly specified depth, even if, admittedly,
it's not really clear what, if any, effect does this have in practice.

Closes #22548.

### DIFF
--- a/src/msw/dib.cpp
+++ b/src/msw/dib.cpp
@@ -760,6 +760,13 @@ bool wxDIB::Create(const wxImage& image, PixelFormat pf, int dstDepth)
                     *dst++ = src[2];
                     *dst++ = src[1];
                     *dst++ = src[0];
+
+                    // If source has no alpha, but the DIB does have it, we
+                    // need to still fill it, even if there is no real
+                    // transparency.
+                    if ( dstDepth == 32 )
+                        *dst++ = wxALPHA_OPAQUE;
+
                     src += 3;
                 }
             }


vadz:
@wsu-cb This amends your changes done for #2032, please let me know if you have any comments. Also, I'm still not sure if it's better to fix the problem like this or by doing something like this instead:
```diff
diff --git a/src/msw/dib.cpp b/src/msw/dib.cpp
index 78656c8b33..d79766f459 100644
--- a/src/msw/dib.cpp
+++ b/src/msw/dib.cpp
@@ -652,7 +652,7 @@ bool wxDIB::Create(const wxImage& image, PixelFormat pf, int dstDepth)
     const bool hasAlpha = image.HasAlpha();
     wxCHECK_MSG(!hasAlpha || dstDepth != 1, false, "alpha not supported in monochrome bitmaps");
     const int srcBpp = hasAlpha ? 32 : 24;
-    dstDepth = dstDepth != -1 ? dstDepth : srcBpp;
+    dstDepth = dstDepth == 1 ? dstDepth : srcBpp;
 
     if ( !Create(w, h, dstDepth) )
         return false;
```
(which also fixes the problem).

wsu-cb:
> @wsu-cb This amends your changes done for #2032, please let me know if you have any comments. Also, I'm still not sure if it's better to fix the problem like this or by doing something like this instead:
> 
> ```diff
> diff --git a/src/msw/dib.cpp b/src/msw/dib.cpp
> index 78656c8b33..d79766f459 100644
> --- a/src/msw/dib.cpp
> +++ b/src/msw/dib.cpp
> @@ -652,7 +652,7 @@ bool wxDIB::Create(const wxImage& image, PixelFormat pf, int dstDepth)
>      const bool hasAlpha = image.HasAlpha();
>      wxCHECK_MSG(!hasAlpha || dstDepth != 1, false, "alpha not supported in monochrome bitmaps");
>      const int srcBpp = hasAlpha ? 32 : 24;
> -    dstDepth = dstDepth != -1 ? dstDepth : srcBpp;
> +    dstDepth = dstDepth == 1 ? dstDepth : srcBpp;
>  
>      if ( !Create(w, h, dstDepth) )
>          return false;
> ```
> 
> (which also fixes the problem).

I think it makes sense to do what the caller said to do;  i.e., I think I prefer your committed fix to this alternative.  Maybe the caller is about to edit the alpha channel, and returning a 24bpp DIB would be a nasty surprise.

PR #22558, opened 2022-06-21
Adjust method wxUILocaleImplUnix::GetLocalizedName

utelle:
On non-Linux-like systems the `_NL_*` symbols for retrieving localized strings from a locale are typically not available, causing compile-time errors.

We now check for Linux-like systems, and resort to accessing our own language database otherwise.

In your PR comment https://github.com/wxWidgets/wxWidgets/pull/22538#issuecomment-1160921624 you mentioned the symbol `__LINUX__` that I now used to detect Linux-like systems. However, elsewhere I often see the lowercase symbol `__linux__` for this purpose. I don't know which one would be preferrable / more reliable.

This PR hopefully does not conflict with PR https://github.com/wxWidgets/wxWidgets/pull/22556.

### DIFF
--- a/src/unix/uilocale.cpp
+++ b/src/unix/uilocale.cpp
@@ -191,11 +191,14 @@ private:
     // Call nl_langinfo_l() if available, or nl_langinfo() otherwise.
     const char* GetLangInfo(nl_item item) const;
 
+#ifdef __LINUX__
     // Call GetLangInfo() using either the native or English item depending on
     // the form needed.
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, will merge soon.

`__LINUX__` is defined by our own configure, while `__linux__` is predefined by gcc. We use the former only right now, so let's continue doing it for consistency, even if it's arguably confusing and it would probably be better to define some wx-specific symbol instead.

PR #22563, opened 2022-06-23
Add a FreeBSD CI build

vadz:
It seems to be nice to test under some non-Linux Unix system to avoid introducing Linuxisms in our code.


vadz:

PR #22565, opened 2022-06-24
Fixes for window resizing after DPI change in wxMSW

vadz:
@MaartenBent Please have a look at this, I don't think this is perfect yet, but AFAICS this does solve a real problem which is responsible for many (although, again, maybe not all) of the problems discussed in #18649.


MaartenBent:
I should be able to look at this Sunday.

vadz:
Great, thanks! If possible, I'd really like to merge this before 3.2.0-rc1 which should normally happen on Monday.

MaartenBent:
When I move the widgets or stc sample between monitors (100% and 150%) its size gets bigger and bigger due to `GetBestSize`.

vadz:
Thanks for testing, I didn't see this, but I didn't test with these samples. This must be due to the same old problem of `SetSize()` changing the best size for the window without any natural best size :-( I still think this is better than the current behaviour, when parts of the window are truncated and become completely invisible, but ideal would be to somehow preserve the best sizes. It looks like we could do it by adding another `wxSize` member to `wxWindow` and using a specific function when updating the size here, but it looks ugly and is a huge change to apply just before the release, so I'm not sure about this...

What do you think?

MaartenBent:
I agree that we should not make lots of new changes. Maybe mention this problem in the event documentation too.

Just to test, I tried blocking the resizing in `HandleDPIChange` and then manually resize the window to `oldSize*newDPI/oldDPI`. But this doesn't work. It gets in an endless loop of DPI changes because the resize moves the center of the window to the other monitor.
In theory this could also happen when using `GetBestSize` to increase the suggested size. Resizing could move the center of the window to another monitor. Maybe we should check for this and also adjust the position?

vadz:
I'm not even sure what exactly do we need to document, to be honest... What would you say about it?

To avoid the problem with moving back to the old monitor, I think we may need to add `wxSIZE_SAME_DISPLAY` flag, as this does seem to be a potentially bad problem. Not sure if I can manage to do it before rc1, I just got back recently today and am pretty tired, but I'll try to.

Thanks for testing!

MaartenBent:
I think the documentation you added in https://github.com/wxWidgets/wxWidgets/pull/22565/commits/f0ea00cd35921435122e689abf4ae29e76f0f275#diff-83e4fc461515404ff77fdb69d9befeff0df7ae6042ffd743978c36112a13464f is already out-of-date, since it still mentions a linear size change, not the adjustment with `GetBestSize`. 
Maybe something like:
To make sure all the content fits in the new window size, the default handler adjusts the size using `GetBestSize`. Note that in some cases this causes the window to grow unexpectedly. In this case you could also use `wxWindow::SetSize` in your handler and scale by the DPI ratio yourself.

vadz:
There is still a problem with updating the min size which is too small after moving from 100% to 200%, for example, but I don't know what to do about this and I just can't do anything more today, so I'll merge this tomorrow and include this in rc-1 and anything else will have to wait for 3.2.1, most likely.

PR #22575, opened 2022-06-28
Recognize current GNOME versions in GetDesktopEnvironment()

vadz:
GNOME session manager currently returns "gnome-session", but we only
checked for "GnomeSM" value, which was apparently used in GNOME 2 or
maybe even earlier.

Check for the presence of "GNOME" substring anywhere in the session
manager name (ignoring case) to cover both values and also do the same
for KDE for consistency.


utelle:
Thanks for making `GetDesktopEnvironment()` operational again for newer versions of Gnome.

PR #22579, opened 2022-06-28
New functionality for drop-down buttons on wxAuiToolBar items

carabsc:
This code allows for a new behavior when using drop-down buttons on toolbar items. Currently its assumed that when clicking on a item's drop-down button a menu will popup up to complete the operation, as implemented on the sample. These changes allow the use of generic transient popup windows that will popup when the drop-down button is clicked on and will close when the item's button is clicked on, without requiring a control on the transient window to complete the operation. The behavior is enabled with the style wxAUI_TB_ALLOW_CLICK_WITH_DROPDOWN and the click event handler will report whether the popup window was open or not when the item's button was pressed through the GetExtraLong() method.
The Aui sample was updated to showcase the new behavior.
NOTE: This code HAS NOT been tested on the MacOS port, but has been tested on the Windows and Linux ports.


vadz:
Thanks for the PR, but I'm rather suspicious of all these `if ( HasCapture() )` checks. Could you please explain why were they added?

The new feature also needs to be documented and the build failures need to be fixed (AFAICS it's just a trivial warning about an unused parameter in the sample).

carabsc:
Regarding the HasCapture calls they're used because the CaptureMouse call only happens if the wxEVT_AUITOOLBAR_TOOL_DROPDOWN event is not handled. Instead of having a boolean to keep track of that I've used the HasCapture to find out if that call ran, since each CaptureMouse call has to be paired with a single ReleaseMouse.
The CaptureMouse call comes from the original code and is used because the wxAuiToolBar has no actual buttons; its a single control that responds to mouse events and as such it may capture the mouse on a left mouse down event, releasing it on left mouse up.
Regarding the documentation, I can add some to the wxAuiToolBar class but will probably have to include some sample code to explain the functionality and that will take some time.
Regarding the build failures I'll have to look into that since I'm new to this Github stuff.

vadz:
I'm sorry, I'm not really sure what's going on with this PR but I think the only good way forward is to split it in separate parts and resubmit because it's just impossible to understand anything here any more.

carabsc:
That PR was superseded by several other commits. Right now I have around 70
commits in my fork that fix bugs, implement missing or new functionalities
and improve several others. Since I haven't merged with upstream since the
fork some commits may need rework to merge properly.
For now unless someone wants to cherrypick commits and apply them the best
course will be to ignore the PR, as right now I don't have time to merge
the upstream with my fork.
Em 29/02/2024 02:35, "VZ" ***@***.***> escreveu:

vadz:
It looks like there could be useful things here, but it's impossible to apply it as is, so I have no choice but to close it.

PR #22583, opened 2022-06-29
Fix ListBox item selection 

alilie:
wxListBox's items are not highlighted anymore after changing the background color.
Both "selection" and "*:selected" options must be activated in the css.


vadz:
Thanks, but I'm not sure what does "any more" mean, did it work in previous wx version or with older GTK versions or something else?

@paulcor Does this look right to you?

paulcor:
Thanks, applied somewhat differently as 8064abb0d6c803f824b5b6d956819b16642ccacb

PR #22584, opened 2022-06-29
Fix size of the requested bitmap bundle in wxToolBar XRC handler

vadz:
We need to pass the physical size, not scaled by DPI, as the "size"
parameter of GetBitmapBundle(), so that a user-defined art provider
could use it to select one of the available bitmap sizes (which will
then be scaled, if necessary, by wxBitmapBundle itself).

This change is not 100% backwards-compatible because, in theory, bitmap
sizes could be specified using dialog units previously, but this is not
supported any more. However in practice this never happened because it
simply doesn't make sense to use dialog units for the bitmap sizes and
it doesn't seem worth complicating the code by adding another XRC type
of "DPI-independent pixels that can be expressed in dialog units",
especially considering that DUs are DPI-dependent by definition.

---

I _still_ hadn't got this toolbar bitmap size right, even after struggling with it for so long. But hopefully this is the really last tweak to it.

@kosh543 Could you please check this?

### DIFF
--- a/docs/doxygen/overviews/xrc_format.h
+++ b/docs/doxygen/overviews/xrc_format.h
@@ -2283,8 +2283,11 @@ No additional properties.
 
 @beginTable
 @hdr3col{property, type, description}
-@row3col{bitmapsize, @ref overview_xrcformat_type_size,
-    Size of toolbar bitmaps (default: not set).}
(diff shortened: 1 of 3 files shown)


vadz:
Merged now as I am reasonably confident that this is the right thing to do, but please let me know if anybody has any problems with this.

PR #22587, opened 2022-06-29
Add reserved virtual functions for forward ABI compatibility

### LINKED ISSUE #22555: Add "reserved" virtual functions
To allow making changes to various classes without breaking ABI in 3.2.x series, we need to add stubs for "reserved" virtual functions that wouldn't do anything right now but could be repurposed for something else later -- without affecting the class ABI.

We need to do this at least for:

- [x] `wxApp`
- [x] `wxEvtHandler`
- [x] `wxWindow`
- [x] `wxTopLevelWindow`
- [x] `wxFileDialog`

Anything else?

vadz:
These functions can be repurposed to do something useful in 3.2 branch
later without breaking ABI-compatibility, as overriding an existing
virtual function doesn't break it -- unlike adding a new one.

Closes #22555.

### DIFF
--- a/include/wx/app.h
+++ b/include/wx/app.h
@@ -537,6 +537,10 @@ protected:
 
     friend class WXDLLIMPEXP_FWD_BASE wxEvtHandler;
 
+    // Stub virtual functions for forward binary compatibility. DO NOT USE.
+    virtual void* WXReservedApp1(void*);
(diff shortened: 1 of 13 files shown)


vadz:
Merged into 3.2 branch.

Issue #22590, opened 2022-07-02
Version 3.2 + Vscode - Cmake could NOT find wxWidgets (missing: net core base)


alex290:
**Describe the problem**

When connecting the wxWidgets 3.2.0 RC1 version, Cmake does not see the library. Moreover, all previous versions work perfectly

`Could NOT find wxWidgets (missing: net core base)`

![](https://i.imgur.com/ObAIWgS.png)

![](https://i.imgur.com/B1nfeqy.png)

Here is the Cmake code

    cmake_minimum_required(VERSION 3.20.0)
    project(Test VERSION 0.1.0)
    
    set(CMAKE_CXX_STANDARD 17) 
    set(CMAKE_CXX_STANDARD_REQUIRED ON)
    set(CMAKE_CXX_EXTENSIONS OFF)
    
    IF (WIN32)
      set(wxWidgets_ROOT_DIR "c:/wxwidgets/3.2.0")
      set(wxWidgets_LIB_DIR "c:/wxwidgets/3.2.0/lib/gcc1210_x64_dll")
    ENDIF()
    
    
    
    find_package(wxWidgets REQUIRED COMPONENTS net core base)
    include(${wxWidgets_USE_FILE})
    
    file(GLOB project
        src/*
    )
    add_executable(${PROJECT_NAME} ${project})
    
    
    if(wxWidgets_FOUND)
        target_link_libraries(${PROJECT_NAME} ${wxWidgets_LIBRARIES})
        target_include_directories(${PROJECT_NAME} PRIVATE ${wxWidgets_INCLUDE_DIRS})
        message(STATUS "Found wxWidgets ${wxWidgets_VERSION_STRING}")
    else()
        message(STATUS "wxWidgets Not Found")
    endif()

If you switch to the previous version, everything is fine

![](https://i.imgur.com/4F5diJZ.png)

Used an assembly MinGW 12.10

vadz:
You need to use the latest version of `findwxWidgets.cmake` (which is not available in any released CMake versions yet AFAIK) or use `find_package(... CONFIG)` to make this work.

Unfortunately there is not much else/more we can do about this, Maarten has already fixed the CMake file to find 3.2, but now CMake people need to release a new version with it.

alex290:
That is, you need to wait until version 3.24 is released?

PBfordev:
[...]

PR #22594, opened 2022-07-02
Warn about using find_package(wxWidgets) with older CMake

vadz:
See #22590.

---

I don't really know what to say here, but @PBfordev is probably right and we should at least mention the problem. @MaartenBent could you please check if this is correct and/or fix it if it isn't? TIA!


MaartenBent:
Seems OK to me. Maybe you can link `FindwxWidgets` (with uppercase F) to https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindwxWidgets.cmake

We'll have to see if it can be included in 3.24 or one of its minor updates.

vadz:
Thanks, updated and will merge soon if there are no other comments.

PR #22601, opened 2022-07-04
Fix non-alphanumeric lang id in Windows resources

### LINKED ISSUE #22589: Load translation from Windows resources whose id contains `@`
**Describe the problem**

My program loads translations from Windows resources using `wxResourceTranslationsLoader`, however some language identifications contains charactor `@`, which is not a valid token in .rc files.
For example:
```
foo_sr@latin    MOFILE  "sr@latin/foo.mo"
```
windres will not compile it.

Are there any solutions?

 - OS and its version: Windows 10.0.22000.376
 - Compiler being used: MSYS2/clang

MikeWang000000:
Fix #22589

### DIFF
--- a/src/common/translation.cpp
+++ b/src/common/translation.cpp
@@ -1886,29 +1886,23 @@ wxMsgCatalog *wxResourceTranslationsLoader::LoadCatalog(const wxString& domain,
     const void *mo_data = NULL;
     size_t mo_size = 0;
 
-    const wxString resname = wxString::Format("%s_%s", domain, lang);
-    // Replace non-alphanumeric characters with `_`
-    wxString resname_safe = resname;
-    for ( wxString::iterator it = resname_safe.begin(); it != resname_safe.end(); ++it )
+    // Language may contain non-alphabetic characters that are not allowed in the
+    // resource names that must be valid identifiers, so sanitize the language
+    // before using it as part of the resource name.
+    wxString lang_sanitized = lang;
+    for ( wxString::iterator it = lang_sanitized.begin(); it != lang_sanitized.end(); ++it )
     {
-        if ( !( (*it >= (wxChar) 'A' && *it <= (wxChar) 'Z') ||
-                (*it >= (wxChar) 'a' && *it <= (wxChar) 'z') ||
-                (*it >= (wxChar) '0' && *it <= (wxChar) '9') ||
-                *it == (wxChar) '_') )
-        {
-                *it = (wxChar) '_';
-        }
+        const wxChar c = *it;
+        if ( !((c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9')) )
+            *it = '_';
     }
+    const wxString resname = wxString::Format("%s_%s", domain, lang_sanitized);
 
     if ( !wxLoadUserResource(&mo_data, &mo_size,
                              resname,
                              GetResourceType().t_str(),
                              GetModule()) )
-        if ( !wxLoadUserResource(&mo_data, &mo_size,
-                                resname_safe,
-                                GetResourceType().t_str(),
-                                GetModule()) )
-            return NULL;
+        return NULL;
 
     wxLogTrace(TRACE_I18N,
                "Using catalog from Windows resource \"%s\".", resname);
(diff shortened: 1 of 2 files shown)


src/common/translation.cpp
```diff
@@ -1887,12 +1887,28 @@ wxMsgCatalog *wxResourceTranslationsLoader::LoadCatalog(const wxString& domain,
...
 
     const wxString resname = wxString::Format("%s_%s", domain, lang);
+    // Replace non-alphanumeric characters with `_`
+    wxString resname_safe = resname;
+    for ( wxString::iterator it = resname_safe.begin(); it != resname_safe.end(); ++it )
+    {
+        if ( !( (*it >= (wxChar) 'A' && *it <= (wxChar) 'Z') ||
+                (*it >= (wxChar) 'a' && *it <= (wxChar) 'z') ||
+                (*it >= (wxChar) '0' && *it <= (wxChar) '9') ||
+                *it == (wxChar) '_') )
```
PBfordev, line 1897: Can't we just use `wxIsalnum()` instead of all those comparisons? I also do not understand the check for underscore here.

vadz: Yes, standard functions do depend on locale, so this is fine, but OTOH all these casts to `wxChar` are rather ugly. Why wouldn't something like
```cpp
const wxChar c = *it;
if ( !((c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9')) )
  *it = '_';
```
work instead?

And @PBfordev is right about the cast to `_` not being needed here, of course.

src/common/translation.cpp
```diff
@@ -1887,12 +1887,28 @@ wxMsgCatalog *wxResourceTranslationsLoader::LoadCatalog(const wxString& domain,
...
     if ( !wxLoadUserResource(&mo_data, &mo_size,
                              resname,
                              GetResourceType().t_str(),
                              GetModule()) )
-        return NULL;
+        if ( !wxLoadUserResource(&mo_data, &mo_size,
+                                resname_safe,
+                                GetResourceType().t_str(),
+                                GetModule()) )
+            return NULL;
```
vadz, line 1911: I don't really think we need to try both, if using non-alnum characters doesn't work in any case, using just the "safe" name should be enough.

But if you really want to do it like this, then it would make sense to move the construction of the safe name here, as otherwise we're not going to use it at all in 99% of cases.

src/common/translation.cpp
```diff
@@ -1887,12 +1887,28 @@ wxMsgCatalog *wxResourceTranslationsLoader::LoadCatalog(const wxString& domain,
     size_t mo_size = 0;
 
     const wxString resname = wxString::Format("%s_%s", domain, lang);
+    // Replace non-alphanumeric characters with `_`
+    wxString resname_safe = resname;
```
vadz, line 1891: This comment is not very useful because it describes what the code does (which is clear enough on its own) but not why does it do it, which is the important part. And, thinking about writing this comment, I've realized that we shouldn't need to ever sanitize the domain part, as it shouldn't contain anything non alphanumeric in the first place, AFAICS. Finally, the variable name is not great neither.

Combining all this together I'd do:
```suggestion
    // Language may contain non-alphabetic characters that are not allowed in the
    // resource names that must be valid identifiers, so sanitize the language
    // before using it as part of the resource name.
    wxString lang_sanitized = lang;
    ...
    const wxString resname = wxString::Format("%s_%s", domain, lang_sanitized);
```

vadz (commented):
Thanks, I agree that it's the right thing to do, but it would be great if you could please address the minor comments below.

TIA!

src/common/translation.cpp
MikeWang000000: The original 'if' code was untouched because I wanted to preserve some backward-compatibility. 

The problem is little complicated, as is described [here](https://github.com/wxWidgets/wxWidgets/issues/22589#issuecomment-1173975641) (This kind of resource id is valid in MSVC by adding quotes around it but not valid in MinGW windres. The naming rule of resource id is undocumented so it varies on different implementations). If some one has already using `"myapp_sr@latin"` in MSVC, their app will not load it in the future version of wxWidgets.

` | `myapp_sr@latin` | `"myapp_sr@latin"` | `myapp_sr_latin`
-----|-----|-----|-----
MSVC rc.exe | ❎ (Syntax error) | ✅ | ✅
GNU windres | ❎ (Syntax error) | ❎ (Buggy, unrelated error output)　| ✅
llvm-rc | ❎ (Syntax error) | ❎ (Syntax error)　| ✅


Yes I think 99% people won't met this case, so I finally decided to remove this check to make the code clean. Now both MSVC and MinGW should follow the last rule (`myapp_sr_latin`).

vadz (approved):
Thanks for the update, will merge soon!

PR #22605, opened 2022-07-05
Replace "’s" with "'s" in interface headers

stahta01:
I am trying to build wxPython using SIP version 6.6.2 under win10 using MinGW64 GCC and this is one of two place with SIP problems.
wxPython build stopped the second time on "interface/wx/webview.h".
I found what I am guessing was a "smart" single quote; replacing it with a normal quote helped.

Edit: The other sip problem, that happened first, was caused by wxPython code issue.

Tim S.


vadz:
Thanks, I'll merge this in master and apply to 3.2.0 too, even if this means regenerating the source archives again (but at least the binaries are not affected).

PR #22620, opened 2022-07-10
Ensure that wxrc gets added to the change_install_names

hmaarrfk:
This is somewhat a WIP since I didn't fully test this but I think it is correct.

In the present order of operations, the file change-install-names gets overwritten when the main portion of it is created in the configure script.

It used to be that the code that tested for XML as "below" the creation of the main script, but it got moved "above" it in https://github.com/wxWidgets/wxWidgets/commit/30915c61631f05a06c0c64503540519ae67a5b3d#diff-bb21aa33a3f69ccb36c68b220f40ad08f29b9cd2c05dfedae7b9e3d5d4d08f6bR2949

it seems like the code move from 7500 to 2500. The file `change-install-names` gets created around line 3000 (in configure.in).

In the proposed implementation, I think that It addresses the issues raised in https://github.com/wxWidgets/wxWidgets/issues/15551

Saw from https://github.com/wxWidgets/wxWidgets/issues/15452#issuecomment-1011007656 that this functionaly came at the request from the MacPorts community. I tried to look through their patches (since it seems they released a 3.2 package as well) but I couldn't find any patch related to the one I am submitting. 

I am not so sure how they get around it. Maybe they simply missed this in their test suite ????

### DIFF
--- a/configure
+++ b/configure
@@ -27356,7 +27356,6 @@ $as_echo "$as_me: WARNING: XML library not built, cannot build wxrc" >&2;}
     USE_XML=0
 else
     USE_XML=1
-    echo "${HOST_PREFIX}install_name_tool \${changes} \${2}/wxrc-${WX_RELEASE}" >> change-install-names
 fi
 
 
@@ -33938,6 +33937,10 @@ done
 for i in \${libnames} ; do
     ${HOST_PREFIX}install_name_tool \${changes} -id \${3}/\${i} \${1}/\${i}
 done
+
+if [[ -f "\${2}/wxrc-${WX_RELEASE}" ]]; then
+    ${HOST_PREFIX}install_name_tool \${changes} \${2}/wxrc-${WX_RELEASE}
+fi
 EOF
             chmod +x change-install-names
             DYLIB_RPATH_INSTALL="\$(wx_top_builddir)/change-install-names \${DESTDIR}\${libdir} \${DESTDIR}\${bindir} \${libdir} \$(wx_top_builddir)/lib"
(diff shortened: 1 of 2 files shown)


hmaarrfk:
I can confirm that this addresses the particular issues i was seeing.

I have to point out that i did not regenerate the configure file, rather i manually edited it. I'm not sure what your policy on updating that file is

oneeyeman1:
Hi,
The configure file is updated thru the "Bakefile" project.

There is a plan to drop it and use something else (most likely CMake) after 3.2.0 release, but as of this writing - "Bakefile" is a way to generate the Makefile.

Thank you.

hmaarrfk:
Thank you for your response. I'm not too sure how to use bakefile. I've never used them before.

Do you hav instructions for me to follow. Alternatively, you could push to my branch

MaartenBent:
This is incorrect. `configure.in` is updated manually, and then `configure` is created by `autoconf`/`autogen.sh`.
Only `Makefile.in` is generated with bakefile.

vadz:
Thanks for noticing and fixing this problem!

I'm going to apply your patch, but use `test` instead of `[[ ... ]]` because `[` and `]` are special in m4 and while the patch works because `[[ ... ]]` becomes `[ ... ]` after passing through autoconf, this is somewhat confusing. AFAICS this shouldn't change anything, but please let me know if it did, i.e. if the version that I'm about to merge into master doesn't work for you.

hmaarrfk:
Thanks!

https://github.com/conda-forge/wxwidgets-feedstock/pull/19

We can monitor that. It should be done in 30 mins

hmaarrfk:
green should mean all good.

hmaarrfk:
Seems green!

vadz:
I'm reopening this because we'll want to backport it to 3.2 branch too, I think (and so it will be closed when this is done).

PR #22622, opened 2022-07-10
Rename the existing changelog to changes_32.txt and add a new one

vadz:
To prevent the changelog file from growing too big, start a new one for
the next release.

Also restore the instructions about updating it from Git notes just
before the release.

---

If nobody has any better/more advanced ideas about the changelog files management, I'm going to do it like this, i.e. simply keep adding new `changes_XY.txt` files.


vadz:
I'm going to just ignore the spell check failure in the new file, it's not going to change anyhow in this branch.

PR #22623, opened 2022-07-12
Fix wxBitmapBundle compatibility for template NSImages

vslavik:
Fix backward compatibility regression where native NSImage bitmaps, such as those loaded via wxArtProvider on macOS, would loose their isTemplate property when implicitly converted to wxBitmap and back. This resulted in broken rendering of templates.

This is a common situation in pre-wxBitmapBundle code. wxArtProvider returns wxBitmapBundle, which is then implicitly converted to a single wxBitmap in wxArtProvider::GetBitmap() and when its later passed to a function that takes wxBitmapBundle, it is converted back using the wxBitmapBundleImplSet implementation.

wxBitmapBundleImplSet serves dual role: it is used both for manually assembling bitmap bundles and to implicitly faciliate the backward compatiblity mechanism described above. To make the latter truly transparent, we need to special-case it and preserve NSImage as is, and this commit does exactly that.

---

@vadz Sorry for not finding it in time for the release; I misremembered the schedule and thought it was planned for early August :(

@csomor Please review the bit where I claim that "wxOSXGetImageFromCGImage […] is essentially identical to wxOSXImageFromBitmap" — I am *not* 100% confident in it.

### DIFF
--- a/src/common/bmpbndl.cpp
+++ b/src/common/bmpbndl.cpp
@@ -312,19 +312,41 @@ wxBitmap wxBitmapBundleImplSet::GetBitmap(const wxSize& size)
 void wxBitmapBundleImplSet::OSXCreateNSImage()
 {
     WXImage image = NULL;
-#if wxOSX_USE_COCOA
-    image = wxOSXImageFromBitmap(m_entries[0].bitmap);
     const size_t n = m_entries.size();
-    for ( size_t i = 1; i < n; ++i )
+    if ( n == 1 )
     {
-        const Entry& entry = m_entries[i];
-        wxOSXAddBitmapToImage(image, entry.bitmap);
+        // The special case of only a single bitmap in the bundle is common and
+        // occurs in older code using wxBitmap instead of wxBitmapBundle. We want
+        // to avoid creating a new NSImage as is done below in this case for two
+        // reasons:
+        //
+        // - performance - returning existing NSImage is much cheaper
+        // - backward compatibility - using existing NSImage preserves existing
+        //   metadata such as image name or, importantly, its isTemplate property
+        //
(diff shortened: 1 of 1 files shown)


vadz (approved):
Thanks, I don't see any problems with this, but I'd feel much better if @csomor could confirm that this is correct too because I could be totally missing something here (just as I missed the existence of any metadata initially) without even knowing what am I missing.

Speaking of metadata, what happens if you use 2 existing `NSImage`s? Is there no problem with losing their metadata in this case? They probably won't (both?) have `isTemplate` set, but there might be some other properties that could be worth preserving, is there any need to do it?

vslavik:
> Speaking of metadata, what happens if you use 2 existing NSImages? Is there no problem with losing their metadata in this case? They probably won't (both?) have isTemplate set, but there might be some other properties that could be worth preserving, is there any need to do it?

I don't think this a case to be concerned about — or that we can do anything meaningful with. None of these properties, including isTemplate, are exposed in wx API and then only way they get involved is when you create a bitmap from native NSImage. If you do that, it is already a multi-resolution representation and you don't need to do any manual assembly of the bundle in the first place. The >1 images case is for bitmaps loaded manually or generated, not for combining multiple NSImages. 

There's more stuff in `NSImage` besides isTemplate (e.g. symbolConfiguration), but it's all related to creation and it's unclear how we'd combine them when merging multiple images.

csomor (approved):
Hi @vslavik thanks a lot for this round-trip-fidelity and optimization. I have checked the execution paths and and ownership counts and I don't see a difference which might cause havoc. Since I only have exposed the isTemplate to the BitmapReference level, I don't see how someone could create a non NSImage based template using the public API.

PR #22627, opened 2022-07-14
Improvements to closing wxDocument

vadz:
This started as a bug fix for the [problem with calling `OnCloseDocument()` twice](https://groups.google.com/g/wx-users/c/RxaqAIS1R1c/m/PcrEdg7QAQAJ), but also includes some related, even though independent, improvements.


vadz:
I'm finally going to merge this because I think it's an improvement, even though I haven't received any feedback about this change.

PR #22631, opened 2022-07-16
Fix wx-config and wxrc links when using DESTDIR with CMake

### LINKED ISSUE #22610: Symlink installations doesn't homor DESTDIR
**Describe the problem**
Cmake doesn't honor DESTDIR when creating the symlinks for wxconfig and wxrc during installation. It always tries to install them to the root filesystem instead of the specified DESTDIR, which fails if the filesystem is write-only

```
$ cmake -B build-dir -S wxWidgets-3.2.0 -DCMAKE_INSTALL_PREFIX=/usr
$ cmake --build build-dir
[...]
$ DESTDIR='foo' cmake --install build-dir
[...]
-- Installing: foo/usr/lib/libwx_baseu_xml-3.2.so
-- Installing: foo/usr/bin/wxrc-3.2
-- Set runtime path of "foo/usr/bin/wxrc-3.2" to ""
CMake Error: failed to create symbolic link '/usr/bin/wxrc': Permission denied
```

 - wxWidgets version you use: 3.2.0
 - wxWidgets port you use: gtk3
 - OS and its version: Linux x86_64
 - Compiler being used: gcc 12.1

vadz:
Don't try installing the symlinks under the root directory when DESTDIR
is specified.

Closes #22610.

### DIFF
--- a/build/cmake/install.cmake
+++ b/build/cmake/install.cmake
@@ -43,9 +43,10 @@ else()
     install(CODE "execute_process( \
         COMMAND ${CMAKE_COMMAND} -E create_symlink \
         ${CMAKE_INSTALL_PREFIX}/lib/wx/config/${wxBUILD_FILE_ID} \
-        ${CMAKE_INSTALL_PREFIX}/bin/wx-config \
+        ${CMAKE_CURRENT_BINARY_DIR}/bin/wx-config \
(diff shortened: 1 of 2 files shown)


vadz:
Superseded by #22653.

PR #22633, opened 2022-07-16
wxGTK/Win: Use wxGetDisplayInfo only under Unix

stahta01:


stahta01:
This PR is needed to build wxGTK/Win when using MinGW64 GCC and likely is needed for MSVC builds.

PR #22634, opened 2022-07-17
Add CI workflow checking ABI stability

vadz:
Add the ABI files generated by libabigail abidw tool and a script
comparing the current ABI with these files.

Also update the documentation to describe using libabigail tools as
abi-compliance-checker doesn't seem to work under the recent systems any
longer: it gives some warnings and doesn't detect ABI-incompatible
changes.


vadz:
Applied as 93577ce1b328f456ea4b865921f3e3f101fb09e2

Issue #22543, opened 2022-06-16
Add support for `IFileDialog::AddPlace`


marekr:
`IFileDialog::AddPlace` exists in IFileDialog and allows you to add folders to the Navigation pane on the left hand side of the file dialogs for users to quickly jump to. Very useful for application folders that are always reused by users.

GTK has `gtk_file_chooser_add_shortcut_folder`

I don't know if macOS has an equivalent because their docs are horrible to read.

vadz:
I don't see anything allowing to do this in [Apple docs](https://developer.apple.com/documentation/appkit/nssavepanel?language=objc) but we could indeed do this relatively easily for MSW and GTK.

vadz:
Please check the linked PR, it seems to work fine for me under both MSW and GTK. If no problems are found with it, it can also be backported, with minor adjustments, to 3.2.

cc @imciner2 @sethhillbrand

vadz:
Backported to 3.2 too now.

marekr:
> Backported to 3.2 too now.

Thanks

PR #22635, opened 2022-07-17
Add wxFileDialog::AddShortcut()

vadz:
See #22543.

### DIFF
--- a/include/wx/filedlg.h
+++ b/include/wx/filedlg.h
@@ -56,6 +56,13 @@ enum
 
 #define wxFD_DEFAULT_STYLE      wxFD_OPEN
 
+// Flags for wxFileDialog::AddShortcut().
+enum
+{
+    wxFD_SHORTCUT_TOP       = 0x0001,
+    wxFD_SHORTCUT_BOTTOM    = 0x0002
+};
+
 extern WXDLLIMPEXP_DATA_CORE(const char) wxFileDialogNameStr[];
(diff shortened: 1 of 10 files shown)


imciner2:
Can you expand on the reason it doesn't work with the extra control creator? Reading the documentation for the IFileDialog in MSW, it doesn't say it is an incompatibility, so I am curious if it was an issue you observed or if you found a different documentation page that explains this. (it would be good to clarify that in a comment above the if check also).

imciner2:
> It doesn't work with the old-style control creator because it forces the use of common dialog, in which this functionality is unavailable. It works perfectly fine with the new customization hook API.

Ah, I had misread the text and thought it didn't work with any additional controls (both newer and older style), not just the old style of adding the controls. Now that I read the API docs again, I see it says to use the newer customize hook with the function and it should work.

marekr:
I'm not setup to test this currently but it looks about what I expect.

The only thing I can think of for documentation is on Windows, the shortcuts will appear under an "Application Links" group in the  dialog, unless the user defines a VERSIONINFO section in their manifest which will cause it to use FileDescription and the app icon instead of "Application Links".

vadz:
Yes, good point, thanks, I'll add this:
```
        @note In wxMSW, the shortcuts appear in a separate section called
            "Application Links" by default. To change the title of this
            section, the application can specify a value of the @c
            FileDescription field of the version information structure in its
            resource file -- if present, this string will be used as the
            section title.
```
to the documentation.

PR #22637, opened 2022-07-18
macOS: wxWebView Remove fullscreen without prefix

TcT2k:
The previous implementation included the fullscreen API with and without
the `webkit` prefix. Safari itself only provides the prefixed version.
This improves compatiblity at least with popular JS library video.js
and should also improve the situation in general as it closer to native
safari functionality (which is not available through `WKWebView`).

This should also be backported to 3.2.


TcT2k:
Currently looking into this a little more, marked as draft till I'm finished.

TcT2k:
Added additional fixes. Tested with various javascript frameworks and sites. Ready for inclusion now, as it might not be a perfect emulation, but much better than the current implementation.

PR #22641, opened 2022-07-18
Fix crash when using wxTextCtrl with TDM-GCC

vadz:


acotty:
Does this change work with MSYS2 Mingw64 and MSYS2 Mingw32 compilers?
Does this change work with the Windows winlib compiler?

vadz:
> Does this change work with MSYS2 Mingw64 and MSYS2 Mingw32 compilers?

What do you mean by "work"? It definitely works in the sense that it doesn't break it (how could it?), but I'm not sure if it's ABI-compatible with 3.2.0, this would need to be tested.

> Does this change work with the Windows winlib compiler?

Sorry, what's this?

Kumazuma:
MinGW-w64 binaries are provided on https://winlibs.com
I tested wxTextCtrl crash on MinGW-w64 10.3 provided there but it is not crashed.

vadz:
Yes, the original problem really seems specific to TDM-GCC libstdc++. I still don't know what it is, but as long as we can avoid it by making the ctors non-inline, I won't spend more time on it.

Also, thinking more about this, I believe this change _is_ ABI-compatible, after all: an existing application can either link to the out-of-line version of the ctor in the DLL, if any, in which case it will still work now because this ctor is still there, or, much more likely, not contain any references to the DLL symbols at all as the trivial ctor body was inlined in the application code -- and in this case it's not affected by the change neither. So I'll backport this to 3.2 too without any changes, and notably without any tests for TDM-GCC.

PR #22653, opened 2022-07-19
Use DESTDIR when creating symlinks in CMake install

### LINKED ISSUE #22610: Symlink installations doesn't homor DESTDIR

vadz:
Prepend $ENV{DESTDIR}, sufficiently quoted to delay its expansion until
the execution of "cmake -E create_symlink" command, to the command
path arguments.

Closes #22610.

### DIFF
--- a/build/cmake/utils/CMakeLists.txt
+++ b/build/cmake/utils/CMakeLists.txt
@@ -30,17 +30,18 @@ if(wxUSE_XRC)
         BUNDLE DESTINATION "bin"
         )
 
-    if(NOT WIN32_MSVC_NAMING)
+    if(NOT WIN32_MSVC_NAMING AND wxBUILD_INSTALL)
         if(IPHONE)
             set(EXE_SUFFIX ".app")
         else()
             set(EXE_SUFFIX ${CMAKE_EXECUTABLE_SUFFIX})
         endif()
 
-        wx_install(CODE "execute_process( \
+        # Don't use wx_install() here to preserve quoting.
+        install(CODE "execute_process( \
             COMMAND ${CMAKE_COMMAND} -E create_symlink \
             ${CMAKE_INSTALL_PREFIX}/bin/${wxrc_output_name}${EXE_SUFFIX} \
-            ${CMAKE_INSTALL_PREFIX}/bin/wxrc${EXE_SUFFIX} \
+            \"\$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/wxrc${EXE_SUFFIX}\" \
             )"
         )
     endif()
--- a/build/cmake/install.cmake
+++ b/build/cmake/install.cmake
@@ -43,7 +43,7 @@ else()
     install(CODE "execute_process( \
         COMMAND ${CMAKE_COMMAND} -E create_symlink \
         ${CMAKE_INSTALL_PREFIX}/lib/wx/config/${wxBUILD_FILE_ID} \
-        ${CMAKE_INSTALL_PREFIX}/bin/wx-config \
+        \"\$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/wx-config\" \
         )"
     )
 endif()


build/cmake/utils/CMakeLists.txt
```diff
@@ -39,8 +39,8 @@ if(wxUSE_XRC)
 
         wx_install(CODE "execute_process( \
```
MaartenBent, line 40:
```suggestion
        install(CODE "execute_process( \
```

MaartenBent (changes requested):

build/cmake/utils/CMakeLists.txt
MaartenBent: It doesn't work with the `wx_install` macro, it resolves the escaped variables I think.

build/cmake/utils/CMakeLists.txt
```diff
@@ -30,17 +30,18 @@ if(wxUSE_XRC)
...
-    if(NOT WIN32_MSVC_NAMING)
+    if(NOT WIN32_MSVC_NAMING AND wxBUILD_INSTALL)
         if(IPHONE)
             set(EXE_SUFFIX ".app")
         else()
             set(EXE_SUFFIX ${CMAKE_EXECUTABLE_SUFFIX})
         endif()
 
-        wx_install(CODE "execute_process( \
+        # Don't use wx_install() here to preserve quoting.
```
MaartenBent, line 40: Just nitpicking, but it is to preserve escaping.

build/cmake/utils/CMakeLists.txt
```diff
@@ -30,17 +30,18 @@ if(wxUSE_XRC)
...
+        install(CODE "execute_process( \
             COMMAND ${CMAKE_COMMAND} -E create_symlink \
             ${CMAKE_INSTALL_PREFIX}/bin/${wxrc_output_name}${EXE_SUFFIX} \
-            ${CMAKE_INSTALL_PREFIX}/bin/wxrc${EXE_SUFFIX} \
+            \"\$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/wxrc${EXE_SUFFIX}\" \
```
MaartenBent, line 44: I think for consistency we should use quotes around both, or none. Not just around one.

vadz: So do you mean that _not_ quoting this line would still work? If so, let's not do this... I'd rather not add quotes around the otherwise unchanged line -- or, if we want to do it, do it everywhere and not just here. But see above about my understanding of CMake quoting.

MaartenBent: The quotes are escaped because they are used within other quotes: `install(CODE " ... \"quoted text\"  ... ")`. This is maybe less obvious because the command is split over multiple lines.

I added the quotes so it is possible to have spaces in the install path. for example the resulting code might look like: `cmake -E create_symlink "/test dir/bin/wxrc-3.2" "/test dir/bin/wxrc"`. Without quotes this will fail.
I've never tried spaces in the install path (or destdir) so there are probably other places that also should be quoted. So for now you could leave them out and it can be part of a PR that adds them everywhere where needed.

The variable `$ENV{DESTDIR}` is escaped so it will _not_ be evaluated during CMake's configure step, and it will appear as-is in the `cmake_install.cmake` script that is executed when installing.


Using `${ARGN}` in the `wx_install` macro seems to cause everything to be re-evaluated, removing the escape. Thats why I added the `wxBUILD_INSTALL` check and use `install` directly.
It is possible to keep using `wx_install` by escaping the escape: `\\\$ENV{DESTDIR}`

vadz: OK, so for now I'll just remove the (escaped) quotes, right?

MaartenBent: yes

MaartenBent: I added the quotes in #22660

vadz:
Sorry for the delay, I was taken by other things but will try to get back to wx and merge your PR a.s.a.p.

MaartenBent:
We still need the `\$ENV{DESTDIR}` fix from this PR though.
Or shall I add it in a commit in my PR?

vadz:
Oh sorry I didn't even notice this. Let's apply this one and then yours on top of it then.

MaartenBent:
Just remember to remove the escaped quotes in this one then. And maybe update the [comment](https://github.com/wxWidgets/wxWidgets/pull/22653#discussion_r924785981).

vadz:
I hope I've done it correctly, finally, but please let me know if there is still anything wrong.

PR #22651, opened 2022-07-19
AUI floatpane drifting on Windows fixed

valid-ptr:
Solves #22533 issue

PR #22663, opened 2022-07-24
Proposed backports to 3.2

vadz:
This backports recent changes to master that seem both important and safe to make in 3.2.


acotty:
Can you include PR #22651?

vadz:
> Can you include PR #22651?

Yes, I'll include it too.

PR #22665, opened 2022-07-24
Don't use system libcurl under Mac with --disable-sys-libs

vadz:
It doesn't seem right that the system zlib (installed by e.g. MacPorts)
can be picked up by configure even when --disable-sys-libs is specified,
so disable its use under Mac, as libcurl is optional there.

Still use the system libcurl under Linux in the same case, however, as
it's required to have wxWebRequest at all there and dependency on it is
not really a problem in practice.

This commit is best viewed with Git --color-moved
--color-moved-ws=ignore-all-space -w options.

----

Any comments are welcome, as usual. I realize that it's weird to handle the same option differently under different platforms, but it really seems to make sense in this case.

### DIFF
--- a/configure
+++ b/configure
@@ -27421,7 +27421,14 @@ if test "$wxUSE_LIBMSPACK" != "no"; then
 fi
 
 
-if test "$wxUSE_WEBREQUEST" = "yes" -a "$wxUSE_LIBCURL" != "no"; then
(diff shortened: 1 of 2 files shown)


vadz:
No comments since almost a week, so merging.

vadz:
applied as 6ed6d97bfeb5a49cb7d2d4a2fec34362488dd553

PR #22670, opened 2022-07-26
Replace raw stat/lstat with wx macros

gbmaster:
In some places, the wxStat macro is not honoured and the original version is used, even if the wxStructStat is used as a parameter.

### DIFF
--- a/src/common/ffile.cpp
+++ b/src/common/ffile.cpp
@@ -336,7 +336,7 @@ bool wxTempFFile::Open(const wxString& strName)
     mode_t mode;
 
     wxStructStat st;
-    if ( stat( (const char*) m_strName.fn_str(), &st) == 0 )
+    if ( wxStat(m_strName, &st) == 0 )
     {
         mode = st.st_mode;
     }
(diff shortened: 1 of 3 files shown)


src/common/ffile.cpp
```diff
@@ -336,7 +336,7 @@ bool wxTempFFile::Open(const wxString& strName)
     mode_t mode;
 
     wxStructStat st;
-    if ( stat( (const char*) m_strName.fn_str(), &st) == 0 )
+    if ( wxCRT_Stat( (const char*) m_strName.fn_str(), &st) == 0 )
```
vadz, line 339: As long as we're doing this anyhow, we should use our wrappers functions instead of `wxCRT_xxx` macros directly, i.e.
```suggestion
    if ( wxStat(m_strName, &st) == 0 )
```
Otherwise there is really not much point, is there?

vadz (changes requested):
Thanks for the cleanup (at least I guess the only reason for doing this is this, but please let us know if there is some actual problem this solves), but could you please go one step further and use the wrapper functions instead of the macros directly?

TIA!

vadz:
Looks good, thanks!

PR #22675, opened 2022-07-29
circle CI build for 3.2.0 debian bullseye packages (arm, arm64, amd64)

mgrouch:
It should work on circle ci right away after accepting my changes and merging it into circleci-project-setup
source code is basically deleted from this branch. as it has only .circleci scripts and debian directory
I've commented out line which uploads to cloudsmith as I do not know your cloudsmith layout.
It should build on circle ci in 15-16 minutes

see pipeline results here https://app.circleci.com/pipelines/github/wxWidgets/wxWidgets

Thanks


vadz:
Sorry, but why did you need to delete all sources?

Anyhow, it looks like just creating a PR doesn't trigger a build on Circle CI... I'll try to look at this and find what does, but it will take a bit longer before I find time for this.

mgrouch:
Because for building a release I need to build it from release source zip. 
PR wouldn’t trigger build. Merging and accepting it should.

vadz:
The idea of integrating the build scripts into the repository would be to build the packages from the sources in the repository -- so that we could do it for testing even before the release (and then build the "real" packages using the tag corresponding to the release). Requiring to have the source archives doesn't make sense to me in this context.

mgrouch:
> 

Exactly what I said

I think release should happen when beta packages on all major platforms are tested. For that you need continuous build of those
test packages. 3.3 and 3.2 branch both would benefit from such continuous build and package process.

The cost of finding bugs increases exponentially as you go into next stage of production. So most needs to be weeded out in development before the release.

vadz:
My main problem right now is that, looking at [the docs](https://circleci.com/docs/triggers-overview), there doesn't seem to be any way to only trigger a workflow manually on Circle CI, which would be a prerequisite for adding any workflow building the packages as we definitely don't want to do this on every commit.

The best I found was to add `type: approval` to the job, but I'm afraid that this will result in an accumulation of thousands of unapproved jobs with time, and I'm not sure if it's such a good idea.

I guess we could have a special `build-debs` branch with `.circleci/config.yml` building packages in it and just doing nothing in master and 3.2 (confusingly, the documentation says that nothing will be done for pushes to the branches without this file, but in practice this results in build failures and email notifications to me, that I'd rather not get for every push neither).

.circleci/build-debian.sh
```diff
@@ -0,0 +1,131 @@
...
+#
+# Build for Debian in a docker container
+#
+
+# bailout on errors and echo commands.
+set -xe
+
+DOCKER_SOCK="unix:///var/run/docker.sock"
+
+echo "DOCKER_OPTS=\"-H tcp://127.0.0.1:2375 -H $DOCKER_SOCK -s overlay2\"" | sudo tee /etc/default/docker > /dev/null
```
vadz, line 12: It would be nice to explain what this does because I have no idea.

.circleci/build-debian.sh
```diff
@@ -0,0 +1,131 @@
...
+sudo service docker restart
+sleep 5;
+
+if [ "$EMU" = "on" ]; then
```
vadz, line 16: Do we need this? `EMU` probably stands for "emulation", but why should we emulate anything?

.circleci/build-debian.sh
```diff
@@ -0,0 +1,131 @@
...
+    libcurl4-openssl-dev                   \
+    libgstreamer1.0-dev                    \
+    libgstreamer-plugins-base1.0-dev       \
+    libgstreamer-plugins-bad1.0-dev        \
+    libwebkit2gtk-4.0-dev                  \
+    libnotify-dev                          \
+    wget                                   \
+    doxygen                                \
+    graphviz                               \
+    meson                                  \
```
vadz, line 60: We should need meson for building... nor cmake (installed above), or xsltproc.

It looks like many packages here are not really needed, it would be nice to restrict it to just the required ones, as it will make it simpler to update in the future.

.circleci/config.yml
```diff
@@ -1,26 +1,126 @@
...
+      resource_class: arm.large
+      enabled: true
+      image: ubuntu-2004:202104-01
+commands:
+  git-checkout:
+    steps:
+      - run: mkdir work
+      - run: cd work && wget -nc -O - https://github.com/wxWidgets/wxWidgets/releases/download/v3.2.0/wxWidgets-3.2.0.tar.bz2 > ../wxwidgets3.2_3.2.0+dfsg.orig.tar.bz2
+      - run: cd work && bzip2 < ../wxwidgets3.2_3.2.0+dfsg.orig.tar.bz2 -cd | tar xvf -; shopt -s dotglob; mv wxWidgets-3.2.0/* .; rmdir wxWidgets-3.2.0
+      - run: cd work && wget -nc -O - https://github.com/bareboat-necessities/wxWidgets/archive/master.tar.gz | gzip -cd - | tar xvf - ; mv *-master/debian .; mv *-master/.circleci ..
```
vadz, line 22: This needs to be replaced with an actual checkout step.

.circleci/config.yml
```diff
@@ -1,26 +1,126 @@
...
       - run:
-          name: "Say hello"
-          command: "echo Hello, World!"
-
-# Invoke jobs via workflows
-# See: https://circleci.com/docs/2.0/configuration-reference/#workflows
+          command: |
+            .circleci/build-debian.sh
+          no_output_timeout: 30m
+      - cloudsmith/ensure-api-key
```
vadz, line 31: This will fail for us, won't it?

vadz (changes requested):
Looking at this, we can't really merge this PR in its current state, but perhaps we could reuse parts of it. AFAICS there are several different things here:

1. Ability to build in the given OS/release image: this is nice, although all this docker stuff looks like black magic to me and I'm not sure if we can manage it. I did have to use some unusual commands in `.github/workflows/ci_msw_cross.yml` for running inside a container, but they seem much simpler and are at least somewhat commented. So while I would like to have a possibility to run the build for the given distro easily, it would be great if this could be simplified/streamlined/documented.
2. Downloading the archives sources and building from them: we don't need/want this at all, we should be using the sources from the branch the config file is in.
3. Having Debian files inside the repository: thinking more about this, as long as we're not going to modify them ourselves, we shouldn't have them here but OTOH we could get _them_ from the Debian repository and this would be actually very useful. I.e. instead of keeping the Debian files here and downloading wx sources, we'd keep wx sources here, but download (git-clone, whatever) the Debian files before building. This would be much more useful, I think.
4. Uploading the results to Cloudsmith: we don't need this neither.

It would be great if we could have something working along these lines, what do you think?

mgrouch:
You can use special branch. Have .circleci directory only in it. Manual merge into that branch would be effectively a manual trigger of circleci workflow

.circleci/build-debian.sh
mgrouch: When CircleCi didn’t have arm support I would be building in qemu docker for armhf and arm64. 
currently not used anymore

.circleci/build-debian.sh
mgrouch: Feel free to remove unnecessary

.circleci/config.yml
mgrouch: Yes

mgrouch: With recursive submodules

.circleci/config.yml
mgrouch: You will need cloudsmith or some other place to upload build results. 
so ppl could pick up packages and test

vadz:

Issue #22649, opened 2022-07-19
cmake: OPENGL_LIBRARIES uses undefined library names


0-wiz-0:
The cmake build framework, in `build/cmake/init.cmake`, has:
```
set(OPENGL_LIBRARIES OpenGL::OpenGL OpenGL::GLU OpenGL::EGL)
```

On NetBSD 9.99.98/amd64 with cmake 3.23.2, these library names are undefined and cause build breakage.

I don't know the proper fix for this, but just removing the line makes the build succeed for me, I guess the GL libraries are pulled in via some other library.

**Platform and version information**
 - wxWidgets version you use: 3.2.0
 - wxWidgets port you use: wxGTK
 - OS and its version: NetBSD 9.99.98/amd64
 - Compiler being used: gcc 10.3.0

Journeyman-dev:
I have this exact issue with Fedora. I commented out the line, and now it generates fine.

vadz:
FWIW under Fedora these libraries must exist, you might just not have them on your system.

For NetBSD, do you know what should be the correct names of the libraries be?

MaartenBent:
I can improve the OpenGL logic a bit and only overwrite `OPENGL_LIBRARIES` if the targets exist.

> I guess the GL libraries are pulled in via some other library.

I suspect they are pulled in here as well. `find_package(OpenGL)` also adds libraries to `OPENGL_LIBRARIES`, but we overwrite them with the EGL targets when `OpenGL_EGL_FOUND` is true.

0-wiz-0:
NetBSD uses the standard Mesa Library, so there is no difference to Linux here.
I grepped the tree after running cmake, but `OPENGL_LIBRARIES` is not set for me at all (after commenting out the line), at least i had no matches in any cmake cache.

MaartenBent:
This variable is not cached. The easiest way to check its value is adding `
message("OPENGL_LIBRARIES=${OPENGL_LIBRARIES}")` to `build/cmake/lib/gl/CmakeLists.txt` or directly after `find_package(OpenGL)`.

I tested my proposed solution on Ubuntu and Windows and there it still works fine.  I'll create a PR later today so you can try it on NetBSD and/or Fedora.

MaartenBent:
Please have a look at #22677

0-wiz-0:
Thanks, @MaartenBent, that fixes the problem for me!

Journeyman-dev:
Fixed for me too. Thank you!

PR #22677, opened 2022-07-29
CMake: Improve linking with OpenGL libraries

MaartenBent:
Don't replace `OPENGL_LIBRARIES`, but prepend existing targets too it.
Also add `OPENGL_EGL_INCLUDE_DIRS` to `OPENGL_INCLUDE_DIR`.

See #22649


vadz:
Thanks, merging in master, will also backport to 3.2 later (unless you tell me not to).

PR #22678, opened 2022-07-30
Use lseek64 and similar functions on Linux

gbmaster:
Using old lseek/lstat/etc functions on linux may prevent properly handling of large files

### DIFF
--- a/include/wx/filefn.h
+++ b/include/wx/filefn.h
@@ -364,20 +364,28 @@ enum wxPosixPermissions
     #define   wxClose      close
     #define   wxRead       ::read
     #define   wxWrite      ::write
-    #define   wxLseek      lseek
-    #define   wxSeek       lseek
+    #if defined(linux) || defined(__linux) || defined(__linux__)
+        #define wxHAS_LARGE_FILES
+        #define   wxLseek      lseek64
+        #define   wxSeek       lseek64
+        #define   wxTell(fd)   lseek64(fd, 0, SEEK_CUR)
+        #define   wxStructStat struct stat64
+        #define   wxCRT_Stat   stat64
+        #define   wxCRT_Lstat  lstat64
+    #else
+        #define   wxLseek      lseek
+        #define   wxSeek       lseek
+        #define   wxTell(fd)   lseek(fd, 0, SEEK_CUR)
+        #define   wxStructStat struct stat
+        #define   wxCRT_Stat   stat
+        #define   wxCRT_Lstat  lstat
+    #endif
     #define   wxFsync      fsync
     #define   wxEof        eof
     #define   wxCRT_MkDir      mkdir
     #define   wxCRT_RmDir      rmdir
 
-    #define   wxTell(fd)   lseek(fd, 0, SEEK_CUR)
-
-    #define   wxStructStat struct stat
-
     #define   wxCRT_Open       open
-    #define   wxCRT_Stat       stat
-    #define   wxCRT_Lstat      lstat
     #define   wxCRT_Access     access
     #define   wxCRT_Chmod      chmod
 


vadz:
Things are definitely wrong here, thanks for looking into it. However I think we should reuse the results of the test already [done in configure](https://github.com/wxWidgets/wxWidgets/blob/667c5c843b89ed46e3a64242f94487866cb8c3be/configure.in#L1757-L1764) and check for `HAVE_LARGEFILE_SUPPORT` instead of checking for Linux, could you please try doing it like this?

Ideal would also be to have a unit test for this as it's somewhat embarrassing that it remained broken for so long, but I don't know how could we find a file bigger than 2GiB. We already use `/proc/kcore` [in a test](https://github.com/wxWidgets/wxWidgets/blob/667c5c843b89ed46e3a64242f94487866cb8c3be/tests/filename/filenametest.cpp#L1039), but I don't think we can use it for these functions, as we normally wouldn't have the permissions for opening it. Does anybody see any other file that could be used for this by chance?

gbmaster:
I agree with you
However I'm quite not experienced with the configure/cmake testing code, so it would be better if somebody else could take care of it :(

Sorry

vadz:
Note that you don't need to modify neither configure nor CMake, they both already predefine `HAVE_LARGEFILE_SUPPORT`. Only the header needs to be changed to test for it instead.

lanurmi:
> Ideal would also be to have a unit test for this as it's somewhat embarrassing that it remained broken for so long, but I don't know how could we find a file bigger than 2GiB. Does anybody see any other file that could be used for this by chance?

Would a sparse file work for testing these functions? I.e. created like `dd if=/dev/zero of=reallybigfile bs=1M count=1 seek=10000000` or similar.

jkj:
This is only required for >2GB files on a 32-bit system. On a 64-bit system which is the vast majority of Linux systems these days) LFS is already the default without any flags. `off_t` is a 64-bit by default. Also if you want to use the LFS-specific functions like stat64 and types like `struct stat64` you also need to ensure you compile with `-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE` which I don't think the current check does.
I am highly adept at autoconf and can fix all of this if you think its worth it if you want to support LFS on 32-bit systems. @vadz let me know how you want me to proceed and I will make a PR.

vadz:
Thanks, you're right, of course, that it's not a problem on 64-bit systems but we still support 32-bit ones too and I believe 32-bit ARM systems are still fairly widespread, so it would definitely be nice to fix it for them. TIA!

MaartenBent:
But the current configure already define `_FILE_OFFSET_BITS=64` So `lseek`, `fseeko` and others are already 64-bit. Or am I missing something? Are there still platforms that do not support this (but do support c++11)?

jkj:
@MaartenBent I had come to the same conclusion. If you specifically want to use `lseek64` for example, you would need to add `-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE` to get the declarations and keep the C++ compiler happy, but on all even approximately modern systems (I mean even UnixWare 7 got this right) just using `-D_FILE_OFFSET_BITS=64` should be enough.

vadz:
I'm sorry, it looks like I was completely wrong when I blithely wrote "Things are definitely wrong here". In fact, things seem indeed to be already correct and I guess the problem might have been just not defining `_FILE_OFFSET_BITS=64`?

@gbmaster Could you please clarify what problem exactly did you encounter?

MaartenBent:
In CMake it is indeed not defined, see https://github.com/wxWidgets/wxWidgets/issues/22750.
I have a fix locally already, I'll try to publish a PR soon.

MaartenBent:
I created #23336 that enables `HAVE_LARGEFILE_SUPPORT` in CMake.

vadz:
I think this can be closed now that Maarten's PR was merged, please let me know if I was wrong.

PR #22679, opened 2022-07-31
Fix bitmap returned from wxMSW wxImageList::GetBitmap()

### LINKED ISSUE #22669: Different background of the same wxBitmap in toobar and menu item
Building with 3.2, the same wxBitmap has different background in toobar and menu item. The wxBitmap was created from xpm which set   transparent backgroud.
On menu, it has a black background.
image
On toolbar, it's OK

Nothing changed in app code

vadz:
This fixes #22669 (~~in the first commit; the rest are just some cleanup~~), but I still don't fully understand why, i.e. what was wrong with the original code. It used 32bpp DDB, rather than DIB, but it should still have worked, AFAICS -- except it clearly didn't. @a-wi Would you have any idea by chance?

In any case, the fix should do no harm and fixes the menu item appearance, so I'll apply it if no better solution can be found.

Edit: Remove "cleanup" commits that broke unit tests again because in fact we still can have bitmaps without alpha in the image list.

### DIFF
--- a/src/msw/imaglist.cpp
+++ b/src/msw/imaglist.cpp
@@ -448,7 +448,11 @@ wxBitmap wxImageList::GetBitmap(int index) const
     int bmp_width = 0, bmp_height = 0;
     GetSize(index, bmp_width, bmp_height);
 
-    wxBitmap bitmap(bmp_width, bmp_height);
+    // Specify the depth to force using a DIB for this bitmap as drawing on a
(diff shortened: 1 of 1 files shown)


vadz:
Replaced by a3fd6cf4e6 (Handle non-DIB wxBitmaps correctly in wxMenuItem, 2022-08-01).

Issue #22654, opened 2022-07-19
MSVC C++20 Compile Errors due to `wxSETUPH_PATH_STR`


k3DW:
I built wxWidgets 3.2.0 using `build\msw\wx_vc17.sln` with Visual Studio 17.2.6. I changed the C++ Language Standard to C++20 in all projects for all configurations, otherwise I changed nothing else. The full batch build was successful.

In a new project, with all the header and library directories properly configured, a simple file consisting of just `#include "wx\wx.h"` fails to compile with C++20 but passes with C++17.

```
Warning	C5103	pasting '/' and 'vc_x64_lib' does not result in a valid preprocessing token                         include\msvc\wx\setup.h	142	
Warning	C5103	pasting 'vc_x64_lib' and '/' does not result in a valid preprocessing token                         include\msvc\wx\setup.h	142	
Warning	C5103	pasting '/' and 'msw' does not result in a valid preprocessing token                                include\msvc\wx\setup.h	142	
Warning	C5103	pasting 'mswu' and '/' does not result in a valid preprocessing token                               include\msvc\wx\setup.h	142	
Error	C1083	Cannot open include file: '../../../lib/ vc_x64_lib / mswu /wx/setup.h': No such file or directory  include\msvc\wx\setup.h	142
```

I think this is about stricter preprocessor compliance. A small change in `include\msvc\wx\setup.h` may do the trick.

vadz:
I don't have this MSVS version yet, I'll try to upgrade to it and test/find a fix for this, but this is problematic because we're going to have a problem constructing the include paths if we can't concatenate strings with `/` any more.

k3DW:
I was able to change line 134 from
`wxCONCAT6(../../../lib/, wxLIB_SUBDIR, /, wxTOOLKIT_PREFIX, wxSUFFIX, /wx/setup.h)`
to
`../../../lib/wxLIB_SUBDIR/wxCONCAT(wxTOOLKIT_PREFIX, wxSUFFIX)/wx/setup.h`
and I think it does the same thing.

paulcor:
> I don't have this MSVS version yet

I suspect you may just need to use `-Zc:preprocessor`.

SpareSimian:
I'm using Microsoft Visual Studio Community 2019 Version 16.11.15 and it doesn't happen with my x64 project. (I added foo.cpp with your one-liner to my project and don't get an error.)

k3DW:
[...]

PR #22680, opened 2022-07-31
Don't use token-pasting with non-identifiers in msvc/wx/setup.h

vadz:
This used to work with the non-standard preprocessor in the older MSVC
versions, but doesn't work with the new standard-compliant one in MSVS
2022 (enabled by /Zc:preprocessor).

Luckily, it seems that we don't really need it neither, as simply
expanding the macros in a /-separated strings work with both the new and
the old preprocessors and has the same effect, so just do this instead.

Also simplify the code a little by defining wxTOOLKIT_FULL and
constructing wxSETUPH_PATH once instead of twice.

See #22654.

Co-Authored-By: Vadim Zeitlin <vadim@wxwidgets.org>

### DIFF
--- a/include/msvc/wx/setup.h
+++ b/include/msvc/wx/setup.h
@@ -127,15 +127,15 @@
     #endif
 #endif // wxTOOLKIT_PREFIX
 
+#ifdef wxSUFFIX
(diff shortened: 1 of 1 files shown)


vadz:
Merged by the commit referenced above.

PR #22693, opened 2022-08-01
Support linking with static runtime library

dconnet:
Create a macro to specify the RuntimeLibrary. This allows a user to create wx libs that are statically linking to the VC runtime by specifying the static versions in wx_local.props via
```
    <wxReleaseRuntimeLibrary>MultiThreaded</wxReleaseRuntimeLibrary>
    <wxDebugRuntimeLibrary>MultiThreadedDebug</wxDebugRuntimeLibrary>
```

### DIFF
--- a/build/msw/wx_setup.props
+++ b/build/msw/wx_setup.props
@@ -13,6 +13,13 @@
     </wxCfg>
     <wxVendor>custom</wxVendor>
     <wxRuntimeLibs>dynamic</wxRuntimeLibs>
+    <wxReleaseRuntimeLibrary>MultiThreadedDLL</wxReleaseRuntimeLibrary>
+    <wxDebugRuntimeLibrary>MultiThreadedDebugDLL</wxDebugRuntimeLibrary>
+  </PropertyGroup>
+  <PropertyGroup Label="UserMacros" Condition="'$(wxRuntimeLibs)'=='static'">
+	<!-- Only these need overriding in wx_local.props -->
+    <wxReleaseRuntimeLibrary>MultiThreaded</wxReleaseRuntimeLibrary>
+    <wxDebugRuntimeLibrary>MultiThreadedDebug</wxDebugRuntimeLibrary>
   </PropertyGroup>
   <PropertyGroup Label="UserMacros" Condition="'$(Platform)'=='Win32'">
     <wxArchSuffix>
--- a/build/msw/wx_setup.props
+++ b/build/msw/wx_setup.props
@@ -121,9 +121,6 @@
     <BuildMacro Include="wxIntRootDir">
       <Value>$(wxIntRootDir)</Value>
     </BuildMacro>
-    <BuildMacro Include="wxIntRootDir">
-      <Value>$(wxIntRootDir)</Value>
-    </BuildMacro>
     <BuildMacro Include="wxReleaseRuntimeLibrary">
       <Value>$(wxReleaseRuntimeLibrary)</Value>
     </BuildMacro>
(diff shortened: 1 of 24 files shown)


build/msw/wx_setup.props
```diff
@@ -119,5 +121,14 @@
     <BuildMacro Include="wxIntRootDir">
       <Value>$(wxIntRootDir)</Value>
     </BuildMacro>
+    <BuildMacro Include="wxIntRootDir">
+      <Value>$(wxIntRootDir)</Value>
+    </BuildMacro>
```
vadz, line 126: This seems to have been duplicated accidentally?

build/msw/wx_setup.props
```diff
@@ -12,6 +12,8 @@
     <wxCfg>
     </wxCfg>
     <wxVendor>custom</wxVendor>
+    <wxReleaseRuntimeLibrary>MultiThreadedDLL</wxReleaseRuntimeLibrary>
+    <wxDebugRuntimeLibrary>MultiThreadedDebugDLL</wxDebugRuntimeLibrary>
```
vadz, line 16: Could we perhaps define `wxRuntimeLibs` and check below if it's `dynamic` (default value) or `static` and set these 2 automatically? It looks like it would be more consistent with the make option used currently and simpler to use.

vadz (commented):
Thanks, I barely had time to post my reply to your message on the list when this PR already came in, like by magic.

I'm ready to apply it as is, but I think it could be nicer to have a single `wxRuntimeLibs` property as suggested below, what do you think?

build/msw/wx_setup.props
dconnet: Oops, yes it was...

build/msw/wx_setup.props
dconnet: I think I understand what you mean here. Let me see what I can come up with...

dconnet: Forgot to mention - this is ready now.

build/msw/wx_setup.props
```diff
@@ -119,5 +120,11 @@
     <BuildMacro Include="wxIntRootDir">
       <Value>$(wxIntRootDir)</Value>
     </BuildMacro>
+    <BuildMacro Include="wxReleaseRuntimeLibrary">
+      <Value>$(wxReleaseRuntimeLibrary)</Value>
+    </BuildMacro>
+    <BuildMacro Include="wxDebugRuntimeLibrary">
+      <Value>$(wxDebugRuntimeLibrary)</Value>
+    </BuildMacro>
```
vadz, line 128: This seems to be unused?

vadz (commented):
Thanks, I'm, again, ready to apply this as is, but I'm afraid I've made you make more changes than I wanted to, I just thought that you'd use the test to set the variables in the `.props` file, i.e. still use `$(wx{Debug,Release}RuntimeLibrary)` as you did in the first version, but set them depending on the `$(wxRuntimeLibs)` value.

I am not sure if this is better or worse than the current version, but if we don't use `$(wx{Debug,Release}RuntimeLibrary)` we shouldn't define them -- and if we do, we shouldn't hardcode `MultiThreaded{Debug,Release}{,DLL}` names.

Or am I missing something here?

build/msw/wx_setup.props
dconnet: Ah, I understand better! I'll change to that (which will make these used)

dconnet: Modified to use wxReleaseRuntimeLibrary/wxDebugRuntimeLibrary in vcxproj files (the original way), plus use wxRuntimeLibs in wx_setup to set initial values. Default is /MD. Setting wxRuntimeLibs to 'static' will use /MT (any other value will leave the setting at /MD)

vadz:
Perfect, thanks! Will merge soon.

PR #22697, opened 2022-08-03
Fix warnings in gcc 12 optimized build

vadz:
This should be backported to 3.2 too.

### DIFF
--- a/include/wx/vector.h
+++ b/include/wx/vector.h
@@ -519,6 +519,13 @@ public:
         const size_t idx = it - begin();
         const size_t after = end() - it;
 
+        // Unfortunately gcc 12 still complains about use-after-free even if
+        // our code is corrects because it actually optimizes it to be wrong,
+        // with -O2 or higher, so use this hack to avoid the warning with it.
+#if wxCHECK_GCC_VERSION(12, 1)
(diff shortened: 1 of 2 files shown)


include/wx/vector.h
```diff
@@ -519,6 +519,13 @@ class wxVector
         const size_t idx = it - begin();
         const size_t after = end() - it;
 
+        // Unfortunately gcc 12 still complains about use-after-free even if
```
PBfordev, line 522:
```suggestion
        // Unfortunately gcc 12 still complains about use-after-free even though
```

include/wx/vector.h
```diff
@@ -519,6 +519,13 @@ class wxVector
...
+        // our code is corrects because it actually optimizes it to be wrong,
```
PBfordev, line 523:
```suggestion
        // our code is correct because it actually optimizes it to be wrong,
```

include/wx/vector.h
vadz: Thanks, will apply this (and the other typo fix) when merging.

vadz:
Merged as cc0c55af7bd56b4512ec88d895bba8dedf8c983c

PR #22703, opened 2022-08-04
Fix lack of key events for wxSearchCtrl under Mac

vadz:
The real fix is in the first commit, but the second one makes the code a bit more palatable.

I think both commits should be safe to backport to 3.2, but if there is any doubt about the latter one, we could backport just the first one, as it's enough to fix the user-visible bug.

### DIFF
--- a/include/wx/osx/cocoa/private.h
+++ b/include/wx/osx/cocoa/private.h
@@ -449,6 +449,17 @@ public:
 
     @end
 
+    @interface wxNSSearchField : NSSearchField
+    {
+        wxNSTextFieldEditor* fieldEditor;
+        BOOL m_withinTextDidChange;
+    }
+
+    - (wxNSTextFieldEditor*) fieldEditor;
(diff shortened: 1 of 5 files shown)


csomor (approved):
looks fine to me, thank you, the only thing that slightly disturbs my eye is a uppercase property name, do we have any 'standard' way starting with a lowercase w... ?

vadz:
> looks fine to me, thank you, the only thing that slightly disturbs my eye is a uppercase property name, do we have any 'standard' way starting with a lowercase w... ?

This is what I had done initially, but the auto-generated setter name capitalizes the first letter automatically resulting in an even uglier `setWxFieldEditor:`, so I've done it like this instead. Another possibility would be `WX_fieldEditor`, I guess, would you prefer this?

But AFAICS we have to use a wx-specific name here (i.e. some form of "wx" prefix) to avoid conflicts with other classes defining their own properties with the same name, but used for something completely different.

PR #22704, opened 2022-08-04
Update "hr.po" file

milotype:
- update Croatian translation with current .pot file
- some corrections in Croatian translation
- removed orphanded/old strings


vadz:
Thanks for the update, will commit and push soon!

Issue #22696, opened 2022-08-03
CMake uses C++-specific options for C files too


vadz:
Doing

    $ cmake -S . -B ~/build/wx/cmake -GNinja
    $ cmake --build ~/build/wx/cmake

results in the following output
```
[2/794] Building C object libs/base/CMakeFiles/wxbase.dir/__/__/__/__/src/common/extended.c.o
cc1: warning: command-line option ‘-Wno-ctor-dtor-privacy’ is valid for C++/ObjC++ but not for C
cc1: warning: command-line option ‘-Woverloaded-virtual’ is valid for C++/ObjC++ but not for C
[5/794] Building C object libs/core/CMakeFiles/wxcore.dir/__/__/__/__/src/gtk/treeentry_gtk.c.o
cc1: warning: command-line option ‘-Wno-ctor-dtor-privacy’ is valid for C++/ObjC++ but not for C
cc1: warning: command-line option ‘-Woverloaded-virtual’ is valid for C++/ObjC++ but not for C
```

It's harmless, but it would be nice if we could avoid these warnings by using separate C and C++ warning options.

MaartenBent:
This can be fixed by adding those warnings only for CXX compiler, like: `$<$<COMPILE_LANGUAGE:CXX>:${cpp_only_warnings}>`

I'll create a PR with this.

vadz:
I think this should be safe to backport to 3.2, shouldn't it?

MaartenBent:
Yes, this can be backported.

vadz:
Thanks, I've actually already done it and forgot to close this (it's really inconvenient not to be able to use `Closes #NNNN` in 3.2 branch).

Issue #22646, opened 2022-07-19
cmake: Unix joystick code is Linux-specific


0-wiz-0:
On NetBSD, cmake by default tries to build the Unix joystick backend, but that uses `linux/joystick.h`, which only exists on Linux.

I use the workaround of `-DwxUSE_JOYSTICK:BOOL=OFF` but it would be good if cmake checked for linux/joystick.h before enabling it.

**Platform and version information**
 - wxWidgets version you use: 3.2.0
 - wxWidgets port you use: wxGTK
 - OS and its version: NetBSD 9.99.98/amd64
 - Compiler being used: gcc 10.3.0
 - Build framework: cmake

vadz:
I don't know how to test for the header existence using CMake. It looks like we should be using `check_include_files()` but I'm not sure how to use the variable it sets later and the only example of its use in our sources doesn't show it, so maybe it's not even the right function to use?

@MaartenBent could you please advise?

MaartenBent:
Something like this, inside the `if(UNIX)` case, would probably work:
```cmake
check_include_files("linux/joystick.h" HAVE_JOYSTICK_H)
if(NOT HAVE_JOYSTICK_H)
     message(WARNING "wxJoystick is not available")
    wx_option_force_value(wxUSE_JOYSTICK OFF)
endif()
```

I can test it and create a PR sometime next week probably.

PR #22705, opened 2022-08-04
CMake: Add joystick check and fix warning

MaartenBent:
Disable `wxUSE_JOYSTICK` if `linux/joystick.h` is not found on Linux systems. See https://github.com/wxWidgets/wxWidgets/issues/22646.

Don't use `-Wno-ctor-dtor-privacy` and `-Woverloaded-virtual` for C files, to fix 'is valid for C++/ObjC++ but not for C' warning. See https://github.com/wxWidgets/wxWidgets/issues/22696.

Also applicable to 3.2 branch.


build/cmake/functions.cmake
```diff
@@ -111,41 +111,45 @@ function(wx_set_common_target_properties target_name)
...
-            -Wno-ignored-attributes
+        endif()
+
+        if("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+            list(APPEND common_gcc_clang_compile_options
+                -Wno-ignored-attributes
             )
+        endif()
+
+        target_compile_options(${target_name} PRIVATE
```
vadz, line 149: Apparently there is an extra space here
```suggestion
        target_compile_options(${target_name} PRIVATE
```

build/cmake/setup.cmake
```diff
@@ -442,6 +442,14 @@ if(UNIX)
         check_symbol_exists(inet_aton arpa/inet.h HAVE_INET_ATON)
         check_symbol_exists(inet_addr arpa/inet.h HAVE_INET_ADDR)
     endif(wxUSE_SOCKETS)
+
+    if(NOT APPLE)
```
vadz, line 446: Can we restrict this to Linux? Checking for `linux/foo` under non-Linux systems doesn't seem to make sense...

vadz (approved):
Thanks for fixing this!

I'll merge it with just the space fix unless you decide to change the platform test before the joystick check.

build/cmake/setup.cmake
MaartenBent: It is. This is all within `if(UNIX)`.

vadz: Sorry, but just to be clear, I meant only _Linux_, not any (other) Unix.

MaartenBent: I don't think CMake has any specific variables for that.
And `unix/joystick.cpp` (that has `#include <linux/joystick.h>` is added inside an `if(UNIX)` check too (via `wx_append_sources(CORE_SRC XWIN_LOWLEVEL)`). But it is also inside `if(WXGTK)`, so I can add that here too.

vadz: There is `CMAKE_SYSTEM_NAME` which should be `Linux` under Linux, but, well, it's not that important.

MaartenBent: If we assume it is available on Linux only, then something like this could also work:
```cmake
    if(WXGTK AND NOT APPLE AND NOT CMAKE_SYSTEM_NAME STREQUAL "Linux")
        message(WARNING "wxJoystick is not available")
        wx_option_force_value(wxUSE_JOYSTICK OFF)
    endif()
```

But I find checking if the header exists more robust.

vadz: It's really not very important, but, as it is right now, you'll be getting this warning message during every cmake configure run under non-Linux systems (e.g. FreeBSD tested by our CI now) and there is nothing you can do to get rid of it, so I thought it would be better to not test for this (Linux-specific) header at all and not warn about its absence neither.

Granted, this also happens when using configure but nobody pays any attention to configure warnings anyhow :-/ and CMake output is a bit tidier, so the warning is more noticeable there.

MaartenBent: Maybe a better solution is then to disable `wxUSE_JOYSTICK` by default on non-linux systems. And if they enable it manually, they get the warning.

vadz:
Thanks, I agree that this is the best solution but I have one last question about this: why do we test for wxGTK here? It looks like joystick support is the same for it, wxX11 and wxQt under Unix.

MaartenBent:
That is indeed wrong. I checked where `unix/joystick.cpp` was included, and that is only inside an `if(WXGTK)` check. But only because CMake does not support wxX11 or motif. wxQt doesn't seem to use it. I'll remove the check.

Issue #22702, opened 2022-08-04
wxEVT_KILL_FOCUS not generated for wxSearchCtrl under Mac


vadz:
#### 2021-12-13 22:56:18: stuart.allie (Stuart Allie) commented

___
I can add that wxEVT_KILL_FOCUS is also not (reliably) sent for wxSearchCtrl on the Mac. It is sent sometimes (eg. changing to a new frame) but not when you simply click away from the wxSearchCtrl. wxEVT_SET_FOCUS seems to work as expected.

_Originally posted by @wxtrac in https://github.com/wxWidgets/wxWidgets/issues/19321#issuecomment-1011322552_

sp00ck:
I have trouble with wxListCtrl
I can click outside and deselect cell, how prevent deselect cell?

before
![Przechwycenie obrazu ekranu_2022-08-06_16-57-48](https://user-images.githubusercontent.com/25770302/183255620-edf23e41-5cda-4417-908c-413de0f2fd45.png)

after click
![Przechwycenie obrazu ekranu_2022-08-06_16-58-44](https://user-images.githubusercontent.com/25770302/183255630-f5bc5639-ab14-44e8-8ef1-71c313deb34d.png)

vadz:
@sp00ck What does this have to do with this issue? Please open new issues for unrelated problems (but in this particular case maybe better start a discussion somewhere because I don't even see how is this an issue).

sp00ck:
this is dependent about event kill focus

vadz:
There is no `wxSearchCtrl` in your screenshot, so it's very clearly a different issue. Please open a new issue and please actually explain what the problem is, including how to reproduce it.

PR #22706, opened 2022-08-04
Generate wxEVT_KILL_FOCUS for wxSearchCtrl under Mac

vadz:
Do it using controlTextDidChange: notification as it's already done for
wxTextCtrl and wxComboBox.

See #22702.

---

It seems like there should be some way to avoid duplicating this code in several places too, but I'm not sure how exactly to do it, so for now let's at least fix the bug.

### DIFF
--- a/src/osx/cocoa/srchctrl.mm
+++ b/src/osx/cocoa/srchctrl.mm
@@ -68,6 +68,14 @@
         impl->controlTextDidChange();
 }
 
+- (void)controlTextDidEndEditing:(NSNotification *) aNotification
+{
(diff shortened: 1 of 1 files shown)


csomor:
> It seems like there should be some way to avoid duplicating this code in several places too, but I'm not sure how exactly to do it, so for now let's at least fix the bug.

some versions unfortunately are a little bit more complicated, but we could have some static methods that we send the self and the notifications over and pack these somewhere  ?

csomor (approved):
exactly, yes, same procedure ...

PR #22707, opened 2022-08-04
Use the appropriate bezel style for wxToggleButton under Mac

### LINKED ISSUE #22140: macOS: wxToggleButton style is unexpected
**Describe the bug**
When creating a wxToggleButton on macOS, the style is standard push button style. This isn't an appropriate use of the push button style on macOS. That style is used for a momentary push button, not a toggle button. A more appropriate use would be square style with on/off type which would also match the wx documentation.

Using standard push button style for a toggle button is unexpected and not following the Mac human interface guidelines.

Create an wxToggleButton on macOS and observe.

A patch to create a more expected toggle button in Cocoa would be to use these flags on the backing button (written in obj-c):

 [_button setButtonType:NSButtonTypePushOnPushOff];
 [_button setBezelStyle:NSBezelStyleShadowlessSquare];

 - wxWidgets version you use: 3.1.5
 - wxWidgets port you use: wxOSX
 - OS and its version: macOS 11.6.4 (20G417)

Note: I do apologize for all the recent macOS bugs; I hope to have some time to submit some PRs to help improve the project! Keep up the great work!

vadz:
It should use NSShadowlessSquareBezelStyle rather than
NSRoundedBezelStyle that it was given by default by the function
SetToggleButtonBezelStyle() used for determining normal push button
bezels.

As a side effect, this function doesn't need to be extern any more and
can be made static.

Closes #22140.

---

I've rebased my old commit on the latest master, as I'd like to either merge it or delete this branch. Stefan, do you think it's an improvement (please see the original issue for the discussion)?

### DIFF
--- a/src/osx/cocoa/button.mm
+++ b/src/osx/cocoa/button.mm
@@ -176,9 +176,7 @@ NSButton *wxButtonCocoaImpl::GetNSButton() const
 // Set bezel style depending on the wxBORDER_XXX flags specified by the style
 // and also accounting for the label (bezels are different for multiline
 // buttons and normal ones) and the ID (special bezel is used for help button).
-//
-// This is extern because it's also used in src/osx/cocoa/tglbtn.mm.
-extern "C"
+static
 void
 SetBezelStyleFromBorderFlags(NSButton *v,
                              long style,
--- a/src/osx/cocoa/tglbtn.mm
+++ b/src/osx/cocoa/tglbtn.mm
@@ -27,8 +27,10 @@
 
 static
 void
-SetToggleButtonBezelStyle(NSButton *v, long style)
+SetToggleButtonStyle(NSButton *v, long style)
 {
+    NSButtonType type = NSOnOffButton;
+
     // This is the appropriate default bezel style for the toggle buttons.
     NSBezelStyle bezel = NSShadowlessSquareBezelStyle;
 
@@ -39,6 +41,7 @@ SetToggleButtonBezelStyle(NSButton *v, long style)
 
         case wxBORDER_NONE:
             [v setBordered:NO];
+            type = NSToggleButton;
             break;
 
         case wxBORDER_SIMPLE:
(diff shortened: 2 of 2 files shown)


csomor (approved):
thanks, I'm fine with it, I guess one reason Apple never updated the style is the fact that this kind is really seldom used nowadays, switches or checkboxes being preferred

vadz:
Yes, I couldn't find a toggle button anywhere to check how it was supposed to look like...

vadz:
Sorry, I forgot that there was another improvement to be done here, thanks to @willco007 for suggesting this one too.

@csomor Could you please look at the latest commit too and tell me if you see anything wrong with doing this? As the commit message says, this allows to visually distinguish the states of toggle buttons without border.

csomor:
@vadz I'll test, I think it's fine, only the naming is too new: for macOS < 10.12 we should use [NSToggleButton](https://developer.apple.com/documentation/appkit/nstogglebutton)

vadz:
Thanks, I'll change it, but this means that none of our CI builds uses macOS 11 SDK, as otherwise it wouldn't have passed. Would it be possible to change at least one of the builds to do it, i.e. can we (still) install this SDK? The machine used by the hosted runner seems to have 12.3 only...

csomor:
> Thanks, I'll change it, but this means that none of our CI builds uses macOS 11 SDK, as otherwise it wouldn't have passed. Would it be possible to change at least one of the builds to do it, i.e. can we (still) install this SDK? The machine used by the hosted runner seems to have 12.3 only...

oh the brave new world of numbering macOS versions ....

we support building against macOS SDK 10.11 (with minimum deployment 10.10) , then came 10.12, 10.13, 10.14, 10.15, only then macOS 11 (Big Sur) and macOS 12 (Monterey), so while we have macos-11 runners which cover macOS SDK 11 we don't have any covering 10.11 .. and our CI mac is on macos 12, covering macOS SDK 12 and we might add an xcode beta for macOS 13 (Ventura)

vadz:
Actually I think my problem here is that my code using the newer constant somehow managed to compile (or did it? it looks like the CI builds are not running again any more...) with the default 10.11 min macOS version. Shouldn't even the latest SDK still give errors when `-mmacosx-version-min=10.11` is used, as it should be in our builds?

PR #22713, opened 2022-08-07
Draft with questions: Add/extend editor folders/files to git ignore list

Randalphwa:
If a user clones or forks the wxWidgets repository, and wants to use a popular editor like Visual Studio Code, they will need to know enough about git to add the editor's required folder name and/or extensions to the repositories `.git/info/exclude` file. While it's unreasonable to add every editor out there to the ignore list, this PR adds support for Eclipse, Sublime, and VSCode.

Microsoft maintains a template for Visual Studio ignores. A lot of these don't apply, and some of them already exist in the various build directories. However, those don't necessarily cover firing up Visual Studio directly in a Sample directory or some of the other directories. I added the standard user-specific file extensions to ignore that Microsoft recommends as well as the .vs/ cache/options directory. These are at the root, so they apply to any directory. If that seems a reasonable approach, then I can remove the duplicates in some of the individual build paths (e.g., remove `/build/msw/*.suo` and `/tests/test.suo`).

While not as popular as VSCode, two other IDEs that might be useful to add are CodeBlock and CodeLite -- this would entail adding:

```
    # CodeBlock files
    *.cbp
    *.depend
    *.layout

    # CodeLite files
    .codelite/
    *.project
    *.workspace
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -421,3 +421,26 @@
 /utils/wxrc/wxrc_vc[789].sln
 
 /3rdparty/webview2
+
+# Eclipse project files
+.project
+.cproject
+
+# SublimeText project files
+*.sublime-project
+*.sublime-workspace
+
+# VSCode files
+.vscode/
+*.code-workspace
+*.code-search
+
+# Visual Studio User-specific files
+*.rsuser
+*.suo
+*.user
+*.userosscache
+*.sln.docstates
+
+# Visual Studio cache/options directory
+.vs/


vadz:
I have traditionally resisted adding such files to our `.gitignore` because it seems to me that they should go into the global per-user `.gitignore` instead. This has 2 benefits:

1. People using C::B (or CodeLite or whatever) can add their extension to this file once instead of relying on each and every project they use to do it.
2. We don't have to maintain the list of popular IDEs (indirectly, via their associated extensions) or, worse, have to determine whether an IDE is popular enough or not to include.

But well, clearly, people don't see (1) as a benefit for some reason, otherwise they would just do it. And (2) is, admittedly, not a huge deal. So if people really, really, really want to have this in wx repository, I'm ready to give up (all while still thinking that it's not the best approach).

Anyone else?

tierra:
> I have traditionally resisted adding such files to our `.gitignore` because it seems to me that they should go into the global per-user `.gitignore` instead.

This has been my position as well.

Randalphwa:
I don't add .vscode/ to my global .gitignore because I _want_ to track it in my private repositories, and I need to ensure it's in the project's .gitignore for any of my public repositories. I always use the local exclude file for forks/cloned repositories, and quite honestly I'd forgotten the global .gitignore file even existed (and initially, why I don't use it).

I had not considered your second point about having to decide down the road whether or not another editor is popular enough to include, and I would agree that it's not a Pandora's box worth opening. I'd still be inclined to add .vscode/ both because it's arguably the most popular IDE and because Microsoft seems to treat it as an alternative to Visual Studio (with the bonus that it's cross platform). However, both the global .gitignore and local exclude files work for those who know about them, so I'm fine leaving the current .gitignore as is.

vadz:
Let's not do anything here if there are no other proponents of adding this.

PR #22719, opened 2022-08-14
More proposed 3.2 backports

vadz:


vadz:
Oops, I've accidentally pushed an updated version of this PR into 3.2 instead of 3.2-backports :-(

I'm closing it now as it doesn't make much sense to keep it open, but please let me know if anybody has any objections to any of the commits backported to 3.2 in d013367c24..467d96c03d range.

PR #22723, opened 2022-08-15
Fix double negatives used with the word 'neither' in docs and comments

lanurmi:
In many cases it should be 'either', and 'nor' should be 'or' accordingly.

No changes to actual code.

### DIFF
--- a/docs/doxygen/mainpages/const_cpp.h
+++ b/docs/doxygen/mainpages/const_cpp.h
@@ -381,7 +381,7 @@ more details.
         if this macro is defined, the compilation options
         don't include compiler flags needed for multithreaded code generation. This
         implies that wxUSE_THREADS is 0 and also that other (non-wx-based) threading
-        packages cannot be used neither.}
+        packages cannot be used either.}
 @itemdef{wxNO_UNSAFE_WXSTRING_CONV,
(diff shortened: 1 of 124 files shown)


tests/test.cpp
```diff
@@ -404,7 +404,7 @@ static int TestCrtReportHook(int reportType, char *message, int *)
 int main(int argc, char **argv)
 {
     // tests can be ran non-interactively so make sure we don't show any assert
-    // dialog boxes -- neither our own nor from MSVC debug CRT -- which would
+    // dialog boxes -- either our own nor from MSVC debug CRT -- which would
```
vadz, line 407: I'm almost sure this is incorrect, "either ... nor" doesn't really work, does it?

vadz (commented):
Most of the changes here are definitely correct, thanks, but while I didn't have time to look at all of them, randomly selecting the first and the last one I'm already not sure about the latter, so it would be really great if somebody (preferably, but not necessarily, a native English speaker) could go over all the changes and double check them. TIA!

tests/test.cpp
swt2c: Yes, this one is correct as-is.

lanurmi: Agreed, there was an error here, and the change has been undone. Please pick another two random changes.

Issue #9264, opened 2004-06-09
wxColourEnumerator


wxtrac:
*Issue migrated from trac ticket # 9264*

**priority:** low

#### 2004-06-09 20:41:52: wyo created the issue

___
Currently it's not possible to retrieve the colours
from the wxColourDatabase. It might be a good idea to
have a wxColourEnumerator class quiet similar to the
wxFontEnumerator class.

vadz:
As discussed in [this thread](https://groups.google.com/g/wx-dev/c/nPhYYOLOL9A), it would be useful to add `GetAllNames()` method and this can be done much simpler than adding a whole new class.

Also, this is not related to `wxFontEnumerator` at all.

Issue #22721, opened 2022-08-15
White button text with dark button becomes unreadable when button is focused (low contrast)


SpareSimian:
Create a button with white text on blue background like this:
```
   m_buttonTest = new wxButton( this, wxID_ANY, wxT("Test\nComms"), wxDefaultPosition, wxSize( 100,100 ), 0 );
   m_buttonTest->SetBackgroundColour(_T("BLUE"));
   m_buttonTest->SetForegroundColour(_T("WHITE"));
```
In wx 3.2 on Win10, mouse over the button, and the blue turns very light blue, almost white. The white text doesn't change color. The result is unreadable. 

Compare to hovering the mouse over the white-on-green "New Issue" button on GitHub.

ardovm:
This issue should be fixed by commit f3ab568

vadz:
Indeed, thanks.

PR #22724, opened 2022-08-16
Issue 9264: Add wxColourDatabase::GetAllNames

SpareSimian:
Includes a VS 2022 sample (based on minimal) to illustrate the usage. Completes issue #9264.


interface/wx/gdicmn.h
```diff
@@ -903,6 +903,11 @@ class wxColourDatabase
         colour is not found in the database.
     */
     wxString FindName(const wxColour& colour) const;
+
+    /**
+        List all known colours by name.
+    */
```
vadz, line 909:
```suggestion
        @since 3.3.0
    */
```

(unless you plan to backport it to 3.2.1, in which case that version should be used).

src/common/gdicmn.cpp
```diff
@@ -455,6 +455,20 @@ wxString wxColourDatabase::FindName(const wxColour& colour) const
     return wxEmptyString;
 }
 
+wxVector<wxString> wxColourDatabase::GetAllNames() const
+{
+   wxColourDatabase * const self = wxConstCast(this, wxColourDatabase);
+   self->Initialize();
+
+   wxVector<wxString> names;
```
vadz, line 463: It won't change the world, but it doesn't cost anything to avoid reallocations by doing
```suggestion
   wxVector<wxString> names;
   names.reserve(m_map->size());
```
neither, so I'd add it here.

vadz (commented):
Thanks, it's indeed nice and simple, but could you please move the code demonstrating it to the drawing sample? It already shows wxSystemSettings colours, so it seems logical to show these ones in it as well.

TIA!

interface/wx/gdicmn.h
SpareSimian: Done. We can change it if/when we backport it.

SpareSimian: You might want to grab a copy of the proposed sample as it illustrates the other recent issue #22721 of buttons with light colours appearing washed out when hoving the mouse over them. I factored the DrawSystemColours method to use the same setup and loop to draw the DB colours.

vadz (approved):
Great, thanks!

I'll merge it soon (but I won't include MSVS files, we'll have to somehow generate them later...).

PR #22726, opened 2022-08-17
CI: fix/simplify httpbin use by standardizing Python version

swt2c:


build/tools/httpbin.sh
```diff
@@ -6,72 +6,9 @@ httpbin_launch() {
...
-            ;;
-
-        *)
-            python3 -m pip install --user --upgrade pip==19.1.1
-            python3 -m pip install --user wheel
-    esac
-
-    echo "Installing using `python3 -m pip --version`"
-
+    pip_explicit_deps='Flask==2.0.3 Werkzeug==2.0.3'
```
vadz, line 11: Do we still need the explicit versions here or could we possibly relax them? If we do, I'd like to add back a comment about it, even if it's just something along the lines of

     // Use versions known to work Python 3.10.

vadz (commented):
Thanks, I agree that this is the best solution for GitHub, but I'm a bit uncomfortable about it because this is also used by the other CIs. I guess it's unlikely we're going to run into any more platforms without working setuptools/pip in the future (this was only needed by Travis AFAIR), but perhaps we should at least test Python version and give a warning if it's out of supported range (3.6..3.10 right now I would guess?)?

In any case, thanks a lot for simplifying it!

build/tools/httpbin.sh
swt2c: It does still need to be limited to those versions (well, basically < 2.1.0).  It's not anything to do with Python versions, but it's all really because httpbin is no longer maintained and is not compatible with newer Flask and Werkzeug.

swt2c:
> Thanks, I agree that this is the best solution for GitHub, but I'm a bit uncomfortable about it because this is also used by the other CIs. I guess it's unlikely we're going to run into any more platforms without working setuptools/pip in the future (this was only needed by Travis AFAIR), but perhaps we should at least test Python version and give a warning if it's out of supported range (3.6..3.10 right now I would guess?)?
> 
> In any case, thanks a lot for simplifying it!

What other CIs use this?  I did a bunch of grepping and I could only find it being used by GitHub.  Did I miss something?

vadz:
Oops, sorry, it was a case of wishful thinking (I did want to use it for both Cirrus and Circle, but never actually found time to do it) and confusion (I forgot that this file was separate from `build/tools/before_install.sh`). So you're right, it is not used by anything, sorry again.

oneeyeman1:
@swt2c ,
Thank you for fixing the build.
Hopefully it can be applied relatively soon and Vadim can push my change as well.

PR #22727, opened 2022-08-18
Fix script parsing in wxLocaleIdent

utelle:
The script part of a locale tag was incorrectly parsed due to a wrong index in the calls to `wxString::Mid`. This resulted in shortened script tags that couldn't be interpreted anymore.

### DIFF
--- a/src/common/uilocale.cpp
+++ b/src/common/uilocale.cpp
@@ -215,7 +215,7 @@ wxLocaleIdent wxLocaleIdent::FromTag(const wxString& tag)
 
         case 4:
             // Must be an ISO 15924 script.
-            locId.m_script = (*it).Left(1).Upper() + (*it).Mid(2).Lower();
+            locId.m_script = (*it).Left(1).Upper() + (*it).Mid(1).Lower();
             break;
 
         default:
@@ -299,7 +299,7 @@ wxLocaleIdent& wxLocaleIdent::Script(const wxString& script)
         script.find_first_not_of(validCharsAlpha) == wxString::npos)
     {
         // Capitalize first character
-        m_script = script.Left(1).Upper() + script.Mid(2).Lower();
+        m_script = script.Left(1).Upper() + script.Mid(1).Lower();
     }
     else if (!script.empty())
     {
--- a/src/common/uilocale.cpp
+++ b/src/common/uilocale.cpp
@@ -215,7 +215,7 @@ wxLocaleIdent wxLocaleIdent::FromTag(const wxString& tag)
 
         case 4:
             // Must be an ISO 15924 script.
-            locId.m_script = (*it).Left(1).Upper() + (*it).Mid(1).Lower();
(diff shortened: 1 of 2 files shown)


src/common/uilocale.cpp
```diff
@@ -215,7 +215,7 @@ wxLocaleIdent wxLocaleIdent::FromTag(const wxString& tag)
 
         case 4:
             // Must be an ISO 15924 script.
-            locId.m_script = (*it).Left(1).Upper() + (*it).Mid(2).Lower();
+            locId.m_script = (*it).Left(1).Upper() + (*it).Mid(1).Lower();
```
vadz, line 218: I didn't even notice it originally, but IMO this should be
```suggestion
            locId.m_script = it->Capitalize();
```
which is much more readable (and the same below, of course).

I'm afraid of making this change myself because, as trivial as it seems, I still don't want to change something without testing it, and this code is obviously not covered by any unit tests (could one be added?), but as you've already tested it, could I please ask you to do it?

TIA!

utelle: > I didn't even notice it originally, but IMO this should be [...] `Capitalize()_`[...] which is much more readable (and the same below, of course).

For sure that is a much more elegant solution. Sorry that I missed the availabilty of method `wxString::Capitalize()`. Its use would have prevented that bug in the first place. I applied and committed this change.

> I'm afraid of making this change myself because, as trivial as it seems, I still don't want to change something without testing it,

_"The world belongs to the brave."_ 😃 

> and this code is obviously not covered by any unit tests (could one be added?),

Yes, this use case was not covered, but should, of course.

> but as you've already tested it, could I please ask you to do it?

I extended the `wxUILocale::FindLanguageInfo` test to cover this use case.

vadz: Great, thanks a lot! I'll merge this and backport to 3.2 when the CI builds finish.

P.S. In response to "The world belongs to the brave.", I also know my proverbs and remember the old common folk wisdom about "break the build once, twice shy" (I have no idea why do I feel that it applies to me, knowing that I broke the build dozens if not hundreds of times and not once, but here you are).

utelle: > Great, thanks a lot!

You are welcome. This is the least I could do. 

> I'll merge this and backport to 3.2 when the CI builds finish.

Thanks. Although locale tags with script parts are not very common, fortunately, it is nevertheless important to backport the fix.

> P.S. In response to "The world belongs to the brave.", I also know my proverbs and remember the old common folk wisdom about "break the build once, twice shy" (I have no idea why do I feel that it applies to me, knowing that I broke the build dozens if not hundreds of times and not once, but here you are).

Oh, I know that from own experience only too well - _"No matter how small the change, for sure an error will be in there."_

oneeyeman1:
Guys,
Remember - even "Hello World" have bugs... ;-)

PR #22730, opened 2022-08-18
Handle Aero Snapped window position in wxPersistentRegisterAndRestore

Randalphwa:
If the app is using wxPersistentRegisterAndRestore, this will correctly save the
current window position if the user closes the top level window after using
an Aero Snap function without restoring the window before closing it.

The wxTLWGeometry class uses ::GetWindowPlacement() and ::SetWindowPlacement() to save/restore a top level window's position. The problem is that ::GetWindowPlacement() returns an incorrect position when the last sizing of the window was an Aero Snap. If the window is not maximized or iconized, then the correct position can be retrieved by a call to ::GetWindowRect(). However this is a different coordinate system (screen coordinates versus work area coordinates), so the rectangle retrieved can _not_ be used directly by ::SetWindowPlacement(). There is no Windows API for converting between the two coordinate systems.

The new code checks to see if the width or height of the rectangle returned by ::GetWindowRect() is the same as that returned by ::GetWindowPlacement(). If it's not the same, then the rectangle retrieved by ::GetWindowPlacement() is updated with the Aero Snapped dimensions.

### DIFF
--- a/include/wx/msw/private/tlwgeom.h
+++ b/include/wx/msw/private/tlwgeom.h
@@ -113,6 +113,38 @@ public:
             return false;
         }
 
+        RECT rect;
+        ::GetWindowRect(tlw->GetHWND(), &rect);
+        // Height and width should be the same unless the user performed
+        // an Aero Snap operation.
+        if ((rect.bottom - rect.top) !=
+            (m_placement.rcNormalPosition.bottom - m_placement.rcNormalPosition.top)
+            ||
+            (rect.right - rect.left) !=
+            (m_placement.rcNormalPosition.right - m_placement.rcNormalPosition.left))
+        {
+            MONITORINFO mi;
+            mi.cbSize = sizeof(mi);
+            if (!::GetMonitorInfo(::MonitorFromWindow(tlw->GetHWND(), 
+                MONITOR_DEFAULTTONEAREST), &mi))
+            {
+                wxLogLastError("GetMonitorInfo");
+                return false;
+            }
+
+            // If the tray is on the top or the left, then the rectangle needs to
+            // be adjusted to match what ::SetWindowPlacement expects.
+            if (mi.rcMonitor.top < mi.rcWork.top || 
+                mi.rcMonitor.left < mi.rcWork.left)
+            {
+                // Negative offset to eliminate the tray width/height.
+                OffsetRect(&rect, (mi.rcMonitor.left - mi.rcWork.left),
+                     (mi.rcMonitor.top - mi.rcWork.top));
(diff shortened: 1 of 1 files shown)


include/wx/msw/private/tlwgeom.h
```diff
@@ -113,6 +113,41 @@ class wxTLWGeometry : public wxTLWGeometryBase
...
+            // Height and width should be the same unless the user performed
+            // an Aero Snap operation.
+            if ((rect.bottom - rect.top) !=
+                (m_placement.rcNormalPosition.bottom - m_placement.rcNormalPosition.top)
+                ||
+                (rect.right - rect.left) !=
+                (m_placement.rcNormalPosition.right - m_placement.rcNormalPosition.left))
+            {
+                MONITORINFO mi;
+                mi.cbSize = sizeof(mi);
```
vadz, line 129: A very minor improvement:
```suggestion
                WinStruct<MONITORINFO> mi;
```

include/wx/msw/private/tlwgeom.h
```diff
@@ -113,6 +113,41 @@ class wxTLWGeometry : public wxTLWGeometryBase
...
+        {
+            RECT rect;
+            ::GetWindowRect(tlw->GetHWND(), &rect);
+            // Height and width should be the same unless the user performed
+            // an Aero Snap operation.
+            if ((rect.bottom - rect.top) !=
+                (m_placement.rcNormalPosition.bottom - m_placement.rcNormalPosition.top)
+                ||
+                (rect.right - rect.left) !=
+                (m_placement.rcNormalPosition.right - m_placement.rcNormalPosition.left))
```
vadz, line 126: Sorry for nitpicking but I think it would be slightly more readable like this:
```suggestion
            const RECT rcNormal = m_placement.rcNormalPosition;
            if ((rcWindow.bottom - rcWindow.top) != (rcNormal.bottom - rcNormal.top) ||
                (rcWindow.right - rcWindow.left) != (rcNormal.right - rcNormal.left))
```
where I also renamed `rect` to `rcWindow` to make it more clear which rect is which and because wxMSW tries (maybe not 100% successfully) to use `rect` for `wxRect` and `rc` for `RECT`.

vadz (approved):
Thanks, if you don't mind, I'd like to make the trivial changes mentioned below, but I can do them myself if you're ok with them.

Of course, the real question is whether there is any better way to do this than by relying on the window rect, but I don't see anything, so if this works, let's do it like this -- the important thing here is to fix the user-visible problem.

Thanks again!

Issue #22734, opened 2022-08-23
wxStyledTextCtrl::CallTipShow with different DPI


ColinTebbutt:
wxWidgets v3.2.0 displays the wxStyledTextCtr call tip differently compared to v3.1.5 on a Windows machine with display magnification greater than 100%. In particular, v3.2.0 shows the call tip text as if the magnification was 100% (i.e. it shows as a small font), but computes the dispaly rectangle with the correct magnification (but this now ends up being too big). I'm using the default style - i.e. I do not use CallTipUseStyle().

The images below are with a display magnification of 125%, and show the issue.
V3.1.5
![v315](https://user-images.githubusercontent.com/60915614/186127572-7ac74ec4-203c-4187-aee5-5bfaa82eb52c.jpg)

v3.2.0
![v320](https://user-images.githubusercontent.com/60915614/186127602-dee15446-e752-4b1c-ba49-8c43191924ca.jpg)

I've had a look at the source code, but cannot see where to fix it.

MaartenBent:
I can reproduce it in the stc sample. The bitmap/MemoryDC used for drawing the calltip is not initialized with the correct DPI. We can probably use the same fix as we did for the [other](https://github.com/wxWidgets/wxWidgets/commit/31de97a2c636d9be49bfecd88cc2e6756ff356a7) bitmap, but I'll have to verify on macOS:
```diff
 src/stc/ScintillaWX.cpp | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

diff --git a/src/stc/ScintillaWX.cpp b/src/stc/ScintillaWX.cpp
index 4e450001b9..98f0c13778 100644
--- a/src/stc/ScintillaWX.cpp
+++ b/src/stc/ScintillaWX.cpp
@@ -115,7 +115,8 @@ public:
 
     void DrawBack(const wxSize& size)
     {
-        m_back = wxBitmap(size);
+        m_back = wxBitmap(ToPhys(size));
+        m_back.SetScaleFactor(GetDPIScaleFactor());
         wxMemoryDC mem(m_back);
         Surface* surfaceWindow = Surface::Allocate(m_swx->technology);
         surfaceWindow->Init(&mem, m_ct->wDraw.GetID());
```

ColinTebbutt:
Thanks for the quick response. That fix works for me (on Windows).

PR #22747, opened 2022-08-28
Fix DPI of wxSTC Calltip

MaartenBent:
See #22734

Can be backported to 3.2.

### DIFF
--- a/src/stc/ScintillaWX.cpp
+++ b/src/stc/ScintillaWX.cpp
@@ -115,7 +115,8 @@ public:
 
     void DrawBack(const wxSize& size)
     {
-        m_back = wxBitmap(size);
+        m_back = wxBitmap(ToPhys(size));
+        m_back.SetScaleFactor(GetDPIScaleFactor());
         wxMemoryDC mem(m_back);
         Surface* surfaceWindow = Surface::Allocate(m_swx->technology);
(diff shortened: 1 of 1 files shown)


vadz:
Thanks! I'll merge this soon, but I wonder (not for the first time I believe, but, unfortunately, I don't remember if we had already reached some decision about it) if we need some
```cpp
bool CreateWithLogicalSize(const wxSize& sz, wxWindow* win);
```
in `wxBitmap` doing this for completeness?

MaartenBent:
You mentioned this in #22465. But since wxSTC was the only place it was used (now twice), we just used `Create()` with `SetScaleFactor()`.

vadz:
Thanks for locating this discussion!

I think we do need to add such a function, because I don't see why would wxSTC be different from any other code using a backing buffer bitmap for drawing. Of course, such code (including wxSTC) should arguably use `wxBufferedDC`, but in practice it will be much simpler to update the existing code to use a bitmap with the correct scale factor than to change it to use that class, so such a function would be useful IMO.

Does anybody have any better suggestions for the name and/or signature than `CreateWithLogicalSize()` shown above? For the record, I also thought about having a function taking both logical and physical sizes and setting the scale factor to their quotient, could this be better/more clear?

PR #22748, opened 2022-08-28
CMake: Use builtin NanoSVG by default

MaartenBent:
Since we do not provide a FindNanoSVG module, we should use the builtin by default.
And prevent warnings about not being able to find the package configuration file.

See for example this build log: https://github.com/wxWidgets/wxWidgets/runs/7920178325?check_suite_focus=true#step:6:64

Can be backported to 3.2


vadz:
Thanks, I had already seen those and did wonder if it was possible to avoid them. I'll merge it and backport to 3.2 too.

Thanks again!

MaartenBent:
Done

Issue #22751, opened 2022-08-28
Crash in radio box_osx when initialized with null list of strings


rmpowell77:
**Describe the bug**
When a `wxRadioBox` is created with no `choices`, it will crash when destroyed.

**Expected vs observed behaviour**
I would not expect a segfault on destruction

**Patch or snippet allowing to reproduce the problem**
This program crashes with a segfault on Mac:

```
#include <iostream>
#include <wx/wx.h>

int main() {
    wxFrame frame { nullptr, wxID_ANY, "" };
    auto window = new wxRadioBox(&frame, wxID_ANY, "", wxDefaultPosition, wxDefaultSize, 0, nullptr, 1, wxRA_SPECIFY_COLS);
    // a crash occurs on frame dtor.
}
```

**To Reproduce**
use the attached sample program (cmake/code) on Mac
[crash.zip](https://github.com/wxWidgets/wxWidgets/files/9440475/crash.zip)

**Platform and version information**
 - wxWidgets version you use: 3.2
 - wxWidgets port you use: wxOSX
 - OS and its version: macOS 12.4

rmpowell77:
possible fix PR here: #22752

vadz:
I agree that we shouldn't crash but I'm not really sure if we should allow creating radio boxes without items -- can this ever be useful? Does it work correctly under the other platforms?

If we do want to allow it, #22752 is probably not sufficient as it's still going to crash at least in `SetFocus()`.

rmpowell77:
Well, we do have this pointy edge in the default constructor:


[...]

PR #22752, opened 2022-08-28
Fix for #22751: Crash in radio box_osx when initialized with null list of str…

rmpowell77:
…ings #22751

Do not attempt to dereference m_radioButtonCycle


vadz:
Sorry, I prefer a slightly different fix is a bit more clear, please see #22755.

rmpowell77:
Yes, that's a much better change!  Very nice.

PR #22753, opened 2022-08-29
Fix crash when using wxNotebook in applications calling `g_log_set_writer_func()` with glib 2.73+

vadz:


AliKet:
From their [documentation](https://docs.gtk.org/glib/func.log_set_writer_func.html) they clearly say:

> Libraries must not call this function — only programs are allowed to install a writer function, as there must be a single, central point where log messages are formatted and outputted.
> 
> There can only be one writer function. It is an error to set more than one.

So this PR is just doing the right thing AFAICS.

interface/wx/app.h
```diff
@@ -1061,6 +1061,23 @@ class wxApp : public wxAppConsole
...
+        logging. You must @e not call it if your application calls @c
+        g_log_set_writer_func() function itself, as this function can be only
+        called once.
+
+        It is recommended to call this function in your overridden version of
+        wxApp::OnInit() to allow wxWidgets to suppress some spurious GTK error
+        messages, e.g. the ones that happen whenever wxNotebook pages are
+        removed with the current GTK versions.
+
+        @since 3.2.1
```
imciner2, line 1077: This should be marked as wxGTK only as well.

interface/wx/app.h
```diff
@@ -1061,6 +1061,23 @@ class wxApp : public wxAppConsole
     */
     static void GTKSuppressDiagnostics(int flags = -1);
 
+    /**
+        Allows wxWidgets to selectively suppress some GTK messages.
+
+        This function can be called to allow wxWidgets to control GTK message
+        logging. You must @e not call it if your application calls @c
```
imciner2, line 1068:
```suggestion
        logging. You must @e not call it if your application calls the @c
```

imciner2 (approved):
Thanks for fixing this. Just two minor comments in the documentation.

interface/wx/app.h
vadz: Oops, thanks for noticing this, fixed (and the other one too).

I won't push the updated version yet to avoid rerunning the CI builds needlessly.

Issue #22745, opened 2022-08-27
cmake build fails with wxUSE_GTKPRINT=ON


antonio-rojas:
**Describe the problem**
Building with cmake and wxUSE_GTKPRINT=ON fails, because the gtkunixprint.h header location is not added to the include dirs
```
/build/wxwidgets/src/wxWidgets-3.2.0/src/gtk/print.cpp:37:10: fatal error: gtk/gtkunixprint.h: No such file or directory
   37 | #include <gtk/gtkunixprint.h>
      |          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

**Platform and version information**
 - wxWidgets version you use: 3.2.0
 - wxWidgets port you use: wxGTK
 - OS and its version: Linux x86_64
 - Compiler being used: GCC 12.2

vadz:
Thanks for reporting this.

We probably just need `pkg_check_modules(gtk+-unix-print-${TOOLKIT_VERSION}.0)` somewhere, but I'm not sure where exactly. Any PRs would be welcome.

MaartenBent:
Indeed. CMake doesn't have any of the `gtkprint` checks that configure has. This part of configure has to be converted to CMake:
https://github.com/wxWidgets/wxWidgets/blob/6ee593c78d082156c4fba2997241e530ba1d24da/configure.in#L5834-L5856

I can create a simple `FindGTKPRINT.cmake` module that has a `pkg_check_modules`, and use that to find the headers and libraries.

vadz:
Fixed in both master and 3.2, thanks again, Maarten!

PR #22762, opened 2022-08-30
CMake: Add GTK Printing support

MaartenBent:
See #22745

And show the extra toolkit options in the CMake summary, fix using libnotify and gnomevfs with gtk3 and fix some unused parameter warnings when using gtk2 or gtk3 on Windows.


vadz:
Thanks!

This should be fine to backport to 3.2 too, shouldn't it?

MaartenBent:
Yes, I guess it is fine to backport.

Or can this cause ABI issues if suddenly the library has GTK printing support (or when it starts using libnotify)? If so, maybe there should be a note in the changelog that these options should be disabled to keep compatibility.

MaartenBent:
Do you still want to backport this? And close #22745?

vadz:
I'm not sure... It looks like GTK print support doesn't even affect the ABI, but libnotify does and, in theory, it could indeed be unexpected that rebuilding 3.2.1 libraries using the same commands as 3.2.0 results in an incompatible binary. OTOH I don't even know if our ABI guarantee covers anything other than the binaries that we produce ourselves, so maybe it's not really a problem?

MaartenBent:
Yes, if the ABI is only for our libraries than we can backport this.
We could still add a note to the changelog so other packagers are aware of it (CMake GTK3 build on Linux now correctly searches for, and enables, libnotify. For 3.2.0 behaviour, build with `-DwxUSE_LIBNOTIFY=0`.

Issue #22763, opened 2022-08-31
Bombs Demo has issues with Easy Corner setting


Night-Wisp:
**Describe the bug**
When running the Bombs Demo, the Easy Corner setting in the menu has 2 minor related bugs.
##### Bug 1:
When the demo asks to confirm the new Easy Corner setting, the displayed text shows "...disable..." when enabling the setting, and vice versa.
##### Bug 2:
When choosing not to confirm toggling the Easy Corner setting, the checkmark by it still toggles.

**Expected vs observed behaviour**
##### Bug 1:
Expected behavior:
When enabling the setting, the dialog should ask if you want to enable the setting.
When disabling the setting, the dialog should ask if you want to disable the setting.
Observed behavior:
When enabling the setting, the dialog asks if you want to disable the setting.
When disabling the setting, the dialog asks if you want to enable the setting.
##### Bug 2:
Expected behavior:
The checkmark by the Easy Corner setting should be visible when enabled, hidden when disabled.
Observed behavior:
When clicking the Easy Corner menu item, the checkmark toggles it's state, regardless of whether the setting is switched.

**To Reproduce**
##### Bug 1:
1. Go to your wxWidgets build directory
2. Go to 'demos/bombs'
3. Run the bombs executable
4. Open the 'File' menu
5. Click 'Easy Corner'
6. The 'Confirm' dialog asks 'Do you really want to _enable/disable_...'
7. '_enable/disable_' doesn't align with the state the setting will be in after confirming
##### Bug 2:
1. Go to your wxWidgets build directory
2. Go to 'demos/bombs'
3. Run the bombs executable
4. Open the 'File' menu
5. Observe the state of the checkmark by 'Easy Corner'
6. Click 'Easy Corner'
7. The 'Confirm' dialog pops up. Select 'No'
8. The checkmark by 'Easy Corner' has switched, but the setting hasn't

**Platform and version information**
 - wxWidgets version you use: 3.0.5
 - wxWidgets port you use: wxGTK
 - OS and its version: Raspbian GNU/Linux 10 (buster)
 - For wxGTK only
   + GTK version: 3.24.5
   + Which GDK backend is used: Unsure

Night-Wisp:
Issue fixed with closing of PR #22764

PR #22764, opened 2022-08-31
Night_Wisp bombs demo patch

Night-Wisp:
* Fix two minor errors in the Bombs demo.

Fix two minor errors in the Bombs demo.

Error 1: When enabling/disabling easy corner, the enable/disable text was mixed up. This resulted in being asked to confirm disabling easy corner, when it was being enabled, and vice versa.
Error 2: When choosing not to switch easy corner, the checkmark in the menu showing it's state still swapped.

Fix 1: Swap the text for enable/disable in the confirm easy mode toggle dialog box.
Fix 2: Call GetMenuBar()->Check(bombsID_EASYCORNER, m_easyCorner); (Check always returns void, and GetMenuBar is only called from the MenuBar, after it's been set up) to set the checkmark's state, when needed.

Added myself in a Modified By section

This fixes issue #22763

### DIFF
--- a/demos/bombs/bombs.cpp
+++ b/demos/bombs/bombs.cpp
@@ -3,7 +3,6 @@
 // Purpose:     Bombs game
 // Author:      P. Foggia 1996
 // Modified by: Wlodzimierz Skiba (ABX) since 2003
-// Modified by: Night_Wisp in 2022
 // Created:     1996
 // Copyright:   (c) 1996 P. Foggia
 // Licence:     wxWindows licence
@@ -197,11 +196,14 @@ void BombsFrame::OnEasyCorner(wxCommandEvent& WXUNUSED(event))
                this
              );
 
-    if(ok!=wxYES)return GetMenuBar()->Check(bombsID_EASYCORNER, m_easyCorner);
+    if(ok!=wxYES)
+    {
+        // Undo the automatic change to the menu bar, to keep it in sync with the actual option value.
+        GetMenuBar()->Check(bombsID_EASYCORNER, m_easyCorner);
+        return;
+    }
 
     m_easyCorner = !m_easyCorner;
-    
-    GetMenuBar()->Check(bombsID_EASYCORNER, m_easyCorner);
 
     NewGame(m_lastLevel, true);
 }
--- a/demos/bombs/bombs.cpp
+++ b/demos/bombs/bombs.cpp
@@ -198,7 +198,7 @@ void BombsFrame::OnEasyCorner(wxCommandEvent& WXUNUSED(event))
 
     if(ok!=wxYES)
     {
-        // Undo the automatic change to the menu bar, to keep it in sync with the actual option value.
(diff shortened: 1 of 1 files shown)


demos/bombs/bombs.cpp
```diff
@@ -196,9 +197,11 @@ void BombsFrame::OnEasyCorner(wxCommandEvent& WXUNUSED(event))
                this
              );
 
-    if(ok!=wxYES)return;
+    if(ok!=wxYES)return GetMenuBar()->Check(bombsID_EASYCORNER, m_easyCorner);
 
     m_easyCorner = !m_easyCorner;
+    
+    GetMenuBar()->Check(bombsID_EASYCORNER, m_easyCorner);
```
vadz, line 204: I don't think this is necessary, the check menu item should have been already toggled, no?

demos/bombs/bombs.cpp
```diff
@@ -196,9 +197,11 @@ void BombsFrame::OnEasyCorner(wxCommandEvent& WXUNUSED(event))
                this
              );
 
-    if(ok!=wxYES)return;
+    if(ok!=wxYES)return GetMenuBar()->Check(bombsID_EASYCORNER, m_easyCorner);
```
vadz, line 200: Minor, but we don't put compound statements on the same line. Also, a comment would probably be useful:
```suggestion
    if(ok!=wxYES)
    {
        // Undo the change to the menu bar, to keep it in sync with the actual option value.
        return GetMenuBar()->Check(bombsID_EASYCORNER, m_easyCorner);
    }
```

demos/bombs/bombs.cpp
```diff
@@ -3,6 +3,7 @@
 // Purpose:     Bombs game
 // Author:      P. Foggia 1996
 // Modified by: Wlodzimierz Skiba (ABX) since 2003
+// Modified by: Night_Wisp in 2022
```
vadz, line 6: We don't really use those any longer since the switch to Git, you will be credited as the commit author anyhow.

vadz (commented):
Thanks for the fixes, they seem correct to me, but I think there are a couple of minor improvements that could be made to make the PR even better.

vadz:
Oh, and there is also [this small issue](https://github.com/wxWidgets/wxWidgets/runs/8105455208?check_suite_focus=true)

demos/bombs/bombs.cpp
Night-Wisp: Yes, it's been removed now.

Night-Wisp:
@vadz I fixed the prior issues. But two different checks failed, without any changes made to any other files. Something is probably causing an intermittent issue.

vadz:
Thanks for the fixes, I'll merge this soon.

The tests failures are unrelated, unfortunately there is some flakiness in `wxWebView` tests. I hope to be able to debug it one day, but for now we just have to live with spurious failures from time to time, sorry about this.

PR #22765, opened 2022-09-01
Gtk diagnostics

joukj:
Added in wxgtk1 GTKSupressDiagnostics & GTKAllowDiagnosticsControl. 
I added the dummy routines (as in wxgtk without wxHAS_GLIB_LOG_WRITER), since OpenVMS (probably the only port still using wxgtk1) does not have this log-writer.

The reason for adding is that some sample rograms (i.e. notebook) need the routines.


vadz:
As discussed on the list, I don't think we want to apply this, sorry.

PR #22766, opened 2022-09-01
Remove useless check for grid size in wxGrid::FreezeTo()

vadz:
This resulted in silently doing nothing when calling FreezeTo() on a
grid which hadn't been resized to its full size yet, which seems wrong
and not useful at all, so simply remove this check and freeze the
requested rows/columns in any case -- it's the user responsibility to
make the grid sufficiently big to allow the user to see the non-frozen
parts.

---

I'd like to make this change (at least in master, it might be too incompatible to backport to 3.2) because the current behaviour doesn't seem useful at all to me because it results in all calls to `FreezeTo()` done before the first layout to just do nothing at all.

Ideal would arguably be to freeze/thaw the grid automatically as it becomes big enough/too small, but this is more difficult and I think this is more useful than the current behaviour.

Any objections?

### DIFF
--- a/interface/wx/grid.h
+++ b/interface/wx/grid.h
@@ -5643,7 +5643,8 @@ public:
 
         Note that this method doesn't do anything, and returns @false, if any
         of the following conditions are true:
-        - Either @a row or @a col are out of range
+        - Either @a row or @a col are out of range, i.e. negative or strictly
+          greater than the total number of grid rows or columns.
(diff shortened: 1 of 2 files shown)


vadz:
Testing my use of this function, I found more problems with it, notably the fact that `FreezeTo(0,0)` may not do anything is very problematic and I think this one should be backported to 3.2, even if the other changes are not.

Issue #22761, opened 2022-08-30
wxIMAGE_QUALITY_HIGH not working with MSVS 2022


dkulp:
**Describe the bug**

The xLights team has moved up to the latest wxWidgets master code (plus some fixes).  However, we are having some problems with scaled images on Windows, but only using optimized builds with Visual Studio.  Debug builds are fine as are the Mac and Linux builds.    The code we are using is similar to:

```
wxBitmapButton* LayoutPanel::CreateModelButton(const char *data[]) {
    wxImage image(data);
    wxImage img24 = image.Scale(24, 24, wxIMAGE_QUALITY_HIGH);
    wxBitmapBundle bmp = wxBitmapBundle::FromBitmaps(img24, image);
    return new wxlBitmapButton(panel, bmp);
 }
```
Basically, the images (which are 48x48) are scaled down to 24x24 and a wxBitmapBundle created with the 24x24 and 48x48 images and a wxBitmapButton created using that.    On Windows optimized builds on non-HiDPI screens, the resulting image is black.   (not sure on HiDPI screens)    The workaround we are using is to use wxIMAGE_QUALITY_BICUBIC on Windows.  I've attached sample images in a zip. (github doesn't support attaching xpm's directly).   data is the xpm data.

See:
https://github.com/smeighan/xLights/issues/3388
https://github.com/smeighan/xLights/issues/3397

**Expected vs observed behaviour**
Scaled images display correctly.

**Patch or snippet allowing to reproduce the problem**

**Platform and version information**
 - wxWidgets 3.3 (master branch as of Aug 24)
 - wxWidgets port you use: wxMSW
 - OS and its version: Windows 10
  
[images.zip](https://github.com/wxWidgets/wxWidgets/files/9452603/images.zip)

PBfordev:
I can confirm this, it seems unrelated to bitmap bundle or XPM and can be reproduced in the image sample with this trivial patch
```diff
diff --git a/samples/image/canvas.cpp b/samples/image/canvas.cpp
index 9a634fed4f..cb62d367e4 100644
--- a/samples/image/canvas.cpp
+++ b/samples/image/canvas.cpp
@@ -96,7 +96,7 @@ MyCanvas::MyCanvas( wxWindow *parent, wxWindowID id,
     }
     else
     {
-        my_horse_png = wxBitmap( image );
+        my_horse_png = image.Scale(128, 128, wxIMAGE_QUALITY_HIGH);
     }
 
[...]

PR #22770, opened 2022-09-02
Change wxImage::ResampleBox() to avoid MSVS 2022 optimizer bug

vadz:
Calculate averaged_pixels directly instead of doing it in a loop as we
somehow end up with a value of 0 for it when using MSVS 2022 optimized
64-bit build.

See #22761.


vadz:
Merged as c6c6dacd640830dccd5ca7ae5820adbed2bc7a4d.

PR #22774, opened 2022-09-05
Improve wxBitmapBundle documentation

PBfordev:
I believe that using an SVG or a custom bitmap source should be mentioned in the class description, SVG in particular. Currently, this can only be deduced from the method list.

Removing that sentence may seem overly aggressive, but IMO there are no plans for this, such promises are often not fulfilled, and thus sentences like this are not useful. In fact I was itching to remove "These limitations will be relaxed in the future wxWidgets versions." from the `FromSVG()` description for the very same reason.


vadz:
This is indeed nice, thanks!

PR #22781, opened 2022-09-08
Add wxWebView::GetNativeConfiguration()

TcT2k:
New method return the pointer to the native configuration used during creation of this control

This allows for customization that would not be available to the application.

When using two-step creation this method can be used to customize configuration options not available via  `GetNativeBackend()` after using `Create()`.

The return value needs to be down-casted to the appropriate type depending on the platform: under macOS, it's a
<a href="https://developer.apple.com/documentation/webkit/wkwebviewconfiguration">WKWebViewConfiguration</a> pointer, under Windows with Edge it's a pointer to <a href="https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2environmentoptions">ICoreWebView2EnvironmentOptions</a>.
With other backends/platforms it's not implemented.

The following pseudo code shows how to use this method with two-step creation to set no user action requirement to  play video in a web view:
```C++
#if defined(__WXMSW__)
#include "webview2.h"
#elif defined(__WXOSX__)
#import "WebKit/WebKit.h"
#endif

wxWebView* webView = wxWebView::New();
#if defined(__WXMSW__)
ICoreWebView2EnvironmentOptions* webViewOptions =
    (ICoreWebView2EnvironmentOptions*) webView->GetNativeConfiguration();
webViewOptions->put_AdditionalBrowserArguments("--autoplay-policy=no-user-gesture-required");
#elif defined(__WXOSX__)
WKWebViewConfiguration* webViewConfiguration =
    (WKWebViewConfiguration*) webView->GetNativeConfiguration();
webViewConfiguration.mediaTypesRequiringUserActionForPlayback = WKAudiovisualMediaTypeNone;
#endif
webView->Create(this, wxID_ANY, "https://www.wxwidgets.org");
```

Obviously this specific configuration sample could be wrapped in wxWidgets but there are so many obscure platform specific configuration options that it wouldn't make sense to provide them all.

ToDo:
- [x] Implement for macOS
- [x] Implement for Edge

### DIFF
--- a/include/wx/private/jsscriptwrapper.h
+++ b/include/wx/private/jsscriptwrapper.h
@@ -33,38 +33,42 @@ public:
     };
 
     wxJSScriptWrapper(const wxString& js, OutputType outputType)
-        : m_escapedCode(js), m_outputType(outputType)
+        : m_outputType(outputType)
     {
         // Adds one escape level.
         const char *charsNeededToBeEscaped = "\\\"\n\r\v\t\b\f";
-        for (
-            size_t pos = m_escapedCode.find_first_of(charsNeededToBeEscaped, 0);
-            pos != wxString::npos;
-            pos = m_escapedCode.find_first_of(charsNeededToBeEscaped, pos)
-        ) {
-            switch (m_escapedCode[pos].GetValue())
+        m_escapedCode.reserve(js.size());
+        for (wxString::const_iterator it = js.begin(); it != js.end(); ++it)
+        {
+            if (wxStrchr(charsNeededToBeEscaped, *it))
             {
-            case 0x0A: // '\n'
-                m_escapedCode[pos] = 'n';
-                break;
-            case 0x0D: // '\r'
-                m_escapedCode[pos] = 'r';
-                break;
-            case 0x0B: // '\v'
-                m_escapedCode[pos] = 'v';
-                break;
-            case 0x09: // '\t'
-                m_escapedCode[pos] = 't';
-                break;
-            case 0x08: // '\b'
-                m_escapedCode[pos] = 'b';
-                break;
-            case 0x0C: // '\f'
-                m_escapedCode[pos] = 'f';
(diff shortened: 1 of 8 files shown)


include/wx/private/jsscriptwrapper.h
```diff
@@ -33,38 +33,42 @@ class wxJSScriptWrapper
...
         // Adds one escape level.
         const char *charsNeededToBeEscaped = "\\\"\n\r\v\t\b\f";
-        for (
-            size_t pos = m_escapedCode.find_first_of(charsNeededToBeEscaped, 0);
-            pos != wxString::npos;
-            pos = m_escapedCode.find_first_of(charsNeededToBeEscaped, pos)
-        ) {
-            switch (m_escapedCode[pos].GetValue())
+        m_escapedCode.reserve(js.size());
+        for (wxString::const_iterator it = js.begin(); it != js.end(); ++it)
```
vadz, line 41: I think the performance gain comes from not calling `insert()` rather than not using `find_first_of()`, doesn't it? I.e. it might be still advantageous to keep the old loop. We could add a small benchmark to `tests/benchmarks` to check this.

vadz (approved):
Thanks, will merge soon.

include/wx/private/jsscriptwrapper.h
vadz: P.S. This commit could/should probably be backported to 3.2 too.

TcT2k: Yes the improvement is probably mainly due to the `insert()` calls. But I found the structure easier with a simple loop over the whole string.
And yes this should be backported to 3.2 as it was quite significant difference in my observation. I was trying to send around 5MB of JSON data to my javascript via `RunScriptAsync()` and it took over 90 seconds in debug mode on a M1 mac. With the new loop it's negligible.

PR #22787, opened 2022-09-09
Fix depth buffer for the OpenGL samples under Linux

buu342:
This PR is being created as a result of the conversation [here](https://forums.wxwidgets.org/viewtopic.php?f=23&t=49603).

Two of the OpenGL samples, specifically the Cube and Penguin sample, did not have a working depth buffer when compiled under Linux. While the cube sample wasn't visually affected by this (only being visible if backface culling was disabled), the visual issues could be visible in the penguin sample under specific rotations.

The issue was caused by mistakenly assuming the depth buffer was enabled by default on Linux, which is not the case. This PR contains two commits, one which adds the depth buffer size to the attribute list in the cube sample (which uses the old integer array method already), and which adds a `wxGLAttributes` object as an argument to the Canvas implementation (which is the preferred method as of wx 3.2).

---
As an aside, the [documentation](https://docs.wxwidgets.org/trunk/classwx_g_l_attributes.html#a547c600aa202a3fdee88ac9c3594c819) of the depth value for `wxGLAttributes` claims that valid values are 0, 16, and 32. However, when I attempt to use the value of 32, it causes the following assertions on the application start:
```
../src/unix/glegl.cpp(372): assert ""Assert failure"" failed in InitVisual(): Failed to get an EGLConfig for the requested attributes.
../src/unix/glegl.cpp(311): assert ""fbc"" failed in wxGLContext(): Invalid EGLConfig for OpenGL
```
Interestingly, a value of 24 works fine. Is this a mistake in the documentation? If so, I can add an extra commit to this PR to correct that.

### DIFF
--- a/samples/opengl/cube/cube.cpp
+++ b/samples/opengl/cube/cube.cpp
@@ -458,9 +458,10 @@ wxEND_EVENT_TABLE()
 MyFrame::MyFrame( bool stereoWindow )
        : wxFrame(NULL, wxID_ANY, "wxWidgets OpenGL Cube Sample")
 {
-    int stereoAttribList[] = { WX_GL_RGBA, WX_GL_DOUBLEBUFFER, WX_GL_STEREO, 0 };
+    int attribList[] = { WX_GL_RGBA, WX_GL_DOUBLEBUFFER, WX_GL_DEPTH_SIZE, 16, 0 };
+    int stereoAttribList[] = { WX_GL_RGBA, WX_GL_DOUBLEBUFFER, WX_GL_DEPTH_SIZE, 16, WX_GL_STEREO, 0 };
 
-    new TestGLCanvas(this, stereoWindow ? stereoAttribList : NULL);
+    new TestGLCanvas(this, stereoWindow ? stereoAttribList : attribList);
 
     SetIcon(wxICON(sample));
 
@@ -481,7 +482,7 @@ MyFrame::MyFrame( bool stereoWindow )
     Show();
 
     // test IsDisplaySupported() function:
-    static const int attribs[] = { WX_GL_RGBA, WX_GL_DOUBLEBUFFER, 0 };
+    static const int attribs[] = { WX_GL_RGBA, WX_GL_DEPTH_SIZE, 16, WX_GL_DOUBLEBUFFER, 0 };
     wxLogStatus("Double-buffered display %s supported",
                 wxGLCanvas::IsDisplaySupported(attribs) ? "is" : "not");
 
--- a/samples/opengl/cube/cube.cpp
+++ b/samples/opengl/cube/cube.cpp
@@ -34,6 +34,10 @@
     #include "../../sample.xpm"
 #endif
 
+#ifndef WGL_STEREO_ARB
+    #define WGL_STEREO_ARB  0x2012
+#endif
+
 // ----------------------------------------------------------------------------
 // constants
 // ----------------------------------------------------------------------------
@@ -309,12 +313,12 @@ wxBEGIN_EVENT_TABLE(TestGLCanvas, wxGLCanvas)
     EVT_TIMER(SpinTimer, TestGLCanvas::OnSpinTimer)
 wxEND_EVENT_TABLE()
 
-TestGLCanvas::TestGLCanvas(wxWindow *parent, int *attribList)
+TestGLCanvas::TestGLCanvas(wxWindow *parent, const wxGLAttributes& attribList)
     // With perspective OpenGL graphics, the wxFULL_REPAINT_ON_RESIZE style
     // flag should always be set, because even making the canvas smaller should
     // be followed by a paint event that updates the entire canvas with new
     // viewport settings.
-    : wxGLCanvas(parent, wxID_ANY, attribList,
+    : wxGLCanvas(parent, attribList, wxID_ANY,
                  wxDefaultPosition, wxDefaultSize,
                  wxFULL_REPAINT_ON_RESIZE),
       m_xangle(30.0),
@@ -323,15 +327,13 @@ TestGLCanvas::TestGLCanvas(wxWindow *parent, int *attribList)
       m_useStereo(false),
       m_stereoWarningAlreadyDisplayed(false)
 {
-    if ( attribList )
+    int i = 0;
+    const int* attributes = attribList.GetGLAttrs();
+    while ( attributes[i] != 0 )
     {
-        int i = 0;
-        while ( attribList[i] != 0 )
-        {
-            if ( attribList[i] == WX_GL_STEREO )
-                m_useStereo = true;
-            ++i;
-        }
+        if ( attributes[i] == WGL_STEREO_ARB )
+            m_useStereo = true;
+        ++i;
     }
 }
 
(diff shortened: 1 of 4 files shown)


samples/opengl/cube/cube.cpp
```diff
@@ -458,9 +458,10 @@ wxEND_EVENT_TABLE()
 MyFrame::MyFrame( bool stereoWindow )
        : wxFrame(NULL, wxID_ANY, "wxWidgets OpenGL Cube Sample")
 {
-    int stereoAttribList[] = { WX_GL_RGBA, WX_GL_DOUBLEBUFFER, WX_GL_STEREO, 0 };
+    int attribList[] = { WX_GL_RGBA, WX_GL_DOUBLEBUFFER, WX_GL_DEPTH_SIZE, 16 };
```
MaartenBent, line 461: This list should be [0-terminated](https://docs.wxwidgets.org/latest/classwx_g_l_canvas.html#a56d0a2a022c0c34b7efa1114da151177), on Windows the cube doesn't even render because of this.

Maybe this is a good moment to start using `wxGLAttributes` here too instead of an `attribList`.

vadz:
Thanks for the patch, but the problem pointed out by Maarten definitely must be fixed before it's merged.

Also, I don't know for sure what are the valid values for the depth attribute, but I suspect it's driver-dependent.

buu342:
Ah whoops, my bad. Fixed. However as pointed out by @MaartenBent, perhaps this is a good opportunity to port the cube sample to use the preferred `wxGLAttributes` object instead?

samples/opengl/cube/cube.cpp
buu342: Fixed.

vadz:
> perhaps this is a good opportunity to port the cube sample to use the preferred `wxGLAttributes` object instead?

Sure, please feel free to add a commit doing this. TIA!

samples/opengl/penguin/penguin.cpp
```diff
@@ -92,7 +92,9 @@ MyFrame::MyFrame(wxFrame *frame, const wxString& title, const wxPoint& pos,
 
     Show(true);
 
-    m_canvas = new TestGLCanvas(this, wxID_ANY, wxDefaultPosition,
+    wxGLAttributes attribList;
+    attribList.PlatformDefaults().Depth(16).EndList();
```
vadz, line 96: Sorry, after actually looking at the code I don't understand why is this necessary -- the defaults already include `Depth(16)` on all platforms. So I think this change should be undone.

vadz (commented):
I'm now even more convinced that we should stop hardcoding the default (non stereo) attributes because otherwise things are too confusing: you need to specify the depth explicitly when doing this but you do _not_ need to do it when using `wxGLAttributes::Defaults()` which already includes it.

samples/opengl/penguin/penguin.cpp
buu342: I am afraid that is not true, this is what happens if I don't specify `Depth(16)` on Linux:
https://i.imgur.com/1f8yPCJ.png

This was the point of contention in the thread that I linked in the OP of this PR. Under Linux, wxWidgets fails to initialize the depth buffer properly.

buu342: Ok upon further inspection of the documentation, I think I now realize that you were probably referring to `Defaults()`, and not `PlatformDefaults()` :sweat_smile:

vadz: OK, this is due to passing `NULL` to `wxGLCanvas` ctor and `wxGLAttributes::AddDefaultsForWXBefore31()` not specifying depth in X11 implementation. The real fix is to stop using the old ctor taking `const int*` and pass it `wxGLAttributes::Defaults()` instead.

Could you please do it like this and check that it works?

buu342: I think my confusion stemmed a little bit from the pyramid sample, which uses both
`vAttrs.PlatformDefaults().Defaults().EndList();` 
AND 
`vAttrs.PlatformDefaults().RGBA().DoubleBuffer().Depth(16).EndList();`

I guess I'll modify the penguin sample to use `PlatformDefaults().Defaults()`.

> The real fix is to stop using the old ctor taking const int* and pass it wxGLAttributes::Defaults() instead.
Are you referring to the cube sample? The code you highlighted in this review is from the penguin sample, which isn't using the integer array.

vadz: > I think my confusion stemmed a little bit from the pyramid sample, which uses both `vAttrs.PlatformDefaults().Defaults().EndList();` AND `vAttrs.PlatformDefaults().RGBA().DoubleBuffer().Depth(16).EndList();`

The second one is clearly not ideal, there is no need to duplicate `Defaults()`. The first one is probably fine, even though I think it would make more sense to specify platform defaults after the global defaults and, in fact, they could be omitted entirely because I don't think the sample would need them and there are no platform-specific defaults at all right now anyhow.
 
> > The real fix is to stop using the old ctor taking const int* and pass it wxGLAttributes::Defaults() instead.
> 
> Are you referring to the cube sample? The code you highlighted in this review is from the penguin sample, which isn't using the integer array.

It does use it, see `TestGLCanvas` ctor in it: it calls `wxGLCanvas` ctor with `NULL`, but it should use the attributes object here instead. If you change this, everything should work without having to specify the depth explicitly.

buu342: > It does use it, see `TestGLCanvas` ctor in it: it calls `wxGLCanvas` ctor with `NULL`, but it should use the attributes object here instead. If you change this, everything should work without having to specify the depth explicitly.

The original does not pass attributes, yes. My modified sample does. You're referring to the original, right? :sweat_smile: 

I have pushed another commit to the penguin sample, which uses `Defaults().PlatformDefaults()` now, please confirm if it is to your liking!

vadz:
Let's do it like this, thanks. Do you plan to push more commits to switch to using `wxGLAttributes` in the cube sample or should we merge this as is?

buu342:
> Let's do it like this, thanks. Do you plan to push more commits to switch to using `wxGLAttributes` in the cube sample or should we merge this as is?

I'll update the cube sample right now.

buu342:
I have ported the cube sample over to use wxGLAttributes, but I don't have a stereo compatible device to test it on. 

I'm also unsure if the loop to check if stereo is enabled is done correctly, since the attributes themselves aren't "exposed" like they were in an integer array (I had to redefine WGL_STEREO_ARB since it's not available in any header).

samples/opengl/cube/cube.cpp
```diff
@@ -34,6 +34,10 @@
     #include "../../sample.xpm"
 #endif
 
+#ifndef WGL_STEREO_ARB
+    #define WGL_STEREO_ARB  0x2012
+#endif
+
```
vadz, line 40: Sorry, this is not going to work under non-MSW systems which don't use `WGL_STEREO_ARB` at all (but `GLX_STEREO` which has a completely different value).

Instead we need to change the logic in the sample and just pass `bool useStereo` to the ctor explicitly (or, better, use `int requestedFlags` with `Requested_Stereo` being the only defined flag for now).

vadz:
@buu342 Could you please check if #22828 solves the original problem?

buu342:
Seems like it does! I compiled his branch of wx, and then built the penguin and cube (with the depth buffer disabled), and the depth buffer works as intended.

vadz:
Thanks for testing! I've now pushed that change to master (see the commit above), so I'm not going to apply the changes here then.

Issue #22694, opened 2022-08-01
Add transparent selection capability to wxGrid control


AliKet:
**Describe the proposed addition**

Any modern software nowadays having grid-like control (e.g. **Microsoft Excel**,
**LibreOffice Calc**,...) uses a transparent selection instead of using solid colour
with inversion to denote selections.

In an effort to bring this capability to **wxGrid** control, here is a patch implementing this
feature besides the old way of selection. In other words, **no intention** to replace, remove or deprecate the old mechanism for the short or long term.

Before doing any PR, i would like from the interested people to try the attached patch (which can be applied to master or wx-3.2) and play with the grid sample:

- You can switch between transparent and solid selection by **Ctrl+Shift+T**
- Try some scrolling while selecting
- Try transparent selection with some frozen columns/rows and do some scrolling.

**The general idea behind the implementation:**

Doing a transparent selection (as implemented in the patch) consists of transforming the selected cells/rectangles (only the visible area) into a one rectangle/polygon/poly-polygon shape and draw it using alpha from inside the paint event.
A DC supporting alpha drawing is required for that matter. Otherwise a graphics context will be used for drawing (e.g. **GTK2** and **MSW**)

- The selected rectangles are retrieved by `wxGrid::GetSelectedRectangles()`
- Converted to poly-polygon (represented by `wxGridSelection::PolyPolygon` type) by a helper class `PolyPolygonHelper` (which can implement the **sweep line algorithm**, as done by **LibreOffice**) and finally draw the result using `wxGrid::DrawSelection()`

**The patch:**

[gridsel.zip](https://github.com/wxWidgets/wxWidgets/files/15316337/gridsel.zip)

(Tested under Linux 18.04 X11/Wayland with wxGTK3/2 and wxQt, and under Windows 10)

Regards.
AliKet.

AliKet:
Testing the above patch under wxOSX is really appreciated, TIA!

vadz:
Thanks, I didn't look at this yet but just wanted to say that I don't think there is any barrier to replacing the current selection display with the new one if it works well. We try not to break the API, unless it's really needed, and then preferably with a long lead time, but we didn't commit to preserving the exact same UI look or behaviour in the future wx versions.

AliKet:
[...]

PR #22795, opened 2022-09-12
wxGrid: minor fixups and cleanups 

AliKet:
This is the first PR in an attempt to implement this feature request (#22694)

In other words, i want to add transparent selection to `wxGrid` , keeping the old mechanism
in place as alternative/user preference (or as a fallback where transparency is not supported by the target system)

**Old:**
Screenshot_2022-09-12_14-42-56

**New:**
Screenshot_2022-09-12_14-43-19

### DIFF
--- a/include/wx/generic/gridsel.h
+++ b/include/wx/generic/gridsel.h
@@ -119,7 +119,7 @@ private:
     {
         SelectBlock(block.GetTopRow(), block.GetLeftCol(),
                     block.GetBottomRow(), block.GetRightCol(),
-                    wxKeyboardState(), false);
+                    wxKeyboardState(), wxEVT_NULL);
     }
 
     // Really select the block and don't check for the current selection mode.
--- a/include/wx/generic/gridsel.h
+++ b/include/wx/generic/gridsel.h
@@ -108,7 +108,7 @@ public:
     wxArrayInt GetRowSelection() const;
     wxArrayInt GetColSelection() const;
 
-    wxGridBlockCoordsVector& GetBlocks() { return m_selection; }
+    const wxGridBlockCoordsVector& GetBlocks() const { return m_selection; }
 
     void EndSelecting();
     void CancelSelecting();
--- a/include/wx/generic/gridsel.h
+++ b/include/wx/generic/gridsel.h
@@ -19,6 +19,8 @@
 
 #include "wx/vector.h"
 
+wxDEPRECATED_MSG("use wxGridBlockCoordsVector instead")
+typedef wxVector<wxGridBlockCoords> wxVectorGridBlockCoords;
 
 // Note: for all eventType arguments of the methods of this class wxEVT_NULL
 //       may be passed to forbid events generation completely.
--- a/include/wx/generic/gridsel.h
+++ b/include/wx/generic/gridsel.h
@@ -19,7 +19,6 @@
 
 #include "wx/vector.h"
 
-typedef wxVector<wxGridBlockCoords> wxVectorGridBlockCoords;
 
 // Note: for all eventType arguments of the methods of this class wxEVT_NULL
 //       may be passed to forbid events generation completely.
@@ -109,7 +108,7 @@ public:
     wxArrayInt GetRowSelection() const;
     wxArrayInt GetColSelection() const;
 
-    wxVectorGridBlockCoords& GetBlocks() { return m_selection; }
+    wxGridBlockCoordsVector& GetBlocks() { return m_selection; }
 
     void EndSelecting();
     void CancelSelecting();
@@ -136,7 +135,7 @@ private:
     // We don't currently check if the new block is contained by several
     // existing blocks, as this would be more difficult and doesn't seem to be
     // really needed in practice.
-    void MergeOrAddBlock(wxVectorGridBlockCoords& blocks,
+    void MergeOrAddBlock(wxGridBlockCoordsVector& blocks,
                          const wxGridBlockCoords& block);
 
     // All currently selected blocks. We expect there to be a relatively small
@@ -146,7 +145,7 @@ private:
     // Selection may be empty, but if it isn't, the last block is special, as
     // it is the current block, which is affected by operations such as
     // extending the current selection from keyboard.
-    wxVectorGridBlockCoords             m_selection;
+    wxGridBlockCoordsVector             m_selection;
 
     wxGrid                              *m_grid;
     wxGrid::wxGridSelectionModes        m_selectionMode;
--- a/tests/controls/gridtest.cpp
+++ b/tests/controls/gridtest.cpp
@@ -659,6 +659,12 @@ TEST_CASE_METHOD(GridTestCase, "Grid::RangeSelect", "[grid]")
     if ( !EnableUITests() )
         return;
 
+#ifdef __WXGTK20__
+    // Works locally, but not when run on Travis CI.
+    if ( IsAutomaticTest() )
+        return;
+#endif
+
     EventCounter select(m_grid, wxEVT_GRID_RANGE_SELECTED);
 
     wxUIActionSimulator sim;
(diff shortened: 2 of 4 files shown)


AliKet:
Any objections to implement `wxGrid::SetCurrentCell()` uniformly under all platforms like this:

```
index 0bcaaa7d5e..e07ed80eec 100644
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -6436,51 +6436,13 @@ bool wxGrid::SetCurrentCell( const wxGridCellCoords& coords )
 
         if ( IsVisible( m_currentCellCoords, false ) )
         {
-#if defined(__WXOSX__) || defined(__WXGTK3__)
             RefreshBlock(m_currentCellCoords, m_currentCellCoords);
-#else
-            wxGridWindow* prevGridWindow = CellToGridWindow(m_currentCellCoords);
-            wxRect r;
-            r = BlockToDeviceRect( m_currentCellCoords, m_currentCellCoords, prevGridWindow );
-            if ( !m_gridLinesEnabled )
-            {
-                r.x--;
-                r.y--;
-                r.width++;
-                r.height++;
-            }
-
-            wxGridCellCoordsArray cells = CalcCellsExposed( r, prevGridWindow );
-
-            // Otherwise refresh redraws the highlight!
-            m_currentCellCoords = coords;
-
-            wxClientDC prevDc( prevGridWindow );
-            PrepareDCFor(prevDc, prevGridWindow);
-
-            DrawGridCellArea( prevDc, cells );
-            DrawAllGridWindowLines( prevDc, r , prevGridWindow);
-
-            if ( prevGridWindow->GetType() != wxGridWindow::wxGridWindowNormal )
-                DrawFrozenBorder(prevDc, prevGridWindow);
-#endif
         }
     }
 
     m_currentCellCoords = coords;
 
-#if defined(__WXOSX__) || defined(__WXGTK3__)
     RefreshBlock(coords, coords);
-#else
-    if ( ShouldRefresh() )
-    {
-        wxGridCellAttrPtr attr = GetCellAttrPtr( coords );
-        wxGridWindow* currentGridWindow = CellToGridWindow(coords);
-        wxClientDC dc( currentGridWindow );
-        PrepareDCFor(dc, currentGridWindow);
-        DrawCellHighlight( dc, attr.get() );
-    }
-#endif
 
     return true;
 }
```
IMHO, i don't see much gain in drawing the grid cursor from outside the `wxEVT_PAINT` event instead of just refreshing the affected rectangle. this makes `wxGrid::SetCurrentCell()` implementation much simpler, especially when adding label-selection/highlight latter ?

AliKet:
Not sure why this is still failing even if it succeeds locally ?

include/wx/generic/gridsel.h
```diff
@@ -19,7 +19,6 @@
 
 #include "wx/vector.h"
 
-typedef wxVector<wxGridBlockCoords> wxVectorGridBlockCoords;
```
vadz, line 22: Not sure how did we end up with duplicate `wxGridBlockCoordsVector` and `wxVectorGridBlockCoords` typedefs, but unfortunately we can't just remove it from the public header any longer, even if it's clearly not needed.

Instead, it should be deprecated, with the message saying to use `wxGridBlockCoordsVector` instead, but left here (in a separate commit).

vadz (commented):
Thanks, this looks good except for removing the existing typedef. Please let me know if you have any problems with deprecating it.

vadz:
> Not sure why this still fails even though it succeeds locally ?

Unfortunately I don't know neither, but if the failure is reproducible, we need to fix it before merging this as we don't want to break the CI builds.

To debug it locally, please try running the test inside Xvfb.

vadz:
> Any objections to implement `wxGrid::SetCurrentCell()` uniformly under all platforms like this:

No, I don't think we really lose anything even under MSW by doing it like this. If you can test it under MSW, please still check that this doesn't produce any noticeable flicker when moving the cursor in the grid (and if you can't, please let me know so that I could do this).

AliKet:
> > Any objections to implement `wxGrid::SetCurrentCell()` uniformly under all platforms like this:
> 
> No, I don't think we really lose anything even under MSW by doing it like this. If you can test it under MSW, please still check that this doesn't produce any noticeable flicker when moving the cursor in the grid (and if you can't, please let me know so that I could do this).

In fact, I tested it before asking and saw no difference (i.e. no flickering and no loss in performance either)

tests/controls/gridtest.cpp
```diff
@@ -659,6 +659,12 @@ TEST_CASE_METHOD(GridTestCase, "Grid::RangeSelect", "[grid]")
     if ( !EnableUITests() )
         return;
 
+#ifdef __WXGTK20__
+    // Works locally, but not when run on Travis CI.
```
vadz, line 663: "Travis CI" is a copy/paste, should be "GitHub CI", right?

vadz:
> @vadz is it possible to rerun just the failing test because i don't see how is it related to this PR ? TIA !

I don't see neither but it definitely looks like it is related because I don't remember this check failing in the past and it seems to consistently fail now. I'd rather like to keep running all the tests, we spent much time on fixing them and it would be a pity to undo this work, even partially.

Note that the current check disables the test for both GTK 2 and 3, I don't care that much about GTK 2 any longer, so maybe we could live with disabling it for it (although it would still be pretty weird that it behaves differently from GTK 3), but I'd really like to keep the test with GTK 3.

Have you already tried running it under Xvfb locally to check if it fails there?

tests/controls/gridtest.cpp
AliKet: Oops! will correct this and the others in a moment, thanks.

AliKet:
> Note that the current check disables the test for both GTK 2 and 3, I don't care that much about GTK 2 any longer, so maybe we could live with disabling it for it (although it would still be pretty weird that it behaves differently from GTK 3), but I'd really like to keep the test with GTK 3.
> 

This is embarrassing because the other tests (i.e. `Grid::Size, Grid::ResizeScrolledHeader` and `Grid::ColumnMinWidth`) actually pass on GitHub CI and the erroneous check for GTK 2 also disables  GTK 3! my bad!

> Have you already tried running it under Xvfb locally to check if it fails there?

I run the test locally **with and without** Xvfb and both GTK 2&3 pass successfully.

AliKet:
Now i don't know what should i do with this failure at all because running the test with ASan under Ubuntu 18.04 succeeds with and without Xvfb. and i don't really know how to debug something working for me in a local machine !!

AliKet:
I've recompiled everything with the exact configuration options as the one using ASAN
but under Ubuntu 18.04 (GTK 3.22.30) and here is the output of the run:

```
$ xvfb-run -a -s '-screen 0 1600x1200x24' ./test_gui [grid]
build: 3.3.0 (wchar_t,compiler with C++ ABI compatible with gcc 4,wx containers,compatible with 3.2)
compiled using gcc 8.4
running under Linux 4.15.0-192-generic i686 as aliket
All tests passed (441 assertions in 41 test cases)

```

The test also succeeds if i run the whole testsuit (just i had to disable ImageTestCase which raise an exception and crash) except this one:
```
ChoiceTestCase
  GetBestSize
../../tests/controls/choicetest.cpp:36

../../tests/controls/choicetest.cpp:117: FAILED:
  CHECK( bestSize.GetHeight() < m_choice->FromDIP(35) )
with expansion:
  35 < 35
```

AliKet:
I think now the following test cases (`Grid::Size, Grid::ResizeScrolledHeader, Grid::ColumnMinWidth and Grid::RangeSelect`) are really covered by the tests on GitHub CI. i.e. : 
- The last one really detects the event from the range-selection and not from cell-selection (or the position of the cursor changes)
- The others were erroneousslly skipped under GTK3

vadz:
Thanks, I think this is probably good enough, even if it would still be interesting to know what's going on in GitHub environment (I hadn't thought about it, but maybe using tmate and ssh'ing into the failing build might give more information...), so I'm going to merge it.

Thanks again!

PR #22797, opened 2022-09-14
WebView handler improvements for advanced requests and Edge handler implementation

TcT2k:
The current `wxWebViewHandler` does only allow simple synchronous GET requests which is fine for many applications. To allow local REST like requests asynchronous POST requests, should be able to be processed.

I've added a new method `wxWebViewHandler::StartRequest()` which allows for more advanced request processing while maintaining backward compatibility through a default implementation that calls `wxWebViewHandler::GetFile()`.

A response can be send via the response object. The response does not have to be finished from this method and it's possible to be finished asynchronously via `wxWebViewHandlerResponse::Finish()`.

The following pseudo code demonstrates a typical implementation:
```C++
void MyAdvancedWebViewHandler::StartRequest(const wxWebViewHandlerRequest& request,
                      wxSharedPtr<wxWebViewHandlerResponse> response) wxOVERRIDE
{
    // Set common headers allowing acces from XMLHTTPRequests()
    response->SetHeader("Access-Control-Allow-Origin", "*");
    response->SetHeader("Access-Control-Allow-Headers", "*");

    // Handle options request
    if (request.GetMethod().IsSameAs("options", false))
    {
        response->Finish("");
        return;
    }

    // Check the post data type
    if (!request.GetHeader("Content-Type").StartsWith("application/json"))
    {
        response->FinishWithError();
        return;
    }

    // Process input data
    wxString postData = request.GetDataString();

    ...

    // Send result
    response->SetContentType("application/json");
    response->Finish("{ result: true }");
}
```

This is currently implemented for **WKWebView** and **Edge** backends. It should be possible for **webkit2** but I will probably not implement this in the first go.

The Edge backend currently does not support custom schemes. The handler implementation relies on "virtual hosts" which make handlers available at something like https://wxfs.wxsite . `LoadURL()` has handling to redirect a call to wxfs:///some.zip#protocol=zip/index.html to https://wxfs.wxsite///some.zip#protocol=zip/index.html . This allows for easy cross platform code in many cases, but obviously doesn't handle cases where the content requests custom schemes. 

In future versions custom schemes will be available via WebView2 but this is an acceptable compromise for now and can be interesting for some cases in the future, where virtual http(s) hosting of local assets might be wished.
The virtual host name defaults to the handlers name + ".wxsite" but can be customized via `wxWebViewHandler::SetVirtualHost()`.

As always any **feedback** is welcome on interface or implementation details. 
The edge backend support can be tested via the _webviewsample_.

ToDo:
- [x] Add macOS implementation
- [x] Add Edge implementation
- [x] Complete documentation
- [x] Integrate advanced webviewhandler into sample

### DIFF
--- a/include/wx/msw/private/comstream.h
+++ b/include/wx/msw/private/comstream.h
@@ -21,7 +21,37 @@ public:
 
     virtual ~wxCOMBaseStreamAdapter() { }
 
-    DECLARE_IUNKNOWN_METHODS;
+    // IUnknown
+    STDMETHODIMP QueryInterface(REFIID riid, void** ppv) wxOVERRIDE
+    {
+        if (riid == IID_IUnknown ||
+            riid == IID_ISequentialStream ||
+            riid == IID_IStream)
+        {
+            *ppv = this;
+            AddRef();
+            return S_OK;
+        }
+
+        *ppv = NULL;
+        return E_NOINTERFACE;
+    }
+
+    STDMETHODIMP_(ULONG) AddRef() wxOVERRIDE
+    {
+        return ++m_cRef;
+    }
+
+    STDMETHODIMP_(ULONG) Release() wxOVERRIDE
+    {
+        if (--m_cRef == wxAutoULong(0))
+        {
+                delete this;
+                return 0;
+        }
+        else
+            return m_cRef;
+    }
 
     // IStream
     virtual HRESULT STDMETHODCALLTYPE Seek(
@@ -110,6 +140,8 @@ protected:
 
         return wxFromStart;
     }
+private:
+    wxAutoULong m_cRef;
 };
 
 class wxCOMInputStreamAdapter : public wxCOMBaseStreamAdapter
@@ -167,13 +199,5 @@ public:
     }
 
 };
-
-BEGIN_IID_TABLE(wxCOMBaseStreamAdapter)
-ADD_IID(Unknown)
-ADD_IID(SequentialStream)
-ADD_IID(Stream)
-END_IID_TABLE;
-
-IMPLEMENT_IUNKNOWN_METHODS(wxCOMBaseStreamAdapter);
 
 #endif // WX_COMSTREAM_H
--- a/include/wx/msw/private/comstream.h
+++ b/include/wx/msw/private/comstream.h
@@ -1,0 +1,179 @@
+///////////////////////////////////////////////////////////////////////////////
+// Name:        wx/msw/private/comstream.h
+// Purpose:     COM Stream adapter to wxStreamBase based streams
+// Created:     2021-01-15
+// Copyright:   (c) 2021 wxWidgets team
+// Licence:     wxWindows licence
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef WX_COMSTREAM_H
+#define WX_COMSTREAM_H
+
+#include "wx/stream.h"
+#include "wx/msw/ole/comimpl.h"
+
+class wxCOMBaseStreamAdapter : public IStream
+{
+public:
+    wxCOMBaseStreamAdapter(wxStreamBase* stream):
+        m_stream(stream)
+    { }
+
+    virtual ~wxCOMBaseStreamAdapter() { }
+
+    DECLARE_IUNKNOWN_METHODS;
+
+    // IStream
+    virtual HRESULT STDMETHODCALLTYPE Seek(
+        LARGE_INTEGER WXUNUSED(dlibMove), DWORD WXUNUSED(dwOrigin),
+        ULARGE_INTEGER *WXUNUSED(plibNewPosition)) wxOVERRIDE
+    {
+        return E_NOTIMPL;
+    }
+
+    virtual HRESULT STDMETHODCALLTYPE SetSize(ULARGE_INTEGER WXUNUSED(libNewSize))
+    {
+        return E_NOTIMPL;
+    }
+
+    virtual HRESULT STDMETHODCALLTYPE CopyTo(IStream *WXUNUSED(pstm),
+        ULARGE_INTEGER WXUNUSED(cb), ULARGE_INTEGER *WXUNUSED(pcbRead),
+        ULARGE_INTEGER *WXUNUSED(pcbWritten)) wxOVERRIDE
+    {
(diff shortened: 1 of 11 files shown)


NewPagodi:
This is great.  I wrote something similar a while back for pulling files from an sqlite database to serve to a webview a while back.  I had it working on  windows with the internet explorer backend and gtk with webkit2gtk.  But I ultimately gave up because the performance was so dreadful on windows due to the ancient backend.  I can't seem to find the files now, but I remember it being surprisingly easy to get it working with webkit2gtk.

include/wx/msw/private/comstream.h
```diff
@@ -0,0 +1,179 @@
...
+
+};
+
+BEGIN_IID_TABLE(wxCOMBaseStreamAdapter)
+ADD_IID(Unknown)
+ADD_IID(SequentialStream)
+ADD_IID(Stream)
+END_IID_TABLE;
+
+IMPLEMENT_IUNKNOWN_METHODS(wxCOMBaseStreamAdapter);
```
vadz, line 177: This is going to result in link errors if this header is included more than once, so it'd be better to move it into some existing file if possible.

interface/wx/webview.h
```diff
@@ -341,7 +569,8 @@ class wxWebViewHandler
     <a href="https://docs.microsoft.com/en-us/microsoft-edge/hosting/webview2">Edge WebView2</a>.
     It is available for Windows 7 and newer.
 
-    This backend does not support custom schemes and virtual file systems.
+    This backend does not support custom schemes. When using handlers see
+    wxWebViewHandler::SetVirtualHost() for more details on how to use handlers.
```
vadz, line 573: Minor, but "when using handlers ... on how to use handlers" seems a bit weird.

src/common/webview.cpp
```diff
@@ -52,6 +54,92 @@ wxDEFINE_EVENT( wxEVT_WEBVIEW_FULLSCREEN_CHANGED, wxWebViewEvent);
...
+wxString wxWebViewHandlerRequest::GetDataString(const wxMBConv& conv) const
+{
+    wxInputStream* data = GetData();
+    if (!data)
+        return wxString();
+
+    size_t length = data->GetLength();
+    wxMemoryBuffer buffer;
+    data->ReadAll(buffer.GetWriteBuf(length), length);
+    wxString dataStr((const char*) buffer.GetData(), conv, length);
```
vadz, line 67: Just to show a good example:
```suggestion
    wxString dataStr(static_cast<const char*>(buffer.GetData()), conv, length);
```

interface/wx/webview.h
```diff
@@ -161,6 +161,155 @@ enum wxWebViewIE_EmulationLevel
...
+    */
+    virtual wxString GetURI() const { return GetRawURI(); }
+
+    /**
+        @return The body data of the request of @c null if nothing was posted.
+    */
+    virtual wxInputStream* GetData() const = 0;
+
+    /**
+        @return The body data of the request as a string.
```
vadz, line 195: Might be worth to mention that the returned string is empty if the encoding is wrong?

interface/wx/webview.h
```diff
@@ -161,6 +161,155 @@ enum wxWebViewIE_EmulationLevel
...
+        The header will be added if it hasn't been set before or replaced
+        otherwise.
+
+        @param name
+            Name of the header
+        @param value
+            String value of the header
+    */
+    virtual void SetHeader(const wxString& name, const wxString& value) = 0;
+    /**
```
vadz, line 281: Extremely minor:
```suggestion

    /**
```

interface/wx/webview.h
```diff
@@ -297,6 +450,81 @@ class wxWebViewHandler
         @since 3.1.5
     */
     virtual wxString GetSecurityURL() const;
+
+    /**
+        When using the edge backend handler urls are https urls with a
+        virtual host. As default @c name.wxsite is used as the virtual hostname.
+        If you customize this host use a non existing site (ideally a reserved
```
vadz, line 457: Minor too but helps readability a bit IMHO
```suggestion
        If you customize this host, use a non existing site (ideally a reserved
```

interface/wx/webview.h
```diff
@@ -297,6 +450,81 @@ class wxWebViewHandler
...
+        The following pseudo code demonstrates a typical implementation:
+        @code
+        void StartRequest(const wxWebViewHandlerRequest& request,
+                              wxSharedPtr<wxWebViewHandlerResponse> response) wxOVERRIDE
+        {
+            // Set common headers allowing access from XMLHTTPRequests()
+            response->SetHeader("Access-Control-Allow-Origin", "*");
+            response->SetHeader("Access-Control-Allow-Headers", "*");
+
+            // Handle options request
```
vadz, line 496: Do we _have_ to do this? I.e. do we get `OPTIONS` request even if we don't generate it ourselves? Or is for demonstration purposes only (but then why does the same do it too)?

vadz (approved):
Thanks, looks excellent, as usual, but just a question: is it feasible to implement this with GTK webkit later or does it not support this functionality at all?

interface/wx/webview.h
TcT2k: It depends on the context, but a `XMLHTTPRequest` generates (regardless of the requests HTTP method) a `OPTIONS` request from the browser to check cross site security.

TcT2k:
> This is going to result in link errors if this header is included more than once, so it'd be better to move it into some existing file if possible.

I've reimplemented the few `IUnknown` methods required inline as the `comimpl.h` macros don't support exported classes and this was an easier fix for now.

Additionally I've integrated your documentation suggestions.

> Thanks, looks excellent, as usual, but just a question: is it feasible to implement this with GTK webkit later or does it not support this functionality at all?

Thanks, from the API documentation I looked at, everything except access to the requests HTTP body (e.g. POST data) should be possible with GTK webkit. As stated I won't implement it at this time. I might add it in the future, if I find the time and nobody beats me too it.

vadz:
Thanks again and it's definitely not a problem to not add webkit implementation right now, I just wanted to know if it were possible in the future, thanks for confirming this.

I'll merge it when the CI builds finish.

PR #22798, opened 2022-09-14
Fix more double negatives used with 'neither' in mainly non-API docs

lanurmi:
In many cases it should be 'either'.

No changes to actual code.

Complements #22723, which focused on API docs and comments in C++ code.

### DIFF
--- a/docs/doxygen/overviews/unicode.h
+++ b/docs/doxygen/overviews/unicode.h
@@ -234,10 +234,10 @@ The problems can be separated into three broad classes:
 Because of the need to support implicit conversions to both @c char and
 @c wchar_t, wxString implementation is rather involved and many of its operators
 don't return the types which they could be naively expected to return.
-For example, the @c operator[] doesn't return neither a @c char nor a @c wchar_t
-but an object of a helper class wxUniChar or wxUniCharRef which is implicitly
-convertible to either. Usually you don't need to worry about this as the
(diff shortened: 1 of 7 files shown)


docs/doxygen/overviews/unicode.h
```diff
@@ -234,7 +234,7 @@ The problems can be separated into three broad classes:
 Because of the need to support implicit conversions to both @c char and
 @c wchar_t, wxString implementation is rather involved and many of its operators
 don't return the types which they could be naively expected to return.
-For example, the @c operator[] doesn't return neither a @c char nor a @c wchar_t
+For example, the @c operator[] returns neither a @c char nor a @c wchar_t
 but an object of a helper class wxUniChar or wxUniCharRef which is implicitly
 convertible to either. Usually you don't need to worry about this as the
```
imciner2, line 239:
```suggestion
For example, the @c operator[] doesn't return either a @c char or a @c wchar_t
and instead returns an object of a helper class wxUniChar or wxUniCharRef that is implicitly
convertible to either. Usually you don't need to worry about this as the
```
I think this is also clearer in the other framing (either/or) with some modifications to the surrounding sentence.

lanurmi: Thanks! I'll squash these changes into the first commit in a moment.

vadz:
Thanks for the changes and for the reviews!

PR #22809, opened 2022-09-18
CMake: Restore wx-config variables

### LINKED ISSUE #22806: wx-config broken after update
**Describe the problem**

After updating my wx revision to the latest, with an out-of-tree CMake build, wx-config started outputting:

```
  *** Error: Directory ''
             no longer exists.
```

Tracked it down to: https://github.com/wxWidgets/wxWidgets/commit/ad75a77bc6deb0e3850b43b3e9384bba819cb3fe 

Without the `    set(abs_top_srcdir ${CMAKE_CURRENT_SOURCE_DIR})` line, then the path is not generated in `wx-config`.

cc @MaartenBent

MaartenBent:
This was erroneously removed in ad75a77bc6 (CMake: Cleanup the wx_configure_script macro, 2021-10-27).

Fixes #22806

This should also be backported to 3.2.

### DIFF
--- a/build/cmake/config.cmake
+++ b/build/cmake/config.cmake
@@ -12,6 +12,10 @@ file(MAKE_DIRECTORY ${wxCONFIG_DIR})
 set(TOOLCHAIN_FULLNAME ${wxBUILD_FILE_ID})
 
 macro(wx_configure_script input output)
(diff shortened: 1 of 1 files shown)


vadz:

PR #22811, opened 2022-09-18
Update subsurface position when a GL canvas is sized.

cth103:
I'm using a `wxGLCanvas` which ends up being laid out towards the right-hand-side of the screen and I find that without this change the GL canvas is rendered much too far left on systems that use Wayland.  I could be missing the point entirely but I thought I'd submit it, as it works for me!

### DIFF
--- a/include/wx/unix/glegl.h
+++ b/include/wx/unix/glegl.h
@@ -122,6 +122,7 @@ public:
     wl_subcompositor *m_wlSubcompositor;
     wl_callback *m_wlFrameCallbackHandler;
     wl_egl_window *m_wlEGLWindow;
+    wl_subsurface *m_wlSubsurface;
 
 private:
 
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for the fix, but I'd like to understand the problem a bit better before applying it -- does this happen in your program only or in wx own samples as well? If it doesn't happen the latter, could you please try to reproduce the problem in one of them, e.g. the cube one?

cc @swt2c

vadz:
I'd like to understand what to do about this one -- I don't see any problem here with Wayland and the cube sample, so I can't do much. @cth103 Could you please check if the cube sample works correctly for you (without your change)?

cth103:
Sorry for the delay in getting back to you. The sample does indeed work fine without the fix. I started trying to make a minimal repro sample but haven't got one yet... Do close this if it's easier and I can resubmit if/when I have code to reproduce.

cth103:
Having looked again - I think I do see the problem with the sample.  If I start `cube`, then double-click its title bar to maximise the window, I see a gap between the window edge and the OpenGL canvas.  With the fix mentioned in the branch, this gap is gone.

vadz:
You're right, thanks, I see it too now and I can confirm that your change fixes this, so I'm going to apply it.

Thanks again!

cth103:
Thank you for your time!

PR #22813, opened 2022-09-20
Fix for missing X11 libraries when linking wxGTK2 statically

vadz:
We must always use them with wxGTK2, so don't check wxHAVE_GDK_X11 for
it, otherwise we're never going to include X11 libraries when using it,
but they are always required.

### DIFF
--- a/build/cmake/toolkit.cmake
+++ b/build/cmake/toolkit.cmake
@@ -116,9 +116,10 @@ if(WXGTK)
 endif()
 
 # We need X11 for non-GTK Unix ports (X11, Motif) and for GTK with X11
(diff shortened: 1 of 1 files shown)


vadz:
@MaartenBent I'm reasonable sure that this is the right thing to do, but I'd still appreciate your confirmation, of course.

FWIW I also thought of always setting `wxHAVE_GDK_X11` for GTK2, but this didn't seem right to me, and it's only used here, so I preferred to make this fix instead.

MaartenBent:
Looks fine to me. Maybe it can even be simplified to `(WXX11 OR WXMOTIF OR WXGTK2 OR wxHAVE_GDK_X11)`.

PR #22814, opened 2022-09-21
Use output directory under wx directory in the build tree

vadz:
CMAKE_CURRENT_BINARY_DIR is the same as CMAKE_BINARY_DIR when building wx directly, but they can be different when including this makefile from a super-project using add_subdirectory() and it makes more sense to use the former.

This also fixes wx-config symlink created in such build, as it the link was broken before because it was created under CMAKE_CURRENT_BINARY_DIR but pointed to a file in a different CMAKE_BINARY_DIR.

---

This is another CMake fix that I'm relatively confident in, for the reasons described above, but a confirmation from @TcT2k and/or @MaartenBent would still be welcome. TIA!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ include(build/cmake/policies.cmake NO_POLICY_SCOPE)
 
 # Initialize variables for quick access to wx root dir in sub dirs
 set(wxSOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
-set(wxBINARY_DIR ${CMAKE_BINARY_DIR})
+set(wxBINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
 set(wxOUTPUT_DIR ${wxBINARY_DIR}/lib)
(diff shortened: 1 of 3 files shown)


MaartenBent:
Maybe we should also replace the other usage of `CMAKE_BINARY_DIR` in `lib/webview/CMakeLists.txt` with `wxBINARY_DIR`.

There is also `CMAKE_CURRENT_SOURCE_DIR` and `CMAKE_CURRENT_BINARY_DIR` in `config.cmake`, for consistency those could also use `wxSOURCE_DIR` and `wxBINARY_DIR`.

vadz:
I'm actually not really sure why do we use our own variables instead of the CMake standard ones, but I've changed these occurrences too for consistency as you suggested now.

The only remaining occurrences of the standard variables are now in Cotire module which I'm not going to touch (BTW, I've tried to use `-DwxBUILD_PRECOMP=COTIRE` just to see what it would do, but it only resulted in errors...).

PR #22815, opened 2022-09-21
Add an example of using wxrc docs with CMake add_subdirectory()

vadz:
There are a couple of non-trivial points when using wxrc in this configuration, so document using $<TARGET_FILE:wxrc> and having to use "DEPENDS wxrc" in this case.

----

@MaartenBent @TcT2k Please correct me if I'm saying anything stupid here.


TcT2k (approved):
Looks good to me and should be helpful to wx+cmake+xrc users

Issue #22818, opened 2022-09-22
Zoomfactor returned by wxZoomGestureEvent::GetZoomFactor() can be incorrect.


s-murphree:
In msw/window.cpp, wxWindowMSW::HandleZoomGesture(), the fingerDistance parmeter is WXDORD (unsigned long) and s_initialFingerDistance is a signed integer.  If the fingerDistance parameter is fed from a negative number, the high bit is set.  Upon the fist gesture event, s_initialFingerDistance is set with (int)fingerDistance and s_initialFingerDistance become negative, which is actually what is desired.  However, the zoom factor is calculated with the unsigned long fingerDistance and that makes the zoom factor a rather large negative number.  The zoom factor calculation should use fingerDistance cast to an integer.

![image](https://user-images.githubusercontent.com/114183135/191823358-ccbf003d-cd45-4806-9853-41ac00727f99.png)

vadz:
Thanks for reporting this and it indeed looks like we ought to handle it as an int if it can be negative.

Please don't hesitate [making a PR](https://www.wxwidgets.org/develop/how-to-submit-patches/) with your changes.

PR #22819, opened 2022-09-22
Update window.cpp

s-murphree:
This PR proposes to correct error reported on issue #22818


vadz:
Applied to both master and 3.2 now, so closing.

s-murphree:
Sorry about the terse comment.  I figured the original ticket comment would move over.  This was my first PR so if there is a next time, it won't be so sloppy.

vadz:
No problem and thanks again for fixing this!

PR #22823, opened 2022-09-24
Simplify wxGrid::SetCurrentCell() implementation

AliKet:
Stop drawing things from outside the paint event handler and just `RefreshBlock()` the current cell under all platforms.

This is much more clear and concise.

I did not see any flickering or regression while testing this under **Windows 10**
(if my testing setup is correct of course)

**Edit:** we can always put `ShouldRefresh()` inside `RefreshBlock()` for safety and do nothing if we shouldn't refresh the blocks right now ?

### DIFF
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -6436,51 +6436,13 @@ bool wxGrid::SetCurrentCell( const wxGridCellCoords& coords )
 
         if ( IsVisible( m_currentCellCoords, false ) )
         {
-#if defined(__WXOSX__) || defined(__WXGTK3__)
             RefreshBlock(m_currentCellCoords, m_currentCellCoords);
-#else
-            wxGridWindow* prevGridWindow = CellToGridWindow(m_currentCellCoords);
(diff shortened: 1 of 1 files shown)


vadz:
I don't think we should change `ShouldRefresh()` checks, they're used for some code which does more than just refreshing a window and it seems tricky to change this, so let's leave it like this. But I'll merge this change in, of course, thanks!

AliKet:
> I don't think we should change `ShouldRefresh()` checks, they're used for some code which does more than just refreshing a window and it seems tricky to change this

Sorry, I mean doing this inside `RefreshBlock()`:
```
wxGrid::RefreshBlock(...)
{
    if ( !ShouldRefresh()  )
        return;
    ...
}
```

vadz:
Yes, I understand, but I think it would be confusing to check for `ShouldRefresh()` both inside and outside the function and as we have to keep the checks outside, it's better to just document (i.e. add a comment explaining it) that this function shouldn't be called if we're inside a batch operation.

PR #22824, opened 2022-09-26
Fix wxGrid::SetCellvalue() incorrectly refreshes the wrong rectangle

AliKet:
First commit fixes what looks like a copy and past error to me!

### DIFF
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -10800,9 +10800,12 @@ void wxGrid::SetCellValue( int row, int col, const wxString& s )
         m_table->SetValue( row, col, s );
         if ( ShouldRefresh() )
         {
+            wxGridWindow* const gridWindow = CellToGridWindow(row, col);
+
             wxRect rect( CellToRect( row, col ) );
-            CalcScrolledPosition(rect.x, rect.y, &rect.x, &rect.y);
-            m_gridWin->Refresh( false, &rect );
+            CalcGridWindowScrolledPosition(rect.x, rect.y, &rect.x, &rect.y, gridWindow);
+            rect.Offset(-GetGridWindowOffset(gridWindow));
+            gridWindow->Refresh( false, &rect );
         }
 
         if ( m_currentCellCoords.GetRow() == row &&
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -10800,11 +10800,8 @@ void wxGrid::SetCellValue( int row, int col, const wxString& s )
         m_table->SetValue( row, col, s );
         if ( ShouldRefresh() )
         {
-            int dummy;
             wxRect rect( CellToRect( row, col ) );
-            rect.x = 0;
-            rect.width = m_gridWin->GetClientSize().GetWidth();
-            CalcScrolledPosition(0, rect.y, &dummy, &rect.y);
+            CalcScrolledPosition(rect.x, rect.y, &rect.x, &rect.y);
             m_gridWin->Refresh( false, &rect );
         }
 


AliKet:
I think the dummy variable is unnecessary as the functions already accept `NULL`
as a valid parameter (for the unused parameter) i.e.:
```
index c556c6566f..bb4202d9ce 100644
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -3745,9 +3745,8 @@ wxArrayInt wxGrid::CalcRowLabelsExposed( const wxRegion& reg, wxGridWindow *grid
 
         // logical bounds of update region
         //
-        int dummy;
-        CalcGridWindowUnscrolledPosition( 0, r.GetTop(), &dummy, &top, gridWindow );
-        CalcGridWindowUnscrolledPosition( 0, r.GetBottom(), &dummy, &bottom, gridWindow );
+        CalcGridWindowUnscrolledPosition( 0, r.GetTop(), NULL, &top, gridWindow );
+        CalcGridWindowUnscrolledPosition( 0, r.GetBottom(), NULL, &bottom, gridWindow );
 
         // find the row labels within these bounds
         const int rowFirst = internalYToRow(top, gridWindow);
@@ -3799,9 +3798,8 @@ wxArrayInt wxGrid::CalcColLabelsExposed( const wxRegion& reg, wxGridWindow *grid
 
         // logical bounds of update region
         //
-        int dummy;
-        CalcGridWindowUnscrolledPosition( r.GetLeft(), 0, &left, &dummy, gridWindow );
-        CalcGridWindowUnscrolledPosition( r.GetRight(), 0, &right, &dummy, gridWindow );
+        CalcGridWindowUnscrolledPosition( r.GetLeft(), 0, &left, NULL, gridWindow );
+        CalcGridWindowUnscrolledPosition( r.GetRight(), 0, &right, NULL, gridWindow );
 
         // find the cells within these bounds
         //
@@ -10594,11 +10592,11 @@ wxGrid::AutoSizeColOrRow(int colOrRow, bool setAsMin, wxGridDirection direction)
             }
             else
             {
-                int cw, ch, dummy;
+                int cw, ch;
                 m_gridWin->GetClientSize( &cw, &ch );
                 wxRect rect ( CellToRect( 0, colOrRow ) );
+                CalcScrolledPosition(rect.x, 0, &rect.x, NULL);
                 rect.y = 0;
-                CalcScrolledPosition(rect.x, 0, &rect.x, &dummy);
                 rect.width = cw - rect.x;
                 rect.height = m_colLabelHeight;
                 GetColLabelWindow()->Refresh( true, &rect );
@@ -10616,11 +10614,11 @@ wxGrid::AutoSizeColOrRow(int colOrRow, bool setAsMin, wxGridDirection direction)
         SetRowSize(colOrRow, extentMax);
         if ( ShouldRefresh() )
         {
-            int cw, ch, dummy;
+            int cw, ch;
             m_gridWin->GetClientSize( &cw, &ch );
             wxRect rect( CellToRect( colOrRow, 0 ) );
+            CalcScrolledPosition(0, rect.y, NULL, &rect.y);
             rect.x = 0;
-            CalcScrolledPosition(0, rect.y, &dummy, &rect.y);
             rect.width = m_rowLabelWidth;
             rect.height = ch - rect.y;
             m_rowLabelWin->Refresh( true, &rect );
```

vadz:
Sorry, just to make sure I'm not missing something here: what is the copy/paste error? It looks like the old code refreshed the entire row, which definitely looks like too much to me, but not, strictly, speaking "wrong". Is there some other problem here I'm missing?

Also, couldn't we just reuse `RefreshBlock()` here?

AliKet:
Thanks for the review.

> Sorry, just to make sure I'm not missing something here: what is the copy/paste error? It looks like the old code refreshed the entire row, which definitely looks like too much to me, but not, strictly, speaking "wrong". Is there some other problem here I'm missing?
> 

I meant to say _just refresh the affected cell instead of refreshing the whole row_, but apparently I chose the **wrong** expression for that! Sorry.

> Also, couldn't we just reuse `RefreshBlock()` here?

Yes, `RefreshBlock()` works correctly. thanks for pointing it out to me.

While writing this i realized that the old code written like that to account for overflowing cells. in this case, we need to refresh the entire row of course, and the old code is correct then. (but adding a comment explaining this is always appreciated).

Still we need to make `SetCellValue()` work correctly with the presence of frozen grid windows.

vadz:
> I meant to say _just refresh the affected cell instead of refreshing the whole row_, but apparently I chose the **wrong** expression for that! Sorry.

No problem at all, I just wanted to be sure I wasn't missing something important.
 
> > Also, couldn't we just reuse `RefreshBlock()` here?
> 
> Yes, `RefreshBlock()` works correctly. thanks for pointing it out to me.
> 
> While writing this i realized that the old code written like that to account for overflowing cells. in this case, we need to refresh the entire row of course, and the old code is correct then. (but adding a comment explaining this is always appreciated).

I'm not so sure about this, doesn't `CellToRect()` already take this into account? If not, maybe it should? But I don't have time to test it right now and it's not really obvious from just looking at the code :-(
 
> Still we need to make `SetCellValue()` work correctly with the presence of frozen grid windows.

Again, maybe I'm overoptimistic, but doesn't `RefreshBlock()` already handle this?

AliKet:
> > While writing this i realized that the old code written like that to account for overflowing cells. in this case, we need to refresh the entire row of course, and the old code is correct then. (but adding a comment explaining this is always appreciated).
> 
> I'm not so sure about this, doesn't `CellToRect()` already take this into account? If not, maybe it should? But I don't have time to test it right now and it's not really obvious from just looking at the code :-(
> 

AFAICT `CellToRect()` knows about merged cells (and handle them correctly)
but not cells that can overflow.

> > Still we need to make `SetCellValue()` work correctly with the presence of frozen grid windows.
> 
> Again, maybe I'm overoptimistic, but doesn't `RefreshBlock()` already handle this?

The answer to this is the same as above. i.e. `RefreshBlock()` although it handles  frozen windows correctly, it fails with overflowing cells.

here is a patch you can apply to test and see `SetCellValue()` behaviour in the
griddemo (Edit->Edit current cell):
```
diff --git a/samples/grid/griddemo.cpp b/samples/grid/griddemo.cpp
index dd6309dbe9..a22941973c 100644
--- a/samples/grid/griddemo.cpp
+++ b/samples/grid/griddemo.cpp
@@ -1436,7 +1436,9 @@ void GridFrame::ClearGrid( wxCommandEvent& WXUNUSED(ev) )
 
 void GridFrame::EditCell( wxCommandEvent& WXUNUSED(ev) )
 {
-    grid->EnableCellEditControl();
+//    grid->EnableCellEditControl();
+    const wxGridCellCoords&  coords = grid->GetGridCursorCoords();
+    grid->SetCellValue(coords.GetRow(), coords.GetCol(), "wxWidgets");
 }
 
 void GridFrame::SetCornerLabelValue( wxCommandEvent& WXUNUSED(ev) )
```

AliKet:
@vadz  I've created this #22826 which addresses the problem this PR tries to solve
and which will be proven useful when we add transparent selection to `wxGrid` later.

TIA for any review on that PR.

PR #22826, opened 2022-09-27
Add RefreshArea() member function to wxGrid

AliKet:
The idea here is to issue a refresh request on the `wxGrid` object itself and rely on the underlying toolkit to dispatch the request to the child windows intersecting the refreshed rectangle. Notice that there is no performance penalty here because the paint handler of the grid itself is trivial (does nothing in fact) and the number of its children is nine (09) at most, and this is not a problem at all (in practice) to know which part needs to be refreshed from the request.

Tested and works correctly under Windows 10 and Linux. but i have no Mac machine to test this on, sorry


include/wx/generic/grid.h
```diff
@@ -1481,6 +1481,22 @@ struct WXDLLIMPEXP_CORE wxGridSizesInfo
     wxUnsignedToIntHashMap m_customSizes;
 };
 
+// ----------------------------------------------------------------------------
+// wxGridArea
```
vadz, line 1485: Do we have to make this public? I.e. do we plan make `RefreshArea()` as part of public API? If yes, both the enum and the function should be documented. If not, could you please move the enum to `include/wx/generic/private/grid.h`?

include/wx/generic/grid.h
```diff
@@ -1481,6 +1481,22 @@ struct WXDLLIMPEXP_CORE wxGridSizesInfo
...
+// ----------------------------------------------------------------------------
+enum wxGridArea
+{
+    Corner      = 0x001, // m_cornerLabelWin
```
vadz, line 1489: The elements of the enum should have `wxGridArea_` prefix as otherwise they will inevitably conflict with something with the names as common as that. This is especially important if they remain in the public header, but they should have a prefix even if they're private because an identifier like `All` just can't be used in the global scope (but if they were private we could put the entire enum into a namespace perhaps).

src/generic/grid.cpp
```diff
@@ -5199,8 +5197,7 @@ void wxGrid::RefreshAfterRowPosChange()
     }
 
     // and make the changes visible
-    m_rowLabelWin->Refresh();
-    m_gridWin->Refresh();
+    RefreshArea(wxGridArea::Cells | wxGridArea::RowLabels);
```
vadz, line 5200: Non-class enum elements don't need to be qualified.

src/generic/grid.cpp
```diff
@@ -5889,6 +5886,56 @@ void wxGrid::RefreshBlock(int topRow, int leftCol,
     }
 }
 
+void wxGrid::RefreshArea(int area, wxRect* rect, bool erasebg)
+{
+    if ( area == wxGridArea::All )
+    {
+        wxScrolledCanvas::Refresh(erasebg);
```
vadz, line 5893: Does this not take `rect` into account intentionally? If so, there should be a comment here explaining why as this doesn't seem obvious.

include/wx/generic/grid.h
```diff
@@ -1693,6 +1709,19 @@ class WXDLLIMPEXP_CORE wxGrid : public wxScrolledCanvas
...
+    //
+    // This function is more convenient/useful (from performance point of view)
+    // to use if the rectangle is already known/calculated, because:
+    // - it transparently handles frozen grid windows for us
+    // - it can deduce the labels area to be refreshed (if we asked to) from _rect_ which
+    //   represents the cells area. e.g. RefreshArea(wxGridArea::Cells | wxGridArea::Labels, ...);
+    // - there is no need to iterate over the selected blocks one more time to know which part of
+    //   the grid needs to be refreshed (for instance: transparent selection rely on this function
+    //   for performance reason)
+    void RefreshArea(int area, wxRect* rect = NULL, bool erasebg = true);
```
vadz, line 1723: Do we really need `erasebg` here? I think we can always erase it, at least I don't see any valid reason not to do it and it shouldn't really have any ill effects.

More importantly, does it really make sense to allow combining different areas _and_ specifying the rectangle? I.e. isn't it valid for just a single area only?

vadz (commented):
Sorry, I find the new function very confusing. Maybe it's just me, but it also looks like it's only used in 2 cases:

1. Refresh several areas completely.
2. Refresh part of a single area.

And this seems to indicate to me that we need 2 different functions here and not one doing everything. Am I missing some reason we need to do all in a single function?

AliKet:
> Sorry, I find the new function very confusing. Maybe it's just me, but it also looks like it's only used in 2 cases:
> 
>     1. Refresh several areas completely.
> 
>     2. Refresh part of a single area.

First, thanks for the review.
Second, here is how `RefreshArea()` works illustrated below:

Screenshot_2022-09-27_17-16-49

In the above image, the nine (09) child windows are all there:
`m_cornerLabelWin, m_{row,col}LabelWin, m_{row,col}FrozenLabelWin,  m_gridwin, m_frozenCornerGridWin, m_frozen{Row,Col}GridWin`

Suppose now that the selection rectangle `rect` surrounded by the green box is already calculated, then calling `RefreshArea(wxGridArea::Cells, &rect)` will correctly refreshes `m_gridwin, m_frozenCornerGridWin, m_frozen{Row,Col}GridWin` windows, **but only the intersection areas** between these windows and `rect`.
Notice that the `rect` as passed to `RefreshArea` has the big black dot as origin and what we should refresh is the offset of this rectangle, i.e. `rect.Offset(GetRowLabelSize(), GetColLabelSize())`

In addition, if the labels need to be refreshed to reflect the selected state, the following `RefreshArea(wxGridArea::Cells | wxGridArea::Labels, &rect)` will also refreshes the red box (which means two windows will be refreshed `m_colLabelWin` and `m_colFrozenLabelWin`). The same for the blue box (`m_rowLabelWin`, `m_rowFrozenLabelWin` will be refreshed).

For me i find this feature extremely useful, i.e. no need to test for any frozen part at all.

include/wx/generic/grid.h
AliKet: I'll use `wxGA_` as prefix then for shorter names unless you prefer the suggested one ?

include/wx/generic/grid.h
AliKet: > Do we really need `erasebg` here? I think we can always erase it, at least I don't see any valid reason not to do it and it shouldn't really have any ill effects.
> 

I agree with this and I'll remove it in the next commits

> More importantly, does it really make sense to allow combining different areas _and_ specifying the rectangle? I.e. isn't it valid for just a single area only?

I understand the confusion here, i.e:
One can effortlessly understand the meaning of the following calls:
```
RefreshArea(wxGA_Cells, &rect);
RefreshArea(wxGA_RowLabels, &rect);
RefreshArea(wxGA_ColLabels, &rect);
```
But the next ones are not !!
```
RefreshArea(wxGA_Labels, &rect); // wxGA_Labels means wxGA_RowLabels | wxGA_ColLabels 
RefreshArea(wxGA_Cells | wxGA_Labels, &rect);
```
What does `rect` represent here ? Well:
`RefreshArea(wxGA_Labels, &rect);` does not make sense and such call should be rejected or at least warn about it.

Whereas:
`RefreshArea(wxGA_Cells | wxGA_Labels, &rect);` means: `rect` is valid for `wxGA_Cells` area, and for the labels the rectangles are to be deduced/calculated from `rect`.

Maybe we should add more comments about the usage of this function to alleviate the confusion about the meaning of the parameter `rect` ?

Notice that the following calls are totally valid and shouldn't be there any confusion about them:
```
RefreshArea(wxGA_Labels); // refresh the entire wxGA_RowLabels area and wxGA_ColLabels area
RefreshArea(wxGA_Cells | wxGA_Labels); // as above, plus refresh the cells area too
```

Hope this is clear now ?

vadz: I agree that having a function refreshing several areas entirely at once is convenient, but it seems that you also agree that using a rectangle and a combination of the areas doesn't make sense. So shouldn't there be 2 functions:

```cpp
void RefreshPartOfArea(wxGridArea area, const wxRect& r); // Refresh part of one area.
void RefreshArea(int areas); // Refresh one or more areas entirely.
```

This would be more clear IMO, what do you think?

AliKet: Yes this is what I've done now and I'll push the changes soon.

src/generic/grid.cpp
```diff
@@ -5889,6 +5886,69 @@ void wxGrid::RefreshBlock(int topRow, int leftCol,
...
+    if ( (areas & wxGA_Corner) != 0 )
+        m_cornerLabelWin->Refresh();
+
+    int cw, ch;
+    GetClientSize(&cw, &ch);
+
+    // Cells area
+    if ( (areas & wxGA_Cells) != 0 )
+    {
+        const wxRect rect = wxRect(GetRowLabelSize(), GetColLabelSize(), cw, ch);
```
vadz, line 5911: Pretty minor, but this could be just
```suggestion
        const wxRect rect(GetRowLabelSize(), GetColLabelSize(), cw, ch);
```
(if this is changed, it should be changed below too, for consistency).

src/generic/grid.cpp
```diff
@@ -5889,6 +5886,69 @@ void wxGrid::RefreshBlock(int topRow, int leftCol,
...
+
+        wxScrolledCanvas::Refresh(true, &rect);
+    }
+}
+
+void wxGrid::RefreshArea(wxGridArea area, const wxRect& rect)
+{
+    wxASSERT_MSG( !rect.IsEmpty(), "RefreshArea expects a valid rectangle to refresh" );
+
+    if ( area == wxGA_Corner )
```
vadz, line 5937: This function doesn't seem to be ever called with this area, is it?

src/generic/grid.cpp
```diff
@@ -5889,6 +5886,69 @@ void wxGrid::RefreshBlock(int topRow, int leftCol,
...
+        // corner area instead
+        m_cornerLabelWin->Refresh();
+    }
+
+    if ( area == wxGA_Cells )
+        wxScrolledCanvas::Refresh(true, &rect);
+    else if ( area == wxGA_RowLabels && GetRowLabelSize() > 0 )
+        wxScrolledCanvas::Refresh(true, &rect);
+    else if ( area == wxGA_ColLabels && GetColLabelSize() > 0 )
+        wxScrolledCanvas::Refresh(true, &rect);
```
vadz, line 5949: Written like this, it seems natural to ask ourselves if we really need to have this function? As it basically just does exactly the same thing as `Refresh()`, and only adds checks for having row/column labels. I think it would be simpler and clearer to just call `Refresh()` directly in the code.

src/generic/grid.cpp
```diff
@@ -9117,26 +9161,28 @@ void wxGrid::SetCornerLabelTextOrientation( int textOrientation )
 
 void wxGrid::SetRowLabelValue( int row, const wxString& s )
 {
-    if ( m_table )
+    if ( m_table && m_rowLabelWidth > 0 )
```
vadz, line 9164: Sorry, this is wrong, setting the value should work, i.e. should update the table, even if the labels are currently hidden.

src/generic/grid.cpp
```diff
@@ -9117,26 +9161,28 @@ void wxGrid::SetCornerLabelTextOrientation( int textOrientation )
...
     {
         m_table->SetRowLabelValue( row, s );
         if ( ShouldRefresh() )
```
vadz, line 9167: OTOH it would be fine to add a check for row labels being shown (BTW, if you wish to add `bool AreRowLabelsShown()` instead of comparing `m_rowLabelWidth` with 0, please do it) here.

include/wx/generic/grid.h
```diff
@@ -88,6 +88,20 @@ enum wxGridRenderStyle
                           wxGRID_DRAW_BOX_RECT
 };
 
+// Flags to pass to wxGrid::RefreshArea()
```
vadz, line 91: We still need to decide whether this is going to be public (and then it needs to be documented) or not (and then it should be moved to a private header, which will be simpler to do if we remove the single area overload).

vadz (changes requested):
Thanks for the changes, they make things clearer, but now it seems even more obvious that we don't actually need a special function for a single area refresh: it's basically only used once and it's more clear to just do it there.

The other overload is, of course, nice and useful to have.

src/generic/grid.cpp
AliKet: Actually the final version of the function is this:

```
void wxGrid::RefreshArea(wxGridArea area, wxRect& rect)
{
    wxASSERT_MSG( !rect.IsEmpty(), "RefreshArea expects a valid rectangle to refresh" );

    if ( area == wxGA_Cells )
    {
        rect.Offset(GetRowLabelSize(), GetColLabelSize());
    }
    else if ( area == wxGA_RowLabels && GetRowLabelSize() > 0 )
    {
        rect.x = 0;
        rect.width = m_rowLabelWidth;
        rect.Offset(0, GetColLabelSize());
    }
    else if ( area == wxGA_ColLabels && GetColLabelSize() > 0 )
    {
        rect.y = 0;
        rect.height = m_colLabelHeight;
        rect.Offset(GetRowLabelSize(), 0);
    }

    wxScrolledCanvas::Refresh(true, &rect);
}
```

i.e. we pass the rectangle by non-const reference and apply the invariants inside the function which will make the code more clearer (and robust?) at the call site.

And in fact, this function is called by SetCellValue(), SetRowLabelValue() and SetColLabelValue().
And will be used later (when we add transparent selection).

src/generic/grid.cpp
AliKet: Oops, sorry!

include/wx/generic/grid.h
AliKet: Sorry again! I think the enum doesn't need to be public and I'll move it to the private header again.

src/generic/grid.cpp
AliKet: Removed now

src/generic/grid.cpp
vadz: Sorry, passing `rect` by non-const reference is _definitely_ a bad idea, nobody expects it to be modified inside this function, which looks like a `Refresh()` variant (and this one never modifies the rectangle passed to it), please let's not do this.

And I still don't understand the purpose of all this at all. E.g. for `wxGA_Cells` case the function seems to take a rectangle in `wxGridWindow` coordinates, which it then offsets by the grid window position and passes to `Refresh()`. How is this better than just calling `Refresh()` on `wxGridWindow` itself directly?

AliKet: > And I still don't understand the purpose of all this at all. E.g. for `wxGA_Cells` case the function seems to take a rectangle in `wxGridWindow` coordinates, which it then offsets by the grid window position and passes to `Refresh()`. How is this better than just calling `Refresh()` on `wxGridWindow` itself directly?

This was true when there were no frozen windows at all. But with the introduction of frozen windows, we have to stop thinking that the cell area is always a simple `wxGridWindow`.

Take for example how `wxGrid::DoSet{Row,Col}Size()` would become ridiculously simpler with `RefreshArea()`:
```
@@ -10085,7 +10123,17 @@ void wxGrid::DoSetRowSize( int row, int height )
 
         int y;
         CalcScrolledPosition(0, GetRowTop(topRow), NULL, &y);
+#if 1
+        int cw, ch;
+        GetClientSize(&cw, &ch);
 
+        wxRect updateRect(0, y, cw - GetRowLabelSize(), ch - y);
+        RefreshArea(wxGA_Cells, updateRect);
+
+        updateRect.y = y;
+        updateRect.height = ch - y;
+        RefreshArea(wxGA_RowLabels, updateRect);
+#else
         if ( topRow < m_numFrozenRows )
         {
             // This row is frozen, refresh the frozen windows.
@@ -10114,6 +10162,7 @@ void wxGrid::DoSetRowSize( int row, int height )
             if ( m_frozenColGridWin )
                 refreshLowerPart(m_frozenColGridWin);
         }
+#endif
     }
 }
 
@@ -10273,7 +10322,22 @@ void wxGrid::DoSetColSize( int col, int width )
 
         int x;
         CalcScrolledPosition(GetColLeft(leftCol), 0, &x, NULL);
+#if 1
+        int cw, ch;
+        GetClientSize(&cw, &ch);
 
+        wxRect updateRect(x, 0, cw - x, ch - GetColLabelSize());
+        RefreshArea(wxGA_Cells, updateRect);
+
+        // Refreshing the native header is unnecessary, as it updates
+        // itself correctly anyhow, and just results in extra flicker.
+        if ( !IsUsingNativeHeader() )
+        {
+            updateRect.x = x;
+            updateRect.width = cw - x;
+            RefreshArea(wxGA_ColLabels, updateRect);
+        }
+#else
         if ( leftCol < m_numFrozenCols )
         {
             FurtherWindowPartRefresher refreshFurtherPart(x);
@@ -10300,6 +10364,7 @@ void wxGrid::DoSetColSize( int col, int width )
             if ( m_frozenRowGridWin )
                 refreshFurtherPart(m_frozenRowGridWin);
         }
+#endif
     }
 }
```

vadz: Could you please explain why can't we use `wxGrid::Refresh()` for doing this?

AliKet: Sure we can and is even more clear and concise.

First, remove the unnecessary block inside `wxGrid::Refresh()`

```
@@ -5707,7 +5704,7 @@ void wxGrid::Refresh(bool eraseb, const wxRect* rect)
     {
         // Refresh to get correct scrolled position:
         wxScrolledCanvas::Refresh(eraseb, rect);
-
+#if 0
         if (rect)
         {
             int rect_x, rect_y, rectWidth, rectHeight;
@@ -5798,6 +5795,7 @@ void wxGrid::Refresh(bool eraseb, const wxRect* rect)
             if ( m_frozenCornerGridWin )
                 m_frozenCornerGridWin->Refresh(eraseb, NULL);
         }
+#endif
     }
 }
``` 
because  `wxScrolledCanvas::Refresh(eraseb, rect)` will correctly refreshes the subwindows intersecting with the rectangle `rect`.

Now, the new `wxGrid::DoSet{Row,Col}Size()` versions look like:

```
@@ -10085,7 +10124,14 @@ void wxGrid::DoSetRowSize( int row, int height )
 
         int y;
         CalcScrolledPosition(0, GetRowTop(topRow), NULL, &y);
+        y += GetColLabelSize();
+#if 1
+        int cw, ch;
+        GetClientSize(&cw, &ch);
 
+        wxRect updateRect(0, y, cw, ch - y);
+        Refresh(true, &updateRect);
+#else
         if ( topRow < m_numFrozenRows )
         {
             // This row is frozen, refresh the frozen windows.
@@ -10114,6 +10160,7 @@ void wxGrid::DoSetRowSize( int row, int height )
             if ( m_frozenColGridWin )
                 refreshLowerPart(m_frozenColGridWin);
         }
+#endif
     }
 }
 
@@ -10273,7 +10320,14 @@ void wxGrid::DoSetColSize( int col, int width )
 
         int x;
         CalcScrolledPosition(GetColLeft(leftCol), 0, &x, NULL);
+        x += GetRowLabelSize();
+#if 1
+        int cw, ch;
+        GetClientSize(&cw, &ch);
 
+        wxRect updateRect(x, 0, cw - x, ch);
+        Refresh(true, &updateRect);
+#else
         if ( leftCol < m_numFrozenCols )
         {
             FurtherWindowPartRefresher refreshFurtherPart(x);
@@ -10300,6 +10354,7 @@ void wxGrid::DoSetColSize( int col, int width )
             if ( m_frozenRowGridWin )
                 refreshFurtherPart(m_frozenRowGridWin);
         }
+#endif
     }
 }
``` 
**The question now is:** what is the advantage of `RefreshArea()` (the one which takes `wxRect`) over just calling `wxGrid::Refresh()` directly ?
**The answer is:** It's more like a helper function that translates the rectangle into the `wxGrid` coordinate space (which `wxGrid::Refresh()` expects) more than anything else.

vadz: IMHO this translation seems less clear than just doing it directly in the code, but if you think it's worth it, why not, but we don't really need the `area` parameter in this one at all then, do we? I.e. this coordinate translation is the same no matter what, isn't it?

Also, could you please document (in a comment before the function) which kind of coordinates exactly does it expect on entry?

AliKet: I'll do these then:
1. keep `RefreshArea(int areas)` function and get rid of the other overload then.
2. Simplify `wxGrid::Refresh()` and use it throughout the code when appropriate.
3. Document (in a comment) how to prepare the rectangle before passing it to `Refresh()`

What do you think ?

AliKet: A question: in the `wxGrid` implementation a call to `wxGrid::Refresh()` is guarded by a call to `ShouldRefresh()` which makes the second call to this function inside `Refresh()` redundant!
so is it ok/safe to remove the second call ? because I am not sure if user codes already call this function or not ?

vadz: > I'll do these then:
> 
>     1. keep `RefreshArea(int areas)` function and get rid of the other overload then.
> 
>     2. Simplify `wxGrid::Refresh()` and use it throughout the code when appropriate.
> 
>     3. Document (in a comment) how to prepare the rectangle before passing it to `Refresh()`
> 
> 
> What do you think ?

Looks to me, thanks!

As for `ShouldRefresh()`, I think it's better to keep it in `Refresh()` because it's just a (probably helpful) internal optimization which doesn't cost much.

src/generic/grid.cpp
```diff
@@ -3702,20 +3687,22 @@ bool wxGrid::Redimension( const wxGridTableMessage& msg )
...
+    {
+        RefreshArea(labelArea);
+        RefreshArea(wxGA_Cells);
 
-    return result;
+        return true;
+    }
+
+    return false;
 }
```
AliKet, line 3706: Oops! we can just `return labelArea != -1;`

AliKet:
@vadz  @paulcor 

FYI in this function [wxGrid::DoSetRowSize()](https://github.com/wxWidgets/wxWidgets/blob/master/src/generic/grid.cpp#L10008) (also _true_ for [wxGrid::DoSetColSize()](https://github.com/wxWidgets/wxWidgets/blob/master/src/generic/grid.cpp#L10186))
the call to `CalcDimensions()` has a side effect of generating a refresh request on the entire widget (under **GTK 3** only) due to [this call](https://github.com/wxWidgets/wxWidgets/blob/master/src/gtk/scrolwin.cpp#L98) (as it calls [gtk_widget_queue_resize()](https://github.com/GNOME/gtk/blob/main/gtk/gtkwidget.c#L6987) internally) which makes this call: `if ( ShouldRefresh() )` useless there (under GTK 3)!

IMHO this is not a problem (i.e. to execute the refresh code resulting in another refresh request to be queued) because GTK 3 is optimized in such a way the drawing is done only once. but the problem is in the **user** not expecting a refresh if `ShouldRefresh()` returns `false` !!

Personally, I haven't had any issues with this, and just wanted to let you know so it can be documented.

Thank you.

vadz:
So should we just move `CalcDimensions()` call inside `if ( ShouldRefresh() )` check? I don't think we need to do it for the currently hidden window (but this needs to be double checked...) and `CalcDimensions()` already does nothing if the updates are disabled anyhow.

AliKet:
> So should we just move `CalcDimensions()` call inside `if ( ShouldRefresh() )` check? I don't think we need to do it for the currently hidden window (but this needs to be double checked...) and `CalcDimensions()` already does nothing if the updates are disabled anyhow.

What I was thinking is to apply something like this:

```
index bca2f1899e..ea1c294f50 100644
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -9939,6 +9939,11 @@ void wxGrid::DoSetRowSize( int row, int height )
 
     CalcDimensions();
 
+    // Notice that if ShouldRefresh() returns true, a refresh request is already
+    // queued for the entire grid under GTK3 by CalcDimensions() above. So save
+    // some valuable CPU cycles and do nothing there.
+#ifndef __WXGTK3__
+
     if ( ShouldRefresh() )
     {
         // We need to check the size of all the currently visible cells and
@@ -9978,6 +9983,8 @@ void wxGrid::DoSetRowSize( int row, int height )
         const wxRect updateRect(0, y, cw, ch - y);
         Refresh(true, &updateRect);
     }
+
+#endif // __WXGTK3__
 }
 
 void wxGrid::SetDefaultColSize( int width, bool resizeExistingCols )
@@ -10082,6 +10089,10 @@ void wxGrid::DoSetColSize( int col, int width )
 
     CalcDimensions();
 
+    // See the comment in DoSetRowSize() for why the following code
+    // is not needed under GTK 3
+#ifndef __WXGTK3__
+
     if ( ShouldRefresh() )
     {
         // This code is symmetric with DoSetRowSize(), see there for more
@@ -10119,6 +10130,8 @@ void wxGrid::DoSetColSize( int col, int width )
         const wxRect updateRect(x, 0, cw - x, ch);
         Refresh(true, &updateRect);
     }
+
+#endif // __WXGTK3__
 }
 
 void wxGrid::SetColMinimalWidth( int col, int width )
```

What do think ?

vadz:
> What do you think ?

I'd rather avoid having platform checks in the generic code.

I also have to wonder how important is it to try to refresh just parts of the window here. If we're using double buffering (and we do), is it really any better than just refreshing everything? And if we do this, we could do it unconditionally and it shouldn't do any harm in wxGTK as 2 refreshes should be coalesced anyhow.

AliKet:
> > What do you think ?
> 
> I'd rather avoid having platform checks in the generic code.
> 

I hate platform checks in the generic code too.

> I also have to wonder how important is it to try to refresh just parts of the window here. If we're using double buffering (and we do), is it really any better than just refreshing everything? And if we do this, we could do it unconditionally and it shouldn't do any harm in wxGTK as 2 refreshes should be coalesced anyhow.

I did some measurments on `wxGrid::DoSetColSize()` (and `wxGrid::DoSetRowSize()`) performence (using a simple glib timer) with and without the following patch:

```
index bca2f1899e..eba6c4b0ab 100644
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -9941,42 +9941,7 @@ void wxGrid::DoSetRowSize( int row, int height )
 
     if ( ShouldRefresh() )
     {
-        // We need to check the size of all the currently visible cells and
-        // decrease the row to cover the start of the multirow cells, if any,
-        // because we need to refresh such cells entirely when resizing.
-        int topRow = row;
-
-        // Note that we don't care about the cells in frozen windows here as
-        // they can't have multiple rows currently.
-        const wxRect rect = m_gridWin->GetRect();
-        int left, right;
-        CalcUnscrolledPosition(rect.GetLeft(), 0, &left, NULL);
-        CalcUnscrolledPosition(rect.GetRight(), 0, &right, NULL);
-
-        const int posLeft = XToPos(left, m_gridWin);
-        const int posRight = XToPos(right, m_gridWin);
-        for ( int pos = posLeft; pos <= posRight; ++pos )
-        {
-            int col = GetColAt(pos);
-
-            int numRows, numCols;
-            if ( GetCellSize(row, col, &numRows, &numCols) == CellSpan_Inside )
-            {
-                // Notice that numRows here is negative.
-                if ( row + numRows < topRow )
-                    topRow = row + numRows;
-            }
-        }
-
-        int y;
-        CalcScrolledPosition(0, GetRowTop(topRow), NULL, &y);
-
-        // Refresh the lower part of the window below the y position
-        int cw, ch;
-        GetClientSize(&cw, &ch);
-
-        const wxRect updateRect(0, y, cw, ch - y);
-        Refresh(true, &updateRect);
+        RefreshArea(wxGA_RowLabels | wxGA_Cells);
     }
 }
 
@@ -10084,40 +10049,7 @@ void wxGrid::DoSetColSize( int col, int width )
 
     if ( ShouldRefresh() )
     {
-        // This code is symmetric with DoSetRowSize(), see there for more
-        // comments.
-
-        int leftCol = col;
-
-        const wxRect rect = m_gridWin->GetRect();
-        int top, bottom;
-        CalcUnscrolledPosition(0, rect.GetTop(), NULL, &top);
-        CalcUnscrolledPosition(0, rect.GetBottom(), NULL, &bottom);
-
-        const int posTop = YToPos(top, m_gridWin);
-        const int posBottom = YToPos(bottom, m_gridWin);
-        for ( int pos = posTop; pos <= posBottom; ++pos )
-        {
-            int row = GetRowAt(pos);
-
-            int numRows, numCols;
-            if ( GetCellSize(row, col, &numRows, &numCols) == CellSpan_Inside )
-            {
-                if ( col + numCols < leftCol )
-                    leftCol = col + numCols;
-            }
-        }
-
-        int x;
-        CalcScrolledPosition(GetColLeft(leftCol), 0, &x, NULL);
-
-        // Refresh the further part of the window to the right (LTR)
-        // or to the left (RTL) of the x position
-        int cw, ch;
-        GetClientSize(&cw, &ch);
-
-        const wxRect updateRect(x, 0, cw - x, ch);
-        Refresh(true, &updateRect);
+        RefreshArea(wxGA_ColLabels | wxGA_Cells);
     }
 }
```

and guess what ?
there is no much gain to try to refresh just parts of the grid compared with
just refreshing the entire cells area plus the affected label (row or col) window.
moreover, the gain is really lost (or vanished) if we are resizing the first row/col labels

So if you (or someone else) don't object to apply the patch above I'll do that ?
(Note that we don't have to change these functions later if/when merged cells in frozen windows are supported.)

vadz:
I think partial refreshing is just generally not that useful when using double buffering as the entire window is going to be blitted to the screen anyhow. I guess there could be some exceptions when drawing itself takes a relatively long time and we want to optimize away redrawing of the unchanged parts even off screen, but normally this shouldn't be the case.

So yes, let's simplify the code here by just refreshing everything.

PR #22827, opened 2022-09-27
Refactor issue templates using HTML comments

Randalphwa:
This PR places the explanatory information in an HTML comment so that the user does not need to manually delete it.

Please note that I did change a bit of the wording in some of these. For example, in the bug_report.md file, I added the suggestion to use the Preview tab to check formatting (it will also show that the comments are automatically removed). I also made some more subtle changes -- I removed the suggested \<t>\</t> around the \`\`\` block since github will display the code section in an indented block without it. Most of the other changes are formatting to force wrap the commented descriptions.


vadz:
Looks good to me, thanks again! Will push soon.

PR #22828, opened 2022-09-27
Remove wxGLAttributes::AddDefaultsForWXBefore31()

Manolo-ES:
OpenGL visual attributes can be passed by wxGLAttibutes or by a list of integers. In the second case a NULL pointer can be passed, which makes wx to ask for some common defaults. These defaults were not exactly the same in each platform, because it was expected that the OGL driver provides such same defaults.
This is proving not true in some modern hardware.
This patch removes wxGLAttributes::AddDefaultsForWXBefore31() and replace its only call at wxGLCanvasBase::ParseAttribList() with a call to wxGLAttibutes::Defaults(), which does ask for those common OGL visual settings.

This PR is a fix for issue PR #22787

### DIFF
--- a/include/wx/glcanvas.h
+++ b/include/wx/glcanvas.h
@@ -162,11 +162,7 @@ public:
     wxGLAttributes& FrameBuffersRGB();
     void EndList(); // No more values can be chained
 
-    // This function is undocumented and cannot be chained on purpose!
-    // To keep backwards compatibility with versions before wx3.1 we add here
-    // the default values used in those versions for the case of NULL list.
-    void AddDefaultsForWXBefore31();
-};
+ };
 
 // ----------------------------------------------------------------------------
 // wxGLContextBase: OpenGL rendering context
--- a/src/common/glcmn.cpp
+++ b/src/common/glcmn.cpp
@@ -158,8 +158,7 @@ bool wxGLCanvasBase::ParseAttribList(const int *attribList,
 
     if ( !attribList )
     {
-        // Default visual attributes used in wx versions before wx3.1
-        dispAttrs.AddDefaultsForWXBefore31();
+        dispAttrs.Defaults();
         dispAttrs.EndList();
         if ( ctxAttrs )
             ctxAttrs->EndList();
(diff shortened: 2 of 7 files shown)


vadz:
Thanks! I like the PRs removing the existing code, so I'd be glad to apply it if it really solves the problem without introducing any compatibility constraints.

We're going to have to keep `AddDefaultsForWXBefore31()` if we backport this to 3.2, even if we don't call it (and it doesn't do anything) because removing it would be an ABI breakage.

Also, I don't know how it happened, but the files modified by this PR have become executable, this will need to be undone when merging.

Manolo-ES:
Some files have the 'executable' attribute. They are the files I worked on Windows and then copied using Nautilus in my Ubuntu 20. I don't know whose to blame, Nautilus or me missing something.

Manolo-ES:
To backport the patch to 3.2, would suffix adding at 'include/wx/glcanvas.h' class wxGLAttributes a line like this?
void AddDefaultsForWXBefore31() {};

vadz:
> Some files have the 'executable' attribute. They are the files I worked on Windows and then copied using Nautilus in my Ubuntu 20. I don't know whose to blame, Nautilus or me missing something.

You need to adjust permissions when copying files from file systems not having them (or having an incompatible set of them) but my recommendation is to never copy files at all but use Git for synchronizing branches between your systems: this is simpler and more robust.

vadz:
Thanks again, I've pushed this now, with some whitespace-only changes and a hopefully more clear commit message (it's important to have informative commit messages and they should be in English), but thinking more about this, I'm really not sure if we should backport this to 3.2 because it would change the behaviour of the existing programs. It could, of course, be argued, that it would change it for the better, but can we totally exclude that some programs would be broken by it? I.e. if any of them really relied on some default attributes that we're now going to overwrite by calling `Defaults()`? I'm not 100% sure about this and at the very least I'd like to wait until 3.3.0 and until people start to actually use this code in master before backporting it.

vadz:
We probably can't backport this to 3.2 as it does change the behaviour and any future changes being discussed don't seem 100% compatible neither. It would be nice to have some simple fix for 3.2 but I'm not sure what could be done there.

vadz:
Closing it as it was applied to master and probably shouldn't be backported to 3.2 in the current state.

PR #22829, opened 2022-09-28
Git mini guide

Manolo-ES:
I've collected some info I think is quite interesting for Git-novices (like me :) )

I hope people reading this to correct and improve this mini-guide, as well as my mediocre English.

I still have doubts, specially on points 7 to 9, such as:
++ Setting the remote. Sometimes 'git@github.com/<github-user>/...' works, while other times 'https://github.com/<github-user>/...' must be used instead.
++ What happens in GitHub when I push the same branch again (with some changes to my previous code) and how it's seen in the PR
++ How to make 'git am PATCH'  to work without downloading the .patch file with my browser.

I don't know where this file fits in. I've trying with docs/contributing, but perhaps it's better to "merge" with https://www.wxwidgets.org/develop/code-repository/

### DIFF
--- a/docs/contributing/Git-mini-guide.md
+++ b/docs/contributing/Git-mini-guide.md
@@ -1,0 +1,196 @@
+This is a mini-quick guide about using Git to share patches for wxWidgets.
+Better and extensive information can be found at
+https://git-scm.com/book/en/v2  and  https://git-scm.com/docs
+
+What is Git?
+------------
+Git is a distributed version control system.
+'Version control' is a system that records changes to a file or set of files
+over time so that you can recall specific versions later.
+'Distributed' means that every developer makes a full copy of the files in the
+repository of the project, including their full history of changes.
+
+What is GitHub?
+---------------
+It's a host of Git repositories. Its site https://github.com is a web interface
+for repositories. wxWidgets is just one of many projects that uses a repository
+that you can browse or download or collaborate with.
+
+1. Install git
+--------------
+Download from https://git-scm.com/downloads the version you like. It can be a
+terminal-based or a GUI one.
+In this guide I use the terminal, just to show the gist of the usage of git.
+
+Git needs to identify you by user/mail because your uploads will carry this info.
+   git config --global user.name "My name"
+   git config --global user.email my@domain.com
+Using '--global' will ask you just this time, no matter what project you work on.
+
+2. Download wxWidgets code... perhaps not yet
+---------------------------------------------
+If you just want the last snap of wx's code, don't plan to modify it, then you
+don't need to create a "divergence" from the "main-stream" of wx. You may
+simply download the repository, compile wx and continue with your business.
+But if you want to contribute, you need a "fork" and a "branch".
+
+3. Create a fork
+----------------
+This is where you tell GitHub you create a divergence to upload code to.
+First, sign in at https://github.com You may need to first create an account.
+I advise to use the same user/email as you did with Git configuration.
+
+A user & password is not enough to upload code. You need other authentication mode.
+You may use SSH keys. You may use PAT.
+See the Git-book and
+https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories#cloning-with-https-urls
+
+
+Now you've signed in, create a fork.
+  Browse to https://github.com/wxWidgets/wxWidgets
+  About the top right corner find and use the "Fork" button. Select "Create a new fork"
+The new fork will be created as github.com/<your_github_name>/wxWidgets.git
+
+This fork is your fork. You don't need another fork to submit changes for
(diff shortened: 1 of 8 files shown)


docs/contributing/Git-mini-guide.md
```diff
@@ -0,0 +1,196 @@
...
+simply download the repository, compile wx and continue with your business.
+But if you want to contribute, you need a "fork" and a "branch".
+
+3. Create a fork
+----------------
+This is where you tell GitHub you create a divergence to upload code to.
+First, sign in at https://github.com You may need to first create an account.
+I advise to use the same user/email as you did with Git configuration.
+
+A user & password is not enough to upload code. You need other authentication mode.
```
oneeyeman1, line 43: It is worse mentioning that those authentications are needed only for pushing the code, i.e. write access.
In the beginning only user/password are sufficient.

docs/contributing/Git-mini-guide.md
```diff
@@ -0,0 +1,196 @@
...
+https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories#cloning-with-https-urls
+
+
+Now you've signed in, create a fork.
+  Browse to https://github.com/wxWidgets/wxWidgets
+  About the top right corner find and use the "Fork" button. Select "Create a new fork"
+The new fork will be created as github.com/<your_github_name>/wxWidgets.git
+
+This fork is your fork. You don't need another fork to submit changes for
+another contribution to wxWidgets.
```
oneeyeman1, line 55: Mention of the second fork is irrelevant.
I believe you will not be able to make it anyway.

docs/contributing/Git-mini-guide.md
```diff
@@ -0,0 +1,196 @@
...
+
+4. Download wxWidgets code, now yes
+-----------------------------------
+Create the directory 'wxFork' (or the name you like).
+
+Option a)
+  git clone --recurse-submodules https://github.com/wxWidgets/wxWidgets.git wxFork
+Option b)
+  git clone --recurse-submodules https://github.com/<your_github_name>/wxWidgets.git wxFork
```
oneeyeman1, line 64: Here you need "option b" only.
You will not get a write access to the main wxWidgets repository anyway.

Just clone your fork.

I believe even Vadim work out of his own fork. ;-)

docs/contributing/Git-mini-guide.md
```diff
@@ -0,0 +1,196 @@
...
+-----------------------
+Not all changed files will be part of your contribution. No compiled files,
+just code files.
+
+  git status -v
+Inspect the output of this command. It will show the files that you changed
+and/or added.
+
+Tell Git the staged files:
+  git add --all
```
oneeyeman1, line 117: You can also use "git add ."

The other useful command is "git diff", especially if you tried to add some printf()/wxMessageBox(), i.e. stuff you did for debugging.

The git diff will produce a "patch-like" output of your current code and the tip of the branch you are currently on. That way you can inspect the code and get rid of all debugging things you put in.

docs/contributing/Git-mini-guide.md
```diff
@@ -0,0 +1,196 @@
...
+or tell one by one:
+  git add file1 file2 file3
+
+
+Git needs not only changes, but snapshots. To do this, commit your changes:
+  git commit -m "My beautiful commit message"
+
+  git diff
+is useful to see only the changes. Review it before committing.
```
oneeyeman1, line 126: You need to review you changes before adding them to the staging area, i.e. before doing "git add...."

docs/contributing/Git-mini-guide.md
lanurmi, line 117: Ironically, this PR itself has proven using `add --all` is a bad idea.

Manolo-ES: No, I didn't use 'git add --all' but added the file.
The issue with this PR (mixing PRs) is that when I created the new branch in my local I forgot to checkout to master first, so the new branch hangs from the other branch instead of the master one.

vadz:
Sorry, but I'm really not sure we want to have Git guides in wx, it just doesn't seem to be the right place for it. We don't do anything unusual, so it would make much more sense to follow one of the many existing guides instead.

Unless someone has any really good arguments in favour of adding it, I'm not going to do it, sorry.

SpareSimian:
Agreed. I'd be OK with links to guides and any special notes on how the wx project diverges from standard practice. The only detail that caused me a problem was the need to init submodules and to repeat that step when switching branches. Most beginner git guides don't cover submodules. The wx contribution docs include mention of this now. Otherwise, the process is pretty vanilla.

vadz:
I'm closing this one, for the previously mentioned reasons.

PR #2575, opened 2021-11-01
gtk: map GDK_KEY_Launch1/2 to WXK_LAUNCH_APP1/2

martinetd:
Hello,

For context, I have a button mapped to KEY_PROG1 (linux input-event-code.h code) which maps to Xf86launch1/GDK_KEY_Launch1 and cannot use it in wxwidget.
The key could eventually be remapped with a bit of work (either xkb config or at driver/evdev level), but I'm probably not the only one with that problem and would like to discuss some possible way forward.

Here's the commit message of my initial suggestion to work on:
```
gtk: map Xf86Launch1/2 to WXK_LAUNCH_APP1/2

X11 defines Xf86Launch[0-9a-f], which are then also defined as
GDK_KEY_Launch[0-9A-F].

Unfortunately, keys which are not mapped are just plain ignored and the
application is then not able to receive these keyboard events at all.

The original PR https://github.com/wxWidgets/wxWidgets/pull/157 mapped
only LaunchA/B to WXK_LAUNCH_APP1/2, this patch also maps Launch1/2 to
these.
It would probably be better to add more LAUNCH_APPx to correspond to all
the available keys, but that might be difficult to map all of these to
other platforms so reuse the symbols here.
```

Note it might also make sense to just map all of the Xf86launchX bindings e.g. odd -> APP1, even -> APP2 or something like that?
As written in the commit message I'm not sure it'd make sense to just add more symbols, but I'd like some sort of event to be able to act on even if it's a bit difficult to grasp -- the current gtk code just ignores all keys it doesn't know about (e.g. ascii or hard coded list) but it's not really practical in the case a button is mapped to some weird key... So another possibility would perhaps be a "RAW" key that just exposes underlying keycode? Application portability is not a concern in my case, so I would have little qualm of hardcoding GDK_KEY_xyz in my application if I could.

Thanks!

PR #22830, opened 2022-09-29
Fix for build failure vs. GTK2 due to PR #2575

mkirkham:
There are a number of GDK_KEY_Launch macros that got put into use in PR #2575 that are apparently not defined in GTK2, and as a result current wxWidgets no longer builds for me. It appears to have been an oversight that these were not added to the gtk2-compat.h file with the others. This patch fixes the build for me.


vadz:
Thanks, I guess we should apply it, but the new names are actually available since GTK 2.21.8 released on 2010-09-13. Do you really use a GTK 2 version earlier than that??

mkirkham:
At the moment... yes, it appears so. At least for one platform. I am working on bringing my many projects/platforms more up to date but this is an incremental process to keep things in a building/working state (one step of which is bringing wxWidgets up to date, hence finding the new version wouldn't build anymore).

vadz:
Thanks, merged now and will also apply to 3.2 soon.

Issue #22807, opened 2022-09-18
wxWidgets inadequately chooses smalles icon from iconset


janopae:
When setting an IconBundle as a window's icon, on my High DPI system (using the Gtk backend), the smallest icon will be chosen, and it will be displayed way too small. As I provide many bigger icons, I expect a bigger icon to be chosen, at least big enough to fill up the title bar.

I use wxwidgets in Python using wxWidgets/Phoenix. However, I don't think the Python bindings cause this problem, so I post this issue here. Sorry for my demonstration being written in Python, I hope you can execute it anyway.

**To Reproduce**
Download the demonstration program: 
[demonstration_wx_icon_issue.zip](https://github.com/wxWidgets/wxWidgets/files/9593923/demonstration_wx_icon_issue.zip)

Extract it and go to the extracted directory

Execute it via `python3 main.py`

See the icon being too small: 
![grafik](https://user-images.githubusercontent.com/8160319/190901459-0d036d3e-e2bb-4f47-a471-49693f50f8f3.png)

**Platform and version information**
 - wxWidgets version you use: 4.2
 - wxWidgets port you use: wxGTK
 - OS and its version: Ubuntu 22.04
 - For wxGTK only
   + GTK version: 3.24.33
   + Which GDK backend is used: Wayland
   + If relevant, i.e. for appearance-related problems: current theme: Yaru

vadz:
For the reference, here is the code:
```python
import os
import wx

def getFullPath(relativePath: str):
    return os.path.join(os.path.dirname(__file__), relativePath)

def main():
    wxapp = wx.App()
    fr = wx.Frame(None, -1, 'Test')

    wx.Image.AddHandler(wx.PNGHandler())

    ib = wx.IconBundle()

    for sz in ("16", "32", "64", "128", "256"):
        ib.AddIcon(wx.Icon(wx.Bitmap(getFullPath(("resources/icon%s.png" % sz)))))

    fr.SetIcons(ib)

    fr.Show(True)
    wxapp.MainLoop()


if __name__ == '__main__':
    main()
```

I haven't tried it yet, but it looks like it ought to work. Is your application configured to use high DPI? I.e. does `wxWindow::GetDPIScaleFactor()` return 2 (or whichever DPI scaling you use)? Not doing this seems like the most likely reason for the problem.

janopae:
[...]

PR #22831, opened 2022-09-29
Update TLW icon when the DPI changes in wxMSW

vadz:
While we selected the icon of the size most appropriate for the current
DPI when the window was created, we didn't update the icon later if the
DPI changed as we should, so do it now.

Note that we abuse the existing MSWUpdateFontOnDPIChange() to make it
simple to backport this fix to 3.2 without breaking the ABI.

See #22807.

### DIFF
--- a/include/wx/msw/toplevel.h
+++ b/include/wx/msw/toplevel.h
@@ -169,6 +169,11 @@ protected:
                                           int& x, int& y,
                                           int& w, int& h) const wxOVERRIDE;
 
+    // override this one to update our icon on DPI change (not quite the same
+    // thing as font, but close enough...)
+    virtual void MSWUpdateFontOnDPIChange(const wxSize& newDPI) wxOVERRIDE;
+
+
     // This field contains the show command to use when showing the window the
     // next time and also indicates whether the window should be considered
     // being iconized or maximized (which may be different from whether it's
@@ -195,6 +200,10 @@ protected:
     wxWindowRef m_winLastFocused;
 
 private:
+    // Part of SetIcons() actually updating the window icons.
+    void DoSetIcons();
+
+
     // The system menu: initially NULL but can be set (once) by
     // MSWGetSystemMenu(). Owned by this window.
     wxMenu *m_menuSystem;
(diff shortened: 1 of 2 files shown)


vadz:
@MaartenBent I don't know if you see any better way of doing this, updating the icons in `MSWUpdateFontOnDPIChange()` is definitely not great, but anything else would have been more difficult/not ABI-compatible.

MaartenBent:
The other option is using `wxDPIChangedEvent`.
Adding `Bind(wxEVT_DPI_CHANGED, &wxTopLevelWindowMSW::OnDPIChanged, this);` in `wxTopLevelWindowMSW::Create()` and `wxTopLevelWindowMSW::OnDPIChanged(wxDPIChangedEvent& event)
{ ... }` where the icons are set.

But I don't know if this is ABI compatible, since a new function is added instead of overriding one.

vadz:
The problem with using an event handler is that it doesn't get to run if a user-defined one doesn't call `event.Skip()`, but I think we do want to always do this (at least I can't think of any reason not to), which is why I wanted to do it in some function. I think we might need something like `MSWOnDPIChanged()`, similar to wxGTK `OnInternalIdle()`, i.e. a function that could be used for executing the code that we always want to execute. This is a bit ugly, but a generic solution, e.g. some kind of "post-handler" at `wxEvtHandler` level, would complicate the already complicated logic there even further, so I'm not sure if we really want to do this...

Anyhow, I'll push this for now, thanks for looking at it!

imciner2:
The DPI changed event documentation (https://docs.wxwidgets.org/stable/classwx_d_p_i_changed_event.html) specifically says that users should skip the event to continue processing by the window, so users who don't have already been warned it may not properly adapt to the new DPI. There is a discussion about top-level window behavior in that event documentation already, so I would think extending that would be the best way to communicate this when it is added to the event handler.

vadz:
Well, yes, exactly: as this documentation explains, it may be sometimes necessary to not skip this event because the default window size scaling with DPI is not perfect and you may want, or need, to do better in your own code. However you do always want to update the icon, which is why I don't think it's a good idea to do it in the default handler.

vadz:
Merged into master, will backport to 3.2 too soon.

Issue #19075, opened 2021-02-09
wxWebView - wxWebViewBackendEdge - Runscript - get stuck


wxtrac:
*Issue migrated from trac ticket # 19075*

**component:** WebView | **priority:** normal | **keywords:** wxWebView,wxWebViewBackendEdge

#### 2021-02-09 20:53:48: hiemann (Rico Hiemann) created the issue

___
Hi,

 I used the wxWebView sample from master branch and wrote a little test html, see below.
 I noticed different behaviour between wxWebViewBackendEdge vs
 wxWebViewBackendIE especially with RunScript.

 Using the backend: wxWebViewBackendIE
 All links change the images as expected:
InternGrey - grey image - OK
InternRed - red image - OK
ExternGrey - grey image - OK
ExternRed - red image - OK

 Than switching to the backend: wxWebViewBackendEdge
 (Compilation and installation worked - backend message from log:
 wxWebViewEdge Version: Microsoft Edge WebView2 88.0.705)

InternGrey - grey image - OK
InternRed - red image - OK
ExternGrey - change image on first click but function get stuck
ExternRed - change image on first click but function get stuck

The internal C++ part:

```
void WebFrame::OnTitleChanged(wxWebViewEvent& evt)
{
    wxString Msg = evt.GetString();
    if (Msg == "testexternred")
    {
        RunScript("setImage('red.jpg');");
    }
    else if (Msg == "testexterngrey")
    {
        RunScript("setImage('grey.jpg');");
    }
}
```

wxWebViewBackendEdge get stuck in line (infinite loop, no error or any response):

```
if (m_browser->RunScript(javascript, &result))
```

 Thanks in advance
 Rico

```
<html>
  <head>
    <meta http-equiv="content-type" content="text/html; charset=UTF-8">
  </head>
  <body>
<script>
function setImage(content) 
{
	document.getElementById('myimage').setAttribute('src', content);
}
function mytitle(id)
{
	document.title = id;	
}	
</script>  
    <nav id="head">
	  <a href="#" id="testinterngrey" onclick="setImage('grey.jpg')">InternGrey</a>
	  <a href="#" id="testinternred" onclick="setImage('red.jpg')">InternRed</a>
	  <a href="#" id="testexterngrey" onclick="mytitle(this.id)">ExternGrey</a>
	  <a href="#" id="testexternred" onclick="mytitle(this.id)">ExternRed</a>
	</nav>
    <main>
[...]

PR #22834, opened 2022-10-01
Queue wxWebViewEdge events

### LINKED ISSUE #22744: Accessing HTML from wxWebViewEdge event handlers causes crash
When using wxWebView Edge backend, calling its methods such as `GetPageText()` or `GetPageSource()` from an event handler for an event such as `wxEVT_WEBVIEW_LOADED`, the webview becomes unstable, and the application crashes with unhandled exception when attempting to use the webview or just closing the application immediately.

Just placing `const wxString pageText = m_browser->GetPageText();` in the event handler is enough. Call stack after the crash:

```
'webviewsample.exe' (Win32): Loaded 'C:\Windows\System32\dbghelp.dll'. 
'webviewsample.exe' (Win32): Unloaded 'C:\Windows\System32\dbghelp.dll'
Exception thrown at 0x00007FFDD49D4FD9 in webviewsample.exe: Microsoft C++ exception: std::length_error at memory location 0x000000A3456FD150.
Exception thrown at 0x00007FFDD49D4FD9 in webviewsample.exe: Microsoft C++ exception: [rethrow] at memory location 0x0000000000000000.
Exception thrown at 0x00007FFDD49D4FD9 in webviewsample.exe: Microsoft C++ exception: std::length_error at memory location 0x000000A3456FF090.
Exception thrown at 0x00007FFDD49D4FD9 in webviewsample.exe: Microsoft C++ exception: [rethrow] at memory location 0x0000000000000000.
Exception thrown at 0x00007FFDD49D4FD9 in webviewsample.exe: Microsoft C++ exception: [rethrow] at memory location 0x0000000000000000.
Exception thrown at 0x00007FFDD49D4FD9 in webviewsample.exe: Microsoft C++ exception: std::length_error at memory location 0x000000A3456FF090.
Exception thrown at 0x00007FFDD49D4FD9 in webviewsample.exe: Microsoft C++ exception: [rethrow] at memory location 0x0000000000000000.
'webviewsample.exe' (Win32): Unloaded 'D:\Dev\Desktop\!Lib\wxWidgets-PB-build-vc17_x64_DLL\lib\vc_x64_dll\WebView2Loader.dll'
```
I tested that the crash occurs for handlers for `wxEVT_WEBVIEW_LOADED`, `wxEVT_WEBVIEW_NAVIGATED`, and `wxEVT_WEBVIEW_TITLE_CHANGED`.  There is no crash and the string seems correct when I use `CallAfter()`, e.g., in the event handler replace
```cpp
const wxString pageText = m_browser->GetPageText();
```
with
```cpp
CallAfter([this] { const wxString pageText = m_browser->GetPageText(); wxLogMessage("Page source %s", pageText);} );
```

**To Reproduce**
1. Patch the webview sample with
```diff
diff --git a/samples/webview/webview.cpp b/samples/webview/webview.cpp
index b0a71da6a0..8818da6742 100644
--- a/samples/webview/webview.cpp
+++ b/samples/webview/webview.cpp
@@ -865,6 +865,8 @@ void WebFrame::OnNavigationComplete(wxWebViewEvent& evt)
   */
 void WebFrame::OnDocumentLoaded(wxWebViewEvent& evt)
 {
+    const wxString pageText = m_browser->GetPageText();
+
     //Only notify if the document is the main frame, not a subframe
     if(evt.GetURL() == m_browser->GetCurrentURL())
     {
```
2. Build the sample, ensuring it uses `wxWebViewEdge` backend.
3. Run the sample and wait until the page is displayed.
4. Close the sample.
5. Observe the unhandled exception error box.

 - wxWidgets version you use: master, built as 64-bit DLL with MSVS 2022 with CMake
 - OS and its version: Windows 10.0.19044.1889
 - WebView2Loader.dll version: 1.0.705.50
 - Edge version: 104.0.1293.63

**EDIT**
Calling `GetPage*()` results in calling `RunScript()` which waits for `RunScriptAsync()` while `wxYield()`ing. 
https://github.com/wxWidgets/wxWidgets/blob/6ee593c78d082156c4fba2997241e530ba1d24da/src/common/webview.cpp#L232-L236

However, something goes wrong when doing this in a `wxWebviewEdge` event handler and the script execution seems to never end properly, with the application forever stuck in that `while` loop.

TcT2k:
Using `HandleWindowEvent()` from the WebView2 event callbacks is a common error source when user code does something blocking like `RunScript()` or `ShowModal()`.

Queue the events to prevent this where possible, in `wxEVT_WEBVIEW_NAVIGATING` at least warn the developer instead of hanging silently.

Closes: #22744, #19075

### DIFF
--- a/src/msw/webview_edge.cpp
+++ b/src/msw/webview_edge.cpp
@@ -18,6 +18,7 @@
 #include "wx/module.h"
 #include "wx/mstream.h"
 #include "wx/log.h"
+#include "wx/scopeguard.h"
 #include "wx/stdpaths.h"
 #include "wx/thread.h"
 #include "wx/tokenzr.h"
@@ -246,6 +247,10 @@ public:
     wxCOMPtr<ICoreWebView2WebResourceRequestedEventArgs> m_args;
 };
 
+#define wxWEBVIEW_EDGE_EVENT_HANDLER_METHOD \
+    m_inEventCallback = true; \
+    wxON_BLOCK_EXIT_SET(m_inEventCallback, false);
+
 wxString wxWebViewEdgeImpl::ms_browserExecutableDir;
 
 wxWebViewEdgeImpl::wxWebViewEdgeImpl(wxWebViewEdge* webview):
@@ -276,6 +281,7 @@ bool wxWebViewEdgeImpl::Create()
 {
     m_initialized = false;
     m_isBusy = false;
+    m_inEventCallback = false;
     m_pendingContextMenuEnabled = -1;
     m_pendingAccessToDevToolsEnabled = 0;
 
@@ -365,6 +371,7 @@ void wxWebViewEdgeImpl::UpdateBounds()
 
 HRESULT wxWebViewEdgeImpl::OnNavigationStarting(ICoreWebView2* WXUNUSED(sender), ICoreWebView2NavigationStartingEventArgs* args)
 {
+    wxWEBVIEW_EDGE_EVENT_HANDLER_METHOD
     m_isBusy = true;
     wxString evtURL;
     wxCoTaskMemPtr<wchar_t> uri;
@@ -389,7 +396,7 @@ HRESULT wxWebViewEdgeImpl::OnSourceChanged(ICoreWebView2 * WXUNUSED(sender), ICo
         // navigation within the current document, send apropriate events
         wxWebViewEvent event(wxEVT_WEBVIEW_NAVIGATING, m_ctrl->GetId(), m_ctrl->GetCurrentURL(), wxString());
         event.SetEventObject(m_ctrl);
-        m_ctrl->HandleWindowEvent(event);
+        m_ctrl->GetEventHandler()->AddPendingEvent(event);
         OnNavigationCompleted(NULL, NULL);
         OnDOMContentLoaded(NULL, NULL);
     }
@@ -441,7 +448,7 @@ HRESULT wxWebViewEdgeImpl::OnNavigationCompleted(ICoreWebView2* WXUNUSED(sender)
             }
         }
         if (!ignoreStatus)
-            m_ctrl->HandleWindowEvent(event);
+            m_ctrl->GetEventHandler()->AddPendingEvent(event);
     }
     else
     {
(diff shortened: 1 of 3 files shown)


src/msw/webview_edge.cpp
```diff
@@ -1080,6 +1099,15 @@ void wxWebViewEdge::MSWSetBrowserExecutableDir(const wxString & path)
     wxWebViewEdgeImpl::ms_browserExecutableDir = path;
 }
 
+bool wxWebViewEdge::RunScript(const wxString& javascript, wxString* output) const
+{
+    if (m_impl->m_inEventCallback)
+    {
+        wxFAIL_MSG("RunScript() cannot be used during event callbacks. Consider using RunScriptAsync()");
+    }
```
vadz, line 1107: Should we `return false` here? Or, alternatively, use
```
wxCHECK_MSG( !m_impl->m_inEventCallback, false, "..." );
```
?

src/msw/webview_edge.cpp
```diff
@@ -246,6 +246,23 @@ class wxWebViewEdgeHandlerResponse : public wxWebViewHandlerResponse
     wxCOMPtr<ICoreWebView2WebResourceRequestedEventArgs> m_args;
 };
 
+class wxWebViewImplCallbackActivator
```
vadz, line 249: I wonder if we should add some `wxScopedVarSetter` that could be used as
```
wxScopedVarSetter set(m_impl->m_inEventCallback, true);
```
as it looks like we do something like this in many places in our code.

There is also a `wxRecursionGuard`, but it's not quite the same thing...

vadz (commented):
Thanks, no real comments and I agree that this is the best thing to do.

I just think it would be better to avoid reentrancies even in release build, i.e. add a `return` there, but please let me know if you disagree.

src/msw/webview_edge.cpp
TcT2k: Sounds good,  obviously this shouldn't happen in release code, but still better that hanging the process.
I've implemented it with `wxCHECK_MSG()`

src/msw/webview_edge.cpp
TcT2k: While looking where to add something like `wxScopedVarSetter` I've came accross `wxON_BLOCK_EXIT_SET()` which seems perfectly suitable for this scenario and I've used that now.

vadz (approved):
Perfect, thanks!

PR #22836, opened 2022-10-01
Remove "Printing " from title of printout documents

Blake-Madden:
When you print something to PDF (e.g., using Microsoft Print to PDF), the title embedded into the PDF will start with "Printing " and looks a bit odd (IMO).  For example, if I print something to PDF and open it in Adobe Reader or Chrome, the title shown in the tab will be something like "Printing Q1 Business Earnings".  Removing "Printing " as I do in this PR will make it appear as "Q1 Business Earnings" as one would expect.


vadz:
Thanks, I agree that using "Printing" doesn't make much sense here. While many implementations just ignore the string passed to `StartDoc()` entirely, MSW one uses it as the value for `DOCINFO::lpszDocName` which should be the document name, clearly, and not some sentence using it. And under Mac `PMPrintSettingsSetJobName()` is used, so "Printing" doesn't seem appropriate there neither.

I'll apply this to both branches soon if there are no objections.

vadz:
This actually was applied quite some time ago but I forgot to close it.

Not sure why there was a merge pushed here, normally this branch should be just deleted.

Issue #22835, opened 2022-10-01
Random crashes in wxBitmapBundle::FromSVG()


wh11204:
### Description

I am getting random crashes and hangs when using wxBitmapBundle::FromSVG().

The first crash is in nsvg_Parse(), looks like this is related to non-english locales and is [fixed](https://github.com/memononen/nanosvg/commit/d2002eeaf20f5b30c41b10a929bf9e368580034b) in nanoSVG trunk but not in wxWidgets.
A workaround to this crash is a temporary locale change before calling FromSVG().

The second crash happens in nsvg_addEdge() with this DrMinGW information:

codeblocks.exe caused an Access Violation at location 706C0498 in module wxmsw32u_gcc_custom.dll Writing to location 00000010.

Registers:
eax=00000000 ebx=0aece0d8 ecx=00000000 edx=00000001 esi=00000002 edi=0ada48d0
eip=706c0498 esp=0081d680 ebp=0aece0d8 iopl=0         nv up ei pl nz na pe nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00210202

706C0498 402F8F8A 3FE141B7 411A9455  wxmsw32u_gcc_custom.dll!nsvg__addEdge
706C77D1 0081D7D8 00000000 00000030  wxmsw32u_gcc_custom.dll!nsvgRasterize

### Platform and version information
- wxWidgets version  you use: wx3.2.1
- wxWidgets port  you use: wxMSW
- OS  and its version: Windows 7 SP1, Windowa 10 21H2, both 64 bits.

vadz:
Could you please provide the SVG which results in this?

In any case, it looks like we should merge the latest nanoSVG, there seem to be quite a few fixes applied to it relatively recently.

wh11204:
This is one that fails, as you see it is very simple. It is loaded using wxBitmapBundle::FromSVG("es_ES.svg", wxSize(20, 20)) and the resulting bundle is used to construct wxStatictBitmap(this, wxID_ANY, bundle). The locale is spanish, with comma as decimal separator. A locale change to default suppress the first crash.

First crash happens inside FromSVG(), second when rasterizing the bundle inside wxStaticBitmap::DoCreate(). Crash frequency ranges from 10 to 75 %.

![es_ES](https://user-images.githubusercontent.com/26407261/193418233-359273f1-4468-4be0-805d-5fd3ab44ae79.svg)

wh11204:
[...]

PR #22838, opened 2022-10-01
Document third wxBitmapBundle::FromSVG() variant

TcT2k:
Add a note that data has to be null terminated with the variants without len parameter and to use the third variant for data that is not null terminated.

See: #22835


vadz:
Thanks, will apply soon with some minor fixes.

vadz:
Applied to both branches now, thanks again.

PR #22839, opened 2022-10-03
Fixes for wxDataViewCtrl header-related issues in wxOSX

vadz:
@csomor I'd be grateful if you could please have a look at these changes. I've tested them and they seem to work but, as usual, I have no idea if this is the right way to do things under Mac and if there is no better one. TIA!

### DIFF
--- a/src/osx/cocoa/dataview.mm
+++ b/src/osx/cocoa/dataview.mm
@@ -2583,8 +2583,29 @@ void wxCocoaDataViewControl::DoSetIndent(int indent)
     [m_OutlineView setIndentationPerLevel:static_cast<CGFloat>(indent)];
 }
 
-void wxCocoaDataViewControl::HitTest(const wxPoint& point, wxDataViewItem& item, wxDataViewColumn*& columnPtr) const
+void wxCocoaDataViewControl::HitTest(const wxPoint& point_, wxDataViewItem& item, wxDataViewColumn*& columnPtr) const
 {
+    // Assume no item by default.
+    columnPtr = NULL;
+    item      = wxDataViewItem();
+
+    // Make a copy before modifying it.
+    wxPoint point(point_);
+
+    // First check that the point is not inside the header area and adjust it
+    // by its offset.
+    if (NSTableHeaderView* const headerView = [m_OutlineView headerView])
+    {
+        if (point.y < headerView.visibleRect.size.height)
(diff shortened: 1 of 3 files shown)


src/osx/cocoa/dataview.mm
```diff
@@ -163,6 +163,46 @@ -(id) copyWithZone:(NSZone*)zone
...
+}
+
+    -(id) initWithDVC:(wxDataViewCtrl*)ctrl;
+    -(void) rightMouseDown:(NSEvent *)theEvent;
+@end
+
+@implementation wxDVCNSHeaderView
+
+-(id) initWithDVC:(wxDataViewCtrl*)ctrl
+{
```
csomor, line 182: I'm surprised you don't have to model along a initWithFrame initialization, since it is a NSView I would have thought the frame should be given already when initializing but apparently enough layout calls are triggered before things are shown ... and nothing vital seems to be in the NSTableHeaderView initWithFrame ...

csomor (approved):
I've only a small comment, but if things work in your tests, my thoughts may be overly cautious ... Thanks

src/osx/cocoa/dataview.mm
vadz: Thanks for looking at this!

I am not sure what I could/should pass to `initWithFrame:`, so I'm going to leave it like this for now, as not calling doesn't seem to result in any problems in practice...

jey5nd6:
Tested in KiCad.  Works nicely.

src/osx/cocoa/dataview.mm
csomor: most of the NSViews get initialized with `InitWithFrame:` so this issue might arise elsewhere, that's why I thought I'd mention it. Here we'd need the headers frame, I've seen places that just calculate the height to be some constant (17). So if the need arises I'd do a `initWithFrame: DVC:` and call `[super initWithFrame:]`

vadz: My problem is that I need to create this header view and I don't know which frame to pass to it when doing it. Currently I conveniently avoid the problem and I'd rather keep it like this until we find something actually requiring doing it.

csomor: in this case, absolutely agreed, if it works that's fine

PR #22840, opened 2022-10-03
Prevent memory bloat when creating new wxThreads.

nietsu:
The log functions in PthreadStart cause a new wxThreadSpecificInfo object to be created after the original one has already been discarded by CallEntry. Perform the wxThreadSpecificInfo::ThreadCleanUp again at the end of PthreadStart, so that it will not needlessly remain there consuming memory till the end of program.

### DIFF
--- a/include/wx/thread.h
+++ b/include/wx/thread.h
@@ -602,9 +602,6 @@ protected:
     // of this thread.
     virtual void *Entry() = 0;
 
-    // use this to call the Entry() virtual method
-    void *CallEntry();
-
     // Callbacks which may be overridden by the derived class to perform some
     // specific actions when the thread is deleted or killed. By default they
     // do nothing.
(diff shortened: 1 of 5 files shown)


vadz (commented):
Thanks for finding and fixing this problem!

I wonder if we should stop calling `ThreadCleanUp()` from `CallEntry()` (and then just remove it completely, as it won't be doing anything useful then) to avoid calling it twice?

nietsu:
> I wonder if we should stop calling `ThreadCleanUp()` from `CallEntry()` (and then just remove it completely, as it won't be doing anything useful then) to avoid calling it twice?

I agree it's a bit messy to call it twice. Still, I think at least Window implementation still relies on the call from CallEntry(), so you probably shouldn't just remove it without creating some other method of calling it on platforms that don't use the implementation in threadpsx.cpp. Just calling it directly from each implementation would be a solution, although a bit fragile one.

vadz:
I've added a commit to do this, could you please check if it works correctly for you? TIA!

nietsu:
Just tested it. Works great!

PR #22845, opened 2022-10-04
3.2 proposed backports

vadz:


PBfordev:
Assuming backporting an isolated commit is easy and simple, I suggest also backporting these, not really important but IMO still useful, commits:

**Code**
- [ ] Add macros for event tables for three wxWebView events: d9bb8c3

I think `wxNODISCARD` could also be backported but even if it really does not introduce a new feature, it could be considered backwards-incompatible, as it can break the builds that worked before (even if that was likely a bug).

**Docs**
- [ ] Update Windows Theme support page: b112dfb
- [ ] Update Multithreading Overview programming guide: 41fc164
- [ ] Update Backwards Compatibility programming guide: cf43c97 (it is written from v3.2 point of view)

vadz:
Thanks, I've cherry picked all the commits you mentioned (but didn't push yet), but `wxNODISCARD` is a bit less clear, so I'm not doing this just yet.

Issue #16318, opened 2014-06-05
wxBitmapComboBox does not emit wxEVT_TEXT_ENTER unless empty


wxtrac:
*Issue migrated from trac ticket # 16318*

**component:** wxMSW | **priority:** normal | **keywords:** wxBitmapComboBox keyboard

#### 2014-06-05 18:02:14: jeff_hanna (Jeff Hanna) created the issue

___
A populated wx.combo.BitmapComboBox will not emit a wx.EVT_TEXT_ENTER event. But, the same control without any items in its list will successfully emit that event. This happens regardless of whether or not the wx.TE_PROCESS_ENTER style flag is set on the control (on Windows).

I've verified this on wx 3.0.0.0 msw (classic). The attached program demonstrates the problem.

PR #22847, opened 2022-10-05
wxBitmapComboBox: Fix for emitting of  wxEVT_TEXT_ENTER event

YuSanka:
Alternative solution for #16318

This code was provided by @xarbit for https://github.com/prusa3d/PrusaSlicer with https://github.com/prusa3d/wxWidgets/commit/24d72953141b5f3acb2bd0b750e73efedd8b8b09#diff-86749244e33368aea5650ada6ac71568b356109f0fab7f3645536e75f308c6eb. 

But most of code in this commit doesn't used now. 
And it would be great to have these changes in the upstream wxWidgets.

### DIFF
--- a/samples/widgets/bmpcombobox.cpp
+++ b/samples/widgets/bmpcombobox.cpp
@@ -425,7 +425,9 @@ void BitmapComboBoxWidgetsPage::CreateCombo()
         flags |= wxCB_SORT;
     if ( m_chkReadonly->GetValue() )
         flags |= wxCB_READONLY;
-
+    else
+        flags |= wxTE_PROCESS_ENTER;
+    
     switch ( m_radioKind->GetSelection() )
     {
         default:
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for submitting this, but manipulating the wndproc like this looks really hackish and I think I'd prefer to use [this patch](https://github.com/wxWidgets/trac-attachments/tree/main/ticket/bae/bae636b833bd3b8cb12b4697e8cfe4e538f037b6/f95183252c0848d43ac8da954c5459cc21b7d99f.patch) from #16318 instead.

But actually it would make even more sense to avoid dealing with the internals of `wxComboBox` in the derived class and provide a way to recreate it in this class itself, so I've created #22850 doing this and replacing this PR. Could you please check that the other PR works for you and comment there if it doesn't? TIA!

Issue #22846, opened 2022-10-04
Enable native double buffering for all non-native windows by default


vadz:
There doesn't seem to be any reason _not_ to enable double buffering, i.e. set `WS_EX_COMPOSITED` for all generic windows in wxMSW (some native windows don't work with this style), so we should try doing it and turning `wxALWAYS_NATIVE_DOUBLE_BUFFER` to 1 for wxMSW as well.

vadz:
This was actually done in #22851, but see #22876 now.

Tommykrul:
Unless I am misunderstanding how to enable proper double buffering in wxWidgets on Windows, I belief this change makes some incorrect assumptions, at least in the current implementation.

We have some internal tools that make liberal use of custom painted widgets. These controls are all implemented using SetBackgroundStyle(wxBG_STYLE_PAINT) / wxAutoBufferedPaintDC() / wxGCDC(); Starting with 3.3.0, all these controls exhibit severe flicking, very slow rendering, and seem to refresh after every individual paint operation (items appear one by one and the view takes several seconds to update). The same code linking against 3.2.0 shows no flickering and performance is fast as expected. The problem only occurs on Windows. macOS renders fine. 

Reading through the code changes, it seems like the assumption is being made that Windows now supports "native double buffering" through calling WS_EX_COMPOSITED (wxWindow::SetDoubleBuffered(true)) on the windows. And as such wxALWAYS_NATIVE_DOUBLE_BUFFER is now turned on for Windows as well. This in turn switches wxAutoBufferedPaintDC from wxBufferedPaintDC to simply wxPaintDC.

The problem as far as I can tell is that does not provide a generalized double buffering solution for windows. While I don't know the exact behavior or implementation of WS_EX_COMPOSITED, from what I can find online the general consensus seems to be that WS_EX_COMPOSITED only double-buffers the individual controls and that it seems to be a custom solve for certain native widgets. It does not seem to turn on double buffering for the entire panel.

We tried to explicitly enable SetDoubleBuffered(true) on any and all panels that do custom drawing using a wxGCDC and nothing we tried resulted in flicker free drawing. Conversely, if we hard code our wxDC to use a wxBufferedPaintDC, all flicking is gone and rendering performance is normal again. 

[...]

PR #22851, opened 2022-10-06
Enable double buffering by default in wxMSW

vadz:
I've simply turned on `WS_EX_COMPOSITED` for all our windows and so far I haven't seen anything bad happening in the few samples I've tried (drawing, grid, notebook, toolbar, widgets). This doesn't mean there are no problems at all, of course, so it would be great if people could please test their applications with this PR _before_ it gets merged.

Please report any problems you see with it!

See #22846.

### DIFF
--- a/src/msw/window.cpp
+++ b/src/msw/window.cpp
@@ -539,6 +539,19 @@ bool wxWindowMSW::CreateUsingMSWClass(const wxChar* classname,
     msflags &= ~WS_BORDER;
 #endif // wxUniversal
 
+    // Enable double buffering by default for all our own, i.e. not the ones
+    // using native controls, classes.
+    //
+    // Note that this function is not used for top-level windows, so we don't
+    // set this style for them, and also that setting it for children of
+    // windows that already have WS_EX_COMPOSITED set doesn't seem to have any
+    // bad effect as the style is just ignored in this case, so we don't bother
+    // checking it it's already set for the parent, even though we could.
+    if ( !classname )
+    {
+        exstyle |= WS_EX_COMPOSITED;
+    }
+
     if ( IsShown() )
     {
         msflags |= WS_VISIBLE;
--- a/src/msw/window.cpp
+++ b/src/msw/window.cpp
@@ -544,6 +544,10 @@ bool wxWindowMSW::CreateUsingMSWClass(const wxChar* classname,
         msflags |= WS_VISIBLE;
     }
 
+    // If the class name is not specified, use the one for generic wxWindow.
+    if ( !classname )
+        classname = GetMSWClassName(style);
+
     if ( !MSWCreate(classname, NULL, pos, size, msflags, exstyle) )
         return false;
 
--- a/include/wx/dcbuffer.h
+++ b/include/wx/dcbuffer.h
@@ -15,12 +15,11 @@
 #include "wx/dcclient.h"
 #include "wx/window.h"
 
-// Split platforms into two groups - those which have well-working
-// double-buffering by default, and those which do not.
-#if defined(__WXMAC__) || defined(__WXGTK20__) || defined(__WXDFB__) || defined(__WXQT__)
-    #define wxALWAYS_NATIVE_DOUBLE_BUFFER       1
-#else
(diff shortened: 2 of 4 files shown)


src/msw/window.cpp
```diff
@@ -539,6 +539,19 @@ bool wxWindowMSW::CreateUsingMSWClass(const wxChar* classname,
...
+    // Note that this function is not used for top-level windows, so we don't
+    // set this style for them, and also that setting it for children of
+    // windows that already have WS_EX_COMPOSITED set doesn't seem to have any
+    // bad effect as the style is just ignored in this case, so we don't bother
+    // checking it it's already set for the parent, even though we could.
+    if ( !classname )
+    {
+        exstyle |= WS_EX_COMPOSITED;
+    }
+
```
AliKet, line 554: Tried this PR with the `wxGrid` control and it doesn't seem to work
because this style `WS_EX_COMPOSITED` is not applied for it (or its children)  !
i.e. `classname` is != NULL and has this value "`wxWindowNR`", what does that mean ?

AliKet: No, it's a false alarm sorry.
I just recompiled everything from the beginning and I confirm that it works perfectly well
sorry again for the noise!

vadz:
I'm not sure if anybody has really tested this, but I think that the only way to really get it tested is to merge it into master and see if anybody complains, so I'm going to do it.

PBfordev:
I think I found first victims already, just by running the bundled samples.
wxAuiToolbar::OnPaint asserts when creating `wxAutoBufferedPaintDC`:
> D:\Dev\Desktop\!Lib\wxWidgets-PB\include\wx/dcbuffer.h(229): assert "win->GetBackgroundStyle() == wxBG_STYLE_PAINT" failed in wxAutoBufferedPaintDC::wxAutoBufferedPaintDC(): You need to call SetBackgroundStyle(wxBG_STYLE_PAINT) in ctor, and also, if needed, paint the background in wxEVT_PAINT handler.

I also see several drawing issues when compared to the recentish master but I am not sure if it is caused by this commit:
1. When dragging an image in dragimg sample, neither image nor cursor are shown.
2. When dragging a card in the forty demo, the card is not shown.
3. When starting listctrl sample, a black rectangle is shown in place of the list for about a half a second. The background can be also observed when resizing the list. IIRC, there was an issue like this with the control, fixed loong time ago.

vadz:
Thanks for testing!

I'll check in the fix for `wxAuiToolBar` problem in a moment, but I'll have to look at the other ones a bit later.

DietmarSchwertberger:
@vadz : I think I found a problem with double buffering.
When dragging a row or col, the drop marker is missing now.

E.g. in the grid sample with "Grid -> Row drag move".
In the screenshot it's between lines 6 and 7:

Before merging this branch:
grafik

After:
grafik


I can work around by commenting out `exstyle |= WS_EX_COMPOSITED;`.


The relevant code for drawing the marker is in `wxGrid::ProcessRowColLabelMouseEvent`.
I would guess that some additional action is required around the ClientDC, but could not see what is missing. Other code like "pressing" the row label looks similar, but does not have problems.
```
if ( markerPos != m_dragLastPos || markerColour != m_dragLastColour )
{
    wxClientDC dc( headerWin );
    oper.PrepareDCForLabels(this, dc);

    int markerLength = oper.Select(headerWin->GetClientSize());

    // Clean up the last indicator if position has changed
    if ( m_dragLastPos >= 0 && markerPos != m_dragLastPos )
    {
        wxPen pen(headerWin->GetBackgroundColour(), 2);
        dc.SetPen(pen);
        oper.DrawParallelLine(dc, 0, markerLength, m_dragLastPos + 1);
        dc.SetPen(wxNullPen);

        int lastLine = oper.PosToLine(this, m_dragLastPos, nullptr, false);

        if ( lastLine != -1 )
            oper.DrawLineLabel(this, dc, lastLine);
    }

    // Draw the marker
    wxPen pen(*markerColour, 2);
    dc.SetPen(pen);
    oper.DrawParallelLine(dc, 0, markerLength, markerPos + 1);
    dc.SetPen(wxNullPen);

    m_dragLastPos = markerPos;
    m_dragLastColour = markerColour;
}
```

PR #22853, opened 2022-10-06
Use wxChoice for month selection in wxGenericCalendarCtrl

lanurmi:
For no obvious reason the generic calendar control used a read-only wxComboBox until now.

While this is more or less identical to a wxChoice on MSW, it is somewhat different on OSX and Gtk: Whereas a wxChoice can be opened by clicking anywhere on it, a wxComboBox requires clicking on the arrows on the right edge. A wxComboBox allows selecting and copying the current value, but this is probably not terribly useful in a calendar.

### DIFF
--- a/include/wx/generic/calctrlg.h
+++ b/include/wx/generic/calctrlg.h
@@ -14,7 +14,7 @@
 #include "wx/control.h"         // the base class
 #include "wx/dcclient.h"        // for wxPaintDC
 
-class WXDLLIMPEXP_FWD_CORE wxComboBox;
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, I agree that `wxChoice` makes more sense here.

I don't know if can backport this to 3.2, in principle `m_comboMonth` is private and replacing it with `m_choiceMonth` shouldn't change the ABI but formally it's not a compatible change...

vadz:
I've created #22971 for 3.2.

Issue #22684, opened 2022-07-31
Drop support for pre-C++11 compilers


vadz:
This will require at least gcc 4.8, clang 3.4, MSVS 2015 which seems reasonable.

See https://en.cppreference.com/w/cpp/compiler_support/11

Some of the things to do:

- [x] Remove MSVS < 2015 support, including the checks for it in the code, all `.vcproj` files (and prevent bakefile from regenerating them while we still use it) and CI jobs using it.
- [x] Remove checks for gcc < 4.8 in the code.
- [x] Replace macros such as `wxOVERRIDE`, `wxNOEXCEPT` etc with their values (still keep the macros themselves, however, as somebody is almost surely using them in their code and there is no real reason to break it).
- [ ] Modernize the code base using clang-tidy (`bear -- make` can be used to generate the compilation database for it, note that it has to be done separately for all ports).
  - [x] modernize-use-nullptr 
  - [ ] modernize-loop-convert
  - [ ] modernize-raw-string-literal
  - [ ] modernize-use-using
- [x] Replace `wxConvertibleTo` from `include/wx/meta/convertible.h` with C++11 equivalent.
- [x] Replace calls to `wxShrinkToFit()` with its expansion.
- [x] Update CATCH to 2.x.
- [x] Use `__func__` instead of `__FUNCTION__`
- [x] Grep for `TODO-C++11` comments in the code.

Kumazuma:
There is any plan dropping support for win xp?

vadz:
Yes, see #22689.

vadz:
I'll close this once #22962 is merged. More can, and should, be done, but let's create separate issues for the other tasks.

PR #22854, opened 2022-10-07
Require C++11

vadz:
See #22684.


vadz:
I think this can be merged now, even if more things remain to be done, obviously -- but this is already a start.

Please let me know if anybody sees any problems with merging.

PBfordev:
This PR removes support for several MSVC versions.  Should this not reflect in the docs as well? I have done quick search and update of the docs ([commit](https://github.com/wxWidgets/wxWidgets/commit/7194871bae059264e3917e74741755cbd867f2da)). I am not sure about MSVS solutions for individual MSVS versions using the version-specific compiler prefixes (e.g., `vc140`, `vc141`) or all using `vc14x`.

I am never sure whether the root `readme.md` applies to the stable or master but if it applies to the master, the compiler information should be updated there as well.

I have also attempted to update the MSVC debug visualisers (https://github.com/wxWidgets/wxWidgets/commit/59890939b498a8239cbc4429a66335560c695c48) but I am not 100% sure about `autoexp.inc` being unusable now.

I assume the update for binaries build and release process will come in a separate PR since there is a lot of time for that.

Losing the ability to open any sample directly in MSVS is not great, but I guess it cannot be helped, maintaining them by hand is probably unfeasible.

vadz:
> This PR removes support for several MSVC versions. Should this not reflect in the docs as well? I have done quick search and update of the docs ([commit](https://github.com/wxWidgets/wxWidgets/commit/7194871bae059264e3917e74741755cbd867f2da)).

Thanks, I've completely forgotten to do this, but you're right, of course. I'll cherry pick your commit.

> I am not sure about MSVS solutions for individual MSVS versions using the version-specific compiler prefixes (e.g., `vc140`, `vc141`) or all using `vc14x`.

We're going to preserve compatibility and use version-dependent suffixes even if personally I find it more convenient to use `vc14x`, which I do in my `build/msw/wx_local.props`.
 
> I am never sure whether the root `readme.md` applies to the stable or master

GitHub doesn't really understand anything except master (or main, but just one single branch), so it applies to master. We could mention or link 3.2 README from the master one perhaps.

> but if it applies to the master, the compiler information should be updated there as well.

Yes, indeed.
 
> I have also attempted to update the MSVC debug visualisers ([5989093](https://github.com/wxWidgets/wxWidgets/commit/59890939b498a8239cbc4429a66335560c695c48))

Another nice change, thank you! 

> but I am not 100% sure about `autoexp.inc` being unusable now.

I'm pretty sure it's not supported by MSVS 2015+, thanks for thinking about this.

> Losing the ability to open any sample directly in MSVS is not great, but I guess it cannot be helped, maintaining them by hand is probably unfeasible.

I'm probably going to end up writing some simple project generation script. I don't want to reinvent the wheel, which is why I thought of [using xmake](https://groups.google.com/g/wx-dev/c/XSFIlZ85fG8), but it seems like it would be more trouble than it's worth.

PBfordev:
I also suggest replacing `wxOVERRIDE` and `wxNOEXCEPT` with the actual C++11 keywords, done here: https://github.com/PBfordev/wxWidgets/commit/435c88f46f600f3beaf7d144708bf189b6c67bbc and in the documentation fix (where I accidentally replaced `wxOVERRIDE` with `override` too) but in the end decided to undocument the macro alltogher since it is not used in wxWidgets code anymore: https://github.com/PBfordev/wxWidgets/commit/d1e87b89f0dfd764cb6af41c67d4bbceb6ede139

But perhaps it would be better to keep it documented as deprecated, since it was documented, I assume it was intended to be used not only by the library but also by the users?

vadz:
Thanks, we definitely want to replace all the macros occurrences in wx itself, so I'll cherry-pick this commit too (but I'll have to tweak the change to `docs/doxygen/mainpages/const_cpp.h` in it, see below).

Concerning the documentation, I don't think we need to keep documenting `wxOVERRIDE` and `wxNOEXCEPT`, as they shouldn't need to be ever used again and their meaning in the existing code should be clear enough. I'm less sure about `wxHAS_NOEXCEPT`, we probably do want to keep it just to say that it's always defined now, as this part might not be 100% obvious. But then it doesn't cost much to keep the other macros in the docs neither, just with "deprecated macro expanding to <C++11 equivalent>"... I.e. I'm not completely sure what to do about them neither.

a-wi:
Maybe it would be good to update Catch to v2.x?

vadz:
> Maybe it would be good to update Catch to v2.x?

Yes, absolutely, this is one of the first things I want to do after merging this.

PBfordev:
The compiler list in root README.md has not been updated yet but perhaps in the end you decided to keep the list for the stable version...

vadz:
Of course I forgot to do it :-( Thanks for the reminder!

Issue #22689, opened 2022-08-01
Remove support for older Windows versions


vadz:
Not sure what the new minimum required version should be, i.e. we could decide to still support Windows 7 as it doesn't cost that much or we could require Windows 10 (which build then?), but we definitely don't need to support XP or Vista any more.

- [x] Remove `wxUSE_COMPILER_TLS` option and just always do it.
- [x] Remove `wxHAS_MSW_TASKDIALOG`

Kumazuma:
if wxWidgets does not need any feature of win10, I suggest to keep support for win7, I hope replace joypad api to xinput. what is we need win10 specific api or feature?

vadz:
It does use Windows 10-specific features, see the various uses of `wxDynamicLibrary` in the code, but I think it wouldn't be a big burden to keep them, at least for now. XP support has to go, though (I actually wouldn't be surprised if it's already broken as I don't think anybody has used wx under XP for the last several years).

TcT2k:
I think (at least for now) Win7 is a easy enough to achieve minimum target.
Maybe when the Microsoft compiler default settings change to Win 10 (like when they dropped XP default support a few years back) the situation would be different.

I haven't tested for a long time on XP, but I wouldn't be surprised if the current codebase still works on XP.

TcT2k:
BTW many uses of `wxDynamicLibrary` in MSW code are there for outdated headers of minGW. But there are a few which are there for WinXP availability.

utelle:
Win7 was widely used (and I'm pretty sure it still is). The (unloved) successor Win8.1 is very similar to Win7.  Taking into account that official Win8.1 support ends only in January 2023, I would propose that wx should still support Win7 and Win8 for a couple of years (as was done for XP for which official support ended in 2014).

vadz:
Yes, I think we can do this. OTOH I'd indeed like to drop support for old MinGW. We will require gcc 5 too (see #22684), but this doesn't buy as much, unfortunately.

Kumazuma:
There is no need about old MinGW less 8.1.0...

vadz:
It's really not clear why should 8.1 be the cut-off and not 7 or 10 or something else.

Also, the real problem is not the compiler version but the MinGW vs MinGW-w64 distinction: the latter includes almost everything we need, but the former lacks a lot of things even in the latest releases (unless this changed very recently). I'm not sure if we can just drop support for the classic MinGW, but it would be nice if we could.

[...]

PR #22857, opened 2022-10-07
Update comments in setup_inc.h

PBfordev:
I have updated some comments but there are still many which I think should be updated but I am not sure how.

**WXWIN_COMPATIBILITY_3_0**
> Default is 1 but please try building your code with 0 as the default will
> change to 0 in the next version and the deprecated functions will disappear
> in the version after it completely.

We have already discussed this but I still believe the comment is incorrect. There is no doubt that the default value is not 1 but 0. Moreover, I do not think that the version part should be identical to that for `WXWIN_COMPATIBILITY_3_2`, as 3.0 compatibility should be abandoned one (stable) version before 3.2 compatibility is.

**wxUSE_EXTENDED_RTTI**
> Recommended setting: 0 (this is still work in progress...)

AFAIK, there has been no progress on this in many years and the feature was even considered for removal?

**wxUSE_COMPILER_TLS**
[The comment](https://github.com/wxWidgets/wxWidgets/blob/8a77fae68b8758bbdcd729e5680e0c4bd874bfb6/include/wx/setup_inc.h#L284) mentions issues with Windows XP but since wxWidgets 3.3 supports only Vista+, not only should the comment be changed but wx/msw/chkconf.h should be modified as well?

**wxUSE_INTL**
> i18n support: _() macro, wxLocale class. Requires wxTextFile.

Does it really require wxTextFile - I cannot see that?

**wxUSE_DIALUP_MANAGER**
> Requires wxUSE_DYNAMIC_LOADER.

This does not seem to be the case.

**wxUSE_DYNAMIC_LOADER**
> experimental, don't use for now

Is it still experimental? Is it even used in wxWidgets code? I found `dynload.h` included in several places (e.g.,` src\msw\gdiplus.cpp`) but AFAICT, the code uses just `wxDynamicLibrary` (`dynlib.h`) and nothing from `dynload.h`?

**wxUSE_XML**
> XML parsing classes. Note that their API will change in the future, so using wxXmlDocument and wxXmlNode in your app is not recommended.

Is the Note correct?

**wxUSE_GEOMETRY**
> Miscellaneous geometry code: needed for Canvas library

Affine transformation seem to be used by `wx{GC}DC`? See also` wxUSE_DC_TRANSFORM_MATRIX` remark below.

**wxUSE_FILEDLG**
> TODO: setting to choose the generic or native one

Remove?

**wxUSE_IPC, wxUSE_HELP, wxUSE_CONSTRAINTS, and others**
I wonder why the comments for these are written like that instead of above the define as usual?

**wxUSE_DC_TRANSFORM_MATRIX**
It seems this is in source code used for detecting availability of `wxAffineMatrix2D`, but `wxAffineMatrix2D` depends on `wxUSE_GEOMETRY`?


include/wx/setup_inc.h
```diff
@@ -1120,16 +1116,16 @@
...
+// then when bitmaps are used, at the expense of somewhat larger library size.
 //
 // Default is 1 under non-GTK ports. Under wxGTK the native art provider using
 // the GTK+ stock icons replaces it so it is normally not necessary.
 //
 // Recommended setting: 1 but can be turned off to reduce the library size.
 #define wxUSE_ARTPROVIDER_TANGO 1
 
-// Hotkey support (currently Windows only)
+// Hotkey support (currently Windows and macOS only)
```
PBfordev, line 1128: Moreover, the description in "wxUSE Preprocessor Symbols" in the docs should be updated and macOS added to the note in `wxWindow::UnRegisterHotKey()` docs.

vadz:
> I have updated some comments but there are still many which I think should be updated but I am not sure how.

Thanks for the updates, I don't have any comments concerning the changes so far, they should all be applied.

> **WXWIN_COMPATIBILITY_3_0**
> 
> > Default is 1 but please try building your code with 0 as the default will
> > change to 0 in the next version and the deprecated functions will disappear
> > in the version after it completely.
> 
> We have already discussed this but I still believe the comment is incorrect. There is no doubt that the default value is not 1 but 0. Moreover, I do not think that the version part should be identical to that for `WXWIN_COMPATIBILITY_3_2`, as 3.0 compatibility should be abandoned one (stable) version before 3.2 compatibility is.
 
Yes, this became outdated after 3.2 and wasn't fully updated. It should instead say something like this:
```cpp
// Default is 0, but you may still set it to 1 if you can't update the existing
// code relying on the deprecated functions. Please do consider updating it if
// at all possible instead of changing this to 1, however, as these functions
// will completely disappear in the next wxWidgets release.
#define WXWIN_COMPATIBILITY_3_0 0
```

> **wxUSE_EXTENDED_RTTI**
> 
> > Recommended setting: 0 (this is still work in progress...)
> 
> AFAIK, there has been no progress on this in many years and the feature was even considered for removal?

Yes, we should probably say
```cpp
// This feature is incompletely and experimental, please only enable it if
// you want to participate in its development.
//
// Recommended setting: 0 (unless you wish to try working on it).
```

> **wxUSE_COMPILER_TLS** [The comment](https://github.com/wxWidgets/wxWidgets/blob/8a77fae68b8758bbdcd729e5680e0c4bd874bfb6/include/wx/setup_inc.h#L284) mentions issues with Windows XP but since wxWidgets 3.3 supports only Vista+, not only should the comment be changed but wx/msw/chkconf.h should be modified as well?

We should drop this option and update the code to not use it any more, I've added this to #22689.

> **wxUSE_INTL**
> 
> > i18n support: _() macro, wxLocale class. Requires wxTextFile.
> 
> Does it really require wxTextFile - I cannot see that?

The only way to be really sure is to try building with `wxUSE_INTL==1` but `wxUSE_TEXTBUFFER==0`. But it will probably fail due to other reasons...

IOW I think the last part can indeed be removed but just because I'm almost certain that building with `wxUSE_TEXTBUFFER==0` must not work anyhow.

> **wxUSE_DIALUP_MANAGER**
> 
> > Requires wxUSE_DYNAMIC_LOADER.
> 
> This does not seem to be the case.

We probably should remove `wxUSE_DIALUP_MANAGER` completely...

> **wxUSE_DYNAMIC_LOADER**
> 
> > experimental, don't use for now
> 
> Is it still experimental? Is it even used in wxWidgets code? I found `dynload.h` included in several places (e.g.,` src\msw\gdiplus.cpp`) but AFAICT, the code uses just `wxDynamicLibrary` (`dynlib.h`) and nothing from `dynload.h`?

I'm almost sure nobody uses this one because it has never been documented or even demonstrated in a sample and so I'd be surprised if anybody really knew what it did (I think it tries to make wxRTTI in loadable plugins work, but I have no idea if it actually succeeds in doing it). AFAICS from a quick search, Audacity name drops it in their `ModuleManager.h` but doesn't reference it otherwise and nobody else really uses it neither.

So I think it should be deprecated too, unless somebody is motivated in reviving it and at least providing documentation and the tests.

> **wxUSE_XML**
> 
> > XML parsing classes. Note that their API will change in the future, so using wxXmlDocument and wxXmlNode in your app is not recommended.
> 
> Is the Note correct?

No, it probably won't change.

> **wxUSE_GEOMETRY**
> 
> > Miscellaneous geometry code: needed for Canvas library
> 
> Affine transformation seem to be used by `wx{GC}DC`? See also` wxUSE_DC_TRANSFORM_MATRIX` remark below.

We should just remove this option and always compile this in, there is relatively little code guarded by this option and nothing is gained by disabling it.

> **wxUSE_FILEDLG**
> 
> > TODO: setting to choose the generic or native one
> 
> Remove?

Yes.

> **wxUSE_IPC, wxUSE_HELP, wxUSE_CONSTRAINTS, and others** I wonder why the comments for these are written like that instead of above the define as usual?

Just because I haven't updated them when rewriting the descriptions of all the options in a more detailed way. I guess it should be done...

> **wxUSE_DC_TRANSFORM_MATRIX** It seems this is in source code used for detecting availability of `wxAffineMatrix2D`, but `wxAffineMatrix2D` depends on `wxUSE_GEOMETRY`?

Again, I didn't check it that carefully, but I think we should just remove this option and always compile the relevant code.

Thanks for looking at this!

PBfordev:
I will update the comments as suggested.

I will also attempt to deprecate `wxUSE_GEOMETRY`, `wxUSE_DC_TRANSFORM_MATRIX`, `wxUSE_DIALUP_MANAGER`, and `wxUSE_DYNAMIC_LOADER`.

But I am not sure what exactly needs to be done:
1. Remove the guards from the source code.
2. Mark the defines as deprecated (e.g., "Deprecated: Not used by wxWidgets since version 3.3.0") in comments in `setup_inc.h` and in the define list in the docs.
3. Anything else?

It's probably best if I delay this at least until  #22854 is merged.

BTW, I assume there is no way I can suspend CI builds for this PR?

vadz:
> I will update the comments as suggested.
> 
> I will also attempt to deprecate `wxUSE_GEOMETRY`, `wxUSE_DC_TRANSFORM_MATRIX`, `wxUSE_DIALUP_MANAGER`, and `wxUSE_DYNAMIC_LOADER`.

Great, TIA!

> But I am not sure what exactly needs to be done:
> 
>     1. Remove the guards from the source code.
> 
>     2. Mark the defines as deprecated (e.g., "Deprecated: Not used by wxWidgets since version 3.3.0") in comments in `setup_inc.h` and in the define list in the docs.
> 
>     3. Anything else?

I think we may just remove all mentions of them, as this is not going to break anything. Am I missing something here?

> It's probably best if I delay this at least until #22854 is merged.

I hope to do it relatively soon...

> BTW, I assume there is no way I can suspend CI builds for this PR?

Removing all the files in `.github/workflows` (or maybe all except for the one checking for the code style and spelling, as this one can still be useful to run), `appveyor.yml`, `.cirrus.yml` and `.circleci/config.yml` will prevent the CI builds from running for this branch. Of course, this commit would need to be reverted later, when merging.

Using `[skip-ci]` in the commit message will disable most, but not all, of them too.

PBfordev:
Thanks for the information on how to easily temporarily suppress the CI.

Regarding the deprecated defines
> I think we may just remove all mentions of them, as this is not going to break anything. Am I missing something here?
I thought we must keep them both defined in `setup.h` and mentioned in the docs because we could not assume there were no users ever using these in their code. But I guess it safe to assume that due to the nature of those?

vadz:
You're right that somebody could possibly check for the values of these symbols in their code, even though I don't remember ever seeing this. But it doesn't cost much to add some "Deprecated options" section to `include/wx/setup_inc.h` and just define them all there with a comment saying that they're defined for compatibility only but that changing them is not supported.

Or we could just do it in `wx/platform.h` itself to avoid having these "options", which can't be changed, in a file which is supposed to be editable.

PBfordev:
I have updated more comments and I may be done with the PR. I have still temporarily disabled CI, in case changes will be requested.

I was thinking about this and the rest of changes discussed here (e.g., deprecating some `wxUSE_xxx`) should be done in a separate PR, as it will also affect the code (the guards will be removed). This PR should really do only what is says, i.e., just update the comments.

Based on recent changes, the comments and default values for standard classes use defines are probably going to change as well, so that would be likely another PR.

vadz:
Thanks, I'm going to push this soon. I'm not sure if this should be backported to 3.2 branch: AFAICS everything here except for `WXWIN_COMPATIBILITY_3_0` comment is valid for it too, but I could be missing something, please let me know if I do.

FWIW I could cherry pick the commits from your branch without any problems (thinking about it again, I could have used `git merge --no-commit` to do it even simpler), but to avoid such problems in the future, please consider rebasing, i.e. running `git rebase master <your-branch>`, rather than merging master into your branch -- this is what you typically want to do if you need to integrate the changes on master into your working branch (but for short-lived branches you typically don't need to do neither the former not the latter, in fact).

PBfordev:
I can't see anything that could not be backported, aside for the already-mentioned `WXWIN_COMPATIBILITY_3_0` comment. But as I wrote before, I think the comments for both `WXWIN_COMPATIBILITY_2_8` and `WXWIN_COMPATIBILITY_3_0` are already wrong there as they are identical, i.e., similar issue as here.

`WXWIN_COMPATIBILITY_2_8` comment claims that the default is `1` while it actually is `0`. The wxWidgets versions where the two compatibilities are going to be removed should be different as well.

FWIW, I thought I did use rebase on my PR branch (with TortoiseGIT via its Fetch & Rebase), apparently not.

PR #22866, opened 2022-10-11
Pdf24

sz5000:
1) Adds method static wxPrintData GetDefaultPrintData(const wxString& printerName = wxEmptyString); to get default print data for a windows printer.

2) Disables wxFAIL_MSG(wxT("should have a top level parent!")); in wxSplitterWindow because we have a case where the parent is not a top level window.


sz5000:
Something went wrong with this pull request.... will add a new one later...

Issue #22870, opened 2022-10-13
MacOS standard "Window" menu is in the wrong place and sometimes moves around


nobugshere:
### Description

MacOS apps have a standard "Window" menu item, whether you want it or not :-).  By convention, it is placed to the left of the "Help" menu but, in apps using wxWidgits, it appears to the right (see first screenshot of the Hello World app). 

![wxDefault](https://user-images.githubusercontent.com/61946508/195725017-34da8cdc-2c20-46f0-b292-f884aac3e1e2.png)

In a more complicated app, the "Window" menu starts off on the right but jumps to the left of the "File" menu when the window is minimized then restored.  I haven't been able to reproduce that behavior in a simple test case.

If I insert code to add the "Window" menu right before adding the "Help" menu it seems to place it where I want it, and it gets the standard system entries.

![Hack](https://user-images.githubusercontent.com/61946508/195725448-8f6c7725-1fdc-400c-90e3-d8cc0b903b75.png)

However, in the complicated app, this trick works at first, but after a minimize/restore cycle, the menu stays in the right place but stops working.

It seems like this is something that should be handled by the framework unless the user needs to customize it.

### Platform and version information

- wxWidgets version  you use:
3.2.1 (installed via Homebrew)
- wxWidgets port  you use:
wxOSX
- OS  and its version:
M1 Max macOS Monterey 12.6 and Xcode 14.0.1

csomor:
Hi

You could turn the auto-window-menu off using `wxMenuBar::SetAutoWindowMenu(false)` but since all the standard window manipulation commands are in there, this should rarely be used ...

Right now I'm not reordering the menus in a menubar, only adding the apple menu in the front and the window menu at the end, so the recommended way to arrive at the correct order right now would be, e.g. if you'd take the minimal sample to add a dummy window menu

```
menuBar->Append(fileMenu, "&File");
menuBar->Append(new wxMenu(), "&Window");
menuBar->Append(helpMenu, "&Help");
```

I could create a PR which does an autoreorder, but I'm a little bit relucatant to do that

vadz:
I think we need to add "Window" menu at the end except if the last menu is "Help", in which case we should put it before it.

But it's not totally clear to me when does the problem arise currently, does it happen in any of our samples? If not, could you please provide a [minimal patch](https://www.wxwidgets.org/develop/how-to-submit-patches/) allowing to reproduce it?

TIA!

csomor:
[...]

PR #22871, opened 2022-10-14
macOS fix automatic menu order for window and help menu

csomor:
see #22870

### DIFF
--- a/src/osx/cocoa/menu.mm
+++ b/src/osx/cocoa/menu.mm
@@ -256,11 +256,29 @@ public :
 
         if ( windowMenu == nil )
         {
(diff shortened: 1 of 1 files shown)


vadz:
@csomor Should I merge this or are you still waiting for something before doing it?

Also, I guess this should be backported to 3.2 too, right?

csomor:
> @csomor Should I merge this or are you still waiting for something before doing it?

I thought I could try to make it work even when menus are removed completely, as in the #22870, but the current version is already better than what we had, so yes, I'm fine with merging this
 
> Also, I guess this should be backported to 3.2 too, right?

yes, thanks

PR #22873, opened 2022-10-15
Fix line numbers in Linux stack walk

vadz:
The old code didn't work for the symbols not in the executable itself, which wasn't very useful when using shared wx libraries.

Use the correct offset for the shared libraries and run addr2line separately for each module to fix this.

A better solution would be to implement addr2line functionality ourselves, which shouldn't be that difficult if we can use libdw, but this will have to wait.


vadz:
This could be backported to 3.2 if anybody is motivated to replace the use of `std::vector` and `std::unordered_map` in this code with `wxVector` (easy) and wx hash map (a bit less so). I probably won't be doing this myself though.

PR #22874, opened 2022-10-15
Switch to using Catch 2.x

vadz:
Now that we can afford it.


vadz:
Hmm, I'm going to postpone this until I replace all `NULL`s with `nullptr` because Catch v2 doesn't compile `CHECK( ptr == NULL )` any longer, so we have to use `nullptr` here.

vadz:
Just a heads up: I'll merge it soon now that master uses `nullptr` (after fixing the remaining warning).

PR #22875, opened 2022-10-15
editors.h: Fix missing include

lbartoletti:
At line https://github.com/wxWidgets/wxWidgets/blob/88d526660f2e7d3038fd0ff6eb0ae9b4c6e96656/include/wx/propgrid/editors.h#L473-L475 the statement will use `wxBitmapBundle` however  `wx/bmpbndl.h` is not included.

### DIFF
--- a/include/wx/propgrid/editors.h
+++ b/include/wx/propgrid/editors.h
@@ -13,6 +13,10 @@
 
 #include "wx/defs.h"
 
+#if wxUSE_BMPBUTTON
+#include "wx/bmpbndl.h"
+#endif
+
 #if wxUSE_PROPGRID
 
 #include "wx/window.h"
--- a/include/wx/propgrid/editors.h
+++ b/include/wx/propgrid/editors.h
@@ -13,14 +13,11 @@
 
 #include "wx/defs.h"
 
-#if wxUSE_BMPBUTTON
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for the fix but it's probably slightly better to forward declare the class here -- which should also be enough, shouldn't it? I.e. I'd rather apply
```diff
diff --git a/include/wx/propgrid/editors.h b/include/wx/propgrid/editors.h
index 9a42ef491e..8feca8c0ba 100644
--- a/include/wx/propgrid/editors.h
+++ b/include/wx/propgrid/editors.h
@@ -17,6 +17,7 @@

 #include "wx/window.h"

+class WXDLLIMPEXP_FWD_CORE wxBitmapBundle;
 class WXDLLIMPEXP_FWD_PROPGRID wxPGCell;
 class WXDLLIMPEXP_FWD_PROPGRID wxPGProperty;
 class WXDLLIMPEXP_FWD_PROPGRID wxPropertyGrid;
```

Could you please confirm that this fixes the problem too?

lbartoletti:
> ```diff
> 
> +class WXDLLIMPEXP_FWD_CORE wxBitmapBundle;
> ```

Indeed! Done.

Issue #22688, opened 2022-08-01
Remove wxMotif and wxGTK 1


vadz:
Nobody uses them any more except @joukj on OpenVMS but their compiler will almost certainly fail to deal with C++14 anyhow, so they will have to stay at 3.2 anyhow.

setdebug:
No, leave wxMotif alone. It isn't hurting anyone.

lanurmi:
> No, leave wxMotif alone. It isn't hurting anyone.

It isn't helping anyone either. Even the minimal sample fails to operate properly with wxMotif.

vadz:
@setdebug Do you use wxMotif? Do you know anybody who does?

vadz:
As an example of why I want -- and will -- do it, there are thousands of occurrences of `NULL` in wxGTK1 and wxMotif sources that I can't replace automatically with clang-tidy and I really don't want to spend hours on replacing them manually.

vadz:
This is done now, so closing.

ProfFan:
Was just trying to use wxMOTIF and it's sad to see this news. Guess I have to use an old commit now.

oneeyeman1:
@ProfFan ,
Motif port is unmaintained since a long time.
Which OS do you have so it comes with the Motif WM?

Thank you.

ProfFan:
@oneeyeman1 macOS, you can just do `brew install openmotif` and it works :)

ProfFan:
* install and run `startx /opt/homebrew/bin/mwm`

lanurmi:
> Was just trying to use wxMOTIF and it's sad to see this news. Guess I have to use an old commit now.

There is absolutely nothing that prevents you from using the 3.2 branch and enjoying wxMotif all day long.

PR #22877, opened 2022-10-16
Remove wxMotif and wxGTK1

vadz:
See #22688.


vadz:
I'd like to merge this relatively soon, please let me know if I should hold it off if anybody would like to do a real review of this change and needs time for it.

But I suspect that nobody is going to look at it really closely anyhow, so if nobody says anything, I'll merge it in a couple of days (after fixing any inevitable CI fallout...).

PR #22878, opened 2022-10-16
Test case for Window menu position and GetMenuCount

nobugshere:
This is a test case for Issue [#22870](https://github.com/wxWidgets/wxWidgets/issues/22870#issue-1408526348).


vadz:
FYI I'm tagging it as "work needed" just to indicate that this shouldn't be merged.

I also think that it's probably better to just attach patches like this to the corresponding ticket...

vadz:
Sorry, I'm going to close it because it is triggering new CI builds with every new merge and this is definitely not something we want.

nobugshere:
No problem.  Sorry if it cause any trouble.

PR #22879, opened 2022-10-16
Proper icon rendering for different DPIs using wxGenericTreeCtrl

gerstrong:
Fixes blurry rendering of icons in TreeCtrl on GTK using Wayland

### DIFF
--- a/samples/treectrl/treetest.cpp
+++ b/samples/treectrl/treetest.cpp
@@ -955,7 +955,7 @@ MyTreeCtrl::MyTreeCtrl(wxWindow *parent, const wxWindowID id,
 {
     m_reverseSort = false;
 
-    CreateImages(16);
+    CreateImages(32);
     CreateStateImageList();
 
     // Add some items to the tree
@@ -979,8 +979,6 @@ void MyTreeCtrl::CreateImages(int size)
     // should correspond to TreeCtrlIcon_xxx enum
     wxIcon icons[5];
 
-    const auto scaleFac = this->GetDPIScaleFactor();
-
     if (m_alternateImages)
     {
         icons[TreeCtrlIcon_File] = wxIcon(icon1_xpm);
@@ -992,10 +990,10 @@ void MyTreeCtrl::CreateImages(int size)
     else
     {
         icons[TreeCtrlIcon_File] =
-        icons[TreeCtrlIcon_FileSelected] = wxArtProvider::GetIcon(wxART_NORMAL_FILE, wxART_LIST, iconSize*scaleFac);
+        icons[TreeCtrlIcon_FileSelected] = wxArtProvider::GetIcon(wxART_NORMAL_FILE, wxART_LIST, iconSize);
         icons[TreeCtrlIcon_Folder] =
         icons[TreeCtrlIcon_FolderSelected] =
-        icons[TreeCtrlIcon_FolderOpened] = wxArtProvider::GetIcon(wxART_FOLDER, wxART_LIST, iconSize*scaleFac);
+        icons[TreeCtrlIcon_FolderOpened] = wxArtProvider::GetIcon(wxART_FOLDER, wxART_LIST, iconSize);
     }
 
     // Make a vector of bundles corresponding to the icons. We use a custom
--- a/samples/treectrl/treetest.cpp
+++ b/samples/treectrl/treetest.cpp
@@ -979,6 +979,8 @@ void MyTreeCtrl::CreateImages(int size)
     // should correspond to TreeCtrlIcon_xxx enum
     wxIcon icons[5];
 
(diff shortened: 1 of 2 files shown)


vadz:
Sorry, I'm not sure why is this is the right thing to do. Could you please explain how exactly do you set up images for your tree control? Ideal, as always, would be to reproduce in the treectrl sample (see `samples/treectrl/treetest.cpp`).

gerstrong:
I tested this against the wxAUI Sample application. Why do you think it is not right to it so?

gerstrong:
Just tried your mentioned test ```samples/treectrl/treetest.cpp```, and yes it looks better, although I had to load image with higher resolutions, since everything is set to 16x16 in order to see the improvement properly.

DiffHDPI

vadz:
Thanks, I do see the problem with your patch to the sample, the image is indeed pixel-doubled with the current sources.

I had questions about the original PR because I thought that `wxImageList` should work for drawing images with the correct scale factor, normally, and that if it didn't, it would be better to fix it rather than changing all the code using it.

But now that I think more about it, I think `wxImageList` just can't be used with the images of the same size in different resolutions without scaling them and we really need to use `wxBitmapBundle` for this, so your change is indeed correct. I'd like to tweak it a bit, however, before merging, to avoid using `GetDPIScaleFactor()` explicitly as this shouldn't be necessary and we should just use `GetBitmapBundle(image).GetBitmapFor(this)` instead. It also should be done for the state and button image lists too for consistency.

If you could please update the PR to do it like this, it would be great, otherwise I'll do it myself when I have time.

TIA!

gerstrong:
Okay, I updated the PR as you suggested. For avoiding ```wxImageList``` everywhere I would have to crawl the whole code if you want me to do that. If that is the case, I'd open another PR, if that is okay for you. Let know your thoughts. Other than that you are welcome.

Best Regards, Gerhard

samples/treectrl/treetest.cpp
```diff
@@ -979,6 +979,8 @@ void MyTreeCtrl::CreateImages(int size)
     // should correspond to TreeCtrlIcon_xxx enum
     wxIcon icons[5];
 
+    const auto scaleFac = this->GetDPIScaleFactor();
```
vadz, line 982: Just not to forget to do it: changes to this file should be reverted, they were useful for testing but not needed permanently.

src/generic/treectlg.cpp
```diff
@@ -2662,12 +2662,13 @@ void wxGenericTreeCtrl::PaintItem(wxGenericTreeItem *item, wxDC& dc)
...
-                                 item->GetX() + state_w,
-                                 item->GetY() +
-                                    (total_h > image_h ? (total_h-image_h)/2
-                                                       : 0),
-                                 wxIMAGELIST_DRAW_TRANSPARENT );
+
+        const wxCoord x = item->GetX() + state_w;
+        const wxCoord y = item->GetY() + (total_h > image_h ? (total_h-image_h)/2 : 0);
+
+        auto &img = GetImages().at(image);
```
vadz, line 2669: We really need to use `GetBitmapBundle()` here, otherwise it will crash, AFAICS, if the existing code uses `wxImageList`-based API and not `wxBitmapBundle`-based one for adding the images. I.e.
```suggestion
        auto &img = GetBitmapBundle(image);
```

vadz (changes requested):
Sorry for being unclear, I didn't meant to change it in all the code, but just in the same file as there are 2 more very similar calls to `Draw()` in it and it feels wrong to change one of them but not the other two.

If you could change the 2 other ones too, it would be great, but please note that, unless I'm missing something, we shouldn't be using `GetImages()` directly like this.

src/generic/treectlg.cpp
```diff
@@ -2381,11 +2381,13 @@ void wxGenericTreeCtrl::CalculateLineHeight()
         // Calculate a m_lineHeight value from the normal Image sizes.
         // May be toggle off. Then wxGenericTreeCtrl will spread when
         // necessary (which might look ugly).
-        int n = GetImageCount();
-        for (int i = 0; i < n ; i++)
+        auto &imgs = GetImages();
+        for ( auto &img : imgs )
```
vadz, line 2385: Sorry, but why did this have to be changed?

I'm afraid it's wrong because it breaks compatibility for code using `wxImageList`: `GetImages()` will be empty in this case, so the size of the images in the list won't be taken into account, unlike before.

gerstrong: Okay, got it. I reverted those lines.

vadz:
OK, thanks, this looks good, but looking at the code around it, there were a few more things I wanted to change, so I've done it in #22887 which also includes (a slightly modified version of) your commit from here. Could you please check if the other PR works for you? If so, I'll merge it soon. TIA!

Issue #22876, opened 2022-10-15
wxMSW regressions after enabling double buffering


vadz:
@vadz, I think I found first victims already, just by running the bundled samples.
`wxAuiToolBar::OnPaint()` asserts when creating `wxAutoBufferedPaintDC`:
> D:\Dev\Desktop\!Lib\wxWidgets-PB\include\wx/dcbuffer.h(229): assert "win->GetBackgroundStyle() == wxBG_STYLE_PAINT" failed in wxAutoBufferedPaintDC::wxAutoBufferedPaintDC(): You need to call SetBackgroundStyle(wxBG_STYLE_PAINT) in ctor, and also, if needed, paint the background in wxEVT_PAINT handler.

I also see several drawing issues when compared to the recentish master but I am not sure if they are caused by this commit:

- [x] When dragging an image in dragimg sample, neither image nor cursor are shown.
- [x] When dragging a card in the forty demo, the card is not shown.
- [x] When starting listctrl sample, a black rectangle is shown in place of the list for about a half a second. The black background can be also observed when resizing the list. IIRC, there was an issue like this with the control, fixed loong time ago.

_Originally posted by @PBfordev in https://github.com/wxWidgets/wxWidgets/issues/22851#issuecomment-1279823432_

vadz:
I've fixed 2 of the 3 problems, but I don't know what to do about forty. It's a nice demo, but it's totally incompatible with the way drawing should be done now and doesn't work at all with Wayland (nor, almost certainly, under macOS) and, in fact, works very poorly with GTK3 even with X11 backend.

I could add a call to `MSWDisableComposited()` to it, I guess, but the right thing to do would be to rewrite it entirely to avoid using `wxClientDC`, `wxDC::Blit()` etc (and also using `wxBitmapBundle`, SVG-based card images etc). Except who has time to do this :-(

vadz:
I've applied the "fix" to the demo too, but it will actually need to be applied to all samples and demos using `wxClientDC`, of which there are quite a few (some of them might not need fixing, but it's not obvious as long as we don't have a separate `wxInfoDC`, see #12486).

vadz:
I've disabled compositing for the demo in 3a7394fe08 (Disable use of composition in forty demo under MSW, 2022-10-17), so I think this can be closed now, even though I wouldn't be surprised at all if there are other/more problems with using it, see [this thread](https://groups.google.com/g/wx-dev/c/irUrLXU3_gQ).

PR #22882, opened 2022-10-17
MSW fixes after enabling WS_EX_COMPOSITED

vadz:
Fix some of the drawing/refresh problems after enabling double buffering, see #22876.

### DIFF
--- a/demos/forty/canvas.cpp
+++ b/demos/forty/canvas.cpp
@@ -35,6 +35,15 @@ FortyCanvas::FortyCanvas(wxWindow* parent, const wxPoint& pos, const wxSize& siz
              m_playerDialog(0),
              m_leftBtnDown(false)
 {
+#ifdef __WXMSW__
+    // Unfortunately the redraw logic here is incompatible with double
+    // buffering under MSW, so we have to disable it. The proper solution would
+    // be to rewrite this logic, as this is also required to make it work under
+    // GTK/Wayland and macOS, where the game currently doesn't update its
+    // display at all.
+    MSWDisableComposited();
+#endif
+
     SetScrollbars(0, 0, 0, 0);
 
 #ifdef __WXGTK__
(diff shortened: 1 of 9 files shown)


AliKet:
I think the following
https://github.com/wxWidgets/wxWidgets/blob/bc07c9a309272b79a8ad5aa4af2b977e8b0c663d/src/msw/notebook.cpp#L183-L219
and the paint handler
https://github.com/wxWidgets/wxWidgets/blob/bc07c9a309272b79a8ad5aa4af2b977e8b0c663d/src/msw/notebook.cpp#L779-L859
can be safely removed after merging this ?

vadz:
@AliKet have you already tested that it worked fine (i.e. without flickering) after setting `USE_NOTEBOOK_ANTIFLICKER` to 0?

Getting rid of this code would definitely be appealing but I'm afraid we may need to keep it for as long as we want to allow using `wxClientDC` with any of the windows in the notebook, as enabling composition for the notebook will probably prevent `wxClientDC` from working anywhere inside it. So we may actually need to still keep it, but only use it when `WS_EX_COMPOSITED` is not set.

AliKet:
> @AliKet have you already tested that it worked fine (i.e. without flickering) after setting `USE_NOTEBOOK_ANTIFLICKER` to 0?
> 

Yes, I tested it and didn't notice any flicker except we would have to keep subclassing the updown button to make it flicker free.

> Getting rid of this code would definitely be appealing but I'm afraid we may need to keep it for as long as we want to allow using `wxClientDC` with any of the windows in the notebook, as enabling composition for the notebook will probably prevent `wxClientDC` from working anywhere inside it. So we may actually need to still keep it, but only use it when `WS_EX_COMPOSITED` is not set.

AliKet:
> > @AliKet have you already tested that it worked fine (i.e. without flickering) after setting `USE_NOTEBOOK_ANTIFLICKER` to 0?
> 
> Yes, I tested it and didn't notice any flicker except we would have to keep subclassing the updown button to make it flicker free.
> 

let me correct myself: the notebook seems to work with no problems **only** when the tabs are in the default position (on top).

So yes the code above should stay as you said and sorry to bother you with that.

PR #22887, opened 2022-10-18
Generic wxTreeCtrl icons high DPI improvements

vadz:
Use the bitmap of appropriate size if available in high DPI instead of always scaling up the standard size image.

### DIFF
--- a/include/wx/withimages.h
+++ b/include/wx/withimages.h
@@ -129,6 +129,37 @@ public:
         return m_imageList;
     }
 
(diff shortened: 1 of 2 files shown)


gerstrong:
Looks good to me. I tested it on my system and the related samples programs render beautifully using gnome 42.2 and wayland with gtk. :-)

vadz:
Thanks for testing (and for the original fix, of course)!

PR #22889, opened 2022-10-18
Add a Docker container for regenerating wx configure

vadz:
This allows doing it on any system with Docker instead of having to do it only under Debian stable that was traditionally used for this.

---

I thought this could be useful to others, so I've created this docker file which seems to work well, i.e. regenerating configure using it doesn't result in any unwanted differences.

Would this actually be useful to have? If so, any comments about better Docker practices or whatever would be welcome, I don't know much about it.

### DIFF
--- a/build/tools/autoconf/Dockerfile
+++ b/build/tools/autoconf/Dockerfile
@@ -1,0 +1,10 @@
+FROM alpine:3.13
+
+RUN apk add autoconf
+
+# wxWidgets sources must be mounted here.
+VOLUME /wx
+
+WORKDIR /wx
+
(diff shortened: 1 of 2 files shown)


vadz:
Well, this didn't seem to have resulted in any enthusiasm, but I'm still going to merge this as it does no harm and I still find it useful, personally.

MaartenBent:
I usually generate it on Windows. I have to replace `autoconf` with `autoconf-2.69` in `autogen.sh`. And in the generated file the `runstatedir` variables are removed so I have to restore those.
I never looked into docker so I don't know if this could benefit me.

Maybe you can make it part of a CI check too? If `configure.in` is modified run the docker script to make sure the generated file is correct?

PR #22893, opened 2022-10-19
wxGrid refresh logic revisited

AliKet:
Replaces #22826 

**Not tested under MacOS!**

By the way, there are two commits not included here (because I need
to double check them before inclusion) which are:

1. This patch:
```
index bca2f1899e..eba6c4b0ab 100644
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -9941,42 +9941,7 @@ void wxGrid::DoSetRowSize( int row, int height )
 
     if ( ShouldRefresh() )
     {
-        // We need to check the size of all the currently visible cells and
-        // decrease the row to cover the start of the multirow cells, if any,
-        // because we need to refresh such cells entirely when resizing.
-        int topRow = row;
-
-        // Note that we don't care about the cells in frozen windows here as
-        // they can't have multiple rows currently.
-        const wxRect rect = m_gridWin->GetRect();
-        int left, right;
-        CalcUnscrolledPosition(rect.GetLeft(), 0, &left, NULL);
-        CalcUnscrolledPosition(rect.GetRight(), 0, &right, NULL);
-
-        const int posLeft = XToPos(left, m_gridWin);
-        const int posRight = XToPos(right, m_gridWin);
-        for ( int pos = posLeft; pos <= posRight; ++pos )
-        {
-            int col = GetColAt(pos);
-
-            int numRows, numCols;
-            if ( GetCellSize(row, col, &numRows, &numCols) == CellSpan_Inside )
-            {
-                // Notice that numRows here is negative.
-                if ( row + numRows < topRow )
-                    topRow = row + numRows;
-            }
-        }
-
-        int y;
-        CalcScrolledPosition(0, GetRowTop(topRow), NULL, &y);
-
-        // Refresh the lower part of the window below the y position
-        int cw, ch;
-        GetClientSize(&cw, &ch);
-
-        const wxRect updateRect(0, y, cw, ch - y);
-        Refresh(true, &updateRect);
+        RefreshArea(wxGA_RowLabels | wxGA_Cells);
     }
 }
 
@@ -10084,40 +10049,7 @@ void wxGrid::DoSetColSize( int col, int width )
 
     if ( ShouldRefresh() )
     {
-        // This code is symmetric with DoSetRowSize(), see there for more
-        // comments.
-
-        int leftCol = col;
-
-        const wxRect rect = m_gridWin->GetRect();
-        int top, bottom;
-        CalcUnscrolledPosition(0, rect.GetTop(), NULL, &top);
-        CalcUnscrolledPosition(0, rect.GetBottom(), NULL, &bottom);
-
-        const int posTop = YToPos(top, m_gridWin);
-        const int posBottom = YToPos(bottom, m_gridWin);
-        for ( int pos = posTop; pos <= posBottom; ++pos )
-        {
-            int row = GetRowAt(pos);
-
-            int numRows, numCols;
-            if ( GetCellSize(row, col, &numRows, &numCols) == CellSpan_Inside )
-            {
-                if ( col + numCols < leftCol )
-                    leftCol = col + numCols;
-            }
-        }
-
-        int x;
-        CalcScrolledPosition(GetColLeft(leftCol), 0, &x, NULL);
-
-        // Refresh the further part of the window to the right (LTR)
-        // or to the left (RTL) of the x position
-        int cw, ch;
-        GetClientSize(&cw, &ch);
-
-        const wxRect updateRect(x, 0, cw - x, ch);
-        Refresh(true, &updateRect);
+        RefreshArea(wxGA_ColLabels | wxGA_Cells);
     }
 }

```
My measurements revealed no difference between the two versions at all (optimized vs non-optimized refresh logic) and I'll leave it to someone else to decide which is better here.

2. The refresh performed in `wxGrid::AutoSizeColOrRow()` looks redundant to me, because it is already
done inside `DoSet{Row,Col}Size()` when `Set{Row,Col}Size()` is called! and I will not touch it for the moment for fear of breaking something inadvertently!!

**TODO:** consider using `RefreshRect()` instead.

### DIFF
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -1693,6 +1693,10 @@ public:
     void RefreshBlock(int topRow, int leftCol,
                       int bottomRow, int rightCol);
 
+    // Refresh one or more areas (a combination of wxGridArea enums) entirely.
+    void RefreshArea(int areas);
+
+
     // ------
     // Code that does a lot of grid modification can be enclosed
     // between BeginBatch() and EndBatch() calls to avoid screen
--- a/include/wx/generic/private/grid.h
+++ b/include/wx/generic/private/grid.h
@@ -31,6 +31,23 @@ WX_DECLARE_HASH_MAP_WITH_DECL(wxLongLong_t, wxGridCellAttr*,
                               wxIntegerHash, wxIntegerEqual,
                               wxGridCoordsToAttrMap, class WXDLLIMPEXP_CORE);
 
+// ----------------------------------------------------------------------------
+// enumerations
+// ----------------------------------------------------------------------------
+
+// Flags to pass to wxGrid::RefreshArea()
+enum wxGridArea
+{
+    wxGA_Corner      = 0x001, // m_cornerLabelWin
+    wxGA_RowLabels   = 0x002, // m_rowLabelWin, [m_rowFrozenLabelWin]
+    wxGA_ColLabels   = 0x004, // m_colLabelWin, [m_colFrozenLabelWin]
+    wxGA_Cells       = 0x008, // m_gridwin, [m_frozenCornerGridWin,
+                              // m_frozenColGridWin, m_frozenRowGridWin]
+
+    wxGA_Heading     = wxGA_ColLabels | wxGA_Corner,
+    wxGA_Labels      = wxGA_RowLabels | wxGA_Heading,
+    wxGA_All         = wxGA_Cells | wxGA_Labels
+};
 
 // ----------------------------------------------------------------------------
 // private classes
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -9122,8 +9122,7 @@ void wxGrid::SetCornerLabelValue( const wxString& s )
         m_table->SetCornerLabelValue( s );
         if ( ShouldRefresh() )
         {
-            wxRect rect = m_cornerLabelWin->GetRect();
-            m_cornerLabelWin->Refresh(true, &rect);
+            m_cornerLabelWin->Refresh();
         }
     }
 }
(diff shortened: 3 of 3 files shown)


AliKet:
@vadz I just tested this with `wxQt` on Linux and it doesn't work there unfortunately!

As a noob on `wxQt` my guess is that `wxWindowQt::Refresh()` doesn't account for child windows as is the case on `wxMSW` (which uses `RDW_ALLCHILDREN` flag) and `wxGTK` (which calls `gdk_window_invalidate_rect( ,  , true /*invalidate_children*/)`).
No idea what `Refresh()` does under `wxOSX` though!

AFAICS `wxQt` is now a **stopper** for this PR unless some qt guys decide to change `wxWindowQt::Refresh()` to follow `wxMSW/wxGTK/(wxOSX?)` behaviours !?

vadz:
Thank you, this looks good to me, but I'd like to test it under Mac before merging -- what's the simplest way to verify that refreshing still works correctly?

Also, if you could please make a PR with your proposed fix for #22899, it would be nice, as we could apply it first, before merging this one, to ensure `wxGrid` in wxQt isn't broken at all instead of being broken now and then fixed some time later...

Thanks again!

AliKet:
> Thank you, this looks good to me, but I'd like to test it under Mac before merging -- what's the simplest way to verify that refreshing still works correctly?
> 

If I haven't missed anything, the grid sample is a great help in testing this as it covers all aspects of the refresh scenarios performed here, i.e.:
- When resizing grid rows/columns
- When some grid attributes (fg/bg colours, fonts, alignments, etc...) are changed
- When showing/hiding cell editors

Also, the refresh will work properly even if part of the grid is frozen.

AFAICS, people are more comfortable to test master than pull requests unfortunately! so the best bet is to merge this and see what they report then...

vadz:
I've tested the grid sample under macOS and it seems to work fine there, so I'm going to merge this -- as you say, people will report problems only once this is merged into master in any case.

PR #22897, opened 2022-10-19
Remove mentions of C++11 in docs

TcT2k:
Since wxWidgets now requires C++11 it's not necessary to point out C++11 features as such explicitly.

BTW maybe it would make sense to deprecate `wxThread`, `wxCondition` and `wxMutex` as the docs now pretty much only recommend not to use them.

### DIFF
--- a/interface/wx/atomic.h
+++ b/interface/wx/atomic.h
@@ -16,8 +16,8 @@
 /**
     This function increments @a value in an atomic manner.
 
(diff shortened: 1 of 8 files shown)


vadz:
Thanks for cleaning this up!

I am not sure we need to deprecate the thread-related classes as the code using them doesn't _need_ to be changed, but OTOH we might consider reimplementing them using the standard ones.

PR #22901, opened 2022-10-21
Add support for wxUSE_TASKBARBUTTON option to configure

### LINKED ISSUE #22900: wxUSE_TASKBARBUTTON is 0 when building wxWidgets with MinGW on Linux
### Description

The problem is wxUSE_TASKBARBUTTON is set to 0 in generated "setup.h" when building wxWidgets with MinGW on Linux.

Reproducing steps:

* download the 3.2.1 release "wxWidgets-3.2.1.zip" and extract the zip
* switch to the directory where the extracted contents are
* do the following:
  ```
  mkdir build_mingw
  cd build_mingw
  ../configure --host=x86_64-w64-mingw32 --with-msw"
  ```
* check the generated "setup.h" and look for wxUSE_TASKBARBUTTON
  * wxUSE_TASKBARBUTTON is defined to be 0

Here's the [config.log](https://github.com/wxWidgets/wxWidgets/files/9837597/config.log)

### Note

wxUSE_TASKBARBUTTON doesn't appear in "configure" at all. Don't know if this is relevant.

- wxWidgets version: 3.2.1 release
- wxWidgets port: wxMSW
- OS: Arch Linux (every package updated to the latest)
- Compiler: x86_64-w64-mingw32-gcc (GCC) 12.2.0

vadz:
Add new --enable-taskbarbutton option and define wxUSE_TASKBARBUTTON as 1 by default instead of always defining it as 0, as was the case before.

Closes #22900.

### DIFF
--- a/configure
+++ b/configure
@@ -1365,6 +1365,7 @@ enable_ico_cur
 enable_dccache
 enable_ps_in_msw
 enable_ownerdrawn
+enable_taskbarbutton
 enable_uxtheme
 enable_wxdib
 enable_webviewie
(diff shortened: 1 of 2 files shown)


vadz:
Merged as ee67242b5f

PR #22902, opened 2022-10-21
Various wxQt fixes

vadz:


MaartenBent:
Should this be added to CMake too?

If I remember correctly you have to manually add the new sections (`QT_UNIX_SRC` and `QT_UNIX_HDR`) to `files.cmake`. Then upmake will fill them with the files.
And then add them to the project the same way as is done for the win32 sources:
https://github.com/wxWidgets/wxWidgets/blob/0a33da8058150df7e0ee4acf30385c3034029e97/build/cmake/lib/core/CMakeLists.txt#L53-L55

vadz:
> Should this be added to CMake too?

Thanks for the reminder, done now.

PR #22904, opened 2022-10-22
Use wxID_HIGHEST+1 in samples

Blake-Madden:
The samples inconsistently assign `wxID_HIGHEST` or `wxID_HIGHEST+1` to the start of ID enums.  The docs state:

> When using an enum to define a number of custom IDs, assigning the value of wxID_HIGHEST+1 to the first element ensures that none of the enum elements will conflict with any standard IDs.

This PR sets everything `wxID_HIGHEST+1`.


Blake-Madden:
FYI, there is one more in the "validate" sample, but there's some sort of bit shifting going on with `wxID_HIGHEST-1` in there and I could not figure out that the intent was.  I just left that sample alone.

vadz:
Thanks, but I have an alternative suggestion: what if we just redefined `wxID_HIGHEST` as the first ID _not_ used by wx? It's a bit weird, but OTOH it looks like plenty of code (and not just our samples) start their own IDs from `wxID_HIGHEST` and not `wxID_HIGHEST+1` (or higher), so it looks like it would be a more useful change.

Blake-Madden:
Yes, using `wxID_HIGHEST` would be better for compatibility with existing code I suppose.  I can make a new PR for the samples and documentation, but I'm not sure what wx internals would need to be changed (if any).

vadz:
I've created #22909 doing this, if you don't see any problems with it, let's merge it and close this one (and if you do, please comment on that PR there, TIA).

vadz:
I've merged the other PR in 99c52403f9ce649ecece891d76bf0d32c2046d19 now, so closing this one.

PR #22905, opened 2022-10-22
Improved DPIs support for wxListbook

gerstrong:
This makes the ```wxListbook``` rendering of images look less blurry, when higher DPI than 1.0 is used.

Let me know your thoughts


gerstrong:
Sorry, The MSW Tests do not pass yet: Checking now how to fix that...

PR #22907, opened 2022-10-23
List ctrl improvements for different DPIs

gerstrong:
This makes ```wxListView``` to render icons in HIDPI look better.

### DIFF
--- a/include/wx/generic/private/listctrl.h
+++ b/include/wx/generic/private/listctrl.h
@@ -640,6 +640,7 @@ public:
     void DrawImage( int index, wxDC *dc, int x, int y );
     void GetImageSize( int index, int &width, int &height ) const;
 
+    void SetImages(wxWithImages *images, const int which);
     void SetImageList( wxImageList *imageList, int which );
     void SetItemSpacing( int spacing, bool isSmall = false );
     int GetItemSpacing( bool isSmall = false );
@@ -813,6 +814,9 @@ protected:
     wxColour            *m_highlightColour;
     wxImageList         *m_small_image_list;
     wxImageList         *m_normal_image_list;
+
+    wxWithImages         *m_normal_images = nullptr;
+
     int                  m_small_spacing;
     int                  m_normal_spacing;
     bool                 m_hasFocus;
--- a/include/wx/generic/private/listctrl.h
+++ b/include/wx/generic/private/listctrl.h
@@ -640,8 +640,7 @@ public:
     void DrawImage( int index, wxDC *dc, int x, int y );
     void GetImageSize( int index, int &width, int &height ) const;
 
-    void SetImages(const wxVector<wxBitmapBundle> &images, const int which);
-    void SetImageList( wxImageList *imageList, int which );
+    void SetImages( wxWithImages *images, const int which );
     void SetItemSpacing( int spacing, bool isSmall = false );
     int GetItemSpacing( bool isSmall = false );
 
@@ -696,7 +695,7 @@ public:
     }
 
     wxImageList* GetSmallImageList() const
-        { return m_small_image_list; }
+        { return m_small_images ? m_small_images->GetImageList() : nullptr; }
 
     // set the scrollbars and update the positions of the items
     void RecalculatePositions();
@@ -812,10 +811,8 @@ protected:
     bool                 m_dirty;
 
     wxColour            *m_highlightColour;
-    wxImageList         *m_small_image_list;
-    wxImageList         *m_normal_image_list;
-
-    wxVector<wxBitmapBundle> m_normal_images;
+    wxWithImages        *m_small_images;
+    wxWithImages        *m_normal_images;
 
     int                  m_small_spacing;
     int                  m_normal_spacing;
--- a/include/wx/generic/private/listctrl.h
+++ b/include/wx/generic/private/listctrl.h
@@ -694,8 +694,12 @@ public:
         SetItem( info );
     }
 
-    wxImageList* GetSmallImageList() const
-        { return m_small_images ? m_small_images->GetImageList() : nullptr; }
+/*    wxImageList* GetSmallImageList() const
+        { return m_small_images ? m_small_images->GetImageList() : nullptr; }*/
+
+    wxWithImages* GetSmallImages() const
+        { return m_small_images; }
+
 
     // set the scrollbars and update the positions of the items
     void RecalculatePositions();
--- a/include/wx/generic/private/listctrl.h
+++ b/include/wx/generic/private/listctrl.h
@@ -640,7 +640,7 @@ public:
     void DrawImage( int index, wxDC *dc, int x, int y );
     void GetImageSize( int index, int &width, int &height ) const;
 
-    void SetImages(wxWithImages *images, const int which);
+    void SetImages(const wxVector<wxBitmapBundle> &images, const int which);
     void SetImageList( wxImageList *imageList, int which );
     void SetItemSpacing( int spacing, bool isSmall = false );
     int GetItemSpacing( bool isSmall = false );
@@ -815,7 +815,7 @@ protected:
     wxImageList         *m_small_image_list;
     wxImageList         *m_normal_image_list;
 
-    wxWithImages         *m_normal_images = nullptr;
+    wxVector<wxBitmapBundle> m_normal_images;
 
     int                  m_small_spacing;
     int                  m_normal_spacing;
--- a/src/common/listctrlcmn.cpp
+++ b/src/common/listctrlcmn.cpp
@@ -296,7 +296,7 @@ void wxListCtrlBase::SetNormalImages(const wxVector<wxBitmapBundle>& images)
 {
     m_imagesNormal.SetImages(images);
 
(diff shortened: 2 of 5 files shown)


include/wx/generic/private/listctrl.h
```diff
@@ -813,6 +814,9 @@ class wxListMainWindow : public wxWindow
     wxColour            *m_highlightColour;
     wxImageList         *m_small_image_list;
     wxImageList         *m_normal_image_list;
+
+    wxVector<wxBitmapBundle> m_normal_images;
```
vadz, line 818: There shouldn't be any need to store these images here, they're already stored in `m_normal_image_list`, why do we need to duplicate them?

vadz (changes requested):
Sorry, but the changes here seem to be wrong to me. It looks like you're (partially) redoing everything `wxWithImages` already does for you: this class encapsulates the difference between using `SetImageList()` and `SetImages()` for you, you just need to use its `GetImageLogicalSize()` and `GetImageBitmapFor()` methods in the code, what's the problem with doing this?

include/wx/generic/private/listctrl.h
gerstrong: @vadz I tried to use ```wxWithImages```. My problem is that you have both normal and small image list objects within one class. If I derive ```wxListMainWindow``` from the class ```wxWithImages``` additionally (like seen for other classes) I would only have one instance I could deal with. What I could do is to create a new class derived from ```wxWithImages``` and create two instances from that (not sure if that would be a good design choice). I also saw ```wxWithImages``` is meant to be not copied (no proper copy-constructors). So I am also a bit unsure about the design of that ```wxWithImages``` on that side. 

Can you guide me a bit on how you think it should be implemented properly seen? I'd be happy to get it coded as intended. At least what I have here works with higher DPI, but only on normal icon size. I tried with small icons but ran into a bunch of crashes, so I left those changes out for now... (but maybe it can be implemented properly as well). I tested this against your samples and it looks good and I also tested it against Money Manager EX and it really gives the application a quality push ;-)

Let me know your thoughts :-)

vadz: We indeed need to do some more refactoring done here, as we must get rid of `m_{small,normal}_image_list` in this class and use `wxListCtrlBase::m_images{Small,Normal}` instead, as we won't be able to deal with high DPI icons correctly as long as `wxImageList` is used directly -- we should only use it via `wxWithImages`. To minimize the number of changes we probably should make the existing `wxImageList*` members `wxWithImages*` instead and make them point to the existing base class objects.

Would this work?

gerstrong: Okay, I tried your suggestions and so far it seems to work. Hope the tests pass. Let's see...

gerstrong: Looking good. Let me know your thoughts.

include/wx/generic/private/listctrl.h
```diff
@@ -695,7 +695,7 @@ class wxListMainWindow : public wxWindow
     }
 
     wxImageList* GetSmallImageList() const
-        { return m_small_image_list; }
+        { return m_small_images ? m_small_images->GetImageList() : nullptr; }
```
vadz, line 698: I think we should get rid of this function, it seems wrong that it's still used.

src/common/listctrlcmn.cpp
```diff
@@ -292,7 +292,7 @@ wxItemAttr* wxListCtrlBase::OnGetItemColumnAttr(long item, long WXUNUSED(column)
 // Images support
 // ----------------------------------------------------------------------------
 
-void wxListCtrlBase::SetNormalImages(const wxVector<wxBitmapBundle>& images)
+void wxListCtrlBase::SetNormalImages(const wxVector<wxBitmapBundle> &images)
```
vadz, line 295: Let's please avoid such whitespace-only changes (I can filter it out myself before merging, this is mostly for the future).

src/generic/listctrl.cpp
```diff
@@ -3302,41 +3302,53 @@ void wxListMainWindow::OnKillFocus( wxFocusEvent &WXUNUSED(event) )
...
 
 void wxListMainWindow::GetImageSize( int index, int &width, int &height ) const
 {
-    if ( HasFlag(wxLC_ICON) && m_normal_image_list )
+    if ( HasFlag(wxLC_ICON) && m_normal_images )
     {
-        m_normal_image_list->GetSize( index, width, height );
+        const auto size = m_normal_images->GetImageLogicalSize(this, index);
+        width = size.GetWidth();
+        height = size.GetHeight();
```
vadz, line 3333: We have an overload taking `width` and `height` directly, i.e. this could be just
```suggestion
        m_normal_images->GetImageLogicalSize(this, index, width, height);
```

(and the same thing below, of course).

src/generic/listctrl.cpp
```diff
@@ -3345,28 +3357,29 @@ void wxListMainWindow::GetImageSize( int index, int &width, int &height ) const
...
     // calc the spacing from the icon size
     int width = 0;
 
-    if ((imageList) && (imageList->GetImageCount()) )
+    if ((images) && (images->HasImages()) )
     {
-        int height;
-        imageList->GetSize(0, width, height);
+        const auto size = images->GetImageLogicalSize(this, 0);
+        width = size.GetWidth();
```
vadz, line 3371:
```suggestion
        int height;
        images->GetImageLogicalSize(this, 0, width, height);
```

src/generic/listctrl.cpp
```diff
@@ -3403,10 +3416,9 @@ wxListMainWindow::ComputeMinHeaderWidth(const wxListHeaderData* column) const
     const int image = column->GetImage();
     if ( image != -1 )
     {
-        if ( m_small_image_list )
+        if ( m_small_images )
         {
-            int ix = 0, iy = 0;
-            m_small_image_list->GetSize(image, ix, iy);
+            const int ix = m_small_images->GetImageLogicalSize(this, image).GetWidth();
```
vadz, line 3421: We could minimize the code churn here as well:
```suggestion
            int ix = 0, iy = 0;
            m_small_images->GetImageLogicalSize(this, image, ix, iy);
```

src/generic/listctrl.cpp
```diff
@@ -4626,10 +4638,10 @@ void wxListMainWindow::InsertItem( wxListItem &item )
     {
         // Reset the buffered height if it's not big enough for the new image.
         int image = item.GetImage();
-        if ( m_small_image_list && image != -1 && InReportView() )
+        if ( m_small_images && image != -1 && InReportView() )
         {
-            int imageWidth, imageHeight;
-            m_small_image_list->GetSize(image, imageWidth, imageHeight);
+            const auto size = m_small_images->GetImageLogicalSize(this, image);
+            const int imageHeight = size.GetHeight();
```
vadz, line 4644: And here:
```suggestion
            int imageWidth, imageHeight;
            m_small_images->GetImageLogicalSize(this, image, imageWidth, imageHeight);
```

vadz (commented):
Thanks a lot, this looks much better!

I think we need to get rid of `GetSmallImageList()`, but otherwise I'd be glad to merge it -- although I'd prefer to apply the suggestions below to minimize the unnecessary changes to the code.

TIA!

Note to self: we'd need to avoid using `auto` if we want to backport this 
to 3.2.

vadz:
Looks good to me, thanks! I'll merge this soon.

src/generic/listctrl.cpp
```diff
@@ -1177,14 +1177,13 @@ void wxListHeaderWindow::OnPaint( wxPaintEvent &WXUNUSED(event) )
...
-                        xAligned + wLabel - ix - HEADER_IMAGE_MARGIN_IN_REPORT_MODE,
-                        HEADER_OFFSET_Y + (h - iy)/2,
-                        wxIMAGELIST_DRAW_TRANSPARENT
-                       );
+            wxBitmap bmp = imageList->GetImageBitmapFor(this, image);
+            dc.DrawBitmap(
+                          bmp,
+                          xAligned + wLabel - ix - HEADER_IMAGE_MARGIN_IN_REPORT_MODE,
+                          HEADER_OFFSET_Y + (h - iy)/2,
+                          wxIMAGELIST_DRAW_TRANSPARENT
```
vadz, line 1185: Just for the record: I'm removing this, it's wrong to use it as a `bool useMask`, and adding `true` here and to all the other places where `DrawBitmap()` is called instead.

include/wx/generic/private/listctrl.h
```diff
@@ -694,8 +694,12 @@ class wxListMainWindow : public wxWindow
         SetItem( info );
     }
 
-    wxImageList* GetSmallImageList() const
-        { return m_small_image_list; }
+/*    wxImageList* GetSmallImageList() const
+        { return m_small_images ? m_small_images->GetImageList() : nullptr; }*/
```
vadz, line 698: Also removing this.

vadz (commented):
I've finally created #22916 subsuming this PR, so closing it and will merge the other one when the CI builds pass.

PR #22909, opened 2022-10-23
Use usual half-open interval for wxID_LOWEST..wxID_HIGHEST

vadz:
The reserved wx IDs were previously in (wxID_LOWEST, wxID_HIGHEST] range, which was very unusual, as C++ always uses intervals half-open in the other direction, i.e. [begin, end).

It also conflicted (in theory, if not in practice, as wxID_HIGHEST wasn't actually used for anything) with the common practice, used even by wx own samples, of starting the custom IDs at wxID_HIGHEST rather than wxID_HIGHEST+1.

So change this to use the standard convention, retroactively making all the code using wxID_HIGHEST valid -- this seems better than insisting on starting the IDs from wxID_HIGHEST+1.

---

This replaces #22904.

### DIFF
--- a/include/wx/defs.h
+++ b/include/wx/defs.h
@@ -1779,10 +1779,10 @@ enum wxStandardID
     wxID_ANY = -1,
 
 
-    /* all predefined ids are between wxID_LOWEST and wxID_HIGHEST */
-    wxID_LOWEST = 4999,
+    /* all predefined ids are between wxID_LOWEST and wxID_HIGHEST (exclusive) */
+    wxID_LOWEST = 5000,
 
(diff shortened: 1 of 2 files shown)


Blake-Madden:
This looks good to me; the standard IDs still work OK for me on Windows at least.
After this is accepted, I'll make a new PR to homogenize the samples to this new convention.

PBfordev:
When moving the ID range, perhaps it could be a good time to fix low-priority, long-standing #18589 (`wxRichTextCtrl` using `wxID_HIGHEST` to `wxID_HIGHEST` + `3`)?

PR #22913, opened 2022-10-23
Update strftime() link in wxDateTime::Format() docs

### LINKED ISSUE #22912: Hyperlink in the wxDateTime page not working
The wxDateTime::Format help links to here,
https://cplusplus.com/reference/clibrary/ctime/strftime.html

But the link is no longet valid and needs to point here
https://cplusplus.com/reference/ctime/strftime/

PBfordev:
Closes #22912.

Could be backported to 3.2

### DIFF
--- a/interface/wx/datetime.h
+++ b/interface/wx/datetime.h
@@ -852,7 +852,7 @@ public:
 
     /**
         This function does the same as the standard ANSI C @c strftime(3)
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, will apply soon.

FWIW I prefer using https://en.cppreference.com/ but I guess it doesn't matter what description of `strftime()` we link to.

PR #22916, opened 2022-10-25
Support high DPI images in generic wxListCtrl

vadz:
This is a slightly cleaned up #22907 with some extra refactorings on top.


vadz:
This could be backported to 3.2 but I'd like to wait a bit to see if no problems are found with these changes.

PR #22917, opened 2022-10-26
Always use `thread_local`

vadz:
Simplify the code and remove our custom implementation of thread-specific variables which is not needed any more with C++11.

See #22689.


eranif:
FYI: one of the reasons I stopped using GCC MinGW is because of crashes I had when using `thread_local`. You can run a quick google search on `thread_local` and mingw gcc and you will see that the issue still occurs

vadz:
@eranif Does this apply only to classic MinGW or to MinGW-w64/TDM-GCC too? If it affects the latter ones too, it's indeed a big problem.

What do you use with them if you don't use `thread_local`? Does wx implementation work well for you?

eranif:
AFAIK, TDM GCC was discontinued so its no relevant. I have stopped using MinGW-w64 a while ago and I am installing all the tools that I need from MSYS2. 

On Windows, I am building wxWidgets + CodeLite using clang64 from MSYS2/pacman. Clang does not have this bug.

I never used the wx implementation:(

vadz:
I don't know what to do about this. At the very least, I think we should use `thread_local` for non-MinGW compilers, but it seems a pity to keep our implementation just for it.

I don't know how to check if this bug affects our use of it (which is limited to only what we do in `src/common/threadinfo.cpp`). If it doesn't, I'd be tempted to merge this nevertheless...

vadz:
I've looked at this closer and it seems that the problem only exists in 32 bits and there are attempts to fix it there too. Moreover, I can't reproduce it with our use of thread-specific variables in `wxThreadSpecificInfo`, at least the following contrived test program, which uses `wxThreadSpecificInfo` from both `wxLog::EnableLogging()` and `wxTranslations::GetUntranslatedString()`, works fine for me:
```cpp
#include <wx/init.h>
#include <wx/crt.h>
#include <wx/log.h>
#include <wx/thread.h>
#include <wx/translation.h>

int main(int argc, char **argv)
{
    wxInitializer init;
    if ( !init.IsOk() ) {
        fprintf(stderr, "Failed to initialize wxWidgets.\n");
        return 3;
    }

    struct MyThread : wxThread {
        MyThread() : wxThread{wxTHREAD_JOINABLE} {}

        void* Entry() override {
            wxLog::EnableLogging(enable_);
            wxLogMessage("Worker thread says %s", wxTranslations::GetUntranslatedString("hi"));
            return nullptr;
        }

        bool enable_ = true;
    } t1, t2;
    t2.enable_ = false;
    t1.Run();
    t2.Run();

    wxLogMessage("Main thread says bye");
    return 0;
}
```

So I'm going to merge this and hope things work at least as well as before.

PR #22918, opened 2022-10-26
Make wxQt::Refresh() compliant with the documentation

AliKet:
`wxWindow::Refresh()` documentation says:
> Causes this window, and all of its children recursively, to be repainted.

This commit add the missing recursion explicitly as there is no way to tell **Qt** to obey to the wx convention of refreshing windows.

see: https://forum.qt.io/topic/7891/how-to-forcibly-update-child-widgets-in-parent


vadz:
Thanks, let's merge this into master -- and, in principle, it could be backported to 3.2 if range loops were replaced with their C++98 equivalents.

PR #22920, opened 2022-10-27
Drop ANSI build

vadz:


vadz:
I'd like to merge this soon (before Sunday). There are a lot of changes here, of course, but most of them are quite straightforward and while I wouldn't be surprised at all if this breaks something, it can/will be fixed later, while I don't think any problems will be found before it's merged.

But please let me know if anybody sees anything really wrong with this.

PBfordev:
I don't see anything really wrong (I did not review the files) but ass usual, here's my few unsolicited nitpicky suggestions:
1. Simplify setting `wxSuffix`  (i.e., always assume Unicode) in _build/msw/wx_setup.props_. Not sure if it is possible to show an error from a prop sheet when Unicode is not the charset used in the project.
2. Remove `wxOnAssert(const wxChar*,...)` overload from _include/wx/debug.h_, its comment says it is used only for 2.8 compatibility or ANSI builds, neither is relevant in 3.3.
3. Adapt texts mentioning `$(unicode)` in _docs/contributing/about-platform-toolkit-and-library-names.md_.

There is also section `overview_mbconv_string` in _docs/doxygen/overviews/mbconvclasses.h_ but it was not updated in a long time, so it perhaps can be left alone again.

vadz:
> I don't see anything really wrong (I did not review the files) but as usual, here's my few unsolicited nitpicky suggestions:

Thanks for looking at this and for the suggestions!
 
>     1. Simplify setting `wxSuffix`  (i.e., always assume Unicode) in _build/msw/wx_setup.props_.

Done (will push soon), thanks!

> Not sure if it is possible to show an error from a prop sheet when Unicode is not the charset used in the project.

It is, `wxwidgets.props` does it in its initial target, but I'm not sure if it's worth doing it in `wx_setup.props` itself, as I guess every initial target is always evaluated. But I'll add it to that file.

>     2. Remove `wxOnAssert(const wxChar*,...)` overload from _include/wx/debug.h_, its comment says it is used only for 2.8 compatibility or ANSI builds, neither is relevant in 3.3.

Yes, it indeed looks like this can be removed, thanks!
 
>     3. Adapt texts mentioning `$(unicode)` in _docs/contributing/about-platform-toolkit-and-library-names.md_.

Done, thanks!
 
> 
> There is also section `overview_mbconv_string` in _docs/doxygen/overviews/mbconvclasses.h_ but it was not updated in a long time, so it perhaps can be left alone again.

I did update it quickly, but it certainly could be improved more...

Thanks again!

PBfordev:
Inspired by the recent commits, I have also grepped for "ANSI", found few parts they may need to be updated (nothing serious)

Simplify https://github.com/wxWidgets/wxWidgets/blob/f4d93e5afddd95eecefb87edcf86ba3ada55e42a/include/wx/build.h#L44-L50

_docs/base/readme.txt_ seems very outdated in many regards (e.g., platforms and compilers supported).

Remove first sentence from
https://github.com/wxWidgets/wxWidgets/blob/f4d93e5afddd95eecefb87edcf86ba3ada55e42a/docs/doxygen/overviews/resyntax.h#L198-L201 but this document was probably not updated after switching to PCRE.

https://github.com/wxWidgets/wxWidgets/blob/f4d93e5afddd95eecefb87edcf86ba3ada55e42a/interface/wx/cmdline.h#L456-L458

https://github.com/wxWidgets/wxWidgets/blob/f4d93e5afddd95eecefb87edcf86ba3ada55e42a/interface/wx/datstrm.h#L195-L201

There are several mentions of  ANSI build in _interface/wx/string.h_.

Few outdated comments regarding ANSI builds in _src/common/string.cpp_, _src/common/strvarag.cpp_ (not sure if it affects code), and in some files "ANSI" should perhaps be changed to "utf-8" in comments regarding builds...

Possible code fix after dropping ANSI build in `wxVariantDataChar::Write()` and `wxVariantDataChar::Read(`), see the comments in _src/common/variant.cpp_.

And lastly: Perhaps  `wxDynamicLibrary::GetSymbolAorW()` is to be deprecated?

vadz:
> Inspired by the recent commits, I have also grepped for "ANSI", found few parts they may need to be updated (nothing serious)
> 
> Simplify
> 
> https://github.com/wxWidgets/wxWidgets/blob/f4d93e5afddd95eecefb87edcf86ba3ada55e42a/include/wx/build.h#L44-L50

Thanks, updated.

> _docs/base/readme.txt_ seems very outdated in many regards (e.g., platforms and compilers supported).
> 
> Remove first sentence from
> 
> https://github.com/wxWidgets/wxWidgets/blob/f4d93e5afddd95eecefb87edcf86ba3ada55e42a/docs/doxygen/overviews/resyntax.h#L198-L201
> but this document was probably not updated after switching to PCRE.

I'm leaving these two for later, they both (but especially the latter) need to be completely rewritten or maybe just removed and replaced with a link to PCRE syntax.

> https://github.com/wxWidgets/wxWidgets/blob/f4d93e5afddd95eecefb87edcf86ba3ada55e42a/interface/wx/cmdline.h#L456-L458
> 
> https://github.com/wxWidgets/wxWidgets/blob/f4d93e5afddd95eecefb87edcf86ba3ada55e42a/interface/wx/datstrm.h#L195-L201

Fixed, thanks.

> There are several mentions of ANSI build in _interface/wx/string.h_.

Fixed with minimal changes too.

> Few outdated comments regarding ANSI builds in _src/common/string.cpp_, _src/common/strvarag.cpp_ (not sure if it affects code), and in some files "ANSI" should perhaps be changed to "utf-8" in comments regarding builds...

Fixed some, but not all of them, I'm afraid I don't have time to deal with format conversions carefully enough right now...

> Possible code fix after dropping ANSI build in `wxVariantDataChar::Write()` and `wxVariantDataChar::Read(`), see the comments in _src/common/variant.cpp_.

I've considered this already, but actually doing it requires adding `wxUniChar` support to the stream classes and I'm not sure if we really need it. Alternatively, we could make `wxVariantDataChar::m_value` a `wchar_t`, but this is a bit weird too, so, finally, I've just left it like this.

> And lastly: Perhaps `wxDynamicLibrary::GetSymbolAorW()` is to be deprecated?

I think this name is not too bad because most Windows programmers should understand what it means and it might be used outside of wx itself, so I've also decided to leave it alone.

Thanks again!

PBfordev:
> > And lastly: Perhaps `wxDynamicLibrary::GetSymbolAorW()` is to be deprecated?
> 
> I think this name is not too bad because most Windows programmers should understand what it means and it might be used outside of wx itself, so I've also decided to leave it alone.

I meant to deprecate the two methods, since their name became deceiving as there is now no way to retrieve `A` symbols.

vadz:
I'll merge this one without doing it for now, but I've created #22930 for later consideration.

PR #22922, opened 2022-10-27
Make sure wxWindow::Refersh() works as documented

AliKet:


tests/controls/windowtest.cpp
```diff
@@ -459,3 +460,81 @@ TEST_CASE_METHOD(WindowTestCase, "Window::SizerErrors", "[window][sizer][error]"
...
+TEST_CASE_METHOD(WindowTestCase, "Window::Refresh", "[window]")
+{
+    wxWindow* const parent = m_window;
+    wxWindow* const child1 = new wxWindow(parent, wxID_ANY, wxPoint(10, 20), wxSize(80, 50));
+    wxWindow* const child2 = new wxWindow(parent, wxID_ANY, wxPoint(110, 20), wxSize(80, 50));
+    wxWindow* const child3 = new wxWindow(parent, wxID_ANY, wxPoint(210, 20), wxSize(80, 50));
+
+    m_window->SetSize(300, 100);
+
+#if wxDEBUG_LEVEL
```
vadz, line 473: Any reason we couldn't always do this? AFAICS it doesn't do any harm in non-debug build build too, does it?

tests/controls/windowtest.cpp
```diff
@@ -459,3 +460,81 @@ TEST_CASE_METHOD(WindowTestCase, "Window::SizerErrors", "[window][sizer][error]"
...
+    child3->SetBackgroundColour(*wxGREEN);
+#endif
+
+    // Notice that using EventCounter here will give incorrect results,
+    // so we have to bind each window to a distinct event handler instead.
+
+    static bool isParentPainted;
+    static bool isChild1Painted;
+    static bool isChild2Painted;
+    static bool isChild3Painted;
```
vadz, line 487: I don't think it's really going to be a problem here, but it's still a bit weird to make these variables static, why not just pass their address to the handler? In fact, by doing this we could define the handler only once instead of doing it 4 times.

I.e. we could do (warning, untested code ahead):
```cpp
const auto setFlagOnPaint = [](wxWindow* win, bool* flag)
{
    win->Bind(wxEVT_PAINT, [=](wxPaintEvent&)
    {
        wxPaintDC dc(win);
        *flag = true;
    });
}; 

setFlagOnPaint(parent, isParentPainted);
setFlagOnPaint(child1, isChild1Painted);
...
```

Wouldn't this work?

vadz (commented):
This is definitely a simple but useful test to have, thanks!

I can merge it as is, but perhaps it could be improved a bit as suggested, please let me know if you see any problems with doing it like this.

Thanks!

tests/controls/windowtest.cpp
AliKet: Indeed, this works and is much more elegant and concise.
Thanks for the suggestion.

tests/controls/windowtest.cpp
AliKet: Removed now

vadz:
Perfect, thanks! Will merge soon.

PR #22923, opened 2022-10-27
Use wxID_HIGHEST instead of wxID_HIGHEST+1 in samples

Blake-Madden:


Blake-Madden:
This is in response to the recent re-alignment of wxID_LOWEST - wxID_HIGHEST.

vadz:
Thanks for the cleanup! I will merge it soon.

PR #22929, opened 2022-10-28
wxQt: wxAnyButton should override GetLabel() too

AliKet:
`wxAnyButton` already overrides `SetLabel()`, but not `GetLabel()` in which case an empty string will be returned if we try to call it because the base class version is trying to return the window title instead of the window text/label and this is obviously wrong.


vadz:

PR #22931, opened 2022-10-30
Always use vendor in CMake-built DLLs regardless of version

PBfordev:
The vendor string was ignored for the stable wxWidgets versions (e.g., 3.2). However, vendor should be always used when available, regardless of wxWidgets version. The only case when vendor should not be used is when creating official builds (currently not done with CMake).

This should be back-ported to 3.2, as that version is currently affected (but it will produce different DLL names than 3.2.0 and 3.2.1...).


vadz:
> This should be back-ported to 3.2, as that version is currently affected (but it will produce different DLL names than 3.2.0 and 3.2.1...).

Not for the official builds created for the releases though, right?

PR #22932, opened 2022-10-30
wxQt fixups

AliKet:
All upcoming fixes are detected by running the tests locally. and instead of creating a giant PR to include them all at once, I decided to create tiny PRs with one or two (or maybe three) commits on each to make the review process easier. TIA for any reviews!

### DIFF
--- a/include/wx/qt/private/winevent.h
+++ b/include/wx/qt/private/winevent.h
@@ -458,4 +458,26 @@ protected:
     }
 };
 
+// RAII wrapper for blockSignals(). It blocks signals in its constructor and in
+// the destructor it restores the state to what it was before the constructor ran.
+class wxQtEnsureSignalsBlocked
+{
+public:
+    // Use QObject instead of QWidget to avoid including <QWidget> from here.
+    wxQtEnsureSignalsBlocked(QObject *widget) :
+        m_widget(widget)
+    {
+        m_restore = m_widget->blockSignals(true);
+    }
+
+    ~wxQtEnsureSignalsBlocked()
+    {
+        m_widget->blockSignals(m_restore);
+    }
+
+private:
+    QObject* const m_widget;
+    bool m_restore;
+};
+
 #endif
--- a/include/wx/qt/private/winevent.h
+++ b/include/wx/qt/private/winevent.h
@@ -478,6 +478,8 @@ public:
 private:
     QObject* const m_widget;
     bool m_restore;
+
+    wxDECLARE_NO_COPY_CLASS(wxQtEnsureSignalsBlocked);
 };
 
 #endif
--- a/src/qt/checklst.cpp
+++ b/src/qt/checklst.cpp
@@ -81,7 +81,7 @@ bool wxCheckListBox::IsChecked(unsigned int n) const
 
 void wxCheckListBox::Check(unsigned int n, bool check )
 {
-    // Prevent the emission of wxEVT_CHECKLISTBOX event by temporarily block all
+    // Prevent the emission of wxEVT_CHECKLISTBOX event by temporarily blocking all
     // signals on m_qtListWidget object.
     wxQtEnsureSignalsBlocked blocker(m_qtListWidget);
     QListWidgetItem* item = m_qtListWidget->item(n);
--- a/src/qt/checklst.cpp
+++ b/src/qt/checklst.cpp
@@ -9,6 +9,7 @@
 #include "wx/wxprec.h"
 
 #include "wx/checklst.h"
+#include "wx/qt/private/winevent.h"
 
 #include <QtWidgets/QListWidgetItem>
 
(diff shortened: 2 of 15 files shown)


AliKet:
The first commit is a copy-and-past from wxGTK sources.

src/qt/checklst.cpp
```diff
@@ -80,8 +80,13 @@ bool wxCheckListBox::IsChecked(unsigned int n) const
 
 void wxCheckListBox::Check(unsigned int n, bool check )
 {
+    // Prevent the emission of wxEVT_CHECKLISTBOX event by temporarily block all
+    // signals on m_qtListWidget object. QSignalBlocker can be used instead when
+    // Qt 5.3 becomes the minimum supprted version.
```
vadz, line 85: FWIW we could easily have our own trivial RAII helper calling `blockSignals()` in ctor and unblocking them (if necessary) in dtor.

And use it not only here but in all the other places were we call `blockSignals()` manually currently.

src/qt/combobox.cpp
```diff
@@ -190,6 +190,18 @@ bool wxComboBox::IsReadOnly() const
     return HasFlag( wxCB_READONLY );
 }
 
+bool wxComboBox::IsEditable() const
+{
+    return !IsReadOnly() && !m_qtComboBox->lineEdit()->isReadOnly();
```
vadz, line 195: It's a bit suspicious that we check for it being null below, but not here. Shouldn't we?

vadz (commented):
Looks good, thanks, I can merge this already if you confirm that we don't need to check for `lineEdit() != nullptr`.

But if you prefer to add this `wxQtSignalBlocker`, please let me know.

src/qt/checklst.cpp
AliKet: I just found this:
https://github.com/wxWidgets/wxWidgets/blob/5002fb0351d3e04a70d15e50f787d1589bf29053/src/qt/treectrl.cpp#L1119-L1137

I'll rename it to `wxQtSignalBlocker` and put it in `qt/private/winevent.h`
is that acceptable ?

vadz: Yes, please. I was actually looking for this class because I thought it already existed, but didn't find it because I was looking in the headers and not in the sources.

TIA!

vadz: P.S. Maybe it's not even necessary to rename it.

src/qt/notebook.cpp
```diff
@@ -178,14 +178,12 @@ bool wxNotebook::DeleteAllPages()
 
     // Block signals to not receive selection changed updates
     // which are sent by Qt after the selected page was deleted.
-    m_qtTabWidget->blockSignals(true);
+    wxQtEnsureSignalsBlocked blocker(m_qtTabWidget);
 
     // Pages will be deleted one by one in the base class.
     // There's no need to explicitly clear() the Qt control.
     bool deleted = wxNotebookBase::DeleteAllPages();
```
vadz, line 185: Very minor, but this could be combined with the next line to just
```suggestion
    return wxNotebookBase::DeleteAllPages();
```
now.

include/wx/qt/private/winevent.h
```diff
@@ -458,4 +458,26 @@ class wxQtEventSignalHandler : public Widget, public wxQtSignalHandler< Handler
...
+
+    ~wxQtEnsureSignalsBlocked()
+    {
+        m_widget->blockSignals(m_restore);
+    }
+
+private:
+    QObject* const m_widget;
+    bool m_restore;
+};
```
vadz, line 481: I'd add this for clarity:
```suggestion
    wxDECLARE_NO_COPY_CLASS(wxQtEnsureSignalsBlocked);
};
```

src/qt/checklst.cpp
```diff
@@ -80,8 +81,11 @@ bool wxCheckListBox::IsChecked(unsigned int n) const
 
 void wxCheckListBox::Check(unsigned int n, bool check )
 {
+    // Prevent the emission of wxEVT_CHECKLISTBOX event by temporarily block all
```
vadz, line 84: Sorry, another nitpick:
```suggestion
    // Prevent the emission of wxEVT_CHECKLISTBOX event by temporarily blocking all
```

src/qt/scrolbar.cpp
```diff
@@ -99,11 +99,10 @@ void wxScrollBar::SetScrollbar(int position, int WXUNUSED(thumbSize),
     // Configure the scrollbar
     if (range != 0 )
     {
+        wxQtEnsureSignalsBlocked blocker(m_qtScrollBar);
         m_qtScrollBar->setRange( 0, range - pageSize );
         m_qtScrollBar->setPageStep( pageSize );
-        m_qtScrollBar->blockSignals(true);
         m_qtScrollBar->setValue( position );
-        m_qtScrollBar->blockSignals(false);
         m_qtScrollBar->show();
```
vadz, line 106: I have no idea if it matters, but now signals are also blocked for `set{Range,PageStep()}` and `show()` too whereas previously they were disabled after and re-enabled before calling these functions. Is this intentional? If you're not sure neither, I'd rather put the `blocker` creation in its own block containing just `setValue()` call.

src/qt/window.cpp
```diff
@@ -724,11 +743,10 @@ void wxWindowQt::SetScrollbar( int orientation, int pos, int thumbvisible, int r
     // scrollBar == nullptr and it is not a problem
     if ( scrollBar )
     {
+        wxQtEnsureSignalsBlocked blocker(scrollBar);
```
vadz, line 746: Same question as for scrollbar code above.

vadz (commented):
Thanks a lot, this is nice simplification!

If you think that extending the scope of blocking the signals is fine, we can merge this already, but if not, or if you're not sure, I'd rather restrict this scope to avoid changing the existing code behaviour, just because I don't know if it can be important or not.

vadz:

vadz:
FYI: I've rebased the commits to squash the last one, but without making any changes to the code.

PR #22934, opened 2022-10-30
Do not trigger non-CMake CI builds with CMake-only changes

PBfordev:


PBfordev:
Unfortunately, I have just noticed there is still a repeated typo in GitHub CI ymls. Wherever there is` 'build/cmake**'`, there should be `'build/cmake**'` (a missing slash before asterisks).

I can fix it in few hours.

vadz:
I must be blind but I don't see what the difference is, so I'll leave fixing this to you, it's not really urgent anyhow. TIA!

PBfordev:
> I must be blind but I don't see what the difference is

You are not blind, I am just being extremely sloppy and missed the slash again. Hopefully all fixed now.

vadz:
Thanks! I'll push this soon and will also apply it to 3.2 because it looks like it should be useful there too.

PR #22936, opened 2022-11-01
Update docs [docs] [low priority]

PBfordev:
There may be some issues still left to improve in the updated docs.

~~I left some code examples in the Archive Formats untouched, as I was not sure how to best improve them.~~
I updated the examples but the one using `ZipCatalog` produces an error when used with an empty zip file.

In the Exceptions programming guide, the section about exceptions in GUI tests could be extended, at least `tests/test.cpp` mentioned as an example of wxWidgets GUI test application setup.

### DIFF
--- a/docs/doxygen/overviews/archive.h
+++ b/docs/doxygen/overviews/archive.h
@@ -181,7 +181,7 @@ entries looked up by name can be opened using the
 wxArchiveInputStream::OpenEntry() method.
 
 @code
-using ZipEntryPtr = std::shared_ptr<wxZipEntry>;
+using ZipEntryPtr = std::unique_ptr<wxZipEntry>;
 using ZipCatalog = std::map<wxString, ZipEntryPtr>;
 ZipEntryPtr entry;
 ZipCatalog cat;
@@ -192,7 +192,7 @@ wxZipInputStream zip(in);
 
 // load the zip catalog
 while (entry.reset(zip.GetNextEntry()), entry.get() != nullptr)
-    cat[entry->GetInternalName()] = entry;
+    cat[entry->GetInternalName()] = std::move(entry);
 
 // open an entry by name
 auto it = cat.find(wxZipEntry::GetInternalName(localName));
--- a/docs/doxygen/overviews/archive.h
+++ b/docs/doxygen/overviews/archive.h
@@ -14,7 +14,7 @@
 The archive classes handle archive formats such as zip.
 Currently zip and tar support is bundled with wxWidgets and
 when external liblzma library is available (see @ref page_build_liblzma),
-XZ format using LZMA2 algorithm is supported.
+XZ format using LZMA2 algorithm is supported as well.
 
 For each archive type, there are the following classes (using zip here as an
 example):
--- a/docs/doxygen/overviews/archive.h
+++ b/docs/doxygen/overviews/archive.h
@@ -67,7 +67,7 @@ ownership).
 Reading from the input stream then returns the entry's data. Eof() becomes
 @true after an attempt has been made to read past the end of the entry's data.
 
-When there are no more entries, GetNextEntry() returns @nullptr and sets Eof().
+When there are no more entries, GetNextEntry() returns @NULL and sets Eof().
 
 @code
 std::unique_ptr<wxZipEntry> entry;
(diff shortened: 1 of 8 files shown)


docs/doxygen/overviews/archive.h
```diff
@@ -11,8 +11,10 @@
 
 @tableofcontents
 
-The archive classes handle archive formats such as zip, tar, rar and cab.
-Currently wxZip, wxTar and wxZlib classes are included.
+The archive classes handle archive formats such as zip.
+Currently zip and tar support is bundled with wxWidgets and
+when external liblzma library is available (see @ref page_build_liblzma),
+XZ format using LZMA2 algorithm is supported.
```
vadz, line 17: Maybe
```suggestion
XZ format using LZMA2 algorithm is supported as well.
```
?

docs/doxygen/overviews/archive.h
```diff
@@ -65,12 +67,12 @@ ownership).
 Reading from the input stream then returns the entry's data. Eof() becomes
 @true after an attempt has been made to read past the end of the entry's data.
 
-When there are no more entries, GetNextEntry() returns @NULL and sets Eof().
+When there are no more entries, GetNextEntry() returns @nullptr and sets Eof().
```
vadz, line 70: I've changed `@NULL` alias in `docs/doxygen/Doxyfile` to expand to `nullptr` too to avoid having to change all of its existing occurrences. I don't know if it's too confusing to keep using it, but even if not, we should replace all of its occurrences with `@nullptr` at once, otherwise we're going to end up with an unholy mix of the two which would be even worse.

IOW I'd like to
```suggestion
When there are no more entries, GetNextEntry() returns @NULL and sets Eof().
```
at least for now.

docs/doxygen/overviews/archive.h
```diff
@@ -179,27 +181,22 @@ entries looked up by name can be opened using the
 wxArchiveInputStream::OpenEntry() method.
 
 @code
-WX_DECLARE_STRING_HASH_MAP(wxZipEntry*, ZipCatalog);
-ZipCatalog::iterator it;
-wxZipEntry *entry;
+using ZipEntryPtr = std::shared_ptr<wxZipEntry>;
```
vadz, line 184: Do we really need to use `shared_`, rather than `unique_`, pointer here? Not a problem if we do, I just don't see why should we.

vadz (commented):
Thanks a lot for updating these docs!

I'm fine with merging it as is, but please let me know if you agree with the proposed changes (and then you can either apply them yourself or I can do it when merging).

docs/doxygen/overviews/archive.h
PBfordev: Fixed.

vadz:
Looks great now, thanks!

I'm not sure if you plan to rebase to get rid of fixups yourself or if I should do it?

Also, a strange question: how do you create the "!fixup" commits? Normally it's done by using `git commit --fixup`, but this should use a subject of an existing commit after the "fixup" part, which doesn't seem to be the case here... Are you creating them manually, by chance?

PBfordev:
> I'm not sure if you plan to rebase to get rid of fixups yourself or if I should do it?

Please do it yourself, I can't speak GIT.

> Also, a strange question: how do you create the "!fixup" commits? Normally it's done by using `git commit --fixup`, but this should use a subject of an existing commit after the "fixup" part, which doesn't seem to be the case here... Are you creating them manually, by chance?

I noticed that sometimes "!fixup" was used in the commit titles for commits like these but I had no idea it was added automatically. So of course, I was creating them manually, see my reply to your first comment above.

vadz:
Just FYI, when you use `git commit --fixup some-commit` you can then use `git rebase --autosquash -i` to let Git rearrange the commits automatically. In fact, you can do `git config rebase.autoSquash true` to make this behaviour the default (as I do) and then you don't need to specify `--autosquash` explicitly.

For now I've squashed the commits manually and will merge this soon, thanks again!

PBfordev:
> Just FYI, when you use `git commit --fixup some-commit`...

Thanks for tip but I am using TortoiseGIT (a GUI-based GIT client) which unfortunately does not have this. 

Last time I thought I rebased my PR I ended up pulling a humongous diff between the master and my branch into the PR, so I'd rather avoid anything unnecessary, sorry to burden you with it.

PR #22939, opened 2022-11-02
Document pen styles, joins, and stipple styles

Blake-Madden:


interface/wx/pen.h
```diff
@@ -37,10 +37,12 @@ enum wxPenStyle
         /**< No pen is used. */
 
     wxPENSTYLE_STIPPLE_MASK_OPAQUE,
-        /**< @todo WHAT's this? */
+        /**< Uses a bitmap as a stipple; the mask is used for blitting
+             monochrome using text foreground and background colors. */
```
vadz, line 41: Is there something missing after "monochrome"? I have trouble parsing this sentence in the current state.

vadz (commented):
Nice, thanks for improving this!

I'm just not completely sure about the `wxPENSTYLE_STIPPLE_MASK_OPAQUE` comment, could you please check it?

interface/wx/pen.h
Blake-Madden: I copied this description from the sibling `wxBRUSHSTYLE_STIPPLE_MASK_OPAQUE` enum value.  I assumed "monochrome" was used as an adverb here and I couldn't wordsmith it any differently...so I just left it as-is.

vadz: The existing description doesn't make any sense to me neither :-(

I also don't think this is actually supported under any platform, so perhaps it would be better to just remove its documentation completely.

Blake-Madden: Should I remove it entirely from the interface file then?

vadz: I think it would be best, unless I'm missing something -- but a quick grep doesn't seem to find support for this in any port. But please double check it.

Issue #4775, opened 2008-04-05
wxNotebook & wxBK_HITTEST_* broken (ignores layout)


wxtrac:
*Issue migrated from trac ticket # 4775*

**component:** wxMSW | **priority:** normal

#### 2008-04-05 22:58:40: arkay created the issue

___
The returned flags should indicate the position where the hit on the control appeared.

This only works if the control uses standard layout (wxNB_TOP and probably wxNB_BOTTOM, didn't test).

If wxNB_LEFT is used, wxBK_HITTEST_ONICON is _always_ returned regardless of where we click even though the image is displayed at the _bottom_ of the tab.

PR #22941, opened 2022-11-03
Add GetTabRect() and GetTabOrientation() to wxNotebook.

skeetor:
Add GetTabOrientation() to return the orientation of the tabs on a wxNotebook.
Add GetTabRect() to return the window coordinates of a tab from a wxNotebook.

### DIFF
--- a/include/wx/msw/notebook.h
+++ b/include/wx/msw/notebook.h
@@ -93,6 +93,9 @@ public:
     // style.
   void SetTabSize(const wxSize& sz) override;
 
+  // Return the position and size of the tab for the given page
+  wxRect GetTabRect(size_t page) const override;
+
     // hit test
   virtual int HitTest(const wxPoint& pt, long *flags = nullptr) const override;
 
--- a/include/wx/notebook.h
+++ b/include/wx/notebook.h
@@ -139,6 +139,13 @@ public:
     // new is wxNOT_FOUND)
     void SendPageChangedEvent(int nPageOld, int nPageNew = wxNOT_FOUND);
 
+    // return wxTOP/wxBOTTOM/wxRIGHT/wxLEFT
+    wxDirection GetTabOrientation() const;
+
+    // return the rectangle in window coordinates of the page selector.
+    // For example, for a wxNotebook this would be its tab.
+    virtual wxRect GetTabRect(size_t page) const;
+
 #if wxUSE_EXTENDED_RTTI
     // XTI accessors
     virtual void AddPageInfo( wxNotebookPageInfo* info );
--- a/interface/wx/notebook.h
+++ b/interface/wx/notebook.h
@@ -191,6 +191,26 @@ public:
     */
     virtual void SetPadding(const wxSize& padding);
 
+    /**
+        This is a convenience function mapping wxBK_TOP etc styles to one of the
+        wxDirection enum elements.
+
+        @since 3.3.0
+    */
+    wxDirection GetTabOrientation() const;
+
+    /**
+        Return the rectangle of the given page tab in window coordinates.
+        This function will always return the rectangle for the specified tab, even
+        if the tab is currently not visible. If the page is invalid the returned
+        rectangle is empty.
+
+        @note Currently only available in Univ and MSW
+
+        @since 3.3.0
+    */
+    virtual wxRect GetTabRect(size_t page) const;
+
     // implementations of pure virtuals
     virtual int GetPageImage(size_t nPage) const;
     virtual bool SetPageImage(size_t page, int image);
@@ -201,6 +221,5 @@ public:
     virtual int ChangeSelection(size_t page);
     virtual bool InsertPage(size_t index, wxWindow * page, const wxString & text,
                             bool select = false, int imageId = NO_IMAGE);
-
 };
 
--- a/src/common/nbkbase.cpp
+++ b/src/common/nbkbase.cpp
@@ -33,6 +33,8 @@
 
 extern WXDLLEXPORT_DATA(const char) wxNotebookNameStr[] = "notebook";
 
+#define IS_VALID_PAGE(nPage) ((nPage) < GetPageCount())
+
 wxDEFINE_EVENT( wxEVT_NOTEBOOK_PAGE_CHANGED, wxBookCtrlEvent );
 wxDEFINE_EVENT( wxEVT_NOTEBOOK_PAGE_CHANGING, wxBookCtrlEvent );
 
@@ -195,6 +197,29 @@ void wxNotebookBase::SendPageChangedEvent(int nPageOld, int nPageNew)
     event.SetOldSelection(nPageOld);
     event.SetEventObject(this);
     GetEventHandler()->ProcessEvent(event);
+}
+
+wxDirection wxNotebookBase::GetTabOrientation() const
+{
+    long style = GetWindowStyle();
+    if (style & wxBK_BOTTOM)
+        return wxBOTTOM;
+    else if (style & wxBK_RIGHT)
+        return wxRIGHT;
+    else if (style & wxBK_LEFT)
+        return wxLEFT;
+
+    // wxBK_TOP == 0 so we don't have to test for it
+    return wxTOP;
+}
+
+wxRect wxNotebookBase::GetTabRect(size_t page) const
+{
+    wxRect r;
+    wxCHECK_MSG(IS_VALID_PAGE(page), r, wxT("invalid notebook page"));
+    wxFAIL_MSG("Not implemented");
+
+    return r;
 }
 
 #endif // wxUSE_NOTEBOOK
--- a/src/msw/notebook.cpp
+++ b/src/msw/notebook.cpp
@@ -499,6 +499,21 @@ void wxNotebook::SetTabSize(const wxSize& sz)
     ::SendMessage(GetHwnd(), TCM_SETITEMSIZE, 0, MAKELPARAM(sz.x, sz.y));
 }
 
+wxRect wxNotebook::GetTabRect(size_t page) const
+{
+    wxRect r;
+    wxCHECK_MSG(IS_VALID_PAGE(page), r, wxT("invalid notebook page"));
+
+    if (GetPageCount() > 0)
+    {
+        RECT rect;
+        if (TabCtrl_GetItemRect(GetHwnd(), page, &rect))
+            r = wxRectFromRECT(rect);
+    }
+
+    return r;
+}
+
 wxSize wxNotebook::CalcSizeFromPage(const wxSize& sizePage) const
 {
     // we can't use TabCtrl_AdjustRect here because it only works for wxNB_TOP
--- a/src/univ/notebook.cpp
+++ b/src/univ/notebook.cpp
@@ -659,21 +659,7 @@ bool wxNotebook::IsVertical() const
     return dir == wxLEFT || dir == wxRIGHT;
 }
 
-wxDirection wxNotebook::GetTabOrientation() const
-{
-    long style = GetWindowStyle();
-    if ( style & wxBK_BOTTOM )
-        return wxBOTTOM;
-    else if ( style & wxBK_RIGHT )
-        return wxRIGHT;
-    else if ( style & wxBK_LEFT )
-        return wxLEFT;
-
-    // wxBK_TOP == 0 so we don't have to test for it
-    return wxTOP;
-}
-
-wxRect wxNotebook::GetTabRect(int page) const
+wxRect wxNotebook::GetTabRect(size_t page) const
 {
     wxRect rect;
     wxCHECK_MSG( IS_VALID_PAGE(page), rect, wxT("invalid notebook page") );
@@ -688,7 +674,7 @@ wxRect wxNotebook::GetTabRect(int page) const
     else
     {
         widthBefore = 0;
-        for ( int n = 0; n < page; n++ )
+        for (size_t n = 0; n < page; n++)
         {
             widthBefore += m_widths[n];
         }
--- a/include/wx/univ/notebook.h
+++ b/include/wx/univ/notebook.h
@@ -93,9 +93,6 @@ public:
     // return true if all tabs have the same width
     bool FixedSizeTabs() const { return HasFlag(wxNB_FIXEDWIDTH); }
 
-    // return wxTOP/wxBOTTOM/wxRIGHT/wxLEFT
-    wxDirection GetTabOrientation() const;
-
     // return true if the notebook has tabs at the sidesand not at the top (or
     // bottom) as usual
     bool IsVertical() const;
@@ -120,6 +117,9 @@ public:
 
     // refresh the currently selected tab
     void RefreshCurrent();
+
+    // get the tab rect
+    wxRect GetTabRect(size_t page) const override;
 
 protected:
     virtual wxNotebookPage *DoRemovePage(size_t nPage) override;
@@ -163,9 +163,6 @@ protected:
 
     // refresh all tabs
     void RefreshAllTabs();
-
-    // get the tab rect (inefficient, don't use this in a loop)
-    wxRect GetTabRect(int page) const;
 
     // get the rectangle containing all tabs
     wxRect GetAllTabsRect() const;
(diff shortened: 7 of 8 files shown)


include/wx/msw/notebook.h
```diff
@@ -93,6 +93,9 @@ class WXDLLIMPEXP_CORE wxNotebook : public wxNotebookBase
     // style.
   void SetTabSize(const wxSize& sz) override;
 
+  // Return the position and size of he tab for the given page
```
PBfordev, line 96:
```suggestion
  // return the position and size of the tab for the given page
```

include/wx/notebook.h
```diff
@@ -139,6 +139,13 @@ class WXDLLIMPEXP_CORE wxNotebookBase : public wxBookCtrlBase
     // new is wxNOT_FOUND)
     void SendPageChangedEvent(int nPageOld, int nPageNew = wxNOT_FOUND);
 
+    // return wxTOP/wxBOTTOM/wxRIGHT/wxLEFT
+    wxDirection GetTabOrientation() const;
+
+    // return the rectangle in windowcoordinates of the page selector.
```
PBfordev, line 145:
```suggestion
    // return the rectangle in window coordinates of the page selector
```

include/wx/notebook.h
```diff
@@ -139,6 +139,13 @@ class WXDLLIMPEXP_CORE wxNotebookBase : public wxBookCtrlBase
...
+    // i.E. for a wxNotebook this would be it's tab.
```
PBfordev, line 146:
```suggestion
    // e.g. for a wxNotebook this would be its tab
```

include/wx/notebook.h
```diff
@@ -139,6 +139,13 @@ class WXDLLIMPEXP_CORE wxNotebookBase : public wxBookCtrlBase
...
+    virtual wxRect GetTabRect(size_t n) const;
```
PBfordev, line 147:
```suggestion
    virtual wxRect GetTabRect(size_t page) const;
```

interface/wx/notebook.h
```diff
@@ -191,6 +191,19 @@ class wxNotebook : public wxBookCtrlBase
     */
     virtual void SetPadding(const wxSize& padding);
 
+    /**
+        Returns wxTOP / wxBOTTOM / wxRIGHT / wxLEFT
+    */
+    wxDirection GetTabOrientation() const;
+
+    /**
+        Return the rectangle in windowcoordinates of the page selector.
```
PBfordev, line 200:
```suggestion
        Returns the rectangle in window coordinates of the page selector.
```

interface/wx/notebook.h
```diff
@@ -191,6 +191,19 @@ class wxNotebook : public wxBookCtrlBase
...
+        i.E. for a wxNotebook this would be it's tab.
```
PBfordev, line 201:
```suggestion
        For example, for a wxNotebook this would be its tab.
```

interface/wx/notebook.h
```diff
@@ -191,6 +191,19 @@ class wxNotebook : public wxBookCtrlBase
...
+
+        @note Currently only available in Univ and MSW
+    */
+    virtual wxRect GetTabRect(size_t n) const;
```
PBfordev, line 205:
```suggestion
    virtual wxRect GetTabRect(size_t page) const;
```

src/common/nbkbase.cpp
```diff
@@ -197,4 +197,25 @@ void wxNotebookBase::SendPageChangedEvent(int nPageOld, int nPageNew)
...
+    else if (style & wxBK_LEFT)
+        return wxLEFT;
+
+    // wxBK_TOP == 0 so we don't have to test for it
+    return wxTOP;
+}
+
+wxRect wxNotebookBase::GetTabRect(size_t WXUNUSED(n)) const
+{
+    wxFAIL_MSG(wxT("wxNotebook::GetTabRect not implemented"));
```
PBfordev, line 216:
```suggestion
    wxFAIL_MSG("not implemented");
```

src/msw/notebook.cpp
```diff
@@ -499,6 +499,25 @@ void wxNotebook::SetTabSize(const wxSize& sz)
...
+wxRect wxNotebook::GetTabRect(size_t n) const
+{
+    wxRect r;
+
+    if (GetPageCount() > 0)
+    {
+        RECT rect;
+        if (TabCtrl_GetItemRect(GetHwnd(), n, &rect))
+        {
+            r.x = rect.left;
```
PBfordev, line 511: I would use `r = wxRectFromRECT(rect);`  instead of all these four lines.

src/univ/notebook.cpp
```diff
@@ -688,7 +674,7 @@ wxRect wxNotebook::GetTabRect(int page) const
     else
     {
         widthBefore = 0;
-        for ( int n = 0; n < page; n++ )
+        for (size_t n = 0; n < page; n++)
```
PBfordev, line 677:
```suggestion
        for ( size_t n = 0; n < page; n++ )
```

PBfordev (commented):
Code specific comments aside...

Not sure if the maintainers accept such methods when implemented only for one of the three major ports. I am MSW-only myself, so I have no idea if it is possible to do the same for macOS and GTK.

Just curious, does it work correctly when the tabs are not on the top (can be easily tested with the notebook sample)? IIRC, there were some issues on MSW with non-top tab placement, you can see in the sample that the tabs look different in other positions than the top.

src/univ/notebook.cpp
skeetor: Already fixed in latest commit

include/wx/msw/notebook.h
skeetor: Fixed

include/wx/notebook.h
skeetor: Fixed

include/wx/notebook.h
skeetor: Fixed

include/wx/notebook.h
skeetor: Fixed

interface/wx/notebook.h
skeetor: Fixed

interface/wx/notebook.h
skeetor: Fixed

interface/wx/notebook.h
skeetor: Fixed

src/common/nbkbase.cpp
skeetor: Fixed

src/msw/notebook.cpp
skeetor: Fixed. Didn't know about this funtion. :)

skeetor:
This code definitely works on MSW and I tested it with all orientations of the tabs. For the univ build, the functions already existed and I didn't change them, so I assume they work or at least will behave as they always had.

I would like to add this to QT and GTK build later on as well, but currently I work on my main implementation of my module and I don't want to get diverted to much by it by testing several other builds as well. However, since these functions were not there before anyway, they don't brake anything as of now.

src/msw/notebook.cpp
```diff
@@ -499,6 +499,20 @@ void wxNotebook::SetTabSize(const wxSize& sz)
...
 
+wxRect wxNotebook::GetTabRect(size_t n) const
+{
+    wxRect r;
+
+    if (GetPageCount() > 0)
+    {
+        RECT rect;
+        if (TabCtrl_GetItemRect(GetHwnd(), n, &rect))
+            wxRectFromRECT(rect);
```
PBfordev, line 510:
```suggestion
            r = wxRectFromRECT(rect);
```

BTW, I meant to change the method argument name from `n` to `page` everywhere it was used (similarly for typos in the comments).

skeetor: Fixed

oneeyeman1:
@skeetor ,
There was a bug in Windows about HitTest not working correctly for non-top tab position. MS confirmed it.

I presume this code works as it should.

Maybe we can implement HitTest somehow with this change for non-top tab position?

skeetor:
I noticed that about HitTest because I tried using that first. Since it didn't work as I expected it, I wrote some code to detect klicks in the off region of the tabs, why I needed this functions. :)
I wrote some code to get the full tab rectangle, as well as getting the rectangle of the off region. So should I add this as well, so it might be used by HitTest?

oneeyeman1:
@skeetor ,
There is a long-withstanding issue/ticket about it.
I told MS about it and they confirmed it.
If you can make HitTest() on MSW work as expected - it would be great.
However, it probably should be another PR and if you can find this issue and reference it it will be even better.

skeetor:
OK! I would also prefer to have it in another PR anyway so this would be fine for me.

skeetor:
The only reference I could find is this one: https://github.com/wxWidgets/wxWidgets/issues/4775

However, with my (upcoming) fix this will work for all orientations, as I needed this to work reliably for my docking module.

interface/wx/notebook.h
```diff
@@ -191,6 +191,19 @@ class wxNotebook : public wxBookCtrlBase
     */
     virtual void SetPadding(const wxSize& padding);
 
+    /**
+        Returns wxTOP / wxBOTTOM / wxRIGHT / wxLEFT
```
vadz, line 195: I'd like to expand the description a bit and we also should indicate that it's only available in 3.3 and later:

```suggestion
        Returns one of wxTOP, wxBOTTOM, wxRIGHT or wxLEFT.
        
        This is a convenience function mapping wxBK_TOP etc styles to one of the
        wxDirection enum elements.
        
        @since 3.3.0
```

interface/wx/notebook.h
```diff
@@ -191,6 +191,19 @@ class wxNotebook : public wxBookCtrlBase
...
+    */
+    wxDirection GetTabOrientation() const;
+
+    /**
+        Return the rectangle in window coordinates of the page selector.
+        For example, for a wxNotebook this would be its tab.
+
+        @note Currently only available in Univ and MSW
+    */
```
vadz, line 204:
```suggestion
        @since 3.3.0
    */
```

interface/wx/notebook.h
```diff
@@ -191,6 +191,19 @@ class wxNotebook : public wxBookCtrlBase
...
     virtual void SetPadding(const wxSize& padding);
 
+    /**
+        Returns wxTOP / wxBOTTOM / wxRIGHT / wxLEFT
+    */
+    wxDirection GetTabOrientation() const;
+
+    /**
+        Return the rectangle in window coordinates of the page selector.
+        For example, for a wxNotebook this would be its tab.
```
vadz, line 201: I'm not sure it makes sense to say "for a wxNotebook" in wxNotebook documentation. Was this supposed to be a `wxBookCtrlBase` function initially?

include/wx/notebook.h
```diff
@@ -139,6 +139,13 @@ class WXDLLIMPEXP_CORE wxNotebookBase : public wxBookCtrlBase
     // new is wxNOT_FOUND)
     void SendPageChangedEvent(int nPageOld, int nPageNew = wxNOT_FOUND);
 
+    // return wxTOP/wxBOTTOM/wxRIGHT/wxLEFT
+    wxDirection GetTabOrientation() const;
+
+    // return the rectangle in window coordinates of the page selector.
+    // For example, for a wxNotebook this would be its tab.
```
vadz, line 146: As also mentioned in the interface file comment, saying "for example for a wxNotebook" in wxNotebook documentation doesn't seem to make much sense, so I'd just write
```suggestion
    // return the rectangle of the given page tab in window coordinates
```

src/common/nbkbase.cpp
```diff
@@ -197,4 +197,25 @@ void wxNotebookBase::SendPageChangedEvent(int nPageOld, int nPageNew)
...
+    else if (style & wxBK_LEFT)
+        return wxLEFT;
+
+    // wxBK_TOP == 0 so we don't have to test for it
+    return wxTOP;
+}
+
+wxRect wxNotebookBase::GetTabRect(size_t WXUNUSED(n)) const
+{
+    wxFAIL_MSG(wxT("Not implemented"));
```
vadz, line 216: Let's not use `wxT()` in the new code:
```suggestion
    wxFAIL_MSG("Not implemented");
```

src/univ/notebook.cpp
vadz: This is pretty minor, but it doesn't seem to have been fixed.

vadz (commented):
Thanks, I'm not really against having this function even if it's MSW-only if it can be implemented for the other ports later, but I have some questions about it that should be answered in the documentation and, ideally, tests:

1. What happens when the tab is scrolled away, i.e. not currently visible at all?
2. What happens if an invalid index (too big) is passed?
3. Do we really want to assert in the other ports or maybe we should just return an empty rectangle silently? I think the latter would be more useful as otherwise I don't really know how would I use this function in portable code.

Having a test for this function in `tests/controls/notebooktest.cpp` would be really, really nice -- it could be used to clarify the answers to the questions above and also make it simpler to check that the function is implemented for the other platforms correctly later.

TIA!

skeetor:
>     1. What happens when the tab is scrolled away, i.e. not currently visible at all?

The official MSW function TabCtrl_GetItemRect() reports the position where the tab would be. There is no indicator that the tab is not visible. So the application has to check this itself if it loops through all tabs.

>     2. What happens if an invalid index (too big) is passed?

I added the checks.

>     3. Do we really want to assert in the other ports or maybe we should just return an empty rectangle silently? I think the latter would be more useful as otherwise I don't really know how would I use this function in portable code.

Personally I prefer to get a big red warning from the library layer instead of silently giving me a wrong answer. If an application developer wants to use this function and it is not implemented on his platform, then he should be aware that it is not working, instead of having to debug it why it doesn't work only to find out that it's not implemented.
I was considering to implement it later on for wxGTK myself, when I get to port my app to linux, but this is still far in the future, so maybe somebody may beat me to it. :) As for other platforms like OSX, somebody, who is interested enough would have to do it.

> Having a test for this function in `tests/controls/notebooktest.cpp` would be really, really nice -- it could be used to clarify the answers to the questions above and also make it simpler to check that the function is implemented for the other platforms correctly later.

OK. I will take a look at it how to test it. I added all your other suggestions as well.

vadz:
> > 1. What happens when the tab is scrolled away, i.e. not currently visible at all?
> 
> The official MSW function TabCtrl_GetItemRect() reports the position where the tab would be. There is no indicator that the tab is not visible. So the application has to check this itself if it loops through all tabs.

This is fine, but please add this to the documentation, as it's not really obvious. I.e. just say that the position of the tab can be beyond the window boundary or something like this.

> > 2. What happens if an invalid index (too big) is passed?
> 
> I added the checks.

Thanks!

> > 3. Do we really want to assert in the other ports or maybe we should just return an empty rectangle silently? I think the latter would be more useful as otherwise I don't really know how would I use this function in portable code.
> 
> Personally I prefer to get a big red warning from the library layer instead of silently giving me a wrong answer. If an application developer wants to use this function and it is not implemented on his platform, then he should be aware that it is not working, instead of having to debug it why it doesn't work only to find out that it's not implemented.

This is a reasonable position, although I'm still not sure if it's the best thing to do in this particular case. But we can start by doing this as we can always remove the assert later if it's really problematic. However it should, again, be clearly documented, i.e. the docs should say that this function is only implemented in wxMSW and wxUniv and asserts if it's called under another platform.

skeetor:
> Having a test for this function in `tests/controls/notebooktest.cpp` would be really, really nice -- it could be used to clarify the answers to the questions above and also make it simpler to check that the function is implemented for the other platforms correctly later.

I'm currently doing the test, but that got me thinking. When I add a testcase for GetTabRect() wont the builds for GTK and OSX fail because of the assert, until these functions are properly implemented?

vadz:
> > Having a test for this function in `tests/controls/notebooktest.cpp` would be really, really nice -- it could be used to clarify the answers to the questions above and also make it simpler to check that the function is implemented for the other platforms correctly later.
> 
> I'm currently doing the test, but that got me thinking. When I add a testcase for GetTabRect() wont the builds for GTK and OSX fail because of the assert, until these functions are properly implemented?

Yes, you need to use `#if defined(__WXMSW__) || defined(__WXUNIVERSAL__)` around these tests for now.

vadz:
Thanks, applied with some minor/cosmetic changes.

PR #22942, opened 2022-11-03
wxQt: wxEVT_TEXT_ENTER processing fixup

AliKet:
Can be tested with this patch:
```
index 501caf9096..759b60f3cc 100644
@@ -186,18 +186,80 @@ void MyFrame::OnQuit(wxCommandEvent& WXUNUSED(event))
     Close(true);
 }
 
+namespace
+{
+
+class TestDialog : public wxDialog
+{
+    enum
+    {
+        Control1_ID = wxID_HIGHEST,
+        Control2_ID,
+        Control3_ID
+    };
+
+public:
+    explicit TestDialog(wxWindow* parent)
+        : wxDialog(parent, wxID_ANY, "Test dialog")
+    {
+        const wxString choices[] = { "foo", "bar", "baz" };
+
+        wxControl* const control1 =
+            new wxComboBox(this, Control1_ID, wxString(),
+                           wxDefaultPosition, wxDefaultSize,
+                           WXSIZEOF(choices), choices);
+        control1->SetToolTip("ProcessEnter_No");
+
+        wxControl* const control2 =
+            new wxComboBox(this, Control2_ID, wxString(),
+                           wxDefaultPosition, wxDefaultSize,
+                           WXSIZEOF(choices), choices, wxTE_PROCESS_ENTER);
+        control2->SetToolTip("ProcessEnter_ButSkip");
+
+        wxControl* const control3 =
+            new wxTextCtrl(this, Control3_ID, wxString(),
+                           wxDefaultPosition, wxDefaultSize, wxTE_PROCESS_ENTER);
+        control3->SetToolTip("ProcessEnter_WithoutSkipping");
+
+        wxSizer* const sizer = new wxBoxSizer(wxVERTICAL);
+        sizer->Add(control1, wxSizerFlags().Expand().Border());
+        sizer->Add(control2, wxSizerFlags().Expand().Border());
+        sizer->Add(control3, wxSizerFlags().Expand().Border());
+        sizer->Add(CreateStdDialogButtonSizer(wxOK));
+        SetSizerAndFit(sizer);
+    }
+
+private:
+    void OnTextEnter(wxCommandEvent& e)
+    {
+        const int id = e.GetId();
+        switch ( id )
+        {
+            case Control1_ID: // ProcessEnter_No
+                wxFAIL_MSG("Shouldn't be getting wxEVT_TEXT_ENTER at all");
+                break;
+
+            case Control2_ID: // ProcessEnter_ButSkip
+                wxLogStatus("ProcessEnter_ButSkip");
+                e.Skip();
+                break;
+
+            case Control3_ID: // ProcessEnter_WithoutSkipping
+                wxLogStatus("ProcessEnter_WithoutSkipping");
+                break;
+        }
+    }
+
+    wxDECLARE_EVENT_TABLE();
+};
+
+wxBEGIN_EVENT_TABLE(TestDialog, wxDialog)
+    EVT_TEXT_ENTER(wxID_ANY, TestDialog::OnTextEnter)
+wxEND_EVENT_TABLE()
+}
+
 {
-                 (
-                    "\n"
-                 ),
-                 this);
+    TestDialog dlg(this);
+    dlg.ShowModal();
 }
```

### DIFF
--- a/include/wx/qt/private/winevent.h
+++ b/include/wx/qt/private/winevent.h
@@ -82,6 +82,11 @@ public:
             return wxQtSignalHandler< Handler >::GetHandler();
     }
 
+    // Hack used by wxTextEntry derived classes (by specializing this function)
+    // to toggle the dialog's default button off and on on focus events.
+    // see src/qt/textentry.cpp for the reason.
+    void ToggleDefaultButtonOnFocusEvent() { }
+
 protected:
     /* Not implemented here: wxHelpEvent, wxIdleEvent wxJoystickEvent,
      * wxMouseCaptureLostEvent, wxMouseCaptureChangedEvent,
@@ -144,6 +149,8 @@ protected:
         if ( !this->GetHandler() )
             return;
 
+        ToggleDefaultButtonOnFocusEvent();
+
         if ( !this->GetHandler()->QtHandleFocusEvent(this, event) )
             Widget::focusInEvent(event);
         else
@@ -155,6 +162,8 @@ protected:
     {
         if ( !this->GetHandler() )
             return;
+
+        ToggleDefaultButtonOnFocusEvent();
 
         if ( !this->GetHandler()->QtHandleFocusEvent(this, event) )
             Widget::focusOutEvent(event);
@@ -457,24 +466,5 @@ protected:
         }
     }
 };
-
-// Used by wxTextEntry derived classes to toggle the dialog's default button
-// off and on. see src/qt/textentry.cpp for the reason.
-#define wxHANDLE_DIALOG_DEFAULT_BUTTON(Widget)              \
-    virtual void focusInEvent(QFocusEvent *e) override      \
-    {                                                       \
-        if ( GetHandler()->HasFlag(wxTE_PROCESS_ENTER) )    \
-            GetHandler()->ToggleDefaultButton();            \
-                                                            \
-        Widget::focusInEvent(e);                            \
-    }                                                       \
-                                                            \
-    virtual void focusOutEvent(QFocusEvent *e) override     \
-    {                                                       \
-        if ( GetHandler()->HasFlag(wxTE_PROCESS_ENTER) )    \
-            GetHandler()->ToggleDefaultButton();            \
-                                                            \
-        Widget::focusOutEvent(e);                           \
-    }
 
 #endif
--- a/include/wx/qt/private/winevent.h
+++ b/include/wx/qt/private/winevent.h
@@ -458,4 +458,23 @@ protected:
     }
 };
 
+// Used by wxTextEntry derived classes to toggle the dialog's default button
+// off and on. see src/qt/textentry.cpp for the reason.
+#define wxHANDLE_DIALOG_DEFAULT_BUTTON(Widget)              \
+    virtual void focusInEvent(QFocusEvent *e) override      \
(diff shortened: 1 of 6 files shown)


include/wx/qt/textentry.h
```diff
@@ -36,6 +36,10 @@ class WXDLLIMPEXP_CORE wxTextEntry : public wxTextEntryBase
     virtual bool IsEditable() const override;
     virtual void SetEditable(bool editable) override;
 
+    // For internal use only, see implementation for explanation.
+    // Only controls with wxTE_PROCESS_ENTER flag use this function.
+    void ToggleDefaultButton();
```
vadz, line 41: Let's use port-specific prefix for non-portable functions:
```suggestion
    void QtToggleDefaultButton();
```

include/wx/qt/private/winevent.h
```diff
@@ -458,4 +458,23 @@ class wxQtEventSignalHandler : public Widget, public wxQtSignalHandler< Handler
     }
 };
 
+// Used by wxTextEntry derived classes to toggle the dialog's default button
+// off and on. see src/qt/textentry.cpp for the reason.
+#define wxHANDLE_DIALOG_DEFAULT_BUTTON(Widget)              \
```
vadz, line 463: Using macro for this is rather ugly, could we use just use a function instead, i.e. `wxQtToggleDefaultButtonOnFocusEvent()`, and override the methods explicitly where this needs to be done? This would be much more clear and IMO the code reuse of the macro is not worth it in this case.

vadz (commented):
Thanks for fixing this! It's a pity that such a hack is needed (how do native Qt applications deal with it, is there really no Qt dialogs with a default button and multiline text?), but let's do it.

I'd prefer to do it without the macro, however.

TIA!

AliKet:
> Thanks for fixing this! It's a pity that such a hack is needed (how do native Qt applications deal with it, is there really no Qt dialogs with a default button and multiline text?), ...

AFAICT the multiline text works correctly. only the singleline text which doesn't work as documented, i.e. pressing Enter key is either processed internally by the control or used to activate the default button of the dialog, if any.
In `wxQt` the default button is always activated even though `wxEVT_TEXT_ENTER` handler doesn't **Skip()** the event which is not what the other ports currently do.

include/wx/qt/private/winevent.h
```diff
@@ -458,4 +467,23 @@ class wxQtEventSignalHandler : public Widget, public wxQtSignalHandler< Handler
...
```
vadz, line 472: This doesn't seem to be used at all any more, shouldn't it be removed?

vadz (commented):
Thanks, using specialization is much nicer than using the macro!

vadz:
BTW, just in case you didn't know about it, there is a dialog using text controls with and without `wxTE_PROCESS_ENTER` in the dialogs sample which could also be used for testing this.

include/wx/qt/private/winevent.h
AliKet: Oops! removed now, thanks.

AliKet:
> BTW, just in case you didn't know about it, there is a dialog using text controls with and without `wxTE_PROCESS_ENTER` in the dialogs sample which could also be used for testing this.

I tested it and it works correctly.

PR #22943, opened 2022-11-04
Use [[nodiscard]] in places with memory leak risk

PBfordev:
I understand if this is rejected, considering the attribute not very useful and just adding noise to a declaration.

I believe it could be used at least in places with a memory leak risk as I did here for MSW-specific code. 

Discarding the return value from any `Get*()`, `Is*()`, (let alone `Clone()`) function almost always indicates a bug, but there are so many such functions that it is probably not realistic to add it everywhere needed.

There may be more functions with a memory leak risk but for now, I have just added the few I could think off the top of my head. If this proposal is accepted, I will look for more candidates.

**EDIT**
It seems that the attribute could be used even with many widespread pre-C++17 compilers, see https://stackoverflow.com/a/45191044/7267315.

### DIFF
--- a/include/wx/defs.h
+++ b/include/wx/defs.h
@@ -249,7 +249,6 @@ typedef short int WXTYPE;
 
 #if wxCHECK_CXX_STD(201703L)
     #define wxFALLTHROUGH [[fallthrough]]
-    #define wxNODISCARD [[nodiscard]]
 #elif defined(__has_warning) && WX_HAS_CLANG_FEATURE(cxx_attributes)
     #define wxFALLTHROUGH [[clang::fallthrough]]
 #elif wxCHECK_GCC_VERSION(7, 0)
@@ -260,9 +259,18 @@ typedef short int WXTYPE;
     #define wxFALLTHROUGH ((void)0)
 #endif
 
-#ifndef wxNODISCARD
+/* wxNODISCARD is used to notate that the function return value should not be ignored */
+
+#if wxCHECK_CXX_STD(201703L)
+    #define wxNODISCARD [[nodiscard]]
+#elif defined(__VISUALC__)
+    #define wxNODISCARD _Check_return_
+#elif defined(__clang__) || defined(__GNUCC__)
+    #define wxNODISCARD __attribute__ ((warn_unused_result))
+#else
     #define wxNODISCARD
 #endif
+
 
 /* these macros are obsolete, use the standard C++ casts directly now */
 #define wx_static_cast(t, x) static_cast<t>(x)
--- a/include/wx/defs.h
+++ b/include/wx/defs.h
@@ -259,7 +259,7 @@ typedef short int WXTYPE;
     #define wxFALLTHROUGH ((void)0)
 #endif
 
-/* wxNODISCARD is used to notate that the function return value should not be ignored */
+/* wxNODISCARD is used to indicate that the function return value must not be ignored */
 
 #if wxCHECK_CXX_STD(201703L)
     #define wxNODISCARD [[nodiscard]]
--- a/include/wx/defs.h
+++ b/include/wx/defs.h
@@ -249,6 +249,7 @@ typedef short int WXTYPE;
 
 #if wxCHECK_CXX_STD(201703L)
     #define wxFALLTHROUGH [[fallthrough]]
+    #define wxNODISCARD [[nodiscard]]
 #elif defined(__has_warning) && WX_HAS_CLANG_FEATURE(cxx_attributes)
     #define wxFALLTHROUGH [[clang::fallthrough]]
 #elif wxCHECK_GCC_VERSION(7, 0)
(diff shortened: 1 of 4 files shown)


SpareSimian:
Should this be used for all singletons, like wxPersistenceManager::Get()? I use this in my wxApp subclass to ensure that the manager had the correct lifetime. (But being a singleton it probably lasts until after exit() destroys all global data, which is all I need.)

Blake-Madden:
You could add a `wxNODISCARD("message")` variant also for when C++20 is defined, which allows for something like this:

`[[nodiscard("Assign the return to avoid a memory leak.")]]`

Personally, I find `nodiscard` very useful; it's helped me find a couple of issues before in my own code.

PBfordev:
I realized that GCC and clang warning cannot be suppressed with just casting the result to `void*` as with MSVC.

This makes the warning rather annoying in the rare cases we really do not care about the return value, like in this PR when detaching the safe array we just "borrowed" with attach earlier.

I am not sure about the variant with the message. Even if this is accepted, the attribute is probably going to be used in wxWidgets sparingly, only in the cases where there is a risk of memory leak so the reason for the warning should be clear even without the message.

PBfordev:
I have added support for warning for more compilers.

But perhaps I should have defined the macro as  `WX_ATTRIBUTE_NODISCARD`, similarly to the already defined `WX_ATTRIBUTE_NORETURN` or `WX_ATTRIBUTE_UNUSED`...

include/wx/defs.h
```diff
@@ -259,6 +259,19 @@ typedef short int WXTYPE;
     #define wxFALLTHROUGH ((void)0)
 #endif
 
+/* wxNODISCARD is used to notate that the function return value should not be ignored */
```
vadz, line 262: I'm sorry, but is "notate" the right word to use here? I'd say "indicate". I'd also use "must" instead of "should".

include/wx/msw/ole/safearray.h
```diff
@@ -367,7 +367,7 @@ class wxSafeArray : public wxSafeArrayBase
             result = sa.ConvertToVariant(variant);
 
         if ( sa.HasArray() )
-            sa.Detach();
+            static_cast<void*>(sa.Detach()); // we do not own the array, just attached it
```
vadz, line 370: Could/should we use `wxUnusedVar()` instead of the cast? It would seem to be more clear, although the most clear would probably be to rename that function (while keeping the current name for compatibility, of course...) to `wxIgnoreUnusedWarning()` as then the code using it would really be self-documenting.

vadz (commented):
Thanks, this is definitely going to be useful!

I can merge it already (with just a tweak to the comment) if you prefer, or you could replace the casts if you agree that it would be better.

FWIW I also like the idea of having `wxNODISCARD_MSG()`, but it's not critical and could always be added later, of course.

include/wx/defs.h
PBfordev: I just copied the comment from `wxFALLTHROUGH`, probably wrongly.

I will change the wording to what you suggested, once we agreed on other changes.

include/wx/msw/ole/safearray.h
PBfordev: As I wrote elsewhere, I realized that the cast cannot be used since the warning is produced with clang and GCC.

I think for now I am going to use `wxUnusedVar()` and the renaming of the function should be left to a different PR.

PBfordev:
> I can merge it already (with just a tweak to the comment) if you prefer, or you could replace the casts if you agree that it 

So the macro should be left as is now and we should not name it use `WX_ATTRIBUTE_NODISCARD` (analogical to already defined `WX_ATTRIBUTE_NORETURN` or `WX_ATTRIBUTE_UNUSED` in the same file)?

include/wx/msw/ole/safearray.h
vadz: I did see that you write it, but then I've also seen that the CI builds passed and so thought you've fixed it by using static cast instead of a plain cast or something like this. If not, I don't understand why did they pass, they're supposed to use `-Werror` and fail if there are any warnings... Do you understand what's going on here or should I look at this?

vadz:
I prefer the current name, I think the `WX_ATTRIBUTE_XXX` ones predate the standard attributes addition and `wxNODISCARD` is consistent with `wxFALLTHROUGH` if not with them.

PBfordev:
BTW, there is a word missing in the initial commit title, it should read Use [[nodiscard]] in places with memory leak **risk**

But perhaps the commit title should be changed entirely, e.g., replace "in places" with "in functions".

include/wx/msw/ole/safearray.h
PBfordev: (Sorry, I missed this comment before).

I did not not know that (some?) CI builds do not allow any warnings. I get a few between the library and samples when building with GCC locally (at least when using CMake).

I am ashamed to admit I cannot reproduce the GCC warning when reverting from `wxUnusedVar` to `static_cast<void*>` anymore, I can't believe that extending the macro definition would change that (tried with both 11 and 17 C++ standards). But it seems that generally the warnings are always not the same (depends on C++ standard, e.g., the string template one from the XML library ), but perhaps I just miss them...

PR #22949, opened 2022-11-05
Add macros for event tables for three wxWebView events

PBfordev:
I am not aware of event tables or even `Connect()` being unsupported any more. But perhaps I just missed this and the macros were not added for a reason.

Should probably be backported if accepted.

### DIFF
--- a/include/wx/webview.h
+++ b/include/wx/webview.h
@@ -391,6 +391,18 @@ typedef void (wxEvtHandler::*wxWebViewEventFunction)
     wx__DECLARE_EVT1(wxEVT_WEBVIEW_TITLE_CHANGED, id, \
                      wxWebViewEventHandler(fn))
 
+#define EVT_WEBVIEW_FULLSCREEN_CHANGED(id, fn) \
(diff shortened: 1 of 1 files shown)


vadz:
No, I think it was just an oversight -- unless @TcT2k really didn't do it intentionally?

However, ideally, the new macros should be also documented, shouldn't they?

TcT2k:
Definitely just an oversight as I never use the macros

PBfordev:
> However, ideally, the new macros should be also documented, shouldn't they?

I thought they all already are: https://docs.wxwidgets.org/latest/classwx_web_view_event.html ?

vadz:
> > However, ideally, the new macros should be also documented, shouldn't they?
> 
> I thought they all already are: https://docs.wxwidgets.org/latest/classwx_web_view_event.html ?

Sorry, I didn't even think about checking it as I didn't think they could be documented in spite of not existing!

Pushing this to master now and will apply to 3.2 too later, thanks!

Issue #22940, opened 2022-11-03
Corner glitch when using wxCheckBox in place of wxStaticBox label


charvey2718:
### Description

When placing a wxCheckBox at the top of a wxStaticBox, as per the constructor, [here](https://docs.wxwidgets.org/3.2.0/classwx_static_box.html#aedd6c9352e976dcc55e8112143cca6f7), there is a break in the surrounding box in the top-right corner.

![2](https://user-images.githubusercontent.com/59702518/199719776-ef9b7bf1-a5ee-4c3e-aa63-a7f351196b14.png)

No break in the surrounding box is expected. The break disappears if I replace the wxCheckbox with a wxString label.

### Minimum working example

The issue is reproduced in the following MWE:
![1](https://user-images.githubusercontent.com/59702518/199721959-48fef7e5-bb42-47fc-8adf-33bbfc2d97ca.PNG)
```
#include <wx/wx.h>

class MyFrame:public wxFrame {
public:
    MyFrame(wxWindow* parent);
};

class MyApp:public wxApp {
public:
    bool OnInit() {
        wxFrame* frame=new MyFrame(NULL);
        SetTopWindow(frame);
        frame->Show();
        return true;                    
    }
};

DECLARE_APP(MyApp);
IMPLEMENT_APP(MyApp);

MyFrame::MyFrame(wxWindow* parent):wxFrame(parent,wxID_ANY,_("StatBoxMWE"),wxDefaultPosition,wxDefaultSize,wxDEFAULT_FRAME_STYLE) {
	SetClientSize(wxSize(100,100));
	wxBoxSizer* mainSizer=new wxBoxSizer(wxVERTICAL);
	
	wxCheckBox* cb=new wxCheckBox(this,wxID_ANY,_("checkbox"),wxDefaultPosition,wxDefaultSize,wxALIGN_LEFT);
	wxStaticBox* sb=new wxStaticBox(this,wxID_STATIC,cb);
	wxStaticBoxSizer *sbs=new wxStaticBoxSizer(sb,wxVERTICAL);
	
	wxStaticText* text=new wxStaticText(sb,wxID_ANY,_("hello"),wxDefaultPosition,wxDefaultSize,wxALIGN_LEFT);
	sbs->Add(text,0,wxALL,5);
	
	mainSizer->Add(sbs,1,wxEXPAND|wxALL,10);
	SetSizer(mainSizer);
}
```

### Platform and version information

[...]

PR #22952, opened 2022-11-08
Fixes to painting wxStaticBox in wxMSW

vadz:
The first commit here is just a bug fix (see #22940) and should definitely be applied. The second one does nothing and the third one should be applied too as it avoids unnecessary triple buffering.

The final one is more questionable: it's tempting to apply it because this gets rid of a lot of non-trivial code which historically created many problems but could, of course, create new problems of its own. I'd like to ask people to test this, but if not, I'm going to merge it and, at worst, revert it later if/when somebody finds any serious problems due to it.

### DIFF
--- a/samples/widgets/static.cpp
+++ b/samples/widgets/static.cpp
@@ -583,7 +583,9 @@ void StaticWidgetsPage::OnBoxCheckBox(wxCommandEvent& event)
 
 void StaticWidgetsPage::OnButtonBoxText(wxCommandEvent& WXUNUSED(event))
 {
-    m_sizerStatBox->GetStaticBox()->SetLabel(m_textBox->GetValue());
+    wxStaticBox* const box = m_sizerStatBox->GetStaticBox();
+    box->SetLabel(m_textBox->GetValue());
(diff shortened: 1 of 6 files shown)


samples/widgets/static.cpp
```diff
@@ -583,7 +583,9 @@ void StaticWidgetsPage::OnBoxCheckBox(wxCommandEvent& event)
 
 void StaticWidgetsPage::OnButtonBoxText(wxCommandEvent& WXUNUSED(event))
 {
-    m_sizerStatBox->GetStaticBox()->SetLabel(m_textBox->GetValue());
+    wxStaticBox* const box = m_sizerStatBox->GetStaticBox();
+    box->SetLabel(m_textBox->GetValue());
+    wxLogMessage("Box label changedm now is '%s'", box->GetLabel());
```
PBfordev, line 588:
```suggestion
    wxLogMessage("Box label changed, now is '%s'", box->GetLabel());
```

vadz: Oops, thanks!

PR #22953, opened 2022-11-08
Add msw.window.no-composited system option

vadz:
For some applications turning off double buffering for individual windows may be infeasible, so allow doing it globally using this system option.

Note that wxAutoBufferedPaintDC still does no buffering in wxMSW now, even when this option is set, so setting it will result in flickering in any code using it. The solution is to use wxBufferedPaintDC directly in the programs that use this option.

----

I think we might need to add this one, even if I don't like doing it -- what do you think?


beENABLED:
+1 from me. Although i never use wxWindowDC or wxClientDC in my code, i do have several redraw-issues, which i have to address one by one with MSWDisableComposited().

Issue #22957, opened 2022-11-10
Setting font for a menu item leads to a wrong size in menu on high DPI displays


Rezakarimi66:
I am using wxMenu in my program both in menu bar and pop up. In some cases I need to change the font size of my menu items. I do it by calling `SetFont` function of every menu item. Everything works fine in normal displays but when I open my application in high DPI display, I see that the menu size is wrong. If I remove the font setting line, everything works again. Below you see a sample photo in a monitor with 175% scaling. 

![image](https://user-images.githubusercontent.com/5058070/201093494-f0e815c5-62fb-4f5c-9301-849d3b7b656d.png)

Is there any workaround for this?

I put the code to reproduce it here:

```
/////////////////////////////////////////////////////////////////////////////
// Name:        minimal.cpp
// Purpose:     Minimal wxWidgets sample
// Author:      Julian Smart
// Modified by:
// Created:     04/01/98
// Copyright:   (c) Julian Smart
// Licence:     wxWindows licence
/////////////////////////////////////////////////////////////////////////////

// ============================================================================
// declarations
// ============================================================================

// ----------------------------------------------------------------------------
// headers
// ----------------------------------------------------------------------------

// For compilers that support precompilation, includes "wx/wx.h".
#include "wx/wxprec.h"


// for all others, include the necessary headers (this file is usually all you
// need because it includes almost all "standard" wxWidgets headers)
#ifndef WX_PRECOMP
    #include "wx/wx.h"
#endif

// ----------------------------------------------------------------------------
// resources
// ----------------------------------------------------------------------------

// the application icon (under Windows it is in resources and even
// though we could still include the XPM here it would be unused)
#ifndef wxHAS_IMAGES_IN_RESOURCES
    #include "../sample.xpm"
#endif

// ----------------------------------------------------------------------------
// private classes
// ----------------------------------------------------------------------------

// Define a new application type, each program should derive a class from wxApp
class MyApp : public wxApp
{
public:
    // override base class virtuals
    // ----------------------------

    // this one is called on application startup and is a good place for the app
    // initialization (doing it here and not in the ctor allows to have an error
    // return: if OnInit() returns false, the application terminates)
    virtual bool OnInit() wxOVERRIDE;
};

// Define a new frame type: this is going to be our main frame
class MyFrame : public wxFrame
{
public:
    // ctor(s)
    MyFrame(const wxString& title);

    // event handlers (these functions should _not_ be virtual)
[...]

PR #22958, opened 2022-11-10
Fix measuring menu items text

vadz:
See #22957.

@MaartenBent please let me know if you see any better way to do this, TIA!


MaartenBent (approved):
This looks good to me.

PR #22959, opened 2022-11-10
Get rid of wxHAS_MSW_TASKDIALOG checks

vadz:
All the necessary declarations should be available in the SDK headers of any still supported compilers.

Note that we still need to keep the run-time checks because the task dialog may be unavailable even in the latest OS versions when using old comctl32.dll, as it happens if the application doesn't have a manifest specifying the use of comctl32.dll v6.

See #22689.

----

Unfortunately we can't simplify this code is as much as I'd like as long as we still support comctl32.dll v5. I wonder if we could require using v6, i.e. having a valid manifest?

### DIFF
--- a/include/wx/msw/private/msgdlg.h
+++ b/include/wx/msw/private/msgdlg.h
@@ -13,19 +13,10 @@
 #include "wx/msw/wrapcctl.h"
 #include "wx/scopedarray.h"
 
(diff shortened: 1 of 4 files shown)


vadz:
Actually scratch that, I'll include this commit into a bigger PR that will also contain other similar changes elsewhere.

PR #22961, opened 2022-11-10
Drop pre Windows 7 versions support

vadz:
See #22689.

### DIFF
--- a/interface/wx/systhemectrl.h
+++ b/interface/wx/systhemectrl.h
@@ -11,9 +11,9 @@
    A helper class making it possible to use system theme for any control.
 
    Under MSW, there an alternative theme available for the list and list-like
-   controls since Windows Vista. This theme us used by Windows Explorer list
-   and tree view and so is arguably more familiar to the users than the standard
-   appearance of these controls.
+   controls. This theme us used by Windows Explorer list and tree view and so
+   is arguably more familiar to the users than the standard appearance of these
+   controls.
 
    This class is used in wxWidgets to enable this system theme in wxTreeCtrl,
    wxListCtrl and wxDataViewCtrl and thus give them the same, familiar look.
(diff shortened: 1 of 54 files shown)


vadz:
I'm not pushing this yet, but I've also found and updated a few more mentions of XP in the following files:

- README.md
- docs/doxygen/mainpages/introduction.h
- docs/doxygen/mainpages/platdetails.h

vadz:

interface/wx/systhemectrl.h
```diff
@@ -11,9 +11,9 @@
    A helper class making it possible to use system theme for any control.
 
    Under MSW, there an alternative theme available for the list and list-like
-   controls since Windows Vista. This theme us used by Windows Explorer list
-   and tree view and so is arguably more familiar to the users than the standard
-   appearance of these controls.
+   controls. This theme us used by Windows Explorer list and tree view and so
```
PBfordev, line 14:
```suggestion
   controls. This theme is used by Windows Explorer list and tree view and so 
```

There is also a missing "is" on line 13, which should read "Under MSW, there **is** an ...")

src/msw/msgdlg.cpp
```diff
@@ -542,12 +542,8 @@ int wxMessageDialog::ShowModal()
 
     wxWindowDisabler disableOthers(this, GetParentForModalDialog());
 
-#ifdef wxHAS_MSW_TASKDIALOG
-    if ( HasNativeTaskDialog() )
+    if ( TaskDialogIndirect_t taskDialogIndirect = GetTaskDialogIndirectFunc() )
```
PBfordev, line 545: I thought declaring variable in the if statement is available only since C++17?

lanurmi: No, it is available since C++98, probably even pre-standard versions. The new C++17 thing is declaring a variable without the declaration itself being the condition. Like this: `if ( int val; getVal(&val) ) ...`

PBfordev: I see, thanks for the explanation and sorry about the noise.

interface/wx/systhemectrl.h
vadz: Thanks, fixed locally (but not pushing yet to avoid rerunning the CI builds needlessly).

PR #22962, opened 2022-11-11
More C++11 fixes

vadz:
Address a few more items of #22684, notably fixes all "TODO-C++11" comments in the code.

### DIFF
--- a/samples/dialogs/dialogs.cpp
+++ b/samples/dialogs/dialogs.cpp
@@ -3062,34 +3062,6 @@ wxBEGIN_EVENT_TABLE(TestDefaultActionDialog, wxDialog)
     EVT_TEXT_ENTER(wxID_ANY,                 TestDefaultActionDialog::OnTextEnter)
 wxEND_EVENT_TABLE()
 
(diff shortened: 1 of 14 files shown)


samples/dialogs/dialogs.cpp
```diff
@@ -3117,6 +3089,27 @@ TestDefaultActionDialog::TestDefaultActionDialog( wxWindow *parent ) :
...
+            : wxTextCtrl(parent, id, value,
+                         wxDefaultPosition, wxDefaultSize, wxTE_PROCESS_ENTER)
+        {
+            Bind(wxEVT_TEXT_ENTER, &EnterHandlingTextCtrl::OnEnter, this);
+
+            SetInitialSize(GetSizeFromTextSize(GetTextExtent(value).x));
+        }
+
+    private:
+        void OnEnter(wxCommandEvent& WXUNUSED(event))
```
PBfordev, line 3107: I wonder if it was not better in new code to use lambdas for one-line event handlers like this...

vadz: Good point, thanks! Done now.

PR #22963, opened 2022-11-11
Add MSWHitTestLeftRight

skeetor:
As discussed in my previous PR with GetTabRect, I now created a fix for issue https://github.com/wxWidgets/wxWidgets/issues/4775.

This now fixes the reporting of the tab->HitTest method for left/right orientation in MSW.

### DIFF
--- a/include/wx/msw/notebook.h
+++ b/include/wx/msw/notebook.h
@@ -192,6 +192,11 @@ protected:
   wxPoint m_bgBrushAdj;
 #endif // wxUSE_UXTHEME
 
+private:
+      // Workaround for Windows bug to detect the correct flags for left/right
+      // tab orientation.
+      // Returns the pageindex or -1 if the point is not on a tab, same as HitTest().
+      int MSWHitTestLeftRight(const wxPoint &pt, long *flags) const;
 
   wxDECLARE_DYNAMIC_CLASS_NO_COPY(wxNotebook);
   wxDECLARE_EVENT_TABLE();
--- a/src/msw/notebook.cpp
+++ b/src/msw/notebook.cpp
@@ -37,6 +37,7 @@
 
 #include <windowsx.h>
 #include "wx/msw/winundef.h"
+#include <wx/display.h>
 
 #if wxUSE_UXTHEME
     #include "wx/msw/uxtheme.h"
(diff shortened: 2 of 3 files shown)


vadz (commented):
Sorry, I don't understand the idea here: why are we adding a new `MSWHitTestLeftRight()` function instead of just fixing `HitTest()` itself?

Also, why was the test for `if ( flags )` removed from `HitTest()`? It's legal to call it with null flags pointer, but now this will just crash, AFAICS.

skeetor:
> Sorry, I don't understand the idea here: why are we adding a new `MSWHitTestLeftRight()` function instead of just fixing `HitTest()` itself?

Because the checks for this are more complex than I thought, so I encapsulated it into a separate function.

> Also, why was the test for `if ( flags )` removed from `HitTest()`? It's legal to call it with null flags pointer, but now this will just crash, AFAICS.

This was an error. I missed that it is valid to pass a nullpointer if you are not interested in the flags, so I'm going to fix that.

vadz:
> Because the checks for this are more complex than I thought, so I encapsulated it into a separate function.

This is fine, but it should be private if it's only meant to be used by `HitTest()` itself, could you please change this?

Also, as with the other PR, it would be great to have (even if MSW-only) test for it to show that it works now, when it didn't work before.

skeetor:
I fixed that and made the function private. I have the impression that my function works far better than the original MS implementation, because when I tested it, I noticed that the reported flags were sometimes different than what I would have expected.

As for the test I try to think about it, because I don't really know how I could verify it. So I have to see if I can come up with something.

src/msw/notebook.cpp
```diff
@@ -729,22 +730,195 @@ bool wxNotebook::InsertPage(size_t nPage,
...
+        {
+            *flags = wxBK_HITTEST_ONLABEL;
+            return true;
+        }
+    }
+
+    if (ir.height)
+    {
+        wxDisplay disp(wxDisplay::GetFromWindow(this));
+        double scale = disp.GetScaleFactor();
```
vadz, line 849: Couldn't we just do
```suggestion
        double scale = GetDPIScaleFactor();
```
and avoid using `wxDisplay`?

src/msw/notebook.cpp
```diff
@@ -729,22 +730,195 @@ bool wxNotebook::InsertPage(size_t nPage,
...
+        int dispIndex = 0;
+        // 1.0 = 3
+        // 1.25 = 8
+        // 1.50 = 14
+        // 1.75 = 3
+        // 2.00 = 6
+        // 2.25 = 12
```
vadz, line 856: Could you please add a comment explaining where does this come from and how can it be (re)measured if this is how you arrived at these numbers?

I also don't understand at all why do we use this `dispIndex` here instead of directly assigning the displacement value, what's the point of doing that?

src/msw/notebook.cpp
```diff
@@ -729,22 +730,195 @@ bool wxNotebook::InsertPage(size_t nPage,
...
+    }
+
+    *flags = wxBK_HITTEST_ONITEM;
+    return true;
+}
+
 int wxNotebook::HitTest(const wxPoint& pt, long *flags) const
 {
+    int item = -1;
+
```
vadz, line 904: This change doesn't seem to be necessary, it's better to restore the original version and initialize the variable in its declaration.

src/msw/notebook.cpp
```diff
@@ -729,22 +730,195 @@ bool wxNotebook::InsertPage(size_t nPage,
...
-    int item = TabCtrl_HitTest(GetHwnd(), &hitTestInfo);
+    item = TabCtrl_HitTest(GetHwnd(), &hitTestInfo);
 
-    if ( flags )
+    if (flags)
     {
         *flags = 0;
 
+        if (MSWHitTestLeftRight(pt, item, flags))
+            return item;
```
vadz, line 915: Sorry, but this still doesn't seem right: now we only call this function if `flags != 0`. But if I understand your intention, we should always call it -- and check if `flags` is valid inside it.

include/wx/msw/notebook.h
```diff
@@ -192,6 +192,11 @@ class WXDLLIMPEXP_CORE wxNotebook : public wxNotebookBase
   wxPoint m_bgBrushAdj;
 #endif // wxUSE_UXTHEME
 
+private:
+      // Workaround for Windows bug to detect the correct flags for left/right
+      // tab orientation.
+      // Returns true if the point was evaluated, otherwise false.
```
vadz, line 198: This doesn't seem right, it seems to only return false if the orientation is not horizontal.

FWIW I'd rather give it the same signature as to `HitTest()` itself and only call it when the orientation is horizontal, this would seem to be more clear.

src/msw/notebook.cpp
```diff
@@ -729,22 +730,195 @@ bool wxNotebook::InsertPage(size_t nPage,
...
+    TabCtrl_GetItem(GetHwnd(), item, &ti);
+
+    wxRect tr;
+    wxRect ir;
+
+    // If the tab has an image, we need to check if the mouse is over it.
+    if (ti.iImage != -1)
+    {
+        wxIcon const &icon = GetImage(ti.iImage);
+        if (!icon.IsNull())
```
vadz, line 794: It's better to avoid double negation:
```suggestion
        if (icon.IsOk())
```

src/msw/notebook.cpp
```diff
@@ -729,22 +730,195 @@ bool wxNotebook::InsertPage(size_t nPage,
...
+
+    TCHAR buffer[256];
+    TCITEM ti;
+    ti.mask = TCIF_IMAGE|TCIF_TEXT;
+    ti.cchTextMax = sizeof(buffer) - 1;
+    ti.pszText = buffer;
+    TabCtrl_GetItem(GetHwnd(), item, &ti);
+
+    wxRect tr;
+    wxRect ir;
```
vadz, line 788: The code would be quite a bit more readable if these variables had more clear names, e.g.
```suggestion
    wxRect rectText;
    wxRect rectIcon;
```

vadz (commented):
Thanks, I appreciate your work on this, but right now the code seems to be needlessly confusing and I'd like to simplify it as much as possible to facilitate maintaining it in the future.

Having a test would also help with maintenance and doesn't seem to be very complicated me: we could simply create a notebook and use text character size to verify that hit testing a point produces the expected results. Is there some problem with it that I'm overlooking?

skeetor:
I applied all your suggestions and also commited a test. I also kept in a debugging functions which helps to debug this if there is a need for it (documented it in the code as well). To properly test this, I had to modify the notebook sample, because it is "to nice" :). The original implementation has for tabs and all of them have an icon and a label, which may be visually more appealing, but for testing it is bad. In my version I removed the icon from one tab and the label from the other, so I have now four tabs where two have an icon + label as before and the other two only have either one. Shall I commit this version as well?

skeetor:
> Thanks for the updates, I think we should merge this now, although I'm not sure if we want to really keep the debugging function in the code. If we do, we should at least use something like `#ifdef WXDEBUG_NOTEBOOK_HITTEST` around it instead of just commenting it out.

That's a good idea, much better than the commenting. 👍  

> One thing which I still wonder about: if the tests fail for top/bottom tabs, does it mean that the standard function is broken? If so, should we use your custom version for all orientations instead of just the left/right ones?

Actually, when I tested the original HitTest function for TOP/BOTTOM on Windows, I had the impression that it is not very accurate. I left it at that because nobody seemed to complain, but when I tested it with the notebook sample app, I had the impression that it often reported flags where I would have expected something different. So my implementation is pretty accurate. I can replace it as well if you want.

vadz:
> Actually, when I tested the original HitTest function for TOP/BOTTOM on Windows, I had the impression that it is not very accurate. I left it at that because nobody seemed to complain, but when I tested it with the notebook sample app, I had the impression that it often reported flags where I would have expected something different. So my implementation is pretty accurate. I can replace it as well if you want.

What exactly are the problems with the native version, i.e. how serious are they? Using native API has one important advantage: it has much higher chances to work in any future Windows versions than our own code, so we should only switch to the latter if the native API is really problematic.

skeetor:
> What exactly are the problems with the native version, i.e. how serious are they? 

I don't know how serious they are. They definitly don't work as you would expect. You can easily verify this bey running the notebook sample and Ctrl-H on different positions in the tab area. It wont match the expectation as described in the MS docs. But since there are no bug reports, it seems that most people don't notice or don't use that flags. TNH some of the flags I wondered about anyway, because I can't really see a usecase where you would have to rely on it. The ONITEM flag doesn't make much sense. Most people would capture the event when a tab is changed which is usually enough. I don't know how many programs really need to know whether a user clicked on the label or the icon. In 99% of the usecases I would expect that a developer just reacts on a tab change/activation and not on the actual click or even click location. Maybe to allow renaming a tab the label might be clickable? Anyway,the bottomline is, there doesn't seem to be complaints that it's not working precise enough, so it maybe doesn't really need to be fixed, at least it wouldn't have a high priority.

> Using native API has one important advantage: it has much higher chances to work in any future Windows versions than our own code, so we should only switch to the latter if the native API is really problematic.

Yeah, that's why I didn't replace it.

src/msw/notebook.cpp
```diff
@@ -729,22 +730,244 @@ bool wxNotebook::InsertPage(size_t nPage,
...
+        // I checked with different resolutions and it seems to be about right.
+        // My assumption was that the icon and the label are both centered in the tab
+        // and this assumption seems to be mostly right with this adjustemnt offest
+        // being required.s
+        int gap = 8;
+
+        TCHAR buffer[256];
+        TCITEM ti;
+        ti.mask = TCIF_IMAGE | TCIF_TEXT;
+        ti.cchTextMax = sizeof(buffer) - 1;
```
vadz, line 818: I just wanted to point out one important change among several I made in this code: `sizeof(buffer)` is wrong here, it's the size in bytes, i.e. double of the actual buffer size in `TCHAR`s and so could theoretically result in a buffer overflow here if we had a label longer than 128 characters.

vadz:
I've finally tested this code myself and, surprisingly, the tests seem to pass for me for the top/bottom tabs too, contrary to what the comment in the test said.

So I've enabled these tests too, simplified the code and (especially) the tests a bit and created #23023 with my version. I'm closing this one because the new PR supersedes it, but please let me know if you see anything wrong with any of my changes. TIA!

Issue #22883, opened 2022-10-17
Get rid of `wxStringImpl`


vadz:
We should just always use `std::[w]string` directly instead, our own implementation is very different (uses COW), almost not tested any longer (as it's disabled by default since a very long time) and there is just no point in avoiding the use of `std::string` any longer.

beENABLED:
FWIW; I have always used and still use the wxStringImpl. I also set wxUSE_STD_CONTAINERS to 0. Mainly because i hate std exceptions and cluttering my code with try/catch everywhere.

vadz:
Sorry, `wxUSE_STD_CONTAINERS==0` won't be supported any longer in master, we're already using `std::vector` and `std::unordered_map` and will probably use more of them soon.

You don't have to put try/catch everywhere, of course, and many/most `std::string` functions don't throw when used with correct parameters (and `wxString` will only use them with the correct values).

paulcor:
> `wxUSE_STD_CONTAINERS==0` won't be supported any longer

Did I miss a discussion where this was decided?

PR #22976, opened 2022-11-16
Remove `wxStringImpl` and always assume `wxUSE_STD_STRING==1`

vadz:
Also `wxUSE_STD_CONTAINERS_COMPATIBLY==1`.


vadz:
I'd like to merge this soon. I don't know if we should remove `wxUSE_STD_IOSTREAM` too, it's a rather weird option, but OTOH it doesn't cost much to keep it if anybody needs it.

cc @paulcor to make sure there are no surprises here (see discussion in #22883).

PR #22977, opened 2022-11-17
wxQt: wxEVT_TEXT_ENTER processing revisited

AliKet:
@vadz sorry for the noise but I think this one is better than the previous
(now merged!) one.

I really want `wxSpinCtrl` to take advantage of this, but unfortunately it doesn't derive from `wxTextEntry`.

### DIFF
--- a/include/wx/qt/private/winevent.h
+++ b/include/wx/qt/private/winevent.h
@@ -22,7 +22,11 @@
 #include <QtWidgets/QGestureEvent>
 #include <QtGui/QCursor>
 
+// redeclare wxEVT_TEXT_ENTER here instead of including "wx/textctrl.h"
+wxDECLARE_EXPORTED_EVENT(WXDLLIMPEXP_CORE, wxEVT_TEXT_ENTER, wxCommandEvent);
+
 class QPaintEvent;
+
 
 template< typename Handler >
 class wxQtSignalHandler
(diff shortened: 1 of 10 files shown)


AliKet:
Why double `WXUNUSED(WXUNUSED(...))` here ?

https://github.com/wxWidgets/wxWidgets/blob/222cf952f4618d4b907193359753e841a854702b/src/qt/spinctrl.cpp#L26-L31

vadz:
Sorry, I'm a bit confused here: so we _can_, finally, prevent Qt from closing the dialog on pressing Enter? I.e. what has changed to allow removing the hack with `g_defaultButton`?

Also, do we really want to deal with SFINAE here when we could simply have `virtual bool QtProcessEnter() { return false; }` in the base class and just override it in the classes that need to do something else? AFAICS this would be much more straightforward and we really don't care about the cost of an extra virtual function call when dispatching key events (people don't type fast enough for this to be relevant).

Finally, the double `WXUNUSED()` is just a mistake IMO.

AliKet:
> Sorry, I'm a bit confused here: so we _can_, finally, prevent Qt from closing the dialog on pressing Enter?

By good luck yes!

> I.e. what has changed to allow removing the hack with `g_defaultButton`?
> 

Previously, we request handling `wxEVT_TEXT_ENTER` from inside `wxQtComboBox::activated()` and `wxQtLineEdit::returnPressed()` signals, and apparently (or in fact, because this is what actually happen) the event (generated by pressing enter) is propagated to the dialog which in turn looks for the default button to forward the event to, and there is no way to prevent this from happening ,if `wxEVT_TEXT_ENTER` has been processed and no `Skip()` was called, other than temporarily toggle the default button off.

Fortunately the above scenario won't happen if the control really handle `wxEVT_TEXT_ENTER` in `keyPressEvent()` and the event was `accept()`ed

> Also, do we really want to deal with SFINAE here when we could simply have `virtual bool QtProcessEnter() { return false; }` in the base class and just override it in the classes that need to do something else? AFAICS this would be much more straightforward and we really don't care about the cost of an extra virtual function call when dispatching key events (people don't type fast enough for this to be relevant).
> 

I chose this approach **1)** for performance reason as you've already guessed and **2)** to avoid code duplication as much as possible.
If these two arguments are not convincing then let's add `virtual bool QtProcessEnter() { return false; }` and the derived classes override this at their will.

> Finally, the double `WXUNUSED()` is just a mistake IMO.

vadz:
Thanks for the explanation! The new approach definitely seems much better (now that I understand why it works).

As for the choice of approach: I really don't think we care that much about performance of handling "Enter". Using a virtual function seems like a much simpler solution and we could avoid code duplication simply by having a helper (inline) function that could be called from the overridden version. So yes, I think it'd be preferable to do it like this to keep things simple and easy to modify.

TIA!

AliKet:
@vadz I hope you like the new implementation!
FYI it works as expected with wxTextCtrl, wxComboBox and wxSpinCtrl[Double]

vadz:
Looks good to me, thanks!

AliKet:
Please don't merge, I have some pending commits that I want to push soon,
TIA!

AliKet:
The samples (including the widgets sample) work correctly with these changes
but a double check or even better, testing this with your application (if you are a `wxQt` guy of course) is really appreciate, TIA!

Issue #22972, opened 2022-11-16
Set up a CI build using clang cross-compiling for MSW


vadz:
I've discovered a nice [xwin tool](https://github.com/Jake-Shadle/xwin) which can be used to compile MSW-specific code using clang. I think it would be nice to set up a CI build doing this, as clang can find bugs in MSW code which are not detected neither by gcc (which also doesn't compile all of it) nor MSVC.

For the record, I had to use the following options to suppress warnings inside the SDK headers themselves (perhaps this could be integrated into wx headers, i.e. push/pop them before/after including `windows.h`):

- `-Wno-nonportable-include-path`: due to `windows.h` vs `Windows.h` mismatch.
- `-Wno-ignored-pragma-intrinsic`: plenty of those in `um/winnt.h`, maybe we do need to include `intrin.h` as the error message recommends.
- `-Wno-ignored-attributes`: `no_init_all` is used for `_CONTEXT` structs in `winnt.h`.

The full command line is something like
```
clang++-14 -Wno-nonportable-include-path -Wno-ignored-pragma-intrinsic -Wno-ignored-attributes --target=x86_64-pc-windows-msvc -I /opt/windows-msvc/crt/include -I /opt/windows-msvc/sdk/include/ucrt -I /opt/windows-msvc/sdk/include/um -I /opt/windows-msvc/sdk/include/shared ...
```
where `/opt/windows-msvc` is the directory with the files downloaded by xwin.

MaartenBent:
Do you specifically want to cross-compile, or is building with clang on Windows also fine?
In that case it might be easier to use the [msys2 github action](https://www.msys2.org/docs/ci/#github-actions-recommended), install `mingw-w64-x86_64-clang`, and build like 
`cmake -G "MinGW Makefiles" -DCMAKE_C_COMPILER=clang.exe -DCMAKE_CXX_COMPILER=clang++.exe -DCMAKE_BUILD_TYPE=Release`.

vadz:
I rather like cross-compiling from Linux (it's more convenient for me and often more flexible), but you're right, using clang under Windows could be an even simpler solution -- after all the main goal here is to fix any bugs that clang warnings could discover in MSW code. Thanks for the suggestion!

vadz:
We're not going to do this now that Maarten has added a native clang build.

PR #22979, opened 2022-11-19
Add GitHub action building with clang on MSW

MaartenBent:
See #22972

This adds a CI job that uses clang on MSW to build the libraries, all samples, tests and demos.
Even though it uses CMake, I added the job to `ci_msw.yml` instead of `ci_cmake.yml`, because the focus is on building with clang on MSW, the system that is used is less important.

Building the zlib library caused lots of warnings like ` warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]`. Since this is an external library, just ignore these.


vadz:

PR #22980, opened 2022-11-20
Remove files that are no longer useful

PBfordev:
The docs update I suggested some time ago will be a separate PR.

BTW, I noticed that masks for file/folder names and file extensions for no longer supporter compilers (older MSVS, BCC, DMC) are still used in some files, notably in _.gitignore_ and _.gitattributes_. Is this intentional?


PBfordev:
The MacOS failure is certainly unrelated (webrequest). I have no idea what happened with MSW build:
```
Run .\vc*msw*\test.exe
  .\vc*msw*\test.exe
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
build: 3.3.0 (wchar_t,Visual C++ 1900,wx containers,compatible with 3.2)
compiled using msvc 1929 (full: 192930146)
running under Windows Server 2016 (build 17763), 64-bit edition as runneradmin
Error: Process completed with exit code 1.
```
but I cannot think of any reason it would be caused by the PR.

vadz:
Mac one is at least somewhat explainable -- test fails with a timeout, so maybe the machine was just too loaded/slow to run it in time. But I have no idea why do things fail without any explanation sometimes, like it happened in the MSW test. It isn't related to your changes, of course.

Concerning the changes themselves, the only reservation I have is about removing `docs/tech/index.txt`, I think it might be better to keep it and just say that this directory used to contain files that are either obsolete or were moved to `docs/contributing`. Like this people following old links won't be completely stumped (e.g. our own wiki still mentions this directory on its "Resources" page). What do you think?

PBfordev:
> Concerning the changes themselves, the only reservation I have is about removing `docs/tech/index.txt`, I think it might be better to keep it and just say that this directory used to contain files that are either obsolete or were moved to `docs/contributing`. Like this people following old links won't be completely stumped (e.g. our own wiki still mentions this directory on its "Resources" page). What do you think?

IMO, wxWiki contains a frightening amount of outdated information and I would not take anything there as a reason for doing or not doing something with the code repository. 

If anything, I would update it there but I could not find the page(s) where the tech notes are referenced.

vadz:
If it were only the wiki, then I'd agree, but I think there are probably other references elsewhere (FWIW here is the [wiki page](https://wiki.wxwidgets.org/Resources) where I found this one) and I just hate following a link only to not find anything at the end -- keeping `index.txt` (or `README.md` or whatever) in this directory would be just nicer and doesn't seem to cost much.

PBfordev:
I do understand your point of view but I have no idea how to best achieve it (i.e., keep the folder and index.txt inside it) GIT-wise, aside from reverting the commit and basically repeat it.

As removing the tech notes was the initial impulse for this PR, perhaps we could just close the PR. Despite the files not having been used in years, there is no harm in keeping them. It's just that bit-rot keeps accumulating and I think it's easier to cut it out whenever possible.

PR #22981, opened 2022-11-20
Use variadic templates for vararg funcs

vadz:
~The last commit here requires C++14 and so can't be merged yet.~

The latest version doesn't require C++14 any more at the price of slightly changing/breaking `%c` handling.


vadz:
I'd like to merge this as I think the new version of the code is much more clear than the macro-based one.

Please let me know if anybody has any strong objections.

Issue #22629, opened 2022-07-15
OSX repaint bug in 3.2.0


JulianSmart:
When refreshing part of a window, such as a wxTreeCtrl or wxListCtrl, especially within focus events the non-updated part the client area can be corrupted, which it makes it look as though text becomes bold and icons aren't draw correctly. This is new since I used 3.1.5.

To Reproduce:

Compile and run the aui sample, auidemo. Click Item 5 in the tree. Then click on the tab wxTextCtrl 2, and then wxTextCtrl 3. When the popup message is shown, you'll see the root item of the tree pane - wxAUI Project - go bold. Resizing the pane clears the problem.

This is with:
 - wxWidgets 3.2.0
 - wxOSX
 - macOS Catalina 10.15.7
 - macOS SDK 11.1

Might it be related to wxPaintDC changes? Unfortunately it happens so much in my app that it's not useable.
Thanks for any insight!

vadz:
I'll try to debug it, but, just to be sure, is this wxAUI-specific or do you also see it without it?

Also, which `wxPaintDC` changes do you mean? I'm afraid I must have missed this. If you already know which change broke this, it would be very helpful to know it, otherwise perhaps you could try using git-bisect between 3.1.5 and 3.2.0 to find it? TIA!

JulianSmart:
Hi Vadim, I couldn't reproduce it in the treectrl sample although maybe I wasn't persistent enough. So I think it might well be triggered by AUI.

I vaguely thought it might be related to this:

https://github.com/wxWidgets/wxWidgets/issues/19108

but only because it was a paint-related change. The bug (if it's the same one) also messes up painting in my Jutoh welcome window - it's supposed to paint shadows around the welcome window, but this is corrupted - the shadow goes enormous/semi-painted and icons are visible through the HTML welcome window.

I may well have to look into bisecting but OTOH my grasp of git is so tenuous (as ever) I'll probably chuck the Mac out of the window before I ever get to grips with that. And it's one heavy 2012 MacBook Pro.

Thanks!

vadz:
[...]

PR #22987, opened 2022-11-23
Apply workaround to mask issues with repainting some controls

a-wi:
This is related to issue #22629.

### DIFF
--- a/src/generic/treectlg.cpp
+++ b/src/generic/treectlg.cpp
@@ -2576,6 +2576,10 @@ void wxGenericTreeCtrl::PaintItem(wxGenericTreeItem *item, wxDC& dc)
 
     int offset = HasFlag(wxTR_ROW_LINES) ? 1 : 0;
 
+#ifdef __WXOSX__
+    dc.DestroyClippingRegion();
+#endif // __WXOSX__
+
     if ( HasFlag(wxTR_FULL_ROW_HIGHLIGHT) )
     {
         int x, w, h;
--- a/src/osx/carbon/graphics.cpp
+++ b/src/osx/carbon/graphics.cpp
@@ -1534,6 +1534,7 @@ private:
     WXWidget m_view;
     bool m_contextSynthesized;
     CGAffineTransform m_initTransform;
+    CGRect m_initClipBox;
     CGAffineTransform m_windowTransform;
     bool m_invisible;
     int m_stateStackLevel;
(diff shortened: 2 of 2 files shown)


vadz:
@a-wi should we merge this or do you still plan to make any changes here?

a-wi:
I'm done with this workaround.
Any feedback from those who could reproduce the issue #22629 (@whalley, @codesquid) would be welcome.

whalley:
Apologies for delay in testing. I'm away at present so will have to look at this after the 12th Dec.

a-wi:
@whalley
Since restoring graphics context attributes didn't help so I think next suspect are font settings. It looks like previous (emulated) version of ResetClip() masked some font issues and new (real) version just revealed them.
I've added some more diagnostic logging to see what are font settings used in wxTreeCtrl and it would be helpful if you could pull new code from this PR, run it in debug mode and check what is logged when you are reproducing the issue.

vadz:
Could we apply the known to work workaround (there was one, right?) to master before continuing to look for a better one? I think we do need to do at least something to avoid regressions in 3.3.0, don't we?

whalley:
Apologies, not had time to spend much more testing on this due to other commitments.... Reverting according to this https://github.com/wxWidgets/wxWidgets/issues/22629#issuecomment-1315175124 works for me.

a-wi:
I we are not in a hurry I wouldn't apply crippled workaround to master. I am going to do more investigations with this issue.

vadz:
> I we are not in a hurry I wouldn't apply crippled workaround to master.

We're not in a hurry (this would need to be fixed before 3.3.0, but I don't know when it's going to happen yet), but I don't like having master broken for longer than necessary, so I'd prefer to apply a workaround which avoids the regression already. Of course, absolutely nothing prevents us from reverting it and applying a better one as soon as it becomes available, so I don't see what we lose here. OTOH having this problem in master for a couple of months already is not ideal IMO.

a-wi:
I'm also not comfortable with this issue being open for so long but the catch with this issue is that it's not universally reproducible and it's hard to isolate the root cause. And you know how looks debugging/fixing the issue if you can't reproduce it.
The catch with so called workaround is that this s not a workaround per se but rather a mask hiding the real problem. And it's not a neutral mask because it has side effects.  And what is worse we have no idea where the problem lies so this mask would be applied completely blindly. So I maybe having an open issue is not that bad option. At least there is a motivation to do something.
I will try to dig more in a new environment where maybe I would be able to reproduce the issue, for a good start.

src/generic/treectlg.cpp
```diff
@@ -2576,6 +2576,10 @@ void wxGenericTreeCtrl::PaintItem(wxGenericTreeItem *item, wxDC& dc)
 
     int offset = HasFlag(wxTR_ROW_LINES) ? 1 : 0;
 
+#ifdef __WXOSX__
+    dc.DestroyClippingRegion();
+#endif // __WXOSX__
```
vadz, line 2581: This is really weird, where is this clipping region set? We only use `wxDCClipper` in this file itself, so how can we have any clipping region here?

a-wi: The issue itself is weird - how is it possible that call to native CGContextResetClip(() impacts font weight  - so the workaround is weird too.
But this is still experiment, so maybe in the next turn something less weird will be found.

a-wi:
Won't be implemented.

PR #22989, opened 2022-11-24
wxQt slider fixes

AliKet:
There is still a problem with drawing ticks similar to [this](https://stackoverflow.com/questions/56662698/qslider-with-ticks-not-properly-drawn) which I don't know how to properly fix unfortunately ?
FYI overriding `DoGetBestSize()` solves the problem for horizontal slider only!


vadz:
Sorry, I don't know how to fix the remaining problem if imposing the minimal size doesn't help.

But I'll merge all these fixes already for now, thanks!

PR #22991, opened 2022-11-24
Make wxRadioButton test pass for wxQt

AliKet:


vadz:
Thanks for fixing this!

Issue #22988, opened 2022-11-24
use of undeclared identifiers NSURLVolumeLocalizedNameKey


crazbot:
mkdir build-cocoa-debug
cd build-cocoa-debug
../configure --enable-debug
make

### Description
build failed with error: use of undeclared identifiers `NSURLVolumeLocalizedNameKey`, `NSURLVolumeIsLocalKey`, `NSURLVolumeIsReadOnlyKey`, etc.

### Platform and version information
- wxWidgets 3.2.1
- wxWidgets wxOSX
- OS macOS 10.11.6
- Compiler Xcode7.3

oneeyeman1:
@crazbot ,
You probably shouldn't close the issue as its not being solved yet.
You did post the solution, but it is not yet available in the mainline and so it should still be open in order core devs to apply the changes.

Thx.

vadz:
Thanks, I don't know why does it compile with later SDKs, but it looks like adding the `#import` shouldn't break anything for them but fix the build with this one.

Could you please create a PR with your changes? If not, I can try making one myself, but I have no Xcode 7 installation, so I wouldn't be able to test it.

crazbot:
Sure. I'll try to create a PR as soon as possible.

imciner2:
> Thanks, I don't know why does it compile with later SDKs, but it looks like adding the #import shouldn't break anything for them but fix the build with this one.

Probably some change in imports used by the other SDK headers made it so the header was brought in for later SDKs transitively. This is unfortunately a very common type of error with standard library headers, and is usually the cause of build failures when new versions of GCC are released as well.

PR #22992, opened 2022-11-25
fix use of undeclared identifier NSURLVolumeLocalizedNameKey

crazbot:
#22988


vadz:
Thanks for the fixes, merging in master now and will also backport to 3.2 a bit later.

PR #22999, opened 2022-11-27
Fix ParseDateTime to handle default 'T' separator

Blake-Madden:
`ParseDateTime` currently fails if there is a 'T' separator in front of time component. `FormatISOCombined()` uses this separator as the default, so `wxDateTime` can't parse its own formatted results by default. Example

```cpp
wxDateTime dt;
wxASSERT(dt.ParseDateTime("1979-10-31 13:37:09"));
wxASSERT(dt.IsValid());

wxDateTime otherDt;
// OK
wxASSERT(otherDt.ParseDateTime(dt.FormatISOCombined(' ')));
// Fails using the defaults (and, according to the help, the standard format)
wxASSERT(otherDt.ParseDateTime(dt.FormatISOCombined()));
```


lanurmi:
A unit test for this should be added to `tests/datetime/datetimetest.cpp`.

vadz:

Issue #22964, opened 2022-11-12
Warn when not using comctl32.dll v6


vadz:
Ideally we'd like to get rid of all code used only to support the old, pre-XP (!), comctl32.dll versions, but we can't do it as long as we still support applications not using the manifest.

The plan is to start warning about this in 3.3.0 and, if there are not _too_ many complaints about it, make this warning an error and remove support for the old DLL in 3.3.1.

See also [this thread](https://groups.google.com/g/wx-dev/c/JP4O27KuypA).

vadz:
Done now, let's see if anybody complains.

magical-obama:
Hi!
I got the message that @vadz added a few commits ago about the manifest not being added. This is probably the wrong place to be asking, but I searched for "wxwidgets manifest" and all sorts of other things. I didn't get any meaningful results. I simply don't know what this manifest is used for, how to add it and how it should look. I'm simply lost with this message everytime I start my app.

Please help!

Max

PBfordev:
@magical-obama, this is odd. If you are using MSVC, the manifest will be created and linked in automatically.

Regardless of the compiler, you are supposed to include <wx/wx.rc> in your resource file (which every application must have, at least for the icons), this will by default result in including the application manifest with GCC or clang. So, the simplest solution would be to add the above-mentioned wx.rc to the .rc file in your project.

The wxWidgets documents this e.g. [here](https://docs.wxwidgets.org/latest/plat_msw_winxp.html).

BTW, application manifests have been in use since 2001, introduced with Windows XP. Without them, the application UI would look very odd.

magical-obama:
Thank you, @PBfordev, for the quick reply!
My situation is the following:
- Currently I'm working on Windows 11 with CLion
- My project is CMake based (since wxWidgets is cross-platform, why not)
- CMake Toolkit: Default for CLion and Included - Ninja
- I included wxWidgets as a git submodule
- In the main CMakeLists I use `add_subdirectory()` and set wxBUILD_SHARED to OFF and wxBUILD_PRECOMP to OFF (It didn't configure at all with precompiled headers enabled)
- Linking and Compiling works fine
[My main CMakeLists.txt](https://github.com/wxWidgets/wxWidgets/files/10355220/CMakeLists.txt)

```
cmake_minimum_required(VERSION 3.24)
project(ContactBook)

set(CMAKE_CXX_STANDARD 17)

set(wxBUILD_SHARED OFF CACHE BOOL "")
[...]

PR #23003, opened 2022-11-28
Warn on startup when not using comctl32.dll v6 in wxMSW

vadz:
There should be no reason to not use a manifest specifying the use of v6 of the common controls library and if everybody did it, we would be able to drop support for pre-v6 versions and simplify the code.

See #22964.


vadz:
This one will need to wait until #23006 is merged to avoid breaking the tests (and any other program) execution when using Wine.

vadz:
This works when using Wine too now, so I'd like to merge this. Please let me know if there are any objections.

PR #23006, opened 2022-12-01
Wine fixes

vadz:
Enable more functionality when using Wine, e.g. use native hyper link and command button controls with it too, as it supports them.

Also enable more tests when using Wine.

### DIFF
--- a/include/wx/msw/hyperlink.h
+++ b/include/wx/msw/hyperlink.h
@@ -43,6 +43,7 @@ public:
                 long style = wxHL_DEFAULT_STYLE,
                 const wxString& name = wxASCII_STR(wxHyperlinkCtrlNameStr));
 
+    virtual ~wxHyperlinkCtrl();
 
     // overridden base class methods
     // -----------------------------
--- a/include/wx/platinfo.h
+++ b/include/wx/platinfo.h
@@ -13,6 +13,8 @@
 
 #include "wx/string.h"
 
+class wxVersionInfo;
+
 // ----------------------------------------------------------------------------
 // wxPlatformInfo enums & structs
 // ----------------------------------------------------------------------------
@@ -436,11 +438,15 @@ protected:
     wxString m_nativeCpuArch;
 };
 
-// Returns true only for MSW programs running under Wine.
+// Return true if running under Wine and fills the provided pointer with
+// its version in this case if it's non-null.
 #ifdef __WINDOWS__
-WXDLLIMPEXP_BASE bool wxIsRunningUnderWine();
+WXDLLIMPEXP_BASE bool wxIsRunningUnderWine(wxVersionInfo* ver = nullptr);
 #else // !__WINDOWS__
-inline bool wxIsRunningUnderWine() { return false; }
+inline bool wxIsRunningUnderWine(wxVersionInfo* WXUNUSED(ver) = nullptr)
+{
+    return false;
+}
 #endif // __WINDOWS__/!__WINDOWS__
 
 #endif // _WX_PLATINFO_H_
--- a/include/wx/versioninfo.h
+++ b/include/wx/versioninfo.h
@@ -43,6 +43,24 @@ public:
     // Default copy ctor, assignment operator and dtor are ok.
 
 
+    // Check if this version is at least equal to the given one.
+    bool AtLeast(int major, int minor = 0, int micro = 0) const
+    {
+        if ( m_major > major )
+            return true;
+
+        if ( m_major < major )
+            return false;
+
+        if ( m_minor > minor )
+            return true;
+
+        if ( m_minor < minor )
+            return false;
+
+        return m_micro >= micro;
+    }
+
     const wxString& GetName() const { return m_name; }
 
     int GetMajor() const { return m_major; }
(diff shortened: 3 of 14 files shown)


vadz:
I have really no idea what's going on with Direct2D tests here and why does `GetObject()` fail to retrieve information about the font. For the reference, the test failures are all similar to this:
```
CoordinatesGCDCDirect2D::DeviceOriginChangedEx
./graphics/coords.cpp:1219

./graphics/coords.cpp:1219: FAILED:
  Failed to obtain font info for MS Shell Dlg-400-regular-8.25-33 (HFONT
  00000000060a0064): 0 in wxD2DFontData() at ./src/msw/graphicsd2d.cpp:3236
```
and both the font and its handle look fine here and, worse, none of the commits in this PR is supposed to affect them in any way.

Ah, and I can't reproduce this locally, with exactly the same Wine version, neither, of course.

a-wi:
Maybe it has something to do with the fact that "MS Shell Dlg" is a logical font? Perhaps installing corefonts with winetricks would help?

vadz:
> Maybe it has something to do with the fact that "MS Shell Dlg" is a logical font? Perhaps installing corefonts with winetricks would help?

Thanks for the suggestion! Do you know if this can be done from the command line by chance? I'm ready to try just about everything at this stage... But I still have no idea how/why could it change due to any of the commits in this PR. I'm going to try to set up tmate and bisect it in the GHA environment when I have time.

a-wi:
"sh winetricks corefonts" should work.
https://wiki.winehq.org/Winetricks#Using_winetricks

vadz:
Bisecting shows that running `HyperlinkCtrlTestCase` before `CoordinatesGCDCDirect2D::*` tests breaks the latter on the CI machine, and this is related to the commits here because they enable the use of the native control rather than the generic one in wxMSW, so reverting this change and switching back to using the generic control with Wine fixes the tests, but I'd rather not do this because the native control seems to work fine (I've tested it interactively).

Unfortunately I still don't know what else to do about it, using `winetricks corefonts` didn't help.

a-wi:
Is creating `wxHyperlinkCtrl` enough to disturb further tests or is it necessary to execute some tests, Colour, Url, CLick?

vadz:
Just creating it is enough. But I can't reproduce it in a simple example, e.g. this:
```diff
diff --git a/samples/minimal/minimal.cpp b/samples/minimal/minimal.cpp
index 501caf9096..2d275e4f42 100644
--- a/samples/minimal/minimal.cpp
+++ b/samples/minimal/minimal.cpp
@@ -26,6 +26,9 @@
     #include "wx/wx.h"
 #endif

+#include "wx/hyperlink.h"
+#include "wx/msw/private.h"
+
 // ----------------------------------------------------------------------------
 // resources
 // ----------------------------------------------------------------------------
@@ -123,6 +126,20 @@ bool MyApp::OnInit()
     // create the main application window
     MyFrame *frame = new MyFrame("Minimal wxWidgets App");

+    new wxHyperlinkCtrl(frame, wxID_ANY, "wxWidgets", "http://wxwidgets.org");
+    LOGFONTW logfont;
+    if ( !GetObjectW(GetHfontOf(frame->GetFont()), sizeof(logfont), &logfont) )
+    {
+       wxPrintf("ERROR!\n");
+    }
+    else
+    {
+       wxPrintf("Face name: %s\n", logfont.lfFaceName);
+    }
+    frame->Destroy();
+    return false;
+
+
     // and show it (the frames, unlike simple controls, are not shown when
     // created initially)
     frame->Show(true);
```
works.

vadz:
Apparently it's _destroying_ `wxHyperlinkCtrl` and not creating it which breaks things because the following is enough to reproduce the problem too:
```diff
diff --git a/tests/graphics/coords.cpp b/tests/graphics/coords.cpp
index bec0c53101..30b5eed07d 100644
--- a/tests/graphics/coords.cpp
+++ b/tests/graphics/coords.cpp
@@ -17,6 +17,9 @@
 #include "wx/dcgraph.h"
 #include "wx/dcmemory.h"

+#include "wx/app.h"
+#include "wx/hyperlink.h"
+

 // ----------------------------------------------------------------------------
 // test class
@@ -35,6 +38,8 @@ class CoordinatesDCTestCaseBase
 public:
     CoordinatesDCTestCaseBase()
     {
+       auto h = new wxHyperlinkCtrl(wxTheApp->GetTopWindow(), wxID_ANY, "wxWidgets", "http://wxwidgets.org");
+       delete h;
         m_bmp.Create(s_dcSize);
         m_dc = nullptr;
     }
```
i.e. with this patch just running `CoordinatesGCDCDirect2D::DeviceOriginChangedE` on its own fails.

I still can't reproduce the problem in the minimal sample however, which is crazy.

a-wi:
Maybe the reason it cannot be reproduced in the minimal sample is that font obtained from the frame is not MS Shell Dlg?

Another idea to help isolating the issue would be to check what would happen if parent of created `wxHyperlinkCtrl' was not top window of the app but some its child? Maybe parent window is somehow impacted by this control and this change interferes with the test because top level windows is shared?

vadz:
I think there is a real bug in Wine as it shouldn't do this:

https://github.com/wine-mirror/wine/blob/0fe562b4ae3f66f9c0e29ab8f27a4abf2ac78501/dlls/comctl32/syslink.c#L1712

I.e. I don't think this control should take ownership of the font.

PR #23010, opened 2022-12-02
wxTaskBarIcon: Prefer GetPopupMenu over CreatePopupMenu

### LINKED ISSUE #23008: wxTaskBarIcon GetPopupMenu not called before CreatePopupMenu on macOS
The [documentation for wxTaskBarIcon](https://docs.wxwidgets.org/3.2/classwx_task_bar_icon.html) states that `CreatePopupMenu()` is "Called by the library when the user requests popup menu if `GetPopupMenu()` is not overridden." And regarding `GetPopupMenu()` states that "On Windows and Unix platforms this happens when the user right-clicks the icon, so overriding this method is the simplest way to implement the standard behaviour of showing a menu when the taskbar icon is clicked." But the behaviour is ambiguous for macOS.

I would expect the `GetPopupMenu()` to be called before `CreatePopupMenu()` on macOS because there appears to be no reason against this behaviour, other than the fact that this is not explicitly stated in the documentation. If this is not the intended behaviour I would expect that to be stated, preferably along with a reason, in the documentation.

The behaviour I observe is that `GetPopupMenu()` is _only_ called before `CreatePopupMenu()` if the `wxTaskBarIcon` is constructed with `wxTBI_DOCK`, i.e when the menu is spawned from the dock icon as opposed to the task bar icon. If the `wxTaskBarIcon` is constructed with `wxTBI_CUSTOM_STATUSITEM` or `wxTBI_DEFAULT_TYPE` the `GetPopupMenu()` method is never invoked.

One can clearly see [in the code here](https://github.com/wxWidgets/wxWidgets/blob/master/src/osx/cocoa/taskbar.mm#L274) that `GetPopupMenu()` is called before `CreatePopupMenu()`, however this is _only_ the behaviour, as described above, for `wxTBI_DOCK`. [This is the code](https://github.com/wxWidgets/wxWidgets/blob/master/src/osx/cocoa/taskbar.mm#L337) that is invoked when the task bar object is `wxTBI_CUSTOM_STATUSITEM` or `wxTBI_DEFAULT_TYPE`, where it's clear that `GetPopupMenu()` is _not_ called before `CreatePopupMenu()`.

I observe this behaviour in versions 3.1.6 and 3.2 and specifically on macOS. I'd be happy to attempt a PR to solve this myself but I'm unsure whether this is a problem with the code or documentation.

williamln:
### What has changed
When clicking a `wxTaskBarIcon` subclass constructed with `wxTBI_CUSTOM_STATUSITEM` or `wxTBI_DEFAULT_TYPE` on macOS the `GetPopupMenu()` method is called before `CreatePopupMenu()`.

### Why this was changed
Before this change the `GetPopupMenu()` method was _not_ called before `CreatePopupMenu()`, which is contradictory to the intended behaviour as described in the documentation.

### Links
This resolves https://github.com/wxWidgets/wxWidgets/issues/23008, more details in the issue description.

### How can a reviewer test this change
I have verified locally that the menu provided in `GetPopupMenu()` is displayed. I would appreciate suggestions for how to create a unit test for this change (to prevent regression) as I'm not familiar with C++ unit tests, nor with the wxWidgets tests.

### DIFF
--- a/src/osx/cocoa/taskbar.mm
+++ b/src/osx/cocoa/taskbar.mm
@@ -334,7 +334,12 @@ bool wxTaskBarIconDockImpl::PopupMenu(wxMenu *WXUNUSED(menu))
 - (void) clickedAction: (id) sender
 {
     wxUnusedVar(sender);
-    wxMenu *menu = impl->CreatePopupMenu();
+    wxMenu *menu = impl->GetPopupMenu();
+    if (menu) {
+        impl->PopupMenu(menu);
+        return;
+    }
+    menu = impl->CreatePopupMenu();
     if (menu)
     {
         impl->PopupMenu(menu);


vadz:
Looks good, thanks, I'll merge it soon.

Concerning the test, it's unfortunately not simple to create one for this as we need to simulate clicking the dock icon but `wxUIActionSimulator` doesn't work under Mac, so we'll have to live without it for now.

williamln:
Thank you, appreciated!

vadz:
Applied to master now and will also cherry pick for 3.2 a bit later, thanks again!

vadz:
Already done in feea01b2bfe579800549055ed174cea92bde0c64

PR #23015, opened 2022-12-04
Always use wxDynamicLibrary in wxMSW and remove checks for it

### LINKED ISSUE #22956: Remove `wxUSE_DYNLIB_CLASS` or at least disallow setting it to 0 in wxMSW
wxMSW code uses `wxDynamicLibrary` in a lot of places to use functions not available in all (still supported) OS versions. It's painful to have to use `#if wxUSE_DYNLIB_CLASS` all the time and even more painful to write the stubs for the never used in practice `#else` branches. So I'd like to get rid of this and assume that `wxUSE_DYNLIB_CLASS` is always 1.

The only question I have is whether we should just remove this preprocessor option completely or keep it, but force it to be 1 in wxMSW in `wx/msw/chkconf.h`. There is no real reason to force it to be on for the other ports but OTOH there is very little reason to turn it off too, the code saving is minuscule.

What do the others think? If there is no sufficient interest in keeping this option, I'd prefer to remove it completely as it will make things even simpler.

vadz:
Force wxUSE_DYNLIB_CLASS to be 1 when using wxMSW and remove checks for it being 0 from the code.

No real changes.

Closes #22956.


vadz:
Applied as c2f3a3db01144e5d8564b7bd0fc59e70fbef5859 and 0f54ec8b36cd3551d7a3ee9593994adebb14cfa3

PR #23016, opened 2022-12-04
Informally deprecate wxDynamicLibrary::GetSymbolAorW()

### LINKED ISSUE #22930: Deprecate `wxDynamicLibrary::GetSymbolAorW()`
> > And lastly: Perhaps `wxDynamicLibrary::GetSymbolAorW()` is to be deprecated?
> 
> I think this name is not too bad because most Windows programmers should understand what it means and it might be used outside of wx itself, so I've also decided to leave it alone.

I meant to deprecate the two methods, since their name became deceiving as there is now no way to retrieve `A` symbols.

_Originally posted by @PBfordev in https://github.com/wxWidgets/wxWidgets/issues/22920#issuecomment-1295752783_

vadz:
This function is useless now, so don't use it nor the related wxDL_INIT_FUNC_AW() macro in wxWidgets own code any more and document it as being deprecated -- even though the function is not really deprecated at the code level because it doesn't cost much to continue to provide it and it's not worth forcing people to change their code, if they happen to use it, when updating wxWidgets.

Closes #22930.

### DIFF
--- a/include/wx/dynlib.h
+++ b/include/wx/dynlib.h
@@ -111,8 +111,10 @@ enum wxPluginCategory
 
 #ifdef __WINDOWS__
 
-// same as wxDL_INIT_FUNC() but appends 'A' or 'W' to the function name, see
-// wxDynamicLibrary::GetSymbolAorW()
+// same as wxDL_INIT_FUNC() but appends 'W' (never 'A' any more) to the
+// function name, see wxDynamicLibrary::GetSymbolAorW()
+//
+// shouldn't be used any longer, just use wxDL_INIT_FUNC() instead
 #define wxDL_INIT_FUNC_AW(pfx, name, dynlib) \
     pfx ## name = (name ## _t)(dynlib).GetSymbolAorW(#name)
 
@@ -292,6 +294,8 @@ public:
 #ifdef __WINDOWS__
     // This function is misnamed now as it always loads "W" symbol because we
     // always use Unicode now, but keeps its old name for compatibility.
+    //
+    // Use RawGetSymbol() directly with the name containing "W" in the new code.
     static void *RawGetSymbolAorW(wxDllType handle, const wxString& name)
     {
         return RawGetSymbol
--- a/interface/wx/dynlib.h
+++ b/interface/wx/dynlib.h
@@ -182,13 +182,14 @@ public:
     void* GetSymbol(const wxString& name, bool* success = 0) const;
 
     /**
-        This function is available only under Windows as it is only useful when
-        dynamically loading symbols from standard Windows DLLs. Such functions
-        have either @c 'A' (in ANSI build) or @c 'W' (in Unicode, or wide
-        character build) suffix if they take string parameters. Using this
-        function, you can use just the base name of the function and the
-        correct suffix is appended automatically depending on the current
-        build. Otherwise, this method is identical to GetSymbol().
+        @deprecated
+
+        Returns the symbol with the given name with "W" appended to it.
+
+        This function was useful to hide the differences between ANSI and
+        Unicode builds but shouldn't be used in the new code, as ANSI build is
+        not supported any longer. Just call GetSymbol() directly instead and
+        pass the name ending with `W` to it.
 
         @onlyfor{wxmsw}
     */
--- a/src/msw/app.cpp
+++ b/src/msw/app.cpp
@@ -402,13 +402,13 @@ private:
     int m_dataLen;              // length data buffer
     int m_dataLine;             // line offset
 
-    typedef DWORD (WINAPI *GetConsoleCommandHistory_t)(LPTSTR sCommands,
-                                                       DWORD nBufferLength,
-                                                       LPCTSTR sExeName);
-    typedef DWORD (WINAPI *GetConsoleCommandHistoryLength_t)(LPCTSTR sExeName);
+    typedef DWORD (WINAPI *GetConsoleCommandHistoryW_t)(LPWSTR sCommands,
+                                                        DWORD nBufferLength,
+                                                        LPCWSTR sExeName);
+    typedef DWORD (WINAPI *GetConsoleCommandHistoryLengthW_t)(LPCWSTR sExeName);
 
-    GetConsoleCommandHistory_t m_pfnGetConsoleCommandHistory;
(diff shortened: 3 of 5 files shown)


PBfordev:
Sorry, if I am being stupid again.

Did you decide that the UTF-8 build is not (going to be) supported on Windows? I have no experience with such s build but perhaps there will be demand for it in the future and we should use the narrow functions when `wxUSE_UNICODE_UTF8` is defined?

I do understand that this will reintroduce the clutter to the code again for little if any future gains, but who knows...

vadz:
I don't know if it's worth supporting `wxUSE_UNICODE_UTF8` build under MSW. It made sense when we had separate and incompatible Unicode and ANSI builds but ever since 3.0 why should people care if we use wide or narrow Windows APIs internally -- all that counts is that they can use any kind of strings in their own applications.

OTOH it's also true that the gain from removing these functions/macros is small and we could keep them... But we need to actually test UTF-8 MSW build then as otherwise it will inevitably get broken with time even if it works today (and I'm not sure if it does).

Does anybody think that supporting it is worthwhile?

PBfordev:
Sorry for continuing the off-topic but FWIW, I tried to build wxWidgets as UTF-8 on MSW, just naively patching the master with
```diff
 build/cmake/options.cmake | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)

diff --git a/build/cmake/options.cmake b/build/cmake/options.cmake
index 2c70d91763..ae884855ea 100644
--- a/build/cmake/options.cmake
+++ b/build/cmake/options.cmake
@@ -81,10 +81,10 @@ wx_option(wxUSE_STL "use standard C++ classes for everything" OFF)
 set(wxTHIRD_PARTY_LIBRARIES ${wxTHIRD_PARTY_LIBRARIES} wxUSE_STL "use C++ STL classes")
 wx_dependent_option(wxUSE_STD_CONTAINERS "use standard C++ container classes" ON "wxUSE_STL" OFF)
 
-if(NOT WIN32)
-    wx_option(wxUSE_UNICODE_UTF8 "use UTF-8 representation for strings (Unix only)" OFF)
-    wx_dependent_option(wxUSE_UTF8_LOCALE_ONLY "only support UTF-8 locales in UTF-8 build (Unix only)" ON "wxUSE_UNICODE_UTF8" OFF)
-endif()
+
+wx_option(wxUSE_UNICODE_UTF8 "use UTF-8 representation for strings" OFF)
+wx_dependent_option(wxUSE_UTF8_LOCALE_ONLY "only support UTF-8 locales in UTF-8 build" ON "wxUSE_UNICODE_UTF8" OFF)
+
 
 if(NOT WIN32)
     wx_option(wxUSE_VISIBILITY "use of ELF symbols visibility")
 include/wx/msw/wx_dpi_aware_pmv2.manifest | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/include/wx/msw/wx_dpi_aware_pmv2.manifest b/include/wx/msw/wx_dpi_aware_pmv2.manifest
index f8a475152d..6c47f5cfd5 100644
--- a/include/wx/msw/wx_dpi_aware_pmv2.manifest
+++ b/include/wx/msw/wx_dpi_aware_pmv2.manifest
@@ -30,6 +30,9 @@
       <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}" />
       <!-- Windows 10 -->
       <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />
+      <windowsSettings>
+        <activeCodePage xmlns="http://schemas.microsoft.com/SMI/2019/WindowsSettings">UTF-8</activeCodePage>
+      </windowsSettings>
     </application>
   </compatibility>
 </assembly>
```

setting `wxUSE_UNICODE_UTF8` to ON, `wxUSE_UTF8_LOCALE_ONLY` to OFF and generating MSVS 2022 solution for the shared build.

The build fails with countless instances of this error (MSVS 17.4.2)
`Error	C2492	'wxString::ms_cache': data with thread storage duration may not have dll interface	base	D:\Dev\Desktop\!Lib\wxWidgets-PB\include\wx\string.h	457	`
where `wxString::ms_cache` is declared like this
```cpp
static wxTHREAD_SPECIFIC_DECL Cache ms_cache;
```
( `Cache` is a struct declared inside `wxString`).

So it appears that MSW UTF-8 build is currently impossible anyway.

SpareSimian:
The thread local storage issue could be addressed by moving the ms_cache member out of the class and into the class implementation file, perhaps as a singleton inside GetCache. https://stackoverflow.com/questions/40744152/thread-local-c11-error-c2492-in-vs2015

PBfordev:
> The thread local storage issue could be addressed by...

Thanks for the information but after the GCC error, I basically gave up on playing with UTF-8 build on Windows.

For example, I do not understand this code
https://github.com/wxWidgets/wxWidgets/blob/32041aa2797c888e60eac52f783193f5cd8175f7/include/wx/string.h#L1751-L1757

I assumed (based on `TCHAR`-using MSW functions) that `wxString::t_str()` returns `wchar_t`-based string when built with `wxUSE_UNICODE_WCHAR` but `char`-based UTF-8-encoded string with `wxUSE_UNICODE_UTF8`. This assumption is obviously wrong.

I also do not get the third code branch in the code above (and there is more like that there), since I thought one of `wxUSE_UNICODE_WCHAR` or `wxUSE_UNICODE_UTF8` must be always set to `1` since ANSI support was dropped.

vadz:
Native UTF-8 support in MSW is relatively recent, UTF-8 build until now just meant "use UTF-8 in wxString", but it still used "W" APIs because this was the only possibility until a couple of years ago (which is nothing, in wx development speed terms :-).

It could, indeed, be interesting to use "A" APIs with UTF-8 directly now, but this would require quite a few changes and I'm not sure I'm really interested in making them. But OTOH this PR definitely goes in the wrong direction if we ever plan to do this. So I'm not so sure if we want to apply it any more. Are there people who want to use UTF-8 build under Windows 10+ only?

Also, the "third code branch" is just a leftover and should be removed, of course.

vadz:
I've created #23137 to address the "third code branch" issue. Concerning this PR itself, I still don't know if it should be applied or not, but for now I won't merge it, let's see if we decide to support UTF-8 wxMSW build in not too distant future.

vadz:
Let's not apply this for now.

PR #23022, opened 2022-12-05
Fix some wxListCtrl tests

AliKet:
The use of `CPPUNIT_ASSERT` is for consistency only. and will be replaced with `CHECK` later when the file is updated.

### DIFF
--- a/tests/controls/listbasetest.cpp
+++ b/tests/controls/listbasetest.cpp
@@ -449,18 +449,23 @@ void ListBaseTestCase::Visible()
     wxListCtrl* const list = GetList();
 
     list->InsertColumn(0, "Column 0");
+    list->InsertItem(0, wxString::Format("string 0"));
 
     int count = list->GetCountPerPage();
 
-    for( int i = 0; i < count + 10; i++ )
+    for( int i = 1; i < count + 10; i++ )
     {
         list->InsertItem(i, wxString::Format("string %d", i));
     }
 
     CPPUNIT_ASSERT_EQUAL(count + 10, list->GetItemCount());
     CPPUNIT_ASSERT_EQUAL(0, list->GetTopItem());
+    CPPUNIT_ASSERT(list->IsVisible(0));
+    CPPUNIT_ASSERT(!list->IsVisible(count + 1));
 
-    list->EnsureVisible(count + 9);
+    CPPUNIT_ASSERT(list->EnsureVisible(count + 9));
+    CPPUNIT_ASSERT(list->IsVisible(count + 9));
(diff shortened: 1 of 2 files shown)


AliKet:
> Thanks, the tests look fine, but consistently fail...

I didn't run the tests under Windows TBH because I didn't expect them to fail there. But I'll do that tomorrow to see what's going on there because they pass under `wxQt` and `wxGTK`!!

AliKet:
Well, the `wxMSW` failures are because `wxListCtrl` items are not editable by default unless `wxLC_EDIT_LABELS` flag has been specified.

Now this makes me think about ports consistency here, because `wxGTK` tests shouldn't pass too (for `wxQt` I'll fix this by adding `HasFlag(wxLC_EDIT_LABELS)`) before trying to edit labels)

Is it a breaking change if we add a check for `wxLC_EDIT_LABELS` to `wxGTK` ?

AliKet:
> Is it a breaking change if we add a check for `wxLC_EDIT_LABELS` to `wxGTK` ?

Oops! the check is already there see https://github.com/wxWidgets/wxWidgets/blob/6e940d1a108bbf4a07586d32f0d8061beb63f927/src/generic/listctrl.cpp#L2665

But why the tests was passing then (when the `wxListCtrl` was created without the `wxLC_EDIT_LABELS` ) ? is it a bug ?

vadz:
I didn't actually know myself that not specifying `wxLC_EDIT_LABELS` prevented `EditLabel()` from working (in addition to preventing the user from editing them interactively), we definitely need to document this.

And I guess we should assert that this style is set in the generic version: this won't break the existing code as it will continue to work (skipping the assert) but should hopefully tell people that they're doing something not portable.

AliKet:
> I didn't actually know myself that not specifying `wxLC_EDIT_LABELS` prevented `EditLabel()` from working (in addition to preventing the user from editing them interactively), we definitely need to document this.
> 

Or, we can document that `wxLC_EDIT_LABELS` is for user interaction only and it's always possible for the program to call `EditLabel()` to start editing if it needs to ? If yes, then we should temporary turn on `LVS_EDITLABELS` flag (if the control was created without `wxLC_EDIT_LABELS`) in `wxMSW::EditLabel()` as it would fail otherwise. what do think ?

AliKet:
> `LVS_EDITLABELS`

You are right about dynamically turning `LVS_EDITLABELS` on and off, I tried to do so and it is not working correctly, i.e. although I could enable the flag dynamically, disabling it again doesn't seem to work as you already expected!

vadz:

PR #23023, opened 2022-12-06
Fix wxNotebook::HitTest() for horizontal orientation in wxMSW

### LINKED ISSUE #4775: wxNotebook & wxBK_HITTEST_* broken (ignores layout)

vadz:
Native control doesn't seem to implement TCM_HITTEST correctly for the tabs on the left or right side, so provide our own implementation of HitTest() in this case.

Also add a unit test checking that HitTest() does something reasonable.

Closes #4775.

cc @skeetor

### DIFF
--- a/include/wx/msw/notebook.h
+++ b/include/wx/msw/notebook.h
@@ -192,6 +192,12 @@ protected:
   wxPoint m_bgBrushAdj;
 #endif // wxUSE_UXTHEME
 
+private:
+  // This function is used instead of TCM_HITTEST for implementing HitTest()
+  // for the notebooks with horizontal orientation.
+  //
+  // It has the same semantics as HitTest().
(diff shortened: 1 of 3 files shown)


vadz:
@skeetor Do you have any objections to merging this? If not, I'll do it soon.

skeetor:
No objections.

PR #23024, opened 2022-12-07
Fix wxTextCtrl creation in wxGenericListCtrl::EditLabel()

AliKet:
Ensure the edit control is created before sending `wxEVT_LIST_BEGIN_LABEL_EDIT` event, as calling `wxListCtrl::GetEditControl()` from inside the handler should always return a valid pointer

### DIFF
--- a/src/generic/listctrl.cpp
+++ b/src/generic/listctrl.cpp
@@ -2385,6 +2385,12 @@ void wxListMainWindow::ChangeCurrent(size_t current)
 
 wxTextCtrl *wxListMainWindow::EditLabel(long item, wxClassInfo* textControlClass)
 {
+    // Calling this function for control without wxLC_EDIT_LABELS flag set
+    // is not portable. i.e. on wxMSW this function cannot edit labels if
+    // the flag is not already set on the control.
+    wxASSERT_MSG( HasFlag(wxLC_EDIT_LABELS),
+                 "should only be called if wxLC_EDIT_LABELS flag is set");
+
     wxCHECK_MSG( (item >= 0) && ((size_t)item < GetItemCount()), nullptr,
                  wxT("wrong index in wxGenericListCtrl::EditLabel()") );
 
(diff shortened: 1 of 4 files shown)


src/generic/listctrl.cpp
```diff
@@ -2391,6 +2391,9 @@ wxTextCtrl *wxListMainWindow::EditLabel(long item, wxClassInfo* textControlClass
     wxASSERT_MSG( textControlClass->IsKindOf(wxCLASSINFO(wxTextCtrl)),
                  wxT("EditLabel() needs a text control") );
 
+    wxTextCtrl * const text = (wxTextCtrl *)textControlClass->CreateObject();
+    m_textctrlWrapper = new wxListTextCtrlWrapper(this, text, item);
+
```
vadz, line 2396: Can we move these lines immediately before `ProcessEvent()` call? This would avoid the problem with orphan text control being left if `wxCHECK_MSG` below fails and would also make it more clear that we need to create the control before calling the handler.

vadz (commented):
I wonder if we could create the text control on demand in `GetEditControl()` instead? The advantage of doing it like this would be that we don't do it needlessly if editing is vetoed, as I'm a bit concerned that creating the control and immediately destroying it could result in some flicker on the screen.

AliKet:
> I wonder if we could create the text control on demand in `GetEditControl()` instead? The advantage of doing it like this would be that we don't do it needlessly if editing is vetoed, as I'm a bit concerned that creating the control and immediately destroying it could result in some flicker on the screen.

In fact, I just followed the comment in `EditLabel()` here: https://github.com/wxWidgets/wxWidgets/blob/32041aa2797c888e60eac52f783193f5cd8175f7/src/msw/listctrl.cpp#L1762-L1768
which shows good reason for why the textctrl is created first before generating the  wxEVT_LIST_BEGIN_LABEL_EDIT event.

AliKet:
The failures are expected now and this PR is dependent on this #23022 for the tests to pass.

AliKet:
This should be ready now.

vadz:
Thanks, I've merged this now but as 2 commits as I wanted to get rid of the second one -- and also slightly update the last one.

PR #23025, opened 2022-12-08
3.2: Queue wxWebViewEdge events to avoid difficult to debug bugs

TcT2k:
Using HandleWindowEvent() from the WebView2 event callbacks is a common error source when user code does something blocking like RunScript() or ShowModal().
Queue the events to prevent this where possible, in wxEVT_WEBVIEW_NAVIGATING at least warn the developer instead of hanging silently.

Closes #22834, #22744, #19075

(cherry picked from commit 0ace9894c787f7ace607bfe2f4c2ae0f9f5708d2)

### DIFF
--- a/include/wx/msw/private/webview_edge.h
+++ b/include/wx/msw/private/webview_edge.h
@@ -56,6 +56,7 @@ public:
 
     bool m_initialized;
     bool m_isBusy;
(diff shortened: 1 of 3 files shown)


vadz:
Great, thanks!

PR #23029, opened 2022-12-11
wxListCtrl fixes for wxQt

AliKet:
The following events are still not implemented by this PR:
```
wxEVT_LIST_BEGIN_DRAG
wxEVT_LIST_BEGIN_RDRAG
wxEVT_LIST_COL_BEGIN_DRAG
wxEVT_LIST_CACHE_HINT
```
other than that, this PR makes `[ListCtrlTestCase]` and `[VirtListCtrlTestCase]` tests pass for me.

### DIFF
--- a/src/qt/listctrl.cpp
+++ b/src/qt/listctrl.cpp
@@ -1376,7 +1376,7 @@ bool wxListCtrl::SetItemData(long item, long data)
 bool wxListCtrl::GetItemRect(long item, wxRect& rect, int WXUNUSED(code)) const
 {
     wxCHECK_MSG(item >= 0 && (item < GetItemCount()), false,
-                 "invalid item in GetSubItemRect");
+                 "invalid item in GetItemRect");
 
     const int columnCount = m_model->columnCount(QModelIndex());
     if ( columnCount == 0 )
@@ -1395,7 +1395,7 @@ bool wxListCtrl::GetItemRect(long item, wxRect& rect, int WXUNUSED(code)) const
 bool wxListCtrl::GetSubItemRect(long item,
                                 long subItem,
                                 wxRect& rect,
-                                int WXUNUSED(code)) const
+                                int code) const
 {
     wxCHECK_MSG(item >= 0 && item < GetItemCount(),
         false, "invalid row index in GetSubItemRect");
@@ -1406,6 +1406,45 @@ bool wxListCtrl::GetSubItemRect(long item,
     const QModelIndex index = m_qtTreeWidget->model()->index(item, subItem);
     rect = wxQtConvertRect(m_qtTreeWidget->visualRect(index));
     rect.Offset(0, m_qtTreeWidget->GetHeaderHeight());
+
+    switch ( code )
+    {
+        case wxLIST_RECT_BOUNDS:
+            // Nothing to do.
+            break;
+
+        case wxLIST_RECT_ICON:
+        case wxLIST_RECT_LABEL:
+            {
+                QVariant var = index.data(Qt::DecorationRole);
+                if ( var.isValid() )
+                {
+                    const int iconWidth = m_qtTreeWidget->iconSize().width();
+
+                    if ( code == wxLIST_RECT_ICON )
+                    {
+                        rect.width = iconWidth;
+                    }
+                    else // wxLIST_RECT_LABEL
+                    {
+                        rect.x += iconWidth;
+                        rect.width -= iconWidth;
+                    }
+                }
+                else // No icon
+                {
+                    if ( code == wxLIST_RECT_ICON )
+                        rect = wxRect();
+                    //else: label rect is the same as the full one
+                }
+            }
+            break;
+
+        default:
+            wxFAIL_MSG(wxS("Unknown rectangle requested"));
+            return false;
+        }
+
     return true;
 }
 
--- a/src/qt/listctrl.cpp
+++ b/src/qt/listctrl.cpp
@@ -955,8 +955,8 @@ class wxQtListTreeWidget : public wxQtEventSignalHandler< QTreeView, wxListCtrl 
     // Data type passed to EmitListEvent() as extra data
     struct ListEventData
     {
-        int m_colOrFirstRow = -1; // column index or first (de)selected row
-        int m_colWidthOrLastRow = -1; // column width or last (de)selected row
+        int m_colOrFirstRow; // column index or first (de)selected row
+        int m_colWidthOrLastRow; // column width or last (de)selected row
     };
 
 public:
--- a/src/qt/listctrl.cpp
+++ b/src/qt/listctrl.cpp
@@ -1138,6 +1138,7 @@ bool wxListCtrl::Create(wxWindow *parent,
 
     m_qtTreeWidget->setRootIsDecorated(false);
     m_qtTreeWidget->setSelectionBehavior(QAbstractItemView::SelectRows);
+    m_qtTreeWidget->setTabKeyNavigation(true);
 
     if ( !QtCreateControl(parent, id, pos, size, style, validator, name) )
         return false;
--- a/src/qt/listctrl.cpp
+++ b/src/qt/listctrl.cpp
@@ -299,6 +299,25 @@ public:
 
             case Qt::TextAlignmentRole:
                 return header.m_align;
+
+            case Qt::DecorationRole:
+            {
+                wxImageList *imageList = GetImageList();
+                if ( imageList == nullptr )
+                    return QVariant();
+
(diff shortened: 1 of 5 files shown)


src/qt/listctrl.cpp
```diff
@@ -2108,6 +2113,36 @@ bool wxListCtrl::SortItems(wxListCtrlCompare fn, wxIntPtr data)
...
 }
 
+void wxListCtrl::ShowSortIndicator(int col, bool ascending)
+{
+    const auto header = m_qtTreeWidget->header();
+    if ( header )
+        header->setSortIndicator(col, ascending ? Qt::DescendingOrder
+                                                : Qt::AscendingOrder);
+}
+
```
AliKet, line 2123: This should be
`        header->setSortIndicator(col, ascending ? Qt::AscendingOrder : Qt::DescendingOrder);`

But there seems to be a bug somewhere which is causing it to not work as expected!

src/qt/listctrl.cpp
```diff
@@ -417,17 +442,37 @@ class wxQtListModel : public QAbstractTableModel
...
+        {
+            // wxLIST_MASK_STATE is the only mask supported by virtual wxListCtrl
+            return true;
+        }
+
         RowItem &rowItem = m_rows[row];
         ColumnItem &columnItem = rowItem[col];
 
-        QVector<int> roles;
-
```
vadz, line 430: Not sure why was this moved above, there doesn't seem to be any reason for doing this?

src/qt/listctrl.cpp
```diff
@@ -911,10 +949,20 @@ class wxQtVirtualListModel : public wxQtListModel
 
 class wxQtListTreeWidget : public wxQtEventSignalHandler< QTreeView, wxListCtrl >
 {
+    using BaseClass = wxQtEventSignalHandler< QTreeView, wxListCtrl >;
+
+    // Data type passed to EmitListEvent() as extra data which is essentially
+    // a pair of two ints with these meanings:
+    // first  = column or first (de)selected row
+    // second = column width or last (de)selected row
```
vadz, line 957: Would it be much more difficult to declare this as a struct with the named fields describing their meanings? I find the code using `pair` much more difficult to use than using named fields.

src/qt/listctrl.cpp
```diff
@@ -980,9 +1039,146 @@ class wxQtListTreeWidget : public wxQtEventSignalHandler< QTreeView, wxListCtrl
...
+    // Event handlers
+    virtual void mousePressEvent(QMouseEvent* event) override
+    {
+        const QModelIndex itemPressed = indexAt(event->pos());
+
+        BaseClass::mousePressEvent(event);
+
+        if ( !itemPressed.isValid() )
+        {
+            // wx: deselect all if clicking on empty space
```
vadz, line 1080: I don't think it's a requirement of wx API, so if Qt applications don't do this, e.g. if clicking on an empty space doesn't change the selection in standard KDE applications, I think it would make more sense not to do it in the wxQt ones neither (and updated/exclude the tests relying on this happening if we have any).

src/qt/listctrl.cpp
```diff
@@ -980,9 +1039,146 @@ class wxQtListTreeWidget : public wxQtEventSignalHandler< QTreeView, wxListCtrl
...
+                setSectionResizeMode(logicalIndex, QHeaderView::Fixed);
+                return;
+            }
+
+            m_isDragging = true;
+        }
+
+        wxQtListTreeWidget* const m_parent;
+
+        bool m_isDragging;
```
vadz, line 1176: I'd initialize it here because there is just no reason not to:
```suggestion
        bool m_isDragging = false;
```
(and then we could remove its initialization from ctor member initializer list, of course).

src/qt/listctrl.cpp
```diff
@@ -994,41 +1190,99 @@ wxQtListTreeWidget::wxQtListTreeWidget( wxWindow *parent, wxListCtrl *handler )
...
+        if ( !index.isValid() && data )
+        {
+            if ( type == wxEVT_LIST_ITEM_SELECTED ||
+                 type == wxEVT_LIST_ITEM_DESELECTED )
+            {
+                // Instead of sending hundreds of (de)selection messages, send only
+                // one for each range which is more efficient (see issue #4541)
+                // data->first is the first row in the (de)selection
+                // data->second is the last row in the (de)selection
+                wxFAIL_MSG("No implemention yet");
```
vadz, line 1222:
```suggestion
                wxFAIL_MSG("No implementation yet");
```

But, more importantly, should we really have this assert at all here? This risks being annoying when using the control as a lot of user actions would result in an assert...

src/qt/listctrl.cpp
```diff
@@ -1133,13 +1385,42 @@ int wxListCtrl::GetColumnWidth(int col) const
...
-    if ( width < 0 )
+    const auto header = m_qtTreeWidget->header();
+
+    if ( header &&
+         col == GetColumnIndexFromOrder(col) &&
+         col == GetColumnCount() - 1 )
     {
-        m_qtTreeWidget->resizeColumnToContents(width);
-        return true;
+        // Always stretch the last section if _with_ is either
```
vadz, line 1394:
```suggestion
        // Always stretch the last section if _width_ is either
```

src/qt/listctrl.cpp
```diff
@@ -1133,13 +1385,42 @@ int wxListCtrl::GetColumnWidth(int col) const
...
+        // wxLIST_AUTOSIZE or wxLIST_AUTOSIZE_USEHEADER
+        header->setStretchLastSection( width < 0 );
     }
 
-    m_qtTreeWidget->setColumnWidth(col, width);
+    if ( width >= 0 )
+        m_qtTreeWidget->setColumnWidth(col, width);
```
vadz, line 1400: Having braces around `else` but not `if` branch looks confusing, let's use them here too:
```suggestion
    {
        m_qtTreeWidget->setColumnWidth(col, width);
    }
```

src/qt/listctrl.cpp
vadz: Do you mean that it works right now, with the inverted logic? If yes, then we must not be using the same meaning for the "ascending" order as Qt does and I guess we just need to add a comment documenting it here.

vadz (commented):
Thanks for your work on this!

No real comments, I'm only not sure about leaving the "not implemented" assert as I think this is more annoying than useful.

src/qt/listctrl.cpp
AliKet: But no assertion is triggered yet and this is just a reminder for the maintainer if he tries to do a call like this:
```
    const ListEventData data { firstSel, lastSel };
    EmitListEvent(wxEVT_LIST_ITEM_SELECTED, QModelIndex(), &data);
```
which is supposed to send only one `wxEVT_LIST_ITEM_SELECTED` message and the handler can use something like: `event.GetRange();` to get the selected range generating the event (likewise for `wxEVT_LIST_ITEM_DESELECTED`)

but this is the subject of resolving this issue #4541

src/qt/listctrl.cpp
AliKet: Yes, with the inverted logic everything is ok.

After building a simple Qt program (using **QtCreator**), I discovered that the sort indicator symbol is indeed reversed!

I'll just add a comment documenting it then.

vadz:
Thanks, should this be merged now or do you plan any more changes here?

Also, I'm a bit worried by the last commit: does MSVS 2015 really have a problem with member initialization? This would be sad, as I really like using it...

AliKet:
> Thanks, should this be merged now or do you plan any more changes here?
> 

Just a commit for `wxEVT_LIST_KEY_DOWN` event generation.

> Also, I'm a bit worried by the last commit: does MSVS 2015 really have a problem with member initialization? This would be sad, as I really like using it...

Sorry for the unclear commit message!

What doesn't really work with MSVS 2015 (later versions do not have this bug) is when you try to brace initialize a `class/struct` which has in-class member initialization. although the two features are supported by that version of compiler see:

https://learn.microsoft.com/en-us/cpp/cpp/initializing-classes-and-structs-without-constructors-cpp?redirectedfrom=MSDN&view=msvc-140

https://learn.microsoft.com/en-us/cpp/cpp/class-member-overview?view=msvc-140

However, I couldn't find a formal bug report for this specific issue!

AliKet:
@vadz  I think i've done with this for now, i.e. the tests and the listctrl sample work as expected. the missing functionalities will be added later, support for drag and drop for example.

Thanks for the review.

vadz:
Merged now, thanks again!

Issue #23007, opened 2022-12-01
wxGrid mouse scroll is slow 


DjPasco:
### Description
`wxGrid` works slow on mouse scroll when custom line draw functionality is used (`wxGrid::DrawGridLine` is overridden).

Custom line drawing functionality looks like this and it draws dashed line:

```
static void DrawLine(HDC hdc, int32_t nX, int32_t nY)
{
    RECT rect;
    rect.left   = nX;
    rect.top    = nY;
    rect.bottom = nY + 1;
    rect.right  = nX + 1;

    ::ExtTextOut(hdc, 0, 0, ETO_OPAQUE, &rect, L"", 0, nullptr);
}

void wxGridEx::DrawGridLine(wxDC &dc, int nFrom, int nTo, int nStaticCoord, int nIndex, bool bVertical)
{
    // Fill rectangle with color
    HDC hdc = dc.GetHDC();
    const COLORREF before = ::GetBkColor(hdc);
    ::SetBkColor(hdc, RGB(0, 0, 0));

    for (int32_t nCoord = nFrom; nCoord < nTo; nCoord += 2) 
    {
        int32_t nX = nCoord;
        int32_t nY = nStaticCoord;
        if ( bVertical ) {
            std::swap(nX, nY);
        }

        DrawLine(hdc, nX, nY);
    }

    ::SetBkColor(hdc, before);
}
```

**Main problem** is that `wxGrid::DrawAllGridWindowLines` is called to many times.
**Result**: when user scrolls grid fast - it scrolls with delay and it feels unnatural compared to other grids in other software (Ms Excel for example).

### Proposed solution (works in my case)
`DrawAllGridWindowLines` is used to redraw grid lines which can change position on scroll. That could be fixed - just config grid to scroll in full rows and not in partial rows.
Current behavior after scroll: two lines scrolled fully and the third one partially
Expected behavior after scroll: all three lines scrolled fully.

When scroll works with the full line functionality - no grid lines redraw will be required and `wxGrid::DrawAllGridWindowLines` call could be omitted. MS Excel works in the same manner - it always scrolls full lines.

Full line scroll could be achieved by calling `SetScrollLineY`:

```
wxStyleGrid::wxStyleGrid(...) 
: wxGrid(..)
{
    EnableGridLines(true);
    const int32_t nHeight = GetRowHeight(1);
    SetScrollLineY(nHeight);
}
```

[...]

PR #23032, opened 2022-12-12
Minimizing grid lines redraw when custom draw function is used

DjPasco:
Related to  issue "wxGrid mouse scroll is slow #23007"

This commit makes the following changes:

1. Default scroll Y value is changed to the row height. That will result in full row scroll instead of partial scroll.
2. Grid lines redraw is disabled when mouse scroll is detected. That is achieved by using `m_win->SetRefData(pRefObjData);`

Grid scroll with mouse is much faster after those changes.

### DIFF
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -2625,6 +2625,9 @@ void wxGridWindow::OnMouseWheel( wxMouseEvent& event )
 {
     if (!m_owner->ProcessWindowEvent( event ))
         event.Skip();
+
+    // To redraw grid after scroll. If not called - some lines are not updated in grids where cell is combined from multiple items.
+    m_owner->Refresh();
 }
 
 // This seems to be required for wxX11/wxGTK otherwise the mouse
@@ -2839,6 +2842,9 @@ void wxGrid::InitPixelFields()
 #else
     m_defaultRowHeight += 4;
 #endif
+
+    // To make full row scroll. In other case rows are scrolled partialy.
+    m_yScrollPixelsPerLine = m_defaultRowHeight;
 
     // Don't change the value when called from OnDPIChanged() later if the
     // corresponding label window is hidden, these values should remain zeroes
@@ -6791,6 +6797,11 @@ void wxGrid::DrawAllGridWindowLines(wxDC& dc, const wxRegion & WXUNUSED(reg), wx
 {
     if ( !m_gridLinesEnabled || !gridWindow )
          return;
+
+    wxObjectRefData *pRefData = this->GetRefData();
+    if (pRefData) {
+        return;
+    }
 
     int top, bottom, left, right;
 
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -6790,6 +6790,8 @@ wxGrid::DrawRangeGridLines(wxDC& dc,
     dc.DestroyClippingRegion();
 }
 
+extern wxWindow *wxTheWindowBeingScrolled;
+
 // This is used to redraw all grid lines e.g. when the grid line colour
 // has been changed
 //
@@ -6798,8 +6800,8 @@ void wxGrid::DrawAllGridWindowLines(wxDC& dc, const wxRegion & WXUNUSED(reg), wx
     if ( !m_gridLinesEnabled || !gridWindow )
          return;
 
-    wxObjectRefData *pRefData = this->GetRefData();
-    if (pRefData) {
+    // Minimizing grid lines redraw count.
+    if (this == wxTheWindowBeingScrolled) {
         return;
     }
 
--- a/src/generic/scrlwing.cpp
+++ b/src/generic/scrlwing.cpp
@@ -474,6 +474,8 @@ void wxScrollHelperBase::SetTargetWindow(wxWindow *target)
 // scrolling implementation itself
 // ----------------------------------------------------------------------------
 
+wxWindow *wxTheWindowBeingScrolled;
+
 void wxScrollHelperBase::HandleOnScroll(wxScrollWinEvent& event)
 {
     int nScrollInc = CalcScrollInc(event);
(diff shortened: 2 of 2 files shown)


src/generic/scrlwing.cpp
```diff
@@ -550,6 +555,10 @@ void wxScrollHelperBase::HandleOnScroll(wxScrollWinEvent& event)
         m_win->Refresh(true, GetScrollRect());
     }
 #endif // __WXUNIVERSAL__
+
+    if (m_win) {
+        m_win->UnRef();
+    }
```
vadz, line 561: And whatever we use for the flag we need a RAII helper (e.g. a scope guard) to ensure that it's reset on function exit instead of doing it manually here.

src/generic/grid.cpp
```diff
@@ -2625,6 +2625,9 @@ void wxGridWindow::OnMouseWheel( wxMouseEvent& event )
 {
     if (!m_owner->ProcessWindowEvent( event ))
         event.Skip();
+
+    // To redraw grid after scroll. If not called - some lines are not updated in grids where cell is combined from multiple items.
+    m_owner->Refresh();
```
vadz, line 2630: Why do we only need to do this when scrolling using the wheel? Couldn't the same thing happen when dragging the scrollbar thumb around with the mouse?

vadz (commented):
Thanks for making this PR to show what you meant by optimizing the grid redraw!

I think it would be useful to provide a way to detect if scrolling is in progress (although ideally it would work for wxGTK too -- which is not the case currently) and also an event indicating when we don't scroll any more (so that the window could have the final refresh). But this needs to be done with some well-defined API and we can't apply the current patch as is, sorry.

I also wonder if it's really a good idea to refresh the entire grid even when you scrolled only a single row into the view. The old approach was to minimize refreshes as much as possible, but I guess that now, with everything being double-buffered, it doesn't matter much any more.

src/generic/grid.cpp
DjPasco: In my experiments I did not noticed any issues when scrollbar was used. Refresh was only required after mouse scroll.

src/generic/scrlwing.cpp
```diff
@@ -485,6 +485,11 @@ void wxScrollHelperBase::HandleOnScroll(wxScrollWinEvent& event)
         return;
     }
 
+    if (m_win) {
+        wxObjectRefData *pRefObjData = new wxObjectRefData;
+        m_win->SetRefData(pRefObjData);
+    }
```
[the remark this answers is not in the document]
DjPasco: Thank You for the review. I already removed wxObjectRefData usage in my proposal and replaced it with global wxTheWindowBeingScrolled.

src/generic/scrlwing.cpp
DjPasco: Completely agree on that remark.

DjPasco:
Hello, I added a suggested wxTheWindowBeingScrolled implementation. But I see, that "continuous-integration/appveyor/pr" test is failing.

vadz:
I've restarted this build, let's see if it was a real failure or just a gremlin...

DjPasco:
> I've restarted this build, let's see if it was a real failure or just a gremlin...

It seems that all checks have passed without errors.

DjPasco:
Checks failed on 
  ./uris/url.cpp:37: FAILED:

Maybe it is possible to restart checks?

vadz:
Thanks, the test failure is clearly a spurious one, so it's not a problem. I'll try to test and apply this, thank you for your patience.

DjPasco:
Hello, I see that one test is failing, does it related to proposed changes?

vadz:
Sorry, I completely forgot about this.

Looking at the code again I'm still quite lost, i.e. I have no idea why is it ok to do what it does. It looks like the explanation basically boils down to "it works better", but this is not really satisfactory. In particular, I don't understand why is it fine not to redraw the grid lines while scrolling. I understand that not drawing anything is faster, but it doesn't look right neither? And, OTOH, the new `Refresh()` looks like a very blunt (and inefficient) solution to the problem, whatever it is, because now the entire window is being repainted all the time.

Setting `m_yScrollPixelsPerLine` is the only thing here which looks definitely correct, but we should get rid of `GRID_SCROLL_LINE_Y` too now as it's not really used any more.

DjPasco:
> Sorry, I completely forgot about this.
> 
> Looking at the code again I'm still quite lost, i.e. I have no idea why is it ok to do what it does. It looks like the explanation basically boils down to "it works better", but this is not really satisfactory. In particular, I don't understand why is it fine not to redraw the grid lines while scrolling. I understand that not drawing anything is faster, but it doesn't look right neither? And, OTOH, the new `Refresh()` looks like a very blunt (and inefficient) solution to the problem, whatever it is, because now the entire window is being repainted all the time.
> 
> Setting `m_yScrollPixelsPerLine` is the only thing here which looks definitely correct, but we should get rid of `GRID_SCROLL_LINE_Y` too now as it's not really used any more.

Current grid vertical scrolling is allowing a partial row scroll. This behavior results in a requirement to always redraw not only internals in the cells, but also a grid lines. Especially grid lines redraw is slow when grid contains several thousands rows.
This can be avoided by always making grid lines position static. This achieved by making scroll distance equal to the row height.
When this logic is used - text or internals of the cells should be redrawn. `Refresh()` function is used for that reason.

vadz:
> Current grid vertical scrolling is allowing a partial row scroll. This behavior results in a requirement to always redraw not only internals in the cells, but also a grid lines.

As I said, changing the vertical scroll step is the only thing here which is obviously correct, so I'm all for doing it.

> Especially grid lines redraw is slow when grid contains several thousands rows. This can be avoided by always making grid lines position static.

But this only works if all rows are of the same height. What happens, with these changes, if they are not?

> This achieved by making scroll distance equal to the row height. When this logic is used - text or internals of the cells should be redrawn. `Refresh()` function is used for that reason.

But when the lines are not static (as now), the text needs to be redrawn too, and yet it happens without a full `Refresh()`?

vadz:
Sorry, but I really can't convince myself that not redrawing the window while it's scrolled can be the right thing to do, so I'm applying the change to the vertical scroll line size (thanks!) and closing this one as it shouldn't be applied unless I'm really missing something here.

PR #23037, opened 2022-12-15
Do use host-specific pkg-config when cross-compiling

vadz:
This re-enables the use of $host-pkg-config after it was disabled again by d6ddfe87e1 (Disable build system pkg-config when cross-compiling to non-Linux, 2022-11-30).

---

@paulcor Does this address [your concerns](https://github.com/wxWidgets/wxWidgets/commit/d6ddfe87e1e75fc26cb50e2e8be215443460d444#commitcomment-92076182)?

### DIFF
--- a/configure
+++ b/configure
@@ -25052,23 +25052,21 @@ $as_echo "no" >&6; }
 fi
 
 if test "$build" != "$host"; then
-    case "${host}" in
-        *-linux-*)
+    case "$PKG_CONFIG" in
+        */$host-pkg-config )
             ;;
 
-        *)
-            disable_build_system_pkg_config=1
-            ;;
+        * )
+            { $as_echo "$as_me:${as_lineno-$LINENO}: WARNING: Not using native pkg-config when cross-compiling." >&5
+$as_echo "$as_me: WARNING: Not using native pkg-config when cross-compiling." >&2;}
+
+
+
+                                                if test -z "$PKG_CONFIG_LIBDIR"; then
+                PKG_CONFIG_LIBDIR=/dev/null
+                export PKG_CONFIG_LIBDIR
+            fi
     esac
-fi
-
-if test "$disable_build_system_pkg_config" = 1; then
-
-
-                if test -z "$PKG_CONFIG_LIBDIR"; then
-        PKG_CONFIG_LIBDIR=/dev/null
-        export PKG_CONFIG_LIBDIR
-    fi
 fi
 
 
--- a/configure.in
+++ b/configure.in
@@ -2098,32 +2098,28 @@ dnl the build system, they are at best useless and can be harmful (e.g. they
 dnl may define options inappropriate for the cross-build, resulting in the
 dnl failure of all the subsequent tests).
 dnl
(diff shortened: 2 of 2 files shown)


vadz:
I get problems cross-compiling for wxMSW all the time because I don't have `x86_64-w64-mingw32-pkg-config` installed (and why would I? I don't have any MSW libraries I want to use on my Linux system anyhow) because `pkg-config` finds libcurl which results in adding `-I/usr/include/x86_64-linux-gnu` to the compilation flags which breaks compiling code targeting MSW as this directory contains Linux-specific headers.

I don't know what's the best solution is here but it seems bad to break the build in this common case, especially as it's not really obvious to understand why it is broken. And the only solution I know of is to set `PKG_CONFIG_LIBDIR=/dev/null` when running configure which is far from obvious neither.

But most importantly, I just don't see when would it ever make sense to use build system `pkg-config`. Can you think of any case when anyone would want to do it? IMNSHO this is just a bad default and autoconf should never fall back to build system tools if a host-specific one can't be found.

paulcor:
> I get problems cross-compiling for wxMSW all the time because I don't have x86_64-w64-mingw32-pkg-config installed (and why would I?

I don't understand your logic here. You would install the cross `pkg-config` so you don't "get problems ... all the time".

vadz:
> I don't understand your logic here. You would install the cross `pkg-config` so you don't "get problems ... all the time".

I'd like to make 2 separate points here:

1. It's very much not obvious that you need to install `$host-pkg-config` to build wxWidgets correctly even though you absolutely do _not_ need to use it at all for actually finding the libraries, as wxWidgets comes with its own built-in versions of them. I hadn't realized it myself and I really don't think many wxWidgets users would think to do it. We could document it, of course, but explaining that you need to install it even if you don't use any external libraries is going to be an interesting logical exercise. How exactly would you justify having a requirement that is not needed on its own? From a purely theoretical viewpoint, I believe it's just wrong to require something that we don't actually use.
2. From a more practical point of view, this PR does nothing if `$host-pkg-config` is installed, so it only avoids difficult to diagnose errors if it happens not to be available. And it basically allows things to work out of the box in a configuration where they mysteriously broke with weird error messages before. How can it be seen as anything other than an improvement?

IMO we only have 2 choices here: either complain and *stop* with an error if `$host-pkg-config` is not available when cross-compiling because we know that things will break without it when targeting MSW and libcurl is installed (which is a pretty common case) or do what this PR does. And the latter seems much more helpful to me.

Do you (or anybody else) disagree with this?

vadz:
I'd appreciate getting a reply to the above or at least a confirmation that you don't violently object to merging this PR. TIA!

paulcor:
You may take my silence as tacit capitulation.

vadz:
> You may take my silence as tacit capitulation.

Thanks, I wasn't sure if it was that or just due to the seasonal pleasantness (or maybe just vacations).

In any case, I really don't think this makes things worse and we can always re-re-revert this later if it turns out to still create any problems.

PR #23038, opened 2022-12-15
Fix scaling of standard MSW platform icons in high DPI

### LINKED ISSUE #23031: wxInfoBar icon to large on High DPI screen
On my High DPI Screen the standard Icon used in the wxInfoBar is to large. My display uses a scaling factor of 2.5. It's a 4k display. Seems that the display factor is applied twice.

```
wxInfoBar* infoBar = new wxInfoBar(this);
infoBar ->ShowMessage(_("Unsupported PDF feature"), wxICON_INFORMATION);
```
The icon is loaded by `wxWindowsArtProvider::CreateBitmap` in file artmsw.cpp.

m_sizeDefault seems to be 40,40 which is 16 x 2.5 and so 40,40 is already scaled. `GetPreferredBitmapSizeAtScale` returns 3.0 and thus `DoGetPreferredSize` returns 120,120. Size 40,40 comes from `wxArtProvider::GetNativeSizeHint(client)`.

Is this behavior intentional?

image

vadz:
Use the correct scaling factor when creating the bitmap in wxWindowsArtProvider, otherwise the bitmap was wrongly rescaled again when actually using it.

It still seems wrong to always use the primary display scaling here, but we just don't have any window to use in this code and at least it is consistent and works when primary display scaling is not 100% now.

Closes #23031.

----

@MaartenBent Please let me know if you see anything wrong with this.

### DIFF
--- a/src/msw/artmsw.cpp
+++ b/src/msw/artmsw.cpp
@@ -23,6 +23,7 @@
 #endif
 
 #include "wx/image.h"
+#include "wx/display.h"
 #include "wx/dynlib.h"
 #include "wx/volume.h"
 #include "wx/msw/private.h"
@@ -109,8 +110,12 @@ MSWGetBitmapFromIconLocation(const TCHAR* path, int index, const wxSize& size)
     // the same anyhow, of course, but if it isn't, the actual icon size would
     // be size.x in both directions as we only pass "x" to SHDefExtractIcon()
     // above.
+    //
+    // Also always use the primary display scale factor here because this is
+    // what SHDefExtractIcon() uses.
     wxIcon icon;
-    if ( !icon.InitFromHICON((WXHICON)hIcon, size.x, size.x) )
+    if ( !icon.InitFromHICON((WXHICON)hIcon, size.x, size.x,
+                             wxDisplay().GetScaleFactor()) )
         return wxNullBitmap;
 
     return wxBitmap(icon);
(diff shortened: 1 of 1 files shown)


MaartenBent (approved):
Looks correct to me.

vadz:
@a-wi It's the first time I see it, but clipping region tests don't fail only on GitHub, they also failed in the initial [AppVeyor build](https://ci.appveyor.com/project/wxWidgets/wxwidgets/builds/45695709/job/9i2t35ilotd6xp5d). I'm afraid we're going to need to disable them for now because they just keep failing sporadically and this is annoying.

a-wi:
Clipping tests are definitely deterministic, all constructs are static, so they should not fail randomly. The only potential for non-deterministic behaviour have tests with `wxClientDC` because testing window is not fully under control - if it's hidden or obscured by testing environment the clipping state can be affected.
Just disabling tests doesn't seem to be a good idea to me. It looks like a turning a blind eye.

vadz:
> Clipping tests are definitely deterministic, all constructs are static, so they should not fail randomly.

I agree, but the fact is that somehow they do.

> The only potential for non-deterministic behaviour have tests with `wxClientDC` because testing window is not fully under control - if it's hidden or obscured by testing environment the clipping state can be affected.

I don't understand how can this happen, but if it does, could we test for it and skip the test if we detect that the window is not visible?

> Just disabling tests doesn't seem to be a good idea to me. It looks like a turning a blind eye.

I'd prefer not to disable them too, but dealing with their failures is just too annoying in the current state, so I'd prefer to disable the tests using `wxClientDC` (especially considering that, as I already mentioned, I think, we want to phase out the use of this class completely anyhow) if we can't do anything better.

vadz:
Applied to master now, should be backported to 3.2 too later.

PR #23039, opened 2022-12-15
Allow selecting and copying text in wxMessageDialog on GTK

imciner2:
Normally, the GTK message dialogs have non-selectable text, but that also means users can't copy the text. This can be inconvenient if the dialog is used for error messages, since then the error message can't be copied.

Instead, provide a new style that enables selection (and copying) on the text labels used in the GTK message dialog to fix that.

To test this, a simple patch to the dialogs sample can be done:
```
─────────────────────────────────────────────────────────────────────────────────────────────┐
• samples/dialogs/dialogs.cpp:978: void MyFrame::MessageBox(wxCommandEvent& WXUNUSED(event)) │
─────────────────────────────────────────────────────────────────────────────────────────────┘
978|978|                            "Message box text",
979|979|                            wxCENTER |
980|980|                            wxNO_DEFAULT | wxYES_NO | wxCANCEL |
981|   |-                           wxICON_INFORMATION);
   |981|+                           wxICON_INFORMATION | wxMESSAGE_SELECT);
982|982| 
983|983|     wxString extmsg;
984|984|     if ( dialog.SetYesNoCancelLabels
```
Without the patch, the dialog text can't be selected or copied, with the patch it can be.

I think this would also be nice to backport to 3.2, since it doesn't change any ABI (the only user-facing change is the new define for the style), and it doesn't change the default behavior (the new behavior is opt-in by providing the style). If that is the case, the interface comment should be updated to the 3.2 version it is included in.

### DIFF
--- a/src/gtk/msgdlg.cpp
+++ b/src/gtk/msgdlg.cpp
@@ -187,10 +187,29 @@ void wxMessageDialog::GTKCreateMsgDialog()
         gtk_window_set_keep_above(GTK_WINDOW(m_widget), TRUE);
     }
 
+    // A GTKMessageDialog usually displays its labels without selection enabled,
+    // so we enable selection to allow the user to select+copy the text out of
+    // the dialog.
+    GtkMessageDialog * const msgdlg = GTK_MESSAGE_DIALOG(m_widget);
+
+    GtkWidget* area = gtk_message_dialog_get_message_area(msgdlg);
+    GList* labels = gtk_container_get_children(GTK_CONTAINER(area));
+
+    for ( GList* elem = labels; elem; elem = elem->next )
+    {
+        GtkWidget* widget = GTK_WIDGET( elem->data );
+
+        if ( GTK_IS_LABEL(widget) )
+        {
+            gtk_label_set_selectable(GTK_LABEL(widget), TRUE);
+        }
+    }
+
+    g_list_free(labels);
+
     // we need to add buttons manually if we use custom labels or always for
     // Yes/No/Cancel dialog as GTK+ doesn't support it natively
     const bool addButtons = buttons == GTK_BUTTONS_NONE;
-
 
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, I totally agree with allowing to select and copy the message box text, but I wonder if we really even need a style for this, i.e. if there is any drawback in just always making the labels selectable?

imciner2:
We could make it always selectable, I just thought it might be better to preserve the default GTK behavior instead of changing it. On the other hand, if macOS and MSW allow the selection/copying by default, then it might make sense to always do it on GTK as well for better consistency.

Can someone check on MSW and macOS on the dialogs sample to see the dialog behavior?

vadz:
MSW doesn't allow copying the contents of the individual fields (unless when using the real message box and not the task dialog due to wx-specific `wxMessageDialog::ReplaceStaticWithEdit()` hack), but does allow copying the entire box contents, for both the old message box and the task dialog by pressing `Ctrl+C` in it (it's one of the best hidden secrets of the Windows UI, but very nice to know).

No idea about Mac, I suppose it doesn't support it, but in any case I don't think it should prevent us from enabling this for GTK by default/always. The only thing that would warrant introducing a new flag would, IMO, be some visible difference in appearance with the selectable labels and without them. But if they look the same, and AFAICS they do, then it shouldn't be a problem to do it and it has the advantage of making the text of all the existing message boxes copyable without any further changes.

imciner2:
Ok, I have gone ahead and pushed the change to make it always allow selection of the text.

vadz:
Thanks! Applying now and I think this should be backported to 3.2 too, if nobody complains.

vadz:
Sorry, forgot to say: I've changed it a little to use `wxGtkList` and avoid manual `g_list_free()` call, I hope you don't mind.

PR #23040, opened 2022-12-15
Fix nullptr access when missing menuitem in OSX menu handler

imciner2:
Sometimes the menu item passed to the command process function might be null, so we can't dereference it to test checkability in that case.

I don't have a clear case from the samples when this happens, but there was already a test for a null menu item when looking for the menu ID, so this guards the rest of the accesses. This was found from a crash we saw in KiCad (https://gitlab.com/kicad/code/kicad/-/issues/13149) which had a stacktrace of
```
hread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libwx_osx_cocoau-3.2.0.1.0.dylib	       0x107c79b04 wxMenu::HandleCommandProcess(wxMenuItem*) + 188
1   libwx_osx_cocoau-3.2.0.1.0.dylib	       0x107c79adc wxMenu::HandleCommandProcess(wxMenuItem*) + 148
2   AppKit                        	       0x1c004d400 -[NSApplication(NSResponder) sendAction:to:from:] + 456
3   AppKit                        	       0x1c0142f84 -[NSMenuItem _corePerformAction] + 444
4   AppKit                        	       0x1c0142c78 -[NSCarbonMenuImpl performActionWithHighlightingForItemAtIndex:] + 100
5   AppKit                        	       0x1c018b40c -[NSMenu performActionForItemAtIndex:] + 196
6   AppKit                        	       0x1c018b330 -[NSMenu _internalPerformActionForItemAtIndex:] + 100
7   AppKit                        	       0x1c018b13c -[NSCarbonMenuImpl _carbonCommandProcessEvent:handlerCallRef:] + 116
8   AppKit                        	       0x1c0126b70 NSSLMMenuEventHandler + 728
9   HIToolbox                     	       0x1c5e63418 DispatchEventToHandlers(EventTargetRec*, OpaqueEventRef*, HandlerCallRec*) + 1116
10  HIToolbox                     	       0x1c5e62884 SendEventToEventTargetInternal(OpaqueEventRef*, OpaqueEventTargetRef*, HandlerCallRec*) + 356
11  HIToolbox                     	       0x1c5e789c4 SendEventToEventTarget + 40
12  HIToolbox                     	       0x1c5ed8ca0 SendHICommandEvent(unsigned int, HICommand const*, unsigned int, unsigned int, unsigned char, void const*, OpaqueEventTargetRef*, OpaqueEventTargetRef*, OpaqueEventRef**) + 420
13  HIToolbox                     	       0x1c5efd9f8 SendMenuCommandWithContextAndModifiers + 56
14  HIToolbox                     	       0x1c5efd988 SendMenuItemSelectedEvent + 352
15  HIToolbox                     	       0x1c5efd7b4 FinishMenuSelection(SelectionData*, MenuResult*, MenuResult*) + 100
16  HIToolbox                     	       0x1c5efe16c MenuSelectCore(MenuData*, Point, double, unsigned int, OpaqueMenuRef**, unsigned short*) + 564
17  HIToolbox                     	       0x1c5efde88 _HandleMenuSelection2 + 416
18  AppKit                        	       0x1bffe63fc _NSHandleCarbonMenuEvent + 288
19  AppKit                        	       0x1bffe61e8 _DPSEventHandledByCarbon + 68
20  AppKit                        	       0x1bfe480a4 -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 3380
21  AppKit                        	       0x1bfe399b4 -[NSApplication run] + 596
22  libwx_osx_cocoau-3.2.0.1.0.dylib	       0x107cf5048 wxGUIEventLoop::OSXDoRun() + 196
23  libwx_osx_cocoau-3.2.0.1.0.dylib	       0x107bec368 wxCFEventLoop::DoRun() + 40
24  libwx_osx_cocoau-3.2.0.1.0.dylib	       0x107b43330 wxEventLoopBase::Run() + 192
25  libwx_osx_cocoau-3.2.0.1.0.dylib	       0x107b13d38 wxAppConsoleBase::MainLoop() + 208
26  libwx_osx_cocoau-3.2.0.1.0.dylib	       0x107c931c4 wxApp::OnRun() + 36
27  kicad                         	       0x104dfad48 APP_KICAD::OnRun() + 28
28  libwx_osx_cocoau-3.2.0.1.0.dylib	       0x107b75e54 wxEntry(int&, wchar_t**) + 88
29  kicad                         	       0x104df9444 main + 52
30  dyld                          	       0x1059390f4 start + 520
```

### DIFF
--- a/src/osx/menu_osx.cpp
+++ b/src/osx/menu_osx.cpp
@@ -361,7 +361,7 @@ bool wxMenu::HandleCommandProcess( wxMenuItem* item )
 {
     int menuid = item ? item->GetId() : 0;
     bool processed = false;
-    if (item && item->IsCheckable())
+    if ( item && item->IsCheckable() )
         item->Check( !item->IsChecked() ) ;
 
     if ( SendEvent( menuid , (item && item->IsCheckable()) ? item->IsChecked() : -1 ) )
--- a/src/osx/menu_osx.cpp
+++ b/src/osx/menu_osx.cpp
@@ -361,10 +361,10 @@ bool wxMenu::HandleCommandProcess( wxMenuItem* item )
 {
     int menuid = item ? item->GetId() : 0;
     bool processed = false;
-    if (item->IsCheckable())
+    if (item && item->IsCheckable())
         item->Check( !item->IsChecked() ) ;
 
-    if ( SendEvent( menuid , item->IsCheckable() ? item->IsChecked() : -1 ) )
+    if ( SendEvent( menuid , (item && item->IsCheckable()) ? item->IsChecked() : -1 ) )
         processed = true ;
 
     if(!processed && item)


vadz:
Thanks, we should apply this, but I wonder why do we call `HandleCommandProcess()` with a null item in the first place when it doesn't seem to do anything useful in this case anyhow and sending `wxEVT_MENU` with id of 0 might actually be unexpected/harmful rather than useful.

The only potential useful thing it does is to call `OSXAfterMenuEvent()` but it doesn't look like we'd lose anything if we didn't do it when we don't have any valid item.

@csomor Would it be fine to skip calling this function (and other related ones, such as `wxMenu::HandleMenuItemHighlighted()`) if there is no valid item or am I missing something here?

csomor:
> @csomor Would it be fine to skip calling this function (and other related ones, such as `wxMenu::HandleMenuItemHighlighted()`) if there is no valid item or am I missing something here?

If there is no other common handling needed at the container level for a handler, then I'd say we should not call a handler if the corresponding wx-target is a nullptr. 

How this can exist at all is not really obvious to me, I'm not seeing everything involved in the stacktrace so it's difficult to pinpoint. I'd really think a `wxNSMenuItem` can only exist if there is a `wxMenuItem` peer and they're both bundled by a `wxMenuItemCocoaImpl` and I cannot imagine yet how this could happen

craftyjon:
Hi all,

While investigating this I've found another problem.  The original problem with the null item may have been a red herring, although it seems to me that the code should still be cleaned up as this function should either check for null in all paths or have a guarantee that item is not null.

The new problem is that `GetInvokingWindow()` can return a stale pointer resulting in a use-after-free crash.

I have started to trace this out but haven't yet figured out:

1) Where is the invoking window supposed to be set for a top-level application menu on wxMac?  I set a breakpoint on SetInvokingWindow and it is never called in our application in the test case that produces the bug (selecting a menu item in a submenu on the application menubar)
2) Even though the invoking window is never set for this menubar, something is overwriting this pointer at some point in the process, since it is initialized to `nullptr` but ends up being invalid memory inside `wxMenu::HandleCommandProcess`.  I haven't yet figured out how this happens.

craftyjon:
Update, my previous investigation was a bit misguided.  I have finally figured out the actual root cause:  Because of the addition of the `OSXAfterMenuEvent` API, it is now implicitly forbidden to destroy a wxMenu inside of a menu event handler, as the client event handler is no longer the last piece of code that runs in the old menu's context.  Unfortunately, KiCad currently does that in a number of places so we'll have to clean that up.

vadz:
We could check if the window is still alive in `wxMenu::HandleCommandProcess()` after executing the handler by binding to its `wxEVT_DESTROY` and only calling `OSXAfterMenuEvent()` on it if it's still there.

OTOH this whole `OSXAfterMenuEvent()` hack is rather ugly and doesn't even solve the problem completely, as explained in the commit message of 0f1d1d2ae9 (Update wxSearchCtrl popup menu state after showing it, 2021-04-04). It would be really nice to do replace it with something else, but I still don't really know what...

vadz:
I am still not sure how to fix this in the best way, but I think we could just call `OSXAfterMenuEvent()` _before_ calling the user-defined handler, thus ensuring that the window is still alive when it's called. This seems like the simplest fix that really shouldn't break anything, does anybody see any problem with doing this?

vadz:
I've created #23100 with my alternative proposed solution after testing that it worked (menu items in the search control menu in the widgets sample are still checked/unchecked correctly) and I think it's preferable to the changes here, so I'm closing this one.

Of course, please let me know if you disagree or if you see any problems with the other PR. TIA!

Issue #23048, opened 2022-12-18
Fix saving TIFF images with MinGW 12 in 32 bits


vadz:
Debian Sid upgraded MinGW gcc version from 10 to 12 and this seems to have broken saving TIFF images, as the corresponding unit tests now consistently fail in the CI builds, see e.g. [this log](https://github.com/wxWidgets/wxWidgets/actions/runs/3725632855/jobs/6318562687).

vadz:
The PR above provides a workaround for now, so I'm going to merge it, but I still have no idea what the real problem is. I can reproduce it locally, in a build using -O0 and the minimal code to reproduce it I could come up with was just this:
```cpp
TEST_CASE("SaveTIFF", "")
{
    wxInitAllImageHandlers();
    wxImage image("horse.png");

    size_t len;
    {
    wxMemoryOutputStream memOut;
    REQUIRE(image.SaveFile(memOut, wxBITMAP_TYPE_TIFF));
    len = memOut.GetLength();
    wxMemoryInputStream memIn(memOut);
    REQUIRE(memIn.IsOk());
    wxImage savedImage(memIn);
    REQUIRE(savedImage.IsOk());
    }
    {
    wxMemoryOutputStream memOut;
    REQUIRE(image.SaveFile(memOut, wxBITMAP_TYPE_TIFF));
    REQUIRE(memOut.GetLength() == len);
    }
}
```

I.e. after reading a TIFF image from the buffer back, saving it the next time doesn't work correctly, with a chunk at the end missing.

Adding some tracing to TIFF code I see that in the first, working call, there is a bunch of largish (~8KiB) writes, then a seek to the end of the buffer and a few more small writes, but in the second call this seek simply doesn't happen and we're left with an incomplete file.

There is something very wrong here because if I simply add a `printf()` before `TIFFWriteScanline()` call, it fixes the problem -- even though it clearly can't affect anything libtiff does. Except that it does...

vadz:
OK, I've finally found the real problem, see the PR above, which was due to some random memory bits just happening to form a NaN pattern in memory sometimes. I love C API :-(

vadz:
And I've just realized that it's actually a relative recent (2+ years) regression due to the changes of b5a554b9a6 (Avoid -Wdouble-promotion warnings, 2020-10-14). Fixing warnings sometimes introduces bugs too :-/

PR #23049, opened 2022-12-18
Enable optimizations for MSW CI cross-builds

vadz:
Attempt to work around #23048.


vadz:
I'm going to merge this to fix the CI builds that are always failing currently, even if this clearly doesn't solve the underlying problem in the non-optimized builds.

PR #23050, opened 2022-12-18
locale: Avoid using glibc specific defines on musl

kraj:
musl does not provide some glibc-only enum members e.g. _NL_ADDRESS_LANG_NAME

Signed-off-by: Khem Raj <raj.khem@gmail.com>


vadz:
Thanks, we probably should just do this to fix the immediate problem, but I think an even better solution would be to check for `_NL_ADDRESS_LANG_NAME` availability in configure (and CMake).

If anybody has any plans to do it soon, please let me know, but otherwise I'll merge this one.

Thanks again!

PR #23053, opened 2022-12-19
Improve wxWidgets naming conventions docs

PBfordev:
BTW, is the discrepancy in DLL names on Windows intentional? I.e., AFAICT, `arch` (as in __x64_ for 64-bit builds) is used only by the MSVS solution and official binaries (both MSVC and GCC) but not even when using MSVC via bundled makefile.vc or CMake...

### DIFF
--- a/docs/contributing/about-platform-toolkit-and-library-names.md
+++ b/docs/contributing/about-platform-toolkit-and-library-names.md
@@ -28,7 +28,7 @@ When an item of the name is empty, `'_'` or `'-'` are omitted.
 
     libwx_$(toolkit)$(widgetset)u_$(flavour)_$(name)-$(version)-$(host).$(lib_extension)
 
-*Windows libraries* (VS solution, makefile.gcc/vc, CMake on Windows/MinGW, shared libraries on Windows):
+*Windows libraries* (VS solution, makefile.gcc/vc, CMake on Windows, shared libraries on Windows):
 
     wx$(toolkit)$(widgetset)$(version)u$(debug)_$(flavour)_$(name)_$(compiler)_$(arch)_$(vendor).$(lib_extension)
 
@@ -69,10 +69,10 @@ Eg. for wxWidgets 3.1.5, `$version` is `315` for Windows shared libraries,
 
 The rationale for this is that under UNIX-like systems it is desirable
 that differently 'minor numbered' releases can be installed together,
-meaning your old 2.2 apps can continue to work even if you migrate
-development to the next stable or unstable release (eg. 2.3, 2.4),
-but binary compatibility is maintained between point releases (those
-with the same major.minor number).
+meaning 3.0 apps can continue to work even if you migrate development
+to the next stable or unstable release (eg. 3.2, 3.3), but binary
+compatibility is maintained between point releases (those with the same
+major.minor number).
 
 A known break in binary compatibility should be addressed by updating
 the library soname (see the notes in configure.in for details on this).
@@ -85,10 +85,10 @@ supported one, but which used to be optional.
 --------------------------------------------------------------------
 
 `$debug` is set to `'d'` for the libraries using debug version of CRT and is empty
-for release libraries. It is only really useful for the libraries created with MSVC
-projects and makefiles, as MSVC debug and release CRT are not ABI-compatible,
-but is also used by `makefile.gcc` under MSW for consistency with `makefile.vc`.
-When using configure under MSW or UNIX, it is always empty.
+for release libraries. It is only really useful for the libraries created with MSVC,
+as MSVC debug and release CRT are not ABI-compatible, but is also used by MinGW
+(makefile.gcc, CMake) for consistency with MSVC builds. When using configure under
+MSW or UNIX, it is always empty.
 
 --------------------------------------------------------------------
 
@@ -105,13 +105,14 @@ It is only added to shared libraries on Windows.
 
 --------------------------------------------------------------------
 
-`$arch` is used by MSVC solutions. It is empty for 32-bit builds and
+`$arch` is used only by MSVS solutions. It is empty for 32-bit builds and
 `'x64'` for 64-bit builds. It is only added to shared libraries.
 
 --------------------------------------------------------------------
 
 `$vendor` is an optional name appended to the library name. It is only
-added to shared libraries on Windows. It defaults to `'custom'`.
+added to shared libraries on Windows. It defaults to `'custom'` and is
+empty for the official Windows binaries.
 
 --------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


PBfordev:
> It should already be used for nmake builds, however, see `ARCH_SUFFIX` in `makefile.vc`. This relis on `CPU=x64` being defined in the environment, but this should be the case in MSVS command prompt window, AFAIK. Of course, you can also set `TARGET_CPU=X64` manually when using these makefiles, but this shouldn't be necessary.

I see, thanks. I think that since at least MSVS 2017, the variable is not set any more when running "x64 Native Tools Command Prompt for VS 201x". I cannot see any similar variable set for v2017 while 2019 and 2022 have `VSCMD_ARG_TGT_ARCH`.

MaartenBent:
If you want to add this to CMake, something like this should work:
```diff
 build/cmake/functions.cmake | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/build/cmake/functions.cmake b/build/cmake/functions.cmake
index ca0bdd9784..01dc2a095b 100644
--- a/build/cmake/functions.cmake
+++ b/build/cmake/functions.cmake
@@ -212,6 +212,9 @@ function(wx_set_target_properties target_name)
     if(wxCOMPILER_PREFIX)
         wx_string_append(dll_suffix "_${wxCOMPILER_PREFIX}")
     endif()
+    if(WIN32_MSVC_NAMING AND wxARCH_SUFFIX)
+        wx_string_append(dll_suffix "${wxARCH_SUFFIX}")
+    endif()
     if(wxBUILD_VENDOR)
         wx_string_append(dll_suffix "_${wxBUILD_VENDOR}")
     endif()
```

vadz:
I still don't know what happened with `CPU` environment variable, I'm almost sure that it used to be defined some time ago, but maybe I'm imagining things. In any case, it doesn't help now and I don't know how can we detect 64-bit compiler from nmake makefile any longer now. I guess we'll just have to tell people to use `CPU=x64` explicitly...

I've created #23071 with Maarten's suggestion from above and will merge this PR itself soon, thanks again!

PBfordev:
>.... I don't know how can we detect 64-bit compiler from nmake makefile any longer now. I guess we'll just have to tell people to use `CPU=x64` explicitly...

@vadz, FWIW, I have checked the system environment variables for 64-bit command prompts in all supported MSVS versions (i.e., 2015-2022). It seems they all set environment variable "Platform" to "x64", which matches the target CPU architecture (other tool prompts can set it e.g. to "x86" or "ARM").  None of them set "TARGET_CPU" or "VISUALSTUDIOPLATFORM" but 2017-2022 also have "VSCMD_ARG_TGT_ARCH" set to the same value as "Platform".

Perhaps the nmake makefiles can check for "Platform" system environment variable and if it is set to "x64" use it for `ARCH_SUFFIX`? I know nothing about makefiles and this probably has to be done in a bakefile anyway, so I did not attempt to do anything with this.

###  Tested MSVSs
**MSVS 2015**
_Version:_ Microsoft Visual Studio Express 2015 for Windows Desktop Version 14.0.25431.01 Update 3
_OS:_ 64-bit Windows 8.1
_Command prompt run:_ VS2015 x86 x64 Cross Tools Command Prompt
_Note:_  There is no native 64-bit prompt there, may be due to the edition?

**MSVS 2017**
_Version:_ Microsoft Visual Studio Express 2017 for Windows Desktop Version 15.9.50
_OS:_ 64-bit Windows 7
_Command prompt run:_ x64 Native Tools Command Prompt for VS 2017

**MSVS 2019**
_Version:_ Microsoft Visual Studio Community 2019 Version 16.11.18
_OS:_ 64-bit Windows 10
_Command prompt run:_ x64 Native Tools Command Prompt for VS 2019

**MSVS 2022**
_Version:_ Microsoft Visual Studio Community 2022 Version 17.4.3
_OS:_ 64-bit Windows 10
_Command prompt run:_ x64 Native Tools Command Prompt for VS 2022

vadz:
This is really weird, but maybe this variable was set for even older compilers, considering that it was actually added back in e40b74ba83 (Add CPU_TARGET option to msvc makefiles for targeting AMD64 and IA64, 2005-10-16)...

Anyhow, looking at this more closely, we already do use `PLATFORM` for setting `ARCH_SUFFIX`, via `VISUALSTUDIOPLATFORM` make variable, since 284b25ce02 (Automatically detect if target architecture is x64 in .vc file, 2020-03-26) (the makefiles were rebaked in a slightly later commit), so I still don't know why it doesn't work for you.

It also looks like we could use `PROCESSOR_ARCHITECTURE` instead of `CPU` as this seems to be defined appropriately in all cases, but, again, it looks like the current makefiles should already work and they do work for me. Would you have any idea what is different in your case?

PBfordev:
@vadz wrote
> It also looks like we could use `PROCESSOR_ARCHITECTURE` instead of `CPU` as this seems to be defined appropriately in all cases...

I may be missing something but when I run "x86 Native Tools Command Prompt for VS 2022" on my 64-bit Windows 10 and execute `set` I get
>Platform=x86
**PROCESSOR_ARCHITECTURE=AMD64**
VSCMD_ARG_HOST_ARCH=x86
VSCMD_ARG_TGT_ARCH=x86

and the same for the MSVS 2019 version run on the same computer and MSVS 2017 version run on a different one (also 64-bit Windows 10).

> but, again, it looks like the current makefiles should already work and they do work for me. Would you have any idea what is different in your case?

Sorry, no idea. All I do is basically
```
cd %WXWIN%\build\msw
set "VSCMD_START_DIR=%cd%"
call "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvars64.bat"
nmake -f makefile.vc BUILD=debug SHARED=1 
```
But if you build in a similar manner and it works for you, I guess the problem may somehow be on my side.

vadz:
Oops, I have
```
PROCESSOR_ARCHITECTURE=x86
PROCESSOR_ARCHITEW6432=AMD64
```
for MSVS 2015 but I indeed also have only
```
PROCESSOR_ARCHITECTURE=AMD64
```
for MSVS 2019.

But if `Platform` is set appropriately -- and it should be, AFAICS, because I could find the following in the maze of MSVS batch files:
```bat
if "%VSCMD_VCVARSALL_INIT%" NEQ "" (
    set Platform=%VSCMD_ARG_TGT_ARCH%
)
```
our `ARCH_SUFFIX` should still be set correctly...

And I think I finally understand what's going on: in https://github.com/wxWidgets/wxWidgets/blob/cd48a182c9980afbb73c04266342876fb511eea3/build/msw/makefile.vc#L1885-L1890
we only use `ARCH_SUFFIX` as part of the DLL name in the official builds, but not otherwise. This was done back in 9043a590ce (Use COMPILER_VERSION and ARCH_SUFFIX in libraries names in Windows makefiles., 2012-07-01) and the commit message explicitly mentions this behaviour but doesn't explain why was it done like this. My guess would be that it was for compatibility, as always...

Anyhow, this is trivial to change, but now the question becomes whether we should indeed change it, knowing that it risks breaking existing distribution scripts and what not...

PBfordev:
> Anyhow, this is trivial to change, but now the question becomes whether we should indeed change it, knowing that it risks breaking existing distribution scripts and what not...

So perhaps we should just change CMakefile accordingly (use ARCH_SUFFIX only for the official build, or just revert the recent change since CMake is not used for the official builds) and adjust the naming docs.

vadz:
Thinking more about this, using different names for the projects and the makefiles really doesn't make much sense and we have to change one of them to avoid them being gratuitously inconsistent with each other, and it's probably better to change the makefiles, so let's try doing this.

SpareSimian:
If the arch is now part of the library name, is there any reason to have separate library _directories_ for each arch? (I'd expect separate directories would still be needed for different compilers, unless that's also to be encoded in the library name. Compare to Boost's library names, which encode quite a bit in each file name, allowing them all to share a single directory on the target system.)

PBfordev:
> If the arch is now part of the library name, is there any reason to have separate library _directories_ for each arch? 

It is a part only of the DLL name, not the import library name (e.g., `wxmsw330ud_core_vc_x64_custom.dll` vs. `wxmsw33ud_core.lib`). This means that no, the different architectures cannot share a folder. 

It would also not be possible due to the name of the build-specific include subfolder, which  is either `mswu` or `mswud`.

PR #23058, opened 2022-12-20
wxOverlay implementation for wxMSW

### LINKED ISSUE #23047: wxOverlay broken by use of `WS_EX_COMPOSITED`
As can be seen in the drawing sample, nothing drawn using `wxOverlay` can be seen any longer by default but setting `wx_msw_window_no_composited=1` allows the expected selection rectangle to be seen.

This is actually not due so much to `wxOverlay` itself as to using `wxClientDC` in the sample but then, of course, this is the way it's going to be used in 99% (or maybe even 100%) of the existing code, so we must find some way to make this work. In wxMSW we could use `wxScreenDC` which should still work even with composited windows, but it would be great to have a better solution.

@AliKet You've worked on this functionality for GTK 3, where `wxClientDC` can't be used neither, would you know of any way to make this work for composited windows under MSW by chance?

AliKet:
Fixes #23047


AliKet:
@vadz I see that master is using bakefile version 0.2.13 which I couldn't find online!
sorry I know you are busy working at other things, but would you help me fixing this
if you please ? TIA.

vadz:
AFAICS 0.2.13 is the latest version on `legacy-0.2-branch` of https://github.com/vslavik/bakefile/, did you have any problems with getting it? In any case, I can rebake before merging, just please let me know if I should.

AliKet:
> AFAICS 0.2.13 is the latest version on `legacy-0.2-branch` of https://github.com/vslavik/bakefile/, did you have any problems with getting it? In any case, I can rebake before merging, just please let me know if I should.

Sorry I thought the [legacy release 0.2.12](https://github.com/vslavik/bakefile/releases/tag/v0.2.12.1) was the latest one, umm...

Anyway, if you could, I'll leave the rebake to you than.

TIA.

Kumazuma:
missing src/msw/overlay.cpp

src/msw/overlay.cpp
```diff
@@ -0,0 +1,227 @@
...
+        m_overlayWindow->Destroy();
+        m_overlayWindow = nullptr;
+        m_window = nullptr;
+    }
+}
+
+wxOverlay::Impl* wxOverlay::Create()
+{
+    wxString value;
+    if ( wxGetEnv(wxASCII_STR("wx_msw_window_no_composited"), &value) && value == wxASCII_STR("1") )
```
PBfordev, line 220: Should not the check here be
```suggestion
    if ( wxSystemOptions::GetOptionInt("msw.window.no-composited") )
```

src/msw/overlay.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+        m_overlayWindow->Destroy();
+        m_overlayWindow = nullptr;
+        m_window = nullptr;
+    }
+}
+
+wxOverlay::Impl* wxOverlay::Create()
+{
+    wxString value;
```
PBfordev, line 220:
```suggestion
```
Sorry for all the nitpicking. 

Talking about nitpicking, I would also unify how Win32 function names are referred in the `wxLogLastError()` messages: Unlike in `wxOverlayWindow::Destroy()`, in `wxOverlayWindow::Create()`  the parentheses after the function name are missing.

include/wx/private/overlay.h
```diff
@@ -17,7 +17,7 @@
     #define wxHAS_NATIVE_OVERLAY 1
 #elif defined(__WXOSX__) && wxOSX_USE_COCOA
     #define wxHAS_NATIVE_OVERLAY 1
-#elif defined(__WXGTK3__)
+#elif defined(__WXGTK3__) || defined(__WXMSW__)
     #define wxHAS_NATIVE_OVERLAY 1
     #define wxHAS_GENERIC_OVERLAY 1
```
vadz, line 22: Is is correct to define this for wxMSW if it doesn't work there?

samples/drawing/drawing.cpp
```diff
@@ -2055,13 +2055,17 @@ void MyCanvas::OnMouseMove(wxMouseEvent &event)
...
-        dc.SetPen( *wxGREY_PEN );
-        dc.SetBrush( wxColour( 192,192,192,64 ) );
-#else
-        dc.SetPen( wxPen( *wxLIGHT_GREY, 2 ) );
-        dc.SetBrush( *wxTRANSPARENT_BRUSH );
-#endif
+        if ( m_overlay.IsNative() )
+        {
+            dc.SetPen( *wxGREY_PEN );
+            dc.SetBrush( wxColour( 192,192,192,64 ) );
```
vadz, line 2061: Does this really work under MSW? I think only ports using wxGC for implementing wxDC (i.e. Mac and GTK 3) support alpha channel in brush colours.

src/msw/overlay.cpp
PBfordev: > feel free to leave any comments even if it's just a nitpick.

Be careful what you wish for. ;) OK, the last one, I promise. The file header in `src/msw/overlay.cpp` misses _Purpose_ and _Created_ lines (and Author's e-mail). 

Not sure if all these fields are still useful, but that's what the current coding guidelines [say](https://wxwidgets.org/develop/coding-guidelines/#wxwin_header).

src/msw/overlay.cpp
```diff
@@ -0,0 +1,227 @@
...
+namespace // anonymous
+{
+static const wxChar* gs_overlayClassName = nullptr;
+
+class wxOverlayWindow : public wxNativeContainerWindow
+{
+public:
+    static wxWindow*  Create(bool fullScreen = false)
+    {
+        static const wxChar* OVERLAY_WINDOW_CLASS = wxS("wxOverlayWindow");
```
vadz, line 26: Could you please add a comment explaining why do we need a separate window class for this window? It doesn't seem obvious to me...

src/msw/overlay.cpp
```diff
@@ -0,0 +1,227 @@
...
+#include "wx/msw/private.h"
+
+namespace // anonymous
+{
+static const wxChar* gs_overlayClassName = nullptr;
+
+class wxOverlayWindow : public wxNativeContainerWindow
+{
+public:
+    static wxWindow*  Create(bool fullScreen = false)
```
vadz, line 24: Do we use, or plan to use, `fullScreen` parameter? If not, I'd rather remove it for clarity.

Also, and less importantly, wx typically uses `Create()` for non-static ctor-like function, so I'd rename this to `New()`, which is used for similar factory functions in a couple of other classes.

src/msw/overlay.cpp
```diff
@@ -0,0 +1,227 @@
...
+    {
+        m_overlayWindow->Destroy();
+        m_overlayWindow = nullptr;
+        m_window = nullptr;
+    }
+}
+
+wxOverlay::Impl* wxOverlay::Create()
+{
+    if ( wxSystemOptions::GetOptionInt("msw.window.no-composited") )
```
vadz, line 220: Do we really need this? I.e. if this implementation works with non-composited windows too (and I don't see why it wouldn't?), let's just always use it.

src/msw/overlay.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+    const wxPoint pt = dc->GetDeviceOrigin();
+    winDC.Blit(0, 0, m_rect.width, m_rect.height, dc, -pt.x, -pt.y);
+}
+
+void wxOverlayImpl::Clear(wxDC* dc)
+{
+    wxCHECK_RET( IsOk(), wxS("overlay not initialized") );
+
+    dc->SetBackground(wxBrush(wxTransparentColour));
```
vadz, line 204: Is this different/better than using `*wxTRANSPARENT_BRUSH`?

vadz (commented):
Thanks a lot, this looks nice and simple to me, so we should definitely use layered windows if they work well (including in the non-composite case).

I just wonder if we couldn't inherit `wxOverlayWindow` from `wxWindow`, using `wxNativeContainerWindow` for something we create ourselves looks very strange to me, it's our own window after all, i.e. not "native" in the sense this class uses.

samples/drawing/drawing.cpp
AliKet: This is true. But for `wxMSW` the alpha channel is just ignored and this is not a problem because the overlay transparency is controlled by this call:
`::SetLayeredWindowAttributes(hwnd, 0, 128, LWA_COLORKEY|LWA_ALPHA)`

BTW if I change **128** to **64** and set the pen to `dc.SetPen( wxColour(20, 20, 20 );` the rubber band looks almost like the native one!

src/msw/overlay.cpp
AliKet: I first added this for debugging purposes. but when I tried to remove it I get an instant crash!

vadz: Err, this is rather worrisome because I don't see why should it crash at all. Do you have any information about this crash, i.e. a stack trace or something?

samples/drawing/drawing.cpp
vadz: I think the current code is better because it's not misleading: people looking at the sample could believe that using alpha works under MSW when, as you say, it's just ignored there.

src/msw/overlay.cpp
AliKet: It crashes because `CreateWindowEx()` refuses to create the window if `lpClassName` is **null** and the error message returned is **invalid parameter value**. I don't know why, but I also don't mind passing a non-null `lpClassName` if that makes `CreateWindowEx()` happy in the end.

vadz: Yes, sure, this makes sense, you need to pass a valid class name to `CreateWindow()`.

The question is why do we need a specific class name and can't just use `wxWindowNR` here (this is returned by `wxWindow::GetMSWClassName()`)?

AliKet: I tried to do this but whenever the overlay window is shown for the first time there is a noticeable blinking on the window underneath (may be due to a window refresh ?) which is definitely undesirable/unacceptable.

vadz:
I've finally returned to this, sorry for the long delay, but after some testing I see that the main window loses activation when selecting now, which wasn't the case before. This is probably not fatal, but it would be still nice to avoid it, so I'm going to look at this -- but please let me know if you already know how to solve this problem (or know why it can't be solved). Thanks!

vadz:
> I've finally returned to this, sorry for the long delay, but after some testing I see that the main window loses activation when selecting now, which wasn't the case before.

OK, I've found the reason for this: we shouldn't use `WS_VISIBLE` when creating the window, but show it later, as this correctly uses `SWP_NOACTIVATE` and prevents it from stealing the activation from the parent.

vadz:
I've rebased this PR on master and simplified its code without -- AFAICS -- any loss in functionality. Please test the #23261 if you can and let me know if you see anything wrong with my changes to your code.

I'm closing this one now as it's included in the other PR.

Thanks again for implementing this!

Issue #23055, opened 2022-12-19
Optimize TIFF images loading/saving from/to memory buffers


vadz:
While looking at #23048 I've realized that we're handling loading/saving TIFF images from/to memory in a very inefficient way, as we use TIFF read/write proc for reading the data instead of using map proc to just provide libtiff with the direct access to the buffer that we already have.

Implementing `wxTIFFMapProc()` should result in a nice performance improvement here, so if anybody is motivated by achieving it, please don't hesitate to submit PRs. And if you do, please add a couple of new benchmarks to `tests/benchmarks/image.cpp` measuring the performance of doing this to see how it is affected by the changes.

PR #23064, opened 2022-12-22
Implement wxTIFFMapProc and wxTIFFUnmapProc

Kumazuma:
see #23055

### DIFF
--- a/include/wx/stream.h
+++ b/include/wx/stream.h
@@ -585,6 +585,7 @@ protected:
 
     wxStreamBuffer *m_i_streambuf;
 
+    wxDECLARE_ABSTRACT_CLASS(wxBufferedInputStream);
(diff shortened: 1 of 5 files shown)


vadz:
Not sure why did this get closed, but, just to be clear, even though using memory mapped files might be a good idea too, I wanted to do something much simpler and provide libtiff with the direct access to an already existing memory buffer when using `wxMemoryStream` classes. This would require either using `wxDynamicCast` or, probably preferably, adding some kind of `Map()` virtual function to `wxStreamBase` itself, which could then be only implemented by these classes.

Kumazuma:
It's not faster enough, but slower. so I closed this pull request.  if it need to access wxMemoryStream's bufffer directly, I reopen pull request after removing code about file mapping.

PR #23068, opened 2022-12-23
Fix for GetPrinter logic bug in WinPrinter class

sz5000:
According to the documentation of GetPrinter, GetPrinter returns false if bufSize is 0 on input. Therefore we can not check for FALSE on the first call. The following if statement always evaluates to true and then false is returned:

```
if ( !::GetPrinter(m_hPrinter, level, nullptr, 0, &bufSize) || !bufSize )
   return false;
```

If bufSize is 0 in the input, we cannot test the return value of GetPrinter for FALSE.

### DIFF
--- a/src/msw/printdlg.cpp
+++ b/src/msw/printdlg.cpp
@@ -87,7 +87,8 @@ public:
             return false;
 
         DWORD bufSize = 0;
-        if ( !::GetPrinter(m_hPrinter, level, nullptr, 0, &bufSize) || !bufSize )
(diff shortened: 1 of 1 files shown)


sz5000:
No idea why all these checks fail. My change shouldn't affect it.

MaartenBent:
You modified the catch submodule, you can see it in the 'files changed' tab.
 Try `git submodule update` and (force) push again.

vadz:
Thanks for noticing and fixing this! Will merge soon.

PR #23074, opened 2022-12-24
Explain how to display text elements from SVG files

Blake-Madden:
I've tested this with a number of SVG files with text objects of various fonts and styles and it works quite well.


vadz:
Thanks, this indeed looks useful to mention. I'll merge it soon.

PR #23079, opened 2022-12-25
Disable use of composition for top most windows

### LINKED ISSUE #23076: wxRichToolTip broken by the use of `WS_EX_COMPOSITED` too
Using "Dialogs > Rich tip dialog..." menu of the dialogs sample and showing the tooltip for the button there shows black background now. Setting `wx_msw_window_no_composited=1` fixes this, so it's due to `WS_EX_COMPOSITED` too.

### LINKED ISSUE #23078: wxGenericNotificationMessage on Windows broken by the use of `WS_EX_COMPOSITED`
When using `wxGenericMessage` on Windows 10, the message window is briefly (under 1 sec) displayed correctly, but then all its parts (icon, title, message, and close button) are displayed just as black rectangles.
wx-generic-user-notification
Just before the message disappears, it is again very briefly displayed properly. Setting system environment variable `wx_msw_window_no_composited` to `1` makes the issue disappear, so it appears to be caused by `WS_EX_COMPOSITED`.

To Reproduce:
1. Run dialogs sample.
2. Click on menu Dialogs/User notifications/User Notification.
3. In the User Notification Test Dialog check "Use generic notifications".
4. Press Show button.
5. See error.

- wxWidgets version: GIT master
- OS and its version: Windows 10 21H2, OS build 19045.2364, Windows Feature Experience Pack 120.2212.4190.0
- Other: light mode, DPI scaling 125% on the primary (2560x1440) and 100% on the secondary (1680x1050) display.

vadz:
WS_EX_COMPOSITED seems to be incompatible with WS_EX_TOPMOST and breaks (re)painting of such windows.

Closes #23076.

Closes #23078.

----

This seems to fix the problem for me under both 7 and 10. It's weird because I couldn't find anything about these styles being incompatible anywhere, so maybe it's due to something we're doing, but I don't see what can it be and disabling `WS_EX_COMPOSITED` in these cases doesn't seem to be a big problem, so for now let's just do it.

### DIFF
--- a/src/msw/window.cpp
+++ b/src/msw/window.cpp
@@ -534,25 +534,42 @@ bool wxWindowMSW::CreateUsingMSWClass(const wxChar* classname,
     // Enable double buffering by default for all our own, i.e. not the ones
     // using native controls, classes.
     //
-    // Note that this function is not used for top-level windows, so we don't
-    // set this style for them, and also that setting it for children of
-    // windows that already have WS_EX_COMPOSITED set doesn't seem to have any
-    // bad effect as the style is just ignored in this case, so we don't bother
-    // checking it it's already set for the parent, even though we could.
-    if ( !classname )
+    // The loop here is a bogus one just to create a block that we can break
+    // from, it never executes more than once.
+    while ( !classname )
     {
+        // WS_EX_COMPOSITED seems to be incompatible with WS_EX_TOPMOST, so
+        // don't use it for:
+
+        // Popup windows that get created with this style themselves: this
+        // seems to work under Windows 10, but doesn't under Windows 7 and
+        // using WS_EX_COMPOSITED for these windows that are temporarily
+        // doesn't seem to be very useful anyhow, so don't bother testing for
+        // the OS version and just always disable it for them.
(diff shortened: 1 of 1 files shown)


src/msw/window.cpp
```diff
@@ -534,25 +534,42 @@ bool wxWindowMSW::CreateUsingMSWClass(const wxChar* classname,
...
+
+        // Children of such windows as this doesn't work neither (see #23078).
+        wxWindow* const tlw = wxGetTopLevelParent(this);
+        if ( tlw && tlw->HasFlag(wxSTAY_ON_TOP) )
+            break;
+
         // We also allow disabling the use of this style globally by setting
         // a system option if nothing else (i.e. turning it off for individual
         // windows) works.
         if ( !wxSystemOptions::GetOptionInt("msw.window.no-composited") )
```
MaartenBent, line 560: This should be inverted. Now you break when the options is not set, so `WS_EX_COMPOSITED` is never enabled.

vadz: Oooops. And I thought I was hallucinating that I was seeing flicker again in the samples.

Thanks a lot for noticing this!

PR #23090, opened 2022-12-28
Fix wrong links in documentation

tohammer:
Using the same link id more than once a page results in all links using the first one.
See e.g. link to NanoSVG here
https://docs.wxwidgets.org/trunk/overview_high_dpi.html

Also add a missing link target

### DIFF
--- a/docs/doxygen/overviews/high_dpi.md
+++ b/docs/doxygen/overviews/high_dpi.md
@@ -270,11 +270,11 @@ bitmap will be the default bundle size, which must be provided when creating
 this kind of bitmap bundle, as SVG image itself doesn't necessarily contain
 this information.
 
-Note that wxWidgets currently uses [NanoSVG][1] library for SVG support and so
+Note that wxWidgets currently uses [NanoSVG][2] library for SVG support and so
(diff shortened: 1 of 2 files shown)


tierra:
Rather than keeping the numeric IDs, can you change these to named IDs?

For example:

```markdown
vector image in [SVG format] instead.

[SVG format]: https://en.wikipedia.org/wiki/Scalable_Vector_Graphics
```

```markdown
Note that wxWidgets currently uses [NanoSVG] library for SVG support and so
doesn't support all SVG standard features and you may need to simplify or
tweak the SVG files to make them appear correctly.

[NanoSVG]: https://github.com/memononen/nanosvg
```

This would avoid this situation happening again with improperly indexed links.

vadz:
Thanks for the fixes but I agree that we should just stop using numbered links, this is convenient for email, but not for the documentation.

But I think the best solution is the simplest one and so I've finally changed the docs to use "implicit links", i.e. `[link text][]` with `[link text]: https://link.url/` defined somewhere else, so I've done this now. Please let me know if anybody sees any problem with this.

PR #23093, opened 2022-12-28
Include arch-specific suffix in the DLLs names in makefile.vc too

vadz:
This is similar to edcfeab29d (Include arch-specific suffix in the DLL names in CMake builds too, 2022-12-23).

See #23053.


PBfordev:
So, we also need to change this sentence in `docs/contributing/about-platform-toolkit-and-library-names.md`
>$arch is used only by MSVS solutions. It is empty for 32-bit builds and 'x64' for 64-bit builds. It is only added to shared libraries.

to something like 
>$arch is used only for shared libraries on Windows when building with MSVC. It is empty for 32-bit builds and 'x64' for 64-bit builds.

Writing the above made me realize that the recent CMake change introduced a discrepancy between CMake and GCC makefile: Unlike with makefile.gcc, DLLs built with CMake will have "x64" in their names for all three compilers (MSVC, GCC, and clang). I wonder if we should change the CMakefile again, so $arch is used only with MSVC...

vadz:
I actually thought `WIN32_MSVC_NAMING` was only true when using MSVC but now I see that it's actually set to 1 for any compiler when not using Cygwin or MSYS... So I agree that we should test for `MSVC` here instead and updated this PR to do it.

@MaartenBent please let me know if you have any objections.

MaartenBent:
`WIN32_MSVC_NAMING` is used to differentiate between the way `msvc/makfile` creates/names files, and the way `configure` does it.
If only MSVC uses the arch suffix, then this fix is correct.

vadz:
Yes, this confirms what I (now) think: "MSVC naming" is actually "non-configure naming". FWIW I find this symbol name rather confusing, but I'm not sure if it's worth renaming it...

PR #23097, opened 2022-12-28
Improve support for dark mode in wxSTC

vadz:


vadz:
cc @eranif @imciner2

src/stc/stc.cpp
```diff
@@ -232,6 +232,22 @@ bool wxStyledTextCtrl::Create(wxWindow *parent,
     SetFontQuality(wxSTC_EFF_QUALITY_DEFAULT);
 #endif
 
+    // Use colours appropriate for the current system colour theme.
+    auto attr = wxTextCtrl::GetClassDefaultAttributes();
+    StyleSetForeground(wxSTC_STYLE_DEFAULT, attr.colFg);
```
eranif, line 237: I would do this a bit different:

```c++
for(int i = 0; i < wxSTC_STYLE_MAX; ++i) {
   ...
}
```

this is how I have been resetting styles in wxSTC for years now and it ensures me that I won't forget any style (like you special handled `wxSTC_STYLE_LINENUMBER` or `wxSTC_STYLE_CALLTIP` below)

vadz: Sorry, I prefer to use `StyleClearAll()` which seems to be provided exactly to do this instead of writing the loop ourselves. It's unfortunate that this function then goes to overwrite `STYLE_LINENUMBER` and `STYLE_CALLTIP`, but this hasn't been changed since many years and seems unlikely to change in the future...

vadz:
If there are no other comments I'm going to merge this soon.

PR #23100, opened 2022-12-29
Fix a possible crash in Mac menu item event handling

vadz:
This replaces #23040 with a better/simpler fix and also cleans things up a little as described there.

AFAICS this could be backported to 3.2 too.


csomor (approved):
Hi @vadz, I'm fine with it, thanks a lot, since the clickedAction does not have an alternative path now, we can write it like this. And if we don't encounter any problems, then I'm for backporting this of couse.

PR #23110, opened 2023-01-03
Mention embedded images in SVG not being supported

Blake-Madden:
At least from my experiments, PNG and BMP files embedded in SVGs do not get rasterized at all by nanosvg.  I only have Inkscape to test with, so perhaps it's something with how Inkscape saves SVG, but I more strongly suspect nanosvg just doesn't support this.  Looking at the code, that seems to be the case. However, please let me know if I am mistaken.

Also, fix minor doxygen formatting in note.

### DIFF
--- a/interface/wx/bmpbndl.h
+++ b/interface/wx/bmpbndl.h
@@ -252,6 +252,7 @@ public:
 
         - Text elements are not supported at all (see note for workaround).
         - SVG 1.1 filters are not supported.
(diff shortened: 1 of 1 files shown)


oneeyeman1:
@Blake-Madden ,
I believe it is nanoSVG's fault.

vadz:
Thanks, I'll apply this soon.

PR #23113, opened 2023-01-05
Restore value of errno after wxString::ToInt()/Long()/etc.

lanurmi:
Such functions modifying `errno` is undocumented, and may come as a surprise to the caller. Consequently also a call to wxString::Format() would modify `errno`.

Steps to reproduce (tested on macOS):
```
@@ -120,6 +120,10 @@ bool MyApp::OnInit()
     if ( !wxApp::OnInit() )
         return false;

+    fopen("asdfg", "r");
+    wxString::Format("aaaaaa");
+    wxLogMessage(wxSysErrorMsgStr());
+
     // create the main application window
     MyFrame *frame = new MyFrame("Minimal wxWidgets App");
```

This code would produce a log message `Unspecified error: 0` or similar. If the line with `wxString::Format` is commented out, the log message becomes `No such file or file or directory`.

The current behavior is not wrong either per se, and the error code/message should have been read as soon as possible, which is not what this code sample does. But many would not think of `wxString::Format` as the kind of call that would cause a system error on failure, nor cause the system error to be reset on a successful call.

### DIFF
--- a/src/common/string.cpp
+++ b/src/common/string.cpp
@@ -1446,6 +1446,7 @@ int wxString::Find(wxUniChar ch, bool bFromEnd) const
 
 #define WX_STRING_TO_X_TYPE_START                                           \
     wxCHECK_MSG( pVal, false, wxT("null output pointer") );                  \
+    int errnoWas = errno;                                                   \
     errno = 0;                                                              \
     const wxStringCharType *start = wx_str();                               \
     wxStringCharType *end;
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for the update! I've added cleanup that I mentioned in my initial comment on top of your changes, please let me know if you see any problems with that. If not, I'll merge it soon.

PR #23116, opened 2023-01-05
Improve comments in samples/sample.rc

PBfordev:
If accepted, could/should be backported to 3.2 branch.


vadz:
Thanks for updating this!

PR #23117, opened 2023-01-05
Update Scintilla to 5.0

### LINKED ISSUE #23080: [wxMSW] regression in wxSTC performance
Recently, I have been encouraged to try wxWidgets from master.
I noticed a **major** performance degradation:

- When I am loading CodeLite with couple of editors opened ("last session") it takes CodeLite roughly 10 seconds to load (the more open editors, the longer it takes to load)
- Reverting back to wx3.2 and CodeLite is loaded under 1 second

I spent hours doing `git bisect`  mainly because some of the offending commits selected by `git bisect bad` or `git bisect good` - can not be built.

`git bisect` output is:

```
# git bisect bad
6c3b0f66f31a78f4b070ca75bedb3ccd7f105f21 is the first bad commit
commit 6c3b0f66f31a78f4b070ca75bedb3ccd7f105f21
Author: Maarten Bent <MaartenBent@users.noreply.github.com>
Date:   Tue Jul 28 22:57:04 2020 +0200

    Update scintilla to 3.21.1
```

But this commit can not be built, the one after it is the actual offending commit
2af3719a4eccff2bf12fc70087b29f1f7ef63e25

OS: `Windows 10, 64 bit`
Compiler: `Clang 15.02/MSYS2`

CC: @MaartenBent 

Thanks!

MaartenBent:
This PR updates Scintilla to 5.0.0, and adds Lexilla 5.0.1.

I replaced c++17 features, so it can be built with c++14 compilers. And another commit to replace c++14 features with c++11, if this will be the maximum c++ version that wxWidgets will support.

This might be the moment to move the Scintilla and Lexilla sources to submodules. Lexilla's upstream is on GitHub. But Scintilla is using Mercurial at Source Forge, so we would have to maintain that ourselves.

@eranif This should also fix #23080

`SCI_SETLEXER`, `SCI_SETLEXERLANGUAGE` and `SCI_LOADLEXERLIBRARY` have been removed from Scintilla, I reimplemented them using the new API

The xcode and msvc projects have not yet been updated.


samples/stc/stctest.cpp
```diff
@@ -672,7 +672,8 @@ AppAbout::AppAbout (wxWindow *parent,
     }
 
     // Get version of Scintilla
-    wxVersionInfo vi = wxStyledTextCtrl::GetLibraryVersionInfo();
+    wxVersionInfo vi1 = wxStyledTextCtrl::GetLibraryVersionInfo();
+    wxVersionInfo vi2 = wxStyledTextCtrl::GetLexerVersionInfo();
```
vadz, line 676: I think we should make the names more readable.
```suggestion
    wxVersionInfo viScintilla = wxStyledTextCtrl::GetLibraryVersionInfo();
    wxVersionInfo viLexilla = wxStyledTextCtrl::GetLexerVersionInfo();
```

src/stc/PlatWX.cpp
```diff
@@ -607,6 +611,11 @@ void SurfaceImpl::Copy(PRectangle rc, Point from, Surface &surfaceSource) {
               wxRound(from.x), wxRound(from.y), wxCOPY);
 }
 
+std::unique_ptr<IScreenLineLayout> SurfaceImpl::Layout(const IScreenLine* WXUNUSED(screenLine))
+{
+    return nullptr;
```
vadz, line 616:
```suggestion
    return {};
```

src/stc/PlatWX.cpp
```diff
@@ -1593,7 +1614,7 @@ void SurfaceD2D::DrawTextTransparent(PRectangle rc, Font &font_,
 XYPOSITION SurfaceD2D::WidthText(Font &font_, const char *s, int len)
 {
     XYPOSITION width = 1.0;
-    wxString tbuf = stc2wx(s,len);
+    wxString tbuf = stc2wx(s, len);
```
vadz, line 1617: These whitespace-only changes are a bit annoying :-(

src/stc/PlatWX.cpp
```diff
@@ -1550,6 +1566,11 @@ void SurfaceD2D::Copy(PRectangle rc, Point from, Surface& surfaceSource)
     }
 }
 
+std::unique_ptr<IScreenLineLayout> SurfaceD2D::Layout(const IScreenLine* WXUNUSED(screenLine))
+{
+    return nullptr;
```
vadz, line 1571:
```suggestion
    return {};
```

src/stc/ScintillaWX.cpp
```diff
@@ -206,6 +207,8 @@ static wxTextFileType wxConvertEOLMode(int scintillaMode)
 
 
 ScintillaWX::ScintillaWX(wxStyledTextCtrl* win) {
+    GetLexerCount();
```
vadz, line 210: It would be nice to have a comment explaining why do we call this function and don't use its result (I guess it's some kind of lexer initialization?).

vadz (commented):
Thanks for the update!

Are there any new features in the new version that would merit being showcases in the sample? I didn't really see anything in https://www.scintilla.org/ScintillaHistory.html but it's not exactly easy to read, so I could have missed it...

As for the submodules, we'd indeed need to maintain our own Git mirror, but it could still be useful to have it in order to track our changes to the upstream sources more easily. In principle, converting from hg to git is simple enough with fast export, but I must have last done it 10+ years ago, when I was still using hg, so I am not totally sure if this is still the case... If you'd like, I could try doing the conversion but we need to decide if we want to use submodules at all first and I think this should be your choice, as you're the one working with STC the most, so please let me know.

src/stc/ScintillaWX.cpp
MaartenBent: I mentioned it in the commit message. There is a bug in Lexilla 5.0.2 where it doesn't initialize the lexers. It is fixed in 5.0.3  but this version has major architecture changes so I can't/won't update to that version.

I think I'll remove this and apply the Lexilla fix instead.

MaartenBent:
Nothing much has changed. 4.4 is mostly the same as 3.21, but with c++17. And 5.0 moves the lexers to a separate project.

I'd like to use submodules. Scintilla is the only external code that doesn't have one yet.
Updating to the next Scintilla version (even 5.0.1) seems to requires some major changes. So I'm not sure when/if I or someone else will do that. We might add the submodules and then never update them. But I still think it is better to have them so they are there when we need them.

src/stc/ScintillaWX.cpp
vadz: Sorry for missing this and thanks for the explanation!

vadz:
I didn't realize that upgrading from 5.0 to 5.0.1 would be more difficult than upgrading from 3 to 5, how quaint...

Anyhow, let's use submodules, I can create the repository (or 2 repositories?) under wxWidgets org for them but if you plan to do it yourself, I think it might actually be simpler/more convenient if you created your own repositories and then moved them under wx org (I'm almost sure you should be able to do it). As long as GitHub Actions support cron-based workflows, we could set one up to update the mirror from hg automatically. Or we could set up a "dispatch" workflow to only do it on demand. Please let me know if you're going to set this up or if you count on me to do it.

MaartenBent:
This should be (almost) done now, only need to sort out the submodules.
I found and used this scintilla mirror: https://github.com/missdeer/scintilla/tree/master. If you think setting up our own automatic hg mirror will be more future proof, you will have to do that. I'm not sure how to do that.

I'll try to move my submodules to the wxWidgets organization. You probably get an email to accept or deny this.

There is no rush, if you want to focus on 3.2.2 first, thats fine.

MaartenBent:
Didn't work: You don’t have the permission to create public repositories on wxWidgets

Maybe you can fork them yourself via the GitHub website (so it will also show `forked from origina/repo`). And create a `wx` branch for both of them.
And then set the `wx` branch as the default branch (might want to do this for the `nanosvg` and `pcre` repositories too).

For the `scintilla` repo, set branch wx to commit `78f398b2eb4bce342197a2c3559ed561837224dd`
For the `lexilla` repo, set branch wx to tag `rel-5-0-1`

vadz:
I think it's because you're not a wx org member? I could add you as one, of course, please let me know if I should.

Otherwise I can, of course, create the forks myself but I'm not sure about the lexilla one: where should I fork it from, your repository with the same name?

MaartenBent:
You can fork the official Lexilla repo: https://github.com/ScintillaOrg/lexilla

vadz:
I've created both repositories now with the branches set as described, please let me know if you see anything wrong, TIA!

MaartenBent:
Looks good, just one small thing. Can you go to the Scintilla/Lexilla project settings, branches, and then set `wx` as the default branch? This makes it easier to find when navigating the repo on GitHub.
Most other submodules have this already, except nanosvg and pcre. Maybe you can change this for them too?

vadz:
No idea why, but I'm trying "Could not change default branch" when trying to do it for lexilla... I'll look at this again tomorrow.

vadz:
> No idea why, but I'm trying "Could not change default branch" when trying to do it for lexilla

Doing exactly the same thing from the desktop browser worked, so I've updated all these repositories to use wx now.

MaartenBent:
I've created https://github.com/wxWidgets/scintilla/pull/1 and https://github.com/wxWidgets/lexilla/pull/1

vadz:
Thanks, merged both of them (using GitHub web UI, hopefully this worked as I thought it would...).

.github/workflows/code_checks.yml
```diff
@@ -90,6 +91,7 @@ jobs:
             if git diff origin/master \
                ':!.github/workflows/code_checks.yml' \
                ':!src/stc/scintilla/' \
+               ':!src/stc/lexilla/' \
```
vadz, line 94: BTW this shouldn't be necessary any longer as git-diff doesn't recurse into submodules by default.

src/stc/ScintillaWX.cpp
```diff
@@ -771,35 +774,58 @@ sptr_t ScintillaWX::WndProc(unsigned int iMessage, uptr_t wParam, sptr_t lParam)
             break;
 #endif
 
-      case SCI_GETDIRECTFUNCTION:
+        case SCI_GETDIRECTFUNCTION:
```
vadz, line 777: Are these whitespace changes intentional?

vadz (approved):
Thanks again for all this work!

Should this be merged if/when the CI builds pass or do you plan more changes here?

src/stc/ScintillaWX.cpp
MaartenBent: Yes, I fixed the indentation in this commit a968b37

.github/workflows/code_checks.yml
MaartenBent: Ok, I'll remove them and see if CI succeeds..

MaartenBent:
> Should this be merged if/when the CI builds pass or do you plan more changes here?

Yes, it is done now (after resolving above comment). I'll remove the draft, and then it should be ready to merge.

vadz:
I'd like to ask people using wxSTC in their applications to test this PR and let us know if there are any important regressions (there are not supposed to be any, but it doesn't mean there are none). It would be much better to fix them before merging it!

vadz:
Thanks again for doing all this work!

I was wondering if you could summarize the main new features of Scintilla 5.0 so that we could include this in the change log and/or maybe a separate news/blog post on the web site if it's worth it. In the latter case and if you'd like to do it, please feel free to just submit one as PR to the website repository, otherwise just a few lines here would be enough. TIA!

MaartenBent:
The main change of 5.0 is the division of the project into Scintilla and Lexilla.

I think the only other big change is the addition of `SetILexer()`/`SCI_SETILEXER`, see https://www.scintilla.org/Scintilla5Migration.html .
It replaces `SetLexer()` / `SCI_SETLEXER`, `SetLexerLanguage()` / `SCI_SETLEXERLANGUAGE` and `LoadLexerLibrary()` / `SCI_LOADLEXERLIBRARY`. This doesn't affect users because I restored their functionality in a72108c657.
When we update again, Scintilla 5.0.1 will also remove those defines. But we can probably keep defining them ourselves as long as Lexilla keeps the API functions.

It is probably not worth a blog post. I suspect the biggest problem will be complaints about missing files because users forget the update the submodules.

vadz:
But we upgraded it from Scintilla 3, have there been no user-visible changes in all the intermediate versions neither?

MaartenBent:
Scintilla 4.4 is the same as Scintilla 3.21, but with c++17. And Scintilla 5.0 is the next version after that.

If I check `stc.h`, the only other difference is support for a F# lexer.

vadz:
I see, thanks. I admit I expected more changes from such a version jump, especially because I remember people really wanting to upgrade to the later versions, but it's still very nice to not be stuck on a really ancient version, of course.

Eric-01:
> I'd like to ask people using wxSTC in their applications to test this PR and let us know if there are any important regressions (there are not supposed to be any, but it doesn't mean there are none). It would be much better to fix them before merging it!

@eranif, you may be interested in this as your CodeLite IDE makes extensive use of wxSTC and wxWidgets 3.3.

eranif:
Already built and updated for macOS  & Windows. Will update minimum Ubuntu tomorrow

PR #23118, opened 2023-01-05
Enable `-Werror` for clang wxMSW CI build

vadz:
@MaartenBent Is this a good way of enabling `-Werror` for CMake builds?


MaartenBent:
Yes, this seems the easiest way to enable it.

Issue #23114, opened 2023-01-05
wxOSX does not fall back properly when setting default language


sethhillbrand:
### Description

When MacOS system is set to use English as its language and a region that is not provided by the locale database (e.g. Germany), calling `wxLocale::Init( wxLANGUAGE_DEFAULT )` will attempt to set the language to `en_DE` even though this language does not exist.

### Expected behavior
According to the documentation, 
 
        wxLANGUAGE_DEFAULT has special meaning: best suitable translation,
        given user's preference and available translations, will be used.

I would expect that if the language/region combination does not exist that the first available matching language would be used regardless of region.  If no matching language is available, I would expect that the base translation (usually english) would be used.

I would never expect that we would see an error message when setting the locale to `wxLANGUAGE_DEFAULT`

#### To Reproduce
Set language and region like the following:
![image](https://user-images.githubusercontent.com/914826/210843671-d3e3b12a-5673-4ac7-8e75-4d1e0df25321.png)

Then run intltest.cpp and set the language to `wxLANGUAGE_DEFAULT`

### Platform and version information

- wxWidgets version  you use: wxWidgets: 3.2.1 (wchar_t,wx containers)
- wxWidgets port  you use: wxOSX
- OS  and its version: macOS Ventura Version 13.1 (Build 22C65)

vadz:
FWIW I think it does make sense to _try_ `en_DE` on such system, but we should definitely fall back on `en_US` if setting it fails. I'll check why it doesn't happen currently...

sethhillbrand:
Agreed.  Only if the language/region combination does not exist should it choose a different one

vadz:
*TL;DR*: There is a minor problem in `wxLocale` but I think the real solution is to just stop using it in your code and use `wxUILocale`, it was added partially in order to make dealing with such less standard situations simpler and more logical.

----

Thinking more about this, I think we definitely need to change one thing here:

- We must handle `IsAvailable(wxLANGUAGE_DEFAULT)` specially as using the default locale may succeed even if we don't recognize it.

However I'm less sure if we should do this:

- Try to chop off the part after `_` in the locale name and retry using it as `wxSetlocale()` argument if `wxSetlocale(LC_ALL, "")` fails.

[...]

PR #23119, opened 2023-01-06
Minor wxLocale-related improvements

vadz:
See #23114.

cc @utelle


utelle (approved):
I'm fine with the applied changes.

Of course, it makes sense to check whether `setlocale` actually works for the default locale, even if no entry for it exists in the wx language database (typically being the case for non-standard locales like `en_DE`).

vadz:
This is now part of #23226,

PR #23123, opened 2023-01-07
Improve manifest docs

vadz:
An attempt to make things even more clear.


docs/msw/install.md
```diff
@@ -478,12 +476,41 @@ instructions here are out of date, you can always simply copy a makefile or
...
-If you are not using Visual Studio please see
-@subpage plat_msw_winxp "Theme Support" to enable visual styles in your
-application.
+
+Using a Manifest                       {#msw_manifest}
+----------------
+
+All Windows applications should use a "manifest", which is a special kind of
+Windows resource with information about the application, as this is required to
+use the "modern", i.e. Windows XP, version 6 of the Common Controls Library.
```
PBfordev, line 485: It may be just me, but I am not a big fan of mentioning Windows XP here. They are not relevant here and some users may not even ever used them.

I would use something like this instead
>... information about the application compatibility, required, among else, for the application UI to look correctly instead of looking very outdated and different from other native applications.

vadz: What about

> use the "modern", i.e. not "classic", appearance provided by the version 6 of the Common Controls Library

? Or do you think that mentioning "classic" is still bad?

PBfordev: > ? Or do you think that mentioning "classic" is still bad?

I just wanted avoid terminology that may be not only outdated but in 2023 also unknown to many people, who unlike us, may have never used older Windows versions.

vadz: Sorry, do you mean "XP" or "classic" is outdated/unknown? I guess both of them are, but I have no idea about how to express what I want to say here without mentioning either of them. We could just say "All modern Windows applications must use a manifest...", but this would be less precise.

PBfordev: > Sorry, do you mean "XP" or "classic" is outdated/unknown? I guess both of them are, but I have no idea about how to express what I want to say here without mentioning either of them. 

FWIW, I provided such a suggestion in my first comment (as a manifest is not only about CC version). 

BTW, the PR should also update the comment in `samples/sample.rc`, which refers to the removed `winxp.md`.

vadz: Thanks for noticing its use in `sample.rc`, I've updated it now and also used your wording with an extra mention of comctl32.dll (because I think it's still important to say it), hopefully it shouldn't seem so antiquated now. Or does it, yet?

vadz: @PBfordev I'd just like to ping you again for one last look before I merge this, please let me know if you think this still needs to be improved.

Issue #23122, opened 2023-01-07
wxrc.exe is missing from the msvc binary dev packages


samizzo:
### Description
I was expecting to find wxrc.exe in the msvc dev packages. Issue #18124 requested this to be added, and it was closed with a fix, but wxrc.exe does not appear to be in the msvc dev packages. I downloaded `wxMSW-3.2.1_vc14x_x64_Dev.7z` and it was not there. I also downloaded a few earlier versions and it didn't seem to be in any of them. It is however in `wxMSW-3.2.1_gcc1220_x64_Dev.7z`

PBfordev:
I have checked development packages for the latest latest MSVC and GCC versions:
```
wxMSW-3.2.1_gcc1220_Dev.7z
wxMSW-3.2.1_gcc1220_x64_Dev.7z
wxMSW-3.2.1_vc14x_x64_Dev.7z
wxMSW-3.2.1_vc14x_Dev.7z
```
wxrc.exe is missing "only" from the 64-bit MSVC one. 

Not sure where the problem is, assuming the executable was really built it `utils\wxrc\%VCver%_mswudll_x64\` it should be copied and packed here
https://github.com/wxWidgets/wxWidgets/blob/f0b76076f4b8ca21ef5acebd5569f41133dcd21b/build/tools/msvs/package.bat#L70-L73

However, when I build 64-bit wxrc with `nmake -f makefile.vc BUILD=release SHARED=1`, the result is in "vc_x64_mswudll", not in "vc_mswudll_x64" (please notice the different position of "mswudll").

vadz:
Thanks for reporting and checking this!

@doscott Could you please have a look at it? I think it might be just as simple as fixing the path used in the copy command, but it would be nice if you could please confirm it. TIA!

doscott:
Will look at this tomorrow.

samizzo:
Thank you! Sorry I should have checked the other zips as well.

PR #23125, opened 2023-01-08
Corrected path for x64 wxrc.exe file

doscott:
Fixes issue #23122. This should be backported to 3.2. I don't know a quick way to do this other than making the change on the 3.2 branch and making another PR, which I am quite willing to do, just let me know.


vadz:
Thanks Danny! Committed to master and cherry-picked to 3.2 branch now.

Issue #23102, opened 2022-12-30
wxDataViewCtrl with wxDV_VARIABLE_LINE_HEIGHT performs incoherent scrolling


CarCan69:
### Description
The controls based on wxDataViewCtrl with wxDV_VARIABLE_LINE_HEIGHT performs incoherent scrolling when scrolling is made with keyboard. 

### Source to reproduce the issue
Build and execute, push "End" button, the list scroll at the and and last item is selected, push "Up" arrow key and when the selected item is "35" the list does not scroll anymore and the selection "climb" the list, it is like when I reach the top of the list. If you contine to play with arrows keys other incoherent scroll are performed.

```
#include "wx/wxprec.h"

#ifndef WX_PRECOMP
#include "wx/wx.h"
#endif

#include <wx/dataview.h>
#include <wx/renderer.h>


class RigaScontrino
{
public:
    RigaScontrino(const wxString& articolo, long quantita, double prezzo, double sconto, double ribasso)
    {
        m_articolo = articolo;
        m_quantita = quantita;
        m_prezzo = prezzo;
        m_sconto = sconto;
        m_ribasso = ribasso;
    }

    ~RigaScontrino()
    {
    }

    bool IsDiminuzione() { return m_sconto != 0 || m_ribasso != 0; }

    double GetTotale() { return m_quantita * m_prezzo; }
    double GetDiminuzione() { return m_sconto == 0 ? -m_ribasso : -GetTotale() * m_sconto / 100; }

    wxString GetTestoDiminuzione() { return m_sconto == 0 ? wxString::Format("Ribasso %.2f", GetDiminuzione()) : wxString::Format("Sconto %.0f%% di %.2f", m_sconto, GetTotale()); }

    wxString m_articolo;
    long m_quantita{ 0 };
    double m_prezzo{ 0 };
    double m_sconto{ 0 };
    double m_ribasso{ 0 };
};



class ScontrinoRender : public wxDataViewCustomRenderer
{
public:
    explicit ScontrinoRender(wxDataViewCellMode mode)
        : wxDataViewCustomRenderer("void*", mode, wxALIGN_LEFT)
    { }


    bool Render(wxRect rect, wxDC* dc, int state) wxOVERRIDE
    {
        int rectHeight{ m_value->IsDiminuzione() ? rect.height / 2 : rect.height };
        int fontHeight = rectHeight * .8;

        if (fontHeight == 0)
            return false;

        wxRect articoloRect;
[...]

PR #23128, opened 2023-01-10
Scrolling with enabled flag wxDV_VARIABLE_LINE_HEIGHT (#23102)

jensgoe:
scroll to last visible item must be calculated based on item position instead of the first visible row.

### DIFF
--- a/src/generic/datavgen.cpp
+++ b/src/generic/datavgen.cpp
@@ -825,7 +825,11 @@ public:
 
     int GetCountPerPage() const;
     int GetEndOfLastCol() const;
+
+    // Returns the position where the given column starts.
+    // The column must be valid.
     int GetColumnStart(int column) const;
+
     unsigned int GetFirstVisibleRow() const;
     wxDataViewItem GetTopItem() const;
 
@@ -3474,9 +3478,9 @@ void wxDataViewMainWindow::ScrollTo( int rows, int column )
     // Take care to not divide by 0 if we're somehow called before scrolling
     // parameters are initialized.
     int sy = y ? GetLineStart( rows )/y : -1;
-    int sx = GetColumnStart(column);
-    if( column != -1 && x )
-        sx = sx / x;
+    int sx = -1;
+    if (column != -1 && x)
+        sx = GetColumnStart(column) / x;
     m_owner->Scroll( sx, sy );
 }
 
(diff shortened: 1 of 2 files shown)


src/generic/datavgen.cpp
```diff
@@ -6450,9 +6450,23 @@ void wxDataViewCtrl::EnsureVisibleRowCol( int row, int column )
...
+        else
+        {
+            // calculate scroll position based on last visible item
+            const int itemStart = m_clientArea->GetLineStart(row);
+            const int itemHeight = m_clientArea->GetLineHeight(row);
+            const int clientHeight = m_clientArea->GetSize().y;
+            int scrollX, scrollY;
+            GetScrollPixelsPerUnit(&scrollX, &scrollY);
+            int scrollPosY = (itemStart + itemHeight - clientHeight) / scrollY;
+            Scroll(-1, scrollPosY + 1); // +1 due to integer division, always show whole item
```
vadz, line 6466: If it's really about rounding up, shouldn't this rather be
```suggestion
            int scrollPosY = (itemStart + itemHeight - clientHeight + scrollY - 1) / scrollY;
            Scroll(-1, scrollPosY);
```
?

src/generic/datavgen.cpp
```diff
@@ -6450,9 +6450,23 @@ void wxDataViewCtrl::EnsureVisibleRowCol( int row, int column )
...
+        }
+    }
+    else if (row == first)
```
vadz, line 6469: We don't need this test, perhaps just leave it as a comment but testing for it is IMO confusing as it seems to indicate that the test might be false -- which is never the case.
```suggestion
    else // row == first
```

vadz (commented):
Thanks!

Should we merge this already (maybe after applying the fixes below) or do you plan to add more commits?

src/generic/datavgen.cpp
jensgoe: I try to explain. Currently the else case scolls always to first row even if row is somewhere between fist and last row. for variable line height this results in the following behaviour (gif):
scroll_up
Because of GetFirstVisibleRow returns the first not fully visible row and ScrollTo(newRow) makes newRow fully visible. Maybe that is already a problem. Indeed it is not necessary (for fixed and variable line height) to scroll if the row is already between the visible rows so i added this test.
But yes the code could look cleaner with:
`
if( row <= first )
    m_clientArea->ScrollTo( row, column );
else if (row > last) {...}
`

src/generic/datavgen.cpp
```diff
@@ -6448,12 +6457,28 @@ void wxDataViewCtrl::EnsureVisibleRowCol( int row, int column )
...
+            // calculate scroll position based on last visible item
+            const int itemStart = m_clientArea->GetLineStart(row);
+            const int itemHeight = m_clientArea->GetLineHeight(row);
+            const int clientHeight = m_clientArea->GetSize().y;
+            int scrollX, scrollY;
+            GetScrollPixelsPerUnit(&scrollX, &scrollY);
+            int scrollPosY = (itemStart + itemHeight - clientHeight + scrollY - 1) / scrollY;
+            int scrollPosX = m_clientArea->GetColumnStart(column);
+            if (column != -1 && scrollX)
+                scrollPosX = scrollPosX / scrollX;
```
vadz, line 6477: Shouldn't we round up here too, i.e.
```suggestion
                scrollPosX = (scrollPosX + scrollX - 1)/ scrollX;
```
?

src/generic/datavgen.cpp
```diff
@@ -825,6 +825,7 @@ class wxDataViewMainWindow: public wxWindow
 
     int GetCountPerPage() const;
     int GetEndOfLastCol() const;
+    int GetColumnStart(int column) const;
```
vadz, line 828: It would be nice to add a comment here saying that this function can be called with `column == -1` and just returns `-1` in this case.

Alternatively, maybe it would be more clear if it could be only called with a valid column (and then always returned a valid value) and then not call it below when column is not specified.

vadz (commented):
Thanks, we should merge this, but I still have a couple of minor remarks below.

jensgoe:
thanks for your feedback. i applied your suggested changes

vadz:
Thanks again for the fixes! I'm going to merge this and we should backport it to 3.2 too if no new problems are found.

PR #23131, opened 2023-01-12
Fix buttons in print preview frame in high DPI

### LINKED ISSUE #23127: wxPrintPreview to large icons on High DPI Display
Tested wxPrintPreview recently and noticed that the Icons in the toolbar are to large. Seems to b another High DPI issue. Screenshot is attached. Code is below, copied from an example.

image

```
wxPrintPreview* preview = new wxPrintPreview(printout1, printout2);
wxPreviewFrame* frame = new wxPreviewFrame(preview, GetCtrl(),
	"Demo Print Preview",
	wxPoint(100, 100),
	wxSize(600, 650));
frame->Centre(wxBOTH);
frame->Initialize();
frame->Show(true);
```

vadz:
Closes #23127.

### DIFF
--- a/src/common/artprov.cpp
+++ b/src/common/artprov.cpp
@@ -191,7 +191,30 @@ protected:
         // ones should just override CreateBitmapBundle() directly), so we only
         // return the original bitmap scale, but hope that perhaps the provider
         // will have other (e.g. x2) scales too, when our GetBitmap() is called.
-        return i++ ? 0.0 : m_bitmapScale;
+        //
+        // We also suppose that the art provider can always return a good (i.e.
+        // not obtained by rescaling) bitmap at scale 1, even if the original
+        // bitmap had a higher scale factor. This is necessary to allow using
+        // bitmaps on a secondary monitor using standard DPI when the primary
+        // monitor uses high DPI, as the art provider always uses the primary
+        // monitor DPI (being global, it can't do anything else).
+        switch ( i++ )
+        {
+            case 0:
+                // As explained above, this is always supported.
+                return 1.0;
+
+            case 1:
+                // If we already have a bitmap in a higher DPI, we can be sure
+                // that we can provide it too.
+                if ( m_bitmapScale != 1.0 )
+                    return m_bitmapScale;
+                wxFALLTHROUGH;
+
+            default:
+                // No other scales supported.
+                return 0.0;
+        }
     }
 
 private:
--- a/src/common/artstd.cpp
+++ b/src/common/artstd.cpp
@@ -232,7 +232,7 @@ wxDefaultArtProvider::CreateBitmapBundle(const wxArtID& id,
     // We currently handle just a single SVG here.
     if ( id == wxART_WX_LOGO )
     {
-        wxSize sizeDef = size != wxDefaultSize ? size : GetSizeHint(client);
+        wxSize sizeDef = size != wxDefaultSize ? size : GetDIPSizeHint(client);
         if ( sizeDef == wxDefaultSize )
         {
             // We really need some default size here.
--- a/src/common/arttango.cpp
+++ b/src/common/arttango.cpp
@@ -209,7 +209,7 @@ wxTangoArtProvider::CreateBitmapBundle(const wxArtID& id,
         if ( entry.id != id )
             continue;
 
-        wxSize sizeDef = size != wxDefaultSize ? size : GetSizeHint(client);
+        wxSize sizeDef = size != wxDefaultSize ? size : GetDIPSizeHint(client);
         if (sizeDef == wxDefaultSize)
         {
             // We really need some default size here, so keep using the same
--- a/src/common/prntbase.cpp
+++ b/src/common/prntbase.cpp
@@ -1517,8 +1517,8 @@ public:
         // We don't use (smaller) images inside a button with a text label but
         // rather toolbar-like bitmap buttons hence use wxART_TOOLBAR and not
         // wxART_BUTTON here.
-        wxBitmap bmp = wxArtProvider::GetBitmap(artId, wxART_TOOLBAR);
-        wxBitmapButton * const btn = new wxBitmapButton(m_parent, btnId, bmp);
+        wxBitmapBundle bb = wxArtProvider::GetBitmapBundle(artId, wxART_TOOLBAR);
+        wxBitmapButton * const btn = new wxBitmapButton(m_parent, btnId, bb);
         btn->SetToolTip(tooltip);
 
         Add(btn);
--- a/src/msw/artmsw.cpp
+++ b/src/msw/artmsw.cpp
@@ -276,14 +276,20 @@ wxSize wxArtProvider::GetNativeDIPSizeHint(const wxArtClient& client)
     }
     else if ( client == wxART_FRAME_ICON )
     {
+        // We're supposed to return a DPI-independent value here, but
+        // ::GetSystemMetrics() uses the primary monitor DPI, so undo this by
+        // explicitly dividing by its scale.
         return wxSize(::GetSystemMetrics(SM_CXSMICON),
-                      ::GetSystemMetrics(SM_CYSMICON));
+                      ::GetSystemMetrics(SM_CYSMICON))
+                / wxDisplay().GetScaleFactor();
     }
     else if ( client == wxART_CMN_DIALOG ||
               client == wxART_MESSAGE_BOX )
     {
+        // As above, we need to convert to DIPs explicitly.
         return wxSize(::GetSystemMetrics(SM_CXICON),
-                      ::GetSystemMetrics(SM_CYICON));
+                      ::GetSystemMetrics(SM_CYICON))
+                / wxDisplay().GetScaleFactor();
     }
     else if (client == wxART_BUTTON)
     {


vadz:
I'm almost certain that the first commit here is correct, but it's a bit surprising that nobody had noticed this until now because it affects _all_ uses of `wxArtProvider::GetBitmap()`, not just this one. But the current code just seems clearly wrong and I think this should be fixed in 3.2 too, even if this is not exactly a backwards-compatible change (the size of the returned bitmaps will change).

Any thoughts/comments?

MaartenBent:
I see also an issue with the `wxWindowsArtProvider`, e.g. when selecting `wxART_ERROR` in the artprov sample.
My main monitor has 144dpi (1.5 scale), and any bitmap returned from `wxArtProvider::GetBitmapBundle(wxART_ERROR, wxART_MESSAGE_BOX, wxDefaultSize)` uses this as minimum size.
For example, when I move the artprov sample to a 96dpi monitor, it returns an image of 72x72 instead of 48x48 .

This is caused by this line, where `m_bitmapScale` has value 1.5, and this is used as minimum for any bitmap that is returned:
https://github.com/wxWidgets/wxWidgets/blob/a4f939aafa8f51a6d0ad16b4634387500340976f/src/common/artprov.cpp#L194



Not sure how to fix this. Probably the best way is to implement `wxWindowsArtProvider::CreateBitmapBundle` which extracts the various windows icons at all the desired sizes (16x16 to ...).
Then this will succeed:
https://github.com/wxWidgets/wxWidgets/blob/a4f939aafa8f51a6d0ad16b4634387500340976f/src/common/artprov.cpp#L465
and the bitmapBundle is no longer created from the (1.5 scaled) bitmap here
https://github.com/wxWidgets/wxWidgets/blob/a4f939aafa8f51a6d0ad16b4634387500340976f/src/common/artprov.cpp#L484-L486

Modifications to the artprov sample while testing this:
(I am aware that the `size: ? x ?` text does not update when switching monitors, I only look at the icon size).
<details>
  <summary>Click me</summary>

```diff
diff --git "a/samples/artprov/artbrows.cpp" "b/samples/artprov/artbrows.cpp"
index e33319c348..39b6b10ee5 100644
--- "a/samples/artprov/artbrows.cpp"
+++ "b/samples/artprov/artbrows.cpp"
@@ -60,62 +60,6 @@ static void FillBitmaps(wxImageList *images, wxListCtrl *list,
                         const wxArtClient& client, const wxSize& size)
 {
     ART_ICON(wxART_ERROR)
-    ART_ICON(wxART_QUESTION)
-    ART_ICON(wxART_WARNING)
-    ART_ICON(wxART_INFORMATION)
-    ART_ICON(wxART_ADD_BOOKMARK)
-    ART_ICON(wxART_DEL_BOOKMARK)
-    ART_ICON(wxART_HELP_SIDE_PANEL)
-    ART_ICON(wxART_HELP_SETTINGS)
-    ART_ICON(wxART_HELP_BOOK)
-    ART_ICON(wxART_HELP_FOLDER)
-    ART_ICON(wxART_HELP_PAGE)
-    ART_ICON(wxART_GO_BACK)
-    ART_ICON(wxART_GO_FORWARD)
-    ART_ICON(wxART_GO_UP)
-    ART_ICON(wxART_GO_DOWN)
-    ART_ICON(wxART_GO_TO_PARENT)
-    ART_ICON(wxART_GO_HOME)
-    ART_ICON(wxART_GOTO_FIRST)
-    ART_ICON(wxART_GOTO_LAST)
-    ART_ICON(wxART_PRINT)
-    ART_ICON(wxART_HELP)
-    ART_ICON(wxART_TIP)
-    ART_ICON(wxART_REPORT_VIEW)
-    ART_ICON(wxART_LIST_VIEW)
-    ART_ICON(wxART_NEW_DIR)
-    ART_ICON(wxART_FOLDER)
-    ART_ICON(wxART_FOLDER_OPEN);
-    ART_ICON(wxART_GO_DIR_UP)
-    ART_ICON(wxART_EXECUTABLE_FILE)
-    ART_ICON(wxART_NORMAL_FILE)
-    ART_ICON(wxART_TICK_MARK)
-    ART_ICON(wxART_CROSS_MARK)
-    ART_ICON(wxART_MISSING_IMAGE)
-    ART_ICON(wxART_NEW);
-    ART_ICON(wxART_FILE_OPEN)
-    ART_ICON(wxART_FILE_SAVE)
-    ART_ICON(wxART_FILE_SAVE_AS)
-    ART_ICON(wxART_DELETE);
-    ART_ICON(wxART_COPY)
-    ART_ICON(wxART_CUT)
-    ART_ICON(wxART_PASTE)
-    ART_ICON(wxART_UNDO)
-    ART_ICON(wxART_REDO)
-    ART_ICON(wxART_PLUS)
-    ART_ICON(wxART_MINUS)
-    ART_ICON(wxART_CLOSE)
-    ART_ICON(wxART_QUIT)
-    ART_ICON(wxART_FIND)
-    ART_ICON(wxART_FIND_AND_REPLACE)
-    ART_ICON(wxART_EDIT)
-    ART_ICON(wxART_FULL_SCREEN)
-    ART_ICON(wxART_HARDDISK)
-    ART_ICON(wxART_FLOPPY)
-    ART_ICON(wxART_CDROM)
-    ART_ICON(wxART_REMOVABLE)
-    ART_ICON(wxART_REFRESH)
-    ART_ICON(wxART_STOP)
 }
 
 
@@ -254,9 +198,10 @@ void wxArtBrowserDialog::OnChooseClient(wxCommandEvent &event)
 
 void wxArtBrowserDialog::SetArtBitmap(const wxArtID& id, const wxArtClient& client, const wxSize& size)
 {
-    wxBitmap bmp = wxArtProvider::GetBitmap(id, client, size);
-    m_canvas->SetSize(bmp.GetWidth(), bmp.GetHeight());
-    m_canvas->SetBitmap(bmp);
-    m_text->SetLabel(wxString::Format("Size: %d x %d", bmp.GetWidth(), bmp.GetHeight()));
+    wxBitmapBundle bb = wxArtProvider::GetBitmapBundle(id, client, size);
+    wxSize s = bb.GetBitmapFor(this).GetSize();
+    m_canvas->SetSize(s.GetWidth(), s.GetHeight());
+    m_canvas->SetBitmap(bb);
+    m_text->SetLabel(wxString::Format("Size: %d x %d", s.GetWidth(), s.GetHeight()));
     Refresh();
 }
```
</details>

vadz:
Thanks for finding this!

I see at least one more bug immediately: `wxArtProvider::GetNativeDIPSizeHint()` is supposed to return the size in DIPs, but it actually doesn't because `::GetSystemMetrics()` uses the DPI of the main screen and so is DPI-dependent. I've added the commit fixing this, but I'm not sure if that's all, i.e. if this fixes all the problems. Could you please retest with the latest commit and let me know if you still see anything wrong?

TIA!

MaartenBent:
I still see the problem. The bitmap that is used to create the bitmap bundle still has scale 1.5. Because these dip sizes are adjusted to the DPI in [wxArtProvider::GetNativeSizeHint](https://github.com/wxWidgets/wxWidgets/blob/61baa7f5fd3b142a529d920c6a651b559be962da/src/common/artprov.cpp#L608) (called from [wxWindowsArtProvider::CreateBitmap](https://github.com/wxWidgets/wxWidgets/blob/61baa7f5fd3b142a529d920c6a651b559be962da/src/msw/artmsw.cpp#L185)).

But I think your last commit does fix the bitmap size. Before your fix it was 72x72, now it is 48x48. Which I think is correct. Assuming the default `wxART_MESSAGE_BOX` size is 32x32, it should be 48x48 at 150% scale.

vadz:
> I still see the problem. The bitmap that is used to create the bitmap bundle still has scale 1.5.

Yes, but why is it a problem? Sorry if I'm missing something obvious here, but it seems to me that `wxBitmapBundleImplArt` deals with this case correctly, as it sets its normal size to `bitmap.GetDIPSize()`, which works for bitmaps using any scale.

> But I think your last commit does fix the bitmap size. Before your fix it was 72x72, now it is 48x48. Which I think is correct. Assuming the default `wxART_MESSAGE_BOX` size is 32x32, it should be 48x48 at 150% scale.

Yes.

So which problem do you still see? AFAICS `wxBitmapBundleImplArt::GetBitmap()` returns the bitmaps of correct size with the correct scale factor now. But, again, I could be missing something here...

MaartenBent:
The problem is that I can't get any smaller bitmaps. When I move the window to a 100% monitor, it still uses the bitmap with 1.5 scale, because this is the smallest size that the bitmap bundle can provide.

vadz:
Oh, of course, I see, sorry.

I think we should always return 1 from `wxBitmapBundleImpl::GetNextAvailableScale()` as the art provider should always return something reasonable at scale 1, shouldn't it? I.e. would
```diff
diff --git a/src/common/artprov.cpp b/src/common/artprov.cpp
index 1ac25a0add..e4ffcbb869 100644
--- a/src/common/artprov.cpp
+++ b/src/common/artprov.cpp
@@ -191,7 +191,20 @@ class wxBitmapBundleImplArt : public wxBitmapBundleImpl
         // ones should just override CreateBitmapBundle() directly), so we only
         // return the original bitmap scale, but hope that perhaps the provider
         // will have other (e.g. x2) scales too, when our GetBitmap() is called.
-        return i++ ? 0.0 : m_bitmapScale;
+        switch ( i++ )
+        {
+            case 0:
+                return 1.0;
+
+            case 1:
+                if ( m_bitmapScale != 1.0 )
+                    return m_bitmapScale;
+                wxFALLTHROUGH;
+
+            default:
+                // No other scales supported.
+                return 0.0;
+        }
     }
 
 private:
```
work?

MaartenBent:
Yes, this seems to work. The icon now gets smaller when moving to the 100% monitor. And when I open the resource browser on the 100% monitor, it shows the correct 32x32 size.

vadz:
Thanks for testing this! I've pushed this commit too now and I think this should be merged -- and also backported to 3.2 as all of the changes here look like bug fixes to me and unlikely to break anything.

But please correct me if I'm wrong (again). TIA!

PR #23132, opened 2023-01-12
Fix for a check of the combobox enable state, when control is recreating

YuSanka:
For check of the ComboBox enable state is better to use IsThisEnable() instead of IsEnable(),  IsEnabled() == IsThisEnabled() && parent.IsEnabled(). As a result it is possible, that when TopLevelWindow is disabled at the moment of a control recreation, then ComboBox is recreated with disable state.


vadz:
Thanks for the fix! I agree that this is the right thing to do and will merge it soon (and also backport to 3.2).

PR #23134, opened 2023-01-12
Avoid expanding the window size by too much on DPI change

### LINKED ISSUE #23091: wxListCtrl can grow extremely wide when moving between displays
When I have a `wxListCtrl` in report mode with a long text in columns and columns width set to ` wxLIST_AUTOSIZE`, it is initially displayed in reasonable size, having scrollbars. However when I try to move the window from primary screen to the secondary one (primary screen has the higher resolution and DPI), once the window moves over half to the secondary screen, the control and the whole frame grow extremely wide so that the content can fit columns, covering almost both screens.

It happens also when `wxListCtrl` is not the sole child of a `wxFrame` but is several levels deep in the sizer hierarchy.

This does not happen with 3.2.1 but I noticed that the frame always grows a bit there when moving to the secondary screen and therefore after moving it between the screens a couple of times, the window size ends being MUCH bigger too.

To Reproduce:
1. Run the patched minimal sample.
2. Attempt to move the frame from primary to secondary screen.
3. Once the frame is over half on the secondary screen, observe the size change.

Minimal sample patch
```diff
diff --git a/samples/minimal/minimal.cpp b/samples/minimal/minimal.cpp
index 501caf9096..d950ce629f 100644
--- a/samples/minimal/minimal.cpp
+++ b/samples/minimal/minimal.cpp
@@ -26,6 +26,8 @@
     #include "wx/wx.h"
 #endif
 
+#include "wx/listctrl.h"
+
 // ----------------------------------------------------------------------------
 // resources
 // ----------------------------------------------------------------------------
@@ -175,6 +177,17 @@ MyFrame::MyFrame(const wxString& title)
     CreateStatusBar(2);
     SetStatusText("Welcome to wxWidgets!");
 #endif // wxUSE_STATUSBAR
+
+    auto longText = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque nec magna at nulla tempor pellentesque. Cras felis lorem, sodales vel sagittis et, vulputate nec sem";
+
+    wxListCtrl* listCtrl = new wxListCtrl(this, wxID_ANY, wxDefaultPosition, wxDefaultSize, wxLC_REPORT);
+
+    listCtrl->AppendColumn("Item");
+    listCtrl->AppendColumn("SubItem");
+    listCtrl->InsertItem(0, longText);
+    listCtrl->SetItem(0, 1, longText);
+    listCtrl->SetColumnWidth(0, wxLIST_AUTOSIZE);
+    listCtrl->SetColumnWidth(1, wxLIST_AUTOSIZE);
 }

```

- wxWidgets version: GIT master
- OS and its version: Windows 10 21H2, OS build 19045.2364, Windows Feature Experience Pack 120.2212.4190.0
- Other: [displays:](https://raw.githubusercontent.com/PBfordev/wxsysinfoframe/master/screenshots/displays.png) DPI scaling 125% on the primary (2560x1440) and 100% on the secondary (1680x1050) display.

vadz:
Don't make the window size at least equal to GetBestSize() when the DPI changes as this could result in it becoming far too big, e.g. if it contains any controls whose current size is significantly smaller than their best size, as could be the case for just about any control using scrollbars and notably wxListCtrl.

Instead amend the changes of b66e63e (Fix windows becoming smaller than best size on DPI change, 2022-06-24) to only do this for the windows using sizers as we can determine the minimal acceptable size for them and we really shouldn't shrinking them beneath it.

Closes #23091.

---

@MaartenBent please have a look at this change if you can, TIA!

### DIFF
--- a/src/msw/nonownedwnd.cpp
+++ b/src/msw/nonownedwnd.cpp
@@ -23,6 +23,7 @@
     #include "wx/dcclient.h"
     #include "wx/frame.h"       // Only for wxFRAME_SHAPED.
     #include "wx/region.h"
+    #include "wx/sizer.h"
     #include "wx/msw/private.h"
 #endif // WX_PRECOMP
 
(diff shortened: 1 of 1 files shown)


MaartenBent:
I've tested it with some samples and don't see any regressions, so this looks good to me.

PBfordev:
I can confirm that it fixes my issue.

vadz:
Cherry-picked in the above commit, thanks for testing!

PR #23141, opened 2023-01-13
Fix web view Edge handler crash 

TcT2k:
Since WebView2 runtime 109.0.1518.49 `CreateWebResourceResponse()` requires `reasonPhrase` not to be `nullptr` during creation.


vadz:

PR #23142, opened 2023-01-13
Fix wxFontList::FindOrCreateFont() for pixel and fractional point sizes

### LINKED ISSUE #23139: wxFontList::FindOrCreateFont() does not work with pixel or fractional sizes
### Description
When calling `wxFontList::FindOrCreateFont(wxFontInfo&)`, only font size in points is considered:
https://github.com/wxWidgets/wxWidgets/blob/78a041f2ed63505d1a563da2885728f4af282b66/include/wx/font.h#L653-L657

However, in `wxFontInfo`, the point size is always `-1` when using the pixel size, so using any pixel size with `wxFontList::FindOrCreateFont(wxFontInfo&)` behaves as if a font with the point size of `-1` was requested.

I am not sure this can be easily fixed as the pixel size may not be available for fonts in the list created using a point size?

I did not test how the method works with  `wxFontInfo` using a fraction point size but I suspect it may also have a (small) problem, i.e., using a rounded integer instead of actual fractional point size.

While this is a low priority issue perhaps not worth spending time to fix it, I think it should be documented as a method's limitation. as the current behaviour is quite unexpected.

PBfordev:
This is an attempt to fix #23139. From the two approaches suggested by vadz in the referenced issue, I have chosen "just make it work" one instead of "make it work and pretty". The upside is that it can be easily backported. Yes that was the reason for my choice. ;P

Regarding the added tests. Instead of adding a new test file for `wxFontList` (which would require rebaking), I added one `TEST_CASE` to existing `fonttest.cpp`.

The new tests are very simple (but IMO sufficient); however, I am not sure if any tests in `fonttest.cpp` are actually run and from which test executable (gui or drawing?). I am also not sure about comparing doubles with `==`  and `Approx()` should probably have been used there.

### DIFF
--- a/include/wx/font.h
+++ b/include/wx/font.h
@@ -640,7 +640,12 @@ public:
                              wxFontWeight weight,
                              bool underline = false,
                              const wxString& face = wxEmptyString,
-                             wxFontEncoding encoding = wxFONTENCODING_DEFAULT);
+                             wxFontEncoding encoding = wxFONTENCODING_DEFAULT)
+        {   return FindOrCreateFont(wxFontInfo(pointSize)
+                                    .Family(family)
+                                    .Style(style).Weight(weight).Underlined(underline)
+                                    .FaceName(face).Encoding(encoding));
+        }
 
     wxDEPRECATED_MSG("use wxFONT{FAMILY,STYLE,WEIGHT}_XXX constants")
     wxFont *FindOrCreateFont(int pointSize, int family, int style, int weight,
@@ -650,11 +655,7 @@ public:
         { return FindOrCreateFont(pointSize, (wxFontFamily)family, (wxFontStyle)style,
                                   (wxFontWeight)weight, underline, face, encoding); }
 
-    wxFont *FindOrCreateFont(const wxFontInfo& fontInfo)
-        { return FindOrCreateFont(fontInfo.GetPointSize(), fontInfo.GetFamily(),
-                                  fontInfo.GetStyle(), fontInfo.GetWeight(),
-                                  fontInfo.IsUnderlined(), fontInfo.GetFaceName(),
-                                  fontInfo.GetEncoding()); }
+    wxFont *FindOrCreateFont(const wxFontInfo& fontInfo);
 };
 
 extern WXDLLIMPEXP_DATA_CORE(wxFontList*)    wxTheFontList;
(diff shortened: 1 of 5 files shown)


PBfordev:
I don't understand the test failures.

Even on Windows it fails: when I request point size 10.5, I get 11: e.g., [here](https://github.com/wxWidgets/wxWidgets/actions/runs/3914869115/jobs/6692417525#step:10:24). However, I get 10.5 on my Windows 10 computer as expected? Are default fonts on CIs not supporting fractional sizes?

I will not attempt to make further changes to the PR and will wait for suggestions on how to continue with this.

vadz:
Err, I really don't understand how can it pass for you, `GetPointSize()` returns an int which can never be equal to `10.5`. Replacing it with `GetFractionalPointSize()` should make it work, so I've pushed a commit doing this to check...

PBfordev:
> Err, I really don't understand how can it pass for you

Sorry, it did not. I tested in a separate code where I actually did use `GetFractionalPointSize()`.

vadz:
OK, this would explain it :-)

The failures in Unix tests are also understandable and I actually wonder if wxMSW shouldn't behave like this too: it makes sense to specify 0 width when creating the font, but not when querying it for size. In any case, we could just compare `size.y` here.

But there is also a failure under Mac which is more worrisome:
```
./font/fonttest.cpp:505: FAILED:
  REQUIRE( font1->GetPixelSize() == pixelSize )
with expansion:
  16*31 == 0*32
```
This one might be a real bug...

Will you be able to look at it or should I try to?

vadz:
I really shouldn't have spent half of my Saturday on this but, well, I did... unfortunately the situation is pretty bad under Mac and I can't fix all the problems there, so we'll have to just work around for them for now (as it was already the case before...), but at least we should be able to merge this if the tests pass now.

PBfordev:
> Will you be able to look at it or should I try to?

Sorry but no, I am Windows only.

vadz:
If the latest version works for you, let's merge it. Please let me know if you have any objections.

PBfordev:
It works. Thanks for the quick fix and sorry for ruining your Saturday.

vadz:
Thanks for testing! And no, it wasn't ruined at all, it's just that I hoped to do something probably more important in wx that I won't have time to do any longer today, but it's solely my fault for being so easily distracted. And the fact that this allowed us to find and already partially fix #23144 means that this was more important than I had thought, too, so maybe it's all for the best.

Anyhow, I'll merge this soon.

PR #23147, opened 2023-01-15
Adjust wxUILocaleImpl::GetPreferredUILanguages

utelle:
Instead of preferring LANGUAGE and falling back to LC_ALL, LC_MESSAGES, or LANG, if LANGUAGE is not set, LC_ALL, LC_MESSAGES, and LANG are checked first, because the first item in LANGUAGE should be equal to LANG anyway.

Only, if none of the variables is set or valid, use English (en_US) as a fallback.


vadz:
Thanks! I'll test this together with my other changes in (the upcoming update of) #23119 and hopefully merge it soon.

vadz:
I've included this in #23226, so closing this one, thanks for fixing this!

PR #23148, opened 2023-01-16
Fix webview edge on Windows 7

TcT2k:
Disable double buffering on the edge host window. With double buffering enabled rendering fails on Windows 7 (possibly Win8). It shouldn't affect performance as the contents are rendered and hosted by the external edge process anyway.


vadz:
Thanks!

PR #23157, opened 2023-01-19
Increase margin of listctrl items on mac

ryanseghers:
Change listctrl to apply an icon offset on mac. The 4px offset looks good on my test app on mac. The offset was already fine on Windows and Linux.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -402,3 +402,6 @@
 /utils/wxrc/wxrc_vc[789].sln
 
 /3rdparty/webview2
+
+# mac os
+.DS_Store/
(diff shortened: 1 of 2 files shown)


ryanseghers:
Here is my test app after the change:
mac-after-change

ryanseghers:
Here is the listctrl sample before the change:
mac-listctrl-sample-before-change

And after the change:
mac-listctrl-sample-after-change

vadz:
Sorry but I'd prefer to rid of the existing `#if`s rather than adding more of them. If we want to increase `EXTRA_BORDER_X` to 4, let's do it for all platforms, I don't see any valid reason to do it only under Mac but not under Linux.

But for now I'll apply #23156 to add a 2px border (which should be scaled by DPI, BTW...) rather than this one.

PR #23163, opened 2023-01-21
Fix initializing wxFont from wxNativeFontInfo in wxQt

AliKet:
This also makes the font test pass for `wxQt`


vadz:
Thanks for fixing this! Pushing now, and I think this should be also backported to 3.2, right?

PR #23167, opened 2023-01-22
Suppress -Wmaybe_unitialized warning

Blake-Madden:
Under GCC, creating an `std::optional<wxString>` causes a `-Wmaybe_unitialized` warning. Seems like a false positive, but using a default null for `m_str` doesn't hurt and suppresses the warning.

### DIFF
--- a/include/wx/string.h
+++ b/include/wx/string.h
@@ -3384,8 +3384,8 @@ private:
           return wxScopedCharTypeBuffer<T>::CreateNonOwned(m_str, m_len);
       }
 
-      T *m_str;     // pointer to the string data
-      size_t m_len; // length, not size, i.e. in chars and without last NUL
+      T *m_str{nullptr};     // pointer to the string data
+      size_t m_len{0}; // length, not size, i.e. in chars and without last NUL
   };
 
 
--- a/include/wx/string.h
+++ b/include/wx/string.h
@@ -241,7 +241,7 @@ public:
 
 private:
     // the wxString this object was returned for
-    const wxString *m_str;
+    const wxString *m_str{nullptr};
     // Offset into c_str() return value. Note that this is *not* offset in
     // m_str in Unicode characters. Instead, it is index into the
     // char*/wchar_t* buffer returned by c_str(). It's interpretation depends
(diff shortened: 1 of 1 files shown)


include/wx/string.h
```diff
@@ -241,7 +241,7 @@ class wxCStrData
 
 private:
     // the wxString this object was returned for
-    const wxString *m_str;
+    const wxString *m_str{nullptr};
```
vadz, line 244: I'd rather avoid doing this as this gives the impression that this pointer can be null, which is not the case, it must always be valid.

include/wx/string.h
```diff
@@ -280,10 +280,10 @@ class WXDLLIMPEXP_BASE wxStringIteratorNode
...
         { clear(); DoSet(str, nullptr, iter); }
 
-    const wxString *m_str;
-    wxStringImpl::const_iterator *m_citer;
-    wxStringImpl::iterator *m_iter;
-    wxStringIteratorNode *m_prev, *m_next;
+    const wxString *m_str{nullptr};
+    wxStringImpl::const_iterator *m_citer{nullptr};
+    wxStringImpl::iterator *m_iter{nullptr};
+    wxStringIteratorNode *m_prev{nullptr}, *m_next{nullptr};
```
vadz, line 286: This could be done, but the ctor above should be replaced with `= default` then, it's useless to initialize the members in both places.

include/wx/string.h
```diff
@@ -3384,8 +3384,8 @@ class WXDLLIMPEXP_BASE wxString
           return wxScopedCharTypeBuffer<T>::CreateNonOwned(m_str, m_len);
       }
 
-      T *m_str;     // pointer to the string data
-      size_t m_len; // length, not size, i.e. in chars and without last NUL
+      T *m_str{nullptr};     // pointer to the string data
+      size_t m_len{0}; // length, not size, i.e. in chars and without last NUL
```
vadz, line 3388: Same as above: this is fine to do, but redundant initialization in the ctor should be removed.

vadz (changes requested):
Sorry, but I'd rather not apply this in its current state for the reasons explained in the comments.

Blake-Madden:
OK, made the suggested changes. Thanks for the feedback.

vadz (approved):

Issue #23172, opened 2023-01-23
wxJoystick::GetProductName() returns blank string on Windows (suggested fix attached)


inkmeister:
On my system of 64-bit Windows 10 (22H2) with wxWidgets 3.2.1, wxJoystick::GetProductName() returns a blank string. This is due to the referenced registry keys not existing in HKEY_LOCAL_MACHINE but HKEY_CURRENT_USER. My simple fix was to replace the two instances of HKLM to HKCU in the aforementioned function.

Unfortunately I don’t know which version of Windows moved these keys, i.e. whether it’s also worth amending the function to try and read from HKCU first and then fall back to reading from HKLM.

vadz:
Unfortunately I don't know this neither and I don't even have any joysticks to test this with, but it looks that trying HKCU first and then falling back in HKLM would work just fine in any case, so if you'd like to make a PR changing the code to do it, I'd be glad to merge it.

MaartenBent:
I can reproduce it with a controller. I need to use `wxRegKey::HKCU` as well.

I also noticed the joystick sample does not work anymore. It seems to require `MSWDisableComposited` (because of `wxClientDC`?) . And drawing doesn't look right at high DPI. I'll fix and update the sample. I can include the registry fix as well.

PR #23175, opened 2023-01-24
Update joystick sample, fix joystick name

MaartenBent:
Fix determining the joystick name by also checking `HKCU` registry.
See #23172 

Update the joystick sample to work with `WS_EX_COMPOSITED` and modernize it a bit.

joystick


vadz:
Nice, thanks! I'll backport the first commit to 3.2 too.

PR #23180, opened 2023-01-27
Modernize and clean up XRC code

vadz:
No real changes here, this is just cleanup in preparation for other things.


src/xrc/xmlres.cpp
```diff
@@ -629,6 +642,10 @@ static void ProcessPlatformProperty(wxXmlNode *node)
             }
         }
 
+        if (isok && c->GetAttribute("", &s))
+        {
+        }
+
```
vadz, line 648: Oops, this is a leftover, will remove.

PR #23181, opened 2023-01-27
macOS: Fix wxWebViewHandlerRequest::GetDataString

TcT2k:
An additional call to `wxWebViewHandlerRequest::GetDataString()` would return an empty string because the underlying memory stream wasn't reset.

(Not applicable to 3.2)


vadz:
Thanks, cherry-picking to master.

PR #23186, opened 2023-01-29
Fix signature of dark-mode related uxtheme.dll functions

### LINKED ISSUE #23169: MSW dark mode misdetected in release MSVC builds
### Description
I have set the value of system environment variable `wx_msw_dark_mode` to `1`, which means that wxWidgets should use dark mode only when the system does. My PC is not set to use dark mode.

Surprisingly, when running an MSVC-built application in the release mode, it behaves as if Windows were in the dark mode (or rather `wx_msw_dark_mode` was set to `2`), i.e., the controls are drawn partially dark. I can reproduce it with the three last MSVS versions (tested 64-bit shared wxWidgets builds). 

The issue does not manifest in the MSVC Debug build (tried only with MSVS 2022) nor gcc or clang release builds. As I find such bug hard to believe, I have restarted the computer several times but nothing has changed, I do not change light/dark setting in Windows (ever) or the value of `wx_msw_dark_mode` after restart. The issue goes away when I set `wx_msw_dark_mode` to `0` (without needing restart), i.e., never use dark mode. Can anyone else reproduce this (the computer I tried this on does not show any problems)?

The following screenshot has the minimal sample built in the release build overlapping the window of the minimal sample built in the debug build.
wxdark-mode issue

Notebook sample built in release mode:
wxnotebook-dark

The wxWidgets and samples were always built with CMake in an out-of-source folder, like this
```
cmake -G "Visual Studio 17 2022" -DwxBUILD_SAMPLES=ALL -S ../wxWidgets-PB -B .
cmake --build . --config Release
```
only the generators were different (`-G "Visual Studio 16 2019"` and `-G "Visual Studio 15 2017" -A x64`).

_MSVS versions (are the latest available for each):_
2022: 17.4.4
2019: 16.11.23
2017: 15.9.49 (minimal sample built with this has also black frame client area?)

Each build was in its own folder and I have tried building each build at least twice, deleting everything in the build folder before another build trial.

- wxWidgets version: current master (commit 842ca1e)
- wxWidgets port: wxMSW
- OS: Windows 10 22H2 build 19045.2486

**EDIT**
I can reproduce it on another computer, so it looks like a real bug.

vadz:
They were wrongly declared as taking (which is not very important) and returning (which was very bad as it resulted to interpreting the return values incorrectly in optimized builds) Win32 4-byte BOOL instead of the C++ 1-byte bool that they actually use.

Closes #23169.

### DIFF
--- a/src/msw/darkmode.cpp
+++ b/src/msw/darkmode.cpp
@@ -92,8 +92,10 @@ namespace wxMSWImpl
 
 // Global pointers of the functions we use: they're not only undocumented, but
 // don't appear in the SDK headers at all.
-BOOL (WINAPI *ShouldAppsUseDarkMode)() = nullptr;
-BOOL (WINAPI *AllowDarkModeForWindow)(HWND hwnd, BOOL allow) = nullptr;
+//
+// Note that, not being public, they use C++ bool type and not Win32 BOOL.
+bool (WINAPI *ShouldAppsUseDarkMode)() = nullptr;
+bool (WINAPI *AllowDarkModeForWindow)(HWND hwnd, bool allow) = nullptr;
 DWORD (WINAPI *SetPreferredAppMode)(DWORD) = nullptr;
 
 bool InitDarkMode()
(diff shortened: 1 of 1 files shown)


PBfordev:
I can confirm that this fixes the issue.

vadz:
Applied to master in the commit referenced above.

PR #23187, opened 2023-01-29
Create wxPGGlobalVars on demand and not on startup

### LINKED ISSUE #23165: wxPropertyGrid use of global state causes problems when using wxPython
### Description

`wxPropertyGrid` makes use of module-level global state (`wxPGGlobalVars`) that can be mutated by the user, for example to register custom editors.

wxPython has several modules (`wx.adv`, `wx.html`, etc) that call `wxPyReinitializeModules` when they are imported.

This call wipes out changes made to the global state, and programs using wxPython alongside wxWidgets therefore can have their custom editors broken just by importing some wxPython modules.  See https://gitlab.com/kicad/code/kicad/-/issues/13280 for details.

There are various workarounds that could be done, but in my opinion the right fix for this is to get rid of global state so that it is safe to call `wxPyReinitializeModules` at any point in time.  It does not seem clear to me that there is any reason that the `wxPropertyGrid` state must be global, rather than local to an instance.  It would be fine for me if I had to register editor classes once per `wxPropertyGrid` instance.

- wxWidgets port  you use: wxMSW, wxGTK, wxOSX
- OS  and its version: Windows 10; Ubuntu (various); macOS 11/12/13

vadz:
This is slightly more efficient as it avoids allocating an object which is possibly not ever going to be used on startup of any application linking with the propgrid library (at the price of slower access to this object, but this should never be done in performance-critical parts of the code, hopefully) and avoids a problem with losing the previously registered editors when re-initializing wx modules, as wxPython does.

Closes #23165.

### DIFF
--- a/include/wx/propgrid/propgrid.h
+++ b/include/wx/propgrid/propgrid.h
@@ -113,7 +113,14 @@ public:
     int HasExtraStyle( int style ) const { return (m_extraStyle & style); }
 };
 
-extern WXDLLIMPEXP_DATA_PROPGRID(wxPGGlobalVarsClass*) wxPGGlobalVars;
+// Internal class providing access to the global wxPGGlobalVars instance.
+class WXDLLIMPEXP_PROPGRID wxPGGlobalVarsPtr
+{
+public:
+    wxPGGlobalVarsClass* operator->() const;
+};
+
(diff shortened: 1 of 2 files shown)


vadz:

PR #23189, opened 2023-01-29
Try forcing core files creation

vadz:
Ubuntu seems to be using apport, which intercepts core files creation and creates them in a different directory. Try to undo this by resetting the kernel core pattern setting to the default value.

---

Trying to fix the problem from https://github.com/wxWidgets/wxWidgets/actions/runs/4038112428/jobs/6941882804

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,11 @@ jobs:
   build:
     runs-on: ${{ matrix.runner }}
     name: ${{ matrix.name }}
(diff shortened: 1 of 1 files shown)


vadz:
This should help when the tests crash the next time. I have no idea why is `/proc/sys/kernel/core_pattern` set to `|apport ...` even in the Docker Ubuntu 18.04 image which doesn't even have apport installed, but it is, so we have to work around it.

vadz:

PR #23190, opened 2023-01-29
Actually return English with wxLOCALE_FORM_ENGLISH in GetLocalizedName

lanurmi:
`GetLocalizedName` on macOS would return the name localized with the current locale instead of English, when English was specifically requested.

Steps to reproduce:
1. Set system language/country to German/Germany.
2. Start the `internat` sample.
3. Observe that UI locale name in English is shown as `Deutsch (Deutschland)`, which is not widely considered to be English.

Bildschirmfoto 2023-01-29 um 23 39 37

If I'm reading [Apple's documentation](https://developer.apple.com/documentation/foundation/nslocale/1642864-localizedstringforlocaleidentifi) correctly, it explicitly states that the code in `case wxLOCALE_FORM_NATIVE` is equivalent to the code in `case wxLOCALE_FORM_ENGLISH`, so it makes sense the output is identical too.

This PR changes the code to create an `en_US`-`NSLocale` object for acquiring the name in English.

### DIFF
--- a/src/osx/core/uilocale.mm
+++ b/src/osx/core/uilocale.mm
@@ -150,6 +150,8 @@ wxString
 wxUILocaleImplCF::GetLocalizedName(wxLocaleName name, wxLocaleForm form) const
 {
     NSString* str = nullptr;
+    NSLocale *enLocale = [[NSLocale alloc] initWithLocaleIdentifier:@"en_US"];
+    [enLocale autorelease];
     switch (name)
     {
         case wxLOCALE_NAME_LOCALE:
@@ -159,7 +161,7 @@ wxUILocaleImplCF::GetLocalizedName(wxLocaleName name, wxLocaleForm form) const
                     str = [m_nsloc localizedStringForLocaleIdentifier:[m_nsloc localeIdentifier]];
                     break;
                 case wxLOCALE_FORM_ENGLISH:
-                    str = [m_nsloc displayNameForKey:NSLocaleIdentifier value:[m_nsloc localeIdentifier]];
+                    str = [enLocale displayNameForKey:NSLocaleIdentifier value:[m_nsloc localeIdentifier]];
                     break;
                 default:
                     wxFAIL_MSG("unknown wxLocaleForm");
@@ -172,7 +174,7 @@ wxUILocaleImplCF::GetLocalizedName(wxLocaleName name, wxLocaleForm form) const
                     str = [m_nsloc localizedStringForLanguageCode:[m_nsloc languageCode]];
                     break;
                 case wxLOCALE_FORM_ENGLISH:
-                    str = [m_nsloc displayNameForKey:NSLocaleIdentifier value:[m_nsloc localeIdentifier]];
+                    str = [enLocale displayNameForKey:NSLocaleIdentifier value:[m_nsloc localeIdentifier]];
                     break;
                 default:
                      wxFAIL_MSG("unknown wxLocaleForm");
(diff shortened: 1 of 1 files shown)


src/osx/core/uilocale.mm
```diff
@@ -150,6 +150,8 @@ int CompareStrings(const wxString& lhs, const wxString& rhs,
 wxUILocaleImplCF::GetLocalizedName(wxLocaleName name, wxLocaleForm form) const
 {
     NSString* str = nullptr;
+    NSLocale *enLocale = [[NSLocale alloc] initWithLocaleIdentifier:@"en_US"];
+    [enLocale autorelease];
```
vadz, line 154: Not sure if it's really an expensive operation, but we could avoid creating this locale until it's really needed relatively easily by just doing
```suggestion
    auto const getEnLocale = []() {
        NSLocale *enLocale = [[NSLocale alloc] initWithLocaleIdentifier:@"en_US"];
        [enLocale autorelease];
        return enLocale;
    };
```
and then using `getEnLocale()` below. Any reason not to do it like this?

src/osx/core/uilocale.mm
```diff
@@ -172,7 +174,7 @@ int CompareStrings(const wxString& lhs, const wxString& rhs,
                     str = [m_nsloc localizedStringForLanguageCode:[m_nsloc languageCode]];
                     break;
                 case wxLOCALE_FORM_ENGLISH:
-                    str = [m_nsloc displayNameForKey:NSLocaleIdentifier value:[m_nsloc localeIdentifier]];
+                    str = [enLocale displayNameForKey:NSLocaleIdentifier value:[m_nsloc localeIdentifier]];
```
vadz, line 177: According to @utelle's changes in #23199 this should be
```suggestion
                    str = [enLocale displayNameForKey:NSLocaleLanguageCode value:[m_nsloc localeIdentifier]];
```

src/osx/core/uilocale.mm
```diff
@@ -185,7 +187,7 @@ int CompareStrings(const wxString& lhs, const wxString& rhs,
                     str = [m_nsloc localizedStringForCountryCode:[m_nsloc countryCode]];
                     break;
                 case wxLOCALE_FORM_ENGLISH:
-                    str = [m_nsloc displayNameForKey:NSLocaleIdentifier value:[m_nsloc localeIdentifier]];
+                    str = [enLocale displayNameForKey:NSLocaleIdentifier value:[m_nsloc localeIdentifier]];
```
vadz, line 190: And this
```suggestion
                    str = [enLocale displayNameForKey:NSLocaleCountryCode value:[m_nsloc localeIdentifier]];
```

vadz (commented):
Please let me know if you think the suggestions here are correct, I haven't tested them myself yet.

src/osx/core/uilocale.mm
utelle: Although it works when using `localeIdentifier` for the value attribute it would be clearer to use `languageCode`:
```c++
str = [enLocale displayNameForKey:NSLocaleIdentifier value:[m_nsloc languageCode]];
```
Even more concise would be:
```c++
str = [enLocale localizedStringForLanguageCode:[m_nsloc languageCode]];
```
Therefore I would recommend to use the latter.

src/osx/core/uilocale.mm
utelle: Same as for language.

Although it works when using `localeIdentifier` for the value attribute it would be clearer to use `countryCode`:
```c++
str = [enLocale displayNameForKey:NSLocaleIdentifier value:[m_nsloc countryCode]];
```
Even more concise would be:
```c++
str = [enLocale localizedStringForCountryCode:[m_nsloc countryCode]];
```
Therefore I would recommend to use the latter.

vadz:
Thanks Ulrich, I agree that this is even better and if Lauri doesn't have any objections, I'm finally going to merge that version -- and without `getEnLocale()` as it's indeed useless if it's used only once.

lanurmi:
No objections 👍

vadz:
Applied to master now and will apply to 3.2 branch too soon, thanks again!

PR #23196, opened 2023-01-30
Improve generic and MSW "About" dialog appearance

vadz:
The new screenshots show how the dialogs look now.


vadz:
BTW, as can be seen in GTK screenshots, the dialog shows the caret by default, which looks a bit ugly and doesn't seem to happen in the other GTK apps. Would anybody know how to get rid of it by chance?

vadz:
Not sure if this is appropriate to backport to 3.2, any thoughts?

PR #23198, opened 2023-01-30
GTK: wxGetKeyState to support WXK_CAPITAL WXK_NUMLOCK WXK_SCROLL

### LINKED ISSUE #23195: wxGetKeyState doesn't support Caps/Num/Scroll lock state under non-X11 GTK (Wayland for example)
### Description
wxGetKeyState raises error when invoked with WXK_CAPITAL/WXK_NUMLOCK/WXK_SCROLL under Wayland -bases system.
Its because current wxGetKeyStateGTK implementation fails when invoked with above keys, however it must be possible to detect them using gdk_keymap_get_caps_lock_state/gdk_keymap_get_num_lock_state/gdk_keymap_get_scroll_lock_state with proper GTK version

- wxWidgets - any (including master)
- wxWidgets port - wxGTK
Ubuntu 22.04, Wayland login

GTK version: any

elfmz:
Fixed #23195
Note that its actually possible to make wxGetKeyStateGTK to support some of modifiers under GTK earlier than 3.4 as gdk_keymap_get_caps_lock_state supported by GTK 2.16 and gdk_keymap_get_num_lock_state supported by GTK 3.0 but i intentionally left check for 3.4 as it was to keep 'silent fail' behavior under older versions and to simplify things...

### DIFF
--- a/src/unix/utilsx11.cpp
+++ b/src/unix/utilsx11.cpp
@@ -2608,34 +2608,46 @@ static bool wxGetKeyStateGTK(wxKeyCode key)
     GdkDisplay* display = gdk_window_get_display(wxGetTopLevelGDK());
     GdkKeymap* keymap = gdk_keymap_get_for_display(display);
 
-    if (key == WXK_ALT)
-        return (gdk_keymap_get_modifier_state(keymap) & GDK_MOD1_MASK) != 0;
+    guint mask = 0;
+    switch (key)
+    {
+        case WXK_ALT:
+            mask = GDK_MOD1_MASK;
+            break;
 
-    if (key == WXK_CONTROL)
-        return (gdk_keymap_get_modifier_state(keymap) & GDK_CONTROL_MASK) != 0;
+        case WXK_CONTROL:
+            mask = GDK_CONTROL_MASK;
+            break;
 
-    if (key == WXK_SHIFT)
-        return (gdk_keymap_get_modifier_state(keymap) & GDK_SHIFT_MASK) != 0;
+        case WXK_SHIFT:
+            mask = GDK_SHIFT_MASK;
+            break;
 
-    if (key == WXK_CAPITAL)
-        return gdk_keymap_get_caps_lock_state(keymap) != FALSE;
+        case WXK_CAPITAL:
+            return gdk_keymap_get_caps_lock_state(keymap) != FALSE;
 
-    if (key == WXK_NUMLOCK)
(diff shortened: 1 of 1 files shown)


src/unix/utilsx11.cpp
```diff
@@ -2607,27 +2607,35 @@ static bool wxGetKeyStateGTK(wxKeyCode key)
...
+    if (key == WXK_SHIFT)
+        return (gdk_keymap_get_modifier_state(keymap) & GDK_SHIFT_MASK) != 0;
+
+    if (key == WXK_CAPITAL)
+        return gdk_keymap_get_caps_lock_state(keymap) != FALSE;
+
+    if (key == WXK_NUMLOCK)
+        return gdk_keymap_get_num_lock_state(keymap) != FALSE;
+
+# if GTK_CHECK_VERSION(3,18,0)
```
vadz, line 2626: We need both compile- and run-time version checks using `gtk_check_version()`, as is done elsewhere, otherwise the program compiled on a current system but running on the old one would just crash.

vadz (commented):
Thanks for fixing this!

I'd prefer to keep `switch` if only because it would let us have the `default` clause, but it's a small detail.

Will you be able to add the run-time version check or should I do it?

src/unix/utilsx11.cpp
elfmz: But what runtime check is missing? There is (was) if (gtk_check_version(3,4,0) != nullptr) in very beginning of wxGetKeyStateGTK that automatically ensures all functions inside except gdk_keymap_get_scroll_lock_state. And gdk_keymap_get_scroll_lock_state has corresponding gtk_check_version(3,18,0).

elfmz: Regarding using switch - i removed it to avoid using of gdk_keymap_get_modifier_state where its not needed. Well, it was possible to add another switch for functions that dont need gdk_keymap_get_modifier_state but for me two switch-es would look more tricky than sequence of if-s...

vadz: Sorry, I'm blind. Or so used to seeing `gtk_check_version()` check on its own line that I somehow skipped it below. No problem here, sorry again.

vadz:
What do you think of restoring the `switch` like this? I find it a bit clearer like this, even if it's not a huge deal. If you agree, I'm going to (squash) merge this into master.

elfmz:
its also fine

vadz:
Thanks, merged now and will backport to 3.2 too soon.

Issue #23199, opened 2023-01-30
macOS: wxUILocale::GetLocalizedName() returns localized names when asked to return English names


utelle:
### Description

On _macOS_ calling method `wxUILocale::GetLocalizedName()` always returns localized names, even when asked to return English names. For example for a German locale

`wxUILocale::GetLocalizedName(wxLOCALE_NAME_LOCALE, wxLOCALE_FORM_ENGLISH)`

returns **"Deutsch (Deutschland)"** instead of the expected **"German (Germany)"**.

This problem can be seen in the _internat_ sample application.

Additionally, for English names always the locale name is returned instead of the language or the country.

### Platform and version information

- wxWidgets version you use: 3.2.1
- wxWidgets port you use: wxOSX
- OS and its version: macOS Ventura 13.2

vadz:
This should have been fixed by 60e67de06ee4b0da2da1ec777d23b6cd7c026042, so closing.

PR #23200, opened 2023-01-30
Fix issue #23199

utelle:
On macOS calling method wxUILocale::GetLocalizedName() always returns localized names, even when asked to return English names. Additionally, for English names always the locale name is returned instead of the language or the country.


lanurmi:
This is essentially the same as PR #23190, isn't it.

vadz:
Yes, indeed, thanks, so I'm closing this one as the other one looks better because it does auto-release.

vadz:
Wait, it looks like there are more fixes here than in the other one actually, as it also switches to using `NSLocaleLanguageCode` and `NSLocaleCountryCode` for `wxLOCALE_NAME_LANGUAGE` and `wxLOCALE_NAME_COUNTRY` respectively, which also looks like the right thing to do, doesn't it?

PR #23204, opened 2023-01-31
OSX default window menu title i18n

csomor:
Uses properly translated title when we have to add a window menu of our own …

### DIFF
--- a/src/osx/cocoa/menu.mm
+++ b/src/osx/cocoa/menu.mm
@@ -271,7 +271,7 @@ public :
                 }
             }
 
-            windowMenu = [[NSMenu alloc] initWithTitle:nsWindowMenuTitle];
+            windowMenu = [[NSMenu alloc] initWithTitle:nsAlternateWindowMenuTitle];
             NSMenuItem* windowMenuItem = [[NSMenuItem alloc] initWithTitle:nsWindowMenuTitle action:nil keyEquivalent:@""];
             [windowMenuItem setSubmenu:windowMenu];
             [windowMenu release];


src/osx/cocoa/menu.mm
```diff
@@ -271,7 +271,7 @@ virtual void MacSetupHelpMenu()
                 }
             }
 
-            windowMenu = [[NSMenu alloc] initWithTitle:nsWindowMenuTitle];
+            windowMenu = [[NSMenu alloc] initWithTitle:nsAlternateWindowMenuTitle];
             NSMenuItem* windowMenuItem = [[NSMenuItem alloc] initWithTitle:nsWindowMenuTitle action:nil keyEquivalent:@""];
```
vadz, line 275: Is it intentional that this doesn't use translated menu title? If so, it would certainly be helpful to have a comment explaining why, as it doesn't seem obvious to me...

vadz:
Thanks, this certainly should be merged, but could we also rename `nsAlternateWindowMenuTitle` to something like `nsTranslatedWindowMenuTitle`? It took me some time to realize what did "alternate" mean here. I can do this myself, of course, if you agree (I don't know but perhaps "alternate" has some mystic Mac-only meaning that I'm not privy to).

csomor:
@vadz please go ahead, yes, no magic name introspection involved ;-)

csomor:
and then the help menu title in the same method could be changed accordingly

src/osx/cocoa/menu.mm
csomor: as you wrote later, the naming is not obvious, `nsAlternateWindowMenuTitle` is the translated one

vadz:
Done in master and also cherry-picked the original commit to 3.2.

PR #23206, opened 2023-01-31
Revert "Use new CoreGraphics API to reset clipping region"

vadz:
This reverts commit dc11bb18cf2ebc3f36a6d5b083a95b5639149c4f as it results in visual artifacts due to CGContextResetClip() apparently not fully resetting everything.

See #22629.


vadz:
Applied to 3.2, see the commit referenced above.

Issue #22805, opened 2022-09-18
wxSimplebook causes link conflict


marekr:
This is a linker bug introduced in 3.2.0, wxSimplebook was derived  from `wxNavigationEnabled<wxBookCtrlBase>` instead of `wxBookCtrlBase`.

But per the comment above it
>// NB: This class doesn't use DLL export declaration as it's fully inline.

The problem is other classes like wxAuiNotebook use wxNavigationEnabled<wxBookCtrlBase>

`class WXDLLIMPEXP_AUI wxAuiNotebook : public wxNavigationEnabled<wxBookCtrlBase>`

thus the wx DLLs end up with their own templated version of `wxNavigationEnabled<wxBookCtrlBase>` exported.

But because wxSimplebook is an inline class, your application will also generate its own `wxNavigationEnabled<wxBookCtrlBase>` expansion.

The result is a conflict if you use wxSimplebook and one of the other wx book controls.

```
E:\kicad\kicad\build\x64-Release\kicad\wxmsw32u_aui.lib(wxmsw32u_aui_vc.dll) : error LNK2005: "public: __cdecl wxNavigationEnabled<class wxBookCtrlBase>::wxNavigationEnabled<class wxBookCtrlBase>(void)" (??0?$wxNavigationEnabled@VwxBookCtrlBase@@@@QEAA@XZ) already defined in panel_display_options_base.cpp.obj
E:\kicad\kicad\build\x64-Release\kicad\wxmsw32u_aui.lib(wxmsw32u_aui_vc.dll) : error LNK2005: "public: virtual __cdecl wxNavigationEnabled<class wxBookCtrlBase>::~wxNavigationEnabled<class wxBookCtrlBase>(void)" (??1?$wxNavigationEnabled@VwxBookCtrlBase@@@@UEAA@XZ) already defined in panel_display_options_base.cpp.obj
 
```

AliKet:
Does exporting `wxSimplebook` solves the problem for you ?

vadz:
Thanks for reporting this, it's clearly a problem, but I'm not so sure it's `wxSimplebook`'s fault -- wouldn't the same thing happen if you derived your own class from e.g. `wxNavigationEnabled<wxWindow>`?

[...]

PR #23207, opened 2023-02-01
Work around link errors when using wxSimplebook in DLL wx build

vadz:
This is an uglier workaround than the one applied in dbbf509e18 (Fix link when using wxSimplebook both directly and indirectly, 2023-01-29) to master but has the advantage of preserving ABI-compatibility.

See #22805.


vadz:
Applied in the commit above.

PR #23212, opened 2023-02-02
Fix non-fuzzy keyboard key names that are clearly wrong for Finnish

lanurmi:
Fix non-fuzzy keyboard key names that are clearly wrong for Finnish

The translations make sense in another context, but not as key names. Also one typo. Also fix UTF-8 of comment lines.


vadz:

PR #23213, opened 2023-02-02
Use translation context for more standard OSX menu items

lanurmi:
Similar items were added context in d85f0e311d40b56732b52905b7ee7d73b8414494,
but some were missed back then.

The context primarily serves as a translator comment, informing that these
are indeed OSX menu item names, and not something else.

Also to match the established translations of OSX itself, for some
languages these may need to be translated using particular terminology
different than on other platforms.

wxstd.pot should be regenerated; not done by this PR.


vadz:
Applied now, thanks!

PR #23215, opened 2023-02-03
Add move operations and rvalue support to wxString

pavel-t:
Ran benchmarks on msvc 2022 and mingw-w64 12 x64 builds, got 55-130% speedup in insert depending on build, 15-44% in sorting, 1-24% in push_back. wxArrayString insert/push_back benchmarks are only affected in STL builds.

### DIFF
--- a/include/wx/string.h
+++ b/include/wx/string.h
@@ -1078,8 +1078,17 @@ private:
 
   static wxString FromImpl(const wxStringImpl& src)
       { return wxString((CtorFromStringImplTag*)nullptr, src); }
+
+  wxString(CtorFromStringImplTag* WXUNUSED(dummy), wxStringImpl&& src) noexcept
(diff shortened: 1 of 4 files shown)


vadz:
Very nice, thanks! Will merge soon.

pavel-t:
Are you interested in backporting this to 3.2? I started working on it before c++11 became a requirement, so the compatibility macros are mostly ready.

vadz:
I can't say that I'm personally interested in this (I mostly use the latest wx version in all applications I work on anyhow...), but I would be happy to apply this to 3.2 too, of course,provided that:

1. It still compiles in C++98 mode (which is probably as simple as adding `wxCHECK_CXX_STD(11)` checks?).
2. It preserves the ABI (which, I think, it should, but I didn't actually check it).

Thanks again!

Issue #23218, opened 2023-02-04
`wxLocale::Init(wxLANGUAGE_DEFAULT)` should fail under Linux if locale is not recognized


vadz:
As discussed in #23114, currently it always succeeds, even if it doesn't do anything because the locale is unknown to the system. This happens both for locales which might be supported, such as `fr_US` or `en_FR` (see #23217) and for the completely bogus locales, e.g. it succeeds even with `LC_ALL=bloordyblop`.

utelle:
> As discussed in #23114, currently it always succeeds, even if it doesn't do anything because the locale is unknown to the system. This happens both for locales which might be supported, such as `fr_US` or `en_FR` (see #23217) and for the completely bogus locales, e.g. it succeeds even with `LC_ALL=bloordyblop`.

IMHO it is a misuse of the environment variables if they are set to values the system doesn't support. The problem in the current wx code is that it is not checked whether the function `setlocale` succeeded or not, when setting the default locale. If `setlocale` fails then reporting `wxLANGUAGE_UNKNOWN` and staying with the C locale seems to be the right thing to do.

vadz:
I agree and pushed the corresponding change to #23226 too.

Issue #23217, opened 2023-02-04
Support mixed language/territory in `LC_ALL`/`LANG` under Linux


vadz:
As discussed in #23114, setting `LC_ALL=en_FR` or `LC_ALL=fr_US` doesn't work at all currently, i.e. neither `wxLocale::Init(wxLANGUAGE_DEFAULT)` nor `wxUILocale::UseDefault()` actually do anything in this case and we just continue using C locale.

It would seem to be better to set it to French locale for `en_FR` -- but keep using English as the language, of course.

A related question is what should be done for `LC_ALL=fr`. Currently this doesn't set the correct locale neither, but it would be more useful to set it to `fr_FR` in this case (or even any supported `fr_XX` if the primary locale is not available).

utelle:
> As discussed in #23114, setting `LC_ALL=en_FR` or `LC_ALL=fr_US` doesn't work at all currently,

These locales are typically unknown, unless the user/administrator of the system configured and installed them.

> i.e. neither `wxLocale::Init(wxLANGUAGE_DEFAULT)` nor `wxUILocale::UseDefault()` actually do anything in this case and we just continue using C locale.

What else could be reasonably done?

> It would seem to be better to set it to French locale for `en_FR` -- but keep using English as the language, of course.

Under Linux there are several environment variables (`LC_*`) that allow to configure various locale settings (number and date formatting and so on).

In the system settings of my system I can select a UI language and different locale settings for number formatting, date formatting and so on. And for all these entries I can choose only from a list of known locales. This results in matching values for the `LC_*` environment variables.

AFAIK the wx code does not distinguish between the various `LC_*` variables. Maybe that should be changed in the long run.

> A related question is what should be done for `LC_ALL=fr`. Currently this doesn't set the correct locale neither,

On most Linux systems `fr` is not a valid locale, because a combination of language and region (like `fr_FR`) is expected. However, our language database contains a reference locale tag for entries with language-only primary tags whereever possible. For example, for the entry `fr` the reference tag is `fr_FR`.

> but it would be more useful to set it to `fr_FR` in this case (or even any supported `fr_XX` if the primary locale is not available).

Using `fr_FR` would be possible via the wx language database - find entry for `fr` and use the reference tag for setting the locale. Choosing anything else would be difficult, unless you check all language database entries that start with the given language code.

vadz:
[...]

PR #23226, opened 2023-02-05
Improvements for handling default locale

vadz:
The changes here fix Mac-specific problems discussed in #23114, i.e.

1. Querying current locale before setting it now returns "C" there, just as everywhere else.
2. `wxLocale::Init(wxLANGUAGE_DEFAULT)` doesn't fail any longer with mixed locales.
3. While `GetSystemLocale()` still returns `wxLANGUAGE_UNKNOWN` for such locales, there is a new `GetSystemLocaleId()` function which allows to get e.g. "en-FR" string for them.

NB: this PR also incorporates #23119 and #23147.

cc @utelle @csomor (Stefan, I've removed Mac-specific code you added back in 93254327a1 (fixing strange bug under OS X (null strings from within CoreFoundation), 2007-12-19), but I'm pretty sure it's the right thing to do)

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -63,6 +63,12 @@ Changes in behaviour not resulting in compilation errors
   mode under MSW, use the new AreAppsDark() or IsSystemDark() to check if the
   other applications or the system are using dark mode.
 
+- wxUILocale::IsSupported() now returns false for unavailable locales under
+  Unix systems without trying to fall back on another locale using the same
+  language in a different region, e.g. it doesn't use fr_FR if fr_BE is not
+  available. If any locale using the given language is acceptable, the region
+  must be left empty, e.g. just "fr" would use any available "fr_XX".
+
 
 Changes in behaviour which may result in build errors
 -----------------------------------------------------
--- a/include/wx/uilocale.h
+++ b/include/wx/uilocale.h
@@ -158,12 +158,18 @@ public:
     // its dtor is not virtual.
     ~wxUILocale();
 
+    // Return the locale ID representing the default system locale, which would
+    // be set is UseDefault() is called.
+    static wxLocaleIdent GetSystemLocaleId();
+
     // Try to get user's (or OS's) preferred language setting.
     // Return wxLANGUAGE_UNKNOWN if the language-guessing algorithm failed
+    // Prefer using GetSystemLocaleId() above.
     static int GetSystemLanguage();
 
     // Try to get user's (or OS's) default locale setting.
     // Return wxLANGUAGE_UNKNOWN if the locale-guessing algorithm failed
+    // Prefer using GetSystemLocaleId() above.
     static int GetSystemLocale();
 
     // Try to retrieve a list of user's (or OS's) preferred UI languages.
--- a/interface/wx/intl.h
+++ b/interface/wx/intl.h
@@ -526,13 +526,15 @@ public:
     /**
         Tries to detect the user's default locale setting.
 
-        Returns the ::wxLanguage value or @c wxLANGUAGE_UNKNOWN if the language-guessing
-        algorithm failed.
+        @note This function is somewhat misleading, as it uses the default
+        system locale to determine its return value, and not just the system
+        language. It is preserved for backwards compatibility, but to actually
+        get the language, and not locale, used by the system by default, call
+        wxUILocale::GetSystemLanguage() instead.
 
-        @note This function works with @em locales and returns the user's default
-              locale. This may be, and usually is, the same as their preferred UI
-              language, but it's not the same thing. Use wxTranslation to obtain
-              @em language information.
+        Returns the ::wxLanguage value or @c wxLANGUAGE_UNKNOWN if the locale
+        is not recognized, as can notably happen when combining any language
+        with a region where this language is not typically spoken.
 
         @see wxTranslations::GetBestTranslation().
     */
--- a/interface/wx/uilocale.h
+++ b/interface/wx/uilocale.h
@@ -289,7 +289,7 @@ public:
               by the operating system (for example, Windows 7 and below), the user's
               default @em locale will be used.
 
-        @see wxTranslations::GetBestTranslation().
+        @see wxTranslations::GetBestTranslation(), GetSystemLocaleId().
     */
     static int GetSystemLanguage();
 
@@ -297,7 +297,8 @@ public:
         Tries to detect the user's default locale setting.
 
         Returns the ::wxLanguage value or @c wxLANGUAGE_UNKNOWN if the locale-guessing
-        algorithm failed.
+        algorithm failed or if the locale can't be described using solely a
+        language constant. Consider using GetSystemLocaleId() in this case.
 
         @note This function works with @em locales and returns the user's default
               locale. This may be, and usually is, the same as their preferred UI
@@ -308,7 +309,19 @@ public:
 
         @see wxTranslations::GetBestTranslation().
     */
-    static int GetSystemLocale();};
+    static int GetSystemLocale();
+
+    /**
+        Return the description of the default system locale.
+
+        This function can always represent the system locale, even when using
+        a language and region pair that doesn't correspond to any of the
+        predefined ::wxLanguage constants, such as e.g. "fr-DE", which means
+        French language used with German locale settings.
+
+        @since 3.3.0
+     */
+};
 
 /**
     Return the format to use for formatting user-visible dates.
--- a/src/common/intl.cpp
+++ b/src/common/intl.cpp
@@ -732,11 +732,36 @@ bool wxLocale::IsAvailable(int lang)
     const wxLanguageInfo *info = wxLocale::GetLanguageInfo(lang);
     if ( !info )
     {
-        // The language is unknown (this normally only happens when we're
-        // passed wxLANGUAGE_DEFAULT), so we can't support it.
-        wxASSERT_MSG( lang == wxLANGUAGE_DEFAULT,
-                      wxS("No info for a valid language?") );
-        return false;
+        // This must be wxLANGUAGE_DEFAULT as otherwise we should have found
+        // the matching entry.
+        wxCHECK_MSG( lang == wxLANGUAGE_DEFAULT, false,
+                     wxS("No info for a valid language?") );
+
+        // For this one, we need to check whether using it later is going to
+        // actually work, i.e. if the CRT supports it.
+        const char* const origLocale = wxSetlocale(LC_ALL, nullptr);
+        if ( !origLocale )
+        {
+            // This is not supposed to happen, we should always be able to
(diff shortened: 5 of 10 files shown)


utelle:
Although I tested only under Windows and Linux up to now, here comes a first feedback. Mac test will follow.

I tested with the _internat_ sample.

- Windows (system language German)
  - With language and locale `de-DE` all results as expected.
  - With language `de-DE` and regional formatting `fr-FR` German translations are used, except for the calendar control where French month and day names are used (but since the system was set to French regional formatting this was to be expected).
  - Testing for availability with `en_DE` results in "locale _en-DE / English (Germany)_ available" - and that's correct.
  - Testing for availability with `fr_DE` results in "locale _fr-DE_ is not available" - and that's correct.

- Linux (system language German)
  - `LC_ALL` **not** set: locale `de_DE.utf-8` is used for UI and C locale, all results as expected (except for a blank field under the heading **Date**)
  - Testing for availability with `en_DE` results in "locale _en_DE_ is available / Identifier _en_GB.utf-8 / English (United Kingdom)_" - and that is somewhat surprising. Why is the locale available, although it can't be set actually? Why is the identifier _en_GB_, although the identifier _en_DE_ is contained in the wx language database?
  - Testing for availability with `fr_DE` results in "locale _fr_DE_ is available / Identifier _fr_FR.utf-8 / French (France)_"" - again that is somewhat surprising. Why is the locale available, although it can't be set actually? Why is the identifier _fr_FR_?

  - `LC_ALL=de_DE`: locale `de_DE` is used for UI and C locale, all results as expected (except for a blank field under the heading **Date**)
  - `LC_ALL=en_DE`:
    - Setting `LC_ALL` like that results in a shell warning _setlocale: LC_ALL: can't change the locale settings (en_DE): File or directory not found
    - The program asks "Would you like to use the program in English (Germany)?" That is expected, because the locale `en_DE` is defined in the wx language database.
    - locale `C` is used for UI and C locale, all results in English as expected (except for a blank field under the heading **Date**)
  - `LC_ALL=fr_DE`:
    - Setting `LC_ALL` like that results in a shell warning _setlocale: LC_ALL: can't change the locale settings (fr_DE): File or directory not found
    - The program asks "Would you like to use the program in English (United States)?" That is expected, because the locale `fr_DE` is **not** defined in the wx language database.
    - locale `C` is used for UI and C locale, all results in English as expected (except for a blank field under the heading **Date**)

  IMHO the locales like `en_DE` and `fr_DE` should be reported as **not available** under Linux, because they can't be set via `setlocale`. If they are accepted, then most likely the expected behaviour would be to use the language part for the UI language and the region part for the regional formatting. However, this doesn't happen, because the actual locale will be _C_.

I will have to look at the code changes for Linux to understand what is going on.

vadz:
Thanks for testing! I think the discussion of support for "mixed" locales under Linux should be restricted to #23217. And there is also the more specific #23218 for the "locale is unexpectedly available" part of the problem.

Other than that it looks like you haven't found any problems, right?

Thanks again!

utelle:
Now I tested also under Mac OS Ventura. Again, I tested with the _internat_ sample.

- system settings `German (Germany)`
  - locale `de_DE` is used for UI, locale `C/UTF-8/C/C/C/C` is used for C locale.
  -  _Window_ menu and all subentries **not** translated (due to commit [f7d0742](https://github.com/wxWidgets/wxWidgets/commit/0a1b0fbe0142128447151c05a13103ed9f1d4fce) - message catalog **wxstd.pot** needs to be regenerated and **all** language catalogs need to be adjusted),
  - the calendar control uses **English** month and day names, numbers correctly use comma as decimal separator,
  -  the date field under the heading **Date** is blank instead of showing the current date,
  -  the locale name in English shows _Deutsch (Deutschland)_ instead of _German (Germany)_ - the required changes to `src/osx/core/uilocale.mm` were not yet applied.
  - Testing for availability with `en_DE` results in "locale _en_DE_ is available / Identifier _en_DE / English (Germany)_"
  - Testing for availability with `fr_DE` results in "locale _fr_DE_ is **not** available_"

- system settings `German (France)`
  -  locale `de_FR` is used for UI, locale `C/UTF-8/C/C/C/C` is used for C locale.
  - Testing for availability with `de_FR` results in "locale _de_FR_ is **not** available_" - although it corresponds to the system settings,
  - for locale name and calendar control same behaviour as above

- system settings `French (Germany)`
  -  locale `fr_DE` is used for UI, locale `C/UTF-8/C/C/C/C` is used for C locale.
  - Testing for availability with `fr_DE` results in "locale _fr_DE_ is **not** available_" - although it corresponds to the system settings,
  - for locale name and calendar control same behaviour as above

- system settings `English (Germany)`
  -  locale `en_DE` is used for UI, locale `C/UTF-8/C/C/C/C` is used for C locale.
  - Testing for availability with `en_DE` results in "locale _en_DE_ is available / Identifier _en_DE / English (Germany)_"
  - for locale name and calendar control same behaviour as above
  -  the date field under the heading **Date** is correctly shown formatted as a German date,

I find it interesting that the test for availability fails for `de_FR` and `fr_DE`, although the default locale was reported with exactly these values for the corresponding test cases. One would expect that a locale that is reported as the default locale would also be reported as available.

utelle:
> I think the discussion of support for "mixed" locales under Linux should be restricted to #23217. And there is also the more specific #23218 for the "locale is unexpectedly available" part of the problem.

IMHO under Linux the problem is that it is not detected when `setlocale` fails - the locale is reported as available nevertheless.

> Other than that it looks like you haven't found any problems, right?

There are a few minor issues:

- Linux: locales are reported as available, even if `setlocale` fails for them
- Linux and MacOS: the date field under the heading **Date** is blank quite often. If the language is English the current date is usually shown correctly.
- MacOS: method `GetLocalizedNames` not yet adjusted,
- MacOS: Window menu entries not translated
- MacOS: calendar control always uses English month and day names
- MacOS: some mixed locales reported as not available, even if the default locale is the same
- MacOS: all wxstd message catalogs need to be regenerated/adjusted

vadz:
Concerning this part:

> Linux: locales are reported as available, even if `setlocale` fails for them

This is due to the code in `TryCreateMatchingLocale()` which explicitly looks for any locale with the same language, even if the region is different. I admit I have no idea why it was done, in spite of doing this myself in 3f8fd90c31 (Support creating Unix wxUILocale for languages without region, 2021-09-01) -- but the actual code doesn't do what the commit says, as it also creates locales even if we do have a different region.

Thinking more about this, I believe the idea here was to allow choosing a different region using the same language, i.e. if `fr_BE` is not supported, fall back on `fr_FR` if it is. This is probably still a good idea (although I'm not even 100% sure about this), but using `en_US` as fallback for `en_FR` almost surely is not, right?

utelle:
> Concerning this part:
> 
> > Linux: locales are reported as available, even if `setlocale` fails for them
> 
> This is due to the code in `TryCreateMatchingLocale()` which explicitly looks for any locale with the same language, even if the region is different. I admit I have no idea why it was done, in spite of doing this myself in [3f8fd90](https://github.com/wxWidgets/wxWidgets/commit/3f8fd90c31e290c8fdd6fbf5370f03ce5b1feea7) (Support creating Unix wxUILocale for languages without region, 2021-09-01) -- but the actual code doesn't do what the commit says, as it also creates locales even if we do have a different region.

IMHO a failing call to `setlocale` is a good indicator that a locale shoud be reported as "not available". 

AFAICR the approach in pre-wxUILocale times was similar. However, I find this approach questionable. I didn't change it to not break wxLocale compatibility, but regarding wxUILocale I would rather drop this fallback "strategy".

I'm not sure what the actual expectations of a user are, when he/she uses a "mixed" locale of which it is known that it is not supported by the OS.

The reason for setting a locale like `fr_US` or `de_US` is usually that the user is located in the given region, but doesn't speak well the region's main language and therefore prefers to use his native language for the UI, but a regional formatting, so that other people do not misinterpret dates and numbers.

> Thinking more about this, I believe the idea here was to allow choosing a different region using the same language, i.e. if `fr_BE` is not supported, fall back on `fr_FR` if it is. This is probably still a good idea (although I'm not even 100% sure about this), but using `en_US` as fallback for `en_FR` almost surely is not, right?

Well, I'm not really convinced that falling back on a locale with the same language, but a different region is such a good idea. Although it seems to be rather plausible to use `fr_FR` instead of `fr_BE` (France and Belgium are neighboring countries after all), it is absolutely not plausible to fall back from `en_DE` to `en_GB` - date and number formatting are very different in Germany and the United Kingdom.

Under Windows many language/region combinations are predefined, but not all. Nevertheless, you can set the locale to almost any combination programmatically, even though the settings dialog doesn't offer it. Windows "constructs" internally a locale that tries to match the language and the region as close as possible. For example, for `fr_US` it would use the French language, but date and number formatting would follow the US rules. This is far better than falling back to a different locale replacing the region in a somewhat obscure manner. However, our code under Windows reports "constructed" locales as "not available". And IMHO that is a good decision.

I think the most logical approach is to allow for date and number formatting (and similar locale dependent things) to set a locale that is actually supported by the OS, and to let the user choose a different language for the UI. This should cover most use cases.

vadz:
The trouble is that I'd like to apply these fixes to 3.2 too and removing this fallback would definitely be a compatibility breakage, so I don't want to do it there. But I agree that we should stop doing this in master, so I've added a commit doing this.

vadz:
Finally, concerning the locales availability under Mac: it turns out `en-FR` is supported by Cocoa, i.e. it's one of the locales returned by `[NSLocale availableLocaleIdentifiers]`, while neither `fr-DE` nor `de-FR` are in [this list](https://github.com/wxWidgets/wxWidgets/files/10670079/macos-12-locales.txt).

I don't know if it's worth handling the current locale specially, maybe it is, but for now I won't do it.

utelle:
> Finally, concerning the locales availability under Mac: it turns out `en-FR` is supported by Cocoa, i.e. it's one of the locales returned by `[NSLocale availableLocaleIdentifiers]`, while neither `fr-DE` nor `de-FR` are in [this list](https://github.com/wxWidgets/wxWidgets/files/10670079/macos-12-locales.txt).

Question: Should we update our language database to include the locales from the above list that are not yet included?

> I don't know if it's worth handling the current locale specially, maybe it is, but for now I won't do it.

Since on MacOS there is no way to set (`Use()`) a locale other than the default locale, the user can always access the current locale to retrieve the information. So, for now it is ok as it is.

vadz:
> Question: Should we update our language database to include the locales from the above list that are not yet included?

I don't know how does macOS determine which locales to support, so probably not -- as I don't see how are we going to maintain it in the future. But if anybody has a source for this data, then perhaps we should...

vadz:
I'm going to merge this without waiting for more reviews/testing because I want to backport it to 3.2 and for this it should be merged into master first, but any feedback would be still very welcome even later!

PR #23229, opened 2023-02-06
Update message catalogs

utelle:
At least commit f7d0742715d16a76bf7fbe125484e5dead0754ea added several messages with context. To make use of their translations it is necessary to regenerate wxstd.pot and update all .po files.

This PR updates wxstd.pot and all .po files. Additionally, the translations of the German message catalog have been updated.


vadz:
Thanks Ulrich! Will merge in a moment.

PR #23234, opened 2023-02-07
Update French translations of the internat sample

vadz:
I almost never use any programs in French, so I'm not sure if the translations here are really the best, and so would be grateful for any corrections. TIA!


PBfordev:
Sorry, I am not going to correct since I don't know any.

I just wanted to ask when are the sample's .po files regenerated? I took a look at the Czech one (CS) and it seems out of sync with the code. Its header is also unfilled and POEdit warned me that it uses "unusual plural forms expression" (and it is different than in `locale/cs.po`), so it probably was not updated in a while.

Perhaps the sample catalogs should be now regenerated too?

vadz:
> I just wanted to ask when are the sample's .po files regenerated?

They aren't :-(

> I took a look at the Czech one (CS) and it seems out of sync with the code. Its header is also unfilled and POEdit warned me that it uses "unusual plural forms expression" (and it is different than in `locale/cs.po`), so it probably was not updated in a while.
> 
> Perhaps the sample catalogs should be now regenerated too?

The README in the sample directory contains the instructions about doing this, but I'm not even sure if this is the best way to do it. To be honest I don't understand the difference between xgettext `-j` option and msgmerge and which one should be preferred, but it seems weird to use one of them here and the other one in `locale/Makefile`.

vadz:
I'm merging this because I'd also like to include this in the upcoming 3.2.2, but any corrections from French users would be still welcome.

PR #23235, opened 2023-02-07
Update Czech message catalog

PBfordev:
~~The matching cs.mo needs to be created.~~

~~I wonder if it could be (realistically) possible to somehow create an automated (GitHub?) task updating the .mo file when its .po file was changed...~~

The macOS Window menu translations are just guesses (I don't have a macOS computer and cursory google search did not find any images with the the translated menu) but I think at least they are not horribly wrong.

Edit: I did not realize the GIT does not have the .mo files at all. BTW, perhaps we could change Cirrus CI settings so that mere a change in a message catalog does not trigger the full build?


vadz:
Applied, thanks.

Also updated Cirrus CI configuration file in 16522b05a83b7a609ab2440d84d6737763ebd921.

Issue #23240, opened 2023-02-08
wxOSXDrawNSImage ignores the current wxCompositionMode


krZll:
Composition mode `wxCOMPOSITION_XOR ` has no effect if an image is drawn using function `wxMacCoreGraphicsContext::DrawBitmap( const wxBitmap &bmp, wxDouble x, wxDouble y, wxDouble w, wxDouble h )` in grahics.cpp with `wxOSX_USE_COCOA `enabled.
You can see this in the sample '**drawing**' if you select '**Mask screen**' under '**Screen**'. The image declared as `wxXOR` is the same as `wxCOPY`. The relevant function in the example is `MyCanvas::DrawImages(wxDC& dc, DrawMode mode)`. Images are drawn onto a `dc `with various raster operations, including `wxXOR`. When the old branch of `wxMacCoreGraphicsContext::DrawBitmap` is called the view is fine.

Tested with v3.1.5 and v3.2.1 on macOS Big Sur

vadz:
`wxOSXDrawNSImage()` hardcodes `operation:NSCompositeSourceOver` in its call to `[NSImage drawInRect:...]`, see

https://github.com/wxWidgets/wxWidgets/blob/620e02c568bb7cc410b10ede1b592c20c02dfb9f/src/osx/carbon/utilscocoa.mm#L307

so I think we need to pass it the correct composition mode to use to fix this.

@csomor What do you think?

csomor:
@vadz, thanks I'll add a composition parameter

csomor:
@krZll could you please test with #23245

PR #23245, opened 2023-02-09
OSX adding composition do wxOSXDrawNSImage

csomor:
see #23240


vadz:
Thanks! Applied to master, see the commit referenced above, and will backport to 3.2 a bit later.

csomor:
> will backport to 3.2 a bit later.

@vadz I had to add an additional parameter, for backwards compatibility should add a default version with one parameter less ?

vadz:
I thought `wxOSXDrawNSImage()` was a private function, but now I see that it's indeed exported... I guess we should indeed add an overload without the new parameter to be safe, thanks for noticing this.

But I also wonder if we shouldn't stop exporting it in master, or is it really useful in the application code?

vadz:
Backported with the extra ABI fix to 3.2, thanks!

PR #23246, opened 2023-02-10
fix non-Unicode issues in src/msw/main.cpp

brechtsanders:
Fix build non-Unicode build errors in `src/msw/main.cpp`

### DIFF
--- a/src/msw/main.cpp
+++ b/src/msw/main.cpp
@@ -193,7 +193,11 @@ int wxEntry(int& argc, wxChar **argv)
 WXDLLIMPEXP_BASE void wxMSWCommandLineInit();
 WXDLLIMPEXP_BASE void wxMSWCommandLineCleanup();
 WXDLLIMPEXP_BASE int& wxMSWCommandLineGetArgc();
+#if wxUSE_UNICODE
 WXDLLIMPEXP_BASE wchar_t** wxMSWCommandLineGetArgv();
+#else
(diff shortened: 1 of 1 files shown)


PBfordev:
Non-Unicode build is not supported in wxWidgets 3.3, see e.g. https://github.com/wxWidgets/wxWidgets/commit/fec8c06a53f5f2eb65e3328eadddd22334e7971f, so `wxUSE_UNICODE` is always defined as `1`.

AFAIK, UTF-8 build (`wxUSE_UNICODE_UTF8`) is not working either on MSW, so could you please explain what is the purpose of this PR?

brechtsanders:
Sorry, I wasn't aware this was no longer supported.

I have always been able to build non-unicode builds with MinGW-w64 up to version 3.2.1.

The reason I do this is because I port a lot of software to Windows, and sometimes this software just isn't made for Unicode.

Anyway, with the proposed patch it still builds just fine.

PBfordev:
> Anyway, with the proposed patch it still builds just fine.

I understand where you are coming from but once again, non-Unicode builds are no longer supported after wxWidgets 3.2.x, `wxUSE_UNICODE` is always defined with a value of 1:https://github.com/wxWidgets/wxWidgets/blob/6674ce7fe6f8407b077350105bfb4ef0d3658c41/include/wx/platform.h#L193-L199

PR #23253, opened 2023-02-12
Fix wxFLEX_GROWMODE_ALL in wxFlexGridSizer with proportions

### LINKED ISSUE #23251: wxFlexGridSizer ignores proportion in the Flexible Direction if NonFlexibleGrowMode is set
### Description

Not sure if this is a bug, or just something not mentioned in the documentation... but does look like a bug to me.

Consider this example:

```cpp
#include <wx/wx.h>

class MyApp : public wxApp
{
public:
    virtual bool OnInit();
};

wxIMPLEMENT_APP(MyApp);

{
public:
    MyFrame(const wxString &title, const wxPoint &pos, const wxSize &size);
};

bool MyApp::OnInit()
{
    MyFrame *frame = new MyFrame("Hello World", wxDefaultPosition, wxDefaultSize);
    frame->Show(true);
    return true;
}

MyFrame::MyFrame(const wxString &title, const wxPoint &pos, const wxSize &size)
    : wxFrame(nullptr, wxID_ANY, title, pos, size)
{
    const int COLS = 4;
    const int ROWS = 3;

    auto sizer = new wxFlexGridSizer(ROWS, COLS, 0, 0);

    sizer->AddGrowableRow(1, 1);
    sizer->AddGrowableRow(2, 4);

    sizer->AddGrowableCol(1, 1);
    sizer->AddGrowableCol(3, 3);

    sizer->SetFlexibleDirection(wxHORIZONTAL);
    sizer->SetNonFlexibleGrowMode(wxFLEX_GROWMODE_SPECIFIED);

    for (int i = 0; i < ROWS; i++)
    {
        for (int j = 0; j < COLS; j++)
        {
            auto square = new wxCheckBox(this, wxID_ANY, wxString::Format("%d, %d", i, j), wxDefaultPosition, wxDefaultSize);

            bool isDark = i % 2 == j % 2;

            square->SetBackgroundColour(isDark ? *wxBLACK : *wxBLUE);
            sizer->Add(square, 1, wxEXPAND);
        }
    }

    this->SetSizer(sizer);
}
```

This looks good: the rows and cols are resized and the proportions are respected.

Screenshot 2023-02-11 at 13 56 27

Now if we change the non-flexible grow mode:

```cpp
sizer->SetNonFlexibleGrowMode(wxFLEX_GROWMODE_ALL);
```

The non-flexible direction (vertical) grows proportionally, and this is expected. However, the flexible direction **ALSO** grows proportionally, ignoring the proportions!

Screenshot 2023-02-11 at 13 56 44

I guess the `SetNonFlexibleGrowMode` should only affect the non-flexible direction. If that's the case, it's a bug.

Similarly for  `wxFLEX_GROWMODE_NONE` -- the proportions for the flexible direction are ignored.

- wxWidgets version  you use: current master
- wxWidgets port  you use: wxMSW, wxGTK, wxOSX

vadz:
Setting wxFLEX_GROWMODE_ALL for the non-flexible direction broke handling of proportions in the flexible direction which were unexpectedly not taken into account at all any more due to not passing them to DoAdjustForGrowables().

Fix this by still respecting the proportions in this case and add a test case to ensure this doesn't get broken again.

Closes #23251.

### DIFF
--- a/src/common/sizer.cpp
+++ b/src/common/sizer.cpp
@@ -2104,16 +2104,23 @@ void wxFlexGridSizer::AdjustForGrowables(const wxSize& sz, const wxSize& origina
     }
 #endif // wxDEBUG_LEVEL
 
+    // Use growable columns proportions if we are flexible in this direction.
+    const wxArrayInt* const growableColsProportions =
+        (m_flexDirection & wxHORIZONTAL) || (m_growMode == wxFLEX_GROWMODE_SPECIFIED)
+            ? &m_growableColsProportions
+            : nullptr;
 
-    if ( (m_flexDirection & wxHORIZONTAL) || (m_growMode != wxFLEX_GROWMODE_NONE) )
+    // And do anything at all with the columns if we're either flexible or must
+    // resize all columns uniformly (otherwise we use wxFLEX_GROWMODE_NONE and
+    // there is nothing to do).
+    if ( growableColsProportions || (m_growMode == wxFLEX_GROWMODE_ALL) )
     {
         DoAdjustForGrowables
         (
             sz.x - minSize.x,
             m_growableCols,
             m_colWidths,
-            m_growMode == wxFLEX_GROWMODE_SPECIFIED ? &m_growableColsProportions
-                                                    : nullptr
+            growableColsProportions
         );
 
         // This gives nested objects that benefit from knowing one size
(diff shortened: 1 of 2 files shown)


vadz:
Applied to master, will also backport to 3.2.

PR #23256, opened 2023-02-13
configure.in: Fix sockets/webRequest build error

bogdro:
Fix build error when sockets are disabled at configure time but webRequest isn't:
```
wxWidgets-3.2.2/src/common/webrequest_curl.cpp:451:23: error: ‘wxSOCKET_T’ has not been declared
  451 |     bool StartPolling(wxSOCKET_T, int);
      |                       ^~~~~~~~~~
```
Since webrequest requires sockets, it needs to be disabled at configure time when sockets are disabled (the required header is there, so it's not the problem, the problem is that wxUSE_SOCKETS = 0 and the declarations are skipped).
Bug present also at least in 3.2.0 and 3.2.1, as far as I remember (maybe also earlier).
Feel free to modify the message shown to the user.

Configuration that causes the problem to appear follows. It's VeraCrypt (https://veracrypt.fr/en/) and its Makefile that cause the error to appear when called as `make NOGUI=1 WXSTATIC=1 WX_ROOT=/path/to/wxWidgets-3.2.2 wxbuild`.
```
--disable-gui --disable-debug_flag --disable-debug_gdb --disable-debug_info
 --enable-unicode -disable-shared --disable-dependency-tracking
 --enable-exceptions --enable-std_string --enable-dataobj
 --enable-mimetype --disable-protocol --disable-protocols --disable-url --disable-ipc
 --disable-sockets --disable-fs_inet --disable-ole --disable-docview --disable-clipboard
 --disable-help --disable-html --disable-mshtmlhelp --disable-htmlhelp
 --disable-mdi --disable-metafile --disable-webview --disable-xrc --disable-aui
 --disable-postscript --disable-printarch --disable-arcstream --disable-fs_archive
 --disable-fs_zip --disable-tarstream --disable-zipstream --disable-animatectrl
 --disable-bmpcombobox --disable-calendar --disable-caret --disable-checklst
 --disable-collpane --disable-colourpicker --disable-comboctrl --disable-datepick
 --disable-display --disable-dirpicker --disable-filepicker --disable-fontpicker
 --disable-grid --disable-dataviewctrl --disable-listbook --disable-odcombobox
 --disable-sash --disable-searchctrl --disable-slider --disable-splitter
 --disable-togglebtn --disable-toolbar --disable-tbarnative --disable-treebook
 --disable-toolbook --disable-tipwindow --disable-popupwin --disable-commondlg
 --disable-aboutdlg --disable-coldlg --disable-finddlg --disable-fontdlg
 --disable-numberdlg --disable-splash --disable-tipdlg --disable-progressdlg
 --disable-wizarddlg --disable-miniframe --disable-splines --disable-palette
 --disable-richtext --disable-dialupman --disable-debugreport --disable-filesystem
 --disable-rearrangectrl --disable-treelist --disable-richmsgdlg --disable-richtooltip 
 --disable-propgrid --disable-stc --without-libnotify --without-gtkprint
 --without-gnomevfs --disable-fsvolume --disable-fswatcher --disable-sound
 --disable-mediactrl --disable-joystick --disable-apple_ieee --disable-gif
 --disable-pcx --disable-tga --disable-iff --disable-gif --disable-pnm
 --disable-svg --without-expat --without-libtiff --without-libjpeg
 --without-libpng -without-regex --without-zlib
```

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -5966,6 +5966,12 @@ if test "$wxUSE_FS_INET" = "yes"; then
 fi
 
 if test "$wxUSE_WEBREQUEST" = "yes"; then
+    if test "$wxUSE_SOCKETS" != "yes"; then
+        AC_MSG_WARN(wxWebRequest classes require sockets... disabled)
+        wxUSE_WEBREQUEST=no
+    fi
+fi
+if test "$wxUSE_WEBREQUEST" = "yes"; then
     if test "$wxUSE_LIBCURL" = "yes"; then
         AC_DEFINE(wxUSE_WEBREQUEST_CURL)
         have_webrequest_backend=1


vadz:
Thanks for reporting this and for the fix, but `wxWebRequest` doesn't _actually_ require sockets, so I think it would be better to just use `curl_socket_t` instead of `wxSOCKLEN_T` here, see #23257 doing this.

Do you see any problems with doing it like this? Of course, you may want to add `--disable-webrequest` to your configure command line anyhow if you're not using it.

Also, if you want to disable as much of optional functionality as possible, you should probably consider using `--disable-all-features`.

bogdro:
Hi.
Thanks for the reply.
I don't know the internals of wxWidgets, I just saw a build error and thought I could help in fixing it.
If it's more proper to use `curl_socket_t`, feel free to do so. I don't see a problem with that, and I don't mind. However, what I saw in the `configure.in` file, just below my fix, is that there may be more "back-ends", so maybe `curl_socket_t` (cURL-specific, I presume) won't actually always be available...
Like I said - I saw a problem and tried to fix it "my way" (which doesn't always need to be "the right way" :) ).
As for the command line, it's VeraCrypt's, not mine. Let's just gently say that their build system "is not perfect" (basically fails on everything except for wxWidgets 3.0.0 built "their way")...

vadz:
Replaced by #23257.

PR #23260, opened 2023-02-14
Stop requesting multi-sampling in `wxGLCanvas` by default and other minor improvements

vadz:
With this change the cube sample can now draw the cube in a VM too.

If we ever decide to revert this later, only 46857a73ba82bb09fcc6bb42d3d4c886354b9243 will need to be reverted, the other commits don't break compatibility and still improve things.


vadz:

PR #23261, opened 2023-02-14
Simplified implementation of native `wxOverlay` for wxMSW

vadz:
See #23058.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -4976,7 +4976,8 @@ COND_TOOLKIT_MSW___LOWLEVEL_SRC_OBJECTS =  \
 	monodll_uuid.o \
 	monodll_msw_evtloop.o \
 	monodll_access.o \
-	monodll_msw_bmpbndl.o
+	monodll_msw_bmpbndl.o \
+	monodll_msw_overlay.o
 @COND_TOOLKIT_MSW@__LOWLEVEL_SRC_OBJECTS = $(COND_TOOLKIT_MSW___LOWLEVEL_SRC_OBJECTS)
 @COND_TOOLKIT_OSX_COCOA@__LOWLEVEL_SRC_OBJECTS = \
 @COND_TOOLKIT_OSX_COCOA@	$(__OSX_LOWLEVEL_SRC_OBJECTS)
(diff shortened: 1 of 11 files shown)


AliKet:
Thanks for taking care of this, this version is much more clean and clear and I'll test it tonight, thank you!

AliKet:
I think I know why there is a flicker (I mentioned here https://github.com/wxWidgets/wxWidgets/pull/23058#discussion_r1056949817) when the overlay window is shown for the first time:
it's due to the **default background colour** of the overlay window which is **white**.
We need to apply something like this:
```
index 78b13e38a9..c6d9766bac 100644
--- a/src/msw/overlay.cpp
+++ b/src/msw/overlay.cpp
@@ -32,6 +32,8 @@ wxWindow* wxCreateOverlayWindow(const wxRect& rect)
                     WS_EX_NOACTIVATE
                 );
 
+    overlayWin->SetBackgroundColour(*wxBLACK);
+
     if ( !::SetLayeredWindowAttributes(GetHwndOf(overlayWin), 0, 128,
                                        LWA_COLORKEY | LWA_ALPHA) )
     {
```
since `LWA_COLORKEY` (the transparent colour) is choosen to be black.

Other than that, everything is working correctly for me.

Thanks again!

vadz:
Thanks, I didn't notice the flicker, somehow -- maybe due to enabling compositing in the meanwhile?

Anyhow, setting the background colour makes sense, but I guess we could also just use `wxBG_STYLE_TRANSPARENT`?

AliKet:
> Anyhow, setting the background colour makes sense, but I guess we could also just use `wxBG_STYLE_TRANSPARENT`?

Doesn't work under my system (Window 10 22H2 32bits), and I get this message (returned by `IsTransparentBackgroundSupported()`)

`Can't use transparent background: This platform does not support background transparency.`

???

vadz:
I do see the flicker after disabling compositing (i.e. with `wx_msw_window_no_composited=1` in the environment), but setting the background style to `wxBG_STYLE_PAINT` (sorry, not `TRANSPARENT`, of course...) gets rid of it and seems preferable to changing the background colour as it's better not to draw the background at all rather than draw it with the transparent colour.

If you don't see any problems with doing it like this, I'm going to merge it soon.

Thanks again!

AliKet:
Works correctly for me now, thanks!

PR #23265, opened 2023-02-16
Use © instead of (c) as the copyright sign in wxGetLibraryVersionInfo()

lanurmi:


src/common/utilscmn.cpp
```diff
@@ -1431,12 +1431,15 @@ wxVersionInfo wxGetLibraryVersionInfo()
...
+    const wxString copyrightSign = wxString::FromUTF8("\xc2\xa9");
+
     return wxVersionInfo(wxS("wxWidgets"),
                          wxMAJOR_VERSION,
                          wxMINOR_VERSION,
                          wxRELEASE_NUMBER,
                          msg,
-                         wxS("Copyright (c) 1992-2023 wxWidgets team"));
+                         wxString::Format(wxS("Copyright %s 1992-2023 wxWidgets team"),
+                                          copyrightSign));
```
vadz, line 1442: Sorry, I don't understand how is this better than just
```suggestion
                         wxString::FromUTF8("Copyright \xc2\xa9 1992-2023 wxWidgets team"));
```
?

lanurmi: It is only better in the sense that the variable name documents what the hex sequence stands for, but admittedly that is quite easy to guess in this context.

vadz: OK, fair enough, it is indeed more clear, and we don't care about the extra overhead of `Format()` here nor, probably, about a few extra bytes it adds to the library... Will apply as is, thanks.

vadz:
Applied to master now, thanks.

Issue #23210, opened 2023-02-02
setlocal for Ukrainian language


YuSanka:
### Description

From version 3.1.6 wxWidgets calls `setlocation(0, wxInfoLanguage->LocaleTag)`, see (https://github.com/prusa3d/wxWidgets/commit/deef116a09748796711d1e3509965ee208dcdf0b#diff-7de25e9a71c4dce61bbf76492c589623d5b93fd1bb105ceaf0662075d15f4472),
where LocaleTag is a Tag of locale in BCP 47 - like notation.
For Ukrainian Language LocaleTag is "uk".
But setlocale(0, "uk") returns "English_United Kingdom.1252" instead of "uk",
and, as a result, locales are set to English_United Kingdom        
    
We use such workaround in PS now. After initialization of the wxLocale, for Ukrainian we call `setlocal `one more time with **GetCanonicalWithRegion()** data.

 ```
if (language_info->CanonicalName == "uk")
          setlocale(0, language_info->GetCanonicalWithRegion().data());
```

But I suppose, that this issue have be resolved in wxWidgets internally

### Platform and version information

- wxWidgets version  you use: 3.2.0
- wxWidgets port  you use: any
- OS  and its version: any

**Comment by vadz** [comment]
It would be really great if you could please provide an example of reproducing the problem because while I understand that the issue has something to do with the confusion between the language and country codes, I have no idea how to actually trigger it.

Please show some minimal code or, again, ideally a patch to the internat sample.

**Comment by vadz** [comment]
I can't reproduce the problem in the internat sample, e.g. here is what it shows under MSW:
```
[Window Title]
Internat Information

[Content]
Locale "uk_UA" is available.
Identifier: uk-UA; Layout: LTR
English name: Ukrainian (Ukraine)
Localized name: українська (Україна)

[OK]
```

Please explain what doesn't work exactly for you and, also whether the changes of #23226 change anything for you. TIA!

**Comment by YuSanka** [comment]
I'm sorry to be late with the reply.

For the repro you can just use your "calendar" sample, but make a small change
```
MyApp::MyApp() :
    // Locale affects on the language used in the calendar, and may affect
    // on the first day of the week.
    m_locale(wxLANGUAGE_UKRAINIAN) //m_locale(wxLANGUAGE_DEFAULT)
{
} 
```
1. Start "Calendar wxWidgets sample" application 
2. Select menu "Calendar" > "Use generic version"
3. You coudn's see any effect, calendar is in English 
[...]

PR #23269, opened 2023-02-17
Fix setting locale to `wxLANGUAGE_UKRAINIAN` (#23210)

utelle:
Use GetCanonicalWithRegion() (instead of LocaleTag)  in the Windows implementation of method wxLanguageInfo::TrySetLocale() - as is done in the non-Windows implementation of this method. (see https://github.com/wxWidgets/wxWidgets/issues/23210)


vadz:
Thanks for the fix! I agree that this is the right thing to do and will apply this to both master and 3.2 soon.

vadz:
Applied to master now and my 3.2-proposed-backports branch, thanks again!

Issue #23271, opened 2023-02-17
wxAuiToolBar::AddTool()  client_data parameter is discarded in code


CBtiger:
Describe the bug:
The wxAui function
 `  wxAuiToolBarItem* wxAuiToolBar::AddTool( int toolId, ...., wxObject *  client_data ) 	`
is described on the documentation as able to store a client_data pointer of type `wxObject*`- similar to `wxToolBar`. Here is the link to that documentation  https://docs.wxwidgets.org/trunk/classwx_aui_tool_bar.html#aa52bcc769ee06a6267b4ee030f76972a

Expected vs observed behaviour:
Thecode should store this pointer. But the source code (trunk) has this function listed as
   ` wxAuiToolBarItem* wxAuiToolBar::AddTool(int tool_id, .... , wxObject* WXUNUSED(client_data))	`
Here the source link. https://github.com/wxWidgets/wxWidgets/blob/master/src/aui/auibar.cpp#988
Also there is no function given to retrieve this `wxObject*` pointer

Patch or snippet allowing to reproduce the problem:
n/a

### Description
The wxAuiToolBar::AddTool() parameter "`client_data`" is ignored in the source code. That is not mentioned in the documentation.
* least fix should be to update the documentation and tell the user that the parameter `client_data` is discarded at this time.
*  better fix is to save that parameter and provide a way to retrieve that data again (this is missing). That would make sense, because he same principle is used by the native `wxToolBar`.

### Platform and version information
wxAui - All versions, all ports.

vadz:
There is indeed no support for the client data in this class, I'll mention it in the (equally non existent...) documentation.

Any PRs implementing this would be welcome, of course,

wh11204:
I have just created a PR fixing this. It is my first PR, I hope it it OK.

wh11204:
I deleted the branch and the PR has closed itself, that is not what I intended to do. Lesson learned...

PR #23276, opened 2023-02-20
Add support for client data to wxAuiToolBar

wh11204:
wxAuiToolBar::AddTool() discards client data parameter (issue #23271).

### DIFF
--- a/include/wx/aui/auibar.h
+++ b/include/wx/aui/auibar.h
@@ -249,8 +249,8 @@ private:
     bool m_active;               // true if the item is currently active
     bool m_dropDown;             // true if the item has a dropdown button
     bool m_sticky;               // overrides button states if true (always active)
-    long m_userData;             // user-specified data
-    wxObject* m_clientData;      // client data
+    long m_userData;             // number associated with the item
+    wxObject* m_clientData;      // pointer to a wxObject associated with the item
     int m_alignment;             // sizer alignment flag, defaults to wxCENTER, may be wxEXPAND or any other
 };
 
@@ -566,7 +566,7 @@ public:
     void SetMargins(int x, int y) { SetMargins(x, x, y, y); }
     void SetMargins(int left, int right, int top, int bottom);
 
-    void SetToolClientData (int tool_id, wxObject *client_data);
+    void SetToolClientData (int tool_id, wxObject* client_data);
     wxObject* GetToolClientData(int tool_id) const;
 
     void SetToolBitmapSize(const wxSize& size);
--- a/include/wx/aui/auibar.h
+++ b/include/wx/aui/auibar.h
@@ -121,6 +121,7 @@ public:
         m_dropDown = true;
         m_sticky = true;
         m_userData = 0;
+        m_clientData = nullptr;
         m_alignment = wxALIGN_CENTER;
     }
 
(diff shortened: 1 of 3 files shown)


include/wx/aui/auibar.h
```diff
@@ -560,6 +566,9 @@ class WXDLLIMPEXP_AUI wxAuiToolBar : public wxControl
     void SetMargins(int x, int y) { SetMargins(x, x, y, y); }
     void SetMargins(int left, int right, int top, int bottom);
 
+    void SetToolClientData (int tool_id, wxObject *client_data);
+    wxObject* GetToolClientData(int tool_id) const;
+
```
vadz, line 571: Do we really need those? It looks like it would be enough to use `FindTool()` and similar (`ByPosition/Index` variants) and then call `{Get,Set}ClientData()` on the returned pointer, just as it's already apparently supposed to be done with `{Get,Set}UserData()`.

include/wx/aui/auibar.h
```diff
@@ -244,7 +249,8 @@ class WXDLLIMPEXP_AUI wxAuiToolBarItem
     bool m_active;               // true if the item is currently active
     bool m_dropDown;             // true if the item has a dropdown button
     bool m_sticky;               // overrides button states if true (always active)
-    long m_userData;            // user-specified data
+    long m_userData;             // user-specified data
+    wxObject* m_clientData;      // client data
```
vadz, line 253: It looks like they're supposed to be used for the same thing and I'd at least describe them better, e.g.
```suggestion
    long m_userData;             // numeric user-specified data
    wxObject* m_clientData;      // pointer to user-specified data
```

But it's still somewhat of a mess :-(

vadz (commented):
Thanks, looks good but you also need to add the new declarations to `interface/wx/aui/auibar.h`. Please use `@since 3.3.0` in their descriptions to indicate that they're new in this version.

include/wx/aui/auibar.h
wh11204: I added them because wxToolBar has them, to keep coherency.

include/wx/aui/auibar.h
CBtiger: What about: "alternative client data, compatible with wxToolBar" for the wxObject* 

Just some feedback on the status quo from a user perspective: I am using those long values as workaround as the orginal `m_clientData` does not work and it works fine. 
There is indeed potential confusion if both exist and they are not the same, but imho the alternative is even more confusing. As long as this is mentioned in the documentation this is one of the more easy traps to master when using AUI, like you can not manage wxFrame or wxDialog derived stuff its just crashing (wasn't obvious to me). There are other things confusing and annoying like different named functions doing the same thing, e.g. `aui::GetToolCount() `and `wx::GetToolsCount()`. But some things are better with AUI, e.g. there is actually a documented Tool class. wxToolBarToolBase is not documented at all (using 3.0.5). 
I just wrote a common base class to those two Toolbars for own additions and I had introduce about 10 virtual functions that were basically doing the same thing but slightly different between the two.

vadz: FWIW we could also change the type of `m_userData` to `wxUIntPtr` and use it to store either long or a pointer. This would be type-unsafe, of course, but then so is using `wxObject*` in the first place anyhow...

vadz:
Thanks, (squash) merged into master now with some cosmetic changes.

CBtiger:
Perfect. Thanks a lot!

PR #23278, opened 2023-02-20
Fix wxCollapsiblePane background inside wxNotebook under MSW

### LINKED ISSUE #23274: wxCollapsiblePane background color does not match parent's background
### Description
On MSW, when you have a CollapsiblePane inside a Notebook page, the background color of the CollapsiblePane contrasts with the background color of the Notebook

image

Expected vs observed behaviour:
I would expect that the colors would match, or at least there would be an easy way to set the background colors to match, but I cannot do it.
I've found a similar problem described here: https://forums.wxwidgets.org/viewtopic.php?t=45014 , but no easy solution can be found there.

Patch or snippet allowing to reproduce the problem:
```#include <wx/wx.h>
#include <wx/frame.h>
#include <wx/notebook.h>
#include <wx/collpane.h>

class MyApp : public wxApp
{
public:
    virtual bool OnInit() {
	wxFrame* frame = new wxFrame(nullptr, wxID_ANY, "Main");
	wxNotebook* notebook = new wxNotebook(frame, wxID_ANY);
	wxPanel* panel = new wxPanel(notebook, wxID_ANY);
	notebook->AddPage(panel, "Page", true);
	wxCollapsiblePane * pane = new wxCollapsiblePane(panel, wxID_ANY, "Collapsible pane");
	wxWindow *paneWindow = pane->GetPane();
    wxSizer *paneSizer = new wxBoxSizer(wxVERTICAL);
    paneSizer->Add(new wxStaticText(paneWindow, wxID_ANY, "Text inside pane"), 1, wxGROW|wxALL, 2);
    paneWindow->SetSizer(paneSizer);
    paneSizer->SetSizeHints(paneWindow);

	frame->Show(true);
	return true;
    }
};
wxIMPLEMENT_APP(MyApp);
```

To Reproduce:
Compare background color under the "Collapsible pane" text with the surrounding background color.

- wxWidgets version you use: current master (0c321cd039ce0e04afc3a63fc17f82339cc45bdc) from GitHub (just after 3.2.2.1 release)
- OS and its version: Windows 10

vadz:
This fixes #23274 for me, but I don't understand how could it work before as I don't see any changes that could have broken it. Maybe it never actually worked?

@TcT2k Please let me know if you see any reason not to do this (speaking about the first commit here, the second one is trivial), TIA!

### DIFF
--- a/include/wx/generic/collheaderctrl.h
+++ b/include/wx/generic/collheaderctrl.h
@@ -44,6 +44,8 @@ public:
     virtual bool IsCollapsed() const override
         { return m_collapsed; }
 
+    virtual bool HasTransparentBackground() override { return true; }
+
 protected:
 
     virtual wxSize DoGetBestClientSize() const override;
--- a/src/generic/collpaneg.cpp
+++ b/src/generic/collpaneg.cpp
@@ -78,8 +78,7 @@ bool wxGenericCollapsiblePane::Create(wxWindow *parent,
 
     // create children and lay them out using a wxBoxSizer
     // (so that we automatically get RTL features)
-    m_pButton = new wxCollapsibleHeaderCtrl(this, wxID_ANY, label, wxPoint(0, 0),
-                             wxDefaultSize);
+    m_pButton = new wxCollapsibleHeaderCtrl(this, wxID_ANY, label, wxPoint(0, 0));
 
     m_sz->Add(m_pButton, wxSizerFlags().Border(wxALL, GetBorder()));
 


TcT2k:
> This fixes #23274 for me, but I don't understand how could it work before as I don't see any changes that could have broken it. Maybe it never actually worked?
> 
I think that's probably true, because the most common use case for a collapsible pane would probably to expand a dialog where it wouldn't be on a notebook.

> @TcT2k Please let me know if you see any reason not to do this (speaking about the first commit here, the second one is trivial), TIA!

I have no objections.

PR #23289, opened 2023-02-22
Ignore extra version numbers in WX_CONFIG_CHECK

voyageur:
Hot-fix releases have an extra version number (like 3.2.2.1) which is reported in `wx-config --version` and messes WX_CONFIG_CHECK: the extra .1 is added to all components (major, minor, micro) and causes autoconf failures like:
test: 3.1: integer expression expected
(test on major version)

This ignores extra version numbers when defining major/minor/micro versions

### DIFF
--- a/wxwin.m4
+++ b/wxwin.m4
@@ -222,18 +222,18 @@ AC_DEFUN([WX_CONFIG_CHECK],
 
     WX_VERSION=`$WX_CONFIG_WITH_ARGS --version 2>/dev/null`
     wx_config_major_version=`echo $WX_VERSION | \
-           sed 's/\([[0-9]]*\).\([[0-9]]*\).\([[0-9]]*\)/\1/'`
+           sed 's/\([[0-9]]*\).\([[0-9]]*\).\([[0-9]]*\).*/\1/'`
     wx_config_minor_version=`echo $WX_VERSION | \
-           sed 's/\([[0-9]]*\).\([[0-9]]*\).\([[0-9]]*\)/\2/'`
+           sed 's/\([[0-9]]*\).\([[0-9]]*\).\([[0-9]]*\).*/\2/'`
     wx_config_micro_version=`echo $WX_VERSION | \
-           sed 's/\([[0-9]]*\).\([[0-9]]*\).\([[0-9]]*\)/\3/'`
+           sed 's/\([[0-9]]*\).\([[0-9]]*\).\([[0-9]]*\).*/\3/'`
 
     wx_requested_major_version=`echo $min_wx_version | \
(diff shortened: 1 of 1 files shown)


voyageur:
Spotted with filezilla 3.63 and wxWidgests 3.2.2.1 on Gentoo Linux: https://bugs.gentoo.org/895642

vadz:
Thanks for the PR but I don't quite understand the problem, as wx-config does _not_ return the fourth version component here:
```
$ git checkout v3.2.2.1
$ ./configure --without-subdirs >/dev/null
$ ./wx-config --version
3.2.2
```
and I don't understand how could it do it considering that this component is only part of `WX_SUBVERSION` which is not used in `wx-config` -- could you please explain how does this happen in Gentoo? Does it use any patches to wx-config by chance?

I think we could still apply the proposed patch as I don't think it can do any harm, but it seems a bit weird to do it when it looks like it shouldn't be needed, so I'd like to have more details please.

Thanks in advance!

voyageur:
Interesting, thanks! This does come from somewhere in Gentoo build:
```
# Git tag checkout
% ./wx-config --version
3.2.2
# Gentoo build in progress
/var/tmp/portage/x11-libs/wxGTK-3.2.2.1/work/wxWidgets-3.2.2.1/configure --prefix=/usr --build=x86_64-pc-linux-gnu --host=x86_64-pc-linux-gnu --mandir=/usr/share/man --infodir=/usr/share/info --datadir=/usr/share --sysconfdir=/etc --localstatedir=/var/lib --datarootdir=/usr/share --disable-dependency-tracking --docdir=/usr/share/doc/wxGTK-3.2.2.1 --htmldir=/usr/share/doc/wxGTK-3.2.2.1/html --with-zlib=sys --with-expat=sys --enable-compat30 --with-sdl --with-liblzma --enable-webrequest --disable-precomp-headers --libdir=${prefix}/lib64 --enable-graphics_ctx --with-gtkprint --enable-gui --with-gtk=3 --with-libpng=sys --with-libjpeg=sys --with-regex=sys --without-gnomevfs --enable-mediactrl --disable-webview --with-libnotify --with-opengl --with-libtiff=sys --disable-secretstore --enable-spellcheck --disable-tests --enable-wayland
[...]
% /var/tmp/portage/x11-libs/wxGTK-3.2.2.1/work/wxWidgets-3.2.2.1-abi_x86_64.amd64/wx-config --version
3.2.2.1
```

I did not see patches relating to versioning, but the ebuild itself may change, let me dig further (and report to wxWidgets downstream maintainers)

voyageur:
Thanks for comments here, Gentoo bug https://bugs.gentoo.org/895982 is now fixed and reports proper versions (tested with filezilla).
I do not think this PR itself brings a lot when wx-config reports expected versions so closing it

PR #23294, opened 2023-02-24
Improve "About" dialog appearance under MSW

vadz:
The current dialog looks rather ugly and it's difficult to recommend people to use it in their applications, so I've tried to improve it to be sufficiently presentable.

Please let me know if anybody has any strong objections to this, otherwise I'll merge this soon into master and I think we might also consider backporting this to 3.2.


vadz:
If there are no objections, I'll merge this one soon.

Issue #23250, opened 2023-02-10
OSX: Adding text control to the toolbar


oneeyeman1:
Hi,

Add following 2 lines to the toolbar sample:

```cpp
    // up for the current DPI, if necessary.
    toolBar->SetToolBitmapSize(FromDIP(sizeBitmap));
+    wxTextCtrl *test = new wxTextCtrl( toolBar, wxID_ANY, "" );
+   toolBar->AddControl( test );
    toolBar->AddTool(wxID_NEW, "New",
                     toolBarBitmaps[Tool_new], wxNullBitmap, wxITEM_DROPDOWN,
                     "New file", "This is help for new file tool");

```

Build and run the toolbar sample on all 3 platforms.

On both Windows and GTK control will have some reasonable sizes.

However, on OSX text control will be very small.

I use OSX 10.13.6 High Sierra and Xcode 9.4 with wxWidgets previous release. (Didn't tried the latest one yet).

![Screen Shot 2023-02-10 at 3 34 10 PM](https://user-images.githubusercontent.com/5044312/218204240-4199dd1f-27d9-4876-9cab-b51289773975.png)

oneeyeman1:
![table](https://user-images.githubusercontent.com/5044312/218204715-7a6d4b21-6eb4-4da5-bc55-d7715649fa7e.svg)

oneeyeman1:
![Screenshot onGTK](https://user-images.githubusercontent.com/5044312/218204809-67453ae1-394e-4a46-91ee-f261a9dfebb1.png)

vadz:
Probably can be worked around by giving some default size to the control but it would, of course, be better if a reasonable size was used by default.

P.S. I've edited the description to use proper markup for text. Could you please think to check the "Preview" button before submitting the next time?

oneeyeman1:
@vadz,
This is actually a regression.
I just built the sample on wx-3.1.3 and ran it on OSX 10.8.5.

The text control has reasonable size.

vadz:
It would be useful to bisect the history to find out when this got broken, if you can do it.

oneeyeman1:
@vadz ,
Doing bisection shows this commit:

```
commit dc2ff0d6b42b93dc84b886571bc6008393721f1e
Author: Stefan Csomor <csomor@advancedconcepts.ch>
Date:   Sat Apr 30 20:34:28 2022 +0200

    wxOSX: Rewiring GetBestSize for single line text control again
    
    see #22374

```

[...]

Issue #22374, opened 2022-04-28
Text controls are not properly laid out when the font is changed.


dconnet:
**Describe the bug**
Text controls are not properly sized (in layouts) when the font is changed.

**Expected vs observed behaviour**
The text control should be properly sized on layout. On Windows, this happens. On Mac is does not. The font is properly set, but the height of the control is not adjusted.

**Patch or snippet allowing to reproduce the problem**

Patch against the 'text' sample on the current trunk.
```
diff --git a/samples/text/text.cpp b/samples/text/text.cpp
index 4d40c8dcfb..227990b51b 100644
--- a/samples/text/text.cpp
+++ b/samples/text/text.cpp
@@ -1123,6 +1123,10 @@ MyPanel::MyPanel( wxFrame *frame, int x, int y, int w, int h )
     m_text->SetInsertionPoint(0);
     m_text->WriteText( "Prepended. " );

+    auto font = m_text->GetFont();
+    font.SetPointSize(font.GetPointSize() * 2);
+    m_text->SetFont(font);
+
 #if wxUSE_SPELLCHECK
     if ( m_text->EnableProofCheck(wxTextProofOptions::Default()) )
     {
```

**Platform and version information**
 - wxWidgets 3.1.6 (also happens on trunk)
 - wxWidgets port: wxOSX
 - OS and its version: OSX 12.2 and 10.15

vadz:
@csomor I could be missing something, but I just don't see any `GetBestSize()` implementation for the single-line (`NSField`-based) `wxTextCtrl`. Do we need to implement it on our own or is there some way to get the size taking the font into account from Cocoa?

csomor:
@vadz I assume then that the sizeToFit which should get called from window.mm either is not implemented, or does not take things like font size into account. I'll look what I can find.

csomor:
Ok, I see more clearly now, it's funny, somewhere in the `wxTextWidgetImpl` extraction process, it lost connection to the `wxWindowMac::DoGetBestSize` implementation, I'll add the few lines now to the sublass responsible for the native `NSTextField`, `wxNSTextFieldControl`

csomor:
@dconnet Dave, could you please retest ? Thanks

dconnet:
Confirmed. Also confirmed when applying this patch to 3.1.6.
thx!

whalley:
[...]

PR #23295, opened 2023-02-25
wxOSX regression: default-widths for empty textfields

csomor:
in the process of fixing #22374 by using native best sizes, empty fields were getting a small, but positive best width, this broke the trigger for using default widths for empty fields, see #23250

### DIFF
--- a/src/osx/textctrl_osx.cpp
+++ b/src/osx/textctrl_osx.cpp
@@ -200,6 +200,7 @@ wxSize wxTextCtrl::DoGetBestSize() const
 wxSize wxTextCtrl::DoGetSizeFromTextSize(int xlen, int ylen) const
 {
     static const int TEXTCTRL_BORDER_SIZE = 5;
+    static const int TEXTCTRL_MAX_EMPTY_WIDTH = 5;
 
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for fixing this! It feels a bit bad to hardcode the size like this, but if you think it's fine, I can certainly live with it... As an alternative, perhaps we could compare it with something like `GetCharWidth()/2` (or would this be too small?)?

csomor:
> Thanks for fixing this! It feels a bit bad to hardcode the size like this, but if you think it's fine, I can certainly live with it... As an alternative, perhaps we could compare it with something like `GetCharWidth()/2` (or would this be too small?)?

the thing is that the constant refers to the pixel insets Apple applies to a text control, I don't think that these would scale down if the font is smaller, or increase if it's bigger, perhaps it may be related to the 'normal' 'smaller' 'larger' style of a control. But this would mean testing and making the code more complex, so at the moment I'd like to keep it hardcoded.

vadz:
Thanks for the explanation! Applied as is to master now and will backport to 3.2 too soon.

Issue #23286, opened 2023-02-21
Shaped windows don't appear correctly in dark mode under Mac


vadz:
While debugging another problem in the dialogs sample, I've noticed that the shaped rich tooltip window doesn't look right in dark mode (it's fine, i.e. really shaped, in light one), as it has grey background in the "transparent" parts, see this screenshot:

![mac_dark_shaped](https://user-images.githubusercontent.com/146917/220440454-a41107db-b710-4993-b8b3-e959fa03b60c.png)

I have no idea why using `[NSColor clearColor]` as background colour in `wxNonOwnedWindowCocoaImpl::SetShap()` doesn't work in dark mode, unfortunately, so I don't know how to fix this.

csomor:
ok, we are overriding the background color of the toplevel window with a solid color, that's why our previous clearColor gets changed ... I'll try to find a way to change the colour on the background view but not on the NSWindow ...

PR #23296, opened 2023-02-25
macOS fix background color on shaped frame (eg rich tooltip)

csomor:
see #23286


vadz:
Thanks for fixing this one too!

I've applied it to master and will cherry pick it into 3.2 too if nobody sees any problems with it.

PR #23297, opened 2023-02-25
Removal of compiler warning because of incompatible integer assignments

hwiesmann:
Pointer arithmetic uses std::ptrdiff_t as the result type but LastUsedCacheElement's type is unsigned int that is unlikely being equivalent to std::ptrdiff_t. Therefore, the pointer arithmetic has to be casted (or compiler warnings may be the result). Casting is safe as the cache size is rather small.

### DIFF
--- a/src/osx/cocoa/window.mm
+++ b/src/osx/cocoa/window.mm
@@ -3311,7 +3311,7 @@ void wxWidgetCocoaImpl::SetBackgroundColour( const wxColour &col )
         wxWindow* peer = GetWXPeer();
         if ( peer->GetBackgroundStyle() != wxBG_STYLE_TRANSPARENT )
         {
-            wxTopLevelWindow* toplevel = wxDynamicCast(peer,wxTopLevelWindow);
+            wxNonOwnedWindow* toplevel = dynamic_cast<wxNonOwnedWindow*>(peer);
 
             if ( toplevel == nullptr || toplevel->GetShape().IsEmpty() )
                 [targetView setBackgroundColor:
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, it looks like this only happens in UTF-8 builds, right? Somehow I don't see the warning, could you please show the warning and/or give the compiler and compiler options you use when you get it?

In particular, I'd like to check this because it looks like this should be applied to 3.2 too, or does this warning not happen there for some reason?

hwiesmann:
Hi,

I am using Apple Clang with Implicit integer to pointer conversions switched on. I believe that it is this warning switch as all the others that are related to integer conversions are switched off (because there are a lot of other warnings otherwise).

BTW: I am not using the standard make or configuration files.

hwiesmann:
Hi,

sorry, my information before was wrong. The warning came from another part of the project (as the source code is located in a header file) where the warning for implicit conversion from 64 bit to 32 bit was switched on.

BTW: If you have switched on this warning while compiling wxWidgets you get countless number of warnings. Mostly the reason is that longs are assigned to ints.
PS: There should be a decision for wxWidgets to use either ints or longs. Using both mostly does not make sense. Even worse: it is often also wrong when comparing numbers (e.g. long (-1) != std::numeric_limits<unsigned int>::max())

vadz:
What exactly is this warning (you should see its name in the full error message)? I'd like to enable it in `allheaders` test.

As for mixing up ints and longs, this is unfortunately done in so many places in the existing code that fixing it is rather hopeless and not turning this warning on for wx itself is much simpler. We do try to make the wx _headers_ compile without any warnings at all, so that the projects using wx could enable all the warnings they want.

hwiesmann:
The warning itself is: -Wshorten-64-to-32. Leading to: "warning: implicit conversion loses integer precision: long to unsigned it."

vadz:
This should be fixed by #23301, please let me know if you still see any problems with it.

PR #23299, opened 2023-02-26
Explain behavior of calling SetSashPosition() with 0

Blake-Madden:
I thought that calling `wxSplitterWindow::SetSashPosition()` with zero would hide the first subwindow, but instead it (by design) splits the subwindows evenly. This PR explains this behavior.


vadz:
Thanks! Will merge into master soon and 3.2 a bit later.

PR #23305, opened 2023-03-01
Fix utf8 iterator going past end of string in PosLenToImpl

Zekkers:
Fix iterator going past end of string in PosLenToImpl which causes std::string exception dereferencing end()

### DIFF
--- a/src/common/string.cpp
+++ b/src/common/string.cpp
@@ -229,7 +229,7 @@ void wxString::PosLenToImpl(size_t pos, size_t len,
             // going beyond the end of the string, just as std::string does
             const const_iterator e(end());
             const_iterator i(b);
-            while ( len && i <= e )
+            while ( len && i < e )
             {
                 ++i;
                 --len;


vadz:
Thanks, this looks correct, but would you have a test case allowing to reproduce the problem, by chance? I'd like to do it to confirm that it's really fixed before merging (of course, if you can add a unit test for this to `tests/strings`, this would be even better).

TIA!

Zekkers:
Yeah, I'll come up with something for tests. Essentially, it was crashing during stack trace collection after building for UTF8 and alongside wxLua.

Zekkers:
Actually, there's already a test that fails with this issue. Seems you're not running these tests against UTF8?

```
DateTimeTestCase
  TestTimeParse
C:\git\cbrx\cambrionix-tool\external\wxWidgets\wxWidgets\tests\datetime\datetimetest.cpp(222)

C:\git\cbrx\cambrionix-tool\external\wxWidgets\wxWidgets\tests\datetime\datetimetest.cpp(222): FAILED:
  CRT assert failure: C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\
  Tools\MSVC\14.36.32323\include\xstring(1906) : Assertion failed: cannot
  dereference string iterator because it is out of range (e.g. an end iterator)

```

There's a few other failed tests running under UTF8 btw:
[wx-test-fails.txt](https://github.com/wxWidgets/wxWidgets/files/10882858/wx-test-fails.txt)

vadz:
We do run tests in the CI build using UTF-8 but only under Unix and we don't catch this problem without ASAN/valgrind there, unfortunately.

So it looks like we need to

1. Add a CI build using UTF-8 under MSW/with MSVS.
1. Enable ASAN for UTF-8 Unix build.

I'll try to do this soon, thanks.

vadz:
Hmm, it seems like there is another problem in `wxString` code, see [this build log](https://github.com/wxWidgets/wxWidgets/actions/runs/4337288559/jobs/7573181759). I'll need to debug this one locally...

vadz:
And there also build error when using MSVS, e.g.

```
..\..\src\msw\registry.cpp(1594): error C2446: ':': no conversion from 'const wxScopedWCharBuffer' to 'nullptr'
```

I guess you must have some custom changes to wx fixing this, can you please make them available so that I don't spend time on re-fixing them myself?

Zekkers:
Hmmm. I've made no changes to wxWidgets source at all. I'm including it in a cmake project via add_subdirectory with the following options:

```
set(wxUSE_PROPGRID OFF)
set(wxUSE_UNICODE_UTF8 ON)
set(wxUSE_UTF8_LOCALE_ONLY ON)
set(wxUSE_STL ON)
set(wxUSE_STD_CONTAINERS ON)
set(wxUSE_LIBTIFF OFF)
set(wxBUILD_MONOLITHIC ON)
set(wxBUILD_SHARED OFF)
```

vadz:
Weird, I don't understand how can the code in `src/msw/registry.cpp` compile. I'll fix it, but not today.

PBfordev:
I am also surprised that UTF-8 build succeeds on MSW, back when I tried, it did not work neither with both with MSVS nor GCC (wxString TLS with MSVS and registry with GCC) : https://github.com/wxWidgets/wxWidgets/pull/23016#issuecomment-1338110136

vadz:
This is really weird because there have been several issues concerning UTF-8 build in just a few days (see #23297 and #23328), so people definitely are using it. I'm ill right now but I'll try to return to #23313 and fix it under MSW if it's so popular...

Issue #23036, opened 2022-12-14
Generalize the use of `wxArrayStringsAdapter`


vadz:
It would be convenient to be able to just pass the list of strings to various functions currently taking either a C array or `wxArrayString`, e.g. `wxListBox` and other such classes ctors. In fact, it should be safe to just replace `wxArrayString` with `wxArrayStringsAdapter` in them, which would allow passing `std::vector<wxString>` directly, without breaking any existing code as there are no implicit ctors in `wxArrayString`, so nothing should rely on them happening.

For even more convenience we could add a converting ctor from an initializer list of strings, although we'd need to be careful not to introduce ambiguities if we do this.

vadz:
After #23309, adding a converting ctor from `std::vector<T>` to `wxArrayString` looks like a better idea, so I've done this now in #23434.

PR #23309, opened 2023-03-02
Add support for initializer_list on arrays

Lotendan:
Hello,

Class `wxArrayString` is used all throughout the project in many different components. This pull-requests intends to improve the developer experience by adding support for `std::initializer_list` and syntactic sugar to simplify the initialization of `wxArrayString` and other arrays.

Exemple:
```
wxArrayString choices;
choices.Add( _T( "Choice 1" ) );
choices.Add( _T( "Choice 2" ) );
wxChoice* choiceControl = new wxChoice( this, wxID_ANY, wxDefaultPosition, wxDefaultSize, choices );

// OR

wxChoice* choiceControl = new wxChoice( this, wxID_ANY );
choiceControl->Append( { _T( "Choice 1" ), _T( "Choice 2" ) } );

// Replaced by:
wxChoice* choiceControl = new wxChoice( this, wxID_ANY, wxDefaultPosition, wxDefaultSize, { _T( "Choice 1" ), _T( "Choice 2" ) } );
```

I wasn't sure about adding a constructor based on `std::vector<std::string>` or more generally `std::vector<T>` too. Any suggestions about this are welcome.

### DIFF
--- a/include/wx/arrstr.h
+++ b/include/wx/arrstr.h
@@ -16,9 +16,7 @@
 #include "wx/dynarray.h"
 
 #include <vector>
-#include "wx/beforestd.h"
 #include <initializer_list>
-#include "wx/afterstd.h"
 
 // these functions are only used in STL build now but we define them in any
 // case for compatibility with the existing code outside of the library which
@@ -84,8 +82,8 @@ public:
     wxArrayString(size_t sz, const char** a);
     wxArrayString(size_t sz, const wchar_t** a);
     wxArrayString(size_t sz, const wxString* a);
-    template<typename _IList>
-    wxArrayString(std::initializer_list<_IList> list) : wxArrayStringBase(list) { }
+    template<typename U>
+    wxArrayString(std::initializer_list<U> list) : wxArrayStringBase(list) { }
 
     int Index(const wxString& str, bool bCase = true, bool bFromEnd = false) const;
 
@@ -188,8 +186,8 @@ public:
     // copy ctor
   wxArrayString(const wxArrayString& array);
     // list constructor
-  template<typename _IList>
-  wxArrayString(std::initializer_list<_IList> list) { Init(false); assign(list.begin(), list.end()); }
+  template<typename U>
+  wxArrayString(std::initializer_list<U> list) { Init(false); assign(list.begin(), list.end()); }
     // assignment operator
   wxArrayString& operator=(const wxArrayString& src);
     // not virtual, this class should not be derived from
--- a/include/wx/arrstr.h
+++ b/include/wx/arrstr.h
@@ -16,6 +16,9 @@
 #include "wx/dynarray.h"
 
 #include <vector>
+#include "wx/beforestd.h"
+#include <initializer_list>
+#include "wx/afterstd.h"
 
 // these functions are only used in STL build now but we define them in any
 // case for compatibility with the existing code outside of the library which
@@ -81,6 +84,8 @@ public:
     wxArrayString(size_t sz, const char** a);
     wxArrayString(size_t sz, const wchar_t** a);
     wxArrayString(size_t sz, const wxString* a);
+    template<typename _IList>
+    wxArrayString(std::initializer_list<_IList> list) : wxArrayStringBase(list) { }
 
     int Index(const wxString& str, bool bCase = true, bool bFromEnd = false) const;
 
@@ -182,6 +187,9 @@ public:
   wxArrayString(size_t sz, const wxString* a);
     // copy ctor
   wxArrayString(const wxArrayString& array);
+    // list constructor
+  template<typename _IList>
+  wxArrayString(std::initializer_list<_IList> list) { Init(false); assign(list.begin(), list.end()); }
     // assignment operator
   wxArrayString& operator=(const wxArrayString& src);
     // not virtual, this class should not be derived from
--- a/interface/wx/arrstr.h
+++ b/interface/wx/arrstr.h
@@ -90,9 +90,9 @@ public:
     wxArrayString(size_t sz, const wxString* arr);
 
     /**
-        Constructs the container with the contents of the initializer_list @list.
+        Constructs the container with the contents of the initializer_list @a list.
 
-        @since 3.2.2
+        @since 3.2.3
     */
     template<typename T>
     wxArrayString(std::initializer_list<T> list);
--- a/interface/wx/arrstr.h
+++ b/interface/wx/arrstr.h
@@ -90,6 +90,14 @@ public:
     wxArrayString(size_t sz, const wxString* arr);
 
     /**
+        Constructs the container with the contents of the initializer_list @list.
+
+        @since 3.2.2
+    */
+    template<typename T>
+    wxArrayString(std::initializer_list<T> list);
+
+    /**
         Destructor frees memory occupied by the array strings. For performance
         reasons it is not virtual, so this class should not be derived from.
     */
--- a/interface/wx/dynarray.h
+++ b/interface/wx/dynarray.h
@@ -272,6 +272,14 @@ public:
     wxObjArray(const wxObjArray& array);
 
     /**
+        Constructs the container with the contents of the initializer_list @list.
+
+        @since 3.2.2
+    */
+    template<typename T>
+    wxArray(std::initializer_list<T> list);
+
+    /**
         Performs a shallow array copy (i.e.\ doesn't copy the objects pointed to
         even if the source array contains the items of pointer type).
     */
--- a/interface/wx/dynarray.h
+++ b/interface/wx/dynarray.h
@@ -272,9 +272,9 @@ public:
     wxObjArray(const wxObjArray& array);
 
     /**
-        Constructs the container with the contents of the initializer_list @list.
+        Constructs the container with the contents of the initializer_list @a list.
 
-        @since 3.2.2
+        @since 3.2.3
     */
     template<typename T>
     wxArray(std::initializer_list<T> list);
--- a/include/wx/dynarray.h
+++ b/include/wx/dynarray.h
@@ -15,9 +15,7 @@
 
 #include "wx/vector.h"
 
-#include "wx/beforestd.h"
 #include <initializer_list>
-#include "wx/afterstd.h"
 
 /*
   This header defines legacy dynamic arrays and object arrays (i.e. arrays
@@ -109,8 +107,8 @@ public:
         : base_vec(first, last)
     { }
 
-    template<typename _IList>
-    wxBaseArray(std::initializer_list<_IList> list) : base_vec(list) {}
+    template<typename U>
+    wxBaseArray(std::initializer_list<U> list) : base_vec(list.begin(), list.end()) {}
 
     void Empty() { this->clear(); }
     void Clear() { this->clear(); }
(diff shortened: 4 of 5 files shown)


include/wx/arrstr.h
```diff
@@ -16,6 +16,9 @@
 #include "wx/dynarray.h"
 
 #include <vector>
```
Lotendan, line 18: Small inconsistency here: I didn't want to include `<vector>` between the `beforestd` and `afterstd` guards. Is that expected?

interface/wx/dynarray.h
```diff
@@ -271,6 +271,14 @@ class wxArray<T>
     */
     wxObjArray(const wxObjArray& array);
 
+    /**
+        Constructs the container with the contents of the initializer_list @list.
+
+        @since 3.2.2
+    */
+    template<typename T>
+    wxArray(std::initializer_list<T> list);
```
Lotendan, line 280: Do we need to add documentation for the other constructors as well? `wxObjArray` and `wxSortedArray`?

include/wx/arrstr.h
vadz: It doesn't really matter in practice because there are no known problems with including `<vector>` outside of them (and very few problems in standard headers remain at all, so we probably could include `<initializer_list>` outside of them too if we really wanted).

include/wx/arrstr.h
```diff
@@ -81,6 +84,8 @@ class WXDLLIMPEXP_BASE wxArrayString : public wxArrayStringBase
     wxArrayString(size_t sz, const char** a);
     wxArrayString(size_t sz, const wchar_t** a);
     wxArrayString(size_t sz, const wxString* a);
+    template<typename _IList>
```
vadz, line 87: We probably should stick to using `T` for the generic template parameters, especially because `_IList` is a bit confusing (it's the type of the initializer list elements and not of the list itself) and is also a reserved identifier.

interface/wx/arrstr.h
```diff
@@ -89,6 +89,14 @@ class wxArrayString : public wxArray
     */
     wxArrayString(size_t sz, const wxString* arr);
 
+    /**
+        Constructs the container with the contents of the initializer_list @list.
+
+        @since 3.2.2
```
vadz, line 95:
```suggestion
        @since 3.2.3
```

interface/wx/dynarray.h
vadz: To be honest, I have no idea what's going on with the documentation of this class as it seems to document several different classes inside a single class declaration... But, to answer the question, I don't think we need to continue doing it like this, just this should be enough, thanks.

interface/wx/dynarray.h
```diff
@@ -271,6 +271,14 @@ class wxArray<T>
     */
     wxObjArray(const wxObjArray& array);
 
+    /**
+        Constructs the container with the contents of the initializer_list @list.
+
+        @since 3.2.2
```
vadz, line 277:
```suggestion
        @since 3.2.3
```

vadz (commented):
Thanks!

I definitely agree that we need to do something here, but I was thinking of doing this "something" at `wxArrayStringsAdapter` level, see #23036, and I'm not sure which approach is better, but they're probably mutually exclusive as we'd have ambiguous overloads if we implemented both proposals.

One advantage of this one is that it can indeed be done even in 3.2 as it's backwards compatible. And, to be honest, I don't see any problems with it.

Lotendan:
Thank you for the first level of review, that’s really appreciated.
The issue #23036 and `wxArrayStringsAdapter` look interesting too, but go much further than I initially planned, especially as this is my first contribution in the repository.
Do you suggest that we still go ahead with this? At least in 3.2? Or take the time to think about it further as is pointed out in #23036? Maybe the overloads it introduces are not that much incompatible/exclusive and I see this more as a first step towards the StringsAdapter. At least the `{ … }` syntax would remain source compatible?

include/wx/arrstr.h
Lotendan: OK, since `T` is the name of the template parameter in some other definitions I opted for a more generic name while being consistent all over the place. I can change that to `U` I guess.
Do you have any documentation about `_IList` being a reserved identifier?

vadz: From https://en.cppreference.com/w/cpp/language/identifiers (which is authoritative enough, but the Standard has the same wording):

Identifiers that appear as appearing as a token or preprocessing token of one of the following forms are reserved:

- identifiers with a double underscore anywhere;
- identifiers that begin with an underscore followed by an uppercase letter;
- in the global namespace, identifiers that begin with an underscore. 

And `_IList` falls in the second category, of course.

vadz:
I think the main advantage of `wxArrayStringsAdapter` is that it could, in theory, avoid copying the strings. However it's probably not really worth it for the amounts of strings that we operate with here -- there are probably never going to be more than, what, a couple of hundreds, maybe, of them? In practice I think the worst I've ever seen was the list of US states, i.e. ~50 (without going into the details of the US federal system).

So, even after thinking about it, I don't really see any problems with doing it like this and I think we should go ahead and do it in master for now and then, if we don't find any problems in practice neither, backport it to 3.2.

Of course, the CI builds need to be fixed first -- could you please look at the errors in them? TIA!

include/wx/arrstr.h
Lotendan: Removed the std guards

include/wx/arrstr.h
Lotendan: Thanks a lot for the details, I was not aware. Very useful to know.
This is fixed and I have replaced `_IList` with `Ty`

interface/wx/arrstr.h
Lotendan: Fixed

interface/wx/dynarray.h
Lotendan: Fixed

vadz:
Annoyingly, it looks like this doesn't compile in STL build (see the failing CI builds). I didn't yet have time to look at this, do you already see a way to fix it there?

Lotendan:
In some cases the compiler fails to infer the template parameters in the `initializer_list` if `T` and `U` are not the same types (and even though `T` is constructible from `U`. I am trying another alternative to make the type inference more explicit to the compiler. With this the tests pass in local.

Lotendan:
It worked :) The trick was to call the base constructor with iterators.
All green now! Please have one last look, and then we are ready to merge.

vadz:
Looks good, thanks!

I'll merge this in master soon. For 3.2 we'll need to add `#if wxCHECK_CXX_STD(201103L)` checks around the new functions as we still support using C++98 there, but I can do it.

Lotendan:
Sure, let me know if there's anything more I can do to help on this, or if you need me to do the PR on 3.2.

vadz:
Sorry for completely forgetting to reply to this but a PR for 3.2 would indeed be nice.

There we still need to test whether we're using C++11 at all but, AFAICS, we don't need `wxABI_VERSION` checks as the functions are inline anyhow -- but it would be nice to check this, e.g. building against the latest 3.2 branch with this PR applied and verifying that the binary still runs when using distribution provided binaries for an earlier 3.2 version.

TIA!

IreneKnapp:
Hi! I'm the nixpkgs maintainer for Tenacity, which uses wxGTK. This initializer_list work came to my attention via https://github.com/NixOS/nixpkgs/pull/266945#issuecomment-1807021222

From preliminary investigation, it appears that this change is responsible for breaking at least Tenacity and Audacity.

I think it's a good change and should be kept. I've already prepared a fix on the Tenacity side of things and will be driving the process to get it upstreamed. Therefore, I'm not making any sort of request from you, I'm just giving you a heads up because I thought you'd like to know, in case it bears on any other planned changes or anything.

vadz:
@IreneKnapp Thanks for reporting this and sorry for breaking it. I am not sure what exactly is the problem in {Tena,Auda}city is, but I guess there is some new overload ambiguity?

It looks like we really can't do any changes in 3.2 headers without breaking something :-( Not sure what to do about it, ideal would be to at least test the changes _before_ the release, but we've never managed to do this. Again, I'd invite all maintainers of the packages using wx to monitor the project communications (we could create a special ultra low traffic mailing list for this or maybe post pre-release announcements to the existing wx-announce too) and help us with discovering such problems before the release. TIA!

IreneKnapp:
yeah there was a place where it uses list initialization and it used to fall back to a copy constructor because there was no initializer list support, but now it tries to use list initialization and gets an error because the parameter it's passing is a const reference and it used to get implicitly copied but now it doesn't, so there's no appropriate method to handle it

and there was another place where it's using an overloaded assignment operator and now there's more options for what it could be so it's ambiguous, as you suggested.

I'll add a link to the Tenacity PR sometime today so you can see the details.

if you start a list like that I'll join it, I can't promise to read everything. please do let me know if us nix people can help in some way, nix is really good at rebuilding downstream stuff after changes, that has to be useful for testing somehow...

IreneKnapp:
I'm told upstream has fixed this; anyway here's my changes that show where the problems were - https://github.com/NixOS/nixpkgs/pull/266945#issuecomment-1807292941

vadz:
Thanks for the explanations! I should have thought of this and kept the new ctor(s) inside some `#if wxUSE_INIT_LIST`... But it's too late for this now, so I'll just try to document it.

As for the list, I've [asked](https://groups.google.com/g/wx-dev/c/-pSZ74GEafo) whether a new one or posting to the existing [wx-announce one](https://groups.google.com/g/wx-announce) would be preferable. If you have any preferences here, please let me know (here or by replying on wx-dev). For now I'm leaning towards just reusing the existing list to avoid creating too many of them.

IreneKnapp:
no preference here, the announce list does make sense to me.

evils:
note that it'd be nice if i could subscribe to the announcements without having to involve a google account
(this seems to work for the [KiCad dev mailing list / google group](https://groups.google.com/a/kicad.org/g/devlist), though maybe that only works when it's in read+write)

vadz:
@evils I don't think you need a Google account to subscribe to the mailing list, at least not if you do it via email, i.e. send a request to wx-announce+subscribe@googlegroups.com. If this doesn't work for you, please contact me and we'll try to figure this out (although dealing with Google groups is difficult...).

evils:
@vadz
i did send a mail there before sending my message here
i just confirmed again, it doesn't work
i do see a "contact owners and managers" button, which pops up a message authoring window, but its content fails to load
so either something is blocking its content, or it also requires a google account

reviewing the KiCad devlist subscription email, maybe i did link it to my google account...
though neither of my google accounts show it is "my groups"

joining does have an option to uncheck "link to my google profile"
but that does add the group to my google groups "my groups" list (and seems to just reuse my avatar image)
and didn't send a join confirmation to the email address i attempted to subscribe from
(my main issue there is which inbox the mails go to, aside from requiring more than an email address to subscribe...)

vadz:
@evils Unfortunately I can confirm that it doesn't work, I've tested it by posting an email to subscribe from a unique address and got a reply directing me to https://groups.google.com/group/wx-announce/subscribe where I can't do anything. I don't see anything I can do about this in the group options neither, unfortunately.

I can only suggest that you contact me by email and I try adding your email to the list directly, I think this might work but I'm not even sure about any more. Sorry.

swt2c:
> I can only suggest that you contact me by email and I try adding your email to the list directly, I think this might work but I'm not even sure about any more. Sorry.

That definitely does work.  I maintain a Google Group for something unrelated and the administrative interface does allow you to add non-Google users manually.  They users can't access the web interface (eg, archives) though.

PR #23311, opened 2023-03-04
AUI Toolbar Changes

imciner2:
This PR makes two sets of changes to the AUI toolbar, which nominally breaks a bit of the API in some places (but a lot of effort was made to preserve API for most of the changes.

The first change is to change the container type used by wxAuiToolBar to be a list of pointers instead of an array of items. This allows wxAuiToolBarItem to be subclassed, and then for those subclasses to be used in the toolbar. This is similar to how a wxMenu operates (and since wxMenu was built that way, I initially thought the toolbar would work that way also). I actually do have a usecase for subclassing a tool, specifically to add multiple kinds of additional data/metadata to a tool (so it is better to subclass it than use a struct in the client data).

This change to a list from an array also allows for quite a bit of modernization of the loops in the AUI toolbar, replacing the older ones using indices with range-based for loops over the items. (the previous array implementation didn't have the proper iterators for range-based for loops for this code).

The second set of changes adds AUI-specific item IDs and actually exports them. The AUI toolbar uses 7 different values for kind, but only 4 were ever exported (since they were also in `wxItemKind`), so the other three were internal-only. This change exports them all in the user-facing API, while giving them the prefix `wxAUI_TB_ITEM_*`.

Overall, I have tried to preserve as much API as I can, but did decide to break the API in the toolbar art class in a non-backwards compatible way (but the fix is simple). This means that I did leave the original overflow item set method as a deprecated method (and marked as such). In order to preserve the API for the AUI-specific item ID change, I modified the toolbar `AddTool` methods to use `ints` instead of `wxitemKind`, allowing either the existing `wxitemKind` values or the new `wxAuiToolBarItemKind` values to be used (and I have kept the numerical values for those two types linked as well so that existing code still works with no changes).

### DIFF
--- a/include/wx/aui/auibar.h
+++ b/include/wx/aui/auibar.h
@@ -255,7 +255,9 @@ private:
 };
 
 #ifndef SWIG
+// wxAuiToolBarItemArray is deprecated but kept for backwards compatibility
 WX_DECLARE_USER_EXPORTED_OBJARRAY(wxAuiToolBarItem, wxAuiToolBarItemArray, WXDLLIMPEXP_AUI);
+WX_DECLARE_EXPORTED_LIST(wxAuiToolBarItem, wxAuiToolBarItemList);
 #endif
 
 
@@ -345,7 +347,7 @@ public:
 
     virtual int ShowDropDown(
                          wxWindow* wnd,
-                         const wxAuiToolBarItemArray& items) = 0;
+                         const wxAuiToolBarItemList& items) = 0;
 
     // Provide opportunity for subclasses to recalculate colours
     virtual void UpdateColoursFromSystem() {}
@@ -433,7 +435,7 @@ public:
     virtual void SetElementSize(int elementId, int size) override;
 
     virtual int ShowDropDown(wxWindow* wnd,
-                             const wxAuiToolBarItemArray& items) override;
+                             const wxAuiToolBarItemList& items) override;
 
     virtual void UpdateColoursFromSystem() override;
 
@@ -494,11 +496,23 @@ public:
     bool SetFont(const wxFont& font) override;
 
 
+    wxAuiToolBarItem* AddTool(wxAuiToolBarItem* item);
+
     wxAuiToolBarItem* AddTool(int toolId,
                  const wxString& label,
                  const wxBitmapBundle& bitmap,
                  const wxString& shortHelpString = wxEmptyString,
-                 wxItemKind kind = wxITEM_NORMAL);
+                 wxItemKind kind = wxITEM_NORMAL)
+    {
+        return AddTool(toolId,
+                label,
+                bitmap,
+                wxBitmapBundle(),
+                kind,
+                shortHelpString,
+                wxEmptyString,
+                nullptr);
+    }
 
     wxAuiToolBarItem* AddTool(int toolId,
                  const wxString& label,
@@ -560,6 +574,7 @@ public:
     bool GetToolFits(int toolId) const;
     wxRect GetToolRect(int toolId) const;
     bool GetToolFitsByIndex(int toolId) const;
+    bool GetToolBarItemFits(wxAuiToolBarItem* item) const;
     bool GetToolBarFits() const;
 
     void SetMargins(const wxSize& size) { SetMargins(size.x, size.x, size.y, size.y); }
@@ -617,8 +632,12 @@ public:
     wxString GetToolLongHelp(int toolId) const;
     void SetToolLongHelp(int toolId, const wxString& helpString);
 
+    // Preserved for API compatibility, shouldn't be used
     void SetCustomOverflowItems(const wxAuiToolBarItemArray& prepend,
                                 const wxAuiToolBarItemArray& append);
+
+    void SetCustomOverflowItems(const wxAuiToolBarItemList& prepend,
+                                const wxAuiToolBarItemList& append);
 
     // get size of hint rectangle for a particular dock location
     wxSize GetHintSize(int dockDirection) const;
@@ -668,7 +687,7 @@ protected: // handlers
 
 protected:
 
-    wxAuiToolBarItemArray m_items;      // array of toolbar items
+    wxAuiToolBarItemList m_items;       // Toolbar items
     wxAuiToolBarArt* m_art;             // art provider
     wxBoxSizer* m_sizer;                // main sizer for toolbar
     wxAuiToolBarItem* m_actionItem;    // item that's being acted upon (pressed)
@@ -678,8 +697,8 @@ protected:
     wxSizerItem* m_overflowSizerItem;
     wxSize m_absoluteMinSize;
     wxPoint m_actionPos;               // position of left-mouse down
-    wxAuiToolBarItemArray m_customOverflowPrepend;
-    wxAuiToolBarItemArray m_customOverflowAppend;
+    wxAuiToolBarItemList m_customOverflowPrepend;
+    wxAuiToolBarItemList m_customOverflowAppend;
 
     int m_buttonWidth;
     int m_buttonHeight;
--- a/include/wx/aui/auibar.h
+++ b/include/wx/aui/auibar.h
@@ -141,7 +141,9 @@ public:
         m_alignment = wxALIGN_CENTER;
     }
 
-    void Assign(const wxAuiToolBarItem& c)
+    virtual ~wxAuiToolBarItem() {}
+
+    virtual void Assign(const wxAuiToolBarItem& c)
     {
         m_window = c.m_window;
         m_label = c.m_label;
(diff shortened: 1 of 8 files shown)


imciner2:
I also realized that `wxAuiToolBar` was still not virtual, so I have modified it to support inheritance by adding a virtual destructor and virtual methods. I have also made the fields protected so that they can be accessed by subclasses.

One thing I am tempted to do is remove the need to make the AUI toolbar a friend of the item and just use the accessor methods instead, allowing subclasses to then modify those as needed. I will note that my want for subclassing comes more from wanting to add more fields to the item, not modify the fields already there.

include/wx/aui/auibar.h
```diff
@@ -125,7 +141,9 @@ class WXDLLIMPEXP_AUI wxAuiToolBarItem
         m_alignment = wxALIGN_CENTER;
     }
 
-    void Assign(const wxAuiToolBarItem& c)
+    virtual ~wxAuiToolBarItem() {}
```
vadz, line 144: A nitpick but this should be
```suggestion
    virtual ~wxAuiToolBarItem() = default;
```

include/wx/aui/auibar.h
```diff
@@ -150,87 +168,87 @@ class WXDLLIMPEXP_AUI wxAuiToolBarItem
...
 
-    bool CanBeToggled() const
+    virtual bool CanBeToggled() const
     {
-        return m_kind == wxITEM_CHECK || m_kind == wxITEM_RADIO;
+        return m_kind == wxAUI_TB_ITEM_CHECK || m_kind == wxAUI_TB_ITEM_RADIO;
     }
 
-private:
+protected:
```
vadz, line 251: Do we really need to do this? We already have accessors to all these fields AFAICS, so this doesn't seem necessary.

include/wx/aui/auibar.h
```diff
@@ -255,7 +273,9 @@ class WXDLLIMPEXP_AUI wxAuiToolBarItem
 };
 
 #ifndef SWIG
+// wxAuiToolBarItemArray is deprecated but kept for backwards compatibility
 WX_DECLARE_USER_EXPORTED_OBJARRAY(wxAuiToolBarItem, wxAuiToolBarItemArray, WXDLLIMPEXP_AUI);
+WX_DECLARE_EXPORTED_LIST(wxAuiToolBarItem, wxAuiToolBarItemList);
```
vadz, line 278: Let's avoid using the macro-based containers for the new code, i.e. please just use `std::vector<wxAuiToolBarItem*>` or maybe `std::vector<std::unique_ptr<wxAuiToolBarItem>>` if it should own the items.

include/wx/aui/auibar.h
```diff
@@ -494,17 +514,29 @@ class WXDLLIMPEXP_AUI wxAuiToolBar : public wxControl
...
 
 
+    wxAuiToolBarItem* AddTool(wxAuiToolBarItem* item);
+
     wxAuiToolBarItem* AddTool(int toolId,
                  const wxString& label,
                  const wxBitmapBundle& bitmap,
                  const wxString& shortHelpString = wxEmptyString,
-                 wxItemKind kind = wxITEM_NORMAL);
+                 int kind = wxAUI_TB_ITEM_NORMAL)
```
vadz, line 523: Why `int` and not `wxAuiToolBarItemKind`?

include/wx/aui/auibar.h
```diff
@@ -560,6 +592,7 @@ class WXDLLIMPEXP_AUI wxAuiToolBar : public wxControl
     bool GetToolFits(int toolId) const;
     wxRect GetToolRect(int toolId) const;
     bool GetToolFitsByIndex(int toolId) const;
+    bool GetToolBarItemFits(wxAuiToolBarItem* item) const;
```
vadz, line 595: Should it be
```suggestion
    bool GetToolBarItemFits(const wxAuiToolBarItem* item) const;
```
?

vadz (commented):
Thanks for the PR, but it seems like this would require changing _all_ the existing code using `wxAuiToolBar`, wouldn't it? If so, i.e. unless I'm missing something, it seems like a too high price to pay.

BTW, if I do miss something and the compatibility breakage is somehow not as bad as it looks, I think we should keep using `wxAuiToolBarItemArray` instead of adding another similar-but-different container class.

But right now I'm 80% sure that we shouldn't break all the existing code like this and we could find some other way to achieve what you want (see my reply to wx-dev).

We could (and should) still apply `wxAuiToolBarItemKind` part of the changes however.

imciner2:
> but it seems like this would require changing all the existing code using wxAuiToolBar

No, this won't require changing existing wxAuiToolBar code that uses the public API - as I said in the initial PR description. The only place in `wxAuiToolBar` this internal type (`wxAuiToolBarItemArray`) was exposed to users was in the `SetCustomOverflowItems` function, which took two arguments of type `wxAuiToolBarItemArray` (one for prepend and one for append). I have kept that version of the function (but marked it as deprecated in the docs) and added a new overload that takes the new `wxAuiToolBarItemList` type for both arguments - so old code setting the overflow items using the old type will still work by calling the existing function, and new code using the new type will call the new overload.

The only place the change breaks the public API is in the function `virtual int wxAuiToolBarArt::ShowDropDown(wxWindow* wnd, const wxAuiToolBarItemList& items) = 0;`, where I replaced the previous version using an argument of type `const wxAuiToolBarItemArray& items` with one of type `const wxAuiToolBarItemList& items`. This one is more tricky to preserve the API for, because it is actually called from the wxAuiToolBar code itself, so after this change we expect the `wxAuiToolBarItemList` version. This is IMO not a large change for downstream users to make, since the number doing custom toolbar art with this function is fairly low, and as a quick experiment, I used the new GitHub code search to look for uses of:

* `: public wxAuiToolBarArt` - no uses outside of people including the upstream wx header in their repositor
* `: public wxAuiGenericToolBarArt` - Only use I found is in https://github.com/open-enterprise-solutions/enterprise/blob/1e399ddf028e8d5e6a453ccd8be8d00210bd49e6/src/engine/core/frontend/theme/luna_auitoolbar.h (they have a submodule pinned to 3.2.2 though, so they seem to be able to update at their own pace)
* `: public wxAuiDefaultToolBarArt` - 35 results, only one I found overloading the specific `ShowDropDown` function was https://github.com/yazici/ginkgocadx/blob/d9241fc2919aec369436c6d82f092f00829d7fc0/cadxcore/wx/wxginkgotoolbar.cpp#L89, which actually looks very similar to the version in the base class (so I can't pick out the difference at first glance).
* `: public wxAuiMSWToolBarArt` - No results

---
Note: To see the public API compatibility for old code and the changes here, you can take the AUI sample from the master branch and replace the AUI sample in this PR, and it still builds and works just fine - no changes needed.

---
Where it gets trickier is the change to the internal type, moving it from `wxAuiToolBarItemArray` to `wxAuiToolBarItemList`. This will affect any downstream users who inherit from `wxAuiToolBar` and then directly access the `m_items` object in the toolbar class. I did another GitHub code search for inheritance from the wxAuiToolBar (e.g. searching for `: public wxAuiToolBar`), and out of around 150 results, the vast majority of subclasses just contain functions calling things like `AddTool`, `AddSeparator`, etc. to do initialization of the toolbars. The cases of code directly accessing the `m_items` member variable that is changing type that I found are:

* https://github.com/manyxu/Phoenix3D_2.1/blob/bfb2e3757bf61ac461aeeda9216bf8c8fdf76d99/Phoenix3D/Tools/Nirvana/PX2wxAui.cpp - a getter function and then a loop over all items.
* https://github.com/open-enterprise-solutions/enterprise/blob/1e399ddf028e8d5e6a453ccd8be8d00210bd49e6/src/engine/core/frontend/controls/toolBar.h#L6 - two functions to insert tools/separators at specific points in the toolbar (this access could be fixed by adding insert functions to the `wxAuiToolBar`, which would also bring it into better feature parity with `wxToolBar`, which has those functions).
* https://github.com/yazici/ginkgocadx/blob/d9241fc2919aec369436c6d82f092f00829d7fc0/cadxcore/wx/wxginkgotoolbar.cpp#L315 - direct access to the tool to implement programmatic setting of radio/check tools.
* https://github.com/yazici/ginkgocadx/blob/d9241fc2919aec369436c6d82f092f00829d7fc0/cadxcore/wx/wxginkgotoolbar.cpp#L381 - Custom paint loop that removes dropdown buttons and just draws normal/check/radio items all the same.
* https://github.com/acotty/CodeBlocks_Unofficial_Testing/blob/37cb5bc6fead6e6b3f0a10320a7c6d2070f4ef7f/src/plugins/contrib/wxSmithAui/wxAuiToolBar/wxSmithAuiToolBar.cpp#L40 - A hit test routine to find the tool at a given point
* https://github.com/kirixresearch/strata/blob/85612770e903f855e9b5c9c9d8b4dc051ecc7d38/appmain/linkbar.cpp - Quite a few uses of `m_items` scattered around, but they appear to have made their own member variable called `m_items` that will shadow the base class version, so I don't think these uses will be affected.
---

>  I think we should keep using wxAuiToolBarItemArray instead of adding another similar-but-different container class.

Doing this would completely break compatibility, because then we lose the ability to have the `wxAuiToolBar::SetCustomOverflowItems` function work with pre 3.3.0 code. To preserve that compatibility, we need two different types (keep the old and add the new).

include/wx/aui/auibar.h
imciner2: `int` is used to preserve compatibility with code using the old `wxITEM_*` identifiers here. Changing the type of the argument to `wxAuiToolBarItemKind` forces the use of `wxAUI_TB_ITEM_*` constants and causes a compile error when any `wxITEM_*` constants are used, because they can't be implicitly converted between the enums. Both enums can be implicitly converted to `int` though, so changing the `AddTool` functions to take `int` allows it to accept both the new `wxAUI_TB_ITEM_*` constants and the old `wxITEM_*` constants that existing code is using. (and note that `wxAuiToolBarItem` already treats kind as an `int`` in its API and storage, so this is the only change needed to make it work).

include/wx/aui/auibar.h
imciner2: So new containers should all be standard library containers now?

Even though the toolbar should own the items, I don't think using the `unique_ptr` is going to work, since we have to pass this object down to the toolbar art provider to draw the dropdown list. A pure vector with the toolbar itself doing the memory management would work.

vadz: > So new containers should all be standard library containers now?

Yes, there is no reason not to use them any more as we require C++11 and this means we can count on them being available.

include/wx/aui/auibar.h
```diff
@@ -345,7 +365,7 @@ class WXDLLIMPEXP_AUI wxAuiToolBarArt
 
     virtual int ShowDropDown(
                          wxWindow* wnd,
-                         const wxAuiToolBarItemArray& items) = 0;
+                         const wxAuiToolBarItemList& items) = 0;
```
vadz, line 368: [It looks like](https://cs.github.com/?scopeName=All+repos&scope=&q=wxAuiToolBar+ShowDropDown+-path%3Awx%2Faui%2Fauibar.h+-path%3Awx%2Faui%2Fbarartmsw.h+-path%3Asrc%2Faui%2Fauibar.cpp+-path%3Asrc%2Faui%2Fbarartmsw.cpp) this change is going to break quite some code, as the existing overloads won't be called any longer.

include/wx/aui/auibar.h
vadz: Yes, but this is really a pity as it's not type-safe at all. Could we provide both overloads, taking the old and the new enum? It's annoying, of course, but I think this would be worth doing.

vadz:
> > but it seems like this would require changing all the existing code using wxAuiToolBar
> 
> No, this won't require changing existing wxAuiToolBar code that uses the public API

OK, sorry, "all" was clearly a dramatic exaggeration, but it seems like quite a lot of existing code will be affected. Before doing into details, I'd like to reiterate the question asked [in this post](https://groups.google.com/g/wx-dev/c/zv9PhneifbA/m/M0i8WoyKAQAJ): why do we need to make all these changes if it looks like the same goal could be easily achieved simply by using the client data inside the tool?

> The only place the change breaks the public API is in the function `virtual int wxAuiToolBarArt::ShowDropDown(wxWindow* wnd, const wxAuiToolBarItemList& items) = 0;`, where I replaced the previous version using an argument of type `const wxAuiToolBarItemArray& items` with one of type `const wxAuiToolBarItemList& items`. This one is more tricky to preserve the API for, because it is actually called from the wxAuiToolBar code itself,

In principle this is possible but it would require making a temporary copy of `m_items` just to call the deprecated function, so this is going to be ugly and inefficient.

> This is IMO not a large change for downstream users to make, since the number doing custom toolbar art with this function is fairly low, 

I think you might be underestimating this number, see the link in my comment. And not all code is on GitHub, of course, so, typically, if I find more than one or two occurrences on GitHub using some code it's a good indication that non-trivial amount of people use it.

> Where it gets trickier is the change to the internal type, moving it from `wxAuiToolBarItemArray` to `wxAuiToolBarItemList`. This will affect any downstream users who inherit from `wxAuiToolBar` and then directly access the `m_items` object in the toolbar class. I did another GitHub code search for inheritance from the wxAuiToolBar 

[My own search](https://cs.github.com/?scopeName=All+repos&scope=&q=wxAuiToolBar+m_items+-path%3Awx%2Faui%2Fauibar.h+-path%3Awx%2Faui%2Fbarartmsw.h+-path%3Asrc%2Faui%2Fauibar.cpp+-path%3Asrc%2Faui%2Fbarartmsw.cpp) finds ~100 of occurrences which is less than 150 you found, but IMO is still quite sufficient to avoid breaking all of it unless we really have no choice.

And, to round back to the beginning, it looks like we do have the choice of using the client data, so why not do it?

> > I think we should keep using wxAuiToolBarItemArray instead of adding another similar-but-different container class.
> 
> Doing this would completely break compatibility, because then we lose the ability to have the `wxAuiToolBar::SetCustomOverflowItems` function work with pre 3.3.0 code. To preserve that compatibility, we need two different types (keep the old and add the new).

OK, yes, this is a good point. But having 2 similar containers is still ugly.

vadz:
To return to this, unless there are some crucial advantage of this compared to using client data, I'm afraid that applying it is not worth the breakage, sorry.

Please reopen if you think this is wrong, but I'd really like to see at least some (and preferably convincing) arguments to justify doing this.

Issue #23314, opened 2023-03-05
Clang ARM64 builds are defining -mthreads in wx-config, but that's an unsupported switch


hmartinez82:
### Description
In MSYS2's CLANGARM64 environment downstream deps are failing to build when they are restrictive towards unsupported command line switches. See https://github.com/msys2/MINGW-packages/issues/15825

I think that the issue comes from this line:
https://github.com/wxWidgets/wxWidgets/blob/master/configure.in#L4631
It's only honoring `-mthreads` not being supported in MINGW64, but it should also not add `-mthreads` when the host is `aarch64-w64-mingw32`

### Platform and version information

- wxWidgets version 3.2.2.1:
- wxMSW
- OS MSYS2 under Windows 11 10.0.22621
- Compiler: 
```
clang version 15.0.7
Target: aarch64-w64-windows-gnu
Thread model: posix
```

hmartinez82:
Correct. When I meant `restrictive towards unsupported command line switches` that's exactly what I meant :)
KiCAD was such an example. I think that `-mthreads`` is only used in 32 bit x86? (Not exactly my expertise these command line switches)

vadz:
It looks like this should have been closed by the commits above, but was just forgotten. Closing it now, please let me know if the problem still exists.

PR #23316, opened 2023-03-06
Don't use `-mthreads` option when building with MinGW

vadz:
This option seems to be obsolete and was only ever needed for classic MinGW32 which is not even supported any more.

This basically reverts a0b9e27fd4 (Applied Fabian Wenzel's patches to add -mthreads for threads to work in MinGW., 2002-08-28).

See #23314.

### DIFF
--- a/configure
+++ b/configure
@@ -35027,53 +35027,6 @@ $as_echo "$wx_cv_type_abi_forced_unwind" >&6; }
         fi
       fi
     fi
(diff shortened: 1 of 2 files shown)


vadz:
I'm still going to merge this one because AFAIK `-mthreads` is just not useful any longer. And if I'm wrong, we can always revert this later.

Issue #23315, opened 2023-03-06
Overriding wxApp::CreateTraits() has no effect under MSW


rcane:
### Description

Since 94e4818 an overridden `CreateTraits()` in a client application is no longer called.

The reason for this is the `if ( !wxSystemOptions::GetOptionInt("msw.no-manifest-check") )` line in the constructor of `wxApp`.
This produces the following method calls:
- `wxSystemOptions::GetOptionInt()`
- `wxSystemOptions::GetOption()`
- `wxTheApp->GetAppName()`
- `wxStandardPathsBase::Get()`
- `wxAppConsoleBase::GetTraitsIfExists()`
- `wxAppConsoleBase::GetTraits()`

The last method finally calls `CreateTraits()`. But since we are coming from the `wxApp` constructor an overridden `CreateTraits()` in a derived class is never called. Instead, `wxAppBase::CreateTraits()` is always called.

Although I have not tried this yet, a possible fix for this would be to move this manifest check from the constructor into `wxApp::Initialize()`.

### Platform and version information

- wxWidgets version: 3.3.0
- wxWidgets port: wxMSW
- OS: Windows 10

vadz:
Thanks for reporting this! The problem can indeed be fixed by doing this later, but it's not the first time this happens (see at least bc334f39ec4479e43e6aecd3925bd3dc516e26b8) and I'd like to find a way to prevent this from happening again in the future, so I've done this in 2be5f429c19cbcc945993112ea484cc579aa0e99, and then slightly improved it in d09245429a8b1dc7f9f56604526682fed948911c, which are both part of PR #23319 now (and which also includes your original suggestion for the fix as fc409d2c569b2161676a67eac6db7ffaa595788f because it's still better to do it there). Unfortunately it's rather clumsy, but I just don't see any better way to do it.

If there are no objections/problems, I'm going to merge this soon.

PR #23319, opened 2023-03-06
Improvements to `wxApp` initialization

vadz:
This is mostly about fixing #23315 (and ensuring that it doesn't happen again), but also includes some cleanup including getting rid of `wx{DECLARE,DEFINE}_SCOPED_PTR` macros globally (and not only in `wxApp` code).


rcane:
I just tested this PR. It does fix #23315 for me.

Thanks for solving this so quickly.

vadz:
Thanks for testing, will merge in a moment.

PR #23325, opened 2023-03-07
Replace -1 with wxCOL_WIDTH_DEFAULT in wxDataViewColumn widths

wh11204:
Also mention wxCOL_WIDTH_AUTOSIZE

### DIFF
--- a/include/wx/dataview.h
+++ b/include/wx/dataview.h
@@ -571,11 +571,11 @@ public:
     // -----------------
 
     wxDataViewColumn *PrependTextColumn( const wxString &label, unsigned int model_column,
-                    wxDataViewCellMode mode = wxDATAVIEW_CELL_INERT, int width = -1,
+                    wxDataViewCellMode mode = wxDATAVIEW_CELL_INERT, int width = wxCOL_WIDTH_DEFAULT,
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, will (squash) merge soon. Ideal would be to also explain the difference between `wxDVC_DEFAULT_WIDTH` and `wxCOL_WIDTH_DEFAULT` but I don't even know what it is myself without looking...

wh11204:
`wxDVC_DEFAULT_WIDTH` is a fixed positive number (80) used for text columns.

vadz:
Yes, sorry, I know/see that much. But it's not clear at all why and when would you use it instead of `wxCOL_WIDTH_DEFAULT` which should, I think, result in roughly the same width.

In fact the real question is whether we shouldn't just make `wxDVC_DEFAULT_WIDTH` a (deprecated) synonym for `wxCOL_WIDTH_DEFAULT`. I'd like to do it because `wxDVC_DEFAULT_WIDTH` doesn't even use logical pixels and so works differently depending on the current DPI which is clearly a bug.

vadz:
I still think we need to do something about `wxDVC_DEFAULT_WIDTH`, but for now I've merged these changes, as they're already an improvement, thanks!

PR #23334, opened 2023-03-09
Add wxWebView child window handling

TcT2k:
When a javascript call to `window.open()` triggers a new window it can manipulate that window and it's contents in a browser.

The current implementation of `wxEVT_WEBVIEW_NEWWINDOW` didn't enable this functionality in `wxWebView`, as there couldn't be connection established in the underlying API to connect two `wxWebView` instances.

This PR adds `wxWebViewEvent::GetTargetWindowInfo()` and a new class `wxWebViewWindowInfo` which provides information about the options from a java script `window.open()` call. It also includes the method `CreateChildWebView()` which **must** be called in the event handler to create the target web view. A web view created this way can than be access from the calling java script just like in a browser. Without a call to `CreateChildWebView()` the event handler behaves the same as the current implementation.

Sample code of a new window event handler:
```C++
m_webView->Bind(wxEVT_WEBVIEW_NEWWINDOW, [](wxWebViewEvent& evt) {
    // Get target window info
    wxWebViewWindowInfo* info = evt.GetTargetWindowInfo();
    // Create a top level window for the child web view
    wxWindow* win = new wxWindow(this, wxID_ANY, info->GetPosition(), info->GetSize());
    wxSizer* sizer = new wxBoxSizer(wxVERTICAL);
    // Create the child web view
    wxWebView* childWebView = info->CreateChildWebView();
    // Place the child web view into the window
    childWebView->Create(win, wxID_ANY);
    sizer->Add(childWebView, 1, wxEXPAND);
});
```

This is currently implemented for Edge and WKWebView.

I would think it could also be implemented for GTK WebKit but I would either add it in another PR or hope that somebody else does it :)


vadz:
Thanks, looks good to me, so I'm going to merge it and hope that someone can indeed come up with an implementation for GTK WebKit later, as this definitely looks like useful functionality.

PR #23335, opened 2023-03-09
Add CMakePresets.json

TcT2k:
I've recently started using [CMakePresets.json](https://cmake.org/cmake/help/v3.26/manual/cmake-presets.7.html) and I think it would be a good addition to include in the wxWidgets repository. (Additional introduction in the [Visual Studio Blog](https://devblogs.microsoft.com/cppblog/cmake-presets-integration-in-visual-studio-and-visual-studio-code/))

Basically what it does is to preconfigure various cmake configuration/generation settings.
This allows for easier CMake usage from the CMake GUI, Visual Studio and Visual Studio Code. Additionally it also allows to specify these presets from the cmake CLI via the --preset command line.

I think this simplifies configuring builds significantly for everyone especially for newcomers (to cmake and/or wxWidgets)
I've included a few presets I think could be useful for contributors and library users alike.

Obviously these should be expanded to contain some other common configurations (Linux, MinGW, ?). Additionally presets for offical binaries could make sense.
But I wanted to check what others think before adding more.

Here are a few screenshots how the presets are currently presented in GUI tools:

How it looks in CMake GUI:
image

In Visual Studio Code (when opening the wxWidgets folder):
image

In Visual Studio (when opening the wxWidgets folder):
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -402,3 +402,6 @@
 /utils/wxrc/wxrc_vc[789].sln
 
 /3rdparty/webview2
+
+/build/cmake/output
+CMakePresetsUser.json
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,136 @@
+{
+    "version": 3,
+    "cmakeMinimumRequired": {
+        "major": 3,
+        "minor": 21,
+        "patch": 0
+    },
+    "configurePresets": [
+        {
+            "name": "common",
+            "hidden": true,
+            "binaryDir": "${sourceDir}/build/cmake/output/${presetName}"
+        },
+        {
+            "name": "with_samples",
+            "hidden": true,
+            "cacheVariables": {
+                "wxBUILD_SAMPLES": "ALL"
+            }
+        },
+        {
+            "name": "msw_base",
+            "description": "Base configuration for MSW",
+            "hidden": true,
+            "inherits": "common",
+            "cacheVariables": {
+                "wxUSE_WEBVIEW_EDGE": "ON"
+            },
+            "condition": {
+                "type": "equals",
+                "lhs": "${hostSystemName}",
+                "rhs": "Windows"
+            }
+        },
+        {
+            "name": "msw_vc17_samples",
+            "inherits": [
+                "with_samples",
+                "msw_base"
+            ],
+            "displayName": "MSW Visual Studio 2022 (with Samples)",
+            "generator": "Visual Studio 17 2022"
+        },
+        {
+            "name": "msw_vc17",
+            "inherits": "msw_base",
+            "displayName": "MSW Visual Studio 2022",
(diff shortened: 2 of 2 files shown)


CMakePresets.json
```diff
@@ -0,0 +1,136 @@
+{
+    "version": 3,
+    "cmakeMinimumRequired": {
+        "major": 3,
+        "minor": 21,
```
vadz, line 5: This is a much higher version than what we require otherwise, do we really need it? I have no problem with this, just curious. FWIW [MSVS docs](https://learn.microsoft.com/en-us/cpp/build/cmake-presets-vs) only recommend requiring 3.14.

.gitignore
```diff
@@ -402,3 +402,8 @@
 /utils/wxrc/wxrc_vc[789].sln
 
 /3rdparty/webview2
+
+/.vs
+/.vscode
```
vadz, line 407: Do we have to have those, i.e. are they somehow related to presets? As usual, I'd rather avoid adding here anything that should be in the global `.gitignore` anyhow.

.gitignore
```diff
@@ -402,3 +402,8 @@
...
+/build_cmake
```
vadz, line 408: A nitpick, but I'd prefer to use `build/cmake/build` or something like this (it's unfortunate that `build` is redundant, but I'd still like to keep everything CMake-related under `build/cmake`).

vadz (commented):
Thanks, I have no experience using CMake presets, so I don't know how useful this will be, but anything that makes wx easier to build for the users is welcome, of course.

One thing which bothers me is having all the MSVS versions here, but I guess we can't avoid it...

CMakePresets.json
TcT2k: This is independent of the version required to build. Version 3.21 is simply the first version to implement support for presets. (Technically it was 3.20 but the condition feature was added in 3.21)

.gitignore
TcT2k: Removed, was unrelated

.gitignore
TcT2k: I've changed it to `build/cmake/output`

vadz:
This could be nice to have in 3.3.0, @MaartenBent do you have any thoughts/opinions about whether we should merge it in its current state?

MaartenBent:
I'd remove the old MSVS presets and only keep vs2022.
And maybe rename `macOS Make` to `macOS Makefiles`? Not sure what the preferred terminology is.

TcT2k:
It's been a while since I added this, but I think it still would make sense for the same reasons stated in the initial comment.

And @MaartenBent  proposed changes definitively make sense.

vadz:
Thanks, I'll remove the MSVS non-2022 presets then (but still keep the base configs, just in case we need to reuse them in the future) and push this soon.

PR #23337, opened 2023-03-10
Move wxPropertyGrid internal stuff to the separate private header

a-wi:


k-serg:
My own code not build anymore, after this commit. Since I'm using **nmake** and **makefile.vc** which not updated by commiter. 
@vadz @a-wi  Can I ask you both, it is Ok to commit into master (that use by some people as production) incomplete things?

vadz:
@k-serg You need to report the errors you're getting to allow us to fix them. Nobody commits incomplete changes to master intentionally and most of the time master does build, but there are sometimes unexpected problems. Except that right now the CI builds pass, including [those using nmake](https://ci.appveyor.com/project/wxWidgets/wxwidgets/builds/46477941/job/he5nibgscusol2y6), so I really have no idea what is the problem. Please open an issue for it with all the details. TIA!

k-serg:
Fixed build issue by adding: 
`#include <wx/propgrid/property.h>`
in addition in:
```
#include <wx/propgrid/propgrid.h>
#include <wx/propgrid/advprops.h>
```
It's quite old code for my part, but it works for decade, and as say old rule: "if it works, don't touch it". Looks like it's time to touch my old code, ha-ha.

@vadz @a-wi  Really sorry, for any concern, which i caused. Accept my apologies.

PR #23338, opened 2023-03-10
Print current page option added to use current page option in Windows print dialog

sz5000:
The pull request makes it possible to use the print current page option of the Windows print dialog.

image

I have added three new members to wxPrintDialogData which can be used to control the new feature:
`int m_printCurrentPage` - the page number of the current page
`bool m_printEnableCurrentPage` - used to enable/disable the current page field in print dialog
`bool m_printSelectCurrentPage` - if set to true, the current page field is selected when the dialog is shown

By default the print current page option is disabled. Therefore the extension is 100% compatible with the old code.

This extension uses the new properties only for the Windows port. I don't know whether an option for a current page is also available in other dialogs so that these ones could be extended as well.

### DIFF
--- a/include/wx/cmndata.h
+++ b/include/wx/cmndata.h
@@ -160,6 +160,7 @@ public:
     int GetMinPage() const { return m_printMinPage; }
     int GetMaxPage() const { return m_printMaxPage; }
     int GetNoCopies() const { return m_printNoCopies; }
+    int GetCurrentPage() const { return m_printCurrentPage; }
     bool GetAllPages() const { return m_printAllPages; }
     bool GetSelection() const { return m_printSelection; }
     bool GetCollate() const { return m_printCollate; }
@@ -170,20 +171,25 @@ public:
     void SetMinPage(int v) { m_printMinPage = v; }
     void SetMaxPage(int v) { m_printMaxPage = v; }
     void SetNoCopies(int v) { m_printNoCopies = v; }
+    void SetCurrentPage(int v) { m_printCurrentPage = v; }
     void SetAllPages(bool flag) { m_printAllPages = flag; }
     void SetSelection(bool flag) { m_printSelection = flag; }
     void SetCollate(bool flag) { m_printCollate = flag; }
     void SetPrintToFile(bool flag) { m_printToFile = flag; }
+    void SetSelectCurrentPage(bool flag) { m_printSelectCurrentPage = flag; }
 
     void EnablePrintToFile(bool flag) { m_printEnablePrintToFile = flag; }
     void EnableSelection(bool flag) { m_printEnableSelection = flag; }
     void EnablePageNumbers(bool flag) { m_printEnablePageNumbers = flag; }
     void EnableHelp(bool flag) { m_printEnableHelp = flag; }
+    void EnableCurrentPage(bool flag) { m_printEnableCurrentPage = flag; }
 
     bool GetEnablePrintToFile() const { return m_printEnablePrintToFile; }
     bool GetEnableSelection() const { return m_printEnableSelection; }
     bool GetEnablePageNumbers() const { return m_printEnablePageNumbers; }
     bool GetEnableHelp() const { return m_printEnableHelp; }
+    bool GetEnableCurrentPage() const { return m_printEnableCurrentPage; }
+    bool GetSelectCurrentPage() const { return m_printSelectCurrentPage; }
 
     // Is this data OK for showing the print dialog?
     bool Ok() const { return IsOk(); }
@@ -201,6 +207,7 @@ private:
     int             m_printMinPage;
     int             m_printMaxPage;
     int             m_printNoCopies;
+    int             m_printCurrentPage;
     bool            m_printAllPages;
     bool            m_printCollate;
     bool            m_printToFile;
@@ -209,6 +216,8 @@ private:
     bool            m_printEnablePageNumbers;
     bool            m_printEnableHelp;
     bool            m_printEnablePrintToFile;
+    bool            m_printEnableCurrentPage;
+    bool            m_printSelectCurrentPage;
     wxPrintData     m_printData;
 
 private:
--- a/include/wx/cmndata.h
+++ b/include/wx/cmndata.h
@@ -160,36 +160,34 @@ public:
     int GetMinPage() const { return m_printMinPage; }
     int GetMaxPage() const { return m_printMaxPage; }
     int GetNoCopies() const { return m_printNoCopies; }
-    int GetCurrentPage() const { return m_printCurrentPage; }
     bool GetAllPages() const { return m_printAllPages; }
     bool GetSelection() const { return m_printSelection; }
     bool GetCollate() const { return m_printCollate; }
     bool GetPrintToFile() const { return m_printToFile; }
+    bool GetCurrentPage() const { return m_printCurrentPage; }
 
     void SetFromPage(int v) { m_printFromPage = v; }
     void SetToPage(int v) { m_printToPage = v; }
     void SetMinPage(int v) { m_printMinPage = v; }
     void SetMaxPage(int v) { m_printMaxPage = v; }
     void SetNoCopies(int v) { m_printNoCopies = v; }
-    void SetCurrentPage(int v) { m_printCurrentPage = v; }
     void SetAllPages(bool flag) { m_printAllPages = flag; }
     void SetSelection(bool flag) { m_printSelection = flag; }
     void SetCollate(bool flag) { m_printCollate = flag; }
     void SetPrintToFile(bool flag) { m_printToFile = flag; }
-    void SetSelectCurrentPage(bool flag) { m_printSelectCurrentPage = flag; }
+    void SetCurrentPage(bool flag) { m_printCurrentPage = flag; }
 
     void EnablePrintToFile(bool flag) { m_printEnablePrintToFile = flag; }
     void EnableSelection(bool flag) { m_printEnableSelection = flag; }
     void EnablePageNumbers(bool flag) { m_printEnablePageNumbers = flag; }
-    void EnableHelp(bool flag) { m_printEnableHelp = flag; }
     void EnableCurrentPage(bool flag) { m_printEnableCurrentPage = flag; }
+    void EnableHelp(bool flag) { m_printEnableHelp = flag; }
 
     bool GetEnablePrintToFile() const { return m_printEnablePrintToFile; }
     bool GetEnableSelection() const { return m_printEnableSelection; }
+    bool GetEnableCurrentPage() const { return m_printEnableCurrentPage; }
     bool GetEnablePageNumbers() const { return m_printEnablePageNumbers; }
     bool GetEnableHelp() const { return m_printEnableHelp; }
-    bool GetEnableCurrentPage() const { return m_printEnableCurrentPage; }
-    bool GetSelectCurrentPage() const { return m_printSelectCurrentPage; }
 
     // Is this data OK for showing the print dialog?
     bool Ok() const { return IsOk(); }
@@ -207,17 +205,16 @@ private:
     int             m_printMinPage;
     int             m_printMaxPage;
     int             m_printNoCopies;
-    int             m_printCurrentPage;
     bool            m_printAllPages;
+    bool            m_printSelection;
+    bool            m_printCurrentPage;
     bool            m_printCollate;
     bool            m_printToFile;
-    bool            m_printSelection;
     bool            m_printEnableSelection;
     bool            m_printEnablePageNumbers;
+    bool            m_printEnableCurrentPage;
     bool            m_printEnableHelp;
     bool            m_printEnablePrintToFile;
-    bool            m_printEnableCurrentPage;
-    bool            m_printSelectCurrentPage;
     wxPrintData     m_printData;
 
 private:
--- a/include/wx/prntbase.h
+++ b/include/wx/prntbase.h
@@ -278,6 +278,7 @@ public:
     virtual bool HasPage(int page);
     virtual bool OnPrintPage(int page) = 0;
     virtual void GetPageInfo(int *minPage, int *maxPage, int *pageFrom, int *pageTo);
+    virtual bool IsPageSelected(int page);
 
     virtual wxString GetTitle() const { return m_printoutTitle; }
 
--- a/interface/wx/cmndata.h
+++ b/interface/wx/cmndata.h
@@ -607,9 +607,12 @@ public:
     void SetSelection(bool flag);
 
     /**
-        Selects the "Current Page" radio button. The effect of printing the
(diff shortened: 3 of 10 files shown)


src/msw/printdlg.cpp
```diff
@@ -911,12 +927,19 @@ bool wxWindowsPrintDialog::ConvertToNative( wxPrintDialogData &data )
     if ( data.GetEnableHelp() )
         pd->Flags |= PD_SHOWHELP;
 
+#if(WINVER >= 0x0500)
```
vadz, line 930: We never use `WINVER` checks in wx code. This particular one is obsolete and can be simply removed anyhow because we only support Windows 7 or later (and probably won't support even Windows 7 for much longer) in master, but if you ever need to perform any checks, they need to be done during run-time and not compile-time because we want the same wx DLL to work on all Windows versions.

src/msw/printdlg.cpp
```diff
@@ -964,6 +991,11 @@ bool wxWindowsPrintDialog::ConvertFromNative( wxPrintDialogData &data )
     data.EnablePageNumbers( ((pd->Flags & PD_NOPAGENUMS) != PD_NOPAGENUMS) );
     data.EnableHelp( ((pd->Flags & PD_SHOWHELP) == PD_SHOWHELP) );
 
+#if(WINVER >= 0x0500)
```
vadz, line 994: Another check which should be removed.

src/msw/printdlg.cpp
```diff
@@ -852,14 +857,16 @@ bool wxWindowsPrintDialog::ConvertToNative( wxPrintDialogData &data )
...
-    PRINTDLG *pd = (PRINTDLG*) m_printDlg;
+    PRINTDLGEX* pd = (PRINTDLGEX*) m_printDlg;
 
     // Shouldn't have been defined anywhere
     if (pd)
         return false;
 
-    pd = new PRINTDLG;
-    memset( pd, 0, sizeof(PRINTDLG) );
+    pd = (LPPRINTDLGEX) GlobalAlloc(GPTR, sizeof(PRINTDLGEX));
```
vadz, line 866: I don't really mind doing this, but I just wonder why did you decide to change this, i.e. switch from `new` to `GlobalAlloc` here? There doesn't seem to be any reason to use the Windows function as we only free this one ourselves, so why?

src/msw/printdlg.cpp
```diff
@@ -964,6 +991,11 @@ bool wxWindowsPrintDialog::ConvertFromNative( wxPrintDialogData &data )
     data.EnablePageNumbers( ((pd->Flags & PD_NOPAGENUMS) != PD_NOPAGENUMS) );
     data.EnableHelp( ((pd->Flags & PD_SHOWHELP) == PD_SHOWHELP) );
 
+#if(WINVER >= 0x0500)
+    data.SetSelectCurrentPage(((pd->Flags & PD_CURRENTPAGE) == PD_CURRENTPAGE));
+    data.EnableCurrentPage(((pd->Flags & PD_NOCURRENTPAGE) != PD_NOCURRENTPAGE));
```
vadz, line 996: I don't think this can be changed by user, can it? I.e. I think this line is just unnecessary.

include/wx/cmndata.h
```diff
@@ -170,20 +171,25 @@ class WXDLLIMPEXP_CORE wxPrintDialogData: public wxObject
...
     void SetSelection(bool flag) { m_printSelection = flag; }
     void SetCollate(bool flag) { m_printCollate = flag; }
     void SetPrintToFile(bool flag) { m_printToFile = flag; }
+    void SetSelectCurrentPage(bool flag) { m_printSelectCurrentPage = flag; }
 
     void EnablePrintToFile(bool flag) { m_printEnablePrintToFile = flag; }
     void EnableSelection(bool flag) { m_printEnableSelection = flag; }
     void EnablePageNumbers(bool flag) { m_printEnablePageNumbers = flag; }
     void EnableHelp(bool flag) { m_printEnableHelp = flag; }
+    void EnableCurrentPage(bool flag) { m_printEnableCurrentPage = flag; }
```
vadz, line 185: This is very minor, but I'd put this just below `EnableSelection()` because these functions are related (while it's completely unrelated to `EnableHelp()`).

I'd also consider renaming it to `EnableSelectCurrentPage()` to be consistent with `SetSelectCurrentPage()`. Alternatively, the other function could be renamed to just `SetCurrentPage()` but this would be less clear.

include/wx/cmndata.h
```diff
@@ -160,6 +160,7 @@ class WXDLLIMPEXP_CORE wxPrintDialogData: public wxObject
     int GetMinPage() const { return m_printMinPage; }
     int GetMaxPage() const { return m_printMaxPage; }
     int GetNoCopies() const { return m_printNoCopies; }
+    int GetCurrentPage() const { return m_printCurrentPage; }
```
vadz, line 163: This seems completely unused to me, as `m_printCurrentPage` is never set (nor used) anywhere. What is the point of having this function? IMO it should be just removed.

include/wx/cmndata.h
```diff
@@ -170,20 +171,25 @@ class WXDLLIMPEXP_CORE wxPrintDialogData: public wxObject
     void SetMinPage(int v) { m_printMinPage = v; }
     void SetMaxPage(int v) { m_printMaxPage = v; }
     void SetNoCopies(int v) { m_printNoCopies = v; }
+    void SetCurrentPage(int v) { m_printCurrentPage = v; }
```
vadz, line 174: This should be removed too AFAICS.

include/wx/cmndata.h
```diff
@@ -170,20 +171,25 @@ class WXDLLIMPEXP_CORE wxPrintDialogData: public wxObject
...
     void EnableSelection(bool flag) { m_printEnableSelection = flag; }
     void EnablePageNumbers(bool flag) { m_printEnablePageNumbers = flag; }
     void EnableHelp(bool flag) { m_printEnableHelp = flag; }
+    void EnableCurrentPage(bool flag) { m_printEnableCurrentPage = flag; }
 
     bool GetEnablePrintToFile() const { return m_printEnablePrintToFile; }
     bool GetEnableSelection() const { return m_printEnableSelection; }
     bool GetEnablePageNumbers() const { return m_printEnablePageNumbers; }
     bool GetEnableHelp() const { return m_printEnableHelp; }
+    bool GetEnableCurrentPage() const { return m_printEnableCurrentPage; }
```
vadz, line 191: Similarly, I'd put this one right below `GetEnableSelection()`.

include/wx/cmndata.h
```diff
@@ -170,20 +171,25 @@ class WXDLLIMPEXP_CORE wxPrintDialogData: public wxObject
...
+    bool GetSelectCurrentPage() const { return m_printSelectCurrentPage; }
```
vadz, line 192: More importantly, I'd move this one to a different function group, together with `GetSelection()` -- as it's a completely kind of flag (this one is user input and not an option set by the program).

include/wx/cmndata.h
```diff
@@ -209,6 +216,8 @@ class WXDLLIMPEXP_CORE wxPrintDialogData: public wxObject
     bool            m_printEnablePageNumbers;
     bool            m_printEnableHelp;
     bool            m_printEnablePrintToFile;
+    bool            m_printEnableCurrentPage;
```
vadz, line 219: Again, for consistency this one should be
```suggestion
    bool            m_printEnableSelectCurrentPage;
```

vadz (changes requested):
Thanks for your work on this, I agree that it would be definitely useful to have support for this, but I think the proposed API is a bit confusing and should be simplified/streamlined a bit.

It would also be very nice if you could please modify the printing sample to show the new API in action -- it's always the best way to see if it works well.

And, finally, once the API is stabilized, with all the functions names settled, we absolutely need documentation for it, so please update `interface/wx/cmndata.h` with the new functions descriptions. Please use `@since 3.3.0` in them.

TIA!

src/msw/printdlg.cpp
sz5000: Ok, changed.

src/msw/printdlg.cpp
sz5000: Ok, changed.

src/msw/printdlg.cpp
sz5000: My last commit has changed this.

include/wx/cmndata.h
sz5000: Was not unsused, but my last commit has changed this too.

include/wx/cmndata.h
sz5000: Ok, changed.

include/wx/cmndata.h
sz5000: Ok, changed.

include/wx/cmndata.h
sz5000: Ok, changed

include/wx/cmndata.h
sz5000: Changed with my last commit.

src/msw/printdlg.cpp
sz5000: Changed.

vadz (commented):
Thanks a lot for the changes, I think the "current page" part is perfectly fine now, but it looks like there is also a new change, related to `IsPageSelected()` addition and now I don't understand how does this work... Who is supposed to override it? Shouldn't this be done by MSW printout using the information about multiple ranges from the dialog?

Again, showing the new functionality in the sample would probably be very useful to demonstrate how it works.

Also, could we perhaps leave this for another PR and merge just the current-page-related parts of this one?

TIA!

sz5000:
"Also, could we perhaps leave this for another PR and merge just the current-page-related parts of this one?"
Yes

IsPageSelected() is required for the selection mode, otherwise the print function does not know which page is selected because currently there is function to get this information. I changed the msw/printWin.cpp function and there IsPageSelected() is used. I would definitively recommend to improve this because currently the print selection mode is pretty useless.

```
if ( !(m_printDialogData.GetAllPages() || m_printDialogData.GetSelection()) )
{
	minPageNum = m_printDialogData.GetFromPage();
	maxPageNum = m_printDialogData.GetToPage();
}
```

minPageNum and maxPageNum is only updated if the print range mode is used. So in selection mode, every page is printed because there is no code that determines selected pages. Therefore the new IsPageSelected() and this new method fixes this. It should be overridden by a PrintOut if selection mode is enabled.

vadz:
I don't understand how is this fix sufficient. I.e. who is supposed to override `IsPageSelected()` and how? It seems to me that this part is still missing, but perhaps it's me who's missing something, could you please explain me how is it supposed to work? Again, an example of using this in the sample would probably do wonders.

But if it's really not fully implemented yet, I'd rather avoid committing incomplete fixes. I can exclude them manually when merging if you're ok with this.

sz5000:
Actually it's really simple. User clicks the pint function and selects the Selection radio button. Then the printing code needs some information from the app on which pages are selected so that only these pages are printed. This part is missing currently. In order so solve this, IsPageSelected() comes into action. If print selection is used, the printing code calls IsPageSelected with the page number and the printout has to respond either with true or false. If the return value it true, then the page is printed. If false is returned, then the page is skipped.

```
class MyPrintOut: public wxPrintout {
...
public:
	bool IsPageSelected(int page) {
		// return true if the page is selected in the APP/GUI, else flase
	}
...
};
```

This new method is used in printwin.cpp, line 207:

```
if (m_printDialogData.GetSelection() && !printout->IsPageSelected(pn))
                continue;
```

In wxWindowsPrinter class, if a document is printed, the implementation loops over the pages (minPageNum, maxPageNum) and prints the pages in that range but it does not check if a page in that range is selected or not. The two code lines above check that and skips pages that are not selected. How pages are selected is completely app specific.

sz5000:
I have also updated the printing sample which shows how to use the new current page feature and how to use the new IsPageSelected() function.

interface/wx/print.h
```diff
@@ -705,6 +705,13 @@ class wxPrintout : public wxObject
         and maximum page values that the user can select, and the required page range to
         be printed.
 
+        If the user has decided to print only selected pages, then pageFrom and pageTo are
```
vadz, line 708:
```suggestion
        If the user has decided to print only selected pages, then @a pageFrom and @a pageTo are
```

interface/wx/print.h
```diff
@@ -705,6 +705,13 @@ class wxPrintout : public wxObject
...
+        used to limit the page range and IsPageSelected() is used to query whether the page
+        is selected.
+
+        If the user only wants to print the current page, then pageFrom and pageTo are used
+        to set the current page by setting pageFrom = pageTo.
```
vadz, line 713:
```suggestion
        If the user only wants to print the current page, then @a pageFrom and @a pageTo are both set to the current page number.
```

vadz (commented):
Thanks, this is much more clear now and, except for some minor docs improvements, I think is ready to be merged.

I'm going to separate it in two commits, one adding current page selection and the other one adding `IsPageSelected()`, just to keep things clean, when merging.

Please let me know if you plan to make any more changes here, otherwise I'll try to do it (probably towards the end of the week).

Thanks again!

interface/wx/cmndata.h
```diff
@@ -489,6 +489,13 @@ class wxPrintDialogData : public wxObject
     */
     void EnableSelection(bool flag);
 
+    /**
+        Enables or disables the "Current Page" radio button.
+
```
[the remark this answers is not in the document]
sz5000: Sounds reasonable if other print dialogs don't have a Current Page option.

include/wx/prntbase.h
```diff
@@ -278,6 +278,7 @@ class WXDLLIMPEXP_CORE wxPrintout: public wxObject
     virtual bool HasPage(int page);
     virtual bool OnPrintPage(int page) = 0;
     virtual void GetPageInfo(int *minPage, int *maxPage, int *pageFrom, int *pageTo);
+    virtual bool IsPageSelected(int page);
```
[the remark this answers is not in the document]
sz5000: I also don't think IsPageSelected changes the Object. So const probably wouldn't be a problem. But in the end it depends more on how the app implements this function and I'm just wondering if we're taking too much liberty with const for a programmer.

sz5000:
I am happy with the current state and I don't see any more problems. So, please merge.

vadz:
Sorry for the delay, I've finally repackaged your changes in 2 separate commits in the linked PR (e93570f681 and 498ed666da). The PR also contains some cleanup which I did after looking at this code, but the commits above contain just your changes with some minor tweaks, mostly to avoid gratuitous differences with the existing code.

Please let me know if you see any problems with the other PR, TIA!

P.S. I'm closing this one as it's a subset of the other one.

Issue #23341, opened 2023-03-12
wxStaticText not wrapping text on Mac (Windows and Linux work OK)


lszl84:
### Description

The code below does not wrap the lines on Mac OS. Works correctly on Windows and Linux (see screenshots).

```cpp
#include <wx/wx.h>

class MyApp : public wxApp
{
public:
    virtual bool OnInit();
};

wxIMPLEMENT_APP(MyApp);

class MyFrame : public wxFrame
{
public:
    MyFrame(const wxString &title, const wxPoint &pos, const wxSize &size);
};

bool MyApp::OnInit()
{
    MyFrame *frame = new MyFrame("Hello World", wxDefaultPosition, wxDefaultSize);
    frame->Show(true);
    return true;
}

const auto paraText = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed euismod, nisl vitae ultricies lacinia, nisl nisl aliquam nisl, vitae aliquam";

MyFrame::MyFrame(const wxString &title, const wxPoint &pos, const wxSize &size)
    : wxFrame(nullptr, wxID_ANY, title, pos, size)
{
    auto paragraph = new wxStaticText(this, wxID_ANY, paraText);
    paragraph->SetFont(wxFont(20, wxFONTFAMILY_DEFAULT, wxFONTSTYLE_NORMAL, wxFONTWEIGHT_NORMAL));
    paragraph->SetBackgroundColour(wxColour(200, 100, 100));
}

```
<img width="529" alt="Screenshot 2023-03-12 at 20 11 03" src="https://user-images.githubusercontent.com/42072702/224567737-9ce3dfbb-49c5-4441-a360-39b528b52d27.png">
<img width="1712" alt="Screenshot 2023-03-12 at 20 09 51" src="https://user-images.githubusercontent.com/42072702/224567743-418aae58-2214-4056-b396-4c3693edcc9f.png">
<img width="1712" alt="Screenshot 2023-03-12 at 20 09 43" src="https://user-images.githubusercontent.com/42072702/224567749-a39251bc-305b-48d8-b2e0-07ebc321e98d.png">

### Platform and version information

- wxWidgets version  you use: 3.2.2
- wxWidgets port  you use: wxOSX
- OS  and its version: Ventura 13.2.1 (22D68)

vadz:
Apparently we need to use `setPreferredMaxLayoutWidth:` for this to work, but I'm really not sure. Maybe @csomor would have some idea?

[...]

PR #23345, opened 2023-03-13
Change Line Break mode of wxStaticText on macOS to match MSW and GTK

csomor:
See #23341


vadz:
Thanks, I've merged into master now and I'd be tempted to also cherry-pick it to 3.2, but this does change the existing behaviour, so I'm not 100% sure about it.

csomor:
> Thanks, I've merged into master now and I'd be tempted to also cherry-pick it to 3.2, but this does change the existing behaviour, so I'm not 100% sure about it.

@vadz yes, I'm a little bit reluctant too, I don't know whether we get really enough coverage when just having it in master, so I wouldn't just backport it 'lightly' ... perhaps when having a RC of a 3.2 somewhen ? or be prepared to add a fix release in case it causes havoc ?

PR #23349, opened 2023-03-14
Add wxWebViewConfiguration

TcT2k:
In 40ff38b63b594e6a82bf6f7845e0ac2cb8ae2449 `wxWebView::GetNativeConfiguration()` was
added which isn't suitable if multiple `wxWebViewEdge` instances are created with
different options (as they must share the same options). To prevent such errors
the previous method is replaced by the new class `wxWebViewConfiguration` which
can be shared between various `wxWebView` instances.

This explicitly breaks API to improve usage (previous API wasn't available before 3.3.0).

In the future the new class could also enable a good way to wrap various common options/configurations available via native API.

### DIFF
--- a/include/wx/msw/webview_edge.h
+++ b/include/wx/msw/webview_edge.h
@@ -25,6 +25,8 @@ public:
 
     wxWebViewEdge();
 
+    explicit wxWebViewEdge(const wxWebViewConfiguration& config);
+
     wxWebViewEdge(wxWindow* parent,
         wxWindowID id,
         const wxString& url = wxWebViewDefaultURLStr,
@@ -103,7 +105,6 @@ public:
     virtual void RegisterHandler(wxSharedPtr<wxWebViewHandler> handler) override;
 
     virtual void* GetNativeBackend() const override;
-    virtual void* GetNativeConfiguration() const override;
 
     static void MSWSetBrowserExecutableDir(const wxString& path);
 
@@ -128,6 +129,7 @@ class WXDLLIMPEXP_WEBVIEW wxWebViewFactoryEdge : public wxWebViewFactory
 {
 public:
     virtual wxWebView* Create() override { return new wxWebViewEdge; }
+    virtual wxWebView* CreateWithConfig(const wxWebViewConfiguration& config) override;
     virtual wxWebView* Create(wxWindow* parent,
         wxWindowID id,
         const wxString& url = wxWebViewDefaultURLStr,
@@ -140,6 +142,7 @@ public:
     }
     virtual bool IsAvailable() override;
     virtual wxVersionInfo GetVersionInfo() override;
+    virtual wxWebViewConfiguration CreateConfiguration() override;
 };
 
 #endif // wxUSE_WEBVIEW && wxUSE_WEBVIEW_EDGE && defined(__WXMSW__)
--- a/include/wx/osx/webview_webkit.h
+++ b/include/wx/osx/webview_webkit.h
@@ -28,27 +28,13 @@
 WX_DECLARE_STRING_HASH_MAP(wxSharedPtr<wxWebViewHandler>, wxStringToWebHandlerMap);
 
 class wxWebViewWindowInfoWebKit;
+class wxWebViewConfigurationImplWebKit;
 
 class WXDLLIMPEXP_WEBVIEW wxWebViewWebKit : public wxWebView
 {
 public:
-    wxDECLARE_DYNAMIC_CLASS(wxWebViewWebKit);
+    explicit wxWebViewWebKit(const wxWebViewConfiguration& config, wxWebViewWindowInfoWebKit* parentWindowInfo = nullptr);
 
-    wxWebViewWebKit(wxWebViewWindowInfoWebKit* parentWindowInfo = nullptr)
-    {
-        m_parentWindowInfo = parentWindowInfo;
-        Init();
-    }
-    wxWebViewWebKit(wxWindow *parent,
-                    wxWindowID winID = wxID_ANY,
-                    const wxString& strURL = wxASCII_STR(wxWebViewDefaultURLStr),
-                    const wxPoint& pos = wxDefaultPosition,
-                    const wxSize& size = wxDefaultSize, long style = 0,
-                    const wxString& name = wxASCII_STR(wxWebViewNameStr))
-    {
-        Init();
-        Create(parent, winID, strURL, pos, size, style, name);
-    }
     bool Create(wxWindow *parent,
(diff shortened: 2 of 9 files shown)


include/wx/msw/webview_edge.h
```diff
@@ -25,6 +25,8 @@ class WXDLLIMPEXP_WEBVIEW wxWebViewEdge : public wxWebView
 
     wxWebViewEdge();
 
+    wxWebViewEdge(std::shared_ptr<wxWebViewConfiguration> config);
```
vadz, line 28:
```suggestion
    explicit wxWebViewEdge(std::shared_ptr<wxWebViewConfiguration> config);
```

include/wx/osx/webview_webkit.h
```diff
@@ -28,27 +28,13 @@
 WX_DECLARE_STRING_HASH_MAP(wxSharedPtr<wxWebViewHandler>, wxStringToWebHandlerMap);
 
 class wxWebViewWindowInfoWebKit;
+class wxWebViewConfigurationWebKit;
 
 class WXDLLIMPEXP_WEBVIEW wxWebViewWebKit : public wxWebView
 {
 public:
-    wxDECLARE_DYNAMIC_CLASS(wxWebViewWebKit);
+    wxWebViewWebKit(std::shared_ptr<wxWebViewConfiguration> config, wxWebViewWindowInfoWebKit* parentWindowInfo = nullptr);
```
vadz, line 36:
```suggestion
    explicit wxWebViewWebKit(std::shared_ptr<wxWebViewConfiguration> config, wxWebViewWindowInfoWebKit* parentWindowInfo = nullptr);
```

include/wx/osx/webview_webkit.h
```diff
@@ -135,15 +120,22 @@ class WXDLLIMPEXP_WEBVIEW wxWebViewWebKit : public wxWebView
...
                               wxWindowID id,
                               const wxString& url = wxWebViewDefaultURLStr,
                               const wxPoint& pos = wxDefaultPosition,
                               const wxSize& size = wxDefaultSize,
                               long style = 0,
                               const wxString& name = wxASCII_STR(wxWebViewNameStr)) override
-    { return new wxWebViewWebKit(parent, id, url, pos, size, style, name); }
+    {
+        auto webView = Create();
+        if (webView->Create(parent, id, url, pos, size, style, name))
```
vadz, line 134: I find it very confusing to see `Create()` called twice, typically this is an error in wx API (you're supposed to either use a non-default ctor or a default ctor and `Create()`). I wonder if there is still any way of renaming something to avoid calling both functions `Create()` or if it's too late?

src/common/webview.cpp
```diff
@@ -458,4 +478,15 @@ void wxWebView::InitFactoryMap()
...
 
+class wxWebViewConfigurationDefault : public wxWebViewConfiguration
+{
+public:
+    wxWebViewConfigurationDefault(): wxWebViewConfiguration(wxWebViewBackendDefault) {}
+};
+
+std::shared_ptr<wxWebViewConfiguration> wxWebViewFactory::CreateConfiguration()
+{
+    return std::shared_ptr<wxWebViewConfiguration>(new wxWebViewConfigurationDefault());
```
vadz, line 489: Should use `make_shared`.

vadz (commented):
My main question here is whether we really need to use `shared_ptr`. As it is now, it seems like we could just copy/pass by value to keep things simple. And if the configuration objects become too big to be copied all the time in the future (will they?) I'd prefer using `shared_ptr` in them internally to avoid copying rather than enforcing their use via shared pointers, which is rather confusing because it doesn't seem really obvious that we need shared ownership here.

If we do keep using shared pointers, I have a couple of nitpicks and would prefer to pass them by const reference instead of by value to the various functions and would also use `{}` instead of `nullptr` for an empty pointer. But, again, I'd very much prefer to not use them at all.

Would this be possible?

TcT2k:
> My main question here is whether we really need to use `shared_ptr`. As it is now, it seems like we could just copy/pass by value to keep things simple. And if the configuration objects become too big to be copied all the time in the future (will they?) I'd prefer using `shared_ptr` in them internally to avoid copying rather than enforcing their use via shared pointers, which is rather confusing because it doesn't seem really obvious that we need shared ownership here.
> 
> If we do keep using shared pointers, I have a couple of nitpicks and would prefer to pass them by const reference instead of by value to the various functions and would also use `{}` instead of `nullptr` for an empty pointer. But, again, I'd very much prefer to not use them at all.
> 
> Would this be possible?

Should be possible, just didn't think about doing it that way. Will certainly make some things easier.

include/wx/osx/webview_webkit.h
TcT2k: I agree it's confusing but the factory method is called `Create()` but i could of course use new here instead of calling `Create()` from the factory itself.

include/wx/webview.h
```diff
@@ -144,6 +144,27 @@ class WXDLLIMPEXP_WEBVIEW wxWebViewHandler
...
 };
 
+class WXDLLIMPEXP_WEBVIEW wxWebViewConfigurationImpl
+{
+public:
+    virtual ~wxWebViewConfigurationImpl() = default;
+    virtual void* GetNativeConfiguration() const { return nullptr; }
+};
+
+class WXDLLIMPEXP_WEBVIEW wxWebViewConfiguration
```
vadz, line 154: As written, this class shouldn't be DLL-exported as it only has inline functions. This would change, however, if `wxWebViewConfigurationImpl` becomes private as `GetNativeConfiguration()` couldn't be implemented inline then.

include/wx/webview.h
```diff
@@ -144,6 +144,27 @@ class WXDLLIMPEXP_WEBVIEW wxWebViewHandler
...
+{
+public:
+    explicit wxWebViewConfiguration(const wxString& backend, wxWebViewConfigurationImpl* impl)
+        : m_backend(backend), m_impl(impl) {}
+    virtual ~wxWebViewConfiguration() = default;
```
vadz, line 159: This class is not meant to be inherited from and so its dtor should _not_ be virtual.

vadz (commented):
Thanks a lot for the changes, this seems much cleaner to me now!

I'd just like to move the `Impl` class into private header if there is no reason to keep it public that I'm missing and then it would be perfect.

TcT2k:
I've moved `wxWebViewConfigurationImpl` to a private header and added some small additions.

vadz:
Thanks again! Will merge in a moment.

PR #23358, opened 2023-03-17
Fix wxPropertyGrid tab traversal issue under wxGTK

### LINKED ISSUE #23354: wxPropertyGrid tab traversal issue under wxGTK
### Description
Summary: Tab traversal from wxPropertyGrid doesn't work across wxPanel instances. It also fails when the property grid is followed by a static control such as wxStaticText or wxStaticLine.

### Detailed Description
Ordinarily a user may navigate between controls, even across different panels, by pressing the tab key. However, adding a property grid can sometimes break the navigation chain such that pressing the tab key won't advance keyboard focus past the property grid. The problem is that to get around peculiarities in GTK's handling of keyboard focus the property grid is programmed to find its next sibling and set focus to it directly, which is fine when the grid is followed by a focusable control but otherwise leads to incorrect behavior.

The file propgrid.cpp (inside src/propgrid) includes the following code (wxWidgets 3.2.2.1):

```
// Tab traversal workaround for platforms on which
// wxWindow::Navigate() may navigate into first child
// instead of next sibling. Does not work perfectly
// in every scenario (for instance, when property grid
// is either first or last control).
#if defined(__WXGTK__)
    wxWindow* sibling = mainControl->GetNextSibling();
    if ( sibling )
        sibling->SetFocusFromKbd();
#else
    Navigate(wxNavigationKeyEvent::IsForward);
#endif
```

This code has two problems. First, the code doesn't check whether the sibling accepts focus or not. So if the property grid is followed by a wxStaticText or wxStaticLine control, for example, it will focus on that instead of the next focusable control. Second, navigation is confined to the current parent; focus may transfer between siblings, but it will never leave the parent control.

The first of these issues can be fixed by iterating through the siblings until either a sibling is found that accepts keyboard focus or the list is exhausted, but that doesn't help with the second issue. What I have found in testing possible fixes is that calling Navigate() on the property grid with no rows selected will work as desired, but calling Navigate() after a row has been selected will either transfer focus to the edit control (which is desirable) or transfer focus from the edit control back to the property grid (which is not desirable, hence the need for a special case for wxGTK in the code above).

Calling SetFocus() on the property grid followed by Navigate() allows focus to proceed as expected:

```
// Tab traversal workaround for platforms on which
// wxWindow::Navigate() may navigate into first child
// instead of next sibling. Does not work perfectly
// in every scenario (for instance, when property grid
// is either first or last control).
#if defined(__WXGTK__)
    SetFocus();
    Navigate(wxNavigationKeyEvent::IsForward);
#else
    Navigate(wxNavigationKeyEvent::IsForward);
#endif
```

... or, more simply:

```
// Tab traversal workaround for platforms on which
// wxWindow::Navigate() may navigate into first child
// instead of next sibling. Does not work perfectly
// in every scenario (for instance, when property grid
// is either first or last control).
#if defined(__WXGTK__)
    SetFocus();
#endif

Navigate(wxNavigationKeyEvent::IsForward);
```

With the code above wxGTK appears to behave as expected, but I haven't tested whether this has any undesired side effects so I'm submitting this as an issue rather than a pull request (which would basically consist of the second or third snippet above together with its counterpart for backwards navigation). I can confirm that the propgrid sample that comes with wxWidgets still works as expected after this change, but I otherwise don't know enough about wxWidgets to know if this would cause any new issues.

### Test program
Patch against samples/propgrid/propgrid_minimal.cpp: 
[propgrid_minimal.cpp.txt](https://github.com/wxWidgets/wxWidgets/files/10998538/propgrid_minimal.cpp.txt)

- wxWidgets version 3.2.2.1.
- wxGTK.
- Linux Mint 20
- GTK 3.24.20-0ubuntu1.1.

adrianlopezroche:
Fix wxPropertyGrid tab traversal issue under wxGTK (tab traversal from a wxPropertyGrid doesn't work across wxPanel instances or when the property grid is followed by a static control such as wxStaticText).

Fixes wxWidgets/wxWidgets#23354.

### DIFF
--- a/src/propgrid/propgrid.cpp
+++ b/src/propgrid/propgrid.cpp
@@ -5609,15 +5609,6 @@ void wxPropertyGrid::HandleKeyEvent( wxKeyEvent &event, bool fromChild )
 
     if ( keycode == WXK_TAB )
     {
-#if defined(__WXGTK__)
-        wxWindow* mainControl;
-
-        if ( HasInternalFlag(wxPG_FL_IN_MANAGER) )
-            mainControl = GetParent();
-        else
-            mainControl = this;
-#endif
-
         if ( !event.ShiftDown() )
         {
             if ( !editorFocused && m_wndEditor )
@@ -5628,16 +5619,12 @@ void wxPropertyGrid::HandleKeyEvent( wxKeyEvent &event, bool fromChild )
             {
                 // Tab traversal workaround for platforms on which
                 // wxWindow::Navigate() may navigate into first child
-                // instead of next sibling. Does not work perfectly
-                // in every scenario (for instance, when property grid
-                // is either first or last control).
+                // instead of next sibling.
             #if defined(__WXGTK__)
-                wxWindow* sibling = mainControl->GetNextSibling();
-                if ( sibling )
-                    sibling->SetFocusFromKbd();
-            #else
-                Navigate(wxNavigationKeyEvent::IsForward);
+                SetFocus();
             #endif
+
+                Navigate(wxNavigationKeyEvent::IsForward);
             }
         }
         else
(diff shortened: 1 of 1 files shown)


vadz:
Thanks! Will merge in a moment.

PR #23360, opened 2023-03-18
wxURI minor update with tests

ghost:
Sync with my local copy, except for Normalization (my local copy is in almost pure C, so it doesn't port over cleanly):

- Update IPVX parsing to RFC, along with a lot of tests
- Authority parsing was reworked for RFC edge cases
- Couple other extreme edge cases

All previous tests were kept, just added all of them from my local copy, which is a few.  It's been a hot minute since I've contributed anything but I wanted to sync it with my local copy as best I could, while following the guidelines. If there's any requests I'll be happy to change it if feasible.

Note: All tests pass except for two tests where wxExec launches Notepad and fails to soft close it due to EnumWindows not getting a window. I tried to rework it heavily to using EnumThreadWindows (and related APIs like Process/Thread32Next), which did get a proper window, but notepad ignored the WM_QUIT message regardless. Either way this isn't due to my changes, just a note.


vadz:
Thanks for your changes! I admit I didn't check all the changes to the parsing code in the details, but it looks globally good and, of course, the fact that all the tests pass gives me a lot of confidence in it. If nobody else has any objections, I'm going to merge this in master soon (I'm not sure if we should backport this to 3.2, while the new behaviour is almost certainly more correct, there are still probably some changes compared to the existing code).

As an aside, it's unfortunate that we hadn't got rid of CppUnit compatibility macros in this test file until now, as using Catch matchers would have been much better than dealing with all these macros, but now that it's done it's not worth rewriting them. But if you plan to make any future non-trivial changes to the tests, it would be better to start by converting it to use Catch directly as many things become simpler then.

Anyhow, thanks again for your work on this!

MetalPlateMage:
Put on the finishing touch of fixing github issue #3350  (done unintentionally as a result of this update).  I just changed the Resolve() assert to take into account legacy file urls/uris so that #3350 goes through clean in debug mode as well. Commit message:

"Fix github issue https://github.com/wxWidgets/wxWidgets/issues/3350 - it was already fixed, but the assert was bad for old code in Resolve(), so I changed it for compatibility purposes
.
Put in a test for https://github.com/wxWidgets/wxWidgets/issues/3350 and the other 4 compat IPv6 tests."

Whether you want to backport is up to you.

PR #23362, opened 2023-03-19
Don't expose internal strings to translation in samples

Blake-Madden:


vadz:
Thanks, I'll merge this soon.

PR #23366, opened 2023-03-21
Document wxGLCanvas::CreateSurface()

dgud:
It is needed to be able to re-create the EGL drawing surface after the window layout have changed, such as after a reparent of the canvas or of it's grandparents.

The functionality is already there, so this is mostly a documentation to keep the functionality available in future versions.

Also destroy the surface before re-creating the surface, so the user don't have to do that.

Would really like if this could be back-ported to 3.2 branch as seen in the since tag.

### DIFF
--- a/interface/wx/glcanvas.h
+++ b/interface/wx/glcanvas.h
@@ -887,6 +887,19 @@ public:
                const wxString& name = "GLCanvas",
                const wxPalette& palette = wxNullPalette);
 
+
+    /**
+       Re-creates EGLSurface. To be used after a reparent or other
+       changes that may invalidate the EGL drawing surface.
+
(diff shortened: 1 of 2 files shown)


dgud:
I don't think the failing build is due to my changes?

vadz:
> I don't think the failing build is due to my changes?

No, this is #23367 and is just due to Fedora updating their Webkit version.

I'll merge this one soon, thanks.

dgud:
Will/can this be merged to 3.2?

I need to know if I should ifdef the erlang wrapper api with 3.3 or 3.2.3

vadz:
Yes, I plan to merge it into 3.2 (see the label), I'd just like to let it spend at least some time in master to see if anybody runs into any problems with it (which seems unlikely, but who knows).

Issue #23364, opened 2023-03-20
Possible GSource leak in wxGTK, especially when using a lot of timers


eevee:
Apologies if this is a little half-baked — C++ isn't my native language, I don't know much about GTK, and the project I'm trying to fix isn't even mine.  :)  But _something_ funny is going on here and at the very least I could use some insight from someone who knows what they're doing.

The project is [SLADE](https://github.com/sirjuddington/SLADE), a Doom editor (specifically [issue 1016](https://github.com/sirjuddington/SLADE/issues/1016)).  A few years ago it started having a problem where, once the map editor had been opened, it would consume more and more CPU and become less and less responsive until it was completely unusable.  

Subsequent on-and-off investigation revealed the following:

- The cause seems to be an accumulation of GSources, to the point that the glib event loop cannot iterate over them fast enough.  This ends up visible in `perf top` as `g_source_ref` eating most of the CPU time (called via `g_main_loop_run` → `g_main_context_iterate` → `g_main_context_check` → `g_source_iter_next`).

- Armed with a debug copy of glib, I found that the extraneous GSources are all timeouts.  SLADE's map editor has a wxTimer that goes off every 15ms to trigger its GL canvas to redraw (by calling `Refresh`), and _sometimes_ the associated GSource gets incref'd one too many times and sits in the GMainContext, in a dead state, forever.

    The timer is a oneshot that's restarted from its own event handler.  I don't know why that's the case, but the obvious next step is:

- I tried making SLADE's timer continuous, and that partially fixed the problem since it would only ever create one GSource...  but then I would _still_ accumulate timeout GSources, this time from GTK itself — they're named `[gtk+] gdk_frame_clock_paint_idle` and [can be seen being set here](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/gdkframeclockidle.c#L331).

This doesn't make any sense — timeouts are handled entirely internally by glib, and nothing outside has any reason to alter their refcount at all (save for decrementing it once).  And yet, they persist.

As far as I can tell, what's actually happening is this:

1. `wxGUIEventLoop::DoRun` calls `gtk_main`, which in turn calls `g_main_loop_run`.

2. `g_main_loop_run` calls `g_main_context_iterate`, which assembles a list of GSources that are ready to be dispatched.  Because the list holds copies of pointers to the GSources, their refcounts are all incremented.

[...]

PR #23368, opened 2023-03-21
Fix GSource leak from idle processing

paulcor:
It is apparently not safe to call gtk_events_pending() from the idle callback, as it can result in GLib failing to properly dereference some GSource objects. This leads to increasingly large amounts of CPU time being spent in GLib processing internal lists of these GSource objects. So avoid calling gtk_events_pending(), as it is not strictly necessary because our idle source will remain installed if wxIdleEvent::RequestMore() is used, and if no events have arrived the idle callback will be invoked again. See #23364

### DIFF
--- a/src/gtk/app.cpp
+++ b/src/gtk/app.cpp
@@ -142,12 +142,9 @@ bool wxApp::DoIdle()
         gs_focusChange = 0;
     }
 
-    bool needMore;
-    do {
-        ProcessPendingEvents();
+    ProcessPendingEvents();
+    const bool needMore = ProcessIdle();
 
-        needMore = ProcessIdle();
-    } while (needMore && gtk_events_pending() == 0);
     gdk_threads_leave();
 
 #if wxUSE_THREADS


eevee:
SLADE (the downstream project I was originally trying to fix) has a whole lot of compilation errors against 3.3 right now, but if I cherry-pick this onto 3.2, it does seem to solve the problem!

Unfortunately it also seems to cause SLADE to spend another 5% CPU just processing idle events, according to `perf top`.  I don't understand why, since it only occasionally uses `RequestMore`, so this might be a little more complicated.

But that side effect might also be something specifically quirky about SLADE, or my use of `LD_PRELOAD`.  Little bit out of my comfort zone here, sorry  :)

vadz:
> has a whole lot of compilation errors against 3.3 right now

Hmm, this is rather unexpected, there were very few backwards-incompatible changes in 3.3 compared to 3.2 so far. If you have any concrete examples of things that you'd expect to work (i.e. they're not explicitly mentioned in [this section of the changelog](https://github.com/wxWidgets/wxWidgets/blob/23ccdb23c84b6b3de74d1cd83c83d7ad01431725/docs/changes.txt#L8)), please open separate issue(s) for them.

> it does seem to solve the problem!

Thanks for confirming this! We'll merge the fix then.

> Unfortunately it also seems to cause SLADE to spend another 5% CPU just processing idle events

Generally speaking, using `wxIdleEvent::RequestMore()` is a very inefficient way of managing CPU resources. It's probably not very simple to change it, but using a background thread for any long running tasks is a much better way since the last -- let's try to be conservative -- 25 years or so.

paulcor:
Actually, I think the overhead is more like 15us. But it doesn't really matter, as continuing idle processing when there are higher priority events waiting is just incorrect. The user code may depend on those other events being handled before it gets another idle.

eevee:
> Hmm, this is rather unexpected, there were very few backwards-incompatible changes in 3.3 compared to 3.2 so far. If you have any concrete examples of things that you'd expect to work (i.e. they're not explicitly mentioned in [this section of the changelog](https://github.com/wxWidgets/wxWidgets/blob/23ccdb23c84b6b3de74d1cd83c83d7ad01431725/docs/changes.txt#L8)), please open separate issue(s) for them.

The specific things I ran into were:

- `wxBitmap::set{Width,Height}`, deprecated since 3.0
- changing wxString to std::string caused an overload ambiguity between `wxArrayString` and `std::vector<char*>` which didn't exist before
- some link errors, mostly around `wxString::{Printf,Format}`, that I now suspect were build contamination somehow (stale precompiled headers??), but that's around where I gave up and tried 3.2

So mostly not unexpected, I think.

> Generally speaking, using `wxIdleEvent::RequestMore()` is a very inefficient way of managing CPU resources. It's probably not very simple to change it, but using a background thread for any long running tasks is a much better way since the last -- let's try to be conservative -- 25 years or so.

I note that SLADE itself never uses `RequestMore()` — it only comes from Scintilla, via wxStyledTextCtrl.  :)

----

I'm also wary of the timeout solution; I really wouldn't expect returning to the glib event loop to have much overhead, and I can't explain why it seems to on SLADE.  That said, SLADE already tends to idle at around 20%, so it might be doing something dubious that's getting exacerbated for incidental reasons.

And I'm now getting memory corruption with a fresh build of a fresh checkout of SLADE that didn't happen two days ago, so it's possible the project is just haunted.  :)

Is there another complex project this patch could be benchmarked against, to see if the overhead is consistent?

vadz:
> The specific things I ran into were:
> 
>     * `wxBitmap::set{Width,Height}`, deprecated since 3.0

This one is indeed not unexpected. FWIW `Set{Width,Height}()` never really worked, so any code using them really needs to be fixed.

>     * changing wxString to std::string caused an overload ambiguity between `wxArrayString` and `std::vector<char*>` which didn't exist before

Not sure about this one, we didn't change anything by default. Could you please give more details about this, probably in a new issue? I.e. some minimal example of code which compiled before but doesn't any longer.

>     * some link errors, mostly around `wxString::{Printf,Format}`, that I now suspect were build contamination somehow (stale precompiled headers??), but that's around where I gave up and tried 3.2

Yes, this does look like a bad build problem, there definitely shouldn't be any problems with those.

> I note that SLADE itself never uses `RequestMore()` — it only comes from Scintilla, via wxStyledTextCtrl. :)

I see. Looking at Scintilla sources, this seems to be difficult to change, unfortunately.

> Is there another complex project this patch could be benchmarked against, to see if the overhead is consistent?

I guess any other project using wxSTC with line wrapping would do.

vadz:
I've merged this fix now, thanks again.

We'll also need to backport it to 3.2, but let's wait a bit to see if any problems with it are found.

Issue #23369, opened 2023-03-22
master (0331057aba) doesn't build with --enable-compat30


dgud:
../configure --enable-debug --enable-compat30

### Description

Build fails:

wxWidgets/MYPR/bk-deps g++ -c -o propgriddll_propgridpagestate.o  -  CUT_INCLUDES  ../src/propgrid/propgridpagestate.cpp
../src/propgrid/propgridiface.cpp: In member function ‘void wxPropertyGridInterface::SetPropertyColoursToDefault(wxPGPropArg)’:
../src/propgrid/propgridiface.cpp:704:73: error: no matching function for call to ‘wxPropertyGridInterface::SetPropertyColoursToDefault(const wxPGPropArgCls&, wxPGPropertyValuesFlags)’
  704 |     SetPropertyColoursToDefault(id, wxPGPropertyValuesFlags::DontRecurse);
      |                                                                         ^
../src/propgrid/propgridiface.cpp:702:6: note: candidate: ‘void wxPropertyGridInterface::SetPropertyColoursToDefault(wxPGPropArg)’
  702 | void wxPropertyGridInterface::SetPropertyColoursToDefault(wxPGPropArg id)
      |      ^~~~~~~~~~~~~~~~~~~~~~~
../src/propgrid/propgridiface.cpp:702:6: note:   candidate expects 1 argument, 2 provided
In file included from ../include/wx/propgrid/propgrid.h:23,
                 from ../src/propgrid/propgridiface.cpp:26:

### Platform and version information

- wxWidgets version master commit 0331057aba
- wxWidgets port wxGTK
- OS Ubuntu-20.04

vadz:
This seems to work for me and I don't understand why it wouldn't, there _is_ an overload taking 2 arguments in master.

You put `CUT_INCLUDES`, but could you be including system-installed older wx headers somehow?

dgud:
Having a bad day can't build 3.2 branch either.

Why is this looking for:  /home/uabdgud/src/wxWidgets/MYPR/lib/wx/include/gtk3-unicode-**3.3**
When I have wx/include/gtk3-unicode-**3.2**

How do I build the latest on 3.2 branch, am I using the wrong branch? 


>git fetch upstream
>git reset --hard upstream/3.2
>rm -rf * && ../configure --enable-debug
>make --debug
GNU Make 4.2.1
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2016 Free Software Foundation, Inc.
[...]

PR #23371, opened 2023-03-22
Fix problems with 3.0 compatibility in wxPG code

vadz:
See #23369.

Also enable 3.0 compatibility for one of the CI builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
           - name: Ubuntu 18.04 wxGTK 3 STL
             runner: ubuntu-latest
             container: ubuntu:18.04
-            configure_flags: --enable-stl --disable-compat30
+            configure_flags: --enable-stl --enable-compat30
             use_xvfb: true
           - name: Ubuntu 20.04 wxGTK 3 with clang
             runner: ubuntu-20.04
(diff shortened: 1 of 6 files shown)


vadz:
@a-wi I wanted to make minimal changes initially, but this was ugly (code had to be duplicated) and didn't allow the sample, using the new API, to compile when compatibility was enabled, so I finally went ahead and did a bigger, but IMO also much better, change in the last commit.

Generally speaking, we really should only add something to the API when compatibility with old version is enabled, but still provide all the new APIs. It's unfortunate that we can't do this for `DrawCaptionSelectionRect()`, the new virtual function really ought to have had a different name to allow it, but we still should do it like this for any new deprecations.

vadz:
@a-wi Do you see anything you'd like to apply from this PR other than the last commit after your 35a8d0f9085f9c9b5aad6fde56bda66124132506 or should it be just closed?

a-wi:
Essentials changes in wxPG headers are already done but adding a additional condition for `wxPG_DEPRECATED_MACRO_VALUE` in property.h would be useful.
Updates to CI are welcome.

vadz:
Hmm, looking at the amount of the existing code overriding this function and seeing that this change is not even mentioned in the change log, I think we do need to keep this working.

vadz:
Replaced by #23375.

Issue #23367, opened 2023-03-21
Fix deprecation warnings in `wxWebViewWebKit` code


vadz:
We've started getting the following warnings in Fedora CI builds:
```
./src/gtk/webview_webkit2.cpp: In function 'gboolean wxgtk_webview_webkit_navigation(WebKitWebView*, WebKitPolicyDecision*, wxWebViewWebKit*)':
./src/gtk/webview_webkit2.cpp:123:71: error: 'const gchar* webkit_navigation_policy_decision_get_frame_name(WebKitNavigationPolicyDecision*)' is deprecated: Use 'webkit_navigation_policy_decision_get_navigation_action' instead [-Werror=deprecated-declarations]
  123 |     wxString target = webkit_navigation_policy_decision_get_frame_name(navigation_decision);
      |                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/webkitgtk-4.0/webkit2/webkit2.h:61,
                 from ./include/wx/gtk/private/webkit.h:15,
                 from ./src/gtk/webview_webkit2.cpp:34:
/usr/include/webkitgtk-4.0/webkit/WebKitNavigationPolicyDecision.h:70:1: note: declared here
   70 | webkit_navigation_policy_decision_get_frame_name        (WebKitNavigationPolicyDecision *decision);
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./src/gtk/webview_webkit2.cpp: In function 'gboolean wxgtk_webview_webkit_new_window(WebKitPolicyDecision*, wxWebViewWebKit*)':
./src/gtk/webview_webkit2.cpp:291:71: error: 'const gchar* webkit_navigation_policy_decision_get_frame_name(WebKitNavigationPolicyDecision*)' is deprecated: Use 'webkit_navigation_policy_decision_get_navigation_action' instead [-Werror=deprecated-declarations]
  291 |     wxString target = webkit_navigation_policy_decision_get_frame_name(navigation_decision);
      |                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
/usr/include/webkitgtk-4.0/webkit/WebKitNavigationPolicyDecision.h:70:1: note: declared here
[...]

PR #23374, opened 2023-03-23
WebKit2GTK: Add wxWebViewConfiguration and child window support

TcT2k:
Add some features to the WebKit2GTK backend that where recently added to Edge and macOS

wxWebViewConfiguration
==

Add support for more in depth configuration before creating a web view. Allows to customize the data path and access to the underlying context.

Add wxWebViewWindowFeatures and event
==

This replaces the previously implemented `wxWebViewWindowInfo`.

It explicitly **breaks** the previous API to enable WebKitGTK
integration and to make the usage in application code less error prone.

A new event `wxEVT_WEBVIEW_NEWWINDOW_FEATURES` is added to allow
access to the window features and the child web view.

New sample handling code looks like this:
```C++
// Bind new window handler
m_webView->Bind(wxEVT_WEBVIEW_NEWWINDOW, [](wxWebViewEvent& evt) {
    if (evt.GetURL() == "http://badwebsite.com")
        evt.Veto(); // Disallow new window for badwebsite.com
    else
        evt.Skip(); // Allow new window for all other websites
});
 
// Bind new window features handler
m_webView->Bind(wxEVT_WEBVIEW_NEWWINDOW_FEATURES, [](wxWebViewEvent& evt) {
    // Get target window features
    wxWebViewWindowFeatures* features = evt.GetTargetWindowFeatures();
    // Create a top level window for the child web view
    wxWindow* win = new wxWindow(this, wxID_ANY, features->GetPosition(), features->GetSize());
    wxSizer* sizer = new wxBoxSizer(wxVERTICAL);
    win->SetSizer(sizer);
    // Get the child web view
    wxWebView* childWebView = features->GetChildWebView();
    // Place the child web view into the window
    childWebView->Create(win, wxID_ANY);
    sizer->Add(childWebView, 1, wxEXPAND);
}
```

### DIFF
--- a/include/wx/gtk/webview_webkit.h
+++ b/include/wx/gtk/webview_webkit.h
@@ -34,15 +34,9 @@ class WXDLLIMPEXP_WEBVIEW wxWebViewWebKit : public wxWebView
 public:
     wxWebViewWebKit();
 
-    wxWebViewWebKit(wxWindow *parent,
-           wxWindowID id = wxID_ANY,
-           const wxString& url = wxWebViewDefaultURLStr,
-           const wxPoint& pos = wxDefaultPosition,
-           const wxSize& size = wxDefaultSize, long style = 0,
-           const wxString& name = wxASCII_STR(wxWebViewNameStr))
-    {
-        Create(parent, id, url, pos, size, style, name);
-    }
+#if wxUSE_WEBVIEW_WEBKIT2
+    wxWebViewWebKit(const wxWebViewConfiguration& config);
+#endif
 
     virtual bool Create(wxWindow *parent,
            wxWindowID id = wxID_ANY,
@@ -200,6 +194,7 @@ private:
     //Used for webkit2 extension
     GDBusServer *m_dbusServer;
     GDBusProxy *m_extension;
+    wxWebViewConfiguration m_config;
 #endif
 
     wxDECLARE_DYNAMIC_CLASS(wxWebViewWebKit);
@@ -216,9 +211,18 @@ public:
                               const wxSize& size = wxDefaultSize,
                               long style = 0,
                               const wxString& name = wxASCII_STR(wxWebViewNameStr)) override
-    { return new wxWebViewWebKit(parent, id, url, pos, size, style, name); }
+    {
+        std::unique_ptr<wxWebView> webView(new wxWebViewWebKit);
+        if (webView->Create(parent, id, url, pos, size, style, name))
+            return webView.release();
+        else
+            return nullptr;
+    }
+
 #if wxUSE_WEBVIEW_WEBKIT2
     virtual wxVersionInfo GetVersionInfo() override;
+    virtual wxWebViewConfiguration CreateConfiguration() override;
+    virtual wxWebView* CreateWithConfig(const wxWebViewConfiguration& config) override;
 #endif
 };
 
--- a/include/wx/gtk/webview_webkit.h
+++ b/include/wx/gtk/webview_webkit.h
@@ -35,6 +35,7 @@ public:
     wxWebViewWebKit();
 
 #if wxUSE_WEBVIEW_WEBKIT2
+    wxWebViewWebKit(WebKitWebView* parentWebView, wxWebViewWebKit* parentWebViewCtrl);
     wxWebViewWebKit(const wxWebViewConfiguration& config);
 #endif
 
--- a/include/wx/webview.h
+++ b/include/wx/webview.h
@@ -344,10 +344,14 @@ private:
     wxDECLARE_ABSTRACT_CLASS(wxWebView);
 };
 
-class WXDLLIMPEXP_WEBVIEW wxWebViewWindowInfo
+class WXDLLIMPEXP_WEBVIEW wxWebViewWindowFeatures
 {
 public:
-    virtual ~wxWebViewWindowInfo() = default;
+    wxWebViewWindowFeatures(wxWebView* childWebView);
+
+    virtual ~wxWebViewWindowFeatures();
+
+    wxWebView* GetChildWebView() const;
 
     virtual wxPoint GetPosition() const = 0;
 
@@ -361,9 +365,10 @@ public:
 
     virtual bool ShouldDisplayScrollBars() const = 0;
 
-    virtual wxWebView* CreateChildWebView() = 0;
+protected:
+    mutable bool m_childWebViewWasUsed;
+    std::unique_ptr<wxWebView> m_childWebView;
 };
-
 
 class WXDLLIMPEXP_WEBVIEW wxWebViewEvent : public wxNotifyEvent
 {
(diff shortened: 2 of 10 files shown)


include/wx/gtk/webview_webkit.h
```diff
@@ -216,9 +212,18 @@ class WXDLLIMPEXP_WEBVIEW wxWebViewFactoryWebKit : public wxWebViewFactory
                               const wxSize& size = wxDefaultSize,
                               long style = 0,
                               const wxString& name = wxASCII_STR(wxWebViewNameStr)) override
-    { return new wxWebViewWebKit(parent, id, url, pos, size, style, name); }
+    { 
+        wxWebView* webView = new wxWebViewWebKit;
+        if (webView->Create(parent, id, url, pos, size, style, name))
+            return webView;
+        else
+            return nullptr;
```
vadz, line 220: Shouldn't we `delete webView` here (or, rather, use `std::unique_ptr<>` for it in the first place)?

src/common/webview.cpp
```diff
@@ -86,6 +87,24 @@ wxString wxWebViewConfiguration::GetDataPath() const
...
+// wxWebViewWindowFeatures
+wxWebViewWindowFeatures::wxWebViewWindowFeatures(wxWebView * childWebView):
+    m_childWebViewWasUsed(false),
+    m_childWebView(childWebView)
+{ }
+
+wxWebViewWindowFeatures::~wxWebViewWindowFeatures()
+{
+    if (m_childWebView && !m_childWebViewWasUsed)
+        delete m_childWebView;
```
vadz, line 99: I wonder if we could avoid this by making `GetChildWebView()` reset the pointer to `nullptr` after returning it? In fact, what about using `unique_ptr<>` here too and returning it by value from `GetChildWebView()` to make it absolutely clear that the caller takes its ownership?

src/gtk/webview_webkit2.cpp
```diff
@@ -608,6 +682,80 @@ wxgtk_authorize_authenticated_peer_cb(GDBusAuthObserver *,
...
 
+//-----------------------------------------------------------------------------
+// wxWebViewConfigurationImplWebKit
+//-----------------------------------------------------------------------------
+class wxWebViewConfigurationImplWebKit : public wxWebViewConfigurationImpl
+{
+public:
+    wxWebViewConfigurationImplWebKit():
+        m_webContext(nullptr)
+#if WEBKIT_CHECK_VERSION(2, 10, 0)
```
vadz, line 693: Should we perhaps define some `HAVE_WEBKIT_WEBSITE_DATA_MANAGER` once instead of checking for the version multiple times? This would be a bit more readable, IMHO, and would make it easier to get rid of this check later when we drop support for the versions < 2.10.

src/gtk/webview_webkit2.cpp
```diff
@@ -608,6 +682,80 @@ wxgtk_authorize_authenticated_peer_cb(GDBusAuthObserver *,
...
+        m_websiteDataManager(nullptr)
+#endif
+    { }
+
+#if WEBKIT_CHECK_VERSION(2, 10, 0)
+    wxString GetDataPath() const override
+    {
+        GetOrCreateContext();
+        if (m_websiteDataManager)
+            return webkit_website_data_manager_get_base_data_directory(m_websiteDataManager);
```
vadz, line 704: We probably need a run-time version check for 2.10 here (and when creating it below), otherwise an app compiled on a newer system will crash when running on an older one and I think we want to support this.

As I've just recently written in [this comment](https://github.com/wxWidgets/wxWidgets/issues/23367#issuecomment-1479377777), I think it would be useful to add `wx_check_webkit_version()` function for doing it, as WebKit itself doesn't seem to provide one.

vadz (commented):
Thanks for your work on this!

The new API looks perfectly reasonable to me and it's great that it now works on all platforms.

I just have some minor comments about (not) deleting raw pointers, I think it would be nice to start following good C++ practices in wx code (better late than never, right) and use `unique_ptr` instead whenever possible.

Thanks again!

include/wx/gtk/webview_webkit.h
TcT2k: Changed to use `std::unique_ptr`

src/gtk/webview_webkit2.cpp
```diff
@@ -73,6 +74,79 @@ bool wxGetStringFromJSResult(WebKitJavascriptResult* js_result, wxString* output
...
+        webkit_window_properties_get_geometry(m_properties, &m_geometry);
+        // Treat 0 as -1 to indicate that the value is not set
+        if (m_geometry.width == 0)
+            m_geometry.width = -1;
+        if (m_geometry.height == 0)
+            m_geometry.height = -1;
+        if (m_geometry.x == 0)
+            m_geometry.x = -1;
+        if (m_geometry.y == 0)
+            m_geometry.y = -1;
```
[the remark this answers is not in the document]
TcT2k: While valid I think it's still the right thing to do. As in most cases it won't be set and it's more in line with the other backends (where we can actually check if a value is available).

src/common/webview.cpp
TcT2k: Using a `std::unique_ptr` would probably make sense. But I intentionally wanted `GetChildWebView()` to be callable multiple times by application code. E.g. it could make sense to do something with the webview before processing the rest of `wxWebViewWindowFeatures` in another method.

As this is a control the application code shouldn't have to worry to much about ownership as that is more or less automatic once it's added to a window.

include/wx/webview.h
```diff
@@ -361,10 +365,11 @@ class WXDLLIMPEXP_WEBVIEW wxWebViewWindowInfo
 
     virtual bool ShouldDisplayScrollBars() const = 0;
 
-    virtual wxWebView* CreateChildWebView() = 0;
+protected:
+    mutable bool m_childWebViewWasUsed; // modified from const method
```
vadz, line 369: Sorry for continuing to nitpick, but `GetChildWebView()` is not actually const, which is why I was wondering why it was mutable in the first place. Did you intend to make it const, or is there some other method modifying this variable somewhere?

TcT2k: I see, I indeed intented it to be const. Will change that.

src/gtk/webview_webkit2.cpp
vadz: Yes, it's called "lazy binding" and is the default under Linux (and the vast majority of the other Unix systems although historically there were some exceptions, I believe). This is also the default behaviour of `dlopen()` (`RTLD_LAZY` is the default, but `RTLD_NOW` can be used to force immediately resolving everything).

src/gtk/webview_webkit2.cpp
```diff
@@ -38,6 +39,20 @@
...
+
+// Function to check webkit version at runtime
+bool wx_check_webkit_version(int major, int minor, int micro)
+{
+    const unsigned int version = webkit_get_major_version() * 10000 +
+                                 webkit_get_minor_version() * 100 +
+                                 webkit_get_micro_version();
+    const unsigned int required = major * 10000 + minor * 100 + micro;
+    return version >= required;
+}
```
swt2c, line 54: Can `wx_check_webkit_version` please be moved into a separate commit, and perhaps be merged separately ASAP?  I could use it for the other webkit2 changes.

TcT2k: I've added it as a separate commit. But this PR can probably be merged soon anyway and would include this change.

swt2c: Thanks - that will make it easier to backport - since all of the changes in this PR will probably not be backportable.

vadz:
Thanks, merged now and, indeed, this won't be backported entirely but we can backport 42ca043.

PR #23375, opened 2023-03-23
Improve and test 3.0 compatibility in wxPG

vadz:
This replaces #23371.

### DIFF
--- a/src/propgrid/property.cpp
+++ b/src/propgrid/property.cpp
@@ -98,6 +98,31 @@ void wxPGCellRenderer::DrawEditorValue( wxDC& dc, const wxRect& rect,
     }
 }
 
+#if WXWIN_COMPATIBILITY_3_0
+
+// This is ugly, but we need to know whether our version of this function was
+// called or not, so we set this global variable to true before calling it and
+// then reset it to false if the base class version was used.
+//
+// To make things even worse, we also use this variable to check if we're
+// called from our own code (in which case we just need to reset it) or from
+// the application (in which case we actually need to draw something).
+namespace
+{
+bool wxPGCellRendererDrawCaptionSelectionRectFlag = false;
+}
+
+void wxPGCellRenderer::DrawCaptionSelectionRect( wxDC& dc,
+                                       int x, int y,
+                                       int w, int h ) const
+{
+    if ( wxPGCellRendererDrawCaptionSelectionRectFlag )
+        wxPGCellRendererDrawCaptionSelectionRectFlag = false;
+    else
+        wxPGDrawFocusRect(nullptr, dc, x, y, w, h);
+}
+#endif // WXWIN_COMPATIBILITY_3_0
+
 void wxPGCellRenderer::DrawCaptionSelectionRect(wxWindow* win, wxDC& dc,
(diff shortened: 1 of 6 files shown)


src/propgrid/property.cpp
```diff
@@ -297,13 +322,34 @@ bool wxPGDefaultRenderer::Render( wxDC& dc, const wxRect& rect,
...
+            }
+            else
+                // This means that our own version was called, so we now need
+                // to call the other overload, to use the user-defined version
+                // if any or to just draw the rectangle in the base class
+                // version otherwise.
+#endif // WXWIN_COMPATIBILITY_3_0
+            DrawCaptionSelectionRect( const_cast<wxPropertyGrid*>(propertyGrid), dc,
+                                      rectCaption.x, rectCaption.y,
+                                      rectCaption.width, rectCaption.height );
```
a-wi, line 352: To me the intent of this this logic is not clear but if you want to do something like this, the better place would be probably  `wxPGDrawFocusRect`, where you know what `DrawCaptionSelectionRect` calls it (win ==/ != nullptr) and where you already have conditional blocks for v3.0 compatibility code.

vadz: The intent here is to call the overridden function with the old signature as this is what's really needed for compatibility sand this wasn't done any longer after the latest changes (before them the old function was called but the new one was not, which wasn't ideal neither).

This couldn't be done solely in `wxPGDrawFocusRect()`.

a-wi: OK, but I still don't get why do you insist on using `DrawCaptionSelectionRect()` with old signature in wx code in v3.0 mode.
In this mode `DrawCaptionSelectionRect()` is defined in two versions: with old and new signature, so why not to use only new one internally? I think version with old signature should be only used (potentially) by user app (in v3.0 mode).

vadz: It's a virtual function, so I assume that it can be overridden in the user code, so we need to call it. If not, why was it virtual in the first place?

a-wi: OK, good point. But maybe simpler implementation would be just to call in v3.0 mode `DrawCaptionSelectionRect()` with old signature from `DrawCaptionSelectionRect()` with new signature? I mean something like this:
```
DrawCaptionSelectionRect(wxWindow*, wxDC& dc,...)
{
#if v3.0
DrawCaptionSelectionRect(dc, ...)
#else
wxPGDrawFocusRect(...)
#endif
}
```

vadz: This means enabling 3.0 compatibility (which might be needed for something completely unrelated) would make wxPG draw uglier looking focus rect by default, which doesn't seem good. I.e. the goal is to always call `wxPGDrawFocusRect()` with a valid window by default, but still call overridden virtual function with 3.0-compatible signature if 3.0 compatibility is enabled.

And I still don't see any simpler/better way to do it than the hack in this PR.

vadz: @a-wi Do you still object to merging this in the current state and/or see a better way to handle this?

vadz: I'm going to push this, we can always change it later if you find a better way to do it.

PR #23390, opened 2023-03-28
CMake: Add libraries of imported targets to wx-config

### LINKED ISSUE #23373: [nanosvg] When NanoSVG is used as a non-header-only library, it's missed from the wx-config file
Ref: https://github.com/microsoft/vcpkg/issues/30361

MaartenBent:
Use the `LOCATION_DEBUG` or `LOCATION_RELEASE` property of an imported target (like NanoSVG) to determine the full library path.

Fixes #23373


vadz:
Thanks, I don't think it's worth waiting longer before merging it, so I'll do it in a moment.

MaartenBent:
Can you backport this one to 3.2 too?

PR #23392, opened 2023-03-28
Fix AppVeyor httpbin

swt2c:


swt2c:
Looks like it worked.  Fixes #23392.

PR #23398, opened 2023-03-29
Try go-httpbin

MaartenBent:
Just a test to see if this will work. When python httpbin fails again, this could be an alternative.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,6 +29,8 @@ skip_commits:
     - build/tools/httpbin.sh
 
 environment:
+  GOVERSION: "1.20.3"
+  GOPATH: c:\gopath
   matrix:
   - TOOLSET: msbuild
     CONFIGURATION: DLL Release
@@ -83,15 +85,21 @@ before_build:
 build_script: c:\projects\wxwidgets\build\tools\appveyor.bat
 
 before_test:
+- cmd: |
+    echo Getting and launching httpbin.
+    rmdir %GOROOT% /s /q
+    mkdir %GOROOT%
+    appveyor DownloadFile https://golang.org/dl/go%GOVERSION%.windows-amd64.msi
+    msiexec /i go%GOVERSION%.windows-amd64.msi INSTALLDIR="%GOROOT%" /q
+    go version
+    go install github.com/mccutchen/go-httpbin/v2/cmd/go-httpbin@v2
+    set PATH=%PATH%;%GOPATH%\bin
 - ps: |
-    Write-Output "Getting and launching httpbin."
-    $env:PATH = "C:\Python38;C:\Python38\Scripts;" + $env:PATH
-    pip.exe --disable-pip-version-check install Flask==2.0.3 Werkzeug==2.0.3 httpbin
-    Start-Job -Name wx_httpbin { python.exe -m httpbin.core 2>&1 > c:\projects\wxwidgets\httpbin.log }
+    Start-Job -Name wx_httpbin { go-httpbin -host 127.0.0.1 -port 8081 2>&1 > c:\projects\wxwidgets\httpbin.log }
     Start-Sleep -Seconds 5
-    curl.exe --silent --show-error http://127.0.0.1:5000/ip > $null
+    curl.exe --silent --show-error http://127.0.0.1:8081/ip > $null
     if ($lastExitCode -eq "0") {
-        $env:WX_TEST_WEBREQUEST_URL="http://127.0.0.1:5000"
+        $env:WX_TEST_WEBREQUEST_URL="http://127.0.0.1:8081"
     }
     else {
         Write-Error "Disabling wxWebRequest tests as launching httpbin failed, log follows:"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,7 +29,7 @@ skip_commits:
     - build/tools/httpbin.sh
 
 environment:
-  GOVERSION: "1.20.3"
+  GOVERSION: "1.22.1"
   GOPATH: c:\gopath
   matrix:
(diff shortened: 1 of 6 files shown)


tierra:
I know @vadz did attempt to use this a long time ago once. I don't recall what issues he ran into with it though.

cbb25f66ac239a1f032230a91c89327e61900b55

MaartenBent:
Yes, I remembered that too, and found the below text in the mailing list. Installing works, the only remaining problem I'm looking at now is that it doesn't seem to run in AppVeyor. CURL can't connect / connection refused.

Conclusions so far:
- running go-httpbin works in all GitHub CI tests
- can't run go-httpbin in old appveyor containers, it needs at least go 1.16(?). so install it manually
- go-httpbin (and https://httpbingo.org/ ) responses are slightly different, causing the tests tofail

---

> Also I really, really hate trying to use python/pip in a portable way, this has taken way too much time. So I tried to switch to a version not requiring Python from https://github.com/mccutchen/go-httpbin but I didn't have better luck with Go -- although here it could be due to my total lack of knowledge of it. The "go install" command given in the README didn't
work, and from the documentation it looks like "go get" should be used instead, but it still doesn't work for some reason, even after setting GO{ROOT,PATH} -- it seems to look for the files only locally instead of downloading them for some reason. If anybody knows how to install this under macOS 10.11, please let me know.

MaartenBent:
success

vadz:
If it works, perhaps we want to switch to it already, without waiting for the next problem?

I haven't compared, but what takes longer: installing Python or Go version?

MaartenBent:
I'll reopen this since it still seems to fail with python: https://ci.appveyor.com/project/wxWidgets/wxwidgets/builds/46818435/job/u8vmba8bh6ey7dmo#L3120

tests/net/webrequest.cpp
```diff
@@ -263,7 +263,7 @@ TEST_CASE_METHOD(RequestFixture,
     pos += strlen(expectedKey);
 
     // There may, or not, be a space after it.
-    while ( wxIsspace(response[pos]) )
+    while ( wxIsspace(response[pos]) || response[pos] == '[' )
```
vadz, line 266: Could you please update the comment to explain why do we need to test for `[` here? I don't really understand it...

appveyor.yml
```diff
@@ -86,15 +88,21 @@ before_build:
 build_script: c:\projects\wxwidgets\build\tools\appveyor.bat
 
 before_test:
+- cmd: |
+    echo Getting and launching httpbin.
+    rmdir %GOROOT% /s /q
+    mkdir %GOROOT%
+    appveyor DownloadFile https://golang.org/dl/go%GOVERSION%.windows-amd64.msi
```
vadz, line 95: Just to be clear: we can't use any of [Go versions](https://www.appveyor.com/docs/windows-images-software/#golang) already available on Appveyor?

vadz (commented):
Thanks again for working on this, I'm more than willing to try using this, but have a couple of minor questions before merging it.

tests/net/webrequest.cpp
```diff
@@ -186,11 +186,11 @@ TEST_CASE_METHOD(RequestFixture,
     if ( !InitBaseURL() )
         return;
 
-    Create("/bytes/65536");
+    Create("/bytes/1000");
```
MaartenBent, line 189: I changed this to 1000, because go-httpbin doesn't add a `content-length` header when the size is over 2048 bytes. Resulting in the functions below to return `-1`.

vadz:
It looks like something [got broken again](https://github.com/wxWidgets/wxWidgets/actions/runs/8252986967/job/22573798475) in Python ecosystem. Can we try switching to this one?

MaartenBent:
Sure. I think I addressed all the comments, though not marked them as resolved yet.

I can rebase to check if it still works. But if you want to play with it, I don't mind if you push to this PR.

vadz:
If you could please rebase it, please do, otherwise I will, but I won't be able to do it today and maybe not tomorrow neither, unfortunately, and in the meanwhile Windows CMake CI seems to be permanently broken (i.e. the error above is not just a fluke, I've already tried rerunning the build and it failed again). TIA!

vadz:
Thanks a lot! I've (squash) merged this now, let's see if it works better in the long run.

MaartenBent:
It does still have some warnings in the action logs: `Restore cache failed: Dependencies file is not found in /__w/wxWidgets/wxWidgets. Supported file pattern: go.sum`

I think we should add this to all the `setup-go`:
```yml
with:
  cache: false
```

vadz:
Thanks, done now.

There is also a lot of output like this when running the tests:

```
time="2024-03-13T13:18:02.9718" status=200 method="POST" uri="/post" size_bytes=604 duration_ms=0.08 user_agent="test/1 wxWidgets/3.3.0" client_ip=127.0.0.1:34276
```

Can they be avoided/redirected and shown only in case of error?

MaartenBent:
I think something goes wrong with piping the output in https://github.com/wxWidgets/wxWidgets/blob/b03c3416c87e799cdd086be49761913e994451bd/build/tools/httpbin.sh#L13

I think `go-httpbin -host 127.0.0.1 -port 8081 &> httpbin.log &` should fix it (from https://askubuntu.com/a/731237 )
When there is an error, `httpbin.log` contents are already shown.

vadz:
We'd need to switch to using `bash` as shell in order to use `&>` which is a non-POSIX construct AFAIR and `2>&1 >` should really work in the same way (as it means "redirect stderr to stdout and then redirect stdout to file") unless Go does something really weird. But then it just might, of course...

PR #23402, opened 2023-03-30
Updated wxwin.m4 macro regex to detect x.x.x[...] versions

mehw:
Hello,

Is it possible to add the possibility to detect wxWidgets versions like `3.2.2.1`?

There are problems compiling [xCHM](https://github.com/rzvncj/xCHM/pull/27) on Gentoo GNU/Linux with `wxGTK-3.2.2.1` installed: https://bugs.gentoo.org/895982

The current `wxwin.m4` macro fails to detect `x.x.x[...]` versions:
```
echo "3.2.2.1" | sed 's/\([0-9]*\).\([0-9]*\).\([0-9]*\)/\1/'
```
This PR simply adds a capture group to glob the `[...]` part in `x.x.x[...]`:
```
echo "3.2.2.1" | sed -r 's/^([0-9]+)\.([0-9]+)\.([0-9]+)(.*)$/\1/'
```
Thanks.

### DIFF
--- a/wxwin.m4
+++ b/wxwin.m4
@@ -222,11 +222,11 @@ AC_DEFUN([WX_CONFIG_CHECK],
 
     WX_VERSION=`$WX_CONFIG_WITH_ARGS --version 2>/dev/null`
     wx_config_major_version=`echo $WX_VERSION | \
-           sed 's/\([[0-9]]*\).\([[0-9]]*\).\([[0-9]]*\)/\1/'`
+           sed -r 's/^([[0-9]]+)\.([[0-9]]+)\.([[0-9]]+)(.*)$/\1/'`
     wx_config_minor_version=`echo $WX_VERSION | \
-           sed 's/\([[0-9]]*\).\([[0-9]]*\).\([[0-9]]*\)/\2/'`
+           sed -r 's/^([[0-9]]+)\.([[0-9]]+)\.([[0-9]]+)(.*)$/\2/'`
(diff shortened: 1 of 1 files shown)


vadz:
This seems to be the same as #23289. As mentioned there, it's due to some change to wx-config in Gentoo, but I still don't know why did they change it nor where exactly.

But I still suggest submitting your fix to Gentoo developers because it doesn't make sense for wx itself, because our wx-config is never going to output the 4th version component, but they should be willing to take it (if they're not willing to simply revert this change). They also should have no problem with using `sed -r`, while we can't do this in wx itself because it is not portable and, for example, wouldn't work at all under macOS.

mehw:
Thank you for the precise answer.  I completely missed the https://github.com/wxWidgets/wxWidgets/pull/23289 PR.

Gentoo's devs fixed the problem just minutes ago:
https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=86d51f8fb503db6f6f023efdb1ebca4590404fa7

Simply put, the fix drops the 4th version component from the release string, so it's never printed by the installed `wx-config`.

I just installed the Gentoo's `wxGTK-3.2.2.1` fixed version and [xCHM](https://github.com/rzvncj/xCHM/pull/27) configures and compiles fine now.

I'm closing the PR.  Thanks ;)

rzvncj:
Great! Thanks for researching it. I'm glad it's all sorted out.

PR #23403, opened 2023-03-30
Fedora CI: enable GUI tests

swt2c:
This runs the GUI tests during Fedora CI using Wayland/Weston compositor.  Unfortunately the tests cannot be run (reliably) with Mutter as there are issues intermittently where the main window loses focus and stops getting painted.

Unfortunately, there's an issue when running some of tests inside the Docker container because there are some namespaces that are unavailable, so I had to use a somewhat hacky workaround (provide a bwrap wrapper) to workaround this limitation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,6 +64,12 @@ jobs:
             WX_TEST_WEBREQUEST_URL="0" ./test
 
       - run:
+          name: Run GUI tests using Xvfb
+          command: |
+            cd tests
+            xvfb-run -a -s '-screen 0 1600x1200x24' ./test_gui -d 1 $wx_tests_selection
+
+      - run:
           name: Show ccache statistics
           when: always
           command: ccache -vv -s
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,23 @@ jobs:
             WX_TEST_WEBREQUEST_URL="0" ./test
 
       - run:
+          name: Run GUI tests
+          command: |
+            cat > /usr/local/bin/bwrap \<<'EOF'
+            #!/bin/sh
+            # glycin's loader sandbox needs user namespaces unavailable in this container;
+            # run glycin image loaders unsandboxed, pass everything else to the real bwrap.
+            for a in "$@"; do case "$a" in /usr/libexec/glycin-loaders/*)
+              while [ $# -gt 0 ]; do case "$1" in /usr/libexec/glycin-loaders/*) exec "$@";; esac; shift; done;;
+            esac; done
+            exec /usr/bin/bwrap "$@"
+            EOF
+            chmod +x /usr/local/bin/bwrap
+
+            cd tests
+            wlheadless-run -- ./test_gui
+
+      - run:
           name: Show ccache statistics
           when: always
           command: ccache -vv -s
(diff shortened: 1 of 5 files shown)


vadz:
@swt2c Do you plan to return to this?

swt2c:
Maybe eventually.  I think the issue I ran into was that when I went to debug this locally, I ran into a bunch more test failures locally.  I'm not really sure why. Perhaps HiDPI related or Wayland related?

In any event, feel free to close this PR if it's been open too long without resolution.  😁

vadz:
We can keep it open, it doesn't really cost us anything, I just wasn't sure if this wasn't completely abandoned.

FWIW the tests using `wxUIActionSimulator` only run under X11, not Wayland, so you either need to run them with `GDK_BACKEND=x11` (to run the tests) or `WX_UI_TESTS=0` (to skip them). You also need to run the tests from the `tests` subdirectory (which is annoying and I keep meaning to fix this and postponing it...). But otherwise they are supposed to work, and do work for me.

swt2c:
@vadz finally got back around to this (as you like to say, better late than never ;-)) after getting (most) of the tests working locally with Wayland on Fedora.

vadz:
Excellent, thanks! This is definitely very good to have, I'll merge it soon.

vadz:
Oops, the tests didn't pass: https://app.circleci.com/pipelines/gh/wxWidgets/wxWidgets/6886/workflows/6eb65f5b-4631-467a-8506-70d42c2b9e03/jobs/6774/parallel-runs/0/steps/0-110

If you can't access the URL, the error is:

```
WebView
  Selection
./controls/webtest.cpp:224

./controls/webtest.cpp:242: FAILED:
  CHECK( m_browser->HasSelection() )
with expansion:
  false

./controls/webtest.cpp:243: FAILED:
  CHECK( m_browser->GetSelectedText() == "Some strong text" )
with expansion:
  "" == "Some strong text"

./controls/webtest.cpp:252: FAILED:
  REQUIRE( selSource.Lower().Matches("*some*<strong*strong</strong>*text*") )
with expansion:
  false
with message:
  Unexpected selection source: ""
```

I have no idea why does it happen, so I'm probably going to just disable this test under Wayland for now unless you tell me how to fix it.

FWIW there is also

```
wxPersistTLW
./persistence/tlw.cpp:84

./waitfor.h:32: warning:
  Timed out waiting for frame to be iconized
```

and we may want to check for a WM (or launch one?) to allow it to be iconized.

And there is this at the end:

```
[16:14:09.360] BUG: finalizing a layer with views still on it.
```

which seems suspicious but I don't know which views are those (is it possible to ask Weston for a screen dump before exiting?).

swt2c:
Seems like your change in a87d72e hopefully resolved the WebView test issue.

vadz:
Weird that it should take more than 50ms under Wayland, but as long as it works, I'm not going to bother looking into this (sometimes never might be better than later :-).

swt2c:
Might be more related to running in the Fedora container than Wayland.  WebKit2 has that separate process where the extension runs, so it's possible that launching that separate process takes more than 50ms...

swt2c:
I looked into the other issues.  I don't think it's possible to get the iconized state on Wayland, so I think the best we could do is skip that check on Wayland also.

About the `[16:14:09.360] BUG: finalizing a layer with views still on it.` it seems that is likely just due to the way Weston is shutting down abruptly after the test process exits, probably not something to worry about.

swt2c:
#26994 for the iconized check.

Issue #23365, opened 2023-03-20
wxKill wxSIGTERM unit tests fail in Windows 11 in Head


MetalPlateMage:
I mentioned this figured I'd document it (feel free to remove if not important).

Results:
```
tests\exec\exec.cpp(202): FAILED:
  CHECK( rc == 0 )
with expansion:
  -1 == 0
exec.cpp(275): FAILED:
  CHECK( rc == 0 )
with expansion:
  -1 == 0
```

It has to do with how Notepad and some modern presentation apps work. EnumWindows fails completely to find a window, going through the process threads finds you a window with EnumThreadWindows (patch attached, I wouldn't apply it its just for show as what I hacked around with on Friday), but then Notepad - or the launching process - ignores WM_QUIT.

This goes into the deeper issue where when you launch a process on windows with modern presentation apps it only launches a hooking process. I'm not sure how to get the right process to kill, so to speak.

EDIT: Also, forgot to mention the wxSIGKILL test doesn't actually close Notepad either.. it remains open despite that test reporting success

MetalPlateMage:
Also, forgot to mention the wxSIGKILL test doesn't actually Notepad either.. it remains open despite that test reporting success

vadz:
This is pretty important and thanks for reporting it, but I'm not sure what to do about it. Do you have more information about the "modern presentation apps" way by chance? There must be _some_ way to terminate the process we launch ourselves, but perhaps we need to switch to using process handles instead of PIDs in order to do it?

MetalPlateMage:
OK, so it's been a while since I've researched how "optimized" WPF apps work... specifically how Microsoft uses them. I mean my main hobby is working on a pet app for Mac system 6+7, which means I can't even use templates most of the time (though Retro68 is a thing now). Anyways....

So, when you launch notepad for the first time, it generally (and I say generally because it is murky and dependent on some algorithm is uses) creates 3 other notepad processes for what I believe is caching - sort of like how 4 socket servers serve a lot of clients these few processes service all notepad windows. The problem is these processes have no valid parent processes because they are launched by something else which doesn't exist anymore, so we can't enumerate them normally. I recall a way to do it in .Net from like 10 years ago, but I never used .Net/WPF in any of the apps I did so I dumped all that knowledge.

The other case (which is easier to handle) is that it spawns a single child process which is the gui process. However, the process launched by wxExec has no message loop so PostMessage won't do anything and killing that process will not kill notepad. Here's what it looks like:
[...]

PR #23408, opened 2023-03-31
Tweak wxExec to send WM_QUIT and change test app to mspaint

ghost:
Change wxKill to PostMessage(WM_QUIT) to each thread of a process, the same as Windows does when it tries to kill a process. If we want to do more of what windows does we'd do a WM_CLOSE and WM_ENDSESSION to each topwindow of a process, but it might be just excessive...

Change the test application to use for tests from Notepad to MSPaint, which at least as the moment doesn't do a process pool or anything so it quits cleanly and I'm not left with Notepads open every time I run the tests.

This technically "resolves" #23365 (https://github.com/wxWidgets/wxWidgets/issues/23365).


vadz:
Thanks!

Just to confirm: does the current code fail to kill mspaint too? I.e. do I understand correctly that just changing the program used in the test is not enough, and that we also need to change `wxKill()` implementation to make it work with mspaint, but that even this change to `wxKill()` is still not enough to make it work with notepad because of its child processes?

MetalPlateMage:
- Windows 11 notepad uses process pooling. Currently the only way I can think of without using WPF/.NET is to hook CreateProcess() when you spawn the process - I'm sure there are other ways, but so far all of them are beyond what wxKill() is stated to do.

- Just switching to mspaint with the current code passes the tests - just tested now (didn't originally since I implemented this first in my quest to kill notepad).

- This way works even if the process doesn't have a top level window created like the parent notepad process... usually. It doesn't kill the ones it spawns that show the GUI though as I mentioned. I can understand not wanting to change stable code.
 
Hope those answer your questions Vadim.

P.S. Also sorry if there were any weird pull requests before this; I have very little usage with git - I've used svn for over 20 years.

vadz:
Thanks for the answers!

I indeed prefer not to change the existing code if it has no immediate benefits. Perhaps we can replace it with something more reliable, that works with notepad too, later, but for now I'll apply just the change to the test to make it pass under Windows 11 but not the rest of them.

PR #23422, opened 2023-04-05
Use C++17 `std::{to,from}_chars()`

vadz:
`strtod()` or `snprintf()` are still faster because they don't need to construct `wxString` at all and so don't allocate, but the difference is smaller now and `FromCDouble()` is now 3+ times faster than `FromDouble()` instead of 1.5 times slower.

### DIFF
--- a/include/wx/string.h
+++ b/include/wx/string.h
@@ -4254,10 +4254,20 @@ void wxStringIteratorNode::DoSet(const wxString *str,
     m_str = str;
     if ( str )
     {
+        // When compiling with optimizations gcc considers "this" pointers used
+        // below dangling, even if this is really not the case.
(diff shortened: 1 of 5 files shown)


vadz:
Some benchmark results for relative speed of `FromCDouble()` and its non-C counterpart:

1. Originally: `FromCDouble()` 103% i.e. slightly slower, as expected, as it does slightly more.
2. With this PR when using C++17: `FromCDouble()` is 25% i.e. 4 times faster.
3. With this PR when not using C++17: `FromCDouble()` is almost exactly the same as `FromDouble()` as there is no more overhead of `wxUILocale`.

vadz:
I'm going to merge this soon if there are no comments/objections.

Issue #23419, opened 2023-04-04
Locale/language info minor bugs


BrianNixon:
Kashmiri (India) has a blank native description (though I don’t know what it should be).

The following are tagged as LTR; I suspect (though am far from certain) they should be RTL:
* Mazanderani
* Northern Luri
* Uzbek (Perso-Arabic)

There are some native descriptions which include mixed-direction text without direction markers:
* Northern Luri (Iran)
* Northern Luri (Iraq)

In some right-to-left layouts, these are rendered wrong because the opening bracket is interpreted in the prevailing right-to-left direction before the direction switches to left-to-right when the English text is encountered. Either ‘Iran’ and ’Iraq’ need translating, or there needs to be left-to-right direction marker for the opening bracket.

That said: my browser gets it right here:
* لۊری شومالی (Iran)
* لۊری شومالی (Iraq)

vadz:
The table in `languageinfo.cpp` is generated from official Unicode consortium data, so I'm not sure why would it be wrong...

cc @utelle this one as well.

BrianNixon:
Could wx be out of date, or have a buggy import?
https://github.com/unicode-org/cldr/blob/release-42/common/main/mzn.xml#L592

utelle:
> The table in `languageinfo.cpp` is generated from official Unicode consortium data, [...]

Actually, `languageinfo.cpp` is generated from `misc/languages/langtabl.txt` (and `misc/languages/scripttabl.txt`) with the Python script `misc/languages/genlang.py`. However, the information in `misc/languages/langtabl.txt` is generated based on data from various sources, among them the [Unicode CLDR](https://cldr.unicode.org/), _Windows 10 locale data_, _ISO 639-3_, and _ISO 3166_.

Currently, the information about the **locale name** (English and native) and the **layout direction** are taken from the _Windows 10 locale data_ (as retrieved via the system functions [`EnumSystemLocalesEx()`](https://learn.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-enumsystemlocalesex) and [`GetLocaleInfoEx()`](https://learn.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-getlocaleinfoex). 

Unfortunately, the Windows system functions return the **layout direction** as **LTR** for the locales

- Mazanderani (_mzn_)
- Northern Luri (_lrc_)
- Uzbek (Perso-Arabic) (_uz-Arab_)

[...]

PR #23426, opened 2023-04-07
Update locale list and fix issue #23419

utelle:
The list of known languages/locales is updated to match the list of known locales in Windows 11. Additionally, the bugs mentioned in issue #23419 are fixed, except the native descriptions of Northern Luri (Luri translations of the territories are not known).


vadz:
Great, thanks!

Would it be possible to add the (sources of the) program you used to regenerate these files somewhere to the repository too to allow someone else doing this later?

Also, in principle, this should be backported to 3.2, but the new constants would have to be inside `wxABI_VERSION` checks there and I'm not really sure it's worth adding them... Opinions?

utelle:
> Would it be possible to add the (sources of the) program you used to regenerate these files somewhere to the repository too to allow someone else doing this later?

In principle, yes. However, at the moment the regeneration process consists of quite a number of steps with various tools on a Windows platform:

1. Generate a textfile with a list of locales supported by Windows 11. This is done with a small wxWidgets based program.
2. Retrieve data files from various sources (mainly from the Unicode CLDR repository), for example, ISO codes for languages and countries. This is curently done manually.
3. Convert the text based data files from step 1 and 2 to SQL. This is done by a number of Lua scripts.
4. Import the SQL generated in step 3 into a SQLite database. This is done with a SQLite shell.
5. Generate the files `langtabl.txt` and `scripttabl.txt`. This is done with a SQLite shell.
6. Update the wxWidgets source files with the Python script `genlang.py` using the files generated in step 5.

Steps 3, 4, and 5 are called from a Windows batch file. Especially in step 5 one can get errors - mainly due to name clashes with generated symbols (`wxLANGUAGE_*`). These have to be resolved manually by adjusting a Lua and/or SQL script (and sometimes the Python script `genlang.py` if new synonyms have to be created), before step 6 can be executed.

The whole process is quite complex, because we need to keep compatibility with prior wxWidgets versions.

> Also, in principle, this should be backported to 3.2, but the new constants would have to be inside `wxABI_VERSION` checks there and I'm not really sure it's worth adding them... Opinions?

Well, IMHO I would not backport these changes to 3.2 - at least not in full. The bugs discovered in #23419 are all minor and affect only rather exotic locales.

Nevertheless, I could provide an adjusted file `langtabl.txt` that does not add any new constants, but only fixes bugs like wrong layout direction. This file could then be used to regenerate the dependent source files for 3.2 with the Python script `genlang.py` of 3.2.

vadz:
I think it would be nice to add the sources of (1) to the repository, as at least this can't (realistically) be done manually. As for the other steps, maybe just documenting them in (the existing) README would be enough? TIA!

Concerning 3.2, I'm fine with not backporting anything to it, but if anybody is motivated to make a PR applying some fixes to the files there, I'm also fine with merging it.

utelle:
> I think it would be nice to add the sources of (1) to the repository, as at least this can't (realistically) be done manually. As for the other steps, maybe just documenting them in (the existing) README would be enough? TIA!

I will try to streamline the process. Actually, I currently use the process for more than just generating `langtabl.txt`. For some of my applications I need translations of language, country, and currency names, and I generate .po files from data gathered from Unicode CLDR data.

I agree that the process of regenerating `langtabl.txt` should be at least documented within the wxWidgets repository. I will do that, but it will take some time. However, that shouldn't be a problem, now that the files are updated.

> Concerning 3.2, I'm fine with not backporting anything to it, but if anybody is motivated to make a PR applying some fixes to the files there, I'm also fine with merging it.

I suppose _anybody_ will mean _me_. I will look into the issue, but I won't promise anything yet.

vadz:
> > I think it would be nice to add the sources of (1) to the repository, as at least this can't (realistically) be done manually. As for the other steps, maybe just documenting them in (the existing) README would be enough? TIA!
> 
> I will try to streamline the process.

TIA!

> I agree that the process of regenerating `langtabl.txt` should be at least documented within the wxWidgets repository. I will do that, but it will take some time. However, that shouldn't be a problem, now that the files are updated.

Sure, I'll still merge this, of course.
 
> > Concerning 3.2, I'm fine with not backporting anything to it, but if anybody is motivated to make a PR applying some fixes to the files there, I'm also fine with merging it.
> 
> I suppose _anybody_ will mean _me_.

Not necessarily! It might be you or @BrianNixon who has reported the original problem or anybody actually using these languages (small chance of this happening, I think, but who knows...).

utelle:
> > > Concerning 3.2, I'm fine with not backporting anything to it, but if anybody is motivated to make a PR applying some fixes to the files there, I'm also fine with merging it.
> > 
> > I suppose _anybody_ will mean _me_.
> 
> Not necessarily! It might be you or @BrianNixon who has reported the original problem or anybody actually using these languages (small chance of this happening, I think, but who knows...).

In the meantime I already updated the file `langtabl.txt` for 3.2 locally. I assume that changes should be applied to the branch `3.2`. My wxWidgets fork does not contain that branch, so I have to find out how to get that branch into my repository.

utelle:
I created PR #23430 to backport the bug fixes without adding new language constants.

PR #23427, opened 2023-04-07
Update wxToolBar documentation

PBfordev:
Remove information relevant only for Pocket PC and also fix a mistyped (German?) word.

Few more things I noticed but did not do anything about:
1. Is the information about color mapping on MSW still true/relevant?
2. `AddControl()` has remark "wxMac: labels are only displayed if wxWidgets is built with wxMAC_USE_NATIVE_TOOLBAR set to 1 " but I could not find that define anywhere in the codebase.

### DIFF
--- a/interface/wx/toolbar.h
+++ b/interface/wx/toolbar.h
@@ -171,10 +171,8 @@ public:
     the menu bar in a wxFrame.
 
     You may create a toolbar that is managed by a frame calling
-    wxFrame::CreateToolBar(). Under Pocket PC, you should always use this
-    function for creating the toolbar to be managed by the frame, so that
-    wxWidgets can use a combined menubar and toolbar. Where you manage your
-    own toolbars, create wxToolBar as usual.
+    wxFrame::CreateToolBar(). Where you manage your own toolbars,
+    create wxToolBar as usual.
 
     There are several different types of tools you can add to a toolbar.
     These types are controlled by the ::wxItemKind enumeration.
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for the update!

To answer the other questions:

1. This is definitely out of date, as no remapping is done by default since a very long time, see fb7e87effa (cleaned up and disabled by default the toolbar buttons colour remapping code, 2006-02-25). I'm almost sure that we can just remove all this code completely as it only makes sense for 8bpp colour modes or pre-XP systems neither of which is supported any longer, so IMO we could just remove all this documentation *and* all its code, but the latter would require at least some testing to ensure nothing gets broken. FWIW [code search](https://github.com/search?type=code&auto_enroll=true&q=%22msw.remap%22+-path%3Asrc%2Fmsw%2Ftoolbar.cpp&l=C%2B%2B) shows quite a few occurrences of setting `msw.remap = 0`, which actually enables remapping (`Remap_Bg`). In any case, the docs definitely need to be changed.
2. `wxMAC_USE_NATIVE_TOOLBAR` got renamed to `wxOSX_USE_NATIVE_TOOLBAR` back in 292e5e1f17 (OSX renaming round 2, 2008-07-30), so the very least we need to do is to rename it here as well. But I think we can just remove this remark, actually, because `wxOSX_USE_NATIVE_TOOLBAR` should be always set to 1. Or maybe warn that this doesn't work correctly if this option is changed from its default value (assuming this is still true, I didn't really bother to check).

Please let me know if you plan to do any of the proposed changes here (not necessarily immediately/urgently, of course) or if I should merge this as is. Thanks!

PBfordev:
Thanks for the quick reply.

As for (1), I think we (i.e., you) need to decide what to do with the option and code using it first. 

As for (2), I have never even seen macOS, so I cannot comment on that.

So, I do not plan any further changes to the PR.

vadz:
OK, thanks, I've done the required minimum for (2) and opened another issue for (1) and will merge this one now.

PR #23432, opened 2023-04-08
Fix deleting taskbar icon in wxMSW wxNotificationMessage

vadz:
Previously, the taskbar icon could remain shown in the taskbar after wxNotificationMessage was destroyed, as wxEVT_TASKBAR_BALLOON_TIMEOUT handler was called on an already dead object.

Weirdly enough this didn't result in the crashes, but it definitely didn't work correctly neither.

Fix this by binding a separate handler for this event, not using the wxNotificationMessage object at all, which is responsible for destroying the icon when it times out.

This also seems to make it completely unnecessary to have a separate m_active flag in wxNotificationMessageImpl, as we can now just delete the "impl" object too directly when the main object itself is deleted.

---

@TcT2k Please let me know if I'm missing something here, but to me the hack with `m_active` just seems unnecessary -- and it is definitely insufficient, as the commit message above explains. With this change I don't get stray taskbar icons in my own applications any longer after testing this for a couple of days.

### DIFF
--- a/include/wx/private/notifmsg.h
+++ b/include/wx/private/notifmsg.h
@@ -14,8 +14,7 @@ class wxNotificationMessageImpl
 {
 public:
     wxNotificationMessageImpl(wxNotificationMessageBase* notification):
-        m_notification(notification),
-        m_active(false)
+        m_notification(notification)
     {
 
     }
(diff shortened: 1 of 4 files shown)


TcT2k:
> @TcT2k Please let me know if I'm missing something here, but to me the hack with `m_active` just seems unnecessary -- and it is definitely insufficient, as the commit message above explains. With this change I don't get stray taskbar icons in my own applications any longer after testing this for a couple of days.

Too long ago that I looked at that code to say anything with confidence, but your assumptions seem right to me from you observations and code changes.

vadz:
Thanks for looking at this, merged now.

PR #23433, opened 2023-04-08
[DO NOT MERGE] Get cursor from bundle resoure part2

stahta01:


MaartenBent:
You probably have to do something similar to what the samples do, so the resource compiler gets the correct include paths:
https://github.com/wxWidgets/wxWidgets/blob/d0e892ac09cee8721980e9b56b88e21c9679435f/build/bakefiles/common_samples_intree.bkl#L28-L31

stahta01:
> You probably have to do something similar to what the samples do, so the resource compiler gets the correct include paths:
> 
> https://github.com/wxWidgets/wxWidgets/blob/d0e892ac09cee8721980e9b56b88e21c9679435f/build/bakefiles/common_samples_intree.bkl#L28-L31

Thank you

oneeyeman1:
@vadz,
Please close this.

Thx.

PR #23436, opened 2023-04-09
Remove most occurrences of wxT() from the docs

PBfordev:
I have removed `wxT()` in most cases, leaving it only in:
1. Where a string literal is used in `wxChar*` arrays in `wxPropertyGrid` overview (e.g., [here](https://github.com/wxWidgets/wxWidgets/blob/master/docs/doxygen/overviews/propgrid.h#L272)).
2. In `wxIMAGE_OPTION_*` defines.

BTW, I think the current description of `wxT()` may be a bit outdated:
https://github.com/wxWidgets/wxWidgets/blob/d0e892ac09cee8721980e9b56b88e21c9679435f/interface/wx/chartype.h#L17-L23

wxWidgets removed non-UNICODE build option in 3.3 and `wxChar` is always defined as `wchar_t` (i.e., regardless of the underlying `wxString` char type, so even in the UTF-8 build), therefore `wxT()` now always adds `L` (unlike `wxS()` which adds it only if `wxUSE_UNICODE_WCHAR`):
https://github.com/wxWidgets/wxWidgets/blob/d0e892ac09cee8721980e9b56b88e21c9679435f/include/wx/chartype.h#L168-L178

BTW2, is it not high time to deprecate or even remove `wxT_2()` macro, its code comment says "wxT_2 exists only for compatibility with wx 2.x"?

### DIFF
--- a/docs/doxygen/overviews/mbconvclasses.h
+++ b/docs/doxygen/overviews/mbconvclasses.h
@@ -99,7 +99,7 @@ wxString::FromUTF8().
 Example 3: Input in KOI8-R. Construction of wxCSConv instance on the fly.
 
 @code
-wxString str(input_data, wxCSConv(wxT("koi8-r")));
+wxString str(input_data, wxCSConv("koi8-r"));
 @endcode
 
 Example 4: Printing a wxString to stdout in UTF-8 encoding.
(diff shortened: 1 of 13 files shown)


vadz:
Thanks, I've updated `wxT()` documentation and pushed it here, please let me know if you have any comments. As for `wxT_2()`, it seems to be used in [a few projects](https://github.com/search?type=code&auto_enroll=true&q=wxT_2+-path%3Awx%2Fchartype.h&l=C%2B%2B), so we shouldn't remove it and so I've left its documentation as well, but made it even more clear that it is obsolete now.

PBfordev:
Thanks, I have no comments.

PR #23438, opened 2023-04-10
Improvements to language tables generation including `since` support

vadz:
@utelle Here is my version, please let me know if you can integrate this with your changes or if you prefer not to use this (I think at least some of the changes here should be useful...).


vadz:
Replaced with #23453.

PR #23439, opened 2023-04-10
Set `wxUSE_STD_CONTAINERS=1` by default

vadz:


vadz:
This _will_ break something, but considering that we almost certainly won't know what exactly until we merge it, I'm going to do it soon if there are no objections.

Of course, it would be great if people could test this PR _before_ it's merged and let me know if it introduces any particularly egregious incompatibilities that might need to be fixed.

PR #23440, opened 2023-04-11
Remove avoidable uses of legacy container classes

vadz:
This is still WIP.


vadz:
This PR is independent of #23439 and can be done even without it. It's also probably less useful and more dangerous, as there is a real potential of breaking something here. But I still think it's worth doing it to simplify the code in the library itself and because having fewer ugly macro-based containers is always better -- even if many of them are still left because they are used in the public API and so can't be changed that easily.

As always, comments and testing would be very welcome, TIA!

PR #23443, opened 2023-04-12
Replace gtk_combo_box_text_insert_text with direct model insertion

imciner2:
There is a lot of overhead in the `gtk_combo_box_text_insert_text` function, so adding a lot of items to a choicebox can be an expensive operation when it is used. We have encountered this in KiCad when trying to use a wxChoiceBox as a font chooser on systems that have a lot of fonts installed (https://gitlab.com/kicad/code/kicad/-/issues/14277).

I traced back some of the slow performance on creation of the wxChoiceBox to the method that was being used to add the items to the combobox. Experiments showed that adding 10000 items to the choicebox takes 6.75s in the current implementation using `gtk_combo_box_text_insert_text`. The other way of adding items is to directly access the underlying data model using `gtk_list_store_insert_with_values`. Using the data model access, experiments showed the time needed to add 10000 items dropped to 438.2ms.

It turns out, that was already the method used in `wxBitmapComboBox`, so this actually collapses that down into the base `wxChoice` class, removing the overriden function from `wxBitmapComboBox`. I tried using the same 2-step insertion pattern from `wxBitmapComboBox` (calling `gtk_list_store_insert` followed by `gtk_list_store_set_value`, but it turns out this is just as slow as the original method, and using the 1-step method by calling `gtk_list_store_insert_with_values` was much faster).

I have split this into two separate commits. The first is ABI-compatible, so it would be good to backport it to 3.2. The second removes the now redundant function from `wxBitmapComboBox`, so it changes the ABI (but leaves the API). I would have preferred to just collapse out `GTKInsertComboBoxTextItem` entirely so each item didn't have to query for the model, but that would remove that from the API, and it could theoretically be being overriden by user's in their own implementations currently.

This doesn't completely solve the performance issues for the large choicebox though, there is still a large performance hit when computing the text size and showing the menu, but that needs to be handled separately (and I haven't quite figured out the best way to do that yet).

These experiments can be seen using the minimal sample with this diff:

```
index 501caf9096..428f3db1b5 100644
@@ -144,6 +144,9 @@ MyFrame::MyFrame(const wxString& title)
     // set the frame icon
     SetIcon(wxICON(sample));
 
+    wxBoxSizer* main_sizer = new wxBoxSizer(wxVERTICAL);
+    SetSizer(main_sizer);
+
 #if wxUSE_MENUBAR
     // create a menu bar
     wxMenu *fileMenu = new wxMenu;
@@ -162,12 +165,13 @@ MyFrame::MyFrame(const wxString& title)
     // ... and attach this menu bar to the frame
     SetMenuBar(menuBar);
 #else // !wxUSE_MENUBAR
+
     // If menus are not available add a button to access the about box
     wxSizer* sizer = new wxBoxSizer(wxHORIZONTAL);
     wxButton* aboutBtn = new wxButton(this, wxID_ANY, "About...");
     aboutBtn->Bind(wxEVT_BUTTON, &MyFrame::OnAbout, this);
     sizer->Add(aboutBtn, wxSizerFlags().Center());
-    SetSizer(sizer);
+    main_sizer->Add(sizer);
 #endif // wxUSE_MENUBAR/!wxUSE_MENUBAR
 
 #if wxUSE_STATUSBAR
@@ -175,6 +179,18 @@ MyFrame::MyFrame(const wxString& title)
     CreateStatusBar(2);
 #endif // wxUSE_STATUSBAR
+
+    wxArrayString labels;
+
+    for( int i=0; i < 10001; i++)
+    {
+        labels.Add( wxString::Format( "Item %d", i ) );
+    }
+
+    wxChoice* choice = new wxChoice(this, wxID_ANY, wxDefaultPosition, wxDefaultSize, labels);
+    main_sizer->Add(choice);
+
+    Layout();
 }

```

The flamegraph from the original implementation is:

Choicebox_Original

The flamegraph from the proposed implementation is (note the different scale on the time axis):
Choicebox_SingleStage

### DIFF
--- a/src/gtk/choice.cpp
+++ b/src/gtk/choice.cpp
@@ -139,17 +139,6 @@ bool wxChoice::GTKHandleFocusOut()
     return wxChoiceBase::GTKHandleFocusOut();
 }
 
-void wxChoice::GTKInsertComboBoxTextItem( unsigned int n, const wxString& text )
-{
-    GtkComboBox* combobox = GTK_COMBO_BOX( m_widget );
-    GtkTreeModel *model = gtk_combo_box_get_model( combobox );
-    GtkListStore *store = GTK_LIST_STORE( model );
-    GtkTreeIter iter;
-
-    gtk_list_store_insert_with_values(store, &iter, n, m_stringCellIndex,
-                                      text.utf8_str().data(), -1);
-}
-
 int wxChoice::DoInsertItems(const wxArrayStringsAdapter & items,
                             unsigned int pos,
                             void **clientData, wxClientDataType type)
@@ -163,6 +152,10 @@ int wxChoice::DoInsertItems(const wxArrayStringsAdapter & items,
 
     int n = wxNOT_FOUND;
 
+    GtkTreeIter iter;
+    GtkTreeModel *model = gtk_combo_box_get_model( GTK_COMBO_BOX( m_widget ) );
+    GtkListStore *store = GTK_LIST_STORE( model );
+
     gtk_widget_freeze_child_notify(m_widget);
 
     for ( int i = 0; i < count; ++i )
@@ -173,7 +166,8 @@ int wxChoice::DoInsertItems(const wxArrayStringsAdapter & items,
         if (m_strings)
             n = m_strings->Add(items[i]);
 
-        GTKInsertComboBoxTextItem( n, items[i] );
+        gtk_list_store_insert_with_values(store, &iter, n, m_stringCellIndex,
+                                          items[i].utf8_str().data(), -1);
 
         m_clientData.Insert( nullptr, n );
         AssignNewItemClientData(n, clientData, i, type);
--- a/src/gtk/choice.cpp
+++ b/src/gtk/choice.cpp
@@ -141,11 +141,13 @@ bool wxChoice::GTKHandleFocusOut()
 
 void wxChoice::GTKInsertComboBoxTextItem( unsigned int n, const wxString& text )
 {
-#ifdef __WXGTK3__
-    gtk_combo_box_text_insert_text(GTK_COMBO_BOX_TEXT(m_widget), n, text.utf8_str());
-#else
-    gtk_combo_box_insert_text( GTK_COMBO_BOX( m_widget ), n,  text.utf8_str() );
-#endif
+    GtkComboBox* combobox = GTK_COMBO_BOX( m_widget );
+    GtkTreeModel *model = gtk_combo_box_get_model( combobox );
+    GtkListStore *store = GTK_LIST_STORE( model );
+    GtkTreeIter iter;
+
+    gtk_list_store_insert_with_values(store, &iter, n, m_stringCellIndex,
+                                      text.utf8_str().data(), -1);
 }
 
 int wxChoice::DoInsertItems(const wxArrayStringsAdapter & items,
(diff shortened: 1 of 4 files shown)


src/gtk/choice.cpp
```diff
@@ -161,6 +163,8 @@ int wxChoice::DoInsertItems(const wxArrayStringsAdapter & items,
 
     int n = wxNOT_FOUND;
 
+    gtk_widget_freeze_child_notify(m_widget);
```
vadz, line 166: Normally we prefer to use RAII helpers like `wxWindowUpdateLocker` for calling pairs of functions like this. Not sure if it's really needed if it's the only place where we're going to use it, but I think it might be useful elsewhere, in which case adding it to `include/wx/gtk/private.h` or some new file under `include/wx/gtk/private/` would be nice.

vadz (commented):
Thanks for improving this!

We can definitely merge this already or you could address the 2 minor remarks below if you'd like.

Concerning the API breakage: functions prefixed with `GTK` are specifically meant to be private and _not_ part of the public API and so we can remove it without problem. Unfortunately we do need to keep it in 3.2 to avoid breaking the ABI...

src/gtk/choice.cpp
```diff
@@ -141,11 +141,13 @@ bool wxChoice::GTKHandleFocusOut()
...
-#else
-    gtk_combo_box_insert_text( GTK_COMBO_BOX( m_widget ), n,  text.utf8_str() );
-#endif
+    GtkComboBox* combobox = GTK_COMBO_BOX( m_widget );
+    GtkTreeModel *model = gtk_combo_box_get_model( combobox );
+    GtkListStore *store = GTK_LIST_STORE( model );
+    GtkTreeIter iter;
+
+    gtk_list_store_insert_with_values(store, &iter, n, m_stringCellIndex,
+                                      static_cast<const char *>(text.utf8_str()), -1);
```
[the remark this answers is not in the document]
imciner2: Thanks, fixed in the fixup commit.

src/gtk/choice.cpp
imciner2: It looks like this is the only use of the `gtk_widget_freeze_child_notify` and the equivalent thaw function, and I am not sure if there will be other uses of it (do you have any thoughts of other places to use it?). I'm tempted to say we leave it as a non-RAII wrapper here unless there are other places we want it right now.

imciner2:
> Concerning the API breakage: functions prefixed with GTK are specifically meant to be private and not part of the public API and so we can remove it without problem. 

Thanks for clarifying this. I have gone ahead and removed it from the class then, collapsing everything into `DoInsertItems`. It is a separate commit, so backporting the core of this change to 3.2 can still be done by taking the first commit and its fixup.

src/gtk/choice.cpp
vadz: I thought it could be useful in `wxListBox::DoInsertItems()` too, but I didn't look at it in any details, so let's leave it like this until we need it there...

vadz:
Thanks for the updates! I think it would be better if you rebased the commits here, so that I could cherry-pick just a single commit to 3.2.

imciner2:
Ok, I have rebased everything so now only 1 commit (the first) needs to be cherry-picked, everything else is master-only.

vadz:
Great, thanks! Will merge (and backport) soon.

PR #23445, opened 2023-04-12
Add boolean operator to wxPGGlobalVars

imciner2:
This is just a minor tweak, but the old API for `wxPGGlobalVars` allowed for users to directly test the variable and find if it was null, so recreate that API with the new system to preserve compatibility.

### DIFF
--- a/include/wx/propgrid/propgrid.h
+++ b/include/wx/propgrid/propgrid.h
@@ -113,6 +113,8 @@ class WXDLLIMPEXP_PROPGRID wxPGGlobalVarsPtr
 {
 public:
     wxPGGlobalVarsClass* operator->() const;
+    bool operator!() const;
+    explicit operator bool() const;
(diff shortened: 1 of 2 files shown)


vadz (commented):
Thanks, I agree that this should be added to improve compatibility, but I wonder if we should also add an `explicit operator bool() const` to it, as I'd expect there to be some code testing it just with `if (wxPGGlobalVars)` too.

imciner2:
Good point. I actually just hit a case where that is needed -`wxASSERT( wxPGGlobalVars );`

vadz:
Thanks, I'll merge this soon if there are no objections from @a-wi

vadz:
Sorry, I've completely forgot about this one, but as there have been no objections since several weeks from @a-wi I'm finally going to merge it soon. Thanks again!

PR #23449, opened 2023-04-15
Remove separate STL build mode

vadz:
We always use standard classes now and don't need a separate build mode for this.


vadz:
I'd like to merge this too soon. Except for the incompatible change to the conversions to `std::[w]string` in STL build, this should be backwards-compatible but simplifies things a bit.

Issue #23379, opened 2023-03-27
Key events do not return key code when non-latin keyboard layout is active


unxed:
Under GTK, the problem [can be worked around](https://github.com/wxWidgets/Phoenix/issues/1391#issuecomment-1484663351) by calling GetRawKeyFlags and then using XGetKeyboardMapping to get all possible key symbols associated with this raw x11 key code. So if a button with a Latin letter is pressed while a non-Latin keyboard layout is enabled, we can still recognize the Latin letter. This forces us to additionally link against the x11 libraries, but at least it works.

Under other platforms, there doesn't seem to be a solution.

Related issues in other projects:
https://github.com/wxWidgets/Phoenix/issues/1391
https://github.com/elfmz/far2l/issues/1562

imciner2:
Did you try using `GetUnicodeKey()` instead of `GetKeyCode()`? The docs for `GetKeyCode()` explicetely say that it doesn't work for non-latin keyboard layouts and that `GetUnicodeKey()` should be used instead (https://docs.wxwidgets.org/stable/classwx_key_event.html#a8214333c112432c4c9df114d3bd9122c).

unxed:
Of course GetUnicodeKey() is working. But it can not give me any information that identify pressed key, not char. But I want to know that key was pressed exactly. Only GetRawKeyFlags() can identify key independently of current selected keyboard layout.

I know it is a documented behavior, so maybe this should be called not a bug report, but a "feature request", like "provide wx key codes in key events for non-latin keyboard layouts also"?

Usage example: I want to use accelerator Ctrl+`>` and want it to work in any keyboard layout. Classical wx accelerators only support Latin keys, so I can not use them for it and need to process keydown events by myself. But in, for example, Russian keyboard layout Ctrl+`>` is Ctrl+`ю`, so keydown event gives `ю` as unicode char and no way to match it to `>` key except for 
GetRawKeyFlags() hack.

vadz:
This seems to be very similar, if not quite the same, as #17643, and has also been discussed on the mailing list quite a few times. The current situation is that wx API indeed doesn't provide a way to uniquely identify a keyboard key portably, i.e. we don't have any portable scan code equivalent.

[...]

PR #23452, opened 2023-04-15
Simplify wxKeyEvent generation in wxGTK a bit

vadz:
Try to make the code generating key events a bit more clear and also avoid duplicating the checks in AdjustCharEventKeyCodes() by returning true from this function if the key code passed them.

This shouldn't result in any real changes in behaviour.

---

@unxed I've seen your (apparently deleted now?) comment in #23379 and while trying to understand it, I've realized that the logic here could be simplified a bit, so I've tried doing it in this PR. Do you see any problems with doing it like this? The new version looks more comprehensible to me, so I'd like to merge this if it still works for you. TIA!

### DIFF
--- a/src/gtk/window.cpp
+++ b/src/gtk/window.cpp
@@ -1268,9 +1268,13 @@ bool SendCharHookEvent(const wxKeyEvent& event, wxWindow *win)
 // Adjust wxEVT_CHAR event key code fields. This function takes care of two
 // conventions:
 // (a) Ctrl-letter key presses generate key codes in range 1..26
-// (b) Unicode key codes are same as key codes for the codes in 1..255 range
-void AdjustCharEventKeyCodes(wxKeyEvent& event)
+// (b) Unicode key codes are same as key codes for the codes in ASCII range
+//
+// Return true if the key code was modified.
+bool AdjustCharEventKeyCodes(wxKeyEvent& event)
 {
+    bool modified = false;
+
     const int code = event.m_keyCode;
 
     // Check for (a) above.
@@ -1286,14 +1290,22 @@ void AdjustCharEventKeyCodes(wxKeyEvent& event)
 
         // Adjust the Unicode equivalent in the same way too.
         if ( event.m_keyCode != code )
+        {
             event.m_uniChar = event.m_keyCode;
+            modified = true;
+        }
     }
 
     // Check for (b) from above.
     //
     // FIXME: Should we do it for key codes up to 255?
     if ( !event.m_uniChar && code < WXK_DELETE )
+    {
         event.m_uniChar = code;
+        modified = true;
+    }
+
+    return modified;
 }
 
 } // anonymous namespace
@@ -1413,16 +1425,7 @@ gtk_window_key_press_callback( GtkWidget *WXUNUSED(widget),
         {
             wxKeyEvent eventChar(wxEVT_CHAR, event);
 
-            // Check for the special case of Ctrl+letter, see comment before
-            // AdjustCharEventKeyCodes().
-            if ( event.ControlDown() &&
-                    ((event.m_keyCode >= 'a' && event.m_keyCode <= 'z') ||
-                     (event.m_keyCode >= 'A' && event.m_keyCode <= 'Z')) )
-            {
-                eventChar.m_keyCode = event.m_keyCode;
-                eventChar.m_uniChar = event.m_uniChar;
-            }
-            else
+            if ( !AdjustCharEventKeyCodes(eventChar) )
             {
                 // use Unicode values
                 eventChar.m_keyCode = key_code;
@@ -1430,8 +1433,6 @@ gtk_window_key_press_callback( GtkWidget *WXUNUSED(widget),
             }
 
             wxLogTrace(TRACE_KEYS, wxT("Char event: %ld"), eventChar.m_keyCode);
-
-            AdjustCharEventKeyCodes(eventChar);
 
             ret = win->HandleWindowEvent(eventChar);
         }


unxed:
My comment was written due to a mistake. I've forgotten that zero keycode in OnChar is ok. On Windows we have character code in Windows ANSI 8-bit code page as Key Code in OnChar event, and it makes no sence reproducing this behavior on Unix platforms.

I compiled the code from your pull request, ran some tests on the keyboard demo app. The difference is that now when you press Alt+RussianLetter, the key code in the OnChar event is not 0, as it was before, but contains the corresponding Latin letter in uppercase. There is some inconsistency in this: if you press a Russian letter without Alt, then the key code in OnChar will still be zero, and if you press the same button in the Latin layout, the key code will contain a lowercase letter.

I think we need to globally rethink the meaning of the KeyCode field in the OnChar event.

If you were to ask me how to do it, I would suggest the following. Currently on Windows there is a character code in Windows ANSI encoding (8 bits) in KeyCode field of OnChar events, and on GTK there is zero there. I don’t see any point in this: no one uses the ANSI encoding today, and zero does not carry any information at all. Moreover, if we need to get both the key code and the character code at the same time, we have to listen to the KeyDown event, remember the key code, and then listen to the OnChar event and match the key code and character code. This is how it works in far2l now. Pretty weird, isn't it? I suggest setting the OnChar event to the same key code value as the KeyDown event on all platforms for all keys and key combinations (except for Ctrl+letters). Unfortunately, I can't offer a proper PR since I don't have Windows to test the changes under it, but the changes are very simple, I think you can easily make them yourself if you think it's reasonable.

unxed:
The suggested change will break the assumption that Unicode key value should be always equal to keycode for latin-1 keys, so this keyboard event test should also be rewritten:
```
    if ( desc.m_keycode < WXK_START )
    {
        // For Latin-1 our key code is the same as Unicode character value.
        CPPUNIT_ASSERT_EQUAL_MESSAGE( "wrong Unicode key in " + msg,
                                      (char)desc.m_keycode,
                                      (char)ev.GetUnicodeKey() );
    }
    else // Special key
    {
        // Key codes above WXK_START don't correspond to printable characters.
        CPPUNIT_ASSERT_EQUAL_MESSAGE( "wrong non-zero Unicode key in " + msg,
                                      0,
                                      (int)ev.GetUnicodeKey() );
    }
```

vadz:
> My comment was written due to a mistake. I've forgotten that zero keycode in OnChar is ok. On Windows we have character code in Windows ANSI 8-bit code page as Key Code in OnChar event, and it makes no sence reproducing this behavior on Unix platforms.

Thanks for the explanation, but I actually disagree here: I think that the behaviour should be the same on all platforms unless there are some really good reasons to diverge from it.
 
> I compiled the code from your pull request, and ran some tests on the keyboard demo app. The difference is that now when you press Alt+RussianLetter, the key code in the OnChar event is not 0, as it was before, but contains the corresponding Latin letter in uppercase. There is some inconsistency in this: if you press a Russian letter without Alt, then the key code in OnChar will still be zero, and if you press the same button in the Latin layout, the key code will contain a lowercase letter.

Yes, this is wrong :-(
 
> I think we need to globally rethink the meaning of the KeyCode field in the OnChar event.

I'm not so sure. It seems to work correctly for me under MSW, AFAICS we just need to make wxGTK work in the same way.

> If you were to ask me how to do it, I would suggest the following. Currently on Windows there is a character code in Windows ANSI encoding (8 bits) in KeyCode field of OnChar events, and on GTK there is zero there (if we are talking about master branch behavior). I don’t see any point in this: no one uses the ANSI encoding today, and zero does not carry any information at all. 

Nobody uses ANSI encoding but plenty of people use US English keyboards and it makes sense to keep providing the key code in `m_keyCode` for them. We should, ideally, do it on all platforms, however.

> Moreover, if we need to get both the key code and the character code at the same time, we have to listen to the KeyDown event, remember the key code, and then listen to the OnChar event and match the key code and character code.

Why does far2l need to do this? The idea is that you need to use `wxEVT_KEY_DOWN` if you care about actual key presses and `wxEVT_CHAR` if you care about text entry. In which situation do you need to handle both?

> I suggest setting the OnChar event to the same key code value as the KeyDown event on all platforms for all keys and key combinations.

This would silently break existing applications currently working correctly under MSW (and probably Mac too), so this is not really a good idea, we try to avoid breaking compatibility unless it's really necessary and try to avoid silent breakages really hard.

> Unfortunately, I can't offer a proper PR since I don't have Windows to test the changes under it, but the changes are very simple, I think you can easily make them yourself if you think it's reasonable.

I can test the changes under MSW (or, if you add more tests to `tests/events/keyboard.cpp`, they could get tested by the CI), but I'm not sure how to make wxGTK behave in the same way as wxMSW...

unxed:
> Nobody uses ANSI encoding but plenty of people use US English keyboards and it makes sense to keep providing the key code in m_keyCode for them. We should, ideally, do it on all platforms, however.

So I just propose to pass the corresponding Latin character in the m_keyCode for character keys, even if non-Latin layout is selected, just as is done now in KeyUp/Down events. Just do the same for OnChar. Also it should always be in upper case, just as it is in KeyUp/Down events.

> Why does far2l need to do this? The idea is that you need to use wxEVT_KEY_DOWN if you care about actual key presses and wxEVT_CHAR if you care about text entry. In which situation do you need to handle both?

far2l is ported from Windows, and uses a kind of compatibility layer to emulate some part of the Windows API. So input events coming from wx are translated into equivalent of Windows WM_KEYDOWN/UP events, which are further processed by existing Far Manager code. I don't know Winodws console API well enough to explain why I had to be done this way. Maybe @elfmz could explain it better.

unxed:
>> I compiled the code from your pull request, and ran some tests on the keyboard demo app. The difference is that now when you press Alt+RussianLetter, the key code in the OnChar event is not 0, as it was before, but contains the corresponding Latin letter in uppercase. There is some inconsistency in this: if you press a Russian letter without Alt, then the key code in OnChar will still be zero, and if you press the same button in the Latin layout, the key code will contain a lowercase letter.

> Yes, this is wrong :-(

First of all, let us fix the first part of the problem: with the patch below, OnChar events:
1) from the press of Russian letter key alone and
2) from the press of the same key with Alt
both have Latin equivalent in Key Code field. This is unlikely to break anything, since the zero keycodes that were there before are probably not used by anyone.

Let's first agree that this is a good solution before discussing anything else — just to avoid mess and misunderstanding.

```
diff --git a/src/gtk/window.cpp b/src/gtk/window.cpp
index 85f01f2a71..77f6e56a4e 100644
--- a/src/gtk/window.cpp
+++ b/src/gtk/window.cpp
@@ -1485,8 +1485,29 @@ bool wxWindowGTK::GTKDoInsertTextFromIM(const char* str)
     for( wxString::const_iterator pstr = data.begin(); pstr != data.end(); ++pstr )
     {
         event.m_uniChar = *pstr;
+        event.m_keyCode = 0;
+
+#ifdef wxHAS_XKB
+        char key_code_str[64];
+        xkb_state_key_get_utf8(gs_xkbData.GetState(),
+                               m_imKeyEvent->hardware_keycode,
+                               key_code_str,
+                               sizeof(key_code_str));
+        if ( strlen(key_code_str) == 1 )
+        {
+            event.m_keyCode = key_code_str[0];
+
+            // If key_code is a Latin char, it should be in upper register to
+            // match the other ports.
+            if (islower(event.m_keyCode))
+            {
+                event.m_keyCode = toupper(event.m_keyCode);
+            }
+        }
+#endif // wxHAS_XKB
+        
         // Backward compatible for ISO-8859-1
-        event.m_keyCode = *pstr < 256 ? event.m_uniChar : 0;
+        event.m_keyCode = *pstr < 256 ? event.m_uniChar : event.m_keyCode;
         wxLogTrace(TRACE_KEYS, wxT("IM sent character '%c'"), event.m_uniChar);
 
         AdjustCharEventKeyCodes(event);

```

unxed:
English/Russian one

vadz:
> First of all, let us fix the first part of the problem: with the patch below, OnChar events:
> 
>     1. from the press of Russian letter key alone and
>     2. from the press of the same key with Alt
> 
> both have Latin equivalent in Key Code field. This is unlikely to break anything, since the zero keycodes that were there before are probably not used by anyone.
> 
> Let's first agree that this is a good solution before discussing anything else — just to avoid mess and misunderstanding.

I'm sorry, I know that I wrote that I agreed with this previously, but I've actually retested this now and the current behaviour is consistent for all 3 platforms: key code is set to 0 when unicode char is a non-ASCII letter (whether Alt is pressed or not).

Knowing this, I think I prefer to keep it like this to avoid introducing new platform differences.

Sorry again for the confusion, but I somehow thought that you wrote that GTK behaviour differed from MSW, when actually they're the same (and Mac behaves like this too).

unxed:
In this case, it is better to reject this PR, because after it, the key code is set to a capital Latin letter in OnChar events in some cases. We can just leave master it as is, that's completely acceptable.

vadz:
I'd still like to avoid duplicating the checks from `AdjustCharEventKeyCodes()` here, so I've updated the PR to do just this (in fact, this was always my intention, but I just misunderstood what the code did and so updated it incorrectly -- which, I guess, proves my point that the code is not that easy to understand...).

AFAICS the latest version should work in the same way as the current master, doesn't it?

unxed:
Retested. Yes, it looks like behavior now is like in master.

vadz:
Thanks for testing! Merged now.

Issue #23437, opened 2023-04-10
Further glitches in the list of known languages/locales


utelle:
In the course of streamlining the process to regenerate the file `misc/languages/langtabl.txt` containing the list of languages/locales with associated `wxLANGUAGE_*` symbols, I discovered 2 glitches:

1. The symbol `wxLANGUAGE_CENTRAL_KURDISH` is listed 2 times, for `ku` and `ku-Arab`.
2. The symbol `wxLANGUAGE_JAVANESE`  is listed 2 times, for `jv` and `jv-Latn`.

These glitches were present since the list of languages was extended (from 234 to 877 entries) in wxWidgets version 3.1.6. This duplication went undetected, because the script `genlang.py` silently eliminates double entries of language symbols.

The affected locales are rather "exotic" and I doubt that any productive wxWidgets application is/was affected. Therefore I would not backport the necessary fixes to version 3.2, because fixes would inevitably introduce new symbols.

Additionally, I'd like to bring to your attention that the list of languages associates 3 generic language entries with certain region specific entries where the list of known locales under Windows uses different associations:

1. Chinese `zh` is associated with `zh_TW` (Taiwan), while under Windows it is associated with `zh_CN` (China).
2. English `en` is associated with `en_GB` (Great Britain, United Kingdom), while under Windows it is associated with `en_US` (USA).
3. Portuguese `pt` is associated with `pt_PT` (Portugal), while under Windows it is associated with `pt_BR` (Brazil).

For wx the associations (`zh` <=> `zh_TW`), (`en` <=> `en_GB`), and (`pt` <=> `pt_PT`) were chosen for compatibilty reasons with prior wxWidgets versions. However, most operating systems use the associations  (`zh` <=> `zh_CN`), (`en` <=> `en_US`), and (`pt` <=> `pt_BR`).

The big question is whether we should keep the current wx associations (for compatibility) or whether we should replace them by the more common ones (as used in most operating systems). Opinions?

vadz:
I am not sure at all what's the right thing to do here, so I'd rather keep the definitions as they are and just document somewhere that `wxLANGUAGE_ENGLISH` etc might not have the expected definition and that the more precise versions such as `wxLANGUAGE_ENGLISH_US` should be used instead.

utelle:
> I am not sure at all what's the right thing to do here, so I'd rather keep the definitions as they are

Keep "as they are" where? As currently in the wxWidgets list? Or as in the Windows defaults?

[...]

PR #23453, opened 2023-04-16
Fix issue #23437, streamline regeneration process

utelle:
- A few glitches (like duplicate language symbols, wrong Windows language ids etc) were fixed
- Scripts were added to allow to regenerate the list of supported locales from the list of known Windows locales
- Version info when language symbols became available was added to the documentation of the language symbol enum
- Synonyms are no longer hard-coded in the script genlang.py


vadz:
Thanks Ulrich!

I didn't have time to look at it yet, but I'd just like to ask: what should we do with #23438 now? Does this one replace it or ...?

utelle:
> I didn't have time to look at it yet, but I'd just like to ask: what should we do with #23438 now? Does this one replace it or ...?

Yes, this one replaces it. I integrated all changes you made to `genlang.py` into the new version provided by this PR.

The wx version `since` is now an additional column in `langtabl.txt` and `synonymtabl.txt`. I updated the version info to mark all languages/locales that were added in version 3.1.6 accordingly.

Additionally, I added the version info to the synonyms, too. However, I didn't check how the generated documentation will look like and how useful this information really is. Actually, the meaning is when the language symbol was introduced the first time (as a normal language symbol, not as a synonym).

In the `README` file I described the regeneration process in some detail.

utelle:
One of the checks, namely the one for C++11 features, fails, because the source of the utility `showlocales` (`showlocales.c`) uses `NULL` for null pointers. Well, the utility is a simple C application, and C doesn't know `nullptr`. Of course, I can replace `NULL` by `0`, to silence the check, if you prefer.

vadz:
Thanks for the changes! I didn't have time to actually run the scripts, but they look good to me, so I'll merge this soon unless somebody has any comments about it.

As for the `NULL` problem, I think we can restrict the check to `*.h` and `*.cpp` files only, which would avoid triggering it in a `*.c` file, see #23459.

vadz:
Thanks again, I've merged this now (and so closing the PR), but I had some changes after rerunning `genlang.py`:
```diff
diff --git a/src/common/languageinfo.cpp b/src/common/languageinfo.cpp
index 0661be5906..e345760402 100644
--- a/src/common/languageinfo.cpp
+++ b/src/common/languageinfo.cpp
@@ -1034,6 +1034,7 @@
     { "Cher", "cherokee" },
     { "Chrs", "chorasmian" },
     { "Copt", "coptic" },
+    { "Cpmn", "cypro_minoan" },
     { "Cprt", "cypriot" },
     { "Cyrl", "cyrillic" },
     { "Deva", "devanagari" },
@@ -1114,6 +1115,7 @@
     { "Orya", "oriya" },
     { "Osge", "osage" },
     { "Osma", "osmanya" },
+    { "Ougr", "old_uyghur" },
     { "Palm", "palmyrene" },
     { "Pauc", "pau_cin_hau" },
     { "Perm", "old_permic" },
@@ -1156,8 +1158,11 @@
     { "Thai", "thai" },
     { "Tibt", "tibetan" },
     { "Tirh", "tirhuta" },
+    { "Tnsa", "tangsa" },
+    { "Toto", "toto" },
     { "Ugar", "ugaritic" },
     { "Vaii", "vai" },
+    { "Vith", "vithkuqi" },
     { "Wara", "warang_citi" },
     { "Wcho", "wancho" },
     { "Xpeo", "old_persian" },
```

They look correct (i.e. correspond to `misc/languages/scripttabl.txt` contents) to me, so I've just added them to the commit, but I wonder why they hadn't been included originally.

vadz:
Yes, I think it would be better to add these files to Git because it doesn't make much sense to have some, but not all, generated files there -- it should be either all or none and we tend to err on the side of "all" in wx. And it is convenient if only because you can see the changes easily with `git diff` after regenerating them... TIA!

utelle:
I added the missing data files and adjusted the local `.gitignore` file (see PR #23465).

The data files downloaded from Unicode, the generated list of known Windows locales, and the current  version of  the files `langtabl.txt`, `scripttabl.txt`, and `synonymtabl.txt` form the basis for the regeneration process.

Issue #23450, opened 2023-04-15
macOS: wxString::Format() fails with some Unicode strings after creating wxLocale


lanurmi:
### Description

Creating a `wxLocale` object, i.e. changing locale, breaks formatting certain (but not all) Unicode strings using `wxString::Format()` on macOS. The string returned by `Format()` will be empty in this case. This is a regression; it did not happen with 3.0.x nor most of 3.1.x versions.

Observations are made on the 3.2 branch, though I strongly suspect the master branch is equally affected.

To Reproduce:
1. Overwrite `samples/internat/internat.cpp` by copying `internat.cpp` *from the 3.0 branch*.
2. Add e.g. `SetTitle(wxString::Format(wxString::FromUTF8("Hmm %d √ 10"), 1));` to the beginning of `MyFrame::MyFrame()` body.
3. Build the sample and run it *via Finder* (not via command line).
4. Choose "System default" as the language. (On my system, the primary language is Finnish, secondary is German (Germany)).
5. Observe that the frame's title is unexpectedly empty.
6. Observe that running the sample via command line (`open internat.app`) results in the expected non-empty title.

Reasons:
It is apparently a well known fact that `LC_CTYPE` needs to be set to `"UTF-8"` to have `wxString::Format()` work as expected, and [`src/common/init.cpp` does precisely this](https://github.com/wxWidgets/wxWidgets/blob/3.2/src/common/init.cpp#L231-L241). Presumably `LC_CTYPE` gets (re)set to something else when creating a `wxLocale` object, and consequent formatting subtly fails for some strings. It does not, however, fail for simple accented letters (`äöüé..`), so the problem may go unnoticed for quite a while even in localized applications.

Fix/workaround:
Calling `setlocale(LC_CTYPE, "UTF-8");` again in `wxLocale::DoCommonPostInit()` seems to remedy the situation. Pull request to be expected soon.

### Platform and version information

- wxWidgets version  you use: 3.2 branch
- wxWidgets port  you use: wxOSX
- OS  and its version: macOS 12.6.5

lanurmi:
Fixed in #23454, a long time ago.

PR #23454, opened 2023-04-16
Fix/workaround for wxLocale breaking Unicode string formatting

lanurmi:
Also attempt to test whether wxLocale breaks formatting Unicode strings.
See #23450.

### DIFF
--- a/src/common/intl.cpp
+++ b/src/common/intl.cpp
@@ -52,6 +52,7 @@
 #include "wx/hashset.h"
 #include "wx/uilocale.h"
 
+#include "wx/private/localeset.h"
 #include "wx/private/uilocale.h"
 
 #ifdef __WIN32__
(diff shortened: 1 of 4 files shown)


src/common/intl.cpp
```diff
@@ -396,6 +396,12 @@ bool wxLocale::DoCommonPostInit(bool success,
             t->AddStdCatalog();
     }
 
+#if defined(__WXOSX__)
+    // LC_CTYPE is set in DoCommonPreInit() (see init.cpp). Set it again here,
```
vadz, line 400: I think we might want have some `wxLocale::OSXFixCRT()` or something like this and call it from both places, even if it's a trivial one-line function because this would allow documenting this issue in a single place -- and hopefully get rid of it easier in the future, too.

vadz (commented):
Thanks for the fix, we should, of course, do this if it doesn't work without it even if it's somewhat unexpected for `LC_CTYPE` to "change on its own" like this (I'd like to document it, but I'm not sure where).

I think this should also be backported to 3.2?

lanurmi:
> Thanks for the fix, we should, of course, do this if it doesn't work without it even if it's somewhat unexpected for `LC_CTYPE` to "change on its own" like this (I'd like to document it, but I'm not sure where).

It is somewhat unexpected, though it has been changing like that already in 2.9.something. From the caller's perspective, this fix prevents it from changing again and retains old behaviour.

> I think this should also be backported to 3.2?

Absolutely, this is an issue that causes subtle bugs with random strings missing for no apparent reason or pattern.

src/common/intl.cpp
vadz: I've done this now, please let me know if you have any objections.

PR #23455, opened 2023-04-17
Branch 3.2.2 hotfix kh

Khalyutin:
When compile with clang 16.0.0.0 on Windows, We are have errors
unknown type name 'uintptr_t' and unknown type name 'intptr_t'
without #include <inttypes.h> in src\stc\scintilla\include\Scintilla.h


.gitignore
```diff
@@ -422,3 +422,4 @@
 /utils/wxrc/wxrc_vc[789].sln
 
 /3rdparty/webview2
+build
```
imciner2, line 425: This is incorrect. This directory is needed by wxWidgets and should not be excluded. Please remove it from the ignore file and restore its contents.

imciner2 (changes requested):

.gitignore
Khalyutin: We are compile with Clang Cmake and vsCode , and after reconfigure and rebuild, We have more then 800 chabges...  
Is "Commint" it correct?

Khalyutin (commented):
revert ignore "build" By instruction of  imciner2

vadz:
Sorry, this is a very strange change. First of all, it shouldn't change anything but `src/stc/scintilla/include/Scintilla.h`, of course. Second, the change there doesn't seem correct neither, as nobody defines `wxNO_INTTYPES_H`, so this will result in including `<inttypes.h>` for all MSVC versions, breaking the build with older ones that don't have this header.

If the problem is clang-specific, it would make much more sense to add a test for `__clang__` and include `<inttypes.h>` for it.

Please redo this PR without the extra changes and not affecting the other MSVC versions.

Also please make it against 3.2 branch and not the short-lived hotfix branch which won't be updated any longer.

Thanks!

PR #23456, opened 2023-04-17
Use wxID_HIGHEST instead of hard-coded numbers in samples

Blake-Madden:


vadz:
Thanks for improving this! I will merge it soon.

Issue #23457, opened 2023-04-17
The implementation of wxRibbonToolBar::GetToolByPos(size_t pos) is broken


mcorino:
### Description
The implementation of `wxRibbonToolBar::GetToolByPos(size_t pos)` is broken for retrieving any tool item beyond the first group.

My intention was to iterate tool items in a `wxRibbonToolBar` by using `wxRibbonToolBar::GetToolCount()`  and `wxRibbonToolBar::GetToolByPos(size_t pos)` but after reading the code in `src/ribbon/toolbar.cpp` I had to conclude that that would not work although the code in `wxRibbonToolBar::GetToolPos(int tool_id)` made clear that it clearly was intended to work.

The code in `wxRibbonToolBar::GetToolPos(int tool_id)` correctly determines the position index for a tool item taking any preceding groups before the group containing the tool item  into account but the code in `wxRibbonToolBar::GetToolByPos(size_t pos)` fails to take those same group offsets into account so no position other than one from the first group will ever return a valid tool item.

To properly function  `wxRibbonToolBar::GetToolByPos(size_t pos)`  should be changed like this:

```C++
wxRibbonToolBarToolBase* wxRibbonToolBar::GetToolByPos(size_t pos)const
{
    size_t group_count = m_groups.GetCount();
    size_t g;
    for(g = 0; g < group_count; ++g)
    {
        wxRibbonToolBarToolGroup* group = m_groups.Item(g);
        size_t tool_count = group->tools.GetCount();
        if(pos<tool_count)
        {
            return group->tools[pos];
        }
        else if(pos==tool_count)
        {
            return nullptr;
        }
        pos -= tool_count+1;  // FIX CODE
    }
    return nullptr;
}
```

### Platform and version information

- wxWidgets version  you use: master
- wxWidgets port  you use: wxMSW
- OS  and its version: Windows 10

vadz:
Sorry, but are you really sure this is the right fix? `pos` is unsigned, and the code above subtracts a value greater than it from it, why would this be the correct thing to do?

We need to add at least an interactive test for this to `samples/ribbon` (or, ideally, an automatic one to the test suite) and check that it really works.

Please don't hesitate to make PRs with the proposed changes.

mcorino:
[...]

PR #23458, opened 2023-04-17
Fix wxRibbonToolBar::GetToolByPos(size_t) 

mcorino:
See #23457


vadz:
Thanks for fixing this, (squash-)merged now!

PR #23460, opened 2023-04-17
Fix comment about wxWebResponse::IsOk()'s return type

Blake-Madden:
`wxWebResponse::IsOk()` returns a boolean, not a pointer.


vadz:
Applied, thanks for fixing this!

PR #23461, opened 2023-04-17
Fix GetMimeType() returning empty

Blake-Madden:
I believe it should be "Content-Type" to request the MIME type from a response header, not "Mime-Type".

https://developer.mozilla.org/en-US/docs/web/http/headers/content-type

As an example, currently wxWebResponse::GetMimeType() will return empty string for any page (e.g., "https://www.libreoffice.org/discover/libreoffice/").  With this fix, I will now get "text/html; charset=utf-8" (tested on Windows).

### DIFF
--- a/interface/wx/webrequest.h
+++ b/interface/wx/webrequest.h
@@ -598,7 +598,8 @@ public:
     /**
         Returns the content type of the response (if available).
 
-        This can be a value such as "text/html; charset=utf-8".
+        This is the full value of the "Content-Type" header of the response,
+        e.g. a value such as "text/html; charset=utf-8".
 
         @since 3.3.0
     */
--- a/interface/wx/webrequest.h
+++ b/interface/wx/webrequest.h
@@ -589,8 +589,17 @@ public:
 
     /**
         Returns the MIME type of the response (if available).
+
+        This can be a value such as "text/html".
     */
     wxString GetMimeType() const;
+
+    /**
+        Returns the content type of the response (if available).
+
+        This can be a value such as "text/html; charset=utf-8".
+    */
+    wxString GetContentType() const
 
     /**
         Returns the status code returned by the server.
--- a/interface/wx/webrequest.h
+++ b/interface/wx/webrequest.h
@@ -598,6 +598,8 @@ public:
         Returns the content type of the response (if available).
 
         This can be a value such as "text/html; charset=utf-8".
+
+        @since 3.3.0
     */
     wxString GetContentType() const
 
--- a/interface/wx/webrequest.h
+++ b/interface/wx/webrequest.h
@@ -590,7 +590,8 @@ public:
     /**
         Returns the MIME type of the response (if available).
 
-        This can be a value such as "text/html".
+        This is just the MIME type part (e.g. "text/html") of the value returned
+        by GetContentType().
     */
     wxString GetMimeType() const;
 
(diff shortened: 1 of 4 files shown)


vadz:
Thanks, I'm not sure what is this function supposed to return, but using `Mime-Type` header is definitely wrong. My question is about whether it should really return the entire `Content-Type` header contents or just the MIME type, i.e. `text/html` part.

@TcT2k would you have any thoughts about this?

Blake-Madden:
curl returns the charset info in this string also. For example, when I do this:

```
char* ct(nullptr);
res = curl_easy_getinfo(curl_handle, CURLINFO_CONTENT_TYPE, &ct);
```
then `ct` will be "text/html; charset=utf-8".

Perhaps we should have separate `GetMimeType()` and `GetContentType()` functions to make this more clear?

interface/wx/webrequest.h
```diff
@@ -589,9 +589,18 @@ class wxWebResponse
...
+
+        This can be a value such as "text/html".
     */
     wxString GetMimeType() const;
 
+    /**
+        Returns the content type of the response (if available).
+
+        This can be a value such as "text/html; charset=utf-8".
+    */
```
vadz, line 601:
```suggestion

        @since 3.3.0
    */
```

src/common/webrequest.cpp
```diff
@@ -636,6 +636,14 @@ void wxWebResponseImpl::Init()
 }
 
 wxString wxWebResponseImpl::GetMimeType() const
+{
+    wxString contentType( GetContentType() );
+    const auto separatorPosition = contentType.find(';');
+    return (separatorPosition != wxString::npos) ?
+        contentType.substr(0, separatorPosition) : contentType;
```
vadz, line 643: In general I'm for using the standard string functions, but in this particular case all this could be replaced with just
```suggestion
    return GetContentType().BeforeFirst(';');
```
which seems worth it as it's shorter and much more readable.

interface/wx/webrequest.h
```diff
@@ -589,9 +589,18 @@ class wxWebResponse
 
     /**
         Returns the MIME type of the response (if available).
+
+        This can be a value such as "text/html".
```
vadz, line 593:
```suggestion
        This is just the MIME type part (e.g. "text/html") of the value returned
        by GetContentType().
```

interface/wx/webrequest.h
```diff
@@ -589,9 +589,18 @@ class wxWebResponse
...
     */
     wxString GetMimeType() const;
 
+    /**
+        Returns the content type of the response (if available).
+
+        This can be a value such as "text/html; charset=utf-8".
```
vadz, line 600:
```suggestion
        This is the full value of the "Content-Type" header of the response,
        e.g. a value such as "text/html; charset=utf-8".
```

vadz (commented):
Thanks for the changes, I agree with this, but I'd like to improve the documentation a bit. If you agree, I can merge the proposed changes myself, otherwise please feel free to incorporate them in the way you prefer.

src/common/webrequest.cpp
Blake-Madden: Totally fine with me.  I was just trying to use the more "std::string-like" API, a habit I have gotten into.  But yes, this is much easier to read.

vadz:
Thanks again, will merge in a moment.

This could be backported to 3.2 if anybody thinks it's worth doing it...

PR #23465, opened 2023-04-18
Add missing data files for regenerating locale list

utelle:
1) Data files downloaded from Unicode
2) Unicode Data Files and Software license
3) List of known Windows locales (generated by showfiles utility)

### DIFF
--- a/misc/languages/data/.gitignore
+++ b/misc/languages/data/.gitignore
@@ -4,9 +4,14 @@
 # exclude database and data files
 *.db3
 *.txt
+*.bak
 
 # exclude everything in temp
-temp/*
+temp/**
 
 # exception to the rule
 !temp/.gitkeep
+
+# exceptions: downloaded and generated data files 
+!unicode/*.txt
+!windows/*.txt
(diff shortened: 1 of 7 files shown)


utelle:
The Whitespace Check fails for the data files downloaded from Unicode. The problem is that it wouldn't help to remove the whitespace manually, because the next download from Unicode would bring it back. Is it somehow possible to exclude those files from the check?

vadz:
I think it's simpler to just remove the whitespace and then either do it again after the next update (which probably won't happen very soon) or remove it in the regeneration scripts (which shouldn't be difficult neither, but Lua is my 17th language so I'm not exactly sure how...).

So if you don't mind, I'll just push this as is (well, after removing whitespace) for now.

And thanks for the update, of course!

utelle:
> I think it's simpler to just remove the whitespace and then either do it again after the next update (which probably won't happen very soon) or remove it in the regeneration scripts (which shouldn't be difficult neither,

You are absolutely right. After downloading the files trailing whitespace can be removed from them with a simple script. I will add that to the download step.

> but Lua is my 17th language

Mine, too (depending on in which order I count them). 😃 

> So if you don't mind, I'll just push this as is (well, after removing whitespace) for now.

Please wait a short time to allow me to add the automatic whitespace removal.

vadz:
I'll merge this in a moment, thanks again!

Issue #23479, opened 2023-04-23
LoadDocument fails if std::istream reaches eof (regression?)


lszl84:
### Description

`wxDocument` framework fails to load the file if std::istream used in `LoadObject` has `eof` flag set.

```cpp
DocumentIstream &DrawingDocument::LoadObject(DocumentIstream &istream)
{
    wxDocument::LoadObject(istream);

    std::string line;
    while (std::getline(istream, line))
    {
        std::cout << line << std::endl;
    }

    return istream;
}
```

`EOF` is ignored for wx streams but not for standard streams for some reason: 

```cpp
// docview.cpp

#if wxUSE_STD_IOSTREAM
    LoadObject(store);
    if ( !store )
#else
    int res = LoadObject(store).GetLastError();
    if ( res != wxSTREAM_NO_ERROR && res != wxSTREAM_EOF )
#endif
    {
        wxLogError(_("Failed to read document from the file \"%s\"."), file);
        return false;
    }
```

It seems that there was a fix for that, but it got lost somehow: https://github.com/wxWidgets/wxWidgets/issues/5308

### Platform and version information

- wxWidgets version  you use: master, 3.2.2
- wxWidgets port  you use: wxOSX, most likely others too

vadz:
Thanks, the fix from that old patch looks indeed correct to me so I'll push it soon.

I do wonder if we shouldn't try harder to preserve compatibility with the existing code not using `wxUSE_STL=1` but using docview though, as the change to the default `wxUSE_STD_IOSTREAM` value is going to break it... Have you been affected by this?

vadz:
No, wait, sorry, I spoke too soon: it doesn't make sense to check for `!stream && !store.eof()` because `!stream` is the same as `stream.fail()` and `fail()` returns false if `eofbit` is the only error bit set. The stream must actually be in _fail_ state, which actually makes sense for the example above because `getline()` sets `failbit` if it doesn't extract anything.

I am not sure what's the right solution here. We could replace `!store` with `store.bad()`, perhaps? For now you can always call `istream.clear()` to make sure the status is correct.

lszl84:
It seems that the behavior of standard streams and wx streams is a bit different when it comes to setting the error bits.

Consider this example, reading from a ZIP using wx streams:

```cpp
    wxFileInputStream in("file.zip");
    wxZipInputStream zipIn(in);
    std::unique_ptr<wxZipEntry> entry(zipIn.GetNextEntry());

[...]

PR #23481, opened 2023-04-23
Restore `wxDocument::{Save,Load}Object()` compatibility with the old non-STL build

vadz:
See #23479.


vadz:
Merged now except for the first commit which was included in this PR accidentally.

Issue #23484, opened 2023-04-24
(git HEAD) gcc 7 build with C++17:  fatal error: charconv: No such file or directory


jolz:
this is my test compile: 
```
git clone --recurse-submodules https://github.com/wxWidgets/wxWidgets
cd wxWidgets

# then compile with gcc 7
cmake -DwxBUILD_CXX_STANDARD=17 -DwxUSE_GUI=0 .. && cmake --build .

# if you currently do not have a gcc-7, you can use an image
podman run -it -v $PWD:$PWD -w$PWD --userns=keep-id conanio/gcc7 sh -c 'rm -rf xx1 ; mkdir xx1 ; cd xx1 && cmake -DwxBUILD_CXX_STANDARD=17 -DwxUSE_GUI=0 .. && cmake --build .'
```

### Description
string.cpp does not compile: 
```
[ 61%] Building CXX object libs/base/CMakeFiles/wxbase.dir/__/__/__/src/common/string.cpp.o
/icpers/hl/work/tecs/hellowx/wxWidgets-master/src/common/string.cpp:1571:10: fatal error: charconv: No such file or directory
 #include <charconv>
          ^~~~~~~~~~
compilation terminated.
libs/base/CMakeFiles/wxbase.dir/build.make:950: recipe for target 'libs/base/CMakeFiles/wxbase.dir/__/__/__/src/common/string.cpp.o' failed
```

the issue sounds like https://stackoverflow.com/questions/49639808/fatal-error-charconv-file-not-found-in-clang-6-0-with-std-c17
- the problem is solved on newer gcc versions (image conanio/gcc8 works)

### Platform and version information

- wxWidgets version master
- wxWidgets port gtk3 (base on linux)
- OS opensuse leap 15.4
- Compiler gcc 7.5.0

jolz:
seems, like the compiler detection may be altered a bit for not-yet-ready compilers: https://github.com/wxWidgets/wxWidgets/blob/master/src/common/string.cpp#L1561

vadz:
I hoped we could avoid configure checks for C++17 compilers, but apparently no such luck :-(

Maybe we could use `__has_include` however, as any C++17 compiler is supposed to have it?

jolz:
created (trivial) pull request as @vadz proposed #23487 
Sorry, I have only tried this with with linux, some gcc/clang versions.  

```bash
# clang 5
[...]

PR #23487, opened 2023-04-24
support gcc7, clang5+6, fixes bug #23484

jolz:
trivial change as proposed by vadz. 

_only tested with some gcc/clang-compiler with linux._


vadz:
Thanks, but I think this can be improved/simplified further, see my follow up commit.

If you don't mind, I'm going to merge this version instead.

PR #23491, opened 2023-04-25
Add `WXK_NUMPAD_CENTER`

### LINKED ISSUE #23478: Numpad "5" key generates different codes in Windows and GTK if NumLock is off
This is wx keyboard sample app logs for pressing `NumPad 5` key with NumLock being off.

On Windows:
```
   Hook           CLEAR   305   ----    0 (U+0000)       12    0x004c0001  (  381,  698)  No
KeyDown           CLEAR   305   ----    0 (U+0000)       12    0x004c0001  (  381,  698)  No
   Char           CLEAR   305   ----    0 (U+0000)       12    0x004c0001  (  381,  698)  No
  KeyUp           CLEAR   305   ----    0 (U+0000)       12    0xc04c0001  (  382,  698)  Yes
```

On GTK:
```
   Hook    NUMPAD_BEGIN   383   ----    0 (U+0000)    65437    0x00000054  (  551,  374)  No
KeyDown    NUMPAD_BEGIN   383   ----    0 (U+0000)    65437    0x00000054  (  528,  358)  No
   Char            HOME   313   ----    0 (U+0000)    65437    0x00000054  (  528,  358)  No
  KeyUp    NUMPAD_BEGIN   383   ----    0 (U+0000)    65437    0x00000054  (  462,  306)  No
```

unxed:


samples/keyboard/keyboard.cpp
```diff
@@ -386,7 +386,9 @@ const char* GetVirtualKeyCodeName(int keycode)
         WXK_(RBUTTON)
         WXK_(CANCEL)
         WXK_(MBUTTON)
-        WXK_(CLEAR)
+#ifdef __WXMSW__
+        WXK_(NUMPAD_CENTER) // = WXK_CLEAR on MSW
+#endif
```
vadz, line 391: Do we really need preprocessor checks here? I.e. why can't we just use `WXK_(NUMPAD_CENTER)` once for all ports?

src/gtk/window.cpp
```diff
@@ -876,7 +876,7 @@ static long wxTranslateKeySymToWXKey(KeySym keysym, bool isChar)
             break;
 
         case GDK_KEY_KP_Begin:
-            key_code = isChar ? WXK_HOME : WXK_NUMPAD_BEGIN;
+            key_code = WXK_NUMPAD_CENTER;
```
vadz, line 879: We actually don't even need this change, AFAICS (nor the one for wxMSW below), as it already works correctly...

vadz (commented):
Thanks, this works fine, but we need to document this in `interface/wx/defs.h` (using `@since 3.3.0` for the constant or `@since 3.2.3` if you plan to backport this to 3.2), could you please add it?

TIA!

src/gtk/window.cpp
unxed: Sorry, don't understand you. Why should we put wxk_home in onChar here?

vadz: What I meant was that we didn't need to change `WXK_NUMPAD_BEGIN` to `WXK_NUMPAD_CENTER` here, as they're one and the same (and same thing with `WXK_CLEAR` in wxMSW).

As for `isChar`, I guess it's never going to be true here as this key code must only be generated for non-char events, but I'm not 100% sure.

unxed: Just in the starting message of this issue I show that onChar of such events have WXK_HOME key code, and this is 100% confusing.

vadz: Sorry, I had missed/forgotten it. Anyhow, I have 0 objections to removing the `isChar` test (except that it might be better to do it in a separate commit), I just was commenting on the fact that if we define `WXK_NUMPAD_CENTER` as a synonym for the existing constant, there is no reason to use the synonym instead of the constant itself in the code.

An alternative -- but, in effect, almost indistinguishable -- approach would be to define `WXK_NUMPAD_CENTER` as a new constant and `WXK_NUMPAD_BEGIN = WXK_NUMPAD_CENTER` for compatibility. And then it would make sense update the code.

But this seems a bit like a mix of the two...

samples/keyboard/keyboard.cpp
unxed: fixed, please review

src/gtk/window.cpp
unxed: fixed, please review

vadz:
Thanks, I think only the documentation is lacking now.

unxed:
Done, please review!

vadz:
Everything looks good, I'll merge this soon with just some minor wording changes. Thanks again!

PR #23492, opened 2023-04-25
Fix typo in grid.h interface file

mcorino:
Replace 'l' by ';' (causes corrupted doxygen output)


vadz:

PR #23494, opened 2023-04-25
Fix Unicode char code in `wxKeyEvent`s for Ctrl+non-Latin-letter events

### LINKED ISSUE #23477: Russian unicode char is irretrievably lost in Ctrl+RussianLetter key events
1. Set two keyboard layouts system wide: English one and Russian one
2. Open Keyboard sample app form wx
3. Switch to Russian layout
4. Press `Ctrl`+`я` (same physical keys as `Ctrl`+`z`)
5. You will see `Ctrl-Z` (26) value in both keycode and unicode key field. So letter `я` is lost forever.

This reproduces in both GTK and Windows backends.

My suggestion: we should put original unicode letter (like `я` in this case) in unicode key field and leave current value in keycode field only.

This can break backward compatibility, but we are breaking it a bit anyway with recent GTK backend keyboard events behavior changes. Also special ctrl+letters parsing is aimed to help in terminal apps development, but the only somewhat popular wx-based terminal app I know is far2l, and we will make necessary adoptions on our side if needed.

unxed:


vadz:
Sorry, but from the test changes it looks like this change the Unicode key for Ctrl+latin-letter events too?

I also think that this paragraph in the docs:

https://github.com/wxWidgets/wxWidgets/blob/8d8f94da641fc95a6d5203276fa2e26abb42e9fc/interface/wx/event.h#L1394-L1403

needs to be updated to mention the handling of non-US/English layouts.

In fact, it would be really great to add a discussion of how non-Latin characters are handled there, e.g. maybe a table explaining the key/Unicode values for the various modifiers?

unxed:
> Sorry, but from the test changes it looks like this change the Unicode key for Ctrl+latin-letter events too?

That's right, this behavior seemed logical to me. If see reasonable to leave the Unicode field values for Latin characters as they are, please tell. I will add the necessary conditions.

vadz:
All this is rather subjective, but the current behaviour doesn't seem illogical to me, so I'd rather keep it. Generally speaking, changing the behaviour of the existing code silently (i.e. without any compilation warnings/errors) is one of the worst things we can do, as library developers, so we really try to avoid doing it unless there is no other reason. And all the incompatible changes in behaviour need to be documented in `docs/changes.txt`.

Anyhow, I'd prefer to keep the existing behaviour unchanged unless there is a very good reason to change it. Thanks!

unxed:
Fixed, please review!

unxed:
Is there anything else that I should do?

vadz:
> Is there anything else that I should do?

No, sorry, I just didn't have time to look at it yet and I'd like to make sure I understand what's going on here before merging it.

src/msw/window.cpp
```diff
@@ -6364,6 +6364,41 @@ wxWindowMSW::CreateCharEvent(wxEventType evType,
...
+        int result = ToUnicode(MapVirtualKey(HIWORD(lParam) & 0xFF, MAPVK_VSC_TO_VK),
+                               HIWORD(lParam) & 0xFF,
+                               keyboardState,
+                               unicodeChar,
+                               1,
+                               0);
+        if (result > 0)
+        {
+            bool ShiftPressed = (GetKeyState(VK_LSHIFT) & 0x8000) != 0 || (GetKeyState(VK_RSHIFT) & 0x8000) != 0;
+            bool capsLockEnabled = (GetKeyState(VK_CAPITAL) & 0x0001) != 0;
```
vadz, line 6381: Couldn't we use [GetKeyboardState()](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getkeyboardstate) to provide the correct keyboard state to `ToUnicode()` and avoid performing these adjustments ourselves? IMO this should be simpler and more robust.

interface/wx/event.h
```diff
@@ -1403,6 +1403,9 @@ enum wxKeyCategoryFlags
     to 31.
     Since version 2.9.2, the enum values @c WXK_CONTROL_A - @c WXK_CONTROL_Z
     can be used instead of the non-descriptive constant values 1-26.
+    Since version 3.3.0, if non-US/English keyboard layout is selected,
+    Unicode field of char events contains original Unicode character
+    as it would be without Control.
```
vadz, line 1408: I think it would be more useful to just document the current behaviour and just say that `GetUnicodeKey()` returns the key code that would be generated by the key press without `Control`, and that the adjustment described here is only done for `GetKeyCode()`.

src/gtk/window.cpp
```diff
@@ -1429,6 +1428,20 @@ gtk_window_key_press_callback( GtkWidget *WXUNUSED(widget),
...
+                    ( uniChar < 'A' || uniChar > 'Z' )) {
+
+                    // for Ctrl+NonLatinLetters use Unicode values also
+                    eventChar.m_uniChar = uniChar;
+
+                } else {
+
+                    // use keycode (1..26) value for Unicode field to mimic MSW behavior
+                    eventChar.m_uniChar = eventChar.m_keyCode;
+                }
```
vadz, line 1444: Would it be possible to keep this inside `AdjustCharEventKeyCodes()`? It's difficult to understand this logic when it's split between two places...

src/msw/window.cpp
```diff
@@ -6364,6 +6364,41 @@ wxWindowMSW::CreateCharEvent(wxEventType evType,
...
+
+            if (( unicodeChar[0] < 'a' || unicodeChar[0] > 'z' ) &&
+                ( unicodeChar[0] < 'A' || unicodeChar[0] > 'Z' )) {
+
+                event.m_uniChar = unicodeChar[0];
+            }
+        }
+    }
+
+    // set keycode to zero for non-ascii characters to match GTK backend behavior
```
vadz, line 6396: Is this a change in the behaviour for wxMSW? If so, it should be described in `docs/changes.txt`.

vadz (commented):
Thanks for your changes, I think they're correct, but it would be great if they could be simplified to make this code more obviously correct and easier to understand because right now -- please let me know what do you think of my suggestions below.

TIA!

interface/wx/event.h
unxed: Updated

src/msw/window.cpp
unxed: Done

vadz:
Is this still relevant after #23818? AFAICS wxGTK now generates `Ctrl-Letter` even for non-US layouts too, so this actually makes wxMSW behave differently from it.

Am I missing something?

vadz:
After looking at this again, I think wxMSW change may still be needed/wanted, but I have trouble understanding what exactly does it change, it would be great if you could please describe it more clearly.

But I don't think wxGTK changes are still needed, are they?

unxed:
As far as I remember (some time has passed, and I may be mistaken in details), Windows generates a special code corresponding to this combination in POSIX terminals when you press Ctrl+letter. Therefore, it is not possible to obtain a Unicode character for such combinations in the standard way. However, some applications require hotkeys like Ctrl+non-English-letter. Therefore, it is preferable to still determine the Unicode character corresponding to the pressed letter key and pass it to the application. This is exactly what the proposed code does.

vadz:
Thanks, but what about wxGTK changes? Should they be dropped?

unxed:
```
    // for Ctrl+letters let's detect corresponding Unicode char
    if (event.m_controlDown && event.m_keyCode <= 26)
    {
```

looks like GTK part solves the same problem with GTK. not sure, time has passed. maybe better to do some testing with Ctrl+letters before and after

vadz:
OK, we need to retest this carefully and document what exactly the changes are. As this doesn't seem very urgent, let's leave it until somebody has time to do it.

unxed:
Probably the good idea is to start documenting kb behavior and it's differences across platforms. Maybe you can suggest good place for such document?

vadz:
I only see `wxKeyEvent` documentation as the place for it. It's already a bit long but maybe we could organize it better, e.g. add a separate section with a table summarizing all cases?

unxed:
Okay! Probably I can start with writing some tests that can disclose some internals of different keyboard  behavior across platforms. Currently all such things are discovered ad-hoc, such approach can not be a stable foundation for development IMHO.

Do you agree?

vadz:
Having tests would be great, but I'm not sure how is this going to work, i.e. I don't think we can use `wxUIActionSimulator` to send layout-dependent key events, can we?

Maybe we could add support for testing this to the keyboard sample, e.g. add a window asking the user to press whichever combination and comparing the received event(s) with the expected one(s)?

In any case, having some kind of test would be great, of course!

unxed:
I think it's worth trying both and seeing what happens! I think I could do this. If you don’t hear from me for a long time, don’t hesitate to remind me, the problem seems important to me.

vadz:
For consistency with `Ctrl+[a-z]` which generate `A-Z` key codes.

PR #23488, opened 2023-04-24
Force D2DRenderTarget's DPI to 96

Kumazuma:
See #23486

PR #23496, opened 2023-04-26
Set DPI to 96 in RenderTargetProperties when creating HwndRenderTarget

Kumazuma:
https://github.com/wxWidgets/wxWidgets/pull/23488

I'm really sorry about poor explanation about this commit.


MaartenBent (approved):
Yes, I think this is the correct thing to do.

PR #23498, opened 2023-04-26
Enable transparent background on Windows

### LINKED ISSUE #23412: What would it take to implement transparency for custom controls?
### Description

Suppose I want to write a library with custom rounded buttons for wxWidgets. Here's a simple idea:

```cpp
#include <wx/wx.h>
#include <wx/dcbuffer.h>

class MyApp : public wxApp
{
public:
    virtual bool OnInit();
};

wxIMPLEMENT_APP(MyApp);

{
public:
    MyFrame(const wxString &title, const wxPoint &pos, const wxSize &size);
};

bool MyApp::OnInit()
{
    MyFrame *frame = new MyFrame("Hello World", wxDefaultPosition, wxDefaultSize);
    frame->Show(true);
    return true;
}

class MyButton : public wxWindow
{
public:
    MyButton(wxWindow *parent, wxWindowID id, const wxPoint &pos, const wxSize &size, long style = 0, const wxString &name = wxPanelNameStr)
        : wxWindow(parent, id, pos, size, style, name)
    {
        this->Bind(wxEVT_PAINT, &MyButton::OnPaint, this);
    }

    void OnPaint(wxPaintEvent &event)
    {
        wxPaintDC dc(this);

        wxColor color = wxColor(100, 100, 200);

        dc.SetPen(wxPen(color, FromDIP(10)));
        dc.SetBrush(wxBrush(color));

        auto rect = this->GetClientRect();
        rect.Deflate(FromDIP(10));
        dc.DrawRoundedRectangle(rect, FromDIP(20));
    }
};

MyFrame::MyFrame(const wxString &title, const wxPoint &pos, const wxSize &size)
    : wxFrame(nullptr, wxID_ANY, title, pos, size)
{

    auto sizer = new wxBoxSizer(wxVERTICAL);

    auto panel = new wxPanel(this);
    panel->SetBackgroundColour({200, 100, 100});

    auto custom = new MyButton(panel, wxID_ANY, wxDefaultPosition, wxDefaultSize);

    auto panelSizer = new wxBoxSizer(wxVERTICAL);
    panelSizer->Add(custom, 1, wxEXPAND | wxALL, FromDIP(50));
    panel->SetSizer(panelSizer);

    sizer->Add(panel, 1, wxEXPAND);
    this->SetSizer(sizer);
}
```

Works well on GTK or Mac but not on Windows. Experimenting with clearing the background with a transparent brush yields mixed results (weird paint artifacts).

Setting the background to the parent's color might be a workaround, but what if the user wants to place my custom button on top of their custom bitmap control? As the library author, I don't have access to the user's controls, so I cannot render its part in my `wxPaintDC.`

TL;DR, it seems such a basic thing as creating a custom rounded rect button is not possible on wxWidgets as of now. What would it take to implement transparency and compositing on Windows, so it can be done?

Screenshot 2023-04-01 at 13 25 23
Screenshot 2023-04-01 at 13 24 34
Screenshot 2023-04-01 at 13 24 09

MaartenBent:
When the `wxWindow` background style is set to `wxBG_STYLE_TRANSPARENT`, set the `WS_EX_TRANSPARENT` style flag.

Closes #23412

### DIFF
--- a/include/wx/msw/window.h
+++ b/include/wx/msw/window.h
@@ -115,6 +115,8 @@ public:
     virtual bool SetCursor( const wxCursor &cursor ) override;
     virtual bool SetFont( const wxFont &font ) override;
 
+    virtual bool IsTransparentBackgroundSupported(wxString* reason = nullptr) const override;
+
     virtual int GetCharHeight() const override;
     virtual int GetCharWidth() const override;
 
--- a/interface/wx/window.h
+++ b/interface/wx/window.h
@@ -2420,12 +2420,12 @@ public:
         problem.
 
 
-        Under wxGTK and wxOSX, you can use ::wxBG_STYLE_TRANSPARENT to obtain
+        Under wxGTK, wxOSX and wxMSW, you can use ::wxBG_STYLE_TRANSPARENT to obtain
         full transparency of the window background. Note that wxGTK supports
         this only since GTK 2.12 with a compositing manager enabled, call
         IsTransparentBackgroundSupported() to check whether this is the case,
         see the example of doing it in @ref page_samples_shaped "the shaped
-        sample".
+        sample". Under wxMSW this is supported since 3.3.0.
 
         Also, in order for @c SetBackgroundStyle(wxBG_STYLE_TRANSPARENT) to
         work, it must be called before Create(). If you're using your own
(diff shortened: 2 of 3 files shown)


vadz:
Thanks! I'd like to test this is a bit before apply, but perhaps you already did and can tell if it works only when compositing is on (in which case we should mention this in the docs too) or if it somehow manages to work without it too?

vadz:
Sorry, I've realized that this was already discussed/tested in #23412 and my own simple test confirms that it works both with and without compositing, so I've just merged this as is.

Thanks again!

PR #23504, opened 2023-04-28
Explain how to specifiy download paths before starting a wxWebRequest

### LINKED ISSUE #22986: User-defined download path for wxWebRequest/wxWebRequestEvent
Would it be possible to add a field for download path to @c wxWebRequest that would be passed to @c wxWebRequestEvent?

Basically, I set the storage to @c Storage_File and connect it to an event.  In the event, I can process the download and copy the file from the event's **temp file path**; however, this requires the user to specify the download path interactively at this point (refer to the webrequest sample). It would be nice if you could provide this path in the original @c wxWebRequest object and get it from the @c wxWebRequestEvent when it is handles later.  This way you could queue numerous downloads and not keep asking the user where they should be save to.

I can try to make a PR myself, but I want to make sure there aren't any issues with doing this.  Thanks!

Blake-Madden:
Adding an overview paragraph explaining how to use wxWebRequest's and wxWebEvent's IDs together to specify a request's download path before starting it. The sample shows how to interactively prompt for local paths once a file downloads, but I was wanting to download a file to a specific location without user prompting. The technique described here worked for me.

If accepted, this could close #22986.

### DIFF
--- a/interface/wx/webrequest.h
+++ b/interface/wx/webrequest.h
@@ -63,6 +63,15 @@
         // Start the request
         request.Start();
     @endcode
+    
+    The location of where files are downloaded can also be defined prior to any request
+    by passing unique IDs to `wxWebSession::GetDefault().CreateRequest()` and processing
+    them in your @c wxEVT_WEBREQUEST_STATE handler. For example, create a map of IDs with
+    their respective download paths prior to creating any requests. For each call to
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for the explanations!

@TcT2k Could you please confirm that this is the right way to do it? It seems reasonable (even if a bit roundabout) to me, but I'd appreciate your opinion. TIA!

vadz:
I'm finally going to just assume this is correct and merge it.

PR #23508, opened 2023-04-29
Remove mentions of no longer supported MSVS versions

PBfordev:
The oldest supported MSVS version for wxWidgets 3.3 is 2015, so remove mentions of limitations of older MSVS versions.


vadz:
Thanks for the update, I will merge it soon.

PR #23510, opened 2023-04-29
Fix updating wxFrame client size in wxEVT_DPI_CHANGED handler

vadz:
This didn't work correctly when switching from higher to lower DPI if the frame had a status bar because this status bar still had its old, bigger height, by the time SetClientSize() was called from the event handler, resulting in the frame size becoming too big.

It notably resulted in having an unwanted gap between the frame contents and the status bar if the event handler called GetSizer()->Fit().

Fix this by calling PositionStatusBar(), which also updates the status bar height to correspond to the new DPI as a side effect, from a new virtual MSWBeforeDPIChangedEvent() function which had to be added to allow wxFrame to customize DPI handling.

Adding this new virtual function just for this isn't great, but the only alternatives seem even worse.

----

@MaartenBent Please let me know if you have any comments/better ideas about how to do it. The TL;DR summary of the commit message above is that we need to somehow call `PositionStatusBar()` before processing `wxDPIChangedEvent` generated in `MSWUpdateOnDPIChange()`.

### DIFF
--- a/include/wx/event.h
+++ b/include/wx/event.h
@@ -3129,7 +3129,12 @@ public:
 
     // Scale the value by the ratio between new and old DPIs carried by this
     // event.
+    wxPoint Scale(wxPoint pt) const;
     wxSize Scale(wxSize sz) const;
+    wxRect Scale(wxRect r) const
+    {
+        return wxRect(Scale(r.GetPosition()), Scale(r.GetSize()));
+    }
 
     int ScaleX(int x) const { return Scale(wxSize(x, -1)).x; }
     int ScaleY(int y) const { return Scale(wxSize(-1, y)).y; }
--- a/include/wx/msw/frame.h
+++ b/include/wx/msw/frame.h
@@ -131,6 +131,8 @@ protected:
     virtual void DoGetClientSize(int *width, int *height) const override;
     virtual void DoSetClientSize(int width, int height) override;
 
+    virtual void MSWBeforeDPIChangedEvent(const wxSize& newDPI) override;
+
 #if wxUSE_MENUS_NATIVE
     // perform MSW-specific action when menubar is changed
     virtual void AttachMenuBar(wxMenuBar *menubar) override;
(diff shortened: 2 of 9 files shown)


vadz:
We definitely need such function because it's also needed in a completely different place (see the last commit in the updated PR), but I'm still not sure about its name and parameter(s): passing it `wxDPIChangedEvent` is convenient but somewhat unusual.

Perhaps we need some `wxDPIChange` class that would just contain the old and new DPI values and provide `Scale()` and then inherit `wxDPIChangedEvent` from it and pass this class to that function? It would be slightly cleaner but it's not clear if it's really worth the trouble...

If nobody expresses any other opinions, I'm just going to merge this soon, as this function is private, we can always change it later, but in the meanwhile adding it fixes a couple of real problems.

MaartenBent:
Sorry for the late reply. I tried some samples with a status bar, but I don't think I see the problem that you described.
But if you indeed need to update the size of these gui elements before the size calculations start, this seems like a correct solution.

I'm only wondering if it would have been possible to do this in `MSWUpdateFontOnDPIChange` if it knew the old DPI.

vadz:
> Sorry for the late reply. I tried some samples with a status bar, but I don't think I see the problem that you described.

You need to have a frame with something entirely filling it and
```cpp
	Bind(wxEVT_DPI_CHANGED, [this](wxDPIChangedEvent& event) {
		m_sizer->Fit(this);
	});
```
to see it. Without this change, the frame size became wrong (too big) when moving from high to low DPI monitor because `Fit()` ended up calling `SetClientSize()` which transformed the size to the full size using the old (bigger) status bar height. The status bar height was then updated, but it was too late and a gap equal to the different in heights before and after remained between the frame contents and the status bar.

> I'm only wondering if it would have been possible to do this in `MSWUpdateFontOnDPIChange` if it knew the old DPI.

I thought about repurposing this function for this, but it didn't seem right to do something not related to the font at all in it. We could rename it, of course, but having this function seems convenient in a few places where it's currently used, so I've finally decided to add a new one. I'm still not sure if it was the correct decision, but OTOH I'm not sure it was a wrong one neither :-/

PR #23511, opened 2023-05-01
Fix wxTextCtrl hint showing for wxUniv

Kvaz1r:
GetValue use text stored in wxTextEntryHintData so when wxTextCtrl get focus it wasn't remove hint as expected.

### DIFF
--- a/src/univ/textctrl.cpp
+++ b/src/univ/textctrl.cpp
@@ -783,7 +783,7 @@ wxTextCtrl::~wxTextCtrl()
 
 void wxTextCtrl::DoSetValue(const wxString& value, int flags)
 {
-    if ( value != GetValue() )
+    if ( value != DoGetValue() )
     {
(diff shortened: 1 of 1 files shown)


vadz:
I think the actual problem here is that `GetValue()` returns the hint text instead of empty string. Could you perhaps consider fixing this instead?

Kvaz1r:
> I think the actual problem here is that `GetValue()` returns the hint text instead of empty string. Could you perhaps consider fixing this instead?

No, issue with hint opposite, GetValue() returns exactly empty string with no hint consideration as should, but DoSetValue get empty string from [wxTextEntryHintData::OnSetFocus](https://github.com/wxWidgets/wxWidgets/blob/master/src/common/textentrycmn.cpp#L134-L141) especially to remove hint appearance. But because GetValue also returns empty string, hint remains showed because m_value isn't updated.

vadz:
Ah, I see, thanks for the explanation. Will merge soon.

PR #23520, opened 2023-05-04
Support wxDP_DROPDOWN mode for native date pickers on macOS

lanurmi:
According to Apple's documentation, the presentsCalendarOverlay property is supported in macOS 10.15.4 and later.

This should be safe to cherry-pick to 3.2, too.

### DIFF
--- a/src/osx/cocoa/datetimectrl.mm
+++ b/src/osx/cocoa/datetimectrl.mm
@@ -26,6 +26,7 @@
 
 #include "wx/osx/core/private/datetimectrl.h"
 #include "wx/osx/cocoa/private/date.h"
+#include "wx/osx/private/available.h"
 
 using namespace wxOSXImpl;
 
@@ -164,7 +165,7 @@ wxDateTimeWidgetImpl::CreateDateTimePicker(wxDateTimePickerCtrl* wxpeer,
                                            const wxDateTime& dt,
                                            const wxPoint& pos,
                                            const wxSize& size,
-                                           long WXUNUSED(style),
+                                           long style,
                                            wxDateTimeWidgetKind kind)
 {
     NSRect r = wxOSXGetFrameForControl(wxpeer, pos, size);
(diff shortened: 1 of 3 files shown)


src/osx/cocoa/datetimectrl.mm
```diff
@@ -187,6 +187,14 @@ long WXUNUSED(style),
 
     [v setDatePickerStyle: NSTextFieldAndStepperDatePickerStyle];
 
+    if ( style & wxDP_DROPDOWN )
+    {
+        if ( @available(macOS 10.15.4, *) )
```
vadz, line 192: I think there was a reason we used the macros from `include/wx/osx/private/available.h` instead of using this directly but I don't remember what it was...

If there really isn't any any more, we should stop using these macros everywhere, but otherwise we should use add some new `WX_IS_MACOS_AVAILABLE_FULL` and use `WX_IS_MACOS_AVAILABLE_FULL(10, 15, 4)` here too for consistency.

vadz (commented):
Thanks, I didn't test this, but we should apply it, of course, if it's supported now.

I just have a question/concern about the direct use of `@available` (cc @csomor).

src/osx/cocoa/datetimectrl.mm
csomor: @vadz I think we need another round of macros
- the native options are Clang only
- in c/c++ we'd use `__builtin_available` natively (this would work because it's available in Xcode 7+)
- only in objective-c or objective-c++ does the `@available` exist
(and only in Xcode 9+ which is newer than our 3.2 minimal Xcode requirement 7.2.1 on 10.10.5)

vadz: Ah, that's right, thanks, I now remember that we've introduced these macros because `@available` is not available (sic) in pure C++ code and we wanted to use the same thing in both C++ and Objective-C++.

So yes, I think we should add another macro. @lanurmi If you could please do it, it would be great, TIA!

lanurmi: Done. Should have guessed there's a reason why I never recall seeing `@available` in wx code. I also realized that an `#if` check for the SDK version is required to avoid breaking compilation with older SDKs.

vadz:
Great, thanks, will push in a moment.

PR #23525, opened 2023-05-05
Multiple print page ranges

sz5000:
With this change you can now define multiple page ranges in the print dialog, which are then printed. Previously, only one page range was possible. I have tested this change with the PDF24 Reader and there it works fine.

image


include/wx/cmndata.h
```diff
@@ -145,6 +145,24 @@ class WXDLLIMPEXP_CORE wxPrintData: public wxObject
...
+ * Contains a range of pages to be printed.
+ */
+
+class WXDLLIMPEXP_CORE wxPrintPageRange
+{
+public:
+    int fromPage = 0;
+    int toPage = 0;
+
+    wxPrintPageRange() : fromPage(0), toPage(0) { }
```
vadz, line 159: This is definitely not needed, we should just use
```suggestion
    wxPrintPageRange() = default;
```

include/wx/cmndata.h
```diff
@@ -145,6 +145,24 @@ class WXDLLIMPEXP_CORE wxPrintData: public wxObject
...
+    wxPrintPageRange(int _fromPage, int _toPage) : fromPage(_fromPage), toPage(_toPage) { }
```
vadz, line 160: Arguably this is not needed neither, we should be able to initialize this class just by using `{from, to}` as it's a simple aggregate, so I'd consider removing both ctors.

include/wx/cmndata.h
```diff
@@ -222,6 +244,9 @@ class WXDLLIMPEXP_CORE wxPrintDialogData: public wxObject
     bool            m_printEnablePrintToFile = true;
     wxPrintData     m_printData;
 
+    int m_maxPageRanges = 64;
+    wxVector<wxPrintPageRange> m_printPageRanges;
```
vadz, line 248: Let's use the standard class directly in the new code:
```suggestion
    std::vector<wxPrintPageRange> m_printPageRanges;
```

include/wx/cmndata.h
```diff
@@ -222,6 +244,9 @@ class WXDLLIMPEXP_CORE wxPrintDialogData: public wxObject
     bool            m_printEnablePrintToFile = true;
     wxPrintData     m_printData;
 
+    int m_maxPageRanges = 64;
```
vadz, line 247: It would be nice to have a comment here explaining where does this value come from and, more generally, why do we need this variable (and the associated functions) at all.

include/wx/cmndata.h
```diff
@@ -201,6 +219,10 @@ class WXDLLIMPEXP_CORE wxPrintDialogData: public wxObject
     wxPrintData& GetPrintData() { return m_printData; }
     void SetPrintData(const wxPrintData& printData) { m_printData = printData; }
 
+    wxVector<wxPrintPageRange>& GetPageRanges() { return m_printPageRanges; }
+    void SetMaxPageRanges(int maxPageRanges) { m_maxPageRanges = maxPageRanges; }
+    int GetMaxPageRanges() { return m_maxPageRanges; }
```
vadz, line 224:
```suggestion
    int GetMaxPageRanges() const { return m_maxPageRanges; }
```

src/msw/printwin.cpp
```diff
@@ -207,9 +232,25 @@ bool wxWindowsPrinter::Print(wxWindow *parent, wxPrintout *printout, bool prompt
...
+                        inRanges = true;
+                        break;
+                    }
+                }
+
+                if (!inRanges)
+                    continue;
+            }
+
+            printCnt += 1;
```
vadz, line 252: A comment explaining why do we do this would be helpful as it's not immediately obvious.

src/msw/printwin.cpp
```diff
@@ -163,16 +163,40 @@ bool wxWindowsPrinter::Print(wxWindow *parent, wxPrintout *printout, bool prompt
...
+            minPageNum = fromPage;
+            maxPageNum = toPage;
+        }
+        totalPageCnt = maxPageNum - minPageNum + 1;
     }
     else if ( !m_printDialogData.GetAllPages() )
     {
-        minPageNum = m_printDialogData.GetFromPage();
-        maxPageNum = m_printDialogData.GetToPage();
+        if (m_printDialogData.GetFromPage() > 0 && m_printDialogData.GetToPage() > 0)
```
vadz, line 179: It's not really clear how does `Get{From,To}Page()` and `GetPageRanges()` work together. Can we have both of them valid at the same time? It looks like the former should be just a special case of the latter, with a single range, could we do it like this or is there something that prevents us from doing it?

src/msw/printdlg.cpp
```diff
@@ -964,6 +984,14 @@ bool wxWindowsPrintDialog::ConvertFromNative( wxPrintDialogData &data )
     {
         data.SetFromPage(pd->lpPageRanges[0].nFromPage);
         data.SetToPage(pd->lpPageRanges[0].nToPage);
+
+        wxVector<wxPrintPageRange>& ranges = data.GetPageRanges();
+        ranges.clear();
+        for (DWORD i = 0; i < pd->nPageRanges; i++) {
+            int fromPage = pd->lpPageRanges[i].nFromPage;
+            int toPage = pd->lpPageRanges[i].nToPage;
+            ranges.push_back(wxPrintPageRange(fromPage, toPage));
```
vadz, line 993: If we remove ctor as I suggested, we could just write
```suggestion
            ranges.emplace_back(fromPage, toPage);
```
here.

src/msw/printdlg.cpp
```diff
@@ -964,6 +984,14 @@ bool wxWindowsPrintDialog::ConvertFromNative( wxPrintDialogData &data )
     {
         data.SetFromPage(pd->lpPageRanges[0].nFromPage);
         data.SetToPage(pd->lpPageRanges[0].nToPage);
+
+        wxVector<wxPrintPageRange>& ranges = data.GetPageRanges();
+        ranges.clear();
+        for (DWORD i = 0; i < pd->nPageRanges; i++) {
```
vadz, line 990: Minor style nitpick:
```suggestion
        for (DWORD i = 0; i < pd->nPageRanges; i++)
        {
```

src/msw/printdlg.cpp
```diff
@@ -964,6 +984,14 @@ bool wxWindowsPrintDialog::ConvertFromNative( wxPrintDialogData &data )
     {
         data.SetFromPage(pd->lpPageRanges[0].nFromPage);
         data.SetToPage(pd->lpPageRanges[0].nToPage);
+
+        wxVector<wxPrintPageRange>& ranges = data.GetPageRanges();
+        ranges.clear();
```
vadz, line 989: This is a bit of a weird pattern, shouldn't we just have `SetPageRanges()` instead?

vadz (commented):
Thanks for your work on this! This is clearly useful and it will be nice to integrate it, but in addition to all the minor comments below, I have 2 more global questions:

1. Could we handle a single from/to case as just having a single range instead of handling it specially/separately from handling ranges?
2. Why do we need the "max ranges" stuff? It looks like we could just the number of ranges that we have, but I may be missing something here.

Thanks again!

include/wx/cmndata.h
sz5000: I actially like ctors like `wxPrintPageRange(int _fromPage, int _toPage)` because it's easier for programming. Programmers can see more easily in the IDE how to create the object. If you do it with `{from, to}`, you always have to check the order of the elements beforehand and this is more error-prone.

include/wx/cmndata.h
sz5000: with m_maxPageRanges  you can set the maximum number of ranges allowed. This values is used for the internal array that is required to hold the ranges. When the print dialog is opened, the user can then define up to m_maxPageRanges  ranges because the internal array can only hold this amount of ranges.

src/msw/printwin.cpp
sz5000: Can be used together and is there because this is backward compatible. The ranges array is currently just an extension. If you want to use more than one range, you have to use the ranges array.

src/msw/printdlg.cpp
sz5000: You could, but actually it is not necessary to create a new vector beforehand....

include/wx/cmndata.h
vadz: Note that you can, of course, write `wxPrintPageRange{from, to}` explicitly too.

`wxPoint` was written for pre-C++98 compilers, things were a bit different back then...

include/wx/cmndata.h
vadz: Ah, I see, thanks for the explanation.

But I believe that the default value of 64 should really be fine, at least I can't imagine anybody needing more ranges than this in any kind of realistic scenario. So my first instinct would be to just leave this field and these constants off and use 64 in wxMSW code.

Of course, if you really want to have them (in order to increase this even further, I guess?), we can keep them, but in this case we should explain in the docs that `SetMaxPageRanges()` can be called to increase the default limit and mention what it is. So it would be just simpler to not do this and not have all this at all.

src/msw/printdlg.cpp
vadz: Is this an efficiency concern? Because I really don't think temporarily allocating extra 1KiB of memory is a problem.

src/msw/printwin.cpp
vadz: `m_printFromPage` and `m_printToPage` are private, so it looks like we could get rid of them and just use the first element of `m_printPageRanges` to implement `{Get,Set}{From,To}Page()` instead, couldn't we?

`GetPageInfo()` is incompatible with using multiple ranges, so we need to extend the API and add some `vector<wxPrintPageRange> GetPageRanges() const` which would be called by wx. The default implementation would call `GetPageInfo()` for compatibility, but it could be overridden to return richer information, with multiple ranges.

This is not going to be that simple, but AFAICS this problem can be solved and I'm almost sure that it has to be solved to allow the application to specify multiple range of pages to print as this still seems to be unsupported with the current version of the PR.

But maybe we need a deeper rethink of how this stuff works because it seems suspicious to have similar but not at all the same information in `wxPrintDialogData` and `GetPageInfo()`... Maybe we should just pass the former to the latter and allow it to modify it? Sorry, I'm running out of time here :-(

include/wx/cmndata.h
sz5000: I think that eventually someone might want to make the number of regions smaller rather than larger.

src/msw/printwin.cpp
sz5000: I have revised the code and made the changes you suggested. Please have a look. I've also added some more comments to make some things a little more clear.

src/msw/printdlg.cpp
sz5000: I'm thinking more of flexibility then on efficiency ..

include/wx/cmndata.h
vadz: This is what I don't understand: why would anybody want to make this smaller or restrict it at all? I.e. I understand that we need to specify some number of ranges in advance because this is how Win32 API works and I can accept that, in theory, any number we specify might be too small for somebody (even though in practice 64 should be more than enough), but what would be the motivation for restricting the number of ranges?

I think the only cases we need handle are:

1. Not accepting ranges at all.
2. Accepting exactly one range.
3. Accepting any number of them.

Am I missing something?

src/msw/printdlg.cpp
vadz: Well, using (const) `Get` + `Set` is exactly as flexible, as you just can make a copy of the vector and then it's the same as using the current (non-const) `Get`. The only difference is efficiency (but I don't think it matters here) and consistency with the rest of the API -- which I think does matter and wx API almost universally uses const `Get` and `Set` instead of allowing to modify things in place (there are some exceptions, of course...).

include/wx/cmndata.h
```diff
@@ -178,8 +178,14 @@ class WXDLLIMPEXP_CORE wxPrintDialogData: public wxObject
...
     virtual ~wxPrintDialogData();
 
-    int GetFromPage() const { return m_printFromPage; }
-    int GetToPage() const { return m_printToPage; }
+    // Retained for backward compatibility.
+    // Returns the from page of the first range or 0 if there is no page range.
+    int GetFromPage() const;
+
+    // Retained for backward compatibility.
+    // Returns the to page of the first range or 0 if there is no page range.
```
vadz, line 186: This is probably not very important in practice, but perhaps we should return `to` of the _last_ range here? So that if the user selects `1-2,8-9` in the dialog, an old application not updated to use the new API yet would print the pages `1-9` rather than `1-2`? Neither is ideal, of course, but the former is perhaps slightly better?

src/common/prntbase.cpp
```diff
@@ -626,15 +626,30 @@ bool wxPrintout::HasPage(int page)
 
 void wxPrintout::GetPageInfo(int *minPage, int *maxPage, int *fromPage, int *toPage)
 {
-    *minPage = 1;
-    *maxPage = DEFAULT_MAX_PAGES;
-    *fromPage = 1;
-    *toPage = 1;
+    if(minPage)
```
vadz, line 629: I don't think we need those, the parameters are guaranteed to be non-null here AFAICS.

interface/wx/print.h
```diff
@@ -705,23 +705,39 @@ class wxPrintout : public wxObject
...
-        @a pageTo should be both set to the current page number.
-
         By default this returns (1, 32000) for the page minimum and maximum values, and
         (1, 1) for the required page range.
 
         @a minPage must be greater than zero and @a maxPage must be greater
         than @a minPage.
+
+        @deprecated
+        The newer method that support multiple ranges should be used instead.
```
vadz, line 715: This is not quite right, as this function is not formally deprecated. I'd probably say something like

```suggestion
        This function allows to indicate only a single range of pages to print,
        consider overriding GetPageInfoRanges() to return multiple ranges.
```

interface/wx/print.h
```diff
@@ -705,23 +705,39 @@ class wxPrintout : public wxObject
...
+
+        By default this returns (1, 32000) for the page minimum and maximum values, and
+        (1, 1) for the required page range.
+
+        @a minPage must be greater than zero and @a maxPage must be greater
+        than @a minPage.
+
+        @since 3.3.0.
+    */
+    virtual void GetPageInfo(int* minPage, int* maxPage, wxVector<wxPrintPageRange>* pageRanges);
```
vadz, line 739: The function name is wrong here:
```suggestion
    virtual void GetPageInfoRanges(int* minPage, int* maxPage, wxVector<wxPrintPageRange>* pageRanges);
```

interface/wx/print.h
```diff
@@ -705,23 +705,39 @@ class wxPrintout : public wxObject
...
+
+        If the user chose to print the current page, then @a pageRanges should contain one range
+        and the range values are both set to the current page number.
+
+        By default this returns (1, 32000) for the page minimum and maximum values, and
+        (1, 1) for the required page range.
+
+        @a minPage must be greater than zero and @a maxPage must be greater
+        than @a minPage.
+
```
vadz, line 736: We need to mention that the default implementation forwards to `GetPageInfo()` somewhere.

include/wx/prntbase.h
```diff
@@ -284,8 +284,11 @@ class WXDLLIMPEXP_CORE wxPrintout: public wxObject
 
     virtual bool HasPage(int page);
     virtual bool OnPrintPage(int page) = 0;
+
     virtual void GetPageInfo(int *minPage, int *maxPage, int *pageFrom, int *pageTo);
-    virtual bool IsPageSelected(int page);
+
+    // the default implementation of this new method uses the older GetPageInfo method above
+    virtual void GetPageInfoRanges(int* minPage, int* maxPage, std::vector<wxPrintPageRange>* pageRanges);
```
vadz, line 291: I think we should return the ranges from here instead of using an output pointer, this would be cleaner:
```suggestion
    virtual void std::vector<wxPrintPageRange> GetPageInfoRanges(int* minPage, int* maxPage);
```

vadz (commented):
Thanks for the updates, this looks much better now, IMO, but there are still a few questions/problems:

1. We need to replace all the other calls to `GetPageInfo()` inside the library (i.e. in wxGTK and wxOSX) with the calls to `GetPageInfoRanges()`, as otherwise the printing sample (and any code copying what it does) would be broken there.
2. I'm not sure what has happened to `IsPageSelected()` here, did you decide to handle the selected page as a single page range instead? I don't mind if this is the case, but I just wonder why hadn't it been done like this this since the beginning, even with just a single range support?
3. I wonder if we may need some `SetEnableMultiplePageRanges()` to explicitly opt-in using them, as allowing to enter multiple ranges in the dialog might not be a great idea if the program doesn't handle this -- i.e. right now, all the existing programs would just ignore all the ranges after the first one, as all of them use `Get{From,To}Page()`. So maybe we need to disable multiple ranges by default, but allow enabling them explicitly, what do you think?

include/wx/cmndata.h
sz5000: Yes, and `SetMaxPageRanges`  covers all three cases you mentioned and additionally a limit number of ranges, depending on the app that uses this code. 

I think we shouldn't limit the space of possibilities beforehand just because we can't currently imagine that there are other cases besides the 3 cases that you mentioned.

I would therefore strongly recommend letting the developer decide how many page ranges the user of an app can enter.

vadz: Sorry, I disagree here. We shouldn't be adding API to do something that we have no realistic use case for. It can always be added later if needed, but removing it is much harder.

So if there is really no concrete scenario in which these functions are useful, let's remove them for now.

sz5000: You mentioned to add the method `SetEnableMultiplePageRanges` to limit the number of ranges to one. I would not do that, I would use `SetMaxPageRanges` because SetEnableMultiplePageRanges is just a sub-case for SetMaxPageRanges. 

We could even limit the default number of ranges to one and then developer can increase this number with  `SetMaxPageRanges` to allow multiple ranges.  So the developer has more flexibility if we have `SetMaxPageRanges`.

vadz: The problem is that using `SetMaxPageRanges()` requires the application programmer to arbitrarily select the maximum number of supported ranges. This makes the API more difficult to use and increases the possibility of someone making a wrong choice and passing 10,000 or something like that to it, which would certainly be a bad idea. I'd like to avoid people having to make this arbitrary choice and just allowing them to indicate that they want (and are ready) to support multiple page ranges, without specifying how many of them exactly do they need because there is just no good answer to this question in general.

include/wx/cmndata.h
sz5000: I tried without ctors but had build issues on mac, so switched back to this version. You can decide vadz, I'm fine with either version.

interface/wx/print.h
sz5000: is now available

include/wx/cmndata.h
sz5000: Of course, we could, if it is better, you can decide..

sz5000:
@vadz How far are we from merging the changes into masters?

vadz:
Sorry, I didn't have time to return to this yet and look at the latest changes. Do they answer my questions above?

FWIW I definitely want to have this in 3.3.0, but it won't happen before September.

vadz:
I'm sorry for still not getting back to it, I'll try to do it soon (before the end of the year), but I'd be grateful if others could please look at and, most importantly, test this PR too.

vadz:
> Hi vadz, do you know when you have time to add this feature?

Sorry, I really hoped to get some testing for it, but it's clearly not going to happen, so I'll make another pass on this to review it and will merge it soon.

Sorry again for the delay.

vadz:
Sorry once again for taking so long to return to this!

I finally did it and quickly realized that this was incomplete: if we want to replace `GetPageInfo()` with a new function we must call this new function in all ports, not just wxMSW, as otherwise the code using it wouldn't work at all elsewhere. So I've started fixing this and as long as I was changing things anyhow, I've also renamed it to a better (IMHO) name and did a few other minor changes as well as updated wxGTK to support multiple page ranges as well.

The result is in #25030 and it would be great if you could please test it — in my tests it does work, under both wxMSW and wxGTK, but I could have missed something. TIA!

BTW, here is the diff to the printing sample which I used to allow using more ranges:

```diff
diff --git a/samples/printing/printing.cpp b/samples/printing/printing.cpp
index 41c2d8b487..117f45dd06 100644
--- a/samples/printing/printing.cpp
+++ b/samples/printing/printing.cpp
@@ -353,7 +353,7 @@ void MyFrame::OnPrint(wxCommandEvent& WXUNUSED(event))
     printDialogData.EnablePageNumbers(true);
     printDialogData.EnableCurrentPage(true);
     printDialogData.SetMinPage(1);
-    printDialogData.SetMaxPage(2);
+    printDialogData.SetMaxPage(20);
     printDialogData.SetFromPage(1);
     printDialogData.SetToPage(2);
     printDialogData.SetAllPages(true);
@@ -560,12 +560,12 @@ wxPrintPageRange MyPrintout::GetPagesInfo(wxPrintPageRanges& ranges)
     //      printed, depending on the ranges contents on entry to this function
 
     // This is the total range of pages that can be printed.
-    return { 1, 2 };
+    return { 1, 20 };
 }
 
 bool MyPrintout::HasPage(int pageNum)
 {
-    return (pageNum == 1 || pageNum == 2);
+    return (pageNum >= 1 && pageNum <= 20);
 }
 
 void MyPrintout::DrawPageOne()
```

vadz:
Oh, and I forgot to say that I'm closing this PR but only because it's part of another one now.

PR #23528, opened 2023-05-06
Fix regression in destroying wxArtProvider objects

### LINKED ISSUE #23513: wxUniv/wxMSW Crash "access violation at address" on exit application
### Description
Crash "access violation at address" on exit application

Debug shows that error happen at [wxArtProvider::CleanUpProviders](https://github.com/wxWidgets/wxWidgets/blob/master/src/common/artprov.cpp#L318) on deleting sm_providers.
Reason - attempt to delete artprovider that was already deleted in [theme destructor](https://github.com/wxWidgets/wxWidgets/blob/master/src/univ/themes/gtk.cpp#L539).
I'm not sure should it been replaced with:
`wxArtProvider::Remove(m_artProvider);`
or just removed after std::unique_ptr<> was introduced in https://github.com/wxWidgets/wxWidgets/commit/4a294caacd73a2825a01e2a2a41fbfc719761d89 

- wxWidgets version: current master
- wxWidgets port: wxUniv/wxMSW
- OS: Windows 10

vadz:
Don't delete the art provider object in Remove() as this contradicted its contract and do restore the call to Remove() from wxArtProvider dtor as it's actually documented to work.

This fixes a regression due to the changes of 4a294caacd (Stop using macro-based wxList in wxArtProvider code, 2023-04-11).

Closes #23513.

### DIFF
--- a/src/common/artprov.cpp
+++ b/src/common/artprov.cpp
@@ -242,7 +242,11 @@ wxArtProviderCache *wxArtProvider::sm_cache = nullptr;
 // wxArtProvider ctors/dtor
 // ----------------------------------------------------------------------------
 
-wxArtProvider::~wxArtProvider() = default;
+wxArtProvider::~wxArtProvider()
+{
+    if ( sm_providers )
+        Remove(this);
+}
 
 // ----------------------------------------------------------------------------
 // wxArtProvider operations on provider stack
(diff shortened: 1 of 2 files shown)


vadz:
I made a typo in the issue number in 4d48e47811728503a043549ea2ab48640a210809 which was supposed to close this one and not #23529 as it corresponds to the commit here.

PR #23530, opened 2023-05-06
Update wx_setup.props to be more comprehensive

MasonEdgar:
I have descriptive names for my different build configurations in VS2022, and the original `wx_setup.props` wouldn't recognize them.  Proposed solution is based on this forum post: https://forums.wxwidgets.org/viewtopic.php?t=47314

### DIFF
--- a/build/msw/wx_setup.props
+++ b/build/msw/wx_setup.props
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <_PropertySheetDisplayName>wxWidgets Setup Options</_PropertySheetDisplayName>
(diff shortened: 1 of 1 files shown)


vadz (commented):
Thanks, generally speaking this looks like a good idea, but I wonder what happens when several conditions match. Do they get applied in order? E.g. what if configuration is something like "Release with debug info", what would `wxSuffixDebug` value be in this case?

MasonEdgar:
That's a good point. I'm not well versed on `MSBuild` conditionals, so I'm sure there's room for improvement from what I have.

vadz:
It would be great if you could please check how it behaves with configuration names like this. If it's reasonable, we can apply the PR in its current state, but it would be nice to at least know what it is. TIA!

MasonEdgar:
I'll do some more research and let you know what I come up with. I'll update this PR with any new info

dconnet:
I'm pretty sure names are exact matches - and you can only match one. And when I originally set up the props file, 'Release' and 'Debug' are static (which is why there are 'Release DLL' and 'Debug DLL')

vadz:

vadz:
I still think this could be useful but I can't apply this without making sure it doesn't break the existing projects, so I have no choice but to close it, sorry.

PR #23535, opened 2023-05-09
Non-wx thread memory leak

nietsu:
Using certain parts of wx from a thread other than wxThread results in a memory leak.

For example, using wxSemaphore causes wxLogTrace calls that create a wxThreadSpecificInfo object that then never gets freed. This was tested using wxGTK 3.2.2.1:
```cpp
#include <wx/wx.h>

#include <future>

int main(int argc, char **argv) {
	wxInitializer initializer;
	wxSemaphore sema;
	for (;;) {
		std::async(std::launch::async, [&sema]() { sema.Post(); });
		sema.Wait();
	}
	return 0;
}
```

This commit tries to work around the problem by only creating the wxThreadSpecificInfo from wxThreads, and otherwise reverting to using common log objects.

Notably in master this would probably be a trivial thing to work around by just creating the wxThreadSpecificInfo thread_local, but I wanted a workaround for 3.2.

Another place which this will affect is wxTranslations::GetUntranslatedString. I switched it to use a common critical section protected hashmap, which should work just find in my understanding. Although I am not 100% sure if the hash map created by WX_DECLARE_HASH_SET keeps the addresses of its items intact when inserting new items. std::unordered_set should be safe and I am assuming this works the same way.

### DIFF
--- a/include/wx/private/threadinfo.h
+++ b/include/wx/private/threadinfo.h
@@ -52,13 +52,6 @@ public:
     wxLocaleUntranslatedStrings untranslatedStrings;
 #endif
 
-#if wxUSE_THREADS
-    // Cleans up storage for the current thread. Should be called when a thread
-    // is being destroyed. If it's not called, the only bad thing that happens
-    // is that the memory is deallocated later, on process termination.
-    static void ThreadCleanUp();
-#endif
-
 private:
     wxThreadSpecificInfo() : logger(NULL), loggingDisabled(false) {}
 };
--- a/src/common/threadinfo.cpp
+++ b/src/common/threadinfo.cpp
@@ -10,86 +10,14 @@
 // For compilers that support precompilation, includes "wx.h".
 #include "wx/wxprec.h"
 
-
-#include "wx/private/threadinfo.h"
-
 #if wxUSE_THREADS
 
-#include "wx/tls.h"
-#include "wx/thread.h"
-#include "wx/sharedptr.h"
-#include "wx/vector.h"
+// The threaded version for wxThreadSpecificInfo::Get() is defined in platform specific
+// thread files: src/msw/thread.cpp or src/unix/threadpsx.cpp
 
-namespace
-{
+#else
 
-// All thread info objects are stored in a global list so that they are
-// freed when global objects are destroyed and no memory leaks are reported.
-
-// Notice that we must be using accessor functions instead of simple global
-// variables here as this code could be executed during global initialization
-// time, i.e. before any globals in this module were initialzied.
-inline wxCriticalSection& GetAllThreadInfosCS()
-{
-    static wxCriticalSection s_csAllThreadInfos;
-
-    return s_csAllThreadInfos;
-}
-
-typedef wxVector< wxSharedPtr<wxThreadSpecificInfo> > wxAllThreadInfos;
-inline wxAllThreadInfos& GetAllThreadInfos()
-{
-    static wxAllThreadInfos s_allThreadInfos;
-
-    return s_allThreadInfos;
-}
-
-// Pointer to the current thread's instance
-inline wxTLS_TYPE_REF(wxThreadSpecificInfo*) GetThisThreadInfo()
-{
-    static wxTLS_TYPE(wxThreadSpecificInfo*) s_thisThreadInfo;
-
-    return s_thisThreadInfo;
-}
-
-#define wxTHIS_THREAD_INFO wxTLS_VALUE(GetThisThreadInfo())
-
-} // anonymous namespace
-
-
-wxThreadSpecificInfo& wxThreadSpecificInfo::Get()
-{
-    if ( !wxTHIS_THREAD_INFO )
-    {
-        wxTHIS_THREAD_INFO = new wxThreadSpecificInfo;
-        wxCriticalSectionLocker lock(GetAllThreadInfosCS());
-        GetAllThreadInfos().push_back(
-                wxSharedPtr<wxThreadSpecificInfo>(wxTHIS_THREAD_INFO));
-    }
-    return *wxTHIS_THREAD_INFO;
-}
-
-void wxThreadSpecificInfo::ThreadCleanUp()
-{
-    if ( !wxTHIS_THREAD_INFO )
-        return; // nothing to do, not used by this thread at all
-
-    // find this thread's instance in GetAllThreadInfos() and destroy it
-    wxCriticalSectionLocker lock(GetAllThreadInfosCS());
-    for ( wxAllThreadInfos::iterator i = GetAllThreadInfos().begin();
-          i != GetAllThreadInfos().end();
-          ++i )
-    {
-        if ( i->get() == wxTHIS_THREAD_INFO )
-        {
-            GetAllThreadInfos().erase(i);
-            wxTHIS_THREAD_INFO = NULL;
-            break;
-        }
-    }
-}
-
-#else // !wxUSE_THREADS
+#include "wx/private/threadinfo.h"
 
 wxThreadSpecificInfo& wxThreadSpecificInfo::Get()
 {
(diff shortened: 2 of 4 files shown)


src/common/log.cpp
```diff
@@ -580,13 +581,16 @@ wxLog *wxLog::SetThreadActiveTarget(wxLog *logger)
...
+        if (oldLogger)
+            oldLogger->Flush();
 
-    wxThreadInfo.logger = logger;
+        wxThreadInfo->logger = logger;
+        return oldLogger;
+    }
 
-    return oldLogger;
+    wxASSERT_MSG( false, "SetThreadActiveTarget() only works when called from a wxThread" );
```
vadz, line 593: This should be a `wxCHECK_MSG( wxThreadInfo, ...)` above (would also preserve the indent of the existing code).

src/common/log.cpp
```diff
@@ -797,14 +801,19 @@ void wxLog::FlushThreadMessages()
...
 /* static */
 bool wxLog::EnableThreadLogging(bool enable)
 {
-    const bool wasEnabled = !wxThreadInfo.loggingDisabled;
-    wxThreadInfo.loggingDisabled = !enable;
+    const bool wasEnabled = IsThreadLoggingEnabled();
+    if ( wxThreadInfo )
+        wxThreadInfo->loggingDisabled = !enable;
+    else
+        wxASSERT_MSG(enable, "Disabling thread logging only works for wxThreads");
```
vadz, line 816: Same remark as above.

src/common/translation.cpp
```diff
@@ -1546,17 +1546,16 @@ wxString wxTranslations::GetBestTranslation(const wxString& domain,
...
 const wxString& wxTranslations::GetUntranslatedString(const wxString& str)
 {
-    wxLocaleUntranslatedStrings& strings = wxThreadInfo.untranslatedStrings;
-
-    wxLocaleUntranslatedStrings::iterator i = strings.find(str);
-    if ( i == strings.end() )
-        return *strings.insert(str).first;
-
-    return *i;
+    static wxCriticalSection s_cs;
```
vadz, line 1555: This is only MT-safe in C++11, in C++98 we'd need to make it global. Not sure if it's worth changing this...

include/wx/private/threadinfo.h
```diff
@@ -47,12 +41,12 @@ class wxThreadSpecificInfo
...
-#if wxUSE_INTL
-    // Storage for wxTranslations::GetUntranslatedString()
-    wxLocaleUntranslatedStrings untranslatedStrings;
-#endif
-
 #if wxUSE_THREADS
+    // Initialize the wxThreadSpecificInfo.
+    // If this is not called from a thread, the wxThreadInfo will remain NULL
+    // in that thread's context. Use ThreadCleanUp to free when thread ends.
+    static void ThreadInitialize();
```
vadz, line 48: It would make sense to have a `wxThreadInfoInitializer` class calling this in its ctor and `CleanUp()` in its dtor now that we have both functions.

vadz (commented):
Thanks for looking into this!

This is definitely a problem which needs to be fixed, but while I've started
writing a reply saying that I didn't see a way to do it in any other way, I've finally deleted it because I think I do see a way, or even a couple of ways, to do it better.

Basically, we just need to ensure that `wxThreadSpecificInfo::ThreadCleanUp()` is called on _any_ thread exit, not just when `wxThread` terminates. And it should be possible to do it:

1. In C++11 I think we could just create a global `thread_local` object whose dtor would be executed -- if my understanding is correct, I didn't test it -- whenever a thread terminates. Then it would be simply a matter of calling `ThreadCleanUp()` from it. And I think the ordering should be correct, i.e. it should happen after all the rest of the thread code ends running.
1. In C++98, if we/you care about fixing it for it too, although this seems much less important considering that it doesn't have `std::thread` and so people using it are much more likely to use `wxThread`, we could still do it in platform-specific way by registering a function to call on thread exit when creating `wxThreadSpecificInfo`:
 * For Win32, we could use [FLS callback](https://learn.microsoft.com/en-us/windows/win32/api/winnt/nc-winnt-pfls_callback_function).
 * For Pthreads, we could use normal TLS, i.e. [pthread_key_create()](https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_key_create.html).

Doing this would be much preferable to this PR because it wouldn't (silently!) change the behaviour of the existing code, which is really problematic, especially in 3.2 branch.

It would be great if you could please try re-fixing this problem in this way, so as to preserve the current behaviour and let `wxLog` etc work as it does now.

TIA!

nietsu:
Alright, thanks for the feedback. I'll see if I can get the ThreadCleanUp call to work on all threads. I actually have an FLS-based TLS implementation somewhere which might be usable here...

I believe with C++11 thread_local would make the entire ThreadCleanUp obsolete: you could just implement the wxThreadSpecificInfo::Get() like:
`thread_local wxThreadSpecificInfo info;`
`return info;`

vadz:
> Alright, thanks for the feedback. I'll see if I can get the ThreadCleanUp call to work on all threads. I actually have an FLS-based TLS implementation somewhere which might be usable here...

Note that this would only be needed to support non-`wxThread` uses from C++98 programs. I think this should be quite rare...
 
> I believe with C++11 thread_local would make the entire ThreadCleanUp obsolete: you could just implement the wxThreadSpecificInfo::Get() like: `thread_local wxThreadSpecificInfo info;` `return info;`

Hmm, yes, it could actually indeed be as simple as this. And if so, we should do it -- please don't hesitate making a PR against master simplifying this. TIA!

nietsu:
> Note that this would only be needed to support non-`wxThread` uses from C++98 programs. I think this should be quite rare...

Well, the problem is there regardless of the C++ version in main program as long as wxWidgets itself uses C++98, is it not? Once wxWidgets switches to C++11 (as it has in master?) the workaround becomes obsolete, but meanwhile I'd like to have a workaround for stable 3.2.

nietsu:
Does this look more like what you had in mind? (overlooking the stray `#include <wx/tls.h>` I seem to have left in thread.cpp...)

I'm looking at master now as well. I think the entire threadinfo.h/cpp files might be obsolete. You could just declare thread_local logger and loggingdisabled in log.cpp and thread_local strings in wxTranslations::GetUntranslatedString. There's a comment about wxThreadSpecificInfo being about preserving TLS slots, and I don't think that's really a concern anymore with thread_local?

Alternatively I could just alter the wxThreadSpecificInfo::Get to use thread_local, remove the ThreadCleanUp, and keep the rest unchanged.

vadz:
> Does this look more like what you had in mind?

Yes, this looks good, thanks! We just need to declare the function and load it dynamically as we can't rely on having it in the headers/import library :-(

> I'm looking at master now as well. I think the entire threadinfo.h/cpp files might be obsolete.

Yes, I think you're right.

> You could just declare thread_local logger and loggingdisabled in log.cpp and thread_local strings in wxTranslations::GetUntranslatedString. There's a comment about wxThreadSpecificInfo being about preserving TLS slots, and I don't think that's really a concern anymore with thread_local?

I don't know how is `thread_local` implemented internally. I suspect that they already do the equivalent of what wx code does, i.e. use a single TLS slot to store all thread local variables as otherwise I don't see how could it work. But I'm not 100% sure.
 
> Alternatively I could just alter the wxThreadSpecificInfo::Get to use thread_local, remove the ThreadCleanUp, and keep the rest unchanged.

This would be the simplest/fastest change, so if you don't have time to do the rest, doing just this would be already nice. But I'd welcome getting rid of it entirely too.

Thanks again!

nietsu:
Another day, another version.

There was a note on Fls function documentation that they are not supported until Vista. I added Tls fallbacks as workaround, but they require the cleanup mechanic again, meaning that the leak is still there on Windows XP. Still an improvement I guess.

vadz:
Thanks for all this work, looks good to me now! I'm going to merge it soon with just some minor reformatting (to better conform to wx style) and rewording.

vadz:
Merged now, thanks again!

PR #23542, opened 2023-05-11
Fix wxMSW wxWindow::GetTextExtent() with given font in high DPI

vadz:
This function returned wrong results when using a non-default font in
non-standard DPI because it didn't adjust the font to use the window
DPI, as needs to be done manually in wxMSW before using it.

Fix this now and do the same thing here as we already did in
wxWindow::SetFont() (via WXAdjustFontToOwnPPI()) or in wxDC::SetFont().

### DIFF
--- a/src/msw/textmeasure.cpp
+++ b/src/msw/textmeasure.cpp
@@ -69,8 +69,25 @@ void wxTextMeasure::BeginMeasuring()
     // also if we're associated with a window because the window HDC created
     // above has the default font selected into it and not the font of the
     // window.
-    if ( m_font || m_win )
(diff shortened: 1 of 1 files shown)


vadz:
@MaartenBent I'm relatively sure that this is correct as otherwise we just get wrong values from `GetTextExtent(font)`, but please let me know if I'm missing anything. TIA!

MaartenBent:
Yes, I see the problem. If I use
```c++
    wxFont f(wxFontInfo(17));
    wxTextMeasure txm(this, &f);
    wxCoord w, h;
    txm.GetTextExtent("hello", &w, &h);
```
in the minimal sample, it returns the same sizes at displays with different DPI. With the patch this is fixed.

PR #23543, opened 2023-05-12
Fix a memory leak when using non-wx threads. (C++11)

nietsu:
A fix to the same memory leak issue as in #23535 

Now that we have C++11 and `thread_local` available, we can replace the entire wxThreadSpecificInfo structure with it. This simplifies code and resolves the problem with freeing thread local objects on all platforms.

While I have not found concrete specification for it, I am assuming that `thread_local` does not have as strict limits as the previous TLS slot designs, and thus it is unnecessary to wrap all thread_local objects into one, thus making `wxThreadSpecificInfo` unnecessary. Alternatively it would be possible to just implement `wxThreadSpecificInfo::Get()` with `tread_local`, but I think this solution is cleaner.

Note that I am not quite familiar with the process of removing source files from the project, and I did some manual removal of references to `threadinfo.cpp`. It's possible I have overlooked something.


vadz:
Thanks again for doing this, this is definitely a very nice simplification!

To remove a file you basically need to follow the same steps as in https://github.com/wxWidgets/wxWidgets/blob/master/docs/contributing/how-to-add-files-to-build-system.md but it doesn't cover a couple of places which you've also updated, so thank you for this too.

I'm applying this with just some minor changes: I made `tl_xxx` symbols static instead of global/extern and renamed them to use `wxPerThread` prefix instead of relatively cryptic `tl_` (of course, this is much longer, but I think it's worth it for such relatively rarely used variables).

Thanks again!

Issue #23512, opened 2023-05-01
Kicad very sluggish when some of its windows are hidden


kuon:
I am trying to find the reason for the following issue.

When using kicad under Xwayland under sway, if one window is hidden, the whole application will be very sluggish.

After some investigation, it seems to be due to `eglSwapBuffers` "slowing down" when window is hidden, [ref][1].

I am still unsure if this should be fixed here. I'll happily help as much as I can if you need more info.

kicad 7.0.1 with wxWidgets 3.2.2

[1]: https://github.com/swaywm/sway/issues/7557

vadz:
Thanks for reporting this, I didn't know that calling `eglSwapBuffers()` on hidden windows is so bad. Of course, it doesn't make much sense to call `wxGLCanvas::SwapBuffers()` on a hidden window anyhow, but I guess this might somehow be happening in KiCad...

Anyhow, it would be simple enough to add a check for whether a window is hidden to https://github.com/wxWidgets/wxWidgets/blob/1093bb492bc90531162490ba1711f9dba09f59e0/src/unix/glegl.cpp#L636-L645

E.g. we could just replace the test for `!m_readyToDraw` with `IsShownOnScreen()`. Would you be able to test if this fixes the problem?

kuon:
Yes, I can definitely test if it fixes the problem. Maybe not today as I'll have to compile everything, but this week I should be able to.

dgud:
How expensive is IsShownOnScreen?  
We don't want do something expensive in every call to SwapBuffers() that might be called very often.

martantoine:
Hello there,
Just to confirm I got the same problem on my laptop running sway, those freezes are really slowing my workflow
Glad to see I'm not the only one with this issue 
kicad 7.0.1

imciner2:
Some searching shows that there is a slightly different recommended way of handling not drawing on a hidden window in the Wayland rendering loop: https://emersion.fr/blog/2018/wayland-rendering-loop/. I am not sure how easily that can be adapted to work on the GL canvas though, so I wonder if it is something we might have to put into downstream KiCad instead.

swt2c:
> Except that we're not doing it correctly, AFAICS, because we never reset m_readyToDraw to false after setting it to true once.

Well, that's because it was originally only intended to avoid the deadlock on the first draw before the surface was realized.

However, your changes make sense to me.  Hopefully they improve things for KiCad.

swt2c:
[...]

PR #23554, opened 2023-05-17
Fix `wxGLCanvas` performance when using Wayland

vadz:
This should hopefully help with #23512, but I haven't actually tested it with Kicad itself (yet).

Note: when backporting to 3.2, the last commit should be omitted or `constexpr` in it changed.

### DIFF
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -638,8 +638,20 @@ void wxGLCanvasEGL::FreeDefaultConfig()
 
 bool wxGLCanvasEGL::SwapBuffers()
 {
-#ifdef GDK_WINDOWING_WAYLAND
     GdkWindow* const window = GTKGetDrawingWindow();
+#ifdef GDK_WINDOWING_X11
+    if (wxGTKImpl::IsX11(window))
+    {
+        if ( !IsShownOnScreen() )
+        {
+            // Trying to draw on a hidden window is useless and can actually be
+            // harmful if the compositor blocks in eglSwapBuffers() in this
+            // case, so avoid it.
+            return false;
+        }
+    }
+#endif // GDK_WINDOWING_X11
+#ifdef GDK_WINDOWING_WAYLAND
     if (wxGTKImpl::IsWayland(window))
     {
         // Under Wayland, we must not draw before we're actually ready to, as
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -471,7 +471,6 @@ bool wxGLCanvasEGL::CreateSurface()
         m_xwindow = GDK_WINDOW_XID(window);
         m_surface = eglCreatePlatformWindowSurface(m_display, *m_config,
                                                    &m_xwindow, nullptr);
-        m_readyToDraw = true;
     }
 #endif
 #ifdef GDK_WINDOWING_WAYLAND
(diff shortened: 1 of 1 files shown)


vadz:
There is not much time left for merging this and backporting it to 3.2 if we want this to be included in 3.2.3.

@imciner2 @kuon Please test this PR if you can.

kuon:
I'm sorry, I broke my leg and couldn't test this PR yet. It's on my list, I'll do it asap.

vadz:
Great, thanks a lot for testing!

I'm going to merge this in master now and will cherry-pick it to 3.2 too.

artizirk:
After thinking about this for a while I think I may have found a small oversight.
`IsShownOnScreen()` uses `m_readyToDraw` as one of the checks under Wayland. Before this PR, `m_readyToDraw` was set to `true` **once** after first `wl_frame_callback_handler`. After this PR now the `m_readyToDraw` is toggled after each frame.

`IsShownOnScreen()` probably shouldn't flap at 60hz?

I hacked it use `gtk_window_is_visible()` for `IsShownOnScreen` check and in my limited testing Kicad renders a bit better.

```
index c35273caddaa..0f009e3f4fb8 100644
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -697,12 +697,13 @@ bool wxGLCanvasEGL::SwapBuffers()
 bool wxGLCanvasEGL::IsShownOnScreen() const
 {
     wxDisplayInfo info = wxGetDisplayInfo();
+    GdkWindow* const window = GTKGetDrawingWindow();
     switch ( info.type )
     {
         case wxDisplayX11:
             return GetXWindow() && wxGLCanvasBase::IsShownOnScreen();
         case wxDisplayWayland:
-            return m_readyToDraw && wxGLCanvasBase::IsShownOnScreen();
+            return gtk_window_is_visible(window) && wxGLCanvasBase::IsShownOnScreen();
         default:
             return false;
     }
```

vadz:
Thanks @artizirk, I think you're right and we shouldn't be using `m_readyToDraw` here, but I'm a bit confused as to how does it actually change anything for redrawing: we don't seem to use `IsShownOnScreen()` anywhere under Wayland (under X11 we use it in `SwapBuffers()`).

What am I missing here?

ShalokShalom:
@vadz Open another issue for that?

vadz:
@ShalokShalom Yes, you're right, we need to fix this. It looks easy, but I still don't understand how does it change anything for KiCad which only uses XWayland AFAIK.

ShalokShalom:
They have a [thread about Wayland support](https://gitlab.com/kicad/code/kicad/-/issues/7207), and it looks like there is even an [Arch package](https://aur.archlinux.org/packages/kicad-wayland) with it.

Manolo-ES:
Sorry, I'm strongly against this PR (internally using `eglSwapInterval`). Other PRs I see affected are https://github.com/wxWidgets/wxWidgets/pull/23909  and https://github.com/wxWidgets/wxWidgets/issues/23512

Setting SwapInterval is a user decision, not a wx one. It may be good for some cases but not for others. See [this Khronos wiki article](https://www.khronos.org/opengl/wiki/Swap_Interval) and its [discussion](https://www.khronos.org/opengl/wiki/Talk:Swap_Interval).
Be aware that any final user can override app's settings by directly selecting it at driver's configuration (some vendors provide it).

Any how, I think SwapInterval is a feature that wx should provide to the user.

vadz:
@Manolo-ES I am not sure if you've followed the previous discussions and bug reports, but without this PR existing code working fine under the other platforms (Windows, Mac and even Linux with X11) makes the program completely unusable under Wayland, i.e. it's a fatal problem and not just something that can be ignored.

I don't think it's reasonable to require all wx programs using OpenGL to call `eglSwapInterval()` themselves and I think it's even less reasonable to expect all users of these programs to change some driver setting.

Issue #23209, opened 2023-02-02
OSX specific: wxLocale: Problem with set locale for any of wxLANGUAGE_CHINESE(...)


YuSanka:
### Description

When we try to switch our application to the some of Chinese languages, then wxWidgets send a message, that cannot set locale for this language.

<img width="279" alt="Screenshot 2023-02-02 at 16 29 04" src="https://user-images.githubusercontent.com/19184891/216372164-7530bf74-ff53-4f15-9e59-de07de6d7021.png">

The bug appears only on macOS

### Platform and version information

- wxWidgets version  you use: 3.2.0
- wxWidgets port  you use: wxOSX
- OS  and its version: tested on Monterey and Ventura

vadz:
Could you please explain what exactly are you doing, i.e. how are you setting the locale? Some code, or, ideally, a diff to the internat sample would be really welcome.

This would at least make it clear if this is related to #23114 or something different.

vadz:
Also, please check if the changes in #23226 change anything for the problem you're seeing.

YuSanka:
Hello, @vadz . Sorry for the late response. But we have to reopen this issue because of it's not fix yet. 
I tried all motioned issues/fixes but, behavior is the same.
To simplest repro just change a wxLANGUAGE_DEFAULT to wxLANGUAGE_CHINESE_TAIWAN for "calendar" sample of wxWidgets.

<img width="552" alt="Screenshot 2023-05-18 at 14 08 21" src="https://github.com/wxWidgets/wxWidgets/assets/19184891/70d18d90-16e7-4ff6-9315-c39a1e3ddf3f">

Run "calendar" application and you will receive a warning which I described.

<img width="1074" alt="Screenshot 2023-05-18 at 14 04 12" src="https://github.com/wxWidgets/wxWidgets/assets/19184891/556b5961-0338-4cd7-b2b0-9aa8693084b1">

vadz:
OK, I can indeed reproduce the problem with a very simple test case:
```cpp
#include <wx/init.h>
#include <wx/crt.h>
#include <wx/datetime.h>
#include <wx/intl.h>

int main(int argc, char **argv)
{
    wxInitializer init;
    if ( !init.IsOk() ) {
        fprintf(stderr, "Failed to initialize wxWidgets.\n");
        return 3;
    }

    wxPrintf("Without locale: %s\n", wxDateTime::Today().FormatDate());
[...]

PR #23556, opened 2023-05-18
Locale-related fixes for macOS and new calendar names localization functions

vadz:
@utelle I'd appreciate your thoughts about this, notably if we should change `FromTag()` to deal with strings of the form `zh_Hant_TW` directly and/or if you see any better way of handling this at `NSLocale` level.

### DIFF
--- a/include/wx/datetime.h
+++ b/include/wx/datetime.h
@@ -265,11 +265,19 @@ public:
         // flags for GetWeekDayName and GetMonthName
     enum NameFlags
     {
-        Name_Full = 0x01,       // return full name
-        Name_Abbr = 0x02        // return abbreviated name
+        Name_Full = 0x01,        // return full name
+        Name_Abbr = 0x02,        // return abbreviated name
+        Name_Shortest = 0x03     // return shortest name
     };
 
-        // flags for GetWeekOfYear and GetWeekOfMonth
+    // context for GetWeekDayName and GetMonthName
+    enum NameContext
+    {
+        Context_Formatting,      // return name for date formatting context
+        Context_Standalone       // return name for standalone context
+    };
+
+    // flags for GetWeekOfYear and GetWeekOfMonth
     enum WeekFlags
     {
         Default_First,   // Sunday_First for US, Monday_First for the rest
@@ -288,6 +296,28 @@ public:
 
     // helper classes
     // ------------------------------------------------------------------------
+
+        // Describes the form of the month or week-day name.
+    class NameForm
+    {
+    public:
+        // Ctor is non-explicit for compatibility.
+        NameForm(NameFlags flags = Name_Full) : m_flags(flags) {}
+
+        // Chainable methods allowing to set various fields.
+        NameForm& Full() { m_flags = Name_Full; return *this; }
+        NameForm& Abbr() { m_flags = Name_Abbr; return *this; }
+        NameForm& Shortest() { m_flags = Name_Shortest; return *this; }
+        NameForm& Formatting() { m_context = Context_Formatting; return *this; }
+        NameForm& Standalone() { m_context = Context_Standalone; return *this; }
+
+        NameFlags GetFlags() const { return m_flags; }
+        NameContext GetContext() const { return m_context; }
+
+    private:
+        NameFlags m_flags;
+        NameContext m_context = Context_Formatting;
+    };
 
         // a class representing a time zone: basically, this is just an offset
         // (in seconds) from GMT
@@ -411,23 +441,23 @@ public:
                                         Calendar cal = Gregorian);
 
 
-        // get the full (default) or abbreviated month name in the current
+        // get the full (default), abbreviated or shortest month name in the current
         // locale, returns empty string on error
     static wxString GetMonthName(Month month,
-                                 NameFlags flags = Name_Full);
+                                 const NameForm& form = {});
 
-        // get the standard English full (default) or abbreviated month name
+        // get the standard English full (default), abbreviated or shortest month name
     static wxString GetEnglishMonthName(Month month,
-                                        NameFlags flags = Name_Full);
+                                        const NameForm& form = {});
 
-        // get the full (default) or abbreviated weekday name in the current
+        // get the full (default), abbreviated or shortest weekday name in the current
         // locale, returns empty string on error
     static wxString GetWeekDayName(WeekDay weekday,
-                                   NameFlags flags = Name_Full);
+                                   const NameForm& form = {});
 
-        // get the standard English full (default) or abbreviated weekday name
+        // get the standard English full (default), abbreviated or shortest weekday name
     static wxString GetEnglishWeekDayName(WeekDay weekday,
-                                          NameFlags flags = Name_Full);
+                                          const NameForm& form = {});
 
         // get the AM and PM strings in the current locale (may be empty)
     static void GetAmPmStrings(wxString *am, wxString *pm);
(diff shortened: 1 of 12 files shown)


utelle:
> @utelle I'd appreciate your thoughts about this, notably if we should change `FromTag()` to deal with strings of the form `zh_Hant_TW` directly

Well, I'm a bit confused. According to the [macOS documentation](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html) locale identifiers which contain a script part should have a **hyphen** to separate the _script_ code from the _language_ code, **not** an **underscore**. Method `FromTag` already handles tags of the form `zh-Hant_TW` correctly.

So, where does the tag with **2** underscore delimiters come from? User input? If yes, the syntax is simply wrong. From the system? In that case I'd say it is a bug in the system.

> and/or if you see any better way of handling this at `NSLocale` level.

I will have to take a closer look whether there exists a better handling option.

vadz:
> So, where does the tag with **2** underscore delimiters come from?

From `[NSLocale availableLocaleIdentifiers]`, which is what this code works with. You could say that it's a bug, or you could say that the documentation is incorrect, but we can't really change it, so we have to deal with it and the only possibilities I see is what I did, which is less disruptive, or changing `FromTag()` to deal with this format (which would involve at least preventing it from "recognizing" `Hant_TW` as Windows sort order...).

> > and/or if you see any better way of handling this at `NSLocale` level.
> 
> I will have to take a closer look whether there exists a better handling option.

TIA!

vadz:
> Throughout all Apple documentation I haven't seen locale identifiers with 2 underscore delimiters.

Yes, I know, which is why I think it could be also a bug in their documentation. But we have to live with what we have...

> `availableLocaleIdentifiers` returns a **very** long list of locale tags with quite a number of exotic locales. The list is completely unsorted. I haven't checked **all** entries, but regarding _Chinese_ I found entries which will make it hard to determine the right entry: `zh`, `zh_Hans`, `zh_Hant`, `zh_Hans_CN`, `zh_Hant_CN` ...
> 
> IMHO the script part should be checked, too, if the tag to be checked contains a script part. I wonder which entry macOS selects if no script part is given.

I couldn't determine this, but I'm also worried by it, see the commit message. The trouble is that we can't really do much without more information than we currently have: if all we've got is `zh_CN`, how are we supposed to know that its default script should be `Hans`? Other than checking for "Simplified" in the description I don't think we can get this information from the current language database contents...

> What we could do, if we find more than one underscore delimiter, is to replace the underscores by hyphens. Then interpreting the tag would work as expected. As far as I can tell underscores are not allowed as a vaild character of a tag part.
> 
> In your code you do just that. It seems to be logical to integrate that approach in method `FromTag`.

I thought about this too but wasn't completely sure if it was safe to do it in general, which is why I preferred to do it in Mac-specific code only. If you think it's ok, I can change this or I can merge this PR as is (because I'd also like to backport it to 3.2) and then you could update it if you'd like. Please let me know, TIA!

utelle:
> > Throughout all Apple documentation I haven't seen locale identifiers with 2 underscore delimiters.
> Yes, I know, which is why I think it could be also a bug in their documentation. But we have to live with what we have...

Yes, there is no other choice. Fortunately, macOS is the only platform using this locale tag representation with 2 underscores. Under Windows and Linux I haven't seen any tags with more than one underscore.

> The trouble is that we can't really do much without more information than we currently have: if all we've got is `zh_CN`, how are we supposed to know that its default script should be `Hans`? Other than checking for "Simplified" in the description I don't think we can get this information from the current language database contents...

If the user does not specify a script, we need to rely on `localeWithLocaleIdentifier` to return the most appropriate locale for the given language and country codes. And most likely that is the case.

I don't think it is worth the effort to extend the language database by the default script of the locales.

> > What we could do, if we find more than one underscore delimiter, is to replace the underscores by hyphens. [...]
> > In your code you do just that. It seems to be logical to integrate that approach in method `FromTag`.
> 
> I thought about this too but wasn't completely sure if it was safe to do it in general, which is why I preferred to do it in Mac-specific code only. If you think it's ok, I can change this or I can merge this PR as is (because I'd also like to backport it to 3.2) and then you could update it if you'd like. Please let me know, TIA!

IMHO it is safe to handle tags with 2 underscores within `FromTag`.

The change to `FromTag` should be rather simple. Only the part handling a Windows sort order needs to be modified. The current code is the following:

```C++
    // 1d. Check for sort order in Windows tag
    //
    // Make sure we don't extract the region identifier erroneously as a sortorder identifier
    {
        wxString tagTemp = tagMain.BeforeFirst('_', &tagRest);
        if (tagRest.length() > 4 && locId.m_modifier.empty() && locId.m_charset.empty())
        {
            // Windows sortorder found
            locId.SortOrder(tagRest);
            tagMain = tagTemp;
        }
    }
```

Changing the line determining `tagTemp` as follows should do the trick:

```C++
        wxString tagTemp = tagMain.BeforeLast('_', &tagRest);
```
If the part that potentially represents a sort order is not long enough, the `tagMain` remains unchanged, and in the next step all underscores are replaced by hyphens.

utelle:
> IMHO it is safe to handle tags with 2 underscores within `FromTag`.

This definitely works ... after adjusting the code for detecting a Windows sort order attribute.

> The change to `FromTag` should be rather simple. Only the part handling a Windows sort order needs to be modified. The current code is the following:
> 
> ```c++
>     // 1d. Check for sort order in Windows tag
>     //
>     // Make sure we don't extract the region identifier erroneously as a sortorder identifier
>     {
>         wxString tagTemp = tagMain.BeforeFirst('_', &tagRest);
>         if (tagRest.length() > 4 && locId.m_modifier.empty() && locId.m_charset.empty())
>         {
>             // Windows sortorder found
>             locId.SortOrder(tagRest);
>             tagMain = tagTemp;
>         }
>     }
> ```
> 
> Changing the line determining `tagTemp` as follows should do the trick:
> 
> ```c++
>         wxString tagTemp = tagMain.BeforeLast('_', &tagRest);
> ```

Using `wxString::BeforeLast()` works, if one takes into account that `tagTemp` will be empty, if the search character (here: underscore) was not found at all. Therefore it is necessary to verify that `tagTemp` is **not** empty, before assuming a sort order attribute:

```c++
        if (tagTemp.length() > 0 && tagRest.length() > 4 &&
            locId.m_modifier.empty() && locId.m_charset.empty())
```

PR #23562 includes this adjustment.

utelle (changes requested):
Unfortunately, I had 2 small glitches in my original PR. In case of falling back to English names the methods must be called with the parameter `form` (not `flags`). This only affects systems that don't have `HAVE_LANGINFO_H` defined.

vadz:
@utelle Just to be sure: this is ok to merge now, right?

utelle:
> @utelle Just to be sure: this is ok to merge now, right?

Yes, AFAICT all related issues are addressed and all checks are successful. So, it should be ok to merge.

marekr:
Curious, is the changes to matching the locale in uilocale.mm going to be backported to 3.2?

We have issues in KiCad where users cannot switch their app language to zh_TW or zh_CN and I wonder if it's because the OS is returning the longer names i.e. zh_Hant_TW

vadz:
@marekr The particular problem you're speaking of is #23209 which is a (strict) subset of the changes in this PR, so I won't backport all of the changes here, but I will indeed backport 6d849f8211977ab969f9fd4a503d6ad362f95e53 and a couple of related commits, thanks for the reminder.

I'll also backport the fix for #23557, i.e. PR #23603.

PR #23562, opened 2023-05-24
Add wxUILocale methods, adjust wxDateTime

utelle:
Implement wxUILocale::GetMonthName() and wxUILocale::GetWeekDayName() on wxMSW, wxGTK, and wxOSX.

Adjust wxDateTime to support a 3rd month and  weekday name representation (Name_Shortest) and a usage context (Context_Formatting and Context_Standalone).

### DIFF
--- a/include/wx/datetime.h
+++ b/include/wx/datetime.h
@@ -265,11 +265,19 @@ public:
         // flags for GetWeekDayName and GetMonthName
     enum NameFlags
     {
-        Name_Full = 0x01,       // return full name
-        Name_Abbr = 0x02        // return abbreviated name
+        Name_Full = 0x01,        // return full name
+        Name_Abbr = 0x02,        // return abbreviated name
+        Name_Shortest = 0x03     // return shortest name
     };
 
-        // flags for GetWeekOfYear and GetWeekOfMonth
+    // context for GetWeekDayName and GetMonthName
+    enum NameContext
+    {
+        Context_Formatting,      // return name for date formatting context
+        Context_Standalone       // return name for standalone context
+    };
+
+    // flags for GetWeekOfYear and GetWeekOfMonth
     enum WeekFlags
     {
         Default_First,   // Sunday_First for US, Monday_First for the rest
@@ -288,6 +296,28 @@ public:
 
     // helper classes
     // ------------------------------------------------------------------------
+
+        // Describes the form of the month or week-day name.
+    class WXDLLIMPEXP_BASE NameForm
+    {
+    public:
+        // Ctor is non-explicit for compatibility.
+        NameForm(NameFlags flags = Name_Full) : m_flags(flags) {}
+
+        // Chainable methods allowing to set various fields.
+        NameForm& Full() { m_flags = Name_Full; return *this; }
+        NameForm& Abbr() { m_flags = Name_Abbr; return *this; }
+        NameForm& Shortest() { m_flags = Name_Shortest; return *this; }
+        NameForm& Formatting() { m_context = Context_Formatting; return *this; }
+        NameForm& Standalone() { m_context = Context_Standalone; return *this; }
+
+        NameFlags GetFlags() { return m_flags; }
+        NameContext GetContext() { return m_context; }
+
+    private:
+        NameFlags m_flags;
+        NameContext m_context = Context_Formatting;
+    };
 
         // a class representing a time zone: basically, this is just an offset
         // (in seconds) from GMT
@@ -411,23 +441,23 @@ public:
                                         Calendar cal = Gregorian);
 
 
-        // get the full (default) or abbreviated month name in the current
+        // get the full (default), abbreviated or shortest month name in the current
         // locale, returns empty string on error
     static wxString GetMonthName(Month month,
-                                 NameFlags flags = Name_Full);
+                                 NameForm form = {});
 
-        // get the standard English full (default) or abbreviated month name
+        // get the standard English full (default), abbreviated or shortest month name
     static wxString GetEnglishMonthName(Month month,
-                                        NameFlags flags = Name_Full);
+                                        NameForm form = {});
 
-        // get the full (default) or abbreviated weekday name in the current
+        // get the full (default), abbreviated or shortest weekday name in the current
         // locale, returns empty string on error
     static wxString GetWeekDayName(WeekDay weekday,
-                                   NameFlags flags = Name_Full);
+                                   NameForm form = {});
 
-        // get the standard English full (default) or abbreviated weekday name
+        // get the standard English full (default), abbreviated or shortest weekday name
     static wxString GetEnglishWeekDayName(WeekDay weekday,
-                                          NameFlags flags = Name_Full);
+                                          NameForm form = {});
 
         // get the AM and PM strings in the current locale (may be empty)
     static void GetAmPmStrings(wxString *am, wxString *pm);
--- a/include/wx/datetime.h
+++ b/include/wx/datetime.h
@@ -444,7 +444,7 @@ public:
         // get the full (default), abbreviated or shortest month name in the current
         // locale, returns empty string on error
     static wxString GetMonthName(Month month,
-                                 NameForm form = {});
+                                 const NameForm& form = {});
 
         // get the standard English full (default), abbreviated or shortest month name
     static wxString GetEnglishMonthName(Month month,
--- a/include/wx/datetime.h
+++ b/include/wx/datetime.h
@@ -311,8 +311,8 @@ public:
         NameForm& Formatting() { m_context = Context_Formatting; return *this; }
         NameForm& Standalone() { m_context = Context_Standalone; return *this; }
 
-        NameFlags GetFlags() { return m_flags; }
-        NameContext GetContext() { return m_context; }
+        NameFlags GetFlags() const { return m_flags; }
+        NameContext GetContext() const { return m_context; }
 
     private:
         NameFlags m_flags;
--- a/include/wx/datetime.h
+++ b/include/wx/datetime.h
@@ -448,16 +448,16 @@ public:
 
         // get the standard English full (default), abbreviated or shortest month name
     static wxString GetEnglishMonthName(Month month,
-                                        NameForm form = {});
+                                        const NameForm& form = {});
 
         // get the full (default), abbreviated or shortest weekday name in the current
         // locale, returns empty string on error
     static wxString GetWeekDayName(WeekDay weekday,
-                                   NameForm form = {});
+                                   const NameForm& form = {});
 
         // get the standard English full (default), abbreviated or shortest weekday name
     static wxString GetEnglishWeekDayName(WeekDay weekday,
-                                          NameForm form = {});
+                                          const NameForm& form = {});
 
         // get the AM and PM strings in the current locale (may be empty)
     static void GetAmPmStrings(wxString *am, wxString *pm);
--- a/include/wx/datetime.h
+++ b/include/wx/datetime.h
@@ -298,7 +298,7 @@ public:
     // ------------------------------------------------------------------------
 
         // Describes the form of the month or week-day name.
-    class WXDLLIMPEXP_BASE NameForm
+    class NameForm
     {
     public:
         // Ctor is non-explicit for compatibility.
--- a/src/osx/core/uilocale.mm
+++ b/src/osx/core/uilocale.mm
@@ -169,7 +169,7 @@ wxUILocaleImplCF::GetName() const
     wxString name = wxCFStringRef::AsString([m_nsloc localeIdentifier]);
 
     // Check for the special case of the "empty" system locale, see CreateStdC()
-    if ( name.empty() )
+    if ( name.empty() || name.IsSameAs("en_US_POSIX") )
         name = "C";
 
     return name;
--- a/src/osx/core/uilocale.mm
+++ b/src/osx/core/uilocale.mm
@@ -20,6 +20,7 @@
 
 #if wxUSE_INTL
 
+#include "wx/log.h"
 #include "wx/uilocale.h"
 #include "wx/private/uilocale.h"
 
@@ -29,6 +30,7 @@
 #import <Foundation/NSArray.h>
 #import <Foundation/NSString.h>
 #import <Foundation/NSLocale.h>
+#import <Foundation/NSDateFormatter.h>
 
 extern wxString
 wxGetInfoFromCFLocale(CFLocaleRef cfloc, wxLocaleInfo index, wxLocaleCategory cat);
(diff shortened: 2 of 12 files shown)


utelle:
Unfortunately, some of the checks fail on macOS. I will have to debug what's going on, because my direct tests with the new methods (that is, calling wxUILocale methods directly) seemed to work as expected.

vadz:
Sorry, I thought that "generic values with little localization" meant POSIX locale, but I was clearly wrong. I think using "en_US_POSIX" is fine, I seem to remember that I simply haven't found this name because I believe that I tried "C" and "POSIX" and neither of them worked (but I could be misremembering something too).

It would be best to make this change in a separate PR, as it should be backported to 3.2 too, I think, but you could also add it here and I'd cherry-pick just this commit into 3.2 later if you think another PR is an overkill.

TIA!

src/osx/core/uilocale.mm
```diff
@@ -86,12 +88,48 @@ explicit wxUILocaleImplCF(NSLocale* nsloc)
         // Surprisingly, localeWithLocaleIdentifier: always succeeds, even for
         // completely invalid strings, so we need to check if the name is
         // actually in the list of the supported locales ourselves.
-        static wxCFRef<CFArrayRef>
-            all((CFArrayRef)[[NSLocale availableLocaleIdentifiers] retain]);
+//        static wxCFRef<CFArrayRef>
+//            all((CFArrayRef)[[NSLocale availableLocaleIdentifiers] retain]);
```
vadz, line 92: This shouldn't remain commented out here, please feel free to cherry pick my commit if it's convenient or let me merge it into master already if you prefer -- just let me know if I should do it or not.

src/osx/core/uilocale.mm
```diff
@@ -185,6 +226,104 @@ int CompareStrings(const wxString& lhs, const wxString& rhs,
...
+    const int idx = ArrayIndexFromFlag(form.GetFlags());
+    if (idx == -1)
+        return wxString();
+
+    NSDateFormatter* df = [NSDateFormatter new];
+    df.locale = m_nsloc;
+
+    NSArray* monthNames = nullptr;
+
+    if (form.GetContext() == wxDateTime::Context_Standalone)
```
vadz, line 241: I'd again prefer to use `switch` here.

src/osx/core/uilocale.mm
```diff
@@ -185,6 +226,104 @@ int CompareStrings(const wxString& lhs, const wxString& rhs,
...
+    {
+        switch (idx)
```
vadz, line 243: It seems strange to me to convert flags to index when we don't use it as an index, finally. And having `case wxDateTime::Name_Full` etc would be more clear than the code we have here now.

src/osx/core/uilocale.mm
```diff
@@ -185,6 +226,104 @@ int CompareStrings(const wxString& lhs, const wxString& rhs,
...
+    }
+
+    NSString* monthName = [monthNames objectAtIndex:(month)];
+    return wxCFStringRef::AsString(monthName);
+}
+
+wxString
+wxUILocaleImplCF::GetWeekDayName(wxDateTime::WeekDay weekday, wxDateTime::NameForm form) const
+{
+    const int idx = ArrayIndexFromFlag(form.GetFlags());
```
vadz, line 281: Same comment as above.

src/unix/uilocale.cpp
```diff
@@ -25,6 +25,7 @@
     #include <langinfo.h>
 #endif
 
+#include "wx/datetime.h"
```
vadz, line 28: Not really needed, as already included from the header.

vadz (commented):
Thanks, this actually looks almost perfect to me, in spite of a number of comments -- all of them are pretty minor and hopefully my suggestions can be helpful, i.e. you could just apply them directly.

Thanks again!

utelle:
> Sorry, I thought that "generic values with little localization" meant POSIX locale, but I was clearly wrong.

Well, this fact about the `systemLocale` wasn't clear to me either. And since the date formatting code used standard C functions to retrieve month and weekday names, it all worked. Only the change adding the method `GetMonthName()` and `GetWeekDayName()`to use system functions in `wxUILocale` revealed this.

> I think using "en_US_POSIX" is fine, I seem to remember that I simply haven't found this name because I believe that I tried "C" and "POSIX" and neither of them worked (but I could be misremembering something too).

You are right that neither "C" nor "POSIX" is contained in the list of available locales. Actually, I stumbled over the locale id **en_US_POSIX** by pure chance. It seems to solve the problem. However, I don't want to expose the locale id **en_US_POSIX** to the outside. At least at the moment the method `wxLocaleIdent::FromTag()` is not able to handle this id format properly. The documentation only speaks of the format `language[-script]_region`. Whether the list of available locales contains any locales using this extended variant, I really don't know.

IMHO we should catch the special case of "C" and "POSIX" in method `FromTag` and elsewhere to handle the standard C locale properly.

> It would be best to make this change in a separate PR, as it should be backported to 3.2 too, I think, but you could also add it here and I'd cherry-pick just this commit into 3.2 later if you think another PR is an overkill.

Hm, the necessary changes affect the very same source files. Therefore I would prefer to add further commits to this PR, if it's ok with you.

> Thanks, this actually looks almost perfect to me, in spite of a number of comments -- all of them are pretty minor and hopefully my suggestions can be helpful, i.e. you could just apply them directly.

I will go through your comments in a moment.

include/wx/datetime.h
```diff
@@ -289,6 +297,28 @@ class WXDLLIMPEXP_BASE wxDateTime
     // helper classes
     // ------------------------------------------------------------------------
 
+        // Describes the form of the month or week-day name.
+    class WXDLLIMPEXP_BASE NameForm
```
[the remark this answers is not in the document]
utelle: > Fully-inline classes don't need to be DLL-exported and doing this results in problems sometimes

Seeing that other inner classes like `TimeZone` used the DLL-export spec, I had just copied it without considering the consequences. Thanks for pointing it out.

src/osx/core/uilocale.mm
utelle: I removed the commented out lines. Additionally, I replaced the `systemLocale` by the locale identified by **en_US_POSIX**. This will hopefully make the checks work again for macOS.

There is still some work to do regarding the handling of the standard C locale, that is, handling of special locale tags "C" and "POSIX".

utelle:
Hopefully this PR has reached its final state.

The date formatting checks using the new methods to retrieve month and weekday names for the generic calendar control now succeed. However, it needs to be checked whether the generic calendar control asks for the right context - IMHO it should be `Context_Standalone`.

The wxOSX implementation uses now the locale **en_US_POSIX** as the standard C locale. However, the locale name **en_US_POSIX** will not be reported to the user, instead the name **C** will be used.

Method `wxLocaleIdent::FromTag()` was modified to check for the special locales **C** and **POSIX** first, so that the standard C locale will be associated with these tags.

This PR addresses some other issues resp PRs:
- Code modifications from PR #23556 (except for the related tests) are included in this PR.
- Code for methods `wxUILocale::GetMonthName()` and `wxUILocale::GetWeekDayName()` is implemented and used in `wxDateTime` methods as discussed in PR #23551.
- Issue #23209 should be solved by this PR, too. However, it should be verified somehow that OS X selects the right locale, if no script is given in the locale tag (for example, **zh_CN** instead of **zh_Hans_CN**). Actually, I expect this to work, because each region has a default script, and the OS should know about it.

vadz:
> Hopefully this PR has reached its final state.

Thanks! I'll need some time to get to merging it, unfortunately, as I'm very busy with something else right now, but I'll get back to it a.s.a.p.

> The date formatting checks using the new methods to retrieve month and weekday names for the generic calendar control now succeed. However, it needs to be checked whether the generic calendar control asks for the right context - IMHO it should be `Context_Standalone`.

Yes, it definitely should be `Context_Standalone` there. You should be able to check that it works correctly, i.e. shows "Травень", in Ukrainian locale which I believe you already have as you had done some fixes for it before.

> The wxOSX implementation uses now the locale **en_US_POSIX** as the standard C locale. However, the locale name **en_US_POSIX** will not be reported to the user, instead the name **C** will be used.
> 
> Method `wxLocaleIdent::FromTag()` was modified to check for the special locales **C** and **POSIX** first, so that the standard C locale will be associated with these tags.

Looks good!

> This PR addresses some other issues resp PRs:
> 
>     * Code modifications from PR [Fix for Chinese locales under macOS #23556](https://github.com/wxWidgets/wxWidgets/pull/23556) (except for the related tests) are included in this PR.

So should the other one be just closed, or should it be merged before merging this one?

>     * Code for methods `wxUILocale::GetMonthName()` and `wxUILocale::GetWeekDayName()` is implemented and used in `wxDateTime` methods as discussed in PR [Fix weekday/month name localization in generic wxCalendarCtrl on macOS #23551](https://github.com/wxWidgets/wxWidgets/pull/23551).
> 
>     * Issue [OSX specific: wxLocale: Problem with set locale for any of wxLANGUAGE_CHINESE(...) #23209](https://github.com/wxWidgets/wxWidgets/issues/23209) should be solved by this PR, too. However, it should be verified somehow that OS X selects the right locale, if no script is given in the locale tag (for example, **zh_CN** instead of **zh_Hans_CN**). Actually, I expect this to work, because each region has a default script, and the OS should know about it.

In principle it would be enough to compare the month names in the calendar sample in 3 different cases:

1. `zh_TW` (known traditional)
1. `zh_SG` (known simplified)
1. `zh_CN`

and check if the last one looks like the first one (which would be wrong) or the second one (which would be correct).

utelle:
> I'll need some time to get to merging it, unfortunately, as I'm very busy with something else right now, but I'll get back to it a.s.a.p.

I will use this time to address the points mentioned below.

> > The date formatting checks using the new methods to retrieve month and weekday names for the generic calendar control now succeed. However, it needs to be checked whether the generic calendar control asks for the right context - IMHO it should be `Context_Standalone`.
> 
> Yes, it definitely should be `Context_Standalone` there. You should be able to check that it works correctly, i.e. shows "Травень", in Ukrainian locale which I believe you already have as you had done some fixes for it before.

Only one source file is affected, namely `generic/calctrlg.cpp`. I will adjust and test the code and will commit the changes to this PR.

> > * Code modifications from PR [Fix for Chinese locales under macOS #23556](https://github.com/wxWidgets/wxWidgets/pull/23556) (except for the related tests) are included in this PR.
> 
> So should the other one be just closed, or should it be merged before merging this one?

The modifications in `src/osx/core/uilocale.mm` are already part of this PR, so I think adding the few extra lines of code in `tests/intl/intltest.cpp` to this PR will be easier than a partial merge with the other PR. (Already done in commit 542d49ebfa60a27ce394277156f574566d31b49f ). So, the other PR can be closed.

BTW, most likely the modification of the other PR can be simplified, because of the change in `wxLocaleIdent::FromTag()`. Replacing the first underscore in a tag by a hyphen shouldn't be necessary anymore.

> > However, it should be verified somehow that OS X selects the right locale, if no script is given in the locale tag (for example, **zh_CN** instead of **zh_Hans_CN**). Actually, I expect this to work, because each region has a default script, and the OS should know about it.
> 
> In principle it would be enough to compare the month names in the calendar sample in 3 different cases:
> 
>     1. `zh_TW` (known traditional)
>     2. `zh_SG` (known simplified)
>     3. `zh_CN`
> 
> and check if the last one looks like the first one (which would be wrong) or the second one (which would be correct).

I will perform this check. If it is doesn't work as we hope it should, I will report it here.

Just as a side note: I generated a list of available locales in macOS Ventura. I will analyze the list how much it differs from the language list we already have. We can discuss later whether it is worth the effort to merge additional entries into the existing list. But this topic is definitely not within the scope of this PR.

utelle:
> > > [...] generic calendar control asks for the right context - IMHO it should be `Context_Standalone`.
> > 
> > Yes, it definitely should be `Context_Standalone` there. You should be able to check that it works correctly, i.e. shows "Травень", in Ukrainian locale which I believe you already have as you had done some fixes for it before.
> 
> Only one source file is affected, namely `generic/calctrlg.cpp`. I will adjust and test the code and will commit the changes to this PR.

Done in commit 1e574e02b3bac1980910c283f3f214ca83863889.

> > In principle it would be enough to compare the month names in the calendar sample in 3 different cases:
> > 1. `zh_TW` (known traditional)
> > 2. `zh_SG` (known simplified)
> > 3. `zh_CN`
> >   
> > and check if the last one looks like the first one (which would be wrong) or the second one (which would be correct).
> 
> I will perform this check. If it is doesn't work as we hope it should, I will report it here.

It works as expected.

Nevertheless I think we have to reconsider how availability of a locale should be checked.

A _**valid**_ locale tag on OS X _must_ have a **language**, and _can_ have a **script** or a **region** or _both_. Currently the code just looks at the values of **language** and **region** and _ignores_ the value of the **script** attribute. IMHO this is not enough. If a **script** value was specified by the user, it has to be taken into account as well.

I made a quick check by specifying locale ids that really don't make much sense like `zh-Latn_CN` or `zh-Arab_TW`. In both cases the locale is **available**, but asking for example for a localized month name results in `M05` for `May`. What else could the system try to do in such a case?

If the locale id to be checked for availability specifies a **script**, it should match, just as the **region** must match if it is given. I have adjusted the code accordingly.

vadz:
> It works as expected.

Great, thanks for confirming this!

> Nevertheless I think we have to reconsider how availability of a locale should be checked.
[...]
> If the locale id to be checked for availability specifies a **script**, it should match, just as the **region** must match if it is given. I have adjusted the code accordingly.

I agree, in principle, but what happens if someone asks for "fr_Latn_FR"? There is no such locale under macOS (at least in the list of all available locales), but clearly this should work. It might be not a problem in practice because I don't know why would anybody bother specifying the script for French explicitly, but to be 100% correct we'd need to have the information about the default language script too, and I don't think we have it right now, do we?

utelle:
> > If the locale id to be checked for availability specifies a **script**, it should match, just as the **region** must match if it is given. I have adjusted the code accordingly.
> 
> I agree, in principle, but what happens if someone asks for "fr_Latn_FR"? There is no such locale under macOS (at least in the list of all available locales), but clearly this should work. It might be not a problem in practice because I don't know why would anybody bother specifying the script for French explicitly, but to be 100% correct we'd need to have the information about the default language script too, and I don't think we have it right now, do we?

Typically, a script is only defined if a language can be written in more than one script - like Chinese (simplified and traditional) or Serbian (cyrillic and latin). That is, in practise no one would specify a script for French.

However, I agree that something like **fr-Latn_FR** should work - at least if the operating system interprets it correctly, too. As mentioned OS X happily creates a **zh-Latn_CN** locale, but then provides for example more or less cryptic month and weekday names. I have no idea what else might work in unexpected ways.

I will check whether setting a locale **fr-Latn_FR** really works. If yes, we will have to discuss how to handle such cases. However, IMHO a locale like **fr-Hans_FR** should not be accepted - that is, simply ignoring the script is not a reasonable solution. BTW, we need to check also, if and how this works on Windows. AFAICR we don't accept "constructed" locales, but I don't know whether Windows handles a locale like **fr-Latn_FR** correctly or not.

It is correct that we don't have the information about the default language script available in wxWidgets right now, but in principle this information **is** available and could be added in a similar way as the list of known scripts.

utelle:
> However, I agree that something like **fr-Latn_FR** should work - at least if the operating system interprets it correctly, too. As mentioned OS X happily creates a **zh-Latn_CN** locale, but then provides for example more or less cryptic month and weekday names. I have no idea what else might work in unexpected ways.
> 
> I will check whether setting a locale **fr-Latn_FR** really works. If yes, we will have to discuss how to handle such cases. However, IMHO a locale like **fr-Hans_FR** should not be accepted - that is, simply ignoring the script is not a reasonable solution. BTW, we need to check also, if and how this works on Windows. AFAICR we don't accept "constructed" locales, but I don't know whether Windows handles a locale like **fr-Latn_FR** correctly or not.

I made a quick check under **Windows**. In the table below are the results:

| Locale | Works? | GetName() | Script | Remark |
| :--- | :---: | :--- | :--- | :--- |
| fr-Latn-FR | ❌ | ➖ | Latin | Constructed |
| zh-CN | ✔️  | zh-CN | Simplified Han | |
| zh-Hans-CN | ✔️ | zh-CN | Simplified Han |
| zh-Hant-CN | ❌ | ➖ | Traditional Han | Constructed |
| zh-HK | ✔️  | zh-HK | Traditional Han | |
| zh-Hant-HK | ✔️  | zh-HK | Traditional Han |
| zh-Hans-HK | ✔️  | zh-Hans-HK | Simplified Han | |
| az | ✔️ | az | Latn |  |
| az-AZ | ❌ | ➖ | Latn _or_ Cyrl | Constructed |
| sr | ✔️ | sr | Latn | |
| sr-BA | ❌ | ➖ | Latn _or_ Cyrl | Constructed |
| sr-Latn-BA | ✔️ | sr-Latn-BA | Latn | |
| sr-Cyrl-BA | ✔️ | sr-Cyrl-BA | Cyrl | |
| sr-RS | ❌ | ➖ | Latn _or_ Cyrl | Constructed |
| sr-Latn-RS | ✔️ | sr-Latn-RS | Latn | |
| sr-Cyrl-RS | ✔️ | sr-Cyrl-RS | Cyrl | |
| en-Latn-US | ❌ | ➖ | Latn | Constructed |
| en-FR | ❌ | ➖ | Latn | Constructed (but works under wxOSX) |

As you can see a locale id is rejected as _constructed_, if

- a script is given that is equal to the primary and only script of the language (_fr-Latn-FR_, _en-Latn-US_)
- a script is given that is neither the primary nor the secondary script of a region (_zh-Hant-CN_)
- a script is **not** given for a region with 2 or more primary scripts (_sr-BA_, _sr-RS_)

In the light of the fact that the locale **fr-Latn-FR** is rejected under Windows, I find it legitimate to reject it under OS X, too. That is, IMHO checking the script part for equality under OS X, if it is given, is acceptable.

The general advice would be to specify the script part only, if a language can be written in more than one script. In the latter case it may be even necessary to specify the script, if the language has more than one primary script.

Looking at the lists of supported locales for Windows and for Mac OSX one can observe that an entry with

1. _language_ alone is always included
2. _language_ and _script_ (without _region_) is always included, if a language has more than one script
3. _language_ and _region_ (without _script_) is included, if a language has only one primary script (in the region), but has regional differences
4. _language_ and _script_ and _region_ is included, if a language has more than one script (in the region) and has regional differences

Regarding (3) and (4) _Chinese_ is handled differently by Windows and Mac OSX, unfortunately. While Mac OSX requires the script to be present (but does the right thing, if it is not), Windows allows the script only, if it differs from the primary script of the region.

In principal, we could add logic to drop the script part from a locale tag, if the script corresponds to the primary and only script of the language. However, I doubt that this is really worth the effort.

In practice, an application usually specifies neither script nor region, unless it is really necessary to differentiate (or the operating system - like Linux/Unix - demands a region).

vadz:
I'm going to merge this, but I'd like to squash together most of the commits in this PR, so my plan is to actually update #23556 with all the changes from here, appropriately (re)combined. Please let me know if you have any objections before I start doing this.

Thanks again!

utelle:
No objections. All these small incomplete commits while handling and accepting your comments are inconvenient. Probably the GitHub interface allows to do that in a better way, and I will try to make it better next time.

vadz:
Please see the updated #23556 -- and I'm closing this one, as discussed.

Thanks again!

PR #23566, opened 2023-05-24
Add EVT_FULLSCREEN event macro

bk138:
..that is mentioned in the docs added by
0ff1bdec09e87d58fd7ffcbb1da50531aeb0061a
but whose implementation was not added by said commit.


vadz:
Thanks for noticing and fixing this! I will merge it soon and probably backport to 3.2 too.

Issue #23568, opened 2023-05-24
win32/master/duplicate definition wxStringFormat::Validate


kfsone:
Commit Hash: (from master)

Build System: CMake 3.24, Windows 11, VS 2019 VisualStudio.16.Release/16.11.25+33423.256, en_US

Build Configuration:
with utf8 set to "ON" or "OFF":
```
    & $cmake -B $BuildDir `
        "$(if ("$Generator") { "${Generator}" })" `
        "-DCMAKE_CONFIGURATION_TYPES=Debug;RelWithDebInfo" `
        -DwxBUILD_BENCHMARKS=OFF `
        -DwxBUILD_CXX_STANDARD=17 `
        -DwxBUILD_DEMOS=OFF `
        -DwxBUILD_INSTALL=OFF `
        -DwxBUILD_MONOLITHIC=ON `
        -DwxBUILD_MSVC_MULTIPROC=ON `
        -DwxBUILD_OPTIMISE=ON `
        -DwxBUILD_PRECOMP=ON `
        -DwxBUILD_SAMPLES=OFF `
        -DwxBUILD_SHARED=OFF `
        -DwxBUILD_STRIPPED_RELEASE=ON `
        -DwxBUILD_TESTS=OFF `
        -DwxUSE_EXPAT=builtin `
        -DwxUSE_LIBJPEG=builtin `
        -DwxUSE_LIBLZMA=OFF `
        -DwxUSE_LIBPNG=builtin `
        -DwxUSE_LIBTIFF=builtin `
        -DwxUSE_MEDIACTRL=OFF `
        -DwxUSE_POSTSCRIPT=OFF `
        -DwxUSE_POSTSCRIPT_ARCHITECTURE_IN_MSW=OFF `
        -DwxUSE_PRINTING_ARCHITECTURE=OFF `
        -DwxUSE_RIBBON=OFF `
        -DwxUSE_SOUND=OFF `
        -DwxUSE_SVG=OFF `
        -DwxUSE_STL=OFF `
        -DwxUSE_ZLIB=builtin `
        -DwxUSE_UNICODE=ON `
        -DwxDEBUG_LEVEL=0 `
        "-DwxUSE_UNICODE_UTF8=${utf8}" `
        $SourceDir
```
(complete script: https://gist.github.com/kfsone/12a9b86ce349ade41879b1a01cc3a28e)

We end up with the following link error:

```
[...]

PR #23569, opened 2023-05-24
Inline wxStringFormat Validate in-class

kfsone:
https://github.com/wxWidgets/wxWidgets/issues/23568

wxStringFormat's Validate method was being provided an empty, inline implementation where wxDEBUG_LEVEL did not evalue to true, but outside of the class. This resulted in multiple definitions of the method in different compilation units, which MSVS 2019 did not like.

This change moves the inlined definition to the declaration site with the same predicate.

### DIFF
--- a/include/wx/strvararg.h
+++ b/include/wx/strvararg.h
@@ -208,12 +208,13 @@ public:
     };
 
 #if wxDEBUG_LEVEL
-    // Validate all format string parameters at once: the vector contains the
-    // format specifiers corresponding to the actually given arguments.
+    // Debugging implementation of Validate that will ralidate all format string
+    // parameters at once: the vector contains the format specifiers
+    // corresponding to the actually given arguments.
     void Validate(const std::vector<int>& argTypes) const;
 #else
-    // Also provide a trivial implementation of Validate() doing nothing in
-    // this case.
+    // Alternately, provide a trivial implementation of Validate() doing nothing
+    // in the non-debug case.
     void Validate(const std::vector<int>& WXUNUSED(argTypes)) const
     {
     }
--- a/include/wx/strvararg.h
+++ b/include/wx/strvararg.h
@@ -207,9 +207,17 @@ public:
         Arg_Unknown     = 0x8000     // unrecognized specifier (likely error)
     };
 
+#if wxDEBUG_LEVEL
     // Validate all format string parameters at once: the vector contains the
     // format specifiers corresponding to the actually given arguments.
     void Validate(const std::vector<int>& argTypes) const;
+#else
+    // Also provide a trivial implementation of Validate() doing nothing in
+    // this case.
+    inline void Validate(const std::vector<int>& WXUNUSED(argTypes)) const
+    {
+    }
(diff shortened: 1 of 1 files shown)


include/wx/strvararg.h
```diff
@@ -207,9 +207,17 @@ class WXDLLIMPEXP_BASE wxFormatString
...
     };
 
+#if wxDEBUG_LEVEL
     // Validate all format string parameters at once: the vector contains the
     // format specifiers corresponding to the actually given arguments.
     void Validate(const std::vector<int>& argTypes) const;
+#else
+    // Also provide a trivial implementation of Validate() doing nothing in
+    // this case.
+    inline void Validate(const std::vector<int>& WXUNUSED(argTypes)) const
```
vadz, line 217: This `inline` seems superfluous now, we should be able to remove it, shouldn't we? I.e. write just
```suggestion
    void Validate(const std::vector<int>& WXUNUSED(argTypes)) const
```

include/wx/strvararg.h
```diff
@@ -207,9 +207,17 @@ class WXDLLIMPEXP_BASE wxFormatString
         Arg_Unknown     = 0x8000     // unrecognized specifier (likely error)
     };
 
+#if wxDEBUG_LEVEL
     // Validate all format string parameters at once: the vector contains the
     // format specifiers corresponding to the actually given arguments.
     void Validate(const std::vector<int>& argTypes) const;
+#else
+    // Also provide a trivial implementation of Validate() doing nothing in
```
vadz, line 215: Sorry for the nitpick, but this "Also" doesn't make sense any more after moving the comment. Let's just say
```suggestion
    // Provide a trivial implementation of Validate() doing nothing in
```

vadz (commented):
Thanks for the fix!

I admit I don't fully understand why it didn't work before, it looks like defining it inside or outside the class shouldn't change anything, but perhaps it does because of the DLL export?

Anyhow, let's merge it if it fixes the problem, just please let me know if we can remove this extra `inline`.

Thanks again!

include/wx/strvararg.h
kfsone: Done!

include/wx/strvararg.h
kfsone: Happy to, and actually stopped a minute to look at the entire #if...#else...#endif block and update both comments from the perspective as reading it as one chunk of code rather than two discrete definitions separated by many lines.

kfsone:
FWIW: The standard allows what you did, http://eel.is/c++draft/dcl.inline, so this is possibly a defect in MSVC 2019 (which we're currently using until we can get all our 3rd-party libraries bumped to 2022; hence the script in the issue ticket. It's still a head-scratcher to me that _powershell_ (dot net) turns out to be the best way to write portable ci scripts)

kfsone:
@vadz muscle memory, I did a forced push, I've made the requested changes if you can re-review

kfsone:
I'm closing this pull request, it appears to be a combination of a tool-version specific defect and an issue with correct setting of wxDEBUG_LEVEL that surfaces for us between 3.2.0 and 3.3.0.

PR #23573, opened 2023-05-26
Correct color table size calculation for RLE-compressed bitmaps in wxMSW

BrianNixon:
With `BI_RLE8` and `BI_RLE4` it’s the _bits_ that are compressed, not the _color table_, so the same calculation as for `BI_RGB` should be used.

This isn’t documented on MSDN, but it can be checked with many of the bitmaps embedded as resources in Windows system DLLs, which use RLE and otherwise display wrongly after conversion with `wxDIB::ConvertToBitmap()`.

### DIFF
--- a/include/wx/msw/dib.h
+++ b/include/wx/msw/dib.h
@@ -64,7 +64,7 @@ public:
     bool Load(const wxString& filename);
 
     // dtor is not virtual, this class is not meant to be used polymorphically
-    ~wxDIB();
+    inline ~wxDIB();
 
 
     // operations
@@ -192,10 +192,10 @@ public:
 
 private:
     // common part of all ctors
-    void Init();
+    inline void Init();
 
     // free resources
-    void Free();
+    inline void Free();
 
     // initialize the contents from the provided DDB (Create() must have been
     // already called)
@@ -271,8 +271,6 @@ inline wxDIB::~wxDIB()
     Free();
 }
 
-#endif
-    // wxUSE_WXDIB
+#endif // wxUSE_WXDIB
 
 #endif // _WX_MSW_DIB_H_
-
--- a/src/common/imagbmp.cpp
+++ b/src/common/imagbmp.cpp
@@ -515,8 +515,8 @@ struct BMPDesc
 
 // Read the data in BMP format into the given image.
 //
-// The stream must be positioned at the start of the palette data for the
-// bitmaps using palettes or at the start of the bitmap data otherwise.
+// The stream must be positioned at the start of the bitmap data
+// (i.e., after any palette data)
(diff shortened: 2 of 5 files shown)


vadz:
Thanks for the fix!

It would be really great to have a test case checking that handling these bitmaps works correctly. The simplest would probable be to provide a pair of images, one as RLE{4,8} BMP and another one in a different format (e.g. PNG) but the same contents and add a check that loading them results in the same `wxImage` to this function:

https://github.com/wxWidgets/wxWidgets/blob/563cc11151ee0d986dae8939c4bb24a3a42c0611/tests/image/image.cpp#L1328-L1341

Of course, if you can make another simpler/better test it would be great too, but we really should cover this case in the tests.

TIA!

BrianNixon:
OK – I’ll have a look at adding a test

BrianNixon:
Test added.

This introduces a mechanism for comparing BMPs loaded via `wxBMPHandler` and via the native `CreateDIBitmap()`. This will also be useful for verifying fixes for some other bugs I discovered while writing this test. I will submit PRs for those once you’ve had a chance to review this one.

vadz:
Sorry, but what is 85fd9ec for?

BrianNixon:
> Sorry, but what is 85fd9ec for?

Resolving the MSW cross-build failures: [1](https://github.com/wxWidgets/wxWidgets/actions/runs/5135650698/jobs/9241370381), [2](https://github.com/wxWidgets/wxWidgets/actions/runs/5135650698/jobs/9241370517)

vadz:
> > Sorry, but what is [85fd9ec](https://github.com/wxWidgets/wxWidgets/commit/85fd9ecc1b72267db90d93774d1884bfbbfb923c) for?
> 
> Resolving the MSW cross-build failures: [1](https://github.com/wxWidgets/wxWidgets/actions/runs/5135650698/jobs/9241370381), [2](https://github.com/wxWidgets/wxWidgets/actions/runs/5135650698/jobs/9241370517)

Ah, I see, thanks. I think I'd prefer to just move the code of these functions actually inline, having to declare them as `inline` seems a bit weird.

If you don't mind, I'm going to do it and merge it like this (with some minor cosmetic changes to the new test, e.g. use `std::vector` instead of `wxScopedArray`).

Thanks again!

vadz:
Done in #23584, I'll merge it once the CI builds pass and will also backport d2ee111eb9 to 3.2 later.

BrianNixon:
> use `std::vector` instead of `wxScopedArray`

Does [the coding standard](https://www.wxwidgets.org/develop/coding-guidelines/#-dont-use-many-standard-library-classes) need updating?

> standard container … classes should not be used inside the library

PR #23574, opened 2023-05-26
Samples mfc update

wsu-cb:
In preparation for trying to add wxWidgets to an existing MFC project, I have been examining `samples/mfc`.  It appears to be out-of-date, and I have tried to update it with this PR.

However, I think the last commit (4336e5c3a6284bb4c8dc52faa28e792b38d85821) in this PR should not actually be applied; I argue that it is a work-around to a bug that I don't know how to solve, so I am marking this PR as a draft so it can't be committed.  (Maybe I should have created separate entries for the PR and the bug-report, but the bug can't be demonstrated without the other commits, so I am putting them both here.)

The bug is that keyboard accelerators for MFC windows don't work in `samples/mfc`.  This can be seen by running `samples/mfc` commit 9c75ec6c8822d3d0854214477743fbd9db774eb7 and using the `F1` key on the MFC window.  It should open the About box, but does not.

I can see that `include/wx/msw/mfc.h` `wxMFCApp<T>::PreTranslateMessage()` calls `wxGUIEventLoop::PreProcessMessage()` with the keyboard accelerator message, and that `wxGUIEventLoop::PreProcessMessage()` always reports that it has consumed the keyboard accelerator message.

I can see that `wxGUIEventLoop::PreProcessMessage()` looks for a `wxWindow` associated with the HWND, and can't find one.  That is to be expected since MFC is creating the frame window in this case.  However, when `wxGUIEventLoop` can't find a `wxWindow`, it asks Win32 `::IsDialogMessage()` if it wants to process the keyboard accelerator,  For reasons I do not understand, `::IsDialogMessage()` reports that it did process the keyboard accelerator, and therefore, the MFC never gets to see it.

The workaround commit modifies `samples/mfc` to give MFC a chance to process the keyboard accelerator before calling `wxMFCApp<T>::PreTranslateMessage()` to demonstrate that this re-enables MFC keyboard accelerator handling, but I don't think every wxWidgets+MFC project should be required to do this; I think wxWidgets should take care of the problem. However, I don't know the right way to handle this since I suspect that wxWidgets is sometimes required to forward the keyboard accelerator to `::IsDialogMessage()`, but doing so in this case appears to be the wrong thing to do.

### DIFF
--- a/include/wx/msw/mfc.h
+++ b/include/wx/msw/mfc.h
@@ -94,6 +94,12 @@ public:
     // Override this to provide messages pre-processing for wxWidgets windows.
     BOOL PreTranslateMessage(MSG *msg) override
     {
+        // As reported in issue #23574, wxGUIEventLoop::PreProcessMessage()
+        // is always returning true, so try BaseApp::PreTranslateMessage()
+        // and hope it doesn't always report true
+        if (BaseApp::PreTranslateMessage(msg))
+            return TRUE;
+
         // Use the current event loop if there is one, or just fall back to the
         // standard one otherwise, but make sure we pre-process messages in any
         // case as otherwise many things would break (e.g. keyboard
@@ -103,10 +109,7 @@ public:
         wxGUIEventLoop evtLoopStd;
         if ( !evtLoop )
             evtLoop = &evtLoopStd;
-        if ( evtLoop->PreProcessMessage(msg) )
-            return TRUE;
-
-        return BaseApp::PreTranslateMessage(msg);
+        return evtLoop->PreProcessMessage(msg);
     }
 
     BOOL OnIdle(LONG lCount) override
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for the fixes! The first 3 commits could be indeed already merged, but for the last one I think that if we need to do this, we should do it in `wxMFCApp` itself. It already overrides `PreTranslateMessage()`, so we'd just change the order in which it uses MFC and wx pre-translate filters.

wsu-cb:
I have moved the re-ordering of the MFC and wx filters into `wxMFCApp`.

vadz:
Thanks again, merging this now!

Some of these fixes could probably be backported to 3.2 but I'm not really sure if it is needed -- please let me know if you do need them there too (of course, making a PR against 3.2 would be very welcome too in this case).

wsu-cb:
I am trying to use the current master code, so I don't need 3.2 backporting.

PR #23580, opened 2023-05-30
Create the IE emulation key in the Windows registry if it does not exist

Rezakarimi66:
The related folder of the IE registry key (FEATURE_BROWSER_EMULATION) should be created if it does not exist. Otherwise some of IE features may not work depending on the level set.

### DIFF
--- a/src/msw/webview_ie.cpp
+++ b/src/msw/webview_ie.cpp
@@ -995,7 +995,8 @@ bool wxWebViewIE::MSWSetEmulationLevel(wxWebViewIE_EmulationLevel level)
         wxT("\\FeatureControl\\FEATURE_BROWSER_EMULATION");
 
     wxRegKey key(wxRegKey::HKCU, IE_EMULATION_KEY);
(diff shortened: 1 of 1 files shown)


Rezakarimi66:
For testing, just remove the folder "HKEY_CURRENT_USER\SOFTWARE\Microsoft\Internet Explorer\Main\FeatureControl\FEATURE_BROWSER_EMULATION" and run your program to see whether it creates the key again or not

vadz:
I agree with [this reply](https://groups.google.com/g/wx-dev/c/UEgWQvel0Ww/m/HSq7bSpsAgAJ), it looks like this `||` should definitely be `&&`. Could you please retest this?

Rezakarimi66:
> I agree with [this reply](https://groups.google.com/g/wx-dev/c/UEgWQvel0Ww/m/HSq7bSpsAgAJ), it looks like this `||` should definitely be `&&`. Could you please retest this?

My bad :) I fixed it

PR #23581, opened 2023-05-30
Backport all changes from PR #23556 to 3.2

utelle:
For month and weekday names only the variants for the date/time formatting context are returned. This is wrong for standalone month names like in the generic calendar control. However, in the prior implementation that was also the case.

### DIFF
--- a/include/wx/private/uilocale.h
+++ b/include/wx/private/uilocale.h
@@ -68,6 +68,14 @@ public:
     // The entries contain platform-dependent identifiers.
     static wxVector<wxString> GetPreferredUILanguages();
 
+    // Helper function used by GetMonthName/GetWeekDayName(): returns 0 if flags is
+    // wxDateTime::Name_Full and 1 if it is wxDateTime::Name_Abbr
+    // or -1 if the flags is incorrect (and asserts in this case)
+    //
+    // the return value of this function is used as an index into 2D array
+    // containing full names in its first row and abbreviated ones in the 2nd one
+    static int ArrayIndexFromFlag(wxDateTime::NameFlags flags);
+
     // Use this locale in the UI.
     //
     // This is not implemented for all platforms, notably not for Mac where the
@@ -80,6 +88,8 @@ public:
     virtual wxLocaleIdent GetLocaleId() const = 0;
     virtual wxString GetInfo(wxLocaleInfo index, wxLocaleCategory cat) const = 0;
     virtual wxString GetLocalizedName(wxLocaleName name, wxLocaleForm form) const = 0;
+    virtual wxString GetMonthName(wxDateTime::Month month, wxDateTime::NameFlags flags) const = 0;
+    virtual wxString GetWeekDayName(wxDateTime::WeekDay weekday, wxDateTime::NameFlags flags) const = 0;
     virtual wxLayoutDirection GetLayoutDirection() const = 0;
     virtual int CompareStrings(const wxString& lhs, const wxString& rhs,
                                int flags) const = 0;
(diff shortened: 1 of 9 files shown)


vadz (changes requested):
Thanks for the changes, but there seem to be 2 problems:

1. This broke `RestoreLocale` unit test under Unix. I didn't look into why did this happen, but this needs to be fixed.
2. This somehow also broke the ABI check, and although this might be just a false positive, we still need to somehow avoid it. I don't know how to tell it that `wxUILocaleImpl` is a private class that should be ignored, unfortunately.

utelle:
> Thanks for the changes, but there seem to be 2 problems:
> 
> 1. This broke `RestoreLocale` unit test under Unix. I didn't look into why did this happen, but this needs to be fixed.

At the moment I have no idea why this happens. I hope I can reproduce it locally.

> 2. This somehow also broke the ABI check, and although this might be just a false positive, we still need to somehow avoid it. I don't know how to tell it that `wxUILocaleImpl` is a private class that should be ignored, unfortunately.

Looking at the messages from the ABI check

```
2 member function insertions:
  'method virtual wxString wxUILocaleImpl::GetMonthName(wxDateTime::Month, wxDateTime::NameFlags) const' at uilocale.h:91:1, virtual at voffset 7/10
  'method virtual wxString wxUILocaleImpl::GetWeekDayName(wxDateTime::WeekDay, wxDateTime::NameFlags) const' at uilocale.h:92:1, virtual at voffset 8/10
2 member function changes (8 filtered):
  'method virtual wxLayoutDirection wxUILocaleImpl::GetLayoutDirection() const' has some sub-type changes:
  the vtable offset of method virtual wxLayoutDirection wxUILocaleImpl::GetLayoutDirection() const changed from 7 to 9
note that this is an ABI incompatible change to the vtable of class wxUILocaleImpl
  'method virtual int wxUILocaleImpl::CompareStrings(const wxString&, const wxString&, int) const' has some sub-type changes:
  the vtable offset of method virtual int wxUILocaleImpl::CompareStrings(const wxString&, const wxString&, int) const changed from 8 to 10
  note that this is an ABI incompatible change to the vtable of class wxUILocaleImpl

!!! INCOMPATIBLE ABI changes detected in libwx_baseu-3.2 !!!
```

I get the impression that the order of methods matters. So, maybe adding the new method definitions **after** `GetLayoutDirection()` solves this problem.

utelle:
Well, the ABI check still fails. However, I don't see a clear hint, what exactly is actually causing the failure - although at least the messages `note that this is an ABI incompatible change to the vtable of class wxUILocaleImpl` are gone.

> > 1. This broke `RestoreLocale` unit test under Unix. I didn't look into why did this happen, but this needs to be fixed.
> 
> At the moment I have no idea why this happens. I hope I can reproduce it locally.

In the meantime I ran the tests locally - without problems. The test tries to activate `wxLocale locEn(wxLANGUAGE_ENGLISH);`. `wxLANGUAGE_ENGLISH` maps to locale tag `en_GB`. Could it be that this locale is not available in the test environment? Maybe only `en_US` is available. In that case the locale would not be changed and the prior locale (here the French locale) would be still active.

Looking at the `Set up build environment` step I see the following list of installed locales:

```
de_CH.UTF-8... done
de_DE.UTF-8... done
en_US.UTF-8... done
fr_FR.UTF-8... done
sv_SE.UTF-8... done
```
So, this explains what happens.

In commit https://github.com/wxWidgets/wxWidgets/commit/1cf59d345ba79719f0c143f5ce1c7047e2b3dea7 a check for an empty region was added in method `wxUILocaleImplUnix::TryCreateMatchingLocale()`. Actually, when called via `wxLocale` the region will (almost) never be empty, because usually always the locale tag with region will be taken from the language database.

I now removed the check for the empty region. This should fix the Unix related problem in the checks.

vadz:
I've squash-merged this, added `version-script.in` changes and, most importantly, used a horrible hack to avoid adding virtual functions to `wxUILocaleImpl` in #23747 -- and this now passes the ABI check, so I'm closing this one and will merge the other one if you don't have any objections.

Issue #23579, opened 2023-05-30
Problem with opengl/penguin sample


gerard-durand:
I have updated and downloaded yesterday, via git, wxwidgets 3.3. It compiles correctly with last gcc 13.1 found with msys2 packages.
Nevertheless, when I try the penguin opengl sample after rebuilding, it crashes : I obtain a gray empty window for one or 2 seconds, then the window is closed without displaying the penguin. Other opengl samples works as expected.
If I replace dxfrenderer.cpp and .h with previous versions (as distributed with 3.2.2.1 wxwidgets), it's OK. The sample works as expected.
I don't know if it is a new dxfrenderer problem or an incompatibility with gcc 13.1. I think that with gcc 12.2, I had no such problems, but I don't remember if I had verified this sample test with that compiler version.

I'm on Window 10 64 bit, last updated version. I compiled with option CXXFLAGS="-std=gnu++20".

oneeyeman1:
@gerard-durand ,
Can you show a backtrack at the  time of the crash?

Thank you.

gerard-durand:
I'm not familiar with backtracing, but this is what I obtain by modifying my make command with CXXFLAGS="-g -std=gnu++20" CFLAGS="-g" and entering in a command line : gdb penguin.exe. The wxwidgets version 3.3.0 is still the release one, not a debug one.

[Backtrace.txt](https://github.com/wxWidgets/wxWidgets/files/11600777/Backtrace.txt)

PBfordev:
I can confirm the issue, I get an assert, happens with both GCC and MSVC. The output from MSVC is:
> D:\Dev\Desktop\!Lib\wxWidgets-PB\samples\opengl\penguin\dxfrenderer.cpp(320): assert "col >= 0 && col <= 255" failed in ACIColourToRGB().

Relevant call stack:
```
penguin.exe!ACIColourToRGB(int col=1084227584) Line 320	C++	Symbols loaded.
penguin.exe!DXFRenderer::Render() Line 649	C++	Symbols loaded.
penguin.exe!TestGLCanvas::OnPaint(wxPaintEvent & __formal={...}) Line 195	C++	Symbols loaded.
```

It does not happen with 3.2.2.1 so perhaps this is the culprit: https://github.com/wxWidgets/wxWidgets/commit/b76ebc61d9817d286ecb38b9ecacc32c7b92f30c (I did not look into this).

vadz:
[...]

PR #23582, opened 2023-05-30
Fix storing DXF entities in penguin sample

PBfordev:
This is a not very pretty but seemingly working fix for #23579.

### DIFF
--- a/samples/opengl/penguin/dxfrenderer.cpp
+++ b/samples/opengl/penguin/dxfrenderer.cpp
@@ -338,8 +338,6 @@ void DXFRenderer::Clear()
 {
     m_loaded = false;
     m_layers.clear();
-    for (auto e : m_entities )
-        delete e;
     m_entities.clear();
 }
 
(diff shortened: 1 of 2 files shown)


vadz:
Oops, thank you, and sorry for completely missing this both when making my changes in b76ebc61d9817d286ecb38b9ecacc32c7b92f30c and now. Storing pointers is definitely the right thing to do if the objects are actually polymorphic.

It's annoying that we don't require C++14 and so can't use `std::make_unique<>`. I wonder if we could define our own `wx::make_unique<>` or something like this and use it instead? If not, we need to revert to writing `new` explicitly here...

PBfordev:
> > Sorry, I thought 3.3 allowed C++14: Is this documented anywhere?
> 
> It's mentioned in the README but we should definitely add this to the requirements somewhere...

Sorry, I missed that (and perhaps the latest supported GCC version listed there should be bumped from 12 to 13; clang seems to be at v16 now too)


> > Could not find it in the Coding Guidelines (which still also have section "Don’t use many standard library classes")
> 
> This entire document is hopelessly outdated.

Mhm, the last four updates of that document are mine, including "modern C++" update from the last fall, so hopefully it is not outdated that hopelessly.

vadz:
Thanks again for fixing this, will merge soon.

PR #23584, opened 2023-05-31
Fix loading RLE bitmaps in wxMSW

vadz:
This is just a slightly repackaged version of #23573.

### DIFF
--- a/tests/image/image.cpp
+++ b/tests/image/image.cpp
@@ -32,6 +32,14 @@
 #include "wx/clipbrd.h"
 #include "wx/dataobj.h"
 
+// Check if we can use wxDIB::ConvertToBitmap(), which only exists for MSW and
+// which assumes the target is little-endian (matching the file format)
+#if defined(__WXMSW__) && wxUSE_WXDIB && wxBYTE_ORDER == wxLITTLE_ENDIAN
+    #define CAN_LOAD_BITMAP_DIRECTLY
+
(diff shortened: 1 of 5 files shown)


tests/image/image.cpp
```diff
@@ -1294,6 +1302,74 @@ TEST_CASE_METHOD(ImageHandlersInit, "wxImage::BMPFlippingAndRLECompression", "[i
...
+    wxImage image2(filename);
+    REQUIRE( image2.IsOk() );
+
+    INFO("Comparing loading methods for " << filename);
+    CompareImage(*wxImage::FindHandler(wxBITMAP_TYPE_BMP), image1, 0, &image2);
+}
+
+TEST_CASE_METHOD(ImageHandlersInit, "wxImage::BMPLoadMethod", "[image][bmp]")
+{
+    CompareBMPImageLoad("image/horse_grey.bmp");
```
BrianNixon, line 1367: This is harmless, but it doesn’t actually test the change since `horse_grey.bmp` isn’t RLE-compressed

vadz: Yes, I've added it just as a sanity test. I.e. I've actually started by applying your tests commit, which failed, as expected, then added this line to make sure it succeeded with non-RLE bitmaps and then applied your actual fix.

As you say it's harmless, and I think it's better to leave it here, just in case we manage to break non-RLE case in the future.

PR #23591, opened 2023-06-01
Remove redundant "const" from wxItemId conversion operator

### LINKED ISSUE #23590: After upgrading to VS 17.6, including itemid.h breaks
With warnings as errors, and most warnings turned on (/Wall, with as small a list of possible of warnings that are ignored), VS17.6 now complains:

>d:\devtools\wx\wxWidgets-3.2.2\include\wx\itemid.h(36,5): warning C5266: 'const' qualifier on return type has no effect

Windows 10, wx3.2.2.1.

vadz:
This "const" is ignored and just results in a warning from MSVS 17.6.

Closes #23590.

### DIFF
--- a/include/wx/itemid.h
+++ b/include/wx/itemid.h
@@ -33,7 +33,7 @@ public:
 
     bool IsOk() const { return m_pItem != nullptr; }
     Type GetID() const { return m_pItem; }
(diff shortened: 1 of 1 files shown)


vadz:
Merged as 946aa6bbc430302265aea41754c13729f931edf4

PR #23598, opened 2023-06-02
Fix `wxSocketImpl::RecvDgram()` when caller requests fewer bytes than are present in datagram

BrianNixon:
On Windows, when the number of bytes present in a datagram is less than what was requested to read, `recvfrom()` will return `SOCKET_ERROR` (not the short count, as on other platforms) and `WSAGetLastError()` will return `WSAEMSGSIZE`. This patch fixes `wxSocketImpl::RecvDgram()` to handle that case.

### DIFF
--- a/build/cmake/tests/base/CMakeLists.txt
+++ b/build/cmake/tests/base/CMakeLists.txt
@@ -50,6 +50,7 @@ set(TEST_SRC
     misc/module.cpp
     misc/pathlist.cpp
     misc/typeinfotest.cpp
+    net/dgramsocket.cpp
     net/ipc.cpp
     net/socket.cpp
     net/webrequest.cpp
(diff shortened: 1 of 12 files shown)


vadz (commented):
Thanks a lot for fixing this and even providing the test case!

I have two questions:

1. Is the new `wxSOCKET_MSGSIZE` supposed to be part of public API? If so, we need to document it, but I'd much prefer to avoid exposing it. Do you see some way we could avoid adding it to the public header somehow?
2. Do we really need a new file for the new test? I'd just add it to `tests/net/socket.cpp` to be honest. But this is a minor point and if you prefer to have a new file, we can keep it too, of course, especially as you've already done all the work for adding it everywhere.

BrianNixon:
1. No: `wxSOCKET_MSGSIZE` doesn’t have to be public, as long as you don’t mind adding platform-specific code to `common/​socket.cpp` to hide the differences in the behaviour of `recvfrom()`.

2. No: we don’t _need_ a new file; it just seemed better to separate lower-level socket tests from the higher-level client/​server stuff that the existing `socket.cpp` focuses on.

It’s probably simpler to discard this branch and start a new one with the much smaller set of changes, so look out for that shortly…

BrianNixon:
Closing in favour of #23604

PR #23599, opened 2023-06-02
Correct delta-RLE bitmap background colour

BrianNixon:
Bitmaps do not necessarily specify every pixel explicitly: the RLE delta escape sequence allows offsetting the current pixel position, so there is an implicit background colour. Windows sets it to black, regardless of whether black is in the image’s colour table. This patch makes wxWidgets match that behaviour (previously it used the first entry in the image’s colour table).


src/common/imagbmp.cpp
```diff
@@ -546,7 +546,11 @@ bool LoadBMPData(wxImage * image, const BMPDesc& desc,
 
     // destroy existing here instead of:
     image->Destroy();
-    image->Create(width, height);
+    // It's important to clear the image because bitmaps do not necessarily
+    // specify every pixel explicitly: the RLE delta escape sequence allows
+    // offsetting the current pixel position, so there is an implicit
+    // background colour. Windows sets it to black, so match that here.
+    image->Create(width, height, true);
```
vadz, line 553: Could it be a worthwhile optimization to do this only when using `BI_RLEn`? I guess clearing even big bitmaps must be pretty fast on modern hardware, but we still don't _need_ to do it in the non-RLE case, which is probably much more common.

I.e. should we do
```suggestion
    image->Create(width, height, desc.comp == BI_RLE4 || desc.comp == BI_RLE8);
```
?

vadz (commented):
Thanks for fixing this!

We can merge this as is but please let me know if my suggestion below has merit.

src/common/imagbmp.cpp
BrianNixon: Yes: that should be fine. I hadn’t changed it as I was afraid of breaking the previous behaviour – but it does appear that in every case other than RLE the routine sets every pixel of the destination image, so clearing it first is a waste of time.

Change committed

vadz:
Thanks for the update! Will (squash) merge this soon.

PR #23600, opened 2023-06-02
Fix wxDC and wxGCDC clippings in wxQt

AliKet:
With this PR all clipping tests are passing successfully now, which wasn't the case before!

### DIFF
--- a/src/common/dcgraph.cpp
+++ b/src/common/dcgraph.cpp
@@ -452,15 +452,6 @@ void wxGCDCImpl::DoSetDeviceClippingRegion( const wxRegion &region )
 void wxGCDCImpl::DestroyClippingRegion()
 {
     m_graphicContext->ResetClip();
-#if !defined(__WXOSX__) && !defined(__WXMSW__) && !defined(__WXGTK__) && !defined(__WXQT__)
-    // currently the clip eg of a window extends to the area between the scrollbars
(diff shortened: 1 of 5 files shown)


src/common/dcgraph.cpp
```diff
@@ -452,7 +452,7 @@ void wxGCDCImpl::DoSetDeviceClippingRegion( const wxRegion &region )
 void wxGCDCImpl::DestroyClippingRegion()
 {
     m_graphicContext->ResetClip();
-#if !defined(__WXOSX__) && !defined(__WXMSW__) && !defined(__WXGTK__)
+#if !defined(__WXOSX__) && !defined(__WXMSW__) && !defined(__WXGTK__) && !defined(__WXQT__)
```
vadz, line 455: Do we really need this code? AFAICS the only still used port not tested here is wxX11 but it has so many other problems that I'm not sure it's worth to cater to it specifically here.

vadz (commented):
Very nice, thanks! I can merge this already, but if we can remove the now almost never used code in `wxGCDCImpl::DestroyClippingRegion()`, let's do it too.

Thanks again!

src/common/dcgraph.cpp
AliKet: Done.

PR #23603, opened 2023-06-03
Restore setting CRT locale in wxLocale under macOS

### LINKED ISSUE #23557: wxDateTime::Format() et al fail to properly format localized dates on macOS
There is an issue with how `wxLocale` operates when using the system default language in wx 3.2 and newer. This issue manifests in such a way that day and month names are represented in English in e.g. the output of `wxDateTime::Format()`, `wxStrftime()`, and consequently the generic calendar control.

To Reproduce:
1. Overwrite `samples/internat/internat.cpp` by copying `internat.cpp` *from the 3.0 branch*.
2. Add e.g. `SetTitle(wxDateTime::Now().Format("%A %B %x %X %c"));` to the beginning of `MyFrame::MyFrame()` body.
3. Build the sample and run it *via Finder* (not via command line).
4. Choose "System default" as the language. (On my system, language/country is German/Germany during this test.)
5. Observe that the date in the title contains day and month names in English.
6. Observe that numerical parts of the output are in the expected localized format.

The key here is to select the system default language; when explicitly selecting something else from the list (even if it is the same as the actual system language), formatting does work.

- wxWidgets version  you use: 3.2 branch
- OS  and its version: macOS 12.6.5

vadz:
This was accidentally broken by the previous changes and the C-level locale was not set at all any more, even when using wxLocale (and not just wxUILocale, which is not supposed to set it) under macOS.

Fix this by explicitly calling setlocale() on this platform to restore the previous, and documented, behaviour.

Closes #23557.

---

@utelle @lanurmi Please let me know if you have any comments about this change. If not, I'll merge it to both master and 3.2.

### DIFF
--- a/src/common/intl.cpp
+++ b/src/common/intl.cpp
@@ -448,8 +448,16 @@ bool wxLocale::Init(int lang, int flags)
     }
 
     // Under (non-Darwn) Unix wxUILocale already set the C locale, but under
-    // the other platforms we still have to do it here.
-#if defined(__WIN32__) || defined(__WXOSX__)
+    // the other platforms we still have to do it.
+#if defined(__WXOSX__)
+    // Locale-related environment variables are typically not set at all under
+    // macOS, so don't rely on them unless we really have no other choice, i.e.
+    // we don't recognize the locale ourselves.
+    if ( !(info ? info->TrySetLocale() : wxSetlocale(LC_ALL, "")) )
+    {
+        ok = false;
+    }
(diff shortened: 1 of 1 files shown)


utelle:
> @utelle @lanurmi Please let me know if you have any comments about this change. If not, I'll merge it to both master and 3.2.

AFAICT the change is ok.

vadz:
Merged now, thanks!

Issue #23594, opened 2023-06-02
wxSocketBase::Peek() works like Read(), bytes become unavailable


GeorgyStolbovoy:
### Description

I am trying to communicate with a server application over UDP protocol. This application responds sequentially in several chunks. Each chunk has some important data and much more. These important data must be extracted before reading the whole chunk. For this purpose I used wxSocket::Peek(). As it turned out, the subsequent call to wxSocket::Read() reads the next chunk! The chunk that Peek() was looking at just got lost! Parts of my code:
```c++
// Initialization in constructor
	Bind(wxEVT_SOCKET, &Recieve, this, socket_id);
	wxIPV4address addr;
	addr.Hostname(wxT("localhost"));
	addr.Service(16801);
	socket = new wxDatagramSocket(addr);
	socket->SetEventHandler(*this, socket_id);
	socket->SetNotify(wxSOCKET_INPUT_FLAG);
	socket->Notify(true);
// ...
// Responce handler
void Frame::Recieve(wxSocketEvent& event)
{
	std::vector<uint8_t> packet;
	uint8_t peek[7];
	socket->Peek(&peek, 7); // Here
	uint16_t w = *reinterpret_cast<uint16_t*>(&peek[3]); // width and height
	uint16_t h = *reinterpret_cast<uint16_t*>(&peek[5]); // of a image
	packet.resize(7 + w*h*2);
	for (uint8_t i(0); i < peek[1]; ++i) // peek[1] is the number of chunks
	{
		socket->Read(packet.data(), packet.size()); // OOPS, this is next chunk
// ...
``` 
If Peek() is not used, then Read() will read the first chunk. If I call Peek() twice, then Read() will read the third chunk. Thus, Peek() changes the state, although it should not do this!

### Platform and version information

- wxWidgets version:  3.2.2.1
- wxWidgets port: wxMSW
- OS: Windows 10

BrianNixon:
On the one hand: this is how the datagram protocol is meant to work. When a single `recvfrom()` does not consume an entire datagram, the remainder is lost.

On the other hand: wx tries to present datagram sockets as though they are streams. To avoid this data loss, `Peek()` on a datagram socket needs to use a buffer big enough to capture the largest possible datagram, not just the number of bytes the caller requested.

BrianNixon:
Here’s a repro. Note that the `LastCount()` checks will fail without #23598; the problem being reported here is the failing `recvbuf[1] == sendbuf1[1]`.

[...]

PR #23604, opened 2023-06-03
Fix peeking and reading datagram sockets when caller requests fewer bytes than are present

BrianNixon:
Simpler version of #23598, plus a fix for #23594


vadz:
Thanks a lot! Merging this into master and it should be backported to 3.2 too later if no problems are found with it.

PR #23607, opened 2023-06-04
Clear grid attribute cache when loading new table

BrianNixon:


BrianNixon:
The test is failing on Mac. Does there need to be something to wait for the update/​paint to complete before the check?

vadz:
Sorry, I don't know what's going on there, but GUI tests in general just don't work under Mac.

Unless you have some idea about fixing it, let's disable/skip this test under Mac as it's already done for many others.

BrianNixon:
I don’t know the first thing about Mac, and don’t have one to learn and test on, and today is not the day to change that…

Test disabled

PR #23613, opened 2023-06-05
CMake: Check and enable use of libxkbcommon

MaartenBent:
This was only implemented for configure, see #23410

### DIFF
--- a/build/cmake/init.cmake
+++ b/build/cmake/init.cmake
@@ -614,7 +614,19 @@ if(wxUSE_GUI)
             wx_option_force_value(wxUSE_CAIRO OFF)
         endif()
     endif()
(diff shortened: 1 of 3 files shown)


vadz:
Thanks! I will merge this in master soon (AFAIR this is not needed in 3.2, is it?).

MaartenBent:
I think 3.2 too, adding xkbcommon via configure is also backported (https://github.com/wxWidgets/wxWidgets/commit/87ae8d5aa7cbeddeb8ead289117c0e114fd72329).

vadz:
Oops, sorry for forgetting that I had backported this. I'll backport this one too then, thanks!

PR #23614, opened 2023-06-05
Customize toolbar on OSX

oneeyeman1:
This is an attempt to introduce the toolbar customization functionality into the library.

However, I have couple of questions:

1. Should this functionality be turned on based on the style/flag?
2. According to the Apple documentation it is recommended to store the new toolbar configuration in the user preferences, so that it can be read it from on the next restart. Should this be turned on by some flag?
3. I don't know if there should be any unit test for this, but I'd like to demonstrate it in the sample? What's the best way to do that? I'm thinking to turn off some buttons and show the customization dialog that will allow to drop the tool on the toolbar. Which one would you recommend?

Please review and give me your thoughts. Only OSX implementation is here - it is supported natively by OS.

MSW will follow and then GTK.

On OSX it is available with the right click and context menu selection.
On MSW it will be available thru the double click.
On GTK I'm thinking to follow OSX way as the GtkToolbar have a "context-menu" signal. And in GTK4 it probably can be done the same way on the GtkBox.

(This PR is not ready to be applied, unless you want it to get some testing on OSX first).

Thank you.


vadz:
I have 2 comments about this:

1. Please test your PRs in your own fork of wxWidgets when they haven't even reached compilable state yet, it's annoying to have to wait until CI builds for this PR finish before being able to run any other ones. In the meanwhile at least switch this one to the "Draft" state until it's really finished.
2. Adding `bool available` parameter is not the right way to do whatever you want to do. Please don't do this, you should be able to add some `wxToolBarTool::MarkAsAvailable()` or whatever instead.

oneeyeman1:
@vadz,
> I have 2 comments about this:
> 
> 1. Please test your PRs in your own fork of wxWidgets when they haven't even reached compilable state yet, it's annoying to have to wait until CI builds for this PR finish before being able to run any other ones. In the meanwhile at least switch this one to the "Draft" state until it's really finished.

I don't have clang installed anywhere but on OSX, so I can't test it under this compiler.
But thank you for pointing out the "draft" switch. Now marked.

How do I unmark it from that? Should I just click on the "work-needed" link?

> 2. Adding `bool available` parameter is not the right way to do whatever you want to do. Please don't do this, you should be able to add some `wxToolBarTool::MarkAsAvailable()` or whatever instead.

OK, thx.
Will do that. Probably name it "MarkAvailable()".
Let me know if the name is not up to coding standard.

Thx as always.

oneeyeman1:
@vadz,
I think it is now ready.

It would be nice if Stefan also could take a look.

Thx.

include/wx/control.h
```diff
@@ -46,7 +46,8 @@ class WXDLLIMPEXP_CORE wxControlBase : public wxWindow
 
     // get the control alignment (left/right/centre, top/bottom/centre)
     int GetAlignment() const { return m_windowStyle & wxALIGN_MASK; }
-
+    void MarkAvailable(bool available) { m_available = available; }
+    bool IsAvailable() const { return m_available; }
```
vadz, line 50: We shouldn't need to add functions that only make sense for toolbar tools to base `wxControl` class. You will have to change the code to use them in `wxToolBatTool` only.

include/wx/event.h
```diff
@@ -724,6 +724,7 @@ wxDECLARE_EXPORTED_EVENT(WXDLLIMPEXP_CORE, wxEVT_COMBOBOX, wxCommandEvent);
 wxDECLARE_EXPORTED_EVENT(WXDLLIMPEXP_CORE, wxEVT_TOOL_RCLICKED, wxCommandEvent);
 wxDECLARE_EXPORTED_EVENT(WXDLLIMPEXP_CORE, wxEVT_TOOL_DROPDOWN, wxCommandEvent);
 wxDECLARE_EXPORTED_EVENT(WXDLLIMPEXP_CORE, wxEVT_TOOL_ENTER, wxCommandEvent);
+wxDECLARE_EXPORTED_EVENT(WXDLLIMPEXP_CORE, wxEVT_TB_CUSTOMIZE, wxCommandEvent);
```
vadz, line 727: Just `TB` is really not clear enough without any context.

```suggestion
wxDECLARE_EXPORTED_EVENT(WXDLLIMPEXP_CORE, wxEVT_TOOLBAR_CUSTOMIZE, wxCommandEvent);
```

include/wx/osx/toolbar.h
```diff
@@ -89,9 +89,11 @@ class WXDLLIMPEXP_CORE wxToolBar: public wxToolBarBase
                                           wxObject *clientData = nullptr,
                                           const wxString& shortHelp = wxEmptyString,
                                           const wxString& longHelp = wxEmptyString) override;
+
     virtual wxToolBarToolBase *CreateTool(wxControl *control,
                                           const wxString& label) override;
 
+    void *GetMacToolbar() { return m_macToolbar; };
```
vadz, line 96: The prevailing convention is to use `OSX` prefix, i.e.
```suggestion
    void *OSXGetToolbar() { return m_macToolbar; };
```

include/wx/osx/toolbar.h
```diff
@@ -89,9 +89,11 @@ class WXDLLIMPEXP_CORE wxToolBar: public wxToolBarBase
                                           wxObject *clientData = nullptr,
                                           const wxString& shortHelp = wxEmptyString,
                                           const wxString& longHelp = wxEmptyString) override;
+
```
vadz, line 92: Please try to avoid whitespace-only changes (even if you make them accidentally, nothing prevents you from looking at the diff and removing them afterwards).

include/wx/tbarbase.h
```diff
@@ -314,7 +319,7 @@ class WXDLLIMPEXP_CORE wxToolBarBase : public wxControl
                                const wxString& shortHelp = wxEmptyString,
                                wxItemKind kind = wxITEM_NORMAL)
     {
-        return AddTool(toolid, label, bitmap, wxBitmapBundle(), kind, shortHelp);
+        return AddTool(toolid, label, bitmap, wxBitmapBundle(), kind, shortHelp, wxEmptyString, nullptr);
```
vadz, line 322: Why?

interface/wx/toolbar.h
```diff
@@ -159,6 +159,19 @@ class wxToolBarToolBase : public wxObject
 
     void SetDropdownMenu(wxMenu *menu);
     wxMenu *GetDropdownMenu() const;
+
+    /**
+     Mark the tool to be not visible.
```
vadz, line 164: This is pretty misleading for a function called like this. You should say it may be used to mark the tool as being not currently visible if `@c available is @false` or something like that.

interface/wx/toolbar.h
```diff
@@ -878,6 +900,25 @@ class wxToolBar : public wxControl
...
+    /**
+        Return whether the tool is displayed by default or not
+        By default all tools are displayed.
+
+        @since 3.3
+    */
+    bool IsAvailable() const;
+
+    /**
+        Calling this function with the parameter set to false will not add the tool to the toolbar
```
vadz, line 913:
```suggestion
        Calling this function with the parameter set to false prevents the tool from being added to the toolbar.
```

samples/toolbar/toolbar.cpp
```diff
@@ -470,6 +470,7 @@ void MyFrame::PopulateToolbar(wxToolBarBase* toolBar)
         combo->Append("combobox with extremely, extremely, extremely, extremely long label");
         combo->Append("in a");
         combo->Append("toolbar");
+        combo->MarkAvailable( false );
```
vadz, line 473: Please create another initially unavailable control instead of hiding this one.

src/osx/iphone/toolbar.mm
vadz, line 1: This file is not changed at all...

interface/wx/toolbar.h
```diff
@@ -271,6 +284,15 @@ class wxToolBarToolBase : public wxObject
         Process a @c wxEVT_TOOL_DROPDOWN event. If unhandled,
         displays the default dropdown menu set using
         wxToolBar::SetDropdownMenu().
+    @event{EVT_TB_CUSTOMIZE(id)}
+        Process a @c wxEVT_TB_CUSTOMIZE event. Under MSW and OSX
+        it will display native customization dialogs. On GTK it will
+        be generic.
```
vadz, line 290: This doesn't seem to be true, there is no implementation at all for MSW and just a stub for GTK. This is not necessarily a problem, but the documentation shouldn't like and you should check that the API you propose is at least in principle implementable under the other platforms.

vadz (commented):
We need to check if this can be made to work elsewhere. Providing at least a generic implementation would be a good way to ensure we can do it, as long as this is only supported under Mac I'm not sure if we can use such API.

oneeyeman1:
@vadz ,
> We need to check if this can be made to work elsewhere. Providing at least a generic implementation would be a good way to ensure we can do it, as long as this is only supported under Mac I'm not sure if we can use such API.

On MSW it is possible to customize toolbar.

GTK does not support that and apparently the devs there are objecting to make it work.

And of course generic implementation we can do whatever we want,
We already have an API to do customization in the generic way..

oneeyeman1:
Also, I started with OSX, because I am not very familiar with OSX.

I will follow you review and apply where appropriate.

However, I'd like to ask you - if this will become mergeable will it be easier to do the MSW here or merge it and then change MSW?

Let me know.

Thx.

PR #23615, opened 2023-06-06
Include macOS 14 name in wxGetOsDescription()

TcT2k:
(Could probably be backported to 3.2)


vadz:
Thanks for the update! I will merge it soon and backport it to 3.2 too.

PR #23618, opened 2023-06-06
3.2 proposed backports (June 2023 edition)

vadz:


Kumazuma:
Can you add ff7476fcd0f3fccad19b849511b73cbfd9a317bd? It seem to be able backport.

vadz:
> Can you add [ff7476f](https://github.com/wxWidgets/wxWidgets/commit/ff7476fcd0f3fccad19b849511b73cbfd9a317bd)? It seem to be able backport.

Sorry, I'm not sure if it's really appropriate: this will change the default bitmap size from 16x15 to 16x16 which might be noticeable and it looks like a wrong thing to do in the stable branch. I could add `FromDIP()`, I guess, as this shouldn't break anything, would this be useful?

a-wi:
I think it would be good to backport 57a9247889ce7501321ef9b69171610582a482e3 - with `#define` (used in 3.2) instead of `constexpr`.

vadz:
@a-wi I don't think we can guarantee that the change of its numeric value doesn't break anything in the existing code, can we? E.g. just storing it (as a numeric value) in the config file and reading it back wouldn't work any more (yes, I agree that it's not the right thing to do, but people still do it and similar things all the time).

docs/changes.txt
```diff
@@ -271,6 +272,10 @@ wxGTK:
 - Fix loading WebKit2 extension when using wxWebView (Scott Talbert, #23497).
 - Fix editing cells without values in wxDataViewCtrl (#23523).
 - Fix uninitialized wxKeyEvent::m_isRepeat (jolz, #23593).
+- Also improve size computation of wxTextCtrl (Alex Shvartzkop, #23610).
```
taler21, line 275: "Also improve" should be replaced by just "Improve" because after moving the changelog entries to the correct place, there is no longer any context for "Also".

taler21:
In the current changelog of the 3.2 branch, the entry for the hotfix version 3.2.2.1 is missing.
I think this should be added for completeness.

src/common/combocmn.cpp
```diff
@@ -2367,6 +2407,14 @@ void wxComboCtrlBase::HidePopup(bool generateEvent)
     OnPopupDismiss(generateEvent);
 }
 
+void wxComboCtrlBase::OnPopupDestroy()
+{
+    m_winPopup = nullptr;
+
+    delete m_popupInterface;
+    m_popupInterface = nullptr;
```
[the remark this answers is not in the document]
vadz: Yes, sorry, this slipped through. I really have to set up a local hook checking for this, to avoid running all the CI builds just to find this out...

docs/changes.txt
vadz: Indeed, thanks for noticing this.

vadz:
> In the current changelog of the 3.2 branch, the entry for the hotfix version 3.2.2.1 is missing. I think this should be added for completeness.

Indeed, thanks for noticing this (and for the review in general!), added now.

Kumazuma:
> I could add FromDIP(), I guess, as this shouldn't break anything, would this be useful?
Yes, it is.

vadz:
> > I could add FromDIP(), I guess, as this shouldn't break anything, would this be useful?
> 
> Yes, it is.

Done now in a2c6620907259ab006ada23d31446cf198e4c887.

Issue #23625, opened 2023-06-09
warning: definition of implicit copy constructor for


rmpowell77:
### Description

Encountering build warnings when building with -Wall -Wextra with C++20.

```
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=20 -DCMAKE_CXX_FLAGS="-Wall -Wextra" -DwxBUILD_SAMPLES=ALL ..
cmake --build . --target auidemo -- -j6
```

```
[ 96%] Building CXX object libs/html/CMakeFiles/wxhtml.dir/__/__/__/__/src/html/helpfrm.cpp.o
In file included from /Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/src/aui/framemanager.cpp:24:
/Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/aui/framemanager.h:676:5: warning: definition of implicit copy constructor for 'wxAuiManagerEvent' is deprecated because it has a user-declared copy assignment operator [-Wdeprecated-copy]
    wxDECLARE_DYNAMIC_CLASS_NO_ASSIGN(wxAuiManagerEvent);
    ^
/Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/rtti.h:142:5: note: expanded from macro 'wxDECLARE_DYNAMIC_CLASS_NO_ASSIGN'
    wxDECLARE_NO_ASSIGN_CLASS(name);                                          \
    ^
/Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/defs.h:3053:20: note: expanded from macro 'wxDECLARE_NO_ASSIGN_CLASS'
        classname& operator=(const classname&) wxMEMBER_DELETE
                   ^
/Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/aui/framemanager.h:649:50: note: in implicit copy constructor for 'wxAuiManagerEvent' first required here
    wxEvent *Clone() const override { return new wxAuiManagerEvent(*this); }
                                                 ^
```

Looks very similar to #22291 .

### Platform and version information

- wxWidgets version  you are building: master, @ db09adf6b5
- wxWidgets port  you are building: wxOSX
- OS  and its version: macOS 13.4 (22F66)
- Compiler  being used: AppleClang 14.0.3.14030022

[build.log](https://github.com/wxWidgets/wxWidgets/files/11702059/build.log)

PR #23626, opened 2023-06-09
Fixing up implicit copy constructor warnings by using wxDECLARE_DYNAM…

rmpowell77:
…IC_CLASS_NO_ASSIGN_DEF_COPY/wxDECLARE_NO_ASSIGN_DEF_COPY.

See #23625

Attaching original build logs and build logs with the fix.
[build_with_fix.log](https://github.com/wxWidgets/wxWidgets/files/11702089/build_with_fix.log)
[build.log](https://github.com/wxWidgets/wxWidgets/files/11702090/build.log)

### DIFF
--- a/include/wx/event.h
+++ b/include/wx/event.h
@@ -3522,7 +3522,7 @@ struct WXDLLIMPEXP_BASE wxEventTableEntry : public wxEventTableEntryBase
     const int& m_eventType;
 
 private:
-    wxDECLARE_NO_ASSIGN_CLASS(wxEventTableEntry);
+    wxDECLARE_NO_ASSIGN_DEF_COPY(wxEventTableEntry);
 };
 
 // an entry used in dynamic event table managed by wxEvtHandler::Connect()
--- a/include/wx/aui/auibar.h
+++ b/include/wx/aui/auibar.h
@@ -98,7 +98,7 @@ private:
     int m_toolId;
 
 private:
-    wxDECLARE_DYNAMIC_CLASS_NO_ASSIGN(wxAuiToolBarEvent);
+    wxDECLARE_DYNAMIC_CLASS_NO_ASSIGN_DEF_COPY(wxAuiToolBarEvent);
 };
 
 
--- a/include/wx/aui/auibook.h
+++ b/include/wx/aui/auibook.h
@@ -79,7 +79,7 @@ private:
 
 #ifndef SWIG
 private:
-    wxDECLARE_DYNAMIC_CLASS_NO_ASSIGN(wxAuiNotebookEvent);
+    wxDECLARE_DYNAMIC_CLASS_NO_ASSIGN_DEF_COPY(wxAuiNotebookEvent);
 #endif
 };
 
--- a/include/wx/aui/framemanager.h
+++ b/include/wx/aui/framemanager.h
@@ -673,7 +673,7 @@ public:
 
 #ifndef SWIG
 private:
-    wxDECLARE_DYNAMIC_CLASS_NO_ASSIGN(wxAuiManagerEvent);
+    wxDECLARE_DYNAMIC_CLASS_NO_ASSIGN_DEF_COPY(wxAuiManagerEvent);
 #endif
 };
 
--- a/include/wx/osx/core/cfdictionary.h
+++ b/include/wx/osx/core/cfdictionary.h
@@ -58,6 +58,14 @@ public:
      */
     wxCFDictionaryRefCommon(const wxCFDictionaryRefCommon&) = default;
 
+    /*! @method     operator=
+        @abstract   Assigns the other ref's pointer to us when the otherRef is the same type.
+        @param otherRef The other ref holder to copy.
+        @discussion The incoming pointer is retained, the original pointer is released, and this object
+                    is made to point to the new pointer.
+    */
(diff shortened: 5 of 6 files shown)


include/wx/event.h
```diff
@@ -3522,7 +3522,7 @@ struct WXDLLIMPEXP_BASE wxEventTableEntry : public wxEventTableEntryBase
     const int& m_eventType;
 
 private:
-    wxDECLARE_NO_ASSIGN_CLASS(wxEventTableEntry);
+    wxDECLARE_NO_ASSIGN_DEF_COPY(wxEventTableEntry);
```
vadz, line 3525: This is the only one I have some doubts about, as I'm not sure whether we really should be copy-constructing objects of this type at all. Would you have the error message with the warning for it, to show where exactly is its copy ctor used? I can't find it easily by just looking at the code.

vadz (commented):
Thanks for fixing this!

Event classes and validators definitely should be copyable, so changing them like you did is definitely right, but I have a question about one other class.

Another question is about 3.2: I don't know if adding defaulted copy ctor is ABI-compatible. It might be, but even if it is, we need to use `wxABI_VERSION` check around it, so this can't be backported "as is" unfortunately.

include/wx/event.h
rmpowell77: The copy ctor is rather hidden.

Here is where it is getting used in the code, in auidemo.cpp:
```
wxBEGIN_EVENT_TABLE(wxSizeReportCtrl, wxControl)
    EVT_PAINT(wxSizeReportCtrl::OnPaint) <----
    EVT_SIZE(wxSizeReportCtrl::OnSize)
    EVT_ERASE_BACKGROUND(wxSizeReportCtrl::OnEraseBackground)
wxEND_EVENT_TABLE()
```

The EVT_PAINT macro entry effectively says "add an wxEventTableEntry to the table".  So that's an implicit copy constructor from the warnings:

```
In file included from /Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/samples/aui/auidemo.cpp:12:
In file included from /Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/wxprec.h:42:
In file included from /Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/wx.h:24:
/Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/event.h:3525:5: warning: definition of implicit copy constructor for 'wxEventTableEntry' is deprecated because it has a user-declared copy assignment operator [-Wdeprecated-copy]
    wxDECLARE_NO_ASSIGN_CLASS(wxEventTableEntry);
    ^
/Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/defs.h:3053:20: note: expanded from macro 'wxDECLARE_NO_ASSIGN_CLASS'
        classname& operator=(const classname&) wxMEMBER_DELETE
                   ^
/Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/samples/aui/auidemo.cpp:253:5: note: in implicit copy constructor for 'wxEventTableEntry' first required here
    EVT_PAINT(wxSizeReportCtrl::OnPaint)
    ^
/Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/event.h:4507:26: note: expanded from macro 'EVT_PAINT'
#define EVT_PAINT(func)  wx__DECLARE_EVT0(wxEVT_PAINT, wxPaintEventHandler(func))
                         ^
/Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/event.h:4475:5: note: expanded from macro 'wx__DECLARE_EVT0'
    wx__DECLARE_EVT1(evt, wxID_ANY, fn)
    ^
/Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/event.h:4473:5: note: expanded from macro 'wx__DECLARE_EVT1'
    wx__DECLARE_EVT2(evt, id, wxID_ANY, fn)
    ^
/Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/event.h:4471:5: note: expanded from macro 'wx__DECLARE_EVT2'
    wxDECLARE_EVENT_TABLE_ENTRY(evt, id1, id2, fn, nullptr),
    ^
/Users/richardpowell/Development/github.com/rmpowell77/wxWidgets.git/include/wx/event.h:87:5: note: expanded from macro 'wxDECLARE_EVENT_TABLE_ENTRY'
    wxEventTableEntry(type, winid, idLast, wxNewEventTableFunctor(type, fn), obj)
    ^
```

I'm not sure we could easily make this type of warning an error, say by removing the default copy-ctor -- that would mean ever call for wxDECLARE_EVENT_TABLE_ENTRY would be an error.

I guess looking at it, while we shouldn't be generally copy-constructing `wxEventTableEntry`, by the way the macros set up the table we effectively are.

vadz: Thanks!

So it's initializing `sm_eventTableEntries` which uses the copy ctor. I guess this could be avoided by using braces instead in `wxDECLARE_EVENT_TABLE_ENTRY` definition but I'm not sure if it doesn't have some other undesirable side effects, so let's just merge this PR as is, finally.

Thanks again!

PR #23629, opened 2023-06-10
Fix a linker error when wxUSE_PROTOCOL_FILE is off

matyalatte:
Tiny fix for a linker error.
It'll allow us to disable `wxUSE_PROTOCOL_FILE` with `wxUSE_PROTOCOL` enabled.


vadz:

PR #23633, opened 2023-06-13
Add support for building WebView with libwebkit2gtk-4.1

swt2c:


swt2c:
@MaartenBent let me know if you have any comments on the cmake changes.

src/gtk/webview_webkit2.cpp
```diff
@@ -252,15 +252,18 @@ wxgtk_webview_webkit_load_failed(WebKitWebView *,
     {
         switch (error->code)
         {
+#if SOUP_MAJOR_VERSION < 3
```
vadz, line 255: Does this mean that we don't get `CANCELLED` error at all any longer or do we still get it in some other way? Looking at https://libsoup.org/libsoup-3.0/migrating-from-libsoup-2.html#status-codes-no-longer-used-for-internal-errors it seems like we should examine `GError` for it but I haven't looked at this at all and just wanted to ask if you already had.

vadz (commented):
Thanks for the changes!

I'm not totally sure about the error handling, but I guess not getting some errors is better than not being able to build at all :-/ so I'll still merge it soon if there are no other comments.

And I guess this should be backported to 3.2 too, right?

swt2c:
> And I guess this should be backported to 3.2 too, right?

Yes, although it's not clear to me whether switching from webkit2gtk-4.0 to webkit2gtk-4.1 might cause an ABI change, so if it does, we might have to change the backport to make 4.1 opt-in.

build/cmake/modules/FindLIBSOUP.cmake
```diff
@@ -31,19 +31,24 @@
 # LibSoup does not provide an easy way to retrieve its version other than its
 # .pc file, so we need to rely on PC_LIBSOUP_VERSION and REQUIRE the .pc file
 # to be found.
+SET(LIBSOUP_VERSION 2.4)
+if(DEFINED LIBSOUP_FIND_VERSION)
+    SET(LIBSOUP_VERSION ${LIBSOUP_FIND_VERSION})
+endif()
+
```
MaartenBent, line 38: I tried to update an existing CMake project from webkitgtk-4.0/libsoup-2.4 to webkitgtk-4.1/libsoup-3.0. It correctly found the packages, but did not update the cached variables. Could you add these lines so the variables will be updated? 
FindWEBKIT2 already has this.

```suggestion

set(LIBSOUP_INCLUDE_DIRS LIBSOUP_INCLUDE_DIRS-NOTFOUND)
set(LIBSOUP_LIBRARIES LIBSOUP_LIBRARIES-NOTFOUND)

```

MaartenBent (commented):
> @MaartenBent let me know if you have any comments on the cmake changes.

CMake changes look good. I left one suggestion for improvement.

build/cmake/modules/FindLIBSOUP.cmake
swt2c: Thanks for the comment.  Incorporated.

vadz:
> > And I guess this should be backported to 3.2 too, right?
> 
> Yes, although it's not clear to me whether switching from webkit2gtk-4.0 to webkit2gtk-4.1 might cause an ABI change, so if it does, we might have to change the backport to make 4.1 opt-in.

I agree that it might be unexpected if the new library build suddenly acquires a new dependency, so if both webkit2gtk-4.0 and webkit2gtk-4.1 are available we should keep using webkit2gtk-4.0 in 3.2. But I don't know what should be done if only webkit2gtk-4.1 is found: can we use it unconditionally or would this still be unexpected? To be on the totally safe side we could issue a warning in configure saying that webkit2gtk-4.1 was detected but you need to use some special configure `--with-webkit2gtk-4.1` option to use it, perhaps?

vadz:
Sorry, forgot to say: for now I'll merge this in master, at least.

MaartenBent:
I noticed an additional CMake issue. If 4.1 cannot be found, it will always show a warning message. Even if 4.0 is found after.
```
-- Could NOT find WEBKIT2 (missing: WEBKIT2_LIBRARY WEBKIT2_INCLUDE_DIR) (Required is at least version "4.1")
```

It should probably use 
```cmake
find_package(WEBKIT2 4.1 QUIET)
```
I wiill create a PR to fix this.

PR #23635, opened 2023-06-13
CMake: Improve finding library name or path of imported targets

### LINKED ISSUE #23632: CMake 3.18+ seems to be required now
I'm getting the following error:
```
CMake Error at build/cmake/config.cmake:50 (get_target_property):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "LOCATION_RELEASE" is not allowed.
  build/cmake/config.cmake:165 (wx_get_dependencies)
  build/cmake/config.cmake:204 (wx_write_config)
  build/cmake/main.cmake:51 (include)
  CMakeLists.txt:103 (include)
```
when building using CMake 3.17.5 on a system where a newer version can't be easily installed.

This is due to #23390 (see the changes of 8102958665 (CMake: Add libraries of imported targets to wx-config, 2023-03-28)) which fixed #23373, which doesn't seem worth bumping up CMake requirement so much, and definitely not in 3.2.

Could this fragment be made conditional on CMake version, as apparently only some later version of CMake can set these properties on interface libraries? Or can we somehow whitelist them in the earlier versions too?

I guess we could also try increasing the required CMake version in master, but, again, I definitely don't think we should drop support for older versions in 3.2 without a very good reason, so if we don't find any other solution we need to at least revert e1d8bc89ba (CMake: Add libraries of imported targets to wx-config, 2023-03-28) on this branch.

MaartenBent:
Checking `LOCATION{_CONFIG}` property on interfaces is only supported since CMake 3.19.
Also check the `LOCATION` and `IMPORTED_LIBNAME` properties for possible library locations.

Fixes https://github.com/wxWidgets/wxWidgets/issues/23632

---

Fix using Cotire for precompiled headers. Cotire doesn't work when `$<COMPILE_LANGUAGE:CXX>` is used.


MaartenBent:
This can/should be backported to 3.2 too.

The fix for Cotire is not needed in 3.2.

PR #23636, opened 2023-06-13
[macOS] BUG FIX - DataviewCtrl's column value can be invalid but checking is not done

hwiesmann:
Although a valid row value has been obtained, the column value can also be invalid. But a check is not done. In those cases the app will crash because of an invalid array index.

Remarks: An invalid column value can be obtained by double-clicking to the far left or right of a table.


csomor:
thanks a lot Hartwig

vadz:
Thanks for the fix! I will merge it in master soon and will also backport to 3.2.

PR #23637, opened 2023-06-13
Removal of unjustified undeclared selector warnings

hwiesmann:
Selectors may exist in case panel is an NSOpenPanel, respectively selectors are created when they do not exist.

### DIFF
--- a/src/osx/cocoa/filedlg.mm
+++ b/src/osx/cocoa/filedlg.mm
@@ -446,10 +446,13 @@ void wxFileDialog::SetupExtraControls(WXWindow nativeWindow)
     {
         [accView removeFromSuperview];
         [panel setAccessoryView:accView];
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wundeclared-selector"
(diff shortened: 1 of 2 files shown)


vadz:
What clang version gives these warnings and why we don't see them in the existing CI builds?

I also think it would be better to use `wxCLANG_WARNING_SUPPRESS()` and `wxCLANG_WARNING_RESTORE()` here instead of using the pragmas unconditionally in case some older (but still supported) clang versions don't have this warning -- and also because the macros are more clear/readable IMO.

hwiesmann:
I did not know that these macros existed.

I am using the Clang version that shipped with Xcode 14.3. But I never use any CMake generated projects of any kind nor do I use CMake to compile anything. I always create my own projects. Therefore, I think that this is more a CMake issue than a Clang issue.

hwiesmann:
I have opened a new pull request with the same topic but by using the suggested macros instead of the pragmas directly.

vadz:
Replaced by #23639.

PR #23639, opened 2023-06-14
Removal of unjustified undeclared selector warnings

hwiesmann:
Modification of previous pull request proposal to same issue: now, usage of macros instead of pragmas.


vadz:
@csomor I'll merge this if you don't mind. I am not sure if there is no better way to handle this (i.e. if we can make the compiler "see" that we're testing for the selector support before using it), but I don't see how can this do any real harm.

csomor:
> @csomor I'll merge this if you don't mind. I am not sure if there is no better way to handle this (i.e. if we can make the compiler "see" that we're testing for the selector support before using it), but I don't see how can this do any real harm.

yes, it's a pity clang does not recognize the pattern, but since it is not, then I'm absolutely fine with adding these. Thanks Hartwig.

Issue #22562, opened 2022-06-23
wxVListBox black lines between items


AndersMunch:
**Describe the bug**

wxVListBox renders black lines between items, overwriting owner-drawn content.

**Expected vs observed behaviour**

In 3.1.6 and older, the list items appeared as drawn by `OnDrawItem`, as expected.  In 3.1.7, black lines appear between list items. 

**Patch or snippet allowing to reproduce the problem**

Sample program to illustrate:
```python
import wx

class MyVListBox(wx.VListBox):
    def __init__(self, parent):
        super().__init__(parent, -1)
        self.SetItemCount(20)
    def OnMeasureItem(self, n):
        return 10
    def OnDrawItem(self, dc, rect, n):
        col = ['#aaaaaa', '#dddddd'][n%2]
        dc.SetBrush(wx.Brush(col))
        dc.DrawRectangle(rect)

class MyFrame(wx.Frame):
    def __init__(self):
        super().__init__(None, -1, wx.version())
        self._vl = vl = MyVListBox(self)
        self.Layout()

if __name__=='__main__':
    app = wx.App()
    fr = MyFrame()
    fr.Show(True)
    app.MainLoop()
```

In 3.1.5 and 3.1.6, it looks like this:
![billede](https://user-images.githubusercontent.com/10682854/175274225-1ded8666-b5b6-41a7-ad6c-d84fe07f4a10.png)

In 3.1.7, it looks like this:
![billede](https://user-images.githubusercontent.com/10682854/175274339-b0ca7052-9fc8-439b-873a-f232bbf1edb5.png)

In both cases, the `wx.Rect` arguments to `OnDrawItem` calls are adjoining: In the sample program it'll be (x=0,height=10) followed by (x=10,height=10), (x=20, height=10) etc., there are no gaps.

**Platform and version information**
 - wxWidgets version you use: 3.1.7
 - wxWidgets port you use: wxMSW via wxPython
 - OS and its version: Windows 10.0.19044.1766

MaartenBent:
These seem indeed regressions from https://github.com/wxWidgets/wxWidgets/pull/2412, though both problems have a different cause.

The first one is caused by the use of `wxDCPenChanger`. IMO this is intended. To fix the border around the drawn rectangles you have to manually set a pen using (or the python equivalent) in your `OnDrawItem`:
```c++
dc.SetPen(*wxTRANSPARENT_PEN);
```

[...]

PR #23641, opened 2023-06-15
Only draw selection rectangle for selected items in wxVListBox

MaartenBent:
See #22562


vadz:
Thanks for fixing this!

Merging into master, and I think it should be backported to 3.2 too, right?

PR #23642, opened 2023-06-15
Improve documentation about parallel builds with MinGW makefile

PBfordev:
In the note about building wxWidgets with MinGW makefile and using parallel build, explain that "-jN" option cannot be used when building "setup_h" target.

The issue was reported on wxForum and I reproduced it.

Should be backported.


vadz:
Thanks for improving this! I will merge (and backport) it soon.

Issue #23401, opened 2023-03-30
[MSW] Without wxUSE_UXTHEME, some controls get the wrong border


BrianNixon:
### Description
Without `wxUSE_UXTHEME`, `wxBORDER_DEFAULT` becomes `wxBORDER_THEME` in `wxControlBase::GetDefaultBorder()`, which is unaffected by `wxWindowMSW::TranslateBorder()` and so in `wxWindowMSW::MSWGetStyle()` effectively means ‘none’. To preserve the native look, controls should have `wxBORDER_SUNKEN`. (Or maybe `wxDEFAULT_CONTROL_BORDER`?) This shows up glaringly on Windows 7 classic.

Affects at least `wxListBox`, `wxListCtrl` and `wxTextCtrl`. (There may be others; I haven’t tested exhaustively.)

Relatedly: `wxSpinCtrl` and `wxTreeCtrl` explicitly set `wxBORDER_SUNKEN` in their `Create()` functions.

### Platform and version information

- wxWidgets version  you use: 3.2.2.1
- wxWidgets port  you use: wxMSW
- OS  and its version: Windows 7 [Version 6.1.7601]

vadz:
Do you have any good reason to disable `wxUSE_UXTHEME`? So many things depend on it that it would be rather painful to fully support not using it, so I wouldn't want to spend any time on this unless it's really needed. In fact, I'd rather drop this option entirely, to be honest.

BrianNixon:
> Do you have any good reason to disable `wxUSE_UXTHEME`?

Not really.

Closing WONTFIX

PR #23644, opened 2023-06-16
Border-related fixes

vadz:


vadz:
I'm going to merge this soon if there are no objections but it would be still nice if this could be tested.

cc @marekr @Tommykrul

BrianNixon:
Incidentally this fixes #23401

PR #23645, opened 2023-06-17
Added a few missing and fixed a few fuzzy French translations

qtnc:
Added a few missing and fixed a few fuzzy French translations

### DIFF
--- a/locale/fr.po
+++ b/locale/fr.po
@@ -1104,35 +1104,33 @@ msgstr "&Enregistrer"
 
 #: ../src/common/docview.cpp:523 ../src/common/docview.cpp:561
 msgid "&Discard changes"
-msgstr ""
+msgstr "&Ne pas enregistrer"
 
 #: ../src/common/docview.cpp:524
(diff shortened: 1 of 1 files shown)


locale/fr.po
```diff
@@ -9966,13 +9963,11 @@ msgstr "Échec de la création de %s « %s »."
 #~ msgid "Print preview"
 #~ msgstr "Aperçu avant impression"
 
-#, fuzzy
 #~ msgid "&Preview..."
-#~ msgstr " Aperçu"
+#~ msgstr " Aperçu..."
```
vadz, line 9967: Just for the record, I don't think it's useful to modify the commented out entries, they're not used anyhow and, in fact, should probably be all removed -- AFAIK gettext tools keep them there just to allow reusing the old translations in the new ones, but I don't think it's very useful in practice (unless tools like Poedit use them too?).

vadz (commented):
Thanks for the updates! I will merge this soon.

PR #23648, opened 2023-06-18
Fix TAB navigation in MSW radio box

vadz:
@qtnc This should fix the problem you've reported.


qtnc:
Yes, the problem is indeed solved with this PR. Tabbing through the radio button group now works fine again.


Thank you very much !

PR #23649, opened 2023-06-18
Fix scrollbar creation for Mac windows not using standard peer

### LINKED ISSUE #23536: wxGLCanvas init crash with SCROLL flag under MACOS 10.15.7
version, 3.2.2.1
wxGLCanvas cannot be initialized with wxVSCROLL flag
	canvas = new wxGLCanvas(
	this
	, wxID_ANY
	, NULL
	, wxPoint(0, 0)
	, wxDefaultSize
	, wxHSCROLL | wxVSCROLL);//crash for scrollbar, BUT it work under windows

how to support scrollbar for glcanvas

vadz:
This avoids crash if wxVSCROLL or wxHSCROLL is used with wxGLCanvas which calls wxWindow::Create() after calling DontCreatePeer() because the Create() tries to use the not (yet) existent peer in MacCreateScrollBars().

Avoid this problem by moving the call to MacCreateScrollBars() to MacPostControlCreate(), which doesn't change anything for the windows that do not call DontCreatePeer(), but ensures that the peer already exists by the time MacPostControlCreate() is called for those that do.

Closes #23536.

### DIFF
--- a/include/wx/osx/window.h
+++ b/include/wx/osx/window.h
@@ -342,7 +342,7 @@ protected:
     virtual bool        MacIsChildOfClientArea( const wxWindow* child ) const ;
 
     bool                MacHasScrollBarCorner() const;
-    void                MacCreateScrollBars( long style ) ;
+    void                MacCreateScrollBars( ) ;
     void                MacRepositionScrollBars() ;
 
     // implement the base class pure virtuals
--- a/src/osx/window_osx.cpp
+++ b/src/osx/window_osx.cpp
@@ -390,19 +390,6 @@ bool wxWindowMac::Create(wxWindowMac *parent,
         MacPostControlCreate(pos, size) ;
     }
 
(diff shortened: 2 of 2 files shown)


csomor (approved):
@vadz thanks a lot !
but doesn't this break binary compat ? or don't you want to backport it to 3.2 anyway

vadz:
We could preserve `MacCreateScrollBars()` parameter in 3.2, although normally this is a private function and not really part of the ABI, but I didn't want to keep a useless parameter in master.

csomor:
> We could preserve `MacCreateScrollBars()` parameter in 3.2, although normally this is a private function and not really part of the ABI, but I didn't want to keep a useless parameter in master.

I was not sure about it, I thought it might be exported because it's protected, but of course chances that's it's linked against from the outside are minimal.

PR #23654, opened 2023-06-19
Support system catch2

apteryks:
Teach the Autoconf/Bakefile build system to use catch2 from the system.

### DIFF
--- a/build/bakefiles/config.bkl
+++ b/build/bakefiles/config.bkl
@@ -408,6 +408,7 @@ compiled .lib files and setup.h under the lib/ toplevel directory.
         <option name="WX_CFLAGS"/>
         <option name="WX_CXXFLAGS"/>
         <option name="WX_LDFLAGS"/>
+        <option name="WX_TEST_CXXFLAGS"/>
         <option name="HOST_SUFFIX"/>
         <option name="DYLIB_RPATH_INSTALL"/>
         <option name="DYLIB_RPATH_POSTLINK"/>
--- a/configure.ac
+++ b/configure.ac
@@ -138,6 +138,7 @@ SAMPLES_CXXFLAGS=
 SAMPLES_RPATH_FLAG=
 DYLIB_RPATH_INSTALL=
 DYLIB_RPATH_POSTLINK=
+WX_TEST_CXXFLAGS=
 
 dnl to support a new system, you need to add its canonical name (as determined
 dnl by config.sub or specified by the configure command line) to this "case"
@@ -512,6 +513,7 @@ dnl ---------------------------------------------------------------------------
 WX_ARG_SYS_WITH(libpng,    [  --with-libpng           use libpng (PNG image format)], wxUSE_LIBPNG)
 WX_ARG_SYS_WITH(libjpeg,   [  --with-libjpeg          use libjpeg (JPEG file format)], wxUSE_LIBJPEG)
 WX_ARG_SYS_WITH(libtiff,   [  --with-libtiff          use libtiff (TIFF file format)], wxUSE_LIBTIFF)
+WX_ARG_SYS_WITH(catch2,    [  --with-catch2           use catch2 (testing framework)], wxUSE_CATCH2)
 
 dnl We don't provide built-in version of this library, so if the use of system
 dnl libraries is explicitly disabled, we can't use it at all.
@@ -2055,6 +2057,7 @@ else
 fi
 
 dnl Find pkg-config outside of any conditional.  Done before any PKG_* call.
+dnl This sets the PKG_CONFIG variable.
 PKG_PROG_PKG_CONFIG
 
 dnl When cross-compiling for another system from Linux, don't use .pc files on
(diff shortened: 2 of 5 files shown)


Apteryks:
> Before looking at it further, this seems to include #23640, right? Does it really need it or could you remove all the other commits except for the last one here?

I wasn't sure, but it seems like it does, at least using Autoconf 2.69.  When attempting to regenerate the configure script with `make -f build/autogen.mk`, I get:
```
[...]
no changes in ../../autoconf_inc.m4
no changes in ../../Makefile.in
0 files modified
make -C build/autoconf_prepend-include

make[1] : on entre dans le répertoire « /home/maxim/src/wxWidgets/build/autoconf_prepend-include »
autom4te -B . --language=Autoconf --freeze --output=autoconf/autoconf.m4f
autoconf/general.m4:301: error: m4_copy: won't overwrite defined macro: AC_PREREQ
autoconf/general.m4:301: the top level
autom4te: /gnu/store/2awzcd1jdd8h65ahdn62ydllrs2s2pk1-m4-1.4.19/bin/m4 failed with exit status: 1
make[1]: *** [Makefile:25 : autoconf/autoconf.m4f] Erreur 1
make[1] : on quitte le répertoire « /home/maxim/src/wxWidgets/build/autoconf_prepend-include »
make: [build/autogen.mk:29 : autoconf_m4f] Erreur 2 (ignorée)
autoconf -B build/autoconf_prepend-include
```

If based on master directly.

vadz:
Sorry but I simply don't see how is this possible as the last commit doesn't do anything that materially changes anything. You must have some local changes/cached files breaking this.

You should create a PR with just this commit (after fixing the problems in it which result in the CI failures).

Apteryks:
> Sorry but I simply don't see how is this possible as the last commit doesn't do anything that materially changes anything. You must have some local changes/cached files breaking this.

Maybe it's my version of Autoconf (2.69); there shouldn't be anything cached in my environment:
```
Sorry but I simply don't see how is this possible as the last commit doesn't do anything that materially changes anything. You must have some local changes/cached files breaking this.
```
I'll try cherry picking the single commit I expect to fix this (stop versioning Autoconf-provided M4 files).

build/bakefiles/config.bkl
```diff
@@ -408,6 +408,7 @@ compiled .lib files and setup.h under the lib/ toplevel directory.
         <option name="WX_CFLAGS"/>
         <option name="WX_CXXFLAGS"/>
         <option name="WX_LDFLAGS"/>
+        <option name="WX_TEST_CXXFLAGS"/>
```
vadz, line 411: I think it's a bit misleading to call it like this, it should probably be called `CATCH2_CXXFLAGS` or something like this instead.

configure.ac
```diff
@@ -8022,21 +8027,41 @@ if test "$wxUSE_TESTS_SUBDIR" != "no"; then
...
-    dnl Check for catch (C++ Automated Test Cases in Headers) availability.
-    AC_MSG_CHECKING([whether catch2/catch.hpp exists])
-    if ! test -f "$srcdir/3rdparty/catch/single_include/catch2/catch.hpp" ; then
-        AC_MSG_RESULT([no])
-        AC_MSG_ERROR([
-    CATCH (C++ Automated Test Cases in Headers) is required, the required file
-    $srcdir/3rdparty/catch/single_include/catch2/catch.hpp couldn't be found.
+   dnl Check for catch (C++ Automated Test Cases in Headers) availability.
+    if test "$wxUSE_CATCH2" != "builtin"; then
+        if test -n "$PKG_CONFIG"; then
```
vadz, line 8032: We don't test for `PKG_CONFIG` anywhere else, so either we shouldn't do it here neither or we should do it in all the other places too... But is it even possible for it to be empty? I.e. won't configure stop if `PKG_PROG_PKG_CONFIG` doesn't find it?

configure.ac
```diff
@@ -8022,21 +8027,41 @@ if test "$wxUSE_TESTS_SUBDIR" != "no"; then
...
+            dnl TODO: Support catch2 version 3, which does not
+            dnl ship a catch2/catch.hpp header anymore.
+            PKG_CHECK_MODULES([CATCH2], [catch2 < 3],, [
```
vadz, line 8035: We should use `CATCH2_CFLAGS` somewhere, otherwise compilation will fail if it's not installed under `/usr/include`.

vadz (changes requested):
Thanks for the update but unfortunately there are still some things that could be improved.

I also think that it would make sense to default to built-in version: unlike the other libraries, this is not used for the binaries being distributed, so I don't see any drawbacks to using the built-in version by default and it would avoid warnings about not finding the system one, which would be annoying to see.

vadz:

vadz:
Just FYI, I'll fix and apply it, finally.

Apteryks:
> Just FYI, I'll fix and apply it, finally.

Thank you.  Sorry for dropping the ball.

PR #23655, opened 2023-06-19
Rename configure.in to configure.ac

apteryks:


acinclude.m4
```diff
@@ -384,7 +384,7 @@ dnl like WX_ARG_WITH but uses AC_ARG_ENABLE instead of AC_ARG_WITH
 dnl usage: WX_ARG_ENABLE(option, helpmessage, var, [enablestring], [default])
 dnl
 dnl enablestring can be omitted or a literal string "disable" and allows to
-dnl show "checking for --disable-foo" message when running configure instead of
+dnl show "checking for --disable-foo" message when running configure.acstead of
```
vadz, line 387: Ugh, sorry, but you really, really shouldn't be running automatic replacements in the entire search tree and if you do it, you should at least be using the right pattern, with `.` metacharacter being escaped.

vadz (changes requested):
Sorry, this is another not ideal PR, it shouldn't include #23640 once again. Please either

1. Wait until the other PR is merged (or rejected).
2. Add this commit to the other PR itself.
3. Cherry pick just the new commit here and create a new PR with it (this will create conflicts when merging however).

Also, please carefully review your changes before submitting them, I'm sorry but I don't have enough time to do manual reviews of automatic replacements.

Thanks in advance.

acinclude.m4
Apteryks: Oops, sorry! I did quickly review the 54 something instances but missed that one.  I've fixed that typo, tightened the regexp and pushed, after validating that it seems to configure and build fine on my end.

PR #23657, opened 2023-06-20
Restore setting of delta-RLE bitmap background to first entry in colour table

### LINKED ISSUE #23638: Inconsistency in setting delta-RLE bitmap background colour
#23599 modified the way wxWidgets sets the background of delta-RLE bitmaps, ostensibly to match the way Windows does it. While that is valid in some cases, it turns out Windows itself is inconsistent.

Given a delta-RLE bitmap, `LoadImage()` with `LR_CREATEDIBSECTION` creates an uncompressed DIB with the same bit depth as the original and with unset pixels given the same colour as the first entry in the colour table. All other bitmap-creation methods (`LoadImage()`/​`LoadBitmap()`/​`CreateDIBitmap()`) create a DDB, which increases the colour depth to 24bpp to match the display device context and makes unset pixels black. 

This difference can be highlighted by adding another image comparison to `CompareBMPImageLoad()` in `tests/​image/​image.cpp`:
```cpp
    wxBitmap bitmap3(filename, wxBITMAP_TYPE_BMP);
    REQUIRE( bitmap3.IsOk() );
    wxImage image3(bitmap3.ConvertToImage());
    REQUIRE( image3.IsOk() );
    INFO("wxBitmap::LoadFile vs wxImage::LoadFile");
    CompareImage(*wxImage::FindHandler(wxBITMAP_TYPE_BMP),
        image3, properties, &image2);
```

In today’s `master`, this test will fail. Before b473163da20c1f36caaf2e4db512bb8fb4d73a93, it would have passed. It is impossible for delta-RLE bitmaps to pass both this test and the existing comparison of `wxDIB::ConvertToBitmap` to `wxImage::LoadFile` (unless the first entry in their colour table happens to be black, which the files in the image tests were presumably deliberately constructed to avoid).

The question is, then: which behaviour should wxWidgets mimic? Both are by some definition ‘correct’. Clearly somebody had made the effort to implement and test the previous version, and that is now broken, so there is a case for undoing the change made in b473163da20c1f36caaf2e4db512bb8fb4d73a93.

BrianNixon:
Revert the change introduced in b473163da20c1f36caaf2e4db512bb8fb4d73a93 and restore the setting of the background of delta-RLE bitmaps to the first entry in the colour table. This retains earlier wxWidgets behaviour and matches Windows `LoadImage()` with `LR_CREATEDIBSECTION`.

Closes #23638.


vadz:
Thanks for (re)fixing this! Will merge soon.

PR #23658, opened 2023-06-20
Restore compatibility with the existing code using custom controls in the "About" dialog

vadz:


Randalphwa:
I've reviewed the code changes as well as the updated description in the header file, and the changes look fine. The change to the sample matches almost exactly what I was doing so I'm confident enough that this will work that I'll wait until it gets merged into the 3.3 sources before rebuilding my app rather than grabbing the PR and making a test build from that.

PR #23661, opened 2023-06-21
Fix problems flagged by ASAN in clipboard/data object code and improve wxMSW HTML data object

vadz:


vadz:
Adding the unit test for the original problem uncovered a bug in wxMSW `wxHTMLDataObject` implementation, so I've fixed it which incidentally involved rewriting most of its code.

mcorino:
I would very much appreciate this being merged.

PR #23662, opened 2023-06-22
Enable `-Wextra` for the Unix CI builds

vadz:


vadz:
This (finally) passes and I'd like to merge it if there are no objections.

PR #23663, opened 2023-06-22
Fix compilation errors when wxUSE_INTL or wxUSE_RADIOBTN is off.

matyalatte:
I noticed [the latest changes of `datetime.cpp`](https://github.com/wxWidgets/wxWidgets/commit/5b424ea181400f1898fd297f97e544d055191a82) won't work with `wxUSE_INTL=0`.
This PR will fix the issue and some compilation errors I found.

- Fixed compilation errors when `wxUSE_INTL` is off.
- Fixed compilation erros for wxFileDialog when `wxUSE_RADIOBTN` is off.

### DIFF
--- a/include/wx/filedlgcustomize.h
+++ b/include/wx/filedlgcustomize.h
@@ -15,7 +15,9 @@
 class wxFileDialogCustomControlImpl;
 class wxFileDialogButtonImpl;
 class wxFileDialogCheckBoxImpl;
+#if wxUSE_RADIOBTN
 class wxFileDialogRadioButtonImpl;
+#endif
 class wxFileDialogChoiceImpl;
 class wxFileDialogTextCtrlImpl;
 class wxFileDialogStaticTextImpl;
@@ -92,6 +94,7 @@ private:
     wxDECLARE_NO_COPY_CLASS(wxFileDialogCheckBox);
 };
 
+#if wxUSE_RADIOBTN
 // A class representing a custom radio button.
 class WXDLLIMPEXP_CORE wxFileDialogRadioButton : public wxFileDialogCustomControl
 {
@@ -110,6 +113,7 @@ private:
 
     wxDECLARE_NO_COPY_CLASS(wxFileDialogRadioButton);
 };
+#endif // wxUSE_RADIOBTN
 
 // A class representing a custom combobox button.
 class WXDLLIMPEXP_CORE wxFileDialogChoice : public wxFileDialogCustomControl
@@ -170,7 +174,9 @@ class WXDLLIMPEXP_CORE wxFileDialogCustomize
 public:
     wxFileDialogButton* AddButton(const wxString& label);
     wxFileDialogCheckBox* AddCheckBox(const wxString& label);
+#if wxUSE_RADIOBTN
     wxFileDialogRadioButton* AddRadioButton(const wxString& label);
+#endif
     wxFileDialogChoice* AddChoice(size_t n, const wxString* strings);
     wxFileDialogTextCtrl* AddTextCtrl(const wxString& label = wxString());
     wxFileDialogStaticText* AddStaticText(const wxString& label);
--- a/include/wx/private/filedlgcustomize.h
+++ b/include/wx/private/filedlgcustomize.h
@@ -38,12 +38,14 @@ public:
     virtual void SetValue(bool value) = 0;
 };
 
+#if wxUSE_RADIOBTN
 class wxFileDialogRadioButtonImpl : public wxFileDialogCustomControlImpl
 {
 public:
     virtual bool GetValue() = 0;
     virtual void SetValue(bool value) = 0;
 };
+#endif // wxUSE_RADIOBTN
 
 class wxFileDialogChoiceImpl : public wxFileDialogCustomControlImpl
 {
@@ -74,7 +76,9 @@ class wxFileDialogCustomizeImpl
 public:
     virtual wxFileDialogButtonImpl* AddButton(const wxString& label) = 0;
     virtual wxFileDialogCheckBoxImpl* AddCheckBox(const wxString& label) = 0;
+#if wxUSE_RADIOBTN
     virtual wxFileDialogRadioButtonImpl* AddRadioButton(const wxString& label) = 0;
+#endif
     virtual wxFileDialogChoiceImpl* AddChoice(size_t n, const wxString* strings) = 0;
     virtual wxFileDialogTextCtrlImpl* AddTextCtrl(const wxString& label) = 0;
     virtual wxFileDialogStaticTextImpl* AddStaticText(const wxString& label) = 0;
--- a/src/common/datetime.cpp
+++ b/src/common/datetime.cpp
@@ -781,10 +781,12 @@ wxString wxDateTime::GetMonthName(Month month,
                                   const NameForm& form)
 {
     wxCHECK_MSG(month != Inv_Month, wxEmptyString, wxT("invalid month"));
+#if wxUSE_INTL
     wxString name = wxUILocale::GetCurrent().GetMonthName(month, form);
-    if (name.empty())
-        name = GetEnglishMonthName(month, form);
-    return name;
+    if (!name.empty())
+        return name;
+#endif // wxUSE_INTL
+    return GetEnglishMonthName(month, form);
 }
 
 /* static */
@@ -812,10 +814,12 @@ wxString wxDateTime::GetWeekDayName(WeekDay wday,
                                     const NameForm& form)
 {
     wxCHECK_MSG(wday != Inv_WeekDay, wxEmptyString, wxT("invalid weekday"));
+#if wxUSE_INTL
     wxString name = wxUILocale::GetCurrent().GetWeekDayName(wday, form);
-    if (name.empty())
-        name = GetEnglishWeekDayName(wday, form);
-    return name;
+    if (!name.empty())
+        return name;
+#endif // wxUSE_INTL
+    return GetEnglishWeekDayName(wday, form);
 }
 
 /* static */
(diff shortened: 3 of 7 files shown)


include/wx/translation.h
```diff
@@ -331,13 +331,13 @@ inline const wxString& wxGetTranslation(const char *str1,
 #if !defined(WXINTL_NO_GETTEXT_MACRO)
     #if !defined(_)
 #ifndef wxNO_IMPLICIT_WXSTRING_ENCODING
-        #define _(s)                 (s)
+        #define _(s)                 wxString(s)
```
vadz, line 334: This looks correct but I find it a bit strange that we had never had problems with this before, i.e. AFAICS `_(s)` was always defined just as `s`, ever since it was added back in 88ac883a0d0.

Could you please explain what errors did this result in, i.e. why did you have to change it?

include/wx/translation.h
```diff
@@ -331,13 +331,13 @@ inline const wxString& wxGetTranslation(const char *str1,
...
 #else
         #define _(s)                 wxASCII_STR(s)
 #endif
     #endif
     #define wxPLURAL(sing, plur, n)  ((n) == 1 ? (sing) : (plur))
-    #define wxGETTEXT_IN_CONTEXT(c, s)                     (s)
+    #define wxGETTEXT_IN_CONTEXT(c, s)                     _(s)
```
vadz, line 340: I'm not sure if this one is right though as it makes this macro inconsistent with `wxGETTEXT_IN_CONTEXT_PLURAL`.

Again, perhaps it would help to see what errors exactly the current version results in?

vadz (commented):
Thanks for the fixes!

The `wxUSE_RADIOBUTTON` commit is fine but I'm not 100% sure about the `wxUSE_INTL` one, could you please show (some of) the errors you were getting without these changes? Thanks!

include/wx/translation.h
matyalatte: [`menu_osx.cpp`](https://github.com/wxWidgets/wxWidgets/blob/master/src/osx/menu_osx.cpp) is using it as wxString, like `wxGETTEXT_IN_CONTEXT(c, s) + "foo"`.
So it should be a string. or need to fix `menu_osx.cpp`.

vadz: I think it would be better (and very certainly safer) to just fix this there.

In fact, I think the code is wrong anyhow because it doesn't translate `Ctrl`, i.e. instead of
```
wxGETTEXT_IN_CONTEXT("macOS menu item", "Preferences...") + "\tCtrl+,"
```
it should really be just
```
_("Preferences...\tCtrtl+,")
```
which would remove the need for using context completely. @lanurmi am I missing something here?

But to avoid any extensive changes in this PR you could just change the above to
```
wxGETTEXT_IN_CONTEXT("macOS menu item", "Preferences...") + wxString("\tCtrl+,")
```
for now.

include/wx/translation.h
lanurmi: While `_(s)` has been unchanged since the beginning, the function that it normally points to, `wxGetTranslation()`, has changed between 2.8 and 3.0, from returning `const wxChar*` to returning `const wxString &`.

It would be consistent that `_()`, `wxPLURAL()`, and `wxGETTEXT_IN_CONTEXT()` would all return a `wxString` also when `wxUSE_INTL` is off.

include/wx/translation.h
lanurmi: > In fact, I think the code is wrong anyhow because it doesn't translate `Ctrl`, i.e. instead of
> 
> ```
> wxGETTEXT_IN_CONTEXT("macOS menu item", "Preferences...") + "\tCtrl+,"
> ```
> 
> it should really be just
> 
> ```
> _("Preferences...\tCtrtl+,")
> ```
> 
> which would remove the need for using context completely. @lanurmi am I missing something here?

First of all, I suspect this particular string is never actually shown to users because of #16888. But let's ignore that for a while...

I don't think you would (or at least I would) want someone to be able change the keyboard shortcut something else than `Ctrl+,` in the standard macOS application menu, because that would mean any wxOSX application running with such a translation in use, would have a completely unusual shortcut for entering preferences.

The context also serves the purpose of documenting that this string's translation should match Apple's terminology.

Also, would the shortcut even work if `Ctrl` is translated? I tested modifying in the `internat` sample's first menu item's shortcut from `Ctrl-T` to `Strg-T` in the translation file, and running the sample in German (Germany). The result is that the shortcut becomes just `T`, and indeed just pressing `T` without any modifiers triggers the menu item.

matyalatte: I understand.
So, if you will adopt lanurmi's opinion, I'll replace
`wxPLURAL(sing, plur, n)  ((n) == 1 ? (sing) : (plur))`
with
`wxPLURAL(sing, plur, n)  ((n) == 1 ? _(sing) : _(plur))`.

If not, I'll remove the changes from `translation.h` and replace `wxGETTEXT_IN_CONTEXT(c, s) + "foo"`
with
`wxGETTEXT_IN_CONTEXT(c, s) + wxASCII_STR("foo")`.
(wxASCII_STR might be better to support the no implicit conversions option?)

I'll follow your decision.

matyalatte: > IMHO it would be highly confusing to say `_(X)` when you really mean `wxString(X)`, even though in this very context they happen to be equal.

`wxPLURAL` and `_(X)` are the macros for translation.
So, it's not coincidence. (at least, for me)
`((n) == 1 ? _(sing) : _(plur))` will work even when `wxUSE_INTL` is on.

Well, I noticed wxASCII_STR might be enough for constructing the key strings.
So, this might be better?
```
    #if !defined(_)
        #define _(s)                 wxASCII_STR(s)
    #endif
    #define wxPLURAL(sing, plur, n)  wxASCII_STR((n) == 1 ? (sing) : (plur))
    #define wxGETTEXT_IN_CONTEXT(c, s)                     wxASCII_STR(s)
```

vadz: > I don't think you would (or at least I would) want someone to be able change the keyboard shortcut something else than `Ctrl+,` in the standard macOS application menu, because that would mean any wxOSX application running with such a translation in use, would have a completely unusual shortcut for entering preferences.

The idea is that `Ctrl` should be translated, not changing the shortcut. So we could also use something like

```cpp
wxString::Format("%s\t%s+,",
    wxGETTEXT_IN_CONTEXT("macOS menu item", "Preferences..."),
    _("Ctrl"));
```

I guess.

> The context also serves the purpose of documenting that this string's translation should match Apple's terminology.

OK, this is a good point.

> Also, would the shortcut even work if `Ctrl` is translated?

Well, I don't know if it does right now, but it very clearly should because surely German users should see `Strg+,` in the actual menu, shouldn't they?

> I tested modifying in the `internat` sample's first menu item's shortcut from `Ctrl-T` to `Strg-T` in the translation file, and running the sample in German (Germany). The result is that the shortcut becomes just `T`, and indeed just pressing `T` without any modifiers triggers the menu item.

This is a bug :-( I'm pretty sure this does work correctly in the other ports.

include/wx/translation.h
vadz: The main (only?) reason for turning off `wxUSE_INTL` is to avoid extra code generation if you're not interested in i18n and `wxGetTranslation()` specifically does _not_ return `wxString` in this case in order to avoid constructing it unnecessarily. Also, there would be a difference (albeit a lesser one of course) between the 2 builds anyhow because the real `wxGetTranslation()` returns `const wxString&` and we couldn't return this in `wxUSE_INTL==0` case.

So I'd rather leave the header unchanged and fix the code to avoid assuming that `_()` or whatever returns a `wxString`. Especially because it seems to me -- and the current example proves it -- that concatenating translated strings with something else is rarely, if ever, the right thing to do anyhow.

include/wx/translation.h
vadz: @matyalatte Sorry for the confusion and all the unrelated deviations, but to address your question: I think we should keep the header unchanged and use `Format("%s...")` as shown above instead of string concatenation.

lanurmi: > > Also, would the shortcut even work if `Ctrl` is translated?
> 
> Well, I don't know if it does right now, but it very clearly should because surely German users should see `Strg+,` in the actual menu, shouldn't they?

In the case of wxOSX, `Ctrl` really means `Cmd` and they should be seeing `⌘,`. But aside from that, yes, I agree with you; they should be seeing `Strg+,`. But could it be normalized and localized at the level of `wxMenu`, even if it was given untranslated?


> > I tested modifying in the `internat` sample's first menu item's shortcut from `Ctrl-T` to `Strg-T` in the translation file, and running the sample in German (Germany). The result is that the shortcut becomes just `T`, and indeed just pressing `T` without any modifiers triggers the menu item.
> 
> This is a bug :-( I'm pretty sure this does work correctly in the other ports.

Not sure if this is because the `internat` sample does not have access to `wxstd` translations on macOS, but even if that is the reason, it seems a bit fragile that an unknown modifier name is interpreted as nothing, and no assert fails.

matyalatte:
Restored the original `translation.h`.
And replaced `wxGETTEXT_IN_CONTEXT(c, s) + "x"` with `wxString::Format("%sx", wxGETTEXT_IN_CONTEXT(c, s))`.
Idk why a test on Windows failed but there is no problem to merge this PR, I think.

@vadz 
I can add another commit to apply the `_(X)` to shortcuts in `menu_osx.cpp`.
(but maybe I shouldn't do it now?)
Or you can merge this PR.

vadz:
Thanks for the update! I'll merge this one soon, the rest will probably need more work as there are apparently other problems, so it would be better to do it a separate PR.

PR #23664, opened 2023-06-22
Assume sigaction::sa_handler always takes "int"

vadz:
This is the standard and there doesn't seem to be any contemporary
systems where this is not the case. The configure check for it was
originally added in 101c20d554 (attempting to fix sa_handler signature
for IRIX, 2000-04-01), but IRIX is not supported/doesn't exist any
longer, so don't waste time on checking for this.

The real motivation for this change is that the test failed in CMake
builds, showing just

-- Performing Test wxTYPE_SA_HANDLER_IS_INT - Failed

in the log and then resulting in -fpermissive warnings when assigning
wxFatalSignalHandler to sa_handler and it's simpler to just remove the
check than to debug it.


vadz:

PR #23665, opened 2023-06-23
Ensure CMake tests don't fail due to -Werror being used

vadz:
Disable warnings while compiling the tests to avoid unnecessary noise in
the CMake output/error files (because we don't really care about
warnings in the test code at all) and also, and more importantly, to
ensure that the tests don't fail even when -Werror is specified in the
CMAKE_CXX_FLAGS globally, as it may happen when wxWidgets is used via
add_subdirectory() from a superproject.

### DIFF
--- a/build/cmake/setup.cmake
+++ b/build/cmake/setup.cmake
@@ -82,6 +82,12 @@ if(wxBUILD_SHARED)
     endif()
 endif() # wxBUILD_SHARED
 
+if(MSVC)
+    set(DISABLE_ALL_WARNINGS "/w")
+else()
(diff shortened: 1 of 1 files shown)


vadz:
@MaartenBent Please let me know what do you think of this hack.

MaartenBent:
CMake provides macros to do this: https://cmake.org/cmake/help/latest/module/CMakePushCheckState.html

vadz:
Thanks for the hint! Updated to use these macros now.

vadz:

PR #23670, opened 2023-06-28
Corrected typo in samples/notebook/notebook.cpp

loremol:
Hi, there is a small error in `notebook.cpp` (`"T&oolbook"` instead of `"&Toolbook"`).


vadz:
Thanks but, as already mentioned above, the current label intentionally doesn't use `T` as mnemonic for this item, so this PR is incorrect and shouldn't be applied.

PR #23675, opened 2023-07-01
Fix typos in changelog

PBfordev:
I do not know if these typos exist in the master as well, I will try to check once the changelog is compiled before 3.3 release.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -237,7 +237,7 @@ Changes in behaviour which may result in build errors
 
 All:
 
-- Add move ctor and assignment to wxString (Pavel Tyunin, #23224).
+- Add move ctor and assignment operator to wxString (Pavel Tyunin, #23224).
 - Enable large file support in Unix CMake builds (Maarten Bent, #22750).
 - Update Georgian translations (NorwayFun, #22673).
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for the fixes! I'll merge this, but I have a problem with one of them (adding "Windows") because it makes the line too long: we're trying to keep the lines under 80 characters here to avoid wrapping in text emails, where they're copied. I've change this to use "OS" instead of "Windows", hopefully this should still be clear enough.

Newbytee:
> Thanks for the fixes! I'll merge this, but I have a problem with one of them (adding "Windows") because it makes the line too long: we're trying to keep the lines under 80 characters here to avoid wrapping in text emails, where they're copied. I've change this to use "OS" instead of "Windows", hopefully this should still be clear enough.

Is "Win" too short?

vadz:
> Is "Win" too long?

We don't use "Win" as abbreviation for "Windows" because this is something that is generally not well seen in the Free Software circles. Our usual abbreviation is "MSW" and we could use it here, but "OS" seems to be a bit more clear (knowing that this is already in the "wxMSW" section).

vadz:
Applied in the above-mentioned commit.

PR #23677, opened 2023-07-02
Create hidden radio button in wxGTK when using wxRB_SINGLE

### LINKED ISSUE #23652: wxRB_SINGLE does not work in wxGTK
### Description
When using the GTK backend for wxWidgets, the `wxRB_SINGLE` style does not work properly. It is impossible to set the radio button to be unchecked (`false`). Sample code:

```cpp
#include <wx/wx.h>

class RadioButtonReproducerApp : public wxApp {
public:
    bool OnInit() override;
};

class RadioButtonReproducerFrame : public wxFrame {
public:
    RadioButtonReproducerFrame();
};

wxIMPLEMENT_APP(RadioButtonReproducerApp);

bool RadioButtonReproducerApp::OnInit() {
    RadioButtonReproducerFrame* frame = new RadioButtonReproducerFrame();
    frame->Show();
    return true;
}

RadioButtonReproducerFrame::RadioButtonReproducerFrame()
    : wxFrame(nullptr, wxID_ANY, "Hello World") {

    wxRadioButton* radio_button = new wxRadioButton(
        this,
        wxID_ANY,
        "Meep 1",
        wxDefaultPosition,
        wxDefaultSize,
        wxRB_SINGLE
    );

    // This should disable the radio button, but it doesn't
    radio_button->SetValue(false);
}
```

I've not specifically tested this on Windows, but the company I work for's internal codebase uses this pattern and it works fine on Windows there.

- OS  and its version: Fedora 38

+ GTK version:  3.24.38

Newbytee:
This lets us deactivate the shown button by activating the hidden one. GTK does not allow deactivating radio buttons by calling them with `gtk_toggle_button_set_active ( ..., FALSE )`, so we need to work around it like this.

Tested with GTK 3.24.38.

Closes https://github.com/wxWidgets/wxWidgets/issues/23652
Related https://github.com/wxWidgets/wxWidgets/issues/17022

### DIFF
--- a/include/wx/gtk/radiobut.h
+++ b/include/wx/gtk/radiobut.h
@@ -55,6 +55,8 @@ protected:
 
 private:
     typedef wxControl base_type;
+    // Only used by wxRB_SINGLE
+    GtkWidget *m_hidden_radio_button = nullptr;
 
     wxDECLARE_DYNAMIC_CLASS(wxRadioButton);
 };
--- a/interface/wx/radiobut.h
+++ b/interface/wx/radiobut.h
@@ -33,9 +33,9 @@
            When this style is used, no other radio buttons will be turned off
            automatically when this button is turned on and such behaviour will
            need to be implemented manually, in the event handler for this
-           button. This style is currently only supported in wxMSW, in the
-           other ports it can be specified, but single radio buttons can't be
-           turned off, making them not very useful.
+           button. This style is currently only supported in wxMSW and wxGTK,
+           in the other ports it can be specified, but single radio buttons
+           can't be turned off, making them not very useful.
     @endStyleTable
 
     @beginEventEmissionTable{wxCommandEvent}
--- a/src/gtk/radiobut.cpp
+++ b/src/gtk/radiobut.cpp
@@ -92,7 +92,22 @@ bool wxRadioButton::Create( wxWindow *parent,
         }
     }
 
-    m_widget = gtk_radio_button_new_with_label( radioButtonGroup, label.utf8_str() );
+    // GTK does not allow a radio button to be inactive if it is the only radio
+    // button in its group, so we need to work around this by creating a second
+    // hidden radio button.
+    if (HasFlag(wxRB_SINGLE))
+    {
+        m_hidden_radio_button = gtk_radio_button_new( nullptr );
+        m_widget = gtk_radio_button_new_with_label_from_widget( GTK_RADIO_BUTTON(m_hidden_radio_button), label.utf8_str() );
+        // Since this is the second button in the group, we need to ensure it
+        // is active by default and not the first hidden one.
+        gtk_toggle_button_set_active( GTK_TOGGLE_BUTTON(m_widget), TRUE );
+    }
+    else
+    {
+        m_widget = gtk_radio_button_new_with_label( radioButtonGroup, label.utf8_str() );
+    }
+
     g_object_ref(m_widget);
 
     SetLabel(label);
@@ -133,9 +148,8 @@ void wxRadioButton::SetValue( bool val )
     }
     else
     {
-        // should give an assert
-        // RL - No it shouldn't.  A wxGenericValidator might try to set it
-        //      as FALSE.  Failing silently is probably TRTTD here.
+        if (HasFlag(wxRB_SINGLE))
+            gtk_toggle_button_set_active( GTK_TOGGLE_BUTTON(m_hidden_radio_button), TRUE );
     }
 
     g_signal_handlers_unblock_by_func(


vadz:
Thanks! Looks good to me and the only thing I'd like to do is to rename the variable to use the usual wx `camelCase` convention for consistency, but I can do it when merging.

This could also be backported to 3.2 if the member variable were replaced with a (global) hash map to avoid changing the ABI if anybody is motivated to do it.

Newbytee:
> This could also be backported to 3.2 if the member variable were replaced with a (global) hash map to avoid changing the ABI if anybody is motivated to do it.

Does such a hashmap exist which I could add it to, or would I have to create it somewhere?

vadz:
> > This could also be backported to 3.2 if the member variable were replaced with a (global) hash map to avoid changing the ABI if anybody is motivated to do it.
> 
> Does such a hashmap exist which I could add it to, or would I have to create it somewhere?

No, sorry for not being clear, you'd have to create it. You'd also have to use `WX_DECLARE_VOIDPTR_HASH_MAP()` in 3.2 to cover all the build variants. But, in principle, it should be simple enough as you just need to declare this as a global variable, add the hidden radio button to it if necessary and then remove it from there in the dtor.

Newbytee:
> But, in principle, it should be simple enough as you just need to declare this as a global variable, add the hidden radio button to it if necessary and then remove it from there in the dtor.

In what file would it be suitable to declare the global variable? And, dtor? As in destructor?

vadz:
> > But, in principle, it should be simple enough as you just need to declare this as a global variable, add the hidden radio button to it if necessary and then remove it from there in the dtor.
> 
> In what file would it be suitable to declare the global variable?

`src/gtk/radiobut.cpp` would seem to be a natural choice.

> And, dtor? As in destructor?

Yes, sorry, I thought it was a standard abbreviation (just like "ctor").

Newbytee:
> Yes, sorry, I thought it was a standard abbreviation (just like "ctor").

It might be, I'm not much of a C++ person.

Makes me wonder, is it an issue that I didn't destroy the GtkButton in the destructor in this code?

vadz:
> Makes me wonder, is it an issue that I didn't destroy the GtkButton in the destructor in this code?

Oops, yes, I think you're right, we probably need to destroy it. The most straightforward way to make sure we do would be to build with ASAN and check that it

1. Reports memory leak when you create a `wxRadioButton` with `wxRB_SINGLE` style now.
2. Doesn't report it any more after adding `gtk_widget_destroy(m_hiddenButton)` call.

Could you please check if it does?

Newbytee:
> Could you please check if it does?

Sure, I can do it, but why did you reopen this?

vadz:
Just to not forget to deal with this. I should have created a new issue but was too lazy/didn't have enough time.

Newbytee:
> > Makes me wonder, is it an issue that I didn't destroy the GtkButton in the destructor in this code?
> 
> Oops, yes, I think you're right, we probably need to destroy it. The most straightforward way to make sure we do would be to build with ASAN and check that it
> 
>     1. Reports memory leak when you create a `wxRadioButton` with `wxRB_SINGLE` style now.
> 
>     2. Doesn't report it any more after adding `gtk_widget_destroy(m_hiddenButton)` call.
> 
> 
> Could you please check if it does?

It's hard to tell since there's so many memory leaks, but regardless of whether I destroy the GtkButton I get `SUMMARY: AddressSanitizer: 504919 byte(s) leaked in 12325 allocation(s).`. So I don't know if it matters.

vadz:
Are you using `misc/suppressions/lsan`?

Newbytee:
> Are you using `misc/suppressions/lsan`?

No, I just used `-fsanitize=address`. I'll have to try that.

Newbytee:
> Are you using `misc/suppressions/lsan`?

So I redid my test with those suppressions, and if I use those I don't get any leak. Just to be sure, I deliberately leaked an int with `int* q = new int` in the `Create` method of `wxRadioButton`, and it still didn't report any leak. So, maybe I'm just testing this wrong? I'm currently building a separate project which then links to wxWidgets like so:

    g++ main.cpp -fsanitize=address `<code directory>/wxWidgets/build-gtk/wx-config --libs --cxxflags` -o repro

And then running it like so:

    LSAN_OPTIONS=suppressions=<code directory>/wxWidgets/misc/suppressions/lsan ./repro

I've verified that the `Create` method and the destructor get run by adding prints in there, and the prints did get printed as I ran my code, so clearly it seems like the leaks either are being cleaned up in some way I don't understand or not detected for some reason.

vadz:
Thanks for testing this! I admit I don't fully understand why the hidden button doesn't leak, but I guess it gets destroyed when its parent is? In any case, if there is no leak, there is no problem and this can be closed again.

PR #23681, opened 2023-07-03
Bug fix in odcombo.cpp. Cached best size should be invalid after inserting new items

Rezakarimi66:


vadz:
Thanks for fixing this, merged now.

PR #23683, opened 2023-07-04
Explain git submodules in gtk install instructions

ForestJohnson:
These instructions are supposed to be easy to follow. But they don't work unless you already know how to manage git submodules, and already have all the git submodules in place at the correct versions. 

So the simple instructions on how to achieve that from scratch should be included

### DIFF
--- a/docs/gtk/install.md
+++ b/docs/gtk/install.md
@@ -11,7 +11,8 @@ The simplest case                      {#gtk_simple}
 -------------------
 
 If you compile wxWidgets on Linux for the first time and don't like to read
-install instructions just do the following in wxWidgets directory:
+install instructions just [obtain the complete source code](https://github.com/wxWidgets/wxWidgets/blob/master/README-GIT.md)
+and then do the following in wxWidgets directory:
 
     $ mkdir buildgtk
     $ cd buildgtk


PBfordev:
IMO, this would seem unnecessary, as it is:
1. Not GTK specific.
2. Already explained here: https://github.com/wxWidgets/wxWidgets/blob/master/README-GIT.md

ForestJohnson:
it's necessary because no one is going to read your entire documentation before they start trying to install it.   They are going to scan the documentation searching for a sequence of shell commands to run, and then run them.

But these commands will fail because 9 times out of 10, the person doing this will not have already known about the submodules, in fact there is a decent chance that they have never heard of a git submodule before in their life. 

So if you want to keep all the git submodule related information in one file, ok, thats fine,  lets link to it instead:

PBfordev:
Disclaimer: I am not a wxWidgets library maintainers, just a regular nosy user.

FWIW, I admit I myself know very little about GIT (but I am not a programmer nor smart or educated) and rely solely on TortoiseGIT GUI, thus being incapable of tasks beyond simple check out/switch branch/commit/push.

Still, I find hard to believe that a vast majority of people using GIT on Linux to clone an unreleased version of a C++ library have no clue about GIT submodules. I am not saying those may not exist but hopefully are the exception to the rule.

In my experience, the problem with missing GIT modules does not come from GIT users; it is usually reported by people who obtained the source using GitHub's "Download ZIP" or "Assets" archives.

OTOH, I do understand that C++ build process is not pleasant and every little thing may help. That is why (I think) _configure_ provides checks for missing submodules, for example like [this for Scintilla/Lexilla](https://github.com/wxWidgets/wxWidgets/commit/dcfe38710229dd7d04316d6834b7dec142a6d746).

ForestJohnson:
Why are you arguing with me? What is the point? 

Projects that  people may want to use, for example https://codeberg.org/forestjohnson/tenacity/src/branch/main/BUILDING.md#linux

will link to wxWidgets,  those people may try to install wxWidgets if its not already packaged for their system. Thats normal. 

Why are you fighting against making it easier to understand how to install wxWidgets?

oneeyeman1:
@ForestJohnson ,
If you are on Linux you better be using the packaged version of the library.
And if you want to work with Git master, you need to be familiar how Git operates. 
However this is not the place to educate people how the Git works - it has its own documentation. And FWIW, wxWidgets does give the information on configure about missing submidules and how to obtain then.

If Vadim think this is useful - he will apply it. Otherwise - thx for the effort.

vadz:
Thanks, I think it would be useful to mention Git submodules in the documentation more prominently as IME people indeed have trouble with them. But this patch is not great because it adds a link to the master version of `README-GIT.md` when this file itself is versioned -- and so it would make more sense to link to `README-GIT.md` of the same version.

It also seemingly implies that you have to use Git to get the sources while the other READMEs (MSW/Mac) still talk only about downloading/uncompressing the source archives. Ideally we'd say the same thing in all of them.

To fix both problems I think it would be better to link to the already existing "Getting the sources" section of the [installation overview](https://docs.wxwidgets.org/3.2/overview_install.html). AFAICS we just need to add an anchor to this section to make it possible. Could you please confirm that this works and update the PR to do it like this? TIA!

PBfordev:
> Why are you arguing with me? What is the point?

I just don't like when people lie, eh, I mean make up an outrageous statistics such as 9/10 GIT users not knowing about submodules, to push their selfish agenda.

But skimming through those outdated docs for Tenacity made me laugh, when I imagined a common computer user attempting to build the application using those instructions, so thanks for that.

vadz:
Hi @ForestJohnson, do you plan to make the changes proposed in [my comment above](https://github.com/wxWidgets/wxWidgets/pull/23683#issuecomment-1620865513) by chance?

PR #23685, opened 2023-07-05
wxWindowsPrintNativeData InitializeDevMode fix

sz5000:
Use a temporary WinPrinter if not specified via unction call in order to not fall back to the default printer code.

Do not use PRINTDLG if a printerName is specified because this returns the default printer which is not necessarily the printer we want. Therefore we have to ensure, that the code above is used and therefore we use a temporary printer if not specified.

### DIFF
--- a/src/msw/printdlg.cpp
+++ b/src/msw/printdlg.cpp
@@ -425,9 +425,8 @@ void wxWindowsPrintNativeData::InitializeDevMode(const wxString& printerName, Wi
     {
         // ensure that we have a printer object here, otherwise we are unable to determine m_devMode
         WinPrinter fallbackPrinter;
-        if (!printer) {
+        if (!printer)
             printer = &fallbackPrinter;
-        }
 
         // Open printer
         if ( printer && printer->Open( printerName ) == TRUE )
--- a/src/msw/printdlg.cpp
+++ b/src/msw/printdlg.cpp
@@ -473,7 +473,7 @@ void wxWindowsPrintNativeData::InitializeDevMode(const wxString& printerName, Wi
         }
     }
 
-    if ( !m_devMode && printerName.IsEmpty())
+    if ( !m_devMode && printerName.IsEmpty() )
     {
         // Use PRINTDLG as a way of creating a DEVMODE object
         PRINTDLG pd;
--- a/src/msw/printdlg.cpp
+++ b/src/msw/printdlg.cpp
@@ -429,7 +429,7 @@ void wxWindowsPrintNativeData::InitializeDevMode(const wxString& printerName, Wi
             printer = &fallbackPrinter;
 
         // Open printer
-        if ( printer && printer->Open( printerName ) == TRUE )
+        if ( printer->Open( printerName ) == TRUE )
         {
             DWORD dwNeeded, dwRet;
 
--- a/src/msw/printdlg.cpp
+++ b/src/msw/printdlg.cpp
@@ -423,6 +423,12 @@ void wxWindowsPrintNativeData::InitializeDevMode(const wxString& printerName, Wi
     // this replaces the PrintDlg function which creates the DEVMODE filled only with data from default printer.
     if ( !m_devMode && !printerName.IsEmpty() )
     {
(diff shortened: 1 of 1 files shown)


src/msw/printdlg.cpp
```diff
@@ -423,6 +423,12 @@ void wxWindowsPrintNativeData::InitializeDevMode(const wxString& printerName, Wi
     // this replaces the PrintDlg function which creates the DEVMODE filled only with data from default printer.
     if ( !m_devMode && !printerName.IsEmpty() )
     {
+        // ensure that we have a printer object here, otherwise we are unable to determine m_devMode
+        WinPrinter fallbackPrinter;
+        if (!printer) {
+            printer = &fallbackPrinter;
+        }
```
vadz, line 430: This should be reformatted to put braces on the correct lines or just remove them:
```suggestion
        if (!printer)
            printer = &fallbackPrinter;
```

src/msw/printdlg.cpp
```diff
@@ -423,6 +423,12 @@ void wxWindowsPrintNativeData::InitializeDevMode(const wxString& printerName, Wi
...
+
         // Open printer
         if ( printer && printer->Open( printerName ) == TRUE )
```
vadz, line 433: This check doesn't make sense any more now:
```suggestion
        if ( printer->Open( printerName ) == TRUE )
```

src/msw/printdlg.cpp
```diff
@@ -468,7 +474,7 @@ void wxWindowsPrintNativeData::InitializeDevMode(const wxString& printerName, Wi
         }
     }
 
-    if ( !m_devMode )
+    if ( !m_devMode && printerName.IsEmpty())
```
vadz, line 477: Another minor style fix:
```suggestion
    if ( !m_devMode && printerName.IsEmpty() )
```

vadz (commented):
Thanks, this looks logical to me but I don't quite understand in which circumstances did we run into a problem with the old code: could you please give an example of a situation in which things didn't work before (and how?) and work correctly now to help me understand this?

TIA!

sz5000:
The problem was the following:

if we call InitializeDevMode like this` InitializeDevMode("TestPrinter")`, then the printer parameter is NULL and this would execute the code block starting with the line  `PRINTDLG pd;`.  But this code block is just for fetching devmode for the default printer and if the default printer is not he printer we want, then we would initialize `m_devMode` with data from a wrong printer.

If we have initialized with a wrong devmode and if we then use `wxWindowsPrintNativeData::TransferFrom`,  which internally does the following
```
DocumentProperties( nullptr,
                printer,
                szPrinterName,
                (LPDEVMODE)hDevMode, // Reuse our buffer for output.
                (LPDEVMODE)hDevMode, // Pass the driver our changes
                DM_IN_BUFFER |       // Commands to Merge our changes and
                DM_OUT_BUFFER );     // write the result.
```

we pass a wrong devmode to the printer and this leads to a crash if the devmode is not like expected.

sz5000:
If we have to initialize the devmode for a specific printer and if the wxPrinter argument is NULL which is the default, then the old code falls back to

```
PRINTDLG pd;

memset(&pd, 0, sizeof(PRINTDLG));
pd.lStructSize    = sizeof(PRINTDLG);

pd.hwndOwner      = nullptr;
pd.hDevMode       = nullptr; // Will be created by PrintDlg
pd.hDevNames      = nullptr; // Ditto

pd.Flags          = PD_RETURNDEFAULT;
```

This code just returns the devmode for the default printer because `PD_RETURNDEFAULT` is used here, but the default printer can also be a different printer than the one for which we need to initialize the devmode. Then we have a wrong devmode and apps can crash, just like in my APP.

vadz:
Yes, I understood your explanation, thanks. What I was asking if it's really enough to select any non-default printer in the dialog to crash? Or if, perhaps, the hack in the commit I linked prevents the crash from happening?

I'd like to know this in order to write an accurate description of this change in the change log.

sz5000:
It depends on the printer driver and the expected driver extra data. My app did not crash with the printers I have used so far, but recently I had to install OneNote because of an issue with the oneno

PR #23691, opened 2023-07-07
add missing const accessor for WXMAC FontDialog

mcorino:
MacOSX implementation misses common, documented const accessor


vadz:
Thanks, I agree that this accessor should be present, but a better way to fix this problem and also avoid other similar problems in the future would be to derive `wxFontDialog` from `wxFontDialogBase`. This looks like a simple change to make, could you please try to update this PR to do it like this? If you run into problems, we can apply this simple fix, of course, but it seems a pity not to make a better change if it's as simple as it seems.

BTW, I think we should be able to drop `wxOSX_USE_EXPERIMENTAL_FONTDIALOG==0` code branch, as `wxOSX_USE_EXPERIMENTAL_FONTDIALOG` was added almost 15 years ago and surely it's not experimental any more? @csomor Do you mind removing it (in master only, of course)?

mcorino:
> Thanks, I agree that this accessor should be present, but a better way to fix this problem and also avoid other similar problems in the future would be to derive `wxFontDialog` from `wxFontDialogBase`. This looks like a simple change to make, could you please try to update this PR to do it like this? If you run into problems, we can apply this simple fix, of course, but it seems a pity not to make a better change if it's as simple as it seems.
> 

Ah, should have seen that right away. Done.

> BTW, I think we should be able to drop `wxOSX_USE_EXPERIMENTAL_FONTDIALOG==0` code branch, as `wxOSX_USE_EXPERIMENTAL_FONTDIALOG` was added almost 15 years ago and surely it's not experimental any more? @csomor Do you mind removing it (in master only, of course)?

Have not done this yet but could if this is ok.

mcorino:
> 
> @mcorino if you can integrate it in the same PR, that would be fine, thanks, otherwise I'll wait and do it after your merge.

@csomor I don't mind doing the work and the C++ stuff is pretty straight forward. But I have no idea what to do with the `src/osx/carbon/fontdlgosx.mm` file which also references the `wxOSX_USE_EXPERIMENTAL_FONTDIALOG` define.

csomor:
> @csomor I don't mind doing the work and the C++ stuff is pretty straight forward. But I have no idea what to do with the `src/osx/carbon/fontdlgosx.mm` file which also references the `wxOSX_USE_EXPERIMENTAL_FONTDIALOG` define.

.mm behave like c++ files so you can just remove everything starting from the `#else` in line 291, but as I said you can also just finish your PR and I'll do it aftwards, no problem at all ..

mcorino:
Cleanup done as well.

mcorino:
> Cleanup done as well.

Okay...that did not work as expected.
Reverted the fontdlgosx.mm changes as that seems more complicated. It seems the `#if wxOSX_USE_EXPERIMENTAL_FONTDIALOG` bracketed code is never built actually (unlike in fontdlg.cpp)  but instead the `#else` branch is.
Someone more knowledgeable than me with respect to these .mm files will have to have a look here.

csomor:
> > Cleanup done as well.
> 
> Okay...that did not work as expected. Reverted the fontdlgosx.mm changes as that seems more complicated. It seems the `#if wxOSX_USE_EXPERIMENTAL_FONTDIALOG` bracketed code is never built actually (unlike in fontdlg.cpp) but instead the `#else` branch is. @csomor Someone more knowledgeable than me with respect to these .mm files will have to have a look here.

@mcorino sorry for the trouble, I'll do that, strange, since I just did a test with line 36 changed to

`#if 1 // wxOSX_USE_EXPERIMENTAL_FONTDIALOG`

and this built correctly, while a `#if 0 `did not, but I'll do a PR

vadz:
@csomor It looks like this resulted in problems in wxiOS builds, see e.g. [this build log](https://github.com/wxWidgets/wxWidgets/actions/runs/5494695671/jobs/10013631448).

Let me merge the first 2 commits here for now, if you don't mind.

csomor:
> @csomor It looks like this resulted in problems in wxiOS builds, see e.g. [this build log](https://github.com/wxWidgets/wxWidgets/actions/runs/5494695671/jobs/10013631448).
> 
> Let me merge the first 2 commits here for now, if you don't mind.

@vadz, yes, that's fine, I've opened #23694 for further exploration

PR #23692, opened 2023-07-07
add delays to make UIActionSimulator work on MacOS Ventura (13.4.1)

mcorino:
On my MacOS Ventura 13.4.1 system wxUIActionSimulator did not work (in tests like Button::Click). 
Based on suggestions from internet search results I added (short) delays after posting events and that seems to do the trick well enough. After these changes the failing tests work again.
I know this is not the prettiest solution but looking at the internet search results it is probably the best I am able to get as it seems a LOT of people have trouble with this functionality on MacOS and none seem to have a better solution.

### DIFF
--- a/src/osx/uiaction_osx.cpp
+++ b/src/osx/uiaction_osx.cpp
@@ -154,6 +154,8 @@ bool wxUIActionSimulatorOSXImpl::MouseDown(int button)
     wxCFEventLoop* loop = dynamic_cast<wxCFEventLoop*>(wxEventLoop::GetActive());
     if (loop)
         loop->SetShouldWaitForEvent(true);
+
+    ::usleep(useconds_t(10 * 1000));
     
     return true;
 }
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, I don't mind adding the delays if this is what it takes to wait (although I'd obviously prefer some more elegant solution, of course, if anybody knows of one), but could we at least add a helper function doing this sleep instead of repeating it multiple times? This shouldn't be difficult and would make the code more maintainable in the future.

TIA!

mcorino:
> Thanks, I don't mind adding the delays if this is what it takes to wait (although I'd obviously prefer some more elegant solution, of course, if anybody knows of one), 

I absolutely agree. Unfortunately I could not find one. Tried one other suggestion that did not work. This did. Also, making the delay too short (1 * 1000) fixed things mostly but not always suggesting this really is some sort of timing issue.

> but could we at least add a helper function doing this sleep instead of repeating it multiple times? This shouldn't be difficult and would make the code more maintainable in the future.
> 

Done.

vadz:
Thanks for the update! Will merge it in a moment.

Issue #23671, opened 2023-06-28
wxCocoa doesn't catch EVT_KEY_DOWN for ⌘+⎇+Letter


Upliner:
### Description
Minimal app to reproduce the problem:
```
#include <wx/wx.h>
struct MyApp : wxApp {
    void OnKeyDown(wxKeyEvent& ev) {
        std::cerr << ev.GetKeyCode() << " " << ev.GetModifiers() << std::endl;
    }
    bool OnInit() override {
        auto frame = new wxFrame(nullptr, wxID_ANY, "Hello World");
        new wxPanel(frame);
        Bind(wxEVT_KEY_DOWN, &MyApp::OnKeyDown, this);
        frame->Show(true);
        return true;
    }
};
wxIMPLEMENT_APP(MyApp);
```
When I press Command+Option+L it prints
```
308 2
307 3
```
When I press L alone it prints `76 0`, so for  Command+Option+L it expected to print `76 3`. Other apps such as Key Codes by manytricks display this key combination correctly, and some other apps has this combination as shortcut for some commands, but I can't use it in my app.

### Platform and version information

- wxWidgets version  you use: 3.2.2
- wxWidgets port  you use: wxOSX
- OS  and its version: macOS Ventura 13.4.1

vadz:
308 and 307 are `WXK_CONTROL` and `WXK_ALT` which are both expected (the first is for pressing Command, the second one is for pressing Option), but I don't know why we don't generate any key events for `L` itself.

Running your program with `WXTRACE=keyevent` in the environment might give some clue and debugging this would be welcome, of course -- we need all the help with wxOSX that we can get.

Upliner:
OK I've run it with wxtrace, wxWidgets sees the keypress, but doesn't send KeyDown event to my widgets. Here's log
[wx-keyevent.txt](https://github.com/wxWidgets/wxWidgets/files/11919531/wx-keyevent.txt)

vadz:
Thanks, but I am afraid I still don't know what goes wrong here. AFAICS from looking at the trace messages, we do get to this line, as expected: https://github.com/wxWidgets/wxWidgets/blob/a5c0a817989aa31390ef79a20938f3f3d00473ac/src/osx/cocoa/window.mm#L3885 but `[NSView interpretKeyEvents:]` apparently doesn't do anything with such keys.

@csomor Should we add some test for modifiers above and call `OSXHandleKeyEvent()` for Command+Option keys ourselves?

[...]

PR #23699, opened 2023-07-10
macOS emulate wxKeyDown

csomor:
Testbed for trying to send wx key down event when interpretKeyEvents does not result in such an event, see #23671


vadz:
Stefan, should this be merged or is it still WIP (maybe it should be converted to "draft" then?)?

csomor:
> Stefan, should this be merged or is it still WIP (maybe it should be converted to "draft" then?)?

no, it's all there is :-) I just thought that the OP might test whether it solves the problem for him before we attempt a merge, but otherwise go ahead ...

vadz:
@Upliner do you plan to test this in the near future?

vadz:
I've merged this now with a small extra `wxLogTrace()` which might be useful for debugging any problems in the future.

This should probably be also backported to 3.2 later if it can be confirmed that it works and doesn't result in any regressions.

Thanks again!

Issue #23700, opened 2023-07-10
Errors compiling some samples against installed wxWidgets on Linux


swt2c:
There appear to be some issues compiling some of the samples against an installed wxWidgets version on Linux (ie, build + install wx 3.2.2 and then use the makefile.unx to compile each sample.)

The console-based samples (archive, console, ipc, sockets) are failing to link with errors like this:
```
`wx-config --cxx` -o console console_console.o     `wx-config --unicode=yes --static=no --toolkit=gtk3 --version=3.2 --libs net,base`
/usr/bin/ld: console_console.o: warning: relocation against `_ZTV17wxMDIClientWindow' in read-only section `.text._ZN17wxMDIClientWindowC2Ev[_ZN17wxMDIClientWindowC5Ev]'
/usr/bin/ld: console_console.o: in function `wxWindowBase::GetBestVirtualSize() const':
console.cpp:(.text._ZNK12wxWindowBase18GetBestVirtualSizeEv[_ZNK12wxWindowBase18GetBestVirtualSizeEv]+0x25): undefined reference to `wxWindowBase::GetBestSize() const'
/usr/bin/ld: console_console.o: in function `wxWindowBase::CanBeFocused() const':
console.cpp:(.text._ZNK12wxWindowBase12CanBeFocusedEv[_ZNK12wxWindowBase12CanBeFocusedEv]+0x31): undefined reference to `wxWindowBase::IsEnabled() const'
/usr/bin/ld: console_console.o: in function `wxWindowBase::HasCapture() const':
console.cpp:(.text._ZNK12wxWindowBase10HasCaptureEv[_ZNK12wxWindowBase10HasCaptureEv]+0xd): undefined reference to `wxWindowBase::GetCapture()'
/usr/bin/ld: console_console.o: in function `wxWindowBase::GetDefaultAttributes() const':
console.cpp:(.text._ZNK12wxWindowBase20GetDefaultAttributesEv[_ZNK12wxWindowBase20GetDefaultAttributesEv]+0x28): undefined reference to `wxWindowBase::GetClassDefaultAttributes(wxWindowVariant)'
/usr/bin/ld: console_console.o: in function `wxWindowBase::DoGetBestClientSize() const':
console.cpp:(.text._ZNK12wxWindowBase19DoGetBestClientSizeEv[_ZNK12wxWindowBase19DoGetBestClientSizeEv]+0xb): undefined reference to `wxDefaultSize'
<snipped>
```

[...]

PR #23701, opened 2023-07-10
Fix building some samples in out of tree build

vadz:
See #23700.


swt2c:
Thanks for the quick fix!  One note on the console fix, the archive sample's makefile.unx doesn't seem to have been updated, so might need a tweak for that.

PR #23712, opened 2023-07-14
Read Linux distribution info from os-release file

swt2c:
The Linux distribution community has somewhat deprecated the lsb_release utility and has standardized on a new file, os-release, that can be simply parsed to get the same information.  Attempt to read this file in /etc/os-release, then /usr/lib/os-release, and finally, fall back to using the lsb_release utility if neither of those files are found.

See: https://www.freedesktop.org/software/systemd/man/os-release.html
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2184391

### DIFF
--- a/src/unix/utilsunx.cpp
+++ b/src/unix/utilsunx.cpp
@@ -60,6 +60,8 @@
 #include "wx/evtloop.h"
 #include "wx/mstream.h"
 #include "wx/private/fdioeventloopsourcehandler.h"
+#include "wx/config.h"
+#include "wx/filename.h"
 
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, this is definitely useful as I've already ran into the problems due to `lsb_release` absence in other (non-wx) contexts myself. I still have no idea why did it have to be deprecated, but I'm pretty sure it's much simpler to apply this patch than to do anything else, so I'll just do this (and backport it to 3.2 too).

Thanks again!

PR #23713, opened 2023-07-14
Add support for building WebView with libwebkit2gtk-4.1 (3.2 backport)

swt2c:
Note that the first commit also contains the cmake fixup in 4704138.


vadz:
Thanks, I'll merge this but I'll regenerate configure with 2.69 instead of 2.71 used here as this results in tons of unwanted changes there.

vadz:
Merged now, thanks again.

PR #23715, opened 2023-07-18
Document that RUNTIME_LIBS works only when building with MSVC

PBfordev:
I considered also mentioning that the same effect can be achieved for GCC (no idea about clang) by passing `LDFLAGS="-static"`, but that is probably outside the document scope.

### DIFF
--- a/docs/msw/install.md
+++ b/docs/msw/install.md
@@ -345,9 +345,8 @@ The full list of the build settings follows:
 
 * `RUNTIME_LIBS=static`
 
-  Links static version of C and C++ runtime libraries into the executable, so
-  that the program does not depend on DLLs provided with the compiler (e.g.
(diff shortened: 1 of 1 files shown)


PBfordev:
Sorry, I had no recollection of editing that document in the master. 

Still, that change was probably suboptimal, as "(VC++ only.)" should have been used there as well, for consistency with other similar options (I must have missed it back then).

I know it is more work for you, sorry, but IMO: (1) This PR should be merged as is and (2) the description for `RUNTIME_LIBS` in the master changed to match that from this PR (of course, I will do it via another PR), if you think it is for the best (the documents in the two branches have already diverged anyway).

vadz:
OK, I'll apply this to master and to 3.2, thanks.

vadz:
This one too.

PR #23723, opened 2023-07-20
CI: Check building samples out of tree

### LINKED ISSUE #23722: Check for building samples out of tree in a CI build
Also, do you think it is worth add a CI check for this?  It looks like the minimal sample is already checked at the very end of `ci.yml` but it shouldn't be too much more effort to check them all.  Probably just run something like:
```sh
make -C samples clean
for f in `find samples -name makefile.unx`; do
	mv $f `dirname $f`/Makefile
done
make -C samples
```

_Originally posted by @swt2c in https://github.com/wxWidgets/wxWidgets/issues/23700#issuecomment-1631086927_

swt2c:
Fixes: #23722

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,6 +120,7 @@ jobs:
             configure_flags: --with-cxx=20 --enable-utf8 --enable-utf8only
             use_asan: true
             use_xvfb: true
+            build_out_of_tree_samples: true
           - name: Ubuntu 18.04 wxX11
             runner: ubuntu-latest
             container: ubuntu:18.04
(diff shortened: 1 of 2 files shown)


swt2c:
OK, so I ended up going with close to my original proposal for a couple reasons:
1) The top-level samples Makefile knows which samples shouldn't be compiled (e.g., taskbarbutton which is MSW only).
2) -j can be used

Unfortunately, this ended up finding another bug, with webrequest (sorry, in my earlier test, I wasn't compiling wx with wxWebRequest support).  I can't see any obvious problem with the bakefile for the webrequest sample as to why the makefile.unx isn't linking with `net`.

vadz:
> OK, so I ended up going with close to my original proposal for a couple reasons:

Sorry for bad advice.

> Unfortunately, this ended up finding another bug, with webrequest (sorry, in my earlier test, I wasn't compiling wx with wxWebRequest support). I can't see any obvious problem with the bakefile for the webrequest sample as to why the makefile.unx isn't linking with `net`.

It took me some time to find out, but it's due to this:

https://github.com/wxWidgets/wxWidgets/blob/44e857c0120ce92f589f6de70e90893e70226c98/build/bakefiles/wxpresets/presets/wx_unix.bkl#L187-L188

So just exchanging the order fixes the problem, I'll do it and will re-apply this PR on top of it soon, thanks!

PR #23726, opened 2023-07-21
Avoid debug warnings when defocusing menu bar in wxGTK

vadz:
At least under Wayland, removing focus from the menu consistently
results in "window lost focus even though it didn't have it" debug
messages because wxMenuBar never gets any focus-in signals but does get
focus-out one.

Avoid this by just not connecting handlers for these signals at all, as
we don't need them for wxMenuBar which isn't supposed to generate any
focus events anyhow.

### DIFF
--- a/src/gtk/window.cpp
+++ b/src/gtk/window.cpp
@@ -2997,7 +2997,15 @@ void wxWindowGTK::PostCreation()
 
     // focus handling
 
-    if (!GTK_IS_WINDOW(m_widget))
+    // Check for GTKNeedsParent() || IsTopLevel() is a hack: it catches the
+    // case of wxMenuBar, which isn't supposed to generate any focus events,
+    // and which is the only non-TLW which returns false from this function.
+    //
+    // The TLW check overlaps with !GTK_IS_WINDOW() check, but it's not 100%
+    // obvious if GTK_IS_WINDOW() and wxWindow::IsTopLevel() are really exactly
+    // equivalent, so for now ensure we don't change the existing check which
+    // only used !GTK_IS_WINDOW().
+    if (!GTK_IS_WINDOW(m_widget) && (GTKNeedsParent() || IsTopLevel()))
     {
         if (m_focusWidget == nullptr)
             m_focusWidget = m_widget;


vadz:
@paulcor Please let me know if I'm doing something stupid here.

vadz:
I've pushed this, but if anybody has any comments, they're still welcome, of course.

src/gtk/window.cpp
```diff
@@ -2997,7 +2997,15 @@ void wxWindowGTK::PostCreation()
...
-    if (!GTK_IS_WINDOW(m_widget))
+    // Check for GTKNeedsParent() || IsTopLevel() is a hack: it catches the
+    // case of wxMenuBar, which isn't supposed to generate any focus events,
+    // and which is the only non-TLW which returns false from this function.
+    //
+    // The TLW check overlaps with !GTK_IS_WINDOW() check, but it's not 100%
+    // obvious if GTK_IS_WINDOW() and wxWindow::IsTopLevel() are really exactly
+    // equivalent, so for now ensure we don't change the existing check which
+    // only used !GTK_IS_WINDOW().
+    if (!GTK_IS_WINDOW(m_widget) && (GTKNeedsParent() || IsTopLevel()))
```
paulcor, line 3008: Sorry I'm a bit late here, but this test looks... weird. If you want to be conservative, `wxMenuBar` could be excluded specifically: `&& !IsKindOf(wxCLASSINFO(wxMenuBar))`. And I don't understand the need for `|| IsTopLevel()`. For `wxMenuBar` both `GTKNeedsParent()` and `IsTopLevel()` return false, while for all others `GTKNeedsParent()` is equivalent to `!IsTopLevel()`, so you have `!IsTopLevel() || IsTopLevel()`.

vadz: Thanks for checking this!

The idea here was to preserve the behaviour of the existing code for all the classes other than `wxMenuBar` because I wasn't sure if `GTK_IS_WINDOW(m_widget)` is always equivalent to `IsTopLevel()`. I do know that if `GTK_IS_WINDOW()` is true, then `IsTopLevel()` is true because only windows of classes deriving from `wxTopLevelWindowBase` create windows in GTK sense. But I'm not sure if the converse is true, i.e. what if something overrides `IsTopLevel()` to return `true`  but the widget is not a `GtkWindow`? In this case the test would pass before, but if I just added `&& GTKNeedsParent()` it wouldn't pass any more, which is why I did this complicated thing.

If we decide that overriding `IsTopLevel()` to return true without actually using `GtkWindow` is not allowed, then this could be simplified to just `if (GTKNeedsParent())` and I'd be glad to do it if you confirm that you don't see any situations in which doing this could make sense. But I preferred to err on the side of caution by default.

PR #23727, opened 2023-07-22
Fixed constant changes in the progress dialog size if the message keeps changging

oleg68:
This PR fixes an undesired side effect of https://github.com/wxWidgets/wxWidgets/commit/4101849b4d0a4e1291e45cd990611eb9408c68a1:

According to the commit message, the author did not want the progress dialog to change in it's size when the message was becoming shorter.

But the original code

```
const wxSize sizeOld = m_msg->GetSize();

        m_msg->SetLabel(newmsg);

        if ( m_msg->GetSize().x > sizeOld.x )
```

had wrong behavior if the message at first was long (L3), then bacame short (L1) and then became middle (L2), L1 < L2 < L3. In this case the size was changing from L3 to L2, because the code compared the new width (L2) with the previous one (L1) instead of with the maximal (L3).

The attached test shows a wrong behavior without this patch and a right behavior with this patch: 
[WxProgressDialogTest.zip](https://github.com/wxWidgets/wxWidgets/files/12137149/WxProgressDialogTest.zip)

### DIFF
--- a/include/wx/generic/progdlgg.h
+++ b/include/wx/generic/progdlgg.h
@@ -182,6 +182,8 @@ private:
     wxGauge *m_gauge;
     // the message displayed
     wxStaticText *m_msg;
+    // the message width in pixels. If the new width is greater, the dialog grows
+    unsigned m_msgWidth;
     // displayed elapsed, estimated, remaining time
     wxStaticText *m_elapsed,
                  *m_estimated,
--- a/src/generic/progdlgg.cpp
+++ b/src/generic/progdlgg.cpp
@@ -82,6 +82,7 @@ void wxGenericProgressDialog::Init()
 
     m_gauge = nullptr;
     m_msg = nullptr;
+    m_msgWidth = 0;
     m_elapsed =
     m_estimated =
     m_remaining = nullptr;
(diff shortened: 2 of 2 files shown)


vadz:
Thank you for the patch!

I agree that the current code is wrong, but I think I know why: it didn't take into account the fact that `wxStaticText` is the only control which changes its size on its own when its label changes. And I think it would be better to just prevent it from changing here, if only because this avoids the addition of a new member variable and means the change can be backported to 3.2 without breaking the ABI there.

So I've pushed an alternative fix which makes your test case work too. Please let me know if you see any problems with it!

rousseldenis:
@oleg68 Could you check the other fix?

oleg68:
@rousseldenis Yes, I've done it. The second fix works.

@vadz Thank you for your fix. Could you submit the same fix to 3.2?

vadz:
> @vadz Thank you for your fix. Could you submit the same fix to 3.2?

Yes, I will backport the fix to 3.2, this was the main reason for doing it, rather than applying your PR.

I'm closing it now, as it has been superseded by the other change.

PR #23730, opened 2023-07-23
Modernize configure

vadz:
This replaces #23640 by doing some of the same changes manually, so that the diffs remain relatively readable.

It also uses pkg-config for GTK detection as there should be no reason not to do it any longer.


vadz:

PR #23735, opened 2023-07-25
Fix samples out of tree build 3.2

swt2c:


vadz:
Looks good, thanks! I will merge it together with the other backports soon.

PR #23744, opened 2023-07-29
Enable wxStyledTextCtrl unit test in static builds too

vadz:
The original problem preventing this from being done may have been fixed
in bde95c388c (Fix linking of the GUI test when using static libraries,
2021-01-30).


vadz:
If nobody sees anything wrong with this, I'll merge it in order to be able to add more wxSTC tests in the future.

PR #23747, opened 2023-08-01
Add wxUILocale::GetMonthName() and GetWeekDayName()

vadz:
Unlike in master, only the variants for the date/time formatting context are returned to reduce the number of changes in this branch. This is wrong for standalone month names like in the generic calendar control. However, in the prior implementation that was also the case.

Use a horrible hack to avoid adding new virtual functions to wxUILocaleImpl class which, although private, still somehow results in ABI breakage.

See #23556, #23581.

cc @utelle

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -247,6 +247,7 @@ All:
 - Fix memory leak when using wx from non-wx threads (Antti Nietosvaara, #23535).
 - Use grep instead of fgrep and egrep in makefiles/scripts (#23537).
 - Use /etc/os-release if lsb_release isn't available (Scott Talbert, #23712).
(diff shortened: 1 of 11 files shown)


utelle:
@vadz Great that you found a way to overcome the ABI breakage.

utelle:
The AppVeyor check for toolset "mingw" fails due to missing definition of symbol `LOCALE_RETURN_GENITIVE_NAMES`. This symbol should be defined for Windows 7 and higher, but obviously the headers coming with mingw do not define the symbol.

If the symbol is not defined the definition `#define LOCALE_RETURN_GENITIVE_NAMES 0x10000000` could be added.

vadz:
Thanks, done this now (and also in master, just in case...).

PR #23752, opened 2023-08-03
Invalidate the current cell coords when inside table changed in wxGrid::SetTable

### LINKED ISSUE #23751: wxGrid memorizes(focus on) old coords when having table changed inside
### Description

There is a fixed wxGrid("grid" in the following) and a dynamic wxGridTableBase("table" in the following) inside of it.
The table size would change due to some biz logic, e.g. apply some filter on it.
The bug occurs when 
1. you have **clicked** a cell with large row number
2. the table shrinks 
3. you try to click the cell with smaller row number.
At that time, it tries to select a block using the old coords which you selected before shrink and which is invalid since the row never exist, and lead to the assertion failure.
The assertion fails at [src\generic\grid.cpp:5124](https://github.com/wxWidgets/wxWidgets/blob/a4dec24c6dab69551f5bc1b39c8e32aae9c890d1/src/generic/grid.cpp#L5124).

- OS  and its version: windows 10

### Reproduction
Demo code:
```cpp
#include <wx/wx.h>
#include <wx/scrolwin.h>
#include <wx/grid.h>


class MyTable : public wxGridTableBase
{
public:
    MyTable(int rows) : _rows{ rows }, _cols{ 4 }  {}

    virtual int GetNumberRows() wxOVERRIDE { return _rows; }
    virtual int GetNumberCols() wxOVERRIDE { return _cols; }
    virtual wxString GetValue(int row, int col) wxOVERRIDE
    {
        return wxString::Format("%d", row*col);
    }
    virtual void SetValue(int, int, const wxString&) wxOVERRIDE
    {
        wxFAIL_MSG("shouldn't be called");
    }
    virtual wxString GetColLabelValue(int col) wxOVERRIDE
    {
        // notice that column parameter here always refers to the internal
        // column index, independently of its position on the screen
        static const char* labels[] = { "Name", "Extension", "Size", "Date" };
        return labels[col];
    }
    virtual void SetColLabelValue(int, const wxString&) wxOVERRIDE
    {
        wxFAIL_MSG("shouldn't be called");
    }
private:
    int _rows;
    int _cols;
};

{
public:
    MyFrame() : wxFrame(nullptr, wxID_ANY, "Test")
    {
        SetBackgroundColour(*wxBLUE);
        wxBoxSizer* vertical = new wxBoxSizer(wxVERTICAL);
        m_grid = new wxGrid(this, wxID_ANY);
        m_table = new MyTable(20);
        m_grid->SetTable(m_table, true);
        wxButton* halve = new wxButton(this, wxID_ANY, "Halve");
        halve->Bind(wxEVT_BUTTON, [&](wxCommandEvent& event)
            {
                m_table = new MyTable(10);
                m_grid->SetTable(m_table, true);
                m_grid->ForceRefresh();
            });
        vertical->Add(halve, 0, wxCENTER);
        vertical->AddSpacer(20);
        vertical->Add(m_grid, 0, wxEXPAND);
        SetSizer(vertical);
        vertical->SetSizeHints(this);
    }
private:
    wxGrid* m_grid = nullptr;
    MyTable* m_table = nullptr;
};

class MyApp : public wxApp
{
public:
    bool OnInit() override
    {
        (new MyFrame())->Show();
        return true;
    }
}; wxIMPLEMENT_APP(MyApp);
```

Steps:
1. select the cell with row larger than 10 to get grid cursor on that
5. click button "halve" to shrink the table size
6. click (or drag) in the cells in the new smaller table
7. Error occurs -> it tried to access the coords in the original big table before thus causing invalid index

CookieLau:
Fix #23751 

In case of table inside grid has been changed, the function call of IsVisible should be cautious of the index validation.

### DIFF
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -2911,6 +2911,7 @@ wxGrid::SetTable(wxGridTableBase *table,
     }
 
     InvalidateBestSize();
+    m_currentCellCoords = wxGridNoCellCoords;
 
     return m_created;
 }


vadz:
Thanks, but unfortunately I don't think this is the correct fix.

I do think a check for the parameters validity is warranted, but IMO it should be using `wxCHECK()`, i.e. it shouldn't be legal to call `IsVisible()` with invalid indices.

And I also think that this just hides the current bug without really fixing it, i.e. the internal grid state is still inconsistent, with something not being updated when the table changes. I think the correct fix must be to just clear/reset whatever we forget to reset currently in `wxGrid::SetTable()`. If you could please try to find what it is and change this PR to do it instead, it would be very welcome!

Thanks in advance!

CookieLau:
Push a fixed version to just invalidate the current cell coords in SetTable as the coords information is useless when it corresponding table has gone.

vadz:
Copying the comment from the issue into the PR just in case you missed it there:

Sorry, after thinking more about this, I wonder if we really want to always reset it here: in some circumstances, it could be unwanted, e.g. we probably do _not_ want to lose the current selection if we just extended the table.

So I think that we need to call `UpdateCurrentCellOnRedim()` here instead of just invalidating the current cell. Could you please confirm that this also fixes the bug? If so, I can commit it directly (or you can update the associated PR to do this).

TIA!

CookieLau:
Hi @vadz , I've replied you here https://github.com/wxWidgets/wxWidgets/issues/23751 , feel free to commit it dierctly.

Issue #23757, opened 2023-08-06
wxMemoryInputStream::Peek does not set last read count


the-nerbs:
### Description
The comments on `wxInputStream::Peek()` indicate that it will cause `wxInputStream::LastRead()` to return 1 when it successfully reads a byte from the stream, or 0 if it does not. From [wx/stream.h](https://github.com/wxWidgets/wxWidgets/blob/a4dec24c6dab69551f5bc1b39c8e32aae9c890d1/include/wx/stream.h#L103C1-L109C25):
```cpp
    // blocks until something appears in the stream if necessary, if nothing
    // ever does (i.e. EOF) LastRead() will return 0 (and the return value is
    // undefined), otherwise 1
```

However, `wxMemoryInputStream`'s implementation of `Peek` does not affect the value returned by `LastRead`. Comparing the implementation ([link](https://github.com/wxWidgets/wxWidgets/blob/a4dec24c6dab69551f5bc1b39c8e32aae9c890d1/src/common/mstream.cpp#L108C1-L120C2)) to the base `wxInputStream` implementation, it looks like `wxMemoryInputStream` might just be missing assignments to `wxStreamBase::m_lastcount` in either the success or failure cases.

Reproduction code:
```cpp
#include <wx/wx.h>
#include <wx/mstream.h>

int main(int argc, char *argv[])
{
    char data[4] = { 'T', 'e', 's', 't' };

    wxMemoryInputStream stream(data, sizeof(data));

    char nextChar = stream.Peek();
    std::size_t readCount = stream.LastRead();

    // fails for readCount being 0
    wxASSERT(readCount == 1);

    // passes
    wxASSERT(nextChar == 'T');
}
```

### Platform and version information

- wxWidgets version you use: 3.2.2.1
- wxWidgets port you use: wxMSW
- OS and its version: Windows 7

PBfordev:
Regarding `Peek()`, it is interesting that:
1. It is not officially documented how it affects `LastRead()`.
2. It setting `LastRead()` to `1` would be different from C++ `istream`'s `peek()` and `gcount()`, where calling `gcount()` after `peek()` [returns `0`](https://en.cppreference.com/w/cpp/io/basic_istream/gcount).

Quick testing with some other simple `wxInputStream`s shows that indeed, `LastRead()` is set to `1`, running
```cpp
#include <wx/wx.h>
#include <wx/mstream.h>
#include <wx/sstream.h>
#include <wx/wfstream.h>

[...]

PR #23758, opened 2023-08-07
Update wxMemoryInputStream::Peek to set last read count.

the-nerbs:
Fix for #23757 

This updates the implementation of `wxMemoryInputStream::Peek` to set the last read count to 1 if it successfully retrieved a byte from the stream, or 0 if it did not. This change is to align the memory stream's behavior with that of the other stream types - `wxFileInputStream`, `wxFFileInputStream`, or `wxStringInputStream` for a few examples.

### DIFF
--- a/tests/streams/bstream.h
+++ b/tests/streams/bstream.h
@@ -253,9 +253,17 @@ protected:
         while (stream_in.IsOk())
         {
             char peekChar = stream_in.Peek();
+            size_t peekLastRead = stream_in.LastRead();
+
             char getChar = stream_in.GetC();
+
+            // Peek and GetC should retrieve the same 0 or 1 characters.
+            CPPUNIT_ASSERT_EQUAL(peekLastRead, stream_in.LastRead());
+
             if (stream_in.LastRead() == 1)
+            {
                 CPPUNIT_ASSERT_EQUAL(getChar, peekChar);
+            }
         }
     }
 
--- a/src/common/mstream.cpp
+++ b/src/common/mstream.cpp
@@ -112,10 +112,12 @@ char wxMemoryInputStream::Peek()
     if ( pos == m_length )
     {
         m_lasterror = wxSTREAM_READ_ERROR;
+        m_lastcount = 0;
 
         return 0;
     }
 
+    m_lastcount = 1;
     return buf[pos];
 }
 


tests/streams/bstream.h
```diff
@@ -253,9 +253,17 @@ template <class TStreamIn, class TStreamOut> class BaseStreamTestCase : public C
         while (stream_in.IsOk())
         {
             char peekChar = stream_in.Peek();
+            size_t peekLastRead = stream_in.LastRead();
```
the-nerbs, line 256: I want to highlight this test update as something I'm not entirely sure about, since I believe it would impose this behavior of `Peek` on all stream types. This seems to be true at the moment, though as PB mentioned on the issue, this behavior is not formally documented.

It is mentioned by the in-source comments on `wxInputStream::LastRead` that `Peek` affects it, and on `wxInputStream::Peek` that the last read count is set to 0 or 1 though, which is why I thought to update the common tests here instead of just adding one against the memory stream specifically.

I'm wondering if I should change this test to just be against the memory stream, or perhaps this change might warrant an update to the doxygen comments for `wxInputStream::Peek` and/or `LastRead` as well?

vadz: It does make sense to impose the same behaviour for all stream types, however, and as all the existing tests pass, it seems that this is already the case, so I think it's fine to do this.

vadz (approved):
Thanks for fixing this, I agree that it's a change for the better and so will merge it in master soon if there are no objections. I probably won't merge it into 3.2 however just in case somebody relies on the old behaviour there (even though I have no idea why would anybody do it).

I'll also update `Peek()` documentation by basically copying the source comment contents there.

PBfordev:
> I'll also update `Peek()` documentation by basically copying the source comment contents there.

FWIW, I now also tested input streams in the socket client sample (in `DoDownload()`), where `LastRead()` after `Peek()` returns `1`.

But I have also tried `wxZipInputStream` where `LastRead()` after `Peek()` returns `0`.  Not sure why, since `wxZipInputStream` is derived from `wxArchiveInputStream` whose `Peek()` just [returns `wxInputStream::Peek()`](https://github.com/wxWidgets/wxWidgets/blob/cc5f5f90b08b72c7cf21fd4238c6d3bc2e156394/include/wx/archive.h#L102). But maybe I am doing something wrong or `Peek()` does not really make sense for a zipped stream. The test code was basically
```cpp
        wxFileSystem::AddHandler(new wxZipFSHandler);

        const wxString            zipFileName("test.zip#zip:\\test.txt");
        wxFileSystem              zipArchive;
        std::unique_ptr<wxFSFile> zipFile(zipArchive.OpenFile(zipFileName, wxFS_READ));

        if ( zipFile )
        {
            std::unique_ptr<wxInputStream> insz(zipFile->DetachStream());

            if ( insz )
                DoPeek(insm, "wxZipInputStream");
        }
```

vadz:
Thanks for testing, this looks like a bug even though I don't see why does it happen neither from just looking at the code. It would be nice to fix it, but this looks like a separate issue, so I'll create one for it and still merge this one.

PR #23764, opened 2023-08-11
Implement wxGTK3 dark mode

ColinKinloch:
Read dark mode state from DBus `org.freedesktop.appearance.color-scheme` and set `GtkSettings` `gtk-application-prefer-dark-theme` property.

### To Do
* ~[ ] Keep DBus connection (`wxDarkModeModule`?)~
  * ~[ ] Listen for `color-scheme` changes~ [libadwaita](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/d7307a4b0da7a47de2cc1e84f23e45d604aa00c7/src/adw-settings-impl-portal.c)
* [x] `IsDark` fallback when DBus is unavailable
* [x] ~Look into GTK2 dark mode~
* [x] Other `wxSystemAppearance` functions?

### DIFF
--- a/src/gtk/settings.cpp
+++ b/src/gtk/settings.cpp
@@ -25,6 +25,22 @@
 #include "wx/gtk/private/win_gtk.h"
 #include "wx/gtk/private/stylecontext.h"
 #include "wx/gtk/private/value.h"
+#include "wx/gtk/private/error.h"
+
+// xdg-desktop-portal provides a DBus interface for dark mode
+// https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Settings
+#define FDO_PORTAL_DESKTOP_BUS_NAME "org.freedesktop.portal.Desktop"
+#define FDO_PORTAL_DESKTOP_OBJECT_PATH "/org/freedesktop/portal/desktop"
+#define FDO_PORTAL_DESKTOP_SETTINGS_INTERFACE "org.freedesktop.portal.Settings"
+
+enum ColorScheme
+{
+    FDO_COLOR_DEFAULT,
+    FDO_COLOR_PREFER_DARK,
+    FDO_COLOR_PREFER_LIGHT
+};
+
+
 
 bool wxGetFrameExtents(GdkWindow* window, int* left, int* right, int* top, int* bottom);
 
@@ -839,6 +855,55 @@ wxFont wxSystemSettingsNative::GetFont( wxSystemFont index )
 
     return font;
 }
+
+#if defined(__WXGTK3__) || defined(__WXGTK__)
+bool wxSystemAppearance::IsSystemDark() const
+{
+    GVariant *out = nullptr;
+    GVariant *child = nullptr;
+    GVariant *ret = nullptr;
+    wxGtkError error;
+
+    bool isDark = IsDark();
+
+    GDBusProxy *proxy = g_dbus_proxy_new_for_bus_sync(G_BUS_TYPE_SESSION,
+        G_DBUS_PROXY_FLAGS_NONE,
+        nullptr,
+        FDO_PORTAL_DESKTOP_BUS_NAME,
+        FDO_PORTAL_DESKTOP_OBJECT_PATH,
+        FDO_PORTAL_DESKTOP_SETTINGS_INTERFACE,
+        nullptr,
+        error.Out());
+
+    if (proxy)
+    {
+        ret = g_dbus_proxy_call_sync(proxy, "Read",
+            g_variant_new ("(ss)", "org.freedesktop.appearance", "color-scheme"),
+            G_DBUS_CALL_FLAGS_NONE, G_MAXINT, nullptr, nullptr);
+
+        if (ret)
+        {
+            g_variant_get(ret, "(v)", &child);
+            g_variant_get(child, "v", &out);
+
+            if (g_variant_get_uint32(out) == FDO_COLOR_PREFER_DARK)
+                isDark = true;
+        }
+    }
+
+    g_variant_unref (child);
+    g_variant_unref (ret);
+    g_variant_unref (out);
+    g_clear_object(&proxy);
+
+    return isDark;
+}
+
+bool wxSystemAppearance::AreAppsDark() const
+{
+    return IsSystemDark();
+}
+#endif // __WXGTK3__ || __WXGTK__
 
 // helper: return the GtkSettings either for the screen the current window is
 // on or for the default screen if window is null
--- a/src/common/settcmn.cpp
+++ b/src/common/settcmn.cpp
@@ -69,7 +69,7 @@ void wxSystemSettings::SetScreenType( wxSystemScreenType screen )
 // ----------------------------------------------------------------------------
 
 // wxMSW has its own implementation of these functions.
-#if !defined(__WXMSW__)
+#if !defined(__WXMSW__) && !defined(__WXGTK3__) && !defined(__WXGTK__)
 
 bool wxSystemAppearance::AreAppsDark() const
 {
@@ -81,7 +81,7 @@ bool wxSystemAppearance::IsSystemDark() const
     return IsDark();
 }
 
-#endif // !__WXMSW__
+#endif // !__WXMSW__ && !__WXGTK3__ && !__WXGTK__
 
 #if !defined(__WXOSX__)
 
--- a/src/gtk/toplevel.cpp
+++ b/src/gtk/toplevel.cpp
@@ -26,6 +26,7 @@
 #ifndef WX_PRECOMP
     #include "wx/frame.h"
     #include "wx/app.h"     // GetAppDisplayName()
+    #include "wx/settings.h"
     #include "wx/icon.h"
     #include "wx/log.h"
 #endif
@@ -895,6 +896,12 @@ bool wxTopLevelWindowGTK::Create( wxWindow *parent,
     // old colours etc.
     g_signal_connect_after(gtk_settings_get_default(), "notify::gtk-theme-name",
         G_CALLBACK(notify_gtk_theme_name), this);
+
+#ifdef __WXGTK3__
+    GtkSettings *settings = gtk_widget_get_settings(GTK_WIDGET(m_widget));
+    g_object_set(settings, "gtk-application-prefer-dark-theme",
+        wxSystemSettings::GetAppearance().IsSystemDark(), nullptr);
+#endif
 
     return true;
 }


src/gtk/settings.cpp
```diff
@@ -840,6 +849,31 @@ wxFont wxSystemSettingsNative::GetFont( wxSystemFont index )
     return font;
 }
 
+    bool wxSystemAppearance::IsDark() const
+    {
+        GVariant* out = NULL;
```
ColinKinloch, line 854:
```suggestion
        GVariant* out = nullptr;
```

src/gtk/settings.cpp
```diff
@@ -840,6 +849,31 @@ wxFont wxSystemSettingsNative::GetFont( wxSystemFont index )
...
+        g_autoptr (GVariant) ret =
+                g_dbus_proxy_call_sync(proxy, "Read",
+                                       g_variant_new ("(ss)", "org.freedesktop.appearance", "color-scheme"),
+                                       G_DBUS_CALL_FLAGS_NONE, G_MAXINT, nullptr, nullptr);
+
+        g_autoptr (GVariant) child = nullptr;
+        g_variant_get(ret, "(v)", &child);
+        g_variant_get(child, "v", &out);
+
+        printf("heloo %s\n", g_variant_print(out, true));
```
ColinKinloch, line 873:
```suggestion
```

vadz:
Thanks for working on this! Listening to the colours change event would be definitely nice, but if it's not easy to implement, perhaps we should merge the changes here already and leave it for a separate PR?

ColinKinloch:
Yeah, I think it would be pretty trivial to implement.
However probably an easy job for someone more familiar with on the codebase than I am.

Here's the relevant file in libadwaita: https://gitlab.gnome.org/GNOME/libadwaita/-/blob/d7307a4b0da7a47de2cc1e84f23e45d604aa00c7/src/adw-settings-impl-portal.c

src/gtk/settings.cpp
```diff
@@ -840,6 +853,53 @@ wxFont wxSystemSettingsNative::GetFont( wxSystemFont index )
     return font;
 }
 
+#if defined(__WXGTK3__) || defined(__WXGTK__)
+bool wxSystemAppearance::IsSystemDark() const
+{
+    GVariant *out = nullptr;
+    GVariant *child = nullptr;
+    GVariant *ret = nullptr;
+    GError *error = nullptr;
```
vadz, line 862: We could/should use `wxGtkError` instead of dealing with raw Glib types.

src/gtk/settings.cpp
```diff
@@ -26,6 +26,19 @@
 #include "wx/gtk/private/stylecontext.h"
 #include "wx/gtk/private/value.h"
 
+#define PORTAL_BUS_NAME "org.freedesktop.portal.Desktop"
+#define PORTAL_OBJECT_PATH "/org/freedesktop/portal/desktop"
+#define PORTAL_SETTINGS_INTERFACE "org.freedesktop.portal.Settings"
+
+enum ColorScheme
```
vadz, line 33: Interesting, the only documentation for this I could find was https://docs.flatpak.org/fr/latest/portal-api-reference.html
but this doesn't seem to be flatpak-specific. Do you know of some better/more authoritative documentation for this? I'd like to add a link to it to a comment here somewhere.

src/gtk/settings.cpp
```diff
@@ -840,6 +853,53 @@ wxFont wxSystemSettingsNative::GetFont( wxSystemFont index )
...
+    if (proxy) {
+        ret = g_dbus_proxy_call_sync(proxy, "Read",
+            g_variant_new ("(ss)", "org.freedesktop.appearance", "color-scheme"),
+            G_DBUS_CALL_FLAGS_NONE, G_MAXINT, nullptr, nullptr);
+
+        if (ret) {
+            g_variant_get(ret, "(v)", &child);
+            g_variant_get(child, "v", &out);
+
+            isDark = g_variant_get_uint32(out) == FDO_PREFER_DARK;
```
vadz, line 884: I'm not 100% sure if we should set it to false for `FDO_DEFAULT`, it could be better to fallback to `IsDark()` in this case.

src/gtk/settings.cpp
```diff
@@ -840,6 +853,53 @@ wxFont wxSystemSettingsNative::GetFont( wxSystemFont index )
...
+    GDBusProxy *proxy = g_dbus_proxy_new_for_bus_sync(G_BUS_TYPE_SESSION,
+        G_DBUS_PROXY_FLAGS_NONE,
+        nullptr,
+        PORTAL_BUS_NAME,
+        PORTAL_OBJECT_PATH,
+        PORTAL_SETTINGS_INTERFACE,
+        nullptr,
+        &error);
+
+    if (proxy) {
```
vadz, line 875:
```suggestion
    if (proxy)
    {
```

src/gtk/settings.cpp
```diff
@@ -840,6 +853,53 @@ wxFont wxSystemSettingsNative::GetFont( wxSystemFont index )
...
+        ret = g_dbus_proxy_call_sync(proxy, "Read",
+            g_variant_new ("(ss)", "org.freedesktop.appearance", "color-scheme"),
+            G_DBUS_CALL_FLAGS_NONE, G_MAXINT, nullptr, nullptr);
+
+        if (ret) {
```
vadz, line 880:
```suggestion
        if (ret)
        {
```

vadz (commented):
I think we should merge this already (after fixing the trivial issues below), please let me know if anybody -- notably @paulcor -- has any objections.

src/gtk/settings.cpp
ColinKinloch: The xdg-desktop-portal code is hosted on [github](https://github.com/flatpak/xdg-desktop-portal) which links to: https://flatpak.github.io/xdg-desktop-portal/

paulcor:
So, the whole point of this is really just to have `wxSystemAppearance::IsSystemDark()` return the correct value, right? I think that could be done much more simply by examining the theme name; if it ends with "-dark", or the `GTK_THEME` environment variable ends with ":dark", it's dark.

ColinKinloch:
`GTK_THEME` isn't in my environment and isn't set by gnome settings.
If `GTK_THEME` were set gtk would ignore `gtk-application-prefer-dark-mode` and select whichever theme/variant it was set to.

ColinKinloch:
Looking at [libadwaita](https://gitlab.gnome.org/search?group_id=8&nav_source=navbar&project_id=14079&repository_ref=main&scope=blobs&search=adw_settings_impl_set_features&search_code=true) there is also the GSettings interface with the field `color-scheme` on the `org.gnome.desktop.interface` schema.

ColinKinloch:
There's also `gsettings get org.gnome.desktop.interface gtk-theme` which yields `'Adwaita'` on my dark mode system.

vadz:
FWIW `gsettings get org.gnome.desktop.interface color-scheme` returns `prefer-dark` for me in dark mode and `default` otherwise with GNOME 43.6 (Debian Bookworm). Should we just use this?

ColinKinloch:
So support GSettings when the portal fails or just not support flatpak?

vadz:
> So try GSettings when the portal fails or just not support flatpak?

Sorry but I don't know much about flatpak -- is GSettings unavailable when using it?

ColinKinloch:
```
flatpak run --command=gsettings org.freedesktop.Platform get org.gnome.desktop.interface color-scheme
```
Yields `'default'`.
I had a look just now, I couldn't figure out the specific incantation of permissions to give it access to the hosts GSettings.

paulcor:
The theme name is available from `GtkSettings` as the `gtk-theme-name` property.

vadz:
> The theme name is available from `GtkSettings` as the `gtk-theme-name` property.

But this one is always "Adwaita" for me, both in light (default) and dark modes.

paulcor:
How are you enabling dark mode?

vadz:
> How are you enabling dark mode?

Going to GNOME Settings and choosing "Dark" style in the "Appearance" page.

imciner2:
Is any of this compatible with non-Gnome desktop environments? Mate doesn't have dedicated settings to select dark mode, so to do dark mode you actually switch the theme to something else. For instance, I have a theme for dark and a theme for light, and it would appear to be defined inside a different gsettings key: `org.mate.interface gtk-theme`, which when I query it returns the theme:
```
$ gsettings get org.mate.interface gtk-theme
'Mint-Y-Dark-Blue'
```
(or at least the controls theme, since I actually have a custom theme that uses that for the controls portion). Querying the gnome one would just give the result for the Gnome DE that I don't use (but still have installed):
```
$ gsettings get org.gnome.desktop.interface color-scheme
'default'
```

So reading the gnome key would be bad for my system, because then it would try to use the appearance of a desktop environment that isn't actually the one being used.

vadz:
@imciner2 Does using portal work under Mate? If so, then we should probably just do it as it's supposed to be the WM/desktop-agnostic way to check this, IIUC.

For the reference, I'm using

```
dbus-send --session --print-reply=literal --dest=org.freedesktop.portal.Desktop /org/freedesktop/portal/desktop org.freedesktop.portal.Settings.Read string:org.freedesktop.appearance string:color-scheme
```

to check this.

ColinKinloch:
The desktop portal for mate is here: https://github.com/linuxmint/xdg-desktop-portal-xapp
There's a commit from March that says Mate doesn't support dark mode then: https://github.com/linuxmint/xdg-desktop-portal-xapp/commit/b342a7fcff2541e7ac67480e0660dd5cf17be575

imciner2:
> Does using portal work under Mate? If so, then we should probably just do it as it's supposed to be the WM/desktop-agnostic way to check this, IIUC.

Using the command you posted, this is the output I get on my Fedora 38 install:

```
$ dbus-send --session --print-reply=literal --dest=org.freedesktop.portal.Desktop /org/freedesktop/portal/desktop org.freedesktop.portal.Settings.Read string:org.freedesktop.appearance string:color-scheme
   variant       variant          uint32 0
```

I'm running a dark color scheme based on the `Mint-Y-Dark-Blue` control theme to give a dark mode. According to the spec (https://github.com/flatpak/xdg-desktop-portal/blob/d7a304a00697d7d608821253cd013f3b97ac0fb6/data/org.freedesktop.impl.portal.Settings.xml#L33-L45), `0` means "No preference," so it doesn't indicate one way or the other on my system, but I would want the app to use a dark mode.

ColinKinloch:
If I'm understanding correctly the portal implementation for Mate returns "No preference" because Mate is yet to implement dark mode.
It looks like Mate gives you a theme setting which this MR wont override.

paulcor:
> > The theme name is available from `GtkSettings` as the `gtk-theme-name` property.
> 
> But this one is always "Adwaita" for me, both in light (default) and dark modes.

Based on the [GTK source code](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtksettings.c#L3311), I'm having difficulty understanding how this is possible. So I have to ask, are you sure? Could you check with this?

```diff
diff --git a/src/gtk/toplevel.cpp b/src/gtk/toplevel.cpp
index 13b1857c6b..4fc1bdd66f 100644
--- a/src/gtk/toplevel.cpp
+++ b/src/gtk/toplevel.cpp
@@ -507,6 +507,15 @@ gtk_frame_window_state_callback( GtkWidget* WXUNUSED(widget),
 }
 }
 
+static void f()
+{
+    char* theme_name;
+    gboolean prefer_dark;
+    g_object_get(gtk_settings_get_default(), "gtk-theme-name", theme_name,
+        "gtk-application-prefer-dark-theme", &prefer_dark, nullptr);
+    fprintf(stderr, "%s %d\n", theme_name, prefer_dark);
+    g_free(theme_name);
+}
 //-----------------------------------------------------------------------------
 // "notify::gtk-theme-name" from GtkSettings
 //-----------------------------------------------------------------------------
@@ -514,6 +523,7 @@ gtk_frame_window_state_callback( GtkWidget* WXUNUSED(widget),
 extern "C" {
 static void notify_gtk_theme_name(GObject*, GParamSpec*, wxTopLevelWindowGTK* win)
 {
+    f();
     wxSysColourChangedEvent event;
     event.SetEventObject(win);
     win->HandleWindowEvent(event);
@@ -896,6 +906,7 @@ bool wxTopLevelWindowGTK::Create( wxWindow *parent,
     g_signal_connect_after(gtk_settings_get_default(), "notify::gtk-theme-name",
         G_CALLBACK(notify_gtk_theme_name), this);
 
+    f();
     return true;
 }
 
```

vadz:
> > > The theme name is available from `GtkSettings` as the `gtk-theme-name` property.
> > 
> > 
> > But this one is always "Adwaita" for me, both in light (default) and dark modes.
> 
> Based on the [GTK source code](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtksettings.c#L3311), I'm having difficulty understanding how this is possible.

Doesn't GNOME use GTK 4 nowadays?

> So I have to ask, are you sure? Could you check with this?

I've checked (after replacing `theme_name` with `&theme_name` to avoid crashing) and it always outputs `Adwaita 0` for me (whether I start it in light or dark mode) and doesn't react to the theme change.

It looks like there is a new mechanism for dark mode support: previously it was indeed just another theme but now they use "color scheme" parameter (and also [accent colours](https://tech.lgbt/@orowith2os/110953058120399161)) without changing the theme at all -- presumably to be more like macOS, as usual...

ColinKinloch:
GTK doesn't integrate with system dark mode. Its development seems to be under the umbrella of libadwaita which is more of a opinionated human interface design library.
Since wx had a dark mode API I think this MR provides a good first step for a wxGTK implementation.

paulcor:
> it always outputs Adwaita 0 for me (whether I start it in light or dark mode) and doesn't react to the theme change.

OK, I didn't realize the app is not reacting to this Gnome "dark style". It turns out GTK3 doesn't work with whatever Gnome is using now. How very typical. GTK4 does work though, and it doesn't appear to be using the portal `color-scheme` thing, nor does it use `libadwaita`. I'd like to understand how GTK4 is getting this dark mode indication before proceeding, and leave the Flatpak issue aside for the moment.

ColinKinloch:
What are you using to test GTK4? I was using `gtk4-demo`

paulcor:
I also used `gtk4-demo`. Although I'm not sure what I did, as I just re-checked and it appears GTK4 doesn't work either...

ColinKinloch:
It occurred to me that `wxTopLevelWindowGTK::Create` probably isn't the correct place to handle dark mode, since `gtk-application-prefer-dark-theme` is a property of `GtkSettings`.
Is there a more appropriate place in wxGTK to handle global settings/events? Somewhere a `GDBusProxy` for the lifetime of an application?

vadz:
Sorry, I'm rather lost here by now. Paul, are you going to take some decision about this or should I?

paulcor:
I'm working on it.

paulcor:
The default `wxSystemAppearance::IsDark()` seems to work fine for wxGTK, so I don't see a strong need to implement that. But using the "color-scheme" setting to get an actual dark theme seems more important, so I've done that now in 9b9ec141fb.

PR #23765, opened 2023-08-14
Improve the combo sample

PBfordev:
In the dialog, create the controls put in the static box sizer with the sizer's static box as the parent instead of the dialog.

Improve appearance of custom drawn items in dark mode.

Could be backported to 3.2.


vadz:
Thanks, will merge soon!

PBfordev:
Please don't merge ATM, I will change the code in one place to use `wxSystemSettings::SelectLightDark()`.

PR #23771, opened 2023-08-14
Prefer binary packages when installing httpbin

swt2c:
This should hopefully work around the issue of greenlet (dependency of httpbin) failing to compile with gcc 4.8.  Installing binary packages is preferably anyway, as it should be faster.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,8 +172,9 @@ jobs:
               #     be built from source with gcc 4.8 any longer (its build
               #     script uses -Wdate-time option unknown to this compiler).
               #   - xvfb is used for running the GUI tests.
+              #   - httpbin's dependencies can't be built by gcc 4.8 any more.
               apt-get update -qq
-              apt-get install -qq coreutils ${compiler-g++} git make pkg-config python3 python3-cffi python3-pip sudo xvfb
+              apt-get install -qq coreutils ${compiler-g++} git make pkg-config python3 python3-cffi python3-pip python3-httpbin sudo xvfb
               ;;
 
             '')
--- a/build/tools/httpbin.sh
+++ b/build/tools/httpbin.sh
@@ -8,12 +8,19 @@ httpbin_launch() {
 
     echo 'Launching httpbin...'
 
+    # Get Ubuntu codename if it exists
+    if command -v lsb_release > /dev/null; then
+        codename=$(lsb_release --codename --short)
+    fi
+
     # Installing Flask 2.1.0 and its dependency Werkzeug 2.1.0 results
     # in failures when trying to run httpbin, so stick to an older but
     # working version.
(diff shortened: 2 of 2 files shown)


vadz:
Is it worth installing pip supporting this option or would it just open yet another can of worms?

swt2c:
> Is it worth installing pip supporting this option or would it just open yet another can of worms?

Maybe, but yes it might open a new can of worms.  The other option I was considering was just installing the distribution packaged version of httpbin (on Linux only) and not install anything with pip.  Any objections to that?

vadz:
Well, we used to do this but stopped because it had problems too, but this was with the really ancient Ubuntu versions that we don't support any more, so maybe it's fine to return to this.

I'll take anything that works at this stage, I have no idea how to get rid of `-Wdate-time` error which breaks all the CI builds right now. As usual, I'm impressed how things just break on their own when we didn't change literally anything (and I even have an alibi: I was on vacation!).

swt2c:
> Well, we used to do this but stopped because it had problems too, but this was with the really ancient Ubuntu versions that we don't support any more, so maybe it's fine to return to this.
> 
> I'll take anything that works at this stage, I have no idea how to get rid of `-Wdate-time` error which breaks all the CI builds right now. As usual, I'm impressed how things just break on their own when we didn't change literally anything (and I even have an alibi: I was on vacation!).

This PR is an attempt to fix the `-Wdate-time` error.  It came up because one of httpbin's dependencies no longer has a wheel for the Python version on Ubuntu 18.04, so it has to be compiled, which it can't with gcc 4.8.

Welcome back from vacation!

vadz:
> This PR is an attempt to fix the `-Wdate-time` error.

I hoped so, but thanks for confirming. In the meanwhile, I had a look at what has really changed and it seems like there was a weird new release of httpbin on August 11, so I wonder if we can just revert to the old one, especially as I have no idea about where did the new release come from, seeing that there haven't been any changes in the official repository.

Unfortunately my PR is still in the queue because ccache has also stopped working while I wasn't looking and so every build now recompiles everything from scratch, which means that it takes an hour instead of ~10 minutes. And, as usual with this CI stuff, I have no idea what has changed.

> Welcome back from vacation!

Thanks, it definitely won't happen again any time soon, I still have ~50 wx emails to look at...

swt2c:
> Unfortunately my PR is still in the queue because ccache has also stopped working while I wasn't looking and so every build now recompiles everything from scratch, which means that it takes an hour instead of ~10 minutes. And, as usual with this CI stuff, I have no idea what has changed.

Closing this one in favor of yours #23800 which actually works.

> > Welcome back from vacation!
> 
> Thanks, it definitely won't happen again any time soon, I still have ~50 wx emails to look at...

Eh, there's always time for vacations...there's only a limited time for those, and an unlimited number of wx things to do.  :-)

PR #23773, opened 2023-08-14
Samples darkmode

PBfordev:
Attempt to ensure that there is always a good contrast between the foreground and background colours both in light and dark color system modes. I left alone the samples that may use a white canvas but the drawings are readable (e.g., the font sample).

There may be better solutions for this (and better hard-coded colours), so suggestions are welcome. I thought wx had an API for getting a contrasting colour for a `wxColour` but I failed to find it.

Tested only on Windows 10. I have run into other issues (see my Issues spree) and some where I am not sure whether the issue is with sample or wxWidgets but I think it is enough for now.

EDIT: The PR (and eventual merge commit) title should be "Improve samples appearance in dark mode".

### DIFF
--- a/samples/ownerdrw/ownerdrw.cpp
+++ b/samples/ownerdrw/ownerdrw.cpp
@@ -297,7 +297,7 @@ OwnerDrawnFrame::OwnerDrawnFrame(wxFrame *frame, const wxString& title,
     for ( ui = 0; ui < WXSIZEOF(aszChoices); ui += 2 )
     {
 #if defined(__WXMSW__) && !defined(__WXUNIVERSAL__)
-        m_pListBox->GetItem(ui)->SetBackgroundColour(wxColor(200, 200, 200));
+        m_pListBox->GetItem(ui)->SetBackgroundColour(*wxBLUE);
 #endif
     }
 
@@ -344,6 +344,11 @@ OwnerDrawnFrame::OwnerDrawnFrame(wxFrame *frame, const wxString& title,
         if ( ui == 4 )
         {
             pListBox->GetItem(ui)->SetBackgroundColour(wxColor(0, 0, 0));
+        }
+        else
+        if ( ui == 5 ) // black on dark gray is barely visible
+        {
+            pListBox->GetItem(ui)->SetBackgroundColour(*wxYELLOW);
         }
     }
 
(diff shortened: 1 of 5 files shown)


samples/popup/popup.cpp
```diff
@@ -109,8 +109,13 @@ SimpleTransientPopup::SimpleTransientPopup( wxWindow *parent, bool scrolled )
                                               wxBORDER_NONE |
                                               wxPU_CONTAINS_CONTROLS )
 {
+    wxColour colour(*wxLIGHT_GREY);
+
+    if ( wxSystemSettings::GetAppearance().IsDark() )
+        colour.Set(90, 90, 90); // dark grey
+
```
vadz, line 116: It would probably be better to use `SelectLightDark()` here, as it's shorter and also better practice, I think.

samples/ownerdrw/ownerdrw.cpp
```diff
@@ -345,6 +345,11 @@ OwnerDrawnFrame::OwnerDrawnFrame(wxFrame *frame, const wxString& title,
         {
             pListBox->GetItem(ui)->SetBackgroundColour(wxColor(0, 0, 0));
         }
+        else
+        if ( ui == 5 ) // black on dark gray is barely visible
+        {
+            pListBox->GetItem(ui)->SetBackgroundColour(*wxYELLOW);
+        }
```
vadz, line 352: This indentation seems strange, let's write it like this
```suggestion
        else if ( ui == 5 ) // black on dark gray is barely visible
        {
            pListBox->GetItem(ui)->SetBackgroundColour(*wxYELLOW);
        }
```

vadz (commented):
Thanks, this is definitely useful, but I'd just like to tweak this slightly before merging -- if you can do it, please do, otherwise I'll do it myself a bit later.

samples/ownerdrw/ownerdrw.cpp
PBfordev: Done.

samples/popup/popup.cpp
PBfordev: Thanks, I somehow missed the function.

samples/popup/popup.cpp
```diff
@@ -216,6 +218,9 @@ void SimpleTransientPopup::OnMouse(wxMouseEvent &event)
     rect.SetWidth(1000000);
     wxColour colour(*wxLIGHT_GREY);
 
+    if ( wxSystemSettings::GetAppearance().IsDark() )
+        colour.Set(90, 90, 90); // dark grey
+
```
vadz, line 223: Sorry for being so annoying, but could we use `SelectLightDark()` here too please?

PBfordev: You have nothing to apologize for, it's me who has to apologize for being so sloppy.

PR #23780, opened 2023-08-18
wxUniChar: Support C++11 char16_t and char32_t

vvs31415:


vadz:
Thanks, but I wonder what is the actual effect of this change. `wxUniChar` itself is somewhat obsolete if we can use `char32_t`, so I'm not sure whether it's worth to add support for the latter to the former... And we need at least some unit tests checking that this actually works as expected, as I'm not sure about the actual level of `char32_t` support. Could you please add something?

Thanks in advance!

vvs31415:
This could be the first step towards the consistent use of `char8_t`, `char16_t`, and `char32_t` types. Here we provide super easy conversions between `wxUniChar` and `char32_t`. It would be fine to get rid of `wxUniChar` altogether, but we still need few functions to check the correctness of its content.

vadz:
I think we need to discuss the global approach here. In particular, what's going to happen to `wxUniCharRef`.

Having a conversion between `wxUniChar` and `char32_t` is probably not a problem but it just doesn't seem that useful to me on its own...

vvs31415:
The problem with `wxUniChar` is that it does nontrivial conversions to and from 8-bit char types. Perhaps we should prohibit such conversions, mandating to use From8bit() and To8bit() explicitly. This way we can have a single char32_t constructor.

vvs31415:
The issue with `wxUniCharRef` is too big for this small ticket, but I would suggest to make `UniChar()` public and strip the class from all convenience members which are using it.
Also, strip both `wxUniChar` and `wxUniCharRef` from all implicit conversions to non-character types.

PR #23782, opened 2023-08-18
CYGWIN: Tests: add test case for Intl

carlo-bramini:
While running the `test_base` on CYGWIN, this message will be printed on the console:

```
/cygdrive/c/Users/carlo/Documents/GitHub/wxWidgets/tests/intl/intltest.cpp:44

/cygdrive/c/Users/carlo/Documents/GitHub/wxWidgets/tests/intl/intltest.cpp:199: FAILED:
  REQUIRE( FRENCH_LONG_DATE_FMT == NormalizeFormat(wxLocale::GetInfo(wxLOCALE_LONG_DATE_FMT)) )
with expansion:
  "%A %d %B %Y" == "%a %e %b %Y"
with message:
  French long date
```

It looks like that NEWLIB needs its own signatures for letting these test to be passed.
This patch adds the right test case for CYGWIN, for letting the execution of the test.


vadz:
And this one too, thanks!

PR #23784, opened 2023-08-19
Fix creating wxRadioBox with no items in wxUniv

Kvaz1r:
Without changes, in case where radiobox haven't items invalid value is passing to SetSelection.


vadz:

PR #23789, opened 2023-08-21
Improve wxFileSystem::AddHandler() documentation

PBfordev:
The note in the method description incorrectly stated that "...the handlers are deleted in wxFileSystem's destructor so that you don't have to care about it." However, unless a handler is removed by calling wxFileSystem::RemoveHandler() (or undocumented wxFileSystem::CleanUpHandlers()), it is deleted only when wxWidgets shuts down in wxFileSystemModule::OnExit().

Update the method documentation that it is correct (and similar to that of wxImage::AddHandler()).

### DIFF
--- a/interface/wx/filesys.h
+++ b/interface/wx/filesys.h
@@ -44,19 +44,19 @@ public:
     wxFileSystem();
 
     /**
-        This static function adds new handler into the list of handlers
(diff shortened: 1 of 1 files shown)


vadz:
Thanks! Will merge soon.

PR #23792, opened 2023-08-23
Update maximum supported compiler versions in main readme

PBfordev:
TBH, I am not sure if this information is useful, I would not expect wxWidgets build break with newer ("unsupported") compiler versions and prefer listing just the minimal supported version (clang is missing it).

Moreover, the readme has: "All C++11 compilers are supported including but not limited to:" I think there are no other compilers besides the three listed (Borland and Intel do not have their own anymore), are they?

~~And now I notice the commit text wrong, with an extra "Update" in it. :(~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,8 @@ This version of wxWidgets supports the following primary platforms:
 All C++11 compilers are supported including but not limited to:
 
 - Microsoft Visual C++ 2015 or later (up to 2022).
-- g++ 4.8 or later (up to 12), including MinGW/MinGW-64/TDM under Windows.
-- Clang (up to 14).
+- g++ 4.8 or later (up to 13), including MinGW/MinGW-64/TDM under Windows.
+- Clang (up to 16).
 
 Please use [3.2 branch](https://github.com/wxWidgets/wxWidgets/tree/3.2) if
 you must use wxWidgets with a C++98 compiler or support Windows XP.


imciner2:
> (Borland and Intel do not have their own anymore)

Intel still has their own - `icpx` and `icx` are their new compilers for C++ and C respectively (built on LLVM like Clang, but with Intel-specific changes and optimizations).

The current head of master builds fine with these as well.

PBfordev:
> > (Borland and Intel do not have their own anymore)
> 
> Intel still has their own - `icpx` and `icx` are their new compilers for C++ and C respectively

I see, thanks. 

But I also meant they do not require anything specific from wxWidgets side (as e.g. BCC or DMC did with build or code paths), they are probably just fine using whatever there is for gcc/clang...

lanurmi:
> And now I notice the commit text wrong, with an extra "Update" in it. :(

So amend the commit and fix the text, and force-push.

PBfordev:
> > And now I notice the commit text wrong, with an extra "Update" in it. :(
> So amend the commit and fix the text, and force-push.

Good idea, thanks. I have already explained my GIT disability many times here. I agree that it is shameful but it is what it is.

vadz:
Sorry, not sure what happened here, but there are now no changes at all. I can apply the other PR but I guess you wanted to make some other changes?

@PB please don't hesitate to ask if you have any questions about Git, there is nothing shameful about not knowing something about it (I don't know everything neither, in spite of using it for close to 15 years daily) and I'd be glad to help.

PBfordev:
> Sorry, not sure what happened here, but there are now no changes at all. 

Should be fixed now.

> @PB please don't hesitate to ask if you have any questions about Git

Thanks, but that's on me. I did read the Git book at least twice, but since I am not using Git regularly (perhaps aside from like the basic four commands), I forget most of it soon.

vadz:
Thanks!

The idea behind mentioning the max versions is to show that we've tested wx with them and that the project is actually maintained and hasn't been abandoned 10 years ago as some people think, so I think it's worth doing.

PR #23797, opened 2023-08-23
Fix undefined behavior in wxImage::Paste with wxIMAGE_ALPHA_BLEND_COM…

### LINKED ISSUE #23791: wxImage::Paste with wxIMAGE_ALPHA_BLEND_COMPOSE relies on undefined NaN to integer conversion
The following test fails on the riscv64 architecture:
```
wxImage::Paste
  Paste semitransparent image over transparent image
./image/image.cpp:1931

./image/image.cpp:1945: FAILED:
  CHECK_THAT( actual, RGBSimilarTo(wxImage("image/paste_result_no_background_square_over_circle.png"), 1) )
with expansion:
  image of size 100*100 with alpha doesn't have the same RGB data as image of
  size 100*100 with alpha: first mismatch is at (0, 0) which has value
  0xffffff00 instead of the expected 0x00000000
```

I expect the test to pass like on other architectures.

### Description

I have tracked the bug to an undefined NaN to integer conversion in wxImage::Paste with wxIMAGE_ALPHA_BLEND_COMPOSE, and more precisely in the following code:

```
                        for (int c = 3 * i; c < 3 * (i + 1); c++)
                        {
                            target_data[c] =
                                (unsigned char)(((source_data[c] * source_alpha +
                                    target_data[c] * light_left) /
                                result_alpha) + 0.5f);
                        }
```

When `source_alpha`, `light_left` and `result_alpha` are all `0.0` like in the testcase, the result is a NaN. It is then casted to an `unsigned char`. This is an undefined operation, which happens to give the expected result on x86 and a few other architectures, but not on riscv64.

The correct way is probably to test for a NaN before casting the value.

- wxWidgets version 3.2.2
- wxWidgets port wxGTK
- OS Debian GNU/Linux sid on riscv64

+ GTK version: 3.24.38
+ Which GDK X11

aurel32:
…POSE

The wxImage::Paste method, when used with wxIMAGE_ALPHA_BLEND_COMPOSE, converts floating point values, which might not be finite, to unsigned char using a simple cast. According to the C++ standard this is an undefined behaviour:

  When a finite value of real floating type is converted to an integer
  type other than _Bool, the fractional part is discarded (i.e., the
  value is truncated toward zero). If the value of the integral part
  cannot be represented by the integer type, the behavior is undefined.

This causes the failure of the "Paste semitransparent image over transparent image" test on a few architectures, including RISC-V.

This commits fixes that by assigning 0 if the value is not finite, to match the x86 behavior.

Fixes: #23791

### DIFF
--- a/src/common/image.cpp
+++ b/src/common/image.cpp
@@ -1743,10 +1743,11 @@ wxImage::Paste(const wxImage & image, int x, int y,
                         alpha_target_data[i] = (unsigned char)((result_alpha * 255) + 0.5f);
                         for (int c = 3 * i; c < 3 * (i + 1); c++)
                         {
-                            target_data[c] =
-                                (unsigned char)(((source_data[c] * source_alpha +
-                                    target_data[c] * light_left) /
-                                result_alpha) + 0.5f);
+                            float target_value =
+                                ((source_data[c] * source_alpha + target_data[c] * light_left) /
+                                result_alpha) + 0.5f;
+                            target_data[c] = wxFinite(target_value) ? (unsigned char)target_value
+                                                                    : 0;
                         }
                     }
                 }


vadz:
Thanks for the fix but I wonder if we shouldn't be handling `result_alpha == 0` case specially? It looks wrong to divide by 0 and then check whether the result is finite, it seems like it would be better to avoid doing it in the first place, wouldn't it?

paulcor:
Yeah, I think it's better to avoid the divide by zero. Done now in 496a9a40a1.

PR #23798, opened 2023-08-23
Enable some warnings

vvs31415:


vadz:
Thanks, I'm a bit surprised we're not getting these warnings any more, but if we don't, we should leave them enabled, of course, so I'll merge this.

PR #23800, opened 2023-08-23
Pin httpbin to the previously working 0.7.0 version

vadz:
The recently released 0.10.0 doesn't seem to have any actual changes but
has new dependencies which broke the previously working builds.


swt2c:
Ah, I missed the fact that httpbin had actually released a new version (didn't even think to check that) so that's why greenlet started being pulled in all of a sudden.  Yeah, this should probably fix it.

PR #23801, opened 2023-08-23
dynamiclib test: Add more glibc versions

sthibaul:
alpha and ia64 use libc.so.6.1, GNU/kFreeBSD uses libc.so.0.1, and GNU/Hurd uses libc.so.0.3


vadz:
Thanks! It would probably be better to rewrite this test in some better way to avoid this brute-force approach, but for now I'll apply this fix.

PR #23808, opened 2023-08-24
Test: Make sure the correct wxString ctor is called with wide literal

vvs31415:
The compiler may decide to pick a char overload when a wchar_t literal like L'A' is supplied:
    void foo(char);
    foo(L'A'); // no warning!
    auto wch = L'A';
    foo(wch); // MSVC warning C4244
If wxString lacked the wchar_t constructor, the test would not notice that.

### DIFF
--- a/tests/strings/strings.cpp
+++ b/tests/strings/strings.cpp
@@ -125,9 +125,10 @@ TEST_CASE("StringConstructors", "[wxString]")
     CHECK( wxString("Hello", 4) == "Hell" );
     CHECK( wxString("Hello", 5) == "Hello" );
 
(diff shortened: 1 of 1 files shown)


vadz:
Sorry, I'm totally lost here too... What does this test actually test for? We didn't have any warnings before this patch, we (hopefully) shouldn't have them with it neither, so what is the point? I'd understand changing the text to use a non-Unicode character rather than `Z` to check that it is handled correctly, but this change just doesn't make sense to me.

vvs31415:
Sorry, I am totally wrong

PR #23810, opened 2023-08-24
Fix CMake build issues after enabling manifest from rc file

MaartenBent:
This should fix the CMake CI build errors.

Note that `test[_gui].vcxproj` also seems to need a similar fix as https://github.com/wxWidgets/wxWidgets/commit/f6255e456b4d83f09168a2731c32da44a1f73ff8 to fix all CI build errors.


vadz:
Thanks, I've fixed the GUI test project too (I don't think the non-GUI version needs it as it doesn't use `sample.rc`), let's see if this has finally been enough.

Thanks again for your help with fixing my breakage!

PR #23812, opened 2023-08-24
Prevent pasting invalid characters when using wxTextValidator

### LINKED ISSUE #10281: wxTextCtrl::Paste does not respect a validator
*Issue migrated from trac ticket # 10281*

**component:** wxGTK | **priority:** normal | **keywords:** wxTextCtrl, Validators

#### 2008-12-17 10:55:42: ivan (Ivan) created the issue

___
I'm using a one line wxTextCtrl created in the following way:

 wxTextCtrl* lTextCtrl = new wxTextCtrl (this, wxID_ANY, "0", wxDefaultPosition, wxDefaultSize, wxTE_RIGHT, wxTextValidator (wxFILTER_NUMERIC));

If I'm typing a text from the keyboard, validation of the inserted text works well (only numbers are allowed). If I'm using a method  wxTextCtrl::Paste, validator is ignored, means any text could be inserted in the control.

Ivan

vadz:
It shouldn't be possible to put characters that can't be entered into
the associated control directly by pasting them, so intercept paste
events too and filter out any invalid characters.

Closes #10281.

### DIFF
--- a/include/wx/valtext.h
+++ b/include/wx/valtext.h
@@ -155,6 +155,8 @@ protected:
     wxArrayString        m_excludes;
 
 private:
+    void OnPaste(wxClipboardTextEvent& event);
+
     wxDECLARE_NO_ASSIGN_CLASS(wxTextValidator);
     wxDECLARE_DYNAMIC_CLASS(wxTextValidator);
     wxDECLARE_EVENT_TABLE();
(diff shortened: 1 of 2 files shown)


vadz:
This looks like such a long-standing but easy to fix (or am I missing something?) problem that I went ahead and fixed it.

It could also be backported to 3.2.

PR #23814, opened 2023-08-24
Revert all recent wxUSE_DPI_AWARE_MANIFEST-related changes

vadz:
This reverts 5d630caabd (Make it enough to predefine only
wxUSE_DPI_AWARE_MANIFEST, 2023-08-23) and all the commits which tried to
fix the breakage caused by it.

While the original change had merit, it seems to be too difficult to fix
all our build systems to avoid embedding manifest when defining this in
the code, like samples/sample.rc does, so revert this change for now.

Maybe it can be reintroduced in the future after switching to some other
build system.

### DIFF
--- a/build/cmake/functions.cmake
+++ b/build/cmake/functions.cmake
@@ -793,19 +793,16 @@ function(wx_add name group)
             set(wxUSE_DPI_AWARE_MANIFEST_VALUE 0)
             if (${wxUSE_DPI_AWARE_MANIFEST} MATCHES "system")
                 set(wxUSE_DPI_AWARE_MANIFEST_VALUE 1)
+                list(APPEND src_files "${wxSOURCE_DIR}/include/wx/msw/wx_dpi_aware.manifest")
             elseif(${wxUSE_DPI_AWARE_MANIFEST} MATCHES "per-monitor")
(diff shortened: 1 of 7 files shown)


vadz:
@MaartenBent Sorry, I give up, MSVC makefiles still don't work and Qt/CMake build on appveyor fails too for some reason, so I'm just going to undo all these changes for now. Sorry for easting your time.

vadz:
I'd like to merge this soon to allow the other CI builds to pass, please let me know if you have any objections.

MaartenBent:
The Qt/CMake build can be fixed my moving `/MANIFEST:NO` outside `if (DEFINED wxUSE_DPI_AWARE_MANIFEST_VALUE)` (which is only defined for `WXMSW`). I didn't look at the other error.

I'm fine with merging this PR. 
Maybe you can create a new PR with the desired `wxUSE_DPI_AWARE_MANIFEST` changes, and we can fix the remaining issues there.

vadz:
> I'm fine with merging this PR. Maybe you can create a new PR with the desired `wxUSE_DPI_AWARE_MANIFEST` changes, and we can fix the remaining issues there.

Yes, I should have done this from the beginning but I had totally failed to realize that this change was so disruptive.

PR #23815, opened 2023-08-25
Changes to fix #23811

### LINKED ISSUE #23811: wxAuiManager event interception prevents propagation of wxWindowDestroyEvent
### Description
The way wxAuiManager intercepts and handles wxWindowDestroyEvent events for it's managed window prevents to propagation of these events to other objects like the global App object.
In wxRuby this causes serious problems as window lifecycle management depends on these events reliably propagating.

Changing the code in `wxAuiManager::OnDestroy` to:

```c++
void wxAuiManager::OnDestroy(wxWindowDestroyEvent& event)
{
    if ( event.GetEventObject() == m_frame )
    {
        wxWindow* const frame = m_frame;

        UnInit();

        // Just calling Skip() would be insufficient in this case, as by
        // removing the event handler from the event handlers chain in UnInit()
        // we'd still prevent the frame from getting this event, so we need to
        // forward it to it manually. Note that this must be done after calling
        // UnInit() to prevent infinite recursion.
        if ( frame )
            frame->ProcessWindowEventLocally(event);

        event.Skip(); // ADDITION
    }
    else
    {
        event.Skip();
    }
}
```

solves the problem for me but I am not sure this is the best solution.

Probably something like:

```c++
        // Just calling Skip() would be insufficient in this case, as by
        // removing the event handler from the event handlers chain in UnInit()
        // we'd still prevent the frame from getting this event, so we need to
        // forward it to it manually. Note that this must be done after calling
        // UnInit() to prevent infinite recursion.
        if ( frame )
        {
            if (frame->ProcessWindowEventLocally(event))
              return;     // actually processed
        }

        event.Skip(); // Not actually processed so skip
```

might be better.

Is there a specific reason why wxAuiManager blocks wxWindowDestroyEvent (and possibly other) event propagation or would this constitute an actual bug?

- OS  and its version: SuSE Linux 15.4

+ GTK version:  3.3
+ Current theme:

mcorino:


mcorino:
It seems the AppVeyor build vms are somewhat unstable. The failures have nothing whatsoever to do with the changes made for this PR.

PR #23817, opened 2023-08-25
Add support for high DPI animations

vadz:
Implement possibility to specify multiple animations, similarly to the
possibility to specify multiple bitmaps, for different resolutions and use the
appropriate one in wxGenericAnimationCtrl (but not the GTK version yet).

Also clean up wxGenericAnimationCtrl code and, notably, stop using wxClientDC
in it.


MaartenBent:
Can you add the new file to CMake too?
https://github.com/wxWidgets/wxWidgets/blob/329d35b922aa893c292d14061fb099aacb4af20b/build/cmake/samples/CMakeLists.txt#L11

vadz:
Done now, thanks for the reminder!

PR #23818, opened 2023-08-26
Try to fix #23680

### LINKED ISSUE #23680: Key events with wrong key code (regression in wxGTK)
### Description

When pressing, for example, the keys `Z` and `Y` on a German keyboard, the key events `EVT_CHAR_HOOK`, `EVT_KEY_DOWN` and `EVT_KEY_UP` are generated with a wrong key code.

Note that the German keyboard layout has the `Z` and `Y` keys swapped compared to the US keyboard layout.
So it seems that these key events are generated according to the US keyboard layout.

With wxWidgets-3.2.2.1 it works correct.
So it is a regression that can be reproduced with both the current master and the current 3.2 branch.

The issue can be seen in the keyboard sample.

With wxWidgets 3.2.2.1 it works correct.

wxKeyEvents-3 2 2 1

With current master and 3.2 branch:

wxKeyEvents-3 2-branch

- wxWidgets version  you use: current master and 3.2 branch
- OS  and its version: openSUSE Leap 15.1

+ GTK version:  2.24.32

unxed:


vadz:
Thanks, this does restore the old behaviour and I don't see any regressions in my testing (I've checked with French and Russian layouts under wxGTK and wxMSW).

PR #23825, opened 2023-08-26
Fix displaying file size in sockets/client sample

PBfordev:
Not sure if it is worth changing: The new code is not the most readable and downloading with wxURL is basically deprecated anyway.

It's just something I ran into when testing input streams behaviours.


vadz:
Thanks, looks like an improvement to me, so I'll push it soon.

PR #23829, opened 2023-08-27
Make GetTextObject() in GTK text controls protected instead of private

Blake-Madden:
This allows users to derive from `wxTextCtrl` and be able to access the underlying `GtkTextBuffer`. This is useful in my case as this allows me to write Pango markup directly to the control efficiently.  I was trying to do this earlier:

```
GtkWidget* text_view = gtk_bin_get_child(GTK_BIN(GTK_SCROLLED_WINDOW(GetHandle())));
GtkTextBuffer* buffer = gtk_text_view_get_buffer(GTK_TEXT_VIEW(text_view) );
do stuff with buffer...
```
and I could do stuff with `buffer`. However, usually my edits get lost because `DoFreeze()` and `DoThaw()` swap the real buffer with an empty temp one. In my situation then, the buffer that `gtk_text_view_get_buffer()` gives me is usually the temp, empty one that eventually gets de-attached.

By exposing `GetTextObject()` as a protected member and using that, all works great as I am now pointing to the real buffer at all times.

I'm leaving this out of the documentation since you should only be using this if you know what you're doing.

### DIFF
--- a/include/wx/gtk/textctrl.h
+++ b/include/wx/gtk/textctrl.h
@@ -186,7 +186,7 @@ protected:
     void GTKSetWrapMode();
     void GTKSetJustification();
 
-    // get the underlying text buffer for multi-line controls, or NULL otherwise
+    // get the underlying text buffer for multi-line controls, or null otherwise
     GtkTextBuffer *GTKGetTextBuffer() { return IsMultiLine() ? m_buffer : nullptr; }
 
 private:
--- a/include/wx/gtk/textctrl.h
+++ b/include/wx/gtk/textctrl.h
@@ -186,6 +186,15 @@ protected:
     void GTKSetWrapMode();
     void GTKSetJustification();
 
+    // returns either m_text or m_buffer depending on whether the control is
+    // single- or multi-line; convenient for the GTK+ functions which work with
+    // both
+    void *GetTextObject() const override
+    {
+        return IsMultiLine() ? static_cast<void *>(m_buffer)
+                             : static_cast<void *>(m_text);
+    }
+
 private:
     void Init();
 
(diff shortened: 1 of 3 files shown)


vadz:
We can expose this, but the function should be called `GTKSomething()` to clearly show that it's GTK-specific and should still be documented, see e.g. the existing `GTKSuppressDiagnostics()`.

I also wonder if it really makes sense to have a single function returning `void*`. Presumably the application would use either a single- or multi-line text control but never one which can be both, so I'd suggest adding `GTKGetTextBuffer()` returning `GtkTextBuffer` or `nullptr` and, if we also need, `GTKGetEditable()` with similar semantics.

Blake-Madden:
OK, I reverted the changes and added `GTKGetTextBuffer()` like you suggested instead.  Much nicer option, less casting on my end :)

Blake-Madden:
...and added `GTKGetEditable()`, which is a wrapper around `GetEditable()`.

I don't think a `GtkGetEntry()` would be useful; there isn't much interesting that you can do with a single-line control. But of course, if you think I should add it I will.

include/wx/gtk/textctrl.h
```diff
@@ -186,6 +186,11 @@ class WXDLLIMPEXP_CORE wxTextCtrl: public wxTextCtrlBase
     void GTKSetWrapMode();
     void GTKSetJustification();
 
+    // get the underlying text buffer for multi-line controls, or null otherwise
+    GtkTextBuffer *GTKGetTextBuffer() { return IsMultiLine() ? m_buffer : nullptr; }
```
vadz, line 190: I think you need to forward-declare it in the beginning of the file, just as it's already done for `GtkTextMark` (which is, ironically, not needed here...).

vadz (commented):
Could please add the new functions to `interface/wx/gtk/textctrl.h` too please?

vadz:
> > Could please add the new functions to `interface/wx/gtk/textctrl.h` too please?
> 
> I didn't see that file,

Sorry, it was a typo, I did mean `interface/wx/gtk/textctrl.h`.

> but I do see where macOS specific functions are documented in `interface\wx\textctrl.h`. I don't see GTK functions documented at all in there, so I'll add those...

Thanks!

I don't think we want to make `UseGTKStyleBase()` part of public API however, so I'll remove this one before merging.

PR #23830, opened 2023-08-27
Update webview.cpp

beached:
Fixed crash when handler was called.


vadz:
Oops, thanks for fixing this! I'll merge it soon, although it actually looks like we could just remove the sizer entirely -- the only dialog child would be resized to cover it entirely anyhow. But it's definitely better not to crash if we do use it...

PR #23831, opened 2023-08-29
Fix wxTextCtrl::EnableProofCheck for WXMSW

mcorino:
fixes return value according to documentation

### DIFF
--- a/src/gtk/textctrl.cpp
+++ b/src/gtk/textctrl.cpp
@@ -1054,7 +1054,7 @@ bool wxTextCtrl::EnableProofCheck(const wxTextProofOptions& options)
         gspell_entry_set_inline_spell_checking(spell, options.IsSpellCheckEnabled());
     }
 
-    return GetProofCheckOptions().IsSpellCheckEnabled();
+    return GetProofCheckOptions().IsSpellCheckEnabled() == options.IsSpellCheckEnabled();
 }
 
 wxTextProofOptions wxTextCtrl::GetProofCheckOptions() const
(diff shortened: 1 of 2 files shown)


vadz:
It's true that the documentation indicates that `EnableProofCheck()` should return `true` even when disabling the check, but currently both wxMSW and wxGTK return `false` in this case (although wxOSX does return `true` unconditionally), so this PR doesn't fully solve the problem.

We need to either change both wxMSW and wxGTK or change the documentation and wxOSX. I am not sure which one is better, i.e. what return value is more useful. It seems to me that the return value when disabling checking doesn't matter all that much (what are you going to do in case of failure, anyhow?), so I'm fine with either but changing the documentation to say that the function returns `true` if checking could be enabled seems slightly more logical to me.

What do you think?

mcorino:
I honestly do not really care either way. I just assumed the author would have thought this through and the implementation should match the documentation.
You say that wxGTK returns `false` on disabling spell checking but my wxRuby regression test cares to differ with that.
The reason I submitted this PR is because I found through this regression test that wxMSW behavior differed from wxOSX **and** wxGTK. I did not check the wxGTK code extensively but at least in my wxRuby regression test wxGTK adheres to the currently documented behavior.
That being so I would prefer to keep the currently documented behavior.

vadz:
Hmm, this might indicate that there is a bug in wxGTK because the code currently does

https://github.com/wxWidgets/wxWidgets/blob/00366cbaae3e881413af5453608ea3ed031e3cb0/src/gtk/textctrl.cpp#L1057

so if it returns `true` after disabling the check, something is wrong...

mcorino:
Ok, I can now confirm `wxTextCtrl::GetProofCheckOptions`  is faulty on wxGTK. After (successfully) disabling spell checking this method will keep reporting that spell checking is enabled.
And I do not think the error is in GTK but rather in the implementation of wxGTK here:
 https://github.com/wxWidgets/wxWidgets/blob/00366cbaae3e881413af5453608ea3ed031e3cb0/src/gtk/textctrl.cpp#L1068
This does not seem to check the setting of the `inline-spell-checking` property (of which I have currently no idea how it should be checked).

mcorino:
Ahh, it seems there are getters for this properties as well (figures).
So a call `gspell_text_view_get_inline_spell_checking` would probably fix behavior.
Do you want me take a stab at this?
And how do you want to end up? According to current documentation?

mcorino:
Changes committed to make implementation behave according to docs on wxGTK as well.
Can confirm that with wxRuby everything now works as expected.

vadz:
Great, thanks! Will merge in a moment.

PR #23834, opened 2023-08-30
Don't clobber std::string_view equality with char *

joanbm:
Small fix for an annoyance I found when trying to build [SLADE](https://github.com/sirjuddington/SLADE) with wxWidgets 3.3:

Trying to compare a `std::string_view` to a `char *` now fails due to ambiguity with the new `wxString(std::string_view)` constructor.

I suggest making it explicit since I guess that is a pretty common use case.

This sample only builds if the constructor is explicit:

 ```c++
#include <wx/string.h>
#include <string_view>

int main() {
  std::string_view view = "abc";
  const char *str = "abc";
  return view == str;
}
```

This is also consistent with `std::string` from the STL, whose constructor from `std::string_view` is explicit (c.f. https://en.cppreference.com/w/cpp/string/basic_string/basic_string, 10th overload)

### DIFF
--- a/include/wx/string.h
+++ b/include/wx/string.h
@@ -1263,7 +1263,7 @@ public:
   #endif
 
 #ifdef wxHAS_STD_STRING_VIEW
-    wxString(std::wstring_view view)
+    explicit wxString(std::wstring_view view)
         { assign(view.data(), view.length()); }
 #endif  // wxHAS_STD_STRING_VIEW
 
@@ -1272,7 +1272,7 @@ public:
       { assign(str.c_str(), str.length()); }
 
     #ifdef wxHAS_STD_STRING_VIEW
-        wxString(std::string_view view)
+        explicit wxString(std::string_view view)
             { assign(view.data(), view.length()); }
     #endif  // wxHAS_STD_STRING_VIEW
 #endif // wxNO_IMPLICIT_WXSTRING_ENCODING
@@ -1881,6 +1881,29 @@ public:
     // from wxScopedCharBuffer
   wxString& operator=(const wxScopedCharBuffer& s)
     { return assign(s); }
+#endif // wxNO_IMPLICIT_WXSTRING_ENCODING
+
+#if wxUSE_UNICODE_WCHAR
+  wxString& operator=(const std::wstring& str) { m_impl = str; return *this; }
+  wxString& operator=(std::wstring&& str) noexcept { m_impl = std::move(str); return *this; }
+#else // wxUSE_UNICODE_UTF8
+  wxString& operator=(const std::wstring& str)
+    { return assign(str.c_str(), str.length()); }
+#endif
+
+#ifdef wxHAS_STD_STRING_VIEW
+  wxString& operator=(std::wstring_view view)
+    { return assign(view.data(), view.length()); }
+#endif  // wxHAS_STD_STRING_VIEW
+
+#ifndef wxNO_IMPLICIT_WXSTRING_ENCODING
+  wxString& operator=(const std::string& str)
+    { return assign(str.c_str(), str.length()); }
+
+    #ifdef wxHAS_STD_STRING_VIEW
+      wxString& operator=(std::string_view view)
+        { return assign(view.data(), view.length()); }
+    #endif  // wxHAS_STD_STRING_VIEW
 #endif // wxNO_IMPLICIT_WXSTRING_ENCODING
 
   // string concatenation
--- a/tests/strings/stdstrings.cpp
+++ b/tests/strings/stdstrings.cpp
@@ -690,5 +690,14 @@ TEST_CASE("StdString::View", "[stdstring]")
     std::string_view strViewInvalidUTF(strInvalidUTF);
 
     CHECK( "" == wxString::FromUTF8(strViewInvalidUTF) );
+
+    /* Ensure we don't clobber comparisons on base types */
+    std::string_view view = "abc";
+    const char *str = "abc";
+    CHECK( view == str );
+
+    std::wstring_view wview = L"abc";
+    const wchar_t *wstr = L"abc";
+    CHECK( wview == wstr );
 }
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by imciner2** [comment]
Welp, that is annoying. Making these explicit seems to cause the assignment from `std::string_view` to `wxString` not to work, though, so we'll need to add those operators manually into the class as well (that is why this change is causing the tests to fail to compile on my machine).

I'd also suggest adding the following to the stdstring tests though:

```
diff --git a/tests/strings/stdstrings.cpp b/tests/strings/stdstrings.cpp
index f3ff37f2ee..fec2536181 100644
--- a/tests/strings/stdstrings.cpp
+++ b/tests/strings/stdstrings.cpp
@@ -690,5 +690,15 @@ TEST_CASE("StdString::View", "[stdstring]")
     std::string_view strViewInvalidUTF(strInvalidUTF);
 
     CHECK( "" == wxString::FromUTF8(strViewInvalidUTF) );
+
+    /* Ensure we don't clobber comparisons on base types */
+    std::string_view view = "abc";
+    const char *str = "abc";
+    CHECK( view == str )
+
+    std::wstring_view wview = L"abc";
+    const wchar *wstr = L"abc";
+    CHECK( wview == wstr )
+
 }
 #endif // wxHAS_STD_STRING_VIEW
```

**Comment by joanbm** [comment]
PR updated with the suggestions.

Adding `operator=(std::string_view)` fixes that build problem, but however, introduces another. The following then fails to build:

```c++
        std::string str = "abc";
        wxString s;
        s = str; // Ambiguous between s = wxString(str)
                                  and s = std::string_view(str)
```

This can be easily fixed by also adding `operator=(const std::string&)`, which is what I did.

However, in general, since we have both `operator=(const wxString&)` and  `operator=(std::string_view)`, this means that any type that has implicit conversions that allow it to use either overload will cause an ambiguity problem (unless it's handled by a more specific overload beforehand). As seen before `std::string` was one of them, but there's at least one more (plus all its wchar_t, etc. variants):

```c++
    wxString s;
    s = {"abc", 2}; // Ambiguous between s = wxString("abc", 2)
                                     and s = std::string_view("abc", 2)
```

This can also be fixed (see [LWG 2946](https://cplusplus.github.io/LWG/issue2946)), but since the case is rather obscure, I'd rather avoid it to keep things simple (and it can always be done later), however, I'm open to your thoughts.

vadz:
Thanks Joan! I'm ready to merge this, we just need to remove the mention of `implicit wxString::wxString(std::string_view) ctor` from the docs, but I can do it when merging.

@imciner2 Please let me know if you have any comments about/objections to this.

imciner2:
Yes, this looks good to me. I'd say that we can consider the example of assigning like `s = {"abc", 2}` to be undesirable, and users should instead be explicit about the creation of the wxString for it instead, so I don't see that as a problem we should try to fix.

@vadz maybe we should add a statement in the "changes that may cause build errors" section of the changes file that says this type of assignment is problematic and that the users should be explicit instead?

vadz:
> @vadz maybe we should add a statement in the "changes that may cause build errors" section of the changes file that says this type of assignment is problematic and that the users should be explicit instead?

As support for `string_view` is fairly new and hasn't been included in any released version yet, I don't think we need to worry about this.

imciner2:
> As support for string_view is fairly new and hasn't been included in any released version yet, I don't think we need to worry about this.

But my understanding was that the assignment `s = {"abc", 2}` would have worked in code building against old wxWidgets versions, but with the new overloads/functions it is problematic. So it would seem to me to be a change in behavior caused by adding this feature.

vadz:
> But my understanding was that the assignment `s = {"abc", 2}` would have worked in code building against old wxWidgets versions

Oh, I didn't think about this. Does this work in 3.2? If so, it indeed needs to be documented...

imciner2:
> Oh, I didn't think about this. Does this work in 3.2? If so, it indeed needs to be documented...

The minimal sample compiles and works in the 3.2 branch with the following patch, so it was previously possible.

```
index 3d71bd276c..4dced469dc 100644
@@ -173,7 +173,9 @@ MyFrame::MyFrame(const wxString& title)
 #if wxUSE_STATUSBAR
     // create a status bar just for fun (by default with 1 pane only)
     CreateStatusBar(2);
-    SetStatusText("Welcome to wxWidgets!");
+    wxString value;
+    value = {"Hello", 5};
+    SetStatusText(value);
 #endif // wxUSE_STATUSBAR
 }
```

Using that patch on the minimal sample compiled against this PR gives the following error though:
```
../../../samples/minimal/minimal.cpp: In constructor ‘MyFrame::MyFrame(const wxString&)’:
../../../samples/minimal/minimal.cpp:177:24: error: ambiguous overload for ‘operator=’ (operand types are ‘wxString’ and ‘<brace-enclosed initializer list>’)
  177 |     value = {"Hello", 5};
      |                        ^
In file included from ../../../include/wx/object.h:23,
                 from ../../../include/wx/wx.h:15,
                 from ../../../samples/minimal/minimal.cpp:26:
../../../include/wx/string.h:1797:13: note: candidate: ‘wxString& wxString::operator=(const wxString&)’
 1797 |   wxString& operator=(const wxString& stringSrc)
      |             ^~~~~~~~
../../../include/wx/string.h:1810:13: note: candidate: ‘wxString& wxString::operator=(wxString&&)’
 1810 |   wxString& operator=(wxString&& stringSrc) noexcept
      |             ^~~~~~~~
../../../include/wx/string.h:1900:13: note: candidate: ‘wxString& wxString::operator=(const std::string&)’
 1900 |   wxString& operator=(const std::string& str)
      |             ^~~~~~~~
../../../include/wx/string.h:1904:17: note: candidate: ‘wxString& wxString::operator=(std::string_view)’
 1904 |       wxString& operator=(std::string_view view)
      |                 ^~~~~~~~
make: *** [Makefile:190: minimal_minimal.o] Error 1
```

PR #23835, opened 2023-08-30
Handle map/unmap events on Wayland's wxGLCanvasEGL

### LINKED ISSUE #22580: wxNotebook can't handle wxGLCanvas correctly on Wayland
**Bug description**
When a wxGLCanvas is inserted as one of the pages of a wxNotebook, the OpenGL window is visible on other pages as well.

The wxGLCanvas content with properly initialized OpenGL context and some rendered content is expected to be visible only when the appropriate page -- into which the canvas was inserted -- is selected.
The observation is that the OpenGL content is visible on other pages too. This is only happening in a Wayland session with EGL support enabled. Setting `GDK_BACKEND=x11` in such environment will hide the issue. 

**Screenshots**
Expected (also when `GDK_BACKEND=x11`): Screenshot from 2022-06-28 17-33-12

Observed: 
Screenshot from 2022-06-28 17-32-25

```c++
#include <iostream>
#include <utility>
#include <memory>
#include <vector>

// For compilers that support precompilation, includes "wx/wx.h".
#include <wx/wxprec.h>
#ifndef WX_PRECOMP
#include <wx/wx.h>
#endif

#include <wx/tglbtn.h>
#include <wx/combobox.h>
#include <wx/msgdlg.h>
#include <wx/glcanvas.h>
#include <wx/notebook.h>

class Renderer {
protected:
    wxGLCanvas *m_canvas;
    std::unique_ptr<wxGLContext> m_context;
public:
    
    Renderer(wxGLCanvas *c): m_canvas{c} {
        m_context = std::make_unique<wxGLContext>(m_canvas);
    }
    
    wxGLContext * context() { return m_context.get(); }
    const wxGLContext * context() const { return m_context.get(); }

    // OpenGL example from https://cs.lmu.edu/~ray/notes/openglexamples/
    // section: Tetrahedron
    
    void set_active()
    {
        m_canvas->SetCurrent(*m_context);
        
        glClearColor(0.1, 0.39, 0.88, 1.0);
        glColor3f(1.0, 1.0, 1.0);

        glEnable(GL_CULL_FACE);
        glCullFace(GL_BACK);

        glMatrixMode(GL_PROJECTION);
        glLoadIdentity();
        glFrustum(-2, 2, -1.5, 1.5, 1, 40);

        glMatrixMode(GL_MODELVIEW);
        glLoadIdentity();
        glTranslatef(0, 0, -3);
        glRotatef(50, 1, 0, 0);
        glRotatef(70, 0, 1, 0);
    }

    void draw_scene(long w, long h)
    {
        glViewport(0, 0, GLsizei(w), GLsizei(h));
        glClear(GL_COLOR_BUFFER_BIT);

        // Draw a white grid "floor" for the tetrahedron to sit on.
        glColor3f(1.0, 1.0, 1.0);
        glBegin(GL_LINES);
        for (GLfloat i = -2.5; i <= 2.5; i += 0.25) {
            glVertex3f(i, 0, 2.5); glVertex3f(i, 0, -2.5);
            glVertex3f(2.5, 0, i); glVertex3f(-2.5, 0, i);
        }
        glEnd();

        // Draw the tetrahedron.  It is a four sided figure, so when defining it
        // with a triangle strip we have to repeat the last two vertices.
        glBegin(GL_TRIANGLE_STRIP);
        glColor3f(1, 1, 1); glVertex3f(0, 2, 0);
        glColor3f(1, 0, 0); glVertex3f(-1, 0, 1);
        glColor3f(0, 1, 0); glVertex3f(1, 0, 1);
        glColor3f(0, 0, 1); glVertex3f(0, 0, -1.4);
        glColor3f(1, 1, 1); glVertex3f(0, 2, 0);
        glColor3f(1, 0, 0); glVertex3f(-1, 0, 1);
        glEnd();

        glFlush();
    }

    void swap_buffers() { m_canvas->SwapBuffers(); }
};

// The top level frame of the application.
class MyFrame: public wxFrame
{
    wxGLCanvas     *m_canvas;
    std::unique_ptr<Renderer> m_renderer;
    
public:
    MyFrame(const wxString &       title,
            const wxPoint &        pos,
            const wxSize &         size);

    wxGLCanvas * canvas() { return m_canvas; }
    const wxGLCanvas * canvas() const { return m_canvas; }
};

class App : public wxApp {
    MyFrame *m_frame = nullptr;
    wxString m_fname;
public:
    bool OnInit() override {

        m_frame = new MyFrame("Wayland wxNotebook issue", wxDefaultPosition, wxSize(1024, 768));
        m_frame->Show( true );
        
        return true;
    }

};

wxIMPLEMENT_APP(App);

MyFrame::MyFrame(const wxString &title, const wxPoint &pos, const wxSize &size):
    wxFrame(nullptr, wxID_ANY, title, pos, size)
{
    wxMenu *menuFile = new wxMenu;
    menuFile->Append(wxID_OPEN);
    menuFile->Append(wxID_EXIT);
    wxMenuBar *menuBar = new wxMenuBar;
    menuBar->Append( menuFile, "&File" );
    SetMenuBar( menuBar );

    auto notebookpanel = new wxPanel(this);
    auto notebook      = new wxNotebook(notebookpanel, wxID_ANY);
    auto maintab       = new wxPanel(notebook);

    m_canvas = new wxGLCanvas(maintab,
                              wxID_ANY,
                              nullptr,
                              wxDefaultPosition,
                              wxDefaultSize,
                              wxWANTS_CHARS | wxFULL_REPAINT_ON_RESIZE);

    m_renderer = std::make_unique<Renderer>(m_canvas);

    wxPanel *control_panel = new wxPanel(maintab);

    auto controlsizer = new wxBoxSizer(wxHORIZONTAL);
    auto console_sizer = new wxBoxSizer(wxVERTICAL);

    std::vector<wxString> combolist = {"One", "Two", "Three"};
    auto combobox = new wxComboBox(control_panel, wxID_ANY, combolist[0],
                                   wxDefaultPosition, wxDefaultSize,
                                   int(combolist.size()), combolist.data());

    auto sz = new wxBoxSizer(wxHORIZONTAL);
    sz->Add(new wxStaticText(control_panel, wxID_ANY, "Choose number"), 0,
            wxALL | wxALIGN_CENTER, 5);
    sz->Add(combobox, 1, wxALL | wxEXPAND, 5);
    console_sizer->Add(sz, 0, wxEXPAND);

    auto btn1 = new wxToggleButton(control_panel, wxID_ANY, "Button1");
    console_sizer->Add(btn1, 0, wxALL | wxEXPAND, 5);

    auto btn2 = new wxToggleButton(control_panel, wxID_ANY, "Button2");
    btn2->SetValue(true);
    console_sizer->Add(btn2, 0, wxALL | wxEXPAND, 5);

    controlsizer->Add(console_sizer, 1, wxEXPAND);
    
    control_panel->SetSizer(controlsizer);
    
    auto maintab_sizer = new wxBoxSizer(wxHORIZONTAL);
    maintab_sizer->Add(m_canvas, 1, wxEXPAND);
    maintab_sizer->Add(control_panel, 0);
    maintab->SetSizer(maintab_sizer);

    notebook->AddPage(maintab, "Main");

    wxTextCtrl* textCtrl1 = new wxTextCtrl(notebook, wxID_ANY, L"Tab 2 Contents");
    notebook->AddPage(textCtrl1, "Dummy");

    auto notebooksizer = new wxBoxSizer(wxHORIZONTAL);
    notebooksizer->Add(notebook, 1, wxEXPAND);
    notebookpanel->SetSizer(notebooksizer);

    auto topsizer = new wxBoxSizer(wxHORIZONTAL);
    topsizer->Add(notebookpanel, 1, wxEXPAND);
    SetSizer(topsizer);
    SetMinSize(size);

    Bind(wxEVT_MENU, [this](wxCommandEvent &) {
        wxFileDialog dlg(this, "Select file",  wxEmptyString,
                         wxEmptyString, "*.*", wxFD_OPEN|wxFD_FILE_MUST_EXIST);
        dlg.ShowModal();
    }, wxID_OPEN);

    Bind(wxEVT_MENU, [this](wxCommandEvent &) {
        Close();
    }, wxID_EXIT);

    Bind(wxEVT_SHOW, [this](wxShowEvent &) {
        m_renderer->set_active();

        m_canvas->Bind(wxEVT_PAINT, [this](wxPaintEvent &){
            wxPaintDC dc(m_canvas);

            const wxSize sz = m_canvas->GetClientSize();
            m_renderer->draw_scene(sz.x, sz.y);
            m_renderer->swap_buffers();
        });
    });
}
```

**To Reproduce**
Switch between the "Main" and "Dummy" tabs.

**Related**
See a potentially related issue here https://github.com/prusa3d/PrusaSlicer/issues/8284

 - wxWidgets version: 3.1.7 or 3.1.6 (where wxGLCanvasEGL is in use)
 - wxWidgets port: wxGTK
 - OS and its version: Manjaro Linux (or any other with a Wayland session)
 - GTK version: 3.24.34
 - GDK backend: Wayland

joanbm:
Fixes hiding a wxGLCanvas on Wayland, either directly (`->Show(false)`) or indirectly (e.g. when it is contained in a wxNotebook).

On Wayland, unlike on X11, the window and surface that we create with `wl_egl_window_create` and `eglCreatePlatformWindowSurface` is detached from the GTK widget associated to the canvas, thus it is not automatically mapped or unmapped when the associated GTK widget is. Rather, we need to manually synchronize it with the widget's state.

As Wayland/EGL does not provide any way to temporally hide the window, we need to fully destroy it on unmap, and re-create it on map.

Fixes #22580

---

Annex: Test case that exercises some tricky cases: The necessity of the `map-event` signal instead of `map` when hiding and re-showing toplevel windows containing canvases, and correct handling of re-parenting of canvases between windows.

```c++
#include <wx/app.h>
#include <wx/glcanvas.h>
#include <wx/dcclient.h>
#include <wx/frame.h>
#include <wx/button.h>
#include <wx/sizer.h>

class MyWxApp : public wxApp
{
	wxGLContext *context;

public:
	bool OnInit() override {

		auto subWindow = new wxFrame(nullptr, -1, wxT("Sub1"));
		subWindow->Bind(wxEVT_CLOSE_WINDOW, [subWindow](wxCloseEvent &evt) {
			subWindow->Show(false);
		});

		auto subWindow2 = new wxFrame(nullptr, -1, wxT("Sub2"));
		subWindow2->Bind(wxEVT_CLOSE_WINDOW, [subWindow2](wxCloseEvent &evt) {
			subWindow2->Show(false);
		});

		auto canvas = new wxGLCanvas(subWindow);
		context = new wxGLContext(canvas);
		canvas->Bind(wxEVT_PAINT, [this, canvas](wxPaintEvent &evt) {
			context->SetCurrent(*canvas);
			wxPaintDC dc(canvas);

			glClearColor(1.0f, 0.0f, 0.0f, 1.0f);
			glClear(GL_COLOR_BUFFER_BIT);

			canvas->SwapBuffers();
		});

		auto buttonToggleWindows = new wxButton(window, wxID_ANY, wxT("Toggle windows"));
		buttonToggleWindows->Bind(wxEVT_BUTTON, [subWindow, subWindow2](wxCommandEvent &evt) {
			subWindow->Show(!subWindow->IsShown());
			subWindow2->Show(!subWindow2->IsShown());
		});

		auto buttonToggleCanvas = new wxButton(window, wxID_ANY, wxT("Toggle canvas"));
		buttonToggleCanvas->Bind(wxEVT_BUTTON, [canvas](wxCommandEvent &evt) {
			canvas->Show(!canvas->IsShown());
		});

		auto buttonReparent = new wxButton(window, wxID_ANY, wxT("Reparent"));
		buttonReparent->Bind(wxEVT_BUTTON, [subWindow, subWindow2, canvas](wxCommandEvent &evt) {
			canvas->Reparent(canvas->GetParent() == subWindow ? subWindow2 : subWindow);
		});

		auto sizer = new wxBoxSizer(wxHORIZONTAL);
		sizer->Add(buttonToggleWindows, 1, wxEXPAND);
		sizer->Add(buttonToggleCanvas, 1, wxEXPAND);
		sizer->Add(buttonReparent, 1, wxEXPAND);
		window->SetSizer(sizer);
		window->Layout();

		window->Bind(wxEVT_CLOSE_WINDOW, [window, subWindow, subWindow2](wxCloseEvent &evt) {
			subWindow->Destroy();
			subWindow2->Destroy();
			window->Destroy();
		});

		window->Show(true);
		SetTopWindow(window);
		return true;
	}

	int OnExit()
	{
		delete context;
		return 0;
	}
};

IMPLEMENT_APP(MyWxApp)
```

### DIFF
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -389,6 +389,14 @@ EGLDisplay wxGLCanvasEGL::GetDisplay()
 // Helper declared as friend in the header and so can access m_wlSubsurface.
 void wxEGLUpdatePosition(wxGLCanvasEGL* win)
 {
+    if ( !win->m_wlSubsurface )
+    {
+        // In some circumstances such as when reparenting a canvas between two hidden
+        // toplevel windows, GTK will call size-allocate before mapping the canvas
+        // Ignore the call, the position will be fixed when it is mapped
+        return;
+    }
+
     int x, y;
     gdk_window_get_origin(win->GTKGetDrawingWindow(), &x, &y);
     wl_subsurface_set_position(win->m_wlSubsurface, x, y);
@@ -439,6 +447,17 @@ static const struct wl_callback_listener wl_frame_listener = {
     wl_frame_callback_handler
 };
 
+static gboolean gtk_glcanvas_map_callback(GtkWidget *, GdkEventAny *, wxGLCanvasEGL *win)
+{
+    win->CreateWaylandSubsurface();
+    return FALSE;
+}
+
+static void gtk_glcanvas_unmap_callback(GtkWidget *, wxGLCanvasEGL *win)
+{
+    win->DestroyWaylandSubsurface();
+}
+
 static void gtk_glcanvas_size_callback(GtkWidget *widget,
                                        GtkAllocation *,
                                        wxGLCanvasEGL *win)
@@ -462,16 +481,16 @@ bool wxGLCanvasEGL::CreateSurface()
         return false;
     }
 
-    if ( m_surface != EGL_NO_SURFACE )
-    {
-        eglDestroySurface(m_surface, m_display);
-        m_surface = EGL_NO_SURFACE;
-    }
-
     GdkWindow *window = GTKGetDrawingWindow();
 #ifdef GDK_WINDOWING_X11
     if (wxGTKImpl::IsX11(window))
     {
+        if ( m_surface != EGL_NO_SURFACE )
+        {
+            eglDestroySurface(m_surface, m_display);
+            m_surface = EGL_NO_SURFACE;
+        }
+
         m_xwindow = GDK_WINDOW_XID(window);
         m_surface = eglCreatePlatformWindowSurface(m_display, *m_config,
                                                    &m_xwindow, nullptr);
@@ -480,10 +499,16 @@ bool wxGLCanvasEGL::CreateSurface()
 #ifdef GDK_WINDOWING_WAYLAND
     if (wxGTKImpl::IsWayland(window))
     {
+        if ( m_wlSurface )
+        {
+            // Already created (can happen when the canvas is un-realized then
+            // re-realized, for example, when the canvas is re-parented)
+            return true;
+        }
+
         int w = gdk_window_get_width(window);
         int h = gdk_window_get_height(window);
         struct wl_display *display = gdk_wayland_display_get_wl_display(gdk_window_get_display(window));
-        struct wl_surface *surface = gdk_wayland_window_get_wl_surface(window);
         struct wl_registry *registry = wl_display_get_registry(display);
         wl_registry_add_listener(registry, &wl_registry_listener, this);
         wl_display_roundtrip(display);
@@ -494,21 +519,25 @@ bool wxGLCanvasEGL::CreateSurface()
         }
         m_wlSurface = wl_compositor_create_surface(m_wlCompositor);
         m_wlRegion = wl_compositor_create_region(m_wlCompositor);
-        m_wlSubsurface = wl_subcompositor_get_subsurface(m_wlSubcompositor,
-                                                         m_wlSurface,
-                                                         surface);
         wl_surface_set_input_region(m_wlSurface, m_wlRegion);
-        wl_subsurface_set_desync(m_wlSubsurface);
-        wxEGLUpdatePosition(this);
         int scale = gdk_window_get_scale_factor(window);
         wl_surface_set_buffer_scale(m_wlSurface, scale);
         m_wlEGLWindow = wl_egl_window_create(m_wlSurface, w * scale,
                                              h * scale);
         m_surface = eglCreatePlatformWindowSurface(m_display, *m_config,
                                                    m_wlEGLWindow, nullptr);
-        m_wlFrameCallbackHandler = wl_surface_frame(surface);
-        wl_callback_add_listener(m_wlFrameCallbackHandler,
-                                 &wl_frame_listener, this);
+
+        // We need to use "map-event" instead of "map" to ensure that the
+        // widget's underlying Wayland surface has been created.
+        // Otherwise, gdk_wayland_window_get_wl_surface may return nullptr,
+        // for example when hiding then showing a window containing a canvas.
+        gtk_widget_add_events(m_widget, GDK_STRUCTURE_MASK);
+        g_signal_connect(m_widget, "map-event",
+                         G_CALLBACK(gtk_glcanvas_map_callback), this);
+        // However, note the use of "unmap" instead of the later "unmap-event"
+        // Not unmapping the canvas as soon as possible causes problems when reparenting
+        g_signal_connect(m_widget, "unmap",
+                         G_CALLBACK(gtk_glcanvas_unmap_callback), this);
         g_signal_connect(m_widget, "size-allocate",
                          G_CALLBACK(gtk_glcanvas_size_callback), this);
 
@@ -535,9 +564,39 @@ wxGLCanvasEGL::~wxGLCanvasEGL()
         eglDestroySurface(m_display, m_surface);
 #ifdef GDK_WINDOWING_WAYLAND
     g_clear_pointer(&m_wlEGLWindow, wl_egl_window_destroy);
-    g_clear_pointer(&m_wlSubsurface, wl_subsurface_destroy);
     g_clear_pointer(&m_wlSurface, wl_surface_destroy);
+#endif
+}
+
+void wxGLCanvasEGL::CreateWaylandSubsurface()
+{
+#ifdef GDK_WINDOWING_WAYLAND
+    GdkWindow *window = GTKGetDrawingWindow();
+    struct wl_surface *surface = gdk_wayland_window_get_wl_surface(window);
+
+    m_wlSubsurface = wl_subcompositor_get_subsurface(m_wlSubcompositor,
+                                                     m_wlSurface,
+                                                     surface);
+    wl_subsurface_set_desync(m_wlSubsurface);
+    wxEGLUpdatePosition(this);
+    m_wlFrameCallbackHandler = wl_surface_frame(surface);
+    wl_callback_add_listener(m_wlFrameCallbackHandler,
+                             &wl_frame_listener, this);
+
+    if ( m_surface == EGL_NO_SURFACE )
+    {
+        wxFAIL_MSG("Unable to create EGL surface");
+        return;
+    }
+#endif
+}
+
+void wxGLCanvasEGL::DestroyWaylandSubsurface()
+{
+#ifdef GDK_WINDOWING_WAYLAND
+    g_clear_pointer(&m_wlSubsurface, wl_subsurface_destroy);
     g_clear_pointer(&m_wlFrameCallbackHandler, wl_callback_destroy);
+    m_readyToDraw = false;
 #endif
 }
 
--- a/include/wx/unix/glegl.h
+++ b/include/wx/unix/glegl.h
@@ -69,6 +69,11 @@ public:
 
     virtual ~wxGLCanvasEGL();
 
+    // Wayland-specific callbacks
+    // --------------------------
+
+    void CreateWaylandSubsurface();
+    void DestroyWaylandSubsurface();
 
     // implement wxGLCanvasBase methods
     // --------------------------------


vadz:
Thanks a lot once again for fixing this!

I will merge this into master and backport to 3.2 soon if there are no objections.

cc @paulcor @swt2c

joanbm:
Grr, I was running some final tests and there's still one test case where it doesn't work (the canvas appears empty), the original `wxNotebook` one: https://github.com/wxWidgets/wxWidgets/issues/22580#issue-1287297419

The problem is that that test case binds the OpenGL context on `wxEVT_SHOW` instead of inside `wxEVT_PAINT` (which I think ought to work):

```c++
    Bind(wxEVT_SHOW, [this](wxShowEvent &) {
        m_renderer->set_active(); // this makes the OpenGL context current

        m_canvas->Bind(wxEVT_PAINT, [this](wxPaintEvent &){
            // ...
        });
    });
```

But the problem is that we wait for GTK's `map-event` signal to create the EGL surface, but `wxEVT_SHOW` happens before it (the order is: 1. `map`, 2. `wxEVT_SHOW`, 3. `map-event`), so at the time it tries to make the OpenGL context current, we haven't yet created the EGL surface. Maybe we should use `map` instead of `map-event`, but this breaks another test case due to the way GDK creates the Wayland surface.

I'll take a look to see what can be done.

vadz:
I'm not totally sure about the difference between `map` and `map-event`, but could we, perhaps, generate `wxEVT_SHOW` from the latter instead of the former?

vadz:
Looking at [the docs](https://docs.gtk.org/gtk3/signal.Widget.map.html), it seems like `map-event` would be more appropriate in some cases, but not all :-(

We could add `wxEVT_SHOWING` (or `wxEVT_WILL_SHOW`) and make existing `wxEVT_SHOW` synonym for `wxEVT_SHOWN` or vice versa but neither is going to be 100% compatible, of course.

joanbm:
Yeah, I suspect that changing `wxEVT_SHOW` in GTK+ to happen after `map-event` is going to lead to all sorts of breakage, since the `map-event` is not emitted synchronous, but rather put in the event loop, so there's a chance for other events to sneak in.

And it's not a hard requirement that we create the EGL window/surface on `map-event` instead of `map`. It's that if you try to do it, you run into a *very* inconvenient GTK/GDK behaviour that makes it hell to re-create the EGL window/surface when the toplevel window containing the canvas is hidden, then re-shown.

For now, I've updated the PR to use `map` again, and a second commit entirely to deal with said GTK/GDK behaviour. However, I think the right solution is to try to get this GTK4 commit: [wayland: Don't destroy the wl_surface on hide()](https://gitlab.gnome.org/GNOME/gtk/-/commit/5d3cec5441ca1fa28e5e77c0572e629eded419c9) backported to GTK3. I'll try to do it, if this can be done, the second commit can be dropped entirely.

With this, all previous tests work, *and* the [wxNotebook sample](https://github.com/wxWidgets/wxWidgets/issues/22580#issue-1287297419) which makes the OpenGL context current on `wxEVT_SHOW` now works on launch.

What *still* doesn't work in that sample, is that if you change to the other tab, and then back to the canvas tab, it no longer render correctly. This is because when changing between tabs, we destroy and re-create the EGL window/surface as needed ([as there is apparently no way to just hide it?](https://stackoverflow.com/q/73941692)). So the context that was bound on `wxEVT_SHOW` becomes stale then. Making the context current on `wxEVT_PAINT` fixes it, as expected.

PS: I also fixed some edge cases I found related to reparenting a wxCanvas to another window.

vadz:
Thanks a lot for all your work on this, I can imagine that dealing with all these "oddities" must have been pretty frustrating!

I'm not sure what do you mean by backporting this commit to GTK3, do you really mean doing it in GTK itself? To the best of my knowledge, GTK3 doesn't get any updates any longer...

Anyhow, I'm more than ready to fix this if it fixes all the issues (I don't doubt that it does, but I just haven't had time to check them myself yet) even in the current form. The code is more complicated than ideal, but actually making things work for the users is definitely more important than code elegance in this case.

Thanks again!

joanbm:
> Thanks a lot for all your work on this, I can imagine that dealing with all these "oddities" must have been pretty frustrating!

Thanks! Indeed.

> I'm not sure what do you mean by backporting this commit to GTK3, do you really mean doing it in GTK itself? To the best of my knowledge, GTK3 doesn't get any updates any longer...

Yes, in GTK itself. I don't think it gets any new features but it [definitely gets some bug fixes](https://gitlab.gnome.org/GNOME/gtk/-/commits/gtk-3-24/). So it *may* be possible to get it in...

---

Finally, I may also have figured out the answer to your StackOverflow question: To hide it, you can just destroy the subsurface (i.e. `m_wlSubsurface` in our code) that is associated to its surface when the GTK widget unmaps, and the EGL surface is immediately gone from the screen. You don't need to destroy the EGL window or surface.

I'd swear that like you, I tried to do this before, and it didn't work. But I was probably doing something wrong. (And unfortunately with those OpenGL things, you're lucky if you get a crash instead of some weird deadlock, graphical glitch...)

I have yet to fully test and review this (I'll update this PR), but with it, the wxNotebook example should fully work on Wayland, including changing tabs, without any code changes.

joanbm:
PR updated, I believe that it's the right approach now.

After understanding that you *just* need to destroy/create the Wayland subsurface to unmap/map the canvas, and not the entire EGL surface and wl_egl window, I finally got the [wxNotebook reproducer](https://github.com/wxWidgets/wxWidgets/issues/22580#issue-1287297419) that binds the OpenGL context only once on `wxEVT_SHOW` to work when switching tabs.

While at the some time, I could go back to the much simpler solution of using `map-event` to ensure that GDK has created the Wayland surface associated to the toplevel window, instead of needing either a backport of [wayland: Don't destroy the wl_surface on hide()](https://gitlab.gnome.org/GNOME/gtk/-/commit/5d3cec5441ca1fa28e5e77c0572e629eded419c9) to GTK3 or quite some ugly code.

I also ran various combinations of programs (wxNotebook reproducer/show-hide-reparent reproducer/SLADE/KiCad) on various compositors (Weston/Sway/GNOME-Shell) and on various OSs (Debian Bullseye VM / Arch Linux baremetal) and could find no problems related to the change.

The only noteworthy testing find is that when hiding then quickly re-showing a toplevel window containing a canvas on GNOME-Shell on my Arch Linux install, it deadlocks inside eglSwapBuffers. But aside from being a minor issue, it's think it's an unrelated issue, and this wasn't working before anyway.

src/unix/glegl.cpp
```diff
@@ -494,21 +519,25 @@ bool wxGLCanvasEGL::CreateSurface()
...
                                                    m_wlEGLWindow, nullptr);
-        m_wlFrameCallbackHandler = wl_surface_frame(surface);
-        wl_callback_add_listener(m_wlFrameCallbackHandler,
-                                 &wl_frame_listener, this);
+
+        // We need to use "map-event" instead of "map" to ensure that the
+        // widget's underlying Wayland surface has been created.
+        // Otherwise, gdk_wayland_window_get_wl_surface may return nullptr,
+        // for example when hiding then showing a window containing a canvas.
+        gtk_widget_add_events(m_widget, GDK_STRUCTURE_MASK);
```
vadz, line 534: Interesting, I was a bit surprised we needed this because I knew we were already getting `map-event` for wxTLW without doing it, but it looks like GTK enables this event mask automatically for TLWs only.

vadz (approved):
Very nice, thanks a lot! Not only it works now, but the code is also pretty clear (separate thanks for all the comments) and straightforward.

If there are no objections I'm going to merge (and backport to 3.2) this in a couple of days.

Thanks again!

swt2c (approved):
@joanbm thanks a lot for your work on this!  The changes look good to me.

swt2c:
> The only noteworthy testing find is that when hiding then very quickly re-showing a toplevel window containing a canvas on GNOME-Shell on my Arch Linux install, it deadlocks inside eglSwapBuffers. But aside from being a minor issue, it's think it's an unrelated issue, and this wasn't working before anyway.

That's strange about the deadlock issue that you saw.  Using the frame callback handler to tell us when it's OK to eglSwapBuffers seems like it ought to take care of that issue (at least that's why I started using it in the first place).

joanbm:
> > The only noteworthy testing find is that when hiding then very quickly re-showing a toplevel window containing a canvas on GNOME-Shell on my Arch Linux install, it deadlocks inside eglSwapBuffers. But aside from being a minor issue, it's think it's an unrelated issue, and this wasn't working before anyway.
> 
> That's strange about the deadlock issue that you saw. Using the frame callback handler to tell us when it's OK to eglSwapBuffers seems like it ought to take care of that issue (at least that's why I started using it in the first place).

Yeah, I took a quick look and I didn’t see anything immediately wrong with the frame callbacks, but it just deadlocks. It’s definitely on my todo list to look at this (though no promises). I just hope at the end of the day it’s something dumb instead of a rabbit hole, I’ve had enough of those recently already :)

imciner2:
> I also ran various combinations of programs (wxNotebook reproducer/show-hide-reparent reproducer/SLADE/KiCad) on various compositors (Weston/Sway/GNOME-Shell) and on various OSs (Debian Bullseye VM / Arch Linux baremetal) and could find no problems related to the change.

Thanks for testing this (especially with KiCad), since one of my worries was if destroying/creating the canvas would lead to issues with existing programs (I was concerned that maybe there would be attempts to draw on it once it was destroyed).

> The only noteworthy testing find is that when hiding then very quickly re-showing a toplevel window containing a canvas on GNOME-Shell on my Arch Linux install, it deadlocks inside eglSwapBuffers.

Does this happen on all programs, or did you just observe it on specific ones you tested?

vadz:
I don't know if it's worth waiting for more reviews, so let's merge this and see if anybody finds any new problems.

Thanks again Joan for all your work on this!

joanbm:
> > > The only noteworthy testing find is that when hiding then very quickly re-showing a toplevel window containing a canvas on GNOME-Shell on my Arch Linux install, it deadlocks inside eglSwapBuffers. But aside from being a minor issue, it's think it's an unrelated issue, and this wasn't working before anyway.
> > 
> > 
> > That's strange about the deadlock issue that you saw. Using the frame callback handler to tell us when it's OK to eglSwapBuffers seems like it ought to take care of that issue (at least that's why I started using it in the first place).
> 
> Yeah, I took a quick look and I didn’t see anything immediately wrong with the frame callbacks, but it just deadlocks. It may have something to do with the window open/close animations (Sway doesn’t have those), as the timing for how quickly how have to click for the deadlock to happen is suspiciously close to the animation length. It’s definitely on my todo list to look at this (though no promises). I just hope at the end of the day it’s something dumb instead of a rabbit hole, I’ve had enough of those recently already :)

FWIW this does not happen anymore now that `eglSwapInterval(display, 0)` from #23909 / #24014 is merged (which makes sense since it disables the logic that can make `eglSwapBuffers()` block), though I'm not sure what the actual root cause is/was.

PR #23840, opened 2023-08-31
Sizer overview documentation fix

jurmc:
Small documentation fix:
image


vadz:
Sorry, the fix is wrong and the existing documentation is correct. The screenshot shows a horizontal sizer, you can't use vertical alignment in a vertical sizer.

PR #23842, opened 2023-08-31
Update screenshots of controls under wxMSW in documentation

PBfordev:
I think it was the high time to update the controls screenshots for MS Windows. Many if not most came from Windows XP, which was IMO rather pretty, but it is 22 years old now and still using it may not make a good impression on some.

I captured all the controls on Windows 11 22H2 at 100% DPI and in light mode.

Remarks:
1. This could (should) be backported to 3.2 (except those two _about_ dialog screenshots).
2. I would prefer if `wxTreeCtrl` was displayed with `wxTR_TWIST_BUTTONS`, but I left it using its defaults.
3. I did not find the code to display the rather complex `wxBannerWindow`, so that one is not updated even if it was probably taken on Windows (7?).
4. For some reason, `wxGenericDirCtrl` docs are not (and have not been) linked from the screenshots page, probably because `Generic` in the class name, I let it be.
5. The screenshotgen utility cannot be built by cmake, moreover, some controls it does not capture correctly (W10/11, regardless of DPI), I did not investigate.


vadz:
Thanks for doing this! Will merge it soon.

PBfordev:
OK, thanks.

I think it will be for the best to update screenshots of `wxTreeCtrl` and `wxGenericDirCtrl` in a separate PR, which I will create once/if #23844 is merged.

PBfordev:
@vadz,  it seems some screenshots were not updated, e.g. wxCheckBox, wxCheckListBox, or wxDataViewCtrl, see https://docs.wxwidgets.org/latest/page_screenshots.html

The files were committed correctly, see e.g. https://github.com/wxWidgets/wxWidgets/blob/master/docs/doxygen/images/appear-checkbox-msw.png

Perhaps it will fix itself during next docs update?

vadz:
It was CloudFlare cache again because clearing it fixed the problem (remember to reload the page without cache in your browser). I have no idea why it didn't happen automatically (cc @tierra just in case you have some idea about why did it happen and how to prevent it from happening again in the future?), but thanks for noticing it!

tierra:
I probably just need to review the `Cache-Control` and `Expires` headers being served from the origin, and ensure that they're configured with more reasonable time. This is the reason you won't find TTL settings for CF edge caches, since they simply use what the origin says to use for that specific page/resource. There is also the browser cache on top of that, and we do override that and just set it to 4 hours period, for everything, but past that 4 hours, if it's still cached on the edge node selected for your location, it'll still use that unless it's past the origin-specified expiration.

Taking a quick look, I do currently have an expiration on assets (png, css, js, etc) configured at 2 weeks, and I'm not sure what's currently set for HTML content (but it's not explicitly configured, so likely some default that isn't terrible). This is probably the reason these required a purge, since the updates were to png images. I've just bumped that down to 24 hours now.

PR #23844, opened 2023-09-01
Change wxTreeCtrl default style under wxMSW

PBfordev:
Make `wxTreeCtrl` on Windows look up-to-date by default: use twisting buttons and full-row highlight and
don't display lines connecting children to the parent.

I understand that this change may seem very radical, but here is my reasoning:
1. The default style for `wxTreeCtrl` is documented as "The set of flags that are closest to the defaults for the native control for a particular toolkit" and the new default style is how a tree control usually looks in "modern" applications on Windows.
2. This is how `wxDataViewTreeCtrl` looks by default.
3. Twisting buttons are already by default used by `wxTreeCtrl` in dark mode.

If accepted, it should probably be mentioned in _Changes in behaviour_ part of the changelog.

BTW, if accepted, this would leave `wxPropertyGrid` as the only control using +/- buttons by default (it also already uses twisting ones in dark mode).


vadz:
Thanks, I see no problems with doing it in master.

PBfordev:
OK, thanks.

I will add updated screenshots of `wxTreeCtrl` and` wxGenericDirCtrl`  as another commit to this PR, once https://github.com/wxWidgets/wxWidgets/pull/23842 is merged.

vadz:
I think we should just change `wxGenericDirCtrl` to use `wxTR_DEFAULT_STYLE` for `treeStyle`, I don't see any reason to deviate from it, do you?

PBfordev:
I have also committed the updated `wxTreeCtrl` screenshot for the docs. The CI failure is unrelated.

As for `wxGenericDirCtrl`, if we were to start with `wxTR_DEFAULT_STYLE`: Since `wxGenericDirCtrl` has platform-specific setup and `wxTreeCtrl` MSW-specific oddities (some styles affect each other), for MSW the code preserving the current behaviour should be
```diff
diff --git a/src/generic/dirctrlg.cpp b/src/generic/dirctrlg.cpp
index cae0d5aeb9..678b651d1b 100644
--- a/src/generic/dirctrlg.cpp
+++ b/src/generic/dirctrlg.cpp
@@ -350,12 +350,12 @@ bool wxGenericDirCtrl::Create(wxWindow *parent,
     SetBackgroundColour(wxSystemSettings::GetColour(wxSYS_COLOUR_3DFACE));
     SetForegroundColour(wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOWTEXT));
 
-    long treeStyle = wxTR_HAS_BUTTONS;
-
-    treeStyle |= wxTR_HIDE_ROOT;
+    long treeStyle = wxTR_DEFAULT_STYLE | wxTR_HIDE_ROOT;
 
 #ifdef __WXGTK__
     treeStyle |= wxTR_NO_LINES;
+#elif defined(__WXMSW__)
+    treeStyle &= ~wxTR_LINES_AT_ROOT;
 #endif
 
     if (style & wxDIRCTRL_EDIT_LABELS)
```
But as I am Windows only, I decided to leave `wxGenericDirCtrl` alone to avoid possible regressions on non-MSW platforms which I cannot test for. 

Or I could just do safe MSW-only change like
```diff
diff --git a/src/generic/dirctrlg.cpp b/src/generic/dirctrlg.cpp
index cae0d5aeb9..75200829e8 100644
--- a/src/generic/dirctrlg.cpp
+++ b/src/generic/dirctrlg.cpp
@@ -356,6 +356,8 @@ bool wxGenericDirCtrl::Create(wxWindow *parent,
 
 #ifdef __WXGTK__
     treeStyle |= wxTR_NO_LINES;
+#elif defined(__WXMSW__)
+    treeStyle |= wxTR_NO_LINES | wxTR_TWIST_BUTTONS | wxTR_FULL_ROW_HIGHLIGHT;
 #endif
 
     if (style & wxDIRCTRL_EDIT_LABELS)
```

vadz:
I think we could just remove `wxTR_LINES_AT_ROOT` unconditionally, this is not platform-specific but just due to the fact that without this style the control looks bad with `wxTR_HIDE_ROOT`? Does it make sense at all to have lines at root with hidden root?

PBfordev:
> I think we could just remove `wxTR_LINES_AT_ROOT` unconditionally, this is not platform-specific but just due to the fact that without this style the control looks bad with `wxTR_HIDE_ROOT`? Does it make sense at all to have lines at root with hidden root?

I am not sure if you are talking about `wxTreeCtrl` or `wxGenericDirCtrl` here?

As documented, on MSW, not having `wxTR_LINES_AT_ROOT` means the root items do not have expanders, so that means no expanders are currently shown for disks in `wxGenericDirCtrl`. TBH, I am personally not a fan (missing an indication whether a node has children or not) but I understand that there must have been a reason for this decision for `wxGenericDirCtrl`. OTOH, I think it would be a bad idea to do that (lose expanders for root items) for `wxTreeCtrl`.

vadz:
I was talking about `wxGenericTreeCtrl` initially and then thought that this could apply to `wxTreeCtrl` itself, but let's leave the latter aside. For the former, I'm not really sure at all there was a good reason to add it, it just appeared in 82c1f2a390 (Make use of new Art.   Add sections to the dir dialog as per the GTK 2.4 file dialog (Home, Desktop, "/")., 2004-11-27) without any explanation and I think we could remove it...

PBfordev:
Sorry, I was talking about the style of `wxTreeCtrl` instance used by `wxGeneric**Dir**Ctrl`, NOT `wxGeneric**Tree**Ctrl `which I have no idea about. But I am probably missing something.

vadz:
> Sorry, I was talking about the style of `wxTreeCtrl` instance used by `wxGeneric**Dir**Ctrl`, NOT `wxGeneric**Tree**Ctrl `which I have no idea about. But I am probably missing something.

No, sorry, it was me, I made a typo, I did mean `wxGenericDirCtrl`. Sorry for the confusion...

PR #23849, opened 2023-09-03
Fix /tmp permissions in Cirrus CI environment

vadz:
For some reason /tmp is not world-writable any more and running apt-get
(even as root) fails. Fix this by ensuring the directory has the correct
permissions.


vadz:
Looks like this was a transient problem.

Issue #23843, opened 2023-09-01
Newly added GTK specific methods for wxTextCtrl interface are incorrectly declared public


mcorino:
The newly listed methods here:

https://github.com/wxWidgets/wxWidgets/blob/8ca3d3bc75e8fa1b851cbaeaef128575fd8af0af/interface/wx/textctrl.h#L1794-L1827

are incorrectly listed as public where they are really protected.

vadz:
Thanks for noticing this.

I think they ought to be public actually, shouldn't they?

cc @Blake-Madden

Blake-Madden:
I'm deriving from `wxTextCtrl` and using them in a protected fashion myself, but...

I could see usefulness in them being public if you wanted to use a `wxTextCtrl` and work with its buffer without needing derivation.

vadz:
I'm going to push a change to make these functions public soon.

mcorino:
That's ok. I'm ignoring them for wxRuby anyway. :wink:

PR #23852, opened 2023-09-06
Document new GTK functions as protected

Blake-Madden:


Blake-Madden:
This would close issue #23843

PR #23854, opened 2023-09-06
Add support for setting proxy to `wxWebView` implementations that support it

vadz:
Unfortunately Mac doesn't, but it's still arguably better to have it only for WebKitGTK and Edge (and Chromium, in another branch soon) than not at all.

The API is intentionally minimal for now.

### DIFF
--- a/src/gtk/webview_webkit2.cpp
+++ b/src/gtk/webview_webkit2.cpp
@@ -994,6 +994,46 @@ bool wxWebViewWebKit::SetUserAgent(const wxString& userAgent)
     return true;
 }
 
+bool wxWebViewWebKit::SetProxy(const wxString& proxy)
+{
+#if WEBKIT_CHECK_VERSION(2, 16, 0)
+    if (wx_check_webkit_version(2, 16, 0))
+    {
+        const auto context = static_cast<WebKitWebContext*>(m_config.GetNativeConfiguration());
+        wxCHECK_MSG( context, false, "no context?" );
+
+        const auto data_manager = webkit_web_context_get_website_data_manager(context);
+        wxCHECK_MSG( data_manager, false, "no data manager?" );
+
+        const auto proxy_settings = webkit_network_proxy_settings_new(
+            proxy.utf8_str(),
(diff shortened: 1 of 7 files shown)


src/gtk/webview_webkit2.cpp
```diff
@@ -994,6 +994,38 @@ bool wxWebViewWebKit::SetUserAgent(const wxString& userAgent)
...
+            proxy_settings
+        );
+
+        webkit_network_proxy_settings_free(proxy_settings);
+    }
+#else // WebKit < 2.16 doesn't support setting proxy
+    wxUnusedVar(proxy);
+#endif // WebKit 2.16+
+
+    return false;
```
Kumazuma, line 1026: it seems to return false always

src/msw/webview_edge.cpp
```diff
@@ -268,6 +268,19 @@ class wxWebViewConfigurationImplEdge : public wxWebViewConfigurationImpl
 #endif
     }
 
+    bool SetProxy(const wxString& proxy)
+    {
+#ifdef __VISUALC__
+        m_webViewEnvironmentOptions->put_AdditionalBrowserArguments(
```
Kumazuma, line 274: it can not call on MinGW?

src/gtk/webview_webkit2.cpp
vadz: Oops, indeed, thanks for noticing this!

Fixed now and I've also added `wxLogError()` calls.

src/msw/webview_edge.cpp
vadz: I am not sure about Edge support status with MinGW and I didn't have time to test it myself, so I've just copied the same test that was already present above. If you (or anybody else) uses Edge with MinGW and can check whether this `put_` call (as well as the one above) works, it would be nice!

TcT2k: The problem isn't the call as such, but the EnvironmentOptions object.
It's defined in an Edge SDK header but it's implementation uses a VisualC only COM implementation templates.

Issue #22325, opened 2022-04-17
Add run-time check for EGL version on canvas creation


imciner2:
Apparently, some older vendor drivers (*cough*, NVIDIA, *cough*) forcefully remap the EGL library on a system to be different from the one used to build wx. This means that while wx has the config check to ensure the EGL library it builds against is version >1.5, in reality it could try to use a different EGL library at runtime, that might not be version >1.5. This was first noticed by a FreeBSD user because their install of KiCad was crashing with an error about `undefined symbol eglGetPlatformDisplay` - https://gitlab.com/kicad/code/kicad/-/issues/11369 and https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=263209#c5.

It appears that EGL is supposed to provide a version at runtime if you query the `eglQueryString` with the `EGL_VERSION` name supplied. The EGL spec (https://www.khronos.org/registry/EGL/specs/eglspec.1.5.pdf, page 19) describes its behavior as:

> The format of the EGL_VERSION string is:
> `<major version.minor version><space><vendor specific info>`
> Both the major and minor portions of the version number are numeric. Their values must match the major and minor values returned by eglInitialize (see section 3.2). The vendor-specific information is optional; if present, its format and contents are implementation specific.
>
> If dpy is EGL_NO_DISPLAY, then the EGL_VERSION string describes the supported client version. If dpy is a valid, initialized display, then the EGL_VERSION string describes the supported EGL version for dpy. The client version indicates that all EGL entry points which are needed for the supported client APIs are available at runtime, while > the display version indicates which EGL functionality is supported for a display.

I think it would be good to add a runtime check to the creation of the GL canvas with an EGL backend to ensure the runtime version of the library is at least the version we build against.

swt2c:
Seems like a reasonable thing to do.  Any thoughts on what should happen if EGL is < 1.5 at runtime?  Warning?  Assert?

vadz:
I think we should use `wxLogError()` for this (BTW, the existing `wxLogMessage()` in `wxGLContext` ctor in `src/unix/glegl.cpp` should be an error too). Definitely not an assert because it's not a programmer's mistake, but something outside of the program control (and I suspect that many of the asserts in this file should be `wxLogError()`s too).

[...]

PR #23855, opened 2023-09-07
[EGL] Fail early if EGL version is less than 1.5

Steelskin:
* Add an explicit error message indicating why the `wxGLCanvas` failed to initialize on systems where the EGL runtime version is older than 1.5.
* Change `wxGLCanvasEGL::GetDisplay()` to use `eglGetDisplay()` rather than `eglGetPlatformDisplay()` when the latter is not available. This is necessary because this has to be called before EGL is initialized, which is the point where we can know the EGL runtime version. `eglGetPlatformDisplay()` is only available on EGL 1.5 and greater.

Bug: #22325

TEST=Tested the OpenGL samples on Wayland with EGL 1.5. Hard-coded the EGL version number to an older version to ensure the samples were failing with the expected error message.

### DIFF
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -370,42 +370,38 @@ EGLDisplay wxGLCanvasEGL::GetDisplay()
     typedef EGLDisplay (*GetPlatformDisplayFunc)(EGLenum platform,
                                                  void* native_display,
                                                  const EGLAttrib* attrib_list);
-    // Static initializers ensure these are only computed once.
-    static const bool s_EGLHasPlatformDisplay(
-        wxGLCanvasBase::IsExtensionInList(
-            eglQueryString(nullptr, EGL_EXTENSIONS),
-            "EGL_EXT_platform_base"));
 
-    static const GetPlatformDisplayFunc s_eglGetPlatformDisplay(
-        []() -> GetPlatformDisplayFunc
+    // Try loading the appropriate EGL function on first use.
+    static GetPlatformDisplayFunc s_eglGetPlatformDisplay = nullptr;
+    static bool s_eglGetPlatformDisplayInitialized = false;
+    if ( !s_eglGetPlatformDisplayInitialized )
+    {
+        s_eglGetPlatformDisplayInitialized = true;
+
+        if ( wxGLCanvasBase::IsExtensionInList(
+                eglQueryString(nullptr, EGL_EXTENSIONS),
+                "EGL_EXT_platform_base") )
         {
-            if ( !s_EGLHasPlatformDisplay )
-            {
-                return nullptr;
-            }
-
-            const GetPlatformDisplayFunc getPlatformDisplayFunc =
-                reinterpret_cast<GetPlatformDisplayFunc>(
+            s_eglGetPlatformDisplay = reinterpret_cast<GetPlatformDisplayFunc>(
                     eglGetProcAddress("eglGetPlatformDisplay"));
-            if ( getPlatformDisplayFunc )
+            if ( !s_eglGetPlatformDisplay )
             {
-                return getPlatformDisplayFunc;
+                // Try the fallback if not available.
+                s_eglGetPlatformDisplay = reinterpret_cast<GetPlatformDisplayFunc>(
+                    eglGetProcAddress("eglGetPlatformDisplayEXT"));
             }
+        }
+    }
 
-            // Fallback or nullptr if not available.
-            return reinterpret_cast<GetPlatformDisplayFunc>(
-                eglGetProcAddress("eglGetPlatformDisplayEXT"));
-        } ());
+    const wxDisplayInfo info = wxGetDisplayInfo();
 
     if ( !s_eglGetPlatformDisplay )
     {
-        // For backward compatibility.
-        return eglGetDisplay(
-            static_cast<EGLNativeDisplayType>(wxGetDisplay()));
+        // Use the last fallback for backward compatibility.
+        return eglGetDisplay(static_cast<EGLNativeDisplayType>(info.dpy));
     }
 
-    wxDisplayInfo info = wxGetDisplayInfo();
-    EGLenum platform;
+    EGLenum platform = 0;
     switch ( info.type )
     {
         case wxDisplayX11:
@@ -414,9 +410,11 @@ EGLDisplay wxGLCanvasEGL::GetDisplay()
         case wxDisplayWayland:
             platform = EGL_PLATFORM_WAYLAND_EXT;
             break;
-        default:
-            return EGL_NO_DISPLAY;
+        case wxDisplayNone:
+            break;
     }
+
+    wxCHECK_MSG( platform, EGL_NO_DISPLAY, "unknown display type" );
 
     return s_eglGetPlatformDisplay(platform, info.dpy, nullptr);
 }
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -364,43 +364,36 @@ bool wxGLCanvasEGL::InitVisual(const wxGLAttributes& dispAttrs)
     return m_config != nullptr;
 }
 
-
-extern "C"
-{
-
-typedef EGLDisplay (*GetPlatformDisplayFunc)(EGLenum platform,
-                                             void* native_display,
-                                             const EGLAttrib* attrib_list);
-
-// Returns the appropriate function pointer for "eglGetPlatformDisplay()" or
-// "eglGetPlatformDisplayEXT()" or nullptr if neither is available at runtime.
-static GetPlatformDisplayFunc get_edl_platform_display_func()
-{
-    // Static initializer ensures this is only computed once.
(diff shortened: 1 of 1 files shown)


swt2c:
Thanks for the PR!  After researching this a little bit, it seems that using `eglGetPlatformDisplay()` is preferred over using `eglGetDisplay()`.  It seems that it is possible to determine whether `eglGetPlatformDisplay()` is available by checking the EGL extensions?  See for example:

https://github.com/WebKit/WebKit/blob/5e7400cc3d50fe4b4657242113ef03de1868bec3/Source/WebCore/platform/graphics/x11/PlatformDisplayX11.cpp#L137

src/unix/glegl.cpp
```diff
@@ -628,11 +615,26 @@ EGLConfig *wxGLCanvasEGL::InitConfig(const wxGLAttributes& dispAttrs)
...
     }
+
+    // The runtime EGL version cannot be known until EGL has been initialized.
+    if ( egl_major < 1 || (egl_major == 1 && egl_minor < 5) )
+    {
+        // Ignore the return value here, we cannot recover at this point.
+        eglTerminate(dpy);
+        wxFAIL_MSG(wxString::Format(
+            "EGL version is %d.%d. EGL version 1.5 or greater is required.",
+            egl_major, egl_minor));
```
vadz, line 634: I wonder if this shouldn't be `wxLogError()`, so that it is reported to the end user in production builds (where `wxFAIL_MSG()` does nothing by default), as it might be something they can fix or at least tell the developer about.

vadz (commented):
Thanks! I'm not merging this yet because I'm not sure about `eglGetDisplay()` vs `eglGetPlatformDisplay()` (when in doubt I'd rather not change the existing code), but otherwise this looks good to me.

src/unix/glegl.cpp
Steelskin: I changed it. I think it's going to be more useful down the line, when we can fallback more graciously for older versions. Anyway it doesn't change much right now since we end up calling `wxFAIL_MSG()` down the line.

Steelskin:
I changed the code to check for the presence of the extension, rather than always using `eglGetDisplay()`.

src/unix/glegl.cpp
```diff
@@ -367,6 +367,17 @@ bool wxGLCanvasEGL::InitVisual(const wxGLAttributes& dispAttrs)
 /* static */
 EGLDisplay wxGLCanvasEGL::GetDisplay()
 {
+    // Static initializer ensures this is only computed once.
+    static const bool s_EGLHasPlatformDisplay =
+        wxGLCanvasBase::IsExtensionInList(
+            eglQueryString(nullptr, EGL_EXTENSIONS),
+            "EGL_EXT_platform_base");
```
swt2c, line 374: I think technically the extension for `eglGetPlatformDisplay()` should be `EGL_KHR_platform_base`.  `EGL_EXT_platform_base` seems to be for `eglGetPlatformDisplayEXT()`.

Steelskin: I am testing on Ubuntu 22.04 running under WSL, and this is what I am getting from `eglQueryString()`:

```
EGL_EXT_device_base
EGL_EXT_device_enumeration
EGL_EXT_device_query
EGL_EXT_platform_base
EGL_KHR_client_get_all_proc_addresses
EGL_EXT_client_extensions
EGL_KHR_debug
EGL_EXT_platform_device
EGL_EXT_platform_wayland
EGL_KHR_platform_wayland
EGL_EXT_platform_x11
EGL_KHR_platform_x11
EGL_EXT_platform_xcb
EGL_MESA_platform_gbm
EGL_KHR_platform_gbm
EGL_MESA_platform_surfaceless
```

In fact, I cannot find `EGL_KHR_platform_base` anywhere in the EGL spec. But I think this code is wrong either way, as `EGL_EXT_platform_base` is defined for EGL 1.4+, and `eglGetPlatformDisplay()` is EGL 1.5. I'm going to spend some more time on this.

Steelskin: Okay, I ended up using `eglGetProcAddress()` instead and added `eglGetPlatformDisplayEXT()` as an intermediate fallback. I think this should work either way but I don't have any 1.4 hardware to test.

src/unix/glegl.cpp
```diff
@@ -364,9 +364,51 @@ bool wxGLCanvasEGL::InitVisual(const wxGLAttributes& dispAttrs)
...
+extern "C"
+{
+
+typedef EGLDisplay (*GetPlatformDisplayFunc)(EGLenum platform,
+                                             void* native_display,
+                                             const EGLAttrib* attrib_list);
+
+// Returns the appropriate function pointer for "eglGetPlatformDisplay()" or
+// "eglGetPlatformDisplayEXT()" or nullptr if neither is available at runtime.
+static GetPlatformDisplayFunc get_edl_platform_display_func()
```
swt2c, line 377: Presumably you meant to name this `get_egl_platform_display_func` ?

src/unix/glegl.cpp
```diff
@@ -364,9 +364,51 @@ bool wxGLCanvasEGL::InitVisual(const wxGLAttributes& dispAttrs)
...
+{
+    // Static initializer ensures this is only computed once.
+    static const GetPlatformDisplayFunc s_getPlatformDisplayFunc([] ()
+    {
+        const GetPlatformDisplayFunc getPlatformDisplayFunc =
+            reinterpret_cast<GetPlatformDisplayFunc>(
+                eglGetProcAddress("eglGetPlatformDisplay"));
```
swt2c, line 384: It looks like just calling `eglGetProcAddress()` might not be enough.  :-(  From the [documentation](https://registry.khronos.org/EGL/sdk/docs/man/html/eglGetProcAddress.xhtml):

> A non-NULL return value does not guarantee that an extension function is actually supported at runtime. The client must also make a corresponding query, such as glGetString(GL_EXTENSIONS) for OpenGL and OpenGL ES extensions; vgGetString(VG_EXTENSIONS) for OpenVG extensions; eglQueryString(display, EGL_EXTENSIONS); or query the EGL or client API version for non-extension functions, to determine if a function is supported by EGL or a specific client API context.

So it seems it might be necessary to call `eglQueryString` also.

swt2c: Here is how GTK seems to do it, for reference:
https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/gdkdisplay.c#L1488

Steelskin: I really think the `EGL_KHR_platform_base` part is incorrect. I can at least attest of one 1.5 EGL implementation that doesn't use it. It is also nowhere to be found in the EGL documentation.

swt2c: > I really think the `EGL_KHR_platform_base` part is incorrect. I can at least attest of one 1.5 EGL implementation that doesn't use it. It is also nowhere to be found in the EGL documentation.

You might be right about `EGL_KHR_platform_base`.  [This](https://registry.khronos.org/EGL/api/egl.xml) seems to suggest that `EGL_EXT_platform_base` should be used instead.

vadz: So is the current version ok to merge, @swt2c?

swt2c: I still think that checks for `EGL_EXT_platform_base` in the query string should be added too.

Also fix the presumed typo I pointed out.

Steelskin: I am now querying for `EGL_EXT_platform_base` before trying `eglGetProcAddress()`. Please, let me know what you think.

src/unix/glegl.cpp
Steelskin: I ended up removing this in favor of embedding the static variable in `GetDisplay()` since I now need to call `wxGLCanvasBase::IsExtensionInList()`, which is protected. I can't use `wxGLCanvasBase::IsExtensionSupported()` from `wxGLCanvasEGL::GetDisplay()` because the former calls the latter.

src/unix/glegl.cpp
swt2c: This looks OK to me now.

My only nitpick would be the lambda function as this can't be backported as-is, but it's not a dealbreaker.  :-)

vadz:
Sorry, but I agree with Scott here: this lambda doesn't seem to be really useful, so I'd rather avoid it as I did in the latest commit.

If you don't mind it, I'll (squash?) merge this and probably backport to 3.2 too -- please let me know if you do. Thanks!

Steelskin:
@vadz I do not mind the change. The main difference is that the initialization is no longer thread-safe (C++ guarantees const static initializers are only called once), but I don't think that matters. Thanks for following up!

vadz:
> @vadz I do not mind the change. The main difference is that the initialization is no longer thread-safe (C++ guarantees const static initializers are only called once), but I don't think that matters. Thanks for following up!

Everything GUI-related in wxWidgets can only be done from the main GUI thread anyhow, so, indeed, I don't think it matters.

I'll merge it when the CI builds finish, thanks again!

PR #23864, opened 2023-09-11
wxBitmap fixes for wxQt

AliKet:
Although the tests are all passing now, testing and reviewing this PR before merging is really appreciated, TIA!

```
> xvfb-run -a -s '-screen 0 1600x1200x24' ./test_gui [bitmap]
build: 3.3.0 (UTF-8,compiler with C++ ABI compatible with gcc 4,STL containers,compatible with 3.2)
compiled using gcc 8.4
running under Linux 4.15.0-213-generic i686 as aliket
Filters: [bitmap]
All tests passed (4046 assertions in 14 test cases)
```

### DIFF
--- a/include/wx/rawbmp.h
+++ b/include/wx/rawbmp.h
@@ -191,6 +191,24 @@ typedef wxPixelFormat<unsigned char, 24, 0, 1, 2> wxImagePixelFormat;
     typedef wxPixelFormat<unsigned char, 24, 0, 1, 2> wxNativePixelFormat;
 
     #define wxPIXEL_FORMAT_ALPHA 3
+
+    template<>
+    struct wxPixelFormat<void, 1, -1, -1, -1, -1, bool>
+    {
+        // the type which may hold the entire pixel value
+        typedef bool PixelType;
+
+        // size of one pixel in bits
+        static const int BitsPerPixel = 1;
+
+        // size of one pixel in ChannelType units (usually bytes)
+        static const int SizePixel = 1;
+
+        // true if we have an alpha channel (together with the other channels, this
+        // doesn't cover the case of wxImage which stores alpha separately)
+        enum { HasAlpha = false };
+    };
+    typedef wxPixelFormat<void, 1, -1, -1, -1, -1, bool> wxMonoPixelFormat;
 #endif
 
 // the (most common) native format for bitmaps with alpha channel
@@ -678,7 +696,7 @@ struct wxPixelDataOut<wxBitmap>
     // private: -- see comment in the beginning of the file
 
         // the bitmap we're associated with
-        wxBitmap m_bmp;
+        wxBitmap& m_bmp;
 
         // the iterator pointing to the image origin
         Iterator m_pixels;
@@ -695,7 +713,7 @@ struct wxPixelDataOut<wxBitmap>
     };
 };
 
-    #if defined(__WXMSW__)
+    #if defined(__WXMSW__) || defined(__WXQT__)
         template <>
         struct wxPixelDataOut<wxBitmap>::wxPixelDataIn<wxMonoPixelFormat> : public wxPixelDataBase
         {
@@ -722,7 +740,11 @@ struct wxPixelDataOut<wxBitmap>
                 operator bool() const
                 {
                     wxByte mask = static_cast<wxByte>(1 << m_bit);
+                #ifndef __WXQT__
                     return (*m_ptr & mask) != 0;
+                #else
+                    return (*m_ptr & mask) == 0;
+                #endif
                 }
 
             private:
@@ -896,7 +918,7 @@ struct wxPixelDataOut<wxBitmap>
         // private: -- see comment in the beginning of the file
 
             // the bitmap we're associated with
-            wxBitmap m_bmp;
+            wxBitmap& m_bmp;
 
             // the iterator pointing to the image origin
             Iterator m_pixels;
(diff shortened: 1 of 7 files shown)


tests/graphics/bitmap.cpp
```diff
@@ -819,7 +819,7 @@ TEST_CASE("BitmapTestCase::DrawAlpha", "[bitmap][draw][alpha]")
 
     // Bitmap to be drawn
     wxBitmap bmp(w, h, 32);
-#if defined(__WXMSW__) || defined(__WXOSX__)
+#if defined(__WXMSW__) || defined(__WXOSX__) || defined(__WXQT__)
     bmp.UseAlpha();
 #endif // __WXMSW__ || __WXOSX__
```
AliKet, line 824: Oops! this call is not needed in `wxQt`.

tests/graphics/bitmap.cpp
```diff
@@ -934,7 +934,7 @@ TEST_CASE("BitmapTestCase::DrawAlphaWithMask", "[bitmap][draw][alpha][withmask]"
 
      // Bitmap with mask to be drawn
      wxBitmap bmp(w, h, 32);
-#if defined(__WXMSW__) || defined(__WXOSX__)
+#if defined(__WXMSW__) || defined(__WXOSX__) || defined(__WXQT__)
      bmp.UseAlpha();
 #endif // __WXMSW__ || __WXOSX__
```
AliKet, line 939: Not needed in wxQt

include/wx/rawbmp.h
```diff
@@ -678,7 +696,7 @@ struct wxPixelDataOut<wxBitmap>
     // private: -- see comment in the beginning of the file
 
         // the bitmap we're associated with
-        wxBitmap m_bmp;
+        wxBitmap& m_bmp;
```
vadz, line 699: Sorry, why? I don't see anything that requires changing this bitmap now.

include/wx/rawbmp.h
```diff
@@ -722,7 +740,11 @@ struct wxPixelDataOut<wxBitmap>
                 operator bool() const
                 {
                     wxByte mask = static_cast<wxByte>(1 << m_bit);
+                #ifndef __WXQT__
```
vadz, line 743: This looks like the indicator that a mask in wxQt is reversed compared to wx conventions which is surely problematic and it looks like it would be better to fix it by inverting the mask when constructing it?

src/qt/bitmap.cpp
```diff
@@ -71,47 +74,63 @@ static wxImage ConvertImage( QImage qtImage )
...
+            const unsigned char g = data[1];
+            const unsigned char b = data[2];
+
+            colour = (a << 24) + (r << 16) + (g << 8) + b;
+
+            if ( !qtMask.isNull() )
+            {
+                if ( qRgb(r, g, b) != maskedColour )
+                    qtMask.setPixel(x, y, Qt::color1); // set to 1
+            }
```
vadz, line 115: I.e. maybe the logic here needs to be reversed?

tests/graphics/bitmap.cpp
```diff
@@ -306,13 +306,13 @@ TEST_CASE("BitmapTestCase::ToImage", "[bitmap][image][convertto]")
             {
                 wxColour bmpc(iBmp.Red(), iBmp.Green(), iBmp.Blue(), iBmp.Alpha());
                 wxColour imgc(image.GetRed(x, y), image.GetGreen(x, y), image.GetBlue(x, y), image.GetAlpha(x,y));
-#if defined(__WXMSW__) || defined(__WXOSX__)
+#if defined(__WXMSW__) || defined(__WXOSX__) //|| defined(__WXQT__)
```
vadz, line 309: It would be nice to leave a comment explaining why is wxQt excluded here unlike everywhere else...

vadz (commented):
Thanks for working on this! I don't know anything about Qt bitmaps/masks formats, but I find it suspicious that we must have wxQt-specific logic in common code. We can live with it if we really have to, but it would be nicer if we could avoid it.

include/wx/rawbmp.h
AliKet: I thought the lack of the call to `AllocExclusive()` in `wxBitmap::GetRawData()` 
was a mistake (unless it's by design) because accessing raw bitmap data should do a **COW** to break data sharing which is not the case currently. So if we (you?) decide to change `GetRawData()` implementation for each port to follow **wxDFB**:
https://github.com/wxWidgets/wxWidgets/blob/eb8a57760e61e72959593200d4df8d8b36dfb87a/src/dfb/bitmap.cpp#L506
then this change is necessary. If not, then I can revert this if you like ?

include/wx/rawbmp.h
AliKet: It's unfortunate that we have to reverse the logic here, because the code checking to see whether the current pixel is transparent or not uses this `bool operator` to determine that. and **Qt** and **wx** seem to operate in reverse manner when applying the mask  to bitmap.

AliKet:
I converted this to draft for now in the hope I can figure out how to get rid of the hack in `operator bool()`

AliKet:
@vadz I managed to get things work without the hack in `operator bool()`. but the `graphbitmap` test fails now. I had to change this line: https://github.com/wxWidgets/wxWidgets/blob/eb8a57760e61e72959593200d4df8d8b36dfb87a/src/generic/graphicc.cpp#L1738C1-L1738C1
 to:
`                if (p.Red()+p.Green()+p.Blue() != 0)`

for the test to succeed. because the mask under wxQt is inverted, which means non-zero values in the mask (white colour) represent transparent regions in the bitmap instead of zero values. other then that everything seems to work correctly with the new changes.

My question now: is it ok to change the aforementioned line as I said above ? if yes, I'll create a new PR soon with the necessary changes only ?

vadz:
Yes, I think it's ok to change this, although we might want to add some helper `IsMasked()` that would be defined differently for wxQt to make this more clear.

Thanks!

AliKet:
Replaced bu #23872

PR #23867, opened 2023-09-11
Added ForceDark

bartvanstiphout:
With this pull request, the ForceDark function is added to the MacOS version. It will force the app to use the Dark Mode color scheme.

### DIFF
--- a/include/wx/settings.h
+++ b/include/wx/settings.h
@@ -183,6 +183,9 @@ public:
     // Return true if this application itself uses a dark theme.
     bool IsDark() const;
 
+    // Forces the Application to use Dark Mode. Currently only availalbe on MacOS
+    void ForceDark() const;
+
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for the addition but I'm not sure if we really want to add this to `wxSystemAppearance` which, like the related `wxSystemSettings` class, has always been limited to querying information about the system so far.

As we already have `wxApp::MSWEnableDarkMode()` in wxMSW, I think it would be better to add `EnableDarkMode()` with the same flags (but without MSW-specific `wxDarkModeSettings`) to `wxApp` on all platforms, as I think it should be also possible to implement it in wxGTK, see #23764.

Also, and in any case, we need the new function to be documented in the corresponding interface header (e.g. `interface/wx/app.h` if we implement the suggestion above).

PR #23870, opened 2023-09-13
Fix page origin in wxGTK when printing in landscape mode

### LINKED ISSUE #23860: Printed image is shifted out of paper bounds on wxGTK.
### Description

I've modified the "printing" sample to print a blue rectangle: [printing.cpp.txt](https://github.com/wxWidgets/wxWidgets/files/12565763/printing.cpp.txt)

When printing on GTK, on landscape A4 paper, the image is shifted by 7.87 mm to the right so part of the rectangle is clipped:
image

The preview from wxWidgets looks fine:
image

The preview from the OS printing dialog when choosing "Print to File" does **not** look fine:
image

PDF from "Print to File":
[wx_print_bug.pdf](https://github.com/wxWidgets/wxWidgets/files/12565759/wx_print_bug.pdf)

Affects real printers too.

### References
https://gitlab.com/kicad/code/kicad/-/issues/1907
https://forum.kicad.info/t/print-from-schematic-editor-is-not-centered-on-paper/44540

In KiCad's PCB Editor (vs Schematic Editor), the printout rendering is re-implemented using Cairo, similar to wxGtkPrinterDCImpl, but in a way that this problem does not appear.

- wxWidgets version: 3.2.2
- wxWidgets port: wxGTK
- OS: Linux Mint 21.1, also reported on Ubuntu 23.04 and others
- GTK version: 3.24.33
- Which GDK: X11

vadz:
Using the "left" and "top" margin doesn't work in this case and we would
actually need to use "bottom" and "left" ones to offset the origin
correctly in landscape orientation (and something else for the reverse
landscape, and reverse portrait one, too), but there actually seems to
be a simpler way to put the origin where we want it without translating
it explicitly: we can just tell GTK that we want to use the full page,
rather than just the printable area, by setting "use-full-page" property
to true, so simply do this instead.

Co-authored-by: Alex Shvartzkop <dudesuchamazing@gmail.com>

Closes #23860.

### DIFF
--- a/src/gtk/print.cpp
+++ b/src/gtk/print.cpp
@@ -702,6 +702,11 @@ int wxGtkPrintDialog::ShowModal()
     gtk_print_operation_set_default_page_setup (printOp, pgSetup);
     g_object_unref(pgSetup);
 
+    // By default the origin of the Cairo context is in the upper left
+    // corner of the printable area, but wx convention is to have it in
+    // the upper left corner of the paper, so change this.
+    gtk_print_operation_set_use_full_page(printOp, TRUE);
+
     // Show the dialog if needed.
     GError* gError = nullptr;
     GtkPrintOperationResult response = gtk_print_operation_run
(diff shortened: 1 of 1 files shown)


vadz:
Just to check, is this ok for you, @dsa-t ?

dsa-t:
Looks good to me. Thanks!

dsa-t:
@vadz can this get merged and backported to 3.2.3?

vadz:
Merged in master and will be part of 3.2.3 when #23891 is merged.

PR #23872, opened 2023-09-13
Use premultiplied ARGB32 format for wxBitmap under wxQt

AliKet:
[bitmap] & [graphbitmap] tests pass successfully with these changes. also the image and drawing samples work as expected.

### DIFF
--- a/include/wx/qt/bitmap.h
+++ b/include/wx/qt/bitmap.h
@@ -23,8 +23,10 @@ public:
     wxBitmap(int width, int height, const wxDC& dc);
     wxBitmap(const char* const* bits);
     wxBitmap(const wxString &filename, wxBitmapType type = wxBITMAP_TYPE_XPM);
+#if wxUSE_IMAGE
     wxBitmap(const wxImage& image, int depth = wxBITMAP_SCREEN_DEPTH, double scale = 1.0);
     wxBitmap(const wxImage& image, const wxDC& dc);
+#endif // wxUSE_IMAGE
 
     // Convert from wxIcon / wxCursor
     wxBitmap(const wxIcon& icon) { CopyFromIcon(icon); }
@@ -82,7 +84,9 @@ protected:
     virtual wxGDIRefData *CloneGDIRefData(const wxGDIRefData *data) const override;
 
 private:
+#if wxUSE_IMAGE
     void InitFromImage(const wxImage& image, int depth, double WXUNUSED(scale));
+#endif
 
     wxDECLARE_DYNAMIC_CLASS(wxBitmap);
 };
--- a/include/wx/qt/bitmap.h
+++ b/include/wx/qt/bitmap.h
@@ -109,7 +109,7 @@ public:
     virtual ~wxMask();
 
     // Construct a mask from QBitmap, takes ownership.
-    wxMask(QBitmap* qtBitmap);
+    explicit wxMask(QBitmap* qtBitmap);
 
     wxBitmap GetBitmap() const;
 
--- a/include/wx/qt/bitmap.h
+++ b/include/wx/qt/bitmap.h
@@ -72,6 +72,9 @@ public:
     // disappear in the future
     bool HasAlpha() const override;
 
+    // Blend mask with alpha channel and remove the mask
+    void QtBlendMaskWithAlpha();
+
     QPixmap *GetHandle() const;
 
 protected:
--- a/include/wx/qt/bitmap.h
+++ b/include/wx/qt/bitmap.h
@@ -105,6 +105,9 @@ public:
     wxMask(const wxBitmap& bitmap);
     virtual ~wxMask();
 
+    // Construct a mask from QBitmap, takes ownership.
+    wxMask(QBitmap* qtBitmap);
+
     wxBitmap GetBitmap() const;
 
     // Implementation
--- a/src/generic/graphicc.cpp
+++ b/src/generic/graphicc.cpp
@@ -586,7 +586,7 @@ protected:
 #ifdef __WXMAC__
     CGContextRef m_cgContext;
 #endif // __WXMAC__
-    
+
 private:
     cairo_t* m_context;
     cairo_matrix_t m_internalTransform;
@@ -1608,6 +1608,21 @@ wxCairoBitmapData::wxCairoBitmapData( wxGraphicsRenderer* renderer, cairo_surfac
     m_buffer = nullptr;
 }
 
+namespace
+{
+#ifdef wxHAS_RAW_BITMAP
+// Return true if the pixel is masked under this platform
+inline bool IsMasked(wxNativePixelData::Iterator& p)
+{
+#ifndef __WXQT__
+    return p.Red()+p.Green()+p.Blue() == 0;
+#else
+    return p.Red()+p.Green()+p.Blue() != 0;
+#endif
+}
+#endif
+}
+
 wxCairoBitmapData::wxCairoBitmapData( wxGraphicsRenderer* renderer, const wxBitmap& bmp )
     : wxGraphicsBitmapData(renderer)
 {
@@ -1622,14 +1637,14 @@ wxCairoBitmapData::wxCairoBitmapData( wxGraphicsRenderer* renderer, const wxBitm
     // different format and iterator than if it doesn't...
     const bool isSrcBpp32 = bmp.GetDepth() == 32;
 
-#if defined(__WXMSW__) || defined(__WXOSX__)
-    // Under MSW and OSX we can have 32 bpp xRGB bitmaps (without alpha).
+#if defined(__WXMSW__) || defined(__WXOSX__) || defined(__WXQT__)
+    // Under MSW, OSX and Qt we can have 32 bpp xRGB bitmaps (without alpha).
     const bool hasAlpha = bmp.HasAlpha();
 #endif
 
     cairo_format_t bufferFormat =
-    // Under MSW and OSX we can have 32 bpp xRGB bitmaps (without alpha).
-#if defined(__WXMSW__) || defined(__WXOSX__)
+    // Under MSW, OSX and Qt we can have 32 bpp xRGB bitmaps (without alpha).
+#if defined(__WXMSW__) || defined(__WXOSX__) || defined(__WXQT__)
         (isSrcBpp32 && hasAlpha) || bmp.GetMask() != nullptr
 #else
         isSrcBpp32 || bmp.GetMask() != nullptr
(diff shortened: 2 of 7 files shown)


tests/graphics/bitmap.cpp
```diff
@@ -50,10 +50,22 @@
...
 #endif // __WXMSW__
-#ifdef __WXOSX__
+#if defined(__WXOSX__) || defined(__WXQT__)
 // 32 bpp xRGB bitmaps are native ones
 typedef wxNativePixelData wxNative32PixelData;
-#endif // __WXOSX__
+#endif // __WXOSX__ || __WXQT__
+
+// masked/unmasked colours are revered in wxQt:
+
```
AliKet, line 59: reversed

vadz:
Looks good to me, thanks!

@a-wi Would you like to take a look at this? Please let me know if I should wait with merging this until you can do it if you'd like.

AliKet:
Oops! last commit should be part of [d9e3efc](https://github.com/wxWidgets/wxWidgets/pull/23872/commits/d9e3efcf204ae5ff6222553e8f8cf6601993d299) sorry!

AliKet:
@vadz now the wxTreeCtrl need to be fixed. I mean we need to apply the mask (if any) nefore drawing the icons like the following patch (which solves the problem btw):
```
diff --git a/src/qt/treectrl.cpp b/src/qt/treectrl.cpp
index 5b6fe397d0..19523f595a 100644
--- a/src/qt/treectrl.cpp
+++ b/src/qt/treectrl.cpp
@@ -20,6 +20,7 @@
 #include <QtWidgets/QTreeWidget>
 #include <QtWidgets/QHeaderView>
 #include <QtWidgets/QScrollBar>
+#include <QtGui/QBitmap>
 #include <QtGui/QPainter>
 
 namespace
@@ -268,7 +269,15 @@ protected:
             const wxBitmap bitmap = imageList->GetBitmap(imageIndex);
             const QRect rect = visualRect(index);
             const int offset = (rect.height() / 2) - (bitmap.GetHeight() / 2);
-            painter->drawPixmap(rect.topLeft() + QPoint(0,offset), *bitmap.GetHandle());
+
+            QPixmap pix = *bitmap.GetHandle();
+            if ( bitmap.GetMask() && bitmap.GetMask()->GetHandle() )
+            {
+                QBitmap qtMask = *bitmap.GetMask()->GetHandle();
+                pix.setMask(qtMask);
+            }
+
+            painter->drawPixmap(rect.topLeft() + QPoint(0,offset), pix);
         }
     }
```
...but I'm asking if you have a better solution than this. like applying the masks when constructing the imagelist and not at the time of drawing the icons ?

Here is the affected treectrl (in the widget sample):
Screenshot from 2023-09-14 17-28-01

The treectrl sample is not affected though!

TIA!

vadz:
Sorry, I'm lost here: as we are using Qt conventions for the mask internally (aren't we?), drawing the bitmap using Qt functions should work without any changes, so I have no idea why it doesn't. Is there a missing `setMask()` call when setting mask at wx level?

AliKet:
> Is there a missing `setMask()` call when setting mask at wx level?

Sorry, what do you mean by `at wx level` ? that we should apply the mask at wxBitmap level itself ?

If so, then none of the other ports do that. But instead, they explicitly request the mask to be applied to bitmap at drawing moment (see [wxDrawImageBitmap()](https://github.com/wxWidgets/wxWidgets/blob/e570d85164af9a2bb1bf6ebb9a061cda97f1ae56/include/wx/generic/private/drawbitmap.h#L20))

Or, for wxMSW case, [wxBitmapToIconOrCursor()](https://github.com/wxWidgets/wxWidgets/blob/e570d85164af9a2bb1bf6ebb9a061cda97f1ae56/src/msw/bitmap.cpp#L1821) takes care of converting wxBitmap with mask to a valid wxIcon.

Moreover, when looking at [wxQtDCImpl::DoDrawBitmap()](https://github.com/wxWidgets/wxWidgets/blob/e570d85164af9a2bb1bf6ebb9a061cda97f1ae56/src/qt/dc.cpp#L691) implementation, it is clear that the associated mask doesn't necessarily have to be applied to bitmap when we get the native handle via `GetHandle()`

So I think we should apply the mask explicitly somewhere before drawing the bitmap!

vadz:
> > Is there a missing `setMask()` call when setting mask at wx level?
> 
> Sorry, what do you mean by `at wx level` ? that we should apply the mask at wxBitmap level itself ?

Yes, looking at

```cpp
                pix.setMask(qtMask);
```

above, it looks like we should do it.

> If so, then none of the other ports do that.

But they don't have any equivalent of `QPixmap::setMask()`. E.g. under MSW a `wxBitmap` is just an `HBITMAP` and it doesn't have any mask, the mask can be only applied when we draw it, which is why we do it like this.

> So I think we should apply the mask explicitly somewhere before drawing the bitmap!

Considering that typically a bitmap would be drawn many more times than it is created, I think we should apply the mask when creating it if possible. Applying it when drawing it should also work, of course, but doesn't seem optimal.

src/generic/graphicc.cpp
```diff
@@ -1608,6 +1608,21 @@ wxCairoBitmapData::wxCairoBitmapData( wxGraphicsRenderer* renderer, cairo_surfac
...
+namespace
+{
+#ifdef wxHAS_RAW_BITMAP
+// Return true if the pixel is masked under this platform
+inline bool IsMasked(wxNativePixelData::Iterator& p)
+{
+#ifndef __WXQT__
+    return p.Red()+p.Green()+p.Blue() == 0;
+#else
+    return p.Red()+p.Green()+p.Blue() != 0;
```
a-wi, line 1620: `wxMask` documentation (https://docs.wxwidgets.org/trunk/classwx_mask.html) states that "masked area is black and the unmasked area is white" and `wxQT` should conform this convention.

AliKet: From the user's POV, wxQt conforms to the documentation. that is, applying the same mask on the same bitmap should give the same result on all platforms.

For the implementation, one should know that Qt applies the mask in opposite way to what wx documentation says. IOW, masked area is white and the unmasked area is black! and you can see how Qt actually applies the mask by the following patch:
```
index 5f32257c6b..7170fd1e19 100644
@@ -26,6 +26,9 @@
     #include "wx/wx.h"
 #endif
 
+#include <QtGui/QPixmap>
+#include <QtGui/QBitmap>
+
 // resources
@@ -188,16 +191,14 @@ void MyFrame::OnQuit(wxCommandEvent& WXUNUSED(event))
 
 {
-                 (
-                    "\n"
-                 ),
-                 this);
+    QPixmap pix;
+    pix.load("colour.bmp");
+    QBitmap mask;
+    mask.load("mask.bmp");
+
+    if ( !pix.isNull() && !mask.isNull() )
+    {
+        pix.setMask(mask);
+        pix.save("colour_with_mask.bmp");
+    }
 }
```
colour.bmp:
colour

mask.bmp:
mask

colour_with_mask.bmp (expected):
colour_with_mask_expected

colour_with_mask.bmp(actual result!):
colour_with_mask_result

So I don't know how to make things more transparent under wxQt without complicating wxBitmap implementation further and making it inefficient at best!

a-wi: I'm not proficient with QT but it seems that in `wxImage ConvertImage(QImage qtImage)` when you convert 1bpp mask to RGB with
`if ( qtImage.depth() == 1 )
      qtImage = qtImage.convertToFormat(QImage::Format_RGB32);
`
you could also invert it using `QImage::invertPixels()`.
I think it shouldn't degrade performance very much (`QImage::converttoFormat()` itself is probably expensive).
With inverting  QImage internally you should be able to handle wx-style mask.

AliKet:
@vadz, @a-wi  I have a question:
If we apply the mask unconditionally, then how about [ConvertToImage()](https://docs.wxwidgets.org/3.2/classwx_bitmap.html#aad1c403e6e9148835907e3970a08e9bf) function ?
because after the call to `setMask()` (**which merges the mask with the pixmap's alpha channel**) the conversion to image can no longer preserve mask information as documented ? if this not a problem, then we should disable some tests under `wxQt` and update documentation to mention the exception.

Alternatively, we can change `GetHandle()` implementation under wxQt like this:

```
// If @mergeMask is true the mask will be merged with the pixmap's alpha channel
// Notice that calling this function with @mergeMask set to true will also delete the
// associated wxMask object so future calls to this function will not try to set the mask
// again.
QPixmap *GetHandle(bool mergeMask = false) const;

QPixmap *wxBitmap::GetHandle(bool mergeMask) const
{
    if ( mergeMask && M_MASK && M_MASK->GetHandle() )
    {
        M_PIXDATA.setMask(*M_MASK->GetHandle());
        wxDELETE(M_MASK);
    }

    return ( m_refData != nullptr ) ? &M_PIXDATA : nullptr;
}
```
I tried the above and it works perfectly well.

What do you think ?

vadz:
I think it would be best to add a separate function returning the Qt bitmap handle with the mask applied to it, seeing `GetHandle(true)` would be very unreadable. I.e. we could have 2 handles in `M_PIXDATA`: one as now and the other one with the mask, that would be initialized on first use and then preserved (until the mask or bitmap changes).

Would this solve the problem?

AliKet:
@vadz I think I'm done with this unless other changes are requested. TIA!

include/wx/qt/bitmap.h
```diff
@@ -105,6 +108,9 @@ class WXDLLIMPEXP_CORE wxMask : public wxMaskBase
     wxMask(const wxBitmap& bitmap);
     virtual ~wxMask();
 
+    // Construct a mask from QBitmap, takes ownership.
+    wxMask(QBitmap* qtBitmap);
```
vadz, line 112: I think it should be
```suggestion
    explicit wxMask(QBitmap* qtBitmap);
```

src/generic/imaglist.cpp
```diff
@@ -106,6 +106,10 @@ wxBitmap wxGenericImageList::GetImageListBitmap(const wxBitmap& bitmap) const
 #endif // wxUSE_IMAGE
     }
 
+#ifdef __WXQT__
+    bmpResized.QtBlendMaskWithAlpha();
+#endif
+
```
vadz, line 112: Just trying to understand: is this just an optimization, i.e. would it still work correctly but just slower because it would have to convert the mask multiple times?

vadz (commented):
This looks much better now, thanks a lot for working on this!

I have just one question left, but whatever the answer to it, this should be good to merge, thanks again!

src/generic/imaglist.cpp
AliKet: > Just trying to understand: is this just an optimization, i.e. would it still work correctly...

No it wouldn't. because we have to apply the mask explicitly in:
1- The `wxBitmap` itself. but again, `wxQt` would diverge from the documented behaviour. i.e. converting `wxBitmap` to `wxImage` wouldn't preserve mask information anymore.
2- `wxTreeCtrl` and `wxListCtrl` classes. and I think you agree that it's not wise to call `setMask()` over and over again each time the icons need to be drawn even if the performance hit won't be noticeable in practice.

So I chose to put the call here (in `wxGenericImageList::GetImageListBitmap()`) as is done in wxMSW port by calling MSWBlendMaskWithAlpha() in its own implementation of wxImageList.

> but just slower because it would have to convert the mask multiple times?

Yes see above.

vadz: OK, the part I didn't realize is that wxQt has native `wx{Tree,List}Ctrl` implementations and so we need to return the native bitmap with mask applied to it when using them. I.e. this wouldn't be necessary if we used generic implementations of these controls, because `wxDC::DrawBitmap()` would draw the bitmap correctly, applying its mask on the fly, but it is needed to work with the native implementations.

If so, i.e. if my understanding is finally correct, then I agree that this is fine. It would, admittedly, be nicer if we added a Qt-specific `wxImageList` implementation too that would inherit from the generic one and just add this `QtBlendMaskWithAlpha()` call, but it would be a lot of work for not that much gain so let's leave it like this.

I'd just like to add a comment like this:
```cpp
// In wxQt the returned bitmap is used with native list/tree controls which require
// the bitmap to already have its mask applied to it, as it's used directly by Qt
// and not via our wxDC::DrawBitmap() which would apply the mask itself, so we must
// take it into account here -- the only alternative would be for the controls to do
// it themselves, but then this would happen every time the bitmap is drawn, which
// would be less efficient than doing it just once here.
```

Again, this is assuming I do understand things correctly now -- if not, please correct me. Otherwise I'll add this comment and merge this soon.

Thanks again!

AliKet: Yes, exactly what you just described and the comment is indeed relevant. and I would be grateful if you add this comment when you merge this. TIA!

PR #23873, opened 2023-09-14
Fix a crash in wxRearrangeCtrl

AliKet:
`wxRearrangeCtrl` ctor trying to use `m_order` before it's actually initialized.


vadz:
Thanks, will merge soon (note to self: CI failure is bogus).

PR #23877, opened 2023-09-16
Fix recent wxGTK regression when using wide string entry point

vadz:
The changes of 5f17915e63 (Store original command line options passed to
main(), 2023-09-02) didn't account for the possibility of the
application using an entry point such as wxEntry(), wxEntryStart() or
wxInitialize() with wide character argv under Unix -- where this is
typically not the case but still possible and if any of the wide string
overloads of these functions were called, wxInitData::arg[cv] remained
uninitialized resulting in a crash in gtk_init_check() later.

Fix this by ensuring we do initialize them in this case too and also
reconstruct the original narrow string argvA from wide string arguments
in this case, which is a bit wasteful but unavoidable if we're not
provided access to the original main() arguments.

### DIFF
--- a/include/wx/private/init.h
+++ b/include/wx/private/init.h
@@ -21,8 +21,13 @@ struct WXDLLIMPEXP_BASE wxInitData
     // Get the single global object.
     static wxInitData& Get();
 
-    // Initialize from ANSI command line arguments.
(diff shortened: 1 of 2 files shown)


dominicletz:
Thanks @vadz tested this and it works perfectly well!

vadz:
Merged now.

PR #23881, opened 2023-09-18
Fix logging bug in the event sample

### LINKED ISSUE #23767: Bug in the events sample
### Description
The event sample demonstrates using gesture events by showing a wxFrame-derived `MyGestureFrame`  along with the sample main frame. 

`MyGestureFrame` takes over the log target with
https://github.com/wxWidgets/wxWidgets/blob/3ae3a4e1f519741132c51e5aa8f83d6f05174501/samples/event/gestures.cpp#L26

This is problematic for two reasons:
1. All the log messages (i.e., even those triggered from the main frame menu / buttons) are shown in the `MyGestureFrame` log.
2. When `MyGestureFrame` is destroyed, so is the log target, so any further use of `wxLogxxx()` makes the sample crash.

I am not sure how to best solve it, perhaps the simplest fix would be converting `MyGestureFrame` to a modal dialog and restore the log target when the dialog is closed?

PBfordev:
The gestures frame created its own log target and deleted the previous one but did not restore it when being closed.

This left the current log target invalid which led to a crash when calling any wxLog*() function after gestures frame was closed.

Fix this by restoring the previous log target when gestures frame is closed.

Closes #23767.

Should be backported to 3.2.

### DIFF
--- a/samples/event/gestures.cpp
+++ b/samples/event/gestures.cpp
@@ -23,7 +23,7 @@ MyGestureFrame::MyGestureFrame()
     SetSizeHints(wxMin(800,dsplySz.GetWidth()), wxMin(600,dsplySz.GetHeight()));
 
     // Log to the text control
-    delete wxLog::SetActiveTarget(new wxLogTextCtrl(m_logText));
+    m_logOld = wxLog::SetActiveTarget(new wxLogTextCtrl(m_logText));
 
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for fixing this! Will merge/backport soon.

vadz:
Done now.

Issue #18227, opened 2018-09-22
AddCatalog doesn't fail with wxLANGUAGE_DEFAULT


wxtrac:
*Issue migrated from trac ticket # 18227*

**component:** wxMSW | **priority:** normal | **keywords:** addcatalog translations locale

#### 2018-09-22 17:47:19: @lanurmi created the issue

___
When `wxLocale` is inited with `wxLANGUAGE_DEFAULT`, and the default language is non-English, `wxTranslations::AddCatalog` returns `true` even if the requested catalog in the *preferred language* was not found. This is specific to Windows, possibly also Mac, but *nix is not affected.

What happens is that on non-English Windows (for example, Windows 7 in Finnish), [GetPreferredUILanguage](https://github.com/wxWidgets/wxWidgets/blob/cc931612eec2e3ea49200ebff45042135f3c3f9c/src/common/translation.cpp#L130) returns *two* languages: `fi_FI` and `en_US`. Now, if the catalog for Finnish was not found, the original msgid language English is still considered to satisfy what the caller asked for, so no error is returned. This is however not what the caller expected, and on *nix an error would be returned.

I also tested this on a Windows 10 installed in German, and again two preferred languages are returned. On neither machine has the user (me) done anything to choose English as a secondary language.

In wx 2.8 `AddCatalog` used to fail when a catalog in the default language was not found. On *nix it still does.

Steps to reproduce in the unmodified *internat* sample:
0. Have something else than English as the preferred language.
1. Remove all **.mo* files, so no catalog exists.
2. Select *System default* as the language in *internat*.
3. Observe there is no error message about failing to load the catalog.

vadz:
After returning to this, I realize that I actually have a problem with adding `msgIdLanguage` to the "available translations" which was done back in 01f953efb2 (Add wxTranslations::GetBestTranslation()., 2012-09-08). This seems wrong because this translation is not actually available -- it is fine not to use it at all, but this is not quite the same thing.

Unfortunately to solve this we'd need to change `AddCatalog()` to return a 3-state enum with possible values of "Loaded", "Not loaded" and "Not needed" which is going to be difficult to implement in a backwards compatible way. So I guess the only thing we can do is to add yet _another_ function behaving like this.

[...]

PR #23886, opened 2023-09-20
Don´t search for wxstd MO files w/ version number, just look for wxstd.mo

Blake-Madden:
Searching for a translated "wxstd-3.3.mo" file is supposed to fallback to searching for a translated "wxstd.mo" file if it can't find it. However, it does not work that way because English is the default `wxLanguage` argument that `AddStdCatalog()` passes to `AddCatalog()`. What happens then is that `AddCatalog()` can't find a translated "wxstd-3.3.mo" file and falls back to English and returns true. Then `AddStdCatalog()` accepts that true response and never attempts to call `AddCatalog()` again with just "wxstd". Hence, my "fix" is to simply look for a translated "wxstd.mo" and not try to look for a file with the version number in it.

I've tested this on Spanish Windows and unless I'm misunderstanding something, it seems like the fallback branch to look for "wxstd.mo" would never get ran--it will always fall back to English and give up if wxstd-3.3.mo isn't found.

### DIFF
--- a/src/common/translation.cpp
+++ b/src/common/translation.cpp
@@ -1312,13 +1312,6 @@ wxArrayString wxTranslations::GetAvailableTranslations(const wxString& domain) c
 
 bool wxTranslations::AddStdCatalog()
 {
-    // Try loading the message catalog for this version first, but fall back to
-    // the name without the version if it's not found, as message catalogs
-    // typically won't have the version in their names under non-Unix platforms
-    // (i.e. where they're not installed by our own "make install").
-    if ( AddCatalog("wxstd-" wxSTRINGIZE(wxMAJOR_VERSION) "." wxSTRINGIZE(wxMINOR_VERSION)) )
-        return true;
-
     if ( AddCatalog(wxS("wxstd")) )
         return true;
 


vadz:
Sorry, if I understand this correctly, this can't be right because it means `AddCatalog()` always returns true. If this is really the case, this must be fixed, but I have some trouble believing translations don't work at all and nobody has noticed it.

Can you reproduce the problem in the internat sample?

Blake-Madden:
The problem is with `AddStdCatalog()`, not `AddCatalog()`.  `AddCatalog()` will always return true if you use the default second argument of `wxLANGUAGE_ENGLISH_US` and the file you are looking for isn't found. (Which is what  `AddStdCatalog()` is doing).

If I rename my Spanish file to "wxstd-3.3.mo" then it loads fine. However, I change it to "wxstd.mo" and `AddStdCatalog()` will not find a Spanish "wxstd-3.3.mo" and then give up. (It should try to look for a Spanish "wxstd.mo" at that point).

The internat example is not using `AddStdCatalog()`, but instead is loading a hard coded "wxstd", so that is probably why it hasn't been noticed yet:

```
// Now try to add wxstd.mo so that loading "NOTEXIST.ING" file will produce
// a localized error message:
trans->AddCatalog("wxstd");
// NOTE: it's not an error if we couldn't find it!
```

Change it to `trans->AddStdCatalog()` and it will fail to load wxstd.mo.

vadz:
> The problem is with `AddStdCatalog()`, not `AddCatalog()`. `AddCatalog()` will always return true if you use the default second argument of `wxLANGUAGE_ENGLISH_US` and the file you are looking for isn't found.

The described behaviour of `AddCatalog()` is wrong. The second argument is not a fallback or anything like this, it's just the language of the message IDs themselves and should be English in 99% of cases.

I don't have time to debug it now, but if this is really how it behaves, it's completely broken :-(

lanurmi:
Likely related to #18227, especially as the issue was observed on Windows.

Blake-Madden:
> Likely related to #18227, especially as the issue was observed on Windows.

Stepping through the code, I think this is exactly what is happening.

vadz:
I'm closing this one because I still don't think it's the right solution, but we do need to solve this bug, of course.

PR #23889, opened 2023-09-20
Document more macOS-only wxTextCtrl methods

PBfordev:
Document previously undocumented macOS methods enabling/disabling automatic character substitutions.

This should have been part of 90e1769 (Add API to control OS X wxTextCtrl’s smart behavior, 2016-10-16).

Considering the three methods are mentioned in the changelog and documented in wxPython, I believe that them being undocumented before was not on purpose.

Should be backported to 3.2.

**EDIT:** I forgot to mention that I have literally zero knowledge of macOS; I was triggered to make the PR from conversations in wxPerl-related Issues and wrote the docs based on the wxWidgets code and Apple docs for `NSTextView` attributes.

### DIFF
--- a/interface/wx/textctrl.h
+++ b/interface/wx/textctrl.h
@@ -1768,7 +1768,7 @@ public:
     ///@{
 
     /**
-        Enable the automatic replacement of new lines characters in a
+        Enables the automatic replacement of new lines characters in a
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for doing this! I will merge this soon.

PR #23891, opened 2023-09-21
Final bunch of proposed backports for 3.2.3

vadz:


vadz:
> I know it's just the pictures in the docs but as I wrote in the PR, I think [the control screenshots updated for Windows 11](https://github.com/wxWidgets/wxWidgets/commit/d22275ce0a73def3320ed3c4284e184ebe34e15f) (mostly from Windows XP/7) should be also backported, as 3.2 is going to be here for many years and the ancient control look may not give best library impression (no idea how up-to-date non-MSW screenshots are).

Thanks for the reminder, I've cherry-picked it too now.

PBfordev:
I guess [Fix using `*.*` and similar patterns...](https://github.com/wxWidgets/wxWidgets/commit/a4d2e36cec16d511e461496f3eecbcb6f3e7343b ) is not a suitable candidate for backporting?

On one hand, it does introduce a subtle runtime change; on the other hand, it is still a bug fixed...

vadz:
> I guess [Fix using `*.*` and similar patterns...](https://github.com/wxWidgets/wxWidgets/commit/a4d2e36cec16d511e461496f3eecbcb6f3e7343b) is not a suitable candidate for backporting?

It is, sorry for forgetting about it and thanks for the reminder.
 
> On one hand, it does introduce a subtle runtime change; on the other hand, it is still a bug fixed...

Yes, I think the old behaviour was clearly wrong and new one is (more) correct, so I've pushed it too now. I'd prefer more testing to make sure that using `PathMatchSpec()` doesn't result in any issues, but I guess we'll learn about this soon after 3.2.3.

PR #23897, opened 2023-09-23
Optimize wxInputStream::SeekI() from start or current position

### LINKED ISSUE #23895: wxInputStream::SeekI does unnecessary Tell/Seek calls in common cases.
### Description
See this code in `wxInputStream::SeekI`:
https://github.com/wxWidgets/wxWidgets/blob/3b7d81b7f28353ed3b932fbcce958cdc81cc1992/src/common/stream.cpp#L963-L970

In most usage cases, `mode` is not `wxFromEnd`.

`size` value is only used if `mode == wxFromEnd`, but `size = GetLength()` is executed in all modes.

For file streams (a common use case), `GetLength` performs multiple Tell/Seek calls, reducing the performance.

vadz:
It's not clear whether it's worth explicitly checking for no-op seeks at
all, but unnecessarily calling GetLength() when doing it is certainly a
pessimization rather than optimization, so don't do it.

And don't bother checking for no-op seeks when seeking from end at all,
as it's not obviously faster than just performing the seek and the
existing code didn't work correctly anyhow, as it compared the current
position with "size-pos" instead of the correct "size+pos", so removing
it certainly shouldn't make things worse.

Closes #23895.

### DIFF
--- a/src/common/stream.cpp
+++ b/src/common/stream.cpp
@@ -962,11 +962,28 @@ wxFileOffset wxInputStream::SeekI(wxFileOffset pos, wxSeekMode mode)
         m_lasterror=wxSTREAM_NO_ERROR;
 
     // avoid unnecessary seek operations (optimization)
-    wxFileOffset currentPos = TellI(), size = GetLength();
-    if ((mode == wxFromStart && currentPos == pos) ||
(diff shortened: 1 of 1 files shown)


vadz:
Merged into master now.

PR #23900, opened 2023-09-24
Fix `wxGLCanvasEGL::IsShownOnScreen()` under Wayland

### LINKED ISSUE #23899: `wxGLCanvasEGL::IsShownOnScreen()` regression under Wayland
After thinking about this for a while I think I may have found a small oversight.
`IsShownOnScreen()` uses `m_readyToDraw` as one of the checks under Wayland. Before this PR, `m_readyToDraw` was set to `true` **once** after first `wl_frame_callback_handler`. After this PR now the `m_readyToDraw` is toggled after each frame.

`IsShownOnScreen()` probably shouldn't flap at 60hz?

I hacked it use `gtk_window_is_visible()` for `IsShownOnScreen` check and in my limited testing Kicad renders a bit better.

```
index c35273caddaa..0f009e3f4fb8 100644
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -697,12 +697,13 @@ bool wxGLCanvasEGL::SwapBuffers()
 bool wxGLCanvasEGL::IsShownOnScreen() const
 {
     wxDisplayInfo info = wxGetDisplayInfo();
+    GdkWindow* const window = GTKGetDrawingWindow();
     switch ( info.type )
     {
         case wxDisplayX11:
             return GetXWindow() && wxGLCanvasBase::IsShownOnScreen();
         case wxDisplayWayland:
-            return m_readyToDraw && wxGLCanvasBase::IsShownOnScreen();
+            return gtk_window_is_visible(window) && wxGLCanvasBase::IsShownOnScreen();
         default:
             return false;
     }
```

_Originally posted by @artizirk in https://github.com/wxWidgets/wxWidgets/issues/23554#issuecomment-1688358313_

vadz:
This should fix #23899.

@joanbm I'd be grateful for your review, as I've changed a check added by your recent commit which seemed misplaced to me, but I don't understand this (sub)surface stuff well enough to be really sure. TIA!

### DIFF
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -750,9 +750,11 @@ bool wxGLCanvasEGL::IsShownOnScreen() const
             return GetXWindow() && wxGLCanvasBase::IsShownOnScreen();
         case wxDisplayWayland:
             return m_readyToDraw && wxGLCanvasBase::IsShownOnScreen();
-        default:
-            return false;
+        case wxDisplayNone:
+            break;
     }
+
+    return false;
 }
 
 #endif // wxUSE_GLCANVAS && wxUSE_GLCANVAS_EGL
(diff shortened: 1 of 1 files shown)


artizirk:
I tested this PR and Kicad under native Wayland seems to work as well as it did with my hacky patch.

PR #23902, opened 2023-09-25
Fix background colour for empty cells in generic wxDataViewCtrl

### LINKED ISSUE #23708: bug: wxDataViewModel::GetAttr will change the color for the whole column, not child nodes only
Hi, in the wxDataViewCtrl sample code, especially in the first page, the "MyMusicTreeModel" model, I would like to change the color of the "child" nodes, not the "Container" nodes. I have add such code:

In the header file, I add this:

~~~~
virtual bool GetAttr( const wxDataViewItem&  item, unsigned int col, wxDataViewItemAttr& attr) const wxOVERRIDE;
~~~~

In the cpp file, I add this:
~~~~
bool MyMusicTreeModel::GetAttr( const wxDataViewItem&  item, unsigned int col, wxDataViewItemAttr& attr) const
{
    if (IsContainer(item))
        return false;

    if (col == 1)
    {
        attr.SetColour(*wxYELLOW);
        attr.SetBackgroundColour(*wxLIGHT_GREY);

        return true;
    }
    else
    {
        return false;
    }
}
~~~~

If you look at the code, I have a condition check that for a "container" like node, I just return "false".

From the document ([wxDataViewModel::GetAttr](https://docs.wxwidgets.org/stable/classwx_data_view_model.html#a28b85b9aca3a1ea803940817d86b0b73))return `false` means it just dose NOT change the color attributes of the input `wxDataViewItemAttr`.

But now, here is the result(see screen shot below)

image

You see that in the screen shot, the column "artist" pointed by the two arrows are expected to be white background, not the gray background.

My guess is that the wxDataViewItemAttr is reused when rendering. So, once the child node get gray color rendered, the parent node(the container node) also get rendered by the same color.

I can even change the code like below:

~~~~
bool MyMusicTreeModel::GetAttr( const wxDataViewItem&  item, unsigned int col, wxDataViewItemAttr& attr) const
{
    if (col == 1)
    {
        if (IsContainer(item))
        {
            attr.SetColour(*wxYELLOW);
            attr.SetBackgroundColour(*wxWHITE);
            return true;
        }
        else
        {
            attr.SetColour(*wxYELLOW);
            attr.SetBackgroundColour(*wxLIGHT_GREY);
            return true;
        }
    }
    else
    {
        return false;
    }
}
~~~~

Which means either the container node or child nodes will have different colors, but the result is still the same, the whole column will using the same background gray color.

Any ideas?

Thanks.

vadz:
There were two related problems: first, any attempts to set the background colour for the cells without values were simply ignored because we didn't call wxDataViewModel::GetAttr() at all from PrepareForItem() in this case and, second, PrepareForItem() itself was not called when repainting the control neither, resulting in the existing attribute being reused for the items without values, meaning that they used the last background colour that was set instead of at least not using any background at all.

Fix both of the problems for the generic version. Unfortunately the GTK one still doesn't allow setting the background for the empty cells, but at least when not doing it the code now consistently leaves them without any background.

The following code can be added to MyMusicTreeModel in the dataview sample for testing this:

```cpp
    bool GetAttr( const wxDataViewItem& item, unsigned int col, wxDataViewItemAttr& attr) const override
    {
        if (col != 1)
            return false;

        if (IsContainer(item))
        {
            if (item != m_pop)
                return false;

            attr.SetBackgroundColour(*wxYELLOW);
        }
        else
        {
            attr.SetColour(*wxYELLOW);
            attr.SetBackgroundColour(*wxLIGHT_GREY);
        }

        return true;
    }
```

Closes #23708.

@asmwarrior @jmoraleda Could you please test whether this works for you and, more importantly, if it doesn't break something else?

### DIFF
--- a/src/common/datavcmn.cpp
+++ b/src/common/datavcmn.cpp
@@ -902,12 +902,13 @@ wxDataViewRendererBase::PrepareForItem(const wxDataViewModel *model,
         }
 
         SetValue(value);
-
-        // Also set up the attributes for this item if it's not empty.
-        wxDataViewItemAttr attr;
-        model->GetAttr(item, column, attr);
-        SetAttr(attr);
     }
+
+    // Also set up the attributes: note that we need to do this even for the
+    // empty cells because background colour is still relevant for them.
+    wxDataViewItemAttr attr;
+    model->GetAttr(item, column, attr);
+    SetAttr(attr);
 
     // Finally determine the enabled/disabled state and apply it, even to the
     // empty cells.
--- a/src/generic/datavgen.cpp
+++ b/src/generic/datavgen.cpp
@@ -2835,7 +2835,6 @@ void wxDataViewMainWindow::OnPaint( wxPaintEvent &WXUNUSED(event) )
             wxDataViewTreeNode *node = nullptr;
             wxDataViewItem dataitem;
             const int line_height = GetLineHeight(item);
-            bool hasValue = true;
 
             if (!IsVirtualList())
             {
@@ -2847,10 +2846,6 @@ void wxDataViewMainWindow::OnPaint( wxPaintEvent &WXUNUSED(event) )
                 }
 
                 dataitem = node->GetItem();
-
-                if ( !model->HasValue(dataitem, col->GetModelColumn()) )
-                    hasValue = false;
-
             }
             else
             {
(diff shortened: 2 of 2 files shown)


asmwarrior:
OK, Thanks for the fix.
I will test it. I think currently, I will extract a patch file from this pull request, and apply the patch in my local wx 3.2.2.1 release source code, and rebuild the wx source.

asmwarrior:
> OK, Thanks for the fix. I will test it. I think currently, I will extract a patch file from this pull request, and apply the patch in my local wx 3.2.2.1 release source code, and rebuild the wx source.

Hi, I just did the above mentioned steps, and I can confirm that your pull request fixes the reported issue #23708.

Thanks!

vadz:
Merged into master now.

Issue #23894, opened 2023-09-22
wxComboCtrl and its derivatives are broken under wxQt


AliKet:
### Description

After this commit 7ebb20fb04d9c975a82e71c59bdb75f982aa90b8 popup window can no longer be shown and is dismissed immediately if you try to show the dropdown window by clicking the arrow button. And can be seen in the combo sample.

### Platform and version information

- wxWidgets version  you use: lateset
- wxWidgets port  you use: wxQt
- OS  and its version: Ubuntu 18.04

vadz:
Sorry for breaking this.

If this worked before, it should be possible to implement a generic `wxPopupWindow` using the code from the `USE_TRANSIENT_POPUP` case removed by that commit working at least as well as it used to. If you could please try doing this, it would be great. TIA!

AliKet:
I'll try to fix this in a couple of days!

AliKet:
Fixed in #23904

PR #23904, opened 2023-09-26
Fix wxPopupWindow implementation under wxQt

AliKet:
also, issue #23894 should be fixed now by this PR


src/qt/popupwin.cpp
```diff
@@ -9,12 +9,56 @@
...
 
+bool wxPopupWindow::Create( wxWindow *WXUNUSED(parent), int style )
+{
+    m_qtWindow = new wxQtPopupWindow(nullptr, this);
+
+    m_qtWindow->setWindowFlag(Qt::Popup);
+    m_qtWindow->setFocusPolicy(Qt::NoFocus);
+
+    m_qtWindow->setMinimumSize(100, 100);
+
```
AliKet, line 53: Notice that we don't have to set the minimum size explicitly here. And it's here just temporarily because it appears that there is a bug in `wxQt` reporting a size of (0, 0) or (1, 1) for a `wxPanel` which prevent the popup from appearing! I'll try to debug this further...

AliKet: I think I found the solution which seems to work correctly so far!
but I need to test it more thoroughly before I can post a PR for it...

AliKet: Fixed in #23912

vadz:
Great, thanks, will merge in a moment.

PR #23909, opened 2023-09-27
Fix setting EGL swap interval under native Wayland too

vadz:
See #23512.

### DIFF
--- a/include/wx/unix/glegl.h
+++ b/include/wx/unix/glegl.h
@@ -122,7 +122,9 @@ public:
     // is responsible for freeing the pointer
     static EGLConfig *InitConfig(const wxGLAttributes& dispAttrs);
 
-    bool m_readyToDraw;
+    // Only called when using Wayland to indicate that we should be redrawn.
+    void OnWLFrameCallback();
+
     wl_compositor *m_wlCompositor;
     wl_subcompositor *m_wlSubcompositor;
     wl_callback *m_wlFrameCallbackHandler;
@@ -138,6 +140,8 @@ private:
     wl_surface *m_wlSurface;
     wl_region *m_wlRegion;
     wl_subsurface *m_wlSubsurface;
+
+    bool m_readyToDraw;
 
     // the global/default versions of the above
     static EGLConfig *ms_glEGLConfig;
--- a/include/wx/unix/glegl.h
+++ b/include/wx/unix/glegl.h
@@ -142,6 +142,7 @@ private:
     wl_subsurface *m_wlSubsurface;
 
     bool m_readyToDraw;
+    bool m_swapIntervalSet = false;
 
     // the global/default versions of the above
     static EGLConfig *ms_glEGLConfig;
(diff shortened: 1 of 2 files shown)


joanbm:
There's a problem using `eglSwapInterval` here because it depends on the currently selected EGL context (from `wxGLContext::SetCurrent` / `eglMakeCurrent`) and there's no guarantee that the appropriate context is bound inside `wxGLCanvasEGL::OnWLFrameCallback()`. This may be a problem if the application is using multiple OpenGL contexts (not as strange as one may think since using e.g. `wxWebView` creates a context).

The failure you observed in https://github.com/wxWidgets/wxWidgets/issues/23512#issuecomment-1737294996 is probably due to the same reason, the right context was not selected.

This should probably be moved inside `wxGLContext::SwapBuffers` (like in #23898) since we can be sure that the right context is selected there.


The call to `eglSwapInterval`

vadz:
> There's a problem using `eglSwapInterval` here because it depends on the currently selected EGL context (from `wxGLContext::SetCurrent` / `eglMakeCurrent`) and there's no guarantee that the appropriate context is bound inside `wxGLCanvasEGL::OnWLFrameCallback`.

Oops. 

> The failure you observed in [#23512 (comment)](https://github.com/wxWidgets/wxWidgets/issues/23512#issuecomment-1737294996) is probably due to the same reason, the right context was not selected.

If so, it seems like we ought to do this in `wxGLContext::SetCurrent()`? But I am not sure that it fails because the right context isn't selected: why would it fail for this other context? I.e. I'd understand if it didn't work, in the sense that it didn't change the swap interval, but the function actually failed, with an error.

> This should probably be moved inside `wxGLContext::SwapBuffers` (like in #23898) since we can be sure that the right context is selected there.

But this is what we had started with and it didn't work. I'm not sure if it's because the context wasn't current or some other reason, but I don't know how to make it work there. If you do, please let me know!

joanbm:
> > The failure you observed in https://github.com/wxWidgets/wxWidgets/issues/23512#issuecomment-1737294996 is probably due to the same reason, the right context was not selected.
> 
> If so, it seems like we ought to do this in wxGLContext::SetCurrent()? But I am not sure that it fails because the right context isn't selected: why would it fail for this other context? I.e. I'd understand if it didn't work, in the sense that it didn't change the swap interval, but the function actually failed, with an error.
> 
> > This should probably be moved inside wxGLContext::SwapBuffers (like in https://github.com/wxWidgets/wxWidgets/pull/23898) since we can be sure that the right context is selected there.
> 
> But this is what we had started with and it didn't work. I'm not sure if it's because the context wasn't current or some other reason, but I don't know how to make it work there. If you do, please let me know!

As far as I can see, on Wayland, we never called `eglSwapInterval` from `wxGLContext::SwapBuffers`, but rather from `wxGLCanvasEGL::CreateSurface`. Calling `eglSwapInterval` from there has be same problem, there's no guarantee that the right EGL context is current. In particular, for simple applications, if you try to call `eglSwapInterval` from `wxGLCanvasEGL::CreateSurface` it will fail with error `EGL_BAD_CONTEXT` because no context has been set yet.

Calling `eglSwapInterval` from either `wxGLContext::SetCurrent` and `wxGLContext::SwapBuffers` should work.

vadz:
I thought I tried calling from `SwapBuffers()` too but maybe you're right and I didn't. Anyhow, it's not difficult to call it from there and it looks preferable to call it from the same place for X11 and Wayland cases anyhow, so I've done it now in #24014.

PR #23910, opened 2023-09-27
Fix using *.* in wxDir under MSW

### LINKED ISSUE #23905: Surprising file name matching of *.* under wxMSW
On Windows, the users are generally used to `*.*` matching all the files, including those without an extension. This can be seen for example with _dir_ in the windows command line or Common File Dialogs (or `wxFileDialog` using `*.*`); same when using Win32 `FindFirstFile()`/`FindNextFile()` API with `*.*`, where files without an extension are returned. 

However, with wxWidgets, files without an extension are not matched with `*.*` (probably because they do not actually have the dot in the name, they are matched with `*`), which can be seen in the unmodified widgets sample, for both FileCtrl and DirCtrl. For example, there are files without an extension in the wxWidgets root folder (e.g., _configure_ and _mkinstalldirs_). `wxMatchWild()` and `wxDir::GetAllFiles()` also behave the same.

Such behavior can be rather unexpected when one wants to get all the files in the folder, not to mention that sometimes the wild card pattern comes from the user. 

Additionally, [the description](https://docs.wxwidgets.org/latest/filedlg_8h.html#a3a48d9b7ad35e6c332acb146f78c1b5d) for `wxFileSelectorDefaultWildcardStr` used by `wxFileCtrl` and `wxFileDialog` has:
> Default wildcard string used in wxFileDialog corresponding to all files. 
> It is defined as `"*.*"` under MSW and `"*"` everywhere else. 

Funnily enough, Microsoft's `FindFirstFile()`/`FindNextFile()` [example](https://learn.microsoft.com/en-us/windows/win32/fileio/listing-the-files-in-a-directory) uses just `*` as the pattern.

For completeness sake: Windows PowerShell’s dir etc. do NOT list files without an extension when called with `*.*` but unlike the default _dir_, they also do not list any folders with this mask.

To conclude, if this oddity (at least to an MSW-only person) cannot be fixed (difficult/incompatible/by design), perhaps it should be at least documented somewhere where there is a chance it can be easily spotted.

vadz:
Using "*.*" as a wildcard is supposed to match everything under MSW, but
ever since the changes of 4daceaacbd (Check that files returned from
wxDir::FindXXX() match the filter., 2013-04-08, see #3432) it only
matched the files with extension because we double-checked the match
returned by the native MSW function (which does handle "*.*" correctly)
using our own wxString::Matches() which doesn't handle it specially.

Fix this by skipping the call to Matches() when "*.*" is used: we know
that this wildcard matches everything, so rechecking the match would be
useless at best, even if it were not actively harmful. And also skip
this call for "*" because calling Matches("*") always succeeds anyhow.

This also fixes the same bug in wxFind{First,Next}File() and any other
code using them such as wx{File,Dir}Ctrl.

Closes #23905.

### DIFF
--- a/src/msw/dir.cpp
+++ b/src/msw/dir.cpp
@@ -28,6 +28,11 @@
 #include "wx/dir.h"
 
 #include "wx/msw/private.h"
+#include <shlwapi.h>
+
+#ifdef __VISUALC__
+    #pragma comment(lib, "shlwapi")
+#endif
 
 // ----------------------------------------------------------------------------
 // define the types and functions used for file searching
@@ -71,23 +76,7 @@ CheckFoundMatch(const FIND_STRUCT* finddata, const wxString& filter)
     if ( filter.empty() )
         return true;
 
-    // Otherwise do check the match validity. Notice that we must do it
-    // case-insensitively because the case of the file names is not supposed to
-    // matter under Windows.
-    wxString fn(finddata.cFileName);
-
-    // However if the filter contains only special characters (which is a
-    // common case), we can skip the case conversion.
-    if ( filter.find_first_not_of(wxS("*?.")) == wxString::npos )
-    {
-        // And maybe even skip the check entirely if we have a filter that we
-        // know matches everything. This is more than just an optimization, as
-        // "*.*" it wouldn't match the files without extension according to
-        // wxString::Matches(), but it should.
-        return filter == wxS("*.*") || filter == wxS("*") || fn.Matches(filter);
-    }
-
-    return fn.MakeUpper().Matches(filter.Upper());
+    return ::PathMatchSpecEx(finddata->cFileName, filter, PMSF_NORMAL) == S_OK;
 }
 
 inline bool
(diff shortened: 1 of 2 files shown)


src/msw/dir.cpp
```diff
@@ -86,7 +86,13 @@ CheckFoundMatch(const FIND_STRUCT* finddata, const wxString& filter)
     // However if the filter contains only special characters (which is a
     // common case), we can skip the case conversion.
     if ( filter.find_first_not_of(wxS("*?.")) == wxString::npos )
-        return fn.Matches(filter);
+    {
+        // And maybe even skip the check entirely if we have a filter that we
+        // know matches everything. This is more than just an optimization, as
+        // "*.*" it wouldn't match the files without extension according to
```
PBfordev, line 92:
```suggestion
        // "*.*" wouldn't match the files without extension according to
```

PBfordev:
BTW, what is the reason of using `PatchMatchSpecEx()` instead of `PatchMatchSpec()` here? I can't see any advantage of the former han and the latter can also be used in 3.2 (both are probably broken under Wine).

PR #23912, opened 2023-09-28
Fix wxWindow::SetClientSize() under wxQt

AliKet:
AFAICS, this does not introduce any regression. Moreover, this solves the problem mentioned [here](https://github.com/wxWidgets/wxWidgets/pull/23904#discussion_r1338711704) because the popup always has a size of (0, 0)


vadz:
Thanks, merging this also.

P.S. Not sure why does the second commit use "fixup", it might be a mistake, but I'll just merge it as is.

PR #23913, opened 2023-09-28
Fix invisible wxStaticLine on macOS >= 13

vslavik:
`NSBox` used as a separator must have its type set to `NSBoxSeparator`.

This apparently didn’t matter before in practice, but starting with macOS 13, separators - i.e. narrow NSBoxes - without the type correctly set are invisible.


vadz:
Thanks for fixing this! I'll merge it and backport to 3.2 if nobody has any objections.

PR #23914, opened 2023-09-28
Ensure that the currently shown wxSimplebook page has focus

vadz:
When calling wxSimplebook::ChangeSelection(), the focus remained on the
previous, now hidden, page, unless it was explicitly set to one of the
controls on the new page.

This was completely unexpected as it could result in the user
(inadvertently) changing the values of the already "accepted" controls
on the last page, so don't let this happen and always set the focus to
the new page explicitly, even if this hasn't been done by
ShowWithEffect() which doesn't do it at least under wxMSW.

### DIFF
--- a/include/wx/simplebook.h
+++ b/include/wx/simplebook.h
@@ -195,9 +195,17 @@ protected:
     virtual void DoShowPage(wxWindow* page, bool show) override
     {
         if ( show )
+        {
             page->ShowWithEffect(m_showEffect, m_showTimeout);
+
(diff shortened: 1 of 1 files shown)


vadz:
This is the least intrusive fix of the problem described. I also thought of doing it in `ShowWithEffect()` itself, but I think this might not be always desirable, so I didn't do it finally in this commit, as I plan to apply it to 3.2 too.

In master we should probably do it there, what do the others think?

PR #23917, opened 2023-09-29
Override Set{Min,Max}Size and DoSetSizeHints() to pass size informati…

AliKet:
…on to Qt

Setting min/max widget size at the wxWidgets level is not enough to take effect unless the information is also set at the Qt level.


AliKet:
Here is a comparison before and after these changes:

**Before:**
Screenshot from 2023-09-29 21-19-30

**After:**
Screenshot from 2023-09-29 21-16-31

The above screenshots are from the combo sample.

vadz:
Merging after fixing the spelling of the comment which triggered the CI check failure.

PR #23918, opened 2023-09-30
Fix `AddStdCatalog()` under MSW and add functions without msg ID fallback

vadz:
This replaces #23886 and somewhat addresses #18227: it doesn't change `AddCatalog()` but adds a new function which behaves as described there.

@vslavik Your review would be highly appreciated because I'm really not sure if I'm doing the right thing here, but, as I explained in #18227, I don't see any other reasonable solution. TIA!

### DIFF
--- a/include/wx/translation.h
+++ b/include/wx/translation.h
@@ -151,16 +151,24 @@ public:
     // get languages available for this app
     wxArrayString GetAvailableTranslations(const wxString& domain) const;
 
-    // find best translation language for given domain
+    // find best available translation language for given domain
+    wxString GetBestAvailableTranslation(const wxString& domain);
+
     wxString GetBestTranslation(const wxString& domain, wxLanguage msgIdLanguage);
     wxString GetBestTranslation(const wxString& domain,
                                 const wxString& msgIdLanguage = wxASCII_STR("en"));
+
+    // add catalog for the given domain returning true if it could be found by
+    // wxTranslationsLoader
+    bool AddAvailableCatalog(const wxString& domain);
 
     // add standard wxWidgets catalog ("wxstd")
     bool AddStdCatalog();
 
     // add catalog with given domain name and language, looking it up via
-    // wxTranslationsLoader
+    // wxTranslationsLoader -- unlike AddAvailableCatalog(), this function also
+    // returns true if this catalog is not needed at all because msgIdLanguage
+    // is an acceptable language to use directly
     bool AddCatalog(const wxString& domain,
                     wxLanguage msgIdLanguage = wxLANGUAGE_ENGLISH_US);
 
@@ -186,7 +194,7 @@ public:
 
 private:
     // perform loading of the catalog via m_loader
-    bool LoadCatalog(const wxString& domain, const wxString& lang, const wxString& msgIdLang);
+    bool LoadCatalog(const wxString& domain, const wxString& lang);
 
     // find catalog by name in a linked list, return nullptr if !found
     wxMsgCatalog *FindCatalog(const wxString& domain) const;
--- a/interface/wx/intl.h
+++ b/interface/wx/intl.h
@@ -390,7 +390,10 @@ public:
     bool AddCatalog(const wxString& domain, wxLanguage msgIdLanguage);
 
     /**
-        Calls wxTranslations::AddCatalog(const wxString&, wxLanguage, const wxString&).
+        Calls wxTranslations::AddCatalog(const wxString&, const wxString&).
+
+        This overload shouldn't be used any longer as @a msgIdCharset is
+        just ignored.
     */
     bool AddCatalog(const wxString& domain, wxLanguage msgIdLanguage,
                     const wxString& msgIdCharset);
--- a/interface/wx/translation.h
+++ b/interface/wx/translation.h
@@ -180,34 +180,6 @@ public:
                     wxLanguage msgIdLanguage = wxLANGUAGE_ENGLISH_US);
 
     /**
-        Same as AddCatalog(const wxString&, wxLanguage), but takes two
-        additional arguments, @a msgIdLanguage and @a msgIdCharset.
-
-        This overload is only available in non-Unicode build.
-
-        @param domain
-            The catalog domain to add.
-
-        @param msgIdLanguage
-            Specifies the language of "msgid" strings in source code
-            (i.e. arguments to GetString(), wxGetTranslation() and the _() macro).
-            It is used if AddCatalog() cannot find any catalog for current language:
-            if the language is same as source code language, then strings from source
-            code are used instead.
-
-        @param msgIdCharset
-            Lets you specify the charset used for msgids in sources
-            in case they use 8-bit characters (e.g. German or French strings).
-
-        @return
-            @true if catalog was successfully loaded, @false otherwise (which might
-            mean that the catalog is not found or that it isn't in the correct format).
-     */
-    bool AddCatalog(const wxString& domain,
-                    wxLanguage msgIdLanguage,
-                    const wxString& msgIdCharset);
-
-    /**
         Check if the given catalog is loaded, and returns @true if it is.
 
         According to GNU gettext tradition, each catalog normally corresponds to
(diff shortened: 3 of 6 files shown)


interface/wx/intl.h
```diff
@@ -390,7 +390,10 @@ class wxLocale
     bool AddCatalog(const wxString& domain, wxLanguage msgIdLanguage);
 
     /**
-        Calls wxTranslations::AddCatalog(const wxString&, wxLanguage, const wxString&).
+        Calls wxTranslations::AddCatalog(const wxString&, const wxString&).
+
+        This overload shouldn't be used any longer as @a msgIdCharset is
+        just ignored.
```
vslavik, line 396: Shouldn't it also be deprecated, then, to avoid confusion?

interface/wx/translation.h
```diff
@@ -124,21 +143,38 @@ class wxTranslations
...
-        By default, it is searched for in standard places (see
+        By default, the catalog is searched for in standard places (see
         wxFileTranslationsLoader), but you may also prepend additional
         directories to the search path with
         wxFileTranslationsLoader::AddCatalogLookupPathPrefix().
 
         All loaded catalogs will be used for message lookup by GetString() for
         the current locale.
 
+        @return
```
vslavik, line 161: I'd explicitly mention the msgIdLanguage/en situation here too - i.e. note that it is normal with some languages (typically English) that there's no catalog.

vslavik (commented):
Seems reasonable to me, my only comments are nitpicky ones.

On that note, there's a typo in 2a93404e64e00e67bd075c636dbce37c96947f5d's commit message: "ditinguish".

> It is still not clear if it is really correct to return "en-US" from the list of preferred languages even if the user has never intentionally configured the OS to indicate that English is acceptable

FWIW, I think it isn't, but I'm not sure if we can distinguish it from the intentional case on MSW?

interface/wx/intl.h
vadz: I've hesitated to do it, as it does no harm to keep it, but done now, hopefully it's not used much anyhow.

include/wx/translation.h
```diff
@@ -151,16 +151,24 @@ class WXDLLIMPEXP_BASE wxTranslations
...
+    // find best available translation language for given domain
+    wxString GetBestAvailableTranslation(const wxString& domain);
+
     wxString GetBestTranslation(const wxString& domain, wxLanguage msgIdLanguage);
     wxString GetBestTranslation(const wxString& domain,
                                 const wxString& msgIdLanguage = wxASCII_STR("en"));
 
+    // add catalog for the given domain returning true if it could be found by
+    // wxTranslationsLoader
+    bool AddAvailableCatalog(const wxString& domain);
```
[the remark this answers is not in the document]
vadz: Yes, as you can probably guess, I spent a long time agonizing over the name, but I couldn't find anything better, so I just went ahead with "Available" because it was already used for `GetAvailableTranslations()` and it seemed better to avoid introducing yet another term.

interface/wx/translation.h
```diff
@@ -89,6 +89,18 @@ class wxTranslations
...
+        Returns the best available translation for the required language.
+
+        For wxLANGUAGE_DEFAULT, this function returns the available translation
+        best matching one of wxUILocale::GetPreferredUILanguages(). Otherwise
+        it simply returns the language set with SetLanguage() if it's available
+        or empty string otherwise.
+
+        @since 3.3.0
+     */
+    wxString GetBestAvailableTranslation(const wxString& domain);
```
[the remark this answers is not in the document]
vadz: Sorry, I think "Loaded" would be wrong -- it's not loaded yet. What we need here is "something that can be loaded" -- which is pretty close to "available", I think.

IMO English is available as UI language, but not as a translation and I'm looking to clearly distinguish between the two.

vadz:
> Seems reasonable to me, my only comments are nitpicky ones.

Thanks, I'm going to commit it then even if I can't say that I'm really happy with it. But I'd like to include this in 3.2.3 to at least fix the problem from #23886 there. And maybe we can improve things in master later. E.g. I still think we might want to add

```cpp
enum { NotNeeded, Loaded, NotFound } InitializeFor(wxString domain, wxString lang = "en-US");
```

(better names for everything are welcome as always!).

> On that note, there's a typo in [2a93404](https://github.com/wxWidgets/wxWidgets/commit/2a93404e64e00e67bd075c636dbce37c96947f5d)'s commit message: "ditinguish".

Oops, thanks, fixed.
 
> > It is still not clear if it is really correct to return "en-US" from the list of preferred languages even if the user has never intentionally configured the OS to indicate that English is acceptable
> 
> FWIW, I think it isn't, but I'm not sure if we can distinguish it from the intentional case on MSW?

I don't see how :-( AFAICS there is nothing to indicate that it was added there by default.

But, in theory, the problem is not limited to this case: even if you intentionally set up your preferred languages to Czech and English, you would presumably still prefer to use wx in Czech if `cs/wxstd.mo` is available, but currently it would never be loaded.

vadz:
I'm really sorry, I've accidentally pushed the branch to a wrong place and then deleted it from a wrong fork and now it looks like I can't reopen this PR and so I had to create #23930 instead.

vslavik:
I'm returning to this because of a breakage neither of us considered. This seems like the best place to do it because the relevant context is right here.

> > > FWIW, I think it isn't, but I'm not sure if we can distinguish it from the intentional case on MSW?
> >
> > I don't see how :-( AFAICS there is nothing to indicate that it was added there by default.

Perhaps "if en-US is at the end of the list, ignore it, otherwise respect it" would be a usable MSW-only heuristic?

> But, in theory, the problem is not limited to this case: even if you intentionally set up your preferred languages to Czech and English, you would presumably still prefer to use wx in Czech if cs/wxstd.mo is available, but currently it would never be loaded.

It turns out we now, with these changes included, have the opposite problem: Czech translation is loaded even when I want English to be used. Specifically, if the language preference list is e.g. `[en-US, en, cs]`, i.e. English preferred, then this preference is ignored and the first non-English language is picked.

Another example: consider `[cs, en, fr]` preference. I strongly prefer English (which I speak) over French (which I sometimes semi-understand). If the application doesn't have a Czech translation, but has French one (very common), French is what I'll get, not English. So the fix is not even as simple as checking for English at the beginning of the list...

That's I think the answer to the question of why English was implicitly included over at https://github.com/wxWidgets/wxWidgets/issues/18227#issuecomment-1741837877 I think English really needs to be considered a properly available "translation" in the logic. It just so happens that it is not stored in a MO file, but in implicit storage in the code.

Notice also that backwards-compatible `wxTranslations::GetBestTranslation()` does this too (and so isn't compatible) -- it first considers translations and only falls back to English if that comes up empty.

I'll try to have another stab at this. Preliminarily, I think `wxstd` and `wxstd-3.2` need to be evaluated/loaded together, not sequentially, and the lookup change from here / #23930 reverted.

vadz:
> Perhaps "if en-US is at the end of the list, ignore it, otherwise respect it" would be a usable MSW-only heuristic?

Can we be sure only `en-US` gets implicitly added like this? I didn't think we could, but if we can, then, yes, this should work.

> It turns out we now, with these changes included, have the opposite problem: Czech translation is loaded even when I want English to be used.

I see... I just hadn't considered the situation in which we may prefer not to load the translations because message ID language is higher in the preferred languages list than the translations language. At least I understand the intent of the original code now, sorry for having missed it before.

> I'll try to have another stab at this.

TIA!

Just to be sure: this doesn't affect 3.2.4 release (which just changes the return value of `AddCatalog()`), does it?

vslavik:
> Just to be sure: this doesn't affect 3.2.4 release (which just changes the return value of AddCatalog()), does it?

I noticed it on the 3.2 branch, where it's present since 3.2.3's d1782237 backport.

vadz:
Yes, sorry, I should have been more clear: I realize it's also the case for 3.2 branch, what I meant was just that the change in 3.2.4 doesn't make it worse.

vslavik:
> what I meant was just that the change in 3.2.4 doesn't make it worse.

I see; right, 4fab20cc has no effect (AFAICT).

PR #23919, opened 2023-09-30
Fix return type of wxBitmap::UseAlpha() under wxMSW

### LINKED ISSUE #23903: Incorrect return type for wxBitmap::UseAlpha() in wxMSW
`wxBitmap::UseAlpha()` is [documented](https://docs.wxwidgets.org/latest/classwx_bitmap.html#a34e03148a76b66cea44a41155ba66618) to return `bool` and the base class does that:
https://github.com/wxWidgets/wxWidgets/blob/df6366ff57371f9d2d41f748514d25259e393203/include/wx/bitmap.h#L255
and so does the OSX one.

However, wxMSW does not (unlike in OSX, in MSW `wxBitmap` does not derive from `wxBitmapBase`):
https://github.com/wxWidgets/wxWidgets/blob/df6366ff57371f9d2d41f748514d25259e393203/include/wx/msw/bitmap.h#L180

Not sure how to best fix it, probably something like this should suffice
```diff
diff --git a/include/wx/msw/bitmap.h b/include/wx/msw/bitmap.h
index 9293963d34..2e3f278260 100644
--- a/include/wx/msw/bitmap.h
+++ b/include/wx/msw/bitmap.h
@@ -177,7 +177,7 @@ public:
     void SetMask(wxMask *mask);
 
     bool HasAlpha() const;
-    void UseAlpha(bool use = true);
+    bool UseAlpha(bool use = true);
     void ResetAlpha() { UseAlpha(false); }
 
     // old synonyms for CreateWithDIPSize() and GetLogicalXXX() functions
 src/msw/bitmap.cpp | 13 +++++--------
 1 file changed, 5 insertions(+), 8 deletions(-)

diff --git a/src/msw/bitmap.cpp b/src/msw/bitmap.cpp
index 80b650087c..24ac83099c 100644
--- a/src/msw/bitmap.cpp
+++ b/src/msw/bitmap.cpp
@@ -1222,16 +1222,13 @@ wxDC *wxBitmap::GetSelectedInto() const
 #endif
 }
 
-void wxBitmap::UseAlpha(bool use)
+bool wxBitmap::UseAlpha(bool use)
 {
-    if ( GetBitmapData() )
-    {
-        // Only 32bpp bitmaps can contain alpha channel.
-        if ( use && GetBitmapData()->m_depth < 32 )
-            use = false;
+    if ( !GetBitmapData() || (use && GetBitmapData()->m_depth < 32) )
+        return false;
 
-        GetBitmapData()->m_hasAlpha = use;
-    }
+    GetBitmapData()->m_hasAlpha = use;
+    return true;
 }
 
 bool wxBitmap::HasAlpha() const
```

PBfordev:
The method should return bool but was returning void.

Closes #23903.

### DIFF
--- a/include/wx/msw/bitmap.h
+++ b/include/wx/msw/bitmap.h
@@ -177,7 +177,7 @@ public:
     void SetMask(wxMask *mask);
 
     bool HasAlpha() const;
-    void UseAlpha(bool use = true);
+    bool UseAlpha(bool use = true);
     void ResetAlpha() { UseAlpha(false); }
 
     // old synonyms for CreateWithDIPSize() and GetLogicalXXX() functions
--- a/src/msw/bitmap.cpp
+++ b/src/msw/bitmap.cpp
@@ -1222,16 +1222,13 @@ wxDC *wxBitmap::GetSelectedInto() const
 #endif
 }
 
-void wxBitmap::UseAlpha(bool use)
+bool wxBitmap::UseAlpha(bool use)
 {
-    if ( GetBitmapData() )
-    {
-        // Only 32bpp bitmaps can contain alpha channel.
-        if ( use && GetBitmapData()->m_depth < 32 )
-            use = false;
+    if ( !GetBitmapData() || (use && GetBitmapData()->m_depth < 32) )
+        return false;
 
-        GetBitmapData()->m_hasAlpha = use;
-    }
+    GetBitmapData()->m_hasAlpha = use;
+    return true;
 }
 
 bool wxBitmap::HasAlpha() const


include/wx/msw/bitmap.h
```diff
@@ -177,7 +177,7 @@ class WXDLLIMPEXP_CORE wxBitmap : public wxGDIImage,
     void SetMask(wxMask *mask);
 
     bool HasAlpha() const;
-    void UseAlpha(bool use = true);
+    bool UseAlpha(bool use = true);
     void ResetAlpha() { UseAlpha(false); }
```
AliKet, line 181: `override` can be added then!

vadz: No because `wxBitmap` still doesn't inherit from `wxBitmapBase` in wxMSW, unfortunately.

src/msw/bitmap.cpp
```diff
@@ -1222,16 +1222,13 @@ wxDC *wxBitmap::GetSelectedInto() const
 #endif
 }
 
-void wxBitmap::UseAlpha(bool use)
+bool wxBitmap::UseAlpha(bool use)
 {
-    if ( GetBitmapData() )
-    {
-        // Only 32bpp bitmaps can contain alpha channel.
```
vadz, line 1229: I think this comment was useful, so I'm going to restore it when merging.

PR #23920, opened 2023-09-30
Fix a crash in wxAUI

AliKet:
`wxAUI` has a special window "`wxTabFrame`" for which `GetHandle()` returns `nullptr`. This is a problem because this commit 3c68696 (Make wxTopLevelWindow::SetSizeHints() work in wxQt) doesn't account for the rare situation in which a `wxWindow` is actually created without the underlying handle, resulting in a crash when we try to call `Set{Min,Max}Size()` for this type of windows.

### DIFF
--- a/src/qt/window.cpp
+++ b/src/qt/window.cpp
@@ -488,6 +488,8 @@ void wxWindowQt::SetFocus()
 
 void wxWindowQt::QtSetMinSize(const wxSize& minSize)
 {
+    if ( !GetHandle() ) return;
+
     if ( minSize.x >= 0 )
         GetHandle()->setMinimumWidth(minSize.x);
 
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, I'll apply this (with just minor reformatting) but I wonder if we shouldn't be setting min/max size when creating the Qt window if it happens after setting it at wx level?

AliKet:
> Thanks, I'll apply this (with just minor reformatting) but I wonder if we shouldn't be setting min/max size when creating the Qt window if it happens after setting it at wx level?

Normally before I publish any PR implementing or correcting something I try to:
1- read the documentation carefully to know the intended behaviour.
2- run the tests locally for fear to break something else inadvertently.
3- In doubt, compare with the other ports.

I somehow missed to do step 2 this time and I have to revert some recent changes here #23925 (in the first two commits) to correct the divergence from documentation and the already established tests.

I'm so sorry for the inconvenience I've made!

PR #23922, opened 2023-10-01
Modernize wxGenericDirCtrl appearance on MSW

PBfordev:
Change wxTreeCtrl style for a control used by wxGenericDirCtrl so that wxGenericDirCtrl looks
more like the folder view in File Explorer (but still don't show expanders for root-level items).

Also update wxGenericDirCtrl screenshot in the docs.

See also  6caaf58 (Use more modern appearance for wxMSW wxTreeCtrl by default, 2023-09-30).

This change was already [discussed](https://github.com/wxWidgets/wxWidgets/pull/23844#issuecomment-1704375512), I opted for the "safe" MSW-only change (instead of changing the default style on all platforms and modifying that).


vadz:
Thanks, merging now with some very minor reformatting.

PR #23923, opened 2023-10-01
Fixes for suppressing bogus GTK diagnostics

vadz:
This got completely broken for 2 different reasons in master, fix this now.

The first commit here also needs to be backported to 3.2.


vadz:
I've also added `WXSUPPRESS_GTK_DIAGNOSTICS` to this PR as this is just too convenient not to have it. Please let me know if anybody has any objections as I'd like to backport this to 3.2 too.

PR #23925, opened 2023-10-01
Preparations to enable automatic tests for wxQt

AliKet:


AliKet:
I am already aware of the reported failures and will be corrected in the upcoming commits

AliKet:
@vadz This [05bf4ee](https://github.com/wxWidgets/wxWidgets/pull/23925/commits/05bf4ee6043f98157cf2e6e7a8e587d3ab206fd2)  is just a temporary commit to show that the generic `wxPalette` implementation works perfectly under `wxQt`. I need bakefile 0.2.13 to make a real commit. I tried to build it from source but no luck sorry!

vadz:
> @vadz This [05bf4ee](https://github.com/wxWidgets/wxWidgets/pull/23925/commits/05bf4ee6043f98157cf2e6e7a8e587d3ab206fd2) is just a temporary commit to show that the generic `wxPalette` implementation works perfectly under `wxQt`.

So should this commit be reverted before merging?

AliKet:
> > @vadz This [05bf4ee](https://github.com/wxWidgets/wxWidgets/pull/23925/commits/05bf4ee6043f98157cf2e6e7a8e587d3ab206fd2) is just a temporary commit to show that the generic `wxPalette` implementation works perfectly under `wxQt`.
> 
> So should this commit be reverted before merging?

I managed to build bakefile 0.2.13 from source (after manually disabling documentation building which was problematic for me) and pushed the updated commit. You can merge this after GitHub CI completed. TIA!

AliKet:
@vadz BTW do you know how to enable these tests under AppVeyor ? what the instructions needed to do that ? TIA.

vadz:
This seems to be determined by https://github.com/wxWidgets/wxWidgets/blob/368a13a1a2b0d8f9a7c31935cb22dc1537e86d0b/build/tools/appveyor-test.bat#L50 so it looks like it might actually be enabled? Anyhow, this is what is being used by `appveyor.yml`.

interface/wx/spinctrl.h
```diff
@@ -136,6 +136,8 @@ class wxSpinCtrl : public wxControl
     /**
         Returns the text in the text entry part of the control.
 
+        @note Always returns non-empty string under wxQt.
```
vadz, line 139: Interesting, so what does it return if the text control is actually empty?

interface/wx/spinctrl.h
```diff
@@ -189,6 +191,8 @@ class wxSpinCtrl : public wxControl
 
         @note Setting a range including negative values is silently ignored
         if current base is set to 16.
+
+        @note Backwards ranges are not supported under wxQt.
```
vadz, line 195: I'd make it more prescriptive:
```suggestion
        @note In wxQt @a minVal must be less than @a maxVal.
```

src/qt/textentry.cpp
```diff
@@ -104,11 +105,20 @@ wxString wxTextEntry::DoGetValue() const
     return wxString();
 }
 
-void wxTextEntry::DoSetValue(const wxString &WXUNUSED(value), int WXUNUSED(flags))
+void wxTextEntry::DoSetValue(const wxString& value, int flags)
 {
+    wxTextEntryBase::DoSetValue(value, flags);
 }
```
vadz, line 111: Maybe we should just stop overriding it? Or do you plan to do something else than calling the base class version here in the (near) future?

src/qt/window.cpp
```diff
@@ -350,6 +350,8 @@ bool wxWindowQt::Create( wxWindowQt * parent, wxWindowID id, const wxPoint & pos
     // that a generic control, like wxPanel, is being created, so we need a very
     // simple control as a base:
 
+    wxSize size_ = size;
```
vadz, line 353: Minor, but `size_` is not very readable. Let's rename it to `initialSize` maybe?

tests/controls/listboxtest.cpp
```diff
@@ -109,12 +109,21 @@ void ListBoxTestCase::Sort()
 
     m_list->Append(testitems);
 
+#ifndef __WXQT__
```
vadz, line 112: The different sort order should ideally be documented (I assume we can't change it).

vadz (commented):
Great, thanks a lot for all your work here!

There are a couple of possible minor improvements/clarifications but otherwise this is ready to be merged AFAICS.

Thanks again!

tests/controls/listboxtest.cpp
AliKet: we can change the default sort order by overriding the  `operator <` :
```
    virtual bool wxQtListWidgetItem::operator<(const QListWidgetItem& other) const override
    {
        const auto str1 = text();
        const auto str2 = other.text();
        return str1.compare(str2, Qt::CaseInsensitive) < 0;
    }
```
Sure it will make the test happy then (as it passes successfully with it) but how about the end user ? would the list be sorted the way he/she wants/expects ? because the above code doesn't take into account the (locale-specific) rules on how to compare  strings!
IMHO `wxQt` and `wxGTK` (before this commit https://github.com/wxWidgets/wxWidgets/commit/625ed743015f2da84db47a6851e6011bd08f4bd2) looks more correct to me because they respect the `LC_COLLATE` setting of the location.

src/qt/textentry.cpp
AliKet: No I don't have any plan about this function. at least for now. So feel free to remove this override when you merge this PR. Thanks!

include/wx/qt/datectrl.h
```diff
@@ -0,0 +1,64 @@
+///////////////////////////////////////////////////////////////////////////////
+// Name:        wx/qt/datectrl.h
+// Purpose:     wxDatePickerCtrl for Qt
+// Author:      Ali Kettab
+// Modified by:
```
vadz, line 5: Not worth updating the PR just for this, but, just FYI, we don't use this line any more in the new files, it doesn't seem worth it with `git shortlog` providing the same information much more reliably.

include/wx/qt/datectrl.h
```diff
@@ -0,0 +1,64 @@
...
+
+// ----------------------------------------------------------------------------
+// wxDatePickerCtrl
+// ----------------------------------------------------------------------------
+
+class WXDLLIMPEXP_ADV wxDatePickerCtrl : public wxDatePickerCtrlBase
+{
+public:
+    // ctors
+    wxDatePickerCtrl() { }
```
vadz, line 23: Again, not worth changing it here, but in the future, consider writing
```suggestion
    wxDatePickerCtrl() = default;
```
instead.

vadz (commented):
Great, thanks, I'll (finally) merge this soon.

PR #23926, opened 2023-10-01
Attempt to fix wxMSW UI tests and other improvements

vadz:
I'm not sure what's going on with AppVeyor but now at least one of a few tests seems to fail in every second build there, which makes it almost unusable (especially considering how long does it take to build there), so I'm trying to find some workaround.

I still can't reproduce any of AppVeyor failures locally, if anybody can, it would be great of they could be debugged...


vadz:
It's perhaps just a coincidence, but AppVeyor builds don't seem to fail any more with this, so let's merge it and see if it really helps.

PR #23931, opened 2023-10-02
Fix wxComboCtrl popup behaviour under some Wayland compositors

### LINKED ISSUE #23908: wxComboCtrl does not work properly on Wayland with Plasma and Sway when out of parent window bounds.
### Description

When the cursor is over a wxComboCtrl window, but goes out of the parent window bounds, seems like the actual cursor position gets constrained to it and you can resize the parent window this way.

KDE Plasma `4:5.27.8-0ubuntu1~ubuntu23.04~ppa1`:

https://github.com/wxWidgets/wxWidgets/assets/37658952/646edba5-d96a-4d8e-8a3f-1e81e688de0d

Also happens on Sway `1.7-6` (couldn't get OBS to record screen in it).

Not an issue on GNOME Wayland sessions.

- wxWidgets version  you use: [3.2.3 (proposed)](https://github.com/wxWidgets/wxWidgets/pull/23891)
- OS  and its version: Ubuntu 23.04
+ GTK version: 3.24.37-1ubuntu1

vadz:
We need to set the window type hint to COMBO in order to use a popup window for it under Wayland and while not doing it seems to still work with GNOME/mutter, it is required with the other compositors such as KDE Plasma or Sway and doesn't seem to do any harm under GNOME, so do it always and hope that it doesn't result in any problems with not combobox-like popups.

This basically undoes 62d934aab8 (Don't set HINT_COMBO as wxPopupWindow is used for different windows as well, 2008-10-07).

Closes #23908.

### DIFF
--- a/src/gtk/popupwin.cpp
+++ b/src/gtk/popupwin.cpp
@@ -106,8 +106,12 @@ bool wxPopupWindow::Create( wxWindow *parent, int style )
     g_object_ref( m_widget );
 
     gtk_widget_set_name( m_widget, "wxPopupWindow" );
-    // wxPopupWindow is used for different windows as well
-    // gtk_window_set_type_hint( GTK_WINDOW(m_widget), GDK_WINDOW_TYPE_HINT_COMBO );
+
+    // While wxPopupWindow is used for different windows as well, we don't
+    // really know how is it going to be used but we do know that without the
+    // hint at all, it doesn't work correctly, at least under Wayland, where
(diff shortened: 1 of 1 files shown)


vadz:
@dsa-t Sorry to bother you again, but I'd also like to merge this one soon in order to include it in the upcoming 3.2.3, so it would be great if you could please confirm that this works for you. TIA!

dsa-t:
It worked last time I tested, so looks good to me.

vadz:
Merged into master and added to #23891, so closing.

PR #23937, opened 2023-10-03
Minor fix to the recently added wxBitmap::QtBlendMaskWithAlpha()

AliKet:
Testing for the alpha channel support was a mistake because it is completely common to initially have a bitmap without an alpha channel ( in which case HasAlpha() returns false ) and Qt correctly converts it to an appropriate format when the mask is set.

Should be part of 1e43b37 (Added QtBlendMaskWithAlpha() to wxBitmap under wxQt)

@vadz Sorry for pushing untested code because adding `HasAlpha()` was a last minute change that I thought was a trivial change, but it's not.

### DIFF
--- a/src/generic/imaglist.cpp
+++ b/src/generic/imaglist.cpp
@@ -107,6 +107,13 @@ wxBitmap wxGenericImageList::GetImageListBitmap(const wxBitmap& bitmap) const
     }
 
 #ifdef __WXQT__
+    // In wxQt the returned bitmap is used with native list/tree controls which require
+    // the bitmap to already have its mask applied to it, as it's used directly by Qt
+    // and not via our wxDC::DrawBitmap() which would apply the mask itself, so we must
+    // take it into account here -- the only alternative would be for the controls to do
+    // it themselves, but then this would happen every time the bitmap is drawn, which
+    // would be less efficient than doing it just once here.
+
     bmpResized.QtBlendMaskWithAlpha();
 #endif
 
--- a/src/qt/bitmap.cpp
+++ b/src/qt/bitmap.cpp
@@ -597,7 +597,7 @@ void wxBitmap::QtBlendMaskWithAlpha()
     // alpha channel, if it has any. Notice that the bitmap can still be
     // converted to wxImage, but without mask information.
 
-    if ( IsOk() && HasAlpha() && M_MASK && M_MASK->GetHandle() )
+    if ( IsOk() && M_MASK && M_MASK->GetHandle() )
     {
         AllocExclusive();
         M_PIXDATA.setMask(*M_MASK->GetHandle());


vadz:
Sorry, I don't really have time to understand the issue right now, but I've merged this, thanks.

AliKet:
> Sorry, I don't really have time to understand the issue right now, but I've merged this, thanks.

The problem is best illustrated with an example:
```
1. wxImageList il(32, 32, true /*mask*/);
2.
3. wxBitmap bmpRGB(wxSize(32, 32), 24);
4. bmpRGB.SetMask(new wxMask(bmpRGB, *wxRED));
5.
6. it.Add(bmpRGB, *wxRED);
```
At line **6** we want to blend the bitmap's mask with the alpha channel (because this is what Qt `setMak()` actually does under the hood) only once before storing the result in the image list.
Now inside `QtBlendMaskWithAlpha()`, the mask will never be applied if we keep `HasAlpha()` test because `bmpRGB.HasAlpha()` is always false. and this is the problem.

BTW, there is another problem in the test now:
```
ImageList:WithMask
  Replace with RGB image
../../tests/graphics/imagelist.cpp:193

../../tests/graphics/imagelist.cpp:205: FAILED:
  CHECK( HasMaskOrAlpha(bmp1) )
with expansion:
  false

```
The test fails because it does this (kind of):
```
1. wxImageList il(32, 32, true /*mask*/);
2.
3. wxBitmap bmpRGB(wxSize(32, 32), 24);
4.
5. it.Add(bmpRGB);
```
The bitmap `bmpRGB` has only two colours: blue (background) and red.
The image list adds a mask to it with this colour `wxSystemSettings::GetColour(wxSYS_COLOUR_BTNFACE) /*light grey*/`
Inside `QtBlendMaskWithAlpha()`, after setMask() we delete the mask immediately. but bitmap is not changed at all at Qt level (i.e. hasAlphaChannel() still returns false) because the mask colour is not valid (i.e. doesn't mask anything)

I propose to change this [line](https://github.com/wxWidgets/wxWidgets/blob/4a22466ed7cf9807765e18daef25e9299ddb0b93/tests/graphics/imagelist.cpp#L76)  to: `mdc.SetBackground(wxSystemSettings::GetColour(wxSYS_COLOUR_BTNFACE));` which makes much more sense. or change `QtBlendMaskWithAlpha()` like this:

```
void wxBitmap::QtBlendMaskWithAlpha()
{
    // For performance reasons, permanently merge the mask with pixmap's
    // alpha channel, if it has any. Notice that the bitmap can still be
    // converted to wxImage, but without mask information.

    if ( IsOk() && M_MASK && M_MASK->GetHandle() )
    {
        AllocExclusive();
        M_PIXDATA.setMask(*M_MASK->GetHandle());
        if ( HasAlpha() )
            wxDELETE(M_MASK); // delete the mask only if it was correctly set!
    }
}
```
which doesn't make sense because why we keep a mask that cannot be applied at all?

AliKet:
Please let me think more about this.

vadz:
I think we could just leave it unspecified whether the bitmap added to the image list has a mask or not, nobody really cares about it, AFAICS. What we do care about is that it is drawn correctly (i.e. respecting the mask), but how it is achieved doesn't really matter.

IOW I'd just disable this test for wxQt with a comment explaining this.

Issue #10924, opened 2009-06-23
Double-clicking to select file in wxFileDialog results in EVT_MOTION event for the window beneath it


wxtrac:
*Issue migrated from trac ticket # 10924*

**component:** wxMSW | **priority:** low | **keywords:** wxFileDialog

#### 2009-06-23 17:59:26: Kerry (Kerry Loux) created the issue

___
The wxGLCanvas' EVT_MOTION event is triggered when it shouldn't be.  As an example:  A file dialog appears from which the user selects a file to load.  The file is loaded and displayed on the GL canvas, but it does not appear as expected.  This is because the EVT_MOTION event handler in the wxGLCanvas was overloaded to allow the user to rotate the object in the scene, and one of the openGL matricies was erroneously transformed prior to the scene being rendered.

This bug occurs when:
1.  A wxFileDialog is placed above a wxGLCanvas.
2.  The user selects a file by double-clicking on it.
3.  At this point, the EVT_MOTION event for the wxGLCanvas object is called.

This bug DOES NOT occur when:
1.  The double-clicking does not occur directly over top of hte wxGLCanvas (if the file name extends beyond the edge of the canvas, and it is selected by double-clicking on the portion of the file name that is not over the canvas).
2.  The user presses Enter or clicks the Open button to select the file.

This bug was observed with wxMSW 2.8.10.  I've attached an example that demonstrates this problem.

charvey2718:
This is maybe a similar workaround to that suggested by @RobinD42, but slightly different I think.
When a FileDialog is created, I set a global flag to suppress mouse clicks. Other controls that handle mouse clicks check the flag and only execute the event handling code if the flag is not set, and then unset the flag.
This means that controls that handle mouse clicks won't react to unconsumed events from the FileDialog.

It's not a general solution, onyl a workaround. And it's fairly simple, seems to work with only a few extra lines of code.

In case this can help anyone else, I have updated my minimum example in #23938 (closed as it duplicates this one) as follows:
````
#include <wx/wx.h>
#include <wx/filedlg.h>

class MyFrame:public wxFrame {
public:
    MyFrame(wxWindow* parent);
private:
    bool suppress_mouse_clicks = false;
};

class MyApp:public wxApp {
public:
    bool OnInit() {
        wxFrame* frame=new MyFrame(NULL);
        frame->Show();
        return true;                    
    }
};

DECLARE_APP(MyApp);
IMPLEMENT_APP(MyApp);

[...]

PR #23944, opened 2023-10-05
Update filedlg.cpp

charvey2718:
Fixes for issue #10924

### DIFF
--- a/src/msw/filedlg.cpp
+++ b/src/msw/filedlg.cpp
@@ -111,6 +111,21 @@ DWORD gs_oldExceptionPolicyFlags = 0;
 bool gs_changedPolicy = false;
 
 /*
+Fixes issue #10924
+Removes all mouse messages from the input queue. Prevents bogus mouse
+messages from being passed to controls positioned underneath the
+FileDialog after it has been destroyed.
+*/
+void DrainMouseMessages()
+{
+    MSG msg;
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, but could you please extract this into a trivial function in the beginning of this file, to make updating it later easier?

And, in fact, thinking more about it, I think we may want to amend this slightly by breaking from the loop after some big (1000? 10000?) number of iterations, just to be 100% sure that we don't get stuck forever if the user has a vibrating mouse generating events all the time or something other like this, which seems impossibly contrived and unrealistic until you run into it in practice...

charvey2718:
No worries - I'll get onto this later.

charvey2718:
@vadz I've done as you requested. Can you see the improvements as part of this PR or do I need to renew it somehow?

charvey2718 is fine for the change log, thanks.

vadz:
Thanks! I'm going to merge this in master with some minor changes (mostly more comments to explain what's going on here and, also, a better commit message) soon and also apply it to 3.2 branch so that it is included in the upcoming 3.2.3 release.

DRNadler:
Please note: This is a general problem for any dialog containing a wxListBox for example (probably any control that processes a double-click). If the handler dismisses the dialog, the left-up or move events may happen after the dialog is dismissed... See #25446

PR #23947, opened 2023-10-07
Fix CHOICE constant values that were reversed

Blake-Madden:
These constants are backwards from how they are documented in the help. I believe the ones in the help are correct, so use that.

It doesn't appear that these constants are actually applied to anything though, so this is mostly a consistency fix.


vadz:
Yes, as you say, they're not really used, so I see no point in updating them. We should add a comment saying that they're not used to the header and stop documenting them in the interface header instead (and update `wxGetSingleChoiceIndex()` documentation to mention that these parameters are not used at all).

Could you please do this instead? TIA!

PR #23948, opened 2023-10-07
Fix CHOICE constant values that were reversed

Blake-Madden:
These constants are backwards from how they are documented in the help. I believe the ones in the help are correct, so use that.

It doesn't appear that these constants are actually applied to anything though, so this is mostly a consistency fix.


vadz:
Sorry, but why did you recreate it when this seems completely identical to #23947?

Blake-Madden:
Sorry, I must have pushed this twice.

PR #23949, opened 2023-10-07
Explain that wxGetSingleChoice()'s size parameters are ignored

Blake-Madden:
Also, remove `wxCHOICE*` constants from the help that are not used for anything.
Finally, explain that you can resize a `wxSingleChoiceDialog` object instead if you need to change the dialog's size.


vadz:
Thanks, I'm pushing this with some minor changes.

PR #23953, opened 2023-10-08
Fix inserting new items into wxListBox under wxQt

AliKet:
Use the right type when inserting new items into the list box. This ensures that the `wxEVT_CHECKLISTBOX` event is emitted for these new items. And in fact, it should be part of this commit 16da1a1 when `wxQtListWidgetItem` was introduced.


AliKet:
The changes here are included in #23925

PR #23954, opened 2023-10-09
Remove wxNODISCARD from wxBasicString::Get() declaration

### LINKED ISSUE #23952: Build fails with clang and configure in MSYS2
Disclaimer: I am not familiar with _configure_, this issue was [reported on wxForum](https://forums.wxwidgets.org/viewtopic.php?f=19&t=50615).

I have installed MSYS2 and two group packages:
[mingw-w64-ucrt-x86_64-toolchain](https://packages.msys2.org/groups/mingw-w64-ucrt-x86_64-toolchain) (GCC 13.2.0)
[mingw-w64-clang-x86_64-toolchain](https://packages.msys2.org/groups/mingw-w64-clang-x86_64-toolchain) (clang 17.0.1)

I launched clang64.exe from the MSYS2 root folder, in the MSYS2 shell changed to the build folder and successfully ran
`../configure --with-msw --disable-dependency-tracking`:
```
Configured wxWidgets 3.3.0 for `x86_64-w64-mingw32'

  Which GUI toolkit should wxWidgets use?                 msw
  Should wxWidgets be compiled into single library?       no
  Should wxWidgets be linked as a shared library?         yes
  Unicode encoding used by wxString?                      UTF-16
  What level of wxWidgets compatibility should be enabled?
                                       wxWidgets 3.0      no
                                       wxWidgets 3.2      yes
  Which libraries should wxWidgets use?
                                       jpeg               builtin
                                       png                builtin
                                       regex              builtin
                                       tiff               builtin
                                       lzma               no
                                       zlib               sys
                                       expat              sys
                                       libmspack          no
                                       sdl                no
```
I then ran  `mingw32-make`, which resulted in a build error:
```
g++ -c -o basedll_filename.o   -I../3rdparty/pcre/src/wx  -ID:/Dev/Desktop/Lib/wxWidgets-PB/build-cl
ang-multi/lib/wx/include/msw-unicode-3.3 -I../include -D_FILE_OFFSET_BITS=64  -D__WXMSW__      -DWXB
UILDING -DwxUSE_GUI=0 -DWXMAKINGDLL_BASE -DwxUSE_BASE=1  -Wall -Wundef -Wunused-parameter -Wno-ctor-
dtor-privacy -Woverloaded-virtual -O2     ../src/common/filename.cpp
In file included from ../src/common/filename.cpp:106:
../include/wx/msw/ole/oleutils.h:107:5: error: 'nodiscard' attribute cannot be applied to types
  107 |     wxNODISCARD BSTR Get() const { return Copy(); }
      |     ^
../include/wx/defs.h:273:27: note: expanded from macro 'wxNODISCARD'
  273 |     #define wxNODISCARD [[nodiscard]]
      |                           ^
1 error generated.
mingw32-make: *** [Makefile:24253: basedll_filename.o] Error 1
```
`wxNODISCARD` is a feature present only in 3.3, added in f1985c6ba2018342de0dab05dd9e934d1627425e.

I noticed that error is in the only `wxBasicString` method that uses not only `wxNODISCARD` but also `wxDEPRECATED_MSG`:
https://github.com/wxWidgets/wxWidgets/blob/4235a018be3c42245f821959d0ee38bb67f43886/include/wx/msw/ole/oleutils.h#L106-L107

After commenting out the ` wxDEPRECATED_MSG` line above and rerunning make, the build finished successfully. 

wxWidgets own CMake-based clang MSYS2 CI succeeds: https://github.com/wxWidgets/wxWidgets/actions/runs/6442401188/job/17493193039 (and also does not have countless warnings the configure has?).

The issue is not present when building with GCC in MSYS2 (using the UCRT MSYS2 environment), so I am not sure if the problem is in wxWidgets or (more likely?) the toolchain.

- wxWidgets version: current GIT master
- wxWidgets port: wxMSW
- OS: Windows 10
- Compiler:
```
clang version 17.0.1
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: C:/msys64/clang64/bin
```
The original report is over a month old, so I think it happened also with clang 16; according to the original reporter, the issue manifests also with clang aarch64.

PBfordev:
Using both wxDEPRECATED_MSG and wxNODISCARD for some reason breaks compilation with clang (when using C++17 or newer standard).

Since wxBasicString::Get() is deprecated, one warning should be enough so just remove wxNODISCARD from the method declaration.

Closes #23952.

### DIFF
--- a/include/wx/msw/ole/oleutils.h
+++ b/include/wx/msw/ole/oleutils.h
@@ -104,7 +104,7 @@ public:
 
     // retrieve a copy of our string - caller must SysFreeString() it later!
     wxDEPRECATED_MSG("use Copy() instead")
-    wxNODISCARD BSTR Get() const { return Copy(); }
+    BSTR Get() const { return Copy(); }
 private:
     // actual string
     BSTR m_bstrBuf;


vadz:
Thanks! I've pushed this commit directly.

PR #23956, opened 2023-10-11
Make macOS date picker control locale aware

utelle:
- Add global variable for pointer to current NSLocale
- Set global NSLocale pointer in method wxUILocaleImplCF::Use()
- Add helper function to retrieve the global NSLocale pointer
- Add function declaration of the helper function in private/uilocale.h
- Call wxUILocaleImpl::Use() in method wxUILocale::GetCurrent(), so that the current wxUILocale instance is always valid, even if wxUILocale::UseDefault() was not explicitly called

### DIFF
--- a/include/wx/osx/private/uilocale.h
+++ b/include/wx/osx/private/uilocale.h
@@ -2,7 +2,7 @@
 // Name:        wx/osx/private/uilocale.h
 // Purpose:     Helper for making pointer to current NSLocale available
 //              for locale-dependent controls under macOS.
-// Author:      Vadim Zeitlin
+// Author:      Ulrich Telle
 // Created:     2023-10-13
 // Copyright:   (c) 2023 Vadim Zeitlin <vadim@wxwidgets.org>
 // Licence:     wxWindows licence
@@ -13,10 +13,8 @@
 
 #if wxUSE_INTL
 
-#ifdef __WXOSX_COCOA__
 // Function returning a pointer to the current NSLocale
 WXDLLIMPEXP_BASE NSLocale* wxGetCurrentNSLocale();
-#endif
 
 #endif
 
--- a/include/wx/osx/private/uilocale.h
+++ b/include/wx/osx/private/uilocale.h
@@ -1,0 +1,23 @@
+///////////////////////////////////////////////////////////////////////////////
+// Name:        wx/osx/private/uilocale.h
+// Purpose:     Helper for making pointer to current NSLocale available
+//              for locale-dependent controls under macOS.
+// Author:      Vadim Zeitlin
+// Created:     2023-10-13
+// Copyright:   (c) 2023 Vadim Zeitlin <vadim@wxwidgets.org>
+// Licence:     wxWindows licence
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef _WX_OSX_PRIVATE_UILOCALE_H_
+#define _WX_OSX_PRIVATE_UILOCALE_H_
+
+#if wxUSE_INTL
+
+#ifdef __WXOSX_COCOA__
+// Function returning a pointer to the current NSLocale
+WXDLLIMPEXP_BASE NSLocale* wxGetCurrentNSLocale();
+#endif
+
+#endif
+
+#endif // _WX_OSX_PRIVATE_UILOCALE_H_
--- a/include/wx/private/uilocale.h
+++ b/include/wx/private/uilocale.h
@@ -23,6 +23,11 @@ const wxLanguageInfos& wxGetLanguageInfos();
 // Function returning hard-coded values for the "C" locale.
 wxString wxGetStdCLocaleInfo(wxLocaleInfo index, wxLocaleCategory cat);
 
+#ifdef __WXOSX_COCOA__
+// Function returning a pointer to the current NSLocale
+WXDLLIMPEXP_BASE void* wxGetCurrentNSLocale();
+#endif
+
 // ----------------------------------------------------------------------------
 // wxUILocaleImpl provides the implementation of public wxUILocale functions
 // ----------------------------------------------------------------------------
--- a/include/wx/private/uilocale.h
+++ b/include/wx/private/uilocale.h
@@ -23,11 +23,6 @@ const wxLanguageInfos& wxGetLanguageInfos();
 // Function returning hard-coded values for the "C" locale.
 wxString wxGetStdCLocaleInfo(wxLocaleInfo index, wxLocaleCategory cat);
 
-#ifdef __WXOSX_COCOA__
-// Function returning a pointer to the current NSLocale
-WXDLLIMPEXP_BASE void* wxGetCurrentNSLocale();
-#endif
-
 // ----------------------------------------------------------------------------
 // wxUILocaleImpl provides the implementation of public wxUILocale functions
 // ----------------------------------------------------------------------------
--- a/src/common/uilocale.cpp
+++ b/src/common/uilocale.cpp
@@ -522,11 +522,14 @@ const wxUILocale& wxUILocale::GetCurrent()
     {
         // Instantiate standard C locale (should never fail) ...
         wxUILocaleImpl* impl = wxUILocaleImpl::CreateStdC();
+#ifdef __WXOSX__
         if (impl)
         {
             // ... and Use() it
+            // (Required to set the global NSLocale pointer)
             impl->Use();
         }
+#endif
         ms_current = wxUILocale(impl);
     }
 
--- a/src/common/uilocale.cpp
+++ b/src/common/uilocale.cpp
@@ -520,7 +520,14 @@ const wxUILocale& wxUILocale::GetCurrent()
     // We initialize it on demand.
     if ( !ms_current.m_impl )
     {
-        ms_current = wxUILocale(wxUILocaleImpl::CreateStdC());
+        // Instantiate standard C locale (should never fail) ...
+        wxUILocaleImpl* impl = wxUILocaleImpl::CreateStdC();
+        if (impl)
+        {
+            // ... and Use() it
+            impl->Use();
+        }
+        ms_current = wxUILocale(impl);
     }
 
     return ms_current;
--- a/src/common/uilocale.cpp
+++ b/src/common/uilocale.cpp
@@ -520,17 +520,7 @@ const wxUILocale& wxUILocale::GetCurrent()
     // We initialize it on demand.
     if ( !ms_current.m_impl )
     {
-        // Instantiate standard C locale (should never fail) ...
-        wxUILocaleImpl* impl = wxUILocaleImpl::CreateStdC();
-#ifdef __WXOSX__
-        if (impl)
-        {
-            // ... and Use() it
-            // (Required to set the global NSLocale pointer)
-            impl->Use();
-        }
-#endif
-        ms_current = wxUILocale(impl);
+        ms_current = wxUILocale(wxUILocaleImpl::CreateStdC());
     }
 
     return ms_current;
--- a/src/osx/core/uilocale.mm
+++ b/src/osx/core/uilocale.mm
@@ -35,6 +35,14 @@
 extern wxString
 wxGetInfoFromCFLocale(CFLocaleRef cfloc, wxLocaleInfo index, wxLocaleCategory cat);
 
+// Helper function to retrieve a pointer to the NSLocale of the current wxUILocale
+static NSLocale* gs_nsloc = nullptr;
+
+void* wxGetCurrentNSLocale()
+{
+  return gs_nsloc;
+}
+
 // ----------------------------------------------------------------------------
 // wxLocaleIdent::GetName() implementation using Foundation
 // ----------------------------------------------------------------------------
@@ -168,6 +176,8 @@ wxUILocaleImplCF::Use()
 {
     // There is no way to start using a locale other than default, so there is
     // nothing to do here.
+    // However, make the associated NSLocale instance accessible for internal use
+    gs_nsloc = m_nsloc;
 }
 
 wxString
--- a/src/osx/core/uilocale.mm
+++ b/src/osx/core/uilocale.mm
@@ -32,15 +32,17 @@
 #import <Foundation/NSLocale.h>
 #import <Foundation/NSDateFormatter.h>
 
+#include "wx/osx/private/uilocale.h"
+
 extern wxString
 wxGetInfoFromCFLocale(CFLocaleRef cfloc, wxLocaleInfo index, wxLocaleCategory cat);
 
 // Helper function to retrieve a pointer to the NSLocale of the current wxUILocale
-static NSLocale* gs_nsloc = nullptr;
+static NSLocale* gs_currentNSLocale = nullptr;
 
-void* wxGetCurrentNSLocale()
+NSLocale* wxGetCurrentNSLocale()
 {
-  return gs_nsloc;
+  return gs_currentNSLocale;
 }
 
 // ----------------------------------------------------------------------------
@@ -177,7 +179,7 @@ wxUILocaleImplCF::Use()
     // There is no way to start using a locale other than default, so there is
     // nothing to do here.
     // However, make the associated NSLocale instance accessible for internal use
-    gs_nsloc = m_nsloc;
+    gs_currentNSLocale = m_nsloc;
 }
 
 wxString
--- a/src/osx/core/uilocale.mm
+++ b/src/osx/core/uilocale.mm
@@ -42,7 +42,12 @@ static NSLocale* gs_currentNSLocale = nullptr;
 
 NSLocale* wxGetCurrentNSLocale()
 {
-  return gs_currentNSLocale;
+    static NSLocale* stdCLocale([[NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"] retain]);
+    if (!gs_currentNSLocale)
+    {
+        gs_currentNSLocale = stdCLocale;
+    }
+    return gs_currentNSLocale;
 }
 
 // ----------------------------------------------------------------------------
--- a/src/osx/core/uilocale.mm
+++ b/src/osx/core/uilocale.mm
@@ -42,9 +42,9 @@ static NSLocale* gs_currentNSLocale = nullptr;
 
 NSLocale* wxGetCurrentNSLocale()
 {
-    static NSLocale* stdCLocale([[NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"] retain]);
     if (!gs_currentNSLocale)
     {
+        static wxCFRef<NSLocale*> stdCLocale([[NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"] retain]);
         gs_currentNSLocale = stdCLocale;
     }
     return gs_currentNSLocale;
--- a/src/osx/cocoa/datetimectrl.mm
+++ b/src/osx/cocoa/datetimectrl.mm
@@ -23,6 +23,8 @@
 
 #include "wx/datetimectrl.h"
 #include "wx/datectrl.h"
+#include "wx/uilocale.h"
+#include "wx/private/uilocale.h"
 
 #include "wx/osx/core/private/datetimectrl.h"
 #include "wx/osx/cocoa/private/date.h"
@@ -187,6 +189,15 @@ wxDateTimeWidgetImpl::CreateDateTimePicker(wxDateTimePickerCtrl* wxpeer,
     [v setDatePickerElements: elements];
 
     [v setDatePickerStyle: NSTextFieldAndStepperDatePickerStyle];
+
+    if ( wxUILocale::GetCurrent().IsSupported() )
+    {
+        NSLocale* nsloc = (NSLocale*) wxGetCurrentNSLocale();
+        if (nsloc)
+        {
+            [v setLocale: nsloc];
+        }
+    }
 
     if ( style & wxDP_DROPDOWN )
     {
--- a/src/osx/cocoa/datetimectrl.mm
+++ b/src/osx/cocoa/datetimectrl.mm
@@ -24,11 +24,11 @@
 #include "wx/datetimectrl.h"
 #include "wx/datectrl.h"
 #include "wx/uilocale.h"
-#include "wx/private/uilocale.h"
 
 #include "wx/osx/core/private/datetimectrl.h"
 #include "wx/osx/cocoa/private/date.h"
 #include "wx/osx/private/available.h"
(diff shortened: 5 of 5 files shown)


src/common/uilocale.cpp
```diff
@@ -520,7 +520,14 @@ const wxUILocale& wxUILocale::GetCurrent()
     // We initialize it on demand.
     if ( !ms_current.m_impl )
     {
-        ms_current = wxUILocale(wxUILocaleImpl::CreateStdC());
+        // Instantiate standard C locale (should never fail) ...
```
vadz, line 523: I'm not 100% sure that calling `Use()` on C locale is a good idea. At least under Unix it's not a NOP.

Why not set `g_nsloc` in Mac-specific `CreateStdC()` instead?

include/wx/private/uilocale.h
```diff
@@ -23,6 +23,11 @@ const wxLanguageInfos& wxGetLanguageInfos();
 // Function returning hard-coded values for the "C" locale.
 wxString wxGetStdCLocaleInfo(wxLocaleInfo index, wxLocaleCategory cat);
 
+#ifdef __WXOSX_COCOA__
+// Function returning a pointer to the current NSLocale
+WXDLLIMPEXP_BASE void* wxGetCurrentNSLocale();
+#endif
+
```
vadz, line 30: I don't think this belongs to this header, it is Mac-, rather than locale-, specific.

Putting it in a (new) header under `include/wx/osx/private` would result in better physical (de)coupling and would allow to use `NSLocale*` as the function return type and so avoid the ugly casts too.

src/osx/core/uilocale.mm
```diff
@@ -35,6 +35,14 @@
 extern wxString
 wxGetInfoFromCFLocale(CFLocaleRef cfloc, wxLocaleInfo index, wxLocaleCategory cat);
 
+// Helper function to retrieve a pointer to the NSLocale of the current wxUILocale
+static NSLocale* gs_nsloc = nullptr;
```
vadz, line 39: This is really trivial, but I'd call it `gs_currentNSLocale` as extra clarity is worth the extra length for global variables IMHO.

vadz (commented):
Thanks for your work on this, but I think it could be improved a bit.

Please let me know if I'm missing anything here.

include/wx/private/uilocale.h
utelle: > I don't think this belongs to this header, it is Mac-, rather than locale-, specific.
> 
> Putting it in a (new) header under `include/wx/osx/private` would result in better physical (de)coupling and would allow to use `NSLocale*` as the function return type and so avoid the ugly casts too.

Ok, this can be done, of course.

src/osx/core/uilocale.mm
utelle: > This is really trivial, but I'd call it `gs_currentNSLocale` as extra clarity is worth the extra length for global variables IMHO.

Ok, I will apply this change.

src/common/uilocale.cpp
utelle: > I'm not 100% sure that calling `Use()` on C locale is a good idea. At least under Unix it's not a NOP.

Yes, `wxUILocale::Use()` is not a NOP under Unix. It sets the C runtime locale via `setlocale`. **But** does that matter in our case? IMHO **no**, because we would call `wxUILocale::Use()` only, if `wxUILocale::Use()` was **not** called. In that case we can assume that the current C runtime locale is the "**C**" locale anyway. Setting it in `wxUILocale::Use()` wouldn't cause any harm, would it?

> Why not set `g_nsloc` in Mac-specific `CreateStdC()` instead?

Unfortunately, this is not possible. `CreateStdC()` can be called from various places, not only from `Use()` or `GetCurrent()`, but also from the `wxUILocale()` constructor. Only in `Use()` you can be sure that the locale is the **current** locale.

include/wx/osx/private/uilocale.h
```diff
@@ -0,0 +1,23 @@
+///////////////////////////////////////////////////////////////////////////////
+// Name:        wx/osx/private/uilocale.h
+// Purpose:     Helper for making pointer to current NSLocale available
+//              for locale-dependent controls under macOS.
+// Author:      Vadim Zeitlin
```
vadz, line 5: Not really :-)

include/wx/osx/private/uilocale.h
```diff
@@ -0,0 +1,23 @@
...
+// Copyright:   (c) 2023 Vadim Zeitlin <vadim@wxwidgets.org>
+// Licence:     wxWindows licence
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef _WX_OSX_PRIVATE_UILOCALE_H_
+#define _WX_OSX_PRIVATE_UILOCALE_H_
+
+#if wxUSE_INTL
+
+#ifdef __WXOSX_COCOA__
```
vadz, line 16: This is not needed here, is it?

src/common/uilocale.cpp
vadz: You're probably right, but the trouble is that it's not immediately obvious and for me it's a bad sign when you need to carefully think about something instead of it just being immediately clear.

And it just seems suspicious to call `Use()` from an accessor. In fact I am still not sure to understand why do we need to remember this locale as the default one here, i.e. why does querying the default locale should change anything?

vadz (commented):
Thanks for the updates and sorry for being so cautious, but I'm still not sure if calling `Use()` from `GetCurrent()` is a good idea. Could you please re-explain why should we do it from there?

utelle:
> Thanks for the updates and sorry for being so cautious, but I'm still not sure if calling `Use()` from `GetCurrent()` is a good idea. Could you please re-explain why should we do it from there?

Sure. 

We have to differentiate between **creating** and **using** a `wxUILocale`. You can _**create** many_, but you only can _**use** one at a time_. Even if it may not be a common use case to create _many_ `wxUILocale` instances, it is perfectly allowed. To make _use_ of a `wxUILocale` we have 2 methods: `wxUILocale::UseDefault()` and `wxUILocale::UseLocaleName()`. An application should typically call one of these methods exactly once.

Both _use methods_ create a `wxUILocaleImpl` instance first, call then its `wxUILocaleImpl::Use()` method, and finally assign it to the global `wxUILocale` instance `ms_current`.  This `wxUILocaleImpl::Use()` method is called from nowhere else, and that is the reason, why this method is the only place where we can be sure that this `wxUILocaleImpl` will be the _current_ used locale. (And we need this fact to assign the current `NSLocale*` to the global variable in the MacOS implementation.)

Now, let us look at the implementation of `wxUILocale::GetCurrent()`. First of all this method is expected to **always** return a reference to a **valid** `wxUILocale`. Usually this method will simply return a reference to the global instance `ms_current`; and this will work, if one of the _use methods_ was called.

However, we have to handle one special case, namely if the application did **not** call one of the _use methods_. In that case the global instance `ms_current` is **not valid**, but we **need** a _valid instance_. 

The only assumption we can make is that the application _**uses**_  (:bangbang:) the standard C locale. So, what should we do to reflect that? Yes, we should call another _use method_, namely `wxUILocale::UseStdC()`. **Oops** ... such a method does not exist... ok ... let's do it manually (as it is necessary to do it only here, and therefore probably not worth to add a method `wxUILocale::UseStdC()`): _**create**_ a standard C locale and _**use**_ it.

The current implementation of `wxUILocale::GetCurrent()` makes the first step (creating a standard C locale) by calling `wxUILocaleImpl::CreateStdC()`, but misses to actually _**use**_ it. But remember: `wxUILocale::GetCurrent()` **expects** that the underlying locale is **used** (usually by calling `wxUILocale::UseDefault()`). Therefore I think it is a mistake to **not** call `wxUILocaleImpl::Use()`.

To cut a long story short: calling `wxUILocaleImpl::Use()` for the standard C locale from `wxUILocale::GetCurrent()` is actually fixing a bug. The alternative would be to create a method `wxUILocale::UseStdC()` and call that from `wxUILocale::GetCurrent()` to make absolutely clear that a call to a _use method_ was not done before.

Yes, calling `wxUILocaleImpl::Use()` is not a NOOP under _Linux_, but this is true for **all** _use methods_. They all call `setlocale` under the hood, but in this special case we just set the standard C locale again. And that should really do no harm at all.

src/common/uilocale.cpp
utelle: > You're probably right, but the trouble is that it's not immediately obvious and for me it's a bad sign when you need to carefully think about something instead of it just being immediately clear.

I fully agree. And you asked for a more detailed explanation in the general comments. However, I hope I could make my point in my answer there.

> And it just seems suspicious to call `Use()` from an accessor. In fact I am still not sure to understand why do we need to remember this locale as the default one here, i.e. why does querying the default locale should change anything?

Well, actually it is the same as with any method that instantiates a static global object when called for the first time: it checks whether the global object already exists and is initialized; if yes, a reference to the global object is returned; if no, the global object will allocated and initialized, before a reference to it will be returned.

Here we have the situation that on the **first** call to `GetCurrent()` the global object `ms_current` is not initialized, because `UseDefault()` was not called. One option would have been to call `UseDefault()` from `GetCurrent()`, because it was not done before by the application. However, that would force the application to use the default user locale, but it is unlikely that that is the intention of the application, when it did not call `UseDefault()`. More likely is that the application wants to stay with the standard C locale. Therefore we would need a method `UseStdC()` to make clear that we want to _**use** the standard C locale_. But we don't have it. Therefore `GetCurrent()` calls `CreateStdC()` ... and misses to call `Use()` for it (which would be part of `UseStdC()`. That is adding the call to `Use()` actually fixes a bug.

include/wx/osx/private/uilocale.h
utelle: No, probably not. At least if that header is not included from some non-Cocoa build. I don't know whether the latter can happen or not.

include/wx/osx/private/uilocale.h
utelle: I just copied the comments from another file, adjusting the _purpose_ but not paying attention to the other items.

vadz:
Thanks for the explanations but I have to say that I very much disagree with the basic assumption underlying them, namely that calling `GetCurrent()` somehow implies that the application wants to use "C" locale because this is not the case at all, using this accessor doesn't mean that and I don't know why do you think it does? It's just a request to get information about the current locale whatever it is. And I still think that calling `setlocale()` from `GetCurrent()` is completely wrong as it modifies the state of the program which is not something that an accessor is supposed to do.

In fact, now I'm quite sure that doing this is wrong because you've allowed me to understand what exactly is wrong with it: calling `GetCurrent()` is absolutely **not** equivalent to what `UseStdC()` would do (if it existed).

vadz:
Oh, and sorry for not putting this in the previous reply, but for the same reason `GetCurrent()` must **not** change the global wxOSX `NSLocale*` pointer neither. In fact, the fix is really trivial: the new code in `wxUILocale::GetCurrent()` should just be removed, i.e. this PR shouldn't change this function at all.

utelle:
> Thanks for the explanations but I have to say that I very much disagree with the basic assumption underlying them, namely that calling `GetCurrent()` somehow implies that the application wants to use "C" locale

Well, obviouly I was not able to make myself clear, unfortunately. The wxWidgets documentation states

>> "Localized applications should call `wxUILocale::UseDefault()` on startup to explicitly indicate that they opt-in using the current UI locale, ..."

That is, for a localized application that **calls** `wxUILocale::UseDefault()` the method `GetCurrent()` will simply return a reference to the current locale.

It is only the special case, if  `wxUILocale::UseDefault()` was **not** called before `GetCurrent()` gets called, we have to deal with. And that's what `GetCurrent()` did already, namely calling `CreateStdC()` and assigning it to the global instance `ms_current`.

IMHO it is a fair assumption in this situation that the current C locale is the _standard C locale_. Actually, that's why `GetCurrent()` assigns the standard C locale to `ms_current`, thus making it valid. And it did so already, before I opened the PR.

Actually, nobody knows which locale the application wants to use. It's up to the developer to call the right methods. For a localized application `UseDefault()` should be called; for a non-localized application most likely it is not called.

If `UseDefault()` is not called, the C runtime locale is typically the standard C locale. So, IMHO it is the right thing to use it as the UI locale (as is the case since `wxUILocale` was introduced.

The only change I introduced in my PR is that additionally `Use()` is called for the standard C locale (if it had to be created). If you don't do it, you will get a locale mix of standard C and user default under MacOS - as can be seen in the `internat` sample: the date field in the grid will be displayed according to the standard C locale, but if you start to edit the field it will be formatted according to the user default locale.

Under MacOS and under Windows the method `Use()` does absolutely nothing when called from the implementation of the standard C locale, except for setting the global `NSLocale*` in the MacOS build in my PR.

Under Linux the net effect is also zero, because each application starts with the standard C locale as the C runtime 

> because this is not the case at all, using this accessor doesn't mean that and I don't know why do you think it does? It's just a request to get information about the current locale whatever it is.

Sorry, we are talking about the special case that a `wxUILocale` was not created before `GetCurrent()` was called. And you have to handle that case in `GetCurrent()`. Otherwise the code calling `GetCurrent()` would have to deal with an invalid `wxUILocale`.

> And I still think that calling `setlocale()` from `GetCurrent()` is completely wrong as it modifies the state of the program which is not something that an accessor is supposed to do.

As explained above this is simply not true. If the application did not instantiate `wxUILocale`, what else could `GetCurrent()` do instead of creating and using the standard C locale? Call `Exit()` and stop the application? Certainly not.

> In fact, now I'm quite sure that doing this is wrong because you've allowed me to understand what exactly is wrong with it: calling `GetCurrent()` is absolutely **not** equivalent to what `UseStdC()` would do (if it existed).

**Again**: we are talking about a _special case_ that `GetCurrent()` **must** handle. And **not** calling `Use()` on the created standard C locale object leads to an inconsistent user experience.

utelle:
> Oh, and sorry for not putting this in the previous reply, but for the same reason `GetCurrent()` must **not** change the global wxOSX `NSLocale*` pointer neither. In fact, the fix is really trivial: the new code in `wxUILocale::GetCurrent()` should just be removed, i.e. this PR shouldn't change this function at all.

`GetCurrent()` will normally change nothing at all, but will simply return the reference to `ms_current`. It is only the _special case_ that `ms_current` was not set and is therefore invalid, that a standard C locale will created and assigned to `ms_current`. And only in this special case method `Use()` is called. Not calling `Use()` is an error, because you get an inconsistent user experience if the global `NSLocale` pointer is not set (as explained in my previous reply). And - also as explained - the net effect on the application state is zero.

vadz:
Just to make it clear: I'm only speaking about the case in which `UseDefault()` wasn't called. It's perfectly fine to call `GetCurrent()` in this case and it's also fine to change `ms_current` then because it has no observable effects. It's however **not** fine at all to call `setlocale()` because this does change the program state.

Sorry, but I still don't see at all why do you think that `GetCurrent()` need to do what you write. It simply returns the current locale, nothing else. If no locale has been set, it returns "C" locale, which is the only logical thing to do (at least we agree here). But it shouldn't do anything else.

I _guess_ (but I could be wrong here, so please feel free to ignore this paragraph in this case...) that you're looking at it from macOS-centric point of view, and macOS is weird because it doesn't use "C" locale by default. This weirdness should, however, be dealt with in wxOSX itself and doesn't change anything for the other platforms and shouldn't result in calling `setlocale()` under Linux.

utelle:
> Just to make it clear: I'm only speaking about the case in which `UseDefault()` wasn't called.

Me too.

> It's perfectly fine to call `GetCurrent()` in this case and it's also fine to change `ms_current` then because it has no observable effects.

Agreed.

> It's however **not** fine at all to call `setlocale()` because this does change the program state.

My point is that the current C runtime locale in this situation is "C". And therefore calling `setlocale` to set it to "C" should not have any effect. But I'm not an expert for `setlocale`. So, maybe I'm wrong, and even if the locale is the same, `setlocale` should not be called.

> Sorry, but I still don't see at all why do you think that `GetCurrent()` need to do what you write. It simply returns the current locale, nothing else. If no locale has been set, it returns "C" locale, which is the only logical thing to do (at least we agree here). But it shouldn't do anything else.

Regarding Windows and Linux I agree. But under MacOS calling `Use()` is required so that the locale-dependent code in wxWidgets uses the right `NSLocale`. We can't set the new global pointer from `CreateStdC` unfortunately.

> I _guess_ (but I could be wrong here, so please feel free to ignore this paragraph in this case...) that you're looking at it from macOS-centric point of view,

Correct.

> and macOS is weird because it doesn't use "C" locale by default. This weirdness should, however, be dealt with in wxOSX itself and doesn't change anything for the other platforms and shouldn't result in calling `setlocale()` under Linux.

Ok. I hope you can accept the compromise I just committed. Namely calling `Use()` only on MacOS.

vadz:
> > It's however **not** fine at all to call `setlocale()` because this does change the program state.
> 
> My point is that the current C runtime locale in this situation is "C". And therefore calling `setlocale` to set it to "C" should not have any effect. But I'm not an expert for `setlocale`. So, maybe I'm wrong, and even if the locale is the same, `setlocale` should not be called.

It shouldn't because the program might call `setlocale()` itself (possibly with some other locale). And possibly overriding the global locale without being asked for is definitely rude (pointedly not looking at GTK).

> Ok. I hope you can accept the compromise I just committed. Namely calling `Use()` only on MacOS.

I could live with this, even though I'd prefer not to have any unnecessary platform checks in the common code, but I still don't understand why are you so convinced that it is necessary. If the program doesn't call `wxUILocale::UseDefault()`, it basically doesn't care about the locale. And it's not clear at all to me that showing the dialog in US English (which is this change accomplishes, AFAIU?) is better than showing it in the system default language (which is how I believe it behaves by default?).

So if it were up to me, I would still just delete this code. And if you really want to use C locale in the dialog by default, it looks like you could easily do it by just handling the case of `gs_currentNSLocale == nullptr` as meaning this, couldn't you?

utelle:
> > Ok. I hope you can accept the compromise I just committed. Namely calling `Use()` only on MacOS.
> 
> I could live with this, even though I'd prefer not to have any unnecessary platform checks in the common code,

IMHO it is not unnecessary.

> but I still don't understand why are you so convinced that it is necessary. If the program doesn't call `wxUILocale::UseDefault()`, it basically doesn't care about the locale. 

My understanding is that we want consistent behaviour as far as it is possible. Since the Standard C locale is assumed, if no locale was explicitly set, it should be used throughout. That is, not mixing C locale and default user locale where possible. 

> And it's not clear at all to me that showing the dialog in US English (which is this change accomplishes, AFAIU?) is better than showing it in the system default language (which is how I believe it behaves by default?).

I speak of number and date formatting. I would find it irritating, if a date is formatted according to US English for display, but according to German for editing.

> So if it were up to me, I would still just delete this code. And if you really want to use C locale in the dialog by default, it looks like you could easily do it by just handling the case of `gs_currentNSLocale == nullptr` as meaning this, couldn't you?

Yes, in principle this could be done, but I find it inconvenient, because you need then to instantiate each time a NSLocale for C, instead of simply reusing the NSLocale of the current wxUILocale.

utelle:
@vadz I hope you will be satisfied by the latest changes. I removed the MacOS-specific code from `wxUILocale::GetCurrent()`. The helper function `wxGetCurrentNSLocale()` now returns a standard C locale if `Use()` was not called for any locale.

include/wx/osx/private/uilocale.h
vadz: Will you change it or should I, when merging?

include/wx/osx/private/uilocale.h
vadz: I'd still remove it.

src/osx/core/uilocale.mm
```diff
@@ -32,9 +32,24 @@
...
+
 extern wxString
 wxGetInfoFromCFLocale(CFLocaleRef cfloc, wxLocaleInfo index, wxLocaleCategory cat);
 
+// Helper function to retrieve a pointer to the NSLocale of the current wxUILocale
+static NSLocale* gs_currentNSLocale = nullptr;
+
+NSLocale* wxGetCurrentNSLocale()
+{
+    static NSLocale* stdCLocale([[NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"] retain]);
```
vadz, line 45: This should be inside `if`, why even instantiate it if it's never going to be needed?

It's also going to leak currently, not sure if it's really important, but using `wxCFRef<NSLocale>` for it would avoid it.

vadz (commented):
Thanks for the update! IMO it's indeed much better now, as there is less code and it's perfectly clear what's going on, so I'd be glad to merge it, just please let me know what do you think of the small issues below.

Thanks again!

src/osx/core/uilocale.mm
utelle: I moved the static locale into the if condition. And I used `wxCFRef`. However, I was not sure whether I have to keep the `retain` keyword or not.

include/wx/osx/private/uilocale.h
utelle: Removed.

include/wx/osx/private/uilocale.h
utelle: I changed the author name.

vadz:
Sorry, somehow I've never got back to this and by now I, of course, don't remember anything any longer :-(

The code changes look fine to me now, thanks for addressing the previous comments.

But, just to be sure I understand the change in behaviour implemented here: do I understand correctly that currently date picker uses the user default locale under macOS (e.g. German on your system, presumably) but with this change, if `wxUILocale::UseDefault()` is _not_ called, it would use English now?

Sorry again for the huge delay, I promise to merge this a.s.a.p. if you could just please confirm that the above is correct. TIA!

utelle:
> Sorry, somehow I've never got back to this and by now I, of course, don't remember anything any longer :-(

Same here. I hardly remember the details.

> The code changes look fine to me now, thanks for addressing the previous comments.
> 
> But, just to be sure I understand the change in behaviour implemented here: do I understand correctly that currently date picker uses the user default locale under macOS (e.g. German on your system, presumably) but with this change, if `wxUILocale::UseDefault()` is _not_ called, it would use English now?

Correct. If `wxUILocale::UseDefault()` was not called explicitly, `wxUILocale::GetCurrent()` will return the standard C locale (corresponding to `en_US@POSIX`). And the date picker control will then use English (instead of German on my system).

Additionally, if the locale was set via `wxLocale` - that is, with method `wxUILocale::UseLocaleName()` -, the date picker control will use the language of that locale.

> Sorry again for the huge delay, I promise to merge this a.s.a.p. if you could just please confirm that the above is correct. TIA!

See above.

PR #23960, opened 2023-10-12
Fixing MSW Dark Mode crash with empty wxListCtrl

andrewfb:
Thanks very much for all the recent work on MSW Dark Mode.

I'm seeing an empty wxListCtrl will crash when Dark Mode is enabled on MSW due to a recent-ish change: 9710592cc4c35f60d52257e8d159e1b2448a268c

If `itemCount` is 0, the local `bottom` index becomes -1, which leads to an assert failing downstream. This small fix should prevent that. I've tried to create a minimal repro which can be pasted over the top of the _minimal_ sample:

```
#include <wx/wx.h>
#include <wx/listctrl.h>
#include <wx/msw/app.h>

class MyFrame : public wxFrame {
  public:
    MyFrame()
        : wxFrame( nullptr, wxID_ANY, "Dark Mode wxListCtrl Crash", wxDefaultPosition, wxSize(800, 600) )
    {
        wxPanel* panel = new wxPanel( this, wxID_ANY );
        mListCtrl = new wxListCtrl( panel, wxID_ANY, wxDefaultPosition, wxDefaultSize, wxLC_REPORT );
    }

    wxListCtrl*     mListCtrl;
};

class MyApp : public wxApp
{
  public:
    virtual bool OnInit()
    {
        wxApp::MSWEnableDarkMode();

        MyFrame* frame = new MyFrame();
        frame->Show(true);
        return true;
    }
};

wxIMPLEMENT_APP(MyApp);
```

### DIFF
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3408,7 +3408,7 @@ void wxListCtrl::OnPaint(wxPaintEvent& event)
     const long top = GetTopItem();
     const long bottom = wxMin(top + countPerPage, itemCount - 1);
 
-    if ( needToErase )
+    if ( needToErase && itemCount )
     {
         wxRect lastRect;
         GetItemRect(bottom, lastRect);


vadz:
Thanks for reporting and fixing this!

However I'd prefer to fix it a bit differently:

```diff
diff --git a/src/msw/listctrl.cpp b/src/msw/listctrl.cpp
index 11d6f5bd95..ecc391096b 100644
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3378,9 +3378,9 @@ void wxListCtrl::OnPaint(wxPaintEvent& event)
     // don't align with the separators drawn by the header control.
     bool needToDraw = false,
          needToErase = false;
-    if ( InReportView() )
+    if ( InReportView() && itemCount )
     {
-        if ( (drawHRules || drawVRules) && itemCount )
+        if ( (drawHRules || drawVRules) )
             needToDraw = true;
         else if ( wxMSWDarkMode::IsActive() && wxGetWinVersion() >= wxWinVersion_11 )
             needToErase = true;
```

Do you see any problems with doing it like this?

andrewfb:
That makes sense, and I can confirm that still fixes my use case. Would you rather receive a fresh PR or just commit your own change and close this PR out?

vadz:
I'll push it soon, thanks for confirming that this works!

PR #23962, opened 2023-10-13
Improve item sorting in generic wxFileCtrl and wxDirCtrl

PBfordev:
When users browse for files or folders on a modern system, they expect to see them sorted in the natural order.
However, to sort its items (files and folders) `wxFileCtrl` used simple string compare and `wxDirCtrl` case-insensitive string compare.

For example, this is how an example folder looks in File Explorer:
sort-explorer
but this is how it looks in `wxFileCtrl` and `wxDirCtrl`, where one can see the unexpected order, e.g., `Z` before `a` (for `čáp` may be only for Czech and Slovaks):
sort-filectrl sort-dirctrl

Therefore, I think it is better to use `wxCmpNatural()` for file and folder names and `wxStricmp()` for file extensions.

### DIFF
--- a/src/generic/dirctrlg.cpp
+++ b/src/generic/dirctrlg.cpp
@@ -234,13 +234,6 @@ bool wxIsDriveAvailable(const wxString& dirName)
 
 #if wxUSE_DIRDLG
 
-// Function which is called by quick sort. We want to override the default wxArrayString behaviour,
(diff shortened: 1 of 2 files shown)


vadz (approved):
This is definitely the right thing to do, thanks for the update!

Will merge it soon.

PR #23963, opened 2023-10-14
Re-enable mfc sample

LegalizeAdulthood:
I was able to build and run the sample without issues.

### DIFF
--- a/build/cmake/samples/CMakeLists.txt
+++ b/build/cmake/samples/CMakeLists.txt
@@ -279,8 +279,7 @@ if(WIN32)
 
     if(WXMSW)
         wx_add_sample(nativdlg nativdlg.cpp nativdlg.h resource.h RES nativdlg.rc)
-        #TODO: reenable when sample is fixed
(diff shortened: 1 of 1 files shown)


vadz:
I think the problem here is that it only builds with MSVS and only when MFC is available (which is not always the case any more). So we'd need to test that MSVS is being used (probably simple) and ideally also that MFC is available (e.g. by checking for one of its headers?), otherwise this will break the build for all MinGW users.

Kumazuma:
it is possible with [MSVC CMake Variable](https://cmake.org/cmake/help/v3.0/variable/MSVC.html) to check compiler is Visual Studio. MFC base header file is afxwin.h

LegalizeAdulthood:
Checking `MSVC` is probably the right thing to do, I'll look at a fix.

vadz:
Testing for MSVC is simple, but testing that MFC is available is not, so somebody would need to do it. For now I'm closing it as it can't be applied as is.

PBfordev:
Using [FindMFC](https://cmake.org/cmake/help/latest/module/FindMFC.html) and adding the MFC sample only when `MFC_FOUND` would seem an obvious solution but I cannot test it.

PR #23966, opened 2023-10-14
Add support for initializer_list on arrays

Lotendan:
Port these changes in 3.2 as was requested in #23309

### DIFF
--- a/include/wx/defs.h
+++ b/include/wx/defs.h
@@ -505,6 +505,14 @@ typedef short int WXTYPE;
     #endif
 #endif /* HAVE_WOSTREAM */
 
+#ifndef wxHAVE_INITIALIZER_LIST
+    #if __cplusplus >= 201103L
+        #define wxHAVE_INITIALIZER_LIST
+    #elif wxCHECK_VISUALC_VERSION(11)
+        #define wxHAVE_INITIALIZER_LIST
+    #endif
+#endif /* wxHAVE_INITIALIZER_LIST */
+
 /*  ---------------------------------------------------------------------------- */
 /*  portable calling conventions macros */
 /*  ---------------------------------------------------------------------------- */
--- a/include/wx/dynarray.h
+++ b/include/wx/dynarray.h
@@ -15,6 +15,8 @@
 
 #include "wx/vector.h"
 
+#include <initializer_list>
+
 /*
   This header defines legacy dynamic arrays and object arrays (i.e. arrays
   which own their elements) classes.
@@ -104,6 +106,9 @@ public:
     wxBaseArray(InputIterator first, InputIterator last)
         : base_vec(first, last)
     { }
+
+    template<typename U>
+    wxBaseArray(std::initializer_list<U> list) : base_vec(list.begin(), list.end()) {}
 
     void Empty() { this->clear(); }
     void Clear() { this->clear(); }
@@ -519,6 +524,8 @@ private:
         name(size_t n, Base::const_reference v) : Base(n, v) { }              \
         template <class InputIterator>                                        \
         name(InputIterator first, InputIterator last) : Base(first, last) { } \
+        template<typename U>                                                  \
+        name(std::initializer_list<U> list) : Base(list.begin(), list.end()) { } \
     }
 
 
--- a/include/wx/dynarray.h
+++ b/include/wx/dynarray.h
@@ -15,7 +15,9 @@
 
 #include "wx/vector.h"
 
-#include <initializer_list>
+#ifdef wxHAVE_INITIALIZER_LIST
+    #include <initializer_list>
+#endif
 
 /*
   This header defines legacy dynamic arrays and object arrays (i.e. arrays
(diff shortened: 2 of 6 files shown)


imciner2:
3.2 doesn't have guaranteed support for standard library features, so this needs to be guarded behind one of the configure flags. Simplest would be guard it with `wxUSE_STL` to guard against just the normal standard library enablement, but I don't know if it would be better for it to be behind one of the other one guarding the containers specifically.

interface/wx/arrstr.h
```diff
@@ -89,6 +89,14 @@ class wxArrayString : public wxArray
     */
     wxArrayString(size_t sz, const wxString* arr);
 
+    /**
+        Constructs the container with the contents of the initializer_list @a list.
+
+        @since 3.2.3
```
vadz, line 95:
```suggestion
        @since 3.2.4
```

interface/wx/dynarray.h
```diff
@@ -271,6 +271,14 @@ class wxArray<T>
     */
     wxObjArray(const wxObjArray& array);
 
+    /**
+        Constructs the container with the contents of the initializer_list @a list.
+
+        @since 3.2.3
```
vadz, line 277:
```suggestion
        @since 3.2.4
```

include/wx/dynarray.h
```diff
@@ -519,6 +524,8 @@ class wxBaseObjectArray : private wxBaseArray<T*>
         name(size_t n, Base::const_reference v) : Base(n, v) { }              \
         template <class InputIterator>                                        \
         name(InputIterator first, InputIterator last) : Base(first, last) { } \
+        template<typename U>                                                  \
+        name(std::initializer_list<U> list) : Base(list.begin(), list.end()) { } \
```
vadz, line 528: This would have to use another helper macro, defined differently in C++11 and not case.

vadz (commented):
Thanks, but unfortunately it's indeed a bit more involved than this because we need to keep C++98 support in 3.2.

I think it would be enough to just add

```cpp
#if __cplusplus >= 201103L
```

checks around the code using `initializer_list` or, maybe better, do something like

```cpp
#if __cplusplus >= 201103L
  #define wxHAVE_INITIALIZER_LIST
#elif wxCHECK_VISUALC_VERSION(11) // Not sure about the exact version?
  #define wxHAVE_INITIALIZER_LIST
#endif

#ifdef wxHAVE_INITIALIZER_LIST
    #include <initializer_list>
#endif
```

include/wx/defs.h
```diff
@@ -505,6 +505,14 @@ typedef short int WXTYPE;
     #endif
 #endif /* HAVE_WOSTREAM */
 
+#ifndef wxHAVE_INITIALIZER_LIST
+    #if __cplusplus >= 201103L
+        #define wxHAVE_INITIALIZER_LIST
+    #elif wxCHECK_VISUALC_VERSION(11)
```
Lotendan, line 511: Visual Studio 10 offers an early support for C++11, but that does not include `initializer_list`.
Real support for it starts with Visual Studio 11.

include/wx/dynarray.h
```diff
@@ -508,18 +517,30 @@ class wxBaseObjectArray : private wxBaseArray<T*>
...
-    {                                                                         \
-        typedef wxBaseArrayFor##name Base;                                    \
-    public:                                                                   \
-        name() : Base() { }                                                   \
-        explicit name(size_t n) : Base(n) { }                                 \
-        name(size_t n, Base::const_reference v) : Base(n, v) { }              \
-        template <class InputIterator>                                        \
-        name(InputIterator first, InputIterator last) : Base(first, last) { } \
-    }
+#define wxDEFINE_TYPEARRAY_WITH_DECL__BASEARRAY(T, name, base, classdecl)         \
```
Lotendan, line 520: I'm not really fond of this split macro, but it is the only way I found to avoid duplicating the code. Let me know if you have a better solution for this.

tests/arrays/arrays.cpp
```diff
@@ -623,6 +636,9 @@ TEST_CASE("wxDynArray::" #name, "[dynarray]")                                 \
     CHECK( b.Index( 5 ) == 2 );                                      \
     CHECK( b.Index( 6 ) == wxNOT_FOUND );                            \
     CHECK( b.Index( 17 ) == 3 );                                     \
+                                                                     \
+    wxArray##name c({1,2,3});                                        \
```
Lotendan, line 640: I imagine I need to duplicate this big macro as well (and this is the reason why the tests are failing).
Any other suggestion?

vadz: Again, although it's still not going to be pretty, adding `WX_TEST_INIT_LIST` helper macro defined as non-empty only if init list is available would make this not too horrible.

vadz (commented):
Thanks for the update! I hope to have answered your questions, but please let me know if I didn't.

PR #23967, opened 2023-10-14
Fix using wxStaticBoxSizer in samples

PBfordev:
If a control is contained in a wxStaticBoxSizer or its child sizer, wxStaticBoxSizer::GetStaticBox() should be used as the control's parent.

However, widgets and dialogs samples did not do that, which resulted in flooding the message log in widgets sample with warnings about this; in dialogs sample the warnings were shown in the debug output.

Fix this by always using the static box as the control's parent when the control is contained (directly or indirectly) in a wxStaticBoxSizer.

The warnings are not there in 3.2 branch, but if backporting this should comprise of just replacing `nullptr` with `NULL`, perhaps it could be backported, to promote to correct usage of `wxStaticBoxSizer`.

### DIFF
--- a/samples/dialogs/dialogs.cpp
+++ b/samples/dialogs/dialogs.cpp
@@ -2421,16 +2421,17 @@ public:
 #endif
         wxSizer * const sizerTop = new wxBoxSizer(wxVERTICAL);
 
-        wxSizer* sizerText = new wxStaticBoxSizer(wxVERTICAL, this, "Notification Texts");
+        wxStaticBoxSizer* sizerText = new wxStaticBoxSizer(wxVERTICAL, this, "Notification Texts");
+        wxStaticBox* const sizerTextBox = sizerText->GetStaticBox();
 
-        sizerText->Add(new wxStaticText(this, wxID_ANY, "&Title:"),
+        sizerText->Add(new wxStaticText(sizerTextBox, wxID_ANY, "&Title:"),
             wxSizerFlags());
-        m_textTitle = new wxTextCtrl(this, wxID_ANY, "Notification Title");
+        m_textTitle = new wxTextCtrl(sizerTextBox, wxID_ANY, "Notification Title");
         sizerText->Add(m_textTitle, wxSizerFlags().Expand());
 
-        sizerText->Add(new wxStaticText(this, wxID_ANY, "&Message:"),
+        sizerText->Add(new wxStaticText(sizerTextBox, wxID_ANY, "&Message:"),
             wxSizerFlags());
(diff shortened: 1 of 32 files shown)


samples/dialogs/dialogs.cpp
```diff
@@ -2421,16 +2421,16 @@ class TestNotificationMessageWindow : public wxFrame
 #endif
         wxSizer * const sizerTop = new wxBoxSizer(wxVERTICAL);
 
-        wxSizer* sizerText = new wxStaticBoxSizer(wxVERTICAL, this, "Notification Texts");
+        wxStaticBoxSizer* sizerText = new wxStaticBoxSizer(wxVERTICAL, this, "Notification Texts");
 
-        sizerText->Add(new wxStaticText(this, wxID_ANY, "&Title:"),
+        sizerText->Add(new wxStaticText(sizerText->GetStaticBox(), wxID_ANY, "&Title:"),
```
vadz, line 2426: I know beauty is not the most important thing in this code, but it still bothers me to repeat `sizerText->GetStaticBox()` so many times here. Could we please do something like

```suggestion
        wxStaticBox* const boxTest = sizerText->GetStaticBox();
        
        sizerText->Add(new wxStaticText(boxText, wxID_ANY, "&Title:"),
```

instead?

vadz (commented):
Thanks for taking time to fix this!

There are some build errors due to the fact that the full declaration of `wxStaticBox` is now needed in places where it wasn't needed before, but this should be simple to fix by just including `"wx/statbox.h"` where needed.

samples/dialogs/dialogs.cpp
PBfordev: Did you mean (a) just there, (b) everywhere in the dialogs sample, or (c) everywhere in both samples?

The current code in widgets sample uses `wxStaticBoxSizer` in two ways:
1. Create `wxStaticBox` first and then use it with `wxStaticBoxSizer` (often originally declared just as `wxSizer`).
2. Just create `wxStaticBoxSizer`.

When doing my changes, I opted for minimum changes to the existing code and used as the parent the existing `wxStaticBox` instance in the first case and `wxStaticBoxSizer::GetStaticBox()` in the second one.

vadz: I think it would be nice to do it everywhere where `sizer->GetStaticBox()` is used more than 2-3 times, as IMO the code becomes even more difficult to read with these calls everywhere in it.

PBfordev: I updated the code, in the spirit of the request. Hopefully it is good enough now.

vadz: Great, thanks for the update! I'll merge this soon.

Issue #23961, opened 2023-10-13
[Wayland] wxGTK crashes with GDK Error 71 on Aegisub


witchymary:
### To Reproduce:

1. Download Aegisub from the Arch repos
2. Launch it with `GDK_BACKEND=wayland aegisub`
3. Go to `Video` -> `Open Video...` and open an mkv file

### Description

Aegisub's video renderer crashes consistently on Wayland with the error: `Gdk-Message: 16:09:57.355: Error 71 (Protocol error) dispatching to Wayland display.` Downgrading to `3.2.2.1` fixes the issue. Here's the `WAYLAND_DEBUG=1` [log](https://pastebin.com/nVfkaN3h).

### Platform and version information

- wxWidgets version you use: 3.2.3
- wxWidgets port you use: wxGTK
- OS: Arch Linux with Sway

+ GTK version: 3.24.38-1 
+ Which GDK backend is used: Wayland

vadz:
The relevant messages seem to be just those at the very end:

```
wl_display@1.error(wl_subcompositor@43, 0, "Cannot reassign role wl_subsurface to wl_surface@40, role object still exists")
Gdk-Message: 16:24:27.856: Error 71 (Protocol error) dispatching to Wayland display.
```

but I don't really know where does this happen. Could you please run the program under gdb, break on `_exit` (which is what I think must be getting called by GTK in this situation) and use `bt` to check if it can provide us with more information?

I suspect this might be due to the changes of 1fb96180f4 (Handle map/unmap events on Wayland's wxGLCanvasEGL, 2023-08-25), see #22835, so cc'ing @joanbm.

witchymary:
This is my first time using gdb, so apologies if I got anything wrong.

```gdb
Gdk-Message: 17:40:42.028: Error 71 (Protocol error) dispatching to Wayland display.

Thread 1 "aegisub" hit Breakpoint 1.1, __GI__exit (status=status@entry=1)
    at ../sysdeps/unix/sysv/linux/_exit.c:27
27	{
(gdb) bt

#0  __GI__exit (status=status@entry=1) at ../sysdeps/unix/sysv/linux/_exit.c:27
#1  0x00007ffff336b868 in _gdk_wayland_display_queue_events (display=<optimized out>)
    at ../gtk/gdk/wayland/gdkeventsource.c:205
#2  _gdk_wayland_display_queue_events (display=<optimized out>)
[...]

PR #23968, opened 2023-10-15
Ignore spurious map-events on Wayland's wxGLCanvasEGL 

joanbm:
Fixes the Aegisub regression reported on issue #23961.

This kind of fix feels a bit dirty but as far as I can tell, there's not much else we can do due to the way GTK manages the Wayland resources associated with the widget and its top-level window.

It's also a simple fix with not much chance of breaking other things, so it should be safe to include it in 3.2.4.

### DIFF
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -627,6 +627,21 @@ wxGLCanvasEGL::~wxGLCanvasEGL()
 void wxGLCanvasEGL::CreateWaylandSubsurface()
 {
 #ifdef GDK_WINDOWING_WAYLAND
+    // It's possible that we get in here unnecessarily in two ways:
+    // (1) If the canvas widget is shown, and then immediately hidden, we will
+    //     still receive a map-event signal, but by that point, the subsurface
+    //     does not need to be created anymore as the canvas is hidden
+    // (2) If the canvas widget is shown, and then immediately hidden, and then
+    //     immediately shown again, we will receive two map-event signals.
+    //     By the second time we get it, the subsurface will already be created
+    // Not ignoring either of the two scenarios will likely cause the subsurface
+    // to be created twice, leading to a crash due to a Wayland protocol error
+    // See https://github.com/wxWidgets/wxWidgets/issues/23961
+    if ( !gtk_widget_get_mapped(m_widget) || m_wlSubsurface )
+    {
+        return;
+    }
+
     GdkWindow *window = GTKGetDrawingWindow();
     struct wl_surface *surface = gdk_wayland_window_get_wl_surface(window);
 


vadz:
Thanks, no comments about the fix (I guess we could set a flag in unmap callback and check it here too, but I don't really see much benefit in doing this), but just a question: can we recommend some workaround to Aegisub or will we need to release 3.2.4 to allow it to work again?

joanbm:
I think the only workaround is to launch Aegisub under the X11 backend, i.e. with `GDK_BACKEND=x11 aegisub`, to use it under Xwayland.

Furthermore, I was doing more tests and I found that there's another similar regression triggered by running Aegisub with a video as an argument (i.e. `aegisub test.avi`) which is *not* fixed by this PR. What happens is that in this case Aegisub does something like `window->Show(true); canvas->Show(false); canvas->Show(true);` which enqueues two "map-event" signals right after each other, which also results in a protocol error. I think this can be fixed precisely with the flag approach you propose so I'll update the PR hopefully later today.

Furthermore something I noticed is that at least on my setup, on wxWidgets 3.2.3, resizing the Aegisub video control using either the scroll wheel over it or the dropdown with the percentages doesn't work consistently, i.e. the control stays at the same size sometimes. This appears to be an unrelated, more minor issue (I think related to 0ded71ff109daae2baa83cd8e835ec775649d32c) but it's something I'd also like to look at.

joanbm:
PR updated using the flag approach, which also easily lets us ignore "duplicate" map-events which happen if you show+hide+show a canvas without going through the event loop.

This fixes the Aegisub crashes, both where  is a video is opened from the UI, and the case where a video is specified on the command line.

include/wx/unix/glegl.h
```diff
@@ -125,6 +125,7 @@ class WXDLLIMPEXP_GL wxGLCanvasEGL : public wxGLCanvasBase
     // Only called when using Wayland to indicate that we should be redrawn.
     void OnWLFrameCallback();
 
+    bool m_wlPendingMap;
```
vadz, line 128: Don't we need to initialize it to `false` here?

vadz (commented):
Thanks, the fix looks good to me, the only problem is that we can't apply it to 3.2 as is because it would break ABI, so instead of adding a new field we're going to need to add a global set `canvasesWithPendingMap` and add/remove this object to/from it instead of setting/unsetting the flag.

include/wx/unix/glegl.h
joanbm: If my understanding is correct, it's not necessary because you're guaranteed to get a `"map"` signal (which initializes the flag) before a `"map-event"` (which reads the flag). Though it's probably not a bad idea to initialize it anyway, just in case...

joanbm:
Hmmmm, of course there's ABI compatibility for 3.2...

A `canvasesWithPendingMap` would certainly work but it seems a bit ugly to manage (aside from getting/setting the flag, we'd need to clean it on the destructor to avoid leaks / dangling pointers, and I'm not sure if we'd need to make it thread safe - or is everything always supposed to come from the UI thread?).

So, I've updated the PR to go back the flag-less approach, but this time with an added condition to check for duplicate map-events.
To my knowledge, this new version should be equivalent to the flag approach, and thus handle both the crash when opening a video in Aegisub manually as well as opening it from the command line.

vadz:
Thanks, I'm perfectly fine with applying this version too (FWIW we already have a global `gs_alreadySetSwapInterval` in the same file, so it's too late to worry about the code not being ugly... but OTOH I don't think we can avoid having this one ).

If there are no comments I'm going to merge it in a couple of days, backport to 3.2 and think about releasing 3.2.4.

Thanks again for fixing and testing this!

vadz:
Sorry, I forgot to apply this one to master, so let me do this already and I'll backport it to 3.2 later when we're happy with the situation in master.

Thanks again for debugging and fixing this!

PR #23971, opened 2023-10-16
Adds missing wxItemContainer API doc.

mcorino:
This method is needed to determine the sorted state of a control derived from wxItemContainer without knowing the
actual derived control class.


vadz:
Thanks, will also merge this one soon.

PR #23965, opened 2023-10-14
Make wxMSW date/time picker controls locale aware

utelle:
- Add helper function to retrieve native format specs from wxUILocale
- Add function declaration of the helper function in msw/private/uilocale.h
- Adjust code for date/time picker to set native format specification

PR #23972, opened 2023-10-16
Document wxUILocale::UseLocaleName()

utelle:
Under certain circumstances it can make sense to use a locale other than the default user locale.

One example would be, if an application wants to use a UI translation with an RTL language (like Arabic) on a non-RTL system. This wouldn't work with the default locale, because the layout direction would be wrong.

Currently setting a non-default locale can already be achieved via wxLocale anyway. Therefore wxUILocale::UseLocaleName() should be made public, to allow for less side effects than using wxLocale.

### DIFF
--- a/interface/wx/uilocale.h
+++ b/interface/wx/uilocale.h
@@ -100,6 +100,41 @@ public:
     static const wxUILocale& GetCurrent();
 
     /**
+        Configure the UI to use the locale corresponding to the given locale name tag.
+
+        If localized applications use this function instead of the recommended
+        function wxUILocale::UseDefault(), it should be called as early as possible
+        during the program startup, e.g. in the very beginning of the overridden
+        wxApp::OnInit().
+
+        @param localeName
+            The locale name tag for which the corresponding locale should be created.
+            Example: "de_DE.UTF-8" - German locale name in POSIX notation.
+            See wxLocaleIdent::FromTag() for more information about the syntax of
+            the @a locale tag string.
+
+        Note that under most Unix systems (but not macOS) this function changes
+        the C locale to the locale specified by the environment variables and
+        so affects the results of calling C functions such as @c sprintf() etc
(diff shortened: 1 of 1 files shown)


wh11204:
You should remove also this comment from uilocale.h:

    // This is a compatibility function used by wxWidgets itself, don't use it
    // in the new code.

wh11204:
I use this method in my applications because IMHO changing the language should change the locale accordingly. An english user in a spanish computer may prefer using the program in english with the english locale, not with the spanish one.

vadz:
I'll merge this, but shouldn't we be more clear the limitations of this function under macOS, i.e. say that it doesn't affect many parts of the UI there (AFAICS it will only affect the date picker, once #23956 is merged)? OTOH we probably shouldn't say that it doesn't affect date picker under Windows if #23965 is merged, right?

utelle:
> I'll merge this, but shouldn't we be more clear the limitations of this function under macOS, i.e. say that it doesn't affect many parts of the UI there (AFAICS it will only affect the date picker, once #23956 is merged)?

Under macOS numbers, date and time will be formatted according to the selected locale. AFAICR only some system dialogs like file selection can't be fully localized if a non-default locale was selected. Maybe some system entries in the main menu could be localized, too, but I'm not a macOS expert.

> OTOH we probably shouldn't say that it doesn't affect date picker under Windows if #23965 is merged, right?

Maybe minor adjustments to the note should be made, after PRs #23956 and #23965 have been merged. However, the note already says that a full localization is impossible for a non-default locale. But IMHO it is better than doing nothing.

utelle:
> But, FWIW, localizing the "Apple" menu items is definitely important under macOS.

I don't remember all the details, but under macOS especially some system menu entries were not localized. Unfortunately, my access to macOS is currently limited. If that changes in the future I will try to look into this.

PR #23973, opened 2023-10-17
Fix wxContextMenuEvent event propagation under wxQt

AliKet:
Also fixes this test case:
https://github.com/wxWidgets/wxWidgets/blob/7fbe3698c5e3eb89693e8a4e9f54e7f89d69ec87/tests/events/propagation.cpp#L655-L697


vadz:
Thanks, merging this one too.

PR #23980, opened 2023-10-20
Fix missing doc for new wxPGSelectPropertyFlags enum

mcorino:


interface/wx/propgrid/propgridpagestate.h
```diff
@@ -6,6 +6,50 @@
...
 
 
+enum class wxPGSelectPropertyFlags : int
+{
+    /**
+      No flags
+    */
+    Null          = 0,
+    /**
+      Focuses to created editor
```
a-wi, line 16: It would be good to add @hideinitializer attribute to all descriptions.

vadz:
Could the documentation of enum itself be added please, as I don't really know what/where is it used for?

It should also have `@since 3.3.0` in it.

TIA!

mcorino:
> Could the documentation of enum itself be added please, as I don't really know what/where is it used for?
> 
> It should also have `@since 3.3.0` in it.
> 

I think that would be a good idea. I suggest the original author of the enum does that because I lack the same knowledge. I only noticed the docs were lacking.

vadz:
Unfortunately I don't know this neither. @a-wi Could you please explain this? TIA!

mcorino:
As there does not seem to be any useful comments forthcoming from the original author would it suffice if I added the following:

```
/**
    Property selection flag values providing information about including/excluding operations of selected properties.
 */
```

This is as much as I can gather from the code using these flags.
These seem to be more for internal use anyway (and were not documented AT ALL before 3.3) but since they are part of the public interface will require documentation to include wxRuby (and probably wxPython).

vadz:
I've looked at this and I think the only public function this enum is used with is `wxPropertyGrid::CommitChangesFromEditor()`, so I'd rather mention this instead.

This might be not the best documentation, but I agree that it's probably not worth waiting for more.

Thanks for documenting it!

PR #23983, opened 2023-10-22
Implement wxSearchCtrl::GetEditHWND() correctly in wxMSW

### LINKED ISSUE #23975: wxSearchCtrl::SetHint() generates an unnecessary assert in 3.3
### Description

Commit 4d76a87 changes srchctrl.h by adding the following:

```C++
#if defined(__WXMSW__) && !defined(__WXUNIVERSAL__)
    private:
        // implement this to return the HWND of the EDIT control
        // can return nullptr here as this TextEntry is just a proxy
        virtual WXHWND GetEditHWND() const override  { wxFAIL_MSG("unreachable"); return nullptr; }
#endif
```

On Windows in a Debug build, if you then call `wxSearchCtrl::SetHint("some text")` you hit the wxFAIL_MSG() assertion. Click continue and the hint is correctly set. That's because wxTextEntry::SetHint uses:

```C++
    if ( wxUxThemeIsActive() )
    {
        if ( ::SendMessage(GetEditHwnd(), EM_SETCUEBANNER,
                             TRUE, (LPARAM)(const wchar_t *)hint.wc_str()) )
            return true;
    }

    return wxTextEntryBase::SetHint(hint);
```

SendMessage will have returned FALSE because the window handle is a nullptr, but that just drops out and calls wxTextEntryBase::SetHint which works fine. All of which means that the assertion makes it _seem_ like you shouldn't call SetHint() under __WXMSW__, but in fact it works fine on 3.3 just as it did in 3.2 which didn't have the assertion. While I see the advantage of providing a virtual method to get the edit control, I don't see why it should generate an assertion dialog if it wasn't overridden.

Of course calling `::SendMessage` with a nullptr for a handle doesn't make sense, but you can't even check for a nullptr first without calling `GetEditHwnd()` which generates the assertion. I.e., without the assertion check, one could write:

```C++
    if ( wxUxThemeIsActive() )
    {
        if ( auto hwnd = GetEditHwnd(); hwnd && ::SendMessage(hwnd, EM_SETCUEBANNER,
                             TRUE, (LPARAM)(const wchar_t *)hint.wc_str()) )
            return true;
    }

    return wxTextEntryBase::SetHint(hint);
```

Or something like that -- I forget which c++ standard allowed that variation of an `if()` statement...

vadz:
Since the changes of 4d76a87015 (Make wxSearchCtrl inherit from wxTextEntry in all ports, 2023-07-09, see #23697) GetEditHWND() returned null for wxSearchCtrl which was considered to be correct because all wxTextEntry functions were supposed to be forwarded to wxSearchTextCtrl anyhow.

This wasn't really correct, however, as GetEditHWND() was also called from SetHint(), resulting in an assert failure and falling back on generic hint implementation which, while mostly working, suffers from some problems that the native implementation doesn't have.

So make this function return the actual HWND used by the associated text entry to fix this and allow SetHint() to work as before and also forestall any problems due to not having the right HWND in the future.

Closes #23975.

### DIFF
--- a/include/wx/generic/srchctlg.h
+++ b/include/wx/generic/srchctlg.h
@@ -175,6 +175,12 @@ protected:
 private:
     friend class wxSearchButton;
 
+#if defined(__WXMSW__) && !defined(__WXUNIVERSAL__)
+    // Implement wxMSW-specific pure virtual function by forwarding it to the
+    // real text entry.
+    virtual WXHWND GetEditHWND() const override;
+#endif
+
     // Implement pure virtual function inherited from wxCompositeWindow.
     virtual wxWindowList GetCompositeWindowParts() const override;
 
--- a/include/wx/srchctrl.h
+++ b/include/wx/srchctrl.h
@@ -27,12 +27,6 @@
         : public wxCompositeWindow< wxNavigationEnabled<wxControl> >,
           public wxTextEntry
     {
-#if defined(__WXMSW__) && !defined(__WXUNIVERSAL__)
-    private:
-        // implement this to return the HWND of the EDIT control
-        // can return nullptr here as this TextEntry is just a proxy
-        virtual WXHWND GetEditHWND() const override  { wxFAIL_MSG("unreachable"); return nullptr; }
-#endif
     };
 #elif defined(__WXMAC__)
     // search control was introduced in Mac OS X 10.3 Panther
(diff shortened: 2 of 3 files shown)


Randalphwa:
I reviewed the changes in the commit and it looks fine, and should indeed solve the problem.

vadz:

PR #23989, opened 2023-10-23
Allow header files to be included multiple times (within a single translation unit).

DoctorNoobingstoneIPresume:
Thank you for considering this and, generally, for your work on wxWidgets !! 🤗 ❤️


lanurmi:
Is there some evidence that including multiple times isn't already possible without this commit?

imciner2:
The headers modified in this PR are in the `interface` folder. These headers are only made for the documentation generation and should not be actually used by developers in their application. To build against wxWidgets, you need to use the headers in the `include` directory instead. There is no need for these modifications.

DoctorNoobingstoneIPresume:
Thank you very much for kind clarifications. I now understand and I agree that there is no need for this change.

Issue #23981, opened 2023-10-20
Crash while initialising a console app


gwilymk:
### Description

Spotted while trying to run `kicad-cli` and found under the debugger that the crash comes from within wxwidgets.

The following code crashes:
```cpp
#include <wx/wx.h>

class App : public wxAppConsole
{
public:
	bool OnInit() override
	{
		return true;
	}
};

IMPLEMENT_APP_CONSOLE(App);
```

compiled simply with
```sh
g++ `wx-config --cxxflags` main.cpp -o main `wx-config --libs`
```

with the following error:

```
(process:37226): GLib-GObject-CRITICAL **: 19:36:01.730: g_object_get: assertion 'G_IS_OBJECT (object)' failed
free(): invalid pointer
fish: Job 1, './main' terminated by signal SIGABRT (Abort)
```

backtrace:

```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007ffff6aac8a3 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007ffff6a5c668 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff6a444b8 in __GI_abort () at abort.c:79
#4  0x00007ffff6a45390 in __libc_message (fmt=fmt@entry=0x7ffff6bbc55d "%s\n") at ../sysdeps/posix/libc_fatal.c:150
#5  0x00007ffff6ab67b7 in malloc_printerr (str=str@entry=0x7ffff6bba02b "free(): invalid pointer") at malloc.c:5765
#6  0x00007ffff6ab8a74 in _int_free (av=<optimized out>, p=<optimized out>, have_lock=have_lock@entry=0) at malloc.c:4500
#7  0x00007ffff6abb353 in __GI___libc_free (mem=<optimized out>) at malloc.c:3391
#8  0x00007ffff79cabbd in UpdatePreferDark(GVariant*) (value=value@entry=0x7fffe00025d0) at /usr/src/debug/wxwidgets/wxWidgets-3.2.3/src/gtk/settings.cpp:201
#9  0x00007ffff79cad0a in wxSystemSettingsModule::OnInit() (this=<optimized out>) at /usr/src/debug/wxwidgets/wxWidgets-3.2.3/src/gtk/settings.cpp:1235
[...]

PR #23990, opened 2023-10-23
Add a test for running wxGTK GUI apps in console mode

vadz:
See #23981.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -344,6 +344,11 @@ jobs:
             exit $rc
           fi
 
+      - name: Testing launching GUI test
+        working-directory: tests
+        run: |
+          WX_TEST_DISABLE_GUI=1 ./test_gui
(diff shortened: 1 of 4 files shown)


vadz:
I've checked that running the test with `WX_TEST_DISABLE_GUI=1` using pre-4cabfa111a (Avoid crash with GTK3 if console program is linked to GUI library, 2023-10-21) version of `src/gtk/settings.cpp` crashes, as expected, but that it works now.

vadz:
Amazingly, as I've discovered while testing this PR, I've already fixed a similar problem in f18b415ee3 (Don't crash on startup of console programs in monolithic wxX11 build., 2011-05-03) and if I had added a unit test back then, we wouldn't have had #23981 today.

We really should adopt "Melius sero quam numquam" as our official motto.

paulcor (approved):

PR #23991, opened 2023-10-23
Fix wxDirectFB build with C++17

vadz:
Not sure if anybody is using it, but fixing this was easy enough, so just do it for now.

### DIFF
--- a/include/wx/dfb/wrapdfb.h
+++ b/include/wx/dfb/wrapdfb.h
@@ -14,8 +14,13 @@
 #include "wx/gdicmn.h"
 #include "wx/vidmode.h"
 
+// "register" is removed in C++17 but used inside these headers.
+#define register
+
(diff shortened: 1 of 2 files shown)


PaavoHelde:
Thanks, I just stumped on the CopyBits failure in my build!

vadz:
Out of curiosity, do you actually use wxDFB? It would add at least some motivation to keep maintaining it if we knew it still had any users.

PaavoHelde:
Yes, sure, years ago this was the only way I was able to figure out how to compile a minimal wxWidgets library on Linux without dragging in excessive X11 support. We are using it just for building the wxsvg library, and this now sits in our codebase. I know next to nothing about wxWidgets, it might be some other approach is better for preparing wxWidgets (or for rendering SVG files, for that matter).

PR #23992, opened 2023-10-24
Set default button in StdDialogButtonSizer::Realize()

bigpilot:
Currently the Default button (the button which is activated when an Enter button is pressed) isn't set.

### DIFF
--- a/src/common/sizer.cpp
+++ b/src/common/sizer.cpp
@@ -3157,6 +3157,7 @@ void wxStdDialogButtonSizer::Realize()
 
         if (m_buttonAffirmative){
             Add((wxWindow*)m_buttonAffirmative, 0, wxALIGN_CENTRE | wxLEFT | wxRIGHT, m_buttonAffirmative->ConvertDialogToPixels(wxSize(4, 0)).x);
+            m_buttonAffirmative->SetDefault();
             tabOrder.Add(m_buttonAffirmative);
         }
 
--- a/src/common/sizer.cpp
+++ b/src/common/sizer.cpp
@@ -3112,6 +3112,7 @@ void wxStdDialogButtonSizer::Realize()
 
         if (m_buttonAffirmative){
             Add((wxWindow*)m_buttonAffirmative, 0, wxALIGN_CENTRE | wxLEFT | wxRIGHT, m_buttonAffirmative->ConvertDialogToPixels(wxSize(2, 0)).x);
+            m_buttonAffirmative->SetDefault();
             tabOrder.Add(m_buttonAffirmative);
         }
 
@@ -3157,7 +3158,6 @@ void wxStdDialogButtonSizer::Realize()
 
         if (m_buttonAffirmative){
             Add((wxWindow*)m_buttonAffirmative, 0, wxALIGN_CENTRE | wxLEFT | wxRIGHT, m_buttonAffirmative->ConvertDialogToPixels(wxSize(4, 0)).x);
-            m_buttonAffirmative->SetDefault();
             tabOrder.Add(m_buttonAffirmative);
         }
 


imciner2:
I don't think we should set a default button in the sizer, because the platform guidance specifies different defaults depending on the action being prompted about. It is better to have the user set the default based on their usage of the dialog and what is best for them.

For instance, the GTK guidelines say (https://developer.gnome.org/hig/patterns/feedback/dialogs.html#general-guidelines):
> Assign the return key to activate the affirmative button. However, this should not be done if its action is irreversible, destructive or otherwise inconvenient to the user. If there is no appropriate button to designate as the default button, do not set one.

bigpilot:
And how about making this platform specific? On Windows it's etiquette to have a Default button set. As a user (developer) I would expect this to be set on Windows.

I don't know what the Mac OS X user interface guidelines say, though.

dconnet:
On Windows, you still have to set which button is default. (In the resource file with `DEFPUSHBUTTON`)

PBfordev:
Based on the OP's thread on wxForum, I think their goal is to mimic what [`wxDialog::CreateStdDialogButtonSizer()`](https://github.com/wxWidgets/wxWidgets/blob/dc6a0c069bdde714f22cb0459efd9ad186ce7179/src/common/dlgcmn.cpp#L271) does in regards to the default button.

In other word, after calling e.g. `CreateStdDialogButtonSizer(wxOK | wxCANCEL)`, the OK button is the default dialog button (and has initial focus) which is not the case when using `wxStdDialogButtonSizer` directly, regardless of how the buttons are added there. Of course, one can always call `SetDefault()` on any button manually...

I have no idea whether there is an intent behind this difference (likely) or it just happened like that.

bigpilot:
I modified my PR only adding the SetDefault() on Windows and removing it for GTK+. 

I tested it on Windows (win32 build) and it works as expected.

vadz:
Sorry, I agree with the others, it is not a good idea to make the affirmative button default by default (sic). Just imagine that your dialog is "Launch the missiles?" with "Yes/No/Cancel" as possible answers and you would probably agree that making "Yes" the default is not the safest choice.

You can use `wxXXX_DEFAULT` flags of `CreateStdDialogButtonSizer()` or just call `GetAffirmativeButton()->SetDefault()` manually. In principle, we could add `wxStdDialogButtonSizer::SetDefaultButton()` but to me even this doesn't seem really necessary.

PBfordev:
> Sorry, I agree with the others, it is not a good idea to make the affirmative button default by default (sic). Just imagine that your dialog is "Launch the missiles?" with "Yes/No/Cancel" as possible answers and you would probably agree that making "Yes" the default is not the safest choice.
> 
> You can use `wxXXX_DEFAULT` flags of `CreateStdDialogButtonSizer()` 

Just for the record, the OK or Yes buttons are automatically [set as default](https://github.com/wxWidgets/wxWidgets/blob/dc6a0c069bdde714f22cb0459efd9ad186ce7179/src/common/dlgcmn.cpp#L331) (convenient but undocumented) when using `wxDialog::CreateStdDialogButtonSizer()`. The only `wxXXX_DEFAULT` flag the method accepts is somewhat the opposite `wxNO_DEFAULT`.

vadz:
Yes, because `wx{OK,YES}_DEFAULT` are both 0, i.e. this is the default behaviour. But you can still change it with `wxNO_DEFAULT`, i.e. it's not unconditional. Arguably, we should have `wxNONE_DEFAULT` or something like this too.

PBfordev:
> Yes, because `wx{OK,YES}_DEFAULT` are both 0, i.e. this is the default behaviour. 

These styles are not listed in the allowed flags in `wxDialog::CreateStdDialogButtonSizer()` docs and AFAICT are not referenced anywhere outside `wxMessage{Box|Dialog}` docs.  I, for one, had no idea they would apply to `wxDialog::CreateStdDialogButtonSizer()` as well. 

Once again, I am not saying the current behavior is wrong but perhaps it could be mentioned in the docs (particularly if there is an argument that automatically setting a default YES/OK button is dangerous).

vadz:
Well, they "apply to it" somewhat tautologically: as they don't anything, they also don't do anything in it too. But yes, it would still be probably useful to mention them in the docs, personally I find that using them makes the code more readable as it makes the intention explicit.

PR #23998, opened 2023-10-25
Make using EGL under Wayland from outside wxEVT_PAINT work again

### LINKED ISSUE #23996: Repainting `wxGLCanvas` from `wxEVT_IDLE` doesn't work in 3.2.3
I took a quick look at why even with the crash fix from #23968, on wxWidgets 3.2.3 moving the mouse inside the video display in Aegisub (which is supposed to render a crosshair centered at the mouse position) is very sluggish.

It turns out that Aegisub does not render the canvas on `wxEVT_PAINT`, but *only* on `wxEVT_IDLE`. But due to the changes from #23554, most attempts to draw on `wxEVT_IDLE` end up doing nothing (due to the flag `m_readyToDraw` being set to false at the time the idle event comes). This test application shows a big difference in fluidity between wxWidgets 3.2.2 and 3.2.3:

```c++
#include <wx/app.h>
#include <wx/glcanvas.h>
#include <wx/dcclient.h>
#include <wx/frame.h>
#include <wx/button.h>
#include <wx/sizer.h>

class MyWxApp : public wxApp
{
	double normX = 0.5;

public:
	bool OnInit() override {
		auto canvas = new wxGLCanvas(window);
		auto context = new wxGLContext(canvas);
		canvas->Bind(wxEVT_IDLE, [this, canvas, context](wxIdleEvent &evt) {
			fprintf(stderr, "Canvas paint\n");
			context->SetCurrent(*canvas);

			glClearColor(normX, 0.0f, 0.0f, 1.0f);
			glClear(GL_COLOR_BUFFER_BIT);

			canvas->SwapBuffers();
		});

		canvas->Bind(wxEVT_MOTION, [this, canvas](wxMouseEvent &evt) {
			fprintf(stderr, "Mouse move\n");
			normX = (double)evt.GetPosition().x / canvas->GetSize().GetWidth();
		});

		window->Layout();
		window->Show(true);
		SetTopWindow(window);
		return true;
	}
};

IMPLEMENT_APP(MyWxApp)
```

_Originally posted by @joanbm in https://github.com/wxWidgets/wxWidgets/issues/23961#issuecomment-1776209016_

vadz:
Since the changes of 194a7be33f (Improve wxGLCanvasEGL refresh logic
under Wayland, 2023-05-17, see #23554) drawing on wxGLCanvas from
wxEVT_IDLE handler without refreshing the window first didn't work any
longer because the "ready to draw" flag was always false in this case.

To make this work again, refresh the window ourselves to ensure that it
does get redrawn. This is less efficient than drawing on it from
wxEVT_PAINT because all the code executed by the application prior to
calling SwapBuffers() is just wasted, but better than not updating the
window at all.

The recommended way of doing things remains, just as it always was, to
draw from wxEVT_PAINT and call Refresh() in the application code when
the window needs to be updated.

Closes #23996.

### DIFF
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -799,6 +799,12 @@ bool wxGLCanvasEGL::SwapBuffers()
         if ( !m_readyToDraw )
         {
             wxLogTrace(TRACE_EGL, "Window %p is not not ready to draw yet", this);
+
+            // When the application draws the window from a timer or idle event
+            // handler, we need to ensure that it's going to be called again if
+            // we didn't do anything this time, so request another repaint.
+            Refresh();
+
             return false;
         }
 


vadz:

PR #24001, opened 2023-10-25
Expand wxDialog::CreateStdDialogButtonSizer docs

PBfordev:
Add a note explaining that when creating a button sizer with this method, one of the buttons will be usually made the default.

BTW, I wonder whether the notes in other button-sizer-creating `wxDialog` methods such as this
https://github.com/wxWidgets/wxWidgets/blob/dc6a0c069bdde714f22cb0459efd9ad186ce7179/interface/wx/dialog.h#L249-L253
are still relevant.


vadz:
Thanks for adding this!

I don't think any PDA-related stuff is still relevant, but OTOH it might become again in the future if we support more devices with physical buttons?

Issue #23893, opened 2023-09-22
wxOSX Cocoa: A problem with starting the application when using UIElement.


DiCode77:
#### Platform and version information

- wxWidgets version  you use: 3.2.2.1
- wxWidgets port  you use: wxOSX, cocoa.
- OS  and its version: macOS(14.0 (23A344)) Sonoma RC1/2.
- IDE: Xcode 15.0 (15A240d).

I would like to note in advance that this error does not exist on macOS Ventura or older systems, the problem appears starting with macOS Sonoma.

# 

#### Description of the problem

The problem is that when I use (UIElement == true), the app only starts on the second try (the UI is rendered).

That is, after the first attempt, only a process is created in the system, but the program does not show signs that it has started.

<img width="951" alt="1" src="https://github.com/wxWidgets/wxWidgets/assets/109686137/b414352f-4fd6-4aad-b806-9bb60ce152e0">

After the second attempt to start the program, the same process starts working, the program interface appears.

<img width="954" alt="2" src="https://github.com/wxWidgets/wxWidgets/assets/109686137/7473ec9d-131d-4b96-b0e9-2a0e2ea5d321">

I want to emphasize that it is not important how you try to launch the program, either with the cursor, or by adding it to the autorun in the system.

With UIElement disabled, the application starts on the first try.

#

#### Trying to identify the problem.

I will note right away that if the application is debugged or only launched through the IDE (Xcode), then the problem does not manifest itself, that is, the launch occurs on the first try.

By the method of research, I understood that the problem is that the `virtual bool OnInit()` method is not called.
That is, for example, the `MyApp()` constructor works in this class, but the `OnInit` method does not.

<img width="366" alt="3" src="https://github.com/wxWidgets/wxWidgets/assets/109686137/c00980b1-eba7-4256-94b3-d580ba44b4d2">

<br> 

That is, we can conclude that when we pass the `MyApp` class to the macro, after creating the object, the `OnInit()` method is not called in wxWidgets itself, and this is the problem.
<br> 
#

#### Workaround method

The first option is to call the `OnInit()` method in the `MyApp()` constructor, you just need to change the OnInit name, otherwise two application interfaces can be created at once.

<img width="209" alt="4" src="https://github.com/wxWidgets/wxWidgets/assets/109686137/a75eb93b-4ef1-4b5c-9e4c-dc25e09c2462">

<br> 
[...]

PR #24003, opened 2023-10-27
macOS Sonoma Background Agent fix

csomor:
see #23893

### DIFF
--- a/src/osx/cocoa/utils.mm
+++ b/src/osx/cocoa/utils.mm
@@ -108,8 +108,9 @@ void wxBell()
             activate = true;
         }
 
-        if ( activate )
-            [NSApp activateIgnoringOtherApps: YES];
+        if ( activate ) {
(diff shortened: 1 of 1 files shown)


csomor:
yes, definitely, that code works back to 10.6 so no problems here ...

ok, so I'll add that line to changes.txt in master directly I guess now that I've merged the PR ?

vadz:
> ok, so I'll add that line to changes.txt in master directly I guess now that I've merged the PR ?

No, sorry, it's more complicated than this, see the note in master version of `docs/changes.txt`. In 3.2 I update it directly however as there are fewer merges (which used to regularly create conflicts if both of them touch the change log), so I'll just add it there when cherry-picking -- please just let me know how would you prefer to describe this. TIA!

csomor:
ok, perhaps something along

wxOSX
- fixing activation (windows not shown upon launch) for background/agent apps under macOS 14 Sonoma

Issue #24002, opened 2023-10-26
Buttons drawn by wxRendererMac::DrawTreeItemButton don't respect dark mode


craftyjon:
### Description

On wxMac, the DrawTreeItemButton renderer always draws in light mode (i.e. dark strokes) even when the system is in dark mode.  I'm still trying to track down why this is, and if there's a simple fix, since the macOS APIs used are old and not really documented.

I've played around with adding a call to `HIThemeSetStroke` before `HIThemeDrawButton`, but this doesn't seem to actually change how the button is drawn (or I'm just doing it wrong).  For example (I've tried various different ThemeBrush colors with no effect)

```
ThemeBrush color = (flags & wxCONTROL_DISABLED) ? kThemeBrushButtonFrameInactive : kThemeBrushButtonFrameActive;
HIThemeSetStroke( color, nullptr, cgContext, kHIThemeOrientationNormal );
```

Perhaps the `HIThemeDrawButton` API just is hard-coded to draw in black, in which case if there is not a newer macOS API (I haven't found one yet) maybe wxOSX should just switch to a generic renderer for these buttons inside a tree control (and other places they are needed)

### Platform and version information

- wxWidgets version  you use: 3.2.3
- wxWidgets port  you use: wxOSX
- OS  and its version: macOS 13.5.1

craftyjon:
I guess the "right" path is to move to a Cocoa NSButton in disclosure mode, but I don't know enough about macOS development / Objective-C to actually attempt that myself (especially since it seems like it might even call for a whole new file in osx/cocoa, and I don't know where to start

vadz:
@csomor If you could please give a hint about how to fix this, it would be very welcome as I have no idea about this neither.

In the worst case, we'd have to use the generic function in the dark mode, as it should be less wrong than drawing the invisible native button...

csomor:
> @csomor If you could please give a hint about how to fix this, it would be very welcome as I have no idea about this neither.

I'll look into it

csomor:
ok, yes, the HIThemeDraw calls seem to respect at certain places the appearance mode, but for the disclosure button this does not work. To really clean this up we'd have to write a new renderer that eg uses NSCells that are allocated on a invisible NSWindow and convinces them to draw into the current cgContext. So for the moment I'd suggest using a generic implementation.

craftyjon:
I tried doing:

```cpp
    // HIThemeDrawButton does not respect dark mode
[...]

PR #24004, opened 2023-10-27
OSX migrate to NSCell based renderer

csomor:
the old HITheme renderer does not support dark mode for the disclosure button, we should move to use NSCells for rendering in the future, see #24002


craftyjon:
Looks good to me!  Thank you :)

image

vadz:
Stefan, is this ready to be merged? If so, I'd like to (squash) merge it removing the old files to get the correct diffs for this commit. Please let me know if I should do this, TIA!

csomor:
@vadz it's not ready, I still have a focus-clear-redraw issue I have to investigate, and I thought I'd post to wx-users to get more testing, but if this looks like it could hinder things, I could introduce a #ifdef internally to have the older code active and allow for testers to enable it on master

vadz:
IME asking people to test things really works, unfortunately. Other than those directly involved in the PR/bug report, few people test new code until it's merged into master. So if you're confident that this is the right way to do it and can fix the bugs you see, I'd merge it first and ask for testing afterwards.

craftyjon:
> Other than those directly involved in the PR/bug report, few people test new code until it's merged into master. So if you're confident that this is the right way to do it and can fix the bugs you see, I'd merge it first and ask for testing afterwards.

For what it's worth, we have this same experience with KiCad and thus also have adopted a "merge early" philosophy

csomor:
> Stefan, I've converted this to draft for now, please make it "Ready for review" when you consider it in an acceptable (even if not perfect) state. TIA!

Thanks, although you are not fond of them, I'm really thinking about adding #ifdefs with the old code - to support better A/B testing

vadz:
> Thanks, although you are not fond of them, I'm really thinking about adding #ifdefs with the old code - to support better A/B testing

If we do this, we should at least enable the new version by default. But I'd indeed prefer to avoid having them, if only because it would be better if this appeared as a change of the existing code (which will be the case if it's replaced by the new version) instead of addition of another copy of it and then removal of the original version.

vadz:
Just to be sure (I think you already know this, but, again, just in case): we can move the file, Git doesn't care about its path (nor extension) and history would still look correct in this case.

csomor:
yes, but I realized that I have started the wrong way in the branch, with the duplicate , I'll create a new one and setup this with the correct history

PR #24008, opened 2023-10-27
Fix wxTLW::SetSizeHints() implementation under wxQt

AliKet:
As documented, the value -1 is special which means no constraints will be used. But, passing -1 as `maxsize` end up with Qt making the window **non-resizable** at all (because `maxsize` will be the same as `minsize`).

In other words, we should use the Qt defaults if -1 was specified.


vadz:

PR #24017, opened 2023-10-29
Undo wxGLCanvas::SwapBuffers() Wayland frame sync.

### LINKED ISSUE #24012: [Wayland] `wxGLCanvas` always repaints continuously since 3.2.3
Since wxWidgets 3.2.3 (in particular PR #23554 / commit 0ded71ff109daae2baa83cd8e835ec775649d32c), applications using `wxGLCanvas` on Wayland (i.e. using the EGL implementation) receive a continuous stream of `wxEVT_PAINT` events, apparently at the refresh rate of the display, even if the application is not requesting a refresh and the user is not interacting with the application.

This is generally harmless from the point of view of the application functioning correctly, but causes extra CPU usage for applications that do not need to update graphics continuously.

Example application to reproduce the problem:

```c++
#include <wx/app.h>
#include <wx/glcanvas.h>
#include <wx/dcclient.h>
#include <wx/frame.h>
#include <wx/stattext.h>
#include <wx/string.h>

class MyWxApp : public wxApp
{
public:
	bool OnInit() override {
		auto counter = window->CreateStatusBar();
		auto canvas = new wxGLCanvas(window);
		auto context = new wxGLContext(canvas);
		canvas->Bind(wxEVT_PAINT, [this, canvas, context, counter](wxPaintEvent &evt) {
			fprintf(stderr, "Canvas paint\n");

			context->SetCurrent(*canvas);
			glClearColor(0.5f, 0.5f, 0.5f, 1.0f);
			glClear(GL_COLOR_BUFFER_BIT);
			canvas->SwapBuffers();

			static int frames = 0;
			counter->SetStatusText(wxString::Format(wxT("%d frames"), ++frames));
		});

		window->Layout();
		window->Show(true);
		SetTopWindow(window);
		return true;
	}
};

IMPLEMENT_APP(MyWxApp)
```

**Behaviour under wxWidgets 3.2.2.1:** The frame counter at the status bar is at a low value and is stable when no interactions are happening.
**Behaviour under wxWidgets 3.2.3:** The frame counter continuously increments.

joanbm:
This reverts the frame synchronization logic that was recently added with the purpose of avoiding performance issues due to `eglSwapBuffers` blocking when the canvas is hidden or occluded.

This logic should be unnecessary after `eglSwapInterval(display, 0)` is called, since `eglSwapBuffers` should never block anymore. Furthermore, as it stands now, it causes the canvas to continuously repaint itself at the refresh rate of the display, which is wasteful for applications which do not need to continuously refresh.

Fixes #24012

### DIFF
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -793,27 +793,13 @@ bool wxGLCanvasEGL::SwapBuffers()
 #ifdef GDK_WINDOWING_WAYLAND
     if (wxGTKImpl::IsWayland(window))
     {
-        // Under Wayland, we must not draw before we're actually ready to, as
-        // this would be inefficient at best and could result in a deadlock at
-        // worst if we're called before the window is realized.
+        // Under Wayland, we must not draw before the window has been realized,
+        // as this could result in a deadlock inside eglSwapBuffers()
         if ( !m_readyToDraw )
         {
             wxLogTrace(TRACE_EGL, "Window %p is not not ready to draw yet", this);
-
-            // When the application draws the window from a timer or idle event
-            // handler, we need to ensure that it's going to be called again if
-            // we didn't do anything this time, so request another repaint.
-            Refresh();
-
             return false;
         }
-
-        // Ensure that we redraw again when the frame becomes ready.
-        m_readyToDraw = false;
-        wl_surface* surface = gdk_wayland_window_get_wl_surface(window);
-        m_wlFrameCallbackHandler = wl_surface_frame(surface);
-        wl_callback_add_listener(m_wlFrameCallbackHandler,
-                                 &wl_frame_listener, this);
     }
 #endif // GDK_WINDOWING_WAYLAND
 


joanbm:
The idea is that after applying #24014, `eglSwapBuffers` should never block anymore, so we can get rid of #23554 (and #23998) without reintroducing the KiCad performance issue (#23512).

With #24014+this PR on top of 3.2.3, I can not reproduce the KiCad performance issue (but I can with 3.2.2.1), but I would like to run some more tests to ensure nothing breaks.

vadz:
Great, this makes things much simpler (and pretty close to the original version, actually... I'm afraid I overcomplicated things way too much due to not understanding the problem correctly).

I'm going to merge this in master as this seems to work with all the different test cases I have too (the ones for #22580, #23961, #24012 and multiple windows in the cube sample, one of which is obscured) and will backport to 3.2 relatively quickly too (as it seems this code doesn't get tested much in master anyhow).

Thanks for all your help once more!

PR #24018, opened 2023-10-29
3.2 EGL fixes backports

vadz:
This collects all the changes done in master and backports them to 3.2.

cc @joanbm

### DIFF
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -641,6 +641,21 @@ wxGLCanvasEGL::~wxGLCanvasEGL()
 void wxGLCanvasEGL::CreateWaylandSubsurface()
 {
 #ifdef GDK_WINDOWING_WAYLAND
+    // It's possible that we get in here unnecessarily in two ways:
+    // (1) If the canvas widget is shown, and then immediately hidden, we will
+    //     still receive a map-event signal, but by that point, the subsurface
+    //     does not need to be created anymore as the canvas is hidden
+    // (2) If the canvas widget is shown, and then immediately hidden, and then
+    //     immediately shown again, we will receive two map-event signals.
+    //     By the second time we get it, the subsurface will already be created
+    // Not ignoring either of the two scenarios will likely cause the subsurface
+    // to be created twice, leading to a crash due to a Wayland protocol error
+    // See https://github.com/wxWidgets/wxWidgets/issues/23961
+    if ( !gtk_widget_get_mapped(m_widget) || m_wlSubsurface )
+    {
+        return;
+    }
+
     GdkWindow *window = GTKGetDrawingWindow();
     struct wl_surface *surface = gdk_wayland_window_get_wl_surface(window);
 
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -843,21 +843,13 @@ bool wxGLCanvasEGL::SwapBuffers()
 #ifdef GDK_WINDOWING_WAYLAND
     if (wxGTKImpl::IsWayland(window))
     {
-        // Under Wayland, we must not draw before we're actually ready to, as
-        // this would be inefficient at best and could result in a deadlock at
-        // worst if we're called before the window is realized.
+        // Under Wayland, we must not draw before the window has been realized,
+        // as this could result in a deadlock inside eglSwapBuffers()
         if ( !m_readyToDraw )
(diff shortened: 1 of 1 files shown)


Manolo-ES:
After re-reading those previous threads, I still stand with **not** calling `SwapInterval`.

When you call it for a hidden or obscured window it is a workaround for Wayland and its 1 Hz weird decision (and not many GPU vendors agree; that's why such MR took so long to merge). It seems to me that WL's team see all us programmers as stupid, we must be corrected by them.

OK, drawing only at vertical blank seems the right thing to do, just because we can be saved from tearing and faster frames will not arrive to the display any how. OTOH the deactivation of vsync by `SwapInterval(...,0)` is used, for example, to benchmark the shaders. Or to get frames into some memory, even they are not shown. Or special tight rendering. In these cases `SwapBuffers` is called, and should not be dismissed.

But, if you deactivate vsync for a not-shown window, why not activate again at show-event?
And, what value do you "recover"? Perhaps the user set `1` (vsync activated), or perhaps not.
MSW (WGL) and GLX (and Apple?) have extensions (search for "swap control" at [Khronos registry](https://registry.khronos.org/OpenGL/extensions/EXT/)) to query the current value of `SwapInterval`. Some vendors provide "swap control tear" extension to allow a negative value (the driver takes control of the interval). Unfortunately, EGL has not a function nor an extension to query this value.

The thing is that playing with `SwapInterval` may well go against those users that do know what they're doing. And if it's the GPU driver who controls it, our call has no effect.

At least, WL team now provides "surface suspended/resumed" events. That's what the [SDL_commit](https://github.com/libsdl-org/SDL/commit/44536b7537fb7b2d896ea313ea19175e641ffb6e) that Vadim found handles.
IMHO that's not enough, because the event may arrive too late, when the app has already called `SwapBufferes`, stalling the CPU during a whole second because of the 1Hz thing. But less is nothing.

The `IsShownOnScreen()` function for GTK relies on `gtk_widget_get_child_visible` which GTK docs say "_This function does not check if the widget is obscured in any way._"

What are the alternatives? Looking into [Mesa code](https://gitlab.freedesktop.org/mesa/mesa/-/commit/2ae30237602ec07c63e245bc07aff3dccddb0267) perhaps we can query some member of the surface struct or from the EGLDisplay.

I hope WL team to get back from this madness. In the while, I propose to tell wx users to take this issue in consideration, and let they play with `SwapInterval` instead of wx.

Lastly, I still think that it would be a good having `bool wxGLCanvas::[Set][Get]SwapInterval(int& val)`.

vadz:
I don't think there is a realistic probability of Wayland developers changing their mind about this and in any case we don't want to fatally break all the existing wxWidgets applications using `wxGLCanvas` until they do. I.e. today it seems painfully obvious to me that we must set the swap interval to 0 ourselves.

Of course, it would be better if we could reliably detect if the window is occluded and only do it in this case and I'd welcome any PRs doing this. And I don't see anything wrong with providing the possibility to set it from the application code (although I don't see any advantage in wrapping it in a wx function when you can already just call `eglSwapInterval()` directly). But this doesn't change for the situation right now, sorry.

joanbm:
@Manolo-ES As I see it, yes, `eglSwapInterval(display, 0)` is a hack, and we'd rather not do it, but it's a hack that allows us to fix existing applications that can't cope with `eglSwapBuffers` blocking for a long time, which is something that didn't use to happen.

And it's something that is reasonably simple to implement and offers a solution for the applications people want to use, now.

So far, I haven't seen any proposed solution to avoid those long `eglSwapBuffers` blocks that works everywhere `eglSwapInterval(display, 0)` does, is reasonably simple to implement, does not have tricky edge cases and is not racy.

For example, the SDL thing you link, appears to depend on a version of the XDG Shell protocol that many Wayland compositors do not yet support. And it's not clear to me that the concept of a surface being "suspended" perfectly matches the situations under which `eglSwapBuffers` blocks for a long time. There's still the race condition where maybe the surface is hidden the instant before you call `eglSwapBuffers`. And even if it works, it's still a lot of work to understand, implement and test the entire thing.

joanbm:
I have run some tests on both GNOME Shell and Sway with KiCad / SLADE / Aegisub plus the various test applications / reproducers I have gathered during debugging and haven't been able to reproduce any of the bugs/regressions, so LGTM.

PR #24030, opened 2023-11-02
Add wxTE_NOHIDESEL to text controls that need it in sample

Blake-Madden:
For wxTextCtrls that use various rich flags (e.g., `wxTE_RICH2`, `wxTE_AUTO_URL`), `wxTE_NOHIDESEL` is also needed. Otherwise, using select all with them exhibits odd behavior. 

For example, in the "text" sample, if you do a "select all" from the context menu for the text control on the far right, then the selection doesn't go away even as you try to select other text. Adding `wxTE_NOHIDESEL` fixes that.


vadz:
This looks like a bug to me. Not sure if it's a regression, but this should be either fixed or we should always add `wxTE_NOHIDESEL` automatically internally because the current behaviour is indeed very wrong and hiding it in the sample is not enough.

vadz:
I've created a bug for it and closing this PR (but thanks for noticing this!).

PR #24036, opened 2023-11-04
Provide MinGW-specific instructions for building applications

vadz:
We had instructions for building wxWidgets with MinGW but not for building the applications using wxWidgets, so add them too now in the hope that it will be helpful to people who struggle to find the options to use for building -- or at the very least to the people who have to answer the questions about this, as they could now just point to these instructions.

---

This is not much and more improvements would be definitely welcome, but better than nothing. Or at least I hope so: please let me know if I wrote anything wrong, as I didn't really test this.

### DIFF
--- a/docs/msw/install.md
+++ b/docs/msw/install.md
@@ -470,11 +470,11 @@ need to:
 
 * Add the list of libraries to link with to the linker input. The exact list
   depends on which libraries you use and whether you built wxWidgets in
-  monolithic or default multlib mode and basically should include all the
-  relevant libraries from the directory above, e.g. `wxmsw31ud_core.lib
-  wxbase31ud.lib wxtiffd.lib wxjpegd.lib wxpngd.lib wxzlibd.lib wxregexud.lib
-  wxexpatd.lib` for a debug build of an application using the core library only
-  (all wxWidgets applications use the base library).
+  monolithic or default multi-lib mode and basically should include all the
+  relevant libraries from the directory above, e.g. `wxmsw34ud_core.lib
+  wxbase34ud.lib wxtiffd.lib wxjpegd.lib wxpngd.lib wxzlibd.lib wxregexud.lib
+  wxexpatd.lib` for a debug build of an application using the core library of
+  wxWidgets 3.4 only (all wxWidgets applications use the base library).
 
 
 For example, to compile your program with gcc using debug wxWidgets DLLs
--- a/docs/msw/install.md
+++ b/docs/msw/install.md
@@ -455,18 +455,22 @@ Here is what you need to do:
   - compiler
   - resource compiler
   include paths.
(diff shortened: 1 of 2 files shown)


docs/msw/install.md
```diff
@@ -472,6 +476,18 @@ need to:
   (all wxWidgets applications use the base library).
 
 
+For example, to compile your program with gcc using x64 debug wxWidgets DLLs
+you would need to use the following options for the compiler (and `windres`
+resource compiler):
+
+    -I$WXWIN/include -I$WXWIN/lib/gcc_x64_dll/mswud -D__WXMSW__ -DWXUSINGDLL
```
PBfordev, line 483: IME, when building wxWidgets with the bundled GCC makefile, `_x64` in the library folder name is not used (unlike with MSVC), i.e., both 32- and 64-bit libraries are by default produced in `$WXWIN/lib/gcc_{lib|dll}` folder.

`_x64` in the folder name is used when building with CMake or with official pre-built binaries.

I would expect people using the makefile have wxWidgets also built by the makefile. I think it would be easiest to just remove "x64" from the text and `_x64`  from the examples?

vadz: You're absolutely right, thanks for noticing this. I'll fix it and push.

PBfordev:
As for the libraries names passed to the linker, I think they should be documented better.

For experienced users, we know what to use, i.e., can derive the names of libraries from the docs and file names. But in my long-time experience on wxForum, many beginners struggle with this. Sure, one can be told to study the samples makefiles (that's what I did), but...

Link order with GCC is another similar thing, the correct order may be difficult to figure out, even with the library dependency scheme available. Lastly, the Windows libraries to link with in the static build are probably not easy to find out.

AFAICT, the library names (and their order with GCC) are not documented anywhere also for CMake.

I do not know how to efficiently (i.e., very briefly yet clearly) describe the above, it takes three pages in [my guide](https://github.com/PBfordev/wxpbguide) (but it's basically just one big table and two screenshots).  

Perhaps splitting the install instructions into two documents, (1) Building wxWidgets and (2) Building applications, would allow for more extensive guide... I could have a try at (2), if you think it could be useful. On the other hand, one can argue that the longer the docs, the slimmer the chance most people want to read them.

Even if non-CMake builds are not going to be available since 3.3, 3.2 will be here for the long time, so I think these instructions are still valuable.

vadz:
I think splitting the documentation into 2 topics could be better, but I'm not sure. It would also be nice to have the instructions directly in the binaries README but they almost totally duplicate the normal instructions (but can be much more brief as they don't have to deal with all the different cases).

For now I'll merge this as it's already a (slight) improvement IMHO but our build documentation is clearly still suboptimal, considering the amount of people who have trouble with it.

PR #24037, opened 2023-11-05
Return true from AddCatalog() if message ID matches language

### LINKED ISSUE #24019: wxTranslations::AddCatalog() returns false for msgid language if it's different from system language
In wxWidgets-3.2.3 there is a regression that was introduced with commit d1782237458a92d084ba67afb063e703496aeab2.

`wxTranslations::AddCatalog()` now returns false if the selected language:
* is that of the "msgid" strings in source code AND
* is different to the system language AND
* has no available catalog (which is perfectly fine for the msgid language)

This issue can be reproduced in the internat sample with the following modifications:

```patch
diff --git "a/samples/internat/internat.cpp" "b/samples/internat/internat.cpp"
index 53ca7ebbb1..d575fccc93 100644
--- "a/samples/internat/internat.cpp"
+++ "b/samples/internat/internat.cpp"
@@ -212,6 +212,14 @@ bool MyApp::OnInit()
         langDesc = langInfo ? langInfo->Description
                             : wxString("the default system language");
 
+    const wxLanguage msgIdLang = wxLANGUAGE_ENGLISH_US;
+
+    const wxLanguageInfo* const
+            msgIdLangInfo = wxUILocale::GetLanguageInfo( msgIdLang );
+    const wxString
+            msgIdLangDesc = msgIdLangInfo ? msgIdLangInfo->Description
+                                          : wxString("the msgid language");
+/*
     if ( m_setLocale == Locale_Ask )
     {
         m_setLocale = wxMessageBox
@@ -227,7 +235,7 @@ bool MyApp::OnInit()
     }
 
     if ( m_setLocale == Locale_Set )
-    {
+*/  {
         if ( !wxUILocale::UseDefault() )
         {
             wxLogWarning("Failed to initialize the default system locale.");
@@ -245,13 +253,14 @@ bool MyApp::OnInit()
 
         // Create the object for message translation and set it up for global use.
         wxTranslations* const trans = new wxTranslations();
+        trans->SetLanguage( msgIdLang );
         wxTranslations::Set(trans);
 
         // Initialize the catalogs we'll be using.
-        if ( !trans->AddCatalog("internat") )
+        if ( !trans->AddCatalog("internat", msgIdLang ) )
         {
             wxLogError(_("Couldn't find/load 'internat' catalog for %s."),
-                       langDesc);
+                       msgIdLangDesc);
         }
 
         // Now try to add wxstd.mo so that loading "NOTEXIST.ING" file will produce

```

- wxWidgets version  you use: 3.2.3

vadz:
When the original messages language matches the language of the locale being used, these strings can be used directly and so AddCatalog() should still return true for them but it didn't do it any more after the changes of 94b1a17aeb (Add AddAvailableCatalog() and use it in AddStdCatalog(), 2023-09-30).

See #18227.

Closes #24019.

### DIFF
--- a/src/common/translation.cpp
+++ b/src/common/translation.cpp
@@ -1346,9 +1346,27 @@ bool wxTranslations::AddCatalog(const wxString& domain,
         return true;
 
     const wxString msgIdLang = wxUILocale::GetLanguageCanonicalName(msgIdLanguage);
-    const wxString domain_lang = GetBestTranslation(domain, msgIdLang);
 
-    if ( msgIdLang == domain_lang )
+    // Check if the original strings can be used directly.
+    bool canUseUntranslared = false;
+    if ( m_lang.empty() )
+    {
+        // If we are using the default language, check if the message ID
+        // language is acceptable for this system.
+        const wxString domain_lang = GetBestTranslation(domain, msgIdLang);
+
+        if ( msgIdLang == domain_lang )
+            canUseUntranslared = true;
+    }
+    else // But if we have a fixed language, we should just check it instead.
+    {
+        // Consider message IDs for another region using the same language
+        // acceptable.
+        if ( msgIdLang.BeforeFirst('_') == m_lang.BeforeFirst('_') )
+            canUseUntranslared = true;
+    }
+
+    if ( canUseUntranslared )
     {
         wxLogTrace(TRACE_I18N,
                     wxS("not using translations for domain '%s' with msgid language '%s'"),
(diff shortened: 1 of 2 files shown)


wh11204:
Typo, _canUseUntranslared_ should be _canUseUntranslated_.

vadz:
> Typo, _canUseUntranslared_ should be _canUseUntranslated_.

Thanks, fixed locally but not updating the PR to avoid rerunning the CI builds without good reason.

PR #24040, opened 2023-11-06
Generate DPI changed event in wxGTK too

vadz:
Non-GTK changes here are just refactoring.

@paulcor Please let me know if you object to this naive implementation (I think the proper way would be to react to the property change, but this seems much simpler to me).


src/gtk/toplevel.cpp
```diff
@@ -34,6 +34,10 @@
 #include "wx/recguard.h"
 #include "wx/sysopt.h"
 
+#ifdef __WXGTK3__
+    #include "wx/display.h"
```
paulcor, line 38: Doesn't seem to be needed.

src/gtk/toplevel.cpp
```diff
@@ -339,6 +343,21 @@ gtk_frame_configure_callback( GtkWidget*,
...
+    const auto newScaleFactor = GetContentScaleFactor();
+    if ( newScaleFactor != m_scaleFactor )
+    {
+        const auto oldScaleFactor = m_scaleFactor;
+
+        // It seems safer to change it before generating the events to avoid
+        // any chance of reentrancy.
+        m_scaleFactor = newScaleFactor;
+
+        WXNotifyDPIChange(oldScaleFactor, newScaleFactor);
```
paulcor, line 357: This is going to fire when a TLW is created on a scaled display, even though it's not really a "change". Not sure if that is your intention.

src/gtk/toplevel.cpp
vadz: Oops, thanks for noticing this, it's a leftover from the version before Mac refactoring, will remove.

src/gtk/toplevel.cpp
vadz: Yes, this is definitely my intention and, in fact, the main motivation for this change: without it, the (generic) `wxTreeCtrl` in the sample is broken on high DPI display because it uses scale factor of 1 which is in effect when it's being initialized and its images are set. Getting this event makes everything work correctly, even if not optimally -- but I have no idea how could we get the correct scale factor before showing the window and I'm not sure if it's even possible.

paulcor: GTK will make a guess for the scale of a window that is not realized yet, which presumably would always be correct for a single monitor or multiple monitors with the same scale, and if wrong would be corrected here. So I think adding `m_scaleFactor = GetContentScaleFactor();` to the end of `Create()` would avoid the extra event most of the time.

vadz: Thanks, you're absolutely right and I should have realized that the fact that it doesn't work in my case (I have no idea why does GTK consider my 100% DPI monitor primary even if I have the main Gnome stuff on my 200% monitor) doesn't mean that it never works -- in fact it probably does in the vast majority of cases. I've implemented your suggestion now, thanks again.

src/gtk/toplevel.cpp
```diff
@@ -675,6 +690,13 @@ bool wxTopLevelWindowGTK::Create( wxWindow *parent,
     gdk_set_program_class(wxTheApp->GetAppDisplayName().utf8_str());
 #endif
 
+    // This value may be incorrect as here it's just set to the scale factor of
+    // the primary monitor, but it's better than setting it to 1, as chances
+    // are that the window will be created on the primary monitor or, maybe,
+    // the other monitors use the same scale factor anyhow. And if this isn't
+    // the case, it will corrected when we get "configure-event" from GTK.
+    m_scaleFactor = GetContentScaleFactor();
```
paulcor, line 698: This is too early, `m_widget` needs to be valid.

vadz: Ugh, yes, of course, thanks again.

PR #24041, opened 2023-11-08
Don't find hidden items when using incremental tree search

vadz:
Only the last commit here is important, the rest is just some minor cleanup.

I don't know if anybody has any objections to this, i.e. likes the current behaviour, but personally it drives me crazy because, as a user, I definitely want to go to the visible item when typing a first couple of the letters in its name and not some item 4 levels deep in the collapsed part of the tree.


vadz:
No objections, so merging.

Issue #24021, opened 2023-10-30
wxWidgets does not appear to show properties for desktop environment or monitor properly under Wayland


joshua-stone:
### Description

Properties regarding desktop environment and monitor appear to be missing when invoking several functions in wxWidgets while running under Wayland. The following code is impacted:

https://github.com/moneymanagerex/moneymanagerex/blob/dd67784392b6a3fefd5ca8abb0e4e4c8c3e4a54c/src/util.cpp#L1451

```
    for (unsigned int i = 0; i < wxDisplay::GetCount(); i++)
    {
        wxSharedPtr<wxDisplay> display(new wxDisplay(i));

        if (display->IsPrimary())
        {

            description << wxString::Format(bull + "%ix%i %ibit %ix%ippi\n",
                display->GetCurrentMode().GetWidth(),
                display->GetCurrentMode().GetHeight(),
                display->GetCurrentMode().bpp,
                display->GetPPI().GetWidth(),
                display->GetPPI().GetHeight()
            );
        }
    }
```

Which displays the following monitor information when running under Wayland:

```
* 0x0 0bit 96x96ppi
```

And `wxPlatformInfo::Get().GetDesktopEnvironment()` doesn't appear to return anything, possibly because it's using [X11-specific codepaths](https://docs.wxwidgets.org/latest/classwx_app_traits.html#ad333680437446304014c087d2fffedcf).

[A workaround has been made which falls back to checking several environment variables to find the current desktop environment.](https://github.com/moneymanagerex/moneymanagerex/commit/cc2714975ebc3a539f95bdac2420e5fe484232cd)

### Platform and version information

- wxWidgets version  you use: **3.2.3**
- wxWidgets port  you use: **wxGTK**

- OS  and its version: **Fedora 38 / Gnome 44**

+ GTK version:  **3.24.38**

+ Which GDK  backend is used: **Wayland**

vadz:
Video modes are indeed not supported but I have no idea if we have this kind of information under Wayland and I don't think it's that important, so I'd settle for just mentioning this in the docs.

`GetDesktopEnvironment()` is also rather strange, I have no idea what is it used for and what is it supposed to return under non-Gnome/KDE.

So all in all while enhancements are always welcome, for me neither of these is a priority.

paulcor:
It would be pretty easy to fake up the current `wxVideoMode`, does this address your problem?
```diff
diff --git a/src/gtk/display.cpp b/src/gtk/display.cpp
[...]

PR #24050, opened 2023-11-10
Improve desktop environment detection under Wayland

joshua-stone:
This change should help address https://github.com/wxWidgets/wxWidgets/issues/24021

### DIFF
--- a/src/gtk/utilsgtk.cpp
+++ b/src/gtk/utilsgtk.cpp
@@ -184,7 +184,11 @@ wxTimerImpl *wxGUIAppTraits::CreateTimerImpl(wxTimer *timer)
 #endif // wxUSE_TIMER
 
 #if wxUSE_DETECT_SM
-static wxString GetSM()
+#if defined(__GNUC__) && !(__GNUC_PREREQ(7, 0))
+[[gnu::unused]] static wxString GetSM()
+#else
+[[maybe_unused]] static wxString GetSM()
+#endif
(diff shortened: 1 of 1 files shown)


imciner2:
I think these environment variables are actually independent of Wayland, so they would also possibly be present under X11.

For instance, I currently have X11 (`wxGetDisplayInfo().type` returns 1 for me), and I have the environment variable `XDG_SESSION_DESKTOP=mate` set. So it would be nice if the check for these variables were actually display-agnostic.

paulcor:
I think checking for `XDG_CURRENT_DESKTOP` should be sufficient, anything old or oddball enough to not have it would likely be covered by the session manager code. I've done this now in 2d07dd25b4bf120a7c85baa44e09b17e611a2c08, please let me know if you find any problems with it.

joshua-stone:
Hey @paulcor, thanks for adding that enhancement!

I think the one major enhancement left for addressing https://github.com/wxWidgets/wxWidgets/issues/24021 is getting monitor properties under Wayland. I was able to use some [Python GTK examples](https://stackoverflow.com/questions/65807310/how-to-get-total-screen-size-in-python-gtk-without-using-deprecated-gdk-screen) to get monitor dimensions, but I haven't successfully ported these over to wxWidgets.

vadz:
AFAICS this should be backported to 3.2 too, so adding the label.

PR #24052, opened 2023-11-10
Add a possibility to get a dynamic library location for wxDynamicLibrary::Load()

oneeyeman1:
Please review and apply.

I added couple of comments to explain the reasoning.

And sorry for too many commits.

I will add the documentation after the good review, unless directed otherwise.

Thank you.

### DIFF
--- a/include/wx/osx/cocoa/stdpaths.h
+++ b/include/wx/osx/cocoa/stdpaths.h
@@ -35,7 +35,7 @@ public:
     virtual wxString MakeConfigFileName(const wxString& basename,
                                         ConfigFileConv conv = ConfigFileConv_Ext
                                         ) const override;
-
+    virtual wxString GetSharedLibrariesDir() const override;
 protected:
     // Ctor is protected, use wxStandardPaths::Get() instead of instantiating
     // objects of this class directly.
--- a/include/wx/stdpaths.h
+++ b/include/wx/stdpaths.h
@@ -257,7 +257,7 @@ public:
     {
         return m_prefix + wxS("/") + basename;
     }
-    virtual wxString GetSharedLibraresDir() const { return m_prefix; }
+    virtual wxString GetSharedLibrariesDir() overrde const { return m_prefix; }
 protected:
     // Ctor is protected because wxStandardPaths::Get() should always be used
     // to access the global wxStandardPaths object of the correct type instead
--- a/include/wx/stdpaths.h
+++ b/include/wx/stdpaths.h
@@ -166,7 +166,7 @@ public:
 
     // return the temporary directory for the current user
     virtual wxString GetTempDir() const;
-
+    virtual wxString GetSharedLibrariesDir() const;
     virtual wxString GetUserDir(Dir userDir) const;
 
     virtual wxString
@@ -257,7 +257,7 @@ public:
     {
         return m_prefix + wxS("/") + basename;
     }
-
+    virtual wxString GetSharedLibraroesPath() const { return m_prefix; }
 protected:
     // Ctor is protected because wxStandardPaths::Get() should always be used
     // to access the global wxStandardPaths object of the correct type instead
--- a/include/wx/stdpaths.h
+++ b/include/wx/stdpaths.h
@@ -257,7 +257,7 @@ public:
     {
         return m_prefix + wxS("/") + basename;
     }
-    virtual wxString GetSharedLibraroesPath() const { return m_prefix; }
+    virtual wxString GetSharedLibraresDir() const { return m_prefix; }
 protected:
     // Ctor is protected because wxStandardPaths::Get() should always be used
     // to access the global wxStandardPaths object of the correct type instead
--- a/include/wx/stdpaths.h
+++ b/include/wx/stdpaths.h
@@ -257,7 +257,10 @@ public:
     {
         return m_prefix + wxS("/") + basename;
     }
-    virtual wxString GetSharedLibrariesDir() overrde const { return m_prefix; }
+    virtual wxString GetSharedLibrariesDir() const override { return m_prefix; }
+#ifdef __WXMSW__
+    static wxString MSWGetShellDir(int csidl) { return wxEmptyString; }
+#endif
 protected:
     // Ctor is protected because wxStandardPaths::Get() should always be used
     // to access the global wxStandardPaths object of the correct type instead
--- a/include/wx/stdpaths.h
+++ b/include/wx/stdpaths.h
@@ -258,8 +258,8 @@ public:
         return m_prefix + wxS("/") + basename;
     }
     virtual wxString GetSharedLibrariesDir() const override { return m_prefix; }
-#ifdef __WXMSW__
-    static wxString MSWGetShellDir(int csidl) { return wxEmptyString; }
+#if defined(__WINDOWS__)
+    static wxString MSWGetShellDir(int WXUNUSED(csidl)) { return wxEmptyString; }
 #endif
 protected:
     // Ctor is protected because wxStandardPaths::Get() should always be used
--- a/interface/wx/stdpaths.h
+++ b/interface/wx/stdpaths.h
@@ -410,8 +410,8 @@ public:
     /**
         Return OS specific directory where project shared liraries are
 
-        - Windows: @c it returns the folder wher the application binary is located
-        - Unix:    @c it rturns the libraies installation path
+        - Windows: @c it returns the folder where the application binary is located
+        - Unix:    @c it returns the libraries installation path
         - Mac:     @c "/Applications/exename.app/Contents/Frameworks/"
 
         @since 3.3.0
--- a/interface/wx/stdpaths.h
+++ b/interface/wx/stdpaths.h
@@ -408,7 +408,11 @@ public:
     virtual wxString GetUserLocalDataDir() const;
 
     /**
-        Return OS specific directory where project shared liraries are
+        Return OS specific directory where project shared liraries are.
+        The function is different from GetPluginsDir() as PlugIns
+        and Shared Libraries are 2 different things (especially on Mac).
+        Plugin is a library that can be used without restartng the applicatioon
+        whereas the library needes to be installed and the application restarted.
 
         - Windows: @c it returns the folder where the application binary is located
         - Unix:    @c it returns the libraries installation path, i.e. /usr/lib
--- a/interface/wx/stdpaths.h
+++ b/interface/wx/stdpaths.h
@@ -409,14 +409,14 @@ public:
 
     /**
         Return OS specific directory where project shared liraries are.
-        The function is different from GetPluginsDir() as PlugIns
-        and Shared Libraries are 2 different things (especially on Mac).
-        Plugin is a library that can be used without restartng the applicatioon
-        whereas the library needes to be installed and the application restarted.
+        The function does the same thing as GetPluginsDir() under non-Mac platforms
+        but differs from it under Mac, where plugins (shared libraries loaded by the
+        application dynamically while it's running) and shared libraries (that the
+        application is statically linked with) are stored in different directories.
 
         - Windows: @c it returns the folder where the application binary is located
         - Unix:    @c it returns the libraries installation path, i.e. /usr/lib
-        - Mac:     @c "/Applications/exename.app/Contents/Frameworks/"
+        - Mac:     @c "/Applications/exename.app/Contents/Frameworks"
 
         @since 3.3.0
     */
--- a/interface/wx/stdpaths.h
+++ b/interface/wx/stdpaths.h
@@ -408,6 +408,17 @@ public:
     virtual wxString GetUserLocalDataDir() const;
 
     /**
+        Return OS specific directory where proect shared liraries are
+
+        - Windows: @c C:\Programs\AppFolder\
+        - Unix:    @c either /usr/lib/ or /usr/local/lib/
+        - OSX:     @c /Applications/exename.app/Contents/Frameworks/
+
+        @since 3.3.0
+    */
+    virtual wxString GetSharedLibrariesDir() const;
+
+    /**
         MSW-specific function to customize application directory detection.
 
         This class supposes that data, plugins &c files are located under the
--- a/interface/wx/stdpaths.h
+++ b/interface/wx/stdpaths.h
@@ -286,7 +286,7 @@ public:
     /**
         Return the program installation prefix, e.g.\ @c /usr, @c /opt or @c /home/zeitlin.
 
-        If the prefix had been previously by SetInstallPrefix(), returns that
+        If the prefix had been previously set by SetInstallPrefix(), returns that
         value, otherwise tries to determine it automatically (Linux only right now)
         and finally returns the default @c /usr/local value if it failed.
 
@@ -408,7 +408,7 @@ public:
     virtual wxString GetUserLocalDataDir() const;
 
     /**
-        Return OS specific directory where proect shared liraries are
+        Return OS specific directory where project shared liraries are
 
         - Windows: @c "C:\Programs\AppFolder\"
         - Unix:    @c either /usr/lib/ or /usr/local/lib/
--- a/interface/wx/stdpaths.h
+++ b/interface/wx/stdpaths.h
@@ -410,9 +410,9 @@ public:
     /**
         Return OS specific directory where proect shared liraries are
 
-        - Windows: @c C:\Programs\AppFolder\
+        - Windows: @c "C:\Programs\AppFolder\"
         - Unix:    @c either /usr/lib/ or /usr/local/lib/
-        - OSX:     @c /Applications/exename.app/Contents/Frameworks/
+        - OSX:     @c "/Applications/exename.app/Contents/Frameworks/"
 
         @since 3.3.0
     */
--- a/interface/wx/stdpaths.h
+++ b/interface/wx/stdpaths.h
@@ -410,9 +410,9 @@ public:
     /**
         Return OS specific directory where project shared liraries are
 
-        - Windows: @c "C:\Programs\AppFolder\"
-        - Unix:    @c either /usr/lib/ or /usr/local/lib/
-        - OSX:     @c "/Applications/exename.app/Contents/Frameworks/"
+        - Windows: @c it returns the folder wher the application binary is located
+        - Unix:    @c it rturns the libraies installation path
+        - Mac:     @c "/Applications/exename.app/Contents/Frameworks/"
 
         @since 3.3.0
     */
--- a/interface/wx/stdpaths.h
+++ b/interface/wx/stdpaths.h
@@ -411,7 +411,7 @@ public:
         Return OS specific directory where project shared liraries are
 
         - Windows: @c it returns the folder where the application binary is located
-        - Unix:    @c it returns the libraries installation path
+        - Unix:    @c it returns the libraries installation path, i.e. /usr/lib
         - Mac:     @c "/Applications/exename.app/Contents/Frameworks/"
 
         @since 3.3.0
--- a/src/common/stdpbase.cpp
+++ b/src/common/stdpbase.cpp
@@ -173,5 +173,5 @@ wxString wxStandardPathsBase::AppendAppInfo(const wxString& dir) const
 
 wxString wxStandardPathsBase::GetSharedLibrariesDir() const
 {
-    return "";
+    return {};
 }
--- a/src/common/stdpbase.cpp
+++ b/src/common/stdpbase.cpp
@@ -171,3 +171,7 @@ wxString wxStandardPathsBase::AppendAppInfo(const wxString& dir) const
     return subdir;
 }
 
+wxString wxStandardPathsBase::GetSharedLibrariesDir() const
+{
+    return "";
+}
--- a/src/msw/stdpaths.cpp
+++ b/src/msw/stdpaths.cpp
@@ -347,7 +347,7 @@ wxStandardPaths::MakeConfigFileName(const wxString& basename,
 wxString wxStandardPaths::GetSharedLibrariesDir() const
 {
     wxFileName fn( GetExecutablePath() );
-    return fn.GetPathWithSep();
+    return fn.GetPat();
 }
 
 // ============================================================================
--- a/src/msw/stdpaths.cpp
+++ b/src/msw/stdpaths.cpp
@@ -370,7 +370,7 @@ wxString wxStandardPathsWin16::GetUserConfigDir() const
 wxString wxStandardPaths::GetSharedLibrariesDir() const
 {
     wxFileName fn( GetExecutablePath() );
-    return fn.GetPath() + wxFileName::GetPathSeparator();
+    return fn.GetPathWithSep();
 }
 
 #endif // wxUSE_STDPATHS
--- a/src/msw/stdpaths.cpp
+++ b/src/msw/stdpaths.cpp
@@ -344,6 +344,12 @@ wxStandardPaths::MakeConfigFileName(const wxString& basename,
     return fn.GetFullName();
 }
 
+wxString wxStandardPaths::GetSharedLibrariesDir() const
+{
+    wxFileName fn( GetExecutablePath() );
+    return fn.GetPathWithSep();
+}
+
 // ============================================================================
 // wxStandardPathsWin16 implementation
 // ============================================================================
@@ -365,12 +371,6 @@ wxString wxStandardPathsWin16::GetUserConfigDir() const
 {
     // again, for wxFileConfig which uses $HOME for its user config file
     return wxGetHomeDir();
-}
-
-wxString wxStandardPaths::GetSharedLibrariesDir() const
-{
-    wxFileName fn( GetExecutablePath() );
-    return fn.GetPathWithSep();
 }
 
 #endif // wxUSE_STDPATHS
--- a/src/msw/stdpaths.cpp
+++ b/src/msw/stdpaths.cpp
@@ -367,4 +367,10 @@ wxString wxStandardPathsWin16::GetUserConfigDir() const
     return wxGetHomeDir();
 }
 
+wxString wxStandardPaths::GetSharedLibrariesDir() const
+{
+    wxFileName fn( GetExecutablePath() );
+    return fn.GetPath() + wxFileName::GetPathSeparator();
+}
+
 #endif // wxUSE_STDPATHS
--- a/src/msw/stdpaths.cpp
+++ b/src/msw/stdpaths.cpp
@@ -347,7 +347,7 @@ wxStandardPaths::MakeConfigFileName(const wxString& basename,
 wxString wxStandardPaths::GetSharedLibrariesDir() const
 {
     wxFileName fn( GetExecutablePath() );
-    return fn.GetPat();
+    return fn.GetPath();
 }
 
 // ============================================================================
--- a/src/msw/stdpaths.cpp
+++ b/src/msw/stdpaths.cpp
@@ -369,8 +369,9 @@ wxString wxStandardPathsWin16::GetUserConfigDir() const
 
 wxString wxStandardPaths::GetSharedLibrariesDir() const
 {
-    wxFileName fn( GetExecutablePath() );
-    return fn.GetPathWithSep();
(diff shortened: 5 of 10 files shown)


oneeyeman1:
I don't understand those failures.
They are completely unrelated to the commits I pushed and I didn't get them locally.

src/unix/stdpaths.cpp
```diff
@@ -365,4 +365,16 @@ wxStandardPaths::MakeConfigFileName(const wxString& basename,
     return fn.GetFullName();
 }
 
+wxString wxStandardPaths::GetSharedLibrariesDir() const
+{
+    char c;
+    wxFileName::GetPathSeparator().GetAsChar( &c );
+    // While debugging libraries usually goes to "/usr/local/lib"
+    // while in production the usual place is "/usr/lib"
+#ifdef DEBUG
```
vadz, line 374: `DEBUG` is not used nor defined anywhere.

vadz:
> I don't understand those failures. They are completely unrelated to the commits I pushed

Of course they're related to the submodules that you (wrongly) updated.

I don't understand, do you know you can _look_ at your changes, even after submitting them (it's better to do it before, but...)? Can't you see that the submodules have been changed? Don't you wonder if this is normal/intentional? I'm just lost here...

oneeyeman1:
@vadz ,
> > I don't understand those failures. They are completely unrelated to the commits I pushed
> 
> Of course they're related to the submodules that you (wrongly) updated.

I didn't update them.

After doing "git fetch origin && git merge origin/master" I had the scintilla submodule unmerged. I thought it will take care of everything, but it didn't.
So I deleted src/stc/scintilla and ran the "git submodule update --init".

Please check the thread on the wx-dev and reply from Maarten..

I don't now why it happened, but after doing it I guess everything was updated to its latest version.

Any idea how to fix it easily?

If not - I will probably have to redo the fork and resubmit.
> 
> I don't understand, do you know you can _look_ at your changes, even after submitting them (it's better to do it before, but...)? Can't you see that the submodules have been changed? Don't you wonder if this is normal/intentional? I'm just lost here...

vadz:
> > Of course they're related to the submodules that you (wrongly) updated.
> 
> I didn't update them.

You *did* update them. The fact that you didn't realize that you updated them is somewhat understandable but the fact that you didn't even look at your changes before pressing "Submit" in spite of being asked to do this literally dozens of time in the past is not. And the fact that you still can somehow claim that you didn't even after I told that you did is just mind boggling.

Again, anybody makes mistakes. But nobody else refuses not only to check for them before submitting but even refuses to see them after submitting. It just drives me crazy that you can look at the compilation errors in the submodules after pushing an update to them and your only conclusion is that something unrelated must be wrong -- and not that you have actually broken it.

You really need to carefully review your next submission, whatever it is, at least twice before submitting it because this colossal waste of time just can't continue any longer.

MaartenBent:
Like I said in wx-dev, your branch has diverged from the wxWidgets branch. I'm 99.9% sure you committed wrong submodules to it.
So even `git submodule update` will update the submodules to that version, not the one in wxWidgets master.

Like I said in wx-dev, you should have done `git reset --hard origin/master` to get rid of it. You can still do this, but you'll loose the commits with the changes of this PR. But they're visible here on GitHub, so just manually update the files again, commit, and then force push.

And like @vadz said, just click the `changed files` tab before creating the PR. You would have seen all of those submodules change. And the `lobrary` typo in the title.

oneeyeman1:
@MaartenBent ,
> Like I said in wx-dev, your branch has diverged from the wxWidgets branch. I'm 99.9% sure you committed wrong submodules to it. So even `git submodule update` will update the submodules to that version, not the one in wxWidgets master.

I don't know how it was possible.
Like I said - all I did is to kill the scintilla folder and ran the "git submidule updaye --init", which should have picked the correct commit from the appropriate submodule.

But that's another mistery that I don't have time to solve right now.

> 
> Like I said in wx-dev, you should have done `git reset --hard origin/master` to get rid of it. You can still do this, but you'll loose the commits with the changes of this PR. But they're visible here on GitHub, so just manually update the files again, commit, and then force push.

I can do that command on my PR branch, right?


> 
> And like @vadz said, just click the `changed files` tab before creating the PR. You would have seen all of those submodules change. And the `lobrary` typo in the title.

MaartenBent:
> I don't know how it was possible. Like I said - all I did is to kill the scintilla folder and ran the "git submidule updaye --init", which should have picked the correct commit from the appropriate submodule.

You could have committed those old submodules weeks or months ago. Not to this branch, but in your master branch. It is not related to the scintilla stuff you did.

> I can do that command on my PR branch, right?

Yes.
`git reset --hard origin/master`
`git update submodules`
Then add all the changes from https://github.com/wxWidgets/wxWidgets/pull/24052/files and commit them
`git push -f`

And because your master branch is diverged, also do the first two commands in that branch. Not now, but definitely before you check out a new branch. Otherwise we'll have the same problem all over again.

dgud:
>>  git reset --hard origin/master
That depends on what his **origin** is set to, shouldn't that be **upstream** or some other name, don't know what you recommend. 

But **origin**  is normally his fork of wxWidgets not the wxWidgets repo.

MaartenBent:
True, I messed up there. It should be `upstream/master`. Or whatever name was used for the upstream remote.

src/unix/stdpaths.cpp
oneeyeman1: @vadz,
It looks like there is o definition I can use on Linux/Mac with configure build.
Neither DEBUG, not DEBUG_INFO or wxUSE_DEBUG{_INFO} are defined

Any suggestions?

src/unix/stdpaths.cpp
```diff
@@ -365,4 +365,15 @@ wxStandardPaths::MakeConfigFileName(const wxString& basename,
     return fn.GetFullName();
 }
 
+wxString wxStandardPaths::GetSharedLibrariesDir() const
+{
+    // While debugging libraries usually goes to "/usr/local/lib"
+    // while in production the usual place is "/usr/lib"
+#ifdef wxUSE_DEBUG
```
MaartenBent, line 372: Using compile definitions and hardcoded paths seem wrong and ununsable to me.
Have you seen `GetInstallPrefix()`?

src/osx/cocoa/stdpaths.mm
```diff
@@ -139,4 +139,13 @@ ConfigFileConv WXUNUSED(conv)) const
     return fn.GetFullName();
 }
 
+wxString wxStandardPaths::GetSharedLibrariesDir() const
+{
+    // Shared libraries on OSX should be stored inside the
+    // <Bundle.app>/Contents/Frameworks
+    wxFileName fn( GetExecutablePath() );
+    wxString temp = fn.GetPath() + wxFileName::GetPathSeparator() + ".." + wxFileName::GetPathSeparator() + "Frameworks" + wxFileName::GetPathSeparator();
```
MaartenBent, line 147: This will have `..` in the result. There should be some way to resolve this path I assume.

src/msw/stdpaths.cpp
```diff
@@ -367,4 +367,10 @@ wxString wxStandardPathsWin16::GetUserConfigDir() const
     return wxGetHomeDir();
 }
 
+wxString wxStandardPaths::GetSharedLibrariesDir() const
+{
+    wxFileName fn( GetExecutablePath() );
+    return fn.GetPath() + wxFileName::GetPathSeparator();
```
MaartenBent, line 373: Have you seen `wxFileName::GetPathWithSep()`?
Also relevant in other places.

interface/wx/stdpaths.h
```diff
@@ -407,6 +407,17 @@ class wxStandardPaths
     */
     virtual wxString GetUserLocalDataDir() const;
 
+    /**
+        Return OS specific directory where proect shared liraries are
```
MaartenBent, line 411: `proect`?

interface/wx/stdpaths.h
```diff
@@ -407,6 +407,17 @@ class wxStandardPaths
...
+
+        - Windows: @c "C:\Programs\AppFolder\"
+        - Unix:    @c either /usr/lib/ or /usr/local/lib/
+        - OSX:     @c "/Applications/exename.app/Contents/Frameworks/"
```
MaartenBent, line 415: Maybe clarify that these are examples. I can install somewhere else on Windows. Or run the macOS .app from a different directory.

And maybe use `appinfo` instead of `exename`, like some other documentation does.

include/wx/stdpaths.h
```diff
@@ -257,7 +257,7 @@ class WXDLLIMPEXP_BASE wxStandardPaths : public wxStandardPathsBase
     {
         return m_prefix + wxS("/") + basename;
     }
-
+    virtual wxString GetSharedLibraroesPath() const { return m_prefix; }
```
MaartenBent, line 260: what is this?

include/wx/osx/cocoa/stdpaths.h
```diff
@@ -35,7 +35,7 @@ class WXDLLIMPEXP_BASE wxStandardPaths : public wxStandardPathsBase
     virtual wxString MakeConfigFileName(const wxString& basename,
                                         ConfigFileConv conv = ConfigFileConv_Ext
                                         ) const override;
-
+    virtual wxString GetSharedLibrariesDir() const override;
```
MaartenBent, line 38: For readability, don't replace the empty line, just add a new line.
Also in other places.

src/unix/stdpaths.cpp
oneeyeman1: @MaartenBent ,
Wow! Thank you for pointing that out. I didn't know about this one. I will look at that and see what it does.

src/osx/cocoa/stdpaths.mm
oneeyeman1: @MaartenBent ,
Fixed in the upcoming commit. Thx.

src/msw/stdpaths.cpp
oneeyeman1: @MaartenBent ,
Didnt know about that. Fixed. Thx.

src/osx/cocoa/stdpaths.mm
```diff
@@ -139,4 +139,17 @@ ConfigFileConv WXUNUSED(conv)) const
...
 }
 
+wxString wxStandardPaths::GetSharedLibrariesDir() const
+{
+    // Shared libraries on OSX should be stored inside the
+    // <Bundle.app>/Contents/Frameworks
+    wxFileName fn( GetExecutablePath() );
+    auto temp = fn.GetPath();
+    auto pos = temp.rfind( "/MacOS" );
+    if( pos != wxString::npos )
```
MaartenBent, line 149: I was thinking of using `wxFileName::RemoveLastDir()`. You seem to come up with the hardest solutions for a simple problem.

MaartenBent: On second thought, you probably should use a similar solution as `GetExecutablePath()`. 
https://developer.apple.com/documentation/foundation/nsbundle/1408900-builtinpluginspath

interface/wx/stdpaths.h
oneeyeman1: @MaartenBent ,
Fixed. Also fixed another typo.

include/wx/stdpaths.h
oneeyeman1: @MaartenBent ,
It is generic implementation.

MaartenBent: No, this is an entire different function.

interface/wx/stdpaths.h
lanurmi: What version of Windows uses literally `C:\Programs`?

oneeyeman1: @lanurmi ,
I don't know. But that's what the documentation says.

src/osx/cocoa/stdpaths.mm
oneeyeman1: @MaartenBent ,
> On second thought, you probably should use a similar solution as `GetExecutablePath()`. https://developer.apple.com/documentation/foundation/nsbundle/1409078-executablepath

That what OSX is using inside GetExecutablePath().

oneeyeman1: @MaartenBent 
> I was thinking of using `wxFileName::RemoveLastDir()`. You seem to come up with the hardest solutions for a simple problem.

I used the solution from the GetInstallPrefix().

oneeyeman1: @MaartenBent 
After thinking more about it I changed it to use your latest suggestion.

Thx.

oneeyeman1:
@MaartenBent ,
Do you think its now ready?

interface/wx/stdpaths.h
lanurmi: > @lanurmi , I don't know. But that's what the documentation says.

I can't find any search results for `C:\Programs` either in 'documentation' or anywhere else on the web.

MaartenBent:
@oneeyeman1 no, there are still open issues, see all the unresolved comments. Here my comments after a code review, I didn't test it so I assume it at least works.
* There is a `GetSharedLibraroesPath`, which has nothing to do with `GetSharedLibrariesDir`.
* On Windows you should move the implementation, it is in the wxStandardPathsWin16 section now. And maybe you could use `GetAppDir`, just like `GetPluginsDir` does, see also the comment in that function.
* On macOS you should use the [designated API](https://developer.apple.com/documentation/foundation/nsbundle/1417226-sharedframeworkspath), not `GetExecutablePath`.
* The documentation needs updating, see the comments. The Windows location is very unrealistic as @lanurmi commented. Shouldn't it be Linux instead of Unix? And the OSX (is it still called that? probably should use macOS) name can be improved. Maybe make them relative to a install path instead of using fixed paths. 
* Less important, but if you look at the diff you can see that you replaced empty lines, you should add the declarations on new lines. It would be nice if you could fix this so @vadz doesn't have to do all the cleanup and can just squash and merge.

And a general question. Is the location in Linux always `/lib`, or could it also be `/lib64`. I suppose it depends on where the packagers decide to install it.

oneeyeman1:
@MaartenBent ,
> There is a GetSharedLibraroesPath, which has nothing to do with GetSharedLibrariesDir.

This will be fixed in an upcoming commit.

> On macOS you should use the [designated API](https://developer.apple.com/documentation/foundation/nsbundle/1417226-sharedframeworkspath), not GetExecutablePath.

According to [Apple docs](https://developer.apple.com/documentation/xcode/embedding-nonstandard-code-structures-in-a-bundle) the folder is called Frameworks. And that'ss where Xcode places the dynamic libraries when it creates Bundle.
Using that API I get the folder called SharedFramework. 

> The documentation needs updating, see the comments. The Windows location is very unrealistic as @lanurmi commented. Shouldn't it be Linux instead of Unix? And the OSX (is it still called that? probably should use macOS) name can be improved. Maybe make them relative to a install path instead of using fixed paths.

Looking at the documentation of GetExecutablePath(), it uses Unix, not Linux, but it does use Mac and not OSX. Will fix. Also I will change the Windows verbiage to explain what will happen.

> And a general question. Is the location in Linux always /lib, or could it also be /lib64. I suppose it depends on where the packagers decide to install it.

I think one is a symlink ro another.
It is the case on my home machine. And it is mute, since I changed the verbiage for it to make it better.

> On Windows you should move the implementation, it is in the wxStandardPathsWin16 section now. And maybe you could use GetAppDir, just like GetPluginsDir does, see also the comment in that function.

I will check the Windows and make sure its fixed as well.

I did look and I didn't find any additional conditional directives inside src/msw/stdpaths.cpp

imciner2:
> I think one is a symlink ro another.
> It is the case on my home machine. And it is mute, since I changed the verbiage for it to make it better.

It is not guaranteed to be a symlink - on Fedora /usr/lib and /usr/lib64 are two different directories. And the verbage would be wrong on Fedora, because the libraries are installed into `/usr/lib64` and you are not returning that directory.

oneeyeman1:
@imciner2 ,

> > I think one is a symlink ro another.
> > It is the case on my home machine. And it is mute, since I changed the verbiage for it to make it better.
> 
> It is not guaranteed to be a symlink - on Fedora /usr/lib and /usr/lib64 are two different directories. And the verbage would be wrong on Fedora, because the libraries are installed into `/usr/lib64` and you are not returning that directory.

Did you try it?
What it returns for you?
And what GetInstallPrefix() returns for you?

Thx.

include/wx/stdpaths.h
oneeyeman1: @MaartenBent 
> No, this is an entire different function.

Sorry, after looking at the code more close it appears to be a function that will be used if generic verion of the class is used.

oneeyeman1:
@MaartenBent ,
Also I just re-checked and wxStandardPathsWin16 does not contain an implementation of my function:

[code]
Igors-MacBook-Air:wxFork igorkorot$ grep -riI wxStandardPathsWin16 *
include/wx/msw/stdpaths.h:// wxStandardPathsWin16: this class is for internal use only
include/wx/msw/stdpaths.h:class WXDLLIMPEXP_BASE wxStandardPathsWin16 : public wxStandardPaths
patch:@@ -367,4 +367,9 @@ wxString wxStandardPathsWin16::GetUserConfigDir() const
src/msw/stdpaths.cpp:// wxStandardPathsWin16 implementation
src/msw/stdpaths.cpp:wxString wxStandardPathsWin16::GetConfigDir() const
src/msw/stdpaths.cpp:wxString wxStandardPathsWin16::GetUserConfigDir() const
Igors-MacBook-Air:wxFork igorkorot$ 
[/code]

Thank you.

oneeyeman1:
@MaartenBent ,
I have  been using this code for about a month and didn't see anything bad.

@vadz ,
Should I update the documentation  @since tag?
Could you please look at it?

Thank you.

MaartenBent:
> Also I just re-checked and wxStandardPathsWin16 does not contain an implementation of my function

I don't mean the class, just the location where you put the code:
Screenshot 2023-11-25 134827


Do you see the difference between these two lines?
`GetSharedLibrariesDir`
`GetSharedLibraresDir`
Check your code because it is in there. The line with the typo should have an `override` keyword so the compiler will throw an error that you are trying to override a function that does not exist.

oneeyeman1:
@MaartenBent ,
Could you please check what else is missing?

Thx

MaartenBent:
The typo in `GetSharedLibraresDir` is still there

oneeyeman1:
@MaartenBent 
> The typo in `GetSharedLibraresDir` is still there

Fixed in the latest commit

Can you apply it?

Thx

oneeyeman1:
@MaartenBent ,
I fixed the typo and also fixed the error message on Windows when USE_STDPATH is 0.

However, in this case I'm getting following warnings:
```

c:\wxfork\include\wx\stdpaths.h(262): warning C4100: 'csidl': unreferenced formal parameter (compiling source file ..\..\src\common\docview.cpp)
c:\wxfork\include\wx\stdpaths.h(262): warning C4100: 'csidl': unreferenced formal parameter (compiling source file ..\..\src\msw\rt\notifmsgrt.cpp)

```

MaartenBent:
I see `MSWGetShellDir` was added back in [2009](https://github.com/wxWidgets/wxWidgets/commit/94aff795d414f4d55ecd7eda44362c4c57023ce4), and only now we notice it is missing in this `wxStandardPaths` implementation. So fixing the warning isn't the highest priority.

It should be simple to fix though using `WXUNUSED`, like is also done for a few other methods in this class.
```c++
static wxString MSWGetShellDir(int WXUNUSED(csidl)) { return wxEmptyString; }
```

oneeyeman1:
@MaartenBent ,
If I do this I get link error about not finding proper functin.
But as you say - it's not high priority.

Can you push that code?

Thank you.

MaartenBent:
I get the error too when building shared libraries. The problem is that `__WXMSW__`  is not defined for the `base` library, only the `core` library.
So I think the best fix is to use `#if defined(__WINDOWS__)`, just like line 221 in the same file.

oneeyeman1:
@MaartenBent ,
You are right.. Just verified and compilation finished fine.
Could you also look at https://github.com/wxWidgets/wxWidgets/pull/23558?

I put some comments in and added some additional fixes to the text code....
Thx.

oneeyeman1:
@MaartenBent ,
Is this now ready to be merged?

Thx

include/wx/stdpaths.h
```diff
@@ -257,7 +257,10 @@ class WXDLLIMPEXP_BASE wxStandardPaths : public wxStandardPathsBase
     {
         return m_prefix + wxS("/") + basename;
     }
-
+    virtual wxString GetSharedLibrariesDir() const override { return m_prefix; }
+#if defined(__WINDOWS__)
+    static wxString MSWGetShellDir(int WXUNUSED(csidl)) { return wxEmptyString; }
+#endif
```
vadz, line 263: What's this and why do we need it? It doesn't seem to be used anywhere.

src/common/stdpbase.cpp
```diff
@@ -171,3 +171,7 @@ wxString wxStandardPathsBase::AppendAppInfo(const wxString& dir) const
     return subdir;
 }
 
+wxString wxStandardPathsBase::GetSharedLibrariesDir() const
+{
+    return "";
```
vadz, line 176:
```suggestion
    return {};
```

interface/wx/stdpaths.h
```diff
@@ -407,6 +407,17 @@ class wxStandardPaths
     */
     virtual wxString GetUserLocalDataDir() const;
 
+    /**
+        Return OS specific directory where project shared liraries are
+
+        - Windows: @c it returns the folder where the application binary is located
+        - Unix:    @c it returns the libraries installation path
```
vadz, line 414: This is not really clear, I'd give `lib` directory as an example.

src/osx/cocoa/stdpaths.mm
```diff
@@ -139,4 +139,13 @@ ConfigFileConv WXUNUSED(conv)) const
     return fn.GetFullName();
 }
 
+wxString wxStandardPaths::GetSharedLibrariesDir() const
+{
+    // Shared libraries on OSX should be stored inside the
+    // <Bundle.app>/Contents/Frameworks
+    wxFileName fn( GetExecutablePath() );
+    fn.RemoveLastDir();
+    return fn.GetPathWithSep() + "Frameworks/";
```
vadz, line 148: I think we need one of the functions from [this page](https://developer.apple.com/documentation/foundation/nsbundle/1408900-builtinpluginspath/) instead although I'm not sure which one. Please try using them to see which of them returns `Frameworks`.

vadz (changes requested):
The documentation needs to explain how is the new function different from the existing `GetPluginsDir()`.

It also should either mention that the path is `/`-terminated or, better, remove it because this seems to be inconsistent with all the other functions and with the empty return value in the base class.

include/wx/stdpaths.h
oneeyeman1: @vadz,
When you turn off wxUSE_STDPATH and compile on Windows this function has to be defined.
Otherwise you will get compiler error

src/common/stdpbase.cpp
oneeyeman1: Fixed

src/osx/cocoa/stdpaths.mm
oneeyeman1: @vadz ,
Found it.
Tested and committed.

Thx

oneeyeman1:
@vadz ,
> The documentation needs to explain how is the new function different from the existing `GetPluginsDir()`.

I will try to put some wording explaining this.

> 
> It also should either mention that the path is `/`-terminated or, better, remove it because this seems to be inconsistent with all the other functions and with the empty return value in the base class.

Will remove.

Thx

oneeyeman1:
@vadz ,
I tried to clarify the difference between those 2 functions.

Feel free to critique my explanation.

Thank you

include/wx/stdpaths.h
vadz: Ah, so this has nothing to do with this PR.

This is not the proper fix for the problem, of course. AFAICS this is only used in `src/msw/rt/notifmsgrt.cpp` and so the right thing to do would be to add `#if wxUSE_STDPATH` around the code there. This really has nothing to do here in any case.

src/osx/cocoa/stdpaths.mm
```diff
@@ -143,9 +143,7 @@ ConfigFileConv WXUNUSED(conv)) const
 {
     // Shared libraries on OSX should be stored inside the
     // <Bundle.app>/Contents/Frameworks
-    wxFileName fn( GetExecutablePath() );
-    fn.RemoveLastDir();
-    return fn.GetPathWithSep() + "Frameworks/";
+    return wxCFStringRef::AsString((CFStringRef)[NSBundle mainBundle].privateFrameworksPath);
```
vadz, line 146: Why do you need the cast here? Using `NSString*` directly should work too, i.e.
```suggestion
    return wxCFStringRef::AsString([NSBundle mainBundle].privateFrameworksPath);
```

interface/wx/stdpaths.h
```diff
@@ -408,10 +408,14 @@ class wxStandardPaths
     virtual wxString GetUserLocalDataDir() const;
 
     /**
-        Return OS specific directory where project shared liraries are
+        Return OS specific directory where project shared liraries are.
+        The function is different from GetPluginsDir() as PlugIns
+        and Shared Libraries are 2 different things (especially on Mac).
+        Plugin is a library that can be used without restartng the applicatioon
+        whereas the library needes to be installed and the application restarted.
```
vadz, line 415: Thanks, this does explain it, but I'd just say
```suggestion

        The function does the same thing as GetPluginsDir() under non-Mac platforms
        but differs from it under Mac, where plugins (shared libraries loaded by the
        application dynamically while it's running) and shared libraries (that the
        application is statically linked with) are stored in different directories.
```

interface/wx/stdpaths.h
```diff
@@ -408,10 +408,14 @@ class wxStandardPaths
...
 
         - Windows: @c it returns the folder where the application binary is located
-        - Unix:    @c it returns the libraries installation path
+        - Unix:    @c it returns the libraries installation path, i.e. /usr/lib
         - Mac:     @c "/Applications/exename.app/Contents/Frameworks/"
```
vadz, line 419: This is wrong now and should be
```suggestion
        - Mac:     @c "/Applications/exename.app/Contents/Frameworks"
```

vadz (commented):
Thanks, this is close to being mergeable now, just a couple of minor things to fix.

include/wx/stdpaths.h
oneeyeman1: @vadz ,
No it doesn't..
However, when I set wxUSE_STDPATH to 0 and tried to compile with MSVC I got an error.

All I was trying to do is make sure the code compiles. And when I saw it didn't - I tried to fix it.

I can remove this piece of code here and add the ifdef in there in another PR fixing it. However, I will not be able to test it

src/osx/cocoa/stdpaths.mm
oneeyeman1: @vadz ,
There is another function that uses this cast, so I just copied it here.

Will check and if its not needed - will remove.

Do you want to fix the other function as well?

interface/wx/stdpaths.h
oneeyeman1: @vadz ,
Fixed in the upcoming commit.

interface/wx/stdpaths.h
oneeyeman1: @vadz ,
Fixed in the upcoming commit.

src/osx/cocoa/stdpaths.mm
oneeyeman1: @vadz ,
I removed the cast, but only for that function.

oneeyeman1:
@vadz ,
I see at least 2 potential PR after this is accepted - remove the cast n all other MAC functions and fix the MSW code.
Could you push this PR as is right now and I'll try to make follow-up same day?

TIA!!!

vadz:
Finally merging with minor changes.

oneeyeman1:
Thx.

PR #24053, opened 2023-11-12
OSX migrate to NSCell based renderer (2nd attempt)

csomor:
the old HITheme renderer does not support dark mode for the disclosure button, we should move to use NSCells for rendering in the future, see #24002, replaces PS #24004 with better history of file renderer.mm


csomor:
Hi @vadz I've tried to setup things correctly for keeping renderer's history, if you think it's ok, then we could merge it, the graphical glitches I've encountered in propgrid (checkbox' focus rects remaining) are already present in the old implementation.

src/osx/cocoa/renderer.mm
```diff
@@ -162,13 +192,62 @@ wxRendererNative& wxRendererNative::GetDefault()
     return s_rendererMac;
 }
 
+wxRendererMac::wxRendererMac()
+{
+#if wxOSX_USE_NSCELL_RENDERER
+    m_nsPushButtonCell = [[NSButtonCell alloc] initTextCell:@""];
```
vadz, line 198: Do you think it could be worth to allocate all these cells on demand only, i.e. only when a renderer method using it is called for the first time? Or are cells so cheap to create that it's not worth optimizing this?

I'm just a bit wary of adding more and more things unconditionally done on every application startup, each of them doesn't cost much but sooner or later they're going to add up to something noticeable.

vadz (commented):
Thanks a lot, I'm all for merging this, just please let me know if you think it could be worth to create the cells on demand.

src/osx/cocoa/renderer.mm
csomor: The CPU on my M1 is 6 msec for the allocating all cells, so at the moment I'd rather leave it like that, if that time is ok

vadz: To be honest, 6ms doesn't seem negligible to me: I've just tested and with the debug build of wxGTK (I was too lazy to rebuild in release, but it should be smaller there) the time-to-first-paint is 60ms, i.e. it's 10% of the total startup time.

If you'd like, I can make the changes to create them on demand, but I'd just like to ask you to test them because I'm not sure how to do it.

csomor: > If you'd like, I can make the changes to create them on demand, but I'd just like to ask you to test them because I'm not sure how to do it.

@vadz I'd very much appreciate that, I'll test it, thanks :-)

vadz: I've pushed my changes now (includes a warning fix I was getting here), please let me know if I did anything stupid. TIA!

csomor:
@vadz Thanks a lot, I've just added a small change, I'm all for merging this ....

Issue #23916, opened 2023-09-29
wxAUI panels don't clip properly with Sonoma SDK/Xcode 15


dkulp:
When wxWidgets is built with the latest Xcode 15 and the Sonoma SDK it ships with, panels within wxAUI miniframes and other things are not clipped properly.    

It can be seen in the wxAUI sample by making the window smaller (or by adding more controls to the panel):
![image](https://github.com/wxWidgets/wxWidgets/assets/91569/aa116ce0-9588-49ad-840c-5ee07c68e7af)
The wxSpinCtrl is not clipped and is over the splitter.  

If you can get ahold of the 13.3 SDK and build with the --with-macosx-sdk= flag to point at the 13.3, then everything displays properly. 

### Platform and version information
- wxWidgets version you use: latest master code
- wxWidgets port  you use: wxOSX
- OS  and its version: Sonoma

vadz:
I suspect this is not wxAUI-specific (at least I don't see any Mac-only code in it that could be responsible for this), but I have no idea what is this due to. Any ideas (or patches) are more than welcome...

tonyvsuk:
I think this is exactly the same thing I'm seeing and reporting in the wx users mailing list.
https://groups.google.com/g/wx-users/c/rf9UfE5ah_o

Here is a patch (I think, I hope) that shows the clipping problem in the aui sample.
[mac_clipping_problem.patch](https://github.com/wxWidgets/wxWidgets/files/13326070/mac_clipping_problem.patch)

@dkulp, how did you get hold of the older SDK? I'm desperate to get this fixed as I've got users struggling with this.

Tony.

dkulp:
@tonyvsuk I still had a Ventura machine with Xcode 14 installed so I grabbed the 13.3 SDK out of there.

tonyvsuk:
I've got an "old" 2017 Mac Air (Monterry/Xcode 14), so hopefully that will work. Downloading Xcode 13.3 as I write this.

@dkulp Did you recompile your entire app with Xcode 13, or just wx? If it's just wx, that would be really handy for me. I usually build wx as a static library (but do intend to go to shared once this problem is out of the way).

dkulp:
For my app, we use LTO and thus everything has to be compiled consistently or all kinds of warnings.   Thus, we do build everything with 13.3.  That said, once the SDK is added to Xcode (a symlink to a location elsewhere works) 
[...]

PR #24054, opened 2023-11-12
OSX Sonoma keep clipping behavior for paint events

csomor:
see #23916


vadz:
AFAICS this should be backported to 3.2 too, so adding the label.

PR #24055, opened 2023-11-12
Try to make bug report instructions even more clear

vadz:
What do people think of changing it like this?

Notably, very few people think of submitting stack traces on their own (thanks a lot to those who do!).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -11,29 +11,49 @@ assignees: ''
 IMPORTANT NOTE
 --------------
 
-Please note that this entry field uses [Markdown][1], which means that you
-need to escape any code snippets pasted here using fenced code blocks,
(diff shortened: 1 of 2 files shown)


MaartenBent:
Maybe mention to put it in code tags, so there is no massive wall of text.

While you're updating the templates, could you also update `build_problem.md` with a question what they use to build wxWidgets?
 Configure, CMake, Visual Studio solution, Xcode project, makefile.gcc/makefile.vc.

vadz:
> Maybe mention to put it in code tags, so there is no massive wall of text.

There is already a mention of this, but I've tried to emphasize it even more in the latest version. Any suggestions for making it even more clear would be welcome, but we also need to keep this brief because nobody reads long texts...
 
> While you're updating the templates, could you also update `build_problem.md` with a question what they use to build wxWidgets? Configure, CMake, Visual Studio solution, Xcode project, makefile.gcc/makefile.vc.

Done too.

Issue #22965, opened 2022-11-12
Use `= default` whenever possible for ctors and dtors


vadz:
We have a number (grep finds 700 of them but there are probably some false positives there) of ctors and dtors with the empty body. We should use `= default` for them instead as this is more clear and would also be more consistent with the new code where it will be used.

There are even more classes doing just trivial initialization of their members in the ctors, where we should initialize the members in their declarations instead and also make the ctor default, but this could be done separately.

PR #24058, opened 2023-11-13
Replacing empty ctors and dtors by = default

BrianDelalex:
Following issue [#22965](https://github.com/wxWidgets/wxWidgets/issues/22965) this change replace empty dtors and ctors by = default whenever possible.


vadz:
Thanks, this is not really reviewable (too many changes...), but a random sampling of them seems good.

The CI errors are all of the form
```
./strings/stdstrings.cpp:56:24: error: unused variable 'i1' [-Werror,-Wunused-variable]
```
and seem due to the fact that clang considers variables with defaulted ctor unused, unlike those with user-defined trivial ctor. The fix would be to add `wxUnusedVar()` calls to this test case.

vadz:

Will merge in a moment.

PR #24061, opened 2023-11-14
Implement maximum text length for multiline text control under GTK

oneeyeman1:
This PR is a redo of the 23928.

### DIFF
--- a/include/wx/gtk/textctrl.h
+++ b/include/wx/gtk/textctrl.h
@@ -59,7 +59,7 @@ public:
     virtual void GetSelection(long* from, long* to) const override;
 
     virtual void Remove(long from, long to) override;
-
+    virtual void SetMaxLength(unsigned long length) override;
     virtual void MarkDirty() override;
     virtual void DiscardEdits() override;
 
--- a/include/wx/gtk/textctrl.h
+++ b/include/wx/gtk/textctrl.h
@@ -133,6 +133,10 @@ public:
     bool SetForegroundColour(const wxColour& colour) override;
     bool SetBackgroundColour(const wxColour& colour) override;
 
+    void HandleMaxLength(wxString &text);
+    bool IsMaxLengthAllowed() const { return m_maxLengthAllowed; }
+    void SetMaxLengthAllowed(const bool &allowed) { m_maxLengthAllowed = allowed; }
+
     GtkWidget* GetConnectWidget() override;
 
     void SetUpdateFont(bool WXUNUSED(update)) { }
@@ -227,7 +231,6 @@ private:
 
     bool m_modified;
     bool m_dontMarkDirty;
-
     int         m_countUpdatesToIgnore;
 
     // Our text buffer. Convenient, and holds the buffer while using
--- a/include/wx/gtk/textctrl.h
+++ b/include/wx/gtk/textctrl.h
@@ -134,8 +134,6 @@ public:
     bool SetBackgroundColour(const wxColour& colour) override;
 
     void HandleMaxLength(wxString &text);
-    bool IsMaxLengthAllowed() const { return m_maxLengthAllowed; }
-    void SetMaxLengthAllowed(const bool &allowed) { m_maxLengthAllowed = allowed; }
 
     GtkWidget* GetConnectWidget() override;
 
--- a/include/wx/gtk/textctrl.h
+++ b/include/wx/gtk/textctrl.h
@@ -233,7 +233,7 @@ private:
 
     bool m_modified;
     bool m_dontMarkDirty;
-
+    
     int         m_countUpdatesToIgnore;
 
     // Our text buffer. Convenient, and holds the buffer while using
--- a/include/wx/gtk/textctrl.h
+++ b/include/wx/gtk/textctrl.h
@@ -233,6 +233,7 @@ private:
 
     bool m_modified;
     bool m_dontMarkDirty;
+
     int         m_countUpdatesToIgnore;
 
     // Our text buffer. Convenient, and holds the buffer while using
--- a/include/wx/gtk/textctrl.h
+++ b/include/wx/gtk/textctrl.h
@@ -133,7 +133,7 @@ public:
     bool SetForegroundColour(const wxColour& colour) override;
     bool SetBackgroundColour(const wxColour& colour) override;
 
-    void HandleMaxLength(wxString &text);
+    void GTKHandleMaxLength(wxString &text);
 
     GtkWidget* GetConnectWidget() override;
 
--- a/include/wx/gtk/textctrl.h
+++ b/include/wx/gtk/textctrl.h
@@ -133,6 +133,10 @@ public:
     bool SetForegroundColour(const wxColour& colour) override;
     bool SetBackgroundColour(const wxColour& colour) override;
 
+    // In GTK only single line text control properly handles max
+    // text length. This function will prevent to enter more
+    // than maximum number of characters in the multiline text
+    // control
     void GTKHandleMaxLength(wxString &text);
 
     GtkWidget* GetConnectWidget() override;
--- a/include/wx/gtk/textentry.h
+++ b/include/wx/gtk/textentry.h
@@ -62,7 +62,7 @@ public:
     void SendMaxLenEvent();
     bool GTKEntryOnInsertText(const char* text);
     bool GTKIsUpperCase() const { return m_isUpperCase; }
-
+    int GetMaxLength() const { return m_maxlen; }
     // Called from "changed" signal handler (or, possibly, slightly later, when
     // coalescing several "changed" signals into a single event) for GtkEntry.
     //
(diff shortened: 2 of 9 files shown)


vadz:
It's good to enable the test, but it would be even better if it actually passed. I'm pretty sure it would fail for you locally too if you ran it.

oneeyeman1:
@vadz,

I fixed the test case.
Please review.

If Paul can take a look as well - it would be great.

After this is applied I will put OSX changes on top.

TIA!!

oneeyeman1:
I have no idea why CI fails to build.

And the last one seems unrelated - it is about WebView.

Thank you.

include/wx/gtk/textctrl.h
```diff
@@ -133,6 +133,8 @@ class WXDLLIMPEXP_CORE wxTextCtrl: public wxTextCtrlBase
     bool SetForegroundColour(const wxColour& colour) override;
     bool SetBackgroundColour(const wxColour& colour) override;
 
+    void HandleMaxLength(wxString &text);
```
vadz, line 136: Should have `GTK` prefix as it's GTK-specific and a comment explaining what it does would be welcome.

It should probably also return `bool` to indicate whether it changed the text to avoid useless calls to `SetValue()`.

include/wx/gtk/textctrl.h
```diff
@@ -227,7 +229,6 @@ class WXDLLIMPEXP_CORE wxTextCtrl: public wxTextCtrlBase
 
     bool m_modified;
     bool m_dontMarkDirty;
-
```
vadz, line 230: Useless change should be undone.

include/wx/gtk/textentry.h
```diff
@@ -62,7 +62,7 @@ class WXDLLIMPEXP_CORE wxTextEntry : public wxTextEntryBase
     void SendMaxLenEvent();
     bool GTKEntryOnInsertText(const char* text);
     bool GTKIsUpperCase() const { return m_isUpperCase; }
-
+    int GetMaxLength() const { return m_maxlen; }
```
vadz, line 65: Should have `GTK` prefix as it's GTK-specific.

include/wx/gtk/textentry.h
```diff
@@ -99,7 +99,7 @@ class WXDLLIMPEXP_CORE wxTextEntry : public wxTextEntryBase
     // Call this from the overridden wxWindow::GTKIMFilterKeypress() to use
     // GtkEntry IM context.
     int GTKEntryIMFilterKeypress(GdkEventKey* event) const;
-
+    int m_maxlen;
```
vadz, line 102: Should have a comment, should be initialized, almost certainly should not be defined in between 2 functions without any blank lines around it. Please try to keep things logically organized and more or less readable.

Oh, and it should also be initialized in the declaration.

include/wx/textctrl.h
```diff
@@ -757,6 +757,8 @@ class WXDLLIMPEXP_CORE wxTextCtrlBase : public wxControl,
     // more readable flag testing methods
     bool IsSingleLine() const { return !HasFlag(wxTE_MULTILINE); }
     bool IsMultiLine() const { return !IsSingleLine(); }
+    bool IsMaxLengthAllowed() const { return m_maxLengthAllowed; }
+    void SetMaxLengthAllowed(const bool &allowed) { m_maxLengthAllowed = allowed; }
```
vadz, line 761: It doesn't make sense to pass `bool` by const ref.

include/wx/textctrl.h
```diff
@@ -757,6 +757,8 @@ class WXDLLIMPEXP_CORE wxTextCtrlBase : public wxControl,
     // more readable flag testing methods
     bool IsSingleLine() const { return !HasFlag(wxTE_MULTILINE); }
     bool IsMultiLine() const { return !IsSingleLine(); }
+    bool IsMaxLengthAllowed() const { return m_maxLengthAllowed; }
```
vadz, line 760: What is this function? If it's public, it must be documented. If not, it shouldn't be declared here, should have a `WX` prefix and still should be documented in a comment, at least.

include/wx/textctrl.h
```diff
@@ -805,7 +807,10 @@ class WXDLLIMPEXP_CORE wxTextCtrlBase : public wxControl,
     }
     virtual void SetValue(const wxString& value) override
     {
+       auto old = m_maxLengthAllowed;
+       m_maxLengthAllowed = false;
```
vadz, line 811: This GTK-specific hack should be in GTK-specific code if it's needed at all, not here.

samples/text/text.cpp
```diff
@@ -1343,6 +1347,19 @@ void MyPanel::DoPasteFromClipboard()
 #if wxUSE_LOG
             *m_log << "Successfully retrieved data from the clipboard.\n";
 #endif // wxUSE_LOG
+#ifdef __WXGTK__
```
vadz, line 1350: What does this code do and what is it supposed to demonstrate?

src/common/textentrycmn.cpp
```diff
@@ -229,8 +229,17 @@ void wxTextEntryBase::ChangeValue(const wxString& value)
 
 void wxTextEntryBase::AppendText(const wxString& text)
 {
+    bool old = false;
+    auto textCtrl = dynamic_cast<wxTextCtrl *>( this );
+    if( textCtrl && textCtrl->IsMultiLine() )
+    {
+        old = textCtrl->IsMaxLengthAllowed();
+        textCtrl->SetMaxLengthAllowed( false );
+    }
```
vadz, line 238: Again, all this seems to be GTK-specific and has nothing to do in the common code.

src/gtk/textctrl.cpp
```diff
@@ -876,7 +910,7 @@ bool wxTextCtrl::Create( wxWindow *parent,
             g_signal_connect (m_buffer, "apply_tag",
                               G_CALLBACK (au_apply_tag_callback), nullptr);
 
-            // Check for URLs in the initial string passed to Create
+            // Check for URLs in the initial string sed to Create
```
vadz, line 913: To undo.

P.S. You still don't look at your own diffs, do you?

src/gtk/textctrl.cpp
```diff
@@ -1769,7 +1819,12 @@ void wxTextCtrl::Paste()
     wxCHECK_RET( m_text != nullptr, wxT("invalid text ctrl") );
 
     if ( IsMultiLine() )
+    {
         g_signal_emit_by_name (m_text, "paste-clipboard");
+        auto value = GetValue();
+        HandleMaxLength( value );
+//        text->SetValue( value );
```
vadz, line 1826: Why is this commented out?

src/gtk/textctrl.cpp
```diff
@@ -2315,4 +2370,17 @@ wxTextCtrl::GetClassDefaultAttributes(wxWindowVariant WXUNUSED(variant))
     return GetDefaultAttributesFromGTKWidget(gtk_entry_new(), true);
 }
 
+void wxTextCtrl::HandleMaxLength(wxString &text)
+{
+    auto count = gtk_text_buffer_get_char_count( m_buffer );
+    if( count == m_maxlen )
+        return;
+    auto fullsize = count + text.length();
+    if( (unsigned int) fullsize > (unsigned int) m_maxlen )
```
vadz, line 2379: You shouldn't need 2 casts here.

src/gtk/textentry.cpp
```diff
@@ -272,6 +272,12 @@ DoHandleClipboardCallback( GtkWidget *widget,
         // ourselves in the event handler
         g_signal_stop_emission_by_name (widget, signal_name);
     }
+    wxTextCtrl *text = dynamic_cast<wxTextCtrl *>( win );
```
vadz, line 275: There should be no need for dynamic casts, these classes provide all the virtual functions you need to avoid them.

vadz (changes requested):
There are still a lot of things that need to be fixed here.

I didn't look at this in the details but you probably don't need the "allowed max length" hack at all, what you need to do instead is to block the signal handler when it shouldn't be used (see existing code for many examples of doing it).

You also need to rebase your branch on master (or at least merge master into it) to avoid the CI failures.

oneeyeman1:
@vadz ,

> There are still a lot of things that need to be fixed here.
> 
> I didn't look at this in the details but you probably don't need the "allowed max length" hack at all, what you need to do instead is to block the signal handler when it shouldn't be used (see existing code for many examples of doing it).

MS made this functionality only for text that user types.
If the text is inserted with "AppendText()", i.e. programmatically - no check is done.

So I don't know what signal you are talking about. Could you please explain?

> 
> You also need to rebase your branch on master (or at least merge master into it) to avoid the CI failures.

Will try to do that...

Thx.

vadz:
> MS made this functionality only for text that user types. If the text is inserted with "AppendText()", i.e. programmatically - no check is done.
> 
> So I don't know what signal you are talking about. Could you please explain?

This PR is about GTK changes and you add a signal handler checking that the length is not bigger than the allowed one. It seems a bit strange to explain to you what your own PR does.

oneeyeman1:
@vadz ,
> This PR is about GTK changes and you add a signal handler checking that the length is not bigger than the allowed one

Yes it is.
But calling WriteText() do not go thru the signal.

However, I will see if it can be done like this - thx.

oneeyeman1:
Superseded by 24751.

PR #24069, opened 2023-11-19
Add bullet number and style support to MSW wxTextCtrol via wxTextAttr

Blake-Madden:
Adds bullet support to `wxTextCtrl` under MSW using `wxTextAttr` objects. I mapped all the `wxTextAttrBulletStyle` values that are available in `PARAFORMAT2` and tested under Windows 10 & 11.

One API quirk is that because `wNumberingStyle` only takes one value, setting the number will override styles like parentheses and periods.

### DIFF
--- a/interface/wx/textctrl.h
+++ b/interface/wx/textctrl.h
@@ -373,12 +373,18 @@ public:
 
     /**
         Returns the bullet number.
+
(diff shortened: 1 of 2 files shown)


vadz:
Looks good, thanks! I'll merge it and cherry-pick to 3.2 too.

Issue #23214, opened 2023-02-03
Check for success when saving TIFF images


vadz:
`TIFFClose()` doesn't return anything and so currently `wxTIFFHandler::SaveFile()` returns `true` even if saving actually failed, which can happen for many possible reasons, including an actual IO error or a bug in our code, such as the one fixed in af923360ab (Fix type of values passed to TIFFSetField(), 2022-12-20).

To prevent such bugs from happening again in the future and to detect real errors we need to call `TIFFFlush()`, which does return an error code, first and return `false` if it fails because if the stream had been already flushed, `TIFFClose()` doesn't do anything other than freeing resources, and so really can't fail.

PR #24072, opened 2023-11-20
Fixing issue #23214 this change add use of "TIFFFlush()" to check for success when saving TIFF Images.

BrianDelalex:


vadz (commented):
Thanks! Will apply soon.

Issue #24067, opened 2023-11-18
macOS 14, wx3.2.4: Scrolling a panels contents moves over a neighbour panel.


rocrail:
macOS: 14.1.1 
CPU: M3
wxWidgets: 3.2.4

Just check the attached screenshots:
<img width="500" alt="Screenshot 2023-11-18 at 13 46 09" src="https://github.com/wxWidgets/wxWidgets/assets/67797224/5525fefc-aea4-4435-8edb-93c06999a855">
<img width="500" alt="Screenshot 2023-11-18 at 13 46 18" src="https://github.com/wxWidgets/wxWidgets/assets/67797224/bb559df6-06c5-437c-b1d7-607136386f34">

rocrail:
I checked it also with macOS 10.15; No issue with the same wxWidgets and application.
So this could be an Apple macOS issue?

vadz:
Is it not the same thing as #23916? Please test with the latest 3.2 branch/master, i.e. with #24054 applied, to check it.

rocrail:
I do not how to checkout the git 3.2 master...
git checkout 3.2.master ???

vadz:
It's _either_ the latest 3.2 _or_ master. If you already have a Git checkout, just update to the latest version by using `git pull`. If you don't, please see the standard Git instructions for creating it and clone 3.2 branch if you want to have 3.2 (by default you'd get master).

rocrail:
The git master is AFAIK 3.3.0
So if the fix is in the master I have to wait for a new tag v3.2.5 with a merge from the master fix?

vadz:
The fix was backported to 3.2 branch (which will become 3.2.5) already. Please test with it or just apply the (simple) fix from the linked PR to your local sources.

rocrail:
I checked the patch and the issue is solved.
Thanks!

rocrail:
Sorry, almost. The scrollbars are not clipped:
![wxscrollbar-not-clipped](https://github.com/wxWidgets/wxWidgets/assets/67797224/43e12942-ed98-48db-a0ff-b7244b64e279)

vadz:
@csomor Stefan, not sure if you've noticed this?

@rocrail Do you have some simple way of reproducing this, e.g. do you see this in the scroll sample?

rocrail:
The scroll example is text based and does not show horizontal scrollbars.

vadz:
[...]

PR #24073, opened 2023-11-20
OSX adding native clip view for scrolled windows

csomor:
Sonoma changed clipping behavior, we must use a native clip view to preserve proper clipping of scrolled windows, see #24067

### DIFF
--- a/include/wx/osx/cocoa/private.h
+++ b/include/wx/osx/cocoa/private.h
@@ -223,7 +223,7 @@ public :
 
     virtual void                AdjustClippingView(wxScrollBar* horizontal, wxScrollBar* vertical) override;
     virtual void                UseClippingView(bool clip) override;
-    virtual WXWidget            GetContainer() const override { return m_osxClipView ? m_osxClipView : m_osxView;} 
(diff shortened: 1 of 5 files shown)


vadz:
Stefan, I took the liberty to merge this and I'll backport it to 3.2 too.

BrutalSoloH:
I tried your fix in the 3.2 branch, but then the scrollbars are not visible. Help is much appreciated.

imciner2:
> I tried your fix in the 3.2 branch, but then the scrollbars are not visible. Help is much appreciated.

Do you mean that you checked out the tip of the 3.2 branch and tested it, or that you took this patch and applied it on top of a different 3.2 branch commit?

BrutalSoloH:
> > I tried your fix in the 3.2 branch, but then the scrollbars are not visible. Help is much appreciated.
> 
> Do you mean that you checked out the tip of the 3.2 branch and tested it, or that you took this patch and applied it on top of a different 3.2 branch commit?

I checked out the tips of the 3.2 branch.
I notice there is also a report on this here: https://github.com/wxWidgets/wxWidgets/issues/24151

PR #24075, opened 2023-11-20
wxStatusBar fixes for wxQt

AliKet:


vadz (commented):
Thanks, but having to use Qt-specific `QtAddWidget()` is a bit disappointing. I think it should be possible to avoid it because you could just store the child handle in `m_qtPanes` in (overridden) `AddChild()`. If it isn't, could you please explain why?

AliKet:
> Thanks, but having to use Qt-specific `QtAddWidget()` is a bit disappointing. I think it should be possible to avoid it because you could just store the child handle in `m_qtPanes` in (overridden) `AddChild()`. If it isn't, could you please explain why?

But how to tell the status bar where to place the added widget?
Let's say I have a `wxStatusBar` with four panes for example and I want to display a custom control in the third pane. how does the status bar get this information without a hint like `QtAddWidget()` currently does?

vadz:
> > Thanks, but having to use Qt-specific `QtAddWidget()` is a bit disappointing. I think it should be possible to avoid it because you could just store the child handle in `m_qtPanes` in (overridden) `AddChild()`. If it isn't, could you please explain why?
> 
> But how to tell the status bar where to place the added widget? Let's say I have a `wxStatusBar` with four panes for example and I want to display a custom control in the third pane. how does the status bar get this information without a hint like `QtAddWidget()` currently does?

Oops, I didn't think about this, sorry -- and thanks for the explanation.

We could try some hacks like letting the user-defined `wxEVT_SIZE` handler to execute and then examine all children rectangles and find which `GetFieldRect()` do they lie in, but it's not going to be pretty :-(

So we probably do need some new API, but in this case could we make it portable? I.e. provide some new `wxStatusBar::AddFieldControl(int n, wxWindow*)` and implement this generically by remembering this window and positioning it in `wxEVT_SIZE` as is done now in the sample?

AliKet:
> 
> So we probably do need some new API, but in this case could we make it portable? I.e. provide some new `wxStatusBar::AddFieldControl(int n, wxWindow*)` and implement this generically by remembering this window and positioning it in `wxEVT_SIZE` as is done now in the sample?

Do you mean something like this: #24092 ?

vadz:
Just to confirm: is it ok to merge this one now?

AliKet:
@vadz This is ready now. TIA!

samples/statbar/statbar.cpp
```diff
@@ -940,6 +939,12 @@ MyStatusBar::MyStatusBar(wxWindow *parent, long style)
 
     m_statbmp = new wxStaticBitmap(this, wxID_ANY, wxIcon(green_xpm));
 
+#ifndef __WXQT__
```
vadz, line 942: I think we should get rid of this and use `AddFieldControl()` everywhere and just use an intermediate parent window that would be center the bitmap inside it.

Or is there some problem with doing this?

Using conditional compilation in the samples should really be avoided unless absolutely necessary IMO...

vadz:
Merged now, thanks again!

PR #24080, opened 2023-11-23
Add native wxOverlay implementation under wxQt

AliKet:
As well as various fixes to `wx{Client,Window,Screen}DC`

Note that `wxQt` uses `wxGenericDragImage` and still **doesn't work** even with this PR.
But if we change [RedrawImage](https://github.com/wxWidgets/wxWidgets/blob/978fa7ed06fbba9901912d8c416a91459ce59133/src/generic/dragimgg.cpp#L390) to use `wxClientDC/wxScreenDC` constructed as temporary stack objects then everything works well. As I am not sure how to proceed to make the needed changes, I'll open an issue to record that, sorry!

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -5534,7 +5534,8 @@ COND_TOOLKIT_QT___GUI_SRC_OBJECTS =  \
 	monodll_qt_treectrl.o \
 	monodll_paletteg.o \
 	monodll_qt_datectrl.o \
-	monodll_qt_timectrl.o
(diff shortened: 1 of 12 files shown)


vadz:
It is a bit strange that using "persistent" DC in `wxGenericDragImage` doesn't work, but if this doesn't have any noticeable drawbacks in the other ports, I wouldn't mind changing it to use local DC objects instead.

In any case, the changes here look good to me, so merging, thanks!

PR #24085, opened 2023-11-24
Fix misspellings/repeated words in overviews and some code

Blake-Madden:


docs/contributing/how-to-release.md
```diff
@@ -125,7 +125,7 @@ Update https://www.wxwidgets.org:
   anything else, the page will dynamically show the release files with the
   specified prefixes).
 * Add a news item. Usually a news item is enough but something
-  more can be called for for major releases
+  more can be called for major releases
```
vadz, line 128: Sorry, I don't think this is correct, "called for" is a verb phrase and second "for" is a preposition. The old wording is maybe somewhat infelicitous but the new one is worse because it doesn't make much sense, AFAICS.

vadz (commented):
Thanks! I'll (squash) merge this but without the change mentioned in the other comment, it can always be fixed later.

PR #24086, opened 2023-11-24
Implement maximum number of characters in Multi line text control under OSX

oneeyeman1:
Please review and apply.

### DIFF
--- a/include/wx/osx/textctrl.h
+++ b/include/wx/osx/textctrl.h
@@ -68,7 +68,7 @@ public:
     // operations
     // ----------
 
-
+    virtual void SetMaxLength(unsigned long length) override;
     // sets/clears the dirty flag
     virtual void MarkDirty() override;
     virtual void DiscardEdits() override;
@@ -147,6 +147,7 @@ public:
     void OSXEnableAutomaticDashSubstitution(bool enable);
     void OSXDisableAllSmartSubstitutions();
 
+    unsigned long GetMaxLen() const { return m_maxlen; }
 protected:
     // common part of all ctors
     void Init();
@@ -166,6 +167,8 @@ protected:
 private :
     wxMenu  *m_privateContextMenu;
 
+    unsigned long m_maxlen;
+    
     wxDECLARE_EVENT_TABLE();
 };
 
--- a/include/wx/osx/textctrl.h
+++ b/include/wx/osx/textctrl.h
@@ -168,7 +168,7 @@ private :
     wxMenu  *m_privateContextMenu;
 
     unsigned long m_maxlen;
-    
+
     wxDECLARE_EVENT_TABLE();
 };
 
--- a/src/osx/cocoa/textctrl.mm
+++ b/src/osx/cocoa/textctrl.mm
@@ -472,6 +472,27 @@ NSView* wxMacEditHelper::ms_viewCurrentlyEdited = nil;
 }
 
 
+- (BOOL)shouldChangeTextInRange:(NSRange)affectedCharRange
+              replacementString:(NSString *)replacementString
+{
+    BOOL ret = YES;
+    wxWidgetCocoaImpl* impl = (wxWidgetCocoaImpl* ) wxWidgetImpl::FindFromWXWidget( self );
+    if ( impl )
+    {
+        wxTextCtrl *textCtrl = (wxTextCtrl *) impl->GetWXPeer();
+        if( textCtrl && textCtrl->GetTextPeer() )
+        {
+            if( textCtrl->HasFlag( wxTE_MULTILINE ) )
+            {
+                auto maxlen = textCtrl->GetMaxLen();
+                ret = [super shouldChangeTextInRange:affectedCharRange replacementString:replacementString] &&
+                      (self.string.length + (replacementString.length - affectedCharRange.length) <= maxlen);
+            }
+        }
+    }
+    return ret;
+}
+
 - (void)changeColor:(id)sender
 {
     wxUnusedVar(sender);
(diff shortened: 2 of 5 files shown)


oneeyeman1:
CI failure is in the grid test that is intermittently fails...

vadz:
I'm not sure but I don't think comparing lengths like this is a good idea because it doesn't take selection into account: if you select some text and paste (or type) something, the selected text is removed. E.g. doing this would probably prevent pasting/typing even a single character in a control filled up to capacity when its entire contents is selected.

oneeyeman1:
@vadz ,
> I'm not sure but I don't think comparing lengths like this is a good idea because it doesn't take selection into account: if you select some text and paste (or type) something, the selected text is removed. E.g. doing this would probably prevent pasting/typing even a single character in a control filled up to capacity when its entire contents is selected.

I just tested it - osx 10.13 + Xcode 9 and it works as expected.
What doesn't work however is the multi-line text control hint. I see a blank control

src/osx/cocoa/textctrl.mm
```diff
@@ -562,6 +583,29 @@ - (void)cut:(id)sender
 
 - (void)paste:(id)sender
 {
+    NSPasteboard *pb = [NSPasteboard generalPasteboard];
```
vadz, line 586: Are you sure this code is needed, i.e. `shouldChangeTextInRange:replacementString:` is really not invoked when pasting? I'd expect it to be...

src/osx/cocoa/textctrl.mm
```diff
@@ -562,6 +583,29 @@ - (void)cut:(id)sender
...
+        if( textCtrl && textCtrl->GetTextPeer() )
+        {
+            if( textCtrl->HasFlag( wxTE_MULTILINE ) )
+            {
+                auto maxlen = textCtrl->GetMaxLen();
+                if( textCtrl->GetValue().length() + pbItem.length > maxlen )
+                {
+                    auto range = maxlen - textCtrl->GetValue().length();
+                    pbItem = [pbItem substringToIndex:range];
+                    [self insertText:pbItem];
```
vadz, line 604: This looks wrong: if we insert text here and call `[super paste:]` below too, the text will be inserted twice (assuming `replacementString:` is not called).

src/osx/textctrl_osx.cpp
```diff
@@ -107,7 +107,7 @@ bool wxTextCtrl::Create( wxWindow *parent,
         SetEditable( false ) ;
 
     SetCursor( wxCursor( wxCURSOR_IBEAM ) ) ;
-
+    m_maxlen = UINT_MAX;
```
vadz, line 110: I'd initialize it to 0 to indicate that it's not set and do it in the declaration.

src/osx/textctrl_osx.cpp
```diff
@@ -843,4 +843,10 @@ wxTextProofOptions wxTextWidgetImpl::GetCheckingOptions() const
 void wxTextWidgetImpl::SetJustification()
 {
 }
+
+void wxTextCtrl::SetMaxLength(unsigned long length)
+{
+    if( HasFlag( wxTE_MULTILINE ) )
```
vadz, line 849:
```suggestion
    if ( HasFlag( wxTE_MULTILINE ) )
```

src/osx/cocoa/textctrl.mm
```diff
@@ -472,6 +472,27 @@ - (nullable NSUndoManager *)undoManagerForTextView:(NSTextView *)view
...
+        {
+            if( textCtrl->HasFlag( wxTE_MULTILINE ) )
+            {
+                auto maxlen = textCtrl->GetMaxLen();
+                ret = [super shouldChangeTextInRange:affectedCharRange replacementString:replacementString] &&
+                      (self.string.length + (replacementString.length - affectedCharRange.length) <= maxlen);
+            }
+        }
+    }
+    return ret;
```
vadz, line 493: This is wrong because you don't call super version for multiline controls any more. You need to do it here and return `NO` above when the change is not allowed.

src/osx/cocoa/textctrl.mm
```diff
@@ -472,6 +472,27 @@ - (nullable NSUndoManager *)undoManagerForTextView:(NSTextView *)view
...
+    if ( impl )
+    {
+        wxTextCtrl *textCtrl = (wxTextCtrl *) impl->GetWXPeer();
+        if( textCtrl && textCtrl->GetTextPeer() )
+        {
+            if( textCtrl->HasFlag( wxTE_MULTILINE ) )
+            {
+                auto maxlen = textCtrl->GetMaxLen();
+                ret = [super shouldChangeTextInRange:affectedCharRange replacementString:replacementString] &&
+                      (self.string.length + (replacementString.length - affectedCharRange.length) <= maxlen);
```
vadz, line 489: Not sure about the types here but it's not obvious that this can't overflow, when `replacementString.length < affectedCharRange.length`. You need to compare them first.

src/osx/cocoa/textctrl.mm
```diff
@@ -1018,6 +1062,25 @@ - (void)controlTextDidEndEditing:(NSNotification *)aNotification
...
+    wxWidgetCocoaImpl* impl = (wxWidgetCocoaImpl* ) wxWidgetImpl::FindFromWXWidget( m_textView );
+    if ( impl )
+    {
+        wxTextCtrl *textCtrl = (wxTextCtrl *) impl->GetWXPeer();
+        if( textCtrl && textCtrl->GetTextPeer() )
+        {
+            if( textCtrl->HasFlag( wxTE_MULTILINE ) )
+            {
+                auto maxlen = textCtrl->GetMaxLen();
+                if( textCtrl->GetValue().length() + st.length() > maxlen )
```
vadz, line 1075: Overflow also seems to be possible here. Comparing `st.length()` with `maxlen` first would make it safe.

src/osx/textctrl_osx.cpp
```diff
@@ -843,4 +843,10 @@ wxTextProofOptions wxTextWidgetImpl::GetCheckingOptions() const
...
+        m_maxlen = length;
```
vadz, line 850: What happens if the current length is greater? Should we truncate it or not? What does wxMSW do here?

include/wx/osx/textctrl.h
```diff
@@ -147,6 +147,7 @@ class WXDLLIMPEXP_CORE wxTextCtrl: public wxTextCtrlBase
     void OSXEnableAutomaticDashSubstitution(bool enable);
     void OSXDisableAllSmartSubstitutions();
 
+    unsigned long GetMaxLen() const { return m_maxlen; }
```
vadz, line 150: If this is wxOSX-specific, it should have `OSX` prefix.

include/wx/osx/textctrl.h
```diff
@@ -68,7 +68,7 @@ class WXDLLIMPEXP_CORE wxTextCtrl: public wxTextCtrlBase
     // operations
     // ----------
 
-
+    virtual void SetMaxLength(unsigned long length) override;
```
vadz, line 71: Minor, but this doesn't belong to the "operations" section.

vadz (commented):
Please make a function taking the current length and the length of the text being added and checking if there is enough space left and call it from all places where it's necessary because verifying that all these comparisons work/don't suffer from overflows is difficult.

PR #24087, opened 2023-11-24
Fix crash in DoShowMultipleLogMessages() with  wxUSE_LOG_DIALOG set to 0

Randalphwa:
This PR fixes a crash when `DoShowMultipleLogMessages()` is called with `wxUSE_LOG_DIALOG` set to 0. The code after the #else statement gets the size of the `wxArrayString` but then instead of accessing that array, it tried to access the `m_aMessages` array which is empty, resulting in a crash in the C runtime. This one-line commit simply gets the strings from the correct array.

I was able to verify the fix in an application I have that has wxUSE_LOG_DIALOG set to 0 and a situation where 26 warnings were in the array passed to DoShowMultipleLogMessages(). It works fine with this change.


vadz:
Thanks for fixing this, will merge and backport to 3.2 soon.

PR #24089, opened 2023-11-25
Restore ToCDouble compatibility with leading spaces/+ characters and hex

imciner2:
The previous ToCDouble() function accepted leading spaces, a starting + sign, and hex strings. std::from_chars() does not accept spaces or +, and requires a special flag to convert hex strings.

Sample strings that didn't work before can be seen in the new test cases. I also added test cases for infinity/nan, since they are also strings that should be ensured are supported.

### DIFF
--- a/src/common/string.cpp
+++ b/src/common/string.cpp
@@ -1678,9 +1678,20 @@ bool wxString::ToCDouble(double *pVal) const
     wxCHECK_MSG( pVal, false, "null output pointer" );
 
     const wxScopedCharBuffer& buf = utf8_str();
-    const auto start = buf.data();
+    auto start = buf.data();
     const auto end = start + buf.length();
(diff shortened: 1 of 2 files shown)


imciner2:
The wxMac build failure seems to be a runner failure. Can someone restart that job?

vadz:
Done, but I don't know why this (and only this, AFAICS) Mac build keeps failing. @csomor Would you have any idea by chance?

csomor:
> Done, but I don't know why this (and only this, AFAICS) Mac build keeps failing. @csomor Would you have any idea by chance?

@vadz not really, I'm sorry, I cannot find out where it got stuck from looking at the logs ... at least we are not the only ones ...
https://github.com/actions/runner/issues/2140

vadz:
Thanks Ian, will merge it in a moment.

PR #24090, opened 2023-11-25
Fix double negatives used with 'neither'

lanurmi:
Replace most 'neither' words introduced after
66916c7 (Fix double negatives used with 'neither' in docs and comments, Mon Aug 15 16:36:37 2022 +0300) with 'either'.

All changes are to comments only.


vadz:
Argh, sorry for most (all?) of those. Will merge soon, thanks.

lanurmi:
> Argh, sorry for most (all?) of those. Will merge soon, thanks.

Thanks! Having observed how you use the word (on mailing lists, too), I think you would get it nearly always right, if you simply typed 'either' whenever you are about to type 'neither'. :)

PR #24094, opened 2023-11-27
Add US Catholic Feasts / Christian holiday authority classes

Blake-Madden:
This includes a static function to calculate Easter that can be used for other authorities.
Document the `wxDateTimeWorkDays` and `wxDateTimeHolidayAuthority` classes and how to connect derived `wxDateTimeHolidayAuthority` classes to the system.
Fix comment in `wxDateTimeHolidayAuthority` that references a base `DoGetHolidaysInRange()` that doesn't exist.

Note that these are only the feasts which are holy days of obligation, so would be considered holidays from work.


vadz (commented):
Great, thank you for your contribution!

And nice to see that a 25 year old TODO comment can still be addressed :-)

Globally this looks good, but at least the error in wxMSW build

```
./src/common/datetime.cpp:2302:26: error: ‘floor’ is not a member of ‘std’
```

seems quite real, so it would need to be fixed by include `<math>`.

include/wx/datetime.h
```diff
@@ -1625,6 +1633,55 @@ class WXDLLIMPEXP_BASE wxDateTimeWorkDays : public wxDateTimeHolidayAuthority
...
+        return m_holyDaysOfObligation.find(dt) != m_holyDaysOfObligation.cend() ||
+            dt.IsSameDate(GetEaster(dt.GetYear())) ||
+            dt.IsSameDate(GetThursdayAscension(dt.GetYear()));
+    }
+
+    size_t DoGetHolidaysInRange(const wxDateTime& dtStart,
+                                const wxDateTime& dtEnd,
+                                wxDateTimeArray& holidays) const override;
+private:
+    static std::map<wxDateTime, wxString, wxHolidayLess> m_holyDaysOfObligation;
```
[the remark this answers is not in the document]
Blake-Madden: Done

src/common/datetime.cpp
```diff
@@ -2271,6 +2271,115 @@ size_t wxDateTimeWorkDays::DoGetHolidaysInRange(const wxDateTime& dtStart,
...
 }
 
+// ----------------------------------------------------------------------------
+// wxDateTimeUSCatholicFeasts
+// ----------------------------------------------------------------------------
+
+std::map<wxDateTime, wxString, wxHolidayLess> wxDateTimeUSCatholicFeasts::m_holyDaysOfObligation =
+{
+    // Feasts with fixed dates
+    { wxDateTime(1, wxDateTime::Month::Jan, 0), _T("Solemnity of Mary, Mother of God") },
```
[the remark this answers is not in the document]
Blake-Madden: I just turned those into comments since it's a vector now.

Blake-Madden:
I also added a new `wxDateTimeChristianHolidays` class, which will just find Easter and Christmas.

include/wx/datetime.h
```diff
@@ -1565,6 +1567,18 @@ using wxDateTimeArray = wxBaseArray<wxDateTime>;
 //     virtual methods to work with the holidays they correspond to.
 // ----------------------------------------------------------------------------
 
+// Compares dates only on day and month.
+// Useful for finding dates within a map of holidays.
+class wxHolidayLess
```
vadz, line 1572: This doesn't seem used any longer now that we don't have a map any more, do you think it's still worth adding it? I'd rather omit it.

src/common/datetime.cpp
```diff
@@ -2271,6 +2272,142 @@ size_t wxDateTimeWorkDays::DoGetHolidaysInRange(const wxDateTime& dtStart,
...
+    // when D is 28
+    else if ((D == 28) && (E == 6))
+    {
+        wxASSERT_MSG(
+            wxDateTime(18, wxDateTime::Apr, year).GetWeekDay() ==
+            wxDateTime::WeekDay::Sun,
+            "Error in Easter calculation!");
+        return wxDateTime(18, wxDateTime::Apr, year);
+    }
+    else {
```
vadz, line 2337: Sorry for nitpicking, but
```suggestion
    else
    {
```

src/common/datetime.cpp
```diff
@@ -2271,6 +2272,142 @@ size_t wxDateTimeWorkDays::DoGetHolidaysInRange(const wxDateTime& dtStart,
...
+size_t wxDateTimeChristianHolidays::DoGetHolidaysInRange(const wxDateTime& dtStart,
+                                                         const wxDateTime& dtEnd,
+                                                         wxDateTimeArray& holidays) const
+{
+    holidays.Clear();
+
+    for (wxDateTime dt = dtStart; dt <= dtEnd; dt += wxDateSpan::Day())
+    {
+        if (dt.IsSameDate(GetEaster(dt.GetYear())) ||
+            (dt.GetMonth() == 12 && dt.GetDay() == 25))
```
vadz, line 2401: Is there any reason not to use `DoIsHoliday()` here? I could understand it if we optimized this by precomputing Easter date only once outside of the loop, but we don't even do this...

vadz (commented):
Nice, thanks for the improvements!

I can merge this already, making the couple of small fixes mentioned below myself, but please let me know if you prefer to do it yourself.

include/wx/datetime.h
Blake-Madden: Removed now.

src/common/datetime.cpp
Blake-Madden: Fixed

src/common/datetime.cpp
Blake-Madden: Got it, fixed it in both classes.

vadz:
Thanks for the final fixes, will (squash)merge in a moment with just a couple of minor cosmetic changes.

PR #24097, opened 2023-11-27
Fix ``wxBitmapBundle::FromSVG()``, which renders incorrectly on GTK

### LINKED ISSUE #24064: wxBitmapBundle from SVG renders incorrectly on wxGTK
### Description
If I set an icon in an generic About box on wxGTK from a wxBitmapBundle that was created from SVG, using any theme with a light-colored background, the anti-aliased edges have dark fringes, as if the objects were rendered on a dark background. This problem doesn't happen on wxOSX/Cocoa or wsMSW. See example below and screenshot attached.

This doesn't appear to be a problem with SVG rendering in general, since the object renders correctly with a background-matching object behind it (see commented out rectangle in example). Also, if I create the wxBitmapBundle from a PNG with alpha-channel transparency, rendering is also correct.

This appears to be a wxGTK bug. A bug fix would be great, but I’m also looking for a workaround that still allows me to use SVG art in the provided About box.

```
#include <wx/wx.h>
#include <wx/bmpbndl.h>
#include <wx/aboutdlg.h>
#include <wx/generic/aboutdlgg.h>

class MyApp: public wxApp
{
public:
    virtual bool OnInit();
};

wxIMPLEMENT_APP(MyApp);

{
public:
    MyFrame();
};

bool MyApp::OnInit()
{
    MyFrame *frame = new MyFrame();
    frame->Show();

    static char sample_icon[] = "<svg viewBox=\"0 0 1000 1000\">"
//      "<rect width=\"100%\" height=\"100%\" fill=\"#d4d4d4\"/>"
      "<circle cx=\"50%\" cy=\"50%\" r=\"45%\" fill=\"#B0E0E6\"/>"
      "</svg>";
    wxBitmapBundle bb = wxBitmapBundle::FromSVG(sample_icon, wxSize(256, 256));

    wxAboutDialogInfo aboutInfo;
    aboutInfo.SetName(" ");
    aboutInfo.SetIcon(bb.GetIcon(wxDefaultSize));

    wxGenericAboutDialog *about_dlg = new wxGenericAboutDialog(aboutInfo, frame);
    about_dlg->Show();

    return true;
}

MyFrame::MyFrame() : wxFrame(nullptr, wxID_ANY, "About test")
{
}
```
About_SVG_bug_ss

- wxWidgets version you use: 3.2.2.1 (stock Fedora 39 package)
- wxWidgets port you use: wxGTK
- OS and its version: Fedora 39 Linux

+ GTK version: 3.24.38
+ Which GDK backend is used: X11
+ Current theme: MATE BlueMenta (but I've tried many others, and also with GNOME)

teliseo:
Due to incorrect premultiplication of raw pixel color values by alpha on GTK, which stores ``wxBitmap`` colors at full value (“straight alpha”), ``wxBitmapBundle::FromSVG()`` renders partial transparency incorrectly, mostly seen as dark anti-aliased edges when a light-colored SVG object is drawn on a light background. This same incorrect premultiplication is also seen in the **Test raw bitmap...** graphic of the _image_ sample application.

In addition to fixing these problems, this PR defines the symbol ``wxHAS_PREMULTIPLIED_ALPHA`` for appropriate platforms (currently MSW, OSX, and Qt) and it is now used to replace platform-specific conditionals. This PR also adds a unit test for the proper application of premultiplication by ``wxBitmapBundle::FromSVG()``.

@vadz has suggested that a better solution would be abstraction of premultiplication in ``wxAlphaPixelData``. This isn’t attempted in this PR, and it needs further discussion.

Resolves #24064.

### DIFF
--- a/docs/doxygen/mainpages/const_cpp.h
+++ b/docs/doxygen/mainpages/const_cpp.h
@@ -210,6 +210,10 @@ Currently the following symbols exist:
     Windows resource files</a> resource files are available on the current platform.
     Usually wxHAS_IMAGE_RESOURCES should be used instead.}
 @itemdef{wxHAS_POWER_EVENTS, Defined if wxPowerEvent are ever generated on the current platform.}
+@itemdef{wxHAS_PREMULTIPLIED_ALPHA, Defined if wxBitmap stores raw pixel color
+    values premultiplied (scaled) by alpha. This is generally used to switch
+    whether to premultiply before storing pixels via wxAlphaPixelData converted
+    from a straight alpha source.}
 @itemdef{wxHAS_RADIO_MENU_ITEMS,
         Defined if the current port supports radio menu items (see wxMenu::AppendRadioItem).}
 @itemdef{wxHAS_RAW_BITMAP, Defined if direct access to bitmap data using the classes in @c wx/rawbmp.h is supported.}
--- a/include/wx/rawbmp.h
+++ b/include/wx/rawbmp.h
@@ -146,12 +146,19 @@ struct wxPixelFormat
 // wxImage format is common to all platforms
 typedef wxPixelFormat<unsigned char, 24, 0, 1, 2> wxImagePixelFormat;
 
+// wxPIXEL_FORMAT_ALPHA is the offset of alpha (defined if the format has
+// alpha).
+
+// wxHAS_PREMULTIPLIED_ALPHA is defined if R, G, and B are stored premultiplied
+// (scaled) by alpha, otherwise they have full value ("straight alpha").
+
 // the (most common) native bitmap format without alpha support
 #if defined(__WXMSW__)
     // under MSW the RGB components are reversed, they're in BGR order
     typedef wxPixelFormat<unsigned char, 24, 2, 1, 0> wxNativePixelFormat;
 
     #define wxPIXEL_FORMAT_ALPHA 3
+    #define wxHAS_PREMULTIPLIED_ALPHA
 
     template<>
     struct wxPixelFormat<void, 1, -1, -1, -1, -1, bool>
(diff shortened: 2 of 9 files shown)


vadz (approved):
Thanks a lot, looks good to me!

I'd just like to add the documentation of the new symbol to `docs/doxygen/mainpages/const_cpp.h` (but I can do it myself when merging) and I'd also like to check with @AliKet if wxQt doesn't use premultiplied alpha as well, because I think it does, and so it should be defined for it too (but this shouldn't delay merging this PR, as it can always be done later, e.g. if it results in any Qt-specific unit test failures).

teliseo:
``wxHAS_PREMULTIPLIED_ALPHA`` is defined for wxQt in my commits, which corresponded to most previous conditionals.

vadz:
> `wxHAS_PREMULTIPLIED_ALPHA` is defined for wxQt in my commits, which corresponded to most previous conditionals.

Oops, sorry, I must have misread the diff.

teliseo:
My changes passed my new ``BitmapBundle::FromSVG-alpha`` test on MSW, OSX, GTK, and Qt, so I think that is good validation. The only remaining platform with ``wxHAS_RAW_BITMAP`` defined is DFB, which has a validation test, but it doesn’t run ``test_gui``.

I also pushed a commit with doc for ``wxHAS_PREMULTIPLIED_ALPHA`` added.

vadz:
Great, thanks a lot for your work, will merge in a moment!

DirectFB doesn't use X11 so it can't be run inside Xvfb but in any case this port is almost dead and nobody is using it AFAIK, so it doesn't matter that much.

teliseo:
This seems to me like the kind of bug that should also be fixed on the 3.2 branch. Should I open a PR for that, since I already have a branch ready?

teliseo:
@vadz I see now that you have a different process for backports, and that you only picked the most essential patch.

vadz:
Yes, AFAICS this should fix the immediate problem and the rest is not critical, so I prefer not to bother with it (and maybe risk breaking something).

I'll merge the backports branch soon, thanks again for your contribution!

PR #24099, opened 2023-11-29
Fix missing wxLC_EDIT_LABELS style for list ctrl

mcorino:


vadz:
Sorry, but doesn't this mean that the labels can now be edited even if `wxEL_ALLOW_EDIT` is not specified?

And if we can't have one without the other, maybe we should just assert in this case and document that `wxEL_ALLOW_NEW` requires `wxEL_ALLOW_EDIT`?

vadz:
> > Sorry, but doesn't this mean that the labels can now be edited even if `wxEL_ALLOW_EDIT` is not specified?
> 
> No, it does not, as without wxEL_ALLOW_EDIT the 'Edit item' button will not be available.

But wouldn't clicking the label start editing it nevertheless?

> Of course the question is; is there a real use case for separate `wxEL_ALLOW_EDIT` and `wxEL_ALLOW_NEW` styles? I am not sure, but as long as the styles can be separately specified I think this patch applies.

I think `EDIT` without `NEW` makes sense, but not vice versa, which is why I believe that `NEW` should either turn on `EDIT` implicitly or fail if it's not specified. The former would be more compatible, so I'd prefer doing this if there is no strong reason not to.

mcorino:
> > Ah, as it turns out it does. I did not notice that at first as that functionality seems to have a problem all of its own. On GTK3 I cannot get it to trigger without a **triple** click!
> 
> This is because you need to use a "slow click" to start editing, i.e. it's _not_ supposed to start on double click, but when you click once and then wait a bit without moving the mouse (too much). I'm not sure if this is the best from the UI point 
of view, but this is how it works in Windows...

Ugh, horrible. No matter, I like the buttons better anyway.

> 
> > I agree. I prefer implicitly adding 'EDIT' for 'NEW'. Would you like me to make that change (and perhaps add a not to the docs)?
> 
> Yes, please. TIA!

The changes are in.
Would be nice to have this backported.

PR #24101, opened 2023-11-29
Clean-up the .props files

PBfordev:
Remove the stuff regarding no long supported MSVS versions, Itanium CPU architecture, and Windows XP platform toolset.


vadz:
Thanks, looks good to me!

PR #24112, opened 2023-12-02
add wxNativeContainerWindow to wxTopLevelWindows

wsu-cb:
to ensure that wxDialog::ShowModal() disables wxNativeContainerWindow


wsu-cb:
~wxTopLevelWindowBase(), a base class for wxNativeContainerWindow, remove itself from the wxTopLevelWindows list, and running in a debugger verifies that it does the job.  (In fact, ~wxWindowBase() redundantly also removes itself from the list.)

vadz:
Thanks for checking/confirming this! I'll merge this in master soon and in principle it could be backported to 3.2, if you think it's worth it.

PR #24113, opened 2023-12-03
Add missing default constructor of wxFileDialog on MSW

Kumazuma:
There is the contructor on other backends but not MSW.


vadz:

PR #24123, opened 2023-12-08
Misc mac fixes

vadz:
Some minor changes to get rid of a few warnings in Mac build using older SDK.

Stefan, do you agree with removing `bool clip`? I could just suppress the warning about it being unused, if you think it's useful to keep it, but as it doesn't do anything, I'd prefer to remove it to make things more clear.

### DIFF
--- a/include/wx/osx/cocoa/private.h
+++ b/include/wx/osx/cocoa/private.h
@@ -221,7 +221,7 @@ public :
     virtual void                controlTextDidChange();
 
     virtual void                AdjustClippingView(wxScrollBar* horizontal, wxScrollBar* vertical) override;
-    virtual void                UseClippingView(bool clip) override;
+    virtual void                UseClippingView() override;
     virtual WXWidget            GetContainer() const override { return m_osxClipView ? m_osxClipView : m_osxView; }
 
 protected:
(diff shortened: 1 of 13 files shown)


csomor (commented):
I don't understand this correctly, the new constants in the new SDKs **are** named differently starting from 11 in AvailabilityVersions.h , shouldn't we then change the place where they are used ?

```
#define  MAC_OS_X_VERSION_10_16                          __MAC_10_16
#define  MAC_OS_VERSION_11_0                             __MAC_11_0

```

csomor (approved):
I usually like having the bool param in a setter, but since we would probably never want to revert it after having set it, I agree that we can simplify things this way

vadz:
> I don't understand this correctly, the new constants in the new SDKs **are** named differently starting from 11 in AvailabilityVersions.h , shouldn't we then change the place where they are used ?
> 
> ```
> #define  MAC_OS_X_VERSION_10_16                          __MAC_10_16
> #define  MAC_OS_VERSION_11_0                             __MAC_11_0
> ```

I didn't realize Apple renamed them. But in our code we only use `MAC_OS_X_VERSION_XXX`, and never `MAC_OS_VERSION_XXX`. IMHO it would be better to use consistent names, so if you prefer to drop the `_X` part, we'd need to remove it from all (27 at the last count) the existing tests. I could do it if you want, but I'm not sure it's worth it...

csomor:
Why can't we just use the versions Apple defines ? starting from 11 the _X_ is dropped, before that it's there, so I think we don't have to change things, do we ? I have not found occurrences

PR #24124, opened 2023-12-08
Add wxGrid::CopySelection()

Blake-Madden:
The was the same code from CTRL+C. I just refactored it into a separate function that the client can call now.

### DIFF
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -6095,7 +6095,12 @@ void wxGrid::OnKeyDown( wxKeyEvent& event )
             case 'C':
                 if ( event.GetModifiers() == wxMOD_CONTROL )
                 {
-                    CopySelection();
+                    if ( !CopySelection() )
+                    {
+                        wxLogWarning(_("Error copying grid to the clipboard. "
+                            "Either selected cells were not contiguous or "
+                            "no cell was selected."));
+                    }
                     break;
                 }
                 wxFALLTHROUGH;
@@ -10826,8 +10831,6 @@ bool wxGrid::CopySelection()
     wxClipboardLocker lockClipboard;
     if (!lockClipboard)
     {
-        // Error message should have been already given and we
-        // don't have much to add.
         return false;
     }
 
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -10787,7 +10787,7 @@ void wxGrid::ClearSelection()
         m_selection->ClearSelection();
 }
 
-void wxGrid::CopySelection()
+bool wxGrid::CopySelection()
 {
 #if wxUSE_CLIPBOARD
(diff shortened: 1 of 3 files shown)


src/generic/grid.cpp
```diff
@@ -10852,6 +10787,77 @@ void wxGrid::ClearSelection()
...
+        if (++iter != blocks.end())
+        {
+            // As we use simple text format, we can't copy more
+            // than one block to clipboard.
+            wxLogWarning
+            (
+                _("Copying more than one selected block "
+                    "to clipboard is not supported.")
+            );
+            return;
```
vadz, line 10827: I think it's not very helpful for the new function to give this warning, so I'd consider changing the code further to make the new function return `bool` indicating whether anything has been copied or not and giving this message in the original event handler (alternative: make the function return the error message, as string, with empty string meaning "no error", but it looks like an overkill here).

If not, the documentation should at least mention that the function logs a warning if there is more than one selected block.

vadz (commented):
Thanks, extracting this function looks like a good idea, but I'm not sure about the warning.

src/generic/grid.cpp
```diff
@@ -6095,72 +6095,7 @@ void wxGrid::OnKeyDown( wxKeyEvent& event )
...
-                                buf += '\t';
-
-                            buf += GetCellValue(row, col);
-                        }
-
-                        buf += wxTextFile::GetEOL();
-                    }
-
-                    wxTheClipboard->SetData(new wxTextDataObject(buf));
+                    CopySelection();
```
vadz, line 6098: Sorry, I think there was a misunderstanding, I meant to do this:
```suggestion
                    if ( !CopySelection() )
                    {
                        wxLogWarning(...);
                    }
```

vadz (commented):
Sorry, I didn't meant to remove the warning, just move it out from this function.

src/generic/grid.cpp
```diff
@@ -10852,6 +10787,74 @@ void wxGrid::ClearSelection()
...
+            // than one block to clipboard.
+            return false;
+        }
+    }
+
+    wxClipboardLocker lockClipboard;
+    if (!lockClipboard)
+    {
+        // Error message should have been already given and we
+        // don't have much to add.
```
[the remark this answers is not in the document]
Blake-Madden: I removed this and make the CTRL+C log message more encompassing (but still informative).

vadz:
Thanks, looks good to me, will merge in a moment.

vadz:
Sorry, after further thought I've realized that I am not sure with the latest commit added here. While I agree that it's useless when copying just a single cell, I think the trailing newline should be preserved when copying multiple cells.

So for now I've (squash) merged the rest, but please resubmit the latest commit as another PR preserving the multiple cells case unless you have a convincing reason to remove it in that case too.

TIA!

PR #24126, opened 2023-12-10
Fix splitter in high DPI under wxGTK

vadz:
cc @MaartenBent

### DIFF
--- a/samples/splitter/splitter.cpp
+++ b/samples/splitter/splitter.cpp
@@ -226,8 +226,7 @@ wxEND_EVENT_TABLE()
 
 // My frame constructor
 MyFrame::MyFrame()
-       : wxFrame(nullptr, wxID_ANY, "wxSplitterWindow sample",
-                 wxDefaultPosition, wxSize(420, 300))
+       : wxFrame(nullptr, wxID_ANY, "wxSplitterWindow sample")
 {
     SetIcon(wxICON(sample));
 
@@ -331,12 +330,14 @@ MyFrame::MyFrame()
     m_splitter->Initialize(m_left);
 #else
     // you can also try -100
-    m_splitter->SplitVertically(m_left, m_right, 100);
+    m_splitter->SplitVertically(m_left, m_right, FromDIP(100));
 #endif
 
 #if wxUSE_STATUSBAR
     SetStatusText("Min pane size = 0", 1);
 #endif // wxUSE_STATUSBAR
+
+    SetInitialSize(FromDIP(wxSize(420, 300)));
 }
 
 MyFrame::~MyFrame()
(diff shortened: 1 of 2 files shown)


samples/splitter/splitter.cpp
```diff
@@ -625,13 +621,13 @@ void MyCanvas::OnDraw(wxDC& dcOrig)
     wxMirrorDC dc(dcOrig, m_mirror);
 
     dc.SetPen(*wxBLACK_PEN);
-    dc.DrawLine(0, 0, 100, 200);
+    dc.DrawLine(0, 0, FromDIP(100), FromDIP(200));
```
MaartenBent, line 624: These should use `dc.FromDIP()`.

MaartenBent (commented):
I would put the `wxHAS_DPI_INDEPENDENT_PIXELS` check inside `OnDPIChanged`, so you only have it in 1 place instead of 3. At the expense of having an unneeded entry in the event table.

On Windows the spiller sample still works correctly (it looks even a bit better).
I don't have a setup to test gtk/osx.

samples/splitter/splitter.cpp
vadz: I think it should be exactly the same (the DC passed here is associated with this window anyhow), but maybe it will be indeed more clear to use `dc.`, thanks.

vadz:
> I would put the `wxHAS_DPI_INDEPENDENT_PIXELS` check inside `OnDPIChanged`, so you only have it in 1 place instead of 3. At the expense of having an unneeded entry in the event table.

I hesitated doing it like this but then decided that there was no reason to have an empty handler on these platforms. It would make sense to have it if we ever needed to handle the DPI change event there too, but I don't see why would we...

> On Windows the spiller sample still works correctly (it looks even a bit better). I don't have a setup to test gtk/osx.

I did test there. Under wxGTK it works now, and under wxOSX it still works but it wasn't broken there even before because it doesn't send a DPI changed event on startup, unlike wxGTK does (and I didn't change the DPI of the system to test this as I'm using it remotely and it's not easy -- or even possible? -- to do it). This probably explains why nobody had noticed it before.

samples/splitter/splitter.cpp
MaartenBent: Yeah, here it will be the same. But for example in the drawing sample the dc can also be a bitmap. So I try to be consistent.

MaartenBent:
> and I didn't change the DPI of the system to test this as I'm using it remotely and it's not easy -- or even possible? -- to do it

I don't know about remotely, but I used to have a Ubuntu virtualbox, and it does allow to change the dpi dynamically.
https://www.omgubuntu.co.uk/2020/04/ubuntu-20-04-fractional-scaling-support-setting

vadz:
> > and I didn't change the DPI of the system to test this as I'm using it remotely and it's not easy -- or even possible? -- to do it
> 
> I don't know about remotely, but I used to have a Ubuntu virtualbox, and it does allow to change the dpi dynamically. https://www.omgubuntu.co.uk/2020/04/ubuntu-20-04-fractional-scaling-support-setting

I'm using Mac via VNC. Under Linux I have no trouble testing it locally.

samples/splitter/splitter.cpp
vadz: Changed now (and I've also followed your other advice to minimize the number of preprocessor checks).

PR #24127, opened 2023-12-11
Generic tree icon fixes

vadz:
This notably fixes broken display of icons in high DPI in legacy code, i.e. code using `wxImageList` and not `wxBitmapBundle`, after 53b4f4ddf2 (cc @craftyjon I don't think this affects your use, but please let me know if I'm wrong).

The rest is just some minor cleanup/appearance improvements.


craftyjon:
I won't be able to test this anytime soon, but just by reading it, I'm not concerned about it, seems fine!

PR #24128, opened 2023-12-11
Mac build and docs improvements

vadz:
Trying to explain to somebody how to build wxWidgets for use with their application, I realized that our documentation, and even our `configure`, is not optimal, to put it mildly.

I still struggle to explain to people how to use wxWidgets from Xcode, so I didn't document this here, but if anybody can provide some good instructions for this, it would be great.

### DIFF
--- a/docs/osx/install.md
+++ b/docs/osx/install.md
@@ -31,22 +31,18 @@ build on modern systems. So in practice you should use a command like this:
 You may also prefer to add `-s` option to avoid normal output from make and/or
 redirect it you to a log file for further inspection.
 
-Build the samples and demos
+You should build at least the smallest possible wxWidgets sample to verify that
+everything is working as intended, by doing
+
+    cd samples/minimal
+    make
+
+and then running `minimal.app` from this directory from Finder.
+
+If you'd like to, you can also build all the other samples and demos
 
     cd samples; make;cd ..
     cd demos;   make;cd ..
-
-After the compilation completes, use Finder to run the samples and demos
-* Go to build-cocoa-debug/samples to experiment with the Cocoa samples.
-* Go to build-cocoa-debug/demos to experiment with the Cocoa demos.
-* Double-click on the executables which have an icon showing three small squares.
-* The source code for the samples is in wxWidgets/samples
-* The source code for the demos is in wxWidgets/demos
-
-More information about building on macOS is available in the wxWiki.
-Here are two useful links
-  * https://wiki.wxwidgets.org/Guides_%26_Tutorials
-  * https://wiki.wxwidgets.org/Development:_wxMac
 
 
 Advanced topics                        {#osx_advanced}
--- a/docs/osx/install.md
+++ b/docs/osx/install.md
@@ -48,6 +48,24 @@ If you'd like to, you can also build all the other samples and demos
 Advanced topics                        {#osx_advanced}
 ===============
 
+Building library for distribution
+---------------------------------
+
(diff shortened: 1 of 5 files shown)


docs/osx/install.md
```diff
@@ -9,39 +9,63 @@ Most OS X developers should start by downloading and installing Xcode
...
 
-Build the samples and demos
+It is recommended to use `-jN` option with the last command, where `N` is a
+number of the processors in your system (which can be checked using `sysctl -n
+hw.ncpu` command if you are not sure), as this will dramatically speed up the
+build on modern systems. So in practice you should use a command like this:
 
-    cd samples; make;cd ..
-    cd demos;   make;cd ..
+    make -j16
```
lanurmi, line 29: Because `-j16` requires quite a high-end Mac system to be useful, and some people are likely to just copy-paste `-j16` without thinking how many cores or how much memory they have, I propose using a smaller figure in this example. Such as 6 or 8.

vadz: OK, done, and added an extra warning about this, thanks.

vadz:
@csomor just a ping to ask you if you could please check that [the change to configure](https://github.com/wxWidgets/wxWidgets/pull/24128/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810) is correct. TIA!

csomor (commented):
Hi @vadz, thanks for this change, I don't know how many people just building universal without indicating archs ... since only from macOS 11 arm64 can be built ... should we make this reflecting the OS it's run on to avoid error messages ?

vadz:
> Hi @vadz, thanks for this change, I don't know how many people just building universal without indicating archs ... since only from macOS 11 arm64 can be built ... should we make this reflecting the OS it's run on to avoid error messages ?

Well, this is the question... I think people using master are very very likely to target arm64, but I can't be sure, of course. If you think it's worth making this conditional on the OS version, I'd need to understand how to test for it (using `sw_vers -productVersion`, I guess, but what are the formats used for different versions? or maybe `uname -r`?) and what should be used for which version.

I definitely think we should remove `i386` because it's not even supported any longer.

csomor:
> > Hi @vadz, thanks for this change, I don't know how many people just building universal without indicating archs ... since only from macOS 11 arm64 can be built ... should we make this reflecting the OS it's run on to avoid error messages ?
> 
> Well, this is the question... I think people using master are very very likely to target arm64, but I can't be sure, of course. If you think it's worth making this conditional on the OS version, I'd need to understand how to test for it (using `sw_vers -productVersion`, I guess, but what are the formats used for different versions? or maybe `uname -r`?) and what should be used for which version.

the first segment of `sw_vers -productionVersion` returning 10 would be true for any system before macOS 11 Big Sur.
Alternatively we could ask for `gcc --version` and treat everything before Xcode 12.2 as not supporting arm64 (this is a little bit more precise as Catalina supports running Xcode up to 12.4, which can produce a universal binary 2.

[Xcode versions](https://trac.macports.org/wiki/XcodeVersionInfo)

But if you think this is more effort than it deserves, we just put a note somewhere we can link to, in case questions arise ..

> I definitely think we should remove `i386` because it's not even supported any longer.

yes

vadz:
Thanks, I've added a test for the OS version as this was simpler than checking gcc version.

vadz:
I'm going to merge this, finally. Please let me know if I broke/misdocumented anything.

PR #24131, opened 2023-12-11
Make wxSTAY_ON_TOP TLWs transient for parent too in wxGTK

vadz:
Normally such windows should be remain on top of all the other windows, but at least under Wayland they do not, so make them at least remain on top of their parent window -- which should do no harm anyhow and might do some good (by allowing the WM to associate them with the correct window) even if they do stay on top of everything.

----

@paulcor Any objections to merging this? Without it, `wxBusyInfo` window can get hidden beyond its parent, at least under Wayland. I could change its code to use `wxFRAME_FLOAT_ON_PARENT`, of course, but it looks like this would be the desired behaviour for any `wxSTAY_ON_TOP` window, wouldn't it?


paulcor:
LGTM

Issue #24129, opened 2023-12-11
wxQt: button state looks odd


davidkaroly:
#### Bug description:

In wxQt, Button state is reported as State_Current when it should be State_Normal i.e. not hovering.

This also results in 'empty buttons' e.g. in the life demo program as the bitmap is not refreshed.

#### Expected vs observed behaviour:

* Bitmap button e.g. in the widgets sample show hover mode instead of  normal mode bitmap.
* Life demo program should show a navigation mini-window with arrow buttons. However, it shows empty buttons.

#### Stack trace:

n/a, no crash

#### Patch or snippet allowing to reproduce the problem:

Build demos and samples with wxQt.

Run demos/life/life. Navigation mini-window shows empty buttons.

Run samples/widgets/widgets.
* Select widget: Button, Set style: Bitmap only. Sample button shows up in hover state even though it should show normal state when we're not hovering.
* Select widget: EditableListbox, Set style: Allow new items, Allow editing items, Allow deleting items. Some of the little buttons on the top right side of the editable list box show up empty.

### Platform and version information

- wxWidgets version: latest master
- wxWidgets port: wxQt (on Wayland)
- OS version: Fedora 39

I've seen similar behavior on previous versions of Fedora as well as on Haiku (just to mention something completely different, which does not rely on Wayland)

### Note:

Commenting out `hasMouseTracking` in `wxAnyButton::QtGetCurrentState()` gives me much better results:

```diff
diff --git a/src/qt/anybutton.cpp b/src/qt/anybutton.cpp
index c1d9e77ce7..fa62660dc4 100644
--- a/src/qt/anybutton.cpp
+++ b/src/qt/anybutton.cpp
@@ -158,7 +158,7 @@ wxAnyButton::State wxAnyButton::QtGetCurrentState() const
         return State_Pressed;
     }
 
-    if ( HasCapture() || m_qtPushButton->hasMouseTracking() || m_qtPushButton->underMouse() )
+    if ( HasCapture() /* || m_qtPushButton->hasMouseTracking() */ || m_qtPushButton->underMouse() )
     {
         return State_Current;
     }
```

![image](https://github.com/wxWidgets/wxWidgets/assets/92124087/3e4bf979-5dd3-4aac-9828-750073b48275)

![image](https://github.com/wxWidgets/wxWidgets/assets/92124087/3e402a4a-a6be-476b-8403-6dd086d22c3b)

vadz:
[...]

PR #24132, opened 2023-12-12
Fix wxAnyButton Current/Normal state under wxQt

davidkaroly:
Do not check for hasMouseTracking in QtGetCurrentState as mouse tracking is always enabled in any widget derived from wxQtEventSignalHandler.

This resulted in an QtGetCurrentState reporting State_Current when it should have reported State_Normal i.e. the button showed hovering state when it was actually not hovering.

After the change, button state will show State_Current only if the mouse pointer is actually hovered on the button.

See #24129.


vadz:

PR #24134, opened 2023-12-13
more convenient API for wxGenericValidator + wxLB_SINGLE

wsu-cb:
For a single-selection wxListBox, using an int to transfer data to/from wxGenericValidator is more convenient than wxArrayInt.

### DIFF
--- a/interface/wx/valgen.h
+++ b/interface/wx/valgen.h
@@ -23,6 +23,7 @@
 
     @since 3.2.5
     A wxLB_SINGLE wxListBox can also use an int.  wxColourPickerCtrl support.
+    A 3-state wxCheckBox can use wxCheckBoxState.
 
     For more information, please see @ref overview_validator.
 
--- a/interface/wx/valgen.h
+++ b/interface/wx/valgen.h
@@ -15,7 +15,7 @@
     - wxButton, wxRadioButton, wxToggleButton, wxBitmapToggleButton, wxSpinButton
     - wxCheckBox, wxRadioBox, wxComboBox, wxListBox, wxCheckListBox
     - wxGauge, wxSlider, wxScrollBar, wxChoice, wxStaticText
-    - wxSpinCtrl, wxTextCtrl, wxColourPickerCtrl
+    - wxSpinCtrl, wxTextCtrl, wxColourPickerCtrl (since wxWidgets 3.3.0 or later).
 
     It checks the type of the window and uses an appropriate type for it.
     For example, wxButton and wxTextCtrl transfer data to and from a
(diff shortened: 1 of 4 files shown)


vadz (commented):
Thanks, this looks good to me but AFAICS you could use `pControl->HasMultipleSelection()` which would be more readable than the existing style checks, would you mind changing the code like this please? TIA!

wsu-cb:
> Thanks, this looks good to me but AFAICS you could use `pControl->HasMultipleSelection()` which would be more readable than the existing style checks, would you mind changing the code like this please? TIA!

I agree HasMultipleSelection() is much more readable, but it is not listed in interface\wx\listbox.h.  May I infer that you are implying that it should be part of the public API?

vadz:
Yes, sure. But, FWIW, even if it's not part of the public API it can still be used by wx itself internally.

wsu-cb:
> But, FWIW, even if it's not part of the public API it can still be used by wx itself internally.

Now that you point this out, I realize that I should perform a mental context switch when changing from wx user to wx contributor.

wsu-cb:
As I have worked on my own project some more, it seems like `wxGenericValidator` could be further extended to support a `wxColourPickerCtrl`/`wxColour` pair.  Would it be OK for me to add support for that to this PR?

vadz:
Yes, you could add it here, or make a separate PR, as you prefer (but let me know if you plan to add it here so that I don't merge it for now). TIA!

wsu-cb:
> Yes, you could add it here, or make a separate PR, as you prefer (but let me know if you plan to add it here so that I don't merge it for now). TIA!

I will add it here.

interface/wx/valgen.h
```diff
@@ -15,12 +15,16 @@
     - wxButton, wxRadioButton, wxToggleButton, wxBitmapToggleButton, wxSpinButton
     - wxCheckBox, wxRadioBox, wxComboBox, wxListBox, wxCheckListBox
     - wxGauge, wxSlider, wxScrollBar, wxChoice, wxStaticText
-    - wxSpinCtrl, wxTextCtrl
+    - wxSpinCtrl, wxTextCtrl, wxColourPickerCtrl
```
vadz, line 18:
```suggestion
    - wxSpinCtrl, wxTextCtrl, wxColourPickerCtrl (since wxWidgets 3.3.0 or later).
```

vadz (commented):
Thanks, this looks good to me, but newer changes can't be backported to 3.2, as they modify the ABI. I can backport `wxListBox`-related commits there however if you'd like to.

wsu-cb:
> Thanks, this looks good to me, but newer changes can't be backported to 3.2, as they modify the ABI. I can backport `wxListBox`-related commits there however if you'd like to.

I am, and will be for some time to come, using master rather than released code, so backporting does not affect me.  However, if anyone else asks for backporting, I think this can be made ABI-compatible by using an extra indirection in the wxGenericValidator object so that it doesn't change size.  I don't think the additional constructors actually break the ABI by themselves.

PR #24138, opened 2023-12-15
wxFileSystemHandler implementation for URLs with the "data" scheme

GlooFox:


vadz:
Thanks, but there are unfortunately a few things missing here:

1. The most obvious one is that the new code is not being compiled at all as the new file hasn't been added to `build/files` nor anywhere else. The details of what needs to be done are [explained here](https://github.com/wxWidgets/wxWidgets/blob/master/docs/contributing/how-to-add-files-to-build-system.md), but if it's too difficult for you to do, I could do it myself.
2. There is no documentation for the new class: this is something we really, really need to have before merging it. Please create `interface/wx/fs_data.h` similar to the existing files there (e.g. `fs_mem.h`) and provide at least minimum instructions for using this class there, ideally with an example.
3. In addition to having an example, it would be really great to have a unit test checking that this class works as intended in `tests/filesys/filesystest.cpp`, otherwise, even if it works now, it risks getting broken in the future without anybody noticing.

TIA!

GlooFox:
Thank you for the advice.
I believe I've finally added the required documentation and tests.
But unfortunately I'm not familiar with this build infrastructure and can't even decide to which category this functionality should belong: BASE_CMN_SRC, BASE_AND_GUI_CMN_SRC, something else... So would be great if someone assist with this stuff a little bit.

vadz:
I've added the files to the build system and made a couple of other tweaks, but the test fails, please fix this, thanks!

vadz:
Please test your code locally before pushing to avoid wasting CI resources for nothing. TIA!

GlooFox:
Sorry for wasting the resources. I promise to try avoiding that.
Unfortunately looks like not all code compiles the same on your environment.
Anyway the solution hopefully works now.

vadz:
Thanks! Looks good to me now, I've pushed some minor fixes and will (squash) merge it once the CI builds finish.

PR #24139, opened 2023-12-15
wxGrid: remove trailing newline if only one cell selected

Blake-Madden:


vadz (commented):
Thanks, this looks fine, even though looking at this code again I can't help wondering why do we add the extra new line only to remove it later instead of doing what the inner loop does and only adding it if necessary.

I.e. I'd rather write it like this

```cpp
bool firstRow = true;
for ( ... row ... )
{
    if ( firstRow )
        firstRow = false;
    else
        buf += wxTextFile::GetEOL();

    ... same code as now, maybe renaming "first" to "firstCol", but without += at the end ...
}
```

Wouldn't this be more clear/consistent?

Blake-Madden:
> Thanks, this looks fine, even though looking at this code again I can't help wondering why do we add the extra new line only to remove it later instead of doing what the inner loop does and only adding it if necessary.
> 
> I.e. I'd rather write it like this
> 
> ```c++
> bool firstRow = true;
> for ( ... row ... )
> {
>     if ( firstRow )
>         firstRow = false;
>     else
>         buf += wxTextFile::GetEOL();
> 
>     ... same code as now, maybe renaming "first" to "firstCol", but without += at the end ...
> }
> ```
> 
> Wouldn't this be more clear/consistent?

Got it. I re-implemented as requested.

vadz:
Looks good, thanks for the update! I will merge it soon.

PR #24141, opened 2023-12-17
Fix use after free in MSW printing

Blake-Madden:
I have ASAN (`/fsanitize=address`) enabled on Windows 11 (using MSVC2022) and was running into a use after deallocation error when showing the page setup dialog.  It appears that a `GlobalPtrLock` is locking an `HGLOBAL` and then unlocking it upon RAII (in its DTOR, as expected). However, in one branch, the `HGLOBAL` is being `GlobalFree`d and then the `GlobalPtrLock` is trying to unlock the already freed `HGLOBAL`.

In this patch, I added an `Unlock` function that does what the DTOR does, but also nulls the pointers. I call this before the `GlobalFree` and when RAII kicks in, it won't try to unlock the freed `HGLOBAL` anymore.

### DIFF
--- a/include/wx/msw/private.h
+++ b/include/wx/msw/private.h
@@ -803,7 +803,7 @@ public:
         Unlock();
     }
 
-    // explicity unlock in case we need to unlock before the DTOR is called
+    // explicitly unlock in case we need to unlock before the DTOR is called
     void Unlock()
     {
         if (m_hGlobal && !GlobalUnlock(m_hGlobal))
--- a/include/wx/msw/private.h
+++ b/include/wx/msw/private.h
@@ -811,6 +811,21 @@ public:
         }
     }
 
+    // explicity unlock in case we need to unlock before the DTOR is called
+    void Unlock()
+    {
+        if (m_hGlobal && !GlobalUnlock(m_hGlobal))
+        {
+            DWORD dwLastError = ::GetLastError();
+            if (dwLastError != NO_ERROR)
+            {
+                wxLogApiError(wxT("GlobalUnlock"), dwLastError);
+            }
+        }
+        m_hGlobal = nullptr;
+        m_ptr = nullptr;
+    }
+
     void *Get() const { return m_ptr; }
     operator void *() const { return m_ptr; }
 
--- a/include/wx/msw/private.h
+++ b/include/wx/msw/private.h
@@ -808,6 +808,7 @@ public:
     {
         if (m_hGlobal && !GlobalUnlock(m_hGlobal))
         {
+            // this might happen simply because the block became unlocked
             DWORD dwLastError = ::GetLastError();
             if (dwLastError != NO_ERROR)
             {
--- a/include/wx/msw/private.h
+++ b/include/wx/msw/private.h
@@ -788,7 +788,7 @@ public:
         m_ptr = GlobalLock(hGlobal);
         if ( !m_ptr )
         {
-            wxLogLastError(wxT("GlobalLock"));
+            wxLogLastError("GlobalLock");
         }
     }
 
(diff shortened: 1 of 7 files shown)


vadz:
Thanks for finding and fixing this bug, it's definitely a bad one!

However I'd prefer to add a new `GlobalPtr::Reset()` function so that we could do

```cpp
lockDevMode.Reset(newDevMode.Release());
```

instead of calling `GlobalFree()` manually. Could you please update the code to do it like this to make it even safer? TIA!

P.S. In master we could add move assignment operator to `GlobalPtr`, but I'd rather add `Reset()`, at least as the first step, to allow backporting this to 3.2 too.

Blake-Madden:
Got it.  Replaced `Unlock `with `Reset `and call `::GlobalFree` in there on the original `HGLOBAL`. Please let me know if I misunderstood anything.

vadz:
I'm really sorry, you understood what I wrote perfectly well, but while applying this patch I've realized that I was wrong because I mixed up `GlobalPtr` and `GlobalPtrLock` classes and that it really doesn't make much sense to reset a lock.

So, finally, what about this, even simpler, solution:

```diff
diff --git a/src/msw/printdlg.cpp b/src/msw/printdlg.cpp
index 4a1f8a6994..21d7681529 100644
--- a/src/msw/printdlg.cpp
+++ b/src/msw/printdlg.cpp
@@ -520,6 +520,7 @@ void wxWindowsPrintNativeData::InitializeDevMode(const wxString& printerName, Wi
     }
 
     // Try to initialize devmode to user or system default.
+    GlobalPtr newDevMode;
     if (m_devMode)
     {
         GlobalPtrLock lockDevMode(m_devMode);
@@ -548,18 +549,21 @@ void wxWindowsPrintNativeData::InitializeDevMode(const wxString& printerName, Wi
                 if ( pDevMode )
                 {
                     DWORD devModeSize = pDevMode->dmSize + pDevMode->dmDriverExtra;
-                    GlobalPtr newDevMode(devModeSize, GMEM_FIXED | GMEM_ZEROINIT);
+                    newDevMode.Init(devModeSize, GMEM_FIXED | GMEM_ZEROINIT);
                     if ( newDevMode )
                     {
                         memcpy(newDevMode, pDevMode, devModeSize);
-
-                        ::GlobalFree(m_devMode);
-                        m_devMode = newDevMode.Release();
                     }
                 }
             }
         }
     }
+
+    if ( newDevMode )
+    {
+        ::GlobalFree(static_cast<HGLOBAL>(m_devMode));
+        m_devMode = newDevMode.Release();
+    }
 }
 
 bool wxWindowsPrintNativeData::TransferFrom( const wxPrintData &data )
```

? It looks like it ought to fix the problem and this is an even more minimal change, which is nice for backporting to 3.2.

As for master, I think we really ought to make `m_devMode` `GlobalPtr` instead of `void*` and get rid of all the explicit alloc/free calls -- but this would require moving it into some kind of `wxWindowsPrintNativeDataImpl` first as `GlobalPtr` can't be used in the public header, i.e. it's a bigger change that will have to wait for another day...

vadz:
Sorry, not sure what happened here, did you close this one intentionally?

I.e. do you agree with fixing the problem with the diff above?

Sorry again for the confusion/misleading advice.

Blake-Madden:
Had to resubmit this as #24146, but the requested change is in there.

Blake-Madden:
> Sorry, not sure what happened here, did you close this one intentionally?
> 
> I.e. do you agree with fixing the problem with the diff above?
> 
> Sorry again for the confusion/misleading advice.

Sorry, I had too much trouble rebasing and undoing all the other changes, so I just made a new branch and PR and used your suggestion.  Anyway, this change works well for me, no ASAN issues.

PR #24144, opened 2023-12-18
More proposed backports for 3.2.5

vadz:


Kumazuma:
Could you add the missing default ctor of wxFileDilalog in wxMSW?

vadz:
> Could you add the missing default ctor of wxFileDilalog in wxMSW?

Done. And while looking at this I've realized that #24113 was actually wrong and resulted in crash if you just did

```cpp
{
   wxFileDialog fd;
}
```

so I've fixed it too.

vadz:
Oops, and somehow the tests crash too now. I can't debug this now, so I'll probably revert this after all.

src/msw/filedlg.cpp
```diff
@@ -1028,6 +1028,11 @@ wxFileDialog::wxFileDialog(wxWindow *parent,
 
 {
     // NB: all style checks are done by wxFileDialogBase::Create
+}
+
```
Kumazuma, line 1032:
```suggestion
    m_data = NULL;
}

```
It is not initialized as NULL.

vadz: Oh, thanks for noticing this, cherry pick removed this accidentally as it doesn't exist in master. Fixed.

PR #24149, opened 2023-12-19
Fix W String conversion

computergeek1507:
when enabling wxUSE_STD_STRING_CONV_IN_WXSTRING, w_char * operator is removed, force conversion with t_str().


vadz:

PR #24152, opened 2023-12-21
Make some WidgetsPages scrollable in the widgets sample

AliKet:
Under some ports, `wxQt` specifically, the left and middle panes of certain `WidgetsPages` must be scrollable. Otherwise, some properties will not be accessible at all.

These changes are still useful for the other ports too. especially when the window is not maximized.

_I found this very handy when working on this PR #24135._

### Here is some screenshots:

Under KDE **without** this patch:
widgets_qt_old

Under KDE **with** this patch:
widgets_qt

Under Win10 with this patch:
widgets

### DIFF
--- a/samples/widgets/bmpcombobox.cpp
+++ b/samples/widgets/bmpcombobox.cpp
@@ -39,6 +39,7 @@
 #include "wx/stattext.h"
 #include "wx/dc.h"
 #include "wx/dcmemory.h"
+#include "wx/scrolwin.h"
 #include "wx/sizer.h"
 #include "wx/icon.h"
 #include "wx/dir.h"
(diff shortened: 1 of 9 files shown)


PBfordev:
I agree that sometimes there is too much content and some of it may not be shown. 

I have actually wanted to do something about it, but my intention was to make `WidgetsPage` derive from `wxScrolledWindow`. In other words, make the whole top right part of the frame scrollable. What do you think about this approach?

AliKet:
> I agree that sometimes there is too much content and some of it may not be shown.
> 
> I have actually wanted to do something about it, but my intention was to make `WidgetsPage` derive from `wxScrolledWindow`. In other words, make the whole top right part of the frame scrollable. What do you think about this approach?

I actually did it this way the first time, but ran into an annoying problem while trying this:
some control being tested became out of sight when scrolling the page. This may not be a big problem after all. so I'll update this PR like so. thanks.

AliKet:
Replaced by #24154

PR #24154, opened 2023-12-21
Make WidgetsPage pages scrollable in the widgets sample

AliKet:
Because some pages in the sample contain too much content to fit entirely in the visible area of the page. Under `wxQt` port, this is true regardless of the state of the window (whether it is maximized or not) some content remains hidden and inaccessible.

`ButtonWidgetsPage` and `ComboboxWidgetsPage` are also fixed after the changes made here:

widgets_sample_button

widgets_sample_combobox

### DIFF
--- a/samples/widgets/button.cpp
+++ b/samples/widgets/button.cpp
@@ -374,7 +374,7 @@ void ButtonWidgetsPage::CreateContent()
 #endif
 
     // right pane
-    m_sizerButton = new wxBoxSizer(wxHORIZONTAL);
+    m_sizerButton = new wxBoxSizer(wxVERTICAL);
     m_sizerButton->SetMinSize(FromDIP(150), 0);
 
     // the 3 panes panes compose the window
@@ -612,7 +612,7 @@ void ButtonWidgetsPage::CreateButton()
 
     m_sizerButton->AddStretchSpacer();
     m_sizerButton->Add(m_button, wxSizerFlags().Centre().Border());
-    m_sizerButton->AddStretchSpacer();
+    m_sizerButton->AddStretchSpacer(2);
 
     m_sizerButton->Layout();
 }
--- a/samples/widgets/combobox.cpp
+++ b/samples/widgets/combobox.cpp
@@ -259,6 +259,15 @@ void ComboboxWidgetsPage::CreateContent()
     wxSizer *sizerTop = new wxBoxSizer(wxHORIZONTAL);
 
     // upper left pane
+    wxStaticBoxSizer *sizerLeftTop = new wxStaticBoxSizer(wxVERTICAL, this, "&Popup");
+    wxStaticBox* const sizerLeftTopBox = sizerLeftTop->GetStaticBox();
+
+    sizerLeftTop->Add(new wxButton(sizerLeftTopBox, ComboPage_Popup, "&Show"),
+                         wxSizerFlags().Border().Centre());
+    sizerLeftTop->Add(new wxButton(sizerLeftTopBox, ComboPage_Dismiss, "&Hide"),
+                         wxSizerFlags().Border().Centre());
+
+    // lower left pane
 
     // should be in sync with ComboKind_XXX values
     static const wxString kinds[] =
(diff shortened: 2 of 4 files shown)


vadz:
Nice, thanks!

paulcor:
Now the `widgets` sample doesn't size itself at all. This is a pretty unimpressive first impression.

paulcor:
Screenshot_2023-12-22_06-42-26

vadz:
I (and I guess @AliKet) didn't notice it because I had saved geometry in the config file, but after erasing it's indeed bad.

We probably just need to impose some reasonable min size to the scrolled window as I don't know how can we compute it accurately now, at least in the vertical direction (it ought to be possible to compute the best width, but clearly this doesn't work neither, I wonder why...).

AliKet:
Sorry for breaking this!

The idea of making `WidgetsPage` scrollable is because:

- It's the only way to make the widget example usable under `wxQt`.
- For the other ports (`wxMSW` & `wxGTK`) I had planned to make the log window height customizable just in case someone (like me) preferred larger log windows, in which case a scrollable `WidgetsPage` is unavoidable.

So IMHO we can just impose a reasonable minimum size (`wxSize(800, 400)` works for me) on `WidgetsPage` and everything should continue to work as before:
```
void WidgetsFrame::OnPageChanged(WidgetsBookCtrlEvent& event)
{
...
        curPage->CreateContent();
        curPage->SetMinSize(wxSize(800, 400));
        curPage->SetScrollRate(10, 10);
        curPage->FitInside();
...
}
```

What do you think ?

AliKet:
Or better yet, this also works well too:
```
void WidgetsFrame::OnPageChanged(WidgetsBookCtrlEvent& event)
{
...
        curPage->CreateContent();
        curPage->SetMinSize(curPage->GetBestSize());
        curPage->SetScrollRate(10, 10);
        curPage->FitInside();
...
}
```

paulcor:
Why is the log window so tall with wxQT?

vadz:
Unless somebody objects to this very much I'm going to push this:

```diff
commit 27585bf2c3f0366153c1aa3e7102959f466039b3
Author: Vadim Zeitlin <vadim@wxwidgets.org>
Date:   2023-12-22 17:40:00 +0100

    Hardcode minimal widget sample notebook size
    
    Not doing this resulted in the window being tiny by default, so it's
    better than nothing, even if it's clearly still not ideal.
    
    See #24154.

diff --git a/samples/widgets/widgets.cpp b/samples/widgets/widgets.cpp
index c2cb886adc..263e8ddac5 100644
--- a/samples/widgets/widgets.cpp
+++ b/samples/widgets/widgets.cpp
@@ -506,8 +506,9 @@ WidgetsFrame::WidgetsFrame(const wxString& title)
     // Uncomment to suppress page theme (draw in solid colour)
     //style |= wxNB_NOPAGETHEME;
 
+    // Give it some reasonably big minimal size by default.
     m_book = new WidgetsBookCtrl(m_panel, Widgets_BookCtrl,
-                                 wxDefaultPosition, wxDefaultSize,
+                                 wxDefaultPosition, FromDIP(wxSize(900, 500)),
                                  style, "Widgets");
 
     InitBook();
```

Better solution can be always implemented later but this should make things usable for now.

AliKet:
> Why is the log window so tall with wxQT?

It is just the default size of the `wxListBox/QListWidget` (the log window type) under this platform!

AliKet:
> Unless somebody objects to this very much I'm going to push this:
> 
> ```diff
> commit 27585bf2c3f0366153c1aa3e7102959f466039b3
> Author: Vadim Zeitlin <vadim@wxwidgets.org>
> Date:   2023-12-22 17:40:00 +0100
> 
>     Hardcode minimal widget sample notebook size
>     
>     Not doing this resulted in the window being tiny by default, so it's
>     better than nothing, even if it's clearly still not ideal.
>     
>     See #24154.
> 
> diff --git a/samples/widgets/widgets.cpp b/samples/widgets/widgets.cpp
> index c2cb886adc..263e8ddac5 100644
> --- a/samples/widgets/widgets.cpp
> +++ b/samples/widgets/widgets.cpp
> @@ -506,8 +506,9 @@ WidgetsFrame::WidgetsFrame(const wxString& title)
>      // Uncomment to suppress page theme (draw in solid colour)
>      //style |= wxNB_NOPAGETHEME;
>  
> +    // Give it some reasonably big minimal size by default.
>      m_book = new WidgetsBookCtrl(m_panel, Widgets_BookCtrl,
> -                                 wxDefaultPosition, wxDefaultSize,
> +                                 wxDefaultPosition, FromDIP(wxSize(900, 500)),
>                                   style, "Widgets");
>  
>      InitBook();
> ```
> 
> Better solution can be always implemented later but this should make things usable for now.

Yes I think this should be ok!

PBfordev:
I believe that:
1. Making the widgets page scrollable was the only solution, as the controls could be basically inaccessible depending on the main frame size, even without any hint they are there. This also did not make a good first impression.
2. The log window should be separated from the widgets tree and page by a splitter, to make its height adjustable.

BTW, are the pixels in `wxScrolled::SetScroll{Rate|bars}()` physical or logical?

vadz:
> I believe that:
> 
>     1. Making the widgets page scrollable was the only solution, as the controls could be basically inaccessible depending on the main frame size, even without any hint they are there. 

I agree with this but, just for the record, with GTK/Gnome there is no hint scrollbars are there neither, so it's still not great.

>     2. The log window should be separated from the widgets tree and page by a splitter, to make its height adjustable.

Yes, absolutely. 
 
> BTW, are the pixels in `wxScrolled::SetScroll{Rate|bars}()` physical or logical?

They're logical (at least I don't see any reason to think otherwise), so we should use `FromDIP()` when using hard-coded values for them (this is done in docview, popup and scroll sample too in addition to this one, as well as in `wxHtmlWindow`).

vadz:
It turns out there is another problem with this: pages supposed to show ellipsization/wrapping don't ever show them any more because the width is never small enough. I'm not sure what to do about it, probably just make these particular pages not scrollable, but this risks reintroducing the same problem that this PR tried to fix...

PR #24155, opened 2023-12-22
Fix missing semicolon

mcorino:
Missing semicolon on latest method addition causes doxygen parsing issues which in turn causes problems for dependent projects like wxRuby and wxPython.


vadz:

PR #24156, opened 2023-12-22
Fix confusing argument name.

mcorino:
Fix an obvious copy/paste error.


vadz:
And this one too, thanks.

PR #24158, opened 2023-12-23
Fix confusing argument names in wxPersistentWindow docs

PBfordev:
Use the same argument name as in the actual code in the class constructor as well as the convenience function, instead of wrongly copypasted "book".

See also 33de6dc (Fix confusing wxPersistentTLW ctor argument name, 2023-12-22).


vadz:

Issue #24082, opened 2023-11-23
Fix `Grid::LabelClick` unit test or disable it under AppVeyor


vadz:
This unit test keeps failing in 50% of CI runs (or maybe even more often), so it needs to be either fixed or disabled.

As the first step, it would be useful to debug it to understand in which section of this test it fails.

vadz:
Another regularly failing test case is ListCtrlTestCase/ColumnDrag.

vadz:
Thanks again for looking at this (unfortunately I won't be able to do it today...)!

I think waiting for it to become active would be fine if it does become it, at least in (more than?) 50% of cases, i.e. the tests still run sometimes. If it never does, then something must be wrong and we'd need to find what it is, ideally, but adding this wait wouldn't be any different from disabling the tests under AppVeyor.

vadz:
I've tried debugging this but I still don't understand what's going on here, I just can't reproduce the problem locally at all. Or, rather, I can, but in a very specific scenario: it happens if I accidentally move the mouse while the test is running. I have no idea why would the mouse move on AppVeyor build machines, but after trying to account for this in #24162 the tests don't fail any more -- but apparently _not_ due to these fixes, see my comment in the PR.

vadz:
I give up, it still happens after my changes, even though apparently much less frequently -- but it still failed in [this run](https://ci.appveyor.com/project/wxWidgets/wxwidgets/builds/48873587/job/2qpgby18fuhbfmkj). So I will just ignore these failures in AppVeyor tests.

PR #24162, opened 2023-12-24
Attempt to fix failing UI tests under AppVeyor

vadz:


vadz:
The AppVeyor tests pass, even after several rebuilds, but I don't see any warning messages that should have been given if they pass thanks to the fixes here, which means that they probably don't really fix anything. But I'm still going to merge this, just in case they do, as this shouldn't do any harm in any case.

See #24082.

PR #24165, opened 2023-12-25
Disable swap interval in GLX wxGLCanvas implementation too

### LINKED ISSUE #24163: Set swap interval to 0 in `wxGLCanvasX11::SwapBuffers()` too
Apparently KiCAD/wxwdigets/glew in Debian is currently build without EGL support (I'm writing a bug report in parallel to change that). Running KiCAD with `vblank_mode=0` environment fixes the issue for me and in the mesa bug report it was suggested to copy the workaround implemented for EGL to GLX.

_Originally posted by @sre in https://github.com/wxWidgets/wxWidgets/issues/23512#issuecomment-1838930605_

vadz:
We need to do it when using XWayland for the same reasons as we had to do it in the EGL version when using either XWayland or Wayland directly: without this, we can block for up to 1 second in glXSwapBuffers() if the window is hidden, see #23512.

Closes #24163.

### DIFF
--- a/include/wx/unix/glx11.h
+++ b/include/wx/unix/glx11.h
@@ -99,6 +99,8 @@ public:
 private:
     GLXFBConfig *m_fbc;
     void* m_vi;
+
+    bool m_swapIntervalSet = false;
 };
 
 // ----------------------------------------------------------------------------
--- a/src/unix/glx11.cpp
+++ b/src/unix/glx11.cpp
@@ -124,6 +124,13 @@ typedef GLXContext(*PFNGLXCREATECONTEXTATTRIBSARBPROC)
 #define GLX_CONTEXT_ES2_PROFILE_BIT_EXT    0x00000004
 #endif
 
+namespace
+{
+
+constexpr const char* TRACE_GLX = "glx";
+
+} // anonymous namespace
+
 // ----------------------------------------------------------------------------
 // wxGLContextAttrs: OpenGL rendering context attributes
 // ----------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


dsa-t:
This might need checks for extension presence in `glXQueryExtensionsString` (see https://gitlab.com/kicad/code/kicad/-/blob/1fdf7d0cf42bace5375ad2898dbad864a928dc7c/include/gl_utils.h#L49)

And should only do anything on Xwayland (check for `WAYLAND_DISPLAY` env variable?)

vadz:
> This might need checks for extension presence in `glXQueryExtensionsString` (see https://gitlab.com/kicad/code/kicad/-/blob/1fdf7d0cf42bace5375ad2898dbad864a928dc7c/include/gl_utils.h#L49)

Wouldn't `glXGetProcAddress()` just return null if it isn't? I can't really test what happens in this case, but I don't see what else could possibly happen...

> And should only do anything on Xwayland (check for `WAYLAND_DISPLAY` env variable?)

This is another thing I was unsure about, i.e. does doing this do any harm for Xorg?

Checking for XWayland reliably is more involved than just checking the env var, AFAICS we need to get the RandR output name and check if it starts with `"XWAYLAND"` string.

dsa-t:
> Wouldn't `glXGetProcAddress()` just return null if it isn't? I can't really test what happens in this case, but I don't see what else could possibly happen...

If the function is supported in the client library, it doesn't mean that the extension works in X11 connection. See:
https://gitlab.com/kicad/code/kicad/-/commit/0e2d3236b50ea52dddf2d911f87ac48c9f0f9f5f
https://gitlab.com/kicad/code/kicad/-/issues/11751

vadz:
> > Wouldn't `glXGetProcAddress()` just return null if it isn't? I can't really test what happens in this case, but I don't see what else could possibly happen...
> 
> If the function is supported in the client library, it doesn't mean that the extension works in X11 connection.

Good point, thanks. But if it doesn't work, what can we do anyhow? I.e. if the function is there, would any harm come from calling it nevertheless? It looks like it just won't do anything in the worst case.

dsa-t:
> I.e. if the function is there, would any harm come from calling it nevertheless? It looks like it just won't do anything in the worst case.

Hopefully it won't do anything.
But when trying to query the drawable like this, it will crash:
```cpp
glXQueryDrawable( dpy, drawable, GLX_SWAP_INTERVAL_EXT, &clampedInterval );
```

vadz:
> > I.e. if the function is there, would any harm come from calling it nevertheless? It looks like it just won't do anything in the worst case.
> 
> Hopefully it won't do anything.

Yes, this looks like the only reasonable thing to do. But, in any case, I'd really prefer if someone could actually test it, i.e. I don't think we should merge it without testing.

> But when trying to query the drawable like this, it will crash:
> 
> ```c++
> glXQueryDrawable( dpy, drawable, GLX_SWAP_INTERVAL_EXT, &clampedInterval );
> ```

Good to know, thanks. But we don't need to do this (for now).

sre:
I just applied the patch from this pull request to the Debian package (locally) and it fixes the problem with Debian/Sway/XWayland/KiCAD.

I also did a quick test using Openbox (native Xorg) and things looked ok to me.

dsa-t:
@sre have you checked GPU/CPU power usage?

This still needs some way of detecting Xwayland

dsa-t:
This should cause tearing on Xorg.

vadz:
> This should cause tearing on Xorg.

Could you please check if it does? I only have Xorg inside a VM now and I'm not sure I'd see tearing there even if it were present.

dsa-t:
I won't have a machine available until later this month unfortunately.

PR #24170, opened 2023-12-27
wxRichToolTip fixes for wxQt port

AliKet:
Screenshot_2023-12-27_17-51-44

### DIFF
--- a/interface/wx/nonownedwnd.h
+++ b/interface/wx/nonownedwnd.h
@@ -62,8 +62,8 @@ public:
         resized using SetSize().
 
         As the overload above, this method is not guaranteed to work on all
-        platforms but currently does work in wxMSW, wxOSX/Cocoa, wxGTK and wxQt
-        (with the appropriate but almost always present X11 extensions) ports.
+        platforms but currently does work in wxMSW, wxOSX/Cocoa, wxGTK (with
+        the appropriate but almost always present X11 extensions) and wxQt ports.
 
         @since 2.9.3
      */
--- a/interface/wx/nonownedwnd.h
+++ b/interface/wx/nonownedwnd.h
@@ -62,8 +62,8 @@ public:
         resized using SetSize().
 
         As the overload above, this method is not guaranteed to work on all
-        platforms but currently does work in wxMSW, wxOSX/Cocoa and wxGTK (with
-        the appropriate but almost always present X11 extensions) ports.
+        platforms but currently does work in wxMSW, wxOSX/Cocoa, wxGTK and wxQt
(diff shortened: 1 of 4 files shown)


interface/wx/nonownedwnd.h
```diff
@@ -62,8 +62,8 @@ class wxNonOwnedWindow : public wxWindow
         resized using SetSize().
 
         As the overload above, this method is not guaranteed to work on all
-        platforms but currently does work in wxMSW, wxOSX/Cocoa and wxGTK (with
-        the appropriate but almost always present X11 extensions) ports.
+        platforms but currently does work in wxMSW, wxOSX/Cocoa, wxGTK and wxQt
+        (with the appropriate but almost always present X11 extensions) ports.
```
vadz, line 66: The text in parentheses is for wxGTK, so it should be
```suggestion
        platforms but currently does work in wxMSW, wxOSX/Cocoa, wxGTK (with
        the appropriate but almost always present X11 extensions) and wxQt ports.
```

src/common/popupcmn.cpp
```diff
@@ -431,8 +431,8 @@ bool wxPopupTransientWindow::Show( bool show )
 
 void wxPopupTransientWindow::Dismiss()
 {
-    Hide();
     PopHandlers();
+    Hide();
```
vadz, line 435: I'm not really comfortable with changing this, it risks breaking the other ports.

It looks like the current code dates from 414f2513fc (For wxMSW capture and release the mouse as the cursor moves out or back in to the transient popup window to enable the transient to be automatically dismissed when a click happens outside of it., 2005-04-15) so at the very least this would need to be tested in the popup sample under MSW to check if it doesn't result in any regressions.

vadz (commented):
Thanks, no real comments about wxQt changes, but I'm afraid this could break something for wxMSW (and maybe others).

If wxQt really requires doing things in different order, we might have to accommodate it using a preprocessor check...

src/common/popupcmn.cpp
AliKet: I just tested `wxMSW` and `wxGTK` with this change and didn't notice any regression. everything work well.

Looking closer at 414f251, I suspect that changing the order of calls in `wxPopupTransientWindow::Dismiss()` was **not** intentional because the role of `wxPopupWindowHandler` and `wxPopupFocusHandler` is to intercept events that cause the popup to disappear. IOW, they have already done their duty if we are really in `Dismiss()`.

vadz:
OK, let's merge this and see if this breaks anything under MSW.

Thanks!

PR #24172, opened 2023-12-29
Fix #24108

### LINKED ISSUE #24108: wxHeaderCtrl  and wxSimpleHeaderCtrl broken, multiple assertions in widgets sample
#### Bug description:

The current implementation of wxHeaderCtrl and wxSimpleHeaderCtrl do not conform to documentation and when used as advertised can cause assertions and failing functionality. This currently happens in the standard widgets sample.

The documentation for wxHeaderCtrl describes two protected virtual methods that  (according to docs) must be overridden in derived classes if called (which will be the case if column reordering and/or hiding is allowed): 

- `virtual void UpdateColumnVisibility(unsigned int idx, bool show)`
- `virtual void UpdateColumnsOrder(const wxArrayInt &order)`

The first problem here is that the documentation states that an override of `UpdateColumnVisibility` does not need to call `UpdateColum()` as that would be done by default by the library code. This is incorrect. The current implementation of wxHeaderCtrl does no such thing. This causes failures to refresh the displayed header when hiding/showing  columns through the popup menu/dialog.

The next problem lies with the default implementation of these methods in wxHeaderCtrl and the missing overrides of those methods in wxSimpleHeaderCtrl.
The default implementations in wxHeaderCtrl trigger assertions to warn the developer that overrides are required. Since wxSimpleHeaderCtrl lacks these overrides using this class underived (as is advertised in the docs and which the standard widgets sample does) will lead to assertions when attempting to reorder or hide columns through the popup menu/dialog.

Reordering through header column dragging does not suffer from this issue as the `UpdateColumnsOrder` is not called in this case. This is another inconsistency which would force a developer to implement either multiple methods to handle associated data changes (above mentioned callbacks and event handling) or forego the callbacks and use event handling only.
The latter would however currently require providing empty overrides for the callbacks as otherwise the assertions will still be triggered for certain user actions.

I am not sure what the best way to fix this would be.
It is clear at least wxSimpleHeaderCtrl needs updates to be usable without derivatives which is it's advertised use.
The changes to this are relatively straight forward as far as fixing suppressing the assertions and the missing trigger for column refreshing (UpdateColumn()) is concerned.
This will however leave the inconsistency between reordering through popup and dragging.
On the one hand the option of handling associated data control changes through the callbacks is an easy way for simple use cases but it always will require using derivatives.
On the other hand the option of handling the same through event handling is more generic and should probably be preferred. In this case It would be best to deprecate the callbacks (possibly marked for internal use only as wxGridHeaderCtrl actually uses them). The assertions should than be removed from the base (so developers of derived classes can ignore them but still keeping backward compatibility) and the implementation of wxSimpleHeaderCtrl updated to reflect this.

#### To Reproduce:

The assertions can be reproduced from the standard widget sample:

1. start sample
2. activate HeaderCtrl page
3. check 'Allow hide' alongside 'Allow reorder'
4. popup the context menu on the header control
5. select a column to hide

Alternatively:

1. start sample
2. activate HeaderCtrl page
3. check 'Allow hide' alongside 'Allow reorder'
4. popup the context menu on the header control
5. choose 'Customize ...'
6. reorder the columns in the dialog
7. select 'Ok'

Checking 'Allow hide' is necessary because otherwise the popup menu and dialog cannot be triggered. Without 'Allow hide' the only way to reorder is dragging which does not trigger an assertion.

- OS  and its version: OpenSuSE 15.5
  
  + GTK version:  3.24
  + Desktop environment : KDE

mcorino:
Updates implementing changes for issue #24108

### DIFF
--- a/include/wx/headerctrl.h
+++ b/include/wx/headerctrl.h
@@ -343,6 +343,9 @@ protected:
     virtual const wxHeaderColumn& GetColumn(unsigned int idx) const override;
     virtual bool UpdateColumnWidthToFit(unsigned int idx, int widthTitle) override;
 
+    virtual void UpdateColumnVisibility(unsigned int idx, bool show) override;
+    virtual void UpdateColumnsOrder(const wxArrayInt& order) override;
+
     // and define another one to be overridden in the derived classes: it
     // should return the best width for the given column contents or -1 if not
     // implemented, we use it to implement UpdateColumnWidthToFit()
--- a/interface/wx/headerctrl.h
+++ b/interface/wx/headerctrl.h
@@ -447,13 +447,15 @@ protected:
 
     /**
         Method called when the columns order is changed in the customization
-        dialog.
+        dialog @em or when the EVT_HEADER_END_REORDER event is not handled after
+        dragging a single column.
 
-        This method is only called from ShowCustomizeDialog() when the user
-        changes the order of columns. In particular it is @em not called if a
-        single column changes place because the user dragged it to the new
-        location, the EVT_HEADER_END_REORDER event handler should be used to
-        react to this.
+        This method is always called from ShowCustomizeDialog() when the user
+        changes the order of columns. In case a single column changes place
+        because the user dragged it to a new location, the EVT_HEADER_END_REORDER
+        event handler can be used to react to this. If this event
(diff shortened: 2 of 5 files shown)


vadz:
Thanks, (squash) merging with just some minor indentation/comment fixes.

PR #24180, opened 2024-01-02
Improve `wxFileConfig` XDG compliance

vadz:
The main goal here is to allow easily creating config files in `~/.config/appname` instead of using ~/.appname` directory for them as `wxCONFIG_USE_SUBDIR` does now. Unfortunately achieving this is not that simple without breaking backwards compatibility -- which is something we really can't do here as this would break all the existing applications that wouldn't find their config files any longer.

So add a new `wxCONFIG_USE_XDG` flag which can be used to explicitly opt-in the new behaviour.

Also add `Dir_Config` which can be used to retrieve this directory location for code not using `wxFileConfig` too.

Also includes some other related cleanup.

### DIFF
--- a/interface/wx/stdpaths.h
+++ b/interface/wx/stdpaths.h
@@ -80,6 +80,18 @@ public:
         Dir_Cache,
 
         /**
+            Directory containing configuration information.
+
+            Example return values:
+            - Unix: `~/.config`
+            - Windows: `C:\Users\username\AppData\Roaming`
+            - Mac: @c `~/Library/Preferences`
+
+            @since 3.3.0
+        */
+        Dir_Config,
+
+        /**
             Directory containing user documents.
 
             Example return values:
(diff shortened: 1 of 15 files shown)


interface/wx/stdpaths.h
```diff
@@ -79,6 +79,18 @@ class wxStandardPaths
         */
         Dir_Cache,
 
+        /**
+            Directory containing configuration information.
+
+            Example return values:
+            - Unix: `~/.config`
+            - Windows: `C:\Users\username\AppData\Roaming`
```
dconnet, line 87: Should there be a Local version of this (for Windows) like there is for userdatadir? (to allow windows users to keep this in appdata\local)

dconnet: Update: Though looking at wxFileConfig closer, I notice it does not support Local, so this may not be too important.

vadz: `GetUserDir(Dir_Config)` returns the same directory as `GetUserConfigDir()` under Windows as I think this makes most sense.

vadz:
I've added a convenient (IMO) function for migrating existing config files to the new location, please let me know if anybody has any comments about this.

I've also fixed a crash in wxGTK, hopefully without breaking anything else. I plan to cherry pick the commit doing this ([055c4cb](https://github.com/wxWidgets/wxWidgets/pull/24180/commits/055c4cbed5006f801d6e01a88eea46d89a67dc39)) to 3.2 too.

vadz:
I'd like to merge this soon, as I have other changes depending on this one, please let me know if there are any comments.

Issue #23804, opened 2023-08-24
Wrong return types on the documentation of wxRealPoint and wxPoint


jank9201:
### Description

The  methods of class wxRealPoint and wxPoint declared in " interface/wx/gdicmn.h"
class wxRealPoint
```
    wxSize operator /(const wxRealPoint& sz, int factor);
    wxSize operator *(const wxRealPoint& sz, int factor);
    wxSize operator *(int factor, const wxSize& sz);
    wxSize& operator /=(int factor);
    wxSize& operator *=(int factor);
```

and class wxPoint

```
    wxSize operator /(const wxPoint& sz, int factor);
    wxSize operator *(const wxPoint& sz, int factor);
    wxSize operator *(int factor, const wxSize& sz);
    wxSize& operator /=(int factor);
    wxSize& operator *=(int factor);
```
These methods are not correct according to the class wxRealPoint  and wxPoint declared in "include/wx/gdicmn.h"

In  "include/wx/gdicmn.h"
the above-menthoned methods in class wxRealPoint are defined as follows :
```
inline wxRealPoint operator/(const wxRealPoint& s, int i)
{
    return wxRealPoint(s.x / i, s.y / i);
}

inline wxRealPoint operator*(const wxRealPoint& s, int i)
{
    return wxRealPoint(s.x * i, s.y * i);
}

inline wxRealPoint operator*(int i, const wxRealPoint& s)
{
    return wxRealPoint(s.x * i, s.y * i);
}

```
the above-menthoned methods in class wxPoint are as follows :

```
inline wxPoint operator/(const wxPoint& s, int i)
{
    return wxPoint(s.x / i, s.y / i);
}

inline wxPoint operator*(const wxPoint& s, int i)
{
    return wxPoint(s.x * i, s.y * i);
}

inline wxPoint operator*(int i, const wxPoint& s)
{
    return wxPoint(s.x * i, s.y * i);
}
```
Compare above methods from different header files,  we can get the conclusion that the return type and some parameter type of these method  listed in " interface/wx/gdicmn.h" are incorrect.

The first three of the class wxRealPoint and wxPoint should be as follows:

class wxRealPoint
 ```
   wxRealPoint operator /(const wxRealPoint& s, int factor);
    wxRealPoint operator *(const wxRealPoint& s, int factor);
    wxRealPoint operator *(int factor, const wxRealPoint & s);
```
and class wxPoint
  ```
  wxPoint operator /(const wxPoint& s, int factor);
[...]

PR #24186, opened 2024-01-03
interface/wx/gdicmn.h: correct wxRealPoint and wxPoint types for documentation

carandraug:
As is described in #23804, the documented types for the overloaded operators of `wxRealPoint` and `wxPoint` are wrong, probably because they were copied from `wxSize`

_Originally posted by @vadz in https://github.com/wxWidgets/wxWidgets/issues/23804#issuecomment-1693282079_:
> I see, thanks, the return types are indeed wrong (due to copy-and-paste probably). If you can please make a PR fixing `interface/wx/gdicmn.h` it would be welcome, TIA!

This is *that* pull request.

---

Since I'm looking at this, I noticed that the `operator/` and `operator*`, are overloaded for `int`, `unsigned int, `long`, and `unsigned long` but only the `int` is documented. I'm guessing that this is on purpose to avoid extra verbose documentation but if not, those can be added.

### DIFF
--- a/interface/wx/gdicmn.h
+++ b/interface/wx/gdicmn.h
@@ -227,11 +227,11 @@ public:
     wxRealPoint& operator +=(const wxSize& sz);
     wxRealPoint& operator -=(const wxSize& sz);
 
-    wxSize operator /(const wxRealPoint& sz, int factor);
-    wxSize operator *(const wxRealPoint& sz, int factor);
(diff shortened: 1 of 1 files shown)


vadz:

vadz:
I forgot to mention this in the commit message, but it was cherry-picked as a8b14edc4072d82eae2a477c2168d0b982b9526b, thanks again.

PR #24188, opened 2024-01-04
Improve C++ standard handling in CMake

vadz:
I had to move things around to allow me to use `wx_check_cxx_source_compiles()` in `init.cmake`, please let me know if there is a better way to do this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,9 +97,6 @@ endif()
 
 project(wxWidgets VERSION ${wxVERSION} LANGUAGES ${wxLANGUAGES})
 
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-
 include(build/cmake/main.cmake)
 
 # Set the default startup project for Visual Studio
--- a/build/cmake/init.cmake
+++ b/build/cmake/init.cmake
@@ -10,6 +10,25 @@
 
 if(DEFINED wxBUILD_CXX_STANDARD AND NOT wxBUILD_CXX_STANDARD STREQUAL COMPILER_DEFAULT)
     set(CMAKE_CXX_STANDARD ${wxBUILD_CXX_STANDARD})
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+else()
+    # If the standard is not set explicitly, check whether we can use C++11
+    # without any special options.
+    include(CheckCXXSourceCompiles)
+    check_cxx_source_compiles("
+        #include <vector>
+        int main() {
+            std::vector<int> v{1,2,3};
+            for (auto& n : v)
+                --n;
+            return v[0];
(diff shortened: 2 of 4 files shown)


vadz:
@MaartenBent I'll merge this soon if you don't see anything wrong with it.

MaartenBent:
I'll have a detailed look later. But maybe you can move the c++11 check to setup.h so all checks are contained in that file?

vadz:
I wanted to do this in `init.cmake` because we have

```cmake
        if(wxUSE_WEBVIEW_CHROMIUM AND CMAKE_CXX_STANDARD LESS 14)
```

check there in wxCEF branch. And I don't think it's a good idea to move this check to `setup.cmake`, it belongs either here or in `options.cmake`.

So the other possibility I see is to move both checks to `options.cmake`, but I'm not really sure if it's better...

MaartenBent:
There are some other checks in `setup.cmake` that disable `wxUSE_` options, like `wxUSE_STACKWALKER`, `wxUSE_SNGLINST_CHECKER`, `wxUSE_JOYSTICK` and more.
So I don't have a problem with moving the c++14 (or 17) check of `wxUSE_WEBVIEW_CHROMIUM` there.

vadz:
It seems strange to split this check from all the other webview options checks currently in `init.cmake`, so should we move them _all_ there? This is going to result in a lot of changes and I'm not really sure if it's worth it... OTOH I don't really understand how are the checks split between `init.cmake` and `setup.cmake` currently (`options.cmake` contains just the options definitions AFAICT), and I guess moving all `wxUSE_XXX`-related stuff into `setup.cmake` would make it more consistent? But, again, this would be a huge change...

Also, we still need `wx_check_cxx_source_compiles()` in `init.cmake`, unless we move `CMAKE_CXX_STANDARD` to `options.cmake` too, but this really doesn't make much sense to me.

MaartenBent:
Keeping it together would indeed be the best. But I'm not a fan of moving the `wx_check_` functions around.

Maybe you don't even need to use these functions, but you can use `check_cxx_source_compiles` directly?
The same is already done in
https://github.com/wxWidgets/wxWidgets/blob/e97d27165a6367ad340fbfbadea67d110b94226e/build/cmake/init.cmake#L320-L323

vadz:
> Keeping it together would indeed be the best. But I'm not a fan of moving the `wx_check_` functions around.

I thought it would make sense to have all `wx_` functions in `functions.make`, but I don't really have any strong opinion here, so I undid this.
 
> Maybe you don't even need to use these functions, but you can use `check_cxx_source_compiles` directly?

Yes, sure, pushed now. Again, I thought using wx function was preferable (because otherwise why would we define it?), but I'm fine with doing it like this too.

vadz:
I'm going to push this if you're ok with it, we can always improve/refactor this later.

PR #24190, opened 2024-01-05
Experiment with not defining operators in global scope

vadz:
Try using "hidden friend" idiom instead.

If this works, we should extend to all the other operators currently defined in global scope.

To get an idea of the scope of the problem:

```console
$ echo "#include <wx/gdicmn.h>\nstruct S{}; bool foo() { S s1, s2; return s1 == s2; }" | g++ -fsyntax-only -Wall `wx-config --cxxflags --libs base,core` -x c++ -
<stdin>: In function ‘bool foo()’:
<stdin>:2:46: error: no match for ‘operator==’ (operand types are ‘S’ and ‘S’)
wx/include/wx/defs.h:791:17: note: candidate: ‘bool operator==(char, const wxUniChar&)’
  791 |     inline bool operator op(T1 x, T2 y) { return y oprev x; }
      |                 ^~~~~~~~
... 420 more lines of similar output snipped ...
```

I.e. there is a clear error message (on 2 lines) and then **424** lines of junk due to global operators defined in wx headers.

### DIFF
--- a/include/wx/bitmap.h
+++ b/include/wx/bitmap.h
@@ -18,6 +18,7 @@
 #include "wx/gdicmn.h"  // for wxBitmapType
 #include "wx/colour.h"
 #include "wx/image.h"
+#include "wx/variant.h"
 
 class WXDLLIMPEXP_FWD_CORE wxBitmap;
 class WXDLLIMPEXP_FWD_CORE wxBitmapHandler;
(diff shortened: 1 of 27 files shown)


vadz:
Somewhat surprisingly, this seems to work even with MSVS 2015, so I'll continue with moving other operators in global scope inside the classes they're used with.

vadz:
As this does seem to work with all the compilers we support, I'd like to merge it. This is an incompatible change, as these operators won't be applied any longer to the types convertible to wx types, but IMO this is a good thing and it definitely enormously improves the situation with the error messages (for MSVC you need to use `/std:c++20` or `/permissive-` to see it, however).

Any objections?

Issue #24177, opened 2023-12-31
Use wxOverlay to implement non-live resizing in wxAUI


AliKet:
### Description
This is a proposal to add non-live resizing to wxAUI using `wxOverlay `+ `wxClientDC` which can be proven  to work under these platforms:
- `wxGTK3` with **Wayland**
- `wxQt`
- `wxMAC` (someone else needs to test the patch under this platform)
- `wxMSW` (see below)

The patch I'm using:
```
diff --git a/include/wx/aui/framemanager.h b/include/wx/aui/framemanager.h
index da9dff9609..72b3c84f11 100644
--- a/include/wx/aui/framemanager.h
+++ b/include/wx/aui/framemanager.h
@@ -24,6 +24,7 @@
 #include "wx/timer.h"
 #include "wx/sizer.h"
 #include "wx/bmpbndl.h"
+#include "wx/overlay.h"
 
 enum wxAuiManagerDock
 {
@@ -616,6 +617,8 @@ protected:
     wxByte m_hintFadeAmt;       // transparent fade amount
     wxByte m_hintFadeMax;       // maximum value of hint fade
 
+    wxOverlay m_overlay;
+
     void* m_reserved;
 
 private:
diff --git a/src/aui/framemanager.cpp b/src/aui/framemanager.cpp
index 8a660e0722..e28149e452 100644
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -280,15 +280,10 @@ static void DrawResizeHint(wxDC& dc, const wxRect& rect)
     wxBitmap stipple = wxPaneCreateStippleBitmap();
     wxBrush brush(stipple);
     dc.SetBrush(brush);
-#ifdef __WXMSW__
-    wxMSWDCImpl *impl = (wxMSWDCImpl*) dc.GetImpl();
-    PatBlt(GetHdcOf(*impl), rect.GetX(), rect.GetY(), rect.GetWidth(), rect.GetHeight(), PATINVERT);
-#else
     dc.SetPen(*wxTRANSPARENT_PEN);
 
     dc.SetLogicalFunction(wxXOR);
     dc.DrawRectangle(rect);
-#endif
 }
 
 
@@ -762,7 +757,7 @@ unsigned int wxAuiManager::GetFlags() const
 
 // With Core Graphics on Mac or GTK 3, it's not possible to show sash feedback,
 // so we'll always use live update instead.
[...]

PR #24193, opened 2024-01-06
Use wxOverlay to show sash feedback in wxAUI and in wxSplitterWindow

AliKet:
See #24177

Tested under: `wxMSW`, `wxGTK3` (**Wayland** & **X11**) and `wxQt`

Not Tested under `wxMac` though!

### DIFF
--- a/interface/wx/overlay.h
+++ b/interface/wx/overlay.h
@@ -27,6 +27,25 @@ public:
        example, when the window content has been changed and repainted.
     */
     void Reset();
+
+    /**
+        Sets or unsets constant opacity of the overlay window.
+
+        If @a alpha is `-1`, use per-pixel alpha blending, otherwise use the
+        given alpha value for all pixels.
+
+        @note Drawing on the overlay window is opaque by default under wxMSW.
+        You have to call SetOpacity(-1) before initializing the overlay and use
+        a graphics context to do the actual drawing if you want per-pixel alpha
+        blending. Or, assuming that SetOpacity(-1) wasn't called, explicitly set
+        the alpha value (at any time) to change the overlay window's opacity.
+
+        @note This is currently only implemented for wxMSW and does nothing in
+        the other ports.
+
+        @since 3.3.0
+    */
+    void SetOpacity(int alpha);
 };
 
 
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -271,31 +271,23 @@ wxIMPLEMENT_DYNAMIC_CLASS(wxPseudoTransparentFrame, wxFrame);
 
 static wxBitmap wxPaneCreateStippleBitmap()
 {
-    // TODO: Provide x1.5 and x2.0 versions.
-    unsigned char data[] = { 0,0,0,192,192,192, 192,192,192,0,0,0 };
+    // Notice that wxOverlay, under wxMSW, uses the wxBLACK colour i.e.(0,0,0)
+    // as the key colour for transparency. and using it for the stipple bitmap
+    // will make the sash feedback totaly invisible if the window's background
+    // colour is (192,192,192) or so. (1,1,1) is used instead.
+    unsigned char data[] = { 1,1,1,192,192,192, 192,192,192,1,1,1 };
     wxImage img(2,2,data,true);
     return wxBitmap(img);
 }
 
 static void DrawResizeHint(wxDC& dc, const wxRect& rect)
 {
-#ifdef __WXMSW__
     wxBitmap stipple = wxPaneCreateStippleBitmap();
     wxBrush brush(stipple);
     dc.SetBrush(brush);
-    wxMSWDCImpl *impl = (wxMSWDCImpl*) dc.GetImpl();
-    PatBlt(GetHdcOf(*impl), rect.GetX(), rect.GetY(), rect.GetWidth(), rect.GetHeight(), PATINVERT);
-#else
-    // Note that we have to use white colour for wxINVERT to work with
-    // wxGraphicsContext-based wxDC implementations, such as used by wxGTK3
-    // (and wxOSX, but this code is never used for the latter because it always
-    // uses live resize).
-    dc.SetPen(*wxWHITE_PEN);
-    dc.SetBrush(*wxWHITE_BRUSH);
+    dc.SetPen(*wxTRANSPARENT_PEN);
 
-    dc.SetLogicalFunction(wxINVERT);
     dc.DrawRectangle(rect);
-#endif
 }
 
 
@@ -767,11 +759,9 @@ unsigned int wxAuiManager::GetFlags() const
     return m_flags;
 }
 
-/* static */ bool wxAuiManager::AlwaysUsesLiveResize(const wxWindow* window)
+/* static */ bool wxAuiManager::AlwaysUsesLiveResize(const wxWindow* WXUNUSED(window))
 {
-    // Not using live resize relies on wxClientDC being usable for drawing, so
-    // we have to use live resize if it can't be used on the current platform.
-    return !wxClientDC::CanBeUsedForDrawing(window);
+    return false;
 }
 
 bool wxAuiManager::HasLiveResize() const
@@ -4453,8 +4443,7 @@ void wxAuiManager::OnLeftUp(wxMouseEvent& event)
         if (!HasLiveResize())
         {
             // get rid of the hint rectangle
-            wxClientDC dc{m_frame};
-            DrawResizeHint(dc, m_actionHintRect);
+            m_overlay.Reset();
         }
         if (m_currentDragItem != -1 && HasLiveResize())
             m_actionPart = & (m_uiParts.Item(m_currentDragItem));
@@ -4569,18 +4558,20 @@ void wxAuiManager::OnMotion(wxMouseEvent& event)
             else
             {
                 wxRect rect(pos, m_actionPart->rect.GetSize());
-                wxClientDC dc{m_frame};
 
                 if (!m_actionHintRect.IsEmpty())
                 {
-                    // remove old resize hint
-                    DrawResizeHint(dc, m_actionHintRect);
                     m_actionHintRect = wxRect();
                 }
 
-                // draw new resize hint
-                DrawResizeHint(dc, rect);
+                wxClientDC dc{m_frame};
+                wxDCOverlay overlaydc(m_overlay, &dc);
+                overlaydc.Clear();
+
+                // draw resize hint
                 m_actionHintRect = rect;
+                rect.SetPosition(rect.GetPosition() + m_frame->GetClientAreaOrigin());
+                DrawResizeHint(dc, rect);
             }
         }
     }
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -33,7 +33,7 @@
     #include "wx/settings.h"
     #include "wx/app.h"
     #include "wx/dcclient.h"
-    #include "wx/dcscreen.h"
+    #include "wx/dcmemory.h"
     #include "wx/toolbar.h"
     #include "wx/image.h"
     #include "wx/statusbr.h"
@@ -64,6 +64,8 @@ wxDEFINE_EVENT( wxEVT_AUI_FIND_MANAGER, wxAuiManagerEvent );
     #include "wx/msw/dc.h"
 #endif
 
+#include "wx/dcgraph.h"
+
 #include <memory>
 
 wxIMPLEMENT_DYNAMIC_CLASS(wxAuiManagerEvent, wxEvent);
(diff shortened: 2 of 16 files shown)


AliKet:
May conflict with #24166

AliKet:
I just tested this under `wxGTK2` and didn't work unfortunately.

AliKet:
@vadz @paulcor  If I apply this patch:
```
diff --git a/src/gtk/dcclient.cpp b/src/gtk/dcclient.cpp
index 69478afc92..c2ed3f7afa 100644
--- a/src/gtk/dcclient.cpp
+++ b/src/gtk/dcclient.cpp
@@ -452,6 +452,11 @@ void wxWindowDCImpl::SetUpDC( bool isMemDC )
     gdk_gc_set_clip_rectangle( m_brushGC, nullptr );
     gdk_gc_set_clip_rectangle( m_textGC, nullptr );
     gdk_gc_set_clip_rectangle( m_bgGC, nullptr );
+
+    gdk_gc_set_subwindow( m_penGC, GDK_INCLUDE_INFERIORS );
+    gdk_gc_set_subwindow( m_brushGC, GDK_INCLUDE_INFERIORS );
+    gdk_gc_set_subwindow( m_textGC, GDK_INCLUDE_INFERIORS );
+    gdk_gc_set_subwindow( m_bgGC, GDK_INCLUDE_INFERIORS );
 }
 
 void wxWindowDCImpl::DoGetSize( int* width, int* height ) const
```
then this PR works correctly under wxGTK2.

vadz:
Sorry, I've just realized that I perhaps shouldn't have merged #24166 before this one. But I can resolve the conflicts, however I'd like to understand the situation with wxGTK2: can we use `wxOverlay` for it or not, finally? I'd like to minimize time spent on it, but AFAIK we don't plan to drop support for it yet, so we still need this to work in at least some way there...

AliKet:
> Sorry, I've just realized that I perhaps shouldn't have merged #24166 before this one. But I can resolve the conflicts, however I'd like to understand the situation with wxGTK2: can we use `wxOverlay` for it or not, finally? I'd like to minimize time spent on it, but AFAIK we don't plan to drop support for it yet, so we still need this to work in at least some way there...

`wxAUI` and `wxSplitterWindow` now work correctly under all supported platforms (can not test under **wxMac** !!)

AliKet:
I also managed to get transparent hint to work with `wxOverlay` (generic and native), so `wxPseudoTransparentFrame` (with all its complications and limitation) is no longer needed.
I just want to know if it's ok to push the changes (depend on the changes done here) here or should I wait until this is approved and merged?

as always I need someone else to test my changes under **wxOSX**. TIA!

src/msw/overlay.cpp
```diff
@@ -100,8 +99,16 @@ void wxOverlayImpl::Init(wxDC* dc, int , int , int , int )
...
+    if ( wxDynamicCast(dc, wxClientDC) )
+    {
+        m_rect = m_window->GetClientRect();
+        // The rectangle should be in screen coordinates
+        m_window->ClientToScreen(&m_rect.x, &m_rect.y);
+    }
+    else // wxWindowDC
+    {
+        m_rect = m_window->GetScreenRect();
+    }
```
vadz, line 111: Why can't we always just use `GetScreenRect()` here? Is it just because we really need client size for `wxClientDC` but full window size otherwise? If so, we could still use something like

```c++
m_rect = wxRect(m_window->GetScreenPosition(), dc->GetSize());
```

I think, without any dynamic casts.

src/common/overlaycmn.cpp
```diff
@@ -140,6 +140,20 @@ void wxDCOverlay::Clear()
 
 #include "wx/window.h"
 
+#if defined(__WXGTK__) && !defined(__WXGTK3__)
```
vadz, line 143: I would just put this in `Init()` itself, i.e. avoid defining the extra function which is used for wxGTK2 only but just inlined its body there inside `#if GTK2` test.

The rationale is that it's not obvious when seeing the call to `wxDontClipSubWindows()` that it only does something for GTK2, and it would be helpful to see it immediately.

This would also be a bit shorter.

src/generic/splitter.cpp
```diff
@@ -71,17 +71,37 @@ wxEND_EVENT_TABLE()
...
     // following the mouse movement while it drags the sash, without it we only
     // draw the sash at the new position but only resize the windows when the
-    // dragging is finished -- but drawing the sash is done using wxClientDC,
-    // so check if we can use it and always use live resizing if we can't
-    return wnd->AlwaysUsesLiveUpdate() || wnd->HasFlag(wxSP_LIVE_UPDATE);
+    // dragging is finished
+    return wnd->HasFlag(wxSP_LIVE_UPDATE);
+}
+
+static wxBitmap wxPaneCreateStippleBitmap()
```
vadz, line 86: We should extract this function somewhere and probably cache the bitmap because even if it's just a 2x2 one, I am still not sure we want to recreate it on every repaint.

The trouble is that we need to clean it up in some module and I don't see anything obviously appropriate. Perhaps we could make this a stock brush? Then it would be cleaned up by `wxStockGDI::DeleteAll()`.

But at least let's factor out this function into some private header, it's not quite trivial and it might need to be changed later (for dark mode or high DPI or I don't know what else) and so I'd rather not have 2 copies of it.

vadz (commented):
Thanks a lot, this is much better and I'd really like to merge it but I'd like to avoid duplicating the function for creating the stipple bitmap. Will you be able to change it or should I do it myself?

vadz:
> I also managed to get transparent hint to work with `wxOverlay` (generic and native), so `wxPseudoTransparentFrame` (with all its complications and limitation) is no longer needed.

Excellent!

> I just want to know if it's ok to push the changes (depend on the changes done here) here or should I wait until this is approved and merged?

Whatever you prefer. I.e. I'm fine with adding them here.

> as always I need someone else to test my changes under **wxOSX**. TIA!

I'll do it before merging, so from this point of view it would be better to remove `wxPseudoTransparentFrame` now, as I could test this too then.

Thanks again!

src/msw/overlay.cpp
AliKet: When `wxClientDC` is used we want the overlay window to be positioned over the client area of the window. `m_window->GetScreenPosition()` returns the upper left corner of the window and not the top left position of the client area.

The overlay window positioned at `m_window->GetScreenPosition()` (**incorrect**): 

overlay_incorrect

The overlay window positioned at `m_window->GetClientRect()` (**correct**):

overlay_correct

vadz: Ah, I see, thanks for the explanation!

I'd still like to avoid dynamic cast, for all the usual reasons. As always, we can just add a virtual function instead and while it's a bit ugly to have window-oriented function in `wxDC`, we do already have `GetWindow()` there, so we could also add some `GetOriginInWindow()` or `GetWindowOffset()` or whatever else you'd prefer that would return `wxPoint(0, 0)` by default but would be overridden to return `GetClientAreaOrigin()` in `wxClientDCImpl`.

Do you see any problems with doing it like this? Thanks again!

src/generic/splitter.cpp
AliKet: Updating the documentation and this refactoring are still not done yet. if you don't mind me asking you to do that when testing under wxOSX, or fixing any regressions introduced (which I haven't noticed any!). TIA!

vadz (commented):
Thanks again for all your work here, this looks really nice but I just have a couple of questions about opacity semantics under MSW, notably its default behaviour, could you please check this? TIA!

src/msw/overlay.cpp
```diff
@@ -100,12 +125,21 @@ void wxOverlayImpl::Init(wxDC* dc, int , int , int , int )
...
+        m_bitmap.CreateWithDIPSize(m_rect.GetSize(), m_window->GetDPIScaleFactor());
+    }
+    else
+    {
+        m_bitmap.CreateWithDIPSize(m_rect.GetSize(), m_window->GetDPIScaleFactor(), 32);
+        m_bitmap.UseAlpha();
+    }
 
-    m_overlayWindow = wxCreateOverlayWindow(m_rect);
+    HWND hWndInsertAfter = ::GetNextWindow(GetHwndOf(m_window), GW_HWNDPREV);
```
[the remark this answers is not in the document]
AliKet: This change is mainly to prevent the overlay window from being placed on top of/covering any floating window.

**Before:**

overlay_bad

**After:**

overlay_good

Unfortunately, this breaks rubberbanding in the drawing sample. That is, the rubberband no longer appears there! What actually solves both problems is doing the following:
```
    const auto win = wxGetTopLevelParent(m_window);
    HWND hWndInsertAfter = ::GetNextWindow(GetHwndOf(win), GW_HWNDPREV);
    m_overlayWindow = wxCreateOverlayWindow(m_rect, m_alpha, hWndInsertAfter);
```
Which I'll commit if you don't see any problem in doing so (of course) ?

include/wx/dc.h
```diff
@@ -713,6 +719,9 @@ class WXDLLIMPEXP_CORE wxDC : public wxObject
     wxWindow *GetWindow() const
         { return m_pimpl->GetWindow(); }
 
+    wxRect GetWindowRect() const
```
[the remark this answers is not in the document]
AliKet: Documented now!

src/aui/framemanager.cpp
```diff
@@ -744,40 +559,23 @@ wxAuiDockUIPart* wxAuiManager::HitTest(int x, int y)
...
-    }
 }
 
 unsigned int wxAuiManager::GetFlags() const
 {
     return m_flags;
 }
 
-/* static */ bool wxAuiManager::AlwaysUsesLiveResize(const wxWindow* window)
+/* static */ bool wxAuiManager::AlwaysUsesLiveResize(const wxWindow* WXUNUSED(window))
```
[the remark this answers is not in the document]
AliKet: Updated.

AliKet:
@vadz I see that [`wxSplitterWindow::AlwaysUsesLiveUpdate`](https://docs.wxwidgets.org/latest/classwx_splitter_window.html#a7a377a089bb00011d00f233b055a5f8f)  is only added in 3.3.0.
Should we remove it completely or update its documentation like we did with `wxAuiManager::AlwaysUsesLiveResize`?

vadz:
> @vadz I see that [`wxSplitterWindow::AlwaysUsesLiveUpdate`](https://docs.wxwidgets.org/latest/classwx_splitter_window.html#a7a377a089bb00011d00f233b055a5f8f) is only added in 3.3.0. Should we remove it completely or update its documentation like we did with `wxAuiManager::AlwaysUsesLiveResize`?

Yes, I think it should be fine to remove it if we never do this any longer. Thanks!

interface/wx/overlay.h
```diff
@@ -27,6 +27,25 @@ class wxOverlay
...
     */
     void Reset();
+
+    /**
+        Sets or unsets constant opacity of the overlay window.
+
+        If @a alpha is `-1`, use per-pixel alpha blending, otherwise use the
+        given alpha value for all pixels.
+
+        @note The overlay window is opaque by default under wxMSW. You have to
```
paulcor, line 37: That's not my understanding of how 'wxOverlay` is supposed to work. It's expected that calling `wxDCOverlay::Clear()` will make the overlay transparent.

AliKet: In fact, it works as you expect. Just the wording is misleading and should say something like this instead:
> Drawing on the overlay window is opaque by default under wxMSW...

because this is how `SetLayeredWindowAttributes` works. i.e. it can only apply a constant alpha on the drawing with no alpha blending at all.

As pictures speak louder than words, this is what I mean with rubberbanding as example:

- Rubberbanding (outline) with `SetLayeredWindowAttributes` + `wxALPHA_OPAQUE` (default):

overlay_opaque_outline

- Rubberbanding (filled) with `SetLayeredWindowAttributes` + `wxALPHA_OPAQUE` :

overlay_opaque

- Rubberbanding (filled) with `SetLayeredWindowAttributes` + alpha (128): 

overlay_translucent

- Rubberbanding (filled) with `UpdateLayeredWindow` + wxGCDC: 

overlay_translucent_gc

include/wx/dc.h
```diff
@@ -176,6 +176,12 @@ class WXDLLIMPEXP_CORE wxDCImpl: public wxObject
 
     void SetWindow(wxWindow* w) { m_window = w; }
 
+    // By default, this function returns the window rectangle in screen
+    // coordinates as returned by GetScreenRect(). It returns the client
+    // rectangle (also in screen coordinates) if the DC is a wxClientDC.
+    // This is mostly meant to be used by wxOverlay.
+    virtual wxRect GetWindowRect() const;
```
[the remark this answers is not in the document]
AliKet: I'll remove this commit then.

BTW, `wxOverlay` and `wxDCOverlay` documentations need to be updated. Also mentione that `wxClientDC` is the only supported DC that should be used to draw on the overlay!

Hope I understood you correctly!!

src/generic/splitter.cpp
AliKet: Would it be better to do the refactoring in a seperate PR instead?

vadz (approved):
Thanks a lot once again for all your work on this!

I've finally tested this under Mac and I found one problem there: the overlay contents continued to be displayed after `Reset()` (although it disappeared as soon as the window was redrawn). I couldn't find any better way to fix this than the brute force approach implemented in 4261e1d07d52c0703a5adf4a436937eb0f8896e2 but at least with this it works for me.

@csomor If you have a moment to look at this commit and let me know if there is some simpler way of removing the overlay contents from screen, please let me know. FWIW I've tried calling `setNeedsDisplay:YES` but this didn't seem to help.

vadz:
I'm merging this as it has been opened for a long time. We can almost surely improve things under Mac, but this can be done later.

Thanks again for all your efforts!

PR #24195, opened 2024-01-07
Use XDG-compliant location by default in wxFileConfig

vadz:
Still use the traditional dot file if it already exists, but prefer using the new location otherwise, i.e. for the new program installation.

Add wxCONFIG_USE_HOME to allow forcing the use of the old location if really necessary.

Also use the new style as default "old style" of MigrateLocalFile() so that calling it even when using XDG layout in wxStandardPaths still works as expected.


vadz:
FYI, I also wrote a [blog post](https://github.com/wxWidgets/website/pull/91) describing this change.

Issue #24121, opened 2023-12-07
MSW: wxFontPickerCtrl doesn't fully support hiDPI


willco007:
### Description

On a 4k monitor (scale at 150%, resolution 3840x2160), when using wxFontPickerCtrl to pick the font, the picker window is blurry and controls are sized incorrectly.

#### Bug description:

wxFontPickerCtrl font picker window isn't supporting hiDPI fully.

#### Expected vs observed behaviour:

Expect, crisp font drawing and checkbox controls drawn at the correct size. Observed, font is burry and checkboxes are too small.

#### Stack trace:

n/a

#### Patch or snippet allowing to reproduce the problem:

_fontPicker = new wxFontPickerCtrl(this, wxID_ANY, wxNullFont, wxDefaultPosition, wxDefaultSize, wxFNTP_USE_TEXTCTRL);
_fontPicker->GetPickerCtrl()->SetLabel(_("Select..."));

Press the select button to show the picker window.

#### To Reproduce:

show font picker window, observe results (see screenshot)

<img width="320" alt="image" src="https://github.com/wxWidgets/wxWidgets/assets/4294739/e4e3a4be-227e-4abf-a640-7f06a4a0180c">

### Platform and version information

- wxWidgets version you use: 3.2.1
- wxWidgets port you use: wxMSW
- OS and its version: Windows 11, 22H2

MaartenBent:
Hi, this is by design. Because these native win32 dialogs (font, colour, file picker) don't support moving between displays with different DPI. Moving it from 250% to 100% would look like this:
<img width="212" alt="Screenshot 2023-12-07 210734" src="https://github.com/wxWidgets/wxWidgets/assets/8088070/ded0566b-229a-47fa-91f0-380e6ed6225f">

The disadvantage is that it always scales from 100%, even if you only have 1 display. 

I suppose we could add a  `wxSystemOptions`, like `msw.dark-mode`, to disable this behavior. And/or only use it when `wxDisplay::GetCount() > 1`.

The font picker is shown with a system-dpi-aware context in:https://github.com/wxWidgets/wxWidgets/blob/96b5c1419bca0519d32688df7dc681ac5f52a2ae/src/msw/fontdlg.cpp#L95-L96

PR #24196, opened 2024-01-08
Improve native MSW dialog DPI awareness

MaartenBent:
Only show the native dialogs using system dpi-awereness, when there are displays with different DPI.

And add the `msw.native-dialogs-pmdpi` option to also show the dialogs using per-monitor dpi-awareness when there are displays with different DPI.

I tested many combinations of dpi and on which display the dialog is created, and it seems to always show and return the correct font size. But some more tests would be nice.

See #24121

### DIFF
--- a/interface/wx/sysopt.h
+++ b/interface/wx/sysopt.h
@@ -81,6 +81,12 @@
         using it, i.e. this has the same effect as calling
         wxApp::MSWEnableDarkMode(). If set to 2, use dark mode unconditionally,
         as if this function were called with wxApp::DarkMode_Always argument.
+    @flag{msw.native-dialogs-pmdpi}
+        Some native win32 dialogs (like the font and colour pickers) are not
+        per-monitor DPI aware, and wxWidgets will forcefully show them as
+        system DPI aware when there are monitors with different DPI connected.
+        If set to 1, these dialogs will always be shown as per-monitor DPI
+        aware (when enabled in the manifest).
     @endFlagTable
 
 
--- a/src/msw/fontdlg.cpp
+++ b/src/msw/fontdlg.cpp
@@ -126,21 +126,22 @@ int wxFontDialog::ShowModal()
         chooseFontStruct.lpfnHook = wxFontDialogHookProc;
     }
 
+    // The native font dialog does not support moving between displays with
+    // different DPIs. Check if it will be shown system-dpi-aware.
+    const bool useSystemDPI = wxMSWImpl::AutoSystemDpiAware::Needed();
+
+    // When the font dialog is system-dpi-aware, it expects the font at 96DPI/100% scaling.
+    // When the font dialog is per-monitor-dpi-aware, it expects a font with the system DPI.
+    const int fontdlgDPI = useSystemDPI ? wxDisplay::GetStdPPIValue() : wxGetDPIofHDC(ScreenHDC()).y;
+
     if ( m_fontData.m_initialFont.IsOk() )
     {
         flags |= CF_INITTOLOGFONTSTRUCT;
         logFont = m_fontData.m_initialFont.GetNativeFontInfo()->lf;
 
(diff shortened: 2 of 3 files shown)


src/msw/fontdlg.cpp
```diff
@@ -126,21 +126,22 @@ int wxFontDialog::ShowModal()
         chooseFontStruct.lpfnHook = wxFontDialogHookProc;
     }
 
+    // The native font chooser does not support moving between displays with
+    // different DPIs. Check if it will be shown system-dpi-aware.
+    const bool useSystemDPI = wxMSWImpl::AutoSystemDpiAware::Needed();
+
+    // When the font chooser is system-dpi-aware, it expects the font at 96DPI/100% scaling.
+    // When the font chooser is per-monitor-dpi-aware, it expects a font with the system DPI.
+    const int chooserDPI = useSystemDPI ? wxDisplay::GetStdPPIValue() : wxGetDPIofHDC(ScreenHDC()).y;
```
vadz, line 135: This is very minor, but I thought this was going to be used to "choose" something. I'd call this just `dialogDPI` or `fontdlgDPI`.

src/msw/window.cpp
```diff
@@ -4820,8 +4820,7 @@ namespace wxMSWImpl
 {
 
 AutoSystemDpiAware::SetThreadDpiAwarenessContext_t
-AutoSystemDpiAware::ms_pfnSetThreadDpiAwarenessContext =
-    (AutoSystemDpiAware::SetThreadDpiAwarenessContext_t)-1;
+    AutoSystemDpiAware::ms_pfnSetThreadDpiAwarenessContext = nullptr;
```
vadz, line 4823: Changing this means we're going to retry loading it every time instead of just trying it once in case it's not available. Not sure if this actually risks happening, but if it does, this doesn't look like a good change.

vadz (commented):
Thanks for improving this!

Changes look good to me but it's, of course, pretty easy to not notice something with these DPI conversions, and so more testing would indeed be welcome -- unfortunately I can't provide it myself.

vadz:
Do you mind if I make the changes mentioned above? It's absolutely not urgent, of course, so if you plan to make them yourself (or object to making them at all), just let me know, but I just think it would be nice to get this merged.

MaartenBent:
I agree with the suggestions. I'll push them now.

interface/wx/sysopt.h
```diff
@@ -81,6 +81,13 @@
         using it, i.e. this has the same effect as calling
         wxApp::MSWEnableDarkMode(). If set to 2, use dark mode unconditionally,
         as if this function were called with wxApp::DarkMode_Always argument.
+    @flag{msw.native-dialogs-pmdpi}
+        If set to 1, native win32 dialogs (like the font and colour pickers)
+        will always be shown as per-monitor DPI aware (when enabled in the manifest).
+        By default, these are only per-monitor DPI aware when all monitors have
```
PBfordev, line 87: Sorry, but what does "By default, these are only per-monitor DPI aware when all monitors have the same DPI, ..." mean?

Is it factually any different then "By default, these are not per-monitor DPI aware."?

MaartenBent: Indeed, if you don't support moving between monitors, they are not per-monitor DPI aware. I'll rephrase.

MaartenBent:
No idea what's going on with the failed test, clipboard is totally unrelated. I ran the test locally and there it also fails. With and without this PR.

vadz:
Thanks again, I'm finally merging this.

AFAICS this should also be backported to 3.2 (after replacing `nullptr` with `NULL`s there), please let me know if it shouldn't (but I won't do it immediately just in case people do find some problem with this change).

MaartenBent:
Yes, this can be backported to 3.2.

PR #24197, opened 2024-01-08
Fix testsuite failures on s390x

jengelh:
Upon ``make check``, we observe:

	./textfile/textfiletest.cpp:351

	./textfile/textfiletest.cpp:354: FAILED:
	  CHECK( f.Open("/sys/power/state") )
	with expansion:
	  false

	./textfile/textfiletest.cpp:355: FAILED:
	  REQUIRE( f.GetLineCount() == 1 )
	with expansion:
	  0 == 1

### DIFF
--- a/tests/file/filetest.cpp
+++ b/tests/file/filetest.cpp
@@ -153,6 +153,7 @@ TEST_CASE("wxFile::Special", "[file][linux][special-file]")
     CHECK( fileProc.ReadAll(&s) );
     CHECK( !s.empty() );
 
+    if ( wxFile::Exists("/sys/power/state") ) {
     // All files in /sys have the size of one kernel page, even if they don't
     // have that much data in them.
     const long pageSize = sysconf(_SC_PAGESIZE);
@@ -163,6 +164,7 @@ TEST_CASE("wxFile::Special", "[file][linux][special-file]")
     CHECK( fileSys.ReadAll(&s) );
     CHECK( !s.empty() );
     CHECK( s.length() < pageSize );
+    }
 }
 
 #endif // __LINUX__
--- a/tests/filename/filenametest.cpp
+++ b/tests/filename/filenametest.cpp
@@ -1040,9 +1040,11 @@ TEST_CASE("wxFileName::GetSizeSpecial", "[filename][linux][special-file]")
     INFO( "size of /proc/kcore=" << size );
     CHECK( size > 0 );
 
+    if ( wxFile::Exists("/sys/power/state") ) {
     // All files in /sys are one page in size, irrespectively of the size of
     // their actual contents.
     CHECK( wxFileName::GetSize("/sys/power/state") == sysconf(_SC_PAGESIZE) );
+    }
 }
 
 #endif // __LINUX__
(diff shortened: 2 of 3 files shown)


vadz:
Thanks, but I wonder why this file is not present there, I'm almost sure think these tests passed on s390x when we still used Travis (which provided access to them) and there also was 2005e50b27 (Fix tests when run on s390x arch, 2020-04-07) which wouldn't have made sense if this file didn't exist at all. Perhaps it's somehow possible to enable it there? Or is there some other file present under this arch as well that we could use?

BTW, I guess you're using some CI system with s390x support? It could be interesting to integrate it with our CI, if possible.

imciner2:
Which OS is being used on this s390x system? I believe that Fedora is building for s390x using this test suite and doesn't seem to have any problems with this check.

vadz:
> Which OS is being used on this s390x system?

It must be Linux (because this code is inside check for it), but I also think it's not Fedora because this test must indeed be passing under it ...

> I believe that Fedora is building for s390x using this test suite and doesn't seem to have any problems with this check.

... after fixing the problem they did have in the commit referenced above.

swt2c:
Yeah, so on Fedora we've been skipping that test for a bit now (sorry for not reporting it or looking into it):
https://src.fedoraproject.org/rpms/wxGTK/blob/rawhide/f/wxGTK.spec#_232

jengelh:
> BTW, I guess you're using some CI system with s390x support? It could be interesting to integrate it with our CI, if possible.

https://build.opensuse.org . technically yes, a github job could trigger an OSC commit that then triggers the OBS build. Obtaining results is practically an asynchronous affair though.

swt2c:
Also skipping this test on several arches (beyond s390x) on Debian:
https://salsa.debian.org/freewx-team/wxwidgets3.2/-/blob/master/debian/rules?ref_type=heads#L9

vadz:
OK, thanks, I guess we should indeed just apply this fix then.

OTOH I still think it would be nice if we could use some file which does exist everywhere, there must be something like this. E.g. what about `/sys/kernel/security/lsm`? It looks pretty arch-independent (speaking of its existence, not contents)?

jengelh:
Hm. Strictly speaking, anything that is a separate mount is optional. Only the root directory and the current working directory of the current process are guaranteed to exist.

If the task is to test the wxFile class for reading, consider creating a file using wxFile. That way, you have a file that is guaranteed to exist.

vadz:
This test was added because some (or all) of these classes didn't work correctly with the files like these ones, which have size of one page (4096) but from which you can't read 4096 bytes of data. It's not easy to create a file like this (I guess it would be possible using sparse files and what not), so using an existing file is much simpler.

jengelh:
Perhaps, then, just walk /sys recursively until you find an arbitrary file that has access(R_OK) and stat::st_size>0?

vadz:
> Perhaps, then, just walk /sys recursively until you find an arbitrary file that has access(R_OK) and stat::st_size>0?

This seems too complicated (or at least much more complicated than replacing one hard-coded file with another one), so let's just skip the tests -- but at least give a warning when doing it (see the upcoming commit).

matoro:
Just for reference, this affects any system with a kernel built with `CONFIG_PM=n`.  This is common in virtual machines since there's no need for power management

cboltz:
For the records: e2cc16ef9c45bdc64d42e4fef4dda46a3077cb35 was slightly broken. #24768 fixes it ;-)

PR #24199, opened 2024-01-09
Add `wxBitmap::CreateWithLogicalSize()` and use it

vadz:
While doing it I've actually realized that many of the existing uses of `CreateWithDIPSize()` were wrong -- or at least I think so, please correct me if I'm missing something, so this is more than just a simplification/making the API easier to use, see the last commit.


MaartenBent:
Sorry I haven't had time to test it earlier. But I see no problems so far in some of the samples on Windows.

PR #24200, opened 2024-01-10
wxQt wxGLCanvas improvements and crash fixes.

dsa-t:
This allows running "cube" and "pyramid" samples and KiCad with OpenGL canvas on wxQt.
Tested on Linux Mint 21.1 (X11, XFCE).

`"wxPaintDC can't be created outside wxEVT_PAINT handler"` assert check didn't work correctly with the wxQt OpenGL canvas, so it was removed.

### DIFF
--- a/include/wx/qt/glcanvas.h
+++ b/include/wx/qt/glcanvas.h
@@ -24,6 +24,9 @@ public:
 
     virtual bool SetCurrent(const wxGLCanvas& win) const override;
 
+private:
+    QGLContext* m_glContext = nullptr;
+
     wxDECLARE_CLASS(wxGLContext);
 };
 
@@ -78,7 +81,9 @@ public:
 
     virtual bool SwapBuffers() override;
 
-    static bool ConvertWXAttrsToQtGL(const int *wxattrs, QGLFormat &format);
+    virtual bool QtCanPaintWithoutActivePainter() const override;
+
+    static bool ConvertWXAttrsToQtGL(const wxGLAttributes &glattrs, const wxGLContextAttrs ctxAttrs, QGLFormat &format);
 
 private:
     wxDECLARE_CLASS(wxGLCanvas);
--- a/src/qt/combobox.cpp
+++ b/src/qt/combobox.cpp
@@ -288,40 +288,48 @@ void wxComboBox::SetSelection( long from, long to )
 
     SetInsertionPoint( from );
     // use the inner text entry widget (note that can be null if not editable)
-    if ( m_qtComboBox->lineEdit() != nullptr )
+    if ( QLineEdit* const lineEdit = m_qtComboBox->lineEdit() )
     {
-        m_qtComboBox->lineEdit()->setSelection( from, to - from );
+        lineEdit->setSelection( from, to - from );
     }
 }
 
 void wxComboBox::SetInsertionPoint( long pos )
 {
-    // check if pos indicates end of text:
-    if ( pos == -1 )
-        m_qtComboBox->lineEdit()->end( false );
-    else
-        m_qtComboBox->lineEdit()->setCursorPosition( pos );
+    if ( QLineEdit* const lineEdit = m_qtComboBox->lineEdit() )
+    {
+        // check if pos indicates end of text:
+        if ( pos == -1 )
+            lineEdit->end( false );
+        else
+            lineEdit->setCursorPosition( pos );
+    }
 }
 
 long wxComboBox::GetInsertionPoint() const
 {
-    long selectionStart = m_qtComboBox->lineEdit()->selectionStart();
+    QLineEdit* const lineEdit = m_qtComboBox->lineEdit();
+
+    if ( !lineEdit )
+        return -1;
+
+    long selectionStart = lineEdit->selectionStart();
 
     if ( selectionStart >= 0 )
         return selectionStart;
 
-    return m_qtComboBox->lineEdit()->cursorPosition();
+    return lineEdit->cursorPosition();
 }
 
 void wxComboBox::GetSelection(long* from, long* to) const
 {
     // use the inner text entry widget (note that can be null if not editable)
-    if ( m_qtComboBox->lineEdit() != nullptr )
+    if ( QLineEdit* const lineEdit = m_qtComboBox->lineEdit() )
     {
-        *from = m_qtComboBox->lineEdit()->selectionStart();
+        *from = lineEdit->selectionStart();
         if ( *from >= 0 )
         {
-            *to = *from + m_qtComboBox->lineEdit()->selectedText().length();
+            *to = *from + lineEdit->selectedText().length();
             return;
         }
     }
--- a/src/qt/dcclient.cpp
+++ b/src/qt/dcclient.cpp
@@ -126,7 +126,7 @@ wxPaintDCImpl::wxPaintDCImpl( wxDC *owner )
 wxPaintDCImpl::wxPaintDCImpl( wxDC *owner, wxWindow *win )
     : wxWindowDCImpl( owner, win )
 {
-    wxCHECK_RET( m_isWindowPainter,
+    wxCHECK_RET( m_isWindowPainter || win->QtCanPaintWithoutActivePainter(),
                  "wxPaintDC can't be created outside wxEVT_PAINT handler" );
 }
 
--- a/src/qt/glcanvas.cpp
+++ b/src/qt/glcanvas.cpp
@@ -16,42 +16,26 @@
 #include <QtWidgets/QGestureRecognizer>
 #include <QtWidgets/QGestureEvent>
 
-#if defined(__VISUALC__)
-    #pragma message("OpenGL support is not implemented in wxQt")
-#else
-    #warning "OpenGL support is not implemented in wxQt"
-#endif
 wxGCC_WARNING_SUPPRESS(unused-parameter)
 
 class wxQtGLWidget : public wxQtEventSignalHandler< QGLWidget, wxGLCanvas >
 {
 public:
     wxQtGLWidget(wxWindow *parent, wxGLCanvas *handler, QGLFormat format)
-        : wxQtEventSignalHandler<QGLWidget,wxGLCanvas>(parent, handler)
-        {
-            setFormat(format);
-            setAutoBufferSwap( false );
-        }
+        : wxQtEventSignalHandler<QGLWidget, wxGLCanvas>(parent, handler)
+    {
+        setFormat(format);
+        setAutoBufferSwap(false);
+        setFocusPolicy(Qt::StrongFocus);
+    }
 
 protected:
-    virtual void showEvent ( QShowEvent * event ) override;
-    virtual void hideEvent ( QHideEvent * event ) override;
     virtual void resizeEvent ( QResizeEvent * event ) override;
     virtual void paintEvent ( QPaintEvent * event ) override;
 
     virtual void resizeGL(int w, int h) override;
     virtual void paintGL() override;
 };
-
-void wxQtGLWidget::showEvent ( QShowEvent * event )
-{
-    QGLWidget::showEvent( event );
-}
-
-void wxQtGLWidget::hideEvent ( QHideEvent * event )
-{
-    QGLWidget::hideEvent( event );
-}
 
 void wxQtGLWidget::resizeEvent ( QResizeEvent * event )
 {
@@ -84,6 +68,7 @@ wxGLContextAttrs& wxGLContextAttrs::CoreProfile()
 {
 //    AddAttribBits(GLX_CONTEXT_PROFILE_MASK_ARB,
 //                  GLX_CONTEXT_CORE_PROFILE_BIT_ARB);
+    AddAttribute(wx_GL_COMPAT_PROFILE);
     SetNeedsARB();
     return *this;
 }
@@ -92,6 +77,9 @@ wxGLContextAttrs& wxGLContextAttrs::MajorVersion(int val)
 {
     if ( val > 0 )
     {
+        AddAttribute(WX_GL_MAJOR_VERSION);
+        AddAttribute(val);
+
         if ( val >= 3 )
             SetNeedsARB();
     }
@@ -102,12 +90,15 @@ wxGLContextAttrs& wxGLContextAttrs::MinorVersion(int val)
 {
     if ( val >= 0 )
     {
+        AddAttribute(WX_GL_MINOR_VERSION);
+        AddAttribute(val);
     }
     return *this;
 }
 
 wxGLContextAttrs& wxGLContextAttrs::CompatibilityProfile()
 {
+    AddAttribute(wx_GL_COMPAT_PROFILE);
     SetNeedsARB();
     return *this;
 }
@@ -168,7 +159,7 @@ wxGLContextAttrs& wxGLContextAttrs::PlatformDefaults()
 
 void wxGLContextAttrs::EndList()
 {
-//    AddAttribute(None);
+    AddAttribute(0);
 }
 
 // ----------------------------------------------------------------------------
@@ -190,6 +181,7 @@ void wxGLContextAttrs::EndList()
 
 wxGLAttributes& wxGLAttributes::RGBA()
 {
+    AddAttribute(WX_GL_RGBA);
     return *this;
 }
 
@@ -197,24 +189,28 @@ wxGLAttributes& wxGLAttributes::BufferSize(int val)
 {
     if ( val >= 0 )
     {
+        AddAttribute(WX_GL_BUFFER_SIZE);
+        AddAttribute(val);
     }
     return *this;
 }
 
 wxGLAttributes& wxGLAttributes::Level(int val)
 {
-//    AddAttribute(GLX_LEVEL);
+    AddAttribute(WX_GL_LEVEL);
     AddAttribute(val);
     return *this;
 }
 
 wxGLAttributes& wxGLAttributes::DoubleBuffer()
 {
+    AddAttribute(WX_GL_DOUBLEBUFFER);
     return *this;
 }
 
 wxGLAttributes& wxGLAttributes::Stereo()
 {
+    AddAttribute(WX_GL_STEREO);
     return *this;
 }
 
@@ -222,6 +218,8 @@ wxGLAttributes& wxGLAttributes::AuxBuffers(int val)
 {
     if ( val >= 0 )
     {
+        AddAttribute(WX_GL_AUX_BUFFERS);
+        AddAttribute(val);
     }
     return *this;
 }
@@ -230,15 +228,23 @@ wxGLAttributes& wxGLAttributes::MinRGBA(int mRed, int mGreen, int mBlue, int mAl
 {
     if ( mRed >= 0)
     {
+        AddAttribute(WX_GL_MIN_RED);
+        AddAttribute(mRed);
     }
     if ( mGreen >= 0)
     {
+        AddAttribute(WX_GL_MIN_GREEN);
+        AddAttribute(mGreen);
     }
     if ( mBlue >= 0)
     {
+        AddAttribute(WX_GL_MIN_BLUE);
+        AddAttribute(mBlue);
     }
     if ( mAlpha >= 0)
     {
+        AddAttribute(WX_GL_MIN_ALPHA);
+        AddAttribute(mAlpha);
     }
     return *this;
 }
@@ -247,6 +253,8 @@ wxGLAttributes& wxGLAttributes::Depth(int val)
 {
     if ( val >= 0 )
     {
+        AddAttribute(WX_GL_DEPTH_SIZE);
+        AddAttribute(val);
     }
     return *this;
 }
@@ -255,6 +263,8 @@ wxGLAttributes& wxGLAttributes::Stencil(int val)
 {
     if ( val >= 0 )
     {
+        AddAttribute(WX_GL_DEPTH_SIZE);
+        AddAttribute(val);
     }
     return *this;
 }
@@ -263,40 +273,44 @@ wxGLAttributes& wxGLAttributes::MinAcumRGBA(int mRed, int mGreen, int mBlue, int
 {
     if ( mRed >= 0)
     {
+        AddAttribute(WX_GL_MIN_ACCUM_RED);
+        AddAttribute(mRed);
     }
     if ( mGreen >= 0)
     {
+        AddAttribute(WX_GL_MIN_ACCUM_GREEN);
+        AddAttribute(mGreen);
     }
     if ( mBlue >= 0)
     {
+        AddAttribute(WX_GL_MIN_ACCUM_BLUE);
+        AddAttribute(mBlue);
     }
     if ( mAlpha >= 0)
     {
+        AddAttribute(WX_GL_MIN_ACCUM_ALPHA);
+        AddAttribute(mAlpha);
     }
     return *this;
 }
 
 wxGLAttributes& wxGLAttributes::SampleBuffers(int val)
 {
-#ifdef GLX_SAMPLE_BUFFERS_ARB
-    if ( val >= 0 && wxGLCanvasX11::IsGLXMultiSampleAvailable() )
+    if ( val >= 0 )
     {
-        AddAttribute(GLX_SAMPLE_BUFFERS_ARB);
+        AddAttribute(WX_GL_SAMPLE_BUFFERS);
         AddAttribute(val);
     }
-#endif
     return *this;
 }
 
 wxGLAttributes& wxGLAttributes::Samplers(int val)
 {
-#ifdef GLX_SAMPLES_ARB
-    if ( val >= 0 && wxGLCanvasX11::IsGLXMultiSampleAvailable() )
+    if ( val >= 0 )
     {
-        AddAttribute(GLX_SAMPLES_ARB);
+        AddAttribute(WX_GL_SAMPLES);
         AddAttribute(val);
     }
-#endif
     return *this;
 }
 
@@ -309,11 +323,12 @@ wxGLAttributes& wxGLAttributes::FrameBuffersRGB()
 
 void wxGLAttributes::EndList()
 {
+    AddAttribute(0);
 }
 
 wxGLAttributes& wxGLAttributes::PlatformDefaults()
 {
-    // No GLX specific values
+    // No Qt specific values
     return *this;
 }
 
@@ -324,15 +339,33 @@ wxGLAttributes& wxGLAttributes::PlatformDefaults()
 
 wxIMPLEMENT_CLASS(wxGLContext, wxWindow);
 
-wxGLContext::wxGLContext(wxGLCanvas *WXUNUSED(win), const wxGLContext* WXUNUSED(other), const wxGLContextAttrs *WXUNUSED(ctxAttrs))
+wxGLContext::wxGLContext(wxGLCanvas *win,
+                         const wxGLContext *other,
+                         const wxGLContextAttrs *ctxAttrs)
 {
+    QGLWidget *qglWidget = static_cast<QGLWidget *>(win->GetHandle());
+    m_glContext = qglWidget->context();
+
+    if (m_glContext != nullptr)
+    {
+        m_isOk = true;
+    }
 }
 
-bool wxGLContext::SetCurrent(const wxGLCanvas&) const
+bool wxGLContext::SetCurrent(const wxGLCanvas& win) const
 {
-// I think I must destroy and recreate the QGLWidget to change the context?
-//    win->GetHandle()->makeCurrent();
-    return false;
+    QGLWidget *qglWidget = static_cast<QGLWidget *>(win.GetHandle());
+    QGLContext *context = qglWidget->context();
+
+    if (context != m_glContext)
+    {
+        // I think I must destroy and recreate the QGLWidget to change the context?
+        wxLogDebug("Calling wxGLContext::SetCurrent with a different canvas is not supported in wxQt");
+        return false;
+    }
+
+    qglWidget->makeCurrent();
+    return true;
 }
 
 //---------------------------------------------------------------------------
@@ -403,8 +436,11 @@ bool wxGLCanvas::Create(wxWindow *parent,
                         const wxString& name,
                         const wxPalette& palette)
 {
-    wxLogError("Missing implementation of " + wxString(__FUNCTION__));
-    return false;
+    const int* attrsList = dispAttrs.GetGLAttrs();
+
+    wxCHECK_MSG(attrsList, false, "wxGLAttributes object is empty.");
+
+    return Create(parent, id, pos, size, style, name, attrsList, palette);
 }
 
 bool wxGLCanvas::Create(wxWindow *parent,
@@ -421,9 +457,22 @@ bool wxGLCanvas::Create(wxWindow *parent,
 #endif // wxUSE_PALETTE
     wxUnusedVar(palette); // Unused when wxDEBUG_LEVEL==0
 
-    QGLFormat format;
-    if (!wxGLCanvas::ConvertWXAttrsToQtGL(attribList, format))
+    // Separate display/context attributes, set defaults.
+    wxGLAttributes dispAttrs;
+    wxGLContextAttrs ctxAttrs;
+    if (!ParseAttribList(attribList, dispAttrs, &ctxAttrs))
         return false;
+
+    QGLFormat format;
+    if (!wxGLCanvas::ConvertWXAttrsToQtGL(dispAttrs, ctxAttrs, format))
+        return false;
+
+    // Return false if any attribute is unsupported
+    if ( !IsDisplaySupported(attribList) )
+    {
+        wxFAIL_MSG("Can't find a pixel format for the requested attributes");
+        return false;
+    }
 
     m_qtWindow = new wxQtGLWidget(parent, this, format);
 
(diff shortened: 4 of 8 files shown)


src/qt/statusbar.cpp
```diff
@@ -102,6 +102,11 @@ bool wxStatusBar::GetFieldRect(int i, wxRect& rect) const
     wxCHECK_MSG( (i >= 0) && ((size_t)i < m_panes.GetCount()), false,
                  "invalid statusbar field index" );
 
+    const_cast<wxStatusBar *>(this)->UpdateFields();
+
+    if (m_qtPanes.empty())
+        return false;
+
```
AliKet, line 109: I don't mind keeping this for safety reasons. But, is the crash reproducible in the **statbar** sample ?

src/qt/dcclient.cpp
```diff
@@ -126,7 +126,5 @@ wxPaintDCImpl::wxPaintDCImpl( wxDC *owner )
 wxPaintDCImpl::wxPaintDCImpl( wxDC *owner, wxWindow *win )
     : wxWindowDCImpl( owner, win )
 {
-    wxCHECK_RET( m_isWindowPainter,
-                 "wxPaintDC can't be created outside wxEVT_PAINT handler" );
```
AliKet, line 130: FWIW, `Qt` won't let you paint anything outside the `paintEvent()` handler. So I don't know how this affects `wxGLCanvas` to be honest!

dsa-t: When the assert message dialog appears in Debug, this causes a recursive repaint and recursive assert for some reason.

wxPaintDC is created inside paint handlers in "cube" and "pyramid" samples, but when wxQtGLWidget paint handlers are called, they didn't set the correct flags/data to make the assert disappear.

src/qt/statusbar.cpp
dsa-t: Here's the usage in KiCad: https://gitlab.com/kicad/code/kicad/-/blob/07d7ccf46fbbaf3c81a56deb8e2e4302ccdbcd47/common/widgets/kistatusbar.cpp#L225

`GetFieldRect` is called before `SetStatusText`, so `m_qtPanes` is empty (if `UpdateFields` isn't called).
"statbar" sample doesn't seem to do that.

if `UpdateFields` _is_ called, `m_qtPanes` will probably never be empty.

src/qt/statusbar.cpp
```diff
@@ -102,6 +102,11 @@ bool wxStatusBar::GetFieldRect(int i, wxRect& rect) const
     wxCHECK_MSG( (i >= 0) && ((size_t)i < m_panes.GetCount()), false,
                  "invalid statusbar field index" );
 
+    const_cast<wxStatusBar *>(this)->UpdateFields();
+
```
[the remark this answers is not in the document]
dsa-t: `UpdateFields` is needed to initialize `m_qtPanes` (see the other comment)

Calling `UpdateFields` here seems ok because other ports initialize data in this method too. I don't know if there exists a better place to put it.

src/qt/statusbar.cpp
AliKet: I see, thanks. so we can rename `UpdateFields()` to `CreateFieldsIfNeeded()` and call it from this function as well as from `DoUpdateStatusText()` and remove this test (i.e. `if ( m_qtPanes.empty() )` ) because the check at the beginning of the function is enough. IOW, if  `m_panes` is empty, the function will assert.

TIA.

src/qt/dcclient.cpp
AliKet: I'll try to look at this a.s.a.p.

src/qt/glcanvas.cpp
```diff
@@ -17,9 +17,9 @@
 #include <QtWidgets/QGestureEvent>
 
 #if defined(__VISUALC__)
-    #pragma message("OpenGL support is not implemented in wxQt")
+    #pragma message("OpenGL support is not implemented fully in wxQt")
```
vadz, line 20: Do we still need this if OpenGL (mostly) works? I'd remove these warnings entirely.

src/qt/glcanvas.cpp
```diff
@@ -403,8 +435,15 @@ bool wxGLCanvas::Create(wxWindow *parent,
...
 {
-    wxLogError("Missing implementation of " + wxString(__FUNCTION__));
-    return false;
+    const int* attrsList = dispAttrs.GetGLAttrs();
+
+    if (!attrsList)
+    {
+        wxFAIL_MSG("wxGLAttributes object is empty.");
+        return false;
+    }
```
vadz, line 444: Minor, but could be just

```suggestion
    wxCHECK_MSG(attrsList, false, "wxGLAttributes object is empty.");
```

src/qt/toolbar.cpp
```diff
@@ -58,7 +58,7 @@ class wxQtToolButton : public QToolButton, public wxQtSignalHandler
 
 public:
     wxQtToolButton(wxToolBar *handler, wxToolBarTool *tool)
-        : QToolButton(handler->GetHandle()),
+        : QToolButton(handler != nullptr ? handler->GetHandle() : nullptr),
```
vadz, line 61: Is it really normal for `handler` to be null here? When does it happen? A comment explaining this would be nice if it really must be supported (otherwise, preventing it from happening would be even nicer).

src/qt/menu.cpp
```diff
@@ -30,7 +30,7 @@ class wxQtActionGroup : public QActionGroup, public wxQtSignalHandler
 
 public:
     explicit wxQtActionGroup( wxMenu* handler )
-        : QActionGroup( handler->GetHandle() ),
+        : QActionGroup( handler != nullptr ? handler->GetHandle() : nullptr ),
```
vadz, line 33: Same question as for the toolbar: is it really normal for `handler` to be null here?

src/qt/dcclient.cpp
vadz: Thanks for looking at this, I'd prefer not to remove this assert as it can catch a common mistake.

However I also think that it would be nice to make asserts in paint handlers non-fatal: currently they always result in a crash because when the assert message box is shown, another paint event is generated and another assert happens and we crash. And this is not wxQt-specific, exactly the same thing happens in wxMSW and wxGTK too. So ideally we should use something like `wxRecursionGuard` to avoid generating this assert if we're already inside it...

vadz (commented):
Thanks for the fixes, but it would be great to understand if the "crash fixes" do the right thing or just workarounds for some bugs elsewhere.

src/qt/dcclient.cpp
AliKet: The real fix is to remove these overrides:
https://github.com/wxWidgets/wxWidgets/blob/5dac42edc07f6e504ecc9b5a9cbb1c559ca9319c/src/qt/glcanvas.cpp#L37-L40

from `wxQtGLWidget` and just let `wxQtEventSignalHandler` (the base class) do its work as normal

src/qt/toolbar.cpp
dsa-t: Searching through the code, it seems like this change and wxQtActionGroup one is not necessary, I will remove them.

src/qt/dcclient.cpp
dsa-t: If I do that, "pyramid" sample only shows dark green canvas for some reason.

src/qt/glcanvas.cpp
dsa-t: Removed.

src/qt/dcclient.cpp
dsa-t: Seems to be working after fixing `wxGLContext::SetCurrent`

src/qt/menuitem.cpp
```diff
@@ -183,7 +183,7 @@ void wxMenuItem::ClearExtraAccels()
 
 wxQtAction::wxQtAction( wxMenu *handler, int id, const wxString &text, const wxString &help,
         wxItemKind kind, wxMenu *subMenu, wxMenuItem *menuItem )
-    : QAction( wxQtConvertString( text ), handler->GetHandle() ),
+    : QAction( wxQtConvertString( text ), handler != nullptr ? handler->GetHandle() : nullptr ),
       wxQtSignalHandler( handler ),
       m_mitemId(menuItem->GetId()), m_isCheckable(menuItem->IsCheckable())
```
AliKet, line 188: Are you sure this fixes the crash you're seeing? because passing `nullptr` to `wxQtSignalHandler` will also cause a crash when `EmitEvent()` is called: https://github.com/wxWidgets/wxWidgets/blob/5dac42edc07f6e504ecc9b5a9cbb1c559ca9319c/include/wx/qt/private/winevent.h#L38-L45

Do have some snippet reproducing the crash by chance ?

BTW, thanks a lot for working on this and making it working.

AliKet: Wait! maybe `EmitEvent()` mentioned above is the **culprit**!! because the comment there says:
>     // We're only called with the objects of class (or derived from) 
>     // wxWindow, so the cast is safe.

which is not true if the handler is wxMenu, because it derives from `wxEvtHandler` and not from `wxWindow`?

We should investigate...

dsa-t: It was crashing when constructing wxMenuItem with nullptr parent, for example:

https://gitlab.com/kicad/code/kicad/-/blob/d0236ca7511b24ef106297482e7c58e65d63d186/common/file_history.cpp#L150-165

I didn't see any other related crashes at runtime.

AliKet: Indeed, constructing `wxMenuItem` with `nullptr` parent is perfectly legal, and is already documented here:
https://docs.wxwidgets.org/3.2/classwx_menu_item.html#a6e9b0e1b786fa84250a42c88d84aed2b:~:text=Menu%20that%20the%20menu%20item%20belongs%20to.%20Can%20be%20NULL%20if%20the%20item%20is%20going%20to%20be%20added%20to%20the%20menu%20later.

And I think we should properly handle this in `wxQt`. i.e.:
- Allow passing `nullptr` to `wxQtAction`.
- Assign a valid parent to `wxQtAction` when the `wxMenuItem` is added or appened to a `wxMenu`.
- Fix `EmitEvent()` to not expect `wxWindow` objects only.

Just let me know if you plan to do this here or in a separate PR. because otherwise, this commit simply hides a potential bug IMHO.

TIA.

dsa-t: >Just let me know if you plan to do this here or in a separate PR.

I'll probably leave it to others, since I don't know all the details here.

AliKet: I'll fix it then. just don't include this commit when you finish working on this.

AliKet: See commit [a021a01](https://github.com/wxWidgets/wxWidgets/pull/24210/commits/a021a01c828198dd192372af8de3fdf15a386e4d) in PR #24210

dsa-t: >just don't include this commit when you finish working on this.

Commit dropped.

src/qt/dcclient.cpp
dsa-t: Seems like Qt messes up the clear color and depth test at each paint if these overrides are removed.

Otherwise, `isosurf` sample, when run from executable's dir looks fine if this is added to `wxQtGLWidget` ctor:
```
        setAttribute(Qt::WA_OpaquePaintEvent, true);
        setAttribute(Qt::WA_NoSystemBackground, true);
```

dsa-t: With latest commits, all samples work with correct background colors and depth testing.

AliKet:
So we definitely need to :
- Override `paintEvent()` in `wxQtGLWidget` because `wxQtEventSignalHandler` (the `wxQtGLWidget` base class) only calls `QGLWidget::paintEvent()` (which, among other things, does important things internally and is responsible for calling `paintGL()` to paint the canvas) if the event was not handled. See https://github.com/wxWidgets/wxWidgets/blob/a2f7a933e861a0d32bbc195a0ed502913ed1fd34/include/wx/qt/private/winevent.h#L315-L324

- Likewise for `resizeEvent()` because the `QGLWidget` should really be called. See this [comment](https://codebrowser.dev/qt6/qtbase/src/openglwidgets/qopenglwidget.cpp.html#1456:~:text=1456-,/*!,1465,-void%20QOpenGLWidget%3A%3A) which is not mentioned in the **Qt** documentation!
- Overriding `showEvent()` and `hideEvent()` shouldn't be necessary IMHO!
- Relaxe the `wxCHECK` in `wxPaintDC` only for `QGLWidget` with something like:
```
    wxCHECK_RET( m_isWindowPainter || win->GetHandle()->inherits("QGLWidget"),
                 "wxPaintDC can't be created outside wxEVT_PAINT handler" );
```

BTW, you can remove this line : https://github.com/wxWidgets/wxWidgets/blob/a2f7a933e861a0d32bbc195a0ed502913ed1fd34/src/qt/glcanvas.cpp#L33

and this line : https://github.com/wxWidgets/wxWidgets/blob/a2f7a933e861a0d32bbc195a0ed502913ed1fd34/src/qt/glcanvas.cpp#L439

and let **Qt** swap the buffer for you if necessary.

Other than that, the changes you made look good to me. Thanks again for your work on this.

src/qt/glcanvas.cpp
```diff
@@ -29,6 +24,9 @@ class wxQtGLWidget : public wxQtEventSignalHandler< QGLWidget, wxGLCanvas >
     wxQtGLWidget(wxWindow *parent, wxGLCanvas *handler, QGLFormat format)
         : wxQtEventSignalHandler<QGLWidget,wxGLCanvas>(parent, handler)
         {
+            setAttribute(Qt::WA_OpaquePaintEvent, true);
+            setAttribute(Qt::WA_NoSystemBackground, true);
+
```
AliKet, line 29: It would be more clear if you do this instead:
- Remove these lines
```suggestion
```
- and in `wxGLCanvas::Create()`, line 495, put this
```
    if ( !wxWindow::Create( parent, id, pos, size, style, name ) )
        return false;

    SetBackgroundStyle(wxBG_STYLE_PAINT);

    return true;

```

dsa-t: `wxBG_STYLE_PAINT` doesn't seem like it enables `WA_NoSystemBackground`

AliKet: > `wxBG_STYLE_PAINT` doesn't seem like it enables `WA_NoSystemBackground`

Sorry, but why do we need this attribute here? what kind of problems does this solve?

dsa-t: Qt changes the clear color and clears the canvas, also I think it disables depth testing.

This causes wrong background color and disabled depth testing in opengl samples, because they only set this state in InitGL.

I tried removing one of the flag setters, but it didn't work correctly.

AliKet: I couldn't see this happening in any of the samples!

Anyway, `WA_OpaquePaintEvent` according to Qt docs:

> Indicates that the widget paints all its pixels when it receives a paint event. Thus, it is not required for operations like updating, resizing, scrolling and focus changes to erase the widget before generating paint events.

So I don't see the benefit that `WA_NoSystemBackground` adds with `WA_OpaquePaintEvent` attribute ?

You may find [this](https://lists.qt-project.org/pipermail/interest/2015-January/014903.html) useful!

But if you still see that we need both for everything to work properly, then let's do it like this.

dsa-t: Will have to double-check tomorrow.

dsa-t: So turns out WA_NoSystemBackground wasn't needed.

dsa-t:
>So we definitely need to :

Suggestions appled.

>and let Qt swap the buffer for you if necessary.

I think applications expect to swap buffers themselves, so I've left it as is.

src/qt/msgdlg.cpp
```diff
@@ -25,19 +25,28 @@ class wxQtMessageDialog : public wxQtEventSignalHandler< QMessageBox, wxMessageD
 
 
 wxMessageDialog::wxMessageDialog( wxWindow *parent, const wxString& message,
-        const wxString& caption, long style, const wxPoint& pos )
+        const wxString& caption, long style, const wxPoint& WXUNUSED(pos) )
```
vadz, line 28: If specifying position worked before, it seems wrong to ignore it now.

src/qt/msgdlg.cpp
```diff
@@ -108,17 +117,14 @@ wxMessageDialog::wxMessageDialog( wxWindow *parent, const wxString& message,
...
-    Centre(wxBOTH | wxCENTER_FRAME);
 }
 
-wxIMPLEMENT_CLASS(wxMessageDialog,wxDialog);
-
 int wxMessageDialog::ShowModal()
 {
     WX_HOOK_MODAL_DIALOG();
+
+    QtCreateMsgDialog();
```
vadz, line 126: Why do we have to call this from here instead of creating it from the ctor as before? It looks like we could just call `setInformativeText()` here if necessary, couldn't we?

This results in (apparently) unnecessary changes and potentially creates problems, e.g. if `ShowModal()` is called twice.

src/qt/msgdlg.cpp
```diff
@@ -108,17 +117,14 @@ wxMessageDialog::wxMessageDialog( wxWindow *parent, const wxString& message,
 
     if ( style & wxSTAY_ON_TOP )
         dlg->setWindowModality( Qt::ApplicationModal );
-
-    PostCreation(false);
```
vadz, line 112: Just checking, was removing this call intentional?

src/qt/combobox.cpp
```diff
@@ -296,15 +296,21 @@ void wxComboBox::SetSelection( long from, long to )
 
 void wxComboBox::SetInsertionPoint( long pos )
 {
-    // check if pos indicates end of text:
-    if ( pos == -1 )
-        m_qtComboBox->lineEdit()->end( false );
-    else
-        m_qtComboBox->lineEdit()->setCursorPosition( pos );
+    if ( m_qtComboBox->lineEdit() != nullptr )
```
vadz, line 299: This is minor, of course, but I'd write it like

```suggestion
    if ( auto edit = m_qtComboBox->lineEdit() )
```

and then use `edit` inside the `if` to avoid calling `lineEdit()` thrice.

src/qt/dcclient.cpp
```diff
@@ -126,7 +126,7 @@ wxPaintDCImpl::wxPaintDCImpl( wxDC *owner )
 wxPaintDCImpl::wxPaintDCImpl( wxDC *owner, wxWindow *win )
     : wxWindowDCImpl( owner, win )
 {
-    wxCHECK_RET( m_isWindowPainter,
+    wxCHECK_RET( m_isWindowPainter || win->GetHandle()->inherits("QGLWidget"),
```
vadz, line 129: Using dynamic cast/RTTI like this doesn't look great. I don't know why do we need to allow creating `wxPaintDC` outside of `wxEVT_PAINT` handler for OpenGL windows, but if we really do, I'd add `virtual bool QtCanPaintWithoutActivePainter() const { return false; }` and override it in `wxGLCanvas` to return true.

include/wx/qt/glcanvas.h
```diff
@@ -24,6 +24,9 @@ class WXDLLIMPEXP_GL wxGLContext : public wxGLContextBase
 
     virtual bool SetCurrent(const wxGLCanvas& win) const override;
 
+private:
+    QGLContext* m_glContext;
```
vadz, line 28: We should really initialize all members in their declarations, i.e.

```suggestion
    QGLContext* m_glContext = nullptr;
```

src/qt/glcanvas.cpp
```diff
@@ -324,15 +338,34 @@ wxGLAttributes& wxGLAttributes::PlatformDefaults()
 
 wxIMPLEMENT_CLASS(wxGLContext, wxWindow);
 
-wxGLContext::wxGLContext(wxGLCanvas *WXUNUSED(win), const wxGLContext* WXUNUSED(other), const wxGLContextAttrs *WXUNUSED(ctxAttrs))
+wxGLContext::wxGLContext(wxGLCanvas *win,
+                         const wxGLContext *other,
+                         const wxGLContextAttrs *ctxAttrs)
+    : m_glContext(nullptr)
```
vadz, line 344: This is unnecessary, it's set anyhow below.

src/qt/glcanvas.cpp
```diff
@@ -532,13 +588,48 @@ bool wxGLCanvas::ConvertWXAttrsToQtGL(const int *wxattrs, QGLFormat &format)
                 // can we somehow indicate if it's not supported?
                 break;
 
+            default:
+                wxLogDebug(wxT("Unsupported OpenGL attribute %d"),
+                           glattrs[arg]);
+                continue;
+        }
+
+        if ( !isBoolAttr ) {
```
vadz, line 597: Style nitpick:

```suggestion
        if ( !isBoolAttr )
        {
```

vadz (commented):
Thanks again!

Most changes look good, but I'm not sure about wxMessageDialog ones. Perhaps we should merge the rest and leave this one for a separate PR?

AliKet:
Can we just follow the **Qt** documentation and make these changes on top of this PR:

- Add `setAutoFillBackground(false);` to `wxQtGLWidget` ctor. In other words, don't follow `QGLWidget` default behaviour and do what `QOpenGLWidget` does currently instead (and IMHO this is the right thing to do in wx) see _Differences to QGLWidget_ [here](https://doc.qt.io/qt-5/qopenglwidget.html#differences-to-qglwidget) .
- As we are not relying on wxQtEventSignalHandler to handle paint events (because the base class version `QGLWidget::paintEvent()` needs to be called anyway) we have to start painting on the `wxQtGLWidget` object in `paintGL()` function (the standard way/place to render the OpenGL scene) before emitting the **wx** paint event like this (remember that we are still inside `wxQtGLWidget::paintEvent()` handler):
```
  void wxQtGLWidget::paintGL()
  {
      auto painter = GetHandler()->QtGetPainter();
      painter->begin(this);

      wxPaintEvent event( GetHandler() );
      EmitEvent(event);
  }
```
- End painting on the painter in `SwapBuffers()`:
```
bool wxGLCanvas::SwapBuffers()
{

    auto painter = QtGetPainter();

    if ( painter->isActive() )
        return false;

    painter->end();

    static_cast<QGLWidget *>(m_qtWindow)->swapBuffers();

    return true;
}
```
Can you confirm that these changes work correctly for you ?

Note that the **Qt** documentation states that we should use
```
    beginNativePainting();
    endNativePainting();
```
around raw **OpenGL** calls. I tried to add them but nothing is drawn on the screen! did I miss something?

src/qt/dcclient.cpp
AliKet: Sorry! I have another idea which makes this change unnecessary. see https://github.com/wxWidgets/wxWidgets/pull/24200#issuecomment-1906787585

src/qt/combobox.cpp
dsa-t: I tried to match style to the rest of the file, e.g. `if ( m_qtComboBox->lineEdit() != nullptr )` already existed in the file.
From a quick search, there's only 4 usages of `if ( auto` in wx code, that's why I didn't prefer it.

include/wx/qt/glcanvas.h
dsa-t: This one also isn't a common pattern in wx, I didn't see many usages other than `include/wx/private/init.h`.
Suggestion applied though.

src/qt/combobox.cpp
vadz: `auto` can only be used in the code added in 3.1, as we didn't use C++11 before, so this explains why it's relatively rare. But my main point wasn't even about using it, but rather about defining a variable instead of having multiple calls to `lineEdit()`. Again, this is pretty minor, but the code seems more readable to me if we don't have multiple repeated calls to the same function.

dsa-t:
> Can you confirm that these changes work correctly for you ?

With `setAutoFillBackground(false);` and `paintGL` and `SwapBuffers` changes, nothing is rendered (a picture of windows behind the canvas is shown in the window):
```
QWidget::repaint: It is dangerous to leave painters active on a widget outside of the PaintEvent
QWidget::repaint: It is dangerous to leave painters active on a widget outside of the PaintEvent
QPainter::begin: A paint device can only be painted by one painter at a time.
QWidget::repaint: It is dangerous to leave painters active on a widget outside of the PaintEvent
QWidget::repaint: It is dangerous to leave painters active on a widget outside of the PaintEvent
QPainter::begin: A paint device can only be painted by one painter at a time.
QWidget::repaint: It is dangerous to leave painters active on a widget outside of the PaintEvent
QWidget::repaint: It is dangerous to leave painters active on a widget outside of the PaintEvent
QPaintDevice: Cannot destroy paint device that is being painted
Segmentation fault (core dumped)
```

AliKet:
> > Can you confirm that these changes work correctly for you ?
> 
> With `setAutoFillBackground(false);` and `paintGL` and `SwapBuffers` changes, nothing is rendered (a picture of windows behind the canvas is shown in the window):
> 
> ```
> QWidget::repaint: It is dangerous to leave painters active on a widget outside of the PaintEvent
> QWidget::repaint: It is dangerous to leave painters active on a widget outside of the PaintEvent
> QPainter::begin: A paint device can only be painted by one painter at a time.
> QWidget::repaint: It is dangerous to leave painters active on a widget outside of the PaintEvent
> QWidget::repaint: It is dangerous to leave painters active on a widget outside of the PaintEvent
> QPainter::begin: A paint device can only be painted by one painter at a time.
> QWidget::repaint: It is dangerous to leave painters active on a widget outside of the PaintEvent
> QWidget::repaint: It is dangerous to leave painters active on a widget outside of the PaintEvent
> QPaintDevice: Cannot destroy paint device that is being painted
> Segmentation fault (core dumped)
> ```

Oops ! change 

```
    if ( painter->isActive() )
        return false;
```
To this : 

```
    if (  ! painter->isActive() )
        return false;
```

dsa-t:
> Oops ! change
> To this :

Now depth testing is screwed (isosurf sample):
image

src/qt/combobox.cpp
dsa-t: I used `QLineEdit* const lineEdit` that has been used in other places, let me know if it's ok.

dsa-t:
Why do we need to use the Qt painter with GL canvas?

> GetHandler()->InitializeGL();

Seems expensive to initialize something at every paint.

AliKet:
> Why do we need to use the Qt painter with GL canvas?
> 
> > GetHandler()->InitializeGL();
> 
> Seems expensive to initialize something at every paint.

So @vadz 's comment above is the only viable solution I see then.
https://github.com/wxWidgets/wxWidgets/pull/24200#discussion_r1462404801

I also think we should update this code :
https://github.com/wxWidgets/wxWidgets/blob/ac3cea5bcd5a3fc3d52648efb4dceacbcc9d0698/src/qt/dcclient.cpp#L38-L55
to not try to create a painter completely if `QtCanPaintWithoutActivePainter()` returns true.

AliKet:
BTW, there is a missing call to `setFocusPolicy(Qt::StrongFocus);` in `wxQtGLWidget` ctor.
Without it, the keyboard interaction in the cube sample wouldn't work.

dsa-t:
> BTW, there is a missing call to `setFocusPolicy(Qt::StrongFocus);` in `wxQtGLWidget` ctor. Without it, the keyboard interaction in the cube sample wouldn't work.

Thanks, I was thinking why it didn't work.

dsa-t:
`QtCanPaintWithoutActivePainter` added.

vadz:
@AliKet Do you have any comments about this or should this be (re)reviewed and merged?

AliKet:
> @AliKet Do you have any comments about this or should this be (re)reviewed and merged?

No other comments other than what I said in this https://github.com/wxWidgets/wxWidgets/pull/24200#issuecomment-1908624364 because I don't see why we need to create a painter that will not be used at all. So I think we can safely do this:
```
wxWindowDCImpl::wxWindowDCImpl( wxDC *owner, wxWindow *win )
    : wxQtDCImpl( owner )
{
    m_window = win;
    m_qtPainter = m_window->QtGetPainter(); // guaranteed to be non-null


    if ( m_qtPainter->isActive() )
    {
        m_isWindowPainter = true;
    }
    else if ( !m_window->QtCanPaintWithoutActivePainter() )
    {
        m_qtPainter = new QPainter();
        m_pict.reset( new QPicture() );
        m_ok = m_qtPainter->begin( m_pict.get() );
        QtPreparePainter();
    }
}
```

@dsa-t do you see any problem with this suggestion ?

dsa-t:
> @dsa-t do you see any problem with this suggestion ?

I'd need more time to check if this works.

vadz:
Do I understand correctly that this still shouldn't be merged (or, at least, that we're not sure if it should)?

AliKet:
> Do I understand correctly that this still shouldn't be merged (or, at least, that we're not sure if it should)?

I'm just waiting for confirmation from @dsa-t because I still think it's better to avoid building QPainter with wxGLCanvas completely and it won't cause any harm as far as I can see.

vadz:
Sorry, I'm still confused, as the two replies seem to be contradictory. @AliKet Are you for merging this now or for waiting until the better version can be tested and confirmed to work?

AliKet:
> Sorry, I'm still confused, as the two replies seem to be contradictory. @AliKet Are you for merging this now or for waiting until the better version can be tested and confirmed to work?

Sure, this can be merged as is for now. we can always improve things later. TIA.

src/qt/glcanvas.cpp
```diff
@@ -16,43 +16,27 @@
...
-        : wxQtEventSignalHandler<QGLWidget,wxGLCanvas>(parent, handler)
-        {
-            setFormat(format);
-            setAutoBufferSwap( false );
-        }
+        : wxQtEventSignalHandler<QGLWidget, wxGLCanvas>(parent, handler)
+    {
+        setFormat(format);
+        setAutoBufferSwap(false);
+        setFocusPolicy(Qt::StrongFocus);
```
vadz, line 29: BTW, I find it surprising that we don't call `setFocusPolicy()` at `wxWindow` level, it ought to take into account the return values of `AcceptsFocus()` and `AcceptsFocusFromKeyboard()` to determine the correct value to call it with.

This doesn't prevent this PR from being merged, of course, it just seems strange to do this only for the OpenGL canvas but not the other windows.

src/qt/glcanvas.cpp
```diff
@@ -440,26 +494,33 @@ bool wxGLCanvas::SwapBuffers()
...
-        return true;
+    const int *glattrs = wxGLAttrs.GetGLAttrs();
+    const int *ctxattrs = wxCtxAttrs.GetGLAttrs();
 
     // set default parameters to false
     format.setDoubleBuffer(false);
     format.setDepth(false);
     format.setAlpha(false);
-    format.setStencil(false);
+    format.setStencil(true);
```
vadz, line 512: This is inconsistent with the comment above now. As [the docs](https://doc.qt.io/qt-5/qglformat.html#setStencil) say that it's enabled by default, maybe we should just remove this line?

```suggestion
```

src/qt/glcanvas.cpp
```diff
@@ -440,26 +494,33 @@ bool wxGLCanvas::SwapBuffers()
...
 {
-    if (!wxattrs)
-        return true;
+    const int *glattrs = wxGLAttrs.GetGLAttrs();
+    const int *ctxattrs = wxCtxAttrs.GetGLAttrs();
 
     // set default parameters to false
     format.setDoubleBuffer(false);
     format.setDepth(false);
     format.setAlpha(false);
```
vadz, line 511: BTW, this seems unnecessary too, as it's disabled by default according to [the docs](https://doc.qt.io/qt-5/qglformat.html#setAlpha), so maybe remove it too?

```suggestion
```

vadz (commented):
I'd like to merge this, but I don't know if I should keep the existing commits, which is not ideal, as some of them are fixups, use squash-merge, which is not ideal as there are several unrelated things here, or try rebasing them -- which would be ideal, but it would be better if it were done by you, as you're more familiar with your changes than I am.

I'd also like to either remove the `setStencil(true)` call or update the comment to explain what exactly are we doing here.

Please let me know if you can do this? TIA!

src/qt/glcanvas.cpp
dsa-t: This should be set to false, now corrected.

src/qt/glcanvas.cpp
dsa-t: Since we call `format.setAlpha(true);` later, I think it's better to set it to false explicitly here.

dsa-t:
> do you see any problem with this suggestion ?

It crashes unfortunately

Screenshot_2024-02-12_15-40-41

AliKet:
> > do you see any problem with this suggestion ?
> 
> It crashes unfortunately
> 
> Screenshot_2024-02-12_15-40-41

Thanks for testing this...

I don't know way `m_ok` is initialized to `true` by default here:
https://github.com/wxWidgets/wxWidgets/blob/57349c2aaaddfd5a835aa968b071e189c8c368cf/src/qt/dc.cpp#L55
If we can't change this to `false` (I didn't test to see the implication) in the base class, we can add **else branch** and explicitly set `m_ok` to `false` in `wxWindowDCImpl` ctor.
and this should fix the crash you see.

Thanks again.

dsa-t:
If `m_ok = true;` is removed, the crash is the same.

AliKet:
> If `m_ok = true;` is removed or set to false, the crash is the same.

I really wanted to avoid creating `QPainter` altogether because it can't be used for drawing anyway. So let this be merged without my suggestion then if it will lead to a crash!

vadz:
As discussed before, let's merge this for now and try to improve it later (FWIW I agree that the situation with the painter is not very clear and could probably be improved).

Thanks for your work on this!

PR #24204, opened 2024-01-11
wxVListBox:  add GetCurrent()/SetCurrent() accessors

wsu-cb:


vadz:
Thanks!

Do you need this in 3.2? If so, we need some extra boilerplate for these functions. If not, I'll just change "@since" lines to say "3.3.0" and merge this (only) into master.

wsu-cb:
> Do you need this in 3.2? If so, we need some extra boilerplate for these functions. If not, I'll just change "@SInCE" lines to say "3.3.0" and merge this (only) into master.

I'm building from master, so I do not need this in 3.2.  Applying this only in 3.3.0 is fine.

I specified 3.2.5 because that appears to be the next version number likely to appear, and I didn't think this could break the ABI.  For my edification, what additional work is needed for 3.2?

wsu-cb:
I see that the wxMSW/Univ build error appears to be a real error (or at least a real warning treated as error) rather than a spurious failure.  However, I do not know what the correct resolution is.  The compiler is reporting that wxUniv has a `virtual bool wxWindow::SetCurrent(bool)`, which is hidden by the new `wxVListBox::SetCurrent()`.  I could (assuming no conflict with some other target) change these functions to `GetCurrentItem()` and `SetCurrentItem()` to solve the problem in this case, but what is the "general" solution to avoiding such a conflict?  Or is this just an annoyance one must live with each time it pops up?

wsu-cb:
The wxX11 and wxDFB problems are repeats of the wxUniv problem, so they will all be fixed together.

The AppVeyor error looks spurious to me, so I hope it will not happen on the build after pushing a fix for the real problem.

vadz:
I'm not sure what to do about `SetCurrent()` clash, but one possibility would be to remove `wxWindow` function to `WXMakeCurrent()` or something like this, i.e. prefix it with "WX" to ensure that it doesn't conflict with any public or user-defined functions.

wsu-cb:
> I'm not sure what to do about `SetCurrent()` clash, but one possibility would be to remove `wxWindow` function to `WXMakeCurrent()` or something like this, i.e. prefix it with "WX" to ensure that it doesn't conflict with any public or user-defined functions.

I just found that `.../include/wx/glcanvas.h` line 261 is handling the same problem by providing an additional declaration of `SetCurrent(bool doit)`.  Should I do that instead?

vadz:
You could do this (or just `using wxWindow::SetCurrent;`) but this is a bit ugly, derived classes shouldn't have to work around differences between ports like this.

wsu-cb:
> I'm not sure what to do about `SetCurrent()` clash, but one possibility would be to remove `wxWindow` function to `WXMakeCurrent()` or something like this, i.e. prefix it with "WX" to ensure that it doesn't conflict with any public or user-defined functions.

I have done this done this, and run the checks twice.  In both runs, the checks failed, but different checks failed on each run, so I think they are spurious failures.  Therefore, I think this is ready for a review.

vadz:

PR #24209, opened 2024-01-11
Add missing accessibility methods to interface

DietmarSchwertberger:
`wxWindow->CreateAccessible()` and `wxWindow->GetOrCreateAccessible()` are missing from `interface/wx/window.h`.
Therefore these methods are currently not added to wxPython builds.

I have tested with a wxPython build and they can be supported there, once the interface is updated.
(I will submit a wxPython PR for this.)

I think, no `@since ...` need to be added, as the methods are only new for wxPython, not for wxWidgets.

### DIFF
--- a/interface/wx/window.h
+++ b/interface/wx/window.h
@@ -3462,6 +3462,17 @@ public:
     */
     void SetAccessible(wxAccessible* accessible);
 
+    /**
+        Override to create a specific accessible object.
+    */
+    virtual wxAccessible* CreateAccessible();
+
+    /**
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, pushed with a minor fix.

DietmarSchwertberger:
Thanks a lot. I have also started working on accessibility support for wxGrid.

swt2c:
@vadz can you backport this one please, since wxPython follows the 3.2 branch?

PR #24210, opened 2024-01-11
wxQt: Some more fixes and cleanups

AliKet:


AliKet:
Oops ! I didn't expect this to fail decause i just tested the menu sample and a modified version of the minimal sample by adding a wxMenuItem created with nullptr parent and everything goes well.
I'll fix the failures tomorrow...

AliKet:
@vadz  This is ready for review

include/wx/qt/menuitem.h
```diff
@@ -46,9 +43,12 @@ class WXDLLIMPEXP_CORE wxMenuItem : public wxMenuItemBase
     virtual void ClearExtraAccels() override;
 #endif // wxUSE_ACCEL
 
+    // implementation
+    void CreateQtAction(wxMenu* parentMenu);
```
vadz, line 47: Very minor, but I'd rename this to `QtCreateAction()` for consistency with the other impl-only functions that use port-specific prefixes.

src/qt/menuitem.cpp
```diff
@@ -47,9 +42,17 @@ class wxQtAction : public QAction, public wxQtSignalHandler
...
-    void onActionToggled( bool checked );
-    void onActionTriggered( bool checked );
+    void onActionToggled( bool checked )
+    {
+        m_handler->Check(m_mitemId, checked);
+    }
 
+    void onActionTriggered( bool checked )
+    {
+        m_handler->SendEvent(m_mitemId, m_isCheckable ? checked : -1 );
```
vadz, line 52: Extreme nitpick:
```suggestion
        m_handler->SendEvent(m_mitemId, m_isCheckable ? checked : -1);
```

vadz (approved):
Thanks for the cleanup, it looks much better like this!

I'd like to rename the function just for consistency but no comments otherwise. Of course, if there are good reasons to keep the current name, it's not a problem neither, just please let me know. Otherwise please do it or tell me if I should do it myself when merging. TIA!

AliKet:
Review suggestions applied.

vadz:
Thanks for the update! Will merge in a moment.

PR #24211, opened 2024-01-12
Add wxRibbonBar::GetPageById()

Blake-Madden:
To find a page by ID currently you have to do a recursive `FindWindow` (which would also look at the child Panel windows). This function allows for finding a page by ID more elegantly.

Also, I have to admit that I initially thought `GetPage()` expected an ID (not an index), so having this function makes the overall API a little more clear (IMO).


vadz:
Thanks, applying now!

PR #24214, opened 2024-01-14
Update nanosvg to the latest upstream version

### LINKED ISSUE #24213: Invalid memory read on nanosvgrast.h
### Description

Good afternoon!
My application crashes due to an error. Error in this line
https://github.com/wxWidgets/nanosvg/blob/ccdb1995134d340a93fb20e3a3d323ccb3838dd0/src/nanosvgrast.h#L1285

The error has been fixed in this comit
https://github.com/memononen/nanosvg/commit/55b64b5070f937ddb65a8dc5985f9948d3c52d37
This is a different fork and I do not know if you can fix this annoying little mistake.

- wxWidgets version  you use: 3.2.2

vadz:
This notably fixes crash with SVGs using 1 stop gradients.

Closes #24213.

### DIFF
--- a/tests/graphics/bmpbundle.cpp
+++ b/tests/graphics/bmpbundle.cpp
@@ -396,7 +396,7 @@ TEST_CASE("BitmapBundle::FromSVG-alpha", "[bmpbundle][svg][alpha]")
     // alpha pixel values in wxImage, allowing for roundoff error.
     CHECK( (int)img.GetRed(0, 1) >= 0x3c );
     CHECK( (int)img.GetRed(0, 1) <= 0x3f );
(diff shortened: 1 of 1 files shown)


vadz:
Hmm, this breaks the recently added tests for anti-aliasing with alpha. There are some changes related to this upstream, but they were supposed to improve rounding, so it's a bit weird that they require relaxing the checks (we get 0x7b instead of expected 0x7c which is supposed to be "something close to 0x7f").

@teliseo Could you please check if the new nanosvg still works well with the images you're using? If so, I'll just make the check larger and merge it, as I don't see what else we can do.

PR #24215, opened 2024-01-14
Add wxXmlParseError structure, update wxXmlDocument::Load() to optionally use it

### LINKED ISSUE #24167: Proposal: add GetLastError() to wxXmlDocument
### Description

When `wxXmlDocument::Load()` fails, it generates a `wxLogError()` message with the description of the error and the line number in the file it was trying to load when the error occurred. That's quite useful if you are running a Debug build of your application where you can access logging. A Release build may not have logging enabled, but it can still be useful to get that error message. The information is, in fact, available to the `Load()` function in Release builds, it's just not available to the caller.

A simple way to save the information for the caller to retrieve would be to add to the header file:

```C++
public:
    const wxString& GetLastError() const { return m_err_msg; }  // Returns error message from last Load() call.

private:
    wxString m_err_msg;
```

In xml.cpp, if there is a failure, you could call:

```C++
            m_err_msg.Printf(_("XML parsing error: '%s' at line %d"),
                             error.c_str(),
                             (int)XML_GetCurrentLineNumber(parser));
```

I've actually done this in a private build, and verified that the the error string does get correctly created with description and line number in a release build.

So my question is, would this be worth submitting a PR for?

Randalphwa:
This PR declares a `wxXmlParseError` structure and adds an optional pointer to this structure to both versions of wxXmlDocument::Load(). If a parsing error occurs, this structure will be filled in with a descriptive message and the line, column and byte offset where the error occurred.

Note that this also fixes a bug in xml.h where `#ifdef WXWIN_COMPATIBILITY_3_2` was used to call the deprecated functions instead of `#if WXWIN_COMPATIBILITY_3_2` which allows setup to use `#define WXWIN_COMPATIBILITY_3_2 0`.

Closes #24167

### DIFF
--- a/include/wx/xml/xml.h
+++ b/include/wx/xml/xml.h
@@ -20,6 +20,7 @@
 #include "wx/list.h"
 #include "wx/textbuf.h"
 #include "wx/versioninfo.h"
+#include "wx/filefn.h"
 
 #include <memory>
 
--- a/include/wx/xml/xml.h
+++ b/include/wx/xml/xml.h
@@ -222,6 +222,16 @@ enum wxXmlDocumentLoadFlag
     wxXMLDOC_KEEP_WHITESPACE_NODES = 1
 };
 
+// Create an instance of this and pass it to wxXmlDocument::Load()
+// to get detailed error information in case of failure. Values
+// are set to -1 if no error occurred.
+struct wxXmlParseError
+{
+    wxString message;
+    int line;
+    int column;
+    int byte_offset;
+};
 
 // This class holds XML data/document as parsed by XML parser.
 
@@ -238,8 +248,8 @@ public:
 
     // Parses .xml file and loads data. Returns TRUE on success, FALSE
     // otherwise.
-    bool Load(const wxString& filename, int flags = wxXMLDOC_NONE);
-    bool Load(wxInputStream& stream, int flags = wxXMLDOC_NONE);
+    bool Load(const wxString& filename, int flags = wxXMLDOC_NONE, wxXmlParseError* err_details = nullptr);
+    bool Load(wxInputStream& stream, int flags = wxXMLDOC_NONE, wxXmlParseError* err_details = nullptr);
 
     // Saves document as .xml file.
     virtual bool Save(const wxString& filename, int indentstep = 2) const;
--- a/include/wx/xml/xml.h
+++ b/include/wx/xml/xml.h
@@ -223,14 +223,13 @@ enum wxXmlDocumentLoadFlag
 };
 
 // Create an instance of this and pass it to wxXmlDocument::Load()
-// to get detailed error information in case of failure. Values
-// are set to -1 if no error occurred.
+// to get detailed error information in case of failure.
 struct wxXmlParseError
 {
     wxString message;
-    int line;
-    int column;
-    int byte_offset;
+    int line = 0;
+    int column = 0;
+    wxFileOffset offset = 0;
 };
 
 // This class holds XML data/document as parsed by XML parser.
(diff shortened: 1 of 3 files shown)


include/wx/xml/xml.h
```diff
@@ -222,6 +222,16 @@ enum wxXmlDocumentLoadFlag
...
 };
 
+// Create an instance of this and pass it to wxXmlDocument::Load()
+// to get detailed error information in case of failure. Values
+// are set to -1 if no error occurred.
+struct wxXmlParseError
+{
+    wxString message;
+    int line;
+    int column;
```
vadz, line 232: I'd initialize those, i.e. do something like

```suggestion
    int line = 0;
    int column = 0;
```

BTW, you used -1 as "invalid" indicator, but as I think lines and columns are 1-based and not 0-based, IMO 0 makes more sense.

include/wx/xml/xml.h
```diff
@@ -222,6 +222,16 @@ enum wxXmlDocumentLoadFlag
...
+    int byte_offset;
```
vadz, line 233: I think this one should use `wxFileOffset` as it's conceivable to work with the files > 2GiB.

If not, the casts to it will at need to be checked.

include/wx/xml/xml.h
```diff
@@ -238,8 +248,8 @@ class WXDLLIMPEXP_XML wxXmlDocument : public wxObject
 
     // Parses .xml file and loads data. Returns TRUE on success, FALSE
     // otherwise.
-    bool Load(const wxString& filename, int flags = wxXMLDOC_NONE);
-    bool Load(wxInputStream& stream, int flags = wxXMLDOC_NONE);
+    bool Load(const wxString& filename, int flags = wxXMLDOC_NONE, wxXmlParseError* err_details = nullptr);
```
vadz, line 251: Very minor, but we use `camelCase` rather than `snake_case` convention, so this should rather be `errDetails` (or just `err` if you'd prefer).

interface/wx/xml/xml.h
```diff
@@ -827,15 +844,24 @@ class wxXmlDocument : public wxObject
         less memory however makes impossible to recreate exactly the loaded text with a
         Save() call later. Read the initial description of this class for more info.
 
+        Create an wxXmlParseError object and pass it to this function to get more
+        information if an error occurred during XML parsing.
+
         Returns true on success, false otherwise.
+
+       @since 3.3.0
```
vadz, line 852: Indentation seems to be wrong here, maybe there is a hard TAB?

src/xml/xml.cpp
```diff
@@ -857,9 +859,23 @@ bool wxXmlDocument::Load(wxInputStream& stream, int flags)
             wxLogError(_("XML parsing error: '%s' at line %d"),
                        error.c_str(),
                        (int)XML_GetCurrentLineNumber(parser));
+            if (err_details)
```
vadz, line 862: I would still put `wxLogError()` in `else` part of this check for the reasons previously discussed.

Maybe I'm missing something, could you please explain why do you think it would be useful to both log and return the error?

src/xml/xml.cpp
```diff
@@ -857,9 +859,23 @@ bool wxXmlDocument::Load(wxInputStream& stream, int flags)
...
+            {
+                err_details->message = error;
+                err_details->line = (int)XML_GetCurrentLineNumber(parser);
+                err_details->column = (int)XML_GetCurrentColumnNumber(parser);
+                err_details->byte_offset = (int)XML_GetCurrentByteIndex(parser);
```
vadz, line 867: This cast here might well be not lossless.

vadz (commented):
Thanks, globally this looks good, but there is still the question of whether we should return error in addition to or instead of logging -- and a couple of other minor remarks.

src/xml/xml.cpp
Randalphwa: No reason I can think of other then my tendency to avoid touching existing code without a good reason. Presumably anyone that uses the structure will have their own message without needing the log message, or they could just call wxLogError themselves.

src/xml/xml.cpp
```diff
@@ -857,9 +859,23 @@ bool wxXmlDocument::Load(wxInputStream& stream, int flags)
...
+            }
             ok = false;
             break;
         }
+        else if (err_details)
+        {
+            err_details->message.clear();
+            err_details->line = -1;
+            err_details->column = -1;
+            err_details->byte_offset = -1;
```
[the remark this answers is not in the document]
Randalphwa: Actually, the more I think about it, I don't think there's any reason to do anything with the structure if no error occurred, particularly if the structure is already initialized when the caller created it. Presumably, the caller only checks the structure if the Load returned false -- the docs do say "if an error occurred".

vadz:
Thanks for the update! Will (squash) merge this soon.

tester0077:
Can I up-vote this issue 
It would be very handy and helpful for me. TIA
What is the expected release this will be a part of?

Randalphwa:
It's in wxWidgets 3.3 -- which doesn't have a release date set yet.

PR #24217, opened 2024-01-14
Fix wxEVT_SET_CURSOR handling under wxQt

AliKet:
Really fix `wxSetCursorEvent` handling for `wxAUI` done in commit f608b34 (Send set cursor events whenever there is mouse movement) without breaking the other controls, i.e. `wxSplitterWindow`, `wxHeaderCtrl` and `wxGrid`.

Also pass the mouse position by const reference to `QtSendSetCursorEvent()`.

### DIFF
--- a/src/qt/window.cpp
+++ b/src/qt/window.cpp
@@ -284,7 +284,7 @@ void wxWindowQt::QtSendSetCursorEvent(wxWindowQt* win, const wxPoint& posClient)
         event.SetId(win->GetId());
         event.SetEventObject(win);
 
-        bool processedEvtSetCursor = win->HandleWindowEvent(event);
+        const bool processedEvtSetCursor = win->HandleWindowEvent(event);
         if ( processedEvtSetCursor && event.HasCursor() )
         {
             win->SetCursor( event.GetCursor() );
--- a/src/qt/window.cpp
+++ b/src/qt/window.cpp
@@ -274,27 +274,33 @@ static const char WINDOW_POINTER_PROPERTY_NAME[] = "wxWindowPointer";
 }
 
 /* static */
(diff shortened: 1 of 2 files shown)


AliKet:
The failure is unrelated to these changes.

src/qt/window.cpp
```diff
@@ -274,27 +274,33 @@ static const char WINDOW_POINTER_PROPERTY_NAME[] = "wxWindowPointer";
...
     {
-        const wxPoint posClient = w->ScreenToClient(posScreen);
-        wxSetCursorEvent event(posClient.x, posClient.y);
-        event.SetEventObject(w);
+        wxSetCursorEvent event( posClient.x , posClient.y );
+        event.SetId(win->GetId());
+        event.SetEventObject(win);
 
-        const bool processedEvtSetCursor = w->ProcessWindowEvent(event);
+        bool processedEvtSetCursor = win->HandleWindowEvent(event);
```
vadz, line 287: Very minor, but looks like `const` could remain:

```suggestion
        const bool processedEvtSetCursor = win->HandleWindowEvent(event);
```

src/qt/window.cpp
```diff
@@ -274,27 +274,33 @@ static const char WINDOW_POINTER_PROPERTY_NAME[] = "wxWindowPointer";
...
         if ( processedEvtSetCursor && event.HasCursor() )
         {
-            win->SetCursor(event.GetCursor());
-            return;
+            win->SetCursor( event.GetCursor() );
```
vadz, line 290: I'd prefer to avoid changing this line too...

vadz (approved):
Thanks for fixing this, will merge soon!

vadz:
Thanks for the update! Will (squash) merge soon.

PR #24220, opened 2024-01-16
Make checks on character entry in numeric validators less relaxed

taler21:
In wxWidgets-3.2.3, precisely with commit cf93b63ca11642bc891570f9b9eef950777b01f6 (Relax checks on character entry in numeric validators, 2023-07-03), the checks on character entry became much more relaxed than necessary.

The input can be restricted at least to not allow values greater than the positive maximum or less than the negative minimum.
In many cases it is not necessary to allow the entry of invalid characters (e.g. if the specified range is 0 to 100).

### DIFF
--- a/include/wx/valnum.h
+++ b/include/wx/valnum.h
@@ -330,6 +330,15 @@ protected:
 
     virtual bool IsInRange(LongestValueType value) const = 0;
 
+    // Check whether the value is in the extended range allowed on input.
+    //
+    // Notice that the minimal and maximal values that have to be accepted on
+    // input may differ from the actually defined range to allow entering
+    // a temporarily invalid number because otherwise it would not be possible
+    // to enter any digits in an initially empty control limited to the values
+    // between "10" and "20".
+    virtual bool IsInInputRange(LongestValueType value) const = 0;
+
     // Implement wxNumValidatorBase pure virtual method.
     virtual bool IsCharOk(const wxString& val, int pos, wxChar ch) const override;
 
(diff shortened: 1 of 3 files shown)


vadz (commented):
Thanks for the changes but I'm not really sure if we want to be so strict here. E.g. while you can enter `1` now in a control with 10..20 range, you can't delete `1` if you have "10" in this control, which seems like a relatively natural thing to do.

So maybe we really should only check for the max value? Because I can't think of any good reason to allow entering values greater than the max, at least for the positive numbers.

And while I don't think anybody uses text controls limited to negative max value, even this would be a problem for them and so to be really safe we probably should only refuse values greater than `abs(max)`.

What do you think?

taler21:
Deleting a character is (and was) always allowed.
https://github.com/wxWidgets/wxWidgets/blob/a2f7a933e861a0d32bbc195a0ed502913ed1fd34/src/common/valnum.cpp#L161-L165

The added check with the (extended) value range only takes place in `IsCharOk()` when a character is entered.
https://github.com/wxWidgets/wxWidgets/blob/a2f7a933e861a0d32bbc195a0ed502913ed1fd34/src/common/valnum.cpp#L179-L181

So in a control with 10..20 range, you still can delete `1` if you have "10" in it.

The case of a negative maximum value I have also taken into account.

At the moment I don't see any input scenario that doesn't work and I also couldn't find any during testing.

vadz:
Sorry, deleting is this indeed allowed. I guess we could imagine the situation in which `10` is selected and you press `0`, but this is probably too artificial.

If nobody else have any objections, I'm going to merge this in master.

Please test if you use numeric validators in your application!

vadz:
No feedback, so merged in the commit above and hopefully people will test it in master.

Thanks again for the improvement!

PR #24222, opened 2024-01-16
Introduce MSW ARM support

lanurmi:
MSVC supports compiling for ARM, but due to not defining any ARCH_SUFFIX for ARM, makefile.vc used to place objects and libraries to the same folder as x86 objects and libraries.

A completely different question is what kind of Windows runs on 32-bit ARM, and whether one can run regular desktop apps on it.

This commit mimics what f69dbaa1 did for ARM64, and adapts it for ARM.

Bakefile has not been run, changes to makefiles are not included, except for manually modified `build/msw/makefile.vc` (in a separate commit, which should not be merged).


vadz:
IMO the question of whether we can run desktop applications on ARM32 is rather important. If we can't, why do we want to add support for it?

lanurmi:
> IMO the question of whether we can run desktop applications on ARM32 is rather important. If we can't, why do we want to add support for it?

Yes, it is. Mainly because there already is implicit support for it, even without this PR. You can compile with `TARGET_CPU=ARM`, and compilation will succeed – but objects and libs are confusingly written to the same folder as those for x86, with mostly clashing names presumably. This is because `ARCH_SUFFIX` is empty by default, and set only for explicitly known architectures, except for x86.

The commit message can be adjusted to emphasize that this PR's main goal is to address that issue with folder names, and any other support is not promised.

src/msw/stdpaths.cpp
```diff
@@ -257,6 +257,7 @@ void wxStandardPaths::IgnoreAppBuildSubDirs()
     IgnoreAppSubDir("Win64");
     IgnoreAppSubDir("x64");
     IgnoreAppSubDir("x86_64");
+    IgnoreAppSubDir("ARM");
```
MaartenBent, line 260: This should probably be in the `#else` section for 32-bit.

MaartenBent (commented):
You can add it to CMake too by mimicking part of 1edb38b54892273f2f932339d49b91f7b60fd3a4 (add two `set(wxPLATFORM_ARCH "arm")`).

build/bakefiles/common.bkl
```diff
@@ -37,6 +37,8 @@
         <if cond="FORMAT=='msvs2008prj' and MSVS_PLATFORM=='win64'">_x64</if>
         <if cond="TARGET_CPU=='amd64'">_x64</if>
         <if cond="TARGET_CPU=='AMD64'">_x64</if>
+        <if cond="TARGET_CPU=='arm'">_arm64</if>
+        <if cond="TARGET_CPU=='ARM'">_arm64</if>
```
vadz, line 41: I'm almost sure that we shouldn't have `64` here, should we?

build/msw/makefile.vc
```diff
@@ -5742,6 +5754,9 @@ __webview_additional_libdirs_arch_FILENAMES = x86
 !if "$(TARGET_CPU)" == "AMD64"
 __webview_additional_libdirs_arch_FILENAMES = x64
 !endif
+!if "$(TARGET_CPU)" == "ARM"
+__webview_additional_libdirs_arch_FILENAMES = arm
+!endif
```
vadz, line 5759: Rebaking removes this part (and the same addition below) and I don't immediately see how to fix this. AFAICS it doesn't really matter because webview is not supported on ARM32 (at least I know that Chromium doesn't support it).

vadz (commented):
Please let me know if I should merge this even without webview changes in the makefile or if you want to try to fix that.

build/bakefiles/common.bkl
lanurmi: That's right, thanks!

build/msw/makefile.vc
lanurmi: Probably it doesn't matter, no. Now the build should at least behave sensibly when compiling without extra components enabled.

I have now squashed my fixes and adjusted the commit message a little. Should be ready to merge after rebaking.

marekr:
FYI, 32-bit ARM is deprecated in Windows and will be removed.

https://www.microsoft.com/en-us/windows/windows-11-specifications#table3

image


https://learn.microsoft.com/en-us/windows/arm/arm32-to-arm64

MaartenBent:
This still needs a `bakefile_gen`. Bakefiles in the repo are currently out-of-date.

vadz:
Sorry, I probably ran `bakefile_gen -b wx.bkl` only, I'll commit/push the changes to the samples etc too soon.

PR #24223, opened 2024-01-17
Added wxRendererNative implementation for wxQt

AliKet:
`render` sample under KDE:

Screenshot_2024-01-15_22-32-53

`dataview` sample under **Xubuntu** (native renderer):

dataview_render

`dataview` sample under **Xubuntu** (generic renderer):

dataview_renderg

### DIFF
--- a/src/qt/renderer.cpp
+++ b/src/qt/renderer.cpp
@@ -1,0 +1,817 @@
+///////////////////////////////////////////////////////////////////////////////
+// Name:        src/qt/renderer.cpp
+// Purpose:     implementation of wxRendererNative for wxQt
+// Author:      Kettab Ali
+// Created:     2024-01-08
+// Licence:     wxWindows licence
+///////////////////////////////////////////////////////////////////////////////
+
+// ============================================================================
+// declarations
+// ============================================================================
+
+// ----------------------------------------------------------------------------
+// headers
+// ----------------------------------------------------------------------------
+
+// for compilers that support precompilation, includes "wx.h".
+#include "wx/wxprec.h"
+
+
+#include "wx/renderer.h"
+
+#ifndef WX_PRECOMP
+    #include "wx/window.h"
+    #include "wx/dcclient.h"
+#endif
+
+#include "wx/headerctrl.h" // for wxHD_BITMAP_ON_RIGHT
+#include "wx/qt/private/converter.h"
+
+#include <QtGui/QPainter>
+#include <QtWidgets/QStyle>
+#include <QtWidgets/QStyleOptionButton>
+#include <QtWidgets/QStyleOptionFrame>
+#include <QtWidgets/QStyleOptionHeader>
+
+namespace
+{
+bool wxIsKDEDesktop()
+{
+    wxString de = wxGetenv(wxS("XDG_CURRENT_DESKTOP"));
+
+    if ( !de.empty() )
+    {
+        // Can be a colon separated list according to
+        // https://wiki.archlinux.org/title/Environment_variables#Examples
+        de = de.BeforeFirst(':');
+    }
+
+    de.MakeUpper();
+
+    return de.Contains(wxS("KDE"));
+}
+}
+
+// ----------------------------------------------------------------------------
+// wxRendererQt: our wxRendererNative implementation
+// ----------------------------------------------------------------------------
+
+class WXDLLEXPORT wxRendererQt : public wxDelegateRendererNative
+{
+public:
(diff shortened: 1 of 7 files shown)


AliKet:
@a-wi @vadz 
Is it ok to apply this patch (adapted from defaults with `wxPG_USE_RENDERER_NATIVE` only set to 1) to `propgrid` :
```
diff --git a/include/wx/propgrid/private.h b/include/wx/propgrid/private.h
index 72c3ca23f9..a776c7a601 100644
--- a/include/wx/propgrid/private.h
+++ b/include/wx/propgrid/private.h
@@ -121,6 +121,35 @@
     // If 1 then controls are refreshed after selected was drawn.
     #define wxPG_REFRESH_CONTROLS 0
 
+#elif defined(__WXQT__)
+
+    // space between vertical line and value text
+    #define wxPG_XBEFORETEXT            5
+    // space between vertical line and value editor widget
+    #define wxPG_XBEFOREWIDGET          1
+
+    // x position adjustment for wxTextCtrl (and like)
+    // left margin cannot be set with wxTextEntry::SetMargins()
+    #define wxPG_TEXTCTRLXADJUST        3
+
+    // comment to use bitmap buttons
+    #define wxPG_ICON_WIDTH             9
+    // 1 if wxRendererNative should be employed
+    #define wxPG_USE_RENDERER_NATIVE    1
+
+    // width of optional bitmap/image in front of property
+    #define wxPG_CUSTOM_IMAGE_WIDTH     20
+
+    // 1 if splitter drag detect margin and control cannot overlap
+    #define wxPG_NO_CHILD_EVT_MOTION    1
+
+    #define wxPG_NAT_BUTTON_BORDER_ANY      0
+    #define wxPG_NAT_BUTTON_BORDER_X        0
+    #define wxPG_NAT_BUTTON_BORDER_Y        0
+
+    // If 1 then controls are refreshed after selected was drawn.
+    #define wxPG_REFRESH_CONTROLS 0
+
 #else // defaults

     // space between vertical line and value text

``` 
Which makes `wxPropertyGrid` look more **Qt**-ish ?

TIA.

src/qt/renderer.cpp
```diff
@@ -0,0 +1,634 @@
...
+// ----------------------------------------------------------------------------
+// list/tree controls drawing
+// ----------------------------------------------------------------------------
+
+int
+wxRendererQt::DrawHeaderButton(wxWindow *win,  wxDC& dc, const wxRect& rect, int flags,
+                               wxHeaderSortIconType sortArrow, wxHeaderButtonParams* params)
+{
+    auto painter = wxGetQtPainter(dc);
+    if ( !painter )
```
vadz, line 161: Is it expected for this to happen or should it be a `wxCHECK()` (here and below)?

vadz (approved):
Thanks for adding this!

No real comments, we definitely should merge it, but just a word about wxPG changes: I think it's ok to add them, as we already do this for the other ports, but it would, of course, be even better to get rid of the hardcoded values for all platforms and use `wxRendererNative::GetWhatever()` instead (adding functions to retrieve the missing values, if necessary).

src/qt/renderer.cpp
AliKet: AFAICT, no, the painter is not expected to be **NULL**. I'll update these and use  `wxCHECK()` instead...

AliKet: I just want to confirm that the painter can only be null if the DC is a `wxMemoryDC` with no bitmap selected. So the added checks are safe for the other types of `wxDC`s and should warn for the `wxMemoryDC` case mentioned above.

AliKet:
@vadz I applied a minor correction here (after your approval, sorry!) about rendering disabled controls. I also implemented `DrawTextCtrl()` and show it in the sample.

And here is the result (under Ubuntu 18.04 _default theme_)...

**Enabled :**

render_ena

**Disabled :**

render_dis

I have no idea why the vertical gauge is not rendered correctly under non-KDE desktop!

I'll try to fix it...

vadz:
Should I wait for the fix to the non-KDE problem mentioned above in this PR or should it be merged?

AliKet:
**Under Ubuntu**

render_norm

**Under Xubuntu**

render_norm_xfce

**Under KDE plasma**

render_norm_kde

AliKet:
> Should I wait for the fix to the non-KDE problem mentioned above in this PR or should it be merged?

Yes, should be ok to merge now. TIA!

vadz:
Thanks to you for implementing this! Will push soon.

PR #24230, opened 2024-01-20
Document CEF_RUNTIME_LIBRARY_FLAG switch for building CEF

vadz:
This must match the options used when building wx itself.

Also mention CEF_USE_SANDBOX while discussing CEF build options.

---

@MaartenBent I think that CMake might want to use this option when configuring CEF instead of using `REGEX_REPLACE` on compile options, but I didn't change `build/cmake/lib/webview_chromium/CMakeLists.txt` myself because I might be missing some reason for which it wouldn't work.

### DIFF
--- a/build/cmake/init.cmake
+++ b/build/cmake/init.cmake
@@ -506,11 +506,36 @@ if(wxUSE_GUI)
             message(WARNING "WebviewChromium libcef_dll_wrapper can only be built with MSVC... disabled")
             wx_option_force_value(wxUSE_WEBVIEW_CHROMIUM OFF)
         endif()
-        if(wxUSE_WEBVIEW_CHROMIUM AND CMAKE_CXX_STANDARD LESS 17)
-            # We shouldn't disable this option as it's disabled by default and
-            # if it is on, it means that CEF is meant to be used, but we can't
-            # continue neither as libcef_dll_wrapper will fail to build.
-            message(FATAL_ERROR "WebviewChromium requires at least C++17 but configured to use C++${CMAKE_CXX_STANDARD}")
+        if(wxUSE_WEBVIEW_CHROMIUM)
+            # Check for C++17 support as it's required by CEF: we trust
+            # CMAKE_CXX_STANDARD if it is defined, but we need to compile a
+            # test program if it is not because the compiler could be
+            # supporting C++17 anyway.
+            if (DEFINED CMAKE_CXX_STANDARD)
+                if (CMAKE_CXX_STANDARD GREATER_EQUAL 17)
+                    set(wxHAVE_CXX17 ON)
+                endif()
+            else()
+                check_cxx_source_compiles("
+                    #if defined(_MSVC_LANG)
+                        #if _MSVC_LANG < 201703L
+                            #error C++17 support is required
+                        #endif
+                    #elif __cplusplus < 201703L
+                        #error C++17 support is required
+                    #endif
+                    int main() {
+                        [[maybe_unused]] auto unused = 17;
+                    }"
+                    wxHAVE_CXX17)
+            endif()
+
+            if (NOT wxHAVE_CXX17)
+                # We shouldn't disable this option as it's disabled by default and
+                # if it is on, it means that CEF is meant to be used, but we can't
+                # continue neither as libcef_dll_wrapper will fail to build.
+                message(FATAL_ERROR "WebviewChromium requires at least C++17 but configured to use C++${CMAKE_CXX_STANDARD}")
+            endif()
         endif()
     endif()
 
--- a/build/cmake/init.cmake
+++ b/build/cmake/init.cmake
@@ -533,8 +533,18 @@ if(wxUSE_GUI)
             if (NOT wxHAVE_CXX17)
                 # We shouldn't disable this option as it's disabled by default and
                 # if it is on, it means that CEF is meant to be used, but we can't
-                # continue neither as libcef_dll_wrapper will fail to build.
-                message(FATAL_ERROR "WebviewChromium requires at least C++17 but configured to use C++${CMAKE_CXX_STANDARD}")
+                # continue neither as libcef_dll_wrapper will fail to build
+                # (actually it may still succeed with CEF v116 which provided
+                # its own stand-in for std::in_place used in CEF headers, but
+                # not with the later versions, so just fail instead of trying
+                # to detect CEF version here, as even v116 officially only
+                # supports C++17 anyhow).
+                if (DEFINED CMAKE_CXX_STANDARD)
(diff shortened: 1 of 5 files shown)


MaartenBent:
Seems indeed like something that can be used, I'll try it out. Maybe the regex was originally added because this option was not yet available.

MaartenBent:
And while you're doing CEF things, I get the following error:
```
CMake Error at build/cmake/setup.cmake:617 (string):
  string no output variable specified
  build/cmake/main.cmake:42 (include)
  CMakeLists.txt:100 (include)


CMake Error at build/cmake/setup.cmake:618 (if):
  if given arguments:

    "STREQUAL" "DEBUG"

  Unknown arguments specified
```

Because `CMAKE_BUILD_TYPE` does not exist with msvc generator, it uses `CMAKE_CONFIGURATION_TYPES`. Can you fix it by using:
```cmake
if(wxUSE_WEBVIEW_CHROMIUM AND DEFINED CMAKE_BUILD_TYPE)
```
Thanks

MaartenBent:
And another one:
`5>cl : command line  error D8021: invalid numeric argument '/Wno-extra'`

```diff
diff --git a/build/cmake/lib/webview_chromium/CMakeLists.txt b/build/cmake/lib/webview_chromium/CMakeLists.txt
index 1c9e28314c..70188ddd45 100644
--- a/build/cmake/lib/webview_chromium/CMakeLists.txt
+++ b/build/cmake/lib/webview_chromium/CMakeLists.txt
@@ -59,7 +59,6 @@ set(USE_SANDBOX OFF)    # Disable usage of sandbox on windows
 add_subdirectory(${CEF_ROOT} ${CEF_BUILD_DIR} EXCLUDE_FROM_ALL)
 set_target_properties(libcef_dll_wrapper PROPERTIES
     FOLDER "Third Party Libraries"
-    COMPILE_FLAGS     "-Wno-extra"
     OUTPUT_NAME "libcef_dll_wrapper"
 )

@@ -68,6 +67,8 @@ if(MSVC)
     get_target_property(cef_compile_options libcef_dll_wrapper COMPILE_OPTIONS)
     string(REGEX REPLACE "/MTd?;" "" cef_compile_options "${cef_compile_options}")
     set_target_properties(libcef_dll_wrapper PROPERTIES COMPILE_OPTIONS "${cef_compile_options}")
+else()
+    set_target_properties(libcef_dll_wrapper PROPERTIES COMPILE_FLAGS "-Wno-extra")
 endif()
```

MaartenBent:
And one last thing :), as mentioned in the chromium PR. If chromium is enabled it uses c++17 `std::in_place_t`. So this check should be updated:
https://github.com/wxWidgets/wxWidgets/blob/2a74f6a6445ea50181f436819bedef6d4fe6d3f6/build/cmake/init.cmake#L513

Often `CMAKE_CXX_STANDARD` is not set and the compiler default will be used. So I think there should be a `wxHAVE_CXX17` check as well. I can add this later.

vadz:
> And another one: `5>cl : command line error D8021: invalid numeric argument '/Wno-extra'`

Thanks, I've applied this but using `COMPILE_OPTIONS` for consistency with the line just above. AFAICS it's the same thing as `COMPILE_FLAGS` in this context (I do wonder if CMake tries to be confusing on purpose, how am I supposed to know which one of these is a list and which one is a string??).

> Often CMAKE_CXX_STANDARD is not set and the compiler default will be used. So I think there should be a wxHAVE_CXX17 check as well. I can add this later.

I've tried adding this too, please let me know if I did anything stupid. TIA!

MaartenBent:
> I've tried adding this too, please let me know if I did anything stupid. TIA!

It seems to work. If I but it on compiler default it runs the test and fails.
But because `CMAKE_CXX_STANDARD` is not defined in this case, the error message just ends in `but configured to use C++`. Possible fix:
```cmake
if (DEFINED CMAKE_CXX_STANDARD)
    set(cxx17_error_details "but configured to use C++${CMAKE_CXX_STANDARD}")
else()
    set(cxx17_error_details "but the compiler defaults do not support this")
endif()
message(FATAL_ERROR "WebviewChromium requires at least C++17 ${cxx17_error_details}")
```

MaartenBent:
Thinking about it, I'm not sure if using preprocessors in the c++17 check is the right thing to do. I tried without it, but the test succeeds in msvc even though it doesn't support `std::in_place_t`.

The best would be using a check using `std::in_place_t` but I didn't find a simple example using it. What about using the following code in the check:
```c++
namespace wx1::wx2::wx3 {}
int main() {
    [[maybe_unused]] auto unused = 17;
}
```

vadz:
> Thinking about it, I'm not sure if using preprocessors in the c++17 check is the right thing to do. I tried without it, but the test succeeds in msvc even though it doesn't support `std::in_place_t`.

Oops, this is news to me. So does it define `_MSVC_LANG == 201703L` even without `/std:c++17`??
 
> The best would be using a check using `std::in_place_t` but I didn't find a simple example using it.

We don't really need to use it, hopefully if it's declared, it should be usable, so just something like

```c++
#include <utility>

int main() {
     [[maybe_unused]] auto unused = std::in_place;
}
```

should work.

> What about using the following code in the check:
> 
> ```c++
> namespace wx1::wx2::wx3 {}
> int main() {
>     [[maybe_unused]] auto unused = 17;
> }
> ```

But we could combine it with this too, of course.

MaartenBent:
Maybe I wasn't clear. I removed the preprocessor checks so only the `main` function remains. Then I use the default compiler (which is `_MSVC_LANG` is `201402L`) and the test passes. If I compile it myself it passes too, just a warning:
 `warning C5051: attribute [[maybe_unused]] requires at least '/std:c++17'; ignored`.

So I suggested to use something that causes an actual compiler error,  instead of using a preprocessor check. I think your suggestion using `[[maybe_unused]] auto unused = std::in_place;` is a good one. We don't need the nested namespaces then.
I verified it with c++20 (pass), c++17 (pass), c++14 (fail) and compiler_default (fail), so it does the job.

vadz:
Sorry, I've indeed misunderstood you, I thought the existing test didn't work. But if it does work, what's the problem with keeping it? Testing for `_MSVC_LANG` is much faster than including `<utility>`, and if it actually works, why replace it?

I've added `[[maybe_unused]]` just as a kind of a sanity check and because we need `main()` anyhow (it's pretty weird to me that `check_cxx_source_compiles()` actually tries to _link_, but this is what it does), but why add more tests? This way we risk ending up with hundreds of lines of test code from `build/aclocal/ax_cxx_compile_stdcxx.m4`...

vadz:
OK, thanks, I'll merge this for now then and we can further improve it later, of course.

Issue #22799, opened 2022-09-14
Some animated gifs not loadable with wxImage


dkulp:
**Describe the bug**
We've had several reports over the years of various animated GIF images not working with xLights.   In general we have the user load the gif into a different editor and resave and that usually "fixes" it and it works.   However, it would be great if wxWidgets could be fixed to support the gifs that currently don't load.

**Expected vs observed behaviour**
The image loads and works.  

Using the attached animated gif, a simple "wxImage::GetImageCount("turkey.gif);" returns 0.   image.LoadFile("turkey.gif", wxBITMAP_TYPE_ANY, 0) returns false indicating it couldn't load it.   

I haven't tried to dig into the gif parsing code to figure out what it's trying to do.   

**Patch or snippet allowing to reproduce the problem**

Problematic gif:
![turkey](https://user-images.githubusercontent.com/91569/190228903-47818762-75bf-4eb7-a376-217fece30099.gif)

Carnildo:
The issue isn't unique to animated GIFs, but that's also where I encountered it.

The problem is that the wxWidgets GIF decoder violates the GIF specification.  From the cover sheet to the [GIF89a spec](https://www.w3.org/Graphics/GIF/spec-gif89a.txt) calling out exactly this mistake:

> There has been confusion about where clear codes can be found in the data stream.  As the specification says, they may appear at anytime.  There is not a requirement to send a clear code when the string table is full.
>
> It is the encoder's decision as to when the table should be cleared.  When the table is full, the encoder can chose to use the table as is, making no changes to it until the encoder chooses to clear it.  The encoder during this time sends out codes that are of the maximum Code Size.

wxGIFDecoder::dgif() errors out as soon as it sees a symbol after the string table fills up.

vadz:
It goes without saying that any fixes/improvements to GIF loading code would be very welcome!

PR #24234, opened 2024-01-21
Fix reading GIF images where the encoder has chosen not to emit a CLE…

Carnildo:
…AR code when the alphabet is full

Fixes bug #22799.

### DIFF
--- a/src/common/gifdecod.cpp
+++ b/src/common/gifdecod.cpp
@@ -443,28 +443,26 @@ wxGIFDecoder::dgif(wxInputStream& stream, GIFImage *img, int interl, int bits)
         // make new entry in alphabet (only if NOT just cleared)
         if (lastcode != -1)
         {
-            // Normally, after the alphabet is full and can't grow any
-            // further (ab_free == 4096), encoder should (must?) emit CLEAR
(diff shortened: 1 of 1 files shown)


vadz:
Thanks! We should apply this, of course, but it would be really nice to add a (small, if possible) GIF showing the problem and a test checking that loading it now succeeds to `tests/image/image.cpp` to ensure that the problem stays fixed in the future. Could you please do it?

Thanks again!

Carnildo:
I don't know how to create a GIF that shows the issue, I can only test on found images.  The ones I'm aware of are dkulp's turkey, at 250kb and an unknown license, and the 600kb image I discovered it with which has a CC-BY-SA 4.0 license.  Is either of these suitable?

vadz:
It's a bit bigger than I'd like, but we have ~150KB images in `tests/image`, so I guess 250KB is not a big deal.

@dkulp Can your image be included in wx test suite, licence-wise?

vadz:
Without response about the turkey image licence we can't include it, so I'm going to apply the patch without the test case, which is not ideal, but better than not resolving the bug at all.

Thanks once again for the fix!

Issue #24225, opened 2024-01-18
Dock info in wxAuiManager should be publicly accessible


craftyjon:
### Description

wxAuiManager stores state in two arrays, `m_panes` and `m_docks`.  There are public APIs to get access to `m_panes` but none for `m_docks`.  This prevents users from being able to save/restore all parts of an AUI state without using Save/LoadPerspective, which we don't want to do for various reasons.

I haven't yet formed an opinion on what APIs should be exposed, but basically we need to be able to get and set the `wxAuiDockInfo` for a given pane.

### Platform and version information

- wxWidgets version  you use: 3.2.1
- wxWidgets port  you use: all
- OS  and its version: all

vadz:
We probably could provide an iterator over `wxAuiDockInfo`, i.e. allow writing

```cpp
for (const wxAuiDockInfo& dockInfo : manager.GetDocks()) {
   ...
}
```

or something like that.

But it would also be nice to have some higher level API for (de)serializing layout information, perhaps we should discuss adding this instead?

craftyjon:
Such an iterator would be fine if we can also stick the information back in (i.e. when we are restoring a saved configuration) somehow.

> But it would also be nice to have some higher level API for (de)serializing layout information, perhaps we should discuss adding this instead?

Perhaps.  In general my thinking is that the client may want control over serialization.  Speaking for KiCad, we just want access to the wx data structures (not serialized) because we have serialization ourselves, so we'd be faced with (/ are faced with, because the only way to do it today is with the save/loadperspective things) converting one serialization format to another.

vadz:
I don't think it's a good idea to provide direct write access to `m_docks`, this would _significantly_ reduce the possibility of changing the implementation later.

By "higher level API" I mean something that would allow you to plug it in your existing serialization format, i.e. it would provide a bridge/adapter that you would plug into -- but would also have some default implementation that could be used by the applications not needing any custom code.

If you could please summarize how does your existing serialization code work, it would be nice. In particular, I wonder how do you identify individual `wxAuiDockInfo` objects, as they don't have any names (should we perhaps add them?), do you do it by their index?

craftyjon:
It doesn't work, at present.  We have no way of saving and restoring floating panes, and we'd like to add it.  The only thing I can think of right now is to extract data from the text format used by SavePerspective.

Perhaps it would be better to get rid of `m_docks` and just make `wxAuiDockInfo` be a child of `wxAuiPaneInfo`?

[...]

PR #24235, opened 2024-01-21
Provide a way to store AUI layouts in any custom format

vadz:
Add wxAuiSerializer and wxAuiDeserializer classes and SaveLayout() and LoadLayout() functions in wxAuiManager using them.

Show how these classes can be used to use XML for storing AUI layout in the sample.

See #24225.

----

This still needs to be documented and the example serializer might be moved to the library itself if we think it can be useful. It also should use DIPs as discussed in #24225.

TODO:

- [x] Rebake
- [x] Document
- [x] Use DIPs


vadz:
@craftyjon @marekr @imciner2 Sorry for mass-tagging but would you have any feedback about this? I'd like to go ahead and finish this, but it would be nice to check that it is fit for purpose, i.e. can be used in KiCad instead of the current serialization system, first. TIA!

craftyjon:
Looks good to me!  Thanks!

marekr:
LGTM

vadz:
I've implemented automatic support for saving (and restoring) all coordinates as DIPs as this is the only thing which really makes sense and it seems better to do it, even if this is inconsistent with the rest of wx API, than keep doing the wrong thing for consistency.

Please let me know if anybody has objections to this, but personally I'd rather try to find some way to use DIPs in `wxPersistentTLW` too in backwards-compatible way.

Any other comments are also welcome, of course!

utelle:
> I've implemented automatic support for saving (and restoring) all coordinates as DIPs as this is the only thing which really makes sense and it seems better to do it, even if this is inconsistent with the rest of wx API, than keep doing the wrong thing for consistency.

It is really great to have this support of saving and restoring AUI layouts in wxWidgets. And IMHO using DIPs is exactly the right thing to do. 👍 

> Please let me know if anybody has objections to this, but personally I'd rather try to find some way to use DIPs in `wxPersistentTLW` too in backwards-compatible way.

I'm fully in favor of using DIPs, even if it were **not** backwards-compatible.

vadz:
Also, while testing this, I've realized that saving/restoring layout doesn't actually restore the window sizes.

I think this should be done but it doesn't seem to be trivial to implement it, at least just saving/restoring `wxAuiPaneInfo::rect` is not enough because this size is overridden by the size computed by the sizers when redoing the layout. I'm going to try to do it, however, as it seems wrong to lose the previously used sizes when restoring a previous layout.

craftyjon:
Wouldn't it be possible to re-apply the saved size after the initial size is computed?  This is what we end up doing as a client application if wx doesn't support save/restore of size internally.

vadz:
> Wouldn't it be possible to re-apply the saved size after the initial size is computed?

I wanted to reapply the saved size even before/instead of computing the new size to avoid wasting time doing it, but I guess we can indeed do it later too, as you're probably not going to restore AUI layouts thousands time per second...

> This is what we end up doing as a client application if wx doesn't support save/restore of size internally.

If you agree to relicense this code under wx licence, I could just copy it instead of rewriting it, please let me know if I can do it (and where this code is). TIA!

craftyjon:
> If you agree to relicense this code under wx licence, I could just copy it instead of rewriting it, please let me know if I can do it (and where this code is). TIA!

Sorry I was not clear, I just mean that we would do such a thing in the future if wxWidgets doesn't do it.  We have various places that do something like this, but not for wxAUI in particular (well, we have some hacks that do some of it, but I think it's not a good universal solution)

To be honest I'm not sure our code should be copied, there are a number of places where we have added "hacks" that seem to work around differences in behavior between different wx platforms that maybe aren't relevant anymore in 3.2, etc.  It all needs a rewrite on our side too.

vadz:
Sorry for a very long delay with this PR, but I'd like to finally merge it.

I'm still not totally happy with it because it looks like it shouldn't be necessary to serialize (and hence deserialize) the docks at all, because they're something created by wxAUI on its own, to accommodate the specified panes, but the way the code is written now, we can only specify the docks initial size when creating them — and so we have to do it when deserializing in order to restore the same geometry. If anybody has any bright idea about how this could be avoided in not too ugly way, I'd really like to hear about it.

But otherwise I think this is as flexible as it gets, the sample shows a possible implementation of XML-based serializer, but using another format would be just as simple or even simpler.

@craftyjon Please let me know if you have any comments!

craftyjon:
This looks really nice, thank you @vadz !

vadz:
@craftyjon I've just opened #24950 because I've realized (well, been told) that we still have a big problem with not saving (and hence not restoring) AUI notebooks layout. Are you affected by this as well or have you found some way to do it on your own? If not, am I right in thinking that you'd be interested in having a solution for this as well?

craftyjon:
Yes, we'd like this.  I haven't started work on KiCad integration with the new storage methods yet because of other KiCad features underway, but we do use notebooks and save/restore for them will be good.

vadz:
I've convinced myself, while working on #24950, that we really shouldn't be saving the docks separately as all the data saved for them is redundant because it can be recovered from the panes geometry (albeit in a rather non-trivial way, which is why I had initially decided to just save the docks directly — but now doing it also for the docks inside notebooks would be even worse, so I'd rather avoid it, finally).

I don't expect anybody to have already started using this, but if anyone did, please keep in mind that the serializer API is going to change soon (and become simpler to implement!).

PR #24237, opened 2024-01-21
Add support for hidden panels in ribbon

Blake-Madden:
If a panel window is hidden, then don't include its size when laying out the ribbon bar. After toggling a panel's visibility, calling Realize() and Layout() will properly include the panel (if shown) or not show it and move panels to the right of it over (without a blank gap in the ribbon bar where the hidden panel would be).

Currently, hiding a panel will result in a blank gap on a ribbon bar. This PR fixes that.

For those curious, this is useful for creating a dynamic ribbon bar, where sections of it change based on context. Destroying and recreating panels can be dangerous if there are pending events, whereas hiding/showing panels is safer and more efficient.

### DIFF
--- a/src/ribbon/page.cpp
+++ b/src/ribbon/page.cpp
@@ -918,7 +918,7 @@ bool wxRibbonPage::ExpandPanels(wxOrientation direction, int maximum_amount)
                   node = node->GetNext(), ++panel_size )
         {
             wxRibbonPanel* panel = wxDynamicCast(node->GetData(), wxRibbonPanel);
-            if(panel == nullptr)
+            if(panel == nullptr || !panel->IsShown())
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, this looks reasonable to me, so I'll merge it soon.

I wonder if this should be documented somewhere or if this can be considered as just a bug fix?

Blake-Madden:
> I wonder if this should be documented somewhere or if this can be considered as just a bug fix?

For another PR, I was thinking about adding `GetPanel`, `GetPanelById`, and maybe `GetPanelCount` functions to `wxRibbonPage`. I could mention panel hiding/showing in the docs for those (and maybe in the overviews too).

PR #24238, opened 2024-01-22
Mention __WXGTK3__ in the preprocessor symbols documentation

taler21:


vadz:

taler21:
Thanks, I think it would make sense to add it in the 3.2 branch too. TIA!

Issue #24023, opened 2023-10-31
wxAuiToolbar's Realize() function causes glitchy rendering


marekr:
This is an old issue that I figured just to throw in the wx issues list

We use Realize() to update toolbars after we may have added or removed buttons from the toolbar.

The problem is, it tries to compute both the horizontal and vertical extents physically on screen, even if it's not actively using one of those layout directions

The result is you can see both directions layed out momentarily on screen before it flashes to the correct direction
![image](https://github.com/wxWidgets/wxWidgets/assets/1838013/ffcadf76-8a4f-4326-b5d7-aa2674e7a4bb)

I actually fixed this in KiCad by just implementing our own Realize() function that avoids computing an unused layout direction.
https://gitlab.com/kicad/code/kicad/-/blob/master/common/tool/action_toolbar.cpp#L772

because users were seeing it
https://gitlab.com/kicad/code/kicad/-/issues/5705

but perhaps there is a smarter way for this to occur in wx?

### Platform and version information

- wxWidgets version  you use: 3.2.3
- wxWidgets port  you use: wxMSW
- OS  and its version:

vadz:
AFAICS the idea not to compute the size not needed by the current orientation is a good optimization in any case because we already call `Realize()` if the orientation changes later, see https://github.com/wxWidgets/wxWidgets/blob/abfdd186ff540e62d18ef3e3d37ce2015843309a/src/aui/auibar.cpp#L2374-L2377

So I don't think we _need_ to do anything smarter. The only other thing I'd do would be to get rid of or at least simplify `RealizeHelper()` as it doesn't need to be passed the orientation at all if it always works with `m_orientation` anyhow. But please let me know if I'm missing something?

Otherwise I'll integrate your changes in wx, thanks!

vadz:
As I realized (sic) later, I did miss something: we need to compute the sizes for both orientations as we return them from `GetHintSize()` which may be called for non-current orientation too (although I'm still not totally sure if this is really needed, as for some reason we intentionally don't show docking hints for the toolbars, so the hint rectangle is not visible anyhow).

So I did do something "smarter", finally, and preserved the calls to `RealizeHelper()` in both orientations but only actually change the window size for the correct one -- as this is not needed to just compute the hint size. Please let me know if you see any problems with #24164.

imciner2:
> we need to compute the sizes for both orientations as we return them from GetHintSize() which may be called for non-current orientation too

[...]

PR #24244, opened 2024-01-24
Restore wxAuiToolBar::RealizeHelper(bool) for compatibility

vadz:
This function was replaced with a more useful function taking wxOrientation and returning wxSize in bbba3b7cb9 (Make wxAuiToolBar::RealizeHelper() return value actually useful, 2023-12-25) but there is some existing code using the old RealizeHelper() from its overridden Realize(), so restore it for compatibility, even if it's quite useless now.

See #24023, #24164.

---

cc @dsa-t @marekr

### DIFF
--- a/include/wx/aui/auibar.h
+++ b/include/wx/aui/auibar.h
@@ -693,6 +693,13 @@ protected:
     bool m_gripperVisible;
     bool m_overflowVisible;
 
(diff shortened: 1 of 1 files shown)


vadz:
Merged by the commit above.

Issue #24246, opened 2024-01-24
wxMSW: wxGrid prevents correct key and focus handling of MenuBar


DietmarSchwertberger:
On the wxpython-users mailing list I got a report about an interference between wxGrid and wxMenuBar which causes accessibility problems for blind users with screen readers.

I'm about to submit a PR to fix this.

#### Bug description:

When pressing ALT (without additional key), a wxMenuBar behaves differently whether a wxGrid has the focus or some other widget.

#### To Reproduce:

You can see the problem from the grid sample.

If you press a shortcut key like ALT-F, the file menu is activated and you can use the arrow keys to navigate around. Everything is fine with this.

When the focus is in the TextCtrl, you can press and release ALT (without an additional key) and after that use the arrow keys to navigate through the menu. Hitting Escape will move the focus back to the TextCtrl.
When the focus is in the Grid and you press just ALT, you can see that the menu is activated (the access keys are underlined).
But, the focus remains in the Grid and the arrow keys are moving the focus between the grid cells.
Also, the 'activation' of the menu bar remains permanent (e.g. the activation keys remain underlined).

### Platform and version information

- wxWidgets version: the current master
- wxWidgets port: wxMSW
- OS: Windows 10 Pro 22H2

PR #24247, opened 2024-01-24
wxGrid: skip EVT_KEY_UP event for pure ALT key

DietmarSchwertberger:
This fixes issue #24246 by skipping EVT_KEY_UP when only the ALT key is pressed.

If you accept the modification, please also port back to 3.2 such that it finds it's way into wxPython soon.

### DIFF
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -6117,9 +6117,12 @@ void wxGrid::OnKeyDown( wxKeyEvent& event )
     }
 }
 
-void wxGrid::OnKeyUp( wxKeyEvent& WXUNUSED(event) )
+void wxGrid::OnKeyUp( wxKeyEvent& event )
 {
-    // try local handlers
+    // try local handlers for anything except pure ALT key
+    // MenuBar needs ALT to enable navigation via arrow keys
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for fixing this, but I have to say that I don't really understand why do we have this handler at all, i.e. why we don't _always_ call `event.Skip()` here or, better yet, remove it entirely (the function would have to remain in 3.2, but we could remove it from the event table).

Do you see any regressions if you do this?

DietmarSchwertberger:
I did a quick check with the modified event table and I could not see issues.

Maybe, apply this PR to 3.2 only, to be on the safe side.
For the current master branch I could submit another PR removing the function completely.

vadz:
I'm removing it in master and disconnecting it from the corresponding event in 3.2, please let me know if you notice any regressions due to this. TIA!

DietmarSchwertberger:
I did a quick tests and did not see an issue.
Please also apply the original PR to the 3.2 branch for wxPython. Thanks.

vadz:
> I did a quick tests and did not see an issue.

Thanks!

> Please also apply the original PR to the 3.2 branch for wxPython. Thanks.

It's part of #24262 and will be merged soon.

vadz:
This actually was already fixed in master, so closing.

PR #24252, opened 2024-01-26
Fix crash due to using wxLog during shutdown in a better way

vadz:
This commit reverts 055c4cbed5 (Fix crash on shutdown if wxConfig couldn't be saved, 2024-01-04) and implements an alternative solution mentioned in that commit log message by setting wxTheApp to null pointer before calling CleanUp().

This ensures that wxLog creates wxLogOutputBest if necessary instead of recreating wxLogGui which can't be safely used once the cleanup starts: under MSW the dialog used by it can't be shown once the GUI objects are destroyed, so logging from CleanUp() still resulted in crashes under this platform, even after the previous fix.

This change might affect some existing code using wxTheApp during shutdown but it seems like the only really safe thing to do.

### DIFF
--- a/src/common/init.cpp
+++ b/src/common/init.cpp
@@ -488,16 +488,10 @@ static void DoCommonPostCleanup()
 
 void wxEntryCleanup()
 {
-    // delete the application object
-    if ( wxTheApp )
-    {
-        wxTheApp->CleanUp();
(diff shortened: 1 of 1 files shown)


vadz:
Merged by the commit above.

Issue #24254, opened 2024-01-26
Non-unique values in wxKeyCode enum


imciner2:
### Description

Inside the wxKeyCode enum, there are two names assigned to the same numerical values of 8, 9, 13 and 27, specifically the `WXK_BACK`, `WXK_TAB`, `WXK_RETURN` and `WXK_ESCAPE` values overlap with various `WXK_CONTROL_*` enum entries.

We noticed this because `WXK_CONTROL_I` is overlapping with `WXK_TAB`, so we see an overlap between the two: https://gitlab.com/kicad/code/kicad/-/issues/12992. Specifically, when we process a key event we call `GetKeyCode()` and `IsKeyInCategory( WXK_CATEGORY_TAB )`. For the case of a user pressing `Ctrl + I`, the key code returned is the numerical value `9`, which maps to both `WXK_TAB` and `WXK_CONTROL_I`, and the second function returns `true` because it checks if the value of `GetKeyCode()` is `WXK_TAB` to match the category. This means the key was identified as a tab key press instead of a letter key press.

This can be seen by examing the defs header file: https://github.com/wxWidgets/wxWidgets/blob/ac3cea5bcd5a3fc3d52648efb4dceacbcc9d0698/include/wx/defs.h#L2168-L2202

Unfortunately, I don't know if this can be fixed in any meaningful way :(
Because changing the enum values is probably going to be breaking (and is probably silently breaking).

### Platform and version information

- wxWidgets version  you use: Master (and 3.2)
- wxWidgets port  you use: wxGTK
- OS  and its version:
  
  + GTK version: 
  + Which GDK  backend is used:
  + Desktop environment :
  + Current theme:

oneeyeman1:
@imciner2 ,
Those 2 are interchangeable.
If you press Ctrl+I and check whether the Ctrl key is pressed - what do you get?

Thank you.

vadz:
This is indeed intentional because `Ctrl-I` is the same thing as TAB, traditionally.

I think this should be easy to fix at KiCad level by just performing this check in `KEY_DOWN` event instead of the `CHAR` one, as no translation is done for the former.

But we can also fix this in `IsKeyInCategory()` itself, albeit in a rather ugly, because it has to be platform-specific, way, see #24258. AFAICS this should fix your problem and is arguably the more correct thing to do, but I'd still recommend not using `wxEVT_CHAR` if you don't want these translations to happen.

imciner2:
[...]

PR #24258, opened 2024-01-27
Exclude `Ctrl+I` from "TAB" key category and `Ctrl+H` from the "cut" one

vadz:
See #24254.


imciner2:
@vadz was this backported to the 3.2 branch, or is it too risky to do that?

vadz:
I think we can backport it, I don't see why would anybody rely on `IsKeyInCategory(WXK_CATEGORY_TAB)` returning true for `Ctrl+I` (I hope I won't regret saying this...).

PR #24260, opened 2024-01-27
Image crash

Randalphwa:
This PR will avoid a crash if `wxImage::InitAlpha()` is called on an empty image (previously it would de-reference a null pointer). 

I also changed MSW `wxImageList::GetImageListBitmaps()` to return immediately if the bitmap is invalid rather than calling `wxImage::InitAlpha()` on it. Otherwise, the function calls `InitAlpha()`, assumes an alpha channel exists and attempts to de-reference a pointer to that channel.

### DIFF
--- a/src/common/image.cpp
+++ b/src/common/image.cpp
@@ -2185,8 +2185,7 @@ unsigned char *wxImage::GetAlpha() const
 
 void wxImage::InitAlpha()
 {
-    if (!IsOk())
-        return;
+    wxCHECK_RET( IsOk(), wxT("invalid image") );
 
     wxCHECK_RET( !HasAlpha(), wxT("image already has an alpha channel") );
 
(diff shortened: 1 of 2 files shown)


src/common/image.cpp
```diff
@@ -2185,6 +2185,9 @@ unsigned char *wxImage::GetAlpha() const
 
 void wxImage::InitAlpha()
 {
+    if (!IsOk())
```
PBfordev, line 2188: As calling basically any method on an invalid `wxImage` is a programmer mistake, I wonder if `wxCHECK()` would not be better here.

Randalphwa: That makes sense -- updated accordingly, copying the same check from the SetAlpha() function above this one.

vadz:
Thanks, I'm merging this in master and also cherry-picking into 3.2 as I think it's not a compatibility break to assert instead of crashing (i.e. it doesn't make things worse).

PR #24261, opened 2024-01-28
Modernize ribbon code

Blake-Madden:
Remove wxT(), add pointer checks, compare against nullptr properly, init pointer members to nullptr, replace __WXMAC__ defines, add explicit, replace macros with lambdas, replace IsEmtpy() with empty, add wxFALLTHROUGHs, homogenize for loop incrementing, use wxNOT_FOUND instead of -1.


include/wx/ribbon/bar.h
```diff
@@ -68,7 +68,7 @@ class WXDLLIMPEXP_RIBBON wxRibbonBarEvent : public wxNotifyEvent
     void SetPage(wxRibbonPage* page) {m_page = page;}
 
 protected:
-    wxRibbonPage* m_page;
+    wxRibbonPage* m_page{ nullptr };
```
vadz, line 71: We probably need to standardize on whether we use `{initializer}` or `= initializer`.

There is no difference between them for primitive types, including pointers, but the former does have the advantage of working better for classes. Subjectively I find the latter more readable, however.

Any thoughts/opinions/preferences?

src/ribbon/art_aui.cpp
```diff
@@ -110,16 +110,19 @@ wxColour wxRibbonAUIArtProvider::GetColour(int id) const
     switch(id)
     {
     case wxRIBBON_ART_PAGE_BACKGROUND_COLOUR:
+        wxFALLTHROUGH;
```
vadz, line 113: Do we really need fall through between consecutive cases? I thought none of the compilers warned about this, do they? If not, I'd rather not add this.

src/ribbon/art_aui.cpp
```diff
@@ -233,34 +247,38 @@ void wxRibbonAUIArtProvider::SetColourScheme(
     // TODO: Remove next line once this provider stops piggybacking MSW
     wxRibbonMSWArtProvider::SetColourScheme(primary, secondary, tertiary);
 
-#define LikePrimary(luminance) \
-    wxRibbonShiftLuminance(primary_hsl, luminance ## f).ToRGB()
-#define LikeSecondary(luminance) \
-    wxRibbonShiftLuminance(secondary_hsl, luminance ## f).ToRGB()
+    const auto LikePrimary = [&primary_hsl](const float luminance)
```
vadz, line 250: We don't use toplevel const in functions declarations, i.e. this should be
```suggestion
    const auto LikePrimary = [&primary_hsl](float luminance)
```

I also wonder if we really need to pass the HSL colour by reference, I'd just copy it for simplicity. But this is even more minor, of course.

src/ribbon/art_aui.cpp
```diff
@@ -233,34 +247,38 @@ void wxRibbonAUIArtProvider::SetColourScheme(
...
+        {
+            return wxRibbonShiftLuminance(primary_hsl, luminance).ToRGB();
+        };
+    const auto LikeSecondary = [&secondary_hsl](const float luminance)
```
vadz, line 254:
```suggestion
    const auto LikeSecondary = [&secondary_hsl](float luminance)
```

src/ribbon/bar.cpp
```diff
@@ -140,7 +140,7 @@ void wxRibbonBar::ShowPanels(bool show)
 void wxRibbonBar::SetWindowStyleFlag(long style)
 {
     m_flags = style;
-    if(m_art)
+    if ( m_art != nullptr )
```
vadz, line 143: I don't think we want to change all the existing `if ( ptr )` to use `if ( ptr != nullptr )` so I'd leave them be.

src/ribbon/panel.cpp
```diff
@@ -711,7 +711,9 @@ bool wxRibbonPanel::Realize()
                 scale = 2.0;
 
             wxImage img(m_minimised_icon.ConvertToImage());
-            img.Rescale(int(scale * bitmap_size.GetWidth()), int(scale * bitmap_size.GetHeight()), wxIMAGE_QUALITY_HIGH);
+            img.Rescale(static_cast<int>(scale * bitmap_size.GetWidth()),
```
vadz, line 714: I wonder if this shouldn't be `wxRound()`...

vadz (commented):
Thanks for the cleanup!

To be honest I'm not really thrilled about the change of all pointer tests and adding `f` to floating point constants (does this really change anything? It seems hard to believe that the compiler would pass a double constant to a function taking float differently than a float one), but we can merge this if you really want to.

The changes getting rid of macros in favour of lambdas and initializing members in their declaration are definitely welcome, but I'd like to standardize on the style of the latter before merging, to avoid having to change it back later.

Blake-Madden:
>  I'm not really thrilled about the change of all pointer tests

I was just trying to be consistent; there are a lot of places in the code that compares against `nullptr`.

> adding `f` to floating point constants (does this really change anything? It seems hard to believe that the compiler would pass a double constant to a function taking float differently than a float one), but we can merge this if you really want to.

I didn't want to do that, but MSVC2022 was issuing "double to float narrowing" warnings until I added the `f`s.

src/ribbon/art_aui.cpp
Blake-Madden: Yeah, I haven't seen warnings; I sort of assume that future compilers will warn about this as they become more pedantic. But I can certainly roll these back if you prefer.

src/ribbon/art_aui.cpp
Blake-Madden: Got it, will make those changes.

src/ribbon/art_aui.cpp
Blake-Madden: Will make those changes.

src/ribbon/art_aui.cpp
Blake-Madden: OK, I'm removing those.

src/ribbon/panel.cpp
Blake-Madden: Will make the change.

src/ribbon/bar.cpp
Blake-Madden: OK, rolling those back...

include/wx/ribbon/bar.h
Blake-Madden: I prefer `{}` myself because it's more "modern" looking, but `=` is totally OK.  I happy to use either.

With `{}`, I've gotten into the habit of using spaces like `{ nullptr }` to make it easier to read.  Just my style, I'm happy to use a different style.

vadz: I find `=` more readable, e.g. I agree that

```c++
T* m_ptr{nullptr};
```

is not very readable without spaces, but 

```c++
T* m_ptr = nullptr;
```

is IMO just fine on its own.

And I think in modern C++, i.e. with guaranteed copy elision, there are few drawbacks to using it compared to `{}`. The latter does forbid narrowing conversions, but wx code is supposed to be compiled with warning options catching the most egregious of them and you're unlikely to have really unwanted conversions when writing simple member initializers IMHO.

So if there are no big objections to it, I'd rather use `=`. TIA!

vadz:
> > I'm not really thrilled about the change of all pointer tests
> 
> I was just trying to be consistent; there are a lot of places in the code that compares against `nullptr`.

Yes, consistency is nice but I'm afraid this is really a place where we'd have to change roughly 50% of the code to achieve it and I don't think it's worth it.
 
> > adding `f` to floating point constants (does this really change anything? It seems hard to believe that the compiler would pass a double constant to a function taking float differently than a float one), but we can merge this if you really want to.
> 
> I didn't want to do that, but MSVC2022 was issuing "double to float narrowing" warnings until I added the `f`s.

I'd change the lambda to take `double` if really needed rather than do this.

include/wx/ribbon/bar.h
manphiz: An OT and outsider's comment: the new initializer `{}` used without any contents will emulate the default "0", "false", "nullptr" initialization for primitive types, and calls the default ctor for objects (which is by default since C++98).  So basically for most type `T`, `T obj{};` should do the reasonable default initialization.  This also matches the behavior in newer languages with default initialization like Go.  Though admittedly this is not a widely applied convention, so just my 2 cents.

vadz: Yes, this is correct but, again, `{}` loses to `= 0` from readability point of view IMO.

I also don't think we're going to switch to writing `int i{};` instead of `int i = 0;` for local variables and using the same syntax for them and member variables is also a (minor) advantage.

manphiz: Indeed, and I suspect that it's exactly because of this difference between old "=" code that prevents its wider use, unless this was done from the beginning.  Totally makes sense for existing codebase not to switch to that.

Blake-Madden: Got it, will use `=`.

Blake-Madden:
> I'd change the lambda to take `double` if really needed rather than do this.

Made the fix, removed the `f` suffixes. That should be everything now.

src/ribbon/art_aui.cpp
vadz: Sorry, it seems they didn't get removed. I guess it was unintentional/forgotten?

src/ribbon/gallery.cpp
```diff
@@ -61,8 +57,8 @@ class wxRibbonGalleryItem
     wxBitmap m_bitmap;
     wxClientDataContainer m_client_data;
     wxRect m_position;
-    int m_id;
-    bool m_is_visible;
+    int m_id{ 0 };
+    bool m_is_visible{ false };
```
vadz, line 61:
```suggestion
    int m_id = 0;
    bool m_is_visible = false ;
```

vadz (commented):
Thanks for the updates, but I'd still remove "fall through" macros between consecutive cases and there are also still a couple of extra `!= nullptr` added here and there.

src/ribbon/gallery.cpp
Blake-Madden: Fixed.

src/ribbon/art_aui.cpp
Blake-Madden: OK, fix them.

src/ribbon/page.cpp
```diff
@@ -1076,17 +1078,20 @@ bool wxRibbonPage::CollapsePanels(wxOrientation direction, int minimum_amount)
                     // come to this panel again anyway)
                     amount = 32;
                 }
-                if(direction & wxHORIZONTAL)
+                if (largest_panel_size)
```
vadz, line 1081: This change (and similar one below) doesn't seem to fall under cleanup/modernization, so if they're really needed, could you please resubmit them as a separate PR or at least a separate commit?

FWIW I think we should test for it above, instead of testing for `largest_panel`, if we really need to check it (the code is not very clear, so I'm not sure).

TIA!

vadz (commented):
Thanks, I'm finally (squash)merging this after dropping a few more whitespace-only changes and removing unnecessary, IMO, `wxFALLTHROUGH`s.

PR #24263, opened 2024-01-28
CMake improvements

MaartenBent:
Lots of small CMake changes and improvements.


build/cmake/functions.cmake
```diff
@@ -324,6 +324,11 @@ function(wx_set_target_properties target_name)
             )
     endif()
 
+    if(WIN32)
+        target_compile_definitions(${target_name} PUBLIC UNICODE)
+    endif()
+    target_compile_definitions(${target_name} PUBLIC _UNICODE)
```
vadz, line 330: Do we really need to predefine this under non-Win32 systems? I don't think it's used there, am I missing something?

samples/minimal/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 #
 
 # Declare the minimum required CMake version
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.5)
```
vadz, line 17: I think this needs to be documented somewhere.

samples/minimal/CMakeLists.txt
```diff
@@ -55,14 +50,20 @@ endif()
 
 # Define the build target for the executable
 add_executable(${PROJECT_NAME} WIN32 MACOSX_BUNDLE ${SRC_FILES})
-# Link required libraries to the executable
-target_link_libraries(${PROJECT_NAME} ${wxWidgets_LIBRARIES})
+# Link executable to the wxWidgets target
+target_link_libraries(${PROJECT_NAME} wxWidgets::wxWidgets)
```
vadz, line 54: I wonder if this target shouldn't be called `wxWidgets::all` or something like for clarity? Or is this name fixed?

vadz (approved):
Thanks for the fixes!

samples/minimal/CMakeLists.txt
MaartenBent: Fixed, I want to use the same name as https://cmake.org/cmake/help/latest/module/FindwxWidgets.html#imported-targets

build/cmake/functions.cmake
MaartenBent: It will be used on any platform. When you link with another target in CMake, your target will inherit the public headers, libraries, and defines.

samples/minimal/CMakeLists.txt
MaartenBent: I'll see if it is documented anywhere now. It is mostly to get rid of the annoying warning everything you configure.
```cmake
Compatibility with CMake < 3.5 will be removed from a future version of CMake.
```

Another way might be:
```cmake
cmake_minimum_required(VERSION 2.8.12...3.28)
```

Though I added `IN_LIST` to the config file, which raises the minimum version to 3.3.

build/cmake/functions.cmake
vadz: Yes, sorry, I understand this, what I don't understand is why should the projects using wx get this define under non-Win32 platforms when it's only used by (MSVC) CRT under Win32 AFAIK.

samples/minimal/CMakeLists.txt
vadz: Just to be clear: I don't mind increasing the version, I do believe that all platforms supported by current master should have (or have no trouble getting) CMake 3, but I just think it would be nice to document this requirement somewhere, probably in CMake build instructions.

build/cmake/functions.cmake
vadz: I'm still not sure about this but if you're confident this is correct, we could merge this, of course. Please let me know if we should?

MaartenBent: No, you are right. I confused the msvc c- runtime with the c++ library. I'll make it win32 only.

vadz (approved):

PR #24264, opened 2024-01-28
wxGenericValidator:  support wxRadioButton group

wsu-cb:
Enable wxGenericValidator(int*) to work with wxRadioButton groups when applied to the first element of the group.  The value will be the index of the selected button within the group.

### DIFF
--- a/src/common/valgen.cpp
+++ b/src/common/valgen.cpp
@@ -172,6 +172,28 @@ bool wxGenericValidator::TransferToWindow()
             pControl->SetValue(*m_pBool) ;
             return true;
         }
+        else if (m_pInt)
+        {
+            wxRadioButton* selected = nullptr;
+            for (int i = 0 ;
+                pControl ;
+                ++i, pControl = pControl != pControl->GetLastInGroup() ? pControl->GetNextInGroup() : nullptr)
+            {
+                wxCHECK_MSG(!pControl->GetValidator() ||
+                            pControl == pControl->GetFirstInGroup(),
+                        false,
+                        "wxRadioButton group validator must be on first item in group");
+                if (i == *m_pInt)
+                {
+                    selected = pControl;
+                }
+            }
+            if (selected)
+            {
+                selected->SetValue(true);
+                return true;
+            }
+        }
     } else
 #endif
 
@@ -497,6 +519,25 @@ bool wxGenericValidator::TransferFromWindow()
         if (m_pBool)
         {
             *m_pBool = pControl->GetValue() ;
+            return true;
+        }
+        else if (m_pInt)
+        {
+            *m_pInt = -1;
+            for (int i = 0 ;
+                pControl ;
+                ++i, pControl = pControl != pControl->GetLastInGroup() ? pControl->GetNextInGroup() : nullptr)
+            {
+                wxCHECK_MSG(!pControl->GetValidator() ||
+                            pControl == pControl->GetFirstInGroup(),
+                        false,
+                        "wxRadioButton group validator must be on first item in group");
+                if (pControl->GetValue())
+                {
+                    *m_pInt = i;
+                }
+            }
+            wxASSERT(*m_pInt >= 0 || !"no selected radio button!");
             return true;
         }
     } else
--- a/src/common/valgen.cpp
+++ b/src/common/valgen.cpp
@@ -174,24 +174,17 @@ bool wxGenericValidator::TransferToWindow()
         }
         else if (m_pInt)
         {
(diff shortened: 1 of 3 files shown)


wsu-cb:
I do NOT need this backported to wx 3.2.

src/common/valgen.cpp
```diff
@@ -499,6 +521,25 @@ bool wxGenericValidator::TransferFromWindow()
...
+                wxCHECK_MSG(!pControl->GetValidator() ||
+                            pControl == pControl->GetFirstInGroup(),
+                        false,
+                        "wxRadioButton group validator must be on first item in group");
+                if (pControl->GetValue())
+                {
+                    *m_pInt = i;
+                }
+            }
+            wxASSERT(*m_pInt >= 0 || !"no selected radio button!");
```
vadz, line 540: We have a more readable form for this:
```suggestion
            wxASSERT_MSG( *m_pInt >= 0, "no selected radio button!" );
```

src/common/valgen.cpp
```diff
@@ -499,6 +521,25 @@ bool wxGenericValidator::TransferFromWindow()
             *m_pBool = pControl->GetValue() ;
             return true;
         }
+        else if (m_pInt)
+        {
+            *m_pInt = -1;
```
vadz, line 526: Very minor, but I think this might be a tad more readable:
```suggestion
            *m_pInt = wxNOT_FOUND;
```

src/common/valgen.cpp
```diff
@@ -172,6 +172,28 @@ bool wxGenericValidator::TransferToWindow()
             pControl->SetValue(*m_pBool) ;
             return true;
         }
+        else if (m_pInt)
+        {
+            wxRadioButton* selected = nullptr;
```
vadz, line 177: I don't think we need this, once we find the selected button we can just call `SetValue()` on it directly and either break or return, there is no need to keep iterating further.

src/common/valgen.cpp
```diff
@@ -172,6 +172,28 @@ bool wxGenericValidator::TransferToWindow()
...
+            for (int i = 0 ;
+                pControl ;
+                ++i, pControl = pControl != pControl->GetLastInGroup() ? pControl->GetNextInGroup() : nullptr)
```
vadz, line 180: Sorry, I find this loop pretty difficult to read. I see a few different ways to simplify it, but I'd probably go with

```suggestion
            for ( int i = 0; ; ++i )
```

here and then add

```c++
if ( pControl == pControl->GetLastInGroup() )
    break;

pControl = pControl->GetNextInGroup()
```

at the end of the loop.

src/common/valgen.cpp
```diff
@@ -499,6 +521,25 @@ bool wxGenericValidator::TransferFromWindow()
             *m_pBool = pControl->GetValue() ;
             return true;
         }
+        else if (m_pInt)
+        {
+            *m_pInt = -1;
+            for (int i = 0 ;
+                pControl ;
+                ++i, pControl = pControl != pControl->GetLastInGroup() ? pControl->GetNextInGroup() : nullptr)
```
vadz, line 529: Same remark as above.

vadz (commented):
Thanks, globally this definitely looks like it ought to work, but I'd like to simplify things a bit as it looks like this code could be clearer. Could you please try to do it?

Also, it finally seems like this would apply cleanly to 3.2 too, so I'll cherry pick it there even after seeing your comment about it -- like this I won't have to change the docs, which speak of this feature being added in 3.2.5.

src/common/valgen.cpp
wsu-cb: I intentionally completed the loop so that I could have the wxCHECK_MSG() verify that the rest of the buttons in the group had no validator installed.  However, if you think that checking this isn't worth the extra bother, I can indeed simplify this code.

vadz: Ah, I had missed that, sorry. I guess we could keep if you added a comment explaining what's going but it doesn't seem to be that important.

In fact, thinking more about this, such a check would be more useful in `SetValidator()` as it would point directly to the place in the program where the mistake is being made instead of triggering at a much later moment and giving you no real idea where the problem is (how would you find the offending radio button?). So what about overriding `SetWindow()` in `wxGenericValidator` and checking that the window is not the non-first radio button in the group there?

wsu-cb:
I have pushed a `commit --fixup` that I hope will address your suggestions.

vadz:
Very nice, thanks! Will merge soon with some minor changes.

PR #24265, opened 2024-01-29
Set rpath for wx libraries themselves too

vadz:


vadz:
I'll merge this soon if there are no objections.

Issue #23109, opened 2023-01-02
wxIPV4address::Hostname(unsigned long) is too slow


SMMB-DEV:
### Description
I have a program with an FTP connection that checks certain folders for new files periodically. (**Not running on the main thread!**)

```
wxFTP ftp = new wxFTP;
//...
while (...)
{
    wxString dir = ...;
    if (ftp->ChDir(dir))
    {
        wxArrayString files;
        if (ftp->GetFilesList(files))
        {
            //...
        }
    }
}
```

I noticed that wxFTP::GetFilesList takes too long most of the time even with an empty directory (~1700ms - min ~100ms). Finally I narrowed it down to this section in the source code:

#### src/common/ftp.cpp:
```
wxSocketBase *wxFTP::GetPassivePort()
{
    //...

    // get the port number and address
    int a[6];
    wxString straddr(m_lastResult, addrStart + 1, addrEnd - (addrStart + 1));
    wxSscanf(straddr, wxT("%d,%d,%d,%d,%d,%d"),
             &a[2],&a[3],&a[4],&a[5],&a[0],&a[1]);

    wxUint32 hostaddr = (wxUint16)a[2] << 24 |
                        (wxUint16)a[3] << 16 |
                        (wxUint16)a[4] << 8 |
                        a[5];
    wxUint16 port = (wxUint16)(a[0] << 8 | a[1]);

    wxIPV4address addr;
    addr.Hostname(hostaddr);    //<-- this is taking too long
    addr.Service(port);

    //...
}
```

which is using `wxIPV4address::Hostname(unsigned long addr)` but if I replace it with `wxIPaddress::Hostname(const wxString& name)` it works fine and sets the hostname almost immediately.
```
addr.Hostname(wxString::Format("%hhu.%hhu.%hhu.%hhu", a[2], a[3], a[4], a[5]));
```

### Platform and version information
- wxWidgets version 3.2.1 (wxMSW)
- Windows 10 Version 10.0.19044 Build 19044

SpareSimian:
First, I suggest reducing the issue title to "wxIPV4address(unsigned long addr) is slow".

The code that's slow is doing some kind of address-to-name lookup using Windows gethostbyaddr, possibly DNS. Your patch eliminates this lookup and tells the FTP code to use use the address without a lookup.

What is the network environment you're using this with? How is the address being converted to a hostname? A local hosts file? DNS? 

[...]

PR #24269, opened 2024-01-30
Unnecessary hostname lookup in wxIPV4address::Hostname(unsigned long addr)

SMMB-DEV:
Removed unnecessary lookup as discussed in issue #23109.

### DIFF
--- a/src/common/sckaddr.cpp
+++ b/src/common/sckaddr.cpp
@@ -849,7 +849,14 @@ bool wxIPV4address::Hostname(unsigned long addr)
         return false;
     }
 
-    m_origHostname = Hostname();
+    const wxUint8
+        addr1 = (addr >> 24) & 0xFF,
+        addr2 = (addr >> 16) & 0xFF,
+        addr3 = (addr >>  8) & 0xFF,
+        addr4 = (addr >>  0) & 0xFF;
+
+    m_origHostname = wxString::Format("%hhu.%hhu.%hhu.%hhu", addr1, addr2, addr3, addr4);
+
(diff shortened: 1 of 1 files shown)


src/common/sckaddr.cpp
```diff
@@ -849,7 +849,14 @@ bool wxIPV4address::Hostname(unsigned long addr)
...
     }
 
-    m_origHostname = Hostname();
+    const wxUint8
+        addr1 = (addr >> 24) & 0xFF,
+        addr2 = (addr >> 16) & 0xFF,
+        addr3 = (addr >>  8) & 0xFF,
+        addr4 = (addr >>  0) & 0xFF;
+
+    m_origHostname = wxString::Format("%hhu.%hhu.%hhu.%hhu", addr1, addr2, addr3, addr4);
```
vadz, line 858: To be honest I'm not sure what's the advantage of using `wxUint8` above and `%hhu` here compared to just using `unsigned` and `%u`, is there something I'm missing?

vadz (commented):
Thanks, I'll merge this, but if you prefer/agree with simplifying things by using `unsigned`/`%u` instead, please let me know.

Another welcome improvement would be to document this `Hostname()` overload in `wxIPV4address` description in `interface/wx/socket.h` and mention that the string returned by `Hostname()` is in this format there.

src/common/sckaddr.cpp
SMMB-DEV: No, there's no advantage; just wanted to be explicit.

PR #24270, opened 2024-01-30
CI: update actions/checkout to v4

swt2c:
GitHub warns that actions/checkout@v3 is using Node.js 16 which is deprecated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -186,7 +186,7 @@ jobs:
           esac
 
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           submodules: 'recursive'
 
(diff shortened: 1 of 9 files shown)


swt2c:
Sigh, looks like Ubuntu 18.04's glibc is too old...

vadz:
I'm not sure if we can drop support for 18.04 already and, in any case, we need it in 3.2 branch so it would be nice to find some solution for it. Any ideas would be welcome!

vadz:
Returning to this, I think the only solution is to use a different checkout step for this platform, i.e. use an `if:` on the existing step and add a manual `git clone` with the inverted `if:` condition.

vadz:
Closing because this is now done in #24377 but Ubuntu 18.04 is still a problem. It would be nice to keep support for it, at least in 3.2, but I think we're going to need to perform the checkout ourselves there, at least I don't see any other solution.

MaartenBent:
fyi, `actions/checkout@v3` is also used in `docs_update.yml` and `make_release.yml` (2x).

PR #24272, opened 2024-01-30
Really fix logging during app cleanup

vadz:
This is yet another commit trying to improve things with logging on app shutdown -- without introducing bad regressions such as the one fixed by the first commit in this PR.

I'll merge this soon because this commit really needs to be merged a.s.a.p. to avoid crashes on exit.

### DIFF
--- a/include/wx/x11/app.h
+++ b/include/wx/x11/app.h
@@ -41,8 +41,6 @@ public:
     // override base class (pure) virtuals
     // -----------------------------------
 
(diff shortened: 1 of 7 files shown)


mcorino:
This fixes all issues in both the wxRuby3 regression tests as well as the changed standard minimal sample with WXGTK.
After you merge I'll restart the wxRuby3 master builds to check that WXMSW is fixed too.

PR #24276, opened 2024-02-02
Fix WebView tests with WebKitGTK 2.43+

swt2c:
Plus a fix for the WebView sample I ran into while troubleshooting this.

### DIFF
--- a/samples/webview/webview.cpp
+++ b/samples/webview/webview.cpp
@@ -1215,7 +1215,10 @@ void WebFrame::OnToolsClicked(wxCommandEvent& WXUNUSED(evt))
     unsigned int i;
     for(i = 0; i < back.size(); i++)
     {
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for fixing this!

I couldn't prevent myself from doing some minor cleanup here, please let me know if you see anything wrong with it, otherwise I'll merge it soon.

swt2c:
Your updates look good.  I'm assuming they are not backportable though?

vadz:
No, they aren't, I hadn't realized you wanted to apply this to 3.2 too.

I'll merge this in master and will just apply your original changes to 3.2 then.

swt2c:
The first commit definitely should go to 3.2 as I'm running into that problem on both Fedora & Debian.  The second one is a nice-to-have as I'll be less annoyed the next time I go to play with the webview sample.  :-)

Issue #24283, opened 2024-02-04
wxButton::GetDefaultSize no longer holds true on modern windows with hidpi


marekr:
We were using ``wxButton::GetDefaultSize` in KiCad to get button sizes to do a custom bitmap button because there was something internally wrong with wxBitmapButton.

Turns out `wxButton::GetDefaultSize` does not hold true anymore on Windows and I say that because it has this inside

```
  // The size of a standard button in the dialog units is 50x14,
        // translate this to pixels.
        //
        // Windows' computes dialog units using average character width over
        // upper- and lower-case ASCII alphabet and not using the average
        // character width metadata stored in the font; see
        // http://support.microsoft.com/default.aspx/kb/145994 for detailed
        // discussion.

        s_sizeBtn.SetAtNewDPI(wxRescaleCoord(50, 14).From(4, 8).To(base));
```

On a standard 1080p screen at 100% scaling, it returns 26 px height instead of 23px.

On a 4K, 250% scaling screen, it returns 86px instead of 65px for height.

I switched kicad to creating a new button, parenting it to the window, setting a single letter, and then destroying that button away right after I get it's dimensions.

MaartenBent:
That comment has been there since the dawn of wxWidgets. The DPI stuff has been added, but it is still the same multiplications.
What caused it to change, a wxWidgets update?

Or maybe you started using a manifest in your application? Then it uses the normal GUI font instead of the ugly system font to determine the default character height.
(for example, with `wxSystemSettings::GetFont(wxSYS_SYSTEM_FONT)` in the code above that code snippet, it still seems to return `100x23`)

marekr:
> That comment has been there since the dawn of wxWidgets. The DPI stuff has been added, but it is still the same multiplications. What caused it to change, a wxWidgets update?
> 
> Or maybe you started using a manifest in your application? Then it uses the normal GUI font instead of the ugly system font to determine the default character height. (for example, with `wxSystemSettings::GetFont(wxSYS_SYSTEM_FONT)` in the code above that code snippet, it still seems to return `100x23`)

We've had a manifest in KiCad for probably a decade hah.

vadz:
We had already fixed this in c6d2f6d9fe (Really fix the standard button size in high DPI under MSW, 2019-10-19) when creating buttons, see #18528, but I forgot to fix it in `GetDefaultSize()` -- done now in #24285, please let me know if you still see any problems with it.

[...]

PR #24285, opened 2024-02-05
Fix `wxButton::GetDefaultSize()` in wxMSW

vadz:
See #24283.

### DIFF
--- a/src/msw/button.cpp
+++ b/src/msw/button.cpp
@@ -163,30 +163,19 @@ wxSize wxButtonBase::GetDefaultSize(wxWindow* win)
 
     if ( s_sizeBtn.HasChanged(win) )
     {
-        wxSize base;
-        if ( win )
-        {
-            wxClientDC dc(win);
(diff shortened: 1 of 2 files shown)


src/msw/button.cpp
```diff
@@ -163,30 +163,19 @@ wxSize wxButtonBase::GetDefaultSize(wxWindow* win)
...
-        // character width metadata stored in the font; see
-        // http://support.microsoft.com/default.aspx/kb/145994 for detailed
-        // discussion.
-
-        s_sizeBtn.SetAtNewDPI(wxRescaleCoord(50, 14).From(4, 8).To(base));
+        // Moreover, it looks like the extra 2px borders around the visible
+        // part of the button are not scaled correctly in higher than normal
+        // DPI, so add them without scaling.
+        s_sizeBtn.SetAtNewDPI(
+            wxWindow::FromDIP(wxSize(73, 21), win) + wxSize(2, 2)
```
marekr, line 177: The only thing I can remark is that 2px error isn't constant. It's more like 

`2 + (scale%/100)`

https://github.com/wxWidgets/wxWidgets/issues/24283#issuecomment-1928558875

PR #2340, opened 2021-04-22
Various list control fixes

imciner2:
This PR contains a few different fixes for the `wxListCtrl` widget including:

* Fix centering of images in wxListCtrl items when selected in icon view (Trac ticket: https://trac.wxwidgets.org/ticket/11331)

Before:
image

After:
image
image

* Introduce ABI shims for the column ordering functions on all ports.

This commit adds pure virtual methods to the base class for these functions and then adds stub methods to the generic port that serve as placeholders. The wxQT port already had stub methods for these, so those are just marked override now.

* Remove the documentation on the old native wxMAC `wxListCtrl` control for report mode.

The native control was using the old Carbon library and was removed in https://github.com/wxWidgets/wxWidgets/commit/5ba67c67e47d069f65d648ab16dfc505c5400bfc, and the only remaining references to it was in the documentation and the sample program. This removes the remaining documentation references and the sample program parts.

PR #24292, opened 2024-02-08
Fix centering of item label in generic `wxListCtrl`

taler21:
With wxWidgets-3.2.4, in a generic `wxListCtrl` the label is not centered on the icon as intended if the icon is wider than the label.

wxListCtrl-master

This PR fixes this issue.

wxListCtrl-PR

### DIFF
--- a/include/wx/generic/private/listctrl.h
+++ b/include/wx/generic/private/listctrl.h
@@ -190,10 +190,11 @@ public:
             wxASSERT_MSG( m_rectAll.width <= w,
                             wxT("width can only be increased") );
 
+            int delta = (w - m_rectAll.width) / 2;
             m_rectAll.width = w;
-            m_rectLabel.x += (w - m_rectLabel.width) / 2;
-            m_rectIcon.x += (w - m_rectIcon.width) / 2;
-            m_rectHighlight.x += (w - m_rectHighlight.width) / 2;
+            m_rectLabel.x += delta;
+            m_rectIcon.x += delta;
+            m_rectHighlight.x += delta;
         }
     };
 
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for fixing this!

Just for the record, has this always been broken or is it a regression?

taler21:
It is a regression introduced in wxWidgets-3.1.6 by commit e0ef3830c13f0d7d7b871c27348de6c042e520d9 (`Fix centering of images in wxListCtrl items when selected, 2021-04-27`).

vadz:
Thanks for checking this!

@imciner2 could you please confirm that this doesn't reintroduce the original problem you fixed?

imciner2:
Yes, this doesn't appear to reintroduce the regression I was trying to fix in https://github.com/wxWidgets/wxWidgets/pull/2340

Comparing the 0th icon to the one from my earlier PR, this still looks to match the intended result.

image

Perhaps it would be good to modify the sample to include some of these very short strings, so this can be obvious when just viewing the sample in the future?
For instance, I just did this small change to get my new image (since I find it easier to judge centering of icons on large text using the blobs instead of the checkmark):
```
diff --git a/samples/listctrl/listtest.cpp b/samples/listctrl/listtest.cpp
index e08687d867..5e610f10b9 100644
--- a/samples/listctrl/listtest.cpp
+++ b/samples/listctrl/listtest.cpp
@@ -650,6 +650,8 @@ void MyFrame::InitWithIconItems(bool withText, bool sameIcon)
             wxString label;
             if ( !(i % 5) )
                 label.Printf("Longer label %d", i);
+            else if ( !(i % 4) )
+                label.Printf("A %d", i);
             else
                 label.Printf("Label %d", i);

```

vadz:
Thanks for testing so quickly!

I'm merging this now with your proposed change to the sample.

PR #24296, opened 2024-02-09
Only add padding for shown panels on ribbon

Blake-Madden:
When laying out panels on a ribbon, don't add padding for panels that won't be drawn and used in the layout calculations.


vadz:
It seems a bit strange that we still adjust the origin for hidden panels, do we really need to do this?

But let's merge it, of course, if this solves a user-visible problem, thanks!

PR #23930, opened 2023-10-02
Fix AddStdCatalog() under MSW and add functions without msg ID fallback (v2)

vadz:
Sorry, I've accidentally closed #23918 by deleting its branch and now there doesn't seem to be any way to reopen it, so creating a new one.

PR #24297, opened 2024-02-10
Restore correct translation language lookup

vslavik:
@vadz sorry this took so long, but it was for the better, I encountered additional problematic cases.

This fixes the problem in PR #23918 / #23930 discussed here: https://github.com/wxWidgets/wxWidgets/pull/23918#issuecomment-1806153817; see also commit messages here.

The lookup bug manifests itself in more situations than the above comment says, e.g.

1. msgid as not-last preference, the above-mentioned case. E.g. msgid `en`, user preference `[cs,en,fr]`, only `fr.po` available: `fr` would be used instead of `en`
2. forced use of `en_GB` translation if available any time `en` or `en_US` should be used if msgid is English (words like "catalogue" poke you in the eyes when you least expect it)
3. this is admittedly a very edge case, but it would pick typography-affecting translations like `en@boldquot` over embedded English when the preference is `en_US`

In this PR, I restore the original behavior, but preserve the split API.

This PR also fixes, in 34a90ba3afabb20283657ecf2ca09d22888b48a4, wrong behavior when `GetBest*Translation()` would, as far as I can tell, *never* return empty string. That behavior, and `GetPreferredUILanguageFallback()`, were present since my initial commit of this code way back and I can't figure out the reason for it. I *think* it is obsolete in modern wx with `wxUILocale`.

I am not sure what to do about `AddAvailableCatalog()`. This PR preserves it, but changes its signature - it simply cannot be correctly implemented without the msgidLanguage argument. My preference would be, given its recency and presumably none to extremely minimal use in the wild, to remove it altogether (keeping it as a deprecated alias for `AddCatalog()`) — IMHO it's not worth the added complexity. But this version of the PR retains it.

This should be backported to 3.2 (and the docs change assumes it will happen).


src/common/translation.cpp
```diff
@@ -168,7 +160,7 @@ wxString GetPreferredUILanguage(const wxArrayString& available)
     if (!langNoMatchRegion.empty())
         return langNoMatchRegion;
 
-    return GetPreferredUILanguageFallback(available);
+    return wxString();
```
vadz, line 163: I was always very confused by the "locale-based as a fallback" code, but as I didn't understand it, I didn't change it (or at least didn't do it intentionally) as I was too afraid to remove it and break something. If we can remove it, this is great, as it makes things simpler to understand, but I can't help wondering how could we have it for all these years without anybody knowing what it was for...

src/common/translation.cpp
```diff
@@ -1452,31 +1438,39 @@ wxString wxTranslations::GetBestTranslation(const wxString& domain,
...
-        }
-        else
-        {
-            wxLogTrace(TRACE_I18N,
-                       "using message ID language '%s' for domain '%s'", lang);
-        }
+        wxLogTrace(TRACE_I18N,
+                   "no available language for domain '%s'", domain);
+    }
+    else if ( lang == msgIdLanguage )
```
vadz, line 1450: It's not very important because it only affects the debugging message, but shouldn't we check for `lang == msgIdLanguage.BeforeFirst('_')` too here?

src/common/translation.cpp
```diff
@@ -1452,31 +1438,39 @@ wxString wxTranslations::GetBestTranslation(const wxString& domain,
...
     }
 
     return lang;
 }
 
 wxString wxTranslations::GetBestAvailableTranslation(const wxString& domain)
 {
-    const wxArrayString available(GetAvailableTranslations(domain));
+    // Determine the best language from the ones with actual translation file:
+    return DoGetBestAvailableTranslation(domain, wxString());
```
vadz, line 1462: Passing empty string here makes me wonder if we shouldn't add `msgIdLanguage` parameter to `GetBestAvailableTranslation()` too?

vadz (commented):
Thanks, at first glance (well, ok, who am I kidding, I must have spent half an hour on this), this looks fine and I do like the idea of passing the available message into `DoGetBestAvailableTranslation()` instead of trying to apply it after calling it.

But, of course, I thought this code was fine before too... I'd really like to find a way to have some tests for it, i.e. assert that, for the given tuple `(msgid, user-preferred-languages-array, actually-available-catalogs)` we get the expected result. Could we refactor this code somehow or maybe add a private function that could be called from the tests to verify that this works as expected? IMO we really need this because this code has been proven to be brittle.

Another thing I'd like to change is to use some

```c++
enum class Translations
{
  NotNeeded = -1,
  NotFound = 0,
  Found = 1
};
```

instead of `bool&`, but this is trivial and I could do it myself too.

Finding a way to test this is much less trivial but would be IMO be very, very useful. Please let me know if you can think of something reasonable here.

TIA!

src/common/translation.cpp
utelle: IMHO passing an empty string is wrong here, because the message language used in the source code **is** _available_. That is, `msgIdLanguage` should be passed.

The only somewhat problematic use case is when the `msgIdLanguage` of the given domain is **not** _English_ and the standard wxWidgets catalog is **not** available in that language. Then you will have a mix of translation languages.

interface/wx/translation.h
```diff
@@ -167,9 +177,10 @@ class wxTranslations
             selected or system-default languages, but is not necessarily an
             error if no translations are needed in the first place.
 
-        @since 3.3.0
+        @since 3.2.5
```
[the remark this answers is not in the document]
vslavik: Sorry, I should have commented on this. I know (the 3.2 branch documents it correctly), but I figured it's a useful white lie. It is *technically* true and because the pre-3.2.5 version was broken by design, and x.y.(z+1) updates are not problematic, it IMHO makes sense to just document the corrected version...

include/wx/translation.h
```diff
@@ -193,6 +193,8 @@ class WXDLLIMPEXP_BASE wxTranslations
     static const wxString& GetUntranslatedString(const wxString& str);
 
 private:
+    bool DoAddCatalog(const wxString& domain, wxLanguage msgIdLanguage, bool& isUsedUntranslated);
```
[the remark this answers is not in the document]
vslavik: Done in 039360a0 (as a fixup for clarity; I'll rebase and force-push the final version).

src/common/translation.cpp
vslavik: Reconstructing from git blame, IIRC the original could get a list of languages only on some platforms (OS X?) and maybe it could fail, hence fallback. The "fallback" implementation was definitely used as the only one on UNIX originally.

The wrong use the fallback, i.e. as the last thing, was a bug I implemented at the beginning, but it didn't demonstrate itself until these API changes. Possibly it was that even older code just returned locale-based single language, so falling through to older logic made sense. In any case, now that we get a list of languages on every platform, there's no point.

src/common/translation.cpp
vslavik: Thanks, fixed in 039360a0

src/common/translation.cpp
vslavik: @utelle I agree, and we had this conversation with @vadz in the linked PR before. This part of *this* PR is attempting to preserve the unfortunate `GetBestAvailableTranslation ()` API. If it didn't become part of 3.2 already, I'd argue for removing this function altogether, as it has little if any value — even `GetStdCatalog()`, which motivated all this, can be implemented w/o it just using `GetAvailableTranslations()`.

@vadz What would be the semantics - and usefulness be - then? How would (or even *could*) differ from `GetBestTranslation()` then?

vadz: Sorry, I got confused (not for the first time...), please disregard my original comment: passing `msgIdLanguage` to this function doesn't make sense because it never uses, or even considers, the original message strings anyhow.

I agree that I wouldn't add `GetBestAvailableTranslation()` today, as the problem using it was supposed to fix is now fixed in a better way. But it's still potentially useful if you really want to find all the catalog files available on the system, for whatever reason, and removing it now would create more problems than it solves.

So, finally, the only thing I'd change would be this:

```suggestion
    // As this function never considers the language of the original messages as being
    // available, pass empty string as message ID language to the helper function.
    return DoGetBestAvailableTranslation(domain, wxString());
```

src/common/translation.cpp
vadz: Thanks for digging into this!

vadz (approved):
Thanks again, I would still love to have the tests, but this is clearly an improvement (both from the correctness and simplicity/clarity points of view), so we should merge it, even if there is still no way to test it.

src/common/translation.cpp
vslavik: > but it's still potentially useful if you really want to find all the catalog files available on the system

`GetAvailableTranslations()` is better for that...

> for whatever reason, and removing it now would create more problems than it solves.

No argument there :/

vslavik:
> But, of course, I thought this code was fine before too... 

...and while I didn't like it and was vaguely suspicious of the change, I failed to pinpoint the problem myself in time :/

> I'd really like to find a way to have some tests for it, i.e. assert that, for the given tuple `(msgid, user-preferred-languages-array, actually-available-catalogs)` we get the expected result. Could we refactor this code somehow or maybe add a private function that could be called from the tests to verify that this works as expected? IMO we really need this because this code has been proven to be brittle.

A lot of brittleness was in interactions within `wxTranslations` code, so I'd be worried that just testing the matching algorithm wouldn't be terribly useful — better to test the user-facing API. We can plug in a mock `wxTranslationsLoader` easily enough to pretend which translations are available, but we also need `wxUILocale::GetPreferredUILanguages()`. I was going to suggest using a mock `wxUILocaleImpl`, but apparently this method is static in it too.

So another alternative would be to have a private function pointer just for `wxTranslations` use:

```diff
diff --git a/src/common/translation.cpp b/src/common/translation.cpp
@@ -76,6 +76,11 @@ const size_t32 MSGCATALOG_MAGIC_SW = 0xde120495;
 // implementation
 // ============================================================================
 
+// The function pointer is only exported because it is used in the unit test, it
+// is not part of the public API.
+WXDLLIMPEXP_BASE
+wxVector<wxString> (*wxTranslations_GetPreferredUILanguages)() = &wxUILocale::GetPreferredUILanguages;
+
 namespace
 {
 
@@ -118,7 +123,7 @@ void LogTraceLargeArray(const wxString& prefix, const wxArrayString& arr)
 
 wxString GetPreferredUILanguage(const wxArrayString& available)
 {
-    wxVector<wxString> preferred = wxUILocale::GetPreferredUILanguages();
+    wxVector<wxString> preferred = wxTranslations_GetPreferredUILanguages();
     LogTraceArray(" - system preferred languages", preferred);
 
     wxString langNoMatchRegion;
```

Or move `GetPreferredUILanguage()` to become a method in `wxTranslations` and add a virtual `GetPreferredUILanguages()` to it that would by default just call its `wxUILocale` counterpart, but could be overridden in the test suite.

vslavik:
> So another alternative would be to have a private function pointer just for wxTranslations use:

Actually, maybe doing that in `wxUILocale:: GetPreferredUILanguages()` would be best as it could be used for testing other things as well?

vadz:
> Actually, maybe doing that in `wxUILocale::GetPreferredUILanguages()` would be best as it could be used for testing other things as well?

I'd make things maximally simple and allow overriding the return value of this function by predefining `WX_LANGUAGE` environment variable (name up to discussion, as always, but this one is consistent with `LANGUAGE` used by glibc) to contain a colon-separated list of languages.

This could be useful for more than testing, too.

vadz:
> > I'd make things maximally simple and allow overriding the return value of this function by predefining WX_LANGUAGE
> 
> I.e. do it for all ports before (and so instead of) calling `wxUILocaleImpl:: GetPreferredUILanguages()`?

Yes (well, if it's defined, of course). I don't see any drawbacks in doing this and it could be used to actually override the language used by any wx app if the user wants to, which looks like a nice feature to have (especially for non-Linux systems without `LANGUAGE`).

src/common/uilocale.cpp
```diff
@@ -743,6 +748,29 @@ int wxUILocale::GetSystemLocale()
...
+    {
+        wxVector<wxString> preferred;
+        wxStringTokenizer tknzr(languageFromEnv, ":");
+        while (tknzr.HasMoreTokens())
+        {
+            const wxString tok = tknzr.GetNextToken();
+            if (const wxLanguageInfo* li = wxUILocale::FindLanguageInfo(tok))
+            {
+                preferred.push_back(li->CanonicalName);
+            }
```
vadz, line 765: Very minor, but maybe add a trace/debug message saying that we ignore unknown `tok` in `WX_LANGUAGE`?

tests/intl/intltest.cpp
```diff
@@ -287,6 +289,60 @@ TEST_CASE("wxTranslations::AddCatalog", "[translations]")
...
+        wxSetEnv("WX_LANGUAGE", "cs:en:fr");
+        CHECK( trans.GetBestTranslation(domain) == "en" );
+
+        // ...But won't be used if there's a suitable translation file.
+        wxSetEnv("WX_LANGUAGE", "fr:en:cs");
+        CHECK( trans.GetBestTranslation(domain) == "fr" );
+        CHECK( trans.GetBestAvailableTranslation(domain) == "fr" );
+    }
+
+    wxUnsetEnv("WX_LANGUAGE");
```
vadz, line 343: I'd use `wxON_BLOCK_EXIT` to do this to ensure this variable doesn't remain set and then affect some subsequent test in very hard to find way later.

vadz (commented):
Thanks a lot!

The only thing I'd like to add is to document the new variable in `docs/doxygen/overviews/envvars.h` (also, looking at the other variables there, I wonder if we shouldn't have called it `WXLANGUAGE` for consistency... but I'll let you decide whether to change or keep the name).

The tests look really good, thanks again!

tests/intl/intltest.cpp
vslavik: What am I missing here? This seems to be the canonical way to do teardown; the code is executed when a check fails and even if an unexpected exception is thrown anywhere within the SECTION...

vadz: Right now it's indeed fine, but if anybody adds a `REQUIRE` test, it would exit without executing this line, so it just seems safer to ensure it does always get executed.

vslavik: I see, thanks!

src/common/uilocale.cpp
vslavik: I took this code from UNIX implementation, so it would make sense to add tracing there, but there's the logic more complicated and would significantly inflate the function... At that point I realized that the trace output lists detected system languages anyway, so any misparsing should be pretty obvious if you compare it to WXLANGUAGE value — which I changed the code to include in the trace message.

vadz (approved):
Great, thanks again for fixing this!

I'll merge it into master already but will wait a bit before backporting to 3.2, just in case we're still missing something. But I'm feeling much more confident about this code correctness now, thanks!

vadz:
This is not completely trivial to backport to 3.2 (doing it naively results in conflicts and we also need to replace the default parameter with overloaded functions to preserve ABI).

Would anybody be interested in doing this by chance? If yes, please don't hesitate to submit PRs against 3.2 with the changes here. TIA!

PR #24298, opened 2024-02-11
Patches for wxQt/Android

dsa-t:
Patches required to get wxQt on Android built. Also allows wxLogDebug to work on Android.

Not sure if `android_*` functions are still needed - nothing seemed to fail when using normal functions instead, but I don't have comprehensive tests.

Build scripts: https://github.com/dsa-t/wx-android-minimal


configure.ac
```diff
@@ -532,7 +532,7 @@ WX_ARG_WITH(opengl,        [  --with-opengl           use OpenGL (or Mesa)], wxU
 WX_ARG_WITH(xtest,         [  --with-xtest            use XTest extension], wxUSE_XTEST)
 WX_ARG_WITH(nanosvg,       [  --with-nanosvg          use NanoSVG for rasterizing SVG], wxUSE_NANOSVG)
 
-if test "$wxUSE_GTK" = 1 -o "$wxUSE_QT" = 1 -o "$wxUSE_X11" = 1; then
+if test "$wxUSE_GTK" = 1 -o "$wxUSE_X11" = 1; then
```
vadz, line 535: This seems to result in a crash in the test, unless it's some weird coincidence.

vadz (commented):
If Android-specific functions can be removed, it would be nice to do it, of course.

But I'm mostly worried by the crash in the test, do you see it locally?

configure.ac
dsa-t: You're right. Rather `--enable-graphics_ctx=no` should be passed to configure.

dsa-t:
Also added patches for building wx for Android via CMake. This way I don't have to use patchelf tricks to remove .so versioning suffixes.

https://github.com/wxWidgets/wxWidgets/pull/24298/commits/1059bd52cd688b286ee0f341c188741b071904f2 is needed because `config.cmake` puts wx libraries in wx-config file like this:
`ldlibs_xml="-lwx_baseu-3.3-Android .../sysroot/x86_64/usr/lib/libexpat.so"`
without the ABI suffix that `Qt5AndroidSupport.cmake` adds.

For example, `CMAKE_SHARED_LIBRARY_SUFFIX_CXX` would've been `_x86_64.so`,
leading to file name `libwx_baseu-3.3-Android_x86_64.so`.

Build scripts for CMake: https://github.com/dsa-t/wx-android-minimal-cmake

vadz (approved):
Thanks, this looks good to me (even though I wonder if we shouldn't handle Android separately and skip all pthreads-related compilation checks there, as it presumably passes them all anyhow), but I'd like to ask @MaartenBent for confirmation concerning CMake-related changes.

MaartenBent:
I'm wondering why you first do `set(CMAKE_USE_PTHREADS_INIT 1)`. But then at every `if(CMAKE_USE_PTHREADS_INIT)` check you exclude `ANDROID`. 

Can't you just set it to 0?
Or not set it at all, normally it is set by `find_package(Threads)` but you've disabled this.

vadz:
@dsa-t Do you plan to simplify CMake logic as proposed?

dsa-t:
> @dsa-t Do you plan to simplify CMake logic as proposed?

Yes, I'll check it.

dsa-t:
`FindThreads.cmake`'s test fails because Android doesn't implement `pthread_cancel`:

<details><summary>Error log</summary>
<p>

```
Performing C SOURCE FILE Test CMAKE_HAVE_LIBC_PTHREAD failed with the following output:
Change Dir: ${builddir}/CMakeFiles/CMakeScratch/TryCompile-ubkGdg

Run Build Command(s):/usr/bin/gmake -f Makefile cmTC_d4f96/fast && /usr/bin/gmake  -f CMakeFiles/cmTC_d4f96.dir/build.make CMakeFiles/cmTC_d4f96.dir/build
gmake[1]: Entering directory '${builddir}/CMakeFiles/CMakeScratch/TryCompile-ubkGdg'
Building C object CMakeFiles/cmTC_d4f96.dir/src.c.o
${toolchain}/bin/clang --target=x86_64-none-linux-android28 --gcc-toolchain=${toolchain} --sysroot=${toolchain}/sysroot -DCMAKE_HAVE_LIBC_PTHREAD  -g -DANDROID -fdata-sections -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security   -fPIE -MD -MT CMakeFiles/cmTC_d4f96.dir/src.c.o -MF CMakeFiles/cmTC_d4f96.dir/src.c.o.d -o CMakeFiles/cmTC_d4f96.dir/src.c.o -c ${builddir}/CMakeFiles/CMakeScratch/TryCompile-ubkGdg/src.c
${builddir}/CMakeFiles/CMakeScratch/TryCompile-ubkGdg/src.c:13:3: warning: implicit declaration of function 'pthread_cancel' is invalid in C99 [-Wimplicit-function-declaration]
  pthread_cancel(thread);
  ^
1 warning generated.
Linking C executable cmTC_d4f96
/usr/bin/cmake -E cmake_link_script CMakeFiles/cmTC_d4f96.dir/link.txt --verbose=1
${toolchain}/bin/clang --target=x86_64-none-linux-android28 --gcc-toolchain=${toolchain} --sysroot=${toolchain}/sysroot -g -DANDROID -fdata-sections -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security   -Wl,--exclude-libs,libgcc.a -Wl,--exclude-libs,libgcc_real.a -Wl,--exclude-libs,libatomic.a -static-libstdc++ -Wl,--build-id -Wl,--fatal-warnings -Wl,--no-undefined -Qunused-arguments -Wl,--gc-sections -llog -Wl,-rpath-link=sysroot/x86_64/usr/lib  CMakeFiles/cmTC_d4f96.dir/src.c.o -o cmTC_d4f96  -latomic -lm 
src.c:13: error: undefined reference to 'pthread_cancel'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[1]: *** [CMakeFiles/cmTC_d4f96.dir/build.make:100: cmTC_d4f96] Error 1
gmake[1]: Leaving directory '${builddir}/CMakeFiles/CMakeScratch/TryCompile-ubkGdg'
gmake: *** [Makefile:127: cmTC_d4f96/fast] Error 2


Source file was:
#include <pthread.h>

static void* test_func(void* data)
{
  return data;
}

int main(void)
{
  pthread_t thread;
  pthread_create(&thread, NULL, test_func, NULL);
  pthread_detach(thread);
  pthread_cancel(thread);
  pthread_join(thread, NULL);
  pthread_atfork(NULL, NULL, NULL);
  pthread_exit(NULL);

  return 0;
}
```

</p>
</details> 

But I agree `CMAKE_USE_PTHREADS_INIT` should be checked for in other places instead of Android.

MaartenBent:
`CMAKE_HAVE_LIBC_PTHREAD` seems like some hidden/undocumented variable in `findThreads`. One that will also affect other compilers.
Either use the official `CMAKE_THREAD_LIBS_INIT`, like in my idea below.
Or define your own variable (just `HAVE_LIBC_PTHREAD` or something).

-----

I have some ideas for further improvements, not needed for this PR.
It might make sense to use `CMAKE_THREAD_LIBS_INIT` instead of using `-pthread`.
Maybe in combination with the `THREADS_PREFER_PTHREAD_FLAG` option.

config.cmake
```cmake
if(CMAKE_THREAD_LIBS_INIT)
    set(WXCONFIG_CFLAGS ${CMAKE_THREAD_LIBS_INIT})
    set(WXCONFIG_LDFLAGS ${CMAKE_THREAD_LIBS_INIT})
endif()
```
setup.cmake
```cmake
set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_THREAD_LIBS_INIT})
```
functions.cmake (maybe we can get rid of the clang check too)
```cmake
if(CMAKE_THREAD_LIBS_INIT)
    target_compile_options(${target_name} PRIVATE ${CMAKE_THREAD_LIBS_INIT})
    set_target_properties(${target_name} PROPERTIES LINK_FLAGS ${CMAKE_THREAD_LIBS_INIT})
endif()
```

vadz:
I agree that it would be better to use `CMAKE_THREAD_LIBS_INIT`, it seems to be there exactly in order to account for the case when pthread functions are in libc.

I also wonder, where is `CMAKE_USE_PTHREADS_INIT` getting defined, exactly? AFAICS it's supposed to be done by `FindThreads`, but where do we use it?

dsa-t:
I also agree that it's better to use `${CMAKE_THREAD_LIBS_INIT}` instead of `"-pthread"`, and some `if` checks can be removed,
However I wanted to keep changes in this PR to a minimum as to not accidentally break other build configurations.

vadz:
Could you perhaps make a separate PR with these changes and then rebase this on top of it? I'd like to avoid introducing known bad CMake changes because I am never confident enough in my CMake knowledge to clean them up later...

dsa-t:
>Could you perhaps make a separate PR with these changes and then rebase this on top of it?

This is for @MaartenBent I assume? I'm not building with Win32/Clang to test the clang++.exe check removal.

>I'd like to avoid introducing known bad CMake changes

I wouldn't call current changes "known bad".

vadz:
Sorry if "bad" is too strong, but AFAIU using `CMAKE_HAVE_LIBC_PTHREAD` is incorrect, and I'd prefer to avoid introducing it, if possible.

If Maarten could please do it, it would be great, of course, but I'm almost sure Maarten wouldn't mind if you did it neither!

MaartenBent:
Yes, I can do that.

For this PR I still suggest to not use `CMAKE_HAVE_LIBC_PTHREAD` because it is an undocumented output from `FindThreads`. Which is probably also set for most gcc/clang compilers.

dsa-t:
> not use CMAKE_HAVE_LIBC_PTHREAD because it is an undocumented output from FindThreads

I didn't really think this is a problem but ok.

Another idea is to provide a custom `FindThreads.cmake` that doesn't check for `pthread_cancel`

MaartenBent:
> I didn't really think this is a problem but ok.

The problem is that it might be different in an older version of CMake, and can change in a new version without any notice.

I suggest to just rename it to `wxHAVE_LIBC_PTHREAD`, then only Android will be affected.
 All this will probably change anyway when using `CMAKE_THREAD_LIBS_INIT`, so we should use a quick and simple solution.

dsa-t:
Rebase done.

PR #24300, opened 2024-02-11
Fix missing "true", small rewording, make other true bold

Blake-Madden:
Rewording explains what "peer verification" means.


vadz:
Thanks for the corrections! I will merge them soon.

PR #24301, opened 2024-02-11
Revert clipping changes under Mac as they break `wxGrid`

### LINKED ISSUE #24201: Cell editors no longer appear in wxGrid (macOS)
#### Bug description:

With the changes introduced in bcbc31e97f246564e4b19ea2d060ae859c302b44 , the grid cell editors no longer appear when editing cells in a wxGrid on OSX.    This can easily be shown with the grid sample.    Trying to edit any cell will display the text "Cell editor shown.", but the editor never appears visible.   Typing text and hitting enter does enter the text, but the user cannot see what they are typing. 

Users would expect to be able to see what they are typing. 

The grid sample "as is" can be used to reproduce.

- wxWidgets version  you use:  latest master code
- wxWidgets port  you use:   macOS
- OS  and its version:   macOS Sonoma in dark mode

vadz:
Stefan, as I don't know how to fix #24201, it looks like we need to revert your fix for the scrollbars under macOS 14.

Please let me know if I should hold off merging this to give you time to see if you can fix the other bug in a better way. TIA!

### DIFF
--- a/include/wx/osx/cocoa/private.h
+++ b/include/wx/osx/cocoa/private.h
@@ -220,14 +220,9 @@ public :
     // from the same pimpl class.
     virtual void                controlTextDidChange();
 
-    virtual void                AdjustClippingView(wxScrollBar* horizontal, wxScrollBar* vertical) override;
-    virtual void                UseClippingView(bool clip) override;
-    virtual WXWidget            GetContainer() const override { return m_osxClipView ? m_osxClipView : m_osxView; }
-
 protected:
     WXWidget m_osxView;
-    WXWidget m_osxClipView;
-
+    
     // begins processing of native key down event, storing the native event for later wx event generation
     void BeginNativeKeyDownEvent( NSEvent* event );
     // done with the current native key down event
--- a/include/wx/osx/cocoa/private.h
+++ b/include/wx/osx/cocoa/private.h
@@ -221,7 +221,7 @@ public :
     virtual void                controlTextDidChange();
 
     virtual void                AdjustClippingView(wxScrollBar* horizontal, wxScrollBar* vertical) override;
-    virtual void                UseClippingView() override;
+    virtual void                UseClippingView(bool clip) override;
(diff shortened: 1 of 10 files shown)


csomor:
I thought about adding some switch to only trigger the replacement with the scroll view for true wxscrolled windows

vadz:
> I thought about adding some switch to only trigger the replacement with the scroll view for true wxscrolled windows

This would be great, of course, but I have no idea how to do this (and I can't experiment as I don't even have macOS 14) and in the meanwhile people complain about not being able to use wxGrid...

csomor:
> > I thought about adding some switch to only trigger the replacement with the scroll view for true wxscrolled windows
> 
> This would be great, of course, but I have no idea how to do this (and I can't experiment as I don't even have macOS 14) and in the meanwhile people complain about not being able to use wxGrid...

ok, so go ahead, revert, and I'll open a separate branch again for this

csomor (approved):
ok, so go ahead, revert, and I'll open a separate branch trying to make this change only for true wxscrolled windows

vadz:
Just to be clear, I'd strongly prefer _not_ to merge this, as reverting the existing commits and then applying them again makes history more difficult to follow and, also, needs to be done in 3.2 branch too. So if we can avoid doing it, it would be much better.

Shall we perhaps leave this one here for a week or two to see if you manage to come out with a better fix?

Also, I don't think I can help here, but please let me know if I can.

csomor:
> Just to be clear, I'd strongly prefer _not_ to merge this, as reverting the existing commits and then applying them again makes history more difficult to follow and, also, needs to be done in 3.2 branch too. So if we can avoid doing it, it would be much better.
> 
> Shall we perhaps leave this one here for a week or two to see if you manage to come out with a better fix?
> 
> Also, I don't think I can help here, but please let me know if I can.

if a week is fine, I'll try to come up with a better solution

vadz:
Replaced by #24342.

PR #24304, opened 2024-02-11
Minor wxPropertyGrid fixes

vadz:
@a-wi I don't even know if we should have separate `m_iconHeight` different from `m_iconWidth`, IMO these icons should always be square anyhow, but as long as we have it, we need to adjust it in high DPI too. And I made this conditional on `wxPG_ICON_WIDTH` but I'd actually like to remove all tests for it and just assume that it's always 1 (which is the case) and also remove `wxPG_USE_RENDERER_NATIVE` and make it always 1 too (which is not the case under non-MSW/GTK/Mac now).

### DIFF
--- a/samples/propgrid/propgrid.xrc
+++ b/samples/propgrid/propgrid.xrc
@@ -6,8 +6,7 @@
     <object class="wxBoxSizer">
         <object class="sizeritem">
             <option>1</option>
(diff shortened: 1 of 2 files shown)


vadz:
I'm going to merge this to fix a user-visible problem in high DPI, but I still think all the unused stuff should be removed -- but would like to get @a-wi opinion before doing it.

PR #24309, opened 2024-02-12
More wxQt fixes.

dsa-t:
Fixes for some more bugs encountered when running KiCad.

### DIFF
--- a/include/wx/qt/radiobox.h
+++ b/include/wx/qt/radiobox.h
@@ -79,6 +79,9 @@ public:
     virtual void SetSelection(int n) override;
     virtual int GetSelection() const override;
 
+    virtual void SetLabel(const wxString &label) override;
+    virtual wxString GetLabel() const override;
+
     virtual QWidget *GetHandle() const override;
 
 private:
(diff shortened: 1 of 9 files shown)


vadz (approved):
Looks good to me, thanks!

@AliKet any comments?

src/qt/bitmap.cpp
```diff
@@ -237,6 +237,9 @@ void wxBitmap::InitFromImage(const wxImage& image, int depth, double WXUNUSED(sc
                  ? QBitmap::fromImage(ConvertImage(image), Qt::ThresholdDither)
                  : QPixmap::fromImage(ConvertImage(image, &mask));
 
+    if (mask)
+        qtImage.setMask(*mask->GetHandle());
+
```
AliKet, line 242: Keeping the _mask_ seperated from the _qtImage_ **is intentional** because this is how the other ports works. IOW, `setMask()` merges the mask with the pixmap's alpha channel. And AFAIK none of the other ports do this by default.
If you really want the mask be merged just be explicit and call `wxBitmap::QtBlendMaskWithAlpha()`

AliKet (approved):

PR #24311, opened 2024-02-13
Update generic `wxListCtrl` after user has edited an item label

taler21:
With wxWidgets-3.2.4, a generic `wxListCtrl` is not automatically updated when an item label was edited by the user.

For example, if the user edits the label of the first item, adds a few characters and completes the editing by pressing the Enter key, the result currently looks like this:
wxListCtrl-master

This PR fixes this issue by triggering a list control update, including a recalculation of sizes and positions, on the next idle.

wxListCtrl-PR

In case you are wondering if this is a regression or has always been broken.
What I can say is that it doesn't work with either wxWidgets 3.2.4 or 3.0.5.

It may have been broken a long time ago, in the year 2002 by commit 62d89eb4c981d1abf2f7aac5ead7d2f862407f2d.
See the changes there in `wxListMainWindow::OnRenameAccept(...)`.
But that's just speculation, i.e. I have not tested that.

### DIFF
--- a/src/generic/listctrl.cpp
+++ b/src/generic/listctrl.cpp
@@ -2395,8 +2395,12 @@ bool wxListMainWindow::OnRenameAccept(size_t itemEdit, const wxString& value)
 
     data->GetItem( 0, le.m_item );
     le.m_item.m_text = value;
-    return !GetParent()->GetEventHandler()->ProcessEvent( le ) ||
-                le.IsAllowed();
+    bool accept = !GetParent()->GetEventHandler()->ProcessEvent( le ) ||
+                      le.IsAllowed();
+    if ( accept )
+        m_dirty = true;
+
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for fixing this!

And, just for completeness, you're under no obligation to dig into wx long history to find out if each correction is a regression fix or not, of course, I'm just asking it sometimes if I have the impression that it worked before in order to mention that it is a regression fix in the change log in case you already know that is. But thanks for checking anyhow!

vadz:
P.S. I'm applying a slightly modified version of this as I think it's a bit easier to read like this, but the effect is the same. I hope you don't mind this change, but please let me know if you do.

taler21:
Improvements to my changes and comments are always welcome. Thanks.

PR #24314, opened 2024-02-13
wxGrid: add drag-move insertion marker to row/col label drawing

### LINKED ISSUE #24310: wxGrid drop marker missing with double buffering on
I think I found a problem with double buffering (on Windows).
When dragging a row or col, the drop marker is missing now.

E.g. in the grid sample with "Grid -> Row drag move".
In the screenshot it's between lines 6 and 7:

Before merging this branch:
grafik

After:
grafik

I can work around by commenting out `exstyle |= WS_EX_COMPOSITED;`.

The relevant code for drawing the marker is in `wxGrid::ProcessRowColLabelMouseEvent`.
I would guess that some additional action is required around the ClientDC, but could not see what is missing. Other code like "pressing" the row label looks similar, but does not have problems.
```
if ( markerPos != m_dragLastPos || markerColour != m_dragLastColour )
{
    wxClientDC dc( headerWin );
    oper.PrepareDCForLabels(this, dc);

    int markerLength = oper.Select(headerWin->GetClientSize());

    // Clean up the last indicator if position has changed
    if ( m_dragLastPos >= 0 && markerPos != m_dragLastPos )
    {
        wxPen pen(headerWin->GetBackgroundColour(), 2);
        dc.SetPen(pen);
        oper.DrawParallelLine(dc, 0, markerLength, m_dragLastPos + 1);
        dc.SetPen(wxNullPen);

        int lastLine = oper.PosToLine(this, m_dragLastPos, nullptr, false);

        if ( lastLine != -1 )
            oper.DrawLineLabel(this, dc, lastLine);
    }

    // Draw the marker
    wxPen pen(*markerColour, 2);
    dc.SetPen(pen);
    oper.DrawParallelLine(dc, 0, markerLength, markerPos + 1);
    dc.SetPen(wxNullPen);

    m_dragLastPos = markerPos;
    m_dragLastColour = markerColour;
}
```

_Originally posted by @DietmarSchwertberger in https://github.com/wxWidgets/wxWidgets/issues/22851#issuecomment-1939732120_

DietmarSchwertberger:
With double-buffering on Windows the insertion markers were not visible any more when a row or col was drag-moved in a wxGrid.
With this PR they are drawn during paint events.

Fixes https://github.com/wxWidgets/wxWidgets/issues/24310

### DIFF
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -2652,8 +2652,9 @@ protected:
     // operation in progress.
     int     m_dragMoveRowOrCol;
 
-    // Last drag marker position while drag-moving a row or column.
+    // Last drag marker position and line while drag-moving a row or column.
     int     m_dragLastPos;
+    int     m_dragLastRowOrCol;
 
     // Last drag marker colour while drag-moving a row or column.
     const wxColour *m_dragLastColour;
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -2652,9 +2652,8 @@ protected:
     // operation in progress.
     int     m_dragMoveRowOrCol;
 
-    // Last drag marker position and line while drag-moving a row or column.
+    // Last drag marker position while drag-moving a row or column.
     int     m_dragLastPos;
-    int     m_dragLastRowOrCol;
 
     // Last drag marker colour while drag-moving a row or column.
     const wxColour *m_dragLastColour;
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -3003,7 +3003,6 @@ void wxGrid::Init()
     m_canDragCell = false;
     m_dragMoveRowOrCol = -1;
     m_dragLastPos  = -1;
-    m_dragLastRowOrCol = -2;
     m_dragLastColour  = nullptr;
     m_dragRowOrCol = -1;
     m_dragRowOrColOldSize = -1;
@@ -4040,7 +4039,6 @@ void wxGrid::ProcessRowColLabelMouseEvent( const wxGridOperations &oper, wxMouse
                     oper.DrawParallelLine(dc, 0, markerLength, markerPos + 1);
                     dc.SetPen(wxNullPen);
 
-                    m_dragLastRowOrCol = lineAt;
                     m_dragLastPos = markerPos;
                     m_dragLastColour = markerColour;
                 }
@@ -4448,7 +4446,6 @@ void wxGrid::DoAfterDraggingEnd()
     m_lastMousePos = wxDefaultPosition;
     // from drag moving row/col
     m_dragMoveRowOrCol = -1;
-    m_dragLastRowOrCol = -2;
     m_dragLastPos = -1;
     m_dragLastColour = nullptr;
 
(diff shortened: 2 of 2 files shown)


vadz:
Thanks, we can/should indeed apply this for now even though I still think we should switch to `wxOverlay` eventually.

Unfortunately this can't be easily backported to 3.2 as this change is not ABI-compatible. If you'd like to apply it there, I think it would be better to switch to using a global variable instead: this should be fine because we can't be dragging anything in more than one grid at a time, so just having `struct { wxGrid* grid; int rowOrCol; } gs_currentDragInfo;` should be ok and this could be cherry-picked to 3.2 without problems.

Would you like to do this or do you prefer to apply this "as is" to master only?

DietmarSchwertberger:
Thanks. My C++ knowledge is still very basic, even though I bought a copy of "A Tour of C++" for Christmas.

I would like to see it in 3.2 as currently I'm working on my switch to Python 3.12 and a more recent version of wxPython. I find the new drag mechanism a huge improvement over the old wxPython specific implementation.

Is the anonymous namespace the right place for the global variable?

```
namespace
{
...
struct DefaultHeaderRenderers
{
...
} gs_defaultHeaderRenderers;


// this holds information about the current position of the insertion
// marker while a row or col drag-move is ongoing
struct
{
    wxGrid* grid = nullptr;
    int rowOrCol = -2;
} gs_currentDragInfo;

} // anonymous namespace
```

Should I submit a separate PR for 3.2 and keep `m_dragLastRowOrCol;` for master branch? 
(Initially I thought about calculating the row or col on the fly from the mouse position, but I dropped that idea as it would probably be quite ugly.)

DietmarSchwertberger:
Actually, it might be better to move out the marker drawing into `DrawRowLabels` and  `DrawColLabels`.
This does avoid the global variable that had only the purpose to avoid drawing the marker many times.

Also, I noticed a problem with the current code: the check `(m_dragLastRowOrCol == row || m_dragLastRowOrCol == row-1 )` is not sufficient when rows or cols were re-orderd.

DietmarSchwertberger:
... or into `GridRowLabelWindow::OnPaint`, after `m_owner->DrawRowLabels( dc, rows );`?

DietmarSchwertberger:
I think `DrawColLabels` is the best place. Code inside `GridRowLabelWindow::OnPaint`would require many `m_owner->...`
I will give it a try.

DietmarSchwertberger:
That looks much nicer... Should have done it so right from the beginning.

vadz:
Thanks for the update! This is indeed nicer and easy to backport (will do both soon).

vadz:
Wait, sorry, I've realized that we probably don't actually need to backport it to 3.2 as double buffering is not used under MSW there. Or does the problem still occur with 3.2 too somehow?

DietmarSchwertberger:
Oh, I did not check, as I was just about to use the Drag-Moving in my application. If double-buffering is not active in 3.2 then we don't need the backport. Anyway, the final, simplified version is also the way to go with the master branch.

DietmarSchwertberger:
I noticed the problem while I was working on wxWidgets itself: accessibility support and row/col label drag resizing for wxGrid. Expect pull requests for this soon.

PR #24316, opened 2024-02-13
Fix dialogs sample never exiting with a persistent taskbar icon

lanurmi:
If the latest user notification shown was shown with "Use persistent taskbar icon" toggled on, the application process would never really exit after closing. So delete m_taskbarIcon in destructor.

Observed on Windows 10.


lanurmi:
There appears to be a similar issue when `MSWUseToasts()` is not enabled. This PR doesn't address that one yet.

PR #24318, opened 2024-02-14
Improve wxSearchCtrl style on wxQt.

dsa-t:
Before:
image

After:
image

(on wxQt)

Removes the border by default and makes button sizes smaller, in line with the text field
Screenshot_2024-02-14_12-47-46

### DIFF
--- a/include/wx/generic/srchctlg.h
+++ b/include/wx/generic/srchctlg.h
@@ -144,6 +144,8 @@ public:
 #endif // wxUSE_MENUS
 
 protected:
+    virtual wxBorder GetDefaultBorder() const override { return wxBORDER_NONE; }
+
     virtual void DoSetValue(const wxString& value, int flags) override;
     virtual wxString DoGetValue() const override;
 
--- a/include/wx/generic/srchctlg.h
+++ b/include/wx/generic/srchctlg.h
@@ -144,8 +144,6 @@ public:
 #endif // wxUSE_MENUS
 
 protected:
-    virtual wxBorder GetDefaultBorder() const override { return wxBORDER_NONE; }
-
     virtual void DoSetValue(const wxString& value, int flags) override;
     virtual wxString DoGetValue() const override;
 
--- a/include/wx/qt/window.h
+++ b/include/wx/qt/window.h
@@ -112,6 +112,8 @@ public:
     virtual int GetScrollThumb( int orient ) const override;
     virtual int GetScrollRange( int orient ) const override;
 
+    virtual wxSize GetWindowBorderSize() const override;
+
         // scroll window to the specified position
     virtual void ScrollWindow( int dx, int dy,
                                const wxRect* rect = nullptr ) override;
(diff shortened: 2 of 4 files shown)


vadz:
Sorry, IMO this looks worse than the current appearance under MSW: we definitely the border there and the images are IMHO too small after this change (this is more subjective, but they look too small compared with the image shown in e.g. Explorer).

I'm not sure why things should be different for wxQt. What does the big screenshot show BTW?

dsa-t:
> this looks worse than the current appearance under MSW
You're right.

wxBORDER_THEME (wxBORDER_DOUBLE) seems too thick on wxQt.

dsa-t:
It calls
`wxSystemSettingsNative::GetMetric(wxSYS_EDGE_Y)` -> 2
`wxSystemSettingsNative::GetMetric(wxSYS_BORDER_Y)` -> 2
So the size increases by (2+2)*2 = 8.

dsa-t:
Now, trying to implement `wxBORDER_THEME` in wxQt, in `wxSearchCtrl::Create` I'm calling:
```cpp
m_qtWindow->setStyleSheet("border: 1px solid blue; background: cyan");
```
But it only changes the editable field:
Screenshot_2024-02-18_07-50-05

Not sure why.

dsa-t:
Here's how it looks now on wxQt:

image

The total height matches normal textfield/combobox/button sizes.

vadz:
This looks good to me now, thanks!

Just one question: what is the state of high DPI support in wxQt? I.e. should we perhaps use `FromDIP()` for these hardcoded values?

dsa-t:
At 200 dpi, borders and check/radiobuttons do not become bigger on wxQt with default style or other system apps. But they are scaled in Qt5 example.

gtk2 (default) style:
Screenshot_2024-02-18_18-45-41

Check/radiobuttons are scaled when using "Fusion" style:
Screenshot_2024-02-18_18-43-14

dsa-t:
The scaling on gtk2 can be enabled by setting env var `QT_AUTO_SCREEN_SCALE_FACTOR=1`

Screenshot_2024-02-18_19-04-22

Screenshot_2024-02-18_19-05-16

vadz:
Sorry, I'm not sure what does all this mean... All I wanted to ask was whether Qt API uses DPI-independent or physical pixels, i.e. whether creating a 100x100 window at 200% scaling creates a window 100 or 200 physical pixels wide/high?

dsa-t:
> whether creating a 100x100 window at 200% scaling creates a window 100 or 200 physical pixels wide/high?

That gives 200 physical pixels when `QT_AUTO_SCREEN_SCALE_FACTOR=1`, otherwise 100, it looks like.

200 dpi, `QT_AUTO_SCREEN_SCALE_FACTOR` not set, `QT_STYLE_OVERRIDE=fusion`:
Screenshot_2024-02-18_19-30-31

vadz:
I have no idea how to deal with that, so I guess we should just merge this PR as is and leave high DPI discussion for another time. But I suspect we might have a problem here as, IIUC, wxQt doesn't behave like any of the other ports with high DPI support (wxMSW vs wxGTK3/wxOSX).

PR #24324, opened 2024-02-15
Fix duplicate accelerator for the font dialog in dialogs sample

lanurmi:
Shift-Ctrl-N was already taken when it was added for the font dialog in 9b25ed0 (Add an accelerator for the font dialog in the dialogs sample, 2022-04-16). Use Shift-Ctrl-J instead.


vadz:

PR #24325, opened 2024-02-15
Correct Hungarian translation of "Shift" key

taler21:
I am not Hungarian, but:

* The original translator had no information that the keyboard key was meant.
  See 42a6fe0e2fcea1aca012e96b46b0d57dcf678a45 (added Hungarian translations, 2002-03-04).
* A Hungarian translator who translated texts for our software called the key "Shift".
* On Windows 10 with Hungarian selected as the display language, "Shift" is used for accelerators in the text editor.
* On the [Hungarian Wikipedia](https://hu.wikipedia.org/wiki/Shift) the key is also called "Shift".

So I am pretty sure that the current translation is wrong and the key should be called "Shift".

### DIFF
--- a/locale/hu.po
+++ b/locale/hu.po
@@ -724,7 +724,7 @@ msgstr "alt"
 #: ../src/common/accelcmn.cpp:190
 msgctxt "keyboard key"
 msgid "shift"
-msgstr "eltol"
+msgstr "shift"
 
 #: ../src/common/accelcmn.cpp:192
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for looking at this!

Hungarian translations seem to be unmaintained since a very long time, so I'll merge this if there are no objections from any Hungarian-speakers who might be reading this.

PR #24326, opened 2024-02-15
Try to fix wxAUI broken layout under wxQt

AliKet:
The `wxAUI` layout is screwed up at start up as can be seen in the **auidemo** sample:

Screenshot_2024-02-15_14-33-01

With this PR applied, the layout is fixed if you select **Perspectives -> Default Startup**:

Screenshot_2024-02-15_14-33-41

There are three possible solutions to this problem, but unfortunately none of them is perfect/ideal:

1. Commenting out line **1581** https://github.com/wxWidgets/wxWidgets/blob/47b011c665b7e07eef90489c22ff426a60b38ddb/src/aui/framemanager.cpp#L1578-L1581 but as you can see, the comment there says that we can't do this! see commit 5c62cb6 and commit dec588b.
2. Apply this patch:
```
index ba3db6c8bb..a31eef27d8 100644
--- a/src/qt/window.cpp
+++ b/src/qt/window.cpp
@@ -434,6 +434,9 @@ bool wxWindowQt::Create( wxWindowQt * parent, wxWindowID id, const wxPoint & pos
     if ( parent )
         parent->AddChild( this );
 
+    if ( IsTopLevel() )
+        GetHandle()->setVisible(true);
+
     wxPoint p;
     if ( pos != wxDefaultPosition )
         p = pos;
``` 

   i.e. _**Temporarily**_ show the top-level window at creation time before calling `DoMoveWindow()` to get an immediate window resize at the **Qt** level. Note that the top-level window will still be created hidden as documented when `PostCreation()` is called.

3. Postpone calling `wxAuiManager::Update()` in the auidemo like this:

```
diff --git a/samples/aui/auidemo.cpp b/samples/aui/auidemo.cpp
index 56f5384bb6..19ef383cfd 100644
--- a/samples/aui/auidemo.cpp
+++ b/samples/aui/auidemo.cpp
@@ -1015,7 +1015,11 @@ MyFrame::MyFrame(wxWindow* parent,
     m_perspectives.Add(perspective_all);
 
     // "commit" all changes made to wxAuiManager
+#ifndef __WXQT__
     m_mgr.Update();
+#else // __WXQT__
+    m_mgr.CallAfter(&wxAuiManager::Update);
+#endif // !__WXQT__
 }
 
 wxAuiDockArt* MyFrame::GetDockArt()
```
4. Any better idea ?

### DIFF
--- a/src/aui/auibar.cpp
+++ b/src/aui/auibar.cpp
@@ -2248,6 +2248,10 @@ void wxAuiToolBar::DoIdleUpdate()
     }
 }
 
+wxSize wxAuiToolBar::DoGetBestSize() const
+{
+    return GetClientSize();
+}
 
 void wxAuiToolBar::OnSize(wxSizeEvent& WXUNUSED(evt))
 {
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -644,6 +644,18 @@ void wxAuiManager::SetManagedWindow(wxWindow* wnd)
     }
 
 #endif
+
+#ifdef __WXQT__
+    Bind(wxEVT_SHOW, [this](wxShowEvent& event)
+    {
+        if ( event.IsShown() )
+        {
+            this->Update();
+        }
+
+        event.Skip();
+    });
+#endif
 }
 
 
@@ -2201,6 +2213,11 @@ void wxAuiManager::GetDockSizeConstraint(double* width_pct, double* height_pct) 
 
 void wxAuiManager::Update()
 {
+#ifdef __WXQT__
+    if ( !GetManagedWindow()->IsShownOnScreen() )
+        return;
+#endif
+
     m_hoverButton = nullptr;
     m_actionPart = nullptr;
 
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -797,17 +797,7 @@ bool wxAuiManager::AddPane(wxWindow* window, const wxAuiPaneInfo& paneInfo)
     if (pinfo.best_size == wxDefaultSize &&
         pinfo.window)
     {
-        pinfo.best_size = pinfo.window->GetClientSize();
-
-#if wxUSE_TOOLBAR
-        if (wxDynamicCast(pinfo.window, wxToolBar))
-        {
-            // GetClientSize() doesn't get the best size for
-            // a toolbar under some newer versions of wxWidgets,
-            // so use GetBestSize()
-            pinfo.best_size = pinfo.window->GetBestSize();
-        }
-#endif // wxUSE_TOOLBAR
+        pinfo.best_size = pinfo.window->GetBestSize();
 
         if (pinfo.min_size != wxDefaultSize)
         {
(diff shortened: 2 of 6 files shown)


vadz:
Would calling `wxAuiManager::Update()` when we get `wxShowEvent` fix this?

AliKet:
> Would calling `wxAuiManager::Update()` when we get `wxShowEvent` fix this?

I added a handler at the end of this function https://github.com/wxWidgets/wxWidgets/blob/7150e6ecb385f6b76c7e2d81ff030305db51b344/src/aui/framemanager.cpp#L610
It works, but only if I also comment out this line https://github.com/wxWidgets/wxWidgets/blob/7150e6ecb385f6b76c7e2d81ff030305db51b344/samples/aui/auidemo.cpp#L1017-L1018
IOW, user code should not call `wxAuiManager::Update()` before the `wxShowEvent` handler is called. I still don't know why though!!

src/qt/window.cpp
```diff
@@ -1128,7 +1128,7 @@ wxSize wxWindowQt::DoGetBestSize() const
 {
     const wxSize size = wxWindowBase::DoGetBestSize();
 
-    if ( dynamic_cast<wxQtWidget*>(GetHandle()) )
+    if ( !GetHandle() || dynamic_cast<wxQtWidget*>(GetHandle()) )
```
vadz, line 1131: Is this change just an optimization? Because AFAICS if it's null, dynamic cast would fail anyhow.

This being said, I think it would be best to have some `QtIsNative()` virtual function instead of using this dynamic cast in the first place (if I understand correctly what it is doing).

src/aui/auibar.cpp
```diff
@@ -2248,6 +2248,10 @@ void wxAuiToolBar::DoIdleUpdate()
     }
 }
 
+wxSize wxAuiToolBar::DoGetBestSize() const
+{
+    return GetClientSize();
```
vadz, line 2253: It's not a good idea to return the current size from best size because it means that if the window grows, it can never be shrunk back. Couldn't we use `RealizeHelper()` here to get the real best size?

src/aui/framemanager.cpp
```diff
@@ -2211,6 +2213,11 @@ void wxAuiManager::GetDockSizeConstraint(double* width_pct, double* height_pct)
 
 void wxAuiManager::Update()
 {
+#ifdef __WXQT__
+    if ( !GetManagedWindow()->IsShownOnScreen() )
```
vadz, line 2217: My only explanation of your comment about this only working if `Update()` isn't it called in the sample is that `IsShownOnScreen()` doesn't work correctly in wxQt, could this be the case?

If so, ideal would be to make it work, of course, but failing that we could still use a flag and set it in `wxEVT_SHOW` handler, assuming this event is generated correctly.

vadz (commented):
I'd really like to understand why is the layout broken if it's done while the window is still hidden, it doesn't look like anything should depend on whether it's shown or not.

And, thinking more about this, redoing layout only when it is shown risks resulting in flicker, so it would be really better to do it before...

AliKet:
Will be replaced by an other PR so closing this one.

PR #24327, opened 2024-02-15
Limit radius of rounded rectangles in graphics rendering.

dsa-t:
Test case (replaces "bombs" demo): [dctest.zip](https://github.com/wxWidgets/wxWidgets/files/14302935/dctest.zip)

Before (GTK3):
gtk_before

After:
gtk_after

### DIFF
--- a/src/common/graphcmn.cpp
+++ b/src/common/graphcmn.cpp
@@ -436,6 +436,9 @@ void wxGraphicsPathData::AddRoundedRectangle( wxDouble x, wxDouble y, wxDouble w
         AddRectangle(x,y,w,h);
     else
     {
+        wxDouble maxR = std::min(w, h) / 2.0;
+        if ( radius > maxR ) radius = maxR;
+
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for improving this, but this should probably be done for Windows backends too, shouldn't it?

dsa-t:
I haven't checked how it renders on Windows.

dsa-t:
Looks fine on MSW:
msw

paulcor:
Why is this necessary? Isn't it up to the user to provide an appropriate radius?

dsa-t:
> Why is this necessary?

To be inline with other Graphics APIs which limit the radius internally.

vadz:
Thanks for checking this, as this seems to make the results more consistent across platforms, I'd like to merge it. @paulcor Any objections?

BTW, it would be nice to have this test in the `drawing` sample to be able to rerun it later.

PR #1139, opened 2019-01-14
Qt graphics renderer

jaynabonne:
My apologies for this being such a large pull request, but it's actually implementing a big thing - a Qt-based graphics renderer. The Qt code on Linux uses Cairo, which doesn't exist on Windows. As such, the tests, samples, etc would crash, since it returned a NULL renderer. This remedies that. In theory, this could be used for Qt on any platform (including Linux), which would hopefully mean consistent behaviour when using Qt. That has not been done yet, but could be (if desired) once this first chunk is accepted.

I did change the behaviour of GetCairoRenderer (making it only available if wxUSE_CAIRO is defined) as it really complicated the conditionals to have it always enabled (only to return a NULL pointer anyway), and I wasn't sure if returning a NULL pointer made more sense than not simply compiling. Since nobody checked for NULL pointers, I suspected nobody ever tried it outside of wxUSE_CAIRO anyway.

There are some unimplemented methods: BeginLayer and EndLayer were left unimplemented because there was no code using them, and I wasn't 100% clear on the expected behaviour; GetPartialTextExtents was left unimplemented as well, since there was similarly no code using it. These can be implemented if necessary.

Also, the gradient stops handling is less than "wX optimal", as Qt doesn't support multiple stops at the same location (to create hard edges). I'm not even sure of a workaround for that.

The only quirky code (in my opinion) is the path code, as wX requires the current position to return to the path beginning when a path is closed, whereas Qt sets it to (0, 0). 

Please let me know the best way to proceed with this. Given that even small pull requests usually get at least some push back, I have zero hope that this will go through first time. :) But I would like to work out a path to getting it accepted, with whatever changes are deemed necessary,

PR #24328, opened 2024-02-15
Improve wxQtDCImpl to better match Cairo rendering + enable anti-aliasing.

dsa-t:
Test case (replaces "bombs" demo): [dctest.zip](https://github.com/wxWidgets/wxWidgets/files/14302935/dctest.zip)

Reference (GTK3 after https://github.com/wxWidgets/wxWidgets/pull/24327):

gtk_after

wxQt, with this PR:
qt_after

wxQt, current master (shrinking non-stroked rectangles caused "holes" in UI, and arcs were incorrect):
qt_before

### DIFF
--- a/src/qt/dc.cpp
+++ b/src/qt/dc.cpp
@@ -83,6 +83,12 @@ void wxQtDCImpl::QtPreparePainter( )
         m_qtPainter->setBrush( wxBrush().GetHandle() );
         m_qtPainter->setFont( wxFont().GetHandle() );
 
+        if (m_qtPainter->device()->depth() > 1)
+        {
+            m_qtPainter->setRenderHints(QPainter::Antialiasing,
+                                        true);
+        }
+
         if (m_clipping)
         {
             m_qtPainter->setClipRegion( m_clippingRegion.GetHandle() );
--- a/src/qt/dc.cpp
+++ b/src/qt/dc.cpp
@@ -924,12 +924,16 @@ void wxQtDCImpl::DoDrawLines(int n, const wxPoint points[],
             path.lineTo(wxQtConvertPoint(points[i]));
         }
 
-        m_qtPainter->translate(xoffset, yoffset);
+        {
+            QtDCOffsetHelper helper(m_qtPainter);
 
-        QBrush savebrush = m_qtPainter->brush();
-        m_qtPainter->setBrush(Qt::NoBrush);
-        m_qtPainter->drawPath(path);
-        m_qtPainter->setBrush(savebrush);
+            m_qtPainter->translate(xoffset, yoffset);
+
+            QBrush savebrush = m_qtPainter->brush();
+            m_qtPainter->setBrush(Qt::NoBrush);
+            m_qtPainter->drawPath(path);
+            m_qtPainter->setBrush(savebrush);
+        }
 
         // Reset transform
         ComputeScaleAndOrigin();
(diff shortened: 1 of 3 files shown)


dsa-t:
The test
```
./graphics/bitmap.cpp:434: FAILED:
  CHECK( unmaskedPixelsCount == numUnmaskedPixels )
with expansion:
  60 == 64
```

fails because of the anti-aliased mask rectangle corners:

Screenshot_2024-02-16_02-16-15

This code doesn't work because it doesn't use wxGraphicsContext.
```cpp
#if wxUSE_GRAPHICS_CONTEXT
            wxGraphicsContext* gc = dc.GetGraphicsContext();
            if (gc)
            {
                gc->SetAntialiasMode(wxANTIALIAS_NONE);
            }
#endif // wxUSE_GRAPHICS_CONTEXT
```

Antialiasing should be disabled for 1bpp images

qt_after_1bpp_black

(1bpp, no AA, with wxGREY_BRUSH -> wxBLACK_BRUSH)

vadz:
> The test [...] fails because of the anti-aliased mask rectangle corners:

Thanks for looking into this, but what would be the solution for this?
 
> This code doesn't work because it doesn't use wxGraphicsContext.
> 
> ```c++
> #if wxUSE_GRAPHICS_CONTEXT
>             wxGraphicsContext* gc = dc.GetGraphicsContext();
>             if (gc)
>             {
>                 gc->SetAntialiasMode(wxANTIALIAS_NONE);
>             }
> #endif // wxUSE_GRAPHICS_CONTEXT
> ```

Err, sorry, why doesn't it use when it looks like it does? 

> Antialiasing should be disabled for 1bpp images

This seems logical, but, as always, I wonder if there is any possibility of breaking something by doing it?

dsa-t:
> Thanks for looking into this, but what would be the solution for this?

I've already disabled AA for 1bpp paint devices in latest commits. But the other test also seems to fail because AA needs to be disabled in `wxQtDCImpl` somehow. 

```
./graphics/clippingbox.cpp:2109

./graphics/clippingbox.cpp:262: FAILED:
explicitly with message:
  Invalid shape of the clipping region
```

> Err, sorry, why doesn't it use when it looks like it does?

`wxQtDCImpl` does not inherit from `wxGCDCImpl`, so `GetGraphicsContext()` always returns nulllptr.

I'm not sure why it doesn't though, and `src/qt/graphics.cpp` that provides `wxQtGraphicsRenderer` instead of Cairo is only built on Windows for some reason.

>This seems logical, but, as always, I wonder if there is any possibility of breaking something by doing it?

Antialiasing doesn't look right anyway in 1bpp mode, so should be fine.

vadz:
Apparently the plan was to use this code for all platforms, see #1139, but somehow this has never happened. If it works under the other platforms, I think we should indeed enable it -- have you been able to check if this is the case already, by chance?

dsa-t:
It works on my Linux distro, Qt prints some asserts though.

dsa-t:
Looks like the tests passed.

vadz:
Thanks for the update!

@AliKet Please let me know if you have any comments about changes here.

AliKet:
> I'm not sure why it doesn't though, and `src/qt/graphics.cpp` that provides `wxQtGraphicsRenderer` instead of Cairo is only built on Windows for some reason.
> 

You have to remove the `defined(__WIN32__)` check here https://github.com/wxWidgets/wxWidgets/blob/f24b3d54833a1c15ae3bba87dead83bf9a90c41d/src/generic/graphicc.cpp#L3695-L3702
to get `wxQtGraphicsRenderer` instead of Cairo.

I worked on this last year, but lost that work due to a hardware malfunction unfortunately. But one issue I noticed at the time (with other minor issues) was the clipping that needed to be fixed for the tests to pass.

AliKet:
> @AliKet Please let me know if you have any comments about changes here.

No comment.

AliKet:
@dsa-t Let me know if you have any plan to work on making `wxQtGraphicsRenderer` available under non-Windows too ?

dsa-t:
> @dsa-t Let me know if you have any plan to work on making `wxQtGraphicsRenderer` available under non-Windows too ?

Not in the near future.

vadz:
Sorry, one more question: why do we enable `Antialiasing` here but not both it and `TextAntialiasing`, as in `wxQtGraphicsContext::SetAntialiasMode()`?

Would it make sense to extract this into a (private) function to make it easier to change this later, if necessary?

dsa-t:
TextAntialiasing is already enabled by default.

dsa-t:
>Would it make sense to extract this into a (private) function to make it easier to change this later, if necessary?

Not sure what would be the right place to put this function.

vadz:
> > Would it make sense to extract this into a (private) function to make it easier to change this later, if necessary?
> 
> Not sure what would be the right place to put this function.

After looking more at wxQt code, I think I would add some (inline) `namespace wxQtImpl { QPainter* CreatePaint(); }` to `include/wx/qt/private/painter.h` and use it instead of `new QPainter()` to ensure that all painters we use are initialized in a consistent way. What do you think?

dsa-t:
Sounds ok to me, I probably won't have time to tackle it though.

vadz:
OK, let's merge it as is then and leave this as a possible future enhancement.

Thanks again!

Issue #22712, opened 2022-08-07
Handle unsplit state better in wxSplitterWindow and wxPersistentSplitter


vadz:
There are several problems with using `wxSplitterWindow` that can be dynamically split and unsplit currently:

1. Calling `Unsplit()` loses the current splitter position, as it resets it to 0 (actually to the minimum pane size, which makes even less sense), so if you want to dynamically split/unsplit a splitter you have to store this position in a separate variable in a program instead of just relying on the splitter itself having it.
2. Destroying an unsplit persistent splitter stores its position as `-1` and doesn't store the actual position at all.

Both problems could be fixed by preserving the position in the splitter even when it's unsplit and saving it in `wxPersistentSplitter` and also saving "unsplit" flag separately. Some compatibility could be provided by handling the position of `-1` if we read it from the existing config.

PR #24329, opened 2024-02-16
Save last sash position in wxSplitterWindow and wxPersistentSplitter

douderg:
Work on issue: #22712

Adds members to wxSplitterWindow for storing the last sash position of a horizontal/vertical split when Unsplit is called. This allows the previous sash position to be restored instead of resetting back to the default position when the window is split again.

Adds setter / getter for the new members which are used by wxPersistentSplitter to save and restore state.

### DIFF
--- a/include/wx/generic/splitter.h
+++ b/include/wx/generic/splitter.h
@@ -227,10 +227,10 @@ public:
     wxDEPRECATED_INLINE( void SetSashSize(int WXUNUSED(width)), return; )
 
     // Get the sash position that was last used before Unsplit
-    void GetDefaultSashPosition(int& horizontal, int& vertical) const;
+    const wxPoint& GetLastSplitPosition() const;
 
     // Set the default initial sash position to use when the splitter is split
-    void SetDefaultSashPosition(int horizontal, int vertical);
+    void SetLastSplitPosition(int horizontal, int vertical);
 
 protected:
     // event handlers
@@ -295,8 +295,7 @@ protected:
     wxPoint     m_ptStart;      // mouse position when dragging started
     int         m_sashStart;    // sash position when dragging started
     int         m_minimumPaneSize;
-    int         m_latestHorizontalSashPosition;
-    int         m_latestVerticalSashPosition;
+    wxPoint     m_lastSplitPosition;
     wxCursor    m_sashCursorWE;
     wxCursor    m_sashCursorNS;
     wxOverlay   m_overlay;
--- a/include/wx/generic/splitter.h
+++ b/include/wx/generic/splitter.h
@@ -226,6 +226,12 @@ public:
     // all any more.
     wxDEPRECATED_INLINE( void SetSashSize(int WXUNUSED(width)), return; )
 
+    // Get the sash position that was last used before Unsplit
+    void GetDefaultSashPosition(int& horizontal, int& vertical) const;
+
+    // Set the default initial sash position to use when the splitter is split
+    void SetDefaultSashPosition(int horizontal, int vertical);
+
 protected:
     // event handlers
 #if defined(__WXMSW__) || defined(__WXMAC__)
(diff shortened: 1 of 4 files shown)


include/wx/generic/splitter.h
```diff
@@ -289,6 +295,8 @@ class WXDLLIMPEXP_CORE wxSplitterWindow: public wxNavigationEnabled<wxWindow>
     wxPoint     m_ptStart;      // mouse position when dragging started
     int         m_sashStart;    // sash position when dragging started
     int         m_minimumPaneSize;
+    int         m_latestHorizontalSashPosition;
+    int         m_latestVerticalSashPosition;
```
vadz, line 299: As mentioned above, I'd use a single `wxPoint` for those and call it `m_lastSplitPosition`.

include/wx/persist/splitter.h
```diff
@@ -23,6 +23,9 @@
 // Special position value of -1 means the splitter is not split at all.
 #define wxPERSIST_SPLITTER_POSITION wxASCII_STR("Position")
 
+#define wxPERSIST_SPLITTER_DEFAULT_HORIZONTAL wxASCII_STR("Horizontal")
+#define wxPERSIST_SPLITTER_DEFAULT_VERTICAL wxASCII_STR("Vertical")
```
vadz, line 27: To be consistent with the naming schema suggested above I'd also use
```suggestion
#define wxPERSIST_SPLITTER_DEFAULT_HORIZONTAL wxASCII_STR("LastHorz")
#define wxPERSIST_SPLITTER_DEFAULT_VERTICAL wxASCII_STR("LastVert")
```
here.

src/generic/splitter.cpp
```diff
@@ -840,6 +843,12 @@ int wxSplitterWindow::ConvertSashPosition(int sashPosition) const
     }
     else // sashPosition == 0
     {
+        int defaultPos = GetWindowSize() / 2;
+        if (m_splitMode == wxSPLIT_VERTICAL) {
+            return m_latestVerticalSashPosition != 0 ? m_latestVerticalSashPosition : defaultPos;
+        } else if (m_splitMode == wxSPLIT_HORIZONTAL) {
+            return m_latestHorizontalSashPosition != 0 ? m_latestHorizontalSashPosition : defaultPos;
+        }
```
vadz, line 851: Minor style fix: please put braces on their own lines for consistency, i.e.
```suggestion
        if (m_splitMode == wxSPLIT_VERTICAL)
        {
            return m_lastSplitPosition.y != 0 ? m_lastSplitPosition.y : defaultPos;
        }
        else if (m_splitMode == wxSPLIT_HORIZONTAL)
        {
            return m_lastSplitPosition.x != 0 ? m_lastSplitPosition.x : defaultPos;
        }
```

src/generic/splitter.cpp
```diff
@@ -871,6 +880,12 @@ bool wxSplitterWindow::Unsplit(wxWindow *toRemove)
         return false;
     }
 
+    if (m_splitMode == wxSPLIT_VERTICAL) {
+        m_latestVerticalSashPosition = m_sashPosition;
+    } else if (m_splitMode == wxSPLIT_HORIZONTAL) {
+        m_latestHorizontalSashPosition = m_sashPosition;
+    }
+
```
vadz, line 888: Same style comment as above:
```suggestion
    if (m_splitMode == wxSPLIT_VERTICAL)
    {
        m_lastSplitPosition.y = m_sashPosition;
    }
    else if (m_splitMode == wxSPLIT_HORIZONTAL)
    {
        m_lastSplitPosition.x = m_sashPosition;
    }

```

vadz (commented):
Thank you for your work!

In spite of the number of comments, this looks good globally and all these comments are pretty minor. But it would be still nice to address them.

And please add the new functions to `interface/wx/splitter.h`, unless you think we really shouldn't export them for some reason.

TIA!

vadz:
I'm merging it now but I've done a couple of minor tweaks to the code before doing it:

1. Take `wxPoint` in `SetLastSplitPosition()` to make it consistent with returning it from `GetLastSplitPosition()`.
2. Don't bother saving last position if it's (0,0) and, more importantly, don't fail when restoring if it wasn't found: this would have resulted in not restoring the splitter position of all the existing applications using wxWidgets versions until now, as they wouldn't have these positions in their config files/registry and so wouldn't be user-friendly.
3. More minor style fixes.

Please let me know if you have any objections to anything I've done and thanks again for your contribution!

PR #24262, opened 2024-01-28
3.2.5 proposed backports

vadz:

PR #24332, opened 2024-02-16
Don't reset selection in `wxGenericListCtrl` while Ctrl or Shift key is pressed

taler21:
The native MSW `wxListCtrl` doesn't reset the selection when the user clicks outside of any item while holding the Ctrl or Shift key.
Since the generic `wxListCtrl` mimics the MSW behaviour, it should behave the same way here.

### DIFF
--- a/src/generic/listctrl.cpp
+++ b/src/generic/listctrl.cpp
@@ -2549,12 +2549,17 @@ void wxListMainWindow::OnMouse( wxMouseEvent &event )
         }
         else if (event.LeftDown())
         {
-            // reset the selection and bail out
-            HighlightAll(false);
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, I agree with this change and think that it makes sense on its own, even without the MSW-compatibility argument -- GTK native control behaves like this too.

taler21:
Do you think this improvement can also be backported to 3.2?
If so, I would kindly ask you to do so. Thanks in advance!

vadz:
Yes, I had actually already cherry picked it into my backports branch some time ago, I've updated #24262 with it now.

taler21:
Great, thank you very much!

Issue #24333, opened 2024-02-16
New wxRibbonPage::GetPanelCount() method documented as 'const' but implemented 'non-const'


mcorino:
### Description

@vadz  @Blake-Madden The new `wxRibbonPage::GetPanelCount()` method  added a few days ago is documented as const in 'interface/wx/ribbon/page.h' but implemented as non-const.

Most likely the implementation should be updated but I'm not sure as I did not inspect the code in detail.

### Platform and version information

- wxWidgets version  you use: master
- wxWidgets port  you use: NA
- OS  and its version: NA

PR #24334, opened 2024-02-16
Make GetPanelCount() const

Blake-Madden:
Fix for #24333


vadz:
Thanks for the quick fix, merging.

PR #24335, opened 2024-02-17
Restore columns' position for wxDataViewCtrl

vadimgrn:
There is no API to change columns order for wxDataViewCtrl. For that reason <wx\persist\dataview.h> saves positions, but does not restore them. This a workaround to restore columns' position. It works and is better than nothing.

### DIFF
--- a/include/wx/persist/dataview.h
+++ b/include/wx/persist/dataview.h
@@ -15,6 +15,7 @@
 #if wxUSE_DATAVIEWCTRL
 
 #include "wx/dataview.h"
+#include <map>
 
 // ----------------------------------------------------------------------------
 // String constants used by wxPersistentDataViewCtrl.
@@ -94,6 +95,7 @@ public:
 
     virtual bool Restore() override
     {
+        std::map<int, wxDataViewColumn*> order;
         wxDataViewCtrl* const control = Get();
 
         for ( unsigned int col = 0; col < control->GetColumnCount(); col++ )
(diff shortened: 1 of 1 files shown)


include/wx/persist/dataview.h
```diff
@@ -94,6 +95,7 @@ class wxPersistentDataViewCtrl : public wxPersistentWindow<wxDataViewCtrl>
 
     virtual bool Restore() override
     {
+        std::map<int, wxDataViewColumn*> order;
```
vadz, line 98: IMO the use of `map` here is confusing because we never actually use its key. Why not just use `std::vector<wxDataViewColumn*>` instead.

include/wx/persist/dataview.h
```diff
@@ -115,7 +117,21 @@ class wxPersistentDataViewCtrl : public wxPersistentWindow<wxDataViewCtrl>
...
                 column->SetWidth(width);
 
-            // TODO: Set the column's view position.
+            // Retrieve column's view position.
+            int pos;
+            if ( RestoreValue(columnPrefix + wxASCII_STR(wxPERSIST_DVC_POS), &pos) )
+                order.emplace(pos, column);
+        }
+
+        // Restore columns' position.
```
vadz, line 126:
```suggestion
        // Restore columns' positions.
```

include/wx/persist/dataview.h
```diff
@@ -115,7 +117,21 @@ class wxPersistentDataViewCtrl : public wxPersistentWindow<wxDataViewCtrl>
...
+        int new_pos = 0; // do not trust saved values from registry
```
vadz, line 127: The values are not necessarily stored in the registry, so I'd say something like "stored values". But this comment would probably be unnecessary if we used vector instead of map anyhow.

vadz (commented):
Sorry, I find the use of `map` here pretty confusing, i.e. I _think_ that this code does the right thing but it's not really obvious.

include/wx/persist/dataview.h
vadimgrn: Map is required for implicit sorting, there's no wxMap. Vector give you unsorted sequence during the iteration or it must be sorted.

vadimgrn:
Done

PR #24337, opened 2024-02-18
Add translucent stipple brush support under wxQt

AliKet:
This is to support **venetian blinds bitmap** creation in `wxAUI`

Screenshot from 2024-02-18 18-24-17


vadz:
Thanks! Will merge too soon.

PR #24340, opened 2024-02-19
Expose reason for notification dismissal through SetInt()

lanurmi:
On all platforms, except macOS, the reason why a notification was dismissed is known. Specifically, it can be useful to distinguish whether the notification closed because of a timeout, or a click by the user.

This is left undocumented on purpose for now. The values passed are platform-dependent, and this is not the most elegant API.

### DIFF
--- a/include/wx/notifmsg.h
+++ b/include/wx/notifmsg.h
@@ -81,6 +81,15 @@ public:
         Timeout_Never = 0   // notification will never time out
     };
 
+    // reasons for dismissal, posted with wxEVT_NOTIFICATION_MESSAGE_DISMISSED events
+    enum class DismissalReason
+    {
+        Unknown,
+        ByUser,
+        ByApp,
+        TimedOut
+    };
+
     // show the notification to the user and hides it after timeout seconds
     // pass (special values Timeout_Auto and Timeout_Never can be used)
     //
--- a/interface/wx/notifmsg.h
+++ b/interface/wx/notifmsg.h
@@ -44,7 +44,10 @@
            is clicked.
     @event{EVT_NOTIFICATION_MESSAGE_DISMISSED(id, func)}
            Process a @c wxEVT_NOTIFICATION_MESSAGE_DISMISSED event, when a notification
-           is dismissed by the user or times out.
+           is dismissed by the user or times out. Since wxWidgets 3.3.0, on platforms
+           supporting it, the reason for dismissal can be obtained from
+           wxCommandEvent::GetInt(), with values corresponding to those of
+           wxNotificationMessage::DismissalReason.
     @event{EVT_NOTIFICATION_MESSAGE_ACTION(id, func)}
            Process a @c wxEVT_NOTIFICATION_MESSAGE_ACTION event, when the user
            selects on of the actions added by AddAction()
@@ -62,6 +65,19 @@ public:
     {
         Timeout_Auto = -1,  ///< Notification will be hidden automatically.
         Timeout_Never = 0   ///< Notification will never time out.
+    };
+
+    /**
+      Reasons for dismissal, posted with wxEVT_NOTIFICATION_MESSAGE_DISMISSED events.
+
+      @since 3.3.0
+    */
+    enum class DismissalReason
+    {
+        Unknown,            ///< Reason unknown, possibly unsupported on this platform.
+        ByUser,             ///< The user dismissed the notification.
+        ByApp,              ///< The application itself closed the notification.
+        TimedOut            ///< The notification expired after reaching its timeout.
     };
 
     /**
--- a/src/generic/notifmsgg.cpp
+++ b/src/generic/notifmsgg.cpp
@@ -265,6 +265,8 @@ void wxNotificationMessageWindow::Set(int timeout)
 void wxNotificationMessageWindow::OnClose(wxCloseEvent& WXUNUSED(event))
 {
     wxCommandEvent evt(wxEVT_NOTIFICATION_MESSAGE_DISMISSED);
+    evt.SetInt(static_cast<int>(wxNotificationMessage::DismissalReason::ByApp));
+
     m_notificationImpl->ProcessNotificationEvent(evt);
 
     if ( m_timer.IsRunning() )
@@ -307,6 +309,7 @@ void wxNotificationMessageWindow::OnNotificationMouseLeave(wxMouseEvent& WXUNUSE
 void wxNotificationMessageWindow::OnCloseClicked(wxCommandEvent& WXUNUSED(event))
 {
     wxCommandEvent evt(wxEVT_NOTIFICATION_MESSAGE_DISMISSED);
+    evt.SetInt(static_cast<int>(wxNotificationMessage::DismissalReason::ByUser));
     m_notificationImpl->ProcessNotificationEvent(evt);
 
     m_notificationImpl->Close();
--- a/src/gtk/notifmsg.cpp
+++ b/src/gtk/notifmsg.cpp
@@ -312,15 +312,33 @@ public:
         // Values according to the OpenDesktop specification at:
         // https://developer.gnome.org/notification-spec/
 
+        auto reason = wxNotificationMessage::DismissalReason::Unknown;
         switch (closeReason)
         {
-            case 1: // Expired
-            case 2: // The notification was dismissed by the user.
+            case 1: // "The notification expired."
+                reason = wxNotificationMessage::DismissalReason::TimedOut;
+                break;
+            case 2: // "The notification was dismissed by the user."
+                reason = wxNotificationMessage::DismissalReason::ByUser;
+                break;
+            case 3: // "The notification was closed by a call to CloseNotification."
+                reason = wxNotificationMessage::DismissalReason::ByApp;
+                break;
+        }
+
+        switch (reason)
+        {
+            case wxNotificationMessage::DismissalReason::TimedOut:
+            case wxNotificationMessage::DismissalReason::ByUser:
             {
                 wxCommandEvent evt(wxEVT_NOTIFICATION_MESSAGE_DISMISSED);
+                evt.SetInt(static_cast<int>(reason));
                 ProcessNotificationEvent(evt);
                 break;
             }
+            case wxNotificationMessage::DismissalReason::ByApp:
+            case wxNotificationMessage::DismissalReason::Unknown:
+                break;
         }
 
     }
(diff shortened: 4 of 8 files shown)


vadz (commented):
Thanks, this can definitely be useful, but we need to name and document these constants instead of just using `2` in the code.

I guess a (scoped) enum could be defined in `wxNotificationMessage` itself?

lanurmi:
> Thanks, this can definitely be useful, but we need to name and document these constants instead of just using `2` in the code.

We can do that, yes. There is just the MSW value `ToastDismissalReason_ApplicationHidden` that does not correspond to anything that the Gtk implementation would currently ever return. But at a closer look it does correspond to reason 3 in the Desktop Notifications Specification, so there is no fundamental mismatch of reason meanings across platforms.

> I guess a (scoped) enum could be defined in `wxNotificationMessage` itself?

I'll do that.

lanurmi:
Done with modifications. Since `SetInt()` is not used for anything yet, might as well use that instead of `SetExtraLong()`. The documentation assumes this feature will be available since 3.2.5.

All commits should probably be squashed into one.

src/generic/notifmsgg.cpp
```diff
@@ -265,6 +265,7 @@ void wxNotificationMessageWindow::Set(int timeout)
 void wxNotificationMessageWindow::OnClose(wxCloseEvent& WXUNUSED(event))
 {
     wxCommandEvent evt(wxEVT_NOTIFICATION_MESSAGE_DISMISSED);
+    evt.SetExtraLong(0);
```
vadz, line 268: This should use the symbolic name now (even though it would require a cast).

src/generic/notifmsgg.cpp
```diff
@@ -307,6 +308,7 @@ void wxNotificationMessageWindow::OnNotificationMouseLeave(wxMouseEvent& WXUNUSE
 void wxNotificationMessageWindow::OnCloseClicked(wxCommandEvent& WXUNUSED(event))
 {
     wxCommandEvent evt(wxEVT_NOTIFICATION_MESSAGE_DISMISSED);
+    evt.SetExtraLong(2);
```
vadz, line 311: Same as above.

src/gtk/notifmsg.cpp
```diff
@@ -312,12 +312,28 @@ class wxLibNotifyMsgImpl : public wxNotificationMessageImpl
...
+            break;
+        case 2: // "The notification was dismissed by the user."
+            reason = wxNotificationMessage::DismissalReason::DismissedByUser;
+            break;
+        case 3: // "The notification was closed by a call to CloseNotification."
+            reason = wxNotificationMessage::DismissalReason::DismissedByApp;
+            break;
+        }
+
         switch (closeReason)
```
vadz, line 330: Sorry, it doesn't make to have 2 `switch (closeReason)` to me in a row. This should probably be changed to `switch (reason)` now?

interface/wx/notifmsg.h
```diff
@@ -64,6 +67,15 @@ class wxNotificationMessage : public wxEvtHandler
         Timeout_Never = 0   ///< Notification will never time out.
     };
 
+    /// Reasons for dismissal, posted with wxEVT_NOTIFICATION_MESSAGE_DISMISSED events.
```
vadz, line 70: Although trivial, we probably still want to do this for consistency:

```suggestion
    /**
      Reasons for dismissal, posted with wxEVT_NOTIFICATION_MESSAGE_DISMISSED events.
      
      @since 3.2.5
    */
```

include/wx/notifmsg.h
```diff
@@ -81,6 +81,15 @@ class WXDLLIMPEXP_CORE wxNotificationMessageBase : public wxEvtHandler
         Timeout_Never = 0   // notification will never time out
     };
 
+    // reasons for dismissal, posted with wxEVT_NOTIFICATION_MESSAGE_DISMISSED events
+    enum class DismissalReason
+    {
+        Unknown,
+        DismissedByUser,
+        DismissedByApp,
```
vadz, line 89: Using `DismissalReason::DismissedByUser` in the code looks a bit heavy, maybe we should call this just `ByUser`/`ByApp`?

vadz (commented):
Thanks for the update!

Is it important to have this in 3.2.5? We don't require C++11 there, so we can't use a scoped enum and would have to replace it with a weak one.

src/gtk/notifmsg.cpp
lanurmi: It makes some sense, as the cases aren't the same in both. But yes, `switch (reason)` for the latter would be better.

include/wx/notifmsg.h
lanurmi: Sounds good.

src/generic/notifmsgg.cpp
lanurmi: Looks like I forgot to modify under `generic/`.

lanurmi:
> Is it important to have this in 3.2.5? We don't require C++11 there, so we can't use a scoped enum and would have to replace it with a weak one.

Not terribly important for me personally, but would be nice to get such simple new features into Linux distributions and such within this decade. There is also the API-forward-compatible approach of using just numerical values in 3.2.x, matching the values of the scoped enum. And only introduce the the scoped enum in 3.3.

lanurmi:
I have now altered the since versions to 3.3.0, as backporting to 3.2.x is not certain yet.

src/gtk/notifmsg.cpp
```diff
@@ -312,15 +312,33 @@ class wxLibNotifyMsgImpl : public wxNotificationMessageImpl
...
+        {
+            case wxNotificationMessage::DismissalReason::TimedOut:
+            case wxNotificationMessage::DismissalReason::ByUser:
             {
                 wxCommandEvent evt(wxEVT_NOTIFICATION_MESSAGE_DISMISSED);
+                evt.SetInt(static_cast<int>(reason));
                 ProcessNotificationEvent(evt);
                 break;
             }
+            default:
```
vadz, line 340: Very minor, but could you please explicitly list the 2 other cases here instead of using `default:`? This would ensure that we get a warning from the compiler if we ever add another enum element and forget to handle it here.

include/wx/notifmsg.h
```diff
@@ -81,6 +81,15 @@ class WXDLLIMPEXP_CORE wxNotificationMessageBase : public wxEvtHandler
         Timeout_Never = 0   // notification will never time out
     };
 
+    // reasons for dismissal, posted with wxEVT_NOTIFICATION_MESSAGE_DISMISSED events
+    enum class DismissalReason
```
vadz, line 85: If we want to backport this to 3.2, we'd need to make this plain (unscoped) enum, e.g.

```cpp
enum DismissalReason
{
  Reason_Unknown,
  ...
};
```

vadz (commented):
Thanks for the update!

I have just one question remaining before merging this into master: are we supposed to get notifications for `ByApp` reason or not? It looks to me (but I didn't test it) that GTK version avoids sending those, but MSW one seems to do it. Shouldn't the latter one avoid them too?

And, as already mentioned, if we/you want to have this in 3.2, we should avoid using C++11 scoped enums. But it's perfectly fine to have this enum for 3.3 only, of course.

lanurmi:
> Thanks for the update!
> 
> I have just one question remaining before merging this into master: are we supposed to get notifications for `ByApp` reason or not? It looks to me (but I didn't test it) that GTK version avoids sending those, but MSW one seems to do it. Shouldn't the latter one avoid them too?

I have no definite answer to this. I have been wondering why the GTK version avoids sending them (maybe there is a reason, but it is undocumented). So I figured changing existing behavior is out of scope for this PR, and chose to let both GTK and MSW keep doing/not doing what they were before.

I have pushed the latest requested change to `default:`, along with some indentation adjustments and two `auto` keywords.

> And, as already mentioned, if we/you want to have this in 3.2, we should avoid using C++11 scoped enums. But it's perfectly fine to have this enum for 3.3 only, of course.

I wouldn't want to be limited by C++98 in 3.3, so let us stick to the scoped enum. The backporting options would be:
1. Just not to introduce any `enum` in 3.2.
2. Or maybe:
```c++
class wxNotificationMessage {
   ...
   struct DismissalReason { enum { Unknown, ByUser, ByApp, TimedOut } };
```
This would be _mostly_ source code compatible with how a scoped enum is used. I'm not 100% sure if adding a `struct` into a `class` changes ABI, even though there is no instance of the `struct`.

vadz:
> > Thanks for the update!

(And again)

> > I have just one question remaining before merging this into master: are we supposed to get notifications for `ByApp` reason or not? It looks to me (but I didn't test it) that GTK version avoids sending those, but MSW one seems to do it. Shouldn't the latter one avoid them too?
> 
> I have no definite answer to this. I have been wondering why the GTK version avoids sending them (maybe there is a reason, but it is undocumented). So I figured changing existing behavior is out of scope for this PR, and chose to let both GTK and MSW keep doing/not doing what they were before.

I'd like to at least decide if we're going to standardize on the same behaviour in the future (as I hope) and, if so, whether we even need `ByApp` dismissal reason: if dismissing the notification by the application itself shouldn't produce any notification, it doesn't seem useful to introduce it.

> I wouldn't want to be limited by C++98 in 3.3, so let us stick to the scoped enum. The backporting options would be:
> 
>     1. Just not to introduce any `enum` in 3.2.
> 
>     2. Or maybe:
> 
> 
> ```c++
> class wxNotificationMessage {
>    ...
>    struct DismissalReason { enum { Unknown, ByUser, ByApp, TimedOut } };
> ```
> 
> This would be _mostly_ source code compatible with how a scoped enum is used.

Yes, we could do this...

> I'm not 100% sure if adding a `struct` into a `class` changes ABI, even though there is no instance of the `struct`.

Adding things is, in general, ABI-compatible.

lanurmi:
> > > I have just one question remaining before merging this into master: are we supposed to get notifications for `ByApp` reason or not? 
> I'd like to at least decide if we're going to standardize on the same behaviour in the future (as I hope) and, if so, whether we even need `ByApp` dismissal reason: if dismissing the notification by the application itself shouldn't produce any notification, it doesn't seem useful to introduce it.

I have added a separate commit that changes GTK's behavior so that it sends the event with all reason values, not just the two.

vadz:
> I have added a separate commit that changes GTK's behavior so that it sends the event with all reason values, not just the two.

Thanks! I'll merge this soon.

PR #24341, opened 2024-02-19
Fix toast notification always posting the click event on MSW

### LINKED ISSUE #24320: wxNotificationMessage always sends click event on MSW with toasts
#### Bug description:

When using `wxNotificationMessage`s on MSW with `MSWUseToasts()` enabled (like the dialogs sample does), clicking the notification or letting it timeout will _both_ result in `wxEVT_NOTIFICATION_MESSAGE_CLICK` being sent.

It is expected that `wxEVT_NOTIFICATION_MESSAGE_DISMISSED` would be sent in case of timeout or explicitly dismissing the notification.

#### To Reproduce:

1. Start 'dialogs' sample on MSW.
2. Open User Notification Test Dialog.
3. Click 'Show'.
4. Observe that whatever you do – click or no click –, the status text will soon show "Notification was clicked".

#### But why?

It looks like regardless of what the user actually did or didn't do, [the overload of `Invoke` taking `IInspectable*` gets called](https://github.com/wxWidgets/wxWidgets/blob/master/src/msw/rt/notifmsgrt.cpp#L420-L431), and that one sends the click event. (`IInspectable` is the base class for `IToast{Dismissed,Failed}EventArgs`, so it makes some sense.) The debugger shows that the contents of the `args` object is actually an `IToastDismissedEventArgs`.

- OS  and its version: Windows 10.0.19045.3996

lanurmi:
No idea why this is needed, but would seem to fix #24320.

### DIFF
--- a/src/msw/rt/notifmsgrt.cpp
+++ b/src/msw/rt/notifmsgrt.cpp
@@ -418,13 +418,31 @@ wxString wxToastNotifMsgImpl::ms_appId;
 wxCOMPtr<IToastNotificationManagerStatics> wxToastNotifMsgImpl::ms_toastMgr;
 
 HRESULT wxToastEventHandler::Invoke(
-    IToastNotification *WXUNUSED(sender),
-    IInspectable *WXUNUSED(args))
+    IToastNotification *sender,
+    IInspectable *args)
 {
     if ( m_impl )
     {
-        wxCommandEvent evt(wxEVT_NOTIFICATION_MESSAGE_CLICK);
-        m_impl->ProcessNotificationEvent(evt);
+        IToastDismissedEventArgs *dea = nullptr;
+        IToastFailedEventArgs *fea = nullptr;
+
+        if ( SUCCEEDED(args->QueryInterface(__uuidof(IToastDismissedEventArgs), (void**)&dea)) )
+        {
+            // FIXME: We have for some unknown reason ended up in this function even though
+            // args is really a IToastDismissedEventArgs*. Call the actual implementation
+            // we wanted.
+            return Invoke(sender, dea);
+        }
+        else if ( SUCCEEDED(args->QueryInterface(__uuidof(IToastFailedEventArgs), (void**)&fea)) )
+        {
+            // FIXME: Same as above.
+            return Invoke(sender, fea);
+        }
+        else
+        {
+            wxCommandEvent evt(wxEVT_NOTIFICATION_MESSAGE_CLICK);
+            m_impl->ProcessNotificationEvent(evt);
+        }
     }
 
     return S_OK;


src/msw/rt/notifmsgrt.cpp
```diff
@@ -418,13 +418,31 @@ wxString wxToastNotifMsgImpl::ms_appId;
...
+    IInspectable *args)
 {
     if ( m_impl )
     {
-        wxCommandEvent evt(wxEVT_NOTIFICATION_MESSAGE_CLICK);
-        m_impl->ProcessNotificationEvent(evt);
+        IToastDismissedEventArgs *dea = nullptr;
+        IToastFailedEventArgs *fea = nullptr;
+
+        if ( SUCCEEDED(args->QueryInterface(__uuidof(IToastDismissedEventArgs), (void**)&dea)) )
```
vadz, line 429: We need to call `Release()` on `dea` in case of success to avoid memory leaks (and the same for `fea` below, of course).

vadz (changes requested):
I think the correct thing to do would be to use 3 different handlers, implementing `DesktopToastActivatedEventHandler`, `DesktopToastDismissedEventHandler` and  `DesktopToastFailedEventHandler` respectively instead of a single one.

This should be straightforward to do, AFAICS and we actually only need 2 of them as the third one would do nothing anyhow.

Could you please try to do it like this? TIA!

lanurmi:
> I think the correct thing to do would be to use 3 different handlers, implementing `DesktopToastActivatedEventHandler`, `DesktopToastDismissedEventHandler` and `DesktopToastFailedEventHandler` respectively instead of a single one.

Do you mean doing it [the way it is already done in master](https://github.com/wxWidgets/wxWidgets/blob/master/src/msw/rt/notifmsgrt.cpp#L63-L70)? The thing is that it fails to work correctly, at least not for me. The overload of `Invoke` taking `IInspectable*` always gets called.

Regardless, some unanswered questions are:
* Does the current master do the wrong thing for others besides me?
* Did the current implementation work correctly at the time it was made?
* Is this compiler-specific?

vadz:
> > I think the correct thing to do would be to use 3 different handlers, implementing `DesktopToastActivatedEventHandler`, `DesktopToastDismissedEventHandler` and `DesktopToastFailedEventHandler` respectively instead of a single one.
> 
> Do you mean doing it [the way it is already done in master](https://github.com/wxWidgets/wxWidgets/blob/master/src/msw/rt/notifmsgrt.cpp#L63-L70)?

No, this code uses a single wxToastEventHandler inheriting from
`Microsoft::WRL::Implements<DesktopToastActivatedEventHandler, DesktopToastDismissedEventHandler, DesktopToastFailedEventHandler>`, while I think we should use 3 (or 2) different wx classes, i.e. `wxToastActivatedEventHandler` and `wxToastDismissedEventHandler`, inheriting/implementing just the single corresponding base class.

> Regardless, some unanswered questions are:
> 
>     * Does the current master do the wrong thing for others besides me?

I didn't test it, but I think it always fails.
 
>     * Did the current implementation work correctly at the time it was made?
> 
>     * Is this compiler-specific?

I don't think so.

vadz:
Sorry, I should have realized that it was yet another problem with broken COM helpers not supporting multiple interfaces. The real fix is in #24826, please check that it works for you but I'm pretty sure it should.

PR #24342, opened 2024-02-19
macOS deactivate automatic use of native scrollview

csomor:
see #24201, alternative for #24301

### DIFF
--- a/include/wx/osx/cocoa/private.h
+++ b/include/wx/osx/cocoa/private.h
@@ -221,7 +221,7 @@ public :
     virtual void                controlTextDidChange();
 
     virtual void                AdjustClippingView(wxScrollBar* horizontal, wxScrollBar* vertical) override;
-    virtual void                UseClippingView(bool forceClippingViewCreation) override;
+    virtual void                UseClippingView() override;
     virtual WXWidget            GetContainer() const override { return m_osxClipView ? m_osxClipView : m_osxView; }
 
 protected:
--- a/include/wx/osx/cocoa/private.h
+++ b/include/wx/osx/cocoa/private.h
@@ -221,7 +221,7 @@ public :
     virtual void                controlTextDidChange();
 
     virtual void                AdjustClippingView(wxScrollBar* horizontal, wxScrollBar* vertical) override;
-    virtual void                UseClippingView() override;
+    virtual void                UseClippingView(bool forceClippingViewCreation) override;
     virtual WXWidget            GetContainer() const override { return m_osxClipView ? m_osxClipView : m_osxView; }
 
 protected:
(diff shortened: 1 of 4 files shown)


vadz:
Sorry if I'm missing something, but it looks like `UseClippingView(false)` does nothing at all, so why are we calling it? I understand the desire to use `UseClippingView(true)` at some time later, but why not just remove the calls to `UseClippingView()` now and add them later to `wxScrolledWindow` code instead?

We can merge this if you prefer, of course, but I have to say I find it very confusing to call a function just to do nothing in it.

csomor:
> Sorry if I'm missing something, but it looks like `UseClippingView(false)` does nothing at all, so why are we calling it? I understand the desire to use `UseClippingView(true)` at some time later, but why not just remove the calls to `UseClippingView()` now and add them later to `wxScrolledWindow` code instead?
> 
> We can merge this if you prefer, of course, but I have to say I find it very confusing to call a function just to do nothing in it.

right now it does nothing at all, but the idea is to find out whether in the non forced case I can setup a heuristic to determine what the best action is. It would be great if I get this to work, and find out how to deal with the scroll request properly, it might be that I'll have to find out about the true type of the native control to find out whether and what's really needed. 

The only situation where we are sure we need it, is the `wxScrolledWindow` case (and only when built against 14 SDK we must do something)

vadz:
Right, but why not just remove the existing calls to it instead of adding bool parameter which makes it do nothing?

csomor:
> Right, but why not just remove the existing calls to it instead of adding bool parameter which makes it do nothing?

because the native scroll view might also be needed later for some controls but not for others, because the bleeding problem of native subcontrols can happen again, and instead of removing those call now and adding the same call later again, I thought I'd just deactivate it now and then add a better heuristic within ...

vadz:
Sorry, I still don't understand at all how is it better to make a parameter to make the function do nothing instead of not adding and just not calling a function. If it needs to be called conditionally, why not just check for the condition in the caller and only call it if we must?

Anyhow, if you really don't see any problem with this, let's merge it, but I'll at least add some comments then because otherwise it risks being very, very frustrating to the next person trying to understand this code (quite probably future me, when I'll have forgotten everything about the explanations here).

csomor:
The places where it's called now I think are correct. I'll give it a try to find out whether I can find a good heuristic so that the outside hint by using the force attribute is not needed anymore and I could get rid of the param.

csomor:
@vadz I think I found the reason, scrolled windows with no scrollbars at all were not adjusting their native scrollview correctly

vadz:
Great, thanks, this change makes much more sense to me! I'll (squash) merge it soon as it surely shouldn't make anything worse.

Thanks a lot once again!

PR #24352, opened 2024-02-24
Fix wxSecretStore for linux

### LINKED ISSUE #24351: wxSecretStore is broken for binary secrets with WXGTK (probably any Linux platform).
#### Bug description:

The current gtk specific implementation of the wxSecretValueImpl class is broken for secrets containing  non-textual, binary data.
In case such a secret is used to store a secret the returned data after loading the secret will contain altered binary sequences in place of the original.

For instance; a secret containing "secret\x01\x86\x02\x99\x03" (size 11) was returned with 4 extra bytes (total size 15) inserted in the binary sequence following the 'secret' part.

The problem is caused by the content type 'text/plain' specified for the secret at line 79 of src/unix/secretstore.cpp.
The comment describing the reason for this:

```
    // Notice that we have to use text/plain as content type and not
    // application/octet-stream which would have been more logical because
    // libsecret accepts only valid UTF-8 strings for the latter, while our
    // data is not necessarily UTF-8 (nor even text at all...).
```

may have held truth at some point in the past but is does not now.

Identical secret returned vs stored.

#### To Reproduce:

The attached extended secretstore sample can be used to reproduce the problem. This sample recognizes two additional optional switches for the 'selftest' operation:

- '-b' to specify running the selftest with binary secrets
- '-v' to specify verbose output on selftest errors (will print the stored and loaded secrets)

[secretstore.tar.gz](https://github.com/wxWidgets/wxWidgets/files/14392961/secretstore.tar.gz)

Replace the standard sample with the file from the attached archive, build and run.

- OS  and its version: OpenSuSE Leap 15.5
  
  + GTK version:  3.24.5
  + Desktop environment : KDE

mcorino:
Fixes #24351

### DIFF
--- a/include/wx/private/secretstore.h
+++ b/include/wx/private/secretstore.h
@@ -46,6 +46,13 @@ public:
         memcpy(m_data, data, size);
     }
 
+    wxSecretValueGenericImpl(const wxScopedCharBuffer& buf)
+        : m_size(buf.length()),
+          m_data(new char[buf.length()])
+    {
+      memcpy(m_data, data, size);
+    }
+
     virtual ~wxSecretValueGenericImpl()
     {
         if ( m_data )
--- a/include/wx/private/secretstore.h
+++ b/include/wx/private/secretstore.h
@@ -50,7 +50,7 @@ public:
         : m_size(buf.length()),
           m_data(new char[buf.length()])
     {
-      memcpy(m_data, data, size);
+      memcpy(m_data, buf.data(), buf.length());
     }
 
     virtual ~wxSecretValueGenericImpl()
--- a/include/wx/private/secretstore.h
+++ b/include/wx/private/secretstore.h
@@ -46,13 +46,6 @@ public:
         memcpy(m_data, data, size);
     }
 
-    wxSecretValueGenericImpl(const wxScopedCharBuffer& buf)
-        : m_size(buf.length()),
-          m_data(new char[buf.length()])
-    {
-      memcpy(m_data, buf.data(), buf.length());
-    }
-
     virtual ~wxSecretValueGenericImpl()
     {
         if ( m_data )
--- a/include/wx/secretstore.h
+++ b/include/wx/secretstore.h
@@ -40,14 +40,15 @@ public:
 
     // Creates a secret value from the given data.
     wxSecretValue(size_t size, const void *data)
-        : m_impl(NewImpl(size, data))
+        : m_impl(NewImpl(size, data, "application/octet-stream"))
     {
     }
 
     // Creates a secret value from string.
     explicit wxSecretValue(const wxString& secret)
-        : m_impl(NewImpl(wxScopedCharBuffer(secret.utf8_str())))
     {
+      const wxScopedCharBuffer buf(secret.utf8_str());
+      m_impl = NewImpl(buf.length(), buf.data(), "text/plain");
     }
 
     wxSecretValue(const wxSecretValue& other);
@@ -87,10 +88,9 @@ public:
     static void WipeString(wxString& str);
 
 private:
-    // These methods are implemented in platform-specific code and must return a
+    // This method is implemented in platform-specific code and must return a
     // new heap-allocated object initialized with the given data.
-    static wxSecretValueImpl* NewImpl(size_t size, const void *data);
-    static wxSecretValueImpl* NewImpl(const wxScopedCharBuffer &buf);
+    static wxSecretValueImpl* NewImpl(size_t size, const void *data, const char* content_type);
 
     // This ctor is only used by wxSecretStore and takes ownership of the
     // provided existing impl pointer.
--- a/include/wx/secretstore.h
+++ b/include/wx/secretstore.h
@@ -46,9 +46,8 @@ public:
 
     // Creates a secret value from string.
     explicit wxSecretValue(const wxString& secret)
+        : m_impl(NewImpl(wxScopedCharBuffer(secret.utf8_str())))
     {
-        const wxScopedCharBuffer buf(secret.utf8_str());
-        m_impl = NewImpl(buf.length(), buf.data());
     }
 
     wxSecretValue(const wxSecretValue& other);
@@ -88,9 +87,10 @@ public:
     static void WipeString(wxString& str);
 
 private:
-    // This method is implemented in platform-specific code and must return a
+    // These methods are implemented in platform-specific code and must return a
     // new heap-allocated object initialized with the given data.
     static wxSecretValueImpl* NewImpl(size_t size, const void *data);
+    static wxSecretValueImpl* NewImpl(const wxScopedCharBuffer &buf);
 
     // This ctor is only used by wxSecretStore and takes ownership of the
     // provided existing impl pointer.
--- a/samples/secretstore/secretstore.cpp
+++ b/samples/secretstore/secretstore.cpp
@@ -25,6 +25,11 @@
 #include "wx/crt.h"
 #include "wx/log.h"
 
+#include <cctype>
+
+bool g_verbose_ {false};
+bool g_binary_ {false};
+
 bool Save(wxSecretStore& store, const wxString& service, const wxString& user)
 {
     char password[4096];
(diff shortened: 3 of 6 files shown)


vadz:
Thanks, using text/plain is clearly a bad idea for binary data, but I wonder if using application/octet-stream for text data doesn't have some drawbacks too?

[Official documentation](https://gnome.pages.gitlab.gnome.org/libsecret/ctor.Value.new.html) is useless, but there are some specific checks for text/plain in libsecret code, so maybe it would be better to keep using it for non-binary (== not containing zero bytes) data?

mcorino:
According to my tests it doesn't as far as I can tell.On 24 Feb 2024 19:30, VZ ***@***.***> wrote:
Thanks, using text/plain is clearly a bad idea for binary data, but I wonder if using application/octet-stream for text data doesn't have some drawbacks too?
Official documentation is useless, but there are some specific checks for text/plain in libsecret code, so maybe it would be better to keep using it for non-binary (== not containing zero bytes) data?

vadz:
I'd still rather keep using text/plain for passwords because this comment

https://github.com/GNOME/libsecret/blob/a86c93d6f9ec4b4476eff1bbb4fe01ef50f03430/libsecret/secret-value.c#L35

clearly says that this is the right thing to do.

Would there be any problem with using application/octet-stream only if the data contains NUL bytes?

mcorino:
Not really, but the code needs to be more structurally changed than since currently passwords and binary secrets are identically handled as far as the libsecret level are concerned.It will also impact all other platforms than.On 24 Feb 2024 20:54, VZ ***@***.***> wrote:
I'd still rather keep using text/plain for passwords because this comment
https://github.com/GNOME/libsecret/blob/a86c93d6f9ec4b4476eff1bbb4fe01ef50f03430/libsecret/secret-value.c#L35
clearly says that this is the right thing to do.
Would there be any problem with using application/octet-stream only if the data contains NUL bytes?

vadz:
> Not really, but the code needs to be more structurally changed

Sorry, I may be missing something here, but I don't understand why. All I propose is to just do `memchr(data, 0, size) ? "application/octet-stream" : "text/plain"`, why not?

mcorino:
On 2/24/24 23:28, VZ wrote:
>     Not really, but the code needs to be more structurally changed
> 
> Sorry, I may be missing something here, but I don't understand why. All 
> I propose is to just do |memchr(data, 0, size) ? 
> "application/octet-stream" : "text/plain"|, why not?

Because this won't work. My example binary secret in the issue for 
example was "secret\x1\x86\x2\x99\x3" (size 11) which does not contain
any NUL character but still is a binary secret that gets thrashed by
using "text/plain".
Binary secrets are not merely characterized by containing NUL characters 
but rather by any combination of byte values that are not recognized as 
proper text (printable ASCII or unicode characters) somehow.

So if you want to separate content types for those you will have to 
separate the interface into 2 paths for 'wxString' and for 'size_t, void*'.

mcorino:
I added the required changes to keep the distinction between text and binary secrets.
Although this affects all platform implementations the changes themselves are rather trivial and do **not** affect the public API.

vadz:
> Because this won't work. My example binary secret in the issue for instance was "secret\x1\x86\x2\x99\x3" (size 11) which does not contain any NUL character but still is a binary secret that gets thrashed by using "text/plain". Binary secrets are not merely characterized by containing NUL characters but rather by any combination of byte values that are not recognized as proper text (printable ASCII or unicode characters) somehow.

I'd really like to understand this "somehow" because now I have no idea what the problem actually is any more. I can understand the problem with not handling data with embedded NULs, but not this. Just in case, are you sure the distinction isn't between valid UTF-8 data and everything else? I.e. if you encode your test case in UTF-8, does it work?

> So if you want to separate content types for those you will have to separate the interface into 2 paths for 'wxString' and for 'size_t, void*'.

If the real distinction is indeed between UTF-8 and non-UTF-8 data, we could still do the simpler change but replacing `memchr()` with a check for UTF-8-well-formedness.

mcorino:
On 2/25/24 15:00, VZ wrote:
>     Because this won't work. My example binary secret in the issue for
>     instance was "secret\x1\x86\x2\x99\x3" (size 11) which does not
>     contain any NUL character but still is a binary secret that gets
>     thrashed by using "text/plain". Binary secrets are not merely
>     characterized by containing NUL characters but rather by any
>     combination of byte values that are not recognized as proper text
>     (printable ASCII or unicode characters) somehow.
> 
> I'd really like to understand this "somehow" because now I have no idea 
> what the problem actually is any more. I can understand the problem with 
> not handling data with embedded NULs, but not this. Just in case, are 
> you sure the distinction isn't between valid UTF-8 data and everything 
> else? I.e. if you encode your test case in UTF-8, does it work?

I'm not quite sure what you do not understand here. The "somehow" refers 
to whatever the GNOME libsecret service storage process does with 
text/plain data. That's where the thrashing occurs.

Please inspect the extended secretstore sample code I attached with the 
issue and include in the PR.
That sample can be run to selftest with binary secrets where I make it 
create a secret like `wxSecretValue(11, "secret\x1\x86\x2\x99\x3")` 
which should be completely valid and no string encoding involved at all.

With the original libsecret based implementation this will get stored 
with content type "text/plain" and results in the thrashed secret being
returned after reloading from the store (you can see for your self if 
you run the extended sample with the current library code).

A complete check for UTF-8-well-formedness might work as well but I do 
not see how that is better than the last update to the PR I submitted
and it damn well won't be better performant.
Also I do not see why you would want to overrule the intent of the user 
who clearly indicates he/she/they want's to store a binary secret by 
using the wxSecretValue  'size_t, void*' ctor which might happen if you 
start to test any data for UTF-8 well-formedness.

vadz:
> I'm not quite sure what you do not understand here. The "somehow" refers to whatever the GNOME libsecret service storage process does with text/plain data.

Well, this "whatever" is not clear to me. I thought you might already know what this "whatever" was, which is why I asked.

> That's where the thrashing occurs. Please inspect the extended secretstore sample code I attached with the issue and include in the PR.

I did see this code, but I don't know what does it return. I'll try to run it and check for myself.

> A complete check for UTF-8-well-formedness might work as well but I do not see how that is better than the last update to the PR I submitted and it damn well won't be better performant. 

I don't think we care that much about performance there, but, in principle, checking for this could avoid making an extra copy.

> Also I do not see why you would want to overrule the intent of the user who clearly indicates he/she/they want's to store a binary secret by using the wxSecretValue 'size_t, void*' ctor which might happen if you start to test any data for UTF-8 well-formedness.

This is a good point, thanks.

To summarize, determining which content type to use depending on `wxSecretValue` ctor selection is right (because of the argument above). But using different `NewImpl()` overloads for doing it seems a bit strange to me, especially considering that their implementation is exactly identical in MSW and Mac versions and almost the same in the Unix one. Could we just pass different content types to `NewImpl()` instead? Then MSW and Mac would still ignore it, but this doesn't seem too bad...

mcorino:
> > I'm not quite sure what you do not understand here. The "somehow" refers to whatever the GNOME libsecret service storage process does with text/plain data.
> 
> Well, this "whatever" is not clear to me. I thought you might already know what this "whatever" was, which is why I asked.

No, I do not really know. I browsed the libsecret code shortly but could not immediately locate it. However I'm not really interested although I assume it's some sort of encoding issue. I do feel it is a really **BAD THING** that a secret store service should even consider touching the content of a secret I store.

> 
> To summarize, determining which content type to use depending on `wxSecretValue` ctor selection is right (because of the argument above). But using different `NewImpl()` overloads for doing it seems a bit strange to me, especially considering that their implementation is exactly identical in MSW and Mac versions and almost the same in the Unix one. Could we just pass different content types to `NewImpl()` instead? Then MSW and Mac would still ignore it, but this doesn't seem too bad...

Done.

vadz:
Thanks for improving/simplifying this! Will merge soon.

mcorino:
Will you backport this for 3.2.5 as well?

vadz:
@mcorino Not sure if you've seen this, but I've added the label to look into backporting this. I'm not quite sure if it's ok to change the signature of (even a private) function, so I'd need to test it, but if not we can always add another overload instead, which is definitely fine.

mcorino:
Thanks for looking into this.On 2 Mar 2024 17:05, VZ ***@***.***> wrote:
@mcorino Not sure if you've seen this, but I've added the label to look into backporting this. I'm not quite sure if it's ok to change the signature of (even a private) function, so I'd need to test it, but if not we can always add another overload instead, which is definitely fine.

vadz:
Included in #24375.

PR #24360, opened 2024-02-26
Some fixes for wxQt

AliKet:
Replaces #24326

The last two commits are the most questionable on this PR. And AFAICT they works perfectly well and I didn't notice any regression so far. But please let me know if you see anything wrong with them. TIA!

### DIFF
--- a/include/wx/aui/auibar.h
+++ b/include/wx/aui/auibar.h
@@ -627,6 +627,12 @@ public:
 protected:
     void Init();
 
+    // Override to return the minimum acceptable size because under wxMSW this
+    // function returns DEFAULT_ITEM_HEIGHT (see wxControl::DoGetBestSize())
+    // which is not suitable as height/width for a horizontal/vertical toolbar
+    // if icon sizes are much smaller than DEFAULT_ITEM_HEIGHT.
+    virtual wxSize DoGetBestSize() const override;
+
     virtual void OnCustomRender(wxDC& WXUNUSED(dc),
                                 const wxAuiToolBarItem& WXUNUSED(item),
                                 const wxRect& WXUNUSED(rect)) { }
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -799,17 +799,7 @@ bool wxAuiManager::AddPane(wxWindow* window, const wxAuiPaneInfo& paneInfo)
     if (pinfo.best_size == wxDefaultSize &&
         pinfo.window)
     {
-        pinfo.best_size = pinfo.window->GetClientSize();
-
-#if wxUSE_TOOLBAR
-        if (wxDynamicCast(pinfo.window, wxToolBar))
-        {
-            // GetClientSize() doesn't get the best size for
-            // a toolbar under some newer versions of wxWidgets,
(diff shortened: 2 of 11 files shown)


include/wx/aui/auibar.h
```diff
@@ -627,6 +627,12 @@ class WXDLLIMPEXP_AUI wxAuiToolBar : public wxControl
 protected:
     void Init();
 
+    // Override to return wxDefaultSize because under wxMSW wxControl::DoGetBestSize()
+    // returns DEFAULT_ITEM_HEIGHT which is not suitable as the default height/width for
+    // a horizontal/vertical toolbar. Notice that the toolbar will be resized correctly
+    // when Realize() is called.
+    virtual wxSize DoGetBestSize() const override { return wxDefaultSize; }
```
vadz, line 634: It seems very suspicious to return `wxDefaultSize` from here. I'm not sure it breaks something, but I'd be surprised if it didn't.

Couldn't we return `m_sizer->GetMinSize()` here instead? If not, let's at least return `wxSize(1,1)` or something like this, i.e. a valid size, even if a wrong one.

src/aui/framemanager.cpp
```diff
@@ -799,17 +799,7 @@ bool wxAuiManager::AddPane(wxWindow* window, const wxAuiPaneInfo& paneInfo)
...
-#if wxUSE_TOOLBAR
-        if (wxDynamicCast(pinfo.window, wxToolBar))
-        {
-            // GetClientSize() doesn't get the best size for
-            // a toolbar under some newer versions of wxWidgets,
-            // so use GetBestSize()
-            pinfo.best_size = pinfo.window->GetBestSize();
-        }
-#endif // wxUSE_TOOLBAR
+        pinfo.best_size = pinfo.window->GetBestSize();
```
vadz, line 802: I'd like to get rid of the toolbar hack (which I don't really understand), but this is a change in behaviour as the pane will use its minimal, and not current, size now.

I don't know if it's really important, but neither am I sure that it isn't :-(

vadz (commented):
Sorry, I've somehow completely forgot to look at this one.

I have no comments about all the previous commits, but the last one is indeed questionable. If you'd like, we could merge all the other ones already and deal with this one later.

Thanks!

include/wx/aui/auibar.h
vadz: Sorry, but returning current size (i.e. `GetClientSize()`) as best size is definitely not a good idea because this means the window can never become smaller than its current size. If `GetMinSize()` has a valid value, let's return it instead. If not, returning `wxSize(1, 1)` at least shouldn't break anything (although could result in the window being too small, of course).

vadz:
Thanks for the update, I'll merge it soon (forgot to do it during the last round of merges, sorry).

AliKet:
> Thanks for the update, I'll merge it soon (forgot to do it during the last round of merges, sorry).

This is fine as I have 2 bug fixes affecting `wxAUI` that I want be included in this PR. TIA!

AliKet:
@vadz `wxQt` and `wxGTK2` CI need to be restarted , TIA!

PR #24365, opened 2024-02-28
Fix problem with disappearing `wxStaticText` text

vadz:
I really don't feel good doing this, but I couldn't find any other way to fix the problem described in the commit message of the second commit: without this, we may simply not show _anything_ if the width is too small and there is a space before the last not-quite-fitting word.

BTW, I didn't mention this in the commit message, but this can be reproduced in raw Win32 code, e.g.

```cpp
  HWND hwnd = CreateWindow(L"STATIC", L" AJ",
               WS_CHILD | WS_VISIBLE | WS_CLIPSIBLINGS,
               5, 5, 30, 30,
               g_hwndMain, (HMENU)-1, g_hInstance, NULL);

  HFONT hFont = (HFONT)GetStockObject(DEFAULT_GUI_FONT);
  SendMessage(hwnd, WM_SETFONT, (WPARAM)hFont, MAKELPARAM(TRUE, 0));
```

doesn't show anything, but replace the width with `31` shows the text correctly.

Again, I realize that this is ugly, but the problem it fixes is really bad: the user of our application just doesn't see anything at all in it right now for some string values.

### DIFF
--- a/interface/wx/splitter.h
+++ b/interface/wx/splitter.h
@@ -491,7 +491,7 @@ public:
 
         @since 3.3.0
     */
-    const wxPoint& GetLastSplitPosition() const;
+    wxPoint GetLastSplitPosition() const;
 
     /**
         Sets the last sash position.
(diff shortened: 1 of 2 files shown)


vadz:
No comments, so merging.

PR #24371, opened 2024-03-03
Avoid showing bogus tooltips in macOS wxDataViewCtrl

### LINKED ISSUE #24370: wxOSX: wxDataViewCustomRenderer has undesirable tooltip with debug info
When using a dataview with a custom cell renderer, in certain situations a tooltip will be displayed when hovering over the cell that contains debug output of the form `wxCustomRendererObject: <pointer>`

To see this, there must not be a global tooltip set on the wxDataViewCtrl.
To reproduce, take the dataview sample, and comment out the line that sets a tooltip (line 902 on my sources):

```cpp
m_ctrl[Page_Music]->SetToolTip("You may edit most of the cells here!");
```

Then build and run the sample, and hover over one of the custom renderer cells (last column)

If you do not see the tooltips, you may need to make the column smaller; it is not always triggered and I cannot figure out why.

This came from a KiCad report where the column does not need to be small to show the problem: https://gitlab.com/kicad/code/kicad/-/issues/17197

I suspect the problem is that the custom renderer does not define a tooltip override, and so the default provided by macOS is used, which is the `description` property, which is intended as debug output and has the form `class name: memory address`.  However I'm not sure, and I'm also not comfortable enough with the macOS sources to know how to add suitable overrides if this is indeed the problem.

- OS  and its version: macOS 12+

vadz:
Cells using custom renderers showed useless tooltips with the debug representation of wxCustomRendererObject, which were worse than useless, so avoid showing them.

Contrary to what might be expected, this can't be done by overriding the specific tooltip-related delegate method, because these tooltips come from NSTextFieldCell itself, so we need to prevent it from showing them.

Closes #24370.

### DIFF
--- a/src/osx/cocoa/dataview.mm
+++ b/src/osx/cocoa/dataview.mm
@@ -1111,6 +1111,20 @@ outlineView:(NSOutlineView*)outlineView
     return cellFrame;
 }
 
+-(NSRect)expansionFrameWithFrame:(NSRect)cellFrame
+    inView:(NSView*)view
+{
+    // We override the default behaviour to avoid showing nonsensical tooltips
+    // for the custom cells: if we don't do this, NSTextFieldCell would show a
+    // tooltip with the debug representation of wxCustomRendererObject which is
+    // not very useful.
+    //
+    // Note: override outlineView:toolTipForCell:rect:tableColumn:item:mouseLocation:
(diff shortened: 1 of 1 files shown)


craftyjon:
Wow, thank you for the fast fix!  I have tested this with KiCad and it works to solve the issue.

csomor (approved):
looks fine to me, thanks

PR #24369, opened 2024-03-02
Lexilla 5.3

MaartenBent:
I changed the submodule URL so I can fix all the old compiler specific c++11 warnings and errors that I missed. 
When all CI are happy I'll create a PR at https://github.com/wxWidgets/lexilla

PR #24373, opened 2024-03-03
Switch to using FreeBSD 14 on Cirrus CI

vadz:
FreeBSD 13.2 doesn't seem to be available any longer.


MaartenBent:
Failed build caused by Lexilla. I'll cherry-pick this into #24369 and fix it if still necessary.

PR #24374, opened 2024-03-03
Implement loading cursor from macOS resources

MaartenBent:
I finally gave in and fixed #23558

TODO: add it to resources in .app created by CMake


oneeyeman1:
@MaartenBent ,
I'm not sure why CMake needs modification. Looks like it works fine

src/osx/carbon/cursor.cpp
```diff
@@ -277,7 +279,26 @@ wxCursor::wxCursor(const wxString& cursor_file, wxBitmapType flags, int hotSpotX
     if ( flags == wxBITMAP_TYPE_MACCURSOR_RESOURCE )
     {
 #if wxOSX_USE_COCOA
-        wxFAIL_MSG( wxT("Not implemented") );
+        wxImage image;
```
PBfordev, line 282: Just curious: So we no longer care about `wxUSE_xx` defines in wxWidgets code (see `wxUSE_IMAGE` below)?

src/osx/carbon/cursor.cpp
```diff
@@ -277,7 +279,26 @@ wxCursor::wxCursor(const wxString& cursor_file, wxBitmapType flags, int hotSpotX
...
+        {
+            image.SetOption( wxIMAGE_OPTION_CUR_HOTSPOT_X, hotSpotX ) ;
+            image.SetOption( wxIMAGE_OPTION_CUR_HOTSPOT_Y, hotSpotY ) ;
+        }
+        else
+        {
+            fileName.SetExt( "cur" );
+            image.LoadFile( fileName.GetFullPath(), wxBITMAP_TYPE_CUR );
+        }
+        if( image.IsOk() )
```
PBfordev, line 294:
```suggestion
        if ( image.IsOk() )
```

src/osx/carbon/cursor.cpp
```diff
@@ -277,7 +279,26 @@ wxCursor::wxCursor(const wxString& cursor_file, wxBitmapType flags, int hotSpotX
     if ( flags == wxBITMAP_TYPE_MACCURSOR_RESOURCE )
     {
 #if wxOSX_USE_COCOA
-        wxFAIL_MSG( wxT("Not implemented") );
+        wxImage image;
+        wxFileName fileName( wxStandardPaths::Get().GetResourcesDir(), cursor_file, "png" );
+        if( image.LoadFile( fileName.GetFullPath(), wxBITMAP_TYPE_PNG ) )
```
PBfordev, line 284:
```suggestion
        if ( image.LoadFile( fileName.GetFullPath(), wxBITMAP_TYPE_PNG ) )
```

tests/test.rc
```diff
@@ -0,0 +1,38 @@
+/////////////////////////////////////////////////////////////////////////////
```
PBfordev, line 1: The file header is not in accordance with the coding guidelines: https://wxwidgets.org/develop/coding-guidelines/#wxwin_header

I know it is incorrect (for historic reasons?) in the sample.rc as well, but I think we should always use the correct header for new files.

Also, can't we just include "../samples/sample.rc" from test.rc instead of duplicating and add only the cursor resource gere?

src/osx/carbon/cursor.cpp
MaartenBent: I don't know. But I will add them here.

tests/test.rc
MaartenBent: I'll give that a try. I don't see why it wouldn't work.

MaartenBent:
> I'm not sure why CMake needs modification. Looks like it works fine

CMake doesn't run any tests.

MaartenBent:
This should be fine to merge now.

oneeyeman1:
@MaartenBent , thx.

tests/controls/windowtest.cpp
```diff
@@ -143,11 +192,29 @@ TEST_CASE_METHOD(WindowTestCase, "Window::FocusEvent", "[window]")
...
+    // from the external file and not creating it from embedded bits/resources
+    // Therefore different path are used for creating embedded cursors
+    wxCursor resCursor1( (const char *) horse, 32, 32, 16, 0, nullptr, wxWHITE, wxBLACK );
+    CHECK( resCursor1.IsOk() );
+#endif
+#ifdef __WXOSX__
+    // Check the cur file
+    wxCursor resCursor2( "bullseye" );
+    CHECK( resCursor2.IsOk() );
+#endif
```
vadz, line 217: I think this test should go into `tests/image/image.cpp` where we already have some similar tests for `wxCursor` and have a test XPM too, it doesn't seem right to have it here.

src/osx/carbon/cursor.cpp
```diff
@@ -17,7 +17,9 @@
     #include "wx/image.h"
 #endif // WX_PRECOMP
 
+#include "wx/filename.h"
 #include "wx/xpmdecod.h"
+#include "wx/stdpaths.h"
```
vadz, line 22: Very minor, but could this line go above, please, to keep the header in alphabetical order?

vadz (commented):
Thanks, this indeed looks good, but could the test be moved to a better place please?

tests/controls/windowtest.cpp
MaartenBent: I will move it. But the `wxCursor` constructor accepts XBM (X BitMap), not XPM (X PixMap) so it can't reuse `g_testfiles[1]`.

MaartenBent: Looks like the horse xbm data is the contents of `horse.cur`, so I'll just read that file to a buffer.

oneeyeman1:
@MaartenBent ,
The idea is to not to use the external file, but use one from the resources or compiled in binary data...

Thank you.

MaartenBent:
It is about testing the wxCursor constructors. The special gtk wxCursor constructor doesn't care if it gets the bytes from heap or stack memory.

I'm also considering adding some tests for loading cursor from file.

MaartenBent:
I did have to define an XBM image in the cpp file. Because, obviously, .cur and .xbm are not the same.
Now it is a valid image instead of the raw data of `horse.cur`, so the cursor actually looks like a horse instead of random white and black pixels.
I'll refactor the commits once all the CI like the new tests.

vadz:

oneeyeman1:
@MaartenBent ,

Thank you.

PR #24381, opened 2024-03-06
Fix build failure on Windows XP.

solemnwarning:
I'm not sure what the official minimum version is currently... the wiki references 9x and I found a forum post talking about XP being supported-but-untested as of a few years ago.

This patch applied to the vcpkg port is enough to get it working for me anyway.

### DIFF
--- a/src/msw/taskbar.cpp
+++ b/src/msw/taskbar.cpp
@@ -248,13 +248,16 @@ wxTaskBarIcon::ShowBalloon(const wxString& title,
 
     wxUnusedVar(icon); // It's only unused if not supported actually.
 
+#if _WIN32_WINNT >= 0x0600
     if ( icon.IsOk() )
     {
         m_balloonIcon = icon.GetIconFor(m_win);
         notifyData.hBalloonIcon = GetHiconOf(m_balloonIcon);
         notifyData.dwInfoFlags |= NIIF_USER | NIIF_LARGE_ICON;
     }
-    else if ( flags & wxICON_INFORMATION )
+    else
+#endif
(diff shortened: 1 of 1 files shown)


vadz:
We don't support XP in master (3.3) any more, but it's still supposed to be supported in 3.2, so we could apply it there.

Just to be clear, does your build system predefine `_WIN32_WINNT` to some lesser value? Because by default it's defined to the highest possible value in `include/wx/msw/winver.h`.

solemnwarning:
> Just to be clear, does your build system predefine `_WIN32_WINNT` to some lesser value? Because by default it's defined to the highest possible value in `include/wx/msw/winver.h`.

Yes, I'm predefining it to 0x0501

vadz:
> > Just to be clear, does your build system predefine `_WIN32_WINNT` to some lesser value? Because by default it's defined to the highest possible value in `include/wx/msw/winver.h`.
> 
> Yes, I'm predefining it to 0x0501

I don't care much about XP, but the general policy of wxMSW is that we always compile with the latest available SDK and all checks are done during run-time, so normally this should be done here too.

> EDIT: Should I rebase my change onto the `3.2` branch?

If this commit can be cherry-picked without conflicts there (and I think it can), there is no need to do this.

vadz:
Actually, trying to cherry pick it I see that we did have a run-time check for Vista in 3.2, which was removed when support for XP/Vista got removed in master.

So I don't think we should apply this, finally, sorry. You shouldn't have any problem with targetting XP in 3.2 (but if you do, please make PRs fixing them there).

PR #24383, opened 2024-03-07
wxrc: XRCWndClassData: minor improvements and `GenerateHeaderCode` bugfix.

DoctorNoobingstoneIPresume:
For classes with more than one ancestor, the generated C++ code used to end the class definition more than once. This has now been fixed.

Also, for these classes, all constructors now have a default value (`nullptr`) for the `parent` parameter.

Other improvements (minor) have been made:

- the `GetWidgetData` member function is now `const`;
- the `CanBeUsedWithXRCCTRL` member function is now `static`;
- the generated C++ code now uses `static_cast` instead of C-style cast when preparing the `parent` argument for the call to `InitWidgetsFromXRC`;
- some string literals are now merged at preprocessing time (as opposed to being added at runtime).

### DIFF
--- a/utils/wxrc/wxrc.cpp
+++ b/utils/wxrc/wxrc.cpp
@@ -102,12 +102,12 @@ public:
         BrowseXmlNode(node->GetChildren());
     }
 
-    const ArrayOfXRCWidgetData& GetWidgetData()
+    const ArrayOfXRCWidgetData& GetWidgetData() const
     {
         return m_wdata;
     }
 
-    bool CanBeUsedWithXRCCTRL(const wxString& name)
+    static bool CanBeUsedWithXRCCTRL(const wxString& name)
     {
         if (name == wxT("tool") ||
             name == wxT("data") ||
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, I'd just like to understand something: how do you end up with multiple base classes here? I.e. would you have a minimum XRC example reproducing the problem?

DoctorNoobingstoneIPresume:
> @DoctorNoobingstoneIPresume I wonder if you have seen the question above?

Thank you for kind replies and please forgive my late reaction...

In `wxrc.cpp`, I have been looking at the constructor of `XRCWndClassData`:

```
    XRCWndClassData(const wxString& className,
                    const wxString& parentClassName,
                    const wxXmlNode* node) :
        m_className(className) , m_parentClassName(parentClassName)
    {
        if ( className == wxT("wxMenu") )
        {
            m_ancestorClassNames.insert(wxT("wxMenu"));
            m_ancestorClassNames.insert(wxT("wxMenuBar"));
        }
        else if ( className == wxT("wxMDIChildFrame") )
        {
            m_ancestorClassNames.insert(wxT("wxMDIParentFrame"));
        }
        else if( className == wxT("wxMenuBar") ||
                    className == wxT("wxStatusBar") ||
                        className == wxT("wxToolBar") )
        {
            m_ancestorClassNames.insert(wxT("wxFrame"));
        }
        else
        {
            m_ancestorClassNames.insert(wxT("wxWindow"));
        }

        BrowseXmlNode(node->GetChildren());
    }
```

=> The `m_ancestorClassNames` container has more than one element in the case when the `className` parameter of the constructor contains the `"wxMenu"` text.

`XRCWndClassData` objects are constructred in the `PrepareTempFiles` function:

```
...
while (node)
{
    if (node->GetName () == wxT ("object") && node->GetAttribute (wxT ("class"), &classValue) && node->GetAttribute (wxT ("name"), &nameValue))
        aXRCWndClassData.Add (XRCWndClassData (nameValue, classValue, node));
    node = node -> GetNext ();
}
...
```

If we use the following input XRC file:

```
<?xml version="1.0" encoding="UTF-8"?>
<resource xmlns="http://www.wxwidgets.org/wxxrc" version="2.5.3.0">
  <object class="This_argument_goes_to_the_parentClassName_parameter" name="wxMenu"></object>
</resource>
```

and we process it using:

```
wxrc -c -e -n 'InitXmlResource' -o 'Example3.cpp' 'Example3.xrc'
```

then the resulting `Example3.h` header file is:

```
//
// This file was automatically generated by wxrc, do not edit by hand.
//

#ifndef __Example3_h__
#define __Example3_h__
class wxMenu : public This_argument_goes_to_the_parentClassName_parameter {
protected:

private:
 void InitWidgetsFromXRC(wxWindow *parent){
  wxXmlResource::Get()->LoadObject(this,parent,wxT("wxMenu"), wxT("This_argument_goes_to_the_parentClassName_parameter"));
 }
public:
wxMenu(){
  InitWidgetsFromXRC(nullptr);
 }
};
wxMenu(wxMenu *parent){
  InitWidgetsFromXRC((wxWindow *)parent);
 }
};
wxMenu(wxMenuBar *parent){
  InitWidgetsFromXRC((wxWindow *)parent);
 }
};

void 
'InitXmlResource'();
#endif

```

and we can see that the class definition for `wxMenu` is (incorrectly) ended multiple times.

If I have misinterpreted the intention of the code, I wish to ask that you kindly let me know (when there is time).

DoctorNoobingstoneIPresume:
I certainly agree with you, and I will close this pull request if it is alright with you, and thank you for patience, that you have given me time so I can learn myself.

(Indeed, I too was surprized, that the name of the object, not the name of the class of the object, had to be `wxMenu` in order for that path to be taken...)

paulcor:
> and while I have absolutely no memory of this patch and can't find it any more

#7433

Issue #24289, opened 2024-02-06
wxOwnerDrawnComboBox: wrong selection when two items have the same name but are written with lower and upper case


vwichmann:
#### Bug description:

If there are two items with the same name in the choice list of a wxOwnerDrawnComboBox, but one is written upper case and the other one lower case, always the first in the list gets selected (second one can't be selected). It works fine when both items are written either upper or lower case.

This relates to an already fixed [issue](https://github.com/wxWidgets/wxWidgets/issues/19289). The fix allowed to correctly select two items with the same name when both are written either lower or upper case. Sorry that we missed to test the upper/lower combination by that time.

#### Expected vs observed behaviour:

It should be possible to select each item in a list, regardless how it is written and if it is a duplicate name or not.

#### Patch or snippet allowing to reproduce the problem:

Attached is a patch to the combo sample which allows to reproduce the problem. It simply replaces three item names with "Test", "Test", and "test". 

[repro-for-combo-selection-bug.txt](https://github.com/wxWidgets/wxWidgets/files/14178685/repro-for-combo-selection-bug.txt)

#### To Reproduce:

1. apply the patch to the combo sample and run it
2. try to select the "Test/Test/test" items from the "OwnerDrawnComboBox with owner-drawn items" list:
3. Select the first upper case "Test" item in the list, log is "EVT_COMBOBOX(id=-31991,selection=6)" and the item is correctly selected
4. Select the second upper case "Test" item in the list, log is "EVT_COMBOBOX(id=-31991,selection=7)" and the item is correctly selected
5. Select the lower case "test" item in the list, log is "EVT_COMBOBOX(id=-31991,selection=8)", but although "selection=8" is reported, the first upper case "Test" item gets selected

### Platform and version information

- wxWidgets version: git master

vwichmann:
Just noticed that the selection works fine in the sample with "OwnerDrawnComboBox with simple dropbottom graphics" which uses wxOwnerDrawnComboBox directly.

It fails with "OwnerDrawnComboBox with owner-drawn items" and "OwnerDrawnComboBox with owner-drawn items and buttom on the left", both being created with the sample's wxPenStyleComboBox derived from wxOwnerDrawnComboBox.

vadz:
Thanks for reporting this!

We must be (wrongly) using case-insensitive comparison somewhere, if you can find where exactly does it happen, it should be quite simple to fix. Please try debugging it if you can in order to do it. TIA!

[...]

PR #24389, opened 2024-03-10
Fix case sensitive item selection in wxOwnerDrawnComboBox

vwichmann:
Currently the selection of items in wxOwnerDrawnComboBox is case insensitive, see #24289. Fix this by making FindItem() in wxVListBoxComboPopup case sensitive.


vadz:
Sorry, this indeed can't be applied as is, as discussed in the linked ticket. It would be useful to allow optionally turning on case-sensitivity here, however.

PR #24390, opened 2024-03-10
Fix crash when wxWindowMac peer was not created

craftyjon:
After testing the fixes for Sonoma clipping problems (bcbc31e97f, 5dffb76d0a, 6e7bbfd17) backported to 3.2 in KiCad, I had a crash on startup because we have at least one control with `m_peer` set to `kOSXNoWidgetImpl`


vadz:
Oops, thanks for the fix, will apply (and backport) it soon.

PR #24396, opened 2024-03-12
Fix preferred language detection on macOS

t123yh:
On macOS, the "preferredLanguages" API returns something like `zh-Hans-CN`, but we want the format of `zh_CN`. This converts the format so that the language can be detected correctly.

This fixes [This kicad issue](https://gitlab.com/kicad/code/kicad/-/issues/16052), please see discussion there.

### DIFF
--- a/src/osx/core/uilocale.mm
+++ b/src/osx/core/uilocale.mm
@@ -362,8 +362,18 @@ wxVector<wxString> wxUILocaleImpl::GetPreferredUILanguages()
     NSArray* preferredLangs = [NSLocale preferredLanguages];
     NSUInteger count = preferredLangs.count;
 
-    for (NSUInteger j = 0; j < count; ++j)
-        preferred.push_back(wxCFStringRef::AsString(preferredLangs[j]));
+    for (NSUInteger j = 0; j < count; ++j) {
+        NSString *langCode = preferredLangs[j];
+        // macOS will return things like `zh-Hans-CN`, but we want format of `zh_CN`.
+        // So, we first replace '-' with '_'
+        langCode = [langCode stringByReplacingOccurrencesOfString:@"-" withString:@"_"];
+        // And then remove the middle part, if exists
+        NSArray *langComponents = [langCode componentsSeparatedByString:@"_"];
+        if ([langComponents count] > 2) {
+            langCode = [NSString stringWithFormat:@"%@_%@", langComponents[0], langComponents[2]];
+        }
+        preferred.push_back(wxCFStringRef::AsString(langCode));
+    }
 
     return preferred;
 }


vadz:
Does the problem only exist in 3.2 or in master too? I really thought I had fixed it in the latter (but then I also thought that I had backported the fixes to 3.2...).

t123yh:
Hi vadz,

Could you please point me to the exact commit hash that fixes this problem, so I can test it?

vadz:
> Could you please point me to the exact commit hash that fixes this problem, so I can test it?

Unfortunately it's not that simple because there was a number of changes in this code and I don't remember the details any more. But I do remember that one of the problems was not handling the language variants (like `Hant`) correctly and I thought it worked now...

Anyhow, if `wxUILocale::GetPreferredUILanguages()` still returns wrong value for you, it must be still broken. Could you please check it in master? This can be done for example by applying the following patch and building and running the "internat" sample:

```diff
diff --git a/samples/internat/internat.cpp b/samples/internat/internat.cpp
index 9a2a8f0ea8..481aa05885 100644
--- a/samples/internat/internat.cpp
+++ b/samples/internat/internat.cpp
@@ -387,6 +387,18 @@ MyFrame::MyFrame()
                       ),
                   wxSizerFlags().Center().Border());
 
+    topSizer->Add(new wxStaticText
+                      (
+                        panel,
+                        wxID_ANY,
+                        wxString::Format
+                        (
+                          _("Preferred UI languages: %s"),
+                          wxJoin(wxUILocale::GetPreferredUILanguages(), ' ')
+                        )
+                      ),
+                  wxSizerFlags().Center().Border());
+
     // create some controls affected by the locale
 
     // this demonstrates RTL layout mirroring for Arabic locales and using
```

TIA!

t123yh:
Hi vadz,

I believe that this issue is not solved in master and 3.2 branch. 

Master:

image

3.2:

image

t123yh:
Investigated a little more, and found that it's actually fixed. It turns out that we should parse the result with `wxLocaleIdent::FromTag`, and then we seem to get the correct result:

```diff
diff --git a/samples/internat/internat.cpp b/samples/internat/internat.cpp
index 53ca7ebbb1..cf4163550f 100644
--- a/samples/internat/internat.cpp
+++ b/samples/internat/internat.cpp
@@ -388,6 +388,28 @@ MyFrame::MyFrame()
                       ),
                   wxSizerFlags().Center().Border());

+    wxVector<wxString> preferred = wxUILocale::GetPreferredUILanguages();
+    wxArrayString langs;
+    for ( wxVector<wxString>::const_iterator j = preferred.begin();
+          j != preferred.end();
+          ++j )
+    {
+        wxLocaleIdent localeId = wxLocaleIdent::FromTag(*j);
+        wxString lang = localeId.GetTag(wxLOCALE_TAGTYPE_POSIX);
+        langs.Add(lang);
+    }
+    topSizer->Add(new wxStaticText
+                      (
+                        panel,
+                        wxID_ANY,
+                        wxString::Format
+                        (
+                          _("Preferred UI languages (in POSIX format): %s"),
+                          wxJoin(langs, ' ')
+                        )
+                      ),
+                  wxSizerFlags().Center().Border());
+
     // create some controls affected by the locale

     // this demonstrates RTL layout mirroring for Arabic locales and using
```

image

vadz:
Actually the commit I had in mind was this one: 9ada6cd2db (Fix parsing Windows sort order in wxLocaleIdent::FromTag(), 2023-05-25) and as you can see it just fixed parsing of the strings with the script (not "variant", as I erroneously wrote above) part.

And I think that returning `zh-Hans-CN` from `GetPreferredUILanguages()` is correct, after all it's the language that should be used (simplified Chinese), the bug is actually with `wxTranslations` wrongly selecting traditional Chinese in this case.

I'll try to fix this there.

vadz:
> Investigated a little more, and found that it's actually fixed. It turns out that we should parse the result with `wxLocaleIdent::FromTag`, and then we seem to get the correct result:

Yes, exactly, thanks (my previous comment had been posted before seeing this one).

The important question for me is whether `wxTranslations` handles this correctly or if it's something that KiCad does?

t123yh:
KiCad has its own fork based on 3.2.4 [here](https://gitlab.com/kicad/code/wxWidgets), so if it's something that happened between 3.2.4 and the current 3.2 branch, then it could be different. I will try to test this example on the forked 3.2.4 code base.

vadz:
Thanks! I think we can close this one because I'm quite sure we should return the real preferred language from here (perhaps someone does want to use `zh-Hant-CN` -- even though I understand that this must be very rare), so whatever the problem is this is not the right fix. But if you can reproduce the problem with the current master and/or 3.2, please open an issue or a PR if you have a fix. TIA!

vadz:
FWIW I think the problem is indeed in wx, at least if I do this:

```diff
diff --git a/tests/intl/intltest.cpp b/tests/intl/intltest.cpp
index a504a9edaa..317a8b6e6f 100644
--- a/tests/intl/intltest.cpp
+++ b/tests/intl/intltest.cpp
@@ -344,6 +344,22 @@ TEST_CASE("wxTranslations::GetBestTranslation", "[translations]")
     }
 }
 
+// This test can be used to check how GetBestTranslation() and
+// GetAvailableTranslations() work with the given preferred languages: set
+// WXLANGUAGE environment variable to the colon-separated list of preferred
+// languages to test before running it.
+TEST_CASE("wxTranslations::ShowAvailable", "[.]")
+{
+    wxFileTranslationsLoader::AddCatalogLookupPathPrefix("./intl");
+
+    const wxString domain("internat");
+
+    wxTranslations trans;
+
+    WARN("Available: " << wxJoin(trans.GetAvailableTranslations(domain), ' '));
+    WARN("Best:      \"" << trans.GetBestTranslation(domain) << "\"");
+}
+
 TEST_CASE("wxLocale::Default", "[locale]")
 {
     const int langDef = wxUILocale::GetSystemLanguage();
```

and create (bogus) `zh_CN` translations, running the test doesn't produce the expected results:

```console
$ env -C tests WXLANGUAGE=zh-Hans-CN ./test wxTranslations::ShowAvailable
  Available: ja fr zh_CN zh_TW
  Best:      "en"
```

vadz:
> FWIW I think the problem is indeed in wx, at least if I do this:

Scratch that, this doesn't prove anything because `WXLANGUAGE` parsing just doesn't recognize `zh-Hans-CN` as valid language. If I fix this, I do get the expected result for both `zh-Hans-CN` and `zh-Hant-TW` in master.

But:

1. This doesn't work in 3.2 yet because #24297 wasn't backported there yet.
2. This still doesn't work correctly for `zh-Hant-CN` which should, I believe, prefer `zh_TW` (Traditional).

t123yh:
To demonstrate this problem in a simpler way, I created another patch:

https://gist.github.com/t123yh/65e302c015fd0b660594db22a1304206

In this patch, I copied implementation of `GetPreferredUILanguage`from `src/common/translation.cpp`. Then, I called that like this, when preferred UI language is `zh-Hans-CN`:

```
wxArrayString available;
available.Add("en");
available.Add("zh_TW");
available.Add("zh_CN");
wxString lll = GetPreferredUILanguage(available); // <===== this gets zh_TW
```

Running this on the master branch, I still get `zh_TW` as the result. So, there must be something wrong. I suspect that an extra `@` is added to the end of the result from `GetTag`, which the logic inside `GetPreferredUILanguage` cannot handle.

image

vadz:
Yes, I've fixed this locally, will create a PR soon.

vadz:
Please check #24399. It should be backported to 3.2 too later.

PR #24395, opened 2024-03-12
EGL wxGLCanvas: prefer configs without alpha when EGL_ALPHA_SIZE is zero or not specified.

dsa-t:
There's an issue with KiCad when using KDE6 on native Wayland with Intel graphics (Mesa 24.0.2, Iris) causing OpenGL buffers to be transparent.

See KiCad issue: https://gitlab.com/kicad/code/kicad/-/issues/17341

Even if `EGL_ALPHA_SIZE` is not set, when using `eglChooseConfig`, Mesa sorts the list in such a way that configs with alpha are put at the start, and wx only used the first one. In the sorting order this list looks like this:

```
     bf lv colorbuffer dp st  ms    vis   cav bi  renderable  supported
  id sz  l  r  g  b  a th cl ns b    id   eat nd gl es es2 vg surfaces 
---------------------------------------------------------------------
0x7a 16  0  5  5  5  1 16  0  0 0 0x00--         y  y  y     win <==== wx picked this one
0x8e 16  0  4  4  4  4 16  0  0 0 0x00--         y  y  y     win
0x7b 16  0  5  5  5  1 24  0  0 0 0x00--         y  y  y     win
0x8f 16  0  4  4  4  4 24  0  0 0 0x00--         y  y  y     win
0x7c 16  0  5  5  5  1 24  8  0 0 0x00--         y  y  y     win
0x90 16  0  4  4  4  4 24  8  0 0 0x00--         y  y  y     win
0x81 16  0  5  5  5  1 16  0  2 1 0x00--         y  y  y     win
0x95 16  0  4  4  4  4 16  0  2 1 0x00--         y  y  y     win
0x85 16  0  5  5  5  1 24  0  2 1 0x00--         y  y  y     win
0x99 16  0  4  4  4  4 24  0  2 1 0x00--         y  y  y     win
0x89 16  0  5  5  5  1 24  8  2 1 0x00--         y  y  y     win
0x9d 16  0  4  4  4  4 24  8  2 1 0x00--         y  y  y     win
0x82 16  0  5  5  5  1 16  0  4 1 0x00--         y  y  y     win
0x96 16  0  4  4  4  4 16  0  4 1 0x00--         y  y  y     win
0x86 16  0  5  5  5  1 24  0  4 1 0x00--         y  y  y     win
0x9a 16  0  4  4  4  4 24  0  4 1 0x00--         y  y  y     win
0x8a 16  0  5  5  5  1 24  8  4 1 0x00--         y  y  y     win
[...]

PR #24397, opened 2024-03-12
Prefer EGL configurations without alpha if it wasn't requested

vadz:
At least Mesa Iris on Wayland is known to put RGBA5551 and RGBA4444 formats at the start of the config list, even if we didn't request alpha, which is unexpected, so filter them out manually in this case to ensure we get the expected RGB444, as with the other drivers.

See #24395.


dsa-t:
Works fine in 3.2, only had to add `#include <memory>`.

>ensure we get the expected RGB444

The first config without alpha with this driver is actually RGB888.

vadz:
Merged (with fixed commit message) in the commit above.

PR #24399, opened 2024-03-12
Locale script fixes

vadz:
The net result of this PR is that `zh_CN` translations should be correctly loaded on macOS systems using simplified Chinese locale, which corresponds to `zh-Hans-CN` being returned from `wxUILocale::GetPreferredUILanguages()`. @t123yh Could you please confirm this?

The rest is just helpers for testing.

I also think that we need to change `wxTranslations::DoGetBestAvailableTranslation()` to compare `wxLocaleIdent` objects rather than using string comparisons, but I haven't done this yet.

@vslavik This is not really related to #24297 (except in the sense that it shows once again that these changes were correct), but I've slightly tweaked `WXLANGUAGE` handling for convenience.

@utelle I think the extra `@` in the tag was an obvious bug, but please let me know if I'm missing something. The thing that I don't understand is why we don't have neither `Hans` nor `Hant` in `misc/languages/scripttabl.txt`? Of course, in a way it's lucky that we don't, because if we did return `zh_CN@Hans` as the POSIX locale name, we wouldn't find `zh_CN` translations for it due to the current code using string comparisons.

### DIFF
--- a/docs/doxygen/overviews/envvars.h
+++ b/docs/doxygen/overviews/envvars.h
@@ -54,9 +54,10 @@ wxWidgets programs.
 @itemdef{WXLANGUAGE,
          This variable can be set to override OS setting of preferred languages
          and make wxUILocale::GetPreferredUILanguages() return the set list
-         instead. The format is same as GNU's <a
+         instead. The format is a superset of GNU's <a
          href="https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html">LANGUAGE</a>
-         variable: a colon-separated list of language codes.}
+         variable: a colon-separated list of language tags (which are, in their
+         simplest form, just the language names).}
 */
 
 @see wxSystemOptions
--- a/samples/internat/internat.cpp
+++ b/samples/internat/internat.cpp
@@ -387,6 +387,18 @@ MyFrame::MyFrame()
                       ),
                   wxSizerFlags().Center().Border());
 
+    topSizer->Add(new wxStaticText
+                      (
+                        panel,
+                        wxID_ANY,
+                        wxString::Format
+                        (
+                          _("Preferred UI languages: [%s]"),
+                          wxJoin(wxUILocale::GetPreferredUILanguages(), ',')
+                        )
+                      ),
+                  wxSizerFlags().Center().Border());
+
     // create some controls affected by the locale
 
     // this demonstrates RTL layout mirroring for Arabic locales and using
--- a/src/common/uilocale.cpp
+++ b/src/common/uilocale.cpp
@@ -404,7 +404,11 @@ wxString wxLocaleIdent::GetTag(wxLocaleTagType tagType) const
             if (!m_charset.empty())
                 tag << '.' << m_charset;
             if (!m_script.empty())
-                tag << '@' << wxUILocale::GetScriptAliasFromName(m_script);
+            {
+                const wxString& script = wxUILocale::GetScriptAliasFromName(m_script);
+                if (!script.empty())
+                    tag << '@' << script;
+            }
             else if (!m_modifier.empty())
                 tag << '@' << m_modifier;
             break;
--- a/src/common/uilocale.cpp
+++ b/src/common/uilocale.cpp
@@ -763,14 +763,25 @@ wxVector<wxString> wxUILocale::GetPreferredUILanguages()
         while (tknzr.HasMoreTokens())
         {
             const wxString tok = tknzr.GetNextToken();
-            if (const wxLanguageInfo* li = wxUILocale::FindLanguageInfo(tok))
+            auto ident = wxLocaleIdent::FromTag(tok);
+            if ( ident.IsEmpty() )
             {
-                preferred.push_back(li->CanonicalName);
+                wxLogTrace(TRACE_I18N, "Invalid language code '%s' in WXLANGUAGE", tok);
+                continue;
             }
+
+            if ( !wxUILocale::FindLanguageInfo(ident.GetLanguage()) )
+            {
+                wxLogTrace(TRACE_I18N, "Unknown language in '%s' in WXLANGUAGE", tok);
+                continue;
+            }
+
+            preferred.push_back(ident.GetTag());
         }
         if (!preferred.empty())
         {
-            wxLogTrace(TRACE_I18N, " - using languages override from WXLANGUAGE: '%s'", languageFromEnv);
+            wxLogTrace(TRACE_I18N, " - using languages override from WXLANGUAGE: [%s]",
+                       wxJoin(preferred, ','));
             return preferred;
         }
     }
--- a/src/unix/uilocale.cpp
+++ b/src/unix/uilocale.cpp
@@ -332,20 +332,7 @@ wxString wxLocaleIdent::GetName() const
 
     wxString name;
     if ( !m_language.empty() )
-    {
-        name << m_language;
-
-        if ( !m_region.empty() )
-            name << "_" << m_region;
-
-        if ( !m_charset.empty() )
-            name << "." << m_charset;
-
-        if ( !m_script.empty() )
-            name << "@" << wxUILocale::GetScriptAliasFromName(m_script);
-        else if ( !m_modifier.empty() )
-            name << "@" << m_modifier;
-    }
+        name = GetTag(wxLOCALE_TAGTYPE_POSIX);
 
     return name;
 }
--- a/tests/intl/intltest.cpp
+++ b/tests/intl/intltest.cpp
@@ -476,6 +476,11 @@ static void CheckTag(const wxString& tag)
     CHECK( wxLocaleIdent::FromTag(tag).GetTag() == tag );
 }
 
+static wxString TagToPOSIX(const char* tag)
+{
+    return wxLocaleIdent::FromTag(tag).GetTag(wxLOCALE_TAGTYPE_POSIX);
+}
+
 TEST_CASE("wxLocaleIdent::FromTag", "[uilocale][localeident]")
 {
     CheckTag("");
@@ -485,6 +490,8 @@ TEST_CASE("wxLocaleIdent::FromTag", "[uilocale][localeident]")
     CheckTag("English");
     CheckTag("English_United States");
     CheckTag("English_United States.utf8");
+
+    CHECK( TagToPOSIX("zh-Hans-CN") == "zh_CN" );
 }
 
 // Yet another helper for the test below.
--- a/tests/intl/intltest.cpp
+++ b/tests/intl/intltest.cpp
@@ -344,6 +344,25 @@ TEST_CASE("wxTranslations::GetBestTranslation", "[translations]")
     }
 }
 
+// This test can be used to check how GetBestTranslation() and
+// GetAvailableTranslations() work with the given preferred languages: set
+// WXLANGUAGE environment variable to the colon-separated list of preferred
+// languages to test before running it.
+TEST_CASE("wxTranslations::ShowAvailable", "[.]")
+{
+    wxFileTranslationsLoader::AddCatalogLookupPathPrefix("./intl");
+
+    const wxString domain("internat");
+
+    wxTranslations trans;
+
+    WARN
+    (
+        "Available: [" << wxJoin(trans.GetAvailableTranslations(domain), ',') << "]\n"
+        "Best:      [" << trans.GetBestTranslation(domain) << "]"
+    );
+}
+
 TEST_CASE("wxLocale::Default", "[locale]")
 {
     const int langDef = wxUILocale::GetSystemLanguage();


t123yh:
Yes, it now gets correct `zh_CN`.

utelle:
> @utelle I think the extra `@` in the tag was an obvious bug, but please let me know if I'm missing something. 

Yes, that was a bug. If no alias is found for the script tag an empty string is returned, and in that case no `@` should be added. But that was not checked, resulting in the trailing `@`. Thanks for fixing this.

> The thing that I don't understand is why we don't have neither `Hans` nor `Hant` in `misc/languages/scripttabl.txt`?

This file contains mappings from a script tag to a script alias. On Unix systems typically the script alias is used for specifying the script by appending `@scriptalias`.

The Unicode table used to generate `scripttabl.txt` does not specify aliases for the script tags `Hans` and `Hant`. Therefore these tags are omitted from `scripttabl.txt` (just as all other script tags for which no alias is defined).

> Of course, in a way it's lucky that we don't, because if we did return `zh_CN@Hans` as the POSIX locale name, we wouldn't find `zh_CN` translations for it due to the current code using string comparisons.

Typically, the script alias is not used in POSIX locale names, if it is equal to the default script of the language in the specified region.

vadz:
> > The thing that I don't understand is why we don't have neither `Hans` nor `Hant` in `misc/languages/scripttabl.txt`?
> 
> This file contains mappings from a script tag to a script alias. On Unix systems typically the script alias is used for specifying the script by appending `@scriptalias`.
> 
> The Unicode table used to generate `scripttabl.txt` does not specify aliases for the script tags `Hans` and `Hant`. Therefore these tags are omitted from `scripttabl.txt` (just as all other script tags for which no alias is defined).

Yes, indeed, but this seemed surprising to me because I didn't understand how were these scripts different from e.g. `Tang` (which is another Chinese writing system, AFAIK, albeit a historical one).

> > Of course, in a way it's lucky that we don't, because if we did return `zh_CN@Hans` as the POSIX locale name, we wouldn't find `zh_CN` translations for it due to the current code using string comparisons.
> 
> Typically, the script alias is not used in POSIX locale names, if it is equal to the default script of the language in the specified region.

Right, which is why just ignoring it works for now. But if wanted to be able to handle `zh-Hant-CN` or (geopolitically more likely?) `zh-Hans-TW`, this wouldn't work. This is obviously a much less important problem, of course, so for now I'll just merge this fix.

Thanks for the review!

vslavik:
I need a bit more time to properly review it. At first (or second...) glance, I don't understand why is POSIX locale even involved on macOS.

That said, the commits in this PR themselves are obviously correct and I have almost [^1] no issue with them. I'm just not sure that they truly and robustly fix #24396 (I suspect not) and I'd like to understand the exact behavior with a debugger first. For example:

0. Why does it even go through this code on Mac?
1. Are commonly used minimal tags, `zh-Hans` and `zh-Hant` handled correctly? (Maximized `zh-Hans-CN` is rare.)
2. Does loading translations named `zh-Hans` work or does it have to be `zh_CN`? (wx supports macOS's `.lproj` resource folder and they use `zh-Han[st].lproj`, so they too need to be checked.)
3. Does it handle `zh-Hant-CN` correctly (mentioned by @vadz in the other issue; it should use Traditional for translation purposes, but not zh_TW *locale*)?
4. Does it fail to load Traditional Chinese translation if Simplified Chinese is unavailable?

While https://github.com/wxWidgets/wxWidgets/pull/24399/commits/f4fb6845682c44438fe5b95b0bd5f52ba9e2a5dd fixes a clear bug, I'm not so sure it's the root cause. Based on the #24396 discussion, this manifested itself as loading Traditional Chinese translation, which echoes what I observed in #24297. In both cases, wxTranslations would fail to find exact match, would reduce the language to just the minimal form of `en` or `zh` and then 
1. pick `en_GB` instead of using msgid `en`; or
2. pick `zh_TW` instead of falling back to English

[This is done on purpose here.](https://github.com/wxWidgets/wxWidgets/blob/b03c3416c87e799cdd086be49761913e994451bd/src/common/translation.cpp#L146) I was aware of that code, but decided against opening that can of worms in #24297 as it seemed harmless enough; I think #24396 proves that opinion wrong. The correct algorithm is to omit this "expanded match" altogether; or, if wx wishes to accommodate people who wrongly use e.g. `de_DE.mo` instead of `de.mo`, to minimize subtags on both available and preferred lists [^2].

> I also think that we need to change wxTranslations::DoGetBestAvailableTranslation() to compare wxLocaleIdent objects rather than using string comparisons, but I haven't done this yet.

Not sure what the rationale is here, but doing that would too make me uneasy because of the extra parsing. 

> But if wanted to be able to handle zh-Hant-CN or (geopolitically more likely?) zh-Hans-TW, this wouldn't work.

This is because it's trying to handle mapping between *languages* and *locales* and there is no 1:1 mapping. `zh-Hans-CN` 's far more commonly used equivalent is `zh-Hans` language, for which there is no POSIX locale; `zh_CN` is used as a substitute, but it's not exactly the same thing. This is more apparent with `zh-Hant`: it is traditionally mapped to `zh_TW`, but the *language* is used in multiple *locales*: zh_TW, zh_SG, zh_HK.

I'm pretty sure Chinese just has to be special-cased in `wxTranslations::DoGetBestAvailableTranslation()`.

[^1]: The one issue I do have is with https://github.com/wxWidgets/wxWidgets/pull/24399/commits/30aba7c966c831a58532b62f22176473cacb5a34: I feel uneasy about adding more parsing and interpretation there, particularly with something as heuristic as `FromTag()` is. It was a mistake to add even the minimal validation there before. GNU does intentionally limit this to languages: *language* is not *locale* and confusing the two doesn't end well (see history of why wxTranslations exists).

[^2]: Doing that however requires CLDR likely subtags data and implementing minimization.

vadz:
> I need a bit more time to properly review it. At first (or second...) glance, I don't understand why is POSIX locale even involved on macOS.

Sorry for the confusion, it isn't really involved. POSIX is only involved inasmuch as I used Linux, with `WXLANGUAGE=zh-Hans-CN` for testing.

> That said, the commits in this PR themselves are obviously correct and I have almost [1](#user-content-fn-1-6cb7fc42275fa7c5a03437222303def6) no issue with them. I'm just not sure that they truly and robustly fix #24396 (I suspect not)

It almost certainly doesn't because it just ignores the script instead of taking it into account. But, at least, it's not completely broken when the script is present any more, even when it's the default script for the chosen language (as is the case here).

> and I'd like to understand the exact behavior with a debugger first. For example:
> 
>     0. Why does it even go through this code on Mac?

Under Mac `wxUILocaleImpl::GetPreferredUILanguages()` returns the string with the script part. I think this is correct, in spite of the function name, because we shouldn't just lose this information. And the name is already inappropriate even without taking script into account because the returned string always could contain the locale (`xx_YY`) and not just the language.

>     1. Are commonly used minimal tags, `zh-Hans` and `zh-Hant` handled correctly?

I'd expect `zh-Hans` to be handled correctly because it should be the same as `zh` and then map to `zh_CN`, but not `zh-Hant` -- which should also be the same as `zh`, and so _fail_ to map to `zh_TW`. I didn't test this however.
 
> (Maximized `zh-Hans-CN` is rare.)

I _believe_ it's the default value for `[NSLocale preferredLanguages]` on Chinese systems, but I could be wrong.

>     2. Does loading translations named `zh-Hans` work or does it have to be `zh_CN`? (wx supports macOS's `.lproj` resource folder and they use `zh-Han[st].lproj`, so they too need to be checked.)

I think it has to be `zh_CN`.

>     3. Does it handle `zh-Hant-CN` correctly (mentioned by @vadz in the other issue; it should use Traditional for translation purposes, but not zh_TW _locale_)?

I'm pretty sure it does not.

>     4. Does it fail to load Traditional Chinese translation if Simplified Chinese is unavailable?

Do we really want this? If so, this makes Chinese unlike any other language where we always prefer to use some variant of the language, even using a different script (or country), rather than a different language.

> While [f4fb684](https://github.com/wxWidgets/wxWidgets/commit/f4fb6845682c44438fe5b95b0bd5f52ba9e2a5dd) fixes a clear bug, I'm not so sure it's the root cause. Based on the #24396 discussion, this manifested itself as loading Traditional Chinese translation, which echoes what I observed in #24297. In both cases, wxTranslations would fail to find exact match, would reduce the language to just the minimal form of `en` or `zh` and then
> 
>     1. pick `en_GB` instead of using msgid `en`; or
> 
>     2. pick `zh_TW` instead of falling back to English
> 
> 
> [This is done on purpose here.](https://github.com/wxWidgets/wxWidgets/blob/b03c3416c87e799cdd086be49761913e994451bd/src/common/translation.cpp#L146)

Yes, we do it on purpose and I'm pretty sure we want to do it for many languages. E.g. a person with `fr-CH` preferred UI language would certainly want to use `fr-FR` or even `fr-BE` (which seems unlikely because who would provide Belgian French translations without providing French ones, but who knows) rather than `en`, which is the only other possibility. It's not really that simple even for European languages, of course, e.g. `de-CH` is a different language from `de-DE` but I do believe we still want to provide such fallback even in this case.

> I was aware of that code, but decided against opening that can of worms in #24297 as it seemed harmless enough; I think #24396 proves that opinion wrong. The correct algorithm is to omit this "expanded match" altogether; or, if wx wishes to accommodate people who wrongly use e.g. `de_DE.mo` instead of `de.mo`, to minimize subtags on both available and preferred lists [2](#user-content-fn-2-6cb7fc42275fa7c5a03437222303def6).

I'm pretty sure that fallback logic could and should be improved, but between the current logic and not having any fallback at all, I would still choose the current one.

I also don't think it's by any means "wrong" to use `de_DE.mo`, it's perfectly ok to have (perhaps slightly) different translations for the same language spoken in different countries. `en_UK` and `en_US` are one obvious example, but looking at my (Debian) system right now I see that e.g. `libgweather-4.0.mo` exists for all of the following locales: `es_AR`, `es_CL`, `es_CO`, `es_CR`, `es_DO`, `es_EC`, `es_ES`, `es_GT`, `es_HN`, `es_MX`, `es_NI`, `es_PA`, `es_PE`, `es_PR`, `es_SV`, `es_UY`, `es_VE`.

> > I also think that we need to change wxTranslations::DoGetBestAvailableTranslation() to compare wxLocaleIdent objects rather than using string comparisons, but I haven't done this yet.
> 
> Not sure what the rationale is here, but doing that would too make me uneasy because of the extra parsing.

Yes, this only makes sense if we want to implement fallback.

> > But if wanted to be able to handle zh-Hant-CN or (geopolitically more likely?) zh-Hans-TW, this wouldn't work.
> 
> This is because it's trying to handle mapping between _languages_ and _locales_ and there is no 1:1 mapping. `zh-Hans-CN` 's far more commonly used equivalent is `zh-Hans` language, for which there is no POSIX locale; `zh_CN` is used as a substitute, but it's not exactly the same thing. This is more apparent with `zh-Hant`: it is traditionally mapped to `zh_TW`, but the _language_ is used in multiple _locales_: zh_TW, zh_SG, zh_HK.
> 
> I'm pretty sure Chinese just has to be special-cased in `wxTranslations::DoGetBestAvailableTranslation()`.

There are surely other examples like this... If we start special-casing things, we're never going to finish.

Sorry, I've got to run now, but, to summarize: I realize that our current code is very far from perfect but I think we all agree that this is a (marginal) improvement and I hope I could explain why it fixes the problem under macOS. Where we disagree is that you seem to think that we should never use anything but the language names for the translations, while I believe that we have no choice but to do it because it makes sense and is also very common practice.

vslavik:
> but I think we all agree that this is a (marginal) improvement 

Again, yes, [f4fb684](https://github.com/wxWidgets/wxWidgets/pull/24399/commits/f4fb6845682c44438fe5b95b0bd5f52ba9e2a5dd) is obviously correct fix for an obvious bug ant this should be merged. It's relation to #24396 feels coincidental, though.

> Where we disagree is that you seem to think that we should never use anything but the language names for the translations, while I believe that we have no choice but to do it because it makes sense and is also very common practice.

I must have done a very poor job of explaining myself (which is mildly annoying, because I've spent non-trivial amount of time on it), because that's not it at all. I'll try again once I had the chance to test things and time to respond in depth.

utelle:
> > The Unicode table used to generate `scripttabl.txt` does not specify aliases for the script tags `Hans` and `Hant`. Therefore these tags are omitted from `scripttabl.txt` (just as all other script tags for which no alias is defined).
> 
> Yes, indeed, but this seemed surprising to me because I didn't understand how were these scripts different from e.g. `Tang` (which is another Chinese writing system, AFAIK, albeit a historical one).

The `Tang` script is a non-default script and the Unicode table contains an alias for it. So, theoretically support for it could be installed using the POSIX modifier `@tangut`.

The problem with the POSIX locale tags is that the meaning of the modifier part is not clearly defined. It can be anything: a currency (like `@euro`), a script (like `@latin`), a region (like `@valencia`), or a sort order etc.

If you inspect under Ubuntu the language support for _Chinese_, a modifier is not used for a single variant. You just find `zh_CN`, `zh_HK`, `zh_SG`, and `zh_TW`. AFAIK the simplified Chinese script is used for China mainland and Singapore, while the traditional Chinese script is used for Hongkong and Taiwan. I have no idea, if and how it is possible to install and use a deviating script.

IMHO using the POSIX variant of the locale tags is not well suited to identify the language (and possibly script) for translations, due to potentially missing POSIX script aliases. Other tag variants do not have this specific problem, because script names are used as-is.

> > > Of course, in a way it's lucky that we don't, because if we did return `zh_CN@Hans` as the POSIX locale name, we wouldn't find `zh_CN` translations for it due to the current code using string comparisons.

Instead of using the POSIX locale name it may be a better idea to use the accessors of the `wxLocaleIdent` class to identify language, region, and script.

> > Typically, the script alias is not used in POSIX locale names, if it is equal to the default script of the language in the specified region.
> 
> Right, which is why just ignoring it works for now. But if wanted to be able to handle `zh-Hant-CN` or (geopolitically more likely?) `zh-Hans-TW`, this wouldn't work. This is obviously a much less important problem, of course, so for now I'll just merge this fix.

Under Windows you have support for `zh-Hans`, `zh-Hant` (for selecting the script independently of the region), and `zh-Hans-HK` and `zh-Hans-MO` (for selecting a non-default script for specific regions). Under macOS the situation is most likely similar. Currently, the POSIX variant of the locale tags has no means to express these selections.

However, wxWidgets expects directory names for language catalogs in POSIX syntax. That is, we will have to discuss how the currently unsupported variations for Chinese should be handled.

vadz:
> If you inspect under Ubuntu the language support for _Chinese_, a modifier is not used for a single variant. You just find `zh_CN`, `zh_HK`, `zh_SG`, and `zh_TW`. AFAIK the simplified Chinese script is used for China mainland and Singapore, while the traditional Chinese script is used for Hongkong and Taiwan. I have no idea, if and how it is possible to install and use a deviating script.

Maybe we should just abandon the idea of supporting non-default script/country combinations then. But I still wonder what should we use for `zh-Hans-TW` if it's possible to set such locale under macOS (and I think it is), i.e. should we prefer mapping it to `zh_CN` or `zh_TW`?

> IMHO using the POSIX variant of the locale tags is not well suited to identify the language (and possibly script) for translations, due to potentially missing POSIX script aliases. Other tag variants do not have this specific problem, because script names are used as-is.

We certainly could allow script names in the translation directories...

> > > > Of course, in a way it's lucky that we don't, because if we did return `zh_CN@Hans` as the POSIX locale name, we wouldn't find `zh_CN` translations for it due to the current code using string comparisons.
> 
> Instead of using the POSIX locale name it may be a better idea to use the accessors of the `wxLocaleIdent` class to identify language, region, and script.

Yes, this is what I meant when I wrote

> I also think that we need to change `wxTranslations::DoGetBestAvailableTranslation()` to compare `wxLocaleIdent` objects rather than using string comparisons, but I haven't done this yet.

in the PR description.

> Under Windows you have support for `zh-Hans`, `zh-Hant` (for selecting the script independently of the region),

This almost certainly still doesn't work correctly, even with this PR... We need some way to determine that a locale using `Hans` script should prefer `CN` while the one using `Hant` should prefer something else (but here we'd just have to choose randomly between `TW`, `HK` etc I'm afraid). Is there anything in the Unicode data allowing us to do this? We need to have at least a way to find the default script for the country.

> and `zh-Hans-HK` and `zh-Hans-MO` (for selecting a non-default script for specific regions). Under macOS the situation is most likely similar. Currently, the POSIX variant of the locale tags has no means to express these selections.
> 
> However, wxWidgets expects directory names for language catalogs in POSIX syntax. That is, we will have to discuss how the currently unsupported variations for Chinese should be handled.

AFAICS there are 2 problems here:

1. Specifying the script should affect the choice of the country if it's not specified explicitly.
1. Specifying both the script and the country may need to override the country if it's default script is different from the given one.

As implied by the word "may", I'm not even sure if we need to do this, so it's probably better to start by fixing the first problem as it's at least clear what needs to be done there.

utelle:
> Maybe we should just abandon the idea of supporting non-default script/country combinations then. But I still wonder what should we use for `zh-Hans-TW` if it's possible to set such locale under macOS (and I think it is), i.e. should we prefer mapping it to `zh_CN` or `zh_TW`?

Well, looking at the language/locale table we currently have in wxWidgets you can see that in principle some non-default combinations are included (albeit they probably won't work properly in the POSIX variant due to the missing aliases for `Hans` and `Hant`):

| wx ID | BCP47 | Canonical | Canonical with region |
| --- | --- | --- | --- |
| wxLANGUAGE_CHINESE                                              | zh                | zh                        | zh_TW
| wxLANGUAGE_CHINESE_SIMPLIFIED_EXPLICIT           | zh-Hans        | zh@Hans                   | zh_CN
| wxLANGUAGE_CHINESE_TRADITIONAL_EXPLICIT       | zh-Hant        | zh@Hant | zh_TW
| wxLANGUAGE_CHINESE_CHINA                                  | zh-CN         | zh_CN | 
| wxLANGUAGE_CHINESE_HONGKONG                        | zh-HK          | zh_HK | 
| wxLANGUAGE_CHINESE_MACAO                                | zh-MO         | zh_MO | 
| wxLANGUAGE_CHINESE_SIMPLIFIED_HONGKONG    | zh-Hans-HK  | zh_HK@Hans | 
| wxLANGUAGE_CHINESE_SIMPLIFIED_MACAO            | zh-Hans-MO | zh_MO@Hans | 
| wxLANGUAGE_CHINESE_SINGAPORE                         | zh-SG            | zh_SG | 
| wxLANGUAGE_CHINESE_TAIWAN                               | zh-TW           | zh_TW | 

Maybe we should handle the scripts `Hans` and `Hant` in the mapping from script id to alias and vice versa explicitly by mapping them to themselves (that is, `Hans` <=> `Hans`, and `Hant` <=> `Hant`).

> > IMHO using the POSIX variant of the locale tags is not well suited to identify the language (and possibly script) for translations, due to potentially missing POSIX script aliases. Other tag variants do not have this specific problem, because script names are used as-is.
> 
> We certainly could allow script names in the translation directories...

If the accessors of `wxLocaleIdent` are used instead of the POSIX tag variant, this should not be too hard to accomplish. However, currently the script alias is used for the directory names, so that the mapping `Hans` <=> `Hans`, and `Hant` <=> `Hant` would be necessary here, too.

> > I also think that we need to change `wxTranslations::DoGetBestAvailableTranslation()` to compare `wxLocaleIdent` objects rather than using string comparisons, but I haven't done this yet.

Yes, that should improve the situation.

> > Under Windows you have support for `zh-Hans`, `zh-Hant` (for selecting the script independently of the region),
> 
> This almost certainly still doesn't work correctly, even with this PR... We need some way to determine that a locale using `Hans` script should prefer `CN` while the one using `Hant` should prefer something else (but here we'd just have to choose randomly between `TW`, `HK` etc I'm afraid).

The wxWidgets language/locale table already includes such mappings. If no region is given in the language/locale tag, I'd stick to `CN` resp `TW`.

> Is there anything in the Unicode data allowing us to do this? We need to have at least a way to find the default script for the country.

There is no such thing as a _default script_ for a country. At most there is a default script for a language used in a country. However, I'm currently not aware of a data source providing this information. (Maybe such information could be extracted from the file `misc/languages/data/windows/win-locale-table-win.txt`, but I doubt that it could be done in a really reliable way.)

> > and `zh-Hans-HK` and `zh-Hans-MO` (for selecting a non-default script for specific regions). Under macOS the situation is most likely similar. Currently, the POSIX variant of the locale tags has no means to express these selections.
> > However, wxWidgets expects directory names for language catalogs in POSIX syntax. That is, we will have to discuss how the currently unsupported variations for Chinese should be handled.
> 
> AFAICS there are 2 problems here:
> 
> 1. Specifying the script should affect the choice of the country if it's not specified explicitly.

For _Chinese_ we have such mappings in the language/locale table.

> 2. Specifying both the script and the country may need to override the country if it's default script is different from the given one.

The script should take precedence over the region, I guess. Then the solution is to drop the given country code and use the mapping to a region in the language/locale table.

> As implied by the word "may", I'm not even sure if we need to do this, so it's probably better to start by fixing the first problem as it's at least clear what needs to be done there.

I agree.

craftyjon:
@vadz I'm a bit unclear on where this PR is going; we would like to fix this issue in KiCad ideally by taking in an upstream commit, but if it will take a while to reach consensus on how to fix this, we will probably put a temporary workaround in our fork (maybe take this PR as-is, maybe something else?) so that at least this bug is fixed for KiCad in our next release.

vadz:
> @vadz I'm a bit unclear on where this PR is going

I think it will be merged in its current form, with more changes possibly coming later. If there are no objections from @vslavik (Vaclav, I _think_ that you don't object to merging this, but I'd like to avoid any misunderstanding and so would appreciate an explicit confirmation) I'd like to do it soon. I'll also backport this to 3.2 after merging into master.

vslavik:
>  If there are no objections from @vslavik (Vaclav, I think that you don't object to merging this, but I'd like to avoid any misunderstanding and so would appreciate an explicit confirmation) I'd like to do it soon.

I'm sorry, I overlooked this request. Yes, this should be merged.

vslavik (approved):

vadz:
Thanks, I'm merging this but please let's discuss how would you like things to work when you have time later because I still don't know how exactly did I misunderstand you. TIA!

Issue #23544, opened 2023-05-13
wxTheClipboard doesnt work under WSL2 (wslg)


elfmz:
Under Windows 11/WSL2 wxClipboard (wxTheClipboard instance) doesn't copy-paste across applications - data copied/pasted only within application itself. However when trying to create wxTextCtrl and copy/paste text by pressing ctrl+c/v inside of it then clipboard works properly.

Patch or snippet allowing to reproduce the problem:
Here is minimalistic example: https://pastecode.io/s/ksm4we9u
It creates window with wxTextCtrl and wxButton.
Button click handler gets text from clipboard pritns it to console and then sets new text string "1234567890" into clipboard.
So when i try to copy/paste between other applications and wxTextCtrl using Ctrl+C/Ctrl+V hotkeys - everything works as expected - text is being get/set across this and other application correctly. However when trying to click the button - then text is not transferred to other application, i.e. clipboard starting to behave like to be 'independently private' for my process.

### Platform and version information
This only happens only under Windows 11/WSL2, no problems under vanilla Ubuntu-s.
However other applications like gedit under WSL seems to work correctly.

Reproduced with wxWidgets GTK3 v3.0.5 (from repository) and v3.2 (built manually).

vadz:
Sorry, I don't have WSL2 and am unlikely to have it, so I can't debug this. I'd recommend starting by setting environment variable `WXTRACE="clipboard"` and checking what is the difference (if any) with the output under normal Linux systems.

elfmz:
Here're logs, what was done: ran app and copied in browser word 'Stable' then clicked button twice. After that pressed in browser Ctrl+V and as a result in Ubuntu got 12345678 but in Windows there is still 'Stable ' pasted in browser, while test application gets 12345678 on second button press.
[ubuntu.log](https://github.com/wxWidgets/wxWidgets/files/11470071/ubuntu.log)
[wsl.log](https://github.com/wxWidgets/wxWidgets/files/11470072/wsl.log)

vadz:
Looking at the diff between the logs it seems that WSL provides `text/plain;charset=utf-8` format instead of `UTF8_STRING`. It should be relatively simple to add support for this format too, but I have no idea why it doesn't use the de facto standard here.

KeyWorksRW:
It's worth nothing that even a very simple call to gtk fails to place the text in the Windows clipboard:

```c++
[...]

PR #24400, opened 2024-03-13
use different clipboard targets for text on Wayland

### LINKED ISSUE #24391: Pasting text from Telegram Desktop not working under Wayland
far2l (Linux port of Windows console file manager called Far Manager or FAR) is using wxWidgets for interface rendering and system clipboard interaction. Unfortunately, on Wayland systems (like Ubuntu 23.10) clipboard copy-paste is not working between far2l and Telegram Desktop.

Steps to reproduce:
1. Install Ubuntu 23.10
2. Install far2l-gui from ppa
3. Run far2l
4. Download, unpack and run Telegram Desktop
5. Copy any string from Telegram Desktop to clipboard
6. Try to paste to far2l
7. Try to copy text from any other app (for example, Firefox)
8. Try to paste to far2l

Workaround:
```
GDK_BACKEND=x11 far2l
```

Corresponding far2l issue:
https://github.com/elfmz/far2l/issues/2053

unxed:
Fixes #24391 (and probably #23544).

### DIFF
--- a/src/gtk/dataobj.cpp
+++ b/src/gtk/dataobj.cpp
@@ -103,15 +103,19 @@ void wxDataFormat::SetType( wxDataFormatId type )
     m_type = type;
 
     if (m_type == wxDF_UNICODETEXT)
+    {
         if (getenv("WAYLAND_DISPLAY"))
             m_format = g_textAtomWayland;
         else
             m_format = g_textAtom;
+    }
     else if (m_type == wxDF_TEXT)
+    {
         if (getenv("WAYLAND_DISPLAY"))
             m_format = g_altTextAtomWayland;
         else
             m_format = g_altTextAtom;
+    }
     else
     if (m_type == wxDF_BITMAP)
         m_format = g_pngAtom;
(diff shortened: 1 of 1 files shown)


unxed:
Still need some debug

vadz:
Sorry, I didn't have time to finish my changes yet, but while I did plan to to something like this, I don't think it's the right approach for 2 reasons:

1. Checking the environment variable is not the right way to test for Wayland, this would fail in too many cases (GTK2 app running in Wayland environment, GTK3 one running with `GDK_BACKEND=x11`, GTK3 one running in `env -i`, and probably more). We have `wxGTKImpl::IsWayland()` which does it correctly, but it needs a display pointer.
2. It's not really obvious to me that we need to test for Wayland at all and, in fact, I don't think we should. Instead, we should just always accept all known text formats, I don't see any possible harm in doing this.

I wanted to do this soon, but other things came up and so I wasn't able to do it yet. If you plan to do it like this, please let me know and I'll wait until you do, otherwise I'll try to do it myself a.s.a.p., hopefully this weekend. Thanks!

unxed:
Already fixed on our side, so you are welcome to implement this the best way possible! Thanks a lot.

vadz:
Replaced with #24423.

Issue #24393, opened 2024-03-10
Make it possible to use CEF in universal macOS binaries


vadz:
Right now this fails because building under x64 builds x64-only CEF wrapper (and downloads x64 version of CEF shared library), so linking doesn't find arm64 symbols.

I hope it is possible to use CEF in this case because Chrome itself does come up as a universal binary under Mac, but right now I don't really know how to do it: should we download both files and manually lipo all the shared libraries together?

And what about the support files, apparently at least the snapshot one is [architecture-specific](https://magpcss.org/ceforum/viewtopic.php?f=6&t=18344)?

Any thoughts/hints would be welcome!

vadz:
I don't see any other way than using `lipo` manually, CEF CMake makefiles explicitly override `CMAKE_OSX_ARCHITECTURES` and set it to `CMAKE_HOST_SYSTEM_PROCESSOR`, so we can't even build the wrapper library as a universal one.

I'll try to do this in my baby-cmake...

vadz:
Without much surprise, my CMake-fu failed me, but I could at least create #24402 which solves this is another way.

PR #24402, opened 2024-03-15
Support building CEF for non-default architecture under macOS

vadz:
While this still doesn't allow to build universal CEF binaries, with these changes it's at least possible to build wx twice and then lipo everything together, so this PR provides a fallback solution for #24393.

For the reference, here is how I do it for my application

```sh
set -e

ARCH=`uname -m`
if [[ `uname` == 'Darwin' ]]; then
	case "$ARCH" in
		arm64)
			OTHER_ARCH=x86_64
			;;

		x86_64)
			OTHER_ARCH=arm64
			;;

		*)
			echo "Unsupported macOS architecture: $ARCH"
			exit 1
			;;
	esac

	# Build for the current architecture.
	cmake -S $MYAPP_SRC -B $MYAPP_BUILD -GNinja -DCMAKE_BUILD_TYPE=Release
	cmake --build $MYAPP_BUILD

	# Then build for the other.
	cmake -S $MYAPP_SRC -B $MYAPP_BUILD-$OTHER_ARCH -GNinja -DCMAKE_OSX_ARCHITECTURES=$OTHER_ARCH -DCMAKE_BUILD_TYPE=Release
	cmake --build $MYAPP_BUILD-$OTHER_ARCH

	# And finally glue the two together.
	myapp_subdir='MyApp/MyApp.app/Contents'
	cef_subdir="$myapp_subdir/Frameworks/Chromium Embedded Framework.framework"
	for f in \
"$cef_subdir/Chromium Embedded Framework" \
"$cef_subdir/Libraries/libEGL.dylib" \
"$cef_subdir/Libraries/libGLESv2.dylib" \
"$cef_subdir/Libraries/libvk_swiftshader.dylib" \
"$myapp_subdir/MacOS/MyApp" \
"$myapp_subdir/Frameworks/MyApp Helper.app/Contents/MacOS/MyApp Helper" \
	; do
		lipo -create "$MYAPP_BUILD/$f" "$MYAPP_BUILD-$OTHER_ARCH/$f" -output "$MYAPP_BUILD/$f.tmp"
		mv "$MYAPP_BUILD/$f.tmp" "$MYAPP_BUILD/$f"
	done

	# Special case of arch-specific v8 snapshot file.
	cp "$MYAPP_BUILD-$OTHER_ARCH/$cef_subdir/Resources/v8_context_snapshot.$OTHER_ARCH.bin" "$MYAPP_BUILD/$cef_subdir/Resources"

	# This file differs between the two architectures, but has a fixed name,
	# so we can only remove it. FWIW it is not present in CEF included in
	# /Application/Google Chrome.app, so it's probably the right thing to do.
	rm "$MYAPP_BUILD/$cef_subdir/Resources/snapshot_blob.bin"
fi
```


MaartenBent:
Out of interest, have you checked what happens if you disable the `CMAKE_OSX_ARCHITECTURES` override by CEF?
Can it then successfully build the wrapper? If so, we could block CEF from changing it like I did for `CMAKE_CONFIGURATION_TYPES`.
Then you only have to deal with the single architecture of the cef library.

vadz:
> Out of interest, have you checked what happens if you disable the `CMAKE_OSX_ARCHITECTURES` override by CEF? Can it then successfully build the wrapper?

No, I didn't test this, but I guess it should be able to do it for a single architecture as the effect of setting `PROJECT_ARCH` is about the same, see its use in `3rdparty/cef/cmake/cef_variables.cmake`. However it still wouldn't be able to build the universal binaries without patching its CMake files (in addition to lipoing the downloaded Chromium files together as shown above).

> If so, we could block CEF from changing it like I did for `CMAKE_CONFIGURATION_TYPES`. Then you only have to deal with the single architecture of the cef library.

I'm not sure what do we gain compared to the changes here, could you please explain if I'm missing something?

MaartenBent:
It will be in addition to (some of) these changes.

If I understand correctly, now you build x86_64 and arm64 separately. And then lipo all the libraries (wx and cef) together.
But (I assume) wx already supports building universal libraries, it is just `libcef_dll_wrapper` that is the problem.
More specifically, `libcef_dll_wrapper` sets `CMAKE_OSX_ARCHITECTURES` so it overwrites your `x86_64;arm64`.

My suggestion is to block `libcef_dll_wrapper` from changing `CMAKE_OSX_ARCHITECTURES`. So it uses whatever architectures that you have set.
Then you should be able to build wx and `libcef_dll_wrapper` as universal libraries.

And you only need to lipo the CEF Library.

You could try adding `set_readonly(CMAKE_OSX_ARCHITECTURES)` after
https://github.com/wxWidgets/wxWidgets/blob/c4112f19a71ecb4a9c5840b5b19a209eb96173b0/build/cmake/lib/webview_chromium/CMakeLists.txt#L103
and see if it works.

The value of `PROJECT_ARCH` doesn't really matter then, the only other place it is used is in `PRINT_CEF_CONFIG`.

vadz:
> If I understand correctly, now you build x86_64 and arm64 separately. And then lipo all the libraries (wx and cef) together.

Yes, exactly.

> But (I assume) wx already supports building universal libraries,

Yes, it does.

> it is just `libcef_dll_wrapper` that is the problem. More specifically, `libcef_dll_wrapper` sets `CMAKE_OSX_ARCHITECTURES` so it overwrites your `x86_64;arm64`.

This is a problem, the other problem is that there are no universal CEF binaries, so we'd need to download both versions and lipo them together before building the wrapper. But this is doable -- I had, in fact, started doing this before I realize that the wrapper always wanted to build a single-architecture library.

> My suggestion is to block `libcef_dll_wrapper` from changing `CMAKE_OSX_ARCHITECTURES`. So it uses whatever architectures that you have set. Then you should be able to build wx and `libcef_dll_wrapper` as universal libraries.

OK, it's probably worth trying doing this because _not_ doing everything I do in the shell fragment above is surely going to be nicer for the library users.

> And you only need to lipo the CEF Library.

Also download both versions in different directories. But I should probably be able to do this.

> The value of `PROJECT_ARCH` doesn't really matter then, the only other place it is used is in `PRINT_CEF_CONFIG`.

This is still going to be confusing, I wonder if we could set it to the actual architectures -- AFAICS it shouldn't break anything as their CMakefile doesn't do anything for unknown values of `PROJECT_ARCH`.

vadz:
I've finally managed to get something which has a chance of working -- at least building wx and the sample as universal app works now. Please let me know if you see anything particularly egregious in what I did, otherwise I'd like to merge this and stop having CMake nightmares.

build/cmake/lib/webview_chromium/CMakeLists.txt
```diff
@@ -62,8 +70,149 @@ if(NOT CEF_ROOT)
...
+                set(lipo_command
+                    lipo -create "Release/${cef_bin_file}"
+                )
+
+                foreach(cef_platform IN LISTS other_cef_platforms)
+                    list(APPEND lipo_command "${cef_platform}/Release/${cef_bin_file}")
+                endforeach()
+
+                list(APPEND lipo_command "-output")
+                list(APPEND lipo_command "Release/${cef_bin_file}.tmp")
```
MaartenBent, line 128: Why is the `.tmp` file and rename needed?

build/cmake/lib/webview_chromium/CMakeLists.txt
```diff
@@ -62,8 +70,149 @@ if(NOT CEF_ROOT)
...
+            endif()
+
+            wx_download_cef(${CEF_PLATFORM})
+        endforeach()
+
+        # Now lipo all the binaries together unless this is a degenerate case
+        # in which CMAKE_OSX_ARCHITECTURES contains only one element.
+        if(other_cef_platforms)
+            set(cef_binaries
+                "cef_sandbox.a"
```
MaartenBent, line 111: Is `cef_sandbox.a` needed to lipo? Because we turn the sandbox off.

build/cmake/lib/webview_chromium/CMakeLists.txt
```diff
@@ -62,8 +70,149 @@ if(NOT CEF_ROOT)
...
     endif()
+endfunction()
+
+if(NOT CEF_ROOT)
+    include(cef_version_info.cmake)
+
+    # Note that CMAKE_OSX_ARCHITECTURES is always defined under macOS, but is
+    # empty if not explicitly specified, so do _not_ use if(DEFINED ...) here.
+    if(CMAKE_OSX_ARCHITECTURES)
+        foreach(osx_arch IN LISTS CMAKE_OSX_ARCHITECTURES)
```
MaartenBent, line 81: I think you can simplify all the foreach a bit to just `foreach(item ${list})`.

build/cmake/lib/webview_chromium/CMakeLists.txt
```diff
@@ -33,13 +31,23 @@ find_path(CEF_ROOT
...
+# variables containing the common root for all of them.
+set(CEF_DOWNLOAD_ROOT ${CMAKE_CURRENT_BINARY_DIR}/cef-download)
+set(CEF_SOURCE_ROOT ${CMAKE_CURRENT_BINARY_DIR}/cef-source)
+set(CEF_BUILD_ROOT ${CMAKE_CURRENT_BINARY_DIR}/cef-build)
 
-if(NOT CEF_ROOT)
-    message("Downloading CEF binary distribution...")
+# But when we don't need multiple directories, just use them directly.
+set(CEF_DOWNLOAD_DIR ${CEF_DOWNLOAD_ROOT})
+set(CEF_SOURCE_DIR ${CEF_SOURCE_ROOT})
```
MaartenBent, line 42: Variable is never used. Either it is not needed, or you use `CEF_SOURCE_ROOT` incorrectly.

build/cmake/lib/webview_chromium/CMakeLists.txt
```diff
@@ -33,13 +31,23 @@ find_path(CEF_ROOT
...
+set(CEF_BUILD_DIR ${CEF_BUILD_ROOT})
```
MaartenBent, line 43: Variable is never used. Either it is not needed, or you use `CEF_BUILD_ROOT` incorrectly.

MaartenBent (commented):
Looks good. I only reviewed the code, I'll try to test it later today.

MaartenBent:
When configuring and with `CEF_SHOW_RESULTS` enabled, the shown `Project architecture` is wrong. You added this to the 3th commit but it was removed in the last one:
https://github.com/wxWidgets/wxWidgets/blob/c2a8b111c92384cde663ceca7da31a66b59f72c5/build/cmake/lib/webview_chromium/CMakeLists.txt#L62-L64

But also this doesn't fully work because it is only set inside `if(NOT CEF_ROOT)`.
So maybe this:
```cmake
if(CMAKE_OSX_ARCHITECTURES)
    set(PROJECT_ARCH ${CMAKE_OSX_ARCHITECTURES}) 
endif()
```

When building I get a few warnings (though png and tiff have a few of these too):
```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: for architecture: x86_64 file: libcef_dll_wrapper.a(cef_lock.cc.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: for architecture: x86_64 file: libcef_dll_wrapper.a(libcef_dll_wrapper2.cc.o) has no symbols
```

And when linking many:
```
ld: warning: object file (/Users/maartenbent/dev/build/wxwidgets/libs/webview_chromium/cef-build/libcef_dll_wrapper/libcef_dll_wrapper.a[x86_64][2](shutdown_checker.cc.o)) was built for newer 'macOS' version (10.15) than being linked (10.10)
```
Should we set this to our version and block it from changing (with `CEF_TARGET_SDK`)?

build/cmake/lib/webview_chromium/CMakeLists.txt
vadz: I don't think lipo can modify files in place, so this creates a temporary file and then renames it to the original one, how else can we do it?

BTW, one problem with this is that once lipo has been run, running it again will fail because it will complain about the other arch being already present in the first source file (which will be a fat one). I don't know if it's worth catering for this...

build/cmake/lib/webview_chromium/CMakeLists.txt
vadz: We don't use it currently but it seems better/safer to lipo all the binaries just in case...

build/cmake/lib/webview_chromium/CMakeLists.txt
vadz: I've copied this from another wx CMakefile but now I see that we actually use both styles. We probably should standardize on one of them, but I don't know which.

build/cmake/lib/webview_chromium/CMakeLists.txt
vadz: Sorry, why do you say this? It's used in `build/cmake/lib/webview_chromium/cef_download.cmake.in`, just as before. I.e. it's used by `configure_file()`.

Am I missing something?

build/cmake/lib/webview_chromium/CMakeLists.txt
vadz: Same as above.

vadz:
Thanks for reviewing and testing!

> When configuring and with `CEF_SHOW_RESULTS` enabled, the shown `Project architecture` is wrong. You added this to the 3th commit but it was removed in the last one:
> 
> https://github.com/wxWidgets/wxWidgets/blob/c2a8b111c92384cde663ceca7da31a66b59f72c5/build/cmake/lib/webview_chromium/CMakeLists.txt#L62-L64

I thought it wasn't needed any more, but I overlooked its use here...

> But also this doesn't fully work because it is only set inside `if(NOT CEF_ROOT)`. So maybe this:
> 
> ```cmake
> if(CMAKE_OSX_ARCHITECTURES)
>     set(PROJECT_ARCH ${CMAKE_OSX_ARCHITECTURES}) 
> endif()
> ```

I'm not sure setting `PROJECT_ARCH` to a list of architectures instead of a single one is not going to break something. Have you already tested that the above works by chance?
 
> When building I get a few warnings (though png and tiff have a few of these too):
> 
> ```
> /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: for architecture: x86_64 file: libcef_dll_wrapper.a(cef_lock.cc.o) has no symbols
> /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: for architecture: x86_64 file: libcef_dll_wrapper.a(libcef_dll_wrapper2.cc.o) has no symbols
> ```

Yes, but this is not new AFAIK and I have no idea what to do about them, so I just ignore them. Of course, any better suggestions would welcome!

> And when linking many:
> 
> ```
> ld: warning: object file (/Users/maartenbent/dev/build/wxwidgets/libs/webview_chromium/cef-build/libcef_dll_wrapper/libcef_dll_wrapper.a[x86_64][2](shutdown_checker.cc.o)) was built for newer 'macOS' version (10.15) than being linked (10.10)
> ```
> 
> Should we set this to our version and block it from changing (with `CEF_TARGET_SDK`)?

I think it really needs 10.15, so we should instead ensure that our own version is 10.15 when using CEF under Mac. I actually do this in my own CMakefile in the application using it.

vadz:
@MaartenBent Do you think there any changes that really must be done before merging this or should we merge it and then improve it later?

MaartenBent:
Its fine to merge.

You could add the `PROJECT_ARCH` thing. On macOS this variable isn't used for anything but the status message.

vadz:
Finally done now, thanks again for the review!

PR #24405, opened 2024-03-15
Add EnablePersistentStorage method to wxWebView

TcT2k:
Allows to disable persistent storage for a _wxWebView_.

The default is to enable it. When the webview is created without persistent storage the data is stored in memory and is lost when the webview is destroyed. In browsers this is equivalent to "private browsing" or "incognito mode".

**Note:** This is only implemented on the Edge, WebKit2GTK+ and macOS backends.


vadz:
Thanks! I've pushed a couple of minor corrections but otherwise this looks good to me and I'll merge it soon if there are no objections.

FWIW it should definitely be possible to implement support for this using CEF and I might do it in the future, but any contributions would be welcome too, of course.

PR #24407, opened 2024-03-16
Add EnablePersistentStorage method to wxWebSession

### LINKED ISSUE #23743: Use euphemeral NSURLSession?
With defaultSessionConfiguration NSURLSession uses local caches and cookies and there is very many cases when you don't want it. There is a special [NSURLSessionConfiguration ephemeralSessionConfiguration] that allows to avoid it. And it makes wxWebRequest usage more consistent across different OSes. Can euphemeral NSURLSession be added at least as an option?

TcT2k:
Currently only implemented for the macOS backend.

Default is changed to non persistent to be consistent with other backends

Fixes: #23743

### DIFF
--- a/include/wx/osx/private/webrequest_urlsession.h
+++ b/include/wx/osx/private/webrequest_urlsession.h
@@ -156,9 +156,9 @@ public:
     WX_wxWebSessionDelegate GetDelegate() { return m_delegate; }
 
 private:
-    WX_NSURLSession m_session;
+    WX_NSURLSession m_session = nullptr;
     WX_wxWebSessionDelegate m_delegate;
-    bool m_persistentStorageEnabled;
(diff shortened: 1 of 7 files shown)


vadz:
Thanks! I've pushed some very minor corrections that I intend to squash when merging, please let me know if you have any objections.

Also, do I understand correctly that this is an incompatible change, i.e. that sessions were persistent before but not any more? If so, we need to mention this in the change log (I'll do it).

Thanks again!

TcT2k:
Changes do look good.

Yes it's an incompatible change, but for consistency I think it makes sense.

vadz:
> Yes it's an incompatible change, but for consistency I think it makes sense.

Yes, I agree, but it still has to be mentioned. Done now and merged, thanks again!

PR #24408, opened 2024-03-17
Fix drawing of wxListCtrl last item in Windows 11 dark mode

### LINKED ISSUE #24024: Last inserted Item in a wxListCtrl is not visible in wxMSW when using dark mode
Windows 11-64bit
wxWidgets 3.3.0 master

The last inserted item in a wxListCtrl is not visible but you can select it.
This is also the case if only one item is inserted.

This happens only in wxMSW dark mode.

PBfordev:
Fix coordinates of the rectangle we use to erase the unwanted separator line in wxListCtrl report view in Windows 11 dark mode.

The rectangle was painted over the last item, instead below it as intended.

Closes #24024.

### DIFF
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3415,7 +3415,7 @@ void wxListCtrl::OnPaint(wxPaintEvent& event)
 
         dc.SetPen(*wxTRANSPARENT_PEN);
         dc.SetBrush(GetBackgroundColour());
-        dc.DrawRectangle(0, lastRect.y, clientSize.x, clientSize.y - lastRect.y);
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for noticing and fixing this! I have no idea why is it only a problem under Windows 11, but this is clearly the right thing to do, and I'll apply it soon.

Thanks again!

PBfordev:
>I have no idea why is it only a problem under Windows 11

The erasing happens only when Win11 is detected: https://github.com/wxWidgets/wxWidgets/commit/9710592cc4c35f60d52257e8d159e1b2448a268c

PR #24409, opened 2024-03-17
Remove Windows XP only code path from wxStaticBox code

PBfordev:
Windows XP is no longer supported, so remove the code that is never going to be executed.


vadz:
Thanks!

PR #24411, opened 2024-03-17
Allow separators in wxChoice popup menus

mojca:
I'm attempting to upstream some changes from the KiCad project (which currently requires a custom fork of wxWidgets to build) in order to simplify packaging.

(It would be great if patches could end up in the 3.2 branch as 3.4 isn't likely to be released any time soon, but for this particular one I'm not sure whether it breaks ABI compatibility or breaks the rules for backporting to 3.2. If it is not allowed to stay, just say so and I'll rebase it and retarget it to the master branch.)

@craftyjon @jey5nd6

### DIFF
--- a/include/wx/osx/core/private.h
+++ b/include/wx/osx/core/private.h
@@ -816,6 +816,8 @@ public :
 
     virtual void InsertItem(size_t pos, int itemid, const wxString& text) = 0;
 
+    virtual void InsertSeparator(size_t WXUNUSED(pos)) {}
+
     virtual void RemoveItem(size_t pos) = 0;
 
     virtual void Clear()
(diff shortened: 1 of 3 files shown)


mojca:
After further discussion it turns out that this should better be done in a more generic way. Closing for now.

vadz:
FWIW if we want to do this in a more generic way, it would have to be done at `wxItemContainer` level, but this would definitely require ABI changes (and so couldn't be applied to 3.2). Or perhaps we can do it just for `wxChoice`, but this is not really generic...

vadz:
Oh, sorry for forgetting about this discussion. Looking forward to your changes, TIA!

PR #24413, opened 2024-03-18
wxrc: Faster&leaner compilation of C++ source file generated by resource compiler.

DoctorNoobingstoneIPresume:
In the C++ source code generated by the resource compiler, for each binary file (XRC file or file referenced from an XRC file), a call to `XRC_ADD_FILE` is placed in the generated file.

Previously, `XRC_ADD_FILE` used to be a function-like macro (similar to a function, with all calls inlined), simply "forwarding" its arguments to `wxMemoryFSHandler::AddFile`.

But some parameters of `wxMemoryFSHandler::AddFile` are of type `const wxString &` while the arguments we have are of type `const wxChar *`, and using the macro used to force at every "call" site the construction and destruction of temporary `wxString` objects.

When there are thousands of such calls, the compilation of the generated C++ source code takes much time and memory:

```
  For  2,000 binary files:
    g++-11 (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0  (`g++-11 -O2 -g`):  42.95 seconds and 1,107,296 KiB of RAM
    g++-14 (GCC) 14.0.1 20240117 (experimental)   (`g++    -O2 -g`):  93.44 seconds and 1,149,840 KiB of RAM

  For 10,000 binary files:
    Visual C++ 2019 (version 19.29.30151 for x64) (`cl /O2 /Zi`):     46.51 seconds
    g++-11 (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0  (`g++-11 -O2 -g`): 151.13 seconds and 3,265,268 KiB of RAM
    g++-14 (GCC) 14.0.1 20240117 (experimental)   (`g++    -O2 -g`): 491.27 seconds and 3,489,140 KiB of RAM
```

We have now changed `XRC_ADD_FILE` into a function. It simply calls `wxMemoryFSHandler::AddFile`, therefore the behaviour is the same. But its parameters are not of type `const wxString &`, but of type `const wxChar *` instead, exactly matching the arguments in the generated C++ source code (and not requiring construction and destruction of temporary `wxString` objects; these constructions and destructions are now centralized in the body of `XRC_ADD_FILE`).

The benchmarks of compiling generated C++ source code are as follows:

```
  For  2,000 binary files:
    g++-11 (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0  (`g++-11 -O2 -g`):   2.22 seconds and   358,936 KiB of RAM
    g++-14 (GCC) 14.0.1 20240117 (experimental)   (`g++    -O2 -g`):   4.52 seconds and   343,872 KiB of RAM

  For 10,000 binary files:
    Visual C++ 2019 (version 19.29.30151 for x64) (`cl /O2 /Zi`):      3.23 seconds
    g++-11 (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0  (`g++-11 -O2 -g`):  13.59 seconds and   803,592 KiB of RAM
    g++-14 (GCC) 14.0.1 20240117 (experimental)   (`g++    -O2 -g`):  31.80 seconds and   858,884 KiB of RAM
```

A Bash script which generates an XRC file referencing many binary files (small PNG images) is now included (`GenerateBigXRCFile.sh`).

Additionally, we are now restricting the size of each binary file not to the limit of `size_t` (because anyway this used to be the `size_t` of the resource compiler, not the `size_t` of the subsequent C++ compiler, and because anyway for Windows, even in 64-bit processes, each module (EXE or DLL) cannot occupy more than 4 GiB of virtual memory (relative virtual addresses are limited to 32-bit, and the file format is not called "PE64", but "PE32+" instead).

Additionally, in the generated C++ source code, the individual bytes of binary files are now represented not as decimal integer literals, but instead as hexadecimal integer literals, 16 on a line, and with 256-byte sections delimited by comments showing the current offset and the total size and the percentage (the "progress").

(The name of the temporary files are also included as comments.)

### DIFF
--- a/utils/wxrc/wxrc.cpp
+++ b/utils/wxrc/wxrc.cpp
@@ -28,6 +28,7 @@
 #include "wx/vector.h"
 
 #include <unordered_set>
+#include <vector>
 
 class XRCWidgetData
 {
@@ -626,6 +627,10 @@ void XmlResApp::MakePackageZIP(const wxArrayString& flist)
 
 static wxString FileToCppArray(wxString filename, int num)
 {
+    // [2024-03-15]
+    //   Some of the `wxString` objects are defined at the beginning of the function
+    //   (before the actual use, and with larger scope than needed)
+    //   in order to avoid repeated allocations and deallocations.
     wxString output;
     wxString tmp;
     wxString snum;
@@ -633,34 +638,94 @@ static wxString FileToCppArray(wxString filename, int num)
     wxFileOffset offset = file.Length();
     wxASSERT_MSG( offset >= 0 , wxT("Invalid file length") );
 
-    const size_t lng = wx_truncate_cast(size_t, offset);
+    // [2024-03-15]
+    //
+    //   There are two environments (processes) to consider:
+    //     - the resource compiler;
+    //     - the process using the resource.
+    //
+    //   Previously, we used to check whether `offset` fitted into `size_t`.
+    //
+    //   But this used to check against the `size_t` of the resource compiler.
+    //   The check might have passed, but this might have been misleading
+    //   (if the resource compiler is a 64-bit process and the process using the resource is a 32-bit process).
+    //
+    //   Also, for Windows at least, the file format for modules (executables and shared libraries)
+    //   which are to be loaded in 64-bit processes
+    //   is not called PE64, but instead PE32+,
+    //   and yes, in this format virtual addresses are indeed 64-bit,
+    //   but in fact relative virtual addresses (and also file offsets) are 32-bit
+    //   (i.e. a single module cannot span for more than 4 GiB in virtual memory).
+    //
+    //   It follows that the check might have also been misleading
+    //   in the case of the resource compiler being a 64-bit Linux process
+    //   and the process using the resource being a Windows process -- even a 64-bit Windows process.
+    //
+    //   Therefore, we now check whether the size of the resource fits
+    //   not into `size_t`, but instead into `uint32_t`,
+    //   i.e. we do not support resources larger than 4 GiB.
+    //
+    const wxUint32 lng = wx_truncate_cast(wxUint32, offset);
     wxASSERT_MSG( static_cast<wxFileOffset>(lng) == offset,
                   wxT("Huge file not supported") );
 
     snum.Printf(wxT("%i"), num);
-    output.Printf(wxT("static size_t xml_res_size_") + snum + wxT(" = %lu;\n"),
-                  static_cast<unsigned long>(lng));
-    output += wxT("static unsigned char xml_res_file_") + snum + wxT("[] = {\n");
+
+    tmp.Printf(wxT("// File '%s':\n"), filename);
+    output += tmp;
+
+    tmp.Printf(wxT("const size_t xml_res_size_") + snum + wxT(" = %u;\n"), lng);
+    output += tmp;
+
+    output += wxT("const unsigned char xml_res_file_") + snum + wxT("[] = {\n");
     // we cannot use string literals because MSVC is dumb wannabe compiler
     // with arbitrary limitation to 2048 strings :(
 
-    unsigned char *buffer = new unsigned char[lng];
-    file.Read(buffer, lng);
+    std::vector<unsigned char> buffer(lng);
+    file.Read(buffer.data(), lng);
 
-    for (size_t i = 0, linelng = 0; i < lng; i++)
+    for (wxUint32 i = 0, linelng = 0, sectionlng = 0; ; )
     {
-        tmp.Printf(wxT("%i"), buffer[i]);
-        if (i != 0) output << wxT(',');
-        if (linelng > 70)
+        if (! i || i >= lng || sectionlng >= 0x100)
         {
-            linelng = 0;
-            output << wxT("\n");
-        }
-        output << tmp;
-        linelng += tmp.length()+1;
-    }
+            if (linelng) wxFAIL;
 
-    delete[] buffer;
+            tmp.Printf
+            (
+                wxT("// Offset %8Xh of %8Xh (%7.2lf%%)%c\n"),
+                static_cast<unsigned>(i),
+                static_cast<unsigned>(lng),
+                lng ? 100.0 * i / lng : 0,
+                i < lng ? wxT(':') : wxT('.')
+            );
+            output << tmp;
+
+            sectionlng = 0;
+        }
+
+        if (i >= lng)
+            break;
+
+        tmp.Printf(wxT("0x%02x"), buffer[i]);
+        output << tmp;
+
+        ++i;
+        ++linelng;
+        ++sectionlng;
+
+        const bool newline{i == lng || linelng >= 0x10};
+        if (i < lng)
+        {
+            output << wxT(',');
+            if (! newline && ! (linelng % 4))
+                output << wxT(' ');
+        }
+        if (newline)
+        {
+            output << wxT('\n');
+            linelng = 0;
+        }
+    }
 
     output += wxT("};\n\n");
 
@@ -688,18 +753,23 @@ void XmlResApp::MakePackageCPP(const wxArrayString& flist)
 "#include <wx/xrc/xmlres.h>\n"
 "#include <wx/xrc/xh_all.h>\n"
 "\n"
-"#if wxCHECK_VERSION(2,8,5) && wxABI_VERSION >= 20805\n"
-"    #define XRC_ADD_FILE(name, data, size, mime) \\\n"
-"        wxMemoryFSHandler::AddFileWithMimeType(name, data, size, mime)\n"
-"#else\n"
-"    #define XRC_ADD_FILE(name, data, size, mime) \\\n"
-"        wxMemoryFSHandler::AddFile(name, data, size)\n"
-"#endif\n"
+"// [2024-03-11]\n"
+"//   In order for C++-compilation to run much faster and consume much less memory\n"
+"//   (especially with g++-14, at least with optimization and debugging information i.e. `-O2 -g`)\n"
+"//   and for the resulting module (executable or shared library) to occupy less space (on disk and in memory),\n"
+"//   we avoid (via an intermediary function with parameters of type pointers to arrays of characters)\n"
+"//   the inlining of construction of (many) wxString parameters\n"
+"//   (especially since (anyway) our arguments (in the generated C++ source code)\n"
+"//   are pointers to arrays of characters, not yet strings):\n"
+"//\n"
+"void XRC_ADD_FILE(const wxChar *filename, const void *binarydata, size_t size, const wxChar *mimetype);\n"
 "\n");
 
+    file.Write("namespace\n{\n");
     for (i = 0; i < flist.GetCount(); i++)
         file.Write(
               FileToCppArray(parOutputPath + wxFILE_SEP_PATH + flist[i], i));
+    file.Write("} // End of anonymous namespace.\n\n");
 
     file.Write(""
 "void " + parFuncname + "()\n"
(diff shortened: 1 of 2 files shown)


utils/wxrc/wxrc.cpp
```diff
@@ -626,41 +627,105 @@ void XmlResApp::MakePackageZIP(const wxArrayString& flist)
 
 static wxString FileToCppArray(wxString filename, int num)
 {
+    // [2024-03-15]
```
vadz, line 630: Please don't add comments with dates like this, we can always use `git blame` to find out when something was added and most of the time it's not really important.

Also, ideally, this change should have been done in a separate commit with the subject "Add explanatory comment to wxrc" and clearly stating that it doesn't change anything in the commit message.

utils/wxrc/wxrc.cpp
```diff
@@ -626,41 +627,105 @@ void XmlResApp::MakePackageZIP(const wxArrayString& flist)
...
+    const wxUint32 lng = wx_truncate_cast(wxUint32, offset);
     wxASSERT_MSG( static_cast<wxFileOffset>(lng) == offset,
                   wxT("Huge file not supported") );
 
     snum.Printf(wxT("%i"), num);
-    output.Printf(wxT("static size_t xml_res_size_") + snum + wxT(" = %lu;\n"),
-                  static_cast<unsigned long>(lng));
-    output += wxT("static unsigned char xml_res_file_") + snum + wxT("[] = {\n");
+
+    tmp.Printf(wxT("// File '%s':\n"), filename);
```
vadz, line 674: This may well be useful, but really should be done in a separate commit too because it's completely unrelated to the other changes (and purely cosmetic).

utils/wxrc/wxrc.cpp
```diff
@@ -626,41 +627,105 @@ void XmlResApp::MakePackageZIP(const wxArrayString& flist)
...
+    //
+    //   It follows that the check might have also been misleading
+    //   in the case of the resource compiler being a 64-bit Linux process
+    //   and the process using the resource being a Windows process -- even a 64-bit Windows process.
+    //
+    //   Therefore, we now check whether the size of the resource fits
+    //   not into `size_t`, but instead into `uint32_t`,
+    //   i.e. we do not support resources larger than 4 GiB.
+    //
+    const wxUint32 lng = wx_truncate_cast(wxUint32, offset);
```
vadz, line 668: Thanks, I agree with the fix, but it's unrelated to the other changes and should be in a separate commit.

vadz (commented):
Thanks for your changes, but it would be really great if could you please break them in more reviewable individual commits, as it would make reviewing them much simpler.

Concerning the main topic of this PR, I'm pretty impressed that you're using wxxrc with thousands of files and I'd be very curious to know what is your use case, but we can certainly improve things by making compiler's job easier in this case as you did, it just would be nice to untangle this change from all the rest.

DoctorNoobingstoneIPresume:
Thank you so much for kind replies !!

You are of course correct, I should certainly change this pull request into smaller, more focused pull requests, which can be reviewed/understood/approved/discarded individually.

I would like to ask for a few more days for me to do this. I am also going to read the comments you have kindly added as review to the current pull request. Thank you for those as well.

My day is more beautiful now !! :)

utils/wxrc/wxrc.cpp
```diff
@@ -626,41 +627,105 @@ void XmlResApp::MakePackageZIP(const wxArrayString& flist)
...
 
     snum.Printf(wxT("%i"), num);
-    output.Printf(wxT("static size_t xml_res_size_") + snum + wxT(" = %lu;\n"),
-                  static_cast<unsigned long>(lng));
-    output += wxT("static unsigned char xml_res_file_") + snum + wxT("[] = {\n");
+
+    tmp.Printf(wxT("// File '%s':\n"), filename);
+    output += tmp;
+
+    tmp.Printf(wxT("const size_t xml_res_size_") + snum + wxT(" = %u;\n"), lng);
```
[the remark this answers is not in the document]
DoctorNoobingstoneIPresume: Thank you.

Indeed, `static` has been removed, but I think these arrays are now placed in an anonymous namespace.
As per https://github.com/DoctorNoobingstoneIPresume/240325_static_vs_anonymous_namespace, they now still have internal linkage.

But if there is a preference for expressing internal linkage via `static`, we can switch back to it. Kindly let me know, please.

Working on separate commits... May I ask whether all new commits should be independent or at least some of them can be grouped into pull request(s), i.e. can depend on each other (in a sequence), please ?

utils/wxrc/wxrc.cpp
```diff
@@ -626,41 +627,105 @@ void XmlResApp::MakePackageZIP(const wxArrayString& flist)
...
+    for (wxUint32 i = 0, linelng = 0, sectionlng = 0; ; )
     {
-        tmp.Printf(wxT("%i"), buffer[i]);
-        if (i != 0) output << wxT(',');
-        if (linelng > 70)
+        if (! i || i >= lng || sectionlng >= 0x100)
         {
-            linelng = 0;
-            output << wxT("\n");
+            if (linelng) wxFAIL;
```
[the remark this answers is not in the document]
DoctorNoobingstoneIPresume: I would like to make an assertion:

```
some_azzert_macro(! linelng);
```

(Perhaps that way of expressing in the source code might have been more clear...)

Strangely enough, the intent has been to clarify for readers of the code that, despite the maybe complex logic, at this point `linelng` should be zero, otherwise the programmer has made an error.

I wanted the azzertion to always be enabled, even in `_NDEBUG` builds (i.e. in "release" builds).

If alright with you, may I ask whether a macro already exists for this, please ?

utils/wxrc/wxrc.cpp
DoctorNoobingstoneIPresume: Thank you.

We now have https://github.com/wxWidgets/wxWidgets/pull/24437 as partial replacement for this pull request (not addressing the main topic) -- for review and suggestions for modifications.

I am going to get back a little later these days with that and also with the rest (the main topic of this pull request is not addressed yet by https://github.com/wxWidgets/wxWidgets/pull/24437).

utils/wxrc/wxrc.cpp
DoctorNoobingstoneIPresume: Thank you... Please see https://github.com/wxWidgets/wxWidgets/pull/24437/commits/eba82f0e94f42071e426733ab91a5a5cb5240b24 when there is time...

utils/wxrc/wxrc.cpp
DoctorNoobingstoneIPresume: > Thanks, I agree with the fix, but it's unrelated to the other changes and should be in a separate commit.

https://github.com/wxWidgets/wxWidgets/pull/24437/commits/6952680b6e7dfbef5957372df75c118349d811be

utils/wxrc/wxrc.cpp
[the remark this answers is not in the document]
DoctorNoobingstoneIPresume: https://github.com/wxWidgets/wxWidgets/pull/24437/commits/c49e8a595f6f7fcde0fa8090ae6a9ffcc19437b3

utils/wxrc/wxrc.cpp
```diff
@@ -688,18 +753,23 @@ void XmlResApp::MakePackageCPP(const wxArrayString& flist)
...
 "#include <wx/xrc/xh_all.h>\n"
 "\n"
-"#if wxCHECK_VERSION(2,8,5) && wxABI_VERSION >= 20805\n"
-"    #define XRC_ADD_FILE(name, data, size, mime) \\\n"
-"        wxMemoryFSHandler::AddFileWithMimeType(name, data, size, mime)\n"
-"#else\n"
-"    #define XRC_ADD_FILE(name, data, size, mime) \\\n"
-"        wxMemoryFSHandler::AddFile(name, data, size)\n"
-"#endif\n"
+"// [2024-03-11]\n"
```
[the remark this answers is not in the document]
DoctorNoobingstoneIPresume: Ok, thank you. Please kindly see https://github.com/wxWidgets/wxWidgets/pull/24437/commits/e82247383ccaebf8782cc54add2df9d9ac6dba23 (or whatever replacement commit is added later in #24437) when there is time...

utils/wxrc/wxrc.cpp
```diff
@@ -626,41 +627,105 @@ void XmlResApp::MakePackageZIP(const wxArrayString& flist)
...
+    std::vector<unsigned char> buffer(lng);
+    file.Read(buffer.data(), lng);
 
-    for (size_t i = 0, linelng = 0; i < lng; i++)
+    for (wxUint32 i = 0, linelng = 0, sectionlng = 0; ; )
     {
-        tmp.Printf(wxT("%i"), buffer[i]);
-        if (i != 0) output << wxT(',');
-        if (linelng > 70)
+        if (! i || i >= lng || sectionlng >= 0x100)
```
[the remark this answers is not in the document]
DoctorNoobingstoneIPresume: This I have not done yet (even in https://github.com/wxWidgets/wxWidgets/pull/24437). 😢

utils/wxrc/wxrc.cpp
```diff
@@ -626,41 +627,105 @@ void XmlResApp::MakePackageZIP(const wxArrayString& flist)
...
         {
-            linelng = 0;
-            output << wxT("\n");
+            if (linelng) wxFAIL;
+
+            tmp.Printf
+            (
+                wxT("// Offset %8Xh of %8Xh (%7.2lf%%)%c\n"),
```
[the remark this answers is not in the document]
DoctorNoobingstoneIPresume: I think it might help with a quick vision of progress, telling us where we are within the file, for example when we scroll up and down (especially at fast speed) within the generated C++ code.

For me personally, I remember (and like) hex dumps in a format which also includes the offset:

```
$ hexdump -Cv "$(which ls)" | head -n 16
00000000  7f 45 4c 46 02 01 01 00  00 00 00 00 00 00 00 00  |.ELF............|
00000010  03 00 3e 00 01 00 00 00  60 61 00 00 00 00 00 00  |..>.....`a......|
00000020  40 00 00 00 00 00 00 00  68 37 02 00 00 00 00 00  |@.......h7......|
00000030  00 00 00 00 40 00 38 00  0b 00 40 00 1e 00 1d 00  |....@.8...@.....|
00000040  06 00 00 00 04 00 00 00  40 00 00 00 00 00 00 00  |........@.......|
00000050  40 00 00 00 00 00 00 00  40 00 00 00 00 00 00 00  |@.......@.......|
00000060  68 02 00 00 00 00 00 00  68 02 00 00 00 00 00 00  |h.......h.......|
00000070  08 00 00 00 00 00 00 00  03 00 00 00 04 00 00 00  |................|
00000080  a8 02 00 00 00 00 00 00  a8 02 00 00 00 00 00 00  |................|
00000090  a8 02 00 00 00 00 00 00  1c 00 00 00 00 00 00 00  |................|
000000a0  1c 00 00 00 00 00 00 00  01 00 00 00 00 00 00 00  |................|
000000b0  01 00 00 00 04 00 00 00  00 00 00 00 00 00 00 00  |................|
000000c0  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
000000d0  38 35 00 00 00 00 00 00  38 35 00 00 00 00 00 00  |85......85......|
000000e0  00 10 00 00 00 00 00 00  01 00 00 00 05 00 00 00  |................|
000000f0  00 40 00 00 00 00 00 00  00 40 00 00 00 00 00 00  |.@.......@......|
```

Including the offset on each line may be a little too much output (of C++ code), so I thought we just include it every 16 lines.

However, I understand that this is very personal and I can only ask that you consider it for a little time. If you prefer otherwise, an empty line or nothing at all, I am going to change the code accordingly.

(A separate commit has been created for this: https://github.com/wxWidgets/wxWidgets/pull/24437/commits/ec7d065c4e618db5c5718c864a232c4994f81072 -- as part of #24437. The exact commit ID's might change as that pull request is changed.)

utils/wxrc/wxrc.cpp
DoctorNoobingstoneIPresume: #24437 (with smaller commits) now fully replaces this pull request...

vadz:
It looks like this was superseded by #24437, so closing.

PR #24417, opened 2024-03-19
WebView sample window improvements

TcT2k:
- Enlarge default size to make menu button more easily accessible, without resize
- Move log window to the buttom of the window initially

### DIFF
--- a/samples/webview/webview.cpp
+++ b/samples/webview/webview.cpp
@@ -454,8 +454,9 @@ WebFrame::WebFrame(const wxString& url, int flags, wxWebViewWindowFeatures* wind
     topsizer->Add(m_info, wxSizerFlags().Expand());
 
     // Create a log window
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, I'll apply this (with some very minor tweaks) but, FWIW, this doesn't work under GTK/Wayland where the log window can't be positioned by the application. If we want to make this really convenient, I think we need to embed the log window into the main frame -- which could also then be used to test that you can have `wxWebView` as a sub-window and not as the only frame contents.

TcT2k:
I've added an additional commit to move the log to a text control on the bottom. It can be shown/hidden by a menu item. I'm not sure if it should be visible by default.

vadz:
Thanks, this works perfectly well under wxGTK too!

And I do think the log should be shown by default, it would be painful to have to show it every time when testing something.

I'll (squash) merge this soon.

PR #24419, opened 2024-03-20
Fixes for wxUSE_OLE=0 and also fix --disable-graphics_d2d on configure

b-fission:
These patches resolve some issues when wxWidgets is built with OLE disabled:
- linker error due to undefined `MSWUsesStandardAutoComplete()`
- runtime error where `wxOleInitModule` fails during app startup

And when using the configure script, the `--disable-graphics_d2d` seemed to have no effect, so Direct2D support was enabled solely on a compile test. The configure script will now actually check if the option is not disabled before testing Direct2D support.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -7243,24 +7243,26 @@ if test "$wxUSE_GRAPHICS_CONTEXT" = "yes"; then
             wx_has_graphics=1
         fi
 
-        AC_CACHE_CHECK([if Direct2D is available], wx_cv_lib_direct2d,
-            [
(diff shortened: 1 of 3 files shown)


vadz:
Thanks, I'll merge this soon but I'd just like to say that IMO building wxMSW without OLE support doesn't make any sense nowadays (and by this I mean the last 25 years or so), it was only really needed for Win16 where using OLE added non-trivial overhead, but since then it is used by everything else anyhow and disabling it results in a loss of a lot of functionality. So I think just removing `wxUSE_OLE` could be a better idea.

PR #24421, opened 2024-03-20
Add support for clearing browsing data in wxWebView

TcT2k:
Add new method `wxWebView::ClearBrowsingData()` for WKWebView, Edge and WebKitGTK

Clears the browsing data of the web view.
This function clears the browsing data of the web view, such as cookies, cache, history, etc. The exact data that is cleared depends on the types parameter and the backend used.

This operation is asynchronous and may take some time to complete. 
When finished a `wxEVT_WEBVIEW_BROWSING_DATA_CLEARED` event is generated.

@vadz Could you please have a look at my usage of `wxDateTime`? From what I understand it would be initialized to min(int) but some of the backends treat 0 as a special value that's why I initialized it to 0. Not sure if there would be a better way.


src/gtk/webview_webkit2.cpp
```diff
@@ -1654,6 +1662,63 @@ void wxWebViewWebKit::RemoveAllUserScripts()
...
+
+bool wxWebViewWebKit::ClearBrowsingData(int types, wxDateTime since)
+{
+#ifdef wxHAVE_WEBKIT_WEBSITE_DATA_MANAGER_CLEAR
+    if (wx_check_webkit_version(2, 16, 0))
+    {
+        WebKitWebsiteDataManager* manager = static_cast<wxWebViewConfigurationImplWebKit*>(m_config.GetImpl())->GetWebsiteDataManager();
+
+        if (types & wxWEBVIEW_BROWSING_DATA_OTHER)
+            return false;
```
vadz, line 1694: Why not map it to all the elements of `WebKitWebsiteDataTypes` not otherwise used below?

src/osx/webview_webkit.mm
```diff
@@ -503,6 +506,50 @@ function createClass(name,rules) { \
...
+        {
+            clearDataTypes = [WKWebsiteDataStore allWebsiteDataTypes];
+        }
+        else
+        {
+            NSMutableSet<NSString*>* dataTypes = [[NSMutableSet alloc] init];
+            if (types & wxWEBVIEW_BROWSING_DATA_COOKIES)
+                [dataTypes addObject:WKWebsiteDataTypeCookies];
+            if (types & wxWEBVIEW_BROWSING_DATA_CACHE)
+                [dataTypes addObjectsFromArray:@[WKWebsiteDataTypeDiskCache, WKWebsiteDataTypeMemoryCache,
```
vadz, line 528: There is also `WKWebsiteDataTypeFetchCache`, is it excluded on purpose?

src/osx/webview_webkit.mm
```diff
@@ -503,6 +506,50 @@ function createClass(name,rules) { \
...
 }
 
+bool wxWebViewWebKit::ClearBrowsingData(int types, wxDateTime since)
+{
+    if ( WX_IS_MACOS_AVAILABLE(10, 11) )
+    {
+        NSSet<NSString*>* clearDataTypes = nil;
+        if (types & wxWEBVIEW_BROWSING_DATA_OTHER)
+        {
+            return false;
```
vadz, line 516: Again, it looks like we could map this to `WKWebsiteDataTypeServiceWorkerRegistrations`?

include/wx/webview.h
```diff
@@ -94,6 +94,15 @@ enum wxWebViewUserScriptInjectionTime
     wxWEBVIEW_INJECT_AT_DOCUMENT_END
 };
 
+enum wxWebViewBrowsingDataTypes
+{
+    wxWEBVIEW_BROWSING_DATA_COOKIES = 1,
+    wxWEBVIEW_BROWSING_DATA_CACHE = 2,
+    wxWEBVIEW_BROWSING_DATA_DOM_STORAGE = 4,
+    wxWEBVIEW_BROWSING_DATA_OTHER = 8,
+    wxWEBVIEW_BROWSING_DATA_ALL = 16
```
vadz, line 103: I think this should actually be 15, i.e. the sum of everything below, as otherwise it's not really clear what is the difference between specifying all of the previous constants together or this one -- by making it equal to their sum, this would clearly show that this is equivalent.

vadz (commented):
Thanks a lot for adding this and sorry for such a long delay with the review!

But, to answer your question, I think that using `wxDateTime(0)` is indeed a bit ugly and I believe it would be better to use `wxInvalidDateTime` and then do something like

```cpp
date.IsValid() ? date.GetTicks() : 0
```

in the implementation code. This would be slightly more verbose but IMHO much more clear at the interface level.

vadz:
@TcT2k Do you plan to return to this or should I try to change it as proposed and merge it?

vadz:
I've created #25016 with the proposed changes, so closing this one.

PR #24423, opened 2024-03-20
Clipboard fixes for wxGTK

vadz:
This addresses #24391 but there are still some problems remaining:

1. I have no idea how to put both `UTF8_STRING` and `STRING` on clipboard when copying. I can also get one _or_ the other (with the corresponding Wayland types, unconditionally added by GTK), but not both. It doesn't seem to do any harm to provide just `UTF8_STRING` a.k.a. `text/plain;charset=utf-8` but other applications do provide both those and `STRING` and `text/plain`, so we probably should too...
2. This does nothing to address this [GTK bug](https://gitlab.gnome.org/GNOME/gtk/-/issues/2307) even though it almost surely affects us as well (but I ran out of time for testing).
3. The code is still ugly and unclear, in spite of my feeble attempts to simplify it.

@paulcor Your review would be really welcome if you could please have a look!


src/gtk/clipbrd.cpp
```diff
@@ -663,10 +674,38 @@ bool wxClipboard::AddData( wxDataObject *data )
     // for explanation
     AddSupportedTarget(g_timestampAtom);
 
+#ifdef __WXGTK3__
+    static const bool isWayland =
```
paulcor, line 678: The `IsWayland()` result is already cached, so there is no need to make this static.

PR #24424, opened 2024-03-20
Further improve wxListCtrl appearance in Windows dark mode

### LINKED ISSUE #24410: No checkboxes in wxListCtrl under MSW in dark mode
### Bug description
Checkboxes are not displayed in `wxListCtrl` report view when running on Windows 10 or 11 dark mode.

I did not look into it, just ran into it when fixing another wxListCtrl dark mode issue.

### To Reproduce
1. Launch the listctrl sample in dark mode.
2. From menu _List_ choose _Enable checkboxes_.
3. Observe that checkboxes are not displayed as expected, only the item selection rect is offset to right, accounting for them.

- wxWidgets version you use: master
- OS: Windows 10 v22H2, Windows 11 v23H2

PBfordev:
## Fix missing checkboxes in wxListCtrl in Windows dark mode

Under Windows dark mode, we paint wxListCtrl in report view by ourselves and so we need to paint also the checkboxes when needed.

There is a small glitch on Windows 11, where a selected checkbox has black instead of transparent pixels inside the rounded corners.

Closes #24410.

----

I tried also @MaartenBent code from #24410 using `wxRendererNative::Get().DrawCheckBox()`: The glitch is smaller (only in the upper corners) but the checkmark is a bit jagged, so I kept my code. I have no idea what to do about this, perhaps I am missing something here. @MaartenBent could you please tell me what you think?
 

## Always paint item selection in wxListCtrl in Windows dark mode

When painting items in wxListCtrl in report view under Windows dark mode,
the selection was painted only when the control had focus.

However, this did not match the light mode, where the selection is always drawn
regardless of the focus (the listview is created with LVS_SHOWSELALWAYS style).

---
@vadz, this goes directly against what you coded, I think this commit makes sense but please review.

## Fix item selection painting in wxListCtrl in Windows dark mode

The item selection/focus rectangle should include not only the item text but also the checkbox and image.

### DIFF
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3435,7 +3435,10 @@ void wxListCtrl::OnPaint(wxPaintEvent& event)
         return;
     }
 
-    wxPen pen(wxSystemSettings::GetColour(wxSYS_COLOUR_3DLIGHT));
+    const wxColour penColour(wxSystemSettings::GetColour(wxMSWDarkMode::IsActive()
+                                                         ? wxSYS_COLOUR_GRAYTEXT
+                                                         : wxSYS_COLOUR_3DLIGHT));
+    wxPen pen(penColour);
     dc.SetPen(pen);
     dc.SetBrush(* wxTRANSPARENT_BRUSH);
 
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3226,23 +3226,8 @@ void HandleItemPaint(wxListCtrl* listctrl, LPNMLVCUSTOMDRAW pLVCD, HFONT hfont)
 
     if ( nmcd.uItemState & CDIS_SELECTED )
     {
-        wxSystemColour syscolFg, syscolBg;
-        if ( ::GetFocus() == hwndList )
-        {
-            syscolFg = wxSYS_COLOUR_HIGHLIGHTTEXT;
-            syscolBg = wxSYS_COLOUR_HIGHLIGHT;
-        }
-        else // selected but unfocused
-        {
-            syscolFg = wxSYS_COLOUR_WINDOWTEXT;
-            syscolBg = wxSYS_COLOUR_BTNFACE;
-
-            // don't grey out the icon in this case either
-            nmcd.uItemState &= ~CDIS_SELECTED;
-        }
-
-        pLVCD->clrText = wxColourToRGB(wxSystemSettings::GetColour(syscolFg));
-        pLVCD->clrTextBk = wxColourToRGB(wxSystemSettings::GetColour(syscolBg));
+        pLVCD->clrText = wxColourToRGB(wxSystemSettings::GetColour(wxSYS_COLOUR_HIGHLIGHTTEXT));
+        pLVCD->clrTextBk = wxColourToRGB(wxSystemSettings::GetColour(wxSYS_COLOUR_HIGHLIGHT));
     }
     //else: not selected, use normal colours from pLVCD
 
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3265,6 +3265,14 @@ WXLPARAM HandleItemPrepaint(wxListCtrl *listctrl,
         // We need to always paint selected items ourselves as they come
         // out completely wrong in DarkMode_Explorer theme, see the comment
         // before MSWGetDarkModeSupport().
+        wxFont font;
+
+        if ( attr && attr->HasFont() )
+        {
+            font = attr->GetFont();
+            font.WXAdjustToPPI(listctrl->GetDPI());
+        }
+
         pLVCD->clrText = attr && attr->HasTextColour()
                             ? wxColourToRGB(attr->GetTextColour())
                             : wxColourToRGB(listctrl->GetTextColour());
@@ -3272,7 +3280,7 @@ WXLPARAM HandleItemPrepaint(wxListCtrl *listctrl,
                             ? wxColourToRGB(attr->GetBackgroundColour())
                             : wxColourToRGB(listctrl->GetBackgroundColour());
 
-        HandleItemPaint(listctrl, pLVCD, nullptr);
+        HandleItemPaint(listctrl, pLVCD, font.IsOk() ? GetHfontOf(font) : nullptr);
         return CDRF_SKIPDEFAULT;
     }
 
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3032,17 +3032,18 @@ bool wxListCtrl::MSWOnNotify(int idCtrl, WXLPARAM lParam, WXLPARAM *result)
 namespace
 {
 
-RECT GetCustomDrawnItemRect(const NMCUSTOMDRAW& nmcd)
+RECT GetCustomDrawnItemRect(const NMCUSTOMDRAW& nmcd, bool excludeIconPart = false)
 {
     RECT rc;
     wxGetListCtrlItemRect(nmcd.hdr.hwndFrom, nmcd.dwItemSpec, LVIR_BOUNDS, rc);
 
-    RECT rcIcon;
-    wxGetListCtrlItemRect(nmcd.hdr.hwndFrom, nmcd.dwItemSpec, LVIR_ICON, rcIcon);
-
-    // exclude the icon part, neither the selection background nor focus rect
-    // should cover it
-    rc.left = rcIcon.right;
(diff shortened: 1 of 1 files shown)


MaartenBent:
Looks good, `ImageList_Draw` is indeed a better solution than using `wxRendererNative`.
I don't have Windows 11 so I cannot see the glitches. Maybe some combination of the `ILD_*` / `IMAGELISTDRAWFLAGS` can fix this?

I have another improvement suggestion. In light-mode the images are vertically centered, in dark-mode not. It can use the same y-coord algorithm as I used for the checkbox:
```diff
 src/msw/listctrl.cpp | 9 +++++----
 1 file changed, 5 insertions(+), 4 deletions(-)

diff --git a/src/msw/listctrl.cpp b/src/msw/listctrl.cpp
index 2e894b3027..451e36be62 100644
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3111,9 +3111,13 @@ bool HandleSubItemPrepaint(wxListCtrl* listctrl, LPNMLVCUSTOMDRAW pLVCD, HFONT h
     HIMAGELIST himl = ListView_GetImageList(hwndList, LVSIL_SMALL);
     if ( himl && ImageList_GetImageCount(himl) )
     {
+        int wImage, hImage;
+        ImageList_GetIconSize(himl, &wImage, &hImage);
+
         if ( it.iImage != -1 )
         {
-            ImageList_Draw(himl, it.iImage, hdc, rc.left, rc.top,
+            int imgY = rc.top + ((rc.bottom - rc.top) / 2 - hImage / 2);
+            ImageList_Draw(himl, it.iImage, hdc, rc.left, imgY,
                            nmcd.uItemState & CDIS_SELECTED ? ILD_SELECTED
                                                            : ILD_TRANSPARENT);
         }
@@ -3124,9 +3128,6 @@ bool HandleSubItemPrepaint(wxListCtrl* listctrl, LPNMLVCUSTOMDRAW pLVCD, HFONT h
         // images align?)
         if ( it.iImage != -1 || it.iSubItem == 0 )
         {
-            int wImage, hImage;
-            ImageList_GetIconSize(himl, &wImage, &hImage);
-
             rc.left += wImage + 2;
         }
     }

```

vadz:
Thanks! I didn't have time to test this yet, but I've pushed Maarten's suggestion and my own fix (which normally shouldn't change the behaviour) and if this looks good to both of you, I'll merge it soon.

src/msw/listctrl.cpp
```diff
@@ -3092,9 +3105,13 @@ bool HandleSubItemPrepaint(LPNMLVCUSTOMDRAW pLVCD, HFONT hfont, int colCount)
     HIMAGELIST himl = ListView_GetImageList(hwndList, LVSIL_SMALL);
     if ( himl && ImageList_GetImageCount(himl) )
     {
+        int wImage, hImage;
+        ImageList_GetIconSize(himl, &wImage, &hImage);
+
         if ( it.iImage != -1 )
         {
-            ImageList_Draw(himl, it.iImage, hdc, rc.left, rc.top,
+            int imgY = rc.top + ((rc.bottom - rc.top) / 2 - hImage / 2);
```
PBfordev, line 3113: Supernitpick: I would rename `imgY` to `yImage`, to be consistent with `wImage` and `hImage`.

PBfordev:
> I don't have Windows 11 so I cannot see the glitches. Maybe some combination of the `ILD_*` / `IMAGELISTDRAWFLAGS` can fix this?

I did not think the flags can change it, from the docs they would not help and I also still I tried combining many of them yesterday, even using `ImageList_DrawEx()`, with no effect. I have investigated the issue in-depth today. 

Firstly, I tested with a Win32 application using listview dark mode, mentioned [here](https://github.com/wxWidgets/wxWidgets/issues/24410#issuecomment-2004286142): It shows the issue. 

Secondly, I used File Explorer, which did NOT manifest the issue.

Then I saved the image and mask bitmaps using` ImageList_GetImageInfo()`: The mask had transparent pixel inside the rounded corners.

I was still playing with the listctrl sample using different DPI scaling (125% being the default) and set it manually to 200% (selection went only to 175%), which required me to reboot. At 200%, there was no glitch. Then I tried 100%: No glitch. OK, understandable. But then I tried 150% and the original 125%: No glitch either. Since using the manual override and rebooting, the glitch is gone, even after reboots. Siiiigh.

I now tested also with a Win11 VM under Win10 VirtualBox (which I don't like for testing visual glitches): No glitch there either. I don't have another Win11 machine on hand to test with, maybe on weekend, if I will remember it.

`</rant>`

PBfordev:
>  I've pushed Maarten's suggestion and my own fix (which normally shouldn't change the behaviour) 

Thanks, I was going to fix the whitespace and add Maarten's code. BTW, how did you make the commit to look as his, when you had only the diff?

I considered using a constant instead of literal for the gap too but I left it be. TBH, I would prefer `GAP_BETWEEN_*IMAGE*_AND_TEXT` instead, since it is not related to just to checkboxes (and a checkbox is technically a listview image anyway). I assume the change from physical to logical pixels was intentional...

Anyway, I'm for merging, as it fixes an IMO non-trivial bug (no checkboxes visible). Unfortunately, I have uncovered many additional darkmode `wxListCtrl` glitches, which I am not sure I can fix but I will still report them, sorry in advance.

vadz:
> Thanks, I was going to fix the whitespace and add Maarten's code. BTW, how did you make the commit to look as his, when you had only the diff?

I applied the patch manually (well, using `git apply`) and then did

```console
$ GIT_AUTHOR_NAME='Maarten Bent' GIT_AUTHOR_EMAIL='MaartenBent@users.noreply.github.com' git commit
```

> I considered using a constant instead of literal for the gap too but I left it be. TBH, I would prefer `GAP_BETWEEN_*IMAGE*_AND_TEXT` instead, since it is not related to just to checkboxes (and a checkbox is technically a listview image anyway). I assume the change from physical to logical pixels was intentional...

It was used with `FromDIP()` in `HandleSubItemPrepaint()`, so I've added `FromDIP()` to the other place too, following the comment saying that it's the same constant, but I'm actually not sure if it's correct. Please let me know if `FromDIP()` should be removed (from both places).

Thanks!

PBfordev:
> > I assume the change from physical to logical pixels was intentional...

> It was used with `FromDIP()` in `HandleSubItemPrepaint()`, so I've added `FromDIP()` to the other place too, following the comment saying that it's the same constant, but I'm actually not sure if it's correct. 

When drawing the checkboxes, I used `FromDIP(2)` between the checkbox and the rest of item (i.e., image and text), copying Maarten's code.

 The original code used 2 physical pixels for the gap between the image and text:
https://github.com/wxWidgets/wxWidgets/blob/e7e833fa316a0cb5dc7a7b35c4fc9ee3455ad464/src/msw/listctrl.cpp#L3106-L3112

I would generally always prefer DIPs, but mimicking native controls seems like alchemy, so I was not sure what was the intention there...

vadz:
Could you please check if the native control uses 2px or 2dip gap here, i.e. is it 2 or 4px in 200% DPI? I suspect it's 2, as Windows doesn't seem to scale any "small" elements at all, but I'm not sure.

TIA!

PBfordev:
> Could you please check if the native control uses 2px or 2dip gap here, i.e. is it 2 or 4px in 200% DPI? 

The gap does not seem to be scaled:
wx-listctrl-200

Sorry for not noticing the offset bug, it cannot be seen in the listctrl sample where the subitem text is not left-aligned.

PBfordev:
Oops, the line color commit is really, really the last one, I promise!

The top is before and the bottom after using a new color for lines:
wx-listctrl-lines-fixed

vadz:
> I am really done with the PR
> 
> The only thing left is to decide where to use and not use DIPs. AFAICT from the screenshots above, in light mode the gap before the checbox is scaled, the gap after is not.

Sorry for the delay, I'm without my Windows PC right now (hardware problem), so I can't test this, but it would be nice to be consistent with the light mode, however ugly it makes the code. If you could please push the really last for this, I'd be glad to merge it blindly, otherwise it will have to wait until I can do the tests myself.

vadz:
Sorry for the delay, I'm finally merging this, thanks a lot for all your work on it!

PBfordev:
Sorry for not being able to add the code mimicking the light mode appearance.

I did not have time, tried a bit but it seems more complicated then I thought. Simply moving moving the text offset messed with the coordinates of rectangle drawn when an item had a custom background colour. I may have missed something and will try again when I have time...

PR #24426, opened 2024-03-21
Fix uniitialized wxGLCanvasEGL Wayland members

craftyjon:
These were left out of dbaa0244, causing crashes


vadz:
Oops, I've no idea how did I miss those but thanks for fixing this!

PR #24435, opened 2024-03-22
Fixes for scaling non-square bitmaps in wxBitmapBundle

vadz:


ahoischen:
Thank you for the rapid turnaround on this! However, this branch introduces scaling issues, where icons are suddenly scaled too large. On the left is the the result of using the vanilla `wxBitmapBundle` implementation, while on the right is using the workaround code from the first issue. The screenshot was taken at 150% scaling. I'm guessing the scaling might be getting applied twice here, I don't have my 300% ready to check though.

image

vadz:
Sorry! I made a stupid mistake and used the wrong function and didn't notice it because this still worked under Linux, where I tested. It should work with `GetDIPSize()` instead of `GetLogicalSize()` (the latest version of this PR after force push).

ahoischen:
That fixed it, thank you!

PR #24437, opened 2024-03-25
240325_wxrc_FileToCppArray_01h

DoctorNoobingstoneIPresume:
wxrc: `FileToCppArray` and `MakePackageCPP`:

- We no longer claim to support resources larger than 4 GiB. => [2024-03-27] Commit message has now been changed to "Resource size is now limited to 4 GiB on all platforms." to reduce negativity, oh no, I mean to enhance positivity. (-:
- We have replaced `new []` and `delete []` with `std::vector <...>`.
- Temp file info (index, size, name) is now printed before defining each byte-array.
- `static` has been replaced by anonymous namespace. `const` has been added.
- Byte-array initialization integer literals are now hex; progress is shown.

This pull request replaces part of https://github.com/wxWidgets/wxWidgets/pull/24413 (as a sequence of more focused commits).

Thank you for considering this and for kind comments !! 🤗


DoctorNoobingstoneIPresume:
I apologize. I am not sure why 8 commits are included in my pull request. Only 5 of them have been intended to be included... :'(

DoctorNoobingstoneIPresume:
The following commit has now been added, beyond the scope of #24413:
- wxrc: `GetInternalFileName`: Call to `Printf`: Format string arg is now constant.

The following commit has now also been added, completing (in my opinion) the replacement of #24413:
- wxrc: `MakePackageCPP`: Faster&leaner compilation of C++ source code generated by our resource compiler.

utils/wxrc/GenerateBigXRCFile.sh
```diff
@@ -0,0 +1,43 @@
...
+
+    local i
+    local n="${1:-10000}"
+
+    printf '<?xml version="1.0" encoding="UTF-8"?>\n'
+    printf '<resource xmlns="http://www.wxwidgets.org/wxxrc" version="2.5.3.0">\n'
+    for ((i = 0; i < n; ++i)); do
+        local pathname_image; printf -v pathname_image "%s/Example_%04Xh.png" "${folder_images}" "$((i))"
+
+        if ((1)); then
```
vadz, line 27: Sorry, what's the purpose of this line?

utils/wxrc/GenerateBigXRCFile.sh
```diff
@@ -0,0 +1,43 @@
...
+            rm -f "${pathname_image}"
+            printf >>"${pathname_image}" "\x89\x50\x4E\x47\x0D\x0A\x1A\x0A\x00\x00\x00\x0D\x49\x48\x44\x52"
+            printf >>"${pathname_image}" "\x00\x00\x00\x10\x00\x00\x00\x10\x08\x02\x00\x00\x00\x90\x91\x68"
+            printf >>"${pathname_image}" "\x36\x00\x00\x00\x09\x70\x48\x59\x73\x00\x00\x0E\xC4\x00\x00\x0E"
+            printf >>"${pathname_image}" "\xC4\x01\x95\x2B\x0E\x1B\x00\x00\x00\x1A\x49\x44\x41\x54\x28\xCF"
+            printf >>"${pathname_image}" "\x63\x6C\x60\xF8\xCF\x40\x0A\x60\x62\x20\x11\x8C\x6A\x18\xD5\x30"
+            printf >>"${pathname_image}" "\x74\x34\x00\x00\xC5\xBF\x01\x9F\x22\x91\xFF\xBD\x00\x00\x00\x00"
+            printf >>"${pathname_image}" "\x49\x45\x4E\x44\xAE\x42\x60\x82"
```
vadz, line 35: Do we really have to compose the file piece-wise like this? Why can't we just add it to the repository?

vadz (commented):
Thanks for the update!

I think I'd rather move the (useful, thank you) explanations of the changes to the commit messages and leave just the brief summary in the comments in the code, but I can do it myself when merging this.

The biggest remaining question is whether we want the extra complexity of dealing with the sections in the output. IMO this is not really useful and I don't understand why would anybody want to look at these files, they're really meant to be only used as an input to the compiler and never read by a human. So I'd like to exclude the commits changing this when merging.

But I'd like to know what do the others think about it, would anybody else find it useful to have the section headers in the generated output?

DoctorNoobingstoneIPresume:
> Thanks for the update!

Thank you for everything !!

> I think I'd rather move the (useful, thank you) explanations of the changes to the commit messages and leave just the brief summary in the comments in the code, but I can do it myself when merging this.

Ok, I was thinking that we do not normally have commit messages which are this long, and also to (maybe by force) get the message across to more people 😈 , but either way is fine with me, and I can modify the comment/commit messages and also you please feel free to do it as you think is best.

> The biggest remaining question is whether we want the extra complexity of dealing with the sections in the output. IMO this is not really useful and I don't understand why would anybody want to look at these files, they're really meant to be only used as an input to the compiler and never read by a human. So I'd like to exclude the commits changing this when merging.
> 
> But I'd like to know what do the others think about it, would anybody else find it useful to have the section headers in the generated output?

My opinion has been in this reply to the previous pull request: https://github.com/wxWidgets/wxWidgets/pull/24413#discussion_r1538999530.

For review of the output, here is the XRC output of running `'./Example_Generator.sh' 100` (i.e. XRC file referencing 100 files):
- https://sharetext.me/go6bnohaon

and here are the .h file and .cpp file generated by processing that XRC file with wxrc:
- https://sharetext.me/gisgnhx5ge
- https://sharetext.me/zheze7mwdm

In the latter file (the .cpp file), we have the C++ array representation of 100 short PNG files and a longer file (the XRC file itself). When scrolling through the representation of the longer file, I think the section headers (every 16 lines, i.e. every 256 bytes) might help.

Also, I think that the C++ array representations of the binary files might be useful for debugging. For example, the author might think she has given an UTF-8 encoded file, and might wonder why she does not see the expected glyphs, and in order to debug she might run `hexdump -Cv input_file.txt` and compare that with the hex representation in the wxrc-generated C++ code (at the exact offsets where non-ASCII code points are expected) => helping her with offsets might be useful.

However, I am willing to remove these section headers if you and/or the community thinks otherwise, or you can also remove them as you see fit if you will.

utils/wxrc/GenerateBigXRCFile.sh
DoctorNoobingstoneIPresume: I was thinking we have a completely independent single-file script, but we can of course move the actual content to an external file if you wish.

utils/wxrc/GenerateBigXRCFile.sh
DoctorNoobingstoneIPresume: Thank you. I have removed it (and the corresponding `fi` of the `if`).

DoctorNoobingstoneIPresume:
> > I think I'd rather move the (useful, thank you) explanations of the changes to the commit messages and leave just the brief summary in the comments in the code, but I can do it myself when merging this.
> 
> Ok, I was thinking that we do not normally have commit messages which are this long, and also to (maybe by force) get the message across to more people 😈 , but either way is fine with me, and I can modify the comment/commit messages and also you please feel free to do it as you think is best.

Perhaps someone wants to later add benchmarks for another toolchain (e.g. clang++) -- then he or she can make a commit which modifies/extends the comments in `wxrc.cpp`. I think this would be more difficult if the benchmarks are in the commit message...

vadz:
Sorry for the delay with coming back to it and thanks for splitting this into separate commits. I'd like to make some comments for each of them:

1. 48c3f2c94f (wxrc: `FileToCppArray`: Resource size is now limited to 4 GiB on all platforms.): I agree with the change, but I think we should keep using `size_t` and just compare that the length of the file is not greater than 4 (or even maybe 2) GiB. I also still think that the explanatory comment could be much shorter and just say that we want the generated file to compile on all systems, including 32 bit ones. Finally, we definitely should use actual error checking and `wxLogError()` instead of `wxASSERT_MSG()` because the failure of the condition being checked doesn't indicate a problem in the program logic but one with its inputs.
2. a3291f8145 (wxrc: `FileToCppArray`: We have replaced `new []` and `delete []` with `std::vector <...>`.): No comments, definitely a good change, thanks.
3. 868c382975 (wxrc: `FileToCppArray`: Temp file info (index, size, name) is now printed before defining each byte-array., 2024-03-25) and 73a99c5c41 (wxrc: `FileToCppArray`: Byte-array initialization integer literals are now hex; progress is optionally printed.): Thanks for making this optional but unless someone else votes for having this, I'd still rather omit these commits, they just seem to add more code without any real benefit for 99.99% of wx users.
4. 4f820010e0 (wxrc: `FileToCppArray`: `static` has been replaced by anonymous namespace. `const` has been added.): `const` is good but I see no real advantage of replacing static. Is there any reason to do it?
5. 9a5d32db95 (wxrc: `GetInternalFileName`: Call to `Printf`: Format string arg is now constant.): Good change, thanks.
6. b8907f2525 (wxrc: `MakePackageCPP`: Faster&leaner compilation of C++ source code generated by our resource compiler.): Also good change, but I'd rather put most of the explanations into the commit message than in the comment in the source.

To summarize, I'd like to apply (2) and (5) as is, (1) and (4) with minor changes and (6) with just some comment reformatting/moving, while omitting (3).

But I'll wait for a bit to see if anybody else wants to have (3). And, of course, please let me know if you have any comments.

DoctorNoobingstoneIPresume:
Thank you for kind answer. My partial answer follows:

> [...]
>     1. [48c3f2c](https://github.com/wxWidgets/wxWidgets/commit/48c3f2c94f8f50ed98fa0da41252a821d2fb22ec) (wxrc: `FileToCppArray`: Resource size is now limited to 4 GiB on all platforms.): I agree with the change, but I think we should keep using `size_t` and just compare that the length of the file is not greater than 4 (or even maybe 2) GiB. I also still think that the explanatory comment could be much shorter and just say that we want the generated file to compile on all systems, including 32 bit ones. Finally, we definitely should use actual error checking and `wxLogError()` instead of `wxASSERT_MSG()` because the failure of the condition being checked doesn't indicate a problem in the program logic but one with its inputs.
> [...]

Please kindly see https://github.com/wxWidgets/wxWidgets/pull/24546.

> [...]
>     5. [9a5d32d](https://github.com/wxWidgets/wxWidgets/commit/9a5d32db95adc191b24fd39217838f6b7f2c091e) (wxrc: `GetInternalFileName`: Call to `Printf`: Format string arg is now constant.): Good change, thanks.
> [...]

Please kindly see https://github.com/wxWidgets/wxWidgets/pull/24547.

vadz:
Your PRs corresponding to (1) and (5) have been merged and I've also pushed 3c41a6876cf8ed574122aacea81ee5014ccfa9a7 which does what (2) do and more.

So, basically, there is just (6) left here.

DoctorNoobingstoneIPresume:
> [...]
>     6. [b8907f2](https://github.com/wxWidgets/wxWidgets/commit/b8907f2525e3ca9c9769a2566b1713b086fe5496) (wxrc: `MakePackageCPP`: Faster&leaner compilation of C++ source code generated by our resource compiler.): Also good change, but I'd rather put most of the explanations into the commit message than in the comment in the source.
> [...]

Please kindly see https://github.com/wxWidgets/wxWidgets/pull/24579...

vadz:
I think there is nothing remaining to merge here any longer, so closing. Thanks!

DoctorNoobingstoneIPresume:
Thank you for support and also for the confidence -- and also for considering even my preferences -- I could see that you have included the benchmarks as C++ comments... thank you so much !! Also, my modest attempts with wxWidgets have brought me happiness during the last few weeks... 😹

PR #24438, opened 2024-03-27
Fixed missing file type filters in ShowWindowModal method (open and s…

gammasoft71:
…ave dialogs)


vadz:
Oh well, I'm applying this in its current state, as ugly as it is, it's still better than not to have this at all and I don't have time to do it properly.

Issue #24440, opened 2024-03-27
Should generic listctrl's DTOR be calling DoDeleteAllItems()?


Blake-Madden:
In the generic list control's DTOR (`~wxListMainWindow()`), it calls `DoDeleteAllItems()`. I don't think this is right because under GTK+ I noticed this behavior:

- I derive from `wxListView` and `Bind` a function in my derived class to `wxEVT_LIST_DELETE_ALL_ITEMS`.
- In my function bound to `wxEVT_LIST_DELETE_ALL_ITEMS`, I do something like read from virtual data.
- When my control is destroyed, the DTOR in my derived class is called. In there, I might delete and/or disconnect from my virtual data.
- Then the base class DTORs are called in reverse order.
- When it gets down to `~wxListMainWindow()`, it calls `DoDeleteAllItems()`.
- This call to `DoDeleteAllItems()` will fire a `wxEVT_LIST_DELETE_ALL_ITEMS` event.
- This event will cause re-entrance into the function I bound to this event in my derived class (that was just destroyed and probably in a suspect state).

I don't think this should be happening and I don't experience this behavior in MSW's native control.

Also, I don't see any sort of necessary DTOR-style cleanup being done in `DoDeleteAllItems()`, so it may not even need to be called here.

======================
Platform: GTK+ (Linux)
wx: 3.3

vadz:
I wonder if the fact that you're not getting `wxEVT_LIST_DELETE_ALL_ITEMS` under Windows is not just another bug, I thought you were supposed to get it there, but looking at the code now it seems that by the time the native control gets `LVN_DELETEALLITEMS`, `wxListCtrl` is already destroyed and its `MSWOnNotify()` is not called any longer...

This, of course, would just confirm that sending events from dtor is not great, but I don't see where else could we do it. From the control's own `wxWindowDestroyEvent` handler, perhaps? This is called just before it is destroyed.

In any case, in your particular case, I think the obvious solution is to unbind your event handler in your derived class dtor.

Blake-Madden:
But if I don't bind to `wxEVT_LIST_DELETE_ALL_ITEMS`, then my derived function that handles clearing out my virtual data never gets called (when explicitly calling `DeleteAllItems()`).

I'll see if moving the event from the DTOR into the destroy event is an option.

vadz:
You should bind to it, but unbind the handler in your class dtor -- this will ensure it's not called from the base class dtor executed later.

[...]

PR #24441, opened 2024-03-27
Document needing to Unbind wxEVT_LIST_DELETE_ALL_ITEMS in derived (generic) listctrl DTORs

Blake-Madden:
`DoDeleteAllItems()` fires an event and can cause a bound function in derived classes to be called after the destruction process starts, causing possibly unexpected behavior. See #24440.

### DIFF
--- a/interface/wx/listctrl.h
+++ b/interface/wx/listctrl.h
@@ -420,9 +420,17 @@ public:
 
         This function does @e not send the @c wxEVT_LIST_DELETE_ITEM
         event because deleting many items from the control would be too slow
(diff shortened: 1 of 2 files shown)


vadz:
Applying this basically makes `wxEVT_LIST_DELETE_ALL_ITEMS` completely useless, as it would be only sent when `DeleteAllItems()` is called and what's the point of generating an event only in reaction to a function call. Perhaps we don't need it at all, of course, but I think it would be better to still send it, just a bit before the destruction.

PR #24442, opened 2024-03-28
Enable/add some datetime tests unrelated to Y2K38

lanurmi:
`TestTimeTicks()` was actually run for very few values; enabled some more of those in the array.

In `TestTimeFormat()` I added some new dates, and found out that any date within a period of DST will fail the test by 1 hour, but only between the years [2031, 2037], and only on MSW and Linux, but not macOS. The timezone of the running environment, and the `TZ` environment variable, also affect this.

For example (on Linux):
`./test DateTimeTestCase` -> fail
`TZ=CET ./test DateTimeTestCase` -> fail
`TZ=CEST ./test DateTimeTestCase` -> no error

As I don't yet know what the reason is, or why 2030/2031 are different, this PR serves both as a PR and an issue with a minimal reproducible example of a problem.

### DIFF
--- a/tests/datetime/datetimetest.cpp
+++ b/tests/datetime/datetimetest.cpp
@@ -180,10 +180,10 @@ struct Date
 static const Date testDates[] =
 {
     {  1, wxDateTime::Jan,  1970, 00, 00, 00, 2440587.5, wxDateTime::Thu,         0 },
-    {  7, wxDateTime::Feb,  2036, 00, 00, 00, 2464730.5, wxDateTime::Thu,        -1 },
-    {  8, wxDateTime::Feb,  2036, 00, 00, 00, 2464731.5, wxDateTime::Fri,        -1 },
-    {  1, wxDateTime::Jan,  2037, 00, 00, 00, 2465059.5, wxDateTime::Thu,        -1 },
-    {  1, wxDateTime::Jan,  2038, 00, 00, 00, 2465424.5, wxDateTime::Fri,        -1 },
+    {  7, wxDateTime::Feb,  2036, 00, 00, 00, 2464730.5, wxDateTime::Thu, 2085955200 },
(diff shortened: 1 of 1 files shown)


lanurmi:
And the tests did not even fail here with CI, although they do fail for me locally. 🤷‍♂️
(Windows 10, Fedora Rawhide; system timezone EET)

vadz:
Sorry, I'm not sure if this one is supposed to be merged or if it's a draft/WIP?

lanurmi:
The tests that were added are simple and completely valid, and tests passed in CI, so from that perspective it should be merged. But as mentioned, tests with the new date in 2031 fail for me locally (and I don't know why), meaning they could fail for others, too. 

Then again, if there is an issue, it is there whether or not this is merged, but without merging the issue is untested, undocumented, and hidden.

vadz:
FWIW I can confirm that this fails for me too locally, so I'm not going to merge it yet, as I want the tests to pass when I run them.

Might be related to #15370.

lanurmi:
Makes sense. 5ad2ca1a alone should be safe to cherry-pick, though.

vadz:
I had a brief look at this and the problem is that the date gets formatted as `1931-07-29 15:28:00` (note the century) and `MakeFromTimezone()` doesn't apply DST to it.

IOW I think things would work if we used a format with 4 digits year instead of `%x`.

vadz:
@lanurmi Do you still plan to update this?

lanurmi:
> @lanurmi Do you still plan to update this?

Probably not, honestly. But as mentioned earlier, 5ad2ca1ab65d277bb4cae772a95470cefaf07b39 on its own is should be harmless to cherry-pick.

vadz:
OK, thanks, I've added it to #25386 and will close this one then.

PR #24448, opened 2024-03-31
Add missing GetDropTarget() method override to wxGrid

### LINKED ISSUE #24447: wxGrid GetDropTarget method always returns null
### Description
I think this is a bug. I noticed GetDropTarget() of wxGrid always returns null even after setting the drop target using the SetDropTarget method.

A quick look at the source code appears to show that the SetDropTarget method is overridden in wxGrid and this calls SetDropTarget on the window returned by the GetGridWindow method. However, the GetDropTarget method does not appear to be overridden to do the same.

I assume the GetDropTarget method should also be overridden to retrieve the one set by SetDropTarget?

I am currently getting around this by calling GetGridWindow()->GetDropTarget() in my application.

Didn't think a how to reproduce etc was necessary for this. Let me know if I'm wrong and I'm happy to add one.

### Platform and Version Information

- wxWidgets version  you use: v3.2.4
- OS  and its version: Windows 10 v22H2 (Build: 19045.4170)

gsxruk:
Added `GetDropTarget()` override method to the `wxGrid` class to match that of the `SetDropTarget(wxDropTarget*)` method.

Fixes #24447

Fingers crossed I have done everything correctly. Let me know if anything requires changing.

### DIFF
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -2376,6 +2376,7 @@ public:
     // ------- drag and drop
 #if wxUSE_DRAG_AND_DROP
     virtual void SetDropTarget(wxDropTarget *dropTarget) override;
+    virtual wxDropTarget* GetDropTarget() const override;
 #endif // wxUSE_DRAG_AND_DROP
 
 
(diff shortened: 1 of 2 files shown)


vadz:

PR #24456, opened 2024-04-04
Fix button size of generic wxColourPickerCtrl in high DPI

taler21:
With wxWidgets-3.2.4, the picker button of the generic `wxColourPickerCtrl` is too large on high DPI screens.

This issue, which can also be seen in the widgets sample, will be fixed by this PR.

The sample under Windows 10 at different DPI scalings:
widgets-sample-colour-picker-win10-hidpi

### DIFF
--- a/src/generic/clrpickerg.cpp
+++ b/src/generic/clrpickerg.cpp
@@ -57,7 +57,9 @@ bool wxGenericColourButton::Create( wxWindow *parent, wxWindowID id,
     // and handle user clicks on it
     Bind(wxEVT_BUTTON, &wxGenericColourButton::OnButtonClick, this, GetId());
 
-    m_bitmap = wxBitmap(FromDIP(defaultBitmapSize));
+    wxBitmap bmp;
+    bmp.CreateWithDIPSize( defaultBitmapSize, GetDPIScaleFactor() );
+    m_bitmap = bmp;
     m_colour = col;
     UpdateColour();
     InitColourData();
@@ -121,7 +123,9 @@ void wxGenericColourButton::OnColourChanged(wxColourDialogEvent& ev)
 
 void wxGenericColourButton::OnDPIChanged(wxDPIChangedEvent& event)
 {
-    m_bitmap = wxBitmap(FromDIP(defaultBitmapSize));
+    wxBitmap bmp;
+    bmp.CreateWithDIPSize( defaultBitmapSize, GetDPIScaleFactor() );
+    m_bitmap = bmp;
     UpdateColour();
 
     event.Skip();


src/generic/clrpickerg.cpp
```diff
@@ -57,7 +57,9 @@ bool wxGenericColourButton::Create( wxWindow *parent, wxWindowID id,
     // and handle user clicks on it
     Bind(wxEVT_BUTTON, &wxGenericColourButton::OnButtonClick, this, GetId());
 
-    m_bitmap = wxBitmap(FromDIP(defaultBitmapSize));
+    wxBitmap bmp;
+    bmp.CreateWithDIPSize( defaultBitmapSize, GetDPIScaleFactor() );
+    m_bitmap = bmp;
```
vadz, line 62: This could be
```suggestion
    m_bitmap.CreateWithDIPSize( defaultBitmapSize, GetDPIScaleFactor() );
```
or is there something I'm missing here?

vadz (commented):
Thanks for fixing this!

I'll apply it to master and backport to 3.2, but I'd like to simplify it as indicated, unless there is some reason not to?

taler21:
This simplification is fine, thanks.


It may also be possible in `wxGenericColourButton::OnDPIChanged()` since `wxBitmap::CreateWithDIPSize()` calls `wxBitmap::Create()` which creates a fresh bitmap, meaning it can also be called on an already created bitmap to recreate it.

https://github.com/wxWidgets/wxWidgets/blob/80653ec06461430f76fe4c5ddf79bc2bddb65e6a/src/msw/bitmap.cpp#L764-L768

https://github.com/wxWidgets/wxWidgets/blob/80653ec06461430f76fe4c5ddf79bc2bddb65e6a/interface/wx/bitmap.h#L445-L458

https://github.com/wxWidgets/wxWidgets/blob/80653ec06461430f76fe4c5ddf79bc2bddb65e6a/src/msw/bitmap.cpp#L785-L791

And that doesn't seem to be specific to wxMSW.


However, it is not explicitly documented that `wxBitmap::CreateWithDIPSize()` recreates the bitmap

https://github.com/wxWidgets/wxWidgets/blob/80653ec06461430f76fe4c5ddf79bc2bddb65e6a/interface/wx/bitmap.h#L481-L512

and so probably one should not assume that this is always the case.

vadz:
Thanks, I think it should be fine to call `CreateWithDIPSize()` multiple times, it recreates the bitmap if it already exists, so I'm merging this with this change.

BTW, the really correct thing to do would be to avoid using hardcoded `defaultBitmapSize` and infer the size of the bitmap from the size of the control, which would address [this old question](https://stackoverflow.com/questions/35992169/how-can-i-make-a-wxcolorpickerctrl-narrower/36022643) and generally make much more sense — if anyone feels like doing it, please feel free to do it!

PR #24457, opened 2024-04-05
Add missing 'textctrl.h' inclusion

Blake-Madden:
Without this, I get a number of compile issues where ` wxTextCtrl* m_userTextCtrl;` is declared. (At least on MSW.)


vadz (commented):
Thanks, there is definitely a problem here, but I'd prefer to use a forward declaration, so I'll add one instead.

PR #24461, opened 2024-04-07
Implement support for requesting light/dark app appearance

vadz:
We've had several requests for this, so I've finally implemented it. This is relatively straightforward, but I had to refactor wxGTK code quite a bit to allow reusing parts of it that I needed.

Any comments about API or the implementation (especially under Mac, where, as usual, I don't really know what I'm doing) would be welcome, TIA!

### DIFF
--- a/interface/wx/app.h
+++ b/interface/wx/app.h
@@ -965,6 +965,69 @@ public:
     */
     bool ProcessMessage(WXMSG* msg);
 
+
+    /**
+        Possible parameters for SetAppearance().
+
+        @since 3.3.0
+    */
+    enum class Appearance
+    {
+        System, ///< Use system default appearance.
+        Light,  ///< Use light appearance.
+        Dark    ///< Use dark appearance.
+    };
+
+    /**
+        Possible values returned by SetAppearance().
+
+        @since 3.3.0
+    */
+    enum class AppearanceResult
+    {
+        Failure,     ///< Changing the appearance failed.
+        Success,     ///< Appearance was successfully changed.
+        CannotChange ///< Appearance can't be changed any more.
+    };
+
+    /**
+        Request using either system default or explicitly light or dark theme
+        for the application.
+
+        Under GTK and macOS applications use the system default appearance by
+        default, and so it is only useful to call this function with either
+        Appearance::Light or Appearance::Dark parameters if you need to
+        override the default system appearance. The effect of calling this
+        function is immediate, i.e. this function returns
+        AppearanceResult::Success, and affects all the existing windows as well
+        as any windows created after this call.
+
+        Under MSW, the default appearance is always light and the applications
+        that want to follow the system appearance need to explicitly call this
+        function with Appearance::System parameter in order to do it. Please
+        note using dark appearance under MSW requires using non-documented
+        system functions and has several known limitations, please see
+        MSWEnableDarkMode() for more details. Also, on this platform the
+        appearance can be only set before any windows are created and calling
+        this function too late will return AppearanceResult::CannotChange.
+
+        Note that to query the current appearance, you can use
+        wxSystemAppearance, see wxSystemSettings::GetAppearance().
+
+        @return AppearanceResult::Success if the appearance was successfully
+            changed or had been already set to the requested value,
+            AppearanceResult::CannotChange if the appearance can't be changed
+            any more because it's too late to do it but could be changed if
+            done immediately on next program launch (only returned by wxMSW
+            currently) or AppearanceResult::Failure if changing the appearance
+            failed for some other reason, e.g. because `GTK_THEME` is defined
+            when using wxGTK of this function is not implemented at all for
+            the current platform.
+
+        @since 3.3.0
+    */
+    bool SetAppearance(Appearance appearance);
+
     /**
         Set display mode to use. This is only used in framebuffer wxWidgets
         ports such as wxDFB.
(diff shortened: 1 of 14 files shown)


csomor (commented):
Thanks, the mac part looks ok to me (although that API has been deprecated already in macOS 11 ... as usual short lived ...)

src/gtk/settings.cpp
```diff
@@ -183,40 +188,132 @@ static void notify_gtk_font_name(GObject*, GParamSpec*, void*)
...
+        // Also don't do anything if the preference didn't change.
+        wxLogTrace(TRACE_DARKMODE, "Dark mode preference didn't change");
+        return false;
+    }
+
+    wxLogTrace(TRACE_DARKMODE, "Turning dark mode preference %s",
+               preferDark ? "on" : "off");
+
+    g_object_set(settings,
+        "gtk-application-prefer-dark-theme", preferDark, nullptr);
```
paulcor, line 255: You can't change the theme this way if `GTK_THEME` is set, as it overrides `GtkSettings`, so `UpdatePreferDarkWithoutRefresh()` should probably not report success in that case.

src/gtk/app.cpp
```diff
@@ -347,6 +351,36 @@ bool wxApp::SetNativeTheme(const wxString& theme)
 #endif
 }
 
+wxApp::AppearanceResult wxApp::SetAppearance(Appearance appearance)
+{
+#ifdef __WXGTK3__
+    gboolean preferDark = FALSE;
```
paulcor, line 357: As far as I know, it's perfectly fine to assign `true`/`false` to an `int`, so I don't see any reason not to prefer those over the old `TRUE`/`FALSE`.

src/gtk/settings.cpp
```diff
@@ -183,40 +188,132 @@ static void notify_gtk_font_name(GObject*, GParamSpec*, void*)
...
+
+} // anonymous namespace
+
+// Functions declared in wx/gtk/private/appearance.h and used by wxApp.
+namespace wxGTKImpl
+{
+
+bool GetPreferDarkFromColorScheme(gboolean* preferDark)
+{
+    GVariant* ret = g_dbus_proxy_call_sync(gs_proxyPortalSettings, "Read",
```
paulcor, line 282: It seems possible for `gs_proxyPortalSettings` to be null here.

src/gtk/settings.cpp
```diff
@@ -183,40 +188,132 @@ static void notify_gtk_font_name(GObject*, GParamSpec*, void*)
...
 
-static bool UpdatePreferDark(GVariant* value)
+namespace
+{
+
+constexpr const char* TRACE_DARKMODE = "darkmode";
+
+// This function fills in the provided pointer (which must be non-null) and
+// returns GtkSettings object which is also used to indicate a failure.
+GtkSettings* GetPreferDark(gboolean* preferDarkPrev)
```
paulcor, line 198: If `preferDarkPrev` must be non-null, it seems like it should be a reference.

interface/wx/app.h
```diff
@@ -965,6 +965,64 @@ class wxApp : public wxAppConsole
...
+    };
+
+    /**
+        Request using either system default or explicitly light or dark theme
+        for the application.
+
+        Under GTK and macOS applications use the system default appearance by
+        default, and so it is only useful to call this function with either
+        Appearance::Light or Appearance::Dark parameters if you need to
+        override the default system appearance. The effect of calling this
```
PBfordev, line 1000: I wonder if we should note the existence of `wxSystemAppearance` here and analogically, reference this method in `wxSystemAppearance` docs.
```suggestion
        override the default system appearance (see wxSystemAppearance). The effect of calling this
```

The wxWidgets API is very extensive and sometimes it may be hard to find the correct method/class that does what is wanted...

src/gtk/settings.cpp
vadz: Thanks, indeed. Now I wonder if we should have different `CannotChangeBecauseOverriddenByTheme` and `CannotChangeWithoutRestart` return codes (except preferably shorter), as the application would probably want to handle them differently.

src/gtk/app.cpp
vadz: IMO this is more clear: even if C++ allows you to assign just about [anything to anything else](https://fosstodon.org/@VZ/112235960416690642), using `false` with `bool` variables and `FALSE` with `int` ones is IMO tidier and I've always done it like this in wxMSW (which has its own `BOOL` which is actually `int`) and would rather keep doing it in wxGTK too (where there are already more than 330 occurrences of these constants) unless you absolutely hate it.

src/gtk/settings.cpp
vadz: We rarely use output parameters but when we do, we typically use pointers to them in order to clearly indicate that they're been modified in the caller. As I find this more readable I'd like, again, to keep it like this unless you object violently.

src/gtk/settings.cpp
vadz: Fixed, thanks.

interface/wx/app.h
vadz: Thanks, I've added xrefs in both directions now.

vadz:
I've addressed the non-style review comments now, thanks!

If there are no other ones, I'll merge this soon.

paulcor:
I think there may be another problem with this, at least for GTK. If the user changes the global DE light/dark setting after `SetAppearance()` is called, the `SetAppearance()` setting will be overridden.

vadz:
> I think there may be another problem with this, at least for GTK. If the user changes the global DE light/dark setting after `SetAppearance()` is called, the `SetAppearance()` setting will be overridden.

Thanks, this was indeed a bug and I've fixed it now but I had to change the code a lot in order to do it and so I've also squashed some commits to get rid of some refactorings that turned out to not be really useful any more.

I also had to introduce RAII `wxGtkVariant` helper here, which I wanted to postpone until later, but without it the code became more difficult to write, so I've finally done it here.

Thanks again for reviewing this!

PR #24464, opened 2024-04-11
Support Y2K38 in wxDateTime

### LINKED ISSUE #24414: tests fail after 2038
#### Bug description:

While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds), I found that our wxWidgets-3_2 (3.2.4) package fails tests when the system clock is set to the year 2038.

test output looks thus:
```
 DateTimeTestCase
   TestTimeZoneParse
 ./datetime/datetimetest.cpp:223
 
 ./datetime/datetimetest.cpp:1014: FAILED: 
   REQUIRE( 13 == dt.GetHour(wxDateTime::UTC) )
 with expansion:
   13 == 14
 
 tar: Malformed extended header: invalid atime=1.07814e+09
 tar: Malformed extended header: invalid ctime=2.21842e+09
 tar: Malformed extended header: invalid atime=1.07814e+09
 tar: Malformed extended header: invalid ctime=2.21842e+09
...
 test cases:     313 |     312 passed | 1 failed
 assertions: 1229575 | 1229574 passed | 1 failed
```

Tests should keep working in the future - at least 16 years would be good.

#### Stack trace:

N/A

#### To Reproduce:

on Debian or openSUSE do
```bash
osc checkout openSUSE:Factory/wxWidgets-3_2 && cd $_
osc build --noservice --vm-type=kvm --build-opt=--vm-custom-opt="-rtc base=2040-04-19T02:01:51" standard x86_64
```

Sometimes reproduction is not reliable - it might depend on the exact moment the test is run and how floating point values are rounded.

- wxWidgets version  you use: 3.2.4
- OS  and its version: openSUSE-Tumbleweed 20240313 (x64_64)

lanurmi:
* Return 64-bit values from `GetTicks()` if `time_t` is 64-bit
* Use CRT for dates up to (the arbitrarily chosen) year 2200
* Add some 64-bit timestamps to test

There are actually fewer changes than I expected to be needed; I hope this is all that it takes.

This is could and probably should be backported to 3.2.

May fix #24414.
Should fix #22561.

### DIFF
--- a/include/wx/datetime.h
+++ b/include/wx/datetime.h
@@ -1728,9 +1728,11 @@ protected:
 
 inline bool wxDateTime::IsInStdRange() const
 {
-    // currently we don't know what is the real type of time_t so prefer to err
-    // on the safe side and limit it to 32 bit values which is safe everywhere
-    return m_time >= 0l && (m_time / TIME_T_FACTOR) < wxINT32_MAX;
+    // if sizeof(time_t) is greater than 32 bits, we assume it
+    // is safe to return values exceeding wxINT32_MAX
+
+    return m_time >= 0l &&
+        ( (sizeof(time_t) > 4 ) || ( (m_time / TIME_T_FACTOR) < wxINT32_MAX) );
 }
 
 /* static */
@@ -1837,7 +1839,7 @@ inline time_t wxDateTime::GetTicks() const
         return (time_t)-1;
     }
 
-    return (time_t)((m_time / (long)TIME_T_FACTOR).ToLong()) + WX_TIME_BASE_OFFSET;
+    return (time_t)((m_time / TIME_T_FACTOR).GetValue()) + WX_TIME_BASE_OFFSET;
 }
 
 inline bool wxDateTime::SetToLastWeekDay(WeekDay weekday,
--- a/src/common/datetime.cpp
+++ b/src/common/datetime.cpp
@@ -1273,14 +1273,13 @@ wxDateTime& wxDateTime::Set(wxDateTime_t day,
     // While the range for 64-bit time_t is billions of years per se,
     // we cannot expect the C runtime to support dates thousands of years
     // in the future. MSVC claims to support dates up to 3000-12-31;
-    // what macOS and *nix support is unknown, but let us pick year 2200
-    // as the maximum for now.
+    // however, we do not set an upper limit for the year.
     static const int yearMinInRange = 1970;
-    static const int yearMaxInRange = sizeof(time_t) > 4 ? 2200 : 2037;
+    static const int yearMaxInRange = sizeof(time_t) > 4 ? -1 : 2037;
 
     // test only the year instead of testing for the exact end of the Unix
     // time_t range - it doesn't bring anything to do more precise checks
-    if ( year >= yearMinInRange && year <= yearMaxInRange )
+    if ( year >= yearMinInRange && (yearMaxInRange == -1 || year <= yearMaxInRange) )
     {
         // use the standard library version if the date is in range - this is
         // probably more efficient than our code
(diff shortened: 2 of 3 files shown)


include/wx/datetime.h
```diff
@@ -1728,9 +1728,11 @@ class WXDLLIMPEXP_BASE wxDateTimeChristianHolidays : public wxDateTimeUSCatholic
...
 inline bool wxDateTime::IsInStdRange() const
 {
-    // currently we don't know what is the real type of time_t so prefer to err
-    // on the safe side and limit it to 32 bit values which is safe everywhere
-    return m_time >= 0l && (m_time / TIME_T_FACTOR) < wxINT32_MAX;
+    // if sizeof(time_t) is greater than 32 bits, we assume it
+    // is safe to return values exceeding wxINT32_MAX
+
+    return m_time >= 0l &&
+        ( (sizeof(time_t) > 4 ) || ( (m_time / TIME_T_FACTOR) < wxINT32_MAX) );
```
vadz, line 1735: It looks like this ought to use use the same limit as the code in `datetime.cpp` (i.e. 2200 right now, but see there), shouldn't it?

src/common/datetime.cpp
```diff
@@ -1269,8 +1270,13 @@ wxDateTime& wxDateTime::Set(wxDateTime_t day,
                       wxT("Invalid date in wxDateTime::Set()") );
 
     // the range of time_t type (inclusive)
+    // While the range for 64-bit time_t is billions of years per se,
+    // we cannot expect the C runtime to support dates thousands of years
+    // in the future. MSVC claims to support dates up to 3000-12-31;
+    // what macOS and *nix support is unknown, but let us pick year 2200
```
vadz, line 1276: I don't really like arbitrary picking up the values out of thin air like this. What's the worst that can happen if try using the standard library for the year 12345678 (i.e. infinitely far in the future)? It's probably not going to do worse than our own code here, is it?

I.e. I'd drop the check for max year here (and then we wouldn't need to change the code in `IsInStdRange()`).

vadz (commented):
Thanks, it's reassuring (even if not really surprising) that the changes are so small, but I'd prefer not to hardcode the 2200 limit.

And, ideally, we'd need more tests, i.e. add dates after 2038 to at least some other tests to check that the relevant functionality keeps working after 2K38 when `sizeof(time_t) == 8` (and skip these tests if it isn't).

src/common/datetime.cpp
lanurmi: > I don't really like arbitrary picking up the values out of thin air like this. What's the worst that can happen if try using the standard library for the year 12345678 (i.e. infinitely far in the future)? It's probably not going to do worse than our own code here, is it?

On MSVC, `localtime` and `mktime` will fail (and return `NULL` and `-1`, respectively) for dates after 3000, per the documentation. But I cannot judge if that is better or worse than what wx's own code does.

> I.e. I'd drop the check for max year here (and then we wouldn't need to change the code in `IsInStdRange()`).

I'll do that.

lanurmi:
> And, ideally, we'd need more tests, i.e. add dates after 2038 to at least some other tests to check that the relevant functionality keeps working after 2K38 when `sizeof(time_t) == 8` (and skip these tests if it isn't).

As far as I see, most arrays of dates in `datetimetest.cpp` already have dates after 2038, but I'll also add some more to those that do not have.

vadz:
Thanks, I've rebased, done some minor changes (mostly to the comments which still mentioned 2200) and will merge it soon.

It also can be backported to 3.2 AFAICS, so let's do it later, if we don't find any problems with it in master.

vadz:
Merged into master by the commit above, so closing.

Issue #24465, opened 2024-04-11
Adding ability to change size of the text control in wxTextEntryDialog


Blake-Madden:
I am showing a multiline `wxTextEntryDialog` for my users to enter a large block of text. However, the default size for this dialog isn't very tall (maybe enough space vertically for two lines of text). You always have to manually resize it to see larger text blocks. The CTOR doesn't have a size param (the control is hard-coded to `wxSize(300, wxDefaultCoord)`) and I don't see any way to change the size of the text control in the current API other than deriving from `wxTextEntryDialog` and accessing its `m_textCtrl` (which is protected).

That's easy enough, but would there be an in interest in having an API for this. For example, adding a `SetTextEntrySize(wxSize)` function?

vadz:
It would make sense to add `wxSize` parameter to the dialog ctor, for consistency with just about everything else. This wouldn't allow precisely setting the text entry size, but do we really need this? IMHO just making the dialog "big enough" should be sufficient, in practice.

PR #24467, opened 2024-04-12
Add size parameter to wxTextEntryDialog

Blake-Madden:
I fit the sizer to the dialog (instead of the other way around) after setting the dialog's min size.

Leaving the default results in the same appearance as before, and providing a custom sizes works.

This implements #24465.

### DIFF
--- a/interface/wx/textdlg.h
+++ b/interface/wx/textdlg.h
@@ -117,7 +117,7 @@ public:
         @param pos
             Dialog position.
         @param sz
-            The size of the dialog.
+            The size of the dialog (this parameter is only available since wxWidgets 3.3.0).
 
         @since 2.9.5
     */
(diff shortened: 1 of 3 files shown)


interface/wx/textdlg.h
```diff
@@ -115,14 +116,17 @@ class wxTextEntryDialog : public wxDialog
             make sense for this dialog, used for text input.
         @param pos
             Dialog position.
+        @param sz
+            The size of the dialog.
```
vadz, line 120:
```suggestion
            The size of the dialog (this parameter is only available since wxWidgets 3.3.0).
```

vadz (approved):
Looks good, thanks!

Will merge soon with the correction below applied.

PR #24468, opened 2024-04-12
Add wxGridCellCoords::IsFullySpecified()

Blake-Madden:
Provides parity with `wxPoint` and an alternative to comparing against `wxGridNoCellCoords`.
Also, cleaned up some stray spaces.

Related question: in a separate PR, should I replace `-1` with `wxDefaultCoord` in this class's API?


vadz:
Thanks, will add soon.

I don't know if replacing `-1`s is really worth it.

PR #24474, opened 2024-04-15
added stock accelerators for Close and SaveAs

wouterbt:
wxWidgets includes a number of stock buttons and menu items, with associated mnemonics, help texts and images (under GTK). Some of these also have stock accelerators. This pull requests adds stock accelerators for the Close (Ctrl+W) and Save As... (Ctrl+Shift+S) menu items.


vadz:
Thanks, this looks reasonable and at least both [GNOME](https://developer.gnome.org/hig/reference/keyboard.html) and [Apple](https://developer.apple.com/design/human-interface-guidelines/keyboards) document these keys as being standard for these commands (I also found a similar [page for Windows](https://learn.microsoft.com/en-us/windows/win32/uxguide/inter-keyboard) but it doesn't document the standard shortcuts).

I'll (squash-)merge this soon if there are no objections.

Issue #24216, opened 2024-01-14
Add an option for a different SVG rendering engine


Randalphwa:
### Description

While nanosvg has the advantage of being easy to build as part of wxWidgets, it has the disadvantage of not correctly displaying a variety of different types of SVG files. The following image pairs shows some of the problems I've encountered with the nanosvg 
rendering (nanosvg on the left, correct rendering on the right):

![comparison](https://github.com/wxWidgets/wxWidgets/assets/38287198/e3e53fe4-f91b-42a4-ab6b-956f3c7a4d99)

Issue #22114 discusses the possibility of using ThorVG, however I'd like to suggest an alternative as well as offering to implement it. I've recently finished a review of lunasvg (https://github.com/sammycage/lunasvg) which is an active repository using an MIT license. I actually used it in the comparisons above since so far I haven't encountered an SVG file that it doesn't render the same as it does in InkScape and Affinity Designer. I've already experimented with integrating it into the wxWidgets build system which simply meant adding the filenames to the build list as well as appending one additional include path. In looking at the bmpsvg.cpp code, I think it should be reasonably straightforward to conditionally integrate it based on a setup.h define (e.g. wxUSE_LUNASVG).

I contacted the author of lunasvg and verified that he would be okay with integrating this as a submodule in wxWidgets. He also offered any assistance needed should there be any problems with the integration.

If this seems like a reasonable idea, I'll start working on a branch to implement it. Note that I'm _not_ suggesting that this replace nanosvg, but rather that it be an alternative that can be selected at build time.

hermet:
If you're exploring a brand new vector formats such as Lottie ([Lottie File Format](https://en.wikipedia.org/wiki/Lottie_(file_format))), you may want to consider using ThorVG. ThorVG supports both SVG and Lottie formats. It's a compact library that is easy to integrate and port across various platforms.

Randalphwa:
The advantage of nanosvg is that it's a single header file, which made it fairly trivial to integrate into wxWidgets. I don't recall whether development stopped before or after that integration -- while crashing bugs have been fixed, there's no actual development.

nanosvg is a lot older -- the original license was created 11 years ago, versus 3 years ago for lunasvg. At the time you were looking for SVG rendering engines, you might not have even heard of lunasvg.

lunasvg does not support Text. The developper is working on it, but from what I gather it's a difficult integration.

[...]

PR #24475, opened 2024-04-15
Add LunaSVG as an optional replacement for NanoSVG

Randalphwa:
This PR adds lunasvg as an optional replacement of nanosvg when rendering SVG files. See #24216 for a detailed explanation of the advantages of lunasvg (better accuracy, equivalent performance). The implementation is designed to use eithr nanosvg or lunasvg, but not both. The choice is controlled by `wxUSE_LUNASVG` in setup.h (defaults to 1).

Lunasvg is a submodule consisting of a C++ library (wxlunasvg) and a C library (wxplutovg).

I am not very familiar with either bakefile or the various ways in which wxWidgets can be built. I _think_ I got the various bakefiles changes made correctly, but a closer review from someone more familiar with the build system would be greatly appreciated.

As a side note, I added PB to the copyright authors in bmpsvg.cpp since I essentially used his code for testing lunasvg with some minor modifications when I integrated it into wxWidgets.


configure.ac
```diff
@@ -530,6 +530,7 @@ WX_ARG_WITH(libnotify,     [  --with-libnotify        use libnotify for notifica
 WX_ARG_WITH(opengl,        [  --with-opengl           use OpenGL (or Mesa)], wxUSE_OPENGL)
 WX_ARG_WITH(xtest,         [  --with-xtest            use XTest extension], wxUSE_XTEST)
 WX_ARG_WITH(nanosvg,       [  --with-nanosvg          use NanoSVG for rasterizing SVG], wxUSE_NANOSVG)
```
vadz, line 532: We should probably turn this off by default now. And maybe give a warning if both it and LunaSVG are turned on.

interface/wx/bmpbndl.h
```diff
@@ -246,15 +246,18 @@ class wxBitmapBundle
...
+        library (https://github.com/memononen/nanosvg) (since wxWidgets 3.1.7)
+        for parsing and rasterizing SVG images which imposes the following
+        limitations:
 
         - Text elements are not supported at all (see note for workaround).
         - SVG 1.1 filters are not supported.
         - Embedded images are not supported (see note for workaround).
 
-        These limitations will be relaxed in the future wxWidgets versions.
+        These limitations may be relaxed in future wxWidgets versions (they are
```
vadz, line 259: Minor, but I'd say
```suggestion
        These limitations may be relaxed in future wxWidgets versions (these features are
```
to avoid implying that limitations are planned in LunaSVG :-)

.circleci/config.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - run:
           name: Checkout required submodules
           command: |
-            git submodule update --init 3rdparty/catch 3rdparty/nanosvg src/stc/scintilla src/stc/lexilla
+            git submodule update --init 3rdparty/catch 3rdparty/nanosvg 3rdparty/lunasvg src/stc/scintilla src/stc/lexilla
```
vadz, line 36: We probably don't need this one any more?
```suggestion
            git submodule update --init 3rdparty/catch 3rdparty/lunasvg src/stc/scintilla src/stc/lexilla
```

.gitmodules
```diff
@@ -38,3 +38,6 @@
 	path = src/stc/lexilla
 	url = https://github.com/wxWidgets/lexilla.git
 	branch = wx
+[submodule "3rdparty/lunasvg"]
+	path = 3rdparty/lunasvg
+	url = https://github.com/sammycage/lunasvg.git
```
vadz, line 43: I've created [wxWidgets/lunasvg](https://github.com/wxWidgets/lunasvg) fork, just in case we need to make some wx-specific changes later, let's use this one here.

vadz (commented):
I didn't have time to look at the build errors yet, do you already know what the problem is or do you need help with finding it?

As for the code itself, it would be nice if we could avoid repeating so much of the existing code, even at the price of having more conditional compilation directives...

Randalphwa:
> I didn't have time to look at the build errors yet, do you already know what the problem is or do you need help with finding it?

Yes I know what the problem is, but I'm not thrilled with the solution. The underlying issue is that lunasvg is being created as a dll when a non-monolithic shared build is created. That makes sense to me for 3rdparty code like scintilla which is reasonably large, but not for a tiny bit of code like lunasvg. I would rather see it compiled directly into wxCore if wxUSE_LUNASVG is true so that devs don't have to change their setup scripts to install yet another wxWidgets dll in their packages. In any event, the current build issue is that the code is in a library and not being linked. Is there any way to conditionalize the source code in the bakefile scripts so that it could be compiled directly into wxCore? 

Digressing somewhat, but I did submit a PR to the lunasvg project with all of the plutovg C code converted to C++ so that a separate library was no longer needed. The PR was rejected on the grounds that it is possible the plutovg code might be changed someday, and the changes would have to be merged into the C++ code. That didn't really make sense to me since plutovg is owned/maintained by the same person, but that's not my call. It _could_ be done in wxWidgets, but of course would require a wxWidgets dev to update it should plutovg change.

.circleci/config.yml
Randalphwa: We still need it for as long as there is an option to use nanosvg instead of lunasvg.

If the wxWidgets/lunasvg fork is used, then perhaps nanosvg could be dropped entirely? Should an issue arise with displaying an SVG file then it could theoretically be fixed in the fork if it doesn't get fixed in the original repository.

vadz:
> Yes I know what the problem is, but I'm not thrilled with the solution. The underlying issue is that lunasvg is being created as a dll when a non-monolithic shared build is created. That makes sense to me for 3rdparty code like scintilla which is reasonably large, but not for a tiny bit of code like lunasvg. I would rather see it compiled directly into wxCore if wxUSE_LUNASVG is true so that devs don't have to change their setup scripts to install yet another wxWidgets dll in their packages. In any event, the current build issue is that the code is in a library and not being linked. Is there any way to conditionalize the source code in the bakefile scripts so that it could be compiled directly into wxCore?

I'm not sure why is a shared library used for it when Lexilla/Scintilla are built as static libraries...

> Digressing somewhat, but I did submit a PR to the lunasvg project with all of the plutovg C code converted to C++ so that a separate library was no longer needed. The PR was rejected on the grounds that it is possible the plutovg code might be changed someday, and the changes would have to be merged into the C++ code. That didn't really make sense to me since plutovg is owned/maintained by the same person, but that's not my call. It _could_ be done in wxWidgets, but of course would require a wxWidgets dev to update it should plutovg change.

I haven't looked at this code at all, so I have no idea about how likely plutovg is to change but you're going to become the (de facto, sorry if you don't want it :-) maintainer of this part of the code, so if you think it's worth merging them, please feel free to do it — I certainly wouldn't object to it if it makes maintaining it simpler.

PBfordev:
I took a look only at the source code, as it is based on mine.

My code was written to be tested against wxWidgets NanoSVG implementation and as such is not best suited for integration into wxWidgets code base as is (as also noted by Vadim). I would change it to match the bundle implementation using NanoSVG (and thus my credit should be removed from the file header). In other words, the bundle should have just one ctor taking `lunasvg::Document` and the string/file manipulation should be done in the static `wxBitmapBundle::From*()` methods (perhaps shared for both implementations?) instead of the bundle ctors. Please also note that the current implementation loading from a file is incorrect, not using `wxNO_SVG_FILE` and possibly (however unlikely) erroring out at build time. All this also means that most wxWidgets/standard includes do not need to be listed separately for each implementation, they should be listed just once inside the `wxHAS_SVG`  block.

Moreover, it seems that the code uses my old rasterization method, which did not scale non-square SVGs properly. While such SVGs may not be commonly used for GUI elements, I think we should always maintain the aspect ratio, see https://github.com/PBfordev/wxtestsvg2/commit/a8953e9601147291953afcfc0e4916941e436e5c

.circleci/config.yml
Randalphwa: Ack, I was thinking optional in general, as in an individual building wxWidgets could choose. You are correct that for the CI build, there is no point to getting the nanosvg submodule.

Randalphwa:
> > Yes I know what the problem is, but I'm not thrilled with the solution. The underlying issue is that lunasvg is being created as a dll when a non-monolithic shared build is created. That makes sense to me for 3rdparty code like scintilla which is reasonably large, but not for a tiny bit of code like lunasvg. I would rather see it compiled directly into wxCore if wxUSE_LUNASVG is true so that devs don't have to change their setup scripts to install yet another wxWidgets dll in their packages. In any event, the current build issue is that the code is in a library and not being linked. Is there any way to conditionalize the source code in the bakefile scripts so that it could be compiled directly into wxCore?
> 
> I'm not sure why is a shared library used for it when Lexilla/Scintilla are built as static libraries...

Probably because I misunderstood the .bkl options, in spite of using lexilla.bkl as the basis for lunasvg.bkl. I'll fix it.

> 
> > Digressing somewhat, but I did submit a PR to the lunasvg project with all of the plutovg C code converted to C++ so that a separate library was no longer needed. The PR was rejected on the grounds that it is possible the plutovg code might be changed someday, and the changes would have to be merged into the C++ code. That didn't really make sense to me since plutovg is owned/maintained by the same person, but that's not my call. It _could_ be done in wxWidgets, but of course would require a wxWidgets dev to update it should plutovg change.
> 
> I haven't looked at this code at all, so I have no idea about how likely plutovg is to change but you're going to become the (de facto, sorry if you don't want it :-) maintainer of this part of the code, so if you think it's worth merging them, please feel free to do it — I certainly wouldn't object to it if it makes maintaining it simpler.

I have a vested interest in ensuring that lunasvg works correctly, since my main app replaced all it's PNG files with 160+ newly created SVG files. I'm fine with with being the de facto maintainer of wxlunasvg for as long as I am able to. I assume updating the widgets fork is the same as widgets itself, i.e., submit a PR?

vadz:
> Probably because I misunderstood the .bkl options, in spite of using lexilla.bkl as the basis for lunasvg.bkl. I'll fix it.

Thanks! And I know that dealing with bkl is painful, I'd be glad to replace it with something else, it's just that I don't have anything (see dozens of previous emails about it...).

> I assume updating the widgets fork is the same as widgets itself, i.e., submit a PR?

Yes, absolutely. TIA!

Randalphwa:
I'm in the process of refactoring the `wxBitmapBundleImplSVG` source code as per PB's suggestions, and I have a question about conflicting comments in the header file.

NanoSVG needs a `FromSVG(char* data` method because it modifies that data, and the other variants of this function make copies of the data. LunaSVG only takes a `const char*` so it will never directly modify the data passed to it. Therefore all the comments about making a copy and needing to modify the data _only_ apply to NanoSVG. Fortunately the interface file doesn't mention this, but I'm not sure how best to reflect the differences between the two systems in the header file in terms of comments.

vadz:
Sorry if I'm missing something, but to me the answer seems rather clear: we should still provide the overload taking `char*` in the public API, but it could just do the same thing as the overload taking `const char*` internally.

Randalphwa:
Sorry I wasn't clear -- the question is about the comments themselves:

```
    // Notice that the data here is non-const because it can be temporarily
    // modified while parsing it.
    wxNODISCARD static wxBitmapBundle FromSVG(char* data, const wxSize& sizeDef);
```

This is true for nanosvg only -- lunasvg does not modify the data.

```
    // This overload currently makes a copy of the data.
    wxNODISCARD static wxBitmapBundle FromSVG(const char* data, const wxSize& sizeDef);
```

This is true for nanosvg only -- lunasvg does not make a copy of the data.

The implication of these two comments is that a user should either provide a pointer to mutable data, or wxWidgets will duplicate the data for them at the cost of double the amount of memory required. While that is true for nanosvg, it is not true for lunasvg.

As for the functions themselves, yes both types are provided to provide backwards compatibility. It is the comments themselves that are a misleading for the lunasvg case. Am I being too nit-picky here?

vadz:
Ah, sorry. I'd just say something like:

```cpp
// Some implementations make a copy of the data that they modify in place and it's more
// efficient to provide them a non-const buffer if it's already available.
```

rcane:
Looking at the PR, I see that there is currently no option to link against an external lunasvg (like with nanosvg). I am asking this because not being able to link against an external lunasvg makes it basically impossible to use wx as a static library and also have your own lunasvg lib in the same application. Or are there plans to add a prefix to all symbols in the forked version of lunasvg so that two versions can coexist without getting symbol clashes?

Randalphwa:
> Looking at the PR, I see that there is currently no option to link against an external lunasvg (like with nanosvg). I am asking this because not being able to link against an external lunasvg makes it basically impossible to use wx as a static library and also have your own lunasvg lib in the same application. Or are there plans to add a prefix to all symbols in the forked version of lunasvg so that two versions can coexist without getting symbol clashes?

Nanosvg was added as a submodule of the original repository with no modifications made to it specifically for wxWidgets. Lunasvg is being added as a fork under wxWidgets so that it _can_ be modified with wxWidgets-specific content. If changes are made to add functionality to the wxWidgets version, then switching to an external library might result in a loss of functionality for wxBitmapBundle rendering.

I haven't tried this yet, but it should work to simply change the namespace in the wxWidgets version to `wxlunasvg` instead of `lunasvg`. That way you could link to both wxlunasvg.lib and lunasvg.lib in the same project.

rcane:
Using `wxlunasvg` as the namespace sound like a good idea in any case. It makes it clear that the fork is connected to wx. But I was actually more concerned about plutovg which is a C library. You mentioned that you had a PR to integrate plutovg into lunasvg. This would fix this problem. But as you and Vadim already realized, it would mean that the burden of maintaining this would fall on the wx community.

src/generic/bmpsvg.cpp
```diff
@@ -19,13 +19,33 @@
 // for compilers that support precompilation, includes "wx.h".
 #include "wx/wxprec.h"
 
-#if wxUSE_LUNASVG
+#if defined(wxHAS_SVG) && wxUSE_LUNASVG
     #if !(__cplusplus >= 201703L || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L))
```
PBfordev, line 23: Could be simplified with `wxCHECK_CXX_STD` macro.

build/cmake/options.cmake
```diff
@@ -111,7 +111,12 @@ wx_add_thirdparty_library(wxUSE_EXPAT EXPAT "use expat for XML parsing" DEFAULT_
 wx_add_thirdparty_library(wxUSE_LIBJPEG JPEG "use libjpeg (JPEG file format)")
 wx_add_thirdparty_library(wxUSE_LIBPNG PNG "use libpng (PNG image format)")
 wx_add_thirdparty_library(wxUSE_LIBTIFF TIFF "use libtiff (TIFF file format)")
-wx_add_thirdparty_library(wxUSE_NANOSVG NanoSVG "use NanoSVG for rasterizing SVG" DEFAULT builtin)
+
+# wx_add_thirdparty_library(wxUSE_LUNASVG LunaSVG "use lunasvg for rasterizing SVG (C++17 required)")
```
PBfordev, line 115: Was the commented-out line kept by accident?

But I am not sure why some 3rd party libraries are added via `wx_add_thirdparty_library` and some by modifying `${wxTHIRD_PARTY_LIBRARIES}` ...

docs/doxygen/overviews/high_dpi.md
```diff
@@ -267,14 +267,16 @@ wxBitmapBundle::FromSVG() or wxBitmapBundle::FromSVGFile(). Such bitmap
...
 resolution. At normal DPI, i.e. without any scaling, the size of the resulting
 bitmap will be the default bundle size, which must be provided when creating
-this kind of bitmap bundle, as SVG image itself doesn't necessarily contain
+this kind of bitmap bundle, as an SVG image itself doesn't necessarily contain
 this information.
 
-Note that wxWidgets currently uses [NanoSVG][] library for SVG support and so
-doesn't support all SVG standard features and you may need to simplify or
-tweak the SVG files to make them appear correctly.
+Note that wxWidgets neither the default [NanoSVG][] library nor the newer
```
PBfordev, line 273: I don't think the wording is correct, I would probably just remove "that wxWidgets" part.

include/wx/setup_inc.h
```diff
@@ -1467,6 +1467,9 @@
 // Set to 1 for TIFF format support (requires libtiff)
 #define wxUSE_LIBTIFF       1
 
+// Set to 1 for SVG rasterizing support using LunaSVG (C++17 compiler required)
+#define wxUSE_LUNASVG       0
```
PBfordev, line 1471: How comes that NanoSVG is the default here as advertised in the docs, but LunaSVG seems to be the default in the actual `setup.h` files?

By default, it seems `OFF` in CMake and `1` in `config.{gcc|vc)`?

src/generic/bmpsvg.cpp
```diff
@@ -28,14 +55,173 @@
...
+}
+
+bool wxBitmapBundleImplSVG::IsOk() const
+{
+    return m_svgDocument != nullptr;
+}
+
+wxBitmap wxBitmapBundleImplSVG::DoRasterize(const wxSize& size)
+{
+    if (IsOk())
```
PBfordev, line 169:
```suggestion
    if ( IsOk() )
```

Per coding guidelines, the condition in parentheses should be "wrapped" in a space, applies to other cases here.

Also, I personally would just return here, instead of adding a huge indented block below.

interface/wx/bmpbndl.h
```diff
@@ -293,6 +296,19 @@ class wxBitmapBundle
     /// @overload
     static wxBitmapBundle FromSVG(const wxByte* data, size_t len, const wxSize& sizeDef);
 
+    /**
```
PBfordev, line 299: Is making the two rasterizer-specific functions public the best idea?

For example, we do not allow the NanoSVG rasterizer to be created with `NSVGimage`...

build/cmake/options.cmake
MaartenBent: `wx_add_thirdparty_library` adds an option with choices `off`, `sys` and `builtin`, so should only be used when the sources are distributed with wxWidgets. Otherwise you can use `wx_option` to add an option that can be turned on/off.

Some items are added manually to the `wxTHIRD_PARTY_LIBRARIES` variable, this is only for it to be listed in the CMake summary (`Which libraries should wxWidgets use?`), it doesn't set any options.

Randalphwa:
This is just a quick update so the PR doesn't get purged as inactive. The medical issues that limited my ability to work have been mostly resolved, and I am working regularly again. My own project is way behind, and I had hoped to tie it to the wxWidgets 3.3 release, so I'm currently focusing on that. Given how close it is to the wxWidgets 3.3 release, my inclination is to work on this _after_ the 3.3.0 release, and target it for a potential 3.3.1 release instead.

vadz:
Thanks for the update, even if I wasn't going to close this one!

And I'm very glad to hear good news about your health, this is far more important than any PRs.

Randalphwa:
Ack, looks like I pushed the branch on the wrong repository -- meant to be pushing an update on wxWidgets/lunasvg not the main wxWidgets branch.

In any event, it does rather highlight that this PR got too stale, and the merge conflicts I have with the main repository are intense -- on top of which lunasvg is now a single submodule instead of two. I've actually been working on a different branch where I've incorporated the code review comments from PBfordev and am currently verifying the latest lunasvg code. After that, I'll start from scratch updating the current build scripts rather than risk breaking something with the merge conflicts. Once that's done, my plan is to delete this PR entirely, and create a new one that's based on the current build script files.

PR #24476, opened 2024-04-16
Change theme when changing appearance in wxGTK if necessary

vadz:
At least under Ubuntu 22.04 (GNOME 42) resetting the dark preference is
not enough to switch to light appearance when the theme is set to the
default value of "Yaru-dark" in system-wide dark mode, so change the
theme to remove the explicit dark theme selector to make this work.

Note that it does not seem to be necessary to add "-dark" back when
switching from light mode to dark, i.e. using "prefer-dark-theme" to
choose the dark mode of the default theme seems to have exactly the same
effect as choosing the dark theme variant.

----

This couldn't be as simple as I thought, of course... More testing showed that changing appearance doesn't work with GNOME 42 under Ubuntu 22.04 (even though it did for me with GNOME 43 under Debian Bookworm) and I don't see any other way to make it work than by changing the theme too. Please let me know if you have any comments/better suggestions!

### DIFF
--- a/src/gtk/settings.cpp
+++ b/src/gtk/settings.cpp
@@ -269,14 +269,21 @@ void DoUpdateColorScheme(wxGTKImpl::ColorScheme colorScheme)
         return;
     }
 
-    // We don't need to enable prefer-dark if the theme is already dark
-    if (strstr(themeName, "-dark") || strstr(themeName, "-Dark"))
-    {
-        wxLogTrace(TRACE_DARKMODE, "Force dark mode for theme \"%s\"", themeName);
-        preferDarkPrev = TRUE;
-    }
+    wxLogTrace(TRACE_DARKMODE, "Current GTK theme is \"%s\"", themeName);
 
+    const wxString theme = wxString::FromUTF8(themeName);
     g_free(themeName);
+
+    // Check if the current theme is a dark variant.
+    constexpr const char* darkVariant = "-dark";
+    constexpr const char* darkVariantU = "-Dark";
+    constexpr size_t lenDark = 5; // strlen(darkVariant) == strlen(darkVariantU)
+    auto posDark = theme.find(darkVariant);
+    if ( posDark == wxString::npos )
+        posDark = theme.find(darkVariantU);
+
+    if ( posDark != wxString::npos )
+        preferDarkPrev = TRUE;
 
     gboolean preferDark = FALSE;
     switch ( colorScheme )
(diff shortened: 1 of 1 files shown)


paulcor:
Changing the theme name this way (or at all really) does not seem like a good idea. First of all, there are few if any theme names ending `-light`. And the "dark" part may not be a suffix, for example on my system there is `Mint-Y-Dark-Aqua` amongst others. I'm not aware of any easy way to know whether a theme name is valid. And if you set an invalid name, you're going to get `Adwaita`, which might not be anything like what the DE is using.

I'm becoming more and more uncomfortable with the whole idea of trying to force the dark setting. The rationale for doing so is dubious at best.

vadz:
> Changing the theme name this way (or at all really) does not seem like a good idea. First of all, there are few if any theme names ending `-light`.

I guess we could just remove the "-dark" suffix too.

> And the "dark" part may not be a suffix, for example on my system there is `Mint-Y-Dark-Aqua` amongst others.

We could also search for any "-dark" substring. But note that the assumption that "-dark" could only be a suffix was there from the beginning, i.e. this is not really specific to this PR.

> I'm not aware of any easy way to know whether a theme name is valid.

Unfortunately me neither.

> And if you set an invalid name, you're going to get `Adwaita`, which might not be anything like what the DE is using.

I'd be fine with documenting that calling this function under GTK could change the current theme. Arguably, showing an application that really wants to use dark mode using dark Adwaita is still better than showing it in the default light theme.

> I'm becoming more and more uncomfortable with the whole idea of trying to force the dark setting. The rationale for doing so is dubious at best.

It's been asked for several times, including for the application I'm working on right now, so there is clearly some request for it. I really don't see any harm in doing this on the best effort basis knowing that people who object to it can just not call this function in the first place.

paulcor:
> But note that the assumption that "-dark" could only be a suffix was there from the beginning, i.e. this is not really specific to this PR.

The current code **does** search for a substring, and so does **not** assume "-dark" can only be a suffix. That assumption is very much specific to the changes in this PR.

> Arguably, showing an application that really wants to use dark mode using dark Adwaita is still better than showing it in the default light theme.

But it **won't be** dark Adwaita, the fallback is just regular Adwaita, not dark.

paulcor:
> what do you think of the latest version?

I think it will work.

> AFAICS there is no dark Adwaita variant (any more), just a dark variant of the same theme. See e.g. GTK Inspector [Adwaita page](https://developer.gnome.org/documentation/tools/inspector.html#adwaita).

Note that that documentation is for GTK4. I'm not sure what you think the difference is between "dark Adwaita variant" and "dark variant of... Adwaita", but there is definitely a (built in) `Adwaita-dark` theme in GTK3 (and a place holder for it in `/usr/share/themes`). A dark "variant" is always a separate theme in GTK3.

paulcor:
> Then GTK3 must be selecting "Adwaita-dark" automatically

Looking into it more carefully, this isn't technically what happens, but it doesn't matter, as you are correct that `Adwaita-dark` is the same thing as `Adwaita` with `gtk-application-prefer-dark-theme==true`. And it seems reasonable to assume the same for other themes.

Issue #23489, opened 2023-04-24
WEBP support for wxImage


beENABLED:
Would there by interest for an image handler for webp?

I have written the actual image handler based on Googles webp library, but i have no idea how to integrate the webp library into the wx build process.

vadz:
Sure, it would be welcome, this is a reasonably popular format (much more so than some of the other ones we support...) and it would be nice to have it.

For libwebp, we have 2 choices:

1. Handle it as an external library, i.e. rely on it being available somewhere on the system. This is perfectly reasonable under Linux, but can also be done under MSW, see e.g. LZMA library example (notably the [installation instructions](https://docs.wxwidgets.org/3.2/page_build_liblzma.html)).
2. Include it as a submodule in wx itself. It's probably not worth doing it, at least not initially, but if you can integrate it with wx in the same way as libpng or libjpeg are, it could be considered as well.

The main drawback of (1) is that webp support would have to be disabled under MSW by default.

vadz:
For configure it should be as simple as adding `PKG_CHECK_MODULES(LIBWEBP)` similar to any other existing check (e.g. `LIBCURL` one). For CMake it would involve using a `FindWebp.cmake` but I don't know much about it myself.

hoehermann:
For reference, I did something similar [here](https://github.com/hoehermann/wxWEBPHandler). I have no idea if I got the code-style right. Support for writing an image with alpha channel is missing, but something which [people want](https://forums.wxwidgets.org/viewtopic.php?f=20&t=39212&p=206253). 
Additionally I probably need to create more tests. Also find a way to build and run on MacOS which I do not have a license to.

vadz:
@hoehermann Nice! If you could please make a PR out of this, we'd be happy to include this in wxWidgets.

hoehermann:
I added support for saving an image with alpha channel and wrote a bunch of tests.

[...]

PR #24479, opened 2024-04-16
Add support for WebP image file format via libwebp.

hoehermann:
This is a pull request for #23489. Please refer to the conceptual discussion there. Discuss suggestions regarding this particular pull-request here.

_Update:_ I consider this ready for merging.


Randalphwa:
Because this requires a library that is _not_ built by wxWidgets itself, I'm assuming that anyone wanting to use this functionality will not be able to use any of the pre-built binaries? Or are you envisioning that all the pre-built binaries should have webp support turned on by default? If the pre-built binaries aren't going to have webp support, then it should be documented that the user must build wxWidgets from source to get this functionality.

I'm not familiar with the Chromium repository system, but I do note that the issue you refered to ( https://chromium-review.googlesource.com/c/webm/libwebp/+/4868215 ) states that the change was merged on 9/19/2023 -- so wouldn't the fix be in the current 1.4 release (released 4/12/2024)?

hoehermann:
I answered to Randalphwa's question at https://github.com/wxWidgets/wxWidgets/issues/23489#issuecomment-2080917306 to keep the discussion in one place.

hoehermann:
@vadz Is this okay or do I need to improve something?

vadz:
Sorry, I've completely missed that it was ready, thanks for the ping. I'll try to look at it but, due personal circumstances, it might take some time.

Any reviews/tests by others would be very appreciated!

build/cmake/init.cmake
```diff
@@ -348,6 +348,14 @@ if(wxUSE_LIBLZMA)
     endif()
 endif()
 
+if(wxUSE_LIBWEBP)
+    find_package(WebP) # TODO: select specific components as soon as upstream-supplied WebPConfig.cmake supports it
```
vadz, line 352: Very minor question, and probably more to @MaartenBent, but do we want to use mixed-case for this one or `WEBP` for consistency with all the other 3rd party libraries?

configure.ac
```diff
@@ -7714,6 +7747,10 @@ if test "$wxUSE_LIBLZMA" = "yes"; then
         WXCONFIG_LIBS="$LZMA_LINK $WXCONFIG_LIBS"
     fi
 fi
+if test "$wxUSE_LIBWEBP" = "yes"; then
+    dnl no need to check for duplicates since webp is disabled in the our builtin tiff
```
vadz, line 7751: Nitpick of the day
```suggestion
    dnl no need to check for duplicates since webp is disabled in our builtin tiff
```

interface/wx/gdicmn.h
```diff
@@ -46,6 +46,7 @@ enum wxBitmapType
     wxBITMAP_TYPE_TGA,
     wxBITMAP_TYPE_MACCURSOR,
     wxBITMAP_TYPE_MACCURSOR_RESOURCE,
+    wxBITMAP_TYPE_WEBP,
```
vadz, line 49:
```suggestion
    wxBITMAP_TYPE_WEBP, ///< Available since wxWidgets 3.3.0.
```

interface/wx/imagwebp.h
```diff
@@ -0,0 +1,34 @@
...
+    This is the image handler for the WebP format.
+
+    See @ref page_build_libwebp for information about libwebp, required in
+    order to use this class.
+
+    @library{wxcore}
+    @category{gdi}
+
+    @see wxImage, wxImageHandler, wxInitAllImageHandlers()
+*/
```
vadz, line 20:
```suggestion
    @since 3.3.0
*/
```

misc/scripts/check_mixed_eol.sh
```diff
@@ -16,7 +16,7 @@ while IFS= read -r -d '' file; do
 
     # skip binary files
     case "$file" in
-    *.ani | *.bmp | *.chm | *.cur | *.dia | *.gif | *.gz | *.hlp | *.icns | *.ico | *.jpg | *.mo | *.mpg | *.pcx | *.pdf | *.png | *.pnm | *.pyc | *.tga | *.tif | *.ttf | *.wav | *.zip )
+    *.ani | *.bmp | *.chm | *.cur | *.dia | *.gif | *.gz | *.hlp | *.icns | *.ico | *.jpg | *.mo | *.mpg | *.pcx | *.pdf | *.png | *.pnm | *.pyc | *.tga | *.tif | *.ttf | *.webp | *.wav | *.zip )
```
vadz, line 19: Oops, another nitpick (they're alphabetically sorted):
```suggestion
    *.ani | *.bmp | *.chm | *.cur | *.dia | *.gif | *.gz | *.hlp | *.icns | *.ico | *.jpg | *.mo | *.mpg | *.pcx | *.pdf | *.png | *.pnm | *.pyc | *.tga | *.tif | *.ttf | *.wav | *.webp| *.zip )
```

src/common/imagwebp.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+#if wxUSE_STREAMS
+
+#include <wx/mstream.h>
+#include <memory>
+#include <functional>
+
+typedef std::unique_ptr<WebPDemuxer, std::function<void(WebPDemuxer*)>> WebPDemuxerPtr;
+
+bool DecodeWebPDataIntoImage(wxImage *image, WebPData *webp_data, bool verbose) {
```
vadz, line 42: And now a formatting nitpick:
```suggestion
bool DecodeWebPDataIntoImage(wxImage *image, WebPData *webp_data, bool verbose)
{
```

src/common/imagwebp.cpp
```diff
@@ -0,0 +1,231 @@
...
+//-----------------------------------------------------------------------------
+
+wxIMPLEMENT_DYNAMIC_CLASS(wxWEBPHandler, wxImageHandler);
+
+#if wxUSE_STREAMS
+
+#include <wx/mstream.h>
+#include <memory>
+#include <functional>
+
```
vadz, line 39: Everything here looks private, so

```suggestion

namespace {
```

src/common/imagwebp.cpp
```diff
@@ -0,0 +1,231 @@
...
+           wxLogError("WebP: GetFeatures not OK.");
+        }
+        return false;
+    }
+    image->Create(features.width, features.height, false); // this allocates memory
+    if (features.has_alpha)
+    {
+        // image has alpha channel. needs to be decoded, then re-ordered.
+        uint8_t * rgba = WebPDecodeRGBA(webp_data->bytes, webp_data->size, &features.width, &features.height);
+        if (nullptr == rgba)
```
vadz, line 58: We generally avoid Yoda comparisons in wx, so it should be either `rgba == nullptr` or just `!rgba`, as you prefer.

src/common/imagwebp.cpp
```diff
@@ -0,0 +1,231 @@
...
+               wxLogError("WebP: WebPDecodeRGBInto failed.");
+            }
+            return false;
+        }
+    }
+    image->SetMask(false); // all examples do this, so I do so as well
+    return true;
+}
+
+bool DecodeWebPFrameIntoImage(wxImage *image, int index, WebPDemuxerPtr & demuxer, bool verbose)
```
vadz, line 98: It's unusual and almost always unnecessary to pass (smart) pointers by reference, why not just pass `WebPDemuxer&` directly here?

src/common/imagwebp.cpp
```diff
@@ -0,0 +1,231 @@
...
+        delete webp_data;
+        delete mos;
+        if (verbose)
+        {
+            wxLogError("WebP: WebPDemux failed.");
+        }
+    }
+    return demux;
+}
+
```
vadz, line 151:
```suggestion
} // anonymous namespace

```

src/common/imagwebp.cpp
```diff
@@ -0,0 +1,231 @@
...
+        {
+           wxLogError("WebP: GetFeatures not OK.");
+        }
+        return false;
+    }
+    image->Create(features.width, features.height, false); // this allocates memory
+    if (features.has_alpha)
+    {
+        // image has alpha channel. needs to be decoded, then re-ordered.
+        uint8_t * rgba = WebPDecodeRGBA(webp_data->bytes, webp_data->size, &features.width, &features.height);
```
vadz, line 57: Ideally this would be a `unique_ptr<uint8_t, WebPFree>`.

src/common/imagwebp.cpp
```diff
@@ -0,0 +1,231 @@
...
+    return ok;
+}
+
+WebPDemuxerPtr CreateDemuxer(wxInputStream& stream, bool verbose = false) {
+    wxMemoryOutputStream * mos = new wxMemoryOutputStream;
+    stream.Read(*mos); // this reads the entire file into memory
+    // TODO: Only read data as needed. WebPDemux can operate on partial data.
+    //       Could save some bandwidth with e.g. DoGetImageCount
+    wxStreamBuffer * mosb = mos->GetOutputStreamBuffer();
+    WebPData * webp_data = new WebPData;
```
vadz, line 123: Does this have to live for as long as `WebPDemuxer` itself lives or can it be destroyed immediately after calling `WebPDemux()` (as I suspect)?

In the latter case, we don't need to create it (nor `mos`) on the heap, which avoids the need to `delete` it too. In the former, I'd strongly prefer to keep it in a single object with `wxMemoryOutputStream` and `WebPDemuxer*` itself rather than hiding them, which are logically part of the same object, inside a custom deleter.

src/common/imagwebp.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+wxIMPLEMENT_DYNAMIC_CLASS(wxWEBPHandler, wxImageHandler);
+
+#if wxUSE_STREAMS
+
+#include <wx/mstream.h>
+#include <memory>
+#include <functional>
+
+typedef std::unique_ptr<WebPDemuxer, std::function<void(WebPDemuxer*)>> WebPDemuxerPtr;
```
vadz, line 40: After looking at the code using this, I think it would be much more clear to have a wrapper class rather than using this unique pointer with a very custom deleter, i.e.

```
class wxWebPDemuxer
{
    WebPDemuxer* m_demuxer = nullptr;
    wxMemoryOutputStream m_mos;
    WebPData m_data;
};
```

Or, if possible, make the last 2 fields local variables in the function using them and then just use `unique_ptr<WebPDemuxer, WebPDemuxDelete>`.

src/common/imagwebp.cpp
```diff
@@ -0,0 +1,231 @@
...
+    const std::string webp = "WEBP";
+    const int buffer_size = 12;
+    char buffer[buffer_size];
+    // it's ok to modify the stream position here
+    stream.Read(buffer, buffer_size);
+    if (stream.LastRead() != buffer_size)
+    {
+        return false;
+    }
+    return std::string(buffer, 4) == riff && std::string(&buffer[8], 4) == webp;
```
vadz, line 226: This is probably not a huge optimization but still, why not just do

```suggestion
    return memcmp(buffer, "RIFF", 4) == 0 && memcmp(buffer + 8, "WEBP", 4) == 0;
```

instead of creating these temporary strings?

tests/image/image.cpp
```diff
@@ -839,9 +844,18 @@ TEST_CASE_METHOD(ImageHandlersInit, "wxImage::CompareLoadedImage", "[image]")
 
 
         wxINFO_FMT("Compare test '%s' for loading", g_testfiles[i].file);
-        CHECK_THAT( actual,
-                    RGBSameAs(g_testfiles[i].bitDepth == 8 ? expected8
-                                                           : expected24) );
+        wxImage & expected = g_testfiles[i].bitDepth == 8 ? expected8 : expected24;
+        int tolerance = 0;
+        switch (g_testfiles[i].type) {
```
vadz, line 849:
```suggestion
        switch (g_testfiles[i].type)
        {
```

vadz (commented):
Thanks a lot once again for your work, it's great to have support for WebP and globally it looks pretty good.

I would like to simplify the code creating `WebPDemux` as explained, if possible, please let me know if you can do it or if I should do it myself.

But an even more important consideration is to ensure that this code is actually getting tested, i.e. we need to modify the CI jobs to install libwebp-dev etc, to at least test it under Unix. If we can test it under Windows it would be great too, of course, but I guess this might be more complicated — but let me know if you see a way to do it.

Sorry again for the delay!

build/cmake/init.cmake
MaartenBent: This depends on the package you want to find. It should match the name of the `FindLibraryName.cmake` or `LibraryNameConfig.cmake` file.

For example, the CMake provided modules are `FindJPEG.cmake` and `FindPNG.cmake` and `FindOpenGL.cmake`, so it should match that name.
I think `Webkit` provides `FindWebP.cmake` so the current package name is correct.

If it does not match, it will show a warning:
```
  The package name passed to `find_package_handle_standard_args` (OpenGL)
  does not match the name of the calling package (OPENGL).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
```

vadz: Sorry, I was somehow under (wrong) impression that this matching was case-insensitive, but if it's case-sensitive then there is no question about what to use here, sorry again for the misunderstanding.

src/common/imagwebp.cpp
hoehermann: > Does this have to live for as long as WebPDemuxer itself lives

I assume, with "this" you refer to the `wxMemoryOutputStream` `mos` is pointing to. As far as I understood, `WebPDemux` does _not_ create a copy of the input data in all cases. At least that is what I make of the implementation of [WebPDemuxInternal](https://chromium.googlesource.com/webm/libwebp/+/refs/tags/v1.4.0/src/demux/demux.c#705). Unfortunately, the documentation is not terribly clear about the expected life-time of the buffers.

To keep the implementation simple for now, I went with the "read entire file into memory" approach. I hope that is okay. Is`wxInputStream::read()` into a `wxMemoryOutputStream` the way it is done in wxWidgets?

> I'd strongly prefer to keep it in a single object

I can do that.

src/common/imagwebp.cpp
hoehermann: All I can say that I seem to suffer from C++ poisoning and am usually unaware that `memcmp` even exists. 😅

hoehermann:
> Thanks a lot once again for your work, it's great to have support for WebP and globally it looks pretty good.

Thank you. ☺

> But an even more important consideration is to ensure that this code is actually getting tested, i.e. we need to modify the CI jobs to install libwebp-dev etc, to at least test it under Unix. If we can test it under Windows it would be great too, of course, but I guess this might be more complicated — but let me know if you see a way to do it.

Does the CI on Windows already employ vcpkg by any chance? I think just executed `vcpkg.exe install --triplet x64-windows-static libwebp`, CMake picked it up automagically and all was fine.

> Sorry again for the delay!

No problem, however…

> I would like to simplify the code creating `WebPDemux` as explained, if possible, please let me know if you can do it or if I should do it myself.

…right now, things are rather stressful on my end. I can do some more work here, but not before November 2024, I am afraid.

src/common/imagwebp.cpp
vadz: I meant both `WebPData` itself and the buffer associated with it.

Looking at the source (thanks for the link), it seems like `data` is only used inside this function, as it's used to initialize the local `MemBuffer` variable and so can't be accessed once the function returns. So it really seems like we don't need these heap allocations (and corresponding `delete`s) and can just make them local variables to simplify things.

vadz:
> Does the CI on Windows already employ vcpkg by any chance?

I'm afraid it doesn't. It could be a good idea adding a build installing the external libraries using vcpkg, but for now I'd settle for just installing the WebP libraries from the distribution repositories under Linux — this should be trivial and, while it wouldn't test Windows at all, still much better than nothing.

> > I would like to simplify the code creating `WebPDemux` as explained, if possible, please let me know if you can do it or if I should do it myself.
> 
> …right now, things are rather stressful on my end.

Sorry to hear this!

> I can do some more work here, but not before November 2024, I am afraid.

There is no particular urgency with this, so if you think you'll have time to return to this later, I'd rather leave it to you. But I'd like to release 3.3.0 before the end of the year (even if I'm far from sure that we'll manage to do it), and it would be nice to have WebP support in it.

src/common/imagwebp.cpp
hoehermann: I now remember (sorry, it has been almost half a year, the details are a bit hazy) that I need the `WebPDemux` in `DoGetImageCount`, too. I did not want redundant code for creating a `WebPDemux` struct from a `wxInputStream`. I do not recall whether `DoGetImageCount` works for me or not, but I did not take the time to create any tests. My comment indicates that is the only reason why `DoGetImageCount` is commented out.

As far as I understand, `data` – or rather the buffer associated with it – is also used later on: The demuxer emits `WebPData` structs pointing to different sections within `data`. Instances include the call to `WebPGetFeatures` for getting the image dimensions. Also the actual decompression with `WebPDecodeRGB…`.

Indeed, it looks like we do not need the heap-allocated `WebPData` from line 123 after `WebPDemux()`. However, I am pretty confident that `data` (aka `wxMemoryOutputStream mos`) must remain valid during `WebPDemux`'s lifetime. I should verify that when touching the code again.

MaartenBent:
> I'm afraid it doesn't. It could be a good idea adding a build installing the external libraries using vcpkg, but for now I'd settle for just installing the WebP libraries from the distribution repositories under Linux — this should be trivial and, while it wouldn't test Windows at all, still much better than nothing.

It might work in the [msys2/mingw64 build](https://github.com/wxWidgets/wxWidgets/blob/master/.github/workflows/ci_msw.yml#L134), when adding the [mingw-w64-x86_64-libwebp](https://packages.msys2.org/packages/mingw-w64-x86_64-libwebp) package to the install section. When I check the contents of this package it seems to install the required webp cmake files.

PR #24482, opened 2024-04-17
Switch to using TTM_RELAY for tooltips in wxMSW to simplify the code and avoid problems with MFC

wsu-cb:
I have discovered that my previous workaround for #23574 causes some wx tooltips not to be displayed in my full application.  Unfortunately, I can't find a small sample that shows the problem.

This PR adds an ugly hack to get the wx tooltips to work in my application, and doesn't obviously make things worse for anything else.  However, since this is ugly, and I can't find a simple way to demonstrate the problem, I will understand if this PR is rejected.

### DIFF
--- a/include/wx/msw/statusbar.h
+++ b/include/wx/msw/statusbar.h
@@ -66,7 +66,6 @@ protected:
     virtual wxSize DoGetBestSize() const override;
     virtual void DoMoveWindow(int x, int y, int width, int height) override;
 #if wxUSE_TOOLTIPS
-    virtual bool MSWProcessMessage(WXMSG* pMsg) override;
     virtual bool MSWOnNotify(int idCtrl, WXLPARAM lParam, WXLPARAM* result) override;
 #endif
 
--- a/include/wx/msw/tooltip.h
+++ b/include/wx/msw/tooltip.h
@@ -54,9 +54,6 @@ public:
     // implementation only from now on
     // -------------------------------
 
-    // should be called in response to WM_MOUSEMOVE
-    static void RelayEvent(WXMSG *msg);
-
     // add a window to the tooltip control
     void AddOtherWindow(WXHWND hwnd);
 
--- a/src/msw/evtloop.cpp
+++ b/src/msw/evtloop.cpp
@@ -78,18 +78,6 @@ bool wxGUIEventLoop::PreProcessMessage(WXMSG *msg)
         return true;
     }
 
-#if wxUSE_TOOLTIPS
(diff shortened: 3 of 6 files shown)


vadz:
We also have a [similar workaround here](https://github.com/wxWidgets/wxWidgets/blob/c53406d6c556692a2796085de898ccb01fa3649a/src/msw/window.cpp#L2787-L2795), I wonder if we should do this in some other way to ensure that it always happens no matter what...

wsu-cb:
> We also have a [similar workaround here](https://github.com/wxWidgets/wxWidgets/blob/c53406d6c556692a2796085de898ccb01fa3649a/src/msw/window.cpp#L2787-L2795), I wonder if we should do this in some other way to ensure that it always happens no matter what...

Are you thinking of something like [this](https://learn.microsoft.com/en-us/windows/win32/winmsg/mouseproc)?  I have no experience with hooks, but I am willing to make an attempt using that technique.

wsu-cb:
Wait a minute:  Wouldn't using `TTF_SUBCLASS` avoid the need for these workarounds (by avoiding the need for `TTM_RELAYEVENT`)?  Why doesn't wxMSW use `TTF_SUBCLASS` instead of requiring `TTM_RELAYEVENT`?

vadz:
> Wait a minute: Wouldn't using `TTF_SUBCLASS` avoid the need for these workarounds (by avoiding the need for `TTM_RELAYEVENT`)? Why doesn't wxMSW use `TTF_SUBCLASS` instead of requiring `TTM_RELAYEVENT`?

I don't remember. I _think_ there was some problem with using it, but this was back in Windows XP (if not earlier!) days, so whatever it was, it might not exist any more. If you'd like to try to change the code to use `TTF_SUBCLASS`, and if it works, I think we can try making this change.

wsu-cb:
I have tried using TTF_SUBCLASS instead of TTM_RELAYMESSAGE, and it fixes the problem I was having, simplifies the code overall, and, as far as I can tell, does not create any new problems.

vadz:
Could you please make a PR with your changes? TIA!

wsu-cb:
> Could you please make a PR with your changes? TIA!

The force-push https://github.com/wxWidgets/wxWidgets/commit/f2862c0c09ea5a76dd27348a3586f8a82ed33d66 is the modified code.

vadz:
Oops, sorry, I hadn't realized this PR was changed.

It looks fine to me and if it works, I'm for merging it. Any objections from anybody?

PR #24484, opened 2024-04-17
[wxQt] Make native wxGraphicsContext implementation available under non wxMSW too

AliKet:


AliKet:
@vadz I suspect that `wxUSE_GRAPHICS_CONTEXT` was set to **0** (and I don't know why if this is the case) when running the tests in CI. Notice that none of the tests defined here https://github.com/wxWidgets/wxWidgets/blob/ee309e078a2294d59c754b79c0407184e7558738/tests/graphics/clippingbox.cpp#L2240-L2242 has run ?

TIA for any hint!

AliKet:
@dsa-t it would be great if you could test this PR and report any issues here before merging, TIA for your help on this!

AliKet:
@vadz  As I suspected `wxUSE_GRAPHICS_CONTEXT` is defined **0** !! and I don't know why ?
This temporary commit [d4b413d](https://github.com/wxWidgets/wxWidgets/pull/24484/commits/d4b413d735fbbb3139fc0dc6d2122cea5847388a) prove it, see
 https://github.com/wxWidgets/wxWidgets/actions/runs/8741380641/job/23987274557?pr=24484

FYI everything works correctly for me locally without any special setup.

vadz:
This is due to the following lines in configure output:

```
checking for CAIRO... no
configure: WARNING: wxGraphicsContext won't be available
```

Which is in turn due to this

https://github.com/wxWidgets/wxWidgets/blob/ee309e078a2294d59c754b79c0407184e7558738/configure.ac#L534-L536

If you just remove the test for Qt there, it should work, I think.

AliKet:
> This is due to the following lines in configure output:
> 
> ```
> checking for CAIRO... no
> configure: WARNING: wxGraphicsContext won't be available
> ```
> 
> Which is in turn due to this
> 
> https://github.com/wxWidgets/wxWidgets/blob/ee309e078a2294d59c754b79c0407184e7558738/configure.ac#L534-L536
> 
> If you just remove the test for Qt there, it should work, I think.

Thanks a lot. I'll push the change and see how it goes...

AliKet:
@vadz This is ready for review, TIA!

src/qt/graphics.cpp
```diff
@@ -169,29 +140,123 @@ class WXDLLIMPEXP_CORE wxQtPenData : public wxGraphicsObjectRefData
...
     {
-        return m_pixmap;
+        return const_cast<wxQtBitmapData*>(this)->m_pixmap;
     }
 
     virtual void* GetNativeBitmap() const override
     {
-        return m_pixmap;
+        auto This = const_cast<wxQtBitmapData*>(this);
+        return &This->m_pixmap;
```
vadz, line 256: Why not

```suggestion
        return const_cast<QPixmap*>(&m_pixmap);
```

which would be more clear IMO?

If not, I'd like to at least rename `This` to `self` for consistency with the same casts done elsewhere, but this I can do myself.

vadz (approved):
Thanks, no real comments but then I don't know Qt API so I can't meaningfully review this anyhow. But as long as the tests pass, this should be good, thanks again for working on this!

dsa-t:
Works as expected in my tests, thanks for implementing this!

PR #24487, opened 2024-04-18
Fix crash on spin page of widgets sample

PBfordev:
Attempting to change the style or alignment of the spin controls in the widgets sample made the sample crash. This was caused by a8b4753 (Reconnect events to the widget after recreating it in the sample, 2024-02-19) which resulted in using windows that were previously destroyed.

Fix this by calling NotifyWidgetRecreation() for each spin control only after they were all recreated.


vadz:
Thanks for fixing this! I will merge it soon.

PR #24488, opened 2024-04-18
Improve wxSpinCtrlDouble docs

PBfordev:
Document all available styles, i.e., just like with wxSpinCtrl, list also wxTE_PROCESS_ENTER and the three text alignment options.

This should be backported to 3.2.


vadz:
Thanks!

PR #24489, opened 2024-04-18
Fix wxTE_PROCESS_ENTER in wxGTK comboboxes with autocomplete

### LINKED ISSUE #24394: In wxGTK, wxTextEntry autocomplete interferes with the wxTE_PROCESS_ENTER style for comboboxes
### Description

In wxGTK, wxTextEntry autocomplete interferes with the wxTE_PROCESS_ENTER style for comboboxes.  If autocomplete is active for a combo box, then the wxTE_PROCESS_ENTER style will be cleared the first time the autocomplete dropdown is shown.

Note that this does not happen with wxTextCtrl.

The issue appears to be that `wxTextAutoCompleteData::ToggleProcessEnter()` cannot be safely called twice in a row with `toggleOff` set to true.  The second time it's called, `m_hadProcessEnterFlag` will be unconditionally set to "false" regardless of what it was originally.

In my testing, `wxTextAutoCompleteData::ToggleProcessEnter()` is called twice in a row with `toggleOff` set to true when showing the autocomplete dropdown for a combobox, and twice with it set to false when hiding it.  For a text control, it's called once with it set to true, and once with it set to false. 

1. Run the sample program.
2. Type the letter "a" in the combo box.
3. Hit "Enter".

Expected result: a dialog box pops up.

Actual result: nothing happens.

[combotest.cpp.gz](https://github.com/wxWidgets/wxWidgets/files/14551973/combotest.cpp.gz)

- wxWidgets version you use: Git head ( @66ea0b2 )

- wxWidgets port you use: wxGTK
- OS and its version: Gentoo Linux
  + GTK version: 3.24.39
  + Which GDK backend is used: X11
  + Desktop environment: XFCE

vadz:
This style was reset and never re-enabled because the comboboxes get multiple grab-notify notifications, unlike text controls, which only get one.

Fix this by adding a counter of the number of times grab-notify was called indicating that the popup was shown or hidden and only update the style when its status actually effectively changes.

Also use a more clear name for the function called when the popup is shown as having a Toggle(bool toggle) function was a bit confusing.

Closes #24394.

----

@AliKet Please let me know if you have any comments as this code was originally written by you.

### DIFF
--- a/src/gtk/textentry.cpp
+++ b/src/gtk/textentry.cpp
@@ -313,39 +313,41 @@ public:
     // for wxTextAutoCompleteFixed.
     virtual bool ChangeCompleter(wxTextCompleter* completer) = 0;
 
-    // We should toggle off wxTE_PROCESS_ENTER flag of our wxTextEntry while
+    // We need to turn off wxTE_PROCESS_ENTER flag of our wxTextEntry while
     // the completion popup is shown to let it see Enter event and process it
     // on its own (e.g. to dismiss itself). This is done by "grab-notify" signal
     // see wxTextCtrl::OnChar()
-    void ToggleProcessEnterFlag(bool toggleOff)
+    void GTKOnPopupShown(bool shown)
     {
         wxWindow* const win = GetEditableWindow(m_entry);
 
-        long flags = win->GetWindowStyleFlag();
-        if ( toggleOff )
+        if ( shown )
         {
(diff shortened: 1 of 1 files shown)


AliKet:
I didn't test this but the changes looks good to me, thanks.

Carnildo:
Works in my testing: fixes comboboxes, and doesn't break text boxes.

PR #24490, opened 2024-04-19
Don't make black transparent when drawing on wxOverlay under MSW

vadz:
Using black as the transparent colour is not a great idea as everything drawn using it on the overlay just doesn't appear at all then.

Use a different colour that will hopefully be much less common.

---

@AliKet please let me know if you have any comments.

### DIFF
--- a/include/wx/generic/private/drawresize.h
+++ b/include/wx/generic/private/drawresize.h
@@ -1,0 +1,43 @@
+///////////////////////////////////////////////////////////////////////////////
+// Name:        wx/generic/private/drawresize.h
+// Purpose:     Helper function for drawing "resize hint".
+// Author:      Vadim Zeitlin
+// Created:     2024-04-20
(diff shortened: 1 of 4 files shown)


AliKet:
Great, this is definitely the right thing to do and it should have been done this way from the beginning. 👍

AliKet:
While you're at it, these can be updated after merging this PR:
https://github.com/wxWidgets/wxWidgets/blob/ee309e078a2294d59c754b79c0407184e7558738/src/aui/framemanager.cpp#L92-L101 and this https://github.com/wxWidgets/wxWidgets/blob/ee309e078a2294d59c754b79c0407184e7558738/src/generic/splitter.cpp#L72-L81

i.e.:
```
static wxBitmap wxPaneCreateStippleBitmap()
{
    unsigned char data[] = { 0,0,0,192,192,192, 192,192,192,0,0,0 };
    wxImage img(2,2,data,true);
    return wxBitmap(img);
}
```

PR #24497, opened 2024-04-21
wx-3.2.x MinGW build tools update

XPsoud:
Update available compilers version:
- MSYS2 GCC-13.2.0 is the last one available 
- MSYS2 GCC-12.2.0 still supported for compatibility with previous 3.2.x version

`/docs/licence.txt` will be added to each created archive file


PBfordev:
Just curious: Do we really need to offer MSYS2 binaries?

 I don't know much about MSYS2 but isn't its purpose to provide the prebuilt libraries and they seem to do that for wxWidgets (before 3.2, they used to have not only the latest 3.0.x version but also the latest 3.1.x release). 

wxWidgets MSYS2 MinGW packages are available for GCC (both MSVCRT and UCRT builds) as well as CLang as x86, x86_x64 (and aarch64 for clang).

build/tools/mingw/parameters.bat
```diff
@@ -14,10 +14,13 @@ if "%2" == "" goto :NOPARAMS
 
 rem Define compiler's bin directory
 if "%1" == "1220" (
-    set COMPBINDIR=G:\msys64\mingw%2\bin
+    set COMPBINDIR=G:\msys64-gcc1220\mingw%2\bin
 ) else (
     set COMPBINDIR=G:\MinGW%1-%2\bin
 )
+if "%1" == "1320" (
```
vadz, line 21: FWIW I think you can use `else if` in BAT files, so this could be done above instead of overwriting the value set in the line 19.

vadz (commented):
Thanks for the update, should this be backported to 3.2 too for the upcoming 3.2.5 release?

vadz:
Sorry, I've realized that this PR was for 3.2 already, so the question above didn't make sense.

I've merged it now, thanks!

PR #24499, opened 2024-04-22
Fix for bind with rtti disabled (fixes #14850)

### LINKED ISSUE #14850: wxEvtHandler::Bind with lambdas without RTTI fail to compile
*Issue migrated from trac ticket # 14850*

**component:** GUI-all | **priority:** low | **keywords:** rtti

#### 2012-11-24 08:11:02: staticinline created the issue

___
Attempting to use pass a C++11 lambda function to wxEvtHandler::Bind can cause compilation errors when RTTI is disabled.  The offending code may take the following form:

```
obj->Bind(wxEVT_COMMAND_BUTTON_CLICKED, [ . . . ](wxCommandEvent &){ . . . });
```

and give the following error (under MSVC++2010):

```
wxwidgets\include\wx/event.h(532): error C2512: '`anonymous-namespace'::<lambda0>' :
no appropriate default constructor available
          wxwidgets\include\wx/event.h(532) : while compiling class template member
function 'wxEventFunctorFunctor<EventTag,Functor>::wxEventFunctorFunctor(void)'
          with
          [
              EventTag=wxEventTypeTag<wxCommandEvent>,
              Functor=`anonymous-namespace'::<lambda0>
          ]
          wxwidgets\include\wx/event.h(3230) : see reference to class template
instantiation 'wxEventFunctorFunctor<EventTag,Functor>' being compiled
          with
          [
              EventTag=wxEventTypeTag<wxCommandEvent>,
              Functor=`anonymous-namespace'::<lambda0>
          ]
          test.cpp(10) : see reference to function template instantiation 'void
wxEvtHandler::Bind<wxEventTypeTag<T>,`anonymous-namespace'::<lambda0>>(const
EventTag &,const Functor &,int,int,wxObject *)' being compiled
          with
          [
              T=wxCommandEvent,
              EventTag=wxEventTypeTag<wxCommandEvent>,
              Functor=`anonymous-namespace'::<lambda0>
          ]
```

Whether the error occurs or not depends what the lambda captures or, more specifically, whether the default constructor on the lambda type can be called.  Consider this:

```
auto F = []() { };
decltype(F) f;
```

That succeeds on MSVC++2010 but fails with the error "lambda closure type has a deleted default constructor" on g++4.6.3.  On MSVC++2010 it fails with the error "error C2512: '`anonymous-namespace'::<lambda0>' : no appropriate default constructor available" (same error as above) if I add a capture to the lambda.

Now, notice in event.h wxEventFunctorFunctor that it defines this:

```
    Functor m_handler;
. . .
// Provide a dummy default ctor for type info purposes
	wxEventFunctorFunctor() { }
    typedef wxEventFunctorFunctor<EventTag, Functor> thisClass;
    WX_DECLARE_TYPEINFO_INLINE(thisClass)
```

Here "Functor" binds to a lambda type, and the default constructor used merely by the TYPEINFO (which simulates RTTI) in turn tries to call the default constructor of m_handler, which doesn't exist, thereby giving the compiler error.

The problem was also reported in https://groups.google.com/forum/?fromgroups#!topic/wx-dev/BG2yEWA-anU .

There are workarounds like this where the lambda is wrapped in a std::function (which does have a default constructor):

```
obj->Bind(wxEVT_COMMAND_BUTTON_CLICKED,
  std::function<void (wxCommandEvent &)>([ . . . ](wxCommandEvent &){ . . . }));
```

but that becomes painful to type.

The original example works if I enable RTTI (which undefines wxNO_RTTI), but I'm trying to avoid that.  I also tried disabling RTTI (/GR-) and forcing wxNO_RTTI to stay undefined (in platform.h).  This causes wxWidgets to use the native C++ "typeid" functionality, which still partly works even when RTTI is disabled.  However, some operations with typeid will then give warnings like "warning C4541: 'typeid' used on polymorphic type 'X' with /GR-; unpredictable behavior may result".  In wxWidgets, the only place I get these C4541 warnings is in events.h inside the "IsMatching" functions (wxTypeID comparisons), which in turn is only used by wxEvtHandler::DoUnbind/Unbind/Disconnect.  (This seems consistent with the comments about RTTI and Unbind in http://docs.wxwidgets.org/trunk/overview_cpp_rtti_disabled.html .)

A more straightforward solution is just to store the lambda in a smart pointer:

```
--- event-original.h
+++ event.h
@@ -33,6 +33,8 @@
     #include "wx/meta/convertible.h"
 #endif

+#include "wx/scopedptr.h"
+
 // forward declarations
@@ -491,7 +493,7 @@
     typedef typename EventTag::EventClass EventArg;

     wxEventFunctorFunctor(const Functor& handler)
-        : m_handler(handler), m_handlerAddr(&handler)
+        : m_handler(new Functor(handler)), m_handlerAddr(&handler)
         { }

     virtual void operator()(wxEvtHandler *WXUNUSED(handler), wxEvent& event)
@@ -503,7 +505,7 @@
         // Unbind( wxEVT_XXX, &EventHandler::method );
         // or
         // Unbind( wxEVT_XXX, &EventHandler::method, NULL )
-        m_handler(static_cast<EventArg&>(event));
+        (*m_handler)(static_cast<EventArg&>(event));
     }

     virtual bool IsMatching(const wxEventFunctor &functor) const
@@ -523,7 +525,7 @@
 private:
     // Store a copy of the functor to prevent using/calling an already
     // destroyed instance:
-    Functor m_handler;
+    wxScopedPtr<Functor> m_handler;

     // Use the address of the original functor for comparison in IsMatching:
     const void *m_handlerAddr;
```

That's what I'm using right now.  The smart pointer is not necessary when RTTI is enabled though, so I don't know if you want to exactly do it that way.

Is it possible instead for the simulated RTTI code in typeinfo.h to eliminate the requirement of having a default constructor?  That may be ideal.

MapleLeaf-X:
This commit fixes the long standing bug report #14850 in a sane way without any additional allocations and it fully supports lambdas with disabled rtti.

### DIFF
--- a/include/wx/typeinfo.h
+++ b/include/wx/typeinfo.h
@@ -24,6 +24,16 @@
 
 #include "wx/defs.h"
 
+//
+// Left for compatibility reasons
+//
+
+#define _WX_DECLARE_TYPEINFO_CUSTOM(CLS, IDENTFUNC)
+#define WX_DECLARE_TYPEINFO_INLINE(CLS)
+#define WX_DECLARE_TYPEINFO(CLS)
+#define WX_DEFINE_TYPEINFO(CLS)
+#define WX_DECLARE_ABSTRACT_TYPEINFO(CLS)
+
 #ifndef wxNO_RTTI
 
 //
@@ -38,13 +48,7 @@
 #endif
 
 #include <typeinfo>
-#include <string.h>
-
-#define _WX_DECLARE_TYPEINFO_CUSTOM(CLS, IDENTFUNC)
-#define WX_DECLARE_TYPEINFO_INLINE(CLS)
-#define WX_DECLARE_TYPEINFO(CLS)
-#define WX_DEFINE_TYPEINFO(CLS)
-#define WX_DECLARE_ABSTRACT_TYPEINFO(CLS)
+#include <cstring>
 
 #if wxTRUST_CPP_RTTI
 
@@ -91,7 +95,7 @@ private:
 
 //
 // When C++ RTTI is not available, we will have to make the type comparison
-// using pointer to a dummy static member function. This will fail if
+// using pointer to a dummy static member variable. This will fail if
 // declared type is used across DLL boundaries, although using
 // WX_DECLARE_TYPEINFO() and WX_DEFINE_TYPEINFO() pair instead of
 // WX_DECLARE_TYPEINFO_INLINE() should fix this. However, that approach is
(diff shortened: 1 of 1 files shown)


MaartenBent:
The commit also changes the file permissions: `include/wx/typeinfo.h 100644 → 100755`

MapleLeaf-X:
Thanks for letting me know. I'm not actually sure how it even happened but I will definitely have to fix that. should I close this and recreate this pr once I have it fixed?

vadz:
Thanks for finally fixing this!

PR #24503, opened 2024-04-25
Fix for potential warnings (C4390 / -Wempty-body and C4706)

MapleLeaf-X:
A small fix for potential warnings C4390 / -Wempty-body and C4706.

for consistency with the rest of the code I've used actual types instead of auto.

PS: I ended up botching my fork so I had to recreate it.


vadz:
Thanks, will merge soon.

PR #24505, opened 2024-04-25
More clipboard fixes for wxGTK under Wayland

vadz:
The important commit here is the last one, the other ones are just cleanups.


tobiolo:
Thanks for the fix! :+1: I tested it under Wayland and it works as expected.

PR #24508, opened 2024-04-26
Reworked RTTI in typeinfo.h and fixed potential problems with dll boundaries

MapleLeaf-X:
Previously I've overlooked RTTI's safety across dll boundaries so this pr resolve such potential problem. a few things to note, the code has been reworked and a few internal macros have been removed as they're no longer needed.

PS: I wasn't entirely sure about the naming of wxPrivateTypeId please let me know if that name is suitable or not and I will change it if isn't.

### DIFF
--- a/include/wx/typeinfo.h
+++ b/include/wx/typeinfo.h
@@ -24,16 +24,6 @@
 
 #include "wx/defs.h"
 
-//
-// Left for compatibility reasons
-//
-
-#define _WX_DECLARE_TYPEINFO_CUSTOM(CLS, IDENTFUNC)
-#define WX_DECLARE_TYPEINFO_INLINE(CLS)
-#define WX_DECLARE_TYPEINFO(CLS)
-#define WX_DEFINE_TYPEINFO(CLS)
-#define WX_DECLARE_ABSTRACT_TYPEINFO(CLS)
-
 #ifndef wxNO_RTTI
 
 //
@@ -49,6 +39,9 @@
 
 #include <typeinfo>
 #include <cstring>
+
+#define WX_DECLARE_TYPEINFO(CLS)
+#define WX_DECLARE_ABSTRACT_TYPEINFO(CLS)
 
 #if wxTRUST_CPP_RTTI
 
@@ -95,12 +88,7 @@ private:
 
 //
 // When C++ RTTI is not available, we will have to make the type comparison
-// using pointer to a dummy static member variable. This will fail if
-// declared type is used across DLL boundaries, although using
-// WX_DECLARE_TYPEINFO() and WX_DEFINE_TYPEINFO() pair instead of
-// WX_DECLARE_TYPEINFO_INLINE() should fix this. However, that approach is
-// usually not possible when type info needs to be declared for a template
-// class.
+// using pointer to a dummy static member variable.
 //
 
 class wxTypeIdentifier
@@ -130,7 +118,7 @@ private:
     wxTypeIdentifier(const char* ptr) : m_ptr{ptr} { }
 
     template<typename T>
-    friend wxTypeIdentifier wxTypeId(const T&);
+    friend wxTypeIdentifier wxPrivateTypeId(const T&);
 
     const char* m_ptr;
 };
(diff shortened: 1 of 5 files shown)


include/wx/typeinfo.h
```diff
@@ -24,16 +24,6 @@
...
-//
-// Left for compatibility reasons
-//
-
-#define _WX_DECLARE_TYPEINFO_CUSTOM(CLS, IDENTFUNC)
-#define WX_DECLARE_TYPEINFO_INLINE(CLS)
-#define WX_DECLARE_TYPEINFO(CLS)
-#define WX_DEFINE_TYPEINFO(CLS)
-#define WX_DECLARE_ABSTRACT_TYPEINFO(CLS)
-
```
vadz, line 36: I think we should still leave those just in case people use it in their own code, even if they're not used by wx itself any longer. It doesn't cost much to do it and people appreciate when their code continues to compile with newer wx versions without errors.

include/wx/typeinfo.h
```diff
@@ -127,10 +115,10 @@ class wxTypeIdentifier
         #endif // __VISUALC__/!__VISUALC__
     };
 
-    wxTypeIdentifier(const char* ptr) : m_ptr{ptr} { }
+    wxTypeIdentifier(const char* ptr) : m_ptr(ptr) { }
```
vadz, line 118: Not sure why this was changed? It's not a big deal, of course, but using `{}` to initialize this member looked fine.

include/wx/typeinfo.h
```diff
@@ -145,11 +133,26 @@ const char wxTypeIdentifier::wxDummy<T>::ms_wxClassInfo{};
 #endif // __VISUALC__/!__VISUALC__
 
 template<typename T>
-wxTypeIdentifier wxTypeId(const T&)
+wxTypeIdentifier wxPrivateTypeId()
 {
-    return wxTypeIdentifier{&wxTypeIdentifier::wxDummy<T>::ms_wxClassInfo};
+    return wxTypeIdentifier(&wxTypeIdentifier::wxDummy<T>::ms_wxClassInfo);
```
vadz, line 138: Same question here?

vadz (commented):
Thinking more about this, I'm really not sure if this works correctly when using shared libraries. I don't believe we have tests exercising this case, i.e. having a class defined in a shared library and using its type id in the application, have you tested this, ideally under both Linux and macOS (because ELF and Mac are different in this aspect)?

It would be definitely nice if we could things as simply as this, of course, but this seems suspicious, it's not for nothing that libstdc++ implements C++ standard RTTI by comparing names as strings and not just pointers.

include/wx/typeinfo.h
MapleLeaf-X: I went off the comments regarding the macros existing for internal use only but I'll make a few changes and restore the missing macros to make it backward compatible.

include/wx/typeinfo.h
MapleLeaf-X: It was mostly done for consistency and for no other reason.

include/wx/typeinfo.h
MapleLeaf-X: Same as above, done for consistency.

MapleLeaf-X:
> Thinking more about this, I'm really not sure if this works correctly when using shared libraries. I don't believe we have tests exercising this case, i.e. having a class defined in a shared library and using its type id in the application, have you tested this, ideally under both Linux and macOS (because ELF and Mac are different in this aspect)?
> 
> It would be definitely nice if we could things as simply as this, of course, but this seems suspicious, it's not for nothing that libstdc++ implements C++ standard RTTI by comparing names as strings and not just pointers.

There are indeed no tests for this but at least from my tests on windows it does work correctly and I'm assuming that if exporting via declspec(dllexport) and the likes works on windows then it should at least in theory work on other platforms as well. unfortunately I can't put it to a real test on linux or osx, however it shouldn't be worse than how it previously worked.

MapleLeaf-X:
It appears that under certain conditions it can still fail so I'll be looking into this for now until I'll have it resolved.

vadz:
Please note that both ELF and macOS are very different from Windows in this aspect, so it would be really great to have a test for this case added to our test suite so that it could be run by the CI on all platforms. I realize that it's tricky to add it due to build system complications, but it still would be very nice to have...

MapleLeaf-X:
I did a bit of manual testing with GCC on linux with -O3 and it did not seem to break so it seem to work. I'll try and look into adding more tests and we'll see how that goes but I can't promise anything.

vadz:
I don't think it's `-Ox` really changes anything (but I could be wrong), it's having the same symbol in multiple places which does. E.g. consider the application using `libfoo` where it is defined and also using `libbar` using `libfoo`. I'm afraid we could have 2 different symbols, with different addresses, in this case.

MapleLeaf-X:
Optimizations absolutely can affect symbols as in the case of MSVS where /Gy and /OPT:ICF will merge constants but that is technically non conforming behavior (merging functions falls under as-is so as long as semantics aren't broken it will be considered conforming). as for multiple symbols I do believe that it should work as intended at least for wxWidgets since the code is mostly identical to what it originally used to be minus the requirement of default constructors.

vadz:
Thanks, comparing this with the pre-a0d6d1816dd87357a33df7b9dc4fbf04722f3e1a version I don't think this can introduce any regressions, so I'm going to merge it.

MapleLeaf-X:
You're welcome.

Issue #23357, opened 2023-03-17
Ensure that clipping is enabled when SetDeviceClippingRegion() is called


AliKet:
### Description

![Screenshot_20230317_181246](https://user-images.githubusercontent.com/7704771/225979222-54df8a7e-4d83-4782-80e1-a69114f2ef0f.png)

From the image above you can see that the multicells (**Big box!** for example) is not rendered correctly!
and the correct rendering should look like this:

![Screenshot_20230317_182054](https://user-images.githubusercontent.com/7704771/225981235-81d34291-76da-45ff-8dfc-863f3b01b280.png)

I still don't know what's causing it, maybe a bug in the `wxGrid` implementation?
but with this patch:
```
diff --git a/src/qt/dc.cpp b/src/qt/dc.cpp
index 62b2b80fc8..40352045c8 100644
--- a/src/qt/dc.cpp
+++ b/src/qt/dc.cpp
@@ -477,6 +477,8 @@ void wxQtDCImpl::DoSetDeviceClippingRegion(const wxRegion& region)
     }
     else
     {
+        m_qtPainter->setClipping( true );
+
         QRegion qregion = region.GetHandle();
         // Save current origin / scale (logical coordinates)
         QTransform qtrans = m_qtPainter->worldTransform();

```
the problem is gone.

In fact, there is another bug related to clipping a multicells to be correctly rendered, by applying a second patch:
```
diff --git a/src/generic/grid.cpp b/src/generic/grid.cpp
index ec5b579a42..2e0ba1a827 100644
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -6841,6 +6841,10 @@ void wxGrid::DrawAllGridWindowLines(wxDC& dc, const wxRegion & WXUNUSED(reg), wx
                 {
                     case CellSpan_Main:
                         rect = CellToRect(j,i);
+#ifdef __WXQT__
+                        rect.width -= 1;
+                        rect.height -= 1;
+#endif
                         rect.Offset(-gridOffset);
                         CalcScrolledPosition( rect.x, rect.y, &rect.x, &rect.y );
                         clippedcells.Subtract(rect);
@@ -6848,6 +6852,10 @@ void wxGrid::DrawAllGridWindowLines(wxDC& dc, const wxRegion & WXUNUSED(reg), wx
 
                     case CellSpan_Inside:
                         rect = CellToRect(j + cell_rows, i + cell_cols);
+#ifdef __WXQT__
[...]

PR #24509, opened 2024-04-26
[wxQt] Some wxGrid fixes

AliKet:
The first commit 71c6dd03b7e3ff61f2e888acb8588efba39b006d also fixes the bug (clipping a multicells which not rendered correctly) mentioned here #23357


AliKet:
> 
> Strange that we still need this wxQt-specific `Deflate()` call, do you have any idea why?

No idea, sorry!

PR #24514, opened 2024-04-28
Simplify test for scrollbars creation in wxOSX

vadz:
Tthe existing test seems unnecessary complicated to me, so I'd suggest simplifying it like this. Am I missing something here?


csomor (approved):
yes, with the current dependency of our wxStatusBar I don't see a logical difference, thanks

PR #24518, opened 2024-05-03
Fix compile issue with credentials dialog

Blake-Madden:
You will get a compile error about `wxTextCtrl` being unknown if you include **creddlg.h** before **textctr.h**. Although `wxTextCtrl` is forward declared in **creddlg.h** (a recent change), that is not enough because `wxTextCtrl` objects are used in this header. Moving that code into the implementation file fixes the compile issue.


vadz:

PR #24524, opened 2024-05-09
Fix dylib symlinks on Apple Silicon Macs

### LINKED ISSUE #23143: Installation failing on Apple Silicon Macs
For some reason several students in my course are having the installation fail for wxWidgets on Apple Silicon Macs. They are doing:

sudo make install

This should install the system into /usr/local/lib. For some users, instead of creating the shared library and links to the library, it is creating the actual library files three times. For example:

-rwxr-xr-x   1 student  staff  15824056 Jan 13 16:36 libwx_osx_cocoau_core-3.2.0.1.0.dylib
-rwxr-xr-x   1 student  staff  15824056 Jan 13 16:36 libwx_osx_cocoau_core-3.2.0.dylib
-rwxr-xr-x   1 student  staff  15824056 Jan 13 16:36 libwx_osx_cocoau_core-3.2.dylib

During the install there are a bunch of lines in their output that say "generating fake signature". I don't see those lines on my output. 

This causes an error due to linking the same library more than once for some reason. What is strange is that some of those systems are configured exactly the same as my system and it works just fine for mine. 

I have temporarily fixed the problem by deleting the files and manually copying the lib directory using cp -a, but want to know what is broken on those systems that makes the install fail?

speleo3:
Fixes https://github.com/wxWidgets/wxWidgets/issues/23143

### DIFF
--- a/configure
+++ b/configure
@@ -31573,6 +31573,10 @@ for dep in \${libnames} ; do
     changes="\${changes} -change \${4}/\${dep} \${3}/\${dep}"
 done
 for i in \${libnames} ; do
+    if test -L \${1}/\${i}; then
+        # skip symbolic links
+        continue
+    fi
     ${HOST_PREFIX}install_name_tool \${changes} -id \${3}/\${i} \${1}/\${i}
 done
 
--- a/configure.ac
+++ b/configure.ac
@@ -3500,6 +3500,10 @@ for dep in \${libnames} ; do
     changes="\${changes} -change \${4}/\${dep} \${3}/\${dep}"
 done
 for i in \${libnames} ; do
+    if test -L \${1}/\${i}; then
+        # skip symbolic links
+        continue
+    fi
     ${HOST_PREFIX}install_name_tool \${changes} -id \${3}/\${i} \${1}/\${i}
 done
 


vadz:
Nice, thanks! I was looking at this too and thought we could achieve the same thing by replacing

```
libnames=\`cd lib ; ls -1 | grep '\.[[0-9]][[0-9]]*\.dylib\$'\`
```

with just something like

```
libnames=\`find lib -type -f -name '*.dylib'\`
```

because I have no idea why do we need the check for digits, but this works too.

speleo3:
Your `find lib -type -f` solution looks more elegant :+1: 

Assuming that no library links to a symlink name (that would need to be changed with `-change`).

speleo3:
This also worked for me:
```
libnames=\`find lib -type f -name '*.dylib' | xargs basename\`
```

vadz:
Please let me know which one do you prefer and if it's the `find` one, could you please (force) push to this PR so that I could be sure that I apply exactly what you tested? TIA!

BTW, a slight simplification would be to use

```
libnames=\`find lib -type f -name '*.dylib' -exec basename {} \; \`
```

but I'm not quite sure about the quoting.

speleo3:
I have a slight preference for the current solution. I like that it's very explicit about not running `install_name_tool ` on symlinks.

So I think this is ready for merging.

vadz:
Thanks again, pushing this to master and will also include it in the upcoming 3.2.5 (just in time!).

vadz:
@speleo3 Could you please check of #25182 works for you? We need to apply it or something similar to restore using correct install names under Mac.

PR #24525, opened 2024-05-13
Fix misleading comment in Thread sample

Blake-Madden:
Fixes a comment that is a bit confusing/misleading,


vadz:

PR #24529, opened 2024-05-14
BUG FIX #24528 - Widgets Demo

### LINKED ISSUE #24528: [macOS] widgets sample crashes immediately
### Description
Try to build the widgets sample on macOS with a newly installed wxWidgets version (configure --enable-debug, followed by make). Afterwards, compile the sample "widgets" using "make" and start the program. It crashes immediately because the pages are not correctly setup.

- wxWidgets version you use: today's most up to date developer built

hwiesmann:
Widgets demo makes incorrectly assumes that when a page has children its content has been created. This is incorrect because some pages create as a default children (e.g. scrollbars etc.). This patch adds a flag checking if a page's content has been created.

### DIFF
--- a/samples/widgets/bmpcombobox.cpp
+++ b/samples/widgets/bmpcombobox.cpp
@@ -410,7 +410,7 @@ void BitmapComboBoxWidgetsPage::CreateContent()
 
     SetSizer(sizerTop);
 
-    this->SetContentCreated();
+    SetContentCreated();
 }
 
 // ----------------------------------------------------------------------------
--- a/samples/widgets/bmpcombobox.cpp
+++ b/samples/widgets/bmpcombobox.cpp
@@ -409,6 +409,8 @@ void BitmapComboBoxWidgetsPage::CreateContent()
     Reset();
 
     SetSizer(sizerTop);
+
+    this->SetContentCreated();
 }
 
 // ----------------------------------------------------------------------------
--- a/samples/widgets/button.cpp
+++ b/samples/widgets/button.cpp
@@ -390,6 +390,8 @@ void ButtonWidgetsPage::CreateContent()
     CreateButton();
 
     SetSizer(sizerTop);
+
+    this->SetContentCreated();
 }
 
 // ----------------------------------------------------------------------------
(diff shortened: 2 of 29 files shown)


PBfordev:
How comes the sample has worked for so many years before: Is this macOS specific? The sample for the current GIT master works fine for me on Windows 10.

Or is the issue a regression due to recent changes, similarly to #24487 (which however manifested only when the widgets were **re**created)?

hwiesmann:
May be that it is macOS specific. Maybe the issue is related to a regression in sizers. I have observed that sizers are not working well since I have upgraded to the latest version of wxWidgets (at least for macOS, I will open an issue on this topic soon, too). Therefore, as the size of a window is not correctly calculated, it may be (weak assumption from my side) that scrollbars are automatically created leading to the actual bug in the widgets demo sample.

vadz:
I think the problem is that `GetChildren()` in wxOSX include scrollbars, which _really_ shouldn't be the case. We already have `IsClientAreaChild()` as a workaround for this, but it's inapplicable to tests such as `GetChildren().empty()`, unfortunately.

I didn't have time to test this yet but I believe this would fix the problem in a much less intrusive way:

```diff
diff --git a/samples/widgets/widgets.cpp b/samples/widgets/widgets.cpp
index cf3ad97986..20823ce7cd 100644
--- a/samples/widgets/widgets.cpp
+++ b/samples/widgets/widgets.cpp
@@ -821,7 +821,18 @@ void WidgetsFrame::OnPageChanged(WidgetsBookCtrlEvent& event)
     // create the pages on demand, otherwise the sample startup is too slow as
     // it creates hundreds of controls
     WidgetsPage *curPage = CurrentPage();
-    if ( curPage->GetChildren().empty() )
+
+    bool hasChildren = false;
+    for ( const auto child : curPage->GetChildren() )
+    {
+        if ( !IsClientAreaChild(child) )
+        {
+            hasChildren = true;
+            break;
+        }
+    }
+
+    if ( !hasChildren )
     {
         wxWindowUpdateLocker noUpdates(curPage);
         curPage->CreateContent();
```

vadz:
I've tested it now and the fix above indeed fixes the problem, so I'm going to push it instead of this PR.

Thanks for reporting this!

PR #24536, opened 2024-05-16
Check wxXRC_NO_RELOADING out of the loop

joro75:
Small optimisation:
Instead of checking if the wxXRC_NO_RELOADING flag is set, inside the loop that checks each XmlResource record, this will check the flag before the loop itself is even started.

### DIFF
--- a/src/xrc/xmlres.cpp
+++ b/src/xrc/xmlres.cpp
@@ -723,7 +723,7 @@ bool wxXmlResource::UpdateResources()
 
     // We never do it if this flag is specified.
     if ( m_flags & wxXRC_NO_RELOADING )
-        return;
+        return rt;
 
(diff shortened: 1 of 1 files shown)


joro75:
Is there someone who can restart the failed check? For some reason something went wrong and the check timed-out. I do expect that restarting the check will let it pass (as the other checks did).

Thanks in advance!

vadz:
Thanks, I'll apply this soon although I'd like to just `return` if this flag is used (but it's a trivial change and I can do it myself).

joro75:
@vadz I already updated the files for you, so it now directly returns from wxXmlResource::UpdateResources if wxXRC_NO_RELOADING is set.

vadz:
Thanks for the update! Will (squash) merge soon.

Issue #24532, opened 2024-05-14
[macOS] Ugly drawing of slider's ticks


hwiesmann:
### Description
Compile widgets sample with default compiler options and select the native slider in the tree book. When the ticks are shown it can be seen that the first half of the ticks is slightly centered above the slider's line, the other half is slight centered below the slider's line (see attached screenshot).

<img width="1267" alt="Bildschirmfoto 2024-05-14 um 22 33 54" src="https://github.com/wxWidgets/wxWidgets/assets/518892/e56b41ec-3ec7-49e8-bb0e-4db0eb12f1b9">

The reason for this behavior is the result line 22 in slider_osx.cpp:

```cpp
 // The dimensions of the different styles of sliders (from Aqua document)

#if wxOSX_USE_COCOA
    #define wxSLIDER_DIMENSIONACROSS_WITHTICKMARKS 28 /* this is the critical line! */
    #define wxSLIDER_DIMENSIONACROSS_ARROW 21
#else
    #define wxSLIDER_DIMENSIONACROSS_WITHTICKMARKS 24
    #define wxSLIDER_DIMENSIONACROSS_ARROW 18
#endif
```

Whenever the defined preprocessor variable has an even value, the slider's ticks are badly centered, choosing an odd value solves the issue. So, 29 is a good value.

BTW: Aqua is pretty much outdated.

- wxWidgets version you use: latest master built from GitHub
- wxWidgets port you use: macOS 14.4.1

vadz:
Thanks, let's indeed fix this but I don't know what are the currently recommended values, @csomor would you know this by chance? If we change this to 29 we should probably use the other branch to use 25?

P.S. I've edited the description to use proper code block (triple single quotes).

csomor:
Actually the 28 is correct for the slider, but the labels calculation lead to a height of 49 when shown, this number leads to the ugly rendering. When neither min max nor value are shown, then IMHO things look fine with 28

hwiesmann:
Hi,

When the min and max value are not shown but only the value the drawing of the slider’s ticks is as ugly as with the min and max values (I use the widgets sample, see screenshot attached).

￼
Nevertheless, we should change the slider's implementation. Instead of changing one of the constants we can also make sure that the overall height is always an even number.

csomor:
> When the min and max value are not shown but only the value the drawing of the slider’s ticks is as ugly as with the min and max values (I use the widgets sample, see screenshot attached). ￼

[...]

PR #24537, opened 2024-05-16
BUG FIX for ticket #24532 - Ticks were not centered with respect to the slider's line

hwiesmann:
To have a proper centering with respect to the slider's line, the size of the slider in the dimension perpendicular to the slider's line has to be an even number. Therefore, when determining the best size, the size is adjusted to an even number when ticks are shown.


vadz:

PR #24538, opened 2024-05-17
Timer doc updates

Blake-Madden:
Remove warning about Windows XP since we don't support that platform anymore. (The support link in the comment doesn't go anywhere now also.)

Mention using the modern method of `Bind()` with `wxEVT_TIMER`.

Format improvement.


vadz:

PR #24540, opened 2024-05-17
BUG FIX #24539 - Make sure that slider value and value label are synchronized

### LINKED ISSUE #24539: [macOS] Slider value and value label do not match
### Description
The slider value and the corresponding value label do not match if
 - the value set by using SetValue(int) is outside the slider's range
 - the range is modified so that the actual value is afterwards outside the range

The bug can be seen when using the widgets sample code and selecting the slider control page.Entering a minimum value of 10 for the slider's range while the actual value is 0 leads to the situation that the slider is positioned at 10 while the label still shows the old value 0.

PS: Calling GetValue() will return 10.
PPS: The value itself is always adjusted so that it is within the slider's range.

- wxWidgets version you use: latest master built

hwiesmann:
Now, the methods SetValue and SetRange have been modified so that the slider's value and the shown value label are consistent. Not necessary code that checks the value against the limits has been removed because macOS does the same work anyway.

### DIFF
--- a/src/osx/slider_osx.cpp
+++ b/src/osx/slider_osx.cpp
@@ -155,15 +155,19 @@ int wxSlider::GetValue() const
 
 void wxSlider::SetValue(int value)
 {
-    if ( m_macValueStatic )
-    {
-        wxString valuestring;
-        valuestring.Printf( wxT("%d"), value );
-        m_macValueStatic->SetLabel( valuestring );
-    }
-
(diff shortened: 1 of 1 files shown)


vadz:
Thanks! I don't know this code well, but if this works correctly in interactive testing, this should be fine to apply.

hwiesmann:
Yes, works in widgets demo program.

PR #24542, opened 2024-05-18
Add wxXmlResource::XRCIDCount() to determine the number of known XRCIDs

joro75:
Added the helper function `XRCIDCount()` to `wxXmlResource` to determine the number of XRCID that are present in the XRC XML Resource. 
This can be helpful to detemine and investigate the number of XRCID in the internal cache. Also tests are added to check the implementation, and other tests that are now possible to verify the internal behaviour of the XRCID functions.

### DIFF
--- a/include/wx/xrc/xmlres.h
+++ b/include/wx/xrc/xmlres.h
@@ -269,6 +269,9 @@ public:
     // it checks all the IDs used in XRC.
     static wxString FindXRCIDById(int numId);
 
+    // Determine the total number of ID's that are present in the XML
+    // resource.
+    static int XRCIDCount();
 
     // Returns version information (a.b.c.d = d+ 256*c + 256^2*b + 256^3*a).
     long GetVersion() const { return m_version; }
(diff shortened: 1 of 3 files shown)


joro75:
Some additional information regarding this suggestion: We have an application with over 300 XRC files (some with several dialogs/panels), and at this moment we have no idea, how many XRCID's are used, and if this is a problem for the internal XRCID cache. This addition is a general addition, and does allow us to get some more information if the `XRCID_Records` should be increased/optimized.

vadz:
Thanks for the explanation but I'm still not entirely sure what are you going to do with this information? I.e. suppose you find that 12,345 IDs are used — now what? Sorry if it seems like a stupid question, but I just don't see how could the result of the new function be useful for anything... If there is a problem with XRC IDs hash performance, wouldn't it be better to explore replacing the extremely simplistic hash table used there now with something more performant?

Also, any new function should be documented in `interface/wx/xmlres.h`, but at the moment this is secondary as I'm not sure at all we really want to add it. Please let me know if I'm missing something.

vadz:
Sorry but I don't think we want to apply this without some good rationale for this change. Please comment/reopen (not sure if you can, but I'll do it if necessary) if you can provide one.

PR #24543, opened 2024-05-18
Fix incorrect documention for wxString::Clear()

Blake-Madden:
The docs say that `wxString::Clear()` frees the string's memory, but that is not correct. It calls the stanards `clear()` function which only sets the string's size to zero.

I added an explanation for needing to call `Shrink()` or `shrink_to_fit()` to actually (request to) shrink the string's capacity.


vadz:
Thanks! Applied as 28a1e55017b98d04a1ee235169c7c826948ca5e4 but forgot to mention this PR number in its message.

PR #24547, opened 2024-05-20
wxrc: `GetInternalFileName`: Call to `Printf`: Format string arg is now constant.

DoctorNoobingstoneIPresume:
In the original code:

```
s.Printf(wxFileNameFromPath(parOutput) + wxT("$%03i-") + name2, i);
```

if `wxFileNameFromPath(parOutput2)` or `name2` had contained `'%'` (format specifiers), the results would have been bad.

These strings are now given as additional arguments to `Printf` (corresponding to `"%s"` format specifiers),
and the format string argument itself is now constant.

This is a replacement of https://github.com/wxWidgets/wxWidgets/commit/9a5d32db95adc191b24fd39217838f6b7f2c091e as per https://github.com/wxWidgets/wxWidgets/pull/24437#issuecomment-2077394039.

### DIFF
--- a/utils/wxrc/wxrc.cpp
+++ b/utils/wxrc/wxrc.cpp
@@ -354,13 +354,14 @@ wxString XmlResApp::GetInternalFileName(const wxString& name, const wxArrayStrin
     name2.Replace(wxT("*"), wxT("_"));
     name2.Replace(wxT("?"), wxT("_"));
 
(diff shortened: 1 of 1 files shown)


vadz:
Applying as well, thanks!

DoctorNoobingstoneIPresume:
You are most welcome, thank you for this and for all your work !! 🌈

PR #24549, opened 2024-05-22
Remove mention of variable in comments that doesn't exist

Blake-Madden:


vadz:

PR #24553, opened 2024-05-23
Remove outdated warning and example about wxString being COW from thread docs

Blake-Madden:
Minor punctuation tweaks also.

### DIFF
--- a/interface/wx/event.h
+++ b/interface/wx/event.h
@@ -416,41 +416,23 @@ public:
         moment).
 
         QueueEvent() can be used for inter-thread communication from the worker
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, but I think it's a pity not to mention `wxThreadEvent` here any more, as it was a nice side effect. What about adding something like

> Note that if you want to pass more data than just a single string and/or an integer carried by wxCommandEvent to the main thread, you may use wxThreadEvent instead.

?

Blake-Madden:
Added your comments and a really simple `wxThreadEvent` example.

PR #24554, opened 2024-05-24
Add __WXFUNCTION_SIG__ macro

Blake-Madden:
If compiling as C++20 or using GCC/clang or MSVC, expands to function signature. Falls back to the function name (using `__func__`) otherwise.

The rationale behind this is that `__func__` can yield names such as `operator()`, which (IMO) is not all that useful. This macro can be much more informative.

### DIFF
--- a/include/wx/cpp.h
+++ b/include/wx/cpp.h
@@ -115,7 +115,7 @@
 #endif /* __WXFUNCTION__ already defined */
 
 /*
-    Display the current function's full signature, if available.
+    Expands to the current function's full signature, if available.
 
     Falls back to the function name (i.e., __func__) if not available.
  */
--- a/include/wx/cpp.h
+++ b/include/wx/cpp.h
@@ -123,7 +123,7 @@
     #if __cplusplus >= 202002L
         #include <source_location>
         // actually displays the signature, not just the name
-        #define __WXFUNCTION_SIG__ std::source_location::current().function_name() 
+        #define __WXFUNCTION_SIG__ std::source_location::current().function_name()
     #elif __VISUALC__
         #define __WXFUNCTION_SIG__ __FUNCSIG__
     #elif defined(__clang__)
--- a/include/wx/cpp.h
+++ b/include/wx/cpp.h
@@ -114,6 +114,26 @@
     #define __WXFUNCTION__ __func__
 #endif /* __WXFUNCTION__ already defined */
 
+/*
+    Display the current function's full signature, if available.
+
+    Falls back to the function name (i.e., __func__) if not available.
+ */
+#ifndef __WXFUNCTION_SIG__
+    #if __cplusplus >= 202002L
+        #include <source_location>
+        // actually displays the signature, not just the name
+        #define __WXFUNCTION_SIG__ std::source_location::current().function_name() 
+    #elif __VISUALC__
+        #define __WXFUNCTION_SIG__ __FUNCSIG__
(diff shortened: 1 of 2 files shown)


include/wx/cpp.h
```diff
@@ -114,6 +114,26 @@
...
+ */
+#ifndef __WXFUNCTION_SIG__
+    #if __cplusplus >= 202002L
+        #include <source_location>
+        // actually displays the signature, not just the name
+        #define __WXFUNCTION_SIG__ std::source_location::current().function_name()
+    #elif defined(__VISUALC__)
+        #define __WXFUNCTION_SIG__ __FUNCSIG__
+    #elif defined(__clang__)
+        #define __WXFUNCTION_SIG__ __func__
```
vadz, line 130: I thought clang supported `__PRETTY_FUNCTION__` too, so do we really need to have a separate test for it here?

interface/wx/cpp.h
```diff
@@ -52,9 +52,20 @@
...
     Please use the standard macro instead.
 
+    @see __WXFUNCTION_SIG__
+
     @header{wx/cpp.h}
 */
 #define __WXFUNCTION__ __func__
 
+/**
+    Display the current function's full signature, if available.
```
vadz, line 62: I think it would be more correct to say
```suggestion
    Expands to the current function's full signature, if available.
```

interface/wx/cpp.h
```diff
@@ -52,9 +52,20 @@
...
+
+    Falls back to the function name (i.e., @c \__func__) if not available.
```
vadz, line 64: Would be nice to show a simple example showing the difference.

interface/wx/cpp.h
```diff
@@ -52,9 +52,20 @@
...
+
+    @header{wx/cpp.h}
+*/
```
vadz, line 67: Please add `@since 3.3.0`.

vadz (commented):
Looks useful, thanks!

interface/wx/cpp.h
Blake-Madden: Darn, I keep forgetting that. Fixed.

include/wx/cpp.h
Blake-Madden: Indeed it does. I read people complaining that it is missing that, but I verified that it does on compiler explorer. Fixed.

interface/wx/cpp.h
Blake-Madden: Fixed

interface/wx/cpp.h
Blake-Madden: Done. Used a member function with a `const` decorator to really show off the difference.

0tkl:
It breaks on apple clang 14 when compiled with C++20.
```
subprojects/wxWidgets-master/include/wx/cpp.h:124:18: fatal error: 'source_location' file not found
        #include <source_location>
                 ^~~~~~~~~~~~~~~~~
1 error generated.
```
What about using [`__cpp_lib_source_location`](https://en.cppreference.com/w/cpp/feature_test#cpp_lib_source_location) macro instead of `__cplusplus` as the condition?

Blake-Madden:
I'll need to add a `__has_include` workaround or something like that. I'll check it out.

Blake-Madden:
`__has_cpp_attribute` seems really quirky under GCC trunk. I may just remove the C++20 stuff and rely on the platform macros.

PR #24559, opened 2024-05-25
Do not scale the size of bitmap in wxMSW Create(size, dc) overload

vadz:
The size passed to Create() should be interpreted in the same way as the size passed to CreateWithLogicalSize() and in wxMSW this means that it should _not_ be scaled.

This makes the size of the bitmap returned by this function overload consistent with wxDC::GetSize(), so it looks like the right thing to do, even if it's a backwards-incompatible change.

----

Please correct me if I'm wrong, but it seems that we have a mistake here, as the size passed to `Create()` is supposed to be in logical == physical (under MSW) pixels and not in DIPs, but this function behaved like `CreateWithDIPSize()` rather than `CreateWithLogicalSize()` as, IMO, it should do. But I know that I've already made several similar mistakes in the past, so please check that I'm not making one again here. TIA!

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -92,6 +92,10 @@ Changes in behaviour not resulting in compilation errors
   made is behaviour there incompatible with the other platforms. Please call
   wxWebRequest::EnablePersistentStorage() explicitly if you need it.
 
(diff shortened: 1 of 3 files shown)


vadz:
No comments, so merging.

PR #24561, opened 2024-05-26
Implement wxGrid's overlay selection

AliKet:
see #22694

### DIFF
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -2324,6 +2324,9 @@ public:
     wxArrayInt GetSelectedRows() const;
     wxArrayInt GetSelectedCols() const;
 
+    // Return the selected blocks (only the visible ones) as a vector of rects.
+    void GetSelectedRectangles(std::vector<wxRect>& rectangles) const;
+
     // This function returns the rectangle that encloses the block of cells
     // limited by TopLeft and BottomRight cell in device coords and clipped
     //  to the client size of the grid window.
@@ -2331,6 +2334,13 @@ public:
     wxRect BlockToDeviceRect( const wxGridCellCoords & topLeft,
                               const wxGridCellCoords & bottomRight,
                               const wxGridWindow *gridWindow = nullptr) const;
+
+    // Similar to BlockToDeviceRect() above but mainly for grid selection use.
+    // i.e. GetSelectedRectangles() uses this function as a helper.
+    void BlockToDeviceRect(const wxGridCellCoords& topLeft,
+                           const wxGridCellCoords& bottomRight,
+                           wxGridWindow* gridWin,
+                           std::vector<wxRect>& rectangles) const;
 
     // Access or update the selection fore/back colours
     wxColour GetSelectionBackground() const
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -1704,6 +1704,16 @@ public:
     void RefreshBlock(int topRow, int leftCol,
                       int bottomRow, int rightCol);
 
+    // This function simply calls RefreshRect() to refresh the cell at coordinates
+    // _coords_ and its corresponding row and column labels for highlighting.
+    void RefreshBlock(const wxGridCellCoords& coords);
+
+    // Refresh the entire grid if _rect_ is null. If not null, then:
+    // - the intersection of the cells area with the rectangle _rect_ will be refreshed.
+    // - the projection of the rectangle _rect_ on the row label window will be refreshed.
+    // - the projection of the rectangle _rect_ on the col label window will be refreshed.
+    void RefreshRect(wxRect* rect);
+
     // Refresh one or more areas (a combination of wxGridArea enums) entirely.
     void RefreshArea(int areas);
 
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -1642,6 +1642,7 @@ public:
     void DrawHighlight(wxDC& dc, const wxGridCellCoordsVector& cells);
     void DrawFrozenBorder( wxDC& dc, wxGridWindow *gridWindow );
     void DrawLabelFrozenBorder( wxDC& dc, wxWindow *window, bool isRow );
+    void DrawSelection( wxDC& dc, wxGridWindow *gridWindow );
 
     void ScrollWindow( int dx, int dy, const wxRect *rect ) override;
 
@@ -2277,6 +2278,12 @@ public:
     // make the cell editable/readonly
     void SetReadOnly(int row, int col, bool isReadOnly = true);
 
+    // Return @true if overlay selection can be used (wxUSE_GRAPHICS_CONTEXT=1)
+    // and DisableOverlaySelection() hadn't been called.
+    bool UsesOverlaySelection() const;
+
+    void DisableOverlaySelection();
+
     // ------ select blocks of cells
     //
     void SelectRow( int row, bool addToSelected = false );
@@ -2435,6 +2442,9 @@ public:
     // unset any existing sorting column
     void UnsetSortingColumn() { SetSortingColumn(wxNOT_FOUND); }
 
+    // Invalidate overlay selection whenever the selection changes or scrolls.
+    void InvalidateOverlaySelection();
+
     // override some base class functions
     virtual void Fit() override;
     virtual void SetFocus() override;
@@ -2506,6 +2516,10 @@ protected:
 
     wxColour    m_selectionBackground;
     wxColour    m_selectionForeground;
+
+#if wxUSE_GRAPHICS_CONTEXT
+    bool m_usesOverlaySelection = true;
+#endif
 
     // NB: *never* access m_row/col arrays directly because they are created
     //     on demand, *always* use accessor functions instead!
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -1658,7 +1658,7 @@ public:
     void DrawHighlight(wxDC& dc, const wxGridCellCoordsVector& cells);
     void DrawFrozenBorder( wxDC& dc, wxGridWindow *gridWindow, const wxRect& renderExtent = {} );
     void DrawLabelFrozenBorder( wxDC& dc, wxWindow *window, bool isRow );
-    void DrawSelection( wxDC& dc, wxGridWindow *gridWindow, const wxRect& renderExtent = {} );
+    void DrawOverlaySelection( wxDC& dc, wxGridWindow *gridWindow, const wxRect& renderExtent = {} );
 
     void ScrollWindow( int dx, int dy, const wxRect *rect ) override;
 
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -1656,9 +1656,9 @@ public:
     void DrawAllGridWindowLines( wxDC& dc, const wxRegion & reg , wxGridWindow *gridWindow);
     void DrawCell( wxDC& dc, const wxGridCellCoords& );
     void DrawHighlight(wxDC& dc, const wxGridCellCoordsVector& cells);
-    void DrawFrozenBorder( wxDC& dc, wxGridWindow *gridWindow );
+    void DrawFrozenBorder( wxDC& dc, wxGridWindow *gridWindow, const wxRect& renderExtent = {} );
     void DrawLabelFrozenBorder( wxDC& dc, wxWindow *window, bool isRow );
-    void DrawSelection( wxDC& dc, wxGridWindow *gridWindow );
+    void DrawSelection( wxDC& dc, wxGridWindow *gridWindow, const wxRect& renderExtent = {} );
 
     void ScrollWindow( int dx, int dy, const wxRect *rect ) override;
 
@@ -2350,7 +2350,7 @@ public:
     wxArrayInt GetSelectedCols() const;
 
     // Return the selected blocks (only the visible ones) as a vector of rects.
-    void GetSelectedRectangles(std::vector<wxRect>& rectangles) const;
+    void GetSelectedRectangles(std::vector<wxRect>& rectangles, const wxRect& renderExtent) const;
 
     // This function returns the rectangle that encloses the block of cells
     // limited by TopLeft and BottomRight cell in device coords and clipped
@@ -2365,7 +2365,8 @@ public:
     void BlockToDeviceRect(const wxGridCellCoords& topLeft,
                            const wxGridCellCoords& bottomRight,
                            wxGridWindow* gridWin,
-                           std::vector<wxRect>& rectangles) const;
+                           std::vector<wxRect>& rectangles,
+                           const wxRect& renderExtent) const;
 
     // Access or update the selection fore/back colours
     wxColour GetSelectionBackground() const
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -613,6 +613,12 @@ public:
                            int horizAlign,
                            int vertAlign,
                            int textOrientation) const;
+
+    // Draw highlighted row/col label.
+    virtual void DrawHighlighted(const wxGrid& grid,
+                                 wxDC& dc,
+                                 wxRect& rect,
+                                 int rowOrCol) const;
 };
 
 // Currently the row/column/corner renders don't need any methods other than
@@ -643,6 +649,11 @@ public:
     virtual void DrawBorder(const wxGrid& grid,
                             wxDC& dc,
                             wxRect& rect) const override;
+
+    virtual void DrawHighlighted(const wxGrid& grid,
+                                 wxDC& dc,
+                                 wxRect& rect,
+                                 int row) const override;
 };
 
 // Column header cells renderers
@@ -653,6 +664,11 @@ public:
     virtual void DrawBorder(const wxGrid& grid,
                             wxDC& dc,
                             wxRect& rect) const override;
+
+    virtual void DrawHighlighted(const wxGrid& grid,
+                                 wxDC& dc,
+                                 wxRect& rect,
+                                 int col) const override;
 };
 
 // Header corner renderer
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -2249,7 +2249,9 @@ public:
     wxGridCellEditor *GetDefaultEditor() const;
     wxGridCellEditor* GetCellEditor(int row, int col) const;
 
-
+    // Determine (in constant time) whether the row/col should be highlighted or not.
+    bool IsRowLabelHighlighted(int row) const;
+    bool IsColLabelHighlighted(int col) const;
 
     // ------ cell value accessors
     //
--- a/include/wx/generic/gridsel.h
+++ b/include/wx/generic/gridsel.h
@@ -139,6 +139,15 @@ private:
     void MergeOrAddBlock(wxGridBlockCoordsVector& blocks,
                          const wxGridBlockCoords& block);
 
+    // Used by Select() and DeselectBlock() to ensure that we always have fewer
+    // blocks selected in m_selection.
+    static void MergeAdjacentBlocks(wxGridBlockCoordsVector& selection);
+
+    // This function attempts to reduce the number of rectangles returned from
+    // wxGrid::GetSelectedRectangles() before trying to convert them to PolyPolygon.
+    // Most of the time this will result in just one rectangle.
+    static void MergeAdjacentRects(std::vector<wxRect>& rectangles);
+
     // All currently selected blocks. We expect there to be a relatively small
     // amount of them, even for very large grids, as each block must be
     // selected by the user, so we store them unsorted.
--- a/include/wx/generic/gridsel.h
+++ b/include/wx/generic/gridsel.h
@@ -114,6 +114,48 @@ public:
     void EndSelecting();
     void CancelSelecting();
 
+    // A simple interface used by wxGrid::DrawSelection() as a helper to draw
+    // the grid selection(s).
+    class PolyPolygon
+    {
+    public:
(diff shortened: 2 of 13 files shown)


AliKet:
@vadz I'm not sure if it's intentional or not, but I noticed that `wxGrid::Render()` doesn't render frozen borders (if any) at all, and I fixed this in [b77b7e4](https://github.com/wxWidgets/wxWidgets/pull/24561/commits/b77b7e4f3d0521081cff477ee89225dff7192197). Just let me know if I should change anything here, TIA

Screenshot from 2024-05-27 20-31-48

Screenshot from 2024-05-27 20-31-04

src/generic/gridsel.cpp
```diff
@@ -907,4 +943,455 @@ void wxGridSelection::MergeOrAddBlock(wxGridBlockCoordsVector& blocks,
...
+              std::back_inserter(m_points));
+}
+
+void wxGridSelection::PolyPolygon::CalcBoundingBox(const std::vector<wxPoint>& points)
+{
+    std::vector<wxPoint>::const_iterator cItr = points.begin(),
+                                         cEnd = points.end();
+    while ( cItr != cEnd )
+    {
+        const wxPoint& point = *cItr++;
```
vadz, line 1365: Why not just
```suggestion
    for ( const auto& point : points )
    {
```
?

src/generic/gridsel.cpp
```diff
@@ -907,4 +943,455 @@ void wxGridSelection::MergeOrAddBlock(wxGridBlockCoordsVector& blocks,
...
+            for ( const auto& p : polygon )
+            {
+                poly.push_back(p.first);
+            }
+
+            std::vector<wxPoint>::const_iterator cItr = poly.begin(),
+                                                 cEnd = poly.end();
+            while ( cItr != cEnd )
+            {
+                const wxPoint& pt = *cItr++;
```
vadz, line 1142: This could be replaced by a range-for too AFAICS.

src/generic/grid.cpp
```diff
@@ -58,6 +61,17 @@ const char wxGridNameStr[] = "grid";
 // Required for wxIs... functions
 #include <ctype.h>
 
+// Drawing transparent selection can only be done via a graphics context,
+// or a DC that has a native support for drawing with alpha, like standard
+// DCs on OSX and GTK3.
```
vadz, line 66: Sorry, it's a pity that you've already done it, but I think there might have been a misunderstanding in one of the previous discussions: I never wanted to imply that you should support using `wxDC` when `wxUSE_G_C==0` in these ports. First of all, I don't even this it can be disabled in them and, second, even if it could, it's definitely not worth it because nobdoy would do it in practice.

So I'd just remove the port checks and use only `#if wxUSE_G_C` (although you might want to `#define wxHAS_TRANSPARENT_SELECTION` in this case and check for this one below to make these checks more readable).

src/generic/grid.cpp
```diff
@@ -6935,6 +7149,116 @@ void wxGrid::DrawRowLabel( wxDC& dc, int row )
...
+    if ( offset.x > 0 )
+        clipRect.x += 1;
+    if ( offset.y > 0 )
+        clipRect.y += 1;
+
+    wxDCClipper clip(dc, clipRect);
+
+    const wxColour colBg = GetSelectionBackground();
+
+#ifdef __WXMSW__
```
vadz, line 7209: I'd appreciate a comment explaining why do we do this only under MSW as it doesn't seem obvious.

vadz (commented):
I haven't tried it yet but globally this looks good, thank you for implementing it!

I wonder if there is some better/more common term for this than "transparent selection", would anybody know how it is called in the existing programs? "Overlay selection" maybe?

In any case, whatever its final name, the new `wxGrid` function for enabling it should be documented.

Finally, ideally, we'd extract (some of?) the new functions in `wx/private/grid.h` in order to allow testing them as it's slightly worrisome that there is a lot of not really obvious new code here which is not covered by the unit tests...

Thanks again for your work!

AliKet:
> I wonder if there is some better/more common term for this than "transparent selection", would anybody know how it is called in the existing programs? "Overlay selection" maybe?
> 
So according to this [page](https://www.infragistics.com/help/winforms/wingrid-enabling-selection-overlay-for-selected-cells):
- I'll change `IsUsingTransparentSelection()` to  `IsUsingSelectionOverlay()` or simply `IsSelectionOverlay()` ?
- Remove `EnableTransparentSelection()` because **selection overlay** will be used by default if supported. But...
- If the environment variable `WXGRID_NO_SELECTION_OVERLAY` is set, then the old why to draw the selection will be used.

What do you think ?

samples/grid/griddemo.cpp
```diff
@@ -74,6 +74,16 @@ class CustomColumnHeaderRenderer : public wxGridColumnHeaderRenderer
         dc.DrawRectangle(rect);
     }
 
+    virtual void DrawHighlighted(const wxGrid& WXUNUSED(grid),
+                                 wxDC& dc,
+                                 wxRect& rect,
+                                 int WXUNUSED(rowOrCol)) const override
+    {
+        dc.SetPen(*wxTRANSPARENT_PEN);
+        dc.SetBrush(wxBrush(wxColour(240, 200, 15, 180)));
```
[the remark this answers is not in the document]
AliKet: The custom header renderer uses hardcoded colours and as such I don't think that's necessary here:

Screenshot from 2024-05-30 15-08-41

src/generic/grid.cpp
AliKet: I wanted to make the selection outline more bold but after thinking more about it I think I'll revert this as it doesn't seem necessary.

Here is how it looks without this dirty hack:

In light mode:
grid_light

In dark mode:
grid_dark

vadz:
> > I wonder if there is some better/more common term for this than "transparent selection", would anybody know how it is called in the existing programs? "Overlay selection" maybe?
> 
> So according to this [page](https://www.infragistics.com/help/winforms/wingrid-enabling-selection-overlay-for-selected-cells):
> 
>     * I'll change `IsUsingTransparentSelection()` to  `IsUsingSelectionOverlay()` or simply `IsSelectionOverlay()` ?

I'd invert the word order as it seems more natural to me (can any native English speakers please comment on whether it's just me?), i.e. would use `IsUsingOverlaySelection()` or `UsesOverlaySelection()` or maybe `HasOverlaySelection()`, but any of these choices are acceptable, so please choose whichever you prefer.

>     * Remove `EnableTransparentSelection()` because **selection overlay** will be used by default if supported. But...
>     * If the environment variable `WXGRID_NO_SELECTION_OVERLAY` is set, then the old why to draw the selection will be used.

I don't see why should we have an environment variable to control something like this, so I'd strongly prefer to avoid it. But we could (should?) provide `DisableOverlaySelection()` to revert to the old appearance for people who would want it, for whatever reason.

Thanks again!

AliKet:
> I'd invert the word order as it seems more natural to me (can any native English speakers please comment on whether it's just me?), i.e. would use `IsUsingOverlaySelection()` or `UsesOverlaySelection()` or maybe `HasOverlaySelection()`, but any of these choices are acceptable, so please choose whichever you prefer.
> 
> >

I chooses `UsesOverlaySelection()` and added `DisableOverlaySelection()` as you suggested. also added documentation for them.

I think this is ready now, but as always I didn't tests this under `wxOSX`. so if you could please do it (or anyone interested) before merging.

TIA!

include/wx/generic/gridsel.h
```diff
@@ -151,6 +162,16 @@ class WXDLLIMPEXP_CORE wxGridSelection
     wxGrid                              *m_grid;
     wxGrid::wxGridSelectionModes        m_selectionMode;
 
+    // Used by wxGrid::DrawSelectionOverlay() to draw a:
+    //
+    // - Simple rectangle (using wxDC::DrawRectangle() if it is empty and the bounding box is valid.
+    // - Simple polygon (using wxDC::DrawPolygon()) if it represents a simple polygon.
+    // - Poly-polygon (using wxDC::DrawPolyPolygon()) if it consists of multi-polygons.
```
vadz, line 169: Really minor, but I'd say:
```suggestion
    // - Poly-polygon (using wxDC::DrawPolyPolygon()) if it consists of multiple polygons.
```

include/wx/generic/gridsel.h
```diff
@@ -18,6 +18,9 @@
 
 #include "wx/vector.h"
 
+// Forward declaration
+namespace wxGridPrivate { class PolyPolygon; }
```
vadz, line 22: I wonder if we should call this class `SelectionShape` instead? This name seems to carry more meaning than the current one (especially because it doesn't even always consists of multiple polygons).

include/wx/generic/gridsel.h
```diff
@@ -114,6 +117,11 @@ class WXDLLIMPEXP_CORE wxGridSelection
     void EndSelecting();
     void CancelSelecting();
 
+    // Return the PolyPolygon object. Call ComputePolyPolygon() if necessary.
```
vadz, line 120: This comment should probably mention that it returns the intersection of the shape of the selection with the given rectangle (if I understand what it does correctly).

include/wx/generic/gridsel.h
```diff
@@ -139,6 +147,9 @@ class WXDLLIMPEXP_CORE wxGridSelection
     void MergeOrAddBlock(wxGridBlockCoordsVector& blocks,
                          const wxGridBlockCoords& block);
 
+    // Called each time the selection changed or scrolled to recompute m_polyPolygon.
+    void ComputePolyPolygon(bool refreshLabelWindows = false, const wxRect& renderExtent = {});
```
vadz, line 151: Sorry, I still think that this function shouldn't have `refreshLabelWindows` argument and that the case of `refreshLabelWindows = true` should be handled in the single place where it's used. I realize that this means we're going to potentially call `Refresh()` twice but I don't think this is a problem in practice, the refresh rectangles should be coalesced anyhow.

And it would make this function much easier to understand.

include/wx/generic/gridsel.h
```diff
@@ -151,6 +162,16 @@ class WXDLLIMPEXP_CORE wxGridSelection
...
 
+    // Used by wxGrid::DrawSelectionOverlay() to draw a:
+    //
+    // - Simple rectangle (using wxDC::DrawRectangle() if it is empty and the bounding box is valid.
+    // - Simple polygon (using wxDC::DrawPolygon()) if it represents a simple polygon.
+    // - Poly-polygon (using wxDC::DrawPolyPolygon()) if it consists of multi-polygons.
+    //
+    std::unique_ptr<wxGridPrivate::PolyPolygon> m_polyPolygon;
+
+    size_t                                      m_isAnyLabelHighlighted = 0;
```
vadz, line 173: I'd also appreciate a comment about the meaning of this one as its name (`isXXX`) implies `bool` type for me, but here it is `size_t`, so clearly there is something else going on here, but I don't really know what.

include/wx/generic/gridsel.h
```diff
@@ -151,6 +162,16 @@ class WXDLLIMPEXP_CORE wxGridSelection
     wxGrid                              *m_grid;
     wxGrid::wxGridSelectionModes        m_selectionMode;
 
+    // Used by wxGrid::DrawSelectionOverlay() to draw a:
+    //
+    // - Simple rectangle (using wxDC::DrawRectangle() if it is empty and the bounding box is valid.
+    // - Simple polygon (using wxDC::DrawPolygon()) if it represents a simple polygon.
+    // - Poly-polygon (using wxDC::DrawPolyPolygon()) if it consists of multi-polygons.
+    //
+    std::unique_ptr<wxGridPrivate::PolyPolygon> m_polyPolygon;
```
vadz, line 171: We should `#include <memory>` in this file instead of relying on it being done implicitly in some other header.

include/wx/generic/grid.h
```diff
@@ -2267,6 +2296,12 @@ class WXDLLIMPEXP_CORE wxGrid : public wxScrolledCanvas
     // make the cell editable/readonly
     void SetReadOnly(int row, int col, bool isReadOnly = true);
 
+    // Return @true if overlay selection can be used (wxUSE_GRAPHICS_CONTEXT=1)
+    // and DisableOverlaySelection() hadn't been called.
+    static bool UsesOverlaySelection();
+
+    static void DisableOverlaySelection();
```
vadz, line 2303: Sorry, but why do these functions have to be global, i.e. class-static? I thought they would be per-grid, so that you could customize the selection of a particular grid without affecting anything else.

Maybe it would make sense to have some global `DisableOverlaySelectionByDefault()` that would set `gs_useOverlaySelection` to `false`, but I'd still add `m_useOverlaySelection`, that would be initialized with `gs_useOverlaySelection` but could be set to `false` using non-static `DisableOverlaySelection()` without affecting any other grids.

include/wx/generic/private/grid.h
```diff
@@ -1227,6 +1236,11 @@ wxGetContentRect(wxSize contentSize,
                  int hAlign,
                  int vAlign);
 
+inline bool operator<(const wxPoint& pt1, const wxPoint& pt2)
```
vadz, line 1239: I'd prefer to define a custom comparator and pass it to `std::{set,map}` explicitly. Comparing points lexicographically doesn't always make sense and while it's unlikely to affect anything right now, when this is declared in a private header, it still might and it could be very annoying to run into problems due to this later.

include/wx/generic/private/grid.h
```diff
@@ -1280,6 +1294,235 @@ TryGetValueAsDate(wxDateTime& result,
...
+    // Return the bounding box of the visible selected blocks. Return empty
+    // rectangle if there is no selection.
+    wxRect GetBoundingBox() const;
+
+    void SetBoundingBox(const wxRect& rect);
+
+private:
+    void CalcBoundingBox(const std::vector<wxPoint>& points);
+
+    // m_counts contains the number of points in each of the polygons in m_points.
```
vadz, line 1327: So the data structure here is a vector of points in all polygons stored as a single continuous vector? I.e. if we have 2 polygons P1 and P2, `m_points` contains P1_first, P1_second, ..., P1_last, P2_first, ..., P2_last?

Is this done for efficiency reasons? Because for me a natural approach would be to just have `vector<vector<wxPoint>> m_polygons` and I wonder why is it not being used here?

include/wx/generic/private/grid.h
```diff
@@ -1280,6 +1294,235 @@ TryGetValueAsDate(wxDateTime& result,
...
+
+    std::vector<int>     m_counts;
+    std::vector<wxPoint> m_points;
+
+    int m_minX = 0, m_maxX = 0,
+        m_minY = 0, m_maxY = 0;
+};
+
+// This class is just a helper that simply converts (using the sweep line algorithm)
```
vadz, line 1336: Sorry but why do we need a class for what looks like just a function? I had a lot of trouble understanding what it does, but now that I (think I) understand it, it looks like it could be just a function of `PolyPolygon`, why do we have to make a separate class for it?

vadz (commented):
Thanks again for all this work, but I still wonder if it can't be simplified a bit more. I'm sorry for all the comments, but wxGrid code is already awfully complicated and so I'd really like to make an effort to reduce additional increase of its complexity as much as possible.

vadz:
I've tested this PR under Mac, seems to work great, thanks!

image

include/wx/generic/gridsel.h
AliKet: Done

include/wx/generic/gridsel.h
AliKet: Done

include/wx/generic/gridsel.h
AliKet: Done

include/wx/generic/private/grid.h
AliKet: Done

include/wx/generic/private/grid.h
AliKet: > So the data structure here is a vector of points in all polygons stored as a single continuous vector? I.e. if we have 2 polygons P1 and P2, `m_points` contains P1_first, P1_second, ..., P1_last, P2_first, ..., P2_last?
> 

Yes.

> Is this done for efficiency reasons? Because for me a natural approach would be to just have `vector<vector<wxPoint>> m_polygons` and I wonder why is it not being used here?

Well, the data structure is modeled in such a way that we can use it to draw a poly-polygon using [DrawPolyPolygon()](https://docs.wxwidgets.org/3.2/classwx_d_c.html#ae760da96008debe42e18195d8ac8df3d) in a simple and straightforward way. I don't see how a `vector<vector<wxPoint>>` would simplify things here to be honest?

include/wx/generic/grid.h
AliKet: Done without adding `DisableOverlaySelectionByDefault()`.

include/wx/generic/gridsel.h
AliKet: `ComputePolyPolygon()` has been renamed to `ComputeSelectionShape()`. Because the new name seems strange to me now! I'm thinking of renaming it to `UpdateSelectionShape()` instead, what do you think ?

include/wx/generic/private/grid.h
vadz: Sorry, I forgot about this `DrawPolyPolygon()` convention, I thought the polygons would be drawn one by one, but if we use this function then storing it like this makes sense, thanks, even if using a vector of vectors would still definitely be simpler (because a polygon is a vector of points, and so many polygons is a vector of such vectors).

include/wx/generic/gridsel.h
vadz: Both `Compute` and `Update` seem fine to me, with maybe a slight preference for the latter, but mostly I just think that `SelectionShape()` is much better than `PolyPolygon` here.

Thanks!

AliKet:
@vadz Unrelated to this PR but I just noticed this function [wxGrid::DrawAllGridLines()](https://github.com/wxWidgets/wxWidgets/blob/7f5ca30e3eebbdb85222fbe329a5100b4da65804/include/wx/generic/grid.h#L1639) which is undocumented and not used or referenced anywhere and I'm wondering if it's safe to get rid of it?

vadz:
> @vadz Unrelated to this PR but I just noticed this function [wxGrid::DrawAllGridLines()](https://github.com/wxWidgets/wxWidgets/blob/7f5ca30e3eebbdb85222fbe329a5100b4da65804/include/wx/generic/grid.h#L1639) which is undocumented and not used or referenced anywhere and I'm wondering if it's safe to get rid of it?

Yes, it seems pretty useless to me, so I think we can remove it, thanks.

vadz:
Just to confirm: this is ok to merge already, right?

AliKet:
> Just to confirm: this is ok to merge already, right?

Yes.

DietmarSchwertberger:
@AliKet 
Hi!
I'm just testing the last wxPython version which now tracks the wxWidgets master branch.
Thanks for the overlay mode.
It looks good, but I would prefer to have a finer grained control on this feature.

E.g. when complete row(s) are selected, I don't want all the column headers to be highlighted.
For some grids I would like to have the overlay mode for the selected cell(s), but I don't want the headers to be highlighted for the current cell or the selection.
For some grids, I would like the current row and col label being highlighted, but only when the grid actually has the focus.

Maybe the active cell should be in blue as well if there is no selection. (I find it a bit inconsistent that the row and col header are blue like for a selection, but the cell is not.)

As of now, I find the implementation being a bit too distracting from other contents and I have to disable it.
Traditionally 'blue' means 'focused' to me (I'm on Windows). Maybe, the overlay colour should switch from blue to grey while not focused?

I'm not sure what would be the best way:
 - add arguments to `DisableOverlaySelection`
 - add style flags?

What do you think?

vadz:
> E.g. when complete row(s) are selected, I don't want all the column headers to be highlighted. For some grids I would like to have the overlay mode for the selected cell(s), but I don't want the headers to be highlighted for the current cell or the selection. For some grids, I would like the current row and col label being highlighted, but only when the grid actually has the focus.

I think the current implementation works as in spreadsheet programs, which means that this is probably the most familiar thing for most users (I've checked LibreOffice Calc. Excel and Google Sheets and they all do it like this).

> Maybe the active cell should be in blue as well if there is no selection. (I find it a bit inconsistent that the row and col header are blue like for a selection, but the cell is not.)
> 
> As of now, I find the implementation being a bit too distracting from other contents and I have to disable it. Traditionally 'blue' means 'focused' to me (I'm on Windows). Maybe, the overlay colour should switch from blue to grey while not focused?

Excel does use grey rather than blue, so maybe this colour should be platform-dependent and grey under Windows...

> I'm not sure what would be the best way:
> 
>     * add arguments to `DisableOverlaySelection`
> 
>     * add style flags?

 I'd rather avoid the use of style flags, but we could add some SetOverlaySelectionPreferences() (better name welcome) taking wxOverlaySelectionPreferences object (ditto) which would provide an extensible way to customize behaviour and colours. I don't plan doing this myself, but if you'd like to make a PR adding this, it would be welcome.

TIA!

DietmarSchwertberger:
Thanks for feedback. Currently I have two other projects, but after these I will look into this and experiment a bit.
Is there an example for a Preferences object? Otherwise I would prefer the method plus private properties for easier implementation. I would think that for any grid the preferences are set only once, so I'm not sure whether more abstraction would really add much value.

vadz:
I was thinking of a `wxSizerFlags`-like object, i.e. something that could be built using a builder pattern (a.k.a. chained function calls on the same object) and would contain all the customizable properties — this is tidier than having half a dozen or more of separate parameters/variables.

Issue #24535, opened 2024-05-15
[macOS] Wrong calculation of slider's vertical label positions (in case of a horizontal slider)


hwiesmann:
The slider's label positions are wrongly calculated:

1. the distance between the slider's knob and the label is insufficient as the text overlaps with the knob when the slider has its best size
2. the label vertical positions in case of a horizontal slider (same holds for a vertical slider) are not adjusted when the slider's vertical size is below its best size

The effects can be seen by using the image sample program and opening the "Image filters test" window. I have attached two screen shots showing the two shortcomings.

Remark: the slider's background has been set by myself to red (I just did some investigations of the root cause and therefore needed a visual feedback) and is not part of the original source code.

<img width="888" alt="Bildschirmfoto 2024-05-15 um 23 56 52" src="https://github.com/wxWidgets/wxWidgets/assets/518892/95ed7e8f-bf73-426a-97d0-0cfdfa060f7c">

This screenshot has been made when the window had been sufficiently enlarged.

<img width="888" alt="Bildschirmfoto 2024-05-15 um 23 57 08" src="https://github.com/wxWidgets/wxWidgets/assets/518892/24a0cdf0-9323-4175-b90a-f8c92a393957">

This screenshot has been made when the height of the window had been reduced.

The reasons for the shortcomings are:
 - wrong calculation of the distance between the slider's knob and the labels' texts
 - not taking the size of the slider into account

PR #24562, opened 2024-05-27
wxSlider layout improvements in wxOSX (#24531 & #24535)

### LINKED ISSUE #24531: [macOS] Alignment of a slider's current value looks strange
### Description
I have set the minimum value of the slider to -100 in the widgets sample (see screenshot attached). 
Bildschirmfoto 2024-05-14 um 22 33 54

The current value is zero. Nevertheless, the label is positioned to the left of the slider's knob. This looks strange. I do not know how it looks like in other OSes  but I prefer a center alignment.
Any objections or should I prepare a fix?

PS: Because of the odd ticks frequency it looks like the slider is not positioned at zero but this is actually the case.

hwiesmann:
Value label is now always centered (#24531) with respect to the slider's moving direction. The labels are adjusted with respect to their position perpendicular to the slider's moving direction in case the slider's across size is smaller than its best size (#24535).

### DIFF
--- a/src/osx/slider_osx.cpp
+++ b/src/osx/slider_osx.cpp
@@ -11,63 +11,63 @@
 
 #if wxUSE_SLIDER
 
-#include <iostream>
-
 #include "wx/slider.h"
 #include "wx/osx/private.h"
+#include "wx/utils.h"
 
 wxBEGIN_EVENT_TABLE(wxSlider, wxControl)
 wxEND_EVENT_TABLE()
 
 namespace
 {
-    // The dimensions of the different styles of sliders (from Aqua document)
+// The dimensions of the different styles of sliders (from Aqua document)
 #if wxOSX_USE_COCOA
-    const int wxSLIDER_DIMENSIONACROSS_WITHTICKMARKS = 28;
-    const int wxSLIDER_DIMENSIONACROSS               = 28;
+constexpr int wxSLIDER_DIMENSIONACROSS_WITHTICKMARKS = 28;
+constexpr int wxSLIDER_DIMENSIONACROSS               = 28;
 #else
-    const int wxSLIDER_DIMENSIONACROSS_WITHTICKMARKS = 24;
-    const int wxSLIDER_DIMENSIONACROSS               = 18;
+constexpr int wxSLIDER_DIMENSIONACROSS_WITHTICKMARKS = 24;
+constexpr int wxSLIDER_DIMENSIONACROSS               = 18;
 #endif
 
-    const int DISTANCE_SLIDER_TEXT = 5; // Distance between slider and text
+constexpr int DISTANCE_SLIDER_TEXT = 5; // Distance between slider and text
 
-    // Returns the best dimension of the slider perpendicular to the slider's moving direction
-    int GetBestSliderAcrossDimension(wxSlider const* slider)
-    {
-        return (((slider->GetWindowStyle() & wxSL_AUTOTICKS) != 0) ?
-                wxSLIDER_DIMENSIONACROSS_WITHTICKMARKS : wxSLIDER_DIMENSIONACROSS);
-    }
-    // Returns the slider's text position perpendicular to the slider's moving direction
-    /*
-     * When the slider's size prependicular to the slider's moving direction becomes
-     * smaller than the best size, the formula for determining the text position is
-     * not valid anymore. Therefor, the text is aligned under these conditions with
-     * respect to the lower (horizontal slider orientation), respectively to the
-     * right (vertical slider orientation) border of the slider.
-     * Remarks: There is still a small glitch in calculating the position, namely when
-     *          wxWidgets' calculation determines a size of zero in the perpendicular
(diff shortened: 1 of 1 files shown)


src/osx/slider_osx.cpp
```diff
@@ -11,23 +11,64 @@
 
 #if wxUSE_SLIDER
 
+#include <iostream>
```
vadz, line 14: AFAICS this isn't really needed.

src/osx/slider_osx.cpp
```diff
@@ -11,23 +11,64 @@
...
 #endif
 
-// Distance between slider and text
-#define wxSLIDER_BORDERTEXT 5
+    const int DISTANCE_SLIDER_TEXT = 5; // Distance between slider and text
+
+    // Returns the best dimension of the slider perpendicular to the slider's moving direction
+    int GetBestSliderAcrossDimension(wxSlider const* slider)
+    {
+        return (((slider->GetWindowStyle() & wxSL_AUTOTICKS) != 0) ?
```
vadz, line 38: This could use more readable `HasFlag()`.

src/osx/slider_osx.cpp
```diff
@@ -11,23 +11,64 @@
...
+    {
+        const wxSize size     = slider->GetSize();
+        const wxSize bestSize = slider->GetBestSize();
+
+
+        if (slider->GetWindowStyle() & wxSL_VERTICAL)
+            if (size.x >= bestSize.x)
+                return GetBestSliderAcrossDimension(slider)+DISTANCE_SLIDER_TEXT;
+            else
+                return MAX(0,size.x-textDimension.x);
```
vadz, line 65: What is this `MAX`? It results in build errors and I don't see where it is supposed to be defined... We have `wxMax()` in `wx/utils.h` and there is, of course, also `std::max()` and one of them should be used instead.

vadz (changes requested):
Thanks for the fixes but so many things have changed here that it's really difficult to understand what's going on. I guess we should just merge it anyhow to fix a user-visible problem but it would be great to separate changes like this in separate commits if possible.

Also, there are some more concrete problems that have to be fixed before this can be merged, could you please do it? TIA!

src/osx/slider_osx.cpp
hwiesmann: No, I have forgotten to delete this. It has been used for debugging.

src/osx/slider_osx.cpp
hwiesmann: Sorry, has been taken from a macOS system header. Will use wxMax instead.

src/osx/slider_osx.cpp
```diff
@@ -11,23 +11,64 @@
...
 
- // The dimensions of the different styles of sliders (from Aqua document)
+namespace
+{
+    // The dimensions of the different styles of sliders (from Aqua document)
 #if wxOSX_USE_COCOA
-    #define wxSLIDER_DIMENSIONACROSS_WITHTICKMARKS 28
-    #define wxSLIDER_DIMENSIONACROSS_ARROW 21
+    const int wxSLIDER_DIMENSIONACROSS_WITHTICKMARKS = 28;
+    const int wxSLIDER_DIMENSIONACROSS               = 28;
```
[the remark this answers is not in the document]
hwiesmann: I have checked the dimensions with Interface builder (20 for the slider itself plus 2 and 6 for the alignment around it). The old numbers come from an old macOS version, I suppose, as the naming indicates "arrow". But there isn't for a long time anymore an arrow for a slider.
I have also checked the commit. It seems to be that the source code has mainly been copied from Carbon.

src/osx/slider_osx.cpp
```diff
@@ -346,33 +383,25 @@ wxSize wxSlider::DoGetBestSize() const
...
-            textheight = maxtheight;
-        else
-            textheight = mintheight;
-
-        if (maxtwidth > mintwidth)
-            textwidth = maxtwidth;
-        else
-            textwidth = mintwidth;
+        textheight = MAX(mintheight,maxtheight);
+        textwidth  = MAX(mintwidth, maxtwidth);
```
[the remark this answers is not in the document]
hwiesmann: I would like to do it but the source code had to be basically be re-written. The source code was old inappropriate Carbon code (see my comments below). A lot had to be changed to make it compatible with Cocoa (with respect to alignment and outlook).

PS: There is some more old code inside that can be removed (it is not necessary to change it as it still works but makes the code more complicated than it actually is). I did not remove it because re-writing everything would have been easier. And I already feared that some comments would pop up because of all these changes.

vadz:
@csomor If you don't object, I'm going to apply this even if I'm not completely sure about a few things here myself.

csomor:
@vadz ok, go ahead, thanks :-)

vadz:
I've created #24609 with the same contents as here, so I'm closing this one, please let me know you see anything wrong with the other PR.

PR #24566, opened 2024-05-28
Fix testing for non-client children (fixes #24565)

### LINKED ISSUE #24565: [macOS] wxHasRealChildren in wincmn.cpp incorrectly determines real children
### Description
wxHasRealChildren only checks for scroll bars. The grow box is also a child and not a real child either but has been forgotten to check.

- wxWidgets version you use: latest master build

hwiesmann:
All non-client area children are covered by testing for real children.
Additionally, the macros related to HasRealChildren are now obsolete as HasRealChildren uses only methods that are available for all ports. This makes it also less error-prone for future changes with respect to children.

### DIFF
--- a/src/common/wincmn.cpp
+++ b/src/common/wincmn.cpp
@@ -634,26 +634,16 @@ void wxWindowBase::FitInside()
 }
 
 // On Mac, scrollbars are explicitly children.
-#if defined( __WXMAC__ ) && !defined(__WXUNIVERSAL__)
 static bool wxHasRealChildren(const wxWindowBase* win)
 {
-    int realChildCount = 0;
-
-    for ( wxWindowList::compatibility_iterator node = win->GetChildren().GetFirst();
-          node;
-          node = node->GetNext() )
+    for ( const auto child : win->GetChildren() )
     {
-        wxWindow *win = node->GetData();
-        if ( !win->IsTopLevel() && win->IsShown()
-#if wxUSE_SCROLLBAR
-            && !wxDynamicCast(win, wxScrollBar)
-#endif
-            )
-            realChildCount ++;
+        if ( !child->IsTopLevel() && child->IsShown()
+            && win->IsClientAreaChild(child) )
+            return true;
     }
-    return (realChildCount > 0);
+    return false;
 }
-#endif
 
 void wxWindowBase::InvalidateBestSize()
 {
@@ -713,11 +703,7 @@ wxSize wxWindowBase::DoGetBestSize() const
         best = wxSize(maxX, maxY);
     }
 #endif // wxUSE_CONSTRAINTS
-    else if ( !GetChildren().empty()
-#if defined( __WXMAC__ ) && !defined(__WXUNIVERSAL__)
-              && wxHasRealChildren(this)
-#endif
-              )
+    else if ( wxHasRealChildren(this) )
     {
         // our minimal acceptable size is such that all our visible child
         // windows fit inside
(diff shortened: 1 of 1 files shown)


src/common/wincmn.cpp
```diff
@@ -634,26 +634,16 @@ void wxWindowBase::FitInside()
 }
 
 // On Mac, scrollbars are explicitly children.
-#if defined( __WXMAC__ ) && !defined(__WXUNIVERSAL__)
```
vadz, line 637: I'm not sure why is this not Mac-only any more?

vadz (commented):
Thanks, we do need to fix this for Mac (even though IMO the ideal fix would still be to avoid putting non client area children in `m_children` to begin with, as discussed in another issue recently...), but I'm not sure why did you decide to generalize this to all ports?

src/common/wincmn.cpp
hwiesmann: There are a few reasons:

1.  There is no (relevant) performance hit.
2. IsClientArea is a public method for all ports. I have not checked whether this is Mac-only or if this method will be Mac-only till the end of wxWidgets. And I do not have to worry: the modification will work in any case (as long as IsClientAreaChild exists).
3. IMHO it is difficult to maintain exceptions in a source code that are related to specific ports only. This can be seen from the fact that in this case a bug has been introduced. Perhaps, somebody was not familiar with the Mac port or forgot that grow boxes are also children. Or the grow box became a child after the implementation of scrollbars as children and the person making the change did not know that somewhere else in the code there is at least one assumption that scrollbars are children but not real ones (actually, the same bug occurred twice(!) in the same source code file, in the second case no port distinction has been made (cast to scrollbars only) and I detected it by pure coincidence).

PS: If scrollbars or grow windows are children or not is another topic, I think. And you agree, I believe, 
fixing this may take much longer. This is an appropriate fix till then.

src/common/wincmn.cpp
```diff
@@ -729,10 +715,10 @@ wxSize wxWindowBase::DoGetBestSize() const
               node = node->GetNext() )
         {
             wxWindow *win = node->GetData();
-            if ( win->IsTopLevel()
-                    || !win->IsShown()
+            if ( win->IsTopLevel() || !win->IsShown()
+                || !IsClientAreaChild(win)
 #if wxUSE_STATUSBAR
-                        || wxDynamicCast(win, wxStatusBar)
+                || wxDynamicCast(win, wxStatusBar)
```
vadz, line 721: I think the new test makes the test for `wxStatusBar` redundant as it's only not part of the client area in `wxFrame`, which overrides `IsClientAreaChild()` to exclude it.

What bothers me is that we don't test for `wxToolBar` here, although it looks like we ought to, for the same reason. And in the new version we will exclude it, which might not be correct, if we want to use its width.

src/common/wincmn.cpp
```diff
@@ -713,11 +703,7 @@ wxSize wxWindowBase::DoGetBestSize() const
         best = wxSize(maxX, maxY);
     }
 #endif // wxUSE_CONSTRAINTS
-    else if ( !GetChildren().empty()
-#if defined( __WXMAC__ ) && !defined(__WXUNIVERSAL__)
-              && wxHasRealChildren(this)
-#endif
-              )
+    else if ( wxHasRealChildren(this) )
```
vadz, line 706: Looking at this code once again, I think we don't actually need this at all as we redo the same tests in the loop inside anyhow.

So we could just remember if we have any children there and fallback to the code in the `else` branch if not. IMO this would make things more clear, as right now we have similar but not quite identical checks in 2 places.

vadz (commented):
As long as we're improving this code — and not just replacing the dynamic cast with `IsClientAreaChild()` call — I'd like to go further and get rid of `wxHasRealChildren()` entirely.

What do you think?

vadz:
@hwiesmann What do you think about the suggestion above?

vadz:
I'm closing this one as I've finally created #24708 which seems like a simpler solution to me. Please let me know if you see any problems with that one. TIA!

PR #24567, opened 2024-05-28
Use bilinear quality when downscaling in wxBitmapHelpers::Rescale.

dsa-t:
This gives the best quality when downscaling.

See KiCad project manager icons:

Nearest:
nearest

Bilinear:
bilinear

Bicubic:
bicubic

Box average:
box

### DIFF
--- a/src/common/bmpbase.cpp
+++ b/src/common/bmpbase.cpp
@@ -79,7 +79,14 @@ void wxBitmapHelpers::Rescale(wxBitmap& bmp, const wxSize& sizeNeeded)
     // the icons for which this function is often used. It's also consistent
     // with what wxDC::DrawBitmap() does, i.e. the fallback method below.
     wxImage img = bmp.ConvertToImage();
-    img.Rescale(sizeNeeded.x, sizeNeeded.y, wxIMAGE_QUALITY_NEAREST);
+    wxImageResizeQuality quality = wxIMAGE_QUALITY_NEAREST;
+
+    // If we're downscaling, bilinear interpolation gives best results.
+    if( img.GetWidth() > sizeNeeded.x && img.GetHeight() > sizeNeeded.y )
+        quality = wxIMAGE_QUALITY_BILINEAR;
+
+    img.Rescale(sizeNeeded.x, sizeNeeded.y, quality);
+
     bmp = wxBitmap(img);
 #else // !wxUSE_IMAGE
     // Fallback method of scaling the bitmap


vadz:
Curious, I'd expect `wxIMAGE_QUALITY_BOX_AVERAGE` (a.k.a. `wxIMAGE_QUALITY_HIGH` in this case) to produce best results when downscaling, have you tried this one?

Also, if we really decide that some other method (either bilinear or box average or bicubic) is definitely best than the "nearest" one, I think we should change `wxIMAGE_QUALITY_NORMAL` to use it by default for not too big images because I have trouble imagining modern CPUs taking noticeable time running any of them for the typical icon sizes, i.e. up to 256 pixels maybe.

Finally I thought that the best algorithm to use for downsampling was Lanczos filter but a quick web search found https://johncostella.com/magic/ that I hadn't heard before but which is supposed to be even better and faster. Maybe someone could add support for it to `wxImage`...

dsa-t:
Added a screenshot with `wxIMAGE_QUALITY_BOX_AVERAGE`, it looks like nearest, but blurry.

dsa-t:
Bicubic is similar to bilinear, but very slightly more blurry.

vadz:
I'd be for changing `wxIMAGE_QUALITY_NORMAL` to behave like `wxIMAGE_QUALITY_BILINEAR` when scaling small images then. Although we probably need to change `wxIMAGE_QUALITY_HIGH` to use bilinear when downscaling too then, as it shouldn't result in worse images than `NORMAL`.

What do the other think?

dsa-t:
I think it's best to **not** use bilinear filtering in `wxIMAGE_QUALITY_NORMAL` for upscaling, otherwise images will be blurry.

vadz:
I've realized that I don't actually know which images are we talking about here: the icons in the left-most column or the ones in the "Project files" pane? I thought it was the latter ones but I don't understand how does "Schematic Editor" icon lose its bottom-right quadrant contents entirely then, this shouldn't happen when scaling AFAICS, so maybe I'm just misunderstanding what's going on here?

Anyhow, my suggestion would finally probably be to do this:

```diff
diff --git a/src/common/image.cpp b/src/common/image.cpp
index 0bf483822f..3230d659f7 100644
--- a/src/common/image.cpp
+++ b/src/common/image.cpp
@@ -487,7 +487,7 @@ wxImage::Scale( int width, int height, wxImageResizeQuality quality ) const
 
         case wxIMAGE_QUALITY_HIGH:
             image = width < old_width && height < old_height
-                        ? ResampleBox(width, height)
+                        ? ResampleBilinear(width, height)
                         : ResampleBicubic(width, height);
             break;
     }
diff --git a/src/common/bmpbase.cpp b/src/common/bmpbase.cpp
index ee9feda5d1..fa7e0a4bd6 100644
--- a/src/common/bmpbase.cpp
+++ b/src/common/bmpbase.cpp
@@ -79,7 +79,7 @@ void wxBitmapHelpers::Rescale(wxBitmap& bmp, const wxSize& sizeNeeded)
     // the icons for which this function is often used. It's also consistent
     // with what wxDC::DrawBitmap() does, i.e. the fallback method below.
     wxImage img = bmp.ConvertToImage();
-    img.Rescale(sizeNeeded.x, sizeNeeded.y, wxIMAGE_QUALITY_NEAREST);
+    img.Rescale(sizeNeeded.x, sizeNeeded.y, wxIMAGE_QUALITY_HIGH);
     bmp = wxBitmap(img);
 #else // !wxUSE_IMAGE
     // Fallback method of scaling the bitmap
```

but the comment in `wxBitmapHelpers::Rescale()` says that `wxIMAGE_QUALITY_NEAREST` is best for preserving sharp lines and I indeed remember that some images with geometric shapes looked very poor after being downscaled using `wxIMAGE_QUALITY_HIGH`, so I'm not quite sure if the latter change is correct.

dsa-t:
> I don't actually know which images are we talking about here

The difference is in the launcher panel icons:

image

dsa-t:
> how does "Schematic Editor" icon lose its bottom-right quadrant contents entirely then

I'm not entirely sure what this refers to.

dsa-t:
> but the comment in `wxBitmapHelpers::Rescale()` says that `wxIMAGE_QUALITY_NEAREST` is best for preserving sharp lines

It's best for preserving sharp lines when upscaling.

> I indeed remember that some images with geometric shapes looked very poor after being downscaled using `wxIMAGE_QUALITY_HIGH`, so I'm not quite sure if the latter change is correct.

Probably because `ResampleBox` was being used when downscaling.

dsa-t:
But upscaling with `wxIMAGE_QUALITY_NEAREST` is not suitable for photos, for example.
And upscaling icons with `ResampleBicubic` will make them blurry.

dsa-t:
So the current `wxIMAGE_QUALITY_HIGH` may be fine for photos, but I haven't tested it.

vadz:
> The difference is in the launcher panel icons:

Wait, but this is about downscaling and these icons are bigger than the small versions in the "Project View". Does this mean that they're downscaled from even bigger versions not shown here?

> So the current `wxIMAGE_QUALITY_HIGH` may be fine for photos, but I haven't tested it.

Maybe we need `wxIMAGE_QUALITY_ICON` and `wxIMAGE_QUALITY_PHOTO` or something like this instead of normal/high?

dsa-t:
> Wait, but this is about downscaling and these icons are bigger than the small versions in the "Project View". Does this mean that they're downscaled from even bigger versions not shown here?

Yes, they're downscaled from 128x128 icons:

icon_eeschema_128 icon_libedit_128 icon_pcbnew_128 icon_modedit_128 icon_bitmap2component_128 icon_pcm_128


> Maybe we need `wxIMAGE_QUALITY_ICON` and `wxIMAGE_QUALITY_PHOTO` or something like this instead of normal/high?

Good idea, but not backportable to 3.2.

vadz:
We really need to take some decision here but I still don't know what to do. What about this proposal:

1. In master, change `wxIMAGE_QUALITY_NORMAL` to have a value different from `wxIMAGE_QUALITY_NEAREST` (i.e. `5`) and make it behave as this PR, i.e. use `NEAREST` when upscaling and `BILINEAR` when downscaling.
2. In 3.2, apply this PR, basically.

Any objections/better ideas?

dsa-t:
> 1. In master, change `wxIMAGE_QUALITY_NORMAL` to have a value different from `wxIMAGE_QUALITY_NEAREST` (i.e. `5`) and make it behave as this PR, i.e. use `NEAREST` when upscaling and `BILINEAR` when downscaling.

I'd preferred explicit ICON/PHOTO separation better (someone would have to compare the quality), but this seems fine to me.

vadz:
I'm not sure if it's really always true that bilinear/bicubic works better for icons/photos, so for now I've just created #24685, please let me know if you see any problems with it. TIA!

dsa-t:
Scaling horse by 1/2.8:

image

Bilinear looks the best to me.
Bicubic is slightly blurrier;
Box average is much more blurry, Maybe there's a bug in the algorithm?

vadz:
I think we're going to need some objective way of determining which image is more blurry (Internet has a few suggestions...) because I have trouble deciding it myself.

And personally I also care much more about icons (with straight lines and sharp contrast) than photos.

dsa-t:
I also noticed that `wxIMAGE_QUALITY_NEAREST` uses `wxImage::ShrinkBy` algorithm for integral scale factors, which looks better than bilinear.
1/2 scaling:

image

Looks like `ShrinkBy` is the same algorithm as `ResampleBox`.

dsa-t:
In KiCad with 125% display scaling, the toolbar icons are scaled from 32px to 30px. This doesn't look good with the nearest neighbor algorithm. Any chance to have bilinear in 3.2?
Screenshot 2025-02-14 092723

Though need to check what's happening with https://github.com/wxWidgets/wxWidgets/issues/24774

vadz:
Do you mean cherry-pick just the change to `src/common/image.cpp` in 46833ebaae2da4f22e15b600942d2dcc8786132c (without the `case wxIMAGE_QUALITY_FAST:` line)? If so, yes, this is doable, but I am just not sure at all that this is not going to result in some regressions... Perhaps we could add a system option to do it, e.g. some `image.use-bilinear-as-normal`? I normally hate those, but it would provide an escape hatch for KiCad if you really need it...

But I'm actually more surprised that any resizing is being done at all here, I thought we would just use normal icons in 125% case. And why are we resizing them _down_ in higher DPI?

dsa-t:
> But I'm actually more surprised that any resizing is being done at all here, I thought we would just use normal icons in 125% case.

That's what our `WX_AUI_TOOLBAR_ART` does. I haven't looked at the default `wxAuiDefaultToolBarArt`.

> And why are we resizing them _down_ in higher DPI?

KiCad provides 24, 32, 48, 64 sizes in the bitmap bundle for these icons, so 32 is picked here.


> Do you mean cherry-pick just the change to `src/common/image.cpp` in [46833eb](https://github.com/wxWidgets/wxWidgets/commit/46833ebaae2da4f22e15b600942d2dcc8786132c) (without the `case wxIMAGE_QUALITY_FAST:` line)? If so, yes, this is doable, but I am just not sure at all that this is not going to result in some regressions... Perhaps we could add a system option to do it, e.g. some `image.use-bilinear-as-normal`? I normally hate those, but it would provide an escape hatch for KiCad if you really need it...

What if we do this for 3.2?

```diff
diff --git a/src/common/bmpbase.cpp b/src/common/bmpbase.cpp
index 558417c5aa..a66424a896 100644
--- a/src/common/bmpbase.cpp
+++ b/src/common/bmpbase.cpp
@@ -74,7 +74,7 @@ void wxBitmapHelpers::Rescale(wxBitmap& bmp, const wxSize& sizeNeeded)
     // the icons for which this function is often used. It's also consistent
     // with what wxDC::DrawBitmap() does, i.e. the fallback method below.
     wxImage img = bmp.ConvertToImage();
-    img.Rescale(sizeNeeded.x, sizeNeeded.y, wxIMAGE_QUALITY_NEAREST);
+    img.Rescale(sizeNeeded.x, sizeNeeded.y, wxIMAGE_QUALITY_HIGH);
     bmp = wxBitmap(img);
 #else // !wxUSE_IMAGE
     // Fallback method of scaling the bitmap
diff --git a/src/common/image.cpp b/src/common/image.cpp
index 96bc1a86f5..b3917e7e0c 100644
--- a/src/common/image.cpp
+++ b/src/common/image.cpp
@@ -485,10 +485,26 @@ wxImage::Scale( int width, int height, wxImageResizeQuality quality ) const
             break;
 
         case wxIMAGE_QUALITY_HIGH:
-            image = width < old_width && height < old_height
-                        ? ResampleBox(width, height)
-                        : ResampleBicubic(width, height);
+        {
+            double shrinkFactorX = double(old_width) / width;
+            double shrinkFactorY = double(old_height) / height;
+
+            // When downsampling, use bilinear algorithm for pre-scaling and
+            // box average for integer part
+            if ( shrinkFactorX > 1 && shrinkFactorY > 1 )
+            {
+                int shrinkInt(wxMin(shrinkFactorX, shrinkFactorY));
+
+                image = ResampleBilinear(width * shrinkInt, height * shrinkInt);
+                image = image.ResampleBox(width, height);
+            }
+            else
+            {
+                image = ResampleBox(width, height);
+            }
+
             break;
+        }
     }
 
     // If the original image has a mask, apply the mask to the new image

```

dsa-t:
Or just this:

```diff
diff --git a/src/common/bmpbase.cpp b/src/common/bmpbase.cpp
index 558417c5aa..cb713f09fb 100644
--- a/src/common/bmpbase.cpp
+++ b/src/common/bmpbase.cpp
@@ -70,11 +70,26 @@ void wxBitmapHelpers::Rescale(wxBitmap& bmp, const wxSize& sizeNeeded)
     wxCHECK_RET( sizeNeeded.IsFullySpecified(), wxS("New size must be given") );
 
 #if wxUSE_IMAGE
-    // Note that we use "nearest" rescale mode here to preserve sharp edges in
-    // the icons for which this function is often used. It's also consistent
-    // with what wxDC::DrawBitmap() does, i.e. the fallback method below.
+
     wxImage img = bmp.ConvertToImage();
-    img.Rescale(sizeNeeded.x, sizeNeeded.y, wxIMAGE_QUALITY_NEAREST);
+
+    double shrinkFactorX = double(bmp.GetWidth()) / sizeNeeded.x;
+    double shrinkFactorY = double(bmp.GetHeight()) / sizeNeeded.y;
+
+    // When downsampling, use bilinear algorithm for pre-scaling and
+    // box average for integer part
+    if ( shrinkFactorX > 1 && shrinkFactorY > 1 )
+    {
+        int shrinkInt(wxMin(shrinkFactorX, shrinkFactorY));
+
+        img.Rescale(sizeNeeded.x * shrinkInt, sizeNeeded.y * shrinkInt, wxIMAGE_QUALITY_BILINEAR);
+        img.Rescale(sizeNeeded.x, sizeNeeded.y, wxIMAGE_QUALITY_BOX_AVERAGE);
+    }
+    else
+    {
+        img.Rescale(sizeNeeded.x, sizeNeeded.y, wxIMAGE_QUALITY_BOX_AVERAGE);
+    }
+
     bmp = wxBitmap(img);
 #else // !wxUSE_IMAGE
     // Fallback method of scaling the bitmap

```

dsa-t:
For 3.2 see: https://github.com/wxWidgets/wxWidgets/pull/25170

Issue #24568, opened 2024-05-28
Fix for issue #24528 is faulty


hwiesmann:
### Description
The fix for determining if a page has real children is faulty. This can be seen when selecting a page, selecting another page and going back to the previously selected page.

The reason is that the method IsClientAreaChild is applied incorrectly. It should read

`if ( curPage->IsClientAreaChild(child) )`

### Platform and version information

- wxWidgets version you use: lastest master branch
- wxWidgets port you use: macOS 14.4.1

PR #24569, opened 2024-05-28
BugFix #24568 for partially wrong fix of issue #24528

hwiesmann:
The previously applied patch uses IsClientAreaChild incorrectly. IsClientAreaChild should be called from the parent and a window has a real child if it is in the client area.


vadz:
Oops, sorry for messing this up. Will push soon.

PR #24572, opened 2024-05-29
Fix __WXFUNCTION_SIG__ under Apple clang 14

Blake-Madden:
Don't use C++20 features. Source location feature isn't always available, and `__has_cpp_attribute` is quirky under some compilers and shouldn't be used here (IMO).

Relying on `__FUNC_SIG__` and `__PRETTY_FUNCTION__` covers most compilers anyway (MSVC, GCC, clang, Apple clang, ICC).

### DIFF
--- a/include/wx/cpp.h
+++ b/include/wx/cpp.h
@@ -120,11 +120,7 @@
     Falls back to the function name (i.e., __func__) if not available.
  */
 #ifndef __WXFUNCTION_SIG__
-    #if __cplusplus >= 202002L
-        #include <source_location>
-        // actually displays the signature, not just the name
-        #define __WXFUNCTION_SIG__ std::source_location::current().function_name()
-    #elif defined(__VISUALC__)
+    #if defined(__VISUALC__)
         #define __WXFUNCTION_SIG__ __FUNCSIG__
     #elif defined(__GNUG__)
         #define __WXFUNCTION_SIG__ __PRETTY_FUNCTION__


rcane:
Since `source_location` is a library feature and not a language feature, the test `#if __cplusplus >= 202002L` is indeed not sufficient to detect its availability. Also, `__cplusplus` is not set properly under MSVC. So we would need to use `wxCHECK_CXX_STD` instead.

If we wanted to keep using `source_location`, the test should look like this:
```cpp
#if !defined(__WXFUNCTION_SIG__) && wxCHECK_CXX_STD(202002L) && __has_include(<version>)
    #include <version>
    #if defined(__cpp_lib_source_location)
        #include <source_location>
        // actually displays the signature, not just the name
        #define __WXFUNCTION_SIG__ std::source_location::current().function_name()
    #endif
#endif
// If source_location was not available...
#ifndef __WXFUNCTION_SIG__
    #if defined(__VISUALC__)
        #define __WXFUNCTION_SIG__ __FUNCSIG__
    #elif defined(__GNUG__)
        #define __WXFUNCTION_SIG__ __PRETTY_FUNCTION__
    #endif
#endif
```
This would detect the missing support in clang 14 (it was actually added in clang 15).

imciner2:
`wxCHECK_CXX_STD(202002L)`

I don't think you need to use this if you are using the feature-test macro from the header. The version header will only define the macro when the standard is appropriate.

Blake-Madden:
I don't think I really need `source_location`, it certainly isn't worth all the hoops to jump through. `__PRETTY_FUNCTION__` is widely available across many compilers.

rcane:
`wxCHECK_CXX_STD(202002L)` is actually necessary because `__has_include` is a C++17 feature which not every compiler supports (I tested this with godbolt). But the version to test could be reduced to `201704L`.

But as @Blake-Madden said (and the output I saw with godbolt confirms that) `__PRETTY_FUNCTION__` produces the same strings as `source_location`. So all this C++20 stuff (albeit cooler ;-)) is not really necessary.

vadz:
Thanks, I agree that this is the pragmatic solution for now. We should use `std::source_location` some day but not necessarily right now.

PR #24573, opened 2024-05-29
wxQt fixes for HiDPI, wxTextCtrl size, wxList cell text alignment, wxCursor hotspot.

dsa-t:


dsa-t:
Not sure why the wxQt job was skipped...

dsa-t:
Tests are now passing.

AliKet:
I don't have any real experience with HiDPI (not even under Qt) but the changes here LGTM in particular the tests are successful, which is great. thanks.

src/qt/listctrl.cpp
```diff
@@ -230,7 +230,7 @@ class wxQtListModel : public QAbstractTableModel
                     : QVariant();
 
             case Qt::TextAlignmentRole:
-                return columnItem.m_align;
+                return QVariant((int)columnItem.m_align);
```
vadz, line 233: Sorry, what does this fix? It seems non-obvious why is this better than relying on an implicit conversion to `QVariant`, so a comment explaining this would be useful.

vadz (approved):
Thanks, I didn't test this at all but the changes look good to me, there is just one minor question.

src/qt/listctrl.cpp
dsa-t: Because the type has changed, without this patch, there's warnings with default C++ version:

```
...\src\qt\listctrl.cpp(233): warning C4927: illegal conversion; more than one user-defined conversion has been implicitly applied
  ...\src\qt\listctrl.cpp(233): note: while calling the constructor 'QVariant::QVariant(int)'
  ...\Qt\5.15.2\msvc2019_64\include\QtCore/qvariant.h(219): note: see declaration of 'QVariant::QVariant'
```

on C++20 there's errors:

```
...\src\qt\listctrl.cpp(233): error C2440: 'return': cannot convert from 'const Qt::Alignment' to 'QVariant'
  ...\src\qt\listctrl.cpp(233): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
```

I just updated it to use `static_cast<int>` instead.

vadz: Thanks for the explanation! Ideal would be to use `std::to_underlying` as it's more readable but it's only in C++23... Maybe we should provide a fallback definition for it in some private header for the future.

PR #24574, opened 2024-05-29
Replace __FUNCTION__ with __func__, remove __FUNCTION__ warnings from docs

Blake-Madden:
`__func__` is used internally for logging and assert features now, so remove warnings about `__FUNCTION__` being used and returning `nullptr`,


vadz:
Looks good, thanks!

PR #24576, opened 2024-05-29
WrapSizer sample source code improvement

hwiesmann:
Windows added to static box sizers should have the static box as their parent.


vadz:

PR #24345, opened 2024-02-21
Semi-support Qt6 for wxQt port.

dsa-t:
Gets wxQt with Qt6 building with CMake, and samples (widgets, cube, pyramid) running.

Sort of implements https://github.com/wxWidgets/wxWidgets/issues/24208.

`glib-2.0` is needed to work around removal of `QAbstractEventDispatcher::hasPendingEvents()` in Qt6.

`wxQtEventSignalHandler::HandleDestroyedSignal` is removed due to assert when using Qt 6.6.2 linux desktop installed via aqt. When using distro's default `6.2.4+dfsg-2ubuntu1.1`, it just hangs.
Scenario: in `widgets` sample, select last entry in "All controls", then hold Up arrow key.

<details><summary>Assert details</summary>
<p>

`ASSERT failure in QTreeWidget: "Called object is not of the correct type (class destructor may have already run)", file .../qt/6.6.2/gcc_64/include/QtCore/qobjectdefs_impl.h, line 129`

```quote
libc.so.6!__pthread_kill_implementation(int no_tid, int signo, pthread_t threadid) (pthread_kill.c:44)
libc.so.6!__pthread_kill_internal(int signo, pthread_t threadid) (pthread_kill.c:78)
libc.so.6!__GI___pthread_kill(pthread_t threadid, int signo) (pthread_kill.c:89)
libc.so.6!__GI_raise(int sig) (raise.c:26)
libc.so.6!__GI_abort() (abort.c:79)
libQt6Core.so.6!qAbort() (Unknown Source:0)
libQt6Core.so.6![Unknown/Just-In-Time compiled code] (Unknown Source:0)
libQt6Core.so.6!QMessageLogger::fatal(char const*, ...) const (Unknown Source:0)
libQt6Core.so.6!qt_assert_x(char const*, char const*, char const*, int) (Unknown Source:0)
libwx_qtu_core-3.3.so.0!QtPrivate::assertObjectType<wxQtEventSignalHandler<QTreeWidget, wxTreeCtrl> >(QObject * o) (/home/alex/qt/6.6.2/gcc_64/include/QtCore/qobjectdefs_impl.h:129)
[...]

PR #24577, opened 2024-05-29
 wxQt: switch to QOpenGLWidget for wxGLCanvas.

### LINKED ISSUE #24208: Consider using QOpenGLWidget instead of QGLWidget under wxQt
### Description
Quoting the Qt documentation [QGLWidget](https://doc.qt.io/qt-5/qglwidget.html):
>Note: This class is part of the legacy [Qt OpenGL](https://doc.qt.io/qt-5/qtopengl-index.html) >module and, like the other QGL classes, should be avoided in the new applications. Instead, >starting from Qt 5.4, prefer using [QOpenGLWidget](https://doc.qt.io/qt-5/qopenglwidget.html) and the QOpenGL classes.

dsa-t:
Switches from QGLWidget to QOpenGLWidget, since the former stopped working on my Manjaro/KDE6/Wayland install (only showing a transparent surface).

Also it was deprecated (and removed in Qt 6).

QOpenGLWidget was introduced in Qt 5.4; I doubt anyone would want to use older Qt versions.

Parts from https://github.com/wxWidgets/wxWidgets/pull/24345
Closes https://github.com/wxWidgets/wxWidgets/issues/24208

### DIFF
--- a/src/qt/glcanvas.cpp
+++ b/src/qt/glcanvas.cpp
@@ -12,20 +12,20 @@
 #include "wx/qt/private/winevent.h"
 #include "wx/glcanvas.h"
 
-#include <QtOpenGL/QGLWidget>
+#include <QOpenGLWidget>
+#include <QSurfaceFormat>
 #include <QtWidgets/QGestureRecognizer>
 #include <QtWidgets/QGestureEvent>
 
 wxGCC_WARNING_SUPPRESS(unused-parameter)
 
-class wxQtGLWidget : public wxQtEventSignalHandler< QGLWidget, wxGLCanvas >
+class wxQtGLWidget : public wxQtEventSignalHandler< QOpenGLWidget, wxGLCanvas >
 {
 public:
-    wxQtGLWidget(wxWindow *parent, wxGLCanvas *handler, QGLFormat format)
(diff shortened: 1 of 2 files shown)


src/qt/glcanvas.cpp
```diff
@@ -490,7 +487,7 @@ bool wxGLCanvas::Create(wxWindow *parent,
 
 bool wxGLCanvas::SwapBuffers()
 {
-    static_cast<QGLWidget *>(m_qtWindow)->swapBuffers();
+    // Not possible
```
vadz, line 490: Wait, what? How does anything gets drawn then?

vadz (approved):
Thanks, no real comments but I just don't understand how can this work without every swapping buffers. I guess it does, but... how?

src/qt/glcanvas.cpp
dsa-t: Everything within `paintGL` renders onto a FBO, that is then converted to texture and composited by Qt.

dsa-t: "Not possible", as in :

>Triggering a buffer swap just for the QOpenGLWidget is not possible since there is no real, onscreen native surface for it. It is up to the widget stack to manage composition and buffer swaps on the gui thread. ...

PR #24582, opened 2024-05-30
Cygwin 64 bit build fixes for wx3.0

stahta01:
Two patches from 3.2 branch applied to fix building Cygwin 64 bit in 3.0 branch; the descriptions look okay to me; but, I think they need reviewed.

The local configure command used to test under Cygwin
`export CXXFLAGS="-std=gnu++14" && mkdir -p build-cygwin && cd build-cygwin && ../configure --with-msw --disable-shared --enable-unicode
`

Tim S.

### DIFF
--- a/include/wx/private/socket.h
+++ b/include/wx/private/socket.h
@@ -66,6 +66,15 @@
     #include <sys/time.h>   // for timeval
 #endif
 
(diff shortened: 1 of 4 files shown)


vadz:
I've pushed this now, thanks.

PR #24584, opened 2024-06-01
Base wxUniChar upon char32_t

vvs31415:
char32_t is a C++11 basic type designed to hold the Unicode chars.

### DIFF
--- a/include/wx/defs.h
+++ b/include/wx/defs.h
@@ -1027,6 +1027,7 @@ typedef double wxDouble;
     m(unsigned short) \
     m(int) \
     m(unsigned int) \
+    m(char32_t) \
     m(long) \
     m(unsigned long) \
     wxIF_LONG_LONG_TYPE( m(wxLongLong_t) ) \
@@ -1057,6 +1058,7 @@ typedef double wxDouble;
     m(unsigned short, arg) \
     m(int, arg) \
     m(unsigned int, arg) \
+    m(char32_t, arg) \
     m(long, arg) \
     m(unsigned long, arg) \
     wxIF_LONG_LONG_TYPE( m(wxLongLong_t, arg) ) \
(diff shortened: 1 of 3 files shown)


vadz:
Could you please explain what is the benefit of this change?

vvs31415:
- wxUniChar is interchangeable with all character types; since C++ support should be added for new types char16_t and char32_t.
- The char32_t is the proper standard type for holding Unicode characters, not some typedef.
- This is initial step for adding similar support for wxString.

vadz:
Sorry, I don't find these reasons compelling because I still don't see any benefit in practice and I'm especially wary of any big changes to `wxString`. This class is basically frozen and we shouldn't change it any more unless really necessary.

Unless other people are in favour of applying this and/or give some examples where this would really make a difference in practice I would be against applying this.

vvs31415:
These legitimate code does not compile:
    wxString s1(u"abc");
    wxString s2(U"abc");

vadz:
> These legitimate code does not compile: wxString s1(u"abc"); wxString s2(U"abc");

Good point, thanks. But does anybody actually use UTF-{16,32} literals? I definitely don't myself and I have trouble seeing what would they be useful for, to be honest. Still, if people need them, maybe we should indeed add support for them.

vvs31415:
Maybe the people don't use them because many frameworks do not support these new types, even after a decade. WX could help the users who want to use the proper Unicode types,

vvs31415:
wchar_t is used on Unix for full Unicode strings, but not on Windows, which is confusing.
char32_t is the same on all platforms.

vadz:
> wchar_t is used on Unix for full Unicode strings, but not on Windows, which is confusing. char32_t is the same on all platforms.

I know this, of course, but in practice everybody is using UTF-8 anyhow (and less we speak about `char8_t`, the better), so I just don't know why would you want to use UTF-32 literals. And I'm pretty sure UTF-16 ones are useless.

vvs31415:
Closing for the lack of interest

PR #24589, opened 2024-06-02
wxQt: support modern-style wxFileDialog, wxDirDialog on Windows.

dsa-t:
Uses the MSW implementation that allows adding custom controls to modern-style dialogs. Qt doesn't have an API for this functionality.

### DIFF
--- a/include/wx/qt/dirdlg.h
+++ b/include/wx/qt/dirdlg.h
@@ -8,6 +8,11 @@
 #ifndef _WX_QT_DIRDLG_H_
 #define _WX_QT_DIRDLG_H_
 
+// Use MSW implementation to support custom controls in modern-style dialogs.
+#ifdef _WIN32
+#include <wx/msw/dirdlg.h>
+#else
+
 class QFileDialog;
 
 class WXDLLIMPEXP_CORE wxDirDialog : public wxDirDialogBase
@@ -42,5 +47,7 @@ private:
 
     wxDECLARE_DYNAMIC_CLASS(wxDirDialog);
 };
+
+#endif // _WIN32
 
 #endif // _WX_QT_DIRDLG_H_
--- a/include/wx/qt/dirdlg.h
+++ b/include/wx/qt/dirdlg.h
@@ -9,7 +9,7 @@
 #define _WX_QT_DIRDLG_H_
 
 // Use MSW implementation to support custom controls in modern-style dialogs.
-#ifdef _WIN32
+#ifdef __WINDOWS__
 #include <wx/msw/dirdlg.h>
 #else
 
@@ -48,6 +48,6 @@ private:
     wxDECLARE_DYNAMIC_CLASS(wxDirDialog);
 };
 
-#endif // _WIN32
+#endif // __WINDOWS__
 
 #endif // _WX_QT_DIRDLG_H_
--- a/include/wx/qt/filedlg.h
+++ b/include/wx/qt/filedlg.h
@@ -8,6 +8,11 @@
 #ifndef _WX_QT_FILEDLG_H_
 #define _WX_QT_FILEDLG_H_
 
+// Use MSW implementation to support custom controls in modern-style dialogs.
+#ifdef _WIN32
+#include <wx/msw/filedlg.h>
+#else
+
 class QFileDialog;
 
 class WXDLLIMPEXP_CORE wxFileDialog : public wxFileDialogBase
@@ -54,5 +59,7 @@ private:
 
     wxDECLARE_DYNAMIC_CLASS(wxFileDialog);
 };
+
+#endif // _WIN32
 
 #endif // _WX_QT_FILEDLG_H_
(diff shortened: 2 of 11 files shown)


include/wx/qt/dirdlg.h
```diff
@@ -8,6 +8,11 @@
 #ifndef _WX_QT_DIRDLG_H_
 #define _WX_QT_DIRDLG_H_
 
+// Use MSW implementation to support custom controls in modern-style dialogs.
+#ifdef _WIN32
```
vadz, line 12: We prefer to use our own symbols for tests, i.e.
```suggestion
#ifdef __WINDOWS__
```

include/wx/qt/filedlg.h
```diff
@@ -8,6 +8,11 @@
 #ifndef _WX_QT_FILEDLG_H_
 #define _WX_QT_FILEDLG_H_
 
+// Use MSW implementation to support custom controls in modern-style dialogs.
+#ifdef _WIN32
```
vadz, line 12: Same as above.

src/msw/filedlg.cpp
```diff
@@ -1141,6 +1141,8 @@ void wxFileDialog::MSWOnInitDone(WXHWND hDlg)
         return;
     }
 
+#ifdef __WXMSW__
```
vadz, line 1144:
```suggestion
// Note that this file can also be compiled as part of wxQt and we don't need
// this code then.
#ifdef __WXMSW__
```

src/qt/app.cpp
```diff
@@ -15,6 +15,11 @@
 #include <QtCore/QStringList>
 #include <QtWidgets/QApplication>
 
+#ifdef _WIN32
+    #include "wx/dynlib.h"
+    #include <Shlwapi.h>
```
vadz, line 20: When cross-compiling the case can matter and, besides, we should be consistent with all the other includes of this header:
```suggestion
    #include <shlwapi.h>
```

vadz (commented):
I can't say I like it because it looks like a hack, but it's probably not that bad... Still, having a dummy `HWND` in all wxQt windows just for this looks quite ugly.

Ideal would be to somehow refactor this code to extract it into a common header (e.g. `include/wx/private/windows_filedlg.h`) and allow abstracting the differences between wxMSW with wxQt in some way. But this would clearly require much more work :-(

I also wonder if everybody really wants to have native dialogs in their wxQt applications, i.e. whether we should do it unconditionally?

dsa-t:
> I also wonder if everybody really wants to have native dialogs in their wxQt applications, i.e. whether we should do it unconditionally?

Qt already shows the native dialog on Windows. But it doesn't implement the API to add custom controls to the `IFileDialog`.
Adding custom controls to the Qt's generic dialog is not implemented in wxQt. But the Qt's generic dialog is not great.

vadz:
Do you plan to try refactoring this to reduce the conditional compilation hacks or should I try to do it myself (maybe just to convince myself that it's not going to work...)?

include/wx/qt/dirdlg.h
dsa-t: Seems like `__WINDOWS__` might not be always defined.

image

vadz: It is defined in `wx/defs.h` (well, `wx/platform.h` included from it to be precise), so it should be defined as soon as any wx header is included. So I guess the problem is that this header doesn't include any headers at all — just add

```cpp
#include "wx/defs.h"
```

to fix this.

dsa-t:
> Do you plan to try refactoring this to reduce the conditional compilation hacks or should I try to do it myself (maybe just to convince myself that it's not going to work...)?

Go ahead, as my refactorings won't be as good.

btw, this hack doesn't work properly with the old `SetExtraControlCreator` API.

AliKet:
> > I also wonder if everybody really wants to have native dialogs in their wxQt applications, i.e. whether we should do it unconditionally?
> 
> Qt already shows the native dialog on Windows.

That's true, and that's why the test here: https://github.com/wxWidgets/wxWidgets/blob/70de12cd320e85e9102f42081c4eb4ce2ccc9c23/tests/controls/dialogtest.cpp#L88 fails under Windows, i.e. The `QFileDialog::selectedFiles()` function (which is called by `GetFilename()`) continues to return an empty string until the native dialog is actually displayed. And I simply disabled it there (see commit  [be147ad](https://github.com/wxWidgets/wxWidgets/pull/24837/commits/be147addc937a6e9da6287b2fe9c122cc5680f02))

> But it doesn't implement the API to add custom controls to the `IFileDialog`. Adding custom controls to the Qt's generic dialog is not implemented in wxQt. But the Qt's generic dialog is not great.

The test passes if we add `GetQFileDialog()->setOption(QFileDialog::DontUseNativeDialog);` to `wxFileDialog::Create()`, but then the dialog doesn't look great as you've already noticed.

vadz:
Better late than never: I've finally created #25281 which implements a somewhat cleaner solution (even though it's still a hack, obviously). Please let me know there if you see any problems with doing it like this. TIA!

PR #24592, opened 2024-06-07
Fix printf command issues in PO catalogs

Blake-Madden:


lanurmi:
The underlying issue is that apparently none of the checks run by GitHub check for PO file brokenness. That is, run `msgfmt -c` for each `.po`, and fail the whole check if `msgfmt` fails.

vadz:
Thanks for the fix but I'm very surprised that `msgfmt -c` does _not_ detect all such issues, even though it's supposed to:

```
% msgfmt -v -c -f -o /dev/null locale/ka.po 
1841 translated messages, 8 fuzzy translations, 116 untranslated messages.
```

This is with "msgfmt (GNU gettext-tools) 0.21". Does anybody have any idea what's going on here?

But OTOH with `-f` it does detect many more of them, e.g.:
```
% msgfmt -v -c -f -o /dev/null locale/vi.po
locale/vi.po:1392: format specifications in 'msgid' and 'msgstr' for argument 3 are not the same
locale/vi.po:1397: format specifications in 'msgid' and 'msgstr' for argument 2 are not the same
locale/vi.po:1402: format specifications in 'msgid' and 'msgstr' for argument 2 are not the same
locale/vi.po:1407: format specifications in 'msgid' and 'msgstr' for argument 2 are not the same
locale/vi.po:1412: format specifications in 'msgid' and 'msgstr' for argument 2 are not the same
locale/vi.po:2503: format specifications in 'msgid_plural' and 'msgstr[0]' for argument 1 are not the same
locale/vi.po:2509: format specifications in 'msgid_plural' and 'msgstr[0]' for argument 2 are not the same
locale/vi.po:2546: number of format specifications in 'msgid' and 'msgstr' does not match
locale/vi.po:2571: number of format specifications in 'msgid' and 'msgstr' does not match
locale/vi.po:3959: number of format specifications in 'msgid' and 'msgstr' does not match
locale/vi.po:4352: format specifications in 'msgid' and 'msgstr' for argument 2 are not the same
locale/vi.po:4357: number of format specifications in 'msgid' and 'msgstr' does not match
locale/vi.po:5509: format specifications in 'msgid' and 'msgstr' for argument 2 are not the same
locale/vi.po:5514: format specifications in 'msgid' and 'msgstr' for argument 1 are not the same
locale/vi.po:5878: format specifications in 'msgid' and 'msgstr' for argument 1 are not the same
locale/vi.po:5903: number of format specifications in 'msgid' and 'msgstr' does not match
locale/vi.po:6101: number of format specifications in 'msgid' and 'msgstr' does not match
locale/vi.po:6390: 'msgid' and 'msgstr' entries do not both end with '\n'
locale/vi.po:6521: number of format specifications in 'msgid' and 'msgstr' does not match
locale/vi.po:9040: number of format specifications in 'msgid' and 'msgstr' does not match
msgfmt: found 20 fatal errors
1665 translated messages, 143 fuzzy translations, 157 untranslated messages.
```

So while I agree that we need to run `msgfmt -c` in a check (PRs welcome!), it's probably not even enough, as I always do it manually before updating the translations...

@Blake-Madden Should I apply this or will you update it to fix the errors found by `msgfmt -f -c` too?

Blake-Madden:
I'll do another pass and add some more fixes here.

I'm working on an experimental tool to find these (and other i18n/l10n) issues. I'll tweak it some more and see if I can catch more of these issues that `msgfmt -f -c` is picking up...

Blake-Madden:
Oh, the strings you show here are `#, fuzzy, c-format` instead of `#, c-format`. Maybe that is why `msgfmt -c` isn't enough? I can see that's why my tool was also missing them.

vadz:
> Oh, the strings you show here are `#, fuzzy, c-format` instead of `#, c-format`. Maybe that is why `msgfmt -c` isn't enough? I can see that's why my tool was also missing them.

Just to be clear:

1. Somehow, `msgfmt -c` does _not_ detect the problems in non-fuzzy entries, even though it's clearly supposed to, i.e. this looks like a gettext bug.
2. I didn't know that `msgfmt -c` ignored fuzzy entries by default and that you need to use `-f` to take them into account. It's possible that it still doesn't detect all the mismatches even then due to the bug above, but this is an independent point.

Blake-Madden:
I just fixed a small handful of fuzzy translations (I left them fuzzy); just the ones that were obvious to me. I think this PR is good to go, the production string fixes are the important part.

> 1. Somehow, `msgfmt -c` does _not_ detect the problems in non-fuzzy entries, even though it's clearly supposed to, i.e. this looks like a gettext bug.

For what it's worth, the tool I'm working on (https://github.com/Blake-Madden/i18n-check) detected these. I can run this from time-to-time to keep an eye on things.

vadz:
For the record, I've [reported this to gettext](https://savannah.gnu.org/bugs/index.php?65858).

vadz:
> For what it's worth, the tool I'm working on (https://github.com/Blake-Madden/i18n-check) detected these. I can run this from time-to-time to keep an eye on things.

Thanks, but ideal would be to add a workflow checking this which would run for any changes to `locale/*.po` files. Do you think you could do it? TIA!

Blake-Madden:
> Thanks, but ideal would be to add a workflow checking this which would run for any changes to `locale/*.po` files. Do you think you could do it? TIA!

Sure, should be easier enough to make a github action for this. I'll test it a little bit more the next couple of days to make sure there aren't any false positives and then send in a PR for it...

Blake-Madden:
Regarding spaces in `printf` commands, they are allowed for (presumably) numbers:

https://en.cppreference.com/w/c/io/fprintf

> space: if the result of a signed conversion does not start with a sign character, or is empty, space is prepended to the result. It is ignored if + flag is present.

Reading this, a space would _not_ make sense for a string (i.e., `%s`). Maybe `printf` would silently allow this (`wxString::Format` does), but this just seems wrong to me (IMO).

Perhaps this is why `msgfmt` doesn't catch these?

Blake-Madden:
Something else that `msgfmt -c` overlooks is this:

source: &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`%d: blah`
translation: `%id: blah`

`d` and `i` are interchangeable and `msgfmt` allows that; however, this clearly is a typo given the stray `d` in the translation. My tool is much less forgiving and that's why it caught it while it slipped past `msgfmt -c`.

vadz:
The problem is not just the presence of the space but the fact that the format specifier is different in the original and the translated string — they really should be the same.

PR #24593, opened 2024-06-07
Correct Kopiëren in nl.po

HelioGuilherme66:
As detected in an app done with wxPython, and reported in https://discuss.wxpython.org/t/context-menu-in-stc-not-showing-e-for-dutch-localization/36958/2


vadz:

PR #24597, opened 2024-06-08
document and test behavior of wxDC::DrawRectangle() with transparent pen

wsu-cb:


wsu-cb:
Can anyone explain why the wxGTK 3 gui tests are failing at `bitmap.cpp:185`?  That is supposed to verify that the bitmap BEFORE running the `DrawRectangle(wxSize(1, 1))` is entirely `wxYELLOW`, but that initial condition is not true!

paulcor:
It's because the default pen join is `wxJOIN_BEVEL`. This is going to blend the corners with whatever was there already. Here is one way to fix this.
```diff
diff --git a/tests/graphics/bitmap.cpp b/tests/graphics/bitmap.cpp
index 7f9e901d5d..55dc057f28 100644
--- a/tests/graphics/bitmap.cpp
+++ b/tests/graphics/bitmap.cpp
@@ -140,7 +140,7 @@ TEST_CASE("BitmapTestCase::ToImage", "[bitmap][image][convertto]")
         wxBitmap bmp(16, 16, 24);
         {
             wxMemoryDC dc(bmp);
-            dc.SetPen(*wxYELLOW_PEN);
+            dc.SetPen(wxPenInfo(*wxYELLOW).Join(wxJOIN_MITER));
             dc.SetBrush(*wxYELLOW_BRUSH);
             dc.DrawRectangle(0, 0, bmp.GetWidth(), bmp.GetHeight());
         }
```

PR #24599, opened 2024-06-09
Add Visual Studio default output folders and CMakeSettings.json to .gitignore

dsa-t:


.gitignore
```diff
@@ -10,6 +10,9 @@
 /setup.h
 /*tags
 /autom4te.cache
+/.vs
+/out
```
vadz, line 14: This isn't MSVS default output folder, so I don't know why should we ignore this one.

.gitignore
```diff
@@ -10,6 +10,9 @@
 /setup.h
 /*tags
 /autom4te.cache
+/.vs
```
vadz, line 13: I'm fighting a losing battle, but I really think that all such standard directories should be ignored in the global `.gitignore` instead of doing it in each and every project-specific file.

vadz (commented):
Personally I wouldn't apply this because I think this belongs to the global `.gitignore` but I guess there is no real harm in applying it neither (except for `out`, this one IMO just shouldn't be ignored at all).

.gitignore
dsa-t: Here's the content of the default CMakeSettings.json that is created when clicking Project -> CMake Settings for wxWidgets:

```json
{
  "configurations": [
    {
      "name": "x64-Debug",
      "generator": "Ninja",
      "configurationType": "Debug",
      "inheritEnvironments": [ "msvc_x64_x64" ],
      "buildRoot": "${projectDir}\\out\\build\\${name}",
      "installRoot": "${projectDir}\\out\\install\\${name}",
      "cmakeCommandArgs": "",
      "buildCommandArgs": "",
      "ctestCommandArgs": ""
    }
  ]
}
```

So `out` is indeed the default build+install dir in this case.

.gitignore
dsa-t: Not in every project the `.vs` dir should be ignored, so a global config can't be applied I think.

VS creates lots of files there, and until I ignore them, VS usually fails to perform Git operations like switching to a branch or a commit:

image 

image

vadz: > Not in every project the .vs dir should be ignored, so a global config can't be applied I think.

Why not? I have it in mine since a very long time and didn't notice any problems due to this.

.gitignore
vadz: Huh, sorry, I didn't know this. Where does this come from? I can't find `out` anywhere under `build/cmake`.

dsa-t: It's predefined in VS: https://learn.microsoft.com/en-us/cpp/build/cmake-predefined-configuration-reference?view=msvc-170#x86-64-predefined-build-configurations

.gitignore
dsa-t: Some projects require commiting specific files inside these dot directories.
And it would be more manual work for a new developer.

vadz:
Any other opinions about ignoring these files?

To summarize, personally I think that:

- `.vs` should be globally ignored (I've never encountered a project requiring committing something there and IMNSHO this is just plain wrong because this is a tool directory that should never be used for the project-specific files).
- `out` should not be ignored at all because it's a common name and I often create temporary files with this name and I do _not_ want to be ignored if I forget to delete them.
- `CMakeSettings.json` seems to be deprecated, according to a quick web search, so I wouldn't bother with it neither.

IOW I still would be inclined to close this PR, but if enough other people think it's worth adding it, I won't argue.

dsa-t:
> * `CMakeSettings.json` seems to be deprecated, according to a quick web search, so I wouldn't bother with it neither.

An alternative `CMakePresets.json` is very new, and it is disabled by default in VS.

vadz:
Sorry, it doesn't seem like anybody else is interested in this, so I'd rather close it.

PR #24600, opened 2024-06-09
[wxQt] Fix  wxEVT_TREE_SEL_{CHANGING,CHANGED} generation/handling

AliKet:


src/qt/treectrl.cpp
```diff
@@ -125,6 +125,61 @@ class ImageState
...
+};
+
+/*static*/
+bool wxQItemSelectionModel::ms_isSelectionChangeAllowed = true;
+
+namespace
+{
+// Used when generating wxEVT_TREE_SEL_{CHANGING,CHANGED} events only
+static wxTreeItemId gs_currentItem;
+static wxTreeItemId gs_previousItem;
```
vadz, line 180: Do these really need to be global? I am not 100% sure it's not going to be a problem, e.g. if a handler of some event changes selection in some other tree control.

src/qt/treectrl.cpp
```diff
@@ -125,6 +125,61 @@ class ImageState
 
 Q_DECLARE_METATYPE(TreeItemDataQt)
 
+// We use a custom selection model because we need to generate the
+// wxEVT_TREE_SEL_CHANGED event only if it is allowed. that is, only
```
vadz, line 129: Nitpick of the day:
```suggestion
// wxEVT_TREE_SEL_CHANGED event only if it is allowed, that is, only
```

(alternatively, keep the period, but capitalize "That")

vadz (commented):
Thanks for the fixes, I think the logic is correct here, but if we could store the items in, say, `wxQTreeWidget`, as member variables, it would be better IMHO.

src/qt/treectrl.cpp
AliKet: forgot to update this

AliKet:
@dsa-t It would be really apreciated if you could test this PR (or this [patch](https://patch-diff.githubusercontent.com/raw/wxWidgets/wxWidgets/pull/24600.diff)) before it get merged because I had to change the implementation a little due to problems encountered with `wxTreeBook` control with the previous version.

TIA.

AliKet:
Oops! this version reintroduced the infinite loop reported by @dsa-t which can be seen with the notebook sample (File -> Type -> Treebook)

AliKet:
This is really strange because the treectrl sample works correctly but the notebook (treebook page) does not! maybe the problem is in the `wxTreeBook` control !?

Needs more debugging...

AliKet:
> This is really strange because the treectrl sample works correctly but the notebook (treebook page) does not! maybe the problem is in the `wxTreeBook` control !?
> 
> Needs more debugging...

Everything seems to work correctly now: the treectrl and notebook (wxTreebook) samples.

dsa-t:
Works good generally, but I also see a small issue with multiple selections:

1. In `wxTR_MULTIPLE` mode, select `File child 1.1` to `File child 1.5`
2. Then, select, for example, `File child 1.2`
3. Inside `OnSelChanged`, `GetSelections` will say "File child 1.1""File child 1.2""File child 1.3""File child 1.4""File child 1.5"
Also the name editor gets triggered (I don't think it should).

AliKet:
> Works good generally, but I also see a small issue with multiple selections in `treetest`:
> 
> 1. In `wxTR_MULTIPLE` mode, select `File child 1.1` to `File child 1.5`
> 2. Then, select, for example, `File child 1.2`
> 3. Inside `OnSelChanged`, `GetSelections` will say "File child 1.1""File child 1.2""File child 1.3""File child 1.4""File child 1.5"
>    Also the name editor gets triggered (I don't think it should).

Thanks for testing, but it works correctly for me!

Screenshot

dsa-t:
> Thanks for testing, but it works correctly for me!

In the 2nd step, I mean select just the 1.2 item, not deselect it:

image

AliKet:
> > Thanks for testing, but it works correctly for me!
> 
> In the 2nd step, I mean select just the 1.2 item, not deselect it:

Indeed, I can reproduce the problem.  And the fix is simple which I'll push soon.
but preventing the editor from being triggered is trickier because this is how Qt interpret `SelectedClicked` flag passed here https://github.com/wxWidgets/wxWidgets/blob/f8043a62f3319719cbd3a400ea454e4dbb3d23ff/src/qt/treectrl.cpp#L1311-L1315 

i.e. It starts editing when clicking on an already selected item

See Qt docs https://doc.qt.io/qt-6/qabstractitemview.html#EditTrigger-enum:~:text=QAbstractItemView%3A%3ASelectedClicked

AliKet:
@dsa-t  this patch:
```C++
diff --git a/src/qt/treectrl.cpp b/src/qt/treectrl.cpp
index ea68681ae0..45be7b0c48 100644
--- a/src/qt/treectrl.cpp
+++ b/src/qt/treectrl.cpp
@@ -330,11 +330,31 @@ protected:
         }
     }
 
+    // For compatibility with MSW, selecting an item that is part of a multiple selection
+    // simply deselects the other items without triggering the edit mode on the item. This
+    // is not the case with Qt because we already tell it to start editing when clicking
+    // on an already selected item with setEditTriggers(QTreeWidget::SelectedClicked) call.
+    bool m_blockEditing = false;
+
+    void selectionChanged(const QItemSelection& selected,
+                          const QItemSelection& deselected) override
+    {
+        m_blockEditing = selected.size() == 0;
+
+        QTreeView::selectionChanged(selected, deselected);
+    }
+
     bool edit(const QModelIndex &index, EditTrigger trigger, QEvent *event) override
     {
         // AllEditTriggers means that editor is about to open, not waiting for double click
         if (trigger == AllEditTriggers)
         {
+            if ( m_blockEditing )
+            {
+                m_blockEditing = false;
+                return false;
+            }
+
             // Allow event handlers to veto opening the editor
             wxTreeEvent wx_event(
                 wxEVT_TREE_BEGIN_LABEL_EDIT,

```
solves the problem with starting edit you mention, but I'm still not sure if we should apply it or not because wxGTK3 also start editing if you do the following:
- Select `File child 1.1`
- Hold `Ctrl` key down and select `File child 1.2`
- Now you have 2 item selected (`File child 1.1` and `File child 1.2`)
- Select `File child 1.2` with or without holding the `Ctrl` key down.
- The control starts editing the item.

Notice that this happens only if you select the focused item. `wxMSW` doesn't behave like this. i.e. it doesn't start editing at all.

@vadz any thoughts about this ?

vadz:
I think wxGTK (i.e. generic implementation) is wrong here, this doesn't seem like a useful behaviour to me.

AliKet:
@vadz  I have no idea what's going on here, or what to do with these failures?

vadz:
I'm afraid that at least the Mac failures are real ones, i.e. actually due to the changes here as they seem to be occurring in the affected code. For the MSW ones I'm less sure and I don't have time to debug this right now, sorry...

AliKet:
@vadz This is ready now. just needs **AppVeyor** restart, TIA!

tests/controls/treectrltest.cpp
```diff
@@ -25,82 +25,55 @@
...
+
+    ~TreeCtrlTestCase()
+    {
+        delete m_tree;
+        m_tree = nullptr;
+
+        m_root =
+        m_child1 =
+        m_child2 =
+        m_grandchild = wxTreeItemId();
```
vadz, line 50: Minor, but this is not really needed in a dtor.

tests/controls/treectrltest.cpp
```diff
@@ -111,134 +84,103 @@ class TreeCtrlTestCase : public CppUnit::TestCase
...
-    m_tree->Refresh();
-    m_tree->Update();
-}
-
-void TreeCtrlTestCase::tearDown()
-{
-    delete m_tree;
-    m_tree = nullptr;
+#ifdef __WXMSW__
+    // Toggling the wxTR_HIDE_ROOT window style has no effect under wxMSW if
```
vadz, line 94: It would be nice to mention this in `wxTR_HIDE_ROOT` docs.

tests/controls/treectrltest.cpp
```diff
@@ -111,134 +84,103 @@ class TreeCtrlTestCase : public CppUnit::TestCase
...
-    m_tree->ExpandAll();
-    m_tree->Refresh();
-    m_tree->Update();
-}
-
-void TreeCtrlTestCase::tearDown()
-{
-    delete m_tree;
-    m_tree = nullptr;
+#ifdef __WXMSW__
```
vadz, line 93: I'm not sure if it's worth doing, but we could get rid of this `#ifdef` by defining `TreeCtrlHideRootTestCase` deriving from `TreeCtrlTestCase` and passing `wxTR_HIDE_ROOT` as (optional) style argument to the base class ctor.

.github/workflows/ci_mac.yml
```diff
@@ -232,7 +232,7 @@ jobs:
...
+        wx_tests_selection='~[.] ~[valnum] ~wxPersistTLW ~wxPersistDVC ~wxTopLevel::Show ~wxFont::Weight ~wxFont::NativeFontInfo ~WebView ~RadioButton::Single ~RadioButton::Focus ~wxHtmlPrintout::Pagination ~wxTextCtrl::GetBestSize ~TextCtrlTestCase ~wxExecute::RedirectUTF8 ~Ellipsization::NormalCase ~SpinCtrl::* ~SpinCtrlDouble::* ~NotebookTestCase ~SettingsTestCase ~Window::Show ~ModalDialogsTestCase ~Button::Click ~Button::Disabled ~wxDVC::GetItemRect ~wxDVC::AppendTextColumn ~Grid::KeyboardSelection ~Grid::CellClick ~Grid::ReorderedColumnsCellClick ~Grid::CellSelect ~wxStyledTextCtrl::AutoComp ~EvtLoopTestCase ~EventPropagationTestCase ~wxTreeCtrl::ItemClick ~wxTreeCtrl::LabelEdit ~wxTreeCtrl::KeyDown ~wxTreeCtrl::CollapseExpandEvents ~wxTreeCtrl::SelectionChange ~wxTreeCtrl::SelectItemMultiInteractive ~wxTreeCtrl::Menu ~wxTreeCtrl::KeyNavigation'
```
vadz, line 235: Are there any actually tests passing under Mac? It's not obvious why we don't exclude `TreeCtrlTestCase` entirely...

vadz (commented):
Thanks, I don't know what's going on with AppVeyor and why it considers this PR unmergeable. Do you plan to rebase it again? If so, it might be enough to force it to run the builds again... If not, let me know and I'll just merge it.

tests/controls/treectrltest.cpp
AliKet: Thanks for the suggestion.

.github/workflows/ci_mac.yml
AliKet: Excluded tests (`wxTreeCtrl::*`) are only those that require `wxUIActionSimulator` to run. Excluding `TreeCtrlTestCase` entirely would exclude the others: `wxTreeCtrl::{HasChildren, GetCount, SelectItemSingle, SelectItemMulti, ...}` to name a few that run successfully.

tests/controls/treectrltest.cpp
AliKet: Oops! I forgot to update the documentation. Please could you do this when merging?

TIA!

Issue #24516, opened 2024-04-28
Scrolling with trackpad in two directions at once doesn't seem to work properly


craftyjon:
### Description

When some widget can be scrolled both horizontally and vertically, wxWidgets seems to get "locked in" to accepting only one scroll direction.  On macOS hardware with a trackpad input, it is easy to input "diagonal" scrolling, but this doesn't really work in wxWidgets, because one or the other direction (horizontal or vertical) will "win" and so diagonal scrolling of the widget can only be accomplished "Manhattan style".

This doesn't seem to be a limitation of the hardware, as other applications (e.g. HTML Canvas in Firefox/Chrome) can accept simultaneous scroll inputs on two axes.

I am not sure if this is a wxMac-specific problem or not, as the only hardware I have that can give this kind of input is a Macbook.

#### To Reproduce:

Run the wxGrid sample and try scrolling around with a trackpad that allows scrolling diagonally (i.e. two mousewheel inputs at once)

Compare behavior to, for example, some of the demos at https://konvajs.org/docs/sandbox/Canvas_Scrolling.html

### Platform and version information

- wxWidgets version  you use: 3.2.4
- wxWidgets port  you use: wxMac
- OS  and its version: macOS 14.4

showier-drastic:
Can confirm this behaviour on wx. And to be specific, demo 3 in your example works diagonally.

vadz:
We don't have a way of representing diagonal scrolling at `wxMouseEvent` level currently as it has `wxMouseWheelAxis` field which can take only one of 2 values: either `wxMOUSE_WHEEL_VERTICAL` or `wxMOUSE_WHEEL_HORIZONTAL`.

Ideal would probably be to add `wxMOUSE_WHEEL_MIXED` or something like this and also add the angle/direction or maybe just add `GetWheelRotationX()` and `GetWheelRotationY()` in addition to the existing `GetWheelRotation()`.

But, alternatively, and simpler, we could send 2 events for the horizontal and vertical components respectively, because I think any program would still handle diagonal scrolling as a sum of H/V scrolling, but I'm not totally sure about this.

AFAICS only Cocoa sends a single event with deltas in both directions, both MSW and GTK seem to only send H/V events, although maybe there is some other way to get trackpad scrolling input that would allow receiving both?

[...]

PR #24601, opened 2024-06-09
Generate both vertical and horizontal mouse wheel events for diagonal scrolling

vadz:
See #24516 for motivation.

@craftyjon Please let me know if this works for you.

@csomor Please let me know if you see a better way to translate a single `NSEvent` to multiple `wxMouseEvent`s than what I did here.

### DIFF
--- a/src/osx/cocoa/window.mm
+++ b/src/osx/cocoa/window.mm
@@ -791,17 +791,28 @@ wxWidgetCocoaImpl::TranslateMouseEvent( NSEvent * nsEvent )
             wxevent.m_linesPerAction = 1;
             wxevent.m_columnsPerAction = 1;
 
-            if ( fabs(deltaX) > fabs(deltaY) )
+            // Should we impose some minimum threshhold here?
+            if ( deltaY != 0 )
             {
+                wxevent.m_wheelRotation = (int)deltaY;
+            }
+
+            if ( deltaX != 0 )
+            {
+                // We can send one or two events depending on whether we have
+                // delta in just one or both directions.
+                wxMouseEvent* wxeventPtr = &wxevent;
+                if ( deltaY != 0 )
+                {
+                    wxevents.push_back(wxevent);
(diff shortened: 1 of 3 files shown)


csomor:
@vadz thanks, that's something that would probably work without changes needed in the client apps, I was thinking whether we'd need some (new?) 'panning' event having two dimensions to avoid Manhattan geometry movements ...

src/osx/cocoa/window.mm
```diff
@@ -3933,9 +3955,18 @@ new wxCocoaGesturesImpl(this, m_osxView, eventsMask)
...
     (void)SetupCursor(event);
 
-    wxMouseEvent wxevent(wxEVT_LEFT_DOWN);
-    SetupMouseEvent(wxevent , event) ;
-    return GetWXPeer()->HandleWindowEvent(wxevent);
+    bool processed = false;
+    for ( auto& wxevent : TranslateMouseEvent(event) )
+    {
+        // Even if this event was processed, still continue with the other
+        // eveents, if any.
```
dsa-t, line 3962: Typo.
```suggestion
        // events, if any.
```

src/osx/cocoa/window.mm
```diff
@@ -787,17 +791,28 @@ - (CGFloat)scrollingDeltaY;
             wxevent.m_linesPerAction = 1;
             wxevent.m_columnsPerAction = 1;
 
-            if ( fabs(deltaX) > fabs(deltaY) )
+            // Should we impose some minimum threshhold here?
```
dsa-t, line 794:
```suggestion
            // Should we impose some minimum threshold here?
```

src/osx/cocoa/window.mm
vadz: Thanks, I'll fix this (and the other) typo(s) when I merge it, if it works.

craftyjon:
@vadz This works well out of the box!  No changes needed on our side.  Thank you :)

PR #24598, opened 2024-06-09
Recognize Windows Server 2019 in wxGetOsDescription()

PBfordev:
For some reason, testing for Windows Server 2019 was missing there.

See also #24594.

PR #24603, opened 2024-06-10
Fix detecting Windows Server 2022 in wxGetOsDescription()

PBfordev:
Windows Server 2022 has build number lower than first Windows 11 build number (20348 vs 22000), so we cannot use Windows 11 build number to test for it here.

Sorry for not noticing this pre-existing issue when working on #24598.

### DIFF
--- a/src/msw/utils.cpp
+++ b/src/msw/utils.cpp
@@ -1205,7 +1205,8 @@ wxString wxGetOsDescription()
 
             if ( str.empty() )
             {
-                str.Printf("Windows %lu.%lu",
+                str.Printf("Windows %s%lu.%lu",
+                           wxIsWindowsServer() == 1 ? "Server " : "",
                            info.dwMajorVersion,
                            info.dwMinorVersion);
             }
--- a/src/msw/utils.cpp
+++ b/src/msw/utils.cpp
@@ -1108,7 +1108,9 @@ int wxIsWindowsServer()
     return -1;
 }
 
+static const int WINDOWS_SERVER2016_BUILD = 14393;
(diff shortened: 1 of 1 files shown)


src/msw/utils.cpp
```diff
@@ -1177,21 +1179,27 @@ wxString wxGetOsDescription()
...
+                            case WINDOWS_SERVER2016_BUILD:
+                                str = "Windows Server 2016";
+                                break;
+                            case WINDOWS_SERVER2019_BUILD:
+                                str = "Windows Server 2019";
+                                break;
+                            case WINDOWS_SERVER2022_BUILD:
+                                str = "Windows Server 2022";
+                                break;
+                        }
```
vadz, line 1195: Should we do something like "Windows Server build %d" for the other, unknown, builds? I understand that this is not supposed to happen, but what if they release such build in the future?

vadz (commented):
Thanks for fixing this, but I think we should use "Windows Server" for the unknown server versions too, either by adding `default:` label as mentioned in the comment or maybe by testing for `wxIsWindowsServer()` in the fallback code below.

PBfordev:
I opted for adding the information for unknown server editions to the generic fallback code, instead of possibly reporting it as the last known v10.0 server edition.

Unlike workstation editions, server editions seem to have the build number locked, so they are easy to identify. Windows Server 2025 should be released this year and I am guessing that its version will be 10.0 as well, but its build version is not locked yet.

I also considered removing code for dealing with pre-Win7 versions (< 6.1) but I decided to let it stay: Not only It would make backporting harder but I am not sure if wxWidgets 3.3 really cannot run on those systems.

Issue #24604, opened 2024-06-11
For wxPoint2DDouble, wxPoint2DInt: operator/= and operator*= for wxDouble, wxInt32 args are undefined


dsa-t:
### Description
When trying to run this code:

```cpp
wxPoint2DDouble() /= 5.0;
```

I get this error:
```
symbol lookup error: libwx_gtk3u_core-3.3.so.0: undefined symbol: _ZN15wxPoint2DDoubledVEd
```
---
VScode shows these as unimplemented:

wxPoint2DInt:
```cpp
    inline wxPoint2DInt& operator*=(wxDouble n);
    inline wxPoint2DInt& operator*=(wxInt32 n);

    inline wxPoint2DInt& operator/=(wxDouble n);
    inline wxPoint2DInt& operator/=(wxInt32 n);
```

wxPoint2DDouble:
```cpp
    inline wxPoint2DDouble& operator*=(wxDouble n);
    inline wxPoint2DDouble& operator*=(wxInt32 n);

    inline wxPoint2DDouble& operator/=(wxDouble n);
    inline wxPoint2DDouble& operator/=(wxInt32 n);
```

### Platform and version information

- wxWidgets version  you use: master
- wxWidgets port  you use: wxGTK
- OS  and its version: Linux Mint

vadz:
I'll add the implementation of the missing operators, thanks, but I am very puzzled by the existence of operator overloads taking 2 points as IMO they makes no sense at all, so I'd prefer to completely delete (well, deprecate for now) those.

@csomor Do you remember why were these added, by chance?

csomor:
> I'll add the implementation of the missing operators, thanks, but I am very puzzled by the existence of operator overloads taking 2 points as IMO they makes no sense at all, so I'd prefer to completely delete (well, deprecate for now) those.
> 
> @csomor Do you remember why were these added, by chance?

I'm sorry, I don't remember. I could have been looking at a 2D geometry library at that point, to find out what they included. The geometrical interpretation could be a scaling on x and y with different factors. But now these would be done using matrices anyway. So yes, go ahead, deprecate and mark for deletion :-)

PR #24608, opened 2024-06-13
Minor geometry classes cleanup and fixes

vadz:
This started as a fix for #24604 but I also couldn't help changing a few other things.

### DIFF
--- a/include/wx/geometry.h
+++ b/include/wx/geometry.h
@@ -513,8 +513,7 @@ inline bool wxPoint2DDouble::operator!=(const wxPoint2DDouble& pt) const
 class WXDLLIMPEXP_CORE wxRect2DDouble
 {
 public:
-    wxRect2DDouble()
-        { m_x = m_y = m_width = m_height = 0; }
+    wxRect2DDouble() = default;
     wxRect2DDouble(wxDouble x, wxDouble y, wxDouble w, wxDouble h)
         { m_x = x; m_y = y; m_width = w;  m_height = h; }
 /*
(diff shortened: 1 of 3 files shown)


vadz:
Oops, sorry for including an unrelated first commit.

vadz:
I've updated the PR to use `wxRound()` when multiplying/dividing ints by doubles, this seems to make more sense than truncating.

include/wx/geometry.h
```diff
@@ -347,15 +360,39 @@ public :
...
+        m_y *= pt.m_y;
+        return *this;
+    }
 
-inline wxPoint2DDouble::wxPoint2DDouble()
-{
-    m_x = 0.0;
-    m_y = 0.0;
-}
+    wxDEPRECATED_MSG("Multiplying points doesn't make sense")
```
dsa-t, line 372: Why doesn't it make sense?
It makes sense when working with non-uniform scale factors, for example.

dsa-t: Because methods like `GetVectorLength` `SetVectorLength`, `Normalize`, `GetDotProduct`, `GetCrossProduct` exist, looks like it was designed to store vector data as well.

vadz: I mean, it _is_ a vector. But you don't multiply vectors like this.

csomor: > Non uniform scale factor is not a point. Mathematically, points are elements of a vector space and there is no multiplication operation in it.

Absolutely agreed, you asked where this might have come from ;-) could be it was something from Taligent, I agree it is a misuse of a point struct to be a two-dimensional scalar factor

dsa-t: `wxPoint2DDouble::operator*=(wxDouble)` is good enough for my purposes, but multiplying vectors like this look totally fine to me.

vadz: Sorry, but this isn't a matter of opinion, I'm speaking in mathematical sense and you just can't define multiplication like this in a vector space... To convince you of this, consider that `(1,0)*(0,1) == (0,0)` with this "multiplication" definition, so you can have `x*y = 0` with `x ≠ 0` and `y ≠ 0` (and, relatedly, neither of these elements has an inverse).

csomor (approved):
Looks fine to me, I know that QuickDraw was always flooring and not rounding, but I don't think this leads to a problem. Thanks a lot.

PR #649, opened 2017-12-24
Add API for raw touch events (see #17077)

e9925248:
First draft. OS X not even build tested

PR #2539, opened 2021-09-27
Fixed missing mouse events when using wxGesture

tbultel:
This fixes https://trac.wxwidgets.org/ticket/19265
The fix consists in two things:

First, it emulates the LeftDown, LeftUp, and Motion
events, that are no longer triggered when using the
"touch-event"

In addition, it removes GDK_BUTTON1_MASK in the press callback,
that is never set when using a mouse, but that is set
for some reason by gtk when using a touchscreen.

Signed-off-by: Thierry Bultel <tbultel@free.fr>

PR #24610, opened 2024-06-13
Add API for raw touch events (updated)

dsa-t:
Rebased version of https://github.com/wxWidgets/wxWidgets/pull/649

Tested with wxGTK on Linux Mint/X11 and with wxMSW on Windows 11 with a touch panel.

Includes wxMSW patch from https://groups.google.com/g/wx-dev/c/JdeuH95_bGQ/m/tcJCXMZUDgAJ


interface/wx/window.h
```diff
@@ -111,7 +111,12 @@ enum
     /**
         Enable all supported gesture events.
      */
-    wxTOUCH_ALL_GESTURES
+    wxTOUCH_ALL_GESTURES,
+
+    /**
+        Enable raw multitouch events
```
vadz, line 117:
```suggestion
        Enable raw multitouch events.
        
        @since 3.3.0
```

samples/touchtest/touchtest.cpp
```diff
@@ -0,0 +1,246 @@
...
+class MyFrame: public wxFrame
+{
+private:
+    struct TouchState
+    {
+        wxTouchSequenceId id;
+        wxPoint last;
+        wxPen pen;
+    };
+    struct TouchState m_TouchPoints[TOUCH_POINTS];
```
vadz, line 33: Very minor, but
```suggestion
    TouchState m_TouchPoints[TOUCH_POINTS];
```

samples/touchtest/touchtest.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+    void OnMouseDown(wxMouseEvent& event);
+    void OnMouseMove(wxMouseEvent& event);
+    void OnMouseUp(wxMouseEvent& event);
+
+    void OnPaint(wxPaintEvent& event);
+    void OnSize(wxSizeEvent& event);
+    void OnQuit(wxCommandEvent& event);
+
+DECLARE_EVENT_TABLE()
```
vadz, line 59:
```suggestion
    DECLARE_EVENT_TABLE()
```

samples/touchtest/touchtest.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+IMPLEMENT_APP(wxMyApp)
+
+bool wxMyApp::OnInit()
+{
+    if ( !wxApp::OnInit() )
+        return false;
+
+    wxFrame* frame = new MyFrame(nullptr, wxT("Multi-touch Test"), wxDefaultPosition,
+        wxSize(500, 400), wxDEFAULT_FRAME_STYLE | wxHSCROLL | wxVSCROLL);
```
vadz, line 70: Hardcoding size in pixels doesn't work nice in high DPI, the simplest thing to do is
```suggestion
        wxDefaultSize, wxDEFAULT_FRAME_STYLE | wxHSCROLL | wxVSCROLL);
```

samples/touchtest/touchtest.cpp
```diff
@@ -0,0 +1,246 @@
...
+    m_TouchPoints[15].pen = wxPen(*wxRED, 2, wxPENSTYLE_DOT);
+
+    wxMenu *file_menu = new wxMenu;
+
+    file_menu->Append(wxID_EXIT);
+
+    wxMenuBar *menu_bar = new wxMenuBar;
+
+    menu_bar->Append(file_menu, wxT("&File"));
+    SetMenuBar(menu_bar);
```
vadz, line 121: This looks rather useless and could be just removed IMO.

samples/touchtest/touchtest.cpp
```diff
@@ -0,0 +1,246 @@
...
+};
+
+IMPLEMENT_APP(wxMyApp)
+
+bool wxMyApp::OnInit()
+{
+    if ( !wxApp::OnInit() )
+        return false;
+
+    wxFrame* frame = new MyFrame(nullptr, wxT("Multi-touch Test"), wxDefaultPosition,
```
vadz, line 69: Minor, but we could do without `wxT()` here and elsewhere.

include/wx/event.h
```diff
@@ -1946,31 +1955,92 @@ class WXDLLIMPEXP_CORE wxSetCursorEvent : public wxEvent
...
+    explicit wxTouchSequenceId(void* pItem) : wxItemId<void*>(pItem) { }
+};
+
+class WXDLLIMPEXP_CORE wxMultiTouchEvent : public wxTouchEventBase
+{
+public:
+    wxMultiTouchEvent(wxWindowID winid = 0, wxEventType type = wxEVT_NULL)
+        : wxTouchEventBase(winid, type)
+    {
+        m_isPrimary = false;
```
vadz, line 2000: Minor, but it would be better to initialize it in the declaration now.

include/wx/event.h
```diff
@@ -1946,31 +1955,92 @@ class WXDLLIMPEXP_CORE wxSetCursorEvent : public wxEvent
...
+    }
+
+    wxMultiTouchEvent(const wxMultiTouchEvent& event) : wxTouchEventBase(event)
+    {
+        m_isPrimary = event.m_isPrimary;
+        m_sequence = event.m_sequence;
+    }
```
vadz, line 2007: AFAICS this should be just
```suggestion
    wxMultiTouchEvent(const wxMultiTouchEvent& event) = default;
```

include/wx/event.h
```diff
@@ -1946,31 +1955,92 @@ class WXDLLIMPEXP_CORE wxSetCursorEvent : public wxEvent
...
+public:
+    wxTouchEventBase(wxWindowID winid = 0, wxEventType type = wxEVT_NULL)
+        : wxEvent(winid, type)
+    {
+    }
+
+    wxTouchEventBase(const wxTouchEventBase& event) : wxEvent(event)
+    {
+        m_pos = event.m_pos;
+    }
```
vadz, line 1973: Could/should be just
```suggestion
    wxTouchEventBase(const wxTouchEventBase& event) = default;
```

include/wx/event.h
```diff
@@ -1946,31 +1955,92 @@ class WXDLLIMPEXP_CORE wxSetCursorEvent : public wxEvent
 
 private:
     wxDECLARE_DYNAMIC_CLASS_NO_ASSIGN(wxSetCursorEvent);
+};
+
+ // Touch Base Event
```
vadz, line 1960:
```suggestion
 // Common base class for wxMultiTouchEvent and wxGestureEvent
```

interface/wx/event.h
```diff
@@ -3892,7 +3892,121 @@ class wxScrollEvent : public wxCommandEvent
...
+*/
+class wxTouchEventBase : public wxEvent
+{
+    /**
+        Constructor.
+    */
+    wxTouchEventBase(wxWindowID winid = 0, wxEventType type = wxEVT_NULL);
+
+    /**
+        Returns the position where the event took effect, in client coordinates: position of Pan event,
```
vadz, line 3917: We should have "brief" description, appearing in the summary in the docs, and for this we need a fullstop here, e.g.
```suggestion
        Returns the position where the event took effect, in client coordinates.
        
         This is the position of Pan event,
```

interface/wx/event.h
```diff
@@ -3892,7 +3892,121 @@ class wxScrollEvent : public wxCommandEvent
...
+        Process all touch events.
+    @endEventTable
+
+    @note Touch events are not generated by default, you must call
+          wxWindow::EnableTouchEvents() with the wxTOUCH_RAW_EVENTS parameter.
+
+    @library{wxcore}
+    @category{events}
+
+    @since 3.1.1
```
vadz, line 3983:
```suggestion
    @since 3.3.0
```

src/gtk/window.cpp
```diff
@@ -3709,6 +3710,44 @@ touch_callback(GtkWidget* widget, GdkEventTouch* gdk_event, wxWindow* win)
...
+        if (type != wxEVT_NULL)
+        {
+            wxMultiTouchEvent event(win->GetId(), type);
+
+            event.SetEventObject(win);
+            event.SetPosition(wxPoint(gdk_event->x, gdk_event->y));
+            event.SetSequenceId(wxTouchSequenceId(gdk_event->sequence));
+            event.SetPrimary(gdk_event->emulating_pointer);
+
+            win->GTKProcessEvent(event);
```
vadz, line 3747: Is it correct to ignore its return value here?

src/msw/window.cpp
```diff
@@ -3557,6 +3601,17 @@ wxWindowMSW::MSWHandleMessage(WXLRESULT *result,
...
         }
         break;
+
+        case WM_TOUCH:
+        {
+            if ( !GestureFuncs::IsOk() )
+                break;
+
+            HandleTouch(wParam, lParam);
+            // Let DefWindowProc handle the touch events too
```
vadz, line 3611: It would be nice to explain why do we need it to always have them?

src/msw/window.cpp
```diff
@@ -6276,6 +6331,40 @@ bool wxWindowMSW::HandlePressAndTap(const wxPoint& pt, WXDWORD flags)
...
     return HandleWindowEvent(event);
 }
+
+bool wxWindowMSW::HandleTouch(WXWPARAM wParam, WXLPARAM lParam)
+{
+    const unsigned count = LOWORD(wParam);
+    wxVector<TOUCHINPUT> info(count);
+    if (GestureFuncs::GetTouchInputInfo() ((HTOUCHINPUT)lParam, count, &info[0], sizeof(TOUCHINPUT)))
+    {
+        for(unsigned i = 0; i < count; i++)
```
vadz, line 6341:
```suggestion
        for ( const auto& input : info )
```
(and then use `input` instead of `info[i]` in the loop body, of course)

src/msw/window.cpp
```diff
@@ -6276,6 +6331,40 @@ bool wxWindowMSW::HandlePressAndTap(const wxPoint& pt, WXDWORD flags)
...
+            else
+                continue;
+
+            wxMultiTouchEvent event( GetId(), type);
+
+            event.SetEventObject( this );
+            event.SetPosition(pos);
+            event.SetSequenceId(wxTouchSequenceId(wxUIntToPtr(info[i].dwID + 1)));
+            event.SetPrimary( (info[i].dwFlags & TOUCHEVENTF_PRIMARY) != 0 );
+            HandleWindowEvent(event);
```
vadz, line 6361: Shouldn't we remember if any of the events were processed and only return true in this case?

src/osx/cocoa/window.mm
```diff
@@ -2153,6 +2185,83 @@ When dragging the bottom part of the DND sample ("Drag text from here!")
...
+}
+
+void wxCocoaGesturesImpl::TouchesCancel(NSEvent* event)
+{
+    RawTouchEvent(event, wxEVT_TOUCH_CANCEL, NSTouchPhaseCancelled);
+}
+
+void* wxCocoaGesturesImpl::GetTouchID(NSObject* id, wxEventType type)
+{
+    for(size_t i = 0; i < m_touchIDs.size(); i++)
```
vadz, line 2201: Here too I'd use range for:
```suggestion
    for ( auto other : m_touchIDs )
```

(the name `other` is confusing to me, so perhaps it would be better to rename it too)

vadz (commented):
Thanks for bringing this up to date!

Globally this looks good to me and I'd definitely like to merge it for 3.3.0. I can address the minor comments I left myself, but if you could please do it, it would be even better, of course. I'd just like to ask you to use "fixup" commits (done by `git commit --fixup`) instead of force-pushing your changes, as this makes seeing what exactly has been changed more difficult.

The only relatively important question I see here concerns `wxTouchEventBase`: I'm just not sure if we really need  it at all. It does save a tiny amount of duplication, but I don't think anybody is actually going to use it, i.e. handle gesture and raw touch events as objects of this class. Maybe we should remove it and just add position to `wxMultiTouchEvent` itself too?

Thanks again!

dsa-t:
> Maybe we should remove it and just add position to `wxMultiTouchEvent` itself too?

Makes sense to me.

src/osx/cocoa/window.mm
dsa-t: `i` is used for return value.

samples/touchtest/touchtest.cpp
dsa-t: With `wxDefaultSize`, the frame becomes too small.

vadz: Then we need to do something like

```cpp
SetClientSize(FromDIP(wxSize(500, 400)));
```

in the ctor (we can't use `FromDIP()` here as the window doesn't exist yet).

dsa-t:
If finger is held for too long on MSW, the OS stops synthesizing mouse move events.

So might need to add a `IsSynthesized` method to mouse events.
The `IsPrimary` call is not reliable in this case.

Also in Qt, the primary touch point detection is not available.

dsa-t:
Also in Qt, a `nomousefromtouch` windows platform parameter can be passed to skip these synthesized events at Qt level
(see https://doc.qt.io/qt-6/qguiapplication.html#platform-specific-arguments)

dsa-t:
On my GTK system, the mouse pointer emulation code from https://github.com/wxWidgets/wxWidgets/pull/2539 doesn't seem to be needed.

If I return `false` from `touch_callback`, GTK sends the mouse events by itself.
If I return `true`, GTK doesn't send any emulated mouse events.

Current `touch_callback` has a `void` return type.

Could it be that the `void` function return value evaluated to something other than `false` in https://github.com/wxWidgets/wxWidgets/issues/19265?

dsa-t:
Another question is if the user app has handled the touch event, should we still emit mouse events synthesized by the OS or block them?

dsa-t:
If we return `false` in `touch_callback`, the event order is interesting:

```
OnMouseMove leftdown 0 pos 224 439

touch_callback Emulating 1 widget 0x555555859120 dev 0x555555634820 dev 0 

OnMouseMove leftdown 1 pos 224 439
OnMouseDown leftdown 1 pos 224 439

touch_callback Emulating 1 widget 0x555555859120 dev 0x555555634820 dev 0 

OnMouseUp leftdown 0 pos 224 439
```

There's a `OnMouseMove` with `LeftIsDown() == true` coming before `OnMouseDown`.

samples/touchtest/touchtest.cpp
dsa-t: Doesn't it cause additional conversion at runtime on wchar systems, which may also be non-threadsafe on macOS?
(we've had to add wxT/wxS to KiCad everywhere because there were random crashes on macOS, and probably no one made a repro case and reported it)

vadz: The tiny overhead of converting to `wchar_t` is not worth the ugliness of `wxT()` in non-performance-sensitive code and constructing `wxString` from `char*` should really be safe, if there are any problems here we really, really ought to fix them, but hopefully there aren't any...

src/gtk/window.cpp
dsa-t: Should we still call `wxEmulateLeftDownEvent`/`wxEmulateMotionEvent`/`wxEmulateLeftUpEvent`, even if the touch event was handled?

Note that MSW sends synthesized mouse events regardless.

vadz:
@dsa-t Do you plan to make more changes here?

src/msw/window.cpp
dsa-t: The value of `processed` doesn't appear to have any effect here.

samples/touchtest/touchtest.cpp
dsa-t: It turned out to be useful to check if the pos is relative to the client rect, not the window rect.

dsa-t: Maybe some options can be added later to the menu.

samples/touchtest/touchtest.cpp
dsa-t: The corresponding commit in KiCad was https://gitlab.com/kicad/code/kicad/-/commit/7601a3385fb326ad795a3a42a2344517f7dfcbd3

Here's one of the stack traces from Feb 2022:

```
==1255500==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7fd74c64a400 at pc 0x7fd75d5c501b bp 0x7fd74c64a3e0 sp 0x7fd74c649b90
READ of size 8 at 0x7fd74c64a400 thread T0
    #0 0x7fd75d5c501a in __interceptor_mbsrtowcs ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:3419
    #1 0x7fd75c9e981d in wxMB2WC(wchar_t*, char const*, unsigned long) (/usr/lib/x86_64-linux-gnu/libwx_baseu-3.0.so.0+0x19381d)
    #2 0x7fd75c99845c in wxMBConv::ToWChar(wchar_t*, unsigned long, char const*, unsigned long) const (/usr/lib/x86_64-linux-gnu/libwx_baseu-3.0.so.0+0x14245c)
    #3 0x7fd75c99955f in wxMBConv::cMB2WC(char const*, unsigned long, unsigned long*) const (/usr/lib/x86_64-linux-gnu/libwx_baseu-3.0.so.0+0x14355f)
    #4 0x7fd75c9a7e1e in wxString::ConvertStr(char const*, unsigned long, wxMBConv const&) (/usr/lib/x86_64-linux-gnu/libwx_baseu-3.0.so.0+0x151e1e)
    #5 0x7fd74940cecf in wxString::ImplStr(char const*, wxMBConv const&) /usr/include/wx-3.0/wx/string.h:477
    #6 0x7fd74940d6a8 in wxString::operator=(char const*) /usr/include/wx-3.0/wx/string.h:1921
    #7 0x7fd74aa517a9 in NESTED_SETTINGS::LoadFromFile(wxString const&) ../../common/settings/nested_settings.cpp:81
    #8 0x7fd74aa53358 in NESTED_SETTINGS::SetParent(JSON_SETTINGS*, bool) ../../common/settings/nested_settings.cpp:164
    #9 0x7fd74aa4f606 in NESTED_SETTINGS::NESTED_SETTINGS(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, JSON_SETTINGS*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ../../common/settings/nested_settings.cpp:31
    #10 0x7fd749ed911b in SCHEMATIC_SETTINGS::SCHEMATIC_SETTINGS(JSON_SETTINGS*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ../../eeschema/schematic_settings.cpp:54
    #11 0x7fd749eceafa in SCHEMATIC::SetProject(PROJECT*) ../../eeschema/schematic.cpp:95
```

[nested_settings.cpp:81](https://gitlab.com/kicad/code/kicad/-/blob/ccf2a63f45f2a33bb69bf35da2f06a694f1b0115/common/settings/nested_settings.cpp#L79)

It might've been some issue in KiCad code as well.

dsa-t:
Another questions are:

- Should we use floating point for precise positioning in high-DPI screens?
- Maybe get rid of `wxTouchSequenceId` and use `int`, like [Qt does](https://doc.qt.io/qt-6/qeventpoint.html#id-prop)?

samples/touchtest/touchtest.cpp
vadz: It's difficult to know what the problem was, but I'm almost certain it's indeed something in KiCad and not wx because constructing `wxString` from `char*` is absolutely supposed to work and is done in many, many places in most of projects using wx.

IOW the commit message

> WxString does not allocate space for wide strings needed during
conversion unless the string is explicitly wide.  This can cause buffer
over/underflow

is wrong and the error message above says about buffer overflow on _read_, i.e. it looks like there is something wrong with the input string (`m_filename`).

vadz:
>     * Should we use floating point for precise positioning on high-DPI screens?

What do the different OS provide to us, i.e. what type of coordinates they use?
 
>     * Maybe get rid of `wxTouchSequenceId` and use `int`, like [Qt does](https://doc.qt.io/qt-6/qeventpoint.html#id-prop)?

I'm always in favour of using opaque types if possible. I'm still not sure what exactly is `wxTouchSequenceId` for, i.e. how will it be used, but it doesn't seem like it should be just a plain integer.

dsa-t:
> What do the different OS provide to us, i.e. what type of coordinates they use?
> 
> >

Logical pixels == DIPs on wxQt, so it would be most noticable on smartphones.




> I'm still not sure what exactly is `wxTouchSequenceId` for, i.e. how will it be used, but it doesn't seem like it should be just a plain integer.

It's an arbitrary value to identify touch points. On wxMSW it's something like:

```
0x0000000000003763
0x0000000000003764
0x0000000000003765
0x0000000000003766
0x0000000000003767
0x0000000000003768
0x0000000000003769
0x000000000000376a
0x000000000000376b
0x000000000000376c
```

on wxQt on Windows it's something like:

```
0x0000000002000001
0x0000000002000002
0x0000000002000003
0x0000000002000004
0x0000000002000005
0x0000000002000006
0x0000000002000007
0x0000000002000008
0x0000000002000009
0x000000000200000a
```

On wxGTK, it's a `GdkEventSequence*` (https://docs.gtk.org/gdk3/struct.EventTouch.html), but it's value it usually not a real pointer - it acts as an int.

vadz:
> > What do the different OS provide to us, i.e. what type of coordinates they use?
> 
> Logical pixels == DIPs on wxQt, so it would be most noticable on smartphones.

Sorry, what I meant was whether it was integer or floating point?
 
> > I'm still not sure what exactly is `wxTouchSequenceId` for, i.e. how will it be used, but it doesn't seem like it should be just a plain integer.
> 
> It's an arbitrary value to identify touch points.

Thanks for the explanation and the examples. If we really can't do anything with it, except comparing them, presumably, it should remain opaque and allow doing just this (and copying).

dsa-t:
> Sorry, what I meant was whether it was integer or floating point?

on GTK, it's 

```
  gdouble x;
  gdouble y;
```

on Qt, it's `qreal` aka `double` in `QPointF`:

```
QPointF TouchPoint::screenPos() const
```

vadz:
In this case I agree that we should use doubles in wx API too.

dsa-t:
Changed position type to `wxPoint2DDouble`.
The way `ScreenToClient` is used is not great.
Should `ScreenToClient` be implemented for floatinig point separately, and is there other suggestions?

CI Error in `Cirrus CI / FreeBSD wxGTK 3` seems unrelated.

dsa-t:
`wxMouseEvent` positions are floating point in GTK, Qt and OSX.
Should wx also use floating point types for these?

samples/touchtest/touchtest.cpp
```diff
@@ -0,0 +1,251 @@
...
+class MyFrame: public wxFrame
+{
+private:
+    struct TouchState
+    {
+        wxTouchSequenceId id;
+        wxPoint2DDouble last;
+        wxPen pen;
+    };
+    TouchState m_TouchPoints[TOUCH_POINTS];
```
vadz, line 34: Minor but

```suggestion
    std::array<TouchState, TOUCH_POINTS> m_TouchPoints;
```

would be a bit nicer.

samples/touchtest/touchtest.cpp
```diff
@@ -0,0 +1,251 @@
...
+wxIMPLEMENT_APP(wxMyApp);
+
+bool wxMyApp::OnInit()
+{
+    if ( !wxApp::OnInit() )
+        return false;
+
+    wxFrame* frame = new MyFrame(nullptr, "Multi-touch Test", wxDefaultPosition,
+        wxDefaultSize, wxDEFAULT_FRAME_STYLE | wxHSCROLL | wxVSCROLL);
+    SetTopWindow(frame);
```
vadz, line 72:
```suggestion
```

samples/touchtest/touchtest.cpp
```diff
@@ -0,0 +1,251 @@
...
+    EVT_LEFT_UP(MyFrame::OnMouseUp)
+
+    EVT_PAINT(MyFrame::OnPaint)
+    EVT_SIZE(MyFrame::OnSize)
+wxEND_EVENT_TABLE()
+
+MyFrame::MyFrame(wxFrame *parent, const wxString& title, const wxPoint& pos,
+    const wxSize& size, const long style)
+    : wxFrame(parent, wxID_ANY, title, pos, size, style),
+    m_MouseId(wxUIntToPtr(0xFFFFFFFF))
```
vadz, line 96: Is there some special significance to using `-1` here?

src/msw/textctrl.cpp
```diff
@@ -2809,7 +2809,7 @@ bool wxTextCtrl::MSWOnNotify(int idCtrl, WXLPARAM lParam, WXLPARAM *result)
                     int x = GET_X_LPARAM(msgf->lParam),
                         y = GET_Y_LPARAM(msgf->lParam);
 
-                    HandleMouseEvent(msg, x, y, flags);
+                    HandleMouseEvent(msg, x, y, flags, ::GetMessageExtraInfo());
```
vadz, line 2812: These changes could be avoided if we called `::GetMessageExtraInfo()` in `HandleMouseEvent()` itself too.

src/osx/cocoa/window.mm
```diff
@@ -28,6 +28,7 @@
 #include "wx/private/bmpbndl.h"
 
 #include "wx/evtloop.h"
+#include "wx/vector.h"
```
vadz, line 31: We can just use `<vector>` on master.

src/msw/window.cpp
```diff
@@ -236,6 +236,27 @@ class GestureFuncs
         return ms_pfnSetGestureConfig;
     }
 
+    typedef BOOL (WINAPI *RegisterTouchWindow_t)(HWND,ULONG);
```
vadz, line 239: Actually we shouldn't need to load these functions dynamically any longer, they're all available since Windows 7 and we don't support anything earlier (or even it) any longer. If you prefer, we can keep this code as is for this PR, but we could (and should) just remove all this dynamic loading stuff later.

src/msw/window.cpp
vadz: If there is no good reason to _not_ use the usual logic, I'd stick to it. I.e. just do

```cpp
processed = HandleTouch(wParam, lParam);
```

src/msw/window.cpp
vadz: This also seems to be still relevant.

vadz (commented):
Thanks for the updates, I think this is pretty close to being merged, but I'd like to understand why do we need to pass the `extra` argument everywhere in the MSW code instead of just getting it when needed. Do we really have to complicate things like this?

vadz:
> `wxMouseEvent` positions are floating point in GTK, Qt and OSX. Should wx also use floating point types for these?

I guess we could add `wxPoint2DDouble m_pos` to `wxMouseState` but we'd still have to keep `m_x` and `m_y` for compatibility and ensure they're always synchronized and I'm not sure if it's worth the complexity — considering how (im)precise the mouse/gestures are, it probably doesn't make much difference in practice unless you use a high-precision tablet. Do you notice any problems due to rounding when testing this?

> The way ScreenToClient is used is not great.
> Should ScreenToClient be implemented for floatinig point separately

Yes, I think it could be nice to add an overload using doubles.

> CI Error in Cirrus CI / FreeBSD wxGTK 3 seems unrelated.

Yes, it was already fixed in master.

Chris574-glitch (approved):

vadz:
I've implemented (most of) my own suggestions, but I can't really test this code as I don't have any multitouch devices. Could you please confirm that it still works for you? If it does, let's merge it.

I'd also like to squash-merge all your commits, if you don't mind (and if you do, please feel free to do interactive rebase yourself to keep those you think it makes sense to keep separate).

TIA!

vadz:
@dsa-t Ping just to check that you've seen this (this is obviously not urgent, but I'd still like to eventually merge it).

dsa-t:
Yes, I'm just a little busy these days. Will get to it eventually.

dsa-t:
Works on Linux Mint/X11 (wxGTK and wxQt)

vadz:
Thanks for testing!

AFAICS there have been no changes and the branch was just rebased on the latest master, right? If so, I'll squash-merge and push it soon.

dsa-t:
> AFAICS there have been no changes and the branch was just rebased on the latest master, right?

Yes.

PR #24614, opened 2024-06-15
Remove GDI+ wrapper

MaartenBent:
Link with gdiplus.lib directly.


vadz:
Hmm, the CI failures seem like a real problem, although I didn't have time to actually look/test it myself yet.

MaartenBent:
And the startup failure is caused by:
https://github.com/wxWidgets/wxWidgets/blob/7f5ca30e3eebbdb85222fbe329a5100b4da65804/src/msw/graphics.cpp#L3052
Because this module doesn't exist anymore.

While searching for gdi, I noticed `dc.cpp` also loads some function from `gdi32.dll` dynamically. What do you think of removing that and using gdi directly?
https://github.com/wxWidgets/wxWidgets/blob/7f5ca30e3eebbdb85222fbe329a5100b4da65804/src/msw/dc.cpp#L177

vadz:
According to the comment, this was done to avoid problems with w32api 4.0.3 which is more than 10 years old, so I think we can indeed drop this now.

We might also start linking with `msimg32.lib` to use `AlphaBlend()` and `GradientFill()` directly, as long as we're adding the new `gdiplus.lib` dependency anyhow...

src/msw/graphics.cpp
vadz, line 1: Should we perhaps add

```cpp
    #pragma comment(lib, "gdiplus")
```

to this file? Then we wouldn't have to modify the tests and utils projects and I don't think this can really do any harm, can it?

vadz (commented):
Thanks!

I think this should be merged but we need to document that the applications not using MSVC now need to link with `gdiplus.lib`. I thought we already had the list of required libraries somewhere in the docs but I can't find it right now, would you remember where it is by chance or have I just imagined that?

MaartenBent:
I don't see any list with libraries, only a similar entry in the changelog;
https://github.com/wxWidgets/wxWidgets/blob/f8043a62f3319719cbd3a400ea454e4dbb3d23ff/docs/changes_32.txt#L158-L163

src/msw/graphics.cpp
MaartenBent: Yes. I'll add this.

src/msw/dc.cpp
```diff
@@ -50,28 +50,9 @@
...
-    // #pragma comment(lib) used above, so we continue loading AlphaBlend()
-    // dynamically, if possible.
-    //
-    // We also load some GDI functions not present in MinGW libraries
-    // dynamically.
-    #include "wx/dynlib.h"
-
-    #define USE_DYNAMIC_GDI_FUNCS
-#endif // _MSC_VER/!_MSC_VER
+    #pragma comment(lib, "gdi32")
```
vadz, line 54: Very minor, but I don't think we ever need this, as we always link with this one anyhow.

vadz (approved):
Thanks again!

Other than the change log entry, this looks good to me, should I merge it?

src/msw/dc.cpp
MaartenBent: There might be some scenarios where it doesn't have gdi32.
For example when building static libraries. And then you use CMake's own FindwxWidgets to find these libraries and link with them.  It doesn't have `gid32` in the win32 [library list](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindwxWidgets.cmake#L470). So without this pragma, the final executable won't know it needs gid32.

There is `include/msvc/wx/setup.h` but I think this is only used if you manually add `include/msvc` as the first include path?

MaartenBent:
Yes, this is ready. You can merge it.

src/msw/dc.cpp
vadz: Sorry, I don't understand how could there be any scenarios in which gdi32 is not linked — if this were possible, we'd be getting link errors already because we use plenty of functions from it non-dynamically. I think CMake might be linking Windows GUI programs with it by default, as this would make sense.

MaartenBent: Ok, I've removed it.

PBfordev:
@MaartenBent, could you please tell me if there is a technical reason why the two libraries should not be added to linking libraries in the official CMake's FindwxWidgets module, similarly to [when uuid and uxtheme started to be required](https://gitlab.kitware.com/cmake/cmake/-/commit/c921ec6112961304824b6bbdf7fb743f676325d0)?

If it would be as simple as adding uuid and uxtheme, I am willing to submit a PR to CMake. Breaking a build is a pain, and every little attempt to make it not happen counts...

MaartenBent:
There is no reason not to add it. Please go ahead and submit a MR/PR there.

PR #24615, opened 2024-06-15
Add MSW VS ARM64 and ARM64EC infra

jkunkee:
This change will, when it's done, add:

* ARM64 and ARM64EC Platform entries to all `build/msw` VS Project files
* ARM64 and ARM64EC Platform entries as appropriate to the `msw_vs{15|16|17}.sln` VS Solution files
* ARM64X note in docs
* ARM64 and ARM64EC entries to the `minimal`, `test`, and `test_gui` projects and solutions
* CI dimensions for ARM64 and ARM64EC builds

In favor of getting it checked in sooner, the following are left to future contributions:

* Attempt to add ARM64EC or X support to the NMake files
* Attempt to add or finish GCC or LLVM support
* Add CI changes to run ARM64 and ARM64EC tests (requires ARM64 GitHub-hosted Action runners)


vadz:
Thanks, I didn't look at all changes, but those that I did look at seem pretty straightforward to me.

I'd extract the changes to `wxregex.vcxproj.filters` in a separate commit as they're not really ARM-related, but otherwise no real comments — please just remove the "Draft" status when you'd like this to get merged.

jkunkee:
I have a few more edits for the docs, then I want to re-scope the description to match, then yes, let's get it in soon. (.filters changes are a matter of preference, so I drop that and review the other files.)

vadz:
No, I the `.filters` change is definitely correct, we (I, probably) just forgot to update it when switching from the old regex library to PCRE2, and it needs to be done, but it would just be better to do it in a separate commit. TIA!

jkunkee:
After some tweaks, this should be ready to merge.

jkunkee:
Hang on, part of this should be quick...

vadz:
It looks like we need to add ARM platform to the minimal sample projects. This shouldn't be difficult to do but I've been thinking about adding project files for all the other samples too (as part of migration from using `makefile.vc`) and updating all of those risks being painful. I wonder if there is some way to centralize the common part of all the samples, somehow, e.g. include some common fragment defining the supported platforms or something?

jkunkee:
I poked around with this a bit. Adding the minimal configuration is hitting an error around not being able to find `wx/setup.h`. As noted with ARM64X build, this bit of infra needs a little love.

1. I can add the `if` logic to the minimal example and test jobs along with the run jobs. This just punts on the configuration problem. I see that Win32 is also missing from the minimal example.
2. I can work through adding the `minimal` configurations now.

Note that NMake already supports ARM64, so that should be usable in the interim.

As for factoring out common parts of project/`vcxproj` files, it's only a little more effort. My personal preference is using the Visual Studio GUI to make a change, then looking at `git diff` to move the common changes into [Directory.Build.props and Directory.Build.targets files (docs link)](https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-by-directory). (I've seen industrial-scale work with lots of recursive property sheets/`.props` files--as long as things are `Import`'d in the correct order, you can do anything--but more diversion from the normal path is more work.)

vadz:
> I poked around with this a bit. Adding the minimal configuration is hitting an error around not being able to find `wx/setup.h`. As noted with ARM64X build, this bit of infra needs a little love.

I was going to write that this needs to be addressed by just adding another arch to `build/msw/wx_setup.props`, but I think you've already done it, so is this still a problem?

>     1. I can add the `if` logic to the minimal example and test jobs along with the run jobs. This just punts on the configuration problem. I see that Win32 is also missing from the minimal example.

Sorry, I don't understand this at all, `samples/minimal/minimal.vcxproj` clearly defines Win32 configurations and I'm pretty sure building them works (even though I stopped using Win32 myself since a couple of years, so I didn't test this recently).
 
> As for factoring out common parts of project/`vcxproj` files, it's only a little more effort.

Yes, this is definitely doable, but it "just" needs to be done. To give some context, we've been wanting to automate the generation of these project files and add them for all the samples, but I've never found a good enough way to do it, so I've finally started just adding them by hand, but it's a laborious process (there are ~100 of them) and I didn't finish yet.

jkunkee:
OK, I finished getting the minimal sample, tests, and GUI tests building, got it rebased onto the merged changes from the other PR, and pushed it. Here's hoping CI is passing now--the tasks pass locally. :)

jkunkee:
> > I poked around with this a bit. Adding the minimal configuration is hitting an error around not being able to find `wx/setup.h`. As noted with ARM64X build, this bit of infra needs a little love.
> 
> I was going to write that this needs to be addressed by just adding another arch to `build/msw/wx_setup.props`, but I think you've already done it, so is this still a problem?

Nope, and it needed to be done anyhow.

> > 1. I can add the `if` logic to the minimal example and test jobs along with the run jobs. This just punts on the configuration problem. I see that Win32 is also missing from the minimal example.
> 
> Sorry, I don't understand this at all, `samples/minimal/minimal.vcxproj` clearly defines Win32 configurations and I'm pretty sure building them works (even though I stopped using Win32 myself since a couple of years, so I didn't test this recently).

I think I misread something in haste (VS isn't consistent about how it orders sections) and yes, it's fine.

> > As for factoring out common parts of project/`vcxproj` files, it's only a little more effort.
> 
> Yes, this is definitely doable, but it "just" needs to be done. To give some context, we've been wanting to automate the generation of these project files and add them for all the samples, but I've never found a good enough way to do it, so I've finally started just adding them by hand, but it's a laborious process (there are ~100 of them) and I didn't finish yet.

This is a hard one. `Directory.Build.props` and `.targets` mean no manual `Import`s and can mean the individual project files can be very small, bit it's still one each and...100 is a lot.

.github/workflows/ci_msw.yml
```diff
@@ -113,6 +116,7 @@ jobs:
             msbuild /noLogo /m '/p:Configuration=${{ matrix.configuration }}' /p:Platform=${{ matrix.platform }} tests\test_vc16.sln
 
       - name: Run tests
+        if: ${{ !contains(matrix.platform, 'arm64') }}
```
vadz, line 119: Do the tests fail under ARM or is there some other reason to exclude them?

jkunkee: (I'm working through setting up and running the tests now, as yes, they should be passing for check-in.)

jkunkee: With the `tests` folder as the working directory and DPI set to 100%, all tests pass for ARM64, ARM64EC, and x64 on ARM64 Windows 11 (26100). (Since the _gui tests rely on pixel counts, non-unity DPI breaks a number of them--but that isn't ARM64-specific and is easy to control for.)

[vc_x64_mswu_test.txt](https://github.com/user-attachments/files/15963715/vc_x64_mswu_test.txt)
[vc_x64_mswu_test_gui.txt](https://github.com/user-attachments/files/15963717/vc_x64_mswu_test_gui.txt)
[vc_arm64_mswu_test.txt](https://github.com/user-attachments/files/15963718/vc_arm64_mswu_test.txt)
[vc_arm64_mswu_test_gui.txt](https://github.com/user-attachments/files/15963719/vc_arm64_mswu_test_gui.txt)
[vc_arm64ec_mswu_test.txt](https://github.com/user-attachments/files/15963720/vc_arm64ec_mswu_test.txt)
[vc_arm64ec_mswu_test_gui.txt](https://github.com/user-attachments/files/15963721/vc_arm64ec_mswu_test_gui.txt)

vadz: Thanks! Do you plan updating this PR to run the tests in the CI builds?

jkunkee: Currently, no. It would require setting up and manually managing runners for the wxWidgets org and moving the tests to a second runner to execute--or just building ARM64 on an ARM64 runner then running the tests locally.

I'm happy to either a) adjust this PR to not touch CI at all until that's possible (or whatever scope you feel is appropriate) or b) wait until GH-hosted ARM64 Windows runners are available. (Technically [they're available now in beta](https://github.blog/2024-06-03-arm64-on-github-actions-powering-faster-more-efficient-build-systems/), but I'm not sure how much you'd have to do from the wxWidgets org perspective to set them up.)

vadz: Thanks, I'll just (squash) merge this for now then (unless you prefer to rebase it to keep your commits).

jkunkee: Sounds good to me; I prefer squash merges myself.

jkunkee: Side note: if no one else beats me to it, I'm happy to come back later and work on setting up ARM64 testing once the GH-hosted runners are available. I'd probably move the ARM64 and ARM64EC builds to ARM64 so there's no shuffling around of generated artifacts. Once some confidence is had, it probably wouldn't be too hard to add ARM64 binaries to releases.

vadz: This would be appreciated, TIA!

vadz:
I wonder if we can do anything about multiple warnings that now appear when opening any MSVS projects in MSVS 2022 version without support for these platforms, e.g.

```
build\msw\wx_base.vcxproj : warning  : Platform 'ARM64EC' referenced in the project file 'base' cannot be found.
build\msw\wx_core.vcxproj : warning  : Platform 'ARM64EC' referenced in the project file 'core' cannot be found.
```

These warnings are harmless but annoying and I think we're going to be getting many questions about them...

jkunkee:
If you are seeing this with earlier VS versions (I can't repro with 2022), then I have a couple of ideas. Let me ask a couple friends some questions and I'll get back to you here.

vadz:
I do use MSVS 2022 but a relatively old version of it because I'm rather conservative with upgrading it (IME this not infrequently results in new problems, such as [this recent one](https://developercommunity.visualstudio.com/t/All-my-std::unique_lock-crashed-after-th/10665376?space=41)), so I'm at

```
Microsoft Visual Studio Enterprise 2022
Version 17.6.5
VisualStudio.17.Release/17.6.5+33829.357
```

jkunkee:
A week to discover the redist has already been fixed isn't too bad, tbh--but that's beside the point. No matter the reason, you're on 17.6.5.

This really shouldn't be a problem since [ARM64EC has been officially supported since 17.3](https://devblogs.microsoft.com/cppblog/official-support-for-arm64ec-is-here/), yet we have this warning.

My guess is that there's some connection in 17.6.5 between the 'what Platforms can exist' logic in Visual Studio and the 'this Platform exists' logic in the potentially absent ARM64EC build tools that's buggy. Do you have the ARM64EC build tools (in current versions it's the ARM64/ARM64EC build tools) installed?

vadz:
No, I don't have ARM64 build tools installed, which, I guess, explains the problem. But I also think that there are many, many other installations without them...

jkunkee:
There are many installations without them, it's true! Note that it does not complain about ARM64, only ARM64EC, and you have neither installed. This is part of why I suspect this is just a straight-up bug in Visual Studio's C/C++ property sheets. If my attempts to build with the ARM64EC tools uninstalled on the latest version was done right, the issue does appear to be fixed.

What could a workaround look like? My first idea would be to add a `Condition` attribute to the ARM64EC `ProjectConfiguration` tag using the `MSBuildVersion` property. (I found this property following [these instructions](https://learn.microsoft.com/en-us/cpp/build/reference/common-macros-for-build-commands-and-properties).) Something like this:

```xml
<ProjectConfiguration Include="Debug|ARM64EC" Condition="$(!MSBuildVersion.StartsWith('17.6.'))">
```

My guess is that the issue is in MSBuild's MSVC property sheets, so MSBuildVersion should be a valid property to use here.

If I didn't make a mistake and it works for you for one project, I'm happy to make the change broadly.

vadz:
Sorry, why should we test for `!StartsWith('17.6')`, wouldn't this pass for 17.5 and earlier?

According to [the docs](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-conditions?view=vs-2022#comparing-versions) we should be able to compare it as version if we provide all 4 components, but my attempts to use something like `$(MSBuildVersion) >= 17.7.0.0` failed: first, `MSBuildVersion` has only 3 components and `MSBuildFullVersion` doesn't seem to exit. And using

```xml
    <ProjectConfiguration Include="Debug|ARM64EC" Condition="'$(MSBuildVersion).0 >= 17.7.0.0'">
```

results in

```
---------------------------
Microsoft Visual Studio
---------------------------
Unable to read the project file "wx_base.vcxproj".
<full-path>\build\msw\wx_base.vcxproj(8,51): Specified condition "'$(MSBuildVersion).0 >= 17.7.0.0'" evaluates to "17.6.3.0 >= 17.7.0.0" instead of a boolean.
---------------------------
OK   
---------------------------
```

Any hints?

vadz:
Ah, wait, as usual it was enough to post to find the solution: it should be

```
Condition="'$(MSBuildVersion).0' >= '17.7.0.0'"
```

and this seems to work, but I still get warnings from other places, trying to find out from where...

vadz:
Unfortunately this doesn't seem to work, even just this "project" file

```xml
<?xml version="1.0" encoding="utf-8"?>
<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <ItemGroup Label="ProjectConfigurations">
    <ProjectConfiguration Include="Debug|ARM64">
      <Configuration>Debug</Configuration>
      <Platform>ARM64</Platform>
    </ProjectConfiguration>
    <ProjectConfiguration Include="Debug|ARM64EC" Condition="'$(MSBuildVersion).0' >= '17.7.0.0'">
      <Configuration>Debug</Configuration>
      <Platform>ARM64EC</Platform>
    </ProjectConfiguration>
  </ItemGroup>
  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
  <ImportGroup Label="ExtensionTargets">
  </ImportGroup>
</Project>
```

results in a warning:

```
build\msw\wx_base.vcxproj : warning  : Platform 'ARM64EC' referenced in the project file 'wx_base' cannot be found.
```

jkunkee:
Just chatted with some friends over in VC++ land, and they had some ideas:

1. Would you be willing to generate and share a binary log of the x64 build hitting this error? The idea is to pinpoint what line of code is generating this warning. Binary logs are incredibly verbose and detailed (think paths and environment variables), though, so no worries if you'd rather not.
2. Have you looked at moving up to the latest 17.6 (17.6.18, I think)? The patch releases have a pretty high bar for testing and release, and it's possible that this is fixed there.
    * Note that 17.6 hits end of life (so, no more updates, even for security) in January 2025 ([lifecycle docs](https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2022)).
    * If this has related functionality problems and they do fix this in 17.6, it would come as just such a patch release.

I'm also looking at spinning up a VM with 17.6.5 to repro the issue and try a couple of things; that version compare looked promising. Not sure when I'll get that done though.

(I used StartsWith because there was a lot of churn in ARM64EC in this timeframe, so I somewhat expect the bug to be specific to 17.6. This is based on vibes, though, and your solution is just semantically better. Also, I didn't know about the version comparison feature. :/ )

vadz:
Just to be sure, I'd like to emphasize that what I'm seeing are not build warnings, i.e. doesn't come from MSBuild at all, AFAICS, but from MSVS itself as these messages only appear when opening the solution, not building it.

As for upgrading, personally, I will do it sooner or later, of course, but I'm also thinking of many people using previous MSVS versions that will have all these warnings when opening wx projects and are going to blame wx for them...

jkunkee:
That is some context I somehow missed earlier. You're right. The message comes from this assembly in Visual Studio (used findstr): `Common7\IDE\CommonExtensions\Microsoft\VC\Project\Microsoft.VisualStudio.Project.VisualC.VCProjectEngine.dll`

Might take me a week or more to get time to poke around with that.

jkunkee:
I fired up a VM this evening and tried this out. With the extra detail provided, I was able to repro the issue.

My repro steps:

1. Use the [fixed-version bootstrapper](https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-history#updating-your-installation-to-a-specific-release) for 17.6.5
2. Select Desktop C++ Development
3. Clone wxWidgets
4. Open `build\msw\wx_vc17.sln`
5. View menu -> Output
6. Output pane -> Output tab
7. Note that "Show output from:" does not list Solution, so the warning is hidden
8. Unload the solution
9. Add a typo to `wx_webview.vcxproj` (selected because it's neither wx_core.vcxproj nor wx_custom_build.vcxproj)
10. Load the solution
11. Observe that Solution shows up in "Show output from:" with the expected parse error and the reported warning for the set Startup Project:

```
C:\Users\jkunkee\source\repos\wxWidgets\build\msw\wx_custom_build.vcxproj : warning  : Platform 'ARM64EC' referenced in the project file '_custom_build' cannot be found.
C:\Users\jkunkee\source\repos\wxWidgets\build\msw\wx_custom_build.vcxproj : warning  : Platform 'ARM64' referenced in the project file '_custom_build' cannot be found.
```

Once VS is open and the Solution output option is visible, the message appears whenever a solution or project is loaded and whenever the Startup Solution is changed. This resets for me when VS is restarted.

This also repros on 17.6.18 and 17.1.11.

A fractional workaround could be to not do anything that causes the Solution output to be open by default. This appears to be the default in VS with tip-of-tree wxWidgets, so I hope that this will, by happenstance, decimate the support burden you expressed concern about.

I consider this a bug. Either the warning should be clear that you have loaded a project with a config for a toolset that isn't installed, or it shouldn't be a warning at all. Unfortunately, since this is out of my ken--I know property sheets and targets files well, but not the bowels of Visual Studio itself--no more workarounds come to mind. The best way to find a workaround and get it fixed is to report it on the [Visual Studio Developer Community](https://developercommunity.visualstudio.com/home). If you make the time to do so, please do share a link to the report here so I can track and even encourage its progress.

PR #24619, opened 2024-06-16
implement GetNewRow/Col for EVT_GRID_ROW/COL_MOVE

DietmarSchwertberger:
replacement for https://github.com/wxWidgets/wxWidgets/pull/24392

This PR implements `GetNewRow()` and `GetNewCol()` for `EVT_GRID_ROW/COL_MOVE` to allow to retrieve the target row/col.
Also, it fixes the bug introduced with https://github.com/wxWidgets/wxWidgets/pull/22260 where `GetRow()` would always return -1.

Implementation detail: e.g. for `EVT_GRID_ROW_MOVE`, m_col is used for the target row.

### DIFF
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -3170,6 +3170,8 @@ public:
 
     int GetRow() const { return m_row; }
     int GetCol() const { return m_col; }
+    int GetNewRow() const { return m_col; }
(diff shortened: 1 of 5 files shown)


vadz:
Looks good to me, thanks! I'll merge it in a moment.

PR #24620, opened 2024-06-17
test ellipse boundary pixels relative to its rectangle coordinates

wsu-cb:


vadz:
Should this be merged, with just the comment being updated to avoid saying that this behaviour is wxMSW-specific?

wsu-cb:
> Should this be merged, with just the comment being updated to avoid saying that this behaviour is wxMSW-specific?

I still think this behavior is weird, and hesitate to enshrine it in the unit tests.  That's why I haven't marked this "Ready for review".

On the other hand, I doubt I can talk the rest of you into changing this, so at least this test will ensure that my workarounds for this behavior won't suddenly cause problems because this behavior changed behind my back.

What do YOU think?

vadz:
I still haven't looked at the details of what's going on here but if we have the same behaviour under all platforms now, we'd have, at the very least, change it under all platforms too, as anything else would clearly be a regression, and I'm not sure if you're really proposing to do this?

Do you have the screenshots of the current and desired behaviour, just to be sure what the proposed change would be? TIA!

wsu-cb:
> I still haven't looked at the details of what's going on here but if we have the same behaviour under all platforms now, we'd have, at the very least, change it under all platforms too, as anything else would clearly be a regression, and I'm not sure if you're really proposing to do this?
> 

It looks "obviously wrong" to me, but it's been this way in wxMSW since 2008, and apparently the other platforms do the same thing, so I doubt I can convince everyone else to change this behavior.

> Do you have the screenshots of the current and desired behaviour, just to be sure what the proposed change would be? TIA!

If one applies [drawing-ellipse-bounds.patch](https://github.com/user-attachments/files/16229577/drawing-ellipse-bounds.patch), the drawing sample will show the rectangles that corresponds to the DrawEllipse() calls in the drawing sample.  In the current code, the result is

ellipse-a

If commit 45c6df33eccc9c91df7057dbb57dee38b418d906 is reverted, the result is

ellipse-b

If one looks closely, one can see that the ellipse left and top and rectangle left and top match in both displays.  However, the ellipse right and bottom extend outside the rectangle in the current display, but not the display with commit 45c6df33eccc9c91df7057dbb57dee38b418d906 reverted.  I expect the ellipse and rectangle bounds to match, even though apparently they don't match in any of the ports.

vadz:
The comment in this commit (also present in the original #9887) explains why this was done: using `w = h = 1` should draw a single point and, apparently, without this commit, it doesn't do it, but this definitely seems desirable. The previous discussion also seems to say that the current behaviour is the same as when using `wxGraphicsContext` under MSW (back then only GDI+-based version existed, I wonder if the Direct2D one also behaves in the same way).

If we somehow can ensure all of the following:

1. Ellipses with `w = h = 1` produce a single point.
2. `wxDC` and `wxGC` produce the same results under MSW.
3. `wxDC` produces the same results under all platforms.
4. Ellipse remains inside a rectangle of the same width/height.

it would be great, of course, but I'm not sure if this is possible...

wsu-cb:
The drawing sample actually has options to use wxDC, GDI+, and Direct2D in the Drawing menu.  At least on my system, using those options in that order causes the overall appearance of the ellipses to become progressively heavier.  However, that's not absolutely true for all specific pixels.  In particular, the Direct2D drawing of the unfilled circle in the second to last row actually isn't completely closed; it's missing the right-most pixel!

Unfortunately for me, in all three cases, the ellipses escape the rectangles, and this test says it is true for all the platforms.  So I will have to accept this.

PR #24622, opened 2024-06-18
build: fix wx-config symlink in out-of-tree installs

yann-morin-1998:
When we install out-of-tree, i.e. with a non-empty DESTDIR, the wx-config symlink is broken: the target of the symlink is not rooted in DESTDIR.

It would seem the obvious solution would be to prefix the target of the symlink with $(DESTDIR) and be done with it. That would solves cross-compilation cases, but would not fix native build when the install is done in a staging area to prepare a binary package (e.g. Linux distributions à-la Debian do that for example).

A way to make the symlink work in both cases is by making it relative. ln has been supporting a --relative option since coreutils 8.16, released 2012-03-26, more than 12 years ago now; it is relatively safe to assume that distributions of today do have an ln that supports that flag.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -13790,7 +13790,8 @@ install-wxconfig:
 	$(INSTALL_DIR) $(DESTDIR)$(bindir)
 	$(INSTALL_DIR) $(DESTDIR)$(libdir)/wx/config
 	$(INSTALL_SCRIPT) lib/wx/config/$(TOOLCHAIN_FULLNAME) $(DESTDIR)$(libdir)/wx/config
-	(cd $(DESTDIR)$(bindir) && rm -f wx-config && $(LN_S) $(libdir)/wx/config/$(TOOLCHAIN_FULLNAME) wx-config || cp -p $(DESTDIR)$(libdir)/wx/config/$(TOOLCHAIN_FULLNAME) wx-config)
+	# Warning! If the --relative symlink fails (e.g. *BSD), then we'd create an absolute symlink, which will be broken if DESTDIR is moved
+	(cd $(DESTDIR)$(bindir) && rm -f wx-config && $(LN_S) --relative $(DESTDIR)$(libdir)/wx/config/$(TOOLCHAIN_FULLNAME) wx-config || $(LN_S) $(DESTDIR)$(libdir)/wx/config/$(TOOLCHAIN_FULLNAME) wx-config || cp -p $(DESTDIR)$(libdir)/wx/config/$(TOOLCHAIN_FULLNAME) wx-config)
 
 locale_install: 
 	$(INSTALL_DIR) $(DESTDIR)$(datadir)/locale
--- a/build/bakefiles/wx.bkl
+++ b/build/bakefiles/wx.bkl
@@ -96,7 +96,8 @@
                 $(INSTALL_DIR) $(DESTDIR)$(BINDIR)
                 $(INSTALL_DIR) $(DESTDIR)$(LIBDIR)/wx/config
                 $(INSTALL_SCRIPT) lib/wx/config/$(TOOLCHAIN_FULLNAME) $(DESTDIR)$(LIBDIR)/wx/config
-                (cd $(DESTDIR)$(BINDIR) &amp;&amp; rm -f wx-config &amp;&amp; $(LN_S) $(LIBDIR)/wx/config/$(TOOLCHAIN_FULLNAME) wx-config || cp -p $(DESTDIR)$(LIBDIR)/wx/config/$(TOOLCHAIN_FULLNAME) wx-config)
+                # Warning! If the --relative symlink fails (e.g. *BSD), then we'd create an absolute symlink, which will be broken if DESTDIR is moved
+                (cd $(DESTDIR)$(BINDIR) &amp;&amp; rm -f wx-config &amp;&amp; $(LN_S) --relative $(DESTDIR)$(LIBDIR)/wx/config/$(TOOLCHAIN_FULLNAME) wx-config || $(LN_S) $(DESTDIR)$(LIBDIR)/wx/config/$(TOOLCHAIN_FULLNAME) wx-config || cp -p $(DESTDIR)$(LIBDIR)/wx/config/$(TOOLCHAIN_FULLNAME) wx-config)
             </command>
         </action>
 


vadz:
Thanks, I agree that this seems reasonable, but let's see if there are any objections to it from anybody before merging.

cc @swt2c as the closest to the packaging expert that we have (I think, sorry if I'm forgetting anybody)

swt2c:
It does seem reasonable.  However, we have users on non-Linux platforms too (macOS, BSDs, Solaris?? etc).  Do those platforms support `--relative` in their ln implementations, too?

yann-morin-1998:
> It does seem reasonable. However, we have users on non-Linux platforms too (macOS, BSDs, Solaris?? etc). Do those platforms support `--relative` in their ln implementations, too?

All the Mac workflows, all the iOS workflows, all the Windows workflows passed. So either they do have an `ln` that supports `--relative`, or they don't use that part of the buildsystem.

I could not spot a BSD-specific workflow, though, so that's an open question... How can we test that?

yann-morin-1998:
An alternative would be to do a hardlink rather than a symlink (assuming `$(LN)` is not `ln -s`):

```
(cd $(DESTDIR)$(BINDIR) &amp;&amp; rm -f wx-config &amp;&amp; $(LN) $(DESTDIR)$(LIBDIR)/wx/config/$(TOOLCHAIN_FULLNAME) wx-config || cp -p $(DESTDIR)$(LIBDIR)/wx/config/$(TOOLCHAIN_FULLNAME) wx-config)
```

With a hardlink, the result would always work:
- if doing cross-compilation, that is working because neither path are moved
- if the paths are moved, either the hardlinks is kept (yeah!), or the hardlink is broken and this is the same as the fallback `cp` we have already, so in the end that would be the same
- if doing out-of-tree staging for packaging à-la distro, then the resulting binary package would install hardlinks (assuming the archver supports hardlink, but it's rather common; debian uses tarball-in-ar and tar has supported hardlinks forever)

If that is deemed a better approach, I can rework the patch.

vadz:
> > It does seem reasonable. However, we have users on non-Linux platforms too (macOS, BSDs, Solaris?? etc). Do those platforms support `--relative` in their ln implementations, too?

Unfortunately they don't, at least `ln` doesn't have such option under macOS 14.

> All the Mac workflows, all the iOS workflows, all the Windows workflows passed. So either they do have an `ln` that supports `--relative`, or they don't use that part of the buildsystem.

It's probably the latter, we only test `make install` in Unix CI workflow. We should add it to the Mac ones too.
 
> I could not spot a BSD-specific workflow, though, so that's an open question... How can we test that?

But I don't really understand this one: we do have https://github.com/wxWidgets/wxWidgets/blob/a2d33fc31e7612e76163c67ebdf05179a96db84f/.cirrus.yml#L91 in the FreeBSD workflow, so either FreeBSD does have this option (unlike macOS) or I'm missing something here.

vadz:
Re using hardlinks: would this work under MSYS2/Cygwin/other Windows environments?

yann-morin-1998:
> > I could not spot a BSD-specific workflow, though, so that's an open question... How can we test that?
> 
> But I don't really understand this one: we do have
> 
> https://github.com/wxWidgets/wxWidgets/blob/a2d33fc31e7612e76163c67ebdf05179a96db84f/.cirrus.yml#L91
> in the FreeBSD workflow, so either FreeBSD does have this option (unlike macOS) or I'm missing something here.

Ah, I missed the FreeBSD workflow. It indeed does not fail, even though `ln` fails:

```
ln: illegal option -- -
usage: ln [-s [-F] | -L | -P] [-f | -i] [-hnv] source_file [target_file]
       ln [-s [-F] | -L | -P] [-f | -i] [-hnv] source_file ... target_dir
       link source_file target_file
```

So, `ln` fails, and we fallback to doing a `cp` as the code did even before my patch: if `ln` fails, do a `cp`. I agree that it is not so nice, and even though it is kinda the intended behaviour, I would understand that's not acceptable.

Re. hardlinks on Windows/Cygwin, I havenot touched a Windows machine in 20+ years now, so I have no idea what the status his...

yann-morin-1998:
It would seem [hardlinks on Windows are possible](https://learn.microsoft.com/en-us/windows/win32/fileio/hard-links-and-junctions).

vadz:
In principle they're possible but I don't know if it's going to work in practice.

Could we solve the problem by just providing the relative path to `ln -s` instead?

yann-morin-1998:
> Could we solve the problem by just providing the relative path to `ln -s` instead?

Do you have a simple way to compute the relative path from `$(bindir)` to `$(libdir)`, that is widely available?

coreutils also provide `realpath`, which can compute that:
```
$ realpath -s --relative-to="$file1" "$file2"
```
but I'm afraid (I did not bother to check) that would be the same issue as with `ln --relative`, though.

Apart from that, it would involved a lot of shell trickery, and it would be quite easy to miss corner cases... :-/

Coming back to the basics: the code currently does a copy if it can do a symlink. Why can't that behaviour be kept and relied upon: the copy is done if we can't do a hardlink (or a rleative symlink)?

vadz:
> > Could we solve the problem by just providing the relative path to `ln -s` instead?
> 
> Do you have a simple way to compute the relative path from `$(bindir)` to `$(libdir)`, that is widely available?

No, I forgot that those could be customized independently and not be set to the default `$prefix/{bin,lib}`...

> Coming back to the basics: the code currently does a copy if it can't do a symlink. Why can't that behaviour be kept and relied upon: the copy is done if we can't do a hardlink (or a rleative symlink)?

I think doing `ln -s --relative` with fall back to just `ln -s` with fall back to `cp` should be fine — and like this we can be sure not to introduce any regressions (it seems unlikely but not totally impossible that some Mac package relies on `wx-config` being a symlink rather than a copy).

yann-morin-1998:
> I think doing `ln -s --relative` with fall back to just `ln -s`

In case `ln -s --relative` but `ln -s` succeeds, that would still create an invalid symlink, i.e. the current situation. I think this would not be a proper fix...

vadz:
It would fix using `DESTDIR` under Linux. I don't know if it's really used much, or at all under Mac.

yann-morin-1998:
@vadz I've now pushed anew iteration that implements your final suggestion, with a big warning about the limitation of that code. Thank you for the whole guidance on this change! 👍 

@swt2c Thanks for the feedback as well! 🙂

yann-morin-1998:
I just fixed a typo in the commit log...

vadz:
Thanks, will apply in a moment with just a slightly edited message (I'm removing the "Note" because it's not the same variable and regenerating `Makefile.in` automatically would give the same result anyhow).

PR #24626, opened 2024-06-19
Add native RTF support to wxTextCtrl (macOS)

Blake-Madden:
This adds:

- `GetRTFValue()` and `SetRTFValue()` functions to get and set raw RTF
- Changes `wxTEXT_TYPE_ANY` into a `wxTextCtrlFileType` enum that includes a new `wxTEXT_TYPE_RTF` option
- `LoadFile()` and `SaveFile()` now use their `fileType` parameter to save as either text or RTF (if supported)
- Demonstrates these new feature in the "text" sample
- Simple unit test

P.S.
An MSW implementation for the RICHEDIT2 control will be forthcoming in another PR if this is accepted.


Blake-Madden:
I think the AppVeyor test failure was just a sporadic issue.

include/wx/osx/cocoa/private/textimpl.h
```diff
@@ -60,6 +60,15 @@ public :
 
     virtual wxString GetStringValue() const override ;
     virtual void SetStringValue( const wxString &str) override ;
+    virtual wxString GetRTFValue() const override
+    {
+        wxFAIL_MSG("GetRTFValue() should only be used with multiline controls.");
+        return wxEmptyString;
+    };
```
vadz, line 67: Minor, but this will result in warnings with gcc `-pedantic`:

```suggestion
    }
```

include/wx/osx/cocoa/private/textimpl.h
```diff
@@ -60,6 +60,15 @@ public :
...
+    virtual void SetRTFValue(const wxString& WXUNUSED(str)) override
+    {
+        wxFAIL_MSG("SetRTFValue() should only be used with multiline controls.");
+    };
```
vadz, line 71:
```suggestion
    }
```

include/wx/textctrl.h
```diff
@@ -103,7 +103,12 @@ const wxTextCoord wxInvalidTextCoord    = -2;
 // wxTextCtrl file types
 // ----------------------------------------------------------------------------
 
-#define wxTEXT_TYPE_ANY     0
+enum wxTextCtrlFileType
```
vadz, line 106: I think we ought to use scoped enum for this one.

include/wx/textctrl.h
```diff
@@ -680,6 +685,17 @@ class WXDLLIMPEXP_CORE wxTextAreaBase
     virtual wxString GetValue() const = 0;
     virtual void SetValue(const wxString& value) = 0;
 
+    // overridden implementations are required to actually handle RTF
+    virtual wxString GetRTFValue() const
+    {
+        wxFAIL_MSG("GetRTFValue() should only be used with rich, multiline controls.");
```
vadz, line 691: Is this message really correct? I think it should say "not implemented for the current platform".

I'd also move this (and the function below) into `src/common/textcmn.cpp`, there doesn't seem to be any good reason to define them inline.

include/wx/textctrl.h
```diff
@@ -680,6 +685,17 @@ class WXDLLIMPEXP_CORE wxTextAreaBase
...
+        return wxEmptyString;
+    }
+    virtual void SetRTFValue(const wxString& WXUNUSED(val))
```
vadz, line 694: I think this should return `bool` as it looks like there could be errors here — e.g. what if the RTF is syntactically invalid?

interface/wx/textctrl.h
```diff
@@ -1623,6 +1641,37 @@ class wxTextCtrl : public wxControl,
...
+    */
+    wxString GetRTFValue() const;
+
+    /**
+        Sets the content of a multiline text control from an RTF
+        (Rich Text Formatted) buffer.
+
+        This offers more granular control of content formatting, as well as a
+        significant performance benefit with larger content. This also provides the
+        ability read an RTF file and move it directly into the control.
```
vadz, line 1662:
```suggestion
        ability to read an RTF file and move it directly into the control.
```

samples/text/text.cpp
```diff
@@ -1668,6 +1698,35 @@ void MyFrame::OnFileLoad(wxCommandEvent& WXUNUSED(event))
...
+{
+    if (m_panel->m_textrich->SaveFile("dummy.rtf", wxTEXT_TYPE_RTF))
+    {
+#if wxUSE_FILE
+        // verify that the file length is correct
+        wxFile file("dummy.rtf");
+        wxLogStatus(this,
+            "Successfully saved file (text len = %lu, file size = %ld)",
+            (unsigned long)m_panel->m_textrich->GetValue().length(),
+            (long)file.Length());
```
vadz, line 1711: We should be able to avoid casts here:
```suggestion
            "Successfully saved file (text len = %zu, file size = %lld)",
            m_panel->m_textrich->GetValue().length(),
            file.Length());
```

vadz (commented):
Thanks, this looks good, but I believe we might want to add some way of checking whether RTF support is available, either at compile- or run-time, i.e. it could be as simple as defining `wxHAS_RTF_IN_TEXTCTRL` or adding `static bool wxTextCtrl::IsRTFSupported()`, what do you think?

Blake-Madden:
> Thanks, this looks good, but I believe we might want to add some way of checking whether RTF support is available, either at compile- or run-time, i.e. it could be as simple as defining `wxHAS_RTF_IN_TEXTCTRL` or adding `static bool wxTextCtrl::IsRTFSupported()`, what do you think?

I'm liking `static bool wxTextCtrl::IsRTFSupported()` myself; it's more modern than another `#define` :)

include/wx/osx/cocoa/private/textimpl.h
Blake-Madden: Got it.

include/wx/osx/cocoa/private/textimpl.h
Blake-Madden: Got it.

interface/wx/textctrl.h
Blake-Madden: Got it.

samples/text/text.cpp
Blake-Madden: Got it

include/wx/textctrl.h
Blake-Madden: I was wanting to do that originally, but that would break the API for `LoadFile(const wxString&, int)` and `SaveFile(const wxString&, int)`; the `int` would need to become a `wxTextCtrlFileType`. I could do some static_cast'ing inside these functions, but that defeats the purpose of using a scoped enum. Thoughts?

include/wx/textctrl.h
Blake-Madden: Got it.

include/wx/textctrl.h
Blake-Madden: The underlying function returns void:

https://developer.apple.com/documentation/appkit/nstext/1527249-replacecharactersinrange

and I've never seen an error state raised when I pass in malformed RTF. It just silently does what it can with the content it seems.  The RICHEDIT2 control behaves the same way.

include/wx/textctrl.h
vadz: It's difficult to know if anybody really passes an `int` to these functions, knowing that there was just a single possible value for it so far, but I guess you're right and somebody must be doing it or, worse, passing just `0` here, so I take this back, let's keep your original approach, sorry.

Blake-Madden:
OK, I think everything is ready now. Thanks.

samples/text/text.cpp
```diff
@@ -283,6 +283,8 @@ class MyFrame: public wxFrame
 #endif // wxUSE_LOG
     void OnFileSave(wxCommandEvent& event);
     void OnFileLoad(wxCommandEvent& event);
+    void OnFileSaveRtf(wxCommandEvent& event);
```
PBfordev, line 286: Why is in these two methods RTF spelled "Rtf", unlike everywhere else?

samples/text/text.cpp
```diff
@@ -1668,6 +1698,35 @@ void MyFrame::OnFileLoad(wxCommandEvent& WXUNUSED(event))
...
+void MyFrame::OnFileSaveRtf(wxCommandEvent& WXUNUSED(event))
+{
+    if (m_panel->m_textrich->SaveFile("dummy.rtf", wxTEXT_TYPE_RTF))
+    {
+#if wxUSE_FILE
+        // verify that the file length is correct
+        wxFile file("dummy.rtf");
+        wxLogStatus(this,
+            "Successfully saved file (text len = %lu, file size = %ld)",
+            (unsigned long)m_panel->m_textrich->GetValue().length(),
```
PBfordev, line 1710: The casts have not been resolved here, only in OnFileSave()?

samples/text/text.cpp
Blake-Madden: Fixed, thanks.

samples/text/text.cpp
Blake-Madden: Fixed, thanks.

vadz:
Thanks for the updates!

I've pushed some minor changes, repeating the commit message (as it will get squashed here):

- Define `wxHAS_TEXTCTRL_RTF` instead of testing for wxOSX.
- Rename `wxTEXT_TYPE_TEXT` to `wxTEXT_TYPE_PLAIN` and make its value different from `wxTEXT_TYPE_ANY`.
- Use `switch` to ensure that we handle all `wxTextCtrlFileType` values.
- Minor style/formatting fixes.

Please let me know if you have any objections to/comments on any of them. Otherwise I think we should be able to merge this.

vadz:
FYI, I've also merged master into this branch and resolved the conflict in the change log to let the CI builds run.

Blake-Madden:
> Thanks for the updates!
> 
> I've pushed some minor changes, repeating the commit message (as it will get squashed here):
> 
>     * Define `wxHAS_TEXTCTRL_RTF` instead of testing for wxOSX.
> 
>     * Rename `wxTEXT_TYPE_TEXT` to `wxTEXT_TYPE_PLAIN` and make its value different from `wxTEXT_TYPE_ANY`.
> 
>     * Use `switch` to ensure that we handle all `wxTextCtrlFileType` values.
> 
>     * Minor style/formatting fixes.
> 
> 
> Please let me know if you have any objections to/comments on any of them. Otherwise I think we should be able to merge this.
> 
> Thanks again for your work on this!

These all sound great, no objections here. Thanks!

PR #24630, opened 2024-06-21
Fix touch gestures test navigation in event sample, wxQt gesture fix.

dsa-t:
In the `event` sample:
- Panning while scaled didn't pan by the correct amount.
- Combination of gesture events didn't work correctly.

This also allows panning via zoom event position.

In wxQt:
- Fixes the zoom factor that was not set correctly (we need total scale factor, not relative).
- Implements the rotate gesture call.
- Minor simplification in panTriggered.


vadz:
I didn't test it, but looks good to me.

@AliKet any comments?

AliKet:
> @AliKet any comments?

Sorry I can't test this right now, but the changes LGTM. sorry again!

PR #24632, opened 2024-06-22
Always automatically define WX_PRECOMP for MSVC when not disabled

### LINKED ISSUE #24353: wxprec.h not working when using wxWidgets 3.2 from vcpkg with Visual Studio 2022
### Build System Used

- [X] vcpkg / CMake
- [ ] configure
- [X] MSBuild (Microsoft Visual Studio solution file)
- [ ] nmake with makefile.vc
- [ ] Xcode

### Description
My application compiles fine with a custom buildscript for wxWidgets 3.0.5 using plain "nmake" under Windows.
I am in the process of updating my application to use wxWidgets 3.2.4 from the vcpkg port.
My application uses precompiled header files and includes `wxprec.h` to include the most common wxWidgets headers.

My problem is that now I get thousands (literally) of undefined symbols, because standard headerfiles such as `wx/stattext.h` or `wx/menu.h`, which used to be included automatically in `wx/wxprec.h` are not longer included. `#include "wx/wxprec.h` doesn't really do anything anymore.

The problem is that WX_PRECOMP will no longer get defined. My guess is that thats because the vcpkg port builds wxWidgets with vcpkg_cmake_configure.

The wxprec.h contains a somewhat cryptic comment: https://github.com/wxWidgets/wxWidgets/blob/f24b3d54833a1c15ae3bba87dead83bf9a90c41d/include/wx/wxprec.h#L16

Blame (https://github.com/wxWidgets/wxWidgets/commit/248eaddf7a14cefa161d514763e8fefb8d6f2f79) shows that this was added in 2005
// If user did not request NOCPH and we're **not building using configure**
    // then assume user wants precompiled headers.
    #if !defined(NOPCH) && !defined(\_\_WX_SETUP_H__)
        #define WX_PRECOMP
    #endif

In other words, if \_\_WX_SETUP_H__ is defined (which is supposedly the case when building with configure), WX_PRECOMP will not be defined and subsequently all the usual wx header files will also not be included.

I couldn't find any comments or infos on why this has been added to `wxprec.h`. I really have no idea why the original author wanted to prevent precompiled headers when building with configure.

My Question: 
- Why is this necessary? 
- And how should the vcpkg port be modified so that wxprec.h will work again when using it from vcpkg with Visual Studio?

As a workaround I can define the WX_PRECOMP symbol in a Directory.build.props for all projects using wxWidgets, but I think vcpkg builds should work like "normal" builds.

- wxWidgets version 3.2.4
- wxWidgets port wxMSW
- OS Windows 11
- Compiler MSVS 2022
- vcpkg January 2024

MaartenBent:
Configure builds were excluded, but this would also exclude CMake because it uses the same `__WX_SETUP_H__` define.

This was done back in 248eaddf7a when there were more Windows compilers, of which some probably used configure. But over time he check has been reduced to only `__VISUALC__` which isn't used with configure. So there is no reason to keep this check.

Fixes #24353

### DIFF
--- a/include/wx/wxprec.h
+++ b/include/wx/wxprec.h
@@ -10,14 +10,10 @@
 // compiler detection; includes setup.h
 #include "wx/defs.h"
 
-// check if to use precompiled headers: do it for most Windows compilers unless
-// explicitly disabled by defining NOPCH
-#if defined(__VISUALC__)
-    // If user did not request NOCPH and we're not building using configure
-    // then assume user wants precompiled headers.
-    #if !defined(NOPCH) && !defined(__WX_SETUP_H__)
-        #define WX_PRECOMP
-    #endif
+// check if to use precompiled headers: do it for MSVC unless explicitly
+// disabled by defining NOPCH
+#if !defined(WX_PRECOMP) && defined(__VISUALC__) && !defined(NOPCH)
+    #define WX_PRECOMP
 #endif
 
 #ifdef WX_PRECOMP


vadz:
Thanks for fixing this!

I think it should be safe to backport to 3.2 too, right?

MaartenBent:
I think it is safe. Don't see how this could break abi/api.

PR #24633, opened 2024-06-22
Avoid AppearanceResult::Success conflict with X11 constant

### LINKED ISSUE #24585: glcanvas.h enum/macro conflict with X.h
### Description

When including `glcanvas.h` on Linux, there's a conflict with a macro defined in `X.h`

```cpp
In file included from kicad_dir/common/draw_panel_gal.cpp:41:
In file included from kicad_dir/include/gal/opengl/opengl_gal.h:42:
In file included from kicad_dir/include/gal/hidpi_gl_canvas.h:31:
In file included from wxWidgets_dir/install-vs-qt/include/wx-3.3/wx/glcanvas.h:17:
wxWidgets_dir/install-vs-qt/include/wx-3.3/wx/app.h:690:9: error: expected identifier
        Success,
        ^
/usr/include/X11/X.h:350:21: note: expanded from macro 'Success'
#define Success            0    /* everything's okay */
                           ^
```

- wxWidgets port  you use: wxQT
- OS  and its version: Linux Mint 21.1

vadz:
X11/X.h #defines Success as 0, resulting in compilation errors due to a name clash if this header happens to be included before wx/app.h.

We could #undef Success here, but it seems better to avoid the problem by just renaming our constant, which was introduced only recently and so shouldn't be used in many places yet.

Closes #24585.

----

If sufficiently many people (e.g. more than 1) already use this constant in their code, we should probably keep the existing name and add the `#undef`, but if nobody is using it yet, I think it's better to rename it. Please let me know if you do use it!

### DIFF
--- a/include/wx/app.h
+++ b/include/wx/app.h
@@ -686,7 +686,7 @@ public:
     enum class AppearanceResult
     {
         Failure,
-        Success,
+        Ok,
         CannotChange
     };
 
--- a/interface/wx/app.h
+++ b/interface/wx/app.h
@@ -985,7 +985,7 @@ public:
     enum class AppearanceResult
     {
         Failure,     ///< Changing the appearance failed.
-        Success,     ///< Appearance was successfully changed.
+        Ok,          ///< Appearance was successfully changed.
         CannotChange ///< Appearance can't be changed any more.
     };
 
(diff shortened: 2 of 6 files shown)


vadz:
Applied in the commit above.

PR #24635, opened 2024-06-22
CMake: Set hidden symbol visibility properties

### LINKED ISSUE #24378: Link warnings when using CEF with static wx libraries under macOS
I'm linking my application with static wx libraries using CMake, i.e. I do

```cmake
set(wxBUILD_SHARED OFF)
set(wxUSE_WEBVIEW_CHROMIUM ON)
add_subdirectory(3rdparty/wx EXCLUDE_FROM_ALL)
...
target_link_libraries(MyApp PRIVATE wx::webview wx::core wx::base)
```

and in this case I'm getting the following warnings when linking:

```
ld: warning: direct access in function 'base::cef_internal::BindState<void (*)(CefResourceManager const*), base::cef_internal::UnretainedWrapper<CefResourceManager const> >::BindState<void (*)(CefResourceManager const*), base::cef_internal::UnretainedWrapper<CefResourceManager const> >(std::__1::integral_constant<bool, false>, void (*)(), void (*&&)(CefResourceManager const*), base::cef_internal::UnretainedWrapper<CefResourceManager const>&&)' from file '3rdparty/wx/libs/webview_chromium/cef-build/libcef_dll_wrapper/libcef_dll_wrapper.a(cef_resource_manager.cc.o)' to global weak symbol 'base::cef_internal::BindState<void (*)(CefResourceManager const*), base::cef_internal::UnretainedWrapper<CefResourceManager const> >::Destroy(base::cef_internal::BindStateBase const*)' from file '3rdparty/wx/lib/libwx_osx_cocoau_webview-3.3.a(webview_chromium.cpp.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
ld: warning: direct access in function 'base::cef_internal::BindState<void (*)(CefResourceManager const*), base::cef_internal::UnretainedWrapper<CefResourceManager const> >::BindState<void (*)(CefResourceManager const*), base::cef_internal::UnretainedWrapper<CefResourceManager const> >(std::__1::integral_constant<bool, false>, void (*)(), void (*&&)(CefResourceManager const*), base::cef_internal::UnretainedWrapper<CefResourceManager const>&&)' from file '3rdparty/wx/libs/webview_chromium/cef-build/libcef_dll_wrapper/libcef_dll_wrapper.a(cef_resource_manager.cc.o)' to global weak symbol 'base::cef_internal::BindState<void (*)(CefResourceManager const*), base::cef_internal::UnretainedWrapper<CefResourceManager const> >::Destroy(base::cef_internal::BindStateBase const*)' from file '3rdparty/wx/lib/libwx_osx_cocoau_webview-3.3.a(webview_chromium.cpp.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
ld: warning: direct access in function '(anonymous namespace)::ToLower(CefStringBase<CefStringTraitsUTF16> const&)' from file '3rdparty/wx/libs/webview_chromium/cef-build/libcef_dll_wrapper/libcef_dll_wrapper.a(cef_zip_archive.cc.o)' to global weak symbol 'towlower(int)' from file '3rdparty/wx/lib/libwx_osx_cocoau_core-3.3.a(textcmn.cpp.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
ld: warning: direct access in function '(anonymous namespace)::ToLower(CefStringBase<CefStringTraitsUTF16> const&)' from file '3rdparty/wx/libs/webview_chromium/cef-build/libcef_dll_wrapper/libcef_dll_wrapper.a(cef_zip_archive.cc.o)' to global weak symbol 'towlower(int)' from file '3rdparty/wx/lib/libwx_osx_cocoau_core-3.3.a(textcmn.cpp.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
```

I've confirmed, using `cmake -- -t commands` (I'm using Ninja as generator), that `libcef_dll_wrapper` files are compiled with `-fvisibility=hidden -fvisibility-inlines-hidden`, as expected, so the problem is almost certainly due to not using these options when compiling wx code.

Now I wonder if the warnings above are actually serious enough to justify compiling all code with these options, even when using static libraries, or if we (well, I) should just ignore them. I'm also rather confused by the appearance of `textcmn.cpp.o` in the messages above, it's certainly not the only file using `wxTolower()` in the library, but for some reason the others don't result in any problems (or maybe just don't get reported?).

Any thought/advice would be appreciated!

MaartenBent:
CMake caches the HAVE_VISIBILITY result, so disable it when build options change.

Fixes #24378

### DIFF
--- a/build/cmake/functions.cmake
+++ b/build/cmake/functions.cmake
@@ -111,6 +111,14 @@ function(wx_set_common_target_properties target_name)
         set_target_properties(${target_name} PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
     endif()
 
+    if(NOT WIN32 AND wxBUILD_SHARED AND wxUSE_VISIBILITY)
+        set_target_properties(${target_name} PROPERTIES
+            C_VISIBILITY_PRESET hidden
+            CXX_VISIBILITY_PRESET hidden
+            VISIBILITY_INLINES_HIDDEN TRUE
+        )
+    endif()
+
     if(MSVC)
         if(wxCOMMON_TARGET_PROPS_DEFAULT_WARNINGS)
             set(MSVC_WARNING_LEVEL "/W3")
--- a/build/cmake/functions.cmake
+++ b/build/cmake/functions.cmake
@@ -111,7 +111,7 @@ function(wx_set_common_target_properties target_name)
         set_target_properties(${target_name} PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
     endif()
 
-    if(NOT WIN32 AND wxBUILD_SHARED AND wxUSE_VISIBILITY)
+    if(NOT WIN32 AND wxUSE_VISIBILITY)
         set_target_properties(${target_name} PROPERTIES
             C_VISIBILITY_PRESET hidden
             CXX_VISIBILITY_PRESET hidden
--- a/build/cmake/setup.cmake
+++ b/build/cmake/setup.cmake
@@ -76,11 +76,11 @@ set(wxINSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}")
 
 check_include_files("stdlib.h;stdarg.h;string.h;float.h" STDC_HEADERS)
 
-if(wxBUILD_SHARED)
-    if(wxUSE_VISIBILITY)
-        check_cxx_compiler_flag(-fvisibility=hidden HAVE_VISIBILITY)
-    endif()
-endif() # wxBUILD_SHARED
+if(NOT WIN32 AND wxBUILD_SHARED AND wxUSE_VISIBILITY)
+    check_cxx_compiler_flag(-fvisibility=hidden HAVE_VISIBILITY)
+else()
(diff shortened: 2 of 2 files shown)


MaartenBent:
Actually this doesn't fix #24378, because that one is about static libraries.

Should the visibility flags be added to shared builds too as suggested in https://groups.google.com/g/wx-dev/c/g1Pz3KGVvPQ/m/scgfXZU1BAAJ ?

imciner2:
> Should the visibility flags be added to static builds too 

What does visibility mean to static builds? I thought it was only a setting for shared libraries, since they actually export symbols and a static library is just an archive of all the compiled functions.

MaartenBent:
Probably not much. But it might resolve the warnings about different visibility settings in the linked issue.

vadz:
Thanks, this seems to have indeed fixed my warnings, so merging this.

PR #24637, opened 2024-06-23
Remove OutDir and IntDir conditions in build/msw/*.vcxproj

jkunkee:
This change simplifies the MSW MSVC build files by removing the Condition attribute from the `OutDir` and `IntDir` Properties in `build/msw/*.vcxproj`. Because these are not correctly forward-propagated to new Platforms, this simplifies adding new Platforms.

While this could have been done in wx_setup.props, that file generally only sets `wx*` Properties for later use in Projects.

### DIFF
--- a/build/msw/wx_adv.vcxproj
+++ b/build/msw/wx_adv.vcxproj
@@ -98,26 +98,12 @@
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup>
     <_ProjectFileVersion>10.0.40219.1</_ProjectFileVersion>
-    <OutDir Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">$(wxOutDir)</OutDir>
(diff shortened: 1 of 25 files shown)


jkunkee:
Marking as draft -- while I love to simplify things, the more "VS"-y way is, I fear, to simply duplicate all of these entries.

jkunkee:
Related to #24615

vadz:
I'm fine with applying this as, AFAICS, this doesn't change anything for the build, but the current situation indeed arose because this is the way MSVS IDE generates the projects — which is annoying to me (because I often edit them by hand) but just the way it always worked, AFAIK.

But I don't see any real benefit with remaining perfectly compatible with manually created projects, please let me know if you can think of something. If not, let's merge this (please just remove the "draft" status in this case).

jkunkee:
Sounds good. Draft mark removed.

IIRC the VS IDE will still understand this, but if it gets changed in the GUI it'll revert it back to the per-arch method--so, here we go.

PR #24642, opened 2024-06-25
Make wxWindow::PopupMenu() call wxMenu::UpdateUI()

### LINKED ISSUE #24641: wxWindow::PopupMenu() does not use wxEVT_UPDATE_UI
#### Bug description:

When `wxWindow::PopupMenu(menu)` is called, `menu` is not updated by `wxEVT_UPDATE_UI`. 

I expect that `wxWindow::PopupMenu(menu)` will update `menu` according to the usual `wxEVT_UPDATE_UI` handlers.  This does not happen unless the window is a `wxFrameBase`.

[popupmenu-missing-updateui-demo.patch](https://github.com/user-attachments/files/15963937/popupmenu-missing-updateui-demo.patch)

This patch adds `wxUpdateUIEvent` handlers to the popup menus for both the frame and the dialog that should disable the menu item to display the About box.  However, the dialog's popup menu still has the About menu item enabled.

#### To Reproduce:

1.  Apply the attached patch
2. Run the menu sample
3. Display the popup menu for the frame, and observe that the About menu item is disabled
4. Open the dialog (e.g., use `Ctrl+D`)
5. Display the popup menu for the dialog, and observe that the About menu item is ENabled

- wxWidgets version you use:  master ~2024/06/24
- wxWidgets port you use:  wxMSW
- OS and its version:  MS Windows 10 22H2

wsu-cb:
Fix #24641

### DIFF
--- a/src/common/wincmn.cpp
+++ b/src/common/wincmn.cpp
@@ -3058,6 +3058,7 @@ bool wxWindowBase::PopupMenu(wxMenu *menu, int x, int y)
         setInvokingWin(*menu, static_cast<wxWindow *>(this));
 
     wxCurrentPopupMenu = menu;
+    menu->UpdateUI();
     const bool rc = DoPopupMenu(menu, x, y);
     wxCurrentPopupMenu = nullptr;
 


vadz:

PR #24643, opened 2024-06-25
wxQt: increase default size for wxFileDialog/wxDirDialog.

dsa-t:
The default size was very small by default (400x250), this increases it wxFileDialog/wxDirDialog to Qt's hint size.

### DIFF
--- a/src/qt/filedlg.cpp
+++ b/src/qt/filedlg.cpp
@@ -94,13 +94,18 @@ bool wxFileDialog::Create(wxWindow *parent,
                           const wxString& wildCard,
                           long style,
                           const wxPoint& pos,
-                          const wxSize& sz,
+                          const wxSize& sizeOrig,
                           const wxString& name)
 {
     m_qtWindow = new wxQtFileDialog( parent, this, message, defaultDir,
                                      defaultFile, wildCard, style);
 
-    return wxTopLevelWindow::Create( parent, wxID_ANY, message, pos, sz, style, name );
+    wxSize size(sizeOrig);
+
+    if (!size.IsFullySpecified())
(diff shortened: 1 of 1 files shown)


AliKet:
IMHO, I think we can do this for all `QDialog` descendants by applying something like this
```C++
index 6cbb6a9f3c..c63f989163 100644
--- a/src/qt/window.cpp
+++ b/src/qt/window.cpp
@@ -442,8 +442,9 @@ bool wxWindowQt::Create( wxWindowQt * parent, wxWindowID id, const wxPoint & pos
         p = pos;
 
     wxSize initialSize = size;
-    initialSize.SetDefaults( IsTopLevel() ? wxTopLevelWindowBase::GetDefaultSize()
-                                          : wxQtGetBestSize( GetHandle() ) );
+    initialSize.SetDefaults( IsTopLevel() && !GetHandle()->inherits("QDialog") ?
+                                wxTopLevelWindowBase::GetDefaultSize() :
+                                wxQtGetBestSize( GetHandle() ) );
 
     DoMoveWindow( p.x, p.y, initialSize.GetWidth(), initialSize.GetHeight() );
``` 
What do you think ?

dsa-t:
> What do you think ?

Your idea seems better, and is working as expected with file dialogs.

vadz:
Should this one be closed and the patch above applied it instead or is/will it be part of some other PR?

dsa-t:
> Should this one be closed and the patch above applied

Sure.

vadz:
Thanks, I've applied this patch (will push soon), but FWIW I think this ought to be a virtual function (e.g. `QtGetInitialSize()`) and not a disguised dynamic cast here.

PR #24644, opened 2024-06-25
wxQt: Reduce default best client size for wxListCtrl.

dsa-t:
wxListCtrl default best client width was too big, can be seen in wxLog* dialog, for example.
This reduces the best size to effectively 75x75.

### DIFF
--- a/src/qt/listctrl.cpp
+++ b/src/qt/listctrl.cpp
@@ -1507,11 +1507,6 @@ wxRect wxListCtrl::GetViewRect() const
     return rect;
 }
 
-wxSize wxListCtrl::DoGetBestClientSize() const
-{
-    return wxQtConvertSize(m_qtTreeWidget->minimumSizeHint());
-}
-
 wxTextCtrl* wxListCtrl::GetEditControl() const
 {
     return m_qtTreeWidget->GetEditControl();
--- a/src/qt/listctrl.cpp
+++ b/src/qt/listctrl.cpp
@@ -1507,6 +1507,11 @@ wxRect wxListCtrl::GetViewRect() const
     return rect;
 }
 
+wxSize wxListCtrl::DoGetBestClientSize() const
+{
+    return wxQtConvertSize(m_qtTreeWidget->minimumSizeHint());
+}
+
 wxTextCtrl* wxListCtrl::GetEditControl() const
 {
     return m_qtTreeWidget->GetEditControl();
--- a/include/wx/qt/listctrl.h
+++ b/include/wx/qt/listctrl.h
@@ -274,8 +274,6 @@ public:
 protected:
     void Init();
 
-    virtual wxSize DoGetBestClientSize() const override;
-
     // Implement base class pure virtual methods.
     virtual long DoInsertColumn(long col, const wxListItem& info) override;
     void DoUpdateImages(int which) override;
--- a/include/wx/qt/listctrl.h
+++ b/include/wx/qt/listctrl.h
@@ -274,6 +274,8 @@ public:
 protected:
     void Init();
 
+    virtual wxSize DoGetBestClientSize() const override;
+
     // Implement base class pure virtual methods.
     virtual long DoInsertColumn(long col, const wxListItem& info) override;
     void DoUpdateImages(int which) override;
--- a/src/qt/dc.cpp
+++ b/src/qt/dc.cpp
@@ -400,7 +400,6 @@ wxCoord wxQtDCImpl::GetCharHeight() const
 
 wxCoord wxQtDCImpl::GetCharWidth() const
 {
-    //FIXME: Returning max width, instead of average
     QFontMetrics metrics(m_qtPainter->isActive() ?
         m_qtPainter->font() : QApplication::font());
     return wxCoord( metrics.averageCharWidth() );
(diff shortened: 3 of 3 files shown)


src/qt/listctrl.cpp
```diff
@@ -1507,6 +1507,11 @@ wxRect wxListCtrl::GetViewRect() const
     return rect;
 }
 
+wxSize wxListCtrl::DoGetBestClientSize() const
+{
+    return wxQtConvertSize(m_qtTreeWidget->minimumSizeHint());
+}
+
```
AliKet, line 1514: Thanks for fixing this, but I think it would be better to call `viewportSizeHint()` here instead of using the minimum size by default. Just add this `using QTreeView::viewportSizeHint;` too to the public interface of `wxQtListTreeWidget` because it is protected in `QTreeView` and we need to call it from `wxListCtrl`.

TIA.

dsa-t: Thanks for the suggestion.

If there's a lot of items, `viewportSizeHint` returns 254x868 here, for example:
image

And because best size == effective min size here, it can't be shrinked.

For reference, on wxGTK it looks like this:

image

dsa-t: On wxQt with `minimumSizeHint` it looks like this:

image

AliKet: I thought we should follow the other ports (I can only test under Windows & Linux) when calculating the best client size. i.e. Both `wxMSW` and `wxGTK` do this:
- the calculated best width is slightly larger than the widest string.
- the calculated best height is limited to 10 items.

AFAICT `minimumSizeHint()` always returns a fixed size, but the problem with `viewportSizeHint()` can be solved by limiting the returned height to 10 items as is done by `wxMSW` and `wxGTK`.

Please let me know if I'm still missing something here? TIA.

dsa-t: When using `averageCharWidth`, in both wxGTK and wxQt, `DoGetBestClientSize` returns 400x160 size

dsa-t: wxGTK:
image

wxQt:
image

AliKet: I don't recall why we had to use `maxWith` instead of `averageCharWidth` TBH, and I tried to run the tests locally with `maxWidth` replaced by `averageCharWidth` and everything runs successfully. So unless anyone remembers why `maxWith` is used instead of `averageCharWidth`, let's change `wxQtDCImpl::GetCharWidth` to use the latter then and see how it goes. TIA.

AliKet:
I couldn't reproduce the failure locally! and here is the test run on my machine:

```
xvfb-run -a -s '-screen 0 1600x1200x24' ./test_gui [ellipsization]
build: 3.3.0 (UTF-8,compiler with C++ ABI compatible with gcc 4,STL containers,compatible with 3.2)
compiled using gcc 8.4
running under Linux 4.15.0-214-generic i686 as aliket
Filters: [ellipsization]
All tests passed (697 assertions in 4 test cases)

```

dsa-t:
Tests pass on my system also, even if I set the font to "Sans Serif":

`dc.SetFont(wxFont(9, wxFONTFAMILY_SWISS, wxFONTSTYLE_NORMAL, wxFONTWEIGHT_NORMAL, false, "Sans Serif"));`

which gives the same description:

`Sans Serif,9,-1,5,50,0,0,0,0,0`

AliKet:
Since the failure is not reproduced locally, I don't see a better way to fix the GitHub CI failure other than adding a one-pixel (1px) tolerance for `wxQt` to pass the tests there, unfortunately. something like this:
```C++
diff --git a/tests/graphics/ellipsization.cpp b/tests/graphics/ellipsization.cpp
index 027cbec6cc..ad72486b06 100644
--- a/tests/graphics/ellipsization.cpp
+++ b/tests/graphics/ellipsization.cpp
@@ -88,6 +88,12 @@ TEST_CASE("Ellipsization::NormalCase", "[ellipsization]")
                     const int
                         width = dc.GetMultiLineTextExtent(displayed).GetWidth();
 
+                    #ifdef __WXQT__
+                    const int TOLERANCE = 1; // 1px
+                    #else
+                    const int TOLERANCE = 0;
+                    #endif
+
                     WX_ASSERT_MESSAGE
                     (
                      (
@@ -99,7 +105,7 @@ TEST_CASE("Ellipsization::NormalCase", "[ellipsization]")
                         width,
                         widthsToTest[w]
                      ),
-                     width <= widthsToTest[w]
+                     width <= widthsToTest[w] + TOLERANCE
                     );
                 }
             }
```

vadz:
It's rather surprising that the failure is not reproducible locally, could it be due to a Qt version difference?

Anyhow, we could add a hack for Qt only if you think it's not worth digging into this further, should I merge it with the patch above?

AliKet:
This is fixed now in [6937f73](https://github.com/wxWidgets/wxWidgets/pull/24837/commits/6937f730e988d8a13955f66dd2122358399ba068)

vadz:
Thanks, so, just to confirm: this one should be closed then?

AliKet:
> Thanks, so, just to confirm: this one should be closed then?

Yes.

PR #24647, opened 2024-06-26
Include macOS 15 name in wxGetOsDescription()

TcT2k:
(Should also be included in 3.2)


vadz:
Thanks, will merge soon.

PR #24653, opened 2024-06-28
Update the Windows versions table in wxGetOsVersion() docs

PBfordev:
I am not sure if the table is really useful.

If accepted, should be backported to 3.2.


vadz:
Sure, this looks useful, thanks!

PR #24655, opened 2024-06-29
wxGrid: implement 'pressed' for highlighted row/column headers while drag-moving a row or column

DietmarSchwertberger:
@AliKet : thanks for implementing PR https://github.com/wxWidgets/wxWidgets/pull/24561

Sorry for being late. I just built wx these days week as I'm working on making row/col drag-move compatible with row/col selection. (At the moment, if selection is possible, you can't drag a row.)

With the new highlighted row and column labels, the visual display of a row being drag-moved does not match the rest any more.

You can check by running the grid sample, `Grid -> Row drag-move` and `Select -> Change selection mode -> Disallow selection` and then drag a row with the mouse.

The current implementation (row 6 is being dragged and the insertion marker is between rows 7 and 8):
grafik

I added a parameter `pressed`to `DrawHighlighted` and the button edge colors are swapped if `pressed` is `true`:
grafik
I'm not 100% satisfied, but at least it looks better than before. What do you think?

Another minor issue with the new highlighting:
Personally, I don't like that if e.g. a row is selected, all column headers will be highlighted.
I'm often using the row selection to select a dataset in an application. All other forms will then refer to the selected row.

### DIFF
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -614,12 +614,18 @@ public:
                            int vertAlign,
                            int textOrientation) const;
 
+    // argument 'flags' for DrawHighlighted
+    enum DrawFlags
+    {
+        Draw_Pressed = 1
+    };
+
     // Draw highlighted row/col label.
     virtual void DrawHighlighted(const wxGrid& grid,
                                  wxDC& dc,
                                  wxRect& rect,
                                  int rowOrCol,
-                                 bool pressed = false) const;
+                                 int flags = 0) const;
 };
 
 // Currently the row/column/corner renders don't need any methods other than
@@ -655,7 +661,7 @@ public:
                                  wxDC& dc,
                                  wxRect& rect,
                                  int row,
-                                 bool pressed = false) const override;
+                                 int flags = 0) const override;
 };
 
 // Column header cells renderers
@@ -671,7 +677,7 @@ public:
                                  wxDC& dc,
                                  wxRect& rect,
                                  int col,
-                                 bool pressed = false) const override;
+                                 int flags = 0) const override;
 };
 
 // Header corner renderer
--- a/include/wx/generic/grid.h
+++ b/include/wx/generic/grid.h
@@ -618,7 +618,8 @@ public:
     virtual void DrawHighlighted(const wxGrid& grid,
                                  wxDC& dc,
                                  wxRect& rect,
-                                 int rowOrCol) const;
+                                 int rowOrCol,
+                                 bool pressed = false) const;
 };
 
 // Currently the row/column/corner renders don't need any methods other than
@@ -653,7 +654,8 @@ public:
     virtual void DrawHighlighted(const wxGrid& grid,
                                  wxDC& dc,
                                  wxRect& rect,
-                                 int row) const override;
+                                 int row,
+                                 bool pressed = false) const override;
 };
 
 // Column header cells renderers
@@ -668,7 +670,8 @@ public:
     virtual void DrawHighlighted(const wxGrid& grid,
                                  wxDC& dc,
                                  wxRect& rect,
-                                 int col) const override;
+                                 int col,
+                                 bool pressed = false) const override;
 };
 
 // Header corner renderer
--- a/samples/grid/griddemo.cpp
+++ b/samples/grid/griddemo.cpp
@@ -77,7 +77,8 @@ public:
     virtual void DrawHighlighted(const wxGrid& WXUNUSED(grid),
                                  wxDC& dc,
                                  wxRect& rect,
-                                 int WXUNUSED(rowOrCol)) const override
+                                 int WXUNUSED(rowOrCol),
+                                 bool WXUNUSED(pressed)) const override
     {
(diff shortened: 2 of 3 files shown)


AliKet:
> @AliKet : thanks for implementing PR #24561
> 

My pleasure to help in any way I can.

> Sorry for being late. I just built wx these days week as I'm working on making row/col drag-move compatible with row/col selection. (At the moment, if selection is possible, you can't drag a row.)
> 

AFAICT, that was the case even before, and I'm wondering if we can bind a key to enable row/col dragging even if selection is possible, right?

> With the new highlighted row and column labels, the visual display of a row being drag-moved does not match the rest any more.
> 
> You can check by running the grid sample, `Grid -> Row drag-move` and `Select -> Change selection mode -> Disallow selection` and then drag a row with the mouse.
> 
> The current implementation (row 6 is being dragged and the insertion marker is between rows 7 and 8): grafik
> 
> I added a parameter `pressed`to `DrawHighlighted` and the button edge colors are swapped if `pressed` is `true`: grafik I'm not 100% satisfied, but at least it looks better than before. What do you think?
> 

Looks good to me too, thanks.

> Another minor issue with the new highlighting: Personally, I don't like that if e.g. a row is selected, all column headers will be highlighted. I'm often using the row selection to select a dataset in an application. All other forms will then refer to the selected row.

Isn't this how other spreadsheet applications behave when the entire row or column is selected?

DietmarSchwertberger:
> AFAICT, that was the case even before, and I'm wondering if we can bind a key to enable row/col dragging even if selection is possible, right?

I will submit a separate PR to start drag-moving when the user drags over a minimum distance (`> DRAG_SENSITIVITY`).

> Looks good to me too, thanks.

Thanks, so I will change from draft to ready.

> Isn't this how other spreadsheet applications behave when the entire row or column is selected?

Yes, but the grid is a general purpose control, not a spreadsheet and the darker background is good for navigation, but not for legibility.
I would find it more useful not to highlight all columns but just the one with the active cell. Maybe that could be used to differentiate: higlight all columns if there is no active cell; highlight only one column if there is an active cell.
(Well I can live with the current state, as in my application, after selecting the row, I can clear the selection again and just change the row label from "NN" to "NN >", as I always did to indicate the current row.)
I just saw that when selecting a row, the active cell is set to the first column of the row, even if the active cell was already in this row. I would consider this behavior a bug. It has been like this all the time, but I never noticed before.

DietmarSchwertberger:
@AliKet : I think the implementations of `IsRow/ColLabelHighlighted` should be changed. They don't work for non-contiguous selections, only for `blocks.size() == 1`.
E.g. run the grid sample, select row 2, then select row 4 while holding the Ctrl button. The selection now consists of two blocks, but none of the row and column labels are highlighted.

AliKet:
> @AliKet : I think the implementations of `IsRow/ColLabelHighlighted` should be changed. They don't work for non-contiguous selections, only for `blocks.size() == 1`. E.g. run the grid sample, select row 2, then select row 4 while holding the Ctrl button. The selection now consists of two blocks, but none of the row and column labels are highlighted.

Yes, this is intentional because:
- For efficiency reasons, these functions are supposed to run in constant time.
- Their implementation would be unnecessarily complicated.

Unfortunately, I don't have a simple, satisfactory solution to these problems, so I've simply decided to imitate the behavior of LibreOffice Calc and others here.

DietmarSchwertberger:
I don't think that the performance is an issue here.
If it was, then other code would have much bigger problems.
E.g. `wxGrid::DrawGridCellArea` calls `wxGrid::DrawCell` which uses `IsInSelection(coords));` to decide whether the cell is to be rendered as selected or not.

include/wx/generic/grid.h
```diff
@@ -618,7 +618,8 @@ class WXDLLIMPEXP_CORE wxGridHeaderLabelsRenderer
     virtual void DrawHighlighted(const wxGrid& grid,
                                  wxDC& dc,
                                  wxRect& rect,
-                                 int rowOrCol) const;
+                                 int rowOrCol,
+                                 bool pressed = false) const;
```
vadz, line 622: This is [my pet peeve](https://www.wxwidgets.org/develop/coding-guidelines/#no_bool_params) but could we please do something like

```suggestion
                                 int flags = 0) const;
```

and define

```cpp
enum DrawFlags
{
  Draw_Pressed = 1
}
```

here and below?

vadz (commented):
Thanks, I agree that the appearance is better with this change, but could you please make the proposed change to avoid having unclear bool params? TIA!

AliKet:
> @AliKet : I think 'DrawHighlighted' needs to be added to `interface/wx/grid.h`. Would you mind submitting a PR once this one is merged? 

Yes I think it should be documented, and sorry for not doing so when it was first added.

>Did you look at a more complete implementation of `IsRow/ColLabelHighlighted`?

Not at the moment, sorry, as I don't have the energy to work on this right now, plus some other work I need to finish.

Regards.

vadz:
Thanks for the update, will merge in a moment.

PR #24658, opened 2024-06-30
grid: enable row/col selection and drag-move at the same time

DietmarSchwertberger:
At the moment, enabling drag-move of rows/cols inhibits row/col selection.
This is due to the `LeftDown` down handler inside `ProcessRowColLabelMouseEvent` using `if { } else { if`.
This PR removes the `else`.

The diff looks rather wild, but actually the only change is the removal of `else` and that the selection cursor is not used if it has been set to MOVE already:
```
if ( !oper.CanDragMove(this) )
    ChangeCursorMode(oper.GetCursorModeSelect(), labelWin);
```

I have experimented with changing the cursor mode to `WXGRID_CURSOR_MOVE_ROW/COL` only after the mouse has moved by some pixels, but I prefered the immediate change of the label to 'pressed'. A delayed setting of the cursor mode is too complicated with the current grid implementation (a proper state and state machine data structure would be needed).

### DIFF
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -4269,59 +4269,57 @@ void wxGrid::ProcessRowColLabelMouseEvent( const wxGridOperations &oper, wxMouse
                 if ( oper.CanDragMove(this) )
                 {
                     ChangeCursorMode(oper.GetCursorModeMove(), headerWin);
-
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, makes sense to me, will merge in a moment.

Issue #24611, opened 2024-06-14
[wxGTK] [Windows] Missing argvA/char version of argv when compiling on Windows


ryancog:
### Build System Used

I build wxWidgets and/or my application using:

- [ ] CMake
- [X] configure
- [ ] mingw32-make with makefile.gcc
- [ ] MSBuild (Microsoft Visual Studio solution file)
- [ ] nmake with makefile.vc
- [ ] Xcodeu

And using MXE to setup/install the cross-compilation environment
```
../configure --disable-shared --prefix=/opt/mxe/usr/x86_64-w64-mingw32.static --host=x86_64-w64-mingw32.static --build=x86_64-linux --with-gtk=3
make -j16
sudo make install
```
I don't believe the config.log is relevant, as from the proprocessor defines, there's nothing unintended occurring, but what is intended fails in this (niche) circumstance.

### Description

Whenever building wxGTK for Windows, there's a compile error in src/gtk/app.cpp first about wxInitData::Get().argvA not existing (specifically the variable argvA) and (after I changed it to argv before understanding the point of the "A") then an error about conversion about converting from wchar_t to char (specifically pointers, but that's besides the point).

At which point I looked into it a bit more and realized that argvA is supposed to be a char** representation of argv, where argv is normally a (as far as I can tell, as least for this build) a wchar_t**. However, specifically when building for Windows, there's a preprocessor if in wxInitData that removes argvA, and I assume another one whereever the implementation details are of filling argvA.

This causes issues as the gtk init requires a char** representation of argv. As a quick workaround just to build my application, I've done the following in app.cpp:

```cpp
#if defined(__WXGTK4__)
    init_result = gtk_init_check() != 0;
#else
#if defined(__WINDOWS__)
    auto ORIG_argc_{argc_};
    auto argvA = new char*[ORIG_argc_];
    for (auto i{0}; i < ORIG_argc_; i++) {
        auto len{wcslen(argv_[i])};
        argvA[i] = new char[len + 1];
        argvA[i][len] = '\0';
        wcstombs(argvA[i], argv_[i], len);
    }
#else // !defined(__WINDOWS__)
    auto argvA = wxInitData::Get().argvA;
#endif // defined(__WINDOWS__)

    int argcGTK = argc_;
    init_result = gtk_init_check( &argcGTK, &argvA ) != 0;

    if ( argcGTK != argc_ )
    {
        // we have to drop the parameters which were consumed by GTK+
[...]

PR #24659, opened 2024-06-30
Fix argvA initialization for wxGTK under Windows

ryancog:
Update conditional directives to enable argvA if `!__WXMSW__` instead of if `!__WINDOWS__` (It seems originally there was confusion/misunderstanding of the roles of the two macros as toolkit and platform indicators respectively)

Fix argvA not being initialized in InitIfNecessary due to argc check returning early.

See issue #24611

### DIFF
--- a/src/common/init.cpp
+++ b/src/common/init.cpp
@@ -183,6 +183,26 @@ void wxInitData::Initialize(int argcIn, char **argvIn)
     argv[wargc] = nullptr;
 }
 
+#ifndef __WXMSW__
+
+// This function builds the argvA
+void wxInitData::BuildArgvA() {
+    // We need to convert from wide arguments back to the narrow ones.
+    argvA = new char*[argc + 1];
+    argvA[argc] = nullptr;
+
+    ownsArgvA = true;
+
+    for ( int i = 0; i < argc; i++ )
+    {
+        // Try to use the current encoding, but if it fails, it's better to
+        // fall back to UTF-8 than lose an argument entirely.
+        argvA[i] = wxConvWhateverWorks.cWC2MB(argv[i]).release();
+    }
+}
+
+#endif // !__WXMSW__
+
 #ifdef __WINDOWS__
 
 void wxInitData::MSWInitialize()
@@ -202,18 +222,7 @@ void wxInitData::MSWInitialize()
     argv = argvMSW;
 
 #ifndef __WXMSW__
-    // We need to convert from wide arguments back to the narrow ones.
-    argvA = new char*[argc + 1];
-    argvA[argc] = nullptr;
-
-    ownsArgvA = true;
-
-    for ( int i = 0; i < argc; i++ )
-    {
-        // Try to use the current encoding, but if it fails, it's better to
-        // fall back to UTF-8 than lose an argument entirely.
-        argvA[i] = wxConvWhateverWorks.cWC2MB(argv[i]).release();
-    }
+    BuildArgvA();
 #endif // !__WXMSW__
 }
 
@@ -245,18 +254,7 @@ void wxInitData::InitIfNecessary(int argcIn, wchar_t** argvIn)
     }
 
 #ifndef __WXMSW__
-    // We need to convert from wide arguments back to the narrow ones.
-    argvA = new char*[argc + 1];
-    argvA[argc] = nullptr;
-
-    ownsArgvA = true;
-
-    for ( int i = 0; i < argc; i++ )
-    {
-        // Try to use the current encoding, but if it fails, it's better to
-        // fall back to UTF-8 than lose an argument entirely.
-        argvA[i] = wxConvWhateverWorks.cWC2MB(argvIn[i]).release();
-    }
+    BuildArgvA();
 #endif // !__WXMSW__
 
 #ifdef __WINDOWS__
--- a/src/common/init.cpp
+++ b/src/common/init.cpp
@@ -200,41 +200,33 @@ void wxInitData::MSWInitialize()
     // argvMSW because it could be allocated by Initialize() if a custom entry
     // point is used.
     argv = argvMSW;
+
+#ifndef __WXMSW__
+    // We need to convert from wide arguments back to the narrow ones.
+    argvA = new char*[argc + 1];
+    argvA[argc] = nullptr;
+
+    ownsArgvA = true;
+
+    for ( int i = 0; i < argc; i++ )
+    {
+        // Try to use the current encoding, but if it fails, it's better to
+        // fall back to UTF-8 than lose an argument entirely.
+        argvA[i] = wxConvWhateverWorks.cWC2MB(argv[i]).release();
+    }
+#endif // !__WXMSW__
 }
 
 #endif // __WINDOWS__
 
 void wxInitData::InitIfNecessary(int argcIn, wchar_t** argvIn)
 {
-#ifndef __WXMSW__
-    // It's possible that the below checks for argc and argcIn will pass
-    // and thus return, despite argvA not being initialized. For this reason,
-    // the argvA routine must be checked and (if necessary) ran first.
-    //
-    // It's unclear to me if there's ever a condition where argvA is init'd
-    // and argc/argva are not, so instead of returning here the entire
-    // routine is wrapped in the if.
-    if ( !argvA ) {
-        // We need to convert from wide arguments back to the narrow ones.
-        argvA = new char*[argc + 1];
-        argvA[argc] = nullptr;
-
-        ownsArgvA = true;
-
-        for ( int i = 0; i < argc; i++ )
-        {
-            // Try to use the current encoding, but if it fails, it's better to
-            // fall back to UTF-8 than lose an argument entirely.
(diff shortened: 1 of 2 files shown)


vadz (commented):
Sorry, I feel uncomfortable about this because I don't understand what's going on here any longer: how can it happen that `argc != 0` but `argvA == 0` (when it's used)? I think this shouldn't happen, i.e. we should respect the invariant `argv == 0 || argvA != 0`.

Am I missing something?

ryancog:
To be honest this doesn't make sense to me either. However in testing when I had things the original way (with the original check and the argvA fill after that), the program would immediately crash with a nullptr dereference from trying to use argvA.

I did some printf debugging and found out that the Initialize function wasn't ever run (at least within the time from start to crash) and that InitIfNecessary was run, but the the check caused it to return before actually doing anything.

I did some `grep -r`ing to find anywhere and everywhere argvA was used, so I don't think I missed enabling some other routine that should've filled it...

> we should respect the invariant `argv == 0 || argvA != 0`

I guess I don't understand... are you suggesting this check should be used instead?

To me, InitIfNecessary's job seems to be to initialize these variables if they haven't been already, so checking if argvA is nullptr (and thus not initialized) and then initializing it if so seems to be expected behavior.

In the case of argvA, in an initialized state, it shouldn't ever be nullptr right? Because at the very least argvA[0] should be executable? At this point perhaps I don't understand how (if at all) wxWidgets should have processed the arguments, and maybe that's where I'm confused and/or wrong in my logic.

vadz:
> To be honest this doesn't make sense to me either. However in testing when I had things the original way (with the original check and the argvA fill after that), the program would immediately crash with a nullptr dereference from trying to use argvA.

When this happened, was `argv` already initialized? If it was, we need to find the place which initialized it and initialize `argvA` there too.

> I did some printf debugging and found out that the Initialize function wasn't ever run (at least within the time from start to crash) and that InitIfNecessary was run, but the the check caused it to return before actually doing anything.

But if it didn't do anything, it must mean that `argc != 0`, so it must have been set somewhere?

> > we should respect the invariant `argv == 0 || argvA != 0`
> 
> I guess I don't understand... are you suggesting this check should be used instead?

 No, sorry for the confusion, this is not a check, but the invariant, i.e., in words, we should always have valid `argvA` (when it's used at all, i.e. not when using wxMSW) if we have valid `argv`.

> In the case of argvA, in an initialized state, it shouldn't ever be nullptr right? Because at the very least argvA[0] should be executable? At this point perhaps I don't understand how (if at all) wxWidgets should have processed the arguments, and maybe that's where I'm confused and/or wrong in my logic.

 It's possible that someone calls `wxInitialize(0,0)` (well, with references instead of values, but this is not the point), so we need to handle this case by allocating 1-element `argv` and `argvA` arrays containing a single null entry (not even the executable name, because this would correspond to `argc == 1`). However if the real entry point (i.e. `main()`) is called, we should have `argc >= 1`.

 I guess the problem is that we use `argc != 0` as the test for checking whether we were initialized, but we should use `argv != nullptr` instead.

ryancog:
> > To be honest this doesn't make sense to me either. However in testing when I had things the original way (with the original check and the argvA fill after that), the program would immediately crash with a nullptr dereference from trying to use argvA.
> 
> When this happened, was `argv` already initialized? If it was, we need to find the place which initialized it and initialize `argvA` there too.

I didn't think to check. Give me a moment and I can see.

> > I did some printf debugging and found out that the Initialize function wasn't ever run (at least within the time from start to crash) and that InitIfNecessary was run, but the the check caused it to return before actually doing anything.
> 
> But if it didn't do anything, it must mean that `argc != 0`, so it must have been set somewhere?

Right, that makes sense. I can try and look...

> this is not a check, but the invariant, i.e., in words, we should always have valid `argvA` (when it's used at all, i.e. not when using wxMSW) if we have valid `argv`.

Ah, ok I see. It seems then I've learned a new word/concept today, thanks! :)

> It's possible that someone calls `wxInitialize(0,0)` (well, with references instead of values, but this is not the point), so we need to handle this case by allocating 1-element `argv` and `argvA` arrays containing a single null entry (not even the executable name, because this would correspond to `argc == 1`). However if the real entry point (i.e. `main()`) is called, we should have `argc >= 1`.

Shouldn't that still be argc == 0 since the array is simply null-terminated, but there's no actual data in it? Or is that what you mean, that adding the executable name would then be argc == 1, but 1-length array w/ a nullptr element would be argc == 0?

> I guess the problem is that we use `argc != 0` as the test for checking whether we were initialized, but we should use `argv != nullptr` instead.

Would then we want `if ( argv || !argvIn )` to be the check? I actually had this thought myself when I was looking at the code that it was slightly odd that argc was checked when in both principal and in practice, it was argv that mattered.

ryancog:
Update: 

In the gtk app init, where argvA is used, and without the rearranging for InitIfNecessary, at that time argvA is nullptr, and argv is valid.

So argv is being initialized somewhere argvA isn't.

ryancog:
Interestingly it seems as though wxInitData::Initialize is never run...

ryancog:
in `src/msw/main.cpp` I've traced the wxEntry that is used on windows (compatible with WinMain), and it calls `wxInitData::Get()` to pass `argc` and `argv` to `wxEntry`. At this point the WinMain wxEntry has called `wxMSWEntryCommon()` which calls `wxInitData().Get().MSWInitialize()`.

`MSWInitialize()` sets argvMSW and... *ding ding ding* argv.

So, wxInitData::Initialize() indeed never is called...
So that means that this block here:
```c
#ifndef __WXMSW__
    argvA = argvIn;
#endif
```
Was an oversight. It doesn't do anything because it's never called. Even if it was, the assert above (rightly) would cause issues because it would mean argv is attempting to be reinitialized.

This makes the structure a bit convoluted. It seems `wxInitData::Initialize()` (and that little `argvA = argvIn`) *is* used under other platforms, so that can't be removed, but the preprocessor conditional also cannot be removed, because of course with wxMSW argvA doesn't exist.

It would seem then that the best thing to do is generate argvA in MSWInitialize with a conditional `#ifndef __WXMSW__`, using the same routine as is in `InitIfNecessary`, but also keep the block that is in `InitIfNecessary` so it fulfills its job should there be an edge case that requires it (and then it will initialize argv and argvA, but argvMSW will be nulled. Whether that's correct I'm unsure).

src/common/init.cpp
```diff
@@ -200,17 +200,33 @@ void wxInitData::MSWInitialize()
...
+
+    ownsArgvA = true;
+
+    for ( int i = 0; i < argc; i++ )
+    {
+        // Try to use the current encoding, but if it fails, it's better to
+        // fall back to UTF-8 than lose an argument entirely.
+        argvA[i] = wxConvWhateverWorks.cWC2MB(argv[i]).release();
+    }
+#endif // !__WXMSW__
```
vadz, line 217: Now the logic indeed makes sense to me, thanks!

But this code seems to exactly duplicate the code below, so what about extracting it into `InitArgsA()` and calling it from both places instead of copying it?

ryancog: > But this code seems to exactly duplicate the code below, so what about extracting it into `InitArgsA()` and calling it from both places instead of copying it?

Makes sense to me, will do!

ryancog: Would you prefer it be a helper function (taking in the values of argc and argv (well, the pointer at least) and a reference to argvA and building it) or a member function of wxInitData?

I'm not sure what the philosophy is here. Since the include is already in a `private` subdirectory, it seems fine to add helpers as members there?

vadz: Yes, it's definitely fine to add a member function to do it, TIA!

ryancog:
Ok, I've created `wxInitData::BuildArgvA()`, guarded it with `#ifndef __WXMSW__` and used it in `InitIfNecessary()` and `MSWInitialize()`

ryancog:
Hmm... it seems one of the CI tests failed... but one that failed claims "success" at the end (so clearly I don't know how to read the log.) and the other one fails to a test related to a completely unrelated GUI element? I'm not sure how I would've managed to break that...

vadz:
Must be an AppVeyor fluke, this is clearly unrelated. I've relaunched this build just to check.

ryancog:
Gotcha. Assuming that that completes without issue, I think that's everything, been playing with the builds a bit as I work on my application and I haven't run into any issues. *knock on wood*

ryancog:
Well the rest of the tests just died on their own because they timed out…

vadz:
I'm finally merging this with just some minor renaming, thanks!

ryancog:
Awesome!

Fwiw I named it `BuildArgvA` instead of `InitArgvA` because there's technically the case where argvA is filled by a char** argv. I don't recall exactly, but I know one of the wxEntry's (or one of the init functions?) did that.

So build made more sense to me since it differentiated those two things... it's still init'ing argvA, but via the process of building it from a wchar.

Not something that really matters that much I figure, but I thought I'd explain why I named it that originally :)

vadz:
Thanks, I agree that this explanation makes sense, but naming all of the functions `InitSomething` seemed appealing from consistency point of view as well...

ryancog:
Understandable :)
I think either way it's plenty logical.

PR #24660, opened 2024-06-30
Fix crash on static page of widgets sample

### LINKED ISSUE #24656: widgets sample can crash when modifying a widget style
In the widgets sample, when using the UI to modify the style of a widget, the current widget(s) is/are destroyed and then recreated.

On pages where there is more than one widget being shown (e.g., on the static text page), this can lead to the sample crashing or freezing. The reason is still the same as in #24487 (i.e., https://github.com/wxWidgets/wxWidgets/commit/a8b4753d1dfa84a41b0aba6ca4c93811abcfa3e2):
1. The widget page keeps a collection of pointers to its widgets.
2. When modifying the widgets, the existing ones are destroyed, making those pointers in the collection invalid.
3.  After recreating each widget, `WidgetsPage::NotifyWidgetRecreation()` update function is called for the widget; however, this ends in `WidgetsPage::SetUpWidget()` where the whole page widgets collection is iterated and the stale pointers to the destroyed widgets accessed.

I am not sure how to best fix it at the sample level, instead of fixing each page, so for now I am just posting an Issue, not a PR. Perhaps using `wxWeakRef` instead of raw pointers in the widget collection would be a simple fix? But the current way, i.e., setting-up all the widgets each time one of them is (re)created, seems not ideal so a whole different approach may be better...

PBfordev:
Attempting to change the style or alignment of the static text in the widgets sample made the sample crash. The bug cause was the same as in 7c125d3 (Fix crash on spin page of widgets sample, 2024-04-18), i.e., a8b4753 (Reconnect events to the widget after recreating it in the sample, 2024-02-19).

Fix the issue by calling NotifyWidgetRecreation() for each control only after they were all recreated.

Closes #24656.

### DIFF
--- a/samples/widgets/static.cpp
+++ b/samples/widgets/static.cpp
@@ -525,13 +525,10 @@ void StaticWidgetsPage::CreateStatic()
 #endif // wxUSE_MARKUP
     }
 
-    NotifyWidgetRecreation(m_statText);
 
     m_statText->SetToolTip("Tooltip for a label inside the box");
 
 #if wxUSE_MARKUP
-    NotifyWidgetRecreation(m_statMarkup);
-
     m_statMarkup->SetLabelMarkup(m_textLabelWithMarkup->GetValue());
 
     if ( m_chkGreen->GetValue() )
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for fixing this, I didn't have time to think about better solution (or even why is this one not the best), but I'll merge this already to fix a crash for now.

PR #24663, opened 2024-07-01
Fix static box bgcolor when on wxPanel (MSW dark mode)

Blake-Madden:
Under MSW dark mode, `wxStaticBox` will have a dark gray background when placed on a `wxPanel`. It seems that this is the color it is pulling from the parent `wxPanel`. Explicitly setting the panel's background color to the proper color under dark mode fixes it.

At first I thought of changing `wxStaticBox` to not use the parent's bgcolor, but perhaps this is a better approach in case their are other controls that also look at the parent's bgcolor. Fixing at the `wxPanel` level will fix all any of those.

BEFORE:

Before

AFTER:

After

### DIFF
--- a/src/common/panelcmn.cpp
+++ b/src/common/panelcmn.cpp
@@ -30,7 +30,11 @@
 #endif
 
 #ifdef __WXMSW__
-    #include "wx/msw/private/darkmode.h"
+    #if wxUSE_UXTHEME
+        #include "wx/msw/uxtheme.h"
+        #include "wx/msw/private.h"
+        #include "wx/msw/private/darkmode.h"
+    #endif
 #endif
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


vadz:
Sorry for being so late to answer here.

I'm aware of the problem and had already tried to fix it myself, but I'm still not sure what the best fix is. I don't think that setting the colour explicitly as this PR does is it, however. My preferred solution would be to just use `wxSYS_COLOUR_WINDOW` as the default window background (which would make sense!) but I'm not sure if not using `wxSYS_COLOUR_BTNFACE` for it by default is not going to break something, so, after agonising over whether it was or was not safe to do it, I've finally created the ugly #24753 which avoids any chance of breaking something by preserving the same behaviour as before in the light mode and only overriding the background in dark mode in wxMSW.

Please let me know if you still see any problems with it and I'm closing this one as I think using the correct colours by default is better than changing it later.

PR #24665, opened 2024-07-02
Improve battery and wxSecureZeroMemory docs

Blake-Madden:
Expand docs about how `wxSecureZeroMemory` may be implemented.
Reword the battery functions descriptions about which platform it words on (it was a bit confusing with how it was worded, IMO). The explanation matches similar wording I see elsewhere in the help.


vadz:
Thanks, merging too (with the other doc PR).

Issue #24669, opened 2024-07-03
wxMAC wxListCtrl visible item range not stable while adding items


alcontre:
#### Bug description:
When using `SetItemCount()` to add items to a virtual wxListCtrl that already contains many items (millions), the calculated range of visible items is slightly different on each repaint. This causes the range of visible items to "dance" around on screen.

I believe this is because, in the wxMAC implementation, the visible range is calculated by using the native control position value in range [0,1] times the maximum number of items in the list. This value is computed and set as a double:

```
    void SetScrollThumb( wxInt32 value, wxInt32 thumbSize ) override
    {   
        double v = ((double) value)/m_maximum;
        double t = ((double) thumbSize)/(m_maximum+thumbSize);
        [(wxNSScroller*) m_osxView setDoubleValue:v];
        [(wxNSScroller*) m_osxView setKnobProportion:t];
    }   
```
I believe the issue is that `GetValue()` retrieves it as a float and rounds, leading to issues with numerical precision.
```
    virtual wxInt32 GetValue() const override
    {   
        return wxRound([(wxNSScroller*) m_osxView floatValue] * m_maximum);
    } 
```

Using `doubleValue` appears to fix the behavior. I briefly tried updating wxOSXScrollBarCocoaImpl to track its integer position as a member variable directly, which seems like a more robust solution, but I wasn't quite sure how to update the integer position a result of click-and-drag on the scroll thumb.

I opened PR https://github.com/wxWidgets/wxWidgets/pull/24670 if this is deemed an acceptable solution.

#### Expected vs observed behaviour:

I would expect the visible range of items to not change, as additional items are added through `SetItemCount()`.

#### Patch or snippet allowing to reproduce the problem:
```
#include <wx/listctrl.h>
#include <wx/timer.h>
#include <wx/wx.h>

const long itemcount = 20000000;

class MyListCtrl : public wxListCtrl {
  public:
   MyListCtrl(wxWindow* parent, wxWindowID id)
       : wxListCtrl(parent, id, wxDefaultPosition, wxDefaultSize,
                    wxLC_REPORT | wxLC_VIRTUAL) {
      AppendColumn("Column", wxLIST_FORMAT_LEFT);

      SetItemCount(itemcount + 100);
      EnsureVisible(itemcount); 
   }    
   wxString OnGetItemText(long item, long column) const override {
      return wxString::Format("Item %ld", item);
   }        
};

class MyFrame : public wxFrame {
  public:
[...]

PR #24670, opened 2024-07-03
wxMAC scrolbar: use doubleValue instead of floatValue

alcontre:
Use doubleValue instead of floatValue to improve precision of GetValue().

Address issue #24669

### DIFF
--- a/src/osx/cocoa/scrolbar.mm
+++ b/src/osx/cocoa/scrolbar.mm
@@ -62,7 +62,7 @@ public :
 
     virtual wxInt32 GetValue() const override
     {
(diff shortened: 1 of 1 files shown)


alcontre:
Rebased to pick up ci_yaml updates for macOS runners.

vadz:
Thanks, this change makes sense to me even if it was like this ever since the original dbeddfb93d (committing current cocoa state just to be safe ..., 2008-08-30) which added this code.

@csomor If you don't mind, I'll apply this to both master and 3.2.

csomor:
@vadz yes, please do so. @alcontre thanks a lot for debugging this!

PR #24671, opened 2024-07-03
Draw checkboxes and radiobuttons at correct size on high DPI.

### LINKED ISSUE #24650: Check box of non-selected `wxBoolProperty` is not vertically centered at high DPI
At high DPI, the check box of a `wxBoolProperty` is not vertically centered unless the property is the selected one.
So the vertical position of the check box differs depending on whether the `wxBoolProperty` is selected in the `wxPropertyGrid` or not.

The issue can also be seen in the propgrid sample.

wxBoolProperty-check-box-300-DPI-scaling

### LINKED ISSUE #24651: Check mark of `wxBoolProperty` looks blurry at high DPI
As can be seen in the propgrid sample, the check mark of a `wxBoolProperty` looks sharp at 100% DPI scaling but blurry at 200% and 300%.

wxBoolProperty-check-mark-100-200-300-DPI-scaling

The check mark of a `wxCheckBox` looks sharp even at 300% DPI scaling.

wxCheckBox-check-mark-300-DPI-scaling

dsa-t:
When the display scale is not 100%, MSW scales these controls badly. This commit changes the logic to draw them at correct size.

Closes #24651.

---

Fix vertical centering of propgrid checkboxes on high DPI.
Closes https://github.com/wxWidgets/wxWidgets/issues/24650.

### DIFF
--- a/src/msw/renderer.cpp
+++ b/src/msw/renderer.cpp
@@ -754,8 +754,6 @@ wxRendererXP::DoDrawButtonLike(wxUxThemeHandle& hTheme,
 {
     wxCHECK_RET( dc.GetImpl(), wxT("Invalid wxDC") );
 
-    RECT r = ConvertToRECT(dc, rect);
-
     // the base state is always 1, whether it is PBS_NORMAL,
     // {CBS,RBS}_UNCHECKEDNORMAL or CBS_NORMAL
     int state = 1;
@@ -787,6 +785,20 @@ wxRendererXP::DoDrawButtonLike(wxUxThemeHandle& hTheme,
     // wxCONTROL_ISDEFAULT flag is only valid for push buttons
     else if ( part == BP_PUSHBUTTON && (flags & wxCONTROL_ISDEFAULT) )
         state = PBS_DEFAULTED;
+
+    wxRect crect = rect;
+
+    // Checkboxes and radiobuttons must be drawn at native size, otherwise
+    // they get blurry on non-unity display scales
+    if ( part == BP_CHECKBOX || part == BP_RADIOBUTTON )
+    {
+        if ( ::IsThemePartDefined(hTheme, part, 0) )
+        {
+            crect = wxRect(hTheme.GetDrawSize(part, state)).CenterIn(rect);
+        }
+    }
+
+    RECT r = ConvertToRECT(dc, crect);
 
     hTheme.DrawBackground(GetHdcOf(dc.GetTempHDC()), r, part, state);
 }
--- a/src/propgrid/editors.cpp
+++ b/src/propgrid/editors.cpp
@@ -1555,7 +1555,7 @@ public:
         // Due to SetOwnFont stuff necessary for GTK+ 1.2, we need to have this
         wxControl::SetFont( parent->GetFont() );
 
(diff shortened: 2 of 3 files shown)


src/propgrid/editors.cpp
```diff
@@ -1555,7 +1555,7 @@ class wxSimpleCheckBox : public wxControl
         // Due to SetOwnFont stuff necessary for GTK+ 1.2, we need to have this
         wxControl::SetFont( parent->GetFont() );
 
-        SetBoxHeight(12);
+        SetBoxHeight(13);
```
vadz, line 1558: Shouldn't this be `wxRendererNative::Get()->GetCheckBoxSize().y`?

src/msw/renderer.cpp
```diff
@@ -788,6 +786,20 @@ wxRendererXP::DoDrawButtonLike(wxUxThemeHandle& hTheme,
     else if ( part == BP_PUSHBUTTON && (flags & wxCONTROL_ISDEFAULT) )
         state = PBS_DEFAULTED;
 
+    wxRect crect = rect;
+
+    // Checkboxes and radiobuttons must be drawn at native size, otherwise
```
vadz, line 791: I'm not sure I agree with this comment, we don't do anything like this for any other elements, why should we do it here? IMO the caller should use the correct size, i.e. `GetCheckBoxSize()`, but if they want to use something else (e.g. `10*GetCheckBoxSize()` because they want a giant checkbox), we should respect their wishes.

vadz (changes requested):
Thanks for looking at this, but I think it would be better to provide the correct size to the renderer instead of trying to compensate in the wxMSW renderer code for this, as this makes it inconsistent with the other platform.

Or is there some problem with doing it like this that I'm missing?

src/msw/renderer.cpp
dsa-t: At 100% scale, if the rect is bigger than the checkbox, it's rendered size is still 13 px, it just get centered (or if we change scale at runtime from 100% to something else). This behaviour is replicated here.

It's probably a Windows bug that it gets scaled on non-100% display scale if we pass a bigger rectangle.

dsa-t: >we don't do anything like this for any other elements, why should we do it here?

Push buttons are not affected by this issue. Checkboxes and radiobuttons are.

src/propgrid/editors.cpp
dsa-t: Maybe. Needs additional checking.

vadz: Actually it doesn't seem to matter, as it's changed later anyhow, so I'll leave it as is but I think it could be removed completely. But then I also think this entire class could, and should, be removed completely...

src/msw/renderer.cpp
vadz: Let's do this, but in wxPG itself and not here, like this it certainly won't break any other uses of this function.

vadz (commented):
Sorry for the delay, but I've finally got back to it and I think #24754 should have the same effect but be safer and not affect anything else, so let's apply it instead.

PR #24675, opened 2024-07-03
Add GetHWND() to wxWindowGTK when on Windows

ryancog:
As the title says, when using wxGTK on Windows there's not a way (that I can tell) to get an HWND from a wxWindow. This uses a couple of gdk and gtk helpers to get the HWND from the underlying GtkWidget.

### DIFF
--- a/include/wx/gtk/window.h
+++ b/include/wx/gtk/window.h
@@ -140,8 +140,12 @@ public:
 
     virtual WXWidget GetHandle() const override { return m_widget; }
 #ifdef __WINDOWS__
-    WXHWND GetHWND() const;
+    WXHWND GTKGetWin32Handle() const;
 #endif
+
+    // If the underlying GTK widget hasn't been realized yet, do so.
+    // Does nothing if the widget is already realized.
+    void GTKRealizeWidget();
 
     // many important things are done here, this function must be called
     // regularly
--- a/include/wx/gtk/window.h
+++ b/include/wx/gtk/window.h
@@ -139,13 +139,12 @@ public:
     // --------------
 
     virtual WXWidget GetHandle() const override { return m_widget; }
+
 #ifdef __WINDOWS__
+    // If on Windows, cut through the GtkWidget abstraction
+    // to get HWND.
     WXHWND GTKGetWin32Handle() const;
 #endif
-
-    // If the underlying GTK widget hasn't been realized yet, do so.
-    // Does nothing if the widget is already realized.
-    void GTKRealizeWidget();
 
     // many important things are done here, this function must be called
     // regularly
--- a/include/wx/gtk/window.h
+++ b/include/wx/gtk/window.h
@@ -139,6 +139,9 @@ public:
     // --------------
 
     virtual WXWidget GetHandle() const override { return m_widget; }
+#ifdef __WINDOWS__
+    WXHWND GetHWND() const;
+#endif
 
     // many important things are done here, this function must be called
     // regularly
--- a/src/gtk/window.cpp
+++ b/src/gtk/window.cpp
@@ -4244,7 +4244,8 @@ bool wxWindowGTK::GTKShowFromOnIdle()
 }
 
 #ifdef __WINDOWS__
-WXHWND wxWindowGTK::GetHWND() const {
+WXHWND wxWindowGTK::GetHWND() const
+{
     // There isn't an underlying HWND if the widget hasn't been realized yet.
     gtk_widget_realize(m_widget);
     return static_cast<WXHWND>(gdk_win32_window_get_handle(gtk_widget_get_window(m_widget)));
--- a/src/gtk/window.cpp
+++ b/src/gtk/window.cpp
@@ -37,6 +37,9 @@
 #ifdef __WXGTK3__
     #include "wx/gtk/dc.h"
 #endif
+#ifdef __WINDOWS__
+    #include <gdk/gdkwin32.h>
+#endif
 
 #include <ctype.h>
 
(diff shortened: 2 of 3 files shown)


ryancog:
To give some context as to why I want this (although I'm sure there's plenty of other uses), in order to set the title bar to be dark when `GTK_CSD=0` (And I always turn CSD off... it's awful under windows), the window's HWND is required:

```c++
#	ifdef __WINDOWS__
    DWORD useDarkMode{Misc::getTheme() == Misc::Theme::DARK_MODE};
    DwmSetWindowAttribute(
            GetHWND(),
            DWMWINDOWATTRIBUTE::DWMWA_USE_IMMERSIVE_DARK_MODE,
            &useDarkMode,
            sizeof(DWORD)
            );
#	endif
```

src/gtk/window.cpp
```diff
@@ -4240,6 +4243,14 @@ bool wxWindowGTK::GTKShowFromOnIdle()
     return false;
 }
 
+#ifdef __WINDOWS__
+WXHWND wxWindowGTK::GetHWND() const {
```
vadz, line 4247: Very minor, but wx style is to put braces on their own lines, i.e.
```suggestion
WXHWND wxWindowGTK::GetHWND() const
{
```

include/wx/gtk/window.h
```diff
@@ -139,6 +139,9 @@ class WXDLLIMPEXP_CORE wxWindowGTK : public wxWindowBase
     // --------------
 
     virtual WXWidget GetHandle() const override { return m_widget; }
+#ifdef __WINDOWS__
+    WXHWND GetHWND() const;
```
vadz, line 143: I'm not sure if it's a good or bad idea to use the same name as for the existing function in wxMSW. On one hand, this would allow referring to it without checking for `__WXGTK__` but OTOH do we really plan to have much code using either wxMSW or wxGTK and having to manipulate the underlying `HWND` directly?

Perhaps we should call this `GTKGetWin32Handle()` instead? Granted, this is much longer but this shouldn't be used very often, hopefully.

src/gtk/window.cpp
```diff
@@ -4240,6 +4243,14 @@ bool wxWindowGTK::GTKShowFromOnIdle()
...
+    // There isn't an underlying HWND if the widget hasn't been realized yet.
+    gtk_widget_realize(m_widget);
```
vadz, line 4249: This is more serious, IMO it's really unexpected for an accessor like this to have significant side effects.

Would it be a problem to just call this function only if the window had been already realized?

vadz (commented):
Thanks, I don't think there is a problem with adding such a function but I'm not sure about the name and it should also be documented in `interface/wx/window.h` if we mean it to be part of the public API now.

src/gtk/window.cpp
ryancog: Whoops, my bad... I didn't realize the difference in styling it seems.

include/wx/gtk/window.h
ryancog: > but OTOH do we really plan to have much code using either wxMSW or wxGTK and having to manipulate the underlying `HWND` directly?

I can't imagine so.

> Perhaps we should call this `GTKGetWin32Handle()` instead? Granted, this is much longer but this shouldn't be used very often, hopefully.

Making it a different, distinguishable name makes sense to me.

Mostly out of curiousity, what's the logic stylistically behind prefixing it with `GTK`?
Since there's other wxWindowGTK functions that aren't...

src/gtk/window.cpp
ryancog: Would need to add an extra check on the first gtk function, since it'll return null if not (and passing that to gdk will cause *bad things*), but that's easy enough.

However, that'd require a way (for my use case at least) to be able to realize the GTK Window through wxWidgets. I'll see if there's a function for that already. If not, I'd want to add one.

ryancog: I don't see a method to "force" realize a gtk window.

Are you aware if there's something like this that exists and/or what sort of side effects would need to be considered by creating another function (GTKRealizeWidget() for example) to do this?

ryancog:
> it should also be documented in interface/wx/window.h if we mean it to be part of the public API now.

Should it go in `interface/wx/window.h` or into a gtk specific one? That file currently seems to be entirely platform-agnostic functions, so I'm not sure how to go about adding that documentation.

vadz:
Thanks for the changes!

@paulcor Do you mind adding `GTKRealizeWidget()`? I'm not totally sure if it's a good idea, but OTOH I don't have any better ones.

paulcor:
A shown window is realized, it might be sufficient just to require that rather than adding `GTKRealizeWidget()`.

ryancog:
That feels a bit hacky... especially if I don't want to show it yet. Would `Show(); Hide();` work? Even then it isn't the most obvious...

paulcor:
`wxWindowCreateEvent` is generated when a window is realized, things requiring the `GdkWindow` can probably be done from a handler for that event.

ryancog:
Ok, that seems reasonable then. Something I think would be reasonable to note writing documentation for `GTKGetWin32Handle()`. 

Speaking of which, I’m still not entirely sure where to put such documentation.

vadz:
> Speaking of which, I’m still not entirely sure where to put such documentation.

In `interface/wx/window.h`, please. TIA!

ryancog:
Ok, I've removed `GTKRealizeWidget()`, and added documentation for `GTKGetWin32Handle()`.

Let me know if there's anything wrong with that documentation... Not sure if it's overly verbose or anything along those lines, and I'm unsure if there's extra formatting I should do. (Closest thing to doxygen I've used is JavaDocs... and even then it was only lightly)

ryancog:
Bumping this.

If there's no issues with the documentation or anything else, I don't have anything more to add.

vadz:
Sorry for the delay, will merge this soon.

Issue #24667, opened 2024-07-02
3.2.5 regression on macOS: wxListCtrl and wxTreeCtrl no longer get focused on mouse click


codesquid:
### Description

This is a new regression in wxWidgets 3.2.5 on macOS, this functionality worked correctly in wxWidgets 3.2.4

When left-clicking wxListCtrl or wxTreeCtrl, the control no longer takes focus. Clicking an item in these controls selects said item, but the control itself stays unfocused:
- The background color of the selected items stays grey
- Changing the selected item with the cursor keys on the keyboard is then not possible
- Slowly clicking an item twice no longer starts the label editing

This can be reproduced in the listctrl and treectrl samples: Start the sample, click an item in the list/tree and then try to use the cursor keys to change the selected item. In 3.2.5, it does not work, in 3.2.4 the selection changes as expected.

It is however still possible to give the control focus by navigating to it using the Tab key.

### Platform and version information

- wxWidgets version  you use: 3.2.5
- wxWidgets port  you use: wxOSX
- OS  and its version: macOS 14.5 (23F79) on Apple M2
- Compiler: Apple clang version 15.0.0 (clang-1500.3.9.4) from XCode 15.4 (15F31d)

vadz:
Thanks for reporting this. Unfortunately I don't see any smoking gun, `git log v3.2.4..v3.2.5 src/osx` shows just a few commits none of which seem to be obviously related. If you can run git-bisect to find the commit which actually broke this, it would be great.

codesquid:
It's this commit: https://github.com/wxWidgets/wxWidgets/commit/abab8b43855127c7556e021b8ddb954871fdbcb5

If I reverse it, the focus on left-click is back to normal.

codesquid:
Upon further testing, reversing that commit only fixes it for the listctrl sample, it does nothing for the treectrl sample.

Furthermore, even in the listctrl sample, there is still a small 20x20 in the top left of the list that upon click doesn't give the control focus.

codesquid:
Reverting both https://github.com/wxWidgets/wxWidgets/commit/abab8b43855127c7556e021b8ddb954871fdbcb5 and https://github.com/wxWidgets/wxWidgets/commit/b8744be566195d9adf91be9fd5bb6c2de944c900 makes both the listctrl and treectrl samples work correctly again.

vadz:
[...]

PR #24678, opened 2024-07-05
OSX fix focus issue for scrolling windows on macOS 14+

csomor:
see #24667

### DIFF
--- a/src/osx/cocoa/window.mm
+++ b/src/osx/cocoa/window.mm
@@ -3245,6 +3245,21 @@ bool wxWidgetCocoaImpl::CanFocus() const
 
 @implementation wxNSClipView
 
++ (void)initialize
+{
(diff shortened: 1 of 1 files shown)


andykt:
@csomor and @vadz: There is still a regression with scrolled windows like wxHtmlWindow not seeing mouse motion/hover events. If you run samples/html/test/test.app in macOS Sonoma you'll find that you have to click in the window first before it will see motion/hover events.  I suspect the code in wxWidgetCocoaImpl::InstallEventHandler needs to check if m_osxClipView is set and do the NSTracking and Associate calls on that view rather than m_osxView.

csomor:
@vadz @andykt, see the sequel PR #24809 which attempts to fix the bug mentioned in #24790

PR #24679, opened 2024-07-05
Add panel to sizer when not using wxCollapsiblePane in wxLogDialog.

dsa-t:
Not using `wxCollapsiblePane` in `wxLogDialog` didn't work correctly before this change.


vadz:

PR #24680, opened 2024-07-05
wxQt: avoid SIGPIPE when debugging on X11.

dsa-t:
See https://stackoverflow.com/questions/56686603/how-to-avoid-sigpipe-due-to-a-timeout-when-debugging-an-x11-program

### DIFF
--- a/src/qt/app.cpp
+++ b/src/qt/app.cpp
@@ -63,7 +63,6 @@ bool wxApp::Initialize( int &argc, wxChar **argv )
     QSurfaceFormat::setDefaultFormat(format);
 
     m_qtApplication.reset(new QApplication(m_qtArgc, m_qtArgv.get()));
-    m_qtApplication->processEvents(); // Avoids SIGPIPE when debugging
 
     if ( m_qtApplication->platformName() == "xcb" )
         m_qtApplication->processEvents(); // Avoids SIGPIPE on X11 when debugging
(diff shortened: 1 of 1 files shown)


vadz:
Shouldn't we only do this when using X11? I have no idea what does `processEvents()` actually do, but it seems we shouldn't call it unless we have to.

dsa-t:
I haven't checked if it's needed on other platforms yet.

AliKet:
It would really be great to guard this change so that it runs only under X11 and if a debugger is attached? is this possible ?

dsa-t:
Added a commit that only calls it on "xcb" platform. Not tested yet.

vadz:
BTW, this is pretty weird because I've been having fun[*] , with libxcb during the last few days and I didn't have any such problems when using it directly, I wonder what does Qt do to trigger this `SIGPIPE`.

But if you still think it's worth applying, I can merge it. Is it?

vadz:
@dsa-t Do you have any thoughts about this one? Also, the current diff looks strange because it adds a `processEvent()` call both unconditionally _and_ conditionally on the platform, is this really intentional?

dsa-t:
> I wonder what does Qt do to trigger this `SIGPIPE`.

A quote from StackOverflow:

**_Apparently, Qt somehow "confirms" application's validity to X11 in response to some event, which then lets the application be as slow as it pleases, without any fatal penalties._**

There's a bug report on Qt bugtracker as well: https://bugreports.qt.io/browse/QTBUG-58709

> But if you still think it's worth applying, I can merge it. Is it?

Sure.

> Also, the current diff looks strange because it adds a `processEvent()` call both unconditionally _and_ conditionally on the platform, is this really intentional?

Fixed.

Issue #24666, opened 2024-07-02
Print preview too small at high DPI


taler21:
As can be seen in the printing sample, the print preview is too small at 200% DPI scaling.

If running the sample DPI unaware and DPI aware, the preview should have the same size on the screen in both cases.
But currently, in case of DPI awareness, the preview is too small by the DPI scaling factor.

![print-preview-DPI-unaware-vs-aware](https://github.com/wxWidgets/wxWidgets/assets/99262969/2024b023-8858-4fbe-8707-e8d8a929ff43)

It seems that the size of `wxPreviewFrame` can be easily fixed as follows

```patch
diff --git "a/src/common/prntbase.cpp" "b/src/common/prntbase.cpp"
index 408c4b6336..54b12ea20b 100644
--- "a/src/common/prntbase.cpp"
+++ "b/src/common/prntbase.cpp"
@@ -940,7 +940,7 @@ wxScrolledWindow(parent, wxID_ANY, pos, size, style | wxFULL_REPAINT_ON_RESIZE,
 
     // Use some reasonable default size for this window, roughly proportional
     // to the paper sheet.
-    SetInitialSize(wxSize(600, 750));
+    SetInitialSize(FromDIP(wxSize(600, 750)));
 }
 
 wxPreviewCanvas::~wxPreviewCanvas()

```

![print-preview-DPI-unaware-vs-aware-with-patched-frame](https://github.com/wxWidgets/wxWidgets/assets/99262969/809b5066-1de8-437d-9413-23a6360f7487)

but I have no idea how to fix the size of its content (the preview of the page).

### Platform and version information

- wxWidgets version  you use: 3.2.5
- wxWidgets port  you use: wxMSW
- OS  and its version: Windows 10

dsa-t:
See https://github.com/dsa-t/wxWidgets/commit/c9cfb7f85f27d9b3a37208a9c87098a7ffd3b362

Not tested on other operating systems.

I noticed that window sizing can be wrong when moving between screens for some reason (unrelated bug).

taler21:
I can confirm that with your changes it looks correct with wxMSW.
But your changes have side effects with wxGTK, where the preview of the page becomes too large on a high-resolution monitor (at 200% scaling).

taler21:
The changes (backported to 3.2.5) also cause my actual application with wxGTK to crash.

Call stack:

```
[...]

PR #24681, opened 2024-07-05
Improve printing preview on high-DPI monitors.

dsa-t:
See https://github.com/wxWidgets/wxWidgets/issues/24666

### DIFF
--- a/src/common/prntbase.cpp
+++ b/src/common/prntbase.cpp
@@ -1984,14 +1984,17 @@ bool wxPrintPreviewBase::UpdatePageRendering()
 {
     static double s_sf = 0.0;
 
-    double newScaleFactor = m_previewCanvas->GetDPIScaleFactor();
-
-    if ( m_previewCanvas && newScaleFactor != s_sf )
+    if ( m_previewCanvas )
     {
-        InvalidatePreviewBitmap();
-        AdjustScrollbars(m_previewCanvas);
+        double newScaleFactor = m_previewCanvas->GetDPIScaleFactor();
 
-        s_sf = newScaleFactor;
+        if ( newScaleFactor != s_sf )
+        {
+            InvalidatePreviewBitmap();
+            AdjustScrollbars(m_previewCanvas);
+
+            s_sf = newScaleFactor;
+        }
     }
 
     if ( m_previewBitmap )
--- a/src/common/prntbase.cpp
+++ b/src/common/prntbase.cpp
@@ -940,7 +940,7 @@ wxScrolledWindow(parent, wxID_ANY, pos, size, style | wxFULL_REPAINT_ON_RESIZE, 
 
     // Use some reasonable default size for this window, roughly proportional
     // to the paper sheet.
-    SetInitialSize(wxSize(600, 750));
+    SetInitialSize(FromDIP(wxSize(600, 750)));
 }
 
 wxPreviewCanvas::~wxPreviewCanvas()
@@ -949,6 +949,9 @@ wxPreviewCanvas::~wxPreviewCanvas()
 
 void wxPreviewCanvas::OnPaint(wxPaintEvent& WXUNUSED(event))
 {
+    if ( m_printPreview )
+        m_printPreview->UpdatePageRendering();
+
     wxPaintDC dc(this);
     PrepareDC( dc );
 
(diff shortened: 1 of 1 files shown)


src/common/prntbase.cpp
```diff
@@ -1976,6 +1982,18 @@ void wxPrintPreviewBase::InvalidatePreviewBitmap()
 
 bool wxPrintPreviewBase::UpdatePageRendering()
 {
+    static double s_sf = 0.0;
+
+    double newScaleFactor = m_previewCanvas->GetDPIScaleFactor();
+
+    if ( m_previewCanvas && newScaleFactor != s_sf )
```
vadz, line 1989: It doesn't make sense to check for a pointer we just dereferenced above. Either this check is not needed at all or it should be done first.

src/common/prntbase.cpp
```diff
@@ -1976,6 +1982,18 @@ void wxPrintPreviewBase::InvalidatePreviewBitmap()
 
 bool wxPrintPreviewBase::UpdatePageRendering()
 {
+    static double s_sf = 0.0;
```
vadz, line 1985: Sorry, why is this static? It's possible to have different scale factors for different windows, so it should be a member variable, AFAICS.

src/common/prntbase.cpp
```diff
@@ -2095,7 +2109,8 @@ bool wxPrintPreviewBase::RenderPage(int pageNum)
 
     if (!m_previewBitmap)
     {
-        m_previewBitmap = new wxBitmap(pageRect.width, pageRect.height);
+        m_previewBitmap = new wxBitmap;
+        m_previewBitmap->CreateWithLogicalSize(pageRect.width, pageRect.height, m_previewCanvas->GetContentScaleFactor());
```
vadz, line 2113: It would be useful to have a comment here explaining why the page rect is in logical size rather than in DIPs because this is not really obvious.

src/common/prntbase.cpp
```diff
@@ -949,6 +949,9 @@ wxPreviewCanvas::~wxPreviewCanvas()
 
 void wxPreviewCanvas::OnPaint(wxPaintEvent& WXUNUSED(event))
 {
+    if ( m_printPreview )
+        m_printPreview->UpdatePageRendering();
+
```
vadz, line 954: I'm really not sure this should be done here. There is a comment saying that `UpdatePageRendering()` is called during idle time only and while I don't fully understand the logic of doing this, it seems wrong to call it both from there and from here.

If we need to update the scale factor, we should just call some (private) `DoUpdateScaleFactorIfNeeded()` but OTOH I don't even know why should we do this from here and not during initialization and then from `wxEVT_DPI_CHANGED` handler?

vadz (changes requested):
Thanks for working on this, this definitely should be improved, but I don't think this is the best solution as written right now.

src/common/prntbase.cpp
dsa-t: > I don't even know why should we do this from here and not during initialization and then from `wxEVT_DPI_CHANGED` handler?

Yes, it should be better to call it from `wxEVT_DPI_CHANGED` handler. I'm just not sure what's the proper way to not break the ABI here.

src/common/prntbase.cpp
dsa-t: True. But I don't know if it will break ABI.

src/common/prntbase.cpp
dsa-t: This can be made to work with DIPs in pageRect in theory, but this was meant more like a proof-of-concept, so feel free to modify it.

src/common/prntbase.cpp
dsa-t: Fixed.

src/common/prntbase.cpp
vadz: Adding a new (non-virtual, private) function doesn't break ABI, so we could add `wxEVT_DPI_CHANGED` handler and `Bind()` to it.

src/common/prntbase.cpp
vadz: In master we don't care about ABI. In 3.2, we need use the standard "emulation" of member variables which consists in adding a hash map with `this` as key and this value in it.

src/common/prntbase.cpp
vadz: Modifying it is simple, it's testing it that takes time...

vadz:
I think this was superseded by #24815, is there still anything useful here?

dsa-t:
> I think this was superseded by #24815, is there still anything useful here?

This one can be closed now.

PR #24683, opened 2024-07-06
wxQt: improve wxGLCanvas latency.

dsa-t:
By setting default swap behavior mode to SingleBuffer, we can get much lower latency when working with wxGLCanvas.

Comparison video of `pyramid` (on Linux Mint/X11, 60 Hz monitor, 125 Hz mouse, captured in 240 fps, at 1/8 speed):

"VSync On, Double (buffering)" is before this PR.
"VSync On, Single (buffering)" is after this PR.

https://github.com/wxWidgets/wxWidgets/assets/37658952/1b569639-e0da-44f2-9cb7-06a989607aab

"VSync Off, Single (buffering)" is possible, but needs to be checked if it's better.

### DIFF
--- a/src/qt/app.cpp
+++ b/src/qt/app.cpp
@@ -14,6 +14,7 @@
 #include "wx/qt/private/converter.h"
 #include <QtCore/QStringList>
 #include <QtWidgets/QApplication>
+#include <QSurfaceFormat>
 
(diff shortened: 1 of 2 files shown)


AliKet:
LGTM, thanks.

vadz:
Thanks, will merge soon.

PR #24684, opened 2024-07-06
Update macos-11 runners to macos-12

alcontre:
Github has deprecated macos-11 runners; update to use the oldest supported runners for CI


alcontre:
Would it be worth adding macos-13 and macos-14 runners as well?

vadz:
Thanks, we indeed need to do this.

It would probably be good to switch one of the builds to macOS 14, but I don't see much point in testing intermediate versions.

PR #24685, opened 2024-07-07
Better image rescale quality when enlarging

vadz:
See #24567.

### DIFF
--- a/interface/wx/image.h
+++ b/interface/wx/image.h
@@ -29,7 +29,14 @@ enum wxImageResolution
  */
 enum wxImageResizeQuality
 {
-    /// Simplest and fastest algorithm.
+    /**
+        Simplest and fastest algorithm.
+
+        In wxWidgets versions before 3.3.0 this used to be the same as
+        wxIMAGE_QUALITY_NORMAL, but this is no longer the case when downscaling
+        the images. You can use the new wxIMAGE_QUALITY_FAST as a synonym for
+        this algorithm if speed is paramount.
+     */
     wxIMAGE_QUALITY_NEAREST,
 
     /// Compromise between wxIMAGE_QUALITY_NEAREST and wxIMAGE_QUALITY_BICUBIC.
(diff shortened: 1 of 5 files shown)


src/common/image.cpp
```diff
@@ -463,6 +463,19 @@ wxImage::Scale( int width, int height, wxImageResizeQuality quality ) const
     // Resample the image using the method as specified.
     switch ( quality )
     {
+        case wxIMAGE_QUALITY_NORMAL:
+            // When downscaling, we prefer to use bilinear resampling as it
+            // results in better quality at reasonable speed.
+            if ( width < old_width && height < old_height )
```
dsa-t, line 469:
```suggestion
            if ( width < old_width || height < old_height )
```

vadz: Sorry, why? My understanding is that we want to avoid bilinear when increasing size, even in one of the directions.

One thing we could probably change here is what to do in case of equality, i.e.

```suggestion
            if ( width <= old_width && height <= old_height )
```

dsa-t: >that we want to avoid bilinear when increasing size, even in one of the directions.

Seems safer to prefer bilinear resampling to me. Doesn't really matter though.

vadz: If this improves things for KiCad icons, let's merge it already and we can always improve it later IMO.

interface/wx/image.h
```diff
@@ -46,16 +53,34 @@ enum wxImageResizeQuality
     wxIMAGE_QUALITY_BOX_AVERAGE,
 
     /**
-    Default image resizing algorithm used by wxImage::Scale(). Currently
-    the same as wxIMAGE_QUALITY_NEAREST.
+        Default image resizing algorithm used by wxImage::Scale().
+
+        Currently this is the same as wxIMAGE_QUALITY_NEAREST when enlarging
+        the image and wxIMAGE_QUALITY_BILINEAR when reducing it, which reduces
```
dsa-t, line 59:
```suggestion
        the image and wxIMAGE_QUALITY_BILINEAR when reducing it, which results
```

vadz: Oops, thanks!

src/common/image.cpp
dsa-t: Sure, thanks!

PR #24691, opened 2024-07-09
wxQt: Stop overriding wxWindow::GetHandle() in derived classes

AliKet:
There is no need to do so because assigning to `m_qtWindow` is enough to return a valid handle.

There is also no need to store an extra pointer just to preserve the actual type of the handle because we can simply get it with the added function `GetQtHandle()`.

There are no real changes other than simplifying and normalizing the code base.

### DIFF
--- a/include/wx/qt/anybutton.h
+++ b/include/wx/qt/anybutton.h
@@ -26,17 +26,15 @@ public:
     virtual void SetLabel( const wxString &label ) override;
     virtual wxString GetLabel() const override;
 
-    virtual QWidget *GetHandle() const override;
-
     // implementation only
     void QtUpdateState();
     virtual int QtGetEventType() const = 0;
 
+    QPushButton* GetQPushButton() const;
+
 protected:
     virtual wxBitmap DoGetBitmap(State state) const override;
     virtual void DoSetBitmap(const wxBitmapBundle& bitmap, State which) override;
-
-    QPushButton *m_qtPushButton;
 
     void QtCreate(wxWindow *parent);
     void QtSetBitmap( const wxBitmapBundle &bitmap );
--- a/include/wx/qt/anybutton.h
+++ b/include/wx/qt/anybutton.h
@@ -18,7 +18,7 @@ class QPushButton;
 class WXDLLIMPEXP_CORE wxAnyButton : public wxAnyButtonBase
 {
 public:
-    wxAnyButton();
+    wxAnyButton() = default;
 
     // implementation
     // --------------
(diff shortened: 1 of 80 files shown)


include/wx/qt/defs.h
```diff
@@ -12,6 +12,19 @@
 
 typedef class QWidget *WXWidget;
 
+/* ------------------------------------------------------------------------- */
+/*  macros to return the real Qt widget type corresponding to wx widget type */
+/* ------------------------------------------------------------------------- */
+
+#define wxDECLARE_GET_QT_HANDLE(qtwidget)     			\
```
vadz, line 19: It seems a pity to add yet another widely used macro, IMHO it really doesn't have enough advantages compared to just writing out the accessor in full to use it.

Also, this might be over-engineering, but it should be possible to use CRTP here instead, i.e.

```cpp
template <typename QTWIDGET, typename WXCLASS>
class wxQtHandleAccessor
{
public:
    using QtWidget = QTWIDGET;
    QtWidget* GetQtHandle() const { return static_cast<QtWidget*>(static_cast<WXCLASS*>(this)->QtGetWindow()); }
};
```

and then derive `wxFoo` from `wxQtHandleAccessor<QtFoo, wxFoo>` to get the implementation of this function.

vadz (changes requested):
Sorry, I'm really not convinced of the need for these macros, at least/especially the declaration one which makes the headers more readable and I'd like to replace it with the actual function declaration.

Whether you prefer to keep the `wxIMPLEMENT` one or not is up to you, but I'd put it in some header under `wx/qt/private` if you do.

TIA!

include/wx/qt/defs.h
AliKet: > Also, this might be over-engineering, but it should be possible to use CRTP here instead, i.e.
> 
> ```c++
> template <typename QTWIDGET, typename WXCLASS>
> class wxQtHandleAccessor
> {
> public:
>     using QtWidget = QTWIDGET;
>     QtWidget* GetQtHandle() const { return static_cast<QtWidget*>(static_cast<WXCLASS*>(this)->QtGetWindow()); }
> };
> ```
> 
> and then derive `wxFoo` from `wxQtHandleAccessor<QtFoo, wxFoo>` to get the implementation of this function.

I did something similar before making this PR but it requires the corresponding Qt header to be included for the cast to succeed.

I just force pushed my initial version of the changes, but if you still have any suggestion to improve it, I'll happily implement it. TIA!

vadz (approved):
Thanks, I see no problems with merging it like this.

Should I do it?

AliKet:
> Thanks, I see no problems with merging it like this.
> 
> Should I do it?

Yes go ahead. thanks.

PR #24692, opened 2024-07-10
Add wxSystemHardwareInfo class

### LINKED ISSUE #24652: Adding GetTotalPhysicalMemory() function
Would adding a `GetTotalPhysicalMemory()` function (which would return the size of installed RAM in bytes) be useful? I have this implemented for MSW, macOS, and remaining Unices if anyone would like for me to add this. I use this for logging system information for crash reports myself; perhaps others would to.

If so, where would be the best place for it? Perhaps `wxPlatformInfo`, although this isn't arguably a platform feature. Or it could be a global `wxGetTotalPhysicalMemory()` function.

P.S.
MS calls it "physically installed system memory" (https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getphysicallyinstalledsystemmemory?redirectedfrom=MSDN), hence the API that I suggest. I could say `GetTotalRAM()` instead.

Blake-Madden:
Adds `GetPhysicalMemory()` (all platforms), `IsRunningNatively()` (all platforms), and `GetRemainingBatteryLifePercent()` (MSW).

Also, wraps `wxGetFreeMemory()`, `wxGetCpuArchitectureName()`, `wxGetNativeCpuArchitectureName()`, `wxThread::GetCPUCount()`, `wxGetBatteryState()`, and `wxGetPowerType()` free functions.

Closes #24652

I designed it as a simple collection of static functions because, unlike `wxPlatformInfo`, most of the things in here are constantly changing.

### DIFF
--- a/include/wx/hwinfo.h
+++ b/include/wx/hwinfo.h
@@ -22,7 +22,7 @@
 class WXDLLIMPEXP_BASE wxSystemHardwareInfo
 {
 public:
-    
+
     static int GetCPUCount();
 
     static wxString GetCPUArchitectureName();
--- a/include/wx/hwinfo.h
+++ b/include/wx/hwinfo.h
@@ -1,0 +1,40 @@
+///////////////////////////////////////////////////////////////////////////////
+// Name:        wx/hwinfo.h
+// Purpose:     declaration of the wxSystemHardwareInfo class
+// Author:      Blake Madden
+// Created:     07.09.2024
+// Copyright:   (c) 2024 Blake Madden
+// Licence:     wxWindows licence
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef _WX_SYSTEMHWINFO_H_
+#define _WX_SYSTEMHWINFO_H_
+
+#include "wx/string.h"
+#include "wx/utils.h"
+#include "wx/power.h"
+
+// ----------------------------------------------------------------------------
+// wxSystemHardwareInfo
+// ----------------------------------------------------------------------------
+
+// Information about the system's hardware
+class WXDLLIMPEXP_BASE wxSystemHardwareInfo
+{
+public:
+    
+    static int GetCPUCount();
+
+    static wxString GetCPUArchitectureName();
+    static wxString GetNativeCPUArchitectureName();
+    static bool IsRunningNatively();
+
+    static wxMemorySize GetPhysicalMemory();
+    static wxMemorySize GetFreePhysicalMemory();
+
+    static wxBatteryState GetBatteryState();
+    static int GetRemainingBatteryLifePercent();
+    static wxPowerType GetPowerType();
+};
+
+#endif // _WX_SYSTEMHWINFO_H_
--- a/include/wx/hwinfo.h
+++ b/include/wx/hwinfo.h
@@ -2,7 +2,7 @@
 // Name:        wx/hwinfo.h
 // Purpose:     declaration of the wxSystemHardwareInfo class
 // Author:      Blake Madden
-// Created:     07.09.2024
+// Created:     2024-07-09
 // Copyright:   (c) 2024 Blake Madden
 // Licence:     wxWindows licence
 ///////////////////////////////////////////////////////////////////////////////
--- a/interface/wx/hwinfo.h
+++ b/interface/wx/hwinfo.h
@@ -77,7 +77,7 @@ public:
     /**
         Returns @c true if the architecture that the program is running
         under is the same as the system's architecture. For example, this
-        will be @c if running a 32-bit application under x64 (Windoes) or
+        will be @c if running a 32-bit application under x64 (Windows) or
         running x86_64 under ARM64 (macOS)
     */
     static bool IsRunningNatively();
--- a/interface/wx/hwinfo.h
+++ b/interface/wx/hwinfo.h
@@ -50,7 +50,7 @@ public:
         The returned string may be empty if the CPU architecture
         couldn't be recognized.
 
-        @see wxGetCpuArchitectureName()
+        @see ::wxGetCpuArchitectureName()
      */
     static wxString GetCPUArchitectureName();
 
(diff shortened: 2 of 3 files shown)


Blake-Madden:
Note: I don't have the new source file included in the build system, I wasn't able to figure out how to do that yet.

include/wx/hwinfo.h
```diff
@@ -0,0 +1,40 @@
+///////////////////////////////////////////////////////////////////////////////
+// Name:        wx/hwinfo.h
+// Purpose:     declaration of the wxSystemHardwareInfo class
+// Author:      Blake Madden
+// Created:     07.09.2024
```
PBfordev, line 5: According to the the [Coding Guidelines](https://wxwidgets.org/develop/coding-guidelines/#wxwin_header), ISO format should be used.
```suggestion
// Created:     2024-07-09
```

The headers in other files are also not in accordance with the guidelines

interface/wx/hwinfo.h
```diff
@@ -0,0 +1,140 @@
...
+        Returns the CPU architecture name.
+
+        This can be, for example, "x86_64", "arm64", or "i86pc".
+        The name for the same CPU running on the same hardware
+        can vary across operating systems.
+
+        The returned string may be empty if the CPU architecture
+        couldn't be recognized.
+
+        @see wxGetCpuArchitectureName()
```
PBfordev, line 53:
```suggestion
        @see ::wxGetCpuArchitectureName()
```
Just a nitpick, to make it same as in the other methods.

PBfordev:
> Note: I don't have the new source file included in the build system, I wasn't able to figure out how to do that yet.

This is described here: https://github.com/wxWidgets/wxWidgets/blob/master/docs/contributing/how-to-add-files-to-build-system.md

I think I had some issues running publicly available build of bakefile, but it was a long time ago, so...

Blake-Madden:
> This is described here: https://github.com/wxWidgets/wxWidgets/blob/master/docs/contributing/how-to-add-files-to-build-system.md

Nice, thanks! If the general consensus is that this interface looks OK I'll work on that next...

include/wx/hwinfo.h
vadz, line 1: Question: do we really want to keep using short(ish) file names? I'd rather call it `wx/hardwareinfo.h`, it's not impossibly long and seems easier to remember.

include/wx/hwinfo.h
```diff
@@ -0,0 +1,40 @@
...
+#include "wx/string.h"
+#include "wx/utils.h"
+#include "wx/power.h"
+
+// ----------------------------------------------------------------------------
+// wxSystemHardwareInfo
+// ----------------------------------------------------------------------------
+
+// Information about the system's hardware
+class WXDLLIMPEXP_BASE wxSystemHardwareInfo
```
vadz, line 22: I think there are 2 choices that generally make sense:

1. Make this a namespace with functions inside it.
2. Make this a class with static `Get()` function and non-static accessors.

The latter makes sense if we ever want/need to cache some information, e.g. if `GetPhysicalMemory()` is called, we could also get the result of `GetFreePhysicalMemory()` together with it.

I'm not sure if we really need it, but having a class with just static functions doesn't seem to have any advantages compared to the first approach, so in this case I'd just use a namespace.

include/wx/hwinfo.h
```diff
@@ -0,0 +1,40 @@
...
+public:
+
+    static int GetCPUCount();
+
+    static wxString GetCPUArchitectureName();
+    static wxString GetNativeCPUArchitectureName();
+    static bool IsRunningNatively();
+
+    static wxMemorySize GetPhysicalMemory();
+    static wxMemorySize GetFreePhysicalMemory();
```
vadz, line 33: Sorry, I don't remember if we discussed this already, but do we really need "Physical" in the names of these functions? I think just `Get{Total,Free}Memory()` would be good enough.

Also, I wonder if we need `GetUsedMemory()`?

include/wx/hwinfo.h
```diff
@@ -0,0 +1,40 @@
...
+
+    static wxBatteryState GetBatteryState();
+    static int GetRemainingBatteryLifePercent();
+    static wxPowerType GetPowerType();
```
vadz, line 37: I'm not sure we really want to include these functions here, they don't really seem to fit ("power type" is not quite a hardware characteristic) and I think we might provide better API in some `wxBattery` class in `wx/power.h` instead.

Let's keep things minimal for now (it's always simpler to add than to remove them later) and omit this.

src/common/hwinfo.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+wxMemorySize wxSystemHardwareInfo::GetFreePhysicalMemory()
+{
+    return wxGetFreeMemory();
+}
+
+int wxSystemHardwareInfo::GetCPUCount()
+{
+    return 	wxThread::GetCPUCount();
```
vadz, line 96: Very nitpicky but
```suggestion
    return wxThread::GetCPUCount();
```

src/common/hwinfo.cpp
```diff
@@ -0,0 +1,112 @@
+///////////////////////////////////////////////////////////////////////////////
+// Name:        wx/hwinfo.h
```
vadz, line 2: Wrong file name.

src/common/hwinfo.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+#ifdef __WXMSW__
+    #include <windows.h>
+    #include <sysinfoapi.h>
+#elif defined(__APPLE__)
+    #include <sys/sysctl.h>
+#elif defined(__UNIX__)
+    #include <sys/resource.h>
+    #include <sys/sysinfo.h>
+#endif
```
vadz, line 35: I think it would be better to have platform-specific files rather than put everything in a single file here.

vadz (commented):
Sorry, I have several questions about this one. I'm not sure what the right answers are, but I'd like to at least check that we have good reasons to do things like this if we do.

BTW, have you looked at similar APIs in other libraries/languages? This might give some ideas...

Issue #13010, opened 2011-03-04
Context menus do not pop up when using Voiceover keys on the Mac.  j


wxtrac:
*Issue migrated from trac ticket # 13010*

**component:** GUI-all | **priority:** normal | **keywords:** contexty menu voiceover speech mac

#### 2011-03-04 07:33:50: kperry@blinksoft.com (Ken Perry) created the issue

___

If you create a context menu with any version of wxwidgets you can not activate it using the voiceover keys.  Voiceover is activated by pressing cmd-f5  then if you are on a control that has a context menu you should be able to do shift-control-option-m to bring up the context menu.  The right click works but the key press does not.  This makes it impossible to bring up a context menu in a wx widget program on the mac for a blind user.  Below is the program I used to test.

```py
#!/usr/bin/python

# contextmenu.py

import wx


class MyPopupMenu(wx.Menu):
    def __init__(self, parent):
        wx.Menu.__init__(self)

        self.parent = parent

        minimize = wx.MenuItem(self, wx.NewId(), 'Minimize')
        self.AppendItem(minimize)
        self.Bind(wx.EVT_MENU, self.OnMinimize, id=minimize.GetId())

        close = wx.MenuItem(self, wx.NewId(), 'Close')
        self.AppendItem(close)
        self.Bind(wx.EVT_MENU, self.OnClose, id=close.GetId())


    def OnMinimize(self, event):
        self.parent.Iconize()

    def OnClose(self, event):
        self.parent.Close()


class ContextMenu(wx.Frame):
    def __init__(self, parent, id, title):
        wx.Frame.__init__(self, parent, id, title, size=(250, 150))

        self.Bind(wx.EVT_CONTEXT_MENU, self.OnRightDown)

        self.Center()
        self.Show()

    def OnRightDown(self, event):
        print "test"

       # self.PopupMenu(MyPopupMenu(self), event.GetPosition())


app = wx.App()
frame = ContextMenu(None, -1, 'context menu')
app.MainLoop()

class MyPopupMenu(wx.Menu):
      def __init__(self, parent):
          wx.Menu.__init__(self)
```

PR #24697, opened 2024-07-12
Fixes to context menus in wxDVC on macOS

### LINKED ISSUE #849: Control-clicking translations list messes up selection on macOS
Steps to reproduce:

1. Show context menu on the list by pressing <kbd>control</kbd> and clicking with mouse.
2. Do not select any item in the popup menu,  click outside of the menu, on the list, to dismiss the menu.
3. Try moving selection to another item in the list and observe totally bizarre selection behavior.

Words fails me trying to describe step 3 accurately, so here's a video recorded by a user who reported the bug:

https://github.com/vslavik/poedit/assets/145881/e12382ca-6dd3-4dbb-a96a-9a7d5a856c03

vslavik:
The primary motivation for this PR was f370e9e8ea5bfbed8a6be0bd4e386ee2a3a4ddfe to fix https://github.com/vslavik/poedit/issues/849, and then I went down the rabbit hole a bit:

1. 37efaacba3dfdfe98520f5a2950ddb75a638b8cd — This is a no-brainer, prevents accidental menu clicks and makes wxOSX consistent with both other ports and native Cocoa context menus. The only question is whether to do it like this or limit oneself to the `wxDefaultPosition` case. I initially did the latter, but this seems more sensible and benefits legacy code that may pass explicit mouse coordinates.

2. f370e9e8ea5bfbed8a6be0bd4e386ee2a3a4ddfe — This fixes the weird behavior when using <kb>control</kbd>-click instead of right-click: there seems to be some corruption of internal state (AppKit, not wx) if `menuForEvent:` doesn't return a menu and we pop one up manually. (When I return a native menu, things behave normally. When I delay `PopupMenu` into next event loop iteration, i.e. after `menuForEvent:` returns, it doesn't help.) This code was there since the very beginning (~2009) and never meaningfully modified, so I don't think there was a particularly deep reason for using `menuForEvent:` [^1]

3. caffdd753a37cecd8be492157e49983a3c6b7f0a — While doing that, I noticed that the mouse position is never used... Apparently, `wxEVT_DATAVIEW_ITEM_CONTEXT_MENU` was always broken and didn't operate on the right-clicked item, but instead used whatever the current selection was. So I fixed that too. 

All of this should be cherry-picked into 3.2 too.

[^1]: In principle, it would be better to use `menuForEvent:` everywhere — see also https://github.com/wxWidgets/wxWidgets/issues/13010#issuecomment-1010846630 — but that would require a new API, e.g. addition of `virtual wxWindow::CreateContextMenu()` to be implementable.

### DIFF
--- a/src/osx/dataview_osx.cpp
+++ b/src/osx/dataview_osx.cpp
@@ -852,6 +852,21 @@ void wxDataViewCtrl::OnSize(wxSizeEvent& event)
   event.Skip();
 }
 
+void wxDataViewCtrl::OnContextMenu(wxContextMenuEvent& WXUNUSED(e))
+{
+    // get the item information;
+    // theoretically more than one ID can be returned but the event can only
+    // handle one item, therefore only the first item of the array is
+    // returned:
+    wxDataViewItem item;
+    wxDataViewItemArray selectedItems;
+    if (GetSelections(selectedItems) > 0)
+        item = selectedItems[0];
+
+    wxDataViewEvent event(wxEVT_DATAVIEW_ITEM_CONTEXT_MENU, this, item);
+    GetEventHandler()->ProcessEvent(event);
+}
+
 wxSize wxDataViewCtrl::DoGetBestSize() const
 {
     wxSize best = wxControl::DoGetBestSize();
@@ -864,6 +879,7 @@ wxIMPLEMENT_DYNAMIC_CLASS(wxDataViewCtrl,wxDataViewCtrlBase);
 
 wxBEGIN_EVENT_TABLE(wxDataViewCtrl,wxDataViewCtrlBase)
   EVT_SIZE(wxDataViewCtrl::OnSize)
+  EVT_CONTEXT_MENU(wxDataViewCtrl::OnContextMenu)
 wxEND_EVENT_TABLE()
 
 #endif // !wxHAS_GENERIC_DATAVIEWCTRL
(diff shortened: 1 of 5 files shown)


csomor:
@vslavik thanks a lot, looking forward :-)

vadz (changes requested):
Thanks! I don't have any comments about this, all the explanations make sense to me.

I believe the original code was written by Hartwig, so I'm cc'ing @hwiesmann in case you have any comments. Otherwise I'll merge (and backport) this soon.

Thanks again!

hwiesmann:
No comments!

src/osx/dataview_osx.cpp
csomor, line 1: I don't know about the expectations in the `wxEVT_DATAVIEW_ITEM_CONTEXT_MENU` Handler, but perhaps using the `wxDataViewEvent` constructor that also takes the `wxDataViewColumn` would allow to differentiate on the column as well if wanted?

csomor (approved):
I'm fine with it, added one possible change about the additional `wxDataViewColumn` info, this wasn't there before either in this implementation, but eg in the generic implementation it is filled when this event occurs.

vslavik:
> I'm fine with it, added one possible change about the additional wxDataViewColumn info,

Thanks!

That's trivial to do and worth doing. I pushed cd5150f2 that fixes this for both wxOSX and wxGTK.

csomor:
> > I'm fine with it, added one possible change about the additional wxDataViewColumn info,
> 
> Thanks!
> 
> That's trivial to do and worth doing. I pushed [42ecd8b](https://github.com/wxWidgets/wxWidgets/commit/42ecd8be4274501d3f7b7f14363a185797766b6a) that fixes this for both wxOSX and wxGTK.

perfect, thanks a lot, 

so @vadz please go ahead and merge when it's convenient for you

PR #24700, opened 2024-07-12
Don't hardcode number of processors in Linux build instructions

Blake-Madden:
Saying `$(nproc)` works fine.


lanurmi:
> Saying `$(nproc)` works fine.

It works fine (even on most non-Linux *nix nowadays), but on the other hand it also complicates the simple example.

vadz:
Yes, I'm not sure if it's worth it, we already mention `nproc` just above, after all.

Any other opinions?

Blake-Madden:
OK, we can just leave this as-is.

PR #24702, opened 2024-07-13
Update Bakefile instructions to mention Docker

vslavik:
Because installing Bakefile is increasingly more complicated due to (but not only) Python 2, I figured I can make the situation a little bit less bad by packaging Bakefile as a Docker container. If you already use Docker, as many do, it can be more convenient way of running it, especially for non-core contributors.

The containers are built automatically for both 0.2 and 1.x branches and are always up to date and are documented in bakefile readmes. 

This PR mentions this option in wx's contributor docs.

### DIFF
--- a/docs/contributing/how-to-add-files-to-build-system.md
+++ b/docs/contributing/how-to-add-files-to-build-system.md
@@ -16,6 +16,12 @@ the `bakefile_gen` tool. Run it from `$(wx)/build/bakefiles` directory and it wi
 regenerate all outdated makefiles. See `$(wx)/build/bakefiles/README` for more
 details.
 
+You can also run Bakefile from a Docker or Podman container and avoid the need
+to install it (expanding "$(wx)" to be the path to wx install):
+
+    docker run --rm -v $(wx):$(wx) -w `pwd`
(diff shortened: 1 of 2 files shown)


docs/contributing/how-to-add-files-to-build-system.md
```diff
@@ -16,6 +16,12 @@ the `bakefile_gen` tool. Run it from `$(wx)/build/bakefiles` directory and it wi
 regenerate all outdated makefiles. See `$(wx)/build/bakefiles/README` for more
 details.
 
+You can also run Bakefile from a Docker or Podman container and avoid the need
+to install it (expanding "$(wx)" to be the path to wx install):
+
+    docker run --rm -v $(wx):$(wx) -w `pwd`
```
vadz, line 22: Sorry, I don't understand the use of `-w` here: does the current working directory on the host really make sense as the working directory inside the container?

vadz (commented):
Thanks, I had no idea about this image existence, but it should indeed make it much simpler for the others to update the makefiles, so I'll merge it as soon as you confirm that `-w` option is really correct.

PR #24708, opened 2024-07-16
Fix and simplify check for "real" children in wxWindow

### LINKED ISSUE #24565: [macOS] wxHasRealChildren in wincmn.cpp incorrectly determines real children

vadz:
The old code was wrong because it only checked for the scrollbars, which appear in the list of children under Mac, but not for the "grow box", which is also there and should be similarly ignored when computing the best window size.

It was also inefficient because it used dynamic cast and iterated over the list of children twice, first in wxHasRealChildren() and then in DoGetBestSize() itself.

Fix this by removing wxHasRealChildren() entirely and using IsClientAreaChild() in DoGetBestSize() to check for the children to skip.

Replaces #24566.

Closes #24565.

### DIFF
--- a/src/common/wincmn.cpp
+++ b/src/common/wincmn.cpp
@@ -633,28 +633,6 @@ void wxWindowBase::FitInside()
     SetVirtualSize( GetBestVirtualSize() );
 }
 
-// On Mac, scrollbars are explicitly children.
-#if defined( __WXMAC__ ) && !defined(__WXUNIVERSAL__)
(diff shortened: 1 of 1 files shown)


vadz:
As there are no comments, I'm merging this, but please let me know if you (@hwiesmann) see anything wrong with it. TIA!

PR #24709, opened 2024-07-16
Add private header wrapping GDK headers inclusion

vadz:
This allows to have the code checking for the various backends in one place only and also ensures that we can disable -Wundef in gdkwayland.h just once instead of doing it in every place where it is included, because the changes of aba1c66575 (Suppress harmless -Wundef in Wayland headers, 2024-07-16) were insufficient to fix the FreeBSD CI build, as the same warning still happens elsewhere.

----

@paulcor We could also add this to the existing "wx/gtk/private/backend.h", as this header is always included when "wrapgdk.h" is, but it seemed a bit cleaner to keep them separate — please let me know if you disagree.


vadz:
I've merged this in order to fix the CI failures, but it's not too late to change something here if anybody has any comments — please let me know.

PR #24710, opened 2024-07-17
Slightly improve wxCollapsiblePane docs

PBfordev:
Improve the example code by (a) omitting the defaults when calling wxSizerFlags()::Border() and (b) calling wxSizer::Add() using
wxSizerFlags-based overload instead of the one using many individual parameters.

Also remove an extraneous "not" in the description of its Create() method.

If 8cd71703ad28adfcb857b1f35b029ae267fb6b82 is back-ported to 3.2, this could be as well.


vadz:
Thanks!

Issue #12486, opened 2010-09-18
Provide a DC class for getting text metrics only


wxtrac:
*Issue migrated from trac ticket # 12486*

**component:** GUI-all | **priority:** low | **keywords:** MemoryDC text-extent simple

#### 2010-09-18 13:30:44: elvisjohndowson (Elvis John Dowson) created the issue

___
Hi,
      I have a problem with wxWidgets/WxPython-2.8.81 and 2.8.11.0, when using it with the GNU Radio library, where calls to memory device contexts to get the text extents fail, like this:

$ python scopesink_gl.py

input rate 1000000.0  v_scale None  t_scale 3e-05
Traceback (most recent call last):
 File "/Users/elvis/Tool/gnuradio-next/gr-wxgui/src/python/plotter/plotter_base.py", line 187, in _on_paint
   for fcn in self._draw_fcns: fcn[1]()
 File "/Users/elvis/Tool/gnuradio-next/gr-wxgui/src/python/plotter/plotter_base.py", line 59, in draw
   self._draw()
 File "/Users/elvis/Tool/gnuradio-next/gr-wxgui/src/python/plotter/grid_plotter_base.py", line 204, in _draw_grid
   for tick in self._get_ticks(self.x_min, self.x_max, self.x_step, self.x_scalar)]
 File "/Users/elvis/Tool/gnuradio-next/gr-wxgui/src/python/plotter/grid_plotter_base.py", line 302, in _get_tick_label
   return gltext.Text(tick_str, font_size=TICK_TEXT_FONT_SIZE)
 File "/Users/elvis/Tool/gnuradio-next/gr-wxgui/src/python/plotter/gltext.py", line 340, in __init__
   self._initText()
 File "/Users/elvis/Tool/gnuradio-next/gr-wxgui/src/python/plotter/gltext.py", line 376, in _initText
   self._centered)
 File "/Users/elvis/Tool/gnuradio-next/gr-wxgui/src/python/plotter/gltext.py", line 73, in __init__
   self.createTexture()
 File "/Users/elvis/Tool/gnuradio-next/gr-wxgui/src/python/plotter/gltext.py", line 160, in createTexture
   ow, oh = dc.GetMultiLineTextExtent(self._text)[:2]
 File "//usr/local/lib/wxPython-unicode-2.8.11.0/lib/python2.6/site-packages/wx-2.8-mac-unicode/wx/_gdi.py", line 3774, in GetMultiLineTextExtent
[...]

PR #24715, opened 2024-07-21
Add `wxInfoDC`

vadz:
This is incomplete as there is no documentation yet, but most of the changes have been done: there is now a new `wxInfoDC` class (see #12486) which can be used when you only need to get information about the DC and so replaces all "valid" uses of `wxClientDC`, with the remaining ones known to be broken unless explicitly guarded by `wxClientDC::CanBeUsedForDrawing()` as in wxAUI code.

One thing which is not great is that we still have to use `wxClientDC` with `wxDCOverlay`, I'd like to change/improve the API here too, but I don't know how yet.

But for now please let me know if you have any ideas/suggestions/criticisms about `wxInfoDC` itself. The rationale for doing it in this way is in the first commit message.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -172,6 +172,11 @@ Changes in behaviour which may result in build errors
 - wxGTK wxDirButton::Create() doesn't have unused "wildcard" parameter any
   longer, please just remove it from your code if you used it.
 
+- Several functions now take wxReadOnlyDC instead of wxDC as argument. While
+  any code calling these functions keeps compiling and working, classes that
+  override these functions must be updated to use the new argument type too.
+  Simply replacing wxDC with wxReadOnlyDC should be sufficient.
+
 
 3.3.0: (released 2022-??-??)
 ----------------------------
--- a/docs/doxygen/overviews/dc.h
+++ b/docs/doxygen/overviews/dc.h
@@ -13,34 +13,55 @@ A wxDC is a @e device context onto which graphics and text can be drawn.
 The device context is intended to represent a number of output devices in a
 generic way, with the same API being used throughout.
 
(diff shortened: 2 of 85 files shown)


vadz:
For anybody wanting to review this, the most important commit here is the first one.

vadz:
As there were no comments, I've updated the docs and plan to merge this soon.

docs/changes.txt
```diff
@@ -172,6 +172,11 @@ Changes in behaviour which may result in build errors
 - wxGTK wxDirButton::Create() doesn't have unused "wildcard" parameter any
   longer, please just remove it from your code if you used it.
 
+- Several functions now take wxReadOnlyDC instead of wxDC as argument. While
```
catalinr, line 175: Some possible name alternatives, using other words instead of `ReadOnly`: `Accessor`, `Access`, `Probe`, `Inquiry`, `Examine`, `Inspect`

vadz: Some of those (e.g. `Inspect`) seem too close to `wxInfoDC`, which is something different, and none of them seems obviously better than `ReadOnly` IMHO. But let me know if anybody really loves one of those (or something else), we can still rename it before merging...

eranif (commented):
Changing the signature of the virtual methods of the various arts, breaks existing code that implemented their own art classes by marking the overrode methods as `override`

vadz:
Yes, this is why this is documented as breaking change in `docs/changes.txt`. I hope fixing this shouldn't be too hard and the only alternative of providing completely new set of functions (because having overloaded virtual functions with the same name has its own problems) seems rather unpalatable.

But if anybody has any better ideas, please let me know.

PR #24716, opened 2024-07-22
Add `wxOverlayDC`

vadz:
This simplifies using `wxOverlay` without much loss of generality in practice because we can't use neither `wxWindowDC` nor `wxScreenDC` (which are the only other classes with which `wxDCOverlay` can be used) portably.

Naming is a bit confusing, but I think the real problem is with `wxDCOverlay` (which we can't change) and `wxOverlayDC` looks like a natural name for this one, so I think we should keep it.

Please let me know if I'm missing something here — I might, because if I don't, then I don't understand why hadn't we done it like this a long time ago.

cc @AliKet, @paulcor, @csomor

### DIFF
--- a/include/wx/generic/private/drawresize.h
+++ b/include/wx/generic/private/drawresize.h
@@ -24,9 +24,8 @@ inline wxBitmap wxPaneCreateStippleBitmap()
 inline void
 wxDrawOverlayResizeHint(wxWindow* win, wxOverlay& overlay, const wxRect& rect)
 {
-    wxClientDC dc{win};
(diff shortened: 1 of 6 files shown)


AliKet:
Nice, I approve these changes, thanks.

And me too, I find `wxOverlayDC` more natural/simpler to use for drawing (just like  `wxWindowDC` and `wxClientDC`)

vadz:
As there have been no other comments (thanks for the review, Ali!), I'm merging this but it's still not to late to comment about this.

PR #24717, opened 2024-07-22
Fix iOS warnings

Blake-Madden:


src/osx/core/bitmap.cpp
```diff
@@ -1004,7 +1004,7 @@ wxBitmap wxBitmap::GetSubBitmap(const wxRect &rect) const
     if ( GetBitmapData()->m_bitmapMask )
     {
         wxMemoryBuffer maskbuf ;
-        int rowBytes = GetBestBytesPerRow( destwidth * kMaskBytesPerPixel );
+        int rowBytes = static_cast<int>(GetBestBytesPerRow( destwidth * kMaskBytesPerPixel ));
```
vadz, line 1007: It looks like we could avoid the cast by just using

```suggestion
        size_t rowBytes = GetBestBytesPerRow( destwidth * kMaskBytesPerPixel );
```

instead (here and below).

src/osx/core/bitmap.cpp
```diff
@@ -1543,23 +1543,23 @@ void wxMask::Init()
 
 void *wxMask::GetRawAccess() const
 {
-    wxCHECK_MSG( m_maskBitmap, nullptr , wxT("invalid mask") ) ;
+    wxCHECK_MSG( m_maskBitmap, nullptr , "invalid mask" ) ;
     return CGBitmapContextGetData(m_maskBitmap);
 }
 
 int wxMask::GetBytesPerRow() const
```
vadz, line 1550: This one should be changed to return `size_t`, it ends up being passed as `size_t` argument to e.g. `CGBitmapContextCreate()`, so it doesn't make much sense to cast to `int` when we need `size_t` anyhow.

vadz (commented):
Sorry, `int`/`size_t` confusion is a bit of a mess and we can't really fix it easily, but I think we can do at least slightly better than this patch using the suggestions below.

Please let me know if I'm missing something.

src/osx/core/bitmap.cpp
Blake-Madden: I'm OK with just removing the "Add casts to suppress implicit conversion warnings" commit. `GetBytesPerRow()` is used in a lot of places where it is assigned to `int` and I would rather not change the API and risk even more warnings.

Blake-Madden: I reverted the int/size_t stuff. The other changes are innocuous.

vadz:

PR #24718, opened 2024-07-22
Add RTF support to MSW's wxTextCtrl

Blake-Madden:


src/msw/textctrl.cpp
```diff
@@ -1114,6 +1114,115 @@ void wxTextCtrl::DoSetValue(const wxString& value, int flags)
...
+    MSG msg;
+    while ( ::PeekMessage(&msg, (HWND)0, 0, 0, PM_REMOVE) )
+    {
+        if (msg.message != WM_QUIT)
+        {
+            ::TranslateMessage(&msg);
+            ::DispatchMessage(&msg);
+        }
+    }
+
```
vadz, line 1134: Sorry, why do we need to do this here? This definitely needs a comment but I really hope we can avoid any message dispatching as this is very dangerous, nobody expects any callbacks to be invoked when calling `wxTextCtrl::GetRTFValue()`!

Also, we never did this while we had `wxRichEditStreamOut` (removed in 4519d8e08a (Remove wxUSE_UNICODE checks as they're always true now, 2022-10-27), see its parent commit for its last version), so I really hope it's just a leftover and can be removed entirely.

src/msw/textctrl.cpp
```diff
@@ -1114,6 +1114,115 @@ void wxTextCtrl::DoSetValue(const wxString& value, int flags)
     }
 }
 
+#if wxUSE_RICHEDIT
+
+DWORD wxCALLBACK MSWEditStreamOutCallback(DWORD_PTR dwCookie, LPBYTE pbBuff,
+    LONG cb, LONG* WXUNUSED(pcb))
```
vadz, line 1120: Don't we need to set `*pcb = cb` to indicate that we handled all the bytes?

src/msw/textctrl.cpp
```diff
@@ -1114,6 +1114,115 @@ void wxTextCtrl::DoSetValue(const wxString& value, int flags)
...
+        if (msg.message != WM_QUIT)
+        {
+            ::TranslateMessage(&msg);
+            ::DispatchMessage(&msg);
+        }
+    }
+
+    // write the text
+    std::string* psEntry = reinterpret_cast<std::string*>(dwCookie);
+    if ( psEntry != nullptr )
```
vadz, line 1137: Can it actually be null here? AFAICS it can't and then this test should be either replaced with a `wxCHECK` or removed entirely.

src/msw/textctrl.cpp
```diff
@@ -1114,6 +1114,115 @@ void wxTextCtrl::DoSetValue(const wxString& value, int flags)
...
+    return wxString(buffer.c_str(), *wxConvCurrent);
+}
+
+void wxTextCtrl::SetRTFValue(const wxString& val)
+{
+    if ( !IsRich() )
+    {
+        wxFAIL_MSG("RTF support is only available for rich controls!");
+        return;
+    }
```
vadz, line 1191: Minor, but this is typically written on a single line using `wxCHECK_RET()`.

src/msw/textctrl.cpp
```diff
@@ -1114,6 +1114,115 @@ void wxTextCtrl::DoSetValue(const wxString& value, int flags)
...
+
+    SETTEXTEX textInfo{ 0 };
+    textInfo.flags = GetRichVersion() > 1 ? (ST_DEFAULT | ST_UNICODE) : ST_DEFAULT;
```
vadz, line 1194: We never used rich edit v1 as it doesn't have Unicode support, so this test is always true. However MSDN says that `EM_SETTEXTEX` is in v3 only, so maybe we need to test for this? But see also the remark about using `EM_STREAMIN`.

interface/wx/textctrl.h
```diff
@@ -1659,8 +1659,8 @@ class wxTextCtrl : public wxControl,
         Returns @c true if text controls support reading and writing RTF (Rich
         Text Format).
 
-        Currently this function only returns true in wxOSX, which is the only
-        port implementing RTF support in wxTextCtrl.
+        This function only returns @true in wxOSX and wxMSW (if wxUSE_RICHEDIT is defined),
```
vadz, line 1662: We probably need to make this function non-static or at least add a non-static overload, that would check that the control uses `wxTE_RICH[2]`, as otherwise RTF-related functions still don't work.

vadz (commented):
Thanks, but I wonder why we don't use `EM_STREAMIN` for `SetRTFValue()` for consistency with `EM_STREAMOUT` in the other direction?

src/msw/textctrl.cpp
Blake-Madden: Yes, this was from Windows XP days when streaming into this control was substantially slower. I'll remove this block.

src/msw/textctrl.cpp
Blake-Madden: No, it couldn't. Just being overly pedantic. I'll remove that check.

src/msw/textctrl.cpp
Blake-Madden: Got it.

src/msw/textctrl.cpp
Blake-Madden: Indeed. I'll fix that.

src/msw/textctrl.cpp
Blake-Madden: OK, I removed that check.

From my investigation (and vague memory), RichEdit 4.1 was released in XP SP1. So v3 should always be available. Please correct me if I'm wrong.

interface/wx/textctrl.h
Blake-Madden: I changed it to non-static.

Blake-Madden:
> Thanks, but I wonder why we don't use `EM_STREAMIN` for `SetRTFValue()` for consistency with `EM_STREAMOUT` in the other direction?

I only use `EM_STREAMOUT` because the `EM_GETTEXTEX` message doesn't support returning RTF. So I have to use a `EDITSTREAMCALLBACK`, which doesn't seem all that elegant. But for setting the text, the `EM_SETTEXTEX` message handles RTF; since that is much more elegant than a `EM_STREAMIN` callback, I use that. It's not consistent, but (IMO) the more elegant solution if an option is preferred.

Thoughts?

Blake-Madden:
The Unix failure is from a WebView test; doesn't seem related to this PR.

vadz:
I've changed a couple of string conversions:

1. I think we need to use UTF-8 and not the current encoding, please correct me if I'm wrong.
2. Using `wc_str()` should work in all builds, while casting `wxString` shouldn't work when `wxNO_IMPLICIT_WXSTRING_CONV_TO_PTR` is defined.

If you're ok with these changes, I'll (squash) merge this soon.

Thanks again for your work!

Blake-Madden:
>     1. I think we need to use UTF-8 and not the current encoding, please correct me if I'm wrong.

I'm thinking this is wrong, because the `char*` buffer that the control is streaming out should be in the ANSI (current code page).

If you specify `(CP_UTF8 << 16) | SF_USECODEPAGE` in the flags, then it will be UTF-8 and the header will say `{\urt` (note the `u`). (please see https://learn.microsoft.com/en-us/windows/win32/controls/em-streamout). I don't do this, though.

By **not** doing that we should be receiving ANSI text. For example, the header that it returns will be something like `{\rtf1\ansi\ansicpg1252` instead of the UTF-8 `{\urt`.

I chose not to use `(CP_UTF8 << 16) | SF_USECODEPAGE` and receive ANSI because Wordpad and Libreoffice cannot load UTF-8 RTF (i.e., `{\urt`) files. I doubt other apps can handle this format, but I gave up testing at that point.

Interestingly, the control will encode extended ASCII characters, so I will get something like `clich\'E9s` when I stream out. Because of that, I can't say any particular encoding is actually being used (the `ansicpg1252` in the header is the only reference to any encoding, and that appears to not be used for anything).

>     2. Using `wc_str()` should work in all builds, while casting `wxString` shouldn't work when `wxNO_IMPLICIT_WXSTRING_CONV_TO_PTR` is defined.

OK, that makes sense.

vadz:
Hmm, what happens if you put in a character not representable in CP1252 into the control, how does it come out (if it comes out at all)?

Blake-Madden:
> Hmm, what happens if you put in a character not representable in CP1252 into the control, how does it come out (if it comes out at all)?

It will encode it in RTF syntax. For example, I copy some Japanese text into a control (I'm on an English system). What gets streamed out looks like `\u12399?\u20998?`, so it renders correctly in Wordpad. (Of interest is that "ansicpg1252" is in the header, not "Shift_JIS".) So, it appears to always stream as a 7-bit stream and encode all extended ASCII characters in RTF syntax.

The code page in the header (I presume) is a relic; by that I mean that the RICHEDIT control no longer uses it and relies on RTF syntax instead of code page conversions.

vadz:
And I guess you've already tried using `SF_UNICODE` and it doesn't work with `SF_RTF` (the docs are ambiguous on whether it can be used with `SF_TEXT` or whether it *only* can be used with it)?

BTW, why do we use `SFF_PLAINRTF`, does this change anything and, if so, what?

Blake-Madden:
> And I guess you've already tried using `SF_UNICODE` and it doesn't work with `SF_RTF` (the docs are ambiguous on whether it can be used with `SF_TEXT` or whether it _only_ can be used with it)?

Correct. `SF_UNICODE` only has an effect with `SF_TEXT`, which in turn will give you plain text instead of RTF. I imagine that's by design, as RTF readers would never expect UTF-16 content; `{rtf` means an ANSI content, `{urt` will be UTF-8 (which, from my testing, is poorly supported by other readers).

> BTW, why do we use `SFF_PLAINRTF`, does this change anything and, if so, what?

My interpretation of this is that `SF_RTF` can emit RTF syntax extensions beyond the RTF spec (which seems odd, since the spec is by Microsoft). In other words, RTF syntax that may not be compatible with other RTF readers or other platforms that use RTF (e.g., macOS). So `SFF_PLAINRTF`, which claims to "stream out only the keywords common to all languages," is a safe bet for interoperability.

I honestly have no idea what these "language-specific keywords" `SF_RTF` can cause are, but it's best IMO to avoid them if we want to produce macOS-compatible output.

vadz:
I see, thanks.

According to [Wikipedia](https://en.wikipedia.org/wiki/Rich_Text_Format) RTF can only have 7-bit characters, so maybe we should use `FromAscii()` instead?

Blake-Madden:
> According to [Wikipedia](https://en.wikipedia.org/wiki/Rich_Text_Format) RTF can only have 7-bit characters, so maybe we should use `FromAscii()` instead?

I'm a little nervous about the possible presence of non-ASCII characters causing UB in `FromAscii()`, but it does seem that MSDN states that the control should always stream out 7-bit text. And I've never seen it do otherwise. I guess if you're OK with it then I am too :)

vadz:
I think getting an assert (in debug build) if we have a non-ASCII character is better than silently converting it, although neither behaviour is optimal, of course. But I hesitate to scan the whole buffer for non-ASCII characters because it could, in principle, be rather big. 

Let's merge it like this (with `FromAscii()`) for now and maybe check later if we actually ever happen to see non-ASCII characters in RTF.

PR #24721, opened 2024-07-23
Correct shared library prefix for CYGWIN

carlo-bramini:
Under CYGWIN, shared libraries using POSIX layer must have "cyg" instead of "lib" as prefix.
For example, correct library name must be `cygwx_baseu-3.2-0.dll` instead of `libwx_baseu-3.2-0.dll`.

EDIT:
The fix has been made only to a CMake script because autotools already generate correct file names for shared libraries.


vadz:
Looks good to me, thanks! I'll merge this soon (to 3.2 too?) if there are no objections.

PR #24723, opened 2024-07-23
Rare segfault fixed while DnD on GTK3>=3.20: 'draw' signal must be di…

valid-ptr:
…sconnected from then 'icon window'

Before this PR: "draw" signal can be received after "GTKDisconnectDragSignals" method call.

### DIFF
--- a/src/gtk/dnd.cpp
+++ b/src/gtk/dnd.cpp
@@ -806,7 +806,7 @@ void wxDropSource::PrepareIcon( int action, GdkDragContext *context )
         cairo_region_destroy(region);
     }
 
-    g_signal_connect(widget, "draw", G_CALLBACK(draw_icon), icon);
+    g_signal_connect(m_iconWindow, "draw", G_CALLBACK(draw_icon), icon);
 
 #else // !__WXGTK3__
 
@@ -919,6 +919,12 @@ wxDragResult wxDropSource::DoDragDrop(int flags)
 
     g_signal_handlers_disconnect_by_func (m_iconWindow,
                                           (gpointer) gtk_dnd_window_configure_callback, this);
+
+#ifdef __WXGTK3__
+    g_signal_handlers_disconnect_matched (m_iconWindow,
+                                          G_SIGNAL_MATCH_FUNC, 0, 0,  NULL, (gpointer) draw_icon, NULL);
+#endif
+
     g_object_unref(m_iconWindow);
     m_iconWindow = NULL;
 


src/gtk/dnd.cpp
```diff
@@ -919,6 +919,12 @@ wxDragResult wxDropSource::DoDragDrop(int flags)
 
     g_signal_handlers_disconnect_by_func (m_iconWindow,
                                           (gpointer) gtk_dnd_window_configure_callback, this);
+
+#ifdef __WXGTK3__
+    g_signal_handlers_disconnect_matched (m_iconWindow,
```
vadz, line 924: This is minor, but why can't we use `g_signal_handlers_disconnect_by_func` here for consistency?

src/gtk/dnd.cpp
```diff
@@ -806,7 +806,7 @@ void wxDropSource::PrepareIcon( int action, GdkDragContext *context )
         cairo_region_destroy(region);
     }
 
-    g_signal_connect(widget, "draw", G_CALLBACK(draw_icon), icon);
+    g_signal_connect(m_iconWindow, "draw", G_CALLBACK(draw_icon), icon);
```
vadz, line 809: I find it a bit suspicious that we connect to "draw" on the parent of the drawing area widget and not the widget itself now, are you sure this is correct?

vadz (commented):
Thanks for the fix, but while I understand your explanation for the addition of the disconnect call, I don't really see it for changing the widget in the connect one, could you please explain this?

I'd also appreciate a confirmation from @paulcor before merging this.

paulcor:
Isn't this the problem fixed by 5c4e3e385a? Which could just be backported to 3.2?

valid-ptr:
https://github.com/wxWidgets/wxWidgets/commit/5c4e3e385aff147de3bd18f46ba98a770b6da801 can be backported to 3.2 as well as this PR can be used in master-branch instead of https://github.com/wxWidgets/wxWidgets/commit/5c4e3e385aff147de3bd18f46ba98a770b6da801. There is just just less of code in this PR.

I used **m_iconWindow** instead of **widget**  for 'draw' signal connection for simple disconnection code.
I can't find visual difference while using **m_iconWindow** instead of **widget** , everything works as expected.

src/gtk/dnd.cpp
valid-ptr: We can't pass data (the latest argument) (It's not a member of class),
So I've decided to use `g_signal_handlers_disconnect_matched` instead.

vadz:
Thanks for the explanations, I'll backport the existing fix, I don't see any reason to make more changes in master if the bug has already been fixed there.

PR #24727, opened 2024-07-24
Remove concerning warranty comments from library READMEs

Blake-Madden:
I assume this was meant for humor, but comments about wxWidgets crashing your harddrive or destroying your monitor a bit unprofessional (IMO).


tierra:
Think I'd probably just rip those paragraphs out altogether personally. It's only really necessary in the license itself.

Blake-Madden:
> Think I'd probably just rip those paragraphs out altogether personally. It's only really necessary in the license itself.

Sounds good to me. I Just removed them.

PR #24731, opened 2024-07-26
Fix macOS and GTK broken links in platform docs

Blake-Madden:
Cocoa link was broken, replaced with the latest overview topic on Apple developer.

Some GTK links were broken, others were getting redirected. I replaced all of them with the main GTK link, which includes links to everything else like Pango and GDK. This way, only one link needs to be maintained, but still provides everything.

### DIFF
--- a/docs/doxygen/mainpages/platdetails.h
+++ b/docs/doxygen/mainpages/platdetails.h
@@ -132,12 +132,7 @@ In such case (or when you want to e.g. write a port-specific patch) it can be
 necessary to use the underlying toolkit API directly:
 
 - wxMSW port uses Win32 API: see MSDN docs at https://docs.microsoft.com/en-us/windows/win32/controls/window-controls
-- wxGTK port uses GTK and other lower-level libraries; see
-  - GTK docs at https://developer.gnome.org/gtk/
-  - GDK docs at https://library.gnome.org/devel/gdk/
-  - GLib docs at https://library.gnome.org/devel/glib/
-  - GObject docs at https://library.gnome.org/devel/gobject/
-  - Pango docs at https://library.gnome.org/devel/pango/
-- wxOSX port uses the Cocoa API: see Cocoa docs at https://developer.apple.com/cocoa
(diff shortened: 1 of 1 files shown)


docs/doxygen/mainpages/platdetails.h
```diff
@@ -132,12 +132,7 @@ In such case (or when you want to e.g. write a port-specific patch) it can be
...
 - wxMSW port uses Win32 API: see MSDN docs at https://docs.microsoft.com/en-us/windows/win32/controls/window-controls
-- wxGTK port uses GTK and other lower-level libraries; see
-  - GTK docs at https://developer.gnome.org/gtk/
-  - GDK docs at https://library.gnome.org/devel/gdk/
-  - GLib docs at https://library.gnome.org/devel/glib/
-  - GObject docs at https://library.gnome.org/devel/gobject/
-  - Pango docs at https://library.gnome.org/devel/pango/
-- wxOSX port uses the Cocoa API: see Cocoa docs at https://developer.apple.com/cocoa
+- wxGTK port uses GTK and other lower-level libraries; see https://docs.gtk.org/
+- wxOSX port uses the Cocoa API: see Cocoa docs at https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CocoaFundamentals/WhatIsCocoa/WhatIsCocoa.html
```
vadz, line 136: Not sure if it's worth linking to "archive" docs. Maybe just link to https://developer.apple.com/ instead?

Blake-Madden: I was wondering about that too. The archive page is a fantastic reference for Cocoa, with all the overviews easy to find. The main developer page is quite a bit more generic and you really have to spend some time exploring to find stuff on something like `NSTextView`. And it's very stingy with example code (IMO).

But I'm OK with either option.

vadz: OK, let's keep the archive link for as long as it still works, thanks.

PR #24740, opened 2024-07-29
Fix text size of dragged `wxDataViewCtrl` item in high DPI

### LINKED ISSUE #24648: Text in dragged item of `wxDataViewCtrl` is too small at high DPI
This issue can also be seen in the dataview sample.
At a DPI scaling of 200%, the text in the dragged item is too small.

dataview-sample_dpi-scale-200

- wxWidgets version  you use: 3.2.5 and master (d1506f40c67453ad5ad6799dc6e71a7e71ff56f4)

taler21:
Fixes #24648.

### DIFF
--- a/src/generic/datavgen.cpp
+++ b/src/generic/datavgen.cpp
@@ -2468,6 +2468,9 @@ wxBitmap wxDataViewMainWindow::CreateItemBitmap( unsigned int row, int &indent )
     width -= indent;
 
     wxBitmap bitmap( width, height );
+#ifdef __WXMSW__
+    bitmap.SetScaleFactor( GetDPIScaleFactor() );
+#endif // __WXMSW__
     wxMemoryDC dc( bitmap );
     dc.SetFont( GetFont() );
     dc.SetPen( *wxBLACK_PEN );
--- a/src/generic/datavgen.cpp
+++ b/src/generic/datavgen.cpp
@@ -1715,8 +1715,8 @@ public:
                                         wxFRAME_FLOAT_ON_PARENT |
                                         wxFRAME_NO_TASKBAR |
                                         wxNO_BORDER );
-            new wxBitmapCanvas( m_hint, ib, ib.GetSize() );
-            m_hint->SetClientSize(ib.GetSize());
+            new wxBitmapCanvas( m_hint, ib, ib.GetLogicalSize() );
+            m_hint->SetClientSize(ib.GetLogicalSize());
             m_hint->SetTransparent(128);
             m_hint->Show();
         }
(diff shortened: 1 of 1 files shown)


src/generic/datavgen.cpp
```diff
@@ -2468,6 +2468,9 @@ wxBitmap wxDataViewMainWindow::CreateItemBitmap( unsigned int row, int &indent )
     width -= indent;
 
     wxBitmap bitmap( width, height );
+#ifdef __WXMSW__
+    bitmap.SetScaleFactor( GetDPIScaleFactor() );
+#endif // __WXMSW__
```
vadz, line 2473: Could you please check that the following code, which should be equivalent, but avoids platform-specific code, work too?

```suggestion
    wxBitmap bitmap;
    bitmap.CreateWithLogicalSize( width, height, GetDPIScaleFactor() );
```

vadz (commented):
Thanks for fixing this, but I'd prefer to do it without `#ifdef`s if possible, could you please check if the alternative below works?

TIA!

taler21:
Okay, here now the working solution without `#ifdef`.

You are right, doing it this way is definitely better.
However, `#ifdef`s must be used in the backport because `wxBitmap::CreateWithLogicalSize()` does not exist in 3.2.

taler21:
The backport of this fix looks like this:

```patch
diff --git "a/src/generic/datavgen.cpp" "b/src/generic/datavgen.cpp"
index 5f26739e47..08286f21ed 100644
--- "a/src/generic/datavgen.cpp"
+++ "b/src/generic/datavgen.cpp"
@@ -1714,8 +1714,8 @@ public:
                                         wxFRAME_FLOAT_ON_PARENT |
                                         wxFRAME_NO_TASKBAR |
                                         wxNO_BORDER );
-            new wxBitmapCanvas( m_hint, ib, ib.GetSize() );
-            m_hint->SetClientSize(ib.GetSize());
+            new wxBitmapCanvas( m_hint, ib, ib.GetLogicalSize() );
+            m_hint->SetClientSize(ib.GetLogicalSize());
             m_hint->SetTransparent(128);
             m_hint->Show();
         }
@@ -2465,7 +2465,13 @@ wxBitmap wxDataViewMainWindow::CreateItemBitmap( unsigned int row, int &indent )
     }
     width -= indent;
 
-    wxBitmap bitmap( width, height );
+    wxBitmap bitmap;
+#ifdef wxHAS_DPI_INDEPENDENT_PIXELS
+    bitmap.CreateWithDIPSize( width, height, GetDPIScaleFactor() );
+#else
+    bitmap.Create( width, height );
+    bitmap.SetScaleFactor( GetDPIScaleFactor() );
+#endif
     wxMemoryDC dc( bitmap );
     dc.SetFont( GetFont() );
     dc.SetPen( *wxBLACK_PEN );

```

vadz:
Thanks a lot for the update, as you say it definitely looks much nicer like this! I'll merge it a.s.a.p. and push to the backports PR too.

PR #24741, opened 2024-07-29
Generic DVC: wxDataViewChoiceByIndexRenderer's 'SetValue' and 'Create…

valid-ptr:
…EditorCtrl' with wxNOT_FOUND value in arguments fix. Caused segfault, but should be empty wxChoice instead.


vadz:
Thanks, crashing is definitely wrong and between asserting and failing and accepting `-1` as meaning "empty value", the latter is probably more useful, so I'll apply this and document this behaviour.

PR #24742, opened 2024-07-30
Fix `wxPropertyGrid` row height in high DPI

### LINKED ISSUE #24607: Height of properties in a `wxPropertyGrid` is too large at high DPI
This issue can also be seen in the propgrid sample.

#### To Reproduce:

1. Start the propgrid sample.
2. Click on 'File' and 'Run minimal sample' in the menu.

At a DPI scaling of 200%, the properties in the `wxPropertyGridManager` have the correct height, but not those in the `wxPropertyGrid`.
propGrid-vs-propGridManager-dpi-scale-200

taler21:
Fixes #24607.

### DIFF
--- a/src/propgrid/propgrid.cpp
+++ b/src/propgrid/propgrid.cpp
@@ -482,7 +482,7 @@ void wxPropertyGrid::Init2()
     m_cursorSizeWE = wxCursor(wxCURSOR_SIZEWE);
 
     // adjust bitmap icon y position so they are centered
-    m_vspacing = FromDIP(wxPG_DEFAULT_VSPACING);
+    m_vspacing = wxPG_DEFAULT_VSPACING;
 
     CalculateFontAndBitmapStuff( m_vspacing );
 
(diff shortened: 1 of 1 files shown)


vadz:
Sorry, could you please explain why is this the right thing to do? AFAICS this should scale with the DPI and the changes of a1c3fa0468 (Fix wxPropertyGrid row height on DPI change, 2019-01-13) done by @MaartenBent look correct to me...

taler21:
First `m_vspacing` isn't a pixel size

https://github.com/wxWidgets/wxWidgets/blob/3737245c5195d84f0c788e650c6ad7992eec03d7/include/wx/propgrid/propgrid.h#L1060-L1065

and second, it should not be reset when moving the window to another screen with a different DPI scaling.
A value set via `SetVerticalSpacing()` should be respected.

vadz:
> First `m_vspacing` isn't a pixel size

Oops, I had no idea about this. It's a really weird API but the PR is obviously correct in the light of this, sorry for missing it.

I'll apply this soon, thanks!

taler21:

PR #24743, opened 2024-07-30
Fix borders and spaces in `wxWizard` at high DPI

taler21:
With wxWidgets-3.2.5, for example, the borders of the `wxWizard` buttons are too small at high DPI.

At 200% DPI scaling under Windows 10

- with wxWidgets-3.2.5:
  wxWizard-button-row-dpi-200-wx325

- with this PR:
  wxWizard-button-row-dpi-200-PR


vadz:

vadz:
BTW, using `wxSizerFlags::Border()` here would be even better as it would avoid hardcoding magical constants everywhere, if anyone would be willing to do it and test...

PR #24579, opened 2024-05-29
wxrc: Faster compilation of generated C++ code, leaner generated machine code.

DoctorNoobingstoneIPresume:
This pull request replaces commit number 6 in https://github.com/wxWidgets/wxWidgets/pull/24437#issuecomment-2077394039...

PR #24744, opened 2024-07-30
Fix `wxrc` generated C++ code linker error

sodevel:
Commit 2d3f89109fdc620b856d6aa8f789c2986d4b8228 turned the `XRC_ADD_FILE()` macro in the generated C++ code in a global function, this results in a linker error when multiple of such generated files are linked into a single executable. Give this function internal linkage instead to prevent this error.


vadz:
Oops, thanks for noticing and fixing this!

cc #24579.

PR #24748, opened 2024-08-03
Fix incorrect handling of ligatures in wxStyledTextCtrl on Linux/GTK

### LINKED ISSUE #13168: Incorrect handling of ligatures in wxStyledTextCtrl
*Issue migrated from trac ticket # 13168*

**component:** wxGTK | **priority:** normal | **keywords:** wxSTC wxStyledTextCtrl wxGTK text

#### 2011-04-27 15:56:20: nielsm (Niels Martin Hansen) created the issue

___
I have only observed this happening on wxGTK. I have observed the issue with wxGTK 2.8.10 (from Ubuntu 10.04) and with trunk 67627.

If the control is using a font that has OpenType ligatures, entering a character sequence that produces a ligature will offset the displayed cursor position.

For example, entering the word "office" with the DejaVu Sans font being used, the three letters "ffi" together form a ligature rendered with a single glyph. Using the cursor keys to move over this word, the displayed cursor will move straight from before "ffi" to straight after, as if it was a single letter. Typing when the cursor visually is straight after "ffi" will insert the new input after the first "f" and break the ligature. Moving the cursor further along the line, similar offset errors are seen.

As mentioned, this depends wholly on the features of the font selected and using e.g. DejaVu Sans Mono will not show this issue, as it does not have ligatures.

This issue should show with any program that uses the STC control, but here is a short sample:

```
#!cpp
#include <wx/app.h>
#include <wx/frame.h>
#include <wx/sizer.h>
#include <wx/stc/stc.h>

class TestApp : public wxApp {
public:
	bool OnInit();
};

DECLARE_APP(TestApp)

IMPLEMENT_APP(TestApp)

bool TestApp::OnInit()
{
	wxFrame *frm = new wxFrame(0, -1, _T("STC test app"));
	wxStyledTextCtrl *stc = new wxStyledTextCtrl(frm, -1);
	wxSizer *box = new wxBoxSizer(wxVERTICAL);
	box->Add(stc, 1, wxEXPAND);
	frm->SetSizerAndFit(box);
	frm->Show();
}
```

Funami580:
Fixes #13168.

Simple program to try out the change:
```cpp
#include <wx/wx.h>
#include <wx/stc/stc.h>

class TestApp : public wxApp {
public:
    virtual bool OnInit();
};

wxIMPLEMENT_APP(TestApp);

bool TestApp::OnInit() {
    wxFrame *frm = new wxFrame(nullptr, wxID_ANY, wxT("STC test app"));
    wxStyledTextCtrl *stc = new wxStyledTextCtrl(frm, wxID_ANY);
    wxSizer *box = new wxBoxSizer(wxVERTICAL);
    box->Add(stc, 1, wxEXPAND);
    frm->SetSizerAndFit(box);
    frm->Show();
    return true;
}
```
Compile with:
```
g++ main.cpp `../wxWidgets/buildgtk/wx-config --cxxflags --libs all` -o test
```

The sequence `fi` triggered the bug for me, due to it being treated as a ligature. I can't guarantee that this change won't introduce new bugs, but I haven't noticed any so far.

### DIFF
--- a/src/generic/graphicc.cpp
+++ b/src/generic/graphicc.cpp
@@ -3086,14 +3086,44 @@ void wxCairoContext::GetPartialTextExtents(const wxString& text, wxArrayDouble& 
 
         ApplyFont(layout, font);
         pango_layout_set_text(layout, data, data.length());
-        PangoLayoutIter* iter = pango_layout_get_iter(layout);
-        PangoRectangle rect;
-        do {
-            pango_layout_iter_get_cluster_extents(iter, nullptr, &rect);
-            w += rect.width;
+
+        // Number of Unicode characters in the the text
+        int num_chars = pango_layout_get_character_count(layout);
+
+        if (num_chars) {
+            // Get attributes for each character
+            const PangoLogAttr* attrs = pango_layout_get_log_attrs_readonly(layout, nullptr);
+
+            PangoLayoutIter* iter = pango_layout_get_iter(layout);
+            PangoRectangle rect;
+            int char_index = 0;
+            int byte_index = 0;
+            while (pango_layout_iter_next_char(iter)) {
+                // We need to know the last byte_index for later, so always get it
+                byte_index = pango_layout_iter_get_index(iter);
+                if (!attrs[++char_index].is_cursor_position)
+                    continue;
+                pango_layout_index_to_pos(layout, byte_index, &rect);
+                if (rect.width < 0) {
+                    w = rect.x + rect.width;
+                } else {
+                    w = rect.x;
+                }
+                widths.Add(PANGO_PIXELS(w));
+            }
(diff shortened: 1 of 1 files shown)


vadz:
Thanks! I don't fully understand this code and, notably, what does the negative width mean here, but it does fix the bug and so I'm merging this with just some minor changes (formatting and added a sanity check).

If no problems are discovered with this in master, it should be backported to 3.2 too.

Thanks again!

Funami580:
The negative width means that the directionality of the grapheme is right-to-left, so pos.x is on the right side of the grapheme (https://docs.gtk.org/Pango/method.Layout.index_to_pos.html). If there are any other questions, feel free to ask!

PR #24752, opened 2024-08-11
Improve WebView documentation

PBfordev:
@vadz, the chromium backend docs has "... currently requires CEF version 116". Does it mean the implementation is locked to that version or should it be "... version 116 or newer"?

### DIFF
--- a/interface/wx/webview.h
+++ b/interface/wx/webview.h
@@ -176,7 +176,7 @@ enum wxWebViewIE_EmulationLevel
 
     Sample JavaScript opening a new window:
     @code
-        window.open("http://www.wxwidgets.org", "newWindow", "width=400,height=400");
+        window.open("https://www.wxwidgets.org", "newWindow", "width=400,height=400");
     @endcode
 
     Sample C++ code handling a new window request:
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for the updates, will merge soon!

As for CEF, 116 was the version tested at the time of writing this but I had used 121 since then and I think many others should work too — but unfortunately it's hard to guarantee it, as sometimes there are backwards-incompatible changes in CEF.

PBfordev:
> As for CEF, 116 was the version tested at the time of writing this but I had used 121 since then and I think many others should 

Then perhaps it could be better to instead write something like the usual "... macOS. The backend was developed against CEF version 116 and should work with newer versions, but of course, the compatibility with future CEF versions cannot be guaranteed."

PBfordev:
I think that `wxWebView::GetNativeBackend()` may need even more information.

To my surprise, when using the Edge backend, it returns `nullptr` not only after being created with `wxWebView::New()`,  but also in the `wxEVT_WEBVIEW_CREATED` handler (and `CallAfter()` does not help)? 

I cannot test other (non-obsolete) backends to see how it works there but perhaps this may not be documentation but code issue? In the edge backend the event is sent from `wxWebViewEdge::Create()`, but maybe it would be better to somehow send it only  from `wxWebViewEdgeImpl::OnWebViewCreated()` (`wxEVT_WEBVIEW_CREATED` was added only in 3.3)?

Again, this behavior maybe intended but perhaps it should be documented how to earliest access the native backends (I wanted to call `ICoreWebView2Settings::put_IsBuiltInErrorPageEnabled(FALSE);` before loading any page).

 @TcT2k, please, any thoughts on the above?

PBfordev:
The last push added information that EVT_WEBVIEW_CREATED is available only since 3.3 and slightly changed the wording for version availability for other events.

PR #24753, opened 2024-08-11
Appearance fixes for wxMSW in dark mode

vadz:
The first two commits fix the wrong background being used for the `wxPanel` children (see #24663), while the latter fixes appearance of disabled `wxStaticText` once again (and hopefully for the last time...).


Blake-Madden:
Looks great under Windows 11. Merci!

Dark:

dark

Light:

light

Disabled static text under dark:

statictext

PR #24754, opened 2024-08-11
Fixes for checkboxes in `wxPropGrid` in high DPI

vadz:
This partially reuses and partially replaces #24671 and fixes the same #24650 and #24651 as it did.


taler21:
Many thanks for these fixes! Please also backport them to 3.2.

taler21:
Is there a reason I am missing that prevents these fixes from being backported?

vadz:
No, they can be backported, I've actually already done it in my local branch, just didn't push it out yet.

PR #24755, opened 2024-08-12
Modernize sizers in samples, make more DPI aware

Blake-Madden:
Use `wxSizerFlags`. Make sizers DPI aware.

### DIFF
--- a/samples/artprov/artbrows.cpp
+++ b/samples/artprov/artbrows.cpp
@@ -171,21 +171,21 @@ wxArtBrowserDialog::wxArtBrowserDialog(wxWindow *parent)
         m_sizes->Append( wxString::Format("%d x %d", *p, *p ) );
     }
     m_sizes->SetSelection(0);
-    subsub->Add(m_sizes, 0, wxALL, 4);
+    subsub->Add(m_sizes, 0, wxALL, FromDIP(4));
 
     m_text = new wxStaticText(this, wxID_ANY, "Size: 333x333");
-    subsub->Add(m_text, 0, wxALL, 4);
+    subsub->Add(m_text, 0, wxALL, FromDIP(4));
 
     m_canvas = new wxStaticBitmap(this, wxID_ANY, wxBitmap(null_xpm));
     subsub->Add(m_canvas);
     subsub->Add(256, 256);
-    subsizer->Add(subsub, 1, wxLEFT, 4 );
+    subsizer->Add(subsub, 1, wxLEFT, FromDIP(4));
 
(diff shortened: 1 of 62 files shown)


samples/aui/auidemo.cpp
```diff
@@ -290,118 +290,118 @@ class SettingsPanel : public wxPanel
     {
         //wxBoxSizer* vert = new wxBoxSizer(wxVERTICAL);
 
-        //vert->Add(FromDIP(1), FromDIP(1), 1, wxEXPAND);
+        //vert->Add(FromDIP(1), FromDIP(1), wxSizerFlags(1).Expand());
```
vadz, line 293: The use of `FromDIP()` for expandable spacers is actually quite useless... And using `Expand()` for spacers doesn't make much sense neither. I.e. I'd write this as

```suggestion
        //vert->Add(FromDIP(1), FromDIP(1), wxSizerFlags(1).Expand());
```
```suggestion
        //vert->AddStretchSpacer(1);
```

(and, of course, the same for all not commented out lines too).

vadz (commented):
Thanks, I agree that adding `FromDIP()` calls is an improvement (although not necessarily for spacers), but getting rid of the numbers entirely would be even nicer, I wonder if it would be possible to do this instead?

samples/aui/auidemo.cpp
Blake-Madden: Got it.

samples/artprov/artbrows.cpp
```diff
@@ -150,15 +150,15 @@ wxArtBrowserDialog::wxArtBrowserDialog(wxWindow *parent)
 
     subsizer = new wxBoxSizer(wxHORIZONTAL);
     subsizer->Add(new wxStaticText(this, wxID_ANY, "Client:"), 0, wxALIGN_CENTER_VERTICAL);
-    subsizer->Add(choice, 1, wxLEFT, 5);
-    sizer->Add(subsizer, 0, wxALL | wxEXPAND, 10);
+    subsizer->Add(choice, 1, wxLEFT, FromDIP(5));
```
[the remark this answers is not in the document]
Blake-Madden: OK, I fixed a bunch of these. I can revisit all this again in a later PR, but I need a wxSizer break for now ;)

vadz:
Just FYI, I'd like to write a script updating all this code automatically as there are still quite a few of calls using old sizer API after this PR, but it will take some time.

Blake-Madden:
I'm envisioning a fantastic regex to find all of these :)

I work on a tool that checks for deprecated things like `wxOVERRIDE`; I could maybe add some sort of check for usage of legacy sizer arguments.

Issue #23853, opened 2023-09-06
Interest in a wxTextCtrl::Find() pull request


Blake-Madden:
I’ve been working on text searching functionality for the text control (using the native APIs) and I was wondering if there would be interest in my submitting this as a PR. (I would be a bit involved, so that’s why I ask.)  I currently have code for:

- macOS
- MSW
- GTK

It supports:

- Searching backwards or forwards
- Case sensitive or insensitive
- Whole-word searching

If you think this would be useful, what would the preferred signature be. I’m thinking one of these:

```
long wxTextCtrl::Find(const wxString& textToFind, const wxFindReplaceFlags searchFlags) const

long wxTextCtrl::Find(const wxString& textToFind,  const bool searchDown,
                      const bool matchWholeWord, const bool caseSensitiveSearch) const
```

Note that all the fields for `wxFindReplaceFlags` are supported.

If you search backwards, it will always stop at the start of the document. If you search down and it doesn’t find the text, it will ask if you would like to wrap around and search from the start of the text. (That seems to be the behavior for various editors).

If the sting is not found, `wxNOT_FOUND` is returned.

Question: if it finds the text, should it automatically select it and ensure that it is visible?  That is what I am doing now, but perhaps this should only find the text and it should be the caller’s responsibility to select and scroll to it?

Thanks for any feedback,
Blake

vadz:
It would be definitely nice to have something like this, thanks!

Concerning the API, it should be, generally speaking, as flexible as possible because you shouldn't assume that people will use this only for interactive selection (so there should be a way to avoid any questions about wrapping around) or for selecting just the next match (so it shouldn't get selected automatically).

To facilitate further API development the new function should definitely take an enum and not a set of individual bools, but this enum probably should be something new rather than `wxFindReplaceFlags` in order to allow having flags controlling wraparound in it, for example.

We also need to return something more than just a `long`: at the very least, it should return the range of matched text, but it would also be useful to indicate whether the wraparound occurred, for example.

So my ideal API would be something like

```cpp
struct wxTextSearch
{
    wxTextSearch(const wxString& text);

    wxTextSearch& IgnoreCase() { ...; return *this; }

    ... other similar functions ...
};

struct wxTextSearchResult
{
    explicit operator bool() const { return start != wxNOT_FOUND; }

    int start = 0;
    int end = 0;
};

class wxTextCtrl
{
    ...
    wxTextSearchResult SearchText(const wxTextSearch& search) const;
};
[...]

PR #24756, opened 2024-08-12
Add wxTextCtrl::SearchText() for MSW

Blake-Madden:
Supports case matching, whole-word matching, and search direction options.

The defaults are to **not** match by case or whole word.

Implements #23853

If this API looks agreeable, I can work on PRs for macOS and GTK later.

### DIFF
--- a/include/wx/textctrl.h
+++ b/include/wx/textctrl.h
@@ -630,7 +630,7 @@ struct wxTextSearch
         return *this;
     }
 
-    wxTextSearch& MatchCase(const bool matchCase = true)
+    wxTextSearch& MatchCase(bool matchCase = true)
     {
         m_matchCase = matchCase;
         return *this;
--- a/include/wx/textctrl.h
+++ b/include/wx/textctrl.h
@@ -642,7 +642,7 @@ struct wxTextSearch
         return *this;
     }
 
-    wxTextSearch& Direction(const Direction direction)
+    wxTextSearch& Direction(const wxTextSearch::Direction direction)
     {
         m_direction = direction;
         return *this;
@@ -658,7 +658,7 @@ struct wxTextSearch
     long                  m_startingPosition = -1;
     bool                  m_matchCase = false;
     bool                  m_wholeWord = false;
-    Direction m_direction = Direction::Down;
+    wxTextSearch::Direction m_direction = wxTextSearch::Direction::Down;
 };
 
 // results from a search operation
--- a/include/wx/textctrl.h
+++ b/include/wx/textctrl.h
@@ -669,7 +669,7 @@ struct wxTextSearchResult
     explicit wxTextSearchResult(long startPos, long endPos) :
         m_start(startPos), m_end(endPos) {}
     wxTextSearchResult() : m_start(wxNOT_FOUND), m_end(wxNOT_FOUND) {}
-
+wxTextSearchResult() = default;
     long m_start = wxNOT_FOUND;
     long m_end = wxNOT_FOUND;
 };
--- a/include/wx/textctrl.h
+++ b/include/wx/textctrl.h
@@ -656,7 +656,7 @@ struct wxTextSearch
 
     wxString              m_searchValue;
     long                  m_startingPosition = -1;
-    bool                  m_matchCase = true;
+    bool                  m_matchCase = false;
     bool                  m_wholeWord = false;
     wxTextSearchDirection m_direction = wxTextSearchDirection::Down;
 };
--- a/include/wx/textctrl.h
+++ b/include/wx/textctrl.h
@@ -668,6 +668,7 @@ struct wxTextSearchResult
     explicit operator bool() const { return m_start != wxNOT_FOUND; }
     explicit wxTextSearchResult(long startPos, long endPos) :
         m_start(startPos), m_end(endPos) {}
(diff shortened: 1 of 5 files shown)


include/wx/textctrl.h
```diff
@@ -608,6 +608,69 @@ class WXDLLIMPEXP_CORE wxTextAttr
...
+{
+    Down,
+    Up
+};
+
+// search options
+// --------------
+struct wxTextSearch
+{
+    wxTextSearch(const wxString& text) : m_searchValue(text) {}
```
vadz, line 625: Minor, but the ctor should be explicit we probably also want to provide the default ctor, especially as we have `SearchValue()`, so:

```suggestion
    explicit wxTextSearch(const wxString& text = wxString{}) : m_searchValue(text) {}
```

include/wx/textctrl.h
```diff
@@ -608,6 +608,69 @@ class WXDLLIMPEXP_CORE wxTextAttr
...
+
+    wxTextSearch& SearchValue(const wxString& value)
+    {
+        m_searchValue = value;
+        return *this;
+    }
+
+    wxTextSearch& MatchCase(const bool matchCase = true)
```
vadz, line 633: We don't use [useless top level const](https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html):

```suggestion
    wxTextSearch& MatchCase(bool matchCase = true)
```

(also below)

include/wx/textctrl.h
```diff
@@ -608,6 +608,69 @@ class WXDLLIMPEXP_CORE wxTextAttr
...
 
+// ----------------------------------------------------------------------------
+// Search features for wxTextCtrl
+// ----------------------------------------------------------------------------
+
+enum class wxTextSearchDirection
+{
+    Down,
+    Up
+};
```
vadz, line 619: Maybe move it inside `wxTextSearch`? This would mean writing `wxTextSearch::Direction::Down` instead of `wxTextSearchDirection::Down`, i.e. 2 extra characters, but it seems more logical to have everything search-related inside this single struct.

src/msw/textctrl.cpp
```diff
@@ -1198,6 +1198,56 @@ void wxTextCtrl::SetRTFValue(const wxString& val)
...
 }
 
+wxTextSearchResult wxTextCtrl::SearchText(const wxTextSearch& search) const
+{
+    // set up the flags
+    WPARAM flags = 0;
+    if (search.m_direction == wxTextSearchDirection::Down)
+    {
+        flags |= FR_DOWN;
+    }
```
vadz, line 1208: One of big benefits of using an `enum` rather than just a `bool` is to ensure that we handle all the case, so could we please write this like

```suggestion
    switch ( search.m_direction )
    {
        case wxTextSearchDirection::Down:
            flags |= FR_DOWN;
            break;
            
        case wxTextSearchDirection::Up:
            // Nothing to do this is (surprisingly) the default.
            break;
    }
```

instead please?

src/msw/textctrl.cpp
```diff
@@ -1198,6 +1198,56 @@ void wxTextCtrl::SetRTFValue(const wxString& val)
...
+        // if going down, then start from 0; otherwise, start from end
+        (search.m_direction == wxTextSearchDirection::Down) ? 0 : GetLastPosition();
+    if (search.m_direction == wxTextSearchDirection::Down)
+    {
+        // go to the end of the text
+        findText.chrg.cpMax = -1;
+    }
+    else
+    {
+        // if at the beginning of the window, then we can't search up
```
vadz, line 1231: Is this worth explicitly testing for? I.e. won't the native control just return "not found" anyhow?

Not a big deal, of course, but it seems strange to test for this but not that we're at the end when searching forward.

vadz (commented):
Looks nice, thanks!

Just a few nitpicks, but otherwise we can merge this already — or would you prefer to add GTK/macOS implementations to this PR too?

src/msw/textctrl.cpp
Blake-Madden: Got it, removed that check.

include/wx/textctrl.h
Blake-Madden: Done.

Blake-Madden:
Alright, I think everything is good now. We can (squash and) merge this now. I'll work on GTK+ support next week in another PR.

vadz:
Sorry for returning to this so lately, but there seems to be a real problem here in the AppVeyor CI build:

```
src\msw\textctrl.cpp(1249): error C2440: 'initializing': cannot convert from 'initializer list' to 'wxTextSearchResult'
```

Could you please check/fix this? TIA!

include/wx/textctrl.h
```diff
@@ -668,6 +668,7 @@ struct wxTextSearchResult
     explicit operator bool() const { return m_start != wxNOT_FOUND; }
     explicit wxTextSearchResult(long startPos, long endPos) :
         m_start(startPos), m_end(endPos) {}
+    wxTextSearchResult() : m_start(wxNOT_FOUND), m_end(wxNOT_FOUND) {}
```
AliKet, line 672:
```suggestion
wxTextSearchResult() = default;
```

vadz:
Merged in the commit above, thanks!

PR #24760, opened 2024-08-14
Add synchronous `wxWebRequest`

vadz:
This is ready for merging as is, but I'm probably going to fix a few other minor things in this branch yet. Still, any comments about the API and testing would be already very appreciated.

In addition to the main topic, there are a couple of other changes here:

- Allow more flexibility when disabling TLS checks.

TODO:

- [x] WinHTTP backend
- [x] `NSURLSession` backend (although the current implementation is suboptimal).
- [x] Documentation

### DIFF
--- a/include/wx/msw/private/webrequest_winhttp.h
+++ b/include/wx/msw/private/webrequest_winhttp.h
@@ -101,12 +101,12 @@ private:
 
     wxWebSessionWinHTTP& m_sessionImpl;
     wxString m_url;
-    HINTERNET m_connect;
-    HINTERNET m_request;
+    HINTERNET m_connect = nullptr;
+    HINTERNET m_request = nullptr;
     wxObjectDataPtr<wxWebResponseWinHTTP> m_response;
     wxObjectDataPtr<wxWebAuthChallengeWinHTTP> m_authChallenge;
     wxMemoryBuffer m_dataWriteBuffer;
-    wxFileOffset m_dataWritten;
+    wxFileOffset m_dataWritten = 0;
 
     void SendRequest();
 
@@ -154,7 +154,7 @@ public:
     }
 
 private:
-    HINTERNET m_handle;
+    HINTERNET m_handle = nullptr;
 
     bool Open();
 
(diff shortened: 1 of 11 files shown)


vadz:
I have more changes planned (while working on this, I've found several other problems...), but I think this might be in a reasonably good shape by now.

As proposed in my [post to wx-dev](https://groups.google.com/g/wx-dev/c/WHIMLdQMzAE/m/zkq0wTn9BgAJ) I've removed support for using `wxWebAuthChallenge` in the synchronous case, as it just seems gratuitously complicated when we can simply put the username and password in the URL. The main drawback of this is that we can't support any other auth schemes, not using passwords, but then we didn't support them before anyhow, so it's not really a loss.

Also, this uncovered the fact that using username and password worked for (async) `wxWebRequest` in libcurl and NSURLSession but _not_ WinHTTP backend, which clearly seems wrong, so I've fixed this too and if you know the username/password in advance you can just put them in the URL instead of bothering with providing them later.

vadz:
Huh, I spoke too soon, `WebRequest::Sync::Auth::Digest` test fails under Ubuntu 18.04 even though it succeeds under later versions. This is pretty weird, especially as the same version of go-httpbin is used everywhere. Must be due to some difference in system libraries, maybe libcurl under 18.04 is too old and doesn't support credentials for the digest authentication in the URL?

Blake-Madden:
`wxWebRequestSync::Execute` states "This function blocks for potentially long time and so must not be used from the main thread."

To be truly synchronous then, one would need to launch this in a `std::jthread` and join it. But then that would defeat the point if `wxWebRequestSync::Execute` gets stuck somehow.

Perhaps a timeout parameter (in milliseconds) could be added to `wxWebRequestSync::Execute`?

vadz:
There is no way to specify a timeout with blocking API (such as `curl_easy_perform()` or various `WinHttpXXX()` functions in synchronous mode), so I don't see how can this be done. We'd need a third mode, using async API in a pseudo-blocking way, without dispatching the events, in order to implement this.

vadz:
Just in case this went unnoticed, I've updated the description after pushing the latest changes.

vadz:
I plan to merge this soon and open a separate PR with proxy-related changes.

PR #24762, opened 2024-08-18
Add proxy support to wxWebRequest

vadz:
This one is supposed to be applied on top of #24760 and adds support for specifying the proxy to use, for both sync and async requests.

There are no new unit tests as deploying a proxy just for testing this would be too cumbersome but local testing using tinyproxy and microsocks (supported by libcurl backend only) works.

Any comments about the API would be especially welcome. Note that I'm aware that it could be extended (e.g. by allowing to configure the exceptions list), but it doesn't pretend to be exhaustive, just to cater for the most common use cases.


vadz:
I'd like to merge this too now that the base branch has been merged.

Please let me know if anybody has any comments!

PR #24764, opened 2024-08-19
allow using wxCHK_UNDETERMINED with wxUpdateUIEvent

wsu-cb:
Note that the diffs here are better viewed with --ignore-space-change.

I have added use of the new `wxUpdateUIEvent::Set3StateValue()` to the toolbar, ribbon, and aui samples.  I'm not convinced all three samples need to demonstrate this feature, so if you want to trim the samples, feel free to do so.

### DIFF
--- a/include/wx/event.h
+++ b/include/wx/event.h
@@ -2926,7 +2926,7 @@ public:
     wxUpdateUIEvent(wxWindowID commandId = 0)
         : wxCommandEvent(wxEVT_UPDATE_UI, commandId)
     {
-        m_checked =
+        m_3checked = wxCHK_UNCHECKED;
         m_enabled =
         m_shown =
         m_setEnabled =
@@ -2934,10 +2934,11 @@ public:
         m_setText =
         m_setChecked = false;
         m_isCheckable = true;
+        m_is3State = false;
     }
     wxUpdateUIEvent(const wxUpdateUIEvent& event)
         : wxCommandEvent(event),
-          m_checked(event.m_checked),
+          m_3checked(event.m_3checked),
           m_enabled(event.m_enabled),
           m_shown(event.m_shown),
           m_setEnabled(event.m_setEnabled),
@@ -2945,10 +2946,12 @@ public:
           m_setText(event.m_setText),
           m_setChecked(event.m_setChecked),
           m_isCheckable(event.m_isCheckable),
+          m_is3State(event.m_is3State),
           m_text(event.m_text)
     { }
 
-    bool GetChecked() const { return m_checked; }
+    bool GetChecked() const { return Get3StateValue() != wxCHK_UNCHECKED; }
+    wxCheckBoxState Get3StateValue() const { return m_3checked; }
     bool GetEnabled() const { return m_enabled; }
     bool GetShown() const { return m_shown; }
     wxString GetText() const { return m_text; }
@@ -2957,14 +2960,18 @@ public:
     bool GetSetEnabled() const { return m_setEnabled; }
     bool GetSetShown() const { return m_setShown; }
 
-    void Check(bool check) { m_checked = check; m_setChecked = true; }
+    void Check(bool check) { Set3StateValue(check ? wxCHK_CHECKED : wxCHK_UNCHECKED); }
+    void Set3StateValue(wxCheckBoxState check);
     void Enable(bool enable) { m_enabled = enable; m_setEnabled = true; }
     void Show(bool show) { m_shown = show; m_setShown = true; }
     void SetText(const wxString& text) { m_text = text; m_setText = true; }
 
     // A flag saying if the item can be checked. True by default.
     bool IsCheckable() const { return m_isCheckable; }
-    void DisallowCheck() { m_isCheckable = false; }
+    void DisallowCheck();
+    // A flag saying if the item can be wxCHK_UNDETERMINED. False by default.
+    bool Is3State() const { return m_is3State; }
+    void Allow3rdState(bool b = true);
 
     // Sets the interval between updates in milliseconds.
     // Set to -1 to disable updates, or to 0 to update as frequently as possible.
@@ -2991,7 +2998,7 @@ public:
     virtual wxEvent *Clone() const override { return new wxUpdateUIEvent(*this); }
 
 protected:
-    bool          m_checked;
+    wxCheckBoxState m_3checked;
     bool          m_enabled;
     bool          m_shown;
     bool          m_setEnabled;
@@ -2999,6 +3006,7 @@ protected:
     bool          m_setText;
     bool          m_setChecked;
     bool          m_isCheckable;
+    bool          m_is3State;
     wxString      m_text;
 #if wxUSE_LONGLONG
     static wxLongLong       sm_lastUpdate;
--- a/include/wx/window.h
+++ b/include/wx/window.h
@@ -1341,6 +1341,11 @@ public:
     // send wxUpdateUIEvents to this window, and children if recurse is true
     virtual void UpdateWindowUI(long flags = wxUPDATE_UI_NONE);
 
+    // do the window-specific processing before processing the update event
+    // (mainly for deciding whether wxUpdateUIEvent::Is3State() is set)
+    virtual void DoPrepareUpdateWindowUI(wxUpdateUIEvent& event) const
+        { event.Allow3rdState(false); }
+
     // do the window-specific processing after processing the update event
     virtual void DoUpdateWindowUI(wxUpdateUIEvent& event) ;
 
--- a/interface/wx/window.h
+++ b/interface/wx/window.h
@@ -4099,6 +4099,22 @@ public:
     */
     virtual void UpdateWindowUI(long flags = wxUPDATE_UI_NONE);
 
+    /**
+        When UpdateWindowUI() runs, it creates instances of
+        wxUpdateUIEvent.  Those instances may vary depending on the
+        window that the wxUpdateUIEvent will control.  For example, a
+        wxCheckBox with wxCHK_3STATE should enable
+        wxUpdateUIEvent::Is3State(), but most other windows should not.
+        This function can be overridden to perform the
+        window-specific initializations, such as enabling setting the
+        checkable state.
+
+        @see wxUpdateUIEvent, UpdateWindowUI()
+
+        @since 3.3.0
+    */
+    virtual void DoPrepareUpdateWindowUI(wxUpdateUIEvent& event) const;
+
     ///@}
 
 
--- a/samples/aui/auidemo.cpp
+++ b/samples/aui/auidemo.cpp
@@ -32,6 +32,7 @@
 #include "wx/msgdlg.h"
 #include "wx/textdlg.h"
 #include "wx/stattext.h"
+#include "wx/checkbox.h"
 
 #include "wx/aui/aui.h"
 #include "../sample.xpm"
@@ -103,6 +104,8 @@ class MyFrame : public wxFrame
         ID_NotebookAlignBottom,
         ID_NotebookNewTab,
         ID_NotebookDeleteTab,
+        ID_3CHECK,
+        ID_UI_2CHECK_UPDATED,
 
         ID_SampleItem,
 
@@ -167,6 +170,8 @@ private:
     void OnNotebookDeleteTab(wxCommandEvent& evt);
 
     void OnPaneClose(wxAuiManagerEvent& evt);
+
+    void OnCheckboxUpdateUI(wxUpdateUIEvent& evt);
 
 private:
 
@@ -660,6 +665,7 @@ wxBEGIN_EVENT_TABLE(MyFrame, wxFrame)
     EVT_AUINOTEBOOK_PAGE_CLOSE(wxID_ANY, MyFrame::OnNotebookPageClose)
     EVT_AUINOTEBOOK_PAGE_CLOSED(wxID_ANY, MyFrame::OnNotebookPageClosed)
     EVT_AUINOTEBOOK_PAGE_CHANGING(wxID_ANY, MyFrame::OnNotebookPageChanging)
+    EVT_UPDATE_UI(ID_UI_2CHECK_UPDATED, MyFrame::OnCheckboxUpdateUI)
 wxEND_EVENT_TABLE()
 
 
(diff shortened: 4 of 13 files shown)


include/wx/event.h
```diff
@@ -2957,14 +2960,32 @@ class WXDLLIMPEXP_CORE wxUpdateUIEvent : public wxCommandEvent
...
     void Enable(bool enable) { m_enabled = enable; m_setEnabled = true; }
     void Show(bool show) { m_shown = show; m_setShown = true; }
     void SetText(const wxString& text) { m_text = text; m_setText = true; }
 
     // A flag saying if the item can be checked. True by default.
     bool IsCheckable() const { return m_isCheckable; }
-    void DisallowCheck() { m_isCheckable = false; }
+    void DisallowCheck()
+    {
+        wxCHECK_RET(!GetSetChecked(), "SetCheck() or Set3StateValue() has already been called");
```
imciner2, line 2979: Why do we need to add this check now when it hasn't been a problem in the past? `DisallowCheck` should only be called in the control code anyway, which should always be before any setting of the check state.

include/wx/event.h
```diff
@@ -2957,14 +2960,32 @@ class WXDLLIMPEXP_CORE wxUpdateUIEvent : public wxCommandEvent
...
 
-    void Check(bool check) { m_checked = check; m_setChecked = true; }
+    void Check(bool check) { Set3StateValue(check ? wxCHK_CHECKED : wxCHK_UNCHECKED); }
+    void Set3StateValue(wxCheckBoxState check)
+    {
+        wxCHECK_RET(IsCheckable(), "IsCheckable() not enabled");
+        wxCHECK_RET(Is3State() || check != wxCHK_UNDETERMINED, "Is3State() not enabled");
+        m_3checked = check;
+        m_setChecked = true;
+    }
```
vadz, line 2970: I don't think we should allow calling `Set3StateValue()`, even with the 2 standard values, for non-3-state checkbox, this is just confusing. I.e. I'd do something like this:

```suggestion
    void Check(bool check)
    {
        wxASSERT_MSG(IsCheckable(), "Shouldn't be called if non-checkable");
        
        m_3checked = check ? wxCHK_CHECKED : wxCHK_UNCHECKED);
        m_setChecked = true;
    }
    void Set3StateValue(wxCheckBoxState check)
    {
        wxASSERT_MSG(Is3State(), "Shouldn't be called if non-3-state-checkable");
        
        m_3checked = check;
        m_setChecked = true;
    }
```

include/wx/event.h
vadz: FWIW I don't mind adding this assert, although I'd probably use `wxASSERT_MSG()` here too as nothing too bad happens if we don't return. It could be useful for catching logical errors in the library itself.

OTOH we probably should make all these functions with asserts in them non-inline (i.e. move them to `event.cpp`) to avoid having these strings in all call sites.

samples/aui/auidemo.cpp
```diff
@@ -1356,6 +1376,17 @@ void MyFrame::OnPaneClose(wxAuiManagerEvent& evt)
     }
 }
 
+// copy state from user-controllable checkbox to disabled checkbox
+void MyFrame::On3CheckUpdateUI(wxUpdateUIEvent& evt)
+{
+#if wxUSE_CHECKBOX
+    wxWindow* tb4 = m_mgr.GetPane("tb4").window;
+    wxWindow* wnd = tb4->GetWindowChild(ID_3CHECK);
+    wxCheckBox* src = wxDynamicCast(wnd, wxCheckBox);
```
vadz, line 1385: I'd use

```suggestion
    wxCheckBox* src = wxCheckCast<wxCheckBox>(wnd);
```

to at least get an assert before crashing if it ever turns out not to be a checkbox.

(here and elsewhere)

src/aui/auibar.cpp
```diff
@@ -2207,60 +2207,67 @@ void wxAuiToolBar::DoIdleUpdate()
         if (item.m_toolId == -1)
             continue;
 
-        wxUpdateUIEvent evt(item.m_toolId);
-        evt.SetEventObject(this);
+        if (item.GetKind() != wxITEM_CONTROL)
```
vadz, line 2210: Do the checks for `item.m_window` inside this check still make sense?

I don't quite follow all the ramifications of this change, e.g. is the window ID always guaranteed to be the same as `item.m_toolId` in this case? If not, we might be sending different events now.

Perhaps it would be safer to add, and call, `item.Is3State()` here instead?

src/common/tbarbase.cpp
```diff
@@ -789,24 +789,31 @@ void wxToolBarBase::UpdateWindowUI(long flags)
         if ( tool->IsSeparator() )
             continue;
 
-        int toolid = tool->GetId();
+        if ( !tool->IsControl() )
```
vadz, line 792: Here too it's not totally obvious this doesn't change the existing behaviour...

Maybe it does it in a good way, e.g. it looks like setting text should work now whereas it didn't before, but it's not really obvious.

vadz (commented):
Thanks, the idea of adding `CreateUpdateUIEvent()` is interesting, but I'm not sure if it's the best way to do it as it's rather inconsistent with the existing use of events in the code.

As we already have `DoUpdateWindowUI()` used for class-specific post-processing, I think it would make sense to add a similar virtual `DoPrepareUpdateWindowUI(wxUpdateUIEvent& event)` that would perform additional initialization of the event (name is, as always, subject to discussion). What do you think?

include/wx/event.h
wsu-cb: > FWIW I don't mind adding this assert, although I'd probably use `wxASSERT_MSG()` here too as nothing too bad happens if we don't return. It could be useful for catching logical errors in the library itself.
> 
OK

> OTOH we probably should make all these functions with asserts in them non-inline (i.e. move them to `event.cpp`) to avoid having these strings in all call sites.

Good catch!

include/wx/event.h
wsu-cb: If this were completely new code, I probably would have done it the way you suggest.  However, `wxCheckBoxBase::Set3StateValue()` explicitly allows itself to be called with the two standard values for `!Is3State()`, so I followed its design.  Do you still want me to change the `wxUpdateUIEvent` code?

samples/aui/auidemo.cpp
wsu-cb: OK

include/wx/event.h
vadz: Sorry, I forgot that `wxCheckBox` function worked like this (and I still wonder why it does, but it's too late to change this, of course...). I agree that it makes sense to make the behaviour of the new function consistent with the existing one, so please disregard this comment, sorry again.

src/aui/auibar.cpp
wsu-cb: 1)  The code for `wxAuiToolBar::AddControl()` implies to me that `item.m_toolId` ought to always be equal to `wxControl::GetId()`
2)  The code for `wxAuiToolBar::AddControl()` and `wxAuiToolBar::AddTool()` imply to me that `item.m_kind` ought to be `wx_ITEM_CONTROL` if and only if `AddControl()` (not `AddTool()`) was used, and also that `item.m_window` ought to be `!= nullptr` if and only if `AddControlI()` was used
3)  Unfortunately, the `wxAuitToolBarItem::SetId()`, `wxAuitToolBarItem::SetKind()`, and `wxAuitToolBarItem::SetWindow()` functions allow these ought-to-be-invariants to be violated.
4)  The unmodified code only respects `wxUpdateUIEvent::Check()` for `wxITEM_CHECK` and `wxITEM_RADIO`.  In other words, the unmodified code doesn't properly handle `wxCHK_2STATE wxCheckBox`es because they are `wxITEM_CONTROL`.
5)  The unmodified code doesn't support `wxUpdateUIEvent::Show()` or `wxUpdateUIEvent::SetText()` for `wxITEM_CONTROL` even though it seems to me that those operations should be supported.

I think the code is currently not in a very good state, but I don't see a way to completely fix it without the breaking change of eliminating `wxAuitToolBarItem::SetId()`, `wxAuitToolBarItem::SetKind()`, and `wxAuitToolBarItem::SetWindow()`.  My opinion is to either entirely ignore those functions (which is what I did since I was only trying to make `wxCheckBox` work), or else add `wxCHECK()`s that would treat changing from a `wxITEM_CONTROL` to a non-`wxITEM_CONTROL` (or vice versa) an error, and also that would enforce `item.m_window != nullptr` if and only `item.m_kind == wxITEM_CONTROL`.

I prefer that all `wxControl`s, not just `wxCHK_3STATE wxCheckBox`es, use `wxWindow::UpdateWindowUI()`, but I don't get to make the final decision.  If this message still hasn't convinced you, I will work through the `item.Is3State()` concept.  On the other hand, if I have convinced you, then I should, as you implied, remove the `item.m_window` checks.

src/common/tbarbase.cpp
wsu-cb: My response here is basically the same as for auibar.cpp, except that here the `m_item` members are protected, rather than effectively public, so there is less opportunity for anyone to have been modifying with them.

src/aui/auibar.cpp
wsu-cb: I also hadn't realized the control `wxUpdateUIEvent` handling was so restricted until after I implemented `wxUpdateUIEvent::Set3StateValue()` and realized it was having any effect.

I don't actually need the additional `wxUpdateUIEvent` functionality for other controls, but knowing these limitations exist is irritating me, so I would prefer a general solution.  Also, it seems weird to me to make a change adding `item.Is3State()` just to submit another PR to remove it again when making the general replacement.  If you would prefer the general `wxITEM_CONTROL` changes as a separate PR from this one (which is reasonable), I am actually inclined to submit that change BEFORE this one so that we don't wind up with this very briefly existing special case for `item.Is3State()`.  Is that OK with you?

vadz: Sure, 2 PRs would be fine, but please also feel free to submit it even as part of the same PR but as a separate commit _preceding_ the rest of the changes — I think this should make things tidier, and this commit might be backported to 3.2 too, if desired.

Thanks again!

wsu-cb: > Sure, 2 PRs would be fine, but please also feel free to submit it even as part of the same PR but as a separate commit _preceding_ the rest of the changes — I think this should make things tidier, and this commit might be backported to 3.2 too, if desired.

OK, I will re-sequence the commits in this PR.

wsu-cb:
> As we already have `DoUpdateWindowUI()` used for class-specific post-processing, I think it would make sense to add a similar virtual `DoPrepareUpdateWindowUI(wxUpdateUIEvent& event)` that would perform additional initialization of the event (name is, as always, subject to discussion). What do you think?

OK

include/wx/event.h
wsu-cb: changes force-pushed

samples/aui/auidemo.cpp
wsu-cb: changes force-pushed

src/aui/auibar.cpp
wsu-cb: changes force-pushed

wsu-cb:
> > As we already have `DoUpdateWindowUI()` used for class-specific post-processing, I think it would make sense to add a similar virtual `DoPrepareUpdateWindowUI(wxUpdateUIEvent& event)` that would perform additional initialization of the event (name is, as always, subject to discussion). What do you think?
> 
> OK

changes force-pushed

wsu-cb:
I have force-pushed a new version of this PR.  Currently:
- The first commit now adds some code to the samples to demonstrate that `wxToolBar` and `wxAuiToolBar` don't handle `wxControl` objects as tools very well.  The second commit is a fix for those problems (as discussed at https://github.com/wxWidgets/wxWidgets/pull/24764#discussion_r1725982935).
- The next three commits implement and document the addition of `wxCHK_UNDETERMINED` support to `wxUpdateUIEvent`.
- The sixth (and currently final) commit adds 3-state `wxCheckBox` objects to the samples to demonstrate the new feature.

I don't know whether all the `wxCheckBox` additions to the samples actually need to be merged to master.  Does anyone have suggestions for what the final form of this PR should be?

include/wx/window.h
```diff
@@ -1341,6 +1341,11 @@ class WXDLLIMPEXP_CORE wxWindowBase : public wxEvtHandler
     // send wxUpdateUIEvents to this window, and children if recurse is true
     virtual void UpdateWindowUI(long flags = wxUPDATE_UI_NONE);
 
+    // do the window-specific processing before processing the update event
+    // (mainly for deciding whether wxUpdateUIEvent::Is3State() is set)
+    virtual void DoPrepareUpdateWindowUI(wxUpdateUIEvent& event) const
+        { event.Allow3rdState(false); }
```
vadz, line 1347: Why bother doing this when it's the default anyhow?

interface/wx/window.h
```diff
@@ -4099,6 +4099,19 @@ class wxWindow : public wxEvtHandler
     */
     virtual void UpdateWindowUI(long flags = wxUPDATE_UI_NONE);
 
+    /**
+        When UpdateWindowUI() runs, it creates instances of
+        wxUpdateUIEvent.  Those instances should vary depending on the
```
vadz, line 4104: I don't think we should be so categoric:
```suggestion
        wxUpdateUIEvent.  Those instances may vary depending on the
```

interface/wx/window.h
```diff
@@ -4099,6 +4099,19 @@ class wxWindow : public wxEvtHandler
...
+        window that the wxUpdateUIEvent will control.  For example, a
+        wxCheckBox with wxCHK_3STATE should enable
+        wxUpdateUIEvent::Is3State(), but most other windows should not.
+
```
vadz, line 4108: This seems a bit unfinished to me, I'd add

```suggestion
This function can be overridden to perform window-specific initializations,
such as enabling setting the checkable state.
```

vadz (commented):
Thanks, I think this is good to merge now.

Concerning the samples, I don't mind adding this to them, especially as you did it. Ideal would be to add a unit test for the new functionality to check that it works in all ports, but this risks being a bit more complicated.

wsu-cb:
> Ideal would be to add a unit test for the new functionality to check that it works in all ports, but this risks being a bit more complicated.

I have no experience in unit testing something like this.  I know there exists `wxUIActionSimulator`, but I have no experience with it.

include/wx/window.h
wsu-cb: I have no good reason for this.  I just liked being explicit about the new state, especially since `IsCheckable()` defaults true, and this defaults false.  I can remove it if you prefer.

(As a separate statement, I think it's strange that `IsCheckable()` defaults true, but changing it would be a breaking change that doesn't seem justified.)

interface/wx/window.h
wsu-cb: OK

interface/wx/window.h
wsu-cb: OK

paulcor:
Please stop closing and re-opening this. Every time you do that, everyone on the `wx-dev` mailing list gets an email.

wsu-cb:
> Please stop closing and re-opening this. Every time you do that, everyone on the `wx-dev` mailing list gets an email.

I apologize for the noise, but I have needed to force the build-checks to re-run.  Do you know a different way to re-run them?

PR #24768, opened 2024-08-20
textfiletest: fix skip condition

cboltz:
... so that condition and skip message match.

As a side effect, it also prevents a test failure if /sys/power/state doesn't exist.


vadz:
Oops, thanks for fixing this, will merge soon.

PR #24769, opened 2024-08-21
Add `wxWebRequest::SetBaseURL()` and some other minor improvements

vadz:
I think it's rather convenient to have this function, which is present in most other web client APIs that I know of, so here it is.


vadz:
No comments, so merging.

PR #24770, opened 2024-08-21
Allow customizing exit code in case of `OnInit()` failure

vadz:
The important commit here is the last one.

Any comments are welcome, as always.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -27,6 +27,10 @@ Changes in behaviour not resulting in compilation errors
 - wxMSW doesn't support versions of Microsoft Windows before Windows 7. If you
   need Windows XP or Vista support, please use wxWidgets 3.2.
 
+- Default exit code in case of fatal error is now consistently 255, including
+  when using MSVC for which it was 127 previously. Use wxApp::SetErrorExitCode()
+  and wxApp::SetFatalErrorExitCode() to change this to old value if needed.
+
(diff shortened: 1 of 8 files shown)


vadz:
Thinking more about this, I believe we really ought to change all `return -1` to `return 255` to make the exit codes consistent between all platforms/CRT variants, exiting the program with either 255 or 127 depending on the compiler it is compiled with (MinGW vs MSVC) really doesn't make sense.

The problem here is that we really need to hardcode this, as at least in some places we don't have a `wxTheApp` to use (e.g. if we exit because we failed to create it). But we could provide `static void wxApp::SetFatalExitCode(int code)`, I guess...

So the choices are:

1. Just replace all hardcoded `-1` with the equally hardcoded `255`.
2. Replace `return -1` with `return wxApp::GetFatalExitCode()` which defaults to `255`.
3. Same as above, but let the default be `-1` for compatibility and let programs that want it to call `SetFatalExitCode()` explicitly.

Thoughts/opinions?

vadz:
I've implemented (2) and (3) with the latter being in a separate (last) commit in this branch. We could either apply it or not, but I think we should apply it as explained in the commit message.

Please let me know if anybody disagrees/sees something better to do. TIA!

vadz:
We have to decide something, so if there are no other opinions, I'm going to merge the entire branch, i.e. put in place the solution (3) from above.

PR #24775, opened 2024-08-22
Improve documentation for convenience input functions and wxDirDialog

PBfordev:
Regarding the `wxDirDialog` commit:
1. Does `wxRESIZE_BORDER` still have to be used at all there?
2. The description of `wxDD_DIR_MUST_EXIST` is IMO still inaccurate. E.g., AFAICT, on Windows presence/absence of this flag has zero effect on the (`IFileDialog`-based) dialog appearance.

### DIFF
--- a/docs/doxygen/overviews/commondialogs.h
+++ b/docs/doxygen/overviews/commondialogs.h
@@ -182,8 +182,8 @@ description for each, such as:
 
 Classes: wxDirDialog
 
-This dialog shows a directory selector dialog, allowing the user to select a
-single directory.
+This dialog shows a directory selector dialog, allowing the user to select
+one or multiple directories.
 
 
 
--- a/interface/wx/colordlg.h
+++ b/interface/wx/colordlg.h
@@ -158,8 +158,8 @@ wxEventType wxEVT_COLOUR_CHANGED;
 
     @header{wx/colordlg.h}
 */
-wxColour wxGetColourFromUser(wxWindow* parent,
-                             const wxColour& colInit,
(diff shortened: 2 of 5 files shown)


vadz:
> Regarding the `wxDirDialog` commit:
> 
>     1. Does `wxRESIZE_BORDER` still have to be used at all there?

Apparently it triggers the use of `BIF_NEWDIALOGSTYLE` in wxMSW. I think it's also used by the generic implementation.
 
>     2. The description of `wxDD_DIR_MUST_EXIST` is IMO still inaccurate. E.g., AFAICT, on Windows presence/absence of this flag has zero effect on the (`IFileDialog`-based) dialog appearance.

Do you mean that `BIF_NEWDIALOGSTYLE` (or its absence) has no effect any more?

PBfordev:
To be honest, I think the user should never see the old style dialog (`::SHBrowseForFolder()`). Am I missing something (aside from catastrophic `IFileDialog` API failure)?

vadz:
> To be honest, I think the user should never see the old style dialog (`::SHBrowseForFolder()`). Am I missing something (aside from catastrophic `IFileDialog` API failure)?

Well, there is the check done in `wxIFileDialog::CanBeUsedWithAnOwner()`... I don't think this has been fixed, but I didn't test.

PBfordev:
> Well, there is the check done in `wxIFileDialog::CanBeUsedWithAnOwner()`... I don't think this has been fixed, but I didn't test.

Sorry, forgot about that one; however, that should still be an edge case.

Issue #19053, opened 2021-01-21
Static text in bold is truncated on the wizard page (GTK3)


wxtrac:
*Issue migrated from trac ticket # 19053*

**component:** wxGTK | **priority:** normal | **keywords:** wxStaticText GTK3

#### 2021-01-21 16:10:30: Dummy created the issue

___
A `wxStaticText` will appear truncated on wizard pages if it is in bold.

Bold text appears correctly on the first page of the wizard, but is truncated on other pages.
It is a GTK3-specific issue because it is displayed correctly with GTK2.

Please see the attached changes for the wizard sample, to reproduce this issue.

With GTK2, the second wizard page looks correct:

[[Image(wizard_wx314_gtk2.png)]]

With GTK3, the text on the second wizard page is truncated:

[[Image(wizard_wx314_gtk3.png)]]

vadz:
Unfortunately I won't have time to do anything about this before 3.2.0, sorry.

edward-dauvergne:
I experiences a similar issue [GTK3 Refresh or Layout bug when hiding and showing multiple panels.](https://github.com/wxWidgets/wxWidgets/issues/22668).  It appears as if the bug was fixed with the v3.2.0 tag.

taler21:
As mentioned in [this comment](https://github.com/wxWidgets/wxWidgets/issues/19053#issuecomment-1011258108), the changes in commit 3217a4e8a26bf30758c00852738c0612ca931909 improved the misbehavior but didn't completely fix it.

With the following changes to the wizard sample, the issue is still visible, even with wxWidgets-3.2.0.

```patch
diff -r -u -p a/samples/wizard/wizard.cpp b/samples/wizard/wizard.cpp
--- a/samples/wizard/wizard.cpp	2022-07-25 14:03:43.333751200 +0200
+++ b/samples/wizard/wizard.cpp	2022-07-25 14:08:40.302630200 +0200
@@ -249,13 +249,9 @@ public:
 
         wxBoxSizer *mainSizer = new wxBoxSizer(wxVERTICAL);
 
-        mainSizer->Add(
-            new wxStaticText(this, wxID_ANY, "Try checking the box below and\n"
-                                       "then going back and clearing it"),
-            0, // No vertical stretching
-            wxALL,
-            5 // Border width
-        );
[...]

PR #24781, opened 2024-08-24
Fixes for computing `wxStaticText` best size in wxGTK

vadz:
See the previous discussion in #19053.


paulcor (approved):

vadz:
I've cherry-picked the first commit to 3.2 already, as it seems safe to me, and I think we could cherry-pick the second one too if we don't find any problems with it. This would require some work as we'd have to use a global map/set instead of adding a new member variable, but it looks doable.

PR #24783, opened 2024-08-26
Fix wxAuiTabArt font veiling after SetArtProvider

archivesilver:
Bug fix. 

**Problem:** 
When a new wxAuiTabArt derivative class object is created and configured to be passed as argument in `wxAuiNotebook::SetArtProvider()`, it veils member variables `m_selectedFont` and `m_normalFont` of art provider. This is not immediately clear because it works while setting the art provider. But wxAuiNotebook has its own member variables with the same names, and `wxAuiNotebook::DoModifySelection()` uses these every time the active tab is switched. 

Solution: 
Create accessor functions in `tabart.h` and `tabart.cpp` for these variables, and call these functions in `wxAuiNotebook::SetArtProvider()`. The bug only seems to affect the active tab's title font, but I also fixed normal font because it is the proper way.

### DIFF
--- a/include/wx/aui/tabart.h
+++ b/include/wx/aui/tabart.h
@@ -57,6 +57,9 @@ public:
     virtual void SetColour(const wxColour& colour) = 0;
     virtual void SetActiveColour(const wxColour& colour) = 0;
 
+    virtual wxFont GetNormalFont() const = 0;
+    virtual wxFont GetSelectedFont() const = 0;
+
     virtual void DrawBorder(
                  wxDC& dc,
                  wxWindow* wnd,
@@ -137,6 +140,9 @@ public:
     void SetMeasuringFont(const wxFont& font) override;
     void SetColour(const wxColour& colour) override;
     void SetActiveColour(const wxColour& colour) override;
+
+    wxFont GetNormalFont() const override;
+    wxFont GetSelectedFont() const override;
 
     void DrawBorder(
                  wxDC& dc,
@@ -240,6 +246,9 @@ public:
     void SetMeasuringFont(const wxFont& font) override;
     void SetColour(const wxColour& colour) override;
     void SetActiveColour(const wxColour& colour) override;
+
+    wxFont GetNormalFont() const override;
+    wxFont GetSelectedFont() const override;
 
     void DrawBorder(
                  wxDC& dc,
--- a/src/aui/auibook.cpp
+++ b/src/aui/auibook.cpp
@@ -1799,6 +1799,8 @@ wxAuiNotebook::~wxAuiNotebook()
 void wxAuiNotebook::SetArtProvider(wxAuiTabArt* art)
 {
     m_tabs.SetArtProvider(art);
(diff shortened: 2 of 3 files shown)


vadz:
Just to be sure I understand the problem: does the bug happen if you call `SetNormalFont()` on the art provider directly, `wxAuiNotebook` still uses its own font? If so, it looks like calling `wxAuiNotebook::SetNormalFont()` instead would be a simple workaround. But I could be misunderstanding something, in particular I don't really know what do you mean by "veiling", do you mean "hiding" or something else? As usual, having a minimal example reproducing the problem could be very useful.

Concerning the changes you propose, the problem with adding new pure virtual functions to `wxAuiTabArt` is that it will break any existing classes deriving from it, and I'd prefer to avoid this, if possible.

archivesilver:
> Just to be sure I understand the problem: does the bug happen if you call `SetNormalFont()` on the art provider directly, `wxAuiNotebook` still uses its own font? If so, it looks like calling `wxAuiNotebook::SetNormalFont()` instead would be a simple workaround. But I could be misunderstanding something, in particular I don't really know what do you mean by "veiling", do you mean "hiding" or something else? As usual, having a minimal example reproducing the problem could be very useful.

Not exactly. Steps to reproduce: 
1. Create a wxAuiNotebook
2. Create an art provider. 
3. Call art provider's SetSelectedFont()
4. Assign the art provider to wxAuiNotebook (this time it will work correctly, art provider sets the font)
5. Switch tabs (wxAuiNotebook will use its own m_selectedFont, which shouldn't happen when a tab artist with font configuration is sent. Art provider's `m_selectedFont` and `m_NormalFont` is shadowed by `wxAuiNotebook::DoModifySelection()` which is called when tabs are switched.)

Basically, "art provider" with font can't set font except at SetArtProvider() call. It is inconsistent behavior, a user setting the art provider's font then calling `SetArtProvider()` will assume it works, but tab switching will replace it. Besides, there is no point in having wxAuiNotebook still hold *init* values in `m_selectedFont` and `m_normalFont` member variables. 

> Concerning the changes you propose, the problem with adding new pure virtual functions to `wxAuiTabArt` is that it will break any existing classes deriving from it, and I'd prefer to avoid this, if possible. 

Understandable. However, the only virtual function is added to the abstract `wxAuiTabArt` and is overwritten in the same commit for the standard classes that are deriving from it, `wxAuiGenericTabArt` and `wxAuiSimpleTabArt`. Further derivations usually derive from `wxGenericTabArt`, for example MSW and GTK versions. It didn't break any build checks, but if that's still a concern, abstract function is not necessary. Changes in `wxAuiTabArt` isn't required. Only wxAuiGenericTabArt and wxAuiSimpleTabArt updated without the keyword `override` will still suffice to fix the bug.

vadz:
Thanks for the explanations!

I think we can fix this by doing what I did in #24829, which adds these functions but not as pure virtual ones.

Please let me know if you have any problems with this there, otherwise I'll merge it soon.

Issue #24639, opened 2024-06-24
"Advise" in the ipc sample will stop working after a few iterations when using TCP for IPC -- wxMSW


jpmattia:
### Description
On windows: "Advise" in the ipc sample will stop working after a few iterations when using TCP for IPC
ie #define wxUSE_DDE_FOR_IPC 0

#### Observed behavior and how to reproduce

1. Start both ipcserver and ipcclient.
2. Start the server on the server gui, click connect on the client gui.
3. Click Start Advise on the client gui
4. Click Advise on the server gui a few times

After one or more clicks of "Advise" on the server, the message stops being passed back to the client. 

5. click Stop Advise and disconnect on the client, and usually I see a crash. When running ipcclient under gdb, here is the backtrace:

```
(gdb) bt
#0  0x00007ff7c5a10470 in wxSocketBase::LastError() const ()
#1  0x00007ff7c5a0eac0 in wxSocketInputStream::OnSysRead(void*, unsigned long long) ()
#2  0x00007ff7c5a8f34f in wxInputStream::Read(void*, unsigned long long) ()
#3  0x00007ff7c5a0da29 in wxTCPEventHandler::Client_OnRequest(wxSocketEvent&) ()
#4  0x00007ff7c5afbfa4 in wxEvtHandler::ProcessEventIfMatchesId(wxEventTableEntryBase const&, wxEvtHandler*, wxEvent&) ()
#5  0x00007ff7c5afc13b in wxEventHashTable::HandleEvent(wxEvent&, wxEvtHandler*) ()
#6  0x00007ff7c5afc84b in wxEvtHandler::TryHereOnly(wxEvent&) ()
#7  0x00007ff7c5afc88c in wxEvtHandler::ProcessEventLocally(wxEvent&) ()
#8  0x00007ff7c5afc992 in wxEvtHandler::ProcessEvent(wxEvent&) ()
#9  0x00007ff7c5afd77d in wxEvtHandler::ProcessPendingEvents() ()
#10 0x00007ff7c5a1a4df in wxAppConsoleBase::ProcessPendingEvents() ()
#11 0x00007ff7c5a459d7 in wxEventLoopManual::ProcessEvents() ()
#12 0x00007ff7c5a45a9c in wxEventLoopManual::DoRun() ()
#13 0x00007ff7c5a457bb in wxEventLoopBase::Run() ()
#14 0x00007ff7c5a1bfbc in wxAppConsoleBase::MainLoop() ()
[...]

PR #24793, opened 2024-09-04
IPC sample: Log window for the client should expand vertically as well as horizontally.

jpmattia:
In the IPC sample, the log window in the gui produced by client.cpp does not expand vertically like it does for server.cpp. This fix makes the creation of window elements in the client mirror that of the server, and the log window now expands vertically.

This fix is in response to a side issue was mentioned in in https://github.com/wxWidgets/wxWidgets/issues/24639#issuecomment-2297094102.


vadz:
You probably meant to open this PR against 3.2, not master, please change the base branch (if possible) or rebase your branch with your changes on master and force push it here (preferable, as all PR are typically applied to master and then backported to 3.2 if necessary).

jpmattia:
Yes, but don't see how to change the target. I'll close and redo it.

PR #24794, opened 2024-09-04
IPC sample: Log window for the client should expand vertically as well as horizontally.

jpmattia:
In the IPC sample, the log window in the gui produced by client.cpp does not expand vertically like it does for server.cpp. This fix makes the creation of window elements in the client mirror that of the server, and the log window now expands vertically.

This fix is in response to a side issue that was mentioned in https://github.com/wxWidgets/wxWidgets/issues/24639#issuecomment-2297094102.

### DIFF
--- a/samples/ipc/client.cpp
+++ b/samples/ipc/client.cpp
@@ -124,87 +124,92 @@ MyFrame::MyFrame(wxFrame *frame, const wxString& title)
         IPC_TOPIC, "..."
     };
 
-    wxBoxSizer *item0 = new wxBoxSizer( wxVERTICAL );
+    wxPanel * const panel = new wxPanel(this);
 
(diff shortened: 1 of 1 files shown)


vadz (commented):
Sorry, I've started reviewing this but couldn't see what has changed, so I had to apply the changes locally and then apply them partially to separate the real changes from those just renaming things.

And when I did it, it didn't make much sense to throw it away, so I've created #24798 with your changes properly split into multiple commits and will close this one.

Please try to make as atomic commits as possible for your future PRs and, especially, please don't mix significant and insignificant (renames, whitespace changes, top-level `const` additions, ...) changes inside the same commit.

Thanks in advance!

vadz:
As mentioned above, replaced with the other PR — please let me know if you see anby problems with the changes there.

PR #24796, opened 2024-09-04
Make wxSocketImpl::GetLastError() be called only once, and store the error

jpmattia:
Fix to address a side issue identified in https://github.com/wxWidgets/wxWidgets/issues/24639#issuecomment-2297094102.

In a socket read, there are potentially two calls going to `WSAGetLastError` (via `wxSocketImplMSW::GetLastError()`). The changes here allow only one call and stores the result, in case any intervening system error occurs.

- `wxSocketBase::DoRead()` and `wxSocketBase::DoWrite()` now use `wxSocketImpl::GetError()` instead of `wxSocketImpl::GetLastError()`. In addition, `Accept()` also used `GetLastError()` so it has been changed as well.
- It occured to me that to prevent this error happening in the future, `GetLastError()` should be protected and should be called only by `wxSocketImpl`, ideally as close to when the error occurs in order to set `m_error`.
- Because `GetLastError()` now sets `m_error`, we remove its `const` declaration. 
-  There was a TODO comment in [include/wx/private/socket.h](https://github.com/wxWidgets/wxWidgets/compare/3.2...jpmattia:wxWidgets:jpm_fix_for_socket_double_call#diff-14b41fe3a98df8b1a8a89a6766cba37d518ba37252248e38f92f77c07588b716) listing m_error as something that should be protected, so it is now protected and a `SetError()` accessor added. The accessor is then used by `wxSocketBase` rather than setting `m_error` directly.

### DIFF
--- a/include/wx/private/socket.h
+++ b/include/wx/private/socket.h
@@ -202,11 +202,8 @@ public:
     const wxSockAddressImpl& GetPeer() const { return m_peer; }
 
     wxSocketError GetError() const { return m_error; }
+    void SetError(wxSocketError error) { m_error = error; }
     bool IsOk() const { return m_error == wxSOCKET_NOERROR; }
-
-    // get the error code corresponding to the last operation
-    virtual wxSocketError GetLastError() const = 0;
-
 
     // creating/closing the socket
     // --------------------------
@@ -306,7 +303,6 @@ public:
 
     wxSockAddressImpl m_local,
                       m_peer;
-    wxSocketError m_error;
 
     bool m_stream;
     bool m_establishing;
@@ -321,6 +317,11 @@ protected:
 
     // get the associated socket flags
     wxSocketFlags GetSocketFlags() const { return m_wxsocket->GetFlags(); }
+
+    // get the error code corresponding to the last operation
+    virtual wxSocketError GetLastError() = 0;
+
+    wxSocketError m_error;
 
     // true if we're a listening stream socket
     bool m_server;
--- a/include/wx/private/socket.h
+++ b/include/wx/private/socket.h
@@ -318,8 +318,8 @@ protected:
     // get the associated socket flags
     wxSocketFlags GetSocketFlags() const { return m_wxsocket->GetFlags(); }
 
-    // get the error code corresponding to the last operation
-    virtual wxSocketError GetLastError() = 0;
+    // update the error code to correspond to the outcome of the last operation
+    virtual void UpdateLastError() = 0;
 
     wxSocketError m_error;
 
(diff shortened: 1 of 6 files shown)


include/wx/private/socket.h
```diff
@@ -322,6 +318,11 @@ class wxSocketImpl
     // get the associated socket flags
     wxSocketFlags GetSocketFlags() const { return m_wxsocket->GetFlags(); }
 
+    // get the error code corresponding to the last operation
+    virtual wxSocketError GetLastError() = 0;
```
vadz, line 322: Sorry, the new function name/semantics don't make much sense to me. AFAICS it should become
```suggestion
    // update the error code to correspond to the outcome of the last operation
    virtual void UpdateLastError() = 0;
```

as we don't use its return value and it modifies (and doesn't just "get") the error.

src/common/socket.cpp
```diff
@@ -531,17 +531,26 @@ wxSocketImpl *wxSocketImpl::Accept(wxSocketBase& wxsocket)
     ReenableEvents(wxSOCKET_INPUT_FLAG);
 
     if ( fd == INVALID_SOCKET )
+    {
+	GetLastError();
```
vadz, line 535: Minor, but wrong indentation is used here and elsewhere below, please check the diff.

vadz (commented):
Thanks, the changes look good, but the function really should be renamed (and any of its remaining uses changed to `GetError()`) if we want to change its semantics like this.

Also, I'd like to apply this to master first and backport it to 3.2 later if no problems are found with it in master.

jpmattia:
OK, `UpdateLastError()` sounds right. Updated the name and return value, and made some changes consistent with its new usage. (The return value was used in a small number of places, but easily worked around.)

vadz:
Sorry, after looking at the whole code and not just the diff, I'm not sure how much better does this make things, as we seem to be resetting `m_error` to the same value as it already has in several places now.

Also, I've realized that this PR was against 3.2 branch but I'd really rather make any changes like this in master to avoid breaking something in the stable branch inadvertently.

So, finally, I've created #24811 against master with just the really important part of the changes, i.e. remembering the last error in `wxSocketImpl` instead of trying to get it more than once. I've omitted `SetError()` addition because I don't see any extra safety in having `Get/SetError()` directly modifying `m_error` or using it directly, there are no invariants to preserve here anyhow. And I've also kept the existing `GetLastError()` to avoid meaningless changes and just added `UpdateLastError()`.

Please do let me know what do you think of the changes there, but I'm closing this one because I'd like to apply this only to master, and not to 3.2 (at least to begin with).

PR #24798, opened 2024-09-04
Improve IPC sample client layout

vadz:
Replaces #24794.


jpmattia:
I'm still learning about the best way to do PRs, so thanks for your help and patience.

PR #24799, opened 2024-09-05
Implement dev tools support for `wxWebViewChromium`

vadz:
Add context menu items for showing/hiding CEF developer tools and inspecting the current element to the context menu if access is enabled.

### DIFF
--- a/src/common/webview_chromium.cpp
+++ b/src/common/webview_chromium.cpp
@@ -490,8 +490,19 @@ public:
         return host->GetWindowHandle();
     }
 
+    // Open the dev tools window.
+    bool ShowDevTools()
+    {
+        if ( !m_browser )
+            return false;
+
+        DoShowDevTools(m_browser, CefPoint{});
+
+        return true;
+    }
+
 private:
-    void ShowDevTools(CefRefPtr<CefBrowser> browser,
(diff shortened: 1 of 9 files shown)


src/common/webview_chromium.cpp
```diff
@@ -1793,6 +1863,18 @@ CefRefPtr<CefResourceHandler> ClientHandler::GetResourceHandler(
     return m_resourceManager->GetResourceHandler(browser, frame, request);
 }
 
+void
+ClientHandler::ShowDevTools(CefRefPtr<CefBrowser> browser,
```
PBfordev, line 1867: Is there a reason why basic `ShowDevTools()` is not public? I think it may be handy to be able to show the tool programmatically.

The `wxWebViewEdge` and `wxWebViewWebKit` also support this, see https://github.com/PBfordev/wxecharts/blob/c3f50dc464308b623272f345b06071a1212f2f4a/mainframe.cpp#L308. I can't use macOS so I don't know about that one.

vadz: Looks like a good idea, thanks, I think I can add it. For now let's keep it simple and just allow opening it without specifying the position of the element to inspect, we can always extend it later if really necessary.

PR #35, opened 2015-06-04
Fixed some typos

IlyaBizyaev:
Addition to [PR by orbitcowboy](https://github.com/wxWidgets/wxWidgets/pull/28) (accept that one first).
Replaced all the occurences of _"hilight"_ with _"highlight"_.

PR #24699, opened 2024-07-12
Restore correct translation language lookup (3.2)

vslavik:
As requested, here's #24297 backported to the 3.2 branch.

The one non-trivially different commit is e1b644341dc2347786e3519b3aa3c6c4fea0bc43 — I hope I got the ABI compatibility things mostly right.

PR #24804, opened 2024-09-09
Fix another bug in UI language matching (regression since 3.0)

vslavik:
This is another bug similar to the one from a few months back (sorry, don't have issue numbers handy). I think I was aware of them back then, but forgot about it after I took a break from the topic and didn't realize the underlying cause is something else. 

This (I think) only affects "well-behaved" apps that use `.lproj` folders on Mac, which are named with BCP-47 language tags. But that's something that wx-3.0 supported and that should be done to support some OS features (language override, localizing system strings, ...).

See this nicely absurd log: the requested translation is available, but won't be used:
```
Trace: (i18n) choosing best language for domain 'poedit'
Trace: (i18n)  - available translations: [de,el,zh-Hans,ja,fa,sq,eu,co,uk,kab,nb,pa,en-GB,es,is,sl,da,gl,it,bg,sk,sr,ms,be,sv,cs,ko,hy,zh-Hant,tg,hu,ga,tr,pl,pt-BR,vi,lt,ru,af,uz,fr,fi,id,nl,th,az,pt,pt-PT,ro,bs,hr,ca,en,en,en]
Trace: (i18n)  - system preferred languages: [zh-Hant]
Trace: (i18n)  => using language 'zh-Hans'
```

Or this one (deceptively this looks OK, but macOS uses `pt.lproj` for pt-BR and so it’s not; in any case, the more specific pt-PT match should have been used regardless):
```
Trace: (i18n) choosing best language for domain 'poedit'
Trace: (i18n)  - available translations: [de,el,zh-Hans,ja,fa,sq,eu,co,uk,kab,nb,pa,en-GB,es,is,sl,da,gl,it,bg,sk,sr,ms,be,sv,cs,ko,hy,zh-Hant,tg,hu,ga,tr,pl,pt-BR,vi,lt,ru,af,uz,fr,fi,id,nl,th,az,pt,pt-PT,ro,bs,hr,ca,en,en,en]
Trace: (i18n)  - system preferred languages: [pt-PT]
Trace: (i18n)  => using language 'pt'
```

This is caused by a combination of ill-defined values and good intentions (I *think* it traces back to 8ab635b by @utelle). `GetPreferredUILanguages()` doesn't say what the returned value *is*; in `wx/private/uilocale.h` it even says "The entries contain platform-dependent identifiers" (this is Very Bad™ !). `GetAvailableTranslations()` doesn't say either, but in practice returns what it finds on the filenames (this traces back to me). In macOS case, when you use `.lproj` resources, both lists use language tags.

But elsewhere the latter is typically POSIX locale-like names, while the former is language tags on Windows too I think. And so to "improve compatiblity", this code was changed to — not unreasonably, but subtly incorrectly — normalize to `wxLOCALE_TAGTYPE_POSIX`... which is not necessarily used by the `available` list values.

### Simple fix

This PR does **NOT** implement proper fix. I'm sorry, I lack the mental energy to do that, do it in 3.2 compatible way and get the patch approved right now. Instead, this PR does the minimal, obviously correct [^1] change to fix this particular situation.

Specifically, it looks for an exact match first. I think this isn't a hack, but a reasonable thing to do: both lists come from the same source, it's not unreasonable to expect them to use the same language-code syntax.

This should be applied to both `master` and `3.2` (and is based off the latter).

### Doing it right

I think the only way to properly untangle this is to normalize to use BCP-47 language tags everywhere and particularly in the functions mentioned above and all related APIs. Then convert down to POSIX locales only when trying to load MO files.

wx certainly *should not* do the opposite and convert things into POSIX locales, because some languages cannot be accurately represented by the latter (e.g. zh-Hant and zh-Hans). (Well, and because the rest of the world standardized on the tags.)

The algorithm we use to determine the best language is naive too. See [Unicode Technical Standard \#35](https://unicode.org/reports/tr35/) for the correct algorithm, particularly the “Likely Subtags” and “Language Matching” sections. The latter is somewhat dense and non entirely trivial, but it only took it two paragraphs to say that the truncation-fallback algorithm (which is what wx does, with subsequent broad expansation) is naive and unable to handle the complexities of the real word.

[^1]: I'm going to regret saying this, am I not?

### DIFF
--- a/src/common/translation.cpp
+++ b/src/common/translation.cpp
@@ -133,6 +133,11 @@ wxString GetPreferredUILanguage(const wxArrayString& available)
           j != preferred.end();
           ++j )
     {
+        // try exact match first:
+        if (available.Index(*j, /*bCase=*/false) != wxNOT_FOUND)
+            return *j;
+
+        // try looking up as a POSIX locale:
         wxLocaleIdent localeId = wxLocaleIdent::FromTag(*j);
         wxString lang = localeId.GetTag(wxLOCALE_TAGTYPE_POSIX);
 


utelle:
> This is another bug similar to the one from a few months back (sorry, don't have issue numbers handy). I think I was aware of them back then, but forgot about it after I took a break from the topic and didn't realize the underlying cause is something else.

Well, it would definitely help to know to which other bug you are referring.

> This (I think) only affects "well-behaved" apps that use `.lproj` folders on Mac, which are named with BCP-47 language tags. But that's something that wx-3.0 supported and that should be done to support some OS features (language override, localizing system strings, ...).

AFAIK the logic used for locating translation catalogs is mostly based on POSIX language tags. And I guess the mapping to `.lproj` folders on Mac works for most languages, because usually you have tags consisting of language and region only. In your example below it is different: the language tag of the preferred language (`zh-Hant`) consists of language and script. POSIX language tags map scripts to modifiers - which are completely ignored in the further logic in `translation.cpp`.

> See this nicely absurd log: the requested translation is available, but won't be used:

The result is less absurd as it seems at first glance, but it is all due to the use of POSIX tags - and a missing piece in the mapping of script tags. The tag of the preferred language is `zh-Hant`. Requesting a POSIX tag for this tag results in `zh`, because no region is specified and no mapping from the script `Hant` to a POSIX modifier is known. And therefore the first entry from the list of available languages which starts with `zh` (namely `zh-Hans`) is returned.

> ```
> Trace: (i18n) choosing best language for domain 'poedit'
> Trace: (i18n)  - available translations: [de,el,zh-Hans,ja,fa,sq,eu,co,uk,kab,nb,pa,en-GB,es,is,sl,da,gl,it,bg,sk,sr,ms,be,sv,cs,ko,hy,zh-Hant,tg,hu,ga,tr,pl,pt-BR,vi,lt,ru,af,uz,fr,fi,id,nl,th,az,pt,pt-PT,ro,bs,hr,ca,en,en,en]
> Trace: (i18n)  - system preferred languages: [zh-Hant]
> Trace: (i18n)  => using language 'zh-Hans'
> ```
> 
> Or this one (deceptively this looks OK, but macOS uses `pt.lproj` for pt-BR and so it’s not; in any case, the more specific pt-PT match should have been used regardless):

Well, the default for `pt` in the wxWidgets language database has always been `pt-PT`. However, I'm not sure why `pt-PT` is not used. Probably also due to the use of POSIX tags.

> ```
> Trace: (i18n) choosing best language for domain 'poedit'
> Trace: (i18n)  - available translations: [de,el,zh-Hans,ja,fa,sq,eu,co,uk,kab,nb,pa,en-GB,es,is,sl,da,gl,it,bg,sk,sr,ms,be,sv,cs,ko,hy,zh-Hant,tg,hu,ga,tr,pl,pt-BR,vi,lt,ru,af,uz,fr,fi,id,nl,th,az,pt,pt-PT,ro,bs,hr,ca,en,en,en]
> Trace: (i18n)  - system preferred languages: [pt-PT]
> Trace: (i18n)  => using language 'pt'
> ```
> 
> This is caused by a combination of ill-defined values

What exactly is **ill-defined** in your opinion?

> and good intentions (I _think_ it traces back to [8ab635b](https://github.com/wxWidgets/wxWidgets/commit/8ab635b451ccd3a168d7d46b8cfdd5db83c32994) by @utelle).

Yes, my intentions were definitely good. *sigh* 

There were many lengthy discussions while the `wxUILocale` class was implemented. And most likely not all decisions along the road were perfect.

> `GetPreferredUILanguages()` doesn't say what the returned value _is_; in `wx/private/uilocale.h` it even says "The entries contain platform-dependent identifiers" (this is Very Bad™ !).

Here I disagree. The problem is that we don't have any tag format that covers **_all_** attributes. For example, BCP47 does not cover POSIX's modifier attribute, and POSIX does not cover BCP47's extension attribute. Therefore the best approach is to keep the strings as returned from the operating system. At least as long as we keep using strings.

IMHO it would be better (and most likely less error prone) to use an array of `wxLocaleIdent` objects for the list of available languages and compare entries by looking at the various separate attributes of `wxLocaleIdent`.

> `GetAvailableTranslations()` doesn't say either, but in practice returns what it finds on the filenames (this traces back to me). In macOS case, when you use `.lproj` resources, both lists use language tags.
> 
> But elsewhere the latter is typically POSIX locale-like names, while the former is language tags on Windows too I think. And so to "improve compatiblity", this code was changed to — not unreasonably, but subtly incorrectly — normalize to `wxLOCALE_TAGTYPE_POSIX`... which is not necessarily used by the `available` list values.

If you look at the implementation of `wxUILocale::GetSystemLanguage()` you'll see that the BCP47 tag type is used in conjunction with the `LocaleTag` attribute of language database entries. This works better than POSIX tags. However, the implementation in `translation.cpp` does not make use of the language database, but uses the string list of available languages.

At least the code in `translation.cpp` needs to be reworked, I fully agree.

> ### Simple fix
> 
> This PR does **NOT** implement proper fix. I'm sorry, I lack the mental energy to do that, do it in 3.2 compatible way and get the patch approved right now. Instead, this PR does the minimal, obviously correct [1](#user-content-fn-1-ed8634f896614117301bb1e9852fd17c) change to fix this particular situation.

I'm not absolutely sure that this change is really _obviously correct_. Language tags can come from various sources which are not guaranteed to be in the same tag format. For example, languages can be specified via environment variable where the user may use any tag form.

> Specifically, it looks for an exact match first. I think this isn't a hack, but a reasonable thing to do: both lists come from the same source, it's not unreasonable to expect them to use the same language-code syntax.
> 
> This should be applied to both `master` and `3.2` (and is based off the latter).

IMHO it would be better to do it right immediately (and not to apply a quick hack that may or may not always work).

> ### Doing it right
> 
> I think the only way to properly untangle this is to normalize to use BCP-47 language tags everywhere and particularly in the functions mentioned above and all related APIs. Then convert down to POSIX locales only when trying to load MO files.

No. As mentioned above we currently don't have any tag format that covers **all** variants. As said above I think it would be better to use lists of `wxLocaleIdent` objects instead of string representations. 

> wx certainly _should not_ do the opposite and convert things into POSIX locales, because some languages cannot be accurately represented by the latter (e.g. zh-Hant and zh-Hans). (Well, and because the rest of the world standardized on the tags.)

Correct. But BCP47 is unfortunately not the universal answer. At least not with our current implementation, where we support different sets of attributes for different platforms.

> The algorithm we use to determine the best language is naive too. See [Unicode Technical Standard #35](https://unicode.org/reports/tr35/) for the correct algorithm, particularly the “Likely Subtags” and “Language Matching” sections. The latter is somewhat dense and non entirely trivial, but it only took it two paragraphs to say that the truncation-fallback algorithm (which is what wx does, with subsequent broad expansation) is naive and unable to handle the complexities of the real word.

Well, determining a _"best"_ language if translations for the default language of the user are not available is certainly not a trivial task.

vslavik:
...and this sort of quote-dispute-everything discussion is why I gave up in frustration the last time around (#24699 + linked through). I realize I invited that by trying to provide some background info and tagging you (which was meant as FYI, not blame!), but can we please keep it pragmatic?

The fact is that
 1. `wxTranslations` explicitly allows "passthrough" of identifiers,
 2. this worked in wx-3.0,
 3. wx-3.2 has a regression that broke it and
 4. this PR is straightforward fix that explicitly checks for and allows identical-string passthrough that was accidentally removed in 3.2.

Whether something *more* should be done is the background I shouldn't have bothered with. This PR concerns itself with fixing the regression. I meant to express that the PR in itself is not the end of the journey and more could/should be done, but have chosen my words poorly.

> At least the code in translation.cpp needs to be reworked, I fully agree.

Hold your horses. This is from before your time, so allow me a very brief history lesson: in the 2.8 times, translations were loaded using `wxLanguage` identifier. This was a source of many subtle matching bugs, very much like this one, due to imperfect parsing, imperfect knowledge and reliance on (imperfect) internal wx's database (sounds familiar?). `wxTranslations` addressed these issues and was intentionally designed (by me) to be more flexible and resistant to parsing and knowledge defficiencies (incl. simply being out of date) in wx. 

wxTranslations' approach is not the problem here. Too little interpretation is not the issue; *too much* of (failed) interpretation and parsing is. UI language is highly visible to the user, and it absolutely needs to be permissive and respective of user config even in ways that wx didn't anticipate. Regressing to the 2.8 ways, as you imply, would be backwards.

> I'm not absolutely sure that this change is really obviously correct. Language tags can come from various sources which are not guaranteed to be in the same tag format.

Come on. The function takes a list of requested languages and a set of available languages. If there's a *perfect match*, then arguing that that perfect match *shouldn't* be returned feels like arguing for agument's sake. It is a no-brainer that it should. If the OS supports language "blabla" that wx doesn't have in its database, but there's a translation file for "blabla", it *should* be loaded. As a matter of fact, that was explicit design goal of `wxTranslations`.

Having smarts *on top of that* is good and welcome, but the *basics* are to just not ignore plain equality.

> IMHO it would be better to do it right immediately (and not to apply a quick hack that may or may not always work).

As a reminder, 3.2.6 is currently regressed compared to 3.0 and this is affecting end users (turns out Portuguese *really* don't like seeing Brazilian Portuguese thrown on their screens) and this needs  to be fixed on `3.2`. That's simply not compatible with any major rework.

I think I have enough cred and history here that it should be clear that I don't *do* "quick hacks" as PRs, but do put consideration into them. This is not a "quick hack". This is a considered minimal change to restore equality check that was present in this code before 8ab635b accidentally removed it.

Whether *more* should also be done is an open question, but *this* PR definitely should, to correct broken backwards compatibility if nothing else (and to be explicit about the intention second).

Look, I'm resigned to having to work off a private patched fork for Poedit (what a sad thing for ex-core dev to say, though), but this is a real bug, breaking the `wxTranslations` API promise and backwards compatibility, and should be fixed for other users than me too.

utelle:
> ...and this sort of quote-dispute-everything discussion is why I gave up in frustration the last time around (#24699 + linked through).

Sorry, it wasn't my intention to increase your frustration.

> I realize I invited that by trying to provide some background info and tagging you (which was meant as FYI, not blame!), but can we please keep it pragmatic?

Sure. Actually, it is up to @vadz to decide what to do with this PR.

I chimed in only, because I'm responsible for big parts of the `wxUILocale` implementation (less so for adjustments of `wxTranslations`) and the update of the language database. And I'm ready to take my share to improve the implementation.

BTW, I'm grateful for any background information that helps to clearly and concisely specify the task to fulfill.

> The fact is that
> 
> 1. `wxTranslations` explicitly allows "passthrough" of identifiers,
> 2. this worked in wx-3.0,
> 3. wx-3.2 has a regression that broke it and
> 4. this PR is straightforward fix that explicitly checks for and allows identical-string passthrough that was accidentally removed in 3.2.

I'm not fully convinced that this PR solves the problem in a generic way on all platforms. Checking for an exact match is certainly a good and correct thing to do, but it doesn't fix the flawed implementation of the remaining algorithm to find a reasonable match if no exact match was found. However, maybe I just lack enough insight.

> Whether something _more_ should be done is the background I shouldn't have bothered with. This PR concerns itself with fixing the regression. I meant to express that the PR in itself is not the end of the journey and more could/should be done, but have chosen my words poorly.

IMHO the algorithm to find a best language match should be improved - at least in the long run.

> > At least the code in translation.cpp needs to be reworked, I fully agree.
> 
> Hold your horses. This is from before your time, so allow me a very brief history lesson: in the 2.8 times, translations were loaded using `wxLanguage` identifier. This was a source of many subtle matching bugs, very much like this one, due to imperfect parsing, imperfect knowledge and reliance on (imperfect) internal wx's database (sounds familiar?).

I'm a user of wxWidgets since 2005. I started with 2.8.x and I struggled with language support quite a lot back then. This was part of my motivation to help with improving the situation in 3.x via `wxUILocale` etc.

> `wxTranslations` addressed these issues and was intentionally designed (by me) to be more flexible and resistant to parsing and knowledge deficiencies (incl. simply being out of date) in wx.
> 
> wxTranslations' approach is not the problem here. Too little interpretation is not the issue; _too much_ of (failed) interpretation and parsing is. UI language is highly visible to the user, and it absolutely needs to be permissive and respective of user config even in ways that wx didn't anticipate. Regressing to the 2.8 ways, as you imply, would be backwards.

What gives you the impression that I imply to regress to the 2.8 ways? That's certainly not my intention.

> > I'm not absolutely sure that this change is really obviously correct. Language tags can come from various sources which are not guaranteed to be in the same tag format.
> 
> Come on. The function takes a list of requested languages and a set of available languages. If there's a _perfect match_, then arguing that that perfect match _shouldn't_ be returned feels like arguing for agument's sake. It is a no-brainer that it should. If the OS supports language "blabla" that wx doesn't have in its database, but there's a translation file for "blabla", it _should_ be loaded. As a matter of fact, that was explicit design goal of `wxTranslations`.

As said above checking for an exact match is certainly correct (and I apologize if I accidentally removed that check while trying to adjust `wxTranslations` to use `wxUILocale`). However, the check for an exact match doesn't fix the rest of the algorithm. And using string representations of language tags (which actually **can** originate from sources in different tag formats) makes it difficult to handle the problem properly.

> Having smarts _on top of that_ is good and welcome, but the _basics_ are to just not ignore plain equality.

Fully agreed.

> Whether _more_ should also be done is an open question, but _this_ PR definitely should, to correct broken backwards compatibility if nothing else (and to be explicit about the intention second).

To make it clear: I'm **not** against applying the check for exact match (this PR). However, I would prefer to have it combined with an improved algorithm for the not exact match situations.

> Look, I'm resigned to having to work off a private patched fork for Poedit (what a sad thing for ex-core dev to say, though), but this is a real bug, breaking the `wxTranslations` API promise and backwards compatibility, and should be fixed for other users than me too.

I know this situation well enough myself, because I had to work with a patched wxWidgets version myself for quite a long time. However, I never was a wx core dev. I mainly contributed wx-based components to the wx community and a few patches/additions to wx itself.

I agree that the problem with `wxTranslations` should be fixed - and better sooner than later. If I can help with further improvements, I'm ready to do so.

vadz:
Sorry for the delay with the reply.

I agree that this looks like an "obviously correct thing to do" (but then I've been known to say this about things that turned out to be incorrect in the past...), so we can certainly apply this, just a pity it couldn't be done in time for 3.2.6. The only thing I'd really like to have is a (failing without this PR) unit test, so that we could confirm that any future reworkings don't break this again because this is clearly not a very obvious failure mode. Vaclav, do you think you could add one please?

FWIW I also agree that the correct approach is to use BCP-47 for everything, but I didn't have time to look at the details yet (I hoped to do it before replying, but I've realized I'm not going to be able to find time for this for at least several more days), so I can't really say anything useful about this.

utelle:
> I agree that this looks like an "obviously correct thing to do" (but then I've been known to say this about things that turned out to be incorrect in the past...),

Adding a check for an exact match will fix the issue @vslavik experiences, and it will not do any harm on any platform. However, the PR does not fix the problem in general. If the list of available languages does not contain an exact match, it is still possible that the algorithm returns a unsuitable language. Especially, I suspect that this can happen rather easily on platforms that do not use POSIX tags under the hood like Windows and macOS.

> FWIW I also agree that the correct approach is to use BCP-47 for everything,

BCP-47 is most likely better than POSIX, but as I pointed out in a previous post BCP-47 isn't the universal solution either, because it doesn't cover in full the POSIX _modifier_ attribute. The current implementation in `wxUILocale` maps BCP-47  _scripts_ to POSIX _modifiers_ and vice versa, if there is a mapping defined (like `Latn` <=> `latin`).

A problem related to the bug this PR is all about is that the Unicode list of scripts does not contain entries which map the _script_ tags `Hans` and `Hant` to long forms that would be valid modifiers for POSIX. Therefore the POSIX form currently completely ignores these _script_ tags. This is certainly not good, but I don't know what else should/could be reasonably done.

For POSIX the equivalent of `zh-Hans` would be `zh_CN`, and the equivalent of `zh-Hant` would be `zh_TW`. In principle, a tag like `zh-Hant-CN` would be valid, too, but I haven't ever seen a language tag like `zh_CN@Hant` on a Linux system.

Probably, the script tag mapping should be adjusted to use identity for `Hans` and `Hant` (as they are used like that in the wx  language database), but this can result in uncommon POSIX language tags.

> but I didn't have time to look at the details yet (I hoped to do it before replying, but I've realized I'm not going to be able to find time for this for at least several more days), so I can't really say anything useful about this.

If you want me to, I could dedicate some time to this task. But if you or @vslavik want to do it yourself I'm fine with that, too.

vadz:
@utelle I'd definitely appreciate any contributions here, I don't expect to have time/motivation to work on this myself any time soon.

My main problem is that I'm very afraid of making changes to this code because clearly our existing tests are not enough, as it's not the first time that we don't notice some breakage even though all tests pass. And I'm making this problem worse by not adding a test for the bug fixed by this change myself too, but I just can't do it today and if I don't apply this PR today, I don't know when will I be able to do it, so I'll merge it soon — but tests would still be very welcome and then, once we have the test that can confirm that we don't break this in the future, a more general and nicer solution would be very welcome too. TIA!

utelle:
> @utelle I'd definitely appreciate any contributions here

Ok. I'll take a look at how the algorithm to find a best match between preferred language and supported language can be improved.

> My main problem is that I'm very afraid of making changes to this code because clearly our existing tests are not enough, as it's not the first time that we don't notice some breakage even though all tests pass. And I'm making this problem worse by not adding a test for the bug fixed by this change myself too, but I just can't do it today and if I don't apply this PR today, I don't know when will I be able to do it, so I'll merge it soon — but tests would still be very welcome

I will look into the test issue, too. However, it would be really nice if @vslavik could contribute test cases.

> and then, once we have the test that can confirm that we don't break this in the future, a more general and nicer solution would be very welcome too. TIA!

The algorithm described in [Unicode Technical Standard #35](https://unicode.org/reports/tr35/) to which @vslavik referred is based on essentially 2 data tables (mapping of "likely subtags" for language tags, rules for "language matching"). Especially the list of "likely subtags" is rather huge, because it contains mappings for more than **7000** 3-letter language tags of mostly rather exotic languages. The language database used in wxWidgets is currently mainly based on the list of languages supported by Windows - and that list contains only roughly **100** 3-letter language tags Iin addition to the usual 2-letter language tags). IMHO it should be enough to restrict the algorithm to use mappings of all 2-letter language tags and the reduced list of 3-letter language tags known and supported in Windows. I would really appreciate if @vslavik would give his opinion on this.

vslavik:
> Vaclav, do you think you could add one please?

Done now, sorry for the delay.

> BCP-47 is most likely better than POSIX, but as I pointed out in a previous post BCP-47 isn't the universal solution either, because it doesn't cover in full the POSIX modifier attribute. 

POSIX *locales* are not fully relevant for *language* choice, the few that are (`@latin`, `@cyrillic`, `ca@valencia` etc.) can and should be special-cased. The rest should go to private subtags so that *roundtrip* from POSIX locale to language tag back to POSIX (for file loading, i.e. definitely important) can be done w/o any loss of information.

W/ private tags, I really think tags *are* universal (for the purpose of describing a language). That's not the only consideration, though, it's also important that they are well-standardized, maintained, widely used and well understood, used by modern macOS and Windows, and that they have well-defined algorithms for them. 

> A problem related to the bug this PR is all about is that the Unicode list of scripts does not contain entries which map the script tags Hans and Hant to long forms that would be valid modifiers for POSIX. Therefore the POSIX form currently completely ignores these script tags. This is certainly not good, but I don't know what else should/could be reasonably done.

I think this is pretty simple, actually. You're just not going to be able to handle this with a generic, exceptions-less algorithm. POSIX uses `zh_CN` and `zh_TW` *traditionally* as substitutes for `zh-Han{s,t}`, because it lacks a way to express the script correctly. It's not formally correct, but it is what people do. Consequently, `zh-Hant` has to be special-cased and mapped to `zh_TW` when converting to a POSIX locale, and it needs to be done *regardless* of whether the tag has a region set or not. (In the opposite direction, `zh_TW` can be converted to `zh-TW`, then maximized to `zh-Hant-TW`, then during lookup minimized appropriately.)

Or perhaps it's better to conclude that the language tag to POSIX locale conversion is a one-to-many relationship, especially if `zh@Hant` etc. (listed in languageinfo.cpp) is actually used in the wild. I have never encountered it myself, did you? If it is used, it would imply that for `zh-Hant-HK` user preference, all of `[zh_HK, zh_TW, zh@Hant]` should be matched.

> For POSIX the equivalent of zh-Hans would be zh_CN, and the equivalent of zh-Hant would be zh_TW. In principle, a tag like zh-Hant-CN would be valid, too, but I haven't ever seen a language tag like zh_CN@Hant on a Linux system.

Not only in principle, in *practice* too. This is new information for me (literally a day after this was merged) and I am yet unclear about the precise circumstances, but sometimes macOS asks for the language as e.g. `zh-Hant-TW` even if you set it as not Taiwan-specific, or `zh-Hant-HK` (this one does make sense) or — this gets worse — `zh-Hant-DE` (DE taken from your region), even if you set the language to *Traditional Chinese (Taiwan)*. So we must handle this case even when it doesn't make sense.

Needless to say, the current algorithm completely chokes on the above macOS example *and* will pick Chinese variant at "random" (read: in filesystem order) if presented with `zh-Hant-HK` [^1]

> IMHO it should be enough to restrict the algorithm to use mappings of all 2-letter language tags and the reduced list of 3-letter language tags known and supported in Windows. I would really appreciate if @vslavik would give his opinion on this.

I am, in principle, unhappy about any code that relies on wx's internal database of languages — it *will* get out of date, it's just a matter of time, and then it *will* break things for somebody using a newly added language (as used to happen in 2.8 days).

But you're right about obscurity and it is some ~230kB of data, so it IMHO does make sense to do as you propose and fall back to passthrough + possibly simpler lookup algorithm (e.g. just try w/o region, w/o script) for languages not in the db.


[^1]: it 1. fails the passthrough from this PR, 2. gets converted to `zh_HK` and having only zh-Han{s,t} or zh_{CN,TW} available, fails to find a match, then 3. uses first match for `zh_*` or fail if it has only `.proj` files...

utelle:
> POSIX _locales_ are not fully relevant for _language_ choice, the few that are (`@latin`, `@cyrillic`, `ca@valencia` etc.) can and should be special-cased.

AFAICT the POSIX modifiers `@latin`, `@cyrillic`, `ca@valencia` are handled correctly by the current implementation in wx. However, the _script_ subtags `Hans` and `Hant` are not, because they are not included in the _script_ mapping due to undefined long forms (used as POSIX modifiers). (Probably I should add an identity mapping for the script tags `Hans` and `Hant`, so that the conversion to POSIX does not result in nonsense, but a language tag that can be looked up in wx's language database.)

> The rest should go to private subtags so that _roundtrip_ from POSIX locale to language tag back to POSIX (for file loading, i.e. definitely important) can be done w/o any loss of information.

In the BCP-47 representation used in `wxLocaleIdent` additional subtags (that is, subtags other than _language_, _script_ and _region_) are saved in the _extension_ attribute. However, the _extension_ attribute can't be translated to the _modifier_ attribute of the POSIX representation, because the _modifier_ attribute can contain only a single value, while the _extension_ attribute can contain any number of subtags. 

> W/ private tags, I really think tags _are_ universal (for the purpose of describing a language). That's not the only consideration, though, it's also important that they are well-standardized, maintained, widely used and well understood, used by modern macOS and Windows, and that they have well-defined algorithms for them.

The algorithms described in the Unicode docs are based only on 3 subtags: language, script and region. And IMHO the lists of desired resp supported languages will typically use only those subtags to specify the languages. However, you may know better from your experiences with poedit.

> > script tags Hans and Hant
> 
> I think this is pretty simple, actually. You're just not going to be able to handle this with a generic, exceptions-less algorithm. POSIX uses `zh_CN` and `zh_TW` _traditionally_ as substitutes for `zh-Han{s,t}`, because it lacks a way to express the script correctly.

In the wx language database we have this mapping in the attribute `canonicalref`. However, `wxLocaleIdent` currently doesn't handle this mapping. But it should, of course.

> It's not formally correct, but it is what people do. Consequently, `zh-Hant` has to be special-cased and mapped to `zh_TW` when converting to a POSIX locale, and it needs to be done _regardless_ of whether the tag has a region set or not. (In the opposite direction, `zh_TW` can be converted to `zh-TW`, then maximized to `zh-Hant-TW`, then during lookup minimized appropriately.)

Maximizing is relatively easy. However, when minimizing typically the region is retained and the script dropped, if it is the default script of the region.

> Or perhaps it's better to conclude that the language tag to POSIX locale conversion is a one-to-many relationship, especially if `zh@Hant` etc. (listed in languageinfo.cpp) is actually used in the wild. I have never encountered it myself, did you?

No, I never saw `zh@Hans` or `zh@Hant` in the wild. These forms are used in `languageinfo.cpp` only so that the _canonical_ attribute has unique values. Under Unix the attribute `canonicalref` will be used if it is specified.

> > For POSIX the equivalent of zh-Hans would be zh_CN, and the equivalent of zh-Hant would be zh_TW. In principle, a tag like zh-Hant-CN would be valid, too, but I haven't ever seen a language tag like zh_CN@Hant on a Linux system.
> 
> Not only in principle, in _practice_ too. This is new information for me (literally a day after this was merged) and I am yet unclear about the precise circumstances, but sometimes macOS asks for the language as e.g. `zh-Hant-TW` even if you set it as not Taiwan-specific, or `zh-Hant-HK` (this one does make sense) or — this gets worse — `zh-Hant-DE` (DE taken from your region), even if you set the language to _Traditional Chinese (Taiwan)_. So we must handle this case even when it doesn't make sense.

For setting the locale something like `zh-Hant-DE` perfectly makes sense, because this would require to use a decimal comma for formatting floating point numbers, while the language will be traditional Chinese.

> Needless to say, the current algorithm completely chokes on the above macOS example _and_ will pick Chinese variant at "random" (read: in filesystem order) if presented with `zh-Hant-HK` [1](#user-content-fn-1-176265973d717314ef1c83d7075d6a09)

Yes, the algorithm has to be fixed to handle Chinese correctly. However, AFAICT Chinese is the only language affected by this problem. And only if the POSIX representation is used. The BCP-47 representation should not exhibit such strange behaviour.

> I am, in principle, unhappy about any code that relies on wx's internal database of languages — it _will_ get out of date, it's just a matter of time, and then it _will_ break things for somebody using a newly added language (as used to happen in 2.8 days).

To handle maximizing and minimizing locale tags while trying to find the best match between desired and supported languages any algorithm will need an excerpt of Unicode data tables ( language, script, and region codes etc). Those Unicode data tables also change over time - for example, many 3-letter language codes are marked as deprecated and new codes are added, if necessary.

So, the problem of possibly outdated data will not go away by **not** using wx's internal language database.

wx's internal language database is currently based on the list of locales officially supported by Windows 11. The list contains about 900 entries. Maybe we could check whether macOS supports additional locales and add them to the wx language database. However, I currently don't have access to a macOS system. Maybe you could provide a list of [availableLocaleIdentifiers](https://developer.apple.com/documentation/foundation/nslocale/1410448-availablelocaleidentifiers?language=objc) that I could use to update the wx language database.

> But you're right about obscurity and it is some ~230kB of data, so it IMHO does make sense to do as you propose and fall back to passthrough + possibly simpler lookup algorithm (e.g. just try w/o region, w/o script) for languages not in the db.

We have to differentiate between setting the UI locale and selecting the best translation catalogs. For setting the UI locale it makes sense to support only actually available locales. However, the translation language could be anything - as long as the system font supports all required code points. So, falling back to a simpler lookup algorithm is certainly a reasonable option.

I will start to work on an improved language matching algorithm soon, taking into account the Unicode information as much as necessary.

utelle:
@vadz: I started to work on this issue. In this course I also updated the wx language database. I detected that the recent Windows 11 version no longer supports the language/locale **Dari**:

```
wxLANGUAGE_DARI              3.1.6   prs      prs      prs_AF  0x8c 0x01 RTL "Dari"
wxLANGUAGE_DARI_AFGHANISTAN  3.1.6   prs-AF   prs_AF   -       0x8c 0x01 RTL "Dari (Afghanistan)"
```

The question is whether we should keep this language/locale in the wx language database. Personally, I would drop these entries, because I highly doubt that anyone has used these entries in a productive wx application.

However, if you prefer to keep those entries I will add them manually.

vslavik:
> AFAICT the POSIX modifiers @latin, @cyrillic, ca@valencia are handled correctly by the current implementation in wx.
> ...
> However, the extension attribute can't be translated to the modifier attribute of the POSIX representation, because the modifier attribute can contain only a single value, while the extension attribute can contain any number of subtags.

Possibly (although I see only mentions of `ca_ES@valencia` in the code, while translations commonly use `ca@valencia`, and the `wxLANGUAGE_SERBIAN_CYRILLIC` entry is wrong; nobody uses `sr@cyrillic`). That's besides the point, I merely said what *should* be done, not what is done, and only in response to your complain that language tags are not sufficient. That's not the case and BCP-47 can - and should - be used thorough: `GetAvailableTranslations()` can return BCP-47 tags even on POSIX systems (e.g. `de-DE-x-euro`), so can `GetPreferredUILanguages()` and finally, `Add*Catalog()` can find `de-DE-x-euro` translations by round-tripping back to `de_DE@euro`. 

> The algorithms described in the Unicode docs are based only on 3 subtags: language, script and region.

Sorry, I no longer understand what your argument w.r.t. BCP-47 suitability even is. Either you think it is not good because it can't represent POSIX modifiers well (as you repeatedly said, in opposition to me me and @vadz), in which case round-tripping through private subtags is the answer. Or you say they are irrelevant to language choice (which is true), but in that case BCP-47 is just fine.

To be as clear as possible about my position:

1. BCP-47 tags should be used consistently everywhere 
2. POSIX locales w/ modifiers should be converted to them in a round-trip safe manner (with hardcoded handling of common cases like Chinese or Serbian/Cyrillic)
3. Language matching should be done in BCP-47 space

> No, I never saw zh@Hans or zh@Hant in the wild. These forms are used in languageinfo.cpp only so that the canonical attribute has unique values.

I see. Made-up canonical values are definitely *very* surprising :(

> Maximizing is relatively easy. However, when minimizing typically the region is retained and the script dropped, if it is the default script of the region.

The algorithm explicitly has region-leaning and script-leaning variants...

> For setting the locale something like zh-Hant-DE perfectly makes sense [...]

Yes, it's a safe bet that everybody here knowns what a locale is, methinks. Again, macOS presents this abomination as the *language*. To be clear, I'm not saying this should necessarily be handle in language matching code (though it can, by trying both region- and script-leaning matches), it could just as well be pre-processed in `GetPreferredUILanguages()`, but it needs to be handled on macOS one way or another.

> > Needless to say, the current algorithm completely chokes on the above macOS example and will pick Chinese variant at "random" (read: in filesystem order) if presented with zh-Hant-HK [1](https://github.com/wxWidgets/wxWidgets/pull/24804#user-content-fn-1-176265973d717314ef1c83d7075d6a09)
> Yes, the algorithm has to be fixed to handle Chinese correctly. However, AFAICT Chinese is the only language affected by this problem. And only if the POSIX representation is used. The BCP-47 representation should not exhibit such strange behaviour.

At the moment, wx fails to find the translation *regardless* of how the translations are enumerated.

Yes, Chinese is the only such case I am aware of, i.e. it can be special-cased.

The only vaguely similar cases requiring special-casing that I am aware of is Cyrl/Latn handling, which is inconsistent between languages. E.g. Serbian defaults to Cyrillic on Linux and has `sr` and `sr@latin`, while Uzbek is the other way around: `uz` and `uz@cyrillic`. Those can be derived from likely subtags, but it may (or not) be better to be explicit.

> So, the problem of possibly outdated data will not go away by not using wx's internal language database.
>
> wx's internal language database is currently based on the list of locales officially supported by Windows 11. The list contains about 900 entries. Maybe we could check whether macOS supports additional locales [...]

Sorry if I was unclear: The issue is not with having the data, but with wx's language database being bespoke (which really is just code for "not CLDR"). It's easy to periodically pull from CLDR; it's less easy to merge that with a bespoke database, particularly when it comes to adding entries to it.

macOS uses ICU (and thus CLDR) under the hood (so does Windows and ~everybody these days) and BCP-47 in the APIs. It doesn't make sense to me to enumerate state of some macOS version like that. Why don't go to the source, i.e. CLDR, instead? (Or even pull it from the OS at runtime, but that would be a major change?)

> We have to differentiate between setting the UI locale and selecting the best translation catalogs. For setting the UI locale it makes sense to support only actually available locales.

(For non-default ones, sure. For the default UI locale, emphatically no for precisely the same reasons as with translations.)

utelle:
> although I see only mentions of `ca_ES@valencia` in the code, while translations commonly use `ca@valencia`, and the `wxLANGUAGE_SERBIAN_CYRILLIC` entry is wrong; nobody uses `sr@cyrillic`.

While working on the class `wxUILocale` I reworked the wx language database. Until November 2021 the language database file `langtabl.txt` located in `misc/languages` was maintained manually. This is no longer the case. The file can now be generated based on a list of supported locales in Windows (extracted from Windows via a small application) and several Unicode data files (downloaded from Unicode resp CLDR).

Windows uses BCP-47 tags and those tags were added to the wx language database in November 2021. Windows uses `ca-ES-valencia` for _Valencian_ (and the earlier wx language database file used `ca_ES@valencia`). I know that the wx file for localization of _Valencian_ is named `ca@valencian`, but I haven't seen a real world wx example using _Valencian_. So, I don't know whether specifying `ca@valencian` currently works for selecting _Valencian_ translations or not.

Regarding _Serbian_ I agree that there is a problem. As said the current wx language database is based on the list of locales/languages supported by Windows. Windows associates the tag `sr` with _Serbian (Latin)` - additional tags always include script and language. OTOH _CLDR_ associates the tag `sr` with `sr-Cyrl-RS`. So, we have contradicting information from different sources.

> That's besides the point, I merely said what _should_ be done, not what is done, and only in response to your complain that language tags are not sufficient.

Well, I don't _complain_, but simply _note_ that there is no direct correspondence between BCP-47 and POSIX. While BCP-47 allows to specify more than one extension subtag, I'm not aware of POSIX support for more than one extension subtag.

The main question is what is actually needed in wx. `wxLocaleIdent` currently tries to be very generic, supporting the tag variants used on Windows, Unix-like systems, and macOS - including extension subtags. Is this really necessary? If yes, to which degree? I really don't know.

> That's not the case and BCP-47 can - and should - be used thorough: `GetAvailableTranslations()` can return BCP-47 tags even on POSIX systems (e.g. `de-DE-x-euro`), so can `GetPreferredUILanguages()` and finally, `Add*Catalog()` can find `de-DE-x-euro` translations by round-tripping back to `de_DE@euro`.

Yes, BCP-47 can return a tag like `de-DE-x-euro` (and `wxLocaleIdent` can handle such tags already). But for a round-trip to a POSIX tag `de_DE@euro` you need to know that `@euro` maps to `x-euro` (in contrast to `@valencian` mapping to just `-valencian`. Up to now I haven't found a source for such mappings.

Is it reasonable to restrict the wx users to specify desired translation languages with a combination of the subtags _language_, _script_ and _region_?

> > The algorithms described in the Unicode docs are based only on 3 subtags: language, script and region.
> 
> Sorry, I no longer understand what your argument w.r.t. BCP-47 suitability even is.

The problem is not BCP-47 in itself. If the BCP-47 tag variant would be used throughout wxWidgets, there wouldn't be a problem, I guess. But wxWidgets uses (or at least used in the past) the POSIX tag variant for locating subdirectories of translation files - at least under Windows and Unix-like systems. And for BCP-47 tags with extensions I'm not aware of a generic mapping algorithm to convert to valid POSIX tags (and vice versa).

The Unicode algorithms just look at the subtags _language_, _script_, and _region_. However, extension tags are retained and readded to the final result.

> Either you think it is not good because it can't represent POSIX modifiers well (as you repeatedly said, in opposition to me and @vadz), in which case round-tripping through private subtags is the answer.

As said above I don't see the problem in BCP-47 itself, but in the mapping between BCP-47 and POSIX. As long as you only have to deal with such mappings internally, it may work correctly. But tags will come from the outside - from the user. How do you map valid BCP-47 tags like `gsw-u-sd-chzh` or `he-IL-u-ca-hebrew-tz-jeruslm` to **valid** POSIX tags without loosing anything?

> Or you say they are irrelevant to language choice (which is true), but in that case BCP-47 is just fine.

The question is how generic we need to be in wxWidgets. If we restrict language selection to tags based on _language_, _script_, and _region_, all would be fine most likely. But what should be done if the user adds modifiers/extensions in his list of desired languages or the list of supported languages includes modifiers/extensions?

Example: A user desires `de-DE-x-euro` and the list of supported languages includes `de_DE@dollar` and `de_DE@gregorian`. Which translation should be used? I know the example is artificial and in practice probably no one would provide different translations based on different currencies or calendars. Nevertheless I hope the example makes clear that is necessary to define how modifiers/extensions should be handled, and that it is necessary to define how the mapping between BCP-47 and POSIX should take place.

> To be as clear as possible about my position:
> 
> 1. BCP-47 tags should be used consistently everywhere

The current mixture of BCP-47 and POSIX is due to the fact that compatibility with prior wxWidgets versions had high priority.

I'm fine with using BCP-47 internally (almost) exclusively. However, if for example directory names for translation files are in POSIX syntax, then a clear unique mapping between the representations must be defined - possibly including modifiers/extensions.

> 2. POSIX locales w/ modifiers should be converted to them in a round-trip safe manner (with hardcoded handling of common cases like Chinese or Serbian/Cyrillic)

As long as only _language_, _script_, and _region_ subtags are involved this should already work.

Problematic are special cases. Take _Chinese_ for example. `zh-Hans` can be mapped to `zh_CN`. Ok. But to what do you map `zh-Hant-CN`? If you want to be precise, you can't use `zh_CN`, because it uses the wrong script. But which modifier should be used to make it clear?

> 3. Language matching should be done in BCP-47 space

I agree. However, this doesn't answer the question how to handle modifiers/extensions, if they are present.

> > No, I never saw zh@Hans or zh@Hant in the wild. These forms are used in languageinfo.cpp only so that the canonical attribute has unique values.
> 
> I see. Made-up canonical values are definitely _very_ surprising :(

In practice, this doesn't impose a problem, because the attribute `canonicalref` (which will contain `zh_CN` resp `zh_TW`) will be used automatically as the POSIX tag.

> > Maximizing is relatively easy. However, when minimizing typically the region is retained and the script dropped, if it is the default script of the region.
> 
> The algorithm explicitly has region-leaning and script-leaning variants...

Yes ... but who decides which variant to use? In practice, region-leaning is used more often.

> The only vaguely similar cases requiring special-casing that I am aware of is Cyrl/Latn handling, which is inconsistent between languages. E.g. Serbian defaults to Cyrillic on Linux and has `sr` and `sr@latin`, while Uzbek is the other way around: `uz` and `uz@cyrillic`. Those can be derived from likely subtags, but it may (or not) be better to be explicit.

As pointed out above, under Windows Serbian defaults to Latin. At least at the locale level. For determining the translation language the CLDR data would default to Cyrillic.

> Sorry if I was unclear: The issue is not with having the data, but with wx's language database being bespoke (which really is just code for "not CLDR"). It's easy to periodically pull from CLDR; it's less easy to merge that with a bespoke database, particularly when it comes to adding entries to it.

As described at the top the database is generated from CLDR and other data. However, there are some manually added modifications in the process to keep compatibility with prior versions of wxWidgets.

Adjusting the modifications should be necessary only on rare occasions.

> macOS uses ICU (and thus CLDR) under the hood (so does Windows and ~everybody these days) and BCP-47 in the APIs.

That's true, but doesn't help much under Unix/Linux, unless wxWidgets starts to use the ICU library.

> It doesn't make sense to me to enumerate state of some macOS version like that. Why don't go to the source, i.e. CLDR, instead?

Indirectly this is already done (with the locale data extracted from Windows). The main obstacle is the required compatibility with prior versions of wxWidgets. Who the hell needs all these `wxLANGUAGE_*` constants? Using BCP-47 strings everywhere would be much better.

> (Or even pull it from the OS at runtime, but that would be a major change?)

In principle, retrieving the language/locale data at runtime would not be a big deal under Windows and macOS. But Linux/Unix doesn't offer an API to query this information (AFAIK). You may be able to retrieve a list of installed locales, but that would be a very short list on most systems.

Even if the langiage/locale data are read from the OS the main problem would be to establish compatibility with prior wxWidgets versions (due to several inconsistencies in the data - which are currently handled in the generating scripts).

vslavik:
> Regarding Serbian I agree that there is a problem. [...] So, we have contradicting information from different sources.
> [...]
> As pointed out above, under Windows Serbian defaults to Latin. At least at the locale level.

Again, GNU gettext uses `sr` for Cyrillic and `sr@latin` for Latin. This has nothing to do with Windows. There is no "contradicting information". These are facts on the ground. 

> Is it reasonable to restrict the wx users to specify desired translation languages with a combination of the subtags language, script and region?

Again, no — not for the only interesting case, i.e. respecting system defaults. 

> As long as you only have to deal with such mappings internally, it may work correctly. But tags will come from the outside - from the user.

Again, internal round-tripping between the two functions mentioned above is the only relevant situation. Again, no, the tags will absolutely *not* come from the user,  but from the OS and from the list of available translations. There isn't any need on generally converting BCP47 to POSIX tags, which you seem to fixate on.

> Problematic are special cases. Take Chinese for example. zh-Hans can be mapped to zh_CN. Ok. But to what do you map zh-Hant-CN?

Again, for crying out loud, Chinese needs to be handled specially. Again, zh-Hant-CN is *not an existing language*. Again, zh-Hant **MUST** be mapped to zh_TW **when looking up translation files** and again, it is **for legacy reasons**. Again, this is not controversial, or a subject of *reasonable* argument. This is dead simple. 

> Yes ... but who decides which variant to use?

Geez, I don't know, the author of the language matching code?

> Example: A user desires de-DE-x-euro and the list of supported languages includes de_DE@dollar and de_DE@gregorian. Which translation should be used? I know the example is artificial

Yes, it is nonsense, far removed from the actual issue (which is respecting user defaults and not doing embarrassingly bad mistakes such as using the wrong script or language, as wx-3.2+ does and wx-3.0 didn't do).  You're making up artificial "issues" to argue about, instead of focusing on getting the basics correct.

Sorry, but this discussion long ago lost its usefulness. I'm off this PR.

I explained the real-world scenarios, edge cases and current breakages above. Happy to review changes or provide additional edge cases (plenty of them above, though).

utelle:
> Sorry, but this discussion long ago lost its usefulness. I'm off this PR.

I fully agree. This PR is already closed. Be happy with it being applied.

Nevertheless I will respond to some of your comments.

> Again, GNU gettext uses `sr` for Cyrillic and `sr@latin` for Latin. This has nothing to do with Windows. There is no "contradicting information". These are facts on the ground.

This is the **first** time you mention _GNU gettext_. So, why are you saying "Again"?

The point I intend to make is that there is no _universal truth_. You say _GNU gettext_ uses `sr` for Cyrillic and that that is a fact. And I say Windows uses `sr` for Latin, and that is also a fact.

Earlier you mentioned _Portuguese_ and that macOS uses `pt` for Brazil (`pt-BR`). Let's see what _GNU gettext_ uses. Ah, _GNU gettext_ uses `pt` for Portugal (`pt-PT`). Ask the people in Portugal and they will agree that _GNU gettext_ is right and macOS wrong.

Another example: English. Almost everywhere nowadays `en` represents `en-US`. wxWidgets is an exception: `en` stands for `en-GB`  according to the language database. Why? Because the original wxWidgets author is Britain. And this mapping was retained for compatibility.

> Again, internal round-tripping between the two functions mentioned above is the only relevant situation. Again, no, the tags will absolutely _not_ come from the user, but from the OS and from the list of available translations. There isn't any need on generally converting BCP47 to POSIX tags, which you seem to fixate on.

Again, I'm not fixated to converting BCP47 to POSIX and vice versa in all generality. However, the class `wxLocaleIdent` not only handles language tags, but also locale tags. For example, POSIX tags under Linux/Unix can contain a charset attribute. Currently `wxLocaleIdent` simply drops the charset attribute if asked to convert to BCP47.

I only ask for the cases and attributes we have to handle on different platforms.

BTW, it is simply not true that the tags will only come from the OS. If your application uses only the new `wxUILocale` class and the default language/locale of the system, you are right. But there is also `wxLocale` which can choose any locale/language and which gets the tag via parameter from the outside - and the outside is not only the OS, but also the user.

> Again, for crying out loud, Chinese needs to be handled specially.

I really don't know why you use this kind of language. I already agreed that the current handling of Chinese is flawed and that it has to be corrected.

> Again, zh-Hant-CN is _not an existing language_.

It is a valid locale. A Taiwanese may choose this locale for his notebook when visiting mainland China.

Again, `wxLocaleIdent` not only handles languages, but also locales. And even a language tag can contain _language_ **and** _script_ **and** _region_ subtags, and it is not immediately clear which tags describe existing languages and which not. You yourself brought the example of language "blabla" which  wxWidgets should handle correctly, even though the tag doesn't conform to BCP47 nor to POSIX.

> I explained the real-world scenarios, edge cases and current breakages above. Happy to review changes or provide additional edge cases (plenty of them above, though).

If you are so smart and I am so dumb, it is most likely better that you implement improved algorithms.

Issue #24790, opened 2024-09-01
EVT_HTML_CELL_HOVER is not seen on macOS Sonoma


andykt:
On my M3 MacBook Air running Sonoma (14.6.1) it seems that
wxHtmlWindow doesn't get any EVT_HTML_CELL_HOVER events.

It's easy to see the bug in samples/html/test/test.app.
The cursor doesn't change when the mouse moves over text or links,
and OnHtmlCellHover is never called (no "Mouse moved over cell ..."
log messages are displayed).

I'm using the latest stable wxOSX 3.2.5.

andykt:
Possibly useful clue: If you click and hold down the mouse button then
the cursor DOES change as it is moved over text or links.
"Mouse moved over cell ..." messages are also displayed.

andykt:
It turns out that this bug is a regression caused by the addition of m_osxClipView in wxOSX 3.2.5 (test.app works fine in 3.2.4).  I've just tried test.app using 3.2.6 and things have improved slightly, but you still have to click once in the window before the cursor will behave correctly.  Maybe the code in wxWidgetCocoaImpl::InstallEventHandler needs to check if m_osxClipView is set and do the NSTracking stuff on that view rather than m_osxView?  Ditto for the Associate call?

csomor:
Hi @andykt, thank you for looking deeper into this, I'll do some tests ...

csomor:
@andykt could you please test against master ?

andykt:
@csomor Can confirm your change fixes the problem -- html/test/test.app now works correctly.  Also tested a couple of my apps that use wxHtmlWindow.  Much thanks!

PR #24809, opened 2024-09-14
OSX fix tracking with clip view on macOS 14 Sonoma

csomor:
see  #24790


vadz:
Thanks Stefan, I'm applying this to master and 3.2 too.

PR #24810, opened 2024-09-14
Fix wxALWAYS_SHOW_SB comment

Blake-Madden:
Comment says that `wxALWAYS_SHOW_SB` is Universal and MSW only, but I don't think that is correct anymore. MSW and OSX (from what I can tell) use generic which supports `wxALWAYS_SHOW_SB`. GTK+ explicitly supports `wxALWAYS_SHOW_SB`.  I can't find any other implementation in code base of `wxScrolled` which does **NOT** support this.


vadz:

PR #24811, opened 2024-09-15
Remember last error in wxSocketImpl

vadz:
Instead of calling GetLastError() multiple times, and possibly getting different errors, make it private and call UpdateLastError(), which remembers the value returned by GetLastError() in m_error, so that we can be sure it doesn't change.

This commit is best viewed with Git --color-moved option.

See #24796.

### DIFF
--- a/include/wx/msw/private/sockmsw.h
+++ b/include/wx/msw/private/sockmsw.h
@@ -47,8 +47,6 @@ public:
 
     virtual ~wxSocketImplMSW();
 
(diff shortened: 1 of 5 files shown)


jpmattia:
I'm fine with the changes. I don't know if the automated tests have error coverage, but I'll be able to manually check them later as part of sckipc work.

Also: From where I sit, there's no need to backport to 3.2.  I started there only because I'm still learning about the PR process.

PR #24812, opened 2024-09-16
Avoid trailing space with wx-config --c[xx]flags

tobil4sk:
When WXCONFIG_C[XX]FLAGS is empty, this ends up being: `echo $._cppflags ""` which results in a trailing space being printed. This is a minor issue but it may confuse tools trying to parse the output of wx-config (https://github.com/haxeui/hxWidgets/pull/113).

Removing the quotes would instead give `echo $._cppflags `, which does not print a trailing space.


vadz:
Thanks for the fix! Looks good to me, so I will merge it soon.

tobil4sk:
Thanks!

vadz:
> Thanks!

Thanks to you for fixing this!

PR #24814, opened 2024-09-17
Fix docs generation for datetime with doxygen 1.11.0

swt2c:
It seems that as of the below commit, doxygen changed its handling of parsing backticks in comments such that it now fails to properly generate documentation for the entire datetime.h file.  Fix this by closing the open backtick.

See: https://github.com/doxygen/doxygen/commit/f18767307be20ca8d2ca81f74cc1f3446205282b

### DIFF
--- a/interface/wx/datetime.h
+++ b/interface/wx/datetime.h
@@ -96,7 +96,7 @@ public:
 
         ///@{
         /// zones from GMT (= Greenwich Mean Time): they're guaranteed to be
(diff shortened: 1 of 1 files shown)


swt2c:
Also if you can add this to the queue for 3.2, it would be appreciated as this is needed for wxPython.

vadz:
Thanks for noticing and fixing this! Will merge soon into master and 3.2.

I think I'll update `include/wx/datetime.h` to match too, even if it doesn't matter that much there.

PR #24815, opened 2024-09-17
Fix print preview in high DPI

### LINKED ISSUE #24666: Print preview too small at high DPI

taler21:
Fixes #24666.

### DIFF
--- a/src/common/prntbase.cpp
+++ b/src/common/prntbase.cpp
@@ -1952,13 +1952,8 @@ void wxPrintPreviewBase::CalcRects(wxPreviewCanvas *canvas, wxRect& pageRect, wx
     canvas->GetSize(&canvasWidth, &canvasHeight);
 
     float zoomScale = m_currentZoom / 100.0f;
-    float screenPrintableWidth = zoomScale * m_pageWidth * m_previewScaleX;
-    float screenPrintableHeight = zoomScale * m_pageHeight * m_previewScaleY;
-
-#ifndef wxHAS_DPI_INDEPENDENT_PIXELS
-    screenPrintableWidth *= canvas->GetDPIScaleFactor();
-    screenPrintableHeight *= canvas->GetDPIScaleFactor();
-#endif // !wxHAS_DPI_INDEPENDENT_PIXELS
+    float screenPrintableWidth = zoomScale * canvas->FromDIP(m_pageWidth) * m_previewScaleX;
+    float screenPrintableHeight = zoomScale * canvas->FromDIP(m_pageHeight) * m_previewScaleY;
 
     wxRect devicePaperRect = m_previewPrintout->GetPaperRectPixels();
     wxCoord devicePrintableWidth, devicePrintableHeight;
--- a/src/common/prntbase.cpp
+++ b/src/common/prntbase.cpp
@@ -906,6 +906,7 @@ wxBEGIN_EVENT_TABLE(wxPreviewCanvas, wxScrolledWindow)
     EVT_PAINT(wxPreviewCanvas::OnPaint)
     EVT_CHAR(wxPreviewCanvas::OnChar)
     EVT_IDLE(wxPreviewCanvas::OnIdle)
+    EVT_DPI_CHANGED(wxPreviewCanvas::OnDPIChanged)
     EVT_SYS_COLOUR_CHANGED(wxPreviewCanvas::OnSysColourChanged)
 #if wxUSE_MOUSEWHEEL
     EVT_MOUSEWHEEL(wxPreviewCanvas::OnMouseWheel)
@@ -940,7 +941,7 @@ wxScrolledWindow(parent, wxID_ANY, pos, size, style | wxFULL_REPAINT_ON_RESIZE, 
 
     // Use some reasonable default size for this window, roughly proportional
     // to the paper sheet.
-    SetInitialSize(wxSize(600, 750));
+    SetInitialSize(FromDIP(wxSize(600, 750)));
 }
 
 wxPreviewCanvas::~wxPreviewCanvas()
@@ -982,6 +983,16 @@ void wxPreviewCanvas::OnIdle(wxIdleEvent& event)
     }
 
     s_inIdle = false;
+}
+
+void wxPreviewCanvas::OnDPIChanged(wxDPIChangedEvent& event)
+{
+    if ( m_printPreview )
+    {
+        m_printPreview->WXUpdateOnDPIChanged();
+    }
+
+    event.Skip();
 }
 
 // Responds to colour changes, and passes event on to children.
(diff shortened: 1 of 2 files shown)


src/common/prntbase.cpp
```diff
@@ -1944,6 +1955,11 @@ void wxPrintPreviewBase::CalcRects(wxPreviewCanvas *canvas, wxRect& pageRect, wx
     float screenPrintableWidth = zoomScale * m_pageWidth * m_previewScaleX;
     float screenPrintableHeight = zoomScale * m_pageHeight * m_previewScaleY;
 
+#ifndef wxHAS_DPI_INDEPENDENT_PIXELS
```
vadz, line 1958: Sorry, it bothers me that we still need to check for `wxHAS_DPI_INDEPENDENT_PIXELS` explicitly, this is not supposed to be necessary with the various DIP-related helpers.

AFAICS you do this here because `FromDIP()` doesn't work with float values, but couldn't we just use `FromDIP(m_pageWidth)` and `FromDIP(m_pageHeight)` when computing `screenPrintable{Width,Height}` above instead?

src/common/prntbase.cpp
```diff
@@ -2144,10 +2162,11 @@ bool wxPrintPreviewBase::DrawBlankPage(wxPreviewCanvas *canvas, wxDC& dc)
         shadowOffset, paperRect.height);
 
     // Draw blank page allowing for 1-pixel border AROUND the actual paper
-    dc.SetPen(*wxBLACK_PEN);
+    dc.SetPen(wxPen(*wxBLACK, dc.FromDIP(1), wxPENSTYLE_SOLID));
```
vadz, line 2165: This is pretty minor, but I thought creating `wxPen` with the default width was supposed to do the right thing in any DPI. And I think it would make sense to make this work...

vadz (commented):
Thanks a lot for the fixes, they look good to me, but I'd just like to avoid explicit `wxHAS_DPI_INDEPENDENT_PIXELS` checks if possible. Could this be done as suggested or am I missing something?

src/common/prntbase.cpp
taler21: Changed to use `FromDIP()` as suggested.

src/common/prntbase.cpp
taler21: `*wxBLACK_PEN` is a 1 pixel wide `wxPen` object that is created and cached in `wxStockGDI`.
It doesn't know the DPI of the `wxDC`.
`*wxThePenList->FindOrCreatePen(*wxBLACK, dc.FromDIP(1), wxPENSTYLE_SOLID)` could be used to cache the pen, but I don't think that is necessary in this case.
So I think that for now, creating a temporary `wxPen` whose width depends on DPI is the right thing to do.

vadz:
Thanks for the update, will merge soon.

taler21:
I don't think the changes regarding `wxEVT_DPI_CHANGED` can be backported to 3.2 in an ABI compatible manner.
But the rest of the changes can and should be backported so that the print preview works correctly in high DPI at least ​​for monitors with the same DPI scaling factor (and in system DPI aware applications).

Such a partial backport looks like this:

```patch
diff --git "a/src/common/prntbase.cpp" "b/src/common/prntbase.cpp"
index 8803ef1bc3..57a8f53880 100644
--- "a/src/common/prntbase.cpp"
+++ "b/src/common/prntbase.cpp"
@@ -939,7 +939,7 @@ wxScrolledWindow(parent, wxID_ANY, pos, size, style | wxFULL_REPAINT_ON_RESIZE,
 
     // Use some reasonable default size for this window, roughly proportional
     // to the paper sheet.
-    SetInitialSize(wxSize(600, 750));
+    SetInitialSize(FromDIP(wxSize(600, 750)));
 }
 
 wxPreviewCanvas::~wxPreviewCanvas()
@@ -1940,8 +1940,8 @@ void wxPrintPreviewBase::CalcRects(wxPreviewCanvas *canvas, wxRect& pageRect, wx
     canvas->GetSize(&canvasWidth, &canvasHeight);
 
     float zoomScale = m_currentZoom / 100.0f;
-    float screenPrintableWidth = zoomScale * m_pageWidth * m_previewScaleX;
-    float screenPrintableHeight = zoomScale * m_pageHeight * m_previewScaleY;
+    float screenPrintableWidth = zoomScale * canvas->FromDIP(m_pageWidth) * m_previewScaleX;
+    float screenPrintableHeight = zoomScale * canvas->FromDIP(m_pageHeight) * m_previewScaleY;
 
     wxRect devicePaperRect = m_previewPrintout->GetPaperRectPixels();
     wxCoord devicePrintableWidth, devicePrintableHeight;
@@ -2094,7 +2094,14 @@ bool wxPrintPreviewBase::RenderPage(int pageNum)
 
     if (!m_previewBitmap)
     {
-        m_previewBitmap = new wxBitmap(pageRect.width, pageRect.height);
+        m_previewBitmap = new wxBitmap();
+#ifdef wxHAS_DPI_INDEPENDENT_PIXELS
+        m_previewBitmap->CreateWithDIPSize( pageRect.width, pageRect.height,
+                                            m_previewCanvas->GetDPIScaleFactor() );
+#else
+        m_previewBitmap->Create( pageRect.width, pageRect.height );
+        m_previewBitmap->SetScaleFactor( m_previewCanvas->GetDPIScaleFactor() );
+#endif
 
         if (!m_previewBitmap || !m_previewBitmap->IsOk())
         {
@@ -2132,7 +2139,7 @@ bool wxPrintPreviewBase::DrawBlankPage(wxPreviewCanvas *canvas, wxDC& dc)
     CalcRects(canvas, pageRect, paperRect);
 
     // Draw shadow, allowing for 1-pixel border AROUND the actual paper
-    wxCoord shadowOffset = 4;
+    wxCoord shadowOffset = dc.FromDIP(4);
 
     dc.SetPen(*wxBLACK_PEN);
     dc.SetBrush(*wxBLACK_BRUSH);
@@ -2143,10 +2150,11 @@ bool wxPrintPreviewBase::DrawBlankPage(wxPreviewCanvas *canvas, wxDC& dc)
         shadowOffset, paperRect.height);
 
     // Draw blank page allowing for 1-pixel border AROUND the actual paper
-    dc.SetPen(*wxBLACK_PEN);
+    dc.SetPen(wxPen(*wxBLACK, dc.FromDIP(1), wxPENSTYLE_SOLID));
     dc.SetBrush(*wxWHITE_BRUSH);
-    dc.DrawRectangle(paperRect.x - 1, paperRect.y - 1,
-        paperRect.width + 2, paperRect.height + 2);
+    wxCoord borderOffset = wxRound(dc.GetPen().GetWidth() / 2.0);
+    dc.DrawRectangle(paperRect.x - borderOffset, paperRect.y - borderOffset,
+        paperRect.width + 2*borderOffset, paperRect.height + 2*borderOffset);
 
     return true;
 }

```

PR #24818, opened 2024-09-17
Add tests for wxTranslations passthrough

vslavik:
This adds the missing test to PR #24804. I verified it fails w/o that patch applied.

It explicitly tests passthrough, not macOS handling. Please notice the `WXLANGUAGE` handling modification to warn, but allow, unrecognized codes.

I'd like to test macOS behavior with .lproj folders by overriding the system, not just `WXLANGUAGE`, but don't know how, and doing the env var only risks testing for behavior that the OS no longer has. And because macOS matching is more thoroughly broken that just the passthrough, it's probably best to address that separately.


vadz:
Thanks a lot, will merge soon.

PR #24819, opened 2024-09-17
Update bundled Expat to the latest 2.6.3 upstream version

vadz:
This fixes a few security problems.


dgud:
Can this be backported to 3.2 ? 
Since it seems every security tool available scream about this now.

vadz:
I am afraid the old MSVS versions still supported in 3.2 might be unable to compile this Expat version. But perhaps someone could make a PR to 3.2 with this change to at least test its build on AppVeyor to check this.

PR #24820, opened 2024-09-18
wxWebView improvements

PBfordev:
The code change in `wxWebViewEdge` (the first commit) should be reviewed.

As for the docs, for WebKit2 there is "In Ubuntu the required package name is libwebkit2gtk-4.0-dev and under Fedora it is webkitgtk4-devel." Using Mint 22 Cinnamon, I was unable to obtain 4.0 package, only 4.1 was available; but I know nothing about Linux, so I may have done something wrong.

### DIFF
--- a/interface/wx/webview.h
+++ b/interface/wx/webview.h
@@ -775,19 +775,18 @@ public:
     @class wxWebView
 
     This control may be used to render web (HTML / CSS / JavaScript) documents.
-    It is designed to allow the creation of multiple backends for each port,
-    although currently just one is available. It differs from wxHtmlWindow in
-    that each backend is actually a full rendering engine, Internet Explorer or Edge on MSW and
-    WebKit on macOS and GTK. This allows the correct viewing of complex pages with
-    JavaScript and CSS.
+    It is designed to allow the creation of multiple backends for each port.
+    It differs from wxHtmlWindow in that each backend is actually a full web
+    engine, provided by a 3rd party (e.g., WebView2 on MSW or WebKit on macOS and GTK).
+    This allows the correct viewing of complex pages with JavaScript and CSS.
 
     @section backend_descriptions Backend Descriptions
 
     This class supports using multiple backends, corresponding to different
-    implementations of the same functionality. Under macOS and Unix platforms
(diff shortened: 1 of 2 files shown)


swt2c:
> The code change in `wxWebViewEdge` (the first commit) should be reviewed.
> 
> As for the docs, for WebKit2 there is "In Ubuntu the required package name is libwebkit2gtk-4.0-dev and under Fedora it is webkitgtk4-devel." Using Mint 22 Cinnamon, I was unable to obtain 4.0 package, only 4.1 was available; but I know nothing about Linux, so I may have done something wrong.

`libwebkit2gtk-4.1-dev` is now the recommended version on Ubuntu and on Fedora it is `webkit2gtk4.1-devel`.

PBfordev:
> `libwebkit2gtk-4.1-dev` is now the recommended version on Ubuntu and on Fedora it is `webkit2gtk4.1-devel`.

Thanks for the information. As I wrote, I know nothing about Linux, so sorry for stupid question: Should I change the version from 4 to 4.1, or are there still supported Linux versions/distribution where only 4.0 is available (i.e., leave it alone or mention both).

swt2c:
> > `libwebkit2gtk-4.1-dev` is now the recommended version on Ubuntu and on Fedora it is `webkit2gtk4.1-devel`.
> 
> Thanks for the information. As I wrote, I know nothing about Linux, so sorry for stupid question: Should I change the version from 4 to 4.1, or are there still supported Linux versions/distribution where only 4.0 is available (i.e., leave it alone or mention both).

I guess the recommendation should be to use 4.1 if available but if not, 4.0 works too.  It looks like, for example, Ubuntu 20.04 only has 4.0.

Also, I just noticed there is a 6.0 out now, so I guess we'll have to support that too.  :-)

PBfordev:
> Also, I just noticed there is a 6.0 out now, so I guess we'll have to support that too. :-)

Thanks again, I will leave that part alone, I don't want to change something I know nothing about.

vadz:
The explanations for the code change look good to me, so I'll merge it, even if I didn't actually test it myself.

Thanks!

PR #24825, opened 2024-09-21
Remark FitInside() must be called after adding or removing elements

wh11204:
If you do not call sizer->FitInside() then the scroll bars will not be updated.


vadz:
Thanks, I'll apply this, but we need to also stop mentioning not existing since a very long time (deprecated back in f944aec0cf (removed SetVirtualSizeHints() and related code, 2007-12-15)!) `SetVirtualSizeHints()`, as it's very confusing...

PR #24826, opened 2024-09-22
Fix toast events in wxMSW

vadz:
Replaces #24341.


lanurmi:
Confirmed: this fixes the issue.

PR #24831, opened 2024-09-24
Allow wxDataViewCustomRenderer::GetSize() return 0 in wxGTK

### LINKED ISSUE #24227: wxDataViewCustomRenderer size not applied on wxGTK in some cases.
See a minimal example:

[minimal.cpp.txt](https://github.com/wxWidgets/wxWidgets/files/13991527/minimal.cpp.txt)

`LIB_TREE_RENDERER` returns height 30 in `GetSize()`, but cells are not sized to height 30 in `Render()`.
Instead, they seem to use height 16 (font: Ubuntu 10).

If DPI is changed in system settings (**Appearance - Fonts - DPI - Custom DPI setting** in Mint), then cells are updated to height 30, but only the ones that are visible.

If you scroll down, you can see other cells still with height 16.

wx is built with `DEBUG_RENDER_EXTENTS` on screenshots.

Expected is for all cells to have height 30.

Initial state:
image

After changing DPI:
image

Scroll down:
image

- wxWidgets version  you use: 3.2; master
- OS  and its version: Linux Mint 21.1
  
  + GTK version: 3.24.33-1ubuntu2
  + Desktop environment : XFCE
  + Current theme:  Mint-Y-Aqua

vadz:
Returning 0 width even once from the overridden GetSize() function in wxDataViewCustomRenderer prevented the height returned by it from being taken into account at all by the GtkTreeView, probably due to some weird caching going on inside GTK.

Work around this by ensuring that both size components are always strictly positive.

Closes #24227.

### DIFF
--- a/src/gtk/dataview.cpp
+++ b/src/gtk/dataview.cpp
@@ -1564,6 +1564,15 @@ gtk_wx_cell_renderer_get_size (GtkCellRenderer *renderer,
 
     wxSize size = cell->GetSize();
 
+    // Somehow returning 0 or negative width prevents the returned height from
+    // being taken into account at all, even if we return strictly positive
+    // width from later calls to GetSize(), meaning that it's enough to return
+    // 0 from it once to completely break the layout for the entire lifetime of
+    // the control.
+    //
+    // As this is completely unexpected, forcefully prevent this from happening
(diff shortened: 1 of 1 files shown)


vadz:

Issue #24625, opened 2024-06-19
Very thick wxSearchCtrl border in the dark mode


DueTo250:
wxSearchCtrl  have a very thick border in the dark mode. Also, the magnifier icon looks so fat
![search-box-dark](https://github.com/wxWidgets/wxWidgets/assets/173275846/5234e6b4-cfe8-4146-ac97-9257559ad044)

OS: Windows 11 23H2

vadz:
The first issue is addressed by the PR above. The second one is more difficult because the current icon drawing code is pretty impenetrable and can't be easily adjusted to high DPI. I think the best would be to get rid of it entirely and use fixed icons for this control instead, e.g. here is a [nice zoom icon](https://uxwing.com/search-line-icon/) and here is a [cancel icon](https://uxwing.com/close-round-line-icon/) similar to what we have now.

vadz:
As we have a separate issue (referenced above) for the second problem, I've finally closed this after merging the PR fixing the first one.

PR #24832, opened 2024-09-25
Don't use themed borders in dark mode in wxMSW

vadz:
They don't look good and while they're not seen often, as many classes already override CanApplyThemeBorder() to return false, when they are, as for wxSearchCtrl, they appear very out of place in the dark mode UI.

See #24625.

----

I'd like to ask people to test this PR before it's merged and let me know if they see any visual regressions due to it in dark mode. AFAICS this can only improve things, but perhaps I'm forgetting about some weird controls which do look good with themed border in dark mode and don't with simple one. Please let me know if you find any!


vadz:

PR #24833, opened 2024-09-25
Remove unused wxString variables

lanurmi:
A few unused were left under tests/, because they may exist for readability. The second commit (which could be squashed, or removed) removes them, too.

There were no cases where the unused variable obviously should have been used for something, though the lines in [tests/controls/propgridtest.cpp](https://github.com/wxWidgets/wxWidgets/compare/master...lanurmi:rm-unused-wxstring?expand=1#diff-ce81c0a061e3e7632d89dbac19ed3b0434eaf33743ac9355c34ddefe957ad29f) are suspicious.

The unused `wxString`s were found by adding the `[[gnu:warn_unused]]` attribute to the `wxString` class. The attribute is not a part of this PR, but something to consider for `wxString` and some other classes.

### DIFF
--- a/include/wx/defs.h
+++ b/include/wx/defs.h
@@ -282,13 +282,13 @@ typedef short int WXTYPE;
 /* wxWARN_UNUSED is used as an attribute to a class, stating that unused instances
    should be warned about (in case such warnings are enabled in the first place) */
 
-#ifdef __has_attribute
-#if __has_attribute(warn_unused)
-#define wxWARN_UNUSED __attribute__((warn_unused))
-#endif
+#ifdef __has_cpp_attribute
+    #if __has_cpp_attribute(warn_unused)
+        #define wxWARN_UNUSED __attribute__((warn_unused))
+    #endif
 #endif
 #ifndef wxWARN_UNUSED
-#define wxWARN_UNUSED
+    #define wxWARN_UNUSED
 #endif
 
 /* these macros are obsolete, use the standard C++ casts directly now */
--- a/include/wx/defs.h
+++ b/include/wx/defs.h
@@ -279,6 +279,17 @@ typedef short int WXTYPE;
     #define wxNODISCARD
 #endif
 
+/* wxWARN_UNUSED is used as an attribute to a class, stating that unused instances
+   should be warned about (in case such warnings are enabled in the first place) */
+
+#ifdef __has_attribute
+#if __has_attribute(warn_unused)
+#define wxWARN_UNUSED __attribute__((warn_unused))
+#endif
+#endif
+#ifndef wxWARN_UNUSED
+#define wxWARN_UNUSED
+#endif
 
 /* these macros are obsolete, use the standard C++ casts directly now */
 #define wx_static_cast(t, x) static_cast<t>(x)
--- a/include/wx/string.h
+++ b/include/wx/string.h
@@ -355,7 +355,7 @@ private:
 };
 #endif // wxUSE_UNICODE_UTF8
 
-class WXDLLIMPEXP_BASE wxString
+class WXDLLIMPEXP_BASE wxWARN_UNUSED wxString
 {
   // NB: special care was taken in arranging the member functions in such order
   //     that all inline functions can be effectively inlined, verify that all
--- a/interface/wx/defs.h
+++ b/interface/wx/defs.h
@@ -1860,6 +1860,16 @@ template <typename T> void wxDELETEA(T*& array);
 #define wxSUPPRESS_GCC_PRIVATE_DTOR_WARNING(name)
 
 /**
+    Expands to the "warn_unused" attribute (also known as [[gnu::warn_unused]])
+    on compilers supporting it (GCC and Clang), otherwise to nothing.
+
+    @header{wx/defs.h}
+
+    @since 3.3.0
+*/
+#define wxWARN_UNUSED [[gnu::warn_unused]]
+
+/**
     Swaps the contents of two variables.
 
     This is similar to std::swap() but can be used even on the platforms where
--- a/interface/wx/defs.h
+++ b/interface/wx/defs.h
@@ -1867,7 +1867,7 @@ template <typename T> void wxDELETEA(T*& array);
 
     @since 3.3.0
 */
-#define wxWARN_UNUSED [[gnu::warn_unused]]
+#define wxWARN_UNUSED __attribute__((warn_unused))
 
 /**
     Swaps the contents of two variables.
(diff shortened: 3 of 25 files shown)


vadz (approved):
Thanks!

I'll merge this soon, but I'd like to add the attribute to `wxString`, it looks useful. What are the drawbacks of doing it, if any?

lanurmi:
> Thanks!
> 
> I'll merge this soon, but I'd like to add the attribute to `wxString`, it looks useful. What are the drawbacks of doing it, if any?

I'm not an expert on attributes, but I'm not aware of drawbacks other than the fact that there isn't a similar attribute for MSVC (that I know of).

vadz:
Let's add this attribute then, after wrapping it in some suitable macro (`wxWARN_UNUSED` looks like an obvious choice).

Would you like to add it to this PR or should I merge it and let you create a new one later? TIA!

lanurmi:
I have now added it to the PR, enclosed in
```
#if defined(__clang__) || defined(__GNUC__)
```
If there is an older Clang or GCC version that supports attributes but not that particular attribute, I'm not sure whether it will issue a warning about it.

NB: The check is indeed for `__GNUC__`, not `__GNUCC__` like a few lines earlier for `wxDISCARD`. `__GNUCC__` does not appear to be defined in my GCC.

vadz:
Hmm, using the attribute in this position doesn't seem to work (see the CI failures).

`__GNUCC__` is a typo, thanks for noticing!

lanurmi:
> Hmm, using the attribute in this position doesn't seem to work (see the CI failures).

That seems to be a problem with Ubuntu, but many other builds (Mac, FreeBSD) failed because of having `-Werror,-Wunused-but-set-variable` enabled, and actually finding a few new unused variables thanks to the attribute. Msw-msys2 similary failed due to finding unused variables in MSW-specific code. I'll fix these.

lanurmi:
After embarassingly many iterations, the CI failures are finally gone, except for AppVeyor. Checking for compiler versions is not a 100% foolproof way to ensure the attribute is only used when supported, though.

paulcor:
What about using `__has_attribute`? This seems to work for GCC 14 and Clang 18:
```
#ifdef __has_attribute
#if __has_attribute(warn_unused)
    #define wxWARN_UNUSED __attribute__((warn_unused))
#endif
#endif
#ifndef wxWARN_UNUSED
    #define wxWARN_UNUSED
#endif
```

lanurmi:
> What about using `__has_attribute`? This seems to work for GCC 14 and Clang 18:

That's an excellent idea, thanks. Applied.

This PR is now probably completed, but I'll do some squashing and rewording later today.

include/wx/defs.h
```diff
@@ -279,6 +279,17 @@ typedef short int WXTYPE;
     #define wxNODISCARD
 #endif
 
+/* wxWARN_UNUSED is used as an attribute to a class, stating that unused instances
+   should be warned about (in case such warnings are enabled in the first place) */
+
+#ifdef __has_attribute
```
vadz, line 285: Why do we use `__has_attribute` here and not `__has_cpp_attribute` that we use everywhere else? Is there really some difference between them (in C++ code, that is)? If not, I'd prefer to use the latter for consistency.

include/wx/defs.h
```diff
@@ -279,6 +279,17 @@ typedef short int WXTYPE;
...
+#if __has_attribute(warn_unused)
+#define wxWARN_UNUSED __attribute__((warn_unused))
+#endif
+#endif
+#ifndef wxWARN_UNUSED
+#define wxWARN_UNUSED
```
vadz, line 291: Also, this is really nitpicky, but I'd indent this block as we do elsewhere, i.e.

```suggestion
    #if __has_attribute(warn_unused)
        #define wxWARN_UNUSED __attribute__((warn_unused))
     #endif
#endif
#ifndef wxWARN_UNUSED
    #define wxWARN_UNUSED

```

vadz (commented):
Thanks again, I'd like to merge this but I'd prefer to use `__has_cpp_attribute` unless there is some reason not to — please let me know if there is!

include/wx/defs.h
paulcor: I suggested `__has_attribute` because `include/wx/defs.h` is a C file, and `__has_cpp_attribute` is not available in C mode with Clang.

vadz: Understood, but I don't think we care about using this in C code (we are not going to do it ourselves and we won't be adding it to 3rd party C code we use), so using C++ check still seems better to me.

Thanks for updating this, Lauri, I'll merge it soon.

lanurmi:
Thinking of 3.2: this obviously cannot be backported as-is, but I wonder if it could and should be backported as an opt-in feature. I.e. defining something like `wxOPT_IN_USE_WARN_UNUSED` would be needed to actually enable it in 3.2.

vadz:
We could define `wxWARN_UNUSED` in 3.2 too, of course. As for optionally using it with `wxString`, I'm not really sure if it's worth it, I don't think many people would enable it.

We could take the risk of doing it by default as, after all, this results in just a warning and not an error, and providing `wxDISABLE_UNUSED` or something like this, but this might be too disruptive... As I'm not sure, I'm probably not going to do anything.

lanurmi:
> We could define `wxWARN_UNUSED` in 3.2 too, of course. As for optionally using it with `wxString`, I'm not really sure if it's worth it, I don't think many people would enable it.

(I would.)

> We could take the risk of doing it by default as, after all, this results in just a warning and not an error, and providing `wxDISABLE_UNUSED` or something like this, but this might be too disruptive...

The biggest issue with enabling it by default in 3.2 is that any project being built with `-Werror=unused-variable` or similar could potentially fail to build because of unused `wxString`s that weren't a problem before. With 3.3 I consider this acceptable and indeed *desirable*, but in 3.2.x it would break too many builds after a minor upgrade.

vadz:
I wouldn't mind merging a version of this PR adding `wxWARN_UNUSED` disabled by default for `wxString`, but I really think that except you there would be very few people who ever use it. But, again, if you make one, it doesn't cost much to apply it.

The only question is about the name of the symbol to use to activate it. For consistency with the existing ones, it should probably be `wxNO_Something`, e.g. `wxNO_UNUSED_VARIABLES`.

lanurmi:
> Thanks again, I'd like to merge this but I'd prefer to use `__has_cpp_attribute` unless there is some reason not to — please let me know if there is!

I didn't know it at the time, but apparently `__has_cpp_attribute(warn_unused)` evaluates to false with Clang, whereas `__has_attribute` does the right thing. See PR #24907.

oneeyeman1:
@lanurmi ,
Maybe its a problem with particular version of Clang?

Thank you.

PR #24837, opened 2024-09-25
Some wxQt fixes

AliKet:
This is a preparation for upcoming commits that allow running CI tests under `Appveyor` for `wxQt`.


oneeyeman1:
@AliKet ,
It is possible to implement max length on multiline text control.

Will you plan to do that?

I'd love to but my Qt machine is really old and slow.

Thank you.

src/qt/listctrl.cpp
```diff
@@ -1380,7 +1370,13 @@ bool wxListCtrl::Create(wxWindow *parent,
 
 wxListCtrl::~wxListCtrl()
 {
-    m_model->deleteLater();
+    // To safely remove and delete the model associated with QTreeView.
+    ClearAll();
+
+    auto sm = GetQListTreeWidget()->selectionModel();
```
vadz, line 1376: Could use `unique_ptr<>` instead of explicit `delete` below.

vadz (commented):
Thanks!

No real comments, but I'd like to run the tests on GitHub rather than on AppVeyor, as the latter is really, really slow.

AliKet:
> @AliKet , It is possible to implement max length on multiline text control.
> 
I think it's doable. but we have to do it ourselves though, because there is no  API for it if you look at the `QTextEdit` documentation.

> Will you plan to do that?
>
Yes, I'll try to implement it after I finish this.

Regards.

oneeyeman1:
@AliKet ,
Yes, there is no API fir that - only single line has it.

And thank you. Hopefully it will be done soon.

AliKet:
@vadz @dsa-t I tried to access the resource file on Windows with the `:/` prefix as described in the **Qt** docs, but to no avail! i.e. something like this: `QIcon icon(":/sample");` will simply return an invalid icon, although the next test succeeds: `wxASSERT( !icon.isNull() );`
It seems that this issue is not new, (see [here](https://stackoverflow.com/questions/6523039/qt-application-and-window-icon-under-windows)) and the commit [1cd623a](https://github.com/wxWidgets/wxWidgets/pull/24837/commits/1cd623a3778167ae248c1d36c7ac77caf19d3405) looks like the right thing to do here.

AliKet:
> 
> No real comments, but I'd like to run the tests on GitHub rather than on AppVeyor, as the latter is really, really slow.

Yes, but, unfortunately, `gridtest` crashes on GitHub Actions and I have no idea why! or how to debug it, any hint or suggestion ?

```
test 2
    Start 2: test_gui

2: Test command: D:\a\wxWidgets\wxWidgets\build_cmake\lib\vc_x64_dll\test_gui.exe
2: Working Directory: D:/a/wxWidgets/wxWidgets/build_cmake/lib/vc_x64_dll
2: Test timeout computed to be: 10000000
2: Test program for wxWidgets GUI features
2: build: 3.3.0 (wchar_t,Visual C++ 1900,STL containers,compatible with 3.2)
2: compiled using msvc 1941 (full: 194134120)
2: running under Windows Server 2022 (build 20348), 64-bit edition as runneradmin
2: 
2: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2: test_gui.exe is a Catch v2.13.9 host application.
2: Run with -? for options
2: 
2: -------------------------------------------------------------------------------
2: Grid::LabelClick
2: -------------------------------------------------------------------------------
2: D:\a\wxWidgets\wxWidgets\tests\controls\gridtest.cpp(602)
2: ...............................................................................
2: 
2: D:\a\wxWidgets\wxWidgets\tests\controls\gridtest.cpp(610): FAILED:
2:   {Unknown expression after the reported line}
2: due to a fatal error condition:
2:   SIGSEGV - Segmentation violation signal
2: 
2: ===============================================================================
2: test cases:   102 |   101 passed | 1 failed
2: assertions: [179](https://github.com/AliKet/wxWidgets/actions/runs/11085949125/job/30802932337?pr=41#step:11:180)39 | 17938 passed | 1 failed
2: 
2/3 Test #2: test_gui .........................***Exception: SegFault 15.98 sec
build: 3.3.0 (wchar_t,Visual C++ 1900,STL containers,compatible with 3.2)
compiled using msvc 1941 (full: 194134120)
running under Windows Server 2022 (build 20348), 64-bit edition as runneradmin
```

AliKet:
<details>
<summary>Patch to enable running CI tests for wxQt under GitHub Actions</summary>

```
diff --git a/.github/workflows/ci_cmake.yml b/.github/workflows/ci_cmake.yml
index f9b567d42d..542ad394ce 100644
--- a/.github/workflows/ci_cmake.yml
+++ b/.github/workflows/ci_cmake.yml
@@ -66,15 +66,18 @@ jobs:
             runner: ubuntu-22.04
             cmake_generator: Unix Makefiles
             cmake_samples: ALL
+            cmake_build_toolkit: gtk3
           - name: macOS 12 wxOSX
             runner: macos-12
             cmake_generator: Xcode
+            cmake_build_toolkit: osx_cocoa
           - name: macOS 12 wxIOS
             runner: macos-12
             cmake_generator: Xcode
             cmake_defines: -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_FIND_ROOT_PATH=/usr/local -DCMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED=NO
             cmake_samples: OFF
             cmake_tests: OFF
+            cmake_build_toolkit: osx_iphone
           - name: Windows MSVC
             runner: windows-latest
             no_sudo: 1
@@ -82,6 +85,15 @@ jobs:
             cmake_generator: Ninja
             cmake_samples: SOME
             cmake_tests: CONSOLE_ONLY
+            cmake_build_toolkit: msw
+          - name: Windows wxQt
+            runner: windows-latest
+            no_sudo: 1
+            cmake_defines: -DCMAKE_C_COMPILER=cl.exe -DCMAKE_CXX_COMPILER=cl.exe
+            cmake_generator: Ninja
+            cmake_samples: SOME
+            cmake_tests: ALL
+            cmake_build_toolkit: qt
 
     env:
       wxGTK_VERSION: ${{ matrix.gtk_version && matrix.gtk_version || 3 }}
@@ -116,6 +128,11 @@ jobs:
           if [ -z $cmake_samples ]; then cmake_samples=SOME; fi
           echo wxCMAKE_SAMPLES=$cmake_samples >> $GITHUB_ENV
 
+          cmake_build_toolkit=${{ matrix.cmake_build_toolkit }}
+          echo wxCMAKE_BUILD_TOOLKIT=$cmake_build_toolkit >> $GITHUB_ENV
+
+          echo QT_USE_NATIVE_WINDOWS=1 >> $GITHUB_ENV
+
           # Setting this variable suppresses "Error retrieving accessibility bus address"
           # messages from WebKit tests that we're not interested in.
           echo NO_AT_BRIDGE=1 >> $GITHUB_ENV
@@ -129,12 +146,26 @@ jobs:
         if: runner.os == 'Windows'
         uses: wxWidgets/gha-setup-vsdevenv@wx
 
+      - name: Install Qt5
+        if: runner.os == 'Windows'
+        uses: jurplel/install-qt-action@v4
+        with:
+          version: '5.15.2'
+          host: 'windows'
+          target: 'desktop'
+          arch: 'win64_msvc2019_64'
+          dir: '${{ github.workspace }}/qt5/'
+          install-deps: 'true'
+          archives: 'qtbase qtsvg qtimageformats'
+          cache: 'true'
+          setup-python: 'false'
+
       - name: Configuring
         run: |
           cmake --version
           mkdir build_cmake
           pushd build_cmake
-          cmake -G "${{ matrix.cmake_generator }}" ${{ matrix.cmake_defines }} -DwxBUILD_SAMPLES=$wxCMAKE_SAMPLES -DwxBUILD_TESTS=$wxCMAKE_TESTS ..
+          cmake -G "${{ matrix.cmake_generator }}" ${{ matrix.cmake_defines }} -DwxBUILD_SAMPLES=$wxCMAKE_SAMPLES -DwxBUILD_TESTS=$wxCMAKE_TESTS -DwxBUILD_TOOLKIT=$wxCMAKE_BUILD_TOOLKIT ..
 
       - name: Building
         working-directory: build_cmake
@@ -165,7 +196,7 @@ jobs:
 
           httpbin_launch
 
-          ctest -V -C Debug --output-on-failure --interactive-debug-mode 0 . || rc=$?
+          ctest -V -C Debug -E test_drawing --output-on-failure --interactive-debug-mode 0 . || rc=$?
           if [ -n "$rc" ]; then
               httpbin_show_log
               exit $rc
```
</details>

vadz:
> Yes, but, unfortunately, `gridtest` crashes on GitHub Actions and I have no idea why! or how to debug it, any hint or suggestion ?

Sorry, no magic solution. Under Linux you could connect via ssh and debug using gdb, but this is not going to work with Windows, so the only thing I can recommend is to simplify the test (e.g. remove the sections one by one) until you can localize where it crashes better.

Another thing which could work would be to create a mini dump file that could be then debugged locally, using either our own `wxDebugReport` or [procdump](https://learn.microsoft.com/en-us/sysinternals/downloads/procdump).

vadz:
Looks good to me, so merging, thanks again for your work on this!

src/qt/window.cpp
```diff
@@ -1510,14 +1518,21 @@ bool wxWindowQt::QtHandleWheelEvent ( QWidget *WXUNUSED( handler ), QWheelEvent
     wxMouseEvent e( wxEVT_MOUSEWHEEL );
 #if (QT_VERSION >= QT_VERSION_CHECK(5, 15, 0))
     QPoint qPt = event->position().toPoint();
+    wxMouseWheelAxis wheelAxis = event->angleDelta().y() > 0
```
dsa-t, line 1521: Seems to be missing abs().

Also I wonder if both x and y could be non-zero. In this case maybe it should send two events similar to https://github.com/wxWidgets/wxWidgets/issues/24516

(repeating because apparently my previous comment was not visible)

AliKet: > Seems to be missing abs().
> 
> Also I wonder if both x and y could be non-zero. In this case maybe it should send two events similar to #24516
> 

Can you please test this patch (which is, unfortunately, for Qt 5.15+ only) and tell me if it works properly for you (TIA!):

```
index 0a5ff12172..ed6655b1af 100644
--- a/src/qt/window.cpp
+++ b/src/qt/window.cpp
@@ -1516,27 +1516,42 @@ bool wxWindowQt::QtHandleResizeEvent ( QWidget *WXUNUSED( handler ), QResizeEven
 bool wxWindowQt::QtHandleWheelEvent ( QWidget *WXUNUSED( handler ), QWheelEvent *event )
 {
     wxMouseEvent e( wxEVT_MOUSEWHEEL );
+    e.m_linesPerAction = 3;
+    e.m_wheelDelta = 120;
+    e.SetEventObject(this);
+
 #if (QT_VERSION >= QT_VERSION_CHECK(5, 15, 0))
-    QPoint qPt = event->position().toPoint();
-    wxMouseWheelAxis wheelAxis = event->angleDelta().y() > 0
-                               ? wxMOUSE_WHEEL_VERTICAL : wxMOUSE_WHEEL_HORIZONTAL;
-    int wheelRotation = wheelAxis == wxMOUSE_WHEEL_VERTICAL
-                      ? (event->angleDelta().y() / 8) : (event->angleDelta().x() / 8);
+    e.SetPosition( wxQtConvertPoint(event->position().toPoint()) );
+
+    bool processed_X = false,
+         processed_Y = false;
+
+    if ( event->angleDelta().y() != 0 )
+    {
+        e.m_wheelAxis = wxMOUSE_WHEEL_VERTICAL;
+        e.m_wheelRotation = event->angleDelta().y() / 8;
+
+        processed_Y = ProcessWindowEvent( e );
+    }
+
+    if ( event->angleDelta().x() != 0 )
+    {
+        e.m_wheelAxis = wxMOUSE_WHEEL_HORIZONTAL;
+        e.m_wheelRotation = event->angleDelta().x() / 8;
+
+        processed_X = ProcessWindowEvent( e );
+    }
+
+    return processed_X || processed_Y;
 #else
-    QPoint qPt = event->pos();
-    wxMouseWheelAxis wheelAxis = event->orientation() == Qt::Vertical
-                               ? wxMOUSE_WHEEL_VERTICAL : wxMOUSE_WHEEL_HORIZONTAL;
-    int wheelRotation = event->delta();
-#endif
-    e.SetPosition( wxQtConvertPoint( qPt ) );
-    e.SetEventObject(this);
+    e.SetPosition( wxQtConvertPoint(event->pos()) );
 
-    e.m_wheelAxis = wheelAxis;
-    e.m_wheelRotation = wheelRotation;
-    e.m_linesPerAction = 3;
-    e.m_wheelDelta = 120;
+    e.m_wheelAxis = event->orientation() == Qt::Vertical
+                  ? wxMOUSE_WHEEL_VERTICAL : wxMOUSE_WHEEL_HORIZONTAL;
+    e.m_wheelRotation = event->delta();
 
     return ProcessWindowEvent( e );
+#endif
 }
 
 

```

dsa-t: Thanks,
Yes, it works with https://github.com/p2rkw/xf86-input-mtrack, athrough I was unable to receive both non-zero x and y in `angleDelta()` at the same time.


Mouse wheel events come through correctly with same values as wxGTK.

In `scroll` sample, horizontal scrolling works in inverse direction compared to wxGTK but that's a different issue.

Issue #24695, opened 2024-07-11
wxInfoBarGeneric hides the button if the width is small enough.


dsa-t:
#### To Reproduce:

1. Open `dialogs` sample
2. Click on Dialogs -> Informative Dialogs -> Simple info bar
3. Reduce the window width
4. The button is not visible anymore:
![image](https://github.com/wxWidgets/wxWidgets/assets/37658952/73857e10-959f-4664-8b19-b08671e3cadd)

---

1. Open `dialogs` sample
2. Click on Dialogs -> Informative Dialogs -> Simple info bar with wrapped text
3. The button is not visible:
![image](https://github.com/wxWidgets/wxWidgets/assets/37658952/4e82056f-0a3c-4bb1-8942-cbc39f8a7fa0)

---

Patch to make it work:

```patch
diff --git a/src/generic/infobar.cpp b/src/generic/infobar.cpp
index 1255d8961e..25f69a67d6 100644
--- a/src/generic/infobar.cpp
+++ b/src/generic/infobar.cpp
@@ -87,8 +87,7 @@ bool wxInfoBarGeneric::Create(wxWindow *parent, wxWindowID winid)
     //     and being preceded by a spacer
     wxSizer * const sizer = new wxBoxSizer(wxHORIZONTAL);
     sizer->Add(m_icon, wxSizerFlags().Centre().Border());
-    sizer->Add(m_text, wxSizerFlags().Centre());
-    sizer->AddStretchSpacer();
+    sizer->Add(m_text, wxSizerFlags().Centre().Proportion(1));
     sizer->Add(m_button, wxSizerFlags().Centre().Border());
     SetSizer(sizer);
 
@@ -233,7 +232,16 @@ void wxInfoBarGeneric::ShowMessage(const wxString& msg, int flags)
     // notice the use of EscapeMnemonics() to ensure that "&" come through
     // correctly
     m_text->SetLabel(wxControl::EscapeMnemonics(msg));
-    m_text->Wrap( GetClientSize().GetWidth() );
+
+    int width = GetClientSize().GetWidth();
+
+    if ( m_icon )
+        width -= m_icon->GetBestSize().y + wxSizerFlags::GetDefaultBorder() * 2;
+
+    if ( m_button )
+        width -= m_button->GetBestSize().y + wxSizerFlags::GetDefaultBorder() * 2;
+
+    m_text->Wrap(width);
 
     // then show this entire window if not done yet
     if ( !IsShown() )

```

[...]

PR #24838, opened 2024-09-25
Fix showing messages in narrow info bars

vadz:
See #24695.

### DIFF
--- a/src/generic/infobar.cpp
+++ b/src/generic/infobar.cpp
@@ -235,6 +235,9 @@ void wxInfoBarGeneric::ShowMessage(const wxString& msg, int flags)
     // use SetLabelText() to ensure that "&" come through correctly
     m_text->SetLabelText(msg);
 
+    // in case it doesn't fit in the window, show the full message as a tooltip
+    m_text->SetToolTip(msg);
+
     // then show this entire window if not done yet
     if ( !IsShown() )
     {
--- a/src/generic/infobar.cpp
+++ b/src/generic/infobar.cpp
@@ -74,7 +74,9 @@ bool wxInfoBarGeneric::Create(wxWindow *parent, wxWindowID winid)
     // the icon is not shown unless it's assigned a valid bitmap
     m_icon = new wxStaticBitmap(this, wxID_ANY, wxNullBitmap);
 
-    m_text = new wxStaticText(this, wxID_ANY, wxString());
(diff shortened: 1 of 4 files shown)


src/generic/infobar.cpp
```diff
@@ -84,11 +84,9 @@ bool wxInfoBarGeneric::Create(wxWindow *parent, wxWindowID winid)
     // button at the very right
     //
     // NB: AddButton() relies on the button being the last control in the sizer
-    //     and being preceded by a spacer
     wxSizer * const sizer = new wxBoxSizer(wxHORIZONTAL);
     sizer->Add(m_icon, wxSizerFlags().Centre().Border());
-    sizer->Add(m_text, wxSizerFlags().Centre());
-    sizer->AddStretchSpacer();
```
dsa-t, line 91: I think `sizer->AddStretchSpacer();` should be left in because some user code (KiCad's) expects to find the spacer to remove buttons after it: https://gitlab.com/kicad/code/kicad/-/blob/8128f4ebd96b45ea517042c6b364e883de7e4ab4/common/widgets/wx_infobar.cpp#L320

vadz: Ugh, KiCad really shouldn't have depended on this internal implementation detail.

@marekr Can you change this code or do we really have to preserve this sizer? I guess if we do, we'll have to document this as part of this class API.

dsa-t: It would cause problems for people building older KiCad versions with latest wxWidgets, when bisecting for example.

marekr: > It would cause problems for people building older KiCad versions with latest wxWidgets, when bisecting for example.

Meh? boost 1.86 already breaks bisects for us. Lol.

vadz: Oh well, I've restored the spacer.

BTW, we definitely could add `RemoveAllButtons()` to `wxInfoBar` itself instead of reimplementing it in KiCad.

imciner2: IMO, if a way to remove all the buttons is added into the generic infobar in wx, then it is fine to remove the spacer for that version of wx - we can just conditionalize the code for this (we have done that a lot in the past anyway, so it isn't a big deal).

vadz:
I'm merging this for now, let's discuss removing the buttons in the issue I've created for it.

Issue #24698, opened 2024-07-12
SVG icons and child windows (controls) are scaled incorrectly under Windows on HiDPI displays


hatelamers:
#### Bug description:

When a wxWidgets app is compiled with `#define wxUSE_DPI_AWARE_MANIFEST 2` and run on HiDPI display with custom scaling (f.i. 200%) all SVG icons loaded from XRC have wrong sizes in menus, data views, on buttons, tabs etc.  Also all child windows (resizeable or restored with `wxPersistenceManager::Get().RegisterAndRestore(control)`) have wrong relative dimensions comparing to 100% scale. As result the GUI looks broken.

#### Expected behaviour:
Relative dimensions and proportions of GUI elements on HiDPI monitor with custom scaling should not be different to those on monitors scaled to 100%. Example of "correct" view on 100% scaled display:
![app-100-percent-scale](https://github.com/user-attachments/assets/2d3b7267-d9a7-4d52-a6c4-eaf67f481021) 

#### Observed behaviour:
Apart from fonts and main window size other GUI elements are scaled incorrectly on display with 200% scaling:
![app-200-percent-scale](https://github.com/user-attachments/assets/513c8d83-faa9-4e62-aff4-ca4d090c0002)

#### To Reproduce:
1. Construct a minimal wxWidgets app with one main window loaded from XRC.
2. Add a menu to the XRC with a couple of items with SVG icons (default size: 16)
3. Add wxSplitterWindow containing 2 wxDataViewCtrl elements  (or wxDataViewListCtrl for simplicity) to the XRC.
4. Add `wxPersistenceManager::Get().RegisterAndRestore(splitter);` in initialization code block of the main frame.
5. Insert some items into data views (`wxDataViewCtrl::AppendIconTextColumn(...);`)
6. Compile
7. Run the app on a HiDPI display scaled to 200%.
8. See icons having wrong sizes.
9. Move the splitter.
10. Close the app.
11. Run it again (with 200% scaling)
12. See the splitter on different position from where you moved it previously.

Use attached sample code: [svgtest.zip](https://github.com/user-attachments/files/16363627/svgtest.zip)

### Platform and version information
- wxWidgets version: 3.2.5
- wxWidgets port: wxMSW (compiled from sources)
- OS: Windows 10, 11

oneeyeman1:
@hatelamers ,
Can you make a minimal possible patch to one of the XRC samples and attach it to the ticket?

Thx.

[...]

PR #24840, opened 2024-09-26
Fix size of SVG images loaded from XRC in high DPI

vadz:
Don't scale the default SVG size by the current scale factor, it is DPI-independent and expressed in DIPs, not pixels, so don't use ParseStringInPixels() for it.

See #24698.


vadz:

Issue #24701, opened 2024-07-13
Clipboard problem under KDE neon


unxed:
Steps to reproduce:

1. Download latest KDE neon ISO:
https://neon.kde.org/download
(I used neon-user-20240530-0715.iso for testing)

2. Install system in virtual machine (I used virt-manager & qemu) with default settings

3. Build wxWidgets with default settings

4. Build and run `text` sample

5. Select `Next 10 characters should be in blue` string in `text` sample window

6. Menu, Clipboard, Copy

7. Run Kate text editor

8. Try to paste

Additionally, you can run Konsole terminal emulator and type
`sudo apt install wl-clipboard`
and
`wl-paste -l`
to see the source of a problem:
```
$ wl-paste -l
UTF8_STRING
```

`UTF8_STRING` is x11 atom, but neon uses Wayland, and most apps do not expect such atom under Wayland, it should be `text/plain;charset=utf-8` instead.

unxed:
Touch #23544, #23650, #24444, #24391, #23650

paulcor:
Not KDE specific, I can reproduce a clipboard problem with Weston. There is also a second problem: `FindFocus()` isn't working, which is making the described test using the `text` sample invalid. You can avoid the `FindFocus()` problem by using the top center control in the `text` sample, which has the text starting "Prepended". Copying text from that control with the menu command doesn't work with Wayland, `wl-paste -l` just prints `TIMESTAMP`.

vadz:
For me selecting the text in the control with "Prepended" and using "Clipboard > Copy" from the menu results in

```console
% wl-paste -l         
text/plain;charset=utf-8
UTF8_STRING
```

as expected. This is under Sway and not Weston or Kwin, however.

However I have no idea how can the behaviour of the code depend on the compositor and it's pretty bad news if it does :-( Are you testing with master?

unxed:
> Are you testing with master?

yep

vadz:
I can reproduce it with Weston, looking.

vadz:
It looks like under Weston calling `gtk_selection_add_target()` only works the first time we do it, somehow, which is why only `TIMESTAMP` appears in `wl-paste -l` output: if I remove the call to `AddSupportedTarget(g_timestampAtom)` I do see `UTF8_STRING`, and if I hack the code to add `text/plain;charset=utf-8` instead, I see it too and copying it works, but this is surely not the right thing to do, as it will break something else.

[...]

PR #24845, opened 2024-09-28
Fix copying text to the clipboard under Wayland with some compositors

vadz:
This works with Weston, but still needs to be tested with kwin (see #24701).


vadz:
Damn, sorry, I've pushed this to master accidentally instead of updating the PR :-(

But it does seem to work for me under both Weston and kwin, so let's hope we're going to get more testing for it like this?

PR #24847, opened 2024-09-28
Fix defining wxNODISCARD for gcc

vadz:
Fix the check in f1985c6ba2 (Use [[nodiscard]] with functions returning object to be freed, 2022-11-04) for gcc which didn't work properly due to a typo.

See #22943.


lanurmi:
The same typo in the same context is also present in `src/stc/lexilla/include/LexillaCompat.h`.

PR #24848, opened 2024-09-29
3.2 backports 202310

vadz:


PBfordev:
On wxForum, I encountered SVGs that are rasterized incorrectly in 3.2 but correctly in the master, probably due to difference in bundled nanosvg library.

Is there a reason why nanosvg is not being updated in 3.2 branch?

vadz:
> Is there a reason why nanosvg is not being updated in 3.2 branch?

I don't think there is some special reason, it's just that somebody needs to do it, i.e. cherry pick 19aee81c1d (Update nanosvg to the latest upstream version, 2024-01-14) into 3.2 (should be simple, unless I'm forgetting something) and check that it works there.

lanurmi:
Requesting 795c02c.

PR #24855, opened 2024-10-01
Enhance algorithm to find the best UI translation language

utelle:
In PR https://github.com/wxWidgets/wxWidgets/pull/24804 the current algorithm to determine the best available UI translation language was characterized as being _naive_. As a reference the [Unicode Technical Standard #35](https://unicode.org/reports/tr35/) was mentioned.

This PR implements a (hopefully) more correct algorithm - following the general route of the Unicode language matching algorithm. This is achieved with the following steps:

- enhancing the wxWidgets language database with additional data from the Unicode CLDR repository which are important for the new algorithm - a list of likely subtags, a list of language matches with distances, a list of region groups
- implementing static `wxLocaleIdent` methods to add or remove likely subtags
- implementing a static `wxLocaleIdent` method to get the best match between a list of desired/preferred languages and a list of supported/available languages
- implementing static `wxUILocale` methods to access the additional language data tables
- adding a test case with several checks to verify expected language matches

### Open questions

1) Additional changes needed to be applied to `translation.cpp`?

    Currently method `wxTranslations::DoGetBestAvailableTranslation()` adds the parameter `additionalAvailableLanguage` (which is either empty or corresponds to the language of the message ids) to the list of available languages. However, the new algorithm would expect this language as a separate default language parameter. If no suitable match was found the default language will be returned.

    In the current implementation the internal function `GetPreferredUILanguage()` which in turn calls now `wxLocaleIdent::GetBestMatch()`, doesn't have a parameter for the language of the message ids. My proposal would be to **not** add the parameter `additionalAvailableLanguage` to the list of available languages, but to use it as the default language.

2. How language tags should be handled that are included in the wx language database (and are used for example by Windows), but are not included in the mapping of likely subtags of CLDR?

    For example, Windows uses the tag "**quz**" for the language "_Quechua_". However, the CLDR likely subtags mapping includes only an entry for "**qu**". Should the algorithm remap "**quz**" to "**qu**" internally for adding likely subtags? Or should it ignore this issue and detect only direct matches? The algorithm currently would do the latter.

    Currently, only 4 language tags are affected: **bh**, **quz**, **sh**, **tw**. According to the CLDR data **bh** could be replaced by **hi**, **quz** by **qu**, **tw** by **ak**. **sh** is problematic, because there is no unique mapping - it could be mapped to **bs**, **hr**, or **sr**.

@vadz , @vslavik: I'd like to ask you to review and/or comment on this PR.


utelle:
Regarding CI checks:

- Code Checks / Check Whitespace fails due to tabs and whitespace issues in Unicode CLDR data files. These files should be excluded from the checks.
- Many builds fail due to marking as an error that a variable is set, but unused. Yes, the variable `bestDesired` is not used as an output parameter, but for debugging purposes it is important to know which index of desired languages is used for the current best match. Therefore I'd like to keep this variable. How can I achieve this without causing this crazy error.

lanurmi:
> Many builds fail due to marking as an error that a variable is set, but unused. Yes, the variable `bestDesired` is not used as an output parameter, but for debugging purposes [...] I'd like to keep this variable. How can I achieve this without causing this crazy error.

Just say `(void) bestDesired;` after its definition.

vadz:
> > Many builds fail due to marking as an error that a variable is set, but unused. Yes, the variable `bestDesired` is not used as an output parameter, but for debugging purposes [...] I'd like to keep this variable. How can I achieve this without causing this crazy error.
> 
> Just say `(void) bestDesired;` after its definition.

We have `wxUnusedVar()` for this, which is more readable.

include/wx/uilocale.h
```diff
@@ -14,6 +14,7 @@
 
 #if wxUSE_INTL
 
+#include "wx/arrstr.h"
```
vadz, line 17: Minor, but we're already using `wxVector<wxString>` in this header/class, let's use it instead of `wxArrayString` for the new functions too.

include/wx/uilocale.h
```diff
@@ -40,6 +41,16 @@ class WXDLLIMPEXP_BASE wxLocaleIdent
     // and region separated by dashes.
     static wxLocaleIdent FromTag(const wxString& tag);
 
+    // Add likely subtags to a given locale identifier.
+    static wxLocaleIdent AddLikelySubtags(const wxLocaleIdent& localeIdent);
+
+    // Remove likely subtags from a given locale identifier, favor region.
+    static wxLocaleIdent RemoveLikelySubtags(const wxLocaleIdent& localeIdent, bool favorRegion = true);
```
vadz, line 48: Sorry, but it's always [the same remark](https://www.wxwidgets.org/develop/coding-guidelines/#no_bool_params): please let's avoid functions with non-obvious boolean arguments. I will have *no* idea what `false` means when I see a call `RemoveLikelySubtags(ident, false)`, so this is a bad API

include/wx/uilocale.h
```diff
@@ -238,6 +249,21 @@ class WXDLLIMPEXP_BASE wxUILocale
     static wxString GetScriptAliasFromName(const wxString& scriptName);
     static wxString GetScriptNameFromAlias(const wxString& scriptAlias);
 
+    // These three methods are for internal use only.
```
vadz, line 252: Do they still need to be public or can we move them to some private header?

misc/languages/data/uni_loadregiongroups.sql
```diff
@@ -0,0 +1,145 @@
+create table if not exists uni_regiongroups (language char, country char, primary key(language, country));
```
vadz, line 1: Just to help me understand: was this created by hand or does this information come from somewhere?

I.e. how should this file be maintained/kept up to date?

include/wx/uilocale.h
```diff
@@ -40,6 +41,16 @@ class WXDLLIMPEXP_BASE wxLocaleIdent
     // and region separated by dashes.
     static wxLocaleIdent FromTag(const wxString& tag);
 
+    // Add likely subtags to a given locale identifier.
+    static wxLocaleIdent AddLikelySubtags(const wxLocaleIdent& localeIdent);
```
vadz, line 45: Does this function really need to be public? It isn't documented, so I'm not exactly sure what is it for, but it looks more like an internal helper, will it really be useful to the application code?

src/common/languageinfo.cpp
```diff
@@ -1177,6 +1223,665 @@ tabScriptData[] =
     { nullptr, nullptr }
 };
 
+// Data table for likely subtags
```
vadz, line 1226: I think it would be nicer to move this data out to a separate file and `#include` it from here.

include/wx/uilocale.h
```diff
@@ -40,6 +41,16 @@ class WXDLLIMPEXP_BASE wxLocaleIdent
...
+
+    // Remove likely subtags from a given locale identifier, favor region.
+    static wxLocaleIdent RemoveLikelySubtags(const wxLocaleIdent& localeIdent, bool favorRegion = true);
+
+    // Find the best match between desired and supported languages/locales.
+    static wxString GetBestMatch(const wxArrayString& desired, const wxArrayString& supported, const wxString& defaultTag);
```
vadz, line 51: Do we use non-empty `defaultTag` anywhere? I can't find it in the diff, but perhaps I'm missing something. But if I don't, can we get rid of this argument?

include/wx/uilocale.h
```diff
@@ -40,6 +41,16 @@ class WXDLLIMPEXP_BASE wxLocaleIdent
     // and region separated by dashes.
     static wxLocaleIdent FromTag(const wxString& tag);
 
+    // Add likely subtags to a given locale identifier.
+    static wxLocaleIdent AddLikelySubtags(const wxLocaleIdent& localeIdent);
+
+    // Remove likely subtags from a given locale identifier, favor region.
+    static wxLocaleIdent RemoveLikelySubtags(const wxLocaleIdent& localeIdent, bool favorRegion = true);
+
+    // Find the best match between desired and supported languages/locales.
```
vadz, line 50: This API is error-prone because it's simple to exchange the parameters by mistake. Maybe it's not really important if it's mostly for internal use anyhow and only used in 1 or 2 places, but if we could make it less symmetrical, it would be nice...

src/common/uilocale.cpp
```diff
@@ -275,6 +275,329 @@ wxLocaleIdent wxLocaleIdent::FromTag(const wxString& tag)
...
+            supportedLanguage = supportedLocalesMax[k].GetLanguage();
+            supportedScript = supportedLocalesMax[k].GetScript();
+            supportedRegion = supportedLocalesMax[k].GetRegion();
+            supportedMinDistance = supportedLocalesMinDistance[k];
+
+            // Check for matching language, script and region
+            if (!directMatch && !desiredLanguage.empty() && !supportedLanguage.empty())
+            {
+                // Check language
+                if (supportedLanguage.IsSameAs(desiredLanguage, false))
```
vadz, line 521: Minor, but I find

```suggestion
                if (supportedLanguage.CmpNoCase(desiredLanguage) == 0)
```

much better (or, as already stated above, boolean parameters are really unreadable).

vadz (commented):
Thanks for your work on this, I didn't have time to look at everything, but the code changes look reasonable and the tests seem to conform to my expectations (unless I overlooked something).

I'm worried by several things here:

1. How exactly will we maintain these files? There are Lua scripts, SQL scripts, Python scripts and XSLT (which I'd gladly never touch again in my life) and I'm just scared. Is there some single simple command to run to update all that needs to be updated? Ideally we'd even run it from GitHub workflow periodically to check that it still works (and maybe be notified if there are any updates).
2. I don't fully understand the first question, but I know that we definitely need `additionalAvailableLanguage` as without this, things don't work (but we should have tests that detect this). So I don't understand how can this PR work without them.

Could you please explain this?

vadz:
> Regarding CI checks:
> 
>     * Code Checks / Check Whitespace fails due to tabs and whitespace issues in Unicode CLDR data files. These files should be excluded from the checks.

This can be done by modifying `.github/workflows/code_checks.yml`. I am not sure if the changes in the branch are used when running the check on the PR, if they're not, please make another PR with the changes to this file and I'll merge it to master to make sure it's used for the next workflow execution in this PR. TIA!

utelle:
> > Just say `(void) bestDesired;` after its definition.
> 
> We have `wxUnusedVar()` for this, which is more readable.

@vadz @lanurmi - Thanks for the pointer. This fixes the compile errors. At the time of writing some checks are still running, but most finished successfully already.

> > Regarding CI checks:
> > ```
> > * Code Checks / Check Whitespace fails due to tabs and whitespace issues in Unicode CLDR data files. These files should be excluded from the checks.
> > ```
> 
> This can be done by modifying `.github/workflows/code_checks.yml`. I am not sure if the changes in the branch are used when running the check on the PR, ...

Yes, they are applied fortunately.

utelle:
> Thanks for your work on this, I didn't have time to look at everything, but the code changes look reasonable and the tests seem to conform to my expectations (unless I overlooked something).

Most tests are taken from the Unicode CLDR test suite. So, I hope the new algorithm is suitable in most cases. The description of the algorithm in the Unicode specs is somewhat incomplete regarding some special cases. However, my implementation seems to work for typical real world examples. 

> I'm worried by several things here:
> 
> 1. How exactly will we maintain these files? There are Lua scripts, SQL scripts, Python scripts and XSLT (which I'd gladly never touch again in my life) and I'm just scared.

Understandable. Fortunately, directory `misc/languages` contains almost all items required to regenerate the files actually used in wxWidgets (`include/wx/language.h`, `interface/wx/language.h`, and `src/common/languageinfo.cpp`) under Windows.

> Is there some single simple command to run to update all that needs to be updated?

The file `misc/languages/README` contains a description of necessary steps to regenerate the wx source files. In directory `misc/languages/data` there are several Windows batch command files, which in turn invoke some tools like SQLite CLI, Lua CLI and XSLT CLI.

The most tricky part is the generation of `wxLANGUAGE_*` symbols, because we have to keep compatibility with older wxWidgets versions. If the locale list extracted from Windows 10/11 contains new locale identifiers, it can happen that conflicts show up, requiring manual adjustments. The scripts try their best to resolve conflicts automatically, but it is not always possible, unfortunately.

Things would be much easier, if we could get rid of `wxLANGUAGE_*` symbols. Personally, I have never used those symbols except for `wxLANGUAGE_UNKNOWN` and `wxLANGUAGE_DEFAULT` (and maybe `wxLANGUAGE_ENGLISH`).

> Ideally we'd even run it from GitHub workflow periodically to check that it still works (and maybe be notified if there are any updates).

Yes, that would be certainly useful. Changes can usually be expected, when a new Unicode standard version is released or when a major Windows update occurs (once or twice per year).

At the moment such a check would not work right out of the box, because the required tools are not included in the repository, For some tools binaries can be downloaded. However, for extracting data from XML files (the Unicode CLDR files required for the language matching algorithm are in XML format) I use the Sablotron XSLT processor, for which I haven't found pre-compiled binaries.

> 2. I don't fully understand the first question, but I know that we definitely need `additionalAvailableLanguage` as without this, things don't work (but we should have tests that detect this). So I don't understand how can this PR work without them.

Well, I think the current test cases don't cover the case where the list of preferred languages consists only of languages for which no translation catalogs are available, including English (which is the typical default language).

Currently, the additional language (namely the msgId language) is appended to the list of available translations, but in the new implementation no match will be found, if the msgId language is not among the preferred languages.

The method `wxTranslations::GetBestTranslation()` adds the msgId language to the list of available languages in 2 variants: as given and truncated to the string before the first underscore. In contrast, the method `wxTranslations::GetBestAvailableTranslation()` sets the additional language to an empty string.

> Could you please explain this?

IMHO the current implementation in `translation.cpp` needs to be changed, so that the additional language (representing the msgId language - always?) is used as the default language, if no match with actual available translation catalogs is found.

If a translation catalog is actually available for the msgId language, it will be included in the list of available languages. And if it is the preferred language, a match will be found. Only, if a match was not found, the default language would be returned.

The latter is a bit different than the current implementation. Currently method `wxTranslations::DoAddCatalog()` returns one of the following 3 values:

1. Translations::NotFound - no match was found for the preferred language, not even for the default language
2. Translations::Found - a match was found for an actually existing translation catalog
3. Translations::NotNeeded - a match was found, but a translation catalog could not be loaded - meaning that the default msgId language will be used

If a non-empty default language is specified for the new algorithm, the first case `Translations::NotFound` would no longer occur, because the algorithm would always return a non-empty result - either a found match or the default language.

So, the question is whether we want to distinguish the 3 cases or not. Actually, case 1 and 3 have the same consequence: strings are **not** translated, but the msgId language is used instead.

Currently, the PR reproduces the behaviour as is (3 cases). Using a default language parameter would reduce the number of cases to cases 2 and 3.

include/wx/uilocale.h
utelle: These methods could be moved to some private headers, of course. They only have to be available, wherever they are needed internally.

misc/languages/data/uni_loadregiongroups.sql
utelle: Yes, the table of region groups was created manually. It affects only 5 languages (Arabic, English, Spanish, Portuguese, and Chinese). It is based on Unicode data, but it is not trivial to extract the relevant data from the Unicode files.

My guess is that these lists are quite static and will not change often. However, I will check whether the lists can be generated automatically with reasonabl effort.

src/common/uilocale.cpp
utelle: Ok, I will change the code accordingly.

include/wx/uilocale.h
utelle: > > Do we use non-empty `defaultTag` anywhere?
> 
> Currently not. But this could change, depending on the answer to my [first question](https://github.com/wxWidgets/wxWidgets/pull/24855#issue-2560038878).

I have to correct myself. At the moment, all tests for method `GetBestMatch()` use the default tag parameter. But this could be changed to check for an empty string instead.

vslavik:
> Currently method wxTranslations::DoGetBestAvailableTranslation() adds the parameter additionalAvailableLanguage (which is either empty or corresponds to the language of the message ids) to the list of available languages. However, the new algorithm would expect this language as a separate default language parameter. If no suitable match was found the default language will be returned.

`wxTranslations` is documented to return empty string if no match is found — then the "default language" fallback is realized simply by wxTranslations not loading any translation. IMNSHO this is the right thing to do, and `GetBestMatch` should do the same, i.e. return empty string to indicate that it failed to find a match. There's value in differentiating between failure and intentional/correct use of a fallback.

> In the current implementation the internal function GetPreferredUILanguage() which in turn calls now wxLocaleIdent::GetBestMatch(), doesn't have a parameter for the language of the message ids. My proposal would be to not add the parameter additionalAvailableLanguage to the list of available languages, but to use it as the default language.

Doing this would re-break language matching. If you `git blame` the code involving `additionalAvailableLanguage`, the commits will point to a PR where this was discussed (including the need to fix a broken-by-design new API). Msgid language is and *must* be treated as first-class citizen in language matching. It is not a second-tier fallback, it is proper language in which the application is localized.

See 2af36e9560c694a750b6aa835d32d12a0393fb83 and 1030147aff1fc3ec7f49ed581c976018e82f2cde for explanation and examples.

vslavik:
> There are Lua scripts, SQL scripts, Python scripts and XSLT (which I'd gladly never touch again in my life) and I'm just scared.

FWIW, CLDR publishes the data [in JSON](https://github.com/unicode-org/cldr-json) as well, which is perhaps more readable and doesn't require pulling in the whole XSLT shebang.

vslavik:
> So, the question is whether we want to distinguish the 3 cases or not.

We do. See the commits I mentioned above, loading `wxstd*` catalogs IIRC relies on the distinction (and it too was part of a regression fix).

tests/intl/intltest.cpp
```diff
@@ -526,6 +526,124 @@ TEST_CASE("wxLocaleIdent::FromTag", "[uilocale][localeident]")
...
+    // nearby languages: Danish matches no
+    CHECK(FindBestMatch({ "da" }, { "en", "no" }, "en-GB") == "no");
+    // nearby languages: Nynorsk to Bokmål
+    CHECK(FindBestMatch({ "nn" }, { "en", "nb" }, "en-GB") == "nb");
+    // nearby languages: Danish does not match nn
+    CHECK(FindBestMatch({ "da" }, { "en", "nn" }, "en-GB") == "en-GB");
+
+    // script fallbacks
+    CHECK(FindBestMatch({ "zh-Hant" }, { "zh-CN", "zh-TW" }, "en-GB") == "zh-TW");
+    CHECK(FindBestMatch({ "zh" }, { "zh-CN", "zh-TW" }, "en-GB") == "zh-CN");
```
vslavik, line 604: Please test explicitly `zh-Hans` as an important case too.

tests/intl/intltest.cpp
```diff
@@ -526,6 +526,124 @@ TEST_CASE("wxLocaleIdent::FromTag", "[uilocale][localeident]")
...
+    CHECK(FindBestMatch({ "zh-Hans-CN" }, { "zh-CN", "zh-TW" }, "en-GB") == "zh-CN");
+    CHECK(FindBestMatch({ "zh-Hant-HK" }, { "zh-CN", "zh-TW" }, "en-GB") == "zh-TW");
```
vslavik, line 606: We also need to test with full locales returned by macOS that are "nonsensical", e.g. `zh-Hant-DE` and `zh-Hans-DE`.

tests/intl/intltest.cpp
```diff
@@ -526,6 +526,124 @@ TEST_CASE("wxLocaleIdent::FromTag", "[uilocale][localeident]")
...
+    CHECK(FindBestMatch({ "es-MX" }, { "en", "en-GB", "es-ES", "es-AR" }, "de") == "es-AR");
+    CHECK(FindBestMatch({ "es-PT" }, { "en", "en-GB", "es-ES", "es-AR" }, "de") == "es-ES");
+
+    // best match for traditional chinese
+    CHECK(FindBestMatch({ "zh-TW" }, { "fr", "zh-Hans-CN", "en-US" }, "en-GB") == "en-GB");
+    CHECK(FindBestMatch({ "zh-Hant" }, { "fr", "zh-Hans-CN", "en-US" }, "en-GB") == "en-GB");
+
+    CHECK(FindBestMatch({ "zh-TW", "en" }, {"fr", "zh-Hans-CN", "en-US"}, "en-GB") == "en-US");
+    CHECK(FindBestMatch({ "zh-Hant-CN", "en" }, {"fr", "zh-Hans-CN", "en-US"}, "en-GB") == "en-US");
+    CHECK(FindBestMatch({ "zh-Hans", "en" }, {"fr", "zh-Hans-CN", "en-US"}, "en-GB") == "zh-Hans-CN");
```
vslavik, line 646: All these tests should be (instead of or in addition too) having just plain `zh-Hans` or `zh-Hant` (and also *both*) in the supported list — that is by far the more common situation that having full locales there.

tests/intl/intltest.cpp
```diff
@@ -526,6 +526,124 @@ TEST_CASE("wxLocaleIdent::FromTag", "[uilocale][localeident]")
...
+
+    // script fallbacks
+    CHECK(FindBestMatch({ "zh-Hant" }, { "zh-CN", "zh-TW" }, "en-GB") == "zh-TW");
+    CHECK(FindBestMatch({ "zh" }, { "zh-CN", "zh-TW" }, "en-GB") == "zh-CN");
+    CHECK(FindBestMatch({ "zh-Hans-CN" }, { "zh-CN", "zh-TW" }, "en-GB") == "zh-CN");
+    CHECK(FindBestMatch({ "zh-Hant-HK" }, { "zh-CN", "zh-TW" }, "en-GB") == "zh-TW");
+
+    // language-specific script fallbacks
+    CHECK(FindBestMatch({ "sr" }, { "en", "sr-Latn" }, "en-GB") == "sr-Latn");
+    CHECK(FindBestMatch({ "sr-Cyrl" }, { "en", "sr-Latn" }, "en-GB") == "sr-Latn");
```
vslavik, line 610: Also should test that `sr-Latn` is correctly handled, including (a) on Windows where IIRC you said `sr` is considered `sr-Latn` (while it is `sr-Cyrl` on Linux) and that this correctly handles the situation where the available list has Linux-style `sr` (Cyrillic) and `sr@latin` translation files.

utelle:
> > There are Lua scripts, SQL scripts, Python scripts and XSLT (which I'd gladly never touch again in my life) and I'm just scared.
> 
> FWIW, CLDR publishes the data [in JSON](https://github.com/unicode-org/cldr-json) as well, which is perhaps more readable and doesn't require pulling in the whole XSLT shebang.

In principle, a good and interesting idea, because the SQLite CLI (that is also used in the process of regenerating the language data files) supports JSON. However, I took a quick look at the state of those JSON data files, and unfortunately I have to say that some of the files we need are in a desolate state. For example, the file `languageMatching.json` (which corresponds to `languageInfo.xml`) is completely broken, because important attributes are missing, and the file is therefore unusable. I have not checked the other essential file `likelySubtags.json`. It looks fairly complete, but I haven't verified that yet.

The official CLDR data are distributed in XML format. Therefore I tend to stick with the official XML data. [Sablotron](https://sourceforge.net/projects/sablotron/files/) is a compact tool (Windows executable less than 500 kB). It is old, but worked for me very well. However, I can try to create a small Lua script to extract the data relevant for us - I've done that in another context in the past.

utelle:
> `wxTranslations` is documented to return empty string if no match is found — then the "default language" fallback is realized simply by wxTranslations not loading any translation. IMNSHO this is the right thing to do, and `GetBestMatch` should do the same, i.e. return empty string to indicate that it failed to find a match. There's value in differentiating between failure and intentional/correct use of a fallback.

Ok. The current use of `GetBestMatch()` in `translations.cpp` works exactly in this way. So, I will not change the behaviour.

tests/intl/intltest.cpp
utelle: > Also should test that `sr-Latn` is correctly handled, including (a) on Windows where IIRC you said `sr` is considered `sr-Latn` (while it is `sr-Cyrl` on Linux) and that this correctly handles the situation where the available list has Linux-style `sr` (Cyrillic) and `sr@latin` translation files.

What I said about handling of `sr` on Windows, is related to locales. For matching translation languages we use the CLDR data, where `sr` maps to `sr-Cyrl`. Nevertheless, I add a test case.

tests/intl/intltest.cpp
utelle: I'll add a test case.

tests/intl/intltest.cpp
utelle: I'll add a test case.

tests/intl/intltest.cpp
utelle: ok, I'll adjust the test cases accordingly.

include/wx/uilocale.h
utelle: The following static methods could be moved from class `wxUILocale` in `include/wx/uilocale.h` to class `wxUILocaleImpl` in `include/wx/private/uilocale.h`:
```c++
   static wxString GetScriptAliasFromName(const wxString& scriptName);
   static wxString GetScriptNameFromAlias(const wxString& scriptAlias);
   static wxString GetLikelySubtags(const wxString & fromTag);
   static int GetMatchDistance(const wxString& desired, const wxString& supported);
   static bool SameRegionGroup(const wxString& language, const wxString& desiredRegion, const wxString& supportedRegion);
```
@vadz Please confirm, if I should make this change.

utelle:
I think I have addressed most items that were mentioned in the reviews.

The following items are still open:

1) Methods `AddLikelySubtags()`, `RemoveLikelySubtags()`, and `GetBestMatch()` public or not?
    IMHO these methods should stay public. For applications dealing with locales and/or languages they could make sense. If the methods stay public, I will add documentation for them.
2) Move static internal methods from `wxUILocale` to `wxUILocaleImpl` - that is, into a private header
    If @vadz confirms, I will move the methods.
3) Document the modified procedure for re-generating the language-related source files
4) Generate separate source files for wx language database, script mappings, likely subtags mapping, region groups mapping?
    If the separate files will be included from `languageinfo.cpp` anyway, I see no big advantage in generating separate files.
5) Investigate whether preparing the region group data via script is doable with reasonable effort
6) One check, namely [Ubuntu 22.04 wxGTK with wx containers](https://github.com/wxWidgets/wxWidgets/actions/runs/11159781864/job/31018877817?pr=24855#logs) fails. I don't experience the reported error locally (and most checks run successfully), therefore I'm unsure how to fix the error 
    ```
    ./src/common/translation.cpp: In member function ‘wxString wxTranslations::DoGetBestAvailableTranslation(const wxString&, const wxString&)’:
    ./src/common/translation.cpp:1456:50: error: invalid initialization of reference of type ‘wxVector<wxString>&’ {aka ‘const std::vector<wxString>&’} from expression of type ‘wxArrayString’
    ```

vadz:
> I think I have addressed most items that were mentioned in the reviews.

Thanks!

> The following items are still open:
> 
>     1. Methods `AddLikelySubtags()`, `RemoveLikelySubtags()`, and `GetBestMatch()` public or not?
>        IMHO these methods should stay public. For applications dealing with locales and/or languages they could make sense. If the methods stay public, I will add documentation for them.

I don't really know in which situations they would be useful for the application, so I wanted to keep them private, but if you think it's better to keep them public, I don't have any real objections to this neither. But they do need to be documented in this case and, ideally, the documentation should mention what would you use them for.

>     2. Move static internal methods from `wxUILocale` to `wxUILocaleImpl` - that is, into a private header
>        If @vadz confirms, I will move the methods.

Yes, please.

>     3. Document the modified procedure for re-generating the language-related source files

Yes, thanks in advance!

>     4. Generate separate source files for wx language database, script mappings, likely subtags mapping, region groups mapping?
>        If the separate files will be included from `languageinfo.cpp` anyway, I see no big advantage in generating separate files.

There is no _big_ advantage, but it would make navigating this source file cleaner. Right now it's not too bad as 90% of it is generated and there is relatively little code, but with these changes, and more code added to it, things start getting really messy with the code and data intermixed, so if they could be separated better, it would be nice.

>     5. Investigate whether preparing the region group data via script is doable with reasonable effort

I guess this data is going to change rarely, so we can keep it like this. I'd just like it to be mentioned somewhere that this file is hand-written and not generated, unlike other similar ones.

>     6. One check, namely [Ubuntu 22.04 wxGTK with wx containers](https://github.com/wxWidgets/wxWidgets/actions/runs/11159781864/job/31018877817?pr=24855#logs) fails.

This is because this build uses wx containers, i.e. `wxUSE_STD_CONTAINERS==0`, and so you need to use `wxArrayString::AsVector()` explicitly to convert `wxArrayString` to `std::vector<std::string>`.

utelle:
> > 1. Methods `AddLikelySubtags()`, `RemoveLikelySubtags()`, and `GetBestMatch()` public or not?
> >    IMHO these methods should stay public. For applications dealing with locales and/or languages they could make sense. If the methods stay public, I will add documentation for them.
> 
> I don't really know in which situations they would be useful for the application, so I wanted to keep them private, but if you think it's better to keep them public, I don't have any real objections to this neither. But they do need to be documented in this case and, ideally, the documentation should mention what would you use them for.

Well, I changed my mind, because in principle I agree with you that the methods are of limited use for applications. And I'm a bit reluctant to document these methods. That is, if you prefer they could be made private. But how?

The problem is that these methods can't be made private in `wxLocaleIdent` - at least not all -, because `wxTranslations` needs to be able to call `GetBestMatch()`. And the tests also call this method to verify that it works as expected.

For now, I made `AddLikelySubtags()` and `RemoveLikelySubtags()` private, but kept `GetBestMatch()` public.

@vadz If you have a more elegant solution, please let me know.

include/wx/private/lang_likely.h
```diff
@@ -0,0 +1,353 @@
+/////////////////////////////////////////////////////////////////////////////
+// Name:        wx/private/lang_likely.h
+// Purpose:     Mapping of likely subtags
+// Author:      Vadim Zeitlin
+// Created:     2024-10-04
+// Copyright:   (c) 1998 Vadim Zeitlin <zeitlin@dptmaths.ens-cachan.fr>
```
PBfordev, line 6: Sorry, I have nothing useful, just two nitpicks:
1. I think you should be listed as "Author" instead of Vadim (and according to the coding guidelines, your e-mail should be there in angle brackets).
2. "Copyright" should be provided as "(c) 2024 wxWidgets development team" (and Vadim's e-mail here is probably quite outdated, vadim@wxwidgets.org is being used elsewhere).

The same applies to other new files.

utelle: > Sorry, I have nothing useful, just two nitpicks:
> 
> 1. I think you should be listed as "Author" instead of Vadim (and according to the coding guidelines, your e-mail should be there in angle brackets).

Well, these files are actually generated by the script misc/languages/genlang.py. Therefore I set the script as the author.

> 2. "Copyright" should be provided as "(c) 2024 wxWidgets development team" (and Vadim's e-mail here is probably quite outdated, [vadim@wxwidgets.org](mailto:vadim@wxwidgets.org) is being used elsewhere).

I adjusted the copyright line of the new files accordingly.

PBfordev: > I adjusted the copyright line of the new files accordingly.

I apologize if I was unclear: I don't think Vadim's e-mail address should be included in the "Copyright: " line: It is not done so in any other files with team copyright. I mentioned the e-mail address in case you wanted to retain Vadim's copyright.

utelle:
Update on the open items.

> > 1. Methods `AddLikelySubtags()`, `RemoveLikelySubtags()`, and `GetBestMatch()` public or not?
> >    IMHO these methods should stay public. For applications dealing with locales and/or languages they could make sense. If the methods stay public, I will add documentation for them.
> 
> I don't really know in which situations they would be useful for the application, so I wanted to keep them private, but if you think it's better to keep them public, I don't have any real objections to this neither. But they do need to be documented in this case and, ideally, the documentation should mention what would you use them for.

As mentioned in a previous post I made methods `AddLikelySubtags()` and `RemoveLikelySubtags()` private. Method `GetBestMatch()` stays public, because it is called from `wxTranslations` and from the tests. However, I moved the declaration down in the header file just before the `private` section and added a comment that it is for internal use only.

> > 2. Move static internal methods from `wxUILocale` to `wxUILocaleImpl` - that is, into a private header
> >    If @vadz confirms, I will move the methods.
> 
> Yes, please.

Done.

> > 3. Document the modified procedure for re-generating the language-related source files

Still to be done, but I'm still working on the procedure.

> > 4. Generate separate source files for wx language database, script mappings, likely subtags mapping, region groups mapping?
> >    If the separate files will be included from `languageinfo.cpp` anyway, I see no big advantage in generating separate files.
> 
> There is no _big_ advantage, but it would make navigating this source file cleaner.

I modified the script `genlang.py` to generate separate header files (now located in `include/wx/private`) for each data table. The main source file `src/common/languageinfo.cpp` just includes these header files.

> > 5. Investigate whether preparing the region group data via script is doable with reasonable effort
> 
> I guess this data is going to change rarely, so we can keep it like this. I'd just like it to be mentioned somewhere that this file is hand-written and not generated, unlike other similar ones.

I found out that the Unicode CLDR data include a suitable file, from which the region data can be extracted. I'm working on a Lua script for this purpose.

tests/intl/intltest.cpp
vslavik: > What I said about handling of sr on Windows, is related to locales. For matching translation languages we use the CLDR data, where sr maps to sr-Cyrl. Nevertheless, I add a test case.

Thanks! I think the overall theme of these is that I would be more comfortable with having strong tests that go through typical scenarios for translations. For `sr`, it is relevant what the OS returns as user preference and that wx interprets it correctly. If it never returns plain `sr`, then there's no issue, of course. 

On a related note, it would be nice to have the tests go through POSIX handling too — in the available list, to make sure enumerations from `wxFilesTranslationsLoader` are parsed correctly.

utelle: > I think the overall theme of these is that I would be more comfortable with having strong tests that go through typical scenarios for translations. For `sr`, it is relevant what the OS returns as user preference and that wx interprets it correctly. If it never returns plain `sr`, then there's no issue, of course.

Unfortunately, I was not able to test this under Windows 11. Although I installed "Serbian (Serbia), Latin" and "Serbian (Serbia), Cyrillic" as additional preferred languages in the language and region settings, the Windows system function `GetUserPreferredUILanguages` (which is called by wxUILocale internally) returns only 2 entries, namely `de-DE` and `en-US` - that is, my primary UI language German plus US-English.

However, I'm pretty sure that Windows will return `sr-Latn-SR` or `sr-Cyrl-SR`, because you have to select a language, script and region for Serbian, when you install the additional languages.

Nevertheless, in the list of known locales Windows returns also the locale identifier `sr` - and that is associated with latin script.

> On a related note, it would be nice to have the tests go through POSIX handling too — in the available list, to make sure enumerations from `wxFilesTranslationsLoader` are parsed correctly.

Method `wxLocaleIdent::FromTag()` accepts all supported locale identifier formats. And I'm pretty confident that this method works correctly. Nevertheless, you are welcome to name some typical examples for the list of available translations, and I will add tests for them.

utelle:
While testing the new algorithm for determining the best translation language, I came across a Windows oddity.

To determine the list of preferred UI languages the method `wxUILocale::GetPreferredUILanguages()` is called, which in turn calls the Windows API function `GetUserPreferredUILanguages()`. 

One would expect that this function returns all languages a user added as _preferred languages_ in the section _Language and Region_ of the _Windows Settings_. However, this is not the case. 

In my development environment I added several preferred languages in addition to my primary UI language, _German_, but `GetUserPreferredUILanguages()` returns only 2 language tags, namely `de-DE` and `en-US`.

After some research I came to the conclusion that this is a known problem. To retrieve the list of preferred languages a user has installed, one could extract the list directly from the Windows registry or one could use the (undocumented) function `GetUserLanguages` in the DLL `bcp47langs.dll` (which is part of Windows since Windows 8).

@vadz Should I change the MSW implementation of `wxUILocale::GetPreferredUILanguages()` to retrieve the _real_ list of preferred UI languages?

vadz:
I had never heard about `GetUserLanguages()` before, but if this is the only way to get the real list of the configured languages, we should probably indeed try to use it — with the fallback to `GetUserPreferredUILanguages()` if we can't find this function or if it fails.

Thanks!

utelle:
> I had never heard about `GetUserLanguages()` before, but if this is the only way to get the real list of the configured languages, we should probably indeed try to use it — with the fallback to `GetUserPreferredUILanguages()` if we can't find this function or if it fails.

In commit 35f5a2752aa4d3b65b7e8c898f607a33e2d78276 I implemented the method that reads the Windows registry. This works as expected. The function `GetUserPreferredUILanguages()` is called afterwards and its result is appended to the list of preferred languages. It does no harm to add a language a second time, but we can change that behaviour, if you prefer.

I tested the method via the undocumented function `GetUserLanguages` in the DLL `bcp47langs.dll`, too. And it also works. However, I decided against using this method, because the function is undocumented after all (as all other function in that DLL) and it requires to dynamically load the DLL (which depends on many other system DLLs) and the result is returned in a `HSTRING` which needs to be freed to avoid memory leaks.

PS: I just saw that one Windows test failed in a non-GUI test, but IMHO it is not related to my code:
```
C:\projects\wxwidgets\tests\strings\strings.cpp(97): FAILED:
  trying to encode undefined Unicode character in FromWChar() at C:\projects\
  wxwidgets\src\common\strconv.cpp:1164
```

utelle:
In the meantime I worked on the underlying regeneration process for the wxWidgets language files. I improved the Lua scripts which are used to extract the required data from the Unicode, Windows and wxWidgets data files. Additionally, I replaced the Windows batch files by PowerShell scripts. I still have to look at the SQL code to check whether diagnostic output can be improved.

In a separate private repository I made some experiments to set up a GitHub action to automate the regeneration process. In principle, the process works. However, the step to extract the list of known locales from Windows should **not** be run on the GitHub runner, because the latest GitHub Windows runner is _Windows Server 2022_. The locale data of _Windows Server 2022_ are from 2020 or so, and therefore rather outdated. The data file currently in the repository were extracted from _ Windows 11 Pro 23H2_, that is, second half of 2023.

If a GitHub action should be set up, it should be an action that is triggered manually. Updates should be made when new Unicode data are released - for example, the official release of _CLDR 46_ will happen in the near future - and after major Windows updates - like the upcoming Windows 11 24H2 update.

@vadz Please let me know if you see further issues which should be addressed, before you consider to merge this PR.

vadz:
@utelle I'll try to look at this and merge it a.s.a.p. but I need more time than I have right now to even _try_ to understand the changes... Maybe I'll manage to find it next week-end.

Sorry for the delay!

utelle:
@vadz 

> I'll try to look at this and merge it a.s.a.p.

Take your time there is no hurry from my side.

> but I need more time than I have right now to even _try_ to understand the changes...

Actually, the changes are not that big after all. There were _additions_ for the new algorithm, but the process to generate the language related source files stayed more or less the same. Just to summarize what was done:

- replace algorithm to find a best match fro translation language with an algorithm similar to the Unicode algorithm mentioned by @vslavik 
- add test cases to check the new algorithm
- add new Unicode data files needed by the new algorithm
- add Lua scripts to generate SQL scripts from the Unicode data files
- streamline process to (re)generate the tabular language data files used by the Python script `genlang.py`
- adjust Python script `genlang.py` to generate separate header files for the various data tables (instead of the monolithic languageinfo.cpp file)
- replace Windows batch command files by PowerShell scripts

I will add the GitHub workflow script for (re)generating the language data source files within the next couple of days. However, IMHO actually replacing the language related source files in wxWidgets should stay a _manual_ task, so that the results of the regeneration process can be checked carefully. The workflow will upload artifacts that can be checked, before simply copying the generated files to the wxWidgets repository. I hope this is a feasible approach.

utelle:
After adding the GitHub workflow `genlangdb.yml` this PR is now complete - unless further issues are reported, that is.

src/common/languageinfo.cpp
```diff
@@ -56,6 +58,47 @@ static std::unordered_map<wxString, wxString> gs_scmap_alias2name;
...
+
+/* static */ bool wxUILocaleImpl::SameRegionGroup(const wxString& language, const wxString& desiredRegion, const wxString& supportedRegion)
+{
+    const auto languageIter = gs_region_groups_map.find(language);
+    if (languageIter != gs_region_groups_map.end())
+    {
+        const auto groupSet = languageIter->second;
+        bool containsDesired = groupSet.find(desiredRegion) != groupSet.end();
+        bool containsSupported = groupSet.find(supportedRegion) != groupSet.end();
+        return (containsDesired && containsSupported) || (!containsDesired && !containsSupported);
```
vadz, line 96: This condition is difficult to read/understand IMO. Maybe writing it like this:

```suggestion
        if (containsDesired && containsSupported)
            return true;
            
        if (!containsDesired && !containsSupported)
            return true;
```

could help a bit, especially if comments explaining what do the 2 cases correspond to are added.

include/wx/uilocale.h
```diff
@@ -28,6 +28,13 @@ enum
     wxCompare_CaseInsensitive = 1
 };
 
+// Flags for wxLocaleIdent::RemoveLikelySubtags().
```
vadz, line 31: Is this supposed to be public? If so, it should be documented.

include/wx/uilocale.h
```diff
@@ -28,6 +28,13 @@ enum
...
+enum
+{
+    wxSubtags_FavorRegion = 0,
+    wxSubtags_FavorScript = 1
```
vadz, line 35: We tend to use British spelling, so I'd call them

```suggestion
    wxSubtags_FavourRegion = 0,
    wxSubtags_FavourScript = 1
```

but it's not that important, especially if they can remain private.

vadz (commented):
Sorry for yet another delay, I've finally tried to look at this again, but I admit I still only did it very superficially and, in particular, didn't make any effort to really understand all these bat/powershell/lua/SQL/Python scripts — I just hope we'll be able to maintain them in the future.

Anyhow, except for some minor remarks here and there, I think this should be merged because the tests show that this improves the behaviour and while I'm quite unhappy about all the added complexity, I guess it's unavoidable.

@vslavik Please let me know in the observable future (say, in a month?) if you have any objections, otherwise I'll go ahead and merge it.

@utelle Thanks again for all your work on this!

src/common/languageinfo.cpp
utelle: I add a comment in front of the original code. IMHO splitting the condition doesn't make it clearer and better to understand. Desired and supported region belong to the same region group, if they are either both **contained** or both **not contained** in the checked region group.

src/common/uilocale.cpp
```diff
@@ -275,6 +275,330 @@ wxLocaleIdent wxLocaleIdent::FromTag(const wxString& tag)
...
+        return wxLocaleIdent();
+    }
+
+    // If locale is the default C locale, return it
+    if (IsDefaultCLocale(language))
+    {
+        return localeIdent;
+    }
+
+    // Remove the script code 'Zzzz' and the region code 'ZZ' if they occur.
```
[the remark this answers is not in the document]
utelle: I added a comment explaining why this is done.

src/common/uilocale.cpp
```diff
@@ -275,6 +275,330 @@ wxLocaleIdent wxLocaleIdent::FromTag(const wxString& tag)
...
+    {
+        toTag = wxUILocaleImpl::GetLikelySubtags(language);
+    }
+
+    if (toTag.empty())
+    {
+        // No match, return invalid locale identifier
+        return wxLocaleIdent();
+    }
+    else
```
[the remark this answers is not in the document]
utelle: Done.

src/common/uilocale.cpp
```diff
@@ -275,6 +275,330 @@ wxLocaleIdent wxLocaleIdent::FromTag(const wxString& tag)
...
+    {
+        // Match found. Update script and/or region subtags of the result.
+        const wxArrayString& subtags = wxSplit(toTag, '-', '\0');
+
+        // Attribute m_tag needs to be cleared, because it contains the tag value of the original identifier
+        locId.m_tag.clear();
+
+        // Fill in script and region subtags, if they were originally empty
```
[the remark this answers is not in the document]
utelle: The likely subtags data retrieved from Unicode CLDR **always** map to fully qualified tags with language, script, **and** region subtags. The code here is only executed, if a match was found (and all subtags are therefore defined).

src/common/uilocale.cpp
```diff
@@ -275,6 +275,330 @@ wxLocaleIdent wxLocaleIdent::FromTag(const wxString& tag)
...
+        if (script.empty())
+        {
+            locId.Script(subtags[1]);
```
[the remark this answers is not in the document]
utelle: Done.

src/common/uilocale.cpp
```diff
@@ -275,6 +275,330 @@ wxLocaleIdent wxLocaleIdent::FromTag(const wxString& tag)
...
+    // Test "language"
+    locId = AddLikelySubtags(trial);
+    if (script.IsSameAs(locId.GetScript()) && region.IsSameAs(locId.GetRegion()))
+    {
+        return trial;
+    }
+
+    // Test "language-region" if favoring region or
+    //      "language-script" if favoring script
+    if (subtagsFavor == wxSubtags_FavorRegion)
```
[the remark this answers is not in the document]
utelle: Changed accordingly.

src/common/uilocale.cpp
```diff
@@ -275,6 +275,330 @@ wxLocaleIdent wxLocaleIdent::FromTag(const wxString& tag)
...
+}
+
+// "Infinity" value needs to be larger than any real language distance
+static const int DISTANCE_INFINITY = 1000;
+
+/* static */
+wxString wxLocaleIdent::GetBestMatch(const wxVector<wxString>& desired, const wxVector<wxString>& supported)
+{
+    // Default distance values
+    static int defaultLanguageDistance = wxUILocaleImpl::GetMatchDistance("*", "*");
```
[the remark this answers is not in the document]
utelle: Yes. Changed accordingly.

src/common/uilocale.cpp
```diff
@@ -275,6 +275,330 @@ wxLocaleIdent wxLocaleIdent::FromTag(const wxString& tag)
...
+    wxLocaleIdent desiredLocaleIn;
+    wxLocaleIdent desiredLocaleMax;
+    wxLocaleIdent supportedLocaleIn;
+    std::vector<wxLocaleIdent> supportedLocalesIn;
+    std::vector<wxLocaleIdent> supportedLocalesMax;
+    std::vector<int> supportedLocalesMinDistance;
+
+    // Indexes of best match
+    int bestDesired = -1;
+    wxUnusedVar(bestDesired);
```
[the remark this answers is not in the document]
utelle: Well, for one, without `bestDesired` it is difficult to debug the code. Secondly, it may be of interest to which of the given desired languages the selected supported language actually matches, although this information is currently not passed onward to the calling method.

utelle (commented):
I think I have addressed all comments.

utelle:
> I've finally tried to look at this again, but I admit I still only did it very superficially and, in particular, didn't make any effort to really understand all these bat/powershell/lua/SQL/Python scripts — I just hope we'll be able to maintain them in the future.

As long as the format of the Unicode data files does not change, the scripts should continue to work.

The most complex part is the creation of unique `wxLANGUAGE_*` identifiers. IMHO wxWidgets should get rid of them in the long run. Except maybe for some special values like `wxLANGUAGE_DEFAULT` or `wxLANGUAGE_UNKNOWN` or `wxLANGUAGE_ENGLISH`. (I have developped wxWidgets-based code for about 20 years now, and the only identifiers I ever used were those mentioned above.)

> Anyhow, except for some minor remarks here and there, I think this should be merged because the tests show that this improves the behaviour and while I'm quite unhappy about all the added complexity, I guess it's unavoidable.

IMHO there is not much added complexity, except for some additional data tables which are required for the new algorithm. However, these data tables are simply extracted from the Unicode data files and converted to C++ syntax. The algorithm itself is a bit more involved than the prior one.

.github/workflows/genlangdb.yml
```diff
@@ -0,0 +1,68 @@
+name: Generate wxWidgets language database
+
+on: workflow_dispatch
```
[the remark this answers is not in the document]
utelle: Well, it does no harm to run this workflow regularly. But I don't see much advantage in doing so. Underlying Unicode or Windows data files maybe change once per year, maybe even less often. If the GitHub action breaks (for example, due to an outdated action), it will have to be fixed, of course. However, it will not be an urgent task - whether some more or less exotic languages are supported a bit earlier or later shouldn't matter too much.

vslavik:
> The most complex part is the creation of unique wxLANGUAGE_* identifiers. IMHO wxWidgets should get rid of them in the long run.

+1

> @vslavik Please let me know in the observable future (say, in a month?) if you have any objections, otherwise I'll go ahead and merge it.

I'm good with it. Thanks @utelle for all the work on this PR!

vadz:
Thanks again, I'm (squash-)merging this now.

PR #24864, opened 2024-10-07
Deprecate `wxScreenDC` even more and remove macOS implementation

vadz:
@csomor This is a bit brutal, but I don't know how to do compile/run-time tests there and it just doesn't seem worth it, so if you don't terribly mind, I'd like to merge this to at least fix compilation, as this is certainly going to affect more and more people.

### DIFF
--- a/docs/doxygen/mainpages/cat_classes.h
+++ b/docs/doxygen/mainpages/cat_classes.h
@@ -292,11 +292,13 @@ Related Overviews: @ref overview_dc
 @li wxBufferedPaintDC: A helper device context for double buffered drawing
     inside @b OnPaint().
 @li wxClientDC: A device context to access the client area outside
-    @b OnPaint() events
+    @b OnPaint() events (doesn't work on many modern systems, deprecated).
 @li wxPaintDC: A device context to access the client area inside @b OnPaint()
     events
-@li wxWindowDC: A device context to access the non-client area
-@li wxScreenDC: A device context to access the entire screen
+@li wxWindowDC: A device context to access the non-client area (Windows-only,
+    deprecated).
+@li wxScreenDC: A device context to access the entire screen (doesn't work on
+    many modern systems, deprecated).
 @li wxDC: The device context base class
 @li wxMemoryDC: A device context for drawing into bitmaps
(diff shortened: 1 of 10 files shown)


vadz:
Stefan, I'll merge this soon as I think more and more people risk running into this.

csomor:
> Stefan, I'll merge this soon as I think more and more people risk running into this.

sorry, I missed this, the reason for this change are the compile erorrs ?

csomor:
ok, as usual there's a screenshot kit now ... is the intention to remove wxScreenDC throughout wx ? otherwise I'd like to make it compile again, and then add a screenshot implementation later for macOS 15

vadz:
> is the intention to remove wxScreenDC throughout wx ?

Of course it would be better to keep it if we can. It still works under MSW, AFAIK, and there might be a way to use some (optional) Wayland protocol to make it work there, I think, but it probably wouldn't be trivial.

> otherwise I'd like to make it compile again, and then add a screenshot implementation later for macOS 15

The trouble is that I have no idea how to make it compile again. But if you could do this, it would be great, of course, especially for 3.2, where removing it would be even more problematic. TIA!

csomor:
ok, thanks, I'll do a PR right now

SpareSimian:
I grepped my code for instances of wxScreenDC and see one usage: Measuring the font size in a wxGrid to set the label sizes. It's based on the Python code here: https://wiki.wxpython.org/wxGrid#Autosizing_labels

vadz:
> I grepped my code for instances of wxScreenDC and see one usage: Measuring the font size in a wxGrid to set the label sizes.

FWIW nowadays you'd use `wxInfoDC` for this.

vadz:
I've merged this branch without the last commit, which is not needed any more after Stefan's changes.

When/if we implement `wxScreenDC` for macOS 15+ (and Wayland?), we could revert some of the commits here, but for now I think they describe the situation accurately.

SpareSimian:
wxInfoDC doesn't appear in my copy of 3.2.5. I surmise that this is a fairly recent addition?

vadz:
Yes, it's 3.3-only.

PR #24869, opened 2024-10-09
autoconf fixes for gcc 14 on Solaris

alanc:
This fixes two autoconf checks that currently produce the wrong results when run with gcc 14 on Solaris 11.4 as they trigger errors in gcc 14 that were only warnings in previous gcc versions.

The first issue causes the configure script to incorrectly decide that x86 systems are big endian instead of little.  The second one causes it to determine that the OSS ioctl SNDCTL_DSP_SPEED is not supported on Solaris 11 when it is.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -269,7 +269,7 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([#include <sys/types.h>
  not big endian
 #endif])], ac_cv_c_bigendian=yes, ac_cv_c_bigendian=no)])
 if test $ac_cv_c_bigendian = unknown; then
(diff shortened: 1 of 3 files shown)


oneeyeman1:
@alanc ,
Is it building successfully with the Solaris native compiler?

alanc:
> @alanc , Is it building successfully with the Solaris native compiler?

The configure script completes successful with the Solaris Studio compilers as they do not issue errors for the things gcc recently promoted from warnings to errors.  I've not tried a full build with them, as we use the gcc compilers for the wxWidgets packages we provide in the Solaris package repository.

PR #24870, opened 2024-10-09
Fix minor typo, small rewording

Blake-Madden:


vadz:

PR #24874, opened 2024-10-11
Fix Missing helptext in status bar when mouse is above submenu-entry

### LINKED ISSUE #24842: Missing helptext in status bar when mouse is above submenu-entry
# Description

As the title says, you don't get the help-text in the status bar when the mouse is over a submenu-entry.
All the other menu-entries do show the helptext, so I am not sure if its my error or a problem in the library.

# To Reproduce:

- edit the samples\statbar\statbar.cpp
- add in line 359 a helptext: 'statbarStyleMenu);' --> 'statbarStyleMenu, "help Status bar style");'
- add in line 403 a helptext: 'statbarPaneStyleMenu);' --> 'statbarPaneStyleMenu, "help Field style");'

Btw already in 2009 there was a question about this, but no answers:
"wxMenu::AppendSubMenu help string not displaying"
[viewtopic.php?p=111705#p111705](https://forums.wxwidgets.org/viewtopic.php?p=111705#p111705)

# Platform and version information:
  - wxMSW 
  - wxWidgets v3.2.6 (same in v3.2.4, v3.2.5)
  - Win10

mxdamien:
closes #24842

### DIFF
--- a/include/wx/event.h
+++ b/include/wx/event.h
@@ -45,6 +45,7 @@ class WXDLLIMPEXP_FWD_BASE wxEventFilter;
 #if wxUSE_GUI
     class WXDLLIMPEXP_FWD_CORE wxDC;
     class WXDLLIMPEXP_FWD_CORE wxMenu;
+    class WXDLLIMPEXP_FWD_CORE wxMenuItem;
     class WXDLLIMPEXP_FWD_CORE wxReadOnlyDC;
     class WXDLLIMPEXP_FWD_CORE wxWindow;
     class WXDLLIMPEXP_FWD_CORE wxWindowBase;
@@ -2576,12 +2577,12 @@ private:
 class WXDLLIMPEXP_CORE wxMenuEvent : public wxEvent
 {
 public:
-    wxMenuEvent(wxEventType type = wxEVT_NULL, int winid = 0, wxMenu* menu = nullptr)
+    wxMenuEvent(wxEventType type = wxEVT_NULL, int winid = 0, wxMenu* menu = nullptr, wxMenuItem* menuItem = nullptr)
         : wxEvent(winid, type)
-        { m_menuId = winid; m_menu = menu; }
+        { m_menuId = winid; m_menu = menu; m_menuItem = menuItem; }
     wxMenuEvent(const wxMenuEvent& event)
         : wxEvent(event)
-    { m_menuId = event.m_menuId; m_menu = event.m_menu; }
+    { m_menuId = event.m_menuId; m_menu = event.m_menu; m_menuItem = event.m_menuItem; }
 
     // only for wxEVT_MENU_HIGHLIGHT
     int GetMenuId() const { return m_menuId; }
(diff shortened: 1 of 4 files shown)


src/msw/window.cpp
```diff
@@ -2577,6 +2579,38 @@ wxMenu* wxWindowMSW::MSWFindMenuFromHMENU(WXHMENU hMenu)
     return nullptr;
 }
 
+wxMenuItem* wxWindowMSW::MSWFindMenuItemFromHMENU(WXHMENU hMenu, int item)
+{
+    MENUITEMINFO mii;
+    mii.cbSize = sizeof(MENUITEMINFO);
```
vadz, line 2585: Minor, but this is a bit shorter and more consistent with the rest of our code:
```suggestion
    WinStruct<MENUITEMINFO> mii;
```

include/wx/frame.h
```diff
@@ -233,6 +233,9 @@ class WXDLLIMPEXP_CORE wxFrameBase : public wxTopLevelWindow
     // return false if there is no help for such item
     bool ShowMenuHelp(int helpid);
 
+    // show the help string for a menu item directly without finding it by id
+    bool ShowMenuHelp(const wxString& helpString);
```
vadz, line 237: I'm not really sure we need this function, it seems to be exactly the same as `DoGiveHelp()`, isn't it?

src/common/framecmn.cpp
```diff
@@ -343,7 +343,15 @@ void wxFrameBase::OnMenuHighlight(wxMenuEvent& event)
 {
     event.Skip();
 
-    (void)ShowMenuHelp(event.GetMenuId());
+    wxMenuItem* menuItem = event.GetMenuItem();
+    if( menuItem )
```
vadz, line 347: Minor, but I'd write this as
```suggestion
    if ( wxMenuItem* menuItem = event.GetMenuItem() )
```

vadz (commented):
Thanks David, this looks good to me and I'd just like to simplify it a tiny little bit before merging it.

include/wx/frame.h
mxdamien: You're right. Of course DoGiveHelp can be used directly here.

mxdamien:
Great! Thank you very much for the review Vadim. I agree on all your remarks

include/wx/event.h
```diff
@@ -48,6 +48,7 @@ class WXDLLIMPEXP_FWD_BASE wxEventFilter;
     class WXDLLIMPEXP_FWD_CORE wxReadOnlyDC;
     class WXDLLIMPEXP_FWD_CORE wxWindow;
     class WXDLLIMPEXP_FWD_CORE wxWindowBase;
+    class WXDLLIMPEXP_FWD_CORE wxMenuItem;
```
PBfordev, line 51: nitpick: The classes are declared in the alphabetical order, so `wxMenuItem` should be moved after `wxMenu`.

mxdamien: Thank you !

PR #24877, opened 2024-10-13
wxQt: Reduce flicker when hovering over pane buttons

AliKet:
The problem can be seen in the **auidemo** sample under **Windows** or **Linux**


vadz:
Thanks for the fix, will merge soon.

PR #24881, opened 2024-10-15
Support multiple HTTP headers with same name

dinkelacker:
This is a draft. It is currently missing the implementation for macOS resp. URLSession.

Closes #24878.

### DIFF
--- a/src/common/webrequest.cpp
+++ b/src/common/webrequest.cpp
@@ -321,7 +321,7 @@ SplitParameters(const wxString& s, wxWebRequestHeaderMap& parameters)
         }
         pvalue.Trim();
         if ( !pname.empty() )
-            parameters[pname] = pvalue;
+            parameters[pname] = { pvalue };
         if ( it != end )
             ++it;
     }
@@ -418,6 +418,13 @@ void wxWebRequestBase::SetHeader(const wxString& name, const wxString& value)
     wxCHECK_IMPL_VOID();
 
     m_impl->SetHeader(name, value);
+}
+
+void wxWebRequestBase::AddHeader(const wxString& name, const wxString& value)
+{
+    wxCHECK_IMPL_VOID();
+
+    m_impl->AddHeader(name, value);
 }
 
 void wxWebRequestBase::SetMethod(const wxString& method)
@@ -702,10 +709,10 @@ wxString wxWebResponseImpl::GetSuggestedFileName() const
     wxString contentDisp = GetHeader("Content-Disposition");
     wxWebRequestHeaderMap params;
     const wxString disp = wxPrivate::SplitParameters(contentDisp, params);
-    if ( disp == "attachment" )
+    if ( disp == "attachment" && !params["filename"].empty())
     {
         // Parse as filename to filter potential path names
-        wxFileName fn(params["filename"]);
+        wxFileName fn(params["filename"].back());
         suggestedFilename = fn.GetFullName();
     }
 
--- a/src/common/webrequest.cpp
+++ b/src/common/webrequest.cpp
@@ -843,6 +843,13 @@ wxString wxWebResponse::GetHeader(const wxString& name) const
     return m_impl->GetHeader(name);
 }
 
+std::vector<wxString> wxWebResponse::GetAllHeaderValues(const wxString& name) const
+{
+    wxCHECK_IMPL( std::vector<wxString>() );
+
+    return m_impl->GetAllHeaderValues(name);
+}
+
 wxString wxWebResponse::GetMimeType() const
 {
     wxCHECK_IMPL( wxString() );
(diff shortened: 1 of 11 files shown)


src/msw/webrequest_winhttp.cpp
```diff
@@ -214,6 +214,38 @@ static wxString wxWinHTTPQueryHeaderString(HINTERNET hRequest, DWORD dwInfoLevel
     return result;
 }
 
+static std::vector<wxString> wxWinHTTPQueryAllHeaderStrings(HINTERNET hRequest, DWORD dwInfoLevel,
```
vadz, line 217: Could this be refactored to reuse the existing `wxWinHTTPQueryHeaderString()` by adding an optional `DWORD* nextIndex = WINHTTP_NO_HEADER_INDEX` parameter to it and passing it `&nextIndex` when calling it from here?

The code dealing with buffers is sufficiently non trivial for me not to want to duplicate it...

src/common/webrequest.cpp
```diff
@@ -702,10 +709,10 @@ wxString wxWebResponseImpl::GetSuggestedFileName() const
     wxString contentDisp = GetHeader("Content-Disposition");
     wxWebRequestHeaderMap params;
     const wxString disp = wxPrivate::SplitParameters(contentDisp, params);
-    if ( disp == "attachment" )
+    if ( disp == "attachment" && !params["filename"].empty())
```
vadz, line 712: Exceedingly nitpicky, but:
```suggestion
    if ( disp == "attachment" && !params["filename"].empty() )
```

src/osx/webrequest_urlsession.mm
```diff
@@ -434,6 +434,12 @@ void Wait()
         return wxString();
 }
 
+std::vector<wxString> wxWebResponseURLSession::GetAllHeaderValues(const wxString& WXUNUSED(name)) const
+{
+    std::vector<wxString> result;
+    return result;
```
vadz, line 440: It's fine not to implement this for macOS if you don't know how to do it but I'd at least call `GetHeader()` here as returning the first header value seems like a more useful fallback than not returning anything at all.

This behaviour should be mentioned in the documentation, of course.

vadz (commented):
Thanks for this!

The API looks good to me and a separate thank you for updating the documentation too!

Once the CI builds are fixed, this can be merged.

src/osx/webrequest_urlsession.mm
```diff
@@ -190,8 +190,10 @@ - (void)URLSession:(NSURLSession *)session
     // Set request headers
     for (wxWebRequestHeaderMap::const_iterator it = m_headers.begin(); it != m_headers.end(); ++it)
     {
-        [req setValue:wxCFStringRef(it->second).AsNSString() forHTTPHeaderField:
-         wxCFStringRef(it->first).AsNSString()];
+        // TODO: URLSession does not support multiple headers with the same name.
```
vadz, line 193: FWIW there apparently is `[NSURLSession addValue:forHTTPHeaderField:]`, so you could try using it here if you're motivated.

src/osx/webrequest_urlsession.mm
```diff
@@ -434,6 +436,13 @@ void Wait()
         return wxString();
 }
 
+std::vector<wxString> wxWebResponseURLSession::GetAllHeaderValues(const wxString& name) const
+{
+    // TODO: URLSession does not support multiple headers with the same name.
```
vadz, line 441: Considering `addValue` description, it would be logical — but I didn't test it at all — if `NSURLSession` also combined multiple header values into a single comma-separated string.

vadz (approved):
Thanks, we can merge this already or I can wait if you'd like to test `addValue` — please let me know if you would.

Thanks again!

vadz:
Sorry, I've just found a big problem with this PR: the previously working code which called

```cpp
    m_session.AddCommonHeader("User-Agent", "Whatever");
```

doesn't work any more because we end up with 2 `User-Agent:` lines in the headers passed to `WinHttpSendRequest()` due to the default "User-Agent" already present there and it just fails immediately when this happens.

This is a silent regression and it really shouldn't work like this, so I'm going to do something about it, but I'm not sure what exactly yet. Making `AddCommonHeader()` behave as `SetHeader()`, rather than as `AddHeader()`, would be very inconsistent. I guess we're going to need to avoid calling it by default in the first place.

vadz:
Moreover, I start to really question if we should be ever sending multiple lines for the same header in the first place. It looks like the only valid uses for multiple headers, like `Accept:`, can be also achieved by combining all their values into a single string using commas and this is something that [WinHttpAddRequestHeaders()](https://learn.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpaddrequestheaders) can do and while it can also do other things, it never adds multiple header lines.

vadz:
Sorry, after re-checking HTTP standard, I think it was a bad idea to merge this PR. It [clearly says](https://datatracker.ietf.org/doc/html/rfc9110#name-field-order)

> This means that, aside from the well-known exception noted below, a sender MUST NOT generate multiple field lines with the same name in a message

where the "well-known exception" is `Set-Cookie`.

I'm going to revert this and if there is any interest, we should add a special function specifically for cookies, but it doesn't make sense to allow sending multiple headers when the RFC explicitly says that we MUST NOT be doing this.

PR #24873, opened 2024-10-10
Add the wxWARN_UNUSED attribute to more classes

lanurmi:
Here is the new `wxWARN_UNUSED` attribute added to some other commonly used classes, besides `wxString`. Not all of these classes have any unused instances in wx's own code base.

The classes affected by this PR are:
```
wxColour
wxStringTokenizer
wxSize, wxRect, wxPoint, wxRealPoint
wxLongLong, wxULongLong
wxArrayString
wxDateTime
```

The changes are in multiple commits to make cherry-picking or dropping easier. Some squashing may be appropriate.

PR #24882, opened 2024-10-16
Backport wxWARN_UNUSED into 3.2, along with removing unused variables

lanurmi:
Backport #24833 and #24873 into 3.2 branch.

This is introduced as an opt-in feature in 3.2, to avoid potentially breaking compilation of projects that use `-Werror=unused*`.

`wxWARN_UNUSED` is defined as nothing, unless `wxNO_UNUSED_VARIABLES` is defined by the user.

`wxWARN_UNUSED` is also automatically opted-in when `WXBUILDING` is defined. If this is seen as a risk, I'm not against not doing so.

`wxNO_UNUSED_VARIABLES` as a name is IMHO rather awkward for enabling something, but couldn't come up with something better that would begin with `wxNO_`.

### DIFF
--- a/docs/doxygen/mainpages/const_cpp.h
+++ b/docs/doxygen/mainpages/const_cpp.h
@@ -406,11 +406,11 @@ more details.
         this symbol is only honoured by wxWidgets 3.2 (since 3.2.5) but not by
         the future versions of the library.}
 @itemdef{wxNO_UNUSED_VARIABLES,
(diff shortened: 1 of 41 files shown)


vadz:
Thanks, looks good to me, but could you please add the new symbol to `docs/doxygen/mainpages/const_cpp.h`? TIA!

vadz:
Thanks, I've rebased to remove fixups and merged this into #24848, so closing this one now.

PR #24883, opened 2024-10-16
wxQt: Fix wxEVT_{ENTER,LEAVE}_WINDOW events generation

AliKet:
Quoting Qt docs:

- enterEvent() is called when the mouse enters the widget's screen space. (This excludes screen space owned by any of the widget's children.)
- leaveEvent() is called when the mouse leaves the widget's screen space. If the mouse enters a child widget it will not cause a leaveEvent().

Which contradicts what wx docs says:

For the wxEVT_ENTER_WINDOW and wxEVT_LEAVE_WINDOW events purposes, the mouse is considered to be inside the window if it is in the window client area and not inside one of its children. In other words, the parent window receives wxEVT_LEAVE_WINDOW event not only when the mouse leaves the window entirely but also when it enters one of its children.

### DIFF
--- a/src/qt/window.cpp
+++ b/src/qt/window.cpp
@@ -1727,7 +1727,7 @@ bool wxWindowQt::QtHandleMouseEvent ( QWidget *handler, QMouseEvent *event )
 
     // Use screen position as the event might originate from a different
     // Qt window than this one.
-    wxPoint mousePos = ScreenToClient(wxQtConvertPoint(event->globalPos()));
+    const wxPoint mousePos = ScreenToClient(wxQtConvertPoint(event->globalPos()));
 
     wxMouseEvent e( wxType );
     e.SetEventObject(this);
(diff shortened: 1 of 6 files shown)


vadz:
Thanks, but, just to be sure, and because we don't seem to have any tests for this in the test suite, did you already check that the other ports actually do behave like this, i.e. conform to the documentation?

vadz:
> I don't know what's going on with the failing tests? They pass on my machine!

Sorry, I don't really know but I wonder if `wxYield()` might not be enough and we may need to use `WaitFor()` instead in more places?

AliKet:
> > I don't know what's going on with the failing tests? They pass on my machine!
> 
> Sorry, I don't really know but I wonder if `wxYield()` might not be enough and we may need to use `WaitFor()` instead in more places?

No matter what I do, `wxGTK2` refuses to pass when run in **GitHub Actions** even though it passes locally.

I think this PR is ready for now, but it would be kind of you to let me know how it behaves under `wxOSX`, TIA!

vadz:
> I think this PR is ready for now, but it would be kind of you to let me know how it behaves under `wxOSX`, TIA!

Sorry, I forgot to ask: what exactly would you like me to test there?

vadz:
After rereading this, I've realized that you probably were asking me to test the diff to the minimal sample under macOS ­and so I just did it. The conclusion is that these events are broken under macOS, see #25055, but I don't think this should prevent us from merging this PR, so I'll do it soon.

Sorry for the delay...

PR #24887, opened 2024-10-17
Update list of locales to install

### LINKED ISSUE #24880: Not all locale files are installed during Makefile build
When running `make install`, not all currently produced locale files actually get installed.

The list that gets installed seems to be this hard-coded list of ~22 which git blame says has not been updated in over 20 years:
https://github.com/wxWidgets/wxWidgets/blob/master/build/bakefiles/wx.bkl#L142

Whereas the locales directory seems to contain ~46.

Is there a way in bakefile-speak to specify a dynamically generated list of files?  If not, I could at the very least propose an update to the hard-coded list.

swt2c:
Fixes: https://github.com/wxWidgets/wxWidgets/issues/24880


vadz:
Thanks, will merge soon.

PR #24889, opened 2024-10-18
Fixes #24888

### LINKED ISSUE #24888: New method wxAuiManager::LoadLayout is incorrectly documented
### Description

The documentation for the new wxAuiManager::LoadLayout in `interface/wx/aui/framemanager.h` declares this method as `const` which does not match the implemenation which is non-const (correctly I think).

- wxWidgets version  you use: 3.3.0 (master)
- wxWidgets port  you use: NA
- OS  and its version: NA
   NA
  + GTK version:  NA
  + Which GDK  backend is used: NA
  + Desktop environment : NA
  + Current theme:  NA

mcorino:


vadz:

PR #24890, opened 2024-10-18
override DoGetBestSize and DoGetBestClientSize of wxTabFrame

sz5000:
override DoGetBestSize and DoGetBestClientSize of wxTabFrame otherwise it is very small when we we move one tab to the left or right position.

There are some other issues with the drag/drop visualization, highlighted area is sometimes not correct.

### DIFF
--- a/src/aui/auibook.cpp
+++ b/src/aui/auibook.cpp
@@ -1565,6 +1565,16 @@ protected:
         *y = m_rect.height;
     }
 
+    wxSize DoGetBestSize() const override
+    {
+        return wxSize(m_rect.width, m_rect.height);
+    }
+
+    wxSize DoGetBestClientSize() const override
+    {
+        return wxSize(m_rect.width, m_rect.height);
+    }
+
 public:
     bool Show( bool WXUNUSED(show = true) ) override { return false; }
 
@@ -2107,7 +2117,7 @@ bool wxAuiNotebook::DeletePage(size_t page_idx)
 bool wxAuiNotebook::RemovePage(size_t page_idx)
 {
     // Lock the window for changes to avoid flicker when
-    // removing the active page (there is a noticeable 
+    // removing the active page (there is a noticeable
     // flicker from the active tab is closed and until a
     // new one is selected) - this is noticeable on MSW
     wxWindowUpdateLocker locker { this };


sz5000:
We have wxAUINotebook, two taps open, drag one of them and move it to the right window edge.
image

With the fix, it looks like this, split in half
image

Without the fix (master branch), it looks like this:
image

Problem is located in framemanager.cpp:
```
 if (pinfo.best_size == wxDefaultSize &&
     pinfo.window)
 {
     pinfo.best_size = pinfo.window->GetBestSize();

     if (pinfo.min_size != wxDefaultSize)
     {
         if (pinfo.best_size.x < pinfo.min_size.x)
             pinfo.best_size.x = pinfo.min_size.x;
         if (pinfo.best_size.y < pinfo.min_size.y)
             pinfo.best_size.y = pinfo.min_size.y;
     }
 }
```

GetBestSize returns wxSize(1,1)

sz5000:
Other issue (not solved by this patch)  is e.g. the following
image

Drop indicator of aui manager (blue rect) is at the wrong position when we try to drag und drop the right tab back to the left one.

sz5000:
The sample in samples/aui/ can be used to reproduce the problems.

1. Drag the tab from the notebook to the right, the tab is then very narrow on the right edge.

2. Drag the tab dragged to the right back into the tab bar, the insertion position is sometimes visualized incorrectly.

image

vadz:
Thanks, I indeed see both problems, although the second one took some effort to reproduce. I believe it happens when the drop point is exactly between the notebook pages.

Let me check if we can find a better fix for the first one, if not we'll apply this PR.

As for the second one, I'll open another issue for it as it seems to be something rather different.

vadz:
What about the following change instead:

```diff
diff --git a/src/aui/framemanager.cpp b/src/aui/framemanager.cpp
index 292bbdbfc8..abfbe7b9cb 100644
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -826,15 +826,15 @@ bool wxAuiManager::AddPane(wxWindow* window, const wxAuiPaneInfo& paneInfo)
     if (pinfo.best_size == wxDefaultSize &&
         pinfo.window)
     {
-        pinfo.best_size = pinfo.window->GetBestSize();
-
-        if (pinfo.min_size != wxDefaultSize)
-        {
-            if (pinfo.best_size.x < pinfo.min_size.x)
-                pinfo.best_size.x = pinfo.min_size.x;
-            if (pinfo.best_size.y < pinfo.min_size.y)
-                pinfo.best_size.y = pinfo.min_size.y;
-        }
+        // It's important to use the current window size and not the best size
+        // when adding a pane corresponding to a previously docked window: it
+        // shouldn't change its size if it's dragged and docked in a different
+        // place.
+        pinfo.best_size = pinfo.window->GetSize();
+
+        // But we still shouldn't make it too small.
+        pinfo.best_size.IncTo(pinfo.window->GetBestSize());
+        pinfo.best_size.IncTo(pinfo.min_size);
     }
 
 
```

? AFAICS it has the same effect, but it makes more sense to me to use the current size of the window when adding the pane than override `GetBestSize()` to return the current size.

vadz:
> If that works too, why not...

Well, it works for me in the sample and I think it should work in general too, but it would be great if you could please check if it works for your application too.

sz5000:
I've just tested your changes in my app. Seems to work. Behavior is as expected.

vadz:
Thanks for testing, I'll merge the other PR soon then.

PR #24893, opened 2024-10-18
Run macOS CI builds using macos-14 GitHub image

vadz:
"macos-12" one is deprecated and will be removed soon.

### DIFF
--- a/.github/workflows/ci_cmake.yml
+++ b/.github/workflows/ci_cmake.yml
@@ -66,11 +66,11 @@ jobs:
             runner: ubuntu-22.04
             cmake_generator: Unix Makefiles
             cmake_samples: ALL
-          - name: macOS 12 wxOSX
-            runner: macos-12
+          - name: macOS 14 wxOSX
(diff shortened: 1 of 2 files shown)


vadz:
Surprisingly, this worked without any other changes, so I'll push this soon if nobody sees anything wrong with it.

imciner2:
Just a note, the `macos-14` runners are all arm64 now, with `macos-13` being the last runners that are x86 native. This means I think the CMake macOS build has silently changed what is being built - switching from the x86_64 build to the arm64 build. (or at least the CMake build shows 
```
    Min OS Version required at runtime:                macOS 10.10 x86_64
```
before this PR, and now has
```
    Min OS Version required at runtime:                macOS 10.10 arm64
```
after it).

vadz:
Thanks for noticing this! We could cross-compile for x86_64 from ARM builders too, but I'm not sure if it's worth it. We still have non-CMake Mac builds using Intel wxMac, so maybe this is enough?

PR #24897, opened 2024-10-20
GTK3: Add support for wxTextCtrl::SearchText

Blake-Madden:
Add boundary test case also

### DIFF
--- a/src/gtk/textctrl.cpp
+++ b/src/gtk/textctrl.cpp
@@ -2077,6 +2077,11 @@ bool wxTextCtrl::GetStyle(long position, wxTextAttr& style)
 #ifdef __WXGTK3__
 wxTextSearchResult wxTextCtrl::SearchText(const wxTextSearch& search) const
 {
+    if ( !IsMultiLine() )
+    {
+        return wxTextSearchResult{};
+    }
+
     const GtkTextSearchFlags flags = static_cast<GtkTextSearchFlags>(
         search.m_matchCase ? GTK_TEXT_SEARCH_TEXT_ONLY :
                              (GTK_TEXT_SEARCH_TEXT_ONLY | GTK_TEXT_SEARCH_CASE_INSENSITIVE));
--- a/src/gtk/textctrl.cpp
+++ b/src/gtk/textctrl.cpp
@@ -2097,7 +2097,7 @@ wxTextSearchResult wxTextCtrl::SearchText(const wxTextSearch& search) const
     gboolean found = false;
 
     // verifies if the current match is whole-word
-    const auto confirmWholeWordMatch =
+    const auto confirmWhileWordMatch =
         [&selectionStart, &selectionEnd, &found, &selectionSearchPos]()
     {
         // we have a match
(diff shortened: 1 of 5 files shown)


Blake-Madden:
Relates to #23853

Blake-Madden:
Failure is from unrelated listctrl dragging test case.

src/gtk/textctrl.cpp
```diff
@@ -2074,6 +2074,108 @@ bool wxTextCtrl::GetStyle(long position, wxTextAttr& style)
...
+    if (search.m_direction == wxTextSearch::Direction::Down)
+    {
+        // search forward, beginning at the start (or user-provided position)
+        selectionSearchPos = textStart;
+        if (search.m_startingPosition != -1)
+            gtk_text_buffer_get_iter_at_offset(m_buffer, &selectionSearchPos,
+                                               static_cast<gint>(search.m_startingPosition));
+        while (!found)
+        {
+            found = gtk_text_iter_forward_search(&selectionSearchPos, search.m_searchValue.mb_str(),
```
vadz, line 2128: Minor but this should be
```suggestion
            found = gtk_text_iter_forward_search(&selectionSearchPos, search.m_searchValue.utf8_str(),
```

src/gtk/textctrl.cpp
```diff
@@ -2074,6 +2074,108 @@ bool wxTextCtrl::GetStyle(long position, wxTextAttr& style)
...
+                                                 flags, &selectionStart, &selectionEnd, nullptr);
+            if (found)
+            {
+                if (!search.m_wholeWord || confirmWholeWordMatch())
+                    break;
+                else
+                    continue;
```
vadz, line 2135: This doesn't look right: this `continue` is the same as `break`, because the loop condition is `!found` and `found` is true now. You probably wanted to do `found = false` but it would probably be better to remove the condition from the loop as having a loop with breaks and continue inside and a loop condition testing the same thing is pretty confusing.

I'd rather structure code like this:

```cpp
for (;;)
{
    found = search.m_direction == wxTextSearch::Direction::Down
            ? gtk_text_iter_forward_search(...)
            : gtk_text_iter_backward_search(...);

    // If we haven't found anything at all, we're done.
    if ( !found )
        break;

    // But if we did find something, we may need to check whether it was
    // a whole word.
    if ( !search.m_wholeWord || confirmWhileWordMatch()) )
        break;

    // Keep searching for the next match, maybe it will be a whole-word one.
}

vadz (commented):
Thanks, this looks good, but the loops are confusing and repeating them is IMHO undesirable. Could it be please simplified a little? TIA!

src/gtk/textctrl.cpp
Blake-Madden: Make suggested changes and test cases are passing.

Blake-Madden:
Failure appears to be an unrelated MSW webview compile error.

vadz:
Sorry for the delay, I finally got back to this and couldn't help simplifying it a bit further. I'm squash merging it now, please let me know if you have any objections to the changes I made.

And thanks again for your work!

PR #24898, opened 2024-10-20
Fix height of wxOwnerDrawnComboBox popup at different DPI

### LINKED ISSUE #24850: Height of `wxEnumProperty` dropdown list is too small at high DPI
The height of the dropdown list of a `wxEnumProperty` is too small at high DPI.

The issue can be reproduced in the propgrid sample with the following changes:

```patch
diff --git "a/samples/propgrid/propgrid.cpp" "b/samples/propgrid/propgrid.cpp"
index 38b5e8e169..ee8850e7b6 100644
--- "a/samples/propgrid/propgrid.cpp"
+++ "b/samples/propgrid/propgrid.cpp"
@@ -1356,6 +1356,12 @@ void FormMain::PopulateWithExamples ()
     wxBitmap bmpVal = wxArtProvider::GetBitmap(wxART_REMOVABLE);
     pg->SetPropertyImage("EnumProperty 5", bmpVal);
 
+    wxArrayString choices;
+    for ( int i = 0; i < 30; ++i )
+        choices.Add( wxString::Format("Item %d", i + 1) );
+    pg->Append( new wxEnumProperty("EnumProperty With Large List",wxPG_LABEL,
+        choices) );
+
     // Password property example.
     pg->Append( new wxStringProperty("Password",wxPG_LABEL, "password") );
     pg->SetPropertyAttribute( "Password", wxPG_STRING_PASSWORD, true );

```

At a DPI scaling of 100%:

wxEnumProperty-DropDown-dpi-100

At a DPI scaling of 200%:

wxEnumProperty-DropDown-dpi-200

MaartenBent:
Don't use a hard-coded height of 400px but estimate the height of 21 rows containing text (the orignal 400px at default DPI resulted in 21 items).
The actual height of a row depends on it's content and (custom) OnGetRowHeight implementation and is not known, and can also be different per row.
Also make the height DPI aware when no preferred height is given, and when the popup is empty.

Fixes https://github.com/wxWidgets/wxWidgets/issues/24850

And fix wxPropGrid editor appearance after a DPI change. 
The height of the propgrid item being edited does not correctly adjust to the new DPI, so recreate the control(s).

### DIFF
--- a/src/common/combocmn.cpp
+++ b/src/common/combocmn.cpp
@@ -116,7 +116,7 @@ wxCONSTRUCTOR_5( wxComboBox, wxWindow*, Parent, wxWindowID, Id, \
 
 #define BMP_BUTTON_MARGIN                       4
 
-#define DEFAULT_POPUP_HEIGHT                    400
+#define DEFAULT_POPUP_ITEMS                     21
 
 #define DEFAULT_TEXT_INDENT                     3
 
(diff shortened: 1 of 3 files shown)


vadz:
Thanks, I'll merge this soon.

It should probably be backported to 3.2 too, right?

MaartenBent:
yes! I think this can be backported.

PR #24900, opened 2024-10-22
wxMSW: Toolbar checked item bg color fix for dark mode

sz5000:
In dark mode and MSW, a checked toolbar item always has this light blue background color as you can see here:

image

This pull request fixes this and the new color better fits the dark design, as you can see here:

image

The flag `CDRF_NOTIFYPOSTPAINT` is actually not required for this change, but but you accept another pull request (#24901) that fixes the dark dropdown arrow issue in dark theme, then it is required.

### DIFF
--- a/src/msw/toolbar.cpp
+++ b/src/msw/toolbar.cpp
@@ -1694,6 +1694,7 @@ bool wxToolBar::MSWOnNotify(int WXUNUSED(idCtrl),
                 nmtbcd->clrTextHighlight = wxColourToRGB(GetForegroundColour());
                 nmtbcd->clrHighlightHotTrack = wxSysColourToRGB(wxSYS_COLOUR_HOTLIGHT);
 
+                // draw custom checked button background
                 if (nmtbcd->nmcd.uItemState & CDIS_CHECKED) {
                     wxColor color = wxSystemSettings::GetColour(wxSYS_COLOUR_HOTLIGHT).ChangeLightness(110);
                     HBRUSH br = CreateSolidBrush(wxColourToRGB(color));
--- a/src/msw/toolbar.cpp
+++ b/src/msw/toolbar.cpp
@@ -1694,6 +1694,15 @@ bool wxToolBar::MSWOnNotify(int WXUNUSED(idCtrl),
                 nmtbcd->clrTextHighlight = wxColourToRGB(GetForegroundColour());
                 nmtbcd->clrHighlightHotTrack = wxSysColourToRGB(wxSYS_COLOUR_HOTLIGHT);
 
+                if (nmtbcd->nmcd.uItemState & CDIS_CHECKED) {
+                    wxColor color = wxSystemSettings::GetColour(wxSYS_COLOUR_HOTLIGHT).ChangeLightness(110);
+                    HBRUSH br = CreateSolidBrush(wxColourToRGB(color));
+                    FillRect(nmtbcd->nmcd.hdc, &nmtbcd->nmcd.rc, br);
+                    DeleteObject(br);
+                    *result = CDRF_DODEFAULT | TBCDRF_USECDCOLORS | TBCDRF_HILITEHOTTRACK | TBCDRF_NOBACKGROUND;
+                    return true;
+                }
+
                 *result = CDRF_DODEFAULT | TBCDRF_USECDCOLORS | TBCDRF_HILITEHOTTRACK;
                 return true;
         }
--- a/src/msw/toolbar.cpp
+++ b/src/msw/toolbar.cpp
@@ -1700,7 +1700,7 @@ bool wxToolBar::MSWOnNotify(int WXUNUSED(idCtrl),
                     HBRUSH br = CreateSolidBrush(wxColourToRGB(color));
                     FillRect(nmtbcd->nmcd.hdc, &nmtbcd->nmcd.rc, br);
                     DeleteObject(br);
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for the fix, but do you understand why the problem isn't visible in the toolbar sample? It has a checkable icon (weirdly enough, it's the "help" one) and it appears correctly in the dark mode. What is different about the item in your case?

Also, if we do need this fix, shouldn't we restrict it to the dark mode only?

sz5000:
Seems the be different on Windows 11 and 10.

Inside a Windows 10 Test VM:

image

Under Windows 11:

image

Maybe it would be a good decision to make it customizable to that the dev can change this color.


Regarding "shouldn't we restrict it to the dark mode only":

Yes, it already is, 

```
case CDDS_PREPAINT:
    if ( !wxMSWDarkMode::IsActive() )
        break;

    *result = CDRF_NOTIFYITEMDRAW;
    return true;
```

CDRF_NOTIFYITEMDRAW is only set, if dark mode is active, so `case CDDS_ITEMPREPAINT` is only executed in dark mode.

sz5000:
I made another change and I think it's ok now and look good on Windows 10 and 11.  Please have a look.

src/msw/toolbar.cpp
```diff
@@ -1694,6 +1694,16 @@ bool wxToolBar::MSWOnNotify(int WXUNUSED(idCtrl),
                 nmtbcd->clrTextHighlight = wxColourToRGB(GetForegroundColour());
                 nmtbcd->clrHighlightHotTrack = wxSysColourToRGB(wxSYS_COLOUR_HOTLIGHT);
 
+                // draw custom checked button background
+                if ((nmtbcd->nmcd.uItemState & (CDIS_CHECKED|CDIS_HOT)) == CDIS_CHECKED) {
```
PBfordev, line 1698:
```suggestion
                if ( (nmtbcd->nmcd.uItemState & (CDIS_CHECKED | CDIS_HOT)) == CDIS_CHECKED ) 
                {
```

src/msw/toolbar.cpp
```diff
@@ -1694,6 +1694,16 @@ bool wxToolBar::MSWOnNotify(int WXUNUSED(idCtrl),
...
+                    wxColor color = wxSystemSettings::GetColour(wxSYS_COLOUR_HOTLIGHT).ChangeLightness(110);
```
PBfordev, line 1699:
```suggestion
                    const wxColor color = wxSystemSettings::GetColour(wxSYS_COLOUR_HOTLIGHT).ChangeLightness(110);
```

src/msw/toolbar.cpp
```diff
@@ -1694,6 +1694,16 @@ bool wxToolBar::MSWOnNotify(int WXUNUSED(idCtrl),
...
+                    HBRUSH br = CreateSolidBrush(wxColourToRGB(color));
+                    FillRect(nmtbcd->nmcd.hdc, &nmtbcd->nmcd.rc, br);
```
PBfordev, line 1701:
```suggestion
                    ::FillRect(nmtbcd->nmcd.hdc, &nmtbcd->nmcd.rc, br);
```

src/msw/toolbar.cpp
```diff
@@ -1694,6 +1694,16 @@ bool wxToolBar::MSWOnNotify(int WXUNUSED(idCtrl),
                 nmtbcd->clrTextHighlight = wxColourToRGB(GetForegroundColour());
                 nmtbcd->clrHighlightHotTrack = wxSysColourToRGB(wxSYS_COLOUR_HOTLIGHT);
 
+                // draw custom checked button background
+                if ((nmtbcd->nmcd.uItemState & (CDIS_CHECKED|CDIS_HOT)) == CDIS_CHECKED) {
+                    wxColor color = wxSystemSettings::GetColour(wxSYS_COLOUR_HOTLIGHT).ChangeLightness(110);
+                    HBRUSH br = CreateSolidBrush(wxColourToRGB(color));
```
PBfordev, line 1700:
```suggestion
                   AutoHBRUSH br(wxColourToRGB(color));
```

src/msw/toolbar.cpp
```diff
@@ -1694,6 +1694,16 @@ bool wxToolBar::MSWOnNotify(int WXUNUSED(idCtrl),
...
+                    FillRect(nmtbcd->nmcd.hdc, &nmtbcd->nmcd.rc, br);
+                    DeleteObject(br);
```
PBfordev, line 1702:
```suggestion
```

PBfordev (commented):
Sorry for all the nitpicking (I won't do the same for your other PR), but I believe that if possible, new code should meet the codebase style requirements, in particular, when being inserted into existing code which does.

src/msw/toolbar.cpp
sz5000: I've changed it.

src/msw/toolbar.cpp
sz5000: I've changed it.

src/msw/toolbar.cpp
sz5000: I've changed it.

src/msw/toolbar.cpp
sz5000: I've changed it.

src/msw/toolbar.cpp
sz5000: I've changed it.

vadz:
Thanks! I'm merging it with just an added comment and another minor style change to avoid overlong lines.

P.S. Sorry for missing the existing dark mode check and thanks for pointing it out.

sz5000:
Thanks, looks good so far.

PR #24902, opened 2024-10-22
Set info bar colors in create only if not set

sz5000:
We cannot set the background and foreground colors of the infobar before we create the infobar because these colors are set in the method for creating the infobar. If we change them later, then the close button will not fit well because it uses the background color of the infobar we want to change, so we would also need to change the background color of the close button, but there is no function to get the button. Simple solution is to just set FG and BG color if this hasn't happened yet.

Before:
image

After:
image


vadz (approved):
Thanks, this looks good to me, will merge soon.

PR #24904, opened 2024-10-22
wxQt: Implement maximum length for the multiline text control

AliKet:
It is unfortunate that `QTextEdit` does not have a simple API to enforce the maximum length (as is the case for `QLineEdit`) of text allowed in the control. And the only way to achieve this is:

1- Install an event filter on `QTextEdit` to monitor key presses and filter them out when the maximum length is reached.
2- Handle everything in the `textChanged()` signal

This PR chooses option (1) because option (2) is too late and simply truncating the text to the required length may result in a completely unexpected new value.

### DIFF
--- a/tests/controls/textctrltest.cpp
+++ b/tests/controls/textctrltest.cpp
@@ -28,9 +28,7 @@
     #include "wx/dataobj.h"
 #endif // wxUSE_CLIPBOARD
 
-#if defined(__WXGTK__) || defined(__WXQT__)
-    #include "waitfor.h"
-#endif
+#include "waitfor.h"
 
 #ifdef __WXQT__
 #include <QtGlobal>
@@ -89,6 +87,7 @@ private:
         // Rerun the text entry tests not specific to single line controls for
         // multiline ones now.
         wxTEXT_ENTRY_TESTS();
+        WXUISIM_TEST( MaxLength );
         SINGLE_AND_MULTI_TESTS();
 
 
@@ -275,6 +274,89 @@ void TextCtrlTestCase::ReadOnly()
 void TextCtrlTestCase::MaxLength()
 {
 #if wxUSE_UIACTIONSIMULATOR
+    wxUIActionSimulator sim;
+
+    if ( ms_style == wxTE_MULTILINE )
+    {
+#if defined(__WXMSW__) /*|| defined(__WXGTK3__)*/ || defined(__WXQT__)
+        delete m_text;
+        CreateText(wxTE_DONTWRAP | wxTE_RICH);
+        EventCounter maxlen(m_text, wxEVT_TEXT_MAXLEN);
+
+        m_text->SetMaxLength(250);
+        m_text->SetFocus();
+        wxYield();
+
+        const wxString linePattern = MakeLinePattern();
+
+        m_text->AppendText(wxString::Format("%s", linePattern));
+        m_text->SelectAll();
+        m_text->Copy();
+
+        m_text->SetInsertionPointEnd();
+
+        sim.Char(WXK_RETURN);
+        sim.Char('v', wxMOD_CONTROL); // Paste copied line.
+        wxYield();
+
+        CPPUNIT_ASSERT_EQUAL(0, maxlen.GetCount());
+
+        m_text->SetInsertionPointEnd();
+
+    #ifndef __WXMSW__
+        sim.Char(WXK_RETURN);
+        wxYield();
+    #endif
+        sim.Char('v', wxMOD_CONTROL); // Paste copied line (2nd time).
+        WaitFor("wxTextCtrl update", [&]() { return maxlen.GetCount() != 0; });
+
+        CPPUNIT_ASSERT_EQUAL(1, maxlen.GetCount()); // Maximum length reached.
+        maxlen.Clear();
+
+        sim.Text("7"); // Should be rejected.
+        WaitFor("wxTextCtrl update", [&]() { return maxlen.GetCount() != 0; });
+
+        CPPUNIT_ASSERT_EQUAL(1, maxlen.GetCount());
+        maxlen.Clear();
+
+        // Depending on the underlying system, the new line (NL) could be
+        // LF, CR or CRLF, and as a consequence the length of the last line
+        // should be 50 - 2*NL.
+
+        CPPUNIT_ASSERT_EQUAL(3, m_text->GetNumberOfLines());
+
+        int lineLength = m_text->GetLineText(0).length(); // 1st line
+        CPPUNIT_ASSERT_EQUAL(100, lineLength);
+
+        lineLength = m_text->GetLineText(1).length(); // 2nd line
+        CPPUNIT_ASSERT_EQUAL(100, lineLength);
+
+        lineLength = m_text->GetLineText(2).length(); // 3rd line
+        CPPUNIT_ASSERT( (lineLength == 46 || lineLength == 48) );
+
+        // Try to paste a long string into a shorter selection:
+
+        m_text->SetSelection(0, 20);  // selection is: 01234567890123456789
+        m_text->Copy();
+        m_text->SetSelection(15, 21); // selection is: 567890
+        m_text->Paste(); // Only the first six characters can actually be pasted.
+        WaitFor("wxTextCtrl update", [&]() { return maxlen.GetCount() != 0; });
+        const auto line = m_text->GetLineText(0);
+        CPPUNIT_ASSERT( (line[15].GetValue() == '0' &&
+                         line[20].GetValue() == '5' &&
+                         line[21].GetValue() == '1') );
+        CPPUNIT_ASSERT_EQUAL(1, maxlen.GetCount());
+        maxlen.Clear();
+
+        // Now, despite the maximum length of 250 set above, adding additional
+        // content to the control programmatically is still possible/allowed:
+        m_text->AppendText(wxString::Format("\n%s", linePattern));
+        CPPUNIT_ASSERT_EQUAL(4, m_text->GetNumberOfLines());
+        CPPUNIT_ASSERT_EQUAL(0, maxlen.GetCount());
+#endif // __WXMSW__ || __WXQT__
+    }
+    else // !wxTE_MULTILINE
+    {
 #ifdef __WXQT__
     #if (QT_VERSION < QT_VERSION_CHECK(5, 12, 0))
         WARN("wxEVT_TEXT_MAXLEN event is only generated if Qt version is 5.12 or greater");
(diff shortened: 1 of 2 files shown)


AliKet:
https://github.com/wxWidgets/wxWidgets/blob/09b6dad2ce647dec0867b020f2d687bfef862774/tests/controls/textctrltest.cpp#L664-L699

Unless I'm mistaken, this test case (as is) shouldn't pass under `wxMSW` because the maximum number of characters that can be entered into the control is **15000**, but somehow it accepts all inputs! Can anyone confirm this?

oneeyeman1:
@AliKet ,
Everything works as it should and it is by design.
Max length only applies to the text entered by user.
It is documented in MSDN.

Calls to Erite/AppendText() is not susceptible to this.

All other ports should follow..

(I was confused as well, but verified the behavior with MSDN).

AliKet:
> @AliKet , Everything works as it should and it is by design. Max length only applies to the text entered by user. It is documented in MSDN.
> 
> Calls to Erite/AppendText() is not susceptible to this.
> 
> All other ports should follow..
> 
> (I was confused as well, but verified the behavior with MSDN).

Yes, you are right... but then we have to:
- mention this (limitation?) in the documentation of `wxTextCtrl::SetMaxLength()`
- for completeness, add more tests to `TextCtrlTestCase::LongText()`
   * try to past some text after the maximum length is reached
   * try to enter some text using `wxUIActionSimulator`

oneeyeman1:
@AliKet ,
> > @AliKet , Everything works as it should and it is by design. Max length only applies to the text entered by user. It is documented in MSDN.
> > Calls to Erite/AppendText() is not susceptible to this.
> > All other ports should follow..
> > (I was confused as well, but verified the behavior with MSDN).
> 
> Yes, you are right... but then we have to:
> 
> * mention this (limitation?) in the documentation of `wxTextCtrl::SetMaxLength()`
> * for completeness, add more tests to `TextCtrlTestCase::LongText()`

Yes we should.

Go ahead and do that.
I'll rebate my GTK/OSX branches and submit them on top.

>   
>   * try to past some text after the maximum length is reached
>   * try to enter some text using `wxUIActionSimulator`

tests/controls/textctrltest.cpp
```diff
@@ -260,51 +276,109 @@ void TextCtrlTestCase::ReadOnly()
...
+        sim.Text("7"); // Should be rejected.
+        wxYield();
+
+        CPPUNIT_ASSERT_EQUAL(1, maxlen.GetCount());
+        maxlen.Clear();
+
+        // Depending on the underlying system, the new line (NL) could be
+        // LF, CR or CRLF, and as a consequence the length of the 1st and
+        // 2nd lines should be 100 + sizeof(NL) each, and the length of the
+        // last line should be 50 - 2*NL.
```
oneeyeman1, line 319: @AliKet ,
Did you check the MSW behavior for the new line character?

AliKet: The comments above describes what the control does internally, i.e. new line characters are also considered and counted by the control to decide whether the maximum length is reached or not. For example, if you have a control with MAXLEN set to 500 with 5 lines of text, and each line contains 100 chars, the last line will always have 100 - 4* sizeof(NL).
- If NL is just a LF or CR the result is 96 chars.
- If NL is CRLF the result is 92 chars.

tests/controls/textctrltest.cpp
```diff
@@ -260,51 +276,109 @@ void TextCtrlTestCase::ReadOnly()
...
+#if defined(__WXMSW__) || defined(__WXQT__)
+        m_text->SetMaxLength(250);
+
+        const wxString linePattern = MakeLinePattern();
+
+        m_text->AppendText(wxString::Format("%s\n", linePattern));
+        m_text->SelectAll();
+        m_text->Copy();
+
+        m_text->SetInsertionPointEnd();
```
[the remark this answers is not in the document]
AliKet: > @AliKet , Can you add the same with the cursor somewhere in the middle?
> 

I will add this test case but with text selected and the control has already reached the maxlen.

> Also, if I paste the text which exceeds the max length it should be allowed to paste up to the max length. I.e. if I have a text with 20 char length and the max length is 25, I should be able to paste {1,5] characters. Everything else on the paste will be cut off.

This is already covered by the added test: The max length is set to 250 and we try to paste a line (with 100 chars + newline) of text into the control multiple times.

tests/controls/textctrltest.cpp
```diff
@@ -275,51 +276,134 @@ void TextCtrlTestCase::ReadOnly()
 void TextCtrlTestCase::MaxLength()
 {
 #if wxUSE_UIACTIONSIMULATOR
+    wxUIActionSimulator sim;
+
+    if ( ms_style == wxTE_MULTILINE )
+    {
+#if defined(__WXMSW__) /*|| defined(__WXGTK3__)*/ || defined(__WXQT__)
```
AliKet, line 283: @oneeyeman1 Your PR works correctly with these tests. Just uncomment this line.

tests/controls/textctrltest.cpp
```diff
@@ -275,51 +274,134 @@ void TextCtrlTestCase::ReadOnly()
...
+        delete m_text;
+        CreateText(wxTE_DONTWRAP | wxTE_RICH);
```
AliKet, line 283: @vadz I don't know what to do here, because `wxMSW` would fail without `wxTE_RICH` or `wxTE_RICH2`?
i.e. SetValue()/ChangeValue() can not write more than `maxlen` if `> 0`. and this is not what would happen if the control is created with `wxTE_RICH[2]` flags?

tests/controls/textctrltest.cpp
oneeyeman1: @AliKet ,
Thx for checking. All there is to do is fix the comments. Will do that sometime next week.

Do you have access to a Mac? If not - I'll try that later...

AliKet: > Do you have access to a Mac? If not - I'll try that later...

Unfortunately not, sorry.

tests/controls/textctrltest.cpp
vadz: Sorry, I'm not sure I understand the problem, but if there is a discrepancy between the control behaviour with and without `wxTE_RICH[2]` flag, then it should be ideally fixed or, if this is impossible, at least documented.

AliKet: First, let me correct myself: `SetValue()/ChangeValue()` always succeed in setting/changing the contents of `wxTextCtrl` (with/without `wxTE_RICH[2]`) regardless of whether `SetMaxLength()` has been called or not.

The problem is with `AppendText()` which works inconsistently (with/without `wxTE_RICH[2]`) when `SetMaxLength()` is already called with non-zero length:

1- With `wxTE_RICH[2]` flag set: `AppendText()` always succeed.
2- Without `wxTE_RICH[2]` flag: `AppendText()` fails if the maximum length is reached.

vadz: You should be able to handle `EN_MAXTEXT`, i.e. tweak `wxTextCtrl::AdjustSpaceLimit()`, to allow appending text into plain EDIT control.

vadz: I don't really understand what's going on here because AFAICS we already had this test for wxMSW and it seems to have worked without `wxTE_RICH` before, but I'm fine with merging this wxQt PR without dealing with any wxMSW problems anyhow.

vadz:
@AliKet Now that you're back (great to see!), I'd like to ask what is the state of this one, should it be merged?

AliKet:
> @AliKet Now that you're back (great to see!), I'd like to ask what is the state of this one, should it be merged?

Thanks. My father just passed away, so I won't be available at the moment, I'm sorry. Moreover, I don't know what has changed in master while I was away so I need to check things again to make sure...

oneeyeman1:
@AliKet 
Sorry to hear this. My condolences.

AliKet:
> @AliKet Sorry to hear this. My condolences.

Thanks. Igor!

tests/controls/textctrltest.cpp
AliKet: I don't know if this has worked before (for the plain edit control) or not, but AFAICS `AdjustSpaceLimit()` is not supposed to change the hard limit (if explicitly set on the control) when handling `EN_MAXTEXT` message. Debugging also shows that Windows doesn't send any `EN_MAXTEXT` message for an edit control with `wxTE_RICH[2]` when calling `AppendText()` (see below) but it does for a plain edit control!
________
The MSDN is clear on the `EM_LIMITTEXT` message and says this: 

> The EM_LIMITTEXT message does not affect the length of the text copied into the edit control by the WM_SETTEXT message.

 But what `AppendText()` ends up sending (inside `DoWriteText()`) is the `EM_REPLACESEL` message and the MSDN says nothing about that when `EM_LIMITTEXT` is in effect.

vadz: Do you have a simple diff to the current master that would show the problem with the plain text control not accepting longer text by chance?

AliKet:
```C++
index 5f32257c6b..30ce88592f 100644
@@ -64,6 +64,8 @@ public:
     void OnQuit(wxCommandEvent& event);
     void OnAbout(wxCommandEvent& event);
 
+    wxTextCtrl* m_text = nullptr;
+
 private:
     // any class wishing to process wxWidgets events must use this macro
     wxDECLARE_EVENT_TABLE();
@@ -170,6 +172,16 @@ MyFrame::MyFrame(const wxString& title)
     SetSizer(sizer);
 #endif // wxUSE_MENUBAR/!wxUSE_MENUBAR
 
+    const long style = wxTE_MULTILINE | wxTE_DONTWRAP /*| wxTE_RICH*/;
+    m_text = new wxTextCtrl(this, wxID_ANY, "", wxDefaultPosition, wxDefaultSize, style);
+    m_text->Bind(wxEVT_TEXT_MAXLEN, [](wxCommandEvent& event)
+        {
+            wxLogMessage("Max length reached!");
+            event.Skip();
+        });
+
+    m_text->SetMaxLength(20);
+
 #if wxUSE_STATUSBAR
     // create a status bar just for fun (by default with 1 pane only)
     CreateStatusBar(2);
@@ -188,16 +200,5 @@ void MyFrame::OnQuit(wxCommandEvent& WXUNUSED(event))
 
 {
-                 (
-                    "\n"
-                 ),
-                 this);
+    m_text->AppendText(wxString("0123456789"));
 }

```
With this patch, continuously pressing the **F1** key will append `"0123456789"` to the control.
If the control is created with the `wxTE_RICH[2]` flag, `AppendText()` will always succeed. Otherwise, a message box will appear when the maximum number of characters allowed is reached.

vadz: Ah, so the problem is that `SetMaxLength()` does _not_ work for rich edit? Because the described behaviour seems correct in the plain edit case... Or am I still missing something?

AliKet: `SetMaxLength()` should only affect text entered interractively by the user. Text entered programmatically using `SetValue()`, `ChangeValue()` or `AppendText()` should still be allowed. Here, `AppendText()` is the only method that fails to append more text than the maximum allowed in a plain edit control. The others always succeed.

vadz: Oops, yes, you're right, of course, sorry. This is a bug in `wxTextCtrl`, I'll create a separate issue for it.

vadz: I think it's better to skip this test under MSW, adding `wxTE_RICH` is not the best workaround because rich controls don't implement max length at all.

vadz:
Sorry to return to this, but I'm still not sure what should be done here?

AliKet:
replaced with #25272

Issue #23940, opened 2023-10-04
crash due to possibly nested wxLogGui::Flush calls


marekr:
This is another strange crash captured in kicad via sentry.

The only thing from KiCad appearing in this stack race is APP_KICAD::OnRun() is just a return wxApp::OnRun();

It appears wxLogGui::Flush gets called twice, once at the beginning and a second time because of the nested event loop handling due to the first call spawning a message box modal.

This runs counter to the statement in wxLogGui::Flush where this is stated
```
    // avoid showing other log dialogs until we're done with the dialog we're
    // showing right now: nested modal dialogs make for really bad UI!
    Suspend();
```
but I'm assuming this scenario wasn't imagined

```
 	wxmsw32u_core_vc_custom.dll!wxWndProc(HWND__ * hWnd, unsigned int message, unsigned __int64 wParam, __int64 lParam) Line 2948	C++
 	[External Code]	
 	[Inline Frame] wxmsw32u_core_vc_custom.dll!wxTopLevelWindowMSW::DoShowWindow(int) Line 549	C++
 	wxmsw32u_core_vc_custom.dll!wxTopLevelWindowMSW::Show(bool show) Line 633	C++
 	wxmsw32u_core_vc_custom.dll!wxDialog::Show(bool show) Line 165	C++
 	wxmsw32u_core_vc_custom.dll!wxDialog::ShowModal() Line 187	C++
>	wxmsw32u_core_vc_custom.dll!wxLogGui::DoShowMultipleLogMessages(const wxArrayString & messages, const wxArrayInt & severities, const wxArrayLong & times, const wxString & title, int style) Line 263	C++
 	wxmsw32u_core_vc_custom.dll!wxLogGui::Flush() Line 329	C++
 	wxbase32u_vc_custom.dll!wxLog::SetActiveTarget(wxLog * pLogger) Line 568	C++
 	[Inline Frame] wxbase32u_vc_custom.dll!DoCommonPreCleanup() Line 403	C++
 	wxbase32u_vc_custom.dll!wxEntryCleanupReal() Line 446	C++
 	wxbase32u_vc_custom.dll!wxEntryCleanup() Line 366	C++
 	wxmsw32u_core_vc_custom.dll!wxApp::OnEndSession(wxCloseEvent & __formal) Line 859	C++
[...]

PR #24906, opened 2024-10-23
Miscellaneous end session fixes

vadz:
This is mostly for wxMSW (see #23940, which this still doesn't fix, but which motivated these changes), but also touches wxGTK because I wanted to have a way to get the number of currently open modal dialogs in common code and this rendered wxGTK-specific code doing the same thing unnecessary.

@vslavik I thought you might be interested in having a look at the changes here, as this is also somewhat related to your changes in #24903.

@marekr I don't think this fixes KiCad problems, but it might be still worth integrating these changes on your side, they shouldn't hurt (famous last words...) and just might help.


vadz:
I think the simplest way to make sure we never process this event twice is to not rely on wx event handling but just hard code what `wxApp::OnEndSession()` does in a function that will be called no matter what, i.e. whether the event is handled or skipped.

What do you think?

But the linked Poedit crash seems to happen due to a completely unexpected reentrancy into the message loop from inside `::DestroyWindow()` and I have no idea what to do about this, nothing here is going to help with this.

vadz:
> Sorry, I'm not sure I follow?

I think this might be the case when the code could be more clear than my explanations, especially because I've simplified it to do only 2(i) from above because, after thinking more about this, I think this makes the rest unnecessary, i.e. it looks like this on its own should take care of the problems identified so far.

But please have a look at the latest version and tell me what do you think about it. TIA!

src/msw/window.cpp
```diff
@@ -4351,9 +4359,14 @@ bool wxWindowMSW::HandleEndSession(bool endSession, long logOff)
     if ( !endSession )
         return false;
```
vslavik, line 4362: This branch should be amended with `gs_gotEndSession = false;` so that any subsequent new attempts to end the session work correctly.

vslavik:
> But please have a look at the latest version and tell me what do you think about it. TIA!

Looks reasonable to me (and better than https://github.com/wxWidgets/wxWidgets/pull/24903), thanks!

src/msw/window.cpp
vadz: We could do it, but I think it's unnecessary and potentially confusing because it contradicts the claim that once `gs_gotEndSession` is set to true the application is sure to exit. Am I missing some scenario in which you think we may get `WM_ENDSESSION(TRUE)` first followed by `WM_ENDSESSION(FALSE)` later?

vadz:
I think we should merge this already and see if it helps, but, just for the record, I still have a couple of ideas about other small changes:

1. We should send `wxEVT_QUERY_END_SESSION` to the window itself in `wxWindow::HandleQueryEndSession()` instead of sending it to wxApp (of course, it will still end up in wxApp if the window doesn't handle it) and have a default handler for it in `wxTLW` which would call `Close()` to ensure that we close all TLWs and not just the main one.
1. If we can confirm that modern Windows versions kill the process once `WM_ENDSESSION` returns, I think we ought to call `exit()` from `wxWindow::HandleEndSession()` itself after the user-defined handler returns because it looks strictly better to execute the global objects destructors than to not do it.

Please let me know if you think it's worth adding either/both of those to this PR, this would be simple to do and I could do it now (unlike more extensive changes, like adding support for the restart manager or providing block reason, which would be definitely worth doing but for which I don't really have time).

Otherwise I'll just merge the current version relatively soon (with or without `gs_gotEndSession` reset discussed in a separate conversation above).

src/msw/window.cpp
vslavik: Sorry, I got confused about `WM_ENDSESSION` vs `WM_QUERYENDSESSION`. You're right, sorry :(

vslavik:
> Please let me know if you think it's worth adding either/both of those to this PR, this would be simple to do

IMHO they are worth doing, but maybe not so simple and not worth delaying this PR?

> We should send wxEVT_QUERY_END_SESSION to the window itself in wxWindow::HandleQueryEndSession() instead of sending it to wxApp

Per-window handling would be convenient to have, but:
1. I am not sure how `WM_QUERYENDSESSION` is sent, if to all TLWs right away or sequentially; MSDN also talks about sending the message to application, so this may be intentionally vague
2. I do know that the first non-vetoed message results in immediately going to the `WM_ENDSESSION` phase, i.e. handling at wxApp level and iterating over TLWs seems safer
3. It's used in wxOSX too (where the native API is application-level)

> If we can confirm that modern Windows versions kill the process once WM_ENDSESSION returns

 I'm not sure how to reliably confirm Windows behavior w.r.t. killing the process

vadz:
> Per-window handling would be convenient to have, but:
> 
>     1. I am not sure how `WM_QUERYENDSESSION` is sent, if to all TLWs right away or sequentially; MSDN also talks about sending the message to application, so this may be intentionally vague

I'm pretty sure it's to all TLWs, sequentially, just as all the other broadcast messages.
 
>     2. I do know that the first non-vetoed message results in immediately going to the `WM_ENDSESSION` phase, i.e. handling at wxApp level and iterating over TLWs seems safer
> 
>     3. It's used in wxOSX too (where the native API is application-level)

Oh, this is a good point, thanks.

So finally we should indeed just iterate over all TLWs ourselves in `wxApp::OnQueryEndSession()`. But we still need to find some way of avoiding generating it more than once and this looks tricky (we could use the same `gs_gotEndSession` when it is not vetoed, but not if it is vetoed...).

> > If we can confirm that modern Windows versions kill the process once WM_ENDSESSION returns
> 
> I'm not sure how to reliably confirm Windows behavior w.r.t. killing the process

In principle this is simple: just open a file, and write to it something from `wxEVT_END_SESSION` handler and from some global dtor (always flushing it after writing), then check what ends up in the file in different scenarios. Missing messages would be a reliable indicator of Windows having killed the process before this point is reached, IMO.

vslavik:
> But we still need to find some way of avoiding generating it more than once and this looks tricky (we could use the same gs_gotEndSession when it is not vetoed, but not if it is vetoed...).

I just use a 5s timeout in user code now. But this is actually where I got my wires crossed earlier: we could have a `g_gotQueryEndSession` that is reset back to false when handling `WM_ENDSESSION(FALSE)` (which is sent when the OS cancels its session-ending attempt).

vadz:
> I just use a 5s timeout in user code now. But this is actually where I got my wires crossed earlier: we could have a `g_gotQueryEndSession` that is reset back to false when handling `WM_ENDSESSION(FALSE)` (which is sent when the OS cancels its session-ending attempt).

Ah, I didn't realize we could use this, thanks.

I've implemented what (I believe) you proposed now, but didn't have time to test it because I have to leave now, so that's all from me for today. Thanks again for the review and advice!

vslavik:
> I've implemented what (I believe) you proposed now

Yes, that's what I meant, thanks! It seems to behave correctly in my testing. I was able to clarify this bit of behavior (at least on Win11, with rmlogotest):

> > I am not sure how `WM_QUERYENDSESSION` is sent, if to all TLWs right away or sequentially; MSDN also talks about sending the message to application, so this may be intentionally vague
>
> I'm pretty sure it's to all TLWs, sequentially, just as all the other broadcast messages.

When I have a modal window shown in response to the first `WM_QUERYENDSESSION`, asking user for confirmation, then if the user doesn't respond within 5 seconds, another `WM_QUERYENDSESSION` is sent to another TLW (and is handled by the inner event loop of the modal message box) even though the first one wasn't responded to yet. So it's not strictly sequential, at least with rmlogotest.

However, I wasn't able to trigger any bad behavior even though I tried: replying to the 2nd message with Allow, as this PR's code does, still resulted in WM_ENDSESSION(FALSE), presumably because the 1st message's handling took so long. This is strange, because it is seemingly contrary to the docs, but it's unclear to me if this is due to using rmlogotest or is "real" behavior.

vadz:
Thanks for testing!

I am going to merge this soon if there are no objections from anybody else.

PR #24907, opened 2024-10-24
Make wxWARN_UNUSED work with Clang (again)

lanurmi:
Checking for availability of 'warn_unused' using `__has_cpp_attribute` evaluates to false with Clang, even when the attribute is actually available. `__has_attribute` gives the correct result, so use it instead, despite the aesthetical inconsitency with other checks.

See also: PR #24833

I wish the CI could be used to ensure that something *does* fail to compile, but no idea if that is possible.


lanurmi:
> You can grep for `failtest` in `tests/test.bkl` to see how to add a test checking that compiling something fails.

Thanks; I'll look into it, but I'll submit a separate PR in case I'm successful. This PR should be ready for merge.

Issue #24892, opened 2024-10-18
AUI drop indicator drawn at completely wrong location


vadz:
Other issue (not solved by this patch)  is e.g. the following
![image](https://github.com/user-attachments/assets/17193453-7f7d-406e-a743-1534ee5abc40)

Drop indicator of aui manager (blue rect) is at the wrong position when we try to drag und drop the right tab back to the left one (mouse position here was some pixels next to the 65_pages tab).

_Originally posted by @sz5000 in https://github.com/wxWidgets/wxWidgets/issues/24890#issuecomment-2422392975_

vadz:
I've tried debugging this but unsuccessfully so far. Or, rather, I only have negative results:

1. This is not due to `CalculateHintRect()` returning a wrong value, it never returns coordinates of such a narrow rectangle.
2. This is not due to timer-based fade in, the problem is still visible when not using it.
3. This is not even wxMSW-specific, I see the same problem under Linux.

It really looks like we're doing something wrong with the overlay somewhere, but I don't see where yet.

MaartenBent:
It happens when drawing the overlay over the tabs instead of over the panel. Converting coordinates in `ClientToScreen` seems to cause it:
https://github.com/wxWidgets/wxWidgets/blob/c301af42f923a9997e99bf8d3b8f00df667f88f9/src/aui/auibook.cpp#L2740

Also it should probably only draw the hint over the tabs, and not continuously switch between the panel and tabs while moving.

vadz:
Thanks! It's nice to get back and see that the problem is already fixed :-)

This line is indeed the culprit, so I'm going to remove it and create a PR with my other related changes soon, but I am not sure how do you suggest to avoid switching between the panel and the tabs? Do you think we shouldn't be showing the hint for dest tabs at all?

MaartenBent:
I am not sure if it should be fixed, or what the fix should be. But when I drag the right panel it keeps switching between these two overlays when the mouse is over the tab bar
<img src="https://github.com/user-attachments/assets/346d7efe-454f-4c9c-8585-2e455b76965a">
<img src="https://github.com/user-attachments/assets/4e86a36a-60ec-48d2-83f7-a8aec42412cc">

vadz:
Yes, I've seen this, and it would be nice to improve this somehow, but it doesn't look _wrong_, just suboptimal.

[...]

PR #24911, opened 2024-10-26
wxAUI notebook pages dragging-related fixes

vadz:
This fixes user-visible #24890 and #24892, but, as discussed in the latter, I'm not sure if we shouldn't restore the old behaviour of `ShowHint()` and pass it the rectangle in screen coordinates.

### DIFF
--- a/src/aui/auibook.cpp
+++ b/src/aui/auibook.cpp
@@ -1550,6 +1550,16 @@ public:
         m_tabCtrlHeight = h;
     }
 
+    // As we don't have a valid HWND, the base class version doesn't work for
+    // this window, so override it to return the appropriate DPI.
+    wxSize GetDPI() const override
+    {
+        if (!m_tabs)
+            return wxWindow::GetDPI();
+
+        return m_tabs->GetDPI();
+    }
+
 protected:
     void DoSetSize(int x, int y,
                    int width, int height,
--- a/src/aui/auibook.cpp
+++ b/src/aui/auibook.cpp
@@ -2709,9 +2709,7 @@ void wxAuiNotebook::OnTabDragMotion(wxAuiNotebookEvent& evt)
 
                 if (nb != this)
                 {
-                    wxRect hint_rect = tab_ctrl->GetClientRect();
-                    tab_ctrl->ClientToScreen(&hint_rect.x, &hint_rect.y);
-                    m_mgr.UpdateHint(hint_rect);
+                    m_mgr.UpdateHint(tab_ctrl->GetScreenRect());
                     return;
                 }
             }
@@ -2745,9 +2743,7 @@ void wxAuiNotebook::OnTabDragMotion(wxAuiNotebookEvent& evt)
 
     if (dest_tabs)
     {
-        wxRect hint_rect = dest_tabs->GetRect();
-        ClientToScreen(&hint_rect.x, &hint_rect.y);
-        m_mgr.UpdateHint(hint_rect);
+        m_mgr.UpdateHint(dest_tabs->GetScreenRect());
     }
     else
     {
--- a/src/aui/auibook.cpp
+++ b/src/aui/auibook.cpp
@@ -2628,7 +2628,6 @@ void wxAuiNotebook::OnTabDragMotion(wxAuiNotebookEvent& evt)
 {
     wxPoint screen_pt = ::wxGetMousePosition();
     wxPoint client_pt = ScreenToClient(screen_pt);
-    wxPoint zero(0,0);
 
     wxAuiTabCtrl* src_tabs = (wxAuiTabCtrl*)evt.GetEventObject();
     wxAuiTabCtrl* dest_tabs = GetTabCtrlFromPoint(client_pt);
@@ -2752,7 +2751,7 @@ void wxAuiNotebook::OnTabDragMotion(wxAuiNotebookEvent& evt)
     }
     else
     {
-        m_mgr.DrawHintRect(m_dummyWnd, client_pt, zero);
+        m_mgr.DrawHintRect(m_dummyWnd, client_pt);
     }
 }
 
@@ -2911,10 +2910,8 @@ void wxAuiNotebook::OnTabEndDrag(wxAuiNotebookEvent& evt)
         }
         else
         {
-            wxPoint zero(0,0);
             wxRect rect = m_mgr.CalculateHintRect(m_dummyWnd,
-                                                  mouse_client_pt,
-                                                  zero);
+                                                  mouse_client_pt);
             if (rect.IsEmpty())
             {
                 // there is no suitable drop location here, exit out
--- a/src/aui/auibook.cpp
+++ b/src/aui/auibook.cpp
@@ -2712,7 +2712,7 @@ void wxAuiNotebook::OnTabDragMotion(wxAuiNotebookEvent& evt)
                 {
                     wxRect hint_rect = tab_ctrl->GetClientRect();
                     tab_ctrl->ClientToScreen(&hint_rect.x, &hint_rect.y);
-                    m_mgr.ShowHint(hint_rect);
+                    m_mgr.UpdateHint(hint_rect);
                     return;
                 }
             }
@@ -2748,7 +2748,7 @@ void wxAuiNotebook::OnTabDragMotion(wxAuiNotebookEvent& evt)
     {
         wxRect hint_rect = dest_tabs->GetRect();
         ClientToScreen(&hint_rect.x, &hint_rect.y);
-        m_mgr.ShowHint(hint_rect);
+        m_mgr.UpdateHint(hint_rect);
     }
     else
     {
--- a/include/wx/aui/auibook.h
+++ b/include/wx/aui/auibook.h
@@ -334,6 +334,8 @@ public:
 
     const wxAuiManager& GetAuiManager() const { return m_mgr; }
 
+    void SetManagerFlags(unsigned int flags) { m_mgr.SetFlags(flags); }
+
     // Sets the normal font
     void SetNormalFont(const wxFont& font);
 
--- a/include/wx/aui/framemanager.h
+++ b/include/wx/aui/framemanager.h
@@ -485,12 +485,12 @@ public:
     wxRect CalculateHintRect(
                  wxWindow* paneWindow,
                  const wxPoint& pt,
-                 const wxPoint& offset);
+                 const wxPoint& offset = wxPoint{});
 
     void DrawHintRect(
                  wxWindow* paneWindow,
                  const wxPoint& pt,
-                 const wxPoint& offset);
+                 const wxPoint& offset = wxPoint{});
 
     void UpdateHint(const wxRect& rect);
 
--- a/include/wx/aui/framemanager.h
+++ b/include/wx/aui/framemanager.h
@@ -492,6 +492,10 @@ public:
                  const wxPoint& pt,
                  const wxPoint& offset);
 
+    void UpdateHint(const wxRect& rect);
+
+    // These functions are public for compatibility reasons, but should never
+    // be called directly, use UpdateHint() above instead.
     virtual void ShowHint(const wxRect& rect);
     virtual void HideHint();
 
--- a/interface/wx/aui/auibook.h
+++ b/interface/wx/aui/auibook.h
@@ -321,6 +321,28 @@ public:
     virtual bool SetFont(const wxFont& font);
 
     /**
+        Sets the flags for the wxAuiManager used by wxAuiNotebook.
+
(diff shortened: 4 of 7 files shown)


AliKet:
[ef2510c](https://github.com/wxWidgets/wxWidgets/pull/24911/commits/ef2510ccea1c6d0084104652e4189aab36bc52c2) shouldn't be included (see https://github.com/wxWidgets/wxWidgets/issues/24892#issuecomment-2439712989)

AliKet:
I'm just wondering if it's a good idea to break the `ShowHint()` API in **wx3.3** and make the rectangle it accepts in client coordinates instead of screen coordinates and update the aui code accordingly?

vadz:
Thanks, I've updated the PR to revert the change in `ShowHint()`.

I don't think it's worth changing its semantics and I even kept it for the new `UpdateHint()` too because it would be confusing to use different coordinates for the two closely related functions.

src/aui/auibook.cpp
```diff
@@ -2712,7 +2712,7 @@ void wxAuiNotebook::OnTabDragMotion(wxAuiNotebookEvent& evt)
                 {
                     wxRect hint_rect = tab_ctrl->GetClientRect();
                     tab_ctrl->ClientToScreen(&hint_rect.x, &hint_rect.y);
-                    m_mgr.ShowHint(hint_rect);
+                    m_mgr.UpdateHint(tab_ctrl->GetRect());
                     return;
```
AliKet, line 2715: I don't understand this change:
- `hint_rect` is calculated but not used!
- We pass a rectangle (in client coordinates) to `UpdateHint()`, but it accepts a rectangle in screen coordinates? or, the intention here is to use `tab_ctrl->GetScreenRect()` instead?

vadz: Oops, thanks for noticing this. This is just a bad rebase conflict resolution, it should be `UpdateHint(hint_rect)`, of course.

I thought about changing the code to use `GetScreenRect()` too, but decided it wasn't worth doing it, but maybe it is...

AliKet:
Please, let me know if I should remove this [line](https://github.com/wxWidgets/wxWidgets/blob/b1442f2d4c5554ee587eb9a707caa95716b8ba94/src/aui/framemanager.cpp#L4494) in PR #24921 because it would be incorrect for `wxQt` when applying the aforementioned PR. Or, would it be much more sense to remove it here?

TIA!

vadz:
@AliKet I don't understand why do we need this line, so it should probably be removed here, but how exactly can I test this code, i.e. what should be done to see the current behaviour?

AliKet:
> @AliKet I don't understand why do we need this line, so it should probably be removed here, but how exactly can I test this code, i.e. what should be done to see the current behaviour?

Just disable **Live Resize Update** under **Options** and try to resize some panes:

Screenshot from 2024-10-28 17-48-24

With this PR and #24921 the resize hint under `wxQt` is drawn correctly at the right offset.

vadz:
No, they actually don't... Well, maybe wxGTK does (I think it always uses origin of (0,0)) but if you apply this:

```diff
iff --git a/samples/aui/auidemo.cpp b/samples/aui/auidemo.cpp
index 96a51c4ba0..48c6ed62cf 100644
--- a/samples/aui/auidemo.cpp
+++ b/samples/aui/auidemo.cpp
@@ -695,6 +695,10 @@ MyFrame::MyFrame(wxWindow* parent,
                  long style)
         : wxFrame(parent, id, title, pos, size, style)
 {
+    auto tb = CreateToolBar(wxTB_TEXT);
+    tb->AddTool(wxID_ABOUT, "About", wxIcon(sample_xpm));
+    tb->Realize();
+
     // tell wxAuiManager to manage this frame
     m_mgr.SetManagedWindow(this);
 
```

to make the origin non-0 in wxMSW (admittedly, it's uncommon to use normal toolbar in a program also using wxAUI, but it's definitely possible), you can see that the position is wrong there too, **both** with and without that line.

To make things work correctly in wxMSW you actually need to use

```cpp
                rect.SetPosition(rect.GetPosition() - m_frame->GetClientAreaOrigin());
```

and now wxQt behaviour makes even less sense to me, how could it have needed `+` here?

AliKet:
> 
> To make things work correctly in wxMSW you actually need to use
> 
> ```c++
>                 rect.SetPosition(rect.GetPosition() - m_frame->GetClientAreaOrigin());
> ```
> 
> and now wxQt behaviour makes even less sense to me, how could it have needed `+` here?

For `wxQt` I missed a subtlety when I made this PR which should have looked like this:

```
diff --git a/src/qt/dcclient.cpp b/src/qt/dcclient.cpp
index 46247c09a1..675d127b20 100644
--- a/src/qt/dcclient.cpp
+++ b/src/qt/dcclient.cpp
@@ -115,10 +115,10 @@ wxClientDCImpl::wxClientDCImpl( wxDC *owner, wxWindow *win )
 {
      if ( m_ok )
      {
-        m_qtPainter->translate( wxQtConvertPoint(win->GetClientAreaOrigin()) );
-
         m_qtPainter->setClipRect( wxQtConvertRect(win->GetClientRect()),
                                   m_clipping ? Qt::IntersectClip : Qt::ReplaceClip );
+
+        m_qtPainter->translate( wxQtConvertPoint(win->GetClientAreaOrigin()) );
     }
 }
 
```

So apparently this line
```c++
                 rect.SetPosition(rect.GetPosition() - m_frame->GetClientAreaOrigin());
```
 is only needed under wxMSW (wxOSX?)

vadz:
Ugh, something is very wrong here. First of all, thinking more about this, using `-` for wxMSW really doesn't make sense: either the offset is accounted for by `wxClientDC` and `wxOverlayDC` based on it and then this line shouldn't be used at all or it isn't and in this case `+` would be correct to compensate for it. Second, somehow using `-` does work correctly when starting the sample but if you undock the AUI toolbars (which do not count to the client area offset, AFAICS), it doesn't work correctly for the splitter between the tree and the notebook any — while it still does for the splitter between them both and the text at the bottom.

So it looks like action part position itself is incorrect, maybe?

But in any case, the first question to ask is what is the point (0,0) of `wxOverlayDC` associated with the window. IMO it should be the same as the point (0,0) of `wxClientDC` and hence must be just below the toolbar, i.e. at `GetClientAreaOrigin()` in "real" coordinates. Is this the case for wxQt?

vadz:
FWIW here is a test showing that currently `wxOverlay` origin is broken under MSW:

```cpp
#include <wx/wx.h>
#include <wx/overlay.h>

{
public:
    MyFrame() : wxFrame(nullptr, wxID_ANY, "Test")
    {
        auto tb = CreateToolBar();
        tb->AddTool(wxID_ABOUT, "About", wxICON(sample));
        tb->Realize();

        Show();

        wxOverlayDC dc(m_overlay, this);
        dc.Clear();
        dc.SetBrush(*wxRED_BRUSH);
        dc.SetPen(*wxTRANSPARENT_PEN);
        dc.DrawRectangle(0, 0, 200, 100);
    }

    wxOverlay m_overlay;
};

class MyApp : public wxApp
{
public:
    bool OnInit() override
    {
        new MyFrame();
        return true;
    }
};
wxIMPLEMENT_APP(MyApp);
```

If you run this, you can see that the overlay is offset from the toolbar by toolbar height.

AliKet:
> But in any case, the first question to ask is what is the point (0,0) of `wxOverlayDC` associated with the window. IMO it should be the same as the point (0,0) of `wxClientDC` and hence must be just below the toolbar, i.e. at `GetClientAreaOrigin()` in "real" coordinates. Is this the case for wxQt?

Yes and here is how it looks like if we give it a cyan colour (with the corrected patch I mentioned above applied):

Screenshot from 2024-10-28 21-44-42

wxGTK3 using Wayland is exactly the same.

AliKet:
> FWIW here is a test showing that currently `wxOverlay` origin is broken under MSW:
> 
> ```c++
> #include <wx/wx.h>
> #include <wx/overlay.h>
> 
> class MyFrame : public wxFrame
> {
> public:
>     MyFrame() : wxFrame(nullptr, wxID_ANY, "Test")
>     {
>         auto tb = CreateToolBar();
>         tb->AddTool(wxID_ABOUT, "About", wxICON(sample));
>         tb->Realize();
> 
>         Show();
> 
>         wxOverlayDC dc(m_overlay, this);
>         dc.Clear();
>         dc.SetBrush(*wxRED_BRUSH);
>         dc.SetPen(*wxTRANSPARENT_PEN);
>         dc.DrawRectangle(0, 0, 200, 100);
>     }
> 
>     wxOverlay m_overlay;
> };
> 
> class MyApp : public wxApp
> {
> public:
>     bool OnInit() override
>     {
>         new MyFrame();
>         return true;
>     }
> };
> wxIMPLEMENT_APP(MyApp);
> ```
> 
> If you run this, you can see that the overlay is offset from the toolbar by toolbar height.

Thanks for the code, I'll try to take a look at it tomorrow as I'm exhausted right now.

Regards.

vadz:
> I'll try to take a look at it tomorrow as I'm exhausted right now.

I'll look at wxMSW myself, something seems very wrong there as `wxClientDC` is not offset correctly neither...

MaartenBent:
Maybe you already found the problem, but I was also intrigued. I think the problem is `GetClientRect`/`GetClientAreaOrigin` adds the `wxToolbar` height (or width) to the origin, but when converting client coordinates to screen coordinates this is not accounted for.
And Windows doesn't know that the origin is shifted, so the screen coordinates will be shifted as well.

vadz:
Thanks, there is indeed this, but also the fact that the adjustment for the client area origin in `wxClientDCImpl::InitDC()` is not done for wxMSW, while IMO it should be.

This won't be difficult to fix but it's strange that this hasn't been noticed until now, I spent some time looking for the changes that could have broken it but it really looks like it always worked like this. I guess people don't use `wx{Client,Paint}DC` with frames with toolbars that often, but still...

vadz:
Just to be clear (as much for my own benefit as for anyone else): fixing `wxOverlay` right now is as simple as doing this:

```diff
diff --git a/src/msw/overlay.cpp b/src/msw/overlay.cpp
index 9d8240efa3..449a906d29 100644
--- a/src/msw/overlay.cpp
+++ b/src/msw/overlay.cpp
@@ -161,7 +161,7 @@ void wxOverlayImpl::Init(wxDC* dc, int , int , int , int )
     m_window = dc->GetWindow();
 
     // The rectangle must be in screen coordinates
-    m_rect = m_window->GetClientRect();
+    m_rect = m_window->GetClientSize();
     m_window->ClientToScreen(&m_rect.x, &m_rect.y);
 
     if ( IsUsingConstantOpacity() )
```

but if you replace `wxOverlayDC dc(m_overlay, this)` in the example above with `wxClientDC dc(this)` you can see that it's also wrong and to fix it, and `wxPaintDC`, we need

```diff
diff --git a/src/msw/dcclient.cpp b/src/msw/dcclient.cpp
index cc2549ca7c..5818562618 100644
--- a/src/msw/dcclient.cpp
+++ b/src/msw/dcclient.cpp
@@ -188,9 +188,8 @@ void wxClientDCImpl::InitDC()
 {
     wxWindowDCImpl::InitDC();
 
-    // in wxUniv build we must manually do some DC adjustments usually
-    // performed by Windows for us
-#if defined(__WXUNIVERSAL__)
+    // Account for the origin of the client area which is non-zero only for
+    // TLWs with (left or top) toolbar: we shouldn't draw over the toolbar.
     wxPoint ptOrigin = m_window->GetClientAreaOrigin();
     if ( ptOrigin.x || ptOrigin.y )
     {
@@ -198,6 +197,9 @@ void wxClientDCImpl::InitDC()
         SetDeviceOrigin(ptOrigin.x, ptOrigin.y);
     }
 
+    // in wxUniv build we must manually do some DC adjustments usually
+    // performed by Windows for us
+#if defined(__WXUNIVERSAL__)
     // clip the DC to avoid overwriting the non client area
     wxSize size = m_window->GetClientSize();
     DoSetClippingRegion(0, 0, size.x, size.y);
```

but if we do this, then the fix above is not enough for `wxOverlay` any more and we need to also do something like

```cpp
    m_rect.Offset(-m_window->GetClientAreaOrigin());
```

in the overlay code (we can't do it before client-to-screen conversion).

But this actually doesn't seem too bad, and looks like the right thing to do, so I'll make a PR doing this, unless somebody quickly tells me why this is a bad idea.

vadz:
I've created the PR promised above now and pushed the commit making things work correctly for me under MSW, even in the presence of a toolbar.

vadz:
I'm merging this and the related PRs to fix the user-visible problems, if necessary we can always improve things (e.g. by switching to child overlay windows) later, but for now this seems good enough.

Issue #24885, opened 2024-10-16
wxWebSession resp. wxWebRequest crashes on localhost


rexkogitans:
This issue is based on the dicussion on the [forum](https://forums.wxwidgets.org/viewtopic.php?p=226124).

On my platform, recent Manjaro Linux, I use stock wxWidgets 3.2.6. Additionally, I downloaded the sample of webrequest, compiled it with debug symbols turned on and run it.

Citing exactly from the wxWidgets forum:

On my local machine in my home directory, there is a file called ~/public_html/Weinreben.jpg. It can be loaded in Firefox with these two URLs: http://127.0.0.1/~robert/Weinreben.jpg and http://localhost/~robert/Weinreben.jpg ([...] only HTTP, there is no listener on HTTPS).

Here is what the sample program does:

-  I can load any image from the Internet.
-  Internet host with non-existing port: It just takes forever, like [https://www.wxwidgets.org:12345/downloa ... blocks.png](https://www.wxwidgets.org:12345/downloads/logos/blocks.png)
-  http://127.0.0.1/~robert/Weinreben.jpg working
-  http://localhost/~robert/Weinreben.jpg crashes immediately
-  http://127.0.0.1:12345/~robert/Weinreben.jpg [crashes] sometimes immediately, sometimes after about 5 seconds
-  http://localhost/~robert/Weinreben.jpg crashes immediately

This is the core dump (if of any help):
```
poll([{fd=3, events=POLLIN|POLLOUT}], 1, -1) = 1 ([{fd=3, revents=POLLOUT}])
writev(3, [{iov_base="5\30\4\0\274\3`\2\v\0`\2\0\2\26\0\213\4\6\0\275\3`\2\274\3`\2)\0\0\0"..., iov_len=448}, {iov_base=NULL, iov_len=0}, {iov_base="", iov_len=0}], 3) = 448
recvmsg(3, {msg_namelen=0}, 0)          = -1 EAGAIN (Die Ressource ist zur Zeit nicht verfügbar)
pipe2([13, 14], 0)                      = 0
fcntl(13, F_SETFD, FD_CLOEXEC)          = 0
fcntl(14, F_SETFD, FD_CLOEXEC)          = 0
fcntl(13, F_GETFL)                      = 0 (flags O_RDONLY)
fcntl(13, F_SETFL, O_RDONLY|O_NONBLOCK) = 0
fcntl(14, F_GETFL)                      = 0x1 (flags O_WRONLY)
[...]

PR #24914, opened 2024-10-26
Fix crash in `wxWebRequestCURL` when connection is refused

vadz:
See #24885.


vadz:

PR #24915, opened 2024-10-27
Set HTML ids and search for them.

### LINKED ISSUE #17728: Support id tag in wxHtml
*Issue migrated from trac ticket # 17728*

**component:** wxHtml | **priority:** normal

#### 2016-11-27 11:01:53: tm created the issue

___
Currently only `<a name="anchor">..</a>` is accepted as an anchor. 
This is deprecated in HTML5. Here the `id` tag is used for create an anchor, e.g. `<h1 id="anchor">..</h1>`
Would it be possible to add support for the `id` tag in wxHtml?

oltolm:
Fixes #17728.

I noticed the problem when trying to use [xCHM](https://github.com/rzvncj/xCHM) - a CHM reader. It is almost unusable because it does not jump to a fragment like `#tab-completion-and-history-editing` in the URL `file:E:\\python.chm#xchm://tutorial/interactive.html#tab-completion-and-history-editing`. `#tab-completion-and-history-editing` is the ID of an HTML element and the browser can jump to such a fragment ID.

### DIFF
--- a/src/html/htmlcell.cpp
+++ b/src/html/htmlcell.cpp
@@ -193,7 +193,7 @@ void wxHtmlCell::Layout(int WXUNUSED(w))
 
 const wxHtmlCell* wxHtmlCell::Find(int condition, const void* param) const
 {
-    return condition == wxHTML_COND_ISANCHOR && !GetId().IsEmpty() && GetId() == (*((const wxString*)param)) ? this : nullptr;
+    return condition == wxHTML_COND_ISANCHOR && HasId() && GetId() == *static_cast<const wxString*>(param) ? this : nullptr;
 }
 
 
@@ -1287,7 +1287,7 @@ const wxHtmlCell* wxHtmlContainerCell::Find(int condition, const void* param) co
             if (r) return r;
         }
     }
-    return condition == wxHTML_COND_ISANCHOR && !GetId().IsEmpty() && GetId() == (*((const wxString*)param)) ? this : nullptr;
+    return condition == wxHTML_COND_ISANCHOR && HasId() && GetId() == *static_cast<const wxString*>(param) ? this : nullptr;
 }
 
 
--- a/src/html/htmlcell.cpp
+++ b/src/html/htmlcell.cpp
@@ -191,9 +191,9 @@ void wxHtmlCell::Layout(int WXUNUSED(w))
 
 
 
-const wxHtmlCell* wxHtmlCell::Find(int WXUNUSED(condition), const void* WXUNUSED(param)) const
+const wxHtmlCell* wxHtmlCell::Find(int condition, const void* param) const
 {
-    return nullptr;
+    return condition == wxHTML_COND_ISANCHOR && !GetId().IsEmpty() && GetId() == (*((const wxString*)param)) ? this : nullptr;
 }
 
 
@@ -1287,7 +1287,7 @@ const wxHtmlCell* wxHtmlContainerCell::Find(int condition, const void* param) co
             if (r) return r;
         }
     }
-    return nullptr;
+    return condition == wxHTML_COND_ISANCHOR && !GetId().IsEmpty() && GetId() == (*((const wxString*)param)) ? this : nullptr;
 }
 
 
--- a/src/html/m_dflist.cpp
+++ b/src/html/m_dflist.cpp
@@ -35,16 +35,17 @@ TAG_HANDLER_BEGIN(DEFLIST, "DL,DT,DD" )
 
         if (tag.GetName() == wxT("DL"))
         {
-            if (m_WParser->GetContainer()->GetFirstChild() != nullptr)
+            if (m_WParser->GetContainer()->GetFirstChild() != nullptr || !m_WParser->GetContainer()->GetId().IsEmpty())
             {
                 m_WParser->CloseContainer();
                 m_WParser->OpenContainer();
             }
+            m_WParser->GetContainer()->SetId(tag.GetParam(wxT("id")));
             m_WParser->GetContainer()->SetIndent(m_WParser->GetCharHeight(), wxHTML_INDENT_TOP);
 
             ParseInner(tag);
 
-            if (m_WParser->GetContainer()->GetFirstChild() != nullptr)
+            if (m_WParser->GetContainer()->GetFirstChild() != nullptr || !m_WParser->GetContainer()->GetId().IsEmpty())
             {
                 m_WParser->CloseContainer();
                 m_WParser->OpenContainer();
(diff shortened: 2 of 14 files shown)


src/html/htmlcell.cpp
```diff
@@ -191,9 +191,9 @@ void wxHtmlCell::Layout(int WXUNUSED(w))
 
 
 
-const wxHtmlCell* wxHtmlCell::Find(int WXUNUSED(condition), const void* WXUNUSED(param)) const
+const wxHtmlCell* wxHtmlCell::Find(int condition, const void* param) const
 {
-    return nullptr;
+    return condition == wxHTML_COND_ISANCHOR && !GetId().IsEmpty() && GetId() == (*((const wxString*)param)) ? this : nullptr;
```
vadz, line 196: A minor comment, but I think it would be nice to have a small inline helper function hiding this ugly cast (which ought to be a `static_cast<wxString*>`...) and use it here and in a couple of other places where the same check is done.

src/html/m_image.cpp
```diff
@@ -757,10 +757,11 @@ TAG_HANDLER_BEGIN(IMG, "IMG,MAP,AREA")
...
             m_WParser->CloseContainer();
             m_WParser->OpenContainer();
-            wxString tmp;
-            if (tag.GetParamAsString(wxT("NAME"), &tmp))
+            wxString name, id;
+            if (tag.GetParamAsString(wxT("NAME"), &name) || tag.GetParamAsString(wxT("ID"), &id))
             {
-                wxHtmlImageMapCell *cel = new wxHtmlImageMapCell( tmp );
+                wxHtmlImageMapCell *cel = new wxHtmlImageMapCell( name );
+                cel->SetId(tag.GetParam(wxT("id")));
```
vadz, line 764: Isn't this the same thing as
```suggestion
                cel->SetId(id));
```
? If not, it would merit a comment explaining why not because this doesn't seem obvious to me...

src/html/m_links.cpp
```diff
@@ -58,10 +58,12 @@ TAG_HANDLER_BEGIN(A, "A")
...
     TAG_HANDLER_PROC(tag)
     {
-        wxString name;
-        if (tag.GetParamAsString(wxT("NAME"), &name))
+        wxString name, id;
+        if (tag.GetParamAsString(wxT("NAME"), &name) || tag.GetParamAsString(wxT("ID"), &id))
         {
-            m_WParser->GetContainer()->InsertCell(new wxHtmlAnchorCell(name));
+            auto cell = new wxHtmlAnchorCell(name);
+            cell->SetId(tag.GetParam(wxT("id")));
```
vadz, line 65: Also
```suggestion
            cell->SetId(id));
```
?

src/html/m_dflist.cpp
```diff
@@ -35,16 +35,17 @@ TAG_HANDLER_BEGIN(DEFLIST, "DL,DT,DD" )
 
         if (tag.GetName() == wxT("DL"))
         {
-            if (m_WParser->GetContainer()->GetFirstChild() != nullptr)
+            if (m_WParser->GetContainer()->GetFirstChild() != nullptr || !m_WParser->GetContainer()->GetId().IsEmpty())
```
vadz, line 38: Similarly, it would probably make sense to add `HasId()` function and use it here as
```suggestion
            if (m_WParser->GetContainer()->GetFirstChild() != nullptr || m_WParser->GetContainer()->HasId())
```

vadz (commented):
Thank you! This is definitely a useful feature to have and we should merge it, but I'd like to simplify it a bit as indicated below.

I also wonder if we should pass `tag` to the different cell class ctors so that they could pass it down to `wxHtmlCell` which would then do `m_id = tag.GetParam("id")` in a single place instead of having to do it in a dozen or more of different places?

src/html/htmlcell.cpp
oltolm: I changed it to `static_cast` here and elsewhere.

src/html/m_image.cpp
oltolm: added a comment

src/html/m_links.cpp
oltolm: here too

src/html/m_dflist.cpp
oltolm: I added `HasId`.

oltolm:
> I also wonder if we should pass `tag` to the different cell class ctors so that they could pass it down to `wxHtmlCell` which would then do `m_id = tag.GetParam("id")` in a single place instead of having to do it in a dozen or more of different places?

I don't think it will work, because one HTML element can correspond to multiple `HtmlCell`s. Which `HtmlCell` should get the id? All of them? Currently I decide which `HtmlCell` will get the id.

vadz:
> > I also wonder if we should pass `tag` to the different cell class ctors so that they could pass it down to `wxHtmlCell` which would then do `m_id = tag.GetParam("id")` in a single place instead of having to do it in a dozen or more of different places?
> 
> I don't think it will work, because one HTML element can correspond to multiple `HtmlCell`s. Which `HtmlCell` should get the id? All of them? Currently I decide which `HtmlCell` will get the id.

Sorry if I'm missing the problem, but it seems to me that we could still do exactly the same thing as you're doing now by passing `tag` to the ctor of this cell only. We would have to make the new parameter optional or add a new overloaded ctor anyhow as we need to preserve the existing one for compatibility, and so you'd just pass it where needed?

oltolm:
> Sorry if I'm missing the problem, but it seems to me that we could still do exactly the same thing as you're doing now by passing `tag` to the ctor of this cell only. We would have to make the new parameter optional or add a new overloaded ctor anyhow as we need to preserve the existing one for compatibility, and so you'd just pass it where needed?

I get it now, but then you would not see where the id is set. And what about other attributes? Should they be set in the constructor too? I'd rather leave the code as is.

oltolm:
Is this going to be merged?

vadz:
> Is this going to be merged?

Yes, but I'd like to look at this once again and check if the change I proposed would really be a simplification. If not, then I'll merge it at it, but I just didn't have time to look at it yet, sorry.

src/html/m_span.cpp
```diff
@@ -38,6 +38,10 @@ TAG_HANDLER_BEGIN(SPAN, "SPAN" )
         int oldunderlined = m_WParser->GetFontUnderlined();
         wxString oldfontface = m_WParser->GetFontFace();
 
+        auto cell = new wxHtmlFontCell(m_WParser->CreateCurrentFont());
```
vadz, line 41: It looks like this is the only place where a new cell is now being added, instead of just setting ID for an existing cell.

Could you please explain why is this necessary, especially when there is no ID at all?

vadz:
Could you please check #24968? I think it does look a bit simpler like this, but please let me know if you disagree, I can also apply that one without the last commit where I implemented the idea about passing tag to the ctors.

In any case, I'm closing this one as it's included in the other PR now.

PR #24916, opened 2024-10-27
wxHtmlPrintout i18n and DPI fixes

Blake-Madden:
Make default titles translatable, don't piece strings together for title
Also, make initial frame size DPI aware
Remove `wxT()` macros

### DIFF
--- a/src/html/htmprint.cpp
+++ b/src/html/htmprint.cpp
@@ -709,7 +709,7 @@ bool wxHtmlEasyPrinting::DoPreview(wxHtmlPrintout *printout1, wxHtmlPrintout *pr
     }
 
     wxPreviewFrame *frame = new wxPreviewFrame(preview, m_ParentWindow,
-                                wxString::Format(_("%s Preview"), m_Name),
+                                wxString::Format(/* TRANSLATORS: %s may be a document title. */_("%s Preview"), m_Name),
                                 wxDefaultPosition,
                                 wxWindow::FromDIP(wxSize(650, 500), m_ParentWindow));
     frame->Centre(wxBOTH);
--- a/src/html/htmprint.cpp
+++ b/src/html/htmprint.cpp
@@ -710,9 +710,9 @@ bool wxHtmlEasyPrinting::DoPreview(wxHtmlPrintout *printout1, wxHtmlPrintout *pr
 
     wxPreviewFrame *frame = new wxPreviewFrame(preview, m_ParentWindow,
         wxGetTranslation(wxString::Format(_("%s Preview"), m_Name), wxString(),
-            "HTML easy printer preview title (%s may be the name of the document)"),
-                                               wxDefaultPosition,
-                                               wxWindow::FromDIP(wxSize(650, 500), m_ParentWindow));
+            wxString("HTML easy printer preview title (%s may be the name of the document)")),
+        wxDefaultPosition,
+        wxWindow::FromDIP(wxSize(650, 500), m_ParentWindow));
     frame->Centre(wxBOTH);
     frame->Initialize();
     frame->Show(true);
--- a/src/html/htmprint.cpp
+++ b/src/html/htmprint.cpp
@@ -709,10 +709,9 @@ bool wxHtmlEasyPrinting::DoPreview(wxHtmlPrintout *printout1, wxHtmlPrintout *pr
     }
 
     wxPreviewFrame *frame = new wxPreviewFrame(preview, m_ParentWindow,
-            wxGETTEXT_IN_CONTEXT(L"HTML easy printer preview title (%s may be the name of the document)",
-                                 wxString::Format(_("%s Preview"), m_Name)),
-        wxDefaultPosition,
-        wxWindow::FromDIP(wxSize(650, 500), m_ParentWindow));
+                                wxString::Format(_("%s Preview"), m_Name),
+                                wxDefaultPosition,
+                                wxWindow::FromDIP(wxSize(650, 500), m_ParentWindow));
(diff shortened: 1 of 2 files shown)


src/html/htmprint.cpp
```diff
@@ -559,22 +559,22 @@ wxString wxHtmlPrintout::TranslateHeader(const wxString& instr, int page)
...
     wxString num;
 
-    num.Printf(wxT("%i"), page);
-    r.Replace(wxT("@PAGENUM@"), num);
+    num.Printf("%i", page);
+    r.Replace("@PAGENUM@", num);
 
-    num.Printf(wxT("%lu"), (unsigned long)(m_PageBreaks.size() - 1));
-    r.Replace(wxT("@PAGESCNT@"), num);
+    num.Printf("%lu", (unsigned long)(m_PageBreaks.size() - 1));
```
vadz, line 565: As long as we're modifying this, we should make it
```suggestion
    num.Printf("%zu", m_PageBreaks.size() - 1);
```

src/html/htmprint.cpp
```diff
@@ -709,8 +709,11 @@ bool wxHtmlEasyPrinting::DoPreview(wxHtmlPrintout *printout1, wxHtmlPrintout *pr
     }
 
     wxPreviewFrame *frame = new wxPreviewFrame(preview, m_ParentWindow,
-                                               m_Name + _(" Preview"),
-                                               wxPoint(100, 100), wxSize(650, 500));
+                                               wxString::Format(_("%s Preview"), m_Name),
+                                               wxPoint(100, 100),
+                                               m_ParentWindow ?
+                                                   m_ParentWindow->FromDIP(wxSize(650, 500)) :
+                                                   wxSize(650, 500));
```
vadz, line 716: This can be written more simply as
```suggestion
                                               wxWindow::FromDIP(wxSize(650, 500), m_ParentWindow));
```

src/html/htmprint.cpp
```diff
@@ -709,8 +709,11 @@ bool wxHtmlEasyPrinting::DoPreview(wxHtmlPrintout *printout1, wxHtmlPrintout *pr
     }
 
     wxPreviewFrame *frame = new wxPreviewFrame(preview, m_ParentWindow,
-                                               m_Name + _(" Preview"),
-                                               wxPoint(100, 100), wxSize(650, 500));
+                                               wxString::Format(_("%s Preview"), m_Name),
+                                               wxPoint(100, 100),
```
vadz, line 713: Probably should use `wxDefaultPosition` here.

vadz (commented):
Thanks, I'll merge it with the changes proposed below if you don't mind.

src/html/htmprint.cpp
Blake-Madden: I thought that too when I saw it centering the window in the next line, but left it alone for fear of appearing too nit-picky :)

src/html/htmprint.cpp
```diff
@@ -709,8 +709,9 @@ bool wxHtmlEasyPrinting::DoPreview(wxHtmlPrintout *printout1, wxHtmlPrintout *pr
     }
 
     wxPreviewFrame *frame = new wxPreviewFrame(preview, m_ParentWindow,
-                                               m_Name + _(" Preview"),
-                                               wxPoint(100, 100), wxSize(650, 500));
+                                               wxString::Format(_("%s Preview"), m_Name),
```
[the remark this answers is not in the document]
Blake-Madden: Got it. I added `wxGETTEXT_IN_CONTEXT` to all the strings. Thanks.

lanurmi (commented):
I don't think this will work at all, for various reasons. Firstly, passing the formatted and localized `"%s Preview"` to `wxGetTranslation()` means a translation will virtually never be found for the argument. `wxGetTranslation()` (or `_()`) within `wxGetTranslation()` cannot be right. 

Also the messages will not be properly picked with context by `xgettext`, if calling `wxGetTranslation()` with multiple arguments. wxGETTEXT_IN_CONTEXT` is what should be used when dealing with string literals.

Blake-Madden:
> wxGETTEXT_IN_CONTEXT is what should be used when dealing with string literals.

Some compilers are failing with `wxGETTEXT_IN_CONTEXT`, complaining about ambiguity between `const char*` and `unsigned int`. I was calling `wxGETTEXT_IN_CONTEXT` the same way it is called throughout the rest of the wx code base, so I have no idea what compiles in other places but not here.

So, for `wxGETTEXT_IN_CONTEXT`, what is the proper way to call it? With an 'L' in front of the literal string? Or wrap it in a `wxString` CTOR? Again, I tried simply calling `wxGETTEXT_IN_CONTEXT` with string literals like I see in many other places, but 4 or so build environments fail.

Blake-Madden:
I cannot get `wxGETTEXT_IN_CONTEXT` to compile with string literals in all build environments. I have tried wxString CTOR wrappers, L wide strings, char*, etc. Rolling back to simple `_()` calls.

Blake-Madden:
AppVeyor "failed" configuration says "All tests passed (1162787 assertions in 385 test cases)".

This seems to be good to go.

lanurmi:
> This seems to be good to go.

It is still entirely possible to add the `/* TRANSLATORS: ... */` _comments_ that I suggested earlier.

MaartenBent:
> This seems to be good to go.

It runs two tests, `test.exe` and `test_gui.exe`. It seems the second one got stuck and AppVeyor timed out after 1 hour. I think this sometimes randomly happens.

Blake-Madden:
Translator comments are in there now :)

PR #24920, opened 2024-10-27
Fix HTML printing under dark mode

### LINKED ISSUE #24919: Printing an HTML control under dark mode results in light gray text on white background
wx3.3 on Windows (may occur on other platforms)

1. Open the htmlprinting sample
2. Add `MSWEnableDarkMode();` to `MyApp::OnInit()`
3. Set your system to dark mode
4. Run the same and open the "test.html" file from the sample folder
5. The window will appear dark with light grey text; looks fine
6. Go to print or print preview and you will see a white background (as you would want for printing), but the text is still light gray and difficult to see.

It has to do with `wxHtmlWinParser` when there is no windowing interface connected to it. `HTMLDCRenderer` uses `wxHtmlWinParser` during print without an windowing interface, and `wxHtmlWinParser` will fall back to a white background--which is fine. But it will *always* use the system text color, which will be light gray.

Blake-Madden:
If HTML parser has no window connected to it, fall back to black text on white background.

Closes #24919

### DIFF
--- a/src/html/winpars.cpp
+++ b/src/html/winpars.cpp
@@ -211,7 +211,7 @@ void wxHtmlWinParser::InitParser(const wxString& source)
         m_ActualColor = *wxBLACK;
         m_ActualBackgroundColor = *wxWHITE;
     }
-    
+
     m_ActualBackgroundMode = wxBRUSHSTYLE_TRANSPARENT;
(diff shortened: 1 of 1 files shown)


vadz:
Looks good to me, thanks! Will merge soon.

PR #24921, opened 2024-10-28
wxQt: translate QPainter of wxClientDC into window client area.

AliKet:
Because Qt considers the menubar and toolbar parts of the client area and we need to apply the same fix done for `wxPaintDC` in this commit 9652958 (translate `QPainter` into window client area.) to `wxClientDC` too.

Without this fix the overlay hint of a `wxAUI` window would be drawn at incorrect offset when this #24911 get merged:

Screenshot from 2024-10-27 19-54-01


vadz:

PR #24922, opened 2024-10-28
Fix compilation of <wx/archive.h> on Clang 19 on Fedora

lanurmi:
Clang 19.1.0 on Fedora Rawhide fails to compile `<wx/archive.h>` because m_rep's members are accessed as if m_rep was not a pointer, but it is.

This is obviously wrong with any compiler, but apparently because `wxArchiveIterator` is a template class, and its `operator=()` is never called in tests, its brokenness went unnoticed.

It has been broken ever since the commit where the code was added, 00375592f92f68c4ca3f44d8e839bcfd47adc4e1, 20 years ago.

### DIFF
--- a/include/wx/archive.h
+++ b/include/wx/archive.h
@@ -217,9 +217,9 @@ public:
 
     wxArchiveIterator& operator =(const wxArchiveIterator& it) {
         if (it.m_rep)
-            it.m_rep.AddRef();
+            it.m_rep->AddRef();
         if (m_rep)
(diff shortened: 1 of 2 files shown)


vadz:
Will merge soon, thanks!

Probably needs to be backported to 3.2 too, doesn't it?

lanurmi:
> Probably needs to be backported to 3.2 too, doesn't it?

Indeed.

PR #24923, opened 2024-10-28
Mention that wxHTML can handle background-color span style

Blake-Madden:
wxHTML can handle `background-color` inside of `span style` sections; mention that in the overview topic.


Blake-Madden:
You can refer to `wxHtmlWinTagHandler::ApplyStyle` to see where this is being processed.

PR #24924, opened 2024-10-28
Fix handling client area offset in wxMSW for different DC classes

vadz:
See the discussion in #24911, notably [this comment](https://github.com/wxWidgets/wxWidgets/pull/24911#issuecomment-2442790382).

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -108,6 +108,11 @@ Changes in behaviour not resulting in compilation errors
   platforms with RTF support. If you need to save/load plain text even for the
   files with the .rtf extension, specify wxTEXT_TYPE_PLAIN explicitly.
 
+- wxClientDC and wxPaintDC now correctly offset their origin by wxFrame
+  toolbar, if any. This is consistent with the behaviour of the other ports
+  but may require removing any compensation for not doing this before that may
+  be present in the application code.
+
 
 Changes in behaviour which may result in build errors
 -----------------------------------------------------
--- a/src/msw/dcclient.cpp
+++ b/src/msw/dcclient.cpp
@@ -188,9 +188,8 @@ void wxClientDCImpl::InitDC()
 {
     wxWindowDCImpl::InitDC();
 
-    // in wxUniv build we must manually do some DC adjustments usually
-    // performed by Windows for us
-#if defined(__WXUNIVERSAL__)
+    // Account for the origin of the client area which is non-zero only for
+    // TLWs with (left or top) toolbar: we shouldn't draw over the toolbar.
     wxPoint ptOrigin = m_window->GetClientAreaOrigin();
     if ( ptOrigin.x || ptOrigin.y )
     {
@@ -198,6 +197,9 @@ void wxClientDCImpl::InitDC()
         SetDeviceOrigin(ptOrigin.x, ptOrigin.y);
     }
 
+    // in wxUniv build we must manually do some DC adjustments usually
+    // performed by Windows for us
+#if defined(__WXUNIVERSAL__)
     // clip the DC to avoid overwriting the non client area
     wxSize size = m_window->GetClientSize();
     DoSetClippingRegion(0, 0, size.x, size.y);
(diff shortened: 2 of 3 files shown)


AliKet:
Thinking more about this I think we can do better by making the overlaywin a child of the associated window (I tried this and it works perfectly well)

overlay

I'll share a patch in a moment...

AliKet:
Patch to make the overlay window a child of the associatied window (For Windows 8.1 and above):

```C++
index 9d8240efa3..c25eea9feb 100644
--- a/src/msw/overlay.cpp
+++ b/src/msw/overlay.cpp
@@ -55,9 +55,10 @@ inline COLORREF GetTransparentColor()
     return RGB(0, 1, 2);
 }
 
-wxWindow* wxCreateOverlayWindow(const wxRect& rect, int alpha, HWND hWndInsertAfter)
+wxWindow* wxCreateOverlayWindow(wxWindow* parent, const wxRect& rect, int alpha)
 {
     auto overlayWin = new wxWindow();
+    overlayWin->SetParent(parent);
 
     overlayWin->MSWCreate
                 (
@@ -65,7 +66,7 @@ wxWindow* wxCreateOverlayWindow(const wxRect& rect, int alpha, HWND hWndInsertAf
                     nullptr, // No title
                     rect.GetPosition(),
                     rect.GetSize(),
-                    WS_POPUP,
+                    WS_CHILD | WS_CLIPSIBLINGS,
                     WS_EX_LAYERED |
                     WS_EX_TRANSPARENT |
                     WS_EX_NOACTIVATE
@@ -87,7 +88,7 @@ wxWindow* wxCreateOverlayWindow(const wxRect& rect, int alpha, HWND hWndInsertAf
 
     // We intentionally don't use WS_VISIBLE when creating this window to avoid
     // stealing activation from the parent, so show it using SWP_NOACTIVATE now.
-    ::SetWindowPos(GetHwndOf(overlayWin), hWndInsertAfter, 0, 0, 0, 0,
+    ::SetWindowPos(GetHwndOf(overlayWin), nullptr, 0, 0, 0, 0,
                    SWP_NOSIZE |
                    SWP_NOMOVE |
                    SWP_NOREDRAW |
@@ -161,8 +162,7 @@ void wxOverlayImpl::Init(wxDC* dc, int , int , int , int )
     m_window = dc->GetWindow();
 
-    m_rect = m_window->GetClientRect();
-    m_window->ClientToScreen(&m_rect.x, &m_rect.y);
+    m_rect = m_window->GetClientSize();
 
     if ( IsUsingConstantOpacity() )
     {
@@ -174,20 +174,20 @@ void wxOverlayImpl::Init(wxDC* dc, int , int , int , int )
         m_bitmap.UseAlpha();
     }
 
-    // We want the overlay window precede the associated window in Z-order, but
-    // under any floating window i.e. be positioned between them. And to achieve
-    // this, we must insert it after the _previous_ window in the same order, as
-    // then it will be just before this one.
-    // GW_HWNDPREV: Returns a handle to the window above the given window. (MSDN)
-    const auto win = wxGetTopLevelParent(m_window);
-    HWND hWndInsertAfter = ::GetNextWindow(GetHwndOf(win), GW_HWNDPREV);
-    m_overlayWindow = wxCreateOverlayWindow(m_rect, m_alpha, hWndInsertAfter);
+    m_overlayWindow = wxCreateOverlayWindow(m_window, m_rect, m_alpha);
 }
 
 void wxOverlayImpl::BeginDrawing(wxDC* dc)
 {
     wxCHECK_RET( IsOk(), wxS("overlay not initialized") );
 
+    if ( m_window->GetScrollHelper() ) // Is this a scrollable window ?
+    {
+        // Adjust position when rubberbanding. Maybe we should only do this
+        // when the window has actually scrolled.
+        m_overlayWindow->SetPosition(dc->GetDeviceOrigin());
+    }
+
     m_memDC.SelectObject(m_bitmap);
 
     auto impl = dc->GetImpl();

```
The patch is just a proof of concept that shows that we can actually create the overlay window as a child of the associated window, which IMHO is a good thing as it simplifies the management of the overlay window.

The patch is not production ready because I noticed a weired behaviour with it in auidemo sample:
- Apply the patch above
- Run the auidemo
- Undock the wxTreeCtrl pane
- Maximize the frame window
- Try to re-dock the wxTreeCtrl pane
- See the bug!!

I didn't try to debug this yet, so I don't know what's happening here?

PR #24925, opened 2024-10-28
Fix wxGETTEXT_IN_CONTEXT* failing with wxNO_IMPLICIT_WXSTRING_ENCODING

lanurmi:
The call to `wxGetTranslation()` was ambiguous. Fix this by using `wxASCII_STR()` for all string literal arguments in
`wxGETTEXT_IN_CONTEXT`/`_PLURAL`, just as it is used for the argument of `_()`.

Also test that all translation macros expand to compilable code when `wxNO_IMPLICIT_WXSTRING_ENCODING` is enabled.

See also: PR #24916


ardovm:
Looks good to me. Thank you @vadz for calling me in!

PR #24931, opened 2024-10-31
Fix phantom GTK mouse release events

vadz:
The only important commit here is the last one, the rest are just debugging helpers or some cleanups. But the last commit is important because it fixes a long-standing bug (I was _sure_ we had an issue for this but I couldn't find it, even after searching for it for quite some time) when the controls in the parent window were unexpectedly activated after dismissing a modal dialog. And I've finally understood what was going on here and described it in the last commit message and implemented a fix for it.

Note that I tried to make it work with `g_signal_connect_after()` for quite some time first, but couldn't do it: as also explained in that commit message, stopping `key-pressed` propagation by returning `TRUE` from its "after" handler breaks built-in keyboard handling, while doing what `wxgtk_tlw_key_press_event()` does in it results in infinite recursion. If anybody sees how to make this work, it would be IMO preferable to the current approach, which is not that obvious, unlike just stopping the signal propagation would be, but for now this is better than nothing.

### DIFF
--- a/src/gtk/evtloop.cpp
+++ b/src/gtk/evtloop.cpp
@@ -46,6 +46,12 @@ GdkWindow* wxGetTopLevelGDK();
 wxGUIEventLoop::wxGUIEventLoop()
 {
     m_exitcode = 0;
+    m_lastEvent = new GdkEvent;
+}
+
+wxGUIEventLoop::~wxGUIEventLoop()
+{
+    delete m_lastEvent;
 }
 
 int wxGUIEventLoop::DoRun()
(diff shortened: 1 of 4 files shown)


paulcor (approved):

src/gtk/evtloop.cpp
```diff
@@ -46,6 +46,12 @@ GdkWindow* wxGetTopLevelGDK();
 wxGUIEventLoop::wxGUIEventLoop()
 {
     m_exitcode = 0;
+    m_lastEvent = new GdkEvent;
```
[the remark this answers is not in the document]
vadz: Indeed, thanks, will do.

src/gtk/window.cpp
```diff
@@ -1910,13 +1960,8 @@ gtk_window_button_release_callback( GtkWidget *WXUNUSED(widget),
...
-    // want to prevent the other handlers from running even if we did process
-    // this event ourselves, there is no real advantage in doing this and it
-    // could actually be harmful, see #16055.
-    (void)win->GTKProcessEvent(event);
-
-    g_lastMouseEvent = nullptr;
+    if ( win->GTKProcessEvent(event) )
+        return TRUE;
 
     return FALSE;
```
paulcor, line 1966:
```suggestion
    return win->GTKProcessEvent(event);
```

vadz: I wanted to minimize changes to the existing code and, also, we are converting `bool` to `gboolean` explicitly in other places, so it would need to be

```suggestion
    return win->GTKProcessEvent(event) ? TRUE : FALSE;
```

which is not great and I'd rather to leave it as is, unless you feel really strongly about it.

Issue #24930, opened 2024-10-31
AUI toolbar squish when minimized and DockFixed


marekr:
We've gotten this bug in KiCad for awhile but at random where toolbars after some minimization of the app and unminization causes them to get drawn squished.

https://gitlab.com/kicad/code/kicad/-/issues/14581

In this one issue, it was due to our self closing infobar messages, if you minimized the window while the infobar was open, and then reopened it 10 seconds later (after the infobar would be removed), the toolbars are squished.

But there's definitely other ways to trigger it as I can get it in one window that doesn't even use a infobar.

I found a repro case using the disappearing infobar class we have (called WX_INFOBAR),

Repro auidemo.cpp here: https://gist.github.com/marekr/029504961b8df372a2b413382e5c5718

The repro works by
1. Click the Customize option for the vertical toolbar, I just picked the first place with a click event for this. Nothing special
2. The infobar message appears
![image](https://github.com/user-attachments/assets/a40fe300-39b1-4caa-a7ad-7f0475d42df6)
3. Minimize the application to task bar

![image](https://github.com/user-attachments/assets/356c062b-df79-4dc0-92be-2295dae739af)
4. Wait ~10 seconds (the timer is 6 seconds)
5. Unminimize the app from taskbar
6. Observe squished toolbar
![image](https://github.com/user-attachments/assets/36f0832d-afb2-444a-bc5d-f70b485845da)

Now, the identifiable cause is the DockFixed( true ) on the wxAuiPaneInfo cofigured for the vertical toolbar

```
    m_mgr.AddPane(tb5, wxAuiPaneInfo().
                  Name("tb5").Caption("Sample Vertical Toolbar").
                  ToolbarPane().Left().Resizable( true ).Movable( false ).*DockFixed( true)*
                .LeftDockable().RightDockable().CaptionVisible( false ));
```

Remove DockFixed( true ) and the behavior does not reproduce.

marekr:
Seems like the issue is related to `wxAuiManager::LayoutAll` firing while being minimized with the fixed mode dock panes.

In particular

```
        if (dock.fixed)
        {
            // always reset fixed docks' sizes, because
            // the contained windows may have been resized
            dock.size = 0;
        }
```

zeros out the dock size before trying to calculate it again. Removing the dock.size = 0; statement also fixes it 


Looking further, it may be related to the cli_size variable which is from GetClientsize(), apparently when its minimized the value returned is junk

[...]

PR #24934, opened 2024-11-01
Fix updating AUI while minimized

vadz:
This is mostly the fix for #24930 but also fix a separate bug in `wxFrame::GetClientSize()` in wxMSW.


marekr:
Fixes the repro case for me.

PR #24937, opened 2024-11-03
Add MSVC projects for all samples

vadz:
After literally months of (on and off) works, I've finally finished creating projects for all wxWidgets samples (the solution file contains 128 of them, although this counts the main libraries projects too).

The organization is still subject to change, currently it looks like this:

image

which is probably not ideal and suggestions for tweaking it are appreciated, but I'd like to keep the following properties of the current organization:

1. The most important samples are in their own folder which comes first. As MSVS orders items alphabetically, this imposes limitation on the name of the first folder.
2. The rest of the folders are roughly in the order of importance.


vadz:
I'm a bit surprised that there are no objections or suggestions for changes here, but this probably just means that nobody has really looked at it. If you plan to do it, please let me know — otherwise I'll merge it soon and then, hopefully, people will test it in master.

PBfordev:
FWIW, I can take a look but it probably won't be in the next few days.

Assuming the new sample projects are similar to the current manually created minimal sample msbuild-based MSVS solution... I personally don't like that when having this sample solution open, I can accidentally trigger rebuilding, or even worse – cleaning the whole library, and not just the sample as intended. 

For some reason, the actual library is somehow "included" in the sample solution: IIRC, it was not like that with bakefile-generated MSVS solutions and TBH, I don't see why it should be like this.

vadz:
This is intentional as it allows to have proper dependencies and is a pretty important part of these changes. Previously, I got into a situation when some fix didn't work just because I had forgotten to rebuild the library (not necessarily core one) after making it and so the sample simply didn't use it.

doublemax:
> I'm a bit surprised that there are no objections or suggestions for changes here, but this probably just means that nobody has really looked at it. If you plan to do it, please let me know — otherwise I'll merge it soon and then, hopefully, people will test it in master.

First of all, thanks for doing this, i really missed the VS solutions for the samples.

I tested a few samples with VS2017 and didn't see any issues.

Just two mentions:
Where the configurations in the solution files for the libs say "Win32", the solutions for the samples say "x86".
And the ARM support is missing in the samples.

vadz:
> Where the configurations in the solution files for the libs say "Win32", the solutions for the samples say "x86".

I have no idea which one is preferred. "x86", I guess? Or just leave it as is because nobody cares about 32-bit code any more?

> And the ARM support is missing in the samples.

Yes, as I mentioned in https://github.com/wxWidgets/wxWidgets/pull/24937/commits/606abaec970d3fc29cf7f76c37bbe4080b6bccba commit message, I had started doing this before ARM support was added and I didn't have the energy/motivation to modify all the projects to add it afterwards. This should be doable with some scripting, maybe later...

doublemax:
> > Where the configurations in the solution files for the libs say "Win32", the solutions for the samples say "x86".
> 
> I have no idea which one is preferred. "x86", I guess? Or just leave it as is because nobody cares about 32-bit code any more?
> 
I think it's ok to leave it as it is, but at least i care about 32bit code. There are still 32bit systems out there, and most applications do not benefit from 64bit enough to justify maintaining two different executables. I'm talking only about Windows though.

PBfordev:
I have used latest versions of MSVS 2017, 2019, and 2022 to create a Windows application project using their built-in template:  They all have "x86". 

I don't have any older MSVS version available.

PBfordev:
I took a look and haven't found anything important, just have few observations:
1. When using a DLL configuration, the samples cannot be run even from MSVS because wxWidgets DLLs are not found. I suppose it is not possible to somehow add a `<sample>.vcxproj.user` file, where it would for DLL configurations modify `LocalDebuggerEnvironment`, setting it to `PATH=.\..\..\lib\$(wxOutDirName)`. Just for convenience.
1. Some samples can't find their data files when running their executable from the shell. I don't remember if it was like this with bakefile-generated projects (i.e., if data files were copied somewhere the executable could find them) or if it can even be easily fixed.
1. MFC sample could be in the Windows-specific folder.
1. XTI sample is missing, not even in Unmaintained Old Samples but I guess it is on purpose.
1. Speaking of Unmaintained Old Samples, when I run the dragimag sample and try dragging the mostly-white image, I get "Couldn't add an image to the image list" error. The error is not there when running the sample from CMake-generated MSVS solution.
1. htlbox sample is from wxHTML library yet it is in Control folders (where all other samples are from wxCore) and not in Non-Core Libraries Samples / HTML. But I guess it makes more sense to have it where it is. TBH, I can't really say why some samples are in Topical and some in Uncategorized (e.g., Power and Sound in Topical and ArtProv, Display, or PopUp in Uncategorized).
1. To me, folder name `Base Samples` indicates the Base library (see Non-Core Libraries Samples folder), I would prefer another name and this is the last nitpick of this post.

vadz:
> I took a look

Thanks!

>     1. When using a DLL configuration, the samples cannot be run even from MSVS because wxWidgets DLLs are not found. I suppose it is not possible to somehow add a `<sample>.vcxproj.user` file, where it would for DLL configurations modify `LocalDebuggerEnvironment`, setting it to `PATH=.\..\..\lib\$(wxOutDirName)`. Just for convenience.

I didn't know how to do it. AFAIK CMake avoids the problem by putting all binaries in the same directory (doesn't it?), but if we could somehow specify the PATH to the DLLs it would be nicer. But then I'd also rather not add `vcxproj.user` files to Git, they're not supposed to be tracked by the VCS. If you know of some other way to specify the path, please let me know (or make a PR if possible)!

>     2. Some samples can't find their data files when running their executable from the shell. I don't remember if it was like this with bakefile-generated projects (i.e., if data files were copied somewhere the executable could find them) or if it can even be easily fixed.

Yes, bakefile did copy files. Again, the best solution would be to run the samples from their source directory and not build one, but I don't know how to do it neither.

>     3. MFC sample could be in the Windows-specific folder.

Yes, I guess so... As you noticed, a lot of samples don't clearly belong to one or the other folder.
 
>     4. XTI sample is missing, not even in Unmaintained Old Samples but I guess it is on purpose.

Yes, it doesn't build out of the box. We probably could add it but exclude from building but it just didn't seem to be worth it (i.e. I was too lazy to do it).

>     5. Speaking of Unmaintained Old Samples, when I run the dragimag sample and try dragging the mostly-white image, I get "Couldn't add an image to the image list" error. The error is not there when running the sample from CMake-generated MSVS solution.

I guess it's another missing/not found file problem?
 
>     6. htlbox sample is from wxHTML library yet it is in Control folder (where all other samples are from wxCore) and not in Non-Core Libraries Samples / HTML. But I guess it makes more sense to have it where it is. TBH, I can't really say why some samples are in Topical and some in Uncategorized (e.g., Power and Sound in Topical and ArtProv, Display, or PopUp in Uncategorized).

I tried to put "more important" samples under "Topical" and the less important ones elsewhere in order not to have too many samples in the same folder but I admit that this is rather arbitrary. Maybe we should just merge them. I really don't know what would be best.

>     7. To me, folder name `Base Samples` indicates the wxBase library (see Non-Core Libraries Samples folder), I would prefer another name and this is the last nitpick of this post.

This is probably my main problem as well. But I really want this folder to come first and I couldn't come up with any other name making at least some sense and being before "Controls" alphabetically. Any ideas are welcome!

PBfordev:
> > 1. When using a DLL configuration...
>
> I didn't know how to do it. AFAIK CMake avoids the problem by putting all binaries in the same directory (doesn't it?). 

Yes, CMake does put all the files in the same binary dir (which also creates problems for some samples). Unfortunately, the way I suggested is the only way I know telling MSVS where to find extra DLLs.

> > 2. Some samples can't find their data files when running their executable from the shell ...
>
> Yes, bakefile did copy files. Again, the best solution would be to run the samples from their source directory and not build one, but I don't know how to do it neither.

Sorry if I was not clear, the samples work when run from MSVS, but not from the shell from the folder with their executable.

> > 5. when I run the dragimag sample and try dragging the mostly-white image, I get "Couldn't add an image to the image list" error.
> 
> I guess it's another missing/not found file problem?

I stopped being lazy and looked into it and this may be an actual problem I missed. Some samples have their own .rc file with their extra resources (and including `sample.rc`) and those .rc files are not included in the MSVS project for all the samples where it should. But I really need to go sleep now, so I cannot check where they are missing today.

vadz:
I did try to use the correct `.rc` files in the samples which had them, but I could well have missed some. If you notice any, please let me know or make a PR fixing this against this branch. This is not urgent at all, of course, I've been trying to finish this for several months, so a few more days won't change anything. TIA!

PBfordev:
The sample projects which should include their own .rc file but do not are (I attach [a patch](https://github.com/user-attachments/files/17719042/wx-samples-sln.patch) modifying the affected projects): `dnd docview dragimag mdi nativdlg sound svgtest`

I think it could be a good idea to mention how were these projects and the solution created in the Contributing Notes in (a) `how-to-add-new-sample.md` and (b) `how-to-add-new-msvs-version.md`.

I was also thinking some more about the solution folder structure. TBH, I have a bit of a problem finding some samples in the current structure. IMO, you are looking at this with “developer-colored glasses” (see also IMO incorrect “live-linking” libraries from sample solution) but for the vast majority of users, dividing the samples by libraries (core, others) may make little sense. I would use as sample root folders just Fundamental Samples (minimal, dialogs, widgets), GUI Samples (everything the user can see and interact with), and Other Samples (e.g., archive, clipboard, and webrequest). Yes, GUI Samples would have a large list and yes, classification of some samples may not be clear (e.g., event, image, drawing, or xrc) but I would still prefer this. Additionally, in an ideal case, CMake build system would be modified so it would generate the MSVS solution with the same structure as the manually created one.

By the way, is there a reason why each sample project has a list of libraries to link with, instead of just autolinking them using the MSVC-specific setup.h? If the sole reason is to show which libraries the sample uses, I don’t think that is a good enough reason: IMO, the samples are a poor starting block on which to build a user project (relative paths, using wxWidgets compiler variables...) and moreover, users themselves may be better off using the MSVC-specific include.

_Additional observation_
When building clean the whole solution, I get errors, fixed by building again:
```
Error	C1083	Cannot open include file: 'wx/setup.h': No such file or directory	my_dll	E:\dev\Desktop\Lib\wxWidgets-PB\include\wx\platform.h	159		
Error	LNK1104	cannot open file 'vc_x64_mswud\\my_dll.lib'	sdk_exe	E:\dev\Desktop\Lib\wxWidgets-PB\samples\dll\LINK	1		
Error	LNK1104	cannot open file 'vc_x64_mswud\\my_dll.lib'	wx_exe	E:\dev\Desktop\Lib\wxWidgets-PB\samples\dll\LINK	1		
```

This is probably caused by missing dependencies on `base` and `core` in `my_dll` project (by design?).

vadz:
> The sample projects which should include their own .rc file but do not are (I attach [a patch](https://github.com/user-attachments/files/17719042/wx-samples-sln.patch) modifying the affected projects): `dnd docview dragimag mdi nativdlg sound svgtest`

Thanks! I definitely missed more of them than I thought...

> I think it could be a good idea to mention how were these projects and the solution created in the Contributing Notes in (a) `how-to-add-new-sample.md` and (b) `how-to-add-new-msvs-version.md`.

Yes, absolutely, I'll do this.

> I was also thinking some more about the solution folder structure. TBH, I have a bit of a problem finding some samples in the current structure. IMO, you are looking at this with “developer-colored glasses” (see also IMO incorrect “live-linking” libraries from sample solution) but for the vast majority of users, dividing the samples by libraries (core, others) may make little sense. I would use as sample root folders just Fundamental Samples (minimal, dialogs, widgets), GUI Samples (everything the user can see and interact with), and Other Samples (e.g., archive, clipboard, and webrequest). Yes, GUI Samples would have a large list and yes, classification of some samples may not be clear (e.g., event, image, drawing, or xrc) but I would still prefer this.

Hmm, I think we'd still want to somehow prioritize some samples and de-emphasize the others. But we could have "Fundamental", "GUI Building Blocks" (instead of "Controls") and "Other Topics", maybe?

> By the way, is there a reason why each sample project has a list of libraries to link with, instead of just autolinking them using the MSVC-specific setup.h?

Just because this is what the minimal project did, and all the others were created by cloning it. But changing all of them now is not going to be simple (unless this can be done by multiselecting them in the IDE?).

> _Additional observation_ When building clean the whole solution, I get errors, fixed by building again:
...
> This is probably caused by missing dependencies on `base` and `core` in `my_dll` project (by design?).

I don't remember if this was really by design, but probably not. Unless adding these dependencies results in some problem?

PBfordev:
> > By the way, is there a reason why each sample project has a list of libraries to link with, instead of just autolinking them using the MSVC-specific setup.h?
> 
> Just because this is what the minimal project did, and all the others were created by cloning it. But changing all of them now is not going to be simple (unless this can be done by multiselecting them in the IDE?).

I was thinking of a relatively simple batch find&replace in the .vcxproj files and then manually tweaking the few that require extra libraries (probably just OpenGL). I.e., I am offering to attempt this if you agree.

> > _Additional observation_ When building clean the whole solution, I get errors, fixed by building again:
> > ...
> > This is probably caused by missing dependencies on `base` and `core` in `my_dll` project (by design?).
> 
> I don't remember if this was really by design, but probably not. Unless adding these dependencies results in some problem?

I am not familiar with the sample (or concept presented there). Its readme has "So we do something else here: the DLL is compiled against *static* build of wx." so I guess there may be something special about this. Regardless, adding the dependencies makes the build errors disappear and both exes work (although SDK does not have any manifest).

Regarding the folder structure. I would prefer it organized by purpose and I came with this. I cheated to work around the order issue (the second level folders are also numbered to ensure Misc always comes last) and of course, some samples (e.g., aui) would naturally belong into more folders and some are not well places (e.g., wizard and shaped in Controls) but they have to be pidgeonholed into just one.
```
1 Fundamental Samples
  minimal
  dialogs
  widgets

2 GUI Controls Samples
  1 Menus, Toolbars & Co
    menu
    ribbon
    statbar
    toolbar
  
  2 Grids, Lists & Trees
    dataview
    htlbox
    grid
    listctrl
    propgrid
    treectrl
    treelist

  3 Text Controls
    combo
    richtext
    stc
    text

  4 Containers
    aui
    collpane
    mdi
    notebook
    
  5 Layout
    layout
    splitter
    wrapsizer
  
  6 HTML
    about
    help
    helpview
    printing
    test
    virtual
    widget
    zip

  7 Miscellaneous
    animate
    artprov
    calendar
    docview
    font
    mediaplayer
    popup
    preferences
    printing
    shaped
    webview
    wizard
  
3 GUI-related Samples
  1 Input
    caret
    joytest
    keyboard

  2 Drawing
    drawing
    erase
    image
    ownerdrw
    render
    scroll
    svg
    vscroll
      
  3 Miscellaneous
    access
    clipboard
    debugrpt
    dnd
    display
    event
    internat
    taborder
    taskbar
    taskbarbutton
    uiaction
    validate
    xrc

4 OpenGL Samples
  cube
  isosurf
  penguin
  pyramid

5 Non-GUI Samples
  archive
  config
  console
  except
  exec
  fswatcher
  power
  secretstore
  sound

6 Communication & Network Samples
  1 IPC
    baseipcclient
    baseipcserver
    ipcclient
    ipcserver

  2 Sockets
    baseclient
    baseserver
    client
    server
      
  3 Misc
    webrequest

6 Windows-only Samples
  dll
  mfc
  nativdlg
  oleauto
  regtest
    
7 Unmaintaned Old Samples
  dialup
  dragimag
  sashtest
  splash
  thread
  typetest
    
8 wxWidgets Libraries
.....
```

infinity77:
Maybe you can take inspiration from the wxPython demo structure:




😊

PBfordev:
@infinity77
> Maybe you can take inspiration from the wxPython demo structure:

Thanks fo the tip. However, I am not familiar with wxPython, but from the screenshot it seems that it has rather different samples. For example, in C++ samples, all the dialogs are in a single sample and we have not had any recent additions/updates. So I think we cannot model C++ sample structure after the Python one.

PBfordev:
I might have figured out the "wxWidgets DLLs not found when running a sample" problem.

I created file `WXWIN/samples/samples.props`:
```xml
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <PropertyGroup Condition="'$(Configuration)'=='DLL Debug' or '$(Configuration)'=='DLL Release'">
    <LocalDebuggerEnvironment>PATH=.\..\..\lib\$(wxOutDirName);$(PATH)</LocalDebuggerEnvironment>
  </PropertyGroup>
</Project>
```
and modified `WXWIN/samples/minimal/minimal.vcxproj`:
```diff
diff --git a/samples/minimal/minimal.vcxproj b/samples/minimal/minimal.vcxproj
index ad54c45467..7a71db04b6 100644
--- a/samples/minimal/minimal.vcxproj
+++ b/samples/minimal/minimal.vcxproj
@@ -69,6 +69,7 @@
   <ImportGroup Label="PropertySheets">
     <Import Project="..\..\build\msw\wx_setup.props" />
     <Import Project="..\..\build\msw\wx_local.props" Condition="Exists('..\..\build\msw\wx_local.props')" />
+    <Import Project="..\samples.props" />
   </ImportGroup>
   <ImportGroup Condition="'$(Configuration)|$(Platform)'=='DLL Release|Win32'" Label="PropertySheets">
     <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
```

vadz:
> I might have solved the "wxWidgets DLLs not found when running a sample" problem, without using vcxproj.user files.

Great news, thanks! I've cherry picked this commit and the fixes to the `.rc` files one and also added the promised instructions.

Would you like to reorganize the samples tree in the way you proposed? It shouldn't be that difficult but I just know that I'll agonize for hours myself trying to decide which folder is more appropriate for some completely insignificant sample, so I'd be more than grateful to delegate this to somebody else :-) Please let me know if you plan to do it (not necessarily right now) and I'll wait for it and then cherry pick your changes here too and finally merge this.

Thanks again!

PBfordev:
> Would you like to reorganize the samples tree in the way you proposed?

Yes, assuming we are talking about a numbered tree I introduced few days ago. I know some samples will be in a "wrong" category (e.g., aui, shaped, wizard), but that's something that can't reallisticalaly be perfect. 

I have also worked out a solution on how to copy the data files to the folder with the sample executable, just as GCC and VC makefiles do. Please tell me what you think. Perhaps it is not useful, the old MSVS projects did not do it. OTOH, I myself find possibility to run sample executables directly quite convenient.

My solution would be to add to each folder with a sample with data files `copyfile.props` file that would look like this (for the image sample):
```xml
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <ItemDefinitionGroup>
    <PostBuildEvent>
      <Command>xcopy /Q /D /I /Y $(ProjectDir)*.ani $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.bmp $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.cur $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.gif $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.ico $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.jpg $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.pcx $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.png $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.pnm $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.svg $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.tga $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.tif $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.xpm $(OutDir)
xcopy /Q /D /I /Y $(ProjectDir)*.xbm $(OutDir)</Command>
    </PostBuildEvent>
    <PostBuildEvent>
      <Message>Copying sample data files</Message>
    </PostBuildEvent>
  </ItemDefinitionGroup>
</Project>
```
and to `sample.props`
```xml
<Import Project="$(MSBuildProjectDirectory)\copyfiles.props" Condition="Exists('$(MSBuildProjectDirectory)\copyfiles.props')" 
```
It has at least one shortcoming: Clean won't clean these files but I can't think of a better simple solution.

SpareSimian:
I recommend using a separate script for PostBuildEvent. It's easier to maintain and test than a mass of commands in the XML file, and you don't have to worry about escaping the redirection operators within a block of XML. Something like "copyfile.cmd $(ProjectDir) $(OutDir)".

PBfordev:
> I recommend using a separate script for PostBuildEvent.

Yes, thanks for the suggestion. If this gets greenlighted, I will try to do that.

vadz:
Just to be clear, I'm perfectly fine with all the mentioned enhancements. Please make PRs against this branch or tell me where to cherry-pick/pull your changes from. Also, after the samples structure is in place, I think we can merge this and then make other PRs for the rest of improvements, as usual.

TIA!

PBfordev:
I have restructured the solution, using categories as agreed and I have also added dependency on `core` and `base` to `my_dll` sample: https://github.com/PBfordev/wxWidgets/commits/samples-vcxproj/

I loaded a solution for each MSVS version in that version. The only issues I noticed were in MSVS 2015 but they are minor and cannot be helped (missing Arm configuration for library, unrecognized manifest element "dpiAwareness"). Building in MSVS 2015 results in a bunch of odd warnings for some global variables (e.g., _warning C4592: 'wxNullGraphicsPen': symbol will be dynamically initialized (implementation limitation)_) and some warnings in Lexilla (C++ version related?) but I guess it's fine.

I hope to add copying sample data files to the folders with their executable by the end of the week.

My last intended change, using MSVC-specific setup.h to autolink the libraries instead of listing them separately, could be added in a separate PR, after this one is merged. It might not be just a simple find&replace and few tweaks (but hopefully will)...

vadz:
Thanks, I've cherry-picked the 2 new commits here too now.

I love the idea of using numbers for the sections, this is so simple, yet solves all the problems with artificially stupid names, thanks!

I still do have some quibbles, of course, how would we live without them, none of them is critical but I'd like to know what do you (and/or others) think:

1. I'd start numbering with "0" instead of "1" just because I'm a nerd.
1. Could it be a bit better to use "2.1 Menus, toolbars" etc, i.e. prepend the number with the outer section one in the nested sections?
1. Do OpenGL samples really deserve their own top-level section, instead of being just a "GUI Samples" subsection?
1. I'd rename "GUI-related samples" to "Other GUI Samples".
1. A number of samples in "GUI Controls Samples/Miscellaneous" are not obviously related to the GUI controls, e.g. anitest, artprov, font, ... I'd move them to "Other GUI Samples" too.
1. OTOH "docview" seems more important and could probably merit its own "Document-based" subsection, where we could also put "mdi" and, forcing the trait a little bit, "printing", perhaps?
1. Maybe combine "Windows-specific" and "Old" into a single section? I struggly to think of a good name for it, however...
1. Speaking of names, "Non-GUI Samples" is a bit confusing because a number of samples here do use GUI. Can we find some better name indicating that they demonstrate non-GUI classes? "Non-GUI Functionality Samples" maybe?

But all these remarks are relatively minor and we could also already merge it like this. But if you find any merit in (some of) them, it would probably be slightly better to change this now rather than later.

Thanks again!

PBfordev:
> 1. I'd start numbering with "0" instead of "1" just because I'm a nerd.

TBH, I am not a fan of this. IMO, the folder samples are more like the chapter headings in a reference book (and even C++ ISO standard chapter numbers start at 1), 0-based numbering just seems odd to me here. Please tell me if you insist and I will change it.
 
> 2. Could it be a bit better to use "2.1 Menus, toolbars" etc, i.e. prepend the number with the outer section one in the nested sections?

Yes, it would be better, see the book chapters analogy above. I originally decided to not use it so that the numbers do not overshadow the names too much but I will change it.

> 3. Do OpenGL samples really deserve their own top-level section, instead of being just a "GUI Samples" subsection?

I do not view OpenGL just as another control but I suppose you are right and I will move it. However, we do not have "GUI Samples" section, we have "GUI Controls Samples" and (renamed) "Other GUI Samples". Please tell me which to use, I slightly favor the latter.

> 4. I'd rename "GUI-related samples" to "Other GUI Samples".

Will do.

> 5. A number of samples in "GUI Controls Samples/Miscellaneous" are not obviously related to the GUI controls, e.g. anitest, artprov, font, ... I'd move them to "Other GUI Samples" too.

AFAICT, anitest shows off `wxAnimationCtrl`, which is a `wxControl` so I would leave it where it is, unless you insist. I will move the other two. And I propose moving also preferences, shaped, and wizard. Please tell me which other samples you want to move (or just move them by yourself before merging).
 
> 6. OTOH "docview" seems more important and could probably merit its own "Document-based" subsection, where we could also put "mdi" and, forcing the trait a little bit, "printing", perhaps?

Will do.

> 7. Maybe combine "Windows-specific" and "Old" into a single section? 

Sorry, I just don't see any logical reason to merge the two. Is having one top-level folder less worth doing that?

>8. Speaking of names, "Non-GUI Samples" is a bit confusing because a number of samples here do use GUI. Can we find some better name indicating that they demonstrate non-GUI classes? "Non-GUI Functionality Samples" maybe?

My reasoning: The samples may use GUI, but they demonstrate the classes that do not use/need GUI. I wanted to avoid word "Classes" since it would then ought to be used in all top-level folder names for samples.

PBfordev:
> > AFAICT, anitest shows off `wxAnimationCtrl`, which is a `wxControl` so I would leave it where it is, unless you insist.

> Oh, no, I didn't think about this. I somehow still think of it more as belonging to "Drawing" category, but I'm not sure at all where would people expect to find it.

My reasoning: If it is a `wxControl`, it belongs to GUI Controls Samples folder.


> 5. ... FWIW "artprov" for me belongs to "Drawing" (maybe rename it to "Graphics"?) category too because it's a bit like "image".

My reasoning: `wxArtProvider`'s reason for existence is to feed standard images to `wxControl`s, i.e., not related to custom drawing.
If we rename Drawing to Graphics, then OpenGL should be there as well, should it not?

> 8. Speaking of names, "Non-GUI Samples" ...
> I'd probably prefer to avoid having the word "GUI" here at all. What about "Cross-platform Abstractions"? "Platform Services"

The way I see it, most wxWidgets GUI controls are cross-platform abstractions over native controls. TBH, I am not really feeling either of the two suggestions, while also agreeing that Non-GUI Samples may not be perfect. I just cannot say that samples like archive, console, or sound in my mind fit into "Platform Services".

To conclude: As you yourself stated earlier, we have to accept that categorization cannot be perfect (while striving to come as close to perfection as reasonably possible) and the PR should not be hold hostage to analysis paralysis. The structure is easy to change, it needs to be done in just one solution, so it could be easily changed whenever later (but most likely will stay as was committed for ever).

PBfordev:
I have tried to improve sample categorization as best as I could, based on our previous conversations. In the end, I used just _Other Samples_ for the "non-GUI" samples folder. Please, feel free to change anything you want, I will not be offended. Unfortunately, I messed the numbering up so the changes are in two, non-consecutive,  commits. I did not try to merge the two commits, being afraid of messing-up my branch.

I have also removed one forgotten dependence on the advanced library from widgets sample (in Debug|Win32 configuration).

Finally, I also implemented copying sample data files to the folders with their executable as a MSBuild post-build event. The command files for copying were generated by [a simple ugly C++ program](https://github.com/PBfordev/wxsamplesvcxprojhelper), so if you want to change something in them (e.g., remove the explanation how to use them), it should not be difficult for me.

All these changes are at https://github.com/PBfordev/wxWidgets/commits/samples-vcxproj/

I think I am done for now. The only other change I would like to do (assuming it would be easy) is use the MSVC-specific setup.h  to autolink the libraries instead of listing them manually. But I'd rather do that after this PR is merged, unless you don't mind this PR marinating for few more weeks (but I'd rather avoid that). I have not tried the new samples projects/solution in 3.2, I assume you will want to backport them, right?

PBfordev:
> after looking at this, I wonder if we really want to add all these command files when there is a potentially simpler way to make this work instead: what if we modified the samples to use `wxStandardPaths`? ....

IMO, it would:
1. Complicate the sample code.
2. The data files still need to be in the folder with the executable in CMake builds (at least on Windows and Linux, the data files are copied to the binary dir with the executable). Few samples are "smart" and use `wxPathList` to look in several folders.
3. I personally don't have a good experience with ignoring MSVS app sub-folders, I think newer MSVS uses a different naming scheme by default (x86 vs x64?) but I may be remembering this wrong.
4. Lastly, AFAIK, the good practice for accessing program's data files is not using an executable-related directory but, as you said, `GetDataDir().`..

Either way, sorry, I am not doing that.

PR #24941, opened 2024-11-05
Fix indent of imageless `wxTreeCtrl` in high DPI under MSW

### LINKED ISSUE #24640: Buttons of an imageless `wxTreeCtrl` are too small at high DPI
At high DPI, the size of the buttons in a `wxTreeCtrl` is too small when the tree is used without images.

This issue can be reproduced with the following changes to the aui sample:

```patch
diff --git "a/samples/aui/auidemo.cpp" "b/samples/aui/auidemo.cpp"
index 56f5384bb6..c8b7dacc45 100644
--- "a/samples/aui/auidemo.cpp"
+++ "b/samples/aui/auidemo.cpp"
@@ -1631,7 +1631,7 @@ wxTreeCtrl* MyFrame::CreateTreeCtrl()
     wxVector<wxBitmapBundle> images;
     images.push_back(wxArtProvider::GetBitmapBundle(wxART_FOLDER, wxART_OTHER, size));
     images.push_back(wxArtProvider::GetBitmapBundle(wxART_NORMAL_FILE, wxART_OTHER, size));
-    tree->SetImages(images);
+    //tree->SetImages(images);
 
     wxTreeItemId root = tree->AddRoot("wxAUI Project", 0);
     wxArrayTreeItemIds items;

```

At a DPI scaling of 200%, the buttons size is correct with the unmodified sample (with images in the `wxTreeCtrl`).

aui_dpi-200_tree-with-images

But **without images** in the `wxTreeCtrl` the **buttons** are **too small**.

aui_dpi-200_tree-without-images

At a DPI scaling of 300%, the buttons are much too small in an imageless `wxTreeCtrl`.

aui_dpi-300_tree-with-vs-without-images

- wxWidgets version  you use: 3.2.5 and master (d1506f40c67453ad5ad6799dc6e71a7e71ff56f4)

taler21:
Fixes #24640 while taking into account that users can set a custom indent via `wxTreeCtrl::SetIndent()`.

### DIFF
--- a/include/wx/msw/treectrl.h
+++ b/include/wx/msw/treectrl.h
@@ -266,6 +266,8 @@ protected:
 
     virtual void MSWUpdateFontOnDPIChange(const wxSize& newDPI) override;
 
+    void OnDPIChanged(wxDPIChangedEvent& event);
+
     // data used only while editing the item label:
     wxTextCtrl  *m_textCtrl;        // text control in which it is edited
     wxTreeItemId m_idEdited;        // the item being edited
@@ -347,6 +349,7 @@ private:
     friend class wxTreeSortHelper;
 
     wxDECLARE_DYNAMIC_CLASS(wxTreeCtrl);
+    wxDECLARE_EVENT_TABLE();
     wxDECLARE_NO_COPY_CLASS(wxTreeCtrl);
 };
 
--- a/include/wx/msw/treectrl.h
+++ b/include/wx/msw/treectrl.h
@@ -349,7 +349,6 @@ private:
     friend class wxTreeSortHelper;
 
     wxDECLARE_DYNAMIC_CLASS(wxTreeCtrl);
-    wxDECLARE_EVENT_TABLE();
     wxDECLARE_NO_COPY_CLASS(wxTreeCtrl);
 };
 
--- a/src/msw/treectrl.cpp
+++ b/src/msw/treectrl.cpp
@@ -717,10 +717,6 @@ bool wxTreeTraversal::Traverse(const wxTreeItemId& root, bool recursively)
     return true;
 }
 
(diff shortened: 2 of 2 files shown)


vadz:
Thanks, I don't remember why did I think to do it in `MSWUpdateFontOnDPIChange()` but this works too, of course, although I'd rather use `Bind()` than add an event table.

However a more important question is this one: from the original issue I understood that the problem only happened for the controls not using images, but this PR seems to do the adjustment unconditionally, wouldn't this result in scaling the indent twice when the images _are_ used? 

Also, I don't think we need to call `SetIndent()` at 100% DPI when creating the control, but this is minor too, the main question is whether we should always scale it on DPI change or only do if it we don't have any images?

taler21:
I think it is correct to always scale the indent.
Windows generally does not scale the indent, with or without images.
With images, it only ensures that the indent is at least large enough for the images to be fully visible.
So obvious display errors mainly appear with imageless trees.

>  I'd rather use Bind() than add an event table.

Is it not possible to backport the variant that uses/adds an event table?
The use of an event table is consistent with `wxListCtrl`, which also uses one.

> I don't think we need to call SetIndent() at 100% DPI when creating the control

Should I add `if ( GetDPIScaleFactor() > 1 )` as condition?

vadz:
> I think it is correct to always scale the indent.

Thanks, I'll merge this soon then (oops, not sure why did it get closed, this wasn't intentional).
 
> > I'd rather use Bind() than add an event table.
> 
> Is it not possible to backport the variant that uses/adds an event table? The use of an event table is consistent with `wxListCtrl`, which also uses one.
> 
> > I don't think we need to call SetIndent() at 100% DPI when creating the control
> 
> Should I add `if ( GetDPIScaleFactor() > 1 )` as condition?

taler21:
No objections, your changes are fine. Thanks!

taler21:

PR #24953, opened 2024-11-12
[macOS] Supporting PNG data on clipboard for macOS

hwiesmann:
PNG data on clipboards is currently not supported by wxWidgets when using macOS. Though, it seems to be that it can be supported basically out of the box.
Patches are included.

### DIFF
--- a/src/common/dobjcmn.cpp
+++ b/src/common/dobjcmn.cpp
@@ -702,9 +702,9 @@ bool wxCustomDataObject::SetData(size_t size, const void *buf)
 #define wxIMAGE_FORMAT_BITMAP_TYPE wxBITMAP_TYPE_PNG
 #define wxIMAGE_FORMAT_NAME "PNG"
 #elif defined(__WXOSX__)
-#define wxIMAGE_FORMAT_DATA wxDF_BITMAP
-#define wxIMAGE_FORMAT_BITMAP_TYPE wxBITMAP_TYPE_TIFF
-#define wxIMAGE_FORMAT_NAME "TIFF"
+#define wxIMAGE_FORMAT_DATA wxDF_PNG
+#define wxIMAGE_FORMAT_BITMAP_TYPE wxBITMAP_TYPE_PNG
+#define wxIMAGE_FORMAT_NAME "PNG"
 #else
 #define wxIMAGE_FORMAT_DATA wxDF_BITMAP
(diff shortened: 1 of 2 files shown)


vadz (commented):
Thanks, but this seems to _replace_ support for TIFF with PNG.

While I agree that PNG is a much more common and so probably more useful format than TIFF, shouldn't we rather support both of them? Or am I wrong and we didn't actually support TIFF before or, somehow, still support it now?

hwiesmann:
No, this should not be the case. If there is a bitmap on the clipboard, the bitmap is used. If there is a PNG on the clipboard, the PNG is used.

This code works for pasting (just an example):

````
     wxImage newImage;

     if (wxTheClipboard->IsSupported(wxDF_BITMAP))
      {
        wxBitmapDataObject bitmapDataObject;

        wxTheClipboard->GetData(bitmapDataObject);
        newImage = bitmapDataObject.GetBitmap().ConvertToImage();
      }
      else if (wxTheClipboard->IsSupported(wxDF_PNG))
      {
        wxImageDataObject imageDataObject;

        wxTheClipboard->GetData(imageDataObject);
        newImage = imageDataObject.GetImage();
      }
      else
        ;

````

vadz:
> No, this should not be the case. If there is a bitmap on the clipboard, the bitmap is used. If there is a PNG on the clipboard, the PNG is used.

But previously, if there was a TIFF in the clipboard, then (I believe) you could get it using `wxImageDataObject`, while now this (probably) doesn't work any more, does it?

Ideal would be to allow pasting any kind of image format using this class.

hwiesmann:
Hello,

Yes, this is correct. But now the implementation is equivalent to the implementation for the other operating systems. The other operating systems - as far as I can see it - accept PNGs only. 

To accept PNGs and bitmaps is rather difficult with the current implementation and without breaking the compatibility completely, I believe. I can implement it (at least for macOS I can also test it) but this requires that wxImageDataObject has to be derived from wxDataObjectComposite or directly from wxDataObject.

Keeping the derivation from wxDataObjectSimple and programming around its limitations can of course also be done but this is really ugly, IMHO.

vadz:
Sorry, I don't think we should apply this in the current state. I'd be still glad to merge a PR adding support for PNG in _addition_ to, rather than instead of, TIFF however, so if you could please make one, it would be very welcome. TIA!

PR #24954, opened 2024-11-13
Remove/fix outdated information in the internat sample and its readme.txt

lanurmi:
Internat sample's readme.txt is severely outdated, and gives unusual instructions for using `xgettext`. Fixing those instructions, and simply removing some outdated paragraphs. Also fixing/removing comments within `internat.cpp`.

### DIFF
--- a/samples/internat/internat.cpp
+++ b/samples/internat/internat.cpp
@@ -531,17 +531,8 @@ void MyFrame::OnPlay(wxCommandEvent& WXUNUSED(event))
     }
     else
     {
-        // this is a more implicit way to write _() but note that if you use it
-        // you must ensure that the strings get extracted in the message
-        // catalog as by default xgettext won't do it; it only knows of _(),
-        // not of wxTRANSLATE(). As internat's readme.txt says you should thus
-        // call xgettext with -kwxTRANSLATE.
-        str = wxGetTranslation(wxTRANSLATE("Bad luck! try again..."));
-
-        // note also that if we want 'str' to contain a localized string
-        // we need to use wxGetTranslation explicitly as wxTRANSLATE just
-        // tells xgettext to extract the string but has no effect on the
-        // runtime of the program!
+        // this is a more verbose way to write _()
+        str = wxGetTranslation("Bad luck! try again...");
     }
 
     wxMessageBox(str, _("Result"), wxOK | wxICON_INFORMATION);
--- a/samples/internat/internat.cpp
+++ b/samples/internat/internat.cpp
@@ -309,8 +309,6 @@ MyFrame::MyFrame()
     test_menu->Append(INTERNAT_TEST_MSGBOX, _("&Message box test"),
                       _("Tests message box buttons labels translation"));
 
-    // Note that all these strings are currently "translated" only in French
-    // catalog, so you need to use French locale to see them in action.
     wxMenu *macro_menu = new wxMenu;
     macro_menu->Append(INTERNAT_MACRO_1, _("item"));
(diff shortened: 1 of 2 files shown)


samples/internat/internat.cpp
```diff
@@ -533,17 +531,8 @@ void MyFrame::OnPlay(wxCommandEvent& WXUNUSED(event))
...
-        // not of wxTRANSLATE(). As internat's readme.txt says you should thus
-        // call xgettext with -kwxTRANSLATE.
-        str = wxGetTranslation(wxTRANSLATE("Bad luck! try again..."));
-
-        // note also that if we want 'str' to contain a localized string
-        // we need to use wxGetTranslation explicitly as wxTRANSLATE just
-        // tells xgettext to extract the string but has no effect on the
-        // runtime of the program!
+        // this is a more verbose way to write _()
+        str = wxGetTranslation("Bad luck! try again...");
```
PBfordev, line 535: IMO, this change may not be for the best. 

Firstly, I think that all the strings for translation (except where `wxTRANSLATE()` is really needed) should be marked just with `_()` macro, i.e., without the need to add a `wxGetTranslation` to gettext keywords. I also don't see any advantage of using the function instead of the macro, unless one is using additional parameters of the function (not demonstrated in the sample?).

samples/internat/readme.txt
```diff
@@ -24,74 +24,54 @@ Q. Why the message when I enter '17' is only partly translated?
...
-A. Please write them - see the next question. French is chosen by default
-   because it's the only translation which is currently available. To test
-   translations to the other languages please run the sample with 2 command line
-   arguments: the full language name and the name of the directory where the
-   message catalogs for this language are (will be taken as 2 first letters of
-   the language name if only 1 argument is given).
-
 Q. How to do translations to other language?
-A. First of all, you will need the GNU gettext tools (see the next questions).
+A. First of all, you will need the GNU gettext tools.
```
PBfordev, line 28: Since poedit is much nicer to work with for people used to GUI tools, I would add 
```suggestion
A. First of all, you will need the GNU gettext tools or, if you prefer GUI applications, a program such as poedit.
```
This would probably be better done somewhere earlier in the readme, but the alternative to gettext tools should be mentioned somewhere. Assuming wxWidgets sample readme is a good place for teaching people this stuff...

samples/internat/internat.cpp
vadz: The original idea was that you don't have to use `_()` if you don't want to/don't like it, so we do want to use `wxGetTranslation()` here.

OTOH I'd keep the part of the comment saying that you need to use `xgettext -kwxGetTranslation`.

samples/internat/readme.txt
vadz: Yes, I agree that it would be worth mentioning Poedit.

As for doing it in this README -- the best would probably be to move its entire contents into i18n overview and add a comment to the sample pointing to it.

samples/internat/internat.cpp
vadz: This is weird, I was absolutely certain that `_` was in the default keywords list, but according [to the docs](https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html) it isn't. I wonder if this could have changed...

In any case, I think we should document `-kwxPLURAL` too.

vadz:
I'm going to merge this, even though the instructions should still be improved IMHO and we  should move the contents of this README into the manual too. But best is the enemy of good and for now this does improve things, thanks.

lanurmi:
Agreed, the instructions could be improved, though as far as listing required `xgettext` keywords goes, I don't think the 540th line in the middle of internat sample is *the* place where people go looking for such information. That's why I concluded that sort of improvement to be outside the scope of this PR.

PR #24960, opened 2024-11-18
[wxOSX] Fix wxEvent timestamp narrowing

artur-so:
`wxEvent::m_timeStamp` is of type `long`, expressed in miliseconds. On macOS this value is assigned after conversion from `[NSEvent timestamp]` which is of type `double`, expressed in seconds. In several such assigment instances the original value is narrowed to `int`, apparently to silence implicit conversion warnings, see e490b0d.

However, the narrowing limits the applicability to `MAX_INT` milliseconds. Given that `int` is 32 bit on mac and that `NSEvent`'s `timestamp` returns the time in seconds since system startup, an interval between two events can only be checked when macOS uptime is less then 24.9 days.

As macs tend to run quite stable for long periods of time without the need for a reboot, this is a real-world problem.

Fix this by casting the source milliseconds value to `long`, which is 64 bit on mac.

### DIFF
--- a/src/osx/cocoa/window.mm
+++ b/src/osx/cocoa/window.mm
@@ -455,7 +455,7 @@ void wxWidgetCocoaImpl::SetupKeyEvent(wxKeyEvent &wxevent , NSEvent * nsEvent, N
     wxevent.m_rawCode = [nsEvent keyCode];
     wxevent.m_rawFlags = modifiers;
 
-    wxevent.SetTimestamp( (int)([nsEvent timestamp] * 1000) ) ;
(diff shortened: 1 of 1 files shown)


artur-so:
Sorry, I have provided no sample that illustrate the problem. This is mainly because external conditions are required to reproduce, see description above.

csomor:
@artur-so thanks a lot

vadz:
Thanks for fixing this, I'll merge it soon.

The only question I have is whether we should perhaps round this, instead of truncating. But I am not sure if it's really going to change anything in practice.

PR #24961, opened 2024-11-19
[wxCocoa] Implement wxCursor::GetHotSpot() method

solemnwarning:


vadz (commented):
Looks good, thanks, will merge soon after replacing `wxOVERRIDE` with `override` unless @csomor has any comments.

csomor (approved):
looks good to me, thanks a lot

Blake-Madden:
You should update `GetHotSpot()` in "interface/cursor.h" also to note that it is supported on macOS now (since 3.3).

PR #24967, opened 2024-11-21
Allow specifying synopsis for `wxCmdLineParser::Usage()` and wrap its output

vadz:
This is a pretty simple change that I think can be generally useful.


vadz:
I couldn't help myself and also added wrapping of help text in `wxCmdLineParser::Usage()` as it looked really ugly with long strings otherwise.

PR #24968, opened 2024-11-23
Add support for HTML ID attribute

vadz:
This is a slightly simplified (IMO) version of #24915.

It contains other cleanups, but the one I originally mentioned in the other PR discussion is in the last commit: instead of setting the ID manually, it just passes the tag to the ctors. I think this is preferable because there is no risk to forget to set the ID like this, and if there are no objections I'm going to squash this commit with the main 24267dad122e774f661231a5ac6a422c5af96674 one and merge this.


oltolm:
Looks good to me. Thank you for finishing it.

vadz:
> Looks good to me. Thank you for finishing it.

Thanks for reviewing! I've fixed a small compilation problem, rebased and will merge it once the CI builds pass.

PR #24972, opened 2024-11-25
wxRibbonToolBar does not remove tool highlight after moving mouse cursor to a disabled tool

RickvanderS:
Steps to reproduce:
1) Compile and run ribbon sample
2) In the toolbar group, hover the mouse cursor over the 'new' button
3) The 'new' button now turns yellow as all enabled buttons do when the mouse hovers it
4) Move the mouse cursor directly to the disabled 'open' button, without leaving the toolbar first
5) The 'new' button remains yellow. 
Expected behavior is that the 'new' button goes back to its normal state.


vadz:
Looks good to me, thanks! Will merge soon.

PR #24978, opened 2024-11-27
MSW: Make wxGetUserName only return false if it fails to load anything

### LINKED ISSUE #24976: wx 3.3 msw wxGetUserName problem
Hi,

wxGetUserName does not deliver the result like wx did in 3.2.6 for msw.
The Windows function NetUserGetInfo does not return a proper result.
Even the MS example code in the NetUserGetInfo documentation does not return a result. I assume if a computer name is not given, there won't be a result from that function.
Also the 3.2.6. code contains a more detailed error message than the 3.3 version which makes identification the problem less straightforward.

Blake-Madden:
If it at least can load the login name, then return `true`, even if the Net API calls to get the full user name fails. (The Net API requires permissions that may not be available for the current user.)

Fixes #24976

### DIFF
--- a/src/msw/utils.cpp
+++ b/src/msw/utils.cpp
@@ -333,7 +333,7 @@ bool wxGetUserName(wxChar* buf, int maxSize)
 
     if ( status != NERR_Success )
     {
-        wxLogWarning(_("Failed to retrieve full user information: %s"),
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, I've (squash) merged this now.

PR #24979, opened 2024-11-27
wxImageDataObject is able to handle wxDF_BITMAP and wxDF_PNG data formats

hwiesmann:
wxImageDataObject has been re-written to handle wxDF_BITMAP and wxDF_PNG data format on all platforms.
As macOS seems to have been the only platform that did not support the wxDF_PNG format this is now also supported on macOS.

As a test case the sample "image" has been used.

PS: As a consequence to handle two different data formats wxImageDataObject is no longer derived from wxCustomDataObject, respectively wxDataObjectSimple, but directly from wxDataObject.


vadz:
I didn't look at the changes yet, but wxMSW CI errors look real.

vadz:
There is still a test failure in `wxImage::Clipboard` which seems valid, i.e. it happens in several different builds. Did you have a chance to check it?

hwiesmann:
Hello,

I do not have access to Windows or GTK platforms at the moment. Therefore, I cannot really test it by myself. But there is, in my opinion, nothing used that is platform specific.

PR #24981, opened 2024-11-27
wxDataObject::IsSupportedFormat optimization

hwiesmann:
In general, wxDataObject::IsSupportedFormat is not documented, though wxDataObjectBase::IsSupported is.

On macOS, qt and X11 wxDataObject::IsSupportedFormat's implementation is identical to wxDataObjectBase::IsSupported. Therefore, on these platforms wxDataObject::IsSupportedFormat is now calling wxDataObjectBase::IsSupported.
On MSW wxDataObject::IsSupportedFormat is not defined as on all the other platforms (missing direction parameter). But the current implementation already calls wxDataObjectBase::IsSupported.
On GTK wxDataObject::IsSupportedFormat is slightly differently implemented and its implementation is therefore kept.

### DIFF
--- a/src/osx/carbon/dataobj.cpp
+++ b/src/osx/carbon/dataobj.cpp
@@ -270,29 +270,7 @@ wxDataObject::wxDataObject()
 
 bool wxDataObject::IsSupportedFormat( const wxDataFormat& rFormat, Direction vDir ) const
 {
-    size_t nFormatCount = GetFormatCount( vDir );
-    bool found = false;
-
-    if (nFormatCount == 1)
-    {
-        found = (rFormat == GetPreferredFormat());
-    }
-    else
-    {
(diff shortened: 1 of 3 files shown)


vadz:
I think wxGTK implementation is actually the same too because `wxDataFormat::operator==()` calls the same `wxGTKIsSameFormat()` as is called there, so I'd really just define `IsSupportedFormat()` as a trivial inline wrapper in `wxDataObjectBase` itself and remove all the other definitions completely. What do you think?

vadz:
> Hi, Actually, I did not have a look at wxDataFormat::operator==(). If it calls wxGTKIsSameFormat(), then wxDataObjectBase::IsSupported can be used for GTK, too.

Just to check: do you plan to update this PR to do it?

PR #24982, opened 2024-11-27
Make wxTextCtrl::SetLabel() consistently do nothing everywhere

vadz:
Previously it did the same thing as SetValue() in wxMSW but didn't do anything in wxGTK (and probably other ports), which made it easy to write code that only worked under Windows but mysteriously didn't show any text under Linux.

Change it to not do anything anywhere and add an assert to help detecting wrong uses of this function.

Also add a unit test checking this behaviour.

----

This is an incompatible change and it would, in principle, avoid it by making `wxTextCtrl::SetLabel()` call `ChangeValue()` instead, but it seems to be cleaner to not allow it unless this is done more commonly than I think — please let me know if you think we should avoid breaking Windows-only programs using `SetLabel()` to change the control's value.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -113,6 +113,10 @@ Changes in behaviour not resulting in compilation errors
   but may require removing any compensation for not doing this before that may
   be present in the application code.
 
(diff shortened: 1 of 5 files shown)


vadz:

PR #24983, opened 2024-11-28
Stop using checkout and other actions for Ubuntu 18.04 CI builds

vadz:
Even actions/checkout@v3 doesn't work in the container any longer, because none of JS actions can run on Ubuntu 18.04 which doesn't have a new enough glibc version any longer, so perform the checkout manually.

Disable ccache for the builds using this OS too for the same reason: even though we could install ccache itself manually too, we don't have any easy way to save and restore its cache without this action.

(cherry picked from commit 9d1c5b108ce96cc61604a3d8b49a171b8a293fed)


vadz:
I'm going to merge this for now to fix the CI builds. See my wx-dev mail about further possibilities.

vadz:

PR #24984, opened 2024-11-28
Fix minor typos in wxAppProgressIndicator help

Blake-Madden:


interface/wx/appprogress.h
```diff
@@ -45,10 +45,10 @@ class WXDLLIMPEXP_CORE wxAppProgressIndicator
         Check if the application progress display is available.
 
         Currently this only returns @true when using wxMSW and running
-        under Windows 7 or later systems (which provide task bar button
+        under Windows 7 or later systems (which provides the task bar button
```
vadz, line 48: We could just remove the mention of the version completely, couldn't we? We don't support anything earlier anyhow.

Blake-Madden: Oh, that's right. I just updated it.

PR #24997, opened 2024-12-09
Allow opting out of catching unhandled C++ exceptions

lanurmi:
Add a system option to control whether the main loop and threads are run
from within a `try`-`catch` block, or outside of it.

Traditionally, wxWidgets has caught all unhandled C++ exceptions
that are about to cause the program to be aborted.

For many situations this is probably a good thing, but the downside
is that it severely complicates debugging the source of such exceptions.
The backtrace will only show the stack since wx's exception handler,
but not where the exception occurred in the first place.

Any crashdumps created by the OS also suffer from the same issue;
backtraces are mostly useless in these cases.

The following classes with `try`-`catch` blocks do not obey the new system
option for now: `wxDataViewRendererBase`, `wxDocTemplate`, `wxIDropTarget`

---

Attached here is [a patch for triggering a crash (abort)](https://github.com/user-attachments/files/18066198/wx-trigger-crash.patch) from a sample, to help demonstrating what happens with the new option set or not set.

Minimal sample: Help -> About
Threads sample: Thread -> Start a new thread

### DIFF
--- a/interface/wx/sysopt.h
+++ b/interface/wx/sysopt.h
@@ -37,6 +37,15 @@
         this option allows changing it without modifying the program code and
         also applies to asserts which may happen before the wxApp object
         creation or after its destruction.
+    @flag{catch-unhandled-exceptions}
+        If set to zero, wxWidgets will not catch unhandled exceptions, but
+        rather lets the default behavior of aborting the program take place.
+        Not catching unhandled exceptions makes debugging easier, as the
+        backtrace is more likely to show what actually happened, and where.
+        The same applies to any crash dumps generated due to unhandled exceptions.
+        By default unhandled exceptions are eventually caught by wxWidgets.
+        This flag should be set very early during program startup, within
+        the constructor of the wxApp derivative.
     @endFlagTable
 
     @section sysopt_win Windows
--- a/src/common/event.cpp
+++ b/src/common/event.cpp
@@ -22,6 +22,7 @@
 #include "wx/event.h"
 #include "wx/eventfilter.h"
 #include "wx/evtloop.h"
+#include "wx/sysopt.h"
 
 #ifndef WX_PRECOMP
     #include "wx/list.h"
@@ -1666,6 +1667,10 @@ bool wxEvtHandler::TryHereOnly(wxEvent& event)
 
 bool wxEvtHandler::SafelyProcessEvent(wxEvent& event)
 {
+    if ( wxSystemOptions::IsFalse("catch-unhandled-exceptions") )
+    {
+        return ProcessEvent(event);
+    }
 #if wxUSE_EXCEPTIONS
     try
     {
--- a/src/common/evtloopcmn.cpp
+++ b/src/common/evtloopcmn.cpp
@@ -243,6 +243,96 @@ bool wxEventLoopManual::ProcessEvents()
     return res;
 }
 
+void wxEventLoopManual::Loop()
+{
+    // this is the event loop itself
+    for ( ;; )
+    {
+        // give them the possibility to do whatever they want
+        OnNextIteration();
+
+        // generate and process idle events for as long as we don't
+        // have anything else to do, but stop doing this if Exit() is
+        // called by one of the idle handlers
+        //
+        // note that Pending() only checks for pending events from the
+        // underlying toolkit, but not our own pending events added by
+        // QueueEvent(), so we need to call HasPendingEvents() to check
+        // for them too
+        while (!m_shouldExit
+                   && !Pending()
+                       && !(wxTheApp && wxTheApp->HasPendingEvents())
+                           && ProcessIdle())
+            ;
+
+        // if Exit() was called, don't dispatch any more events here
+        if ( m_shouldExit )
+            break;
+
+        // a message came or no more idle processing to do, dispatch
+        // all the pending events and call Dispatch() to wait for the
+        // next message
+        if ( !ProcessEvents() || m_shouldExit )
+            break;
+    }
+
+    // Process any still pending events.
+    for ( ;; )
+    {
+        bool hasMoreEvents = false;
+
+        // We always dispatch events pending at wx level: it may be
+        // important to do it before the loop exits and e.g. the modal
+        // dialog possibly referenced by these events handlers is
+        // destroyed. It also shouldn't result in the problems
+        // described below for the native events and while there is
+        // still a risk of never existing the loop due to an endless
+        // stream of events generated from the user-defined event
+        // handlers, we consider that well-behaved programs shouldn't
+        // do this -- and if they do, it's better to keep running the
+        // loop than crashing after leaving it.
+        if ( wxTheApp && wxTheApp->HasPendingEvents() )
+        {
+            wxTheApp->ProcessPendingEvents();
+            hasMoreEvents = true;
+        }
+
+        // For the underlying toolkit events, we only handle them when
+        // exiting the outermost event loop but not when exiting nested
+        // loops. This is required at least under MSW where, in case of
+        // a nested modal event loop, the modality has already been
+        // undone as Exit() had been already called, so all UI elements
+        // are re-enabled and if we dispatched events from them here,
+        // we could end up reentering the same event handler that had
+        // shown the modal dialog in the first place and showing the
+        // dialog second time before its first instance was destroyed,
+        // resulting in a lot of fun.
+        //
+        // Also, unlike wx events above, it should be fine to dispatch
+        // the native events from the outer event loop, as any events
+        // generated from outside the dialog itself (necessarily, as
+        // the dialog is already hidden and about to be destroyed)
+        // shouldn't reference the dialog. Which is one of the reasons
+        // we still dispatch them in the outermost event loop, to
+        // ensure they're still processed. Another reason is that if we
+        // do have an endless stream of native events, e.g. because we
+        // have a timer with a too short interval, it's arguably better
+        // to keep handling them instead of exiting.
+        if ( gs_eventLoopCount == 1 )
+        {
+            if ( Pending() )
+            {
+                Dispatch();
+                hasMoreEvents = true;
+            }
+        }
+
+        if ( !hasMoreEvents )
+            break;
+    }
+
+}
+
 int wxEventLoopManual::DoRun()
 {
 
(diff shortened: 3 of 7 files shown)


src/common/evtloopcmn.cpp
```diff
@@ -254,95 +345,17 @@ int wxEventLoopManual::DoRun()
 #if wxUSE_EXCEPTIONS
     for ( ;; )
     {
+        if ( wxSystemOptions::IsFalse("catch-unhandled-exceptions") )
+        {
+            Loop();
+            // exit the outer loop as well
+            break;
+        }
+        else
```
vadz, line 354: This looks weird to me, why do we put this inside a loop only to break out of it unconditionally? Why not just put

```cpp
if ( ...propagate exceptions... )
{
    Loop();
    return m_exitcode;
}
```

?

src/common/init.cpp
```diff
@@ -553,26 +576,14 @@ int wxEntryReal(int& argc, wxChar **argv)
         return wxApp::GetFatalErrorExitCode();
     }
 
+    if ( wxSystemOptions::IsFalse("catch-unhandled-exceptions") )
+    {
+        return DoEntryReal();
+    }
+    else
```
vadz, line 583: Here I'd just remove this `else`.

src/msw/thread.cpp
```diff
@@ -500,34 +504,50 @@ class wxThreadKeepAlive
 /* static */
 void wxThreadInternal::DoThreadOnExit(wxThread *thread)
 {
+    if ( wxSystemOptions::IsFalse("catch-unhandled-exceptions") )
+    {
+        thread->OnExit();
+    }
+    else
```
vadz, line 511: And remove this one too.

src/msw/thread.cpp
```diff
@@ -452,6 +453,9 @@ class wxThreadInternal
     // really start the thread (if it's not already dead)
     static THREAD_RETVAL DoThreadStart(wxThread *thread);
 
+    // really really start the thread, without catching exceptions
+    static THREAD_RETVAL DoDoThreadStart(wxThread *thread);
```
vadz, line 457: Sorry but I really hate these "DoDo" names. Let's call it `DoThreadStartWithoutExceptionHandling()` or `DoThreadStartWithoutEH()` or even `DoThreadStartNoEH()` if it's too long.

Alternatively we could avoid having to come up with a name for it by using a lambda inside the function.

interface/wx/sysopt.h
```diff
@@ -37,6 +37,15 @@
         this option allows changing it without modifying the program code and
         also applies to asserts which may happen before the wxApp object
         creation or after its destruction.
+    @flag{catch-unhandled-exceptions}
```
vadz, line 40: I don't have a strong opinion about this, but it seems a bit strange to have an option which is 1 by default. Maybe we should have `allow-exceptions-escape` or something like that with the opposite meaning instead?

vadz (commented):
Thanks, I agree that it can be useful to disable built-in exception handling in some situations, but I have a few questions about the details.

interface/wx/sysopt.h
lanurmi: I don't think it is that strange. IMO nothing in system options' documentation or the name itself suggests that options are inherently 0 by default. (Also cf. Cocoa's [NSUserDefaults](https://developer.apple.com/documentation/foundation/nsuserdefaults)). In fact, there are two existing system options that are 1 by default: `msw.notebook.themed-background` and `msw.staticbox.optimized-paint`.

Enforcing 0-by-default is bound to lead to more or less awkward naming with made up opposite meanings. E.g. in `allow-exceptions-escape`, 'escape' is not a term that the C++ standard uses with exceptions at all.

src/msw/thread.cpp
lanurmi: But then `thread->OnExit()` would get called twice.

src/common/evtloopcmn.cpp
lanurmi: Can be done.

interface/wx/sysopt.h
vadz: We could use `disable-catch-unhandled`.

Anyhow, as I said, I don't feel that strongly about this, so if you do, we can leave it like this but I'd still be curious if nobody else really finds this awkward, as I do.

src/common/event.cpp
```diff
@@ -1658,6 +1659,11 @@ bool wxEvtHandler::TryHereOnly(wxEvent& event)
 
 bool wxEvtHandler::SafelyProcessEvent(wxEvent& event)
 {
+    if ( wxSystemOptions::IsFalse("catch-unhandled-exceptions") )
+    {
+        return ProcessEvent(event);
+    }
+    else
```
vadz, line 1666: Minor, but to avoid any weird indentation-related problems I'd just do this

```suggestion
```

src/msw/thread.cpp
vadz: Well, I meant to remove it after putting a `return` in the `if` above, i.e. make it unnecessary, not just mechanically remove it.

src/msw/thread.cpp
```diff
@@ -500,34 +504,50 @@ class wxThreadKeepAlive
...
 /* static */
 THREAD_RETVAL wxThreadInternal::DoThreadStart(wxThread *thread)
 {
     wxON_BLOCK_EXIT1(DoThreadOnExit, thread);
 
     THREAD_RETVAL rc = THREAD_ERROR_EXIT;
 
+    if ( wxSystemOptions::IsFalse("catch-unhandled-exceptions") )
+    {
+        rc = DoThreadStartWithoutExceptionHandling(thread);
```
vadz, line 545:
```suggestion
        return DoThreadStartWithoutExceptionHandling(thread);
```

and remove `else` below (and move this above `rc` declaration).

src/unix/threadpsx.cpp
```diff
@@ -879,6 +880,15 @@ void *wxThreadInternal::PthreadStart(wxThread *thread)
                    wxT("Thread %p about to enter its Entry()."),
                    THR_ID(pthread));
 
+        if ( wxSystemOptions::IsFalse("catch-unhandled-exceptions") )
+        {
+            pthread->m_exitcode = thread->Entry();
+
+            wxLogTrace(TRACE_THREADS,
+                       wxT("Thread %p Entry() returned %lu."),
+                       THR_ID(pthread), wxPtrToUInt(pthread->m_exitcode));
```
vadz, line 889: I know this is a much smaller piece of code, but it would be more consistent to move it into a function too. Or maybe we could use the helper taking lambda, see the other comment.

src/unix/threadpsx.cpp
```diff
@@ -1734,6 +1744,11 @@ void wxThread::Exit(ExitCode status)
     // might deadlock if, for example, it signals a condition in OnExit() (a
     // common case) while the main thread calls any of functions entering
     // m_critsect on us (almost all of them do)
+    if ( wxSystemOptions::IsFalse("catch-unhandled-exceptions") )
```
vadz, line 1747: IMHO it would be nice if we could replace all this section (including the `wxTRY` below) with something like this:

```cpp
CallPossiblyHandlingExceptions([this]() { OnExit(); });
```

where `CallPossiblyHandlingExceptions()` could be a template or just take `std::function<void ()>` (or maybe a template taking `std::function<T ()>` and returning a value of type `T`) defined in the straightforward way.

It's probably not worth to use it for big chunks of code (although maybe it is?), but for one line, or a few line (like above in this file) fragments it looks nicer.

vadz (commented):
Thanks, with the new functions this looks much better but I rather like (sorry for conceitedness) the idea of `CallPossiblyHandlingExceptions()`, what do you think?

vadz:
@lanurmi What do you think of the change proposed above?

src/unix/threadpsx.cpp
lanurmi: I'm not sure how useful a `CallPossiblyHandlingExceptions()` would be in practice. Calling something conditionally with or without exception handling is not a terribly common pattern. It would be used at about 7 places (at most) in code, and after all it would replace pretty simple and readable `if` statements with something less obvious, and only save a few lines of code, total.

lanurmi:
Oh yes, I've been meaning to get back to this PR. I'm pushing most proposed changes as fixups in a moment. (I notice not all the fixups squash cleanly, I can do a squash + rebase later.)

src/unix/threadpsx.cpp
vadz: I dislike having duplicated code, even if it's just a few lines, I've seen too many times just one version of it getting modified later (and it's difficult to see that something that should be modified is _not_, when looking at the diffs during the code review).

And "calling with or without exception handling" looks like a pattern, i.e. something it could be useful to provide a higher level abstraction for.

Note that this could also probably be done with macros, e.g.

```cpp
wxMAYBE_TRY
{
   ... code ...
}
wxMAYBE_CATCH_ALL( ... handler ... );
```

which would be even simpler but probably rather unclear.

lanurmi: So thinking more about this, `CallPossiblyHandlingExceptions()` should not only take an `std::function` that it calls, but also another `std::function` for the actions that are to be executed `CATCH_ALL` block (because not all `CATCH_ALL` blocks are identical). And not all `catch` blocks are `CATCH_ALL`, so there would need to be, possibly, a third `std::function` that is called for exceptions of type `abi::__forced_unwind&`. And how nicely does [this](https://github.com/wxWidgets/wxWidgets/blob/c8980d531e83dabff5c278c4e4b35eb13d7fe65f/src/common/evtloopcmn.cpp#L351-L380) turn into a nice little lambda function.

vadz: It wouldn't have to be a lambda, of course, the same criteria used for using lambdas in other context would apply, i.e. if it's too long, it should be a real function.

lanurmi: Lambda or not, I still don't know how could one extract a `try-catch-catch-catch` logic into a function, decoupled from the actual function one wanted to call in the first place.

Since you have a clear vision of what `CallPossiblyHandlingExceptions()` should be like, I think your time is better spent implementing it, rather than trying to explain to me how it should be implemented. I consider this PR to be completed for my part.

vadz:
Replaced by #25257.

Issue #23518, opened 2023-05-04
wxColour names are handled differently by FromString() than by GetAsString() on Linux


taler21:
### Description

The functionality of `wxColour::FromString()` with wxGTK is incompatible with `wxColour::GetAsString()` and is also incompatible with other platforms (e.g. wxMSW).

The incompatible behaviour can be reproduced with the following changes to the minimal sample:
```patch
diff --git "a/samples/minimal/minimal.cpp" "b/samples/minimal/minimal.cpp"
index 5f32257c6b..d317968316 100644
--- "a/samples/minimal/minimal.cpp"
+++ "b/samples/minimal/minimal.cpp"
@@ -120,6 +120,15 @@ bool MyApp::OnInit()
     if ( !wxApp::OnInit() )
         return false;
 
+    wxColour col1( 0, 255, 0 );
+    wxString name = col1.GetAsString(); // Returns the color name 'green'
+    wxColour col2( name );              // With wxGTK (0, 128, 0) instead of (0, 255, 0)
+    if ( col1 != col2 )
+    {
+        wxLogError("Mismatch of color and text: %s -> %s -> %s", col1.GetAsString(wxC2S_CSS_SYNTAX),
+                                                           name, col2.GetAsString(wxC2S_CSS_SYNTAX) );
+    }
+
     // create the main application window
     MyFrame *frame = new MyFrame("Minimal wxWidgets App");
 

```

I think the fix will be to call `FromString()` of the base class first and the platform specific function only as a fallback.
```patch
diff --git "a/src/gtk/colour.cpp" "b/src/gtk/colour.cpp"
index 41b2ae2a11..89441210e0 100644
--- "a/src/gtk/colour.cpp"
+++ "b/src/gtk/colour.cpp"
@@ -260,6 +260,9 @@ wxColour::operator const GdkRGBA*() const
 
 bool wxColour::FromString(const wxString& str)
 {
+    if ( wxColourBase::FromString(str) )
+        return true;
+
 #ifdef __WXGTK3__
     GdkRGBA gdkRGBA;
     if (gdk_rgba_parse(&gdkRGBA, str.utf8_str()))
@@ -276,5 +279,5 @@ bool wxColour::FromString(const wxString& str)
     }
 #endif
 
-    return wxColourBase::FromString(str);
+    return false;
 }

```

### Platform and version information

- wxWidgets version  you use: 3.2.2.1
- wxWidgets port  you use: wxGTK
[...]

PR #25002, opened 2024-12-13
Change values of standard colours to follow CSS standard

vadz:
This implements my proposal from #23518. It is obviously backwards-incompatible but I've tried to make it as compatible as possible and hope this should be good enough. Please try to test your applications with this PR and let me know if you see something really bad and/or if using `UseScheme(Traditional)` doesn't work for you to restore the original appearance.

@paulcor Do you think we should remove wxGTK-specific `wxColour::FromString()` now? The effect of doing it would be to make wxGTK use the same colours as wxMSW used to use if `UseScheme(Traditional)` is called, while continuing to use the same colours as today by default. I'm not 100% if this is desirable, but I think it might be.


paulcor:
> Do you think we should remove wxGTK-specific wxColour::FromString() now?

Yes, I think it's better to just do it one way.

PR #25004, opened 2024-12-15
Fix restoring previously maximized pane when loading wxAUI layout

### LINKED ISSUE #24990: Restoring wxAUI state with a maximized pane doesn't work correctly
To reproduce:

1. Run aui sample.
2. Maximize the "tree" pane on the left.
3. Press `Ctrl-C` to copy layout co clipboard.
4. Exit the sample.
5. Run it again.
6. Press `Ctrl-V`.

It gets in a weird state when the tree pane is logically maximized but doesn't appear as such.

To fix this, we need to call `MaximizePane()` on the pane which should be maximized instead of just setting `m_hasMaximized` to `true` in `LoadLayout()` (NB: this code was copied from `LoadPerspective()` but it almost certainly doesn't work there neither).

vadz:
Setting m_hasMaximized=true, as the existing code in LoadPerspective() did, is not enough to make the pane really maximized, and doing this resulted in broken layout when restoring a layout with a maximized pane.

Fix this by calling MaximizePane() instead.

Note that this function only works if passed a reference to one of the panes managed by wxAuiManager and not its copy, so we must use GetPane() to get the proper pane to use.

Closes #24990.

### DIFF
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -1456,10 +1456,8 @@ void wxAuiManager::LoadLayout(wxAuiDeserializer& deserializer)
 {
     deserializer.BeforeLoad();
 
-    // Note that we don't reset m_hasMaximized here but use a local variable so
-    // as to avoid modifying m_hasMaximized in case one of the deserializer
-    // functions called below throws an exception.
-    bool hasMaximized = false;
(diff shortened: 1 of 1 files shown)


vadz:
Merged in the commit above.

PR #25008, opened 2024-12-17
Save from STC using wxConvAuto, not current code page

Blake-Madden:
I think STC should be saving as UTF-8, not the current character set. That seems to be the safest option, and UTF-8 tends to be the default elsewhere.

To demonstrate the issue:

Windows 11, latest wx 3.3

1. Open the STC sample
2. Open "de.po" from the locale folder
3. Type a letter to make the document dirty
4. Select Save As and save it
5. Note that the file that was saved to is totally empty

### DIFF
--- a/src/stc/stc.cpp
+++ b/src/stc/stc.cpp
@@ -5203,7 +5203,7 @@ wxStyledTextCtrl::DoSaveFile(const wxString& filename, int WXUNUSED(fileType))
     wxFile file(filename, wxFile::write);
 #endif
 
-    if ( file.IsOpened() && file.Write(GetValue(), wxConvUTF8) )
+    if ( file.IsOpened() && file.Write(GetValue()) )
     {
         SetSavePoint();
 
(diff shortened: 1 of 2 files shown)


vadz:
Can we just remove this parameter to use the default `wxConvAuto` instead? This would do exactly the same thing but be shorter and more consistent with `ReadAll()` (where this parameter could also be removed).

Finally, please modify `stc.cpp.in` and run `gen_iface.py` to avoid losing these changes the next time this file is regenerated. TIA!

Blake-Madden:
> Can we just remove this parameter to use the default `wxConvAuto` instead? This would do exactly the same thing but be shorter and more consistent with `ReadAll()` (where this parameter could also be removed).

Done. My tests checked out OK.

> Finally, please modify `stc.cpp.in` and run `gen_iface.py` to avoid losing these changes the next time this file is regenerated. TIA!

Done, `stc.cpp.in` is updated. I ran `gen_iface.py`, but the interface and include .h files were the same, so nothing to check in for those.

PR #25012, opened 2024-12-17
Fix syntax errors in gen_iface.py

Blake-Madden:
Add `r` prefix in front of each segment of raw string; otherwise, these segments aren't seen as raw strings and causes a syntax error. Refer to https://stackoverflow.com/questions/46003452/how-to-correctly-write-a-raw-multiline-string-in-python.

Tested with Python 3.13.1 on Windows 11 (which is where I got syntax errors for each line the first time I ran this file, runs fine after change).


vadz:
I had no idea this has changed, but thanks for fixing it, will merge soon unless any Pythonistas object.

vadz:
I'm merging this but I still wonder why I don't see any errors here, with Python 3.11 too.

PR #25013, opened 2024-12-18
Fix numbers in translations

Blake-Madden:
Mostly just paper sizes and code pages

### DIFF
--- a/locale/ms.po
+++ b/locale/ms.po
@@ -2655,7 +2655,7 @@ msgstr "Helai A4, 210 x 297 mm"
 
 #: ../src/common/paper.cpp:95
 msgid "C sheet, 17 x 22 in"
-msgstr "Helai D, 17 x 22 in"
+msgstr "Helai C, 17 x 22 in"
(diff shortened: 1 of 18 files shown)


locale/ms.po
```diff
@@ -2655,7 +2655,7 @@ msgstr "Helai A4, 210 x 297 mm"
 
 #: ../src/common/paper.cpp:95
 msgid "C sheet, 17 x 22 in"
-msgstr "Helai D, 22 x 34 in"
+msgstr "Helai D, 17 x 22 in"
```
lanurmi, line 2658: Here presumably also D should be C.

Blake-Madden: Got it, thanks.

vadz:
Nice, thanks! Did you find all of them on your own or do you have a tool detecting this kind of errors?

Blake-Madden:
I'm working on a tool to detect i18n issues like this...built with wxWidgets, of course :)

PR #25015, opened 2024-12-21
Fix malformed accelerator keys in translations

Blake-Madden:
Had `%` instead of `&` in translations.


vadz:

PR #25020, opened 2024-12-23
Clean up wxGTK `wxDataViewCtrl` code

vadz:
Use range-for loops, `std::vector`, `std::sort()` and other minor cleanup.


vadz:
There is a number of changes here but none of them should have any user-visible effect.

Of course, please let me know if this "should" doesn't work.

PR #25030, opened 2024-12-28
Add support for printing multiple page ranges

vadz:
This includes #23525 but also implements support for this in wxGTK as well as fixing a couple of other problems in this port and cleaning up the code in both it and wxOSX (but without implementing multiple pages support under Mac).

Any reviews of the changes and _especially_ of the API would be welcome!

### DIFF
--- a/include/wx/cmndata.h
+++ b/include/wx/cmndata.h
@@ -145,6 +145,32 @@ private:
 };
 
 /*
+ * wxPrintPageRange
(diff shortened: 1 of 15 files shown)


vadz:
@sz5000 Please let me know if you have objections to merging this, otherwise I'll do it soon.

sz5000:
Hi, I have no objections.

PR #25034, opened 2024-12-30
remove unwanted extension appending

### LINKED ISSUE #24949: wxFileDialog(wxFD_SAVE) handling of extension on entering new file name without extension is inconsistent over platforms
### Description
wxFileDialog(wxFD_SAVE) handling of extension on entering new file name without extension is inconsistent over platforms.

#### Bug description:

When using a wxFileDialog for acquiring the name of the file to save to, the behavior with respect to the returned file path is inconsistent when comparing wxMSW to wxGTK or wxOSX.

Both on wxGTK and wxOSX the `wxFileDialog::GetPath` method will *never* append a default extension whenever  a file name without extension is entered by the user. In these cases `GetPath` will simply return the full path for the file name without extension as entered by the user. 

On wxMSW however `GetPath` will always append an extension based on the default filter index as set with `wxFileDialog::set_filter_index` before showing the dialog *except* when the default filter has a wildcard extension ('.*').

I would expect the same behavior on all platforms and a generic way to either suppress extension appending when this behavior is the default or a generic way to activate extension appending when not appending is the default. 

At a minimum I would expect the wxMSW implementation to honor the presence of a filter with wildcard extension ('.*') by not appending an extension whenever the wildcard extension filter was selected by the user irrespective of the filter set as the default as the presence of a wildcard extension filter should be a clear indication that the app developer intended to allow the user to enter any file name (incl. file names without extension).

As it is now it is all but impossible to implement some clean, multiplatform, code that allows an app to have a file dialog to select a file name without extension to save data to when there are also preferred (but not *mandatory*) extensions involved.

#### To Reproduce:

The current standard 'dialogs' sample demonstrates clearly the platform inconsistencies.
Running the [File operations] \ [Save file] option from this sample on wxMSW will *always* return a filename with extension (automatically appending the default extension '.doc' for any file name without extension entered).
Updating the sample by adding an 'Any files (*.*)|*.*' filter to the `wildCard` argument for the wxFileDialog constructor in method `MyFrame::FileSave` will show that even when the user selects this 'Any files' filter and enters a (non-existing) file name without extension a file name with the default extension (first from the default filter) is returned. When selecting an *existing* file without extension however the dialog will return that file name as is (without extension) making the behavior even more confusing for the user.

Running this sample on wxGTK or wxOSX however will show that no matter what, entering a file name without extension (existing or non-existing) will result in a file name without extension being returned even without a 'Any files (*.*)' filter present.

- wxWidgets port  you use:  wxMSW, wxGTK (3), wxOSX
- OS  and its version: Windows 11, OpenSuSE 15.5, macOS 14

mcorino:
fixes #24949

### DIFF
--- a/src/msw/filedlg.cpp
+++ b/src/msw/filedlg.cpp
@@ -1740,9 +1740,6 @@ int wxFileDialog::ShowIFileDialog(WXHWND hWndParent)
         }
         else // Single selected file is in m_path.
         {
-            // Append the extension if necessary.
-            m_path = AppendExtension(m_path, wildFilters[m_filterIndex]);
-
             const wxFileName fn(m_path);
             m_dir = fn.GetPath();
             m_fileName = fn.GetFullName();


doublemax:
I don't like this as it silently changes application behavior. I also don't think that using one Windows application as sample is sufficient. Could we at least have a flag that controls this? As a user of an application, i expect the default extension to get added when i don't enter one. As a user of wxWidgets, i would like the library to do that for me, otherwise i'll have to add the check all over the place, or subclass wxFileDialog to get the same result.

mcorino:
> I don't like this as it silently changes application behavior. I also don't think that using one Windows application as sample is sufficient. Could we at least have a flag that controls this? As a user of an application, i expect the default extension to get added when i don't enter one. As a user of wxWidgets, i would like the library to do that for me, otherwise i'll have to add the check all over the place, or subclass wxFileDialog to get the same result.

The changes are not based on the behavior of a single app. More were tested. Also, adding the default extension does not disappear with this change as the default Windows behavior already takes care of this when appropriate.
Before commenting further PLEASE read the referenced issue first.

vadz:
Thanks!

I will merge this soon, after just adding a comment saying that we don't need to muck with the extension in this case to be totally clear about what's going on here, e.g.

```cpp
// Note that we intentionally do not call AppendExtension() here because
// IFileDialog already does it if necessary and doing it again would be wrong,
// see the discussion in #24949.
```

but I wonder if we should backport it to 3.2 too? OT1H, it very clearly looks like a bug fix, but OTOH it does change the application behaviour and while I can't really imagine anyone depending on the current behaviour, our policy is to avoid making such changes in the stable branch. What do you think?

mcorino:
IMHO this is clearly a bugfix and I can't imagine any one/app really relying on the current behavior. But as you say the policy is to not make these behavior changes on a stable branch I can live with not backporting this.

PR #25037, opened 2024-12-31
Explain difference between wxGetTranslation and _() in internat sample

Blake-Madden:
Also, more reference to `wxT()` macro in overview topic

### DIFF
--- a/interface/wx/translation.h
+++ b/interface/wx/translation.h
@@ -581,10 +581,10 @@ public:
 
     This function is thread-safe.
 
-    @note This function is not suitable for literal strings using wxT() macro
+    @note This function is not suitable for literal strings
           since this macro is not recognised by @c xgettext, and so such
           strings are not extracted to the message catalog. Instead, use the
-          _() and wxPLURAL() macro for all literal strings.
+          _() and wxPLURAL() macros for all literal strings.
 
     @see wxGetTranslation(const wxString&, const wxString&, unsigned, const wxString&)
 
--- a/interface/wx/translation.h
+++ b/interface/wx/translation.h
@@ -581,12 +581,12 @@ public:
 
     This function is thread-safe.
 
-    @note This function is not suitable for literal strings
-          since this macro is not recognised by @c xgettext, and so such
-          strings are not extracted to the message catalog. Instead, use the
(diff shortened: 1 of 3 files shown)


vadz (changes requested):
Sorry, I don't think this is right. The old note was due to the fact that you couldn't use `wxGetTranslation(wxT("foo"))` even if you did pass `-kwxGetTranslation` to `xgettext` and removing `wxT()` from its text means that it loses its meaning.

We should just say that:

1. If you want literal strings inside `wxGetTranslation()` to be translated, you need to use `-kwxGetTranslation`.
2. You shouldn't use `wxT()` inside `wxGetTranslation()`.

interface/wx/translation.h
```diff
@@ -581,10 +581,10 @@ class wxMsgCatalog
...
     This function is thread-safe.
 
-    @note This function is not suitable for literal strings using wxT() macro
+    @note This function is not suitable for literal strings
           since this macro is not recognised by @c xgettext, and so such
           strings are not extracted to the message catalog. Instead, use the
-          _() and wxPLURAL() macro for all literal strings.
+          _() and wxPLURAL() macros for all literal strings.
 
     @see wxGetTranslation(const wxString&, const wxString&, unsigned, const wxString&)
```
PBfordev, line 589: BTW, this is not linked in [the generated docs](https://docs.wxwidgets.org/latest/group__group__funcmacro__string.html#ga49dcdc17d82a627b7a50d401f1a37d69) and I wonder if this could be fixed by using a correct signature (adding the context argument):
```suggestion
    @see wxGetTranslation(const wxString&, const wxString&, unsigned, const wxString&, const wxString &)
```
[The other override description](https://docs.wxwidgets.org/latest/group__group__funcmacro__string.html#ga3fca2c2319d2481568aef51678e8021e) suffers from the same problem, see the first line.

Blake-Madden: Thanks, PB. I saw a few more missing links, so I fixed those too.

Blake-Madden:
OK, all suggestions applied and fixed a few links also.

vadz:
Nice, thanks! Will merge in a moment.

PR #25040, opened 2025-01-01
Use PCRE2 function for counting the number of captures

vadz:
Don't try parsing the regex ourselves, which is not just slower but also more buggy: in particular, we didn't handle regexes with trailing backslash correctly and could keep reading after the end of the buffer in this case.

cc @oetr


vadz:
Applied to master, so closing. Will backport to 3.2 soon too.

PR #25042, opened 2025-01-01
3.2 backports 202501

vadz:


lanurmi:
I wonder if e2de7dc625d66cc1b3d336868a82c45ab8fa516f could be backported without breaking ABI. Would be useful.

vadz:
It can, if we make the exit code global — which should be fine, I think.

vadz:
I've accidentally pushed fixups for this branch into 3.2 instead of here, sorry for the confusion...

PR #25049, opened 2025-01-07
Use native GTK colours for wxInfoBarGeneric with GTK 3.24

### LINKED ISSUE #25048: wxInfoBar maybe should not take colors from ToolTip on GTK
### Description

On wxGTK, the wxInfoBar widget derives its default colors (`wxSYS_COLOUR_INFOTXT` and `wxSYS_COLOUR_INFOBK`) from a ToolTip widget.

On some popular Gnome themes such as Adwaita, the colors chosen for ToolTips are light text on dark background even in "light mode".

This leads to issues when wxInfoBar contains child widgets, for example wxHyperLink.  See https://gitlab.com/kicad/code/kicad/-/issues/19506

I'm not sure whether or not this issue should be fixed in wxWidgets (KiCad can hack around it, as noted in the discussion thread) but thought it was worth commenting on -- Gnome probably did not anticipate that ToolTips would ever have a need for different colors of text (or maybe any other child widgets), so it's possible that ToolTip is the wrong element to be sampling the infobar colors from.  I'm not sure what would be a better choice, though.

- OS  and its version: Linux (many)
  
  + GTK version: Any
  + Which GDK  backend is used: Any
  + Desktop environment : Gnome
  + Current theme: Adwaita

vadz:
This makes the generic implementation, which can be useful even in wxGTK, as it allows adding custom controls to the info bar, more similar to the native one in appearance and also avoids unfortunate clashes between the tooltip colours and the native colours of other controls, such as hyperlinks.

Closes #25048.

### DIFF
--- a/src/generic/infobar.cpp
+++ b/src/generic/infobar.cpp
@@ -36,6 +36,13 @@
 #include "wx/artprov.h"
 #include "wx/scopeguard.h"
 
+#ifdef __WXGTK3__
+    #include "wx/gtk/private/wrapgtk.h"
+
+    #include "wx/gtk/private/gtk3-compat.h"
+    #include "wx/gtk/private/stylecontext.h"
+#endif
+
 wxBEGIN_EVENT_TABLE(wxInfoBarGeneric, wxInfoBarBase)
     EVT_BUTTON(wxID_ANY, wxInfoBarGeneric::OnButton)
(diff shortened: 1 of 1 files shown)


vadz:
Applied to master and to the 3.2 backports branch (as well as a bunch of other `wxInfoBar`-related commits that this one turned out to depend on).

PR #25050, opened 2025-01-07
Document all font encodings

Blake-Madden:
Document all of them and make explanations more consistent.

Also, `wxFONTENCODING_MACVIATNAMESE` was misspelled. Replaced it with `wxFONTENCODING_MACVIETNAMESE` and kept `wxFONTENCODING_MACVIATNAMESE` as an alias for backward compatibility.

Fixed missing "!<" in doxygen comment.


vadz:
Thanks for cleaning this up!

Issue #24785, opened 2024-08-26
wxUILocale does not affect number/date formatting


PBfordev:
The first sentence in the wxUILocale docs reads "UI locale determines all culture-dependent conventions used in the user interface, including numbers, currencies and dates formatting". This made me think that `wxUILocale` is de-facto the modern drop-in replacement for `wxLocale` (translations aside), i.e., all wxWidgets methods will work as if `wxLocale` was used.

However, this does not seem the case, for example number and date formatting are still in the C locale even when `wxUILocale` was set to Czech, despite its `GetInfo()` returning  "," as the decimal separator and datetime format string being correct as well (tested Windows only).

`wxString` methods for converting floating point numbers (`{To|From}Double()`) or the `printf()`-like functions still use decimal point instead of comma; `wxDateTime::Format()` also ignores the locale setting.

When running the patched sample (see below) using `wxLocale`, all is as expected:
> Using wxLocale.
> Locale name: 'cs_CZ'
> Locale decimal separator: ','
> Datetime format: '%d.%m.%Y %H:%M:%S'
> wxNumberFormatter decimal separator: ','
> wxString::FromDouble(3.14): '3,14'
> Today is: '26.08.2024 17:29:29'
> wxString::ToDouble() with '3,14' succeeded: 3,14.
> wxNumberFormatter::FromString() with '3,14' succeeded: 3,14.

When using `wxUILocale` instead of `wxLocale`, only `wxNumberFormatter` works:
> Using wxUILocale.
> Locale name: 'cs-CZ'
> Locale decimal separator: ','
> Datetime format: '%d.%m.%Y %H:%M:%S'
> wxNumberFormatter decimal separator: ','
> **wxString::FromDouble(3.14): '3.14'
> Today is: 'Mon Aug 26 17:28:54 2024'
> Error: wxString::ToDouble() with '3,14' failed.**
> wxNumberFormatter::FromString() with '3,14' succeeded: **3.14**.

If this is a bug (hardly likely when manifesting in such common scenarios), it should be fixed. If not, I believe that the documentation should better explain how to use `wxUILocale` (e.g., emphasize that `wxUILocale` complements, not replaces `wxLocale`; and how the two may interact/conflict). Or, am I doing something wrong?

### How to reproduce
1. Patch (see below), build, and  run the minimal sample on a system where the user locale uses decimal comma, for example, any continental Europe language.
2. Observe the output in the log window.

```diff
[...]

PR #25051, opened 2025-01-07
Clarify the continued use of wxLocale

### LINKED ISSUE #25035: Clarifying the continued use of wxLocale::Init
Reading this overview:

https://docs.wxwidgets.org/latest/classwx_locale.html

and also the `wxUILocale` overview, I'm not 100% certain if you still need to call `wxLocale::Init()` on Windows to change the locale.

I read that on Linux, `wxUILocale::UseDefault()` will set the locale to the user's region for you (including number formatting). So `wxUILocale::UseDefault()` should be all you need, no need to call `wxLocale::Init()`.

On macOS, `wxUILocale::UseDefault()` will not set the locale. However, if the user's systems is > 12.2 , is it recommended to call `wxLocale::Init()` to get number formatting that the user expects?

And on Window, do you need to call `wxLocale::Init()` to set the locale to the user's region? Or does `wxUILocale::UseDefault()` do that for you?

When I read the help, I was under the impression that `wxUILocale::UseDefault()` is a full replacement for `wxLocale::Init()`, but now I'm not so sure. Thanks for any clarification.

Blake-Madden:
See #24785
Closes #25035

### DIFF
--- a/interface/wx/intl.h
+++ b/interface/wx/intl.h
@@ -287,7 +287,8 @@ enum wxLocaleForm
         11.0 and 12.2 inclusive are affected by a bug when changing C locale can
         break display of the application menus. Because of this, it is
         recommended to use wxUILocale instead of this class for the
-        applications targeting macOS.
+        applications targeting macOS. In general, this class should only be used
+        if requiring C runtime functions (@c strtod) to be localized.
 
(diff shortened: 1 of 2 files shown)


vadz:
Looks good now, thanks!

PR #25060, opened 2025-01-09
Don't use _() with variable, use wxGetTranslation

Blake-Madden:
If I understand correctly, only string literals should go into `_()`; anything else should go into `wxGetTranslation()`.


vadz:
Yes, indeed, thanks!

PR #25062, opened 2025-01-10
Add darkmode support to MSW ribbon, fixes to AUI dark mode ribbon

Blake-Madden:
# MSW Ribbon

MSW ribbon before would use the blue theme under dark mode. Now it will appear as this:

New MSW ribbon dark

# AUI Ribbon

## Before

AUI ribbon under dark mode looked like this:

Old AUI under dark

The tabs and hovered items would be a light gradient (and the text was black), making it difficult to see things.

## After

Now it will appear as such:

New AUI under dark

The ribbon demo has been updated to apply dark mode under MSW.

### DIFF
--- a/samples/ribbon/ribbondemo.cpp
+++ b/samples/ribbon/ribbondemo.cpp
@@ -195,10 +195,6 @@ bool MyApp::OnInit()
     if(!wxApp::OnInit())
         return false;
 
-#ifdef __WXMSW__
-    MSWEnableDarkMode();
-#endif
-
     wxFrame* frame = new MyFrame;
     frame->Show();
 
--- a/samples/ribbon/ribbondemo.cpp
+++ b/samples/ribbon/ribbondemo.cpp
@@ -195,6 +195,10 @@ bool MyApp::OnInit()
     if(!wxApp::OnInit())
         return false;
 
+#ifdef __WXMSW__
+    MSWEnableDarkMode();
+#endif
+
     wxFrame* frame = new MyFrame;
     frame->Show();
 
--- a/src/ribbon/art_aui.cpp
+++ b/src/ribbon/art_aui.cpp
@@ -252,11 +252,18 @@ void wxRibbonAUIArtProvider::SetColourScheme(
 #ifdef __WXOSX__
     m_tab_label_colour = wxSystemSettings::GetColour(wxSYS_COLOUR_CAPTIONTEXT);
 #else
-    m_tab_label_colour = LikePrimary(0.1);
+    if (wxSystemSettings::GetAppearance().IsDark())
+    {
+        m_tab_label_colour = wxSystemSettings::GetColour(wxSYS_COLOUR_CAPTIONTEXT);
+    }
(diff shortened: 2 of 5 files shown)


samples/ribbon/ribbondemo.cpp
```diff
@@ -195,6 +195,10 @@ bool MyApp::OnInit()
     if(!wxApp::OnInit())
         return false;
 
+#ifdef __WXMSW__
```
vadz, line 198: I'd rather omit this, we have a universal way of enabling dark mode for all samples by setting `wx_msw_dark_mode` environment variable and I'd like to have consistent behaviour for all of them.

src/ribbon/art_aui.cpp
```diff
@@ -252,11 +252,18 @@ void wxRibbonAUIArtProvider::SetColourScheme(
...
 #else
-    m_tab_label_colour = LikePrimary(0.1);
+    if (wxSystemSettings::GetAppearance().IsDark())
+    {
+        m_tab_label_colour = wxSystemSettings::GetColour(wxSYS_COLOUR_CAPTIONTEXT);
+    }
+    else
+    {
+        m_tab_label_colour = LikePrimary(0.1);
+    }
```
vadz, line 262: This can be written more briefly and, IMHO, nicely as

```suggestion
    m_tab_label_colour = wxSystemSettings::SelectLightDark()
        ? LikePrimary(0.1)
        : wxSystemSettings::GetColour(wxSYS_COLOUR_CAPTIONTEXT);
```

src/ribbon/art_aui.cpp
```diff
@@ -281,7 +288,8 @@ void wxRibbonAUIArtProvider::SetColourScheme(
     m_panel_label_background_colour = LikePrimary(0.85);
     m_panel_label_background_gradient_colour = LikePrimary(0.97);
     m_panel_hover_label_background_gradient_colour = secondary_hsl.ToRGB();
-    m_panel_hover_label_background_colour = secondary_hsl.Lighter(0.2f).ToRGB();
+    m_panel_hover_label_background_colour = wxSystemSettings::GetAppearance().IsDark() ?
```
vadz, line 291: I'd also use `SelectLightDark()` here.

vadz (commented):
Thanks, this is clearly a very nice improvement and I'll merge it but I think making `SelectLightDark()` is a bit nicer — please let me know if you disagree.

samples/ribbon/ribbondemo.cpp
Blake-Madden: Done.

src/ribbon/art_aui.cpp
Blake-Madden: Done.

vadz:
Thanks again, I'll merge this soon with some minor changes: I've added `wxRibbonHSLColour::AdjustLuminance()` which does the right thing depending on the light/dark mode and so makes the code using it a bit simpler.

PR #25063, opened 2025-01-10
Also accept wide string literals as translation macro arguments

lanurmi:
Not accepting them was an undeliberate regression introduced in PR #24957. There weren't any such calls in wx's own tests or samples, so now adding some to the internat sample to prevent a similar regression in the future.

[Pointed out](https://github.com/wxWidgets/wxWidgets/commit/3dddaeb8f2bb8fcfc07d44331a8a2224e24db34a#commitcomment-151169404) by @tobiolo.

### DIFF
--- a/samples/internat/internat.cpp
+++ b/samples/internat/internat.cpp
@@ -320,12 +320,6 @@ MyFrame::MyFrame()
     macro_menu->Append(INTERNAT_MACRO_8, wxGETTEXT_IN_CONTEXT_PLURAL("context_2", "sing", "plur", 1));
     macro_menu->Append(INTERNAT_MACRO_9, wxGETTEXT_IN_CONTEXT_PLURAL("context_2", "sing", "plur", 2));
 
-    // Also wide strings can be used:
-    _(L"item");
-    wxGETTEXT_IN_CONTEXT(L"context", L"item");
-    wxPLURAL(L"sing", L"plur", 3);
-    wxGETTEXT_IN_CONTEXT_PLURAL(L"context", L"sing", L"plur", 3);
-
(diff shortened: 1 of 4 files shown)


samples/internat/internat.cpp
```diff
@@ -320,6 +320,12 @@ MyFrame::MyFrame()
     macro_menu->Append(INTERNAT_MACRO_8, wxGETTEXT_IN_CONTEXT_PLURAL("context_2", "sing", "plur", 1));
     macro_menu->Append(INTERNAT_MACRO_9, wxGETTEXT_IN_CONTEXT_PLURAL("context_2", "sing", "plur", 2));
 
+    // Also wide strings can be used:
```
vadz, line 323: Shouldn't this rather go into `tests/allheaders.cpp` as the other checks you added there?

Or, perhaps, move all of these checks to `tests/intl/intltest.cpp`?

lanurmi: > Shouldn't this rather go into `tests/allheaders.cpp` as the other checks you added there?

`allheaders.cpp` is special, because it defines `wxNO_IMPLICIT_WXSTRING_ENCODING`, and therefore doesn't test the regular use case. I also hypothesized that wide strings are inherently incompatible with `wxASCII_STR`, which the wrappers call in that case, but didn't test what actually happens.

> Or, perhaps, move all of these checks to `tests/intl/intltest.cpp`?

Yes, the new ones I added to the sample can be moved.

lanurmi: When you put it that way, yes. I suppose this PR is complete now, after your latest changes.

PR #25064, opened 2025-01-11
Fix tests compilation with Clang 19.1.6 on Fedora Rawhide

lanurmi:
There was an error about an incompatible cast:
```
In file included from ../../tests/allheaders.cpp:369: In file included from ../../tests/testprec.h:5:
In file included from ../../include/wx/evtloop.h:13: ../../include/wx/event.h:141:12: error: cast from 'void (wxEvtHandler::*)(wxFocusEvent &)' to 'wxEventFunction' (aka 'void (wxEvtHandler::*)(wxEvent &)') converts to incompatible function type [-Werror,-Wcast-function-type-mismatch]
  141 |     return reinterpret_cast<wxEventFunction>(func);
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
This warning/error was already suppressed on GCC; just enable the same `wxGCC_WARNING_SUPPRESS_CAST_FUNCTION_TYPE()` macro for Clang also.

### DIFF
--- a/include/wx/defs.h
+++ b/include/wx/defs.h
@@ -669,7 +669,7 @@ typedef short int WXTYPE;
 #endif
 
 /* Specific macros for -Wcast-function-type warning new in gcc 8. */
-#if wxCHECK_GCC_VERSION(8, 0)
+#if wxCHECK_GCC_VERSION(8, 0) || defined(__clang__)
     #define wxGCC_WARNING_SUPPRESS_CAST_FUNCTION_TYPE() \
         wxGCC_WARNING_SUPPRESS(cast-function-type)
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, but as [the error](https://github.com/wxWidgets/wxWidgets/actions/runs/12724170727/job/35470053038?pr=25064) shows, we need to use `wxCLANG_WARNING_SUPPRESS()` which checks that the warning is valid before suppressing it.

vadz:
Thanks! Merged into master, will cherry pick to 3.2 too.

PR #25068, opened 2025-01-11
MSW console output fixes

vadz:
I've decided to drastically simplify/remove most of the code dealing with console output, as I don't think it's justified to have it for something that can be done much simpler. It also seems like this code has never been actually used for anything as there was a bug which prevented it from working in non-console applications, so we're not losing much.

On the bright side, using `stderr` works now if we have an associated console.

### DIFF
--- a/src/common/msgout.cpp
+++ b/src/common/msgout.cpp
@@ -167,6 +167,10 @@ void wxMessageOutputDebug::Output(const wxString& str)
     out.Replace(wxT("\t"), wxT("        "));
     out.Replace(wxT("\n"), wxT("\r\n"));
     ::OutputDebugString(out.t_str());
+
+    // If we have a console under Windows, send this there too (if not, this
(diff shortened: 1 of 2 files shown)


marekr:
Seems ok to me except for should "\t" still be getting replaced with spaces even for the normal stderr?

vadz:
> Seems ok to me except for should "\t" still be getting replaced with spaces even for the normal stderr?

I don't see why... Printing `\t` to stderr should work fine and, again, if we're trying to be consistent with `fprintf(stderr)` here (which I think we are), then it makes sense to avoid gratuitous deviations from what it does.

marekr:
Ah nvm I didn't pay attention to variable names clearly.

vadz:
Ah, yes, I should have said that I pass `str`, and not `out`, to `wxMessageOutputStderr` intentionally here.

Anyhow, I'll merge this soon if it works for you.

PR #25076, opened 2025-01-13
Add support for multiple tabs rows in `wxAuiNotebook`

vadz:
The main commit here is the last one, which looks like this (when enabled in the sample menu and after rearranging some tabs by dragging them around) under Windows:

image

And Linux:

image

Note that while MSW look is not too bad IMO, the look with the default (generic) AUI tab art is not great because it reserves some space between tab rows. Unfortunately it doesn't seem simple to modify it to avoid this, so I wonder if we should provide some kind of "flat" theme for AUI that would look better with this style.


vadz:
I'm going to merge this soon if there are no comments/objections.

PR #25091, opened 2025-01-17
make samples/ownerdrw sample buildable for wxUniv

Wlodzimierz-ABX-Skiba:
ownerdrw is wxMSW specific but wxMSW has makefiles which allow wxUniv variant. This little change allows building it for wxUniv so the global makefile for samples does not stop on this sample


vadz:
Thanks, will merge this one too soon.

PR #25093, opened 2025-01-18
don't draw mnemonics in controls in wxUniv

Wlodzimierz-ABX-Skiba:
I don't know when and where the regression happened but in the past mnemonics weren't drawn in wxUniv builds. This patch fixes it.

Before patch it looks as follow: 
wxUniv_broken_mnemonics

After patch if looks as follow:
wxUniv_fixed_mnemonics

### DIFF
--- a/include/wx/univ/control.h
+++ b/include/wx/univ/control.h
@@ -64,8 +64,11 @@ public:
     // accelerator char (the one immediately after '&') into m_chAccel
     virtual void SetLabel(const wxString& label) override;
 
-    // return the current label
-    virtual wxString GetLabel() const override { return m_label; }
+    // return the current label with mnemonics
+    virtual wxString GetLabel() const override { return wxControlBase::GetLabel(); }
+
+    // return the current label without mnemonics
+    virtual wxString GetLabelText() const override { return m_label; }
 
     // wxUniversal-specific methods
 
--- a/include/wx/univ/control.h
+++ b/include/wx/univ/control.h
@@ -65,7 +65,7 @@ public:
     virtual void SetLabel(const wxString& label) override;
 
     // return the current label
-    virtual wxString GetLabel() const override { return wxControlBase::GetLabel(); }
+    virtual wxString GetLabel() const override { return m_label; }
 
     // wxUniversal-specific methods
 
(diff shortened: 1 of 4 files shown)


vadz:
I'm not sure about this one, it seems wrong to pass the label containing mnemonics to a function taking `indexAccel` in the first place. I.e. we should work with _either_ strings with "&" _or_ with pairs of string-without-"&" and accel index, but here we're mixing them up.

IOW I think we need to modify `wxControlRenderer::DrawButtonLabel()` and similar functions instead to pass `GetLabelText()` to the renderer functions, what do you think?

BTW, we could also override `GetLabelText()` in wxUniv to just return `m_label` as a micro-optimization.

Wlodzimierz-ABX-Skiba:
wxControlBase::SetLabelText does EscapeMnemonics() so internally all & are && and later in drawing && are restored to & and that's why it was wrongly drawn. Microoptimisation in GetLabelText didn't work, only GetLabel does the trick

Wlodzimierz-ABX-Skiba:
Any idea why wxX11 build failed on all headers test? I did change header but wxMSWUniv build went fine

vadz:
Looks like a problem on GitHub Actions, the PCH file is not supposed to disappear like this :-( I've relaunched the job.

include/wx/univ/control.h
```diff
@@ -65,7 +65,7 @@ class WXDLLIMPEXP_CORE wxControl : public wxControlBase, public wxInputConsumer
     virtual void SetLabel(const wxString& label) override;
 
     // return the current label
-    virtual wxString GetLabel() const override { return wxControlBase::GetLabel(); }
+    virtual wxString GetLabel() const override { return m_label; }
```
vadz, line 68: Sorry, but this is wrong. `GetLabel()` should return the label _with_ mnemonics (i.e. `&Foo`), but `m_label` shouldn't contain them, i.e. it should be returned by `GetLabelText()`.

This should probably break a number of unit tests too — except that we don't even run them for wxUniv, I think, because so many of them fail...

Wlodzimierz-ABX-Skiba:
Thanks, the problem was in renderer which used GetLabel instead of GetLabelText

vadz:
Looks good now, thanks, will merge soon.

PR #25094, opened 2025-01-18
Refresh control when only position of mnemonic in label is changed

Wlodzimierz-ABX-Skiba:
The problem is visible in samples/widgets sample. In this sample, in "Button" page, on the most right there is text field with "&Press me!" text and button presenting it. When you change "&Press me!" to "Press &me!" and click "Change label" button, the button with this label is not refreshed. This patch fixes it.


vadz:

PR #25095, opened 2025-01-18
Disallow dragging wxAuiNotebook tab if BEGIN_DRAG was vetoed

### LINKED ISSUE #25074: Allow forbidding dragging individual wxAuiNotebook tabs
> (BTW, can locked tabs still be moved to another position in the "locked tabs" group or should even this be forbidden?).

We would like to forbid this.  Might be nice to make this configurable in case some other users don't want to forbid it.

_Originally posted by @craftyjon in https://github.com/wxWidgets/wxWidgets/issues/25072#issuecomment-2585795475_

vadz:
Let the application forbid dragging some of the tabs.

Closes #25074.


vadz:
As discussed in #25074, this is not actually useful, so closing.

PR #25102, opened 2025-01-20
[RFC] Replace `wxLog::SetActiveTarget(new ...)` with slightly safer alternative.

DoctorNoobingstoneIPresume:
In order to prevent a rare possibility of object leak (given that `SetActiveTarget` might fail with an exception, especially since it calls a virtual function which might do who-knows-what),

I wish to suggest that we suggest to users to replace:

```
wxLog::SetActiveTarget(new wxLogDerivedClass{...});
```

with the uglier:

```
std::unique_ptr<wxLog> spNew{stD::make_unique<wxLogDerivedClass(...)};
wxLog.SetActiveTarget(spNew.get());
spNew.release();
```

Also, I wonder whether `ms_pLogger` should eventually be deleted.

Currently, this pull request is a request for comments/feedback. Many thanks, and Happy New Year, Happy New Year...

### DIFF
--- a/samples/xrc/objrefdlg.cpp
+++ b/samples/xrc/objrefdlg.cpp
@@ -32,7 +32,11 @@
 
 #include "wx/xrc/xmlres.h"              // XRC XML resouces
 
+//-----------------------------------------------------------------------------
+// For unique_ptr etc.
+//-----------------------------------------------------------------------------
 
+#include <memory>
 
 //-----------------------------------------------------------------------------
 // Public members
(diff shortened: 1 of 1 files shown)


vadz:
Sorry, I don't think it's a good idea. `Flush()` is not supposed to throw, but if it's a real problem, `wxLog::SetActiveTarget()` should wrap it in `try/catch` or maybe just call it after changing `ms_pLogger`.

It would be nice to provide `SetActiveTarget()` overload taking `unique_ptr<wxLog>`, to clearly indicate that it takes ownership of it, but there are many similar functions in wx, taking a pointer and taking ownership of it, and changing just one of them while leaving the others is going to be inconsistent, while changing all of them (and doing it in backwards compatible way) is probably too difficult.

`ms_pLogger` is deleted on program exit/library shutdown.

Finally, it would be preferable to use wx-dev for discussions, rather than do it here. Please post there if you'd like to discuss this more. TIA!

vadz:
Normally in `wxAppBase::CleanUp()` when `delete wxLog::SetActiveTarget(nullptr);` is called.

PR #25103, opened 2025-01-20
Add macOS support for wxTextCtrl::SearchText()

Blake-Madden:
Also mention that it is supported under wxQT in the docs.


vadz:
Thanks, this looks good to me but it seems that the code can be simplified just a bit by avoiding the test for `found.length` (AFAICS it can never be 0 if location is valid) and using earlier returns to avoid too much nesting — please see my commits.

If you don't have any objections, I'll (squash) merge this soon.

Thanks again!

Blake-Madden:
@vadz Sounds great, thanks. I tend to be overly redundant with my condition checking, so by all means simplify away.

PR #25107, opened 2025-01-21
placement of check mark repaired in wxUnivGTK theme

Wlodzimierz-ABX-Skiba:
I noticed placement of check marks was broken with wxUnivGTK theme. Before this fix it looked like: 
wxUniv_broken_check_menu_icon
and afer this change it looks like: 
wxUniv_fixed_check_menu_icon


vadz:

PR #25109, opened 2025-01-21
wxUniv accel support for controls

Wlodzimierz-ABX-Skiba:
This change adds accel support for mnemonics used in control labels. Previously accel chars were displayed but didn't work. You can see result of this PR in menu or minimal samples. Just build and run sample, press F1 to invoke menu accel (worked fine) and open about dialog, then press O to invoke control accel (implemented in this PR) and cause OK button event.

### DIFF
--- a/include/wx/univ/control.h
+++ b/include/wx/univ/control.h
@@ -31,6 +31,7 @@ typedef wxString wxControlAction;
 // in the controls headers)
 
 #define wxACTION_NONE    wxT("")           // no action to perform
+#define wxEMPTY_ACCEL_CHAR    wxT('\0')
 
 // ----------------------------------------------------------------------------
 // wxControl: the base class for all GUI controls
@@ -52,6 +53,8 @@ public:
 
         Create(parent, id, pos, size, style, validator, name);
     }
+
+    virtual ~wxControl();
 
     bool Create(wxWindow *parent,
                 wxWindowID id,
@@ -78,7 +81,7 @@ public:
     // return the accel char itself or 0 if none
     wxChar GetAccelChar() const
     {
-        return m_indexAccel == -1 ? wxT('\0') : (wxChar)m_label[m_indexAccel];
+        return m_indexAccel == -1 ? wxEMPTY_ACCEL_CHAR : (wxChar)m_label[m_indexAccel];
     }
 
     virtual wxWindow *GetInputWindow() const override
--- a/include/wx/univ/control.h
+++ b/include/wx/univ/control.h
@@ -31,7 +31,7 @@ typedef wxString wxControlAction;
 // in the controls headers)
 
 #define wxACTION_NONE    wxT("")           // no action to perform
-#define wxEMPTY_ACCEL_CHAR    wxT('\0')
+#define wxNO_ACCEL_CHAR    wxT('\0')
 
 // ----------------------------------------------------------------------------
 // wxControl: the base class for all GUI controls
@@ -81,7 +81,7 @@ public:
     // return the accel char itself or 0 if none
     wxChar GetAccelChar() const
     {
-        return m_indexAccel == -1 ? wxEMPTY_ACCEL_CHAR : (wxChar)m_label[m_indexAccel];
+        return m_indexAccel == -1 ? wxNO_ACCEL_CHAR : (wxChar)m_label[m_indexAccel];
     }
 
     virtual wxWindow *GetInputWindow() const override
--- a/src/univ/control.cpp
+++ b/src/univ/control.cpp
@@ -74,15 +74,15 @@ wxControl::~wxControl()
 #if wxUSE_ACCEL
     wxChar accelChar = GetAccelChar();
 
-    for ( wxWindow *win = this; win; win = win->GetParent() )
+    if ( accelChar != wxEMPTY_ACCEL_CHAR )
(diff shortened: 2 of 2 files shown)


src/univ/control.cpp
```diff
@@ -85,9 +104,37 @@ void wxControl::UnivDoSetLabel(const wxString& label)
...
+            {
+                if ( accelCharOld != wxEMPTY_ACCEL_CHAR )
+                {
+                    wxAcceleratorEntry accelEntryOld(wxACCEL_NORMAL, (int)accelCharOld);
+                    win->GetAcceleratorTable()->Remove(accelEntryOld);
+                }
+
+                if ( accelChar != wxEMPTY_ACCEL_CHAR )
+                {
+                    wxAcceleratorEntry accelEntryNew(wxACCEL_NORMAL, (int)accelChar, GetId());
```
vadz, line 130: Shouldn't this be
```suggestion
                    wxAcceleratorEntry accelEntryNew(wxACCEL_ALT, (int)accelChar, GetId());
```
? Mnemonics are supposed to be activated using `Alt-<Letter>`, not just the letter.

src/univ/control.cpp
```diff
@@ -85,9 +104,37 @@ void wxControl::UnivDoSetLabel(const wxString& label)
...
     m_indexAccel = FindAccelIndex(label, &m_label);
 
+#if wxUSE_ACCEL
+    wxChar accelChar = GetAccelChar();
+
+    if ( accelCharOld != accelChar )
+    {
+        for ( wxWindow *win = this; win; win = win->GetParent() )
+        {
+            if ( win->IsTopLevel() )
```
vadz, line 120: This could be replaced with a call to `wxGetTopLevelParent()`.

vadz (changes requested):
Thanks, but this doesn't seem quite right unless I'm missing something.

src/univ/control.cpp
Wlodzimierz-ABX-Skiba: I used wxACCEL_NORMAL because first I investigated wxMSW builds and noticed there that pressing key without ALT worked fine. But now when I look into these samples I cannot find this case. I can live with wxACCEL_ALT.

src/univ/control.cpp
Wlodzimierz-ABX-Skiba: Thank you, somehow I forgot that I used it in the past. Will refactor.

src/univ/control.cpp
```diff
@@ -69,6 +69,25 @@ bool wxControl::Create(wxWindow *parent,
...
+    wxChar accelChar = GetAccelChar();
+
+    if ( accelChar != wxEMPTY_ACCEL_CHAR )
+    {
+        wxWindow *win = wxGetTopLevelParent(this);
+        if ( win )
+        {
+            wxAcceleratorEntry accelEntry(wxACCEL_ALT, (int)accelChar);
+            wxAcceleratorTable *accelTable = win->GetAcceleratorTable();
+            if ( accelTable && accelTable->IsOk() )
```
vadz, line 84: Why do we test for this here but not when adding accelerators below? It looks like it we should either test in both places or not need to test in either.

src/univ/control.cpp
```diff
@@ -69,6 +69,25 @@ bool wxControl::Create(wxWindow *parent,
...
 
+wxControl::~wxControl()
+{
+#if wxUSE_ACCEL
+    wxChar accelChar = GetAccelChar();
+
+    if ( accelChar != wxEMPTY_ACCEL_CHAR )
+    {
+        wxWindow *win = wxGetTopLevelParent(this);
+        if ( win )
```
vadz, line 80: Pretty minor but I think writing it like this:

```suggestion
        if ( wxWindow* win = wxGetTopLevelParent(this) )
```

is more clear (and one line less).

include/wx/univ/control.h
```diff
@@ -31,6 +31,7 @@ typedef wxString wxControlAction;
 // in the controls headers)
 
 #define wxACTION_NONE    wxT("")           // no action to perform
+#define wxEMPTY_ACCEL_CHAR    wxT('\0')
```
vadz, line 34: Extremely minor, but for consistency with the above and general clarity I'd call it `wxNO_ACCEL_CHAR`.

vadz (commented):
Thanks, looks good now, but I'm a bit worried by the absence of checks for the accel table, do we really not need them, i.e. can we be certain that it's never null? And, if so, why do we still need to test for this in the dtor?

src/univ/control.cpp
Wlodzimierz-ABX-Skiba: Good point! I added the test because when I switched to wxGetTopLevelParent it started to assert probablu due to calling it from dtor. I'll add it below as well to be on safe side.

src/univ/control.cpp
Wlodzimierz-ABX-Skiba: Sorry, that's lack of C++ over 20 years. Will change.

include/wx/univ/control.h
Wlodzimierz-ABX-Skiba: Fine for me, will change.

vadz:
Looks good now, thanks! Will merge soon.

PR #25110, opened 2025-01-22
Correct German translation of "Ctrl" in internat sample

taler21:
Should be backported to 3.2 as well.


lanurmi:
There's the slight downside that after this this correction, the shortcut fails to work properly, i.e. it silently changes to just `T` (without modifiers), unless `wxstd.mo` is available – and it isn't available by default in the sample.

Before this correction, the shortcut worked with and without `wxstd.mo` present, and yet was properly shown as `Strg-T` in the menu, despite being `Ctrl-T` in the translation.

PR #25111, opened 2025-01-22
Add international currency symbol support to locale classes

Blake-Madden:
Adds `wxLOCALE_INTERNATIONAL_CURRENCY_SYMBOL` constant to `wxLocaleInfo`.
Adds support to `wxUILocale` and `wxLocale::GetInfo` for this constant.


Blake-Madden:
These symbols are just too spotty on UNIX and macOS.

PR #25112, opened 2025-01-22
Update submodules

MaartenBent:
Test updating expat, jpeg, png and zlib submodules. I'll try to add some more.


MaartenBent:
Before merging this, maybe we can also add a libwwebp submodule?
I was testing #24479 and added support for a builtin (CMake) version.

vadz (approved):

I'd prefer to merge this already and deal with libwebp later in a separate commit, but if you really want to include it in this one, please let me know.

MaartenBent:
Fixed the copy-paste mistake in the instructions, and added a line about updating the Xcode projects.

Lets merge this now then and do the webp stuff some other time.

PR #25118, opened 2025-01-24
Add missing wxTextCtrl styles to its XRC handler

sodevel:
A wxFormBuilder user discovered that the wxTextCtrl XRC handler does not support the style `wxTE_CENTER` (it only supports the `wxTE_CENTRE` variant). Although the documentation does not mention the `wxTE_CENTER` variant it does in fact exist in the code, so i added this style to the XRC handler. I found two additional missing styles and added these as well. I did not update the documentation because i have no idea how to do this.


vadz:
Thanks, will apply soon.

PR #25120, opened 2025-01-25
Add translator comments

Blake-Madden:
Adds some (IMO) helpful comments to explain how to translate various strings.


vadz:
Thanks, I'll apply this, but I wonder what should be the rule about using context vs using special comments, in general, as it's not really clear to me.

Any thoughts?

Blake-Madden:
My thoughts are:

If the string is a readable message (e.g., "Help"), but one may not be sure where it is used, then context. In the example of something like "Help", a "menu" context.

If the string is an unfamiliar acronym, contains a lot of `printf` commands, or is a strange looking token ("*)"), then a descriptive comment is warranted. That's what I has looking for in this particular PR.

PR #25121, opened 2025-01-26
bind event handlers before creating window

wsu-cb:
Reorder operations so that `wxEVT_ACTIVATE` handler gets called when `wxDocChildFrameAny<>`'s `ChildFrame` is created.  (See https://groups.google.com/g/wx-dev/c/XOLR2mD_OO8/m/hGSMGQtRCAAJ)


vadz:

Issue #25105, opened 2025-01-21
Failure in BitmapComboBoxTestCase


Wlodzimierz-ABX-Skiba:
#### Bug description:

(I'm trying to setup building tests for the first time so please forgive me if I'm doing something wrong or miss something. This is my first report of gui failures I found to check if I'm reporting it right way)

I'm getting this in BitmapComboBoxTestCase

```
controls/bitmapcomboboxtest.cpp:100: FAILED:
  REQUIRE( wxSize(16, 16) == m_combo->GetBitmapSize() )
with expansion:
  16*16 == 17*17
```

(I'm also getting build error in allheaders test build but I'm unsure if it is expected behaviour)

#### Expected vs observed behaviour:

I expect m_combo->GetBitmapSize() to return `16*16`

#### To Reproduce:

just build and run gcc_mswud\test_gui.exe BitmapComboBoxTestCase

### Platform and version information

- wxWidgets version you use: fresh master (3.3)
- wxWidgets port you use: wxMSW with makefile.gcc BUILD=debug SHARED=0 MONOLITHIC=0 USE_GUI=1 with fresh MinGW installation
- OS and its version: Windows 10 (build 19045)

vadz:
This is weird, the test passes in the CI jobs and for me locally, so there must be something different about your system which makes it fail, but I can't see what.

AFAICS `m_usedImgSize`, which is returned by `GetBitmapSize()`, is only set in `wxBitmapComboBoxBase::UpdateInternals()`, can you check how does it manage to return 17x17 for you?

Wlodzimierz-ABX-Skiba:
@vadz No, it is not the only place where it is set. It is also set in `wxBitmapComboBoxBase::OnAddBitmap` and debugging it I found that 17 comes from `wxBitmapBundleImpl::DoGetPreferredSize` in src/common/bmpbndl.cpp where in last line `scaleBest=1.5` and `GetDefaultSize().x=11`
That 1.5 probably comes from the fact that I have Windows display set in settings to 150%

vadz:
Ah, I see, thanks.

I'm not sure how to fix this. Using just `FromDIP()` around sizes in the test should do it, but I'm not sure if there is no problem in `wxBitmapComboBox` itself... Still, if you can confirm that adding `FromDIP()`s fixes it for you, we could at least do this for now.

Wlodzimierz-ABX-Skiba:
By "around sizes in the test" you mean to change assert to:
[...]

PR #25123, opened 2025-01-26
FromDIP fix in BitmapComboBoxTestCase

Wlodzimierz-ABX-Skiba:
per discussion in Issue #25105
this fixes assert when screen zoom is like 150%


vadz:
Thanks, I'll apply it (or, rather, commit and push my own stashed change) soon.

PR #25124, opened 2025-01-27
FromDIP fix in PropertyGridTestCase

Wlodzimierz-ABX-Skiba:
like in PR #25123 but for PropertyGridTestCase

### DIFF
--- a/tests/controls/propgridtest.cpp
+++ b/tests/controls/propgridtest.cpp
@@ -1461,7 +1461,7 @@ TEST_CASE("PropertyGridTestCase", "[propgrid]")
     SECTION("SetSplitterPosition")
     {
 #ifndef __WXQT__
-        const int trySplitterPos = 50;
+        const int trySplitterPos = wxTheApp->GetTopWindow()->FromDIP(50);
 #else
         const int trySplitterPos = 51; // FIXME!
 #endif


vadz:
Sorry, this doesn't seem right... In the other PR there was conversion between different (logical/physical) sizes, but here `SetSplitterPosition()` and `GetSplitterPosition()` really must be operating with the same position (expressed in logical pixels), so even if using `FromDIP()` would be more correct, the comparison should work no matter what.

Which check fails for you and with which values?

Wlodzimierz-ABX-Skiba:
@vadz it fals as follow when I have zoom of screen set to 150%

PropertyGridTestCase
  SetSplitterPosition
controls/propgridtest.cpp:1461

controls/propgridtest.cpp:1473: FAILED:
  CHECK( pgManager->GetGrid()->GetSplitterPosition() == trySplitterPos )
with expansion:
  67 == 50

controls/propgridtest.cpp:1481: FAILED:
  CHECK( pgManager->GetGrid()->GetSplitterPosition() == trySplitterPos )
with expansion:
  67 == 50

vadz:
This is definitely wrong, somehow it scales the position by the DPI, but it shouldn't.

But I still have no idea why does it scale 50 but doesn't scale 67 (when you pass it to it in your patch).

Anyhow, this certainly needs to be debugged, but I'm quite sure this PR is not correct and shouldn't be applied, sorry, so I'm closing it.

Wlodzimierz-ABX-Skiba:
@vadz I think neither 50 nor 67 is just simple single value. First of all test assumes thet setting 50 should result in 50 but in wxPropertyGridPageState::DoSetSplitter you have clear comment stating "Actual adjustment can be different from demanded." so if we set splitter to 50 like in test it doesn't mean it will be set to 50. It is computed in wxPropertyGridPageState::DoGetSplitterPosition. For screen zoom set to 100% in this method splitter position is summary of 15+35 (and meets expected 50). But for screen set to 150% it is summary of 22+45=67.  This 67 is a minimal possible value for splitter position when screen is set to 150%. I verified it debugging samples/propgrid sample. So in test there is wrong assumption. I believe that my PR was the simplest solution to this wrong assumption but meybe there is better. I'm open to any suggestions.

vadz:
Oh, I see, thanks for the explanation. So actually this 50 shouldn't ideally be hardcoded at all and instead computed from the column widths somehow? But I guess that this might be not that simple and using `FromDIP()` with the hardcoded value is better than nothing, so I'll merge this soon then.

PR #25126, opened 2025-01-29
#include all headers in tests/allheaders.h

lanurmi:
Despite its description, tests/allheaders.h did not `#include` all current headers.

`<wx/webview_chromium_impl.h>` is left out as excluded.


vadz:
Nice, thanks!

Ideally we should do it for (some) subdirectories too, e.g. I forgot to make the recently added `include/wx/aui/serializer.h` self-contained (I'll push a fix for this soon) because AUI headers are not checked by this test.

Issue #25131, opened 2025-01-31
mac: SetCursor works but cursor is reset to default on next mouse event


unawarez:
#### Bug description:

On macOS, wxWindow::SetCursor to a non-default cursor works, but only until the first motion, click, or scroll (so I assume any) mouse event, and then it resets back to the default cursor. I tried a couple different wxStockCursor values and also the static wxHOURGLASS_CURSOR but they all behaved the same.

#### Expected vs observed behaviour:

New cursor should remain until the next SetCursor. Instead it is reset after the next mouse event.

#### Patch or snippet allowing to reproduce the problem:

```c++
#include "wx/wx.h"

class MyApp : public wxApp {
  public:
    virtual bool OnInit() override;
};

class MyFrame : public wxFrame {
  public:
    MyFrame(const wxString &title);
    void OnClick(wxMouseEvent &event);
    void OnQuit(wxCommandEvent &event);

  private:
    wxDECLARE_EVENT_TABLE();
    bool cursor_hidden{false};
    wxStaticText *hidden_label{nullptr};
};

wxBEGIN_EVENT_TABLE(MyFrame, wxFrame) wxEND_EVENT_TABLE()
    wxIMPLEMENT_APP(MyApp);

bool MyApp::OnInit() {
    if (!wxApp::OnInit())
        return false;
    MyFrame *frame = new MyFrame("Minimal wxWidgets App");
    frame->Show(true);
    return true;
}

MyFrame::MyFrame(const wxString &title) : wxFrame(nullptr, wxID_ANY, title) {
    wxSizer* sizer = new wxBoxSizer(wxVERTICAL);
    sizer->Add(new wxStaticText(this, wxID_ANY, "Cursor hiding (left click to toggle):"));
    hidden_label = new wxStaticText(this, wxID_ANY, "normal");
    sizer->Add(hidden_label);
    SetSizer(sizer);

    Bind(wxEVT_LEFT_UP, &MyFrame::OnClick, this);
}

void MyFrame::OnClick(wxMouseEvent &evt) {
    cursor_hidden = !cursor_hidden;
    if (cursor_hidden) {
        SetCursor(wxCursor(wxCURSOR_CROSS));
        // same behavior: SetCursor(wxCursor(wxCURSOR_BLANK));
        // same behavior: SetCursor(*wxHOURGLASS_CURSOR);
    } else {
        SetCursor(wxNullCursor);
    }
[...]

PR #25134, opened 2025-02-01
wxOSX: fix cursor getting reset to default on next mouse event

unawarez:
Fix for issue #25131 where custom cursors get reset to default on the next mouse event.

The first line of wxWidgetCocoaImpl::DoHandleMouseEvent is to always call wxWidgetCocoaImpl::SetupCursor. Ultimately this goes to wxWindow::MacSetupCursor. It looks like the final `if` was always overwriting the choice of m_cursor with wxSTANDARD_CURSOR. Changing to `else if` fixes the issue for me. Can't say this is the perfect fix because there is a lot of logic happening before this point that isn't totally clear to me.

This also shows that the cursor set logic happens on every mouse event and thus also posts a wxSetCursor event to the app on every mouse event. This seems suboptimal to me but is way beyond my ability to try and change.

### DIFF
--- a/src/osx/window_osx.cpp
+++ b/src/osx/window_osx.cpp
@@ -1798,11 +1798,11 @@ bool wxWindowMac::MacSetupCursor( const wxPoint& pt )
             // if the user code caught EVT_SET_CURSOR() and returned nothing from
             // it - this is a way to say that our cursor shouldn't be used for this
             // point
-            if ( !processedEvtSetCursor && m_cursor.IsOk() )
+            if ( !processedEvtSetCursor && m_cursor.IsOk() ) {
                 cursor = m_cursor ;
-
-            if ( !wxIsBusy() && !GetParent() )
+            } else if ( !wxIsBusy() && !GetParent() ) {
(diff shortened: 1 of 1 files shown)


vadz:
Thanks! I agree that the existing logic seems wrong, which means I was also wrong when I wrote that it used to work before — because this code didn't change since 2008...

Actually I don't understand why do we have windows without parent specially at all here neither.

@csomor Am I missing something here or could we just remove the lines

```cpp
            if ( !wxIsBusy() && !GetParent() )
                cursor = *wxSTANDARD_CURSOR ;
```

completely?

vadz:
@csomor As I haven't heard back from you, I'm going to apply this but not backport to 3.2 for now. Please let me know if you can confirm that this is safe to backport and/or can be simplified further. TIA!

csomor:
> @csomor As I haven't heard back from you, I'm going to apply this but not backport to 3.2 for now. Please let me know if you can confirm that this is safe to backport and/or can be simplified further. TIA!

@vadz sorry, yes, I've marked the issue but couldn't yet think it through

PR #25098, opened 2025-01-19
CMake: Fix using static runtime with msvc

MaartenBent:
This was broken with recent CMake versions, see wx-dev discussion https://groups.google.com/g/wx-dev/c/0ECwYNf8TfU

PR #25139, opened 2025-02-04
3.2 backports 202502

vadz:


MaartenBent:
Can you backport #25098 (and the fix-up commit), or add the backport tag? You asked about it, but it is not yet included.
Looking at my previous PRs, I think #24635 can be backported as well.

vadz:
> Can you backport #25098 (and the fix-up commit), or add the backport tag? You asked about it, but it is not yet included. Looking at my previous PRs, I think #24635 can be backported as well.

Both done now, thanks! There was a conflict, but I think I've resolved it correctly ­— please let me know if not.

MaartenBent:
Thanks, looks fine. Seems you added `CMP0079` to the policies when resolving the conflict, but that shouldn't hurt anything (it was needed for libcef_dll_wrapper in wx3.3).

PR #25140, opened 2025-02-05
Mark property (in property grid) dirty when SetPropertyValue() is called

Blake-Madden:
With a property grid, when I call `SetPropertyValue()` programmatically, `IsPropertyModified()` currently returns false when called on the property whose value I just changed. IMO, the modified flag should be getting set anytime the various `DoSetPropertyValue...` functions get called.


imciner2:
Would this end up marking the items in the grid as dirty when the grid is initialized programmatically (such as when adding a `wxPGProperty` that has its value set)? If so, that seems less than ideal, since it would then mean the entire grid is dirty from the start.

Blake-Madden:
That's a good point, thanks. I'll submit another PR documenting this instead.

PR #25144, opened 2025-02-07
for dialog editors, provide concrete derivation of wxVListBox

wsu-cb:
Dialog editors need to create instances of controls for users to manipulate, but wxVListBox is an abstract class.  So provide wxXRCPreviewVListBox to support Dialog editors, and modify wxVListBoxXmlHandler to return wxXRCPreviewVListBox when in wxXRC_NO_SUBCLASSING mode.


vadz:

Issue #23717, opened 2023-07-19
Add support for `wxPowerEvent` generation under Linux


vadz:
It would be nice to have `wxEVT_POWER_XXX` events under Linux and it looks like D-Bus provides the signals we need to do it.

Some links for future reference:

- Related [Firefox code](https://hg.mozilla.org/integration/autoland/rev/aa2f75bd6480)
- Which was based on this [code from NetworkManager](https://github.com/lcp/NetworkManager/blob/240f47c892b4e935a3e92fc09eb15163d1fa28d8/src/nm-sleep-monitor-systemd.c)

vadz:
I have code which works but it's somewhat annoying is that we need to subscribe to a DBUS signal in any case, whether we need to do it or not, as we don't know if anybody going to be interested in these notifications or not.

I wonder if we should add some way to opt-in into the generation of these events under Linux. We could use `OnDynamicBind()` to do it by default when using `Bind()` to connect the handlers, but it would need to be called manually when using event tables.

Or am I overthinking this and it's not a big deal to call `g_dbus_connection_signal_subscribe()` on startup of all wx programs?

vadz:
Hmm, seeing that we now call `g_dbus_proxy_new_for_bus_sync()` unconditionally in 9b9ec141fbcd604aa8bfc83da6f16fc9bf5c32e0 it looks like @paulcor doesn't consider it to be a problem to do this unconditionally. The code there uses `G_BUS_TYPE_SESSION` while I need `G_BUS_TYPE_SYSTEM`, but I guess it should be fine to connect to the latter as well.

paulcor:
I measured the time for the `g_dbus_proxy_new_for_bus_sync()` plus `g_dbus_proxy_call_sync()`, it was around 0.015s. Kind of a lot of processor time, but not detectable in human time. I judged it acceptable.

vadz:
Thanks for measuring it! I admit I didn't bother doing it because I thought my results wouldn't be really representative: I have a fast and mostly idle machine, while I'd be worried about it taking more time on slower and busier ones. And I also worry that adding small delays on startup may be imperceptible for each individual one but become noticeable as they add up. But, again, I'm probably worrying too much...

vadz:
[...]

Issue #22396, opened 2022-05-05
wxPowerResource support on linux


willco007:
**Describe the proposed addition**

I was looking at adding linux support to `wxPowerResource` and the API that seems like it will work is `gtk_application_inhibit`. However, this API requires an instance of `GtkApplication`. It seems there is no way to get an instance of GtkApplication in the WX code. Is that the case, or am I missing something?

vadz:
We don't use `GtkApplication` currently and I'm not sure if we can fit into its approach.

FWIW `gtk_application_inhibit()` is implemented by using DBUS under Linux AFAICS, so we could do the same thing in our own code, but it's a lot of work, of course...

willco007:
I was afraid of that. As I'm sure you're keenly aware, the best code is code you don't have to write. Trying to maintain distro specific power management oddities within WX sounds like a pretty awful time. :)

vadz:
It's not distro-specific, GTK code just uses DBUS API. But yes, it would be a pity to have to duplicate it.

Anyhow, I'd welcome any PRs using `GtkApplication` in wxGTK, I'm not sure if we can do it, but the only way to answer it is by trying to do it.

willco007:
DBUS is universal, but then you get [bugs like this](https://gitlab.xfce.org/xfce/xfce4-session/-/issues/71). I'd love to hack on WX to attempt to get GtkApplication working, but yeah, it's always time...sigh.

PR #25147, opened 2025-02-08
Implement support for power events and preventing sleep/suspend in wxGTK

vadz:
This mainly addresses #23717 but also happens to implement #22396 as a side effect.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -4927,6 +4927,7 @@ COND_TOOLKIT_GTK_TOOLKIT_VERSION_2___LOWLEVEL_SRC_OBJECTS =  \
 	monodll_win_gtk.o \
 	monodll_gtk_window.o \
 	monodll_gtk_mimetype.o \
+	monodll_gtk_power.o \
 	monodll_gtk_dcclient.o \
 	monodll_gtk_dcmemory.o \
 	monodll_gtk_dcscreen.o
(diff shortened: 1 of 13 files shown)


MaartenBent:
I'm getting link errors:
```
/usr/bin/ld: ../lib/libwx_gtk3u_core-3.3.so.0.0.0: undefined reference to `vtable for wxPowerEvent'
/usr/bin/ld: ../lib/libwx_gtk3u_core-3.3.so.0.0.0: undefined reference to `wxEVT_POWER_SUSPENDED'
/usr/bin/ld: ../lib/libwx_gtk3u_core-3.3.so.0.0.0: undefined reference to `wxEVT_POWER_RESUME'
```

I think caused by this line. The events are implemented in `powercmn.cpp`, part of the base library. This doesn't have the `__WXGTK__` define. It works when I change it to `__UNIX__`, though that maybe breaks macOS.
https://github.com/wxWidgets/wxWidgets/blob/aa1f7db472325a831e14ea4a0f3084e4267caed6/include/wx/power.h#L43

vadz:
Oops, it indeed looks like this can't work (so I wonder how does it manage to work nevertheless for me and in the CI builds too, and what's the difference with your build).

Worse, I'm not sure how to fix it. I think we're going to need to move all `wxPowerXXX` stuff to core from base. Or we'd need to virtualize it via `wxAppTraits`, but this would be more complicated and probably not really worth it, or would it be?

MaartenBent:
Not sure what is different, my build should be the same as `CMake builds - Ubuntu 22.04 wxGTK 3`. The libraries build fine, it is the samples that fail to link.
I'm ok with moving it to core. But I don't know if there are cases where power events are used in console / base-only applications.

lanurmi:
It would make perfect sense for a Windows service (using base only) to prevent sleep. However, I'm unsure whether `SetThreadExecutionState()` actually works with services as the documentation suggests "desktop apps only". Nonetheless, such code can exist and does exist.

vadz:
It's going to be a lot of work to make this available in MSW console applications and I don't think it's worth it, so for now I've only created #25174. If we really, really need them in base, they could be virtualized later but, again, I just don't think it's worth investing time into this.

PR #25148, opened 2025-02-09
HiDPI fixes in samples and sizer overview

Blake-Madden:
Remove redundant 5 arguments to Border, replace Border 10 with DoubleBorder, add FromDIP in other places.


vadz (approved):

PR #25150, opened 2025-02-09
Improve image sample

MaartenBent:
Don't use hard-coded sizes and positions.
Make sizes DPI aware.
Fix initial size of Filter Test window.
Fix showing degree symbol in Filter Test window.

This should make it a lot easier to add new image formats to the canvas.

### DIFF
--- a/samples/image/canvas.cpp
+++ b/samples/image/canvas.cpp
@@ -16,24 +16,14 @@
     #include "wx/wx.h"
 #endif
 
-#include "wx/image.h"
 #include "wx/file.h"
-#include "wx/filename.h"
 #include "wx/mstream.h"
-#include "wx/wfstream.h"
-#include "wx/quantize.h"
-#include "wx/stopwatch.h"
 
-#if wxUSE_CLIPBOARD
-    #include "wx/dataobj.h"
-    #include "wx/clipbrd.h"
-#endif // wxUSE_CLIPBOARD
+#include "canvas.h"
 
 #include "smile.xbm"
 #include "smile.xpm"
 #include "cursor_png.c"
-
-#include "canvas.h"
 
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


samples/image/image.cpp
```diff
@@ -783,7 +783,7 @@ class MyFiltersFrame : public wxFrame
         sizerFlags4.Border();
 
         wxStaticBoxSizer *sizerHue = new wxStaticBoxSizer(new wxStaticBox(this,
-                                     wxID_ANY, wxS("Hue (°)")), wxVERTICAL);
+                                     wxID_ANY, wxS("Hue (\u00b0)")), wxVERTICAL);
```
vadz, line 786: Shouldn't this `wxS` be `L`?

samples/image/canvas.cpp
```diff
@@ -44,27 +34,23 @@ wxBEGIN_EVENT_TABLE(MyCanvas, wxScrolledWindow)
...
     my_horse_ani = nullptr;
-    m_ani_images = 0 ;
+    m_ani_images = 0;
 
-    SetBackgroundColour(* wxWHITE);
+    SetBackgroundColour(*wxWHITE);
 
-    wxBitmap bitmap( 100, 100 );
+    wxBitmap bitmap;
+    bitmap.CreateWithLogicalSize(FromDIP(wxSize(100, 100)), GetDPIScaleFactor());
```
vadz, line 48: Wouldn't `CreateWithDIPSize()` work here?

samples/image/canvas.cpp
```diff
@@ -396,266 +384,348 @@ MyCanvas::~MyCanvas()
...
+wxBitmap const& MyCanvas::GetPngBitmap() const
+{
+    return my_horse_png;
+}
+
+wxSize MyCanvas::GetDrawingSize() const
+{
+    // Aproximate the size used in OnPaint.
+    const int width = 4 * FromDIP(250);
+    const int height = 3100;
```
vadz, line 409: Strange to use `FromDIP()` for width but not for height, is this intentional?

samples/image/image.cpp
```diff
@@ -1161,11 +1161,13 @@ MyFrame::MyFrame()
     SetStatusWidths( 2, widths );
 #endif // wxUSE_STATUSBAR
 
-    m_canvas = new MyCanvas( this, wxID_ANY, wxPoint(0,0), wxSize(10,10) );
-
-    // 500 width * 2750 height
-    m_canvas->SetScrollbars( 10, 10, 50, 275 );
+    m_canvas = new MyCanvas(this);
+    wxSize cz = m_canvas->GetDrawingSize();
+    m_canvas->SetScrollbars(10, 10, cz.GetWidth() / 10, cz.GetHeight() / 10);
```
vadz, line 1166: No need for `FromDIP()` here?

vadz (commented):
Thanks, looks good, but I have a couple of minor questions, please let me know if you're going to change anything or if you think this should be merged as is.

samples/image/image.cpp
MaartenBent: No, the `SetScrollbars` arguments (`pixelsPerUnit` * `noUnits`) should match the number of pixels of the canvas.

It is actually not per-monitor DPI aware, so I'll add a DPI event handler and update `SetScrollbars` when the canvas size changes.

samples/image/image.cpp
MaartenBent: `wxS` expands to `L`, at least for me. Is there a better / preferred way?

vadz: It expands to nothing in UTF-8 build, which is the whole reason for having it in addition to wxT — which does always expand to `L`.

So there are several alternative ways to write this, none of which is ideal:

1. `wxT("\u00b0")`
2. `L"\u00b0"`
3. `wxString::FromUTF8("\xc2\xb0")`

Ideal would be to write either `L"°"` or `wxString::FromUTF8("°")` but this would require using UTF-8 encoding for the sources under all platforms and I'm not sure if we do this (in fact, I'm almost sure that we don't as we don't use `/utf-8` with MSVC).

samples/image/canvas.cpp
MaartenBent: I reverted these changes and always create a bitmap with scale 1. This way bitmap loaded from file and the bitmap created for the canvas will always have the same scale, even after DPI events.

vadz:
Maarten, should this be merged now?

MaartenBent:
Yes please. I don't have any more changes.

Issue #25104, opened 2025-01-21
Open/save file dialogs are not native on kde


generalloki:
#### Bug description:

WxWidgets shows generic open/save file dialogs instead of native dialogs.

#### Expected vs observed behaviour:

Expected open file dialog:
![Image](https://github.com/user-attachments/assets/32f28680-fa33-4f56-9bbc-87d18f8aaacd)

Shown open file dialog:
![Image](https://github.com/user-attachments/assets/42a42105-0de9-4871-a2ea-6b62f3b66e1a)

### Platform and version information

- WxWidgets Version 3.2.6
- KDE Plasma Version 5.27.11
- KDE Frameworks Version 5.115.0
- Qt Version 5.15.13
- Kernel Version 6.12.5-1-liquorix-amd64
- Graphics Platform x11
- Distro Kubuntu 24.04

vadz:
We should use `GtkFileChooserNative`, at least optionally, in wxGTK to show native dialogs, but this is not currently implemented and probably won't be very simple to do. Any help/PRs doing it would be welcome!

Currently to get native Qt dialogs you would have to build using wxQt port — which then wouldn't look natively in Gnome, of course.

Shadowblitz16:
@vadz I am the reason why the issue was posted.
Is there anyway to detect whether an os is qt or gtk?
Is there anyways to compile both gtk and qt dialogs for linux?

I am using kubuntu 24.04 and it would be nice for my native dialogs to show up without removing them from gnome users.

vadz:
It's possible to detect the current platform, but it won't do much good because you won't be able to show the one for the toolkit different from the one you compiled with anyhow.

As I wrote, the way to use the appropriate dialogs here is to use `GtkFileChooserNative`. This is not trivial because it doesn't support all features of `wxFileDialog`, so we'd have to dynamically decide whether to use it or not depending on what dialog options are used, but should be doable if anybody is motivated to work on it.

paulcor:
> We should use `GtkFileChooserNative`, at least optionally

I'm not sure this is feasible. `GtkFileChooserNative` is not a `GtkWidget`, so it's not any kind of a window.

vadz:
Not sure if `wxFileDialog` really needs to be a window, so if we could just ensure that all the window functions assert (but not crash) when called for it, it could be good enough.

[...]

Issue #24486, opened 2024-04-18
Use GtkFileChooserNative as file picker dialog


gudvinr:
### Description

With `GtkFileChooserNative` it is possible to override GTK+ dialog with `GTK_USE_PORTAL=1`, so applications will use native dialogs

That is probably only useful for KDE Plasma users though, but nice to have nonetheless.

vadz:
Yes, this would be nice to have, any contributions would be welcome.

FliegendeWurst:
I started working on this: https://github.com/FliegendeWurst/wxWidgets/commit/747bf40303d1c1f447554c1697173cc50fbd97b1

Currently it works well enough to use in PrusaSlicer. Did not yet test more advanced use cases. (And the code still throws dozens of *CRITICAL* Gtk/Glib warnings.)

PR #25152, opened 2025-02-09
Use GtkFileChooserNative for wxFileDialog/wxDirDialog when possible

paulcor:
This allows "native" dialogs with KDE. Not supported for extra control or wxFD_PREVIEW. See #24486, #25104

### DIFF
--- a/include/wx/gtk/dirdlg.h
+++ b/include/wx/gtk/dirdlg.h
@@ -6,8 +6,10 @@
 // Licence:     wxWindows licence
 /////////////////////////////////////////////////////////////////////////////
 
-#ifndef __GTKDIRDLGH__
-#define __GTKDIRDLGH__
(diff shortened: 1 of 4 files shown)


vadz (approved):
This looks really nice, thanks a lot!

One question: do we really support `Show()`ing these dialogs modelessly? FWIW wxMSW doesn't support this.

paulcor:
> One question: do we really support Show()ing these dialogs modelessly?
> FWIW wxMSW doesn't support this.
>

I have no idea, so I was conservative. It would be simpler to ignore that
possibility though.

>

vadz:
> One question: do we really support Show()ing these dialogs modelessly? FWIW wxMSW doesn't support this.
> I have no idea, so I was conservative. It would be simpler to ignore that possibility though.

Yes, I think we could just drop overridden Show().

PR #25153, opened 2025-02-10
Implement subitem for wxGenericListCtrl::HitTest()

lanurmi:
(I'm not sure if I'm missing something. It is surprising that something that can be implemented in literally 5 lines of code has remained unimplemented for 19 years.)

Pre-emptively assumed to be backported to 3.2.


vadz:
This looks good and, indeed, sometimes things are pretty simple to fix — thanks for (finally) doing it!

PR #25157, opened 2025-02-12
wxAuiToolBar:  implement statusbar help text

wsu-cb:


vadz:
Thanks, I'm merging this but with an extra test for `frame`, just in case we miss some situation where there might not be any.

PR #25161, opened 2025-02-12
Update GitHub Actions to stop using Ubuntu 20.04

vadz:
GitHub Actions will drop Ubuntu 20.04 images on 2025-04-01, so stop using
them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -282,11 +282,6 @@ jobs:
           df -h /
           echo
 
-      - name: Work around Ubuntu 22.04 runner problem
-        if: matrix.runner == 'ubuntu-22.04'
-        # See https://github.com/actions/runner-images/issues/9491#issuecomment-1989718917
-        run: sudo sysctl vm.mmap_rnd_bits=28
-
       - name: Configuring
         run: |
           wxCONFIGURE_OPTIONS="--disable-optimise $wxCONFIGURE_FLAGS"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,7 +101,8 @@ jobs:
             configure_flags: --enable-compat30
             use_xvfb: true
           - name: Ubuntu 20.04 wxGTK 3 with clang
-            runner: ubuntu-20.04
+            runner: ubuntu-latest
+            container: ubuntu:20.04
             compiler: clang++-10
             configure_flags: --disable-sys-libs
             use_xvfb: true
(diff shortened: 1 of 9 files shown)


vadz:
I have no idea what's going on with clang under newer Ubuntu, it seems to crash whenever `pthread_exit()` is called in our code, but I can't reproduce it locally in a simple example. If anybody has any ideas, they would be welcome...

vadz:
This is really bad because I can reproduce the crash locally too (although there is also some weirdness with clang in the GitHub Ubuntu 24.04 which has libllvm18 `1:18.1.3-1ubuntu1` while clang-18 depends on `libllvm18 = 1:18.1.3-1` and I had to uninstall libllvm18 `1:18.1.3-1ubuntu1` which had been installed as a dependency of something else to even be able to install clang-18 here, but it looks like this versions mismatch/difference doesn't really matter).

The stack for the crash is

```
* thread #2, name = 'test', stop reason = signal SIGSEGV: invalid permissions for mapped object (fault address: 0x7ffff55ff630)
  * frame #0: 0x00007ffff55ff630
    frame #1: 0x00007ffff73f7845 libunwind.so.1`unw_get_proc_info + 133
    frame #2: 0x00007ffff73fd06a libunwind.so.1`_Unwind_GetLanguageSpecificData + 42
    frame #3: 0x00007ffff7430115 libc++abi.so.1`__gxx_personality_v0 + 245
    frame #4: 0x00007ffff73e9ce5 libgcc_s.so.1`___lldb_unnamed_symbol340 + 197
    frame #5: 0x00007ffff73ea3c0 libgcc_s.so.1`_Unwind_ForcedUnwind + 304
    frame #6: 0x00007ffff70a57a4 libc.so.6`__GI___pthread_unwind(buf=<unavailable>) at unwind.c:130:3
    frame #7: 0x00007ffff709dd22 libc.so.6`__GI___pthread_exit [inlined] __do_cancel at pthreadP.h:271:3
    frame #8: 0x00007ffff709dd03 libc.so.6`__GI___pthread_exit(value=0x0000000000058980) at pthread_exit.c:36:3
    frame #9: 0x00007ffff7c1753d libwx_baseu-3.3.so.0`wxThread::Exit(this=0x00007fffffffcd70, status=0x0000000000058980) at threadpsx.cpp:1764:5
    frame #10: 0x00007ffff7c1727e libwx_baseu-3.3.so.0`wxThreadInternal::PthreadStart(thread=0x00007fffffffcd70) at threadpsx.cpp:946:17
    frame #11: 0x00007ffff7c18205 libwx_baseu-3.3.so.0`wxPthreadStart(ptr=0x00007fffffffcd70) at threadpsx.cpp:834:12
    frame #12: 0x00007ffff709caa4 libc.so.6`start_thread(arg=<unavailable>) at pthread_create.c:447:8
    frame #13: 0x00007ffff7129c3c libc.so.6`__clone3 at clone3.S:78
```

so this looks like some kind of libunwind mismatch, but I still have no idea what to do about it nor why it's not reproducible with a simple program like this:

```cpp
#include <pthread.h>

void* thread_func(void* arg) {
    pthread_exit(0);
}

int main() {
    pthread_t thread;
    pthread_create(&thread, NULL, thread_func, NULL);
    pthread_join(thread, NULL);
    return 0;
}
```

vadz:
FWIW I've tried with all clang versions available in Ubuntu 24.04 (from 14 to 18) and they all crash in the same way.

vadz:
I'm going to merge this just to avoid problems when GHA drops 20.04, even if the clang `wxThread` bug is pretty worrisome.

PR #25167, opened 2025-02-15
provide better default size for wxXRCPreviewVListBox

wsu-cb:
Without this change, the default size is really tiny (wxSize(1, 1)?)


vadz:
Looks good, thanks. Can this be applied (if so, please just remove the "Draft" mark from it) or do you plan to add more commits here?

wsu-cb:
> Looks good, thanks. Can this be applied (if so, please just remove the "Draft" mark from it) or do you plan to add more commits here?

I think I should mark PRs as draft until after the CI tests succeed.  However, I see that they have, so it is ready to apply.

PR #25169, opened 2025-02-15
Improve default wxImage scaling quality.

### LINKED ISSUE #24774: wxSearchCtrl icons have poor resolution
### Description

This issue is visible building "widgets" sample ...
image

- wxWidgets version  you use: 3.3.0 (master 2024-08-16)
- OS  and its version: Windows 10, Windows 11

dsa-t:
For downscaling, using just ResampleBilinear produces too sharp details if the shrink factor is significant.
Using bilinear then box-averaging produces the best results.

For non-integer upscaling, ResampleBox produces less jagged edges than ResampleNearest

Closes https://github.com/wxWidgets/wxWidgets/issues/24774

See https://github.com/wxWidgets/wxWidgets/pull/24567

This PR: 
image

Current `master` code:
image

### DIFF
--- a/src/common/image.cpp
+++ b/src/common/image.cpp
@@ -464,14 +464,21 @@ wxImage::Scale( int width, int height, wxImageResizeQuality quality ) const
     switch ( quality )
     {
         case wxIMAGE_QUALITY_NORMAL:
-            // When downscaling, we prefer to use bilinear resampling as it
-            // results in better quality at reasonable speed.
-            if ( width <= old_width && height <= old_height )
             {
-                image = ResampleBilinear(width, height);
-                break;
-            }
+                double shrinkFactorX = double(old_width) / width;
+                double shrinkFactorY = double(old_height) / height;
 
+                // When downsampling, use bilinear algorithm for pre-scaling and
+                // box average for integer part
+                if ( shrinkFactorX >= 1 && shrinkFactorY >= 1 )
+                {
+                    int shrinkInt(wxMin(shrinkFactorX, shrinkFactorY));
+
+                    image = ResampleBilinear(width * shrinkInt, height * shrinkInt);
+                    image = image.ResampleBox(width, height);
+                    break;
+                }
+            }
             // Otherwise use NEAREST for upscaling.
             wxFALLTHROUGH;
 
--- a/src/common/image.cpp
+++ b/src/common/image.cpp
@@ -476,11 +476,14 @@ wxImage::Scale( int width, int height, wxImageResizeQuality quality ) const
 
                     image = ResampleBilinear(width * shrinkInt, height * shrinkInt);
                     image = image.ResampleBox(width, height);
-                    break;
                 }
+                else
+                {
+                    image = ResampleBox(width, height);
+                }
+
+                break;
             }
-            // Otherwise use NEAREST for upscaling.
-            wxFALLTHROUGH;
 
         case wxIMAGE_QUALITY_FAST:
         case wxIMAGE_QUALITY_NEAREST:


vadz (commented):
So there are 3 changes here AFAICS:

1. Use bilinear only if we're shrinking the image in both directions (before it would be applied when shrinking it horizontally only, for example).
2. Use `BILINEAR` with int factor and `BOX` for the rest when shrinking.
3. Make `QUALITY` behave as `BOX` instead of `NEAREST` when not shrinking.

At least (2) and (3) seem to be independent, so maybe we should split this in 2 commits (I can do it) to be able to revert one of them if we run into any problems with them later?

vadz:
BTW, I still think it would be great to have a way to objectively measure an image sharpness (or bluriness), either in the tests or in `wxImage` itself. It looks like it would be relatively straightforward to do it (computing the Laplacian of the image using convolution with a 3x3 matrix) and being able to compare the numbers for the test images before/after changing the algorithm should be much simpler and faster than painstakingly zooming into the images.

dsa-t:
> painstakingly zooming into the images.

I open two image tabs at 100% scale in the browser and use Ctrl+Tab and Ctrl+Shift+Tab to compare them.

dsa-t:
The algorithm works like this:

Say we want to scale 128 -> 24.
```text
shrinkFactor = 5.33333333333
shrinkInt = 5

```
1. Apply Bilinear: 128 -> 120 (24 * 5)
2. Apply Box-average: 120 -> 24 (shrink by integer 5)

dsa-t:
> 1. Use bilinear only if we're shrinking the image in both directions (before it would be applied when shrinking it horizontally only, for example).

Fixed.

dsa-t:
> BTW, I still think it would be great to have a way to objectively measure an image sharpness (or bluriness), either in the tests or in `wxImage` itself. It looks like it would be relatively straightforward to do it (computing the Laplacian of the image using convolution with a 3x3 matrix)

That would take me much more time than just looking at the images unfortunately.

dsa-t:
> At least (2) and (3) seem to be independent, so maybe we should split this in 2 commits (I can do it) to be able to revert one of them if we run into any problems with them later?

Split done.

vadz:
> > BTW, I still think it would be great to have a way to objectively measure an image sharpness (or bluriness), either in the tests or in `wxImage` itself. It looks like it would be relatively straightforward to do it (computing the Laplacian of the image using convolution with a 3x3 matrix)
> 
> That would take me much more time than just looking at the images unfortunately.

I've cheated and just used `wxExecute(wxString::Format("convert %s -statistic StandardDeviation 5x5 -format '%%[fx:maxima]\n' info:"), file)` to get the results, shown below: old/new refer to master and this PR and the numbers themselves are sharpness (i.e. the output of that command) as percentage of the original:

| Size | Old 128 | Old 64 | Old 48 | New 128 | New 64 | New 48 |
| ---: | ------: | -----: | -----: | ------: | -----: | -----: |
| 16   |  94 |  91  |  91  |  91 |  93 |  94
| 20   |  93 |  93  |  91  |  90 |  88 |  90
| 21   |  90 |  90  |  93  |  90 |  95 |  90
| 22   |  93 |  94  |  92  |  89 |  91 |  90
| 24   |  93 |  91  |  93  |  88 |  89 |  93
| 30   |  92 |  92  |  92  |  89 |  89 |  92
| 32   |  93 |  94  |  90  |  92 |  94 |  95
| 40   |  94 |  94  |  92  |  87 |  90 |  87
| 42   |  94 |  92  |  97  |  90 |  91 |  86
| 43   |  97 |  98  |  93  |  88 |  91 |  88
| 48   |  95 |  91  | 100  |  87 |  87 | 100
| 64   |  97 | 100  | 104  |  89 | 100 |  80
| 72   |  96 | 101  | 104  |  88 |  80 |  82
| 128  | 100 | 106  | 117  | 100 |  87 |  81

I have no idea how to interpret this, however... I can't say that this unambiguously shows that new is better (or worse). Upscaling is probably better now as we were getting values  sharper than the original before, which can't be good, but is it really better to have blurrier images when downscaling?

dsa-t:
> I can't say that this unambiguously shows that new is better (or worse).

Yes, it's not trivial to develop a good metric here.

> but is it really better to have blurrier images when downscaling?

If you look at the 128 -> 24 or 48 -> 16 cases for example, the old one is so sharp that it loses details and has jagged edges. This metric doesn't tell the full story.

dsa-t:
Rendering the SVG at the exact resolutions and comparing it might be a better metric.

SVGs: 
[icon_pcm](https://github.com/user-attachments/assets/4996545f-3b34-4a2b-8ecc-8862b1b68719)
[icon_pcm_24](https://github.com/user-attachments/assets/1d902a8d-6581-4763-a2c3-766b2b8d4986)
[update_pcb_from_sch](https://github.com/user-attachments/assets/2272be5d-605a-4adf-914c-94c8627a25c4)
[hv45mode](https://github.com/user-attachments/assets/23837ebc-5746-4b3d-a95f-0619d724c34f)

KiCad renders them like this:
```bash
inkscape --export-area-snap --export-type=png ${inputFile} \
       --export-filename ${outFile} -w ${pngWidth} -h ${pngHeight}
```

vadz:
Hmm, so your suggestion would be to compute the difference between the image rendered by Inkscape and the one resized by wx? I guess I could try doing this...

dsa-t:
Yes, if you really want a metric.

vadz:
I've implemented your proposal as a new pseudo test in [this commit](https://github.com/wxWidgets/wxWidgets/commit/91ee31a13ae10c95b9d102dc1beaa05844908f94) and running it as

```
$ WX_TEST_SVG_PATH=icon_pcm.svg WX_TEST_INITIAL_SIZE=64 WX_TEST_TARGET_SIZE=32 ./tests/test_gui wxImage::ResizeQuality
```

etc allows to compare rendering the given SVG at 32x32 directly and by rescaling it from 64x64.

The results are partially encouraging: for downscaling your ("combined") method always results in smaller error and in much smaller ones for not-exact-shrink cases. However for upscaling "box" doesn't really look better than "nearest", although sometimes it does and, notably and importantly, it is better for 1.5 factor.

There could be many things wrong here, of course, the choice of using the simple distance in RGB space is pretty questionable, but this seems to give at least some good argument for changing the code for downscaling — at least for those icons. But for upscaling, I wonder if we shouldn't apply box averaging only for some factors and not always?

Also, I have no idea why does bicubic come out so badly using this metric, could I have done something really stupid here?

dsa-t:
> But for upscaling, I wonder if we shouldn't apply box averaging only for some factors and not always?

Nearest looks acceptable for factors >= 3 where edge jaggedness matters less. So does the slight edge blur from box-average.


> Also, I have no idea why does bicubic come out so badly using this metric, could I have done something really stupid here?

It should be slightly more blurry than Bilinear, so more pixels should differ.

vadz:
I'm still paralyzed by indecision concerning this PR. The improvements don't seem _that_ obvious to me and while I don't see any obvious regressions either, I'm afraid there could be some cases in which they would appear.

Right now I think to merge this in master but not in 3.2, but this doesn't seem very satisfactory.

Any thoughts from others?

vadz:
> Right now I think to merge this in master but not in 3.2, but this doesn't seem very satisfactory.

As there have been no other suggestions, I've created #25252 with the slightly modified versions of the commits here, documentation updates and the change adding the pseudo-test to explore the resizing quality.

I'm closing this one because I will merge the other one when the CI jobs pass there (but please let me know if I broke anything, of course).

Thanks again!

PR #25170, opened 2025-02-15
Improve wxBitmap scaling quality. (3.2)

dsa-t:
For downscaling, using just Bilinear produces too sharp details if the shrink factor is significant. Using Bilinear then Box-averaging produces the best results.

For non-integer upscaling, Box-averaging produces less jagged edges than Nearest-neighbor.

See #24567

This PR:
image

Current 3.2:
image

---

### DIFF
--- a/src/common/bmpbase.cpp
+++ b/src/common/bmpbase.cpp
@@ -92,7 +92,7 @@ void wxBitmapHelpers::Rescale(wxBitmap& bmp, const wxSize& sizeNeeded)
     }
     else
     {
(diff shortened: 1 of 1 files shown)


dsa-t:
Updated to `if ( shrinkFactorX >= 1 && shrinkFactorY >= 1 )`

vadz:
I was too afraid to break something to apply this for 3.2.7,  so I'm closing it, let's see if we get any complaints after merging #25252 and releasing 3.3.0 and, if not, we can backport that PR into a later 3.2 release.

PR #25171, opened 2025-02-15
Prevent last pixels cut-off in wxImage::ResampleNearest() (3.2)

dsa-t:
This PR:
image

Current 3.2:
image

---


vadz (commented):
Thanks, this looks correct to me, but I can't see the difference between the 2 versions of `tests/image/cross_nearest_neighb_256x256.png`, what is it?

Also, it looks like this should be applied to master too and not just 3.2?

dsa-t:
> but I can't see the difference between the 2 versions of `tests/image/cross_nearest_neighb_256x256.png`, what is it?

It's shifted by 1px. You can open them in two browser tabs to compare.



> Also, it looks like this should be applied to master too and not just 3.2?

Yes.

vadz:
> > but I can't see the difference between the 2 versions of `tests/image/cross_nearest_neighb_256x256.png`, what is it?
> 
> It's shifted by 1px. You can open them in two browser tabs to compare.

Sorry, I have no idea how do you manage to see a single pixel difference, regardless of the number of tabs used, you must have superb eyesight.

But I've compared the old and new files by converting both of them to XPM and now I see the differences:

1. The cross branches are now 1 pixel narrower (which is good as they should be 256/9 ≈ 28.4 pixels wide).
2. The background is now `#000001` instead of black as before.

The second change doesn't look correct to me, am I missing something?

dsa-t:
GIMP shows 000001 in the RGB image too... I'm not sure why that is.

vadz:
So, just to be clear, the output of the new code really contains these not-quite-black pixels? I don't see how could this be explained by your changes... Could the old code already be producing it too? This would still pass the test because there is a tolerance of 1 in it:

https://github.com/wxWidgets/wxWidgets/blob/9995edb3a3f0c52e9482f794f60ad675575ce102/tests/image/image.cpp#L1461

dsa-t:
So the XPM decoder uses #000001 as a mask color.

Before saving the image I've removed the mask to make it opaque.

Should I have left the transparent background instead?

vadz:
OK, thanks, I've pushed this to master and plan to cherry-pick 1415cbf234ed79d5ba0683d3d1cef96851beb6f3 to 3.2.

PR #25172, opened 2025-02-16
Use SVG bitmaps for the generic `wxSearchCtrl`

vadz:
The main commit here is the last but one, the previous ones are just refactoring/cleanup and the last one is a hack which I had to make to get the correct background for the "cancel" button: I have no idea why, but without this it showed greyish (`#f0f0f0`) background instead of white in light mode.

Speaking of light/dark, I'm also not proud of black/white replacement, but I don't see how else can the bitmaps be adjusted for the dark mode. Any better ideas?

### DIFF
--- a/include/wx/generic/srchctlg.h
+++ b/include/wx/generic/srchctlg.h
@@ -187,8 +187,8 @@ private:
 #endif // wxUSE_MENUS
 
     void RecalcBitmaps();
-    wxBitmap RenderSearchBitmap( int x, int y, bool renderDrop );
-    wxBitmap RenderCancelBitmap( int x, int y );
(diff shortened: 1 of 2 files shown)


dsa-t:
Before:
image

After:
image

dsa-t:
The arrow seems too close to the magnifying glass for me, but I'm not a graphic designer.

vadz:
I'm definitely not a designer neither, so if anybody can improve this, please let me know.

But part of the reason I want to apply this is that I think it should be _much_ simpler to get nicer SVG images from someone or somewhere than to get somebody rewrite/improve the drawing code.

PR #25175, opened 2025-02-17
Show also toolkit micro version in wxGetLibraryVersionInfo() output

lanurmi:
Instead of outputting just "10.0", output e.g. "10.0.19045".

On Windows, at least, today the micro version has more significance than it perhaps did 17 years ago when the code was added.


vadz:
Thanks, I agree that this is more useful and will merge it soon (and backport to 3.2 too).

PR #25176, opened 2025-02-17
Fix compilation of Qt port

lanurmi:
Add missing `#include "wx/geometry.h"`.

Broken by 734face (Add support for raw touch events, 2025-01-08).

### DIFF
--- a/include/wx/qt/private/converter.h
+++ b/include/wx/qt/private/converter.h
@@ -16,6 +16,7 @@
 #include "wx/kbdstate.h"
 #include "wx/gdicmn.h"
 #include "wx/colour.h"
+#include "wx/geometry.h"
 
 #include <QtCore/QRect>
 #include <QtCore/QSize>


vadz:
Thanks!

Do you know why this wasn't caught by the CI build?

vadz:
Thanks, this indeed explains it. Not sure if we should turn the PCH off in the CI build or maybe turn it on for wxQt by default — it depends on how expensive to compile Qt headers are, so it would be nice to measure the time with/without it.

paulcor:
> it depends on how expensive to compile Qt headers are

Qt headers are not in the PCH.

vadz:
> > it depends on how expensive to compile Qt headers are
> 
> Qt headers are not in the PCH.

Oh, I thought we'd include them in the PCH when using it, like we do with `windows.h`. If we don't, then using PCH is not worth it and I'll turn it off for the CI builds.

PR #25186, opened 2025-02-21
Serialize and restore wxAUI pane visibility too

vadz:
Panes may be hidden by user, so it makes sense to store their visibility as part of the serialized representation to be able to restore the same visual layout later.


vadz:
@rudolfwalter FYI this is merged now, please let me know if you see any problems with it!

Issue #25072, opened 2025-01-12
wxAuiNotebook should have the ability to show close buttons on only some tabs


craftyjon:
### Description

KiCad would like to use wxAuiNotebook in situations where there are some permanent tabs that the user should not be able to close, and some transient tabs that can be opened and closed by user action.  It doesn't seem to be possible to show a close button on only some tabs right now.  We propose that there be some kind of API added to set close button behavior per-page instead of only the global style flags.

vadz:
Is it the same as the concept of "pinned" tabs or something different? We've had requests for adding support for pinned tabs, that have a special "pin" icon, always appear to the left (well, "near", taking RTL languages into account) side of the bar and are treated differently by some operations, e.g. "Close all" may not close the pinned tabs.

However at least in MSVS, which is a typical example of a complex TDI app, pinned tabs still have a close button...

So do we need "normal", "pinned" and "pinned and not closeable" ("fixed"?) states perhaps?

craftyjon:
I think it's a different concept.  Pinning tabs is a user action, like when the user decides they want to mark some tabs as special and then have them shown differently or not affected by some actions.

What we are talking about is removing the ability for the user to interact with certain tabs.  Maybe "locked" state or something like this.  No close button, and less important but a nice follow on, being able to disable drag and drop on these tabs might also be nice.

craftyjon:
@imciner2 points out that if wxAuiNotebook were refactored to be more overrideable in client code, it would be easier for us to just implement this special logic inside KiCad

imciner2:
> @imciner2 points out that if wxAuiNotebook were refactored to be more overrideable in client code, it would be easier for us to just implement this special logic inside KiCad

Yep, this comes back to storing objects vs. pointers in the AUI controls (e.g., similar to the previous problem with AUI toolbars discussed here: https://github.com/wxWidgets/wxWidgets/pull/23311). In the `wxAuiNotebook`, the tabs are stored using an object of type `wxAuiTabContainer` directly, not a pointer. This means we can't just implement a new tab container that overrides the behavior of `wxAuiTabContainer::GetCloseButtonState()` to hide the button on the tabs we want, because there is no way to actually have the `wxAuiNotebook` accept/use our custom class.

vadz:
[...]

PR #25187, opened 2025-02-21
Add support for pinned and locked tabs to wxAUI

vadz:
This implements support for the usual concept of pinned tabs (shown before all the other normal tabs) as well as the locked tabs from #25072 (shown even before the pinned tabs and can't be closed nor drag-moved by the user, even among the tabs of the given group).

After reviewing handling of the pinned tabs in different programs, I've discovered that there are quite a few different ways in which they're handled:

- Some programs don't show any special button for them at all, only allow [un]pinning them from a menu.
- Some programs show the pin/unpin button only on the current tab.
- Some programs show the pin button only on the current tab, but unpin button on all pinned tabs.

And this is why we have `wxAUI_NB_PIN_ON_ACTIVE_TAB` and `wxAUI_NB_UNPIN_ON_ALL_PINNED` styles that provide support for all of the possibilities above. Note that I couldn't find any program that always showed pin button on all tabs and experimenting with it showed why: this just makes the display too visually busy. OTOH MSVS shows pin/unpin buttons on all tabs, but only when the mouse is hovering over it and this looks quite nice (but wastes space on the not shown buttons) and is something we could implement later.

As always, any comments/remarks/suggestions are welcome!

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -121,6 +121,13 @@ Changes in behaviour not resulting in compilation errors
   under all platforms instead of behaving as SetValue() under MSW and doing
   nothing elsewhere.
 
+- If you have a custom class deriving from wxAuiGenericTabArt, possibly
+  indirectly, e.g. via wxAuiMSWTabArt, you need to update it to override the
+  new virtual functions DrawPageTab() and GetPageTabSize() instead of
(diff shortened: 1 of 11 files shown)


craftyjon:
I did a simple test of this branch and it works well for what I need!  Thank you.

Issue #25006, opened 2024-12-16
AUI borders too thick on high DPI


rudolfwalter:
### Description
On the master branch version, the borders of AUI panes (including toolbars) are too thick on high DPI displays.
Specifically, the borders used to be 1px thick, and now they are more than that (see screenshots).

From what I can tell, this was changed in #24844. If I'm reading the diff correctly, it is now impossible to set the thickness back to 1px from client code, because it gets scaled up unconditionally.

wx 3.2 (good):
![image](https://github.com/user-attachments/assets/0681875f-6e52-4a02-90c6-89a965bdb7db)

wx 3.3 master (bad):
![image](https://github.com/user-attachments/assets/5c7735e8-b4ec-4c28-bd86-b6487c6729fc)

### Platform and version information

- wxWidgets version: 3.3 master (bfa27c2)
- wxWidgets port: wxMSW
- OS: Windows 11

MaartenBent:
One could argue that 2px is correct, because everything will double at 200%. But it would indeed be nice to allow 1px (or any px) as well.

What about an extra argument in SetMetric/GetMetric with an enum that specifies the type (dip/px)?
Or use negative values for px, and positive for dip?

Side note: I notice wxAuiGenericToolBarArt GetElementSize/SetElementSize should get the same behavior.

vadz:
I still don't see any really good solution, but what about this:

1. Add `virtual int GetMetricForWindow(int id, wxWindow* window);` to `wxAuiDockArt` which would, by default, just do `return window->FromDIP(GetMetric(id))`.
2. Replace all existing calls to `FromDIP(m_art->GetMetric())` with calls to this function.
3. Document that it can be overridden in a custom `wxAuiDefaultDockArt`-derived class to avoid the call to `FromDIP()` (possibly for some metrics only).

Alternatively, maybe don't apply `FromDIP()` in the default implementation if the argument is either `wxAUI_DOCKART_PANE_BORDER_SIZE` or `wxAUI_DOCKART_SASH_SIZE`. This would still be rather weird, but should behave better by default and it's less bad if this behaviour can be overridden.

If nobody sees anything better, I'll implement this soon.

vadz:
@rudolfwalter #25191 should address this issue, please let me know if you still see anything wrong with it.

rudolfwalter:
Looks good to me!

dsa-t:
[...]

PR #25191, opened 2025-02-22
Avoid scaling small AUI dock art metrics by DPI

vadz:
This addresses #25006.

@MaartenBent Please let me know if you have any objections.

### DIFF
--- a/include/wx/aui/dockart.h
+++ b/include/wx/aui/dockart.h
@@ -38,6 +38,13 @@ public:
     virtual ~wxAuiDockArt() = default;
 
     virtual wxAuiDockArt* Clone() = 0;
+
+    // This function should be used for querying metrics in the new code, as it
+    // will scale them by the DPI of the provided window if necessary. The
+    // older GetMetric() function is kept for compatibility and shouldn't be
+    // used outside of this class itself.
+    virtual int GetMetricForWindow(int id, wxWindow* window);
+
     virtual int GetMetric(int id) = 0;
     virtual void SetMetric(int id, int newVal) = 0;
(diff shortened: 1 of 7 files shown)


MaartenBent:
No objections.

Do you want to address `wxAuiGenericToolBarArt` `GetElementSize`/`SetElementSize` as well? And add a `GetElementSizeForWindow()`. Right now they are initialized on startup, so only high-dpi aware, not per-monitor dpi aware.
It  doesn't have small values that should remain the same for different DPI. All can be scaled.
https://github.com/wxWidgets/wxWidgets/blob/2b6bce1d49929b1863c794d2e45aa3add0b6d69c/src/aui/auibar.cpp#L140-L143

`wxAuiMSWToolBarArt` initializes the element sizes with values from uxTheme. After some testing this seems to always return values at 100% DPI, no matter what the actual DPI is. Except the dropdown size, which doesn't use the uxTheme.
https://github.com/wxWidgets/wxWidgets/blob/2b6bce1d49929b1863c794d2e45aa3add0b6d69c/src/aui/barartmsw.cpp#L38-L48

vadz:
Yes, I can make a similar PR for `wxAuiToolBarArt` too, if we agree that this is the approach to use. Thanks!

MaartenBent:
Yes, I think this is the best approach. Get/Set values should be DPI-unaware, same as it is now for `wxAuiDockArt`.

For `wxAuiToolBarArt` we could use `FromDIP(GetElementSize())` directly, but it is probably best to use the same approach as here. And allow a user to override it.

PR #25196, opened 2025-02-24
Fix showing labels of newly added wxNotebook pages in wxOSX

### LINKED ISSUE #25194: wxNotebook on macOS: label is wrong size on last tab
### Description

On wxMac, wxNotebook tabs that are added after the initial creation of a control have the behavior that the last tab always has an incorrect label width calculated, resulting in a truncated label.

#### To Reproduce:

Run the notebook sample, then use the Pages > Add Page action from the main menu (the select or don't select variants both work).

First run:

Image

Added page:

Image

Added another page:

Image

- wxWidgets version  you use: 3.2.6 (with KiCad patches, but nothing in this area).  Also have seen this in recent 3.3 from upstream.
- OS  and its version: macOS 14.7.2

vadz:
The existing code added new tabs before setting their label, which resulted in them being shown initially truncated.

Fix this by adding new tabs with their correct label (and image), as this seems to make more sense and also fixes the problem.

Also don't override SetMaximum() in wxCocoaTabView, this doesn't make sense as wxNotebook doesn't have a "maximum" property unlike wxSpinCtrl or wxGauge, for which this function (and the associated SetMinimum()) exists, and just do what this function did in SetupTabs() itself, which is the function explicitly intended to be overridden in wxNotebook.

Closes #25194.

### DIFF
--- a/src/osx/cocoa/notebook.mm
+++ b/src/osx/cocoa/notebook.mm
@@ -224,7 +224,7 @@ public:
             return [slf indexOfTabViewItem:selectedItem]+1;
     }
 
-    void SetMaximum( wxInt32 maximum ) override
+    void SetupTabs( const wxNotebook& notebook) override
     {
         wxNSTabView* slf = (wxNSTabView*) m_osxView;
         int cocoacount = [slf numberOfTabViewItems ];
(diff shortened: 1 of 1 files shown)


vadz:
Merged in the commit referenced above.

Issue #25053, opened 2025-01-08
Key Events report incorrect character on Dvorak keymap on GNOME


agrif:
### Description

Pressing keys while in the Dvorak keyboard map on GNOME produces incorrect key events on 3.2.6. The character reported by KeyUp and KeyDown appears to have no relation at all to either the letter typed or the label on the key being pressed. This seems to only happen on X. Wayland sessions report the correct keys.

This appears to be a regression introduced by 1997154abfd8b276c33337c366926d66dfbb2c8a and PR #24773. This issue exists on wx built with this commit, and it does not exist on the parent commit.

This issue is part of my investigation into [this KiCad issue](https://gitlab.com/kicad/code/kicad/-/issues/19545).

#### Expected vs observed behaviour:

**Expected:** Pressing the key that produces 'B' on my keyboard should report 'B' in key events. Or, at least, it should report either 'B' or the label on the key (which happens to be 'N').

**Observed:** Pressing the key that produces 'B' on my keyboard produces KeyUp and KeyDown events for 'X', which is neither 'B' nor 'N'.

Here is a screenshot of the keyboard sample program after I attempted to type 'abc' on a Dvorak layout. Note that on a QWERTY keyboard, these letters are labeled 'ANI', and these are not the letters used in the events either (except for A, which is the same between both layouts).

![after](https://github.com/user-attachments/assets/a60a1448-9aa7-42d7-a64e-0c01c43d0caa)

#### To Reproduce:

1. Build wxGTK.
2. Use the Dvorak keymap.
3. Run the keyboard sample program.
4. Type 'b'.
5. Observe the events saying 'X'.

### Platform and version information

- wxWidgets version: 3.2.6
- wxWidgets port: wxGTK
- O:S Debian trixie
  + GTK version: 3.24.43
  + tested on X. Appears to work fine on Wayland.
  + Desktop Environment: GNOME

agrif:
A minor update: It looks like wx is accidentally pushing an already-mapped key through the map *again*. It reports 'X' when I type 'B' because 'X' is bound to the 'B' key in my keymap. Same with J/C. So the process looks something like this:

 * I press my 'N' key, which in my keymap produces a 'B'.
 * wx receives the 'B' and maps it again through my keymap, producing an 'X'.

vadz:
Sorry, I didn't have time to look at this yet, but I will, thanks for reporting this.

[...]

Issue #25199, opened 2025-02-24
Wrong key code in `wxEVT_KEY_DOWN` events under Wayland


vadz:
The key code in the event generated for pressing `5` with `Shift` in the pressed state is `5` when using wxMSW or wxGTK/X11, but `%` when using wxGTK/Wayland because the relevant code in `wxTranslateGTKKeyEventToWx()` is simply not used when not using X11 because it uses X-specific functions such as `XKeysymToKeycode()`.

It should be possible to use XKB (which is also used with Wayland) instead to do it correctly.

PR #25200, opened 2025-02-25
Rewrite key events generation in wxGTK

vadz:
This fixes recently reported #25053 and #25199 which I noticed while debugging this, hopefully without breaking any of the previously fixed problems listed in the last commit.

@paulcor Any help with reviewing/testing this would be welcome!


vadz:
Sorry, another change, this one affecting only `wxEVT_CHAR` generation to make it work correctly for `Ctrl+^` and `Ctrl+_` too.

Right now `wxEVT_KEY_DOWN` events seem to be consistently generated for all keys in wxMSW and wxGTK and the only discrepancies between them for `wxEVT_CHAR` I still see are the following:

1. `Ctrl-Backspace` generates `CHAR` event with both key and Unicode code set to `DEL` in wxMSW but wxGTK uses just `Backspace` for both.
2. `Ctrl-Return` generates `CHAR` event with both key and Unicode code set to `Ctrl-J` in wxMSW but wxGTK uses just `Return` for both.
3. When using Cyrillic layout, `Ctrl-Letter` generates `CHAR` event using the same values as in US layout in wxMSW but doesn't generate any `CHAR` events in wxGTK.

In all cases wxGTK behaviour seems more logical to me, but I'm more worried about changing wxMSW, which almost certainly always behaved like this, than wxGTK which keeps changing...

agrif:
Just a quick note, this PR does indeed fix #25053 for me.

unxed:
2 seems the best for me. For the next major release of course.

vadz:
Concerning `Ctrl-BS`: I've also compared the events it generates in wxMSW with those generated by `Ctrl-DEL` and there is really no logic here, as we get `DEL/DEL` for key/Unicode for the former and `DEL/0` for the latter (which, coincidentally, the same as in wxGTK with this PR).

So maybe it's not worth following wxMSW here neither. And the choice then is between modifying it to behave logically and generate events with `BS/BS` as wxGTK does or documenting this as a known incompatibility.

vadz:
For now I'm just documenting the existing incompatibilities and merging this into both master and 3.2.

PR #25207, opened 2025-02-28
wxQt: Fix Qt objects destruction

AliKet:
Fix crashes start to appear when porting to Qt version 6.x


vadz (approved):
Not sure what the actual change/problem is, but this looks reasonable, so I'm going to merge it if there are no objections.

Thanks!

vadz:
Oops, sorry, I didn't notice that `wxTreeCtrl::SelectItemSingle` unit test failed in the CI job. I am not sure why did it happen/how is this related, but it probably is?

AliKet:
@vadz Thanks for reviewing this.

I'm trying to fix `wxQt` port for Qt6.x series (I'm working with the latest version 6.8.2) and enable one CI Actions with that version.
Everything goes well locally so far, so I think I should have marked this PR as WIP for now a will notfiy you when this is ready.

Thanks again.

AliKet:
@vadz @dsa-t

Any comment about the last commit ? my tests show no regression or any problem at all with it. nor the wx samples, for both (Qt5 and Qt6).

See https://lists.qt-project.org/pipermail/interest/2019-June/033365.html?utm_source=chatgpt.com

vadz:
Sorry, I don't like this commit. Not only because of `Pending()` becoming useless, but because it seems that `Dispatch()` has very different semantics to the other ports, in which it's really supposed to block until the next event arrives.

I think it might make more sense to make `Pending()` always return `false` in wxQt and tell people to use `DispatchTimeout(0)` if they want to check if there is really an event to be dispatched already waiting. We could also deprecate `Pending()`, which is probably not used much, and add some 3-valued `int IsReadyForDispatch()` return `true`, `false` or `-1` for "don't know", but I'm not really sure if it's worth it.

AliKet:
> Sorry, I don't like this commit. Not only because of `Pending()` becoming useless, but because it seems that `Dispatch()` has very different semantics to the other ports, in which it's really supposed to block until the next event arrives.
> 
Adding `QEventLoop::WaitForMoreEvents` to `processEvents()` will fix `wxQt` non-conforming  `Dispatch()`:
https://github.com/wxWidgets/wxWidgets/blob/14b62a27f547518f80a86d1f9e5e9c140cc5db8c/src/qt/evtloop.cpp#L133-L137

> I think it might make more sense to make `Pending()` always return `false` in wxQt and tell people to use `DispatchTimeout(0)` if they want to check if there is really an event to be dispatched already waiting. We could also deprecate `Pending()`, which is probably not used much, and add some 3-valued `int IsReadyForDispatch()` return `true`, `false` or `-1` for "don't know", but I'm not really sure if it's worth it.

I'll make these changes than, i.e.: make `Pending()` always return `false` in wxQt and fix the non-conforming  `Dispatch()` there and update the documentation.

vadz:
Hmm, [the docs](https://doc.qt.io/qt-6/qeventloop.html) say

> Specifying the [WaitForMoreEvents](https://doc.qt.io/qt-6/qeventloop.html#ProcessEventsFlag-enum) flag makes no sense and will be ignored.

which doesn't make much sense to me (what is even the purpose of `WaitForMoreEvents` then?), but would seem to indicate that the solution might be not so simple as this...

AliKet:
@vadz  I'm trying to add support for building wxQt with Qt6 or Qt5 (using CMake only) and add a CI Actions for Qt6.
After that, I'll try to fix [94d2944](https://github.com/wxWidgets/wxWidgets/pull/25207/commits/94d2944a796326aed1e8d203f96033bdeab20a38) commit.

dsa-t:
> I'm trying to add support for building wxQt with Qt6 or Qt5 (using CMake only)

You can also take a look at how I did it prevoisly: https://github.com/wxWidgets/wxWidgets/pull/24345/files#diff-37710dd5ae4de235cd4cfd793f966d29e6a11dac7b7ac61f6c6499d65957378a

AliKet:
> [WaitForMoreEvents](https://doc.qt.io/qt-6/qeventloop.html#ProcessEventsFlag-enum)

Sorry, but the right function we are calling in wx is this https://doc.qt.io/qt-6/qeventloop.html#processEvents which is a wrapper for https://doc.qt.io/qt-6/qabstracteventdispatcher.html#processEvents

AliKet:
@vadz Unfortunately, runnning the tests in the CI Actions stil crashes with Qt6. So I just give up and added a CI under Appvyor here d82f827 instead and it works.

AliKet:
> > I'm trying to add support for building wxQt with Qt6 or Qt5 (using CMake only)
> 
> You can also take a look at how I did it prevoisly: https://github.com/wxWidgets/wxWidgets/pull/24345/files#diff-37710dd5ae4de235cd4cfd793f966d29e6a11dac7b7ac61f6c6499d65957378a

Thanks, but I did it here cc49571 just like what the Qt documentation suggests.

vadz:
Thanks for your work on this, I think this is good to merge, but I'd prefer to avoid another job to AppVeyor: it's already the slowest part of the CI and making it even slower is really not very appealing.

What exactly prevents us from adding the new CI job to GitHub Actions? And, if we can't do this for Qt6, perhaps we could add one for Qt5 (or just keep the existing one under Unix) and replace the Qt5 AppVeyor build with the Qt6 one?

AliKet:
@vadz I think this is ready now.

src/qt/renderer.cpp
```diff
@@ -623,8 +623,10 @@ wxRendererQt::DrawGauge(wxWindow* win, wxDC& dc, const wxRect& rect,
 
     if ( flags & wxCONTROL_SPECIAL )
     {
+#if QT_VERSION_MAJOR < 6
```
vadz, line 626: Curious, Qt6 really doesn't support vertical progress bars any longer?

src/qt/stattext.cpp
```diff
@@ -35,6 +35,11 @@ wxStaticText::wxStaticText(wxWindow *parent,
     Create( parent, id, label, pos, size, style, name );
 }
 
+wxStaticText::~wxStaticText()
+{
+    GetQLabel()->setBuddy( nullptr );
```
vadz, line 40: I'm not familiar with wxQt details, but it might be worth adding a comment explaining why do we need to do this here?

src/qt/window.cpp
```diff
@@ -1737,7 +1724,11 @@ bool wxWindowQt::QtHandleMouseEvent ( QWidget *handler, QMouseEvent *event )
 
     // Use screen position as the event might originate from a different
     // Qt window than this one.
+#if QT_VERSION_MAJOR >= 6
+    const wxPoint mousePos = ScreenToClient(wxQtConvertPoint(event->globalPosition().toPoint()));
+#else
```
vadz, line 1729: It looks like this is done in a few places, perhaps it could be worth adding a function hiding the difference between `globalPos()` and `globalPosition()` and just call it from here and elsewhere?

vadz (approved):
Thanks again for your work!

I have no real comments, but I'll wait for a bit before merging this just in case someone else does or you decide to make any changes corresponding to the minor remarks below.

How would you describe changes of this PR in the change log?

src/qt/window.cpp
AliKet: I added `wxQtGetEventPosition()` to hide mouse position retrieval in c12363e. Regarding `globalPosition()`/`globalPos()`, there are only two: this one and the other `globalPosition()`/`screenPos()` in `winevent.h`, which I left unchanged.

src/qt/renderer.cpp
AliKet: Sure, Qt6 does support vertical progress bars.  I'll fix it...

AliKet: Sorry, in [432060b](https://github.com/wxWidgets/wxWidgets/pull/25207/commits/432060bbb56353c84caf97b5b24fea9ecda1da8d) I just fixed the progress bar's rendering under Windows (10!) only and here is the result of that commit (Light&Dark modes):
render
render_dark

Since I don't (yet) have a modern enough KDE desktop to test this, I suggest (for anyone want to help) to test this PR as is. If the progress bar doesn't render correctly, apply this changes:
```
diff --git a/src/qt/renderer.cpp b/src/qt/renderer.cpp
index 6649f5da40..436a889e93 100644
--- a/src/qt/renderer.cpp
+++ b/src/qt/renderer.cpp
@@ -635,7 +635,14 @@ wxRendererQt::DrawGauge(wxWindow* win, wxDC& dc, const wxRect& rect,

     if ( isKDE )
     {
+#if QT_VERSION_MAJOR < 6
+        qtStyle->drawControl(QStyle::CE_ProgressBar, &option, painter, qtWidget);
+#else
+        painter->save();
+        painter->rotate(-90);
         qtStyle->drawControl(QStyle::CE_ProgressBar, &option, painter, qtWidget);
+        painter->restore();
+#endif
     }
     else
     {

```
Sorry again!

AliKet: @vadz please take this commit [5948583](https://github.com/wxWidgets/wxWidgets/pull/25207/commits/59485833347223767fa3b3b7105d63ca595081f9) 
 with a grain of salt, because rendering the progress bar in Qt was not fun at all! and I don't know why there isn't a simple and straightforward way (with minimal divergence at least) to render it under all platforms.

Tested and verified under:
- Windows 10 64bits + Qt 6.8.2
- Windows 10 32bits  + Qt 5.15.2
- KDE Neon (VirtualBox) + Qt 6.8.2
- Ubuntu, Kubuntu, Xubuntu, LXQT 18.04 + Qt 5.9.0

src/qt/renderer.cpp
```diff
@@ -586,85 +567,153 @@ wxRendererQt::DrawFocusRect(wxWindow* win, wxDC& dc, const wxRect& rect, int WXU
...
 }
 
-void
-wxRendererQt::DrawGauge(wxWindow* win, wxDC& dc, const wxRect& rect,
-                        int value, int max, int flags)
+namespace
 {
-    auto painter = wxGetQtPainter(dc);
+#if QT_VERSION_MAJOR < 6
+bool wxIsKDEDesktop()
```
vadz, line 573: Could we reuse `wxGUIAppTraits::GetDesktopEnvironment()` from wxGTK in wxQt under Unix too?

vadz:
Sorry, I have really no idea about what's going on with the progress bars here, but let's merge it like this if this is the way it needs to be. Reviews from people familiar with Qt would be welcome, of course, but I guess it would be better to merge this before 3.3.0, wouldn't it?

AliKet:
> 
> How would you describe changes of this PR in the change log?

Two things:
- Mention that `wxQt` port can now be built and run using the latest **Qt** version (i.e.  Qt 6.8.2 LTS)
- `wxEventLoop::Pending()` is useless under `wxQt` and shouldn't be called to check for pending events.

src/qt/renderer.cpp
AliKet: Done here 5225e3c

AliKet:
> Sorry, I have really no idea about what's going on with the progress bars here, but let's merge it like this if this is the way it needs to be. Reviews from people familiar with Qt would be welcome, of course, but I guess it would be better to merge this before 3.3.0, wouldn't it?

I would say yes! because people are less likely to test PRs anyway!!

TIA.

PR #25209, opened 2025-02-28
Some #include cleanup in wx headers

lanurmi:
* Remove unused C `#include`s in headers
* Do not `#include "wx/list.h"` for no reason in various headers

These changes can break compilation of some existing code.

### DIFF
--- a/include/wx/filesys.h
+++ b/include/wx/filesys.h
@@ -16,6 +16,7 @@
 #include "wx/stream.h"
 #include "wx/datetime.h"
 #include "wx/filename.h"
+#include "wx/list.h"
 
 #include <unordered_map>
 
--- a/include/wx/datetime.h
+++ b/include/wx/datetime.h
@@ -19,8 +19,6 @@
 
 #include <vector>
 
-#include <limits.h>             // for INT_MIN
-
 #include "wx/longlong.h"
 #include "wx/anystr.h"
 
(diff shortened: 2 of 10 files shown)


include/wx/filesys.h
```diff
@@ -16,6 +16,7 @@
 #include "wx/stream.h"
 #include "wx/datetime.h"
 #include "wx/filename.h"
+#include "wx/list.h"
```
vadz, line 19: Unrelated, but I think we could change type of `m_Handlers` to `std::list<>` without breaking compatibility: nobody is supposed to use it directly anyhow.

vadz (approved):
I think we can indeed live with breaking the existing code using `wxList` without including `wx/list.h`, so I'll merge this, thanks.

Out of curiosity, how did you do it, did you use IWYU to do it (in which case I'm surprised it didn't find more unneeded headers) or did you do it manually?

lanurmi:
> Out of curiosity, how did you do it, did you use IWYU to do it (in which case I'm surprised it didn't find more unneeded headers) or did you do it manually?

I did it manually, by skimming through all `#include`s under `include/wx/` and making educated guesses about which ones look like unnecessary, then trying to ensure they also are unnecessary. It's not impossible some technically necessary `#include` has been removed in the process, but that doesn't make a huge practical difference as everything still compiles.

As for wx's own headers, `wx/list.h` just stood out as suspicious, and that's the only one I investigated. Could be nice attempting a more comprehensive and automated cleanup some time using IWYU (that I know of, but haven't tried myself).

PR #25211, opened 2025-03-02
Update Corsican translation on 2025-03

Patriccollu:
Hello,

This is an update of **Corsican** (co) localization to take the latest strings into account.

I also modified several other strings.

Best regards,
Patriccollu.


vadz:
Thanks for the update! Pushed now.

Issue #25210, opened 2025-03-01
wxSearchCtrl default sizing problems on GTK3


ojwb:
#### Bug description:

The default sizing behaviour of wxSearchCtrl in a toolbar in wxGTK3 makes it close to unusable by the user.

#### Expected vs observed behaviour:

Observed:

Adding a `wxSearchCtrl` to a toolbar in a wxGTK3 build (using either git master and the Debian 3.2.6+dfsg-2 packages) adds a little rounded cornered white rectangle with a magnifying glass icon inside.  Reasonable so far - modern search controls often start shrunk like this.

* There seems to be a very narrow rectangle I need to click in to give it focus which isn't great for usability.

* The much worse problem is that if I enter text then the search control's cancel icon appears and the rectangle gets wider to accommodate that, but only just enough wider to accommodate the icon - it doesn't grow in width to actually show any of the text being entered.  (If the wxSearchCtrl is created with a value set, as in the toolbar sample, then the cancel icon is there to start with, but it still doesn't resize to show either the initial text or any text subsequently entered.)

Expected:

* I'd expect it not to be so fiddly to give search control focus.

* I'd expect entering text to cause to cause the search control to grow up to the control's maximum width (and probably then scroll to show the text around the caret position).

#### To Reproduce:

Steps to reproduce the behaviour, please make them as detailed as possible.
For example:

1. Run the `toolbar` sample
2. Maximise the window
3. Ctrl+F (to add a wxSearchCtrl to the right end of the toolbar)
4. Notice you can't see any initial value set on the control (which is `needle` according to the sample source code) but the "cancel" icon is present which shows there is a value
5. Click between the two icons in the search control to give it focus (and notice the area to click in is quite narrow)
6. Type text (and notice you can't see any of the text you enter either

- wxWidgets version either git master and the Debian 3.2.6+dfsg-2 packages
- wxWidgets port wxGTK
- Debian unstable
  + GTK version: 3.24.41-1
  + Which GDK  backend is used: X11
  + Desktop environment : mate
  + Current theme: Menta

vadz:
Thanks, I can indeed see this and it's not X11-specific as I see it with Wayland too. I'll try to have a look.

vadz:
This is fixed now, so closing. Thanks again for reporting!

ojwb:
The fix "Respect best control size when adding it to wxGTK wxToolBar" was backported to 3.2.7:

[...]

PR #25212, opened 2025-03-02
Search control and toolbar size fixes in wxGTK

vadz:
See #25210.

### DIFF
--- a/src/gtk/toolbar.cpp
+++ b/src/gtk/toolbar.cpp
@@ -644,6 +644,14 @@ bool wxToolBar::DoInsertTool(size_t pos, wxToolBarToolBase *toolBase)
 #else
             tool->m_item = GTK_TOOL_ITEM(gtk_widget_get_parent(gtk_widget_get_parent(control->m_widget)));
 #endif
+            // The widget size is not controlled by wx, so at least make sure
+            // that its minimal size is respected by GTK.
+            const wxSize minSize = control->GetBestSize();
+            if ( minSize != wxDefaultSize )
+            {
+                gtk_widget_set_size_request(control->m_widget, minSize.x, minSize.y);
+            }
+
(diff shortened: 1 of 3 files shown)


vadz:
Note to self: if/when backporting, replace `constexpr` in the first commit with `const`.

paulcor:
I think it would be better to implement `DoGetSizeFromTextSize()`. I'm looking into it.

vadz:
Thanks for looking at this, but I'm not sure what exactly do you propose. Do you mean implement it and call it from `DoGetBestSize()`? Because otherwise I don't think it would be taken into account...

I thought about accounting for the text extent here, but decided it might not be the best idea for the search text control, which typically has a fixed size and doesn't change it depending on e.g. the latest string you searched for. But I'm not totally sure about this.

src/gtk/toolbar.cpp
```diff
@@ -644,6 +644,14 @@ bool wxToolBar::DoInsertTool(size_t pos, wxToolBarToolBase *toolBase)
 #else
             tool->m_item = GTK_TOOL_ITEM(gtk_widget_get_parent(gtk_widget_get_parent(control->m_widget)));
 #endif
+            // The widget size is not controlled by wx, so at least make sure
+            // that its minimal size is respected by GTK.
+            const wxSize minSize = control->GetBestSize();
+            if ( minSize != wxDefaultSize )
```
paulcor, line 650: Minor, but the check isn't needed. -1 means the same thing on both sides, and nothing will happen if the values aren't changing.

paulcor (approved):
Second commit looks like the right thing.

vadz:
First commit replaced by f2faedcb96b2ff6be7304370065803dcd0d2a980, second one applied as 8f9f1a42ec127b5aa800d49843eff0855ccc8945.

PR #25213, opened 2025-03-02
Improve wxAUI hint appearance and API

vadz:
This PR disables fade-in effect by default, as proposed in [this thread](https://groups.google.com/g/wx-dev/c/WC_mZJWDBfI) but also shows the hints when dragging tabs between different tab controls in a better (IMO) way.

Any comments and, especially, testing of this PR would be welcome!


vadz:
No objections, so merging.

PR #25215, opened 2025-03-03
Fix wxSearchCtrl best size with GTK

paulcor:
Implement GetSizeFromTextSize() and use that to get a reasonable width.
See #25210

### DIFF
--- a/src/gtk/srchctrl.cpp
+++ b/src/gtk/srchctrl.cpp
@@ -16,12 +16,13 @@
 
 #ifndef WX_PRECOMP
     #include "wx/menu.h"
+    #include "wx/utils.h"
 #endif //WX_PRECOMP
 
-#include "wx/utils.h"
-#include "wx/gtk/private.h"
+#include "wx/gtk/private/wrapgtk.h"
 #include "wx/gtk/private/event.h"
 #include "wx/gtk/private/gtk3-compat.h"
+#include "wx/gtk/private/stylecontext.h"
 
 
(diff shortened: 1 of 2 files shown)


src/gtk/srchctrl.cpp
```diff
@@ -384,6 +385,43 @@ void wxSearchCtrl::PopupSearchMenu()
 
 #endif // wxUSE_MENUS
 
+wxSize wxSearchCtrl::DoGetBestSize() const
+{
+    int w;
+    GetTextExtent(wxS("0"), &w, nullptr);
+    return DoGetSizeFromTextSize(w * 8);
```
vadz, line 392: This doesn't change much, but why not just

```suggestion
    return DoGetSizeFromTextSize(GetCharWidth() * 8);
```

? Or, if you really prefer to use "0" (even though I don't see why would it be a particularly good choice for a search control), at least

```cpp
    return DoGetSizeFromTextSize(GetTextExtent(wxS("0")).x * 8);
```

vadz (approved):
Thanks!

As this will replace the first commit of #25212, could you please let me know what do you think about the second one, so that I knew whether it should be just closed or partially applied?

src/gtk/srchctrl.cpp
paulcor: No reason, just didn't think of it.

paulcor:
Applied via f2faedcb96b2ff6be7304370065803dcd0d2a980

PR #25216, opened 2025-03-03
3.2 backports 202503

vadz:


vadz:
FYI, I'm dropping the commit resulting in ABI incompatibility reports, this is not ideal as `wxSearchCtrl` still doesn't have the correct size by default in 3.2, but I don't want to lose this check.

PR #25217, opened 2025-03-03
Try to make code operating with `GdkWindow` more clear

vadz:
There are no real changes here, but this hopefully makes it more clear to understand which widget is used where.

### DIFF
--- a/include/wx/gtk/bmpcbox.h
+++ b/include/wx/gtk/bmpcbox.h
@@ -124,7 +124,7 @@ public:
     virtual bool IsEditable() const override;
     virtual void SetEditable(bool editable) override;
 
-    virtual GtkWidget* GetConnectWidget() override;
+    virtual GtkWidget* GetConnectWidget() const override;
 
 protected:
     virtual GdkWindow *GTKGetWindow(wxArrayGdkWindows& windows) const override;
--- a/include/wx/gtk/combobox.h
+++ b/include/wx/gtk/combobox.h
@@ -123,7 +123,7 @@ public:
 
     virtual void GTKDisableEvents() override;
     virtual void GTKEnableEvents() override;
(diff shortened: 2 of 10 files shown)


paulcor:
Looks fine, but I'm not in love with the name `GTKGetMainWindow()`. The term "main window" isn't really used in the code, so I don't find that name very helpful. `GTKGetWindow()` already exists, so you wanted to distinguish from that. Maybe `GTKGetGdkWindow()`? Ugly, but a little better IMO. Or maybe `GTKWidgetGetWindow()`, as our version of `gtk_widget_get_window()`.

vadz:
> Looks fine, but I'm not in love with the name `GTKGetMainWindow()`.

Thanks for looking at this and yes, I don't like it neither, it was just the least bad one I could find.

> The term "main window" isn't really used in the code, so I don't find that name very helpful. `GTKGetWindow()` already exists, so you wanted to distinguish from that.

Exactly.

> Maybe `GTKGetGdkWindow()`? Ugly, but a little better IMO. Or maybe `GTKWidgetGetWindow()`, as our version of `gtk_widget_get_window()`.

It has to be distinguished from `GTKGetDrawingWindow()` and `GTKGetConnectWindow()` so I think there really should be some adjectif here. I thought about `GTKGetValidWindow()` which is different from the former, but not from the latter.

What about `GTKGetDefaultConnectWindow()` to show that it is the same as `GTKGetConnectWindow()` by default/as long as `GetConnectWidget()` is not overridden?

vadz:
@paulcor Any objections to using `GTKGetDefaultConnectWindow()`?

paulcor:
I like `GTKGetDefaultConnectWindow()` even less. Just go with this as-is.

PR #25218, opened 2025-03-03
Run ABI check workflow under Ubuntu 24.04

vadz:
This uses newer version of libabigail.


vadz:
But, unsurprisingly, this is incompatible with the ABI files generated under 22.04 with a different gcc version.

PR #25232, opened 2025-03-10
Make wxGetGtkPaperSize accessible

Blake-Madden:
Move `wxGetGtkPaperSize()` declaration into header file. This can be useful for those doing low-level printing code under wxGTK.

### DIFF
--- a/include/wx/gtk/print.h
+++ b/include/wx/gtk/print.h
@@ -26,6 +26,8 @@ typedef struct _GtkPageSetup GtkPageSetup;
 
 typedef struct _cairo cairo_t;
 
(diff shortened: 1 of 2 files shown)


vadz (commented):
I'm not sure how wist it is to expose this function, to be honest, its semantics is not very obvious (AFAICS it's "use paper ID if it's valid or paper size otherwise", in particular just ignoring the size if the ID is given) and we need to at least document that the returned pointer needs to be freed using the appropriate GTK function.

On balance, I think you'd be better just copying this function code (or the part of it that you actually need) in your own application instead of exposing it.

@paulcor Do you have any opinion about this?

vadz:
Sorry, I don't think we want to expose this like this.

PR #25235, opened 2025-03-11
Use smart pointers in wxGTK code instead of manual memory management

vadz:
No real changes here, just get rid of a number of `g_free()` and similar calls.

Also optimize `wxGtkCollatedArrayString` implementation a little.

### DIFF
--- a/include/wx/gtk/private/glibptr.h
+++ b/include/wx/gtk/private/glibptr.h
@@ -20,7 +20,20 @@ class wxGlibPtr
 public:
     wxGlibPtr() = default;
     explicit wxGlibPtr(const T *p) : m_ptr(const_cast<T*>(p)) { }
+    wxGlibPtr(wxGlibPtr&& other) : m_ptr(other.m_ptr) { other.m_ptr = nullptr; }
     ~wxGlibPtr() { g_free(m_ptr); }
+
+    wxGlibPtr& operator=(wxGlibPtr&& other)
+    {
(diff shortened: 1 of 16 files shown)


paulcor (approved):

include/wx/gtk/private/glibptr.h
```diff
@@ -0,0 +1,57 @@
...
+    {
+        if ( &other != this )
+        {
+            g_free(m_ptr);
+            m_ptr = other.m_ptr;
+            other.m_ptr = nullptr;
+        }
+
+        return *this;
+    }
```
Ayxan13, line 36: We could avoid branching here by swapping the pointers. The moved-from object will be destroyed eventually:

```suggestion
    wxGlibPtr& operator=(wxGlibPtr&& other)
    {
        std::swap(m_ptr, other.m_ptr);
        return *this;
    }
```

include/wx/gtk/private/glibptr.h
```diff
@@ -0,0 +1,57 @@
...
+
+    const T* get() const { return m_ptr; }
+
+    operator const T*() const { return m_ptr; }
```
Ayxan13, line 40:
```suggestion
    explicit operator const T*() const { return m_ptr; }
```

vadz: Have you tried compiling the library after applying this suggestion?

PR #25236, opened 2025-03-11
Move wxGetLibraryVersionInfo() to core from base

lanurmi:
Move determining compile-time GTK+/QT versions to wxAppTraits, which is presumably the only reason why wxGetLibraryVersionInfo() needed to be in core.

Revert "Move wxGetLibraryVersionInfo() to core from base." This reverts commit 9aea2510431d7109ff7055bce02576ebd80f593b.

### DIFF
--- a/include/wx/apptrait.h
+++ b/include/wx/apptrait.h
@@ -171,6 +171,10 @@ public:
     virtual wxString GetAssertStackTrace();
 #endif // wxUSE_STACKWALKER
 
+    // Text to be appended to the description. Currently used for getting
+    // compile-time versions of GTK+ and Qt.
+    virtual wxString GetLibversionExtraText() const = 0;
+
 private:
     static wxSocketManager *ms_manager;
 };
@@ -238,6 +242,7 @@ public:
 
     virtual bool IsUsingUniversalWidgets() const override { return false; }
     virtual wxString GetDesktopEnvironment() const override { return wxEmptyString; }
+    virtual wxString GetLibversionExtraText() const override { return wxEmptyString; }
 };
 
 // ----------------------------------------------------------------------------
@@ -278,6 +283,7 @@ public:
     }
 
     virtual wxString GetDesktopEnvironment() const override { return wxEmptyString; }
+    virtual wxString GetLibversionExtraText() const override { return wxEmptyString; }
 };
 
 #endif // wxUSE_GUI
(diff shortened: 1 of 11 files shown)


interface/wx/utils.h
```diff
@@ -231,8 +231,6 @@ void wxInfoMessageBox(wxWindow* parent);
     @see wxVersionInfo
 
     @header{wx/utils.h}
-
-    @library{wxcore}
```
vadz, line 235: This should probably be replaced with `@library{wxbase}` rather than removed?

include/wx/apptrait.h
```diff
@@ -171,6 +171,10 @@ class WXDLLIMPEXP_BASE wxAppTraitsBase
     virtual wxString GetAssertStackTrace();
 #endif // wxUSE_STACKWALKER
 
+    // Text to be appended to the description. Currently used for getting
+    // compile-time versions of GTK+ and Qt.
+    virtual wxString GetLibversionExtraText() const = 0;
```
vadz, line 176: Sorry for nitpicking, but I find the name of this function difficult to parse. Could we use `GetLibVersionText()`, i.e. capitalize "Version"?

Also, it's not really clear which "Lib" is this about, so maybe use `GetGUIToolkitVersion()` instead?

vadz (commented):
Thanks, no real comments, but I'd really like to rename the new function.

include/wx/apptrait.h
lanurmi: > Sorry for nitpicking, but I find the name of this function difficult to parse. Could we use `GetLibVersionText()`, i.e. capitalize "Version"?

Sure.

> Also, it's not really clear which "Lib" is this about,

It is the "Lib" and "Version" in wxGet*Lib*rary*Version*Info(), and it is to be parsed as "extra text for library version info description". Though in any case it is an intentionally undocumented function.

> so maybe use `GetGUIToolkitVersion()` instead?

But it is not the GUI toolkit version. It is free-form text, which contains the GUI toolkit version at an unspecified position. Calling it "extra text" also doesn't restrict its use for GUI toolkit version only, and when one day somebody needs some more platform-specific text added to the description, the same function can be used for that.

vadz: What about something like `GetPlatformDescription()` then? I see why you used "Lib" here after your explanation, but I still think it's confusing. Mentioning that this description is used for the library version info in a comment should be enough, but if you think it wouldn't, we could call it `GetPlatformDescriptionForLibVersionText()`.

vadz:
Thanks for the update, will (squash) merge in a moment.

PR #25237, opened 2025-03-11
Fix using C++ objects from block in wxWebRequestURLSession code

### LINKED ISSUE #25231: wxWebSessionSync URLSession (macOS) Crashes after calling Execute()
### Description

My program uses a wxWebSessionSync from a worker thread to download a couple of files, and when using the default URLSession backend on macOS, creating a request and calling Execute() results in a (delayed) EXC_BAD_ACCESS error and crash.

This occurs whether the request succeeds or fails, and only happens with the URLSession backend. Switching to the CURL backend (Which I had some trouble with not all that long ago in #24969 which was resolved) completes without issue.

#### Stack trace:

```
* thread #12, stop reason = EXC_BAD_ACCESS (code=1, address=0x3eaddcbf5060)
    frame #0: 0x00007ff80edc9a4a libobjc.A.dylib`objc_release + 42
  * frame #1: 0x0000000100b3fb88 libwx_baseu_net.dylib`wxWebRequestURLSession::Execute()::Semaphore::~Semaphore(this=0x00006000017890b0) at webrequest_urlsession.mm:240:13
    frame #2: 0x0000000100b3c865 libwx_baseu_net.dylib`wxWebRequestURLSession::Execute()::Semaphore::~Semaphore(this=0x00006000017890b0) at webrequest_urlsession.mm:239:9
    frame #3: 0x0000000100b3c849 libwx_baseu_net.dylib`__Block_byref_object_dispose_((null)=0x0000600001789080) at webrequest_urlsession.mm:0
    frame #4: 0x00007ff80ee961cd libsystem_blocks.dylib`_Block_object_dispose + 153
    frame #5: 0x0000000100b3ca42 libwx_baseu_net.dylib`__destroy_helper_block_e8_32r40r((null)=0x0000600000cda610) at webrequest_urlsession.mm:0
    frame #6: 0x00007ff80ee96752 libsystem_blocks.dylib`_call_dispose_helpers_excp + 37
    frame #7: 0x00007ff80ee9657e libsystem_blocks.dylib`_Block_release + 233
    frame #8: 0x00007ff813a41951 CFNetwork`___lldb_unnamed_symbol5083 + 111
    frame #9: 0x00007ff80edcd26c libobjc.A.dylib`AutoreleasePoolPage::releaseUntil(objc_object**) + 168
    frame #10: 0x00007ff80edca719 libobjc.A.dylib`objc_autoreleasePoolPop + 227
    frame #11: 0x00007ff80ede7b06 libobjc.A.dylib`objc_tls_direct_base<AutoreleasePoolPage*, (tls_key)3, AutoreleasePoolPage::HotPageDealloc>::dtor_(void*) + 140
    frame #12: 0x000000010078e424 libsystem_pthread.dylib`_pthread_tsd_cleanup + 600
    frame #13: 0x00000001007868af libsystem_pthread.dylib`_pthread_exit + 70
    frame #14: 0x00000001007867f6 libsystem_pthread.dylib`_pthread_start + 136
    frame #15: 0x000000010078ebef libsystem_pthread.dylib`thread_start + 15
```

#### To Reproduce:

Even the simplest example causes this crash. For the sake of simplicity, the request can be placed in the main thread. (My application is in a worker thread but it makes no difference).

It also makes no difference whether the default or a new session is used, so long as it uses the URLSession backend.

```cpp
class MyApp : public wxApp {
public: 
    virtual bool OnInit() override {
        const auto uri{wxURI{ /* URL */ }.BuildURI()};
        auto request{wxWebSessionSync::GetDefault().CreateRequest(uri)};
        auto result{request.Execute()};
        
        return false;
    }
}
```

Of course, this should exit normally, but it crashes as described above.

- wxWidgets version you use: git master (commit b218237e)
- wxWidgets port you use: wxOSX
- OS and its version: macOS Ventura 13.7.1

vadz:
The code added in cd05d0a788 (Partially implement sync requests support in NSURLSession backend, 2024-08-16) was wrong as it copied Semaphore object into the block storage, resulting in destroying the dispatch semaphore twice.

Fix this by making Semaphore non-copyable, as it should have always been, and using a unique_ptr<> to access it.

Also implement copying properly for TaskResult, which is also used both in the block and outside of it.

Closes #25231.

### DIFF
--- a/src/osx/webrequest_urlsession.mm
+++ b/src/osx/webrequest_urlsession.mm
@@ -227,7 +227,7 @@ wxWebRequestURLSession::DoPrepare(void (^completionHandler)(NSData*, NSURLRespon
 wxWebRequest::Result wxWebRequestURLSession::Execute()
 {
     // Define the variables used inside the completion handler.
-    __block class Semaphore
+    class Semaphore
     {
     public:
         Semaphore()
@@ -252,19 +252,38 @@ wxWebRequest::Result wxWebRequestURLSession::Execute()
 
     private:
         dispatch_semaphore_t m_sem;
-    } sem;
 
-    __block struct TaskResult
+        wxDECLARE_NO_COPY_CLASS(Semaphore);
+    };
+
+    // Using __block with Semaphore object itself doesn't work for some reason:
+    // the compiler still wants to copy it. So wrap it in a (smart) pointer.
+    __block std::unique_ptr<Semaphore> sem(new Semaphore());
+
+    struct TaskResult
     {
         NSData* data = nil;
         NSError* error = nil;
+
+        TaskResult() = default;
+
+        TaskResult(const TaskResult& other)
+        {
+            data = [other.data retain];
+            error = [other.error retain];
+        }
+
+        TaskResult& operator=(const TaskResult&) = delete;
 
         ~TaskResult()
         {
             [data release];
             [error release];
         }
-    } taskResult;
+    };
+
+    // Here we can allow copying the object.
+    __block TaskResult taskResult;
 
     // Initialize the task with the completion handler that will wake us up
     // after copying the result into local variables.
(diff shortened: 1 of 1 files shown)


vadz:

PR #25239, opened 2025-03-13
in old SDK or system, MAC_OS_VERSION_14_0 not defined

lencil:
In some old environment or osxcross, MAC_OS_VERSION_14_0 might not be defined. It will cause incorrect linking error.

### DIFF
--- a/src/osx/cocoa/window.mm
+++ b/src/osx/cocoa/window.mm
@@ -4244,7 +4244,7 @@ void wxWidgetCocoaImpl::UseClippingView()
 {
 // starting from Sonoma child windows are bleeding through under the scrollbar
 // use native scrollviews therefore
-#if __MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_14_0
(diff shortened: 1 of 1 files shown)


vadz:
Sorry, this doesn't seem right, even if it's not defined by the system header, it should be defined here: https://github.com/wxWidgets/wxWidgets/blob/96c72db0d7888509d29fd1b4df6ca364398a7641/include/wx/platform.h#L470-L472

Please check what's going on in your build more carefully, I don't see how is it possible for this symbol to be undefined.

In any case, we don't want to add such checks everywhere, so whatever the problem is, it should be fixed in `wx/platform.h` and not here.

lencil:
I found what was going on. I use v3.2.6 tag in which version platform.h does not include MAC_OS_VERSION_14_0 define. But 3.2.6 does use MAC_OS_VERSION_14_0 in src\osx\cocoa\window.mm.

vadz:
Thanks, I'll backport bbe28bdb7b248f71c1c7fe7e84566db27b8f80c0 into 3.2 before 3.2.7.

PR #25248, opened 2025-03-17
wxGTK: fix build on macOS

barracuda156:
Closes https://github.com/wxWidgets/wxWidgets/issues/23475

Tested on 14.7.1 arm64 and 10.6.8 powerpc.

Few notes:

1. I did not add https://github.com/wxWidgets/wxWidgets/commit/6585c10d0b12283310b32289ee6cce965cb88f1b commit by @MaartenBent here, but we do need it.
2. CMake build bakes in wrong dylib paths; I fixed that with `install_name_tool` in destroot. This is a separate issue and it is not addressed here.
3. I passed a few flags to the build. Not sure if that is to be handled in the source or rather left to the end-user or downstream distribution like MacPorts.

```
    if {${os.major} < 12} {
        configure.cxxflags-append \
                    -F/System/Library/Frameworks/ApplicationServices.framework/Versions/A/Frameworks
    }

    # FIXME: a hack for the following error:
    # /System/Library/Frameworks/CoreServices.framework/Frameworks/CarbonCore.framework/Headers/MachineExceptions.h:115:4:
    # error: 'vector' does not name a type
    platform powerpc {
        configure.cxxflags-append \
                    -U__APPLE_ALTIVEC__
    }

    configure.ldflags-append \
                    -undefined dynamic_lookup
```

@vadz I make this a draft first; perhaps, these can be improved before merging.


vadz:
Sorry, there are too many non-trivial changes here, I'm not 100% sure all of them are safe (as a matter of fact, some of them are clearly not, as they broke some CI jobs), so I won't merge this for 3.2.7. Let's close this one and please make a new PR against master when you have time for it. Also don't hesitate to make several PRs if you're more sure about some fixes than the others and/or if you think it makes sense to separate them, smaller PRs would be simpler to merge. Thanks!

barracuda156:
@vadz No drama, I am fine with local fixes for now, but it would be nice if you and @MaartenBent find time to look into this. We only need to fix this correctly for modern macOS, so this is easily testable on CI and locally.

Provided wxGTK build uses `uilocale` and `stdpaths` from Unix and not Cocoa, fixing the whole thing for legacy macOS and powerpc becomes trivial. The patch for that restored earlier code as a fallback, I believe. See https://github.com/wxWidgets/wxWidgets/pull/25248/commits/f748da8281b2a95c77063a47914733f3f634875f

Everything non-trivial here is not about legacy systems and can be tested on modern macOS. (Well, I myself did most of the work on Sonoma arm64.)

PR #25252, opened 2025-03-18
Improve `wxImage` default rescale algorithm

vadz:
See #25169.

### DIFF
--- a/interface/wx/image.h
+++ b/interface/wx/image.h
@@ -56,9 +56,10 @@ enum wxImageResizeQuality
         Default image resizing algorithm used by wxImage::Scale().
 
         Currently this is the same as wxIMAGE_QUALITY_NEAREST when enlarging
-        the image (in at least one direction) and wxIMAGE_QUALITY_BILINEAR when
-        reducing it, which produces relatively good results for the images
-        typically used for the icons in the GUI applications.
+        the image (in at least one direction) and a combination of using
+        wxIMAGE_QUALITY_BILINEAR and wxIMAGE_QUALITY_BOX_AVERAGE when reducing
+        it, which produces relatively good results for the images typically
+        used for the icons in the GUI applications.
     */
     wxIMAGE_QUALITY_NORMAL,
 
--- a/interface/wx/image.h
+++ b/interface/wx/image.h
@@ -55,11 +55,13 @@ enum wxImageResizeQuality
     /**
         Default image resizing algorithm used by wxImage::Scale().
 
-        Currently this is the same as wxIMAGE_QUALITY_NEAREST when enlarging
-        the image (in at least one direction) and a combination of using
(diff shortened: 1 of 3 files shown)


vadz:
@dsa-t Any objections to merging this?

interface/wx/image.h
```diff
@@ -55,10 +55,13 @@ enum wxImageResizeQuality
     /**
         Default image resizing algorithm used by wxImage::Scale().
 
-        Currently this is the same as wxIMAGE_QUALITY_NEAREST when enlarging
-        the image (in at least one direction) and wxIMAGE_QUALITY_BILINEAR when
-        reducing it, which produces relatively good results for the images
-        typically used for the icons in the GUI applications.
+        Currently this algorithm uses wxIMAGE_QUALITY_BILINEAR with an integer
+        factor when reducing the image size (in both directions) and, either
```
dsa-t, line 59: Bilinear scales with the non-integer factor in fact.

vadz: I meant to say that bilinear is applied with an integer scaling factor, which seems correct (it's called `shrinkInt` and has type `int`) or am I missing something here?

dsa-t: Bilinear shrinks it to an intermediate size which is an integer multiple of the target size. Box average then shrinks it by the integer factor to the target size.

"uses wxIMAGE_QUALITY_BILINEAR with an integer factor when reducing the image size" doesn't read as that to me.

src/common/image.cpp
```diff
@@ -464,16 +464,24 @@ wxImage::Scale( int width, int height, wxImageResizeQuality quality ) const
     switch ( quality )
     {
         case wxIMAGE_QUALITY_NORMAL:
-            // When downscaling, we prefer to use bilinear resampling as it
-            // results in better quality at reasonable speed.
+            // When downscaling, use bilinear algorithm for shrinking by the
+            // integer factor and box average for the rest.
```
dsa-t, line 468: Maybe something like this:
```cpp
            // When downscaling, use bilinear algorithm for shrinking by the
            // fractional part and box average for the integer part.
```

interface/wx/image.h
vadz: OK, yes, I agree, this is confusing. What about

```
Currently this algorithm uses wxIMAGE_QUALITY_BILINEAR when reducing the image size
(in both directions) to resize the image to an integer multiple of the target size
and, after doing this, or as the only step when enlarging the image, applies
wxIMAGE_QUALITY_BOX_AVERAGE to obtain the desired size.
```

?

dsa-t: Sounds good.

vadz:
Merged with the documentation/comment changes in the commit above.

PR #25257, opened 2025-03-20
Allow not catching unhandled exceptions

vadz:
Replaces #24997, see the discussion there.

### DIFF
--- a/include/wx/app.h
+++ b/include/wx/app.h
@@ -307,6 +307,9 @@ public:
     // OnExceptionInMainLoop()
     virtual void OnUnhandledException();
 
+    // Call OnUnhandledException() on the current wxTheApp object if it exists.
+    static void CallOnUnhandledException();
+
     // Function called if an uncaught exception is caught inside the main
     // event loop: it may return true to continue running the event loop or
     // false to stop it. If this function rethrows the exception, as it does by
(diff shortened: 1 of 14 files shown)


vadz:
@lanurmi Does this address your use case, i.e. should we merge this?

lanurmi:
Thank you for your work! I have skimmed through the changes of the PR, and finally today also tested it myself on MSW and Linux. Looks good to me, and most importantly exceptions are indeed left uncaught with the new system option set.

This is still something I'd hope to see in 3.2 branch also. I think the changes of #24997 might even apply cleanly to 3.2 without conflicts.

vadz:
Let's merge this in master and see if we can get any feedback about it.

I have to admit that I'm rather reluctant to complicate the code in 3.2 and diverge it from the master version for this, but if this really proves useful, maybe we could still do it...

vadz:
It turns out that this adds a pretty noticeable overhead to event handling, making things much slower as we call a relatively expensive `wxSystemOptions::IsFalse()` for *every* event.

The simplest fix would be to cache the results of environment variables lookup to `wxSystemOptions`. This will still leave some overhead, but hopefully it shouldn't be that bad...

vadz:
See #25399.

solemnwarning:
This looks just like what I want - I've been fiddling with `OnExceptionInMainLoop()` and `OnUnhandledException()` trying to come up with ways to circumvent them so the crash handler on macOS can properly handle uncaught exceptions.

Are there any changes besides this and #25399 I should look at backporting into my 3.2 build?

PR #25261, opened 2025-03-21
Qualify _cmd parameter in wxOSX_touchesCancel as WXUNUSED

tobiolo:


tobiolo:
Thanks for your review! As I am unsure about that, I added the WXUSED qualifier.

vadz:
This one looks good too, thanks!

PR #25269, opened 2025-03-25
wxQt: Fix passing command line arguments to QApplication and to wxWidgets

AliKet:
For example: "`./minimal -style fusion`" or "`./minimal -stylesheet myStylesheet`" didn't work before, but it work now with this commit.

Explanation:
------------
The old code was not correct: it just crashes when trying to delete "`argv_`" and if it doesn't, it tries to recreate it with the already consumed args which doesn't make sense to wxWidgets.

Additionally, it makes unnecessary memory allocations for the arguments, which simply results in memory leaks.

-----------------------
FYI: http://bugreports.qt.nokia.com/browse/QTBUG-7551 no longer available and I don't know what the problem was with the reported bug?

### DIFF
--- a/include/wx/qt/app.h
+++ b/include/wx/qt/app.h
@@ -9,8 +9,6 @@
 #ifndef _WX_QT_APP_H_
 #define _WX_QT_APP_H_
 
-#include <wx/scopedarray.h>
-
 #include <memory>
 
 class QApplication;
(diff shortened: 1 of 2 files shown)


lanurmi:
> FYI: http://bugreports.qt.nokia.com/browse/QTBUG-7551 no longer available and I don't know what the problem was with the reported bug?

https://bugreports.qt.io/browse/QTBUG-7551

AliKet:
> > FYI: http://bugreports.qt.nokia.com/browse/QTBUG-7551 no longer available and I don't know what the problem was with the reported bug?
> 
> https://bugreports.qt.io/browse/QTBUG-7551

Thanks for the link, but AFAICS the reported bug has been closed as invalid a long time ago. So I don't think it is still relevant for us, is it ?

vadz:
Thanks, will push this soon.

PR #25270, opened 2025-03-25
Various warning fixes

paulcor:


src/osx/carbon/colordlgosx.mm
```diff
@@ -55,7 +55,7 @@ @implementation wxCPWCDelegate : NSObject
 
 - (id)init
 {
-    if ( self = [super init] )
+    if ((self = [super init]))
```
vadz, line 58: IMHO this looks bad/unidiomatic. If we can't suppress the warning, I'd rather write this as

```suggestion
    self = [super init];
    if (self)
```

src/osx/cocoa/dataview.mm
```diff
@@ -3454,17 +3452,13 @@ - (NSPoint)cellBaselineOffset
...
 
-    wxCheckBoxState checkedState ;
+    wxCheckBoxState checkedState = wxCHK_UNCHECKED;
     switch ( ObjectToLong(value) )
     {
         case 1:
             checkedState = wxCHK_CHECKED;
             break;
 
-        case 0:
```
vadz, line 3464: I'd leave this case for clarity, even if it does nothing useful any more.

vadz:
Thanks for all these fixes, it's definitely nice to avoid so many warnings, but I have no idea if all the casts added here are correct, e.g. if we really need to truncate or round when converting between floating point and integer types. Of course, the changes don't change the meaning of the code, so they can't introduce any new bugs, but, in theory, these warnings could be used to point out the places with suspicious conversions.

paulcor:
I'm dropping most of the `float-conversion` suppressions, for now. 

> in theory, these warnings could be used to point out the places with suspicious conversions

True, but realistically no one is ever going to check them, there are too many, and most have been sitting there unexamined for 10-20 years. But if all the current warnings were avoided, then the warnings could be enabled by default, and at least new code would get flagged and (hopefully) checked.

PR #25273, opened 2025-03-27
Fix wxDataViewCtrl scrollbar overlapping last item

KittenWo0f:
## Problem Description
In wxGTK3 applications using `wxDataViewCtrl`, the overlay scrollbar (a GTK3 feature where scrollbars appear as floating bars) overlaps the last row of content, making it partially obscured and hard to interact with.

### Affected Versions
- wxWidgets 3.2.x
- GTK3 environments (all major Linux distributions)

### Before Fix
image_2025-03-26_16-43-04
*Scrollbar covers bottom row content*

### After Fix
image_2025-03-26_16-41-24
*Scrollbar properly reserves space, leaving content fully visible*

## Root Cause Analysis
The issue occurs because:
1. GTK3 enables overlay scrollbars by default
2. These scrollbars render on top of content rather than reducing client area
3. wxDataViewCtrl's implementation doesn't account for this behavior

## Solution Implementation
The fix involves disabling GTK3's overlay scrolling feature for wxDataViewCtrl.

### DIFF
--- a/src/gtk/dataview.cpp
+++ b/src/gtk/dataview.cpp
@@ -4686,6 +4686,9 @@ bool wxDataViewCtrl::Create(wxWindow *parent,
     gtk_tree_view_set_rules_hint( GTK_TREE_VIEW(m_treeview), (style & wxDV_ROW_LINES) != 0 );
     wxGCC_WARNING_RESTORE()
 #endif
+#ifdef __WXGTK3__
+    gtk_scrolled_window_set_overlay_scrolling(GTK_SCROLLED_WINDOW(m_widget), FALSE);
+#endif
 
     gtk_scrolled_window_set_policy (GTK_SCROLLED_WINDOW (m_widget),
         GTK_POLICY_AUTOMATIC, GTK_POLICY_AUTOMATIC);


paulcor:
The scroll bar behavior is native. Generally the overlay scroll bars will disappear if the mouse is not moving. If they don't, it's likely an issue with your theme. There are several ways you can disable overlay scroll bars if you don't like them, but I'm not at all convinced we should be doing it for everyone.

KittenWo0f:
@paulcor, you are right, the scrollbar does disappear if you move the mouse away from the bottom of the table. The problem is that when working with the table, if you need to change the last row, half of the row is covered by the scrollbar. Because of this, when working with the last row you sometimes click on the scrollbar, which is annoying.

If somehow it is possible to disable the hovering scrollbar from the client part of the code, please tell me how to do it.

KittenWo0f:
Thanks for the answers. I found a way to disable the hover scroll bar for my program. It can be done by defining the environment variable “GTK_OVERLAY_SCROLLING=0”. For the whole system or for an individual program by creating a shortcut with this variable.

vadz:
Yes, I forgot about this, but using the env var indeed already provides a way to do what I suggested at the app level.

PR #25274, opened 2025-03-27
Use hues of blue for highlight colours in MSW dark mode

vadz:
Using grey colours meant that we didn't have almost any non-grey system colours in the dark mode, unlike in the light one, in which all the high/hot light colours were bluish.

Change these colours to (different) hues of blue in dark mode as well to make them more appropriate for actually highlighting something.

----

I'd like to merge this if nobody has any objections because otherwise we just have no nice highlight colours in dark mode under MSW.


vadz:
Merged in the commit above.

PR #25275, opened 2025-03-28
Add new flat AUI tab art

vadz:
As discussed in [this post](https://groups.google.com/g/wx-dev/c/mLeHQZ69ELs) the current situation with AUI art providers is not optimal. I've added a new art provider which looks a bit better (although I admit that it doesn't look that great, in spite of all my efforts), while being simpler (which hopefully means that maybe others will be able to improve it).

Here some screenshots of it under Linux in light image
and dark
image
modes and also under Windows, where I also show it in multiline mode:
image
image

I'd like to make this the default provider for 3.3.0, any thoughts?

Note that this includes #25274 which I had submitted as a separate PR as it's not logically part of this one, even if I did while working on this.


AliKet:
I tested this PR under Windows 10 + wxQt 6.8.2 (light & dark modes):

aui-light

aui-dark

vadz:
Thanks!

Is there some colour that would work better for highlighting than `HOTLIGHT` in wxQt? Or maybe it should be changed there, as in wxMSW dark mode? Because black definitely doesn't look like a great choice for highlighting, especially in dark mode.

AliKet:
`wxSYS_COLOUR_HOTLIGHT` is not implemented under `wxQt`:

> wxSYS_COLOUR_HOTLIGHT: Colour for a hyperlink or hot-tracked item

and here is a patch that fixes it:

```C++
diff --git a/src/qt/settings.cpp b/src/qt/settings.cpp
index 04cd02d6bc..661e59a995 100644
--- a/src/qt/settings.cpp
+++ b/src/qt/settings.cpp
@@ -120,6 +120,9 @@ wxColour wxSystemSettingsNative::GetColour(wxSystemColour index)
             break;

         case wxSYS_COLOUR_HOTLIGHT:
+            color = pal.color(QPalette::Link);
+            break;
+
         case wxSYS_COLOUR_GRADIENTACTIVECAPTION:
         case wxSYS_COLOUR_GRADIENTINACTIVECAPTION:
             // TODO


```

I have updated the images above to reflect this change.

vadz:
Looks much nicer, thanks! I'll push the wxQt fix separately.

Blake-Madden:
Looks good under Windows 11 (wxMSW).

PR #25276, opened 2025-03-28
Print fixes

vadz:
Restore compatibility with the existing code overriding `GetPageInfo()` by ignoring values returned by it (sic).

Also a couple of minor wxMSW-specific printing-related bug fixes/improvements.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -141,6 +141,10 @@ Changes in behaviour not resulting in compilation errors
 - wxAUI_MGR_HINT_FADE is not included in default wxAuiManager style any longer,
   please add it explicitly if you really want to use it.
 
+- Behaviour of wxPrintDialogData::SetAllPages() and SetSelection() has
+  changed when called with "false" argument, please review the documentation
+  and update your code if you called them with "false" (which is rarely done).
+
 Changes in behaviour which may result in build errors
 -----------------------------------------------------
 
--- a/include/wx/cmndata.h
+++ b/include/wx/cmndata.h
@@ -194,11 +194,10 @@ public:
     int GetMinPage() const { return m_printMinPage; }
     int GetMaxPage() const { return m_printMaxPage; }
     int GetNoCopies() const { return m_printNoCopies; }
-    bool GetAllPages() const { return m_printAllPages; }
-    bool GetSelection() const { return m_printSelection; }
-    bool GetCurrentPage() const { return m_printCurrentPage; }
-    bool GetSpecifiedPages() const
-        { return !(m_printAllPages || m_printSelection || m_printCurrentPage); }
+    bool GetAllPages() const { return m_printWhat == Print::AllPages; }
+    bool GetSelection() const { return m_printWhat == Print::Selection; }
+    bool GetCurrentPage() const { return m_printWhat == Print::CurrentPage; }
+    bool GetSpecifiedPages() const { return m_printWhat == Print::SpecifiedPages; }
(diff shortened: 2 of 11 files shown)


taler21:
Thanks, but "Print all pages" doesn't work correctly.

The cause is that `wxWindowsPrintDialog::ConvertFromNative( wxPrintDialogData &data )` sets a `wxPrintPageRange`

https://github.com/wxWidgets/wxWidgets/blob/12b29979d3ba9a5a08b2b3e44ad775ebe3ea18af/src/msw/printdlg.cpp#L1013

and `wxWindowsPrinter::Print()` uses it,

https://github.com/wxWidgets/wxWidgets/blob/12b29979d3ba9a5a08b2b3e44ad775ebe3ea18af/src/msw/printwin.cpp#L152-L156

both regardless of whether the user actually selected "Print pages" (`PD_PAGENUMS`) in the print dialog.

So I think there are two possible solutions:

1. `ConvertFromNative()` calls `wxPrintDialogData::SetPageRanges()` passing an empty `wxPrintPageRange` vector if `PD_PAGENUMS` was not selected by the user in the print dialog.
2. `wxWindowsPrinter::Print()` only initializes page ranges with the value from the dialog, if "Print pages" was selected by the user.
    ```cpp
    std::vector<wxPrintPageRange> pageRanges;
    if ( (!m_printDialogData.GetAllPages()) && (!m_printDialogData.GetSelection()) && (!m_printDialogData.GetCurrentPage()) )
        pageRanges = m_printDialogData.GetPageRanges();
    ```

vadz:
> Thanks, but "Print all pages" doesn't work correctly.

Oops, sorry for not noticing that this was broken now :-(

> So I think there are two possible solutions:

I think the solution (2) is clearly better, we should still copy the page ranges specified by user (if only to restore them the next time the dialog is used, although the sample doesn't do this correctly and reuses only `wxPrintData`, but not `wxPrintDialogData`, which contains this information, for subsequent printings), but we shouldn't use them if this is not the active choice. I've implemented this now and pushed this change, thanks! (I just couldn't avoid the urge to add a helper function because the test above looks really ugly).

<sub>The rest is just some musings about problems not directly related to this issue and can be skipped, but I'd like to write this down at least somewhere.</sub>

-----

While looking at this code, I got confused once again by thinking that it would be correct to call `GetPagesInfo()` *before* showing the dialog because I wrongly thought that the values returned from it should be used to initialize the dialog. And this is not the first time I was confused by this, and it's not just me because I had applied a patch from @troels-k back in 41514cc474 (Set up pages range in the wxMSW print dialog correctly., 2011-01-03), see #12819, only to revert it very quickly afterwards in ff4713940d (Revert "Set up pages range in the wxMSW print dialog correctly.", 2011-01-29)., see #12911. And, to be complete, except for less than one month between these commits, this had been like this even since d2b354f9c9 (Printing improvements: GetPageInfo() gets called after the DC has been set and after OnPreparePrinting() has been called [...], 2003-07-23) and `OnPreparePrinting()` was moved after `GetPageInfo()` even before it, in 1044a386f0 (Misc. 16-bit compilation fixes, 1999-09-05).

And this is correct because this function is supposed to be able to use the printing DC to paginate the document correctly, but this DC can't be created before showing the dialog. So the values returned by it can't be used to initialize the dialog, and only the values from `wxPrintDialogData` can be. And this is, again, done wrongly in the sample: it should preserve them between calls. I've changed this now and updated the documentation to make this more clear.

PR #25277, opened 2025-03-29
Fix gtk3 font scaling on Windows

MaartenBent:
Scaling issues seem to never end. When using gtk3 on Windows with a high-dpi screen, all controls that are custom drawn (treectrl, statusbar, dc) have the wrong font size.
Fix by not applying an additional `fontScalingFactor`, the font has already the correct size for the DPI of the display.

beforeafter

before2after2


vadz:

PR #25278, opened 2025-03-29
fix some event macro arguments in interface declarations

DietmarSchwertberger:
For these event macros there are mismatches in the argument list:
```
EVT_MAXIMIZE(func)
EVT_ACTIVATE(func)
EVT_FULLSCREEN(func)
```
They are correct in `event.h`, but have an additional `id` argument in `interface/toplevel.h` and `interface/window.h`.

(Two samples are using e.g. `EVT_ACTIVATE(MyChild::OnActivate)`, so I assume that `event.h` is correct. I did not actually build them, as I'm just parsing the interface files for the event informations.)

This PR fixes the mismatch and also removes trailing backslashes from the argument list in `interface/mediactrl.h`, e.g. `@event{EVT_MEDIA_LOADED(id\, func)}`.

### DIFF
--- a/interface/wx/mediactrl.h
+++ b/interface/wx/mediactrl.h
@@ -48,10 +48,10 @@ enum wxMediaCtrlPlayerControls
     Event wxMediaCtrl uses.
 
     @beginEventTable{wxMediaEvent}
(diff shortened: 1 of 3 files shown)


vadz:

PR #25279, opened 2025-03-29
Make handling of non-ASCII keys more consistent 

vadz:
This is a backwards incompatible change, but I think we can make it in 3.3.0 and it should make the behaviour as consistent and logical as possible: the key codes are now always ASCII characters (and not things like `ù` or `²` for example) that correspond either to the actual value of the key or to the value of the same physical key in US layout if it doesn't produce an ASCII character in the current layout.

Any comments from people, especially those using non-US Latin keyboard layouts (this PR doesn't change much for non-Latin layouts like Cyrillic ones) would be very welcome!


vadz:
@unxed Will you be able to test this?

taler21:
> Any comments from people, especially those using non-US Latin keyboard layouts

I tested it with a German keyboard layout and can confirm that it works as described.

However, I cannot see the benefit (for non-US Latin keyboard layouts).

Until now, with a German keyboard layout, the key code of the keys `Ä`, `Ö`, `Ü` corresponded to the physical keys (`ä`, `ö`, `ü`), i.e. the labeling of the keys visible to the user.
But with this PR, the key code of the keys `Ä`, `Ö`, `Ü` would be `'`, `;`, `[`.

In the previous scenario, as it works so far at least under Windows and Linux, an accelerator like `Strg+ü` would be possible (although not recommended) in a German application.
By the way, Visual Studio in the German edition actually uses such an accelerator (`Umschalt+Alt+ü`).

But if the key code represents the physical key of the US keyboard instead, it becomes useless.
Because an accelerator with a key that is labeled differently on the keyboard is not understandable for users.

However, I am not decidedly against this change, since we do not use exotic (non-ASCII) accelerators in our applications anyway.
But the previous behavior seems to make more sense to me (for non-US Latin keyboard layouts).

vadz:
> > Any comments from people, especially those using non-US Latin keyboard layouts
> 
> I tested it with a German keyboard layout and can confirm that it works as described.

Thanks for testing!
 
> However, I cannot see the benefit (for non-US Latin keyboard layouts).
> 
> Until now, with a German keyboard layout, the key code of the keys `Ä`, `Ö`, `Ü` corresponded to the physical keys (`ä`, `ö`, `ü`), i.e. the labeling of the keys visible to the user. But with this PR, the key code of the keys `Ä`, `Ö`, `Ü` would be `'`, `;`, `[`.

I had originally made this change for the French layout in which we were getting key events for `²` which seemed just weird. But I admit that this logic works much less well for the German layout, so I'll probably revert 34345e2343 (Change key code for non-ASCII Latin-1 characters in wxMSW, 2025-03-30)...

vadz:
P.S. One thing which _really_ doesn't make sense even for German is that the key code is `ä` and not `Ä`, which is inconsistent with all the other letters. OTOH we couldn't generate the uppercase version of `ß` which doesn't exist and I don't know if it's worth breaking compatibility for `ù`, `ä` etc...

unxed:
Will take look at it tomorrow

vadz:
> Will take look at it tomorrow

Thanks, but please note that I'll push a commit reverting the handling of high-ASCII characters soon (not sure if today, however).

vadz:
> > Will take look at it tomorrow
> 
> Thanks, but please note that I'll push a commit reverting the handling of high-ASCII characters soon (not sure if today, however).

I did do it today, finally. This should be ready for testing now.

vadz:
Sorry, I've just realized that I didn't update wxGTK behaviour to match the new (old) wxMSW behaviour, so I've just done it now. Hopefully you haven't tested under Linux yet, but if you did, please retest with the latest version because in it wxGTK also generates `ù` and `ä` etc key codes, as wxMSW does.

unxed:
today was too long day unfortunately. will try to test it tommorow or a day after

unxed:
And what about Macs? Unfortunately have no Mac to test

taler21:
I tested it again with the German keyboard layout and can confirm that it now works again as with 3.2.

> One thing which really doesn't make sense even for German is that the key code is `ä` and not `Ä`, which is inconsistent with all the other letters.

Yes, that's what I thought too, since the physical keys `Ä`, `Ö`, `Ü` on the keyboard are also labeled in capital letters.

> OTOH we couldn't generate the uppercase version of ß which doesn't exist and I don't know if it's worth breaking compatibility for `ù`, `ä` etc...

Probably not worth breaking compatibility. Especially since Visual Studio also uses a lowercase `ü` in its accelerator (`Umschalt+Alt+ü`).

vadz:
As I haven't had any other negative feedback, I'm going to merge this soon.

unxed:
You'll laugh, but I only just now found a free minute. I'll check it out, then, already in the master - better than nothing

PR #25281, opened 2025-03-30
Use wxMSW `wxFileDialog` in wxQt

vadz:
This replaces #24589.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -2387,7 +2387,9 @@ COND_PLATFORM_WIN32_1_QT_PLATFORM_HDR =  \
 	wx/msw/ole/safearray.h \
 	wx/msw/ole/oleutils.h \
 	wx/msw/ole/comimpl.h \
-	wx/msw/ole/uuid.h
+	wx/msw/ole/uuid.h \
+	wx/msw/filedlg.h \
+	wx/msw/dirdlg.h
 @COND_PLATFORM_WIN32_1@QT_PLATFORM_HDR = $(COND_PLATFORM_WIN32_1_QT_PLATFORM_HDR)
 COND_PLATFORM_MACOSX_1_GTK_PLATFORM_HDR =  \
 	wx/generic/caret.h \
@@ -5576,7 +5578,10 @@ COND_PLATFORM_WIN32_1___QT_PLATFORM_SRC_OBJECTS =  \
 	monodll_safearray.o \
 	monodll_msw_sound.o \
 	monodll_automtn.o \
-	monodll_msw_power.o
+	monodll_msw_power.o \
+	monodll_msw_filedlg.o \
+	monodll_msw_dirdlg.o \
+	monodll_utilswin.o
 @COND_PLATFORM_WIN32_1@__QT_PLATFORM_SRC_OBJECTS = $(COND_PLATFORM_WIN32_1___QT_PLATFORM_SRC_OBJECTS)
 COND_TOOLKIT_DFB___LOWLEVEL_SRC_OBJECTS_1 =  \
 	monodll_fontmgrcmn.o \
@@ -7339,7 +7344,10 @@ COND_PLATFORM_WIN32_1___QT_PLATFORM_SRC_OBJECTS_1 =  \
 	monolib_safearray.o \
 	monolib_msw_sound.o \
 	monolib_automtn.o \
-	monolib_msw_power.o
+	monolib_msw_power.o \
+	monolib_msw_filedlg.o \
+	monolib_msw_dirdlg.o \
+	monolib_utilswin.o
 @COND_PLATFORM_WIN32_1@__QT_PLATFORM_SRC_OBJECTS_1 = $(COND_PLATFORM_WIN32_1___QT_PLATFORM_SRC_OBJECTS_1)
 COND_TOOLKIT_DFB___LOWLEVEL_SRC_OBJECTS_3 =  \
 	monolib_fontmgrcmn.o \
@@ -9248,7 +9256,10 @@ COND_PLATFORM_WIN32_1___QT_PLATFORM_SRC_OBJECTS_2 =  \
 	coredll_safearray.o \
 	coredll_msw_sound.o \
 	coredll_automtn.o \
-	coredll_msw_power.o
+	coredll_msw_power.o \
+	coredll_msw_filedlg.o \
+	coredll_msw_dirdlg.o \
+	coredll_utilswin.o
 @COND_PLATFORM_WIN32_1@__QT_PLATFORM_SRC_OBJECTS_2 = $(COND_PLATFORM_WIN32_1___QT_PLATFORM_SRC_OBJECTS_2)
 COND_TOOLKIT_DFB___LOWLEVEL_SRC_OBJECTS_5 =  \
 	coredll_fontmgrcmn.o \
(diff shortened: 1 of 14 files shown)


vadz:
cc @AliKet @dsa-t

AliKet:
After applying the follwing patch:
```C++
diff --git a/include/wx/msw/filedlg.h b/include/wx/msw/filedlg.h
index 1c0f1c5c4d..22a31578c2 100644
--- a/include/wx/msw/filedlg.h
+++ b/include/wx/msw/filedlg.h
@@ -87,6 +87,10 @@ private:
     WXHWND GetHWND() const { return m_hWnd; }
     void SetHWND(WXHWND hwnd) { m_hWnd = hwnd; }
 
+    // We need to reparent child to m_hWnd if the old customization API is used.
+    // This is not called when the new customization API is used.
+    void AddChild(wxWindowBase* child) override;
+
     WXHWND m_hWnd = nullptr;
 #endif // __WXQT__
 
diff --git a/src/msw/filedlg.cpp b/src/msw/filedlg.cpp
index 2fffaebe00..e6f29c83b5 100644
--- a/src/msw/filedlg.cpp
+++ b/src/msw/filedlg.cpp
@@ -1750,4 +1750,21 @@ int wxFileDialog::ShowIFileDialog(WXHWND hWndParent)
 
 #endif // wxUSE_IFILEOPENDIALOG
 
+#ifdef __WXQT__
+void wxFileDialog::AddChild(wxWindowBase* child)
+{
+    if ( ::IsWindow(m_hWnd) && !GetHandle() )
+    {
+        // This is a dummy widget to be able to destroy child in wxFileDialog dtor.
+        m_qtWindow = new QWidget(GetParent()->GetHandle());
+
+        auto winId = child->GetHandle()->winId();
+        child->GetHandle()->setWindowFlag(Qt::FramelessWindowHint);
+        ::SetParent(reinterpret_cast<WXHWND>(winId), m_hWnd);
+    }
+
+    wxWindowBase::AddChild( child );
+}
+#endif // __WXQT__
+
 #endif // wxUSE_FILEDLG
diff --git a/src/qt/toplevel.cpp b/src/qt/toplevel.cpp
index 2c3fed11a5..8e8e687634 100644
--- a/src/qt/toplevel.cpp
+++ b/src/qt/toplevel.cpp
@@ -92,12 +92,12 @@ void wxTopLevelWindowQt::Iconize(bool iconize )
 
 bool wxTopLevelWindowQt::IsMaximized() const
 {
-    return GetHandle()->isMaximized();
+    return GetHandle() && GetHandle()->isMaximized();
 }
 
 bool wxTopLevelWindowQt::IsIconized() const
 {
-    return GetHandle()->isMinimized();
+    return GetHandle() && GetHandle()->isMinimized();
 }
 
 
index 9fa13baf54..7d97113757 100644
--- a/src/qt/window.cpp
+++ b/src/qt/window.cpp
@@ -1063,8 +1063,11 @@ void wxWindowQt::DoThaw()
 void wxWindowQt::DoGetPosition(int *x, int *y) const
 {
     QWidget *qtWidget = GetHandle();
-    *x = qtWidget->x();
-    *y = qtWidget->y();
+    if ( qtWidget )
+    {
+        *x = qtWidget->x();
+        *y = qtWidget->y();
+    }
 }
 
 namespace
@@ -1089,10 +1092,13 @@ inline wxSize wxQtSetClientSize(QWidget* qtWidget, int width, int height)
 
 void wxWindowQt::DoGetSize(int *width, int *height) const
 {
-    const QSize size =  GetHandle()->frameSize();
+    if ( GetHandle() )
+    {
+        const QSize size = GetHandle()->frameSize();
 
-    if (width)  *width = size.width();
-    if (height) *height = size.height();
+        if (width)  *width = size.width();
+        if (height) *height = size.height();
+    }
 }
 
 
@@ -1118,7 +1124,9 @@ void wxWindowQt::DoSetSize(int x, int y, int width, int height, int sizeFlags )
             height = BEST_SIZE.y;
     }
 
-    if ( !GetHandle()->isVisible() && QtGetClientWidget() != GetHandle() )
+    const auto qtWidget = GetHandle();
+
+    if (qtWidget && !qtWidget->isVisible() && QtGetClientWidget() != qtWidget)
     {
         if ( width != -1 && height != -1 )
         {
```
everything works correctly for me when testing dialogs sample.


Please test the dialogs sample with/without this patch to see the problem when using the old customization API.

Thanks for your work on this.

AliKet:
FYI, Qt has [QWindow::fromWinId](https://doc.qt.io/qt-6/qwindow.html#fromWinId) and [QWidget::createWindowContainer](https://doc.qt.io/qt-6/qwidget.html#createWindowContainer) which can be used to embed native windows into a QWidget.
But here we need to do the opposite, i.e. embed our `QWidget`s  into the native window (HWND), which is done by these lines:
```C++
        auto winId = child->GetHandle()->winId();
        child->GetHandle()->setWindowFlag(Qt::FramelessWindowHint);
        ::SetParent(reinterpret_cast<WXHWND>(winId), m_hWnd);
```
There is already [QWinWidget](https://github.com/qtproject/qt-solutions/blob/master/qtwinmigrate/src/qwinwidget.h) doing exactly this, but I don't think we need to complicate wxWidgets just to support this specific use case!

AliKet:
This is the open dialog using the old customization API (with the patch above applied)

opendlg

Without the patch, the custom control would be displayed in a separate window.

AliKet:
Please note that the above patch is modified to fix a memory leak and also removes the following message (reported by Qt):

> External WM_DESTROY received for  QWidgetWindow(0x20288844410, name="QWidgetClassWindow") , parent:  QWindow(0x0) , transient parent:  QWindow(0x0)

AliKet:
> I'm not really sure we want to add these hacks to `wxWindow` itself just to let the old customization API to work, is it really worth it? I'd just merge the existing PR to allow the new customization API to work for now, unless you feel strongly about adding these changes too.
> 

It's okay to only support the new customization API under `wxQt`. I just thought the old one should also be supported, especially if `wxIFileDialog::CanBeUsedWithAnOwner()` returns false, in which case the custom controls will appear in a separate window from the dialog itself.

> Or are any of them needed even for the new API to work? Because I tested it and it seemed to work fine without them for me, but I did it with Qt 5.something, so maybe it's different with 6?

Both Qt versions (5&6) crash with something like:
```
wxqt330ud_core_vc_x64_custom.dll!wxTopLevelWindowQt::IsMaximized() Line 96	C++
wxqt330ud_core_vc_x64_custom.dll!wxTopLevelWindowBase::DoCentre(int dir) Line 246	C++
wxqt330ud_core_vc_x64_custom.dll!wxFileDialog::MSWOnInitDone(HWND__ * hDlg) Line 1144	C++
wxqt330ud_core_vc_x64_custom.dll!wxFileDialogMSWData::OnTypeChange(IFileDialog * pfd) Line 823	C++
...
wxqt330ud_core_vc_x64_custom.dll!wxMSWImpl::wxIFileDialog::Show(HWND__ * owner, int options, wxArrayString * pathsOut, wxString * pathOut) Line 335	C++
wxqt330ud_core_vc_x64_custom.dll!wxFileDialog::ShowIFileDialog(HWND__ * hWndParent) Line 1713	C++
wxqt330ud_core_vc_x64_custom.dll!wxFileDialog::ShowModal() Line 1273	C++
```
because `GetHandle()` always returns `nullptr` as it doesn't create any. I also tested `wxMSW` and it returns `nullptr` too when `GetHandle()` is called, it just handles it properly.

vadz:
> Both Qt versions (5&6) crash with something like:

Oops, sorry, I've only tested with ad hoc code in the minimal sample (because `dialogs` one is not built by default when using CMake...) and so didn't have `dialog.CentreOnParent();` line in my code. And with it it indeed crashes :-(

I've now disabled this code for wxQt, it's ugly, but it doesn't make much sense to do all this when it has no chance of actually working there. Does this look good (enough to merge) to you now?

Thanks!

PR #25282, opened 2025-03-30
Fix wxTreeCtrl indent regression

MaartenBent:
The changes from https://github.com/wxWidgets/wxWidgets/issues/24640 don't seem to work when the size of the images in the image list is too small.
The win32 control seems to expect an image size of at least `FromDIP(16)`px. When the images are smaller, for example just `16`px, the extra indent causes the expand/collapse hitbox to be shifted to the right and not overlap with the twist (or plus-min) button.
Fix this by reducing the indent when the images are too small.

Note: this also happens when manually applying a large `SetIndent`, can't do anything about that.

In red I marked where the hitbox is approximately (at 150% DPI):
before after

I also updated the widgets sample to use a `wxBitmapBundle` for the images, instead of using hardcoded 16x16px.
after2

-----

I made the indent and spacing DPI independent, so get/set return the same values no matter the DPI. Is this desired or should it be possible to keep pixel-perfect control?

-----

3.2 has this problem too, but this fix cannot easily be backported, because adding functions and variables breaks ABI.  One way would be to fix the indent when setting the image list, taking into account corrections for any previous image lists. I can create a PR with this, but maybe someone has a better idea. (Like documenting the image list should have at least size `FromDIP(16)`.)
```diff
diff --git "a/src/msw/treectrl.cpp" "b/src/msw/treectrl.cpp"
index 8baabda684..7f9f16b13b 100644
--- "a/src/msw/treectrl.cpp"
+++ "b/src/msw/treectrl.cpp"
@@ -922,10 +922,33 @@ void wxTreeCtrl::SetIndent(unsigned int indent)
 
 void wxTreeCtrl::SetAnyImageList(wxImageList *imageList, int which)
 {
+    int indent = GetIndent();
+
+    HIMAGELIST oldImgList = TreeView_GetImageList(GetHwnd(), which);
+    if (oldImgList != nullptr)
+    {
+        int oldWidth = 0, oldHeight = 0;
+        if (ImageList_GetIconSize(oldImgList, &oldWidth, &oldHeight))
+        {
+            int diff = oldWidth - FromDIP(16);
+            if (diff < 0)
+                indent -= diff;
+        }
+    }
+
     // no error return
     (void) TreeView_SetImageList(GetHwnd(),
                                  imageList ? imageList->GetHIMAGELIST() : 0,
                                  which);
+
+    wxImageList* newImgList = GetImageList();
+    if (newImgList != nullptr)
+    {
+        int diff = newImgList->GetSize().GetWidth() - FromDIP(16);
+        if (diff < 0)
+            indent += diff;
+    }
+    SetIndent(indent);
 }
 
 void wxTreeCtrl::SetImageList(wxImageList *imageList)
```

### DIFF
--- a/samples/widgets/widgets.cpp
+++ b/samples/widgets/widgets.cpp
@@ -631,10 +631,8 @@ WidgetsFrame::WidgetsFrame(const wxString& title)
 
 void WidgetsFrame::InitBook()
 {
-    wxImageList *imageList = new wxImageList(ICON_SIZE, ICON_SIZE);
-
-    wxImage img(sample_xpm);
-    imageList->Add(wxBitmap(img.Scale(ICON_SIZE, ICON_SIZE)));
+    wxVector<wxBitmapBundle> imageList;
+    imageList.push_back(WidgetsPage::CreateBitmapBundle(sample_xpm));
 
 #if !USE_TREEBOOK
     WidgetsBookCtrl *books[MAX_PAGES];
@@ -702,7 +700,7 @@ void WidgetsFrame::InitBook()
 
     GetMenuBar()->Append(menuPages, "&Page");
 
-    m_book->AssignImageList(imageList);
+    m_book->SetImages(imageList);
 
     for ( cat = 0; cat < MAX_PAGES; cat++ )
     {
@@ -746,6 +744,7 @@ void WidgetsFrame::InitBook()
 
     wxTreeItemIdValue cookie;
     tree->EnsureVisible(tree->GetFirstChild(tree->GetRootItem(), cookie));
+    tree->SetMinSize(wxSize(tree->GetBestSize().GetWidth() * 1.1, wxDefaultCoord));
 #else
     if ( !pageSet || !m_book->GetCurrentPage() )
     {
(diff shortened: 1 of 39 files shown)


samples/widgets/widgets.cpp
```diff
@@ -1363,16 +1362,57 @@ WidgetsPageInfo::WidgetsPageInfo(Constructor ctor, const wxString& label, int ca
 // WidgetsPage
 // ----------------------------------------------------------------------------
 
+namespace
+{
+    class FixedSizeImpl : public wxBitmapBundleImpl
```
vadz, line 1367: Sorry, I don't understand why do we need this, i.e. how is it better than just using `wxBitmapBundleImplSet` as is done by default?

vadz (commented):
Thanks, I guess it indeed makes more sense to specify the indent in DIPs than in physical pixels, so let's so do it like this — even though this will be yet another exception to the general rule that all sizers use pixels.

No real comments otherwise, thanks for fixing this!

samples/widgets/widgets.cpp
MaartenBent: Because the source images (`.xpm`) are 32x32px but the default size of the bundle should be 16x16px.

With this the smallest size of the bundle is 32x32px, but `wxTreeCtrl` should have 16px as default.
```c++
wxBitmapBundle::FromImage(wxImage(icon));
```

With this the default size is 16px, but it doesn't have high quality images (32px) for 200%. It just scales the 16px up.
```c++
wxBitmapBundle::FromImage(wxImage(icon).Rescale(wxSize(16,16)))
```

With this it has 16px as default, and high quality 32px images. But I use 150% DPI and then it doesn't have an exact match and uses 16px instead of 24px.
```c++
    std::vector<wxBitmap> images;
    images.push_back(wxImage(icon).Scale(wxSize(16, 16)));
    images.push_back(wxImage(icon));
    wxBitmapBundle::FromBitmaps(images);
```
I can add another bitmap for the current DPI. But then it is still possible there are displays with a different DPI, or the DPI of the current display is changed.
```c++
    images.push_back(wxImage(icon).Scale(FromDIP(wxSize(16, 16))));
```

So in the end I copied the `FixedSizeImpl` that is also used in the [treectrl sample](https://github.com/wxWidgets/wxWidgets/blob/master/samples/treectrl/treetest.cpp#L969).

vadz: Ah, thanks, I remember that this is why I wrote this comment there:

```cpp
    // Make a vector of bundles corresponding to the icons. We use a custom
    // bundle implementation here as we always scale the icons, even at 100%
    // DPI, to ensure they are of the desired size.
```

If we need it in more than one place, I wonder if we should make it public class. Or if, perhaps, we should somehow make the default implementation deal with it correctly, maybe by adding an option size parameter to `FromBitmap()` or adding a new `FromBitmapWithNonDefaultSize()` (but, as always, with a better name)?

MaartenBent: If extending the default implementation, it would need two things:
1) A way to specify the default size if it is different from the provided (first) bitmap.
2) A way to override/change `wxBitmapBundleImpl::DoGetPreferredSize` so it always returns `GetDefaultSize() * scaleTarget`.

So it's probably better/easier to make this `FixedSizeImpl` publicly available.

Issue #24694, opened 2024-07-11
Memory leaks from wxInitData::InitArgvA on wxMSW


dsa-t:
### Description

Visual Studio reports memory leaks when closing a program.

```
Detected memory leaks!
Dumping objects ->
{5856} normal block at 0x0000028E41625540, 67 bytes long.
 Data: <C:\Build\wxWidge> 43 3A 5C 42 75 69 6C 64 5C 77 78 57 69 64 67 65 
C:\Build\wxWidgets\src\common\init.cpp(191) : {5855} normal block at 0x0000028E4161C560, 16 bytes long.
 Data: <@UbA            > 40 55 62 41 8E 02 00 00 00 00 00 00 00 00 00 00 
Object dump complete.
The program '[29996] minimal.exe' has exited with code 0 (0x0).
```

---

Allocation call stack for 0x0000028E41625540:

```
wxbase330ud_vc_x64_custom.dll!wxMBConv::cWC2MB() - Line 81
ucrtbased.dll!heap_alloc_dbg_internal() - Line 359
ucrtbased.dll!heap_alloc_dbg() - Line 450
ucrtbased.dll!_malloc_dbg() - Line 496
ucrtbased.dll!malloc() - Line 27
wxbase330ud_vc_x64_custom.dll!wxCharTypeBuffer<char>::wxCharTypeBuffer<char>() - Line 330
wxbase330ud_vc_x64_custom.dll!wxCharBuffer::wxCharBuffer() - Line 428
wxbase330ud_vc_x64_custom.dll!wxMBConv::DoConvertWC2MB() - Line 509
wxbase330ud_vc_x64_custom.dll!wxInitData::InitArgvA() - Line 200
wxbase330ud_vc_x64_custom.dll!wxInitData::MSWInitialize() - Line 227
wxmsw330ud_core_vc_x64_custom.dll!wxMSWEntryCommon() - Line 207
wxmsw330ud_core_vc_x64_custom.dll!wxEntry() - Line 227
minimal.exe!WinMain() - Line 105
minimal.exe!invoke_main() - Line 107
minimal.exe!__scrt_common_main_seh() - Line 288
minimal.exe!__scrt_common_main() - Line 331
minimal.exe!WinMainCRTStartup() - Line 17
kernel32.dll!BaseThreadInitThunk() - Line 0

```

[...]

PR #25283, opened 2025-03-31
Fix leak of wxInitData::argv in wxMSW

vadz:
When not using MSWInitialize(), e.g. in console MSW applications, we need to free argv elements ourselves as they're not the same as argvMSW and so are not freed when freeing it.

This is different from the leak fixed in #24694 but is in the same code.

----

cc @MaartenBent just in case I'm missing something here


vadz:
Thanks for testing!

Issue #25288, opened 2025-04-03
Strategy for prebuilt windows binaries for 3.3


PBfordev:
Seeing as the first release in the 3.3 branch comes close, I think it could be a good idea to discuss possible changes in the binary release process, possibly simplifying it (besides going from supporting five MSVS versions to one).

Even if no one asked, here are my two cents on the topic.

Firstly, is there a reason to still provide 32-bit builds in 2025 and beyond, targeting Windows 7 and newer? Download stats for 3.2.7 on the website show similar numbers as for the 64-bit ones; however, seeing as they all hover around 500 for basically all compilers and architectures, I wonder how reliable/representative those are...

Secondly (and more importantly) about MinGW/GCC-based binaries. wxWidgets 3.2.7 offers prebuilt binaries for five toolchains (GCC is currently at v14.2):
1. MSYS MinGW-w64 13.2
2. MSYS MinGW-w64 12.2
3. TDM-GCC 10.3
4. MinGW-w64 8.1
5. MinGW-w64 7.3

As for 1-2 (GCC vs clang debate aside), IMO, if you are using MSYS2 and don't want to build a library by yourself, you will probably use its package manager to get it. MSYS2 provided also "unstable" 3.1.x releases via pacman. Moreover, we are playing a catch-up game with MSYS2 releases of GCC. Last but not least: Incompatible msvcrt vs ucrt library builds with the latter likely being the better choice now. Is there really a point in providing MSYS2-based binaries?

As for 3-5, these compiler packages have been long abandoned. AFAIK, there are currently two maintained standalone GCC-based toolchains (nuwen had last release in 01/2024): [WinLibs](https://winlibs.com) and [niXman](https://github.com/niXman/mingw-builds-binaries). I am not sure if the two have enough traction to support them.

Of course, in addition to building a library by ourselves or getting its binary from the library maintainers, we now have even on Windows popular and convenient package managers such as MSYS2 and vcpkg, covering all three compilers and probably decreasing the demand on the "official" binaries.

I am interested in hearing your thoughts and/or counter-arguments to my points?

vadz:
Thanks for raising this question, this indeed needs to be decided, although mostly by the current binary maintainers, so I'd like to ping @doscott and @XPsoud (sorry if you have already noticed this).

1. I agree that it should be safe to drop 32 bit binaries, people shouldn't be using them in 2025.
[...]

PR #25289, opened 2025-04-03
Purge mentions of no longer supported MSVC versions from docs

PBfordev:
Remove the note from wxEvtHandler::Bind().

Update "How to use wxMSW binaries" guide.

----------
The  "How to use wxMSW binaries" guide is heavily MSVS-focused. I also don't really see the purpose of the "Building samples with nmake" there, the samples sources will not be obtained with the binaries.

See also #25288.

BTW, the sentence in the binaries guide reading
> You also need to decide whether you use the `x64` files for 64-bit development
or the ones without this suffix for the **still more common** 32-bit builds.

should probably be changed in the 3.2 branch to (I don't trust myself to make a PR against 3.2 branch)
> You also need to decide whether you use the `x64` files for 64-bit development
or the ones without this suffix for the 32-bit builds.

### DIFF
--- a/docs/msw/binaries.md
+++ b/docs/msw/binaries.md
@@ -7,35 +7,31 @@ Supported Compilers
 -------------------
 We provide pre-built binary files for the following compilers:
 
-* Microsoft Visual C++ compiler versions 9.0, 10.0, 11.0, 12.0, 14.0, 14.1 and 14.2
-  (corresponding to marketing product names of Microsoft Visual Studio 2008, 2010, 2012, 2013, 2015, 2017 and 2019 respectively). Please note that MSVC 14.x versions are ABI-compatible and the same set of binaries is used for all of them.
+* Microsoft Visual C++ compiler versions 14.0, 14.1, 14.2 and 14.3
+  (corresponding to marketing product names of Microsoft Visual Studio 2015, 2017, 2019 and 2022 respectively).
+  Please note that MSVC 14.x versions are ABI-compatible and the same set of binaries is used for all of them.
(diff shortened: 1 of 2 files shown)


docs/msw/binaries.md
```diff
@@ -7,35 +7,31 @@ Supported Compilers
...
-determining the combination of suffixes you need, you should download the
-"Dev" and the "ReleaseDLL" files in addition to the "Headers" one above,
-e.g. for 32-bit MSVS 2017 development you need
-`wxMSW-3.2.0_vc14x_Dev.7z` and `wxMSW-3.2.0_vc14x_ReleaseDLL.7z`.
+`wxWidgets-3.3.0-headers.7z` one but the "Dev" and the "ReleaseDLL"
+files depend on your compiler and CPU architecture.
+For example, for MSVS 2015/2017/2019/2022 and 64-bit (x86_64) architecture
+you will need `wxMSW-3.3.0_vc14x_x64_Dev.7z` and `wxMSW-3.3.0_vc14x_x64_ReleaseDLL.7z`.
+If you are using MinGW-w64 v8.1 and 32-bit (x86) architecture, you will need
+`wxMSW-3.3.0_gcc810_Dev.7z` and `wxMSW-3.3.0_gcc810_ReleaseDLL.7z` (32-bit
```
vadz, line 26: I wonder if we should start including `x86` in the names of these archives? Because I suspect that quite a few people might be downloading them by mistake, as the number of downloads is almost the same as for `x64` ones, which is definitely not what I'd expect.

Of course, this assumes we don't just drop them entirely.

vadz (commented):
Thanks for updating this!

I'll merge it, but perhaps we should wait for the results of the discussion in the other issue first.

PR #25292, opened 2025-04-03
Fixes for wxGTK on macOS – 2 (fix condition for X11)

barracuda156:
Cocoa builds are already excluded by GTK requirement in the macro, but for GTK builds on Apple we need X11.

### DIFF
--- a/build/cmake/toolkit.cmake
+++ b/build/cmake/toolkit.cmake
@@ -119,7 +119,7 @@ endif()
 # support, but not for Wayland-only GTK (necessarily 3 or later), which is why
 # we have to do this after find_package(GTKx) above, as this is what sets
 # wxHAVE_GDK_X11.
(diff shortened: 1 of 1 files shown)


build/cmake/toolkit.cmake
```diff
@@ -119,7 +119,7 @@ endif()
 # support, but not for Wayland-only GTK (necessarily 3 or later), which is why
 # we have to do this after find_package(GTKx) above, as this is what sets
 # wxHAVE_GDK_X11.
-if(UNIX AND NOT APPLE AND NOT WIN32 AND (WXX11 OR WXGTK2 OR (WXGTK AND wxHAVE_GDK_X11)))
+if(UNIX AND AND NOT WIN32 AND (WXX11 OR WXGTK2 OR (WXGTK AND wxHAVE_GDK_X11)))
```
swt2c, line 122: Looks like you've got a double AND here...

vadz:
Thanks, this looks fine to me, so I'll merge it if @MaartenBent has no objections.

MaartenBent (approved):

PR #25294, opened 2025-04-03
Fixes for wxGTK on macOS – 4 (minor fixes to macOS macros in generic files)

barracuda156:


include/wx/defs.h
```diff
@@ -2699,7 +2697,7 @@ typedef HIShapeRef WXHRGN;
 
 #endif // __WXMAC__
 
-#if defined(__WXMAC__)
+#if defined(__WXMAC__) || defined(__OBJC__)
```
vadz, line 2700: This doesn't make sense, there are tests for `__OBJC__` inside this branch.

Maybe we need to test for `__APPLE__` or `__DARWIN__` here?

include/wx/defs.h
```diff
@@ -2621,7 +2617,9 @@ typedef unsigned short  WXWORD;
 
 typedef WX_OPAQUE_TYPE(PicHandle ) * WXHMETAFILE ;
 
+#if !defined(__WXX11__) && !defined(__WXGTK__)
 typedef void*       WXDisplay;
+#endif
```
vadz, line 2622: Maybe we should just remove this one? I don't understand why would we need to have `WXDisplay` defined when building wxOSX, it should only be used when using X or GTK.

vadz (commented):
Sorry, the changes in `wx/defs.h` seem suspicious to me.

The other one is good, although I'm not sure if `__APPLE__` or `__DARWIN__` is the right symbol to use. AFAICS currently they're equivalent.

include/wx/defs.h
barracuda156: Your suggestion sounds good and should work. I will try it.

include/wx/defs.h
barracuda156: I am fine with that. With wxGTK having it here creates a conflict, so removing it unconditionally works.

include/wx/defs.h
barracuda156: Fixed

include/wx/defs.h
barracuda156: Removed this one

include/wx/defs.h
```diff
@@ -2699,7 +2693,7 @@ typedef HIShapeRef WXHRGN;
 
 #endif // __WXMAC__
 
-#if defined(__WXMAC__)
+#if defined(__WXMAC__) || defined(__APPLE__)
```
vadz, line 2696: I think this should be just
```suggestion
#ifdef __DARWIN__
```

We don't need `||` (when `__WXMAC__` is defined, `__APPLE__` definitely is too) and using `__DARWIN__` is more consistent with the test above (I think it and `__APPLE__` are synonyms nowadays).

Could you please confirm that it works for you with this change?

barracuda156: Sure, I will verify this tomorrow once back to the machine. Together with an attempt to build Cocoa stdpaths.

P. S. On macOS `__APPLE__` covers all cases, but I have no idea what iOS uses (and whether it should be considered in any given case).

barracuda156:
@vadz This should be good, once CI pass.

PR #25297, opened 2025-04-05
Make MSW buttons more closely match common controls

vslavik:
This is apparently a never-ending story, with attempts to follow MSFT docs (which is wrong), using Vista+ API to get the size (also wrong), eventually settling on doing it ourselves and matching native controls. For reference: #18528, #24283, #24285, 48f8e95 ("Fix wxButton::GetDefaultSize() in high DPI under MSW"), c6d2f6d ("Really fix the standard button size in high DPI under MSW").

I wanted to call this PR "Hopefully finally actually fix the standard button size under MSW". I no longer hold that hope, but I do think doing this, or something similar, is the least bad thing.

In #24283, careful measurements were taken, a matching expression arrived at, and it was a joyous occasion and much optimism was expressed. The only problem was the methodology:

> **the button sizes in Explorer**

It turns out that Explorer uses *atypically small buttons* (...which also have slightly different sizes among themselves). Compare this (from wx code, matching Explorer, at 200%):

Untitled

to this (common file dialog at 200%):

Screenshot 2025-04-05 at 12 49 23

(This also shows why I care: win32 buttons have vertically misaligned labels and the smaller the button is, the more noticeable that is.)

So I went ahead to measure it, expecting to find the correct, non-Explorer size. That wasn't meant to be, see the following table. Note that I didn't bother to measure at >200% for reasons that will soon become apparent. This was done on current Windows 11, but I spot-checked some of the values against Windows 10 and they matched. Notice the sizes are 2px off  from #24283 measurements — I measured the visible size, i.e. including the blue border, but the actual control size is apparently indeed +2px. I measured "Explorer" column using the OK button in Properties; some other buttons there are slightly different, and some values were off compared to #24283; not sure why. I included native WinUI buttons for illustration, they appear to be simply 32pt.

Without further ado, let me present the results in pixels:

| Scale | Explorer / wx | *File dialog* | *Message box* | Run… dlg | WinUI | 24×scale (proposed) |
|----|-----------|-------------|--------------|------|-------|-----------|
| 100%          | 21          | 24           | 21   | 24    | 32        | 24.0 |
| 125%          | 26          | 31           | 27   | 33    | -         | 30.0 |
| 150%          | 33          | 37           | 35   | 42    | -         | 36.0 |
| 175%          | 40          | 44           | 41   | 51    | -         | 42.0 |
| 200%          | 42          | 50           | 46   | 54    | 64        | 48.0 |

That is to say, I found no less than four different "standard" sizes (not counting Explorer's deviations) and two different sizes within common dialogs!

I'm not gonna lie. That's when I threw in the towel and just applied this patch to Poedit to get more reasonable, but not-quite-native-sized buttons. It still doesn't match either file dialogs or message boxes,  but there's no hope of matching *both*, so picking a value that is easy to work with and is somewhere between the two seems reasonable. The difference is not going to be very noticeable at HiDPI.

I'm not sure if anything better can be done (probably) or what (no clue), but IMHO the common dialogs should be a reference point, not Explorer, because the are the ones commonly presented in the context of the same application as wxButton is.

P.S. wx-3.2 form needs to touch 2 places: https://github.com/poedit/wxWidgets/commit/6a2f29094de7c6b55e012558ec9d6353c92a1677

### DIFF
--- a/src/msw/button.cpp
+++ b/src/msw/button.cpp
@@ -174,7 +174,7 @@ wxSize wxButtonBase::GetDefaultSize(wxWindow* win)
         // part of the button are not scaled correctly in higher than normal
         // DPI, so add them without scaling.
         s_sizeBtn.SetAtNewDPI(
-            wxWindow::FromDIP(wxSize(73, 21), win) + wxSize(2, 2)
+            wxWindow::FromDIP(wxSize(86, 24), win) + wxSize(2, 2)
         );
     }
 


vadz:
I think that Explorer sizes are the "historically correct" ones because this is what you find in the other places where old code is used, e.g. the standard "Find", "Font", "About" dialog etc. But I also agree that they look too thin compared to the file dialog buttons. OTOH those look much too big to me :-(

So getting close to them may be good, but with this change we're now different from everything else on the system.

And I still wonder how does Microsoft arrive at the values in the table. For the message box the simplest formula I can find is `round(22*S)+3*round(S-1)-1` (where `S` is scale), which just might be plausible but for the file dialog ones I can't find anything.

I reserve the right to change my mind once again, but I think we should use message box button sizes. This keeps the same size at 100% DPI which is important because adding extra 3px to layouts with many buttons may have noticeable consequences and so changing this in 3.2 would be a bad idea IMO. And this brings us closer to the file dialog buttons while still being consistent with at least something else. What do you think?

cc @marekr just in case you have any thoughts about this

vslavik:
> I think that Explorer sizes are the "historically correct" 

Seems likely to me, with the caveat that "historically" implies "predating non-rare HiDPI use". 

> OTOH those look much too big to me :-(

Could it be that you're unusually exposed to seeing `wxButton`...? To me, the slightly larger size (file dialog, not message box) looks better on modern Windows even at 100% DPI.

> And I still wonder how does Microsoft arrive at the values in the table. For the message box the simplest formula I can find is round(22*S)+3*round(S-1)-1 (where S is scale)

Could the message box one be calculated in dialog units? It likely is some old code... The formula above definitely feels like over-fitting to me.

> I reserve the right to change my mind once again, but I think we should use message box button sizes. [...] What do you think?

I have no problem with that. An argument for using file dialog size would be that it is probably (?) more commonly seen by users than message boxes, so should be prioritized for aiming consistency at, by I get the high-density UI argument.

marekr:
> cc @marekr just in case you have any thoughts about this

A px shouldn't be a big deal as long as all wxButtons take the new numbers. Otherwise we'll have mismatches if they don't actually follow GetDefaultSize

vadz:
> > I think that Explorer sizes are the "historically correct"
> 
> Seems likely to me, with the caveat that "historically" implies "predating non-rare HiDPI use".

Yes, exactly.

> > OTOH those look much too big to me :-(
> 
> Could it be that you're unusually exposed to seeing `wxButton`...? To me, the slightly larger size (file dialog, not message box) looks better on modern Windows even at 100% DPI.

It's probably just because I got used to the "historical" buttons. But this is all subjective anyhow, so I don't attach much importance to it.

> > And I still wonder how does Microsoft arrive at the values in the table. For the message box the simplest formula I can find is `round(22*S)+3*round(S-1)-1`  (where S is scale)
> 
> Could the message box one be calculated in dialog units? It likely is some old code... The formula above definitely feels like over-fitting to me.

Yes, I don't like it much either, but this is what we need to use if we want to get exactly the same sizes, I don't see how to achieve them with dialog units (but, admittedly, I didn't try very hard). The only alternative I see is to just hardcode the sizes given above and choose the closest one to the current DPI.

> > I reserve the right to change my mind once again, but I think we should use message box button sizes. [...] What do you think?
> 
> I have no problem with that. An argument for using file dialog size would be that it is probably (?) more commonly seen by users than message boxes, so should be prioritized for aiming consistency at, by I get the high-density UI argument.

OK, I'll push the change with that formula then, unless you (or somebody else) has a better idea.

It's not for the first time I wish I could look at shell32.dll source code :-(


> > cc @marekr just in case you have any thoughts about this
> 
> A px shouldn't be a big deal as long as all wxButtons take the new numbers. Otherwise we'll have mismatches if they don't actually follow GetDefaultSize

Yes, all buttons will get this size by default, but there is more than a pixel difference: at 200% DPI, it's 4px or almost 10%.

vadz:
Sorry, I've realized that I don't know what to use for width. Vaclav, would you already have the numbers for message box button widths too, so that I could avoid spending time on remeasuring them? TIA!

Of course, if you prefer to update this PR to use the numbers matching the message box, it would be great too.

vslavik:
> Vaclav, would you already have the numbers for message box button widths too

Well, brace yourself: which one? 

An "OK" button has 66px at 100%; a "Save" or "Cancel" has 70px (and wx currently has 73px). So I went with the Cancel one and measured it at all [^1] scale factors:

| Scale | Width | Height |
|-------|-------|--------|
| 100%  | 70    | 21     |
| 125%  | 85    | 27     |
| 150%  | 103   | 35     |
| 175%  | 121   | 41     |
| 200%  | 141   | 46     |
| 225%  | 156   | 52     |
| 250%  | 176   | 59     |
| 300%  | 211   | 70     |
| 350%  | 246   | 83     |
| 400%  | 282   | 95     |

So the formula for height doesn't hold over 200% and I couldn't figure out anything for the width either...



[^1]: Except 450% — I couldn't measure that one because buttons won't render properly and just show as gray or black rectangles.

vadz:
Thanks for doing this!

The best I can find for the width is `floor(71*scale - 2)` but it differs by 1px for all the usual scales, so maybe `round(71*scale - 1.5)` is better, as it reproduces the values for 100 and 200 scales, even though it differs by 2px for all the intermediate scales?

For height the best fit I can find is `round(25*scale - 4)` which is off by at most 1px for all scales and coincides with them for 100 and 200, which will just have to be good enough.

Any other thoughts?

vadz:
As there have been no replies, I've pushed the commit above which slightly increases the button sizes. With it we have the following values:

Height | Diff Shell | Diff Message Box |   | Width | Diff Shell | Diff Message Box
-- | -- | -- | -- | -- | -- | --
23 | 0 | 0 |   | 75 | 0 | 3
29 | 1 | 0 |   | 93 | -7 | 6
36 | 3 | -1 |   | 112 | -1 | 7
42 | 0 | -1 |   | 130 | -8 | 7
48 | 4 | 0 |   | 148 | -2 | 5
54 | 3 | 0 |   | 166 | -9 | 8
61 | 7 | 0 |   | 185 | -15 | 7
73 | 8 | 1 |   | 221 | -17 | 8

which is not great but, short of using different formulas for scaling <= 2 and > 2 (maybe we should?), seems to be the best I can do.

I'm closing this PR but please let me know if you see any other possible improvements.

PR #25300, opened 2025-04-07
Add mouse scrolling to wxSpinCtrlGenericBase

### LINKED ISSUE #24935: Generic wxSpinCtrlDouble doesn't handle mouse wheel
#### Bug description:
A wxSpinCtrlDouble object placed on a wxFrame derived parent.
When compiled on linux on **wxGtk the wxSpinCtrlDouble spins** nicely when I roll my mouse wheel. The exact same code compiled on **wxMSW** behaves differently - rolling the **mouse wheel doesn't do anything** to the wxSpinCtrlDouble. 

Am I right in the assumption that there should be no difference in widget behavior among the different platforms? So there should be  wxSpinCtrlDouble spin on mouse wheel roll on wxMSW as well, right?

- wxWidgets port  you use: wxMSW, wxGTK
- OS  and its version: Linux manjaro, Windows 10 & 11
  
  + GTK version:  3.24.43
  + Desktop environment : KDE Plasma Version: 6.1.5, KDE Frameworks Version: 6.6.0

mxdamien:
closes #24935


src/generic/spinctlg.cpp
```diff
@@ -476,6 +477,28 @@ void wxSpinCtrlGenericBase::OnTextChar(wxKeyEvent& event)
...
 
+void wxSpinCtrlGenericBase::OnMouseWheel(wxMouseEvent& event)
+{
+    double value = m_value;
+
+    if( event.GetWheelRotation() > 0 )
+    {
+        value += m_increment;
+    }
+    else
```
lanurmi, line 488: I'm not sure if it can cause actual problems here, but the case of `.GetWheelRotation()` returning zero is not handled. At many other places within wx's source the case of zero is checked.

Also for the whole function, it should be checked that `.GetWheelAxis() == wxMOUSE_WHEEL_VERTICAL`, otherwise also horizontal scrolling modifies the value. Not dangerous, but a little unexpected maybe.

mxdamien: Hi lanurmi,

thanks for your review.

I see your point regarding wxMOUSE_WHEEL_VERTICAL. As for GetWheelRotation() returning zero, I don't see a concrete issue there, but for the sake of completeness, I think the case should still be explicitly ignored.

PR #25306, opened 2025-04-08
Remove obsolete assert about menuitem ID zero on Mac

lanurmi:
The assert, introduced in 524c47a (osx new layout, 2008-09-02), says that menuitem ID zero does not work on Mac, yet empirically it seems to work just fine.

If it really doesn't work, then this assert should be present in all other ports too, to help writing portable code.

### DIFF
--- a/src/osx/menuitem_osx.cpp
+++ b/src/osx/menuitem_osx.cpp
@@ -36,8 +36,6 @@ wxMenuItem::wxMenuItem(wxMenu *pParentMenu,
                        wxMenu *pSubMenu)
            :wxMenuItemBase(pParentMenu, id, t, strHelp, kind, pSubMenu)
 {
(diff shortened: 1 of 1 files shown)


vadz:
I remember wondering about why it didn't work, if it does work this would answer that question.

Just to be sure, have you tested this for both the menus in the menu bar and popup menus? I.e. does using ID of 0 work in both of these cases?

Thanks!

vadz:
BTW, this assert was really added in 408aef5ec1 (as we cannot handle zero commands properly, we have to warn, 2005-01-04). I have no idea what the discussion on wx-dev was about, unfortunately :-(

lanurmi:
> Just to be sure, have you tested this for both the menus in the menu bar and popup menus? I.e. does using ID of 0 work in both of these cases?

Yes, I tested both the menu bar and popup menu in the menu sample, and I didn't encounter anything unusual using 0.

I'll update the commit message to reference to the commit that really added the assert. So it was indeed added into the Carbon implementation, which might have been very different from today's Cocoa one.

PR #25313, opened 2025-04-10
Define ssize_t under Win32 in the same way as the platform SDK

### LINKED ISSUE #25309: ssize_t defination is weird on 32-bit Windows platform.
### Build System Used

- [x] CMake
- [ ] configure
- [ ] nmake with makefile.vc
- [ ] Xcode

### Description

I've a problem when integrate wxWidgets with cpp-httplib on 32-bit Windows target, as it's report that wxWidgets defines `ssize_t` as `int` but cpp-httplib defines `ssize_t` as `long` (but also 32-bit on 32-bit Windows platform)
wxWidgets:
https://github.com/wxWidgets/wxWidgets/blob/195dc18d4f25ba87a8b2f1b713bf03fc864f38f7/include/wx/types.h#L332
cpp-httplib:
https://github.com/yhirose/cpp-httplib/blob/787a34ad7f01f20922a237d5142aae469828be72/httplib.h#L171
According to a pull request of cpp-httplib https://github.com/yhirose/cpp-httplib/pull/1337 , MS define `SSIZE_T` as an alias of `LONG_PTR` which is an alias of `long`, that should be ABI-compatible with other third-party codes on Windows platform.

I found `SSIZE_T` was defined as `long` since Visual C++ 2003 but not sure if it was in 2002 or early version.
Image

Should it be change before 3.3.0 release or keep it?

- wxWidgets version  you are building: 3.2.7
- wxWidgets port  you are building: wxMSW
- Compiler  being used: MSVC 2022
- Non-default compiler options, if any :

vadz:
For compatibility with the other libraries, define ssize_t in the same way as SSIZE_T is defined in the platform headers: this was also true for Win64, but for Win32 we defined ssize_t as int while everybody else defined it as long, so do it too now.

Closes #25309.

### DIFF
--- a/include/wx/types.h
+++ b/include/wx/types.h
@@ -316,7 +316,14 @@ typedef wxULongLong_t wxUint64;
         #include <sys/types.h>
     #endif
 #else /* !HAVE_SSIZE_T */
-    #if SIZEOF_SIZE_T == 4
+    /* Under Windows use definitions compatible with SSIZE_T in windows.h */
+    #ifdef __WINDOWS__
+        #ifdef __WIN64__
+            typedef __int64 ssize_t;
+        #else
+            typedef long ssize_t;
+        #endif
+    #elif SIZEOF_SIZE_T == 4
         typedef wxInt32 ssize_t;
     #elif SIZEOF_SIZE_T == 8
         typedef wxInt64 ssize_t;


vadz:

PR #25316, opened 2025-04-10
Make wxAuiFlatTabArt the default wxAuiNotebook art provider

vadz:
Add wxAuiNativeTabArt which can be used to restore the previous default behaviour.

----

Please let me know if anybody has any objections.


vadz:
Merged in the commit above.

PR #25318, opened 2025-04-11
Add "@USER@" macro to HTML and RichText printers

Blake-Madden:
Adding `@USER@` to your header or footer with `wxHtmlEasyPrinting` or `wxRichTextPrintout` will now expand to the current user name. For example, `"Printed by @USER@ on @DATE@"` could expand to something like `"Printed by bmadden on 04/11/2025"`.

Also, minor formatting fix in interface documentation.

### DIFF
--- a/interface/wx/html/htmprint.h
+++ b/interface/wx/html/htmprint.h
@@ -325,7 +325,7 @@ public:
          - @@PAGESCNT@ is replaced by total number of pages
          - @@TIME@ is replaced by the current time in default format
          - @@TITLE@ is replaced with the title of the document
-         - @@USER@@: is replaced with the user's name.
+         - @@USER@: is replaced with the user's name.
 
         @param footer
             HTML text to be used as footer.
--- a/interface/wx/html/htmprint.h
+++ b/interface/wx/html/htmprint.h
@@ -320,11 +320,12 @@ public:
 
     /**
         Set page footer. The following macros can be used inside it:
-         @@DATE@ is replaced by the current date in default format
-         @@PAGENUM@ is replaced by page number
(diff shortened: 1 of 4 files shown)


interface/wx/html/htmprint.h
```diff
@@ -320,11 +320,12 @@ class wxHtmlEasyPrinting : public wxObject
...
-         @@PAGENUM@ is replaced by page number
-         @@PAGESCNT@ is replaced by total number of pages
-         @@TIME@ is replaced by the current time in default format
-         @@TITLE@ is replaced with the title of the document
+         - @@DATE@ is replaced by the current date in default format
+         - @@PAGENUM@ is replaced by page number
+         - @@PAGESCNT@ is replaced by total number of pages
+         - @@TIME@ is replaced by the current time in default format
+         - @@TITLE@ is replaced with the title of the document
+         - @@USER@@: is replaced with the user's name.
```
PBfordev, line 328: Unless this is somehow by design
```suggestion
         - @@USER@: is replaced with the user's name.
```
Applies to all the occurrences of `@@USER@@` in the docs.

I am not sure how useful the information returned by the `wxUserName()` is for printouts, but I guess it is on the library user to decide if they want to use it.

BTW, `wxGetUserName()` , at least on Windows, is not a trivial function, is it?

Blake-Madden: `wxGetUserName()` is a little "expensive", but header and footer macros are expanded just once in `OnPreparePrinting()`. So even if you are printing a thousand pages, it shouldn't be calling `wxGetUserName()` more than four times.

Personally, I find this really useful because I store these header/footer strings in a settings file, and it would be nice to have dynamic user names support.

PR #25319, opened 2025-04-11
Document using MSBuild for building, not nmake

### LINKED ISSUE #24830: Update MSVC instructions to use msbuild
docs/msw/install.md shows use of nmake to build the libraries. I believe the current practice is to use msbuild, instead. Example for VS2022 (assumes vcvarsall.bat has been run to set up tool paths):
```
msbuild /m /p:Configuration=Debug /p:Platform=x64 wx_vc17.sln
msbuild /m /p:Configuration=Release /p:Platform=x64 wx_vc17.sln
```

vadz:
MSBuild is strongly recommended, as it's much faster (and will also remain supported, unlike nmake).

Closes #24830.

----

I've done the strict minimum here, but it's better than nothing...

### DIFF
--- a/docs/doxygen/mainpages/samples.h
+++ b/docs/doxygen/mainpages/samples.h
@@ -35,16 +35,15 @@ and often it is even more useful to experiment by making a small change to a
 sample and checking how it works when running the sample.
 
 Before being able to run them, you need to be able build the sample you're
-interested in. The right way to do it depends on the way you built wxWidgets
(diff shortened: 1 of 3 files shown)


taler21:
We still use `nmake` because it generates the `build.cfg` file required by a third-party library we use.

Is there a way to generate this `build.cfg` file when using `msbuild`?

vadz:
Merged in the commit above.

PR #25320, opened 2025-04-11
Format page numbers with thousands separators when printing

Blake-Madden:
Applies to `wxHtmlPrintout` and `wxRichTextPrintout`.

Also, add `const` and pages count size check. I noticed the first time `wxHtmlPrintout::TranslateHeader` is called from the html printing sample, the page breaks are empty and the page count would be formatted to an enormous value.

### DIFF
--- a/src/html/htmprint.cpp
+++ b/src/html/htmprint.cpp
@@ -29,6 +29,7 @@
 #include "wx/wxhtml.h"
 #include "wx/wfstream.h"
 #include "wx/infobar.h"
+#include "wx/numformatter.h"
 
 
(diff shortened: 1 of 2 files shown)


vadz (approved):
I wonder if it might be better to ensure that `m_PageBreaks` is never empty, but using `size() - 1` without checking for emptiness was definitely a bad idea, thanks for fixing this!

The other changes look good too, thanks!

vadz:
Pushed now after breaking into 2 commits because I'm going to cherry-pick the first one in 3.2 too.

Thanks!

PR #25323, opened 2025-04-12
Fix assigning max size_t (0 - 1) to int

Blake-Madden:
Sorry, just noticed one more `m_PageBreaks.size() - 1` when `m_PageBreaks.size()` is zero. The cast from before would probably be fixing it, but just the same I think it's better to do it this way.

Also, replace `INT_MAX` with `numeric_limits` and use `static_cast`s.


vadz:
Thanks, I'll apply this soon but I've finally given up and implemented `wxSsize()` because using `std::ssize()` is so much nicer than all those casts to `int`, so I'll rebase it on top of #25326 and use that in it.

Issue #25327, opened 2025-04-14
Possible ABI change in 3.2.7


swt2c:
### Description

There seems to be a possible ABI change in wxWidgets 3.2.7.  On Debian Unstable and Fedora Rawhide, when updating wxWidgets to 3.2.7 but using wxPython compiled against wxWidgets 3.2.6, I'm seeing this error:

```
$ python3 
Python 3.13.2 (main, Feb  6 2025, 00:00:00) [GCC 15.0.1 20250204 (Red Hat 15.0.1-0)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import wx
Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    import wx
  File "/usr/lib64/python3.13/site-packages/wx/__init__.py", line 17, in <module>
    from wx.core import *
  File "/usr/lib64/python3.13/site-packages/wx/core.py", line 12, in <module>
    from ._core import *
ImportError: /usr/lib64/python3.13/site-packages/wx/_core.cpython-313-x86_64-linux-gnu.so: undefined symbol: _ZN14wxTranslations27GetBestAvailableTranslationERK8wxString, version WXU_3.2
```

There were a few changes in the translations code, but nothing seems to have changed the interface of GetBestAvailableTranslation.  Could this change possibly be related?

https://github.com/wxWidgets/wxWidgets/commit/ed7defadd299a649f08af2436db193d7bd5a4180

### Platform and version information

- wxWidgets version  you use: 3.2.7
- wxWidgets port  you use: wxGTK
- OS  and its version: Debian Unstable and Fedora Rawhide
  
  + GTK version: 
  + Which GDK  backend is used:
  + Desktop environment :
  + Current theme:

vadz:
Thanks for noticing this, I did indeed trust the ABI check to detect any breakage, but I hadn't realized it only checked compatibility with 3.2.0 but not the later 3.2.x :-(

Of course, this almost surely means that all the other functions whose versions were changed in ed7defadd299a649f08af2436db193d7bd5a4180 are affected as well (except for `wxFileDialog::AddShortcut()` because I've still managed to get this one wrong).

I think the commit still needs to be reverted and then all the existing contents of the version script except for the fallback version at the end and the version assigned to `wxApp::GTKAllowDiagnosticsControl()` should be just removed, as they don't apply anyhow. This will allow 3.2.8 be ABI-compatible with 3.2.6 again.

[...]

PR #25329, opened 2025-04-14
Fix versioned symbols version in 3.2

vadz:
This fixes an ABI breakage due to the changes of ed7defadd299a649f08af2436db193d7bd5a4180, see #25327.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -552,7 +552,6 @@ AC_DEFUN([WX_VERSIONED_SYMBOLS],
             ])
 
             if test $wx_cv_version_script = yes ; then
-                AC_DEFINE(wxHAVE_VERSION_SCRIPT)
                 LDFLAGS_VERSIONING="-Wl,--version-script,$1"
                 if test $wx_cv_undefined_version = yes ; then
                     LDFLAGS_VERSIONING="$LDFLAGS_VERSIONING -Wl,--undefined-version"
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -552,6 +552,7 @@ AC_DEFUN([WX_VERSIONED_SYMBOLS],
             ])
 
             if test $wx_cv_version_script = yes ; then
+                AC_DEFINE(wxHAVE_VERSION_SCRIPT)
                 LDFLAGS_VERSIONING="-Wl,--version-script,$1"
                 if test $wx_cv_undefined_version = yes ; then
                     LDFLAGS_VERSIONING="$LDFLAGS_VERSIONING -Wl,--undefined-version"
--- a/configure
+++ b/configure
@@ -33818,6 +33818,8 @@ fi
 $as_echo "$wx_cv_version_script" >&6; }
 
             if test $wx_cv_version_script = yes ; then
+                $as_echo "#define wxHAVE_VERSION_SCRIPT 1" >>confdefs.h
+
                 LDFLAGS_VERSIONING="-Wl,--version-script,\$(wx_top_builddir)/version-script"
                 if test $wx_cv_undefined_version = yes ; then
                     LDFLAGS_VERSIONING="$LDFLAGS_VERSIONING -Wl,--undefined-version"
--- a/configure
+++ b/configure
@@ -1156,6 +1156,7 @@ enable_macosx_arch
 enable_compat28
 enable_compat30
 enable_rpath
+enable_symver
 enable_visibility
 enable_tls
 enable_repro_build
@@ -2131,6 +2132,7 @@ Optional Features:
   --enable-compat28       enable wxWidgets 2.8 compatibility
   --disable-compat30      disable wxWidgets 3.0 compatibility
   --disable-rpath         disable use of rpath for uninstalled builds
+  --disable-symver        disable use of ELF symbols versioning even if supported
   --disable-visibility    disable use of ELF symbols visibility even if supported
   --disable-tls           disable use of compiler TLS support
   --enable-repro-build    enable reproducible build mode
@@ -6401,6 +6403,35 @@ fi
 
           eval "$wx_cv_use_rpath"
 
+
+
+          enablestring=disable
+          defaultval=
+          if test -z "$defaultval"; then
+              if test x"$enablestring" = xdisable; then
+                  defaultval=yes
+              else
+                  defaultval=no
+              fi
+          fi
+
+          # Check whether --enable-symver was given.
+if test "${enable_symver+set}" = set; then :
+  enableval=$enable_symver;
+                          if test "$enableval" = yes; then
+                            wx_cv_use_symver='wxUSE_ELF_SYMVER=yes'
+                          else
+                            wx_cv_use_symver='wxUSE_ELF_SYMVER=no'
+                          fi
+
+else
+
+                          wx_cv_use_symver='wxUSE_ELF_SYMVER=${'DEFAULT_wxUSE_ELF_SYMVER":-$defaultval}"
+
+fi
+
+
+          eval "$wx_cv_use_symver"
 
 
           enablestring=disable
@@ -33719,6 +33750,7 @@ fi
 
 if test "$wxUSE_SHARED" = "yes"; then
 
+        if test "$wxUSE_ELF_SYMVER" != "no"; then
 
     case "${host}" in
         *-*-cygwin* | *-*-mingw* )
@@ -33828,6 +33860,7 @@ $as_echo "$wx_cv_version_script" >&6; }
             ;;
     esac
 
+    fi
 
             if test "$wxUSE_VISIBILITY" != "no"; then
 
--- a/configure
+++ b/configure
@@ -33860,6 +33860,9 @@ $as_echo "$wx_cv_version_script" >&6; }
 
 
         if test "$wx_cv_version_script" = "yes"; then
+                                                                                    saveCppflags="$CPPFLAGS"
+            CPPFLAGS="$saveCppflags -I${srcdir}/include -DwxHAVE_ELF_SYMVER"
+
             saveLdflags="$LDFLAGS"
             LDFLAGS="$saveLdflags -Wl,--version-script,conftest.sym"
 
@@ -33874,8 +33877,9 @@ else
 /* end confdefs.h.  */
 
 
-                            __asm__(".symver foo,foo@lib_old");
-                            __asm__(".symver foo,foo@@lib_new");
+                            #include "wx/private/elfversion.h"
+                            wxELF_SYMVER("foo","foo@lib_old")
+                            wxELF_SYMVER("foo","foo@@lib_new")
                             void foo() {}
                             int main() { return 0; }
 
@@ -33909,7 +33913,8 @@ else
 /* end confdefs.h.  */
 
 
-                                __asm__(".symver foo,foo@@lib_new");
+                                #include "wx/private/elfversion.h"
+                                wxELF_SYMVER("foo","foo@@lib_new")
                                 void foo() {}
                                 int main() { return 0; }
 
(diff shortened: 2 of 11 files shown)


vadz:
@swt2c Could you please check if this fixes the original breakage? TIA!

I'm still not sure why are we allowed to assign a version to the symbol in the version script and using `.symver`, resulting in having 2 versions of the same symbol in the binary, but not to assign 2 different versions in the version script nor to use `.symver` twice, so perhaps I'm exploiting some bug here, as normally we should have 2 different symbols mapped to different versions, but this seems to work for me.

vadz:
OK, so this doesn't always work, as shown by the CI failures... Some of them are understandable (this shouldn't be done under macOS nor when building the static libraries) but I have no idea why does it fail with CMake when it succeeds with the similar autoconf builds and, also, why does it fail in UNICODE=0 build, which shouldn't affect this at all.

MaartenBent:
Maybe because CMake doesn't do anything with the version file. It'll need something equivalent to configure I suppose.
```make
            if test $wx_cv_version_script = yes ; then
                LDFLAGS_VERSIONING="-Wl,--version-script,\$(wx_top_builddir)/version-script"
                if test $wx_cv_undefined_version = yes ; then
                    LDFLAGS_VERSIONING="$LDFLAGS_VERSIONING -Wl,--undefined-version"
                fi
            fi
```

vadz:
Thanks, this would indeed explain it. I think CMake build should use versions too, although I don't know if it's going to be enough to make it ABI-compatible with configure build. Should I try to do it or do you prefer me to not touch CMake stuff?

I've also finally realized the difference between ANSI and Unicode build, thanks to `mold` which gives a tad better error message:

```
mold: error: basedll_common_uilocale.o: symbol wxUILocale::GetSystemLocaleId() has undefined version WXU_3.2
``` 

Of course the version is `WX_3.2` in this case and not `WXU_3.2` :-(

MaartenBent:
> Should I try to do it or do you prefer me to not touch CMake stuff?

You can give it a go, then I don't have to boot up a slow VM.

I had a quick look and it shouldn't be too hard. `version-script` needs to be generated. This can be in `config.cmake`, or in a new file. Set the `WX_VERSION_TAG` variable, and then use `configure_file`.

And add the linker flags to `CMAKE_SHARED_LINKER_FLAGS` in `init.cmake`, similar to `STATIC_LINKER_FLAGS`. Not sure if it needs availability tests, or if recent gcc/clang versions all support these flags.

swt2c:
> @swt2c Could you please check if this fixes the original breakage? TIA!

I'm seeing this error when trying to compile wx 3.2.7 + this PR in Fedora:
```
/tmp/ccstNCgy.s: Assembler messages:
/tmp/ccstNCgy.s: Error: invalid attempt to declare external version name as default in symbol `_ZN14wxTranslations19AddAvailableCatalogERK8wxString@@WXU_3.2'
make[1]: *** [/tmp/ccancKJ9.mk:2: /tmp/ccmUqldV.ltrans0.ltrans.o] Error 1
make[1]: *** Waiting for unfinished jobs....
lto-wrapper: fatal error: make returned 2 exit status
compilation terminated.
/usr/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
make: *** [Makefile:14299: /builddir/build/BUILD/wxGTK-3.2.7-build/wxWidgets-3.2.7/bld_gtk3/lib/libwx_baseu-3.2.so.0.4.0] Error 1
```

vadz:
Oh... I've seen that `.symver` doesn't work with LTO, but I didn't realize we used it.

Apparently we can use `symver` attribute instead, with gcc 10 or later, but it's not supported by clang.

Can we check if we're using LTO at compile-time, i.e. does LTO build predefine something to indicate that it's being used? Then I could try using the attribute for it when using gcc but still use inline asm for clang (and gcc < 10). Although I probably could also just do it unconditionally...

swt2c:
> Oh... I've seen that `.symver` doesn't work with LTO, but I didn't realize we used it.
> 
> Apparently we can use `symver` attribute instead, with gcc 10 or later, but it's not supported by clang.
> 
> Can we check if we're using LTO at compile-time, i.e. does LTO build predefine something to indicate that it's being used? Then I could try using the attribute for it when using gcc but still use inline asm for clang (and gcc < 10). Although I probably could also just do it unconditionally...

It might be that Fedora is enabling LTO?  But anyway, could you check if the `symver` attribute is available, and if so, use it?

Something like is done in this patch at the very bottom?
https://sourceware.org/pipermail/binutils/2022-March/120018.html

vadz:
Let's try doing this.

Note that I didn't do anything for CMake but its build should pass now because it won't be affected by these changes at all as it doesn't define the (just added) `wxHAVE_VERSION_SCRIPT`. I'll try to add support for version script in the CMake builds on master and backport it to 3.2 later, but this is not the focus of this PR.

swt2c:
Looks like CI isn't happy, but I just build the [f3eb19c](https://github.com/wxWidgets/wxWidgets/commit/f3eb19ce2efa839c0d3e8dc922445cbb49bd9e5d) version of this on Fedora and it seems to resolve the issue.  wx 3.2.7 + this patch now works fine with wxPython compiled against 3.2.6.

vadz:
Apparently binutils 2.30 used in Ubuntu 18.04 can't deal with multiple versions defined like this, even if later versions (e.g. 2.34 in 20.04) can.

I should be able to add a check for this in configure, even if this is getting more and more ugly :-( Alternatively, I've checked that assigning just `@@WXU_3.2.N` with this binutils version results in creating the symbol both with this _and_ `@@WXU_3.2` versions, i.e. having 2 default versions. This is weird, but seems to work with both the old and new binutils.

But I've also discovered that neither approach works with `mold`: for the current version, [it only defines](https://github.com/rui314/mold/issues/1442) `foo@WXU_3.2` but not `foo@@WXU_3.2.N` which is fatal, as now `foo` can't be found at link time. And with the attempted workaround above it doesn't emit any symbols at all!

swt2c:
Sorry this has turned into such a mess.  :(  If this is starting to be too much of a rabbit hole for something that (hopefully) won't be used much in the future, would it make sense to instead just make a new 3.2.8 (or 3.2.7.1?) soonish with the ABI change reverted back to 3.2.6 (but breaking with 3.2.7) and just tell people not to use 3.2.7 if they haven't already?

vadz:
There is nothing to be sorry about and, on the contrary, thanks for reporting this. It was entirely my fault and the most impressive thing is that I broke this in spite of there already being a comment in `version-script.in` which warned exactly against doing what I did :-(

Anyhow, it's too late for regrets and I agree that we need to release 3.2.8 a.s.a.p., especially as it's not the only important fix since 3.2.7 (see 1f5d7fb894 and 8a3460c97f, for example).

And I think this is ready to merge now, any objections?

vadz:
BTW, do you know if it's still possible to get 3.2.8 into Trixie? I know it's very late, but the bug fixes in it are really important and it would be much better to have it in the next Debian version than 3.2.7 (even 3.2.6 would be better than 3.2.7, arguably).

swt2c:
> BTW, do you know if it's still possible to get 3.2.8 into Trixie? I know it's very late, but the bug fixes in it are really important and it would be much better to have it in the next Debian version than 3.2.7 (even 3.2.6 would be better than 3.2.7, arguably).

Yes, it should still be possible.  Today, the soft freeze starts, but it should still be possible to get an update in, it just has to wait 10 days for migration now, I believe.

swt2c:
> And I think this is ready to merge now, any objections?

No objections, but if you want me to re-test on Fedora (and Debian), I can do it but probably not for 8 hours or so.

vadz:
This is absolutely not that urgent, but I'd indeed feel better if you could please retest it — it _should_ still work, but nothing replaces testing.

I'll wait until you can do it, and it doesn't need to be done today, I probably won't be able to release 3.2.8 before Friday anyhow and maybe it will even have to wait until the next week.

swt2c:
OK, so I tested this latest patch on Fedora.  The test case with wxPython compiled with 3.2.6 worked.  However, the test with wxPython compiled with 3.2.7 did NOT work.  It looks like the "alias" symbols were missing.  Looking further, it shows this in the configure logs:

```
checking if using multiple ELF symbol versions is supported... no
checking if using ELF symbol versions is supported... no
```

Fedora builds with LTO enabled by default, so perhaps this is relevant to the failed configure test?

vadz:
Could you please check if you see anything in `config.log` (searching for "ELF symbol versions" should find the relevant fix)?

I've also installed Fedora 42 into [toolbox](https://github.com/containers/toolbox) here, but I don't know how do you/does it build wx, could you please give me some guidance here?

swt2c:
Here's the error in config.log:

```
configure:35773: checking if using multiple ELF symbol versions is supported
configure:35791: gcc -o conftest -I/usr/include/webp -DWITH_GZFILEOP -D_GNU_SOURCE -O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m64 -march=x86-64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -mtls-dialect=gnu2 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer    -Wl,-z,relro -Wl,--as-needed  -Wl,-z,pack-relative-relocs -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -Wl,--build-id=sha1 -specs=/usr/lib/rpm/redhat/redhat-package-notes  -Wl,--version-script,conftest.sym conftest.c -lcurl -llzma  -lm >&5
{standard input}: Assembler messages:
{standard input}: Error: invalid attempt to declare external version name as default in symbol `foo@@lib_new'
make: *** [/tmp/ccwduwxg.mk:2: /tmp/ccNVmZeV.ltrans0.ltrans.o] Error 1
lto-wrapper: fatal error: make returned 2 exit status
compilation terminated.
/usr/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
```

The answer about "how do you build wx" is unfortunately somewhat complicated.  It is built in a clean, minimal chroot and a bunch of compile flags are set, presumably that's what's causing the configure test to fail in this case.  If you really want to reproduce exactly, you'll need to have the `fedpkg` package install (`dnf install fedpkg`).  Then, `fedpkg clone -a wxGTK` (this will clone the packaging repository, currently just wx 3.2.7.  Then apply any patches (see wxGTK.spec).  Then you should be able to run `fedpkg mockbuild --no-cleanup-after` (this will build the package and NOT erase the build directory afterwards).  The build directory will be in `/var/lib/mock/fedora-rawhide-x86_64/root/builddir/build/BUILD/wxGTK-3.2.7-build/wxWidgets-3.2.7/`.

You might be able to reproduce outside the chroot by just adding compiler flags.  I tried doing this last night but didn't figure out which ones were relevant yet.

vadz:
Thanks, I could finally look at this and just using `-O2 -flto=auto` causes the test to fail with this error — and I don't even need Fedora for this, it also fails with gcc 14 under Debian.

I'll push the fix soon, thanks in advance for retesting it once again!

vadz:
FYI: the CI failure is due to Fedora image update and is unrelated to the changes here (and is/will be fixed by #25331).

swt2c:
OK, this latest version [a22e124](https://github.com/wxWidgets/wxWidgets/pull/25329/commits/a22e12486f454fc6e6d1383f96a4f5aa0f233657) works with wxPython compiled both with wx 3.2.6 and 3.2.7.  :+1:

vadz:
Great, thanks for testing it (and for originally discovering the problem, of course).

PR #25331, opened 2025-04-15
More 3.2.8 backports

vadz:
I am not 100% sure if we want to include 3ddb4f98a50c39be7877e7aa4b1bf3d631e71851, but I think it's better, on balance, to do the right thing compatible with 3.3 than to leave this unchanged but broken for compatibility, considering how many other changes to these events we've already made in 3.2.

@unxed Please let me know if this fixes the regressions you've seen with 3.2.7.


unxed:
Works for me

PR #25332, opened 2025-04-15
wxQt MDI fixes

AliKet:

PR #25336, opened 2025-04-17
fix wxAuiManager interfering with docview event propagation

wsu-cb:


wsu-cb:
The failed check says that awk is missing from the build system.  I'm pretty sure that is unrelated to this change.

src/common/event.cpp
```diff
@@ -1884,6 +1885,8 @@ bool wxEvtHandler::SearchDynamicEventTable( wxEvent& event )
 
     DynamicEvents& dynamicEvents = *m_dynamicEvents;
 
+    static wxRecursionGuardFlag s_flag;
```
vadz, line 1888: I don't think it's correct to use a global flag here: we could be in a nested call but on a different object and if we postponed pruning in this case, they might never be pruned at all.

What we probably need is a (`unordered_`) set of all event handlers whose `m_dynamicEvents` is currently being used. Or, alternatively, a flag in `wxEvtHandler` itself.

include/wx/aui/framemanager.h
```diff
@@ -654,7 +654,8 @@ class WXDLLIMPEXP_AUI wxAuiManager : public wxEvtHandler
     bool m_updateOnRestore = false;
 
 #ifndef SWIG
-    wxDECLARE_EVENT_TABLE();
+    void FrameBind();
+    void FrameUnbind();
```
vadz, line 658: Do we really need these functions? We seem to be calling them from a single place only...

If we do, I think they should be called something like `{Bind,Unbind}FrameEventHandlers()` as "frame bind" reads awkwardly to me.

vadz (changes requested):
Thanks, but while I have just a minor comment about the first commit, I don't think the second one is correct.

include/wx/aui/framemanager.h
wsu-cb: They are indeed only called from one place each, but I created the FrameBind() function to closely parallel the original event table to make clear how the old and new code were related.  However, I have no problem with moving the body of those functions to their callers, and will do soon.

src/common/event.cpp
wsu-cb: I was aware of this issue when I wrote the code, but I considered permanently postponing the pruning as only a "pessimization", not an actual bug.

On the other hand, I really like the idea of putting the `wxRecursionGuardFlag` in `wxEvtHandler`.  I assume I need to preserve the current structure layout of `wxEvtHandler` to maintain ABI compatibility with 3.2, right?  So how do you feel about changing `m_dynamicEvents` from a `wxVector<>*` to a pointer-to-structure that contains the current vector plus the `wxRecursionGuard`?  I'm pretty sure that in practice all pointers are the same size, so this should preserve ABI compatibility, at least as long as no one is actually using the fact that `m_dynamicEvents` is protected, not private, to make use of `m_dynamicEvents` from a derived class.  For what it's worth, it looks like no wxWidgets code outside `wxEvtHandler` accesses `m_dynamicEvents`.  And no inline function accesses `m_dynamicEvents`, so we're safe from clients accessing it that way.

wsu-cb: I have just realized this isn't as simple as it initially sounded.  Because the `wxEvtHandler` can be deleted within the `wxEvtHandler::SearchDynamicEventTable()` call, the `wxRecursionGuardFlag` owned by the `wxEvtHandler` might be deleted out from under the `wxRecursionGuard`.  I need to do some more thinking...

wsu-cb: It looks like `wxSharedPtr<T>` always has the same size as `T*`, so making `m_dynamicEvents` into a `wxSharedPtr<>` would allow solving the lifetime problem.  I feel like this solution is cleaner than managing the content of a static `unordered_multiset<wxEvtHandler>`.

include/wx/aui/framemanager.h
wsu-cb: Done in https://github.com/wxWidgets/wxWidgets/pull/25336/commits/95daea623880c1ba841c18e62f7f6b507b1ae832

src/common/event.cpp
wsu-cb: @vadz Nudge

vadz: Using shared pointer for `m_dynamicEvents` seems strange to me because it's really not supposed to be shared between anything. But you absolutely have a good point about a flag in `wxEvtHandler` not being a good solution either due to the possibility of deleting `wxEvtHandler` itself.

But I wonder if we could avoid all this business with delayed deletion of the vector entries if we used shared pointers: if iterating over the vector were enough to keep it alive, we could just allocate a new vector instead of doing all this stuff. This would be appealing...

wsu-cb: I think I can argue that each recursive call to `wxEvtHandler::SearchDynamicEventTable()` is "something", so `m_dynamicEvents` is being shared among multiple "things".  As such, I think it is reasonable to make it a shared pointer.  Also, I think that's a relatively low additional cost per call:  each call needs to make a new shared pointer instance (to increment the shared pointer lock count), and convert a simple pointer dereference into a shared pointer dereference.  The creation and cleanup code become a little more expensive, but those aren't done on every event handler.

vadz: I guess we can live with using a shared pointer for it, I'd just like to explain in a comment somewhere that shared pointer is used not because the ownership of this vector is really shared but just to extend its lifetime for long enough by making a local copy of it.

vadz:
BTW, I don't know if I should apply the first commit (with its fixup) already?

wsu-cb:
> BTW, I don't know if I should apply the first commit (with its fixup) already?

I would be inclined to wait until we get the iteration cleanup fixed as well, since I'm pretty sure applying the Unbind() part of the first commit will cause more cases of iteration problems.

include/wx/event.h
```diff
@@ -4065,8 +4067,18 @@ class WXDLLIMPEXP_BASE wxEvtHandler : public wxObject
...
+        wxVector<wxDynamicEventTableEntry*> m_entries;
+        wxRecursionGuardFlag m_flag = 0;
+    };
+    // use wxSharedPtr so that SearchDynamicEventTable() can use another
+    // instance of wxSharedPtr to extend the life of the wxRecursionGuardFlag
+    // to outlive wxRecursionGuard
+    wxSharedPtr<DynamicEvents> m_dynamicEvents;
+    // ensure new m_dynamicEvents has same layout as wxWidgets 3.2 m_dynamicEvents
+    static_assert(sizeof(wxSharedPtr<DynamicEvents>) == sizeof(DynamicEvents*), "wxSharedPtr<> has wrong size");
+    static_assert(alignof(wxSharedPtr<DynamicEvents>) == alignof(DynamicEvents*), "wxSharedPtr<> has wrong alignment");
```
vadz, line 4081: Just FYI, I'll remove this because we can't backport this to 3.2 anyhow (as we can't even use C++11 there) and this is not ABI compatible in any case because `DynamicEvents` has changed.

src/common/event.cpp
```diff
@@ -1226,7 +1226,7 @@ wxEvtHandler::~wxEvtHandler()
             delete entry->m_callbackUserData;
             delete entry;
         }
-        delete m_dynamicEvents;
+        m_dynamicEvents.reset();
```
vadz, line 1229: Do we even need this?

src/common/event.cpp
```diff
@@ -1882,24 +1882,29 @@ bool wxEvtHandler::SearchDynamicEventTable( wxEvent& event )
     wxCHECK_MSG( m_dynamicEvents, false,
                  wxT("caller should check that we have dynamic events") );
 
+    // extend the life of the wxRecursionGuardFlag
+    wxSharedPtr<DynamicEvents> addRef(m_dynamicEvents);
```
vadz, line 1886: I'd call this `localCopy` or something like this.

src/common/event.cpp
```diff
@@ -1882,24 +1882,29 @@ bool wxEvtHandler::SearchDynamicEventTable( wxEvent& event )
     wxCHECK_MSG( m_dynamicEvents, false,
                  wxT("caller should check that we have dynamic events") );
 
+    // extend the life of the wxRecursionGuardFlag
```
vadz, line 1885: I'd like to explain this in more details, e.g. something like:
```suggestion
    // Make sure that m_dynamicEvents remains valid until the end of this function,
    // even if this object itself gets deleted by one of the event handlers called
    // from here.
    //
    // We need to use m_dynamicEvents as we use its m_flag to record whether we're
    // iterating over its entries and check it at the end, at which time this object
    // itself might not exist any more.
    //
    // Note that we can't just use a local variable instead, as this function can
    // also be called recursively, if an event handler dispatches another event while
    // processing this one.
```

vadz (approved):
Thanks, this looks good to me now!

Ideal would be to add a test (to `tests/events/propagation.cpp` or some other file in this directory) verifying that this now works correctly, but I am not sure if there is a relatively simple way to do it. If you can think of anything, please consider adding it because this code is clearly rather fragile and having a test checking that this doesn't get broken again would be very useful.

TIA!

include/wx/event.h
wsu-cb: I don't need 3.2 support, but I think it would be nice to fix this problem there as well.  Are you sure this isn't ABI compatible?  `DynamicEvents` is not referenced using inline functions, so I think its layout should only matter to event.cpp, not an application that's dynamic linked to wxWidgets.  The issue of `static_assert` being unavailable in 3.2 is unfortunate, but we could at least put `wxASSERT()`s somewhere in event.cpp for (only) 3.2.

src/common/event.cpp
wsu-cb: You're right, that's a dtor, so `.reset()` is redundant.

Fixed in https://github.com/wxWidgets/wxWidgets/pull/25336/commits/981aba5174615695c5aae259d9c3236b53f2162d

src/common/event.cpp
wsu-cb: Fixed in https://github.com/wxWidgets/wxWidgets/pull/25336/commits/b318c500f7c33c7b2c49a056f1a883612c18b9ee

wsu-cb:
I have successfully created a `DocViewAui` test (https://github.com/wxWidgets/wxWidgets/pull/25336/commits/3471668976c2920790e0c5e1b18125739cf6b5f1) that fails without the changes in this PR, and succeeds with them.  The test fails with a CRT assert on wxMSW in a debug build (due to array bounds error) in the intermediate state where the `wxAuiManager` change is present but the `wxEvtHandler` change is not.

However, there are two questions I have about this test:

1)  I have made no changes at all to the `DocView` test so far, but the `DocViewAui` test is largely a copy of the `DocView` one with just the code for creating windows changed.  Should I modify the tests to share code, or keep them independent?
2)  I have avoided force-pushing anything in this branch.  However, I think it makes sense to make the test the first commit in the branch so that it can be seen to fail before the fixes are applied, and succeed after the fixes are applied.  What do you think?

AliKet:
FYI, the `DocView` test was recently changed (see b4d80a7) after fixing MDI issues in `wxQt`. So in my opinion, it's better to rebase this PR on master and try to refactor common code if possible.

include/wx/event.h
vadz: I'm not sure that this is ABI-incompatible, but I suspect it is and I'm almost sure this is going to provoke errors from the ABI checker tool that we use in 3.2 (you can try making a PR with just this change against 3.2 to run it).

tests/events/propagation.cpp
```diff
@@ -637,6 +647,110 @@ void EventPropagationTestCase::DocView()
...
+
+        std::unique_ptr<wxDocChildFrameAny<wxAuiMDIChildFrame, wxAuiMDIParentFrame>>
+            child(new wxDocChildFrameAny<wxAuiMDIChildFrame, wxAuiMDIParentFrame>(doc, view, parent.get(), wxID_ANY, "Child"));
+
+        wxMenu* const menuChild = CreateTestMenu(child.get());
+
+        // Ensure that the child that we've just created is the active one.
+        child->Activate();
+
+    #ifdef __WXGTK__
```
vadz, line 698: Is this hack/comment relevant for AUI?

tests/events/propagation.cpp
```diff
@@ -637,6 +647,110 @@ void EventPropagationTestCase::DocView()
 #endif // wxUSE_TOOLBAR
 }
 
+#if wxUSE_AUI
+    void EventPropagationTestCase::DocViewAui()
```
vadz, line 651: Trivial, but could you please remove the indent of this function?

vadz:
> I have successfully created a `DocViewAui` test

Great, thanks!

> However, there are two questions I have about this test:
> 
>     1. I have made no changes at all to the `DocView` test so far, but the `DocViewAui` test is largely a copy of the `DocView` one with just the code for creating windows changed.  Should I modify the tests to share code, or keep them independent?

I am not sure if sharing the code is not going to make things more/too complicated. If I'm wrong, and the merged test remains relatively simple, then, sure, let's merge them. But I'm much less allergic to code duplication in the tests than in the main library, so I'm also fine with leaving this as is.
 
>     2. I have avoided force-pushing anything in this branch.  However, I think it makes sense to make the test the first commit in the branch so that it can be seen to fail before the fixes are applied, and succeed after the fixes are applied.  What do you think?

I don't think this is necessary but you can still rebase this branch on master, if only to squash the fixup commits and fix the CircleCI build (which was fixed on master since this PR was opened).

Thanks!

wsu-cb:
> FYI, the `DocView` test was recently changed (see [b4d80a7](https://github.com/wxWidgets/wxWidgets/commit/b4d80a7e23c7cc954fe25a3ac408023e6c48c336)) after fixing MDI issues in `wxQt`. So in my opinion, it's better to rebase this PR on master and try to refactor common code if possible.

Done.  See https://github.com/wxWidgets/wxWidgets/pull/25336/commits/12a20d786492e998d4518ac7d8ab21eee270ad90

tests/events/propagation.cpp
wsu-cb: Based on experimentation, it appears not to be required.  I have put a `CHECK()` in my current version of the test to ensure it stays unnecessary.

For what it's worth, it seems a little strange to me that the hack is required for `wxDocMDIChildFrame`, and even stranger that it's required for `wxDocMDIChildFrame` but not for `wxAuiMDIChildFrame`, but not strange enough for me to bother trying to change that.

tests/events/propagation.cpp
wsu-cb: Done.  See https://github.com/wxWidgets/wxWidgets/pull/25336/commits/12a20d786492e998d4518ac7d8ab21eee270ad90

wsu-cb:
> I am not sure if sharing the code is not going to make things more/too complicated. If I'm wrong, and the merged test remains relatively simple, then, sure, let's merge them. But I'm much less allergic to code duplication in the tests than in the main library, so I'm also fine with leaving this as is.
...
> I don't think this is necessary but you can still rebase this branch on master, if only to squash the fixup commits and fix the CircleCI build (which was fixed on master since this PR was opened).
> 

I have rebased on master, and also added a new commit https://github.com/wxWidgets/wxWidgets/pull/25336/commits/b303e22312cdd88f98b2efeeee6ca602b5d8ba4a that merges the tests.  I will leave it to you to decide whether this is "relatively simple".  If it is not, then feel free to drop that commit.

AliKet:
Please apply this patch:
```C++
diff --git a/src/qt/frame.cpp b/src/qt/frame.cpp
index 7f670a6045..6f5ae97d71 100644
--- a/src/qt/frame.cpp
+++ b/src/qt/frame.cpp
@@ -226,7 +227,10 @@ void wxFrame::RemoveChild( wxWindowBase *child )
 // excluding any menubar and toolbar if any.
 wxPoint wxFrame::GetClientAreaOrigin() const
 {
-    return wxQtConvertPoint( GetQMainWindow()->centralWidget()->pos() );
+    if ( GetQMainWindow() )
+        return wxQtConvertPoint( GetQMainWindow()->centralWidget()->pos() );
+
+    return wxPoint();
 }
 
 QMainWindow *wxFrame::GetQMainWindow() const

```
which should fix the crash and should have been part of https://github.com/wxWidgets/wxWidgets/pull/25332

AliKet:
See [c07c636](https://github.com/wxWidgets/wxWidgets/pull/25332/commits/c07c6365404142c132f15c6ff3153479be5b331e)

wsu-cb:
> Please apply this patch:

It looks like the patch fixes the problem on Ubuntu, but not Windows.

AliKet:
> > Please apply this patch:
> 
> It looks like the patch fixes the problem on Ubuntu, but not Windows.
> 
> **Edit:** From the `WARN()` logging I have temporarily added to the test, it looks like `child->Activate()` crashes when the child is `wxDocMDIChildFrame`.

#25368 should fix the crashes seen here.

BTW [495caaf](https://github.com/wxWidgets/wxWidgets/pull/25336/commits/495caaf1aad256cfa0abcf79a111df6ef704bf84) is not needed.

wsu-cb:
> #25368 should fix the crashes seen here.

Yes, that fixed it.  All the checks passed!

> BTW [495caaf](https://github.com/wxWidgets/wxWidgets/pull/25336/commits/495caaf1aad256cfa0abcf79a111df6ef704bf84) is not needed.

OK.

include/wx/event.h
wsu-cb: I have created https://github.com/wxWidgets/wxWidgets/pull/25378 which successfully backports this PR to 3.2.

PR #25205, opened 2025-02-27
WebP animations

MaartenBent:
This extends #24479 by addressing some of the remarks made there, adding libwebp as a built-in submodule, and adding WebP animation support.
Also see #23489

https://github.com/user-attachments/assets/c5190d7b-ead8-4168-8d6b-1b91c82b91c4

I think it is mostly finished. Except for updating all the build files (only CMake is up-to-date currently). And documentation. And addressing some more remarks from the linked PR. And maybe adding support for saving WebP animations.

PR #25339, opened 2025-04-17
WebP in project files

MaartenBent:
An extension of #25205

I created a new webp project that is built by the Visual Studio solutions. It does not seem to need an compile or link options.

Because of the huge diff I put this in a separate PR to keep the other one readable.

This sets `#define wxUSE_LIBWEBP 1` in all `setup.h`, so it might break some other builds.
Next step is to create `webp.bkl` so it is added to `makefile.gcc` and `makefile.vc` as well.


MaartenBent:
Xcode projects always defines `HAVE_CONFIG_H`, causing webp to look for a non-existing file.

https://github.com/wxWidgets/wxWidgets/blob/9495129c5eda0b0a1147cf90f122408787dd4fde/build/osx/wx.xcconfig#L16

Adding just an empty `src/web/config.h` to the libwebp repo should fix this. But we need a libwebp repo for that.

@vadz can you fork https://github.com/webmproject/libwebp (or create a clone of https://chromium.googlesource.com/webm/libwebp), create a `wx` branch, and make it the default branch? Thanks.
With that, I should also be able to finalize #25205.

vadz:
I guess the proper fix would be to use `wxHAVE_CONFIG_H` for our use (and the same for all the other `HAVE_XXX`, of course), but for now I've done what you proposed, i.e. created https://github.com/wxWidgets/libwebp/ with default `wx` branch.

TIA!

MaartenBent:
Thanks. Maybe before anyone thinks of using it, can you force the `wx` branch to tag [v1.5.0](https://github.com/wxWidgets/libwebp/releases/tag/v1.5.0) ?

PR #25342, opened 2025-04-21
build/files: install forgotten stdpaths.h for wxGTK on Apple

### LINKED ISSUE #24720: wx/osx/cocoa/stdpaths.h: incorrect install condition
The source uses  on all macOS, regardless of GUI toolkit (which is correct, I believe): https://github.com/wxWidgets/wxWidgets/blob/f63d759e3cffd273923fd3d007110bfa6d113940/include/wx/stdpaths.h#L226-L228

Yet the header is installed only for Cocoa GUI: https://github.com/wxWidgets/wxWidgets/blob/f63d759e3cffd273923fd3d007110bfa6d113940/Makefile.in#L3360

Which leaves any other toolkit broken.

barracuda156:
Closes: https://github.com/wxWidgets/wxWidgets/issues/24720

@vadz We forgot about this one, and I got an error trying to build FileZilla (again, LOL). Since now native stdpaths are being used, we obviously should install corresponding header too.

### DIFF
--- a/build/files
+++ b/build/files
@@ -144,6 +144,7 @@ BASE_COREFOUNDATION_SRC =
 
 BASE_COREFOUNDATION_HDR =
     wx/osx/carbon/region.h
+    wx/osx/cocoa/stdpaths.h
     wx/osx/core/cfarray.h
     wx/osx/core/cfdataref.h
     wx/osx/core/cfdictionary.h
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, I'll push this but I wonder if we shouldn't move `stdpaths.mm` from `cocoa` to `core` subdirectory, as it doesn't make much sense to keep something used by wxGTK under Cocoa. @csomor Do you agree this would make more sense?

csomor:
> Thanks, I'll push this but I wonder if we shouldn't move `stdpaths.mm` from `cocoa` to `core` subdirectory, as it doesn't make much sense to keep something used by wxGTK under Cocoa. @csomor Do you agree this would make more sense?

yes, I agree, if wxGTK needs it as well, core is the right place

PR #25343, opened 2025-04-21
Add GCC version in MSW binaries information fo 3.2

PBfordev:
Add GCC 14.2.0 to the list of MSYS GCC versions for which the Windows prebuilt binaries are available.


vadz:
Thanks!

PR #25321, opened 2025-04-12
Fixes for wxGTK on macOS – 7 (minor fix to Carbon headers parsing on powerpc)

barracuda156:
@vadz Could we tolerate an additional minor fix?

This is the error being addressed:
```
In file included from /opt/local/var/macports/build/_opt_local_ppcports_graphics_wxgtk-devel/wxgtk-devel/work/wxWidgets-99f77b47e91b1aa4860d9ff09fd959c9c91f854a/include/wx/osx/private.h:4,
                 from /opt/local/var/macports/build/_opt_local_ppcports_graphics_wxgtk-devel/wxgtk-devel/work/wxWidgets-99f77b47e91b1aa4860d9ff09fd959c9c91f854a/src/osx/cocoa/[utils_base.mm:24](http://utils_base.mm:24/):
/opt/local/var/macports/build/_opt_local_ppcports_graphics_wxgtk-devel/wxgtk-devel/work/wxWidgets-99f77b47e91b1aa4860d9ff09fd959c9c91f854a/include/wx/osx/core/private.h:42:22: warning: "wxOSX_USE_IPHONE" is not defined, evaluates to 0 [-Wundef]
   42 | #if ( !wxUSE_GUI && !wxOSX_USE_IPHONE ) || wxOSX_USE_COCOA_OR_CARBON
      |                      ^~~~~~~~~~~~~~~~
/opt/local/var/macports/build/_opt_local_ppcports_graphics_wxgtk-devel/wxgtk-devel/work/wxWidgets-99f77b47e91b1aa4860d9ff09fd959c9c91f854a/include/wx/osx/private.h:6:5: warning: "wxOSX_USE_IPHONE" is not defined, evaluates to 0 [-Wundef]
    6 | #if wxOSX_USE_IPHONE
      |     ^~~~~~~~~~~~~~~~
/opt/local/var/macports/build/_opt_local_ppcports_graphics_wxgtk-devel/wxgtk-devel/work/wxWidgets-99f77b47e91b1aa4860d9ff09fd959c9c91f854a/include/wx/osx/private.h:8:7: warning: "wxOSX_USE_COCOA" is not defined, evaluates to 0 [-Wundef]
    8 | #elif wxOSX_USE_COCOA
      |       ^~~~~~~~~~~~~~~
[...]

PR #25344, opened 2025-04-21
Revert "osx/core/private.h: fix parsing included CarbonCore header on powerpc

barracuda156:
This reverts commit c3141dfe3e065c6913f9c3655a649240375f408d

While it worked to fix compilation, it did not address the cause of the issue, and was a suboptimal solution. To avoid parsing error on powerpc, disable precompiled headers usage in configure args.

See: https://github.com/wxWidgets/wxWidgets/pull/25321

I have verified the build of current master and 3.2.7.1 release with this change applied, and instead passing `-DwxBUILD_PRECOMP=OFF` to CMake.
https://github.com/macos-powerpc/powerpc-ports/commit/d05573fa72591f53484ebbb2ae31d8b955f6dfab 

It remains unclear at the moment why precompiled headers lead to that error (see the discussion in the PR), but not using precompiled headers is a valid choice and is preferable to hacking a header.

### DIFF
--- a/include/wx/osx/core/private.h
+++ b/include/wx/osx/core/private.h
@@ -12,12 +12,6 @@
 #ifndef _WX_PRIVATE_CORE_H_
 #define _WX_PRIVATE_CORE_H_
 
-// Without this MachineExceptions.h of CarbonCore cannot be parsed
-// when C++11 or higher standard is used.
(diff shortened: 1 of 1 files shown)


vadz:
OK, thanks, will merge soon.

PR #25346, opened 2025-04-24
macOS: Do not activate CLOSE button on Cmd-C

nobugshere:
PR #15678 and commit https://github.com/wxWidgets/wxWidgets/commit/7a45b7948a17bf4a258a44f6066e9ca1f91bbf31 corrected a problem on macOS where Cmd-C caused the dialog to close if it had a wxID_CANCEL button.

I encountered the same issue with the wxID_CLOSE button.  This PR simply extends the fix to include wxID_CLOSE.

This can be reproduced with the dialogs sample using the "Tip of the Day" dialog (Cmd-T).
Note: This issue exists in the 3.2 branch as well as master.


vadz:
Makes sense to me, thanks! I'll merge this soon unless there any objections from @vslavik (author of the above-mentioned commit) or @csomor.

vslavik:
LGTM.

csomor (commented):
@thanks, good idea

PR #25349, opened 2025-04-28
GTK decoration size-related improvements

vadz:


src/gtk/toplevel.cpp
```diff
@@ -1541,24 +1564,37 @@ void wxTopLevelWindowGTK::GTKUpdateDecorSize(const DecorSize& decorSize)
...
+        // correctly because we didn't have the correct decorations sizes when
+        // we did it -- but now that we do, we can adjust the size to the
+        // desired value.
+        //
+        // Note that we can't test for m_deferShow itself here because it may
+        // have been already reset to false if a WM had generated a
+        // notification with wrong _NET_REQUEST_FRAME_EXTENTS values first (as
+        // mutter does, at least in GNOME 3.48, where it sends 0 values for the
+        // not yet mapped window).
+        if (m_deferShowAllowed)
```
paulcor, line 1579: Yeah, this is the fix. This one line.

src/gtk/toplevel.cpp
```diff
@@ -1541,24 +1564,37 @@ void wxTopLevelWindowGTK::GTKUpdateDecorSize(const DecorSize& decorSize)
...
-                gtk_window_resize(GTK_WINDOW(m_widget), w, h);
-                if (!gtk_window_get_resizable(GTK_WINDOW(m_widget)))
-                    gtk_widget_set_size_request(GTK_WIDGET(m_widget), w, h);
-                resized = true;
+                gtk_window_resize(GTK_WINDOW(m_widget), size.x, size.y);
+                if (!isResizeable)
+                    gtk_widget_set_size_request(GTK_WIDGET(m_widget), size.x, size.y);
             }
         }
-        if (!resized)
```
paulcor, line 1559: I'm pretty sure this still needed.

src/gtk/toplevel.cpp
vadz: Thanks for confirming this!

src/gtk/toplevel.cpp
vadz: OK, thinking more about this, I think you're right, I'll revert this change and will add a comment explaining why.

src/gtk/toplevel.cpp
```diff
@@ -1282,12 +1282,18 @@ void wxTopLevelWindowGTK::DoMoveWindow(int WXUNUSED(x), int WXUNUSED(y), int WXU
...
+void
+wxTopLevelWindowGTK::GTKDoGetSize(bool isResizeable,
+                                  int *width, int *height) const
 {
     wxSize size(m_width, m_height);
-    size.x -= m_decorSize.left + m_decorSize.right;
-    size.y -= m_decorSize.top + m_decorSize.bottom;
-    size.IncTo(wxSize(0, 0));
+
+    if ( isResizeable || !HasClientDecor(m_widget) )
```
vadz, line 1293: BTW, does anybody (meaning @paulcor probably) know why do we check for not using CSD only for non-resizeable windows? I'd like to add a comment here explaining why do we do this, but I have absolutely no idea...

paulcor: 8f47d9ad48d3c59090d3d6f8c9c4bb4eab9dce0f

PR #25352, opened 2025-04-29
Ensure that we include built-in 3rd party libraries headers

### LINKED ISSUE #25350: BUG: When using built-in png and zlib, its include path has low priority, causing wrong zlib-prefix.
### Description

I have custom built libs (including zlib) in a custom path, `/custom` for example.
Now I want to build wxwidgets that depend on one of my custom built libs, but use wxwidget's builtin zlib/png and not depend on my custom zlib.
I exported PKG_CONFIG_PATH to point to my custom libs (`/custom/lib/pkg-config`), so the ./configure script (which may call pkg-config) can find the depended custom lib, and instroduces include path `/custom/include`. Note this include path also includes other custom built libs' header, including the custom zlib's header.

When running `./configure --with-zlib=builtin --with-libpng=builtin ...... && make`, the actual compile command may be things like:

`gcc -c -o wxpng_pngread.o    ......    -I/wx-src/include    -I......    -I......    -I/custom/include    -I......    -I......    -I/wx-src/src/zlib    ......    /wx-src/src/png/pngread.c`

In this case, wx itself's include headers (/wx-src/include) has high priority, overwriting custom include path. This is OK.

But the builtin zlib header files have low priority, got overwritten by custom zlib in `/custom/include`.
Since builtin zlib and png's source code is customized by wx, and uses custom symbol prefix, it behave differently from standard zlib/png.
The builtin png source code incorrectly includes the custom zlib's header rather than builtin zlib's header in this case.
Thus it refers to symbols like `inflateEnd`, but the builtin zlib actually provides symbol `wx_zlib_inflateEnd`. This results in linking error.

For most users using system zlib, their header path is usually in /usr/include, this is the lowest priority path, and not specified in gcc `-I` command line. So there is no problem in this case.

Tested on lastes wxWidgets version 3.2.8
wxWidgets port wxGTK
OS Linux
Compiler gcc

vadz:
Move -I options for built-in 3rd party libraries forward in the compiler command line to ensure that they're getting included even when the system libraries are also available.

Closes #25350.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -87,10 +87,9 @@ WX_RELEASE_NODOT = 33
 WX_VERSION = $(WX_RELEASE).0
 WX_VERSION_NODOT = $(WX_RELEASE_NODOT)0
 LIBDIRNAME = $(wx_top_builddir)/lib
-WXREGEX_CFLAGS = $(WX_CPPFLAGS) -DNDEBUG \
-	-I$(wx_top_builddir)/3rdparty/pcre/src -D__WX__ -DHAVE_CONFIG_H \
-	-DPCRE2_CODE_UNIT_WIDTH=$(wxPCRE2_CODE_UNIT_WIDTH) $(WX_CFLAGS) \
-	$(____SHARED) $(CPPFLAGS) $(CFLAGS)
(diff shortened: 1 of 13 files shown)


vadz:
Merged in the commit above.

PR #25354, opened 2025-04-30
Add wxNODISCARD to all Clone*() functions that return a pointer

lanurmi:
Calling a Clone*() function without using its return value is in most (even all) cases a certain way to leak memory.


vadz:
Looks good, thanks! I will merge it soon.

PR #2240, opened 2021-02-18
wxGTK: Expose GtkTreeView drop position

remyhorton:
Within a wxDataViewCtrl drag-and-drop the underlying GtkTreeView
in wxGTK provides visual hints whether the drop is before/on/after
the destination row, but this information is not exposed to the
end developer. This proof-of-concept adds this functionality.

The main change is connecting to the drag-data-get and
drag-data-received signals which as far as I know were introduced
in GTK3, and are able to provide more information then
GtkTreeDragSourceIface and GtkTreeDragDestIface.

PR #25358, opened 2025-05-01
Set proposed drop index in wxGTK wxDataViewCtrl

vadz:
Use "drag-data-received" signal instead of GtkTreeDragDestIface vfunc to be notified about drops on GtkTreeView, as the signal provides more information, notably the exact position of the drop and not just the nearest item to it, which allows to provide the proposed drop index in wxGTK too.

Closes #2240.


paulcor (approved):

vadz:
Merged in the commit above.

PR #25366, opened 2025-05-05
Expand wxTranslations topic

Blake-Madden:
- Mention `wxLocale` being deprecated (perhaps "unadvised" is a better word)
- Mention `wxUILocale` and how it is also independent of `wxTranslations`. Explain how you can use `wxUILocale` and `wxTranslations` together.

### DIFF
--- a/interface/wx/translation.h
+++ b/interface/wx/translation.h
@@ -11,7 +11,7 @@
 
     In wxWidgets, this class manages message catalogs which contain the
     translations of the strings used for the current language.
-    Unlike the deprecated wxLocale, it isn't bound to locale.
+    Unlike the (unadvised) wxLocale, it isn't bound to locale.
     It can be used either independently of, or in
     conjunction with wxLocale. In the latter case, you should initialize
     wxLocale (which creates the wxTranslations instance) first; in the former, you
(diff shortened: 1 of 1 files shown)


PBfordev:
Regarding `wxLocale` vs `wxUILocale`, see also the discussion in #24785.

For example, if your translatable string is "Value must be between %.2f and %.2f" and used with `wxString::Format/Printf()`, I think you need `wxLocale`, don't you?

Blake-Madden:
> For example, if your translatable string is "Value must be between %.2f and %.2f" and used with `wxString::Format/Printf()`, I think you need `wxLocale` for the locale-specific decimal separator to be used, don't you?

Personally, I've gotten into the habit of using `wxNumberFormatter` everywhere these days (just to be certain). So instead of relying on floating-point formatting inside of `wxString::Format`, I use a `%s` in `wxString::Format` and then use `wxNumberFormatter` to format my number in the argument to `wxString::Format`. For example:

```
wxString::Format(_("Value must be between %s and %s"),
   wxNumberFormatter::ToString or wxNumberFormatter::Format ...);
```

Perhaps this should be the recommended way to format floating-point in general? That's the impression that I get. Thoughts?

vadz (commented):
Maybe it's just me, but I don't find this explanation really clear, as the mention of wxLocale comes seemingly out of nowhere. I would reformulate this to say that this class can, and should, be used on its own and that programs using it would also typically use wxUILocale. And then, at the end, mention that this class is also used by wxLocale under the hood, but that using wxLocale is not recommended any longer.

Blake-Madden:
OK, just majorly reworded the overview.

vadz:
Looks much better now, thanks! Will merge in a moment.

PR #25372, opened 2025-05-07
Implement GetMainWindowOfCompositeControl for wxDataViewCtrl

vslavik:
Implement `GetMainWindowOfCompositeControl()` in the generic version of `wxDataViewCtrl` so that focus is reported correctly.

AFAICT should be safe to backport to 3.2 as-is (well, sans `override`), as these are private classes not exposed in any headers, let alone public API ones.


vadz:
Thanks, will merge and backport soon.

Issue #24749, opened 2024-08-03
Build wxWidgets 3.3 on Chinese language Windows system failed


starwing:
#### Bug description:

just checkout the master branch of wxWidgets, and build it with cmake: on Chinese language system:
```
git checkout master && cd ..
mkdir wx-build
cd wx-build
cmake ..\wxWidgets
cmake --build . --config Release
```

it fails with:
```
C:\Work\Sources\wxWidgets\src\stc\lexilla\lexers\LexJulia.cxx(273,46): error C2143: 语法错误: 缺少“)”(在“;”的前面) [C:\Work\Sourc
es\wx-build\libs\stc\wxlexilla.vcxproj]
C:\Work\Sources\wxWidgets\src\stc\lexilla\lexers\LexJulia.cxx(273,46): error C2059: 语法错误:“;” [C:\Work\Sources\wx-build\
libs\stc\wxlexilla.vcxproj]
C:\Work\Sources\wxWidgets\src\stc\lexilla\lexers\LexJulia.cxx(328,1): error C2226: 语法错误: 意外的“size_t”类型 [C:\Work\Sources
\wx-build\libs\stc\wxlexilla.vcxproj]
C:\Work\Sources\wxWidgets\src\stc\lexilla\lexers\LexJulia.cxx(341,28): error C2065: “opsuffs_len”: 未声明的标识符 [C:\Work\Sou
rces\wx-build\libs\stc\wxlexilla.vcxproj]
```

That because the `LexJulia.cpp` file is the `UTF-8` encoding, but `cl.exe` defaults using the code page (in Chinese system, it's cp936) to read sources, I have tried to add these to `build/cmake/init.cmake` (line 54):
```
if(MSVC)
    wx_string_append(CMAKE_C_FLAGS " /utf-8")
    wx_string_append(CMAKE_CXX_FLAGS " /utf-8")
```

But it seems doesn't work. I can only add `/utf-8` to CMakeCache.txt manually:
```
//Flags used by the CXX compiler during RELEASE builds.
CMAKE_CXX_FLAGS_RELEASE:STRING=/MD /O2 /Ob2 /DNDEBUG /utf-8
```

And it does the trick.

#### To Reproduce:

First you should in a Windows that default code page is not `UTF-8` or `latin1` (all CJK system could reproduce), and just build the code.

### Platform and version information

- wxWidgets version  you use: master (3.3)
- wxWidgets port  you use: wxMSW
[...]

PR #25373, opened 2025-05-07
Apply source file charset to utf-8 in MSVC

Kumazuma:
Resolved a build failure on a Korean-language Windows system caused by MSVC interpreting the source file as CP949 rather than UTF-8, by adding the /utf-8 option.

```
FAILED: libs/stc/CMakeFiles/wxlexilla.dir/__/__/__/__/src/stc/lexilla/lexers/LexJulia.cxx.obj
C:\PROGRA~1\MICROS~2\2022\COMMUN~1\VC\Tools\MSVC\1443~1.348\bin\Hostx64\x64\cl.exe  /nologo /TP -DUNICODE -DWX_PRECOMP -D_CRT_SECURE_NO_DEPRECATE=1 -D_SCL_SECURE_NO_WARNINGS=1 -D_UNICODE -IC:\Users\qwead\proj\wxWidgets-1\out2\lib\vc_x64_lib\mswud -IC:\Users\qwead\proj\wxWidgets-1\src\stc\lexilla\access -IC:\Users\qwead\proj\wxWidgets-1\src\stc\lexilla\include -IC:\Users\qwead\proj\wxWidgets-1\src\stc\lexilla\lexlib -IC:\Users\qwead\proj\wxWidgets-1\src\stc\scintilla\include -IC:\Users\qwead\proj\wxWidgets-1\src\stc\scintilla\src /DWIN32 /D_WINDOWS /EHsc /MP /Ob0 /Od /RTC1 -MDd -Zi /W3 /YuC:/Users/qwead/proj/wxWidgets-1/out2/libs/stc/CMakeFiles/wxlexilla.dir/cmake_pch.hxx /FpC:/Users/qwead/proj/wxWidgets-1/out2/libs/stc/CMakeFiles/wxlexilla.dir/./cmake_pch.cxx.pch /FIC:/Users/qwead/proj/wxWidgets-1/out2/libs/stc/CMakeFiles/wxlexilla.dir/cmake_pch.hxx /showIncludes /Folibs\stc\CMakeFiles\wxlexilla.dir\__\__\__\__\src\stc\lexilla\lexers\LexJulia.cxx.obj /Fdlibs\stc\CMakeFiles\wxlexilla.dir\wxlexilla.pdb /FS -c C:\Users\qwead\proj\wxWidgets-1\src\stc\lexilla\lexers\LexJulia.cxx
C:\Users\qwead\proj\wxWidgets-1\src\stc\lexilla\lexers\LexJulia.cxx(1): warning C4819: 현재 코드 페이지(949)에서 표시할 수 없는 문자가 파일에 들어 있습니다. 데이터가 손실되지 않게 하려면 해당 파일을 유니코드 형식으로 저장하십시오.
C:\Users\qwead\proj\wxWidgets-1\src\stc\lexilla\lexers\LexJulia.cxx(270): error C2143: 구문 오류: ')'이(가) ';' 앞에 없습니다.
C:\Users\qwead\proj\wxWidgets-1\src\stc\lexilla\lexers\LexJulia.cxx(270): error C2059: 구문 오류: ';'
C:\Users\qwead\proj\wxWidgets-1\src\stc\lexilla\lexers\LexJulia.cxx(325): error C2226: 구문 오류: 예기치 않은 형식 'size_t'입니다.
C:\Users\qwead\proj\wxWidgets-1\src\stc\lexilla\lexers\LexJulia.cxx(338): error C2065: 'opsuffs_len': 선언되지 않은 식별자입니다.
```

### DIFF
--- a/build/cmake/functions.cmake
+++ b/build/cmake/functions.cmake
@@ -149,6 +149,8 @@ function(wx_set_common_target_properties target_name)
             set_target_properties(${target_name} PROPERTIES MSVC_RUNTIME_LIBRARY ${msvc_runtime})
         endif()
 
+        target_compile_options(${target_name} PRIVATE "/source-charset:utf-8")
+        set_target_properties(${target_name} PROPERTIES QT_NO_UTF8_SOURCE TRUE)
     elseif(NOT wxCOMMON_TARGET_PROPS_DEFAULT_WARNINGS)
         set(common_gcc_clang_compile_options
             -Wall
--- a/build/cmake/init.cmake
+++ b/build/cmake/init.cmake
@@ -112,9 +112,6 @@ if(MSVC)
         wx_string_append(CMAKE_CXX_FLAGS " /MP")
     endif()
 
-    wx_string_append(CMAKE_C_FLAGS " /source-charset:utf-8")
-    wx_string_append(CMAKE_CXX_FLAGS " /source-charset:utf-8")
-
     if(NOT POLICY CMP0092)
         string(REGEX REPLACE "/W[0-4]" "" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
         string(REGEX REPLACE "/W[0-4]" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
--- a/build/cmake/init.cmake
+++ b/build/cmake/init.cmake
@@ -112,6 +112,9 @@ if(MSVC)
         wx_string_append(CMAKE_CXX_FLAGS " /MP")
     endif()
 
+    wx_string_append(CMAKE_C_FLAGS " /source-charset:utf-8")
+    wx_string_append(CMAKE_CXX_FLAGS " /source-charset:utf-8")
+
     if(NOT POLICY CMP0092)
         string(REGEX REPLACE "/W[0-4]" "" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
         string(REGEX REPLACE "/W[0-4]" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
--- a/build/msw/wx_adv.vcxproj
+++ b/build/msw/wx_adv.vcxproj
@@ -234,6 +234,7 @@
       <WarningLevel>Level4</WarningLevel>
       <SuppressStartupBanner>true</SuppressStartupBanner>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
+      <AdditionalOptions>/source-charset:utf-8</AdditionalOptions>
     </ClCompile>
     <ResourceCompile>
       <PreprocessorDefinitions>_DEBUG;_CRT_SECURE_NO_DEPRECATE=1;_CRT_NON_CONFORMING_SWPRINTFS=1;_SCL_SECURE_NO_WARNINGS=1;__WXMSW__;_UNICODE;WXBUILDING;%(PreprocessorDefinitions)</PreprocessorDefinitions>
@@ -269,6 +270,7 @@
       <WarningLevel>Level4</WarningLevel>
       <SuppressStartupBanner>true</SuppressStartupBanner>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
+      <AdditionalOptions>/source-charset:utf-8</AdditionalOptions>
     </ClCompile>
     <ResourceCompile>
       <PreprocessorDefinitions>_DEBUG;_CRT_SECURE_NO_DEPRECATE=1;_CRT_NON_CONFORMING_SWPRINTFS=1;_SCL_SECURE_NO_WARNINGS=1;__WXMSW__;_UNICODE;WXBUILDING;%(PreprocessorDefinitions)</PreprocessorDefinitions>
@@ -305,6 +307,7 @@
       <WarningLevel>Level4</WarningLevel>
       <SuppressStartupBanner>true</SuppressStartupBanner>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
+      <AdditionalOptions>/source-charset:utf-8</AdditionalOptions>
     </ClCompile>
     <ResourceCompile>
       <PreprocessorDefinitions>_DEBUG;_CRT_SECURE_NO_DEPRECATE=1;_CRT_NON_CONFORMING_SWPRINTFS=1;_SCL_SECURE_NO_WARNINGS=1;__WXMSW__;_UNICODE;WXBUILDING;%(PreprocessorDefinitions)</PreprocessorDefinitions>
@@ -341,6 +344,7 @@
       <WarningLevel>Level4</WarningLevel>
       <SuppressStartupBanner>true</SuppressStartupBanner>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
+      <AdditionalOptions>/source-charset:utf-8</AdditionalOptions>
     </ClCompile>
     <ResourceCompile>
       <PreprocessorDefinitions>_DEBUG;_CRT_SECURE_NO_DEPRECATE=1;_CRT_NON_CONFORMING_SWPRINTFS=1;_SCL_SECURE_NO_WARNINGS=1;__WXMSW__;_UNICODE;WXBUILDING;%(PreprocessorDefinitions)</PreprocessorDefinitions>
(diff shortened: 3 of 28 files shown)


vadz:
If we do this, we need to add the same option to all the existing `*.vcxproj` files too for consistency.

Also, I'm not sure how did this manage to break the unit tests, but this needs to be fixed, of course.

MaartenBent:
This problem was mentioned previously in #24749

My preference is to use `target_compile_options` instead of modifying `CXX_FLAGS`.

It seems Qt6 is consciously leaking `-utf-8`, from their CMake script: `By default enable utf8 sources for both Qt and Qt consumers`. Which causes the build to fail. It says it is possible to turn it off with `QT_NO_UTF8_SOURCE`. Maybe like this:

```diff
diff --git "a/build/cmake/functions.cmake" "b/build/cmake/functions.cmake"
index 3a4c195747c..8fb1ed529ce 100644
--- "a/build/cmake/functions.cmake"
+++ "b/build/cmake/functions.cmake"
@@ -149,6 +149,11 @@ function(wx_set_common_target_properties target_name)
             set_target_properties(${target_name} PROPERTIES MSVC_RUNTIME_LIBRARY ${msvc_runtime})
         endif()
 
+        target_compile_options(${target_name} PRIVATE "/source-charset:utf-8")
+        set_target_properties(${target_name} PROPERTIES
+            QT_NO_UTF8_SOURCE TRUE
+        )
+
     elseif(NOT wxCOMMON_TARGET_PROPS_DEFAULT_WARNINGS)
         set(common_gcc_clang_compile_options
             -Wall
```

The test seem to fails because it expects `utf-8` data (`0xC3 0xA4` for `ä`), but because the file is already read as `utf-8`, it gets just the 'normal' character (`0x00E4` in this case because MSVC uses Unicode Character Set). 
Other `FromUTF8` usages all seem to explicitly specify the `utf-8` bytes. So this and the other characters in the test probably need to be changed to:
```c++
CHECK( l.CompareStrings("a",  u8("\xC3\xA4") /* ä */) == -1);
```

Kumazuma:
@MaartenBent 

I think it is better then hexadecimal character
```c++
CHECK( l.CompareStrings("a",  from_u8(u8"ä")) == -1); //  renames u8
```

`u8` string literal has been available since C++11.

It show build warning but the tests is passed.

vadz:
Thanks for the updates, but replacing UTF-8 characters with hex escapes really feels like a step in the wrong direction. I hoped that by always using UTF-8 as the source encoding we would, on the contrary, be able to use Unicode characters everywhere, and I still don't fully understand why this doesn't work, at least with `u8` or `L` prefix.

I also wonder if we're not going to have problems with `u8` and C++20, as its type has changed there to `char8_t*`. It seems like the simplest would be to just use `L` — can we do this?

Finally, could we add this option in `build/msw/wx_setup.props` instead of modifying all the project files to use it? Besides, perhaps this answers my question about why `u8` doesn't work: the tests projects have not been modified, so they probably don't use the new option (unless I'm missing something). Adding it in `wx_setup.props` would ensure that everything does.

MaartenBent:
`u8` here is not the c++ literal, it is an inline function to `wxString::FromUTF8()`. In my suggestion I chose to keep this to keep the number of changes minimal.
The CMake test projects did get the `/source-charset:utf-8` option, and also failed.

The best fix is probably to remove it:
```c++
CHECK( l.CompareStrings("a", "ä") == -1);
```
With or without `L` doesn't matter. It'll either be a 2-byte `char` or a 1-byte `wchar_t` that is converted to wxString.

vadz:
Oops, sorry, I've completely misread the code.

But why wouldn't `wxString::FromUTF8("ä")` work with `/source-charset:utf-8`? It does for me and I think it's still right to keep it (or `u8()`) because without it the test would fail if the _execution_ charset is not UTF-8, as a wrong `wxMBConv` object would be used for constructing the string from bytes in UTF-8 encoding.

vadz:
> While debugging the wxUILocale::CompareStrings test, I found that the representation of the character `ä` depends on the encoding of the source file.
> 
>     * When `/source-charset:utf-8` is specified, `ä` is represented as `0x3F` (`?`).

This seems completely crazy, if the file is in UTF-8 encoding and this option is specified, it really should be represented by its bytes in UTF-8 encoding.
 
> After changing the encoding of initl.cpp to UTF-16BE:

We don't want to do this anyhow...

> This demonstrates that the value of `ä` is influenced by the encoding of the source file.

This is to be expected, but the encoding should be UTF-8. Is this not the case on your system?

> A string literal without `/source-charset:utf-8` is directly copied from the source file, whereas with `/source-charset:utf-8`, MSVC interprets the raw string literal according to the system's current code page.

This just doesn't make any sense to me :-(

vadz:
FWIW just applying

```diff
diff --git a/build/msw/wx_setup.props b/build/msw/wx_setup.props
index 5731b643c8..7d0c6bbc8e 100644
--- a/build/msw/wx_setup.props
+++ b/build/msw/wx_setup.props
@@ -87,6 +87,7 @@
   <ItemDefinitionGroup>
     <ClCompile>
       <AdditionalOptions Condition="('$(Configuration)'=='Release' or '$(Configuration)'=='DLL Release') and '$(VisualStudioVersion
)' >= '14.0'">/Zc:throwingNew %(AdditionalOptions)</AdditionalOptions>
+      <AdditionalOptions>/source-charset:utf-8 %(AdditionalOptions)</AdditionalOptions>
     </ClCompile>
     <Link>
       <AdditionalDependencies>kernel32.lib;user32.lib;gdi32.lib;gdiplus.lib;msimg32.lib;comdlg32.lib;winspool.lib;shell32.lib;shlwa
pi.lib;ole32.lib;oleaut32.lib;uuid.lib;advapi32.lib;version.lib;comctl32.lib;rpcrt4.lib;ws2_32.lib;wininet.lib;winmm.lib;%(Additiona
lDependencies)</AdditionalDependencies>
```

works for me, i.e. the `wxUILocale::CompareStrings` test case still compiles and works fine. Should we apply this and add `target_compile_options(${target_name} PRIVATE "/source-charset:utf-8")` to CMake? These changes make sense to me, while the rest of them do not.

MaartenBent:
It doesn't work for me, and neither the CI. Both CMake and `test_vc17.sln` fail with errors like:

```
wxUILocale::CompareStrings
  German
D:\dev\wxWidgets\tests\intl\intltest.cpp(468)

D:\dev\wxWidgets\tests\intl\intltest.cpp(477): FAILED:
  CHECK( l.CompareStrings("a", u8("Σ")) == -1 )
with expansion:
  1 == -1
```

MaartenBent:
If I use `/utf-8` it works fine, because this will also set the `execution-charset` to utf-8.

Do you maybe have utf-8 in Windows enabled or a different codepage than `1252` that causes it to work for you?
Running `[System.Text.Encoding]::Default` in PowerShell shows lots of codepage info.

vadz:
Sorry, you're right, my system does use UTF-8... Apparently it's due to this checkbox being checked?

image

I don't understand how does the run-time encoding affect this, however... The `"ä"` in the test is represented by a 3 byte string `"\xc3\a4\0"` when the program is executing.

vadz:
After unchecking the checkbox and rebooting I now have:

```powershell
PS C:\Users\zeitlin> [System.Text.Encoding]::Default


IsSingleByte      : True
BodyName          : iso-8859-1
EncodingName      : Western European (Windows)
HeaderName        : Windows-1252
WebName           : Windows-1252
WindowsCodePage   : 1252
IsBrowserDisplay  : True
IsBrowserSave     : True
IsMailNewsDisplay : True
IsMailNewsSave    : True
EncoderFallback   : System.Text.InternalEncoderBestFitFallback
DecoderFallback   : System.Text.InternalDecoderBestFitFallback
IsReadOnly        : True
CodePage          : 1252

PS C:\Users\zeitlin> & $env:WXWIN\tests\vc_x64_mswud\test.exe wxUILocale::CompareStrings
build: 3.3.1 (wchar_t,Visual C++ 1900,STL containers,compatible with 3.2)      
compiled using msvc 1936 (full: 193632537)
running under Windows 10 (build 19044), 64-bit edition as zeitlin
Filters: wxUILocale::CompareStrings
All tests passed (17 assertions in 1 test case)
```

What could be the difference?

vadz:
OK, sorry, the difference is that MSVS got confused after my change of `wx_setup.props` :-( Rebuilding everything from scratch does show the error(s).

MaartenBent:
Indeed, so your execution charset was already utf-8.

From the [documentation](https://learn.microsoft.com/en-us/cpp/build/reference/source-charset-set-source-character-set?view=msvc-170):

> The source character set is the encoding used to interpret the source text of your program. It's converted into the internal representation used as input to the preprocessing phases before compilation. The internal representation is then converted to the execution character set to store string and character values in the executable

But [also](https://learn.microsoft.com/en-us/cpp/build/reference/execution-charset-set-execution-character-set?view=msvc-170)
> The execution character set is the encoding used for the text of your program that is input to the compilation phase after all preprocessing steps.


Now you should also be able to see that removing those `u8()`/`wxString::FromUTF8` does pass the test.

But maybe we should just enable `/utf-8`?

vadz:
And debugging it now shows that the string in the binary is indeed a 2-byte string `"\xe4\0"`.

And reading [`/source-charset`](https://learn.microsoft.com/en-us/cpp/build/reference/source-charset-set-source-character-set) documentation carefully I realize now that it doesn't work as I thought it did, and the encoding of the string in the binary is specified by `/execution-charset` which doesn't result in any assumptions about the charset actually used when the program is executing.

So, you're right, we should just use `/utf-8` after all!

vadz:
I've created the new #25549 replacing this one, so closing it.

@Kumazuma Could you please confirm that using it fixes your original problem too? TIA!

PR #25374, opened 2025-05-07
Add `wxCursorBundle` to make supporting high resolution cursors simpler

### LINKED ISSUE #24564: Support scale factor in wxCursor.
### Description

It would be great to have the ability to provide higher-resolution custom cursors in HiDPI environment.

Currently on wxGTK (X11/XFCE), the bitmaps are scaled.
On wxMSW, they are not scaled.
wxOSX - not tested, but looks like `wxBitmap`'s scale factor needs to be set in `src\osx\carbon\cursor.cpp`
wxQt - `wxBitmap`'s scale factor needs to be set

Prototype patch for wxGTK: [cursor-scale.patch](https://github.com/wxWidgets/wxWidgets/files/15463105/cursor-scale.patch)

vadz:
Closes #24564.


vadz:
I think this should be good enough for now.

vadz:
@dsa-t I know that this doesn't leave a lot of time, but we plan to release 3.3.0 soon and I'd like to merge this a.s.a.p. to prepare for that release, so if you see any issues with this API, please let me know. TIA!

dsa-t:
Thanks for your work! I'll try it in KiCad today.

dsa-t:
There's at least one problem I see with `imagesample` on Linux Mint 21.1, X11, XFCE with window scaling 2x:

- `gtk-cursor-theme-size` is 24
- `wxSYS_CURSOR_Y` returns 48
- `m_bitmaps.GetPreferredBitmapSizeAtScale(double(48) / 32);`  returns 32
- So I get a 32x32 cursor that is scaled 2x by the system and looks blurry.

dsa-t:
(I think the theme cursor size shoudn't matter at that point (`wxCursorBundleMultiImpl::GetCursorFor`), only the display scaling factor)

dsa-t:
And I think `wxSYS_CURSOR_X`/`wxSYS_CURSOR_Y` shouldn't be scaled.

vadz:
I don't know why is the cursor getting scaled for you. I.e. I think we should be able to use 32x32 cursor as is...

dsa-t:
> I don't know why is the cursor getting scaled for you. I.e. I think we should be able to use 32x32 cursor as is...

GTK(?) does it automatically on my system, does it not on your system?

vadz:
It doesn't seem to... This is using Sway, I'll test under GNOME.

vadz:
Anyhow, the problem is that I'm really not sure what can we do if the cursor size is restricted and we don't have a bitmap with this size. Scaling it ourselves is not going to look (much) better than system-performed scaling.

vadz:
> It doesn't seem to... This is using Sway, I'll test under GNOME.

I take my words back, after viewing it with a magnifier (which I can only do in a VM, I wonder if there is some simpler way of viewing the cursor directly on the system), it is indeed doubled.

But this still doesn't help with understanding how can we prevent this from happening :-(

vadz:
At least under MSW the user can change the cursor size independently of the DPI and it's my understanding that we want to respect this setting, don't we? And if we do it for MSW it seems strange not to do it for GTK...

vadz:
I think something goes wrong with the scaling factor when creating the cursor, i.e. it's just not taken into account, so whichever bitmap we use, it's always doubled when DPI is 200%.

We might need to use `gdk_cursor_new_from_surface()` instead of `gdk_cursor_new_from_pixbuf()`...

dsa-t:
I think we can let the app manage this too, and in wx, only handle display scaling changes.

Or maybe we can let the app provide some `std::vector<wxBitmapBundle>` with different default sizes for different custom cursor sizes - but not sure it's worth it.

dsa-t:
> I think something goes wrong with the scaling factor when creating the cursor, i.e. it's just not taken into account, so whichever bitmap we use, it's always doubled when DPI is 200%.
> 
> We might need to use `gdk_cursor_new_from_surface()` instead of `gdk_cursor_new_from_pixbuf()`...

Check out the patch from https://github.com/wxWidgets/wxWidgets/issues/24564 -  it uses `cairo_surface_set_device_scale`

vadz:
> Check out the patch from #24564 - it uses `cairo_surface_set_device_scale`

Right, I've completely missed it there and was just writing something like this myself now. I'll push it here soon.

vadz:
I've spent an embarrassingly long time on deciding how to handle it and the result is in the last commit. I _think_ it does what we want, i.e. selects the best cursor size which makes sense for the current system taking into account both the DPI and the user-defined cursor size.

It feels very weird to me, however, to have to reset the scale factor manually like this, so it's quite possible that I'm still missing something... But it seems to work in my tests. What do you think?

src/gtk/cursor.cpp
```diff
@@ -21,6 +21,7 @@
 #include "wx/gtk/private/wrapgtk.h"
```
dsa-t, line 21: > This fixes unwanted cursor scaling by GTK/Wayland which made the cursor
both blurry and of a wrong size.

Note that this is same on X11.

dsa-t:
I still think the `wxSYS_CURSOR_X` / `wxSYS_CURSOR_Y` values shouldn't depend on display scaling.

Changes seem ok for me on wxMSW, though the default cursor switches to 2x on 150%+ scaling, while the wx cursor switches on 175%+.

On wxGTK I don't like the hardcoded values too much, but I guess that'll do it, unless GTK scales the cursor automatically to the custom size (not tested with latest commits).

src/gtk/cursor.cpp
vadz: True, but I had only tested this with Wayland. Anyhow, I don't think it's worth rewriting history/rerunning the CI once again, so I'll leave it as is.

vadz:
> I still think the `wxSYS_CURSOR_X` / `wxSYS_CURSOR_Y` values shouldn't depend on display scaling.

But why? I mean, if the user selects 64x64 cursor under MSW, shouldn't we use the high DPI version instead of showing the blurry upscaled 32x32 one? AFAIK plenty of people increase their cursors size, so it makes sense to try to behave nicely in this case.

> Changes seem ok for me on wxMSW, though the default cursor switches to 2x on 150%+ scaling, while the wx cursor switches on 175%+.

Yes, this is because the default `GetPreferredBitmapSizeAtScale()` implementation prefers smaller size in case of a tie. I had initially decided to leave it like this, but the argument about the default cursor makes me think it's worth the extra ugliness of 11c3a2fc9cae5f3026654746492cb0d202faf72f.
 
> On wxGTK I don't like the hardcoded values too much, but I guess that'll do it, unless GTK scales the cursor automatically to the custom size (not tested with latest commits).

24 as default cursor size value comes from X11 and I don't think it's going to change in the observable future, so I'm not too worried. I'm more worried about other ports, e.g. wxQt now and maybe something else in the future, but I'm still not worried that much about this, and we can always virtualize this via platform-specific function if really needed.

dsa-t:
> > I still think the `wxSYS_CURSOR_X` / `wxSYS_CURSOR_Y` values shouldn't depend on display scaling.
> 
> But why? I mean, if the user selects 64x64 cursor under MSW, shouldn't we use the high DPI version instead of showing the blurry upscaled 32x32 one? AFAIK plenty of people increase their cursors size, so it makes sense to try to behave nicely in this case.

Other metrics in wxGTK don't depend on a window scaling factor, so I expect wxSYS_CURSOR_* to also not depend on it.

On MSW, e.g. `wxSYS_ICON_X` seems to change with display scaling, so it's probably fine.

vadz:
> Other metrics in wxGTK don't depend on a window scaling factor, so I expect wxSYS_CURSOR_* to also not depend on it.

Hmm, this is true, thanks for pointing this out.

> On MSW, e.g. `wxSYS_ICON_X` seems to change with display scaling, so it's probably fine.

So we have a choice between making this consistent with the other metrics under GTK or with the same metric under MSW. Looking at `src/osx/cocoa/settings.mm` I see that wxOSX also returns logical values, so I guess I'll have to conform to this, even if I don't really like it.

I've just pushed 901d31104af5b75d015d1f442d101a24b50af73a doing this.

Do you see any other remaining problems?

dsa-t:
Seems ok for me.

PR #25378, opened 2025-05-09
Auimgr docview fix backport

wsu-cb:
Backport https://github.com/wxWidgets/wxWidgets/pull/25336 to 3.2.

### DIFF
--- a/include/wx/event.h
+++ b/include/wx/event.h
@@ -29,6 +29,8 @@
 #include "wx/typeinfo.h"
 #include "wx/any.h"
 #include "wx/vector.h"
+#include "wx/recguard.h"
+#include "wx/sharedptr.h"
 
 #include "wx/meta/convertible.h"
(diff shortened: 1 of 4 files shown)


wsu-cb:
Please, could someone who understands the ABI check look at this?  The ABI check is failing, but this PR's current state makes NO CHANGES in `*.h` files!

wsu-cb:
> I think link errors in the other builds provide a clue: `wxAuiManager::GetEventTable()` is not found any more, i.e. a symbol previously exported from the library has disappeared. This is an ABI break.

Thank you, restoring the IMPLEMENTATION of the event table fixed the ABI check.  I had assumed that the ABI check would only examine the `*.h` files.

vadz:
As mentioned above, I'd rather not merge this, so closing. Sorry again for not thinking about it before you spent your time on this.

PR #25380, opened 2025-05-09
Mention distributing wxstd.mo files for AddStdCatalog

Blake-Madden:
Windows (at the very least) won't have wxWidgets catalog installed anywhere, so mention distributing them with your own progam as an option.


vadz:
Thanks, will merge soon.

PR #25381, opened 2025-05-09
Document GetPreferredUILanguages()

Blake-Madden:


Blake-Madden:
Yes, this actually piqued my curiosity and I was going to see how it works on the various platforms :)
I'll experiment with it when I get a chance and I'll expand the docs in a later PR.

PR #25383, opened 2025-05-10
Fix typo in wxWebViewChromium error message

PBfordev:
Since the typo was in a `_()` string, fixing this should also include message catalogs being updated but I don't know the procedure (probably not triggered automatically?).


vadz:
Yes, message catalogs are/can be updated by running `make -C locale allpo`. I'll do it and also post to wx-translators asking to update the translations before the upcoming release, thanks.

PR #25391, opened 2025-05-11
Make errors shown to the users more readable

vadz:
Don't use internal function names in the error messages that can be seen by the end user.

----

I'd like to ask translators to update the existing translations, but before doing it I'd like to fix some of the new messages that don't make much sense to me IMO. I've tried to make them at least somewhat understandable for the end user while keeping them specific enough — please let me know if anybody sees anything wrong with it or can provide better messages.


vadz:

PR #25392, opened 2025-05-11
Move dummy libwebp config file

MaartenBent:
Move it to a separate wx directory, so it won't be in the default include path and other build systems won't include it.


vadz:
Should this be (squash) merged?

MaartenBent:
Not yet, it is now using my test submodule. I'll need to create a PR in [wxWidgets/libwebp](https://github.com/wxWidgets/libwebp) first.

vadz:
I guess it should be good now, but please let me know if I'm still missing anything remaining to be done. Thanks!

PR #25394, opened 2025-05-12
Add support for custom checkbox to wxInfoBar

Blake-Madden:
For the generic `wxInfoBar`, adds a feature to show a custom checkbox at the bottom.

By intercepting the `wxID_CLOSE` button event, the client can check the state of this checkbox and then know whether they should show it again or not.

image

The inclusion of this checkbox can be toggled and is not shown by default.


include/wx/generic/infobar.h
```diff
@@ -75,6 +75,21 @@ class WXDLLIMPEXP_CORE wxInfoBarGeneric : public wxInfoBarBase
...
+    // time of the window being closed.
+    // This should be called in a client's handler for the
+    // wxID_CLOSE button being clicked.
+    bool IsDontShowAgainChecked() const { return m_dontShowAgain; }
+
+    // Whether the "Do not show this again" checkbox is being shown.
+    bool IsIncludingDontShowAgainCheckbox() const { return m_includeDontShowAgain; }
+
+    // Sets whether the "Do not show this again" checkbox should be shown.
+    void IncludeDontShowAgainCheckbox(const bool includeCheckbox)
```
vadz, line 89: Top-level const is ignored and shouldn't be used.
```suggestion
    void IncludeDontShowAgainCheckbox(bool includeCheckbox)
```

include/wx/generic/infobar.h
```diff
@@ -75,6 +75,21 @@ class WXDLLIMPEXP_CORE wxInfoBarGeneric : public wxInfoBarBase
...
+    {
```
vadz, line 90: This function shouldn't be inline, there is really no reason to put it in the header.

src/generic/infobar.cpp
```diff
@@ -115,19 +117,37 @@ bool wxInfoBarGeneric::Create(wxWindow *parent, wxWindowID winid)
     if(!m_hasFgCol)
         m_text->SetForegroundColour(colFg);
 
-    m_button = wxBitmapButton::NewCloseButton(this, wxID_ANY);
+    m_button = wxBitmapButton::NewCloseButton(this, wxID_CLOSE);
     m_button->SetToolTip(_("Hide this notification message."));
 
+    m_dontShowAgainCheckbox =
+        new wxCheckBox(this, wxID_ANY, _(L"Do not show this again."), wxDefaultPosition,
```
vadz, line 124: Why `L`?
```suggestion
        new wxCheckBox(this, wxID_ANY, _("Do not show this again."), wxDefaultPosition,
```

vadz (dismissed):
Thanks, I agree that this could be useful, but I think it would make more sense to

1. Not show the checkbox by default.
2. Add `AddCheckbox()` or, if we want to be consistent with `wxRichMessageDialogBase`, `ShowCheckBox()` function to show it.

Importantly, GTK version should check for the checkbox being shown and fall back to the generic one if it is.

src/generic/infobar.cpp
Blake-Madden: Sorry, force of habit. I use `L` in my code usually, but not in wx PRs.

include/wx/generic/infobar.h
Blake-Madden: Yes, I see the compile error now. Will fix.

Blake-Madden:
>     1. Not show the checkbox by default.

Yes, it is hidden by default.

>     2. Add `AddCheckbox()` or, if we want to be consistent with `wxRichMessageDialogBase`, `ShowCheckBox()` function to show it.

Got it. Now uses the same API as `wxRichMessageDialogBase`.

> Importantly, GTK version should check for the checkbox being shown and fall back to the generic one if it is.

Not sure how to approach this because `wxInfoBar::Create` constructs the control either generically or natively. At that point, `ShowCheckBox()` couldn't have been called yet (that would have to come later) and by then it is too late. I don't see a way to force it to be generic from `ShowCheckBox()` because by then it is already created natively.

I could add an warning in the GTK version of `ShowCheckBox()` if it is running natively, saying that you should use the generic version.

vadz:
To be able to decide which version to use we'd need to use a style (styles are available in `Create()`). I'm not sure if it's worth the extra complexity, but forcing the user to use `wxInfoBarGeneric` with wxGTK is not ideal, it would be really nice to provide some way to use just `wxInfoBar` everywhere.

Blake-Madden:
I added a new window style `wxINFOBAR_GENERIC` to control this. I'm going to do some more testing on Linux and then change it back from draft...

vadz:
> I added a new window style `wxINFOBAR_GENERIC` to control this.

Calling the style `wxINFOBAR_CHECKBOX` would be better, like this we could switch to using native version on GTK (or some other platform) in the future if it adds support for this — while with the current style we'd always have to use the generic one.

Blake-Madden:
OK, just added `wxINFOBAR_CHECKBOX`. It is not on by default, but if you call `ShowCheckBox()` without this style you will get a "helpful" assert message.

include/wx/generic/infobar.h
```diff
@@ -96,10 +110,14 @@ class WXDLLIMPEXP_CORE wxInfoBarGeneric : public wxInfoBarBase
     // should be called when we're shown or hidden)
     void UpdateParent();
 
+    long m_style = 0;
```
vadz, line 113: I don't think we need this one, we should store the style in `wxWindow::m_windowStyle`.

vadz (commented):
Thanks, looks good now, I'd just like to remove `m_style` because it's confusing to have both it and `m_windowStyle`.

Of course, please let me know if it's really needed for some reason I'm missing. Thanks!

vadz (commented):
Sorry, I've realized there is something else I forgot about: it would be nice to add the new style to `wxInfoBarXmlHandler` in `src/xrc/xh_infobar.cpp` (and also to `docs/doxygen/overviews/xrc_format.h`). Could you please do it? TIA!

Blake-Madden:
OK, I added the style and parameters to `ShowCheckbox()` via XRC.

vadz:
Thanks! I was going to merge this but realized I forgot to ask you to add the new attributes to the schema, and so started doing it, but then realized that `checkboxvalue` was rather inconsistent with the similar attributes names and so renamed it to `checkboxlabel` and also changed its semantics slightly: it now always enables `wxINFOBAR_CHECKBOX` because there just doesn't seem any reason not to do it.

I've pushed my changes into your branch, please let me know if you have any remarks, otherwise I'll merge this soon.

Blake-Madden:
Sounds good to me, thanks!

Yeah, I wasn't sure if it was OK to enable that by default since it would cause it to go generic under GTK, but that seems OK to me.

PR #25397, opened 2025-05-13
Call OnExit to receive the OnEventLoopExit notification

alilie:
Unless I'm missing something `OnExit()` is not called after `OSXDoRun()` finishes and we don't get the `OnEventLoopExit` notification.
`OnExit()` is called only if an exception is thrown.

The corresponding code on linux is this:
https://github.com/wxWidgets/wxWidgets/blob/8b6deb74cf6795b4a4a343609efdb577a3b622b7/src/gtk/evtloop.cpp#L58-L92

Here `OnExit()` is called from wxScopeGuard:
https://github.com/wxWidgets/wxWidgets/blob/8b6deb74cf6795b4a4a343609efdb577a3b622b7/src/common/evtloopcmn.cpp#L336-L374

### DIFF
--- a/src/osx/core/evtloop_cf.cpp
+++ b/src/osx/core/evtloop_cf.cpp
@@ -325,6 +325,8 @@ int wxCFEventLoop::DoRun()
 
             OSXDoRun();
 
+            OnExit();
+
 #if wxUSE_EXCEPTIONS
             // exit the outer loop as well
             break;


vadz:
Thanks, but I wonder what happens if `Exit()` is called to exit the loop -- wouldn't `OnExit()` be called twice then?

This code seems to have been copied from `wxEventLoopManual::DoRun()` (old version of it, before it has been updated to use `wxSafeCall()` in ec12651b9b (Add wxSafeCall() function replacing wxTRY/wxCATCH_ALL macros, 2025-03-20)) and there we explicitly dismiss the guard when exiting normally. But a manual event loop can only exit if `Exit()` was called, while macOS and GTK event loops may probably exit both because it was called and because the loop terminated for some other reason (I'm not quite sure about this, especially for Mac) and so we probably need to check if `Exit()` has been called or not?

alilie:
https://github.com/wxWidgets/wxWidgets/blob/3e6b6647628c944ba9d5e599d599081e58ad2ac4/src/osx/core/evtloop_cf.cpp#L261-L303

> but I wonder what happens if `Exit()` is called to exit the loop -- wouldn't `OnExit()` be called twice then?

> so we probably need to check if `Exit()` has been called or not?

From what I can see, `OSXDoRun()` only stops if `m_shouldExit` is true, and `m_shouldExit` is true only when `wxEventLoopBase::Exit` is called.

I expect `Exit()` to have been called when we reach `OnExit()`.

And I don't see `OnExit()` being called from void `wxEventLoopBase::Exit(int rc)`.

I'm not too sure either. I just saw that our overriden `OnEventLoopExit` method wasn't being called on Mac and I noticed the lack of the `OnExit()` call unlike the other 2 places (GTK and LoopManual).

vadz:
After looking into this, I think we should really call `OnExit()` from `ScheduleExit()` in all ports to ensure that it is always called.

Could you please test #25409 which changes things to work like this?

PR #25399, opened 2025-05-14
Reduce overhead of checking for  "catch-unhandled-exceptions"

vadz:
Also optimize `wxSystemOptions` more generally (at the cost of not taking into account changes to the environment variables while the program is running).


vadz:
@lanurmi If you don't have any objections, I'd like to merge this soon.

PR #25402, opened 2025-05-14
Add mfc.h to file list and enable MFC sample

MaartenBent:


vadz:
Thanks, will merge soon.

PR #25403, opened 2025-05-15
Update translation for Romanian

catalinr:


vadz:
Thanks, but this is probably going to conflict with #25400, won't it?

catalinr:
> this is probably going to conflict with https://github.com/wxWidgets/wxWidgets/pull/25400

Yes, most likely. Then I'll wait for it to be merged and rebase on `master` after that.

vadz:

PR #25406, opened 2025-05-15
Fix building libwebp in universal macOS build

MaartenBent:
Fix building libwebp in universal macOS build. The SIMD flags that libwebp adds (like `-mno-sse2`) do not work with it.

Also add a CMake option to disable use of system libraries, and thus use all built-in libraries.

And add a CI job to check that the universal macOS build now succeeds. I don't think we had any job building `wxWebviewChromium`, so enable that as well.


vadz:
Thanks for tracking this one down!

Issue #25396, opened 2025-05-13
wxDisplay object becomes dangling after monitor disconnection on Windows


Marcel-Pham:
**Summary:**
A `wxDisplay` object created before a monitor configuration change (e.g., unplugging an external display or closing a laptop lid) becomes dangling due to an invalidated internal implementation pointer. Any subsequent call on such an object (e.g., `GetScaleFactor()`) results in a memory access error and crashes the application.

**Platform:**
Windows 11, 10, likely all that support `WM_DISPLAYCHANGE`)

**wxWidgets Version:**
Git master

**Steps to Reproduce:**

1. Create a `wxDisplay` object.
2. While the object is still in scope, trigger a monitor setup change (e.g., disconnect an external monitor or close/reopen a laptop lid).
3. This causes the `WM_DISPLAYCHANGE` message to be sent.
4. `wxDisplay::InvalidateCache()` is triggered, which clears the internal display implementation list (`wxDisplayFactory::m_impls`), but does not update existing `wxDisplay` objects.
5. Any further use of the `wxDisplay` instance results in dereferencing a dangling pointer.

**Expected Behavior:**
`wxDisplay` should either:

- Detect invalidation and gracefully fail (e.g., return invalid results or rebind to a valid impl), or
- Be internally updated when the display configuration changes, or
- Mark existing instances as invalid to avoid undefined behavior.

**Proposed Solutions (Suggestions):**

- Track all wxDisplay instances and nullify their m_impl pointers upon invalidation.
- Add validity checks inside member functions to guard against use-after-invalidate.
- Use shared ownership, reference counting or weak references for display impls to prevent dangling.

**Code Sample (Minimal Reproducible Example):**
```cpp #include "wx/wxprec.h"

#ifndef WX_PRECOMP
    #include "wx/wx.h"
#endif

#include <wx/display.h>

class MyApp : public wxApp
{
public:
    bool OnInit() override;
};

class MyFrame : public wxFrame
{
public:
    MyFrame(const wxString& title);
};

wxIMPLEMENT_APP(MyApp);

bool MyApp::OnInit()
{
    if (!wxApp::OnInit())
        return false;
    MyFrame* frame = new MyFrame("Minimal wxWidgets App");
    frame->Show(true);
    return true;
}

MyFrame::MyFrame(const wxString& title)
    : wxFrame(nullptr, wxID_ANY, title)
{
    CallAfter([this]() -> void
    {
[...]

PR #25407, opened 2025-05-16
wxDisplay object becomes dangling after monitor disconnection on Windows

Marcel-Pham:
This is a PR for the solution discussed in #25396:

- `wxDisplayImpl` derives now from `wxObjectRefData`
- The raw pointers `wxDisplay::m_impl` and `wxDisplayFactory::m_impls` are replaced with shared pointers
- `wxDisplayImpl` has a new member `m_isConnected` with access functions
- `wxDisplay` returns default values if the connection to the monitor is broken
- `wxDisplayFactor::InvidalidateCache` will not delete the display objects, if the displays are still connected but update display information

### DIFF
--- a/include/wx/display.h
+++ b/include/wx/display.h
@@ -57,7 +57,9 @@ public:
 
     // dtor is not virtual as this is a concrete class not meant to be derived
     // from
-
+    // dtor must not be inline because the implementation of
+    // wxDisplayImpl::DecRef() is unknown at this place 
+    ~wxDisplay();
 
     // return the number of available displays, valid parameters to
     // wxDisplay ctor are from 0 up to this number
@@ -78,6 +80,9 @@ public:
 
     // return true if the object was initialized successfully
     bool IsOk() const { return m_impl != nullptr; }
+
+    // return true if the display has not been unplugged yet
+    bool IsConnected() const;
 
     // get the full display size
     wxRect GetGeometry() const;
@@ -159,7 +164,7 @@ private:
 
 
     // the real implementation
-    wxDisplayImpl *m_impl;
+    wxObjectDataPtr<wxDisplayImpl> m_impl;
 
 
     wxDECLARE_NO_COPY_CLASS(wxDisplay);
--- a/include/wx/display.h
+++ b/include/wx/display.h
@@ -58,7 +58,7 @@ public:
     // dtor is not virtual as this is a concrete class not meant to be derived
     // from
     // dtor must not be inline because the implementation of
-    // wxDisplayImpl::DecRef() is unknown at this place 
+    // wxDisplayImpl::DecRef() is unknown at this place
     ~wxDisplay();
 
     // return the number of available displays, valid parameters to
--- a/include/wx/private/display.h
+++ b/include/wx/private/display.h
@@ -68,9 +68,6 @@ private:
     // On-demand populated vector of wxDisplayImpl objects.
     wxVector<wxObjectDataPtr<wxDisplayImpl> > m_impls;
 
-    // The count of connected displays retrieved on last display access.
-    unsigned m_countOnLastAccess = 0u;
-
     wxDECLARE_NO_COPY_CLASS(wxDisplayFactory);
 };
 
--- a/include/wx/private/display.h
+++ b/include/wx/private/display.h
@@ -26,38 +26,7 @@ public:
 
     // Create the display if necessary using CreateDisplay(), otherwise just
     // get it from cache.
-    wxObjectDataPtr<wxDisplayImpl> GetDisplay(unsigned n)
-    {
-        // Normally, m_impls should be cleared if the number of displays in the
-        // system changes because InvalidateCache() must be called. However in
-        // some ports (e.g. Mac right now, see #18318), cache invalidation never
-        // happens, so we can end up with m_impls size being out of sync with
-        // the actual number of monitors. Compensate for this here by checking
-        // if the index is invalid and invalidating the cache at least in this
-        // case.
-        //
-        // Note that this is still incorrect because we continue using outdated
-        // information if the first monitor is disconnected, for example. The
-        // only real solution is to ensure that InvalidateCache() is called,
-        // but for now this at least avoids crashes when a new display is
-        // connected.
-        if ( n >= m_impls.size() )
-        {
-            // This strange two-step resize is done to clear all the existing
-            // elements: they may not be valid any longer if the number of
-            // displays has changed.
-            m_impls.resize(0);
-            m_impls.resize(GetCount());
-        }
-        else if ( m_impls[n] )
-        {
-            // Just return the existing display if we have it.
-            return m_impls[n];
-        }
-
-        m_impls[n] = CreateDisplay(n);
-        return m_impls[n];
-    }
+    wxObjectDataPtr<wxDisplayImpl> GetDisplay(unsigned n);
 
     // Return the primary display object, creating it if necessary.
     wxObjectDataPtr<wxDisplayImpl> GetPrimaryDisplay();
@@ -102,6 +71,9 @@ private:
 
     // On-demand populated vector of wxDisplayImpl objects.
     wxVector<wxObjectDataPtr<wxDisplayImpl> > m_impls;
+
+    // The count of connected displays retrieved on last display access.
+    unsigned m_countOnLastAccess = 0u;
 
     wxDECLARE_NO_COPY_CLASS(wxDisplayFactory);
 };
--- a/include/wx/private/display.h
+++ b/include/wx/private/display.h
@@ -90,7 +90,7 @@ protected:
     // if this is the case
     //
     // mark display as disconnected and return false otherwise
-    virtual bool UpdateDisplayChange(wxDisplayImpl *impl) const;
+    virtual bool UpdateDisplayChange(wxDisplayImpl &impl) const;
 
 private:
     // Delete all the elements of m_impls vector and clear it.
--- a/include/wx/private/display.h
+++ b/include/wx/private/display.h
@@ -47,7 +47,7 @@ public:
     virtual int GetFromWindow(const wxWindow *window);
 
     // Trigger recreation of wxDisplayImpl when they're needed the next time.
-    virtual void InvalidateCache() { UpdateDisplayChanges(); }
+    virtual void UpdateOnDisplayChange();
 
 protected:
     // create a new display object
@@ -59,15 +59,11 @@ protected:
     // if this is the case
     //
     // mark display as disconnected and return false otherwise
-    virtual bool UpdateDisplayChange(wxDisplayImpl &impl) const;
+    virtual bool UpdateOnDisplayChange(wxDisplayImpl &impl) const;
 
 private:
     // Delete all the elements of m_impls vector and clear it.
     void ClearImpls();
-
-    // Trigger recreation of wxDisplayImpl or update its properties if
-    // a reference to it is still in scope.
-    void UpdateDisplayChanges();
 
     // On-demand populated vector of wxDisplayImpl objects.
     wxVector<wxObjectDataPtr<wxDisplayImpl> > m_impls;
--- a/include/wx/private/display.h
+++ b/include/wx/private/display.h
@@ -162,14 +162,14 @@ public:
 
 protected:
     // create the object providing access to the display with the given index
-    wxDisplayImpl(unsigned n) : m_index(n), m_isConnected(true) { }
+    wxDisplayImpl(unsigned n) : m_index(n) { }
 
 
     // the index of this display (0 is always the primary one)
     unsigned m_index;
 
     // true, if this display is still connected physically to system
-    bool m_isConnected;
+    bool m_isConnected = true;
 
     friend class wxDisplayFactory;
 
--- a/include/wx/private/display.h
+++ b/include/wx/private/display.h
@@ -127,10 +127,10 @@ public:
 #endif // wxUSE_DISPLAY
 
     // return true, if this display is still connected physically to system
-    virtual bool IsConnected() const { return m_isConnected; }
+    bool IsConnected() const { return m_isConnected; }
 
     // indicate that this display is not connected to system anymore
-    virtual void Disconnect() { m_isConnected = false; }
+    void Disconnect() { m_isConnected = false; }
 
 protected:
     // create the object providing access to the display with the given index
--- a/include/wx/private/display.h
+++ b/include/wx/private/display.h
@@ -26,7 +26,7 @@ public:
 
     // Create the display if necessary using CreateDisplay(), otherwise just
     // get it from cache.
-    wxDisplayImpl* GetDisplay(unsigned n)
+    wxObjectDataPtr<wxDisplayImpl> GetDisplay(unsigned n)
     {
         // Normally, m_impls should be cleared if the number of displays in the
         // system changes because InvalidateCache() must be called. However in
@@ -60,7 +60,7 @@ public:
     }
 
     // Return the primary display object, creating it if necessary.
-    wxDisplayImpl* GetPrimaryDisplay();
+    wxObjectDataPtr<wxDisplayImpl> GetPrimaryDisplay();
 
     // get the total number of displays
     virtual unsigned GetCount() = 0;
@@ -78,7 +78,7 @@ public:
     virtual int GetFromWindow(const wxWindow *window);
 
     // Trigger recreation of wxDisplayImpl when they're needed the next time.
-    virtual void InvalidateCache() { ClearImpls(); }
+    virtual void InvalidateCache() { UpdateDisplayChanges(); }
 
 protected:
     // create a new display object
@@ -86,12 +86,22 @@ protected:
     // it can return a null pointer if the display creation failed
     virtual wxDisplayImpl *CreateDisplay(unsigned n) = 0;
 
+    // check if the input display is still connected and update properties
+    // if this is the case
+    //
+    // mark display as disconnected and return false otherwise
+    virtual bool UpdateDisplayChange(wxDisplayImpl *impl) const;
+
 private:
     // Delete all the elements of m_impls vector and clear it.
     void ClearImpls();
 
+    // Trigger recreation of wxDisplayImpl or update its properties if
+    // a reference to it is still in scope.
+    void UpdateDisplayChanges();
+
     // On-demand populated vector of wxDisplayImpl objects.
-    wxVector<wxDisplayImpl*> m_impls;
+    wxVector<wxObjectDataPtr<wxDisplayImpl> > m_impls;
 
     wxDECLARE_NO_COPY_CLASS(wxDisplayFactory);
 };
@@ -100,7 +110,7 @@ private:
 // wxDisplayImpl: base class for all wxDisplay implementations
 // ----------------------------------------------------------------------------
 
-class wxDisplayImpl
+class wxDisplayImpl : public wxObjectRefData
 {
 public:
     // virtual dtor for this base class
@@ -144,14 +154,22 @@ public:
     virtual bool ChangeMode(const wxVideoMode& mode) = 0;
 #endif // wxUSE_DISPLAY
 
+    // return true, if this display is still connected physically to system
+    virtual bool IsConnected() const { return m_isConnected; }
+
+    // indicate that this display is not connected to system anymore
+    virtual void Disconnect() { m_isConnected = false; }
+
 protected:
     // create the object providing access to the display with the given index
-    wxDisplayImpl(unsigned n) : m_index(n) { }
+    wxDisplayImpl(unsigned n) : m_index(n), m_isConnected(true) { }
 
 
     // the index of this display (0 is always the primary one)
-    const unsigned m_index;
+    unsigned m_index;
 
+    // true, if this display is still connected physically to system
+    bool m_isConnected;
 
     friend class wxDisplayFactory;
 
--- a/interface/wx/display.h
+++ b/interface/wx/display.h
@@ -208,6 +208,21 @@ public:
     static wxSize GetStdPPI();
 
     /**
+        Returns @true if the display has not been unplugged.
+
+        This function is only implemented on MSW. After the display is unplugged,
+        or if the corresponding laptop lid is closed, the display object becomes
+        invalid on the next event queue execution if the object is still in scope.
+        All display property access functions will then return fallback values.
+
+        To receive notifications about display unplugging or other display changes,
+        connect a top-level window to the wxDisplayChangedEvent.
+
+        @since 3.3.0
+    */
+    bool IsConnected() const;
+
+    /**
         Returns @true if the display is the primary display. The primary
         display is the one whose index is 0.
     */
--- a/interface/wx/display.h
+++ b/interface/wx/display.h
@@ -210,15 +210,16 @@ public:
     /**
         Returns @true if the display has not been unplugged.
 
-        This function is only full implemented in MSW and just always returns
-        @true under the other platforms currently.
-
-        Under MSW, it returns @false if the physical display that this C++
-        object corresponds to has become unavailable, e.g. because it was
+        On MSW, this function returns @false if the physical display that this
+        C++ object corresponds to has become unavailable, e.g. because it was
         unplugged or otherwise removed from the system. Such disconnected
         display object can still be used by the program, but all its accessor
         functions return special fallback values (e.g. 0 for the width and
         height).
+
+        This function is fully implemented only on MSW. On Apple devices, it always
+        returns @false when a display configuration change is detected. On other
+        platforms, it currently always returns @true.
 
         To receive notifications about display unplugging or other display changes,
         connect a top-level window to the wxDisplayChangedEvent.
--- a/interface/wx/display.h
+++ b/interface/wx/display.h
@@ -210,10 +210,15 @@ public:
     /**
         Returns @true if the display has not been unplugged.
 
-        This function is only implemented on MSW. After the display is unplugged,
-        or if the corresponding laptop lid is closed, the display object becomes
-        invalid on the next event queue execution if the object is still in scope.
-        All display property access functions will then return fallback values.
+        This function is only full implemented in MSW and just always returns
+        @true under the other platforms currently.
+        
+        Under MSW, it returns @false if the physical display that this C++
+        object corresponds to has become unavailable, e.g. because it was
+        unplugged or otherwise removed from the system. Such disconnected
+        display object can still be used by the program, but all its accessor
+        functions return special fallback values (e.g. 0 for the width and
+        height).
 
         To receive notifications about display unplugging or other display changes,
         connect a top-level window to the wxDisplayChangedEvent.
--- a/interface/wx/display.h
+++ b/interface/wx/display.h
@@ -212,7 +212,7 @@ public:
 
         This function is only full implemented in MSW and just always returns
         @true under the other platforms currently.
-        
+
         Under MSW, it returns @false if the physical display that this C++
         object corresponds to has become unavailable, e.g. because it was
         unplugged or otherwise removed from the system. Such disconnected
--- a/src/common/dpycmn.cpp
+++ b/src/common/dpycmn.cpp
@@ -269,7 +269,7 @@ void wxDisplayFactory::UpdateDisplayChanges()
         if ( UpdateDisplayChange(*impl) )
             impls.push_back(impl);
     }
-    m_impls = impls;
+    m_impls = std::move(impls);
 }
 
 wxObjectDataPtr<wxDisplayImpl> wxDisplayFactory::GetDisplay(unsigned n)
--- a/src/common/dpycmn.cpp
+++ b/src/common/dpycmn.cpp
@@ -257,22 +257,53 @@ void wxDisplayFactory::ClearImpls()
 
 void wxDisplayFactory::UpdateDisplayChanges()
 {
-    const unsigned newCount = GetCount();
-    wxVector<wxObjectDataPtr<wxDisplayImpl> > impls((size_t)newCount);
+    m_countOnLastAccess = GetCount();
+    wxVector<wxObjectDataPtr<wxDisplayImpl> > impls;
+    impls.reserve(m_countOnLastAccess);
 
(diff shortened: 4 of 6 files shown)


src/common/dpycmn.cpp
```diff
@@ -216,33 +242,55 @@ bool wxDisplay::ChangeMode(const wxVideoMode& mode)
...
+
 void wxDisplayFactory::ClearImpls()
 {
+    m_impls.clear();
+}
+
+void wxDisplayFactory::UpdateDisplayChanges()
+{
+    const unsigned newCount = GetCount();
+    wxVector<wxObjectDataPtr<wxDisplayImpl> > impls((size_t)newCount);
```
vadz, line 261: I'm not sure if creating new vector of fixed size is correct, it looks like we could end up with empty objects at the end of it. If we used default ctor and `reserve(newCount)` here and `push_back()` below, we could be certain that `impls` only contains valid elements at the end.

src/common/dpycmn.cpp
```diff
@@ -216,33 +242,55 @@ bool wxDisplay::ChangeMode(const wxVideoMode& mode)
...
     for ( size_t n = 0; n < m_impls.size(); ++n )
     {
-        // It can be null, that's ok.
-        delete m_impls[n];
-    }
+        wxObjectDataPtr<wxDisplayImpl> &impl = m_impls[n];
 
-    m_impls.clear();
+        // There are wxDisplay objects referring to impl still in scope.
+        if ( impl && impl->GetRefCount() > 1 )
```
vadz, line 268: Testing for reference count is a code smell, do we really have to do it?

src/msw/display.cpp
```diff
@@ -653,6 +668,21 @@ int wxDisplayFactoryMSW::GetFromWindow(const wxWindow *window)
 #endif
 }
 
+bool wxDisplayFactoryMSW::UpdateDisplayChange(wxDisplayImpl *impl) const
+{
+    wxDisplayMSW *implMsw = dynamic_cast<wxDisplayMSW *>(impl);
```
vadz, line 673: I think this could be `static_cast`, we're not going to have any other kind of `wxDisplayImpl` in wxMSW code.

include/wx/display.h
```diff
@@ -79,6 +81,9 @@ class WXDLLIMPEXP_CORE wxDisplay
     // return true if the object was initialized successfully
     bool IsOk() const { return m_impl != nullptr; }
 
+    // return true if the display has not been unplugged yet
+    bool IsConnected() const;
```
vadz, line 85: The new function should be documented in `interface/wx/display.h` and it would be also nice to happen the invalidation of displays which happens on their disconnection in the class description there too.

include/wx/private/display.h
```diff
@@ -78,20 +78,30 @@ class wxDisplayFactory
...
     // create a new display object
     //
     // it can return a null pointer if the display creation failed
     virtual wxDisplayImpl *CreateDisplay(unsigned n) = 0;
 
+    // check if the input display is still connected and update properties
+    // if this is the case
+    //
+    // mark display as disconnected and return false otherwise
+    virtual bool UpdateDisplayChange(wxDisplayImpl *impl) const;
```
vadz, line 93: This should probably take a reference and not a pointer, as it can never be null.

include/wx/private/display.h
```diff
@@ -144,14 +154,22 @@ class wxDisplayImpl
...
-    wxDisplayImpl(unsigned n) : m_index(n) { }
+    wxDisplayImpl(unsigned n) : m_index(n), m_isConnected(true) { }
 
 
     // the index of this display (0 is always the primary one)
-    const unsigned m_index;
+    unsigned m_index;
 
+    // true, if this display is still connected physically to system
+    bool m_isConnected;
```
vadz, line 172: We prefer to initialize members in their declarations in the new code:
```suggestion
    bool m_isConnected = true;
```

(and then its initialization in the ctor above becomes unnecessary, of course)

src/common/dpycmn.cpp
```diff
@@ -125,52 +127,70 @@ wxDisplay::wxDisplay(const wxWindow* window)
...
+}
 
 wxRect wxDisplay::GetGeometry() const
 {
     wxCHECK_MSG( IsOk(), wxRect(), wxT("invalid wxDisplay object") );
 
-    return m_impl->GetGeometry();
+    if ( m_impl->IsConnected() )
+        return m_impl->GetGeometry();
+    return wxRect();
```
vadz, line 143: This is not critical, but I'd put the return in exceptional case first, for consistency with the precondition check above, i.e.

```suggestion
    if ( !m_impl->IsConnected() )
        return wxRect();
        
    return m_impl->GetGeometry();
```

(and the same for all the other functions, of course)

vadz (commented):
Thank you!

This looks mostly good, but there are 2 issues:

1. Documentation needs to be updated.
2. The code in wxMSW update function doesn't look obviously correct, it would be nice to make it more so, if possible.

Thanks again!

src/common/dpycmn.cpp
Marcel-Pham: As far as I understand the design of `wxDisplayFactory::m_impls`, empty objects are allowed because the display implementation objects are created delayed on first access.

After function call, the size of the array must equal the new number of displays, even if some elements are kept empty. If not the `wxDisplayFactory::GetDisplay` function will clear the whole container again in `include/wx/private/display.h`, line 49-50 and create the empty objects at this place.

vadz: Sorry, I just don't have time to dive into this right now, but just the fact that you need to make an effort to understand whether it's correct or not is not great IMHO. It would be better if we could (easily) ensure that there are no empty objects at the end of the vector to begin with, rather than checking that they're valid.

Marcel-Pham: Sorry, this were the lines of code I reffered to which didn't came from me 
```cpp
 wxObjectDataPtr<wxDisplayImpl> GetDisplay(unsigned n)
    {
        // Normally, m_impls should be cleared if the number of displays in the
        // system changes because InvalidateCache() must be called. However in
        // some ports (e.g. Mac right now, see #18318), cache invalidation never
        // happens, so we can end up with m_impls size being out of sync with
        // the actual number of monitors. Compensate for this here by checking
        // if the index is invalid and invalidating the cache at least in this
        // case.
        //
        // Note that this is still incorrect because we continue using outdated
        // information if the first monitor is disconnected, for example. The
        // only real solution is to ensure that InvalidateCache() is called,
        // but for now this at least avoids crashes when a new display is
        // connected.
        if ( n >= m_impls.size() )
        {
            // This strange two-step resize is done to clear all the existing
            // elements: they may not be valid any longer if the number of
            // displays has changed.
            m_impls.resize(0);
            m_impls.resize(GetCount());
        }
        else if ( m_impls[n] )
        {
            // Just return the existing display if we have it.
            return m_impls[n];
        }

        m_impls[n] = CreateDisplay(n);
        return m_impls[n];
    }
```
So suppose we have 4 displays and the user only retrieves the 2nd display, the array `m_impl` will be in this state:
`(empty, 2nd, empty, empty)`
So there will be always empty objects on multi-display setups at the beginning or the end of the array and we have to support this.

I don't want to argue about this... There are two possible solutions to resolve this:

1. Adding more comments to the function, since it seems obviously not to be clear enough
2. Breaking the concept of delayed creation in `UpdateDisplayChanges` and calling `CreateDisplay` on each empty object

But a smaller array size as the number of displays will obviously break the whole implementation...

src/common/dpycmn.cpp
Marcel-Pham: Fixed in dd3e5e9.

include/wx/private/display.h
Marcel-Pham: Fixed in 74f59d2.

include/wx/private/display.h
Marcel-Pham: Fixed in 00b5385.

src/msw/display.cpp
Marcel-Pham: Fixed in 00b5385.

include/wx/display.h
Marcel-Pham: Documentatin added in e847b2f. Please have a look on it, since I'm not a native English speaker. Thanks.

src/common/dpycmn.cpp
Marcel-Pham: Fixed in 7d459c3.

src/common/dpycmn.cpp
Marcel-Pham: Please have a look on 7d459c3. Since `m_impls` is only used in 3 functions, I decided to change its sematic.

The array is no lookup table anymore. This means that we have to iterate over all elements to get a display with a specific index, but this shouldn't be a big deal.

This means also that no empty objects are required anymore to cover the whole lookup range.

The functional behavior should be exactly the same as before, also on the other platforms.

interface/wx/display.h
```diff
@@ -207,6 +207,21 @@ class wxDisplay
      */
     static wxSize GetStdPPI();
 
+    /**
+        Returns @true if the display has not been unplugged.
+
+        This function is only implemented on MSW. After the display is unplugged,
+        or if the corresponding laptop lid is closed, the display object becomes
+        invalid on the next event queue execution if the object is still in scope.
+        All display property access functions will then return fallback values.
```
vadz, line 216: I don't think it's useful to distinguish between unplugging a display and closing the laptop lid, but I believe it would be useful to document the behaviour of this function under the other platforms too, so I'd say:

```suggestion
        This function is only full implemented in MSW and just always returns
        @true under the other platforms currently.
        
        Under MSW, it returns @false if the physical display that this C++
        object corresponds to has become unavailable, e.g. because it was
        unplugged or otherwise removed from the system. Such disconnected
        display object can still be used by the program, but all its accessor
        functions return special fallback values (e.g. 0 for the width and
        height).
```

include/wx/private/display.h
```diff
@@ -144,14 +126,22 @@ class wxDisplayImpl
     virtual bool ChangeMode(const wxVideoMode& mode) = 0;
 #endif // wxUSE_DISPLAY
 
+    // return true, if this display is still connected physically to system
+    virtual bool IsConnected() const { return m_isConnected; }
+
+    // indicate that this display is not connected to system anymore
+    virtual void Disconnect() { m_isConnected = false; }
+
```
vadz, line 134: Minor, but it doesn't seem like either of these functions needs to be `virtual`.

As it's a private class, we can always add `virtual` later if necessary, but it's better not to add it without need, so I'd just remove it:

```suggestion
    // return true, if this display is still connected physically to system
    bool IsConnected() const { return m_isConnected; }

    // indicate that this display is not connected to system anymore
    void Disconnect() { m_isConnected = false; }

```

include/wx/private/display.h
```diff
@@ -78,20 +47,33 @@ class wxDisplayFactory
     virtual int GetFromWindow(const wxWindow *window);
 
     // Trigger recreation of wxDisplayImpl when they're needed the next time.
-    virtual void InvalidateCache() { ClearImpls(); }
+    virtual void InvalidateCache() { UpdateDisplayChanges(); }
```
vadz, line 50: Does it really make sense to have both `InvalidateCache()` and `UpdateDisplayChanges()`? They seem to be one and the same to me and it just makes things more confusing...

Could we just put the new code in `InvalidateCache()` itself?

Or, if you don't like its name, we could rename it to `UpdateOnDisplayChange()` (I think the extra "On" makes the name much more clear). But I don't see any good reason to have 2 functions — please let me know if I'm missing something?

src/common/dpycmn.cpp
```diff
@@ -216,33 +242,86 @@ bool wxDisplay::ChangeMode(const wxVideoMode& mode)
...
     {
-        // It can be null, that's ok.
-        delete m_impls[n];
+        wxObjectDataPtr<wxDisplayImpl> &impl = m_impls[n];
+
+        // Try to update display state or mark it as disconnected.
+        if ( UpdateDisplayChange(*impl) )
+            impls.push_back(impl);
     }
+    m_impls = impls;
```
vadz, line 272: Very minor but this could be made a bit more efficient by doing

```suggestion
    m_impls = std::move(impls);
```

include/wx/private/display.h
```diff
@@ -78,20 +47,33 @@ class wxDisplayFactory
...
     // create a new display object
     //
     // it can return a null pointer if the display creation failed
     virtual wxDisplayImpl *CreateDisplay(unsigned n) = 0;
 
+    // check if the input display is still connected and update properties
+    // if this is the case
+    //
+    // mark display as disconnected and return false otherwise
+    virtual bool UpdateDisplayChange(wxDisplayImpl &impl) const;
```
vadz, line 62: Sorry, but I've realized another thing which bothers me here: why is this function member of `wxDisplayFactory` and not a (virtual) function in `wxDisplayImpl`? Moving it there would avoid the need for any casts (dynamic or static) in its wxMSW implementation.

I would also rename it to `UpdateOnDisplayChange()`.

src/common/dpycmn.cpp
```diff
@@ -216,33 +242,86 @@ bool wxDisplay::ChangeMode(const wxVideoMode& mode)
...
+    impls.reserve(m_countOnLastAccess);
+
     for ( size_t n = 0; n < m_impls.size(); ++n )
     {
-        // It can be null, that's ok.
-        delete m_impls[n];
+        wxObjectDataPtr<wxDisplayImpl> &impl = m_impls[n];
+
+        // Try to update display state or mark it as disconnected.
+        if ( UpdateDisplayChange(*impl) )
```
vadz, line 269: To continue my other comment, this should basically just become

```suggestion
        if ( impl->UpdateOnDisplayChange() )
```

include/wx/private/display.h
```diff
@@ -78,20 +47,33 @@ class wxDisplayFactory
...
+    // Trigger recreation of wxDisplayImpl or update its properties if
+    // a reference to it is still in scope.
+    void UpdateDisplayChanges();
+
     // On-demand populated vector of wxDisplayImpl objects.
-    wxVector<wxDisplayImpl*> m_impls;
+    wxVector<wxObjectDataPtr<wxDisplayImpl> > m_impls;
+
+    // The count of connected displays retrieved on last display access.
+    unsigned m_countOnLastAccess = 0u;
```
vadz, line 76: Sorry, but I just can't figure out what this is and why do we need it. It seems to store the number of displays before the last change, but why?

E.g. consider that we have a system with 2 displays, then 1 gets disconnected. `m_countOnLastAccess` gets set to 2 in `UpdateDisplayChanges()` and then we only invalidate the cache if `GetDisplay(n)` is called with `n >= 2` but clearly we should be invalidating it if it's called with `n >= 1`, where 1 is the number of currently known displays? And, if so, what do we need `m_countOnLastAccess` for when we can just compare `n` with `m_impls.size()`?

vadz (commented):
Thank you for your continued work on this!

I think there is at least one point to improve: the function checking for display validity should clearly (IMO) be in `wxDisplayImpl` itself and not in `wxDisplayFactory` as explained below.

It also seems like we may not need `m_countOnLastAccess` but perhaps I'm missing something here.

include/wx/private/display.h
Marcel-Pham: > why is this function member of `wxDisplayFactory` and not a (virtual) function in `wxDisplayImpl`?

**From the viewpoint of design pattern application:**
`wxDisplayMSW` stores the display handle and further properties to implement the `wxDisplay` interface. But is has not the knowledge, where these information came from, i.e. how to get the handle and properties.

Retrieving this information is the task of `wxDisplayFactoryMSW`. Since on updapte we have to get the same information from the same source at the same way as on creation, it would be a good idea to keep this task in the responsibility of `wxDisplayFactoryMSW`.

**To be more concrete in the case of MSW:**

- On creation we need to call `::EnumDisplaySettings` to get the handles and indices of each available display and call `::GetMonitorInfo` to retrieve the properties. This is done by the factory.
- On update we need to call `::EnumDisplaySettings` to check, if the handle is still available and if the index has changed and call `::GetMonitorInfo` to update the properties. Do you really want me to move this implementation to `wxDisplayMSW`?

vadz: Sorry, I'm not sure we're speaking about the same thing. `::EnumDisplaySettings()` call should, of course, remain in the factory. But updating the state of an individual display object is clearly a responsibility of the display object itself and not of the factory, which is why I think that this function (not `UpdateDisplayChanges()`) should be its member.

Marcel-Pham: Hm.. now I am confused.
If the individual display object is not allowed to call `::EnumDisplaySettings()` and has no access to `wxDisplayFactoryMSW::m_displays`, where the result of `::EnumDisplaySettings()` is stored, in which form and from which source should the display object receive the target information for update?

This is the current implementation:
```C++
bool wxDisplayFactoryMSW::UpdateDisplayChange(wxDisplayImpl &impl) const
{
    wxDisplayMSW &implMsw = static_cast<wxDisplayMSW &>(impl);

    for ( size_t n = 0; n < m_displays.size(); ++n )
    {
        if (implMsw.UpdateStateIfHandleMatches((unsigned) n, m_displays[n]))
            return true;
    }

    impl.Disconnect();
    return false;
}
```
In this function, the display receives the index and display properties from `m_displays[n]`. If I move the implementation to a new function `wxDisplayMSW::UpdateDisplayChange`, I will have no access to `wxDisplayFactoryMSW::m_displays` anymore...
Do you have a specific solution in mind, how to solve this?

vadz: OK, I see the problem, thanks: we need access to all `HMONITOR`s to be able to check if the display is still connected. I think we might be able to use `::GetMonitorInfo()` to check for the monitor validity instead, but it sure looks a bit wrong to not be able to write this without using external calls.

Otherwise we need to pass `m_displays` to `wxDisplayImpl::UpdateOnDisplayChange()` somehow. Using `const void* data` for it would work but would still require a cast, of course and I'm not sure if it's really going to be much better than the current version.

If you decide to keep the current version, could we please at least replace `UpdateStateIfHandleMatches(unsigned, wxDisplayInfo)` with `UpdateOnDisplayChange(vector<wxDisplayInfo>)`, i.e. move the loop inside `wxDisplayMSW`? This would at least avoid having to pass the index and the object separately around which definitely feels awkward.

include/wx/private/display.h
Marcel-Pham: Yes you are right. I would agree, that we have to improve this somehow.
But since the behaviour always have been inconsistent or even wrong esp. on Mac we have to discuss about the right solution.
So again this was initially the function we are talking about:

```C++
wxObjectDataPtr<wxDisplayImpl> GetDisplay(unsigned n)
    {
        // Normally, m_impls should be cleared if the number of displays in the
        // system changes because InvalidateCache() must be called. However in
        // some ports (e.g. Mac right now, see #18318), cache invalidation never
        // happens, so we can end up with m_impls size being out of sync with
        // the actual number of monitors. Compensate for this here by checking
        // if the index is invalid and invalidating the cache at least in this
        // case.
        //
        // Note that this is still incorrect because we continue using outdated
        // information if the first monitor is disconnected, for example. The
        // only real solution is to ensure that InvalidateCache() is called,
        // but for now this at least avoids crashes when a new display is
        // connected.
        if ( n >= m_impls.size() )
        {
            // This strange two-step resize is done to clear all the existing
            // elements: they may not be valid any longer if the number of
            // displays has changed.
            m_impls.resize(0);
            m_impls.resize(GetCount());
        }
        else if ( m_impls[n] )
        {
            // Just return the existing display if we have it.
            return m_impls[n];
        }
        m_impls[n] = CreateDisplay(n);
        return m_impls[n];
    }
```
I think the first if-case and especially its condition is the problem here.
So why did the initial author add this if-case? I think the answer is

1. to preinitialize the lookup-table `m_impls` on first access. This is not required anymore, since we are using `push_back` instead of direct indexing right now.
2. to handle potentially access error esp. on Mac, if the number of displays increases. This is not required anymore, since we iterate over the `m_impls` instead of doing the lookup access right now.
3. to update the cached displays in the obvious case, when the display size increases, at the time, when this case is anyway detected and handled, i.e. on the access to the new display.

I think the comment
```C++
// Note that this is still incorrect because we continue using outdated
// information if the first monitor is disconnected, for example.
```
already mentions the case your are describing right now.

> And, if so, what do we need `m_countOnLastAccess` for when we can just compare `n` with `m_impls.size()`?

The reason, why I replaced `m_impls.size()` with `m_countOnLastAccess` is that `m_impls.size()` is not the number of displays anymore, because I removed the empty objects from `m_impls`. So this suggestion doesn't make sense to me at all.

But I agree that in the new implementation the if-case condition is even more confusing, even if the functionality stays the same as before.

We can remove it completely without having a program crash, since point 1 and 2 is solved otherwise now.
But then no cache update is performed at all on Mac.

We can replace it with `if ( GetCount() !=  m_countOnLastAccess )`, then the cache update is done every time immediately, when the display count changes. Then the behaviour would be less inpredictable at least. But it may be still wrong on some ports if the display properties change but not the count.

What is your opinion? Sorry for the long comment...

include/wx/private/display.h
Marcel-Pham: Function name updated to `UpdateOnDisplayChange` in cdc7987

include/wx/private/display.h
Marcel-Pham: I kept the current version and added the proposed changes from your last comment in cdc7987 .

src/common/dpycmn.cpp
Marcel-Pham: Function name updated to `UpdateOnDisplayChange` in cdc7987

include/wx/private/display.h
vadz: Hmm, so we've replaced the number of the displays with a separate variable now. I'm not really sure this is better, why do you think it is preferable to do it like this?

BTW, I think cache invalidation does happen in wxOSX now. It still doesn't seem to happen in wxGTK, however...

Marcel-Pham: > Hmm, so we've replaced the number of the displays with a separate variable now. I'm not really sure this is better, why do you think it is preferable to do it like this?

I introduced `m_countOnLastAccess` because the last applied number of displays is not stored anymore.

This is the current master-branch state of `GetDisplay` (with modified code comments):
```C++
wxDisplayImpl *GetDisplay(unsigned n)
{  
    //  We compare n with the last applied number of display here, which is stored in m_impls.size()
    if ( n >= m_impls.size() )
    {           
        m_impls.resize(0); // <- Here we did the cache invalidation, but not on all required cases.
        m_impls.resize(GetCount()); // <- Look here, m_impls.size() is the last applied number of displays now.
    }
    else if ( m_impls[n] )
    {       
         return m_impls[n];
    }
    m_impls[n] = CreateDisplay(n);
    return m_impls[n];
}
```
This is the last commit I did in my PR (same functional bahaviour as above):
```C++
wxObjectDataPtr<wxDisplayImpl> GetDisplay(unsigned n)
{  
    // We compare n with the last applied number of display here, which is stored in m_countOnLastAccess 
    if ( n >= m_countOnLastAccess )
    {
        // On cache invalidation we assign m_countOnLastAccess = GetCount()
        // But m_impls is NOT resized to GetCount() anymore, because you claimed that having empty objects in the array is a bad design.
        UpdateOnDisplayChange();
    }

    for ( size_t i = 0; i < m_impls.size(); ++i )
    {
        if ( m_impls[i]->GetIndex() == n ) // Here we have no empty objects, so index lookup is not possible anymore
            return m_impls[i];
    }
    wxObjectDataPtr<wxDisplayImpl> impl(CreateDisplay(n));
    if ( impl )
        m_impls.push_back(impl); // The size is the count of accessed displays not the count of displays in system now
    return impl;
```
**1. Improvement suggestion: Drop if-case**

```C++
wxObjectDataPtr<wxDisplayImpl> GetDisplay(unsigned n)
{  
    // Drop the if-case, so m_countOnLastAccess can be removed. But no cache invalidation is done anymore for ports which do not invalidate on corresponding system message
    for ( size_t i = 0; i < m_impls.size(); ++i )
    {
        if ( m_impls[i]->GetIndex() == n ) 
            return m_impls[i];
    }
    wxObjectDataPtr<wxDisplayImpl> impl(CreateDisplay(n));
    if ( impl )
        m_impls.push_back(impl); 
    return impl;
```

**2. Improvement suggestion: Change if-case condition**

```C++
wxObjectDataPtr<wxDisplayImpl> GetDisplay(unsigned n)
{  
    // Do cache invalidation only if count of displays changed.
    // Depending on the concrete port implementation some cases mey still not be handled, if the count of displays does not change, but properties changed.
    if ( GetCount() != m_countOnLastAccess )
    {
          // On cache invalidation we assign m_countOnLastAccess = GetCount()
          UpdateOnDisplayChange();
    }

    for ( size_t i = 0; i < m_impls.size(); ++i )
    {
        if ( m_impls[i]->GetIndex() == n ) 
            return m_impls[i];
    }
    wxObjectDataPtr<wxDisplayImpl> impl(CreateDisplay(n));
    if ( impl )
        m_impls.push_back(impl); 
    return impl;
```

**3. Improvement suggestion: Do not chache at all, if no system update message supported**

```C++
wxObjectDataPtr<wxDisplayImpl> GetDisplay(unsigned n)
{  
   // Here we introduce a new virtual function. Each port has to tell us if update by system message is possible.
    if ( SystemNotifiesOnDisplayChange() )
    {
        for ( size_t i = 0; i < m_impls.size(); ++i )
        {
            if ( m_impls[i]->GetIndex() == n ) 
                return m_impls[i];
        }
        wxObjectDataPtr<wxDisplayImpl> impl(CreateDisplay(n));
        if ( impl )
            m_impls.push_back(impl); 
        return impl;
    }
   // Else do not cache anything at all.
    else 
        return CreateDisplay(n);
}
```

vadz: Sorry, I'm afraid I sent you on a wild goose chase with my comment about leaving empty objects in `m_impls`: I completely forgot that we _already_ had them there, even without your changes, and thought that this is something this PR introduced. Rethinking this now, I realize that there is no real problem with having empty objects in `m_impls`.

So if the only reason to introduce `m_countOnLastAccess` was to address my comment then it's not a good reason to do it. I'm sorry again for missing the big(ger) picture at first and wasting your time.

But it looks like the best would be to return to the version without `m_countOnLastAccess` (but still containing all the other modifications in the later versions). Do you think you could do this?

Also, I don't think we need to do anything special for the ports not calling `InvalidateCache()`, the right thing to do is to implement this for them. wxMSW and wxOSX already do it and for wxGTK we could use `GdkDisplay` `monitor-added` and `monitor-removed` signals to implement this too.

Marcel-Pham: >  I completely forgot that we already had them there, even without your changes, and thought that this is something this PR introduced.

Yes this was exactly what I was trying to tell you in my explanations... 

Ok, so I reverted `GetDisplay` to its orginal implementation state from master branch, removed `m_countOnLastAccess` and adjusted `UpdateOnDisplayChange` in 64e1f72.

vadz (commented):
Thanks for updating this and sorry again for missing the way how this code worked during the initial review.

I think this is broadly fine now and I'd really like to merge this a.s.a.p. so that it gets into the upcoming 3.3.0, but I still have a couple of questions below, could you please answer them? TIA!

interface/wx/display.h
```diff
@@ -207,6 +207,26 @@ class wxDisplay
      */
     static wxSize GetStdPPI();
 
+    /**
+        Returns @true if the display has not been unplugged.
+
+        This function is only full implemented in MSW and just always returns
+        @true under the other platforms currently.
```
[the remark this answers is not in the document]
Marcel-Pham: To be honest, I'm not sure about the exact behaviour on the other ports. I only have a test environment for MSW available, so I can't debug the behaviour on OSX or Linux.

In general if the port doesn't trigger `InvalidateCache` on operating system display change notification the function will always return `true`.
If the port implementation triggers `InvalidateCache` but doesn't overload the `UpdateOnDisplayChange` function it will always return `false`, even for the displays, which are still connected.

The quoted documentation exept was your change proposal. You told me later that for wxOSX display change notification is implemented but I can't find the implementation in `src\osx\core\display.cpp`. Maybe it is implemented somewhere outside or I missed the implementation.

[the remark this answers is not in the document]
Marcel-Pham: I found the following hit, when I searched the whole folder for `InvalidateCache` call:
```
\wxWidgets\src\osx\carbon\app.cpp (1 Treffer)
	Line 327:     wxDisplay::InvalidateCache();
```
So `IsConnected` will return `false` on all apple devices after system notification?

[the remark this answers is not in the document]
vadz: Yes, as I wrote above:

> AFAICS it will always return false for all displays created before a display change notification in wxOSX. 

IMO this is fine, it's better to return false when we're not sure rather than true, the application can always recreate `wxDisplay` object but using an invalid object would be bad.

vadz:
Thanks again for all your work here, I'm going to (squash)merge it soon.

interface/wx/display.h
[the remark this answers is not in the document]
Marcel-Pham: I updated the description in 86d5ac8.

vadz:
After one final review, I've done a few minor cosmetic changes, with the most important one being the renaming of one of `UpdateOnDisplayChange()` overloads to `RefreshOnDisplayChange()` as I thought it was confusing to use the same name for these 2 functions — I hope you don't mind, but please let me know if you do.

Sorry again for the confusion during the initial review and thanks a lot for fixing this bug!

Marcel-Pham:
Sorry, I did a last test with the final version in the master branch and discovered an error that crept in during the refactoring steps:

at src/msw/display.cpp, Line 510:

https://github.com/wxWidgets/wxWidgets/blob/826228430eff768441a845d8cfb86f06e998b821/src/msw/display.cpp#L510

This of course should be
```C++
if (  m_info.hmon == displays[n].hmon )
```

instead of 
```C++
if (  displays[n].hmon == displays[n].hmon )
```

What is the easiest way to fix this? Can you correct this directly in master branch? Thanks.

vadz:
Thanks for noticing this, I'll indeed push the fix for it directly to master soon.

PR #25409, opened 2025-05-16
Ensure that all ports call wxEventLoop::OnExit()

vadz:
Make ScheduleExit() non-virtual and call a new virtual DoStop() from it and update all ports to override the new virtual function and not override ScheduleExit() itself to ensure that the base class version, which calls OnExit(), is used.

See #25397.


vadz:
@AliKet I've updated wxQt but without really testing it, please let me know if I broke anything.

AliKet:
> @AliKet I've updated wxQt but without really testing it, please let me know if I broke anything.

LGTM. thanks.

vadz:
Applied in the commit above.

PR #25413, opened 2025-05-17
Some fixes related to RTL layout under wxMSW

AliKet:


src/msw/overlay.cpp
```diff
@@ -162,7 +162,18 @@ void wxOverlayImpl::Init(wxDC* dc, int , int , int , int )
...
+    {
+        m_window->ClientToScreen(&m_rect.x, &m_rect.y);
+    }
+    else // wxLayout_RightToLeft
+    {
+        RECT rc;
+        wxCopyRectToRECT(m_rect, rc);
+        wxMapWindowPoints(GetHwndOf(m_window), HWND_DESKTOP, &rc);
+        wxCopyRECTToRect(rc, m_rect);
+    }
```
AliKet, line 177: See this [answer](https://stackoverflow.com/questions/15734528/client-rectangle-coordinates-on-screen#comment22378286_15734569) by [Raymond Chen](https://stackoverflow.com/users/902497/raymond-chen)

AliKet:
The second commit [7f1c4c4](https://github.com/wxWidgets/wxWidgets/pull/25413/commits/7f1c4c446dc14c11f1c6c18e0f4581552fea6aa7) is only needed under `wxMSW`! I'll fix it...

src/generic/scrlwing.cpp
```diff
@@ -984,6 +984,14 @@ void wxScrollHelperBase::HandleOnMouseLeave(wxMouseEvent& event)
         if ( !m_targetWindow->HasScrollbar(orient) )
             return;
 
+#ifdef __WXMSW__
```
vadz, line 987: Is this really MSW-specific? It looks like we might need to do it for all ports...

src/msw/overlay.cpp
```diff
@@ -162,7 +162,18 @@ void wxOverlayImpl::Init(wxDC* dc, int , int , int , int )
 
     // The rectangle must be in screen coordinates
     m_rect = m_window->GetClientSize();
-    m_window->ClientToScreen(&m_rect.x, &m_rect.y);
+
+    if ( m_window->GetLayoutDirection() == wxLayout_LeftToRight )
+    {
+        m_window->ClientToScreen(&m_rect.x, &m_rect.y);
```
vadz, line 168: Should we perhaps always use the version in the `else` branch?

Or, alternatively, change the basic `ClientToScreen()` itself?

vadz (commented):
Thanks for fixing this, I am fine with merging this as is because I guess you did test it and this fixes actual problems, but it's not really clear why are these fixes only needed here/only for MSW...

src/generic/scrlwing.cpp
vadz: I think the code in `wxAutoScrollTimer::Notify()` doesn't handle RTL windows correctly. But this still doesn't explain why it works in wxGTK.

IMO it would be highly desirable to understand what is really going on here and where does the discrepancy in behaviour between different ports really come from in order to fix it at the source.

AliKet: Indeed; the problem was in `wxWindowMSW::GenerateMouseLeave()` fixed now by 
b419237

vadz (approved):
Nice, thanks, this change makes much more sense and I'll merge it soon.

I still have a question about whether `ClientToScreen()` in overlay code should work, but this is less important and we can leave it for later.

Thanks again!

vadz:
Oops, sorry and thanks for noticing this. Re-reverting this now.

PR #25424, opened 2025-05-19
Update MSVS build files for vs14x only builds.

doscott:
I'm not sure what happened on the previous pull request. I'm fairly certain this one is OK, but I am unable to build to test.

### DIFF
--- a/build/tools/msvs/officialbuild.bat
+++ b/build/tools/msvs/officialbuild.bat
@@ -28,39 +28,8 @@ if "%1" == "vc14x" (
   set compvers=vc14x
   call "%VS140COMNTOOLS%VsDevCmd.bat"
 )
-if "%1" == "vc120" (
-  @echo Building for vc120 / vs2013
-  set comp=120
-  set compvers=vc120
-  call "%VS120COMNTOOLS%VsDevCmd.bat"
-)
-if "%1" == "vc110" (
-  @echo Building for vc110 / vs2012
-  set comp=110
-  set compvers=vc110
-  call "%VS110COMNTOOLS%VsDevCmd.bat"
-)
-if "%1" == "vc100" (
-  @echo Building for vc100 / vs2010
-  set comp=100
-  set compvers=vc100
-  if "%WINDOWS71SDK%" == "" set WINDOWS71SDK=C:\Program Files\Microsoft SDKs\Windows\v7.1\Bin\
-  call "%WINDOWS71SDK%SetEnv.cmd"
-)
-if "%1" == "vc90" (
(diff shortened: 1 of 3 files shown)


vadz:
LGTM, thanks!

build/tools/msvs/officialbuild.bat
```diff
@@ -28,66 +28,26 @@ if "%1" == "vc14x" (
...
-del %compvers%x86_Release.txt
 del %compvers%x64_Debug.txt
 del %compvers%x64_Release.txt
 
-if "%compvers%" == "vc14x" call "%VS140COMNTOOLS%..\..\VC\vcvarsall.bat" x64
-if "%compvers%" == "vc120" call "%VS120COMNTOOLS%..\..\VC\vcvarsall.bat" x86_amd64
-if "%compvers%" == "vc110" call "%VS110COMNTOOLS%..\..\VC\vcvarsall.bat" x86_amd64
-if "%compvers%" == "vc100" call "%WINDOWS71SDK%SetEnv.Cmd" /X64 /Release
-if "%compvers%" == "vc90"  call "%WINDOWS61SDK%SetEnv.Cmd" /X64 /Release
+if "%compvers%" == "vc14x" call "%VS140COMNTOOLS%..\..\VC\vcvarsall.bat" x64 10.0.17763.0
```
PBfordev, line 50: @doscott @vadz 
Not sure if hardcoding the winsdk_version here is the best. Perhaps we could add it as an optional argument to `officialbuild.bat`, which could be passed to `vcvarsall.bat`?

Additionally, we can check currently used SDK version inside `officialbuild.bat`, after calling  `vcvarsall.bat` and before running nmake, looking at the system environment variable `WindowsSDKVersion` and warn when the version is newer than the last known working one (not sure if possible within a batch file)...

Of course, I am not saying to do that for 3.3.0 release, just a suggestion for future consideration. I could try doing it.

vadz: Yes, hardcoding the SDK version is not great, but as these scripts are only used by Danny currently, I'm not too bothered by it. Your proposal looks good, however!

doscott: It could be passed in as an argument from buildall.bat. I don't have any time right now and I wouldn't object to someone else taking over the roll of making the official binaries. I will try to find time in the next few days to see if the new Windows 10 SDK works ok. It will also be interesting to see if a binary produced with the current SDK plays well when used with a VS2022 application built with the latest Windows 11 SDK that caused problems with VS2015.

Issue #25428, opened 2025-05-20
A tiny bug in `samples/widgets/widgets.cpp`


ChipZhang:
In the git master version, variable `imageList` here is changed from type `wxImageList *` (in v3.2) to type `wxVector`:

https://github.com/wxWidgets/wxWidgets/blob/05e5092ccb10e75686945f41b8851433158346eb/samples/widgets/widgets.cpp#L632-L635

But later, `imageList` is used in method `SetImageList()` (the same as v3.2, not updated). This code should be updated to use `SetImages()` instead. Or it causes compile error `cannot convert ‘wxVector<wxBitmapBundle>’ to ‘wxImageList*’`, if `#else` directive met (treebook not used):

https://github.com/wxWidgets/wxWidgets/blob/05e5092ccb10e75686945f41b8851433158346eb/samples/widgets/widgets.cpp#L707-L712

ChipZhang:
I confirm this fix works, both for `#if` or `#else`.

PR #25429, opened 2025-05-20
Fix a tiny bug in samples/widgets/widgets.cpp (imageList)

ChipZhang:
Fixes issue #25428


vadz:
Thanks for finding and fixing this!

PR #25430, opened 2025-05-20
Add missing includes in include/wx/longlong.h

SagaciousFish:
Added `#include "wx/txtstrm.h"` ~in `#if wxUSE_STREAMS` blocks~

Otherwise, it would lead to compile errors:
- wxTextInputStream/wxTextOutputStream has not been declared
- wxTextInputStream/wxTextOutputStream does not name a type

~p.s. The new lines are put in the conditional blocks to avoid unnecessary includes. Maintainers, please advise if the include lines should be put at the top of the file for simplicity.~ fixed thanks to the reminder from @lanurmi!

### DIFF
--- a/include/wx/longlong.h
+++ b/include/wx/longlong.h
@@ -260,6 +260,7 @@ public:
     wxString& operator<<(wxString&, const wxLongLong&);
 
 #if wxUSE_STREAMS
+    #include "wx/txtstrm.h"
     friend WXDLLIMPEXP_BASE
     class wxTextOutputStream& operator<<(class wxTextOutputStream&, const wxLongLong&);
     friend WXDLLIMPEXP_BASE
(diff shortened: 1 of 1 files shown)


lanurmi:
> p.s. The new lines are put in the conditional blocks to avoid unnecessary includes. Maintainers, please advise if the include lines should be put at the top of the file for simplicity.

I'm not a maintainer, but it cannot be a generally good idea to `#include` something *within* a class declaration.

SagaciousFish:
I was operating on 3.2.8 release, and it seems the master branch has evolved a bit and the checks do not seem to be compiling.

This PR can be closed if the modification does not fit into the current development. 

Hope this would be a helpful reminder though.

SagaciousFish:
> > p.s. The new lines are put in the conditional blocks to avoid unnecessary includes. Maintainers, please advise if the include lines should be put at the top of the file for simplicity.
> 
> I'm not a maintainer, but it cannot be a generally good idea to `#include` something _within_ a class declaration.

Yes, you are absolutely correct! I'm new to this library and blindly assumed preprocessors are not included in classes.

PR #25431, opened 2025-05-20
Add RTL support to wxGCDC when using GDI+ backend

### LINKED ISSUE #25427: wxGCDC doesn't respect the RTL layout when using GDI+  backend
### Description

To see the problem:

- Run the drawing demo (with GDI+ renderer): drawings are still LTR
- Try selecting some cells or columns in the griddemo: the selection is not correct

#### Possible fix:

This is dirty patch which solves the issue (not to be applied):
```C++
diff --git a/src/msw/graphics.cpp b/src/msw/graphics.cpp
index 7ebb09bb04..f64ba798a6 100644
--- a/src/msw/graphics.cpp
+++ b/src/msw/graphics.cpp
@@ -1929,6 +1929,9 @@ void wxGDIPlusContext::Init(Graphics* graphics, int width, int height, const Mat
     m_height = height;
     m_internalTransform = new Matrix();
 
+    m_context->ScaleTransform(-1, 1);
+    m_context->TranslateTransform(-width, 0);
+
     m_context->SetTextRenderingHint(TextRenderingHintSystemDefault);
     m_context->SetPixelOffsetMode(PixelOffsetModeHalf);
     m_context->SetSmoothingMode(SmoothingModeHighQuality);

```

  + GTK version: 
  + Which GDK  backend is used:
  + Desktop environment :
  + Current theme:

AliKet:
Closes #25427

### DIFF
--- a/src/msw/graphics.cpp
+++ b/src/msw/graphics.cpp
@@ -92,7 +92,7 @@ StringFormat* gs_drawTextStringFormat = nullptr;
 // Get the string format used for the text drawing and measuring functions:
 // notice that it must be the same one for all of them, otherwise the drawn
 // text might be of different size than what measuring it returned.
-inline StringFormat* GetDrawTextStringFormat()
+inline StringFormat* GetDrawTextStringFormat(wxLayoutDirection dir = wxLayout_Default)
 {
     if ( !gs_drawTextStringFormat )
     {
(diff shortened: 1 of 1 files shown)


vadz (commented):
Thanks, I didn't test this, but if it fixes the problem, let's apply it. Just please avoid `bool` parameters.

AliKet:
I have updated this PR as suggested, and in fact:
- The previouse version wasn't quite right.
- I tested the new version with the drawing sample (changing multiple parameters and transforms) and it worked correctly (just I don't know why we should multiply by 2 when drawing the text un-mirrored?)

AliKet:
**Drawing sample with this PR:**

Using GDI+:

drawing_rtl_gdiplus

Using DirectD2 (need to set [DWRITE_READING_DIRECTION_RIGHT_TO_LEFT](https://learn.microsoft.com/en-us/windows/win32/api/dwrite/ne-dwrite-dwrite_reading_direction#:~:text=DWRITE_READING_DIRECTION_RIGHT_TO_LEFT%20%3D%201%2C) to work properly):

drawing_rtl_direct2d

MaartenBent:
> just I don't know why we should multiply by 2 when drawing the text un-mirrored?

Because you need to compensate for the x-coordinate of `DrawString`.

For example if you draw text at `x=10`, rtl should put it at `x=-10`.
`wxDCMirrorDisabler` translates to `20`, then flips the x-axis so it becomes `-20`. And then `DrawString` draws it at `10`, which with the applied translation becomes `-10`.

vadz:
Thanks, I'm applying this without the "why" comment which seems to have been answered above.

PR #25432, opened 2025-05-20
Add missing 'override'

lanurmi:
This is the only line that prevented compilation using -Werror=suggest-override.

### DIFF
--- a/src/common/curbndl.cpp
+++ b/src/common/curbndl.cpp
@@ -61,7 +61,7 @@ public:
     {
     }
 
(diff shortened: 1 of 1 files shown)


vadz:
Oops, sorry for forgetting this one, will merge soon.

How comes our clang CI build didn't catch it?

vadz:
> > How comes our clang CI build didn't catch it?
> 
> I attempted to find a CI build that passes `-Wsuggest-override` or anything similar to the compiler, but couldn't find one. Could be I'm just bad at searching.

I forgot that it had to be explicitly enabled. I'll try doing it.

Issue #24977, opened 2024-11-27
Kicad: High load with EGL support enabled


baldrianbandit:
### Description
When moving certain symbols on the screen in Kicad, CPU load ramps up to 100% easily while framerate drops.

Using builds of wxwidgets and Kicad without EGL support, it runs fine.

#### Versions
wxwidgets-common-git 3.2.6.r45.gfd0f60daea-1
wxwidgets-gtk3-git 3.2.6.r45.gfd0f60daea-1
xorg-server 21.1.14-1
xf86-video-amdgpu 23.0.0-2 
6.11.9-arch1-1

#### To Reproduce:
See https://gitlab.com/kicad/code/kicad/-/issues/19149

vadz:
I'm really not sure what can/are we supposed to do here. If the only difference is using EGL vs using GLX, then it seems like EGL implementation (i.e. driver) is suboptimal, but there is nothing we can do about it at this level. If the problem is something that wxGTK itself does, it would be great to have a practical way to reproduce it, i.e. a patch to one of our OpenGL samples.

baldrianbandit:
Thanks for reaching out @vadz. May I ask for your opinion: I'm seeing this on all kind of machines with Intel / AMD consumer / professional grade GPUs running Arch Linux and Xorg. Do you think it might relate to either Xorg or drivers or the way Kicad uses Wxwidgets?

vadz:
Sorry, I can't really answer that. My first instinct would be to say that if the same code works correctly when using GLX, then the problem should be in the EGL driver. But, of course, I could be missing something here, and either KiCad or wx could conceivable also have some bug in their OpenGL code which somehow doesn't manifest itself with GLX. It's just really difficult to know without having a simple way of reproducing the problem.

dsa-t:
wx 3.2.6 disables vsync when using EGL, while with GLX the vsync is enabled.

We should be drawing no faster than OP's mouse poll rate (just 125 Hz), which still may or may not cause this high CPU usage, because KiCad might be doing more than one draw call per mouse move.

With GLX the draw rate is limited to 60 Hz.

dsa-t:
To prevent EGL+Wayland from breaking, maybe wx can introduce a swap buffers call that uses Wayland protocols to wait.

While EGL+Xorg should use the normal swap interval method.

dsa-t:
[...]

PR #25449, opened 2025-05-25
Add wxGLCanvasX11::GLXSetSwapInterval()

vadz:
Allow setting the swap interval value or not setting it at all instead of always forcing it to 0.

The changes of 3dde6bdeb0 (Disable swap interval in GLX wxGLCanvas implementation too, 2023-12-25) are not always desirable, so allow to customize the behaviour.

See #24163, #24165.

### DIFF
--- a/include/wx/glcanvas.h
+++ b/include/wx/glcanvas.h
@@ -212,7 +212,15 @@ class WXDLLIMPEXP_GL wxGLCanvasBase : public wxWindow
 {
 public:
     // Disable changing swap interval or indicate that it is unknown.
-    static constexpr int DefaultSwapInterval = -1;
+    static constexpr int DefaultSwapInterval = INT_MAX;
+
+    // Return values of SetSwapInterval().
+    enum class SwapInterval
+    {
+        NotSet = 0,
+        Set = 1,
+        NonAdaptive = 2
+    };
 
 
     // default ctor doesn't initialize the window, use Create() later
@@ -245,12 +253,18 @@ public:
 
     // Set swap interval to the specified value.
     //
-    // Special value of 0 means to disable vsync and DefaultSwapInterval means
-    // to disable automatically disabling vsync by default, as needs to be done
+    // Special value of 0 means to disable VSync and DefaultSwapInterval means
+    // to disable automatically disabling VSync by default, as needs to be done
     // under some platforms currently.
     //
+    // Negative values may be used to enable adaptive VSync if supported by the
+    // implementation.
+    //
     // Return true if the swap interval was set successfully, false if not.
-    virtual bool SetSwapInterval(int WXUNUSED(interval)) { return false; }
+    virtual SwapInterval SetSwapInterval(int WXUNUSED(interval))
+    {
+        return SwapInterval::NotSet;
+    }
 
 
     // accessors
--- a/include/wx/glcanvas.h
+++ b/include/wx/glcanvas.h
@@ -211,6 +211,10 @@ protected:
 class WXDLLIMPEXP_GL wxGLCanvasBase : public wxWindow
 {
 public:
+    // Disable changing swap interval or indicate that it is unknown.
+    static constexpr int DefaultSwapInterval = -1;
+
+
     // default ctor doesn't initialize the window, use Create() later
     wxGLCanvasBase();
 
@@ -239,6 +243,15 @@ public:
     // flush the back buffer (if we have it)
     virtual bool SwapBuffers() = 0;
 
+    // Set swap interval to the specified value.
+    //
+    // Special value of 0 means to disable vsync and DefaultSwapInterval means
+    // to disable automatically disabling vsync by default, as needs to be done
+    // under some platforms currently.
+    //
+    // Return true if the swap interval was set successfully, false if not.
+    virtual bool SetSwapInterval(int WXUNUSED(interval)) { return false; }
+
 
     // accessors
     // ---------
@@ -250,6 +263,10 @@ public:
 #if wxUSE_PALETTE
     const wxPalette *GetPalette() const { return &m_palette; }
 #endif // wxUSE_PALETTE
+
+    // Return the current swap interval.
+    virtual int GetSwapInterval() const { return DefaultSwapInterval; }
+
 
     // miscellaneous helper functions
     // ------------------------------
--- a/include/wx/msw/glcanvas.h
+++ b/include/wx/msw/glcanvas.h
@@ -90,7 +90,7 @@ public:
 
     // implement wxGLCanvasBase methods
     virtual bool SwapBuffers() override;
-    virtual bool SetSwapInterval(int interval) override;
+    virtual SwapInterval SetSwapInterval(int interval) override;
     virtual int GetSwapInterval() const override;
 
 
--- a/include/wx/msw/glcanvas.h
+++ b/include/wx/msw/glcanvas.h
@@ -90,6 +90,8 @@ public:
 
     // implement wxGLCanvasBase methods
     virtual bool SwapBuffers() override;
+    virtual bool SetSwapInterval(int interval) override;
+    virtual int GetSwapInterval() const override;
 
 
     // MSW-specific helpers
--- a/include/wx/osx/glcanvas.h
+++ b/include/wx/osx/glcanvas.h
@@ -103,7 +103,7 @@ public:
     virtual bool SwapBuffers() override;
 
 #ifdef __WXOSX_MAC__
-    virtual bool SetSwapInterval(int interval) override;
+    virtual SwapInterval SetSwapInterval(int interval) override;
     virtual int GetSwapInterval() const override;
 #endif
 
--- a/include/wx/osx/glcanvas.h
+++ b/include/wx/osx/glcanvas.h
@@ -102,6 +102,11 @@ public:
     // implement wxGLCanvasBase methods
     virtual bool SwapBuffers() override;
 
+#ifdef __WXOSX_MAC__
+    virtual bool SetSwapInterval(int interval) override;
+    virtual int GetSwapInterval() const override;
+#endif
+
     // Mac-specific functions
     // ----------------------
 
@@ -129,6 +134,10 @@ protected:
 
     WXGLPixelFormat m_glFormat = nullptr;
     wxGLAttributes m_GLAttrs;
+
+private:
+    // The value of swap interval to set or DefaultSwapInterval.
+    int m_swapIntervalToSet = DefaultSwapInterval;
 
     wxDECLARE_CLASS(wxGLCanvas);
 };
--- a/include/wx/unix/glcanvas.h
+++ b/include/wx/unix/glcanvas.h
@@ -53,6 +53,8 @@ public:
     virtual ~wxGLCanvasUnix();
 
     virtual bool SwapBuffers() override;
+    virtual bool SetSwapInterval(int interval) override;
+    virtual int GetSwapInterval() const override;
 
     virtual bool IsShownOnScreen() const override;
 
--- a/include/wx/unix/glcanvas.h
+++ b/include/wx/unix/glcanvas.h
@@ -53,7 +53,7 @@ public:
     virtual ~wxGLCanvasUnix();
 
     virtual bool SwapBuffers() override;
-    virtual bool SetSwapInterval(int interval) override;
+    virtual SwapInterval SetSwapInterval(int interval) override;
     virtual int GetSwapInterval() const override;
 
     virtual bool IsShownOnScreen() const override;
--- a/include/wx/unix/private/glcanvas.h
+++ b/include/wx/unix/private/glcanvas.h
@@ -46,14 +46,18 @@ public:
 
     virtual bool SwapBuffers() = 0;
 
-    bool SetSwapInterval(int interval)
+    wxGLCanvas::SwapInterval SetSwapInterval(int interval)
     {
         if ( !HasWindow() )
         {
             // We don't have a window yet, so we can't set the swap interval
             // right now. Remember the value to set it later in SwapBuffers().
             m_swapIntervalToSet = interval;
-            return true;
+
+            // We don't know whether we will be able to set it or not, so
+            // optimistically report success just because returning failure
+            // would be even worse.
+            return wxGLCanvas::SwapInterval::Set;
         }
 
         // We will try to set it now and we shouldn't try setting it again in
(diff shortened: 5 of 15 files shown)


vadz:
I've completely forgot about this one because it was marked as "Draft" and by now I don't remember why any more.

@dsa-t I think this was requested by you, would this still be useful?

In any case, we need to either merge or close it.

dsa-t:
I think `SetSwapInterval` needs to be implemented in each GL backend, not just `wxGLCanvasX11::GLXSetSwapInterval`.

And I don't think the interval should be changed inside `SwapBuffers`.

For Wayland, I think a solution similar to SDL can be implemented (see https://github.com/wxWidgets/wxWidgets/issues/24977#issuecomment-2981633368)

dsa-t:
Also I'm not sure if [wxGLSetSwapInterval](https://github.com/wxWidgets/wxWidgets/blob/dac1996ebe162386071c0b9148079fabf3a950e3/src/unix/glx11.cpp#L781) checks for the extension correctly.

`glXSwapIntervalEXT` function may exist in the client library, but `GLX_EXT_swap_control`/`GLX_EXT_swap_control_tear` extensions may not supported by the driver.

See how it's implemented in KiCad: https://gitlab.com/kicad/code/kicad/-/blob/c5a2b86379d3583a47e0e4ee5ef64dc898d68e75/include/gal/opengl/gl_utils.h#L49

vadz:
> I think `SetSwapInterval` needs to be implemented in each GL backend, not just `wxGLCanvasX11::GLXSetSwapInterval`.

I may have forgotten (or just misunderstood) things but I thought that we wanted to add it to allow customizing setting of the swap interval to 0 done right now in Unix (GLX and EGL) backends. Do we really need it elsewhere? It wouldn't be that difficult to call `wglSwapIntervalEXT()` under Windows and apparently there is `CGLSetParameter(kCGLCPSwapInterval)` for macOS, but would this be really useful there? Will KiCad use this?

> And I don't think the interval should be changed inside `SwapBuffers`.

Note that this is only done once. It's a way to allow setting the interval before it can be applied. Is it still a problem?
 
> For Wayland, I think a solution similar to SDL can be implemented (see [#24977 (comment)](https://github.com/wxWidgets/wxWidgets/issues/24977#issuecomment-2981633368))

This seems complicated and fragile :-( The original commit message speaks about "upcoming protocols that should improve the situation", would anybody know if anything has happened since then (2021)?

> `glXSwapIntervalEXT` function may exist in the client library, but `GLX_EXT_swap_control/GLX_EXT_swap_control_tear` extensions may not supported by the driver.

Thanks, this is definitely a good point, I've updated this PR with a better check. I've also added a check for MESA extension, although I'm unable to test this because all my systems support `EXT_swap_control` — but it looks like it shouldn't do any harm. I didn't add support for `GLX_SGI_swap_control`, even though it would be easy to do, of course, because it looks like this one shouldn't be relevant any longer.

Finally, I've rebased everything on master.

Next steps are:

1. Decide if we need this under Windows (easy to do, questionable utility).
2. And under macOS (slightly less clear how to do, same question about utility).
3. Implement at least `GLXDontSetSwapInterval()` for EGL backend (simple to do, might be useful?).
4. Think about whether we really need `GLXSetSwapInterval()` with arbitrary interval for EGL too and do something like what SDL does, but I'm still not sure if this is really a wise thing to do — and would definitely welcome any better ideas!

vadz:
This would need to be redone once #26023 is merged, but I am still not sure about (4) (and also (1) and (2)...) from above.

vadz:
Still not sure about this one and if we want to have this in 3.3.2, something needs to be done about it very soon.

@dsa-t Any thoughts by chance?

dsa-t:
> Do we really need it elsewhere? It wouldn't be that difficult to call `wglSwapIntervalEXT()` under Windows and apparently there is `CGLSetParameter(kCGLCPSwapInterval)` for macOS, but would this be really useful there? Will KiCad use this?

> 1\. Decide if we need this under Windows (easy to do, questionable utility).
> 2\. And under macOS (slightly less clear how to do, same question about utility).


We'll be happy to replace our `SetSwapInterval` with the wx function ;)



> Note that this is only done once. It's a way to allow setting the interval before it can be applied. Is it still a problem?

We probably should set the egl swap interval in wx's `SetSwapInterval` and `MakeCurrent`, like SDL, and implement `SwapBuffers` like SDL too.

> This seems complicated and fragile :-( The original commit message speaks about "upcoming protocols that should improve the situation", would anybody know if anything has happened since then (2021)?

Not sure about protocols, but there's been some changes in SDL, see current code:

https://github.com/libsdl-org/SDL/blob/339a91c5ba1a66b979afab0c31b3bca31cee1917/src/video/wayland/SDL_waylandopengles.c#L105


> 4\. Think about whether we really need `GLXSetSwapInterval()` with arbitrary interval for EGL too and do something like what SDL does, but I'm still not sure if this is really a wise thing to do — and would definitely welcome any better ideas!



For X11 + GLX + Wayland (Xwayland), the compatibility layer should be implemented in XWayland I think. 
Otherwise we have to disable vsync on Xwayland and get high CPU usage because it's rendering as fast as possible.

Though there appear to be X11 "present" extensions but i've not looked at them.

dsa-t:
I'm thinking it can be a flag for `SetSwapInterval` whether to use the workarounds.

Or another function to set the swap interval without the workarounds.

vadz:
Please check the latest version which just implements `SetSwapInterval()` for all the 3 major platforms. Compared to KiCad version, it doesn't support using negative values, i.e. doesn't use GLX/WGL_EXT_swap_control_tear extension, but this could be added if we believe this would be universally useful.

Also, I couldn't really test this in wxMSW as my VM uses generic OpenGL driver which doesn't support any extensions. If somebody could please check if the new menu items behave as expected in a more fully functional MSW installation, I'd be grateful.

I think we'll have to leave any smart SDL-like solutions avoiding disabling vsync under Wayland for later, I still don't fully understand what needs to be done and almost certainly won't have time to do it before 3.3.2.

dsa-t:
Thanks!

Adaptive vsync would be useful to have as this is what KiCad currently requests for balance between latency and tearing.

There's also an adaptive vsync implementation for macOS in SDL: https://github.com/libsdl-org/SDL/blob/3dc48a4890d37560e4159f50b7a6eadc9b9c1936/src/video/cocoa/SDL_cocoaopengl.m#L498

I didn't test on MSW yet.

> I think we'll have to leave any smart SDL-like solutions avoiding disabling vsync under Wayland for later, I still don't fully understand what needs to be done and almost certainly won't have time to do it before 3.3.2.

It's the only solution I see for Wayland currently, but it's fine to do later.

vadz:
I've added the possibility to request adaptive VSync now, please let me know if you have any comments.

I've hesitated for a long time about the API to use, but finally kept a simple `int interval` parameter for this function instead of adding `int flags` and passing some `Adaptive` flag to it because it looks like using -1/0/1 is standard practice in OpenGL programming and so it probably doesn't make sense to deviate from it.

I still changed the return type to be a bit more type-safe. The alternative here would be to just return `false` if adaptive VSync is not supported, but this would force all applications to do

```cpp
if ( !SetSwapInterval(-1) )
    SetSwapInterval(1);
```

as I don't think there are any cases where you'd want to use adaptive VSync but not want to enable standard VSync if it's not available, so it seems better to do it inside the library.

Finally, please note that I believe that the code in KiCad `GL_UTILS::SetSwapInterval()` is wrong as it queries `GLX_LATE_SWAPS_TEAR_EXT` before requesting adaptive VSync and AFAICS it's always going to be 0 in this case. But please let me know if I'm missing something, of course.

vadz:
P.S. Testing under MSW would still be much appreciated!

dsa-t:
> Finally, please note that I believe that the code in KiCad `GL_UTILS::SetSwapInterval()` is wrong as it queries `GLX_LATE_SWAPS_TEAR_EXT` before requesting adaptive VSync and AFAICS it's always going to be 0 in this case. But please let me know if I'm missing something, of course.

Looks like Mesa errors out if `EXT_swap_control_tear_bit` is not set at `glXSwapIntervalEXT` call: https://gitlab.freedesktop.org/mesa/mesa/-/blob/51d4803a6fbae1d280885ed6b70683efadd12e81/src/glx/glxcmds.c#L1674

I presume these bits are set internally by mesa and not requested by the user.

dsa-t:
Looking at SDL, the behaviour was(is?) also different between Mesa and NVIDIA: https://github.com/libsdl-org/SDL/blob/f07ba5b954518fb2ab7083158bbcae8e40ab5fab/src/video/x11/SDL_x11opengl.c#L1023-L1049

vadz:
I don't know anything about this bit, but doesn't it just indicate that the extension is present?

To be clear about why I wrote that KiCad code was wrong: according to everything I could find online and in my (admittedly, limited) testing `GLX_LATE_SWAPS_TEAR_EXT` is just never set until the swap interval is actually set to -1. And [Mesa code](https://gitlab.freedesktop.org/mesa/mesa/-/blob/51d4803a6fbae1d280885ed6b70683efadd12e81/src/glx/glx_pbuffer.c#L305) confirms this. So testing for it before setting the interval means that it's actually never set to -1 at all.

dsa-t:
On my Linux Mint (XFCE, X11) machine with `Mesa Intel(R) Graphics (RPL-S), 4.6 (Compatibility Profile) Mesa 23.2.1-1ubuntu3.1~22.04.3` the value received from `GLX_LATE_SWAPS_TEAR_EXT` is correctly 1

dsa-t:
Which GPUs and drivers were you testing on?

dsa-t:
Anyways I think you're right. 

Since the extension check was added, the drawable check is not needed.

vadz:
> Which GPUs and drivers were you testing on?

I was using `amdgpu` kernel module. I also have an integrated Intel graphics controller on this machine but, embarrassingly, I don't even know how to use it (`i915` module is shown as being "in use" but I don't think it's really used for anything).

But I've now retested on a notebook with integrated Intel and you're right, applying this:

```diff
diff --git a/src/unix/glx11.cpp b/src/unix/glx11.cpp
index c31198d8bd2..cc42a8c5f17 100644
--- a/src/unix/glx11.cpp
+++ b/src/unix/glx11.cpp
@@ -852,6 +852,10 @@ wxGLSetSwapInterval(Display* dpy, GLXDrawable drawable, int interval)
             result = wxGLCanvas::SwapInterval::NonAdaptive;
         }

+        unsigned int value = 0;
+        glXQueryDrawable(dpy, drawable, GLX_LATE_SWAPS_TEAR_EXT, &value);
+        wxLogTrace(TRACE_GLX, "GLX_LATE_SWAPS_TEAR_EXT=%u", value);
+
         wxLogTrace(TRACE_GLX, "Setting GLX swap interval to %d", interval);

         s_glXSwapIntervalEXT(dpy, drawable, interval);
```

outputs 1 even before calling `glXSwapIntervalEXT()` there. So KiCad is only broken for AMD (assuming Nvidia works like this too, but I can test it with it as well) and not always, but it's still not good.

> Since the extension check was added, the drawable check is not needed.

So does this look good enough to be used by KiCad to you? Notably are you (reasonably...) happy with the API?

dsa-t:
API looks good for me. In KiCad it would be used like this:

```
    wxGLCanvas::SetSwapInterval( -1 );
    m_swapInterval = wxGLCanvas::GetSwapInterval();
```

On Win11 (NVIDIA GPU), `wglGetProcAddress("wglGetExtensionsStringARB")` in `wxGLCanvasBase::IsExtensionSupported` returns null.
As well as other `wglGetProcAddress` calls in `wxGLContext::wxGLContext`. So the context is created via "legacy" `wglCreateContext`.

dsa-t:
Apparently the "GDI Generic", version "1.1.0" driver is used at `wxGLContext::wxGLContext`.
But when running KiCad (instead of wx samples), it recognizes the NVIDIA driver correctly.

vadz:
> API looks good for me. In KiCad it would be used like this:
> 
> ```
>     wxGLCanvas::SetSwapInterval( -1 );
>     m_swapInterval = wxGLCanvas::GetSwapInterval();
> ```

Just to understand, why do you need the actual swap interval back? How do you use it and will all (or at least most) applications need it to? If so, maybe it would make sense to return it from `SetSwapInterval()`, after all... I just thought it would be something most applications wouldn't really care about because what can you do if you've failed to set adaptive VSync, there doesn't seem to be any reasonable fallback other than using standard VSync, is there?

> On Win11 (NVIDIA GPU), `wglGetProcAddress("wglGetExtensionsStringARB")` in `wxGLCanvasBase::IsExtensionSupported` returns null. As well as other `wglGetProcAddress` calls in `wxGLContext::wxGLContext`. So the context is created via "legacy" `wglCreateContext`.

In my VM this happens because it's using the generic GDI driver (as also confirmed by running `wglinfo`, i.e. it's not something wx-specific), but I hoped it wouldn't be the case in a physical machine (or maybe better configured VMs, as I don't care about graphics much I've never really looked into this).

>  But when running KiCad (instead of wx samples), it recognizes the NVIDIA driver correctly.

So in your case `wglinfo` sees Nvidia but wx cube sample still shows "GDI Generic" in its "About"? This would be a bug in wx...

dsa-t:
> So in your case `wglinfo` sees Nvidia but wx cube sample still shows "GDI Generic" in its "About"?

Yes
image

dsa-t:
`pyramid` recognizes Intel GPU though:
image

dsa-t:
If I do this:
`wxGLAttributes attribs = wxGLAttributes().PlatformDefaults();`

Instead of this:
`wxGLAttributes attribs = wxGLAttributes().Defaults();`

then `cube` uses NVIDIA driver 
image

And swap interval functions work!

vadz:
I see, thanks. Unfortunately using `PlatformDefaults()` doesn't help in my VM, as it genuinely doesn't have any OpenGL drivers. It looks like I'm going to either have to install Windows natively somewhere or spend time on making GPU passthrough work with KVM if I want to really test this. I could also install Mesa under Windows but I don't know how useful this would be, nobody is probably really using it there.

So, to summarize: if you use `PlatformDefaults()` in cube, do the new menu items work for you, i.e. can you set swap interval to 1/0/-1 there successfully?

If so, and if you don't think it's worth changing the API to return the actually set swap interval (i.e. you need it for some KiCad-specific reasons and not because any code calling this function would need it), then I'd like to finally merge this.

dsa-t:
> Just to understand, why do you need the actual swap interval back? How do you use it and will all (or at least most) applications need it to?

Well it's unused in KiCad.
But one might want to limit the rendering speed differently depending on swap interval, e.g.:
- with value 0, limit to arbitrary frame rate, don't render as fast as CPU/GPU allows 
- with values -1 or 1+, render as fast as display allows

> So, to summarize: if you use `PlatformDefaults()` in cube, do the new menu items work for you, i.e. can you set swap interval to 1/0/-1 there successfully?

Yes.

> If so, and if you don't think it's worth changing the API to return the actually set swap interval

The API is fine for me either way.

PR #25453, opened 2025-05-27
Keep wxUSE_WEBVIEW_EDGE turned off in CMake Preset

Blake-Madden:
When I compile out-of-the-box with Visual Studio 2022 (Windows 11) using its CMake capabilities (I just open the folder from VS), I get this error:

> Cannot open include file: 'WebView2.h': No such file or directory

My process is:

- Clone from VS2022 (this will get all submodules)
- Open the wxWidgets folder in VS2022
- Select the "MSW Visual Studio 2022" target
- Build All or Install wxWidgets

The new CMake preset file is enabling `wxUSE_WEBVIEW_EDGE` by default, which I don't believe was the default before. Setting this to `OFF` fixes the error and I build fine after that.

Also, the docs imply that when using CMake, the WebView2 SDK will be somehow be installed for you and you don't need to do anything to manage it. Is it supposed to work that way, because that is not my experience (hence, the build error)? I do see code in "webview/CMake.txt" that should download the SDK, but it doesn't seem to be working for me or this CMake.txt isn't being included in the build process?

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -24,7 +24,7 @@
             "hidden": true,
             "inherits": "common",
             "cacheVariables": {
-                "wxUSE_WEBVIEW_EDGE": "ON"
+                "wxUSE_WEBVIEW_EDGE": "OFF"
             },
             "condition": {
                 "type": "equals",
--- a/interface/wx/webview.h
+++ b/interface/wx/webview.h
@@ -840,14 +840,13 @@ public:
     handler provided URLs.
 
     This backend is not enabled by default, to build it follow these steps:
-    - With CMake just enable @c wxUSE_WEBVIEW_EDGE
-    - When not using CMake:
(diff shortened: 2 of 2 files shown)


MaartenBent:
You could try deleting the `.vs` folder and output folder, get rid of all the CMake cache.
And add some `message("debug 1")`/`message("webview2 = ${WEBVIEW2_PACKAGE_DIR}")` lines to the webview CMakeLists.txt file, to see which check fails and why it doesn't reach the downloading part.

Blake-Madden:
I had the ".../libs/webview/packages/Microsoft.Web.WebView2.1.0.1722.45" folder, but it was empty. So that's why it wasn't downloading it. I must have been disconnected the first time I ran CMake, it made the folder but couldn't download anything since I was disconnected. Then after that point I had an empty folder and the script assumed I had what I needed and wouldn't try to re-download anything.

PR #25454, opened 2025-05-27
Fix tests build when using built-in CATCH

vadz:
This was broken in 01ca081d57 (Allow using system CATCH2 library in configure, 2023-06-15) where a wrong include path was used in this case.

See #23654.


vadz:
@lanurmi I'm going to merge this now as I'm reasonably sure this should fix the problem, but please let me know if it doesn't.

PR #25457, opened 2025-05-28
Check result of downloading WebView2 from CMake

Blake-Madden:
Immediately fail if the result of the SDK downloading wasn't `0` (success).

This will prevent making an empty SDK folder that will later confuse CMake into thinking the SDK was downloaded. Now you will keep getting the same error about failing to download the SDK when configuring CMake (at least until you actually download it).


vadz:
Thanks, this makes sense to me. @MaartenBent do you have any comments?

MaartenBent:
Looks good. Just a nitpick to use 4 spaces indentation instead of 2, like the rest of the CMake files.

PR #25458, opened 2025-05-28
Explain how to include an out-of-tree wxWidgets folder in CMake

Blake-Madden:
I download wxWidgets side-by-side with my projects. This way I can have all of them point to the same wxWidgets folder, rather than having multiple (submodule) wxWidgets in each project.

Anyway, these instructions explain the extra parameter needed to do this.

### DIFF
--- a/docs/doxygen/overviews/cmake.md
+++ b/docs/doxygen/overviews/cmake.md
@@ -171,7 +171,7 @@ target_link_libraries(myapp wx::net wx::core wx::base)
 
 This can be useful if you have multiple projects using wxWidgets.
 This way, you can place wxWidgets side-by-side with your other projects and
-have their Cmake scripts all point to the same wxWidgets folder.
+have their CMake scripts all point to the same wxWidgets folder.
 
 Using XRC
 ---------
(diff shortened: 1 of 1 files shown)


PBfordev (commented):
Just curious: Is sharing a library built with `add_subdirectory()` among projects a good practice?

What advantage does it have over `find_package()`, possibly with pointing the CMake to a specific folder (e.g., via setting `wxWidgets_ROOT`)?

Blake-Madden:
> Just curious: Is sharing a library built with `add_subdirectory()` among projects a good practice?

I agree, that would be bad practice. In my example, I am placing the build directory *inside* of the current project ("${CMAKE_CURRENT_SOURCE_DIR}/wxWidgets_lib", rather than "${CMAKE_CURRENT_SOURCE_DIR}/../wxWidgets_lib"). That way, every project would be pointing to the same wxWidgets source folder, but when they build they are copying the wxWidgets build into the project. So each project would have their own copy of the build files.

Blake-Madden:
> What advantage does it have over `find_package()`, possibly with pointing the CMake to a specific folder (e.g., via setting `wxWidgets_ROOT`)?

I used to do it that way (sing `wxWidgets_ROOT`), but this just seems so much easier. Just one line of code and you don't have to mess with setting a variable. Just my opinion though.

MaartenBent:
`add_subdirectory` supports a relative or absolute path to anywhere, it is not limited to a sub-folder of a project.

Blake-Madden:
I have to admit, it is misleadingly named function :)

But the official [CMake docs](https://cmake.org/cmake/help/latest/command/add_subdirectory.html) do say that absolute paths are supported.

PR #25461, opened 2025-05-28
Fix wxINFOBAR_CHECKBOX style getting lost under MSW

Blake-Madden:
Apply the style when calling `wxWindow::Create()`. Otherwise, `wxINFOBAR_CHECKBOX` will get lost when the window is actually created and trigger an assert from `ShowCheckBox()` (under MSW).


vadz (commented):
Thanks for the fix, but I'd like to check the answer to the question below before applying it.

PR #25462, opened 2025-05-28
CMake: Enable wxUSE_WEBVIEW_EDGE by default

MaartenBent:
After the discussion in #25453, I thought it would be a good idea to enable `wxUSE_WEBVIEW_EDGE` by default in every CMake build.

Probably warrants an entry in the changelog.


vadz:
Thanks, I guess it's indeed a good idea and it's probably best to do it before 3.3.0 to gather more feedback, so I'll merge it soon.

MaartenBent:
I added another CMake fix to this PR, for #25463

PR #25465, opened 2025-05-28
Fix wxGLCanvasEGL buffer scale desyncing from the GTK scale factor

### LINKED ISSUE #23733: EGL windows weirdness on high DPI screens with Wayland
While testing #23554 I've found that showing cube sample on high DPI screen works weirdly:

1. On startup, the cube itself is twice bigger than it should be.
2. After switching to another application and back things become even weirder: it becomes smaller, but squashed, jagged and the canvas becomes bigger than the window containing it (!) resulting in BeOS style window like this: 
Screenshot from 2023-07-24 18-05-26

The second problem might be related to the [GTK/Gnome bug](https://gitlab.gnome.org/GNOME/gtk/-/issues/5972), but the first one probably isn't.

Popax21:
Previously, `wxGLCanvasEGL` only set the buffer scale of the `wl_surface` once during the creation of the canvas. However, the GTK widget scale factor may change at any time (for example when the window is moved across monitors), which results in the EGL window size becoming out-of-sync with the surface's buffer scale when the canvas is resized. This patch fixes this issue by updating the `wl_surface` buffer size when the canvas widget is resized.

This fixes #23733, which was also reported externally in https://github.com/visualboyadvance-m/visualboyadvance-m/issues/1364, https://github.com/flathub/org.kicad.KiCad/pull/276#issuecomment-1949303117 and on the [KiCad GitLab](https://gitlab.com/kicad/code/kicad/-/issues/18079) (there probably also are other reports I missed).

### DIFF
--- a/include/wx/unix/glegl.h
+++ b/include/wx/unix/glegl.h
@@ -148,6 +148,7 @@ private:
     static EGLConfig ms_glEGLConfig;
 
     friend void wxEGLUpdatePosition(wxGLCanvasEGL* win);
+    friend void wxEGLSetScale(wxGLCanvasEGL* win, int scale);
 };
 
 // ----------------------------------------------------------------------------
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -430,6 +430,12 @@ void wxEGLUpdatePosition(wxGLCanvasEGL* win)
     wl_subsurface_set_position(win->m_wlSubsurface, x, y);
 }
 
+// Helper declared as friend in the header and so can access m_wlSurface.
+void wxEGLSetScale(wxGLCanvasEGL* win, int scale)
+{
+    wl_surface_set_buffer_scale(win->m_wlSurface, scale);
+}
+
 extern "C"
 {
 
@@ -490,6 +496,7 @@ static void gtk_glcanvas_size_callback(GtkWidget *widget,
                          win->m_height * scale, 0, 0);
 
     wxEGLUpdatePosition(win);
+    wxEGLSetScale(win, scale);
 }
 
 } // extern "C"


vadz:
Oh, thanks a lot for fixing this, this seems so obvious retrospectively, but I had completely failed to realize that we had to update the scale later too.

The only question I have is whether it's really a good idea to call `wl_surface_set_buffer_scale()` on every resize, i.e. whether it's cheap to call in the case when the scale didn't really change, or whether we should cache the last set scale and skip calling it if it didn't change? I probably should be able to find this out myself, but if you already know the answer, by chance, please let me know.

Thanks again!

Popax21:
> The only question I have is whether it's really a good idea to call `wl_surface_set_buffer_scale()` on every resize, i.e. whether it's cheap to call in the case when the scale didn't really change, or whether we should cache the last set scale and skip calling it if it didn't change?

I was actually investigating this a bit myself, but was unable to find a conclusive answer; my current best guess is that it doesn't - the `wl_surface:set_buffer_scale` Wayland API call only takes effect the next time `wl_surface:commit` is invoked, which in practice should be during the next `eglSwapBuffers` call. Additionally, from my understanding of the related API the compositor should only really need to keep track of the scale and not need to perform any additional work on scale changes. However, I can add a cached version of the scale to compare against if we want to play it safe.

vadz:
Thanks for looking at this!

I'll merge this soon.

Popax21:
Thanks for the quick turn-around! Do you by chance have a timeline for when this becomes part of the next release / is backported to 3.2.x? There's no time pressure (no worries!), I'm just asking so that I can potentially coordinate a bump to said version in downstream repositories (mainly nixpkgs since that is what I am using personally); still, it's totally fine if you can't make any promises regarding the timeline.

vadz:
3.3.0 should be out soon (on 6/6), we don't have any date for 3.2.9 yet, I had initially envisioned to make it in September but if any other important bugs are fixed in it soon, we might want to advance it.

PR #25466, opened 2025-05-29
Some fixes for building with CMake on MSYS2/MinGW-w64

MehdiChinoune:


MaartenBent:
As far as I can see the 'breaking' changes are:

* Always using `copy` instead of `create_symlink` on Windows. I'm fine with this, but maybe it is better to make this a build option. Something like `wxBUILD_INSTALL_USE_SYMLINK` with options `NO`, `YES` and  `DEFAULT` (no on Windows, yes on Linux/macOS).
* Add `$ENV{DESTDIR}` to symlink/copy source. This looks like a bug fix.

And other changes:
* Making `WIN32_MSVC_NAMING` an option, fine with me.
* Enabling `wxUSE_DBGHELP` for MinGW, fine with me.
* Changing `runtime_default_dir`. Seems logical to have this dependent on `WIN32_MSVC_NAMING`. AFAIK it doesn't change behaviour, except maybe for `CMAKE_CROSSCOMPILING`, which is probably fine. Note that you can also override this with `wxBUILD_INSTALL_RUNTIME_DIR`.

I think this needs more testing than I can do before 3.3.0, so might be best to not include it. And since the PR has been closed maybe it is not needed anymore...

Nonetheless these are some good suggestions so I'll try to test them and create a PR myself in the near future.

PR #25483, opened 2025-06-01
add wxAuiPaneInfo::FloatingClientSize()

wsu-cb:
Using wxWindow::SetClientSize() is more platform-independent than wxWindow::SetSize().  Provide that same functionality for wxAuiPaneInfo.

### DIFF
--- a/docs/doxygen/overviews/xrc_format.h
+++ b/docs/doxygen/overviews/xrc_format.h
@@ -733,7 +733,7 @@ wxAuiPaneInfo objects have the following properties:
     Sets the size of the floating pane.}
 @row3col{floating_client_size, @ref overview_xrcformat_type_size,
     Sets the client size of the floating pane. This has precedence over
-    floating_size.}
+    floating_size. This attribute is available since wxWidgets 3.3.1.}
 @row3col{min_size, @ref overview_xrcformat_type_size,
     Sets the minimum size of the pane.}
 @row3col{max_size, @ref overview_xrcformat_type_size,
--- a/docs/doxygen/overviews/xrc_format.h
+++ b/docs/doxygen/overviews/xrc_format.h
@@ -731,6 +731,9 @@ wxAuiPaneInfo objects have the following properties:
     Sets the ideal size for the pane.}
 @row3col{floating_size, @ref overview_xrcformat_type_size,
     Sets the size of the floating pane.}
+@row3col{floating_client_size, @ref overview_xrcformat_type_size,
+    Sets the client size of the floating pane. This has precedence over
+    floating_size.}
 @row3col{min_size, @ref overview_xrcformat_type_size,
     Sets the minimum size of the pane.}
 @row3col{max_size, @ref overview_xrcformat_type_size,
--- a/include/wx/aui/framemanager.h
+++ b/include/wx/aui/framemanager.h
@@ -152,6 +152,7 @@ public:
         , max_size(wxDefaultSize)
         , floating_pos(wxDefaultPosition)
         , floating_size(wxDefaultSize)
+        , floating_client_size(wxDefaultSize)
     {
         window = nullptr;
         frame = nullptr;
@@ -246,6 +247,8 @@ public:
     wxAuiPaneInfo& FloatingPosition(int x, int y) { floating_pos.x = x; floating_pos.y = y; return *this; }
     wxAuiPaneInfo& FloatingSize(const wxSize& size) { floating_size = size; return *this; }
     wxAuiPaneInfo& FloatingSize(int x, int y) { floating_size.Set(x,y); return *this; }
+    wxAuiPaneInfo& FloatingClientSize(const wxSize& size) { floating_client_size = size; return *this; }
+    wxAuiPaneInfo& FloatingClientSize(int x, int y) { floating_client_size.Set(x,y); return *this; }
     wxAuiPaneInfo& Fixed() { return SetFlag(optionResizable, false); }
     wxAuiPaneInfo& Resizable(bool resizable = true) { return SetFlag(optionResizable, resizable); }
     wxAuiPaneInfo& Dock() { return SetFlag(optionFloating, false); }
@@ -391,6 +394,8 @@ public:
 
     wxPoint floating_pos; // position while floating
     wxSize floating_size; // size while floating
+    // this has precedence over floating_size
+    wxSize floating_client_size; // client size while floating
     int dock_proportion;  // proportion while docked
 
     wxRect rect;              // current rectangle (populated by wxAUI)
--- a/include/wx/aui/framemanager.h
+++ b/include/wx/aui/framemanager.h
@@ -470,7 +470,7 @@ public:
     // Older functions using bespoke text format, prefer using the ones using
     // wxAuiSerializer and wxAuiDeserializer above instead in the new code.
     wxString SavePaneInfo(const wxAuiPaneInfo& pane);
-    void LoadPaneInfo(wxString panePart, wxAuiPaneInfo &pane);
+    bool LoadPaneInfo(wxString layoutVersion, wxString panePart, wxAuiPaneInfo &pane);
     wxString SavePerspective();
     bool LoadPerspective(const wxString& perspective, bool update = true);
 
--- a/include/wx/aui/framemanager.h
+++ b/include/wx/aui/framemanager.h
@@ -470,7 +470,10 @@ public:
     // Older functions using bespoke text format, prefer using the ones using
     // wxAuiSerializer and wxAuiDeserializer above instead in the new code.
     wxString SavePaneInfo(const wxAuiPaneInfo& pane);
-    bool LoadPaneInfo(wxString layoutVersion, wxString panePart, wxAuiPaneInfo &pane);
+    void LoadPaneInfo(wxString panePart, wxAuiPaneInfo &pane);
+private:
+    bool LoadPaneInfoVersioned(wxString layoutVersion, wxString panePart, wxAuiPaneInfo& pane);
+public:
     wxString SavePerspective();
     bool LoadPerspective(const wxString& perspective, bool update = true);
 
--- a/interface/wx/aui/framemanager.h
+++ b/interface/wx/aui/framemanager.h
@@ -782,7 +782,7 @@ public:
 
         @see wxWindow::SetClientSize
 
-        @since 3.3.0
+        @since 3.3.1
     */
     wxAuiPaneInfo& FloatingClientSize(const wxSize& size);
     wxAuiPaneInfo& FloatingClientSize(int x, int y);
--- a/interface/wx/aui/framemanager.h
+++ b/interface/wx/aui/framemanager.h
@@ -765,7 +765,9 @@ public:
     //@{
     /**
         FloatingSize() sets the size of the floating pane.
-        FloatingClientSize() has precedence over this.
+
+        FloatingClientSize() has precedence over this, i.e. this size is ignored
+        if the floating client size is specified.
     */
     wxAuiPaneInfo& FloatingSize(const wxSize& size);
     wxAuiPaneInfo& FloatingSize(int x, int y);
@@ -774,7 +776,9 @@ public:
     //@{
     /**
         FloatingClientSize() sets the client size of the floating pane.
-        This has precedence over FloatingSize().
+
+        This has precedence over FloatingSize(), i.e. FloatingSize() is ignored
+        if this is specified.
 
         @see wxWindow::SetClientSize
 
--- a/interface/wx/aui/framemanager.h
+++ b/interface/wx/aui/framemanager.h
@@ -429,7 +429,7 @@ public:
         @see SavePaneInfo()
         @see SavePerspective
     */
-    void LoadPaneInfo(wxString pane_part, wxAuiPaneInfo& pane);
+    void LoadPaneInfo(wxString layoutVersion, wxString pane_part, wxAuiPaneInfo& pane);
 
     /**
         Loads a saved perspective.
--- a/interface/wx/aui/framemanager.h
+++ b/interface/wx/aui/framemanager.h
@@ -765,9 +765,23 @@ public:
     //@{
     /**
         FloatingSize() sets the size of the floating pane.
+        FloatingClientSize() has precedence over this.
     */
     wxAuiPaneInfo& FloatingSize(const wxSize& size);
     wxAuiPaneInfo& FloatingSize(int x, int y);
+    //@}
+
+    //@{
+    /**
+        FloatingClientSize() sets the client size of the floating pane.
+        This has precedence over FloatingSize().
+
+        @see wxWindow::SetClientSize
+
+        @since 3.3.0
+    */
+    wxAuiPaneInfo& FloatingClientSize(const wxSize& size);
+    wxAuiPaneInfo& FloatingClientSize(int x, int y);
     //@}
 
     /**
--- a/interface/wx/aui/framemanager.h
+++ b/interface/wx/aui/framemanager.h
@@ -429,7 +429,7 @@ public:
         @see SavePaneInfo()
         @see SavePerspective
     */
-    void LoadPaneInfo(wxString layoutVersion, wxString pane_part, wxAuiPaneInfo& pane);
+    void LoadPaneInfo(wxString pane_part, wxAuiPaneInfo& pane);
 
     /**
         Loads a saved perspective.
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -1164,8 +1164,22 @@ wxString wxAuiManager::SavePaneInfo(const wxAuiPaneInfo& pane)
 }
 
 // Load a "pane" with the pane information settings in pane_part
-bool wxAuiManager::LoadPaneInfo(wxString layoutVersion, wxString pane_part, wxAuiPaneInfo &pane)
+void wxAuiManager::LoadPaneInfo(wxString pane_part, wxAuiPaneInfo &pane)
 {
+    if (LoadPaneInfoVersioned("layout3", pane_part, pane))
+    {
+        return;
+    }
+    // This is redundant since layout3
+    // recognizes all the layout2 keys,
+    // so should this be removed?
+    LoadPaneInfoVersioned("layout2", pane_part, pane);
+}
+
+bool wxAuiManager::LoadPaneInfoVersioned(wxString layoutVersion, wxString pane_part, wxAuiPaneInfo& destination)
+{
+    // don't overwrite destination unless pane_part is valid
+    wxAuiPaneInfo pane(destination);
     // replace escaped characters so we can
     // split up the string easily
     pane_part.Replace(wxT("\\|"), wxT("\a"));
@@ -1240,6 +1254,7 @@ bool wxAuiManager::LoadPaneInfo(wxString layoutVersion, wxString pane_part, wxAu
     pane_part.Replace(wxT("\a"), wxT("|"));
     pane_part.Replace(wxT("\b"), wxT(";"));
 
+    destination = pane;
     return true;
 }
 
@@ -1347,7 +1362,7 @@ bool wxAuiManager::LoadPerspective(const wxString& layout, bool update)
         pane_part.Replace(wxT("\a"), wxT("|"));
         pane_part.Replace(wxT("\b"), wxT(";"));
 
-        if (!LoadPaneInfo(layoutVersion, pane_part, pane))
+        if (!LoadPaneInfoVersioned(layoutVersion, pane_part, pane))
         {
             return false;
         }
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -1157,8 +1157,8 @@ wxString wxAuiManager::SavePaneInfo(const wxAuiPaneInfo& pane)
     result += wxString::Format(wxT("floaty=%d;"), pane.floating_pos.y);
     result += wxString::Format(wxT("floatw=%d;"), pane.floating_size.x);
     result += wxString::Format(wxT("floath=%d;"), pane.floating_size.y);
-    result += wxString::Format(wxT("cli_floatw=%d;"), pane.floating_client_size.x);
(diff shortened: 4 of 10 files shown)


src/aui/framemanager.cpp
```diff
@@ -1154,7 +1156,9 @@ wxString wxAuiManager::SavePaneInfo(const wxAuiPaneInfo& pane)
     result += wxString::Format(wxT("floatx=%d;"), pane.floating_pos.x);
     result += wxString::Format(wxT("floaty=%d;"), pane.floating_pos.y);
     result += wxString::Format(wxT("floatw=%d;"), pane.floating_size.x);
-    result += wxString::Format(wxT("floath=%d"), pane.floating_size.y);
+    result += wxString::Format(wxT("floath=%d;"), pane.floating_size.y);
+    result += wxString::Format(wxT("cli_floatw=%d;"), pane.floating_client_size.x);
+    result += wxString::Format(wxT("cli_floath=%d"), pane.floating_client_size.y);
```
vadz: I wouldn't bother extending the old serialization format, but if we do, we probably need to change the layout version, i.e. use `layout3` instead of `layout2`.

interface/wx/aui/framemanager.h
```diff
@@ -765,11 +765,25 @@ class wxAuiPaneInfo
     //@{
     /**
         FloatingSize() sets the size of the floating pane.
+        FloatingClientSize() has precedence over this.
```
vadz: I think we need a blank line here for the docs to appear correctly, or did Doxygen relax its rules about it? In any case, it doesn't hurt the readability and we could also improve the description a bit.

```suggestion

        FloatingClientSize() has precedence over this, i.e. this size is ignored
        if the floating client size is specified.
```

(also below)

src/aui/framemanager.cpp
```diff
@@ -2532,10 +2544,22 @@ void wxAuiManager::Update()
             {
                 // frame already exists, make sure its position
                 // and size reflect the information in wxAuiPaneInfo
-                if ((p.frame->GetPosition() != p.floating_pos) || (p.frame->GetSize() != p.floating_size))
+                // give floating_client_size precedence over floating_size
```
vadz, line 2563: But we still end up calling `SetSize()`? I thought the whole idea was to call `SetClientSize()`.

src/xrc/xh_aui.cpp
```diff
@@ -193,6 +193,7 @@ wxObject *wxAuiXmlHandler::DoCreateResource()
             if ( HasParam(wxS("floatable")) )       paneInfo.Floatable( GetBool(wxS("floatable")) );
 // Sizes
             if ( HasParam(wxS("floating_size")) )   paneInfo.FloatingSize( GetSize(wxS("floating_size")) );
+            if ( HasParam(wxS("floating_client_size")) )   paneInfo.FloatingClientSize( GetSize(wxS("floating_client_size")) );
```
vadz, line 196: Could you please also add this to `docs/doxygen/overviews/xrc_format.h` and `misc/schema/xrc_schema.rnc`, near the existing `floating_size`?

src/aui/framemanager.cpp
```diff
@@ -3684,8 +3709,15 @@ void wxAuiManager::OnFloatingPaneResized(wxWindow* wnd, const wxRect& rect)
...
     wxASSERT_MSG(pane.IsOk(), wxT("Pane window not found"));
+    // if frame isn't fully set up, don't stomp on pos/size info
+    if (!pane.frame)
+    {
+        return;
+    }
 
-    pane.FloatingSize(rect.GetWidth(), rect.GetHeight());
+    /* client size has precedence over wnd size,
+        so I don't think we need to set wnd size */
```
vadz: Minor, but let's make this comment C++ and more authoritative:

```suggestion
    // Setting floating client size is enough, there is no need to set floating
    // size, as it won't be used if the client size is set.
```

vadz (commented):
Thanks for your work on this!

I think it's mostly good but please see the comments below and please check if using "layout2" still works — I think it doesn't, i.e. existing applications would give an error when restoring the new data. And the simplest way to fix it would be to just not update this code at all, we have `wxAuiSerializer` which is more flexible.

wsu-cb:
> I think it's mostly good but please see the comments below and please check if using "layout2" still works — I think it doesn't, i.e. existing applications would give an error when restoring the new data. And the simplest way to fix it would be to just not update this code at all, we have `wxAuiSerializer` which is more flexible.

What do you mean by "existing applications" in this context?  The addition of the `floating_client_size` member already makes the new wxWidgets library ABI-incompatible, so existing applications compiled against old libraries can't possibly work with the new wxWidgets code no matter what we do with `layout2`.

If you mean an old application that reads an external data source containing a layout string generated by the new code, then I'm almost certain you are correct that the old application would wxFAIL(), so I agree that we should change to `layout3` to instead result in `return false` from `LoadPerspective()`.

However, I don't want to abandon that old format.  My demonstration of how the new `FloatingClientSize()` works, and that it works as intended, is to use the `auidemo` `Perspectives | All Panes` menu item, and that uses the `SavePerspective()`/`LoadPerspective()` code.

interface/wx/aui/framemanager.h
wsu-cb: I have never used Doxygen, and have no reason to think its syntax has changed, so I will apply your suggestions here and elsewhere.

src/aui/framemanager.cpp
wsu-cb: The code uses `ClientToWindowSize()` to convert `floating_client_size` to a window size, and then calls `SetSize()`.  The reason to use `SetSize()` is that it can set both position and size, but I don't see a way for `SetClientSize()` to set the window position.  (My reading of the `SetClientSize(const wxRect&)` implementation is that it ignores the position!)  Are you suggesting rewriting the code to use separate `SetPosition()` and `SetClientSize()` calls?

interface/wx/aui/framemanager.h
wsu-cb: Done.  See 8c9c2d608514df13b13a486553d100f492762d34.

src/xrc/xh_aui.cpp
wsu-cb: Done.  See 8c9c2d608514df13b13a486553d100f492762d34.

src/aui/framemanager.cpp
wsu-cb: Done.  See 8c9c2d608514df13b13a486553d100f492762d34,

src/aui/framemanager.cpp
vadz: I think using `SetPosition()` and `SetClientSize()` might be a safer bet, because while I _hope_ `ClientToWindowSize()` works, I'm pretty sure that `SetClientSize()` does.

vadz:
> If you mean an old application that reads an external data source containing a layout string generated by the new code, then I'm almost certain you are correct that the old application would wxFAIL(), so I agree that we should change to `layout3` to instead result in `return false` from `LoadPerspective()`.

Yes, this is what I meant, but while I agree with the change to "layout3", I think we still need to keep support for reading "layout2" in the code: otherwise upgrading an application to newer wx version will lose the saved layout information, which is definitely not what we want. AFAICS, this should be simple to do (we just need to accept either on input).

docs/doxygen/overviews/xrc_format.h
```diff
@@ -731,6 +731,9 @@ wxAuiPaneInfo objects have the following properties:
     Sets the ideal size for the pane.}
 @row3col{floating_size, @ref overview_xrcformat_type_size,
     Sets the size of the floating pane.}
+@row3col{floating_client_size, @ref overview_xrcformat_type_size,
+    Sets the client size of the floating pane. This has precedence over
+    floating_size.}
```
vadz: Sorry, a minor addition:
```suggestion
    floating_size. This attribute is available since wxWidgets 3.3.1.}
```

wsu-cb:
> > If you mean an old application that reads an external data source containing a layout string generated by the new code, then I'm almost certain you are correct that the old application would wxFAIL(), so I agree that we should change to `layout3` to instead result in `return false` from `LoadPerspective()`.
> 
> Yes, this is what I meant, but while I agree with the change to "layout3", I think we still need to keep support for reading "layout2" in the code: otherwise upgrading an application to newer wx version will lose the saved layout information, which is definitely not what we want. AFAICS, this should be simple to do (we just need to accept either on input).

1)  Should the parser `wxFAIL()` if a `layout2` string contains `floating_client_size` data?  That would require passing the layout version to `LoadPaneInfo()`.  Or can we allow `LoadPaneInfo()` to accept any field name it recognizes independent of layout version?

2)  This discussion of versioning has made me realize that old (external) `wxAuiSerializer::SavePane()` implementations will silently ignore `wxAuiPaneLayoutInfo::floating_client_size`, and that data will be lost.  I don't know how to avoid that.

src/aui/framemanager.cpp
wsu-cb: Done.  See 15c3a445fbeca1513ab0fe7851ffba746fbcb309

docs/doxygen/overviews/xrc_format.h
wsu-cb: Done.  See bb1a1942593def39c54c518d939c99ae04227b37

wsu-cb:
> > > If you mean an old application that reads an external data source containing a layout string generated by the new code, then I'm almost certain you are correct that the old application would wxFAIL(), so I agree that we should change to `layout3` to instead result in `return false` from `LoadPerspective()`.
> > 
> > 
> > Yes, this is what I meant, but while I agree with the change to "layout3", I think we still need to keep support for reading "layout2" in the code: otherwise upgrading an application to newer wx version will lose the saved layout information, which is definitely not what we want. AFAICS, this should be simple to do (we just need to accept either on input).
> 
> 1. Should the parser `wxFAIL()` if a `layout2` string contains `floating_client_size` data?  That would require passing the layout version to `LoadPaneInfo()`.  Or can we allow `LoadPaneInfo()` to accept any field name it recognizes independent of layout version?
> 2. This discussion of versioning has made me realize that old (external) `wxAuiSerializer::SavePane()` implementations will silently ignore `wxAuiPaneLayoutInfo::floating_client_size`, and that data will be lost.  I don't know how to avoid that.

@vadz Did you overlook this?  Do you have any guidance on these issues?

src/aui/framemanager.cpp
```diff
@@ -3684,8 +3709,15 @@ void wxAuiManager::OnFloatingPaneResized(wxWindow* wnd, const wxRect& rect)
     // try to find the pane
     wxAuiPaneInfo& pane = GetPane(wnd);
     wxASSERT_MSG(pane.IsOk(), wxT("Pane window not found"));
+    // if frame isn't fully set up, don't stomp on pos/size info
```
vadz, line 3727: Why not, actually? I don't really know how can `frame` be null here, but shouldn't we still remember the restored values to be able to use them when it is finally created?

vadz:
Sorry, I just didn't have time to return to this, but I just had another quick look and unfortunately some things are still not obvious to me.

FWIW I don't think the code should assert/fail when given any external data.

src/aui/framemanager.cpp
```diff
@@ -1273,11 +1281,12 @@ bool wxAuiManager::LoadPerspective(const wxString& layout, bool update)
     // check layout string version
     //    'layout1' = wxAUI 0.9.0 - wxAUI 0.9.2
     //    'layout2' = wxAUI 0.9.2 (wxWidgets 2.8)
+    //    'layout3' = wxWidgets 3.3
     part = input.BeforeFirst(wxT('|'));
     input = input.AfterFirst(wxT('|'));
     part.Trim(true);
     part.Trim(false);
-    if (part != wxT("layout2"))
+    if (part != wxT("layout3"))
```
vadz: Ideal would be to recognize both "layout2" (for importing existing saved data) and "layout3" (and only try to read new fields when it is used).

wsu-cb:
> FWIW I don't think the code should assert/fail when given any external data.

Before this PR, `wxAuiManager::LoadPaneInfo()` returns `void`, but `wxFAIL`s when the string starts with a valid `layout2` prefix, but  nevertheless contains bad data.  Would you like me to rewrite this code so that `wxAuiManager::LoadPaneInfo()` returns `bool` and does not wxFAIL?

FWIW, my own view is that it is quite reasonable to `wxFAIL()` when the passed string is invalid because that string might well have been generated by incorrect wxWidgets code.

vadz:
> > FWIW I don't think the code should assert/fail when given any external data.
> 
> Before this PR, `wxAuiManager::LoadPaneInfo()` returns `void`, but `wxFAIL`s when the string starts with a valid `layout2` prefix, but nevertheless contains bad data. Would you like me to rewrite this code so that `wxAuiManager::LoadPaneInfo()` returns `bool` and does not wxFAIL?

I think this would be better, yes, but I don't necessarily want to ask you to fix problems in the existing code. I'd like to avoid introducing, or reproducing, them in the new code, however.

> FWIW, my own view is that it is quite reasonable to `wxFAIL()` when the passed string is invalid because that string might well have been generated by incorrect wxWidgets code.

My position is that an assert should only be used for programming error, not for any bad data coming from the external sources. Sure, it's possible that this data was generated incorrectly. But it's also possible that it was corrupted in some other way and we must not assert in this case. And because we can't distinguish the two cases, we shouldn't assert in any case.

src/aui/framemanager.cpp
wsu-cb: > Why not, actually? I don't really know how can `frame` be null here, but shouldn't we still remember the restored values to be able to use them when it is finally created?

`p.frame` is null here during the floating pane setup when a docked pane is dragged to become floating because `wxAuiManager::Update()` calls `frame->SetPaneWindow()` before setting `p.frame = frame`.  `SetPaneWindow()` sets the window size, so it reaches `OnFloatingPaneResized()` when `p.frame` is null.  I tested setting `p.frame = frame` before `frame->SetPaneWindow()`, but that causes problems because then the `if (!p.IsFloating() && p.frame)` test thinks we are UN-floating a floating pane, and destroys windows.

Also, even if those problems were solved, when the pane has a `MinSize()`, `SetPaneWindow()` does some size operations before applying the pane's floating size, and that overwrites the floating size the pane had at the beginning of `SetPaneWindow()`, so when it does actually apply the pane's floating size, it is has been stomped on.

src/aui/framemanager.cpp
wsu-cb: Done.  See https://github.com/wxWidgets/wxWidgets/pull/25483/commits/ebd7585b7ceabf29f1471e38df56ddc490a2e717

src/aui/framemanager.cpp
```diff
@@ -1154,13 +1156,15 @@ wxString wxAuiManager::SavePaneInfo(const wxAuiPaneInfo& pane)
     result += wxString::Format(wxT("floatx=%d;"), pane.floating_pos.x);
     result += wxString::Format(wxT("floaty=%d;"), pane.floating_pos.y);
     result += wxString::Format(wxT("floatw=%d;"), pane.floating_size.x);
-    result += wxString::Format(wxT("floath=%d"), pane.floating_size.y);
+    result += wxString::Format(wxT("floath=%d;"), pane.floating_size.y);
+    result += wxString::Format(wxT("cli_floatw=%d;"), pane.floating_client_size.x);
+    result += wxString::Format(wxT("cli_floath=%d"), pane.floating_client_size.y);
```
vadz: This is pretty minor, but why change the order of words in "floating client size" here? I'd use something like

```suggestion
    result += wxString::Format(wxT("floatw_cli=%d;"), pane.floating_client_size.x);
    result += wxString::Format(wxT("floath_cli=%d"), pane.floating_client_size.y);
```

for these keys.

include/wx/aui/framemanager.h
```diff
@@ -465,7 +470,7 @@ class WXDLLIMPEXP_AUI wxAuiManager : public wxEvtHandler
     // Older functions using bespoke text format, prefer using the ones using
     // wxAuiSerializer and wxAuiDeserializer above instead in the new code.
     wxString SavePaneInfo(const wxAuiPaneInfo& pane);
-    void LoadPaneInfo(wxString panePart, wxAuiPaneInfo &pane);
+    bool LoadPaneInfo(wxString layoutVersion, wxString panePart, wxAuiPaneInfo &pane);
```
vadz: Sorry, but I don't think we should change this function, which is part of the public API, in an incompatible way — especially when it should be simple to keep it by doing the following:

1. Add a new function, e.g. `LoadVersionedPaneInfo(const wxString& version, const wxString& panePart, wxAuiPaneInfo& pane).
2. Call it from `LoadPaneInfo()` passing `"layout2"` as its first argument.

vadz (commented):
This looks almost good, but let's not make backwards-incompatible changes unnecessary. TIA!

include/wx/aui/framemanager.h
wsu-cb: > Sorry, but I don't think we should change this function, which is part of the public API, in an incompatible way — especially when it should be simple to keep it by doing the following:
> 
> 1. Add a new function, e.g. `LoadVersionedPaneInfo(const wxString& version, const wxString& panePart, wxAuiPaneInfo& pane).
> 2. Call it from `LoadPaneInfo()` passing `"layout2"` as its first argument.

But this PR modifies `SavePaneInfo()` to generate `layout3` data that 2.'s implementation of `LoadPaneInfo()` can't read.  That doesn't seem like a sensible API to me.  Are you implying that there should also be `SaveVersionedPaneInfo()`?

vadz: Hmm, good point. I'm not sure how/if `LoadPaneInfo()` is used outside of wx (I wouldn't have made it public in the first place), but I think it may be used if the application stores pane information in some custom storage and so it should be able to load the data in v2 format currently stored there. But it should be also able to read the data in v3 format which will be written by `SavePaneInfo()`... 

So we need to have a function which can load either v3 (this could be called internally when we know that we're using v3) or accept v2-or-v3, which `LoadPaneInfo()` would need to do.

Sorry about all the complications, this is why I had initially said that it would be better to avoid modifying this code at all. But if you want to do it, we really need to avoid breaking the existing code ("first, do no harm").

src/aui/framemanager.cpp
wsu-cb: Done.  See https://github.com/wxWidgets/wxWidgets/pull/25483/commits/09466113489fef6757ff32a3ce1211f6f97514aa

include/wx/aui/framemanager.h
wsu-cb: What do you think of https://github.com/wxWidgets/wxWidgets/pull/25483/commits/e775f314fcb71a0d44a04ffee0c50b09636aee74?

src/aui/framemanager.cpp
```diff
@@ -1268,16 +1291,18 @@ wxString wxAuiManager::SavePerspective()
...
-    part = input.BeforeFirst(wxT('|'));
+    //    'layout3' = wxWidgets 3.3.1
+    layoutVersion = input.BeforeFirst(wxT('|'));
     input = input.AfterFirst(wxT('|'));
-    part.Trim(true);
-    part.Trim(false);
-    if (part != wxT("layout2"))
+    layoutVersion.Trim(true);
+    layoutVersion.Trim(false);
+    if (layoutVersion != wxT("layout3") &&
```
vadz: This is a typo, right? Should be

```suggestion
    if (layoutVersion != wxT("layout2") &&
```

src/aui/framemanager.cpp
vadz: Do I understand correctly that this change (i.e. just this conditional return) is completely independent of the other ones and fixes a separate bug which (again, IIUC) means that currently we lose the pane size when floating it?

src/aui/framemanager.cpp
```diff
@@ -1154,14 +1156,30 @@ wxString wxAuiManager::SavePaneInfo(const wxAuiPaneInfo& pane)
...
 void wxAuiManager::LoadPaneInfo(wxString pane_part, wxAuiPaneInfo &pane)
 {
+    if (LoadPaneInfoVersioned("layout3", pane_part, pane))
+    {
+        return;
+    }
+    // This is redundant since layout3
+    // recognizes all the layout2 keys,
+    // so should this be removed?
+    LoadPaneInfoVersioned("layout2", pane_part, pane);
```
vadz: Yes, I think it should be removed, but a comment added above saying that we use "layout3" here because it is capable of loading both "layout2" and "layout3" data and we need to do it in this function for compatibility.

vadz (commented):
Thanks, I _think_ it should be good now.

Looking at the latest diff, it's clear that we could make our life much simpler by just always loading pane info as "layout3" (as it's a strict superset of "layout2"), this would only lose us detecting errors due to extra "cli" fields in "layout2" strings, which is probably not a big deal... But I don't know if you want to change this further.

src/aui/framemanager.cpp
wsu-cb: No, this is a new problem.  The original code was using window size, not client size, and the `wxAuiFloatingFrame` ctor can take window size as a parameter, and that allows everything to work.  However, there is no good way to get from client size to window size before the window exists, and client size can't be passed to the ctor.  This means that setting the client size doesn't get done until later, and if `min_size` is set, the `min_size` operations trigger `OnFloatingPaneResized()` before the client size gets applied to `wxAuiFloatingFrame`.  This conditional return is the least-invasive change I could find that avoided problems.

src/aui/framemanager.cpp
wsu-cb: You are correct.  See https://github.com/wxWidgets/wxWidgets/pull/25483/commits/aab24ad28074e180d6e5a5ffcafc90aa57989179

src/aui/framemanager.cpp
wsu-cb: Done.  See https://github.com/wxWidgets/wxWidgets/pull/25483/commits/8a9666479e1267f46b4d6edcd2d3788380c5c24e

vadz:
Great, thanks! This looks finally good to go, so I'm merging and pushing it.

Issue #24106, opened 2023-11-30
wxSlider thumb is clipped and value label misaligned on wxGTK


dsa-t:
The wxSlider thumb is clipped on wxGTK. Also the value label is not vertically centered on vertical sliders. `widgets` demo:

Top side clipped:
![image](https://github.com/wxWidgets/wxWidgets/assets/37658952/0b923a1e-fe1c-430d-994f-bf02501fffd2)

Top and left side clipped:

![image](https://github.com/wxWidgets/wxWidgets/assets/37658952/970c38f6-e9f9-4b9b-a191-48726e8248a5)

### Platform and version information

- wxWidgets version you use: 3.2 tip (https://github.com/wxWidgets/wxWidgets/commit/e8d51afb7871ceead40f1f8aa58e904c241828d9)
- wxWidgets port you use: wxGTK
- OS  and its version:
  
  + GTK version: 3.24.33-1ubuntu2
  + Which GDK  backend is used: X11
  + Desktop environment : Xfce 4.16.0
  + Current theme: Mint-Y-Aqua

vadz:
This should be closed by the commits above (thanks!), should we backport it to 3.2 too?

paulcor:
The clipping is a problem with the theme, it's trying to draw outside the widget's allocation. We can't fix that. The alignment is fixed in ea2beaaf5b. My thinking is that it probably should not be backported.

dsa-t:
> The clipping is a problem with the theme

I see the slider slightly clipped on the left even on Adwaita

![Image](https://github.com/user-attachments/assets/d9f35034-2306-4f25-abc3-994bc8d20949)

PR #25485, opened 2025-06-01
Place wxGTK wxSlider min/max labels the same way wxMSW does

paulcor:
The labels are in line with the slider, not adjacent to the ends.
See #24106

### DIFF
--- a/src/gtk/slider.cpp
+++ b/src/gtk/slider.cpp
@@ -312,32 +312,24 @@ bool wxSlider::Create(wxWindow *parent,
         return false;
     }
 
-    const bool isVertical = (style & wxSL_VERTICAL) != 0;
+    const bool isVertical = (style & (wxSL_LEFT | wxSL_RIGHT | wxSL_VERTICAL)) != 0;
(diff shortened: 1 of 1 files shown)


src/gtk/slider.cpp
```diff
@@ -312,32 +312,24 @@ bool wxSlider::Create(wxWindow *parent,
         return false;
     }
 
-    const bool isVertical = (style & wxSL_VERTICAL) != 0;
+    const bool isVertical = (style & (wxSL_LEFT | wxSL_RIGHT | wxSL_VERTICAL)) != 0;
```
vadz, line 315: This should be ideally renamed to something else (`hasHorzLabels`?) as the name is misleading now and it's not totally clear what/why are we testing here.

vadz (approved):
Thanks for fixing this!

Will you push it or should I merge it?

src/gtk/slider.cpp
paulcor: This is how wxMSW determines the orientation. `wxSL_LEFT/RIGHT` imply `wxSL_VERTICAL`.

Issue #24668, opened 2024-07-02
An API to define app_id for wayland


xeniarose:
### Description

See <https://gitlab.com/kicad/code/kicad/-/issues/18322> for some context.

On wayland, there's [an association called `app_id`](https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id) between application windows and their desktop files that's important for the window system to be able to associate opened window back to the desktop file. On KDE plasma 6, as i've found out with kicad, what happens when the association is broken is that KDE doesn't know what the correct icons for the windows are, or when the app is done opening its first window, and this results in some mildly broken/inconsistent UI. Essentially, `app_id` is similar to WM_CLASS in X11, but the two aren't compatible (for some reason).

In GTK+ 3, the field that defines the `app_id` for wayland purposes is `argv[0]` passed to [gtk_init_check](https://docs.gtk.org/gtk3/func.init_check.html). Alternatively, it can be defined in `GtkApplication`, which doesn't appear to be used in wxGTK. Thus for the kicad issue, the hotfix i've applied to my own versions is just rewriting `argv[0]` to match the desktop file name (without the `.desktop` part), and this fully makes the broken UI go away. There's similar facilities in Qt (but i'm not sure what the current state of wxQt is).

I would like wxWidgets to have an api for the `app_id` concept, which would map to the above low-level APIs on wxGTK and perhaps wxQt, and would be a no-op on mac and windows (unless there's some useful mapping there -- i'm not familiar with those APIs). I think some possible designs for this could be

1. extend wxIMPLEMENT_APP with a second field, eg
```cpp
wxIMPLEMENT_APP(MY_APP_CLASS, "org.me.myapp")
```
2. define an optional or non-optional macro 
```cpp
wxDECLARE_APP_ID("org.me.myapp")
// might expand in user code to eg
const char* wxGetAppId() {
    return "org.me.myapp";
}
// this could then be called by wxEntry/wxInit to retrieve and apply the user app_id
```
3. some other option. i'm happy to brainstorm solutions and draft any needed PRs for this issue

Another point that was raised on the KiCad issue was the possibility of wxWidgets implementing the [recently merged xdg-toplevel-icon](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/269) for wayland wxGTK. I can open a separate issue for that and draft PRs if there's interest.

vadz:
[...]

PR #25486, opened 2025-06-01
Allow setting Wayland app_id via wxApp:SetClassName()

paulcor:
See #24668

### DIFF
--- a/src/gtk/toplevel.cpp
+++ b/src/gtk/toplevel.cpp
@@ -429,6 +429,15 @@ void wxTopLevelWindowGTK::GTKHandleRealized()
     }
 #endif // 3.10
 
+#if defined(GDK_WINDOWING_WAYLAND) && GTK_CHECK_VERSION(3,24,22)
+    if (wxGTKImpl::IsWayland(window) && gtk_check_version(3,24,22) == nullptr)
+    {
+        const wxString className(wxApp::GetInstance()->GetClassName());
+        if (!className.empty())
+            gdk_wayland_window_set_application_id(window, className.utf8_str());
+    }
+#endif
+
     gdk_window_set_decorations(window, (GdkWMDecoration)m_gdkDecor);
     gdk_window_set_functions(window, (GdkWMFunction)m_gdkFunc);
 


imciner2:
~So, Windows calls this thing a `AppUserModelID` (set using `::SetCurrentProcessExplicitAppUserModelID`), so IMO it should use a more portable name than `Class`, because that is very X11 specific, actually.~

I see you are doing this slightly differently from what I expected. Is this way able to be controlled by the app after it is created, or does it require having it set properly at the very start?

paulcor:
I prefer `{Set,Get}ClassName()` as it can work for 3.2 as well.

`wxApp::SetClassName()` would have to be called before the TLW is shown.

vadz:
Thanks, but I'd rather postpone this until 3.3.1 in order to have time to implement it for wxMSW too if you don't mind.

This would also allow to check if wxMSW version can be backported too.

Also, just as a reminder to self, this would need to be documented in any case.

vadz:
So I've returned to this and this doesn't actually work for me even with Wayland, at least when using Sway: the current patch does nothing because neither `display_server.xdg_toplevel` nor `display_server.zxdg_toplevel_v6` are set in `window->impl` when `gdk_wayland_window_set_application_id()` is called, so the app ID is not set.

If I wrap this code in `CallAfter()`, or move it to `map-event` callback the app ID does gets set. But the window becomes invisible, for some reason. Nothing in [xdg_toplevel_set_app_id()](https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id) documentation says that it hides the window...

FWIW it does work with mutter, but I think it implements `xdg_shell_unstable_v6` (how do I check this easily? `wayland-info` doesn't output it, but then it doesn't output `xdg_shell` either, when I'm pretty sure it's supported) and so uses a different code path.

vadz:
OK, please disregard what I wrote, I had a rule with filter `[app_id="M"]` in my Sway config and this accidentally applied to `"MinimalTestID"` string I was testing with. Using a different string (or fixing the rule to use `^M$`) works fine. This still needs to be done in map callback, however, and not in the realized one, so I'll create a new PR based on this commit doing it there.

vadz:
This is part of #25548 now, so closing this one.

PR #25487, opened 2025-06-01
Enable -Wcast-qual in the CI builds using gcc or clang

vadz:
This is a useful warning not enabled by -Wextra, so enable it separately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -274,7 +274,7 @@ jobs:
 
           if [ -z ${{ matrix.allow_warnings }} ]; then
             if [ -z ${{ matrix.allow_extra_warnings }} ]; then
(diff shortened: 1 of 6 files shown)


vadz:
So there are a couple of warnings in 3rd party code. The one in nanosvg is easy to work around/suppress, but I don't know what to do about the ones in Lexilla/Scintilla. It's probably not really worth modifying the submodules just to fix them, is it? But there doesn't seem to be any simple way to use different warning flags for these files and our ones at the build system level... Maarten, would you have any advice by chance? TIA!

vadz:
Wait, I know, I think I can address both this problem and Paul's concerns about being able to enable the CI-like warnings locally: if we enable this (and any other warnings not enabled by default that we want to enable) from our own `wx/defs.h`, it should work, right? Let me try...

PR #25491, opened 2025-06-02
wxGTK on macOS: a follow-up fix to include volumes.mm in the build

barracuda156:
@vadz When you have time, please take a look at this. Eventually it should fix https://github.com/wxWidgets/wxWidgets/issues/25490
Not mergeable right now, therefore a draft.

1. `file`-related commit is a placeholder, I do not know how to add a category here correctly. Locally I just put the file into common base & GUI sources, which is non-portable, of course. What we need is for `volumes.mm` to be in base & GUI for all macOS, not only for Cocoa. I did not find an existing category fitting this requirement.
2. I do not know ObjC++ syntax, and relied on Copilot rewriting for it. Please check if it does any non-sense :) This compiles fine with gcc, and it allowed a dependent app to build and be usable. However I am not sure it is correct.

### DIFF
--- a/build/files
+++ b/build/files
@@ -176,6 +176,8 @@ BASE_OSX_SHARED_HDR =
 # Base and GUI files used by OS X
 BASE_AND_GUI_OSX_COCOA_SRC =
     src/osx/cocoa/utils.mm
+
+BASE_AND_GUI_DARWIN_SRC =
     src/osx/volume.mm
 
 # files used by non-wxMac OS X builds
--- a/src/osx/volume.mm
+++ b/src/osx/volume.mm
@@ -22,6 +22,11 @@
 
 #include "wx/volume.h"
 
+// gcc does not yet support modern Apple ObjC++ syntax.
+#ifndef __clang__
+#define USE_LEGACY_OBJC
+#endif
+
 #ifndef WX_PRECOMP
     #if wxUSE_GUI
         #include "wx/icon.h"
@@ -35,16 +40,33 @@
 #import <Foundation/NSFileManager.h>
 #import <Foundation/NSURL.h>
 
+// Fallbacks for legacy systems
+#ifndef NSURLVolumeLocalizedNameKey
+#define NSURLVolumeLocalizedNameKey @"NSURLVolumeLocalizedNameKey"
+#endif
+#ifndef NSURLVolumeIsLocalKey
+#define NSURLVolumeIsLocalKey @"NSURLVolumeIsLocalKey"
+#endif
+#ifndef NSURLVolumeIsReadOnlyKey
+#define NSURLVolumeIsReadOnlyKey @"NSURLVolumeIsReadOnlyKey"
+#endif
+#ifndef NSURLVolumeIsRemovableKey
+#define NSURLVolumeIsRemovableKey @"NSURLVolumeIsRemovableKey"
+#endif
+
 //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 // wxFSVolume
 //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
 wxArrayString wxFSVolumeBase::GetVolumes(int flagsSet, int flagsUnset)
 {
+#ifdef USE_LEGACY_OBJC
+    NSArray* nativeVolumes = [[NSFileManager defaultManager]
+#else
     auto nativeVolumes = [[NSFileManager defaultManager]
+#endif
                           mountedVolumeURLsIncludingResourceValuesForKeys:nil
                           options:NSVolumeEnumerationSkipHiddenVolumes];
-
 
     wxArrayString volumePaths;
     if ( nativeVolumes == nil )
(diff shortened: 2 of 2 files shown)


src/osx/volume.mm
```diff
@@ -22,6 +22,11 @@
 
 #include "wx/volume.h"
 
+// gcc does not yet support modern Apple ObjC++ syntax.
+#ifndef __clang__
+#define USE_LEGACY_OBJC
```
vadz, line 27: It is not necessary to have 2 versions of the code, if we need to use gcc-compatible syntax, let's just use it.

src/osx/volume.mm
```diff
@@ -35,17 +40,34 @@
...
+#define NSURLVolumeIsRemovableKey @"NSURLVolumeIsRemovableKey"
+#endif
+
 //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 // wxFSVolume
 //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
 wxArrayString wxFSVolumeBase::GetVolumes(int flagsSet, int flagsUnset)
 {
+#ifdef USE_LEGACY_OBJC
```
vadz, line 63: This change doesn't make sense to me, Objective C code is the same in both branches.

build/files
```diff
@@ -176,6 +176,8 @@ BASE_OSX_SHARED_HDR =
 # Base and GUI files used by OS X
 BASE_AND_GUI_OSX_COCOA_SRC =
     src/osx/cocoa/utils.mm
+
+BASE_AND_GUI_DARWIN_SRC =
```
vadz, line 180: I think this file should be just moved to `BASE_OSX_SHARED_SRC`, i.e. sources used in wxBase in all ports under Mac, and `wxFSVolume::InitIcons()` and `wxFSVolume::GetIcon()` be implemented inline in `include/wx/volume.h` under Darwin, as they're just stubs anyhow (in fact, we should really remove `InitIcons()` entirely, it's useless).

src/osx/volume.mm
barracuda156: If that is acceptable, sure.

build/files
barracuda156: Is it compiled with `wxUSE_GUI`? Initially I added it to base files in corefoundation, I think, and that did not produce a working library (i.e. the specific app which needed that code failed to launch).

But getting rid of unneeded stubs is preferable, of course.

vadz: See #25492 for a start.

vadz:
BTW, if we're serious about supporting wxGTK under macOS, we really need to have a CI job checking for this, it would be great if you could find some way to set this up.

barracuda156:
> BTW, if we're serious about supporting wxGTK under macOS, we really need to have a CI job checking for this, it would be great if you could find some way to set this up.

We can probably borrow from MacPorts (BSD license), it has CI for three latests macOS versions. They do not have a port for wxGTK 3.2, but I do.

P. S. Setting up CI to build with gcc and libstdc++ may be non-trivial, at least on the latest macOS (Apple keeps breaking something, gcc upstream tries to catch up to fix that, so that is not guaranteed to work at any given moment), but libc++ can be used with clang and gcc both.

MaartenBent:
The GitHub Actions macOS images have GCC pre-installed, see the readme of the different images: 
https://github.com/actions/runner-images/tree/main/images/macos

So it might be as simple as adding a new job to the matrix with the correct configure/cmake arguments in 
https://github.com/wxWidgets/wxWidgets/blob/master/.github/workflows/ci_mac.yml (for configure) or
https://github.com/wxWidgets/wxWidgets/blob/master/.github/workflows/ci_cmake.yml (for (CMake)

barracuda156:
@MaartenBent I would be very surprised if they use libstdc++.

vadz:
> @MaartenBent I would be very surprised if they use libstdc++.

How else would gcc work? AFAIK it can't use libc++.

barracuda156:
> > @MaartenBent I would be very surprised if they use libstdc++.
> 
> How else would gcc work? AFAIK it can't use libc++.

It can, since gcc10, I think. Certainly with recent versions. It can use libc++ even on powerpc, though I may be the only one who tried it LOL

With libstdc++ the issue is not to build gcc as such, but to build other stuff against it. Suffice to say that it was painful enough for me to try and give up twice after fixing a few broken ports on the way (on 10.15 x86_64 and 14.x arm64).

All recent gcc versions in MacPorts use libc++ on x86 and arm64. (Well, a bootstrap version of gcc10 does not, since it is not supposed to be used to build anything besides bootstrapping toolchain on some systems.)

MaartenBent:
> @MaartenBent I would be very surprised if they use libstdc++.

Does it matter? The CI should test wxGTK on macOS.
And if it could use gcc instead if appleclang, that would allow to test two things at once.

barracuda156:
> > @MaartenBent I would be very surprised if they use libstdc++.
> 
> Does it matter? The CI should test wxGTK on macOS.
> 
> And if it could use gcc instead if appleclang, that would allow to test two things at once.

That was not to say that CI won’t work, of course. But setting them with clang is enough, IMO, since the meaningful set-up with gcc is problematic for CI. (C++ runtime matters a lot.)

csomor:
I moved volume.mm also in #25919

barracuda156:
@csomor Yeah, I will test your PR today (within a couple of hours, I guess). We will probably need to deal with ObjC(++) syntax incompatibilities with some of those components though.

csomor:
> BTW, if we're serious about supporting wxGTK under macOS, we really need to have a CI job checking for this, it would be great if you could find some way to set this up.

#25924 but only with clang, so for other setups more entries in the matrix and more steps will be needed

barracuda156:
@csomor We don’t need this anymore, right?

csomor:
> @csomor We don’t need this anymore, right?

no, that's in #25919

barracuda156:
Closing, since hopefully redundant now.

PR #25493, opened 2025-06-02
Fix Infobar button regression issues

Blake-Madden:
There was no unit tests for `wxInfoBar`, so I added a new file for it. I'm not sure if I added it correctly to everything, though, but I did manually try these tests.

### DIFF
--- a/src/generic/infobar.cpp
+++ b/src/generic/infobar.cpp
@@ -138,7 +138,6 @@ bool wxInfoBarGeneric::Create(wxWindow *parent, wxWindowID winid, long style)
     firstRowSizer->Add(m_icon, wxSizerFlags{}.Centre().Border());
     firstRowSizer->Add(m_text, wxSizerFlags{}.Proportion(1).Centre());
     firstRowSizer->AddSpacer(0); // This spacer only exists for compatibility.
-    firstRowSizer->Add(m_button, wxSizerFlags{}.Centre().Border());
 
     wxBoxSizer* const secondRowSizer = new wxBoxSizer(wxHORIZONTAL);
     secondRowSizer->Add(m_checkbox, wxSizerFlags{}.CentreVertical().Border());
@@ -148,6 +147,7 @@ bool wxInfoBarGeneric::Create(wxWindow *parent, wxWindowID winid, long style)
     defaultControlSizer->Show(m_checkbox, !m_checkbox->GetLabel().empty(), true);
 
     sizer->Add(defaultControlSizer, wxSizerFlags{ 1 }.Expand());
+    sizer->Add(m_button, wxSizerFlags{}.Centre().Border());
 
     SetSizer(sizer);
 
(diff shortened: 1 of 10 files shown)


MaartenBent:
> I'm not sure if I added it correctly to everything

The new file should be added to CMake too, in https://github.com/wxWidgets/wxWidgets/blob/master/build/cmake/tests/gui/CMakeLists.txt

Did you manually change the `.in`/`.gcc`/`.vc` files? These files are generated from `test.bkl` using bakefile. You can add the new file here https://github.com/wxWidgets/wxWidgets/blob/master/tests/test.bkl, and then regenerate.

Only the `vcxproj` files need manual editing. (And the `.mss` file, if this is still in use).

Blake-Madden:
Closes #25489

Blake-Madden:
Thanks, @MaartenBent. I should have it added to everything now, and I see it is running in here.

tests/controls/infobar.cpp
```diff
@@ -0,0 +1,45 @@
...
+#ifndef WX_PRECOMP
+    #include "wx/app.h"
+#endif // WX_PRECOMP
+
+#include "wx/infobar.h"
+
+#include <memory>
+
+namespace
+{
```
vadz, line 25: This seems inconsistent with the other tests, so I'll remove it if you don't mind.

tests/controls/infobar.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+    TEST_CASE("wxInfoBar", "[wxInfoBar]")
+    {
+        SECTION("Buttons")
+        {
+            const std::unique_ptr<wxInfoBar>
+                info(new wxInfoBar(wxTheApp->GetTopWindow(), wxID_ANY, wxINFOBAR_CHECKBOX));
+
+            CHECK(0 == info->GetButtonCount());
```
vadz, line 34: Just FYI, you don't need to use Yoda comparison with CATCH.

src/generic/infobar.cpp
```diff
@@ -355,9 +355,10 @@ size_t wxInfoBarGeneric::GetButtonCount() const
     {
         const wxSizerItem * const item = node->GetData();
 
-        // if we reached the spacer separating the buttons from the text
-        // break the for-loop.
-        if ( item->IsSpacer() )
+        // If we reached the spacer separating the buttons from the text
+        // break the for-loop. Same for if we get to the sizer holding
+        // the icon and text.
+        if ( item->IsSpacer() || item->IsSizer() )
```
vadz, line 361: I wonder if it would be worth extracting this test into a helper function instead of repeating it several times. The advantage would be that we could use some descriptive name for it, e.g. `IsButtonItem()` (here the test for it would be reversed).

What do you think?

vadz (approved):
Thanks a lot, having a test for this is really reassuring!

I can apply it as is, or we could add a helper function as suggested. I'm not sure if it's worth it, so I'm leaving it up to you to decide, please let me know.

tests/controls/infobar.cpp
Blake-Madden: Yeah, I saw other files doing that so I wasn't sure. Totally OK with removing that.

tests/controls/infobar.cpp
Blake-Madden: Sorry, old habit :)

src/generic/infobar.cpp
Blake-Madden: I'm OK with it as it is, but I can make that change if you really want me to. This way is a little more readable at least (you can see that it is checking for the sizer or spacer). Adding a `IsButtonItem()` (that isn't really checking to see if it is actually a `wxButton`) seems a little misleading to me. IMHO

PR #25494, opened 2025-06-03
Don't duplicate LIBS and LDFLAGS in EXTRALIBS

vadz:
LIBS and LDFLAGS are used in all linker commands anyhow, there is no need to add them to EXTRALIBS, especially because this triggers an annoying warning about "ignoring duplicate libraries" from macOS ld.

Closes #25054.


vadz:
Merged in the commit above.

PR #25495, opened 2025-06-03
MSVS CI jobs reorganization

vadz:
Remove MSVS 2019 CI jobs from GitHub Actions as windows-2019 image is being retired.


vadz:
Hmm, so I've added a wxQt5 build to `ci_msw.yml` without noticing that there was already a wxQt6 build in `ci_cmake.yml`. I should probably move this build to `ci_cmake.yml` too, but what about "msw-msys2-clang" build which also uses CMake, should it be moved as well? Logically it makes sense, but could this make CMake workflow take longer? Or it doesn't matter how we split our jobs between workflows? I'm not really sure how does this work on GitHub, so any advice would be welcome.

MaartenBent:
I think all jobs run in parallel, so it doesn't matter how many jobs are added to the matrix in one CI file. 
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#defining-the-maximum-number-of-concurrent-jobs

It would indeed make sense to move `msw-msys2-clang` to `ci_cmake`, and use `ci_msw` only for the Visual Studio solutions.
Are `makefile.gcc`/`makefile.vc` tested anywhere? Maybe those can be in `ci_msw` too.

PBfordev:
> Are `makefile.gcc`/`makefile.vc` tested anywhere? 

It seems that `makefile.vc` is used [by AppVeyor](https://github.com/wxWidgets/wxWidgets/blob/cfff7223ad9ec0d33c2c56b2f4ba1b94a326bec4/appveyor.yml#L42); however,  `makefile.gcc` is not (although there is still a support for it in appveyor.bat).

MaartenBent:
Indeed, looks like it used to have [builds](https://github.com/wxWidgets/wxWidgets/blob/master/build/tools/appveyor.bat#L30) with `makefile.gcc` and `mingw64+configure` in the past, but those got removed in 04f30658cdf83f47d822177993d99af2ddc17831

Maybe they could be re-added in `ci_msw`, and move the `makefile.vc` build there as well. Because AppVeyor is still the slowest CI with only sequential builds.

vadz:
I think we need to get rid of `makefile.{vc,gcc}` anyhow and especially of the VC ones as this is the worst way to build wx: there is no parallelism in nmake, so it takes much longer. And while GNU make is more or less fine, we have a long standing problem with `-j`-unsafed setup.h target... So I am not sure we need CI builds for them or, at least, not ones running for every push/PR — but it could be nice to run them manually to check that they still work before the release.

I'll move the CMake builds as suggested, but I probably won't have time to do it before the end of the week. If you'd like to do it sooner, please feel free to make your own PR replacing this one, just please let me know if you decide to do it. TIA!

vadz:
I've fixed it now, but I still don't understand why the build using native clang 18.1.8 which was accidentally used when I forgot to use msys2 shell resulted in tons of warnings (which were fatal due to `-Werror`) while the build using MSYS2 clang 20.1.5 gives many fewer warnings (but still gives some, which are somehow not fatal, even though we still use `-Werror` with it).

But I'm going to merge this anyhow if there are no comments/objections.

MaartenBent:
I checked past actions and found at least these two that are using `C:/Program Files/LLVM/bin/clang.exe`:
https://github.com/wxWidgets/wxWidgets/actions/runs/15714441700/job/44280535223
https://github.com/wxWidgets/wxWidgets/actions/runs/15713446147/job/44277197704

All the CRT warnings are probably because it uses msvc win32 sdk instead of the msys2/mingw sdk.

One error is `-Wunused-local-typedef`, even though it has `WX_ATTRIBUTE_UNUSED`. It seems the native clang doesn't define `__GNUC__` so `WX_ATTRIBUTE_UNUSED` is empty instead of `__attribute__ ((unused))`.

The other error is `-Wunused-variable`, which is enclosed in a `#if defined(__VISUALC__)` segment. Which is not defined in mingw.

I guess native clang, and probably clang-cl too, define themselves as `__clang__` and `__VISUALC__` instead of `__clang__` and `__GNUC__`.

vadz:
Thanks for looking into this, Maarten! I still didn't have time to actually check/confirm this, but your hypothesis makes sense and if native clang doesn't define `__GNUC__`, we really need to review our checks for clang to not rely on it being detected as gcc. This is obviously not urgent, but Id' actually like to be able to use it under Windows, so I've opened #25581 for this.

PR #25497, opened 2025-06-06
Improve wxRect2DDouble

Blake-Madden:
Add functions to give it parity with `wxRect`: explicit `wxRect `CTOR, `GetX`, `GetY`, `Set/GetWidth`, `Set/GetHeight`, `Offset` overload, `Inflate`, `Deflate`, `ToRect` conversion helper.

Add `wxNODISCARD `where appropriate.

Added as much documentation as I could to interface file.

Behaviour differing from `wxRect` is now explained.

### DIFF
--- a/include/wx/geometry.h
+++ b/include/wx/geometry.h
@@ -519,7 +519,7 @@ inline bool wxPoint2DDouble::operator!=(const wxPoint2DDouble& pt) const
     return !(*this == pt);
 }
 
-// wxRect2Ds are an axis-aligned rectangles, each side of the rect is parallel to the x- or m_y- axis. The rectangle is either defined by the
+// wxRect2Ds are axis-aligned rectangles, each side of the rect is parallel to the x- or m_y- axis. The rectangle is either defined by the
 // top left and bottom right corner, or by the top left corner and size. A point is contained within the rectangle if
 // left <= x < right  and top <= m_y < bottom , thus it is a half open interval.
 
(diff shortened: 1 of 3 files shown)


Blake-Madden:
Note that I did NOT add `SetX` and `SetY` functions to avoid confusion. Because `SetLeft` and `SetTop` behave differently between `wxRect` and `wxRect2DDouble`, I wasn't sure how `SetX` and `SetY` would be expected to work. But the current `Move` functions will do what `SetX` and `SetY` do in `wxRect` and that is all documented now.

include/wx/geometry.h
```diff
@@ -529,18 +529,48 @@ class WXDLLIMPEXP_CORE wxRect2DDouble
     wxRect2DDouble() = default;
     wxRect2DDouble(wxDouble x, wxDouble y, wxDouble w, wxDouble h)
         { m_x = x; m_y = y; m_width = w;  m_height = h; }
+
+    explicit wxRect2DDouble(const wxRect& rect)
```
vadz, line 533: For once, I think, the conversion could be made implicit: it seems to make sense to allow passing `wxRect` where `wxRect2DDouble` is expected, why not?

include/wx/geometry.h
```diff
@@ -529,18 +529,48 @@ class WXDLLIMPEXP_CORE wxRect2DDouble
...
+        m_x = static_cast<wxDouble>(rect.x);
+        m_y = static_cast<wxDouble>(rect.y);
+        m_width = static_cast<wxDouble>(rect.width);
+        m_height = static_cast<wxDouble>(rect.height);
+    }
+
+    wxNODISCARD wxRect ToRect() const
+    {
+        return wxRect(static_cast<int>(m_x), static_cast<int>(m_y),
+                      static_cast<int>(m_width), static_cast<int>(m_height));
```
vadz: I'm never sure about this, but maybe we should use `wxRound()` instead?

interface/wx/geometry.h
```diff
@@ -126,83 +127,406 @@ wxPoint2DDouble operator/(const wxPoint2DDouble& pt , wxInt32 n);
...
+    */
+    wxRect2DDouble(wxDouble x, wxDouble y, wxDouble width, wxDouble height);
 
-    // single attribute accessors
+    /**
+        Constructs a wxRect2DDouble from a wxRect.
 
+        @since 3.3.0
+    */
+    wxRect2DDouble(const wxRect& rect);
```
vadz: If this ctor stays `explicit`, it should be documented as such too.

vadz (commented):
Thanks, no real comments, but just a couple of minor questions below.

include/wx/geometry.h
Blake-Madden: I originally tried that, but got a ton of ambiguity errors when building my other project against it. Explicit seems like the best option for compatibility and avoiding compile issues.

interface/wx/geometry.h
Blake-Madden: Got it.

include/wx/geometry.h
Blake-Madden: Guess that will make it more deterministic. Done.

vadz:
Thanks! Will merge soon.

PR #25501, opened 2025-06-07
Fix building webrequest and webview samples

PBfordev:
Using a const char* and a wxString in the two branches of the ternary
operator resulted in compile-time errors when using C++20 standard
since the result type of the conditional expression was ambiguous,
so add explicit conversions to fix it.

See also #19355.

---------
This should be backported to 3.2.

### DIFF
--- a/samples/webrequest/webrequest.cpp
+++ b/samples/webrequest/webrequest.cpp
@@ -55,7 +55,7 @@ public:
         mainSizer->Add(new wxStaticText(this, wxID_ANY, "Request URL:"),
             wxSizerFlags().Border());
         m_urlTextCtrl = new wxTextCtrl(this, wxID_ANY,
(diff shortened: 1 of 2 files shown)


vadz:
Sorry, I don't understand what's going on here: the (only) commit in this PR seems to be empty?

PBfordev:
Sorry, I messed up when fixing the commit, should be OK now.

vadz:
Thanks, the changes look good but I wonder how did our C++20 CI job passed without them...

[checking...]

So it seems that only MSVS 2022 fails to compile this, g++ 14.2 is happy with it even with `-std=c++20` (and also `-std=gnu++20`).

The conclusion is that I need to make sure #25495 compiles the samples with MSVS 2022 in C++20 mode too.

Issue #25443, opened 2025-05-23
tab navigation doesn't wrap around back to a wxNotebook like is standard behavior


trypsynth:
#### Bug description:
When putting some controls on a panel for a notebook and then trying to tab through them, my focus gets stuck on the last control, and I'm unable to press tab again to wrap back around to the notebook, like is standard win32 behavior.

#### Expected vs observed behaviour:
The tab navigation should work normally, like it would in any other Windows app. If I press tab from the second button, it should wrap back around and put my focus on the notebook.

#### Patch or snippet allowing to reproduce the problem:

```
#include <wx/wx.h>
#include <wx/notebook.h>

class MyApp : public wxApp {
public:
	bool OnInit() override {
		auto* frame = new wxFrame(nullptr, wxID_ANY, "Notebook", wxDefaultPosition, wxSize(300, 200));
		auto* notebook = new wxNotebook(frame, wxID_ANY);
		auto* panel = new wxPanel(notebook, wxID_ANY);
		auto* sizer = new wxBoxSizer(wxVERTICAL);
		sizer->Add(new wxButton(panel, wxID_ANY, "Button 1"), 0, wxALL | wxEXPAND, 5);
		sizer->Add(new wxButton(panel, wxID_ANY, "Button 2"), 0, wxALL | wxEXPAND, 5);
		panel->SetSizer(sizer);
		notebook->AddPage(panel, "Tab");
		frame->Show();
		return true;
	}
};

wxIMPLEMENT_APP(MyApp);
```

#### To Reproduce:
1. Run NVDA Or Narrator.
2. Run the provided sample.
3. Tab around.

### Platform and version information

- wxWidgets version you use: latest from master
- wxWidgets port you use: wxMsw
- OS and its version: Windows 10 21H2 (AMD64) build 19044.5854

trypsynth:
Can confirm it also happens with wxPython:
```
import wx


class MyApp(wx.App):
    def OnInit(self):
        frame = wx.Frame(None, title="Notebook", size=(300, 200))
        notebook = wx.Notebook(frame)
        panel = wx.Panel(notebook)
        sizer = wx.BoxSizer(wx.VERTICAL)
        sizer.Add(wx.Button(panel, label="Button 1"), 0, wx.ALL | wx.EXPAND, 5)
        sizer.Add(wx.Button(panel, label="Button 2"), 0, wx.ALL | wx.EXPAND, 5)
        panel.SetSizer(sizer)
[...]

PR #25507, opened 2025-06-10
3.2.9 backports

vadz:


vadz:
Merged into 3.2.

MaartenBent:
There is also #25443 that has the `backport-3.2` label and can be backported, but has already closed. 

And there are a few more closed issues with this label: https://github.com/wxWidgets/wxWidgets/issues?q=state%3Aclosed%20label%3A%22backport-3.2%22

vadz:
> There is also #25443 that has the `backport-3.2` label and can be backported, but has already closed.

Thanks, it was closed by the submitter, I've added it to the new backport PR #25674.

> And there are a few more closed issues with this label: https://github.com/wxWidgets/wxWidgets/issues?q=state%3Aclosed%20label%3A%22backport-3.2%22

I'm actually only removing the label after backporting, so it's normal for there to be already closed issues with it, but I've just added 3 of them to the same PR too. The last one requires more work...

PR #25523, opened 2025-06-14
Add copyright info to libpng and libjpeg's GetLibraryVersionInfo(), fix libtiff's

Blake-Madden:
Also adds lib name to PCRE's description to make it consistent with other libraries. Below are how the current descriptions appear:

image

Also, fixes libtiff's copyright. Currently, it looks like this:

"Copyright (c) 1988-1996 Sam LefflerCopyright (c) 1991-1996 Silicon Graphics, Inc."

Note the "LefflerCopyright". Newlines will now be preserved.

Also, use `wxString::Trim` instead of using pointer arithmetic that makes assumptions when getting _libpng_ description.

### DIFF
--- a/src/common/imagjpeg.cpp
+++ b/src/common/imagjpeg.cpp
@@ -506,7 +506,6 @@ bool wxJPEGHandler::DoCanRead( wxInputStream& stream )
 {
     struct jpeg_error_mgr err{};
     jpeg_std_error(&err);
-    err.last_jpeg_message = JMSG_COPYRIGHT;
 
 #if defined(JPEG_LIB_VERSION_MAJOR) && defined(JPEG_LIB_VERSION_MINOR)
     return wxVersionInfo("libjpeg", JPEG_LIB_VERSION_MAJOR, JPEG_LIB_VERSION_MINOR,
--- a/src/common/imagjpeg.cpp
+++ b/src/common/imagjpeg.cpp
@@ -35,7 +35,9 @@
 #endif
 
 #include "jpeglib.h"
+#if wxHAS_CXX17_INCLUDE("jversion.h")
 #include "jversion.h"
+#endif
 
 #include "wx/filefn.h"
 #include "wx/wfstream.h"
--- a/src/common/imagjpeg.cpp
+++ b/src/common/imagjpeg.cpp
@@ -35,6 +35,7 @@
 #endif
 
 #include "jpeglib.h"
+#include "jversion.h"
 
 #include "wx/filefn.h"
 #include "wx/wfstream.h"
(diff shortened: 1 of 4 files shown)


src/common/imagpng.cpp
```diff
@@ -851,14 +851,30 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
...
+            copyRight.Trim(false);
+        }
+    }
 
     return wxVersionInfo("libpng",
                          PNG_LIBPNG_VER_MAJOR,
                          PNG_LIBPNG_VER_MINOR,
                          PNG_LIBPNG_VER_RELEASE,
-                         str);
+                         0,
```
MaartenBent: You could use `PNG_LIBPNG_VER_BUILD` here.

Blake-Madden: Got it, thanks.

src/common/imagjpeg.cpp
```diff
@@ -35,6 +35,9 @@
 #endif
 
 #include "jpeglib.h"
+#if wxHAS_CXX17_INCLUDE("jversion.h")
```
vadz: Do we really need this? It looks like this file exists since at least 1991, so why check for it?

src/common/imagpng.cpp
```diff
@@ -851,14 +851,30 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
...
+    copyRight.Trim(true).Trim(false);
+    // The copyright string may include the version info in the first line.
+    // If it does, then remove it.
+    if (copyRight.starts_with("libpng"))
+    {
+        size_t firstNewLine = copyRight.find(L'\n');
+        if (firstNewLine != wxString::npos)
+        {
+            copyRight.erase(0, firstNewLine);
+            copyRight.Trim(false);
```
vadz: Looks like this could be removed if we used `firstNewLine + 1` above?

vadz (commented):
Thanks, looks good, I'd just like to check the `jversion.h` thing.

src/common/imagjpeg.cpp
Blake-Madden: When I would just include the header file, I got errors on some of the UNIX builds saying that the file was missing. My guess was that these configurations weren't using the builtin libjpeg but the system one instead and for whatever reason this file was missing in the system header files. So, wrapping it in a `wxHAS_CXX17_INCLUDE` prevented that.

src/common/imagpng.cpp
Blake-Madden: Got it.

src/common/imagjpeg.cpp
vadz: Yes, sorry, you're right, it's not present in libjpeg installed from the package.

But the reason for this seems to be that it's an internal header which is not supposed to be used from the outside. Looking at `jerror.h`, it looks like there should be some way to get it as it's the message associated with (public) `JMSG_COPYRIGHT` code, can you please check if we can't get to it using public libjpeg API?

Blake-Madden: I changed it to call `jpeg_std_error` to fill an object with the libjpeg's string table and then access the `JMSG_COPYRIGHT` from there.

I'm reading the string table directly instead of calling `format_message`. I think that is better so that I can just copy the text into a `wxString` instead of creating a temp buffer that gets routed to `sprintf`. Ultimately, all `format_message` does is the same as reading the index from the string table.

vadz:
Thanks, I think this is good now, will merge soon.

PR #25524, opened 2025-06-14
Fix a few comment misspellings

Blake-Madden:


src/common/zipstrm.cpp
```diff
@@ -346,7 +346,7 @@ size_t wxStoredOutputStream::OnSysWrite(const void *buffer, size_t size)
 /////////////////////////////////////////////////////////////////////////////
 // wxRawInputStream
 //
-// Used to handle the unusal case of raw copying an entry of unknown
+// Used to handle the unusual case of raw copying an entry of unknown
 // length. This can only happen when the zip being copied from is being
 // read from a non-seekable stream, and also was original written to a
```
ojwb: There's another typo here - should be "was originally written".

Blake-Madden: Got it, thanks!

vadz:
Thanks, will merge soon.

PR #25525, opened 2025-06-14
Add an example to wxVersionInfo

Blake-Madden:
I think this example is useful in particular because it shows which submodules (at least at this moment) support version info. It is also useful as it shows the libraries where you have to call this a little differently (e.g., `wxGetZlibVersionInfo()`, `wxWebView::GetBackendVersionInfo()`).

### DIFF
--- a/interface/wx/versioninfo.h
+++ b/interface/wx/versioninfo.h
@@ -33,6 +33,52 @@ enum class wxVersionContext
     provide version information about your own libraries, or other libraries
     that you use.
 
+    The following example shows how to get most of wxWidgets's library information
+    and format it all into a string.
+    (Note that this assumes all libraries are being linked and @c wxInitAllImageHandlers()
+     has already been called.)
+
+    @code
+    // Consistently format library info with the name and version number.
+    const auto formatLibInfo = [](const auto& libInfo)
+    {
+        return wxString::Format(_("%s - version %s"), libInfo.GetName(),
+                                libInfo.GetNumericVersionString());
+    };
+
+    // Get wxWidgets's version--as well as its submodules--and sort them.
+    std::vector<wxString> allLibInfo;
+    allLibInfo.push_back(formatLibInfo(wxGetLibraryVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxTIFFHandler::GetLibraryVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxJPEGHandler::GetLibraryVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxPNGHandler::GetLibraryVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxWEBPHandler::GetLibraryVersionInfo()));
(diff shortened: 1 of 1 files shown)


vadz:
This looks a bit complicated for an example, but I could be overthinking this. Do the other think we should have it?

Maybe it would be useful to include this somewhere else, although it's going to be hard to find a sample linking to all these libraries. Maybe put it into a "manual only" test?

Blake-Madden:
I could provide a much simpler example of creating a `wxVersionInfo` for something like Lua. Let me know if you prefer that.

vadz:
I think a simpler example would be better, thanks!

interface/wx/versioninfo.h
```diff
@@ -33,6 +33,52 @@ enum class wxVersionContext
...
+    allLibInfo.push_back(formatLibInfo(wxTIFFHandler::GetLibraryVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxJPEGHandler::GetLibraryVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxPNGHandler::GetLibraryVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxWEBPHandler::GetLibraryVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxRegEx::GetLibraryVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxXmlDocument::GetLibraryVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxStyledTextCtrl::GetLibraryVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxWebSession::GetDefault().GetLibraryVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxGetZlibVersionInfo()));
+    allLibInfo.push_back(formatLibInfo(wxWebView::GetBackendVersionInfo()));
```
Jarod42:
```suggestion
    std::vector<wxString> allLibInfo {
        formatLibInfo(wxGetLibraryVersionInfo()),
        formatLibInfo(wxTIFFHandler::GetLibraryVersionInfo()),
        formatLibInfo(wxJPEGHandler::GetLibraryVersionInfo()),
        formatLibInfo(wxPNGHandler::GetLibraryVersionInfo()),
        formatLibInfo(wxWEBPHandler::GetLibraryVersionInfo()),
        formatLibInfo(wxRegEx::GetLibraryVersionInfo()),
        formatLibInfo(wxXmlDocument::GetLibraryVersionInfo()),
        formatLibInfo(wxStyledTextCtrl::GetLibraryVersionInfo()),
        formatLibInfo(wxWebSession::GetDefault().GetLibraryVersionInfo()),
        formatLibInfo(wxGetZlibVersionInfo()),
        formatLibInfo(wxWebView::GetBackendVersionInfo())};
```
unless C++03 support is really required ;-)

Blake-Madden: Old habits die hard :-)

Blake-Madden:
Alright, much simpler example is in there now. And it actually flows better with the previous paragraph too.

vadz:
Thanks, will push soon.

PR #25538, opened 2025-06-17
Fix creating empty accelerator table in wxMSW

vadz:
See #25517.

### DIFF
--- a/interface/wx/accel.h
+++ b/interface/wx/accel.h
@@ -200,7 +200,8 @@ public:
         Initializes the accelerator table from an array of wxAcceleratorEntry.
 
         @param n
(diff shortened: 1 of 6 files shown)


vadz:
So it turns out that wxMSW was an outlier and that creating `wxAccelTable` with 0 entries creates a valid (its `IsOk()` returns `true`) object in all the other ports. wxMSW behaviour still makes (more) sense to me, so I plan to update all the other ports, but please let me know if anybody disagrees and think it's wxMSW that should be changed (note that while this PR fixes the assert, which is a recent regression, it always created invalid object as `::CreateAcceleratorTable()` fails when passed 0 entries).

PR #25539, opened 2025-06-17
Fix wxPersistent(CheckBox|RadioButton) version availability in docs

PBfordev:
wxPersistentCheckBox and wxPersistentRadioButton did not make it into wxWidgets 3.3.0, they are available only since 3.3.1.

### DIFF
--- a/interface/wx/persist/checkbox.h
+++ b/interface/wx/persist/checkbox.h
@@ -22,7 +22,7 @@
     If the checkbox is checked, it will be checked again after the application
     restart.
 
-    @since 3.3.0
+    @since 3.3.1
  */
 class wxPersistentCheckBox : public wxPersistentWindow<wxCheckBox>
 {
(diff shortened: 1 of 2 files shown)


vadz:
Oops, you're right, of course, thanks for noticing and fixing this!

PBfordev:
@vadz, I just noticed that these two are missing in the list of supported classes in https://forum.wxwidgets.org/trunk/overview_persistence.html#persistence_using

However, `wxPersistentComboBox` is missing there as well. Perhaps instead of maintaining the list there, it could be simpler to remove it from Using Persistent Windows and from there just refer to the inheritance diagram in https://forum.wxwidgets.org/trunk/classwx_persistent_object.html

vadz:
Thanks for noticing, I'll add them to the overview for now, I think it's more readable as text rather than in the image.

Issue #25529, opened 2025-06-14
What is the risk if I patch v3.3 to force enable dark theme for Windows 10 LTSC 2019 (based on 1809)


ChipZhang:
Windows 10 LTSC 2019 (based on 1809) is a long term support version with at least 10 years life cycle. While the last Win10 LTSC version, LTSC 2021, has only five years support. This makes LTSC 2019 being the last one that reaches end of life among all Win10 versions. As a result, many enterprise users are using LTSC 2019.

So I am just curious, is it possible to enable dark mode for 1809. I notices the follows in file `darkmode.cpp`:

```c++
    // Note: for simplicity, we support dark mode only in Windows 10 v2004
    // ("20H1", build number 19041) and later, even if, in principle, it could
    // be supported as far back as v1809 (build 17763) -- but very few people
    // must still use it by now and so it just doesn't seem to be worth it.
    if ( !wxCheckOsVersion(10, 0, 19041) )
    {
        wxLogTrace(TRACE_DARKMODE, "Unsupported due to OS version");
        return false;
    }
```

I patched this code to check against 17763 (1809) rather than 19041 (20H1), and compiled demos/samples. In my non-thorough test, it seems those example exe files run well in Windows 10 LTSC 2019 with dark mode enabled. The only different behavior compared to 20H1, is that, title bar is black in 20H1 while in LTSC 2019 it is white.

Is there any potential risk or bug of doing this? If not, why didn't you just check against 1809 in the code to make dark mode available for LTSC 2019?

vadz:
I didn't have such system to test with it, so I preferred to not enable support for using undocumented APIs on an OS version I couldn't test. If things work fine there, we could enable support for it under 10.0.17763 too ­— it's not like we have any less guarantees about these APIs working there than under later Windows versions (in both cases we have none).

ChipZhang:
At least in my test (thought not very thorough), things works fine for LTSC 2019 (based on 1809, build 17763), except for the title bar color.

Perhaps someone else that has such Win10 versions may report more test results.

ChipZhang:
I just noticed, with this patch applied, build the samples or a hello world program with `-mconsole` (force attach a console window), and run them in Win10 1809, it prints error message:

> darkmode.cpp at line 426: 'DwmSetWindowAttribute(USE_IMMERSIVE_DARK_MODE)' failed with error 0x80070057

But the program actually works fine, both the functionality and dark theme work fine, despite this message.

[...]

PR #25541, opened 2025-06-18
Enable MSW dark mode support for v1809

ChipZhang:
Change the minimum required Windows version from 20H1 to v1809 (which Windows 10 LTSC 2019 is based on), and fix dark title bar support for v1809.

Also update the doc to note that dark mode support for versions before Windows 10 20H1 is not fully tested.


ChipZhang:
This PR closes #25529

vadz:
Thanks, will push soon.

PR #25545, opened 2025-06-18
MSW notebook fixes

vadz:


marekr:
Tab background color seems to work fine. I actually forgot until I tested that Windows 11 added 4 high contrast themes total and not just black/white

vadz:
Thanks for testing!

Just for the record, I have 4 high contrast themes under Windows 10 v19044 too, so it's not a Windows 11 thing.

PR #25546, opened 2025-06-19
wxGrid and wxHTML fixes

Kees-van-der-Oord-Nikon:
fixes for two flaws:
- in grid.cpp fix mix up of wxALIGN_CENTER, wxALIGN_CENTER_VERTICAL and wxALIGN_CENTER_HORIZONTAL
- in htmlpars.cpp added support for HTML charset attribute

### DIFF
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -7699,6 +7699,7 @@ void wxGrid::DrawTextRectangle(wxDC& dc,
                 x = rect.x + (rect.width - textWidth - GRID_TEXT_MARGIN);
             break;
 
+        case wxALIGN_CENTER:
         case wxALIGN_CENTER_VERTICAL:
             if ( textOrientation == wxHORIZONTAL )
                 y = rect.y + ((rect.height - textHeight) / 2);
@@ -7740,6 +7741,7 @@ void wxGrid::DrawTextRectangle(wxDC& dc,
                     y = rect.y + lineWidth + GRID_TEXT_MARGIN;
                 break;
 
+            case wxALIGN_CENTER:
             case wxALIGN_CENTRE_HORIZONTAL:
                 if ( textOrientation == wxHORIZONTAL )
                     x = rect.x + ((rect.width - lineWidth) / 2);
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -7699,7 +7699,7 @@ void wxGrid::DrawTextRectangle(wxDC& dc,
                 x = rect.x + (rect.width - textWidth - GRID_TEXT_MARGIN);
             break;
 
-        case wxALIGN_CENTRE:
+        case wxALIGN_CENTER_VERTICAL:
             if ( textOrientation == wxHORIZONTAL )
                 y = rect.y + ((rect.height - textHeight) / 2);
             else
@@ -7740,7 +7740,7 @@ void wxGrid::DrawTextRectangle(wxDC& dc,
                     y = rect.y + lineWidth + GRID_TEXT_MARGIN;
                 break;
 
-            case wxALIGN_CENTRE:
+            case wxALIGN_CENTRE_HORIZONTAL:
                 if ( textOrientation == wxHORIZONTAL )
                     x = rect.x + ((rect.width - lineWidth) / 2);
                 else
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for the fixes, but could you please explain how to reproduce the bug fixed by the first commit (in `wxGrid`)? It doesn't seem that obvious to me which alignment constant should be used here, so I'd really like to test the fix before merging it.

TIA!

Kees-van-der-Oord-Nikon:
Dear Vadz,

I fixed this in a rather large project that include the xlnt library to load .xlsx files, it would take me considerable time to make a minimal example. I hope the arguments below are sufficient for you to understand the details.

When calling wxGrid::DrawTextRectangle(dc,text,wxALIGN_CENTER_HORIZONTAL,wxALIGN_CENTER_VERTICAL,0),
the current implementation does not align the text correctly. The first pull request fixed that but it broke compatibility with the situation that the function is called like this:
wxGrid::DrawTextRectangle(dc,text,wxALIGN_CENTER,wxALIGN_CENTER,0);

I just pushed a second pull request that restores the original case line and only adds case lines to support wxALIGN_CENTER_HORIZONTAL and wxALIGN_CENTER_VERTICAL.

WR,

Kees

vadz:
The problem is that it's really not clear whether the parameter should be `wxALIGN_CENTRE` or `wxALIGN_CENTRE_{HORIZONTAL,VERTICAL}`. Accepting both of them should be fine, but perhaps it would be even better to change the caller using the latter constants to use `wxALIGN_CENTRE` instead?

How/when exactly does this function end up being called with `wxALIGN_CENTRE_{HORIZONTAL,VERTICAL}` in your application?

Kees-van-der-Oord-Nikon:
Dear Vadz,

The library that parses the xlsx returns perfectly correct values for
the horizontal and vertical alignment:

Horizontal: wxALIGN_LEFT, wxALIGN_CENTER_HORIZONTAL, wxALIGN_RIGHT

Vertical: wxALIGN_TOP, wxALIGN_CENTER_VERTICAL, wxALIGN_BOTTOM

That the constant wxALIGN_CENTER is defined as a convenient value to indicate
both horizontal and vertical alignment is an extra feature, on top of the basics above.

IMHO, the function wxGrid::DrawTextRectangle() should at least support the basic,
on top of that also handle the case when the align parameter also has irrelevant bits set.

WR,

Kees


Dr. Kees van der Oord
European Support Specialist
T: +31207099246
M: +31623367589
E: ***@***.***
Website
Nikon Europe B.V.
Stroombaan 14
1181 VX Amstelveen
Netherlands

Registered Chamber of Commerce Amsterdam No. 34139593
VAT No. NL 8091.99.889.B.01

This e-mail may contain information that is PRIVILEGED and/or CONFIDENTIAL. This e-mail is intended solely for the individual or entity to whom it is addressed. If you are not, or have reason to believe you are not the intended recipient of this e-mail please contact us by reply e-mail as soon as possible. Immediately after contacting us, please destroy this e-mail (including attachment(s)) from your computer system(s). Any unauthorized copying, disclosure or distribution of this e-mail (including attachment(s)) is not allowed. Nikon Europe BV hereby disclaims any and all liabilities which may arise in relation to any of its electronic communication. No obligation is entered into by virtue of this e-mail, unless confirmed in writing and duly signed by an authorized representative of Nikon Europe BV

vadz:
> The library that parses the xlsx returns perfectly correct values for the horizontal and vertical alignment: Horizontal: wxALIGN_LEFT, wxALIGN_CENTER_HORIZONTAL, wxALIGN_RIGHT Vertical: wxALIGN_TOP, wxALIGN_CENTER_VERTICAL, wxALIGN_BOTTOM That the constant wxALIGN_CENTER is defined as a convenient value to indicate both horizontal and vertical alignment is an extra feature, on top of the basics above. IMHO, the function wxGrid::DrawTextRectangle() should at least support the basic, on top of that also handle the case when the align parameter also has irrelevant bits set.

But in `wxGrid` code only `wxALIGN_CENTRE` is used... `DrawTextRectangle()` is unfortunately not documented, so it's not clear what its parameters are required to be, but it seems that it really expects `wxALIGN_CENTRE`.

Anyhow, it should probably be safe to apply this, even if it does make things rather messy.

PR #25548, opened 2025-06-19
Use `wxApp` class name as app ID

### LINKED ISSUE #24668: An API to define app_id for wayland

vadz:
Replaces #25486.

Closes #24668.

### DIFF
--- a/include/wx/taskbarbutton.h
+++ b/include/wx/taskbarbutton.h
@@ -223,7 +223,7 @@ typedef wxVector<wxTaskBarJumpListCategory*> wxTaskBarJumpListCategories;
 class WXDLLIMPEXP_CORE wxTaskBarJumpList
 {
 public:
-    wxTaskBarJumpList(const wxString& appID = wxEmptyString);
+    explicit wxTaskBarJumpList(const wxString& appID = wxEmptyString);
     virtual ~wxTaskBarJumpList();
     void ShowRecentCategory(bool shown = true);
     void HideRecentCategory();
--- a/interface/wx/app.h
+++ b/interface/wx/app.h
@@ -758,13 +758,18 @@ public:
         Sets the class name of the application.
 
         The class name is used in a platform specific manner. Currently it is
-        used as "Application User Model ID" under Windows, "app ID" when
-        using wxGTK with Wayland (provided GTK version is 3.24.22 or greater)
-        and is unused under the other platforms.
+        used as "Application User Model ID" under Windows (see [Microsoft
+        documentation][microsoft-docs]), "app ID" when using wxGTK 3.24.22 or
+        later with Wayland (see [Wayland documentation][wayland-docs]) and is
+        unused under the other platforms.
+
+        [microsoft-docs]: https://learn.microsoft.com/en-us/windows/win32/shell/appids
+        [wayland-docs]: https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id
 
         When it is used, the class name purpose is to allow the system to
         handle all windows with the same ID as belonging to the same
-        application, e.g. to group them together in the taskbar. By default the
+        application, e.g. to group them together in the taskbar (so the value
+        set here is used by wxTaskBarJumpList constructor). By default the
         application executable name is used as its ID, so it is not necessary
         to set the class name, but it may be useful to do it to specify a more
         unique string (typically by using a reverse domain name notation with
--- a/interface/wx/app.h
+++ b/interface/wx/app.h
@@ -755,8 +755,40 @@ public:
     void SetAppName(const wxString& name);
 
     /**
-        Sets the class name of the application. This may be used in a platform specific
-        manner to refer to the application.
+        Sets the class name of the application.
+
+        The class name is used in a platform specific manner. Currently it is
+        used as "Application User Model ID" under Windows, "app ID" when
+        using wxGTK with Wayland (provided GTK version is 3.24.22 or greater)
+        and is unused under the other platforms.
+
+        When it is used, the class name purpose is to allow the system to
+        handle all windows with the same ID as belonging to the same
+        application, e.g. to group them together in the taskbar. By default the
+        application executable name is used as its ID, so it is not necessary
+        to set the class name, but it may be useful to do it to specify a more
+        unique string (typically by using a reverse domain name notation with
+        the domain unique to the application vendor) or by specifying the same
+        ID in different applications that should be handled as a single one at
+        UI level.
+
+        Please note that SetClassName() must be called as early as possible and
+        definitely before creating any top-level windows to have an effect.
+        Typically it should be called in the constructor of the class derived
+        from wxApp, e.g.
+
+        @code
+        class MyApp : public wxApp
+        {
+        public:
+            MyApp() {
+                // Constructor shouldn't perform any non-trivial iniailization
+                // as the GUI is not available yet, but this function is fine
+                // to call.
+                SetClassName("com.example.myapp");
+            }
+        };
+        @endcode
 
         @see GetClassName()
     */
--- a/interface/wx/taskbarbutton.h
+++ b/interface/wx/taskbarbutton.h
@@ -616,11 +616,12 @@ public:
         Constructs the jump list.
 
         @param appID
-            Specifies a unique identifier for the application jump list, can be
-            empty by default.
+            Specifies a unique identifier for the application jump list. If it
+            is empty (default value), the global application user model ID,
+            which can be changed by using wxApp::SetClassName(), is used.
 
             See <a href="https://learn.microsoft.com/en-us/windows/win32/shell/appids">
-            Application User Model IDs</a> for further details.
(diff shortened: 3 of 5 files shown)


paulcor (approved):

PBfordev:
On MSW, is this the same `appID` as used in [wxTaskBarJumpList](https://docs.wxwidgets.org/trunk/classwx_task_bar_jump_list.html#ac14d758d41e574368d4db6664b75ab9d)?

If so, we probably can't use the `wxApp` one when passed an empty one to `wxTaskBarJumpList` now, but perhaps it could be cross-referenced in the docs?

vadz:
Good catch, thanks! It's indeed the same app ID and this should indeed be mentioned in the docs, I've updated this PR to do it.

Note that I don't think any changes in the code are needed as by leaving the `appID` parameter empty we should be already using the previously set global app ID.

marekr:
FYI

This creates a behavior change on Windows.

In particular, the Shift+Middle click function of the taskbar will no longer launch a new instance of the process when a explicit app user model id is set.

This can be fixed by also allowing the specification of the relaunch command on the windows's shell property store (SHGerPropertyStoreForWindow)

Probably needs to be explained to a user or documented. And it may not be wise to automatically set the relaunch command in case custom args are required. 


(Note kicad uses app user model id but we did it our own way)

vadz:
> This creates a behavior change on Windows.

Thanks, I had no idea...
 
> In particular, the Shift+Middle click function of the taskbar will no longer launch a new instance of the process when a explicit app user model id is set.
> 
> This can be fixed by also allowing the specification of the relaunch command on the windows's shell property store (SHGerPropertyStoreForWindow)

Do you already have code doing this somewhere I could steal^H^H^H^H^H use as an inspiration?

> Probably needs to be explained to a user or documented. And it may not be wise to automatically set the relaunch command in case custom args are required.

Yes, we could need to add some `wxApp::MSWSetRelaunchCommand()`.
 
> (Note kicad uses app user model id but we did it our own way)

Will you be able to switch to wx way once you require 3.3 or is there still anything missing here from your point of view?

marekr:
>Do you already have code doing this somewhere I could steal^H^H^H^H^H use as an inspiration?


Sure. This is actually hacky because I started looking at this due to a user complaint. But this does work.

The main bits:
- The property store is per window
- The app user model id must be configured on the property store as a separate step from SetCurrentProcessExplicitAppUserModelID. *REQUIRED*
- It may be very undesirable to support relaunch on some random sub-dialog, so I am only going to do this on KiCad's main frames
- PKEY_AppUserModel_RelaunchCommand configures the relaunch behavior. It is used both for pinned taskbar icons AND shift-middle click on running processes. *REQUIRED*
- win32 ::GetCommandLine() could be set to Relaunch command, however that may not be desirable for all applications, I am evaluating my options here in kicad, including appending a new arg to avoid content on already open files
- PKEY_AppUserModel_RelaunchDisplayNameResource is not required. However, it becomes the "tooltip" name for pinned applications. Instead of allowing it to pick the app name by default I believe

```
#include <propkey.h>
#include <propvarutil.h>


void KIPLATFORM::ENV::SetAppDetailsForWindow( wxWindow* aWindow, const wxString& aRelaunchCommand,
                                              const wxString& aRelaunchDisplayName )
{
    IPropertyStore* pps;
    HRESULT         hr = ::SHGetPropertyStoreForWindow( aWindow->GetHWND(), IID_PPV_ARGS( &pps ) );
    if( SUCCEEDED( hr ) )
    {
        PROPVARIANT pv;

        // This is required for any the other properties to actually work
        hr = ::InitPropVariantFromString( GetAppUserModelId(), &pv );

        if( SUCCEEDED( hr ) )
        {
            hr = pps->SetValue( PKEY_AppUserModel_ID, pv );
            PropVariantClear( &pv );
        }


        if( !aRelaunchCommand.empty() )
        {
            hr = ::InitPropVariantFromString( aRelaunchCommand.wc_str(), &pv );
        }
        else
        {
            // empty var
            ::PropVariantInit( &pv );
        }

        if( SUCCEEDED( hr ) )
        {
            hr = pps->SetValue( PKEY_AppUserModel_RelaunchCommand, pv );
            PropVariantClear( &pv );
        }

        if( !aRelaunchDisplayName.empty() )
        {
            hr = ::InitPropVariantFromString( aRelaunchDisplayName.wc_str(), &pv );
        }
        else
        {
            // empty var
            ::PropVariantInit( &pv );
        }

        if( SUCCEEDED( hr ) )
        {
            hr = pps->SetValue( PKEY_AppUserModel_RelaunchDisplayNameResource, pv );
            PropVariantClear( &pv );
        }

        pps->Release();
    }
}
```

marekr:
>Will you be able to switch to wx way once you require 3.3 or is there still anything missing here from your point of view?

I need to explore if we are using wxApp::GetClassName for anything else. For the app user model id we have a microsoft formatted string of `<organization>.<product>.<subproduct>.<version>` which we then pass onto our external third party plugin ecosystem to keep things unified. I don't want to blow things up because we were using GetclassName for something else

vadz:
Thanks for the code, but it looks like this might require adding some new API, as it doesn't seem possible to decide about the right thing to do inside wx itself.

I wonder if this could/should be tied up with some API wrapping [RegisterApplicationRestart()](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-registerapplicationrestart) which I think would be also useful to have in wx, wouldn't it?

marekr:
Register application restart is also something we handle in kicad. Basically to offer post-update restart when used with the session restoration.

But in kicad's case, implemented the relaunchcommand with a "-n"  arg specifically because a user can trigger it while kicad is already open with a project. If you launch kicad again, it would otherwise try to reopen that exist same project. The -n arg in my case tells it to start blank.

But in the register application restart case, I would want it to open the previous project.

marekr:
Also, there's another behavior change by app model id. That I am also fixing in kicad.

The MRU file list in windows for shortcuts in your start menu that point to your application stop working.

Fixing it requires implementing SHAddToRecentDocs to add recent items to the shell MRU using app model IDs.

https://gitlab.com/kicad/code/kicad/-/commit/743988bd57b2f74097ed09a5c46899c10522e441#83968e0062c0e041190aca75cc1be793ab662965_484_485

And using setting the app model id within the shortcut itself which is actually not trivial, Microsoft only intended for people to do this via MSI installs and doing it through something like nsis is a little bit of win32 api adventure.

vadz:
> Also, there's another behavior change by app model id. That I am also fixing in kicad.
> 
> The MRU file list in windows for shortcuts in your start menu that point to your application stop working.

This looks like something `wxFileHistory` could do, I think. Would it be useful for you if it took care of this?

For the restart problem, unless you'd be interested in discussing restart API more in depth (in which case we need to open a new issue and do it there), I think we'll just have to settle for documenting the limitation for now as I can't say it's obvious to me what should wx do here.

PR #25549, opened 2025-06-19
Use UTF-8 with MSVC and for string literals

vadz:
Replaces #25373.


MaartenBent:
AppVeyors Visual Studio 2015 seems to fail on tests that are missing `wxString::FromUTF8()`.

Kumazuma:
It looks good to me.

vadz:
> AppVeyors Visual Studio 2015 seems to fail on tests that are missing `wxString::FromUTF8()`.

It looks like it doesn't handle Unicode characters in the wide strings, I've tried to work around it just for it. But maybe we should just drop support for it, anecdotally it seems that some people still use MSVS 2017 but I don't know anybody still using 2015...

PR #25554, opened 2025-06-22
Link libwebp when using MSVC-specific setup.h

PBfordev:
Recently added libwebp was not added to the libraries MSVC automatically links with when adding "$(WXWIN)\include\msvc" to the include paths.

This should have been part of ad6f05a (Add support for WebP image format, including animations, 2025-05-11).


vadz:
Good catch, thanks! I'll merge it soon and will also add documentation of this `wxNO_` symbol to `docs/doxygen/mainpages/const_cpp.h`.

sins921 (approved):

Issue #25556, opened 2025-06-23
Store and retreive string in PNG


G-e-n-e-v-e-n-s-i-S:
Is it possible to store some string data inside a wxImage, save it to disk as a PNG image, then later load it and retrieve the string?

I tried using wxImage::SetOption() and wxImage::GetOption() but these don't seem to persist upon writing to disk.

I also haven't found any function to access the iTXt, tEXt, or zTXt chunks.

I'm open to any kind of ugly/platform specific hack (except storing it in the filename).

Thanks!

vadz:
No, wxImage doesn't provide support for working with PNG-specific chunks. In principle, it should be possible to handle this using the existing options mechanism by defining more PNG-specific options corresponding to such chunks, but it risks being really ugly. If you need full support of the PNG format, using libpng directly is probably a better idea.

G-e-n-e-v-e-n-s-i-S:
I don't need full PNG support. I really just need the bare minimum functionality of storing a string inside the file when saving it on disk, and retrieving it on load.

I'm trying to avoid adding yet another library dependency to my project just for a single function. You're saying it can be done with Options in principle? Would I need to override the wxImage::SaveFile() function? I'm willing to give it a go if you can start me in the right direction.

Thanks for your reply.

vadz:
You already have an implicit dependency on libpng, as it's used by `wxImage`, so using it shouldn't be a big change.

Anyhow, you can't do this outside of wx. What I meant by "it can be done with options in principle" is that it should be relatively simple to add some `wxIMAGE_OPTION_PNG_tEXt` and handle it in the code by loading it from/saving it to `tEXt` chunk. This still needs to actually be done.

G-e-n-e-v-e-n-s-i-S:
Solved by https://github.com/wxWidgets/wxWidgets/pull/25565

PR #792, opened 2018-04-28
Custom meta-data for PNG

warniaha:
Added ability to store custom meta-data within PNG images.  samples/image updated to edit custom meta-data.

PR #25565, opened 2025-06-25
Store and retrieve PNG Description text chunk

G-e-n-e-v-e-n-s-i-S:
This is an attempt at implementing the suggestion in https://github.com/wxWidgets/wxWidgets/issues/25556

I'm a bit out of my depth here, so I'm hoping for some helpful suggestions.

This is inspired by the handling of `wxIMAGE_OPTION_GIF_COMMENT`, as well as by this PR: https://github.com/wxWidgets/wxWidgets/pull/792 , while trying to keep everything to the bare minimum.

### DIFF
--- a/include/wx/imagpng.h
+++ b/include/wx/imagpng.h
@@ -20,7 +20,8 @@
 #include "wx/image.h"
 #include "wx/versioninfo.h"
 
-#define wxIMAGE_OPTION_PNG_DESCRIPTION wxT("PngDescription")
+#define wxIMAGE_OPTION_PNG_DESCRIPTION     wxT("PngDescription")
+#define wxIMAGE_OPTION_PNG_DESCRIPTION_KEY wxString("Description")
 
 #define wxIMAGE_OPTION_PNG_FORMAT    wxT("PngFormat")
 #define wxIMAGE_OPTION_PNG_BITDEPTH  wxT("PngBitDepth")
--- a/include/wx/imagpng.h
+++ b/include/wx/imagpng.h
@@ -21,7 +21,6 @@
 #include "wx/versioninfo.h"
 
 #define wxIMAGE_OPTION_PNG_DESCRIPTION     wxT("PngDescription")
-#define wxIMAGE_OPTION_PNG_DESCRIPTION_KEY wxString("Description")
 
 #define wxIMAGE_OPTION_PNG_FORMAT    wxT("PngFormat")
 #define wxIMAGE_OPTION_PNG_BITDEPTH  wxT("PngBitDepth")
--- a/include/wx/imagpng.h
+++ b/include/wx/imagpng.h
@@ -20,6 +20,8 @@
 #include "wx/image.h"
 #include "wx/versioninfo.h"
 
+#define wxIMAGE_OPTION_PNG_DESCRIPTION wxT("PngDescription")
+
 #define wxIMAGE_OPTION_PNG_FORMAT    wxT("PngFormat")
 #define wxIMAGE_OPTION_PNG_BITDEPTH  wxT("PngBitDepth")
 #define wxIMAGE_OPTION_PNG_FILTER    wxT("PngF")
--- a/include/wx/imagpng.h
+++ b/include/wx/imagpng.h
@@ -20,8 +20,6 @@
 #include "wx/image.h"
 #include "wx/versioninfo.h"
 
-#define wxIMAGE_OPTION_PNG_DESCRIPTION     wxT("PngDescription")
-
 #define wxIMAGE_OPTION_PNG_FORMAT    wxT("PngFormat")
 #define wxIMAGE_OPTION_PNG_BITDEPTH  wxT("PngBitDepth")
 #define wxIMAGE_OPTION_PNG_FILTER    wxT("PngF")
@@ -29,6 +27,7 @@
 #define wxIMAGE_OPTION_PNG_COMPRESSION_MEM_LEVEL   wxT("PngZM")
 #define wxIMAGE_OPTION_PNG_COMPRESSION_STRATEGY    wxT("PngZS")
 #define wxIMAGE_OPTION_PNG_COMPRESSION_BUFFER_SIZE wxT("PngZB")
+#define wxIMAGE_OPTION_PNG_DESCRIPTION             wxT("PngDescription")
 
 enum
 {
--- a/interface/wx/image.h
+++ b/interface/wx/image.h
@@ -148,6 +148,18 @@ enum wxImagePNGType
 #define wxIMAGE_OPTION_PNG_COMPRESSION_MEM_LEVEL        wxString("PngZM")
 #define wxIMAGE_OPTION_PNG_COMPRESSION_STRATEGY         wxString("PngZS")
 #define wxIMAGE_OPTION_PNG_COMPRESSION_BUFFER_SIZE      wxString("PngZB")
+
+/**
+Key used for setting an image option that, upon saving to a PNG file, will be
+converted to an iTXt chunk with the key: "Description", and stored in the file.
+
+Uncompressed tEXt and iTXt chunks with the key: "Description" are automatically
+retrieved upon loading of a PNG file, and stored in this image option. If
+multiple chunks with this key are present, only the first is retrieved.
+
+@since 3.3.1
+@see wxImage::GetOption()
+*/
 #define wxIMAGE_OPTION_PNG_DESCRIPTION                  wxString("PngDescription")
 
 #define wxIMAGE_OPTION_TIFF_BITSPERSAMPLE               wxString("BitsPerSample")
--- a/interface/wx/image.h
+++ b/interface/wx/image.h
@@ -1324,8 +1324,14 @@ public:
             a GIF it will not be repeated in other frames.
 
         Options specific to wxPNGHandler:
-        @li @c wxIMAGE_OPTION_PNG_DESCRIPTION: The Description text chunk that
-            is read from or written to the PNG file.
+        @li @c wxIMAGE_OPTION_PNG_DESCRIPTION: The contents of this option will
+            be converted to an uncompressed iTXt chunk with the key: "Description",
+            and written to the PNG file upon saving.
+            Contents of uncompressed tXTt and iTXt chunks with the key: "Description"
+            are also automatically retrieved upon loading a PNG file, and stored in
+            this option. If multiple chunks with this key are present, only the first
+            is retrieved.
+            @since 3.3.1
 
         @param name
             The name of the option, case-insensitive.
--- a/interface/wx/image.h
+++ b/interface/wx/image.h
@@ -148,18 +148,6 @@ enum wxImagePNGType
 #define wxIMAGE_OPTION_PNG_COMPRESSION_MEM_LEVEL        wxString("PngZM")
 #define wxIMAGE_OPTION_PNG_COMPRESSION_STRATEGY         wxString("PngZS")
 #define wxIMAGE_OPTION_PNG_COMPRESSION_BUFFER_SIZE      wxString("PngZB")
-
-/**
-Key used for setting an image option that, upon saving to a PNG file, will be
-converted to an iTXt chunk with the key: "Description", and stored in the file.
-
-Uncompressed tEXt and iTXt chunks with the key: "Description" are automatically
-retrieved upon loading of a PNG file, and stored in this image option. If
-multiple chunks with this key are present, only the first is retrieved.
-
-@since 3.3.1
-@see wxImage::GetOption()
-*/
 #define wxIMAGE_OPTION_PNG_DESCRIPTION                  wxString("PngDescription")
 
 #define wxIMAGE_OPTION_TIFF_BITSPERSAMPLE               wxString("BitsPerSample")
--- a/interface/wx/image.h
+++ b/interface/wx/image.h
@@ -1339,7 +1339,7 @@ public:
         @li @c wxIMAGE_OPTION_PNG_DESCRIPTION: The contents of this option will
             be converted to an uncompressed iTXt chunk with the key "Description",
             and written to the PNG file upon saving.
-            Contents of uncompressed tXTt and iTXt chunks with the key: "Description"
+            Contents of uncompressed tXTt and iTXt chunks with the key "Description"
             are also automatically retrieved upon loading a PNG file, and stored in
             this option. If multiple chunks with this key are present, only the first
             is retrieved.
--- a/interface/wx/image.h
+++ b/interface/wx/image.h
@@ -148,6 +148,7 @@ enum wxImagePNGType
 #define wxIMAGE_OPTION_PNG_COMPRESSION_MEM_LEVEL        wxString("PngZM")
 #define wxIMAGE_OPTION_PNG_COMPRESSION_STRATEGY         wxString("PngZS")
 #define wxIMAGE_OPTION_PNG_COMPRESSION_BUFFER_SIZE      wxString("PngZB")
+#define wxIMAGE_OPTION_PNG_DESCRIPTION                  wxString("PngDescription")
 
 #define wxIMAGE_OPTION_TIFF_BITSPERSAMPLE               wxString("BitsPerSample")
 #define wxIMAGE_OPTION_TIFF_SAMPLESPERPIXEL             wxString("SamplesPerPixel")
@@ -1321,6 +1322,10 @@ public:
             or written to the GIF file. In an animated GIF each frame can have
             its own comment. If there is only a comment in the first frame of
             a GIF it will not be repeated in other frames.
+
+        Options specific to wxPNGHandler:
+        @li @c wxIMAGE_OPTION_PNG_DESCRIPTION: The Description text chunk that
+            is read from or written to the PNG file.
 
         @param name
             The name of the option, case-insensitive.
--- a/interface/wx/image.h
+++ b/interface/wx/image.h
@@ -1337,7 +1337,7 @@ public:
 
         Options specific to wxPNGHandler:
         @li @c wxIMAGE_OPTION_PNG_DESCRIPTION: The contents of this option will
-            be converted to an uncompressed iTXt chunk with the key: "Description",
+            be converted to an uncompressed iTXt chunk with the key "Description",
             and written to the PNG file upon saving.
             Contents of uncompressed tXTt and iTXt chunks with the key: "Description"
             are also automatically retrieved upon loading a PNG file, and stored in
--- a/interface/wx/image.h
+++ b/interface/wx/image.h
@@ -1324,13 +1324,13 @@ public:
             a GIF it will not be repeated in other frames.
 
         Options specific to wxPNGHandler:
-        @li @c wxIMAGE_OPTION_PNG_DESCRIPTION: The contents of this option will
-            be converted to an uncompressed iTXt chunk with the key "Description",
-            and written to the PNG file upon saving.
-            Contents of uncompressed tXTt and iTXt chunks with the key "Description"
-            are also automatically retrieved upon loading a PNG file, and stored in
-            this option. If multiple chunks with this key are present, only the first
-            is retrieved.
+        @li @c wxIMAGE_OPTION_PNG_DESCRIPTION: The contents of this option
+            will be converted to an uncompressed iTXt chunk with the
+            key: "Description", and written to the PNG file upon saving.
+            Contents of tXTt and iTXt chunks with the key: "Description" are
+            also automatically retrieved upon loading a PNG file, and stored
+            in this option. If multiple chunks with this key are present,
+            only one is retrieved.
             @since 3.3.1
 
         @param name
--- a/interface/wx/imagpng.h
+++ b/interface/wx/imagpng.h
@@ -5,6 +5,16 @@
 // Licence:     wxWindows licence
 ////////////////////////////////////////////////////////////////////////////
 
+/**
+Key used for setting an image option that, upon saving to a PNG file, will be
+converted to an iTXt chunk with the key: "Description", and stored in the file.
+
+Uncompressed tEXt and iTXt chunks with the key: "Description" are automatically
+retrieved upon loading of a PNG file, and stored in this image option. If
+multiple chunks with this key are present, only the first is retrieved.
+
+@since 3.3.1
+*/
 #define wxIMAGE_OPTION_PNG_DESCRIPTION              wxT("PngDescription")
 
 #define wxIMAGE_OPTION_PNG_FORMAT                   wxT("PngFormat")
--- a/interface/wx/imagpng.h
+++ b/interface/wx/imagpng.h
@@ -5,6 +5,14 @@
 // Licence:     wxWindows licence
 ////////////////////////////////////////////////////////////////////////////
 
+#define wxIMAGE_OPTION_PNG_FORMAT                   wxT("PngFormat")
+#define wxIMAGE_OPTION_PNG_BITDEPTH                 wxT("PngBitDepth")
+#define wxIMAGE_OPTION_PNG_FILTER                   wxT("PngF")
+#define wxIMAGE_OPTION_PNG_COMPRESSION_LEVEL        wxT("PngZL")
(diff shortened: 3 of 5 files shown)


interface/wx/imagpng.h
```diff
@@ -5,6 +5,8 @@
 // Licence:     wxWindows licence
 ////////////////////////////////////////////////////////////////////////////
 
+#define wxIMAGE_OPTION_PNG_DESCRIPTION              wxT("PngDescription")
```
vadz: This would need to be really documented to explain how it works because, unlike most of the other options, it's not really obvious.

Please also use `@since 3.3.1` in the Doxygen comment when you document it.

src/common/imagpng.cpp
```diff
@@ -342,6 +342,23 @@ wxPNGImageData::DoLoadPNGFile(wxImage* image, wxPNGInfoStruct& wxinfo)
         return;
 
     png_read_image( png_ptr, lines );
+
+    // load "Description" text chunk
+    png_textp text_ptr;
+    int num_comments = png_get_text(png_ptr, info_ptr, &text_ptr, NULL);
```
vadz:
```suggestion
    int num_comments = png_get_text(png_ptr, info_ptr, &text_ptr, nullptr);
```

src/common/imagpng.cpp
```diff
@@ -342,6 +342,23 @@ wxPNGImageData::DoLoadPNGFile(wxImage* image, wxPNGInfoStruct& wxinfo)
...
+    for (int i = 0; i < num_comments; ++i)
+    {
+        if (text_ptr[i].compression != PNG_TEXT_COMPRESSION_NONE) continue; // ignoring compressed ones cause I don't know how to handle them
```
vadz: You should check for `PNG_ITXT_COMPRESSION_NONE` here too, this can be handled easily.

src/common/imagpng.cpp
```diff
@@ -342,6 +342,23 @@ wxPNGImageData::DoLoadPNGFile(wxImage* image, wxPNGInfoStruct& wxinfo)
...
+
+        // key and text can either be Latin-1 or UTF-8, and are guaranteed to be zero-terminated C strings, so I think From8BitData is appropriate here?
```
vadz: Please see section 4.2.3 of [PNG chunks documentation](https://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html), it explains that `tEXt` is always Latin-1 while `iTXt` uses UTF-8.

src/common/imagpng.cpp
```diff
@@ -342,6 +342,23 @@ wxPNGImageData::DoLoadPNGFile(wxImage* image, wxPNGInfoStruct& wxinfo)
...
+        wxString key = wxString::From8BitData(text_ptr[i].key);
+        if (key == _("Description")) { // "Description" is listed as a predefined chunk by the png standard. is hardcoding it like this fine?
```
vadz: Hard coding it is fine but you should define a constant for it. Using `_()` is definitely not, you do _not_ want to translate it here, it should be used exactly as it is.

src/common/imagpng.cpp
```diff
@@ -342,6 +342,23 @@ wxPNGImageData::DoLoadPNGFile(wxImage* image, wxPNGInfoStruct& wxinfo)
...
+            wxString text = wxString::From8BitData(text_ptr[i].text);
+            image->SetOption(wxIMAGE_OPTION_PNG_DESCRIPTION, text);
+            break;
```
vadz: It's not really clear whether we can have both `tEXt` and `iTXt` chunks with the same name. If it's possible, we shouldn't break after the first one.

src/common/imagpng.cpp
```diff
@@ -737,6 +754,19 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
         png_set_pHYs( png_ptr, info_ptr, resX, resY, PNG_RESOLUTION_METER );
 
     png_set_sBIT( png_ptr, info_ptr, &sig_bit );
+
+    // save "Description" text chunk
+    if (image->HasOption(wxIMAGE_OPTION_PNG_DESCRIPTION)) {
```
vadz: Minor, but please use wx style, i.e.:

```suggestion
    if (image->HasOption(wxIMAGE_OPTION_PNG_DESCRIPTION))
    {
```

src/common/imagpng.cpp
```diff
@@ -737,6 +754,19 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
...
+        wxString value = image->GetOption(wxIMAGE_OPTION_PNG_DESCRIPTION);
+
+        wxSharedPtr<png_text> text_ptr(new png_text[1]); // is wxSharedPtr appropriate here?
```
vadz: `wxSharedPtr` is absolutely not needed here, you're not sharing the ownership with anybody, why would you use it.

src/common/imagpng.cpp
```diff
@@ -737,6 +754,19 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
...
+        text_ptr.get()[0].compression = PNG_TEXT_COMPRESSION_NONE;
+        text_ptr.get()[0].key =  const_cast<char*>((const char*)_("Description").mb_str(wxConvUTF8)); // again a hard coded string, is this fine?
```
vadz: The same constant as above should be reused, also without `_()`.

src/common/imagpng.cpp
```diff
@@ -737,6 +754,19 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
         png_set_pHYs( png_ptr, info_ptr, resX, resY, PNG_RESOLUTION_METER );
 
     png_set_sBIT( png_ptr, info_ptr, &sig_bit );
+
+    // save "Description" text chunk
+    if (image->HasOption(wxIMAGE_OPTION_PNG_DESCRIPTION)) {
+        wxString value = image->GetOption(wxIMAGE_OPTION_PNG_DESCRIPTION);
+
+        wxSharedPtr<png_text> text_ptr(new png_text[1]); // is wxSharedPtr appropriate here?
+        text_ptr.get()[0].compression = PNG_TEXT_COMPRESSION_NONE;
```
vadz: This is generally wrong because you can't use `TEXT` for UTF-8, only for Latin-1. If you wanted to get fancy, you could try encoding the value of the option as Latin-1 and save it as `TEXT` if it can be represented in it, but it's simpler and might be actually better by now to always use `ITXT` and UTF-8.

src/common/imagpng.cpp
```diff
@@ -737,6 +754,19 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
...
+        text_ptr.get()[0].key =  const_cast<char*>((const char*)_("Description").mb_str(wxConvUTF8)); // again a hard coded string, is this fine?
+        text_ptr.get()[0].text = const_cast<char*>((const char*)value           .mb_str(wxConvUTF8));
```
vadz:
```suggestion
        text_ptr.get()[0].text = const_cast<char*>(value.utf8_str().data());
```

tests/image/image.cpp
```diff
@@ -1069,6 +1069,32 @@ TEST_CASE_METHOD(ImageHandlersInit, "wxImage::SavePNG", "[image]")
...
+    wxMemoryInputStream memIn(memOut);
+    REQUIRE(image.LoadFile(memIn));
+
+    CHECK(image.GetOption(wxIMAGE_OPTION_PNG_DESCRIPTION) == description);
+}
+
+TEST_CASE_METHOD(ImageHandlersInit, "wxImage::PNGDescription", "[image]")
+{
+    // Test writing a description and reading it back.
+    TestPNGDescription("Providing the PNG a pneumatic puma as a present");
```
vadz, line 1089: You should test with non-ASCII strings too.

vadz (changes requested):
Thanks for the PR, but there seems to be some confusion about how PNG chunks work here. You really need to handle `tEXt` and `iTXt` chunks differently, both when reading and when writing them. Hopefully it shouldn't be very difficult to do.

src/common/imagpng.cpp
G-e-n-e-v-e-n-s-i-S: It's apparently possible. Can images also have multiple options with the same key? If so how would one iterate over them?

vadz: There is an ugly way: concatenate all values together using `NUL` (i.e. `\0`) when reading them and then split them on `\0` when writing.

A better way would be add some `SetMultiValuedOption(const wxString& name, const std::vector<wxString>& values)` and the corresponding accessor, but this would require more work.

And it still wouldn't provide full access to the file contents, as you wouldn't be able to get all the names of the available options. This would need adding some more flexible "metadata" API, which should be ideally cover more than just PNG format, and would probably be quite useful to have, but would need even more thought/time to implement.

Of course, it's also possible to just leave it as is and document it as the limitation of the current approach. Whether it's a good idea or not depends on how often does this happen in the wild, if it's something rare we probably could get away with it...

G-e-n-e-v-e-n-s-i-S: I indeed believe the cases where an image has multiple meaningfully different descriptions are rare. I will leave it as is for now, and document it.

In the future, when I level up my skills, I may come back to this and implement that flexible "metadata" API you speak of.

G-e-n-e-v-e-n-s-i-S:
I tried to address all your comments. I remain of course ready to tackle new ones.

Thank you for your availability and patience!

interface/wx/image.h
```diff
@@ -149,6 +149,19 @@ enum wxImagePNGType
...
+converted to an iTXt chunk with the key: "Description", and stored in the file.
+
+Uncompressed tEXt and iTXt chunks with the key: "Description" are automatically
+retrieved upon loading of a PNG file, and stored in this image option. If
+multiple chunks with this key are present, only the first is retrieved.
+
+@since 3.3.1
+@see wxImage::GetOption()
+*/
+#define wxIMAGE_OPTION_PNG_DESCRIPTION                  wxString("PngDescription")
```
PBfordev, line 151: I would remove the whole comment block here, other options don't have it here either.

interface/wx/image.h
```diff
@@ -1322,6 +1335,16 @@ class wxImage : public wxObject
             its own comment. If there is only a comment in the first frame of
             a GIF it will not be repeated in other frames.
 
+        Options specific to wxPNGHandler:
+        @li @c wxIMAGE_OPTION_PNG_DESCRIPTION: The contents of this option will
+            be converted to an uncompressed iTXt chunk with the key: "Description",
```
PBfordev:
```suggestion
            be converted to an uncompressed iTXt chunk with the key "Description",
```

interface/wx/image.h
```diff
@@ -1322,6 +1335,16 @@ class wxImage : public wxObject
...
+            and written to the PNG file upon saving.
+            Contents of uncompressed tXTt and iTXt chunks with the key: "Description"
```
PBfordev:
```suggestion
            Contents of uncompressed tXTt and iTXt chunks with the key "Description"
```

interface/wx/imagpng.h
```diff
@@ -13,6 +13,19 @@
...
+Key used for setting an image option that, upon saving to a PNG file, will be
+converted to an iTXt chunk with the key: "Description", and stored in the file.
+
+Uncompressed tEXt and iTXt chunks with the key: "Description" are automatically
+retrieved upon loading of a PNG file, and stored in this image option. If
+multiple chunks with this key are present, only the first is retrieved.
+
+@since 3.3.1
+@see wxImage::GetOption()
+*/
```
PBfordev: I would remove the comment block here as well.

G-e-n-e-v-e-n-s-i-S: This was added in response to this change request: https://github.com/wxWidgets/wxWidgets/pull/25565#discussion_r2166708435

PBfordev: But does it make sense to explain it exactly the same in three different places?

interface/wx/image.h
vadz: I agree, it's not necessary to duplicate the docs here, it might have made sense to document the options themselves too, but we didn't do it like this and only document them in `GetOption[Int]()` for all the other ones, so let's be consistent with them.

include/wx/imagpng.h
```diff
@@ -20,6 +20,9 @@
 #include "wx/image.h"
 #include "wx/versioninfo.h"
 
+#define wxIMAGE_OPTION_PNG_DESCRIPTION     wxT("PngDescription")
+#define wxIMAGE_OPTION_PNG_DESCRIPTION_KEY wxString("Description")
```
vadz: We don't need to make this constant public, this is only used in the implementation and should be defined in `imagpng.cpp`.

interface/wx/imagpng.h
vadz: As this shouldn't be public at all, we shouldn't document it either.

src/common/imagpng.cpp
```diff
@@ -737,6 +767,20 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
         png_set_pHYs( png_ptr, info_ptr, resX, resY, PNG_RESOLUTION_METER );
 
     png_set_sBIT( png_ptr, info_ptr, &sig_bit );
+
+    // save "Description" text chunk
+    if (image->HasOption(wxIMAGE_OPTION_PNG_DESCRIPTION))
+    {
+        wxString text = image->GetOption(wxIMAGE_OPTION_PNG_DESCRIPTION);
```
vadz: Very minor, but we should avoid copying a potentially long string without any good reason:
```suggestion
        const wxString& text = image->GetOption(wxIMAGE_OPTION_PNG_DESCRIPTION);
```

src/common/imagpng.cpp
```diff
@@ -737,6 +767,20 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
...
+
+        wxScopedPtr<png_text> text_ptr(new png_text[1]);
```
vadz: Sorry, I should have been more precise in my comment on the previous version: we don't need any pointers at all here, shared or scoped. This should be just

```suggestion
        png_text text;
```

and then use `text` instead of `text_ptr.get()[0]` below and `&text` instead of `text_ptr.get()`.

src/common/imagpng.cpp
```diff
@@ -342,6 +343,35 @@ wxPNGImageData::DoLoadPNGFile(wxImage* image, wxPNGInfoStruct& wxinfo)
         return;
 
     png_read_image( png_ptr, lines );
+
+    // load "Description" text chunk
+    png_textp text_ptr;
+    int num_comments = png_get_text( png_ptr, info_ptr, &text_ptr, nullptr );
```
vadz: Extremely minor, but it can be nice to show that this doesn't change

```suggestion
    const int num_comments = png_get_text( png_ptr, info_ptr, &text_ptr, nullptr );
```

src/common/imagpng.cpp
```diff
@@ -342,6 +343,35 @@ wxPNGImageData::DoLoadPNGFile(wxImage* image, wxPNGInfoStruct& wxinfo)
...
+            if (key == wxIMAGE_OPTION_PNG_DESCRIPTION_KEY)
+            {
+                wxString text = wxString::From8BitData(text_ptr[i].text);
+                image->SetOption(wxIMAGE_OPTION_PNG_DESCRIPTION, text);
+                break;
+            }
+        }
+        else if (text_ptr[i].compression == PNG_ITXT_COMPRESSION_NONE) // UTF-8 encoding
+        {
+            wxString key = wxString::FromUTF8(text_ptr[i].key);
```
vadz: Sorry, this is wrong, PNG specification says:

> The keyword must be at least one character and less than 80 characters long. Keywords are always interpreted according to the ISO/IEC 8859-1 (Latin-1) character set [[ISO/IEC-8859-1](https://www.libpng.org/pub/png/spec/1.2/PNG-References.html#B.I-8859-1)]. They must contain only printable Latin-1 characters and spaces; that is, only character codes 32-126 and 161-255 decimal are allowed.

So the key should always be passed to `From8BitData()`.

src/common/imagpng.cpp
```diff
@@ -342,6 +343,35 @@ wxPNGImageData::DoLoadPNGFile(wxImage* image, wxPNGInfoStruct& wxinfo)
         return;
 
     png_read_image( png_ptr, lines );
+
+    // load "Description" text chunk
+    png_textp text_ptr;
+    int num_comments = png_get_text( png_ptr, info_ptr, &text_ptr, nullptr );
+    for (int i = 0; i < num_comments; ++i)
+    {
```
vadz, line 352: The code looks almost good now but a bit verbose/redundant. I would write it like this:

```cpp
const wxString& key = wxString::From8BitData(text_ptr[i].key);
if ( key == wxIMAGE_OPTION_PNG_DESCRIPTION_KEY )
{
    wxString description;
    switch ( text_ptr[i].compression )
    {
        case PNG_TEXT_COMPRESSION_NONE:
            // tEXt chunk: uses Latin-1 encoding.
            description = wxString::From8BitData(text_ptr[i].text);
            break;

        case PNG_ITXT_COMPRESSION_NONE:
            // iTXt chunk: uses UTF-8 encoding.
            description = wxString::FromUTF8(text_ptr[i].text);
            break;

        case PNG_TEXT_COMPRESSION_zTXt:
        case PNG_ITXT_COMPRESSION_zTXt:
            // Not supported currently, but should be.
            break;

        default:
            // Invalid type, should we report it? Probably not worth it.
            break;
    }

    if ( !description.empty() )
        image->SetOption(wxIMAGE_OPTION_PNG_DESCRIPTION, description);
}
```

vadz (commented):
Thanks, it looks mostly good now, but the CI failures need to be dealt with, of course. Have you already tried to look into what is causing them?

src/common/imagpng.cpp
```diff
@@ -783,11 +783,22 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
         const wxString& description = image->GetOption(wxIMAGE_OPTION_PNG_DESCRIPTION);
 
         png_text text;
+        text.key = const_cast<char*>(wxIMAGE_OPTION_PNG_DESCRIPTION_KEY.ToAscii().data());
+        text.lang = NULL;
```
PBfordev: I think we are supposed to use `nullptr` instead of `NULL` everywhere...

tests/image/image.cpp
```diff
@@ -1088,8 +1088,10 @@ TEST_CASE_METHOD(ImageHandlersInit, "wxImage::PNGDescription", "[image]")
     // Test writing a description and reading it back.
     TestPNGDescription("Providing the PNG a pneumatic puma as a present");
 
+#ifdef PNG_iTXt_SUPPORTED
```
PBfordev: Is the libpng define available here?

G-e-n-e-v-e-n-s-i-S: It appears it is not, how would I exclude a test based on it?

src/common/imagpng.cpp
G-e-n-e-v-e-n-s-i-S: I changed the code according to your rewrite. It seems pnglib actually handles the decompression, so we don't have to do anything special.

G-e-n-e-v-e-n-s-i-S:
I had to modify this line:
https://github.com/wxWidgets/wxWidgets/blob/008dc91b93ac863df56f23f38af1847c1d308458/src/common/imagpng.cpp#L194

...to allow it to handle UTF8. Due to my lack of familiarity with the codebase, I do not grasp the full consequences of this change.

src/common/imagpng.cpp
```diff
@@ -342,6 +344,42 @@ wxPNGImageData::DoLoadPNGFile(wxImage* image, wxPNGInfoStruct& wxinfo)
         return;
 
     png_read_image( png_ptr, lines );
+
+    // load "Description" text chunk
+#if defined(PNG_READ_iTXt_SUPPORTED) || defined(PNG_READ_tEXt_SUPPORTED)
```
vadz: Sorry, I don't understand the purpose of this check. AFAIK `PNG_READ_tEXt_SUPPORTED` is always defined and `PNG_READ_iTXt_SUPPORTED` is available since libpng v1.0.5h from 1999, so why do we need to test for them?

Also, it definitely doesn't make sense to test for _either_ of them being supported and then using _both_ below, the test must use `&&` if it should be there at all.

src/common/imagpng.cpp
```diff
@@ -737,6 +775,34 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
...
+        text.lang_key = nullptr;
+
+#ifdef PNG_WRITE_iTXt_SUPPORTED
+        text.compression = PNG_ITXT_COMPRESSION_NONE;
+        text.text = const_cast<char*>(description.utf8_str().data());
+        text.text_length = 0;
+        text.itxt_length = strlen(text.text);
+#else
+        text.compression = PNG_TEXT_COMPRESSION_NONE;
+        text.text = const_cast<char*>(description.mb_str(wxConvISO8859_1).data());
```
vadz: This is inconsistent with the use of `From8BitData()` above, either we should use `To8BitData()` here or use `wxConvISO8859_1` above as well.

src/common/imagpng.cpp
```diff
@@ -737,6 +775,34 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
         png_set_pHYs( png_ptr, info_ptr, resX, resY, PNG_RESOLUTION_METER );
 
     png_set_sBIT( png_ptr, info_ptr, &sig_bit );
+
+    // save "Description" text chunk
+#if defined(PNG_WRITE_iTXt_SUPPORTED) || defined(PNG_WRITE_tEXt_SUPPORTED)
```
vadz: Those also seem to be defined since very, very long time.

src/common/imagpng.cpp
```diff
@@ -737,6 +775,34 @@ bool wxPNGHandler::SaveFile( wxImage *image, wxOutputStream& stream, bool verbos
...
+        const wxString& description = image->GetOption(wxIMAGE_OPTION_PNG_DESCRIPTION);
+
+        png_text text;
+        text.key = const_cast<char*>(wxIMAGE_OPTION_PNG_DESCRIPTION_KEY);
+        text.lang = nullptr;
+        text.lang_key = nullptr;
+
+#ifdef PNG_WRITE_iTXt_SUPPORTED
+        text.compression = PNG_ITXT_COMPRESSION_NONE;
+        text.text = const_cast<char*>(description.utf8_str().data());
```
vadz: Sorry, rereading this code again I realize that this is a critical problem as the pointer to the buffer data becomes dangling when the temporary buffer returned by `utf8_str()` is destroyed. This must be changed to

```cpp
const auto& buf = description.utf8_str();
text.text = const_cast<char*>(buf.data());
text.itxt_length = buf.length();
``` 

(the last line is just a micro-optimization avoiding an extra call to `strlen()`, but we *must* keep the buffer alive for long enough)

vadz (commented):
Sorry, I don't understand the preprocessor checks, could you please explain why are they needed? TIA!

src/common/imagpng.cpp
G-e-n-e-v-e-n-s-i-S: I have no idea why these are needed either. I am copying patterns I find elsewhere in the codebase without understanding them.

I'll remove them if you deem them useless.

vadz (approved):
Looks good now, thanks! I'll merge it soon.

And, of course, if you see something you don't understand, please ask, but don't copy it without understanding what it does (although, to be honest, I have no idea where did you copy the checks that were removed now from, I don't see anything like this in the existing code...).

G-e-n-e-v-e-n-s-i-S:
Nice! Thanks again vadz and PBfordev for your patience and guidance.

PR #25570, opened 2025-06-25
Turn on NSScrollView borders for listbox, dataview and textctrl on macOS #25560

### LINKED ISSUE #25560: wxDataViewCtrl on macOS defaults to no border, doesn't draw correct border style
#### Bug description:

wxDataViewCtrl on macOS doesn't draw a border by default. That is opposite of Win/GTK3 which does draw a border by default. Furthermore, if you pass in `wxBORDER_DEFAULT` it doesn't draw the correct native control border as expected.

Expect border to draw by default and be the native style. Observed no border is drawn.

The bug stems from [this line](https://github.com/wxWidgets/wxWidgets/blob/23762d0c65b7cd013e4cf8b87869c882ac5a8d97/src/osx/cocoa/dataview.mm#L1989). It should be looking at the style value coming in and turning on the scrollview border.

Furthermore, if you set `wxBORDER_DEFAULT` as the border it should turn on the border of the scrollview, not draw a custom border that isn't native which it is currently doing.

#### To Reproduce:

Create a data view like so:
`tree_view = new wxDataViewCtrl(this, wxID_ANY, wxDefaultPosition, wxDefaultSize);`

- wxWidgets version you use: 3.2.7
- wxWidgets port you use: wxMacOS
- OS and its version: macOS 15.5

willco007:
Resolves #25560 by enabling the scrollview borders by default on macOS.

### DIFF
--- a/src/osx/cocoa/dataview.mm
+++ b/src/osx/cocoa/dataview.mm
@@ -1985,8 +1985,13 @@ wxCocoaDataViewControl::wxCocoaDataViewControl(wxWindow* peer,
 {
     // initialize scrollview (the outline view is part of a scrollview):
     NSScrollView* scrollview = (NSScrollView*) GetWXWidget();
+    wxBorder border = (wxBorder)(style & wxBORDER_MASK);
 
-    [scrollview setBorderType:NSNoBorder];
+    if ( (border == wxBORDER_DEFAULT) || (border == wxBORDER_THEME) )
+        [scrollview setBorderType:NSBezelBorder];
+    else
+        [scrollview setBorderType:NSNoBorder];
+
     [scrollview setHasVerticalScroller:YES];
     [scrollview setHasHorizontalScroller:YES];
     [scrollview setAutohidesScrollers:YES];
--- a/src/osx/cocoa/dataview.mm
+++ b/src/osx/cocoa/dataview.mm
@@ -1985,17 +1985,13 @@ wxCocoaDataViewControl::wxCocoaDataViewControl(wxWindow* peer,
 {
     // initialize scrollview (the outline view is part of a scrollview):
     NSScrollView* scrollview = (NSScrollView*) GetWXWidget();
-    wxBorder border = (wxBorder)(style & wxBORDER_MASK);
-
-    if ( (border == wxBORDER_DEFAULT) || (border == wxBORDER_THEME) )
-        [scrollview setBorderType:NSBezelBorder];
-    else
-        [scrollview setBorderType:NSNoBorder];
 
     [scrollview setHasVerticalScroller:YES];
     [scrollview setHasHorizontalScroller:YES];
     [scrollview setAutohidesScrollers:YES];
     [scrollview setDocumentView:m_OutlineView];
+
+    ApplyScrollViewBorderType();
 
     // initialize the native control itself too
     InitOutlineView(style);
--- a/src/osx/cocoa/window.mm
+++ b/src/osx/cocoa/window.mm
@@ -152,6 +152,27 @@ NSRect wxOSXGetFrameForControl( wxWindowMac* window , const wxPoint& pos , const
     return wxToNSRect( sv, bounds );
 }
 
+
+void wxWidgetCocoaImpl::ApplyScrollViewBorderType()
+{
+    wxWindowMac* peer = GetWXPeer();
+
+    if ( peer && [m_osxView isKindOfClass:[NSScrollView class]] )
+    {
+        wxBorder border = peer->GetBorder();
+
+        // Enable the scrollview border for native behavior
+        // otherwise turn off the border and allow WX to draw
+        // the borders using wxWindowMac::MacPaintBorders
+
+        if ( border == wxBORDER_DEFAULT || border == wxBORDER_THEME )
+            [(NSScrollView*)m_osxView setBorderType:NSBezelBorder];
+        else
+            [(NSScrollView*)m_osxView setBorderType:NSNoBorder];
+    }
+}
+
+
 @interface wxNSView : NSView
 {
 }
--- a/src/osx/listbox_osx.cpp
+++ b/src/osx/listbox_osx.cpp
@@ -72,8 +72,10 @@ bool wxListBox::Create(
     DontCreatePeer();
     m_blockEvents = false;
 
+#if !defined(__WXOSX_COCOA__)
     if ( ! (style & wxNO_BORDER) )
         style = (style & ~wxBORDER_MASK) | wxSUNKEN_BORDER ;
+#endif
 
     wxASSERT_MSG( !(style & wxLB_MULTIPLE) || !(style & wxLB_EXTENDED),
                   wxT("only a single listbox selection mode can be specified") );
@@ -131,10 +133,10 @@ void wxListBox::FreeData()
 
 void wxListBox::DoSetFirstItem(int n)
 {
-    // osx actually only has an implementation for ensuring the visibility of a row, it does so  
+    // osx actually only has an implementation for ensuring the visibility of a row, it does so
     // by scrolling the minimal amount necessary from the current scrolling position.
-    // in order to get the same behaviour I'd have to make sure first that the last line is visible, 
-    // followed by a scrollRowToVisible for the desired line 
+    // in order to get the same behaviour I'd have to make sure first that the last line is visible,
+    // followed by a scrollRowToVisible for the desired line
     GetListPeer()->ListScrollTo( GetCount()-1 );
     GetListPeer()->ListScrollTo( n );
 }
--- a/src/osx/listbox_osx.cpp
+++ b/src/osx/listbox_osx.cpp
@@ -72,11 +72,6 @@ bool wxListBox::Create(
     DontCreatePeer();
     m_blockEvents = false;
 
-#if !defined(__WXOSX_COCOA__)
-    if ( ! (style & wxNO_BORDER) )
-        style = (style & ~wxBORDER_MASK) | wxSUNKEN_BORDER ;
-#endif
-
     wxASSERT_MSG( !(style & wxLB_MULTIPLE) || !(style & wxLB_EXTENDED),
                   wxT("only a single listbox selection mode can be specified") );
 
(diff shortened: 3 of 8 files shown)


vadz (commented):
Thanks, I see what you meant now but I'm surprised that we never call `[NSView setBorderType:]` elsewhere — how do borders work for plain `wxWindow`s under Mac?

I also wonder why do we map `wxBORDER_STATIC`, `wxBORDER_SIMPLE` etc to `NSNoBorder`? I'd probably use `[NSLineBorder](https://developer.apple.com/documentation/appkit/nsbordertype/lineborder?language=objc)` for them, but even `NSBezelBorder` would seem to be better than no border at all.

@csomor Your thoughts about this would be very welcome, TIA!

willco007:
IIRC the various borders on macOS are faked by WX, but I haven't really dug into them since they aren't part of the native controls. FWIW `NSLineBorder` etc are all obsolete and shouldn't be used, the only style that is kept up by Apple is `NSBezelBorder` which is the default style.

There is no concept of a border on a standard NSView, which is probably why they are synthesized by WX.

csomor:
> Thanks, I see what you meant now but I'm surprised that we never call `[NSView setBorderType:]` elsewhere — how do borders work for plain `wxWindow`s under Mac?
> 
> I also wonder why do we map `wxBORDER_STATIC`, `wxBORDER_SIMPLE` etc to `NSNoBorder`? I'd probably use `[NSLineBorder](https://developer.apple.com/documentation/appkit/nsbordertype/lineborder?language=objc)` for them, but even `NSBezelBorder` would seem to be better than no border at all.
> 
> @csomor Your thoughts about this would be very welcome, TIA!

I didn't have proper border types for many situations way back when, therefore for these cases I set the border to no border and draw them in wxWindowMac::MacPaintBorders using the proven HITheme calls. For fields like buttons, textcontrols etc. the native versions are used, only turned off when not appropriate `setBordered:NO` or `setBezeled:NO`. For controls like buttons and text controls `setBezelStyle` is used to indicate the border type.

willco007:
Testing more controls that use NSScrollView, `wxTextCtrl` also has this issue. I'll add it to this MR if you don't mind.

src/osx/textctrl_osx.cpp
```diff
@@ -89,8 +89,10 @@ bool wxTextCtrl::Create( wxWindow *parent,
     DontCreatePeer();
     m_editable = true ;
 
+#if !defined(__WXOSX_COCOA__)
```
vadz: This doesn't make sense to me, if `wxTextCtrl` without a border can work, we should just remove this.

src/osx/listbox_osx.cpp
```diff
@@ -131,10 +133,10 @@ void wxListBox::FreeData()
 
 void wxListBox::DoSetFirstItem(int n)
 {
-    // osx actually only has an implementation for ensuring the visibility of a row, it does so  
+    // osx actually only has an implementation for ensuring the visibility of a row, it does so
```
vadz, line 131: Please configure your editor to avoid automatically stripping whitespace. TIA!

src/osx/cocoa/dataview.mm
```diff
@@ -1985,8 +1985,13 @@ -(BOOL) becomeFirstResponder
 {
     // initialize scrollview (the outline view is part of a scrollview):
     NSScrollView* scrollview = (NSScrollView*) GetWXWidget();
+    wxBorder border = (wxBorder)(style & wxBORDER_MASK);
+
+    if ( (border == wxBORDER_DEFAULT) || (border == wxBORDER_THEME) )
+        [scrollview setBorderType:NSBezelBorder];
+    else
+        [scrollview setBorderType:NSNoBorder];
```
vadz: I still think that we should use `wxWindow::GetBorder()` here and that we should extract this into some reusable function (`OSXApplyBorder()`?).

Moreover, as Stefan explained, the rest of the code uses `setBordered:NO`, so I'd rather do this here too for consistency.

Combining both of these thoughts, maybe this could be replaced with just

```cpp
if ( peer->GetBorder() == wxBORDER_NONE )
    [scrollview setBordered:NO];
```

? Or do we need to call `setBezeled:YES` too?

vadz (commented):
Sorry, but I'd like to clean this up a bit more before applying.

src/osx/textctrl_osx.cpp
willco007: I left this in for backwards compatibility for the non-cocoa implementation (same with listbox_osx.cpp).  Is it needed anymore, probably not, but it was there for a reason I figure.

src/osx/cocoa/dataview.mm
willco007: Bordered is different than bezel type in this case. Bordered is hiding the bezel, but we still need to set the type.

src/osx/textctrl_osx.cpp
vadz: There is only iOS implementation now, but it's probably completely different border-wise anyhow.

src/osx/cocoa/dataview.mm
vadz: OK, but the point was that we should be consistent with the rest of wxOSX code and use `setBezeled:` here too (unless we want to change all the rest of the code to use `setBorderType:`).

willco007: `setBorderType` and `setBordered` do different things and not all controls respond to both calls, there isn't a way to make it 100% consistent.

vadz: Oh. Thanks, I hadn't realized that `setBorderType:NSNoBorder` and `setBordered:NO` could be doing different things.

Then I'd just like to extract this into `wxWindow::OSXApplyBorder(NSView* v)` and call it from all the places where it's needed.

willco007: The TL;DR is scroll view uses `borderType` while other buttons use `setButtonType` and `setBordered`.

vadz: So the helper function should probably be called `OSXApplyBorderType()`...

willco007: What do you have in mind for that function? Currently the macOS controls translate the wxBORDER styles passed to what makes sense for the control type (or ignored if they don't make sense).  Trying to abstract that into a helper function doesn't seem particularly useful since each control is very specific.  However, perhaps, abstracting the above scrollview styling into a helper function might make more sense since the styling needs to be applied to several scrollviews in the same way.

vadz: I'd just like to have a function for this because I think we may want to change this in all places where it's used in the future and because this would allow to have a single place where to write a comment explaining what's going on here (which I personally really need and I think it could be useful to the others as well).

willco007: That makes sense, however I'm not sure I'm familiar enough with the codebase to make all those changes.

vadz: The new function can be added to `include/wx/osx/window.h` (it would be called as `peer->OSXApplyBorderType(view)` from the places where you're doing it now), AFAICS this should be simple to do, but please let me know if I'm missing something?

I could do it myself, but I wouldn't be able to write a good comment for it, which is what I'm after!

willco007: I've been poking around with this and adding a function to window.h and then passing in the NSView results in a lot of `[view isKindOfClass:]` calls to determine what type of control it is to then apply the correct border style to. I'm not sure it's worth doing this way when we know the type of the view already at creation time and can apply it there which it is doing already. Thoughts? Comments?

vadz: Sorry, I didn't mean to ask you to do anything complicated, i.e. write a function that would work for all kinds of windows. I really just wanted to perform a trivial refactoring and put the code you added into a function that would be only used for the views that need `setBorderType:` to be called and call it for them only.

willco007: AH! Well, that would certainly be easier. :) OK, will do that.

src/osx/listbox_osx.cpp
willco007: BTW, the .editorconfig file in the wxWidgets project has `trim_trailing_whitespace` enabled which is how this happened.

vadz: Hmm, we do want to avoid whitespaces in the new code. But we don't want to mix up removing them with the other, significant, code changes. The solution is probably to trim all the existing whitespace (I think it's only present in Mac sources...).

src/osx/cocoa/dataview.mm
willco007: @vadz Take a look at my latest push, hopefully it is acceptable. :)

vadz:
This looks good, thanks! I'll merge it soon with just some cosmetic changes.

src/osx/cocoa/window.mm
```diff
@@ -152,6 +152,27 @@ NSRect wxOSXGetFrameForControl( wxWindowMac* window , const wxPoint& pos , const
     return wxToNSRect( sv, bounds );
 }
 
+
+void wxWidgetCocoaImpl::ApplyScrollViewBorderType()
+{
+    wxWindowMac* peer = GetWXPeer();
+
+    if ( peer && [m_osxView isKindOfClass:[NSScrollView class]] )
```
vadz, line 160: I wonder if the `isKindOfClass:` call is (still) needed: isn't it supposed to be always true because we only call this function for the scrollview-derived views? I.e. maybe it should be changed to an assert? Or just removed completely?

willco007: It's purely for protection here, we could turn it into an assert if you prefer.

vadz: AFAIK nothing really bad happens if we send `setBorderType:` to an object not supporting it, right? If so, I'd indeed make it just an assert (I can do it myself, no need to update the PR) to make it clear that it is the precondition for calling this function.

willco007: Oh it does, it throws a runtime exception.  You don't want to call selectors on objects that don't support them. :)

willco007: Perhaps just add a debug assert in the failure case noting that the function was called on a control that isn't a scrollview. That way, if the container class is changed by some future OS (doubtful) or someone is messing around and accidentally calls it it doesn't blow up in production.

PR #25579, opened 2025-06-26
Improve expander in wxPropertyGrid

### LINKED ISSUE #25571: [3.3.0 wxPropertyGrid] group collapse/expand arrow images are too small
#### Bug description:

It seems to me that group collapse/expand arrow images are too small in wxPropertyGrid on normal dpi.

Image

#### Expected vs observed behavior:

I expect them to be in the same size as tree view expand/collapse images. 

- wxWidgets version 3.3.0
- wxWidgets port wxMSW
- OS Windows 11

MaartenBent:
* Make wxPropertyGrid a wxSystemThemedControl so it uses the default expander
* Don't use hard-coded expander sizes in wxPropertyGrid so the expander gets the correct size, fixes #25571
* Improve margin of sub-groups in wxPropertyGrid so the expander does not overlap with the margin

### DIFF
--- a/include/wx/propgrid/propgrid.h
+++ b/include/wx/propgrid/propgrid.h
@@ -17,6 +17,7 @@
 #include "wx/scrolwin.h"
 #include "wx/recguard.h"
 #include "wx/time.h" // needed for wxMilliClock_t
+#include "wx/systhemectrl.h"
 
 #include "wx/propgrid/property.h"
 #include "wx/propgrid/propgridiface.h"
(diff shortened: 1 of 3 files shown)


MaartenBent:
I haven't tested if this looks correctly on linux/macOS.

vadz:
Unfortunately this doesn't look great under Linux (wxGTK, 200% DPI): this is how it looks in master:

image

and this is how it looks with this PR: 

image

The arrows are clearly scaled, but now they also have too big margins around them. Can this be easily fixed (I didn't look at the code at all to be honest)?

MaartenBent:
> The arrows are clearly scaled, but now they also have too big margins around them. Can this be easily fixed (I didn't look at the code at all to be honest)?

It now uses the image width (16dip px I guess), and left and right of the image a 'gutter' of `max(3, imagewidth/3)`. Previously it was a fix image width of 9 px.
Maybe the gutter size can be reduced. I'll try some things.

vadz:
Sorry, this is not urgent at all but I wanted to ask if you still plan to try something here or should I find some way to reduce the margins myself?

MaartenBent:
I committed a fix. Seems the gutter is mostly useless now after using the real size of the expander icons.

vadz:
Great, thanks a lot! Will merge soon.

PR #25580, opened 2025-06-27
MacOS: wxFontPickerCtrl trim off font encoding from font display name

willco007:
Trim off font encoding on Mac and match Windows behavior.


vadz:
Makes sense, thanks!

vadz:
BTW, you've used everydaysoftware.net and panic.com as email for your commits in the recent PRs, do you want to keep them like this or should I rewrite them to use the same email and, if so, which one?

willco007:
@vadz Oh, sorry about that, working on different machines. You can keep it as-is if it's OK with you. I'll update my machine to use the same emails.

PR #25583, opened 2025-06-30
Do not draw Spinbutton when resized to zero

copymark:
With MSW DarkMode ONLY:

When you resize the parent frame of a spinbutton to zero (either width or height), then it crashes on creation of the Bitmap. 

The previous check only checked whether width and height are BOTH zero.

The new check returns when either one of those is 0.

### DIFF
--- a/src/msw/spinbutt.cpp
+++ b/src/msw/spinbutt.cpp
@@ -176,7 +176,7 @@ void wxSpinButton::OnPaint(wxPaintEvent& event)
         const RECT rc = wxGetClientRect(GetHwnd());
         const wxSize size{rc.right - rc.left, rc.bottom - rc.top};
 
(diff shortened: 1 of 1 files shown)


copymark:
Hi @vadz 
Is this a flaky CI run?
I don't see any error related to my changes in the logs :?

vadz:
Thanks for fixing this! CI error is indeed just some unrelated problem on AppVeyor, I'll merge this soon, although I'll probably replace the test with `!IsEmpty()` as it seems a bit more readable to me.

vadz:
Oops, except, of course `wxSize` doesn't have `IsEmpty()`, only `wxRect` does. But this seems wrong, I think we should add it to it... Anyhow, for now I've applied your PR unchanged, thanks again.

PR #25586, opened 2025-07-01
Update MSW dark mode and wxWebViewEdge docs

PBfordev:
Extend the list of MSW dark mode limitations and improve documentation for building wxWebViewEdge.

Regarding `wxWebViewEdge` building instruction, it seems that the procedure for _configure_ is missing.

### DIFF
--- a/interface/wx/app.h
+++ b/interface/wx/app.h
@@ -1429,12 +1429,13 @@ public:
         - The following dialogs wrapping common windows dialogs don't support
           dark mode: wxColourDialog, wxFindReplaceDialog, wxFontDialog,
           wxPageSetupDialog, wxPrintDialog.
-        - wxDatePickerCtrl and wxTimePickerCtrl don't support dark mode and
-          use the same (light) background as by default in it.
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, will push soon.

I've never built `wxWebViewEdge` with gcc (and hence configure), so I don't know how to do it and if it even works (I think it does, as I remember seeing checks for MSVC in this code, but this is hardly a proof).

PBfordev:
> I've never built `wxWebViewEdge` with gcc (and hence configure), so I don't know how to do it and if it even works 

It can be be built with CMake. Code related to `ICoreWebView2EnvironmentOptions` is for some reason in MSVC-only guards, maybe due to missing some WRL stuff? https://github.com/wxWidgets/wxWidgets/blob/d2372005239f57a8b5c78baacd084ada413ccb5b/src/msw/webview_edge.cpp#L35-L41

E.g.
https://github.com/wxWidgets/wxWidgets/blob/d2372005239f57a8b5c78baacd084ada413ccb5b/src/msw/webview_edge.cpp#L265-L276

PR #25588, opened 2025-07-02
remove wxNavigationEnabled<> from wxVListBox

wsu-cb:


vadz:
There are some actual errors here:

```
In file included from ./src/generic/vlbox.cpp:24:
./include/wx/vlbox.h: In member function ‘virtual bool wxVListBox::IsCurrent() const’:
./include/wx/vlbox.h:95:65: error: cannot call member function ‘virtual bool wxWindow::IsCurrent() const’ without object
   95 |     bool IsCurrent() const { return wxVScrolledWindow::IsCurrent(); }
      |                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
make: *** [Makefile:29505: coredll_vlbox.o] Error 1
```

It looks like `wxVListBox` doesn't inherit from `wxVScrolledWindow` any more, at least in some ports?

interface/wx/vscroll.h
```diff
@@ -631,15 +631,17 @@ class wxVarHVScrollHelper : public wxVarVScrollHelper,
     @library{wxcore}
     @category{miscwnd}
 
+    @see wxVScrolledWindow, wxVScrolledCanvas
```
vadz, line 634: If you don't mind, I'll add a note about `wxVScrolled` only available since 3.3.1 here too, to ensure people browsing the documentation don't start using it in the code they want to compile with 3.2.

vadz (approved):
Looks good, thanks!

I'll merge it as part of the next push.

PR #25591, opened 2025-07-02
Minor cleanup in GTK file picker

### LINKED ISSUE #25590: `wxFilePickerCtrl::GetFileName()` always returns empty string in 3.3.0
The problem can be seen in the widgets sample after applying

```
diff --git a/samples/widgets/filepicker.cpp b/samples/widgets/filepicker.cpp
index bc522910cf..8b4d3abcc5 100644
--- a/samples/widgets/filepicker.cpp
+++ b/samples/widgets/filepicker.cpp
@@ -313,8 +313,11 @@ void FilePickerWidgetsPage::OnButtonReset(wxCommandEvent& WXUNUSED(event))
 
 void FilePickerWidgetsPage::OnFileChange(wxFileDirPickerEvent& event)
 {
-    wxLogMessage("The file changed to '%s' ! The current working directory is '%s'",
-                 event.GetPath(), wxGetCwd());
+    wxLogMessage("The file changed to '%s' (the control has '%s')."
+                 "The current working directory is '%s'",
+                 event.GetPath(),
+                 m_filePicker->GetFileName().GetFullPath(),
+                 wxGetCwd());
 }
 
 void FilePickerWidgetsPage::OnCheckBox(wxCommandEvent &event)
```

It is due to the changes of 071b4dd0f1 (Use GtkFileChooserNative for wxFileDialog/wxDirDialog when possible, 2025-02-09) as can be seen by reverting it, after reverting 9036a0b0a5 (Fix EndModal() for wxFileDialog/wxDirDialog when using GtkFileChooserNative, 2025-02-22) first to avoid a merge conflict.

First reported in [this thread](https://groups.google.com/g/wx-users/c/pGpKQW0VLdg).

vadz:
This was opened to fix #25590 but it was fixed by another PR, so now this one only contains minor cleanup.

### DIFF
--- a/include/wx/gtk/filedlg.h
+++ b/include/wx/gtk/filedlg.h
@@ -80,6 +80,11 @@ private:
     void OnSize(wxSizeEvent&);
     virtual void AddChildGTK(wxWindowGTK* child) override;
 
+    const wxGtkFileChooser& GetFileChooser() const
+    {
+        return m_fcNative ? *m_fcNative : m_fc;
+    }
+
     wxGtkFileChooser    m_fc;
     wxGtkFileChooser* m_fcNative = nullptr;
     GtkFileChooser* m_fileChooserNative = nullptr;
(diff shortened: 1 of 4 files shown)


paulcor:
I think it would be better to use a more self-contained solution that does not require `wxFilePickerCtrl` to know about the internals of `wxFileDialog`. See #25597

vadz:
I agree, thanks. I'll still push the other commits from here (after the other PR is merged).

PR #25596, opened 2025-07-04
Add wxVERSION and wxCOPYRIGHT into the CMake cache

Blake-Madden:
If including wx as part of your program's build process, this is a nice way to check wxWidgets's version number.

I kept them `INTERNAL` so that they don't appear in CMake GUI.


vadz:
I don't know how common (or unusual) it is to do this, especially for the copyright, but why not.

Maarten, please let me know if you have any objections/comments, otherwise I'll merge this.

Blake-Madden:
The `wxCOPYRIGHT` probably isn't needed, I just included it for completeness. Please feel free to remove that if you are opposed to it.

`wxVERSION` is what I would find useful; it allows me to enforce a version check when statically including wx into my project.

MaartenBent (approved):
Looks good to me. I was afraid it might prevent updating the cached values when copyright or version got updated, but `INTERNAL`  also implies `FORCE`, so no issue there.

PR #25597, opened 2025-07-05
Fix wxFilePickerCtrl with GTK >= 3.20

paulcor:
GtkFileChooserNative will not be used when wxFileDialog is used by wxFilePickerCtrl, so make sure values will be retrieved from GtkFileChooserDialog and not the unused "native" chooser.
See #25590


vadz (approved):
Thanks, this indeed looks better.

Ideal, I think, would be to avoid creating both choosers and then either deleting or not using one of them, but I couldn't find a simple way to apply all attributes to the one that ends up being created.

PR #25599, opened 2025-07-05
Add a unit test for wxComboBox::SetStringSelection() events

vadz:
Check that it doesn't produce any,

See #25553.


vadz:

PR #25601, opened 2025-07-05
Make disabled bitmaps darker, not lighter, in dark mode in wxMSW

vadz:
Use wxImage::ChangeLightness(66) instead of ConvertToDisabled() in dark mode, as the latter makes the image lighter (by alpha blending it with white) and we want to make it darker (by alpha blending it with black, as this function does).

Closes #25575.

### DIFF
--- a/src/msw/anybutton.cpp
+++ b/src/msw/anybutton.cpp
@@ -40,6 +40,7 @@
 #include "wx/imaglist.h"
 #include "wx/stockitem.h"
 #include "wx/msw/private/button.h"
+#include "wx/msw/private/darkmode.h"
 #include "wx/msw/private/dc.h"
 #include "wx/msw/private/winstyle.h"
 #include "wx/msw/uxtheme.h"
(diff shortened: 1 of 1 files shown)


src/msw/anybutton.cpp
```diff
@@ -87,6 +88,25 @@ using namespace wxMSWImpl;
 extern wxWindowMSW *wxWindowBeingErased; // From src/msw/window.cpp
 #endif // wxUSE_UXTHEME
 
+// Create a disabled bitmap from the normal one taking the application mode
+// (light or dark) into account.
+static wxBitmap CreateDisabledBitmap(const wxBitmap& bmp)
```
PBfordev, line 93: Just curious: Why handle it like this, instead of making `wx(Image|Bitmap)::ConvertToDisabled()` and `wxColor::MakeDisabled()` work also in (MSW) dark mode, even if dark mode had to be user set as a parameter?

Disclaimer: I have no idea how disabling works.

vadz: It seems wrong to have such simple functions behave differently depending on some global state (`IsDark()`), e.g. because they could be used for something other than showing the bitmap on the screen — just imagine a command line graphics conversion/manipulation program à la ImageMagick whose results would depend on whether the system is running in light or dark mode.

PBfordev: > It seems wrong to have such simple functions behave differently depending on some global state

I did not meant for them to depend on a state, I meant them having an optional parameter (defaulting to light).

vadz: This would be doable, but I didn't want to provide public API I don't really know how to implement — because I'm not sure at all that the current version is optimal (from whichever point of view).

vadz:

PR #25610, opened 2025-07-06
Add wxOwnerDrawnComboBox ctor overload not taking any items

vadz:
Surprisingly, similar Create() overload already existed, but ctor one did not.

Fix this inconsistency and provide ctor with the same signature as Create() too.


vadz:

Issue #25612, opened 2025-07-07
'internat' sample crashes or shows rubbisch behind "Preferred UI languages:"


BusyHarry:
### Description
wxWidgets V3.3.0, windows10, all locally compiled with VS2019

If it doesn't crash, it shows one language, and then a long line of rubbisch.
Debugging showed that the function 'GetUserPreferredLanguagesFromRegistry()' in '.\src\msw\uilocale.cpp' does not loop correctly over the buffer in line 113: 'while (*buf != 0)'
At least on my system, the buffer is NOT zero terminated, so the loop finds language(s) that are not there.
Microsoft says: "if the value being queried is a string (REG_SZ, REG_MULTI_SZ, and REG_EXPAND_SZ) the value returned is NOT guaranteed to be null-terminated."
The solution would be to zero the buffer before use, or add a terminating zero at the end of the data: the 'RegQueryValueEx()' does return the actual number of BYTES in the buffer. So a simpel 'buf[valueSize/sizeof(WCHAR)] = 0;' before the loop would suffice (assumption: buffer not filled upto the last byte!)

```diff
diff --git a/src/msw/uilocale.cpp.org b/src/msw/uilocale.cpp
index 52618f2..fa2504a 100644
--- a/src/msw/uilocale.cpp.org
+++ b/src/msw/uilocale.cpp
@@ -109,6 +109,7 @@ void GetUserPreferredLanguagesFromRegistry(wxVector<wxString>& userLanguages)
         {
             // Extract languages from multi-string value
             WCHAR* buf = languages.get();
+            buf[valueSize/sizeof(WCHAR)] = 0;
             while (*buf != 0)
             {
                 const wxString language(buf);
```

BusyHarry:
A solution that always works:

```diff
index 52618f2..84229fc 100644
--- a/src/msw/uilocale.cpp.org
+++ b/src/msw/uilocale.cpp
@@ -104,11 +104,12 @@ void GetUserPreferredLanguagesFromRegistry(wxVector<wxString>& userLanguages)
         // Retrieve the "Languages" value from the key
         DWORD type = REG_SZ;
         DWORD valueSize = 256;
-        wxScopedArray<WCHAR> languages(valueSize);
+        wxScopedArray<WCHAR> languages(valueSize+1);
[...]

PR #25613, opened 2025-07-07
Fix/improve `GetUserPreferredLanguagesFromRegistry()` in wxMSW code

vadz:
Most importantly, fix the buffer overrun reported in #25612.

cc @utelle

### DIFF
--- a/include/wx/msw/registry.h
+++ b/include/wx/msw/registry.h
@@ -139,6 +139,8 @@ public:
   bool  IsOpened() const { return m_hKey != nullptr; }
     // for "if ( !key ) wxLogError(...)" kind of expressions
   operator bool()  const { return m_dwLastError == 0; }
+    // return the associated HKEY, possibly null if the key is not opened
+  WXHKEY GetHkey() const { return m_hKey; }
 
   // operations on the key itself
     // explicitly open the key (will be automatically done by all functions
--- a/interface/wx/msw/registry.h
+++ b/interface/wx/msw/registry.h
@@ -237,6 +237,15 @@ public:
     bool GetFirstValue(wxString& strValueName, long& lIndex);
 
     /**
+        Return the associated HKEY handle.
+
+        It may be null if the key is not opened.
+
+        @since 3.3.1
+     */
+    WXHKEY GetHkey() const;
+
+    /**
         Gets information about the key. Returns @true if successful.
 
         @param pnSubKeys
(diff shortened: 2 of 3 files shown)


utelle:
> Most importantly, fix the buffer overrun reported in #25612.

Strangely enough I never experienced the behaviour described in #25612 myself up to now. Otherwise I would have certainly detected my mistake of overlooking the remark about a potentially missing final NUL character.

I compiled wx 3.3.0 with VS2022 under Windows 11. On my development system I get a proper list with 6 locale id entries (without the fix, that is). Admittedly, I haven't tested under Windows 10.

‼️BTW, it may be necessary to append _**2**_ (**!**) NUL characters to be on the safe side - one for the last language entry, and one for finalizing the language list. Microsoft's warning isn't absolutely precise about `REG_MULTI_SZ` values.

vadz:
I don't see this either (with Windows 10), but apparently this does happen — and it's better not to rely on something explicitly documented not to be guaranteed anyhow.

And you're right, even after this fix we still rely on a single NUL being present, thanks. I've fixed this in the last commit here.

@BusyHarry Please test this on your system (which is the only known one which exhibits the behaviour that resulted in the error), TIA!

vadz:
Sorry, had to force-push (without any real changes) to restart the CI builds with fixed master.

BusyHarry:
I have tested it and its working.
In my system (Dutch), the query returns only 1 language with 1 nul at the end.
ps had some problems getting the file compiled: the class wxRegKey did not have the member GetHkey(), so I had to add it.

vadz:
> I have tested it and its working.

Thanks, I'll merge this soon then.

> ps had some problems getting the file compiled: the class wxRegKey did not have the member GetHkey(), so I had to add it.

The commit adding it is included in this PR. You need to either checkout the branch the branch or merge it (possibly with `--squash` or `--no-commit`) in your current branch to test it.

BusyHarry:
one remark:  you are testing 
```
if ( p - languages >= numChars )
```
This should probably be:
```
if ( p - languages >= valueSize/sizeof(WCHAR))
```
Sorry for the late reaction :(

vadz:
Ugh, you're right, thanks for noticing this. I've force-pushed once again to fix this.

utelle:
> And you're right, even after this fix we still rely on a single NUL being present, thanks. I've fixed this in the last commit here.

Hm, I don't understand why you don't simply append **2** (in words: **two**) NUL characters instead of appending just one plus explicitly checking whether the pointer `p` points beyond the buffer end. Appending **2** NUL characters instead of **1** costs just 2 bytes extra space and would make checking the pointer in each loop iteration unnecessary.

vadz:
I just find this check more explicit.

PR #1781, opened 2020-04-06
Use locale grouping to format numbers

nkottary:
Currently numbers are grouped in groups of 3 regardless of locale. This PR implements grouping by looking at locale. I took some screenshots by using `wxIntegerValidator` on `wxTextCtrl`:

Default grouping or grouping with "en_US", "en_UK" etc
grouping_us

Grouping with "en_IN"
grouping_in

PR #25614, opened 2025-07-08
Make wxNumberFormatter's helper functions publically available

Blake-Madden:
No real changes, just exposes `wxNumberFormatter`'s `AddThousandsSeparators`, `RemoveThousandsSeparators`, and `RemoveTrailingZeroes` publicly.

I could use `RemoveTrailingZeroes()` myself for cleaning up strings from my data importer, and IMO there is real value is making these functions available.

### DIFF
--- a/include/wx/numformatter.h
+++ b/include/wx/numformatter.h
@@ -68,10 +68,6 @@ public:
     // function returns true.
     static bool GetThousandsSeparatorIfUsed(wxChar *sep);
 
-private:
-    // Post-process the string representing an integer.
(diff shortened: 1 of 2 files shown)


vadz:
I'm not sure about the thousands separators functions, we may need to change their signature if #1781 is finally merged, so I'd rather not make them public now.

Blake-Madden:
Only `RemoveTrailingZeroes` will be public now. This is the function I'm really interested, could be useful to others.

vadz:
Thanks, will merge soon.

PR #25620, opened 2025-07-09
Don't change invalid current `wxGrid` cell in `SetTable()`

### LINKED ISSUE #25498: wxGrid::CreateGrid generates wxEVT_GRID_SELECT_CELL in 3.3
`wxGrid::CreateGrid()` generates a synchronous `wxEVT_GRID_SELECT_CELL` event when called with nonzero arguments in wx 3.3, while it used to not do that in wx 3.2.

This is a breaking change because the event handler might be accessing variables that get initialized after the `CreateGrid()` call. If this change was intentional then please mention it in the changes.txt.

Easy repro using the grid sample: change the setup code in the `GridFrame` constructor from `CreateGrid(0, 0)` to `CreateGrid(1, 1)`, and set a breakpoint in `GridFrame::OnSelectCell()`. Notice that with 3.2 the breakpoint is hit thru `AppendRows()`, but with 3.3 it is reached earlier, thru `CreateGrid()`.

Tested on wxMSW.

vadz:
Changing the table shouldn't make the cell (0, 0) current if we hadn't had any current cell before, this was unintentional side effect of e570d85164 (Ensure current cell remains valid when wxGrid table changes, 2023-09-15) which resulted in an unwanted wxEVT_GRID_SELECT_CELL event when creating the grid.

Fix this by only calling UpdateCurrentCellOnRedim() if we have a valid current cell to update.

See #23751.

Closes #25498.

----

cc @CookieLau

### DIFF
--- a/src/generic/grid.cpp
+++ b/src/generic/grid.cpp
@@ -3135,7 +3135,12 @@ wxGrid::SetTable(wxGridTableBase *table,
 
     InvalidateBestSize();
 
-    UpdateCurrentCellOnRedim();
+    // If we already have a valid current cell, ensure that it is in valid
+    // range for the new table with a possibly different number of rows/columns
+    // but don't do anything if the current cell is invalid, setting the table
+    // shouldn't automatically select the cell at (0, 0).
+    if ( m_currentCellCoords != wxGridNoCellCoords )
+        UpdateCurrentCellOnRedim();
 
     return m_created;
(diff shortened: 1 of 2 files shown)


vadz:
I'm going to merge this soon if there are no objections (cc @rudolfwalter).

vadz:

PR #25628, opened 2025-07-11
Add opacity state functions to wxColour

Blake-Madden:
Adds `IsTransparent()`, `IsOpaque()`, and `IsTranslucent()` to `wxColour`.

These are trivial function, but can make code more readable IMO.


vadz:
Thanks, I agree that this is nice to have, I will merge the PR soon.

PR #25629, opened 2025-07-11
Clarify & fix wxRect2DDouble docs, fix minor comment typos

Blake-Madden:


src/dfb/dc.cpp
```diff
@@ -479,7 +479,7 @@ wxCoord wxDFBDCImpl::GetCharWidth() const
 
     int w = -1;
     GetCurrentFont()->GetStringWidth("H", 1, &w);
-    // VS: YDEV is corrent, it should *not* be XDEV, because font's are only
+    // VS: YDEV is correct, it should *not* be XDEV, because font's are only
```
PBfordev:
```suggestion
    // VS: YDEV is correct, it should *not* be XDEV, because fonts are only
```

src/dfb/dc.cpp
```diff
@@ -505,7 +505,7 @@ void wxDFBDCImpl::DoGetTextExtent(const wxString& string, wxCoord *x, wxCoord *y
 
     if ( f->GetStringExtents(string.utf8_str(), -1, &rect, nullptr) )
     {
-        // VS: YDEV is corrent, it should *not* be XDEV, because font's are
+        // VS: YDEV is correct, it should *not* be XDEV, because font's are
```
PBfordev:
```suggestion
        // VS: YDEV is correct, it should *not* be XDEV, because fonts are
```

src/generic/dbgrptg.cpp
```diff
@@ -126,7 +126,7 @@ class wxDumpOpenExternalDlg : public wxDialog
 public:
     wxDumpOpenExternalDlg(wxWindow *parent, const wxFileName& filename);
 
-    // return the command chosed by user to open this file
+    // return the command chose by user to open this file
```
PBfordev:
```suggestion
    // return the command chosen by user to open this file
```

interface/wx/geometry.h
```diff
@@ -388,8 +386,7 @@ class wxRect2DDouble
      */
     void SetLeftTop( const wxPoint2DDouble &pt );
     /**
-       Set the top-left point of the rectangle, while preserving the
-       width and height of the rectangle.
+       Set the top-left point of the rectangle, preserving the width.
```
PBfordev: We are changing both x and y coordinate, so shouldn't we state what happens to the height (as before)?

src/aui/tabmdi.cpp
```diff
@@ -458,12 +458,12 @@ wxAuiMDIChildFrame::wxAuiMDIChildFrame(wxAuiMDIParentFrame *parent,
 {
     Init();
 
-    // There are two ways to create an tabbed mdi child fram without
+    // There are two ways to create an tabbed mdi child frame without
```
PBfordev:
```suggestion
    // There are two ways to create a tabbed mdi child frame without
```

src/msw/notebook.cpp
```diff
@@ -621,7 +621,7 @@ wxNotebookPage *wxNotebook::DoRemovePage(size_t nPage)
         if ( selNew != wxNOT_FOUND )
         {
             // No selection change, just refresh the current selection.
-            // Because it could be that the slection index changed
+            // Because it could be that the selection index changed
             // we need to update it.
             // Note: this does not mean the selection it self changed.
```
PBfordev:
```suggestion
            // Note: this does not mean the selection itself changed.
```

PBfordev (commented):
I dared to fix a few I noticed around the proposed changes. ;)

interface/wx/geometry.h
Blake-Madden: You're right, rolled back to original description. Thanks!

Blake-Madden: Applied all additional fixes, thanks.

vadz:

PR #25632, opened 2025-07-12
Restore compatibility with code using wxPG_XXX flags

### LINKED ISSUE #25627: Undocumented build-breaking change in wxPropertyGrid?
_(Based on [a wxForum post](https://forums.wxwidgets.org/viewtopic.php?f=19&t=52164))_

In January 2024, there was a commit [Use enum class to implement wxPGPropValFormatFlags as bitmask](https://github.com/wxWidgets/wxWidgets/commit/ca3acd7a03ac9dce4f4195b717288cd96aa3ab8f), which started using `enum class` instead of `int` for flags.

The changes make code valid in 3.2 such as `int flags = wxPG_RECURSE;` or `propGrid->HideProperty(propArg, true, wxPG_RECURSE);` not compile anymore in 3.3 with MSVS 2022 (`WXWIN_COMPATIBILITY_3_2` set to `1`). 

For example
```cpp
int flags = wxPG_RECURSE;
```
gives errors
```
Error	C4996	'wxPG_RECURSE': use wxPGPropertyValuesFlags::Recurse instead
Error	C2440	'initializing': cannot convert from 'const wxPGPropertyValuesFlags' to 'int'	
```

I don't understand why 
```cpp
propGrid->HideProperty(propArg, true, wxPG_RECURSE);`
```
gives error
```
Error	C4996	'wxPG_RECURSE': use wxPGPropertyValuesFlags::Recurse instead	`
```
It should be just a deprecation warning (I have `Treat Warnings As Errors` set to `No`). It builds with MinGW GCC...

I could not find this change in the [Changes in behaviour which may result in build errors](https://github.com/wxWidgets/wxWidgets/blob/09f433faf39aab3f25b3c564b82448bb845fae56/docs/changes.txt#L156). If it cannot be fixed (I think the original intention was to keep it working and just display the deprecation message?), it should at least be documented.

vadz:
Defining wxPG_XXX of type different from (and not convertible to) "int" wasn't enough to make all code using these constants compile, as it could rely on storing them in int variables.

Change the type of these compatibility constants to "int" to be really compatible and also remove deprecation messages from them, as getting such message when passing them to any function taking a new style enum is enough, there is no need for multiple deprecation messages for the same line of code.

Closes #25627.

### DIFF
--- a/include/wx/propgrid/advprops.h
+++ b/include/wx/propgrid/advprops.h
@@ -159,7 +159,7 @@ protected:
 
 #if WXWIN_COMPATIBILITY_3_2
 // If set, then match from list is searched for a custom colour.
-constexpr int wxPG_PROP_TRANSLATE_CUSTOM = static_cast<int>(wxPGPropertyFlags::Reserved_1);
+constexpr int wxPG_PROP_TRANSLATE_CUSTOM = wxPG_PROP_CLASS_SPECIFIC_1;
 #endif // WXWIN_COMPATIBILITY_3_2
 
 // Has dropdown list of wxWidgets system colours. Value used is
--- a/include/wx/propgrid/advprops.h
+++ b/include/wx/propgrid/advprops.h
@@ -159,8 +159,7 @@ protected:
 
 #if WXWIN_COMPATIBILITY_3_2
 // If set, then match from list is searched for a custom colour.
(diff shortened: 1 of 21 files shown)


vadz:
For reference, I also found https://github.com/pkulchenko/wxlua/issues/124 due to the changes of f1273ce152 (Use class enums to implement bitmask types, 2023-02-19). We need to make even more changes to repair this...

PR #25634, opened 2025-07-12
Rewrite event handling in `wxAuiNotebook` to make it simpler and more correct

### LINKED ISSUE #25544: Counterintuitive behaviour of wxAuiNotebook events
In response to #24118, some prose was added in 18bba7b70be1f225009931937a8a52d342308593 about the fact that some events emitted by wxAuiNotebook don't propagate to the parent, and can therefore only be used with Bind() called on the notebook itself.

On certain events, I'm noticing behaviour which is almost the opposite of the above. It took me an hour or two to figure out what was going on, so I thought I'd bring it up.

Here's my test code, as a patch for the AUI sample:

```patch
diff --git a/samples/aui/auidemo.cpp b/samples/aui/auidemo.cpp
index 0ab730c3aa..f340a54ad9 100644
--- a/samples/aui/auidemo.cpp
+++ b/samples/aui/auidemo.cpp
@@ -2611,6 +2611,14 @@ wxAuiNotebook* MyFrame::CreateNotebook()
                                     wxPoint(clientSize.x, clientSize.y),
                                     FromDIP(wxSize(430,200)),
                                     m_notebookStyle);
+
+   ctrl->Bind(wxEVT_AUINOTEBOOK_TAB_RIGHT_DOWN, [](wxAuiNotebookEvent& event) {
+       char s[50];
+       snprintf(s, 50, "right clicked %d\n", event.GetSelection());
+       OutputDebugStringA(s); // or: puts(s);
+       event.Skip();
+   });
+
    ctrl->Freeze();

    wxBitmapBundle pageBmp = wxArtProvider::GetBitmapBundle(wxART_NORMAL_FILE, wxART_OTHER, wxSize(16,16));
```

The first unexpected thing for me was that for every right click, this handler will be called twice. A small amount of digging reveals that there is an internal handler for `wxEVT_AUINOTEBOOK_TAB_RIGHT_DOWN` (namely `wxAuiNotebook::OnTabRightDown()`), which processes the event (does not call `Skip()`), then emits a new one in its place.

What took me a while to realize was that my event handler has priority over the internal one, which means that I always get the original unprocessed event. In this case I also get the second event from the internal handler -- but if I don't call `Skip()`, then the internal handler will never run, so I only get the first event.

Now in the case of `wxEVT_AUINOTEBOOK_TAB_RIGHT_DOWN` (and several of its siblings), the first/original event's `GetSelection()` returns a visual position, while the second/processed one returns the logical index. You can imagine how this can lead to buggy client code.

The only solution I can see is to `Bind()` the event handler to the _parent_ of the notebook (ie, `this->Bind()` instead of `ctrl->Bind()` above). This works because the first event gets consumed internally, while the second one can propagate.

(In my actual code, this was happening in a subclass of wxAuiNotebook, so in that case I have to replace `this->Bind()` with `GetParent()->Bind()`.)

I suppose if it were possible to use different events internally and across the interface you would already be doing that. Barring that option, at least a little more prose would be nice about this case as well. Having some warning in the "Events emitted by this class" section of the wxAuiNotebook page itself would be nice too, perhaps with a link to the text in the wxAuiNotebookEvent page.

I already have the workaround described above, so please feel free to close this whenever you see fit.

vadz:
This is mostly to fix #25544.

### DIFF
--- a/include/wx/aui/auibook.h
+++ b/include/wx/aui/auibook.h
@@ -339,6 +339,11 @@ private:
 
 
 
+// This class is internal and can only be used outside of the library as a
+// pointer to an opaque object, i.e. "wxAuiTabCtrl*" returned by wxAuiNotebook
+// functions can be compared with each other and passed to wxAuiNotebook
+// functions taking them, but not otherwise.
+
 class WXDLLIMPEXP_AUI wxAuiTabCtrl : public wxControl,
                                      public wxAuiTabContainer
 {
--- a/include/wx/aui/auibook.h
+++ b/include/wx/aui/auibook.h
@@ -343,12 +343,9 @@ class WXDLLIMPEXP_AUI wxAuiTabCtrl : public wxControl,
                                      public wxAuiTabContainer
 {
 public:
-
-    wxAuiTabCtrl(wxWindow* parent,
-                 wxWindowID id = wxID_ANY,
-                 const wxPoint& pos = wxDefaultPosition,
-                 const wxSize& size = wxDefaultSize,
-                 long style = 0);
+    // This constructor is only used internally by the library, applications
+    // never create objects of this type.
+    wxAuiTabCtrl(wxAuiNotebook* parent, wxWindowID id);
 
     ~wxAuiTabCtrl();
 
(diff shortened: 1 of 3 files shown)


vadz:
@rudolfwalter If you could please test this, it would be great.

PR #25636, opened 2025-07-13
fixed macOS IME supported

hugozhi:
Fixed a long-standing problem with wxStyledTextEdit that caused the input method's candidate boxes to not follow on macOS, and may need to continue to fix wxRichTextEdit.

### DIFF
--- a/src/osx/cocoa/window.mm
+++ b/src/osx/cocoa/window.mm
@@ -19,6 +19,8 @@
     #include "wx/scrolbar.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, but we can't depend on wxSTC in the core library (it's the stc library which depends on core). So this code needs to be somehow integrated into wxSTC itself, probably by putting it in `src/stc/PlatWXcocoa.mm`. Could you please look into doing this? TIA!

PR #25640, opened 2025-07-14
fixed macOS IME supported

hugozhi:
Tried to fix the long-standing problem that the Input Method candidate box doesn't follow under macOS system.


src/stc/PlatWXcocoa.mm
```diff
@@ -13,6 +13,121 @@
 
 #include "wx/osx/private.h"
 #include "PlatWXcocoa.h"
+#include "wx/stc/stc.h"
+
+@interface wxNSView : NSView
```
vadz, line 18: Do you intentionally use the same interface name as in `src/osx/cocoa/window.mm`? This clearly doesn't work and results in link errors.

lanurmi:
@hugozhi FYI: You do not need to create a new pull request every time you wish to change something in a PR. You can simply push/force-push more commits to your branch.

hugozhi:
@vadz @lanurmi Thank you, but I don't want to spend any more energy on this. Feel free to copy the code and use it in your own projects.

vadz:
Unfortunately I have no idea how to make this code work. Do you mean that it works for you somehow? Are you using monolithic build? I don't see how it could work otherwise...

src/stc/PlatWXcocoa.mm
hugozhi: Because the underlying wxNSView has implemented NSTextInputClient protocol interface, but it is not complete. So I re-implemented the NSTextInputClient protocol interface in wxStyledTextCtrl to override the underlying implementation, and get the current editor cursor position in the protocol interface method "firstRectForCharacterRange". The current code compiles and runs fine in my environment (macOS 15.5, c++20).

Issue #25503, opened 2025-06-07
wxUSE_HOTKEY defaults to OFF for wxGTK with configure, but to ON with CMake


phileimer:
### Build System Used

I build wxWidgets using:

- [X] CMake
- [X] configure
- [ ] mingw32-make with makefile.gcc
- [ ] MSBuild (Microsoft Visual Studio solution file)
- [ ] nmake with makefile.vc
- [ ] Xcode

`cmake -B BUILD -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr -DwxUSE_LIBLZMA=on`

`../configure --prefix=/opt/wx --with-liblzma`

### Description
Moving from 'configure' to 'CMake' build gives incompatibility, preventing existing applications from working without rebuilding.
They fail with 'different size symbols':
```
./BUILD/src/wxEphe: Symbol `_ZTV11wxAnyButton' has different size in shared object, consider re-linking
./BUILD/src/wxEphe: Symbol `_ZTV8wxButton' has different size in shared object, consider re-linking
./BUILD/src/wxEphe: Symbol `_ZTV12wxButtonBase' has different size in shared object, consider re-linking
./BUILD/src/wxEphe: Symbol `_ZTV11wxStaticBox' has different size in shared object, consider re-linking
./BUILD/src/wxEphe: Symbol `_ZTV23wxDatePickerCtrlGeneric' has different size in shared object, consider re-linking
./BUILD/src/wxEphe: Symbol `_ZTV10wxCheckBox' has different size in shared object, consider re-linking
./BUILD/src/wxEphe: Symbol `_ZTV7wxFrame' has different size in shared object, consider re-linking
./BUILD/src/wxEphe: Symbol `_ZTV9wxControl' has different size in shared object, consider re-linking
./BUILD/src/wxEphe: Symbol `_ZTV7wxPanel' has different size in shared object, consider re-linking
./BUILD/src/wxEphe: Symbol `_ZTV16wxTimePickerCtrl' has different size in shared object, consider re-linking
./BUILD/src/wxEphe: Symbol `_ZTV16wxDatePickerCtrl' has different size in shared object, consider re-linking
./BUILD/src/wxEphe: Symbol `_ZTV6wxGrid' has different size in shared object, consider re-linking
./BUILD/src/wxEphe: Symbol `_ZTV17wxMDIClientWindow' has different size in shared object, consider re-linking
[...]

PR #25643, opened 2025-07-15
CMake improvements

MaartenBent:
Fixes for #25466, #25503 and #25592


vadz:

pauldmccarthy:
Is the use of `wxUSE_MSVC_NAMING`  at [build/cmake/functions.cmake:485](https://github.com/MaartenBent/wxWidgets/blob/f5ee95f88c6c9806a1b938cf39f4e2ffea0ceebc/build/cmake/functions.cmake#L485) intentional? Should it be `WIN32_MSVC_NAMING`? This change has caused the conda-forge wxwidgets build to break, as it was expecting the `.dll` files to be installed into `lib\vc_x64_dll\`, but as of 3.3.1 they are being installed into `bin\vc_x64_dll\` ..

MaartenBent:
That is indeed a stupid mistake 🤦‍♂️. It should be `WIN32_MSVC_NAMING`.

pauldmccarthy:
No problem! We can work around it for 3.3.1, and revert if/when it's fixed here. FWIW I've opened a PR here (feel free to ignore/close if you intend to fix it by other means):

 https://github.com/wxWidgets/wxWidgets/pull/25655

PR #25648, opened 2025-07-19
Add hardware query functions

Blake-Madden:
My updated attempt of a `wxSystemHardwareInfo` set of functions. I applied all feedback from an older draft. Any other thoughts?


Blake-Madden:
Would it be better to use `wxULongLong` instead of `wxMemorySize` (aka, `wxLongLong`) for the memory function return types?

I only use `wxMemorySize` because `wxGetFreeMemory()` does that. However, `wxULongLong` would be more forward thinking and make it compatible with `wxFileName::GetHumanReadableSize()`.

PR #25649, opened 2025-07-19
Fix a few issues in the translations

Blake-Madden:
Fix malformed pipes, incorrect numbers, wrong translation, missing trailing punctuations, missing trailing spaces


vadz:
Nice, thanks!

PR #25655, opened 2025-07-22
BF: Use WIN32_MSVC_NAMING and not wxUSE_MSVC_NAMING

pauldmccarthy:
This PR fixes a small typo in `build/cmake/functions.cmake` which causes `.dll` files to be installed into the wrong location in some circumstances. See https://github.com/wxWidgets/wxWidgets/pull/25643#issuecomment-3103610781


MaartenBent (approved):
Thanks for the PR

vadz:
Thanks indeed, the PR may well be trivial, but finding and reporting the bug is still very much appreciated!

I'll push your fix soon.

PR #25662, opened 2025-07-25
Remove broken assert in wxSocketBase::SetFlags.

jherning:
This does not correctly check for incompatible flag combinations leading to erroneous failed assertions.


vadz:
I'm going to merge #25665, replacing this one. Please reopen/comment if you still have problems.

PR #25666, opened 2025-07-28
CMake improvements for consuming it via `add_subdirectory()`

vadz:
This started as [this thread](https://groups.google.com/g/wx-dev/c/XvuI422sGE8) but grew out of it to contain several QoL improvements for consuming the library from a superproject using CMake.

cc @MaartenBent


vadz:
I've added another commit fixing a problem with using built-in libpng when libpng-dev package is installed under Linux: without this, the system `pnglibconf.h` was used, resulting in link errors due to the mismatch between the `png_xxx` symbols declared in it and `wx_png_xxx` symbols actually defined in `wxpng` library.

I've tried being smart and allow calling the existing `wx_lib_include_directories()` with an optinal `BEFORE` keyword but ran into problems doing this, so I've finally given up and just added a separate macro (why is it a macro and not a function, BTW?). Please let me know if you think this could/should be improved in some way.

build/cmake/functions.cmake
```diff
@@ -706,6 +706,17 @@ function(wx_add_thirdparty_library var_name lib_name help_str)
 endfunction()
 
 function(wx_print_thirdparty_library_summary)
+  # Avoid printing out the message if we're being reconfigured and nothing has
+  # changed since the previous run, so check if the current summary differs
+  # from the cached value.
+  foreach(entry IN LISTS wxTHIRD_PARTY_LIBRARIES)
+    # TODO-CMAKE-VERSION: Use list(JOIN) with CMake 3.12+.
+    string(APPEND wxTHIRD_PARTY_SUMMARY_NOW "${entry}-")
```
MaartenBent: This doesn't work. `wxTHIRD_PARTY_LIBRARIES` is a list where the uneven entries are the variable name and the even entries the description, like:
```
0 - wxUSE_REGEX
1 - enable support for wxRegEx class
2 - wxUSE_ZLIB
3 - use zlib for LZW compression
4 - wxUSE_EXPAT
5 - use expat for XML parsing
...
```

I think you want the cached string to look like:
 ```
wxUSE_REGEX=builtin wxUSE_ZLIB=builtin wxUSE_EXPAT=builtin wxUSE_LIBJPEG=builtin wxUSE_LIBPNG=builtin wxUSE_LIBTIFF=builtin wxUSE_LIBWEBP=builtin wxUSE_NANOSVG=builtin wxUSE_LIBLZMA=OFF wxUSE_WEBVIEW=ON 
```

Then you'll need to do something similar as is done later in this function, like this:
```cmake
  set(var_name)
  foreach(entry IN LISTS wxTHIRD_PARTY_LIBRARIES)
    if(NOT var_name)
      set(var_name ${entry})
    else()
      string(APPEND wxTHIRD_PARTY_SUMMARY_NOW "${var_name}=${${var_name}} ")
      set(var_name)
    endif()
  endforeach()
```

But now the function iterates `wxTHIRD_PARTY_LIBRARIES` twice. Maybe it is simpler to cache the value of `${message}` and compare against that. At the expense of always creating the full print.

build/cmake/functions.cmake
```diff
@@ -706,6 +706,17 @@ function(wx_add_thirdparty_library var_name lib_name help_str)
...
+  endforeach()
+
+  if(NOT "${wxTHIRD_PARTY_SUMMARY_NOW}" STREQUAL "$CACHE{wxTHIRD_PARTY_SUMMARY}")
```
MaartenBent: `$CACHE{VAR}` is added in version 3.13. I don't think it is needed here and you can just use `"${wxTHIRD_PARTY_SUMMARY}"`

build/cmake/main.cmake
```diff
@@ -73,6 +73,20 @@ endif()
...
 
+# Avoid printing out the message if we're being reconfigured and nothing has
+# changed since the previous run, so check if the current summary differs from
+# the cached value.
+set(wxSUMMARY_NOW
+    "${CMAKE_SYSTEM_NAME}-${wxVERSION}-${wxREQUIRED_OS_DESC}-"
+    "${wxBUILD_TOOLKIT}-${wxTOOLKIT_VERSION}-${wxTOOLKIT_EXTRA}-"
+    "${wxBUILD_MONOLITHIC}-${wxBUILD_SHARED}-${wxBUILD_COMPATIBILITY}-"
+)
+if("${wxSUMMARY_NOW}" STREQUAL "$CACHE{wxSUMMARY}")
```
MaartenBent: See `$CACHE{wxTHIRD_PARTY_SUMMARY}` comment

build/cmake/functions.cmake
```diff
@@ -1025,6 +1048,15 @@ function(wx_option name desc)
     endif()
 endfunction()
 
+# A convenient wrapper for wx_option() for an option set to AUTO by default but
+# also allowing ON and OFF values, with AUTO meaning that the option is set to
+# ON if possible (e.g. required support for it is detected) and turned OFF with
+# just a warning otherwise, while ON means that an error is given if it can't
+# be enabled.
+function(wx_option_auto name desc)
+  wx_option(${name} ${desc} AUTO STRINGS ON OFF AUTO)
```
MaartenBent: Our CMake files are mostly indented with 4 spaces.

build/cmake/locale/CMakeLists.txt
```diff
@@ -1,16 +1,20 @@
-if(wxBUILD_LOCALES STREQUAL "AUTO")
-    find_package(Gettext QUIET)
-elseif(wxBUILD_LOCALES)
-    find_package(Gettext REQUIRED)
+if(wxBUILD_LOCALES)
+    find_package(Gettext)
```
MaartenBent: Why did you remove `QUIET` in the `AUTO` option? Now it pollutes the output every reconfigure with
```
Could NOT find Gettext (missing: GETTEXT_MSGMERGE_EXECUTABLE GETTEXT_MSGFMT_EXECUTABLE)
```

MaartenBent (changes requested):
> why is it a macro and not a function, BTW?

Functions are scoped, macros are not. With a function, `wxMONO_INCLUDE_DIRS` would be reset every time.

> Please let me know if you think this could/should be improved in some way.

I'm fine with this, but you could add a `BEFORE` parameter and use `cmake_parse_arguments`:

```cmake
wx_lib_include_directories(wxcore BEFORE ${${lib}_INCLUDE_DIRS})
```
```cmake
macro(wx_lib_include_directories name)
    cmake_parse_arguments(_INC_DIR "BEFORE" "" "" ${ARGN})
    if(wxBUILD_MONOLITHIC)
        list(APPEND wxMONO_INCLUDE_DIRS ${ARGN})
        set(wxMONO_INCLUDE_DIRS ${wxMONO_INCLUDE_DIRS} PARENT_SCOPE)
    else()
        if(_INC_DIR_BEFORE)
            target_include_directories(${name} BEFORE PRIVATE ${ARGN})
        else()
            target_include_directories(${name} PRIVATE ${ARGN})
        endif()
    endif()
endmacro()
```

build/cmake/functions.cmake
vadz: Fixed, thanks (I wasn't sure about the indentation to use initially as I've seen a mix of both 2 and 4, but I'll configure my editor to use 4 now).

build/cmake/functions.cmake
vadz: Thanks, I didn't notice this and I don't know why did I use `$CACHE{}` originally... I think I wanted to make sure I was using the original value of the variable, but it looks like it's not really necessary, so I've just removed it, thanks.

build/cmake/main.cmake
vadz: Removed `CACHE` too.

build/cmake/locale/CMakeLists.txt
vadz: Sorry, I wanted to simplify the code but didn't realize that it would start giving warnings in AUTO case when it's not available (I've only tested this here, where it is).

I've restored it now, I don't think using a temp variable and passing it to `find_package()` is better than doing it with just an `if()`, is it?

build/cmake/functions.cmake
vadz: Oops, thanks, I've totally missed this.

I've changed the code to build the summary cache as part of the first iteration (the code already iterated twice, I had added a third one), please let me know if you still see any problems with it.

vadz:
> I'm fine with this, but you could add a `BEFORE` parameter and use `cmake_parse_arguments`:

I _think_ I've tried something very similar to this and ran into problems, but I could have made some stupid mistake. Anyhow, if you're fine with the current version, let's leave it, but if you'd like to change it, please let me know or just do it as you prefer.

I've addressed your other comments in the new commits, thanks for the review!

vadz:
Oops, my zlib fix actually doesn't work because `HAVE_UNISTD_H` is not defined yet when `build/cmake/lib/zlib.cmake` is included from `build/cmake/lib/CMakeLists.txt` because it's only done in `build/cmake/setup.cmake` which is processed later.

I couldn't see the right way to fix this, so I've made #25677 instead and will omit [e898ab9](https://github.com/wxWidgets/wxWidgets/pull/25666/commits/e898ab907187fe4d6763c9807ef54875c28ee4db) when merging this if the CI builds there pass.

vadz:
> I couldn't see the right way to fix this, so I've made #25677 instead

This didn't work, so I've returned to this and fixed it in a different way. I've also rebased the branch and added one other minor fix (last commit).

Please let me know if you think this can be merged now.

MaartenBent (approved):
Looks good now!

PR #25669, opened 2025-07-29
Build wxNet without wxSocket if wxWebRequest is used

vadz:
I'm thinking about deprecating `wxSocket` eventually and while it's not going to be done yet, I'd like to be able to build without it — but still with `wxWebRequest`. And this PR makes this work.

### DIFF
--- a/build/cmake/lib/CMakeLists.txt
+++ b/build/cmake/lib/CMakeLists.txt
@@ -45,7 +45,14 @@ endmacro()
 
 # Define base libraries
 set(LIBS base)
-add_opt_lib(net wxUSE_SOCKETS)
+
+# wxNet should be built if either wxUSE_SOCKETS or wxUSE_WEBREQUEST is on.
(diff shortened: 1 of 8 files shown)


imciner2:
> I'm thinking about deprecating wxSocket eventually

Would this then also deprecate `wxSocketServer`? We do use that in KiCad for an IPC mechanism to send messages between our two main frames when people open then separately. We have been talking about if we need to replace it, but haven't put much effort into that yet.

vadz:
_If_ we deprecate `wxSocket`, it would mean deprecating both `wxSocketServer` and `wxSocketClient`, but this is not about happen soon (i.e. not in 3.4). It's true that `wxSocketServer` is probably more useful than `wxSocketClient`, but this code is very old and in spite of all the simplifications can't be improved much further because it tries to support many different socket use models in a single class which is just a crazy thing to do and makes it unmaintainable.

Issue #25307, opened 2025-04-08
configurable timeout in wxWebRequest


mariuszste:
### Description

It would be useful if we could configure a custom timeout when using wxWebRequest

Things to consider:
Connect timeout vs max total time

Curl has both, I'm not sure about other platforms

PR #25672, opened 2025-07-30
Adding configurable timeouts to wxWebRequest

BrianDelalex:
Summary
----------------------------------------------------------------
As discussed in [issue 25307](https://github.com/wxWidgets/wxWidgets/issues/25307). This pr implement configurable timeout for wxWebRequest.

Implementation
--------------------
Adding two functions to wxWebRequestBase, one to set the timeouts concerning the connection phase and another one to set the timeouts for the whole request.
There are the prototypes for the functions:
```
void SetConnectionTimeouts(long resolveTimeoutMs, long connectTimeoutMs);
void SetTimeouts(long resolveTimeoutMs, long connectTimeoutMs, long sendTimeoutMs, long receiveTimeoutMs);
```
As those two functions have been added to wxWebRequestBase we can use it both on wxWebRequest and wxWebRequestSync.

Those functions have been also added to wxWebRequestImpl as pure virtual function.

Those two functions are implemented for wxWebRequestCURL (Linux) and wxWebRequestWinHTTP (Windows).
As the functions are defined as pure virtual in wxWebRequestImpl we must declare it for each class inheriting from it so, for wxWebRequestURLSession (mac) the functions are defined but only call a fail message to signal that the feature is no implemented.

Documentation
-------------------
The functions have been added to the documentation for the two classes concerned and mentioned the OS specificity to take into account when calling one of those.


BrianDelalex:
I had to close the pr it look like I've done some mistake when pushing (extra commit).

vadz:
Note: you can force push (`git push --force`) into your branch if you made a mistake. This should be avoided once somebody starts looking at it, but is acceptable if you do it immediately.

PR #25673, opened 2025-07-30
Adding configurable timeout for wxWebRequest.

BrianDelalex:
Summary
----------------------------------------------------------------
As discussed in [issue 25307](https://github.com/wxWidgets/wxWidgets/issues/25307). This pr implement configurable timeout for wxWebRequest.

Implementation
--------------------
Adding two functions to wxWebRequestBase, one to set the timeouts concerning the connection phase and another one to set the timeouts for the whole request.
There are the prototypes for the functions:
```
void SetConnectionTimeouts(long resolveTimeoutMs, long connectTimeoutMs);
void SetTimeouts(long resolveTimeoutMs, long connectTimeoutMs, long sendTimeoutMs, long receiveTimeoutMs);
```
As those two functions have been added to wxWebRequestBase we can use it both on wxWebRequest and wxWebRequestSync.

Those functions have been also added to wxWebRequestImpl as pure virtual function.

Those two functions are implemented for wxWebRequestCURL (Linux) and wxWebRequestWinHTTP (Windows).
As the functions are defined as pure virtual in wxWebRequestImpl we must declare it for each class inheriting from it so, for wxWebRequestURLSession (mac) the functions are defined but only call a fail message to signal that the feature is not implemented.

Documentation
-------------------
The functions have been added to the documentation for the two classes concerned and mentioned the OS specificity to take into account when calling one of those.

### DIFF
--- a/include/wx/msw/private/webrequest_winhttp.h
+++ b/include/wx/msw/private/webrequest_winhttp.h
@@ -85,6 +85,8 @@ public:
 
     void Start() override;
 
+    void SetTimeouts(long connectionTimeoutMs, long dataTimeoutMs) override;
+
     wxWebResponseImplPtr GetResponse() const override
         { return m_response; }
 
@@ -120,6 +122,7 @@ private:
     // returned later and this argument must be null.
     wxNODISCARD Result DoWriteData(DWORD* bytesWritten = nullptr);
 
+    void DoSetTimeouts();
 
     wxWebSessionWinHTTP& m_sessionImpl;
     wxString m_url;
@@ -139,6 +142,12 @@ private:
     // Proxy credentials (if any) are stored in the session, but we need store
     // the same flag for them as for the server credentials here.
     bool m_tryProxyCredentials = false;
+
+    // Store timeouts, in order to set these before sending request.
+    bool m_isTimeoutsSet = false;
+    long m_connectionTimeoutMs = 60000;
+    long m_dataTimeoutMs = 30000;
+
 
 
     wxNODISCARD Result SendRequest();
--- a/include/wx/msw/private/webrequest_winhttp.h
+++ b/include/wx/msw/private/webrequest_winhttp.h
@@ -144,9 +144,8 @@ private:
     bool m_tryProxyCredentials = false;
 
     // Store timeouts, in order to set these before sending request.
-    bool m_isTimeoutsSet = false;
-    long m_connectionTimeoutMs = 60000;
-    long m_dataTimeoutMs = 30000;
+    long m_connectionTimeoutMs = wxWebRequest::Timeout_Default;
+    long m_dataTimeoutMs = wxWebRequest::Timeout_Default;
 
 
 
--- a/interface/wx/webrequest.h
+++ b/interface/wx/webrequest.h
@@ -228,6 +228,21 @@ public:
     };
 
     /**
+        Constant timeout values. Can be use as argument when calling wxWebRequest::SetTimeouts.
+    */
+    enum Timeout
+    {
+        /**
+            Use the default value for the current timeout and implementation.
+        */
+        Timeout_Default,
+        /**
+            Set the timeout to infinite (No timeout).
+        */
+        Timeout_Infinite
+    };
+
+    /**
         Default constructor creates an invalid object.
 
         Initialize it by assigning wxWebSession::CreateRequest() to it before
@@ -423,22 +438,26 @@ public:
     void SetStorage(Storage storage);
 
     /**
-        Sets the timeouts in milliseconds for the webrequest.
+        Set the timeouts for the connection and total request time.
 
         @param connectionTimeoutMs
-            Maximum time in milliseconds allowed for name resolution and connection to the server.
+            Maximum time in milliseconds allowed for connection to the server, including its name resolution.
         @param dataTimeoutMs
-            Maximum time in milliseconds allowed to send and to receive data.
+            Maximum time in milliseconds allowed for the remaining operations (post connection).
 
-        @note On Linux those parameters can't be distinguish. They are added to form one value used to set
-        the maximum time in seconds that you allow the entire transfer operation to take. The whole thing, from start to end.
+        @note Use wxWebRequest::Timeout_Default to set the timeout to the default value (default depends on implementation).
+        Use wxWebRequest::Timeout_Infinite to set an infinite timeout.\n
 
-        @remarks The default timeouts may vary depending on Implementation.
+        @note On Linux @c connectionTimeoutMs can't be set to infinite instead the timeout will be set to the maximal value that can be stored in a long.
+        @c connectionTimeoutMs and @c dataTimeoutMs are added to form one value used to set the maximum time in milliseconds that you allow the entire transfer operation to take. The whole thing, from start to end.
+
+        @remarks The default timeouts vary depending on implementation.
             For more details take a look at the natives function (@ref descriptions).
             - For WinHTTP backend, <a target=_new href="https://learn.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpsettimeouts">WinHttpSetTimeouts</a> function.
-            - For CURL backend, <a target=_new href="https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT_MS.html">curl_easy_setopt</a> function with CURLOPT_TIMEOUT_MS option.
+            - For CURL backend, <a target=_new href="https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT_MS.html">curl_easy_setopt</a> function with CURLOPT_TIMEOUT_MS and CURLOPT_CONNECTTIMEOUT_MS options.
             - For macOS backend, this is not implemented yet.
 
+        @see Timeout
     */
     void SetTimeouts(long connectionTimeoutMs, long dataTimeoutMs);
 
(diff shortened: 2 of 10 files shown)


vadz (commented):
I know we've discussed this a lot already, but I am still surprised by the proposed API. IMO the current `SetTimeouts()` is not very good because:

1. It half-duplicates `SetConnectionTimeouts()` which is very unusual and undesirable.
2. It has 4 parameters of the same type which makes it hard to distinguish between them.

I'd strongly recommend replacing it with something like `SetDataTimeout(long timeoutMs)` which would set the given timeout as both `nSendTimeout` and `nReceiveTimeout` with WinHTTP (I really don't see why would you need to use different values for different directions and a quick GitHub search shows that everybody sets them to the same value anyhow) and add the given timeout to the connection timeout with libcurl.

Alternatively, replace both of these functions with the single `SetTimeouts()`. But in this case I'd prefer it to take 2, and not 3, parameters: connection timeout and data timeout. For WinHTTP connection timeout would be used for each of resolve/connect timeouts, for libcurl it would be used as is. And the data timeout would be used for both send/receive WinHTTP timeouts and summed with the connection timeout (and maybe doubled?) for the total timeout with libcurl. This is probably the simplest solution if you really want to set everything.

Note that libcurl `CURLOPT_TIMEOUT` is still very different from WinHTTP send/receive timeouts as the latter apply to each request while the former applies to the entire operation which could involve multiple requests (e.g. with redirects) and this would need to be mentioned in the docs.

BrianDelalex:
> I know we've discussed this a lot already, but I am still surprised by the proposed API. IMO the current `SetTimeouts()` is not very good because:
> 
>     1. It half-duplicates `SetConnectionTimeouts()` which is very unusual and undesirable.
> 
>     2. It has 4 parameters of the same type which makes it hard to distinguish between them.
> 
> 
> I'd strongly recommend replacing it with something like `SetDataTimeout(long timeoutMs)` which would set the given timeout as both `nSendTimeout` and `nReceiveTimeout` with WinHTTP (I really don't see why would you need to use different values for different directions and a quick GitHub search shows that everybody sets them to the same value anyhow) and add the given timeout to the connection timeout with libcurl. 
>

I did it that way because WinHttpSetTimeouts don't give me an options to set only some timeouts. I have to set all at once or to store the connect and resolve timeout to put them back in a future call to SetDataTimeout and I would prefer not storing data here.
But I understand you concerns so I think I'll opt for the solution below ( a single SetTimeouts() with only 2 arguments).

> Alternatively, replace both of these functions with the single `SetTimeouts()`. But in this case I'd prefer it to take 2, and not 3, parameters: connection timeout and data timeout. For WinHTTP connection timeout would be used for each of resolve/connect timeouts, for libcurl it would be used as is. And the data timeout would be used for both send/receive WinHTTP timeouts and summed with the connection timeout (and maybe doubled?) for the total timeout with libcurl. This is probably the simplest solution if you really want to set everything.
> 
> Note that libcurl `CURLOPT_TIMEOUT` is still very different from WinHTTP send/receive timeouts as the latter apply to each request while the former applies to the entire operation which could involve multiple requests (e.g. with redirects) and this would need to be mentioned in the docs.

Thank you for your reply.

vadz:
> I did it that way because WinHttpSetTimeouts don't give me an options to set only some timeouts. I have to set all at once or to store the connect and resolve timeout to put them back in a future call to SetDataTimeout and I would prefer not storing data here.

FWIW I think that if we want to have multiple functions, we absolutely do need to store the already set timeouts in the implementation.

> But I understand you concerns so I think I'll opt for the solution below ( a single SetTimeouts() with only 2 arguments).

But this is a simpler and, IMHO, quite sufficient approach, so let's indeed settle on it.

Thanks!

BrianDelalex:
Finally, it's look like I'll be obliged to store the timeouts at least for the Windows implementation. As the request handle is initialized when wxWebRequestWinHTTP::execute() is called, the WinHttpSetTimeouts can't be called before it.

BrianDelalex:
Summary
-------------
As discussed yesterday, here is the new implementation 

Implementation
--------------------
The timeouts are configurable using one single function "void SetTimeout(long connectionTimeoutMs, long dataTimeoutMs);".

Adding the pure virtual function SetTimeouts to wxWebRequestImpl.
Implementing SetTimeouts to wxWebRequestBase.
Implementing SetTimeouts to wxWebRequestCurl.
Implementing SetTimeouts and DoSetTimeout to wxWebRequestWinHTTP.
Adding definition and declaration of SetTimeout to wxWebRequestURLSession. (Not implemented)

Documentation
-------------------
The function has been added to the documentation for the two classes concerned and mentioned the OS specificity to take into account when calling one of those.

include/wx/msw/private/webrequest_winhttp.h
```diff
@@ -140,6 +143,12 @@ class wxWebRequestWinHTTP : public wxWebRequestImpl
     // the same flag for them as for the server credentials here.
     bool m_tryProxyCredentials = false;
 
+    // Store timeouts, in order to set these before sending request.
+    bool m_isTimeoutsSet = false;
```
vadz: I'd call this `m_setTimeoutsCalled` which would be more grammatical ("is" and "timeouts" can't be used together). Or maybe I'd just use some special value (`0` if it doesn't have any special meaning for WinHTTP or `-1` maybe) for the timeouts by default and remove it completely.

include/wx/msw/private/webrequest_winhttp.h
```diff
@@ -140,6 +143,12 @@ class wxWebRequestWinHTTP : public wxWebRequestImpl
...
+    long m_connectionTimeoutMs = 60000;
```
vadz: Probably worth adding a comment saying that these are the values documented as WinHTTP defaults (if you switch to using 0 here, we'd still need these values in the code, so the comment would be needed there).

interface/wx/webrequest.h
```diff
@@ -422,6 +422,26 @@ class wxWebRequest
     */
     void SetStorage(Storage storage);
 
+    /**
+        Sets the timeouts in milliseconds for the webrequest.
```
vadz:
```suggestion
        Set the timeouts for the connection and total request time.
```

(the difference is "Set" instead of "Sets", for consistency with the other functions, and I think the fact that "ms" is mentioned in the parameters descriptions is enough).

interface/wx/webrequest.h
```diff
@@ -422,6 +422,26 @@ class wxWebRequest
...
+
+        @param connectionTimeoutMs
+            Maximum time in milliseconds allowed for name resolution and connection to the server.
```
vadz:
```suggestion
            Maximum time in milliseconds allowed for connection to the server, including its name resolution.
```

interface/wx/webrequest.h
```diff
@@ -422,6 +422,26 @@ class wxWebRequest
...
+        @param dataTimeoutMs
+            Maximum time in milliseconds allowed to send and to receive data.
+
+        @note On Linux those parameters can't be distinguish. They are added to form one value used to set
```
vadz: This is not quite correct if the remark in CURL backend code is acted on.

interface/wx/webrequest.h
```diff
@@ -422,6 +422,26 @@ class wxWebRequest
...
+        the maximum time in seconds that you allow the entire transfer operation to take. The whole thing, from start to end.
+
+        @remarks The default timeouts may vary depending on Implementation.
```
vadz:
```suggestion
        @remarks The default timeouts vary depending on implementation.
```

src/common/webrequest_curl.cpp
```diff
@@ -524,6 +524,13 @@ void wxWebRequestCURL::Start()
     StartRequest();
 }
 
+void wxWebRequestCURL::SetTimeouts(long connectionTimeoutMs,
+                                   long dataTimeoutMs)
+{
+    long timeout = connectionTimeoutMs + dataTimeoutMs;
+    wxCURLSetOpt(m_handle, CURLOPT_TIMEOUT_MS, timeout);
```
vadz: We need to set `CURLOPT_CONNECTTIMEOUT_MS` too, the documentation even mentions it, but the code doesn't do it for some reason?

src/osx/webrequest_urlsession.mm
```diff
@@ -327,6 +327,12 @@ void Wait()
     [m_task resume];
 }
 
+void wxWebRequestURLSession::SetTimeouts(long WXUNUSED(connectionTimeoutMs),
+                                         long WXUNUSED(dataTimeoutMs))
+{
+    wxFAIL_MSG(wxT("Unimplemented"));
```
vadz: It would be better to just do nothing here, otherwise you'd need to use `#ifndef __WXOSX__` in the application code to be able to call this function — and the effect would be the same as just not doing anything under Mac, so let's not do anything here.

IOW:
```suggestion
```

vadz (commented):
Sorry, I should have mentioned this before, but I think it makes sense to allow using 0 for one of the timeouts (mostly the second one) to allow only setting the other one and leaving the one for which 0 is used at the default value.

I.e. I think it could be quite useful to set connection timeout to something relatively short, but leave the default timeout unchanged and, in particular, not set it when using libcurl.

BrianDelalex:
> Sorry, I should have mentioned this before, but I think it makes sense to allow using 0 for one of the timeouts (mostly the second one) to allow only setting the other one and leaving the one for which 0 is used at the default value.
> 
> I.e. I think it could be quite useful to set connection timeout to something relatively short, but leave the default timeout unchanged and, in particular, not set it when using libcurl.

Great idea. I'll work on this and also apply all the changes that you suggested.
Thank you.

vadz:
Sorry for the delay with reviewing the latest version, I've finally done it and couldn't prevent myself from tweaking it a bit, please see the commit which I've pushed directly into your branch.

If you don't have any objections to the changes there, I'll (squash) merge this soon.

Thanks again!

BrianDelalex (commented):
I don't have any objections to the changes.

Thanks!

PR #25677, opened 2025-07-31
Update zlib submodule to fix CMake build with built-in library

vadz:
Ensure that unistd.h gets included if it was detected by wx configure to avoid errors when using standard functions, such as lseek(), declared there.


vadz:
OK, this doesn't work, time for a plan B :-(

Issue #25678, opened 2025-08-02
[Dark Mode] Tab row background has wrong color, most borders are invisible/gone


RippeR37:
#### Bug description:

When switching to a dark mode, if application uses a notebook, the background of the row with all the tabs you can switch between (not the tabs itself) has a wrong color.
Furthermore, The notebook's pages don't have border anymore. Only the top border (between the row with tabs and page itself) has a border as well as the tab (the ones you can click on to switch between them) have correct borders.

#### Expected vs observed behaviour:

In light mode, the area next to switchable tabs has the same color as the wxNotebook's background (so you can set it). In dark mode, setting the notebook's background doesn't change color of this area at all. Color should change also in dark mode to make it customizable.
Most of the borders are gone which results in a weird look. Other elements such as wxStaticBox or wxDataViewListCtrl have their border correct.

Image showing the problematic parts:
<img width="1897" height="706" alt="Image" src="https://github.com/user-attachments/assets/d92f0b60-77f5-4c6e-8928-6644d52250f4" />

### Reproduction code

```cpp
auto* notebook = new wxNotebook(...);
notebook->SetBackgroundColour(wxColour{0x00, 0xFF, 0x00});

// and then run this in both Light and Dark mode and you will notice
// that in Light green is also in the tab bar's background and in Dark
// it's only in the 'page' part of the notebook
```

### Platform and version information

- wxWidgets version: v3.3.1
- wxWidgets port: wxMSW
- OS: Windows 10 22H2 (19045.6093)

### Other interesting data

I noticed that I can adjust color of that background by overriding color returned for `wxSYS_COLOUR_BTNFACE` in `wxDarkModeSettings::GetColour()`. This is however "global" and not per notebook so severely limits customization options.

vadz:
Yes, I agree that the background colour is wrong. This should be easy to fix, I'll try to do it when I have time but if anybody else wants to submit a PR doing it, it would be welcome.

For the borders, I'm less sure what's going on. I can't see the problem in the notebook sample (unlike with the above one), can you make a [minimal patch](https://www.wxwidgets.org/develop/how-to-submit-patches/) to the notebook sample showing it please?

RippeR37:
> This should be easy to fix, I'll try to do it when I have time but if anybody else wants to submit a PR doing it, it would be welcome.

I could try to look around for this, but I don't know this codebase or what to look for, so I'd need some tips where to even start and what to look for.

[...]

PR #25679, opened 2025-08-02
[#25678][MSW] Use wxNotebook's background color for tab background if specified

RippeR37:
This commit updates the logic that draws wxNotebook on MSW in Dark mode. Previously, the row with tabs was cleared with the background color set as `wxSYS_COLOUR_BTNFACE` which may looked okay by default, but didn't allow for customizing this as with the Light theme. In Light theme, the color of backgrounds of both "page area" and "tab row" can be overwritten with calling `wxNotebook::SetBackgroundColor()` which didn't work in Dark mode.

With this commit, if the background color is specified, it will be used for that tab row area as well, which should match behavior with Light mode.

### DIFF
--- a/src/msw/notebook.cpp
+++ b/src/msw/notebook.cpp
@@ -1288,7 +1288,10 @@ void wxNotebook::MSWNotebookPaint(wxDC& dc)
         rectTabArea.SetRight(sizeWindow.x);
     else
         rectTabArea.SetBottom(sizeWindow.y);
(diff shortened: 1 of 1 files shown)


RippeR37:
Okay, it seems to need a further fix. This is also overriding the color of the 'active tab' on Dark mode, while on Light mode it works a bit different (active tab has its own color, etc.).

Differences:
obraz

PR #25684, opened 2025-08-05
Fix MSVS debug visualizer of wxStrings

vslavik:
Fix the visualizer to display full Unicode string, instead of just its first character.

I'm not surprised the visualizer misbehaved - it used the `sb` format specifier, [which is for `const char*` string](https://learn.microsoft.com/en-us/visualstudio/debugger/format-specifiers-in-cpp?view=vs-2022). I'm surprised it (I think) worked until relatively recent MSVS versions.

Since wx uses Unicode strings internally for some time now, it can be simply changed into `sub`.

Harmless for 3.2 branch too.

### DIFF
--- a/misc/msvc/wxWidgets.natvis
+++ b/misc/msvc/wxWidgets.natvis
@@ -12,11 +12,12 @@ More information can be found at https://learn.microsoft.com/en-us/visualstudio/
     <!-- We want to avoid showing 'L' before the string, this is useless and
          makes the display less readable. But we still do want to put quotes
          around the string to show that it is, in fact, a string. So we use
-         "sb" qualifier to get rid of L"..." around the string and then add
-		 the quotes back. This also has a (nice) side effect of not doubling
-		 the backslashes inside the string. -->
+         "sub" or "s8b" qualifier to get rid of L"..." around the string and
+         then add the quotes back. This also has a (nice) side effect of not
+         doubling the backslashes inside the string. -->
     <Type Name="wxString">
-        <DisplayString>"{m_impl,sb}"</DisplayString>
+        <DisplayString Condition="sizeof(m_impl[0])==2">"{m_impl,sub}"</DisplayString> <!-- wxUSE_UNICODE_WCHAR -->
+        <DisplayString Condition="sizeof(m_impl[0])==1">"{m_impl,s8b}"</DisplayString> <!-- wxUSE_UNICODE_UTF8 -->
         <StringView>m_impl</StringView>
     </Type>
 


Randalphwa:
You are assuming that wxUSE_UNICODE_UTF8 is set to 0 -- if it's set to 1, then the internal representation is UTF8 strings. So yes, you will fix it for UTF16 and break it for UTF8 -- essentially the opposite of the way it is broken now. That said, I'm not aware of a way to place a pre-compiler condition within a .natvis file. You might at least add a comment that it requires wxUSE_UNICODE_UTF8 be set to 0.

vslavik:
> You are assuming that wxUSE_UNICODE_UTF8 is set to 0

Which is the default (on only choice w/o performance penalty) on MSW, yes. I'm pretty sure use of `wxUSE_UNICODE_UTF8` *on MSW* rounds to "nobody".

Randalphwa:
If no one was ever going to use it, then the dev team never would have made the effort to add it. The performance _gain_ is significant for wxXML processing, XRC processing, Scintilla processing, and for non-English apps that use wxWidgets to translate their strings. I fail to see why it would be a problem for you to add the comment about wxUSE_UNICODE_UTF8  in the .natvis file to avoid any potential confusion for those who do use UTF8 in their apps to enhance string processing performance.

RobertRoeb:
I planned to (test to) use a UTF8 build also under Windows. If it doesn't work, then I will need use UTF16, obviously.

vadz:
Some people do use UTF-8 build under Windows and it's certainly supposed to work.

However many more people use the default (UTF-16) build.

Unfortunately I don't think it's possible to somehow dynamically select the visualizer to use depending on the build settings, so we probably should apply this and tell people using UTF-8 to remove "u" themselves. Does anybody see a better solution?

vslavik:
> Unfortunately I don't think it's possible to somehow dynamically select the visualizer 

Ye of little faith :)

> Does anybody see a better solution?

Yes. If you think use of UTF-8 builds is non-marginal, then there are two ways to do it:

1. Dynamically choose the visualizer at runtime through some clever (if I may say so myself) hacking of [conditional visualization](https://learn.microsoft.com/en-us/visualstudio/debugger/create-custom-views-of-native-objects?view=vs-2022#BKMK_Condition_attribute). I think this doesn't even qualify as an abuse of the mechanism and while it ought to have some runtime cost, it's presumably trivial. This is what I did in the now-force-pushed 1bcc28c.

2. Given that natvis [can be included in the PDB](https://learn.microsoft.com/en-us/cpp/build/reference/natvis-add-natvis-to-pdb?view=msvc-170) these days, it would be possible to conditionally pick the correct one at compile-time. But that would be more work.

P.S. Note that `sb` was wrong for `wxUSE_UNICODE_UTF8` too, albeit less dramatically so.

vadz:
> > Does anybody see a better solution?
> 
> Yes. If you think use of UTF-8 builds is non-marginal, then there are two ways to do it:
> 
>     1. Dynamically choose the visualizer at runtime through some clever (if I may say so myself)

It's pretty clever indeed, thanks!

I'm going to merge this soon unless there are other objections, thanks again!

PR #25688, opened 2025-08-08
Improve support for iOS, in particular iPhones and allow command line cross-compilation on Silicon Macs

RobertRoeb:
Improve support for iOS, in particular iPhones

- Allow compiling for iPhone simulator from command line by introducing the new flag --enable-iphonesimulator
- Implemented wxNotebook to be used as the main screen for the app with button at the bottom
- Implemented, enabled or improved wxBitmapBundle, wxStaticBitmap, wxBitmapButton, wxChoice, wxChoiceBook, wxBookCtrl
- Implemented wxScrolledWindow using native UIScrollView to enable native scrolling with scrolling physics
- Implemented label next to wxCheckbox
- Enable wxHtmlWindow
- Enable wxDataViewCtrl
- Added wxART_PREV_SCREEN and use it as "back" button emitting wxCloseEvent since frames don't have any other close button in iOS
- Enable wxVListBox

wxIPhone

### DIFF
--- a/include/wx/dc.h
+++ b/include/wx/dc.h
@@ -169,8 +169,10 @@ public:
 class WXDLLIMPEXP_CORE wxDCImpl: public wxObject
 {
 public:
-    wxDCImpl( wxDC *owner );
-    virtual ~wxDCImpl();
+    wxDCImpl( wxDC *owner ) : m_owner(owner)
+    {
+    }
+    virtual ~wxDCImpl() = default;
 
     wxDC *GetOwner() const { return m_owner; }
     void SetOwner(wxDC* owner) { m_owner = owner; }
@@ -237,6 +239,17 @@ public:
     virtual wxSize LogicalToDeviceRel(int x, int y) const;
 
     // bounding box
+
+    void DisableAutomaticBoundingBoxUpdates()
+    {
+        m_isBBoxValid = false;
+        m_updateBBox = false;
+    }
+
+    bool AreAutomaticBoundingBoxUpdatesEnabled() const
+    {
+        return m_updateBBox;
+    }
 
     virtual void CalcBoundingBox(wxCoord x, wxCoord y)
     {
@@ -634,67 +647,96 @@ protected:
 
 
     // window on which the DC draws or nullptr
-    wxWindow   *m_window;
+    wxWindow   *m_window = nullptr;
 
     // flags
-    bool m_colour:1;
-    bool m_ok:1;
-    bool m_clipping:1;
-    bool m_isInteractive:1;
-    bool m_isBBoxValid:1;
+    bool m_colour = true;
+    bool m_ok = true;
+    bool m_clipping = false;
+    bool m_isInteractive = false;
+    bool m_isBBoxValid = false;
 
     // coordinate system variables
 
-    wxCoord m_logicalOriginX, m_logicalOriginY;
-    wxCoord m_deviceOriginX, m_deviceOriginY;           // Usually 0,0, can be change by user
+    wxCoord m_logicalOriginX = 0,
+            m_logicalOriginY = 0;
+    wxCoord m_deviceOriginX = 0,
+            m_deviceOriginY = 0;
 
-    wxCoord m_deviceLocalOriginX, m_deviceLocalOriginY; // non-zero if native top-left corner
-                                                        // is not at 0,0. This was the case under
-                                                        // Mac's GrafPorts (coordinate system
-                                                        // used toplevel window's origin) and
-                                                        // e.g. for Postscript, where the native
-                                                        // origin in the bottom left corner.
-    double m_logicalScaleX, m_logicalScaleY;
-    double m_userScaleX, m_userScaleY;
-    double m_scaleX, m_scaleY;  // calculated from logical scale and user scale
+    // Non-zero for Postscript where the native origin in the bottom left
+    // corner.
+    wxCoord m_deviceLocalOriginX = 0,
+            m_deviceLocalOriginY = 0;
 
-    int m_signX, m_signY;  // Used by SetAxisOrientation() to invert the axes
+    double m_logicalScaleX = 1.0,
+           m_logicalScaleY = 1.0;
+    double m_userScaleX = 1.0,
+           m_userScaleY = 1.0;
 
-    double m_contentScaleFactor; // used by high resolution displays (retina)
+    // Calculated from logical scale and user scale.
+    double m_scaleX = 1.0,
+           m_scaleY = 1.0;
+
+    // Used by SetAxisOrientation() to invert the axes.
+    int m_signX = 1,
+        m_signY = 1;
+
+    // > 1 on high resolution displays.
+    double m_contentScaleFactor = 1.0;
 
     // Pixel per mm in horizontal and vertical directions.
     //
     // These variables are computed on demand by GetMMToPX[xy]() functions,
     // don't access them directly other than for assigning to them.
-    mutable double m_mm_to_pix_x,
-                   m_mm_to_pix_y;
+    mutable double m_mm_to_pix_x = 0.0,
+                   m_mm_to_pix_y = 0.0;
 
     // bounding and clipping boxes
-    wxCoord m_minX, m_minY, m_maxX, m_maxY; // Bounding box is stored in device units.
-    wxCoord m_clipX1, m_clipY1, m_clipX2, m_clipY2;  // Some derived classes operate directly on clipping box given in logical units.
 
-    wxRasterOperationMode m_logicalFunction;
-    int m_backgroundMode;
-    wxMappingMode m_mappingMode;
+    // Bounding box is stored in device units.
+    wxCoord m_minX = 0,
+            m_minY = 0,
+            m_maxX = 0,
+            m_maxY = 0;
+
+    // Some derived classes use clipping box given in logical units directly.
+    wxCoord m_clipX1 = 0,
+            m_clipY1 = 0,
+            m_clipX2 = 0,
+            m_clipY2 = 0;
+
+    wxRasterOperationMode m_logicalFunction = wxCOPY;
+    int m_backgroundMode = wxBRUSHSTYLE_TRANSPARENT;
+    wxMappingMode m_mappingMode = wxMM_TEXT;
 
     wxPen             m_pen;
     wxBrush           m_brush;
     wxBrush           m_backgroundBrush;
-    wxColour          m_textForegroundColour;
-    wxColour          m_textBackgroundColour;
+    wxColour          m_textForegroundColour = *wxBLACK;
+    wxColour          m_textBackgroundColour = *wxWHITE;
     wxFont            m_font;
 
 #if wxUSE_PALETTE
     wxPalette         m_palette;
-    bool              m_hasCustomPalette;
+    bool              m_hasCustomPalette = false;
 #endif // wxUSE_PALETTE
 
 private:
     // Return the full DC area in logical coordinates.
     wxRect GetLogicalArea() const;
 
-    wxCoord m_devClipX1, m_devClipY1, m_devClipX2, m_devClipY2;  // For proper calculations of clipping box we need to store it in device units.
-    bool m_useDevClipCoords;
+    // For proper calculations of clipping box we need to store it in device units.
+    wxCoord m_devClipX1 = 0,
+            m_devClipY1 = 0,
+            m_devClipX2 = 0,
+            m_devClipY2 = 0;
+
+    bool m_useDevClipCoords = false;
+
+    // If true (the default), the bounding box is updated automatically by
+    // all drawing operations, otherwise it is not to save some time if it's
+    // not needed anyhow.
+    bool m_updateBBox = true;
 
     wxDECLARE_ABSTRACT_CLASS(wxDCImpl);
 };
@@ -985,6 +1027,11 @@ public:
         { m_pimpl->EndPage(); }
 
     // bounding box
+
+    void DisableAutomaticBoundingBoxUpdates()
+        { m_pimpl->DisableAutomaticBoundingBoxUpdates(); }
+    bool AreAutomaticBoundingBoxUpdatesEnabled() const
+        { return m_pimpl->AreAutomaticBoundingBoxUpdatesEnabled(); }
 
     void CalcBoundingBox(wxCoord x, wxCoord y)
         { m_pimpl->CalcBoundingBox(x,y); }
--- a/interface/wx/dc.h
+++ b/interface/wx/dc.h
@@ -1567,8 +1567,43 @@ public:
 
     /**
         @name Bounding box functions
+
+        By default, wxDC maintains the coordinates of the bounding box
+        containing all the drawing operations performed on it. This can be
+        useful to determine the area of the DC that was modified, for example
+        to optimize the redrawing of a window.
+
+        If the application code doesn't need this information, it can save some
+        time by calling DisableAutomaticBoundingBoxUpdates() to disable the
+        default behaviour. Note that even in this case, the bounding box can
+        still be updated by calling CalcBoundingBox() manually, but it won't be
+        done automatically by the drawing functions.
     */
     ///@{
+
+    /**
+        Disable automatic bounding box updates.
+
+        Such updates are enabled by default but can be disabled to make the
+        drawing functions slightly faster.
+
+        @see AreAutomaticBoundingBoxUpdatesEnabled()
+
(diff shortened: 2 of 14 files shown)


lanurmi (commented):
I'd like to point out that the commit message of the latest commit (64d0e4d)  and various previous ones does not follow the ["standard rules"](https://wxwidgets.org/develop/code-repository/) used in wxWidgets's git. That is, a short title, followed by an empty line, and a longer description.

(Also note that commit messages can be modified without redoing everything or creating a new PR.)

Wlodzimierz-ABX-Skiba:
Hi @RobertRoeb ! I'm unsure segmented buttons should be implemented as wxChoice. This does not look like wxChoice is supposed to look in general. I believe segmented buttons should be rather implemented as some new control with some future general implementation added later for other platforms. I would rather add this as separate PR since this is new feature rather than simple improvement. Regards, ABX

include/wx/bookctrl.h
```diff
@@ -437,7 +437,7 @@ typedef void (wxEvtHandler::*wxBookCtrlEventFunction)(wxBookCtrlEvent&);
     #define EVT_BOOKCTRL_PAGE_CHANGING(id, fn)     EVT_NOTEBOOK_PAGE_CHANGING(id, fn)
 #else
     // dedicated to Smartphones
-    #include "wx/choicebk.h"
+    // #include "wx/choicebk.h"
```
vadz: This doesn't look intentional.

configure.ac
```diff
@@ -425,6 +425,7 @@ WX_ARG_ONLY_WITH(directfb,      [  --with-directfb         use DirectFB], [wxUSE
 WX_ARG_ONLY_WITH(x11,           [  --with-x11              use X11], [wxUSE_X11="$withval" wxUSE_UNIVERSAL="yes" CACHE_X11=1 TOOLKIT_GIVEN=1])
 WX_ARG_ONLY_WITH(qt,            [  --with-qt               use Qt], [wxUSE_QT="$withval" CACHE_QT=1 TOOLKIT_GIVEN=1])
 WX_ARG_ENABLE(nanox,       [  --enable-nanox          use NanoX], wxUSE_NANOX)
+WX_ARG_ENABLE(iphonesimulator,  [  --enable-iphonesimulator   commpile for iPhone simulator], wxUSE_IPHONESIMULATOR)
```
vadz: Could we avoid using this option and just check if `host` matches `*apple-darwin_sim` instead? This would seem to be a simpler and more natural way to handle this than using artificially different (IIUC) `host` and `build` triplets in this case by specifying the OS version for only one but not the other.

include/wx/osx/iphone/chkconf.h
```diff
@@ -58,11 +58,6 @@
 #define wxUSE_TOOLTIPS 0
 #endif
 
-#if wxUSE_DATAVIEWCTRL
```
vadz: Does it actually work there? It really doesn't look so... And in this case this fragment should remain.

include/wx/osx/iphone/private.h
```diff
@@ -38,7 +38,7 @@ wxBitmapBundle WXDLLIMPEXP_CORE wxOSXCreateSystemBitmapBundle(const wxString& id
 class WXDLLIMPEXP_CORE wxWidgetIPhoneImpl : public wxWidgetImpl
 {
 public :
-    wxWidgetIPhoneImpl( wxWindowMac* peer , WXWidget w, int flags = 0 ) ;
+    wxWidgetIPhoneImpl( wxWindowMac* peer , WXWidget w, int flags = 0, void *c = NULL  ) ;
```
vadz: A comment explaining what `c` would be welcome but if not then at least let's give an actual name, e.g.

```suggestion
    wxWidgetIPhoneImpl( wxWindowMac* peer , WXWidget w, int flags = 0, void* controller = nullptr ) ;
```

Using some type different from `void*` would be great too...

include/wx/osx/iphone/private.h
```diff
@@ -124,8 +126,11 @@ public :
 
     virtual void        controlAction(void* sender, wxUint32 controlEvent, WX_UIEvent rawEvent);
     virtual void         controlTextDidChange();
+
+    void*               GetController() { return m_controller; }
```
vadz:
```suggestion
    void*               GetController() const { return m_controller; }
```

include/wx/osx/iphone/private.h
```diff
@@ -225,13 +230,12 @@ protected :
 
     @end
 
-    @interface wxUIView : UIView
+    @interface wxUIView : UIScrollView
```
vadz: I'm not sure what is `wxUIView` used exactly for, but it looks weird to make all views scrollable. Are you sure we need to do it?

include/wx/osx/iphone/private.h
```diff
@@ -225,13 +230,12 @@ protected :
...
     {
     }
 
     @end // wxUIView
 
-
```
vadz: It would be great to avoid whitespace-only changed intermingled with significant ones please.

include/wx/settings.h
```diff
@@ -140,7 +140,11 @@ enum wxSystemMetric
     wxSYS_DCLICK_MSEC,
     wxSYS_CARET_ON_MSEC,
     wxSYS_CARET_OFF_MSEC,
-    wxSYS_CARET_TIMEOUT_MSEC
+    wxSYS_CARET_TIMEOUT_MSEC,
+    wxSYS_DEVICE_AREA_TOP,
+    wxSYS_DEVICE_AREA_BOTTOM,
+    wxSYS_DEVICE_AREA_LEFT,
+    wxSYS_DEVICE_AREA_RIGHT
```
vadz: We really should add this to `wxDisplay` if it's not already covered by its `GetClientArea()` and not here, this is basically a legacy interface.

include/wx/simplebook.h
```diff
@@ -11,6 +11,7 @@
 #define _WX_SIMPLEBOOK_H_
 
 #include "wx/compositebookctrl.h"
+#include "wx/choicebk.h"
```
vadz: This definitely doesn't seem right.

art/prev_screen.xpm
vadz, line 1: We typically use PNGs or SVGs for new images, not XPM.

interface/wx/artprov.h
```diff
@@ -212,6 +214,7 @@ const char* wxART_WX_LOGO;
      @li @c wxART_CDROM
      @li @c wxART_REMOVABLE
      @li @c wxART_WX_LOGO (since 3.1.6)
+     @li @c wxART_PREV_SCREEN (since 3.1.6)
```
vadz:
```suggestion
     @li @c wxART_PREV_SCREEN (since 3.3.2)
```

vadz: Also, it would be nice to explain how is this one different from `wxART_GO_BACK`.

interface/wx/choice.h
```diff
@@ -15,6 +15,8 @@
     @beginStyleTable
     @style{wxCB_SORT}
            Sorts the entries alphabetically.
+    @style{wxCH_BUTTONS}
```
vadz: Shouldn't this be a style of `wxToolBar`? `wxChoice` really isn't supposed to look like this IMO.

src/osx/carbon/frame.cpp
```diff
@@ -60,6 +66,20 @@ bool wxFrame::Create(wxWindow *parent,
     if ( !wxTopLevelWindow::Create(parent, id, title, pos, size, style, name) )
         return false;
 
+#ifdef __WXOSX_IPHONE__
+    if (parent != NULL) {
```
vadz:
```suggestion
    if (parent != nullptr)
    {
```

src/osx/carbon/frame.cpp
```diff
@@ -60,6 +66,20 @@ bool wxFrame::Create(wxWindow *parent,
...
+        // We are on the next screen, provide a back button and title
+        wxToolBar *tb = CreateToolBar();
+        // tb->AddTool( wxID_CLOSE, wxEmptyString, wxArtProvider::GetBitmap(wxART_PREV_SCREEN) ); doesn't work for some reason
+        tb->AddTool( wxID_CLOSE, wxEmptyString, wxBitmap( prev_screen_xpm ) );
+        tb->AddStretchableSpace();  // doesn't yet stretch on iOS, so at least at two
```
vadz:
```suggestion
        tb->AddStretchableSpace();  // doesn't yet stretch on iOS, so at least add two
```

vadz: Also, maybe we should modify the function itself to make the stretchable spaces twice as wide?

src/osx/iphone/checkbox.mm
```diff
@@ -37,29 +38,43 @@ + (void)initialize
...
-    void SetValue( wxInt32 v )
+    void SetValue( wxInt32 v ) override
     {
         [m_control setOn:v != 0 animated:NO];
     }
+
+    void Move(int x, int y, int width, int height) override
+    {
+        wxWidgetIPhoneImpl::Move( x, y, width, height );
+        m_label->SetPosition( wxPoint(x + 65, y+5) );  // TODO: align properly
```
vadz: It's not just alignment, it's the horizontal offset too. We really should at least use `GetChar{Width,Height}()` here instead of hardcoding these values.

src/osx/iphone/choice.mm
```diff
@@ -119,6 +129,82 @@ void SetItem(int pos, const wxString& text) override
...
+    {
+        wxCFStringRef cftext(text);
+        [((wxUISegmentedControl*)m_osxView) insertSegmentWithTitle:cftext.AsNSString() atIndex:pos animated:NO];
+        [((wxUISegmentedControl*)m_osxView) setEnabled: YES forSegmentAtIndex: pos ];
+    }
+
+    size_t GetNumberOfItems() const override
+    {
+        int value = [((wxUISegmentedControl*)m_osxView) numberOfSegments];
+        wxLogMessage( "GetNumberOfItems()", value );
```
vadz:
```suggestion
```

src/osx/iphone/choice.mm
```diff
@@ -128,13 +214,21 @@ wxWindowID WXUNUSED(id),
...
+                                    long style,
                                     long WXUNUSED(extraStyle))
 {
     CGRect r = wxOSXGetFrameForControl( wxpeer, pos , size ) ;
-    wxUIPickerView* v = [[wxUIPickerView alloc] initWithFrame:r];
 
-    wxChoiceIPhoneImpl* c = new wxChoiceIPhoneImpl( wxpeer, v );
+    wxWidgetImplType *c = NULL;
+
+    if ((style & wxCH_BUTTONS) != 0) {
```
vadz:
```suggestion
    if ((style & wxCH_BUTTONS) != 0)
    {
```

src/osx/iphone/choice.mm
```diff
@@ -128,13 +214,21 @@ wxWindowID WXUNUSED(id),
...
+        wxUISegmentedControl* v = [[wxUISegmentedControl alloc] initWithFrame:r];
+        // [v setMomentary: NO];
+        c = new wxSegmentedChoiceIPhoneImpl( wxpeer, v );
+    } else {
```
vadz:
```suggestion
    }
    else
    { 
```

src/osx/iphone/dialog.mm
```diff
@@ -19,6 +19,7 @@
 #endif // WX_PRECOMP
 
 #include "wx/osx/private.h"
+#include "wx/log.h"
```
vadz:
```suggestion
```

src/osx/iphone/evtloop.mm
```diff
@@ -32,6 +32,7 @@
 #endif
 
 #include "wx/osx/private.h"
+#include "wx/log.h"
```
vadz:
```suggestion
```

src/osx/iphone/nonownedwnd.mm
```diff
@@ -13,6 +13,8 @@
 
 #include "wx/nonownedwnd.h"
 #include "wx/frame.h"
+#include "wx/settings.h"
+#include "wx/log.h"
```
vadz:
```suggestion
```

and remove `wxLogMessage()` calls below.

vadz (changes requested):
Sorry, I ran out of steam before the end but it's clear that while there are a lot of potentially useful things here, there are also potentially quite a few things to change, e.g. I'm not at all convinced that we want to represent `wxChoice` with segmented buttons.

It would be great to split this PR into independent changes in order to discuss them separately as discussing all of them here will probably become unmanageable pretty quickly.

TIA!

RobertRoeb:
Screenshots from major progress on drawing and scrolling support

wxVListBox
wxDataViewCtrl
SetTargetWindow
wxHtmlWindow

RobertRoeb:
>. I'm not at all convinced that we want to represent 
> `wxChoice` with segmented buttons.

The current wxChoice uses the picker control, which surely does not look like a MSW choice either. I can create a new wxSegmentedButton control with no implementation elsewhere. Then how should I implement wxBookCtrl? This control needs a segmented button on top, wxChoiceBook uses wxChoice, so this was an easy solution. Anyways, to get this further - shall I 
- add an iPhone specific wxSegmentedButtons
-  re-implement wxNotebook using it
-  cast wxBookCtrl to wxNotebook instead of wxChoiceBook on iPhone 
- implement wxNavigationCtrl using native control on iPhone and cast to wxNotebook with bar at the bottom on other platforms

configure.ac
RobertRoeb: I tried that but it failed in configure - I forgot in which step, possibly not recognising as cross-compilation, so this was the only way.

include/wx/osx/iphone/chkconf.h
RobertRoeb: Yes, wxDataViewCtrl works now.

include/wx/osx/iphone/private.h
RobertRoeb: Makes sense- will try to add this to the PR - never done that before.

include/wx/osx/iphone/private.h
RobertRoeb: Need to check if that is actually possible

RobertRoeb:
> It would be great to split this PR into independent 
> changes in order to discuss them separately 

I am not sure I know how to do that and some of the changes depend on each other. Can I create a PR out of e.g. 2 commits?

vadz:
> I am not sure I know how to do that and some of the changes depend on each other. Can I create a PR out of e.g. 2 commits?

Yes, typically you would create a branch (you should remain your current `master` to something else, like `my-wip`, because `master` should typically remain synchronized with the main repository master) and then use `git cherry-pick` to apply individual commits in this branch — and then push this branch to GitHub to create a PR for it.

vadz:
> The current wxChoice uses the picker control

Sorry, I have no idea what is the picker control, so I can't comment on this, but my specific concern with this implementation of `wxChoice` is that this control is supposed to be good for reasonably many items, e.g. it could be used to select a US state in the address entry form, and you definitely couldn't use segmented buttons for this!

>     * add an iPhone specific wxSegmentedButtons
> 
>     * re-implement wxNotebook using it
> 
>     * cast wxBookCtrl to wxNotebook instead of wxChoiceBook on iPhone
> 
>     * implement wxNavigationCtrl using native control on iPhone and cast to wxNotebook with bar at the bottom on other platforms

I am not sure, but from the application developer point of view I think just using `wxNotebook` normally would be the best.

RobertRoeb:
I am still learning git the hard way and I learnt that I need to move all my changes from master to a new branch. So I will do:
git branch iphone_updates
git checkout master
git reset --hard HEAD~24
git checkout iphone_updates
This will probably ruin this PR, but I will create new ones by cherry picking. I hope this will work....

vadz:
The above commands will preserve exactly the same changes as are here now in `iphone_updates` branch, so if this is what you want to do, they are indeed the right commands to use. Ideally, you'd do all of this except for the last command and do something like

```console
$ git checkout -b iphone-change-1
$ git cherry-pick <some commits to go into this branch>
```

to create branches with more focused changes.

As for "ruining this PR": pushing your master to GitHub after these commands will indeed do it (I'm not sure if PR will get closed automatically, but in any way, it won't contain anything useful), but this is fine.

And just a small hint at the end: instead of hard-coding 24 (the number of commits) you may use `@{u}` where "u" stands for "upstream" and is a shorthand for `HEAD@{u}` or `origin/master`.

PR #25690, opened 2025-08-09
Add tests preset to CMake

Blake-Madden:
Sets `wxBUILD_TESTS` to `ALL`.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -19,6 +19,13 @@
             }
         },
         {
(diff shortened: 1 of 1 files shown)


vadz:
I have no experience whatsoever with CMake presets, so I don't know how much sense does it make, but if @MaartenBent doesn't have any objections to this, why not.

MaartenBent (approved):
No objections from me.

PR #25694, opened 2025-08-12
allow pformatetc->lindex = 0 for CFSTR_FILECONTENTS

sz5000:
This pull requests allows pformatetc->lindex = 0, which is required for CFSTR_FILECONTENTS data objects. This change is required for dragging and dropping virtual files, e.g. dropping to Windows explorer.

Widows Explorer queries for CFSTR_FILEDESCRIPTORW and CFSTR_FILECONTENTS data. This is currently not supported by wxWidgets itself, but can be easily implemented in an application.

Virtual file drag & drop is required for files that are not available when the drag & drop operation is started.

If Windows Explorer request a CFSTR_FILECONTENTS data object, then the lindex is no longer -1. The lindex here is an index of the requested file contents. Since GetDataHere does not support indexed contents, I have limited lindex to the first item which means lindex = 0.


vadz:
Thanks, the explanation makes sense and even though I think we should ideally add real support for such data types to wx itself, for now I'll just merge this PR with only some minor cosmetic changes.

Issue #25681, opened 2025-08-04
Combobox dropdown button doesn't honour dark mode on MacOS


jey5nd6:
### Description
Dropdown arrow button is drawn in light mode for both light mode and dark mode.

#### Bug description:

The issue is surfaced in Kicad through `wxBitmapComboBox` and `wxComboCtrl` with `UseAltPopupWindow()` set.

It also shows up in the samples for `wxBitmapComboBox` and `wxOwnerDrawnComboBox`:

<img width="1768" height="796" alt="Image" src="https://github.com/user-attachments/assets/2e380e17-de98-4372-8fff-24f55aa63f54" />

### Platform and version information

wxWidgets 3.2.6 and 3.3.1
wxWidgets port wxOSX
OS macOS Sonoma Version 14.6.1 (Build 23G93), 64 bit, Little endian, wxMac
OpenGL: Apple, Apple M1 Max, 2.1 Metal - 88.1

vadz:
The trouble is that we use the native `HIThemeDrawButton()` function for drawing this button and I think (@csomor please correct me if I'm wrong) that it just doesn't support dark mode.

I wonder if we could just invert the bitmap in dark mode. If not, the only solution I can see is to fall back to the generic implementation of `wxRendererNative::DrawComboBoxDropButton()` — this won't look native, but will at least use the correct colours.

Any other ideas would be welcome, of course.

csomor:
@vadz yes, this is correct, and while we can render an entire combobox with a NSCell subclass, there is no special subclass for the button of a combobox itself, and the buttoncell used is private, I'm testing some ideas now

PR #25696, opened 2025-08-14
OSX implement dark mode aware ComboBoxDropButton rendering

csomor:
see #25681


vadz (approved):
Thanks a lot Stefan, this looks good to me, of course, but I didn't test it yet. @jey5nd6 Could you please confirm that it works for you?

jey5nd6:
Yep, works for both our cases (`wxBitmapComboBox` and `wxComboCtrl` with `UseAltPopupWindow()` set).

Issue #25698, opened 2025-08-15
2 GDI objects are created for each wxStaticBitmap.


dsa-t:
### Description
In KiCad we create a panel with `wxStaticBitmap` elements. This causes GDI object exhaustion when there's too many items:

<img width="157" height="302" alt="Image" src="https://github.com/user-attachments/assets/e421f8b0-2390-47a0-a72c-587c26d92903" />

Looks like it's possible to reduce the GDI object to 1 per `wxStaticBitmap` by freeing `wxStaticBitmap`'s handle if a copy was created internally:

```
void wxStaticBitmap::MSWReplaceImageHandle(WXHANDLE handle)
{
    HGDIOBJ oldHandle = (HGDIOBJ)::SendMessage(GetHwnd(), STM_SETIMAGE,
        m_icon.IsOk() ? IMAGE_ICON : IMAGE_BITMAP, (LPARAM)handle);

    // detect if this is still the handle we passed before or
    // if the static-control made a copy of the bitmap!
    if (oldHandle != 0 && oldHandle != (HGDIOBJ)m_currentHandle)
    {
        // the static control made a copy and we are responsible for deleting it
        ::DeleteObject((HGDIOBJ)oldHandle);
    }

    HGDIOBJ newHandle = (HGDIOBJ)::SendMessage(GetHwnd(), STM_GETIMAGE,
        m_icon.IsOk() ? IMAGE_ICON : IMAGE_BITMAP, 0);

    if (newHandle != (HGDIOBJ)m_currentHandle) {
        // a copy was created, so we don't need to keep our own handle
        Free();
    }
}
```

(Tested with bitmaps but not with icons)

`layout` sample patch for testing:

[layout-gdi.patch](https://github.com/user-attachments/files/21793075/layout-gdi.patch)

---

KiCad reference (invisible for now due to confidentiality): https://gitlab.com/kicad/code/kicad/-/issues/21386

dsa-t:
Also using wxIcon minimizes the GDI object count:

```cpp
    wxImage img;
    img.LoadFile("C:/Build/wxWidgets/samples/toolbar/bitmaps/help.png");
    wxBitmapBundle bmplabel = wxBitmapBundle::FromImage(img);

    wxIcon icon;
    icon.CopyFromBitmap(bmplabel.GetBitmapFor(this));

    for (int i = 0; i < 200; i++) {
        auto statbmp = new wxStaticBitmap(p, wxID_ANY, icon);

[...]

PR #25699, opened 2025-08-15
Free unused bitmaps in wxMSW `wxStaticBitmap`

vadz:
See #25698.

### DIFF
--- a/src/msw/statbmp.cpp
+++ b/src/msw/statbmp.cpp
@@ -285,10 +285,10 @@ void wxStaticBitmap::DoUpdateImage(const wxSize& sizeOld, bool wasIcon)
 
 
     // Update the handle used by the native control.
+    const WPARAM imageType = m_icon.IsOk() ? IMAGE_ICON : IMAGE_BITMAP;
+
     const HGDIOBJ oldHandle = (HGDIOBJ)
-        ::SendMessage(GetHwnd(), STM_SETIMAGE,
-                      m_icon.IsOk() ? IMAGE_ICON : IMAGE_BITMAP,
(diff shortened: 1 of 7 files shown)


src/msw/statbmp.cpp
```diff
@@ -296,7 +283,30 @@ void wxStaticBitmap::DoUpdateImage(const wxSize& sizeOld, bool wasIcon)
...
+    {
+        // the static control made a copy and we are responsible for deleting it
+        ::DeleteObject(oldHandle);
+    }
+
+    // Also check if we need to keep our current handle, it may be unnecessary
+    // if the native control doesn't actually use it.
+    const HGDIOBJ newHandle = (HGDIOBJ)
+        ::SendMessage(GetHwnd(), STM_GETIMAGE, imageType, 0);
+    if ( newHandle != (HGDIOBJ)m_currentHandle )
```
dsa-t, line 305: White spaces around `(HGDIOBJ)` look inconsistent, but otherwise looks good!

dsa-t:
It's probably a separate bug, but if we sequentially set two bitmaps without alpha, does the handle of the first bitmap get deleted by `::DeleteObject(oldHandle);`?

It appears to be the case in https://gitlab.com/kicad/code/kicad/-/issues/22767

dsa-t:
(the above is likely a result of https://github.com/wxWidgets/wxWidgets/commit/3e32a9abe1e05f0cddb17c791a59977bed8e8cff)

dsa-t:
And looks like it's backported to 3.2.9 too https://github.com/wxWidgets/wxWidgets/pull/25674

PR #25710, opened 2025-08-20
Make restoring window size with XFCE/MATE work again

vadz:
Workaround for Mutter from 9537141500 (Fix handling window total size under GNOME with X11, 2025-04-28) broke the well behaving WMs used by other desktop environments, such as XFCE, as it resulted in the window size being mistakenly adjusted after the decorations size was changed due to restoring a previous maximized window.

Prevent this from happening by not adjusting the size after the first correct decorations size change.

See #25348, #25349.

Closes #25708.

PR #25711, opened 2025-08-20
Fix TLW size after restoring from fullscreen with X11

paulcor:


vadz:
Is this better than #25710? I prefer the other PR because it's simpler and easier to backport to 3.2.

paulcor:
I think it's better, as it's more general and therefore possibly safer. Yeah, it doesn't apply cleanly to 3.2. I can do the backport, or we can just apply your simpler fix there.

vadz:
I prefer to avoid using different versions in 3.3 and 3.2, this would make any further changes even more difficult, so let's apply the second commit from this PR to 3.2 then and drop the other one.

Although I still wonder if we shouldn't reset `m_deferShowAllowed` to false after using it, it doesn't make much sense to use it more than once, does it?

PR #25716, opened 2025-08-21
Update wxAuiToolBar after wxSysColourChangedEvent

MaartenBent:
The background bitmap of the toolbar is cached and should be recreated after the colour change. Otherwise it will not be updated until a wxSizeEvent is generated.


vadz:
Thanks! Will merge soon and this is probably needed in 3.2 too, isn't it?

MaartenBent:
Yes, seems like the code in 3.2 is the same, so that can get the fix too.

vadz:
Oops, no, it's not the same, `UpdateBackgroundBitmap()` doesn't exist in 3.2, so I won't backport it there even though it possibly could be — but I think it's much more important in master, due to dark mode support.

PR #25718, opened 2025-08-22
Make wxTopLevelWindow label same as title in all ports

vadz:
This was previously done for wxGTK in a03b38ef68 (make Set/GetLabel() set and return title in wxTLW, fixes #12371: Dialog::GetLabel() Inconsistent behaviour across operating systems, 2010-08-21) and wxOSX in 033f86db5f (adding SetLabel -> SetTitle redirects, solves missing title updates using wxDocument/wxView, 2012-08-11) but wasn't done for the other ports, including wxMSW and updating the page title in wxAuiMDIChildFrame in docview framework didn't work because of it.

While using SetLabel() for setting the title is suspicious and it might be a good idea to change docview code too, for now make it work by ensuring that this works consistently in all ports.

See #12371.

Closes #25712.

### DIFF
--- a/include/wx/gtk/toplevel.h
+++ b/include/wx/gtk/toplevel.h
@@ -71,9 +71,6 @@ public:
     virtual void SetTitle( const wxString &title ) override;
     virtual wxString GetTitle() const override { return m_title; }
 
-    virtual void SetLabel(const wxString& label) override { SetTitle( label ); }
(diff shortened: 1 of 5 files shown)


vadz:
Applied as d6af1f50caafcbc00a38f13213fefc3a1e713f8a

Issue #23778, opened 2023-08-16
wxWindow::WarpPointer does nothing on GTK w/Wayland


scottb88:
### Description
Since Fedora switched to Wayland, WarpPointer does nothing.

My app uses WarpPointer at various points, always in response to mouse events, in a way that works the way the user is expecting, doesn't surprise them, saves them time, and is always within the main app window. Works great on Mac and windows.

### Platform and version information
I'm using 
- wxWidgets version 3.1.5
- I checked the change log and looked at the source on Master and it appears to be the same as 3.1.5.

+ GTK version: 4.10.4-1.fc38 (all versions in the last 2 years)
+ GDK == Wayland    works perfect on X11

scottb88:
Wayland is security conscious, and allowing an app free reign of the mouse is said to be a security risk. An old (2014) discussion stated that relative_pointer must be used:
https://wayland.app/protocols/relative-pointer-unstable-v1

I'm not sure if gdk_seat_grab () would help. That would restrict the warp to the window, and be safer. I have zero experience with these APIs.

GTK/Wayland finally came up with a secure and efficient way to capture screen movement to video. It took years. That's about all I know.

gdk_seat_grab ()
--


[gdk_seat_grab](https://www.manpagez.com/html/gdk3/gdk3-3.22.0/GdkSeat.php#gdk-seat-grab) ()

gdk_seat_grab ()
--


[gdk_seat_grab](https://www.manpagez.com/html/gdk3/gdk3-3.22.0/GdkSeat.php#gdk-seat-grab) ()

imciner2:
I went down the wayland mouse warping rabbit hole a year ago, and it is basically discouraged by upstream (not necessarily for security reasons, but because they think it is bad UI/UX for the application to move the mouse on its own). My results are summarized here: https://gitlab.com/kicad/code/kicad/-/issues/9785#note_745301721.

That said, apparently there is an extension that some compositors are implementing called `zwp_pointer_lock_v1.set_cursor_position_hint`, which is supposed to allow programmatic movement of the cursor (extension defined here: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/unstable/pointer-constraints/pointer-constraints-unstable-v1.xml#L218). However, even that extension seems to be implemented in different ways in different window managers/compositors, with this comment saying Weston and Mutter apply it at different times https://gitlab.freedesktop.org/xorg/xserver/-/issues/734#note_749474.

[...]

PR #25721, opened 2025-08-22
Use Wayland protocol to warp pointer

vadz:
See #23778.

This has been only tested with Jay Wayland compositor as it's the only one I could find that implements this protocol already. Unfortunately this doesn't work quite right there:

1. The position of the cursor seems to be off by the decorations size.
2. When the mouse moves after warp, it jumps back to the old position.

I could work around (1) but (2) makes this rather unusable, so I hope that it's a bug in Jay (at least in the scenario in which I'm using it, as a nested compositor), and I wonder if (1) couldn't be a bug too, so it would be great if somebody could please test this with a more common compositor supporting this protocol.


mahkoh:
>When the mouse moves after warp, it jumps back to the old position. 
>
>at least in the scenario in which I'm using it, as a nested compositor

The X11 backend that you are using here is essentially unmaintained. In this particular case, the issue is that it receives absolute cursor positioning events from the X server. You can warp the cursor, but the next time you move the mouse the position will be overwritten by the position the X server sends us. I don't think that is necessarily a bug, if you used another absolute positioning input method, such as a touch screen or a tablet, you would see the same behavior and it would not be surprising.

include/wx/gtk/private/wayland.h
```diff
@@ -18,7 +18,13 @@
 // GTK version to compile the code using it.
 #if defined(GDK_WINDOWING_WAYLAND) && GTK_CHECK_VERSION(3,20,0)
     #ifdef wxHAVE_WAYLAND_CLIENT
-        #define wxHAVE_WAYLAND_PROTOCOLS
+        #include <wayland-client-protocol.h>
+
+        // Old headers don't define this flag used in
+        // pointer-warp-v1-client-protocol.h, so we can't compile it with them.
+        #ifdef WL_MARSHAL_FLAG_DESTROY
+            #define wxHAVE_WAYLAND_PROTOCOLS
```
mahkoh, line 26: You should generate these files at compile time with the wayland-scanner version installed on the system. This will ensure that the generated code is in sync with the installed libwayland. 

What you should do is check the required protocol xml files into the wxwidgets repository.

mahkoh:
>The position of the cursor seems to be off by the decorations size.

I have not looked at the code but I think what is happening here is that the gtk coordinates that you are working with are not surface-local coordinates but rather xdg_surface geometry coordinates or something similar.

Since the protocol requires surface-local coordinates, you have to find a way to transform your coordinates into surface-local coordinates, or you can't use the protocol.

It could be a bug in the compositor but when I tested the code with SDL and blender it worked fine.

mahkoh:
You can verify this by e.g. creating an example program that warps the cursor to 100x100 and then look at the WAYLAND_DEBUG output.

include/wx/gtk/private/wayland.h
vadz: Sorry, I know we're supposed to do this, but I really don't want to even think about adding support for running wayland-scanner to bakefile-generated makefiles. At least for now, we'll live with these files in the repository.

vadz:
Thanks for looking at this!

Are you able to test this with some other compositor with support for this protocol?

> Since the protocol requires surface-local coordinates, you have to find a way to transform your coordinates into surface-local coordinates, or you can't use the protocol.

Is the difference between these coordinates just the decorations size or is there something more to it? I don't see anything coordinates-related in `xdg_surface`...

> You can verify this by e.g. creating an example program that warps the cursor to 100x100 and then look at the WAYLAND_DEBUG output.

Well, this is what the modified sample does, but I'm not sure what am I supposed to conclude from this.

mahkoh:
You can move the cursor over the point that you think is the 100x100 point and the events will tell you the surface-local coordinates.

vadz:
> You can move the cursor over the point that you think is the 100x100 point and the events will tell you the surface-local coordinates.

If you mean the coordinates from `pointer_handle_motion` (part of pointer listener), then I get back exactly the coordinates I pass to it, but I'm not sure if they are "surface local".

In the meanwhile, I've also tested with mutter by running it in another VT as I still have no idea how to run it as nested compositor (again). In my uninformed opinion it looks like its implementation of the protocol is totally broken, see [my comment there](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4525#note_2533401).

vadz:
I've fixed the offset with Jay by using `gdk_window_get_origin()`, see e8e1433d1f (Account for the window origin when warping pointer using Wayland, 2025-08-29).

I'd still be very grateful if somebody could please test this with KDE, I'd rather avoid having to install (or build?) it here myself just for this...

vadz:
I'm probably going to merge this, even if it's not very useful for now (but if [my mutter MR](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4614) gets merged it will at least work in GNOME when a mouse button is pressed, which is better than never working), if there are no other comments.

vadz:
My mutter MR did get merged, so merging this too. It's not perfect, but better than nothing.

include/wx/gtk/private/wayland.h
vadz: @mahkoh Just FYI, I think I've finally found a relative painless way to run `wayland-scanner` during the build without dealing with our build system, see #25769.

Issue #25707, opened 2025-08-19
wxSVGFileDC: text spacing incorrect and different in X11 (native or XWayland) and Wayland


lallulli:
### Description

SVG files produced by wxSVGFileDC under wxGTK in Wayland have wrong horizontal text spacing and size, both in Wayland and in X11/XWayland.

In X11 (both in a native X11 session and under Wayland/XWayland) letter spacing is too large:
![Image](https://github.com/user-attachments/assets/286d3ade-3fc2-4c9d-b28a-ad92b152acf2)

In a native Wayland session letter spacing is too narrow:
![Image](https://github.com/user-attachments/assets/6de4b031-abc8-4aac-abee-fa73952cbc8d)

#### Patch or snippet allowing to reproduce the problem:

```python
# test.py
import wx
app = wx.App()
dc = wx.SVGFileDC('test.svg', 200, 200)
dc.DrawText('Hello World', 50, 50)
```

Run this code fragment in an Wayland session:
```bash
python test.py
mv test.svg test_wayland.svg
GDK_BACKEND=x11 python test.py
mv test.svg test_x11.svg
```

### Platform and version information

- wxWidgets version: 3.2.8.1 (also tested on 3.2.7 and earlier versions)
- wxWidgets port: wxGTK
- OS: Manjaro Linux
  + GTK version: 3.24.49
  + Which GDK: X11, Wayland and XWayland
  + Desktop environment: KDE Plasma 6.3.6, Qt 6.9.1, KDE Frameworks 6.17.0

vadz:
Isn't this a font and/or SVG viewer problem? SVG doesn't specify any spacing AFAICS, here is what this generates here:

```svg
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.0//EN" "http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg10.dtd">

<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="11.29cm" height="8.47cm" viewBox="0 0 320 240">
<title></title>
<desc>Picture generated by wxSVG v0101</desc>

<g style="fill:black; stroke:black; stroke-width:1">
</g>
[...]

PR #22390, opened 2022-05-03
Improve wxSVGFileDC on macOS

MaartenBent:
On macOS everything is based on 72DPI, while the SVG is based on 96DPI. As a result, fonts in point-size are too big and squeezed together (squeezing due too `textLength`).

Fix this by taking the standard screen DPI into account in `FromDIP` and `ToDIP`, so all coordinates and sizes will be in SVG DPI.

`wxSVGFileDC` should be created with a DIP size, so use `FromDIP` to get the correct `viewBox` size and clear-rectangle size.

png:
<a href="https://user-images.githubusercontent.com/8088070/166565270-bdab3cfe-f54b-4b0f-befb-5bfc81034ee5.png">png</a>
old svg:
<a href="https://user-images.githubusercontent.com/8088070/166565296-f90ce2af-db54-4a5c-adfb-f511f2f810c3.svg">old svg</a>
new svg:
<a href="https://user-images.githubusercontent.com/8088070/166565291-098b7cb3-2c10-452d-a318-9652946840c6.svg">new svg</a>

And make all pages in the drawing sample DPI aware.

PR #25723, opened 2025-08-23
wxSVGFileDC improvements

MaartenBent:
Multiple fixes and improvements to `wxSVGFileDC`. Some I mentioned back in https://github.com/wxWidgets/wxWidgets/pull/22390#issuecomment-1116722724

1) Avoid using the style attribute in wxSVGFileDC because sometimes browsers will not parse this when there is not Content Security Policy, see example in the linked comment.

2) Try to fix text metric issues, see #25707, by using `wxMemoryDC` instead of `wxScreenDC`.

3) Fix incorrect `FromDIP` usage, basically reverting parts of #22390. The goal of `wxSVGFileDC` should be to create the same SVG file on each platform. This was broken on macOS, all coordinates would be scaled with a factor 1.33 (96/72). The text metrics still needs to be scaled, because they are based on 72dpi.

4) Update the drawing and svg examples, save the SVG with full drawing size.

With these changes, the following SVG are created on Windows, Ubuntu and macOS, at 200%/retina scaling. (I also tested Windows and Ubuntu at 100% and 150%, and Ubuntu with and without `GDK_BACKEND=x11`). The SVG are almost the same, but the text metrics are slightly different per platform.

<a href="https://github.com/user-attachments/assets/0f2bf306-70d8-40ba-94b1-482b64fce07a">windows</a> <a href="https://github.com/user-attachments/assets/837bb224-e6b0-4f60-a882-49852b5a108d">ubuntu</a> <a href="https://github.com/user-attachments/assets/ff7ba656-6192-495b-b12b-541085840d65">macos</a>

wxDC reference:
<a href="https://github.com/user-attachments/assets/f8fc0e27-3adc-487b-aa03-dacb667842fe">windows</a> <a href="https://github.com/user-attachments/assets/87664900-aa7f-4195-b532-90b827a7ee02">ubuntu</a> <a href="https://github.com/user-attachments/assets/29eb4e68-e38f-4732-bcf2-67ede3eb1ae6">macos</a>

Note that the macOS screenshot exposes a problem with the font size. It is set to `9` and `18` points in the code, but it is too small in the `wxDC`. Previously it used `*wxSWISS_FONT` which looked better, but that has a point size of 13. Which is way too big for the SVG. 
I am not sure how to deal with this. It seems like some kind of 72/96 conversion problem, because with 12 point (9*96/72) the wxDC looks almost the same as Windows and Ubuntu.

And with 13 point (`*wxSWISS_FONT`) it looks like this:
<a href="https://github.com/user-attachments/assets/200c24b8-db1c-40c9-9bc6-cdb75859c6b2">macos wxSWISS_FONT</a>


include/wx/dcsvg.h
```diff
@@ -93,7 +95,7 @@ class WXDLLIMPEXP_CORE wxSVGFileDCImpl : public wxDCImpl
 
     virtual ~wxSVGFileDCImpl();
 
-    bool IsOk() const override { return m_OK; }
+    bool IsOk() const override { return m_OK || m_filename.empty(); }
```
vadz: Sorry, but this is really confusing. After looking at the source, I see that this is indeed correct, but when seeing it initially I was sure that there was `!` missing here or something else was wrong.

If we keep it like this, I'd really like to add a comment explaining what's going on here, but I think it could be preferable to rename `m_OK` to `m_writeOk` (or just `m_saved`?) and change the logic for setting it, so that it remains true if there is no output file. Or, maybe even more clearly, replace `m_OK` with `!m_writeError` and only set `m_writeError = true` when writing failed (and so never if there is no output file).

samples/drawing/drawing.cpp
```diff
@@ -2784,13 +2784,16 @@ void MyFrame::OnSave(wxCommandEvent& WXUNUSED(event))
...
 #endif
-            wxSVGFileDC svgdc(dlg.GetPath(),
-                              canvasSize.GetWidth(),
-                              canvasSize.GetHeight(),
-                              72,
-                              "Drawing sample");
-            svgdc.SetBitmapHandler(new wxSVGBitmapEmbedHandler());
-            m_canvas->Draw(svgdc);
+            wxSize svgSize;
+            wxSVGFileDC tempSvgDC("", svgSize.x, svgSize.y, 72, "Drawing sample");
```
vadz: Probably should provide ctor overload taking `wxSize`...

src/common/dcsvg.cpp
```diff
@@ -86,7 +84,7 @@ wxString Col2SVG(wxColour c, float* opacity = nullptr)
 wxString GetPenStroke(const wxColour& c, int style = wxPENSTYLE_SOLID)
 {
     float opacity;
-    wxString s = wxS("stroke:") + Col2SVG(c, &opacity) + wxS(";");
+    wxString s = wxS("stroke=\"") + Col2SVG(c, &opacity) + wxS("\"");
```
vadz: I think it would be more readable written like this:

```suggestion
    wxString s = wxString::Format(wxS(R"(stroke="%s")"), Col2SVG(c, &opacity));
```

src/common/dcsvg.cpp
```diff
@@ -348,43 +346,40 @@ wxString CreateBrushFill(const wxBrush& brush, wxSVGShapeRenderingMode mode)
...
+        s += wxString::Format(wxS("    <path stroke=\"%s\" stroke-opacity=\"%s\" %s %s %s/>\n"),
             brushColourStr, NumStr(opacity), brushStrokeStr, pattern, GetRenderMode(mode));
         s += wxS("  </pattern>\n");
     }
 
     return s;
 }
 
-void SetScaledScreenDCFont(wxScreenDC& sDC, const wxFont& font)
+std::unique_ptr<wxMemoryDC> GetTextMetricDC(const wxFont& font)
```
vadz: Using `unique_ptr` just because we can't return `wxDC` looks awkward. Best would probably be to make `wxDC` moveable to allow returning it, but for now perhaps it would be better to make this `void SetSVGFont(wxMemoryDC& dc)` even if this means that the caller would have to create `wxMemoryDC` (IMHO not a big deal)?

vadz (commented):
Thanks for all the fixes!

I'd like to make the code a bit more clear or at least add some comments before merging it, but otherwise this looks good to me.

Would you have a short sentence describing the changes in this PR for the change log? TIA!

include/wx/dcsvg.h
MaartenBent: I like the `m_writeError` idea. I'll implement this.

src/common/dcsvg.cpp
MaartenBent: The caller used to create the `wxScreenDC`, but I tried to get rid of the code duplication. I can restore it and just change `SetScaledScreenDCFont(wxScreenDC& sDC, ...)` to `SetScaledDCFont(wxDC& dc, ...)`.

vadz: It looks like making `wxDC` moveable might be easier than I thought, see #25726. If this looks fine to you, let's merge it and then this function could just return `wxMemoryDC`?

samples/drawing/drawing.cpp
```diff
@@ -1806,7 +1806,7 @@ void MyCanvas::DrawSystemColours(wxDC& dc)
     }
 
     int lineHeight = textSize.GetHeight();
-    wxCoord x(FromDIP(10));
+    wxCoord x(dc.FromDIP(10));
```
vadz, line 1809: Maybe we should move this function our of `MyCanvas`, so that it wouldn't be able to use window scale factor accidentally?

samples/drawing/drawing.cpp
```diff
@@ -2790,13 +2790,16 @@ void MyFrame::OnSave(wxCommandEvent& WXUNUSED(event))
...
-                              72,
-                              "Drawing sample");
-            svgdc.SetBitmapHandler(new wxSVGBitmapEmbedHandler());
-            m_canvas->Draw(svgdc);
+            wxSize svgSize;
+            wxSVGFileDC tempSvgDC(svgSize);
+            m_canvas->Draw(tempSvgDC);
+
+            svgSize = wxSize(tempSvgDC.MaxX(), tempSvgDC.MaxY());
+            svgSize.IncBy(15); // account for wxPen width exceeding bounds
```
vadz, line 2798: This "15" seems pretty arbitrary, but I guess we can't do any better?

samples/svg/svgtest.cpp
```diff
@@ -230,204 +226,213 @@ void MyFrame::FileSavePicture(wxCommandEvent& WXUNUSED(event))
...
+    dc.SetFont(wxFontInfo(9).FaceName("Arial").Family(wxFONTFAMILY_SWISS));
     dc.SetPen(*wxGREEN_PEN);
 
     switch (m_index)
     {
         case Page_Lines:
             // draw lines to make a cross
-            dc.DrawLine(0, 0, 200, 200);
-            dc.DrawLine(200, 0, 0, 200);
+            dc.DrawLine(dc.FromDIP(0), dc.FromDIP(0), dc.FromDIP(200), dc.FromDIP(200));
```
vadz, line 272: Similarly to the remark above, maybe we should move this code into a function to make sure it doesn't use `MyPage` scaling factor?

vadz (approved):
Thanks for all these improvements!

Should this be merged now or do you plan to do more here?

If it should be merged, I'd still welcome a good message for the merge commit as I have trouble summarizing all the changes here.

samples/drawing/drawing.cpp
MaartenBent: This applies to all the `MyCanvas::Draw*` functions. I think this is a too big and unrelated change to make in this PR.

samples/svg/svgtest.cpp
MaartenBent: Done. Do you agree with this approach?

samples/drawing/drawing.cpp
MaartenBent: It is a bit arbitrary, but it seems to work for the drawing and svg samples, while `10` still had some lines cut-off.

samples/svg/svgtest.cpp
vadz: Yes, nice, thank you!

PR #25726, opened 2025-08-25
Make `wxDC`-derived objects moveable

vadz:
This can be nice as it allows to return them from functions.


vadz:
First commit here is unrelated and was included accidentally, sorry.

vadz:
Any objections to merging this? If not, I'd like to do it, I've just found another place where it would be useful to have a function returning a `wxDC`-derived object.

MaartenBent:
Works good in the `wxSVGFileDC` case.

PR #25729, opened 2025-08-26
Test only one wxDateTimeHolidayAuthority at a time

lanurmi:
`wxDateTimeChristianHolidays` was tested with a `wxDateTimeUSCatholicFeasts` already existing and taking precedence. I.e. the tests for Christmas in ChristianHolidays actually tested (again) the USCatholicFeasts implementation.

This reveals an off-by-one bug in `wxDateTimeChristianHolidays::DoIsHoliday()`, which doesn't use the `enum` symbol for month, and declares Christmas to be on the 25th day of the 13th month.

Further, this raises the question whether it is a bad idea to have the authority implementations in a header. Fixing bugs like this would require recompiling the application, instead of simply updating the DLL/dylib.

### DIFF
--- a/tests/datetime/datetimetest.cpp
+++ b/tests/datetime/datetimetest.cpp
@@ -2550,6 +2550,7 @@ TEST_CASE("US Catholic Holidays", "[datetime][holiday]")
     }
     SECTION("Fixed date feasts")
     {
+        wxDateTimeHolidayAuthority::ClearAllAuthorities();
         wxDateTimeHolidayAuthority::AddAuthority(new wxDateTimeUSCatholicFeasts);
         CHECK(wxDateTimeHolidayAuthority::IsHoliday(wxDateTime( 1, wxDateTime::Month::Jan, 2024)));
         CHECK(wxDateTimeHolidayAuthority::IsHoliday(wxDateTime(15, wxDateTime::Month::Aug, 2023)));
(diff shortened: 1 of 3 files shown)


tests/datetime/datetimetest.cpp
```diff
@@ -2550,6 +2550,7 @@ TEST_CASE("US Catholic Holidays", "[datetime][holiday]")
     }
     SECTION("Fixed date feasts")
     {
+        wxDateTimeHolidayAuthority::ClearAllAuthorities();
```
vadz: Could we do it once at the beginning of the test? In fact, ideal would be to just do

```suggestion
    wxON_BLOCK_EXIT0(wxDateTimeHolidayAuthority::ClearAllAuthorities);
```

at the start of the function, outside of any `SECTION` — wouldn't this work?

vadz (commented):
Thanks for fixing this embarrassing bug, I can merge it already but it looks like the tests could could be simplified a bit, perhaps?

tests/datetime/datetimetest.cpp
lanurmi: It would work with `wxON_BLOCK_EXIT0`, but preferably the authorities should still be cleared at the beginning also, because otherwise there's the `wxDateTimeWorkDays` object that exists during test(s).

vadz: Every test should clean up after itself so I don't think of these objects should exist in the beginning.

lanurmi: > Every test should clean up after itself so I don't think of these objects should exist in the beginning.

Try setting a breakpoint at `wxDateTimeWorkDays::DoIsHoliday` in current master, and you'll notice it gets called from both US Catholic and Christian test cases. And that's because `wxDateTimeHolidaysModule::OnInit()` always creates such an authority during startup.

vadz: OK, I forgot about this, but this means that we actually need some `AddDefaultAuthorities()` too and do `ClearAllAuthorities()` on the test entry and `ClearAllAuthorities()` followed by the call to `AddDefaultAuthorities()` on exit, right?

lanurmi: Adding more (member?) functions (to the public API?), ones that would only be used by these two tests, makes things needlessly complex IMHO. After all, there is only one default authority, and I don't imagine there ever being more in the foreseeable future.

I've added code to restore the default authority, and comments explaining what is going on.

vadz:
Thanks, will merge and push soon.

vadz:
Done now.

PR #25730, opened 2025-08-26
Suppress clang static analysis warnings

Blake-Madden:
Don't appear to be real issues, but doesn't hurt to init values and add a div by zero check.

### DIFF
--- a/src/common/statbar.cpp
+++ b/src/common/statbar.cpp
@@ -226,7 +226,7 @@ wxArrayInt wxStatusBarBase::CalculateAbsWidths(wxCoord widthTotal) const
                 widths.Add(pane.GetWidth());
             else
             {
-                int nVarWidth = widthExtra > 0 ? (widthExtra * (-pane.GetWidth())) / nVarCount : 0;
+                int nVarWidth = (widthExtra > 0 && nVarCount != 0) ? (widthExtra * (-pane.GetWidth())) / nVarCount : 0;
                 nVarCount += pane.GetWidth();
                 widthExtra -= nVarWidth;
                 widths.Add(nVarWidth);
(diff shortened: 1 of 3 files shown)


src/common/statbar.cpp
```diff
@@ -226,7 +226,7 @@ wxArrayInt wxStatusBarBase::CalculateAbsWidths(wxCoord widthTotal) const
                 widths.Add(pane.GetWidth());
             else
             {
-                int nVarWidth = widthExtra > 0 ? (widthExtra * (-pane.GetWidth())) / nVarCount : 0;
+                int nVarWidth = (widthExtra > 0 && nVarCount != 0) ? (widthExtra * (-pane.GetWidth())) / nVarCount : 0;
```
vadz, line 229: Sorry, this is not unlikely, this is impossible because the loop above makes sure `nVarCount` is non-zero in this case. I also dislike warnings, but this makes the code harder to understand, so I don't like this fix either.

Can we use something like `__builtin_assume(nVarCount > 0)` (wrapped in a macro, as usual) instead?

vadz (commented):
Thanks, initializing variables is fine, but adding always true checks to the code doesn't seem like a good idea, can we fix this warning in some other way?

src/common/statbar.cpp
Blake-Madden: I can only get this (false positive) warning from a deep clang static analysis. I'm OK with just not including anything new regarding the (false) div by zero. Something like `__builtin_assume(nVarCount > 0)` seems like an unnecessary cluttering of the code.

vadz: Using "assume" may be good because it clearly establishes the invariant, so I could live with it. But I can also live with doing nothing, of course.

Blake-Madden: I'm OK with doing nothing. Macro decorations will make it harder to read than my original change.

vadz:
I've merged this without the last commit, thanks.

PR #25740, opened 2025-08-30
Add wxString::AssignFromUTF8() function

vadz:
This can be used to optimize repeated conversions from UTF-8 by reusing the string buffer.


vadz:
Applied in the commit above.

PR #2555, opened 2021-10-20
OSX remove NSImage from public API of wxBitmapBundleImpl

csomor:

PR #25744, opened 2025-08-30
macOS: Fixes to incorrect NSImage creation from wxBitmapBundle and missing menu bitmaps

vslavik:
This PR contains two commits that appear unrelated at first, but the underlying problem in both of them is the same thing: misunderstanding of what `NSImage` is (a direct equivalent of `wxBitmapBundle`) and creating it through a lossy conversion to `wxBitmap` (which unlike on other platforms *doesn't* have a native equivalent).

### wxMenu bitmaps

The primary commit throws out all of the `SetupBitmaps()` scaffolding from wxOSX and reverts most of 337940f, 616e7c8, plus bits of c0dbe80; 338cd95 is also related). It then replaces all that code with a straightforward implementation that just directly sets the bundle/NSImage in `wxMenuItemImpl`.

This fixes multiple bugs, introduced by 337940f, in wxOSX implementation of menu icons:

1. Bitmaps set on items in the global Mac menu were ignored.
2. So were the ones set in the (also Mac-specific) app menu.
3. Loosing compatibility with `NSImage`-backed `wxBitmap` and rendering them without appearance awareness and in low resolution in the global menu.

It's the last one that finally tipped me off that something is *seriously* wrong, because starting with a native resolution-independent, appearance-independent SF Symbol and ending up with a blurry dark-on-dark bitmap is something that takes non-obvious code to accomplish!

### wxDataViewCtrl

The other commit fixes similar problem in wxDVC code, introduced by 33d8510. I didn't encounter these bugs in real use, but found them when reviewing wx code looking for other instances of this class of issues [^1].

### 3.2

Menu item bitmaps are important on macOS 26, so I think this part at least should be backported.

wx-3.2 version that I think is ABI-safe (it just adds an override to vtable) is here: https://github.com/poedit/wxWidgets/commit/011f9a2da063a6d756ec6e6a46c278bab6f6c5eb

### NSImage and wxBitmapBundle

 I think the removal of `NSImage` from `wxBitmapBundle` API (#2555, merged as 4eb4c17 and considered improvement) was a mistake — and the bugs addressed in this PR would all be harder to write if only the developer had `wxBitmapBundle::GetNSImage()` at their disposal.

As it is, the conversion has to be done with semi-hidden private API that even wx developers can overlook, and which is less ergonomic to use. For *user* code it is not just hard, but *impossible* to interop between `wxBitmapBundle` and native code w/o using the private API. 

Ideally, this PR would contain two more commits:

1. Introduction of `GetNSImage()` and `FromNSImage()`.
2. Change of wx's internal uses of the private helpers to use these.

Happy to do it, but the comments in #2555 suggest I can expect some pushback, so didn't want to do the work without an explicit ACK first.

[^1]:  In wxOSX basically any use of `wxBitmapBundle::GetBitmapFor()` or `GetIconFor()` is either wrong or highly suspect. Even in drawing code it's at least inefficient and would ideally be avoided.

### DIFF
--- a/include/wx/menu.h
+++ b/include/wx/menu.h
@@ -260,7 +260,7 @@ public:
     //
     // NB: avoid calling SetInvokingWindow() directly if possible, use
     //     wxMenuInvokingWindowSetter class below instead
-    virtual void SetInvokingWindow(wxWindow *win);
+    void SetInvokingWindow(wxWindow *win);
     wxWindow *GetInvokingWindow() const { return m_invokingWindow; }
 
     // the window associated with this menu: this is the invoking window for
--- a/include/wx/osx/core/private.h
+++ b/include/wx/osx/core/private.h
@@ -161,7 +161,7 @@ public :
     }
 
     virtual ~wxMenuItemImpl() ;
-    virtual void SetBitmap( const wxBitmap& bitmap ) = 0;
+    virtual void SetBitmap( const wxBitmapBundle& bitmap ) = 0;
     virtual void Enable( bool enable ) = 0;
     virtual void Check( bool check ) = 0;
     virtual void SetLabel( const wxString& text, wxAcceleratorEntry *entry ) = 0;
--- a/include/wx/osx/menu.h
+++ b/include/wx/osx/menu.h
@@ -35,8 +35,6 @@ public:
 
     virtual void SetTitle(const wxString& title) override;
 
-    virtual void SetInvokingWindow(wxWindow* win) override;
-
     bool ProcessCommand(wxCommandEvent& event);
 
     // get the menu handle
@@ -66,12 +64,6 @@ public:
     // at given position belongs. Return false if there is no radio group
     // containing this position.
     bool OSXGetRadioGroupRange(int pos, int *start, int *end) const;
-
-#if wxUSE_MENUBAR
-    virtual void Attach(wxMenuBarBase *menubar) override;
-#endif
-
-    void SetupBitmaps();
 
 protected:
     // hide special menu items like exit, preferences etc
@@ -167,10 +159,6 @@ public:
     static wxMenuBar* MacGetInstalledMenuBar() { return s_macInstalledMenuBar ; }
     static void MacSetCommonMenuBar(wxMenuBar* menubar) { s_macCommonMenuBar=menubar; }
     static wxMenuBar* MacGetCommonMenuBar() { return s_macCommonMenuBar; }
-
-    virtual void Attach(wxFrame *frame) override;
-    void SetupBitmaps();
-
 
     static WXHMENU MacGetWindowMenuHMenu() { return s_macWindowMenuHandle ; }
 
--- a/include/wx/osx/menuitem.h
+++ b/include/wx/osx/menuitem.h
@@ -37,6 +37,7 @@ public:
 
     // override base class virtuals
     virtual void SetItemLabel(const wxString& strName) override;
+    virtual void SetBitmap(const wxBitmapBundle& bitmap) override;
 
     virtual void Enable(bool bDoEnable = true) override;
     virtual void Check(bool bDoCheck = true) override;
--- a/src/osx/cocoa/dataview.mm
+++ b/src/osx/cocoa/dataview.mm
@@ -3287,7 +3287,7 @@ bool wxDataViewIconTextRenderer::MacRender()
     cell = (wxImageTextCell*) GetNativeData()->GetItemCell();
     iconText << GetValue();
     const wxDataViewCtrl* const dvc = GetOwner()->GetOwner();
-    [cell setImage:iconText.GetBitmapBundle().GetBitmapFor(dvc).GetNSImage()];
+    [cell setImage:wxOSXGetImageFromBundle(iconText.GetBitmapBundle())];
     [cell setStringValue:wxCFStringRef(iconText.GetText()).AsNSString()];
     return true;
 }
@@ -3401,7 +3401,7 @@ bool wxDataViewCheckIconTextRenderer::MacRender()
     {
         wxNSTextAttachmentCellWithBaseline* const attachmentCell =
             [[wxNSTextAttachmentCellWithBaseline alloc]
-             initImageCell: icon.GetBitmapFor(GetOwner()->GetOwner()).GetNSImage()];
+             initImageCell: wxOSXGetImageFromBundle(icon)];
         NSTextAttachment* const attachment = [NSTextAttachment new];
         [attachment setAttachmentCell: attachmentCell];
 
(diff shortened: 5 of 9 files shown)


vslavik:
One more thing: this restores the logic to pre- 337940f and does the same as the other wxMenuItem::Set/Update methods do:
```cxx
void wxMenuItem::SetBitmap(const wxBitmapBundle& bitmap)
{
    wxMenuItemBase::SetBitmap(bitmap);
    UpdateItemBitmap();
}

void wxMenuItem::UpdateItemBitmap()
{
    if ( !m_parentMenu )  // ??????????
        return;

    if ( m_bitmap.IsOk() )
    {
        GetPeer()->SetBitmap(m_bitmap);
    }
}
```
However, @csomor, I don't understand why the `!m_parentMenu` check is there? I don't see any corresponding code to create native representations in `DoInsertOrAppend` (they weren't there before the reverted changes either)? It *seems* to be that this never worked correctly for detached menu items and is unnecessary: the peer object is already there and has a perfectly good `NSMenuItem`, so this check is unnecessary in all the `UpdateItemXXX` methods, no? Am I missing something here?

vadz (approved):
Thanks, the changes indeed make sense to me and it's, of course, nice to simplify the code while fixing bugs.

Concerning `NSImage` in the public API: it looked wrong to me to provide access to impl-specific details in `wxBitmapBundler` itself. If we need a way to do it from outside wx, perhaps we should add `void* wxBitmapBundleImpl::GetNativeHandle() const { return nullptr; }` and override it in `wxOSXImageBundleImpl` to return the `NSImage`? We could probably also get rid of `gs_nativeImages` then...

vslavik:
> If we need a way to do it from outside wx, perhaps we should add void* wxBitmapBundleImpl::GetNativeHandle() const { return nullptr; } 

Honestly, why reinvent the wheel? Anything untyped or different from `wxBitmap` is not worth the hassle. That is:

```cxx
#ifdef __WXOSX__
WX_NSImage GetNSImage() const  { return wxOSXGetImageFromBundle(*this); }
static wxBitmapBundle FromNSImage(WX_NSImage img) { return wxOSXMakeBundleFromImage(img); }
#endif
```
This allows you to use `NSImage` easily :`SetBitmap([NSImage imageName:@"..."]);`

wxOSX is a special case here. It is the only platform where native handle makes sense, so why over-generalize it?

> and override it in wxOSXImageBundleImpl to return the NSImage? 

Note the that the global mapping - as far as I could understand - is not limited to `wxOSXImageBundleImpl` and the design of these classes necessitates hooking OSX-specific code into other implementations (`wxBitmapBundleImplSet` here):
https://github.com/wxWidgets/wxWidgets/blob/fd0c23ee24c37035a4f87c963e4844b1e0bdb231/src/common/bmpbndl.cpp#L236-L238

vadz:
> Honestly, why reinvent the wheel?

Having the accessor function in `wxBitmapBundle` itself means that we'll have to always use `gs_nativeImages` which doesn't seem great (adding a static `FromNSImage()` is fine though).

> Note the that the global mapping - as far as I could understand - is not limited to wxOSXImageBundleImpl and the design of these classes necessitates hooking OSX-specific code into other implementations (wxBitmapBundleImplSet here):

I would like to change this too, of course, but I'm too afraid of touching this code...

csomor:
> However, @csomor, I don't understand why the `!m_parentMenu` check is there? I don't see any corresponding code to create native representations in `DoInsertOrAppend` (they weren't there before the reverted changes either)? It _seems_ to be that this never worked correctly for detached menu items and is unnecessary: the peer object is already there and has a perfectly good `NSMenuItem`, so this check is unnecessary in all the `UpdateItemXXX` methods, no? Am I missing something here?

No, you are not :-) we are missing the Carbon implementation where there was no native menu item and everything had to be done via the parent menu. So these checks are not needed neither for Cocoa nor UIKit.

csomor (approved):
Thank you, looks fine to me, I have not yet discovered where we really do lose the native NSImage that was stored in the wxBitmap, but I think having the bundle in the API makes things more safe in regard to keeping multi-res native. So +1 from me

vadz:
Vaclav, do you want to remove the parent menu check as part of this PR or should I merge/push it as is? There is no hurry either way, I just don't want to want this prematurely nor to leave it unmerged for too long unnecessarily.

vslavik:
> Vaclav, do you want to remove the parent menu check as part of this PR or should I merge/push it as is? 

Sorry, I was AFK. I added the removal in 4437050 now.

> I have not yet discovered where we really do lose the native NSImage that was stored in the wxBitmap

This was because the `SetupBitmaps` code called `wxMenuItemBase::GetBitmap()`, which under the hood uses `GetBitmapFromBundle()`, which eventually calls this code where forcing scale factor does the destruction in`wxBitmapBundle::GetBitmap(size)`:
https://github.com/wxWidgets/wxWidgets/blob/170c38631458dcef648c31b9d98dcc20a8c57f9c/src/common/bmpbndl.cpp#L587-L590

But I didn't verify this hypothesis once I realized the fix was to sidestep all this...

vslavik:
> Having the accessor function in wxBitmapBundle itself means that we'll have to always use gs_nativeImages which doesn't seem great (adding a static FromNSImage() is fine though).

I don't see how the global map is related to having `wxOSXGetImageFromBundle` exposed publicly? Note that the map and all return values are [implicitly ref-counted](https://clang.llvm.org/docs/AutomaticReferenceCounting.html) by the Objective-C++ compiler and any user of the API will necessarily invoke it from Obj-C++, so while it *looks* like passing raw pointers, it's best thought of as `shared_ptr`, i.e. it would be possible to reimplement `NSView* GetNSImage()` w/o the global map.

> > Note the that the global mapping - as far as I could understand - is not limited to wxOSXImageBundleImpl and the design of these classes necessitates hooking OSX-specific code into other implementations (wxBitmapBundleImplSet here):
>
> I would like to change this too, of course, but I'm too afraid of touching this code...

I'm not a fan of neither the global map nor this, but AFAIK both are consequences of 
1. the design choice to allow multiple implementations, instead of having one per platform and 
2. the fact that you will always need *efficient* way to convert any `wxBitmapBundle` instance into `NSImage` on macOS
 
"Efficient" implies you can't just do it every time a conversion is done. And point 1 forces the implementation to be off-band and not part of `wxOSXImageBundleImpl` ...

vadz:
> * the design choice to allow multiple implementations, instead of having one per platform and

This was my decision and I still think it makes sense, at least for the other platforms that don't provide any standard/native way of doing what this class does. But it would almost certainly be awkward to use a different approach in wxOSX.
 
> * the fact that you will always need _efficient_ way to convert any `wxBitmapBundle` instance into `NSImage` on macOS

We could add an extra virtual member function returning `NSImage` to `wxBitmapBundleImpl` under Mac to make it possible to handle this efficiently in `wxOSXImageBundleImpl` or am I missing something?

Anyhow, for now I'll merge this already, thanks again!

PR #25746, opened 2025-08-31
CMake: Install wx/config and wx/include into wxBUILD_INSTALL_LIBRARY_DIR

cjmayo:
I've tried this on 3.3.1 and 3.2.8.1 + f5ee95f88c6c9806a1b938cf39f4e2ffea0ceebc.

This plus #25736 is enough for me to install a copy of wxWidgets that I can build applications on with existing scripting.

Some cautions: 
- `DESTINATION "lib/wx/config"` was a relative path - the change doesn't seem to matter
- There are still some uses of `lib/wx/config/` in  `wx-config-inplace.in`, `build/tools/build-wxwidgets.py` and `build/cmake/config.cmake`

Is a trailing `/` guaranteed in `library_dir` or was that just me?

### DIFF
--- a/build/cmake/install.cmake
+++ b/build/cmake/install.cmake
@@ -44,9 +44,9 @@ if(WIN32_MSVC_NAMING)
     # non-existent path when only Release or Debug build has been installed
     set(lib_unicode "u")
     install(DIRECTORY
-        DESTINATION "lib/${wxPLATFORM_LIB_DIR}/${wxBUILD_TOOLKIT}${lib_unicode}")
+        DESTINATION "${library_dir}/${wxBUILD_TOOLKIT}${lib_unicode}")
     install(DIRECTORY
-        DESTINATION "lib/${wxPLATFORM_LIB_DIR}/${wxBUILD_TOOLKIT}${lib_unicode}d")
+        DESTINATION "${library_dir}/${wxBUILD_TOOLKIT}${lib_unicode}d")
     install(
         DIRECTORY "${wxSETUP_HEADER_PATH}"
         DESTINATION "lib/${wxPLATFORM_LIB_DIR}")
--- a/build/cmake/install.cmake
+++ b/build/cmake/install.cmake
@@ -49,11 +49,11 @@ if(WIN32_MSVC_NAMING)
         DESTINATION "${library_dir}/${wxBUILD_TOOLKIT}${lib_unicode}d")
     install(
         DIRECTORY "${wxSETUP_HEADER_PATH}"
-        DESTINATION "lib/${wxPLATFORM_LIB_DIR}")
+        DESTINATION "${library_dir}")
 else()
     install(
         DIRECTORY "${wxSETUP_HEADER_PATH}"
-        DESTINATION "lib/wx/include")
+        DESTINATION "${library_dir}/wx/include")
 
     install(
         FILES "${wxOUTPUT_DIR}/wx/config/${wxBUILD_FILE_ID}"
--- a/build/cmake/install.cmake
+++ b/build/cmake/install.cmake
@@ -64,7 +64,7 @@ else()
     install(DIRECTORY DESTINATION "bin")
     install(CODE "execute_process( \
         COMMAND ${CMAKE_COMMAND} -E create_symlink \
-        \"\$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/lib/wx/config/${wxBUILD_FILE_ID}\" \
+        \"${CMAKE_INSTALL_PREFIX}/lib/wx/config/${wxBUILD_FILE_ID}\" \
         \"\$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/wx-config\" \
         )"
     )
--- a/build/cmake/install.cmake
+++ b/build/cmake/install.cmake
@@ -36,6 +36,8 @@ if(MSVC)
     )
 endif()
 
+wx_get_install_dir(library "lib")
+
 # setup header and wx-config
 if(WIN32_MSVC_NAMING)
     # create both Debug and Release directories, so CMake doesn't complain about
@@ -55,7 +57,7 @@ else()
 
     install(
         FILES "${wxOUTPUT_DIR}/wx/config/${wxBUILD_FILE_ID}"
-        DESTINATION "lib/wx/config"
+        DESTINATION "${library_dir}/wx/config"
         PERMISSIONS OWNER_EXECUTE OWNER_WRITE OWNER_READ
                     GROUP_EXECUTE GROUP_READ
                     WORLD_EXECUTE WORLD_READ
@@ -64,7 +66,7 @@ else()
     install(DIRECTORY DESTINATION "bin")
     install(CODE "execute_process( \
         COMMAND ${CMAKE_COMMAND} -E create_symlink \
-        \"${CMAKE_INSTALL_PREFIX}/lib/wx/config/${wxBUILD_FILE_ID}\" \
+        \"${CMAKE_INSTALL_PREFIX}/${library_dir}/wx/config/${wxBUILD_FILE_ID}\" \
         \"\$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/wx-config\" \
         )"
     )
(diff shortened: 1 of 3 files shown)


MaartenBent:
> DESTINATION "lib/wx/config" was a relative path - the change doesn't seem to matter

All install destinations should be relative paths to the  `CMAKE_INSTALL_PREFIX` directory.
Only `install(CODE ...)` needs full paths.

You removed `${CMAKE_INSTALL_PREFIX}` from the command creating the symlink. Is this intended? It seems wrong to me because then it creates a symlink to a target that will only exists when installing directly into `/`.

Instead `$ENV{DESTDIR}` should be removed. After reading again about `DESTDIR`, I think I erroneously added it (in #25466) so f5ee95f88c6c9806a1b938cf39f4e2ffea0ceebc should be reverted. 

> There are still some uses of lib/wx/config/ in wx-config-inplace.in, build/tools/build-wxwidgets.py and build/cmake/config.cmake

The first one is used by `configure` too, is that broken as well?
I don't know about `build-wxwidgets.py`, but based on its git history it is an unmaintained file from before wxWidgets 2.9.

`build/cmake/config.cmake` should indeed be fixed. I think the following lines should also use the result of `wx_get_install_dir`:
```cmake
    set(libdir "\${exec_prefix}/lib")
    set(bindir "\${exec_prefix}/bin")
```


> Is a trailing / guaranteed in library_dir or was that just me?

No, it is not guaranteed. `wx_get_install_dir` only returns a folder name. By default it is the 'lib' or 'bin' argument that is given as second argument. If you specify a different folder using `wxBUILD_INSTALL_[RUNTIME/LIBRARY/ARCHIVE]_DIR` options, this folder is returned.

cjmayo:
> > DESTINATION "lib/wx/config" was a relative path - the change doesn't seem to matter
> 
> All install destinations should be relative paths to the `CMAKE_INSTALL_PREFIX` directory. Only `install(CODE ...)` needs full paths.
> 
> You removed `${CMAKE_INSTALL_PREFIX}` from the command creating the symlink. Is this intended? It seems wrong to me because then it creates a symlink to a target that will only exists when installing directly into `/`.

I had just copied my configuration over from using `--libdir` with autotools and hence
 `-DwxBUILD_INSTALL_LIBRARY_DIR=/usr/lib64` which then lead me to remove `${CMAKE_INSTALL_PREFIX}`.

After switching to `-DwxBUILD_INSTALL_LIBRARY_DIR=lib64` all the libraries are now installing in my build directory - so I need to investigate further and then update this PR.

cjmayo:
Some updates - I force pushed because the change is so small and it seemed simpler.

- Added the revert of https://github.com/wxWidgets/wxWidgets/commit/f5ee95f88c6c9806a1b938cf39f4e2ffea0ceebc first.

- Put ` ${CMAKE_INSTALL_PREFIX}` back.

I believe I've solved the installing in the build directory #25756 

The second part of `wx_get_install_dir` was adding the trailing `/`:

```cmake
    if(wxBUILD_INSTALL_PLATFORM_SUBDIR)
        if(${artifact}_dir)
            wx_string_append(${artifact}_dir ${GEN_EXPR_DIR})
        endif()
        wx_string_append(${artifact}_dir "${wxPLATFORM_LIB_DIR}")
    endif()
```
I can prevent that by specifying `-DwxBUILD_INSTALL_PLATFORM_SUBDIR=OFF`

I'm not sure how to proceed with wx-config-inplace.in and build/cmake/config.cmake. I haven't really understood what they are doing yet.

MaartenBent:
> I can prevent that by specifying -DwxBUILD_INSTALL_PLATFORM_SUBDIR=OFF

That should not be necessary. I see it indeed adds a trailing slash via `${GEN_EXPR_DIR}` when `wxPLATFORM_LIB_DIR` is empty (i.e. for non-msvc builds).
I think the best fix would be to only add it when `wxPLATFORM_LIB_DIR` is used, by changing the first `if` to:
```cmake
if(wxBUILD_INSTALL_PLATFORM_SUBDIR AND wxPLATFORM_LIB_DIR)
```

> I'm not sure how to proceed with wx-config-inplace.in and build/cmake/config.cmake. I haven't really understood what they are doing yet.

`build/cmake/config.cmake` set all the variables that are needed for `wx-config.in` and `wx-config-inplace.in`. All the `@variable_name@` in these files are placeholders that get replaced by its actual value when `wx-config` and `wx-config-inplace` is created.
So if the wrong lib folder is used in these files, the variable in CMake needs to get the correct value and then that will be filled in in the generated file.

In the other pull request you mentioned you compared the output againts autotools/configure. The goal for CMake should be to have the same output.

MaartenBent:
After seeing the errors in #25757, I think `GEN_EXPR_DIR` is not needed at all. It is only used for the build output directory. It is not relevant for installing. 
```cmake
    if(wxBUILD_INSTALL_PLATFORM_SUBDIR AND wxPLATFORM_LIB_DIR)
        wx_string_append(${artifact}_dir "/${wxPLATFORM_LIB_DIR}")
    endif()
```

cjmayo:
Some more added. That achieves a few things for me:

```diff
--- gtk3-unicode-3.3
+++ gtk3-unicode-3.3
@@ -404,7 +404,7 @@
 # Determine the base directories we require.
 prefix=${input_option_prefix-${this_prefix:-/usr}}
 exec_prefix=${input_option_exec_prefix-${input_option_prefix-${this_exec_prefix:-${prefix}}}}
-wxconfdir="${exec_prefix}/lib/wx/config"
+wxconfdir="${exec_prefix}/lib64/wx/config"
 
 installed_configs=`cd "$wxconfdir" 2> /dev/null && ls | grep -v "^inplace-"`
 
@@ -940,7 +940,7 @@
 exec_prefix=${this_exec_prefix-$exec_prefix}
 
 includedir="${prefix}/include"
-libdir="${exec_prefix}/lib"
+libdir="${exec_prefix}/lib64"
 bindir="${exec_prefix}/bin"
 
 # Trivial queries we can answer now.
```

I don't understand  wx-config-inplace.in so that commit is a guess. With autotools does it reflect `--libdir='${prefix}'/64` in  wx-config-inplace.in?

cjmayo:
I've got my head round inplace now (I was 100% focussed on what ended up in DESTDIR and hadn't looked into the build directory).

In the build directory when using `-DwxBUILD_INSTALL_LIBRARY_DIR=lib64` both files are still under lib/wx/config:

```
$ ls lib/wx/config/
gtk3-unicode-3.3  inplace-gtk3-unicode-3.3
```

No need for changes related to inplace. Deleted the last commit.

vadz:
@MaartenBent I'd like a confirmation from you before merging this please. It looks fine to me FWIW but this doesn't mean much concerning CMake...

build/cmake/install.cmake
```diff
@@ -36,6 +36,8 @@ if(MSVC)
     )
 endif()
 
+wx_get_install_dir(library "lib")
```
MaartenBent, line 39: I think you can also use `${library_dir}` in the `WIN32_MSVC_NAMING` section instead of the current `lib/${wxPLATFORM_LIB_DIR}`.

And `${wxSETUP_HEADER_PATH}` is still installed in `"lib/wx/include"`, should this also be `${library_dir}`? What does configure do?

cjmayo: > I think you can also use `${library_dir}` in the `WIN32_MSVC_NAMING` section instead of the current `lib/${wxPLATFORM_LIB_DIR}`.

Added a commit for that.

> And `${wxSETUP_HEADER_PATH}` is still installed in `"lib/wx/include"`, should this also be `${library_dir}`? What does configure do?

configure installs `/usr/lib64/wx/include/gtk3-unicode-3.3/wx/setup.h`. Added a commit for that.

MaartenBent (approved):
I think it looks good now.

vadz:
@cjmayo Would you like this to be merged as is or should I squash merge it? I.e. are the intermediate commits important to preserve?

cjmayo:
> @cjmayo Would you like this to be merged as is or should I squash merge it? I.e. are the intermediate commits important to preserve?

Squashing is find by me.

Might want to change the title as it now also covers wx-config and the header.
It does start with a revert which gives a link back to where it was first introduced, if you want that history.

vadz:
> Might want to change the title as it now also covers wx-config and the header.

Could you please propose the right commit message to use then?

cjmayo:
> > Might want to change the title as it now also covers wx-config and the header.
> 
> Could you please propose the right commit message to use then?

PR title updated.

PR #25754, opened 2025-09-02
Catch change of monitor scaling under GTK+

### LINKED ISSUE #25733: No Re-Layout on wxGTK when the DPI changes
### Description
In my application I have a dialog with the notebook and a buttons below. This dialog is tall and will be cut short on higher DPI.

When the DPI changes I'd expect the dialog to be Re-layout so it is displayed in full or in short.

#### Stack trace:

N/A
It is visual issue.

I ave only one laptop with X11/GNOME and the monitor there supports 100 and 200 scaling.

When the program have a tall dialog it will not be relayout on the DPI change

Following scenario happening:

1. Set the monitor to 200%
2. Start the program.
3. Go to the dialog
4. Change the DPI to 100%
5. Observe that the buttons are not displayed correctly.
6. Exit the app and start it again leaving the DPI at 100.
7. Go to the dialog
8. Observe that the button displays correctly.

In one of the panels of the notebook there is a GtkFontPanel class so the dialog is very tall.
But when the DPI is going to 100% the dialog is not resized to display the buttons.

I will see if I can make a patch to the dialog sample with the reprod, but at this point I'd like to know if this behavior is correct.

I can also try to make a screenshot of my application dialog on both 200%, 100% and how it behaves when started on 100%.

If this how the program should behave on wxGTK, I can catch DPI change event at call Layout there, probably conditionally.

- wxWidgets version  you use:3.3.1
- OS  and its version: Gentoo Linux
  
  + GTK version:  3.24
  + Desktop environment : GNOME
  + Current theme:  Default (Adwaita)

RobertRoeb:
Allows to relayout controls when scaling is changed in large monitor Previous code assumed the configure event would be triggered by GTK+ Hopefully fixes #25733

### DIFF
--- a/include/wx/gtk/toplevel.h
+++ b/include/wx/gtk/toplevel.h
@@ -96,6 +96,7 @@ public:
     virtual void GTKHandleRealized() override;
 
     void GTKConfigureEvent(int x, int y);
+    void GTKScaleFactorChanged();
 
     // do *not* call this to iconize the frame, this is a private function!
     void SetIconizeState(bool iconic);
--- a/src/gtk/toplevel.cpp
+++ b/src/gtk/toplevel.cpp
@@ -333,7 +333,15 @@ gtk_frame_configure_callback( GtkWidget*,
 }
 }
 
-void wxTopLevelWindowGTK::GTKConfigureEvent(int x, int y)
+extern "C" {
+static void
+notify_gtk_scale_factor(GObject*, GParamSpec*, wxTopLevelWindowGTK* win)  // "notify::scale-factor"
+{
+    win->GTKScaleFactorChanged();
+}
+}
+
+void wxTopLevelWindowGTK::GTKScaleFactorChanged()
 {
 #ifdef __WXGTK3__
     // First of all check if our DPI has changed.
@@ -349,6 +357,10 @@ void wxTopLevelWindowGTK::GTKConfigureEvent(int x, int y)
         WXNotifyDPIChange(oldScaleFactor, newScaleFactor);
     }
 #endif // __WXGTK3__
+}
+
+void wxTopLevelWindowGTK::GTKConfigureEvent(int x, int y)
+{
 
     wxPoint point;
 #ifdef GDK_WINDOWING_X11
@@ -840,6 +852,9 @@ bool wxTopLevelWindowGTK::Create( wxWindow *parent,
     // for wxMoveEvent
     g_signal_connect (m_widget, "configure_event",
                       G_CALLBACK (gtk_frame_configure_callback), this);
+
+    g_signal_connect_after(m_widget, "notify::scale-factor",
+        G_CALLBACK(notify_gtk_scale_factor), this);
 
     // activation
     g_signal_connect_after (m_widget, "focus_in_event",


paulcor:
The `scale-factor` notification is coming from "configure-event" handling anyway:
https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkwindow.c?ref_type=heads#L7975
so I don't think this PR would make any difference.

The current code works fine for me changing the scale programmatically with `gdk_x11_display_set_window_scale()`, and we have no independent verification that it doesn't work, the details provided in #25733 being questionable.

oneeyeman1:
@paulcor ,
Did you try what I described?

Build the library with --enable-debug --with-gtk
Build the dialog sample.
Change the monitor scale to 200%
Start the sample and run the Rich text dialog.
Change the monitor scale to 100 leaving the dialog open.
Observe the size of the dialog didnt change.

As said - im running only one monitor with different scales. Using X with GNOME non-classic interface (no menu bar at the top).

Can you run this and see the problem?

Thx.

RobertRoeb:
Strange - I did test the code on my Ubuntu and choosing a different scale in the GNOME control centre did not cause a configure event, but new code was triggered.

vadz:
FWIW this is not going to fix #25733 on its own anyhow, the problem is not the lack of event (if it's lacking) but that we don't change the TLW size when the DPI changes and it's not obvious whether we should do it and how.

Also, I do get DPI change events when dragging the window from one monitor (using 200% DPI) to the other (using 100%), this can be seen just by adding

```cpp
    Bind(wxEVT_DPI_CHANGED, [](wxDPIChangedEvent& event) {
        wxLogDebug("DPI changed, new DPI: %d", event.GetNewDPI().x);
        event.Skip();
    });
```

to `MyFrame` ctor in the minimal sample.

I don't use GNOME, however, so I can't (easily) test changing the scale using its control center. Maybe it doesn't result in a reconfigure event?

vadz:
I'm not sure what to do about this one.

@RobertRoeb Can you please confirm that adding the snippet above to minimal sample doesn't show you "DPI changed" messages without this PR but does it with it?

oneeyeman1:
@vadz ,
I'd also ask Paul if he can reproduce it without this PR, and cant with it.

Thank you.

RobertRoeb:
I checked again with fresh rebuilds - and my observation was wrong. When changing the scaling using the GNOME control centre, the event get triggered in both arms, printing out the same numbers (new DPI 192) and when going back (new DPI 96). We can close this PR.

oneeyeman1:
@RobertRoeb ,
Did you try the instructions I provided in the bug report for reproduce case?
Do you see the problem?

Thank you.

RobertRoeb:
This PR attempted to correct catching a signal when the DPI changes, but it didn't change anything in my test case of using the GNOME control centre. I don't have multiple monitors to test anything and I didn't hear from you that the patch fixed anything

oneeyeman1:
@RobertRoeb ,
I’m sorry I didn’t have time to look at it.

I’m getting ready to move as I lost my contract.

Can you run the instructions I provided in the ticket with and without this PR? I suspect that only frame windows are handled and not the dialogs.

BTW, as I wrote in the ticket, I also don’t have second monitor to test with and use GNOME Display Settings to change scale. I’m using mode GNOME UI with X. (Not classic one with the menu bar on top).

Thank you.

oneeyeman1:
> FWIW this is not going to fix #25733 on its own anyhow, the problem is not the lack of event (if it's lacking) but that we don't change the TLW size when the DPI changes and it's not obvious whether we should do it and how.
> 
> Also, I do get DPI change events when dragging the window from one monitor (using 200% DPI) to the other (using 100%), this can be seen just by adding
> 
> ```c++
>     Bind(wxEVT_DPI_CHANGED, [](wxDPIChangedEvent& event) {
>         wxLogDebug("DPI changed, new DPI: %d", event.GetNewDPI().x);
>         event.Skip();
>     });
> ```
> 
> to `MyFrame` ctor in the minimal sample.
> 
> I don't use GNOME, however, so I can't (easily) test changing the scale using its control center. Maybe it doesn't result in a reconfigure event?

Vadim,
I just added that handler and I also did get that event:

```
igor@WaylandGnome ~/wxWidgets/buildGTK/samples/minimal $ ./minimal 
15:47:56: Debug: DPI changed, new DPI 96
15:50:35: Debug: DPI changed, new DPI 192
15:54:29: Debug: DPI changed, new DPI 96
```

Did you try the dialog sample with the ReachToolTip as described in the bug report?

What WM do you use?

Thank you.

PR #25756, opened 2025-09-02
CMake: Fix wxBUILD_INSTALL_*_DIRs being converted to absolute paths

cjmayo:
When the user specifies a relative directory CMake is prepending the build directory to the variable because it is of type PATH.

They should be relative to CMAKE_INSTALL_PREFIX.


vadz:
Thanks, the explanation makes sense to me.

@MaartenBent OK to apply this one?

On an unrelated note, any idea why do your PRs (and AFAICS only yours) result in this Circle CI error?

MaartenBent (approved):
Yes, looks good to me!

PR #25757, opened 2025-09-02
CMake: Install cmake files into a versioned directory

cjmayo:
Under wxBUILD_INSTALL_LIBRARY_DIR.

---

To allow both wxWidgets 3.2 and 3.3 to be installed at the same time. That seems like it is going to be useful - I tried compiling 2 applications against 3.3 and both failed, although I could force one to complete using `-DwxNO_REQUIRE_LITERAL_MSGIDS`.

CMake can handle this, including finding a specific version e.g.:

```
cmake_minimum_required(VERSION 3.10)
project(test-wx-cmake)
find_package(wxWidgets 3.3 REQUIRED COMPONENTS net core base CONFIG)
message("wxWidgets_VERSION " ${wxWidgets_VERSION})
```


vadz:
Just in case you didn't notice it, these changes result in many CI errors.

cjmayo:
Installation fixed on MS Windows.

Just [MSW/MSVC wxMSW](https://github.com/wxWidgets/wxWidgets/actions/runs/17442782455/job/49529621047?pr=25757#logs) finding the files when testing now.

MaartenBent:
I think it fails now because it installs the cmake files in `lib/vc_x64_lib/cmake/wxWidgets-3.3`. It should be in `lib/cmake/wxWidgets-3.3`. Otherwise the default [search rules](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure) won't find it.

So in this case `wx_get_install_dir` should not append `wxPLATFORM_LIB_DIR`.

Maybe you can temporary disable `wxBUILD_INSTALL_PLATFORM_SUBDIR` so it doesn't get added.
```cmake
set(install_subdir_temp ${wxBUILD_INSTALL_PLATFORM_SUBDIR})
set(wxBUILD_INSTALL_PLATFORM_SUBDIR OFF)
wx_get_install_dir(library "lib")
set(wxBUILD_INSTALL_PLATFORM_SUBDIR ${install_subdir_temp})
```

Or alternatively add a 3th parameter to `wx_get_install_dir` with `TRUE/FALSE` to indicate if it should be added.

cjmayo:
As far as I can see you can't have default values for arguments. I suggest splitting into two macros instead, new commit:

CMake: Add a wx_get_build_install_dir macro

On MSW/MSVC with wxMSW .cmake files were being installed into:
C:/Program Files (x86)/wxWidgets/lib/vc_x64_dll/cmake/wxWidgets-3.3/

which is not in the search path.

Add a wx_get_build_install_dir macro which allows a directory to be
retrived without wxPLATFORM_LIB_DIR appeneded, even on platforms that
use it.

vadz:
Thinking more about this, I don't know if we're really supposed to do this, shouldn't it be possible to use 3.3 CMake file to find both 3.3 and 3.2? And could this break something relying on finding the files in the directory without version in it? And which version will be used if there is 3.2 version of CMake files in the unversioned directory and 3.3 one in the versioned directory?

On an unrelated note, if you find any unintentional (i.e. not documented on the top of the change log) incompatible changes in 3.3, please open issues for them.

cjmayo:
> Thinking more about this, I don't know if we're really supposed to do this, shouldn't it be possible to use 3.3 CMake file to find both 3.3 and 3.2?

No, they are unique per version. Just wxWidgetsConfigVersion.cmake for example:

```diff
@@ -10,13 +10,13 @@
 # The variable CVF_VERSION must be set before calling configure_file().
 
 
-set(PACKAGE_VERSION "3.2.8")
+set(PACKAGE_VERSION "3.3.1")
 
 if(PACKAGE_VERSION VERSION_LESS PACKAGE_FIND_VERSION)
   set(PACKAGE_VERSION_COMPATIBLE FALSE)
 else()
 
-  if("3.2.8" MATCHES "^([0-9]+)\\.([0-9]+)")
+  if("3.3.1" MATCHES "^([0-9]+)\\.([0-9]+)")
     set(CVF_VERSION_MAJOR "${CMAKE_MATCH_1}")
     set(CVF_VERSION_MINOR "${CMAKE_MATCH_2}")
 
@@ -27,7 +27,7 @@
       string(REGEX REPLACE "^0+" "" CVF_VERSION_MINOR "${CVF_VERSION_MINOR}")
     endif()
   else()
-    set(CVF_VERSION_MAJOR "3.2.8")
+    set(CVF_VERSION_MAJOR "3.3.1")
     set(CVF_VERSION_MINOR "")
   endif()
 ```

> And could this break something relying on finding the files in the directory without version in it? And which version will be used if there is 3.2 version of CMake files in the unversioned directory and 3.3 one in the versioned directory?

For the directory name all that matters is that is starts `wxWidgets`. CMake looks for `name*`:
https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure
And as above it's all about the contents of the .cmake files.

MaartenBent:
Yes, in general finding wxWidgets like this should still work fine.

Only existing projects that have configured `wxWidgets_DIR` to point directly to `wxinstalldir/lib/cmake/wxWidgets` will not find it anymore. But normally it should just point to `wxinstalldir` and then it should find the new directory automatically.

I tested what happens when using `find_package(wxWidgets CONFIG)` when there are `wxWidgets`, `wxWidgets-3.2` and `wxWidgets-3.3` and it seems to always find unnumbered first, and then `3.2`. So results are probably ordered alphanumeric, not by latest available version.
When specifying the desired version (`find_package(wxWidgets 3.3 CONFIG)`) it finds the desired one.

So this change should probably be in the changelog. Maybe even as an incompatible change?
Just to inform people that install/overwrite wxWidgets in the same directory, to remove the old non-versioned directory.

PR #25758, opened 2025-09-02
More CppUnit tests upgraded to Catch2

Blake-Madden:


tests/fontmap/fontmaptest.cpp
```diff
@@ -26,29 +26,7 @@
...
-
-// register in the unnamed registry so that these tests are run by default
-CPPUNIT_TEST_SUITE_REGISTRATION( FontMapperTestCase );
-
-// also include in its own registry so that these tests can be run alone
-CPPUNIT_TEST_SUITE_NAMED_REGISTRATION( FontMapperTestCase, "FontMapperTestCase" );
-
-
-void FontMapperTestCase::NamesAndDesc()
+TEST_CASE("Names And Desc", "[fontmapper]")
```
vadz, line 29: I'd rather avoid using spaces in the test names as this makes it less convenient to run them from command line (they need to be quoted). And it is also convenient to use common prefix for the related tests. So maybe something like this:

```suggestion
TEST_CASE("wxFontMapper::NamesAndDesc", "[fontmapper]")
```

tests/formatconverter/formatconvertertest.cpp
```diff
@@ -54,155 +53,185 @@ using std::string;
...
-                                      const wxString& expectedWcharUnix,
-                                      const wxString& expectedWcharWindows);
-};
-
-void FormatConverterTestCase::format_d()
+    // all of them are unused in some build configurations
+    wxUnusedVar(expectedScanf);
+    wxUnusedVar(expectedUtf8);
+    wxUnusedVar(expectedWcharUnix);
+    wxUnusedVar(expectedWcharWindows);
```
vadz, line 65: It could be useful to use `INFO()` to log all of those: this will give useful information in case of a test failure and will avoid the need for these lines.

tests/formatconverter/formatconvertertest.cpp
```diff
@@ -54,155 +53,185 @@ using std::string;
...
+    CHECK(wxString::Format(wxT("%05d"), 255) == wxT("00255"));
+    CHECK(wxString::Format(wxT("% 5d"), 255) == wxT("  255"));
+    CHECK(wxString::Format(wxT("% 5d"), -255) == wxT(" -255"));
+    CHECK(wxString::Format(wxT("%-5d"), -255) == wxT("-255 "));
+    CHECK(wxString::Format(wxT("%+5d"), 255) == wxT(" +255"));
+    CHECK(wxString::Format(wxT("%*d"), 5, 255) == wxT("  255"));
 }
 
-void FormatConverterTestCase::format_hd()
+TEST_CASE("format_hd", "[formatconverter]")
```
vadz, line 147: For consistency I'd use

```suggestion
TEST_CASE("Format::hd", "[formatconverter]")
```

here and below.

tests/geometry/region.cpp
```diff
@@ -55,63 +55,39 @@ std::ostream& operator<<(std::ostream& os, const wxRegion& rgn)
...
+TEST_CASE("Validity", "[region]")
 {
     wxRegion r;
 
-    CPPUNIT_ASSERT_MESSAGE
+    CHECK_FALSE
     (
-        "Default constructed region must be invalid",
-        !r.IsOk()
+        // Default constructed region must be invalid
```
vadz, line 64: Minor, but I don't think it makes sense to keep comments inside `CHECK`s: messages had to be there because they were used by the macro, but this is ignored by `CHECK`, so could just as well move it above it to make the check shorter and more readable.

tests/graphics/affinematrix.cpp
```diff
@@ -27,30 +27,7 @@
...
-};
-
-// register in the unnamed registry so that these tests are run by default
-CPPUNIT_TEST_SUITE_REGISTRATION( AffineTransformTestCase );
-
-// also include in its own registry so that these tests can be run alone
-CPPUNIT_TEST_SUITE_NAMED_REGISTRATION( AffineTransformTestCase, "AffineTransformTestCase" );
-
-void AffineTransformTestCase::InvertMatrix()
+TEST_CASE("Invert Matrix", "[affine-transform]")
```
vadz, line 30:
```suggestion
TEST_CASE("AffineTransform::InvertMatrix", "[affine-transform]")
```

tests/thread/tls.cpp
```diff
@@ -72,54 +72,31 @@ class TLSTestThread : public wxThread
 // test class
 // ----------------------------------------------------------------------------
 
-class TLSTestCase : public CppUnit::TestCase
+TEST_CASE("Int", "[tls]")
```
vadz, line 75:
```suggestion
TEST_CASE("TLS::Int", "[tls]")
```

vadz (commented):
Thanks for the updates!

Could you please just make the names of the tests more consistent with the existing ones?

lanurmi:
FYI: You do *not* need to create a new PR if you want to change something in an existing one. Just push/force-push to your branch.

PR #25759, opened 2025-09-02
Upgrade tests to Catch2

Blake-Madden:


Blake-Madden:
No idea why adding CHECK in a file causes floating point errors in unrelated files. Will try to do this PR more piecemeal.

lanurmi:
FYI: You do *not* need to create a new PR if you want to change something in an existing one. Just push/force-push to your branch.

PR #25760, opened 2025-09-02
Upgrade tests to Catch2

Blake-Madden:
Here is my batch of Catch2 upgrades for the week, ready for review.


vadz:
Thanks! Merged with a couple of other minor changes for the things I've noticed while looking at this.

PR #25763, opened 2025-09-04
CMake: Support finding system libwebp using pkg-config

cjmayo:
Installing libwebp using autotools is still supported.

---

The autotools libwebp build allows control over individual CPU flags, whereas the CMake one has a SIMD option which then automatically selects the CPU flags supported on the build machine.

### DIFF
--- a/build/cmake/modules/FindWebP.cmake
+++ b/build/cmake/modules/FindWebP.cmake
@@ -1,0 +1,8 @@
+find_package(WebP CONFIG)
+
+if(NOT WebP_FOUND)
+    find_package(PkgConfig)
+    if(PKG_CONFIG_FOUND)
+        pkg_check_modules(WebP libwebp)
+    endif()
+endif()
--- a/build/cmake/modules/FindWebP.cmake
+++ b/build/cmake/modules/FindWebP.cmake
@@ -1,7 +1,7 @@
 find_package(WebP CONFIG)
 
 if(NOT WebP_FOUND)
-    find_package(PkgConfig)
+    find_package(PkgConfig QUIET)
     if(PKG_CONFIG_FOUND)
         pkg_check_modules(WebP libwebp)
     endif()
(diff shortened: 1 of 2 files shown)


build/cmake/modules/FindWebP.cmake
```diff
@@ -0,0 +1,8 @@
+find_package(WebP CONFIG)
+
+if(NOT WebP_FOUND)
+    find_package(PkgConfig)
```
vadz: We do it like this elsewhere AFAICS:
```suggestion
    find_package(PkgConfig QUIET)
```

vadz (commented):
Maybe it's just me, but I find it confusing to use `find_package(WebP)` but then actually using `pkg_check_modules()` for it under the hood. It would, IMHO, be more clear to do it directly in `build/cmake/lib/webp.cmake`.

build/cmake/modules/FindWebP.cmake
cjmayo: Commit added.

cjmayo:
> Maybe it's just me, but I find it confusing to use `find_package(WebP)` but then actually using `pkg_check_modules()` for it under the hood. It would, IMHO, be more clear to do it directly in `build/cmake/lib/webp.cmake`.

Not just you, that was my first thought - but it needs to be this way to make this work:
https://github.com/wxWidgets/wxWidgets/blob/df9d6af1bda7dfd224dd03df57ef5f63bfb315e5/build/cmake/options.cmake#L148

vadz:
Thanks for the link, I indeed didn't know that `wx_add_thirdparty_library()` unconditionally used `find_package()`. But this just results in a new question: why do we use `find_package()` in `build/cmake/lib/webp.cmake`, shouldn't we remove it from there?

cjmayo:
> why do we use `find_package()` in `build/cmake/lib/webp.cmake`, shouldn't we remove it from there?

Yes, commit added.

PR #25764, opened 2025-09-04
Improve HitTest in generic wxListCtrl

MaartenBent:
Detect when a checkbox is hit and return `wxLIST_HITTEST_ONITEMSTATEICON`.
Determine the correct location of the icon, it might be preceded by a checkbox.
Return only the icon rectangle in `GetSubItemRect` with `wxLIST_RECT_ICON`, not the padding or margins above/below the icon.


vadz:
Thanks for improving this!

PR #25766, opened 2025-09-07
Improve wxStatusBar appearance on macOS 11-26

vslavik:
This fixes incorrect appearance of `wxStatusBar` on modern macOS. It's a not a *dramatic* change, but still worth doing. It's a generic control for what is effectively legacy thing on macOS (only Finder uses it, literally nothing else), so I took the same approach as before, which is mimicking native appearance within reasonable limits.

It comes in two parts:

[Improve wxStatusBar appearance on macOS 11-26](https://github.com/wxWidgets/wxWidgets/commit/12dfc19f8ec7bcbf12e6bd4c17b62e2a31ca0212)

This is clear-cut, just visual tweaks. It doesn't change how things work in this part of the code, just updates the visuals. It presents some difficulty for 3.2 backporting due to new fields. Perhaps using global variables would be fine here?

[Don't put wxStatusBar fields into corners on macOS](https://github.com/wxWidgets/wxWidgets/commit/f323430a087fa389a2fa4458eec044dd59148b96)

I have mixed feelings about this one, and don't have a strong opinion on whether it should be merged (master only I guess, too difficult to backport) or rejected. Eventually, I decided I'm not going to use it myself in Poedit and that I'll want to center the statusbar, as Finder does (which is more involved than just centering, because I have a wxSplitterWindow-based sidebar in the app and the statusbar should be centered under the area that isn't the sidebar, not under the whole window).

The issue is that macOS 26 Tahoe has *very* rounded corners (macOS 11+ has too, but to much lesser extent):

toolbar_before

So for wx applications to took better, I added indentation:

toolbar_after

Notice how it is still not very good in the right-side window. There are *differently-rounded* windows in macOS 26! However, normal wxWidgets applications shouldn't be able to end with the large radius (it requires native toolbar, which is something I do on top of wxOSX), so it should be fine. There is a native API for determining the "safe area", but I think it would be overkill to use it for just a legacy control and just some obscure uses like Poedit's.

I *do* think, however, that if this is merged, it should be in this form where the offset is configurable from user code, in the spirit of doing a reasonable thing by default, but allowing being more native.

### DIFF
--- a/include/wx/generic/statusbr.h
+++ b/include/wx/generic/statusbr.h
@@ -87,6 +87,9 @@ protected:
     // returns the position and the size of the size grip
     wxRect GetSizeGripRect() const;
 
+    // returns the width available for fields drawing given total width
+    virtual int GetAvailableWidthForFields(int width) const;
+
     // common part of all ctors
     void Init();
 
--- a/include/wx/osx/statusbr.h
+++ b/include/wx/osx/statusbr.h
@@ -28,13 +28,22 @@ public:
     // Implementation
     virtual void MacHiliteChanged() override;
     void OnPaint(wxPaintEvent& event);
+    virtual bool GetFieldRect(int i, wxRect& rect) const override;
+
+    void MacSetCornerInset(int inset);
+    int MacGetCornerInset() const { return m_cornerInset; }
 
 protected:
     virtual int GetEffectiveFieldStyle(int WXUNUSED(i)) const override { return wxSB_NORMAL; }
 
     virtual void InitColours() override;
 
+    void InitCornerInset();
+
+    virtual int GetAvailableWidthForFields(int width) const override;
+
 private:
+    int m_cornerInset;
     wxColour m_textActive, m_textInactive, m_bgActive, m_bgInactive, m_separator;
 
     wxDECLARE_DYNAMIC_CLASS(wxStatusBarMac);
--- a/include/wx/osx/statusbr.h
+++ b/include/wx/osx/statusbr.h
@@ -35,7 +35,7 @@ protected:
     virtual void InitColours() override;
 
 private:
-    wxColour m_textActive, m_textInactive;
+    wxColour m_textActive, m_textInactive, m_bgActive, m_bgInactive, m_separator;
 
     wxDECLARE_DYNAMIC_CLASS(wxStatusBarMac);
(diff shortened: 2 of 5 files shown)


RobertRoeb:
This looks very good, thank you. I also support the idea to configure the indentation with platform defaults.

csomor:
@vslavik thanks a lot, looks a lot petter, apparently Tahoe has some dependency on the upper window part for the radius

https://wwdcnotes.com/documentation/wwdcnotes/wwdc25-310-build-an-appkit-app-with-the-new-design/

vslavik:
> thanks a lot, looks a lot petter, apparently Tahoe has some dependency on the upper window part for the radius

Yes, I mention it above — the top radius is a function of the presence of a native (possibly unified) toolbar, which wx doesn't use.

csomor:
> > thanks a lot, looks a lot petter, apparently Tahoe has some dependency on the upper window part for the radius
> 
> Yes, I mention it above — the top radius is a function of the presence of a native (possibly unified) toolbar, which wx doesn't use.

Actually wx does have a native toolbar implementation, perhaps not fully and as nice as you do in Poedit, but the toolbar sample also shows a unified toolbar, and has the big radius (when built against 26 SDK), so I thought you might extend your heuristics, that's why I added the link with the measurements, but I'm fine with that as well and I'll add it later if possible.

vslavik:
> Actually wx does have a native toolbar implementation,

OK, I used a different sample than "toolbar" there (because that one crashed and I didnt' have time to investigate). My bad. That would mean it makes sense to extend the heuristic, yes, but it also means it can't be be done at ctor time, when the toolbar may not exist in the parent yet :(

Thanks for correcting me!

csomor:
> OK, I used a different sample than "toolbar" there (because that one crashed and I didnt' have time to investigate). 

yes, when using Xcode you have to add the pngs and the svg and add a bundle-resource phase that merges them in, otherwise it cannot load the bitmaps

csomor:
> That would mean it makes sense to extend the heuristic, yes, but it also means it can't be be done at ctor time, when the toolbar may not exist in the parent yet :(

oh dear, yes, so perhaps just commit your current state, and we might add a handler for the ShowEvent later and adjust things there ...

vadz:
Looks good to me too, thank you! Will merge in a moment.

PR #25767, opened 2025-09-07
Assorted controls fixes for macOS 26 Tahoe

vslavik:
This PR covers the rest of Tahoe breakages I am aware of (via Poedit). It is best reviewed individually, see the commits' explanations.

This most prominently fixes single-line `wxTextCtrl`, which was outright invisible on macOS 26 Contrast-Is-For-Amateurs Tahoe. `wxSpinCtrl` was subtly broken too, but that's a much less prominent control.

The big commit is the 1st one, [Don't set NSView.clipsToBounds = YES globally](https://github.com/wxWidgets/wxWidgets/commit/3273d289d7ca710a1fd4833c748facbc248e367b) — this slightly undoes the changes to fix wxAUI rendering (and no doubt much more, #23916 was just the canary), by limiting  the `clipsToBounds` change to non-native windows only.

The alternative would be to do it individually for controls that are known to be mis-rendered, but I somewhat strongly prefer doing it like this: we should mess with native controls as little as possible. Unless I'm missing something (@csomor?) it should be safe to differentiate native-vs-wx with this in `wxWidgetCocoaImpl::wxWidgetCocoaImpl`:
https://github.com/wxWidgets/wxWidgets/blob/3273d289d7ca710a1fd4833c748facbc248e367b/src/osx/cocoa/window.mm#L2738-L2739

This assumption is crucial; if I'm wrong about that, it should be done per-control (i.e. so far at least for `wxTextCtrl` and `wxSpinButton`).

Screenshot before:

before_full

And after:

after_full

And  just the spin control without focus ring:

Before:

before_spin

After:

after_spin

In the process of doing this, I spotted some inaccuracies on macOS 15 too, mostly in sizing, so I fixed them there too; it is obvious from the commits where.

---

For [Fix wxTextCtrl sizes on macOS, particularly 26](https://github.com/wxWidgets/wxWidgets/commit/26ee6d248b6e5c96f811ffb34a67c73168157643) my methodology for verifying sizes was as follows:
1. Create test view in wx for measuring, and also create a native xib in Interface Builder, using Xcode 16 for macOS 15 and Xcode 26 for Tahoe, to avoid Xcode 26's metrics changes.
2. I then measures heights of focused text controls with xScope, and also verified regular and small changes against controls present in Apple Safari and Apple Mail.
3. After verifying all sizes were correct, I commented out the code to call `GetPeer()->GetBestSize()` to make sure the fallback switch statement was executed, and again measured all controls on both macOS versions.

Which is to say that I'm reasonably confident in them.

### DIFF
--- a/src/generic/spinctlg.cpp
+++ b/src/generic/spinctlg.cpp
@@ -49,11 +49,15 @@ wxIMPLEMENT_DYNAMIC_CLASS(wxSpinDoubleEvent, wxNotifyEvent);
 // ----------------------------------------------------------------------------
 
 // The margin between the text control and the spin: the value here is the same
-// as the margin between the spin button and its "buddy" text control in wxMSW
-// so the generic control looks similarly to the native one there, we might
-// need to use different value for the other platforms (and maybe even
-// determine it dynamically?).
+// as the margin between the spin button and its "buddy" text control in wxMSW,
+// and the commonly used spacing on macOS, so the generic control looks
+// similarly to the native one there, we might need to use different value for
+// other platforms (and maybe even determine it dynamically?).
+#ifdef __WXOSX__
+static const wxCoord MARGIN = 4;
+#else
 static const wxCoord MARGIN = 1;
+#endif
 
 #define SPINCTRLBUT_MAX 32000 // large to avoid wrap around trouble
 
@@ -240,6 +244,10 @@ bool wxSpinCtrlGenericBase::Create(wxWindow *parent,
         if ( DoTextToValue(value, &d) )
             m_value = AdjustAndSnap(d);
     }
+
+#ifdef __WXOSX__
+    MacClipsToBounds(false);
+#endif
 
     m_textCtrl   = new wxSpinCtrlTextGeneric(this, DoValueToText(m_value), style);
     m_spinButton = new wxSpinCtrlButtonGeneric(this, style);
--- a/src/osx/cocoa/window.mm
+++ b/src/osx/cocoa/window.mm
@@ -2734,7 +2734,9 @@ wxWidgetImpl( peer, flags )
     if ( m_osxView )
         CFRetain(m_osxView);
     [m_osxView release];
-    m_osxView.clipsToBounds = YES;
+
+    if ( IsUserPane() )
+        ClipsToBounds(true);
 }
 
 
@@ -4290,6 +4292,11 @@ void wxWidgetCocoaImpl::UseClippingView()
         }
     }
 #endif
+}
+
+void wxWidgetCocoaImpl::ClipsToBounds(bool clip)
+{
+    m_osxView.clipsToBounds = clip;
 }
 
 
--- a/include/wx/osx/cocoa/private.h
+++ b/include/wx/osx/cocoa/private.h
@@ -226,6 +226,8 @@ public :
     // from the same pimpl class.
     virtual void                controlTextDidChange();
 
+    virtual void                ClipsToBounds(bool clip) override;
+
     virtual void                AdjustClippingView(wxScrollBar* horizontal, wxScrollBar* vertical) override;
     virtual void                UseClippingView() override;
     virtual WXWidget            GetContainer() const override { return m_osxClipView ? m_osxClipView : m_osxView; }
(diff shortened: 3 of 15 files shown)


csomor:
> The alternative would be to do it individually for controls that are known to be mis-rendered, but I somewhat strongly prefer doing it like this: we should mess with native controls as little as possible. Unless I'm missing something (@csomor?) it should be safe to differentiate native-vs-wx with this in `wxWidgetCocoaImpl::wxWidgetCocoaImpl`:
> 
> https://github.com/wxWidgets/wxWidgets/blob/3273d289d7ca710a1fd4833c748facbc248e367b/src/osx/cocoa/window.mm#L2738-L2739
> 
> This assumption is crucial; if I'm wrong about that, it should be done per-control (i.e. so far at least for `wxTextCtrl` and `wxSpinButton`).

yes, that's correct, you might also just use the `m_isUserPane` that is already set at that moment using the same condition as your code uses

csomor (approved):
Thanks @vslavik, yes, I think it's fine to go ahead, and collect feedback whether we might have to fine tune because there is some nesting of controls that might bite us

csomor (approved):

csomor (approved):

csomor:
one comment: when running an app that was built on earlier SDK and is now running on Tahoe, I think, the view is different, ie 'old', so I'd prefer to bracket the `WX_IS_MACOS_AVAILABLE` with the `__MAC_OS_X_VERSION_MAX_ALLOWED `

```
#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 260000
    if ( WX_IS_MACOS_AVAILABLE(26, 0) ) 
```

src/osx/cocoa/window.mm
```diff
@@ -2734,7 +2734,9 @@ void wxOSXCocoaClassAddWXMethods(Class c, wxOSXSkipOverrides skipFlags)
     if ( m_osxView )
         CFRetain(m_osxView);
     [m_osxView release];
-    m_osxView.clipsToBounds = YES;
+
+    if ( flags & Widget_IsUserPane )
```
vadz: Very minor, but if you amend this to use `IsUserPane()` instead, please also do `s/they're/their/` in the commit message as this triggers my OCD :-/

src/generic/spinctlg.cpp
```diff
@@ -49,11 +49,15 @@ wxIMPLEMENT_DYNAMIC_CLASS(wxSpinDoubleEvent, wxNotifyEvent);
...
 
 // The margin between the text control and the spin: the value here is the same
-// as the margin between the spin button and its "buddy" text control in wxMSW
-// so the generic control looks similarly to the native one there, we might
-// need to use different value for the other platforms (and maybe even
-// determine it dynamically?).
+// as the margin between the spin button and its "buddy" text control in wxMSW,
+// and the commonly used spacing on macOS, so the generic control looks
+// similarly to the native one there, we might need to use different value for
+// other platforms (and maybe even determine it dynamically?).
```
vadz, line 55: Maybe we ought to move this into `wxRendererNative`... But this can definitely be done after merging this PR, of course.

vadz:
No real comments from, but not merging yet just in case you decide to amend it following Stefan's comments.

Thanks a lot for fixing all these problems, I realize this must have required spending quite some time on them!

vslavik:
> one comment: when running an app that was built on earlier SDK and is now running on Tahoe, I think, the view is different, ie 'old', so I'd prefer to bracket the WX_IS_MACOS_AVAILABLE with the __MAC_OS_X_VERSION_MAX_ALLOWED 

You're right, thanks! I assumed this wouldn't trigger if compiled with older compiler, but it does. That's a problem, because code compiled with Xcode pre-26 uses old metrics and appearance.

csomor (approved):
apart from the build-time condition I mentioned, this looks good to me!
@vslavik, thanks a lot for investing your time on these important details

vslavik:
Updated and rebased onto `master` so that I can add the `__MAC_OS_X_VERSION_MAX_ALLOWED` checks to statusbar code as well.

Thanks for the reviews!

vadz:
Merged, thanks again!

csomor:
> Thanks for the reviews!

thanks for your work 🙏

PR #25769, opened 2025-09-07
Use wayland-scanner at build time instead of pre-generated files

vadz:
Should fix problems due to the pre-generated files in the repository being incompatible with the build system Wayland version. It also allows to drop some hacks that were put in place to deal with such incompatibilities before.

@MaartenBent I'm not sure if using an `ALL` custom target is really a good idea, maybe we should just generate these files during configure step, what do you think?


MaartenBent:
Indeed, probably better to create them during configuring. I think that is what `configure / AC_CONFIG_COMMANDS` does too? 

If not, you'll need to add an dependency of `core` to this custom target, because these files need to be created before building `core`. In the `core` cmake file something like:
```cmake
wx_add_dependencies(wxcore wayland_protocols)
```

I haven't checked the details of this (or the previous) PR, but does the generated header need to be installed as well or is it only used privately when building?

Most of the `wxUSE_*` variables also have an entry in `setup.h[.in]`. Would that be useful here too?

vadz:
> Indeed, probably better to create them during configuring. I think that is what `configure / AC_CONFIG_COMMANDS` does too?

Yes, it does, but it also allows you to recreate them later by running `./config.status wayland-protocols` in the build directory, which is why I wanted to have a matching CMake target. But as Wayland XML files change very rarely, if ever, it's probably not necessary after all.
 
> If not, you'll need to add an dependency of `core` to this custom target

This is not really relevant any more, but I thought that `ALL` custom targets were made before anything else, weren't they?

> I haven't checked the details of this (or the previous) PR, but does the generated header need to be installed as well or is it only used privately when building?

No, it's only used privately.
 
> Most of the `wxUSE_*` variables also have an entry in `setup.h[.in]`. Would that be useful here too?

I decided it wasn't, as I don't see what would it be useful for, but if anybody thinks we need it we can always do it later.

I suspect that [this CI failure](https://github.com/wxWidgets/wxWidgets/actions/runs/17534666873/job/49796172463?pr=25769) is due to `WAYLAND_CLIENT_CFLAGS` containing something other than definitions, so I'm probably going to have to revert this commit...

vadz:
Continuing the last remark, I can confirm that under Ubuntu 24.04 I get

```
-- WAYLAND_CLIENT_CFLAGS=-I/usr/include
-- WAYLAND_CLIENT_INCLUDE_DIRS=/usr/include
```

so we definitely can't put `CFLAGS` into `wxTOOLKIT_DEFINITIONS`. I'll probably just ignore them, I don't see why would wayland-client need to provide any definitions...

MaartenBent:
> This is not really relevant any more, but I thought that ALL custom targets were made before anything else, weren't they?

No I don't think so. It doesn't add a dependency. As far as I know it just means this target will be built when you build the full project, the order is not defined. But since it has no dependencies on its own, it is usually built first.

If I only build core (`cmake --build . --target core`) it won't build this custom target, only core and it dependencies (base, jpeg, ...).

> but it also allows you to recreate them later

I think cmake will create the files every time you (re)configure, or when a changed cmake file triggers a reconfigure.

> so we definitely can't put CFLAGS into wxTOOLKIT_DEFINITIONS. I'll probably just ignore them, I don't see why would wayland-client need to provide any definitions

I agree.

vadz:
This should be good to try now, @Blake-Madden could you please confirm that it fixes the problem you reported? TIA!

Blake-Madden:
@vadz Success! Using this PR's branch in my GH action fixed it. Thanks!

PR #25771, opened 2025-09-08
Suppress false positive warning from MSVC

Blake-Madden:
When compiling a release build in MSVC, I sometimes get a (false positive) warning about this line:

`*value = static_cast<LongestValueType>(uvalue);`

saying that `uvalue` may be uninitialized. `FromString` actually sets this value, but initializing this value to zero probably isn't a bad idea and will suppress the warning.


vadz:
Thanks, will merge soon.

PR #25773, opened 2025-09-08
Define main as 'extern "C"'

LegalizeAdulthood:
Fixes #25772

### DIFF
--- a/include/wx/app.h
+++ b/include/wx/app.h
@@ -878,7 +878,7 @@ public:
 #if defined(__WINDOWS__)
     #ifdef __VISUALC__
         #define wxIMPLEMENT_WXWIN_MAIN_CONSOLE                                \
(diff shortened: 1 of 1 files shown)


LegalizeAdulthood:
See the discussion on the bug; if it's preferred that the constraints on `wxIMPLEMENT_APP` be mentioned in the documentation instead, I'll submit a different pull request approaching things from that perspective.

vadz:
I think

```
D:\a\wxWidgets\wxWidgets\samples\archive\archive.cpp:354:1: error: 'main' should not be 'extern "C"' [-Werror,-Wmain]
```

from the CI failure is pretty self-explanatory, so this clearly shouldn't be applied.

LegalizeAdulthood:
Yes, I'll make a PR with a documentation change instead.  Thanks.

PR #25775, opened 2025-09-08
Document that wxIMPLEMENT_APP/wxDECLARE_APP should be called at global scope

LegalizeAdulthood:


vadz:
Thanks, will apply soon, although I think I'll change "invoked" to "used" as I think it would be easier to understand for an average reader.

LegalizeAdulthood:
I can update the commit and PR if you like, it's an easy change.

LegalizeAdulthood:
OK, never mind I see on the bug you've already done it `:)`

PR #25778, opened 2025-09-09
OSX - support POWER events suspend(ed)/resume in wxWdgets

tisj:
Register observers with notifications center, with callbacks for NSWorkspaceWillSleepNotification and NSWorkspaceDidWakeNotification notifications.

These callbacks inform the top level window to handle POWER SUSPENDED and/or RESUME.

### DIFF
--- a/src/osx/cocoa/power.mm
+++ b/src/osx/cocoa/power.mm
@@ -13,6 +13,11 @@
 
 #include "wx/wxprec.h"
 
+#ifndef WX_PRECOMP
+  #include "wx/module.h"
+  #include "wx/window.h"
+#endif
+
 #include "wx/power.h"
 #include "wx/atomic.h"
 #include "wx/osx/private.h"
@@ -105,3 +110,82 @@ void wxPowerResource::Release(wxPowerResourceKind kind)
             break;
     }
 }
+
+#import <Cocoa/Cocoa.h>
+#import "wx/power.h" // Assuming you have wxPowerEvent defined or a custom event
+
+@interface CocoaPowerEventsObserver : NSObject
+@end
+
+@implementation CocoaPowerEventsObserver
+
+- (id)init {
+    self = [super init];
+    if (self) {
+        NSNotificationCenter *center = [[NSWorkspace sharedWorkspace] notificationCenter];
+
+        // Register for "will sleep" notification
+        [center addObserver:self
+                   selector:@selector(handleWillSleep:)
+                       name:NSWorkspaceWillSleepNotification
+                     object:nil];
+
+        // Register for "did wake" notification
+        [center addObserver:self
+                   selector:@selector(handleDidWake:)
+                       name:NSWorkspaceDidWakeNotification
+                     object:nil];
+    }
+    return self;
+}
+
+- (void)dealloc {
+    [[[NSWorkspace sharedWorkspace] notificationCenter] removeObserver:self];
+    [super dealloc];
+}
+
+- (void)handleWillSleep:(NSNotification *)notification {
+    wxPowerEvent event(wxEVT_POWER_SUSPENDED);
+
+    for ( auto tlw : wxTopLevelWindows )
+        tlw->HandleWindowEvent(event);
+}
+
+- (void)handleDidWake:(NSNotification *)notification {
+    wxPowerEvent event(wxEVT_POWER_RESUME);
+
+    for ( auto tlw : wxTopLevelWindows )
+        tlw->HandleWindowEvent(event);
+}
+
+@end
+
+namespace {
+// add power observers on init and remove on exit
+class wxCocoaPowerModule : public wxModule
+{
+wxDECLARE_DYNAMIC_CLASS(wxCocoaPowerModule);
+
+public:
+    bool OnInit() override
+    {
+         m_powerObserver = [[CocoaPowerEventsObserver alloc] init];
+         return true;
+    }
+
+    void OnExit() override
+    {
+         if (m_powerObserver != nullptr)
+         {
+             [m_powerObserver release];
+             m_powerObserver = nullptr;
+         }
+    }
+
+private:
+    CocoaPowerEventsObserver* m_powerObserver;
+};
+
+wxIMPLEMENT_DYNAMIC_CLASS(wxCocoaPowerModule, wxModule)
+
+}; // anonymous namespace
--- a/src/osx/cocoa/power.mm
+++ b/src/osx/cocoa/power.mm
@@ -114,10 +114,10 @@ void wxPowerResource::Release(wxPowerResourceKind kind)
 #import <Cocoa/Cocoa.h>
 #import "wx/power.h" // Assuming you have wxPowerEvent defined or a custom event
 
-@interface CocoaPowerEventsObserver : NSObject
+@interface wxCocoaPowerEventsObserver : NSObject
 @end
 
-@implementation CocoaPowerEventsObserver
+@implementation wxCocoaPowerEventsObserver
 
 - (id)init {
     self = [super init];
@@ -169,7 +169,7 @@ wxDECLARE_DYNAMIC_CLASS(wxCocoaPowerModule);
 public:
     bool OnInit() override
     {
-         m_powerObserver = [[CocoaPowerEventsObserver alloc] init];
+         m_powerObserver = [[wxCocoaPowerEventsObserver alloc] init];
          return true;
     }
 
@@ -183,7 +183,7 @@ public:
     }
 
 private:
-    CocoaPowerEventsObserver* m_powerObserver;
+    wxCocoaPowerEventsObserver* m_powerObserver;
 };
 
 wxIMPLEMENT_DYNAMIC_CLASS(wxCocoaPowerModule, wxModule)
--- a/src/osx/cocoa/power.mm
+++ b/src/osx/cocoa/power.mm
@@ -80,28 +80,8 @@ static wxAtomicInt g_powerResourceSystemRefCount = 0;
 @end
 
 namespace {
-// wrapper around obj-c interface, usable as std::unique_ptr
-class wxCocoaPowerEventsObserverWrapper
-{
-public:
-    wxCocoaPowerEventsObserverWrapper()
-        : m_powerEventsObserver( [[wxCocoaPowerEventsObserver alloc] init] )
-    {
-    }
-
-    ~wxCocoaPowerEventsObserverWrapper()
-    {
-        [m_powerEventsObserver release];
-    }
-
-private:
-    wxCocoaPowerEventsObserver* m_powerEventsObserver;
-
-    wxDECLARE_NO_COPY_CLASS(wxCocoaPowerEventsObserverWrapper);
-};
-
 // the global power events observer
-static std::unique_ptr<wxCocoaPowerEventsObserverWrapper> g_powerEventsObserver;
+static wxCFRef<wxCocoaPowerEventsObserver*> g_powerEventsObserver;
 
 // add power observers on init and remove on exit
 class wxCocoaPowerModule : public wxModule
@@ -113,10 +93,7 @@ public:
         // There should be no other threads by now, so no locking is needed.
         if ( g_powerResourceSystemRefCount )
         {
-            if ( g_powerEventsObserver )
-            {
-                delete g_powerEventsObserver.release();
-            }
+            g_powerEventsObserver.reset();
 
             g_powerResourceSystemRefCount = 0;
         }
@@ -164,9 +141,9 @@ bool UpdatePowerResourceUsage(wxPowerResourceKind kind, const wxString& reason)
             [[NSProcessInfo processInfo]
              endActivity:(id)g_processInfoActivity];
             g_processInfoActivity = nil;
-
-            return true;
         }
+
+        g_powerEventsObserver.reset();
     }
 
     return true;
@@ -183,9 +160,9 @@ wxPowerResource::Acquire(wxPowerResourceKind kind,
         {
             wxAtomicInc(g_powerResourceSystemRefCount);
 
-            if ( !g_powerEventsObserver )
+            if ( g_powerEventsObserver.get() == nullptr )
             {
-                g_powerEventsObserver.reset( new wxCocoaPowerEventsObserverWrapper() );
+                g_powerEventsObserver = [[wxCocoaPowerEventsObserver alloc] init];
             }
         }
 
--- a/src/osx/cocoa/power.mm
+++ b/src/osx/cocoa/power.mm
@@ -25,94 +25,13 @@
 
 #include <IOKit/pwr_mgt/IOPMLib.h>
 
-// ----------------------------------------------------------------------------
-// wxPowerResource
-// ----------------------------------------------------------------------------
+#import <Cocoa/Cocoa.h>
 
 static wxAtomicInt g_powerResourceSystemRefCount = 0;
-static NSObject* g_processInfoActivity = nil;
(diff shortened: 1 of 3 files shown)


src/osx/cocoa/power.mm
```diff
@@ -105,3 +110,81 @@ bool UpdatePowerResourceUsage(wxPowerResourceKind kind, const wxString& reason)
             break;
     }
 }
+
+#import <Cocoa/Cocoa.h>
+#import "wx/power.h" // Assuming you have wxPowerEvent defined or a custom event
+
+@interface CocoaPowerEventsObserver : NSObject
```
vadz: This probably should be called `wxCocoaPowerEventsObserver` for consistency and also maybe to avoid any conflicts with application-defined identifier when using static linking (I'm not sure what are the visibility rules for Objective C interfaces, but using "wx" prefix can never hurt).

src/osx/cocoa/power.mm
```diff
@@ -105,3 +110,81 @@ bool UpdatePowerResourceUsage(wxPowerResourceKind kind, const wxString& reason)
...
+public:
+    bool OnInit() override
+    {
+         m_powerObserver = [[CocoaPowerEventsObserver alloc] init];
+         return true;
+    }
+
+    void OnExit() override
+    {
+         if (m_powerObserver != nullptr) {
```
vadz: Minor style pick, wx style puts braces on their own lines:

```suggestion
         if (m_powerObserver != nullptr)
         {
```

vadz (commented):
Nice, thanks for implementing this!

The documentation in `interface/wx/power.h` should be updated to mention that this functionality is now available under macOS too.

I also wonder if we could/should delay `CocoaPowerEventsObserver` construction until `wxPowerResourceBlocker` is instantiated: this is required to get these events under Linux and even if it isn't under macOS, it could be nice to encourage writing more portable code. What do you think?

src/osx/cocoa/power.mm
tisj: I didn't use wx prefix here since it is a local symbol.

These are the symbols in an application binary: 

> 00000001002fa570 t -[CocoaPowerEventsObserver dealloc]
> 00000001002fa6d0 t -[CocoaPowerEventsObserver handleDidWake:]
> 00000001002fa5e0 t -[CocoaPowerEventsObserver handleWillSleep:]
> 00000001002fa4b0 t -[CocoaPowerEventsObserver init]
> 000000010077c420 S _OBJC_CLASS_$_CocoaPowerEventsObserver
> 000000010077c448 S _OBJC_METACLASS_$_CocoaPowerEventsObserver
> 00000001007791d8 s __OBJC_$_INSTANCE_METHODS_CocoaPowerEventsObserver
> 0000000100779240 s __OBJC_CLASS_RO_$_CocoaPowerEventsObserver
> 0000000100779190 s __OBJC_METACLASS_RO_$_CocoaPowerEventsObserver
> 00000001002ca30c t -[CocoaPowerEventsObserver dealloc]
> 00000001002ca468 t -[CocoaPowerEventsObserver handleDidWake:]
> 00000001002ca37c t -[CocoaPowerEventsObserver handleWillSleep:]
> 00000001002ca244 t -[CocoaPowerEventsObserver init]
> 0000000100713b98 S _OBJC_CLASS_$_CocoaPowerEventsObserver
> 0000000100713bc0 S _OBJC_METACLASS_$_CocoaPowerEventsObserver
> 00000001004cd798 s __OBJC_$_INSTANCE_METHODS_CocoaPowerEventsObserver
> 0000000100710ab0 s __OBJC_CLASS_RO_$_CocoaPowerEventsObserver
> 0000000100710a68 s __OBJC_METACLASS_RO_$_CocoaPowerEventsObserver

However  wx modules also seem local..

> 0000000100787be8 b __ZN12_GLOBAL__N_118wxCocoaPowerModule12ms_classInfoE
> 00000001002fa790 t __ZN12_GLOBAL__N_118wxCocoaPowerModule14wxCreateObjectEv
> 0000000100787c18 b __ZN12_GLOBAL__N_118wxCocoaPowerModule15m_powerObserverE
> 00000001002fa850 t __ZN12_GLOBAL__N_118wxCocoaPowerModule6OnExitEv
> 00000001002fa820 t __ZN12_GLOBAL__N_118wxCocoaPowerModule6OnInitEv
> 00000001002fa7e0 t __ZN12_GLOBAL__N_118wxCocoaPowerModuleD0Ev
> 00000001002fa7d0 t __ZN12_GLOBAL__N_118wxCocoaPowerModuleD1Ev
> 00000001002fa810 t __ZNK12_GLOBAL__N_118wxCocoaPowerModule12GetClassInfoEv
> 0000000100718750 s __ZTIN12_GLOBAL__N_118wxCocoaPowerModuleE
> 00000001005949a8 s __ZTSN12_GLOBAL__N_118wxCocoaPowerModuleE
> 0000000100718708 s __ZTVN12_GLOBAL__N_118wxCocoaPowerModuleE
> 000000010071f220 b __ZN12_GLOBAL__N_118wxCocoaPowerModule12ms_classInfoE
> 00000001002ca51c t __ZN12_GLOBAL__N_118wxCocoaPowerModule14wxCreateObjectEv
> 000000010071f250 b __ZN12_GLOBAL__N_118wxCocoaPowerModule15m_powerObserverE
> 00000001002ca5b0 t __ZN12_GLOBAL__N_118wxCocoaPowerModule6OnExitEv
> 00000001002ca57c t __ZN12_GLOBAL__N_118wxCocoaPowerModule6OnInitEv
> 00000001002ca558 t __ZN12_GLOBAL__N_118wxCocoaPowerModuleD0Ev
> 00000001002ca554 t __ZN12_GLOBAL__N_118wxCocoaPowerModuleD1Ev
> 00000001002ca570 t __ZNK12_GLOBAL__N_118wxCocoaPowerModule12GetClassInfoEv
> 00000001006b0500 s __ZTIN12_GLOBAL__N_118wxCocoaPowerModuleE
> 0000000100536800 s __ZTSN12_GLOBAL__N_118wxCocoaPowerModuleE
> 00000001006b04b8 s __ZTVN12_GLOBAL__N_118wxCocoaPowerModuleE

We actually had a problem using the power management in static builds, as the module and observer were not getting linked in automatically. Only when we used the (wxPowerResource) acquire and release, all symbols in the same file got included (and thus the module is registered with wx).

tisj:
If i understand it correctly the wxResourcePowerBlocker is defined in include/wx/power.h and gets instantiated when the application includes this file (directly or indirectly)?

In Linux these events are observed on the DBus system bus via login manager, but that implementation resides in src/gtk/power.cpp - and thus part of wx library.

No sure how the magic is happening here ;)

src/osx/cocoa/power.mm
vadz: Just to confirm: are you sure that if the application defines an interface called `CocoaPowerEventsObserver` it doesn't conflict with this one? I think Objective C uses strings for matching somewhere, so I'd really prefer to just prefer this type to have a `wx` prefix than to think carefully about whether it is or not a problem.

vadz:
> If i understand it correctly the wxResourcePowerBlocker is defined in include/wx/power.h and gets instantiated when the application includes this file (directly or indirectly)?

No, it's just a normal class, there is no magic involved. The application needs to create an object of this class in order to receive power events under Linux/GTK, see the example in `wxPowerEvent` documentation.

My thinking is that we could also only send these events in macOS implementation if/after `wxPowerResourceBlocker` was/is created, even if this is unnecessary under this platform. But by forcing the application to do it we would

1. Avoid creating the observer unnecessarily (for the applications that don't care about such events).
2. Ensure that the code using these events developed under macOS also works under Linux (instead of silently doing nothing there).

P.S. Please don't force push until and unless you're pushing the very final version of the PR, force pushing makes it difficult to see what has changed, so please use "fixup" commits as created by `git commit --fixup` or `--squash` and then force push just once at the end. TIA!

vadz:
@tisj Just to check, do you plan to change anything here? I could merge it after renaming `CocoaPowerEventsObserver`, even if I would prefer to do what I proposed above, but I don't want to do it if you have any upcoming modifications.

src/osx/cocoa/power.mm
tisj: Just pushed a fixup

tisj:
@vadz I think I made all the changes you requested

In the last fixup, a unique instance will be created upon request (acquire of power resource system).

I used this small program to test

```c++
// For compilers that support precompilation, includes "wx/wx.h".
#include <wx/wxprec.h>

#include <wx/frame.h>
#include <wx/power.h>

#ifndef WX_PRECOMP
    #include <wx/wx.h>
#endif

class MyApp : public wxApp
{
public:
    virtual bool OnInit();
};

class MyMainFrame : public wxFrame
{
public:
    MyMainFrame();

private:
    wxPowerResourceBlocker m_powerDelaySleep;
};
 
MyMainFrame::MyMainFrame()
    : wxFrame(NULL, wxID_ANY, "Hello World"),
      m_powerDelaySleep(wxPOWER_RESOURCE_SYSTEM,
                        "MyApp needs to close network connections",
                        wxPOWER_DELAY)
{
    Bind(wxEVT_POWER_SUSPENDED, [](wxPowerEvent&) {
        // Close network connections here.
        std::cout << "Power suspended, need to close nw" << std::endl;
    });
 
    Bind(wxEVT_POWER_RESUME, [](wxPowerEvent&) {
        // Restore previously closed network connections here.
        std::cout << "Power resumed, need to open nw" << std::endl;
    });
}

wxIMPLEMENT_APP(MyApp);

bool MyApp::OnInit() {
    MyMainFrame *frame = new MyMainFrame();
    frame->Show(true);
    return true;
}
```

With some debug added to src/osx/cocoa/power.mm :

> % ./test
> Power resource acquire
> Power Event observer created
> will sleep now
> Power suspended, need to close nw
> will wake up now
> Power resumed, need to open nw
> Power resource release
> Power Event observer destroyed

src/osx/cocoa/power.mm
```diff
@@ -13,18 +13,125 @@
...
+
 #include "wx/power.h"
 #include "wx/atomic.h"
 #include "wx/osx/private.h"
 #include "wx/osx/private/available.h"
 
 #include <IOKit/pwr_mgt/IOPMLib.h>
 
+#import <Cocoa/Cocoa.h>
+#import "wx/power.h" // Assuming you have wxPowerEvent defined or a custom event
```
vadz: Sorry, I don't understand why do we use `#import` with the header `#include`d above and I don't understand the comment. What's going on here?

src/osx/cocoa/power.mm
```diff
@@ -13,18 +13,125 @@
...
+    bool OnInit() override { return true; }
+    void OnExit() override
+    {
+        // There should be no other threads by now, so no locking is needed.
+        if ( g_powerResourceSystemRefCount )
+        {
+            if ( g_powerEventsObserver )
+            {
+                delete g_powerEventsObserver.release();
+            }
```
vadz: This is trivial, but it really should be done with just
```suggestion
            g_powerEventsObserver.reset();
```

src/osx/cocoa/power.mm
```diff
@@ -13,18 +13,125 @@
...
+
+@end
+
+namespace {
+// wrapper around obj-c interface, usable as std::unique_ptr
+class wxCocoaPowerEventsObserverWrapper
+{
+public:
+    wxCocoaPowerEventsObserverWrapper()
+        : m_powerEventsObserver( [[CocoaPowerEventsObserver alloc] init] )
```
vadz: This should be
```suggestion
        : m_powerEventsObserver( [[wxCocoaPowerEventsObserver alloc] init] )
```
of course.

src/osx/cocoa/power.mm
```diff
@@ -13,18 +13,125 @@
...
+    wxPowerEvent event(wxEVT_POWER_RESUME);
+
+    for ( auto tlw : wxTopLevelWindows )
+        tlw->HandleWindowEvent(event);
+}
+
+@end
+
+namespace {
+// wrapper around obj-c interface, usable as std::unique_ptr
```
vadz: All this would be unnecessary if we used `wxCFRef<wxCocoaPowerEventObserver>` directly instead. IMO it would make sense to simplify the code by using it instead of `unique_ptr<>` with this wrapper. And if you really prefer using `unique_ptr<>`, why not just use it with custom deleter instead? It would still be more concise and arguably more clear than using this wrapper.

vadz (commented):
Great, thanks for implementing this!

AFAICS this should work exactly as desired, except perhaps that we should destroy the power observer when the counter drops to 0 too.

But we obviously need to fix the build first and maybe address the remarks below.

Thanks!

src/osx/cocoa/power.mm
tisj: It was generated by gemini

image

Probably picked it up from some other project I guess

src/osx/cocoa/power.mm
tisj: I have no preference, I used it because it was done so in src/gtk/power.cpp. And I thought it was used in a few other places.. but honestly i am not too familiar with wxWidget code base.

I will look into wxCFRef

src/osx/cocoa/power.mm
tisj: :ashamed: .. I used an older implementation (before the rename) to compile and test, then cp to git and thought to have done the rename completely

src/osx/cocoa/power.mm
tisj: I removed the line

src/osx/cocoa/power.mm
tisj: done

vadz (approved):
Looks good now, thanks!

Will merge soon.

tisj:
Just a question... When you would acquire the system power resource on windows (delay),it would just return true without incrementing g_powerResourceSystemRefCount.

If then you release that resource, it will see g_powerResourceSystemRefCount is 0 and issue a fail "System power resource was not acquired", which is not desirable when writing platform independent code.

Wouldn't it be better to always increment on acquire and decrement on release?

vadz:
> Just a question... When you would acquire the system power resource on windows (delay),it would just return true without incrementing g_powerResourceSystemRefCount.
> 
> If then you release that resource, it will see g_powerResourceSystemRefCount is 0 and issue a fail "System power resource was not acquired", which is not desirable when writing platform independent code.
> 
> Wouldn't it be better to always increment on acquire and decrement on release?

Yes, indeed, there is a problem here, thanks for raising it. But handling `wxPOWER_DELAY` as `wxPOWER_PREVENT` under MSW doesn't look great either... The simplest non-horrible fix seems to be to pass `blockKind` to `Release()` too.

But maybe this just shows that having `wxPowerBlockKind` was a bad idea in the first place as it's not really orthogonal to `wxPowerResourceKind` under any platform: under Linux and macOS (with your patch) `wxPOWER_DELAY` can't be combined with `wxPOWER_RESOURCE_SCREEN`, so maybe we should just drop `wxPowerBlockKind` completely and add `wxPOWER_RESOURCE_SYSTEM_DELAY` or something like this with a better name which would work as `wxPOWER_DELAY` does now?

Any change like this would/should be done in a separate PR, of course.

tisj:
For me the blockKind parameter is not intuitive and rather confusing, as my use case is: I want to be notified about system power changes (suspend/resume) at any time, I am not "delaying" or "preventing" suspend or resume.

In that sense would i be acquiring/release system power "notifications", rather than delaying or preventing? Although technically, when handling such notifications, I probably am delaying the process ofcourse.

PR #25781, opened 2025-09-09
CMake4 Compatibility

cjmayo:
I've been using CMake 4.0.3 for all these PRs without problems.

CMakeParseArguments isn't a problem with CMake4, that is just a tidy-up.

The last one raises the minimum to CMake 3.10. This was released in late 2017. All previous versions are deprecated now. It's not strictly required at this time, although it saves a big warning from CMake 4 and presumably further changes like this for a while.


MaartenBent:
I don't see this deprecation warning with CMake 4.0. We added the `VERSION <min>...<max>` syntax to silence this deprecation warning. And it saves us from having to update the minimum version every year when a new range gets deprecated.
 Or do you see the warning in the cef/cotire script that doesn't specify this `<max>` version?

PBfordev:
I get deprecation warning when building with `wxWebViewChromium`, and assumed it coming from
https://github.com/wxWidgets/wxWidgets/blob/6f4586aa125e7e65cb9194b6dffd01415f193627/build/cmake/lib/webview_chromium/cef_download.cmake.in#L10 (see the first commit in the PR).

I did not report it but thought it could be better to set the required version only in the main CMakeLists.txt and remove it from everywhere else.

MaartenBent:
> I did not report it but thought it could be better to set the required version only in the main CMakeLists.txt and remove it from everywhere else.

Those CMake scripts are all run as independent projects, so require their own `cmake_minimum_required`.
(Except maybe `cotire.cmake`, but I'm not a fan of modifying external modules unless necessary).

cjmayo:
There was a reason why I posted this last, and listed these in order so they could be dropped in order too!

The first cotire commit one is needed if building `-DwxBUILD_PRECOMP=COTIRE`. Even that I'm not actually using so I don't personally mind about.

The second commit obviously just removes useless lines.

The third, yes as you say, I had stripped off the `...3.31` during my investigations and that's what causes the visible warnings.

So, happy to drop any or all. Just say.

MaartenBent:
The first two commits are fine. I think the third commit should add a version range to all the files instead. And this can probably be `VERSION 3.5...4.0`. Thanks!

cjmayo:
I think `3.5...4.0` might be something different. Frankly I don't understand it so I've just dropped the 3rd commit.

vadz:
Maarten, is it ok to merge this one?

MaartenBent:
Yes, this is OK to merge.

PR #25782, opened 2025-09-09
Allow for cross compilation on iPhone on Silicon Macs

RobertRoeb:
Added compile flag --enable-iphonesimulator
  Documented this build instructions

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -425,7 +425,7 @@ WX_ARG_ONLY_WITH(directfb,      [  --with-directfb         use DirectFB], [wxUSE
 WX_ARG_ONLY_WITH(x11,           [  --with-x11              use X11], [wxUSE_X11="$withval" wxUSE_UNIVERSAL="yes" CACHE_X11=1 TOOLKIT_GIVEN=1])
 WX_ARG_ONLY_WITH(qt,            [  --with-qt               use Qt], [wxUSE_QT="$withval" CACHE_QT=1 TOOLKIT_GIVEN=1])
 WX_ARG_ENABLE(nanox,       [  --enable-nanox          use NanoX], wxUSE_NANOX)
-WX_ARG_ENABLE(iphonesimulator,  [  --enable-iphonesimulator   commpile for iPhone simulator], wxUSE_IPHONESIMULATOR)
+WX_ARG_ENABLE(iphonesimulator,  [  --enable-iphonesimulator   compile for iPhone simulator], wxUSE_IPHONESIMULATOR)
 
 dnl check that no more than one toolkit is given and that if none are given that
 dnl we have a default one
--- a/configure.ac
+++ b/configure.ac
@@ -427,7 +427,6 @@ WX_ARG_ONLY_WITH(directfb,      [  --with-directfb         use DirectFB], [wxUSE
 WX_ARG_ONLY_WITH(x11,           [  --with-x11              use X11], [wxUSE_X11="$withval" wxUSE_UNIVERSAL="yes" CACHE_X11=1 TOOLKIT_GIVEN=1])
 WX_ARG_ONLY_WITH(qt,            [  --with-qt               use Qt], [wxUSE_QT="$withval" CACHE_QT=1 TOOLKIT_GIVEN=1])
 WX_ARG_ENABLE(nanox,       [  --enable-nanox          use NanoX], wxUSE_NANOX)
-WX_ARG_ENABLE(iphonesimulator,  [  --enable-iphonesimulator   commpile for iPhone simulator], wxUSE_IPHONESIMULATOR)
 
 dnl check that no more than one toolkit is given and that if none are given that
 dnl we have a default one
--- a/configure.ac
+++ b/configure.ac
@@ -425,6 +425,7 @@ WX_ARG_ONLY_WITH(directfb,      [  --with-directfb         use DirectFB], [wxUSE
 WX_ARG_ONLY_WITH(x11,           [  --with-x11              use X11], [wxUSE_X11="$withval" wxUSE_UNIVERSAL="yes" CACHE_X11=1 TOOLKIT_GIVEN=1])
 WX_ARG_ONLY_WITH(qt,            [  --with-qt               use Qt], [wxUSE_QT="$withval" CACHE_QT=1 TOOLKIT_GIVEN=1])
 WX_ARG_ENABLE(nanox,       [  --enable-nanox          use NanoX], wxUSE_NANOX)
+WX_ARG_ENABLE(iphonesimulator,  [  --enable-iphonesimulator   commpile for iPhone simulator], wxUSE_IPHONESIMULATOR)
 
 dnl check that no more than one toolkit is given and that if none are given that
 dnl we have a default one
@@ -1339,7 +1340,11 @@ fi
 
 if test "x$wxUSE_MACOSX_VERSION_MIN" != "x"; then
     if test "$wxUSE_OSX_IPHONE" = 1; then
-        MACOSX_VERSION_MIN_OPTS="-miphoneos-version-min=$wxUSE_MACOSX_VERSION_MIN"
+        if test "$wxUSE_IPHONESIMULATOR" = "yes"; then
+            MACOSX_VERSION_MIN_OPTS="-miphonesimulator-version-min=$wxUSE_MACOSX_VERSION_MIN"
+        else
+            MACOSX_VERSION_MIN_OPTS="-miphoneos-version-min=$wxUSE_MACOSX_VERSION_MIN"
+        fi
     else
         MACOSX_VERSION_MIN_OPTS="-mmacosx-version-min=$wxUSE_MACOSX_VERSION_MIN"
     fi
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -425,6 +425,7 @@ WX_ARG_ONLY_WITH(directfb,      [  --with-directfb         use DirectFB], [wxUSE
 WX_ARG_ONLY_WITH(x11,           [  --with-x11              use X11], [wxUSE_X11="$withval" wxUSE_UNIVERSAL="yes" CACHE_X11=1 TOOLKIT_GIVEN=1])
 WX_ARG_ONLY_WITH(qt,            [  --with-qt               use Qt], [wxUSE_QT="$withval" CACHE_QT=1 TOOLKIT_GIVEN=1])
 WX_ARG_ENABLE(nanox,       [  --enable-nanox          use NanoX], wxUSE_NANOX)
+WX_ARG_ENABLE(iphonesimulator,  [  --enable-iphonesimulator   commpile for iPhone simulator], wxUSE_IPHONESIMULATOR)
```
vadz:
```suggestion
WX_ARG_ENABLE(iphonesimulator,  [  --enable-iphonesimulator   compile for iPhone simulator], wxUSE_IPHONESIMULATOR)
```

docs/ios/install.md
```diff
@@ -25,3 +25,12 @@ The library can also be build via configure/make:
...
         --with-macosx-sdk=$(xcrun --sdk iphonesimulator --show-sdk-path)
     make
+
+On a recent Silicon Mac in the 3.3 branch you need this configure
+to compile for the iPhone simulator (exchange 24.6.0 with your 
+actual host system):
+
+    ../configure --with-osx_iphone --enable-iphonesimulator --enable-monolithic  \
+        --disable-shared --with-macosx-sdk=$(xcrun --sdk iphonesimulator --show-sdk-path)  \
+        --host=aarch64-apple-darwin24.6.0 --build=aarch64-apple-darwin --without-libtiff
```
vadz: Using `--host=$($wx/config.guess)` would make it unnecessary to tell people to replace 24.6.0 above and AFAICS doesn't have any real drawbacks. I'm also rather surprised that you have to specify both `host` and `build` that are essentially the same...

vadz (commented):
Thanks, I guess it's an improvement but it still doesn't look optimal to specify all of `--with-osx_iphone`, `--enable-iphonesimulator` and `--with-macosx-sdk`. I'm pretty sure we could have `--with-osx_iphonesimulator` replacing the first 2, but I also wonder if we can't either deduce that simulator should be used from the SDK provided (i.e. if it contains `iPhoneSimulator` in its name) or, maybe more logically, choose `iphonesimulator` SDK if the option to build for the simulator is used? I.e. why can't we just make specifying *only* `--with-osx_iphonesimulator` work?

docs/ios/install.md
RobertRoeb: I don't have an environment variable like "wx" set, so I don't understand how this could find the config.guess script. What did work was this:  
../wxWidgetsRR/configure --with-osx_iphone --enable-iphonesimulator --enable-monolithic  \
        --disable-shared --with-macosx-sdk=$(xcrun --sdk iphonesimulator --show-sdk-path)  \
        --host=$(../wxWidgetsRR/config.guess) --build=aarch64-apple-darwin --without-libtiff
but then I have to write the path twice. I have not found a trick to get the main directory into the options.

RobertRoeb:
The title of the last commit is wrong, it now works with just "--with-iphonesimulator" instead of two options

configure.ac
```diff
@@ -415,9 +415,11 @@ dnl shouldn't default to wxUSE_ALL_FEATURES
 AC_ARG_WITH(gtk,          [[  --with-gtk[=VERSION]    use GTK+, VERSION can be 3 (default), 2, or "any"]], [wxUSE_GTK="$withval" CACHE_GTK=1 TOOLKIT_GIVEN=1])
 WX_ARG_ONLY_WITH(osx_cocoa,     [  --with-osx_cocoa        use macOS (Cocoa)], [wxUSE_OSX_COCOA="$withval" CACHE_OSX_COCOA=1 TOOLKIT_GIVEN=1])
 WX_ARG_ONLY_WITH(osx_iphone,    [  --with-osx_iphone       use iOS], [wxUSE_OSX_IPHONE="$withval" CACHE_OSX_IPHONE=1 TOOLKIT_GIVEN=1])
+WX_ARG_ONLY_WITH(iphonesimulator, [  --with-osx_iphonesimulator  use iPhone simulator], [wxUSE_OSX_IPHONE="$withval" wxUSE_IPHONESIMULATOR="yes" CACHE_OSX_IPHONE=1 TOOLKIT_GIVEN=1])
```
vadz: Sorry, I don't understand the purpose of having 2 options doing apparently the same thing? Other than the concern of our configure script not having enough options, I really don't see why would we want to add 2 options when one (with either spelling) would do?

docs/ios/install.md
vadz: We use `$WXWIN` as env var, sorry. Even if it's not defined, it should be pretty clear what it means.

vadz (commented):
Could you please update `.github/workflows/ci_mac.yml` to use the new option? We really need a way to check that it works (and keeps working). TIA!

configure.ac
RobertRoeb: I just followed the convention from the other osx_ variants. None of those add any value and can be removed.

RobertRoeb:
> Could you please update `.github/workflows/ci_mac.yml` to use the new option? We really need a way
> to check that it works (and keeps working). TIA!

No idea what that is, but will look.

RobertRoeb:
> We use $WXWIN as env var, sorry. Even if it's not defined, it should be pretty clear what it means.

It took me a little moment to figure that out. For the sake of increasing the chance that people can compile this successfully, I prefer to use paths like this

../wxWidgets/configure --with-osx_iphone --enable-iphonesimulator --enable-monolithic
--disable-shared --with-macosx-sdk=$(xcrun --sdk iphonesimulator --show-sdk-path)
--host=$(../wxWidgets/config.guess) --build=aarch64-apple-darwin --without-libtiff

vadz:
> > We use $WXWIN as env var, sorry. Even if it's not defined, it should be pretty clear what it means.
> 
> It took me a little moment to figure that out. For the sake of increasing the chance that people can compile this successfully, I prefer to use paths like this
> 
> ../wxWidgets/configure --with-osx_iphone --enable-iphonesimulator --enable-monolithic --disable-shared --with-macosx-sdk=$(xcrun --sdk iphonesimulator --show-sdk-path) --host=$(../wxWidgets/config.guess) --build=aarch64-apple-darwin --without-libtiff

Sure, this works too. The main idea is to avoid hardcoding the OS version which will change again in a few months.

vadz:
> > Could you please update `.github/workflows/ci_mac.yml` to use the new option? We really need a way
> > to check that it works (and keeps working). TIA!
> 
> No idea what that is, but will look.

This is a YAML file describing the steps in (one of) the CI workflow(s), this one building wxiOS. This workflow runs for every PR and push to master and allows to ensure that whatever it is doing doesn't get broken by the future changes.

configure.ac
vadz: I think we used to have `--with-carbon` and `--with-cocoa` and then `--with-osx_cocoa` was added to make it more clear what to use for macOS 10 and then the former was removed and `--with-osx` added as it became unambiguous.

Anyhow, they only exist due to history, but IMO this is not a reason to keep duplicating them.

RobertRoeb: Which one do you want to keep?

vadz: I think we had originally introduced `--with-osx_xxx` to disambiguate between class Mac OS and Mac OS X (as macOS was then known), see e.g. f10f4c8785 (Rename --with-cocoa configure switch to --with-old_cocoa to avoid confusion., 2009-12-13). This distinction has, of course, been irrelevant for many years by now, so I think `--with-osx_xxx` should be considered compatibility options and we should document only `--with-xxx` ones.

RobertRoeb:
I added a wxIOSSImmulator build based on guessing parameters and updated configure and install.md. Am I supposed to commit the entire configure?

vadz:
> I added a wxIOSSImmulator build based on guessing parameters and updated configure and install.md. Am I supposed to commit the entire configure?

Yes, you have to as long as we keep it in the repository. Please see [the docs](https://github.com/wxWidgets/wxWidgets/blob/master/build/tools/autoconf/README.md) for the recommended way to update it. TIA!

RobertRoeb:
> > I added a wxIOSSImmulator build based on guessing parameters and updated configure and install.md
> > Am I supposed to commit the entire configure?
> 
> Yes, you have to as long as we keep it in the repository. Please see [the docs]
> (https://github.com/wxWidgets/wxWidgets/blob/master/build/tools/autoconf/README.md)
> for the recommended way to update it. TIA!

I tried that on my Mac and and it either asked me to connect to a Linux VM (which I don't have on
my Mac) or to install and run Docker (which seems overkill for an autoconf script, in my opinion).
So I copied the configure.ac to my Linux computer, regenerated the configure there and copied 
it back. The intro says "Generated by GNU Autoconf 2.69 for wxWidgets 3.3.2." so I think this is
right.

vadz:
Sorry, I had just implicitly assumed you had a Linux box at hand.

Anyhow, thanks, will (squash) merge soon.

Issue #25739, opened 2025-08-29
wxTextEntryDialog resizes in unexpected ways


willco007:
#### Bug description:

When using wxTextEntryDialog, you can resize the window down to 0, 0 which you shouldn't be able to do. Furthermore in single line mode you can resize it taller than the bottom buttons which causes an unexpected space at the bottom of the window.  Lastly it would be nice to have a fully non-resizable style flag as well.

#### Expected vs observed behaviour:

Expect minimum size to be the fitting size.
Expect maximum height to be to the bottom of the OK button + padding in single line mode.

Observed no min size constraint
Observed no height constraint in single line mode

#### To Reproduce:

Show text entry dialog and resize it:

```C++
wxTextEntryDialog dialog(this,
     "Join our email list.\nStay informed of the latest important news.\n\nEmail Address:",
     "Developer eList",
     "");

dialog.ShowModal()
```

### Platform and version information

- wxWidgets version you use: 3.3.1
- wxWidgets port you use: wxMSW (reproduces in all)
- OS and its version: Windows 11

vadz:
We should definitely set the min size, and this is simple to do.

As for max size, setting it is simple too but I'm not sure if it actually works, would need to check it.

Blake-Madden:
@vadz I'll take a look at this this week if you want to assign it to me.

vadz:
Sure, TIA!

PR #25784, opened 2025-09-09
Fix wxTextEntryDialog from being resizable to 0, 0 (and other tweaks)

Blake-Madden:
I don't see a way to force a max height (only size, which we don't want since the width should always be resizable). However, for a single-line dialog I fixed it so that the button row remains at the bottom of the dialog when resized. That seems much nicer looking then before.
#25739

### DIFF
--- a/src/generic/textdlgg.cpp
+++ b/src/generic/textdlgg.cpp
@@ -82,7 +82,7 @@ bool wxTextEntryDialog::Create(wxWindow *parent,
     }
 
     if (sz.IsFullySpecified())
-    SetMinSize( sz );
+        SetMinSize( sz );
 
     m_dialogStyle = style;
     m_value = value;
@@ -120,7 +120,7 @@ bool wxTextEntryDialog::Create(wxWindow *parent,
     SetSizer( topsizer );
 
     if (sz.IsFullySpecified())
-    topsizer->Fit( this );
+        topsizer->Fit( this );
     else
         topsizer->SetSizeHints( this );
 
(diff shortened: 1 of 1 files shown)


src/generic/textdlgg.cpp
```diff
@@ -81,7 +81,8 @@ bool wxTextEntryDialog::Create(wxWindow *parent,
         return false;
     }
 
-    SetMinSize( sz );
+    if (sz.IsFullySpecified())
```
vadz, line 84: Sorry, why do we need this?

src/generic/textdlgg.cpp
```diff
@@ -98,24 +99,30 @@ bool wxTextEntryDialog::Create(wxWindow *parent,
...
     {
+        if ((style & wxTE_MULTILINE) == 0)
+            topsizer->AddStretchSpacer();
         topsizer->Add(buttonSizer, wxSizerFlags().Expand().DoubleBorder());
     }
 
     SetSizer( topsizer );
 
-    topsizer->Fit( this );
+    if (sz.IsFullySpecified())
```
vadz, line 122: This also seems suspicious, why should we do this for this particular dialog and not any other ones?

vadz (commented):
Thanks, but I hoped using `wxTLW::SetMaxSize(wxSize(-1, whatever))` would work, have you tried it and confirmed that it didn't? It would be probably better to make it work, it should be possible, I think...

vadz:
Sorry, but I've just tested and `SetMaxSize()` does work, at least in wxMSW and wxGTK, so we should definitely use it to prevent from making the dialog too tall instead. If you could please make a PR doing this, it would be great, but I don't think this one is the right approach, so I'm closing it.

PR #25785, opened 2025-09-09
Introduce system color for GRIDLINES, map to NSColor gridColor on macOS.

dkulp:
This introduces a new system color enum value for wxSYS_COLOUR_GRIDLINES which maps to BTNFACE on platforms other than macOS.   On macOS, it will use the NSColor gridColor.

This also adjusts the property grid categories when in dark mode.   If the color is too dark, it will lighten it slightly.  This produces a fairly good looking property grid in dark mode:

Screenshot 2025-09-09 at 7 07 35 PM

### DIFF
--- a/include/wx/settings.h
+++ b/include/wx/settings.h
@@ -80,6 +80,7 @@ enum wxSystemColour
     wxSYS_COLOUR_MENUBAR,
     wxSYS_COLOUR_LISTBOXTEXT,
     wxSYS_COLOUR_LISTBOXHIGHLIGHTTEXT,
+    wxSYS_COLOUR_GRIDLINES,
 
     wxSYS_COLOUR_MAX,
 
@@ -91,6 +92,7 @@ enum wxSystemColour
     wxSYS_COLOUR_3DHIGHLIGHT = wxSYS_COLOUR_BTNHIGHLIGHT,
     wxSYS_COLOUR_3DHILIGHT = wxSYS_COLOUR_BTNHIGHLIGHT,
     wxSYS_COLOUR_FRAMEBK = wxSYS_COLOUR_BTNFACE
+    
 };
 
 // possible values for wxSystemSettings::GetMetric() index parameter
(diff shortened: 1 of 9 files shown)


include/wx/settings.h
```diff
@@ -91,6 +92,7 @@ enum wxSystemColour
     wxSYS_COLOUR_3DHIGHLIGHT = wxSYS_COLOUR_BTNHIGHLIGHT,
     wxSYS_COLOUR_3DHILIGHT = wxSYS_COLOUR_BTNHIGHLIGHT,
     wxSYS_COLOUR_FRAMEBK = wxSYS_COLOUR_BTNFACE
+
```
vadz:
```suggestion
```

interface/wx/settings.h
```diff
@@ -115,7 +115,13 @@ enum wxSystemColour
      */
     wxSYS_COLOUR_LISTBOXHIGHLIGHTTEXT,
 
-
+    /**
+        On macOS, this maps to [NSColor gridLines], on other platforms
+        it's mapped to wxSYS_COLOUR_BTNFACE
+        @since 3.3
```
vadz:
```suggestion
        @since 3.3.2
```

src/propgrid/propgrid.cpp
```diff
@@ -1434,18 +1434,28 @@ void wxPropertyGrid::RegainColours()
...
-            m_colCapBack = wxPGAdjustColour(col,-colDec);
+        wxColour col = wxSystemSettings::GetColour( wxSYS_COLOUR_GRIDLINES );
+    #ifdef __WXOSX__
+        if (wxSystemSettings::GetAppearance().IsDark())
+        {
+            // Make sure colour is light enough
+            int colDec = wxPGGetColAvg(col);
+            if (colDec < 30)
+                col = wxPGAdjustColour(col, 20);
+        }
```
vadz, line 1445: To be honest, I find this code quite incomprehensible, but then the existing version was already so. I wonder if we can abstract this to some function with a readable name?

vadz (commented):
I'm not sure how to reconcile this with your comment about `gridColor` not really working, but if it works, all the better!

Please just fix the build failures and then we could merge this.

Thanks!

dkulp:
> I'm not sure how to reconcile this with your comment about `gridColor` not really working, but if it works, all the better!
> 
> Please just fix the build failures and then we could merge this.
> 
> Thanks!

The main issue is that gridColor was too dark for the PropertyGrid categories.   It's "kind of" too dark (IMO) for wxGrid in dark mode, but it's at least visible.    gridColor does work quite well in light mode.  


Screenshot 2025-09-09 at 7 42 37 PM
Screenshot 2025-09-09 at 7 42 24 PM

vadz:
OK, finally got time to fix the remaining problems myself, so merging.

PR #25790, opened 2025-09-11
CMake: Set libdir for use in wx-config.in

cjmayo:
See linked issue.

### DIFF
--- a/build/cmake/config.cmake
+++ b/build/cmake/config.cmake
@@ -12,6 +12,9 @@ file(MAKE_DIRECTORY ${wxCONFIG_DIR})
 set(TOOLCHAIN_FULLNAME ${wxBUILD_FILE_ID})
 
 macro(wx_configure_script input output)
+    # variable used in wx-config.in
+    wx_get_install_dir(library "lib")
+    set(libdir "\${exec_prefix}/${library_dir}")
     # variables used in wx-config-inplace.in
     set(abs_top_srcdir ${wxSOURCE_DIR})
     set(abs_top_builddir ${wxBINARY_DIR})


MaartenBent:
I do not understand this change, `libdir` is already assigned this exact value in `wx_write_config` here:
https://github.com/wxWidgets/wxWidgets/blob/771d4df4b7019d5231787df53adab4720b05c28c/build/cmake/config.cmake#L114
And then `wx_configure_script` is called from this function.

And for `wx_write_config_inplace` this is not needed, because it doesn't contain `libdir`.

cjmayo:
Adding this patch:

```diff
--- a/build/cmake/config.cmake
+++ b/build/cmake/config.cmake
@@ -12,9 +12,11 @@ file(MAKE_DIRECTORY ${wxCONFIG_DIR})
 set(TOOLCHAIN_FULLNAME ${wxBUILD_FILE_ID})
 
 macro(wx_configure_script input output)
+    message("1 " ${output} " " ${libdir})
     # variable used in wx-config.in
     wx_get_install_dir(library "lib")
     set(libdir "\${exec_prefix}/${library_dir}")
+    message("2 "  ${output} " " ${libdir})
     # variables used in wx-config-inplace.in
     set(abs_top_srcdir ${wxSOURCE_DIR})
     set(abs_top_builddir ${wxBINARY_DIR})
```

I get:

```
1 inplace-gtk3-unicode-3.3 
2 inplace-gtk3-unicode-3.3 ${exec_prefix}/lib64
1 gtk3-unicode-3.3 ${exec_prefix}/lib
2 gtk3-unicode-3.3 ${exec_prefix}/lib64
```

cjmayo:
Yes, not needed since 771d4df4b7019d.

PR #25794, opened 2025-09-12
Fix shared CMake build using external WebP library

cjmayo:
Building an application against wxWidgets built as a shared
library using libwep found with pkg-config fails with:

```
x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libwx_gtk3u_core-3.3.so: undefined reference to `WebPDemuxDelete'
x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libwx_gtk3u_core-3.3.so: undefined reference to `WebPAnimDecoderReset'
x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libwx_gtk3u_core-3.3.so: undefined reference to `WebPAnimDecoderOptionsInitInternal'
x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libwx_gtk3u_core-3.3.so: undefined reference to `WebPDemuxInternal'
x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libwx_gtk3u_core-3.3.so: undefined reference to `WebPAnimDecoderGetNext'
x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libwx_gtk3u_core-3.3.so: undefined reference to `WebPDemuxGetI'
x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libwx_gtk3u_core-3.3.so: undefined reference to `WebPAnimDecoderHasMoreFrames'
x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libwx_gtk3u_core-3.3.so: undefined reference to `WebPAnimDecoderDelete'
x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libwx_gtk3u_core-3.3.so: undefined reference to `WebPAnimDecoderGetInfo'
x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libwx_gtk3u_core-3.3.so: undefined reference to `WebPAnimDecoderNewInternal'
x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libwx_gtk3u_core-3.3.so: undefined reference to `WebPDemuxReleaseIterator'
x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libwx_gtk3u_core-3.3.so: undefined reference to `WebPDemuxGetFrame'
```

### DIFF
--- a/build/cmake/modules/FindWebP.cmake
+++ b/build/cmake/modules/FindWebP.cmake
@@ -3,7 +3,7 @@ find_package(WebP CONFIG)
 if(NOT WebP_FOUND)
     find_package(PkgConfig QUIET)
     if(PKG_CONFIG_FOUND)
-        pkg_check_modules(WebP libwebp)
+        pkg_check_modules(WebP libwebp libwebpdemux)
     endif()
 endif()
 


build/cmake/modules/FindWebP.cmake
```diff
@@ -3,7 +3,7 @@ find_package(WebP CONFIG)
 if(NOT WebP_FOUND)
     find_package(PkgConfig QUIET)
     if(PKG_CONFIG_FOUND)
-        pkg_check_modules(WebP libwebp)
+        pkg_check_modules(WebP libwebpdemux)
```
vadz: Sorry but don't we need both `-lwebp` and `-lwebpdemux`?

MaartenBent: `libwebpdemux` has an internal dependency on `libwebp` so it'll probably returns both libraries:
```
Requires.private: libwebp >= 0.2.0
```

But to be sure, you could add both of them to `pkg_check_modules`.

vadz: This is a private dependency, so it's not taken into account when returning the libraries needed to be linked with, this can be seen by running pkg-config

```console
$ pkg-config --libs libwebpdemux
-lwebpdemux
```

After looking at this again, I'm pretty sure we need both `libwebpdemux` and `libwebp` here. @cjmayo Can you confirm that this works for you?

cjmayo: Yes. Let's do that. Updated.

PR #25796, opened 2025-09-14
Miscellaneous CMake improvements

MaartenBent:
* Update cotire to 1.8.1
* Remove duplicate find_package
* Don't recreate existing targets
* Mark more variables as advanced
* Add/update the version range
* Use GNUInstallDirs variables when installing


vadz:
Judging from the last commit, this is not ready to merge yet, right?

MaartenBent:
CI passed, I'll cleanup/split the last commit and after that it can be merged.

MaartenBent:
If CI passes, this can be merged.
I also added the `wxWidgets::Foo` targets that you mentioned a while back on wx-dev. Both for `FindPackage` and `add_subdirectory`.

PR #25805, opened 2025-09-16
Explain that clip paths don't work in NanoSVG, offer workaround

Blake-Madden:
Clip path not being supported was mentioned in a bug report, and I also ran into this issue. The workaround I explain here worked for me.

I also removed "These limitations will be relaxed in the future wxWidgets versions" statement. NanoSVG is no longer supported (according to its repo's README).

### DIFF
--- a/interface/wx/bmpbndl.h
+++ b/interface/wx/bmpbndl.h
@@ -248,13 +248,14 @@ public:
 
         Please note that the current implementation uses NanoSVG library
         (https://github.com/memononen/nanosvg) for parsing and rasterizing SVG
-        images which imposes the following limitations:
+        images, which does not support the following:
 
-        - Text elements are not supported at all (see note for workaround).
-        - SVG 1.1 filters are not supported.
-        - Embedded images are not supported (see note for workaround).
+        - Text elements
+        - SVG 1.1 filters
+        - Embedded images
+        - Clip paths
 
-        These limitations will be relaxed in the future wxWidgets versions.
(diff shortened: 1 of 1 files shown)


interface/wx/bmpbndl.h
```diff
@@ -278,12 +279,16 @@ class wxBitmapBundle
...
             Inkscape. (In Inkscape, select a text object and choose
-            "Object to Path" from the "Path" menu.)\n
+            "Object to Path" from the "Path" menu.)\n\n
             Converting embedded images to paths from an SVG editor will
             allow them to be rasterized. For example, selecting "Trace Bitmap"
             from the "Path" menu in Inkscape can perform this. This is only
             recommended for simple images, however, as more complex images
-            may not rasterize well.
+            may not rasterize well.\n\n
+            To work around clip paths, you can manually cut and remove portions
```
vadz: Isn't there something missing here?
```suggestion
            To work around lack of support for clip paths, you can manually cut and remove portions
```

vadz (commented):
Thanks, I don't really like tying the documentation to specific programs, but as we're already doing it here, I guess it's ok.

Blake-Madden:
> Thanks, I don't really like tying the documentation to specific programs, but as we're already doing it here, I guess it's ok.

I sort of thought that too, but it was difficult describing a workaround without offering tool-specific details. Plus, since Inkscape is open-source kindred I think it's OK ;)

PR #25806, opened 2025-09-17
Force standard button foreground color when highlighted on MSW

ardovm:
When a button is being hovered, we paint its background colour according to UxTheme, on Windows. 
IMHO we shall follow UxTheme for text colour as well.

Addresses issue #22721.


vadz:
I agree that this looks like a good idea, thanks. Will merge soon.

Issue #25797, opened 2025-09-15
High CPU usage on Windows when using a wxNotebook with long tab titles.


trypsynth:
#### Bug description:
When I run the provided code sample on Windows, I see the CPU usage of my app rising quite high, and my screen reader navigation lags. If I shorten the length of the title strings, the CPU usage goes way down.

#### Expected vs observed behaviour:
CPU usage should remain low, as we're doing nothing intensive.

#### Patch or snippet allowing to reproduce the problem:
```cpp
\#include <memory>
#include <wx/bookctrl.h>
#include <wx/wx.h>

class app : public wxApp {
public:
	bool OnInit() override {
		frame = new wxFrame(nullptr, wxID_ANY, "Test");
		auto* panel = new wxPanel(frame);
		notebook = new wxNotebook(panel, wxID_ANY);
		auto* sizer = new wxBoxSizer(wxVERTICAL);
		sizer->Add(notebook, 1, wxEXPAND | wxALL, 10);
		panel->SetSizer(sizer);
		auto* panel1 = create_tab_panel("This is a test.");
		auto* panel2 = create_tab_panel("This is a test.");
		notebook->AddPage(panel1, "This is a test of a very long title This is a test of a very long title This is a test of a very long title");
		notebook->AddPage(panel2, "This is a test of a very long title This is a test of a very long title This is a test of a very long title");
		frame->Show(true);
		return true;
	}

	wxPanel* create_tab_panel(const wxString& content) {
		wxPanel* panel = new wxPanel(notebook, wxID_ANY);
		auto* sizer = new wxBoxSizer(wxVERTICAL);
		long style = wxTE_MULTILINE | wxTE_READONLY | wxTE_RICH2 | wxTE_DONTWRAP;
		auto* text_ctrl = new wxTextCtrl(panel, wxID_ANY, "", wxDefaultPosition, wxDefaultSize, style);
		sizer->Add(text_ctrl, 1, wxEXPAND | wxALL, 5);
		panel->SetSizer(sizer);
		text_ctrl->SetValue(content);
		return panel;
	}

private:
	wxFrame* frame{nullptr};
	wxNotebook* notebook{nullptr};
};

wxIMPLEMENT_APP(app);
```

#### To Reproduce:
Run the compiled code sample, and then view CPU usage in task manager. Depending on how many cores you have, you'll see different usage values as wxWidgets only is allowed to use one core, but it's trying to use most of it.

### Platform and version information

[...]

PR #25426, opened 2025-05-19
RTL layout fixes for wxMSW and wxGTK3

AliKet:
Currently `wxMSW` is broken in RTL layout:
- Try scrolling `wxGrid` (in griddemo)
- Try switching to Regions screen in drawing sample to see the problem.

I tested this PR in RTL layout and it works correctly for me.

PR #25808, opened 2025-09-17
Don't enable `WS_EX_COMPOSITED` by default in wxMSW any longer

### LINKED ISSUE #25025: wxDataViewCtrl slow to (re)draw in wxWidgets 3.3.0 compared to 3.2.6
#### Bug description:

Some time between wxWidgets 3.2.6 and 3.3.0, something changed in wxDataViewCtrl that causes it to redraw much slower, introducing a 'lag' when scrolling the list with the mousewheel or keyboard. This happens in wxMSW only as far as I can tell (which I think uses the generic wxDataViewCtrl implementation).

I'm wondering if this could be related to #23585?

I've tried to record the difference using the snipping tool, rapidly scrolling the mousewheel up and down, hopefully it shows the issue clearly enough:

wxWidgets 3.2.6:

https://github.com/user-attachments/assets/43bb6370-3e5d-4f53-9ff4-095d68904931

and wxWidgets 3.3.0:

https://github.com/user-attachments/assets/835cab29-a9dc-4828-8bbd-e09d3efdc1b3

#### Stack trace:

N/A

N/A

#### To Reproduce:

Rapidly scroll the mousewheel within a wxDataViewCtrl (see the videos posted above)

- wxWidgets version you use: 3.3.0
- OS and its version: Windows 11

vadz:
This restores pre-3.3 behaviour of _not_ using composition (native double buffering) for all windows and removes many workarounds that are not necessary any more. It wasn't simple to find all such workarounds and I'm not sure if I found all of them, please let me know anything else that could/should be reverted now.

This should fix #25025, #25797 and maybe others I'm forgetting about.

Please test this if you can, TIA!

### DIFF
--- a/demos/forty/canvas.cpp
+++ b/demos/forty/canvas.cpp
@@ -34,15 +34,6 @@ FortyCanvas::FortyCanvas(wxWindow* parent, const wxPoint& pos, const wxSize& siz
              m_playerDialog(0),
              m_leftBtnDown(false)
 {
-#ifdef __WXMSW__
-    // Unfortunately the redraw logic here is incompatible with double
-    // buffering under MSW, so we have to disable it. The proper solution would
-    // be to rewrite this logic, as this is also required to make it work under
-    // GTK/Wayland and macOS, where the game currently doesn't update its
-    // display at all.
-    MSWDisableComposited();
-#endif
-
     SetScrollbars(0, 0, 0, 0);
 
 #ifdef __WXGTK__
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -11,19 +11,6 @@ INCOMPATIBLE CHANGES SINCE 3.2.x:
 Changes in behaviour not resulting in compilation errors
 --------------------------------------------------------
 
-- wxMSW now uses double buffering by default, meaning that updating the
-  windows using wxClientDC doesn't work any longer, which is consistent with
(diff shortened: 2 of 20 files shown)


AliKet:
I'm just wondering if the `wxGrid` control will be affected by these changes. I don't have time to check it myself, as I'm preparing a pull request to add **RTL** support to the `wxQt` port, and also trying to complete #25426.

vadz:
At first glance the grid sample seems to work fine, is there anything more concrete you think should be tested in it? I've checked selection, editing, scrolling.

wsu-cb:
If I understand correctly, wxMSW behaves more like other ports when this PR is NOT applied.  Is that correct?  If so, with this PR applied, how should I configure my wxMSW application to go back to the more-portable behavior?

vadz:
> If I understand correctly, wxMSW behaves more like other ports when this PR is NOT applied. Is that correct? If so, with this PR applied, how should I configure my wxMSW application to go back to the more-portable behavior?

You'll just have to avoid the use of `wxClientDC`. Maybe we could add some `wxNO_CLIENT_DC` preprocessor symbol that could be predefined by the application to make its declaration inaccessible to them, in order to prevent them from accidentally relying on it working.

vadz:
I'm going to merge this soon if there are no other comments.

PR #25809, opened 2025-09-18
Pass toolbar button ID in `wxContextHelpEvent`

StevenPHirshman:
Modified cshelp.cpp file so that context help click on toolbar button returns button ID, NOT toolbar ID.


vadz:
Thanks, but I'd really prefer to have a virtual function instead of using dynamic cast like this.

And, of course, as you can see from the CI failures, the current version doesn't compile due to missing wx/toolbar.h include (but it won't be necessary if/when you change the code to use the virtual function).

StevenPHirshman:
Strange – I was able to build the library with no errors. Any suggestion WHERE the virtual function could go?

***@***.***

vadz:
> Strange – I was able to build the library with no errors.

You are probably using PCH. Please open this PR in your browser, you will see all the CI errors.

> Any suggestion WHERE the virtual function could go?

In `wxWindow`, of course, as I wrote in [this comment](https://github.com/wxWidgets/wxWidgets/issues/25802#issuecomment-3300599605).

P.S. If you reply by email, could you please trim the quoted text because it looks really poorly on the web page (again, please open it in your browser to see it).

StevenPHirshman:
Added a virtual function to window.h called SetHelpEventPos. It does nothing except for the wxtoolbar base case, where it parses the button clicked. No static or other casts are needed.

StevenPHirshman:
I uploaded my changes (using virtual functions):

https://github.com/StevenPHirshman/wxWidgets/commit/45aefbf1eaa8232abaed1338f1e78d3bba7e072e


***@***.***

include/wx/window.h
```diff
@@ -1448,6 +1448,9 @@ class WXDLLIMPEXP_CORE wxWindowBase : public wxEvtHandler
         return GetHelpTextAtPoint(wxDefaultPosition, wxHelpEvent::Origin_Unknown);
     }
 
+	//sph091825
```
vadz: Please don't do this, we use version control for this. Just imagine what this file would look like if each of hundreds of contributors to it left such lines in it.
```suggestion
	//sph091825
```

include/wx/window.h
```diff
@@ -1448,6 +1448,9 @@ class WXDLLIMPEXP_CORE wxWindowBase : public wxEvtHandler
...
+	virtual void SetHelpEventPos(wxHelpEvent& WXUNUSED(event)) { return; }
```
vadz: This is trivial, but why have this useless `return`?
```suggestion
	virtual void SetHelpEventPos(wxHelpEvent& WXUNUSED(event)) { }
```

vadz (changes requested):
Thanks, I don't really understand what is the advantage of doing it like this rather than using `GetHelpIdAtPoint()` as I thought to do it, but why not.

The real problem for me is that:

1. This function name is misleading, it sets the ID and (not just the) position.
2. IMO it shouldn't set the position at all, it doesn't make much sense to have it in client coords for the toolbars and screen coords for everything else.

So could you please rename the function to something like `AdjustHelpEvent()` and remove `SetPosition()` call from it unless there is some good reason to do it that I'm missing?

StevenPHirshman:
I’ve changed the method name to SetHelpIdAtPoint. It change the event id to the button clicked, and not the toolbar id. No need to change the position.

***@***.***

vadz:
Please reset the draft status when the builds pass.

BTW, I forgot to mention this, but it's probably worth mentioning this behaviour in `wxHelpEvent` documentation in `interface/wx/event.h`. TIA!

StevenPHirshman:
Added the following description at line 3833 of event.h

    Modified 091825: Added a virtual function SetHelpIdAtPoint in windows.h to convert ToolBar Id in  context help mode to actual pressed button Id for derived wxToolBar class

Will update the PR shortly.

***@***.***

StevenPHirshman:
What do I need to do about some of the build error messages, like this one:

Mac builds / wxiOS
Failed in 13 minutes and 44 seconds

***@***.***

StevenPHirshman:
Will add a mention in the documentation.

***@***.***

StevenPHirshman:
Updated the wxHelpEvent document and removed any comments in the code (which I had put there so I could easily find the modified sections).

Let me know anything else is required.

***@***.***

vadz:
Just for the recorded, you (or your editor) have completely reformatted all the files you have touched. You need to configure it to prevent this from happening.

Please make the changes in your fork of the repository, confirm that they look good to you by comparing them with this one and then, and only then, push them here.

Thanks in advance!

StevenPHirshman:
> Just for the recorded, you (or your editor) have completely reformatted all the files you have touched. You need to configure it to prevent this from happening.
> 
> Please make the changes in your fork of the repository, confirm that they look good to you by comparing them with this one and then, and only then, push them here.
> 
> Thanks in advance!

I updated the files. Now there are no EOL/White Space issues, but ALL compilation tests are failing now. I don't understand why that is happening.

PR #25814, opened 2025-09-18
Fix SetCursor(wxNullCursor) behaviour in wxOSX

### LINKED ISSUE #25811: Suspected platform inconsistency with wxCursor between macOS and Windows that effects wxAUI gripper behavior
### Description

I noticed on macOS when the cursor moves over a wxAuiDockUIPart::typeGripper and then is moved over a wxAuiDockUIPart::typeCaption element, the mouse will continue to show wxCURSOR_SIZING instead of resetting back to the system default cursor. This misleads the user into thinking that they can resize the panel while in fact is they click they are actually over the caption and thus will instead tear off the panel. 

I tracked this down to `wxAuiManager::OnSetCursor(wxSetCursorEvent& event)`. Here the code will do a hit test and modify the cursor accordingly:

```    // determine cursor
    wxAuiDockUIPart* part = HitTest(event.GetX(), event.GetY());
    wxCursor cursor;

    if (part)
    {
         ...
    }

    event.SetCursor(cursor);
```

The code here handles the dock sizer, pane sizer, and gripper to modify the cursor to a resize icon. If non of the conditions are met though, it will just leave cursor unchanged and thus will subsequently call event.SetCursor with a default constructed wxCursor.

My hunch is that there is a difference in behavior between macOS and Windows here. It seems that on Windows, calling event.SetCursor(wxCursor()) will reset the cursor to the system default. While on macOS, doing this will leave the cursor unchanged. 

Changing the above code to initialize the cursor to the system default makes both platforms behave identical. Like this:

```
    // determine cursor
    wxAuiDockUIPart* part = HitTest(event.GetX(), event.GetY());
    wxCursor cursor(*wxSTANDARD_CURSOR);
```

Alternatively this can be patched by explicitly setting the cursor to wxSTANDARD_CURSOR when we are over a wxAuiDockUIPart::typeCaption. Like this:

```
        ...
        else if (part->type == wxAuiDockUIPart::typeGripper)
        {
            cursor = wxCursor(wxCURSOR_SIZING);
        }
        else if (part->type == wxAuiDockUIPart::typeCaption)
        {
            cursor = wxCursor(*wxSTANDARD_CURSOR);
        }
```

That said, I am not certain what the desired fix should be. If the issue is indeed a difference in behavior between macOS and Windows when setting the cursor to a default constructor cursor, we may want to make all platforms consistently reset the cursor to the system default when setting it to a default constrctor (wxNullCursor) cursor. 

* wxWidgets version: latest master branch as of 9/18/2025 but bug has been present for many years
* wxWidgets port: Inconsistency between wxMSW and wxOSX
* OS and version: All versions of macOS

vadz:
This should set the default cursor for the window instead of doing nothing.

Also simplify the code by overriding just the new WXUpdateCursor() member function instead of SetCursor() itself: this avoids the need for checking if the cursor has really changed.

Closes #25811.

### DIFF
--- a/include/wx/osx/window.h
+++ b/include/wx/osx/window.h
@@ -82,7 +82,6 @@ public:
     virtual void Update() override;
     virtual void ClearBackground() override;
 
-    virtual bool SetCursor( const wxCursor &cursor ) override;
     virtual bool SetFont( const wxFont &font ) override;
     virtual bool SetBackgroundColour( const wxColour &colour ) override;
     virtual bool SetForegroundColour( const wxColour &colour ) override;
@@ -129,6 +128,8 @@ public:
 
     // implementation from now on
     // --------------------------
+
+    virtual void WXUpdateCursor() override;
 
     void MacClientToRootWindow( int *x , int *y ) const;
 
--- a/src/osx/window_osx.cpp
+++ b/src/osx/window_osx.cpp
@@ -844,29 +844,14 @@ void wxWindowMac::DoGetClientSize( int *x, int *y ) const
     }
 }
 
-bool wxWindowMac::SetCursor(const wxCursor& cursor)
+void wxWindowMac::WXUpdateCursor()
 {
-    if (m_cursor.IsSameAs(cursor))
-        return false;
-
-    if (!cursor.IsOk())
-    {
-        if ( ! wxWindowBase::SetCursor( *wxSTANDARD_CURSOR ) )
-            return false ;
-    }
-    else
-    {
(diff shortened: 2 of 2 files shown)


vadz:
For the record, I'm still going to merge this even if it does _not_ fix #25811, finally, see the discussion in #25813. But the current behaviour of `SetCursor(wxCursor{})` still seems incorrect and this fixes it.

vadz:
Merged as the commit above.

PR #25817, opened 2025-09-19
Upgrade downloading WebView2 (Edge) from 1.0.1722.45 to 1.0.3485.44

### LINKED ISSUE #25815: WebView2 build issue in GH action (maybe need to update WebView2 version)
### Build System Used

Windows latest in GH action.
CMake static build.

### Description

On my local Windows 11 all works well, but now I get these errors when I statically build a project on Windows-latest in a GH action:

`wxWidgets_lib/libs/webview/packages/Microsoft.Web.WebView2.1.0.1722.45/WebView2.idl:2027: warning: included file 'AppWindow.cpp' is not found.`

Sleuthing around I see mention of a bug in this IDL file in some versions (?) of 1.0.1722.45 of WebView2 (although my local build looks OK and doesn't try to import 'AppWindow.cpp' anywhere).

The recommendation is to upgrade to 1.0.1901.177, which is the newest LTS release.

Anyway that this can be done in our build system? I see where this is done in "build\cmake\lib\webview\CMakeLists.txt", but I don't know how the SHA256 is generated.

- wxWidgets: trunk
- Windows latest on GH
- CMake

Blake-Madden:
Document how to update this in the build script
Closes #25815
Part of fix for ##25816

### DIFF
--- a/include/wx/msw/private/webview_edge.h
+++ b/include/wx/msw/private/webview_edge.h
@@ -20,7 +20,7 @@
 #include <WebView2.h>
 
 #ifndef __ICoreWebView2Settings3_INTERFACE_DEFINED__
-    #error "WebView2 SDK version 1.0.3485.44 or newer is required"
+    #error "WebView2 SDK version 1.0.1722.45 or newer is required"
 #endif
 
 #ifndef __VISUALC__
--- a/include/wx/msw/private/webview_edge.h
+++ b/include/wx/msw/private/webview_edge.h
@@ -20,7 +20,7 @@
 #include <WebView2.h>
 
 #ifndef __ICoreWebView2Settings3_INTERFACE_DEFINED__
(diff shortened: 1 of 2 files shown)


include/wx/msw/private/webview_edge.h
```diff
@@ -20,7 +20,7 @@
 #include <WebView2.h>
 
 #ifndef __ICoreWebView2Settings3_INTERFACE_DEFINED__
-    #error "WebView2 SDK version 1.0.1722.45 or newer is required"
+    #error "WebView2 SDK version 1.0.3485.44 or newer is required"
```
MaartenBent: I don't think we need to update this error message. The code still compiles when using WebView2 `1.0.1722.45`. Only when we starting using newer interfaces we need to update the minimum version.

Blake-Madden: Fair enough, I rolled that message back.

PR #25820, opened 2025-09-20
Generate help events with the IDs of wxToolBar tools

vadz:
This is more useful than generating the events with the ID of the toolbar itself, as help is typically associated with a particular tool and not the entire toolbar.

Add wxWindow::GetHelpIdAtPoint() virtual function to allow customizing this for the other controls as well.

Closes #25802.

### DIFF
--- a/include/wx/tbarbase.h
+++ b/include/wx/tbarbase.h
@@ -509,6 +509,12 @@ public:
     // Argument is wxID_ANY if mouse is exiting the toolbar.
     virtual void OnMouseEnter(int toolid);
 
+#if wxUSE_HELP
+    // override to return the ID of the tool at the given point
+    virtual int GetHelpIdAtPoint(const wxPoint& pt) override;
+#endif // wxUSE_HELP
+
+
     // more deprecated functions
     // -------------------------
 
--- a/include/wx/window.h
+++ b/include/wx/window.h
@@ -1448,6 +1448,12 @@ public:
         return GetHelpTextAtPoint(wxDefaultPosition, wxHelpEvent::Origin_Unknown);
     }
 
+        // may be overridden to return a more appropriate ID for the help event
+        // generated at the given point (in client window coordinates)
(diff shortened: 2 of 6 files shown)


StevenPHirshman:
I tested it on my PC and it worked. The checks completed successfully (1 pending, 42 successful).m Thanks for your help. Obviously you know the structure of the repository much better than I do.

StevenPHirshman:
Thank you for your help. Any thought about the related issue mentioned originally, namely, getting context help to open a menu item?

***@***.***

vadz:
> Thank you for your help. Any thought about the related issue mentioned originally, namely, getting context help to open a menu item?

I don't know of any way to do this, sorry.

StevenPHirshman:
> I don't know of any way to do this, sorry.

We had discussed a partial solution before you helped incorporate the virtual function GetHelpIdAtPoint into the context help file (which fixed the context help for toolbars). Namely, if I bypass the cshelp mechanism entirely when the context help button is clicked, and replace the frame call to 

wxContextHelp contextHelp(this);

with

m_bContext=true;
//add code here to change the cursor to the context "?" cursor; change back in menu/toolbar handler

Now the cursor is changed to "?" and when the user clicks a toolbar button OR a file menu item, the event handler for that item is invoked (NOT the wxEVT_HELP event since we haven't change to cs mode). In each event handler, one just needs to add a small function (IsContext, which returns m_bContext) to check the ciontext state and invoke the ShowHelp function.

The ONLY problem with this is that when the cursor leaves the frame area, it reverts to the normal (pointer) cursor. I can still click any menu header (like File or Edit) and it opens the menu list, and then click a menu item, and that works too. However, I can't figure out how to maintain the "?" cursor when the pointer leaves the main window. I tried setting the global cursor and not just the local cursor, but it still changes when the pointer leaves the main window. Do you know any way to KEEP the context cursor from changing (until the click is subsequently handled)?

PR #25822, opened 2025-09-21
wxQt RTL fixes

AliKet:
### Some screenshots of the result of this PR:

**Widgets demo:**
widgets_rtl

**Grid demo:**
griddemo_rtl

**AUI demo:**
auidemo_rtl

### DIFF
--- a/include/wx/qt/dc.h
+++ b/include/wx/qt/dc.h
@@ -38,6 +38,11 @@ public:
     virtual void SetPalette(const wxPalette& palette) override;
 #endif // wxUSE_PALETTE
 
+    virtual wxPoint DeviceToLogical(wxCoord x, wxCoord y) const override;
+    virtual wxPoint LogicalToDevice(wxCoord x, wxCoord y) const override;
+    virtual wxSize DeviceToLogicalRel(int x, int y) const override;
+    virtual wxSize LogicalToDeviceRel(int x, int y) const override;
+
     virtual void SetLogicalFunction(wxRasterOperationMode function) override;
 
     virtual wxCoord GetCharHeight() const override;
--- a/include/wx/qt/dc.h
+++ b/include/wx/qt/dc.h
@@ -117,6 +117,14 @@ public:
 
     virtual void* GetHandle() const override { return (void*) m_qtPainter; }
 
+    // LTR/RTL related functions
+    // -------------------------
+    //
+    // get or change the layout direction (LTR or RTL) for this dc,
+    // wxLayout_Default is returned if layout direction is not supported
+    virtual wxLayoutDirection GetLayoutDirection() const override;
+    virtual void SetLayoutDirection(wxLayoutDirection dir) override;
+
 protected:
     virtual QPixmap *GetQPixmap() { return m_qtPixmap; }
 
@@ -126,6 +134,8 @@ protected:
     wxRegion m_clippingRegion;
 
     bool m_isClipBoxValid = false;
+
+    wxLayoutDirection m_layoutDir = wxLayout_Default;
 
 private:
     enum wxQtRasterColourOp
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -3478,14 +3478,6 @@ wxRect wxAuiManager::CalculateHintRect(wxWindow* pane_window,
 
     m_frame->ClientToScreen(&rect.x, &rect.y);
 
-#if !defined(__WXMSW__) && !defined(__WXGTK3__) && !defined(__WXQT__)
-    if ( m_frame->GetLayoutDirection() == wxLayout_RightToLeft )
-    {
-        // Mirror rectangle in RTL mode
-        rect.x -= rect.GetWidth();
-    }
-#endif // !__WXMSW__ && !__WXGTK3__ && !__WXQT__
-
     return rect;
 }
 
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -3478,13 +3478,13 @@ wxRect wxAuiManager::CalculateHintRect(wxWindow* pane_window,
 
     m_frame->ClientToScreen(&rect.x, &rect.y);
 
-#if !defined(__WXMSW__) && !defined(__WXGTK3__)
+#if !defined(__WXMSW__) && !defined(__WXGTK3__) && !defined(__WXQT__)
     if ( m_frame->GetLayoutDirection() == wxLayout_RightToLeft )
     {
         // Mirror rectangle in RTL mode
         rect.x -= rect.GetWidth();
     }
-#endif // !__WXMSW__ && !__WXGTK3__
+#endif // !__WXMSW__ && !__WXGTK3__ && !__WXQT__
 
     return rect;
 }
--- a/src/qt/dc.cpp
+++ b/src/qt/dc.cpp
@@ -864,42 +864,30 @@ void wxQtDCImpl::DoDrawBitmap(const wxBitmap &bmp, wxCoord x, wxCoord y,
 
 void wxQtDCImpl::DoDrawText(const wxString& text, wxCoord x, wxCoord y)
 {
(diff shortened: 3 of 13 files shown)


RobertRoeb:
Amazing!

include/wx/qt/dc.h
```diff
@@ -117,6 +122,14 @@ class WXDLLIMPEXP_CORE wxQtDCImpl : public wxDCImpl
 
     virtual void* GetHandle() const override { return (void*) m_qtPainter; }
 
+    // LRT/RTL related functions
```
vadz: Typo:
```suggestion
    // LTR/RTL related functions
```

src/aui/framemanager.cpp
```diff
@@ -3478,11 +3478,13 @@ wxRect wxAuiManager::CalculateHintRect(wxWindow* pane_window,
 
     m_frame->ClientToScreen(&rect.x, &rect.y);
 
+#ifndef __WXQT__
```
vadz: I was always suspicious of this piece of code, if it's not needed for wxQt, it seems to confirm that it shouldn't be necessary for the other ports as well... Do you have any idea what could be wrong in/with them to require it?

src/qt/dc.cpp
```diff
@@ -462,10 +496,10 @@ void wxQtDCImpl::Clear()
     int width, height;
     DoGetSize(&width, &height);
 
-    m_qtPainter->eraseRect( DeviceToLogicalX(0),
-                            DeviceToLogicalY(0),
-                            DeviceToLogicalXRel(width),
-                            DeviceToLogicalYRel(height) );
+    const wxPoint pos = DeviceToLogical(0, 0);
```
vadz, line 497: Just to confirm: this doesn't really change anything, does it? I.e. `DeviceToLogicalX()` etc work too?

src/qt/overlay.cpp
```diff
@@ -100,10 +108,14 @@ void wxOverlayImpl::Init(wxDC* dc, int , int , int , int )
     m_overlay = new wxOverlayWindow(m_target ? m_target->GetHandle() : nullptr);
 }
 
-void wxOverlayImpl::BeginDrawing(wxDC* WXUNUSED(dc))
+void wxOverlayImpl::BeginDrawing(wxDC* dc)
 {
     wxCHECK_RET( IsOk(), "wxOverlay not initialized" );
 
+    // For correct result in RTL layout, the drawing is done in a non-mirrored
+    // DC and the mirroring is done in paintEvent() above.
```
vadz, line 116: Sorry, I feel dumb, but I don't really understand why do we need to do this. I.e. what is wrong with using RTL layout for the DC itself instead of mirroring it later?

src/qt/window.cpp
```diff
@@ -824,13 +833,69 @@ void wxWindowQt::ScrollWindow( int dx, int dy, const wxRect *rect )
...
+        }
+    }
+
+    Qt::LayoutDirection qtDir = Qt::LeftToRight;
+
+    switch ( dir )
+    {
+    default:
+        wxFALLTHROUGH;
+
```
vadz: I'd remove this to get a warning in the (unlikely) event of more elements being added to this enum.

tests/graphics/clippingbox.cpp
```diff
@@ -1033,9 +1033,9 @@ static void OneDevRegionRTL(wxDC& dc, const wxBitmap& bmp, bool useTransformMatr
         return;
     }
 
-#ifdef __WXGTK__
+#if defined(__WXGTK__) || defined(__WXMSW__)
```
vadz: Could we perhaps add a check for `__WXMSW__` below instead of disabling this test entirely for it?

vadz (commented):
Thanks, this looks good and could be merged AFAICS, I'm not sure why is this in the "draft" state yet?

Also, unfortunately I have no idea why did `SliderTestCase` start consistently failing with Qt 6.8 just recently, but this is  unrelated to this change.

src/aui/framemanager.cpp
AliKet: This code is also not needed when this [PR](#25426) is complete. see this [commit](https://github.com/wxWidgets/wxWidgets/pull/25426/commits/ee9c8e01197de157eb304006a4a8928dae210897)

src/qt/overlay.cpp
AliKet: If we don't do this (i.e.: forcing LTR on the DC) the rubberbanding is not drawn correctly, i.e.: the rubberband is missing near the right side of the window (image 01) and is drawn over the vertical scrollbar (image 02):

Image 01:
drawing_1

Image 02:
drawing_2

I haven't found the reason for this problem. I'll add a comment stating that this is to correct drawing the rubberband on a scrolled window.

tests/graphics/clippingbox.cpp
AliKet: The test is enabled for both `wxMSW` and `wxGTK3` in this [PR](#25426) (see [4f49395](https://github.com/wxWidgets/wxWidgets/pull/25426/commits/4f49395d979303efafea9661a420719a633a3b5d) and [403e9f2](https://github.com/wxWidgets/wxWidgets/pull/25426/commits/403e9f28d977d1c2c9c96498e53c45ee6f513abe))

I think I should merge the changes done here in that PR and close this one! what do you think ?

src/qt/dc.cpp
AliKet: `DeviceToLogical{X,Y,...}()` return different values ​​than their counterpart in the other ports (i.e.: `wxMSW` and `wxGTK3`), which makes `OneDevRegionRTL` test case in `wxQt` fail!

tests/graphics/clippingbox.cpp
vadz: Yes, probably, but whatever you decide please let me know how do you want me to process them, i.e. in which order to merge them if you don't combine them both. TIA!

src/qt/dc.cpp
vadz: Hmm, but which one is correct?

AliKet:
@vadz  Should I completely remove the code in [55f5433](https://github.com/wxWidgets/wxWidgets/pull/25822/commits/55f54331c94e11df7f75dd0cd57d674a05997041) because it seems unnecessary at least under `wxMSW`, `wxGTK3` and `wxQt`?

vadz:
Ideal would be to check if this code is needed under Mac and, if it is, and assuming the bug can't be fixed in wxOSX itself, replace the preprocessor condition with a check for `__WXOSX__`. TIA!

AliKet:
> Ideal would be to check if this code is needed under Mac and, if it is, and assuming the bug can't be fixed in wxOSX itself, replace the preprocessor condition with a check for `__WXOSX__`. TIA!

I don't have a Mac device to test this on it. so I'll leave the check as is then.

vadz:
FWIW we have a wxMac which is specifically there to be used for testing by people without access to a Mac. Please contact Stefan to get access to it.

In the meanwhile, let me know what do I need to do to test for it myself and I could do it here.

AliKet:
> 
> In the meanwhile, let me know what do I need to do to test for it myself and I could do it here.

1- Just hardcode the `wxAppBase::GetLayoutDirection()` return value (for testing purposes) in src/common/appcmn.cpp

```C++
wxLayoutDirection wxAppBase::GetLayoutDirection() const
{
    return wxLayout_RightToLeft;
}
```

2- Run the **auidemo** and see if the hint rectangle is correctly position in RTL layout.


aui

aui2

TIA!

vadz:
wxOSX doesn't seem to be affected by changing the return value of `wxAppBase::GetLayoutDirection()` at all, i.e. everything still shows up the same. Unsurprisingly, removing the code in question doesn't affect it either.

I think you should remove it and when/if anybody fixes RTL support in wxOSX, they could fix the bug that this worked around, if it's still there by then.

Thanks!

PR #25826, opened 2025-09-22
Correct name of user CMake presets file

LegalizeAdulthood:
The file is `CMakeUserPresets.json` not `CMakePresetsUser.json`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -456,4 +456,4 @@
 /3rdparty/webview2
 
 /build/cmake/output/
-CMakePresetsUser.json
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, this looks correct to me, I'll merge it soon.

LegalizeAdulthood:
I noticed it when I added a user presets in a checked out copy of wxWidgets because I wanted to tweak some things in the CMake configuration.  At first I was going to add it to the .gitignore then I noticed it was just spelled incorrectly there.

PR #25827, opened 2025-09-24
Implement native scrolling on iOS

RobertRoeb:
-  Supports scrolling with finger scrolling/panning, has single pixel precision 
  - Supports SetTargetWindow() used by e.g. wxDataViewCtrl
  - Draws before window is visible (I assume for off-screen drawing into buffer)
  - Added localDeviceOrigin like in GTK+, since the native scroll window moves windows up and this needs to be corrected in wxWidgets when e.g. reporting mouse event coordintes, update region and child window positions
  - Supports native scrolling physics beyond limits, then bounces back (like standard on iOS)
  - Supports wxDataViewCtrl with no changes in its code


vadz (commented):
Unfortunately I'm not competent enough to review this, @csomor could you please look at this?

FWIW I'm a bit horrified by the hacks in common code, I would hope we could somehow hide this in wxOSX implementation instead of requiring `#ifdef`s in any code using scrolling rate (IIUC).

RobertRoeb:
> I'm a bit horrified by the hacks in common code

These "hacks" are there to enforce that when scrolling with fingers, there are only single pixel scroll steps and this is done when setting the scrollbars. Nobody's finger uses steps of 20pixels. 
There is one other #ifdef referríng to scroll physics. The current code block scrolling outside of the requested range, but on iOS, the standard is to allow overscrolling and then to bounce back. The code does this perfectly.
Unfortunately, there is one case needed to support SetTargetWindow(). I don't think it is possible get this away, but I will look at this again.

vadz:
> These "hacks" are there to enforce that when scrolling with fingers, there are only single pixel scroll steps and this is done when setting the scrollbars. Nobody's finger uses steps of 20pixels.

If you say so — this doesn't seem at all obvious to me because I definitely never want to scroll in steps of 1px, whichever scrolling method is used. But I mostly would hope that we could add some `GetEffectiveScrollRate()` or whatever function and use it without having to write `#ifdef`s.

> There is one other #ifdef referríng to scroll physics. The current code block scrolling outside of the requested range, but on iOS, the standard is to allow overscrolling and then to bounce back.

Again, I would hope this could be something `wxScrolled` could take into account internally...

RobertRoeb:
> Again, I would hope this could be something `wxScrolled` could take into account internally...

It would be easy to add a wxScrolled::AllowScrollingBeyondRange( bool allow= true );

vadz:
> > Again, I would hope this could be something `wxScrolled` could take into account internally...
> 
> It would be easy to add a wxScrolled::AllowScrollingBeyondRange( bool allow= true );

Why couldn't it be built-in? I.e. allow under iOS (and Android...) but not elsewhere?

RobertRoeb:
> this doesn't seem at all obvious to me 

I tried a different approach first: keep the scroll steps at e.g. 10px and allow finger scrolling
with 1x. The problem is that the wxScrollWinEvent will report the new position in steps to the
window and that determines how much PrepareDC() changes drawing. So your finger would
slowly move pixel by pixel over the screen, but drawing would jump by 10px. Even worse, the
native scrolling widget will first scroll slowly, and then the drawing code will make it jump back.

> because I definitely never want to scroll in steps of 1px

I believe that on your phone, you always do and always want to scroll in steps of 1px. I mean,
1px precision. You surely expect the content on screen to move with your finger, don't you?

BTW, the same applies to finger scrolling on touch screens of laptops. I am not quite sure
how we handle that, actually. If the mouse wheel is used, I understand it is expected to 
scroll by a step (of e.g. 10px) for each mouse wheel movement. But does wxWidgets react
to finger scrolling at all? So basically the wxPanGestureEvent. In my opinion, wxScrolled
should react to wxPanGestureEvent by default - but it currently does not.

RobertRoeb:
> > > Again, I would hope this could be something `wxScrolled` could take into account internally...
> > 
> > 
> > It would be easy to add a wxScrolled::AllowScrollingBeyondRange( bool allow= true );
> 
> Why couldn't it be built-in? I.e. allow under iOS (and Android...) but not elsewhere?

Now I am confused. This scroll physics is default on iOS, but not Android. that is why I put
it into #ifdef __OSXIPHONE__ so it _is_ built-in in my PR. However, that is what you asked
to change initially. Let me know which way is preferred.

vadz:
You wrote:

> There is one other #ifdef referríng to scroll physics. 

I'd like to avoid having ifdefs by just handling this in wxiOS itself.

RobertRoeb:
> But does wxWidgets react to finger scrolling at all? So basically the wxPanGestureEvent

New idea: on iOS, UIScrollView captures the gesture events and then translates that into a 
new (scroll) position (actually in both x and y directions at once). So maybe we need a 
wxEVT_SCROLLWIN_PAN event with GetPixelX() and GetPixelY() (and GetPosX() and 
GetPosY() which would be the pixels positions divided by the scroll steps). On iOS, the
code would generate that directly, on other platforms, we might choose to catch the 
wxPanGestureEvent and translate that into wxEVT_SCROLLWIN_PAN in a second step.
In wxScrollHelperBase::HandleOnScroll(wxScrollWinEvent& event) we could process the
wxEVT_SCROLLWIN_PAN and translate that into a two-directional scroll event with single
pixel precision and we adapt OnPrepareDC() as well as CalcScrolledPosition() and 
CalcUnscrolledPosition() to add the small single-pixel offset. I have no computer with 
a touch screen - maybe I can simulate wxPanGestureEvent somehow.

RobertRoeb:
This seems to work. I added a new wxScrollWinPanEvent that is handled by wxScrolled. Based on that, it should be simple to add wxPanGestureEvent support to wxScrolled and there are almost no #ifdefs anymore. I will document if approved.

csomor:
> Unfortunately I'm not competent enough to review this, @csomor could you please look at this?

I cannot comment on the cross-platform parts, but I'm happy to support the approach. I'll look through things and try to propose extensions where the osx common code needs adjustments because we don't have native offsets elsewhere in osx. E.g. in the scroll sample the drawing part of MyCanvas is not in synch with the scrolling, while the buttons and panels are.

RobertRoeb:
> E.g. in the scroll sample the drawing part of MyCanvas is not in synch with the scrolling, while the buttons and panels are

Are you seeing that now on macOS or iOS or both?

csomor:
> > E.g. in the scroll sample the drawing part of MyCanvas is not in synch with the scrolling, while the buttons and panels are
> 
> Are you seeing that now on macOS or iOS or both?

iOS only, vertical scrolling keeps the text in sync, while horizontal does not, I don't know whether it's related to have only the VERTICAL event being triggered, but your code setting the device origin looks fine to me, so I don't yet know where this gets lost ...

moving things left makes the rendered parts stay in place

horz

but pushing things up keeps the y offset correctly

vert

csomor:
> > > E.g. in the scroll sample the drawing part of MyCanvas is not in synch with the scrolling, while the buttons and panels are
> > 
> > 
> > Are you seeing that now on macOS or iOS or both?
> 
> iOS only, vertical scrolling keeps the text in sync, while horizontal does not, I don't know whether it's related to have only the VERTICAL event being triggered, but your code setting the device origin looks fine to me, so I don't yet know where this gets lost ...

<del>ok, found it, it's the fact that you currently only trigger y thumbtracks, its handler then updates the scrollposition members, I'll comment this, if I do it for x and y everything stays in sync<del>

Ok, I see you've already added the PAN event, that's perfect

csomor:
> > > > E.g. in the scroll sample the drawing part of MyCanvas is not in synch with the scrolling, while the buttons and panels are
> > > 
> > > 
> > > Are you seeing that now on macOS or iOS or both?
> > 
> > 
> > iOS only, vertical scrolling keeps the text in sync, while horizontal does not, I don't know whether it's related to have only the VERTICAL event being triggered, but your code setting the device origin looks fine to me, so I don't yet know where this gets lost ...
> 
> ok, found it, it's the fact that you currently only trigger y thumbtracks, its handler then updates the scrollposition members, I'll comment this, if I do it for x and y everything stays in sync

btw I'm using the scroll sample, commented the parts we don't support on iOS yet out, and then in the `MyApp::OnInit` modify things a little to switch the different cases:

```
    MyFrame* myFrame = new MyFrame();

    // new MySimpleFrame(myFrame);
    new MyCanvasFrame(myFrame);
    // new MySizerFrame(myFrame);
    // new MySubFrame(myFrame);
    // new MyAutoFrame(myFrame);
```

and in the MyFrame Constructor I commented out the 3 lines between `panel->SetSizer(topsizer)` and `Show()` so that it stays fullscreen

RobertRoeb:
> Ok, I see you've already added the PAN event, that's perfect

Yes, scrolling works in both directions now, but initially, on iOS, the text is drawn in the wrong place and then in jumps (I copied MyCanvas into my test app). I cannot see where that happens.

csomor:
> > Ok, I see you've already added the PAN event, that's perfect
> 
> Yes, scrolling works in both directions now, but initially, on iOS, the text is drawn in the wrong place and then in jumps (I copied MyCanvas into my test app). I cannot see where that happens.

Didn't catch that, `wxScrollHelperBase::DoPrepareReadOnlyDC` has `m_yScrollPositionPixelOffset` at 0 on the first paint, but 5 the second time it is called, but the offset seems bigger.

When I add a rectangle at 0,0 with 10,10 size, then this is the display after a few scrolls up and down:
top
So I think we have some permanent y offset, `contentOffset` is at 1/-18 for some reason

csomor:
> > > Ok, I see you've already added the PAN event, that's perfect
> > 
> > 
> > Yes, scrolling works in both directions now, but initially, on iOS, the text is drawn in the wrong place and then in jumps (I copied MyCanvas into my test app). I cannot see where that happens.
> 
> Didn't catch that, `wxScrollHelperBase::DoPrepareReadOnlyDC` has `m_yScrollPositionPixelOffset` at 0 on the first paint, but 5 the second time it is called, but the offset seems bigger.
> 
> When I add a rectangle at 0,0 with 10,10 size, then this is the display after a few scrolls up and down: top So I think we have some permanent y offset, `contentOffset` is at 1/-18 for some reason

ok, we must turn an automatic behaviour off by issuing 

```
    wxUIView* v = [[wxUIView alloc] initWithFrame:r];
    v.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentNever;
```

then everything snaps to 0,0 again

RobertRoeb:
> v.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentNever;

Thanks, I would not have found that in a week

RobertRoeb:
> So I think we have some permanent y offset,

The toolbar should be lower, the little square is the right place, I think. That is for another PR

csomor:
I see the windows are using erase background style, but on iPhone we'd need to clear the background, but I have to make sure we don't get other artifacts, just for your tests, eg in the MyFrame scrollable areas, you can add dc.Clear() at the beginning so that the scrolled texts are not overwriting each other.

csomor:
`MySubFrame` gives another challenge, I think `::ScrollWindow` has to be implemented somehow, perhaps directly manipulating contentOffset, as in this sample the scrolling is directed also to the row and column header for their respective dimensions.

csomor:
Ok, I've tried to find a minimal change `wxWindowMac::ScrollWindow` needs at least a `Refresh` before the return, and the `wxScrollHelperBase::HandleOnPanScroll` needs to call `m_targetWindow->ScrollWindow` as well, so that the call can be intercepted. Sorry for misleading ...

RobertRoeb:
> `wxScrollHelperBase::HandleOnPanScroll` needs to call `m_targetWindow->ScrollWindow` as well

Indeed, and I did just that. However, since UIScrollView already scrolled, ScrollWindow() will not actually
do anything in this case, but it is called to allow being overridden like in the sample - which works now.
Except for the background issue, which is separate. I need to figure out how I prevented double scrolling
in wxGTK back then.

RobertRoeb:
As mentioned above, I just added support for pan gestures to wxScrolled<>. I don't actually have a touch screen that can do this, so I simulated this using mouse dragging. To clarify, this is for the desktop versions and should now work on all platforms which emit the wxPanGestureEvent.

src/osx/iphone/window.mm
```diff
@@ -550,6 +608,81 @@ - (void) WX_valueChangedAction:(id)sender event:(UIEvent*)event
...
+
+    CGPoint position = [view contentOffset];
+
+    position.x -= dx;
+    position.y -= dy;
+
+    m_blockScrollEvents = true;
+    [view setContentOffset: position];    
+    m_blockScrollEvents = false;
+
```
csomor: Sorry, I've been mistaken wich my comment on the above implementation, so I deleted it, but I've linked this part below to the wrong source part now :-(

at the moment please add this to get proper layered drawing in `window_osx.cpp`, in `wxWindowMac::MacDoRedraw` at the end of the `if ( bgStyle == wxBG_STYLE_ERASE )` clause in line 1962
```
                    if (!UseBgCol())
                        dc.Clear();
```

csomor:
The fact that we always have layered windows on iOS lets me see another background issue, I don't want to add too much to this PR, but this one seems important to test with scroll sample on macOS and iOS and compare, so if you agree I'd propose to add one additional change, inheriting background colors only when they are set, otherwise Clear() does always work as intended . 

in `dcclient.h` the `wxWindowDCImpl` gets an overload

```
virtual void InheritAttributes(wxWindow *win) override;
```
with an implementation in `dcclient.cpp`
```
void wxWindowDCImpl::InheritAttributes(wxWindow *win)
{
    wxCHECK_RET( win, "window can't be null" );

    SetFont(win->GetFont());

    SetTextForeground(win->GetForegroundColour());
    if ( win->UseBackgroundColour() ) {
        SetTextBackground(win->GetBackgroundColour());
        SetBackground(win->GetBackgroundColour());
    }
    SetLayoutDirection(win->GetLayoutDirection());
}
```

and use this in `wxWindowDCImpl::wxWindowDCImpl( wxDC *owner, wxWindow *window )`

after the     `DoSetClippingRegion( 0 , 0 , m_width , m_height ) ;` remove the two setters and replace with

```
    InheritAttributes(window);
```

src/osx/iphone/window.mm
```diff
@@ -550,6 +608,81 @@ - (void) WX_valueChangedAction:(id)sender event:(UIEvent*)event
...
+    if (orient == wxVERTICAL)
+    {
+        size = CGSizeMake( size.width, range*thumb );
+        position = CGPointMake( position.x, pos*thumb );
+    } else
+    if (orient == wxHORIZONTAL)
+    {
+        size = CGSizeMake( range*thumb, size.height );
+        position = CGPointMake( pos*thumb, position.y );
+    }
```
csomor, line 718: In my scroll sample, the `MyFrame` has a scroll area that is way too big. So I think - since we must adjust this to points - that we could take the `thumb` as the part representing the view's `bounds` and expand this to `range`
or `pos` 
```
    if ( thumb == 0 )
        return;

    CGPoint position = [view contentOffset];
    CGSize size = [view contentSize];
    CGSize viewSize = [view bounds].size;

    if (orient == wxVERTICAL)
    {
        CGFloat unit = viewSize.height/thumb;
        size = CGSizeMake( size.width, range*unit );
        position = CGPointMake( position.x, pos*unit );
    }
    else if (orient == wxHORIZONTAL)
    {
        CGFloat unit = viewSize.width/thumb;
        size = CGSizeMake( range*unit, size.height );
        position = CGPointMake( pos*unit, position.y );
    }
```

would this also work in your test samples ?

csomor (approved):
@RobertRoeb thanks a lot for your efforts, I'm fine with the PR as it is now for the entire iOS part, I've tested things with scroll sample on iOS and on macOS

src/osx/iphone/window.mm
```diff
@@ -194,8 +244,15 @@ + (void)initialize
...
 
+typedef void (*wxOSX_touchEventHandlerPtr)(UIView* self, SEL _cmd, NSSet* touches, UIEvent *event );
+
 void wxOSX_touchEvent(UIView* self, SEL _cmd, NSSet* touches, UIEvent *event )
 {
+    // Call the superclass handler first as that is e.g. handling mouse events for UISegementedControls
+    wxOSX_touchEventHandlerPtr superimpl = 
+        (wxOSX_touchEventHandlerPtr) [[self superclass] instanceMethodForSelector:_cmd];
+    superimpl(self, _cmd, touches, event );
+
```
csomor: @RobertRoeb can you please give me a sample on where this is needed ? Thanks

csomor:
@vadz @RobertRoeb are there open issues on this ? I'm holding back my own iOS PR because parts of it were added to this PR, in order to make testing easier, but now I'm stuck because I don't want to break merging and I see Robert has opened another PR, which also touches on parts of this PR I'd like add ...

include/wx/osx/iphone/private.h
```diff
@@ -38,7 +38,7 @@ wxBitmapBundle WXDLLIMPEXP_CORE wxOSXCreateSystemBitmapBundle(const wxString& id
 class WXDLLIMPEXP_CORE wxWidgetIPhoneImpl : public wxWidgetImpl
 {
 public :
-    wxWidgetIPhoneImpl( wxWindowMac* peer , WXWidget w, int flags = 0 ) ;
+    wxWidgetIPhoneImpl( wxWindowMac* peer , WXWidget w, int flags = 0, void *controller = NULL  ) ;
```
vadz:
```suggestion
    wxWidgetIPhoneImpl( wxWindowMac* peer , WXWidget w, int flags = 0, void *controller = nullptr ) ;
```

include/wx/osx/iphone/private.h
```diff
@@ -124,8 +128,21 @@ public :
...
+
+    void*               GetController() { return m_controller; }
+    bool                GetBlockScrollEvents() const { return m_blockScrollEvents; }
+    void                SetBlockScrollWindow( bool block ) { m_blockScrollWindow = block; } 
+    bool                GetBlockScrollWindow() const { return m_blockScrollWindow; }
+
 protected:
-    WXWidget m_osxView;
+    WXWidget          m_osxView;
+    void             *m_controller;
```
vadz: Minor, but this is inconsistent with everything else here:
```suggestion
    void*             m_controller;
```

include/wx/osx/iphone/private.h
```diff
@@ -104,7 +104,11 @@ public :
     wxInt32             GetMaximum() const;
     int                 GetIncrement() const { return 1; }
     void                PulseGauge();
+    void                SetScrollbar( int orient, int pos, int thumb, int range, bool refresh );
+    int                 GetScrollPos(int orient) const;
+    void                SetScrollPos(int orient, int pos);
     void                SetScrollThumb( wxInt32 value, wxInt32 thumbSize );
+    void     	        ScrollWindow (int dx, int dy, const wxRect *rect=NULL);
```
vadz:
```suggestion
    void     	        ScrollWindow (int dx, int dy, const wxRect* rect = nullptr);
```

include/wx/scrolwin.h
```diff
@@ -349,6 +351,13 @@ class WXDLLIMPEXP_CORE wxScrollHelperBase : public wxAnyScrollHelperBase
     int                   m_yScrollPixelsPerLine;
     int                   m_xScrollPosition;
     int                   m_yScrollPosition;
+    // ScrollPosition is using steps or lines, but if we need single pixel
+    // precisio, then ScrollPositionPixelOffset gets added to
```
vadz:
```suggestion
    // precision, then ScrollPositionPixelOffset gets added to 
```

include/wx/scrolwin.h
```diff
@@ -349,6 +351,13 @@ class WXDLLIMPEXP_CORE wxScrollHelperBase : public wxAnyScrollHelperBase
     int                   m_yScrollPixelsPerLine;
     int                   m_xScrollPosition;
     int                   m_yScrollPosition;
+    // ScrollPosition is using steps or lines, but if we need single pixel
+    // precisio, then ScrollPositionPixelOffset gets added to 
+    // ScrollPosition * ScrollPixelsPerLine
+    // GetPos() { return ScrollPosition * ScrollPixelsPerLine; }
+    // GetPosPrecise() { return (ScrollPosition * ScrollPixelsPerLine) + ScrollPositionPixelOffset; }
```
vadz: I don't really understand this comment :-( What is `GetPosPrecise()` supposed to mean?

interface/wx/event.h
```diff
@@ -1846,6 +1846,67 @@ class wxJoystickEvent : public wxEvent
...
+
+    /**
+        Sets the horizontal scroll position in pixels.
+    */
+    void SetX( int pos );
+
+    /**
+        Sets the vertical scroll position in pixels.
+    */
+    void SetY( int pos );
```
vadz: Those are only used by wx so probably should _not_ be part of the public API.

interface/wx/scrolwin.h
```diff
@@ -134,9 +134,18 @@ enum wxScrollbarVisibility
...
+    @beginEventEmissionTable{wxScrollWinPanEvent}
+    @event{EVT_SCROLLWIN_PAN(func)}
+        Process @c wxEVT_SCROLLWIN_PAN pan events after finger scrolling on touch screens.
+    @endEventTable
+
     @note
         Don't confuse wxScrollWinEvents generated by this class with
         wxScrollEvent objects generated by wxScrollBar and wxSlider.
+    @note
+        Since version 3.3.1 wxScrolled<> also intercepts wxPanGestureEvents
```
vadz:
```suggestion
        Since version 3.3.2 wxScrolled<> also intercepts wxPanGestureEvents
```

src/generic/scrlwing.cpp
```diff
@@ -472,8 +505,77 @@ void wxScrollHelperBase::SetTargetWindow(wxWindow *target)
...
+void wxScrollHelperBase::HandleOnPanGesture(wxPanGestureEvent& event)
+{
+    int xOldPrecisePosition = m_xScrollPixelsPerLine * m_xScrollPosition + m_xScrollPositionPixelOffset;
+    int yOldPrecisePosition = m_yScrollPixelsPerLine * m_yScrollPosition + m_yScrollPositionPixelOffset;
+    int xNewPrecisePosition = xOldPrecisePosition + event.GetDelta().x;
+    int yNewPrecisePosition = yOldPrecisePosition + event.GetDelta().y;
+
+    wxScrollWinPanEvent winPanEvent( wxEVT_SCROLLWIN_PAN, xNewPrecisePosition, yNewPrecisePosition );
+    event.SetEventObject( m_win );
+    // m_win->GetEventHandler()->ProcessEvent( winPanEvent );  cannot do this because it is protected
```
vadz: If we need to call `ProcessEvent()`, `ProcessWindowEvent()` should be used. If we don't, then this line should be just removed.

src/osx/carbon/dcclient.cpp
```diff
@@ -47,8 +47,10 @@ wxWindowDCImpl::wxWindowDCImpl( wxDC *owner, wxWindow *window )
     m_ok = true ;
 
     m_window->GetSize( &m_width , &m_height);
+#ifndef __WXOSX_IPHONE__
```
vadz: It would be nice to have a comment explaining why is this needed.

src/osx/carbon/dcclient.cpp
```diff
@@ -96,6 +104,20 @@ wxWindowDCImpl::~wxWindowDCImpl()
...
 }
 
+void wxWindowDCImpl::InheritAttributes(wxWindow *win)
+{
+    wxCHECK_RET( win, "window can't be null" );
+
+    SetFont(win->GetFont());
+
+    SetTextForeground(win->GetForegroundColour());
+    if ( win->UseBackgroundColour() ) {
```
vadz: Is this check the whole point for overriding this function in the first place? I don't really understand why we need to check for this but not for `UseForegroundColour()`...

Also:
```suggestion
    if ( win->UseBackgroundColour() )
    {
```

src/osx/iphone/window.mm
```diff
@@ -57,6 +57,56 @@ CGRect wxOSXGetFrameForControl( wxWindowMac* window , const wxPoint& pos , const
...
+    return self;
+}
+
+-(void)scrollViewDidScroll:(UIScrollView *)scrollView
+{
+    wxWidgetIPhoneImpl* viewimpl = (wxWidgetIPhoneImpl* ) wxWidgetImpl::FindFromWXWidget( scrollView );
+    if ( viewimpl )
+    {
+        wxWindowMac* wxpeer = (wxWindowMac*) viewimpl->GetWXPeer();
+        if (wxpeer) {
```
vadz:
```suggestion
        if (wxpeer)
        {
```

src/osx/iphone/window.mm
```diff
@@ -57,6 +57,56 @@ CGRect wxOSXGetFrameForControl( wxWindowMac* window , const wxPoint& pos , const
...
+            CGPoint position = [scrollView contentOffset];
+
+            viewimpl->SetDeviceLocalOrigin( wxPoint( -1*position.x, -1*position.y ) );
```
vadz: Why not just
```suggestion
            viewimpl->SetDeviceLocalOrigin( wxPoint( -position.x, -position.y ) );
```

src/osx/iphone/window.mm
```diff
@@ -57,6 +57,56 @@ CGRect wxOSXGetFrameForControl( wxWindowMac* window , const wxPoint& pos , const
...
+
+            if (viewimpl->GetBlockScrollEvents())
+                return; 
+
+            // wxString name = wxpeer->GetClassInfo()->GetClassName();
+            // wxLogMessage( "scrollViewDidScroll from %s  y %i", name, (int)(-1*position.y) );
+
```
vadz:
```suggestion
```

src/osx/iphone/window.mm
```diff
@@ -361,12 +414,18 @@ - (void) WX_valueChangedAction:(id)sender event:(UIEvent*)event
 
 void wxWidgetIPhoneImpl::SetNeedsDisplay( const wxRect* where )
 {
+    /*
```
vadz, line 425: Please remove code instead of commenting it out.

src/osx/iphone/window.mm
```diff
@@ -550,6 +610,111 @@ - (void) WX_valueChangedAction:(id)sender event:(UIEvent*)event
 {
 }
 
+void wxWidgetIPhoneImpl::SetScrollPos(int orient, int pos)
+{
+    if (m_blockScrollWindow) return;
```
vadz:
```suggestion
    if (m_blockScrollWindow)
        return;
```

src/osx/iphone/window.mm
```diff
@@ -550,6 +610,111 @@ - (void) WX_valueChangedAction:(id)sender event:(UIEvent*)event
...
+    wxUIView* view = (wxUIView*)m_osxView;  // wxUIView derives from UIScrollView
+
+    CGPoint position = [view contentOffset];
+
+    if (orient == wxVERTICAL)
+    {
+        if (m_yScrollPixelsPerLine == 0) return;
```
vadz:
```suggestion
        if (m_yScrollPixelsPerLine == 0)
            return;
```

src/osx/iphone/window.mm
```diff
@@ -550,6 +610,111 @@ - (void) WX_valueChangedAction:(id)sender event:(UIEvent*)event
...
+        position = CGPointMake( position.x, pos*m_yScrollPixelsPerLine );
+    }
+    else if (orient == wxHORIZONTAL)
+    {
+        if (m_xScrollPixelsPerLine == 0) return;
```
vadz:
```suggestion
        if (m_xScrollPixelsPerLine == 0)
            return;
```

src/osx/iphone/window.mm
```diff
@@ -550,6 +610,111 @@ - (void) WX_valueChangedAction:(id)sender event:(UIEvent*)event
...
+    }
+
+    return 0; 
+    
+}
+
+
+void wxWidgetIPhoneImpl::ScrollRect( const wxRect *rect, int dx, int dy )
+{
+    wxLogMessage( "wxWidgetIPhoneImpl::ScrollRect" );
```
vadz: We probably don't want this.

src/osx/iphone/window.mm
```diff
@@ -550,6 +610,111 @@ - (void) WX_valueChangedAction:(id)sender event:(UIEvent*)event
...
+}
+
+void wxWidgetIPhoneImpl::ScrollWindow( int dx, int dy, const wxRect *rect )
+{
+    if (m_blockScrollWindow) return;
```
vadz:
```suggestion
    if (m_blockScrollWindow)
        return;
```

src/osx/iphone/window.mm
```diff
@@ -742,7 +908,13 @@ - (void) WX_valueChangedAction:(id)sender event:(UIEvent*)event
 void wxWidgetIPhoneImpl::controlAction(void* sender, wxUint32 controlEvent, WX_UIEvent rawEvent)
 {
     if ( controlEvent == UIControlEventTouchUpInside )
+    {
         GetWXPeer()->OSXHandleClicked(0);
+    } else 
+    if ( controlEvent == UIControlEventValueChanged )
```
vadz:
```suggestion
    }
    else if ( controlEvent == UIControlEventValueChanged )
```

src/osx/iphone/window.mm
```diff
@@ -194,8 +244,15 @@ + (void)initialize
 
 @end // wxUIView
 
+typedef void (*wxOSX_touchEventHandlerPtr)(UIView* self, SEL _cmd, NSSet* touches, UIEvent *event );
+
 void wxOSX_touchEvent(UIView* self, SEL _cmd, NSSet* touches, UIEvent *event )
 {
+    // Call the superclass handler first as that is e.g. handling mouse events for UISegementedControls
```
vadz:
```suggestion
    // Call the superclass handler first as that is e.g. handling mouse events for UISegmentedControls
```

vadz (commented):
Some trivial fixes, please see the comment below for the more important questions.

vadz:
Except for the trivial issues of CI checks not passing, some typos and tons of formatting issues (I've left corrections for those I noticed but it surely would be nice if I didn't have to do it...), I'm still confused by the whole `wxScrollWinPanEvent` idea: why do we need it, when it seems that it can be made to work under iOS without any changes other than those to `wxScrolled` itself? I.e. how/why would you handle it specifically instead of letting `wxScrolled` deal with it? And, if the latter is possible, why doesn't it just do it without bothering with the new event? And if we do need the new event, it would be really great to test it on at least one other platform to ensure we don't add something we'll regret later.

If it's possible to apply just the wxiOS parts, perhaps we should do this for now to unblock further progress?

src/osx/carbon/dcclient.cpp
csomor: the problem on macOS is the fact that there may be no background set at all, but the GetBackgroundColor always tries to deliver some reasonable color, that way layered drawing which uses transparent views does not work anymore, while the foreground is never a problem. There it is reasonable to have a color, also on macOS.

csomor:
> Except for the trivial issues of CI checks not passing, some typos and tons of formatting issues (I've left corrections for those I noticed but it surely would be nice if I didn't have to do it...), I'm still confused by the whole `wxScrollWinPanEvent` idea: why do we need it, when it seems that it can be made to work under iOS without any changes other than those to `wxScrolled` itself? I.e. how/why would you handle it specifically instead of letting `wxScrolled` deal with it? And, if the latter is possible, why doesn't it just do it without bothering with the new event? And if we do need the new event, it would be really great to test it on at least one other platform to ensure we don't add something we'll regret later.
> 
> If it's possible to apply just the wxiOS parts, perhaps we should do this for now to unblock further progress?

Thanks for looking at this Vadim., I understand. 

There is IMHO a need for a bidirectional scroll event, I've mentioned this some years ago, then we've discussed that we at the moment I should be emitting two events, one for the x and one for the y axis. 

That could be the workaround for iOS at the moment as well, if you want to move this out to a separate PR.  But I really think we should have this on the longer term. And the default fall through IMHO should be to send two separate scroll events if the `wxScrollWinPanEvent` has not been handled. That way we don't force adoption.

As for using another event, I really don't know. We have the `wxPanGestureEvent` but that one only indicates deltas, so it is close to the mouse wheel events, while `wxScrollWinPanEvent` indicates absolute positions, so it is close to the scroll events....

RobertRoeb:
> when it seems that it can be made to work under iOS without any changes other than those to `wxScrolled` itself?

Without a bi-directional scroll event, the code obviously didn't work in both directions. I initially simplified it to only scroll vertically because that is what is usually used on smartphones.

RobertRoeb:
> And if we do need the new event, it would be really great to test it on at least one other platform to ensure
> we don't add something we'll regret later.

I did test it simulating wxPanGestureEvents by catching mouse drag events and emitting wxPanGestureEvents then and that works as it should with smooth pixel precision scrolling on wxMac. 
If something doesn't work, we can just block the code in wxScrolled<> that catches wxPanGestureEvents - but I think that wxScrolled<> absolutely should manage finger scrolling on touch screens by default. With the new wxScrollWinPanEvent I have separeted the new behaviour into a distinct class and function. That is as clean as it can get and there is almost not #ifdef left.

For the record, UIScrollVIew catches the systems pan gesture and translates it into native scrolling. So in wxScrolled<> we do
1) Catch wxPanGestureEvent
2) Translate to wxScrollWinPanEvent
3) Call ScrollWindow
on iOS all of that is done by the native scroll view and we just update the widget to reflect the change. I am very pleased that we have been able to wrap these different behaviours into a single API that only extended the old API. Without this, we can just remove iOS support. There are no scrollbars to click on.

There is still one funny place where the old (units based) and new (pixel based) logic shows the difference. In the scroll sub example with the row and col like for a spreadsheet, the row and col scroll in units (steps of 10 or 20px) when your finger scrolls in single pixels, because the code queries GetViewStart() which uses units, not pixels. We should add a GetViewStartInPixels() GetViewStartPrecise() that adjusts for the offset pixels. That is what GetPos() and GetPosPrecise() are supposed to mean.

vadz:
So do I understand correctly that the new event and new code in `wxScrolled<>` is needed to only implement iOS-like behaviour for the other platforms but not wxiOS itself, where this is already handled by iOS itself?

I'm also still confused by this "precise" scrolling stuff: how is it different from always using line size of 1? And if it isn't, why do we need extra support for it when using line size of 1 should just work?

vadz:
Independently of this, our macOS runner seems to be dead again :-(

csomor:
> Independently of this, our macOS runner seems to be dead again :-(

oh dear, I cannot connect either, I'll check ....

csomor:
ok, only teamviewer cannot connect, but Remote Desktop can, I'll probably reboot

RobertRoeb:
> So do I understand correctly that the new event and new code in `wxScrolled<>` is needed to
> only implement iOS-like behaviour for the other platforms but not wxiOS itself, where this is
> already handled by iOS itself?

The event is sent to wxScrolled<> so that it knows where it has scrolled, but it doesn't have to
scroll the window anymore (the code only call ScrollWindow() because this call was overloaded
in the scrollsub sample to scroll the rows and column according to the main window). On other
platforms, it will allow scrolling with pan gestures.

> I'm also still confused by this "precise" scrolling stuff: how is it different from always using
> line size of 1? And if it isn't, why do we need extra support for it when using line size of 1 
> should just work?

Current code uses units of 10 or 15 or 20 everywhere. This code will continue to work
unchanged, including wxHtmlWindow, wxDataViewCtrl etc. The new code allows for
scrolling between those steps (BTW, theoretically also with means other than finger panning.
We might want to enable smooth scrolling in an animated form: the mousewheel scrolls
down one unit, but it is shown on screen in small steps creating a smooth scroll effect).

Assume scroll units to be 20. so scrolling with scrollbar and mousewheel can scroll
to pixel position 20, 40 etc. ViewStartPos() will return 1, 2, etc. and ViewStartPosPrecise() 
would return 20, 40, etc. Finger panning can scroll you to pixel position 25. In this case
ViewStartPos() would return 1 (in units), ViewStartPosPrecise()  would return 25 (in pixels).
As soon as you use scrollbars or mousewheel for scrolling, the code will go back to 20, 40
and remove the "precise" offset.

I think it is very likely that you will continue to use 20px lines so that pressing on the
scrollbar arrows or using the mousewheel will scroll 20px per step.

RobertRoeb:
> ViewStartPos() would return 1 (in units), ViewStartPosPrecise()

GetViewStartPixel() seems the better name.

https://docs.wxwidgets.org/3.2/classwx_scrolled.html#a3ced29907b1e914f841f979f272277c8

vadz:
> > So do I understand correctly that the new event and new code in `wxScrolled<>` is needed to
> > only implement iOS-like behaviour for the other platforms but not wxiOS itself, where this is
> > already handled by iOS itself?
> 
> The event is sent to wxScrolled<> so that it knows where it has scrolled, but it doesn't have to scroll the window anymore (the code only call ScrollWindow() because this call was overloaded in the scrollsub sample to scroll the rows and column according to the main window). On other platforms, it will allow scrolling with pan gestures.

So this will still be handled entirely by `wxScrolled` in any case? Again, I'm trying to convince myself that we really need a new event, but so far I'm completely failing...

> > I'm also still confused by this "precise" scrolling stuff
>
> Current code uses units of 10 or 15 or 20 everywhere. This code will continue to work unchanged, including wxHtmlWindow, wxDataViewCtrl etc.

My question is why we can't ignore setting scroll pixels per unit in wxiOS and always keep them at 1. Wouldn't the code still continue to work unchanged?

OTOH I am afraid that allowing positions in between scroll units will break the existing code which doesn't expect it.

> I think it is very likely that you will continue to use 20px lines so that pressing on the scrollbar arrows or using the mousewheel will scroll 20px per step.

Sure, under the platforms not using panning to scroll.

csomor:
Perhaps another way to look at the events involved is the level of abstraction. 

`wxPanGestureEvent` is rather low level, like a mouse wheel event. Our wheel events only have one direction, i.e. if I'm using it to move a selected object I need two mouse wheel events - manhattan wise - first one then the other coordinate - ie two events, two redraws etc. While I'm having one action be the user, this gets translated into two separated events, or I'm putting them into one 2D `wxPanGestureEvent` which combines these two into one, one event, one redraw.

Now on the higher level we have scrolling, dragging scrollthumbs via mouse events in a scrollbar this mouse event gets translated into a scroll event. One direction only. If I'm having a pan gesture on a scrollable surface this would be translated into a 2-D scroll event which we don't yet have, we'd have to emulate it using two scroll events, one horizontal, one vertical, again two events, two redraws like above, or we get a 2D`wxScrollWinPanEvent` combining this.

csomor:
> ok, only teamviewer cannot connect, but Remote Desktop can, I'll probably reboot

@vadz after reboot I could only use TeamViewer, it was somehow blocking things on the firewall, I don't assume someone has changed remote management or screen sharing properties / firewallrules, so I'm rather at loss what happened...

RobertRoeb:
> we'd have to emulate it using two scroll events, one horizontal, one vertical, again two events, two
> redraws like above, or we get a 2D`wxScrollWinPanEvent` combining this.

It can be rewritten to use two SCROLLWIN_THUMB events when we add a field for the pixel precision. But given the different origin (from a pan gesture), the pixel precision vs. unit precision and the 2D nature, the event seems sufficiently different to use a new event. I don't know if on some platforms two ScrollWindow() calls would be less efficient than one, but I suspect, it could be. It would not matter on iOS, as the call does nothing in this scenario.

vadz:
Sorry, I'm completely overwhelmed with many different things right now, so maybe I'm just not spending enough time on this to understand it properly, but I still think that this PR is trying to do too many different and not necessarily related things at once and that it would be better to separate them. First thing is native scrolling under iOS which, AFAIU, can be implemented without implementing any new events at wx level. Second thing is "smooth" scrolling (i.e. scrolling by pixels instead of scroll units), which can or not be desirable on the other systems too (and might be achieved there by using thumb position to determine scroll position without rounding?) but should IMO be subject of its own PR and might need to be explicitly enabled, unless we're completely sure that this is not going to break the existing code which relies on always starting a new line at physical (0, 0) position. And third thing is 2D scrolling, i.e. scrolling at once in both directions.

If we could extract and apply just the first part, it would at the very least give me some breathing space instead of being under pressure to merge this urgently because this is blocking other changes. Can we do this please?

csomor:
> If we could extract and apply just the first part, it would at the very least give me some breathing space instead of being under pressure to merge this urgently because this is blocking other changes. Can we do this please?

I'm fine with this if Robert sees a subset, that would work correctly, need not be perfect for the first step. If this is not possible - how granular can a PR be merged ? at the file level, or can individual chunks be picked from a file ?

src/osx/carbon/dcclient.cpp
csomor: @vadz alternatively we might have a getter on `wxWindow` that indicates that this OS and this widget supports transparent backgrounds, and then - on such an instance - not generating a color if `GetBackgroundColour` is asked on a window that doesn't have it set, or just not calling the method, ie moving this code to base either to `GetBackgroundColour` or to `InheritAttributes`, and extend it with something like

```
if ( !win->SupportsTransparentBackground() && win->UseBackgroundColor() )
{
        SetTextBackground(win->GetBackgroundColour());
        SetBackground(win->GetBackgroundColour());
}
```

Or I'd have to modifiy the default attributes in the base class when running on OSX to always have `wxTransparentColour` by default as background, button face is not a good choice for these systems.

vadz: It would be better to handle this at the base class level, of course, there is even a comment saying this dating back from eb96212b6c (do not inherit background colour in InheritAttributes() as it totally breaks themed backgrounds, 2004-12-19) in the similar `wxWindow` function.

But OTOH I'm not sure that not calling `SetTextBackground()` at all is the right thing to do because `wxDCImpl::m_textBackgroundColour` is initialized to "white" by default, not "transparent".

RobertRoeb:
Here we go: wxScrollWinPanEvent removed, wxPanGestureEvent support removed. IOS scrolling now uses THUMBTRACK event to update wxScrolled<>, but I had to add a m_pixelOffest field to it to hold the single pixel precision data.

include/wx/scrolwin.h
vadz: This should now be changed to `GetViewStart()` and `GetViewStartPixels()`, but this is not actually the part I didn't understand. What is unclear, and the comment doesn't really help, is why the offset is only added to pixel positions but not the position in scroll units. Is it because it's always less than (one half of?) a unit?

include/wx/event.h
```diff
@@ -663,6 +663,7 @@ class WXDLLIMPEXP_FWD_CORE wxSetCursorEvent;
 class WXDLLIMPEXP_FWD_CORE wxScrollEvent;
 class WXDLLIMPEXP_FWD_CORE wxSpinEvent;
 class WXDLLIMPEXP_FWD_CORE wxScrollWinEvent;
+class WXDLLIMPEXP_FWD_CORE wxScrollWinPanEvent;
```
vadz, line 666: This should be removed too.

include/wx/event.h
```diff
@@ -805,6 +806,9 @@ wxDECLARE_EXPORTED_EVENT(WXDLLIMPEXP_CORE, wxEVT_SCROLLWIN_PAGEDOWN, wxScrollWin
 wxDECLARE_EXPORTED_EVENT(WXDLLIMPEXP_CORE, wxEVT_SCROLLWIN_THUMBTRACK, wxScrollWinEvent);
 wxDECLARE_EXPORTED_EVENT(WXDLLIMPEXP_CORE, wxEVT_SCROLLWIN_THUMBRELEASE, wxScrollWinEvent);
 
+    // Gesture based scroll events from wxWindow
```
vadz: And this.

include/wx/event.h
```diff
@@ -4278,6 +4286,7 @@ class WXDLLIMPEXP_CORE wxEventBlocker : public wxEvtHandler
 typedef void (wxEvtHandler::*wxCommandEventFunction)(wxCommandEvent&);
 typedef void (wxEvtHandler::*wxScrollEventFunction)(wxScrollEvent&);
 typedef void (wxEvtHandler::*wxScrollWinEventFunction)(wxScrollWinEvent&);
+typedef void (wxEvtHandler::*wxScrollWinPanEventFunction)(wxScrollWinPanEvent&);
```
vadz: And this and other changes below in this file.

include/wx/scrolwin.h
```diff
@@ -192,6 +192,16 @@ class WXDLLIMPEXP_CORE wxScrollHelperBase : public wxAnyScrollHelperBase
         return pt;
     }
 
+    // Get the view start in pixels
+    void GetViewStartPixels(int *x, int *y) const { DoGetViewStartPixels(x, y); }
+
+    wxPoint GetViewStartPixels() const
```
vadz, line 198: This should probably be documented?

include/wx/osx/core/private.h
```diff
@@ -354,6 +359,11 @@ public :
     virtual int         GetIncrement() const = 0;
     virtual void        PulseGauge() = 0;
     virtual void        SetScrollThumb( wxInt32 value, wxInt32 thumbSize ) = 0;
+    virtual void        SetScrollbar( int WXUNUSED(orient), int WXUNUSED(pos), int WXUNUSED(thumb), 
+                                      int WXUNUSED(range), bool WXUNUSED(refresh) ) {}
+    virtual int         GetScrollPos(int WXUNUSED(orient)) const { return 0; }
+    virtual void        SetScrollPos(int WXUNUSED(orient), int WXUNUSED(pos)) {}
+    virtual void 	    ScrollWindow (int WXUNUSED(dx), int WXUNUSED(dy), const wxRect *WXUNUSED(rect)=NULL) {}
```
vadz:
```suggestion
    virtual void 	    ScrollWindow (int WXUNUSED(dx), int WXUNUSED(dy), const wxRect *WXUNUSED(rect) = nullptr) {}
```

src/osx/iphone/window.mm
```diff
@@ -550,6 +620,113 @@ - (void) WX_valueChangedAction:(id)sender event:(UIEvent*)event
...
+    }
+
+    return 0; 
+    
+}
+
+
+void wxWidgetIPhoneImpl::ScrollRect( const wxRect *rect, int dx, int dy )
+{
+    wxLogMessage( "wxWidgetIPhoneImpl::ScrollRect" );
```
vadz:
```suggestion
```

vadz (commented):
Thanks, there is still the issue of background colour inheriting remaining, but I guess we'll deal with (or not...) later.

I'll address the remaining issues (i.e. remove the leftovers) later if you don't beat me to it.

csomor:
> Thanks, there is still the issue of background colour inheriting remaining, but I guess we'll deal with (or not...) later.

We can leave that part out from this, I'll do a separate PR trying to find a better solution

src/osx/carbon/dcclient.cpp
csomor: > It would be better to handle this at the base class level, of course, there is even a comment saying this dating back from [eb96212](https://github.com/wxWidgets/wxWidgets/commit/eb96212b6ca08230668db1bbab173fb7fb55e696) (do not inherit background colour in InheritAttributes() as it totally breaks themed backgrounds, 2004-12-19) in the similar `wxWindow` function.
> 
> But OTOH I'm not sure that not calling `SetTextBackground()` at all is the right thing to do because `wxDCImpl::m_textBackgroundColour` is initialized to "white" by default, not "transparent".

ok, another reason for trying to get this transparent colour thing working..

and yes, inheriting background from a parent window is definitely the wrong path, but inheriting those in a dc from the corresponding window seems reasonable, as long as transparency is preserved

vadz:
I'll apply and merge it now.

PR #25752, opened 2025-09-01
Proposed backports to 3.2

vadz:

PR #25828, opened 2025-09-25
Silence warning about non-virtual dtor on VC++ <=2017

lanurmi:
Explicitly defining a destructor in `wxMSWOwnerDrawnButtonBase` causes precisely the kind of compiler warnings (C4265) that the definition was meant to suppress on GCC.

C4265 is not enabled by default in VC++, but if you have it enabled, you'll get plenty of pointless warnings through anything that #includes `wx/msw/ownerdrawnbutton.h`. This is also the only spot in wx headers that triggers C4265.

The issue appears to be 3.2-specific.

### DIFF
--- a/include/wx/msw/ownerdrawnbutton.h
+++ b/include/wx/msw/ownerdrawnbutton.h
@@ -10,6 +10,15 @@
 #ifndef _WX_MSW_OWNERDRAWNBUTTON_H_
 #define _WX_MSW_OWNERDRAWNBUTTON_H_
 
+// With MSVC 2017 and earlier, explicitly defining the destructor below
+// _causes_ exactly the kind of compiler warnings (C4265) the definition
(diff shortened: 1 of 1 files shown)


vadz (commented):
If the warning doesn't happen in master which defaults this ctor, maybe we could just conditionally define it as defaulted instead? Or does the version of MSVC you use lack support for C++11 too?

lanurmi:
I use MSVC 2017, no problem with C++11 there. My proposed fix would solve this for all the MSVC versions 3.2.x claims to support, though. Also I can't immediately tell whether changing the dtor to defaulted would change the ABI, and whether changing the ABI would matter with this class, which is said to be an internal implementation detail.

vadz:
OK, I've added it to #25752, so closing this one, thanks.

PR #25833, opened 2025-09-26
po: Update Georgian translation

NorwayFun:


vadz:
Thanks, will push soon.

PR #25834, opened 2025-09-26
wxAUI: fix dock buttons not refreshed under some platforms

### LINKED ISSUE #25245: wxAUI dock buttons are not refreshed under platforms without working `wxClientDC` (e.g. Wayland and macOS)
`wxAuiManager::UpdateButtonOnScreen()` just calls `DrawPaneButton()` with the temporary state such as `wxAUI_BUTTON_STATE_HOVER` or `wxAUI_BUTTON_STATE_PRESSED` when using `wxClientDC` but otherwise it simply refreshes the window, which has no effect because this state is not stored anywhere.

We probably need to add `state` field to `wxAuiDockUIPart` and use it when calling `DrawPaneButton()` from `wxAuiManager::OnRender()`.

AliKet:
Under platforms without working wxClientDC (e.g. Wayland, macOS and wxQt) dock buttons are never refreshed to reflect current state of the button.

Closes #25245

### DIFF
--- a/include/wx/aui/framemanager.h
+++ b/include/wx/aui/framemanager.h
@@ -572,7 +572,7 @@ protected:
     void Repaint(wxDC* dc = nullptr);
     void ProcessMgrEvent(wxAuiManagerEvent& event);
     void UpdateButtonOnScreen(wxAuiDockUIPart* buttonUiPart,
-                              const wxMouseEvent& event);
+                              int state = wxAUI_BUTTON_STATE_NORMAL);
     void GetPanePositionsAndSizes(wxAuiDockInfo& dock,
                               wxArrayInt& positions,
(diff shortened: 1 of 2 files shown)


AliKet:
@vadz I did this in a hurry, and tested it only under wxQt and wxGTK3 (Wayland and X11)
feel free to modify it as you willl

vadz:
Thanks, but I don't really understand how could this work for you: you did add `state`, but you never used it anywhere...

Anyhow, I've actually done this now, also added its initialization and corrected a couple of other minor things and will push this soon.

AliKet:
Oops! My bad I I deleted it by mistake when applying the patch!! sorry and thanks for the corrections.

PR #25841, opened 2025-09-28
Correct client area origin and navigation back from secondary wxFrame on IOS

RobertRoeb:
- Add extra space for iPhone status bar, currently hard coded as a temporary fix until we can calculate it
  - Use it in GetClientAreaOrigin() and use that, respectively, in wxTopLevelWindow::Layout to place the client window correctly
  - Add default wxToolbar in all secondary frames as they have no close button with an [X], but most usually a < like sign as a way to return to the previous wxFrame = "screen"


art/prev_screen.xpm
```diff
@@ -0,0 +1,121 @@
+/* XPM */
```
vadz, line 1: We should be really using PNGs for any new icons IMNSHO.

src/common/toplvcmn.cpp
```diff
@@ -496,7 +496,13 @@ bool wxTopLevelWindowBase::Layout()
             int clientW, clientH;
             DoGetClientSize(&clientW, &clientH);
 
-            child->SetSize(0, 0, clientW, clientH);
+#ifndef __WXOSX_IPHONE__
```
vadz: Can we please avoid this, either by using `GetClientAreaOrigin()` under all platforms if this is indeed the right thing to do or by adding another virtual function, e.g. `GetUsableAreaOrigin()` if it is not. But using such preprocessor checks is just not a good idea.

src/osx/carbon/frame.cpp
```diff
@@ -60,6 +66,20 @@ bool wxFrame::Create(wxWindow *parent,
     if ( !wxTopLevelWindow::Create(parent, id, title, pos, size, style, name) )
         return false;
 
+#ifdef __WXOSX_IPHONE__
+    if (parent != NULL) {
```
vadz:
```suggestion
    if (parent != nullptr)
    {
```

src/osx/carbon/frame.cpp
```diff
@@ -60,6 +66,20 @@ bool wxFrame::Create(wxWindow *parent,
...
+        // We are on the next screen, provide a back button and title
+        wxToolBar *tb = CreateToolBar();
+        // tb->AddTool( wxID_CLOSE, wxEmptyString, wxArtProvider::GetBitmap(wxART_PREV_SCREEN) ); doesn't work for some reason
+        tb->AddTool( wxID_CLOSE, wxEmptyString, wxBitmap( prev_screen_xpm ) );
+        tb->AddStretchableSpace();  // doesn't yet stretch on iOS, so at least add two of them
+        tb->AddStretchableSpace();
+        tb->AddControl( new wxStaticText( tb, -1, title ) );
```
vadz:
```suggestion
        tb->AddControl( new wxStaticText( tb, wxID_ANY, title ) );
```

src/osx/carbon/frame.cpp
```diff
@@ -60,6 +66,20 @@ bool wxFrame::Create(wxWindow *parent,
     if ( !wxTopLevelWindow::Create(parent, id, title, pos, size, style, name) )
         return false;
 
+#ifdef __WXOSX_IPHONE__
+    if (parent != NULL) {
+        // We are on the next screen, provide a back button and title
+        wxToolBar *tb = CreateToolBar();
+        // tb->AddTool( wxID_CLOSE, wxEmptyString, wxArtProvider::GetBitmap(wxART_PREV_SCREEN) ); doesn't work for some reason
```
vadz: It really should, but if it doesn't, it would be worth opening an issue about it and removing this commented out line.

src/osx/carbon/frame.cpp
```diff
@@ -60,6 +66,20 @@ bool wxFrame::Create(wxWindow *parent,
...
+        tb->AddTool( wxID_CLOSE, wxEmptyString, wxBitmap( prev_screen_xpm ) );
+        tb->AddStretchableSpace();  // doesn't yet stretch on iOS, so at least add two of them
```
vadz: Again, instead of working around this here, why not make `AddStretchableSpace()` in wxiOS implementation add 2 spaces?

src/osx/carbon/frame.cpp
```diff
@@ -397,8 +417,22 @@ void wxFrame::PositionToolBar()
         else
         {
 #if !wxOSX_USE_NATIVE_TOOLBAR
+#ifdef __WXOSX_IPHONE__
```
vadz: This should also be a virtual function in `wxWidgetImpl`.

src/osx/nonownedwnd_osx.cpp
```diff
@@ -253,6 +253,20 @@ wxPoint wxNonOwnedWindow::GetClientAreaOrigin() const
 {
     int left, top, width, height;
     m_nowpeer->GetContentArea(left, top, width, height);
+
+#ifdef __WXOSX_IPHONE__
```
vadz: And by having it in a virtual function we'd at least avoid duplicating this.

src/osx/nonownedwnd_osx.cpp
```diff
@@ -470,6 +484,17 @@ void wxNonOwnedWindow::DoGetClientSize( int *width, int *height ) const
     // status bar, therefore we use the content view's area
 #ifdef __WXOSX_IPHONE__
     GetPeer()->GetContentArea(left, top, w, h);
+
+    // TODO, find a way to find the safe area not covered at the top
```
vadz: Sorry, triplicating.

vadz (changes requested):
Sorry but this is really more of a draft-quality than a ready-to-merge PR. Could you please polish it at least a little bit? TIA!

art/prev_screen.xpm
RobertRoeb: Not sure why- all icons that we currently ship are XPM

src/common/toplvcmn.cpp
RobertRoeb: Done

src/osx/carbon/frame.cpp
RobertRoeb: Done

src/osx/carbon/frame.cpp
RobertRoeb: Done

src/osx/carbon/frame.cpp
RobertRoeb: Tried, but failed

src/osx/nonownedwnd_osx.cpp
RobertRoeb: Done

src/osx/carbon/frame.cpp
RobertRoeb: Done

RobertRoeb:
I tested the changes on wxMac and wxGTK and didn't see anything broken. Indeed, I reverted all or nearly all changes in non-OSX or non-IOS code.
BackNavi IOS

RobertRoeb:
Here is my hack over a hack to allow stretchable spacer in iOS toolbar. The other solution would be to remove both hacks and add a flag to CreateTool() in all ports. I didn't do that because someone must have thought about that when he added the hack that is currently in the code and decided against the clean solution. 

BTW, I tried really hard to also use Realize() by removing the unstretchable spacers in the toolbar and inserting stretchable ones if they were marked with MarkStretchable(). That _almost_ worked but ruined the layout of the toolbar for reasons I could not figure out after 2h.

```cpp
virtual wxToolBarToolBase *CreateTool(int toolid,
                                          const wxString& label,
                                          const wxBitmapBundle& bmpNormal,
                                          const wxBitmapBundle& bmpDisabled = wxBitmapBundle(),
                                          wxItemKind kind = wxITEM_NORMAL,
                                          wxObject *clientData = nullptr,
                                          const wxString& shortHelp = wxEmptyString,
                                          const wxString& longHelp = wxEmptyString,
                                          bool stretchable = false ) = 0;
```

<<<< This is in the PR right now

```cpp
wxToolBarToolBase *wxToolBarBase::InsertStretchableSpace(size_t pos)
{
#ifdef __WXOSX_IPHONE__
   // The hack below does not work on iPhone as the tools are
   // created in the constructor not Realize(). I hack around the hack
   // by adding a distinct ID
   const int id = wxID_STRETCHABLE_SEPARATOR;
#else
   const int id = wxID_SEPARATOR;
#endif
    wxToolBarToolBase * const tool = CreateSeparator( id );
    if ( tool )
    {
        // this is a hack but we know that all the current implementations
        // don't really use the tool when it's created, they will do it
        // InsertTool() at earliest and maybe even in Realize() much later
        //
        // so we can create the tool as a plain separator and mark it as being
        // a stretchable space later
        tool->MakeStretchable();
    }

    return DoInsertNewTool(pos, tool);
}
```

vadz:
The "clean" solution is not really clean because `bool stretchable` only makes sense when combined with `wxITEM_SEPARATOR` and nothing else. A decent API shouldn't allow specifying nonsensical combinations of parameters, like `wxITEM_CHECK` and `stretchable=true`. Of course, `wxToolBar` is already guilty of it, but it's not a reason to make things worse.

The only alternative API I see is to add `wxITEM_STRETCHABLE_SEPARATOR` but this doesn't seem appealing. Making an existing separator stretchable doesn't seem very out of place as it's similar to making an existing check tool checked etc. Of course, if this _really_ can't be implemented for iOS we may want to revisit it, but for now I guess we can live with `wxID_STRETCHABLE_SEPARATOR` hack, although I think making `CreateSeparator()` virtual and overriding it in wxiOS `wxToolBar` would be slightly less ugly than checking for iOS in common code.

csomor:
> The "clean" solution is not really clean because `bool stretchable` only makes sense when combined with `wxITEM_SEPARATOR` and nothing else. A decent API shouldn't allow specifying nonsensical combinations of parameters, like `wxITEM_CHECK` and `stretchable=true`. Of course, `wxToolBar` is already guilty of it, but it's not a reason to make things worse.
> 
> The only alternative API I see is to add `wxITEM_STRETCHABLE_SEPARATOR` but this doesn't seem appealing. Making an existing separator stretchable doesn't seem very out of place as it's similar to making an existing check tool checked etc. Of course, if this _really_ can't be implemented for iOS we may want to revisit it, but for now I guess we can live with `wxID_STRETCHABLE_SEPARATOR` hack, although I think making `CreateSeparator()` virtual and overriding it in wxiOS `wxToolBar` would be slightly less ugly than checking for iOS in common code.

I'll try to move things into a separate PR. @vadz what about making `MakeStretchable` virtual ? I think I could recreate the native item.

vadz:
Sure, making `MakeStretchable()` virtual would be perfectly fine. TIA!

csomor:
As the different parts have been merged individually I'll close this soon if nobody objects

RobertRoeb:
All relevant code has been merged

PR #25843, opened 2025-09-30
Fix wxGetBestTreeSize function

### LINKED ISSUE #25839: Can't set width of `wxTreeCtrl` in `wxTreeBook`
In example `wxWidgets\samples\notebook\notebook.cpp`, I try to set the width, but failed. It always use the internal rectangle to do it.

Image

Image

ssrlive:
This PR will close #25839 . 

It also includes the invisible child nodes of the tree control in the best estimate width calculation.

The demo file is here [hello.zip](https://github.com/user-attachments/files/22613813/hello.zip)

And the fixed screenshot is here.

image

### DIFF
--- a/src/common/treebase.cpp
+++ b/src/common/treebase.cpp
@@ -220,7 +220,8 @@ wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size)
 {
     wxRect rect;
 
-    if ( treeCtrl->GetBoundingRect(id, rect, true /* just the item */) )
+    if ( treeCtrl->GetBoundingRect(id, rect, true /* just the item */)
+         && rect.width > 0 && rect.height > 0 ) // check for valid rect
     {
         // Translate to logical position so we get the full extent
 #if defined(__WXMSW__) && !defined(__WXUNIVERSAL__)
--- a/src/common/treebase.cpp
+++ b/src/common/treebase.cpp
@@ -209,7 +209,7 @@ wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size,
     wxRect rect;
 
     if ( treeCtrl->GetBoundingRect(id, rect, true /* just the item */)
-         && rect.width > 0 && rect.height > 0 ) // check for valid rect
+         && !rect.IsEmpty() ) // check for valid rect
     {
         // Translate to logical position so we get the full extent
 #if defined(__WXMSW__) && !defined(__WXUNIVERSAL__)
--- a/src/common/treebase.cpp
+++ b/src/common/treebase.cpp
@@ -26,6 +26,7 @@
 #include "wx/imaglist.h"
 #include "wx/dcclient.h"
 #include "wx/string.h"
+#include <algorithm>
 
 extern WXDLLEXPORT_DATA(const char) wxTreeCtrlNameStr[] = "treeCtrl";
 
@@ -244,9 +245,10 @@ wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size)
         if (treeCtrl->GetImageList() && imageIndex != -1)
             treeCtrl->GetImageList()->GetSize(imageIndex, iconWidth, h);
         int depth = wxGetTreeItemDepth(treeCtrl, id);
+        int height = std::max(h, textSize.GetHeight());
         int extraPadding = 20; // experience-based value, can be adjusted
         int totalWidth = indent * depth + iconWidth + textSize.GetWidth() + extraPadding;
-        size.IncTo(wxSize(totalWidth, size.GetHeight()));
+        size.IncTo(wxSize(totalWidth, height));
     }
 
     wxTreeItemIdValue cookie;
--- a/src/common/treebase.cpp
+++ b/src/common/treebase.cpp
@@ -203,21 +203,8 @@ void wxTreeCtrlBase::SetItemState(const wxTreeItemId& item, int state)
     DoSetItemState(item, state);
 }
 
-// Helper to get depth of a tree node
-static int wxGetTreeItemDepth(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id)
-{
-    int depth = 0;
-    wxTreeItemId parent = treeCtrl->GetItemParent(id);
-    while (parent.IsOk())
-    {
-        depth++;
-        parent = treeCtrl->GetItemParent(parent);
-    }
-    return depth;
-}
-
 static void
-wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size)
+wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size, int depth)
 {
     wxRect rect;
 
@@ -244,7 +231,6 @@ wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size)
         int imageIndex = treeCtrl->GetItemImage(id);
         if (treeCtrl->GetImageList() && imageIndex != -1)
             treeCtrl->GetImageList()->GetSize(imageIndex, iconWidth, h);
-        int depth = wxGetTreeItemDepth(treeCtrl, id);
         int height = std::max(h, textSize.GetHeight());
         int extraPadding = 20; // experience-based value, can be adjusted
         int totalWidth = indent * depth + iconWidth + textSize.GetWidth() + extraPadding;
@@ -256,7 +242,7 @@ wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size)
           item.IsOk();
           item = treeCtrl->GetNextChild(id, cookie) )
     {
-        wxGetBestTreeSize(treeCtrl, item, size);
+        wxGetBestTreeSize(treeCtrl, item, size, depth + 1);
     }
 }
 
@@ -293,7 +279,7 @@ wxSize wxTreeCtrlBase::DoGetBestSize() const
         // iterate over all items recursively
         wxTreeItemId idRoot = GetRootItem();
         if ( idRoot.IsOk() )
-            wxGetBestTreeSize(this, idRoot, size);
+            wxGetBestTreeSize(this, idRoot, size, 0);
     }
 
     // need some minimal size even for empty tree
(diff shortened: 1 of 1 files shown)


vadz:
To compute the size including the hidden items we probably could use `GetIndent()*depth + GetTextExtent()`.

I also wonder if we could just expand the tree in `wxTreebook` before computing its best size and then collapse it to show only the top level items?

ssrlive:
I think we can add a parameter to the constructor of `wxTreeCtrl` to indicate whether to expand the nodes during initialization.

I think whether the child nodes in the tree are expanded or not will not affect the current implementation of `wxGetBestTreeSize`.

vadz:
> I think we can add a parameter to the constructor of `wxTreeCtrl` to indicate whether to expand the nodes during initialization.

No, why should it be a ctor parameter? It's enough to call `Expand()` recursively after constructing it.
 
> I think whether the child nodes in the tree are expanded or not will not affect the current implementation of `wxGetBestTreeSize`.

It should, because I guess that `GetBoundingRect()` doesn't return the correct rectangle for collapsed items, but if they're expanded, it really should work.

ssrlive:
I agree.

ssrlive:
Why do so many tests fail?

vadz:
> Why do so many tests fail?

If you click on the links, you will see the error messages.

ssrlive:
I don't know the reason.
```
make[1]: Leaving directory '/__w/wxWidgets/wxWidgets/utils/wxrc'
make: Target 'all' not remade because of errors.
Error: Process completed with exit code 2.
```

vadz:
Please search for error above. E.g. use search box on GitHub to search for `***` or ` error` etc.

ssrlive:
I don't think these failures are caused by my changes. I'd appreciate if you could check the build script.

vadz:
> I don't think these failures are caused by my changes. I'd appreciate if you could check the build script.

Sorry but you're wrong. I haven't looked at all of the errors, but I *had* looked at some of them before commenting and there are plenty of them due to missing header in the code you added. You really should look at the build logs carefully, this is not exactly rocket science.

src/common/treebase.cpp
```diff
@@ -215,6 +230,23 @@ wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size)
...
+        wxClientDC dc(const_cast<wxTreeCtrlBase*>(treeCtrl));
+        dc.SetFont(treeCtrl->GetFont());
+        wxSize textSize = dc.GetTextExtent(label);
+        int indent = treeCtrl->GetIndent();
+        int iconWidth = 0, h = 0;
+        int imageIndex = treeCtrl->GetItemImage(id);
+        if (treeCtrl->GetImageList() && imageIndex != -1)
+            treeCtrl->GetImageList()->GetSize(imageIndex, iconWidth, h);
+        int depth = wxGetTreeItemDepth(treeCtrl, id);
+        int extraPadding = 20; // experience-based value, can be adjusted
```
vadz: This should at least use `treeCtrl->FromDIP()` but it really shouldn't be necessary in the first place. Can we add an extra `indent` instead, perhaps?

src/common/treebase.cpp
```diff
@@ -215,6 +230,23 @@ wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size)
...
+    {
+        // Estimate width for collapsed (invisible) node
+        wxString label = treeCtrl->GetItemText(id);
+        wxClientDC dc(const_cast<wxTreeCtrlBase*>(treeCtrl));
+        dc.SetFont(treeCtrl->GetFont());
+        wxSize textSize = dc.GetTextExtent(label);
+        int indent = treeCtrl->GetIndent();
+        int iconWidth = 0, h = 0;
+        int imageIndex = treeCtrl->GetItemImage(id);
+        if (treeCtrl->GetImageList() && imageIndex != -1)
```
vadz:
```suggestion
        if ( treeCtrl->HasImages() && imageIndex != wxTreeCtrl::NO_IMAGE )
```

src/common/treebase.cpp
```diff
@@ -215,6 +230,23 @@ wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size)
 
         size.IncTo(wxSize(rect.GetRight(), rect.GetBottom()));
     }
+    else if (id != treeCtrl->GetRootItem())
+    {
+        // Estimate width for collapsed (invisible) node
+        wxString label = treeCtrl->GetItemText(id);
+        wxClientDC dc(const_cast<wxTreeCtrlBase*>(treeCtrl));
```
vadz: We should avoid using `wxClientDC` in the new code.
```suggestion
        wxInfoDC dc(const_cast<wxTreeCtrlBase*>(treeCtrl));
```

src/common/treebase.cpp
```diff
@@ -215,6 +230,23 @@ wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size)
...
+        dc.SetFont(treeCtrl->GetFont());
```
vadz: This is unnecessary, the DC already inherits window's font.
```suggestion
```

src/common/treebase.cpp
```diff
@@ -215,6 +230,23 @@ wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size)
...
+        wxSize textSize = dc.GetTextExtent(label);
+        int indent = treeCtrl->GetIndent();
+        int iconWidth = 0, h = 0;
+        int imageIndex = treeCtrl->GetItemImage(id);
+        if (treeCtrl->GetImageList() && imageIndex != -1)
+            treeCtrl->GetImageList()->GetSize(imageIndex, iconWidth, h);
```
vadz:
```suggestion
            iconWidth = treeCtrl->GetImageLogicalSize(treeCtrl, imageIndex).x;
```

src/common/treebase.cpp
```diff
@@ -200,6 +202,19 @@ void wxTreeCtrlBase::SetItemState(const wxTreeItemId& item, int state)
     DoSetItemState(item, state);
 }
 
+// Helper to get depth of a tree node
+static int wxGetTreeItemDepth(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id)
```
vadz: Use of this function results in `O(N^2)` algorithm which is probably not catastrophic for `wxTreebook` but still seems quite unnecessary: just add `int depth = 0` parameter to `wxGetBestTreeSize()` and call it with `depth + 1` when calling it recursively, then you don't have to compute the depth for each item at all.

vadz (commented):
Thanks for fixing the problems, but there are still a few things that could be improved here.

Most problematic is, of course, the hardcoded 20px value. Did you only test this under Windows? And, if so, at which DPI (100%, 200%)?

ssrlive:
After patch https://github.com/wxWidgets/wxWidgets/pull/25843/commits/8103d93cbfa5e32d48502d5790d6f6052f7d55a3 , the behave in Linux looks fine.

image

ssrlive:
After this patch, it looks fine in macOS. https://github.com/wxWidgets/wxWidgets/pull/25843/commits/e0b1d6d79a7d41bbf5a35d25e20b66b71bfce871

image

src/common/treebase.cpp
ssrlive: Finished in https://github.com/wxWidgets/wxWidgets/pull/25843/commits/239d84bb57679a044469e97fbb9bd303298ec1ac

src/common/treebase.cpp
```diff
@@ -201,11 +204,12 @@ void wxTreeCtrlBase::SetItemState(const wxTreeItemId& item, int state)
...
 
 static void
-wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size)
+wxGetBestTreeSize(const wxTreeCtrlBase* treeCtrl, wxTreeItemId id, wxSize& size, int depth)
 {
     wxRect rect;
 
-    if ( treeCtrl->GetBoundingRect(id, rect, true /* just the item */) )
+    if ( treeCtrl->GetBoundingRect(id, rect, true /* just the item */)
+         && rect.width > 0 && rect.height > 0 ) // check for valid rect
```
PBfordev:
```suggestion
         && !rect.IsEmpty() )
```

src/common/treebase.cpp
ssrlive: Apply in https://github.com/wxWidgets/wxWidgets/pull/25843/commits/2ded700ec22543d5b12de86b4920fd6c438653fd

vadz (commented):
Thanks, looks good now but I'm still bothered by this hard-coded 20. Have you tried using extra `indent` instead?

ssrlive:
I have test Linux macOS, if I set extra padding to 0 them all ok. But it is
not okay in windows.

ssrlive:
Indent is 19 in windows, so we can use it to replace extra padding.

vadz:
What about this idea: get the rect of the visible item (there must be at least one), then compute its estimated size using your method and then compute the width (and maybe even height?) difference. And then add this difference to all the sizes computed using your method.

AFAICS this should work reliably in all cases.

ssrlive:
In order to make the code as simple as possible, I think it's a good idea
to use indent directly. Anyway, there should be at least one indent on the
left side of any node, and adding an indent on the right side of the
longest entry is a good reason for aesthetic reasons.

vadz:
There is clearly too much space in your Mac screenshot, so I think it would be worth making the code a bit more complex if this would be more precise. And I don't think it's that complicated anyhow, it's enough to just extract the code for computing the item extents into a separate function and call it twice.

ssrlive:
Here's what I've done.
I've replaced extraPadding with indent.
If you don't like it, feel free to change it accordingly.
You can make changes directly on my fork, commit them to my branch, and then merge them into the `wxWidgets` main branch.

https://github.com/ssrlive/wxWidgets/blob/fix-tree-ctrl/src/common/treebase.cpp#L234

ssrlive:
After the patch https://github.com/wxWidgets/wxWidgets/pull/25843/commits/c6d9248ff59e16e58a603872834d844ba96a8a44

It looks the same behaviors.

image

image

image

vadz:
I've finally returned to this and it seems that we don't compute the vertical component of the best size correctly at all: we just take the item height into account but not its position. It's not important for this particular use case, but it could still be useful to do it correctly as well. AFAICS this could be done by keeping track of running height... I'll try to do this and merge this then.

vadz:
I've finally fixed height computation (I think) and made #25930 replacing this PR. Please test this change and let me know (there) if you see any problems with it. Thanks!

PR #25850, opened 2025-10-01
CMake: Use 'find_dependency()' in Config file

vilkov:
It is designed to be used in a Package Configuration File (<PackageName>Config.cmake). find_dependency() forwards the correct parameters for QUIET and REQUIRED which were passed to the original find_package() call.

See
 https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html

### DIFF
--- a/build/cmake/wxWidgetsConfig.cmake.in
+++ b/build/cmake/wxWidgetsConfig.cmake.in
@@ -147,20 +147,22 @@ foreach(libname @wxLIB_TARGETS@)
     endif()
 endforeach()
 
+include(CMakeFindDependencyMacro)
+
 if(TARGET wx::wxnet AND @wxUSE_WEBREQUEST_CURL@)
     # make sure CURL targets are available:
     # The link interface of target "wx::wxnet" contains: CURL::libcurl_shared
-    find_package(CURL QUIET)
+    find_dependency(CURL)
 endif()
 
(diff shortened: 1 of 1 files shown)


vadz:
Something probably needs to be included to make this usable.

PBfordev:
> Something probably needs to be included to make this usable.

Probably
`include(CMakeFindDependencyMacro)`

vilkov:
Thanks @vadz @PBfordev, missed that one

MaartenBent:
Finding OpenGL doesn't work on iOS, see also our workaround here: https://github.com/wxWidgets/wxWidgets/blob/16af160f7f04442d1a7d9364e24c4b6b8e05bea0/build/cmake/init.cmake#L506-L511

But because we don't use `find_package(OpenGL)` for iOS, there wil also be no `OpenGL::<target>` in the config files, and we don't need to search for them to fix any unknown target errors.

So I think you can simply skip it, something like this will probably work:

```cmake
if (NOT (APPLE AND IOS))
    find_dependency(OpenGL)
endif()
```

vadz:
This can't be applied before the CI error is fixed, please check if Maarten's suggestion above works.

vilkov:
Done. BTW, @MaartenBent, thanks for 'CMake:' prefix, it helps a lot.

Issue #16428, opened 2014-08-17
wxTreeCtrl line height doesn't change with the font size


wxtrac:
*Issue migrated from trac ticket # 16428*

**component:** wxGTK | **priority:** low | **keywords:** wxTreeCtrl

#### 2014-08-17 22:03:40: kelvin_miller created the issue

___
When making the font size of the wxTreeCtrl control smaller, the height of the item remains the same.

How to reproduce: 
compile attached example, launch the application. Press the "Restore Font" button. As a result, the height of the item of the right control will  be bigger than height of the item of the left control.

What i expect: i expect that height of the item should change together with the font size.

PR #25851, opened 2025-10-02
(#16428) allow to set custom font for generic tree ctl (gtk)

basiliscos:
Hi.

This allows to set custom font the tree component on linux. 

Also should fix the issue https://github.com/wxWidgets/wxWidgets/issues/16428

### DIFF
--- a/src/generic/treectlg.cpp
+++ b/src/generic/treectlg.cpp
@@ -1056,11 +1056,19 @@ void wxGenericTreeCtrl::InitVisualAttributes()
 
     m_dottedPen = wxPen(wxSystemSettings::GetColour(wxSYS_COLOUR_GRAYTEXT), 1, wxPENSTYLE_DOT);
 
+    if ( !m_hasExplicitFont )
+    {
 #if defined(__WXOSX__)
-    m_normalFont = wxFont(wxOSX_SYSTEM_FONT_VIEWS);
+        m_normalFont = wxFont(wxOSX_SYSTEM_FONT_VIEWS);
 #else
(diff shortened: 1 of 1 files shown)


vadz (commented):
Thanks, this is probably correct, although the interaction of this check with the check for `!m_hasExplicitFont` just above is rather confusing.

I do have to wonder why do we need `m_normalFont` at all instead of just using `m_font`...

basiliscos:
> I do have to wonder why do we need m_normalFont at all instead of just using m_font

I have no idea. If you are k, I can remove all usages of `m_normalFont` (+ probably `m_hasExplicitFont`)

vadz:
Hmm, actually this doesn't fix the problem reported in #16428 for me, after applying it the item height is still not decreased after resetting the font.

P.S. OTOH that problem is fixed by the following trivial change:

```diff
diff --git a/src/generic/treectlg.cpp b/src/generic/treectlg.cpp
index fc40ba1745..4bb4108d34 100644
--- a/src/generic/treectlg.cpp
+++ b/src/generic/treectlg.cpp
@@ -1318,6 +1318,8 @@ bool wxGenericTreeCtrl::SetFont( const wxFont &font )
     m_normalFont = font;
     m_boldFont = m_normalFont.Bold();
 
+    CalculateLineHeight();
+
     if (m_anchor)
         m_anchor->RecursiveResetTextSize();
 
```

vadz:
Finally, I just don't understand what problem this PR fixes, so I'd rather not apply this. However I made #25927 which fixes a couple of problems, including the original bug and maybe whatever problem you had. Please test that PR and if it doesn't fix your problem, please describe it there. Thanks!

PR #25852, opened 2025-10-02
Clarify wxLog::SetActiveTarget

Blake-Madden:
This explanation is a bit redundant with the rest of the docs, but really helps clarify how `wxLog::SetActiveTarget` should work.


vadz:
Thanks, I'll apply and push this soon.

PR #25853, opened 2025-10-02
OSX iOS support appearance colours

csomor:


src/osx/iphone/colour.mm
```diff
@@ -0,0 +1,201 @@
+/////////////////////////////////////////////////////////////////////////////
+// Name:        src/osx/iphone/colour.mm
+// Purpose:     Conversions between UIColor and wxColour
+// Author:      Vadim Zeitlin
+// Created:     2015-11-26 (completely replacing the old version of the file)
+// Copyright:   (c) 2015 Vadim Zeitlin
```
vadz: My name shouldn't be here.

src/osx/iphone/settings.mm
```diff
@@ -18,9 +18,55 @@
...
+
+// ----------------------------------------------------------------------------
+// wxSystemAppearance
+// ----------------------------------------------------------------------------
+
+wxString wxSystemAppearance::GetName() const
+{
+    UITraitCollection* traitCollection = [UIScreen mainScreen].traitCollection;
+    if (traitCollection.userInterfaceStyle == UIUserInterfaceStyleDark)
+        return wxT("Dark");
```
vadz: We shouldn't use `wxT()` in new code, it should be either `wxS()` or nothing at all.

```suggestion
        return wxS("Dark");
```

src/osx/iphone/settings.mm
```diff
@@ -18,9 +18,55 @@
...
+    else if (traitCollection.userInterfaceStyle == UIUserInterfaceStyleLight)
+        return wxT("Light");
```
vadz:
```suggestion
        return wxS("Light");
```

src/osx/iphone/settings.mm
```diff
@@ -18,9 +18,55 @@
...
+    else
+        return wxT("Unspecified");
```
vadz:
```suggestion
        return wxS("Unspecified");
```

vadz (commented):
I am not sure why all trailing whitespace got removed from `Makefile.in`, but we probably don't want to do this, as it will be added back during next regeneration.

src/osx/iphone/colour.mm
csomor: sorry, copied the headers over from the cocoa version ...

csomor:
> I am not sure why all trailing whitespace got removed from `Makefile.in`, but we probably don't want to do this, as it will be added back during next regeneration.

oh, sorry, my BBEdit seems to have messed this up

PR #25857, opened 2025-10-03
Implement wxNotebook for iOS

RobertRoeb:
- Supports wxBK_TOP using UISegmentedControl
- and wxBK_BOTTOM using UITabView


vadz:
Thanks, @csomor do you have any remarks about this one or should it be merged? Notably, perhaps you know of a way to avoid hardcoding these 35 and 4 in the page size determination function?

csomor:
> Thanks, @csomor do you have any remarks about this one or should it be merged? Notably, perhaps you know of a way to avoid hardcoding these 35 and 4 in the page size determination function?

Once I have widgets sample running, I can see what we can get from the controls themselves

PR #25860, opened 2025-10-05
OSX make iOS and macOS share more drawing code

csomor:


src/generic/regiong.cpp
```diff
@@ -375,7 +375,9 @@ bool wxRegionGeneric::DoOffset(wxCoord x, wxCoord y)
 
 bool wxRegionGeneric::IsEmpty() const
 {
-    wxASSERT(m_refData);
+    if ( m_refData == nullptr )
```
vadz, line 378: Do we really want to do this? I think invalid and empty regions are not the same thing, just as transparent and invalid pens/brushes, for example.

csomor: Sorry, I wrote this twice as I've somehow managed to loose a set of commits, the reason behind it is the documented behavior:

https://docs.wxwidgets.org/3.3/classwx_region.html#a1edc6768118cf02749b46774a0ca37f9

a freshly created wxRegion is invalid but also empty

and apparently other platforms behave like this, already, I just ran into an assertion when setting the background color on a wxFrame on iOS, which doesn't have a native region class.

vadz: Ha, I had even added this comment to the docs myself back in dd4eefcb29 (Document and test behaviour of wxRegion methods when it is invalid., 2011-10-18) but, of course, completely forgot about it. And, as the commit says, there is even a unit test ("wxRegion::Validity") checking for this, so this would have been discovered if we ran unit tests under iOS...

Anyhow, sorry, for not checking this first.

csomor: no problem, I really appreciate you looking at things closely

csomor:
@vadz Sorry, but I need your help, I don't understand why the cirrus ci / FreeBSD suddenly fails, I don't see how I could have changed something in GTK land, can you see a reason ?

vadz:
It looks like they have updated GTK version which now (since 2.44 which is too new to even have docs for it online apparently) deprecates a bunch of `gdk_pixbuf_loader_xxx()` functions. This is completely unrelated to your changes and I'll try to fix this as soon as I can in master.

csomor:
> It looks like they have updated GTK version which now (since 2.44 which is too new to even have docs for it online apparently) deprecates a bunch of `gdk_pixbuf_loader_xxx()` functions. This is completely unrelated to your changes and I'll try to fix this as soon as I can in master.

Thanks a lot, ok I ll  dare to squash merge then

vslavik:
The wxArtProvider part of these changes broke loading of system (incl. stored in `Resources` or asset catalog) bitmaps, which was part of wx since at least 3.0. Specifically, the removal of 

```cpp
 return wxOSXCreateSystemBitmapBundle(id, client, size);
```

from `CreateBitmapBundle`: https://github.com/wxWidgets/wxWidgets/commit/20ff462703a327b0d2b5c344e71dbf23b2a88610#diff-3acbe1af29f31859a4eb1e08c767105fb787e8f23e4cf62add7a05605c8db887L40-L44

Can the old behavior please be restored?

csomor:
Hi @vslavik sorry about that, I'll look into it.

csomor:
@vslavik could you please verify with #26489, thanks

PR #25864, opened 2025-10-06
Fix nested markup conversion to NSAttributedString

vslavik:
`wxMarkupToAttrStringBase` implementation of markup parsing failed to correctly apply nested attributes due to applying - and overwriting - attributes to a span in `OnAttrEnd()`, which is called first for the inner span, and then for the outer one, overwriting all inner changes.

For example, consider this markup:
```html
<i>This <b>word</b> is bold.</i>
```
Here, the entire text would be rendered in italics, but not bold, as the code would first apply bold+italics font to “word”, and then proceed to apply italics-only to the entire “This word is bold.” range.

The fix is to instead apply the currently effective attributes to text in `OnText()`, similarly to how `wxMarkupParserRenderOutput` does it. As a side effect, this also eliminates the need to apply default font to the entire range first.

(This is v2 of the fix. My first implementation was to collect the spans with their attributes in a list and then “replay” them in sequential order at the end of `Parse()`. But this PR’s version is simpler and I think even more efficient overall, as `addAttribute(s):` is only called once for any character in the range, with no overwrites or overlaps.)

### DIFF
--- a/include/wx/osx/cocoa/private/markuptoattr.h
+++ b/include/wx/osx/cocoa/private/markuptoattr.h
@@ -37,12 +37,6 @@ protected:
 
         [m_attrString beginEditing];
 
-        // First thing we do is change the default string font: as mentioned in
-        // Apple documentation, attributed strings use "Helvetica 12" font by
-        // default which is different from the system "Lucida Grande" font. So
-        // we need to explicitly change the font for the entire string.
-        ApplyFont(font, NSMakeRange(0, [m_attrString length]));
-
         // Now translate the markup tags to corresponding attributes.
         wxMarkupParser parser(*this);
         parser.Parse(markup);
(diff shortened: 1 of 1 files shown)


vadz:
This looks much nicer and simpler, thanks!

I guess this should be backported to 3.2 too, right (after `s/override/wxOVERRIDE/`)?

vslavik:
> I guess this should be backported to 3.2 too, right (after s/override/wxOVERRIDE/)?

I think it’s safe to, yes. The 3.2 version [^1] is here: https://github.com/poedit/wxWidgets/commit/9a16422a57b1ff39206a8e4a384ecdee384854f6

[^1]: There were differences in presence of `override` too, and `OSXGetWXColor()` didn’t exist yet.

PR #25865, opened 2025-10-06
Document ribbon bar functions

Blake-Madden:
These appear to have all been added ~3.1.0.

### DIFF
--- a/interface/wx/ribbon/bar.h
+++ b/interface/wx/ribbon/bar.h
@@ -462,6 +462,57 @@ public:
     */
     wxRibbonDisplayMode GetDisplayMode() const;
 
+    /**
+        Returns whether the bar contains multiple pages.
+
+        The default implementation always returns @c true.
+
+        @since 3.1.0
+    */
+    virtual bool HasMultiplePages() const;
+
+    /**
+        Returns @c true if the toggle button is currently moused over.
+
+        @since 3.1.0
+    */
+    bool IsToggleButtonHovered() const;
+
+    /**
+        Returns @c true if the help button is currently moused over.
+
+        @since 3.1.0
+    */
+    bool IsHelpButtonHovered() const;
+
+    /**
+        Hide the panels if its display mode is @c wxRIBBON_BAR_EXPANDED.
+
+        @since 3.1.0
+    */
+    void HideIfExpanded();
+
+    /**
+        Sets the wxRibbonBarOption flags for the ribbon bar.
+
+        @since 3.1.0
+    */
+    void SetWindowStyleFlag(long style);
+
+    /**
+        Returns the ribbon bar's wxRibbonBarOption flags.
+
+        @since 3.1.0
+    */
+    long GetWindowStyleFlag() const;
+
+    /**
+        Return the image list containing images of the given size, creating it
+        if necessary.
+
+        @since 3.1.0
+    */
+    wxImageList* GetButtonImageList(wxSize size);
 
     /**
         Perform initial layout and size calculations of the bar and its
--- a/interface/wx/ribbon/bar.h
+++ b/interface/wx/ribbon/bar.h
@@ -466,51 +466,21 @@ public:
         Returns whether the bar contains multiple pages.
 
         The default implementation always returns @c true.
-
-        @since 3.1.0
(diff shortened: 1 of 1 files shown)


interface/wx/ribbon/bar.h
```diff
@@ -462,6 +462,57 @@ class wxRibbonBar : public wxRibbonControl
     */
     wxRibbonDisplayMode GetDisplayMode() const;
 
+    /**
+        Returns whether the bar contains multiple pages.
+
+        The default implementation always returns @c true.
+
+        @since 3.1.0
```
vadz: This function seems to exist since the very first wxRibbon version.

interface/wx/ribbon/bar.h
```diff
@@ -462,6 +462,57 @@ class wxRibbonBar : public wxRibbonControl
...
+    */
+    bool IsToggleButtonHovered() const;
+
+    /**
+        Returns @c true if the help button is currently moused over.
+
+        @since 3.1.0
+    */
+    bool IsHelpButtonHovered() const;
```
vadz: These functions are under "Implementation only." comment, I don't think they should be documented.

interface/wx/ribbon/bar.h
```diff
@@ -462,6 +462,57 @@ class wxRibbonBar : public wxRibbonControl
...
+
+    /**
+        Hide the panels if its display mode is @c wxRIBBON_BAR_EXPANDED.
+
+        @since 3.1.0
+    */
+    void HideIfExpanded();
```
vadz, line 477: This one seems private too, but it might make sense to make it public and document it.

However it was added in 42d7394119 (Add help button support to wxRibbonBar., 2012-09-15) which was included in 2.9.5.

interface/wx/ribbon/bar.h
```diff
@@ -462,6 +462,57 @@ class wxRibbonBar : public wxRibbonControl
...
+        @since 3.1.0
+    */
+    void SetWindowStyleFlag(long style);
+
+    /**
+        Returns the ribbon bar's wxRibbonBarOption flags.
+
+        @since 3.1.0
+    */
+    long GetWindowStyleFlag() const;
```
vadz: Those are inherited from wxWindow, is there a good reason to redocument them here?

interface/wx/ribbon/bar.h
```diff
@@ -462,6 +462,57 @@ class wxRibbonBar : public wxRibbonControl
...
+
+    /**
+        Return the image list containing images of the given size, creating it
+        if necessary.
+
+        @since 3.1.0
+    */
+    wxImageList* GetButtonImageList(wxSize size);
```
vadz, line 485: This one was added in 3.1.4 but I am not sure if it should be made public/documented, especially because it is not `wxBitmapBundle`-friendly...

vadz (changes requested):
Sorry but I have several questions about these changes.

interface/wx/ribbon/bar.h
Blake-Madden: Fixed

interface/wx/ribbon/bar.h
Blake-Madden: OK, I removed this.

interface/wx/ribbon/bar.h
Blake-Madden: OK, removed those.

interface/wx/ribbon/bar.h
Blake-Madden: This one is public currently. I fixed the change version.

interface/wx/ribbon/bar.h
Blake-Madden: I fixed the changed version.

So, this is the function that caught my eye and wanted to document it. I was hoping I could later add a default `initialSize` parameter to this (defaulting to the current `1`). That would be a nice optimization when needing to load in a lot of icons like I do.

vadz:
Thanks for the updates.

I'm still not sure about `GetButtonImageList()` but I guess we won't replace it in the near future, so let's document it until this can be done.

PR #25867, opened 2025-10-07
OSX iOS Notebook addition

csomor:
can you try it with this, it uses the build int best size for the UISegmented control, this leaves only the border around as a wx-defined const, which is arbitrary so we might just as well start with 4 until we find a reference for the best value

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -5611,6 +5611,7 @@ COND_TOOLKIT_OSX_IPHONE___GUI_SRC_OBJECTS =  \
 	monodll_iphone_gauge.o \
 	monodll_iphone_msgdlg.o \
 	monodll_iphone_nonownedwnd.o \
+	monodll_iphone_notebook.o \
 	monodll_iphone_scrolbar.o \
 	monodll_iphone_slider.o \
 	monodll_iphone_stattext.o \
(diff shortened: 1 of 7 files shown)


csomor:
@vadz I don't know whether I did this correctly, this would be my change to get rid of the hardcoded height, I thought I could commit into an existing PR but apparently I cannot and opened a new PR. But with this change, I'd be fine with Robert's version and you could merge if that's ok

vadz:
> @vadz I don't know whether I did this correctly, this would be my change to get rid of the hardcoded height,

Thanks, but I actually meant the height in `src/osx/notebook_osx.cpp` and not in `src/osx/iphone/notebook.mm`. Anyhow, let me merge this already and let you improve it later.

> I thought I could commit into an existing PR but apparently I cannot and opened a new PR.

I think you should be able to push to Robert's repository as you're a maintainer of this one, and this would be the best way to add something to an existing PR. In any case, it's better to avoid pushing branches directly to this repository (as happened with `pullrequests/RobertRoeb/iphone_notebook` branch here) and I don't think using "pullrequests" prefix has any special meaning — it's still just a branch (which I will delete after merging).

csomor:
> > @vadz I don't know whether I did this correctly, this would be my change to get rid of the hardcoded height,
> 
> Thanks, but I actually meant the height in `src/osx/notebook_osx.cpp` and not in `src/osx/iphone/notebook.mm`. Anyhow, let me merge this already and let you improve it later.

yes, I'll do that, I'll also add the parts for the other controls, so that I can then improve and consolidate things, once we have a more or less complete build of widgets sample

> > I thought I could commit into an existing PR but apparently I cannot and opened a new PR.
> 
> I think you should be able to push to Robert's repository as you're a maintainer of this one, and this would be the best way to add something to an existing PR. In any case, it's better to avoid pushing branches directly to this repository (as happened with `pullrequests/RobertRoeb/iphone_notebook` branch here) and I don't think using "pullrequests" prefix has any special meaning — it's still just a branch (which I will delete after merging).

Ok, thanks and sorry for the mess ...

PR #25868, opened 2025-10-07
Add support for reserving X number of images for image list (including ribbon's image list)

Blake-Madden:


interface/wx/ribbon/bar.h
```diff
@@ -480,9 +480,14 @@ class wxRibbonBar : public wxRibbonControl
         Return the image list containing images of the given size, creating it
         if necessary.
 
+        @param size The size of the images going into this image list.
+        @param initialCount The initial number of images to reserve for the
+            image list.
```
vadz:
```suggestion
            image list (this parameter is available since wxWidgets 3.3.2).
```

interface/wx/ribbon/bar.h
```diff
@@ -480,9 +480,14 @@ class wxRibbonBar : public wxRibbonControl
...
+        @return The existing (or newly created) image list.
+
         @since 3.1.4
     */
-    wxImageList* GetButtonImageList(wxSize size);
+    wxImageList* GetButtonImageList(wxSize size, int initialCount);
```
vadz:
```suggestion
    wxImageList* GetButtonImageList(wxSize size, int initialCount = 1);
```

src/generic/imaglist.cpp
```diff
@@ -58,12 +58,14 @@ int wxGenericImageList::GetImageCount() const
...
 }
 
-bool wxGenericImageList::Create( int width, int height, bool mask, int WXUNUSED(initialCount) )
+bool wxGenericImageList::Create( int width, int height, bool mask, int initialCount )
 {
     // Prevent from storing negative dimensions
     m_size = wxSize(wxMax(width, 0), wxMax(height, 0));
     m_useMask = mask;
 
+    m_images.reserve(initialCount);
```
vadz: Maybe

```suggestion
    if ( initialCount > 1 )
        m_images.reserve(initialCount);
```

to avoid a useless (I believe) call by default?

vadz (commented):
As I said, I'd rather stop using `wxImageList` completely, but as long as don't have any alternative here, I don't see any problem with adding this.

PR #25870, opened 2025-10-08
Document wxBUILD_DEBUG_LEVEL in CMake overview

Blake-Madden:
Can be useful if you want to crank up debugging support throughout your wx build when statically linking (or bumping it done to 0 for a release build if you don't want any asserts).


vadz:
Thanks, will merge soon.

PR #25877, opened 2025-10-12
Fix bug in drawing sample when using wxGC to draw on a scrolled window

AliKet:
A clip region set using SetDeviceClippingRegion() on a DC should not scroll when the canvas we are drawing on scrolls.  The drawing sample behaves like this if the DC is a wxDC instance but not if it is a wxGC instance (because it is not properly prepared when the canvas is scrolled, i.e.: GetDeviceOrigin() always returns (0, 0) for a wxGC). So this bug should be fixed by preparing the correct DC that we are actually using for drawing.

### DIFF
--- a/samples/drawing/drawing.cpp
+++ b/samples/drawing/drawing.cpp
@@ -2100,13 +2100,11 @@ void MyCanvas::OnPaint(wxPaintEvent &WXUNUSED(event))
     if ( m_useBuffer )
     {
         wxBufferedPaintDC bpdc(this);
-        PrepareDC(bpdc); // Adjust scrolled contents.
         Draw(bpdc);
     }
     else
     {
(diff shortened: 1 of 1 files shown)


RobertRoeb:
Do the different ports behave the same then? I remember you reported a platform difference initially

AliKet:
> Do the different ports behave the same then?

Yes, but we should pass the `wxGC` to `PrepareDC()` and not the original `wxDC` for things to work properly.

vadz:
Thanks, I'm not quite sure that there is no other bug here, i.e. I'd expect preparing (= shifting origin) of the wxDC to be inherited by wxGC too and this doesn't seem to happen, but preparing wxGC itself should definitely work in any case, so the new code is definitely correct.

Will push in a moment, thanks again.

PR #25880, opened 2025-10-15
Fix macOS compile warning (unused var)

Blake-Madden:


vadz:

PR #25881, opened 2025-10-15
Ribbon: add bitmap check to avoid assert/bad image handling

Blake-Madden:


Blake-Madden:
Perhaps this can be rolled into the October backport as well.

vadz:
Thanks, will push soon with a minor optimization as long as we're changing this code.

PR #25882, opened 2025-10-15
Improve HiDPI padding in generic wxDataViewCtrl

vslavik:
This seems to be a case of forgotten `FromDIP` — the padding looks almost nonexistent at 200%.

(3.2 is also affected.)


vadz:
Thanks, will merge and cherry pick.

PR #25884, opened 2025-10-15
fix wxGridColumnOperations IsLineInSelection

DietmarSchwertberger:
row/col were swapped for the selection check.


vadz:
Oops. Thanks for noticing and fixing this!

PR #25885, opened 2025-10-15
Minor fixes for generic wxTreeCtrl on OSX

RobertRoeb:
- When not having focus, the selection rectangle is light grey and then we have to use the standard black pen to draw over the selection rect, not the white pen used over the blue selection rectangle, when the control has the focus
  - toggle the tree also with DClick events like the native control and as you would expect anyways

### DIFF
--- a/src/generic/treectlg.cpp
+++ b/src/generic/treectlg.cpp
@@ -3903,11 +3903,7 @@ void wxGenericTreeCtrl::OnMouse( wxMouseEvent &event )
             {
                 // only toggle the item for a single click, double click on
                 // the button doesn't do anything (it toggles the item twice)
-                // Reply: wxWidgets only sends one LeftDown() event and if you
-                // keep clicking, a series of DClick events are sent - so the tree
-                // branch doesn't toggle if you click mulitple times. Therefore,
-                // I re-added || event.LeftDClick()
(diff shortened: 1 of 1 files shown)


src/generic/treectlg.cpp
```diff
@@ -3902,7 +3903,11 @@ void wxGenericTreeCtrl::OnMouse( wxMouseEvent &event )
             {
                 // only toggle the item for a single click, double click on
                 // the button doesn't do anything (it toggles the item twice)
-                if ( event.LeftDown() )
+                // Reply: wxWidgets only sends one LeftDown() event and if you
```
vadz: Thanks, I'm going to merge this but I don't understand the format of this comment, what is this replying to?

vadz:
Taking into account your post about missing `wxEVT_LEFT_DOWN` events in wxOSX, I take my words back: we probably shouldn't apply this if we can fix that bug instead.

I can still apply the first part if you'd like.

RobertRoeb:
> I can still apply the first part if you'd like.

Exactly, I just removed the rest

vadz:
Thanks, I'll merge this soon but FWIW I think this is still wrong, we really shouldn't be hard coding "white" anywhere, as it must not work in dark mode.

The same applies to `wxBLACK_PEN` used under non-Mac platforms too, of course...

PR #25888, opened 2025-10-17
Fix DrawLabel under Retina display

Blake-Madden:
Use the scaled versions of bitmap width/height so that `wxDC::DrawLabel` correctly positions the text under macOS Retina displays.

Before:

Screenshot 2025-10-17 at 8 03 12 AM

After:

Screenshot 2025-10-17 at 7 59 25 AM

I use floor and the cast for deterministic rounding and to avoid compile warnings, sorry that it looks so much more cryptic now.

### DIFF
--- a/src/common/dcbase.cpp
+++ b/src/common/dcbase.cpp
@@ -1131,8 +1131,8 @@ void wxDC::DrawLabel(const wxString& text,
     wxCoord width, height;
     if ( bitmap.IsOk() )
     {
-        width = widthText + static_cast<wxCoord>(std::floor(bitmap.GetScaledWidth()));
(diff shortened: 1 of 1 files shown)


vadz (commented):
Thanks, the change is indeed correct, but could you please use `GetLogicalXXX()` instead of `GetScaledXXX()`? As the comment says, the former names are preferred.

I'm also not sure about the use of `floor()` rather than `wxRound()` (which would also avoid the casts), it looks like it could be better to make width/height 1px bigger than 1px smaller?

Blake-Madden:
OK, applied suggestions.

vadz:

PR #25894, opened 2025-10-19
Add SetFormat to wxDatePickerCtrlGeneric

szpeak:
1.Cut DateValidator and wxCalendarComboPopup from .cpp to .h
2.Chang wxCalendarComboPopup::SetFormat to public method
3.Add SetFormat to wxDatePickerCtrlGeneric


vadz:
Sorry but why did you decide to move all this code into the public header? This shouldn't be necessary at all.

Also note that

1. New public member function must be documented.
2. Its documentation should clearly state that it's only available in the generic version.

Because of (2) it might also be better to rename it to `SetGenericFormat()` or something like this to show more clearly that it doesn't work for the native versions.

vadz:
Sorry, this is unreviewable due to a huge number of whitespace-only changes and you still didn't explain why did you move private code into the header — please don't do this.

I'm closing this as it's probably going to be too difficult to salvage it, please open a new PR with correct changes, i.e. without moving the code into the header and without making unrelated whitespace-only changes. Please make sure to review the diff of your changes before opening the PR to check that it looks reasonable. Thanks in advance!

PR #25900, opened 2025-10-20
Ribbon review

Blake-Madden:
Various small refinements to ribbon code

### DIFF
--- a/src/ribbon/gallery.cpp
+++ b/src/ribbon/gallery.cpp
@@ -458,7 +458,7 @@ bool wxRibbonGallery::ScrollPixels(int pixels)
 
 void wxRibbonGallery::EnsureVisible(const wxRibbonGalleryItem* item)
 {
-    if(item == nullptr || !item->IsVisible() || IsEmpty())
+    if(item == nullptr || !item->IsVisible() || IsEmpty() || m_art == nullptr )
         return;
 
     if(m_art->GetFlags() & wxRIBBON_BAR_FLOW_VERTICAL)
@@ -518,8 +518,9 @@ void wxRibbonGallery::OnPaint(wxPaintEvent& WXUNUSED(evt))
         else
             offset_pos.SetLeft(offset_pos.GetLeft() - m_scroll_amount);
         m_art->DrawGalleryItemBackground(dc, this, offset_pos, item);
-        dc.DrawBitmap(item->GetBitmap(), offset_pos.GetLeft() + padding_left,
-            offset_pos.GetTop() + padding_top);
+        if (item->GetBitmap().IsOk() )
+            dc.DrawBitmap(item->GetBitmap(), offset_pos.GetLeft() + padding_left,
+                offset_pos.GetTop() + padding_top);
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/ribbon/gallery.cpp
```diff
@@ -543,7 +544,7 @@ wxRibbonGalleryItem* wxRibbonGallery::Append(const wxBitmap& bitmap, int id)
 
     wxRibbonGalleryItem *item = new wxRibbonGalleryItem;
     item->SetId(id);
-    item->SetBitmap(bitmap);
+    item->SetBitmap(bitmap.IsOk() ? bitmap : wxBitmap{});
```
paulcor: This doesn't do anything. "Not ok" is the same as "empty".

Blake-Madden: Rolled that back.

vadz (commented):
Thanks, most of the changes look good to me but I'm not sure about 62384e2da56a48c2d757094aff08e733d04ac12d: why should we use `wxDynamicCast()` here if we don't check its result anyhow? If the parent is always `wxRibbonBar`, as I think it should be, then `wxStaticCast()` is better if only because it's more clear about the fact that it is always supposed to succeed.

Blake-Madden:
> why should we use `wxDynamicCast()` here

`wxDynamicCast(GetParent(), wxRibbonBar)` is used every else in these files, but without the post-cast pointer check also. I would be happy to add pointer checks to all of these. I tend to be overly pedantic and dynamic cast would be my preference (with pointer check) to absolutely be sure things are what we think they are when casting and using them.

If you prefer static cast, I should change that elsewhere too to make things more consistent in these files.

vadz:
Using `wxStaticCast()` is preferable for the casts which must always succeed, so I'd prefer to do this. Alternatively, we could do something like this:

```cpp
    wxCHECK_RET( wxDynamicCast(ptr, T), "pointer of wrong type?" );
    ... just use static_cast<T*>(ptr) ...
```

src/ribbon/page.cpp
```diff
@@ -1275,7 +1277,8 @@ wxSize wxRibbonPage::DoGetBestSize() const
 
 void wxRibbonPage::HideIfExpanded()
 {
-    wxStaticCast(m_parent, wxRibbonBar)->HideIfExpanded();
+    wxCHECK_RET(wxDynamicCast(GetParent(), wxRibbonBar), "pointer of wrong type?");
+    static_cast<wxRibbonBar*>(GetParent())->HideIfExpanded();
```
vadz: This does the right thing now, thanks, but writing both casts like this seems very heavy, I wonder if we could improve this somehow. E.g. maybe add a function allowing to do this:

```cpp
if ( auto* const bar = wxCheckedStaticCast<wxRibbonBar>(GetParent()) )
    bar->HideIfExpanded();
```

? This looks very similar to the existing `wxCheckCast` but is not quite the same.

Blake-Madden: OK, applied suggestions. Let me know if my implementation of `wxCheckedStaticCast` should work differently.

vadz:

PR #25906, opened 2025-10-21
wxOSX: Enable threaded animation for smooth gauge style

willco007:
Enable threaded animation on macOS when using `wxGA_SMOOTH` style.  This allows the indeterminate progress bar to animate correctly while the main thread is blocked doing work.


vadz:
Nice, thanks! Are there any drawbacks of using this? I.e. do you have any idea why is it not on by default?

willco007:
No drawbacks that I'm aware of, I've never figured out why it's not enabled by default in the OS. Probably legacy from the Nextstep days. :)

vadz:
@csomor No objections if I merge this?

csomor:
> @csomor No objections if I merge this?

@vadz no, this looks fine to me, and thanks to @willco007

PR #25911, opened 2025-10-21
Move wxWebViewBackendChromium name into common webview header

MaartenBent:
This allows to check for Chromium back-end availability without needing `wxUSE_WEBVIEW_CHROMIUM` guards.


vadz:
Ah, yes, this makes more sense than what I did, thanks!

PR #25912, opened 2025-10-21
Fix crash in wxMSW wxToolInfo when compiling with UBSan

MaartenBent:
See https://forums.wxwidgets.org/viewtopic.php?p=229882#p229882

This is a simplified version of the patch from the topic above. @Brickmaster1 does this work for you?

### DIFF
--- a/src/msw/tooltip.cpp
+++ b/src/msw/tooltip.cpp
@@ -89,6 +89,13 @@ class wxToolInfo : public TOOLINFO
 {
 public:
     wxToolInfo(HWND hwndOwner, unsigned int id, const wxRect& rc)
+#if defined(__has_feature)
+    #if __has_feature(memory_sanitizer)
+        // https://forums.wxwidgets.org/viewtopic.php?p=229882#p229882
+        // Fix UBSan sanitizer causing crash in debug mode
+        __attribute__((no_sanitize("undefined")))
+    #endif
+#endif
     {
         // initialize all members
         wxZeroMemory(*this);


Brickmaster1:
Thanks for integrating the patch! I meant to get around to doing this myself, but I kind of forgot about it!

Brickmaster1:
I am booting my other computer with Windows to test it right now.

vadz:
Sorry, I didn't read the entire forum thread, but is there some explanation for this? I.e. does it work around a bug in our code correctly flagged by UBSAN (in which case I'd, of course, prefer to fix it) or is it really a bug in UBSAN (in which case I guess we don't have much choice, but it would be worth reporting it to UBSAN then)?

Brickmaster1:
I'm honestly not too sure which side causes the problem, I just knew that my patch was able to avoid it by disabling the UBSAN for the instruction.

vadz:
I really don't feel well about disabling UBSAN for this function, this shouldn't be necessary if it's compiled correctly. Rereading the forum thread, it looks like the real error is

```
thread 349144 panic: member access within null pointer of type 'TTTOOLINFOW' (aka 'tagTOOLINFOW')
```

at the assignment to `cbSize` in `wxToolInfo` ctor which seems to be impossible.

It looks like a problem with zig UBSAN implementation, if you can reproduce the problem with a simple struct inheriting from `TOOLINFO` (and I think you should be able to), you should submit this as bug to zig team and hopefully they can fix it, but it seems wrong to work around it in wx code.

vadz:
Maarten, unless you'd really like to merge this, I'd rather leave this out for the reasons explained above.

MaartenBent:
No problem, I'll close this PR then.

PR #25913, opened 2025-10-21
Remove duplicate headers from file list

MaartenBent:
When installing with CMake, I noticed two lines: `Up-to-date: include/wx/scopeguard.h` and `Up-to-date: include/wx/msw/init.h`.
These headers are installed two times, for base and for core. I removed them from core.


vadz:
Thanks for noticing!

PR #25920, opened 2025-10-24
FileDialog wildcard documentation fix and workaround note for macOS

julbean:
FileDialog wildcard documentation fix and bug note with workaround

### DIFF
--- a/interface/wx/filedlg.h
+++ b/interface/wx/filedlg.h
@@ -20,6 +20,11 @@ enum
 
 #define wxFD_DEFAULT_STYLE      wxFD_OPEN
 
+// set this system option to 1 in order to always show the filetypes popup in
+// file open dialogs if possible
+
+#define wxOSX_FILEDIALOG_ALWAYS_SHOW_TYPES wxT("osx.openfiledialog.always-show-types")
+
 /**
     Default wildcard string used in wxFileDialog corresponding to all files.
 
@@ -112,7 +117,7 @@ const char wxFileSelectorDefaultWildcardStr[];
     display of the filter choice set the corresponding wxSystemOptions before calling
     the file open dialog:
     @code
-         wxSystemOptions::SetOption('osx.openfiledialog.always-show-types', 1)
+         wxSystemOptions::SetOption(wxOSX_FILEDIALOG_ALWAYS_SHOW_TYPES, 1)
     @endcode
     But in contrast to Windows and Unix, where the file type choice filters only
     the selected files, on Mac macOS even in this case the dialog shows all files
--- a/interface/wx/filedlg.h
+++ b/interface/wx/filedlg.h
@@ -121,11 +121,8 @@ const char wxFileSelectorDefaultWildcardStr[];
     @endcode
     But in contrast to Windows and Unix, where the file type choice filters only
     the selected files, on Mac macOS even in this case the dialog shows all files
-    matching all file types. The files which does not match the currently selected
-    file type are greyed out and are not selectable. Also, currently the Mac macOS
-    implementation is bugged, causing all filters after the first to also have the
-    first filter applied, so ensure your first filter is a superset of all other
-    wildcard filters to avoid any issues.
+    matching all file types. The files which do not match the currently selected
+    file type are greyed out and are not selectable.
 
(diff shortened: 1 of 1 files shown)


interface/wx/filedlg.h
```diff
@@ -112,12 +112,15 @@ const char wxFileSelectorDefaultWildcardStr[];
     display of the filter choice set the corresponding wxSystemOptions before calling
     the file open dialog:
     @code
-         wxSystemOptions::SetOption(wxOSX_FILEDIALOG_ALWAYS_SHOW_TYPES, 1)
+         wxSystemOptions::SetOption('osx.openfiledialog.always-show-types', 1)
```
swt2c: I think the proper fix here (for the downstream wxPython issue [https://github.com/wxWidgets/Phoenix/issues/2368]) is that the `#define wxOSX_FILEDIALOG_ALWAYS_SHOW_TYPES` from `include/wx/osx/filedlg.h` should be replicated in this file.  That would make the constant available to wxPython.

vadz: If this fixes the problem, I'd prefer to do this. Using a symbolic constant is preferable to using a string for all the usual reasons.

interface/wx/filedlg.h
```diff
@@ -112,12 +112,15 @@ const char wxFileSelectorDefaultWildcardStr[];
...
     @endcode
     But in contrast to Windows and Unix, where the file type choice filters only
     the selected files, on Mac macOS even in this case the dialog shows all files
     matching all file types. The files which does not match the currently selected
-    file type are greyed out and are not selectable.
+    file type are greyed out and are not selectable. Also, currently the Mac macOS
+    implementation is bugged, causing all filters after the first to also have the
+    first filter applied, so ensure your first filter is a superset of all other
+    wildcard filters to avoid any issues.
```
vadz: Sorry, "is bugged" is just not the right way to say it. If I understand the issue correctly, I would rather say something like

```
@note Due to a macOS limitation, the first filter should be a catch-all one,
      i.e. use `*` as the wildcard, as it is always applied in addition to
      all the other filters, when another one is selected.
```

Of course, if it's possible to fix this, it would be even better. @csomor Do you know if this is really a macOS limitation or just wxOSX bug?

julbean: I only wrote "is bugged" because the example given above it with BMP, GIF, and PNG does not work, so it seemed unintended

julbean: here's where I documented that: https://github.com/wxWidgets/Phoenix/issues/2806

interface/wx/filedlg.h
julbean: Sounds good, I'm just a newbie when it comes to C code so didn't know about the better way.  Changed it to use the #define

interface/wx/filedlg.h
csomor: > Of course, if it's possible to fix this, it would be even better. @csomor Do you know if this is really a macOS limitation or just wxOSX bug?

I'll have to think things through again with the current version. The native expected behavior of a file open dialog would be to have any file that can be opened to be active, forcing the user to use a popup to select a different type was considered a poor experience. But I don't think we should enforce this if a file filter is shown. If it's not shown of course, then everything should be selectable that makes sense.

julbean:
Reverted the workaround note, as csomor has fixed the bug for future versions (although I haven't been able to test it to confirm).
Now this PR just fixes the ALWAYS_SHOW_TYPES issue (https://github.com/wxWidgets/Phoenix/issues/2368) and now also fixes a small typo.

interface/wx/filedlg.h
```diff
@@ -20,6 +20,11 @@ enum
 
 #define wxFD_DEFAULT_STYLE      wxFD_OPEN
 
+// set this system option to 1 in order to always show the filetypes popup in
```
vadz: Does this actually work, i.e. does this expose this symbol in wxPython? Or do we need to use `///` or `/**` comments for it to be picked up?

swt2c: Just adding the `#define` should be enough to make it available in wxPython.

PR #25923, opened 2025-10-24
wxGenericTreeCtrl: Reduce casts

vvs31415:


vadz:
Thanks, will merge soon, but I'll put the helper function in the source file too, there doesn't seem to be any reason to have it in the header.

PR #25924, opened 2025-10-25
add a wxGTK on macOS build action

csomor:
right now only manually triggered

### DIFF
--- a/.github/workflows/ci_mac_gtk.yml
+++ b/.github/workflows/ci_mac_gtk.yml
@@ -28,7 +28,7 @@ jobs:
         - name: wxGTK on macOS
           runner: macos-latest
           arch: arm64
-          configure_flags: --disable-uiactionsim --with-cxx=11 --without-liblzma --enable-debug --disable-sys-libs --with-gtk=3 --with-opengl=no --enable-detect_sm=no
+          configure_flags: --disable-uiactionsim --with-cxx=11 --enable-debug --disable-sys-libs --with-gtk=3 --with-opengl=no --enable-detect_sm=no
           skip_samples: true
           skip_testing: true
           allow_warnings: true
--- a/.github/workflows/ci_mac_gtk.yml
+++ b/.github/workflows/ci_mac_gtk.yml
@@ -1,0 +1,190 @@
+name: Mac with Gtk3 builds
+
+on:
+  # Allows you to run this workflow manually from the Actions tab, right now no other triggers until proven
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  
+jobs:
+  build-and-test:
+    defaults:
+      run:
+        shell: /usr/bin/arch -arch ${{ matrix.arch }} /bin/bash --noprofile --norc -eo pipefail {0}
+
+    runs-on: ${{ matrix.runner }}
+
+    name: ${{ matrix.name }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - name: wxGTK on macOS
+          runner: macos-latest
+          arch: arm64
+          configure_flags: --disable-uiactionsim --with-cxx=11 --without-liblzma --enable-debug --disable-sys-libs --with-gtk=3 --with-opengl=no --enable-detect_sm=no
+          skip_samples: true
+          skip_testing: true
+          allow_warnings: true
+        
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Set up Homebrew
+        id: set-up-homebrew
+        uses: Homebrew/actions/setup-homebrew@main
+      
+      - name: Install gtk+3
+        run: brew install gtk+3
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_mac_gtk.yml
```diff
@@ -0,0 +1,190 @@
...
+          export ASAN_OPTIONS='fast_unwind_on_malloc=0 detect_container_overflow=0'
+          # Exclude tests that are currently known to fail
+          wx_tests_selection='~[.] ~[valnum] ~wxPersistTLW ~wxPersistDVC ~wxTopLevel::Show ~wxFont::Weight ~wxFont::NativeFontInfo ~WebView ~RadioButton::Single ~RadioButton::Focus ~wxHtmlPrintout::Pagination ~wxTextCtrl::GetBestSize ~TextCtrlTestCase ~wxExecute::RedirectUTF8 ~Ellipsization::NormalCase ~SpinCtrl::* ~SpinCtrlDouble::* ~NotebookTestCase ~SettingsTestCase ~Window::Show ~ModalDialogsTestCase ~Button::Click ~Button::Disabled ~wxDVC::GetItemRect ~wxDVC::AppendTextColumn ~Grid::KeyboardSelection ~Grid::CellClick ~Grid::ReorderedColumnsCellClick ~Grid::CellSelect ~wxStyledTextCtrl::AutoComp ~EvtLoopTestCase ~EventPropagationTestCase ~wxTreeCtrl::ItemClick ~wxTreeCtrl::LabelEdit ~wxTreeCtrl::KeyDown ~wxTreeCtrl::CollapseExpandEvents ~wxTreeCtrl::SelectionChange ~wxTreeCtrl::SelectItemMultiInteractive ~wxTreeCtrl::Menu ~wxTreeCtrl::KeyNavigation'
+          ./test_gui.app/Contents/MacOS/test_gui $wx_tests_selection
+  
+      - name: Building samples
+        if: matrix.skip_testing != true && matrix.skip_samples != true
```
vadz: Probably should be
```suggestion
        if: matrix.skip_samples != true
```

.github/workflows/ci_mac_gtk.yml
```diff
@@ -0,0 +1,190 @@
...
+        run: |
+          make -k $wxBUILD_ARGS "CXXFLAGS=$wxMAKEFILE_CXXFLAGS" "LDFLAGS=$wxMAKEFILE_LDFLAGS" samples
+  
+      - name: Installing
+        if: matrix.skip_testing != true
```
vadz: I think we should always test installation and compiling `minimal` using the installed version, this doesn't take much time and really should work.

I.e. I'd remove this `if:` and the one below.

.github/workflows/ci_mac_gtk.yml
```diff
@@ -0,0 +1,190 @@
...
+          if [ -n "$rc" ]; then
+            httpbin_show_log
+            exit $rc
+          fi
+  
+      - name: Testing GUI
+        if: matrix.skip_testing != true
+        working-directory: tests
+        run: |
+          # We currently need to disable container overflow detection as we get
```
vadz, line 167: I'm not sure if all this is relevant for this build...

.github/workflows/ci_mac_gtk.yml
```diff
@@ -0,0 +1,190 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Set up Homebrew
+        id: set-up-homebrew
+        uses: Homebrew/actions/setup-homebrew@main
+      
+      - name: Install gtk+3
```
vadz: This is very nitpicky but they don't use "+" in their name any more, i.e.
```suggestion
      - name: Install GTK 3
```

.github/workflows/ci_mac_gtk.yml
```diff
@@ -0,0 +1,190 @@
...
+    name: ${{ matrix.name }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - name: wxGTK on macOS
+          runner: macos-latest
+          arch: arm64
+          configure_flags: --disable-uiactionsim --with-cxx=11 --without-liblzma --enable-debug --disable-sys-libs --with-gtk=3 --with-opengl=no --enable-detect_sm=no
```
vadz: Just out of curiosity, is OpenGL broken in this build?

Also, not sure why we need to disable LZMA, it will just be auto-disabled if support is not found?

vadz (approved):
Looks good, thanks!

.github/workflows/ci_mac_gtk.yml
csomor: I thought perhaps more functionality might be used once, so I've copied the steps from our other command line mac builds, but of course we could remove things

.github/workflows/ci_mac_gtk.yml
csomor: the LZMA was a left over from an earlier configure line of mine, I'll remove that. With OpenGL I ran into a build problem, apparently the homebrew version of gtk+ uses Quartz and not X11 anymore.

.github/workflows/ci_mac_gtk.yml
csomor: Thanks, did that

PR #25926, opened 2025-10-25
Make wxItemId::m_pItem private

vvs31415:
Call public wxItemId::GetID() instead

### DIFF
--- a/include/wx/treebase.h
+++ b/include/wx/treebase.h
@@ -82,9 +82,9 @@ class wxArrayTreeItemIds : public wxArrayTreeItemIdsBase
 {
 public:
     void Add(const wxTreeItemId& id)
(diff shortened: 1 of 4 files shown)


include/wx/itemid.h
```diff
@@ -40,9 +40,7 @@ class wxItemId
 
     void Unset() { m_pItem = nullptr; }
 
-    // This field is public *only* for compatibility with the old wxTreeItemId
```
vadz, line 43: We can apply the rest of this PR, but I don't want to change this, I'm almost certain that there is existing code using `m_pItem` and it's not worth breaking it without any real gain.

vadz:
Thanks, I'll merge and push this soon.

PR #25929, opened 2025-10-26
No longer deactivate wxDataViewCtrl, wxHtml and more on iOS.

RobertRoeb:
Remove the checks that deactivate these controls. They actually work now.


vadz:
Just in case you missed it, this broke a couple of CI builds.

RobertRoeb:
Wiill check. Didn't actually expect this after changes in the iPhone checks

vadz:
Merged, thanks.

PR #25930, opened 2025-10-26
Improve precise wxTreeCtrl best size calculation

### LINKED ISSUE #25839: Can't set width of `wxTreeCtrl` in `wxTreeBook`

vadz:
Try to compute the size of the items manually, even if GetBoundingRect() doesn't return it, as it happens when the items are collapsed.

This ensures the tree control best size is big enough to show the tree fully expanded.

Closes #25839.

Replaces #25843.

### DIFF
--- a/src/common/treebase.cpp
+++ b/src/common/treebase.cpp
@@ -24,6 +24,7 @@
 
 #include "wx/treectrl.h"
 #include "wx/imaglist.h"
+#include "wx/dcclient.h"
 
 extern WXDLLEXPORT_DATA(const char) wxTreeCtrlNameStr[] = "treeCtrl";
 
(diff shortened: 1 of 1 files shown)


ssrlive:
image

image

image

vadz:
Sorry but I'd really appreciate if you could reformulate your comment in a text form because I have no idea what do you mean by this image.

ssrlive:
Here are the results of my tests on three OSs using your PR code. Looks good.

vadz:
Merged in the commit above.

ssrlive:
After I deleted this code, it ran like the picture below. I think it is safe to delete this code.

https://github.com/wxWidgets/wxWidgets/blob/bd14cf2dbf5d32dfee63e4bd8821471bd1186f6f/src/common/treebase.cpp#L262-L265

image

vadz:
> After I deleted this code, it ran like the picture below. I think it is safe to delete this code.

When using 200% DPI it looks bad without this extra indent:

image

so I'd rather keep it.

PR #25933, opened 2025-10-28
Improve wxImageHandler constructor

vvs31415:


vadz:
Thanks, I'm applying this but after restoring default ctor in `wxImageHandler`: you can't make incompatible changes (the class had default ctor before but wouldn't have it any longer if this PR were applied as is) like this, i.e. without any good reason and documenting them. This is a public class, so changing all places where it is used in wx is not enough, as it can (and is) also be used in applications.

PR #917, opened 2018-09-03
BUG FIX - Windows menu potentially multiple times added to menu bar

hwiesmann:
As the windows menu was not known two wxWidgets' menu structure two problems arose:
 - the wrong search algorithm was used to find the windows menu (actually it could never be found);
 - user were not able to detect whether a windows menu existed or not.

Furthermore, as the windows menu was not cleared from menu items when reusing it OS X added multiple times the names of the existing windows to the menu.

Solving the first issue required a re-write of two methods the second problem was solved at the same time.

PR #25935, opened 2025-10-29
Moved the "Show Document Map" & "Mini Editor" menu

eranif:
on macOS, these items are placed in the Window menu and can not be accessed.

### DIFF
--- a/samples/stc/stctest.cpp
+++ b/samples/stc/stctest.cpp
@@ -540,6 +540,9 @@ void AppFrame::CreateMenu ()
     menuView->AppendCheckItem (myID_WHITESPACE, _("Show white&space"));
     menuView->AppendSeparator();
     menuView->Append (myID_USECHARSET, _("Use &code page of .."), menuCharset);
+    menuView->AppendSeparator();
+    menuView->Append(myID_WINDOW_MINIMAL, _("&Minimal editor"));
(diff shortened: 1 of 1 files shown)


vadz:
Wow, does this mean that you can't create menu called "Window" at all in portable applications? Or do I misunderstand why they can't be accessed?

Anyhow, I'll merge this soon, but this is news to me.

Also, does the minimap work for you on Mac? I haven't tested it there...

eranif:
I have a standard Apple's 'Window' menu. 
The minimap works well on macos

Will post a screenshot later

eranif:
This is how the "Window" menu looks without this PR on macOS:

Screenshot 2025-10-30 at 7 31 06

csomor:
As there is a standard window menu on macOS which is handled by the OS, I did try to detect the usage of a window menu in the app and merge things, I'll check what the current status of this is. Of course this can be turned off (`MacSetAutoWindow(false)`) but this is the last line and should not be necessary.

csomor:
originated here https://github.com/wxWidgets/wxWidgets/pull/917, I'll try to find a better approach and still not produce a regression.

vadz:
Stefan, should I apply this (it's not really a problem) or wait until this gets resolved due to your fixes to the "Window" menu handling?

csomor:
@vadz please apply it, so things are useable at least, thanks

PR #25938, opened 2025-10-31
Make ulong CTOR for wxColour explicit

Blake-Madden:
Helps avoid surprising conversions. Hopefully won't break backward compatibility for most, I just had to update one sample file in wx.

`wxColour(const wchar_t *colourName)` (and others) should probably remain non-explicit though for backward compatibility. Those would likely be the much more common CTORs being used.

What prompted this was me discovering that I could pass `true` to a function that expected a `wxColour` without complaint. This probably still won't fixed that because of `wxColour(const wchar_t *colourName)` being non-explicit, but doesn't seem like a bad idea to at least make this small change just the same.

### DIFF
--- a/include/wx/colour.h
+++ b/include/wx/colour.h
@@ -34,7 +34,7 @@ class WXDLLIMPEXP_FWD_CORE wxColour;
              ChannelType blue,                                                \
              ChannelType alpha = wxALPHA_OPAQUE)                              \
         { Init(); Set(red, green, blue, alpha); }                             \
-    explicit wxColour(unsigned long colRGB) { Init(); Set(colRGB    ); }      \
+    wxColour(unsigned long colRGB) { Init(); Set(colRGB    ); }               \
     wxColour(const wxString& colourName) { Init(); Set(colourName); }         \
     wxWXCOLOUR_CTOR_FROM_CHAR                                                 \
     wxColour(const wchar_t *colourName) { Init(); Set(colourName); }
--- a/include/wx/colour.h
+++ b/include/wx/colour.h
@@ -34,7 +34,7 @@ class WXDLLIMPEXP_FWD_CORE wxColour;
              ChannelType blue,                                                \
              ChannelType alpha = wxALPHA_OPAQUE)                              \
         { Init(); Set(red, green, blue, alpha); }                             \
-    wxColour(unsigned long colRGB) { Init(); Set(colRGB    ); }               \
+    explicit wxColour(unsigned long colRGB) { Init(); Set(colRGB    ); }      \
     wxColour(const wxString& colourName) { Init(); Set(colourName); }         \
     wxWXCOLOUR_CTOR_FROM_CHAR                                                 \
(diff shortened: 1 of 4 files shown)


include/wx/colour.h
```diff
@@ -34,7 +34,7 @@ class WXDLLIMPEXP_FWD_CORE wxColour;
              ChannelType blue,                                                \
              ChannelType alpha = wxALPHA_OPAQUE)                              \
         { Init(); Set(red, green, blue, alpha); }                             \
-    wxColour(unsigned long colRGB) { Init(); Set(colRGB    ); }               \
+    explicit wxColour(unsigned long colRGB) { Init(); Set(colRGB    ); }      \
```
lanurmi: Corresponding change in `interface/wx` missing.

Blake-Madden: done

vadz (commented):
This is a backwards incompatible change which risks breaking a lot of code as I'd expect passing hex BGR representation of a colour to be relatively common.

What about adding `wxColour(bool) = delete` instead?

Blake-Madden:
Unfortunately, MSVC can't seem to resolve between `bool` and integers very well. Deleting a `bool` CTOR results in this `wxColour(c.GetRGB())` being complained about as a missing function (it resolves this as a `bool` argument).

It's a weird edge case that I bumped into, I suppose we can just live with it if all agree with not wanting to use `explicit` here.

vadz:
We need to add overloads for `unsigned int` doing the same thing as `unsigned long` one. This should allow most code to keep working, but still break anything that uses `int` (or `long` or, but I think this is rather unlikely, any form of `short`). If we want to be maximally cautious we could provide overloads for all of `[unsigned] {int,long}` and delete the rest.

Blake-Madden:
I added this: `wxColour(unsigned int colRGB) { Init(); Set(colRGB    ); }`

but then this example `return wxColour(0x202020)` gets resolved as `int` and then MSVC falls back to the deleted `bool` CTOR and fails.

I just don't think it's going to work.

vadz:
It certainly will if we provide overload for `int` and we should indeed do it because I stupidly forgot that literals were ints, not unsigneds.

BTW, you don't have to duplicate the code, you can just delegate to the existing ctor.

Blake-Madden:
Got it, overloads and deleted `bool` CTOR are in there now.
I had to use casts so that the proper CTOR delegation took place, just FYI.

Blake-Madden:
I may be wrong, but I think this change uncovered an issue in "univ/themes/win32.cpp":

`dc.SetBrush(wxBrush(stipple_xpm));`

This must have been converting `stipple_xpm` into a `wxColour` before, and overloading the integer CTORs and deleting the bool CTOR in `wxColour` brought this to the surface. Anyway, I construct a `wxBitmap` explicitly from `stipple_xpm` now.

vadz:
Thanks, I've included this as part of #25943 which I hope to merge once I fix all the build errors, so closing this one.

PR #25939, opened 2025-10-31
Fix issue #25937

mcorino:


vadz:
Sorry, I didn't have to address this yet but I'd rather remove `const` from the documentation as I think the derived class may need to modify some of its members in their implementation of this function.

I'll push the trivial fix soon, thanks for reporting this.

PR #25940, opened 2025-10-31
Fix OnEditUpdate to Propagate wxEVT_STC_UPDATEUI Event

eranif:
Insert a call to `event.Skip()` at the start of `wxStyledTextCtrlMiniMap::OnEditUpdate` so the event continues its propagation.

Without this, subclass handlers are never get called...

* src/stc/minimap.cpp


vadz:
Oops, yes, of course, thanks for the fix, will merge soon.

PR #25942, opened 2025-10-31
Add Clang-Format Configuration File for wxWidgets Coding Style

eranif:
Introduce a .clang-format configuration file to enforce consistent code formatting across the project. This configuration is based on the wxWidgets coding style and will enable automatic code formatting with clang-format.

The configuration enforces the following key formatting rules:
- 4-space indentation (no tabs)
- Custom brace wrapping with braces on new lines for most constructs (classes, functions, control statements, etc.)
- 80-character line limit
- Spaces inside parentheses for conditional statements
- Left-aligned pointers and references
- No short-form single-line functions, if statements, or loops
- One parameter per line for function declarations

This will help maintain consistent code style across contributors and reduce manual formatting effort during code reviews. Developers can now use clang-format to automatically format their code according to project standards.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,89 @@
+---
+# clang-format configuration based on wxWidgets coding style
+Language: Cpp
+BasedOnStyle: LLVM
+
+# Indentation
+IndentWidth: 4
+TabWidth: 4
+UseTab: Never
+ContinuationIndentWidth: 4
+AccessModifierOffset: -4
+ConstructorInitializerIndentWidth: 4
+
+# Bracing
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterCaseLabel: true
+  AfterClass: true
+  AfterControlStatement: Always
+  AfterEnum: true
+  AfterFunction: true
+  AfterNamespace: true
+  AfterStruct: true
+  AfterUnion: true
+  AfterExternBlock: true
+  BeforeCatch: true
+  BeforeElse: true
+  BeforeWhile: false
+  IndentBraces: false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+
+# Function declaration/definition
+AlwaysBreakAfterReturnType: None
+AllowShortFunctionsOnASingleLine: None
+BinPackParameters: OnePerLine
+
+# Spacing
+SpaceBeforeParens: ControlStatements
+SpacesInParens: Custom
+SpacesInParensOptions:
+  ExceptDoubleParentheses: false
+  InConditionalStatements: true
+  InEmptyParentheses: false
+
+SpaceAfterCStyleCast: false
+SpaceBeforeAssignmentOperators: true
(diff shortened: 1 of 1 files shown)


vadz:
Sorry, I'm against it. We may discuss this more on wx-dev, but I have 2 problems:

1. We need to reformat all the existing code for this to be really useful and it will introduce an enormous amount of changes and break any still pending PRs/patches (hopefully few, but non zero) and complicate cherry picking fixes in 3.2.
2. Clang-format is, how to put it politely, not very smart. It often reformats the code, especially multiline statements, in a very ugly and unreadable way. It's certainly much worse than any other auto-formatter I've used for other languages :-(

We could probably live with (2), but (1) is just a too high price to pay IMO.

eranif:
Do you have an alternative tool? it's hard to follow wxWidgets coding style :), I saw `.editorconfig` in the root folder but it seems no relevant. What tool do you use? (if you use one at all...)

eranif:
I would like to contribute more, but I don't like iterating PRs because of coding style ...
For example in the recent `wxStyledTextCtrlMinMap` - I want to add a constructor that allows the caller to control the "mini-map" size, i.e. I would like to add another constructor (or modify the current one) so I can pass: pos, size, style etc). Adding such a long constructor got me thinking: do I need to place each param on a line of its own, or?

So having a tool (does not have to be clang-format) - would be great...

vadz:
I use directory-specific autocommands in Vim, but this doesn't take care of everything, just the basics. Mostly I just don't find it hard to write the code in the ambient style at all.

But I realize the value of the formatters. It's just that to really use them we need to set up a check that all new code follows the rules. And this requires reformatting all the existing code. And that, in turn, just seems a too high price to pay. FWIW all my new projects do use clang-format, for better or (mostly) worse. But the only time I've tried to switch an existing project (smaller and much younger) to it, it was a catastrophe — not because of clang-format itself but because of all the later problems it caused.

Weird idea of the day: any LLM should probably be able to "reformat the code in wxWidgets style" for you without any problems.

eranif:
> Weird idea of the day: any LLM should probably be able to "reformat the code in wxWidgets style" for you without any problems.

How do you think I created that `.clang-format` ? 🤣`Claude` did that for me ...

Pastes a chunk of wxWidgets code and asked it to generate `.clang-format` to match it...

But it is an interesting idea, will add it CodeLite ...

vadz:
> I would like to contribute more, but I don't like iterating PRs because of coding style ...

Certainly nothing prevents you from having this file in your fork, and setting up clang-format to run it on just the new code.

> I want to add a constructor that allows the caller to control the "mini-map" size, i.e. I would like to add another constructor (or modify the current one) so I can pass: pos, size, style etc). Adding such a long constructor got me thinking: do I need to place each param on a line of its own, or?

Basically just looking at any existing `wxWindow`-derived class ctor shows that the answer is "yes".

vadz:
I don't think we're going to merge this, so closing, but it could be still nice to make this available somewhere if anybody else finds it useful.

PR #25945, opened 2025-11-01
CMake: Improve creating wayland protocol files

MaartenBent:
Create the files in a temporary directory and only copy them to the final directory if the files are different.
This prevents the source files from being rebuilt every time cmake is run and the protocol files are replaced.

Exclude the protocols folder from being installed.

The entire setup header directory is installed, so exclude `protocols`.
see e.g. [this](https://github.com/wxWidgets/wxWidgets/actions/runs/18975885071/job/54195276677#step:10:956) build, and copied below for reference:
```
-- Installing: /usr/local/lib/wx/include/gtk3-unicode-3.3
-- Installing: /usr/local/lib/wx/include/gtk3-unicode-3.3/wx
-- Installing: /usr/local/lib/wx/include/gtk3-unicode-3.3/wx/setup.h
-- Installing: /usr/local/lib/wx/include/gtk3-unicode-3.3/wx/protocols
-- Installing: /usr/local/lib/wx/include/gtk3-unicode-3.3/wx/protocols/pointer-warp-v1-client-protocol.h
-- Installing: /usr/local/lib/wx/include/gtk3-unicode-3.3/wx/protocols/pointer-warp-v1-client-protocol.c
```

### DIFF
--- a/build/cmake/init.cmake
+++ b/build/cmake/init.cmake
@@ -464,6 +464,7 @@ if(wxUSE_GUI)
                     pkg_get_variable(WAYLAND_SCANNER wayland-scanner wayland_scanner)
                     if(WAYLAND_SCANNER)
                         set(wx_protocols_input_dir ${wxSOURCE_DIR}/src/unix/protocols)
+                        set(wx_protocols_temp_dir ${wxOUTPUT_DIR}/wx/protocols)
                         set(wx_protocols_output_dir ${wxSETUP_HEADER_PATH}/wx/protocols)
 
                         # Note that we need multiple execute_process()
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, I didn't think about it, but it was obviously wrong to do what I did, so thank you for cleaning it up.

Will merge soon.

MaartenBent:
I actually asked about installing them in https://github.com/wxWidgets/wxWidgets/pull/25769#issuecomment-3264116282, but assumed as well they wouldn't be installed automatically.

Seems I copy-pasted wrong, let me fix it...

PR #25949, opened 2025-11-02
Use current, not best, size for controls in wxGTK toolbars

vadz:
The changes of 8f9f1a42ec (Respect best control size when adding it to wxGTK wxToolBar, 2025-03-02) did in wxGTK what earlier b0ad9ccffd (Use control current, not best, size in wxMSW wxToolBar layout code, 2019-03-31) undid in wxMSW, so tweak them to work as in wxMSW and use the current control size if it's specified instead of the best size to make sure the control doesn't take too much space in the toolbar.

This makes appearance of the toolbar containing wxChoice in the toolbar sample much more reasonable.

-----

Not sure if anybody has any comments about this, but it doesn't seem right to have very different behaviour in wxMSW and wxGTK and the (current) wxMSW one seems more reasonable.


vadz:
Applied in the commit above.

PR #25961, opened 2025-11-07
`wxListCtrl` fixes in wxMSW

vadz:
This is #25950 with minor changes and split into several commits.

### DIFF
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -1382,11 +1382,23 @@ bool wxListCtrl::GetSubItemRect(long item, long subItem, wxRect& rect, int code)
 
     wxCopyRECTToRect(rectWin, rect);
 
-    // there is no way to retrieve the first sub item bounding rectangle using
-    // wxGetListCtrlSubItemRect() as 0 means the whole item, so we need to
-    // truncate it at first column ourselves
+    // We can't use wxGetListCtrlSubItemRect() for the 0th subitem as 0 means
+    // the entire row for this function, so we need to calculate it ourselves.
     if ( subItem == 0 && code == wxLIST_RECT_BOUNDS )
+    {
+        // Because the columns can be reordered, we need to sum the widths of
+        // all preceding columns to get the correct x position.
+        rect.x = 0;
+        for ( auto col : GetColumnsOrder() )
+        {
+            if ( col == 0 )
+                break;
+
+            rect.x += GetColumnWidth(col);
+        }
+
         rect.width = GetColumnWidth(0);
+    }
 
     return true;
 }
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3047,9 +3047,21 @@ RECT GetCustomDrawnItemRect(const NMCUSTOMDRAW& nmcd)
     return rc;
 }
 
-constexpr int GAP_BETWEEN_CHECKBOX_AND_TEXT = 2;
+// These values try to replicate the native listctrl rendering as close as
+// possible. Notice that they are not scaled with DPI because the native
+// control doesn't do it either.
+constexpr int PADDING_LEFT_SIDE = 1;
+constexpr int PADDING_RIGHT_SIDE = 2;
+constexpr int GAP_AFTER_CHECKBOX = 2;
+constexpr int GAP_BEFORE_IMAGE = 2;
+constexpr int GAP_BEFORE_CHECKBOX = 4;
+constexpr int PADDING_LEFT_FOR_TEXT = 1;
 
-bool HandleSubItemPrepaint(wxListCtrl* listctrl, LPNMLVCUSTOMDRAW pLVCD, HFONT hfont, int colCount)
+bool
+HandleSubItemPrepaint(wxListCtrl* listctrl,
+                      LPNMLVCUSTOMDRAW pLVCD,
+                      HFONT hfont,
+                      RECT rc)
 {
     NMCUSTOMDRAW& nmcd = pLVCD->nmcd;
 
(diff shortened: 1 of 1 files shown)


src/msw/listctrl.cpp
```diff
@@ -3218,32 +3288,45 @@ void HandleItemPaint(wxListCtrl* listctrl, LPNMLVCUSTOMDRAW pLVCD, HFONT hfont)
...
     {
         pLVCD->iSubItem = col;
-        HandleSubItemPrepaint(listctrl, pLVCD, hfont, colCount);
+
+        // get the rectangle of this entire subitem
+        wxRect rectSubItem;
+        if ( !listctrl->GetSubItemRect(item, col, rectSubItem, wxLIST_RECT_BOUNDS) )
+            continue;
+        RECT rcSubItem;
+        wxCopyRectToRECT(rectSubItem, rcSubItem);
```
MaartenBent, line 3341: Can get the `RECT` directly like:
```suggestion
        RECT rcSubItem;
        if ( !wxGetListCtrlSubItemRect(nmcd.hdr.hwndFrom, item, col, LVIR_BOUNDS, rcSubItem) )
            continue;
```

src/msw/listctrl.cpp
```diff
@@ -3218,32 +3288,45 @@ void HandleItemPaint(wxListCtrl* listctrl, LPNMLVCUSTOMDRAW pLVCD, HFONT hfont)
...
+
+        HandleSubItemPrepaint(listctrl, pLVCD, hfont, rcSubItem, pLVCD->clrTextBk);
```
MaartenBent: You already pass `pLVCD` to the function, so `pLVCD->clrTextBk` is redundant.

src/msw/listctrl.cpp
Crementif: ListCtrl::GetSubItemRect() does additional math to normalize the output of wxGetListCtrlSubItemRect, since its fully incorrect.

MaartenBent: Ignore my suggestion, it won't work with the fixes in `GetSubItemRect` about column order.

MaartenBent: > ListCtrl::GetSubItemRect() does additional math to normalize the output of wxGetListCtrlSubItemRect, since its fully incorrect.

Sorry missed your message. Indeed, I also noticed wrong x-offset when reordering columns.

src/msw/listctrl.cpp
vadz: Thanks, I agree that it's simpler to remove this parameter, force-pushed now.

vadz:
@Crementif If you don't have any objections, I'll merge this soon.

MaartenBent:
I have one more suggestion. In https://github.com/wxWidgets/wxWidgets/pull/25950 I mentioned the right-aligned text padding. What about fixing it like this.
Result: https://imgsli.com/NDI3NDY3
```diff
diff --git "a/src/msw/listctrl.cpp" "b/src/msw/listctrl.cpp"
index b800fdb9b2e..71500d78f85 100644
--- "a/src/msw/listctrl.cpp"
+++ "b/src/msw/listctrl.cpp"
@@ -3055,7 +3055,7 @@ constexpr int PADDING_RIGHT_SIDE = 2;
 constexpr int GAP_AFTER_CHECKBOX = 2;
 constexpr int GAP_BEFORE_IMAGE = 2;
 constexpr int GAP_BEFORE_CHECKBOX = 4;
-constexpr int PADDING_LEFT_FOR_TEXT = 1;
+constexpr int PADDING_FOR_TEXT = 1;
 
 bool
 HandleSubItemPrepaint(wxListCtrl* listctrl,
@@ -3204,6 +3204,7 @@ HandleSubItemPrepaint(wxListCtrl* listctrl,
         {
             case LVCFMT_LEFT:
                 fmt |= DT_LEFT;
+                rc.left += PADDING_FOR_TEXT;
                 break;
 
             case LVCFMT_CENTER:
@@ -3212,13 +3213,12 @@ HandleSubItemPrepaint(wxListCtrl* listctrl,
 
             case LVCFMT_RIGHT:
                 fmt |= DT_RIGHT;
+                rc.right -= PADDING_FOR_TEXT;
                 break;
         }
     }
     //else: failed to get alignment, assume it's DT_LEFT (default)
 
-    rc.left += PADDING_LEFT_FOR_TEXT;
-
     if ( rc.left < rc.right )
         DrawText(hdc, text, -1, &rc, fmt);

```

vadz:
Oops, sorry, I had missed that. Pushed too now.

src/msw/listctrl.cpp
```diff
@@ -3050,41 +3074,52 @@ bool HandleSubItemPrepaint(wxListCtrl* listctrl, LPNMLVCUSTOMDRAW pLVCD, HFONT h
...
-        rc.right = rc2.left;
-    }
+    // there's some padding on both sides of the column item
+    rc.left += PADDING_LEFT_SIDE;
+    rc.right -= PADDING_RIGHT_SIDE;
+
+    const int availableWidth = wxMax(rc.right - rc.left, 0);
 
-    if ( !col && listctrl->HasCheckBoxes() )
+    if ( !col && listctrl->HasCheckBoxes() && availableWidth > 0 )
```
Crementif: I guess it wasn't very clear in my PR, but a large part of what I was also intending to fix (for Cemu) was that columns would still correctly (invisibly) render when their width is set to 0.

I think this code would fail when the availableWidth is set to a value between 1-3. After GAP_BEFORE_CHECKBOX (4 pixels of padding) gets added, the ImageList_DrawEx should really not be called at all, since there's no column width left.

I personally think its better to move this around the ImageList_DrawEx check, like I had.

src/msw/listctrl.cpp
```diff
@@ -3050,41 +3074,52 @@ bool HandleSubItemPrepaint(wxListCtrl* listctrl, LPNMLVCUSTOMDRAW pLVCD, HFONT h
...
+            int cbWidth, cbHeight;
             ImageList_GetIconSize(himl, &cbWidth, &cbHeight);
-            cbX = listctrl->FromDIP(GAP_BETWEEN_CHECKBOX_AND_TEXT);
-            cbY = rc.top + ((rc.bottom - rc.top) / 2 - cbHeight / 2);
+
+            // center checkbox vertically
+            int cbY = rc.top + ((rc.bottom - rc.top) / 2 - cbHeight / 2);
+
+            // prevent drawing checkbox farther than the column width
+            cbWidth = wxClip(cbWidth, 0, availableWidth);
```
Crementif: availableWidth isn't the current remaining space that's available in the column, since we've added GAP_BEFORE_CHECKBOX to it already. The checkbox is now able to overflow the next column by 4 pixels.

src/msw/listctrl.cpp
```diff
@@ -3137,6 +3203,7 @@ bool HandleSubItemPrepaint(wxListCtrl* listctrl, LPNMLVCUSTOMDRAW pLVCD, HFONT h
         {
             case LVCFMT_LEFT:
                 fmt |= DT_LEFT;
+                rc.left += PADDING_FOR_TEXT;
```
Crementif, line 3221: I think `PADDING_FOR_TEXT` isn't very clear by itself. I think re-adding the comment about the padding being used as the inner padding for an individual row-item's background color would be useful for someone looking to understand all of these magic offsets.

Might just be a bit of a nitpick, so you can also leave it if that's clearer for others/you.

Crementif (changes requested):

src/msw/listctrl.cpp
vadz: Sorry, I've indeed missed that. Why is it important to still render them when the width is 0? I thought it would be better to just skip this in this case, what do we gain by still doing it?

src/msw/listctrl.cpp
vadz: I think it's relatively clear as it's only used in this one place. I did remove

```cpp
// text is drawn 1 unit away from the background rectangle
```

comment intentionally because it seemed to just repeat what the code was quite clearly doing.

OTOH looking at this code again, I think we should apply padding (at both sides) for centered text too, shouldn't we?

src/msw/listctrl.cpp
vadz: Oops, thanks, I've indeed missed it. But how to fix this depends on the fix for the comment above, i.e. I would just replace `availableWidth > 0` with `availableWidth > GAP_BEFORE_CHECKBOX` and use `availableWidth - GAP_BEFORE_CHECKBOX` here, but not if we remove that check completely...

src/msw/listctrl.cpp
Crementif: I meant the opposite, sorry. Ultimately it should follow the native white mode behavior, which is that it always cuts the content off correctly. The bug still stands however, at least from what I can see.

src/msw/listctrl.cpp
Crementif: Hard to say without testing, but maybe?

src/msw/listctrl.cpp
vadz: I might be missing something (it's getting late here...), but I don't see any obvious difference in behaviour between light and dark modes with the current version, the checkbox does get clipped when the column width becomes very small. Maybe it's because you've also changed code to overdraw the columns by the following ones?

Anyhow, what about this diff?

```diff
diff --git a/src/msw/listctrl.cpp b/src/msw/listctrl.cpp
index e77b4450d5..9a219ba11c 100644
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3079,8 +3079,9 @@ HandleSubItemPrepaint(wxListCtrl* listctrl,
     rc.right -= PADDING_RIGHT_SIDE;
 
     const int availableWidth = wxMax(rc.right - rc.left, 0);
+    const int availableForCheckbox = availableWidth - GAP_AFTER_CHECKBOX;
 
-    if ( !col && listctrl->HasCheckBoxes() && availableWidth > 0 )
+    if ( !col && listctrl->HasCheckBoxes() && availableForCheckbox > 0 )
     {
         const HIMAGELIST himl = ListView_GetImageList(hwndList, LVSIL_STATE);
 
@@ -3095,7 +3096,7 @@ HandleSubItemPrepaint(wxListCtrl* listctrl,
             int cbY = rc.top + ((rc.bottom - rc.top) / 2 - cbHeight / 2);
 
             // prevent drawing checkbox farther than the column width
-            cbWidth = wxClip(cbWidth, 0, availableWidth);
+            cbWidth = wxClip(cbWidth, 0, availableForCheckbox);
 
             // When using style flag ILD_SELECTED or ILD_FOCUS, the checkboxes
             // for selected items are drawn with a blue background, which we want to avoid.
```

I don't see any visual differences but it does seem to be more logical to do it like this.

src/msw/listctrl.cpp
Crementif: The latter solution would introduce a very minor, but present behavior where it'd draw an image or the text if it can't fit GAP_BEFORE_CHECKBOX. I am not sure if this is what the white mode does, but if I had to guess it doesn't do any "if X doesn't fit, render Y" tricks and it would be inaccurate to do that.

You could still reduce the amount of code it has to run if you split it up into 2 if statements, and I think it'd be a lot more accurate. But I am speculating.

vadz: What about the latest version? I've removed `availableWidth` to get a bit closer back to the original version, but I've also added returns to make sure we don't try drawing anything which is certain not to fit.

src/msw/listctrl.cpp
Crementif: I think the latest version looks good to be merged.

Issue #25633, opened 2025-07-12
wxMSW:darkmode - wxTreeCtrl, wxDataViewCtrl and wxListCtrl


herb52:
Hello,

First of all: Thanks for the darkmode feature in wxMSW.

I have updated my Windows 10 project from wxWidgets 3.2.1 to 3.3.0 and
I started some tests with darkmode.

For my used controls I have seen the following in darkmode:

    wxTreeCtrl
    for selected and focused item it uses a light gray
    for selected but unfocused item it uses a dark gray
    which is nearly identcal to background

    wxDataViewCtrl
    for selected and focused item it uses a dark blue
    for selected and unfocused item it uses black
    which is also nearly identcal to used background

    wxListCtrl
    for selected item it uses blue

I also have retested this behaviour with the released samples for dataview, treeview and listview.
The only modification is: MSWEnableDarkMode(wxApp::DarkMode_Always, nullptr); inside OnInit()

Now I was asked whether it is possible
- to enhance the visibility of selected unfocused items and
- to harmonize the used colors

This leads to question:
Is it possible to change/modify ONLY these used colors?
If yes, any hint where is appreciated.

I know that wxWidgets uses native Windows controls when possible, so maybe some colors cannot be changed.
But I do not have an overview.
I think wxDataViewCtrl is not a native control.

Thanks for your answers/comments in advance
Best regards
herb

PBfordev:
Back then when I was trying to fix some `wxListCtrl` issues, I thought that we need to introduce some pseudo-system named colors for Win32 dark mode, e.g., to `wxDarkModeSettings` or elsewhere. The system-provided ones may not be enough these days.

Now we may have them spread hardcoded all over the code base. I originally meant colors for grid lines (`wxListView` or `wxGrid`),  but of course as in this Issue for various states of items in different controls too...

vadz:
As you notice, it should be possible to improve things for wxDVC as it's entirely under our control. Doing it for `wx{Tree,List}Ctrl` risks being trickier, especially for the list one, as I spent a lot of time on it and the current look was the best I could find, even if it has known problems, see this comment:

https://github.com/wxWidgets/wxWidgets/blob/09f433faf39aab3f25b3c564b82448bb845fae56/src/msw/listctrl.cpp#L611-L621

Of course, any improvements would be welcome.

Takasa:
First, fantastic work on darkmode, it's great!

[...]

PR #25962, opened 2025-11-09
Use correct column attributes when drawing wxListCtrl

MaartenBent:
When a column is custom drawn, it uses the `wxItemAttr` of the column being updated to redraw all columns. Fix this by requesting the column attributes of each column that is drawn.

See #25633 (not sure if you can close this, because it is about more than `wxListCtrl`)

cc @Crementif This shouldn't affect your fixes from #25950 but I'd appreciate if you could review this to make sure all still works fine. Thank you.

You can reproduce the problem with the following patch. Open the listctrl sample in dark mode, and select the virtual view (f7). Without this PR, all columns have the same colour, with this PR columns have the correct different colour.

```diff
 samples/listctrl/listtest.cpp | 12 ++++++++++++
 samples/listctrl/listtest.h   |  1 +
 2 files changed, 13 insertions(+)

diff --git a/samples/listctrl/listtest.cpp b/samples/listctrl/listtest.cpp
index 6cb49bcf871..48dac3cef12 100644
--- a/samples/listctrl/listtest.cpp
+++ b/samples/listctrl/listtest.cpp
@@ -1536,6 +1536,18 @@ int MyListCtrl::OnGetItemColumnImage(long item, long column) const
     return -1;
 }
 
+wxItemAttr* MyListCtrl::OnGetItemColumnAttr(long item, long column) const
+{
+    static wxItemAttr attr(*wxRED, wxNullColour, wxNullFont);
+    if (column == 0)
+        attr.SetTextColour(wxColour(0xFF, 0x00, 0x00));
+    if (column == 1)
+        attr.SetTextColour(wxColour(0x00, 0x00, 0xFF));
+    if ((item + column) % 5 == 0)
+        attr.SetTextColour(wxColour(0x00, 0xFF, 0x00));
+    return &attr;
+}
+
 wxItemAttr *MyListCtrl::OnGetItemAttr(long item) const
 {
     // test to check that RefreshItem() works correctly: when m_updated is
diff --git a/samples/listctrl/listtest.h b/samples/listctrl/listtest.h
index ffa4190548a..2b1e8369560 100644
--- a/samples/listctrl/listtest.h
+++ b/samples/listctrl/listtest.h
@@ -84,6 +84,7 @@ private:
     virtual wxString OnGetItemText(long item, long column) const override;
     virtual bool OnGetItemIsChecked(long item) const override;
     virtual int OnGetItemColumnImage(long item, long column) const override;
+    virtual wxItemAttr *OnGetItemColumnAttr(long item, long column) const override;
     virtual wxItemAttr *OnGetItemAttr(long item) const override;
 
     long m_updated;
```

### DIFF
--- a/include/wx/msw/listctrl.h
+++ b/include/wx/msw/listctrl.h
@@ -379,6 +379,10 @@ public:
                                     WXWPARAM wParam,
                                     WXLPARAM lParam) override;
 
+    // public function to get ItemColumnAttr
+    wxItemAttr* MSWGetItemColumnAttr(long item, long column) const
+        { return DoGetItemColumnAttr(item, column); }
+
 protected:
     // common part of all ctors
     void Init();
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3258,9 +3258,7 @@ void HandleItemPostpaint(NMCUSTOMDRAW nmcd)
 // also called when using dark mode as we have to draw the selected item
 // ourselves when using it, and if we do this, we have to paint all the items
 // for consistency.
-//
-// pLVCD->clrText and clrTextBk should contain the colours to use
-void HandleItemPaint(wxListCtrl* listctrl, LPNMLVCUSTOMDRAW pLVCD, HFONT hfont)
+void HandleItemPaint(wxListCtrl* listctrl, LPNMLVCUSTOMDRAW pLVCD)
 {
     NMCUSTOMDRAW& nmcd = pLVCD->nmcd; // just a shortcut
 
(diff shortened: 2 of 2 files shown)


vadz (commented):
Thanks for fixing this, and it's a minor detail, but could we perhaps pass a `vector<wxItemAttr*>` instead of a function/lambda? It seems weird to use the latter for this, if using vector is difficult/undesirable, I'd rather consider making these functions member functions (or maybe just renaming `DoGetItemColumnAttr()` to `MSWGetItemColumnAttr()` and making it public so that they could call it).

MaartenBent:
I tried using `vector` at first, but it doesn't work with my example, where `wxItemAttr` is static. Because all pointers reference the same object and will have the values of the last called `GetItemColumnAttr`.

I'll try the `MSWGetItemColumnAttr` suggestion.

Crementif:
Sure, I can test whether it breaks anything once you've tried the suggestion that vadz mentioned.
Reading it, I think these changes shouldn't affect anything, as long as the background color clear checks whether the row/item is selected.

MaartenBent:
> I think these changes shouldn't affect anything, as long as the background color clear checks whether the row/item is selected

Thanks, I reverted my change, the selection colour was indeed wrong behind images.

PR #25965, opened 2025-11-09
Restore ANSI win32 API functions

MaartenBent:
wxWidgets is always build in `UNICODE` using the `WIDE` win32 API functions. But applications using it might be built without `UNICODE`, using the `ANSI` win32 functions.

Right now wxWidgets `#undef`s some of the win32 API functions and defines them as inline functions using the `WIDE` win32 API. Making these unavailable for applications that want to use the `ANSI` versions later.

Fix this by inlining the functions as `WIDE` only when `UNICODE` was already defined, not when wxWidgets defines it themselves in `platform.h`.

See https://github.com/wxWidgets/wxWidgets/issues/25803

cc @jamarr81  would this work?


include/wx/msw/winundef.h
```diff
@@ -15,30 +15,52 @@
...
-                             LPCTSTR pTemplate,
+                             wxLPCTSTR pTemplate,
                              HWND hwndParent,
                              DLGPROC pDlgProc)
     {
+    #if wxUSE_UNICODE_WINDOWS_H
         return CreateDialogW(hInstance, pTemplate, hwndParent, pDlgProc);
+    #else
+        return CreateDialogA(hInstance, pTemplate, hwndParent, pDlgProc);
+    #endif
```
vadz: Could we just define our own `wxFUNC_AW()` macro once, depending on `wxUSE_UNICODE_WINDOWS_H` value, and then replace all these fragments with just

```suggestion
        return wxFUNC_AW(CreateDialog)(hInstance, pTemplate, hwndParent, pDlgProc);
```

?

include/wx/msw/winundef.h
```diff
@@ -15,30 +15,52 @@
...
+#if defined(UNICODE) && (defined(WXBUILDING) || !defined(wxUNICODE_SET_BY_WXWIDGETS))
+    // inline the WIDE Win32 API functions when building wxWidgets or when
+    // UNICODE is not defined by wxWidgets (i.e. the user has defined UNICODE).
+    #define wxUSE_UNICODE_WINDOWS_H 1
+
+    #define wxLPCTSTR LPCWSTR
+    #define wxLPSTR LPWSTR
+    #define wxLPWNDCLASS LPWNDCLASSW
+    #define wxDOCINFO DOCINFOW
+    #define wxLPFINDREPLACE LPFINDREPLACEW
```
vadz: Minor, but these could be `typedef`s (or even `using`) if we try to minimize the use of preprocessor (possibly not really helpful in this header...).

include/wx/msw/winundef.h
```diff
@@ -15,30 +15,52 @@
...
+#else
+    // inline the ANSI Win32 API functions when UNICODE was defined by
+    // wxWidgets (i.e. the user has not defined UNICODE).
```
vadz: I'd add "and we're included from the application code and not wx itself" for extra clarity.

vadz (commented):
Thanks, I think this should work and I can't think of anything else that this would break, but I'd prefer to condense all the wrapper definitions using a helper macro before merging this.

include/wx/msw/winundef.h
MaartenBent: `typedef`/`using` doesn't work when these types are not defined. Apparent in the non-PCH builds.

This can happen for example when `winundef.h` is included but `windows.h` is not, like in `wrapcdlg.h` or `ioswrap.h` or `iosfwrap.h`.

MaartenBent:
Not sure how to fix these last errors in CI. In this case `UNICODE` is not defined by the build system / user. So wxWidgets defines is. Then `windows.h` is included (with `UNICODE` defined), resulting in `TEXT` and other defines/macros becoming the `WIDE` variant.

But later in the application code, after `winundef.h`, the ANSI variants will be used. But `TEXT` is still the `WIDE` variant.

I think I'll just let it default to the `WIDE` variant, unless the application defines `wxWIN32_UNDEF_AS_ANSI` before including wx headers:

```c++
#if defined(WXBUILDING) || !defined(wxWIN32_UNDEF_AS_ANSI)
   // wide
#else
   // ansi
#endif
```

vadz:
We could define something (`_UNICODE`?) in the sample to work around this problem there but it seems indeed safer and better to require predefining some symbol if you really need `A` versions of the API. But arguably, `wxNO_WIN32_WIDE_FUNCTIONS` (or just `wxNO_WIN32_W`) would be more in line with our preprocessor symbol naming conventions.

jamarr81:
@MaartenBent @vadz tysm for following up on this! I haven't had a chance to revisit it; however, these changes look good to me on paper.

include/wx/msw/private.h
```diff
@@ -847,13 +847,13 @@ class ClassRegistrar
     bool IsRegistered() const { return m_registered == 1; }
 
     // try to register the class if not done yet, return true on success
-    bool Register(const WNDCLASS& wc)
+    bool Register(const WNDCLASSW& wc)
```
vadz, line 850: It would be good to add a test for compiling this file with `wxNO_WIN32_W` to at least one MSW CI job to catch any errors that could be introduced into it later by forgetting to explicitly use `W` symbol versions.

vadz (approved):
Looks good, thanks!

It would be great if this could be tested in some application which needed this, but if nobody does it soon, I'll just merge this and hope that it works for them.

MaartenBent:
> It would be great if this could be tested in some application which needed this, but if nobody does it soon, I'll just merge this and hope that it works for them.

I tested it in the widgets sample, by adding this to the start of `native.cpp`. But I don't think this is the correct place to test it. 

I can try to add a new test file, similar to `native.cpp`, that starts with this. And build/run it with and without `UNICODE`.

```c++
#if WIN32
    #define _WINSOCK_DEPRECATED_NO_WARNINGS
    #define _CRT_SECURE_NO_WARNINGS
    #include <winsock2.h>
    #include <Windows.h>
    #ifndef UNICODE
        #define wxNO_WIN32_W
    #endif
#endif

#include "wx/wxprec.h"
```

vadz:
I think it could be acceptable to do this in the existing `native.cpp` if we added a comment explaining that this is just for testing, but it would be nicer to do it in some other place. Maybe as part of the (GUI) test suite? We just need a test case calling some harmless Win32 function existing in both A and W versions, e.g. `GetWindowText()`.

include/wx/msw/private.h
MaartenBent: This is now caught by the newly added test.
I also include `allheaders.h` to make sure no other headers use the wrong symbols/functions in the future.

MaartenBent:
I added a separate test, building with and without `UNICODE`, that includes and uses `Windows.h`.

PR #25967, opened 2025-11-11
Fix wrong origin offset when using wxRenderer with wxGCDC

vadz:
ConvertToRECT() in wxRenderer code accounted for the origin shift twice when using wxGCDC since the changes of 4f9186f1a1 (Increase usable scrolling range in wxMSW by a factor of 10,000, 2022-04-30) which added Offset(dc.GetDeviceOrigin()) call to it as for wxGCDC this offset is already accounted for by MSWApplyGDIPlusTransform().

Fix this by replacing MSWApplyGDIPlusTransform() with a more generic MSWApplyWXTransform() which does the same thing as before for wxGCDC but applies the origin offset for other wxDCs. This ensures that this offset is properly accounted for in both cases.

Closes #25966.

### DIFF
--- a/include/wx/dc.h
+++ b/include/wx/dc.h
@@ -483,14 +483,11 @@ public:
 
 #ifdef __WXMSW__
     // Native Windows functions using the underlying HDC don't honour GDI+
-    // transformations which may be applied to it. Using this function we can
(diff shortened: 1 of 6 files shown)


vadz:
Merged in the commit above.

Issue #25969, opened 2025-11-12
wxQt fails to build on Arch Linux with Qt 6.10 and Clang 21.1.5


generic-pers0n:
### Build System Used

I build wxWidgets and/or my application using:

- [x] CMake
- [ ] configure
- [ ] mingw32-make with makefile.gcc
- [ ] MSBuild (Microsoft Visual Studio solution file)
- [ ] nmake with makefile.vc
- [ ] Xcode

The full configure command I used is `cmake -G Ninja -DwxBUILD_TOOLKIT=qt ../wxwidgets`.

### Description
On Arch Linux with Qt 6.10 and Clang 21.1.5, building the wxQt port fails due to no known conversion for `Qt::Key` in `QChar`'s constructor. Here's the full output below:

```
[1/25] Building CXX object libs/core/CMakeFiles/wxcore.dir/__/__/__/__/src/qt/uiaction.cpp.o
FAILED: [code=1] libs/core/CMakeFiles/wxcore.dir/__/__/__/__/src/qt/uiaction.cpp.o 
[...]

PR #25970, opened 2025-11-12
Fix wxQt build on Arch Linux under Clang 21

generic-pers0n:
Fixes ~~#25969~~ #25484 (although this was tested with a different setup).

This PR makes a simple fix to get wxWidgets building under Arch Linux with Clang 21. It simply adds a cast inside `SimulateKeyboardKey()` inside `src/qt/uiaction.cpp` to convert the given `Qt::Key` to a character.

Note that in the issue I used a C-style cast as a proposed fix. In this PR, to follow the [coding guidelines](https://wxwidgets.org/develop/coding-guidelines), I used a `static_cast` instead of a regular C-style cast.


vadz:

PR #25977, opened 2025-11-17
add GetDirectory override to GTK3 

computergeek1507:
add GetDirectory override to GTK3 to fix #25976


vadz:
Thanks for fixing this!

@paulcor No objections from you?

PR #25978, opened 2025-11-18
make wxScrolled<> autoscroll region configurable

wsu-cb:
This is an initial attempt at configuring `wxScrolled<>`'s autoscroll region.  The default configuration emulates the existing behavior, but can be overridden to change the autoscroll region.  In particular, my personal requirement is for the region to be inside the window, but near the edge, but the configuration functions allow for some other possibilities.

Would anyone prefer a different API?  Or have any other comments?

### DIFF
--- a/include/wx/scrolwin.h
+++ b/include/wx/scrolwin.h
@@ -152,6 +152,27 @@ public:
     // returns 0 if no scrollbars are there.
     int GetScrollLines( int orient ) const;
 
+    // wxWidgets <= 3.3.1 autoscrolled exactly when the (captured) mouse cursor
+    // was outside the entire window.  Starting with wxWidgets 3.3.2, the
+    // region where the mouse cursor triggers autoscrolling is configurable.
+    // If the inner scroll zone is default, then the autoscroll region is
+    // relative to the WINDOW rect, and has width of size outer scroll zone.
+    // If the outer scroll zone width is default, then it extends infinitely
+    // outward from the WINDOW rect, and the behavior is the same as
+    // wxWidgets 3.3.1.  If the inner scroll zone is non-default, then the
+    // autoscroll region is relative to the CLIENT rect since that avoids
+    // requiring the user to figure out how the width of scroll region
+    // interacts with the scrollbar.
+    void SetInnerScrollZone(wxCoord innerZone);
+    void SetOuterScrollZone(wxCoord outerZone);
+
+    // Check whether clientPt triggers autoscrolling in each direction: return
+    // true if it does and fill the corresponing output parameter with the
+    // event type to generate.
+    bool AutoscrollTest(wxPoint clientPt,
+                        wxEventType& evtHorzScroll,
+                        wxEventType& evtVertScroll) const;
+
     // Set the x, y scrolling increments.
     void SetScrollRate( int xstep, int ystep );
 
@@ -263,8 +284,10 @@ public:
     // the methods to be called from the window event handlers
     void HandleOnScroll(wxScrollWinEvent& event);
     void HandleOnSize(wxSizeEvent& event);
-    void HandleOnMouseEnter(wxMouseEvent& event);
-    void HandleOnMouseLeave(wxMouseEvent& event);
+    void OnMotion(wxMouseEvent& event);
+    void OnLeftDown(wxMouseEvent& event);
+    void OnLeaveAutoScrollRegion();
+    void OnEnterAutoScrollRegion();
 #if wxUSE_MOUSEWHEEL
     void HandleOnMouseWheel(wxMouseEvent& event);
 #endif // wxUSE_MOUSEWHEEL
@@ -344,6 +367,10 @@ protected:
         return true;
     }
 
+
+    wxCoord               m_innerScrollZone = wxDefaultCoord;
+    wxCoord               m_outerScrollZone = wxDefaultCoord;
+    bool                  m_inAutoScrollRegion = false;
 
     double                m_scaleX;
     double                m_scaleY;
--- a/include/wx/scrolwin.h
+++ b/include/wx/scrolwin.h
@@ -167,7 +167,7 @@ public:
     void SetOuterScrollZone(wxCoord outerZone);
 
     // Check whether clientPt triggers autoscrolling in each direction: return
-    // true if it does and fill the corresponing output parameter with the
+    // true if it does and fill the corresponding output parameter with the
     // event type to generate.
     bool AutoscrollTest(wxPoint clientPt,
                         wxEventType& evtHorzScroll,
--- a/include/wx/scrolwin.h
+++ b/include/wx/scrolwin.h
@@ -155,16 +155,11 @@ public:
     // wxWidgets <= 3.3.1 autoscrolled exactly when the (captured) mouse cursor
     // was outside the entire window.  Starting with wxWidgets 3.3.2, the
     // region where the mouse cursor triggers autoscrolling is configurable.
-    // If the inner scroll zone is default, then the autoscroll region is
-    // relative to the WINDOW rect, and has width of size outer scroll zone.
-    // If the outer scroll zone width is default, then it extends infinitely
-    // outward from the WINDOW rect, and the behavior is the same as
-    // wxWidgets 3.3.1.  If the inner scroll zone is non-default, then the
-    // autoscroll region is relative to the CLIENT rect since that avoids
-    // requiring the user to figure out how the width of scroll region
-    // interacts with the scrollbar.
-    void SetInnerScrollZone(wxCoord innerZone);
-    void SetOuterScrollZone(wxCoord outerZone);
+    // Allow autoscrolling when the mouse is inside the window but within
+    // insideWidth pixels of the edge.
+    void EnableAutoScrollInside(wxCoord insideWidth);
+    // Forbid autoscrolling when the mouse is outside the window
+    void DisableAutoScrollOutside();
 
     // Check whether clientPt triggers autoscrolling in each direction: return
     // true if it does and fill the corresponding output parameter with the
@@ -368,8 +363,8 @@ protected:
     }
 
 
-    wxCoord               m_innerScrollZone = wxDefaultCoord;
-    wxCoord               m_outerScrollZone = wxDefaultCoord;
+    wxCoord               m_innerScrollWidth = 0;
+    bool                  m_outerScrollEnabled = true;
     bool                  m_inAutoScrollRegion = false;
 
     double                m_scaleX;
--- a/src/generic/scrlwing.cpp
+++ b/src/generic/scrlwing.cpp
@@ -77,18 +77,13 @@ class wxAutoScrollTimer : public wxTimer
 {
 public:
     wxAutoScrollTimer(wxWindow *winToScroll,
-                      wxScrollHelperBase *scroll,
-                      wxEventType eventTypeToSend,
-                      int pos, int orient);
+                      wxScrollHelperBase *scroll);
 
     virtual void Notify() override;
 
 private:
     wxWindow *m_win;
     wxScrollHelperBase *m_scrollHelper;
-    wxEventType m_eventType;
-    int m_pos,
-        m_orient;
 
     wxDECLARE_NO_COPY_CLASS(wxAutoScrollTimer);
 };
@@ -102,15 +97,10 @@ private:
 // ----------------------------------------------------------------------------
 
 wxAutoScrollTimer::wxAutoScrollTimer(wxWindow *winToScroll,
-                                     wxScrollHelperBase *scroll,
-                                     wxEventType eventTypeToSend,
-                                     int pos, int orient)
+                                     wxScrollHelperBase *scroll)
 {
     m_win = winToScroll;
     m_scrollHelper = scroll;
-    m_eventType = eventTypeToSend;
-    m_pos = pos;
-    m_orient = orient;
 }
 
 void wxAutoScrollTimer::Notify()
@@ -122,20 +112,44 @@ void wxAutoScrollTimer::Notify()
     }
     else // we still capture the mouse, continue generating events
     {
+        // where is the mouse?
+        // client coords
+        const wxPoint pt = m_win->ScreenToClient(wxGetMousePosition());
+
+        wxEventType horizontalEvent, verticalEvent;
+        // if no event needed, stop auto-scroll
+        if ( !m_scrollHelper->AutoscrollTest(pt, horizontalEvent, verticalEvent) )
+        {
+            Stop();
+            return;
+        }
+
         // first scroll the window if we are allowed to do it
-        wxScrollWinEvent event1(m_eventType, m_pos, m_orient);
-        event1.SetEventObject(m_win);
-        event1.SetId(m_win->GetId());
-        if ( m_scrollHelper->SendAutoScrollEvents(event1) &&
-                m_win->GetEventHandler()->ProcessEvent(event1) )
+        bool needMotion = false;
+        const auto orientations = {
+            std::make_pair(horizontalEvent, wxHORIZONTAL),
+            std::make_pair(verticalEvent, wxVERTICAL),
+        };
+        for (const auto& orientation : orientations)
+        {
+            if (orientation.first != wxEVT_NULL)
+            {
+                wxScrollWinEvent event1(orientation.first, 0, orientation.second);
+                event1.SetEventObject(m_win);
+                event1.SetId(m_win->GetId());
+                if ( m_scrollHelper->SendAutoScrollEvents(event1) &&
+                     m_win->GetEventHandler()->ProcessEvent(event1) )
+                {
+                    needMotion = true;
+                }
+            }
+        }
+
+        if (needMotion)
         {
             // and then send a pseudo mouse-move event to refresh the selection
             wxMouseEvent event2(wxEVT_MOTION);
-            event2.SetPosition(wxGetMousePosition());
-
-            // the mouse event coordinates should be client, not screen as
-            // returned by wxGetMousePosition
-            m_win->ScreenToClient(&event2.m_x, &event2.m_y);
+            event2.SetPosition(pt);
 
             event2.SetEventObject(m_win);
 
@@ -152,10 +166,6 @@ void wxAutoScrollTimer::Notify()
 
             m_win->GetEventHandler()->ProcessEvent(event2);
         }
-        else // can't scroll further, stop
-        {
-            Stop();
-        }
     }
 }
 #endif
@@ -170,6 +180,17 @@ void wxAutoScrollTimer::Notify()
 bool wxScrollHelperEvtHandler::ProcessEvent(wxEvent& event)
 {
     wxEventType evType = event.GetEventType();
+
+    // always process these mouse events ourselves, even if the user code handles
+    // them as well, as we need to autoscroll
+    if ( evType == wxEVT_LEFT_DOWN )
+    {
+        m_scrollHelper->OnLeftDown((wxMouseEvent&)event);
+    }
+    else if ( evType == wxEVT_MOTION )
+    {
+        m_scrollHelper->OnMotion((wxMouseEvent&)event);
+    }
 
     // Pass it on to the real handler: notice that we must not call
     // ProcessEvent() on this object itself as it wouldn't pass it to the next
@@ -248,14 +269,6 @@ bool wxScrollHelperEvtHandler::ProcessEvent(wxEvent& event)
         }
     }
 
-    if ( evType == wxEVT_ENTER_WINDOW )
-    {
-        m_scrollHelper->HandleOnMouseEnter((wxMouseEvent &)event);
-    }
-    else if ( evType == wxEVT_LEAVE_WINDOW )
-    {
-        m_scrollHelper->HandleOnMouseLeave((wxMouseEvent &)event);
-    }
 #if wxUSE_MOUSEWHEEL
     // Use GTK's own scroll wheel handling in GtkScrolledWindow
 #ifndef __WXGTK__
@@ -696,6 +709,83 @@ void wxScrollHelperBase::DoPrepareReadOnlyDC(wxReadOnlyDC& dc)
     dc.SetUserScale( m_scaleX, m_scaleY );
 }
 
+// see scrolwin.h for description
+void wxScrollHelperBase::SetInnerScrollZone(wxCoord innerZone)
+{
+    wxCHECK_RET( innerZone == wxDefaultCoord || innerZone >= 0,
+            "arg should be non-negative or wxDefaultCoord");
+    m_innerScrollZone = innerZone;
+}
+
+// see scrolwin.h for description
+void wxScrollHelperBase::SetOuterScrollZone(wxCoord outerZone)
+{
+    wxCHECK_RET( outerZone == wxDefaultCoord || outerZone >= 0,
+            "arg should be non-negative or wxDefaultCoord");
+    m_outerScrollZone = outerZone;
+}
+
+// check whether clientPt triggers autoscrolling in each direction
+bool
+wxScrollHelperBase::AutoscrollTest(wxPoint clientPt,
+                                   wxEventType& evtHorzScroll,
+                                   wxEventType& evtVertScroll) const
+{
+    // is mouse in autoscroll region?
+    wxRect inner, outer;
+    if ( m_innerScrollZone == wxDefaultCoord )
+    {
+        inner = m_win->GetRect();
+        if ( m_outerScrollZone != wxDefaultCoord )
+        {
+            outer = m_win->GetRect().Inflate(m_outerScrollZone);
+        }
+    }
+    else
+    {
+        inner = m_win->GetClientRect().Deflate(m_innerScrollZone);
+        if ( m_outerScrollZone != wxDefaultCoord )
+        {
+            outer = m_win->GetClientRect().Inflate(m_outerScrollZone);
+        }
+    }
+
+    bool inScrollRegion = !inner.Contains(clientPt) &&
+                          (outer == wxRect() || outer.Contains(clientPt));
+    if ( !inScrollRegion )
+    {
+        return false;
+    }
+
+    // can window can be scrolled in this direction?
+    if ( m_win->HasScrollbar(wxHORIZONTAL) )
+    {
+        if ( clientPt.x < inner.GetLeft() )
+        {
+            evtHorzScroll = wxEVT_SCROLLWIN_LINEUP;
+        }
+        else if (clientPt.x >= inner.GetRight() )
+        {
+            evtHorzScroll = wxEVT_SCROLLWIN_LINEDOWN;
+        }
+    }
+
+    // can window can be scrolled in this direction?
+    if ( m_win->HasScrollbar(wxVERTICAL) )
+    {
+        if ( clientPt.y < inner.GetTop() )
+        {
+            evtVertScroll = wxEVT_SCROLLWIN_LINEUP;
+        }
+        else if ( clientPt.y >= inner.GetBottom() )
+        {
+            evtVertScroll = wxEVT_SCROLLWIN_LINEDOWN;
+        }
+    }
+
+    return true;
+}
+
 void wxScrollHelperBase::SetScrollRate( int xstep, int ystep )
 {
     int old_x = m_xScrollPixelsPerLine * m_xScrollPosition;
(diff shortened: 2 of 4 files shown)


src/generic/scrlwing.cpp
```diff
@@ -171,6 +171,17 @@ bool wxScrollHelperEvtHandler::ProcessEvent(wxEvent& event)
 {
     wxEventType evType = event.GetEventType();
 
+    // always process these mouse events ourselves, even if the user code handles
+    // them as well, as we need to autoscroll
+    if ( evType == wxEVT_LEFT_DOWN )
+    {
+        m_scrollHelper->HandleOnLButtonUp((wxMouseEvent&)event);
```
vadz: It seems counterintuitive to process `wxEVT_LEFT_DOWN` events in a function called `OnLButtonUp`. I think wx convention would be to use just `OnLeftDown` (with "Mouse" and "Button" being implicit).

include/wx/scrolwin.h
```diff
@@ -152,6 +152,20 @@ class WXDLLIMPEXP_CORE wxScrollHelperBase : public wxAnyScrollHelperBase
     // returns 0 if no scrollbars are there.
     int GetScrollLines( int orient ) const;
 
+    // wxWidgets <= 3.3.1 autoscrolled exactly when the (captured) mouse cursor
```
vadz, line 155: I admit I have a lot of trouble parsing this. Maybe I could manage to do it, but it's easier to just ask questions, so here I go:

1. Why do we have 2 functions if they are not independent (or are they?)? It would be better to have a single function with 2 parameters instead, I think. It's annoying that they're of the same type and so it would be simple to give them in wrong order, but it would be also simpler to check for invalid combinations (e.g. I think that valid inner and invalid outer are invalid?).
2. Perhaps we should have 2 functions but based on whether they operate on window or client coordinates/rectangle basis? I think this would also make it simpler to use this.
3. Can this be used to disable auto-scroll entirely? It seems like it should be possible to, but it's not obvious how.
4. What is your actual use case, i.e. how do you use it?

vadz (commented):
Sorry, I'm really confused by the API, so I'd appreciate some clarification/examples. TIA!

src/generic/scrlwing.cpp
wsu-cb: Fixed in https://github.com/wxWidgets/wxWidgets/pull/25978/commits/77389a3fa03353c3fd894cc9a4b92e1212a54dc9

include/wx/scrolwin.h
wsu-cb: It is entirely possible that this interface is overly complicated by attempting to be both (unnecessarily?) general and default to backward compatible behavior.  Personally, I don't like the original behavior, but I assumed it had to remain the default.  The only behavior that I actually care about is having the window autoscroll when the mouse is in a picture-frame-shaped region (16 pixels wide in the example in the drawing sample) that covers the outermost portion of the client rectangle.  However, for generality, I chose an interface that allows using a picture-frame-shaped region outside the (window) rectangle rather than the original behavior's infinitely large region.  Also for generality, the interface allows for the autoscroll region to extend both inside and outside the client rectangle.  If it would make things easier to understand,  I would be happy to restrict the options to the original behavior and the picture-frame-in-client-space region.  (I would be even happier to remove the old behavior!)

To answer the specific questions about the proposed interface:

1:  The two functions in this interface are mostly, but not entirely, independent.  The dependency is that I want the inside-the-window autoscroll region to be based on the CLIENT rect so that no one has to worry about how to deal with making the autoscroll region account for the scrollbars.  However, the original code has the autoscroll region based on the WINDOW rect.  My attempt to resolve this inconsistency is to have the outside-the-window autoscroll region be relative to the WINDOW rect by default, but relative to the CLIENT rect if there is an inside-window autoscroll region.

1:  I have no objection to a single function, but since I considered the zones mostly independent, I thought it was plausible that some API users would only want to change one of them.

1:  I don't know what you mean by invalid combinations.  The invalid values are negative sizes other than wxDefaultCoord, and that is invalid no matter what the other parameter is.  In the case of a strictly positive value `K` inner region and a wxDefaultCoord outer region, then the inner region is the expected picture-frame with width `K`, and the outer region is the entire plane outside the client rect, so the total autoscroll region is the entire plane minus a  rectangular hole in the middle of the client rect.

2:  We could have both client and window rect based versions of the zones.  I don't want to define an inside-the-window zone that extends into the scroll bar, but, now that you suggest it, I see that Windows Explorer's autoscroll zone includes the scrollbar,  I had not previously noticed that.

3:  Yes, this can disable autoscroll.  I have added a demonstration of that in https://github.com/wxWidgets/wxWidgets/pull/25978/commits/16ea5d5d39645aa2401c87fa17eb781c8ffc59fc.

4:  My use case is dragging objects both within a window and from one window to another.  Therefore, like Windows Explorer (see https://groups.google.com/g/wx-dev/c/9tJ3mpkJ4NQ/m/ysvIr7llAQAJ), I want the autoscroll region to be INSIDE the window, not outside it.  While dragging inside the original window, the original window should autoscroll, but when dragging out of the window, the original window should stop autoscrolling, and when the dragged object is over a destination window, the destination window should autoscroll.  This PR is only taking care of changing the autscrolling of the original window; once this is done, I will have to implement autoscrolling the destination window.

trogmaniac:
If the autoscroll code is touched at all, i think it should also fix the (IMHO) only flaw it currently has: It can't scroll in both directions at the same time. E.g. if you drag and move the mouse outside at the bottom, it starts scrolling down. But if you then also move the mouse outside to the right side, it won't start also scrolling in that directions.

wsu-cb:
> If the autoscroll code is touched at all, i think it should also fix the (IMHO) only flaw it currently has: It can't scroll in both directions at the same time. E.g. if you drag and move the mouse outside at the bottom, it starts scrolling down. But if you then also move the mouse outside to the right side, it won't start also scrolling in that directions.

I have made this change in https://github.com/wxWidgets/wxWidgets/pull/25978/commits/81103a58cd078417fd6780e1d85cb244d5e3198f, and then re-built the autoscroll zone changes on top of that.

vadz:
Sorry for the delay, I'll try to get to this before 3.3.2.

include/wx/scrolwin.h
```diff
@@ -152,6 +152,23 @@ class WXDLLIMPEXP_CORE wxScrollHelperBase : public wxAnyScrollHelperBase
...
+    // outward from the WINDOW rect, and the behavior is the same as
+    // wxWidgets 3.3.1.  If the inner scroll zone is non-default, then the
+    // autoscroll region is relative to the CLIENT rect since that avoids
+    // requiring the user to figure out how the width of scroll region
+    // interacts with the scrollbar.
+    void SetInnerScrollZone(wxCoord innerZone);
+    void SetOuterScrollZone(wxCoord outerZone);
+    // check whether clientPt triggers autoscrolling in each direction
+    std::pair<wxEventType /*horz*/,
+              wxEventType /*vert*/> AutoscrollTest(wxPoint clientPt) const;
```
vadz: This is relatively minor, but I really dislike using `std::pair` as `first` and `second` are horrible field names. I thought about replacing it with an ad hoc `struct ScrollEvents { wxEventType horz, vert; }` but after looking at how this function is used, I think it would make even more sense to give it this signature:

```suggestion
    bool AutoscrollTest(wxPoint clientPt, wxEventType& evtHorzScroll, wxEventType& evtVertScroll) const;
```

to avoid tests for both of the events being `wxEVT_NULL` as the function would just return `false` in this case.

What do you think?

include/wx/scrolwin.h
```diff
@@ -345,6 +364,10 @@ class WXDLLIMPEXP_CORE wxScrollHelperBase : public wxAnyScrollHelperBase
     }
 
 
+    wxCoord               m_innerScrollZone = wxDefaultCoord;
+    wxCoord               m_outerScrollZone = wxDefaultCoord;
+    bool                  m_inNonScrollRegion = true;
```
vadz: Why use negation here? Wouldn't


```suggestion
    bool                  m_inAutoScrollRegion = false;
```

be clearer?

src/generic/scrlwing.cpp
```diff
@@ -122,20 +112,46 @@ void wxAutoScrollTimer::Notify()
     }
     else // we still capture the mouse, continue generating events
     {
+        // where is the mouse?
+        // client coords
+        wxPoint pt = m_win->ScreenToClient(wxGetMousePosition());
```
vadz: Very minor, but could be made more clear that it's not going to change:
```suggestion
        const wxPoint pt = m_win->ScreenToClient(wxGetMousePosition());
```

src/generic/scrlwing.cpp
```diff
@@ -122,20 +112,46 @@ void wxAutoScrollTimer::Notify()
...
+        }
+
         // first scroll the window if we are allowed to do it
-        wxScrollWinEvent event1(m_eventType, m_pos, m_orient);
-        event1.SetEventObject(m_win);
-        event1.SetId(m_win->GetId());
-        if ( m_scrollHelper->SendAutoScrollEvents(event1) &&
-                m_win->GetEventHandler()->ProcessEvent(event1) )
+        bool needMotion = false;
+        auto orientations = {
```
vadz: As above
```suggestion
        const auto orientations = {
```

src/generic/scrlwing.cpp
```diff
@@ -122,20 +112,46 @@ void wxAutoScrollTimer::Notify()
...
+            std::make_pair(horizontalEvent, wxHORIZONTAL),
+            std::make_pair(verticalEvent, wxVERTICAL),
+        };
+        for (auto& orientation : orientations)
```
vadz: And here too


```suggestion
        for ( const auto& orientation : orientations )
```

src/generic/scrlwing.cpp
```diff
@@ -122,20 +112,46 @@ void wxAutoScrollTimer::Notify()
...
+        {
+            if (orientation.first != wxEVT_NULL)
+            {
+                wxScrollWinEvent event1(orientation.first, 0, orientation.second);
+                event1.SetEventObject(m_win);
+                event1.SetId(m_win->GetId());
+                needMotion = ( m_scrollHelper->SendAutoScrollEvents(event1) &&
+                                  m_win->GetEventHandler()->ProcessEvent(event1) )
+                             ||
+                             needMotion;
```
vadz: I think this would be simpler to parse:

```suggestion
                if ( m_scrollHelper->SendAutoScrollEvents(event1) &&
m_win->GetEventHandler()->ProcessEvent(event1) )
                    needMotion = true;
```

src/generic/scrlwing.cpp
```diff
@@ -696,6 +711,93 @@ void wxScrollHelperBase::DoPrepareReadOnlyDC(wxReadOnlyDC& dc)
...
+// was outside the entire window.  Starting with wxWidgets 3.3.2, the
+// region where the mouse cursor triggers autoscrolling is configurable.
+// If the inner scroll zone is default, then the autoscroll region is
+// relative to the WINDOW rect, and has width of size outer scroll zone.
+// If the outer scroll zone width is default, then it extends infinitely
+// outward from the WINDOW rect, and the behavior is the same as
+// wxWidgets 3.3.1.  If the inner scroll zone is non-default, then the
+// autoscroll region is relative to the CLIENT rect since that avoids
+// requiring the user to figure out how the width of scroll region
+// interacts with the scrollbar.
```
vadz: I'd rather not duplicate exactly the same comment both in the header and here as they will inevitably get out of sync later resulting in confusion. Let's have it in one place only (not sure where it's more appropriate) and just refer to it in the other one.

include/wx/scrolwin.h
```diff
@@ -263,8 +280,10 @@ class WXDLLIMPEXP_CORE wxScrollHelperBase : public wxAnyScrollHelperBase
     // the methods to be called from the window event handlers
     void HandleOnScroll(wxScrollWinEvent& event);
     void HandleOnSize(wxSizeEvent& event);
-    void HandleOnMouseEnter(wxMouseEvent& event);
-    void HandleOnMouseLeave(wxMouseEvent& event);
+    void OnMotion(wxMouseEvent& event);
+    void OnLeftDown(wxMouseEvent& event);
+    void OnEnterNonScrollRegion();
+    void OnLeaveNonScrollRegion();
```
vadz: I think these function should be private.

Also, perhaps it would be more clear to avoid using "Non" in their names, e.g.

```suggestion
    void OnLeaveAutoScrollRegion();
    void OnEnterAutoScrollRegion();
```

?

vadz (commented):
Sorry for the long delay, I've finally looked at this again and I think I understand the logic better now, so I'm ready to merge this, but I wonder if this could be improved a bit first. Notably I'd really like to avoid using `std::pair` and could make the proposed change myself if you agree.

I think trying to make things more clear is worth it as this is not the most obvious part of code already...

include/wx/scrolwin.h
wsu-cb: `wxScrollHelperEvtHandler` calls `OnMotion()`, etc., but these functions are in `wxScrollHelperBase`, so they can't be private unless `wxScrollHelperEvtHandler` becomes a `friend`.  Is that your intention?

vadz: Sorry, I had overlooked that they were called from `wxScrollHelperEvtHandler` to be honest, so please disregard the first part of this comment (the second one would still be worth considering IMO).

This being said, I rather like the idea of making `wxScrollHelperEvtHandler` friend of this class instead of making all these functions public. But this is not really related to this PR and probably would be better done in a separate one.

include/wx/scrolwin.h
wsu-cb: I actually initially wrote this in positive form, and then switched it to negative because the original code used `OnEnterWindow()`, and to keep the `OnEnter` equivalent, I used `OnEnterNonScrollRegion()`.  However, I agree that `OnLeaveAutoScrollRegion()` is clearer if one isn't trying to match the original code.

Fixed in https://github.com/wxWidgets/wxWidgets/pull/25978/commits/563f042d114c901616e2aef8b6219c16ea4632cb

src/generic/scrlwing.cpp
wsu-cb: Do you want local variables marked `const` (when possible) in general?

Fixed in https://github.com/wxWidgets/wxWidgets/pull/25978/commits/e3c116d6fdd95321f39e145c5316c1b355580657

src/generic/scrlwing.cpp
wsu-cb: Fixed in https://github.com/wxWidgets/wxWidgets/pull/25978/commits/e3c116d6fdd95321f39e145c5316c1b355580657

src/generic/scrlwing.cpp
wsu-cb: Fixed in https://github.com/wxWidgets/wxWidgets/pull/25978/commits/e3c116d6fdd95321f39e145c5316c1b355580657

src/generic/scrlwing.cpp
wsu-cb: Fixed in https://github.com/wxWidgets/wxWidgets/pull/25978/commits/36f0c5bee54441d21fb61b9b9aee3184d216c4a4

src/generic/scrlwing.cpp
wsu-cb: Fixed in https://github.com/wxWidgets/wxWidgets/pull/25978/commits/c04d8bff1a5723398edbd3028b87e1114f3c9ad2

include/wx/scrolwin.h
wsu-cb: Second part of comment fixed in https://github.com/wxWidgets/wxWidgets/pull/25978/commits/563f042d114c901616e2aef8b6219c16ea4632cb

vadz (approved):

wsu-cb:
> Looks good, thanks! Will merge soon.

I'm certainly not going to complain if you merge now, but, now that you have approved the implementation,I was planning to document `SetInnerScrollZone()` and `SetInnerScrollZone()` in `interface/wx/scrolwin.h` before merging.

wsu-cb:
Commit https://github.com/wxWidgets/wxWidgets/pull/25978/commits/d313c5d0c6718f187b4d8d4aa94b6279fdf1e086 fixes some typos unrelated to this pull request.

vadz:
Thanks, I've indeed forgotten that this still needed to be documented. But while applying the docs commit (I'm afraid I've started rewriting it trying to make it more clear...) I've realized that I can't explain the difference between inner zone being set to `0` or `-1`: both values behave the same, don't they?

OTOH setting outer zone to `0` is very different from setting it to `-1` and this doesn't seem good.

vadz:
Maybe we should handle `-1` as `INT_MAX`, i.e. enable autoscrolling anywhere inside the window when the inner scroll zone is set to this value? This doesn't seem very useful, but it would be consistent with the outer zone.

Of course, if we do this, its default value would need to be changed to 0.

vadz:
My attempt at documenting the existing behaviour:

```diff
diff --git a/interface/wx/scrolwin.h b/interface/wx/scrolwin.h
index 7eb4774d1b..67da5ea2e8 100644
--- a/interface/wx/scrolwin.h
+++ b/interface/wx/scrolwin.h
@@ -95,6 +95,28 @@ enum wxScrollbarVisibility
     window out of the visible area), the child window will report a position
     of (10,-90).
 
+    @section scrolled_autoscroll Automatic scrolling
+
+    Scrolled windows implement automatic scrolling behaviour: when the mouse is
+    captured by the application and the user holds the mouse button down, the
+    window may start and keep scrolling even if the mouse doesn't move,
+    provided that it is in the "autoscroll zone". This is convenient to allow
+    extending selection to the parts of the window currently outside of the
+    visible area, for example.
+
+    By default, autoscrolling is triggered when the mouse is anywhere outside
+    of the window, i.e. the window starts scrolling when the (captured) mouse
+    pointer leaves the window and stops when it re-enters the window. However,
+    this behaviour can be customized using SetInnerScrollZone() and
+    SetOuterScrollZone() functions. The first of them allows to enable
+    autoscrolling even when mouse is inside the window, but close to its
+    border, while the second one allows to disable autoscrolling when mouse is
+    outside the window.
+
+    To disable autoscrolling completely, call SetOuterScrollZone() with 0
+    argument.
+
+
     @beginStyleTable
     @style{wxHSCROLL}
            If this style is specified and ::wxVSCROLL isn't, the window will be
@@ -603,6 +625,47 @@ public:
     void SetScrollPageSize(int orient, int pageSize);
     int GetScrollLines( int orient ) const;
 
+    /**
+        Set the width of the autoscroll zone inside the client rectangle.
+
+        Setting inner scroll zone to a non-zero value enables autoscrolling if
+        the distance between the mouse and the closest edge of the client
+        rectangle is less than the given value.
+
+        By default, autoscrolling is not triggered when the mouse is inside the
+        window.
+
+        @see @ref scrolled_autoscroll, SetOuterScrollZone()
+
+        @param innerZone
+            The width of the inner scroll zone in pixels. If this parameter is
+            set to either 0 or ::wxDefaultCoord, there is no inner scroll zone
+            and autoscrolling may be triggered only when the mouse is outside
+            of the window (unless it is disabled too).
+
+        @since 3.3.2
+    */
+    void SetInnerScrollZone(wxCoord innerZone);
+
+    /**
+        Set the width of the auto-scroll zone outside the window.
+
+        By default, autoscrolling is triggered when the mouse is anywhere
+        outside of the window.
+
+        @see @ref scrolled_autoscroll, SetInnerScrollZone()
+
+        @param outerZone
+            The width of the outer scroll zone in pixels. If this parameter is
+            set to ::wxDefaultCoord, autoscrolling is triggered when the mouse
+            is anywhere outside of the window. May be set to 0 to disable
+            autoscrolling when the mouse is outside of the window entirely.
+            It is typically not useful to set it at any other value.
+
+        @since 3.3.2
+    */
+    void SetOuterScrollZone(wxCoord outerZone);
+
     /**
         Set the scaling factor for the window.
 
```

I think this is correct (but please let me know if it isn't), but this shows at least 2 problems with the current API:

1. Previously mentioned inconsistency between the value of `wxDefaultCoord` for inner and outer zone: in one case, it disables autoscroll, in the other one it allows it anywhere.
2. Lack of any real use for the `SetOuterScrollZone()` parameter: the only useful value for it is 0, it just doesn't make sense to call it with anything else.

I think that perhaps we should have the following functions instead:

- `EnableInnerAutoScroll(int width)` where `width > 0`.
- `DisableOuterAutoScroll()`.

This is still not consistent, but at least it's pretty clear what the functions do and they don't allow doing anything useless, such as calling `SetOuterScrollZone(5)`.

What do you think?

wsu-cb:
> Thanks, I've indeed forgotten that this still needed to be documented. But while applying the docs commit (I'm afraid I've started rewriting it trying to make it more clear...) I've realized that I can't explain the difference between inner zone being set to `0` or `-1`: both values behave the same, don't they?
> 

Setting the inner zone to `-1` (aka `wxDefaultCoord`) is not the same as `0`, but I admit you might not like the difference:  the default `wxDefaultCoord` defines the OUTER scroll zone relative to the WINDOW rect, but setting the inner scroll zone to `0` defines the OUTER scroll zone relative to the CLIENT rect.

> OTOH setting outer zone to `0` is very different from setting it to `-1` and this doesn't seem good.

I don't understand why `0` and `-1` being different is objectionable.

vadz:
> Setting the inner zone to `-1` (aka `wxDefaultCoord`) is not the same as `0`, but I admit you might not like the difference: the default `wxDefaultCoord` defines the OUTER scroll zone relative to the WINDOW rect, but setting the inner scroll zone to `0` defines the OUTER scroll zone relative to the CLIENT rect.

Oh, I see (after looking at the code again). This is still really confusing and not at all consistent with how these values are used for the outer zone, however.

Shouldn't outer zone always cover everything outside of the client area?

> > OTOH setting outer zone to `0` is very different from setting it to `-1` and this doesn't seem good.
> 
> I don't understand why `0` and `-1` being different is objectionable.

This is not objectionable, of course. It's the difference between interpretations of `-1` between zones which is: for the inner zone it is (roughly) the same as `0`, but for the outer zone it's (roughly) exactly the opposite of `0`.

wsu-cb:
> Maybe we should handle `-1` as `INT_MAX`, i.e. enable autoscrolling anywhere inside the window when the inner scroll zone is set to this value? This doesn't seem very useful, but it would be consistent with the outer zone.
> 
> Of course, if we do this, its default value would need to be changed to 0.

I prefer that the default value be `wxDefaultCoord`, but, as maintainer, you have the final decision.

vadz:
> I prefer that the default value be `wxDefaultCoord`, but, as maintainer, you have the final decision.

I'd rather reach some consensus. I understand that it's better to use `wxDefaultCoord` as default value, but this is in conflict with inner autoscroll being disabled by default and outer being enabled. And IMO it's worse for the same value to mean 2 different things in related contexts.

What do you think about `EnableInnerAutoScroll(width)` and `DisableOuterAutoScroll()` idea? The names could be improved, e.g. `EnableAutoScrollInside()` and `DisableAutoScrollOutside()` might be better.

wsu-cb:
> Shouldn't outer zone always cover everything outside of the client area?

I think this is the crux of the issue.  As I wrote in https://github.com/wxWidgets/wxWidgets/pull/25978#discussion_r2595800569, it is entirely possible that I have attempted too much generality.  Because I don't like the outer scroll zone at all, I am unqualified to determine whether a partial outer scroll zone is ever useful.

> This is not objectionable, of course. It's the difference between interpretations of `-1` between zones which is: for the inner zone it is (roughly) the same as `0`, but for the outer zone it's (roughly) exactly the opposite of `0`.

I see your point now, and can't dispute it.

For my purposes, the suggested API in https://github.com/wxWidgets/wxWidgets/pull/25978#issuecomment-3902554264 would work well.  I chose a general API to avoid objections about it being inconsistent to permit a partial inner scroll zone but not a partial outer scroll zone, but if no one values that generality, I actually prefer the simpler API as well.

vadz:
If nobody else (if anybody else is reading this in the first place...) has any better ideas, I'll change the API like this — or, alternatively, please let me know if you'd prefer to do it yourself. I've pushed the commits I was about to merge in configurable-auto-scroll branch in my fork, if you make any changes, please base them either on master or on that branch because it has some changes compared to this PR.

wsu-cb:
> If nobody else (if anybody else is reading this in the first place...) has any better ideas, I'll change the API like this — or, alternatively, please let me know if you'd prefer to do it yourself. I've pushed the commits I was about to merge in configurable-auto-scroll branch in my fork, if you make any changes, please base them either on master or on that branch because it has some changes compared to this PR.

I will look at your branch and get back to you...

wsu-cb:
> I will look at your branch and get back to you...

After experimenting, I now think the additional complexity in my previous design of sometimes setting the autoscroll region relative to the client rectangle is also not worthwhile.  I now think I will revert to the wxWidgets 3.2 behavior of always basing the autoscroll region on the window rectangle (although the implementation is unrelated to the 3.2 implementation using `wxEVT_ENTER_WINDOW`/`wxEVT_LEAVE_WINDOW`).

wsu-cb:
Commit https://github.com/wxWidgets/wxWidgets/pull/25978/commits/4ab8c812228ce041ce2b094668f2e86bf1b4f4c8 modifies `interface/wx/scrolwin.h` to match the recent discussion.  Does this look good to you?  If so, I will modify the implementation to match.

vadz:
Yes, thank you, this looks good, the only remaining question is whether we should allow passing `wxDefaultCoord` to `EnableAutoScrollInside()`. As long as it works in the same way as `0`, I would say that we shouldn't and that the function should assert that its argument is `>= 0`. And setting it to 0 would disable inner auto scroll region, of course.

wsu-cb:
> Yes, thank you, this looks good, the only remaining question is whether we should allow passing `wxDefaultCoord` to `EnableAutoScrollInside()`. As long as it works in the same way as `0`, I would say that we shouldn't and that the function should assert that its argument is `>= 0`. And setting it to 0 would disable inner auto scroll region, of course.

As noted in https://github.com/wxWidgets/wxWidgets/pull/25978#issuecomment-3902538102, my preference is that the default value be `wxDefaultCoord`, even if it acts the same as `0`, and even if it means a little bit of extra overhead in the implementation.

vadz:
But now there is no more distinction between client and window, so `-1` and `0` behave exactly the same, IIUC. Why have 2 ways to do the same thing?

Alternative would be to forbid passing 0 to this function, but this doesn't look logical.

wsu-cb:
> But now there is no more distinction between client and window, so `-1` and `0` behave exactly the same, IIUC. Why have 2 ways to do the same thing?

It bothers me for `wxDefaultCoord` to be illegal, and it seems perfectly reasonable to me for `wxDefaultCoord` to mean something that can also be specified explicitly.  In fact, I would EXPECT that to be true in most cases.

> Alternative would be to forbid passing 0 to this function, but this doesn't look logical.

I agree that forbidding `0` here would be even worse than forbidding `wxDefaultCoord`.

vadz:
> it seems perfectly reasonable to me for `wxDefaultCoord` to mean something that can also be specified explicitly

I can't think of any occurrence of this in wxWidgets API. `wxDefaultXXX` always means "unspecified" or, well, "default". There is no reasonable default value for the inner auto scroll zone size, but I'd actually expect setting it to `wxDefaultCoord` to do exactly this, i.e. choose some predefined value like `FromDIP(5)`, rather than disable it entirely.

In fact, you've helped me put my finger on it: this API looks wrong to me because it uses `wxDefaultCoord` as meaning "none" and this is just not how it's used anywhere else.

wsu-cb:
> I can't think of any occurrence of this in wxWidgets API. `wxDefaultXXX` always means "unspecified" or, well, "default". There is no reasonable default value for the inner auto scroll zone size, but I'd actually expect setting it to `wxDefaultCoord` to do exactly this, i.e. choose some predefined value like `FromDIP(5)`, rather than disable it entirely.

This appears to be where we disagree:  I think `0` is the default value.  The difference between `1` and `0` is enormous in a logical sense, but for the user, a 1-pixel wide scroll zone is almost as unhittable as a 0-pixel wide scroll zone.

> 
> In fact, you've helped me put my finger on it: this API looks wrong to me because it uses `wxDefaultCoord` as meaning "none" and this is just not how it's used anywhere else.

vadz:
I'd still keep `wxDefaultCoord` for the future use as meaning "use default width for the inner zone" and would let use 0 to disable it now (not that it will be often useful, as it's disabled by default and I don't see why would you ever enable it first and then disable it later). But I won't argue any more.

vadz:
One way or another, this needs to be finalized soon if it's to be included in 3.3.2.

wsu-cb:
> One way or another, this needs to be finalized soon if it's to be included in 3.3.2.

I have pushed a new version with the simplified API discussed above.  (This implementation forbids `wxDefaultCoord`.)

vadz:
Thanks, I'll merge and push this soon, it can always be improved/extended later if necessary.

PR #25982, opened 2025-11-21
Remove unused, commented-out wxGetCurrentDir() that was never used

lanurmi:
This implementation was surrounded by `#if 0` ever since it was added in 7c07201 (don't disable hidden windows in wxWindowDisabler, it's at best useless, 2003-06-21)


vadz:
Thanks, will merge soon.

Issue #23841, opened 2023-08-31
wxAutoBufferedPaintDC can cause very slow rendering if drawing bitmaps with alpha on MSW


vadz:
wxAutoBufferedPaintDC can cause very slow rendering if drawing bitmaps with alpha on MSW.

The reason is that the backing store bitmap used by wxAutoBufferedPaintDC is a DDB and not a DIB.  If you draw an alpha bitmap onto the DC then AlphaBlt will use wxAlphaPixelData which when not using a DIB will make a copy of the whole backing store.  It does this for each bitmap draw (can be pretty bad on wxAuiToolBar).
I've worked around it by making my own double buffering solution that uses this:

```cpp
            // Convert to DIB.  This way, if transparent images are drawn then 
            // you do not get a DDB=>DIB=>DDB conversion via temporary
            // expensive resource allocations!
            store.ConvertToDIB();
```
Thats meant I have to derive wxAuiToolBar and re-implement OnPaint.  Would be nice to be able to avoid doing that.

_Originally posted by @petebannister in https://github.com/wxWidgets/wxWidgets/issues/23585#issuecomment-1699280806_

vadz:
For the record, this is probably the same problem as alluded to here:

https://github.com/wxWidgets/wxWidgets/blob/3861ad6b6d9b202dcb0f1cb56bcd52a7c375290a/src/richtext/richtextctrl.cpp#L3025-L3029

Except that using depth of 24 doesn't work when alpha is effectively used, of course. OTOH fixing this would allow to remove the workaround in the code above.

dsa-t:
KiCad also spends 90% of time drawing bitmaps while resizing

![image](https://github.com/wxWidgets/wxWidgets/assets/37658952/e7064e21-4eb9-4706-bdb8-a062b87a0dea)

dsa-t:
After updating from 3.2.4 to 3.2.5, `aDc.GetAsBitmap().ConvertToDIB();` causes a (recursive) assert
"deleting bitmap still selected into wxMemoryDC"

vadz:
It's normal that destroying a bitmap selected into a memory DC results in an assert, this really shouldn't be done.

[...]

PR #25983, opened 2025-11-21
Avoid slow drawing of bitmaps with alpha in wxAUI

vadz:
See #23841.

If no problems are found with this, it should be backported to 3.2 by just adding `24` to the code creating the bitmap (i.e. without the first commit here).

### DIFF
--- a/include/wx/dcbuffer.h
+++ b/include/wx/dcbuffer.h
@@ -90,7 +90,7 @@ public:
     {
         InitCommon(dc, style);
 
-        UseBuffer(area.x, area.y);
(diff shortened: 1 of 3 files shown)


marekr:
So far the individual that experienced the slowdowns the heaviest reports it seems to be fixed with this patch applied to our 3.2 based builds. Letting it sit awhile longer since it wasn't always the most consistently reproduced issue

vadz:
I'm going to merge this, as I'd like to backport it to 3.2 too, but any feedback would still be welcome, of course!

PR #25985, opened 2025-11-23
Do not log bogus system error in wxRenameFile() on non-Windows

lanurmi:
On non-Windows, there is no failed system call in the case when destination file exists, so there is no relevant system error available either.

### DIFF
--- a/src/common/filefn.cpp
+++ b/src/common/filefn.cpp
@@ -601,7 +601,7 @@ wxRenameFile(const wxString& file1, const wxString& file2, bool overwrite)
 
     if ( !overwrite && wxFileExists(file2) )
     {
-        wxLogSysError
+        wxLogError
         (
             _("Failed to rename the file '%s' to '%s' because the destination file already exists."),
             file1.c_str(), file2.c_str()


vadz:
Maybe I'm confused, but I think that there is no system call under Windows either: this error is only given if `!overwrite` but `::ReplaceFile()` is called only if `overwrite`.

So shouldn't this `wxLogSysError()` be just replaced with `wxLogError()` unconditionally?

lanurmi:
Oh, apparently I was confused and misread the code. I'll push a fixup that calls `wxLogError()`, and additionally logs a system error if `::ReplaceFile()` failed.

src/common/filefn.cpp
```diff
@@ -593,6 +593,14 @@ wxRenameFile(const wxString& file1, const wxString& file2, bool overwrite)
         {
             return true;
         }
+        else
+        {
+            wxLogSysError
+            (
+                _("Failed to rename the file '%s' to '%s' by overwriting the already existing destination file"),
+                file1.c_str(), file2.c_str()
```
PBfordev: Do we need those two `c_str()` calls here? I thought it has been safe to pass `wxString` directly to **wx** vararg functions since at least v2.9?

Also, by changing the message, we are losing its translations (the new message is also missing the end period)...

vadz: We definitely don't need `c_str()`s, there are just unfortunately many of them left over.

But this is not a changed message, it's a new one, so this is not a problem.

What **is** a problem, however, is that we give an error before potentially succeeding in renaming the file ourselves below, which is definitely confusing and I would say wrong. I think we'll just have to live without a specific error in this case because otherwise you'd need more extensive changes (like remembering last error here and using it below, for wxMSW only).

PR #25986, opened 2025-11-23
Support EGL 1.4 instead of previously required 1.5

### LINKED ISSUE #22325: Add run-time check for EGL version on canvas creation

vadz:
Check for the availability of eglCreatePlatformWindowSurfaceEXT() if EGL version is < 1.5 and use it if it's available, as it allows applications using wxGLCanvasEGL to work on some (many?) systems without EGL 1.5 support.

Closes #22325.

### DIFF
--- a/include/wx/unix/glegl.h
+++ b/include/wx/unix/glegl.h
@@ -131,6 +131,15 @@ public:
     wl_egl_window *m_wlEGLWindow = nullptr;
 
 private:
+    // Call eglCreatePlatformWindowSurface() when using EGL 1.5 or later,
+    // otherwise try eglCreatePlatformWindowSurfaceEXT() if it's available and
+    // fall back on eglCreateWindowSurface() otherwise.
+    //
+    // This function uses m_display and m_config which must be initialized
+    // before using it and should be passed either m_xwindow or m_wlEGLWindow
+    // depending on whether we are using X11 or Wayland.
+    EGLSurface CallCreatePlatformWindowSurface(void *window) const;
+
 
     EGLConfig m_config = nullptr;
     EGLDisplay m_display = nullptr;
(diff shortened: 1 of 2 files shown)


dsa-t:
Looks good for me (I don't have a test setup though)

I think the commit message should say "Support EGL 1.4 instead of previously required 1.5"

vadz:
Merged in the commit above.

PR #25991, opened 2025-11-25
Fix initial scale of wxGLCanvas under EGL/Wayland

vadz:
The scale of the canvas was set up correctly only once we received "size-allocate" signal from GTK, but this doesn't necessarily happen when the window is first shown, e.g. the cube sample doesn't get this signal, at least when using Sway.

Ensure that we still use the correct scale by catching "draw" signal as well and updating the scale if it has changed.

See #23733, #25465.

----

This fixes the problem, but I wonder if there is some better way to do it, maybe @paulcor knows of it?

Also cc @Popax21 just FYI as this is related to your changes.

### DIFF
--- a/include/wx/unix/glegl.h
+++ b/include/wx/unix/glegl.h
@@ -141,6 +141,7 @@ private:
     wl_region *m_wlRegion = nullptr;
     wl_subsurface *m_wlSubsurface = nullptr;
 
+    int m_scale = 1;
     bool m_readyToDraw = false;
     bool m_swapIntervalSet = false;
 
@@ -148,7 +149,8 @@ private:
     static EGLConfig ms_glEGLConfig;
 
     friend void wxEGLUpdatePosition(wxGLCanvasEGL* win);
-    friend void wxEGLSetScale(wxGLCanvasEGL* win, int scale);
+    friend void wxEGLUpdatePositionAndScale(wxGLCanvasEGL* win, int scale);
+    friend void wxEGLUpdateScaleIfChanged(wxGLCanvasEGL* win, int scale);
 };
 
 // ----------------------------------------------------------------------------
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -437,9 +437,21 @@ void wxEGLUpdatePosition(wxGLCanvasEGL* win)
 }
 
 // Helper declared as friend in the header and so can access m_wlSurface.
-void wxEGLSetScale(wxGLCanvasEGL* win, int scale)
+void wxEGLUpdatePositionAndScale(wxGLCanvasEGL* win, int scale)
 {
+    wl_egl_window_resize(win->m_wlEGLWindow, win->m_width * scale,
+                         win->m_height * scale, 0, 0);
+
+    wxEGLUpdatePosition(win);
+
     wl_surface_set_buffer_scale(win->m_wlSurface, scale);
+    win->m_scale = scale;
+}
+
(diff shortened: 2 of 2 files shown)


paulcor:
The "size-allocate" definitely does occur, it's just coming in before the handler is connected. And it's not specific to Sway, I can see it with Weston.

This all seems unnecessarily complicated, I don't think there is any need to cache the scale. A less ugly version of the following would probably work:
```diff
diff --git a/src/unix/glegl.cpp b/src/unix/glegl.cpp
index 78f8f3a6e0..609f5c287e 100644
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -585,6 +585,9 @@ bool wxGLCanvasEGL::CreateSurface()
         return false;
     }
 
+#ifdef GDK_WINDOWING_WAYLAND
+    gtk_glcanvas_size_callback(m_widget, nullptr, this);
+#endif
     return true;
 }
 
```

vadz:
> The "size-allocate" definitely does occur, it's just coming in before the handler is connected.

Sorry, I don't see it. Even if I move the connecting call to the very beginning like this, I still don't get it:

```diff
diff --git a/src/unix/glegl.cpp b/src/unix/glegl.cpp
index 78f8f3a6e0..e9b7d7fb22 100644
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -542,6 +542,9 @@ bool wxGLCanvasEGL::CreateSurface()
             return true;
         }
 
+        g_signal_connect(m_widget, "size-allocate",
+                         G_CALLBACK(gtk_glcanvas_size_callback), this);
+
         int w = gdk_window_get_width(window);
         int h = gdk_window_get_height(window);
         struct wl_display *display = gdk_wayland_display_get_wl_display(gdk_window_get_display(window));
@@ -574,8 +577,6 @@ bool wxGLCanvasEGL::CreateSurface()
         // Not unmapping the canvas as soon as possible causes problems when reparenting
         g_signal_connect(m_widget, "unmap",
                          G_CALLBACK(gtk_glcanvas_unmap_callback), this);
-        g_signal_connect(m_widget, "size-allocate",
-                         G_CALLBACK(gtk_glcanvas_size_callback), this);
     }
 #endif
 
```

> This all seems unnecessarily complicated, I don't think there is any need to cache the scale.

I fully agree with this and I'd like to have another solution.

> A less ugly version of the following would probably work:

Unfortunately this doesn't work, at this time the scale factor is still 1.

paulcor:
> Sorry, I don't see it. Even if I move the connecting call to the very beginning like this, I still don't get it:

Still too late. Try checking it here:
https://github.com/wxWidgets/wxWidgets/blob/dac1996ebe162386071c0b9148079fabf3a950e3/src/gtk/window.cpp#L2574

I have concerns with using "draw":
- this handler will be called after the one that does the painting, here:
https://github.com/wxWidgets/wxWidgets/blob/dac1996ebe162386071c0b9148079fabf3a950e3/src/gtk/glcanvas.cpp#L22
- it's bad practice to do anything other than painting in a paint handler.
- this handler will get called for every paint, when (I think?) it's only needed once.

paulcor:
>at this time the scale factor is still 1

Then it must change at some point. Maybe "notify::scale-factor" could be used.

vadz:
> Then it must change at some point. Maybe "notify::scale-factor" could be used.

Yes, of course, thanks! I stupidly forgot about the existence of this notification again...

Done in #25994 now.

PR #25994, opened 2025-11-26
 Fix initial scale of wxGLCanvas under EGL/Wayland

vadz:
This replaces #25991 with a much simpler and better fix.

### DIFF
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -436,9 +436,17 @@ void wxEGLUpdatePosition(wxGLCanvasEGL* win)
     wl_subsurface_set_position(win->m_wlSubsurface, x, y);
 }
 
-// Helper declared as friend in the header and so can access m_wlSurface.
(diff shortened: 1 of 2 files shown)


vadz:
Oops, the first commit doesn't belong to this PR, sorry, but I'm leaving it here.

src/unix/glegl.cpp
```diff
@@ -497,12 +504,17 @@ static void gtk_glcanvas_size_callback(GtkWidget *widget,
...
 {
-    int scale = gtk_widget_get_scale_factor(widget);
-    wl_egl_window_resize(win->m_wlEGLWindow, win->m_width * scale,
-                         win->m_height * scale, 0, 0);
+    wxEGLUpdateGeometry(widget, win);
+}
 
-    wxEGLUpdatePosition(win);
-    wxEGLSetScale(win, scale);
+static gboolean
```
paulcor: Notify handler is void return type.

vadz: Fixed, thanks.

PR #25999, opened 2025-11-29
Improve wxListCtrl in MSW darkmode

MaartenBent:
Draw the background color outside the list-items area on all Windows version, both below and behind the items, and when the list is empty.

Set the text foreground and background colors when custom drawing. So the disabled list will use the correct colors in list or icon view.

Fixes #25990

### DIFF
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3410,6 +3410,14 @@ WXLPARAM HandleItemPrepaint(wxListCtrl* listctrl, LPNMLVCUSTOMDRAW pLVCD)
     }
 
     wxItemAttr* attr = listctrl->MSWGetItemColumnAttr(pLVCD->nmcd.dwItemSpec, pLVCD->iSubItem);
+
+    pLVCD->clrText = attr && attr->HasTextColour()
+        ? wxColourToRGB(attr->GetTextColour())
+        : wxColourToRGB(listctrl->GetTextColour());
+    pLVCD->clrTextBk = attr && attr->HasBackgroundColour()
+        ? wxColourToRGB(attr->GetBackgroundColour())
+        : wxColourToRGB(listctrl->GetBackgroundColour());
+
     if ( !attr )
     {
         // nothing to do for this item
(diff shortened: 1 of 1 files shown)


src/msw/listctrl.cpp
```diff
@@ -3471,16 +3476,16 @@ void wxListCtrl::OnPaint(wxPaintEvent& event)
     const bool drawVRules = HasFlag(wxLC_VRULES);
 
     // Check if we need to do anything ourselves: either draw the rules or, in
-    // case of using dark mode under Windows 11, erase the unwanted separator
+    // case of using dark mode under Windows, erase the unwanted separator
```
vadz: Very minor, but it doesn't make sense to say "under Windows" in this Windows-only code:

```suggestion
    // case of using dark mode, erase the unwanted separator
```

vadz (approved):
I didn't test it yet, but this looks good, so I'll try to test it and merge it soon, thank you!

src/msw/listctrl.cpp
MaartenBent: Agreed, can you update this when merging?

MaartenBent:
Found another issue, when hRules/vRules are enabled, we also have to erase the background.

MaartenBent:
And a regression of 4b2a57c0 (#25950). When resizing the column, and then moving the scrollbar right and left, it does not redraw correctly, investigating...

Screenshot 2025-11-30 000815

MaartenBent:
This is getting more and more changes, so testing is highly appreciated.

I added hot tracking in dark mode, but I don't really like the current colour. It is the closest colour I could find to the colour when hovering in the header. Maybe it should be more subtle/darker?

This change required using `Explorer` theme, to get all the custom draw calls. I don't see any side effects in Report view. In the other views items now have the bluish hover colour again, I could not find a way to use the same gray colour.
It is also possible to use `Explorer` in Report mode and `DarkMode_Explorer` in other modes, but I think having a bluish colour is better than having nothing.

vadz:
Sorry, I still didn't have time to test this. I'll try to do it, but if others could please test it, it would be great. Please comment here if you did!

vadz:
Thanks again, Maarten!

I didn't find any obvious regressions and there are clearly a lot of improvements, so I'm merging this.

PR #26002, opened 2025-11-30
Correct wxSpinEvent GetPosition() if start value is non-zero

RobertRoeb:
- Set wxSpinButton start value to start value of wxSpinCtrl, otherwise the generic control used in OSX would report a wrong value in EVT_SPIN
  - Very minor doc update
  - Limit explicit border to be around text control, not the spin button, which look very wrong

### DIFF
--- a/interface/wx/spinctrl.h
+++ b/interface/wx/spinctrl.h
@@ -244,7 +244,7 @@ public:
 
         @since 3.1.6
     */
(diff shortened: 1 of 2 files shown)


vadz:
Looks good, thanks!

vadz:
I've split this into several commits and merged in the commit mentioned above, thanks.

PR #26003, opened 2025-11-30
Allow images in higher resolution to be stored in own folder

RobertRoeb:
- The is an alternative used in other toolkits and I think it is convenient and does not harm

### DIFF
--- a/src/common/bmpbndl.cpp
+++ b/src/common/bmpbndl.cpp
@@ -529,6 +529,12 @@ wxBitmapBundle wxBitmapBundle::FromFiles(const wxString& path, const wxString& f
             fn.SetName(wxString::Format("%s_%dx", filename, dpiFactor));
         }
 
+        if ( !fn.FileExists() && dpiFactor != 1 )
+        {
+            // try /2.0x/image.png naming scheme
+            fn.AppendDir( wxString::Format("%d.0x", dpiFactor));
+        }
+
         if ( fn.FileExists() )
         {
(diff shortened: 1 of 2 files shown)


src/common/bmpbndl.cpp
```diff
@@ -529,6 +529,12 @@ wxBitmapBundle wxBitmapBundle::FromFiles(const wxString& path, const wxString& f
             fn.SetName(wxString::Format("%s_%dx", filename, dpiFactor));
         }
 
+        if ( !fn.FileExists() && dpiFactor != 1 )
```
vadz, line 532: This is minor, but we're now calling `FileExists()` twice completely unnecessarily when dpiFactor is 1. Let's use a temp var instead, i.e.

```cpp
bool found = fn.FileExists();
if ( !found && dpiFactor != 1 )
{
    fn.AppendDir(...);
    found = fn.FileExists();
}

if ( found )
```

vadz (commented):
This looks a bit strange to me, if I used subdirectories, I would have `1.0x` and `2.0x` and not `.` and `2.0x` as now, but if this is an established convention, I guess why not.

I also wonder if we need to add support for this to wxOSX, I think it was using "@2x" explicitly somewhere...

RobertRoeb:
For reference, here is a description:

.../my_icon.png       (mdpi baseline)
.../1.5x/my_icon.png  (hdpi)
.../2.0x/my_icon.png  (xhdpi)
.../3.0x/my_icon.png  (xxhdpi)
.../4.0x/my_icon.png  (xxxhdpi)

PR #26004, opened 2025-11-30
Fix to make tree buttons use the right colour on OS X

RobertRoeb:
- This may not be the right fix, but I am not sure if changing wxRendererNative is right and what the consequences would be of changing ApplyMacControlFlags() in wxRenderer. This change only affects the tree button on OSX and it now looks correct in light and dark mode.

### DIFF
--- a/src/generic/treectlg.cpp
+++ b/src/generic/treectlg.cpp
@@ -2870,7 +2870,7 @@ wxGenericTreeCtrl::PaintLevel(wxGenericTreeItem *item,
 #ifdef __WXMAC__
                 // this flag is required under OSX as this will make it
                 // choose the NSCell.highlighted mode. The chevron will
-                // otherwise be light grey on white background and be 
+                // otherwise be light grey on white background and be
                 // barely visible
                 flag |= wxCONTROL_PRESSED;
 #endif
--- a/src/generic/treectlg.cpp
+++ b/src/generic/treectlg.cpp
@@ -2867,14 +2867,6 @@ wxGenericTreeCtrl::PaintLevel(wxGenericTreeItem *item,
                 if (item == m_underMouse)
                     flag |= wxCONTROL_CURRENT;
 
-#ifdef __WXMAC__
-                // this flag is required under OSX as this will make it
-                // choose the NSCell.highlighted mode. The chevron will
-                // otherwise be light grey on white background and be
-                // barely visible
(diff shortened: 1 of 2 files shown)


vadz (changes requested):
It's a pity to have to do this because `wxRendererNative` was created specifically in order to avoid platform-specific code in generic controls.

If the result of `DrawTreeItemButton()` without this flag is invisible (in some mode only?), then I really think it should use it by default. This would also fix `wxPropertyGrid::DrawExpanderButton()` which is presumably broken in the same way currently.

RobertRoeb:
It is too bright in normal/light mode. Barely visible, but it is obviously wrong. There should be a cleaner way to alter wxRendererNative- that is what it us made for. But that code is hard to read. I don't know in which other cases ApplyMacControlFlags() is used. Maybe I can make it less ugly

vadz:
> It is too bright in normal/light mode. Barely visible, but it is obviously wrong.

So you're saying that `wxRendererMac::DrawTreeItemButton()` is wrong and hence needs to be fixed.

> There should be a cleaner way to alter wxRendererNative- that is what it us made for. But that code is hard to read. I don't know in which other cases ApplyMacControlFlags() is used. Maybe I can make it less ugly

Maybe, but in the meanwhile just doing

```diff
diff --git a/src/osx/cocoa/renderer.mm b/src/osx/cocoa/renderer.mm
index 666a0ac11a..2b385aac26 100644
--- a/src/osx/cocoa/renderer.mm
+++ b/src/osx/cocoa/renderer.mm
@@ -422,7 +422,7 @@ void wxRendererMac::DrawTreeItemButton( wxWindow *win,
     {
 #if wxOSX_USE_NSCELL_RENDERER
         NSControlStateValue stateValue = (flags & wxCONTROL_EXPANDED) ? NSControlStateValueOn : NSControlStateValueOff;
-        DrawMacCell(win, dc, GetDisclosureButtonCell(), rect, flags, stateValue);
+        DrawMacCell(win, dc, GetDisclosureButtonCell(), rect, flags | wxCONTROL_PRESSED, stateValue);
 #else
         // now the wxGCDC is using native transformations
         const wxCoord x = rect.x;
```

should be better than this PR AFAICS (with an appropriate comment explaining why we turn it on), shouldn't it?

RobertRoeb:
Yes, I just came up with the same solution and that works

PR #26005, opened 2025-11-30
Fix wxGrid overlay selection not refreshing correctly during drag-select

AliKet:
See:
https://github.com/user-attachments/assets/25410a0e-303b-4e44-b2b3-6a38df2c7ec5


vadz:
Thanks for fixing this!

PR #26006, opened 2025-12-01
GTK: Fix crash when deleting wxTextCtrl in TEXT_ENTER handler

RobertRoeb:
- Key handler collects events and sends a message in an "event-after" signal. This event arrives after deletion from a TEXT_ENTER  handler, leading to a crash.
  - Disconnect from that signal solves the issue
  - added test in text sample

### DIFF
--- a/include/wx/gtk/textentry.h
+++ b/include/wx/gtk/textentry.h
@@ -73,10 +73,6 @@ public:
     // Helper functions only used internally.
     wxTextCoalesceData* GTKGetCoalesceData() const { return m_coalesceData; }
 
-    // avoid event handling after deletion
-    void GTKTextEntrySetBeingDeleted() { m_isBeingDeleted = true; }
-    bool GTKTextEntryIsBeingDeleted() const { return m_isBeingDeleted; }
-
 protected:
     // This method must be called from the derived class Create() to connect
     // the handlers for the clipboard (cut/copy/paste) events.
@@ -148,7 +144,6 @@ private:
     mutable wxTextCoalesceData* m_coalesceData;
 
     bool m_isUpperCase;
-    bool m_isBeingDeleted;
 };
 
 // We don't need the generic version.
--- a/include/wx/gtk/textentry.h
+++ b/include/wx/gtk/textentry.h
@@ -73,6 +73,10 @@ public:
     // Helper functions only used internally.
     wxTextCoalesceData* GTKGetCoalesceData() const { return m_coalesceData; }
 
+    // avoid event handling after deletion
+    void GTKTextEntrySetBeingDeleted() { m_isBeingDeleted = true; }
+    bool GTKTextEntryIsBeingDeleted() const { return m_isBeingDeleted; }
+
 protected:
     // This method must be called from the derived class Create() to connect
     // the handlers for the clipboard (cut/copy/paste) events.
@@ -144,6 +148,7 @@ private:
     mutable wxTextCoalesceData* m_coalesceData;
 
     bool m_isUpperCase;
+    bool m_isBeingDeleted;
 };
 
 // We don't need the generic version.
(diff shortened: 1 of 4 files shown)


paulcor:
It would be appropriate to have a description here of how the crash can be reproduced. Maybe a stack trace too.

How about we do it like this instead?
```diff
diff --git a/src/gtk/textentry.cpp b/src/gtk/textentry.cpp
index 200d2e727f..b3d8f00cf8 100644
--- a/src/gtk/textentry.cpp
+++ b/src/gtk/textentry.cpp
@@ -43,6 +43,7 @@ class wxTextCoalesceData
 public:
     wxTextCoalesceData(GtkWidget* widget, gulong handlerAfterKeyPress)
         : m_handlerAfterKeyPress(handlerAfterKeyPress)
+        , m_widget(widget)
     {
         m_inKeyPress = false;
         m_pendingTextChanged = false;
@@ -52,12 +53,17 @@ public:
         g_signal_handler_block(widget, m_handlerAfterKeyPress);
     }
 
-    void StartHandlingKeyPress(GtkWidget* widget)
+    ~wxTextCoalesceData()
+    {
+        g_signal_handler_disconnect(m_widget, m_handlerAfterKeyPress);
+    }
+
+    void StartHandlingKeyPress()
     {
         m_inKeyPress = true;
         m_pendingTextChanged = false;
 
-        g_signal_handler_unblock(widget, m_handlerAfterKeyPress);
+        g_signal_handler_unblock(m_widget, m_handlerAfterKeyPress);
     }
 
     bool SetPendingIfInKeyPress()
@@ -70,9 +76,9 @@ public:
         return true;
     }
 
-    bool EndHandlingKeyPressAndCheckIfPending(GtkWidget* widget)
+    bool EndHandlingKeyPressAndCheckIfPending()
     {
-        g_signal_handler_block(widget, m_handlerAfterKeyPress);
+        g_signal_handler_block(m_widget, m_handlerAfterKeyPress);
 
         wxASSERT( m_inKeyPress );
         m_inKeyPress = false;
@@ -89,6 +95,7 @@ private:
     bool m_inKeyPress;
     bool m_pendingTextChanged;
     const gulong m_handlerAfterKeyPress;
+    GtkWidget* const m_widget;
 
     wxDECLARE_NO_COPY_CLASS(wxTextCoalesceData);
 };
@@ -120,14 +127,14 @@ extern "C" {
 // to send a single wxEVT_TEXT even if we received several (typically two, when
 // the selected text in the control is replaced by new text) "changed" signals.
 static gboolean
-wx_gtk_text_after_key_press(GtkWidget* widget,
+wx_gtk_text_after_key_press(GtkWidget*,
                             GdkEventKey* WXUNUSED(gdk_event),
                             wxTextEntry* entry)
 {
     wxTextCoalesceData* const data = entry->GTKGetCoalesceData();
     wxCHECK_MSG( data, FALSE, "must be non-null if this handler is called" );
 
-    if ( data->EndHandlingKeyPressAndCheckIfPending(widget) )
+    if ( data->EndHandlingKeyPressAndCheckIfPending() )
     {
         entry->GTKOnTextChanged();
     }
@@ -976,7 +983,7 @@ void wxTextEntry::GTKEntryOnKeypress(GtkWidget* widget) const
         m_coalesceData = new wxTextCoalesceData(widget, handler);
     }
 
-    m_coalesceData->StartHandlingKeyPress(widget);
+    m_coalesceData->StartHandlingKeyPress();
 }
 
 int wxTextEntry::GTKEntryIMFilterKeypress(GdkEventKey* event) const
```

RobertRoeb:
Thanks, that one works as well - and looks as if it would not break backward comp, so it could be backported.

RobertRoeb:
I have applied your patch and added a test to the text sample

vadz:
Sorry, I'm not going to apply the patch to the sample, this makes no sense to anybody else running it. This code really should be part of a unit test instead.

I will apply and backport the fix itself, thanks.

Issue #23986, opened 2023-10-22
wxAui: Allow minimization of docked panes; add minimized pane "status bar"


craftyjon:
### Description

It would be helpful to implement minimization of docked panes.  Minimization basically means the manager draws an additional “status bar” (or "notebook tab bar") in a docked pane area with buttons containing the title of each pane (with optional icon).  Panes can be “minimized” to this toolbar either by clicking the button or (in some implementations) by clicking a minimize button in the pane titlebar.  If no un-minimized panes remain, the docked pane area collapses to just the width of the toolbar.

Each docked pane space with such a toolbar can contain multiple panes that act as tabs in a tab control.  The minimization toolbar also acts as a tab bar to switch between panes.

Each pane can have a different size in the growable direction of the docked area.  For example, a vertical docked area (on the left or right of a main window) can have panes with different widths.

Example of this style of UI in JetBrains IDEs:

![image1](https://github.com/wxWidgets/wxWidgets/assets/116737/6df3035a-a7c8-4447-8b51-39d22ea3aa7a)

vadz:
I've finally (sorry for the very, very long delay...) started working on this and implemented something more or less functional:

<img width="1634" height="1286" alt="Image" src="https://github.com/user-attachments/assets/c8b46eda-6eab-4b6d-8664-e9be96f338f7" />

but I've realized that I have plenty of questions. The main one is probably whether we want to have multiple docking bars and, if so, how do we want to determine where should a particular window minimize: automatically, by choosing the dock closest to it, or by manually associating it with the given dock when adding it to AUI manager?

And if want to have just one dock, should it be created automatically, when the first window which can be minimized is added to the manager, or should we make this more explicit (which would also make customizing it, e.g. specifying where/how it should be positioned, more natural)?

There are also plenty of other, more minor questions:

- Do we want to always show the icons in this dock? As the screenshot shows, this doesn't look great when some panes have icons but others do not. Maybe we should use a placeholder icon for the latter?
- Do we want to rotate the icon, as I do currently?
- Do we want to always show full text which as, again, the screenshot shows, doesn't look great when it is long or maybe truncate/ellipsize it?
- Should floating AUI panes still minimize to this dock or should they minimize as normal top level windows?

@craftyjon (and anybody else interested in using this feature!) Please let me know what do you think, TIA!

craftyjon:
Thanks for working on this!

[...]

PR #26009, opened 2025-12-01
Add support for minimizing panes to wxAUI

vadz:
See #23986.

### DIFF
--- a/include/wx/aui/auibar.h
+++ b/include/wx/aui/auibar.h
@@ -108,46 +108,13 @@ class WXDLLIMPEXP_AUI wxAuiToolBarItem
     friend class wxAuiToolBar;
 
 public:
+    wxAuiToolBarItem() = default;
 
-    wxAuiToolBarItem()
-    {
-        m_window = nullptr;
-        m_sizerItem = nullptr;
-        m_spacerPixels = 0;
-        m_toolId = 0;
-        m_kind = wxITEM_NORMAL;
-        m_state = 0;  // normal, enabled
-        m_proportion = 0;
-        m_active = true;
-        m_dropDown = true;
-        m_sticky = true;
-        m_userData = 0;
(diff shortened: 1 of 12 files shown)


AliKet:
Thanks for implementing this.
I have no real comments, just an observation after testing this with `wxGTK3` and `wxQt`: 
If I resize a pane and then minimize it, it will reset to its default value after being restored. This behavior differs from that observed when it is maximized and then restored.

vadz:
Thanks for noticing this, fixed now.

vadz:
There has been no other comments, so I'm thinking about merging it and then improving it further in subsequent PRs later. I'll wait for a few more days just in case anybody notices something that really should not be merged in this one.

MaartenBent:
Works ok in the aui sample on Windows.
Is there a way to get rid of the special toolbar once you have minimized a pane? I can make it go away when undocking both panes and then redocking them. But maybe it can be shown only if one ore more panes are minimized, and hiden once all are maximized?

vadz:
> Is there a way to get rid of the special toolbar once you have minimized a pane?

This was one of the questions I asked in #23986. On one hand, this would make sense because currently we have a hysteresis effect: minimizing and restoring a pane bring the state to the original one, but the appearance is different, and I don't like this. OTOH it looks rather weird when restoring a single pane makes the dock containing multiple panes disappear. This wouldn't be a problem if we showed only minimized panes in the dock, but we want to (at least optionally) show all of then, so we can't avoid this.

I'm still not sure what is better/less worse here (I probably have a very slight preference towards removing the dock when there are no more minimized panes) and I don't think @craftyjon (who had originally requested this feature) has a definite opinion about this either.

vadz:
I don't think it makes sense to wait for much longer, so I'll merge this tomorrow if nobody objects.

PR #26012, opened 2025-12-04
Fix creation of ES2 context with wxUSE_GLCANVAS_EGL defined

timothytylee:
Two changes inside **src/unix/glegl.cpp** are required to allow the creation of OpenGL ES contexts:
1. Invoking `eglBindAPI()` with either `EGL_OPENGL_API` or `EGL_OPENGL_ES_API` depending on the required type of context. See https://registry.khronos.org/EGL/sdk/docs/man/html/eglBindAPI.xhtml
2. Filter out the `EGL_RENDERABLE_TYPE` attribute when creating the context. This is an invalid attribute for `eglCreateContext()` that causes the context creation to fail. See https://registry.khronos.org/EGL/sdk/docs/man/html/eglCreateContext.xhtml

### DIFF
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -300,14 +300,40 @@ wxGLContext::wxGLContext(wxGLCanvas *win,
         contextAttribs = win->GetGLCTXAttrs().GetGLAttrs();
     }
 
+    // Check which rendering API is required, and filter the attribute list
+    EGLenum api = EGL_OPENGL_API;
+    wxVector<int> attribs;
+    for ( int i = 0; contextAttribs[i] != EGL_NONE; i += 2 )
+    {
+        if ( contextAttribs[i] == EGL_RENDERABLE_TYPE )
+        {
+            if ( contextAttribs[i + 1] == EGL_OPENGL_ES2_BIT )
+            {
+                api = EGL_OPENGL_ES_API;
+            }
+        }
+        else
+        {
+            attribs.push_back(contextAttribs[i]);
+            attribs.push_back(contextAttribs[i + 1]);
+        }
+    }
+    attribs.push_back(EGL_NONE);
+
     m_isOk = false;
+
+    if ( !eglBindAPI(api) )
+    {
+        wxFAIL_MSG("eglBindAPI failed");
+        return;
+    }
 
     EGLConfig fbc = win->GetEGLConfig();
     wxCHECK_RET( fbc, "Invalid EGLConfig for OpenGL" );
 
     m_glContext = eglCreateContext(wxGLCanvasEGL::GetDisplay(), fbc,
                                    other ? other->m_glContext : EGL_NO_CONTEXT,
-                                   contextAttribs);
+                                   &*attribs.begin());
 
     if ( !m_glContext )
         wxLogMessage(_("Couldn't create OpenGL context"));


vadz (commented):
Thanks, reading the documentation confirms that your points are correct, but I think that it would be better to fix this by reimplementing `wxGLContextAttrs::ES2()` and notably avoid using `EGL_RENDERABLE_TYPE` attribute in the context attributes list, where it is not allowed, as this seems just wrong (or am I missing something?). AFAICS we should add a field to `wxGLContextAttrs` and store the requested API in it directly and then use it to call `eglBindAPI()` in EGL code.

OTOH I wonder if we need to add `wxGLAttributes::ES2()` to set this attribute where it is allowed? Or is it not necessary/useful?

Also, how can I test using ES? Can our pyramid sample be used to do it?

cc @swt2c

timothytylee:
> Thanks, reading the documentation confirms that your points are correct, but I think that it would be better to fix this by reimplementing `wxGLContextAttrs::ES2()` and notably avoid using `EGL_RENDERABLE_TYPE` attribute in the context attributes list, where it is not allowed, as this seems just wrong (or am I missing something?). AFAICS we should add a field to `wxGLContextAttrs` and store the requested API in it directly and then use it to call `eglBindAPI()` in EGL code.
> 
> OTOH I wonder if we need to add `wxGLAttributes::ES2()` to set this attribute where it is allowed? Or is it not necessary/useful?

I wrote the fix hoping to keep the existing ABI, which is shared by X11, OSX, MSW and QT implementations. However a quick look at the relevant source files suggest that ES2 context support is broken for OSX, QT.

> Also, how can I test using ES? Can our pyramid sample be used to do it?

ES2 uses a different shader language, and offers a reduced set of `glXXX()` functions. The existing pyramid sample would need to be restructured to support ES2 context.

I've attached a minimal test program for testing ES2 support:
[minimal-es2.cpp](https://github.com/user-attachments/files/23953985/minimal-es2.cpp)

vadz:
> I wrote the fix hoping to keep the existing ABI, which is shared by X11, OSX, MSW and QT implementations.

We don't really care about ABI in master, so I've created non ABI-preserving but better (IMO) #26015 doing what I proposed. Do you see any problems with it? If so, please comment there, but I'll close this one because as the newer PR replaces it.

Also, I still don't know if we need an attribute corresponding to `EGL_RENDERABLE_TYPE` in our API or if it's not important?

> However a quick look at the relevant source files suggest that ES2 context support is broken for OSX, QT.

Unfortunately I don't really have time/knowledge to fix these ports, but any fixes to them would be welcome too, of course!
 
> > Also, how can I test using ES? Can our pyramid sample be used to do it?
> 
> ES2 uses a different shader language, and offers a reduced set of `glXXX()` functions. The existing pyramid sample would need to be restructured to support ES2 context.
> 
> I've attached a minimal test program for testing ES2 support: [minimal-es2.cpp](https://github.com/user-attachments/files/23953985/minimal-es2.cpp)

Thanks, this was very useful, I could check that the expected triangle appeared with my changes too. BTW, I've also tried removing `.ES2()` to see what will happen and got a crash (in a worker thread linking shaders if it matters). This is not the most graceful error mode, but at least it definitely proves that ES support works.

Issue #26014, opened 2025-12-06
UseNativeColHeader() results in blank column labels when wxGrid is wider than window


jey5nd6:
### Description
`wxGrid::UseNativeColHeader( true )` results in blank column labels when `wxGrid` is wider than window.

Potentially MacOS-only.

Ref KiCad bug: https://gitlab.com/kicad/code/kicad/-/issues/18198

#### To Reproduce:

Add the line:

```
    grid->UseNativeColHeader( true );
```

after:

```
    grid->CreateGrid( 0, 0 );
```

in `griddemo.cpp`.  (Should be around line 598.)

Build the sample.

Scroll the grid to the left.  Note that newly-exposed columns on the right have blank column headers.

### Platform and version information

- wxWidgets version  you use: 3.2.8
- wxWidgets port  you use: wxOSX
- OS  and its version: macOS Sonoma Version 14.6.1

```
Application: KiCad Schematic Editor arm64 on arm64

Version: 9.99.0-4297-gf7d04c29ea-dirty, debug build

Libraries:
	wxWidgets 3.2.8
	FreeType 2.13.3
	HarfBuzz 11.0.1
	FontConfig 2.14.0
	libcurl/8.7.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.12 nghttp2/1.61.0

Platform: macOS Sonoma Version 14.6.1 (Build 23G93), 64 bit, Little endian, wxMac
OpenGL: Apple, Apple M1 Max, 2.1 Metal - 88.1

Build Info:
	Date: Dec  6 2025 13:00:29
	wxWidgets: 3.2.8 (wchar_t,wx containers)
	Boost: 1.88.0
	OCC: 7.9.0
	Curl: 8.7.1
	ngspice: 44.2
	Compiler: Clang 16.0.0 with C++ ABI 1002

Build settings:
	KICAD_IPC_API=ON
	KICAD_USE_PCH=OFF
	KICAD_STDLIB_DEBUG=OFF
	KICAD_STDLIB_LIGHT_DEBUG=OFF
	KICAD_SANITIZE_ADDRESS=OFF
	KICAD_SANITIZE_THREADS=OFF
```

vadz:
Unfortunately this is indeed Mac-specific, this diff

```diff
diff --git a/samples/grid/griddemo.cpp b/samples/grid/griddemo.cpp
index a93defedf0..0ee5f63421 100644
--- a/samples/grid/griddemo.cpp
[...]

PR #26016, opened 2025-12-06
Generic `wxHeaderCtrl` fixes

vadz:
This is the result of debugging and "fixing" (working around, actually) #26014.

@csomor I wasn't able to understand why isn't the device origin respected in `wxRendererMac::DrawMacHeaderCell()`. I've checked that it uses the same `CGContext` for which we had called `wxMacCoreGraphicsContext::SetTransform()` with the translation transform before, yet somehow it is just not taken into account in this code.

### DIFF
--- a/src/generic/headerctrlg.cpp
+++ b/src/generic/headerctrlg.cpp
@@ -528,9 +528,8 @@ void wxHeaderCtrl::OnPaint(wxPaintEvent& WXUNUSED(event))
     // account for the horizontal scrollbar offset in the parent window
     dc.SetDeviceOrigin(m_scrollOffset, 0);
 
-    const unsigned int count = m_numColumns;
     int xpos = 0;
-    for ( unsigned int i = 0; i < count; i++ )
+    for ( unsigned int i = 0; i < m_numColumns; i++ )
     {
         const unsigned idx = m_colIndices[i];
         const wxHeaderColumn& col = GetColumn(idx);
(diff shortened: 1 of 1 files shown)


csomor:
@vadz I'll look at it if possible tomorrow

vadz:
Thanks, there is nothing urgent about it and my workaround works fine, but I just wonder if you immediately see something obviously wrong with using `wxDC::SetDeviceOrigin()` and then using native functions on the underlying graphics context because this could be a source of other bugs elsewhere...

csomor:
it's not the offset, but a clipping problem, if you scroll and make the window large enough you see some additional headers being properly drawn, but only as long as they are part of the "original" clipping area, so it's as if somewhere an additional structure exists corresponding to the part to be redrawn, and this one is not in device coordinates but got transformed as well ...

csomor:
When I undo the device origin offsetting and instead move the native NSCell drawing frame accordingly, then things are drawn correctly, I'll have to think about this a little more

ie after this `CGContextSaveGState(cgContext);` in renderer.mm line 763, I added
```
        wxPoint offset = dc.GetDeviceOrigin();
        CGContextTranslateCTM( cgContext, -offset.x, -offset.y );
        controlRect = NSOffsetRect(controlRect, offset.x, offset.y);
```

vadz:
> it's not the offset, but a clipping problem,

Ah, I should have realized it (because "I" column was still visible after scrolling), thanks.

As I said, I think we can live with the new version of the code avoiding clipping in `wxHeaderCtrl` (it's more efficient, too, even if this could be done while still setting the device origin, of course), but it would probably still be good to adjust the clipping region when it the device origin is changed — if this is even possible?

csomor:
I've written a PR #26018 with this change

PR #26018, opened 2025-12-07
OSX NSCell Rendering (native renderer) fix

csomor:
The NSCell drawing routines seem to use a different clipping area than the one of the current context. This workaround undoes the offsetting of the dc but then moves the drawing frame accordingly, see #26014


vadz:
I'll merge this into master but do you think this should be included in 3.2.9 too (as at least its source archives will apparently need to be redone anyhow because of the other issues found in it) or is the risk of breaking something too great?

csomor:
> I'll merge this into master but do you think this should be included in 3.2.9 too (as at least its source archives will apparently need to be redone anyhow because of the other issues found in it) or is the risk of breaking something too great?

it's very tempting, to get the coverage, but I fear it might have some consequences which I haven't thought of ...

vadz:
Thanks, let's see if we run into any problems after doing this in master first then and backport this in 3.2.10 if none are found.

Issue #26017, opened 2025-12-07
macOS wxFontPicker not returning bold/italic/underline/etc...


dkulp:
#### Bug description:

With macOS 26 (not sure about other versions), the wxFontPicker is not picking up ANY of the other attributes other than base font name and size.  Bold, Italic, Underline, etc... are not coming through.

#### Expected vs observed behaviour:

Expect that the font that is picked is actually returned and is queryable with the properties the user selected.

#### To Reproduce:

This is easy to reproduce in the "widgets" sample.   Open the widgets same, select the FontPicker, set the font to something like "Arial Bold 24" and all that is returned is just "Arial 24".

NOTE:  a possible separate bug:  in the font picker that pops up, if you click on the button to enable the Typographical panel, the widgets app completely crashes.

### Platform and version information

- wxWidgets version  you use: close to latest master
- wxWidgets port  wxOSX
- OS  macOS 26

vadz:
This looks bad, I'll try to have a look but, thanks for reporting.

vadz:
Sorry, I forgot that I still can't really debug this because I still haven't updated to macOS 26.

@csomor Might I ask for your help again here?

csomor:
@vadz I'll look at it

csomor:
@dkulp  Thanks for your report!

Just to make sure we are talking about the same thing: 

- in the widgets sample you click on the left in the pickers section 
- take the font picker and then click on the button 
- then you select 
   - font Arial 
   - variant Bold and 
   - size 24 ? 

the log information currently does not log the numeric weight at all, nor the style, so it should be logged, with something like

```
    wxLogMessage("The font changed to '%s' with size %d, style %s, weight %d !",
                 event.GetFont().GetFaceName(), event.GetFont().GetPointSize(),
                 event.GetFont().GetStyle() == wxFONTSTYLE_NORMAL ? "regular" : "italic",
                 event.GetFont().GetNumericWeight());
```

Now on Tahoe the selection of the variant always jumps back to Standard, I'll try to find out what events we must support here... I do not get a crash with the typography panel, but as we try to get a modal behavior this does not fit in right now.

dkulp:
Using your log message, I'm ALWAYS getting style regular and weight 400 no matter if I select italic or bold or not.

csomor:
> Using your log message, I'm ALWAYS getting style regular and weight 400 no matter if I select italic or bold or not.

[...]

PR #26022, opened 2025-12-09
OSX change wxFontPicker implementation 

csomor:
macOS 26 Tahoe breaks our current implementation, see #26017


vadz:
Thanks, I'll merge this soon, but just a question: does this also fix the crash mentioned in the issue when switching to "Typographical" page?

csomor:
> Thanks, I'll merge this soon, but just a question: does this also fix the crash mentioned in the issue when switching to "Typographical" page?

I couldn't reproduce the crash, but I'm on RC 2 of 26.2, so perhaps we look at a reproducible way once 26.2 is out ?

vadz:
Thanks, I'm merging this into master and I think this will also need to be backport into 3.2 for 3.2.10, right?

csomor:
@vadz yes, this should go into 3.2.10 as well, thanks

PR #26023, opened 2025-12-10
Make it possible to choose during run-time between using GLX and EGL

vadz:
We've discussed this several times in the past and I've finally implemented the necessary refactoring to allow selecting between GLX and EGL at run-time, e.g. to allow users having problems with EGL to try using GLX (which is only possible when also using X11, of course, so this is probably much less useful now that Wayland is much more widespread than when we started talking about doing this, but better late than never, isn't it).

I'd like to clean up a couple of things here but I think this can already be tested and any such testing before I merge this would be really appreciated!

cc @dsa-t @swt2c


vadz:
This should be good to merge now, AFAICS. Please let me know if anybody sees any problems with it!

swt2c:
Thanks a lot for doing this!  Will try to take a deeper look later today.

PR #26025, opened 2025-12-10
Fix wxButton::SetBitmap() to keep correct button style in OS X

RobertRoeb:
- Before, when setting a new bitmap (e.g. from Play bitmap to Pause bitmap) the button would be recreated in pure text button style, much too narrow.
  - Then secondly, when taking into account the bitmap, it switch from slightly rounded to unrounded corners. I assume that the initial look with slightly rounded corners was intented (and intended to be kept) and now it is preserved after setting a new bitmap

### DIFF
--- a/src/osx/cocoa/button.mm
+++ b/src/osx/cocoa/button.mm
@@ -93,8 +93,7 @@ void wxButtonCocoaImpl::SetBitmap(const wxBitmapBundle& bitmap)
     // switch bezel style for plain pushbuttons
     if ( bitmap.IsOk() )
     {
-        if ([GetNSButton() bezelStyle] == NSRoundedBezelStyle)
-            [GetNSButton() setBezelStyle:NSRegularSquareBezelStyle];
+        [GetNSButton() setBezelStyle:NSRegularSquareBezelStyle];
     }
     else
     {
@@ -226,7 +225,7 @@ SetBezelStyleFromBorderFlags(NSButton *v,
             case wxBORDER_RAISED:
             case wxBORDER_THEME:
                 bezel = isSimpleText ? NSRoundedBezelStyle
-                                     : NSSmallSquareBezelStyle;
+                                     : NSRegularSquareBezelStyle;
                 break;
         }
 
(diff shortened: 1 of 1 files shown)


src/osx/cocoa/button.mm
```diff
@@ -93,8 +93,7 @@ - (NSControlSize)controlSize;
     // switch bezel style for plain pushbuttons
     if ( bitmap.IsOk() )
     {
-        if ([GetNSButton() bezelStyle] == NSRoundedBezelStyle)
-            [GetNSButton() setBezelStyle:NSRegularSquareBezelStyle];
+        [GetNSButton() setBezelStyle:NSRegularSquareBezelStyle];
```
vadz: Should we perhaps use `SetBezelStyleFromBorderFlags()` here too to ensure that the bezel is always set consistently?

vadz (commented):
Thanks, I didn't test this but from reading the code it looks like it should do the right thing. @csomor Do you have any objections to merging this?

csomor:
@RobertRoeb Thanks! I haven't fully understood the sequence yet, where would it be recreated with pure text, I thought that SetBitmap() would call the implementation's SetBitmap without first re-creating something. But since it happens to you, then I understand that unconditionally changing the style make sense. But I'd also prefer to call the common `SetBezelStyleFromBorderFlags()` to keep all our HIG wisdom at one place and consistent, would just calling this instead of manipulating the style directly also solve your problem ?

According to the docs [`NSRegularSquareBezelStyle` and `NSSmallSquareBezelStyle`](https://developer.apple.com/documentation/appkit/nsregularsquarebezelstyle) are perfect for bitmaps and other cases because they can scale to any size, while [NSRoundedBezelStyle a.k.a NSBezelStylePush](https://developer.apple.com/documentation/appkit/nsroundedbezelstyle) are for text content, so I think it's decision tree is still valid. Otherwise I'd prefer to fix it there.

RobertRoeb:
Yes, that works as well. While testing that, I noticed  that when changing the button from "play" to "pause", since "play" is shorter, "pause" gets wrapped. So can we invalidate the size calculation somehow?
[Edit myself]: I realize that I was under the assumption that the control would resize itself after setting a new, longer label and DoGetBestSize() correctly reported the new size. But I actually had to call the parent window's wxSizer::Layout() for the button to change size. This, sort of, makes sense, but I anticipated an immediate change of the button size. Anyways.

vadz:
@csomor Stefan, should this be merged as well?

csomor:
> @csomor Stefan, should this be merged as well?

@vadz yes, please

PR #26026, opened 2025-12-10
wxDialogValidateEvent as an alternative to Validate()

RobertRoeb:
- Replaces identical PR that got mixed with other commits
   - adds documentation
   - This adds the above mentioend event that is sent after pressing OK or APPLY to validate the dialog data and or transfer its data
   - Derived from wxNotifyEvent to allow veto
   - It is the sister event of wxInitDialog that can be used to transfer data to the dialog
   - This event is helpful if deriving from wxDialog and overriding wxWindow::Validate() is not (easily) possible and it happens to be useful for language bindings

### DIFF
--- a/include/wx/event.h
+++ b/include/wx/event.h
@@ -689,6 +689,7 @@ class WXDLLIMPEXP_FWD_CORE wxPaletteChangedEvent;
 class WXDLLIMPEXP_FWD_CORE wxJoystickEvent;
 class WXDLLIMPEXP_FWD_CORE wxDropFilesEvent;
 class WXDLLIMPEXP_FWD_CORE wxInitDialogEvent;
+class WXDLLIMPEXP_FWD_CORE wxDialogValidateEvent;
 class WXDLLIMPEXP_FWD_CORE wxUpdateUIEvent;
 class WXDLLIMPEXP_FWD_CORE wxClipboardTextEvent;
 class WXDLLIMPEXP_FWD_CORE wxHelpEvent;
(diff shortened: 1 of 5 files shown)


RobertRoeb:
Discussion from the other PR was:

> Either way, as I wrote, we have the validators and I would really like to know if they actually are used.

Yes, extensively.

> There is also wxInitDialogEvent - not sure when you would need that instead of just adding data during dialog creation,

I don't know about this one either.

> I propose an event that would get sent across native and generic dialogs when pressing OK and or some other affirmative > button. And events are easier for language bindings than overriding virtual functions across languages. wxWidgets does
> almost everything by events. It should do this as well, I think.

Language bindings are a good argument, but in C++ overriding virtual functions is much more convenient. E.g. it allows to call the base class member first, which may be important, while with the events you can only process the event in the derived class and skip it to let the base class have it later.

Anyhow, my fear is that by adding events to the existing mix of virtual functions and validators we risk making things even more confusing as there will be I don't know how many different ways of handling this in a derived class (e.g. "process the event, then skip it to let the base class function use validators for the inherited controls"). Other than language bindings I really don't see a good reason to complicate things further here. I wonder how does wxPython and co handle this now, i.e. do they provide some way of overriding TransferData{To,From}Window() or a replacement mechanism?

vadz:
Sorry, I don't have much to add here: I think it's a bad idea to add even more ways to handle validation in the dialogs, so I wouldn't apply this. But I'm ready to be overruled if others think it's worth having — so please comment here if you find this useful and would like to have it in wx. Thanks!

oneeyeman1:
@vadz ,
Why not deprecate Validate() and use event?

Thank you.

RobertRoeb:
Thank you for your support, but that would break more or less all existing applications, so that was not the intention. I think that adding this event is right because you can do everything with events when handling dialogs, but validation is an exception that bit me.

vadz:
One month later, I think it's time to close this one.

PR #26027, opened 2025-12-11
Don't hang handling wxEVT_CHAR_HOOK while IME is used in wxMSW

### LINKED ISSUE #22473: Freeze with Microsoft IME input (japanese/chinese character input)
**Describe the bug**
Application freezes with Microsoft IME Chinese / Japanese.
Issue can be worked around by using IME compatibility mode, or using an old enough version of IME.

Not freezing

Steps to reproduce in KiCad: https://gitlab.com/kicad/code/kicad/-/issues/11502#note_934148979
Steps to reproduce in Audacity: https://github.com/audacity/audacity/issues/805

 - wxWidgets version you use: Bug seen in the wild with wxMSW 3.1.3, 3.1.5 and 3.1.6
 - OS and its version:  Observed on Windows 8.1, 10, and 11.

vadz:
Not skipping the event in wxEVT_CHAR_HOOK event handler could completely freeze the program due to IME getting in some broken state. Avoid this by ensuring we always let the message hook it installs have the keyboard messages when it is active.

Closes #22473.

### DIFF
--- a/src/msw/window.cpp
+++ b/src/msw/window.cpp
@@ -103,6 +103,7 @@
 
 #include <string.h>
 
+#include <imm.h>
 #include <shellapi.h>
 #include <mmsystem.h>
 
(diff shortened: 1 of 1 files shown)


vadz:
I'm merging this in order to have a chance of it being tested before backporting to 3.2.

PR #26029, opened 2025-12-13
wxTreeCtrl chevron/triangle colour and behaviour fix generic control

RobertRoeb:
- The chevron is now drawn in the correct colour if the item is selected on GTK+
- The click area is now bigger, not just the triangle/chevron itself, as in the native controls. I believe this is the right thing to do on all ports, IMO, but I tested GTK+ and OSX


vadz:
Thanks, the fixes look good and I'll merge them soon but I'll break them in 2 commits corresponding to the 2 bullet points above.

Also, Robert, could you please set up a local Git hook that would notify you about the trailing whitespace? It's a bit tiresome to have to remove it manually every time. To do this it's enough to create `.git/hooks/pre-commit` containing this:

```sh
#!/bin/sh
exec git diff-index --check --cached HEAD --
```

and run

```console
$ git config core.whitespace blank-at-eol,space-before-tab,cr-at-eol
```

PR #26030, opened 2025-12-13
Make the progressbar in wxDataViewCtrl thin in GTK+

RobertRoeb:
- This is slightly more in line with look of the native progress bar

### DIFF
--- a/src/gtk/renderer.cpp
+++ b/src/gtk/renderer.cpp
@@ -65,6 +65,8 @@ public:
 
     virtual int GetHeaderButtonMargin(wxWindow *win) override;
 
+    virtual void DrawGauge(wxWindow* win, wxDC& dc, const wxRect& rect, int value, int max, int flags = 0) override;
+
 
     // draw the expanded/collapsed icon for a tree control item
     virtual void DrawTreeItemButton(wxWindow *win,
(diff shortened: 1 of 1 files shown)


src/gtk/renderer.cpp
```diff
@@ -200,6 +202,24 @@ static GdkWindow* wxGetGTKDrawable(wxDC& dc)
 }
 #endif
 
+// ----------------------------------------------------------------------------
+// progress bar in wxDataViewCtrl
```
vadz, line 206: This is not really related to wxDVC.

vadz (commented):
Thanks, this is probably better than nothing but isn't there a way to draw it using GTK theme? As long as we're changing this, it would be nice to make it look good rather than just a bit better...

src/gtk/renderer.cpp
RobertRoeb: It is, to my knowledge, only used in wxDVC

vadz: In wx, yes. But it's a public function which can be used for whatever the application wants.

vadz:
Thinking more about this, I'm really not sure we want to hardcode 8 as height of all progress bars, so I'm not going to apply this, sorry. If we really need to do this, let's add a function returning the native progress bar width (which could be hardcoded as 8 if there is really no better way) and use it in wxDVC code, but we shouldn't be changing the behaviour of any applications using this function.

Issue #24520, opened 2024-05-06
Request to merge a set of fix and improvements (done for our application)


JCYang:
Hi, there.
We've developed a commercial cross-platform app using wxWidgets framework for windows/mac/linux, all three mainstream OS are targeted.
And it's a non-trivial app that utilize many features of the framework, we may be one of the most heavy users.
So on the way, we found so many bugs.
We were on rapid development cycles and didn't have the resources to share our fixes to the upstream project on a regular basis, but now we find that we can no longer maintain the fixes, considering the framework is far from mature enough to allow us to stay on one LTS like version, we'd like to request you, the project maintainers to merge in most of our fixes/improvements.

Because there's too many, and in the foreseeable future, more fixes are expected and coming, we'd like to unify the set of works here for the maintainers to review.

Each of the fixes are now in their own branch, created for the sake of the pull request.

1. Fix wxRibbonButtonBar's layout overall size computation.
https://github.com/JCYang/wxWidgets/commit/a28b8db720cba565c3a5143462832f419456c5d8
2.  wxGTK: Fix UI refreshing delay when dragging the header border of generic wxDataViewCtrl to resize column.
https://github.com/JCYang/wxWidgets/commit/381c608eaf7f432429165bb24e802ba8c0903857
3. wxGTK: Fix improper wxEVT_LEAVE_WINDOW event 
https://github.com/JCYang/wxWidgets/commit/92868899c833261c728d3edf191031de71e04b3a
4. MacOSX: Change the two dummy messages to use systemUpTime as timestamp to fix CEF renderer process crash.
https://github.com/JCYang/wxWidgets/commit/af1f103cf8d10fa43e77da8d16f8091ddb6cacbb
5.  MacOSX: Keep DoDragDrop() from re-entering to enable a proper working CEF.
https://github.com/JCYang/wxWidgets/commit/e998db7cea963d9f2cc4d19057e48a87a01a3973
6. wxOSX: Fix wxPopupTransientWindow's mouse event handling. 
https://github.com/JCYang/wxWidgets/commit/8d2ddec05c3149f60f46ba303d9cec6b3843fa43
7. MacOSX: Make wxWebviewWebkit capture hotkeys less eagerly.
https://github.com/JCYang/wxWidgets/commit/f4eb6c3b041056355b41f634dc4c055860f0ff7a
8. wxOSX: Add support for wxDropSource icons.
[...]

PR #26032, opened 2025-12-14
OSX CEF Fix Support DnD Cursor

csomor:
separate piece from #24520


csomor:
This is PR is a work in progress, it needs some additional reflection and verification, created just to make sure it does not get lost.

PR #26033, opened 2025-12-14
Document wxObjectRefData as a typedef (vs class)

swt2c:
While this is actually more true to reality, this also helps work around an issue with Doxygen 1.15.0 (and wxPython builds).  See: https://github.com/doxygen/doxygen/issues/11889

### DIFF
--- a/interface/wx/object.h
+++ b/interface/wx/object.h
@@ -5,7 +5,7 @@
 // Licence:     wxWindows licence
 /////////////////////////////////////////////////////////////////////////////
 
-/** @class wxObjectRefData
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, will push soon.

Sorry you had to go bug hunting in Doxygen :-/

vadz:
Oh, and I guess this should be cherry-picked to 3.2 too, so I'll do that as well.

swt2c:
> Thanks, will push soon.
> 
> Sorry you had to go bug hunting in Doxygen :-/

Thanks, unfortunately it does happen from time-to-time.

PR #26035, opened 2025-12-16
OSX: choose grey tone for selection background in dark mode

RobertRoeb:
- After hours of testing all available NSColor styles I could not find one that would work in my test cases and/or  is available on all supported version of OSX.
  - 110,110,110 is well readable on super dark grey of e.g. wxListBox and less grey used in other cases, without being too bright like the existing grey

### DIFF
--- a/src/osx/cocoa/renderer.mm
+++ b/src/osx/cocoa/renderer.mm
@@ -587,7 +587,7 @@ wxRendererMac::DrawItemSelectionRect(wxWindow * WXUNUSED(win),
     else
 #endif
     {
-        col = wxColour( wxMacCreateCGColorFromHITheme( (flags & wxCONTROL_FOCUSED) 
+        col = wxColour( wxMacCreateCGColorFromHITheme( (flags & wxCONTROL_FOCUSED)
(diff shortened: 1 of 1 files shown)


RobertRoeb:
I have now added OSX 14 code using NSColor (which adapts to dark mode itself) and the mix of wxMacCreateCGColorFromHITheme and hardcoded dark mode.

csomor:
> I have now added OSX 14 code using NSColor (which adapts to dark mode itself) and the mix of wxMacCreateCGColorFromHITheme and hardcoded dark mode.

Thanks a lot @RobertRoeb

csomor:
@RobertRoeb a question: the constants are available starting from macOS 10.14+ ie Mojave but according to your tests, only macos 14+ ie Sonoma need ? or support ? this solution ?

RobertRoeb:
I may have misread the docs. I tested the "old" code by removing the new one. Feel free to correct the runtime and compile test version.

vadz:
@csomor Is this ok to merge?

csomor:
@vadz yes, please, I'll change the system versions at a later date, too many things to do right now ...

PR #26036, opened 2025-12-16
OSX: allow setting GL context on any thread

Hydr8gon:
I'm working on an OpenGL renderer for my wxWidgets-based project [3Beans](https://github.com/Hydr8gon/3Beans), and I need to maintain a GL context that is potentially used across different threads. The addition of `wxGLContext::ClearCurrent()` allowed this to work flawlessly on Linux and Windows, but on macOS it was impossible to set the GL context on a non-UI thread. It crashes with the error `[NSOpenGLContext setView:] must be called from the main thread`. This change allows the context to be set without touching the UI if it's off the main thread, which makes threaded OpenGL work perfectly on macOS as well.

### DIFF
--- a/src/osx/cocoa/glcanvas.mm
+++ b/src/osx/cocoa/glcanvas.mm
@@ -199,8 +199,12 @@ bool wxGLContext::SetCurrent(const wxGLCanvas& win) const
     if ( !m_glContext )
         return false;
 
-    [m_glContext setView: win.GetHandle() ];
-    [m_glContext update];
+    // View can only be set on the UI thread, but context can be set on any thread
+    if (wxIsMainThread())
+    {
+        [m_glContext setView: win.GetHandle() ];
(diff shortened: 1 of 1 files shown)


vadz:
I am not sure if it's really safe to use `wxGLContext` from non-UI thread(s), if it is, we need to document what thread-safety guarantees exactly does it provide (but I don't know what they are myself, unfortunately) and add an example of doing this to one of the OpenGL samples.

Of course, adding this check shouldn't do any harm, but I just worry about making promises that we won't be able to keep...

Hydr8gon:
From what I've gathered online, GL contexts should be thread-safe as long as they're cleared from any other thread before being set on a new one. Of course, that doesn't account for any interactions a wxGLContext might have with a wxGLCanvas. Another solution might be to separate context and canvas setting into different functions.

PR #26042, opened 2025-12-20
Fix wxMkDir call by adding missing mode_t argument

Proshanto-1234:
Updated wxMkDir call to include mode_t parameter for directory creation, which fixes error:
src/common/filefn.cpp:663:17: error: too few arguments to function ‘int wxMkDir(const wxString&, mode_t)’

### DIFF
--- a/src/common/filefn.cpp
+++ b/src/common/filefn.cpp
@@ -660,7 +660,18 @@ bool wxMkdir(const wxString& dir, int perm)
   #endif
 #else  // MSW and VC++
     wxUnusedVar(perm);
-    if ( wxMkDir(dir) != 0 )
+
+    // Specifies the mode_t parameter of wxMkDir
+    // wxS_DIR_DEFAULT is specified if exists,
+    // otherwise defaults to 0777 (Full permission)
+    // Fixes the following issue:
+    // /src/common/filefn.cpp:663:17: error: too few arguments to function ‘int wxMkDir(const wxString&, mode_t)’
+    #ifdef wxS_DIR_DEFAULT
+    if ( wxMkDir(dir, wxS_DIR_DEFAULT) != 0 )
+    #else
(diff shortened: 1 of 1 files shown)


swt2c:
On which platform and compiler are you seeing that compile error?  `wxMkDir()` is declared in the header with a default argument for `mode_t`, so it shouldn't be needed:
https://github.com/wxWidgets/wxWidgets/blob/master/include/wx/filefn.h#L448

vadz:
This is a Cygwin-only issue, apparently: for some reason `mode` doesn't have a default value there. The right fix is almost certainly to add it to the declaration instead of this hack.

@Proshanto-1234 Can you check that this diff fixes the problem too?

```diff
diff --git a/include/wx/filefn.h b/include/wx/filefn.h
index 0abd2abe0c..e79c8f67c5 100644
--- a/include/wx/filefn.h
+++ b/include/wx/filefn.h
@@ -381,7 +381,7 @@ inline int wxRmDir(const wxString& path)
 inline int wxMkDir(const wxString& path, mode_t WXUNUSED(mode) = 0)
     { return wxCRT_MkDir(path.fn_str()); }
 #else
-inline int wxMkDir(const wxString& path, mode_t mode)
+inline int wxMkDir(const wxString& path, mode_t mode = 0777)
     { return wxCRT_MkDir(path.fn_str(), mode); }
 #endif
 
```

Proshanto-1234:
@swt2c  the issue is in Cygwin on Windows. The generator is Unix Generator, and the build system is CMake, before running make -j$(nproc)

Proshanto-1234:
@vadz inline int wxMkDir(const wxString& path, mode_t mode = 0777) works! Here is it:
[ 36%] Building CXX object libs/base/CMakeFiles/wxbase.dir/__/__/__/__/src/common/filefn.cpp.obj

PR #26048, opened 2025-12-22
Improvements to `wxWebRequest` handling of redirects

vadz:
The net effect of these changes is that `POST` works correctly when using authentication with libcurl, however there are more changes than strictly necessary to just fix this:

1. We now resend request data after redirect when using libcurl.
2. We avoid extra requests in the first place by allowing to preemptively use basic authentication.

So far there is no support for preemptive use of proxy authentication or using any authentication method other than "Basic" just because I don't need this, but this could be easily added later too.

### DIFF
--- a/include/wx/private/webrequest.h
+++ b/include/wx/private/webrequest.h
@@ -90,6 +90,11 @@ public:
 
     virtual void SetTimeouts(long connectionTimeoutMs, long dataTimeoutMs) = 0;
 
+    void UseBasicAuth(const wxWebCredentials& cred)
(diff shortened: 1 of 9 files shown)


vadz:
OK, so I've changed my mind myself after seeing the tests breakage. I'll add `UseBasicAuth()` instead to explicitly opt in into using it.

vadz:
I think this should be relatively uncontroversial now, but please let me know if anybody has any objections. Otherwise I'll merge this soon.

Issue #26044, opened 2025-12-21
Regression: GetBestTranslation() returns suboptimal and/or inconsistent results


arch1t3cht:
#### Bug description:

After 57e1c9e1ee156ff0332f51b08aef65d52adaa4e7, `wxTranslations::GetBestTranslation()` can return a language that is not the user's display language. What language is returned can change depending on when the function is called.

Specifically, consider the following situation:
- On Windows, the user's display language is set to `English (United Kingdom)`, but the user also has a (say) Japanese language pack installed, e.g. for a Japanese keyboard.
- A Japanese translation is available for the catalog in question
- `(new wxTranslations())->GetBestTranslation()` is called very early on. I'm not quite sure *how* early it needs to be, but for example it needs to come before any call to `wxMessageBox()`.

In such a case, `GetBestTranslation()` will return `ja` even though this is not the user's display language. (I'm guessing this is due to the fact that `en_GB` does not match `en` exactly, but `ja` does match `ja` exactly?)

If `wxMessageBox()` is called before `(new wxTranslations())->GetBestTranslation()`, the call returns `en` instead, so that the results are inconsistent.

#### Patch or snippet allowing to reproduce the problem:

Apply the following diff to current master:
```diff
diff --git a/samples/internat/internat.cpp b/samples/internat/internat.cpp
index 5f6072e7f8..bd36dede4f 100644
--- a/samples/internat/internat.cpp
+++ b/samples/internat/internat.cpp
@@ -203,6 +203,15 @@ bool MyApp::OnInit()
     if ( !wxApp::OnInit() )
         return false;

+    wxFileTranslationsLoader::AddCatalogLookupPathPrefix(".");
+
+    // wxMessageBox("Hello");
+
+    wxTranslations* const trans = new wxTranslations();
+    wxMessageBox(wxString::Format("Best translation is %s. %d available translations.", trans->GetBestTranslation("internat"), (int)trans->GetAvailableTranslations("internat").size()));
+
+    return false;
+
     // For demonstration purposes only, ask the user if they want to run the
     // program using the current system language. In real programs, we would do
     // it unconditionally for localized programs -- or never do it at all for
```

#### To Reproduce:

<img width="761" height="330" alt="Image" src="https://github.com/user-attachments/assets/f17ebeda-cd72-45d6-9019-7cb82b221004" />

1. Configure the Windows system language as in the above screenshot.
2. Apply the above patch and compile and run `samples/internat`.
[...]

PR #26050, opened 2025-12-23
Fix issue #26044 - GetBestTranslation

utelle:
Method GetBestTranslation() uses internally static methods of wxUILocaleImpl which in turn require an initialized language database. Calls to wxUILocaleImpl::CreateLanguagesDB() are added to guarantee this.

### DIFF
--- a/src/common/languageinfo.cpp
+++ b/src/common/languageinfo.cpp
@@ -42,6 +42,7 @@ static std::unordered_map<wxString, wxString> gs_scmap_alias2name;
 
 /* static */ wxString wxUILocaleImpl::GetScriptAliasFromName(const wxString& scriptName)
 {
+    wxUILocaleImpl::CreateLanguagesDB();
     const auto scIter = gs_scmap_name2alias.find(scriptName);
     if (scIter != gs_scmap_name2alias.end())
         return scIter->second;
@@ -51,6 +52,7 @@ static std::unordered_map<wxString, wxString> gs_scmap_alias2name;
 
 /* static */ wxString wxUILocaleImpl::GetScriptNameFromAlias(const wxString& scriptAlias)
 {
+    wxUILocaleImpl::CreateLanguagesDB();
     const auto scIter = gs_scmap_alias2name.find(scriptAlias);
     if (scIter != gs_scmap_alias2name.end())
         return scIter->second;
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, I'll merge this if/when it's confirmed that this fixes #26044, even though I'm not particularly happy with the fact that we need to remember to call this function instead of relying on some way for it to be called automatically — but this will be something to improve in 2026.

Thanks again!

utelle:
> Thanks, I'll merge this if/when it's confirmed that this fixes #26044,

It _**fixes**_ #26044. As described there the _correct_ result (`en`) was returned, when `wxMessageBox` was called. And as explained a side effect of calling `wxMessageBox` is that the language database is initialized.

> even though I'm not particularly happy with the fact that we need to remember to call this function instead of relying on some way for it to be called automatically — but this will be something to improve in 2026.

Yes, being forced to call `CreateLanguagesDB()` is far from ideal. The change introduced by this PR is a very local modification (affecting only a single source file). Nevertheless, initializing the language database automatically is most likely preferrable.

In file `common/intl.cpp` there is already a `wxLocaleModule` class that is used to automatically _destroy_ the language database. Why not use that class to automatically _create_ the language database?

vadz:
> In file `common/intl.cpp` there is already a `wxLocaleModule` class that is used to automatically _destroy_ the language database. Why not use that class to automatically _create_ the language database?

The only reason I can see is to postpone initializing it until it's really needed. In principle, it's possible that it is never used, I think?

utelle:
> > In file `common/intl.cpp` there is already a `wxLocaleModule` class that is used to automatically _destroy_ the language database. Why not use that class to automatically _create_ the language database?
> 
> The only reason I can see is to postpone initializing it until it's really needed. In principle, it's possible that it is never used, I think?

In principle, you are right. However, if you take into account that even a simple function call like `wxMessageBox()` already triggers the initialization of the language database (as a side effect of calling `GetLayoutDirection()` internally), if symbol `wxUSE_INTL` is defined, it is justifiable to initialize the language database unconditionally.

utelle:
> However, if you take into account that even a simple function call like `wxMessageBox()` already triggers the initialization of the language database (as a side effect of calling `GetLayoutDirection()` internally), if symbol `wxUSE_INTL` is defined, it is justifiable to initialize the language database unconditionally.

OTOH the static methods in `languageinfo.cpp` now take care of initializing the language database. And the static global data structures are not referenced from anywhere else. So, applying this PR efficiently solves the issue.

vadz:
Thanks again, I'll merge it soon.

PR #26052, opened 2025-12-23
Prevent crash in wxD2DRenderer::CreateContext* if BindDC() call fails.

dhansel:
Instead return NULL and let user code handle the error situation.
This fixes #[26041](https://github.com/wxWidgets/wxWidgets/issues/26041)

### DIFF
--- a/src/msw/graphicsd2d.cpp
+++ b/src/msw/graphicsd2d.cpp
@@ -757,7 +757,12 @@ public:
 
         DoAcquireResource();
 
-        wxCHECK_RESOURCE_HOLDER_POST();
+        // do NOT use wxCHECK_RESOURCE_HOLDER_POST
+        // here since DoAcquireResource may have failed
+        // for "legitimate" reasons (e.g. region too large)
+        // that should be handled in user code once we return.
+        // It should not cause an assertion fail here.
+        //wxCHECK_RESOURCE_HOLDER_POST();
     }
 
     void ReleaseResource() override
@@ -3759,11 +3764,17 @@ protected:
         wxCHECK_RET( status != ERROR, wxS("Error retrieving DC dimensions") );
 
         hr = renderTarget->BindDC(m_hdc, &r);
-        wxCHECK_HRESULT_RET(hr);
-        renderTarget->SetTransform(
-                       D2D1::Matrix3x2F::Translation(-r.left, -r.top));
 
-        m_nativeResource = renderTarget;
+        // BindDC may fail with E_INVALIDARG if the given RECT is too large.
+        // It is better to let the user code handle this condition than
+        // to produce an assertion error here.
+        if (SUCCEEDED(hr))
+        {
+            renderTarget->SetTransform(
+                D2D1::Matrix3x2F::Translation(-r.left, -r.top));
+
+            m_nativeResource = renderTarget;
+        }
     }
 
 private:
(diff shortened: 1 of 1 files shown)


src/msw/graphicsd2d.cpp
```diff
@@ -3759,11 +3764,17 @@ class wxD2DDCRenderTargetResourceHolder : public wxD2DRenderTargetResourceHolder
         wxCHECK_RET( status != ERROR, wxS("Error retrieving DC dimensions") );
 
         hr = renderTarget->BindDC(m_hdc, &r);
-        wxCHECK_HRESULT_RET(hr);
-        renderTarget->SetTransform(
-                       D2D1::Matrix3x2F::Translation(-r.left, -r.top));
 
-        m_nativeResource = renderTarget;
+        // BindDC may fail with E_INVALIDARG if the given RECT is too large.
+        // It is better to let the user code handle this condition than
```
vadz, line 3769: The trouble is that currently there is no way for the application code to even know that an error occurred... Could we propagate the error from here somehow?

src/msw/graphicsd2d.cpp
```diff
@@ -757,7 +757,12 @@ class wxD2DResourceHolder: public wxManagedResourceHolder
 
         DoAcquireResource();
 
-        wxCHECK_RESOURCE_HOLDER_POST();
+        // do NOT use wxCHECK_RESOURCE_HOLDER_POST
```
vadz, line 760: This macro (and then this comment) should be just removed entirely as it's not used anywhere else.

src/msw/graphicsd2d.cpp
```diff
@@ -4174,19 +4185,24 @@ void wxD2DContext::Init()
...
+    ID2D1RenderTarget* target = GetRenderTarget();
+
     // Remove all layers from the stack of layers.
     while ( !m_layers.empty() )
     {
         LayerData ld = m_layers.top();
         m_layers.pop();
 
-        GetRenderTarget()->PopLayer();
+        if (target != nullptr) target->PopLayer();
```
vadz, line 4196: Minor, but in wx coding style this should be written as

```suggestion
        if ( target != nullptr )
            target->PopLayer();
```

(here and below, of course)

src/msw/graphicsd2d.cpp
```diff
@@ -5169,7 +5189,13 @@ wxD2DRenderer::~wxD2DRenderer()
 
 wxGraphicsContext* wxD2DRenderer::CreateContext(const wxWindowDC& dc)
 {
-    return new wxD2DContext(this, m_direct2dFactory, dc);
+    wxD2DContext* d2d = new wxD2DContext(this, m_direct2dFactory, dc);
```
vadz, line 5192: It would be nice to make a factory function to be able to write just

```cpp
return wxD2DContext::New(this, m_direct2dFactory, dc);
```

here and below instead of repeating the same code in multiple places.

vadz (commented):
Thanks, this looks like a good change, but it would be nice to polish it a bit, please do it if you can, otherwise I'll try to do later (next year).

src/msw/graphicsd2d.cpp
dhansel: My idea was that the wxD2DRenderer::CreateContext() function (which is called from application code) will now return NULL instead of crashing - so the application can see there was a problem. Am I misunderstanding/missing something?

src/msw/graphicsd2d.cpp
dhansel: Sorry, not quite understanding this. The "dc" in the different wxD2DContext() constructor calls is a different type every time. So wouldn't that mean I would need a wxD2DContext::New() function for each? Maybe you can point me to a similar construct in existing code so I can take a closer look?

vadz: Sorry, I don't have an example at hand but we could make a template `New()`, forwarding the arguments to the appropriate ctor (which would become private).

Alternatively, and maybe even simpler, just add some helper `CheckIfOk(wxD2DContext*)` function that would do this and wrap all `new wxD2DContext(...)` calls in it.

A template `New()` is safer as it would make it impossible to use a context which couldn't be created.

vadz:
I'd just like to check if you plan any more changes here or if I should merge it (maybe with some changes, as discussed above)? Thanks!

dhansel:
I had hoped I could get to it but work has come in the way. If you want to make some changes and merge it that would be very helpful. I'm not sure when I will get to it.

vadz:
I've created #26169 replacing this one, please let me know if you see any problems with it.

dhansel:
That looks good to me - I see now what you meant by using a factory to share code between the different  CreateContextFrom* calls. Have to admit I did not know about variadic templates. There's always more to learn...

Thanks for fixing this!

PR #26055, opened 2025-12-27
Windows Accessibility: Report checked checked state of items in wxCheckListBox

### LINKED ISSUE #25948: Accessibility: wxCheckListBox doesn't convey checked/unchecked state of items on Windows
The standard Windows ListBox class has no concept of checked/unchecked state of items. WXWidgets obviously overcomes this with owner drawn items so that the checked boxes can be drawn. But this info gets lost when using the standard Windows IAccessible ListBox implementation since said implementation has no concept of  checked state for individual items.

I have a fix in-process  for which I will  offer up a pull request in the next couple of weeks.

glenrgordon:
The standard Windows ListBox class has no concept of checked/unchecked
state of items. WXWidgets obviously overcomes this with owner drawn
items so that the checked boxes can be drawn. But this info gets lost
when using the standard Windows IAccessible ListBox implementation
since said implementation has no concept of checked state for
individual items.

Fixes https://github.com/wxwidgets/wxwidgets/issues/25948

### DIFF
--- a/include/wx/msw/checklst.h
+++ b/include/wx/msw/checklst.h
@@ -62,6 +62,9 @@ public:
     // so we need to override these functions
     virtual wxOwnerDrawn *CreateLboxItem(size_t n) override;
     virtual bool MSWOnMeasure(WXMEASUREITEMSTRUCT *item) override;
+#if wxUSE_ACCESSIBILITY
+    virtual wxAccessible* CreateAccessible() override;
+#endif
 
 protected:
     virtual wxSize MSWGetFullItemSize(int w, int h) const override;
@@ -87,5 +90,16 @@ protected:
     wxDECLARE_EVENT_TABLE();
     wxDECLARE_DYNAMIC_CLASS_NO_COPY(wxCheckListBox);
 };
+
+#if wxUSE_ACCESSIBILITY
+class WXDLLIMPEXP_CORE wxCheckListBoxAccessible : public wxWindowAccessible
+{
+public:
+    explicit wxCheckListBoxAccessible(wxCheckListBox* winm);
+
+    virtual wxAccStatus GetRole(int childId, wxAccRole* role) override;
+    virtual wxAccStatus GetState(int childId, long* state) override;
+};
+#endif
 
 #endif    //_CHECKLST_H
--- a/include/wx/msw/checklst.h
+++ b/include/wx/msw/checklst.h
@@ -91,15 +91,4 @@ protected:
     wxDECLARE_DYNAMIC_CLASS_NO_COPY(wxCheckListBox);
 };
 
-#if wxUSE_ACCESSIBILITY
-class WXDLLIMPEXP_CORE wxCheckListBoxAccessible : public wxWindowAccessible
-{
-public:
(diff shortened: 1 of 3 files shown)


include/wx/msw/checklst.h
```diff
@@ -88,4 +91,15 @@ class WXDLLIMPEXP_CORE wxCheckListBox : public wxCheckListBoxBase
     wxDECLARE_DYNAMIC_CLASS_NO_COPY(wxCheckListBox);
 };
 
+#if wxUSE_ACCESSIBILITY
+class WXDLLIMPEXP_CORE wxCheckListBoxAccessible : public wxWindowAccessible
```
vadz: Does this class need to be public? I'd hide it in `src/msw/checklst.cpp` and not export it (i.e. remove `WXDLLIMPEXP_CORE`) as I don't think the application code should ever use it directly.

include/wx/msw/checklst.h
```diff
@@ -88,4 +91,15 @@ class WXDLLIMPEXP_CORE wxCheckListBox : public wxCheckListBoxBase
...
+{
+public:
+    explicit wxCheckListBoxAccessible(wxCheckListBox* winm);
```
vadz: This is very minor, but why is this parameter called `winm`? Just `win` is fine, of course, but why `m`?

src/common/wincmn.cpp
```diff
@@ -3955,7 +3955,7 @@ wxAccStatus wxWindowAccessible::GetParent(wxAccessible** parent)
         if (*parent)
             return wxACC_OK;
         else
-            return wxACC_FAIL;
+            return wxACC_NOT_IMPLEMENTED;
```
vadz: Is this related to the other changes? If not, it should be ideally extracted into a separate commit, with the explanation of why is this necessary in its commit message.

src/msw/checklst.cpp
```diff
@@ -276,6 +276,10 @@ void wxCheckListBox::Check(unsigned int uiIndex, bool bCheck)
 
     GetItem(uiIndex)->Check(bCheck);
     RefreshItem(uiIndex);
+#if wxUSE_ACCESSIBILITY
+    wxAccessible::NotifyEvent(wxACC_EVENT_OBJECT_STATECHANGE, this, wxOBJID_CLIENT, uiIndex+1);
```
vadz: Extremely minor, but for consistency with the similar line just below:

```suggestion
    wxAccessible::NotifyEvent(wxACC_EVENT_OBJECT_STATECHANGE, this, wxOBJID_CLIENT, uiIndex + 1);
```

vadz (commented):
Great, thanks for implementing this!

I didn't test it, but the code looks correct to me, it would just be nice to address the minor comments/questions below before merging it.

Thanks again!

glenrgordon:
Thanks for the reivew!
I think I've addressed all of the issues. Let me know if I've missed
anything.
Will add a new PR for the portion that got removed because it's  only
indirectly related to this change.

vadz:
Looks good now, thanks! I'll merge this soon with just some very minor cosmetic changes.

PR #26058, opened 2025-12-29
Fix wxChoice::GetBestSize() on macOS 26

vslavik:
I either didn’t notice initially or this was adjusted in macOS 26.2, but `wxChoice` best size was not wide enough:
 
CleanShot 2025-12-22 at 22 39 29@2x

This updates the code to match native metrics exactly.


vadz:
Thanks, will push soon (but probably next year...).

PR #26060, opened 2025-12-31
Update generic wxTreeCtrl state images when DPI changes

### LINKED ISSUE #26059: State images for wxTreeCtrl not updating for DPI changes when no images are set
Hi Vadim, others.

Using wxWidgets f3aa944b01 from git on Mac Tahoe 26.2

- I'm using wxTreeCtrl with only State Images by `wxTreeCtrl::SetStateImages()` with a `wxVector<wxBitmapBundle>` with 16x16 (in red) and 32x32 (in green for test) wxBitmaps
- I set states with `wxTreeCtrl::SetItemState()` for each Child Item (for test)
- I start the application on a HDPI display and it will use the 32x32 Bitmaps (1st image with 32x32 green state images)
Image

- I then drag the application to a non-HDPI display and it will still use the 32x32 Bitmaps, which are now too big (2nd image with large 32x32 green state images)
Image

- In debugger with wxWidgets source I think I traced it down to the `wxWithImages` function:

`void WXHandleDPIChanged(wxDPIChangedEvent& event)`
`    {`
`        if ( HasImages() )`
`            OnImagesChanged();`
`        event.Skip();`
`    }`

- Because I only set State Images, `HasImages()` returns false and hence virtual `wxGenericTreeCtrl::OnImagesChanged()` is never called, which should handle DPI changes for the state images by `m_imagesState.GetUpdatedImageListFor(this)`
- If during initialisation apart from `wxTreeCtrl::SetStateImages(BitmapBundleList)` I also call `wxTreeCtrl::SetImages(BitmapBundleList)` (even though I don't use any Item Images), then the State Images will become DPI-aware correctly as expected (3rd images with 16x16 red state images)
`wxGenericTreeCtrl::OnImagesChanged()` will be used to handle DPI changes for State Images
Image

- So there is a workaround, but it might be worthwhile fixing ? (although it might be difficult in `wxWithImages` to judge that `wxTreeCtrl` will use OnImagesChanged() to update it State Images... ?)

vadz:
Ensure that wxTreeCtrl using only state images, without the normal ones, is still updated correctly when the DPI changes: this didn't happen previously because OnImagesChanged() was only called for the control with normal images.

As changing HasImages() to return true if there are only state images is almost certainly a bad idea for compatibility reasons (existing code may assume that it can use normal images if it returns true), add a new HasAnyImages() function and override it in wxTreeCtrl to account for the state images too.

Closes #26059.

### DIFF
--- a/include/wx/treectrl.h
+++ b/include/wx/treectrl.h
@@ -435,6 +435,10 @@ protected:
                                         int& flags) const = 0;
 
 
+    // override the base class version to take state images into account
+    virtual bool HasAnyImages() const override;
+
+
     // Usually we inherit from this class, rather than aggregating it, but we
     // need two different sets of images here, so we do both.
     wxWithImages m_imagesState;
--- a/include/wx/withimages.h
+++ b/include/wx/withimages.h
@@ -262,6 +262,15 @@ public:
     }
 
 protected:
+    // This is the same as HasImages() for all controls except for wxTreeCtrl
+    // which has a special "state image list" in addition to the normal one and
+    // this function takes it into account, unlike HasImages() itself, which
+    // doesn't and can't be modified to do so for compatibility reasons.
+    virtual bool HasAnyImages() const
+    {
+        return HasImages();
+    }
+
     // This function is called when the images associated with the control
     // change, due to either SetImages() or SetImageList() being called.
     //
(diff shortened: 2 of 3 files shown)


vadz:

Issue #26043, opened 2025-12-21
Linking with `wx::base` still defines `-D__WX${TOOLKIT}__`


vadz:
I've just realized that having a console application using `target_link_libraries(myapp PRIVATE wx::base)` results in this application being compiled with the toolkit symbol, e.g. `__WXGTK__`, defined.

This is a problem because it means using GUI functions is possible at compile-time and any errors are only flagged at link-time and also because when using `-fsanitize=undefined`, UBSAN adds dependencies on the type info objects for GUI classes that are not satisfied when linking, i.e. this results in link errors unless you additionally use `-fno-sanitize=vptr` (which disables a useful check).

I'm not sure how to fix this, my first attempt was:

```diff
diff --git a/build/cmake/functions.cmake b/build/cmake/functions.cmake
index f8d77b0ff8..ae556d4476 100644
--- a/build/cmake/functions.cmake
+++ b/build/cmake/functions.cmake
@@ -388,8 +388,13 @@ function(wx_set_target_properties target_name)
         target_link_libraries(${target_name}
             PUBLIC ${wxTOOLKIT_LIBRARIES})
     endif()
+    if(wxTARGET_IS_BASE)
+        target_compile_definitions(${target_name}
+            PUBLIC __WXBASE__ wxUSE_GUI=0)
+    else()
         target_compile_definitions(${target_name}
             PUBLIC ${wxTOOLKIT_DEFINITIONS})
+    endif()
 
     if(wxBUILD_SHARED)
         string(TOUPPER ${target_name_short} target_name_upper)

```

but this doesn't work because linking with `wx::base` and `wx::core` now results in `wxUSE_GUI` redefinition: I hoped that having `-DwxUSE_GUI=0 -DwxUSE_GUI=1` on the command line would work as expected, with the latter overriding the former, but it doesn't and results in a warning (which becomes an error with `-Werror`).

Does anybody see a way to somehow define `wxUSE_GUI=0` only if _only_ `wx::base` is being used but not any other target? I don't think this is possible, as all dependencies are independent at CMake level, but I could be missing something.

If not, I guess we need some new symbol `wxUSE_BASE_BUT_NOT_IF_OVERRIDDEN_BY_USE_GUI_LATER` (maybe with a slightly shorter name) that would make `wx/defs.h` define `wxUSE_GUI=0` only if it's not already defined. Please let me know if anybody sees a better solution — or why this wouldn't work.

MaartenBent:
[...]

PR #26061, opened 2025-12-31
CMake: Don't add toolkit defines to base libraries

MaartenBent:
And group all toolkit variables together.

See #26043

### DIFF
--- a/build/cmake/functions.cmake
+++ b/build/cmake/functions.cmake
@@ -398,11 +398,6 @@ function(wx_set_target_properties target_name)
             $<INSTALL_INTERFACE:${wxINSTALL_INCLUDE_DIR}>
         )
 
-    if(wxTOOLKIT_INCLUDE_DIRS AND NOT wxTARGET_IS_BASE)
-        target_include_directories(${target_name}
-            PRIVATE ${wxTOOLKIT_INCLUDE_DIRS})
-    endif()
-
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, but I don't think (I haven't tested yet, but I just don't see how it could) this fixes the wrong `wxUSE_GUI` value for console apps. Can I push commits attempting to fix this here or do you plan to add more commits yourself?

MaartenBent:
I forgot to handle the case when the library is built with `wxUSE_GUI=OFF` and the base libraries should get the `__WXBASE__` toolkit define.
I'll push one more commit to fix this. After that you can add commits.

MaartenBent:
I don't think this is the correct approach, at least not when building the libraries. I compared against the vcxproj and configure on macOS and they both have defines for the toolkit (`__WXMSW__`/`__WXOSX_COCOA__`) when building base libraries.

It is possible to use different defines for building (`target_compile_definitions(... PRIVATE ...)`) and when linking with it (`target_compile_definitions(... INTERFACE ...)`). But `wx-config --cppflags base` includes the toolkit as well, so I don't think we should do something different in CMake.

`wx-config --cppflags base` does add the `-DwxUSE_GUI=0` flag. I guess this is what you suggested initially in #26043. I'll investigate that further.

PR #26066, opened 2026-01-02
wxRichTextCtrl: Enhance macOS key bindings

torstenvl:
Updated key bindings for macOS to more closely match native behavior.

Currently, Cmd+Left goes to the left word boundary, and Cmd+Right goes
to the right word boundary, like on Windows (wxWidgets treats Cmd on
Mac as if it were Ctrl for keyboard modifier key purposes). Alt/Opt+
Left and Alt/Opt+Right do nothing. 

This change maps Cmd+Left to the start of the line, and Cmd+Right to
the end of the line.  It also maps Alt/Opt+Left to WordLeft() and Alt/
Opt+Right to WordRight(), which is conventional on native Cocoa
controls (including other wxWidgets controls that map directly to Cocoa
controls).

### DIFF
--- a/src/richtext/richtextctrl.cpp
+++ b/src/richtext/richtextctrl.cpp
@@ -1655,8 +1655,10 @@ Left:       left one character
 Right:      right one character
 Up:         up one line
 Down:       down one line
-Ctrl-Left:  left one word
-Ctrl-Right: right one word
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, will push soon (and cherry pick to 3.2 too as it seems to be safe to do).

PR #26071, opened 2026-01-04
fix wxTipWindow construction sequencing

wsu-cb:


interface/wx/tipwin.h
```diff
@@ -25,7 +25,77 @@ class wxTipWindow : public wxWindow
...
+        wxTipWindow may close itself, so provide a smart pointer
+        that acts as a weak reference to wxTipWindow.
+
+        This is a move-only type.
+
+        Note that this is not a wxWeakRef<> because this is set
+        to @NULL when wxTipWindow is closed, which may be
+        "long" before wxTipWindow is destroyed, and wxWeakRef<>
+        is set to @NULL on object destruction.
+    */
```
vadz, line 43: Please add

```suggestion
        
        @since 3.3.2
    */
```

here and for the new functions below.

samples/dialogs/dialogs.cpp
```diff
@@ -2858,21 +2854,20 @@ void MyFrame::OnShowTip(wxCommandEvent& WXUNUSED(event))
...
-                            &m_tipWindow
+                            FromDIP(60)
                           );
     }
 }
 
 void MyFrame::OnUpdateShowTipUI(wxUpdateUIEvent& event)
 {
-    event.Check(m_tipWindow != nullptr);
+    event.Check(bool(m_tipWindow));
```
vadz: I think this shows that we need `IsOk()` member function in `Ref`.

Or, maybe, implement `operator!=()` for it?

src/html/helpctrl.cpp
```diff
@@ -314,11 +314,11 @@ bool wxHtmlHelpController::DisplayTextPopup(const wxString& text, const wxPoint&
         s_tipWindow->SetTipWindowPtr(nullptr);
         s_tipWindow->Close();
     }
-    s_tipWindow = nullptr;
+    s_tipWindow.Reset();
```
vadz: In theory we could also keep this working and avoid this change by providing overloaded `operator=(std::nullptr_t)`. Not sure if it's really much better than `Reset()`.

include/wx/tipwin.h
```diff
@@ -26,15 +26,56 @@ class WXDLLIMPEXP_FWD_CORE wxTipWindowView;
...
+    public:
+        Ref() { *p = nullptr; }
+
+        void Reset() { *p = nullptr; }
+
+        explicit operator bool() const { return *p; }
+        wxTipWindow* operator->() const { wxASSERT(*p);  return *p; }
+
+    private:
+        std::unique_ptr<wxTipWindow*> p { new wxTipWindow* };
```
vadz: Sorry, I may be missing something obvious but why do we need to use `std::unique_ptr` here when it looks like we could just have `wxTipWindow* m_ptr;` and replace `*p` above with `m_ptr`?

include/wx/tipwin.h
```diff
@@ -26,15 +26,56 @@ class WXDLLIMPEXP_FWD_CORE wxTipWindowView;
...
+    // is set to nullptr on object destruction
+    class Ref
+    {
+    public:
+        Ref() { *p = nullptr; }
+
+        void Reset() { *p = nullptr; }
+
+        explicit operator bool() const { return *p; }
+        wxTipWindow* operator->() const { wxASSERT(*p);  return *p; }
```
vadz: This `wxASSERT()` is pretty useless, it's going to crash in any case and it doesn't add anything helpful in debug builds (asserting or crashing at this line is about the same).

include/wx/tipwin.h
```diff
@@ -26,15 +26,56 @@ class WXDLLIMPEXP_FWD_CORE wxTipWindowView;
 class WXDLLIMPEXP_CORE wxTipWindow : public wxPopupTransientWindow
 {
 public:
-    // the mandatory ctor parameters are: the parent window and the text to
+    // wxTipWindow may close itself, so provide a smart pointer
+    // that acts as a weak reference to wxTipWindow.
+    //
+    // Note that this is a move-only type.
```
vadz: Can it actually be moved? I think you need to provide move ctor/assignment operator (but they can be `= default`).

vadz (commented):
Thanks, this looks nice, but I think that adding `IsOk()` or `operator!=(nullptr_t)` is needed to make using this even more natural.

Also, if the use of `unique_ptr` for `Ref::p` is really required, please add a comment explaining why as I just don't see it (and please rename `p` to something with `m_` prefix in any case).

TIA!

samples/dialogs/dialogs.cpp
```diff
@@ -2858,21 +2854,20 @@ void MyFrame::OnShowTip(wxCommandEvent& WXUNUSED(event))
     }
     else
     {
-        m_tipWindow = new wxTipWindow
+        m_tipWindow = wxTipWindow::New
```
wsu-cb, line 2857: This line proves that wxTipWindow::Ref is moveable.

include/wx/tipwin.h
wsu-cb: See https://github.com/wxWidgets/wxWidgets/pull/26071/files#r2663388851.  (The std::unique_ptr<> provides move ctor/assignment, and any class ctor or assignment automatically uses its members ctor or assignment unless they are blocked somehow.  See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rc-zero).

interface/wx/tipwin.h
wsu-cb: done

samples/dialogs/dialogs.cpp
wsu-cb: implemented operator!=(nullptr_t)

src/html/helpctrl.cpp
wsu-cb: implemented operator=(nullptr_t)

include/wx/tipwin.h
wsu-cb: `wxASSERT()` removed

include/wx/tipwin.h
wsu-cb: done

include/wx/tipwin.h
vadz: Yes, you're right, sorry for overlooking this.

I know that this class is hardly performance-critical, but it still makes me cringe a little that we use an extra heap allocation just to move this movable. It looks like using a raw pointer and deleting special copy member functions and implementing move ones shouldn't be that difficult and would avoid it...

vadz:

PR #26072, opened 2026-01-05
Fix disabled wxListCtrl background colour

### LINKED ISSUE #26068: Disabled wxListCtrl cell backgrounds are white if a font/color/background is set on any cell.
If I call 

```cpp
m_listCtrl->Enable(false);
``` 

in the `listtest` sample, the backgrounds of the cells are still white. I expect them to have the darker color:

Image

If I remove all `SetBackgroundColour`, `SetTextColour`, `SetFont` calls, the cell backgrounds looks correct then:

Image

- wxWidgets version  you use: master (3.3)

MaartenBent:
Fixes #26068

### DIFF
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -3305,6 +3305,16 @@ void DrawGridLines(wxListCtrl* listctrl, int item, int gap = 0)
     }
 }
 
+wxColour GetListCtrlBgCol(wxListCtrl* listctrl)
+{
+    if ( listctrl->IsEnabled() )
+        return listctrl->GetBackgroundColour();
+    else if ( wxMSWDarkMode::IsActive() )
+        return wxMSWDarkMode::GetColour(wxSYS_COLOUR_BTNFACE);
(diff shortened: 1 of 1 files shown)


src/msw/listctrl.cpp
```diff
@@ -3305,6 +3305,16 @@ void DrawGridLines(wxListCtrl* listctrl, int item, int gap = 0)
     }
 }
 
+wxColour GetListCtrlBgCol(wxListCtrl* listctrl)
```
vadz, line 3308: Maybe rename this to `GetEffectiveBackgroundColour()` or something like this to make it more clear that this is different from member `GetBackgroundColour()`?

(I could change this myself when applying if you agree)

vadz (approved):
Thanks for fixing this, will merge soon (with a possible rename).

src/msw/listctrl.cpp
MaartenBent: Fine with me 👍

PR #26078, opened 2026-01-08
[msw/font] If the face name might be truncated, try to get the full face name

### LINKED ISSUE #25333: wxFont::GetFaceName result is truncated
### Description

On Windows, wxFontEnumerator::GetFacenames and wxFont::GetFaceName use GDI, which truncates the face name to 31 characters. It would be nice if we could get the non-truncated face name.

We can retrieve the non-truncated name via [GetOutlineTextMetricsW](https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-getoutlinetextmetricsw). I'm not sure if we should use otmpFamilyName or otmpFaceName.

For a example to how to implement it, see this PR: https://github.com/arch1t3cht/Aegisub/pull/155

- wxWidgets version: 3.1.4
- OS: Windows 10.0.26100.3775

moi15moi:
Fix: https://github.com/wxWidgets/wxWidgets/issues/25333
LOGFONT.lfFaceName has a defined size (32). Because of that, the face name can be truncated (see the issue for a font with a family name longer then 31 characters). To avoid this problem, we can just call GetOutlineTextMetrics and use it's otmpFamilyName which is not truncated.

To test this, I used https://github.com/wxWidgets/wxWidgets/tree/1b582af60d99388b56b383fe240b647387a5e4b4/samples/font

In the issue, you mentionned I should just edit ``wxFontEnumerator``, but actually, I also needed to edit ``wxFontRefData`` because with the [sample](https://github.com/wxWidgets/wxWidgets/tree/1b582af60d99388b56b383fe240b647387a5e4b4/samples/font), if I use ``Select -> Select font``, the textbox will contain the truncated face name if I don't add a check with ``wxFontRefData``.

### DIFF
--- a/src/msw/font.cpp
+++ b/src/msw/font.cpp
@@ -46,6 +46,43 @@
 static const int PITCH_MASK = FIXED_PITCH | VARIABLE_PITCH;
 
 // ----------------------------------------------------------------------------
+// global functions implementation
+// ----------------------------------------------------------------------------
+
+wxString wxGetMSWFaceNameFromHFONT(HFONT hFont)
+{
+    ScreenHDC hdc;
+    SelectInHDC selectFont(hdc, hFont);
+
+    UINT otmSize = GetOutlineTextMetrics(hdc, 0, nullptr);
+    if ( !otmSize )
+    {
+        wxLogLastError("GetOutlineTextMetrics(nullptr)");
+        return wxString();
+    }
+
+    OUTLINETEXTMETRIC * const
+        otm = static_cast<OUTLINETEXTMETRIC *>(malloc(otmSize));
+    wxON_BLOCK_EXIT1( free, otm );
+
+    otm->otmSize = otmSize;
+    if ( !GetOutlineTextMetrics(hdc, otmSize, otm) )
+    {
+        wxLogLastError("GetOutlineTextMetrics()");
+        return wxString();
+    }
+
+    // in spite of its type, the otmpFamilyName field of OUTLINETEXTMETRIC
+    // gives an offset in _bytes_ of the face (not family!) name from the
+    // struct start while the name itself is an array of TCHARs
+    //
+    // FWIW otmpFaceName contains the same thing as otmpFamilyName followed
+    // by a possible " Italic" or " Bold" or something else suffix
+    return reinterpret_cast<wxChar *>(otm) +
+                wxPtrToUInt(otm->otmpFamilyName)/sizeof(wxChar);
+}
+
+// ----------------------------------------------------------------------------
 // wxFontRefData - the internal description of the font
 // ----------------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


include/wx/fontutil.h
```diff
@@ -290,4 +290,9 @@ WXDLLIMPEXP_CORE bool wxGetNativeFontEncoding(wxFontEncoding encoding,
 // return true if such font(s) exist, false otherwise
 WXDLLIMPEXP_CORE bool wxTestFontEncoding(const wxNativeEncodingInfo& info);
 
+#if defined(__WXMSW__)
+// Helper function to get the non-truncated face name from an HFONT
+WXDLLIMPEXP_CORE wxString GetMSWFaceNameFromHFONT(HFONT hFont);
+#endif // __WXMSW__
+
```
vadz: This should go to `include/wx/msw/private.h` or maybe a new `include/wx/msw/private/font.h`, I don't think there is any need to make it public.

src/msw/fontenum.cpp
```diff
@@ -90,6 +91,22 @@ class wxFontEnumeratorHelper
 // private functions
 // ----------------------------------------------------------------------------
 
+
+// Helper function to get the non-truncated face name from a LOGFONT.
+static wxString GetMSWFaceNameFromLogFont(const LOGFONT* lf)
+{
+    HFONT hFont = ::CreateFontIndirect(lf);
```
vadz:
```suggestion
    AutoHFONT hFont(lf);
```

and then there is no need for `wxON_BLOCK_EXIT`.

src/msw/font.cpp
```diff
@@ -115,7 +115,9 @@ class WXDLLEXPORT wxFontRefData: public wxGDIRefData
     wxString GetFaceName() const
     {
         wxString facename = m_nativeFontInfo.GetFaceName();
-        if ( facename.empty() )
+
+        // If the facename.size() is 31, the facename might be truncated.
+        if ( facename.empty() || facename.size() == (LF_FACESIZE - 1) )
```
vadz: Maybe worth adding a helper for this, e.g. `IsFullMSWFaceName()`?

vadz (commented):
Thanks, this looks good to me and could be merged and even backported to 3.2, as it doesn't affect ABI.

It would be nice if you could please address the various minor comments below. TIA!

include/wx/fontutil.h
moi15moi: I added it in ``include/wx/msw/private.h``

src/msw/fontenum.cpp
moi15moi: Fixed in last commit

src/msw/font.cpp
moi15moi: Why not. I added it in the last commit.

vadz:
I've made a couple of minor changes to simplify this and pushed it in this branch — please let me know if you have any objections, otherwise I'll merge this soon. Thanks!

moi15moi:
If it work and is better in term of maintenance, go for it.

PR #26082, opened 2026-01-08
Add `wxDisplay::GetRawPPI()`

vadz:
See #26080.

### DIFF
--- a/interface/wx/display.h
+++ b/interface/wx/display.h
@@ -168,6 +168,19 @@ public:
     wxSize GetPPI() const;
 
     /**
+        Returns raw display resolution in pixels per inch, i.e.\ without
+        applying any scaling.
+
+        Horizontal and vertical resolution are returned in @c x and @c y
+        components of the wxSize object respectively.
+
+        If the resolution information is not available, returns `wxSize(0, 0)`.
+
+        @since 3.3.2
+     */
+    wxSize GetRawPPI() const;
+
+    /**
         Returns scaling factor used by this display.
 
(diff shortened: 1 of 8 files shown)


interface/wx/display.h
```diff
@@ -167,6 +167,19 @@ class wxDisplay
      */
     wxSize GetPPI() const;
 
+    /**
+        Returns raw display resolution in pixels per inch, i.e.\ without
```
PBfordev, line 171: Should the slash after "i.e." be there?

BTW, how are "raw" pixels different from "physical" pixels, "physical" used [elsewhere in wxWidgets](https://docs.wxwidgets.org/3.2/overview_high_dpi.html#high_dpi_pixel_conversions) when contrasting the "logical"? Is this just device PPI vs pixels in windows and bitmaps terminology?

vadz: > Should the slash after "i.e." be there?

Yes, this is needed to prevent the "brief" part of the comment from stopping at the period.
 
> BTW, how are "raw" pixels different from "physical" pixels, "physical" used [elsewhere in wxWidgets](https://docs.wxwidgets.org/3.2/overview_high_dpi.html#high_dpi_pixel_conversions) when contrasting the "logical"? Is this just device PPI vs pixels in windows and bitmaps terminology?

This is unrelated to the pixels terminology. I agree that this is somewhat confusing but I'm not sure what else to use here.

dsa-t:
Thanks, this looks good to me.

PR #26083, opened 2026-01-09
wxPopupTransientWindow:  don't destroy twice

wsu-cb:
On wxMSW, if the dialogs sample with this [tipwin-dismiss-race.patch](https://github.com/user-attachments/files/24519934/tipwin-dismiss-race.patch) runs in debug mode, then wxWidgets will soon `wxASSERT()` due to destroying a `wxTipWindow` instance twice.

To fix this problem, do not destroy an already-destroyed `wxPopupTransientWindow` .

### DIFF
--- a/include/wx/tipwin.h
+++ b/include/wx/tipwin.h
@@ -34,7 +34,7 @@ public:
     //
     // Note that this is not a wxWeakRef<> because this is set to nullptr when
     // wxTipWindow is closed, which may be "long" before wxTipWindow is
-    // destroyed, bug wxWeakRef<> is set to nullptr on object destruction
+    // destroyed, but wxWeakRef<> is set to nullptr on object destruction
     class WXDLLIMPEXP_CORE Ref
     {
     public:
--- a/src/msw/popupwin.cpp
+++ b/src/msw/popupwin.cpp
@@ -22,6 +22,7 @@
 #if wxUSE_POPUPWIN
 
 #ifndef WX_PRECOMP
+    #include "wx/app.h"
 #endif //WX_PRECOMP
 
 #include "wx/popupwin.h"
@@ -238,14 +239,17 @@ wxPopupTransientWindow::MSWHandleMessage(WXLRESULT *result,
         case WM_ACTIVATE:
             if ( wParam == WA_INACTIVE )
             {
-                // We need to dismiss this window, however doing it directly
-                // from here seems to confuse ::ShowWindow(), which ends up
-                // calling this handler, and may result in losing activation
-                // entirely, so postpone it slightly.
-                //
-                // Also note that the active window hasn't changed yet, so we
-                // postpone calling it until DismissOnDeactivate() is executed.
-                CallAfter(&wxPopupTransientWindow::DismissOnDeactivate);
+                if (!wxTheApp->IsScheduledForDestruction(this))
+                {
+                    // We need to dismiss this window, however doing it directly
+                    // from here seems to confuse ::ShowWindow(), which ends up
+                    // calling this handler, and may result in losing activation
+                    // entirely, so postpone it slightly.
+                    //
+                    // Also note that the active window hasn't changed yet, so we
+                    // postpone calling it until DismissOnDeactivate() is executed.
+                    CallAfter(&wxPopupTransientWindow::DismissOnDeactivate);
+                }
             }
             break;
     }


vadz:
Thanks, but this seems a bit fishy to me, if we can get `WM_ACTIVATE` again, what guarantees that we get it before the window is scheduled for destruction, i.e. couldn't we still end up calling `DismissOnDeactivate()` twice? Is it really harmless?

Anyhow, I'll push this now as this seems to fix a real problem, but I'm far from sure that this is the best (or full) fix...

wsu-cb:
> Thanks, but this seems a bit fishy to me, if we can get `WM_ACTIVATE` again, what guarantees that we get it before the window is scheduled for destruction, i.e. couldn't we still end up calling `DismissOnDeactivate()` twice? Is it really harmless?
> 

A second call to `DismissOnDeactivate()` is NOT harmless:  it will trigger another call to `Destroy()` which is what triggers the `wxASSERT()` this PR is trying to avoid.  I thought about simply changing the `wxASSERT()` to silently return (without adding the `wxWindow` to the schedule-for-destruction list) instead, but I thought the report of suspicious behavior was too useful to remove, which is why I used this approach.

> Anyhow, I'll push this now as this seems to fix a real problem, but I'm far from sure that this is the best (or full) fix...

I won't be surprised if another way to double-`Destroy()` gets found, but I haven't seen it yet.

vadz:
Sorry, I made a typo, I meant to write: "what guarantees that we _don't_ get it before the window is scheduled for destruction". I.e. it looks like there is still a window (of time, not popup) when it can happen... But without debugging this to understand how does it happen exactly I don't know if it really can or not.

sz5000:
wxTipWindow / wxPopupTransientWindow: "Shouldn't destroy the popup twice" assert fires when a queued `CallAfter(DismissOnDeactivate)` runs after the popup has already been Destroy()

```
wxPopupTransientWindowBase::Destroy()                  src/common/popupcmn.cpp:215
wxTipWindow::Close()                                   src/generic/tipwin.cpp:285
wxTipWindow::OnDismiss()                               src/generic/tipwin.cpp:261
wxPopupTransientWindowBase::DismissAndNotify()         src/common/popupcmn.cpp:118
wxPopupTransientWindow::DismissOnDeactivate()          src/msw/popupwin.cpp:200
wxAsyncMethodCallEvent0<…>::Execute()                  include/wx/event.h:1453
wxEvtHandler::TryHereOnly() / TryBeforeAndHere() / ProcessEventLocally() / ProcessEvent()
wxEvtHandler::SafelyProcessEvent()
wxEvtHandler::ProcessPendingEvents()                   src/common/event.cpp:1368
wxAppConsoleBase::ProcessPendingEvents()
wxApp::MSWProcessPendingEventsIfNeeded()
wxIdleWakeUpModule::MsgHookProc()
```

## Sequence that triggers it
A wxTipWindow is on screen (read‑mode annotation hover tooltip in our case).
The user clicks somewhere outside the popup. Windows posts `WM_ACTIVATE WA_INACTIVE` to the popup, and [wxPopupTransientWindow::MSWHandleMessage](vscode-webview://0iaplihkgk4s2c8a4kmccuh801clid4upsfu5k12e8upgevkok79/src/msw/popupwin.cpp#L239-L258) queues

`CallAfter(&wxPopupTransientWindow::DismissOnDeactivate);`
The existing guard

```
if ( wxTheApp->IsScheduledForDestruction(this) )
    break;
```

here only prevents new` WM_ACTIVATE` messages from queueing another CallAfter. It does not look at, or cancel, calls that are already in the pending‑event queue.
Application code then explicitly closes the tooltip:

```
wxTipWindow::Close()      // src/generic/tipwin.cpp:269
    └── Destroy()         // first call — succeeds, appends to wxPendingDelete
```

Idle loop processes pending events. The async `DismissOnDeactivate` from step 2 finally fires on a popup that is already in wxPendingDelete:

```
DismissOnDeactivate()
    └── DismissAndNotify()
        └── wxTipWindow::OnDismiss()
            └── Close()
                └── Destroy()   // SECOND call — wxCHECK_MSG asserts
```

## Reproducer
Minimal MSW pattern:

```
auto* tip = wxTipWindow::New(parent, "hello", 400);
// User clicks outside — Windows posts WM_ACTIVATE WA_INACTIVE,
// CallAfter(DismissOnDeactivate) is queued on `tip`.
tip->Close();   // synchronous Close → Destroy → wxPendingDelete.Append(tip)
```

// Some time later, idle loop processes the queued CallAfter →
// DismissOnDeactivate runs on `tip` → Close → Destroy → assert.
Any code path that calls wxTipWindow::Close() (or otherwise destroys a wxPopupTransientWindow) between the time the OS posted `WM_ACTIVATE WA_INACTIVE` and the time the queued DismissOnDeactivate runs hits this assert.


## Suggested fix
Guard inside DismissOnDeactivate itself, mirroring the guard already present in MSWHandleMessage:

```
void wxPopupTransientWindow::DismissOnDeactivate()
{
    // If we were destroyed between the time WM_ACTIVATE queued this
    // CallAfter and now, do nothing — the popup is already on its way out.
    if ( wxTheApp && wxTheApp->IsScheduledForDestruction(this) )
        return;

    DismissAndNotify();
    // …existing owner re-activation code…
}
```

This keeps the existing `MSWHandleMessage` guard (cheap fast path that avoids queueing) and adds a symmetrical guard at the execution side. Both `IsScheduledForDestruction `checks rely on the same wxPendingDelete list `Destroy()` populates, so they stay in sync.

wsu-cb:
> ## Reproducer
> Minimal MSW pattern:
> 
> ```
> auto* tip = wxTipWindow::New(parent, "hello", 400);
> // User clicks outside — Windows posts WM_ACTIVATE WA_INACTIVE,
> // CallAfter(DismissOnDeactivate) is queued on `tip`.
> tip->Close();   // synchronous Close → Destroy → wxPendingDelete.Append(tip)
> ```
> 
> // Some time later, idle loop processes the queued CallAfter → // DismissOnDeactivate runs on `tip` → Close → Destroy → assert. Any code path that calls wxTipWindow::Close() (or otherwise destroys a wxPopupTransientWindow) between the time the OS posted `WM_ACTIVATE WA_INACTIVE` and the time the queued DismissOnDeactivate runs hits this assert.
> 

After the user click, the idle loop should run "very soon".  How is the application code able to call `tip->Close()` between the user click and the idle loop?  Is there an event handler for the user click?

PR #26084, opened 2026-01-09
Make selection rect rounded as per OS X look

RobertRoeb:
- So far only do this for tree control single line
- Up for discussion, it is a matter of native look vs taste

### DIFF
--- a/include/wx/renderer.h
+++ b/include/wx/renderer.h
@@ -77,7 +77,7 @@ enum
     wxCONTROL_DIRTY      = 0x80000000,
 
     // used to indicate a round selection on e.g. macOS
-    wxCONTROL_SELECTION_ROUND  = 0x40000000,
+    wxCONTROL_SELECTION_GROUP  = 0x40000000,
 
     // used to indicate a sub item within e.g. selections or header controls parts
     wxCONTROL_ITEM_FIRST       = wxCONTROL_SPECIAL,
--- a/include/wx/renderer.h
+++ b/include/wx/renderer.h
@@ -77,13 +77,12 @@ enum
     wxCONTROL_DIRTY      = 0x80000000,
 
     // used to indicate a round selection on e.g. macOS
-    wxCONTROL_SELECTION_GROUP  = 0x10000000,
+    wxCONTROL_SELECTION_GROUP  = 0x20000000,
 
     // used to indicate a sub item within e.g. selections or header controls parts
-    wxCONTROL_ITEM_FIRST       = 0x20000000,
-    wxCONTROL_ITEM_MIDDLE      = 0x40000000,
+    wxCONTROL_ITEM_FIRST       = 0x40000000,
     wxCONTROL_ITEM_LAST        = wxCONTROL_DIRTY,
-    wxCONTROL_ITEM_MASK        = wxCONTROL_ITEM_FIRST | wxCONTROL_ITEM_MIDDLE | wxCONTROL_ITEM_LAST
+    // wxCONTROL_ITEM_MIDDLE   = !wxCONTROL_ITEM_FIRST && !wxCONTROL_ITEM_LAST,
 };
 
 // title bar buttons supported by DrawTitleBarBitmap()
--- a/include/wx/renderer.h
+++ b/include/wx/renderer.h
@@ -74,17 +74,16 @@ enum
 
     // this is a pseudo flag not used directly by wxRenderer but rather by some
     // controls internally
-    wxCONTROL_DIRTY      = 0x80000000
-};
+    wxCONTROL_DIRTY      = 0x80000000,
 
-// control state flags for different selection rectangle (on e.g. macOS)
-enum
-{
-    wxRENDER_SELECTION_SINGLE_ROUND  = 0x00001000,  // single round selection
-    wxRENDER_SELECTION_FIRST_ROUND   = 0x00002000,  // top of multiple round selection
-    wxRENDER_SELECTION_MIDDLE_ROUND  = 0x00004000,  // mid of multiple round selection
-    wxRENDER_SELECTION_LAST_ROUND    = 0x00008000,  // bottom of multiple round selection
-    wxRENDER_SELECTION_MASK          = 0x0000F000,  // bottom of multiple round selection
+    // used to indicate a round selection on e.g. macOS
+    wxCONTROL_SELECTION_ROUND  = 0x40000000,
+
+    // used to indicate a sub item within e.g. selections or header controls parts
+    wxCONTROL_ITEM_FIRST       = wxCONTROL_SPECIAL,
+    wxCONTROL_ITEM_MIDDLE      = 0x20000000,
+    wxCONTROL_ITEM_LAST        = wxCONTROL_DIRTY,
+    wxCONTROL_ITEM_MASK        = wxCONTROL_ITEM_FIRST | wxCONTROL_ITEM_MIDDLE | wxCONTROL_ITEM_LAST
 };
 
 // title bar buttons supported by DrawTitleBarBitmap()
--- a/include/wx/renderer.h
+++ b/include/wx/renderer.h
@@ -77,11 +77,11 @@ enum
     wxCONTROL_DIRTY      = 0x80000000,
 
     // used to indicate a round selection on e.g. macOS
-    wxCONTROL_SELECTION_GROUP  = 0x40000000,
+    wxCONTROL_SELECTION_GROUP  = 0x10000000,
 
     // used to indicate a sub item within e.g. selections or header controls parts
-    wxCONTROL_ITEM_FIRST       = wxCONTROL_SPECIAL,
-    wxCONTROL_ITEM_MIDDLE      = 0x20000000,
+    wxCONTROL_ITEM_FIRST       = 0x20000000,
+    wxCONTROL_ITEM_MIDDLE      = 0x40000000,
     wxCONTROL_ITEM_LAST        = wxCONTROL_DIRTY,
     wxCONTROL_ITEM_MASK        = wxCONTROL_ITEM_FIRST | wxCONTROL_ITEM_MIDDLE | wxCONTROL_ITEM_LAST
 };
--- a/include/wx/renderer.h
+++ b/include/wx/renderer.h
@@ -77,6 +77,16 @@ enum
     wxCONTROL_DIRTY      = 0x80000000
 };
 
+// control state flags for different selection rectangle (on e.g. macOS)
+enum
+{
+    wxRENDER_SELECTION_SINGLE_ROUND  = 0x00001000,  // single round selection
+    wxRENDER_SELECTION_FIRST_ROUND   = 0x00002000,  // top of multiple round selection
+    wxRENDER_SELECTION_MIDDLE_ROUND  = 0x00004000,  // mid of multiple round selection
+    wxRENDER_SELECTION_LAST_ROUND    = 0x00008000,  // bottom of multiple round selection
+    wxRENDER_SELECTION_MASK          = 0x0000F000,  // bottom of multiple round selection
+};
+
 // title bar buttons supported by DrawTitleBarBitmap()
 //
 // NB: they have the same values as wxTOPLEVEL_BUTTON_XXX constants in
--- a/interface/wx/renderer.h
+++ b/interface/wx/renderer.h
@@ -62,7 +62,14 @@ enum
     wxCONTROL_CHECKABLE  = 0x00000080,
 
     /** (Check) undetermined state. */
-    wxCONTROL_UNDETERMINED = wxCONTROL_CHECKABLE
+    wxCONTROL_UNDETERMINED = wxCONTROL_CHECKABLE,
+
+    /** Indicated that the control might display a group selection across several lines */
+    wxCONTROL_SELECTION_GROUP  = 0x20000000,
+
+    /** For group selections indicates the position in the group */
+    wxCONTROL_ITEM_FIRST       = 0x40000000,
+    wxCONTROL_ITEM_LAST        = 0x80000000,
 };
 
 /**
--- a/src/generic/listctrl.cpp
+++ b/src/generic/listctrl.cpp
@@ -636,11 +636,12 @@ void wxListLineData::SetAttr(wxItemAttr *attr)
 void wxListLineData::ApplyAttributes(wxDC *dc,
                                      const wxRect& rectHL,
                                      bool highlighted,
-                                     bool current)
+                                     bool current,
+                                     long item )
 {
     const wxItemAttr * const attr = GetAttr();
 
-    wxWindow * const listctrl = m_owner->GetParent();
+    wxListCtrl * const listctrl = (wxListCtrl* const) m_owner->GetParent();
 
     const bool hasFocus = listctrl->HasFocus();
 
@@ -690,9 +691,43 @@ void wxListLineData::ApplyAttributes(wxDC *dc,
             flags |= wxCONTROL_FOCUSED;
         if (current)
            flags |= wxCONTROL_CURRENT;
-        if (listctrl->HasFlag(wxLC_SINGLE_SEL)) {
-            // TODO: extend this later to selecting groups of items
-            flags |= wxCONTROL_SELECTION_GROUP;
+        wxListMainWindow *mainWindow = listctrl->m_mainWin;
+
+        if (listctrl->HasFlag(wxLC_SINGLE_SEL) || (listctrl->GetItemCount() < 2)) 
+        {
+            if (mainWindow->GetClientSize().x >= mainWindow->GetVirtualSize().x)
+            {
+                flags |= wxCONTROL_SELECTION_GROUP;
+            }
+        }
+        else
+        {
+            if (mainWindow->GetClientSize().x >= mainWindow->GetVirtualSize().x)
+            {
+                bool isTop = ((item == 0) ||
+                            (listctrl->GetItemState( item-1, wxLIST_STATE_SELECTED) == 0));
+                bool isBottom = ((item == listctrl->GetItemCount()-1) ||
+                                (listctrl->GetItemState( item+1, wxLIST_STATE_SELECTED) == 0));
+                bool isMiddle = ((item != 0) && (item != listctrl->GetItemCount()-1) &&
+                                (listctrl->GetItemState( item-1, wxLIST_STATE_SELECTED) != 0) &&
+                                (listctrl->GetItemState( item+1, wxLIST_STATE_SELECTED) != 0) );
+                if (isTop && isBottom)
+                {
+                    flags |= wxCONTROL_SELECTION_GROUP;
+                }
+                else if (isTop)
+                {
+                    flags |= wxCONTROL_ITEM_FIRST;
+                }
+                else if (isBottom)
+                {
+                    flags |= wxCONTROL_ITEM_LAST;
+                }
+                else if (isMiddle)
+                {
+                    flags |= wxCONTROL_ITEM_MIDDLE;
+                }
+            }
         }
         wxRendererNative::Get().
             DrawItemSelectionRect( m_owner, *dc, rectHL, flags );
@@ -715,11 +750,11 @@ void wxListLineData::ApplyAttributes(wxDC *dc,
 #endif
 }
 
-void wxListLineData::Draw(wxDC *dc, bool current)
+void wxListLineData::Draw(wxDC *dc, bool current, long line )
 {
     wxCHECK_RET( !m_items.empty(), wxT("no subitems at all??") );
 
-    ApplyAttributes(dc, m_gi->m_rectHighlight, IsHighlighted(), current);
+    ApplyAttributes(dc, m_gi->m_rectHighlight, IsHighlighted(), current, line );
 
     wxListItemData* const item = &m_items[0];
     if (item->HasImage())
@@ -745,7 +780,8 @@ void wxListLineData::DrawInReportMode( wxDC *dc,
                                        const wxRect& rectHL,
                                        bool highlighted,
                                        bool current,
-                                       bool checked )
+                                       bool checked,
+                                       long line )
 {
     // TODO: later we should support setting different attributes for
     //       different columns - to do it, just add "col" argument to
@@ -753,7 +789,7 @@ void wxListLineData::DrawInReportMode( wxDC *dc,
 
     // Note: GetSubItemRect() needs to be modified if the layout here changes.
 
-    ApplyAttributes(dc, rectHL, highlighted, current);
+    ApplyAttributes(dc, rectHL, highlighted, current, line );
 
     wxCoord x = rect.x;
     wxCoord yMid = rect.y + rect.height/2;
@@ -1893,6 +1929,12 @@ bool wxListMainWindow::HighlightLine( size_t line, bool highlight, SendEvent sen
 
 void wxListMainWindow::RefreshLine( size_t line )
 {
+#ifdef __WXOSX__
+    // A change in the selection of a row can influence
+    // the look of the selection of another row so we
+    // always need to refresh() the whole window
+    Refresh();
+#else
     if ( InReportView() )
     {
         size_t visibleFrom, visibleTo;
@@ -1906,10 +1948,17 @@ void wxListMainWindow::RefreshLine( size_t line )
 
     GetListCtrl()->CalcScrolledPosition( rect.x, rect.y, &rect.x, &rect.y );
     RefreshRect( rect );
+#endif
 }
 
 void wxListMainWindow::RefreshLines( size_t lineFrom, size_t lineTo )
 {
+#ifdef __WXOSX__
+    // A change in the selection of a row can influence
+    // the look of the selection of another row so we
+    // always need to refresh() the whole window
+    Refresh();
+#else
     // we suppose that they are ordered by caller
     wxASSERT_MSG( lineFrom <= lineTo, wxT("indices in disorder") );
 
@@ -1949,10 +1998,17 @@ void wxListMainWindow::RefreshLines( size_t lineFrom, size_t lineTo )
             RefreshLine(line);
         }
     }
+#endif
 }
 
 void wxListMainWindow::RefreshAfter( size_t lineFrom )
 {
+#ifdef __WXOSX__
+    // A change in the selection of a row can influence
+    // the look of the selection of another row so we
+    // always need to refresh() the whole window
+    Refresh();
+#else
     if ( InReportView() )
     {
         // Note that we don't compare lineFrom with the last visible line
@@ -1984,10 +2040,17 @@ void wxListMainWindow::RefreshAfter( size_t lineFrom )
         // TODO: how to do it more efficiently?
         m_dirty = true;
     }
+#endif
 }
 
 void wxListMainWindow::RefreshSelected()
 {
+#ifdef __WXOSX__
+    // A change in the selection of a row can influence
+    // the look of the selection of another row so we
+    // always need to refresh() the whole window
+    Refresh();
+#else
     if ( IsEmpty() )
         return;
 
@@ -2011,6 +2074,7 @@ void wxListMainWindow::RefreshSelected()
         if ( line != m_current && IsHighlighted(line) )
             RefreshLine(line);
     }
+#endif
 }
 
 void wxListMainWindow::OnPaint( wxPaintEvent &WXUNUSED(event) )
@@ -2115,7 +2179,8 @@ void wxListMainWindow::OnPaint( wxPaintEvent &WXUNUSED(event) )
                                              GetLineHighlightRect(line),
                                              IsHighlighted(line),
                                              line == m_current,
-                                             IsItemChecked(line) );
+                                             IsItemChecked(line),
+                                             line );
         }
 
         if ( HasFlag(wxLC_HRULES) )
@@ -2180,7 +2245,7 @@ void wxListMainWindow::OnPaint( wxPaintEvent &WXUNUSED(event) )
         size_t count = GetItemCount();
         for ( size_t i = 0; i < count; i++ )
         {
-            GetLine(i)->Draw( &dc, i == m_current );
+            GetLine(i)->Draw( &dc, i == m_current, i );
         }
     }
 
--- a/src/generic/listctrl.cpp
+++ b/src/generic/listctrl.cpp
@@ -709,9 +709,6 @@ void wxListLineData::ApplyAttributes(wxDC *dc,
                             (listctrl->GetItemState( item-1, wxLIST_STATE_SELECTED) == 0));
                 bool isBottom = ((item == listctrl->GetItemCount()-1) ||
                                 (listctrl->GetItemState( item+1, wxLIST_STATE_SELECTED) == 0));
-                bool isMiddle = ((item != 0) && (item != listctrl->GetItemCount()-1) &&
-                                (listctrl->GetItemState( item-1, wxLIST_STATE_SELECTED) != 0) &&
-                                (listctrl->GetItemState( item+1, wxLIST_STATE_SELECTED) != 0) );
                 if (isTop && isBottom)
                 {
                     flags |= wxCONTROL_SELECTION_GROUP;
@@ -724,7 +721,7 @@ void wxListLineData::ApplyAttributes(wxDC *dc,
                 {
                     flags |= wxCONTROL_ITEM_LAST;
                 }
-                else if (isMiddle)
+                else
                 {
                     flags |= wxCONTROL_ITEM_MIDDLE;
                 }
@@ -1928,14 +1925,16 @@ bool wxListMainWindow::HighlightLine( size_t line, bool highlight, SendEvent sen
     return changed;
 }
 
-void wxListMainWindow::RefreshLine( size_t line )
-{
 #ifdef __WXOSX__
+void wxListMainWindow::RefreshLine( size_t WXUNUSED(line) )
+{
     // A change in the selection of a row can influence
     // the look of the selection of another row so we
     // always need to refresh() the whole window
     Refresh();
 #else
+void wxListMainWindow::RefreshLine( size_t line )
+{
     if ( InReportView() )
     {
         size_t visibleFrom, visibleTo;
@@ -1952,14 +1951,16 @@ void wxListMainWindow::RefreshLine( size_t line )
 #endif
 }
 
-void wxListMainWindow::RefreshLines( size_t lineFrom, size_t lineTo )
-{
 #ifdef __WXOSX__
+void wxListMainWindow::RefreshLines( size_t WXUNUSED(lineFrom), size_t WXUNUSED(lineTo) )
+{
     // A change in the selection of a row can influence
     // the look of the selection of another row so we
     // always need to refresh() the whole window
     Refresh();
 #else
+void wxListMainWindow::RefreshLines( size_t lineFrom, size_t lineTo )
+{
     // we suppose that they are ordered by caller
     wxASSERT_MSG( lineFrom <= lineTo, wxT("indices in disorder") );
 
@@ -2004,12 +2005,6 @@ void wxListMainWindow::RefreshLines( size_t lineFrom, size_t lineTo )
 
 void wxListMainWindow::RefreshAfter( size_t lineFrom )
 {
-#ifdef __WXOSX__
-    // A change in the selection of a row can influence
-    // the look of the selection of another row so we
-    // always need to refresh() the whole window
-    Refresh();
-#else
     if ( InReportView() )
     {
         // Note that we don't compare lineFrom with the last visible line
@@ -2022,6 +2017,11 @@ void wxListMainWindow::RefreshAfter( size_t lineFrom )
 
         if ( lineFrom < visibleFrom )
             lineFrom = visibleFrom;
+#ifdef __WXOSX__
+        // Deleting a line can change the look of the selection of the 
+        // line above on macOS
+        if (lineFrom) lineFrom--;
+#endif
 
         wxRect rect;
         rect.x = 0;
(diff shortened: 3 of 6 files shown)


src/osx/cocoa/renderer.mm
```diff
@@ -603,6 +604,14 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
 
     wxDCPenChanger setPen(dc, *wxTRANSPARENT_PEN);
     wxDCBrushChanger setBrush(dc, selBrush);
+    if (win->IsKindOf(wxCLASSINFO(wxTreeCtrl)))
```
vadz: Sorry but let's please not do this. AFAICS this function doesn't use `wxCONTROL_SPECIAL` so the simplest alternative would be to add it when using multiple selections. Or, if this doesn't work, we could add `wxCONTROL_MULTIPLE`. But using RTTI for this is just wrong for the reasons which I think are pretty clear.

vadz (changes requested):
Please change this to drive the behaviour from the caller instead of using RTTI. TIA!

src/osx/cocoa/renderer.mm
RobertRoeb: I didn't want to re-use an unspecific flag, so I introduced new ones. Otherwise did as you suggested and moved the test to the tree control.

src/osx/cocoa/renderer.mm
```diff
@@ -603,7 +604,14 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
 
     wxDCPenChanger setPen(dc, *wxTRANSPARENT_PEN);
     wxDCBrushChanger setBrush(dc, selBrush);
-    dc.DrawRectangle( rect );
+    if ((flags & wxRENDER_SELECTION_SINGLE_ROUND) != 0)
+    {
+        dc.DrawRoundedRectangle( rect, 8 );
```
vadz: I wonder if there is some way to get this `8` from the system?

include/wx/renderer.h
```diff
@@ -77,6 +77,16 @@ enum
...
+// control state flags for different selection rectangle (on e.g. macOS)
+enum
+{
+    wxRENDER_SELECTION_SINGLE_ROUND  = 0x00001000,  // single round selection
+    wxRENDER_SELECTION_FIRST_ROUND   = 0x00002000,  // top of multiple round selection
+    wxRENDER_SELECTION_MIDDLE_ROUND  = 0x00004000,  // mid of multiple round selection
+    wxRENDER_SELECTION_LAST_ROUND    = 0x00008000,  // bottom of multiple round selection
+    wxRENDER_SELECTION_MASK          = 0x0000F000,  // bottom of multiple round selection
+};
+
```
vadz: Sorry, I don't understand why:

1. Is this separate from the main enum, when it's passed inside the same `int`.
2. Do we need all the unused flags.

I'd like to keep just the first one and move it into the enum above and call it `wxCONTROL_MULTIPLE` as proposed before because it's not the fact that it's rounded which is important (this could be different in implementations for the other platforms or even in the future versions of macOS itself), but that it's the selection style used for multiple selection.

I can do this myself if you agree. And if you don't, please explain why.

vadz (commented):
Thanks for the changes but I still have the natural questions about the new flags. Please let me know if I'm missing something which explains why was this approach taken instead of the obvious one.

include/wx/renderer.h
RobertRoeb: I don't care in which enum you move it, but one day, we will need all of these as the first selected item (of a continuous group of selected items) only has rounded corners on top, the last selected item (of the same continuous group of selected items) only has rounded corners at the bottom and all middle ones have no rounded corners, but may look different than classical non-rounded ones (maybe not).

MultipleSelectionsOSX

vadz: There is no way we're going to handle this by calling `DrawItemSelectionRect()` with different flags for the first, middle and last items, it clearly should be called just once for each contiguous selection fragment to draw the entire rectangle at once, so there is definitely not a good reason to have these elements.

Also, while I definitely see it, I can't help wondering _why_ you don't care about which enum is used when there is obviously the natural and, dare I say, only reasonable, choice of using a single enum for all the flags passed as a single parameter and whatever you did. Could you please care a bit more about either choosing the dull and obvious solution or explaining why you don't do it instead of leaving me questioning my sanity?

RobertRoeb: It is because of the name wxCONTROL_SOMETHING. I need different selection drawing based on items position in the same control

RobertRoeb: > There is no way we're going to handle this by calling `DrawItemSelectionRect()` with different flags for the first, middle and last items, 

Well, that would have been easy to do. Not sure what the problem is

vadz: > It is because of the name wxCONTROL_SOMETHING. I need different selection drawing based on items position in the same control

What's wrong with `wxCONTROL_SELECTION_FIRST` etc if you still think you need this?

paulcor:
`DrawHeaderButton()` uses `wxCONTROL_SPECIAL` to indicate the first button, and `wxCONTROL_DIRTY` to indicate the last.

RobertRoeb:
> `DrawHeaderButton()` uses `wxCONTROL_SPECIAL` to indicate
> the first button, and `wxCONTROL_DIRTY` to indicate the last.

That does not seem very intuitive to me, so may I propose this:

   // used to indicate a round selection on e.g. macOS
    wxCONTROL_SELECTION_ROUND  = 0x40000000,

    // used to indicate a sub item within e.g. selections or header controls parts
    wxCONTROL_ITEM_FIRST       = wxCONTROL_SPECIAL,
    wxCONTROL_ITEM_MIDDLE      = 0x20000000,
    wxCONTROL_ITEM_LAST        = wxCONTROL_DIRTY,
    wxCONTROL_ITEM_MASK        = wxCONTROL_ITEM_FIRST | wxCONTROL_ITEM_MIDDLE | wxCONTROL_ITEM_LAST

RobertRoeb:
This is what how it can look now, side by side with native:
Multi wx
Multi Native

src/generic/listctrl.cpp
```diff
@@ -690,6 +690,9 @@ void wxListLineData::ApplyAttributes(wxDC *dc,
             flags |= wxCONTROL_FOCUSED;
         if (current)
            flags |= wxCONTROL_CURRENT;
+        if (listctrl->HasFlag(wxLC_SINGLE_SEL)) {
```
vadz: Isn't this backwards?

include/wx/renderer.h
```diff
@@ -74,7 +74,16 @@ enum
 
     // this is a pseudo flag not used directly by wxRenderer but rather by some
     // controls internally
-    wxCONTROL_DIRTY      = 0x80000000
+    wxCONTROL_DIRTY      = 0x80000000,
+
+    // used to indicate a round selection on e.g. macOS
+    wxCONTROL_SELECTION_ROUND  = 0x40000000,
```
vadz: Sorry but I'd still really, really like to call it `wxCONTROL_MULTIPLE` (or `_SINGLE` if you prefer) because the flags should indicate the semantics, not the appearance.

src/generic/treectlg.cpp
```diff
@@ -2562,12 +2562,20 @@ void wxGenericTreeCtrl::PaintItem(wxGenericTreeItem *item, wxDC& dc)
...
 
     if ( item->IsSelected() )
     {
         dc.SetBrush(m_hasFocus ? m_hilightBrush : m_hilightUnfocusedBrush);
         drawItemBackground = true;
+
+        if (!HasFlag(wxTR_MULTIPLE))
+        {
+            // Allow macOS renderer to draw a rounded rect, but only
+            // when no horizontal scrolling is needed
```
vadz: Can we check this in the renderer instead? Again, this is clearly just not the right place/layer to do it.

src/osx/cocoa/renderer.mm
```diff
@@ -601,9 +602,56 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
     }
     wxBrush selBrush( col );
 
+    // macOS NSTableView has whitespace left and right of the first and last
+    // data columns and only draws the border where the actual columns begins.
+    // wxWidgets generic controls don't do this, so add a bit of distance to
+    // the border
+    int distanceFromBorder = 8;
```
vadz:
```suggestion
    constexpr int distanceFromBorder = 8; 
```

src/osx/cocoa/renderer.mm
```diff
@@ -601,9 +602,56 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
...
+    // wxWidgets generic controls don't do this, so add a bit of distance to
+    // the border
+    int distanceFromBorder = 8; 
+
     wxDCPenChanger setPen(dc, *wxTRANSPARENT_PEN);
     wxDCBrushChanger setBrush(dc, selBrush);
-    dc.DrawRectangle( rect );
+    if ((flags & wxCONTROL_SELECTION_ROUND) != 0)
+    {
+        dc.DrawRoundedRectangle( rect, 8 );
```
vadz: Is this a different `8`? Or should it be `distanceFromBorder`?

Same question for all the other `8`s below.

src/osx/cocoa/renderer.mm
```diff
@@ -601,9 +602,56 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
...
+    }
+    else
+    if ((flags & wxCONTROL_ITEM_FIRST) != 0)
```
vadz:
```suggestion
    else if ((flags & wxCONTROL_ITEM_FIRST) != 0)
```

src/osx/cocoa/renderer.mm
```diff
@@ -601,9 +602,56 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
...
+    {
+        dc.DrawRoundedRectangle( rect, 8 );
+        wxRect bottomrect = rect;
+        bottomrect.y = bottomrect.y + bottomrect.height-8;
+        bottomrect.height = 8;
+        dc.DrawRectangle( bottomrect );
+    }
+    else
+    if ((flags & wxCONTROL_ITEM_LAST) != 0)
```
vadz:
```suggestion
    else if ((flags & wxCONTROL_ITEM_LAST) != 0)
```

src/osx/cocoa/renderer.mm
```diff
@@ -601,9 +602,56 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
...
+        dc.DrawRoundedRectangle( rect, 8 );
+        if (rect.width > distanceFromBorder*2)
+        {
+            wxColour lineColour = col.ChangeLightness( ((flags & wxCONTROL_FOCUSED) != 0) ? 125 : 95 );
+            dc.SetPen( wxPen( lineColour ));
+            dc.DrawLine( rect.x + distanceFromBorder, rect.y, rect.x + rect.width - distanceFromBorder, rect.y );
+        }
+    }
+    else
+    if ((flags & wxCONTROL_ITEM_MIDDLE) != 0)
```
vadz:
```suggestion
    else if ((flags & wxCONTROL_ITEM_MIDDLE) != 0)
```

src/osx/cocoa/renderer.mm
```diff
@@ -27,6 +27,7 @@
 #include "wx/graphics.h"
 #include "wx/dcgraph.h"
 #include "wx/splitter.h"
+#include "wx/treectrl.h"
```
vadz:
```suggestion
```

vadz (commented):
Is there really no native function for drawing the selection rectangle? I'm absolutely sure that this code will result in the selection looking out of place in the next macOS version as this is exactly the kind of things they like tweaking.

include/wx/renderer.h
```diff
@@ -74,7 +74,16 @@ enum
...
+
+    // used to indicate a sub item within e.g. selections or header controls parts
+    wxCONTROL_ITEM_FIRST       = wxCONTROL_SPECIAL,
+    wxCONTROL_ITEM_MIDDLE      = 0x20000000,
```
paulcor: It seems like "middle" can be inferred as the lack of "first" or "last"?

src/generic/listctrl.cpp
RobertRoeb: No, this tests for the easy case, single selection (and possibly no horizontal scrolling) in which case it is round in macOS. I excluded multiple selection mode in a previous variant of the PR because it  patch could not handle multiple row yet.

src/generic/treectlg.cpp
RobertRoeb: The renderer does not know anything about the scrolling state of the parent window. As you wrote further up, this should be done in the caller.

include/wx/renderer.h
RobertRoeb: I cannot promise I have found all scenarios. There is also full row highlight vs. partial, scrolled vs. fits on client area. Tree vs list.

src/osx/cocoa/renderer.mm
RobertRoeb: One is the radius of the selection rectangle that I figured out testing all individual numbers and using a 4x magnification, the other is the white space that the native control leaves left of the actual data. Looking at this again right one, this white space might be the unused space from rows that have no chevron. Not easy to deduce the logic...

vadz:
To summarize:

1. I'm **strongly** opposed to adding `wxCONTROL_SELECTION_ROUND` flag. Not only it won't describe what it does under macOS when they change the selection style to use animated marquee in the next version, but it is also completely unhelpful for the caller as nobody has any idea when to use it and when not to. This conforms to the letter of the rule about using `wxRendererNative` for platform-specific UI elements while completely disregarding its spirit and is really worse than not using `wxRendererNative` at all. If there is really no rule for when it should be used (i.e. if it can't be called `wxCONTROL_MULTIPLE_SELECTION`), just forget about the renderer, add a private `wxDrawSelectionRectangleUnderMac()` function and call it from the generic code inside `#if` check.
2. Please give some names to the constants used in the drawing code, it is completely incomprehensible and unmaintainable without them. It won't be great with them either, but the fact that this code is messy is a good reason to make a special effort to make it more clear, not a reason to abandon any attempt to do so.

paulcor:
> > `DrawHeaderButton()` uses `wxCONTROL_SPECIAL` to indicate
> > the first button, and `wxCONTROL_DIRTY` to indicate the last.
> 
> That does not seem very intuitive to me

True, but somewhat ironic, as it seems to have been done by you...

> wxCONTROL_ITEM_FIRST       = wxCONTROL_SPECIAL,

It looks like `wxCONTROL_SPECIAL` is already aliased to `wxCONTROL_CELL`, which is also used by `DrawItemSelectionRect()`, so `wxCONTROL_CELL` would presumably need to get a different value.

RobertRoeb:
> > > `DrawHeaderButton()` uses `wxCONTROL_SPECIAL` to indicate
> > > the first button, and `wxCONTROL_DIRTY` to indicate the last.
> > 
> > 
> > That does not seem very intuitive to me
> 
> True, but somewhat ironic, as it seems to have been done by you...

I have a very stable personality: already in the early Middle Ages, when
I wrote that code, I tried to avoid lengthy discussions on the naming
of constants.

RobertRoeb:
> Not only it won't describe what it does under macOS when they change the 
> selection style to use animated marquee in the next version

Fair enough. I guess the rounded selection graphic is to indicate a group,
so I changed it to that.

vadz:
> > Not only it won't describe what it does under macOS when they change the
> > selection style to use animated marquee in the next version
> 
> Fair enough. I guess the rounded selection graphic is to indicate a group, so I changed it to that.

Sorry, I don't see any change here, did you push it?

RobertRoeb:
I had forgotton to push the change. Here is how it looks now in wxListCtrl.

ListCtrl Tahoe

include/wx/renderer.h
```diff
@@ -74,7 +74,16 @@ enum
 
     // this is a pseudo flag not used directly by wxRenderer but rather by some
     // controls internally
-    wxCONTROL_DIRTY      = 0x80000000
+    wxCONTROL_DIRTY      = 0x80000000,
+
+    // used to indicate a round selection on e.g. macOS
+    wxCONTROL_SELECTION_GROUP  = 0x40000000,
```
paulcor: This isn't really needed at all, it's logically just `wxCONTROL_ITEM_FIRST | wxCONTROL_ITEM_LAST`

src/generic/listctrl.cpp
```diff
@@ -690,6 +691,45 @@ void wxListLineData::ApplyAttributes(wxDC *dc,
             flags |= wxCONTROL_FOCUSED;
         if (current)
            flags |= wxCONTROL_CURRENT;
+        wxListMainWindow *mainWindow = listctrl->m_mainWin;
+
+        if (listctrl->HasFlag(wxLC_SINGLE_SEL) || (listctrl->GetItemCount() < 2) ||
+            !listctrl->HasFlag(wxLC_REPORT) )
+        {
+            if (mainWindow->GetClientSize().x >= mainWindow->GetVirtualSize().x)
```
paulcor: I believe the whole row is always drawn, and the `wxDC` is offset to account for horizontal scrolling, so this check should not be needed.

include/wx/renderer.h
```diff
@@ -74,7 +74,16 @@ enum
...
+
+    // used to indicate a sub item within e.g. selections or header controls parts
+    wxCONTROL_ITEM_FIRST       = wxCONTROL_SPECIAL,
```
paulcor: As I said earlier, `wxCONTROL_ITEM_FIRST` and `wxCONTROL_CELL` can't share the same value, they are both used by `DrawItemSelectionRect()`.

src/generic/listctrl.cpp
```diff
@@ -690,6 +691,45 @@ void wxListLineData::ApplyAttributes(wxDC *dc,
...
+            {
+                bool isTop = ((item == 0) ||
+                            (listctrl->GetItemState( item-1, wxLIST_STATE_SELECTED) == 0));
+                bool isBottom = ((item == listctrl->GetItemCount()-1) ||
+                                (listctrl->GetItemState( item+1, wxLIST_STATE_SELECTED) == 0));
+                bool isMiddle = ((item != 0) && (item != listctrl->GetItemCount()-1) &&
+                                (listctrl->GetItemState( item-1, wxLIST_STATE_SELECTED) != 0) &&
+                                (listctrl->GetItemState( item+1, wxLIST_STATE_SELECTED) != 0) );
```
paulcor: I'm pretty sure this is just `bool isMiddle = !isTop && !isBottom`

src/generic/listctrl.cpp
```diff
@@ -690,6 +691,45 @@ void wxListLineData::ApplyAttributes(wxDC *dc,
...
+                }
+                else if (isTop)
+                {
+                    flags |= wxCONTROL_ITEM_FIRST;
+                }
+                else if (isBottom)
+                {
+                    flags |= wxCONTROL_ITEM_LAST;
+                }
+                else if (isMiddle)
```
paulcor: If I'm right about `isMiddle`, then this is always true.

src/generic/listctrl.cpp
```diff
@@ -1980,10 +2041,17 @@ void wxListMainWindow::RefreshAfter( size_t lineFrom )
         // TODO: how to do it more efficiently?
         m_dirty = true;
     }
+#endif
 }
 
 void wxListMainWindow::RefreshSelected()
 {
+#ifdef __WXOSX__
+    // A change in the selection of a row can influence
```
paulcor: `RefreshSelected()` is only called for a focus change, the selection is not changing here.

src/generic/listctrl.cpp
```diff
@@ -1945,10 +1999,17 @@ void wxListMainWindow::RefreshLines( size_t lineFrom, size_t lineTo )
...
+#endif
 }
 
 void wxListMainWindow::RefreshAfter( size_t lineFrom )
 {
+#ifdef __WXOSX__
+    // A change in the selection of a row can influence
+    // the look of the selection of another row so we
+    // always need to refresh() the whole window
+    Refresh();
```
paulcor: `RefreshAfter()` is called for a deletion, so it seems like extending the refresh by one line would be sufficient, something like
```
if (lineFrom) lineFrom--;
```

src/generic/listctrl.cpp
RobertRoeb: No, macOS converts to full row and square selection boxes also on the side that is not scrolled off screen

include/wx/renderer.h
RobertRoeb: Will do

src/generic/listctrl.cpp
RobertRoeb: DrawItemSelectionRect() needs to know if is 
- a single line with a normal rectangle and no divider line above
- a single line with a normal rectangle and a divider line above
- a rounded rect single line
- a rounded rect top line
- a rounded rect bottom line
We need more than 2 bits to determine that and I think the flags names now a good.

src/generic/listctrl.cpp
RobertRoeb: Right

src/generic/listctrl.cpp
RobertRoeb: I think you are right

src/generic/listctrl.cpp
RobertRoeb: Right

src/generic/listctrl.cpp
paulcor: > DrawItemSelectionRect() needs to know if is
> 
>     * a single line with a normal rectangle and no divider line above

This would be no `wxCONTROL_ITEM_*` flags

>     * a single line with a normal rectangle and a divider line above

This would be `wxCONTROL_ITEM_MIDDLE`

>     * a rounded rect single line

This would be `wxCONTROL_ITEM_FIRST | wxCONTROL_ITEM_LAST`

>     * a rounded rect top line

This would be `wxCONTROL_ITEM_FIRST`

>     * a rounded rect bottom line

This would be `wxCONTROL_ITEM_LAST`

src/generic/listctrl.cpp
paulcor: > macOS converts to full row and square selection boxes also on the side that is not scrolled off screen

Really? So if you resize a list control window horizontally so that a scrollbar is needed, the rounding will disappear from the **left** side as well? That seems... weird. And do the divider lines also disappear?

src/generic/listctrl.cpp
RobertRoeb: > wxCONTROL_ITEM_FIRST | wxCONTROL_ITEM_LAST

If that is what we want, OK. My problem is that for a very common case of a single selection control, we would need to set these two flags, because macOS uses round selection then as well, see e.g. the system settings, which let you select only one. Not entirely sure if that is the same control , but I think so. 

I think I prefer a distinct flag like it is the PR now. Who knows, maybe Apple changes the radius for this case compared to the "single line, but it could be a group" case. 

Tahoe single sel system settings

src/generic/listctrl.cpp
RobertRoeb: I checked earlier today and they stay

RobertRoeb: Here is a screen capture.

Rect vs Round

RobertRoeb:
Any further comments on this PR, or the naming of the flags?

src/generic/listctrl.cpp
```diff
@@ -1961,6 +2017,11 @@ void wxListMainWindow::RefreshAfter( size_t lineFrom )
 
         if ( lineFrom < visibleFrom )
             lineFrom = visibleFrom;
+#ifdef __WXOSX__
+        // Deleting a line can change the look of the selection of the
```
vadz, line 2007: It seems that it could also modify the look of other lines by the same logic as in `RefreshLines()`, e.g. when deleting an unselected line in the middle between selected ones, so shouldn't we refresh the entire window here too?

I think we could probably always refresh everything in the ports using double buffering anyhow, I don't see what does selective refreshing bring us there...

src/generic/treectlg.cpp
```diff
@@ -2562,12 +2562,20 @@ void wxGenericTreeCtrl::PaintItem(wxGenericTreeItem *item, wxDC& dc)
 
     int total_h = GetLineHeight(item);
     bool drawItemBackground = false,
-         hasBgColour = false;
+         hasBgColour = false,
+         drawRounded = false;
```
vadz: Could we please rename this to `drawGroupSelection`?

src/generic/treectlg.cpp
vadz: I still think this part should be done in the renderer. It does get the window, as argument, and so wxOSX renderer could perfectly well compare its client and virtual width (or check for the presence of the horizontal scrollbar, maybe?) and decide to ignore this flag if necessary.

src/osx/cocoa/renderer.mm
```diff
@@ -601,9 +601,60 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
...
+    }
+    else if ((flags & wxCONTROL_ITEM_LAST) != 0)
+    {
+        wxRect toprect = rect;
+        toprect.height = radius;
+        dc.DrawRectangle( toprect );
+        dc.DrawRoundedRectangle( rect, radius );
+        if (rect.width > distanceFromBorder*2)
+        {
+            // wxColour::ChangeLightness( 125 ) actually adds grey to the colour.
```
vadz, line 647: I don't understand this comment :-( It seems to say that this doesn't work, but then still does it for the focused item.

src/generic/listctrl.cpp
vadz: All I can say is, once again, that platform-specific checks like this should be done in the renderer. It should ignore the `SELECTION_GROUP` flag if it doesn't apply, generic code must not have to do this.

include/wx/renderer.h
```diff
@@ -74,7 +74,16 @@ enum
...
     // controls internally
-    wxCONTROL_DIRTY      = 0x80000000
+    wxCONTROL_DIRTY      = 0x80000000,
+
+    // used to indicate a round selection on e.g. macOS
+    wxCONTROL_SELECTION_GROUP  = 0x10000000,
+
+    // used to indicate a sub item within e.g. selections or header controls parts
+    wxCONTROL_ITEM_FIRST       = 0x20000000,
+    wxCONTROL_ITEM_MIDDLE      = 0x40000000,
```
vadz: I think the suggestion to drop this one and just use the absence of `FIRST` or `LAST` as indication of middle had merit. Why do we need all of them?

include/wx/renderer.h
```diff
@@ -74,7 +74,16 @@ enum
...
+    wxCONTROL_ITEM_LAST        = wxCONTROL_DIRTY,
```
vadz, line 84: I think we can break compatibility with the never documented use of `DIRTY` to mean last and assign a separate value for `LAST` if we want to. But this is minor, please do (or do not) whatever you prefer.

include/wx/renderer.h
```diff
@@ -74,7 +74,16 @@ enum
...
+    wxCONTROL_ITEM_MASK        = wxCONTROL_ITEM_FIRST | wxCONTROL_ITEM_MIDDLE | wxCONTROL_ITEM_LAST
```
vadz: This doesn't seem to be used anywhere, do we really need it?

include/wx/renderer.h
```diff
@@ -74,7 +74,16 @@ enum
 
     // this is a pseudo flag not used directly by wxRenderer but rather by some
     // controls internally
-    wxCONTROL_DIRTY      = 0x80000000
+    wxCONTROL_DIRTY      = 0x80000000,
+
+    // used to indicate a round selection on e.g. macOS
+    wxCONTROL_SELECTION_GROUP  = 0x10000000,
```
vadz: This (and the other flags below) should be added to `DrawItemSelectionRect()` documentation in `interface/wx/renderer.h`.

vadz (changes requested):
Thanks for the improvements, but I still have the idea of checking for horizontal scrolling in the caller, this totally breaks separation of concerns between the control and the renderer. Please change this or explain why do you think it's right to do it like this and wrong it in the renderer because I just don't see it.

src/osx/cocoa/renderer.mm
RobertRoeb: I believe that wxColour::ChangeLightness() is wrong, but I would need to tackle that separately. In this case, it should create light blue, but it seems to create a mix of light blue and grey. Again, this is for another PR and will probably require a new function for wxColour()

include/wx/renderer.h
RobertRoeb: It is not used right now, but I added because I initially thought I would need it and it usual across wxWidgets. I can remove.

include/wx/renderer.h
RobertRoeb: Yes, once we agree on them

include/wx/renderer.h
RobertRoeb: Ok

RobertRoeb:
Next version is in.

src/generic/listctrl.cpp
vadz: Is the above wrong?

include/wx/renderer.h
```diff
@@ -74,7 +74,15 @@ enum
...
-    wxCONTROL_DIRTY      = 0x80000000
+    wxCONTROL_DIRTY      = 0x80000000,
+
+    // used to indicate a round selection on e.g. macOS
+    wxCONTROL_SELECTION_GROUP  = 0x20000000,
+
+    // used to indicate a sub item within e.g. selections or header controls parts
+    wxCONTROL_ITEM_FIRST       = 0x40000000,
+    wxCONTROL_ITEM_LAST        = wxCONTROL_DIRTY,
+    // wxCONTROL_ITEM_MIDDLE   = !wxCONTROL_ITEM_FIRST && !wxCONTROL_ITEM_LAST,
```
vadz, line 85: This doesn't seem to correspond to how these flags are actually used.

include/wx/renderer.h
```diff
@@ -74,7 +74,15 @@ enum
 
     // this is a pseudo flag not used directly by wxRenderer but rather by some
     // controls internally
-    wxCONTROL_DIRTY      = 0x80000000
+    wxCONTROL_DIRTY      = 0x80000000,
+
+    // used to indicate a round selection on e.g. macOS
+    wxCONTROL_SELECTION_GROUP  = 0x20000000,
+
+    // used to indicate a sub item within e.g. selections or header controls parts
```
vadz, line 82: If we use these flags as initially suggested, I'd describe them like this:
```suggestion
    // used to draw either first or last item specially (if neither flag is
    // specified, item is drawn as a generic item in the middle)
    //
    // may be combined with wxCONTROL_SELECTION_GROUP
```

interface/wx/renderer.h
```diff
@@ -62,7 +62,14 @@ enum
     wxCONTROL_CHECKABLE  = 0x00000080,
 
     /** (Check) undetermined state. */
-    wxCONTROL_UNDETERMINED = wxCONTROL_CHECKABLE
+    wxCONTROL_UNDETERMINED = wxCONTROL_CHECKABLE,
+
+    /** Indicated that the control might display a group selection across several lines */
```
vadz, line 67:
```suggestion
    /**
        Use to allow the control to display selection across multiple items.
        
        @since 3.3.2
    */
```

interface/wx/renderer.h
```diff
@@ -62,7 +62,14 @@ enum
...
+    wxCONTROL_SELECTION_GROUP  = 0x20000000,
+
+    /** For group selections indicates the position in the group */
```
vadz, line 70:
```suggestion
    /**
        Use for the first item, which may be drawn specially.
        
        Currently only used in combination with ::wxCONTROL_SELECTION_GROUP.
        
        If neither this flag nor ::wxCONTROL_ITEM_LAST is used, the item
        is considered to be neither first nor last, i.e. in the middle.

        @since 3.3.2
     */
```

interface/wx/renderer.h
```diff
@@ -62,7 +62,14 @@ enum
...
+    wxCONTROL_ITEM_FIRST       = 0x40000000,
+    wxCONTROL_ITEM_LAST        = 0x80000000,
```
vadz, line 72:
```suggestion
    /**
        Use for the last item, which may be drawn specially.
        
        @see wxCONTROL_ITEM_FIRST
        
        @since 3.3.2
    */
    wxCONTROL_ITEM_LAST        = 0x80000000,
```

src/generic/listctrl.cpp
```diff
@@ -690,6 +691,28 @@ void wxListLineData::ApplyAttributes(wxDC *dc,
...
+        wxListMainWindow *mainWindow = listctrl->m_mainWin;
+
+        if (listctrl->HasFlag(wxLC_SINGLE_SEL) || (listctrl->GetItemCount() < 2) ||
+            !listctrl->HasFlag(wxLC_REPORT) )
+        {
+            flags |= wxCONTROL_SELECTION_GROUP | wxCONTROL_ITEM_FIRST | wxCONTROL_ITEM_LAST;
+        }
+        else
+        {
+            if (mainWindow->GetClientSize().x >= mainWindow->GetVirtualSize().x)
```
vadz, line 703: Can we please get rid of this?

src/osx/cocoa/renderer.mm
```diff
@@ -601,9 +601,68 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
...
+    {
+        dc.DrawRectangle( rect );
+        return;
+    }
+
+    if (((flags & wxCONTROL_ITEM_FIRST) != 0) && ((flags & wxCONTROL_ITEM_LAST) != 0))
+    {
+        dc.DrawRoundedRectangle( rect, radius );
+    }
+    else if ((flags & wxCONTROL_ITEM_FIRST) != 0)
```
vadz, line 631: I don't know who benefits from all these parenthesis, I find this much more readable:
```suggestion
    else if (flags & wxCONTROL_ITEM_FIRST)
```

src/osx/cocoa/renderer.mm
```diff
@@ -601,9 +601,68 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
...
+    {
+        dc.DrawRoundedRectangle( rect, radius );
+        wxRect bottomrect = rect;
+        bottomrect.y = bottomrect.y + bottomrect.height-radius;
+        bottomrect.height = radius;
+        dc.DrawRectangle( bottomrect );
+    }
+    else if ((flags & wxCONTROL_ITEM_LAST) != 0)
```
vadz, line 639: And
```suggestion
    else if (flags & wxCONTROL_ITEM_LAST)
```

src/osx/cocoa/renderer.mm
```diff
@@ -601,9 +601,68 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
...
+    {
+        wxRect toprect = rect;
+        toprect.height = radius;
+        dc.DrawRectangle( toprect );
+        dc.DrawRoundedRectangle( rect, radius );
+        if (rect.width > distanceFromBorder*2)
+        {
+            // wxColour::ChangeLightness( 125 ) actually adds grey to the colour.
+            // It should be just lighter.
```
vadz, line 648:
```suggestion
            // wxColour::ChangeLightness() is not quite right, as it adds grey to
            // the colour, but will do for now.
```

src/osx/cocoa/renderer.mm
```diff
@@ -601,9 +601,68 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
...
+            wxColour lineColour = col.ChangeLightness( ((flags & wxCONTROL_FOCUSED) != 0) ? 125 : 95 );
```
vadz, line 649:
```suggestion
            wxColour lineColour = col.ChangeLightness(flags & wxCONTROL_FOCUSED ? 125 : 95);
```

src/osx/cocoa/renderer.mm
```diff
@@ -601,9 +601,68 @@ wxRect rect( (int) splitterRect.origin.x, (int) splitterRect.origin.y, (int) spl
...
+    {
+        dc.DrawRectangle( rect );
+        if (rect.width > distanceFromBorder*2)
+        {
+            // wxColour::ChangeLightness( 125 ) actually adds grey to the colour.
+            // It should be just lighter.
+            wxColour lineColour = col.ChangeLightness( ((flags & wxCONTROL_FOCUSED) != 0) ? 125 : 95 );
+            dc.SetPen( wxPen( lineColour ));
+            dc.DrawLine( rect.x + distanceFromBorder, rect.y, rect.x + rect.width - distanceFromBorder, rect.y );
+        }
```
vadz, line 664: This seems to be identical to the code above, can't it be reused by moving it below the `if`? Or am I missing something?

vadz (changes requested):
Sorry, I still don't understand the meaning of the flags here. You seem to be using `wxCONTROL_ITEM_FIRST | wxCONTROL_ITEM_LAST` as meaning `MIDDLE`? If I got this correctly, why?? It seems much more logical to use absence of either flag (as originally suggested) to mean `MIDDLE`...

RobertRoeb:
As we cannot agree on the meaning of the flags, I close this PR-

vadz:
> As we cannot agree on the meaning of the flags, I close this PR-

The latest version of the PR is inconsistent with _itself_, it has a comment saying that `MIDDLE = !FIRST && !LAST` but then actually uses `MIDDLE = FIRST || LAST` which is exactly the negation. And while the former seems much more natural to me ("an item which is not first nor last is in the middle"), I could live with the latter (especially if you could explain why is it better), but it does need to be consistent.

PR #26085, opened 2026-01-09
Disallow wxSplitterWindow to get the focus

RobertRoeb:
Overrides AcceptsFocus() so that the child windows keeps the focus during a resize operation


vadz:
Thanks, I'll apply this even though I am not entirely sure if this is not going to result in any problems under the other platforms (it seems pretty unusual to have mouse capture but not focus...), but I guess we'll only find out once we merge it.

PR #26086, opened 2026-01-09
Add debug logging to `wxWebRequest`

vadz:
This is a draft as there is no documentation yet and I'd also like to log at least something (even though I don't think we can information at the same level of details as libcurl provides) for the WinHTTP backend.

I'd welcome comments about the API, notably I don't know if it really makes sense to allocate a new logger for each request, but if we want to reuse the same one, we probably need to pass the request to it?

### DIFF
--- a/include/wx/msw/private/webrequest_winhttp.h
+++ b/include/wx/msw/private/webrequest_winhttp.h
@@ -170,6 +170,10 @@ private:
         return Fail(operation, ::GetLastError());
     }
 
+    // Log all the headers of the response if debug logging is enabled.
+    void LogResponseHeadersIfNecessary();
+
+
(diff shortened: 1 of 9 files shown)


vadz:
Thinking more about this, would it, perhaps, make more sense to enable debug logging at the session level? This is less flexible, but, in practice, I think I'm going to enable it for all requests anyhow when I need more information for troubleshooting.

vadz:
I've moved setting the logger to the session as this seems more convenient and provided partial implementation for WinHTTP.

I plan to merge this if there are no objections, as this should be already useful when debugging any download-related problems.

vadz:
No comments, so merging.

vadz:
Applied in the commit above.

PR #26088, opened 2026-01-10
OS X, correct background of wxStaticText in a wxStaticBox

RobertRoeb:
background
background3
  - Manually set background if static text is in static box
  - Problem is that static text is not transparent in its current implementation on OS X
  - Same problem applies to wxCheckBox, wxRadioButton and all items of wxRadioBox, possibly more
  - Better fix would be to find a way to make these labels truely transparent
  - NSTextField labelWithAttributedString might do this (?)
  - This is to start a debate - the patch works in dark and light mode

The attached first screenshot shows the wrong background colour, in this case in a wxRadioBox in dark mode. The second screenshot shows the corrected background colour of a wxStatictext in a wxStaticBox, but the uncorrected wrong colour of a wxCheckBox in the same wxStaticBox.

It also shows the macOS system settings dialog of macOS Tahoe 26.2 on the left, which clearly uses a different colour theme altogether, so the wxWidgets app uses some legacy rendering, which eventually we should get rid of (maybe due to overriding drawRect as has been suggested in a different patch).

### DIFF
--- a/src/osx/stattext_osx.cpp
+++ b/src/osx/stattext_osx.cpp
@@ -19,11 +19,16 @@
     #include "wx/dc.h"
     #include "wx/dcclient.h"
     #include "wx/settings.h"
+    #include "wx/statbox.h"
 #endif // WX_PRECOMP
 
 #include "wx/osx/private.h"
 
 #include <stdio.h>
+
+wxBEGIN_EVENT_TABLE(wxStaticText, wxStaticTextBase)
+    EVT_SYS_COLOUR_CHANGED( wxStaticText::OnSysColourChanged )
+wxEND_EVENT_TABLE()
 
 
 bool wxStaticText::Create( wxWindow *parent,
@@ -52,7 +57,31 @@ bool wxStaticText::Create( wxWindow *parent,
         SetInitialSize(size);
     }
 
+    DoUpdateBackground();
+
     return true;
+}
+
+void wxStaticText::DoUpdateBackground()
+{
+#if wxOSX_USE_COCOA_OR_CARBON
+    if ((GetParent() != nullptr) && GetParent()->IsKindOf(wxCLASSINFO(wxStaticBox)))
+    {
+        if (wxSystemSettings::GetAppearance().IsDark())
+        {
+            SetBackgroundColour( wxColour( 37, 37, 37 ) );
+        }
+        else
+        {
+            SetBackgroundColour( wxColour( 247, 247, 247 ) );
+        }
+    }
+#endif
+}
+
+void wxStaticText::OnSysColourChanged(wxSysColourChangedEvent& WXUNUSED(event))
+{
+    DoUpdateBackground();
 }
 
 void wxStaticText::SetLabel(const wxString& label)
--- a/include/wx/osx/stattext.h
+++ b/include/wx/osx/stattext.h
@@ -50,6 +50,12 @@ protected :
 #endif // wxUSE_MARKUP && wxOSX_USE_COCOA
 
     wxDECLARE_DYNAMIC_CLASS_NO_COPY(wxStaticText);
+    
+    void DoUpdateBackground();
+    void OnSysColourChanged(wxSysColourChangedEvent& event);
+
+private:
+    wxDECLARE_EVENT_TABLE();
 };
 
 #endif


src/osx/stattext_osx.cpp
```diff
@@ -19,12 +19,17 @@
...
 #endif // WX_PRECOMP
 
 #include "wx/osx/private.h"
 
 #include <stdio.h>
 
+wxBEGIN_EVENT_TABLE(wxStaticText, wxStaticTextBase)
+    EVT_SYS_COLOUR_CHANGED( wxStaticText::OnSysColourChanged )
+wxEND_EVENT_TABLE()
+
```
MaartenBent, line 32: Hi, I'm not sure what the best practices are here, but I think new code always tries to use `Bind`/`Unbind` instead of event tables.

src/osx/stattext_osx.cpp
```diff
@@ -52,9 +57,33 @@ bool wxStaticText::Create( wxWindow *parent,
...
+#if wxOSX_USE_COCOA_OR_CARBON
+    if ((GetParent() != nullptr) && GetParent()->IsKindOf(wxCLASSINFO(wxStaticBox)))
+    {
+        if (wxSystemSettings::GetAppearance().IsDark())
+        {
+            SetBackgroundColour( wxColour( 37, 37, 37 ) );
+        }
+        else
+        {
+            SetBackgroundColour( wxColour( 247, 247, 247 ) );
```
MaartenBent, line 76: Are there maybe `wxSystemColour` constants that map to these colours?

src/osx/stattext_osx.cpp
```diff
@@ -52,9 +57,33 @@ bool wxStaticText::Create( wxWindow *parent,
...
 
+    DoUpdateBackground();
+
     return true;
 }
 
+void wxStaticText::DoUpdateBackground()
+{
+#if wxOSX_USE_COCOA_OR_CARBON
+    if ((GetParent() != nullptr) && GetParent()->IsKindOf(wxCLASSINFO(wxStaticBox)))
```
vadz, line 68: Parent can never be null for a non-TLW, so the first half of the test is not necessary.

Second part of the test would, of course, be better written as a call to some virtual `OSXGetChildBackground()` or similar.

src/osx/stattext_osx.cpp
```diff
@@ -52,9 +57,33 @@ bool wxStaticText::Create( wxWindow *parent,
...
+        }
+        else
+        {
+            SetBackgroundColour( wxColour( 247, 247, 247 ) );
+        }
+    }
+#endif
+}
+
+void wxStaticText::OnSysColourChanged(wxSysColourChangedEvent& WXUNUSED(event))
```
vadz, line 82: Should call `event.Skip()`, especially if/when this is converted to use `Bind()`.

vadz (changes requested):
I won't merge this without approval from @csomor as I'm far from sure that this is correct — it would be far better to handle all these controls (i.e. those with `HasTransparentBackground()` returning `true`) in some generic way somewhere else.

Also, as written, this overwrites any custom background set for `wxStaticText` which should, of course, be avoided.

csomor:
these controls should have their background style be `wxBG_STYLE_TRANSPARENT` but right now it's `wxBG_STYLE_ERASE`, adding a  
`SetBackgroundStyle(wxBG_STYLE_TRANSPARENT);`
 in the constructor shows the fix, but I don't think they always had `wxBG_STYLE_ERASE` as style in the first place, I'll try to find out

vadz:
Closing in favour of #26090, please test if it fixes the problem.

HenkDemper:
By coincident was fixing background of controls in wxStaticBox last week for both wxStaticText and wxSlider (both macOS and Windows).

After the fix Windows is completely Ok, but for wxSlider on macOS 26.2 still need `pSlider->SetBackgroundColour(wxColour(0,0,0,wxALPHA_TRANSPARENT))` to make it correct.

Sliders

@csomor I assume that this is expected because macOS sliders are native ? Or would the fix need to be extended to more controls ?

csomor:
> @csomor I assume that this is expected because macOS sliders are native ? Or would the fix need to be extended to more controls ?

#26092 is the PR collecting all the controls that need their transparency set to true, could you try doing the same to the wxSlider

PR #26090, opened 2026-01-10
OSX transparent controls background

csomor:
macOS has transparent backgrounds for controls like statictext, or the text part of checkboxes etc, also nested group boxes get different shades of gray, avoids background erase, see wxWidgets#26088

### DIFF
--- a/src/osx/checkbox_osx.cpp
+++ b/src/osx/checkbox_osx.cpp
@@ -35,8 +35,6 @@ bool wxCheckBox::Create(wxWindow *parent,
 
     WXValidateStyle( &style );
 
-    SetBackgroundStyle(wxBG_STYLE_TRANSPARENT);
-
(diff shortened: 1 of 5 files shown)


vadz:
Should I merge this and close #26088 or do you plan more changes here?

Also, should this be backported to 3.2?

csomor:
@vadz it should replace it yes, but first we need confirmation from @RobertRoeb that it solves his problems as well, and if we get enough coverage on master, then backporting it would be a good idea

RobertRoeb:
Thanks, it works! 
BTW, I had tried setting that BG style myself, and even just tried again, but in a slightly different stage of control creation process, and it had/has no effect.

csomor:
> Thanks, it works! BTW, I had tried setting that BG style myself, and even just tried again, but in a slightly different stage of control creation process, and it had/has no effect.

perfect, thanks a lot for testing

PR #26091, opened 2026-01-10
CMake: Search for GTK2 using pkg-config first

MaartenBent:
Add a copy of the FindGTK3 module, but modified for GTK2. If it fails, use the CMake's default FindGTK2.

Fixes #26040


vadz:
Thanks, applied as 312645b9057afc8750e3f839f9ebf1f7b15d1b01 (which forgot to use `#` before the PR number).

PR #26092, opened 2026-01-11
OSX add explicit transparency where missing

csomor:
some controls were missing the info, and #26090 does not get traction

### DIFF
--- a/include/wx/osx/radiobut.h
+++ b/include/wx/osx/radiobut.h
@@ -37,6 +37,8 @@ public:
     virtual void SetValue(bool val) override;
     virtual bool GetValue() const override;
 
+    virtual bool HasTransparentBackground() override { return true; }
+
     // implementation
 
     void Command(wxCommandEvent& event) override;
--- a/include/wx/osx/radiobox.h
+++ b/include/wx/osx/radiobox.h
@@ -73,6 +73,8 @@ public:
     virtual wxString GetString(unsigned int item) const override;
     virtual void SetString(unsigned int item, const wxString& label) override;
 
+    virtual bool HasTransparentBackground() override { return true; }
+
     // protect native font of box
     virtual bool SetFont( const wxFont &font ) override;
 // Other external functions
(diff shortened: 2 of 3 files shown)


include/wx/osx/radiobut.h
```diff
@@ -37,6 +37,8 @@ class WXDLLIMPEXP_CORE wxRadioButton: public wxRadioButtonBase
     virtual void SetValue(bool val) override;
     virtual bool GetValue() const override;
 
+    virtual bool HasTransparentBackground() override { return true; }
```
vadz, line 40: I think we should put this in `wxRadioButtonBase`, hopefully this shouldn't break anything in the other ports.

csomor: > I think we should put this in `wxRadioButtonBase`, hopefully this shouldn't break anything in the other ports.

@vadz that's what I didn't dare to do ...

RobertRoeb: Just tested on macOS, and the override in wxRadioBox is not required given that it is composed of wxRadioButtons anyways. It is probably still correct to do this. BTW, I am not quite sure what the difference is between HasTransparentBackground() and wxBG_STYLE_TRANSPARENT.

HenkDemper:
From #26090:

> @csomor: https://github.com/wxWidgets/wxWidgets/pull/26092 is the PR collecting all the controls that need their transparency set to true, could you try doing the same to the wxSlider

That indeed works: if I add in `slider.h` for `wxSlider`:
`virtual bool HasTransparentBackground() override { return true; }`
& rebuild libs then it is solved for sliders as well and `SetBackgroundColour()` is not needed anymore 👍🏻

vadz:
Stefan, should this be merged too already or do you plan to push more changes here?

csomor:
> Stefan, should this be merged too already or do you plan to push more changes here?

Yes, thanks, I think we merge it, and will do another PR for the controls that come up later

PR #26094, opened 2026-01-12
Enable macOS 26 Tahoe background

RobertRoeb:
- Don't override drawRect (and as such disable user code drawing in all native controls) to avoid legacy rendering
  - Catch binding to wxEVT_PAINT in static tables and dynamically and add overriding of drawRect only then
  - Disable wxSIMPLE_BORDER and NO_BORDER in wxTextCtrl to avoid legacy rendering
  - Multi-line wxTextCtrl still uses brown background
  - Adapt system colour wxCOLOUR_LISTBOX to report the correct opaque colours in light and dark mode under Tahoe

### DIFF
--- a/include/wx/defs.h
+++ b/include/wx/defs.h
@@ -1446,7 +1446,7 @@ wxALLOW_COMBINING_ENUMS(wxSizerFlagBits, wxStretch)
 #define wxCLIP_SIBLINGS         0x20000000
 
 /* Obsolete */
-#define wxTRANSPARENT_WINDOW    0x00000000
+#define wxTRANSPARENT_WINDOW    0x00100000
 
 /*  Add this style to a panel to get tab traversal working outside of dialogs */
 /*  (on by default for wxPanel, wxDialog, wxScrolledWindow) */
--- a/include/wx/defs.h
+++ b/include/wx/defs.h
@@ -1445,8 +1445,8 @@ wxALLOW_COMBINING_ENUMS(wxSizerFlagBits, wxStretch)
 /*  for subwindows/controls */
 #define wxCLIP_SIBLINGS         0x20000000
 
-/* This style is obsolete and doesn't do anything. */
-#define wxTRANSPARENT_WINDOW    0
+/* This style is re-used for semi-transparent backgrounds. */
+#define wxTRANSPARENT_WINDOW    0x00100000
 
 /*  Add this style to a panel to get tab traversal working outside of dialogs */
 /*  (on by default for wxPanel, wxDialog, wxScrolledWindow) */
--- a/include/wx/defs.h
+++ b/include/wx/defs.h
@@ -1445,8 +1445,8 @@ wxALLOW_COMBINING_ENUMS(wxSizerFlagBits, wxStretch)
 /*  for subwindows/controls */
 #define wxCLIP_SIBLINGS         0x20000000
 
-/* This style is re-used for semi-transparent backgrounds. */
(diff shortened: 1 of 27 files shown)


RobertRoeb:
Further work, see below. I still do not know how to query the new background colours, of which there at least two, one in the background of a e.g. wxListBox and one as the background of e.g. wxPanel and wxDialog. The colour of wxRadioBox and wxStaticBox background look good now. But even better would be if the windows, also those that do override doDraw, would get the right background automatically (and the right background on older systems as well, of course)

Tahoe Light
Tahoe Dark

RobertRoeb:
It seems that macOS has no background colour(s) anymore, it has a background that consists of NSVisualEffectView using as parameters:
1) material (basically the overall style or template) : the NSVisualEffectMaterialWindowBackground for an opaque background, NSVisualEffectMaterialSidebar would eg. be for the left part of a wxTreeBook.
2) blending mode: NSVisualEffectBlendingModeWithinWindow or NSVisualEffectBlendingModeBehindWindow (See below for explanation screenshot, here "Window" seem to be what we call wxTopLevelWindow )
3) vibrancy: as if material and blending were not enough, you can add vibrancy by overriding [BOOL] allowsVibrancy; to return true

As mentioned before, such windows must not override [drawRect:] or [updateLayer]

Looking at Windows 11, I speculate that MSW is also getting some degree of transparency, shine through features. 

Thinking about and API, we currently already have 
   SetBackgroundStyle( wxBG_STYLE_TRANSPARENT );
which has to be called before Create() - so maybe it should rather be a window style flag - and if we use that in wxPanel, wxDialog or even wxWindow, then we'd need to insert a NSVisualEffectView to get these backgrounds. When this flag is used, no wxPaintEvent would be sent as drawRect cannot be overridden.

Here are the material to choose from:
https://developer.apple.com/documentation/appkit/nsvisualeffectview/material-swift.enum?language=objc

For the interested reader, here is a SO conversation on the topic:

https://stackoverflow.com/questions/67304592/how-to-reliably-retrieve-a-windows-background-color-on-macos-with-swiftui

image

RobertRoeb:
I have updated the PR to check if the windows intercepts wxEVT_PAINT events either in the static event tables or using the dynamic event system (which required making wxEvtHandler::doBind() virtual). If a paint event handler is present, the background on macOS 26 Tahoe will fall back the legacy rendering of previous macOS versions, but otherwise it will display the native background.
This hopefully also allow us to use the complex backgrounds show above in the future.

RobertRoeb:
That is what you can do with it now:


https://github.com/user-attachments/assets/4278fe2a-ced6-4601-bd51-e2b225779b1a

MaartenBent:
That looks nice!

I didn't really follow the mailing list about this, but does it work by adding the `wxTRANSPARENT_WINDOW` flag? To just the `wxPanel` or all controls in it?

If it is not a default style, you'll have to add many default parameters to the constructor, like you did in `notebook.cpp` in the diff (`new wxPanel(parent, -1, wxDefaultPosition, wxDefaultSize, wxTRANSPARENT_WINDOW);`).
Is it possible to enable it by calling a function/`SetWindowStyleFlag` after creating the panel?
 I'm not sure how it works, but maybe you can always create the `visualEffectView` and toggle its state?

RobertRoeb:
Just the flag in wxPanel. I don't know yet how we would pass parameters. I think you can create the effect layer once and alter it later

PR #26097, opened 2026-01-13
Corrected wxTreeCtrl::DoGetBestSize(), added DoGetBestClientWidth

RobertRoeb:
For the generic wxTreeCtrl:
- Corrected ::DoGetBestSize(), added DoGetBestClientWidth
 - Adapted both to the same calculation of the required virtual size that is used in AdjustScrollbars() removing the need for adding magic 4
 - Now, importantly, wxTreeBook actually calls a real DoGetBestClientWidth() based on a known height and can correctly figure out if a vertical scrollbar is needed
  - This removes the ugly horizontal scrollbar in wxTreeBook when a the horizontal scrollbar appears.
  - Tested on macOS 26

### DIFF
--- a/src/generic/treectlg.cpp
+++ b/src/generic/treectlg.cpp
@@ -2491,18 +2491,26 @@ void wxGenericTreeCtrl::AssignButtonsImageList(wxImageList *imageList)
 // helpers
 // -----------------------------------------------------------------------------
 
+wxSize GetTotalAreaFromClientArea(wxSize size)
+{
+    int x = size.x + PIXELS_PER_UNIT+2; // one more scrollbar unit + 2 pixels
+    int y = size.y + PIXELS_PER_UNIT+2; // one more scrollbar unit + 2 pixels
+    return wxSize( x, y);
+}
+
 void wxGenericTreeCtrl::AdjustMyScrollbars()
 {
     if (m_anchor)
     {
         int x = 0, y = 0;
         m_anchor->GetSize( x, y, this );
-        y += PIXELS_PER_UNIT+2; // one more scrollbar unit + 2 pixels
-        x += PIXELS_PER_UNIT+2; // one more scrollbar unit + 2 pixels
+        wxSize totalSize = GetTotalAreaFromClientArea( wxSize( x, y) );
         int x_pos = GetScrollPos( wxHORIZONTAL );
         int y_pos = GetScrollPos( wxVERTICAL );
-        SetScrollbars( PIXELS_PER_UNIT, PIXELS_PER_UNIT,
-                       x/PIXELS_PER_UNIT, y/PIXELS_PER_UNIT,
+        SetScrollbars( PIXELS_PER_UNIT, 
+                       PIXELS_PER_UNIT,
+                       totalSize.x/PIXELS_PER_UNIT, 
+                       totalSize.y/PIXELS_PER_UNIT,
                        x_pos, y_pos );
     }
     else
@@ -4189,7 +4197,7 @@ void wxGenericTreeCtrl::DoDirtyProcessing()
     AdjustMyScrollbars();
 }
 
-wxSize wxGenericTreeCtrl::DoGetBestSize() const
+wxSize wxGenericTreeCtrl::DoGetBestClientSize() const
 {
     // make sure all positions are calculated as normally this only done during
     // idle time but we need them for base class DoGetBestSize() to return the
@@ -4202,10 +4210,16 @@ wxSize wxGenericTreeCtrl::DoGetBestSize() const
     // will not have scrollbars
 
     // Use the calculation from AdjustMyScrollbars()
-    size.y += PIXELS_PER_UNIT+2; // one more scrollbar unit + 2 pixels
-    size.x += PIXELS_PER_UNIT+2; // one more scrollbar unit + 2 pixels
+    size = GetTotalAreaFromClientArea( size );
     size.x = (size.x / PIXELS_PER_UNIT) * PIXELS_PER_UNIT;
     size.y = (size.y / PIXELS_PER_UNIT) * PIXELS_PER_UNIT;
+
+    return size;
+}
+
+wxSize wxGenericTreeCtrl::DoGetBestSize() const
+{
+    wxSize size = DoGetBestClientSize();
 
     // add the border
     const wxSize& borderSize = GetWindowBorderSize();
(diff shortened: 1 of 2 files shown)


RobertRoeb:
This image illustrates the problem: when all branches are expanded, a horizontal scrollbar appears because the current code miscalculated the width. This looks broken and you cannot even add some manual extra space.

Treebook Tahoe

src/generic/treectlg.cpp
```diff
@@ -4198,23 +4198,49 @@ wxSize wxGenericTreeCtrl::DoGetBestSize() const
...
-    int dx = (size.x - borderSize.x) % PIXELS_PER_UNIT;
-    if ( dx )
-        size.x += PIXELS_PER_UNIT - dx;
-    int dy = (size.y - borderSize.y) % PIXELS_PER_UNIT;
-    if ( dy )
-        size.y += PIXELS_PER_UNIT - dy;
+    // add the border
+    const wxSize& borderSize = GetWindowBorderSize();
+    size.x += borderSize.x;
+    size.y += borderSize.y;
```
vadz, line 4227: Could we avoid this by overriding `DoGetBestClientSize()` instead of `DoGetBestSize()`?

And if not, this should be written simply as

```cpp
size += GetWindowBorderSize();
```

src/generic/treectlg.cpp
```diff
@@ -4198,23 +4198,49 @@ wxSize wxGenericTreeCtrl::DoGetBestSize() const
...
+    wxSize size = wxTreeCtrlBase::DoGetBestSize();
+
+    // DoGetBestClientWidth calculates if a vertical scrollbar is
+    // needed based on the given height
+
+    // Use the calculation from AdjustMyScrollbars()
+    size.y += PIXELS_PER_UNIT+2; // one more scrollbar unit + 2 pixels
+    size.x += PIXELS_PER_UNIT+2; // one more scrollbar unit + 2 pixels
+    size.x = (size.x / PIXELS_PER_UNIT) * PIXELS_PER_UNIT;
+    size.y = (size.y / PIXELS_PER_UNIT) * PIXELS_PER_UNIT;
```
vadz: This is just an exact copy of `DoGetBestClientSize()`, right? Why not simply call it from here?

src/generic/treectlg.cpp
```diff
@@ -4198,23 +4198,49 @@ wxSize wxGenericTreeCtrl::DoGetBestSize() const
...
-    // I'm not really sure where does it come from -- but without this, the
-    // scrollbars appear in a tree with default/best size
-    size.IncBy(4, 4);
+    // DoGetBestSize assumes we can stretch out completely and therefore
+    // will not have scrollbars
 
-    // and the border has to be rounded up to a multiple of PIXELS_PER_UNIT or
-    // scrollbars still appear
-    const wxSize& borderSize = GetWindowBorderSize();
+    // Use the calculation from AdjustMyScrollbars()
```
vadz, line 4212: Could we, perhaps, add a function like `wxSize GetTotalAreaFromClientArea(wxSize)` and use it both here and in `AdjustMyScrollbars()`?

src/generic/treectlg.cpp
```diff
@@ -4198,23 +4198,49 @@ wxSize wxGenericTreeCtrl::DoGetBestSize() const
...
+    size.y += PIXELS_PER_UNIT+2; // one more scrollbar unit + 2 pixels
+    size.x += PIXELS_PER_UNIT+2; // one more scrollbar unit + 2 pixels
+    size.x = (size.x / PIXELS_PER_UNIT) * PIXELS_PER_UNIT;
+    size.y = (size.y / PIXELS_PER_UNIT) * PIXELS_PER_UNIT;
```
vadz, line 4215: This rounds up, unlike `AdjustMyScrollbars()` which rounds down. A comment explaining why this is the right thing to do here but not there would be quite helpful because this doesn't seem obvious to me.

vadz (changes requested):
Thanks but, again, it looks like there is an "obvious" (unless I'm missing something, which is perfectly possible, but please explain what to me then) better way of doing this by defining `DoGetBestClientSize()` and reusing it from the new function.

src/generic/treectlg.cpp
RobertRoeb: I think I am using the same calculations, so I must be overlooking something. Please point me to it.

src/generic/treectlg.cpp
RobertRoeb: So maybe for background, this is the sequence of calls that I am trying to fix:
a) wxBookCtrlBase::DoSize() wants to lay out the controller (tree control on the left) and the page (right)
b) it calls wxBookCtrlBase::GetControllerSize() to determine the size of the controller
c) that calls m_bookctrl->GetBestWidth(sizeClient.y); to allow the controller to figure out how much it has to stretch out horizontally
d) wxWindowBase::GetBestWidth(int height) calls DoGetBestClientWidth(height);

Therefore, I had to override wxGenericTreeCtrl::DoGetBestClientWidth(height) to make it properly report the width of the tree depending on whether or not it has a vertical scrollbar. This function needs to know what is essentially the primary direction of expansion, in our case the tree should expand horizontally to not have a horizontal scrollbar and should use scrolling vertically (if it does not fit, obviously).

DoGetBestSize() calculates the size so that (or as if) no scrollbar are present in either direction, that is a different task.

src/generic/treectlg.cpp
vadz: Sorry, it looks like I misread the code, either here or there, because rereading it now I don't see the difference. But this just (completely inadvertently) proves my point about having `GetTotalAreaFromClientArea()`: if we called it from both places, it would be perfectly obvious that the calculation is the same.

src/generic/treectlg.cpp
vadz: I understand that it's not enough to override only `DoGetBestSize()`. The comment above only said that we could override `DoGetBestClientSize()` instead of it to avoid having to add borders manually.

Overridden version of `DoGetBestClientWidth()` would still remain (but hopefully reuse this one instead of duplicating it).

vadz:
Should I make the changes discussed above myself or is it not desirable to make them for some reason I'm missing?

src/generic/treectlg.cpp
RobertRoeb: There was no DoGetBestClientSize() so this question puzzled me for a while. I just added it and made DoGetBestSize() and DoGetClientBestWidth() use it.

src/generic/treectlg.cpp
RobertRoeb: I added this as a local function since it has no general use

src/generic/treectlg.cpp
```diff
@@ -4198,23 +4206,41 @@ wxSize wxGenericTreeCtrl::DoGetBestSize() const
...
-    int dx = (size.x - borderSize.x) % PIXELS_PER_UNIT;
-    if ( dx )
-        size.x += PIXELS_PER_UNIT - dx;
-    int dy = (size.y - borderSize.y) % PIXELS_PER_UNIT;
-    if ( dy )
-        size.y += PIXELS_PER_UNIT - dy;
+    return size;
+}
+
+wxSize wxGenericTreeCtrl::DoGetBestSize() const
```
vadz, line 4220: This is just completely unnecessary unless I'm missing something. `GetBestSize()` doesn't call `DoGetBestSize()` if `DoGetBestClientSize()` is implemented.

src/generic/treectlg.cpp
```diff
@@ -2491,18 +2491,26 @@ void wxGenericTreeCtrl::AssignButtonsImageList(wxImageList *imageList)
 // helpers
 // -----------------------------------------------------------------------------
 
+wxSize GetTotalAreaFromClientArea(wxSize size)
```
vadz, line 2494:
```suggestion
static wxSize GetTotalAreaFromClientArea(wxSize size)
```

vadz (commented):
Thanks, I'll apply this after removing `DoGetBestSize()`, please let me know if I shouldn't.

vadz (commented):
Wait, actually I'm still confused: why do we add size for the scrollbars in `DoGetBestClientSize()`, they are not supposed to be part of the window client size. are they in wxGTK?

RobertRoeb:
I work on macOS now. Maybe there is a difference there

vadz:
I think we need to add a unit test to check that the client size for a scrolled window with scrollbars is as expected (i.e. total size minus border size and scroll bar size, maybe with some tolerance) because I'm not sure if this is really the case from just reading the code but it really should be (or, alternatively, it should be the same as total size minus borders, but then also in all ports).

vadz:
I've created #27043 replacing this one, so closing. Please test the other one.

Issue #26095, opened 2026-01-12
On Mac, wxWidgets applications initially not shown till mouse moved or keyboard pressed


HenkDemper:
- Using latest git master build 8ada351adf on latest macOS Tahoe 26.2
- Build (ARM) for Mac both make & CMake, both Debug & Release (4 sets)
- I experience below with both most sample applications (for instance try auidemo.app) and my own application
- Open sample application like auidemo.app by either keyboard Cmd-O or double-click mouse (but don't move mouse, keep still)
- Application will launch, icon visible in Dock, but no GUI is visible
- Wait short or long time, no GUI visible...
- When keyboard is pressed OR mouse is moved, GUI will appear (and stay)
- Mostly (if not always for at least tested auidemp.app & some others tested, if keep mouse still ?) happens in Debug/Release and with make/CMake builds
- I tried adding extra direct/delayed Show/Refresh commands for MainFrame, but to no avail
 
Problem does not happen (GUI directly visible) for either:
- Windows builds of same application
- When launched same Debug build from Xcode (with Cmd-R)
- Made Universal binary ARM/Intel Make/Release build and run on Intel MBP 2014 with macOS Mojave 10.14.6

Not sure if this is something related to my Mac ARM/Tahoe setup, perhaps others can confirm above works Ok for them ?

csomor:
Thanks Henk, unfortunately reproduced when launching a plain minimal.app built with Xcode 26.2 with Cmd-O from the Finder under macOS 26.3 (a), not happening when launching from Xcode

torstenvl:
Does this also happen if there's a complete .app package with PkgInfo and Info.plist?

Does this also happen if you codesign the binary?

csomor:
Note: the issue comes from the fact that the event loop `[NSApp run] `started in `wxApp::CallOnInit` does not get stopped when `[NSapp stop:nil]` gets called from `applicationDidFinishLaunching`, so I'll try adding another event there to keep the loop running and being able to process stop

csomor:
@HenkDemper could you please test whether modifying `utils.mm` `applicationDidFinishLaunching`

by adding

```
    NSEvent *event = [NSEvent otherEventWithType:NSApplicationDefined
                                location:NSMakePoint(0.0, 0.0)
                           modifierFlags:0
                               timestamp:0
                            windowNumber:0
                                 context:nil
                                 subtype:0 data1:0 data2:0];
    [NSApp postEvent:event atStart:FALSE];
[...]

PR #26098, opened 2026-01-14
OSX fix app boot

csomor:
sometimes an application was not properly brought to front if no other events were happening, see #26095


csomor:
@vadz if I merge this, will the backport label still be visible for you ?

vadz:
Yes, but I can also merge this myself and cherry pick it to 3.2 immediately.

One thing I wonder about: would it be possible to add a comment explaining why is this necessary? Because it doesn't seem very obvious to me :-(

vadz:
Thanks! Will merge this a bit later.

PR #26102, opened 2026-01-16
Deprecate wxXmlInitResourceModule() and do nothing in it

vadz:
This function seems to do more harm than good, so stop doing anything in it and deprecate it as it doesn't seem like calling it could ever be useful.

Closes #26039.

### DIFF
--- a/include/wx/xrc/xmlres.h
+++ b/include/wx/xrc/xmlres.h
@@ -695,8 +695,10 @@ public:
        variable->Hide();
 
 
-// FIXME -- remove this $%^#$%#$@# as soon as Ron checks his changes in!!
-WXDLLIMPEXP_XRC void wxXmlInitResourceModule();
+// This function didn't do anything useful and is now preserved just to avoid
+// breaking compilation of the code using it.
+wxDEPRECATED_MSG("Simply remove the calls to this function from your code")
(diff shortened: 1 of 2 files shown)


vadz:
@swt2c I think wxPython is the only user of this function, please let me know if you see anything wrong with this change.

swt2c:
It looks like wxPython has been calling this function basically for 20 years.  The only reason I can possibly thing of is that the `wx.xml` module can be imported at a later time than the main wxPython module (`wx`) so it may have been related to that?

vadz:
After looking closer into this, I am afraid that this does indeed break wxPython in the case when

```python
import wx.xrc
```

occurs after creating `wx.App()` (which is perfectly legal/possible AFAIK) because `wxXmlResourceModule` is indeed not initialized in this case.

It probably doesn't break it very badly, as the only thing that this module does is to register `wxXmlSubclassFactoryCXX` which is unlikely to be used in Python programs (it's simpler to subclass things using `XmlSubclassFactory_Python` which still gets registered), but in theory this could happen in programs using both wxPython and wx directly, such as KiCad.

So I'm finally closing this one and creating #26111 which doesn't break wxPython and may fix whatever the original problem was.

PR #26104, opened 2026-01-17
CMake: Add CI job to test GTK3 on macOS

MaartenBent:
And prevent building `wxUIActionSimulator` and `wxGLCanvas` when unavailable.
See #26089

@barracuda156 Do you see any problems with this? I think you already disabled `wxUIActionSimulator`. Do you use/build OpenGL/`wxGLCanvas` successfully, I couldn't get them to work due to missing `gdk/gdkx.h`.

Note there is already a similar CI job for configure (#25924) but it does not run automatically.


.github/workflows/ci_cmake.yml
```diff
@@ -79,6 +79,12 @@ jobs:
             cmake_samples: ALL
             cmake_tests: OFF
             cmake_build_toolkit: osx_cocoa
+          - name: macOS latest wxGTK 3 Unix Makefiles
+            runner: macos-latest
+            cmake_generator: Unix Makefiles
```
vadz, line 84: Just curious: any reason not to use Ninja here? I don't see any real advantage in using make with CMake compared to it.

vadz (approved):
Great, thanks a lot!

Should I (squash?)merge this already or do you plan more changes here?

.github/workflows/ci_cmake.yml
MaartenBent: XCode and Ninja were already used for macOS, so I thought lets use Unix Makefiles this time.

MaartenBent:
Yes, you can merge this.

Do you want to squash the last two commits? I did notice the makefile generation between them is not entirely correct.

vadz:
I'll squash them and push, thanks.

Issue #22771, opened 2022-09-03
wxRichTextCtrl Selected Text Shifts


lsri2:
**Describe the bug**
When you are selecting text in the control, you can see other text along the line "shift".  

**Expected vs observed behaviour**
The text shouldn't shift.

**Patch or snippet allowing to reproduce the problem**
Using the wxRichTextCtrl sample, start selecting the line beginning with "What can you do with this thing?".  After you select the W, you will see other text along that line shift.

**Platform and version information**
 - wxWidgets version you use: 3.2.0
 - wxWidgets port you use: wxMSW
 - OS and its version: Windows 7 Pro

PBfordev:
I cannot reproduce this on native Win10 or Win7 under VirtualBox. 

However, when I run the sample and move it from the primary screen at 125% DPI to the secondary one with the standard DPI and attempt to select the sentence, selecting behaves erratically, selecting other parts than those dragged over by mouse. Sometimes it does work though.

catalinr:
IIRC (from what was an old discussion on the same subject which nor I nor google can find) selected text is drawn separately from the rest of text. For partial word selection, it behaves as if there are 2 words next to each other, which renders differently from a single big word. A solution might be to change all that so that the background with the selected color was drawn before/under the text, rather than drawing the selected text separately from the unselected one.
IIRC this required significant changes to text rendering.

torstenvl:
Also experiencing this on wxWidgets 3.3.1 on macOS Ventura. It's not a big deal, but is a little unpolished. 

Video here:
https://github.com/user-attachments/assets/82c7f436-9b93-4844-bdc6-b04044ce8a76

vadz:
Yes, this looks bad and should definitely be fixed, but unfortunately this library doesn't have an active maintainer and I don't know it enough and don't use it myself, so it's hard to find the motivation to spend a lot of time on doing it. Of course, if somebody could contribute a fix for this, it would be great.

lsri2:
Hi, guys, yeah I wasn't sure of a fix.  What I did that seems to help it a bit was comment out this code in msw\dc.cpp DoGetTextExtent().  Not a real fix, but I just wanted to share.

```
  /*  const wxFont font2 = GetFont();
   
  if ( font2.IsOk() && font2.GetStyle() != wxFONTSTYLE_NORMAL && len > 0 )
    {
        ABC width;
        const wxChar chFirst = *string.begin();
[...]

PR #26105, opened 2026-01-21
wxRichTextCtrl: Fix selection jitter on macOS

torstenvl:
On macOS, selecting text in wxRichTextCtrl caused visible horizontal jitter as the text shifted left and right during selection. This was caused by splitting DrawText() calls at selection boundaries, which made each text chunk get shaped independently with different kerning.

This commit introduces DrawTabbedStringWithPartialSelection() for macOS, which renders text in a single pass (splitting only at tabs), eliminating the kerning discontinuities that caused jitter.

The implementation matches native macOS behavior where selected text maintains its original styling (like TextEdit in Rich Text mode), with the selection highlight drawn as a background rectangle that completely overrides any text background colors in the selected range.

See #22771 (unable to reproduce on Windows, but confirmed on macOS)

Selection in TextEdit:

PR 915e340 TextEdit Selection

Selection in wxRichTextCtrl with this commit:

PR 915e340 wxRTC Selection

Demonstration of lack of left-right jitter:

https://github.com/user-attachments/assets/754bd6e8-ce9a-4b24-8ea9-72b5d38573cf

### DIFF
--- a/include/wx/richtext/richtextbuffer.h
+++ b/include/wx/richtext/richtextbuffer.h
@@ -4765,10 +4765,7 @@ public:
     wxNODISCARD virtual wxRichTextObject* Clone() const override { return new wxRichTextPlainText(*this); }
 
 private:
-    bool DrawTabbedString(wxDC& dc, const wxRichTextAttr& attr, const wxRect& rect, wxString& str, wxCoord& x, wxCoord& y, bool selected);
-#ifdef __WXMAC__
-    bool DrawTabbedStringWithPartialSelection(wxDC& dc, const wxRichTextAttr& attr, const wxRect& rect, wxString& str, wxCoord& x, wxCoord& y, int selStart, int selEnd);
-#endif
+    void DrawTabbedString(wxDC& dc, const wxRichTextAttr& attr, const wxRect& rect, wxString& str, wxCoord& x, wxCoord& y, ssize_t selStart, ssize_t selEnd);
 
 protected:
     wxString    m_text;
--- a/include/wx/richtext/richtextbuffer.h
+++ b/include/wx/richtext/richtextbuffer.h
@@ -4766,6 +4766,9 @@ public:
 
 private:
     bool DrawTabbedString(wxDC& dc, const wxRichTextAttr& attr, const wxRect& rect, wxString& str, wxCoord& x, wxCoord& y, bool selected);
+#ifdef __WXMAC__
+    bool DrawTabbedStringWithPartialSelection(wxDC& dc, const wxRichTextAttr& attr, const wxRect& rect, wxString& str, wxCoord& x, wxCoord& y, int selStart, int selEnd);
+#endif
 
 protected:
     wxString    m_text;
--- a/src/richtext/richtextbuffer.cpp
+++ b/src/richtext/richtextbuffer.cpp
@@ -6934,172 +6934,74 @@ bool wxRichTextPlainText::Draw(wxDC& dc, wxRichTextDrawingContext& context, cons
         y = rect.y + (rect.height - charHeight - (descent - m_descent));
     }
 
-    // TODO: new selection code
-
     // (a) All selected.
     if (allSelected)
     {
-#ifdef __WXMAC__
-        DrawTabbedStringWithPartialSelection(dc, textAttr, rect, stringChunk, x, y, 0, stringChunk.Len() - 1);
-#else
-        DrawTabbedString(dc, textAttr, rect, stringChunk, x, y, true);
-#endif
+        DrawTabbedString(dc, textAttr, rect, stringChunk, x, y, 0, stringChunk.Len() - 1);
     }
     // (b) None selected.
     else if (noneSelected)
     {
-#ifdef __WXMAC__
-        DrawTabbedStringWithPartialSelection(dc, textAttr, rect, stringChunk, x, y, -1, -1);
-#else
         // Draw all unselected
-        DrawTabbedString(dc, textAttr, rect, stringChunk, x, y, false);
-#endif
+        DrawTabbedString(dc, textAttr, rect, stringChunk, x, y, -1, -1);
     }
     else
     {
-#ifdef __WXMAC__
-        int selStart = wxMax(selectionRange.GetStart(), range.GetStart()) - range.GetStart();
-        int selEnd = wxMin(selectionRange.GetEnd(),   range.GetEnd()) - range.GetStart();
-        DrawTabbedStringWithPartialSelection(dc, textAttr, rect, stringChunk, x, y, selStart, selEnd);
-#else
-        // (c) Part selected, part not
-        // Let's draw unselected chunk, selected chunk, then unselected chunk.
-
-        const wxString& str = stringWhole;
-
-        dc.SetBackgroundMode(wxBRUSHSTYLE_TRANSPARENT);
-
-        // 1. Initial unselected chunk, if any, up until start of selection.
-        if (selectionRange.GetStart() > range.GetStart() && selectionRange.GetStart() <= range.GetEnd())
-        {
-            int r1 = range.GetStart();
-            int s1 = selectionRange.GetStart()-1;
-            int fragmentLen = s1 - r1 + 1;
-            if (fragmentLen < 0)
-            {
-                wxLogDebug(wxT("Mid(%d, %d"), (int)(r1 - offset), (int)fragmentLen);
-            }
-            wxString stringFragment = str.Mid(r1 - offset, fragmentLen);
-
-            DrawTabbedString(dc, textAttr, rect, stringFragment, x, y, false);
-
-#if USE_KERNING_FIX
-            if (stringChunk.Find(wxT("\t")) == wxNOT_FOUND)
-            {
-                // Compensate for kerning difference
-                wxString stringFragment2(str.Mid(r1 - offset, fragmentLen+1));
-                wxString stringFragment3(str.Mid(r1 - offset + fragmentLen, 1));
-
-                wxCoord w1, h1, w2, h2, w3, h3;
-                dc.GetTextExtent(stringFragment,  & w1, & h1);
-                dc.GetTextExtent(stringFragment2, & w2, & h2);
-                dc.GetTextExtent(stringFragment3, & w3, & h3);
-
-                int kerningDiff = (w1 + w3) - w2;
-                x = x - kerningDiff;
-            }
-#endif
-        }
-
-        // 2. Selected chunk, if any.
-        if (selectionRange.GetEnd() >= range.GetStart())
-        {
-            int s1 = wxMax(selectionRange.GetStart(), range.GetStart());
-            int s2 = wxMin(selectionRange.GetEnd(), range.GetEnd());
-
-            int fragmentLen = s2 - s1 + 1;
-            if (fragmentLen < 0)
-            {
-                wxLogDebug(wxT("Mid(%d, %d"), (int)(s1 - offset), (int)fragmentLen);
-            }
-            wxString stringFragment = str.Mid(s1 - offset, fragmentLen);
-
-            DrawTabbedString(dc, textAttr, rect, stringFragment, x, y, true);
-
-#if USE_KERNING_FIX
-            if (stringChunk.Find(wxT("\t")) == wxNOT_FOUND)
-            {
-                // Compensate for kerning difference
-                wxString stringFragment2(str.Mid(s1 - offset, fragmentLen+1));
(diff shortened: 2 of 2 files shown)


torstenvl:
Two force-pushed updates:

1. To remove a single trailing space in a comment, which caused a failure in a code check.
2. No changes, force-pushed only to force restart of checks after the appveyor build failed due to (what appears to be) a transient network issue installing the Golang MSI. 

IMG_0164

src/richtext/richtextbuffer.cpp
```diff
@@ -7036,11 +7049,171 @@ bool wxRichTextPlainText::Draw(wxDC& dc, wxRichTextDrawingContext& context, cons
...
+
+    wxColour highlightBackgroundColor(wxSystemSettings::GetColour(wxSYS_COLOUR_HIGHLIGHT));
+    wxColour standardTextColor = (attr.HasTextColour() && attr.GetTextColour().IsOk()) ? attr.GetTextColour() : wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOWTEXT);
+    wxCoord originx = x;
+    wxArrayInt startpos;  // Character starting positions
+    wxArrayInt endpos;    // Character ending positions (cumulative widths)
+    wxArrayInt tabstops;  // Array of tab stop positions
+    wxPen savedpen;
+
+    int strikethrough = (attr.HasTextEffects() && (attr.GetTextEffects() & wxTEXT_ATTR_EFFECT_STRIKETHROUGH)) ? 1 : 0;
```
vadz: This looks like a `bool`, why does it have `int` type?

```suggestion
    const bool strikethrough = (attr.GetTextEffects() & wxTEXT_ATTR_EFFECT_STRIKETHROUGH)) != 0;
```

src/richtext/richtextbuffer.cpp
```diff
@@ -7036,11 +7049,171 @@ bool wxRichTextPlainText::Draw(wxDC& dc, wxRichTextDrawingContext& context, cons
...
+    int strikeheight = (int) (rect.y + rect.GetHeight()/2.0 + 0.5);  // Cast truncates; +0.5 forces rounding
```
vadz: Or could just use `wxRound()`:

```suggestion
    const int strikeheight = wxRound(rect.y + rect.GetHeight()/2.0);
```

src/richtext/richtextbuffer.cpp
```diff
@@ -7036,11 +7049,171 @@ bool wxRichTextPlainText::Draw(wxDC& dc, wxRichTextDrawingContext& context, cons
...
+    int si = 0;  // string iterator
+    int ti = 0;  // tabstop iterator
+    int pi = 0;  // position iterator
+
+
+    ///////////////////////////////////////////////////////////////////////////
+    //                             Initial setup                             //
+    ///////////////////////////////////////////////////////////////////////////
+
+    if (str.Len() == 0) return true; // Nonsensical but it's what the original does
```
vadz: If it's really nonsensical, i.e. if we can guarantee that the string is never empty, I'd just remove it. Otherwise

```suggestion
    if (str.empty())
        return true; // Nonsensical but it's what the original does
```

src/richtext/richtextbuffer.cpp
```diff
@@ -7036,11 +7049,171 @@ bool wxRichTextPlainText::Draw(wxDC& dc, wxRichTextDrawingContext& context, cons
...
+
+    // Get character ending positions
+    dc.GetPartialTextExtents(str, endpos);
+    startpos.SetCount(endpos.GetCount());
+
+    startpos[0] = originx;
+    for (i = 0; i < endpos.GetCount() - 1; i++) {
```
vadz: wx style mandates putting `{` on its own line, i.e.

```suggestion
    for (i = 0; i < endpos.GetCount() - 1; i++)
    {
```

(here and below).

include/wx/richtext/richtextbuffer.h
```diff
@@ -4766,6 +4766,9 @@ class WXDLLIMPEXP_RICHTEXT wxRichTextPlainText: public wxRichTextObject
 
 private:
     bool DrawTabbedString(wxDC& dc, const wxRichTextAttr& attr, const wxRect& rect, wxString& str, wxCoord& x, wxCoord& y, bool selected);
+#ifdef __WXMAC__
+    bool DrawTabbedStringWithPartialSelection(wxDC& dc, const wxRichTextAttr& attr, const wxRect& rect, wxString& str, wxCoord& x, wxCoord& y, int selStart, int selEnd);
```
vadz: AFAICS the return value is always ignored, why have it?

src/richtext/richtextbuffer.cpp
```diff
@@ -7036,11 +7049,171 @@ bool wxRichTextPlainText::Draw(wxDC& dc, wxRichTextDrawingContext& context, cons
...
+    // settings... but the separate DrawText() calls cause left-right jitter during selection because of aggressive
+    // kerning.
+    if (attr.HasBackgroundColour() && attr.GetBackgroundColour().IsOk()) {
+        wxCoord w, h;
+        int backgroundWidth = endpos[endpos.GetCount() - 1] - startpos[0] + 1; // Avoid kerning gaps
+        savedpen = dc.GetPen();
+        wxCheckSetBrush(dc, wxBrush(attr.GetBackgroundColour()));
+        wxCheckSetPen(dc, wxPen(attr.GetBackgroundColour()));
+        dc.GetTextExtent(str, &w, &h);  // Get actual text height
+        wxRect bgRect(startpos[0], y, backgroundWidth, h);
```
vadz: `GetTextExtent()` overload return `wxSize` is more convenient and concise to use:

```suggestion
        wxRect bgRect(startpos[0], y, backgroundWidth, dc.GetTextExtent(str).y);
```

(and `w` and `h` don't need to be declared any more).

src/richtext/richtextbuffer.cpp
```diff
@@ -7036,11 +7049,171 @@ bool wxRichTextPlainText::Draw(wxDC& dc, wxRichTextDrawingContext& context, cons
...
+    }
+
+    ///////////////////////////////////////////////////////////////////////////
+    //                        Draw selection rectangle                       //
+    ///////////////////////////////////////////////////////////////////////////
+
+    if (selStart >=0 && selStart < endpos.GetCount()) {
+        if (selEnd >= selStart && selEnd < endpos.GetCount()) {
+            int selectionWidth = endpos[selEnd] - startpos[selStart] + 1; // Avoid kerning gaps
+            savedpen = dc.GetPen();
```
vadz: This really ought to be done using `wxDCPenChanger`, i.e.

```suggestion
            wxDCPenChanger changePen(dc, highlightBackgroundColor);
```

(and remove `wxCheckSetPen()` below).

This is not completely identical because we lose the checks done in `wxCheckSetPen()` but they don't seem relevant anyhow here because the pen colour does change.

src/richtext/richtextbuffer.cpp
```diff
@@ -7036,11 +7049,171 @@ bool wxRichTextPlainText::Draw(wxDC& dc, wxRichTextDrawingContext& context, cons
...
+            wxCheckSetBrush(dc, wxBrush(highlightBackgroundColor));
+            wxCheckSetPen(dc, wxPen(highlightBackgroundColor));
+            wxRect selRect(startpos[selStart], rect.y, selectionWidth, rect.GetHeight());
+            dc.DrawRectangle(selRect);
```
vadz: This could also be simplified to just

```suggestion
            dc.DrawRectangle(startpos[selStart], rect.y, selectionWidth, rect.GetHeight());
```

(unless you want to have this variable to show that it's the selection rect? but this seems clear enough anyhow...)

src/richtext/richtextbuffer.cpp
```diff
@@ -7036,11 +7049,171 @@ bool wxRichTextPlainText::Draw(wxDC& dc, wxRichTextDrawingContext& context, cons
...
+            if (selStart <= si && si <= selEnd && strikethrough) {
+                // Matching original behavior of only showing strikethrough on tabs when selected
+                dc.DrawLine(startpos[si], strikeheight, endpos[si], strikeheight);
+            }
+            printstart = si + 1;
+        }
+    }
+    if (printstart < str.Len()) {
+        dc.DrawText(str.Mid(printstart), startpos[printstart], y);
+        if (strikethrough) dc.DrawLine(startpos[printstart], strikeheight, endpos[endpos.GetCount()-1], strikeheight);
```
vadz: Please don't put compound statements on the same line.

```suggestion
        if (strikethrough)
            dc.DrawLine(startpos[printstart], strikeheight, endpos[endpos.GetCount()-1], strikeheight);
```

include/wx/richtext/richtextbuffer.h
```diff
@@ -4766,6 +4766,9 @@ class WXDLLIMPEXP_RICHTEXT wxRichTextPlainText: public wxRichTextObject
 
 private:
     bool DrawTabbedString(wxDC& dc, const wxRichTextAttr& attr, const wxRect& rect, wxString& str, wxCoord& x, wxCoord& y, bool selected);
+#ifdef __WXMAC__
```
vadz: It's not great to have checks for Mac, especially in the headers. Could you please change this to:

1. Keep the current function name ("WithPartialSelection" is not really useful because it is also called when there is no selection and everything is selected).
2. Replace its `bool selected` parameter with `int selStart, int selEnd` (alternatively, you could use `wxRange selection`, see `wx/range.h`).
3. Only use `#ifdef __WXMAC__` to selection between the implementations in the source file.

I think this would make things a bit tidier.

src/richtext/richtextbuffer.cpp
```diff
@@ -6939,16 +6939,29 @@ bool wxRichTextPlainText::Draw(wxDC& dc, wxRichTextDrawingContext& context, cons
...
         DrawTabbedString(dc, textAttr, rect, stringChunk, x, y, false);
+#endif
     }
     else
     {
+#ifdef __WXMAC__
+        int selStart = wxMax(selectionRange.GetStart(), range.GetStart()) - range.GetStart();
+        int selEnd = wxMin(selectionRange.GetEnd(),   range.GetEnd()) - range.GetStart();
+        DrawTabbedStringWithPartialSelection(dc, textAttr, rect, stringChunk, x, y, selStart, selEnd);
+#else
```
vadz: All the code in this `#else` branch would need to be moved to the new version of `DrawTabbedString()`.

vadz (commented):
Thanks for your work on this! We definitely need to merge this to fix the user-visible problem, but it would be nice to clean this up a bit before merging, notably to reduce the number of preprocessor checks.

Please let me know if you think you could do it. TIA!

torstenvl:
Not a problem! As I'm a new contributor, I was conservative about existing code in a lot of places, but I can absolutely do some refactoring to make the end result nicer. 

Especially with the existing code's line 6937 `// TODO: new selection code` (from Julian Smart, 16 years ago) 😂🤣 

Understand the style preference not to have single line `if` statements and I can work with that.

src/richtext/richtextbuffer.cpp
torstenvl: No issues, I just used the existing code (and added the comment since I kept forgetting why they did it this way). Original in DrawTabbedString is `dc.DrawLine(x, (int) (y+(h/2)+0.5), x+w, (int) (y+(h/2)+0.5));`

src/richtext/richtextbuffer.cpp
torstenvl: The original doesn't use partial text extents, so there's no subscripting of the character positions, so they don't early return. However, it returns `true` under all code paths, so I borrowed that behavior. As you point out, the original is a `bool` but the return value is never used (and it's a private method, so the end-user won't use the return value either). Will change to `void` during refactor.

torstenvl:
Did a complete refactor of `wxRichTextPlainText::DrawTabbedString` to minimize platform-specific code and localize it in one place (with an explanation for why we took different, platform-specific approaches).  Verified that selection now works correctly on Mac, Windows, and Linux/GTK, with no left-right jitter bug. 

Mac:

https://github.com/user-attachments/assets/3f360e0e-b150-49ae-a2e9-5e093b58fc63

Windows:

https://github.com/user-attachments/assets/7bb9285d-55da-419b-9f21-1c26c1ccdad8

Linux:

https://github.com/user-attachments/assets/0c872bc2-1d50-497e-8819-e875cc1d2c51

vadz:
Thanks for the update!

I've (squash) merged it now with only just some very minor formatting changes (to avoid really  overlong lines) and use of `std::string`-like `length()` instead of old `Len()` for string lengths.

Thanks again for fixing this!

PR #26109, opened 2026-01-24
Add wxSYS_COLOUR_LISTBOXHIGHLIGHT

RobertRoeb:
- There was no way to query the main selection colour on OSX as wxSYS_COLOUR_HIGHLIGHT returns the colour of text selection, not item selection
  - Also queries the GtkTreeView selected item background colour on  GTK3 and GTK2
  - This new colour now reports what is essentially the main theme or accent colour on all platforms
  - Blindly mapped to wxSYS_COLOUR_HIGHLIGHT on other platforms


vadz:
Thanks, looks good, but this breaks wxGTK2, wxX11 and wxMSW as you can see, so it would need to be fixed before being applied.

vadz (commented):
Thanks for the fixes, this can be merged but I wanted to check if the value of this was correct for MSW dark mode before merging and realized that I can't because this is not used anywhere yet — shouldn't it be actually used somewhere?

Also, it should be added to `sysColours` in `samples/drawing/drawing.cpp` (but I can do this myself when merging).

RobertRoeb:
Fair point. I use it in my own wxDataViewProgressRenderer as the native ones look - not so good.

vadz:
So what should I compare it with under Windows? Background of which item shown here?

image

("tests" is hot, i.e. under mouse currently, ".git-blame-revs" is current, i.e. focused, the other highlighted items are just selected)

RobertRoeb:
I think it should be cirrus.yml - but it is all grey here
.. And sure, someone can create a rainbow theme where everything is psychedelic. But I assume most standard themes do have a common selection and theme colour in list box and elsewhere and that is what I think should go in here

vadz:
> I think it should be cirrus.yml - but it is all grey here

This is `wxSYS_COLOUR_MENUBAR`, not `wxSYS_COLOUR_MENUHILIGHT`, so I'll change the new constant to be equivalent to the former, not the latter.

> .. And sure, someone can create a rainbow theme where everything is psychedelic.

This is standard dark theme of Windows Explorer, so it's definitely not something out of the ordinary.

PR #26110, opened 2026-01-24
Fix wrong destruction of wxBitmap in wxMSW wxStaticBitmap

### LINKED ISSUE #26106: Regression in wxMSW `wxStaticBitmap`
It's probably a separate bug, but if we sequentially set two bitmaps without alpha, does the handle of the first bitmap get deleted by `::DeleteObject(oldHandle);`?

It appears to be the case in https://gitlab.com/kicad/code/kicad/-/issues/22767

(the above is likely a result of https://github.com/wxWidgets/wxWidgets/commit/3e32a9abe1e05f0cddb17c791a59977bed8e8cff)

And looks like it's backported to 3.2.9 too https://github.com/wxWidgets/wxWidgets/pull/25674

_Originally posted by @dsa-t in https://github.com/wxWidgets/wxWidgets/issues/25699#issuecomment-3768725219_

vadz:
Due to a regression introduced in 3e32a9abe1 (Don't bother resetting wxStaticBitmap image when destroying it, 2025-06-12), which was part of #25518, the handle of wxBitmap passed to wxStaticBitmap::SetBitmap() was destroyed when another bitmap was passed to SetBitmap() later.

Fix this by ensuring that we don't overwrite the value of the current handle (to be deleted) prematurely in DoUpdateImage().

Add a simple unit test for wxStaticBitmap checking that the problem is really fixed and, under MSW, also that it doesn't leak bitmap handles.

Closes #26106.

### DIFF
--- a/src/msw/statbmp.cpp
+++ b/src/msw/statbmp.cpp
@@ -231,6 +231,7 @@ void wxStaticBitmap::DoUpdateImage(const wxSize& sizeOld, bool wasIcon)
 
     // For the icons we just use its HICON directly, but for bitmaps we create
     // our own temporary bitmap and need to delete its handle manually later.
+    WXHANDLE currentHandle = 0;
     if ( !m_icon.IsOk() )
     {
         wxBitmap bitmap = m_bitmapBundle.GetBitmapFor(this);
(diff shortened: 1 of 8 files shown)


vadz:
@dsa-t Please check if you have any problems with this. I could reproduce the problem with the new unit test added in this PR before the fix and it passes now, so I'm relatively confident that it should fix the problem for you too, but any additional review/testing would be welcome as this code is not that simple (I hoped to find some way to simplify it, but it's not really obvious how to do it considering the weird ownership rules being used here).

dsa-t:
Thanks, looks good!

Hopefully we can test it more widely once wx 3.3.2 and KiCad 10.0.0-rc1 is released.

MaartenBent:
Can you add the new file to `build/cmake/tests/gui/CMakeLists.txt`?

vadz:
> Can you add the new file to `build/cmake/tests/gui/CMakeLists.txt`?

Oops, sorry for forgetting about this. Done locally, will push soon.

PR #26111, opened 2026-01-24
Make wxXmlInitResourceModule() safer to use

vadz:
This function unconditionally added a new wxXmlResourceModule object to the modules list which could result in problems if it had been already initialized.

Ensure that we do this only if necessary by calling the new function added to wxModule which checks if the module is already known and adds and initializes only this module if it is not.

See #26039.


vadz:
Merged in the commit above.

PR #26116, opened 2026-01-25
Refactor wxBitmapBundleImplSVG

MaartenBent:
Reduce code duplication by adding `wxBitmapBundleLunaSVG` and `wxBitmapBundleNanoSVG` that implement `DoRasterize()`.

Allow to use SVG size in `wxBitmapBundle::FromSVG()`.
If no default size is provided, get the size from the SVG.

### DIFF
--- a/src/generic/bmpsvg.cpp
+++ b/src/generic/bmpsvg.cpp
@@ -80,10 +80,13 @@ public:
     }
 
     virtual bool IsOk() const = 0;
+    virtual wxSize GetSVGSize() const = 0;
     virtual wxBitmap DoRasterize(const wxSize& size) = 0;
 
     virtual wxSize GetDefaultSize() const override
     {
+        if ( m_sizeDef.IsEmpty() )
+            return GetSVGSize();
         return m_sizeDef;
     }
 
@@ -150,6 +153,13 @@ public:
     virtual bool IsOk() const override
     {
         return m_svgDocument != nullptr;
+    }
+
+    virtual wxSize GetSVGSize() const override
+    {
+        if ( IsOk() )
+            return wxSize(m_svgDocument->width(), m_svgDocument->height());
+        return wxDefaultSize;
     }
 
     virtual wxBitmap DoRasterize(const wxSize& size) override
@@ -306,6 +316,13 @@ public:
         // the data is not SVG at all, e.g. without this check creating a bundle
         // from any random file with FromSVGFile() would "work".
         return m_svgImage && m_svgImage->width != 0 && m_svgImage->height != 0 && m_svgImage->shapes;
+    }
+
+    virtual wxSize GetSVGSize() const override
+    {
+        if ( IsOk() )
+            return wxSize(m_svgImage->width, m_svgImage->height);
+        return wxDefaultSize;
     }
 
     virtual wxBitmap DoRasterize(const wxSize& size) override
--- a/src/generic/bmpsvg.cpp
+++ b/src/generic/bmpsvg.cpp
@@ -19,13 +19,14 @@
 // for compilers that support precompilation, includes "wx.h".
 #include "wx/wxprec.h"
 
-// This is required by NanoSVG headers. Include it before any wx headers to
-// avoid potential conflicts (see comments in nanosvg implementation section).
-#include <stdio.h>
-
 #ifdef wxHAS_SVG
 
 #include "wx/bmpbndl.h"
+
+#ifndef WX_PRECOMP
+    #include "wx/log.h"
+    #include "wx/utils.h"                   // Only for wxMin()
+#endif
 
 #if wxUSE_FFILE
     #include "wx/ffile.h"
@@ -36,10 +37,6 @@
 #endif
 
 #include "wx/rawbmp.h"
-#include "wx/private/bmpbndl.h"
-
-#include "wx/buffer.h"
-#include "wx/log.h"
 
 // ============================================================================
 // private helpers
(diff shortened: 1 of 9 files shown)


src/generic/bmpsvg.cpp
```diff
@@ -69,131 +60,108 @@ static wxCharBuffer LoadSVGFile(const wxString& path)
...
+private:
+    wxBitmap m_cachedBitmap;
+    const wxSize m_sizeDef;
 
-    virtual wxSize GetDefaultSize() const override;
-    virtual wxSize GetPreferredBitmapSizeAtScale(double scale) const override;
+    wxDECLARE_NO_COPY_CLASS(wxBitmapBundleImplSVG);
 
-    virtual wxBitmap GetBitmap(const wxSize& size) override;
+protected:
```
vadz: Extremely minor, but why do we have an empty `protected:` section?

src/generic/bmpsvg.cpp
```diff
@@ -69,131 +60,108 @@ static wxCharBuffer LoadSVGFile(const wxString& path)
...
 
-    bool IsOk() const;
+public:
+    wxBitmapBundleImplSVG(const wxSize& sizeDef)
```
vadz: Also minor because it's a private class, but I'd still use `explicit`:

```suggestion
    explicit wxBitmapBundleImplSVG(const wxSize& sizeDef)
```

src/generic/bmpsvg.cpp
```diff
@@ -69,131 +60,108 @@ static wxCharBuffer LoadSVGFile(const wxString& path)
...
-private:
-    wxBitmap DoRasterize(const wxSize& size);
+    virtual bool IsOk() const = 0;
+    virtual wxSize GetSVGSize() const = 0;
+    virtual wxBitmap DoRasterize(const wxSize& size) = 0;
 
-    std::unique_ptr<wxlunasvg::Document> m_svgDocument;
+    virtual wxSize GetDefaultSize() const override
+    {
+        if ( m_sizeDef.IsEmpty() )
```
vadz, line 88: Is `wxDefaultSize` getting translated to `wxSize{}` somewhere before? Because the docs say that either of them can be used, but we only check for the latter here.

src/generic/bmpsvg.cpp
```diff
@@ -69,131 +60,108 @@ static wxCharBuffer LoadSVGFile(const wxString& path)
...
+        return GetDefaultSize() * scale;
+    }
 
-    // Cache the last used bitmap (may be invalid if not used yet).
-    //
-    // Note that we cache only the last bitmap and not all the bitmaps ever
-    // requested from GetBitmap() for the different sizes because there would be
-    // no way to clear such cache and its growth could be unbounded, resulting
-    // in too many bitmap objects being used in an application using SVG for all
-    // of its icons.
```
vadz, line 132: I think this comment was useful, why remove it?

src/generic/bmpsvg.cpp
```diff
@@ -8,25 +8,19 @@
 // Licence:     wxWindows licence
 ///////////////////////////////////////////////////////////////////////////////
 
-// ============================================================================
-// declarations
-// ============================================================================
-
-// ----------------------------------------------------------------------------
-// headers
-// ----------------------------------------------------------------------------
```
vadz, line 17: Very minor, but why remove this, especially while leaving other similar comments below?

src/generic/bmpsvg.cpp
```diff
@@ -69,131 +60,108 @@ static wxCharBuffer LoadSVGFile(const wxString& path)
...
-    wxBitmapBundleImplSVG(const char* data, const wxSize& sizeDef);
-
-    // data must be 0 terminated. wxBitmapBundleImplSVG doesn't take ownership
-    // so it can be deleted after the ctor is called.
-    wxBitmapBundleImplSVG(char* data, const wxSize& sizeDef);
-
-    // wxBitmapBundleImplSVG doesn't take ownership of data so it can be deleted
-    // after the ctor is called.
-    wxBitmapBundleImplSVG(const wxByte* data, size_t data_len, const wxSize& sizeDef);
+private:
```
vadz: I know that some people prefer to put data first in the class declaration, but in wx convention we always put public part first, not sure if it's worth deviating from it here (in this class and the ones derived from it).

src/generic/bmpsvg.cpp
```diff
@@ -359,158 +281,125 @@ wxGCC_WARNING_RESTORE(cast-qual)
...
 /* static */
 wxBitmapBundle wxBitmapBundle::FromSVG(char* data, const wxSize& sizeDef)
 {
-    NSVGimage* const svgImage = nsvgParse(data, "px", 96);
-    if ( !svgImage )
-        return wxBitmapBundle();
-
-    // Somewhat unexpectedly, a non-null but empty image is returned even if
-    // the data is not SVG at all, e.g. without this check creating a bundle
-    // from any random file with FromSVGFile() would "work".
```
vadz, line 506: And I think this comment is worth preserving as well.

vadz (approved):
Thanks, this is definitely nicer and not requiring to provide the size is good too.

No real comments but a couple of nitpicks (that I could also address myself when merging if you prefer).

Thanks again!

src/generic/bmpsvg.cpp
MaartenBent: I was copying relevant parts from the old file to my new file, and didn't include everything. I'll restore this and the other comments.

src/generic/bmpsvg.cpp
MaartenBent: A left-over from experimenting with the default size, I'll remove it.

src/generic/bmpsvg.cpp
MaartenBent: `IsEmpty()` checks for `x <= 0 || y <= 0`, so this includes `wxDefaultSize`/`wxSize(-1,-1)`.

vadz: Oh, sorry for missing this, I either didn't know or completely forgot that it worked like this.

src/generic/bmpsvg.cpp
MaartenBent: Forgot this one. Will push it after seeing the wxQt test results

vadz:
I forgot to cc @Randalphwa on this just in case you'd like to review it.

Randalphwa:
Looks fine to me.

vadz:
Thanks again, will merge in a moment.

OT: AppVeyor regularly fails to download `https://go.dev/dl/go%GOVERSION%.windows-amd64.msi` now, no idea what's going on :-(

PR #26117, opened 2026-01-25
Use bitmapbundle for wxRibbonControl

Randalphwa:
This is a draft. I have updated the code and verified that the ribbon sample works correctly. I want to submit this first for code review. Then once the code passes review, I will update the sample to add SVG files, multi-resolution bitmap bundles, etc. Lastly, I'll update the documentation.


include/wx/ribbon/bar.h
```diff
@@ -161,8 +161,10 @@ class WXDLLIMPEXP_RIBBON wxRibbonBar : public wxRibbonControl
 
     void HideIfExpanded();
 
-    // Return the image list containing images of the given size, creating it
-    // if necessary.
+    // Deprecated: wxRibbonButtonBar now uses wxWithImages internally for
+    // DPI-aware bitmap management. This method is maintained for backward
+    // compatibility but is no longer used by wxRibbonButtonBar.
+    wxDEPRECATED_MSG("wxRibbonButtonBar now uses wxWithImages for DPI support")
```
vadz: Does it really use `wxWithImages`? It doesn't look so...

include/wx/ribbon/buttonbar.h
```diff
@@ -16,6 +16,8 @@
 #include "wx/ribbon/art.h"
 #include "wx/ribbon/control.h"
 #include "wx/dynarray.h"
+#include "wx/bmpbndl.h"
+#include "wx/vector.h"
```
vadz: Not really necessary because already included from `wx/bmpbndl.h` but, also, it's fine to just include `<vector>` in master/3.3.

include/wx/ribbon/buttonbar.h
```diff
@@ -208,9 +214,15 @@ class WXDLLIMPEXP_RIBBON wxRibbonButtonBar : public wxRibbonControl
     bool m_lock_active_state;
     bool m_show_tooltips_for_disabled;
 
+    wxVector<wxBitmapBundle> m_bundlesLarge;         // Large button icons
+    wxVector<wxBitmapBundle> m_bundlesSmall;         // Small button icons
+    wxVector<wxBitmapBundle> m_bundlesLargeDisabled; // Disabled large icons
+    wxVector<wxBitmapBundle> m_bundlesSmallDisabled; // Disabled small icons
```
vadz: Let's definitely not use `wxVector` any more:

```suggestion
    std::vector<wxBitmapBundle> m_bundlesLarge;         // Large button icons
    std::vector<wxBitmapBundle> m_bundlesSmall;         // Small button icons
    std::vector<wxBitmapBundle> m_bundlesLargeDisabled; // Disabled large icons
    std::vector<wxBitmapBundle> m_bundlesSmallDisabled; // Disabled small icons
```

vadz (commented):
Thanks for this!

I know this is still a draft, but I had a quick look just to check and left a couple of minor comments below, but globally this looks very good, thanks again!

Randalphwa (commented):
> Does it really use `wxWithImages`? It doesn't look so... 

No, wxRibbonButtonBar does NOT use wxWithImages. Instead, it uses a custom approach with std::vector<wxBitmapBundle> member variables.

Storage: It stores bitmap bundles in four std::vector containers:

m_bundlesLarge - Large button icons
m_bundlesSmall - Small button icons
m_bundlesLargeDisabled - Disabled large icons
m_bundlesSmallDisabled - Disabled small icons

I have changed the deprecated comment.

vadz:
FYI: I've converted this to draft to indicate its status because I don't think this should be considered for merging yet, but please let me know if I misunderstood.

Randalphwa:
Thanks, I had *not* intended it to be mergeable yet -- I still have a bit more work to do on the sample before it's ready for merging.

Randalphwa:
I finally got a second monitor setup again so that I could verify that DPI scaling properly changes when the ribbon demo is dragged from one monitor to the other. As far as I can tell, it's working fine. I also checked in on an existing app that had a ribbon panel, and that continued to work fine with no changes.

src/ribbon/gallery.cpp
```diff
@@ -518,9 +520,10 @@ void wxRibbonGallery::OnPaint(wxPaintEvent& WXUNUSED(evt))
         else
             offset_pos.SetLeft(offset_pos.GetLeft() - m_scroll_amount);
         m_art->DrawGalleryItemBackground(dc, this, offset_pos, item);
-        if (item->GetBitmap().IsOk() )
-            dc.DrawBitmap(item->GetBitmap(), offset_pos.GetLeft() + padding_left,
-                offset_pos.GetTop() + padding_top);
+        // Resolve bitmap bundle for current DPI
```
vadz, line 523: We've lost the `IsOk()` check here, is it intentional?

src/ribbon/panel.cpp
```diff
@@ -704,19 +711,24 @@ bool wxRibbonPanel::Realize()
...
-            img.Rescale(scale * bitmap_size, wxIMAGE_QUALITY_HIGH);
-            m_minimised_icon_resized = wxBitmap(img, -1, scale);
-        }
-        else
-        {
-            m_minimised_icon_resized = m_minimised_icon;
+                wxImage img(icon_bitmap.ConvertToImage());
+                img.Rescale(scale * bitmap_size, wxIMAGE_QUALITY_HIGH);
+                m_minimised_icon_resized = wxBitmap(img, -1, scale);
+            }
```
vadz: We shouldn't do this any longer, using `bundle.GetBitmap(bundle.GetPreferredBitmapSizeFor(this))` should be strictly better.

samples/ribbon/ribbondemo.cpp
```diff
@@ -260,30 +262,204 @@ EVT_RIBBONBUTTONBAR_CLICKED(ID_CHANGE_LABEL, MyFrame::OnChangeLabel)
...
+// The MakeSvgBundle() helper guarantees a valid (non-empty) bundle is always
+// returned, falling back to wxArtProvider when SVG support is unavailable.
+// ----------------------------------------------------------------------------
+
+namespace
+{
+
+// -- Shape icons -----------------------------------------------------------
+
+static const char circle_svg[] =
```
vadz, line 280: Do we intentionally embed the SVGs here but also add SVG files in `vector` subdirectory (which I'd probably call just `images` BTW)? I'd do one _or_ the other, but not both...

vadz (commented):

It looks good (although, admittedly, I may have skimmed over some files), but I still have a couple of questions, please check them and let me know if you're going to make any changes or if this should be merged as is. TIA!

src/ribbon/gallery.cpp
Randalphwa: Definitely a mistake -- I added the conditional back.

src/ribbon/panel.cpp
Randalphwa: Good point, refactored

samples/ribbon/ribbondemo.cpp
Randalphwa: CMake could actually do both -- create a svg_data.inc file at configure time that would be compiled into the executable. Alas, it *only* works with CMake, so not a viable option given the other build systems that we support.

Currently, these svg files are quite small and the advantage of embedding them in the source code is the executable an run without needing a path to external files.

I removed the vector directory.

vadz:
Thanks, will (squash) merge soon.

PR #26119, opened 2026-01-26
Update wxFrame documentation for macOS style

### LINKED ISSUE #26114: Is wxDIALOG_EX_METAL still relevant?
On macOS, if I call `SetExtraStyle()` with `wxDIALOG_EX_METAL`, I get an assert that I can't use Metal after the window has been created. However, thinking about it, it seems like `wxDIALOG_EX_METAL` is the old brushed metal theme under OSX that has long since been deprecated, right? Should this style be avoided and marked as deprecated? Thanks.

Blake-Madden:
Document deprecation of metallic look style on macOS.
Closes #26114

### DIFF
--- a/interface/wx/frame.h
+++ b/interface/wx/frame.h
@@ -134,6 +134,7 @@
     @style{wxFRAME_EX_METAL}
            On macOS, frames with this style will be shown with a metallic
            look. This is an extra style.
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, will push soon.

PR #26120, opened 2026-01-26
Fix `wxPGMultiButton` bitmap appearance 

vadz:
Unless I'm missing something, there is just no reason at all to have all this code in `wxPGMultiButton` as simply using `wxButton` with `wxBU_EXACTFIT` works just as well — and, in fact, significantly better because the bitmap in the sample is not completely ugly due to rescaling (at least in 200% DPI) any more.

Compare the old and new appearance of the buttons in the sample for GTK:

before: pg-gtk-old and now: pg-gtk-new

and MSW:

before: pg-msw-old and now: pg-msw-new

### DIFF
--- a/samples/widgets/button.cpp
+++ b/samples/widgets/button.cpp
@@ -511,7 +511,8 @@ void ButtonWidgetsPage::CreateButton()
         }
         else
         {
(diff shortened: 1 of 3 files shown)


dsa-t:
Thanks, looks good!

PR #26065, opened 2026-01-02
2026-01 backports to 3.2

vadz:
As usual, please comment to point out more commits to add to this PR.

PR #26121, opened 2026-01-27
Backport PR #25767 to 3.2

nobugshere:
This is a backport of PR #25767 to the wxWidgets 3.2 branch

Refer to #25767 and #26065

Compiled and did some testing on:
macOS M1Max Sequoia 15.7.3, Xcode 16.4
macOS Tahoe 26.2 VMware, Xcode 26.2

### DIFF
--- a/include/wx/osx/statusbr.h
+++ b/include/wx/osx/statusbr.h
@@ -36,7 +36,7 @@ protected:
     virtual void InitColours() wxOVERRIDE;
 
 private:
-    wxColour m_textActive, m_textInactive;
+    wxColour m_textActive, m_textInactive, m_bgActive, m_bgInactive, m_separator;
 
     wxDECLARE_DYNAMIC_CLASS(wxStatusBarMac);
     wxDECLARE_EVENT_TABLE();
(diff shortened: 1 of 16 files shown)


include/wx/osx/statusbr.h
```diff
@@ -36,7 +36,7 @@ class WXDLLIMPEXP_CORE wxStatusBarMac : public wxStatusBarGeneric
     virtual void InitColours() wxOVERRIDE;
 
 private:
-    wxColour m_textActive, m_textInactive;
+    wxColour m_textActive, m_textInactive, m_bgActive, m_bgInactive, m_separator;
```
vadz: This is an ABI-incompatible change, we can't do this in 3.2.

The simplest would be to just make the new variables globals in the source file as they have constant values and can't be changed anyhow.

vadz (changes requested):
Thanks, I didn't test this but I agree that it would be nice to have macOS 26 support in 3.2 too. However the ABI compatibility needs to be restored before this can be applied.

include/wx/osx/statusbr.h
nobugshere: Fixed as suggested.  Thank you.

vadz:
Thanks, I've included this in #26065, so closing this one.

I'd like to ask people using 3.2 under macOS to test these changes by applying either this PR or #26065 or just updating their 3.2 branch in a couple of days under Tahoe as I can't do it myself. TIA!

PR #26122, opened 2026-01-27
Fix type of wxAuiDefaultToolBarArt::Get{Label,Tool}Size "dc" parameters

swt2c:
They are now wxReadOnlyDC and not wxDC.


vadz:
Thanks! Will push soon.

PR #26123, opened 2026-01-27
Show generic `wxTreeCtrl` in the sample and improve its colours under MSW

vadz:
This addresses #26118 (also for generic list control) and also slightly improves the generic control appearance in dark mode, even though it is still not at all the same as Explorer, unfortunately.


dsa-t:
Thanks! Seems to work as expected.

PR #26124, opened 2026-01-28
Fix issues in the toolbar sample

0tkl:
The second commit is trivial. The first commit aims to fix a bug where the horizontal toolbar uses `wxBITMAP_BUNDLE_2` for proper HiDPI support, but the vertical toolbar icons are still using `wxBITMAP_PNG`, resulting in an inconsistent appearance, as shown below:

### DIFF
--- a/samples/toolbar/toolbar.cpp
+++ b/samples/toolbar/toolbar.cpp
@@ -815,12 +815,12 @@ void MyFrame::OnToggleAnotherToolbar(wxCommandEvent& WXUNUSED(event))
 
         m_tbar->SetMargins(4, 4);
 
-        m_tbar->AddRadioTool(IDM_TOOLBAR_OTHER_1, "First", wxBITMAP_PNG(new));
-        m_tbar->AddRadioTool(IDM_TOOLBAR_OTHER_2, "Second", wxBITMAP_PNG(open));
-        m_tbar->AddRadioTool(IDM_TOOLBAR_OTHER_3, "Third", wxBITMAP_PNG(save));
+        m_tbar->AddRadioTool(IDM_TOOLBAR_OTHER_1, "First", wxBITMAP_BUNDLE_2(new));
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, but while the first commit makes sense, I don't see at all why is the second one needed: what problem does it solve? It shouldn't be necessary to call `Realize()` after deleting a tool, our documentation for `DeleteTool()` explicitly says:

```
        @note It is unnecessary to call Realize() for the change to take
            place, it will happen immediately.
```

so why would we need to do it here?

0tkl:
If you want to reproduce the issue corresponding to the second commit, build and run the sample, click “Tools > Insert print button”, here's what I get:

image

And then click “Tools > Delete print button”, here's what I get before applying this patch:

image

Is the behavior shown here correct or incorrect?

vadz:
It seems "correct" to me: you insert a new print button (shown) and then delete it (not shown any more). It is confusing, however, as it's not clear which print button should be deleted.

Of course, if you don't add a new print button and just delete it, it will delete the (only) print button just fine.

0tkl:
We seem to have a slightly different understanding. What looks more natural and intuitive to me is that after adding and then removing that print button, the paddings in all buttons return to their original size. That's why I suggested to call `Realize()` again to refresh the padding.

image

vadz:
I didn't look at the padding at all, just at the buttons. In fact, even looking at it now, I'm not sure what exactly changes?

In any case, neither adding nor removing buttons should affect the padding at all and if it does, it's a bug, but calling `Realize()` shouldn't be necessary, it's not a bug in the sample.

I'll apply the first commit from this PR but the second one should probably be replaced by an issue and fixed in `wxToolBar` itself.

0tkl:
I didn't realize that was an internal implementation issue rather than a sample issue. I have now dropped that commit and opened a corresponding issue.

PR #26125, opened 2026-01-28
Remove implementation details from `wxGetLinuxDistribution()` docs

lanurmi:
An introduction to `lsb_release` is unnecessary, and saying it is very common on modern Linux distributions is an exaggeration, at least today.

And besides those, the details are not even accurate since aef7df6 (Read Linux distribution info from os-release file, 2023-07-14) (#23712), which made `lsb_release` the fallback method for obtaining information.

A candidate for backporting to 3.2.

### DIFF
--- a/interface/wx/platinfo.h
+++ b/interface/wx/platinfo.h
@@ -120,8 +120,7 @@ enum wxEndianness
 };
 
 /**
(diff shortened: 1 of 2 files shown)


vadz:
Makes sense, thanks.

PR #26126, opened 2026-01-28
Add print support to wxQt

AliKet:
The changes done to `configure` (related to `runstatedir`) were (mistakenly?) remove [here](6228eaf4494702a13d4538a1694daebab4ddb0b7) and now restored by the `autogen.sh`

### DIFF
--- a/build/cmake/toolkit.cmake
+++ b/build/cmake/toolkit.cmake
@@ -132,9 +132,9 @@ if(WXQT)
     find_package(QT NAMES Qt6 Qt5 REQUIRED COMPONENTS Core)
 
     if(QT_VERSION_MAJOR EQUAL 5)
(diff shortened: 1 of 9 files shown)


vadz:
Thanks, there is too much code here to really review it, but it looks superficially good and definitely seems better than nothing, so I'll merge it and let others test it.

AliKet:
> Thanks, there is too much code here to really review it, but it looks superficially good and definitely seems better than nothing, so I'll merge it and let others test it.

Thanks for merging this PR.

I admit I should have divided it into smaller commits to facilitate review. I was rushing to meet the deadline for the next release, and it ended up like this, sorry.

Anyway, the changes here are mostly inspired by the code in the `wxMSW` port, along with some helper functions for mapping between `wxWidgets` and `Qt` types.

Thanks again.

PR #26131, opened 2026-01-29
Add support for MSVS 2026

PBfordev:
Update wx_config.props, add MSVS 2026 solutions for the library and samples, and update the docs.

I have tried to unify using MSVS (IDE) and MSVC (compiler) in the matching context.

I have also slightly changed wording when it comes to the CRT (to not mention the old msvcrt explicitly).

### DIFF
--- a/docs/msw/install.md
+++ b/docs/msw/install.md
@@ -47,18 +47,24 @@ All makefiles and project are located in `build\msw` directory.
 Microsoft Visual C++ Compilation       {#msw_build_msvs}
 ----------------------------------------------------------------
 
+Note. All the instructions for building wxWidgets and using it in
+applications with Microsoft Visual Studio (MSVS) assume that Microsoft
+Visual C++ (MSVC) is being used as the C++ compiler (not Clang/LLVM).
+
 ### From the IDE
 
-Ready to use project files are provided for VC++ versions 2015, 2017, 2019 and 2022.
+Ready to use project files are provided for MSVS versions 2015, 2017, 2019,
+2022 and 2026.
 
-Simply open `wx_vcN.sln` (for N=14, 15, 16 or 17) file,
+Simply open `wx_vcN.sln` (for N=14, 15, 16 or 17 matching MSVS versions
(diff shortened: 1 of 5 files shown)


docs/msw/install.md
```diff
@@ -377,28 +383,28 @@ The full list of the build settings follows:
...
   usable .pdb files with debug information) and this setting makes it
   possible.
 
-* `TARGET_CPU=X64|ARM|ARM64|IA64`
+* `TARGET_CPU=X64|ARM|ARM64`
 
-  (VC++ only.) Set this variable to build for x86_64 systems. If unset, x86
-  build is performed.
+  (MSVC only.) Set this variable to build for the target architecture. If unset,
+  the the default architecture for the active Developer Command Prompt is used.
```
PBfordev, line 397:
```suggestion
  then the default architecture for the active Developer Command Prompt is used.
```

vadz:
Nice, thanks! I'll merge it soon and we should add CI builds using this compiler as soon as it becomes available (I'll subscribe to [this issue](https://github.com/actions/runner-images/issues/13291) and [that one](https://github.com/appveyor/ci/issues/3946) to be notified when it happens).

PBfordev:
@vadz, thanks for the quick merge.

I believe that at least the `wx_config.props` change should be backported, otherwise the solutions open with some ancient (and probably unavailable) toolset version sert and all projects need to be retargeted.

vadz:
I would be fine for backporting all/most of this PR, so if you have time to test it with 3.2 to confirm that it works there and open a PR against 3.2 branch, it would be great, TIA!

PBfordev:
Thomas Krebs wrote on wx-dev
> You can set the toolset version to $(DefaultPlatformToolset), then the wx_config_props.props would not be necessary. 

I wonder if we really (still) need to set all the stuff there, does MSBuild not select a reasonable default when it is not set?

vadz:
See #26165 for the CI job.

@PBfordev Would it be possible to add the solution files for the tests (non-GUI and GUI ones) and wxrc too? TIA!

PR #26134, opened 2026-01-30
wxQt: Fix window modality when using wxWindowDisabler

AliKet:
To see the problem:
- Run the printing sample
- Display the print preview window:  File->Print preview... or Ctrl+V
- The window is incorrectly in a disabled state (see the first image below)

printdemo_bad

and should look like this:

printdemo_good

I'll update the documentation once these changes are approved.

TIA!

### DIFF
--- a/include/wx/utils.h
+++ b/include/wx/utils.h
@@ -681,6 +681,10 @@ public:
     // ctor disables all windows except the given one(s)
     explicit wxWindowDisabler(wxWindow *winToSkip, wxWindow *winToSkip2 = nullptr);
 
+    // ctor disables all windows except the given one. Or disables the parent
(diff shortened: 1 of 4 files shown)


vadz (commented):
Sorry, I don't particularly like adding another ctor to `wxWindowDisabler` as this will make it even more confusing to use (and it's already not great) and I also don't understand why do we need to use this class, whose original purpose was to disable all windows, to disable just the parent?

What is the underlying problem in wxQt that this fixes? I.e. why does disabling the parent doesn't work there? Could we work around this at `wxWindow` level, somehow?

AliKet:
Replaced by #26137

PR #26136, opened 2026-01-30
Use custom configuration name in object folder name also with MSBuild

PBfordev:
When using MSBuild, use the custom configuration name (property wxCfg) in the object files folder name (property wxIntRootDir), just as it is done in the NMake makefile.

Closes #26135.


vadz:
Thanks, I'll merge this soon.

PR #26137, opened 2026-01-31
wxQt: Add support for window modality to TLWs

AliKet:


include/wx/qt/toplevel.h
```diff
@@ -47,6 +47,18 @@ class WXDLLIMPEXP_CORE wxTopLevelWindowQt : public wxTopLevelWindowBase
     virtual void SetWindowStyleFlag( long style ) override;
     virtual long GetWindowStyleFlag() const override;
 
+    // Values of this enum are synchronized with Qt::WindowModality ones
+    enum wxQtWindowModality
```
vadz, line 51: Especially if we make this enum public (in wxTLW), I'd make it a scoped enum called just `Modality` with values `None`, `Window` and `App`.

src/common/prntbase.cpp
```diff
@@ -1831,6 +1835,18 @@ void wxPreviewFrame::InitializeWithModality(wxPreviewFrameModalityKind kind)
             // Nothing to do, we don't need to disable any windows.
             break;
     }
+#else // __WXQT__
+    wxQtWindowModality qtWindowModality;
+
+    if ( m_modalityKind == wxPreviewFrame_AppModal )
```
vadz, line 1841: Minor, but I always prefer using `switch` for enums, and it would be more consistent with the code above.

vadz (commented):
Thanks, I'm ok with merging this, but it looks like this approach points to even better solution: add `wxTLW::SetWindowModality()` (or maybe `ShowWithModality()`?) and implement it different for wxQt (using the code from this PR) and the other ports (using the existing `wxWindowDisabler`-based code from print preview).

Please let me know if you'd like to update this PR to do this or prefer to apply it as is.

Thanks!

AliKet:
> Thanks, I'm ok with merging this, but it looks like this approach points to even better solution: add `wxTLW::SetWindowModality()` (or maybe `ShowWithModality()`?) and implement it different for wxQt (using the code from this PR) and the other ports (using the existing `wxWindowDisabler`-based code from print preview).
> 
> Please let me know if you'd like to update this PR to do this or prefer to apply it as is.
> 

I'am working on this and will push the changes soon...

PR #26139, opened 2026-01-31
BUG FIX for issuse #26138

hwiesmann:
As mentioned in issue #26138 the second parameter to the constructor has to be "item" instead of parent. Only "item" is to be marked as being deleted.


vadz:
Thanks, this looks correct, so I'll merge this (and backport to 3.2).

Calling the item stored in `wxOSXDVCDeleting` `m_parent` seems confusing to me, I think we should rename it to just `m_item`, or am I missing something here?

PR #26140, opened 2026-01-31
Add support for MSVS 2026

PBfordev:
Update wx_config.props, add MSVS 2026 solutions for the library and the minimal sample, and update the docs.


vadz:
Pushed, thanks!

PR #26141, opened 2026-01-31
Generic wxListCtrl: Fix colors when system theme changes

rjricken:
This PR addresses several issues when switching the system theme (e. g. light to dark mode) while using the generic `wxListCtrl` (default on GTK).

1. `wxListHeaderWindow`: Added `EVT_SYS_COLOUR_CHANGED` handler to update foreground/background colors and fixed the sort arrow color, which now uses the window's foreground color to ensure visibility on dark headers.
2. `wxListMainWindow`: Added `EVT_SYS_COLOUR_CHANGED` handler to update standard colors and the selection brushes (`m_highlightBrush`, `m_highlightUnfocusedBrush`).

#### Known Issue / Advice Wanted
While the content and headers now update perfectly, the border of the `wxGenericListCtrl` container does not update its color when the theme changes. It retains the color of the theme active at creation.

I tried many different approaches to force a refresh, but the underlying native GTK border does not seem to react to the system color change event. While I'm a long time user of wxWidgets, this is my first time working in the library internals, so I would appreciate any guidance on how to fix the border issue. Just point me in the right direction and I can give it a try.

### DIFF
--- a/include/wx/generic/private/listctrl.h
+++ b/include/wx/generic/private/listctrl.h
@@ -618,6 +618,8 @@ public:
 
     void OnPaint( wxPaintEvent &event );
 
+    void OnSysColourChanged( wxSysColourChangedEvent &event );
+
     void OnChildFocus(wxChildFocusEvent& event);
 
     void DrawImage( int index, wxDC *dc, int x, int y );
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for the fixes, they look good to me and we can merge this and probably backport it to 3.2 too.

Unfortunately I don't know what's going on with the borders neither.

rjricken:
Sounds good. I can do the 3.2 backport next.
As for the borders, I'll give it another try.

PR #26143, opened 2026-02-01
Document wxGLAttributes::Default no longer using multi-sampling

arch1t3cht:
This was changed in 46857a73ba82bb09fcc6bb42d3d4c886354b9243.

---

Just a small fix I noticed when reading through the docs.


vadz:
Thanks for noticing and fixing this, will merge and push soon.

PR #26144, opened 2026-02-01
Fix broken MSVS 2015 build

PBfordev:
The ancient MSVS 2015 is used to build the official binaries. 

I suppose no CI service used supports it anymore...


vadz:
Thanks, I'll commit this but we do have a MSVS 2015 CI job on AppVeyor, so this must be due to not using a recent enough SDK version.

PBfordev:
> Thanks, I'll commit this but we do have a MSVS 2015 CI job on AppVeyor, so this must be due to not using a recent enough SDK version.

I freshly installed MSVS Express 2015 for Windows Desktop (v14.0.25431.01 Update 3), using Windows SDK v8.1. I don't think this version can update SDKs on its own...

vadz:
Is it not possible to install standalone SDK package? Anyhow, AppVeyor somehow manages to use newer SDK, apparently...

PBfordev:
I see. Since I have no idea how to fix the mixed EOL issue; and more importantly, if there are people still actually using MSVS 2015 they probably do not use the very outdated SDK with it, so I guess it would be for the best to close the Issue. Sorry for the noise.

vadz:
I've already fixed the EOLs (you must have edited the file which was in Unix format under Windows?) and applied this locally and will push it, it does no harm, and if it can help some poor soul still stuck on MSVS 2015, why not.

Issue #9215, opened 2003-05-29
Add `wxGLCanvas::GetProcAddress()`


wxtrac:
*Issue migrated from trac ticket # 9215*

**priority:** normal

#### 2003-05-29 00:55:09: anonymous created the issue

___
Hi!

Would it be possible to implement a platform
independent function for GetProcAddress in the wxGLCanvas?
To replace: wglGetProcAddress() on Windows and
glxGetProcAddressARB() in Linux.

Or is it already made? HELP please!

-- 
Thank you,
Balázs

dsa-t:
Would be great to have `GetProcAddress` in `wxGLCanvas`, similar to Qt: https://doc.qt.io/archives/qt-5.15/qopenglcontext.html#getProcAddress

vadz:
@dsa-t Do you have an example of how would it be used that I could, perhaps, incorporate in a sample and use for testing it myself? I.e. what would be the function that you could load like this without checking for the current backend first in any case?

dsa-t:
For example, it should make it easier to initialize GLAD (like https://gitlab.kitware.com/vtk/vtk/-/blob/04661f4bcdf84195a75e4e91a542b9e6e8bd8e37/Rendering/OpenGL2/vtkOpenGLRenderWindow.cxx#L788) without having to load the functions separately.

I don't have a wx sample yet.

vadz:
According to [SDL wiki](https://wiki.libsdl.org/SDL3/SDL_GL_GetProcAddress) this function can't be static in `wxGLContext` under Windows as the available extensions are context-specific. Is it ok for your planned use of it or should we make it static and let people decide themselves if they want to follow the documentation and make a context current before calling it or not?

dsa-t:
Non-static is fine for me.

vadz:
Now I wonder if this should be in `wxGLContext` or in `wxGLCanvas`. I'd say that the former makes more sense, but we already have a clearly related `IsExtensionSupported()` in the latter (albeit as static), so maybe we should put this there too for consistency? But then it would need to take `wxGLContext` as argument, which is probably not what people expect?

Sorry for all the questions, but it's genuinely difficult to decide on the API when I still don't really know how it's going to be used.

dsa-t:
I think it's fine to implement in `wxGLContext` then.

PR #26146, opened 2026-02-01
Enable building wxGTK without GLX

### LINKED ISSUE #26145: wxGTK fails to build if X11/GLX is not available
### Description

Building master for GTK fails when EGL/GDK_Wayland is available but GLX/GDK_X11 is not.

 Seen with 7ba5acc72c22d5cf36f33ac927b4b11f0ffa477e, didn't happen with ace3d4d936115f4b34a0d8c240a29f5cd7c23743.
 Presumably due to d9df7db130ac.

#### Bug description:

```
src/unix/glcanvas.cpp:235:13: error: ‘wxGLBackendX11’ has not been declared; did you mean ‘wxGLBackend’?
  235 |     return &wxGLBackendX11::Get();
      |             ^~~~~~~~~~~~~~
      |             wxGLBackend

...

src/gtk/glcanvas.cpp:52:15: error: ‘XVisualInfo’ does not name a type
   52 |         const XVisualInfo* xvi = static_cast<XVisualInfo*>(win->GetXVisualInfo());
      |               ^~~~~~~~~~~
src/gtk/glcanvas.cpp:54:13: error: ‘GDK_VISUAL_XVISUAL’ was not declared in this scope; did you mean ‘GDK_IS_VISUAL’?
   54 |         if (GDK_VISUAL_XVISUAL(visual)->visualid != xvi->visualid)
      |             ^~~~~~~~~~~~~~~~~~
      |             GDK_IS_VISUAL
src/gtk/glcanvas.cpp:54:53: error: ‘xvi’ was not declared in this scope
   54 |         if (GDK_VISUAL_XVISUAL(visual)->visualid != xvi->visualid)
      |                                                     ^~~
src/gtk/glcanvas.cpp:57:22: error: ‘gdk_x11_screen_lookup_visual’ was not declared in this scope; did you mean ‘gdk_screen_list_visuals’?
   57 |             visual = gdk_x11_screen_lookup_visual(screen, xvi->visualid);
      |                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                      gdk_screen_list_visuals
 ```

#### To Reproduce:

```
-DCMAKE_INSTALL_PREFIX=/usr -DwxUSE_SYS_LIBS=ON -DwxBUILD_COMPATIBILITY=3.2 -DwxUSE_XRC=ON -DwxUSE_LIBLZMA=yes -DwxUSE_SECRETSTORE=ON -DwxUSE_WEBREQUEST=no -DwxBUILD_PRECOMP=OFF -DwxUSE_GLCANVAS_EGL=ON -DwxUSE_GRAPHICS_CONTEXT=ON -DwxUSE_GTKPRINT=ON -DwxUSE_GUI=ON -DwxDEFAULT_TOOLKIT=gtk3 -DwxUSE_LIBPNG=sys -DwxUSE_LIBJPEG=sys -DwxUSE_LIBMSPACK=ON -DwxUSE_LIBWEBP=sys -DwxUSE_NANOSVG=sys -DwxUSE_LIBGNOMEVFS=OFF -DwxUSE_LIBSDL=yes -DwxUSE_MEDIACTRL=yes -DwxUSE_WEBVIEW=yes -DwxUSE_LIBNOTIFY=yes -DwxUSE_OPENGL=yes -DwxUSE_SPELLCHECK=yes -DwxUSE_WAYLAND=yes -DwxBUILD_TESTS=OFF -DwxUSE_UIACTIONSIMULATOR=no -DwxUSE_XTEST=no -DCMAKE_BUILD_TYPE=RelWithDebInfo
```

- wxWidgets version: master
- wxWidgets port wxGTK
- OS Gentoo Linux
  
  + GTK version: 3.24.51
  + GDK Wayland backend is used
  + Desktop environment GNOME

### PR DESCRIPTION
**Pull request by cjmayo**
Set wxHAS_GLX if GTK 3 with GDK_X11.

---

I've added non-configurable support to configure in a separate commit due to my limited autoconf skills - but this at least maintains the current situation for configure.

Closes #26145

### CODE REVIEW

**Comment by cjmayo** [comment]
What to do about wxGTK 2 - is it reasonable to set wxHAS_GLX?

**Review by vadz — COMMENTED** [review]
Thanks for the fix, I have indeed completely forgotten that GLX might not be available either.

I'm not sure why does the PR only define `wxHAS_GLX` for wxGTK3, surely it should do it if OpenGL headers/libraries are found, for any port (and then stop doing it for wxX11 in the header)?

**Comment by vadz** [comment]
> What to do about wxGTK 2 - is it reasonable to set wxHAS_GLX?

Yes, and for wxX11 too, please. I.e. just don't test for the toolkit at all — if OpenGL headers/libraries are available, define this symbol.

**Comment by cjmayo** [comment]
OK, so this v2:

- wxHAS_GLX only defined in the build system
- toolkit.cmake is already setting X11_FOUND so use that and check for GLX
- wxHAS_GLX always defined in configure if OpenGL is found
- One other change in the source - a bit ugly, but needed to avoid a compiler warning `control reaches end of non-void function [-Wreturn-type]` and could be seen as a slight optimisation for EGL only.

```diff
--- a/src/unix/glcanvas.cpp
+++ b/src/unix/glcanvas.cpp
@@ -231,7 +231,9 @@ wxGLBackend* wxGLBackend::Init()
         return &wxGLBackendEGL::Get();
 #endif // GTK 3 with Wayland
 
+#ifdef wxHAS_GLX
     if ( !(wxGLBackendPreferGLX || wxSystemOptions::IsFalse("opengl.egl")) )
+#endif // wxHAS_GLX
         return &wxGLBackendEGL::Get();
 #endif // wxHAS_EG
```

vadz (commented):
Thanks, this looks good to me, but I think we shouldn't compile the code for defining and setting `wxGLBackendPreferGLX` too when GLX support is off.

And maybe dealing with 3 cases:

1. Both EGL and GLX are available.
2. Only EGL is.
3. Only GLX is.

in `wxGLBackend::Init()` could make its code slightly more legible?

Also, @MaartenBent could you please check the CMake part? I think it's ok to use `OpenGL_GLX_FOUND` because we already use `OpenGL_EGL_FOUND` but both [are documented](https://cmake.org/cmake/help/latest/module/FindOpenGL.html) to be in CMake 3.10+ only, so maybe this means we need to bump the minimum CMake version to this?

MaartenBent:
Haven't tried it, but looks fine. No need to bump the minimum CMake version, `OpenGL_GLX_FOUND` will just evaluate to `OFF` with older versions.

cjmayo:
> Thanks, this looks good to me, but I think we shouldn't compile the code for defining and setting `wxGLBackendPreferGLX` too when GLX support is off.

Added a commit for this

> And maybe dealing with 3 cases:
> 
>     1. Both EGL and GLX are available.
> 
>     2. Only EGL is.
> 
>     3. Only GLX is.
> 
> 
> in `wxGLBackend::Init()` could make its code slightly more legible?

I've added a commit to improve readability a bit. The problem I can see in going further is that
```c++
#if defined(__WXGTK3__) && defined(GDK_WINDOWING_WAYLAND)
```
is need for both 1. and 2.. To me duplicating that would make it more work to maintain - but just an opinion.

vadz:
> Haven't tried it, but looks fine. No need to bump the minimum CMake version, `OpenGL_GLX_FOUND` will just evaluate to `OFF` with older versions.

Sorry for being unclear, but this is precisely what I was worried about: this means the test would silently work incorrectly with old CMake. IMO it would be better to require a newer one rather than this.

vadz:
> I've added a commit to improve readability a bit.

Thanks! I agree that this code is not trivial to untangle, so I'll just merge this for now and maybe return to it later.

Thanks again for fixing this!

MaartenBent:
> Sorry for being unclear, but this is precisely what I was worried about: this means the test would silently work incorrectly with old CMake. IMO it would be better to require a newer one rather than this.

You could add a CMake version check, so it will have the old behavior when old CMake is used.
```diff
-            if(X11_FOUND AND OpenGL_GLX_FOUND)
+            if(X11_FOUND AND (OpenGL_GLX_FOUND OR CMAKE_VERSION VERSION_LESS "3.10"))
```

vadz:
> You could add a CMake version check, so it will have the old behavior when old CMake is used.

But it would still be wrong, i.e. result in build errors if GLX is not available. We'd need to write our own test for it to get this right and while this is definitely possible, is it really worth doing? It looks like CMake 3.10 was released pretty soon after 3.5 (I can't find the exact dates, their [release page](https://github.com/kitware/cmake/releases?page=27) seems to say that 3.10 was released _before_ 3.5, in November and December 2018 respectively, which doesn't make sense), so would it be a problem to just start requiring 3.10 instead?

MaartenBent:
Requiring 3.10 as a minimum should be fine. It was released [November 2017](https://www.kitware.com/cmake-3-10-0-available-for-download/), 3.5 was released [March 2016](https://www.kitware.com/cmake-3-5-0-available-for-download/). The github tags were probably pushed later.

vadz:
Thanks, I'll do this then. Looking at `CMAKE_VERSION` checks, I see that we could remove all of them if we required 3.18, which is from 2020. I won't do this for now, but perhaps for 3.4.0?

MaartenBent:
> I won't do this for now, but perhaps for 3.4.0?

Definitely.
Maybe even newer, depending on what distros it will support and which CMake version comes with it.

dsa-t:
Now that the user app doesn't have to care about GLX/EGL as much, it would be great to have `GetProcAddress` (https://github.com/wxWidgets/wxWidgets/issues/9215)

PR #26147, opened 2026-02-02
Add window modality support to wxFrame

AliKet:
`wxDialog` already support modality:

https://github.com/wxWidgets/wxWidgets/blob/7ba5acc72c22d5cf36f33ac927b4b11f0ffa477e/include/wx/dialog.h#L54-L79

Therefore, the changes here are for `wxFrame` class.

If the API looks good, then I'll  update the documentation.

TIA!

### DIFF
--- a/include/wx/frame.h
+++ b/include/wx/frame.h
@@ -74,6 +74,18 @@ public:
     // if the frame has a toolbar) in client coordinates
     virtual wxPoint GetClientAreaOrigin() const override;
 
+    // frame modality
+    // ------------------
+
+    enum class Modality
+    {
+        App,
+        Window,
+        None
+    };
+
+    // Changing the frame modality after it has been shown has no effect
+    virtual void SetWindowModality(Modality modality);
 
     // menu bar functions
     // ------------------
@@ -248,6 +260,9 @@ protected:
 
     wxToolBar *m_frameToolBar = nullptr;
 #endif // wxUSE_TOOLBAR
+
+    // The frame is not modal by default.
+    Modality m_modality = Modality::None;
 
 #if wxUSE_MENUS
     wxDECLARE_EVENT_TABLE();
--- a/include/wx/frame.h
+++ b/include/wx/frame.h
@@ -214,7 +214,7 @@ protected:
     static bool ShouldUpdateMenuFromIdle();
 
 #if wxUSE_MENUBAR
-    wxMenuBar *m_frameMenuBar;
+    wxMenuBar *m_frameMenuBar = nullptr;
 #endif // wxUSE_MENUBAR
 
 #if wxUSE_STATUSBAR && (wxUSE_MENUS || wxUSE_TOOLBAR)
@@ -235,18 +235,18 @@ protected:
     // return false if there is no help for such item
     bool ShowMenuHelp(int helpid);
 
-    wxStatusBar *m_frameStatusBar;
+    wxStatusBar *m_frameStatusBar = nullptr;
 #endif // wxUSE_STATUSBAR
 
 
-    int m_statusBarPane;
+    int m_statusBarPane = 0;
 
 #if wxUSE_TOOLBAR
     // override to update status bar position (or anything else) when
     // something changes
     virtual void PositionToolBar() { }
 
-    wxToolBar *m_frameToolBar;
+    wxToolBar *m_frameToolBar = nullptr;
 #endif // wxUSE_TOOLBAR
 
 #if wxUSE_MENUS
--- a/include/wx/prntbase.h
+++ b/include/wx/prntbase.h
@@ -456,10 +456,6 @@ protected:
     wxPreviewCanvas*      m_previewCanvas = nullptr;
     wxPreviewControlBar*  m_controlBar = nullptr;
     wxPrintPreviewBase*   m_printPreview;
-    wxWindowDisabler*     m_windowDisabler = nullptr;
-
-    wxPreviewFrameModalityKind m_modalityKind = wxPreviewFrame_NonModal;
-
 
 private:
     void OnChar(wxKeyEvent& event);
--- a/include/wx/prntbase.h
+++ b/include/wx/prntbase.h
@@ -54,18 +54,23 @@ enum wxPrinterError
     wxPRINTER_ERROR
 };
 
+// The following symbols are preserved only for compatibility.
+// Please use the new wxFrame::Modality instead in any new code.
+
 // Preview frame modality kind used with wxPreviewFrame::Initialize()
-enum wxPreviewFrameModalityKind
-{
-    // Disable all the other top level windows while the preview is shown.
-    wxPreviewFrame_AppModal,
+using wxPreviewFrameModalityKind = wxFrameBase::Modality;
 
-    // Disable only the parent window while the preview is shown.
-    wxPreviewFrame_WindowModal,
+// Disable all the other top level windows while the preview is shown.
+constexpr wxPreviewFrameModalityKind
+wxPreviewFrame_AppModal = wxFrameBase::Modality::App;
 
-    // Don't disable any windows.
-    wxPreviewFrame_NonModal
-};
+// Disable only the parent window while the preview is shown.
+constexpr wxPreviewFrameModalityKind
+wxPreviewFrame_WindowModal = wxFrameBase::Modality::Window;
+
+// Don't disable any windows.
+constexpr wxPreviewFrameModalityKind
+wxPreviewFrame_NonModal = wxFrameBase::Modality::None;
 
 //----------------------------------------------------------------------------
 // wxPrintFactory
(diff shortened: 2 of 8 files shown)


include/wx/frame.h
```diff
@@ -74,6 +74,26 @@ class WXDLLIMPEXP_CORE wxFrameBase : public wxTopLevelWindow
...
+
+    // Changing the frame modality after it has been shown has no effect
+    virtual void SetWindowModality(ModalityKind modalityKind);
+
+    // Show the frame modally if modalityKind is App or Window. Otherwise
+    // show the frame normally.
+    void ShowWithModality(ModalityKind modalityKind = ModalityKind::App)
+    {
+        SetWindowModality(modalityKind);
+        Show();
```
vadz: I'm not sure if this function can always be implemented like this, e.g. I can imagine that on some platforms the only way to set window modality would be to actually show it in this way.

So I'd rather remove `SetWindowModality()` and make this function virtual instead — if the derived class needs to implement it by doing something and then calling `Show()` it wouldn't be a problem.

include/wx/frame.h
```diff
@@ -235,20 +255,23 @@ class WXDLLIMPEXP_CORE wxFrameBase : public wxTopLevelWindow
...
 #if wxUSE_TOOLBAR
     // override to update status bar position (or anything else) when
     // something changes
     virtual void PositionToolBar() { }
 
-    wxToolBar *m_frameToolBar;
+    wxToolBar *m_frameToolBar = nullptr;
 #endif // wxUSE_TOOLBAR
 
+    // The frame is modaless by default.
```
vadz:
```suggestion
    // The frame is modeless by default.
```

(or "not modal" if you prefer)

include/wx/prntbase.h
```diff
@@ -55,17 +55,19 @@ enum wxPrinterError
 };
 
 // Preview frame modality kind used with wxPreviewFrame::Initialize()
```
vadz, line 60: I'd mention that these symbols are preserved for compatibility but that new `wxFrame::Modality` ones should be used instead in the new code.

include/wx/frame.h
```diff
@@ -74,6 +74,26 @@ class WXDLLIMPEXP_CORE wxFrameBase : public wxTopLevelWindow
     // if the frame has a toolbar) in client coordinates
     virtual wxPoint GetClientAreaOrigin() const override;
 
+    // frame modality
+    // ------------------
+
+    enum class ModalityKind
```
vadz: Maybe just

```suggestion
    enum class Modality
```

? "Kind" doesn't seem to add anything useful and this is shorter.

samples/printing/printing.cpp
```diff
@@ -487,7 +487,7 @@ void MyFrame::OnAngleDown(wxCommandEvent& WXUNUSED(event))
 void MyFrame::OnPreviewFrameModalityKind(wxCommandEvent& event)
 {
     m_previewModality = static_cast<wxPreviewFrameModalityKind>(
-                            wxPreviewFrame_AppModal +
+                            (int)wxPreviewFrame_AppModal +
```
vadz:
```suggestion
                            static_cast<int>(wxPreviewFrame_AppModal) +
```

because it's weird to mix C++ and C style casts in the same expression.

FWIW ideal would be to use `std::to_underlying()`, but this is C++23. Maybe we should provide our own `wxToUnderlying()` helper to be able to use it for now...

src/common/framecmn.cpp
```diff
@@ -48,6 +51,17 @@ wxBEGIN_EVENT_TABLE(wxFrameBase, wxTopLevelWindow)
...
 
+// ----------------------------------------------------------------------------
+// globals
+// ----------------------------------------------------------------------------
+namespace
+{
+// Global stack used to track all active wxWindowDisablers for the wxFrames
+// currently shown modally (those with ModalityKind::App as modality kind).
+// E.g.: a frame shown modally from an other modal frame.
+wxStack<wxWindowDisabler*> gs_windowDisablers;
```
vadz: I'd use `std::stack`, there is no reason to use `wxStack` in the new code.

More importantly, why do we use pointers here, couldn't we just store the objects themselves? `wxWindowDisabler` is not copyable but it could be made movable with minimal effort.

And if we really need pointers, I'd at all use `std::unique_ptr<wxWindowDisabler>` for the element type instead of manually deleting raw pointers.

src/common/framecmn.cpp
```diff
@@ -48,6 +51,17 @@ wxBEGIN_EVENT_TABLE(wxFrameBase, wxTopLevelWindow)
...
 wxEND_EVENT_TABLE()
 
+// ----------------------------------------------------------------------------
+// globals
+// ----------------------------------------------------------------------------
+namespace
+{
+// Global stack used to track all active wxWindowDisablers for the wxFrames
+// currently shown modally (those with ModalityKind::App as modality kind).
+// E.g.: a frame shown modally from an other modal frame.
```
vadz: Very minor:
```suggestion
// E.g.: a frame shown modally from another modal frame.
```

src/common/framecmn.cpp
```diff
@@ -142,25 +156,29 @@ wxCONSTRUCTOR_6( wxFrame, wxWindow*, Parent, wxWindowID, Id, wxString, Title, \
...
+
+wxFrameBase::~wxFrameBase()
 {
-#if wxUSE_MENUBAR
-    m_frameMenuBar = nullptr;
-#endif // wxUSE_MENUS
+    switch ( m_modalityKind )
+    {
+        case ModalityKind::App:
+            if ( !gs_windowDisablers.empty() )
```
vadz: Should we assert that it's not empty perhaps? It looks like it shouldn't be, if there was an app-modal frame before...

vadz (commented):
Thanks, this looks good but I had a number of minor comments, please let me know if I'm missing anything.

include/wx/frame.h
AliKet: I just removed this function entirely for now and let add it when we really need it !

include/wx/utils.h
```diff
@@ -697,6 +710,9 @@ class WXDLLIMPEXP_CORE wxWindowDisabler
     wxVector<wxWindow*> m_windowsToSkip;
     bool m_disabled;
 
+    // move assignment deleted as it doesn't make sense for this class
+    wxWindowDisabler& operator=(wxWindowDisabler&& other) noexcept = delete;
```
vadz: Very minor, but IMHO specifying `noexcept` for a deleted function doesn't make much sense, so


```suggestion
    wxWindowDisabler& operator=(wxWindowDisabler&& other) = delete;
```

src/common/framecmn.cpp
```diff
@@ -142,25 +158,29 @@ wxCONSTRUCTOR_6( wxFrame, wxWindow*, Parent, wxWindowID, Id, wxString, Title, \
...
-#endif // wxUSE_STATUSBAR
-
-    m_statusBarPane = 0;
-}
+wxFrameBase::wxFrameBase() = default;
 
 wxFrameBase::~wxFrameBase()
 {
+#ifndef __WXQT__
+    switch ( m_modality )
```
vadz: I wonder if we shouldn't do this when the frame is hidden, rather than when it's destroyed. As TLWs are destroyed during the next event loop iteration, this means that the other windows may remain disabled for longer than could be expected, which could be problematic under MSW (it could break focus, Z-order, ...).

If doing it when hiding the frame doesn't create any problems, I'd rather do it then.

src/common/framecmn.cpp
```diff
@@ -142,25 +158,29 @@ wxCONSTRUCTOR_6( wxFrame, wxWindow*, Parent, wxWindowID, Id, wxString, Title, \
...
+    {
+        case Modality::App:
+            wxASSERT_MSG( !gs_windowDisablers.empty(),
+                "A window with Modality::App MUST have a wxWindowDisabler" );
+            gs_windowDisablers.pop();
```
vadz: We're going to have a crash after this assert if it ever triggers, so let's make this safer:

```suggestion
            if ( !gs_windowDisablers.empty() )
            {
                gs_windowDisablers.pop();
            }
            else
            {
                wxFAIL_MSG( "A window with Modality::App MUST have a wxWindowDisabler" );
            }
```

src/common/framecmn.cpp
```diff
@@ -265,6 +285,42 @@ void wxFrameBase::RemoveChild(wxWindowBase *child)
...
 }
 
+void wxFrameBase::SetWindowModality(Modality modality)
+{
+    wxCHECK_RET( !IsShown(),
+                 "SetWindowModality() must be called before showing the window" );
+
+    m_modality = modality;
+
+#ifndef __WXQT__
```
vadz: Could we avoid this `#ifndef` if we just didn't call the base class function from wxQt? It seems `m_modality` is not really used in wxQt.

vadz (commented):
Thanks, this looks really good now, but I just wonder if we shouldn't re-enable the other windows sooner.

vadz (approved):
Thanks, will merge this soon.

PR #26151, opened 2026-02-02
OSX fixing file dialog with only one filter

csomor:
see #26148

### DIFF
--- a/src/osx/cocoa/filedlg.mm
+++ b/src/osx/cocoa/filedlg.mm
@@ -287,6 +287,10 @@ void wxFileDialog::ShowWindowModal()
     {
       m_firstFileTypeFilter = GetMatchingFilterExtension(m_fileName);
     }
+    else
+    {
+        m_firstFileTypeFilter = 0;
+    }
 
     if ( HasFlag(wxFD_SAVE) )
     {
@@ -606,6 +610,10 @@ int wxFileDialog::ShowModal()
     else if ( m_useFileTypeFilter )
     {
         m_firstFileTypeFilter = GetMatchingFilterExtension(m_fileName);
+    }
+    else
+    {
+        m_firstFileTypeFilter = 0;
     }
 
     OSXBeginModalDialog();
--- a/src/osx/cocoa/filedlg.mm
+++ b/src/osx/cocoa/filedlg.mm
@@ -247,10 +247,6 @@ NSArray* GetTypesFromFilter( const wxString& filter, wxArrayString& names, wxArr
 
 void wxFileDialog::ShowWindowModal()
 {
-    wxCFStringRef cf( m_message );
-    wxCFStringRef dir( m_dir );
-    wxCFStringRef file( m_fileName );
-
     wxNonOwnedWindow* parentWindow = nullptr;
 
     m_modality = wxDIALOG_MODALITY_WINDOW_MODAL;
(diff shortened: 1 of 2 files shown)


src/osx/cocoa/filedlg.mm
```diff
@@ -287,6 +287,10 @@ - (void)setAllowedExtensions:(const wxArrayString &)extensions
     {
       m_firstFileTypeFilter = GetMatchingFilterExtension(m_fileName);
     }
+    else
+    {
+        m_firstFileTypeFilter = 0;
```
vadz: Shouldn't we just initialize it to 0 above instead? This would be more clear IMHO.

Note that `GetMatchingFilterExtension()` already returns 0 instead of -1 if there is no match.

vadz: If you agree, I can just make this change myself and push it to both branches.

src/osx/cocoa/filedlg.mm
```diff
@@ -607,6 +611,10 @@ - (void)setAllowedExtensions:(const wxArrayString &)extensions
     {
         m_firstFileTypeFilter = GetMatchingFilterExtension(m_fileName);
     }
+    else
```
vadz: It would also be great to refactor this code, at least in master, to avoid having big chunks of same code in both `wxFileDialog::ShowWindowModal()` and `ShowModal()`...

src/osx/cocoa/filedlg.mm
csomor: yes that would be fine, thanks

vadz:
@csomor Sorry, just noticed that after this change this code

https://github.com/wxWidgets/wxWidgets/blob/7ba5acc72c22d5cf36f33ac927b4b11f0ffa477e/src/osx/cocoa/filedlg.mm#L685-L689

doesn't make sense any more as `m_firstFileTypeFilter` is always positive or zero.

Does it mean that we should set it to 0 only when saving? AFAICS this test is more or less equivalent to `m_useFileTypeFilter` but I don't fully understand what's going on here :-(

csomor:
I'll clean up

vadz (approved):
Thanks a lot!

If this is the final version, I'll merge it and backport to 3.2 (adding non-virtual private member function should be ok even there).

csomor:
> Thanks a lot!
> 
> If this is the final version, I'll merge it and backport to 3.2 (adding non-virtual private member function should be ok even there).

Yes, I think it's done. You're very welcome - the code duplication irked me in the past as well - so it was a good moment ... and in a way to make sure we can backport it.

vadz:
I've pushed this as 7c6300acfb21b4105650576ff8a16c18020e8f4f and 17dbee8a464685122fcd0ed0fd105ed2bac4062e and then added 5b96b76a791118beff590d2fc1a7e23d201ff43e to try to clarify things a bit. As I wrote there, it's still not clear to me what does setting `m_filterExtensions` to `-1` achieve, but at least the code works without asserts/crashes now.

If you don't see any problems with what I did, I'll backport at least 17dbee8a464685122fcd0ed0fd105ed2bac4062e to 3.2 too.

Thanks again!

nobugshere:
I'm not sure if it's related to these changes, but I tested it from current master and the save files dialog does not give me a default file name extension nor the drop-down list to select from the options.  (dialogs sample, Dialogs->File Operations->Save file) It works as expected in the 3.2 branch.
macOS 15.7.3/Xcode 16.4

csomor:
@nobugshere thanks I'll test

csomor:
reproduced, I'll add a new PR

PR #26153, opened 2026-02-02
Fix using webview configurations for selecting IE backend

vadz:
See #26152.


dsa-t:
Looks good for me, thanks!

PR #26156, opened 2026-02-03
WebView Edge: ensure keyboard focus is inside the  webview document if the WebView window gets focus before MS Webview2 is fully initialized

glenrgordon:
If focus moved to the wxWebView window before the Edge instance was
fully created, we couldn't call m_webViewController->MoveFocus at that
time. If focus is on the wxWebview at the time we get the
wxWebViewEdgeImpl::OnWebViewCreated callback, we now call MoveFocus to
ensure that focus is inside of the web content window. This helps
screen readers, since they use the focus window to influence their
behavior.

Fixes https://github.com/wxWidgets/wxWidgets/issues/26155

### DIFF
--- a/src/msw/webview_edge.cpp
+++ b/src/msw/webview_edge.cpp
@@ -856,6 +856,14 @@ HRESULT wxWebViewEdgeImpl::OnWebViewCreated(HRESULT result, ICoreWebView2Control
     m_webViewController->put_IsVisible(true);
     m_ctrl->NotifyWebViewCreated();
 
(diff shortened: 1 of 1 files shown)


vadz:

trypsynth:
@glenrgordon Just wanted to personally thank you for this. This has been a bug since I started using wxWidgets, and I just worked around it with a super super hacky UIActionSimulator. This is much better though. Godly work.

glenrgordon:
Thanks so much for the positive feedback!

PR #26157, opened 2026-02-03
OSX add exception rethrowing

csomor:
see #26150

### DIFF
--- a/src/osx/cocoa/evtloop.mm
+++ b/src/osx/cocoa/evtloop.mm
@@ -341,6 +341,13 @@ void wxGUIEventLoop::OSXDoRun()
         }
     }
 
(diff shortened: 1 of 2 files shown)


csomor:
Right now `ProcessPendingEvents` is called from `wxCFEventLoop::CommonModeObserverCallBack` but I don't think this would be a good place to rethrow, I'd rather do this from an event loop iteration, but in `wxGUIEventLoop::OSXDoRun` the main loop `[NSApp run]` does not return until the app exits, so right now only then the exception gets rethrown and eventually logged to stderr.

vadz:
Sorry, I still didn't have time to actually test this, but the code changes look fine, so I'll just merge it, especially because the same changes seem to have been independently redone in #26174.

PR #26158, opened 2026-02-04
wxPropertyGridInterface::SetPropertyImage incorrectly documented

mcorino:
BitmapBundle argument is implemented as const


vadz:

PR #26166, opened 2026-02-06
Some wxQt fixes

AliKet:
[Commit](e37bb7987338b604a0b56687599698b664547432) fixes the problem shown below with the print preview window :

https://github.com/user-attachments/assets/9f5de19d-729e-49d6-b0d4-3309353bebf6

### DIFF
--- a/src/qt/window.cpp
+++ b/src/qt/window.cpp
@@ -653,6 +653,18 @@ void wxWindowQt::Refresh( bool WXUNUSED( eraseBackground ), const wxRect *rect )
     }
 }
 
+void wxWindowQt::ClearBackground()
+{
+    if ( !GetHandle()->autoFillBackground() )
+    {
+        // Rely on Qt to do the right thing with clearing the background.
+        GetHandle()->setAutoFillBackground(true);
+        GetHandle()->setAutoFillBackground(false);
+    }
+    // else: No need to do anything because Qt will fill the background
+    //       of the widget before invoking the paint event anyhow.
+}
+
 bool wxWindowQt::SetCursor( const wxCursor &cursor )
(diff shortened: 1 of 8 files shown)


AliKet:
@vadz @paulcor if the guard here 9e2561fddc981d12305a6371f7dfcb96b9fbbef7 is intentional, then I suggest moving the code in the base class to wxMSW and leave the base class version empty. What do you think ?

vadz:
Yes, I think this code could be moved into wxMSW. But then I also think we should stop using it in the samples and document it as being obsolete/useless.

src/qt/window.cpp
```diff
@@ -653,6 +653,17 @@ void wxWindowQt::Refresh( bool WXUNUSED( eraseBackground ), const wxRect *rect )
     }
 }
 
+void wxWindowQt::ClearBackground()
+{
+    if ( !GetHandle()->autoFillBackground() )
+    {
+        GetHandle()->setAutoFillBackground(true);
+        GetHandle()->setAutoFillBackground(false);
+    }
```
vadz, line 663: I feel like this code would merit some explanation as it looks like it's not doing anything, yet it probably does?

vadz (commented):
Thanks, this could be already merged but I'd rather exclude 9e2561fddc981d12305a6371f7dfcb96b9fbbef7 because it's not related to wxQt and doesn't really change anything -- if we want to change something concerning this function, then I'd prefer to do what you proposed in your comment.

AliKet:
@vadz  No further commits from my side. This PR is ready for another round of review.

TIA!

vadz (approved):

PR #26167, opened 2026-02-06
Add MSVS 2026 solutions for tests and wxrc

PBfordev:
BTW, I have realized that tests and wxrc suffer from the same problem the samples did until late 2024 (see eba1064): The executables in the DLL builds cannot be run from the IDE because wxWidgets DLL are not found.

Not sure if this should be fixed as well for the tests (in another PR), perhaps running tests from the IDE is not as important as when playing with samples...

### DIFF
--- a/tests/test_gui_vc18.slnx
+++ b/tests/test_gui_vc18.slnx
@@ -1,0 +1,38 @@
+<Solution>
+  <Configurations>
+    <BuildType Name="Debug" />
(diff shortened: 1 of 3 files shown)


vadz:
Thanks, I've integrated this in #26165.

And yes, it would be nice to allow running tests without any extra manual configuration in the DLL builds, so if you could please make the equivalent of the samples commit for them too, it would be welcome. TIA!

PR #26171, opened 2026-02-07
Allow running tests from MSVS in DLL configurations

PBfordev:
Create file tests/tests.props which adds the DLL build directory to MSVS PATH in <LocalDebuggerEnvironment> property and add the file to test.vcxproj and test_gui.vcxproj MSVC projects.

See also eba1064.


vadz:

Issue #26172, opened 2026-02-08
`wxApp::OnExit()` may be called before all windows are destroyed


solemnwarning:
#### Bug description:

I'm not sure how this is getting triggered in the wild, but crash dumps I have on macOS are suggesting that under some condition, my window destructors are running after my app's `OnExit()` method has been called. I can trigger this on both Linux and macOS by calling `wxGetApp().ExitMainLoop()` directly from a command handler, which crashes my application because the window destructor tries accessing objects destroyed by my `OnExit()` method.

#### Expected vs observed behaviour:

The manual for `wxAppConsole::OnExit()` states the following:

> OnExit is called after destroying all application windows and controls, but before wxWidgets cleanup. Note that it is not called at all if [OnInit()](https://docs.wxwidgets.org/3.2/classwx_app_console.html#a99953775a2fd83fa2456e390779afe15) failed.

The above is not true if `wxApp::ExitMainLoop()` is called manually, or seemingly, under some unknown condition on macOS.

### Platform and version information

- wxWidgets version  you use: 3.2.8, 3.2.8.1
- wxWidgets port  you use: wxGTK, wxOSX

solemnwarning:
I'm not sure if these backtraces will be of much help, but this is the call stack when the condition occurs due to a call to `wxApp::ExitMainLoop()`:

```
#0  0x00005555559792aa in REHex::MainWindow::~MainWindow (this=0x5555563b8930, __in_chrg=<optimized out>) at src/mainwindow.cpp:676
#1  0x000055555597937e in REHex::MainWindow::~MainWindow (this=0x5555563b8930, __in_chrg=<optimized out>) at src/mainwindow.cpp:678
#2  0x00007ffff746afd6 in wxAppBase::DeleteAllTLWs() () at /lib/x86_64-linux-gnu/libwx_gtk3u_core-3.2.so.0
#3  0x00007ffff746b0a8 in wxAppBase::CleanUp() () at /lib/x86_64-linux-gnu/libwx_gtk3u_core-3.2.so.0
#4  0x00007ffff7b39522 in wxEntryCleanup() () at /lib/x86_64-linux-gnu/libwx_baseu-3.2.so.0
#5  0x00007ffff7b3a04f in wxEntry(int&, wchar_t**) () at /lib/x86_64-linux-gnu/libwx_baseu-3.2.so.0
[...]

PR #26173, opened 2026-02-08
Document that wxApp::OnExit() should be called from overrides.

solemnwarning:
This was found while digging into #26172, but does not fix the issue of wxApp::OnExit() being called before windows are destroyed.


vadz:
Thanks, looks good, will merge soon.

PR #26174, opened 2026-02-08
Fix over-zealous exception catching on macOS.

solemnwarning:
I think this should have been done as part of ec12651b9b1, without it any exceptions thrown within the main loop when the exception handler is disabled appear to come from wxCFEventLoop::DoRun(), unfortunately this still doesn't get us all the way to the real exception source in a debugger because CFRunLoopInMode() installs a handler too, but this gets us one step closer at least.


vadz:
This seems to be the same (well, strict subset of) #26157 or am I missing something?

solemnwarning:
@vadz hah! I didn't see the other PR or the referenced bug, closing this one.

PR #26175, opened 2026-02-09
Update zh_CN translation

CloverGit:


vadz:
Thanks for the update!

I'm unable to review this, of course, but I'll merge it and let @0tkl, @hjiawei and any other previous translators let me know if there is anything they disagree with.

PR #26176, opened 2026-02-09
Add `wxTextEntryDialog::SetHint()`

lanurmi:
Sets the hint text for the underlying `wxTextCtrl`.

Also add a corresponding, intentionally undocumented `GetHint()`.

### DIFF
--- a/include/wx/generic/textdlgg.h
+++ b/include/wx/generic/textdlgg.h
@@ -63,6 +63,9 @@ public:
     void SetValue(const wxString& val);
     wxString GetValue() const { return m_value; }
 
+    void SetHint(const wxString &hint);
+    wxString GetHint() const;
+
     void SetMaxLength(unsigned long len);
 
     void ForceUpper();
--- a/include/wx/generic/textdlgg.h
+++ b/include/wx/generic/textdlgg.h
@@ -64,7 +64,6 @@ public:
     wxString GetValue() const { return m_value; }
 
     void SetHint(const wxString &hint);
-    wxString GetHint() const;
 
     void SetMaxLength(unsigned long len);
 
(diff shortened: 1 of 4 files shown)


vadz:
I'd like to fix the bug mentioned here, can it be reproduced in `tests/controls/textentrytest.cpp`, i.e. the unit test modified to show it?

lanurmi:
> I'd like to fix the bug mentioned here, can it be reproduced in `tests/controls/textentrytest.cpp`, i.e. the unit test modified to show it?

If I change `TextCtrlTestCase::CreateText()` to unconditionally create using multiline and rich mode, i.e.
 `textctrltest.cpp:186`:
```diff
-    const long style = ms_style | extraStyles;
+    const long style = ms_style | extraStyles | wxTE_MULTILINE | wxTE_RICH2;
```

then the existing test `TextEntryTestCase::Hint()` will fail exactly the desired way:
```
wxWidgets\tests\controls\textentrytest.cpp(325): FAILED:
  REQUIRE( "" == GetTestEntry()->GetValue() )
with expansion:
  "" == "This is a hint"
```

This is clearly not the correct way to modify the tests, because also a bunch of other tests fail then, but allows reproducing.

vadz:
Should be fixed by the PR linked above, please let me know if you see any problems with it and please update this one to remove the workaround. TIA!

vadz:
Sorry, I should have mentioned this from the beginning, but why do we need `GetHint()` if it's not even useful enough to merit documenting? After thinking about this for a moment, I just can't imagine why would I ever want to call this function — it's a modal dialog, so it is constructed (and then I have the hint string because I call `SetHint()` with it), shown and then destroyed (and then I don't need the hint string any more), so what would be a possible use case for it?

If you can't think of one either, please just remove it (or I can do it when applying if you prefer).

lanurmi:
TBH I don't know when would one want to call `wxTextEntry::GetHint()` either – except in unit tests. I only added `GetHint()` here for completeness, in case it is needed for the same unknown reason.

PR #26180, opened 2026-02-10
Add forgotten support for %T to our own wxDateTime::Format() code

### LINKED ISSUE #26179: wxDateTime::Format broken in 3.2.5+ MSVC
### Description

In the 3.2.5 releases and the rest of the 3.2 series has a broken wxDateTime::Format if you use the %F specifier and %T in the same format string on Windows.

It looks like a fix from 3.3.0 dd76ab3 #24175 was incorrectly cherry picked b97aee4. In 3.3.0 this was a fix for mingw32 but on 3.2.5+ it was applied to all Windows setups which broke MSVC.

Tbh looking at the implementation of Format i think it's also broken on mingw32 on 3.3+ if you use %F and %T but I've not tested that.

I think format probably needs to add a %T handler.

Also the documentation for the function has not been updated to indicate that %F will use the wxWidgets method instead of the strftime version.

#### Bug description:

```
auto str = wxDateTime::Now().Format("%F %T");
```

In the above str will be "2026-02-10 T". %T is incorrectly not been parsed as a format specifier and just converted to T instead of the time.

- OS  and its version: Win11

vadz:
"%T" was only handled when strftime() could be used but not by our own implementation, resulting in failing to handle it for format string including both "%T" and one of the specifiers not supported by the standard function, such as "%l" or several others when using MinGW.

Closes #26179.

### DIFF
--- a/src/common/datetimefmt.cpp
+++ b/src/common/datetimefmt.cpp
@@ -634,6 +634,10 @@ wxString wxDateTime::Format(const wxString& formatp, const TimeZone& tz) const
                     res += wxString::Format(fmt, tm.sec);
                     break;
 
+                case wxT('T'):       // time as %H:%M:%S
+                    res += wxString::Format(wxT("%02d:%02d:%02d"), tm.hour, tm.min, tm.sec);
+                    break;
+
                 case wxT('U'):       // week number in the year (Sunday 1st week day)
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
The other item of my issue was the documentation doesn't mention that %F uses wxWidgets implementation on mingw and currently states "e.g. MSVC" but that should be updated to "Mingw"

PR #26187, opened 2026-02-13
Fix check boxes and radio buttons acting like generic buttons to screen readers when dark mode is enabled. Closes #26184

### LINKED ISSUE #26184: Dark mode makes checkboxes owner-drawn, breaking screen reader accessibility
## Description

When dark mode is enabled via `wxApp::SetAppearance(Appearance::System)`, checkboxes become owner-drawn, which causes screen readers to report them as "buttons" instead of "checkboxes". The checked/unchecked state is also not properly exposed. Switching back to light mode restores the correct "checkbox" role.

## Root Cause

I traced the issue through the following code path:

1. `wxControl::MSSubclassWindow()` (`src/msw/control.cpp`) checks `MSWGetDarkModeSupport()` when dark mode is active
2. `wxCheckBox::MSWGetDarkModeSupport()` (`src/msw/checkbox.cpp:90-100`) sets `support.setForeground = true`:
   ```cpp
   bool wxCheckBox::MSWGetDarkModeSupport(MSWDarkModeSupport& support) const
   {
       wxCheckBoxBase::MSWGetDarkModeSupport(support);
       support.setForeground = true;
       return true;
   }
   ```
3. This triggers `SetForegroundColour()` in `control.cpp:138-139`
4. `SetForegroundColour()` calls `MSWMakeOwnerDrawnIfNecessary()` -> `MSWMakeOwnerDrawn(true)` (`control.cpp:422-428`)
5. `MSWMakeOwnerDrawn()` replaces the window style with `BS_OWNERDRAW`, stripping the native `BS_CHECKBOX` style:
   ```cpp
   updateStyle.TurnOff(BS_TYPEMASK).TurnOn(BS_OWNERDRAW);
   ```

Screen readers rely on the `BS_CHECKBOX` window style to identify the control role. When it is replaced with `BS_OWNERDRAW`, the control is reported as a generic "button" and the checked state won't be communicated.

## Steps to Reproduce

1. Enable application dark mode on Windows 10/11
2. Create a wxWidgets application that calls `SetAppearance(wxApp::Appearance::System)`
3. Create a dialog containing `wxCheckBox` controls
4. Navigate to a checkbox using a screen reader (tested with NVDA)
5. The screen reader announces "button" instead of "check box"

## Expected Behavior

Checkboxes should retain their "check box" role and expose their checked/unchecked state to screen readers regardless of whether dark mode is active.

## Possible Solutions

Set an appropriate MSAA/UIA accessible role (`ROLE_SYSTEM_CHECKBUTTON` / `UIA_CheckBoxControlTypeId`) on owner-drawn checkboxes so screen readers can still identify them correctly. Alternatively, find a way to change the foreground color in dark mode without switching to owner-drawn rendering (e.g., handling `WM_CTLCOLORSTATIC`).

## Environment

- wxWidgets version: 3.3.1
- OS: Windows 10/11
- Screen readers tested: NVDA, Narrator
- Discovered in a Rust application using wxWidgets via [wxDragon](https://github.com/nicebyte/wxdragon) bindings

trypsynth:


vadz (commented):
Thanks, if this works (I assume you've already tested it?), it should be merged but I'd like to change 2 things:

1. Only use our custom `CreateAccessible()` in dark mode: if there are no problems in light mode, I don't want to override the default implementation which may have some extra features compared to ours.
2. Reuse the code by adding some `wxControlAccessible` class in `include/wx/private/accessible.h` and putting the common parts into it.

Could you please change it like this? If not, I can do it myself later. TIA!

trypsynth:
@vadz I tested it with a small C++ program that showed two check boxes, a radio group, and buttons. The controls read correctly in both light and dark modes. I fired off this PR right before starting work at my dayjob, but I'll try to get around to it this weekend if you don't :)

trypsynth:
@vadz done

vadz:
Thanks, looks good and I'll merge it soon with some minor changes:

- Move the header to `msw` subdirectory as it is wxMSW-specific currently.
- Get rid of `GetRole()` and just specify it as template parameter.
- Some reformatting (more blank lines for better readability).

Thanks again!

PR #26189, opened 2026-02-13
Ensure windows are destroyed before wxApp::OnExit() is called

### LINKED ISSUE #26172: `wxApp::OnExit()` may be called before all windows are destroyed

vadz:
In some circumstances we could call wxApp::OnExit() before destroying the application windows, which violated the documented (and reasonably expected) behaviour.

Fix this by adding a wrapper CallOnExit() function which destroys all windows before calling OnExit() and, for good measure, then does it again in case the user-overridden version of this function created more windows when it was called.

This also has a nice side effect of freeing the user code from the need to call the base OnExit(), as it is now trivial and doesn't do anything any more.

Closes #26172.

### DIFF
--- a/include/wx/app.h
+++ b/include/wx/app.h
@@ -482,6 +482,10 @@ public:
     // returns true for GUI wxApp subclasses
     virtual bool IsGUI() const { return false; }
 
+    // Perform the always needed cleanup before and after calling possibly
+    // overridden OnExit().
+    int CallOnExit();
+
 
     // command line arguments (public for backwards compatibility)
     int argc;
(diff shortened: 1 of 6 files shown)


vadz:
Cherry picked in the commit referenced above.

PR #26196, opened 2026-02-15
Change return type of wxGetTranslation

MaartenBent:
Change it from `const wxString&` to just `wxString` so there is no need to keep a copy around of untranslated strings.
This allows to remove `thread_local UntranslatedStringHolder` used by `GetUntranslatedString`, which causes many problems with MinGW.

Fixes https://github.com/wxWidgets/wxWidgets/issues/26195

### DIFF
--- a/include/wx/intl.h
+++ b/include/wx/intl.h
@@ -210,16 +210,16 @@ public:
     //
     // domains are searched in the last to first order, i.e. catalogs
     // added later override those added before.
-    const wxString& GetString(const wxString& origString,
-                              const wxString& domain = wxEmptyString) const
+    wxString GetString(const wxString& origString,
+                       const wxString& domain = wxEmptyString) const
     {
         return wxGetTranslation(origString, domain);
     }
     // plural form version of the same:
-    const wxString& GetString(const wxString& origString,
-                              const wxString& origString2,
-                              unsigned n,
-                              const wxString& domain = wxEmptyString) const
+    wxString GetString(const wxString& origString,
+                       const wxString& origString2,
+                       unsigned n,
+                       const wxString& domain = wxEmptyString) const
     {
         return wxGetTranslation(origString, origString2, n, domain);
     }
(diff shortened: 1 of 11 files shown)


vadz:
Looks good to me, thanks!

Pinging @vslavik just in case I'm forgetting any reason not to apply this.

vslavik:
> Pinging @vslavik just in case I'm forgetting any reason not to apply this.

I don't remember the reasons for this code (maybe git blame can provide hints, but probably not after all that time), but the only ones I can think of are performance (seems unlikely though) and API compatibility back when we unified ANSI and Unicode  builds. Plausibly the function returned wxChar* back then and there was a risk of it being stored. If wxGetTranslation returned a copy and its wxChar* conversion was implicitly called, there would be use-after-free.

I don't think this is concern nowadays.

taler21:
Such a change needs to be documented as a potential incompatibility in `Changes in behaviour not resulting in compilation errors`.

For example, in our project there is in some places code like this
```cpp
const wxString& GetText() const { return wxGetTranslation( ms_texts[m_index] ); }
```
that needs to be adapted accordingly.

MaartenBent:
Good point. I mistakenly thought the changelog was only updated whenever a release is made, but incompatible changes should be added immediately.

Depending on settings, compilers do give a warning or error after this change, like:
```
error C2220: the following warning is treated as an error
warning C4172: returning address of local variable or temporary
```
But this is not the default, so putting it in `Changes in behaviour not resulting in compilation errors` seems correct to me.

vadz:
I thought compilers would warn about this but if they don't, this change is worse than I thought :-( It's probably still not enough to revert it.

I'll add a note to the changelog.

taler21:
@vadz, after this change has been reverted in df0d6be9ca28b571e652a226cb151f2d1dfe2465 (`Revert "Change return type of wxGetTranslation", 2026-05-03`), 079b891f3b3d02ae546207d9a6c3e1204870c6de (`Document change of wxGetTranslation() return type to wxString, 2026-02-18`) could/should probably also be reverted.

vadz:
Thanks for the reminder, will revert too.

PR #26200, opened 2026-02-17
Fix install name change under macOS

### LINKED ISSUE #25675: Build failure on macOS 11 and earlier due to use of `readlink -f`
### Build System Used

- [ ] CMake
- [x] configure
- [ ] nmake with makefile.vc
- [ ] Xcode

### Description

wxWidgets 3.2.8.1 doesn't build on macOS versions earlier than 12, at least when installed using MacPorts. This problem didn't occur in 3.2.4, the previous version we used in MacPorts.

The error is:

```
readlink: illegal option -- f
usage: readlink [-n] [file ...]
```

followed by errors from `install_name_tool`. If you want a full build log, see https://trac.macports.org/ticket/72751.

I believe the problem was introduced in 65ad6849d1adb80cbfa6902a281f4ec410b2b01f, which introduced the use of `readlink -f`. `readlink` doesn't have a `-f` flag until macOS 12.

- wxWidgets version  you are building: 3.2.8.1
- wxWidgets port  you are building: wxOSX
- OS  and its version: macOS 11 and earlier
- Compiler  being used: clang
- Non-default compiler options, if any : probably not relevant

vadz:
This should really fix #25675.

Please test, TIA!

### DIFF
--- a/configure
+++ b/configure
@@ -34272,24 +34272,28 @@ rm -f core conftest.err conftest.$ac_objext \
             cat <<EOF >change-install-names
 #!/bin/sh
 set -e
-libdir=\$(cd lib ; pwd -P)
-libnames=\$(cd lib ; ls -1 libwx*${WX_RELEASE}.dylib)
+[ -z "\${WXDEBUG_X}" ] || set -x
+destdir=\$1
+libdir=\$2
+bindir=\$3
+build_libdir=\$4
+libnames=\$(cd \${build_libdir}; printf '%s\n' libwx*${WX_RELEASE}.dylib)
 changes=''
-for dep in \${libnames} ; do
-    target=\$(readlink \${4}/\${dep})
-    changes="\${changes} -change \${libdir}\${target} \${3}/\${dep}"
+for libname in \${libnames} ; do
+    target=\$(realpath \${build_libdir}/\${libname})
+    changes="\${changes} -change \${target} \${libdir}/\$(basename \${target})"
 done
-for i in \${libnames} ; do
-    lib=\$(readlink \${1}/\${i})
-    ${HOST_PREFIX}install_name_tool \${changes} -id \${3}/\${i} \${libdir}/\${lib}
+for libname in \${libnames} ; do
+    target=\$(realpath \${build_libdir}/\${libname})
+    ${HOST_PREFIX}install_name_tool \${changes} -id \${libdir}/\${libname} \${destdir}\${libdir}/\$(basename \${target})
 done
 
-if test -f "\${2}/wxrc-${WX_RELEASE}" ; then
-    ${HOST_PREFIX}install_name_tool \${changes} \${2}/wxrc-${WX_RELEASE}
+if test -f \${destdir}\${bindir}/wxrc-${WX_RELEASE} ; then
+    ${HOST_PREFIX}install_name_tool \${changes} \${destdir}\${bindir}/wxrc-${WX_RELEASE}
 fi
 EOF
             chmod +x change-install-names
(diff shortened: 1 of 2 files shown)


ravidavi:
This works great for my use case (the NASA GMAT project). After doing a `make install`, the dylibs in the install directory have the correct install names.

vadz:
Thanks for testing!

I'm going to push this soon if there are no other comments.

pauldmccarthy:
@vads @ravidavi I haven't looked closely at the logic, but this patch seems to work for the wxwidgets 3.2.9 conda-forge package - thanks!

vadz:
Thanks for confirmation, I'll push this to master too soon, so that it becomes part of the upcoming 3.3.2.

vadz:
Hmm, I'm getting

```
/Users/xcodeserver/actions-runner/_work/wxWidgets/wxWidgets/change-install-names: line 11: realpath: command not found
```

in [this CI run](https://github.com/wxWidgets/wxWidgets/actions/runs/22302473813/job/64513448987) under macOS 12.6.9 so it's clearly not available in this version and we need make more changes to avoid breaking building there :-(

pauldmccarthy:
It sounds like `readlink -f` was added in 12.3, and `realpath` was added in 13.0: https://apple.stackexchange.com/questions/450035/is-the-unix-realpath-command-distributed-with-macos-ventura#comment689104_450116

Sorry I didn't check (although in my defence, Apple does not make things easy :/).

How about going back to basics? `cd`+`dirname`+`pwd`? https://github.com/JeffersonLab/coatjava/pull/120/changes

_edit_ no, that'll only work for directories, not files..

vadz:
Using `$(readlink ${build_libdir}/$(readlink ${build_libdir}/${libname}))` should work, I'll push this to master soon, please retest before I backport this to 3.2.

pauldmccarthy:
> Using `$(readlink ${build_libdir}/$(readlink ${build_libdir}/${libname}))` should work, I'll push this to master soon, please retest before I backport this to 3.2.

Won't that throw an error if `${build_libdir}` is not a symbolic link?

```
➜  ~ pwd                             
/Users/paulmc
➜  ~ readlink /Users/paulmc/; echo $?
1
```

From the 15.5 manpage:
> When invoked as readlink, only the target of the symbolic link is printed.  If the given argument is not a symbolic link and the -f option is not specified, readlink will print nothing and exit with an error. 

I don't have access to older versions, but https://ss64.com/mac/stat.html says:

> When invoked as readlink, only the target of the symbolic link is printed. If the given argument is not a symbolic link, readlink will print nothing and exit with an error.

What a pain!

vadz:
> > Using `$(readlink ${build_libdir}/$(readlink ${build_libdir}/${libname}))` should work, I'll push this to master soon, please retest before I backport this to 3.2.
> 
> Won't that throw an error if `${build_libdir}` is not a symbolic link?

All path components don't have to be symlinks, only the last one does, and this is the case here, so in theory this should work. But more testing would be very welcome (I've pushed my changes now).
 
> What a pain!

I can only agree with this.

pauldmccarthy:
How about adding this at the top?

```
resolve_path() {
    # macos >= 13.0
    if which realpath > /dev/null; then
        cmd="realpath"
    # macos < 13.0
    else
        cmd="readlink"
    fi
    $cmd $1
}

resolve_path $1
```

I've opened #26230, and am presently testing on macOS 15.5

pauldmccarthy:
> All path components don't have to be symlinks, only the last one does, and this is the case here, so in theory this should work. But more testing would be very welcome (I've pushed my changes now).

Aah, no worries - I've closed #26320, and will test your `readlink (readlink)` suggestion.

vadz:
No, this won't work at all, the commands don't return the same result, one returns the full path while another just returns the link target. I wish it were as samples as this...

PR #26207, opened 2026-02-19
Cairo: Add support for Cygwin DLL

carlo-bramini:
This patch just adds the right name of the Cairo shared library on CYGWIN.

EDIT: I would like to backport this tiny patch to 3.2, what is the branch to use for this purpuse?

### DIFF
--- a/src/common/cairo.cpp
+++ b/src/common/cairo.cpp
@@ -330,6 +330,8 @@ wxCairo::wxCairo()
 
 #ifdef __WXMSW__
     wxString cairoDllStr("libcairo-2.dll");
+#elif defined(__CYGWIN__)
+    wxString cairoDllStr("cygcairo-2.dll");
 #elif defined(__WXOSX__)
     wxString cairoDllStr("libcairo.2.dylib");
 #else


src/common/cairo.cpp
```diff
@@ -330,6 +330,8 @@ wxCairo::wxCairo()
 
 #ifdef __WXMSW__
     wxString cairoDllStr("libcairo-2.dll");
+#elif defined(__CYGWIN__)
```
vadz, line 333: Hmm, isn't `__WXMSW__` defined for Cygwin too? I.e. shouldn't this be _before_ the test for `__WXMSW__`?

vadz (commented):
I can cherry pick this commit to 3.2, no problem.

src/common/cairo.cpp
carlo-bramini: > Hmm, isn't `__WXMSW__` defined for Cygwin too? I.e. shouldn't this be _before_ the test for `__WXMSW__`?

I'm not expert of wxWidgets internal code, but probably `__WXMSW__` is not.
Actually, I built both 3.2.9 and 3.3.1 with Gtk+3 and Qt5.
I have seen that `__WXGTK20__` / `__WXGTK__` and `__WXQT__ ` exist in the code, so I suspect that `__WXMSW__` won't be declared in those cases, is it correct?

vadz: Yes, if you build wxGTK or wxQT, `__WXMSW__` won't be defined. But if you build wxMSW with Cygwin, it would be. But I don't actually know which DLL we want to use for Cygwin build of wxMSW...

carlo-bramini: Thank you very much, I updated the patch according to your request.

PR #26209, opened 2026-02-19
Add wxGLContext::GetProcAddress()

### LINKED ISSUE #9215: Add `wxGLCanvas::GetProcAddress()`

vadz:
Implement this function as a trivial wrapper for wglGetProcAddress(), glXGetProcAddressARB() or eglGetProcAddress() depending on the platform (currently not implemented for macOS).

Account for the known bug in some Windows OpenGL drivers by checking for manifestly invalid return values and handling them as null.

Use the new function in wxGL code itself and in the pyramid sample.

Closes #9215.

### DIFF
--- a/include/wx/glcanvas.h
+++ b/include/wx/glcanvas.h
@@ -182,6 +182,8 @@ public:
 // wxGLContextBase: OpenGL rendering context
 // ----------------------------------------------------------------------------
 
+using wxGLExtFunction = void (*)();
+
 class WXDLLIMPEXP_GL wxGLContextBase : public wxObject
 {
 public:
@@ -199,6 +201,20 @@ public:
     static void ClearCurrent();
 
     bool IsOK() const { return m_isOk; }
+
+    // Get pointer to OpenGL extension function, return nullptr if not found.
+    static wxGLExtFunction GetProcAddress(const wxString& name);
+
+    // Same as above, but returns the function of the specified type.
+    template <typename T>
+    static T GetProcAddress(const wxString& name)
+    {
+        wxGCC_WARNING_SUPPRESS_CAST_FUNCTION_TYPE()
+
+        return reinterpret_cast<T>(GetProcAddress(name));
+
+        wxGCC_WARNING_RESTORE_CAST_FUNCTION_TYPE()
+    }
 
 protected:
     bool m_isOk;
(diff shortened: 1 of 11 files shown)


vadz:
@dsa-t I've finally done the simplest thing possible and just mentioned the Windows complications in the documentation because I'm not sure we should impose the extra burden of selecting the context first on all library users.

Please let me know if this minimal version is sufficient for your needs and, if not, what is missing. TIA!

dsa-t:
Thanks, looks good for me!

On macOS, GLEW and libepoxy load "/System/Library/Frameworks/OpenGL.framework/Versions/Current/OpenGL"

https://github.com/nigels-com/glew/blob/ee5586bbd99c2976bafee5de97bbacc38c71a103/auto/src/glew_head.c#L85-L101

https://github.com/anholt/libepoxy/blob/1b6d7db184bb1a0d9af0e200e06a0331028eaaae/src/dispatch_common.c#L178

GLAD tries to load from:

```
    static const char *NAMES[] = {
        "../Frameworks/OpenGL.framework/OpenGL",
        "/Library/Frameworks/OpenGL.framework/OpenGL",
        "/System/Library/Frameworks/OpenGL.framework/OpenGL",
        "/System/Library/Frameworks/OpenGL.framework/Versions/Current/OpenGL"
    };
```

According to https://yairchu.github.io/posts/intercept-to-fix, "/System/Library/Frameworks/OpenGL.framework/OpenGL" is gone in macOS 11.0

vadz:
I still have `/System/Library/Frameworks/OpenGL.framework/OpenGL` (weirdly enough, it's a symlink pointing at non existing `Versions/Current/OpenGL`, but this doesn't prevent `dlopen()` from opening it) on my macOS 14.4.1 system, so we could probably use this. But I'm not sure if it's worth doing this without knowing if it actually works or not (although I checked that `glGenBuffers` I randomly used for my example can be loaded from it). It might, but I really don't like adding code without testing it at all so I'm tempted to merge this and then we can add macOS implementation when you or anybody else needs it, which will hopefully also allow us to test that it actually works.

vadz:
Merged, we can always add macOS support later.

PR #26213, opened 2026-02-20
Fix for wxTextEntryDialog resizing in unexpected ways

### LINKED ISSUE #25739: wxTextEntryDialog resizes in unexpected ways

Blake-Madden:
Resizes width and height with multiline control, but doesn't go down to 0,0.
Single-line only resizes horizontally now.
Closes #25739

### DIFF
--- a/src/generic/textdlgg.cpp
+++ b/src/generic/textdlgg.cpp
@@ -117,6 +117,15 @@ bool wxTextEntryDialog::Create(wxWindow *parent,
 
     topsizer->Fit( this );
 
+    // Set minimum size to the fitted size so dialog can't shrink below content
+    SetMinSize(GetSize());
+
+    // For single-line entry, also constrain max height to prevent vertical expansion
+    if ( !(style & wxTE_MULTILINE) )
+    {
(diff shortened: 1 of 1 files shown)


vadz:
Thanks! Will push soon with a minor change to avoid calling `GetSize()` as this size is already returned by `Fit()` anyhow.

PR #26214, opened 2026-02-20
Edge Webview fixes

Blake-Madden:
Use `ICoreWebView2_22::AddWebResourceRequestedFilterWithRequestSourceKinds()` when available (#25816)

Fix memory file system in issues:

In the Webview sample using Edge (Windows 11), select "Menu" -> "Handler Examples" -> "Memory File System". Note that you will see this and the "Page 2" link doesn't work.

image

This PR fixes that, as well as the deprecation warning from #25816


vadz (approved):
Thanks, looks fine to me.

Pinging @TcT2k just in case you have any comments.

If not, I'll merge this soon.

TcT2k:
While I haven't looked into the updated API, the changes look good. Correct handling for iframes in custom protocol handlers is definitely good.

PR #26215, opened 2026-02-20
OSX memory leak

csomor:
see #26208

### DIFF
--- a/src/osx/core/uilocale.mm
+++ b/src/osx/core/uilocale.mm
@@ -309,7 +309,7 @@ wxUILocaleImplCF::GetMonthName(wxDateTime::Month month, wxDateTime::NameForm for
 wxString
 wxUILocaleImplCF::GetWeekDayName(wxDateTime::WeekDay weekday, wxDateTime::NameForm form) const
 {
-    NSDateFormatter* df = [[NSDateFormatter new] autorelease;
+    NSDateFormatter* df = [[NSDateFormatter new] autorelease];
     df.locale = m_nsloc;
 
     NSArray* weekdayNames = nullptr;
--- a/src/osx/core/uilocale.mm
+++ b/src/osx/core/uilocale.mm
@@ -264,7 +264,7 @@ wxUILocaleImplCF::GetLocalizedName(wxLocaleName name, wxLocaleForm form) const
 wxString
 wxUILocaleImplCF::GetMonthName(wxDateTime::Month month, wxDateTime::NameForm form) const
 {
-    NSDateFormatter* df = [NSDateFormatter new];
+    NSDateFormatter* df = [[NSDateFormatter new] autorelease];
     df.locale = m_nsloc;
 
     NSArray* monthNames = nullptr;
@@ -309,7 +309,7 @@ wxUILocaleImplCF::GetMonthName(wxDateTime::Month month, wxDateTime::NameForm for
 wxString
 wxUILocaleImplCF::GetWeekDayName(wxDateTime::WeekDay weekday, wxDateTime::NameForm form) const
 {
-    NSDateFormatter* df = [NSDateFormatter new];
+    NSDateFormatter* df = [[NSDateFormatter new] autorelease;
     df.locale = m_nsloc;
 
     NSArray* weekdayNames = nullptr;
--- a/src/osx/core/uilocale.mm
+++ b/src/osx/core/uilocale.mm
@@ -264,7 +264,7 @@ wxUILocaleImplCF::GetLocalizedName(wxLocaleName name, wxLocaleForm form) const
 wxString
 wxUILocaleImplCF::GetMonthName(wxDateTime::Month month, wxDateTime::NameForm form) const
 {
-    NSDateFormatter* df = [[NSDateFormatter new] autorelease];
+    NSDateFormatter* df = [NSDateFormatter new];
     df.locale = m_nsloc;
 
     NSArray* monthNames = nullptr;
@@ -303,13 +303,15 @@ wxUILocaleImplCF::GetMonthName(wxDateTime::Month month, wxDateTime::NameForm for
     }
 
     NSString* monthName = [monthNames objectAtIndex:(month)];
-    return wxCFStringRef::AsString(monthName);
+    wxCFStringRef cf(monthName);
+    [df release];
+    return cf.AsString();
 }
 
 wxString
 wxUILocaleImplCF::GetWeekDayName(wxDateTime::WeekDay weekday, wxDateTime::NameForm form) const
 {
-    NSDateFormatter* df = [[NSDateFormatter new] autorelease];
+    NSDateFormatter* df = [NSDateFormatter new];
     df.locale = m_nsloc;
 
     NSArray* weekdayNames = nullptr;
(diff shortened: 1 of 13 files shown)


vadz:
Thanks Stefan! Two questions:

1. Does this really fix the leak, i.e. do we not need to release `m_osxClipView` finally?
2. Can this be backported to 3.2 as is?

csomor:
@vadz this is a work in progress, I'll have to think through the different cycles, it's not correct as of now, hope I'll have it done over the weekend

src/osx/core/uilocale.mm
```diff
@@ -264,7 +264,7 @@ int CompareStrings(const wxString& lhs, const wxString& rhs,
 wxString
 wxUILocaleImplCF::GetMonthName(wxDateTime::Month month, wxDateTime::NameForm form) const
 {
-    NSDateFormatter* df = [NSDateFormatter new];
+    NSDateFormatter* df = [[NSDateFormatter new] autorelease];
```
vadz: Should we use `autorelease` here? This can be used in a non-GUI code and I'm not sure if an auto release pool is always available in it. It looks like we could use `wxCFRef<NSDateFormatter>` instead for more deterministic behaviour.

vadz (commented):
Thanks Stefan, is this ok to (squash) merge now?

Also, this almost surely doesn't apply cleanly to 3.2, but at least the changes in `window.mm` are needed there too, right?

src/osx/core/uilocale.mm
csomor: ah ok, I'll change that

csomor:
> Thanks Stefan, is this ok to (squash) merge now?
> 
> Also, this almost surely doesn't apply cleanly to 3.2, but at least the changes in `window.mm` are needed there too, right?

I'd think whenever the diff matches, then it could be applied to the files in 3.2 as well, but at least window.mm, yes

vadz:
Thanks again, Stefan!

This still has "work needed" label, do you plan to make more changes or did you just forget to reset it perhaps?

Also, I'll probably squash merge this because it doesn't make sense to keep all commits here, but perhaps it would be useful to re-split it instead? I can do if you can tell me what kind of logically atomic/independent changes are there. Or we can just make one big "fix leaks" commit, of course...

csomor:
> Thanks again, Stefan!
> 
> This still has "work needed" label, do you plan to make more changes or did you just forget to reset it perhaps?
> 
> Also, I'll probably squash merge this because it doesn't make sense to keep all commits here, but perhaps it would be useful to re-split it instead? I can do if you can tell me what kind of logically atomic/independent changes are there. Or we can just make one big "fix leaks" commit, of course...

label removed

re spliting would be most elegant thanks

The groups would be

for removing gc code and fixing the clip view leak
ea8bb27
038da4e
48ecd61

for fixing leaks related to synthesized properties
4882bfa
036ef29
364c3e3
0a642ab

for different leaks found by Xcode analysis
82e1a41
b35ed9b
c01e7f1

and the rest of the commits for the 3 other separate leaks

Thanks a lot!

vadz:
Thanks, I've recombined the commits and will push them soon. For the record, here are the commit messages, please let me know if I got anything wrong:

```
commit 5e2caa4db96b2da3b78298a304fb0c802fd7e18d
Author: Stefan Csomor <csomor@advancedconcepts.ch>
Date:   2026-02-20 23:34:27 +0100

    Remove GC-aware code and fix clip view leak
    
    Remove "GC-aware" code as GC is not supported by the OS any longer.
    
    Fix leak of wxWindow::m_osxClipView by releasing it after adding it as
    subview.

commit b581980bef7684e6ab40ed29de3449d3390036bd
Author: Stefan Csomor <csomor@advancedconcepts.ch>
Date:   2026-02-21 11:59:46 +0100

    Fix miscellaneous memory leaks related to synthesized properties
    
    When using @property(retain), we need to reset the pointer to prevent it
    from being leaked when the containing object is destroyed.

commit 985a64f11da1d4f2d524e042372a3cd7757f6731
Author: Stefan Csomor <csomor@advancedconcepts.ch>
Date:   2026-02-21 15:17:31 +0100

    Fix memory leaks reported by Xcode Memory Analysis
    
    Mostly just add autorelease to avoid various objects being leaked, but
    also add manual calls to release for the code which is not necessarily
    executed inside a GUI application and so can't rely on autorelease pool
    existence.

commit d23d483241210029b3ec8cc6324c018eb239556b
Author: Stefan Csomor <csomor@advancedconcepts.ch>
Date:   2026-02-21 17:08:08 +0100

    Fix leak of buttons in wxFontDialog
    
    Release buttons after adding them to superview.

commit e69aa968d1f7825f16f78495b66c98a09b57b4e5
Author: Stefan Csomor <csomor@advancedconcepts.ch>
Date:   2026-02-21 17:25:05 +0100

    Fix memory leak of NSPasteboardItem in wxOSXPasteboard code
    
    Add missing release.

commit a2bcc1479cc0dfd89ea2e32ca54413da7ecfe117
Author: Stefan Csomor <csomor@advancedconcepts.ch>
Date:   2026-02-21 17:56:40 +0100

    Fix tracking area memory leak
```

(I failed to come up with an explanation for the last commit as I'm not sure what's going on here, notably I don't understand why is it safe to call `removeTrackingArea:` on a pointer which may be dangling if we had released it before...).

csomor:
> Thanks, I've recombined the commits and will push them soon. For the record, here are the commit messages, please let me know if I got anything wrong:

They are perfect, thanks a lot

> commit a2bcc1479cc0dfd89ea2e32ca54413da7ecfe117
> Author: Stefan Csomor <csomor@advancedconcepts.ch>
> Date:   2026-02-21 17:56:40 +0100
> 
>     Fix tracking area memory leak
> ```
> 
> (I failed to come up with an explanation for the last commit as I'm not sure what's going on here, notably I don't understand why is it safe to call `removeTrackingArea:` on a pointer which may be dangling if we had released it before...).

The rules of memory management (Ownership Policy) in NextStep/OpenStep/Cocoa mean : 

```
    m_trackingArea = [[NSTrackingArea alloc] initWithRect: [self bounds]
                                                  options: options
                                                    owner: self
                                                 userInfo: nil];
```
now freshly created, the refcount (aka retainCount) is now 1

```
    [self addTrackingArea:m_trackingArea];
```
now "this" takes the ownership, refcount is now 2
https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmPractical.html#//apple_ref/doc/uid/TP40004447-SW12

```
    [m_trackingArea release];
```

we can release here, refcount is now 1, all is in the hand of "this" now, when it gets deallocated it deallocates the tracking area as well

(now in the case of a later call due to relayout), the former if clause gets called, 

```
    if( m_trackingArea != nil )
    {
        [self removeTrackingArea:m_trackingArea];
    }
```
now the removeTrackingArea removes the object, releasing it, now the "old" tracking area has a refcount of 0 and gets deallocated

csomor:
funny enough, our former version of the tracking area code is modeled exactly after the docs, which IMHO always will leak the last instance of a TrackingArea https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/TrackingAreaObjects/TrackingAreaObjects.html

vadz:
Applied now, thanks again! Will try to backport the parts that make sense to 3.2.

PR #26216, opened 2026-02-21
OSX: fix missing axes in wxJoystick

Hydr8gon:
This fixes an issue I was having where only 3 joystick axes were available on macOS. I also added some code to handle D-pads as X/Y axes, which is the behaviour I observe when using wxJoystick on Linux.

### DIFF
--- a/src/osx/core/hidjoystick.cpp
+++ b/src/osx/core/hidjoystick.cpp
@@ -64,6 +64,19 @@ enum
     wxJS_AXIS_RUDDER,
     wxJS_AXIS_U,
     wxJS_AXIS_V,
+    wxJS_AXIS_EXT,
+};
+
+enum
+{
+    wxJS_DPAD_UP = 0,
+    wxJS_DPAD_UPRIGHT,
+    wxJS_DPAD_RIGHT,
+    wxJS_DPAD_RIGHTDOWN,
+    wxJS_DPAD_DOWN,
+    wxJS_DPAD_DOWNLEFT,
+    wxJS_DPAD_LEFT,
+    wxJS_DPAD_LEFTUP,
 };
 
 //---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/osx/core/hidjoystick.cpp
```diff
@@ -285,27 +287,21 @@ wxString wxJoystick::GetProductName() const
 //---------------------------------------------------------------------------
 int wxJoystick::GetNumberButtons() const
 {
-    int nCount = 0;
-
-    for(int nIndex = 0; nIndex < 40; ++nIndex)
+    for (int i = 40; i > 0; i--)
```
vadz: This is not new, but it looks like returning anything `> wxJS_MAX_BUTTONS` would result in problems, so I wonder if it could make sense to start from `wxJS_MAX_BUTTONS` instead of 40 (but then I have no idea where does this constant come from...).

src/osx/core/hidjoystick.cpp
```diff
@@ -285,27 +287,21 @@ wxString wxJoystick::GetProductName() const
...
-
-    return nCount;
+    return 0;
 }
 int wxJoystick::GetNumberAxes() const
 {
-    int nCount = 0;
-
-    for(int nIndex = 40; nIndex < 50; ++nIndex)
+    for (int i = 50; i > 40; i--)
```
vadz: Same comment as above but for `wxJS_MAX_AXES`.

src/osx/core/hidjoystick.cpp
```diff
@@ -905,8 +930,56 @@ uint64_t MachTimeToNanoseconds(uint64_t machTime)
...
             pThis->m_axe[2] = hidevent.value;
         }
+        else if (nIndex == m_hid->m_nDpadIdx)
+        {
+            // Remap the D-pad state to reasonable axis values
+            wxevent.SetEventType(wxEVT_JOY_MOVE);
+            int *axe = &pThis->m_axe[nIndex - wxJS_MAX_BUTTONS - 1];
+            switch (hidevent.value)
+            {
+                case 0: // Up
```
vadz: Are there no symbolic constants for the values in this switch?

vadz (approved):
Thanks, I'm unable to test this, but if it works for you, let's merge it. This code could probably be made more clear by using more symbolic constants and fewer hardcoded values but this is a problem that clearly predates this PR.

PR #26217, opened 2026-02-21
CMake backports 02-2026

MaartenBent:
Remove searching for Qt6 in CMake, as discussed in #26065.
And add the `FindGTK2` module that first uses PkgConfig to find GTK2 and if that fails, the built-in module.


vadz:
Thanks, merged, will push soon.

vadz:
This was merged in 3.2 branch now and is part of 3.2.10, so closing.

PR #26219, opened 2026-02-21
[3.2] wxQt: fix undefined reference to cairo_surface_get_device_scale()

carlo-bramini:
This patch fixes issue #26218.


carlo-bramini:
Excuse me, it looks like I did some mistakes because this PR is applied to master rather than 3.2.

PR #26220, opened 2026-02-21
[3.2] wxQt: fix undefined reference to cairo_surface_get_device_scale()

carlo-bramini:
This patch fixes issue #26218.

### DIFF
--- a/src/common/cairo.cpp
+++ b/src/common/cairo.cpp
@@ -186,7 +186,9 @@
     m( cairo_user_to_device_distance, \
        (cairo_t* cr, double *dx, double* dy), (cr, dx, dy) ) \
     m( cairo_surface_mark_dirty, \
-       (cairo_surface_t* surface), (surface))
+       (cairo_surface_t* surface), (surface)) \
+    m( cairo_surface_get_device_scale, \
+       (cairo_surface_t *surface, double *x_scale, double *y_scale), (surface, x_scale, y_scale))
 
 #ifdef __WXMAC__
 #define wxCAIRO_PLATFORM_METHODS(m) \


vadz:
AFAICS this should be applied to master too, even if this renderer is not used there by default, it can still be used if explicitly chosen, right?

AliKet:
> AFAICS this should be applied to master too, even if this renderer is not used there by default, it can still be used if explicitly chosen, right?

Yes 👍

vadz:
Thanks for the confirmation, I'll apply this to both branches.

BTW, testing compilation here I get

```
/usr/include/cairo/cairo.h:70:9: warning: "cairo_public" redefined
   70 | #define cairo_public _cairo_api extern
      |         ^~~~~~~~~~~~
src/common/cairo.cpp:19:9: note: this is the location of the previous definition
   19 | #define cairo_public
      |         ^~~~~~~~~~~~
```

and looking at the header, it looks like we should be predefining `CAIRO_WIN32_STATIC_BUILD` instead, at least with Cairo 1.18.4 installed here.

carlo-bramini:
> Thanks for the confirmation, I'll apply this to both branches.
> 
> BTW, testing compilation here I get
> 
> ```
> /usr/include/cairo/cairo.h:70:9: warning: "cairo_public" redefined
>    70 | #define cairo_public _cairo_api extern
>       |         ^~~~~~~~~~~~
> src/common/cairo.cpp:19:9: note: this is the location of the previous definition
>    19 | #define cairo_public
>       |         ^~~~~~~~~~~~
> ```
> 
> and looking at the header, it looks like we should be predefining `CAIRO_WIN32_STATIC_BUILD` instead, at least with Cairo 1.18.4 installed here.

Yes, that's right, this error also happens and it can happen on both CYGWIN and in my opinion also on WIN32 (if activated in some way) because the particular behaviour of DLLs.
This is how cairo handles the dllexport/dllimport modifiers:
https://cgit.freedesktop.org/cairo/tree/src/cairo.h#n53
```
#if (defined(_WIN32) || defined(__CYGWIN__)) && !defined(CAIRO_WIN32_STATIC_BUILD)
# define _cairo_export __declspec(dllexport)
# define _cairo_import __declspec(dllimport)
#elif defined(__GNUC__)
# define _cairo_export __attribute__((__visibility__("default")))
# define _cairo_import
#else
# define _cairo_export
# define _cairo_import
#endif

#ifdef CAIRO_COMPILATION
# define _cairo_api _cairo_export
#else
# define _cairo_api _cairo_import
#endif

#define cairo_public _cairo_api extern
```
So, declaring `cairo_public` doesn't avoid the use of dllexport/dllimport.
At the time of writing, I confirm that hacking the use of `CAIRO_WIN32_STATIC_BUILD` is the only way to build it, otherwise will will get errors from the linker like these ones:
```
[0/12] Linking CXX shared library lib/cygwx_qtu_core-3.2-0.dll
ld: libs/core/CMakeFiles/wxcore.dir/__/__/__/__/src/generic/graphicc.cpp.o: in function `wxCairoPathData::MoveToPoint(double, double)':
src/generic/graphicc.cpp:1293:(.text+0x1b7): undefined reference to `__imp_cairo_move_to'
ld: libs/core/CMakeFiles/wxcore.dir/__/__/__/__/src/generic/graphicc.cpp.o: in function `wxCairoPathData::AddLineToPoint(double, double)':
src/generic/graphicc.cpp:1298:(.text+0x1c7): undefined reference to `__imp_cairo_line_to'
ld: libs/core/CMakeFiles/wxcore.dir/__/__/__/__/src/generic/graphicc.cpp.o: in function `wxCairoPathData::~wxCairoPathData()':
src/generic/graphicc.cpp:1261:(.text+0x1e4): undefined reference to `__imp_cairo_destroy'
...etc
```

vadz:
I'm almost sure that predefining `cairo_public` worked before, but they broke it in more recent versions (don't know if intentionally or not).

This is annoying because we can't test for Cairo version before including this header and while always predefining `CAIRO_WIN32_STATIC_BUILD` is harmless, defining `cairo_public` is not but still needed by old versions. We'll probably have to ignore the warning, too...

carlo-bramini:
> We'll probably have to ignore the warning, too...

Sorry, I clicked "Comment" too early without finishing the message... ^^;
At the time of writing, the build process fails anyways when linking, so just ignoring the warning doesn't seem the right solution, at least at first sight.

vadz:
Yes, we need to define `CAIRO_WIN32_STATIC_BUILD`, this is not in question.

But while we still support Cairo versions without it, we still need to keep `cairo_public` too, which is why it would be nice to disable the warning in addition to defining `CAIRO_WIN32_STATIC_BUILD`.

vadz:
OK, so the check for `cairo_public` was removed in Cairo ba4d5fbd5 (Build with hidden symbols by default, 2023-01-31) by ebassi without even bothering to explain why this was done (really surprising knowing how much importance Gnome guys attach to backwards compatibility) which is relatively recent: it was only included in 1.18.0, so we definitely can't rely on it being available in all versions.

carlo-bramini:
I opened a dedicated issue for handling this malfunction at #26224.
So, the build error on `cairo_surface_get_device_scale()` could be handled immediately.

vadz:
This was applied as 7a23d47745ac243436579558f2278fe1720d1773 which is part of #26221.

PR #26222, opened 2026-02-22
Restore wxDataViewCtrl columns view position

vadimgrn:
- wx/persist/dataview.h saves wxPERSIST_DVC_POS for each column in Save(), but does not restore them in Restore()
- This patch restores column positions
- It works well on Windows, other platforms not tested


vadz:
Thanks for implementing this!

I'll push it soon with just some minor amendments.

PR #26223, opened 2026-02-22
Added support for tablet events in MSW

undergraver:
I finished the implementation related to tablet events for Windows.

The sample if working fine and takes into account pressure and eraser features. My tablet doesn't have tilt, rotation to test this. Any feedback is welcome.

Tested on Windows 11.
Don't know exactly if the documentation from `inteface` is proper, I gave my best try in copy pasting from other places and adapt it.


include/wx/event.h
```diff
@@ -2203,6 +2209,50 @@ class WXDLLIMPEXP_CORE wxPressAndTapEvent : public wxGestureEvent
     wxDECLARE_DYNAMIC_CLASS_NO_ASSIGN(wxPressAndTapEvent);
 };
 
+class WXDLLIMPEXP_CORE wxTabletEvent : public wxEvent
+{
+public:
+    wxTabletEvent(wxWindowID winid = 0, wxEventType type = wxEVT_NULL)
+        : wxEvent(winid, type), m_pressure(-1.0), m_tiltX(-360.0), m_tiltY(-360.0), m_rotation(-1.0), m_pos(-1,-1), m_usingEraser(false)
```
vadz: Minor, but we should really initialize members in their declarations in the new code (this is not done in the vast majority of existing code because it predates C++11 by a couple of decades).

include/wx/event.h
```diff
@@ -2203,6 +2209,50 @@ class WXDLLIMPEXP_CORE wxPressAndTapEvent : public wxGestureEvent
...
+    void SetUsingEraser(bool eraser) { m_usingEraser = eraser; }
+
+    wxNODISCARD virtual wxEvent* Clone() const override { return new wxTabletEvent(*this); }
+
+protected:
+    wxDouble m_pressure; // range [ 0.0, 1.0 ]
+    wxDouble m_tiltX;    // range [ -90.0, 90.0 ]
+    wxDouble m_tiltY;    // range [ -90.0, 90.0 ]
+    wxDouble m_rotation; // range [ 0, 360.0 ]
+    wxPoint m_pos;        // position in pixels relative to the Window receiving the event
```
vadz: It would be useful say if this is in client or window coordinates?

include/wx/event.h
```diff
@@ -2203,6 +2209,50 @@ class WXDLLIMPEXP_CORE wxPressAndTapEvent : public wxGestureEvent
...
+
+    const wxPoint& GetPosition() const { return m_pos; }
+    void SetPosition(const wxPoint& pos) { m_pos = pos; }
+
+    bool IsUsingEraser() const { return m_usingEraser; }
+    void SetUsingEraser(bool eraser) { m_usingEraser = eraser; }
+
+    wxNODISCARD virtual wxEvent* Clone() const override { return new wxTabletEvent(*this); }
+
+protected:
```
vadz: I would make this private, this class is not really supposed to be inherited from anyhow, is it?

interface/wx/event.h
```diff
@@ -5159,6 +5159,88 @@ class wxSetCursorEvent : public wxEvent
...
+        Process a @c wxEVT_POINTER_DOWN event.
+    @event{EVT_POINTER_UP(func)}
+        Process a @c wxEVT_POINTER_UP event.
+    @event{EVT_POINTER_UPDATE(func)}
+        Process a @c wxEVT_POINTER_UPDATE event.
+    @endEventTable
+
+    @library{wxcore}
+    @category{events}
+ */
```
vadz, line 5188: Please add this:

```suggestion

    @since 3.3.2
 */
```

interface/wx/event.h
```diff
@@ -5159,6 +5159,88 @@ class wxSetCursorEvent : public wxEvent
...
+
+    A wxTabletEvent is generated from wxWindow when the graphical pen is used.
+
+    This event gives the application the chance to perform specific tablet events
+    processing based on the current position of the pen, pressure and other
+    parameters.
+
+    @beginEventTable{wxTabletEvent}
+    @event{EVT_POINTER_DOWN(func)}
+        Process a @c wxEVT_POINTER_DOWN event.
```
vadz: We really need to have some explanation of what this event corresponds to because I am honestly not even sure myself. Is it tablet button event? Tablet stylus being lifted event? Something else?

The same goes for the 2 other events below, of course.

interface/wx/event.h
```diff
@@ -5159,6 +5159,88 @@ class wxSetCursorEvent : public wxEvent
...
+     */
+    wxTabletEvent(wxWindowID winid = 0, wxEventType type = wxEVT_NULL);
+    /**
+        Get the pressure value. Range is from 0 to 1.
+     */
+    wxDouble GetPressure() const;
+    /**
+        Set the pressure value.
+     */
+    void SetPressure(wxDouble p);
```
vadz, line 5203: I'm not really sure we want to document all the `SetXXX()` functions, aren't they all supposed to be used only by wxWidgets itself?

samples/samples_vc17.sln
vadz, line 1: We need to make the same changes in all the other `samples_vcNN.sln` files too if we add a new sample.

src/msw/window.cpp
```diff
@@ -6338,6 +6344,81 @@ bool wxWindowMSW::HandleTouch(WXWPARAM wParam, WXLPARAM lParam)
...
+            type = wxEVT_POINTER_DOWN;
+            break;
+        case WM_POINTERUP:
+            type = wxEVT_POINTER_UP;
+            break;
+        case WM_POINTERUPDATE:
+            type = wxEVT_POINTER_UPDATE;
+            break;
+        default:
+            wxLogLastError(wxT("Unexpected pointer message"));
```
vadz: Use of `wxLogLastError()` is incorrect here, there is no last error (`::GetLastError()`) as this is a logic bug in wx code, not a Win32 error. I would either remove it entirely or replace it with `wxFAIL_MSG("Unknown pointer message")``

src/msw/window.cpp
```diff
@@ -6338,6 +6344,81 @@ bool wxWindowMSW::HandleTouch(WXWPARAM wParam, WXLPARAM lParam)
     return allHandled;
 }
 
+bool wxWindowMSW::HandlePointer(WXUINT message, WXWPARAM wParam, WXLPARAM lParam)
+{
+    wxUnusedVar(lParam);
+    wxEventType type;
+    switch(message)
```
vadz: This is minor, of course, but it would be nice to format all new code consistently with the existing coding style, i.e.

```suggestion
    switch ( message )
```

src/msw/window.cpp
```diff
@@ -6338,6 +6344,81 @@ bool wxWindowMSW::HandleTouch(WXWPARAM wParam, WXLPARAM lParam)
...
+        // set the eraser flag
+        event.SetUsingEraser( (penInfo.penFlags & PEN_FLAG_ERASER) != 0 );
+
+        const POINT& pt = penInfo.pointerInfo.ptPixelLocation;
+        wxPoint pos(pt.x, pt.y);
+        event.SetPosition( ScreenToClient(pos) );
+    }
+
+    bool handled = HandleWindowEvent(event);
+    return handled;
```
vadz: Very minor, but why do this instead of

```suggestion
    return HandleWindowEvent(event);
```

?

src/msw/window.cpp
```diff
@@ -6338,6 +6344,81 @@ bool wxWindowMSW::HandleTouch(WXWPARAM wParam, WXLPARAM lParam)
...
+            type = wxEVT_POINTER_UPDATE;
+            break;
+        default:
+            wxLogLastError(wxT("Unexpected pointer message"));
+            return false;
+    }
+
+    const UINT32 pointerId = GET_POINTERID_WPARAM(wParam);
+    POINTER_INPUT_TYPE pType = 0;
+    if (GetPointerType(pointerId, &pType))
```
vadz: We typically use explicit scope for Win32 functions to make it easier to distinguish them visually:


```suggestion
    if ( ::GetPointerType(pointerId, &pType) )
```

vadz (commented):
Thanks, this is a good start, but I have a couple of global questions in addition to the small remarks below.

1. Terminology: is the word "tablet" actually clear? Windows doesn't seem to use it (it uses only "pointer" AFAICS), GTK seems to use "pad" (but I'm not even sure if it's for the same thing) and Cocoa also seems to use "pointing device". So is it going to be really clear what this is about or should we, perhaps, do a global rename of "Tablet" to "Pointer" or something else?
2. I'm not sure if I had already mentioned this and sorry if I didn't, but I'd like to avoid another sample just for this, especially considering that currently it's completely useless under non-Windows systems. I think it would be better to add another frame to either `event` or `touchtest` sample instead.

undergraver:
> 1. Terminology: is the word "tablet" actually clear? 

Based on my investigation on both GTK https://docs.gtk.org/gdk3/enum.InputSource.html and Windows https://learn.microsoft.com/en-us/windows/win32/api/winuser/ne-winuser-tagpointer_input_type I believe that "tablet" is the right choice since I am filtering only messages from PT_PEN pointer type and the event information is held in a `POINTER_PEN_INFO` structure ( https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-pointer_pen_info ).

Of course, another solution would be to use the `pointer` terminology but I consider it too confusing.

LE: About the naming I am open to any suggestions. I also checked the Qt and they have a QTabletEvent and saw it is almost identical to what I have proposed.

> 2. .... I think it would be better to add another frame to either `event` or `touchtest` sample instead.

I am not against this approach; personally I found out the `touchtest` sample way too confusing for a copy paste user that wants something ready to modify. Based on recent experiences I would say that users want to go "straight to business" without filtering parts of source code to extract their desired code excerpt.

RobertRoeb:
Stylus?

samples/samples_vc17.sln
undergraver: I don't have the means to test this unfortunately.

undergraver:
> Stylus?

I would consider it as viable.

I personally would also think of `wxPenEvent` because "pen" seems to be found in both gtk/msw documentation.

samples/samples_vc17.sln
vadz: It's ok to change them blindly, they're almost identical to the MSVS 2022 version anyhow. But, of course, not adding a new sample would obviate this :-)

vadz:
> > 1. Terminology: is the word "tablet" actually clear?
> 
> Based on my investigation on both GTK https://docs.gtk.org/gdk3/enum.InputSource.html and Windows https://learn.microsoft.com/en-us/windows/win32/api/winuser/ne-winuser-tagpointer_input_type I believe that "tablet" is the right choice since I am filtering only messages from PT_PEN pointer type and the event information is held in a `POINTER_PEN_INFO` structure ( https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-pointer_pen_info ).
> 
> Of course, another solution would be to use the `pointer` terminology but I consider it too confusing.

I agree, and `wxPenEvent` is also confusing because there is `wxPen` which has nothing to do with this. But I do like Robert's suggestion, so why not `wxStylusEvent`?
 
> > 2. .... I think it would be better to add another frame to either `event` or `touchtest` sample instead.
> 
> I am not against this approach; personally I found out the `touchtest` sample way too confusing for a copy paste user that wants something ready to modify. Based on recent experiences I would say that users want to go "straight to business" without filtering parts of source code to extract their desired code excerpt.

This is true, but OTOH it's good to avoid having too many samples. I think having all tablet/stylus-related code in a single frame (shown by a menu command from the main one) would be acceptable.

undergraver:
> This is true, but OTOH it's good to avoid having too many samples. I think having all tablet/stylus-related code in a single frame (shown by a menu command from the main one) would be acceptable.



Alright, I agree then with `wxStylusEvent` but I would still like to have a separate sample.

Related to .sln files. Couldn't we generate them somehow with `cmake` or similar stuff, I find it dreadful to edit .sln files?

vadz:
> Alright, I agree then with wxStylusEvent

Just in case this advice might be useful: the way I do such global renames in new code is to make a patch (`git format-patch` or just `git diff`), load it into a text editor and do a search and replace on all new lines (in Vim with Abolish extension it's just `%g/^+/S/Table/Stylus/g`), then reset and apply the changed patch (`git am` or `git apply`). It's simpler/faster than doing it in all of the files.

> but I would still like to have a separate sample.

This is not a hill I'm going to die on, but it definitely feels strange to have a sample doing absolutely nothing on the platforms which don't implement support for generating these events.

BTW, we probably should define `wxHAS_STYLUS_EVENTS` for the platforms that do support them, to avoid programs using `__WXMSW__` for testing for this.

> Related to .sln files. Couldn't we generate them somehow with `cmake` or similar stuff, I find it dreadful to edit .sln files?

It's not much fun but I don't know of any tool that would be able to generate what we want, i.e. using the current structure, without making things even worse.

vadz:
Sorry, I got distracted by the releases, but now that they're done, I'd like to merge this — is this ready or do you plan more changes (seeing "phase 1" above)?

MaartenBent:
If it will be a separate sample, please add it to CMake as well (`wx_add_sample(tabletpen)` in `build/cmake/samples/CMakeLists.txt`, alphabetical order).

I think `samples.bkl` is also intended to be in alphabetical order.

undergraver:
I only have tested gtk3 sample which showed that stylus works also there so I might be able to make the port of wxGTK.

My questions is:

What are the other major ports that need this? I already played with a python QT sample and the offerings were similar.

undergraver:
I've updated the source code with:
* cmake for samples
* alphabetical order in .bkl file
* updated since to 3.3.3


I don't plan to add anything right now, but I might add support for other platforms, just let me know which

gtk can be developed on the machines I have access to from what I know.

let me know what other ports can be developed/tested on Linux / Windows and I'll do my best.

build/cmake/samples/CMakeLists.txt
```diff
@@ -139,6 +139,7 @@ wx_add_sample(statbar DEPENDS wxUSE_STATUSBAR)
 wx_add_sample(stc stctest.cpp edit.cpp prefs.cpp edit.h defsext.h prefs.h
     DATA stctest.cpp NAME stctest LIBRARIES wxstc DEPENDS wxUSE_STC)
 wx_add_sample(svg svgtest.cpp RES svgtest.rc DEPENDS wxUSE_SVG)
+wx_add_sample(tabletpen tabletpen.cpp)
```
vadz: Sorry but this needs to be renamed too for consistency, i.e.


```suggestion
wx_add_sample(stylus stylus.cpp)
```

include/wx/event.h
```diff
@@ -4327,6 +4377,7 @@ typedef void (wxEvtHandler::*wxTwoFingerTapEventFunction)(wxTwoFingerTapEvent&);
 typedef void (wxEvtHandler::*wxLongPressEventFunction)(wxLongPressEvent&);
 typedef void (wxEvtHandler::*wxPressAndTapEventFunction)(wxPressAndTapEvent&);
 typedef void (wxEvtHandler::*wxFullScreenEventFunction)(wxFullScreenEvent&);
+typedef void (wxEvtHandler::* wxStylusEventFunction)(wxStylusEvent&);
```
vadz: Extremely minor nitpick

```suggestion
typedef void (wxEvtHandler::*wxStylusEventFunction)(wxStylusEvent&);
```

interface/wx/event.h
```diff
@@ -5159,6 +5159,93 @@ class wxSetCursorEvent : public wxEvent
     */
     void SetCursor(const wxCursor& cursor);
 };
+/**
+    @class wxStylusEvent
+
+    A wxStylusEvent is generated from wxWindow when the graphical pen (stylus) is used.
+
+    This event gives the application the chance to perform specific tablet events
```
vadz: There is also still "tablet" here. "chance" seems a bit funny to me ("Your last chance to process stylus event, don't miss it!"), so I'd rephrase this as


```suggestion
    Application can handle this event to process stylus actions based on
    the current of the pen, pressure and other parameters.
```

samples/tabletpen/tabletpen.cpp
```diff
@@ -0,0 +1,254 @@
+/////////////////////////////////////////////////////////////////////////////
+// Name:        tabletpen.cpp
+// Purpose:     Tablet drawing sample
```
vadz:
```suggestion
// Name:        styles.cpp
// Purpose:     Stylus sample
```

samples/samples.bkl
```diff
@@ -77,6 +77,7 @@
     <subproject id="statbar" template="sub"/>
     <subproject id="stc" template="sub" cond="USE_STC=='1'"/>
     <subproject id="svg" template="sub"/>
+    <subproject id="tabletpen" template="sub"/>
```
vadz:
```suggestion
    <subproject id="stylus" template="sub"/>
```

vadz (commented):
Thanks for the updates, I'd just like to perform one last renaming, of the sample files themselves. Please let me know if you're going to do it or if I should do it myself.

Thanks!

undergraver:
I updated based on your suggestions. you can add more changes if needed.

Thanks!

vadz:
Thanks again! I'll merge this soon.

PR #26225, opened 2026-02-22
Cairo 1.18 fixes

### LINKED ISSUE #26224: wxQt: fails to link because function names are decorated with import library exports from Cairo.
### Build System Used

- [X] CMake
- [ ] configure
- [ ] nmake with makefile.vc
- [ ] Xcode

### Description

Testing the build will print this warning:
```
/usr/include/cairo/cairo.h:70:9: warning: "cairo_public" redefined
   70 | #define cairo_public _cairo_api extern
      |         ^~~~~~~~~~~~
src/common/cairo.cpp:19:9: note: this is the location of the previous definition
   19 | #define cairo_public
      |         ^~~~~~~~~~~~
```
and later the final link will fail with many messages like these ones:
```
[0/12] Linking CXX shared library lib/cygwx_qtu_core-3.2-0.dll
ld: libs/core/CMakeFiles/wxcore.dir/__/__/__/__/src/generic/graphicc.cpp.o: in function `wxCairoPathData::MoveToPoint(double, double)':
src/generic/graphicc.cpp:1293:(.text+0x1b7): undefined reference to `__imp_cairo_move_to'
ld: libs/core/CMakeFiles/wxcore.dir/__/__/__/__/src/generic/graphicc.cpp.o: in function `wxCairoPathData::AddLineToPoint(double, double)':
src/generic/graphicc.cpp:1298:(.text+0x1c7): undefined reference to `__imp_cairo_line_to'
ld: libs/core/CMakeFiles/wxcore.dir/__/__/__/__/src/generic/graphicc.cpp.o: in function `wxCairoPathData::~wxCairoPathData()':
src/generic/graphicc.cpp:1261:(.text+0x1e4): undefined reference to `__imp_cairo_destroy'
...etc
```
This error happens because the particular behaviour of DLLs.
This is how cairo handles the dllexport/dllimport modifiers:
https://cgit.freedesktop.org/cairo/tree/src/cairo.h#n53
```
#if (defined(_WIN32) || defined(__CYGWIN__)) && !defined(CAIRO_WIN32_STATIC_BUILD)
# define _cairo_export __declspec(dllexport)
# define _cairo_import __declspec(dllimport)
#elif defined(__GNUC__)
# define _cairo_export __attribute__((__visibility__("default")))
# define _cairo_import
#else
# define _cairo_export
# define _cairo_import
#endif

#ifdef CAIRO_COMPILATION
# define _cairo_api _cairo_export
#else
# define _cairo_api _cairo_import
#endif

#define cairo_public _cairo_api extern
```
So, declaring `cairo_public` doesn't avoid the use of dllexport/dllimport.
At the time of writing, I confirm that hacking the use of `CAIRO_WIN32_STATIC_BUILD` is the only way to build it.

However, it may be working in the past because a different behaviour of `cairo_public` into older versions of Cairo, so it must be inspected from which version this has changed and if such older versions are still able to be used by the current code.

- wxWidgets version  you are building: 3.2.9
- wxWidgets port  you are building: wxQt
- OS  and its version: CYGWIN/WIN32
- Compiler  being used: gcc-13.4.0
- Non-default compiler options, if any : None.

vadz:
Cairo 1.18 broke using cairo_public redefinition hack used before to avoid declaring Cairo functions as being dll-imported. Use the new way to prevent this from happening but still define cairo_public when using older versions.

Note that, unfortunately, we can't do it unconditionally because this results in a macro redefinition warning with newer versions which is impossible to disable using gcc.

Closes #26224.


vadz:
There is a simpler way to do this, see discussion in #26224.

PR #26226, opened 2026-02-22
wxScrolled<>:  fix testing mouse position in autoscroll zone

wsu-cb:
wxWindow::GetRect() does not always report client coordinates, so use wxWindow::GetScreenRect() and wxWindow::ClientToScreen() to test mouse and autoscroll zone with consistent coordinate system


vadz:
Thanks, will push soon.

PR #26227, opened 2026-02-23
Make `wxGetLinuxDistributionInfo()` available on all Unices

lanurmi:
Now that `wxGetLinuxDistributionInfo()` primarily obtains its data from `/etc/os-release`, there is nothing fundamentally Linux-specific about the implementation.

FreeBSD comes with `/etc/os-release` since FreeBSD 13.

### DIFF
--- a/include/wx/utils.h
+++ b/include/wx/utils.h
@@ -144,7 +144,7 @@ WXDLLIMPEXP_BASE wxString wxGetCpuArchitectureName();
 // Get native machine CPU architecture
 WXDLLIMPEXP_BASE wxString wxGetNativeCpuArchitectureName();
 
-#ifdef __LINUX__
+#ifdef __UNIX__
 // Get linux-distro information
 WXDLLIMPEXP_BASE wxLinuxDistributionInfo wxGetLinuxDistributionInfo();
 #endif
--- a/interface/wx/platinfo.h
+++ b/interface/wx/platinfo.h
@@ -120,7 +120,8 @@ enum wxEndianness
 };
 
 /**
-    A structure containing information about a Linux distribution.
+    A structure containing information about a Linux distribution,
+    or another Unix-style operating system (FreeBSD).
 
     See wxGetLinuxDistributionInfo() or wxPlatformInfo::GetLinuxDistributionInfo()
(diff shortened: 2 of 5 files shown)


vadz (changes requested):
Sorry, I don't understand how can you say that there is nothing Linux-specific about this function when it has "Linux" in its name (and also in the name of the struct returned by it).

If it makes sense to extend this to other Unix systems, we need to rename both the function and the struct (keeping the old names for compatibility, of course), otherwise this would be very confusing.

vadz:
Is this replaced by #26231?

lanurmi:
> Is this replaced by #26231?

No, these two are unrelated.

lanurmi:
> Sorry, I don't understand how can you say that there is nothing Linux-specific about this function when it has "Linux" in its name (and also in the name of the struct returned by it).

I said there is nothing _fundamentally_ Linux-specific about the implementation, and I can easily say that because "Linux" is just a name, and the implementation consist of reading text files from paths that aren't too Linux-specific.

> If it makes sense to extend this to other Unix systems, we need to rename both the function and the struct (keeping the old names for compatibility, of course), otherwise this would be very confusing.

Even if we ignore all the hassle of renaming something, the choice of a new name isn't easy or obvious. Something like `UnixDistributionInfo` sounds like pretending `/etc/os-release` doesn't originate from Linux. `OsInfo` would be too general, and the expected output would overlap a lot with `wxGetOsDescription()`/`wxGetOsVersion()` on Windows, macOS, and Unices that do not have `/etc/os-release`.

vadz:
I still think we need a new function name, returning FreeBSD version from `wxGetLinuxDistributionInfo()` is just completely unexpected and I would be hard-pressed to explain it to any confused users.

I would probably just go with `wxOsDistributionInfo` and `GetOsDistributionInfo()` and try to fill it under as many platforms as possible (e.g. I could see this being useful under macOS too actually) and document that it doesn't return anything useful for the platforms where we have nothing to put in it.

But, again, I'd rather not apply this in the current state, sorry.

vadz:
I don't think it is useful to leave this open any longer.

PR #26231, opened 2026-02-23
Make Linux distribution `wxGetOsDescription()`'s main output

lanurmi:
When available, acquire the name and release of the Linux distribution and return it as `wxGetOsDescription()`'s output, followed by the near equivalent of the traditional `uname -srm` output.

The uname output, something like `Linux 6.19.0-5-generic x86_64`, is not the best description for a Linux system, since it leaves up to the reader to figure out what kind of a system this actually is.

The output changes significantly from what it was earlier, but the output was never meant to be machine-readable, nor was it consistent across different operating systems.

### DIFF
--- a/interface/wx/utils.h
+++ b/interface/wx/utils.h
@@ -843,7 +843,9 @@ bool wxGetUserName(char* buf, int sz);
 /**
     Returns the string containing the description of the current platform in a
     user-readable form. For example, this function may return strings like
-    "Windows 10 (build 10240), 64-bit edition" or "Linux 4.1.4 i386".
+    "Windows 10 (build 10240), 64-bit edition",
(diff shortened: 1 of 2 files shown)


vadz (changes requested):
Thanks, I agree with the idea but:

1. Could you please update this functions documentation in `interface/wx/utils.h` with an example of newly returned string?
2. Use normal wx `camelCase` convention for the variable names instead of capitalizing many of them?

Thanks!

vadz:
Thanks, but I don't understand why this one should be Linux-specific — there is no problem with trying to read `os-release` under other Unix systems if it's present, is there?

I've pushed a commit with my changes enabling this for not just Linux and also a few more renamings as I had trouble reading `unFoo` as anything other than "anti Foo".

What do you think?

vadz:
@lanurmi I'm going to merge this soon if you have no objections.

lanurmi:
It's fine, no objections.

PR #26235, opened 2026-02-23
Scale wxAuiToolBar packing, padding and margins by DPI factor

### LINKED ISSUE #26076: wxAuiToolBar should update packing/padding/margins on DPI change on MSW
### Description

`wxAuiToolBar::Create` contains initialization of some packing/padding/margins values which are not updated on DPI change. Would be great to have them updated on DPI change.

https://github.com/wxWidgets/wxWidgets/blob/1b582af60d99388b56b383fe240b647387a5e4b4/src/aui/auibar.cpp#L954-L967

Probably needs some flags to detect if user calls `SetToolPacking`/`SetToolBorderPadding`/`SetMargins` to disable these automatic updates.

This only matters on MSW.

- wxWidgets version  you use: master (3.3)

vadz:
Store these values in DIPs internally and apply FromDIP() when using them.

This is simpler and possibly more precise than rescaling them when getting wxEVT_DPI_CHANGED event and the only awkward part is that we need to use ToDIP() in the setter functions, but this is arguably not very important as these functions are hardly ever used anyhow.

Closes #26076.

### DIFF
--- a/src/aui/auibar.cpp
+++ b/src/aui/auibar.cpp
@@ -951,8 +951,9 @@ bool wxAuiToolBar::Create(wxWindow* parent,
 
     m_windowStyle = style;
 
-    m_toolPacking = FromDIP(2);
-    m_toolBorderPadding = FromDIP(3);
+    // Tool packing and padding are in DIPs, i.e. do not use FromDIP() here.
+    m_toolPacking = 2;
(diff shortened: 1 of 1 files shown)


dsa-t:
Looks ok to me, thanks!

vadz:
Applied as the commit above.

PR #26237, opened 2026-02-24
Support drag-and-drop autoscroll

wsu-cb:
As stated in https://github.com/wxWidgets/wxWidgets/pull/25978#discussion_r2595800569, I want to implement autoscroll for drag-and-drop operations, and https://github.com/wxWidgets/wxWidgets/blob/61300d78d215e576576e7dc951f3147ef1785db2/samples/dnd/d_and_d.txt#L163 implies that I'm not the only one who wants this feature.

Now that the `wxScrolled<>` autoscroll region is configurable, this PR is a suggestion for another `wxScrolled<>` extension to support autoscrolling during drag-and-drop.  Commit 905430223af2482eb0e7ac00eb2ce9a6ab6ca551 implements the support, and commit e2439c10f4ac8a30cd93d51e717a44877ef92bed modifies the `dnd` sample to demonstrate dragging a `DnDShape` from one `DnDShapeFrame` to another `DnDShapeFrame` with the destination `DnDShapeFrame` autoscrolling.

I realize that commit e2439c10f4ac8a30cd93d51e717a44877ef92bed leaves the sample with some imperfections, but I think the current form is sufficient to demonstrate the API.  If the API is approved, then I will go back and refine the sample.

### DIFF
--- a/build/cmake/samples/CMakeLists.txt
+++ b/build/cmake/samples/CMakeLists.txt
@@ -35,13 +35,13 @@ endif()
 wx_add_sample(dialogs ${SAMPLE_DIALOGS_SRC} DATA tips.txt)
 wx_add_sample(dialup nettest.cpp LIBRARIES wxnet DEPENDS wxUSE_DIALUP_MANAGER)
 wx_add_sample(display)
-wx_add_sample(dnd dnd.cpp RES dnd.rc DATA wxwin.png DEPENDS wxUSE_DRAG_AND_DROP)
+wx_add_sample(dnd dnd.cpp RES dnd.rc DATA wxwin.png DEPENDS wxUSE_DRAG_AND_DROP IMPORTANT)
 wx_add_sample(docview docview.cpp doc.cpp view.cpp docview.h doc.h view.h
-    RES docview.rc PLIST Info.plist.in LIBRARIES wxaui DEPENDS wxUSE_DOC_VIEW_ARCHITECTURE)
+    RES docview.rc PLIST Info.plist.in LIBRARIES wxaui DEPENDS wxUSE_DOC_VIEW_ARCHITECTURE IMPORTANT)
 wx_add_sample(dragimag dragimag.cpp dragimag.h RES dragimag.rc
     DATA backgrnd.png shape01.png shape02.png shape03.png
     DEPENDS wxUSE_DRAGIMAGE)
-wx_add_sample(drawing DATA pat4.bmp pat35.bmp pat36.bmp image.bmp mask.bmp NAME drawingsample)
+wx_add_sample(drawing DATA pat4.bmp pat35.bmp pat36.bmp image.bmp mask.bmp NAME drawingsample IMPORTANT)
 wx_add_sample(erase)
 wx_add_sample(event event.cpp gestures.cpp gestures.h chessboard.cpp chessboard.h)
 wx_add_sample(except DEPENDS wxUSE_EXCEPTIONS)
--- a/build/cmake/samples/CMakeLists.txt
+++ b/build/cmake/samples/CMakeLists.txt
@@ -37,11 +37,11 @@ wx_add_sample(dialup nettest.cpp LIBRARIES wxnet DEPENDS wxUSE_DIALUP_MANAGER)
 wx_add_sample(display)
 wx_add_sample(dnd dnd.cpp RES dnd.rc DATA wxwin.png DEPENDS wxUSE_DRAG_AND_DROP)
 wx_add_sample(docview docview.cpp doc.cpp view.cpp docview.h doc.h view.h
-    RES docview.rc PLIST Info.plist.in LIBRARIES wxaui DEPENDS wxUSE_DOC_VIEW_ARCHITECTURE IMPORTANT)
+    RES docview.rc PLIST Info.plist.in LIBRARIES wxaui DEPENDS wxUSE_DOC_VIEW_ARCHITECTURE)
 wx_add_sample(dragimag dragimag.cpp dragimag.h RES dragimag.rc
     DATA backgrnd.png shape01.png shape02.png shape03.png
     DEPENDS wxUSE_DRAGIMAGE)
-wx_add_sample(drawing DATA pat4.bmp pat35.bmp pat36.bmp image.bmp mask.bmp NAME drawingsample IMPORTANT)
+wx_add_sample(drawing DATA pat4.bmp pat35.bmp pat36.bmp image.bmp mask.bmp NAME drawingsample)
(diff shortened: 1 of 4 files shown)


wsu-cb:
> If the API is approved, then I will go back and refine the sample.

And add documentation to `interface/wx/scrolwin.h`.

build/cmake/samples/CMakeLists.txt
```diff
@@ -35,13 +35,13 @@ endif()
 wx_add_sample(dialogs ${SAMPLE_DIALOGS_SRC} DATA tips.txt)
 wx_add_sample(dialup nettest.cpp LIBRARIES wxnet DEPENDS wxUSE_DIALUP_MANAGER)
 wx_add_sample(display)
-wx_add_sample(dnd dnd.cpp RES dnd.rc DATA wxwin.png DEPENDS wxUSE_DRAG_AND_DROP)
+wx_add_sample(dnd dnd.cpp RES dnd.rc DATA wxwin.png DEPENDS wxUSE_DRAG_AND_DROP IMPORTANT)
```
vadz: I'm not sure we want to do this for this one, of all the modified samples I think only `drawing` potentially deserves being "IMPORTANT".

I'd make it and `dialogs` important and remove it from `dataview`, but I'm not really sure.

Any (other) opinions?

samples/dnd/dnd.cpp
```diff
@@ -756,13 +756,27 @@ class DnDShapeDropTarget : public wxDropTarget
...
 #if wxUSE_STATUSBAR
         m_frame->SetStatusText("Mouse left the frame");
 #endif // wxUSE_STATUSBAR
+        m_frame->DisableAutoscrollWithoutCapture();
+    }
+    virtual wxDragResult OnDragOver(wxCoord x, wxCoord y, wxDragResult def) override
+    {
+        wxMouseEvent event(wxEVT_MOTION);
+        event.SetPosition(wxPoint(x, y));
+        m_frame->ProcessWindowEvent(event);
```
vadz: I'd really like to avoid having to do it in application code, e.g. by adding some `InformMousePositionForAutoscroll()` or something with better name.

The rationale is that we can't guarantee that wx windows always react to events in the expected way (e.g. native controls do not and can't), so we advise to never do this and so this is showing a bad example.

include/wx/scrolwin.h
```diff
@@ -168,6 +168,18 @@ class WXDLLIMPEXP_CORE wxScrollHelperBase : public wxAnyScrollHelperBase
...
 
+    // wxWidgets <= 3.3.1 mostly restricted autoscroll to the
+    // window holding the mouse capture.  However, when dragging
+    // objects between windows, the destination window should be
+    // the window that is scrolling (to allow positioning the
+    // dragged object).  The intent is that a window will call
+    // EnableAutoscrollWithoutCapture() when processing a
+    // drag-enter, and DisableAutoscrollWithoutCapture() when
+    // processing a drag-exit or drag-drop.
+    bool GetAutoscrollWithoutCapture() const;
```
vadz, line 179: Does this have to be public? It looks like it's only used (and useful) internally?

vadz (commented):
Thanks, the implementation looks good but I'm not sure about the API. Could it be better to have an object which enable autoscroll without capture in its ctor, disables it in its dtor and has a member function for informing the window about the mouse position change? Not sure if it's worth it, but I'd like to at least have such a function somewhere, i.e. either in this object or in `wxScrolled` itself.

build/cmake/samples/CMakeLists.txt
wsu-cb: Oops.  I turned on `IMPORTANT` to configure the sample to build on my local system, but I didn't intend to commit this.

Undone in https://github.com/wxWidgets/wxWidgets/pull/26237/commits/4e79f0721f843cfa5234e686f9b743c6f3b31d69.

include/wx/scrolwin.h
wsu-cb: It's declared in `wxScrollHelperBase`, but it's called from both `wxScrollHelperBase` and `wxAutoScrollTimer`.  Similar to https://github.com/wxWidgets/wxWidgets/pull/25978#discussion_r2804027627, after the desired functionality is complete, we can discuss `public` vs `friend`.

samples/dnd/dnd.cpp
wsu-cb: MSW has `TTM_RELAYEVENT`.  How about `void RelayPosition(wxPoint pt)`?  Or do you think `Autoscroll` is a necessary part of the name?  `AutoscrollRelayPosition`?

wsu-cb:
> Thanks, the implementation looks good but I'm not sure about the API. Could it be better to have an object which enable autoscroll without capture in its ctor, disables it in its dtor and has a member function for informing the window about the mouse position change? Not sure if it's worth it, but I'd like to at least have such a function somewhere, i.e. either in this object or in `wxScrolled` itself.

Can you explain how you think this object would be used?  Since the enable and disable are not called within a single function invocation, I don't see the advantage to RAII-style.

samples/dnd/dnd.cpp
wsu-cb: Actually, now that I look at this again, I don't think there's any reason for this code.  The `wxAutoScrollTimer::Notify()` takes care of tracking the mouse position.

Unnecessary code removed in https://github.com/wxWidgets/wxWidgets/pull/26237/commits/64eab60005bc86c7000267d1572729cfa21d907d

vadz:
> Can you explain how you think this object would be used? Since the enable and disable are not called within a single function invocation, I don't see the advantage to RAII-style.

Yes, as I wrote, I wasn't sure if it was really worth it, and with the code for forwarding mouse events removed, I'm even less sure now (i.e. I don't think it's worth it any more). But, for the record, this could be still used by using `std::unique_ptr<>` and calling `reset()` on it — not automatic, as with RAII, but still maybe a bit more convenient than remembering to call specific initialization/cleanup functions.

Anyhow, I'll merge this as is now, thanks!

wsu-cb:
> Anyhow, I'll merge this as is now, thanks!

This can be merged now without causing any problems.  However, as noted in earlier comments, I plan to improve the implementation of the `dnd` sample and the `interface` documentation, so I suggest waiting to merge this until I finish that work.

vadz:
Oops, sorry, I had missed that comment and already pushed 2484d14ec6 (wxScrolled<>: add support for autoscroll during drag, 2026-02-22) which merged this, but forgot to mention the PR number in its commit message.

I'll close it now, please open a new PR for further improvements, sorry for the inconvenience...

PR #26238, opened 2026-02-24
Add check for config like in wxHtmlHelpController::ReadCustomization

mcorino:
This way the used config can be safely reset to `nullptr` as clearly the intention looking at wxHtmlHelpController.

Currently a segfault occurs when calling `wxHtmlHelpController::UseConfig(nullptr)` (or `wxHtmlHelpWindow::UseConfig(nullptr)` directly).

### DIFF
--- a/include/wx/html/helpwnd.h
+++ b/include/wx/html/helpwnd.h
@@ -132,7 +132,8 @@ public:
         {
             m_Config = config;
             m_ConfigRoot = rootpath;
-            ReadCustomization(config, rootpath);
+            if (config)
+                ReadCustomization(config, rootpath);
         }
 
     // Saves custom settings into cfg config. it will use the path 'path'


mcorino:
@vadz It would be nice if this could still be included in 3.3.2 as this causing me grieve in wxRuby3 regression tests for the Wx::ConfigBase classes. A 'normal' app would probably not muck about with setting / resetting the global configuration instance and mixing that with HtmlHelpController tests etc. but the tests do and the API allows it so it would seem prudent to be able to guard against that. Unfortunately the only option currently to protect the HtmlHelpController against having a reference to an non-existing configuration instance at the moment it is destroyed (when it will attempt to save it's custom config) is to explicitly reset it's config instance to `nullptr` if not for the problem with HtmlHelpWindow.

vadz:
This looks safe to apply, so I'll do it soon, thanks.

I'll also update the docs to explicitly mention that `nullptr` can be passed here, although I'm not quite sure about the implications: the docs for the same member function in `wxHtmlHelpController` say

```
        If you do not use UseConfig(), wxHtmlHelpController will use the default
        wxConfig object if available (for details see wxConfigBase::Get and
        wxConfigBase::Set).
```

but here it looks like the use of config is completely disabled instead?

mcorino:
If UseConfig is called with nullptr after the help windows have been created by the controller the effect will indeed be that config use is disabled.On 24 Feb 2026 16:47, VZ ***@***.***> wrote:vadz left a comment (wxWidgets/wxWidgets#26238)
This looks safe to apply, so I'll do it soon, thanks.
I'll also update the docs to explicitly mention that nullptr can be passed here, although I'm not quite sure about the implications: the docs for the same member function in wxHtmlHelpController say
        If you do not use UseConfig(), wxHtmlHelpController will use the default
        wxConfig object if available (for details see wxConfigBase::Get and
        wxConfigBase::Set).

but here it looks like the use of config is completely disabled instead?

PR #26239, opened 2026-02-24
Add new Print function to wxWebView that takes custom printer settings

Blake-Madden:
Supports Edge, GTK, and macOS backends. IE and Chromium backends silently fallback to default `Print()`.

Tested under Windows 11, macOS Tahoe, and Linux Mint.


Blake-Madden:
Under Linux Mint, I noticed that printing as landscape to a PDF renders blank pages, but I get the same effect when calling the original `Print()` function too. I'll explore that issue separate from this.

include/wx/webview.h
```diff
@@ -266,6 +267,9 @@ class WXDLLIMPEXP_WEBVIEW wxWebView : public wxControl
     virtual bool IsEditable() const = 0;
     virtual void LoadURL(const wxString& url) = 0;
     virtual void Print() = 0;
+#if wxUSE_PRINTING_ARCHITECTURE
+    virtual void Print(const wxPrintData& printData, bool showHeaderFooter = false);
```
vadz: Sorry, as usual, I [hate boolean parameters](https://wxwidgets.org/develop/coding-guidelines/#no_bool_params), please replace this with an enum or `int flags`.

vadz (commented):
Thanks, this looks useful and safe to merge even with little time remaining before 3.3.2, but I'd really like to avoid committing to the API with `bool header` parameter.

Another small thing: it doesn't make much sense to me to both provide the default implementation for the new function and manually forward it to the base class in the backends which don't implement it. Either make the new function pure virtual or, simpler, just remove `Print(wxPrintData)` overrides from the backends that don't do anything useful there and use a using directive to avoid virtual function hiding warnings.

Thanks!

include/wx/webview.h
Blake-Madden: Done.

include/wx/msw/webview_ie.h
```diff
@@ -96,6 +96,7 @@ class WXDLLIMPEXP_WEBVIEW wxWebViewIE : public wxWebView
     virtual bool CanSetZoomType(wxWebViewZoomType) const override;
 
     virtual void Print() override;
+    using wxWebView::Print;
```
MaartenBent, line 99: Are these `using` statements needed? IE/Chromium seem to build fine for me without it.
I don't know about the one in `webkit`.

vadz: I think you'd get warnings about virtual function hiding without it.

Blake-Madden: Yes, indeed. I was getting errors about hidden virtual functions without this (from GCC I think).

vadz (commented):
Thanks for the update, but 2 more questions:

1. According to [the docs](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2printsettings?view=webview2-1.0.3719.77#get_shouldprintheaderandfooter) the default for `WebView` is not to print the header and the footer and because this is also the behaviour under the other platforms, I think it makes sense to default to it under Windows too.
2. The new enum needs to be documented in `interface` file as otherwise we'll probably break wxPython build.

vadz:
P.S. It's probably better, finally, to postpone this until 3.3.3, unless you really, really want to have it in 3.3.2.

Blake-Madden:
>     1. According to [the docs](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2printsettings?view=webview2-1.0.3719.77#get_shouldprintheaderandfooter) the default for `WebView` is not to print the header and the footer and because this is also the behaviour under the other platforms, I think it makes sense to default to it under Windows too.
>     2. The new enum needs to be documented in `interface` file as otherwise we'll probably break wxPython build.

Done.

Blake-Madden:
> P.S. It's probably better, finally, to postpone this until 3.3.3, unless you really, really want to have it in 3.3.2.

That's fine, not a critical issue.

I have `@since 3.3.2` in the interface file, so if you decide for this to be 3.3.3 we would just need to update that.

include/wx/msw/webview_edge.h
```diff
@@ -68,6 +68,9 @@ class WXDLLIMPEXP_WEBVIEW wxWebViewEdge : public wxWebView
     virtual bool CanSetZoomType(wxWebViewZoomType type) const override;
 
     virtual void Print() override;
+#if wxUSE_PRINTING_ARCHITECTURE
+    virtual void Print(const wxPrintData& printData, int flags = wxWEBVIEW_PRINT_HIDE_HEADER_FOOTER) override;
```
vadz, line 72: I think we should add a using declaration here too, if only for consistency:

```suggestion
    virtual void Print(const wxPrintData& printData, int flags = wxWEBVIEW_PRINT_HIDE_HEADER_FOOTER) override;
    using wxWebView::Print;
```

include/wx/osx/webview_webkit.h
```diff
@@ -51,6 +51,9 @@ class WXDLLIMPEXP_WEBVIEW wxWebViewWebKit : public wxWebView
     virtual void Stop() override;
 
     virtual void Print() override;
+#if wxUSE_PRINTING_ARCHITECTURE
+    virtual void Print(const wxPrintData& printData, int flags = wxWEBVIEW_PRINT_HIDE_HEADER_FOOTER) override;
```
vadz, line 55: Same as above:


```suggestion
    virtual void Print(const wxPrintData& printData, int flags = wxWEBVIEW_PRINT_HIDE_HEADER_FOOTER) override;
    using wxWebView::Print;
```

interface/wx/webview.h
```diff
@@ -172,18 +172,33 @@ enum wxWebViewIE_EmulationLevel
...
     /** Other browsing data like history, settings, auto fill, passwords, etc. */
     wxWEBVIEW_BROWSING_DATA_OTHER = 0x08,
     /** All browsing data, including data corresponding to all the other constants. */
     wxWEBVIEW_BROWSING_DATA_ALL = 0x0f
 };
 
+/**
+    Types of settings that can be applied to print operations.
+
+    @since 3.3.2
```
vadz: Just a reminder:

```suggestion
    @since 3.3.3
```

interface/wx/webview.h
```diff
@@ -1252,11 +1267,40 @@ class wxWebView : public wxControl
...
+        settings including header/footer control. The GTK and macOS backends
+        support paper size, orientation, and copies but ignore @a flags.
+        The IE and Chromium backends fall back to Print().
+
+        @param printData
+            The print settings to use.
+        @param flags
+            A combination of wxWebViewPrintFlags values.
+
+        @since 3.3.2
```
vadz:
```suggestion
        @since 3.3.3
```

src/gtk/webview_webkit2.cpp
```diff
@@ -1418,6 +1418,112 @@ void wxWebViewWebKit::Print()
...
+    "om_folio",      // wxPAPER_FOLIO
+    "na_quarto",     // wxPAPER_QUARTO
+    "na_10x14",      // wxPAPER_10X14
+    "na_ledger",     // wxPAPER_11X17
+};
+
+GtkPaperSize* wxWebViewGetGtkPaperSize(wxPaperSize paperId)
+{
+    // Use the named GTK paper size if we have a mapping
+    if (paperId > 0 && static_cast<size_t>(paperId) < sizeof(gs_webviewPaperList)/sizeof(gs_webviewPaperList[0]))
```
vadz: Minor, but we typically use a macro for this:

```suggestion
    if (paperId > 0 && static_cast<size_t>(paperId) < WXSIZEOF(gs_webviewPaperList))
```

src/gtk/webview_webkit2.cpp
```diff
@@ -1418,6 +1418,112 @@ void wxWebViewWebKit::Print()
...
+}
+
+} // anonymous namespace
+
+void wxWebViewWebKit::Print(const wxPrintData& printData, int WXUNUSED(flags))
+{
+    WebKitPrintOperation* printop = webkit_print_operation_new(m_web_view);
+
+    // Use GtkPageSetup for paper size and orientation
+    GtkPageSetup* pageSetup = gtk_page_setup_new();
```
vadz: I wonder if some of the code here couldn't be shared with `src/gtk/print.cpp`? E.g. have some helper functions in `include/wx/gtk/private/print.h` (to be created) and call them from both places?

src/gtk/webview_webkit2.cpp
```diff
@@ -1418,6 +1418,112 @@ void wxWebViewWebKit::Print()
...
+    }
+
+    gtk_print_settings_set_use_color(settings, printData.GetColour());
+
+    webkit_print_operation_set_print_settings(printop, settings);
+
+    webkit_print_operation_run_dialog(printop, nullptr);
+
+    g_object_unref(settings);
+    g_object_unref(printop);
```
vadz: Using `wxGtkObject` wrapper would be better than calling those manually.

src/osx/webview_webkit.mm
```diff
@@ -461,12 +461,90 @@ function createClass(name,rules) { \
         // in my tests, the progress bar always freezes and it stops the whole
         // print operation. do not turn this to true unless there is a
         // workaround for the bug.
-        [op setShowsProgressPanel: false];
+        [op setShowsProgressPanel: YES];
```
vadz: This seems to be unrelated to the rest of the changes and goes against the comment just above. If the bug was fixed in macOS, the comment should probably be just removed.

vadz (commented):
Thanks, this mostly looks good but the pattern of falling back to calling `Print(void)` when there is an error in the new `Print()` overload is a bit annoying, I think it could be better to have some private `bool PrintWithSettings(wxPrintData)` (or whatever you prefer to call it) and call `Print(void)` from the base class `Print(wxPrintData)` if `PrintWithSettings()` fails. But well, this is minor.

Let me know if you plan any more changes here, otherwise I could apply the suggestions and merge myself.

src/osx/webview_webkit.mm
Blake-Madden: It says `YES` now.

vadz: Err, I don't understand your reply. You've changed it from false to true (`YES` is same as true), while the comment above explicitly cautions against doing it. Maybe the comment is out of date, but in this case it should be removed.

src/gtk/webview_webkit2.cpp
Blake-Madden: done

src/gtk/webview_webkit2.cpp
Blake-Madden: Might be best as a separate (future) PR. Seems like a really deep change.

Blake-Madden:
@vadz I'm all all done making changes.

src/osx/webview_webkit.mm
Blake-Madden: Sorry, I confused myself. I set this back to NO (false) to play it safe. YES was stable in my tests, but best to keep with the safer logic just in case.

vadz:
Thanks for the updates, will merge soon.

PR #26241, opened 2026-02-24
Fix wxGTK gesture handling

filip-hejsek:
This PR fixes several issues with gesture handling in wxGTK. The most severe issue is #26211, which is that gesture cancellation can result in spurious gesture end events. Another issue is that gesture handling used several global variables, even though there can be multiple independent gestures of the same type active in different windows.

I have tested the changes with both a touchpad and a touchscreen and everything seems to work perfectly. Since I don't have an actual touchscreen, I did the testing by configuring my system to recognize my touchpad as a touchscreen. I have tested:
- GNOME Wayland and XWayland with touchpad and touchscreen
- Weston Wayland and XWayland with touchscreen (Weston doesn't seem to support touchpad gestures)
- Xorg with touchpad and touchscreen

I was able to trigger two pan gestures at once in Weston by modifying the sample to open two windows, and using a touchscreen in one window and a mouse in the other. The existing code was horribly broken in this scenario, this PR fixes it.

### DIFF
--- a/src/gtk/window.cpp
+++ b/src/gtk/window.cpp
@@ -3612,6 +3612,8 @@ rotate_gesture_begin_callback(GtkGesture* gesture, GdkEventSequence* WXUNUSED(se
     event.SetPosition(wxPoint(wxRound(x), wxRound(y)));
     event.SetGestureStart();
 
+    data->m_lastAngleDelta = 0;
+
     // Save this point because the point obtained through gtk_gesture_get_bounding_box_center()
     // in the "end" signal is not a rotation center
     data->m_lastGesturePoint = wxPoint(wxRound(x), wxRound(y));
--- a/src/gtk/window.cpp
+++ b/src/gtk/window.cpp
@@ -428,7 +428,7 @@ static double gs_lastOffset = 0;
 static gdouble gs_lastScale = 1.0;
 
 // This is used to set the angle when rotate gesture ends.
-static gdouble gs_lastAngle = 0;
+static gdouble gs_lastAngleDelta = 0;
 
 // Last Zoom/Rotate gesture point
 static wxPoint gs_lastGesturePoint;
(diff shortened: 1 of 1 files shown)


src/gtk/window.cpp
```diff
@@ -3593,31 +3599,45 @@ extern "C" {
...
         return;
     }
 
+
     wxRotateGestureEvent event(win->GetId());
 
     event.SetEventObject(win);
     event.SetPosition(wxPoint(wxRound(x), wxRound(y)));
     event.SetGestureStart();
+    event.SetRotationAngle(0);
```
vadz: Not very important, but this is the default, why call it explicitly?

vadz (approved):
Thanks, I'm unable to test this (if anybody else can, please leave a comment!), but the changes look good to me.

Ideal would be to split this into several commits, e.g. one for moving `gs_xxx` into `wxWindowGesturesData`, maybe another one for the angle confusion fix and then the other changes, so perhaps I'll try doing this when applying it. But if I fail to do it quickly, I'll just apply this as is as it seems important to have it in 3.3.2.

Thanks again!

src/gtk/window.cpp
filip-hejsek: To make it extra clear that the value is the same as what we assign to `m_lastAngleDelta`. Fell free to remove this (along with `event.SetZoomFactor(1);` in the zoom callback).

filip-hejsek: Removed now.

filip-hejsek:
> Ideal would be to split this into several commits, e.g. one for moving `gs_xxx` into `wxWindowGesturesData`, maybe another one for the angle confusion fix and then the other changes, so perhaps I'll try doing this when applying it. But if I fail to do it quickly, I'll just apply this as is as it seems important to have it in 3.3.2.

I did the extra work to separate the changes. I didn't at all test the intermediate states this created though.

vadz:
Nice, thanks a lot for doing this! Will merge soon but probably not today.

PR #26252, opened 2026-02-27
Let wxAuiToolBars with stretchable elements stretch

vadz:
Previously, all fixed panes without a minimum size were forced to have a proportion of 0 in wxAUI layout code, which meant that a wxAuiToolBar with stretchable elements (typically spaces, but labels and controls could be stretchable too) couldn't stretch unless its pane was given some non-default minimum size, even if it was just wxSize(-1,1).

This was completely undiscoverable and not intuitive at all, so let wxAuiToolBar stretch if it has any stretchable elements, regardless of its pane min size instead.

This required adding wxAuiToolBar::CanStretch() and using an ugly dynamic cast. We could probably avoid this by using some kind of heuristic based on e.g. comparing min and best size, but this is simple and guaranteed not to break anything else, so do it like this for now.

Closes #26249.

### DIFF
--- a/include/wx/aui/auibar.h
+++ b/include/wx/aui/auibar.h
@@ -636,6 +636,9 @@ public:
     // Override to call DoIdleUpdate().
     virtual void UpdateWindowUI(long flags = wxUPDATE_UI_NONE) override;
 
+    // This function is for internal use only, don't call it from your code.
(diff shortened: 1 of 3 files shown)


vadz:
Merged in the commit above.

PR #26253, opened 2026-02-28
Add wxRibbonMSWFlatArtProvider, a (Fluent UI) flat ribbon art provider

Blake-Madden:
Also, add pointer and underflow checks and proper `wxDCClipper` usage in other art providers, add descriptions for all of them to the help.

Old MSW provider:

image

New one:

image

### DIFF
--- a/src/ribbon/art_msw.cpp
+++ b/src/ribbon/art_msw.cpp
@@ -1449,6 +1449,9 @@ void wxRibbonMSWArtProvider::DrawTab(
         }
     }
 
+    if ( tab.page == nullptr )
+        return;
+
     if(m_flags & wxRIBBON_BAR_SHOW_PAGE_ICONS)
     {
         wxBitmap icon = tab.page->GetIcon();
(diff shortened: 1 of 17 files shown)


Blake-Madden:
I'm not wild about the hard coded numbers throughout all three art providers, but that will be a separate battle for another day...

src/ribbon/art_msw.cpp
```diff
@@ -1948,7 +1951,7 @@ void wxRibbonMSWArtProvider::DrawPanelBackground(
             {
                 // Room for some characters and ...
                 // Display as many characters as possible and append ...
-                for(size_t len = label.Len() - 1; len >= 3; --len)
+                for( int len = static_cast<int>(label.Len()) - 1; len >= 3; --len )
```
vadz, line 1954: Not sure why this had to be changed, but as long as it did, I'll at least use `wxSsize()` here instead of the cast.

vadz (approved):
Thanks, I don't use ribbon myself but I guess it's a nice possibility for those who do, so I'll merge this soon.

PR #26254, opened 2026-02-28
Fix wxWebViewChromium initialization

MaartenBent:
The wxWebViewChromiumModule did not load, this got broken by #25911

### DIFF
--- a/src/common/webview.cpp
+++ b/src/common/webview.cpp
@@ -15,6 +15,7 @@
 #include "wx/webview.h"
 #include "wx/filesys.h"
 #include "wx/mstream.h"
+#include "wx/link.h"
 #include "wx/private/webview.h"
 
 #if defined(__WXOSX__)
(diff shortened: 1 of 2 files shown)


MaartenBent:
Note this only affects a static build. I guess without the `extern` declaration inside `webview_chromium.cpp`, nothing is loaded from the file and it won't be linked in, thus the module will never be created.

vadz:
If I understand the problem correctly, I think it would be better (because this would be more explicit and more clear) to solve it by using the macros from `wx/link.h` as usual, i.e. add `wxFORCE_LINK_THIS_MODULE(WebViewChromium)` to `src/common/webview_chromium.cpp` and `wxFORCE_LINK_MODULE(WebViewChromium)` inside `#if` check to `src/common/webview.cpp`.

Could you please check if doing this works?

MaartenBent:
That works too, and is indeed a clearer solution.

vadz:
Great, thanks, will merge and push soon.

vadz:
Applied as 4fdc4be96e5d5f795d65825e2db74f141db21a41 (forgot to mention this issue in the commit message, sorry). Thanks again for fixing this!

PR #26255, opened 2026-03-01
Don't unconditionally set m_clipping to true in wxMSWDCImpl::SetHDC()

AliKet:
Account for application-defined clipping regions only (those identified by the SelectClipRgn function) when setting m_clipping to true. IOW, clipping regions that are set by the system during BeginPaint are not considered.

Closes #26251

### DIFF
--- a/src/msw/dc.cpp
+++ b/src/msw/dc.cpp
@@ -339,6 +339,37 @@ void wxMSWDCImpl::SelectOldObjects(WXHDC dc)
     m_selectedBitmap = wxNullBitmap;
 }
 
+void wxMSWDCImpl::SetHDC(WXHDC dc, bool bOwnsDC)
+{
+    m_hDC = dc;
+    m_bOwnsDC = bOwnsDC;
+
+    m_isClipBoxValid = false;
+
+    if ( m_hDC != 0 )
+    {
+        // Notice that m_clipping is initialized to true only if there is
+        // an application-defined clipping region already set on the device.
+        // Quoting the MSDN:
+        //  "An application-defined clipping region is a clipping region
+        //   identified by the SelectClipRgn function. It is not a clipping
+        //   region created when the application calls the BeginPaint function."
+        //
+        HRGN hrgn;
+        int result = ::GetClipRgn((WXHDC)m_hDC, hrgn);
+        if ( result == -1 )
+        {
+            wxLogLastError("GetClipRgn");
+        }
+        else
+        {
+            // We might have a pre-existing clipping region; make sure that we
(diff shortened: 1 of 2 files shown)


src/msw/dc.cpp
```diff
@@ -339,6 +339,37 @@ void wxMSWDCImpl::SelectOldObjects(WXHDC dc)
...
+    {
+        // Notice that m_clipping is initialized to true only if there is
+        // an application-defined clipping region already set on the device.
+        // Quoting the MSDN:
+        //  "An application-defined clipping region is a clipping region
+        //   identified by the SelectClipRgn function. It is not a clipping
+        //   region created when the application calls the BeginPaint function."
+        //
+        HRGN hrgn = 0;
+        int result = ::GetClipRgn((WXHDC)m_hDC, hrgn);
```
vadz, line 359: Minor, but:

```suggestion
        int result = ::GetClipRgn((HDC)m_hDC, hrgn);
```

(although the fact that it compiles as it is seems to indicate that it's fine to not have any cast at all)

src/msw/dc.cpp
```diff
@@ -339,6 +339,37 @@ void wxMSWDCImpl::SelectOldObjects(WXHDC dc)
...
+        if ( result == -1 )
+        {
+            wxLogLastError("GetClipRgn");
+        }
+        else
+        {
+            // We might have a pre-existing clipping region; make sure that we
+            // return it if asked for.
+            m_clipping = result;
```
vadz, line 368: Maybe

```suggestion
            m_clipping = result == 1;
```

? Personally I'd probably rewrite this as

```cpp
switch ( result )
{
    case 0:
        // No clipping region.
        m_clipping = false;
        break;

    case 1:
        // There is an existing clipping region.
        m_clipping = true;
        break;

    case -1:
        // This function is not documented to set last error.
        wxLogDebug("GetClipRgn() failed");
        break;

    default:
        wxLogDebug("unexpected GetClipRgn() return value");
}
```

vadz (commented):
Sorry, I'm not sure which problem does this fix, i.e. why is it a problem to have `m_clipping == true` even when there is no clipping region? AFAICS the only thing which happens in this case is that `DestroyClippingRegion()` may be called uselessly, but it's not clear if this is really a problem?

But if we do have to do this (I'd just like to understand why), the change looks good and I just have 2 small nitpicks (which I may apply myself when merging if you agree with them).

AliKet:
Sorry this fix is WRONG and will be replaced with a correct one soon...
Sorry again!

PR #2517, opened 2021-09-10
Implement Move To Trash feature

oneeyeman1:
This is a second attempt to push "Move to Trash" feature to the library.

I made this function static and I hope I fixed all the documentation things.

Please review.

Next will be an asynchronous implementation of this feature.

Thank you and sorry for such a long delay.

PR #26256, opened 2026-03-01
Add wxMoveToTrash function

Blake-Madden:
Moves a file or directory (recursively) into the trash/recycle bin. If it fails, then returns `false` and leaves the file alone; I thought calling `wxRemoveFile` as an alternative path upon failure would be too aggressive and arbitrary.

I added implementation details in the docs; we seem to do that with system-level functions like this.

Tested on:

- Windows 11
- Linux Mint
- macOS Catalina

### DIFF
--- a/interface/wx/filefn.h
+++ b/interface/wx/filefn.h
@@ -217,7 +217,7 @@ void wxSplitPath(const wxString& fullname,
 /**
     Returns time of last modification of given file.
 
-    The function returns <tt>(time_t)-1</tt> if an error occurred (e.g. file
+    The function returns <tt>(time_t)-1</tt> if an error occurred (e.g., file
     not found).
 
     @header{wx/filefn.h}
@@ -362,7 +362,7 @@ bool wxRemoveFile(const wxString& file);
     <a href="https://specifications.freedesktop.org/trash/latest/">FreeDesktop
     Trash specification</a> directly.
 
-    @note On Unix, moving files across filesystem boundaries (e.g. from an
+    @note On Unix, moving files across filesystem boundaries (e.g., from an
     external drive) is not supported and will return @false.
 
     @since 3.3.3
--- a/interface/wx/filefn.h
+++ b/interface/wx/filefn.h
@@ -349,6 +349,29 @@ bool wxConcatFiles(const wxString& src1,
 bool wxRemoveFile(const wxString& file);
 
 /**
+    Moves @a path to the system trash or recycle bin, returning @true if
+    successful.
+
+    This works for both files and directories. The item is not permanently
+    deleted and can be restored by the user from the platform's trash
+    facility.
+
+    Under Windows, this uses the shell file operation with @c FOF_ALLOWUNDO.
+    Under macOS, this uses @c NSFileManager's @c trashItemAtURL method.
+    Under Unix systems (including Linux and BSD), this implements the
+    <a href="https://specifications.freedesktop.org/trash/latest/">FreeDesktop
+    Trash specification</a> directly.
+
+    @note On Unix, moving files across filesystem boundaries (e.g. from an
+    external drive) is not supported and will return @false.
+
+    @since 3.3.3
+
+    @header{wx/filefn.h}
+*/
+bool wxMoveToTrash(const wxString& path);
+
+/**
     File permission bit names.
 
     We define these constants in wxWidgets because S_IREAD &c are not standard.
--- a/interface/wx/filefn.h
+++ b/interface/wx/filefn.h
@@ -363,7 +363,7 @@ bool wxRemoveFile(const wxString& file);
     @note On Unix, moving files across filesystem boundaries (e.g., from an
     external drive) is not supported and will return @false.
 
-    @availability{wxHAS_MOVE_TO_TRASH}
+    Requires @c wxHAS_MOVE_TO_TRASH.
 
     @since 3.3.3
 
--- a/interface/wx/filefn.h
+++ b/interface/wx/filefn.h
@@ -363,6 +363,8 @@ bool wxRemoveFile(const wxString& file);
     @note On Unix, moving files across filesystem boundaries (e.g., from an
     external drive) is not supported and will return @false.
 
+    @availability{wxHAS_MOVE_TO_TRASH}
+
     @since 3.3.3
 
     @header{wx/filefn.h}
--- a/interface/wx/filefn.h
+++ b/interface/wx/filefn.h
@@ -363,8 +363,6 @@ bool wxRemoveFile(const wxString& file);
     @note On Unix, moving files across filesystem boundaries (e.g., from an
     external drive) is not supported and will return @false.
 
-    Requires @c wxHAS_MOVE_TO_TRASH.
-
     @since 3.3.3
 
     @header{wx/filefn.h}
--- a/interface/wx/filefn.h
+++ b/interface/wx/filefn.h
@@ -356,11 +356,9 @@ bool wxRemoveFile(const wxString& file);
     deleted and can be restored by the user from the platform's trash
     facility.
 
-    Under Windows, this uses the shell file operation with @c FOF_ALLOWUNDO.
+    Under Windows, this uses @c SHFileOperation with @c FOF_ALLOWUNDO.
     Under macOS, this uses @c NSFileManager's @c trashItemAtURL method.
-    Under Unix systems (including Linux and BSD), this implements the
-    <a href="https://specifications.freedesktop.org/trash/latest/">FreeDesktop
-    Trash specification</a> directly.
+    Under Unix systems (including Linux and BSD), this uses @c g_file_trash.
 
     @note On Unix, moving files across filesystem boundaries (e.g., from an
     external drive) is not supported and will return @false.
--- a/interface/wx/filefn.h
+++ b/interface/wx/filefn.h
@@ -360,9 +360,6 @@ bool wxRemoveFile(const wxString& file);
     Under macOS, this uses @c NSFileManager's @c trashItemAtURL method.
     Under Unix systems (including Linux and BSD), this uses @c g_file_trash.
 
-    @note On Unix, moving files across filesystem boundaries (e.g., from an
-    external drive) is not supported and will return @false.
-
     @since 3.3.3
 
     @header{wx/filefn.h}
--- a/src/common/filefn.cpp
+++ b/src/common/filefn.cpp
@@ -681,7 +681,7 @@ bool wxMoveToTrash(const wxString& path)
 {
     if ( !wxFileExists(path) && !wxDirExists(path) )
     {
-        wxLogError(_("'%s' doesn't exist and can't be moved to trash"), path);
+        wxLogError(_("'%s' doesn't exist and can't be moved to trash."), path);
         return false;
     }
 
--- a/src/common/filefn.cpp
+++ b/src/common/filefn.cpp
@@ -641,6 +641,160 @@ bool wxRemoveFile(const wxString& file)
     return res == 0;
 }
 
+#if defined(__WXMAC__)
+extern bool wxMoveToTrashOSX(const wxString& path);
+#endif
+
+#if defined(__UNIX__) && !defined(__WXMAC__)
+#include <fcntl.h>
+
+// Percent-encode a file path per RFC 2396 for the FreeDesktop Trash spec.
+static wxString wxTrashUrlEncodePath(const wxString& path)
+{
+    const wxScopedCharBuffer utf8 = path.utf8_str();
+    const char* cur = utf8;
+    wxString result;
+
+    while ( *cur )
+    {
+        const unsigned char ch = static_cast<unsigned char>(*cur);
+        // Unreserved characters per RFC 2396: A-Z a-z 0-9 - _ . ~ /
+        if ( (ch >= 'A' && ch <= 'Z') ||
+             (ch >= 'a' && ch <= 'z') ||
+             (ch >= '0' && ch <= '9') ||
+             ch == '-' || ch == '_' || ch == '.' || ch == '~' || ch == '/' )
+        {
+            result += static_cast<char>(ch);
+        }
+        else
+        {
+            result += wxString::Format("%%%02X", ch);
+        }
+        ++cur;
+    }
+
+    return result;
+}
+#endif // __UNIX__ && !__WXMAC__
+
+bool wxMoveToTrash(const wxString& path)
+{
+    if ( !wxFileExists(path) && !wxDirExists(path) )
+    {
+        wxLogError(_("'%s' doesn't exist and can't be moved to trash"), path);
+        return false;
+    }
+
+#if defined(__WINDOWS__)
+    // SHFileOperation needs double null termination string
+    // but without separator at the end of the path
+    wxString pathStr(path);
+    if ( pathStr.Last() == wxFILE_SEP_PATH )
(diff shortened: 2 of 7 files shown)


src/common/filefn.cpp
```diff
@@ -641,6 +641,160 @@ bool wxRemoveFile(const wxString& file)
...
+    SHFILEOPSTRUCT fileop;
+    wxZeroMemory(fileop);
+    fileop.wFunc = FO_DELETE;
+    fileop.pFrom = pathStr.t_str();
+    fileop.fFlags = FOF_ALLOWUNDO | FOF_SILENT | FOF_NOCONFIRMATION | FOF_NOERRORUI;
+
+    int ret = SHFileOperation(&fileop);
+    if ( ret != 0 || fileop.fAnyOperationsAborted )
+    {
+        wxLogDebug(wxS("SHFileOperation(FO_DELETE with FOF_ALLOWUNDO) failed: error 0x%08x"),
```
PBfordev: Why was `wxLogDebug()` and not `wxLogError()` used here (and for macOS)? The error is reported to the user in the Linux code path and other wx file functions.

src/common/filefn.cpp
```diff
@@ -641,6 +641,160 @@ bool wxRemoveFile(const wxString& file)
...
+
+    return result;
+}
+#endif // __UNIX__ && !__WXMAC__
+
+bool wxMoveToTrash(const wxString& path)
+{
+    if ( !wxFileExists(path) && !wxDirExists(path) )
+    {
+        wxLogError(_("'%s' doesn't exist and can't be moved to trash"), path);
```
PBfordev:
```suggestion
        wxLogError(_("'%s' doesn't exist and can't be moved to trash."), path);
```

AFAICT, the sentences in `wxLogError()` (but not `wxLogSysError()`) end with a period elsewhere. Applies to the one at the end of the function as well.

PBfordev (commented):
I suppose introducing a new function instead of adding a flag to the existing remove file/folder functions seemed better.

src/common/filefn.cpp
Blake-Madden: That's how it's done in `wxFileName::Rmdir` (src/common/filename.cpp). Not saying that's right or wrong, I just re-used the code from there.

oneeyeman1:
@Blake-Madden ,
Please look at https://github.com/wxWidgets/wxWidgets/pull/2517 and try not to repeat the same mistakes I did... ;-)

Thank you.

src/common/filefn.cpp
```diff
@@ -641,6 +641,160 @@ bool wxRemoveFile(const wxString& file)
...
 
+#if defined(__WXMAC__)
+extern bool wxMoveToTrashOSX(const wxString& path);
+#endif
+
+#if defined(__UNIX__) && !defined(__WXMAC__)
+#include <fcntl.h>
+
+// Percent-encode a file path per RFC 2396 for the FreeDesktop Trash spec.
+static wxString wxTrashUrlEncodePath(const wxString& path)
```
vadz: Could we reuse `wxURI::BuildURI()` here? It's already supposed to have the correct logic for this.

src/common/filefn.cpp
vadz: I think it was done like this in `wxFileName::Rmdir()` because it doesn't show user-visible log messages in any case (why is it so is another question...). This function does give errors in the Linux/Unix version and so I agree that this should be a `wxLogError()` too.

src/common/filefn.cpp
```diff
@@ -641,6 +641,160 @@ bool wxRemoveFile(const wxString& file)
...
+    if ( !wxGetEnv("XDG_DATA_HOME", &xdgData) || xdgData.empty() )
+        xdgData = wxFileName::GetHomeDir() + "/.local/share";
+
+    wxString trashFiles = xdgData + "/Trash/files";
+    wxString trashInfo  = xdgData + "/Trash/info";
+
+    if ( !wxFileName::Mkdir(trashFiles, wxS_DIR_DEFAULT, wxPATH_MKDIR_FULL) )
+        return false;
+    if ( !wxFileName::Mkdir(trashInfo, wxS_DIR_DEFAULT, wxPATH_MKDIR_FULL) )
+        return false;
```
vadz: Should we create them? Maybe just check if they already exist and bail out if not? Not really sure, maybe this is the right thing to do, but some reference to the docs or whatever would be welcome here.

src/common/filefn.cpp
```diff
@@ -641,6 +641,160 @@ bool wxRemoveFile(const wxString& file)
...
+    wxFileName fileName(path);
+    fileName.MakeAbsolute();
+    wxString fullPath = fileName.GetFullPath();
+
+    // Pick a unique trash name atomically using O_EXCL per the spec
+    wxString baseName = fileName.GetFullName();
+    wxString trashName = baseName;
+    wxString infoPath;
+
+    for ( int attempt = 2; ; ++attempt )
```
vadz: It seems a bit dangerous to never stop, I'd put an upper bound on the number of attempts here (even if very high, e.g. 1,000,000). But again, I'm not sure what the spec says.

src/common/filefn.cpp
```diff
@@ -641,6 +641,160 @@ bool wxRemoveFile(const wxString& file)
...
+            return false;
+        }
+
+        wxString encodedPath = wxTrashUrlEncodePath(fullPath);
+        wxString info = wxString::Format(
+            "[Trash Info]\nPath=%s\nDeletionDate=%s\n",
+            encodedPath,
+            wxDateTime::Now().FormatISOCombined()
+        );
+        infoFile.Write(info);
```
vadz: Should check for error.

src/common/filefn.cpp
```diff
@@ -641,6 +641,160 @@ bool wxRemoveFile(const wxString& file)
...
+        wxFile infoFile;
+        if ( !infoFile.Open(infoPath, wxFile::write) )
+        {
+            wxRemoveFile(infoPath);
+            return false;
+        }
+
+        wxString encodedPath = wxTrashUrlEncodePath(fullPath);
+        wxString info = wxString::Format(
+            "[Trash Info]\nPath=%s\nDeletionDate=%s\n",
```
vadz: Minor, but this might look better/more clear as a raw string:


```suggestion
R"[Trash Info]
Path=%s
DeletionDate=%s
",
```

src/common/filefn.cpp
```diff
@@ -641,6 +641,160 @@ bool wxRemoveFile(const wxString& file)
...
+        infoFile.Write(info);
+    }
+
+    // Move the file or directory into Trash/files/
+    wxString destPath = trashFiles + "/" + trashName;
+    if ( wxRename(fullPath, destPath) != 0 )
+    {
+        // Move failed (e.g. cross-device) — clean up the .trashinfo
+        wxRemoveFile(infoPath);
+        wxLogSysError(_("'%s' couldn't be moved to trash"), path);
```
vadz: This won't log the correct error because it will have been overwritten by `wxRemoveFile()`, you need to call this first.

Note that using a scope guard for removing the file would take care of this automatically, as the dtor of the guard would be executed after any normal code.

src/osx/cocoa/utils_base.mm
```diff
@@ -389,3 +389,32 @@ bool wxMacInitCocoa()
...
+
+    NSError *error = nil;
+    BOOL ok = [[NSFileManager defaultManager] trashItemAtURL:fileURL
+                                            resultingItemURL:nil
+                                                       error:&error];
+    if ( !ok )
+    {
+        if ( error )
+        {
+            wxLogDebug("NSFileManager trashItemAtURL failed: %s",
```
vadz: Those should also be replaced with user-readable error messages (i.e. "'%s' couldn't be moved to trash: %s").

vadz (commented):
Thanks, this looks useful, but the error reporting needs to be streamlined.

I'm also not enamored of keeping the code for all platforms but macOS in the same function, it probably would be cleaner to have some `wxMoveToTrashImpl()` and implement it in MSW/Unix-specific file. But I realize that `src/common/filefn.cpp` is such a mess that it probably doesn't matter much.

imciner2:
Why have you implemented the freedesktop spec directly instead of using something like `g_file_trash` (https://docs.gtk.org/gio/method.File.trash.html), which I believe implemented that spec also. (note in KiCad we implemented the move to trash functionality by just calling `g_file_trash`, https://gitlab.com/kicad/code/kicad/-/blob/master/libs/kiplatform/os/unix/environment.cpp?ref_type=heads#L68).

Blake-Madden:
> Why have you implemented the freedesktop spec directly instead of using something like `g_file_trash` (https://docs.gtk.org/gio/method.File.trash.html), which I believe implemented that spec also. (note in KiCad we implemented the move to trash functionality by just calling `g_file_trash`, https://gitlab.com/kicad/code/kicad/-/blob/master/libs/kiplatform/os/unix/environment.cpp?ref_type=heads#L68).

I tried that, but would get link errors. Seems that we need to add a GLib/GIO dependency for that, so I went with a different approach.

vadz:
> I tried that, but would get link errors. Seems that we need to add a GLib/GIO dependency for that, so I went with a different approach.

Even wxBase links with glib under Linux, so I'm not sure what the problem was... In any case, if we made this a GUI function (which might make sense?), there definitely should be no problems with using glib function in wxGTK.

src/common/filefn.cpp
Blake-Madden: Removed this, using `g_file_trash` now.

src/common/filefn.cpp
Blake-Madden: Removed this, using `g_file_trash` now.

src/common/filefn.cpp
Blake-Madden: Removed this, using `g_file_trash` now.

src/common/filefn.cpp
Blake-Madden: Removed this, using `g_file_trash` now.

src/osx/cocoa/utils_base.mm
Blake-Madden: Fixed.

src/common/filefn.cpp
Blake-Madden: Removed this, using `g_file_trash` now.

Blake-Madden:
Using `g_file_trash` now. Seems simple enough now to keep MSW and GTK paths in one file, let me know if anyone disagrees.

wh11204:
I think these "unknown error" messages should be translatable

interface/wx/filefn.h
```diff
@@ -348,6 +348,29 @@ bool wxConcatFiles(const wxString& src1,
...
+    Moves @a path to the system trash or recycle bin, returning @true if
+    successful.
+
+    This works for both files and directories. The item is not permanently
+    deleted and can be restored by the user from the platform's trash
+    facility.
+
+    Under Windows, this uses the shell file operation with @c FOF_ALLOWUNDO.
+    Under macOS, this uses @c NSFileManager's @c trashItemAtURL method.
+    Under Unix systems (including Linux and BSD), this implements the
```
PBfordev: Just curious (I know nothing about Unix): Is the Unix part still still valid after changing the code?

I personally would also change the Windows part, replacing "the shell file operation" with the actual function name, i.e., `SHFileOperation`, as it is done for macOS.

Blake-Madden: fixed

Blake-Madden:
> I think these "unknown error" messages should be translatable

Fixed.

src/common/filefn.cpp
```diff
@@ -641,6 +641,71 @@ bool wxRemoveFile(const wxString& file)
...
+    GFile* file = g_file_new_for_path(path.fn_str());
+
+    bool ok = g_file_trash(file, nullptr, &err);
+    if ( !ok )
+    {
+        wxLogError(_("'%s' couldn't be moved to trash: %s"),
+                   path, err ? err->message : _("unknown error"));
+    }
+
+    g_clear_error(&err);
```
vadz: We have `wxGtkError` in `include/wx/gtk/private/error.h`, please use it instead (see the examples of existing use in wxGTK).

src/common/filefn.cpp
```diff
@@ -641,6 +641,71 @@ bool wxRemoveFile(const wxString& file)
...
+    g_object_unref(file);
```
vadz: We also have `wxGtkObject` in `include/wx/gtk/private/object.h` to avoid having to do (or, worse, forget to do) this manually.

src/common/filefn.cpp
```diff
@@ -641,6 +641,71 @@ bool wxRemoveFile(const wxString& file)
...
+#include <gio/gio.h>
+#endif
+
+bool wxMoveToTrash(const wxString& path)
+{
+    if ( !wxFileExists(path) && !wxDirExists(path) )
+    {
+        wxLogError(_("'%s' doesn't exist and can't be moved to trash."), path);
+        return false;
+    }
```
vadz: I wonder if this is actually needed, wouldn't we get a similar error from the native function anyhow in this case?

vadz (commented):
Thanks, I definitely prefer the new version using the glib function. However I still think it would be better to implement this function in platform-specific file instead of the current approach.

I would advise doing the following:

1. Add (and document in `docs/doxygen/mainpages/const_cpp.h`) `wxHAS_MOVE_TO_TRASH` preprocessor symbol.
2. Define it for the platforms really implementing this function in the header and implement it in the platform-specific files.
3. Provide fallback function just returning false (maybe without logging anything, as I don't think this error message is really useful) in the common code `#ifndef wxHAS_MOVE_TO_TRASH`.

And, of course, the CI failures need to be fixed (you should use `.utf8_str()`).

src/common/filefn.cpp
Blake-Madden: Indeed. I tend to be overly pedantic :) Removing this now.

src/common/filefn.cpp
Blake-Madden: Making change now.

src/common/filefn.cpp
Blake-Madden: Fixing now...

Blake-Madden:
@vadz All requested changes applied.

PBfordev:
The error reporting on MSW still uses `wxLogDebug()` instead of `wxLogError()` the other two ports use?

"unknown error" in the macOS code path is not localized?

As for the docs:
* Does the Unix note still apply after changing the API (and the rest of the Unix docs)?
* I don't think I saw something like "Requires @c wxHAS_MOVE_TO_TRASH." for another API, this seems to be an implementation detail. The function is supported on all three platforms (Qt is not usually mentioned in the docs); I would not mention the availability at all here.

Sorry for all the nagging.

Blake-Madden:
Thanks, @PBfordev, I've applied the updates.

PBfordev:
> Thanks, @PBfordev, I've applied the updates.

I would use the same error reporting on MSW as for the other two, i.e., (1) Make the message translatable and (2) Use the same basic message as in the other two ports (IMO, the current one is overly technical for the end user).

Blake-Madden:
> I would use the same error reporting on MSW as for the other two, i.e., (1) Make the message translatable and (2) Use the same basic message as in the other two ports (IMO, the current one is overly technical for the end user).

Done

vadz (approved):
Looks perfect now, thanks! Will merge soon.

PR #26267, opened 2026-03-06
Remove STC string conversion functions

filip-hejsek:
This PR replaces Scintilla-specific string conversion functions by direct calls to `FromUTF8()`/`utf8_str()`, as suggested by this comment:
```c++
// For historical reasons, we use Scintilla-specific conversion functions, we
// should probably just call FromUTF8()/utf8_str() directly instead now.
```

Where possible, I changed the variables holding the result to `wxScopedCharBuffer`. I suggest reviewers pay extra attention to this as a mistake there could lead to memory bugs. (Though I believe my changes are correct.)


src/stc/PlatWX.cpp
```diff
@@ -2921,7 +2920,7 @@ void wxSTCListBox::Select(int n)
 
 void wxSTCListBox::GetValue(int n, char *value, int len) const
 {
-    strncpy(value, wx2stc(m_labels[n]), len);
+    strncpy(value, m_labels[n].utf8_str(), len);
```
vadz, line 2923: Unrelated, but it would be good to replace this with (safer) `wxStrlcpy()`.

src/stc/PlatWX.cpp
```diff
@@ -3227,19 +3226,17 @@ class wxSTCListBoxD2D : public wxSTCListBox
         // characters from the end of the label until it's short enough.
         wxString ellipsizedLabel = label;
 
-        wxCharBuffer buffer = wx2stc(ellipsizedLabel);
-        int ellipsizedLen = wx2stclen(ellipsizedLabel, buffer);
-        int curWidth = surface.WidthText(tempFont, buffer.data(),ellipsizedLen);
+        wxCharBuffer buffer = ellipsizedLabel.utf8_str();
```
vadz, line 3229: This could be a `wxScopedCharBuffer` too, it's not used outside of `ellipsizedLabel` scope and it would avoid an extra copy of a potentially long string.

vadz (commented):
Looks good to me, thanks for simplifying this!

Will merge soon.

src/stc/PlatWX.cpp
filip-hejsek: I didn't use `wxScopedCharBuffer` because `ellipsizedLabel` is later reassigned, which would invalidate the buffer. But I didn't realize it's actually fine because every time `ellipsizedLabel` is reassigned, `buffer` is reassigned too.

Although, if the string is long, there ale bigger problems than one extra copy - in non-UTF8 builds it will repeatedly convert prefixes to UTF8 until one is found that fits, which will create a lot of copies. And measuring prefixes repeatedly like this is an O(n²) algorithm.

Also this is a pretty bad algorithm for ellipsizing strings. It completely disregards the complexities of Unicode like combining characters etc. But I don't feel like I want to spend much effort fixing this.

filip-hejsek: And the `WidthText` method immediately converts the string back to `wxString`

vadz: FWIW I agree with all your points above, there is definitely scope for more improvements here... (and not just here, ellipsization algorithm used in the main library is the same)

PR #26268, opened 2026-03-06
wxFindWindowAtPoint with specified window exclusion

sulakm:
Hi,

I added overloaded wxFindWindowAtPoint with extra parameter for dragged wxWindow to be excluded from the search.
On MSW it uses httransparent in nchittest temporarily during search. On other platforms it simply continue iterating windows.

regards, Martin

### DIFF
--- a/include/wx/msw/window.h
+++ b/include/wx/msw/window.h
@@ -744,6 +744,9 @@ protected:
     }
 
 private:
+    friend wxWindow* wxFindWindowAtPoint(const wxPoint& pt, wxWindow* skip);
+    bool m_hideFromFind;
+
     // common part of all ctors
     void Init();
 
--- a/include/wx/msw/window.h
+++ b/include/wx/msw/window.h
@@ -744,7 +744,7 @@ protected:
     }
 
 private:
-    friend wxWindow* wxFindWindowAtPoint(const wxPoint& pt, wxWindow* skip);
+    friend WXDLLIMPEXP_CORE wxWindow* wxFindWindowAtPoint(const wxPoint& pt, wxWindow* skip);
     bool m_hideFromFind;
 
     // common part of all ctors
--- a/include/wx/utils.h
+++ b/include/wx/utils.h
@@ -637,7 +637,9 @@ WXDLLIMPEXP_CORE int wxFindMenuItemId(wxFrame *frame, const wxString& menuString
 
 // Find the wxWindow at the given point. wxGenericFindWindowAtPoint
 // is always present but may be less reliable than a native version.
+WXDLLIMPEXP_CORE wxWindow* wxGenericFindWindowAtPoint(const wxPoint& pt, wxWindow* skip);
 WXDLLIMPEXP_CORE wxWindow* wxGenericFindWindowAtPoint(const wxPoint& pt);
+WXDLLIMPEXP_CORE wxWindow* wxFindWindowAtPoint(const wxPoint& pt, wxWindow* skip);
 WXDLLIMPEXP_CORE wxWindow* wxFindWindowAtPoint(const wxPoint& pt);
 
 // NB: this function is obsolete, use wxWindow::FindWindowByLabel() instead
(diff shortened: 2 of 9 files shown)


sulakm:
That Unix build does not show any particular error in code, only timeout in setup phase before build: Error: retrieving gpg key timed out.
Aren't gpg keys related to apt ? Other tests on the same machine went well, so it maybe was a network blackout. Too bad there is no rerun option.

include/wx/msw/window.h
```diff
@@ -744,6 +744,9 @@ class WXDLLIMPEXP_CORE wxWindowMSW : public wxWindowBase
     }
 
 private:
+    friend WXDLLIMPEXP_CORE wxWindow* wxFindWindowAtPoint(const wxPoint& pt, wxWindow* skip);
+    bool m_hideFromFind;
```
vadz, line 748: I don't like the idea of adding a new bool to all window objects just to support one particular function which may not even be called at all.

vadz (commented):
Thanks for your contribution, but it's not clear why exactly is this needed/useful, could you please explain it? It would also need to be explained in the documentation (to be added to `interface/wx/utils.h`).

include/wx/msw/window.h
sulakm: The use-case is drag&drop of a toplevel window. wxFindWindowAtPoint is not useful for finding drop destination because it return the dragged window. Similar problem was discussed on stackoverflow: https://stackoverflow.com/questions/37446702/find-all-windows-beneath-a-point.

Initially I planned to rewrite wxAuiNotebook for undocking pages like in popular web browser. It is not however to redock undocked window by interception onMoving, especially on older versions of MacOS. This feature is ready half way through and not commited on github. Now I aiming at dragging just a hint transparent rectangle in a temporary window. For this I still need wxFindWindowAtPoint with option to skip a dragged window.

As for extra boolean variable: it is for temporary make window transparent. Maybe bit in style would suffice. Or a static pointer for wxWindows to be made transparent if thread-safety is not a concern. Normally this is done in gui thread. Member bool variable is not thread-safe either because I did not think of thread. Atomic counter would be better.

vadz: This is another thing which is going to be unimplementable with Wayland... Not sure if we want to encourage doing it. It definitely would be better to have a high level window dragging API instead, as discussed elsewhere.

Thread-safety is not a concern because GUI functions can only be called from a GUI thread. But FWIW I think looking at the next window in Z-order after the window being dragged could be even simpler and not require any hacks at all.

sulakm: Does not Wayland port use [wxGenericFindWindowAtPoint](https://github.com/wxWidgets/wxWidgets/pull/26268/changes/b851d80459bf4b6bcdfe56a3908974efcf2021b3#diff-c17c0f8a3d3d5659cb1cc771d1a49efcbff854dd1ca3149c7af1af74786a2302L1340) ?

Did you refer to a native API with Z-order or it is available after calling wxFindWindowAtPoint ?

BTW why only MSW has a native version ? There is probably no performance issue with the generic function. Does wxTopLevelWindows have wrong Z-order ?

vadz: Under Wayland you can't find the window under mouse when it's outside of your own window, so it doesn't matter what it uses, it's not going to work anyhow. The only way to make dragging windows work there is via the drag and drop protocol and this is what we need to wrap, in a higher level API.

I don't know/remember why MSW has a native version, but as long as it works I wouldn't change it as I'm not sure if `::WindowFromPoint()` is exactly the same as that of our function.

sulakm: wxGenericWindowFromPoint doesn't work on Wayland ? This function doesn't use any system API. It simply iterates in-process C++ models of all  top level windows created by wxWidgets. Am I wrong ?

vadz: I think it's broken because it doesn't take Z-order into account. And AFAIK there is no way to get it, even for your own windows.

sulakm: You mean generic version is broken ? So wxFindWindowAtPoint only works on MSW ?
And what about Z-order of child windows ?

vadz:
Returning to this, I'm tempted to reject this PR because of the considerations above. Does anyone else think that we should apply it instead?

vadz:
No votes in favour, so closing, sorry.

PR #26270, opened 2026-03-06
Fix the accessible label of checkboxes.

trypsynth:
I noticed an absolutely critical bug when updating to wxWidgets 3.3.2. All of my checkboxes now have their accessible labels replaced with "check"!
The problem: wxWindowAccessible::GetName() has a special case for wxButton -> GetLabel(), but falls through to GetWindow()->GetName() for everything else. GetName() returns the internal wx window name ("check"), not the user-visible label. Before, owner-drawn checkboxes fell back to Windows' own IAccessible which reads window text (the label) directly. Once CreateAccessible() was wired up, it started going through wxWindowAccessible::GetName() and returning the wrong thing. The fix adds GetName() to wxOwnerDrawnAccessible returning ctrl->GetLabel() when owner-drawn, and wxACC_NOT_IMPLEMENTED otherwise to pass through to the standard Windows accessible as before.
Is there a release policy for absolutely breaking bugs like this? wxWidgets 3.3.2 is utterly unusable for me because of this despite having many fixes I want. I get it if I have to wait for 3.3.3, but that would be unfortunate too.

### DIFF
--- a/include/wx/msw/private/accessible.h
+++ b/include/wx/msw/private/accessible.h
@@ -53,21 +53,6 @@ public:
         return wxACC_OK;
     }
 
-    virtual wxAccStatus GetName(int childId, wxString* name) override
-    {
-        if ( childId != wxACC_SELF )
-            return wxACC_NOT_IMPLEMENTED;
-
-        TControl* const ctrl = wxStaticCast(GetWindow(), TControl);
-        wxCHECK(ctrl, wxACC_FAIL);
-
-        if ( !ctrl->MSWIsOwnerDrawn() )
-            return wxACC_NOT_IMPLEMENTED;
-
-        *name = ctrl->GetLabel();
-        return wxACC_OK;
-    }
-
     virtual wxAccStatus GetState(int childId, long* state) override
     {
         if ( childId != wxACC_SELF )
--- a/include/wx/msw/private/accessible.h
+++ b/include/wx/msw/private/accessible.h
@@ -53,6 +53,21 @@ public:
         return wxACC_OK;
     }
 
+    virtual wxAccStatus GetName(int childId, wxString* name) override
+    {
+        if ( childId != wxACC_SELF )
+            return wxACC_NOT_IMPLEMENTED;
+
+        TControl* const ctrl = wxStaticCast(GetWindow(), TControl);
+        wxCHECK(ctrl, wxACC_FAIL);
+
+        if ( !ctrl->MSWIsOwnerDrawn() )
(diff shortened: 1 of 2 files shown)


vadz:
Sorry for the breakage, but the thing I don't understand is that this seems to have been your own change, so how did this manage to pass unnoticed??

Anyhow, I don't know if we're going to make 3.3.2.1 just for this, do you really need to use an official release? As you're building wx on your own, why can't you use a fixed version from master, once this is fixed?

But, speaking about the fix, I actually dislike making things even more complicated instead of fixing the underlying problem, which is that `wxWindowAccessible::GetName()` handles the buttons specially. Shouldn't it just use `GetLabel()` for all windows instead? This would fix the bug too, AFAICS, and be much simpler and avoid such problem happening again.

trypsynth:
Frankly, I have utterly no idea how this wasn't noticed. I just ran my little wxWidgets test app again, and the two checkboxes I put in that window read, based off the initial commit I made to that PR. However, checking out latest master, it breaks. I have utterly no idea what the hell is happening here, other than that maybe the checkbox improvements for lists broke something or...something? I have no idea, but my own test app on that initial commit doesn't show this bug.
I'm not building wxWidgets myself, I'm using wxDragon, which ahrdcodes a particular wx version.

trypsynth:
I do agree about how to make it simpler though, that I can do

trypsynth:
@vadz I've made the change you suggested.

trypsynth:
I understand if it's not wanted to make a wxWidgets patch release just for one bug, but would it at least maybe be possible to put the possibility of one soon on the table, in case there are a few more bugs that pop up, to make it a slightly larger release? I don't want to have to wait 6 months to be able to use this change. Alternatively, I might just figure out how to make wxDragon support wxWidgets from master, but that's more complicated.

trypsynth:
@vadz It was indeed the commit that merged in my pull request that broke this behavior, but that is very odd to me as a copy of wxWidgets I have cloned with that branch still in its local git seems to be working? In any case it doesn't matter much, this fixes it, just sad I missed it before 3.3.2.

vadz:
I don't understand how could I have broken this while merging, but, as you say, it's too late anyhow.

I could create v3.3.2.1 tag in Git if this could help you, what I want to avoid is asking people to rebuild binaries again...

trypsynth:
Fair enough, yeah, that makes complete sense. I'll ask the wxDragon maintainer what would be easiest here and get back to you. Sorry for the hassle.

PR #26278, opened 2026-03-09
Simplify clipboard operations in GTK

davidedmundson:
When changing the clipboard contents the current code emits a lot of noise on the X11 connection, it clears the contents and verifies it's empty before settings the new content.

This non-atomic operation is problematic for desktop environments with clipboard managers that try to replace an empty clipboard. It becomes racey with us setting our new contents.

See #26265

### DIFF
--- a/src/gtk/clipbrd.cpp
+++ b/src/gtk/clipbrd.cpp
@@ -646,8 +646,6 @@ bool wxClipboard::SetData( wxDataObject *data )
 
     wxCHECK_MSG( data, false, wxT("data is invalid") );
 
-    Clear();
-
     return AddData( data );
 }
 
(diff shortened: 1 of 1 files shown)


vadz:
Thanks David! ~~Other than (presumably accidental?) change to submodules~~ (Edit: this was fixed while I was writing this), this looks fine to me, so I'll merge this if there are no objections.

And this is probably safe to backport to 3.2 too, at least I can't think of anything this change could break, can anyone else?

davidedmundson:
>And this is probably safe to backport to 3.2 too, 

Yeah, I did my main real-world testing for the actual bugfix against 3.2.  
I can't see anything that would break.

pvantrap:
Great work! Which 3.2 will contain this backport @vadz ?

vadz:
The next one, i.e. 3.2.11.

PR #26284, opened 2026-03-12
Some more RTL fixes

AliKet:
Rubberbanding is broken without these changes. Now, it's working correctly again under `wxMSW`, `wxGTK3` and `wxQt`

https://github.com/user-attachments/assets/11cf2948-8c5e-4a11-8ae8-82a495400315

### DIFF
--- a/src/common/wincmn.cpp
+++ b/src/common/wincmn.cpp
@@ -1082,6 +1082,17 @@ void wxWindowBase::DoGetScreenPosition(int *x, int *y) const
         *y = 0;
 
     ClientToScreen(x, y);
+
+    if ( x && GetLayoutDirection() == wxLayout_RightToLeft )
+    {
+        // In RTL layout, ClientToScreen(0, 0) correctly returns the upper-right corner
+        // of the window (for non TLWs). But the window position relative to the desktop
+        // surface should be expressed by its upper-left corner because position (0, 0)
+        // of the desktop surface is always at the upper-left even in RTL.
+        int width;
+        DoGetSize(&width, nullptr);
+        *x -= width;
+    }
 }
 
 void wxWindowBase::SendSizeEvent(int flags)
--- a/src/msw/graphics.cpp
+++ b/src/msw/graphics.cpp
@@ -1905,6 +1905,17 @@ wxGDIPlusContext::wxGDIPlusContext( wxGraphicsRenderer* renderer, HDC hdc, wxDou
     : wxGraphicsContext(renderer)
     , m_layoutDir((::GetLayout(hdc) & LAYOUT_RTL) ? wxLayout_RightToLeft : wxLayout_LeftToRight)
 {
+    if ( m_layoutDir == wxLayout_RightToLeft )
+    {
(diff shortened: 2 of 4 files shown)


src/msw/graphics.cpp
```diff
@@ -92,7 +92,7 @@ StringFormat* gs_drawTextStringFormat = nullptr;
 // Get the string format used for the text drawing and measuring functions:
 // notice that it must be the same one for all of them, otherwise the drawn
 // text might be of different size than what measuring it returned.
-inline StringFormat* GetDrawTextStringFormat(wxLayoutDirection dir = wxLayout_Default)
+inline StringFormat* GetDrawTextStringFormat(bool isRTL = false)
```
vadz: I still hate bool parameters and would prefer to do this:

```suggestion
inline StringFormat* GetDrawTextStringFormat(DWORD layoutDir = 0)
```

(and check for `layoutDir & LAYOUT_RTL` below).

src/msw/graphics.cpp
```diff
@@ -1903,16 +1903,33 @@ public :
...
     : wxGraphicsContext(renderer)
-    , m_layoutDir((::GetLayout(hdc) & LAYOUT_RTL) ? wxLayout_RightToLeft : wxLayout_LeftToRight)
+    , m_layoutDir(::GetLayout(hdc))
 {
+    if ( IsRTL() )
+    {
+        // For expected results, always perform drawing operations on an LTR HDC.
+        // A transformation matrix will be applied to this context to achieve the
+        // necessary mirroring effects.
+        ::SetLayout(hdc, 0);
```
vadz: I'm not sure what the expected results are, but it looks like using RTL HDC should give the same results as using LTR one and then transforming it, so it seems to me that there could still be another bug somewhere, if this is not the case...

src/common/wincmn.cpp
```diff
@@ -1082,6 +1082,16 @@ void wxWindowBase::DoGetScreenPosition(int *x, int *y) const
         *y = 0;
 
     ClientToScreen(x, y);
+
+    if ( x && GetLayoutDirection() == wxLayout_RightToLeft )
+    {
+        // In RTL layout, ClientToScreen(0, 0) correctly returns the upper-right corner
+        // of the window (for non TLWs). But wxWidgets expects GetScreenPosition() to
+        // always return the upper-left corner instead to work with.
```
vadz: I think `ClientToScreen(0,0)` should indeed return the upper-right corner of the window when using RTL layout. Why/where do we rely on it being the upper-left corner?

vadz (commented):
You're the de factor RTL expert, so I can merge it if you're sure this is the right thing to do, but these changes seem very suspicious to me, i.e. I believe they work around some other bug(s) elsewhere and shouldn't be needed.

AliKet:
I’ve updated the comments and hope they are clear now.

vadz:
Thanks, the comments are indeed clear now and I'll merge this soon.

Thanks again!

PR #26289, opened 2026-03-14
Translations

ovari:
https://github.com/wxWidgets/wxWidgets/issues/26279#issuecomment-4036997498

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,9 @@ of an application using the library even in the case of static linking.
 Translating
 -----------
 
-Translations are usually updated via pull requests: https://github.com/wxWidgets/wxWidgets/commits/master/locale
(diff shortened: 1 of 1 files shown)


PBfordev:
Translations are documented here: https://wxwidgets.org/about/translations/

But I am not sure how up-to-date that is. It seems that at least the names of translators for the individual languages are not being updated/synced with GIT.

vadz:
Thanks, I agree that it's useful to have some indication about how to contribute translations in the README, but I don't think the link to Git repository is especially helpful, so I'll use the link from above instead and, more importantly, make this part of the new "Contributing" section instead of having a separate section just for translations.

I'll push my changes replacing this PR soon.

PR #26290, opened 2026-03-14
Fix static build regression on GTK (wxMoveToTrash)

Blake-Madden:
Static builds on Linux (GTK), `wxMoveToTrash` will cause a link error because the stub version will be included in `wx::base` (because `__GTK__` isn't defined), but then the actual GTK implementation is built into `wx::core` (where `__GTK__` is defined; this is all because of the GIO dependency).

So we need to ensure that under GTK, only the actual GTK function is compiled (into `wx::core`) and the stub is left out. Under macOS and Windows, it continues to be built into `wx::base`.

This does mean that if `wxHAS_MOVE_TO_TRASH` is not defined, then this function will not be available now. IMO, that actually seems better than a stub function that silently fails.

P.S.
This behavior is likely by design because of https://github.com/wxWidgets/wxWidgets/issues/26043

### DIFF
--- a/include/wx/filefn.h
+++ b/include/wx/filefn.h
@@ -440,7 +440,13 @@ WXDLLIMPEXP_BASE bool wxRemoveFile(const wxString& file);
     #define wxHAS_MOVE_TO_TRASH
 #endif
 
-WXDLLIMPEXP_BASE bool wxMoveToTrash(const wxString& path);
+#ifdef wxHAS_MOVE_TO_TRASH
+    #ifdef __WXGTK__
+        WXDLLIMPEXP_CORE bool wxMoveToTrash(const wxString& path);
+    #else
+        WXDLLIMPEXP_BASE bool wxMoveToTrash(const wxString& path);
+    #endif
+#endif // wxHAS_MOVE_TO_TRASH
 
 // Rename file
 WXDLLIMPEXP_BASE bool wxRenameFile(const wxString& oldpath, const wxString& newpath, bool overwrite = true);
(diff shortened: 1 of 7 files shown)


vadz:
Oh, I should have thought about this, it's not the first time something like this happens, thanks for noticing and fixing it.

However I think we should define this function in core library for all platforms for consistency, it's weird to have it in base on some platforms and core on other ones.

Alternative, but more complicated, solution would be to virtualize it via `wxAppTraits`, which would allow having the function in base on all platforms — but in wxGTK it would just return false in console applications and only really work in GUI ones.

Blake-Madden:
Would that just be a matter of moving the macOS and MSW implementations to different source files (that `core` includes) and using `WXDLLIMPEXP_CORE` in the header? Would that be the best approach?

vadz:
> Would that just be a matter of moving the macOS and MSW implementations to different source files (that `core` includes) and using `WXDLLIMPEXP_CORE` in the header? 

Yes.

> Would that be the best approach?

"Best", I don't know, but definitely the simplest acceptable one.

Blake-Madden:
I moved the masOS and MSW implementations to core files.
The UNIX errors seem to be a GH action certificate issue.
I'm not sure what the issues are with the 3 failing mac ones, I don't see any actual errors except the final step.

MaartenBent:
> I'm not sure what the issues are with the 3 failing mac ones, I don't see any actual errors except the final step.

I usually search for `error: ` in the top-right box (browser search doesn't really work) and that leads to:
```
./src/osx/carbon/utilscocoa.mm:713:9: error: use of undeclared identifier 'wxLogError'
        wxLogError(_("'%s' couldn't be moved to trash: %s"),
        ^
1 error generated.
```

vadz:
Thanks, will merge soon.

PR #26292, opened 2026-03-14
documentation: add hint for reparenting a notebook page

DietmarSchwertberger:
This would have saved me an hour...
I could not see, however, where in RemoveTab the page is actually hidden.

The info is from the forum https://forums.wxwidgets.org/viewtopic.php?t=28934
Maybe, the hint should be formulated such that the user always calls Show if there is a problem?
E.g. a separate line "In many cases you need to call Show for the window afterwards."


vadz:
Sure, this could be applied, but I'm a bit surprised that notebook is mentioned in `Reparent()` documentation at all. Maybe it would be better to mention in `wxNotebook` documentation that

1. You need to `RemovePage()` if you want to reparent the window elsewhere.
2. And document that `RemovePage()` hides it.

?

DietmarSchwertberger:
I think the place is the right one. I found the requirement for `RemovePage` when looking at `Reparent`. I would not have searched for `RemovePage` in the beginning. (It would have been fine for me to have an empty page and being able to move the window back there when the user selects the now empty page.)

Will add the note to `RemovePage`.

DietmarSchwertberger:
Yes, the hint could be added to the notebook AddPage documentation.
I would suggest not to do this, though. Reparenting is a niche usage and it would rather confuse newcomers.

PR #26293, opened 2026-03-15
Fix showing `wxProgressDialog` with `wxPD_CAN_SKIP` in wxMSW

vadz:
Also make this code more robust.


vadz:
I'll merge this soon even in absence of feedback because it fixes a user-visible bug.

PR #26295, opened 2026-03-15
ArtProvider docs missing two art ids

mcorino:


vadz:
Thanks, will push soon.

PR #26296, opened 2026-03-15
Incorrect use of @onlyfor in toplevel.h doc

mcorino:
`wxTopLevelWindow::EnableFullScreenView` is incorrectly documented with @onlyfor{wxosx} as this method is  available for all platforms but only actually works on MacOS (as documented).

### DIFF
--- a/interface/wx/toplevel.h
+++ b/interface/wx/toplevel.h
@@ -719,8 +719,6 @@ public:
         Only @c ::wxFULLSCREEN_NOTOOLBAR and @c ::wxFULLSCREEN_NOMENUBAR will be
         used when using the fullscreen API (other values are ignored).
 
-        @onlyfor{wxosx}
-
         @see ShowFullScreen(), wxFullScreenEvent
 
         @since 3.1.0


vadz:
Shouldn't we document that this function doesn't do anything under non-macOS? I'll do this in addition to removing `@onlyfor`...

mcorino:
As I said the behaviour on other platforms is already correctly documented ;⁠-⁠)On 15 Mar 2026 21:31, VZ ***@***.***> wrote:vadz left a comment (wxWidgets/wxWidgets#26296)
Shouldn't we document that this function doesn't do anything under non-macOS? I'll do this in addition to removing @onlyfor...

vadz:
> As I said the behaviour on other platforms is already correctly documented

I must be blind, but I just don't see it. Anyhow, adding an explicit note about it shouldn't harm.

mcorino:
> > As I said the behaviour on other platforms is already correctly documented
> 
> I must be blind, but I just don't see it. Anyhow, adding an explicit note about it shouldn't harm.

I would have said this line was clear enough:
```
This function is currently only implemented in wxOSX and simply returns false in the other ports.
```
;-)

vadz:
Wait, I may be blind but not crazy (I think?). This sentence is part of my commit which added it, it hadn't been there before!

mcorino:
Oh my, lol. You are absolutely right.
The line I saw before was the description of the return value:

```
@return @true if the button behaviour has been changed, @false if running
        under another OS.
```

I mixed that up with what you added. My bad.

PR #26297, opened 2026-03-15
CMake: Automatically update outdated WebView2 package

MaartenBent:
And update the versions numbers and compile-time version check to the required version.

See #26291


vadz:

PR #26303, opened 2026-03-16
Add missing null check in wxAffineMatrix2D::Get

SerdarPe:
Variable "mat2D" wasn't being checked for null, despite "tr" was checked for it.

### DIFF
--- a/interface/wx/affinematrix2d.h
+++ b/interface/wx/affinematrix2d.h
@@ -28,9 +28,11 @@ public:
     /**
         Get the component values of the matrix.
 
+        At least one of the parameters must be non-NULL.
+
         @param mat2D
-            The rotational components of the matrix (upper 2 x 2), must be
-            non-null.
+            The rotational components of the matrix (upper 2 x 2),
+            may be @NULL.
         @param tr
             The translational components of the matrix, may be @NULL.
     */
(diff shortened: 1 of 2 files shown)


vadz:
This is consistent with the documentation, see interface/wx/affinematrix2dbase.h

```cpp
    /**
        Get the component values of the matrix.

        @param mat2D
            The rotational components of the matrix (upper 2 x 2), must be
            non-null.
        @param tr
            The translational components of the matrix, may be @NULL.
    */
    virtual void Get(wxMatrix2D* mat2D, wxPoint2DDouble* tr) const = 0;
```

Crashing is still not very user-friendly, but AFAICS it makes no sense to call this function with null matrix (do you do it intentionally?), so we should probably add a `wxCHECK_RET()` to it, but not silently ignore the invalid calls.

SerdarPe:
It occured when I wanted to get **just** the translation part of the affine matrix. I didn't need the transformation matrix in that specific part of my code.

vadz:
OK, but then we need to update the documentation (under `interface/wx`) to not say that this parameter is required any longer, could you please do it?

And I'd still add at least a `wxASSERT_MSG()` checking that both of parameters are not null at the same time.

SerdarPe:
Alright, I've updated the commit.

vadz:
Thanks, will merge soon.

PR #26304, opened 2026-03-17
wxScrolled<>:  documentation and sample improvements

wsu-cb:


samples/dnd/dnd.cpp
```diff
@@ -702,17 +688,30 @@ class DnDShapeDialog : public wxDialog
...
 
-class DnDShapeFrame : public wxScrolled<wxFrame>
+class DnDShapeFrame : public wxFrame
 {
 public:
     DnDShapeFrame(wxFrame *parent);
-    ~DnDShapeFrame();
+
+protected:
+    bool TryBefore(wxEvent& event) override;
```
vadz, line 701: I think overriding this function is sufficiently rare to merit a comment explaining why do we do this.

interface/wx/scrolwin.h
```diff
@@ -541,6 +553,23 @@ class wxScrolled : public T
     */
     void Scroll(const wxPoint& pt);
 
+    /**
+        Set this window to autoscroll even if it has not captured the mouse
+        (assuming the mouse cursor is in its autoscroll zone).  This is
+        intended to be called on drag-enter to support drag and drop between
+        windows.
+
+        @since 3.3.0
```
vadz:
```suggestion
        @since 3.3.3
```

interface/wx/scrolwin.h
```diff
@@ -541,6 +553,23 @@ class wxScrolled : public T
...
+    */
+    void EnableAutoscrollWithoutCapture();
+    /**
+        Undo EnableAutoscrollWithoutCapture().  This is intended to be called on
+        drag-exit and drag-drop when supporting drag and drop between windows.
+
+        @since 3.3.0
```
vadz:
```suggestion
        @since 3.3.3
```

samples/dnd/dnd.cpp
```diff
@@ -867,20 +866,20 @@ wxEND_EVENT_TABLE()
 
 #if wxUSE_DRAG_AND_DROP
 
-wxBEGIN_EVENT_TABLE(DnDShapeFrame, wxFrame)
-    EVT_MENU(Menu_Shape_New,    DnDShapeFrame::OnNewShape)
-    EVT_MENU(Menu_Shape_Edit,   DnDShapeFrame::OnEditShape)
-    EVT_MENU(Menu_Shape_Clear,  DnDShapeFrame::OnClearShape)
+wxBEGIN_EVENT_TABLE(DnDShapeCanvas, wxScrolledCanvas)
```
vadz, line 873: Not sure why did all these functions have to be moved to another class. I don't see anything wrong with it, per se, but this results in a lot of changes and I don't understand why do we need them?

vadz (commented):
Thanks, changes to the documentation are good, but I'm less sure about the changes to the sample (using `unique_ptr` is good, of course, but moving the handlers to another class is less obviously so).

samples/dnd/dnd.cpp
wsu-cb: The problem I am addressing is that, if I don't create a `DnDShapeCanvas` window that performs the scrolling, then the scrolling is part of the `DnDShapeFrame`, and that causes the status bar to move when the frame scrolls.  At that point, I think there are actually fewer changes by having the shape operations implemented by `DnDShapeCanvas` rather than `DnDShapeFrame`.

samples/dnd/dnd.cpp
wsu-cb: Commented in https://github.com/wxWidgets/wxWidgets/pull/26304/commits/b51b388f9a5d08ec9f744dd855e42ef528eec346

interface/wx/scrolwin.h
wsu-cb: I apologize for the careless mistake.

Fixed in https://github.com/wxWidgets/wxWidgets/pull/26304/commits/2e6689f8eecc8fe9855de7e61138c1fba341ca96

interface/wx/scrolwin.h
wsu-cb: At least I was consistent.

Fixed in https://github.com/wxWidgets/wxWidgets/pull/26304/commits/2e6689f8eecc8fe9855de7e61138c1fba341ca96

vadz:
Merged, thanks!

PR #26307, opened 2026-03-17
Backport wxNO_IMPLICIT_WXSTRING_CONV_TO_PTR to 3.2

vslavik:
This PR backports `wxNO_IMPLICIT_WXSTRING_CONV_TO_PTR` from 889845fbc4a89e2fba62c96e04d26b0cf5639cd1 to the 3.2 branch, so that code that may need to be compiled against e.g. distro-provided wx packages can rely on its effect.

Also serves as a mitigation for https://github.com/wxWidgets/wxWidgets/pull/26306

### DIFF
--- a/docs/doxygen/overviews/string.h
+++ b/docs/doxygen/overviews/string.h
@@ -248,7 +248,7 @@ is not appropriate for the current software and platform. The macro @c
 wxNO_IMPLICIT_WXSTRING_ENCODING disables all implicit conversions, and
 forces the code to explicitly indicate the encoding of all C strings.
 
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, will apply.

I was going to say that we should also document this in the [preprocessor constants documentation page](https://docs.wxwidgets.org/latest/page_cppconst.html), but realized that we don't do it in master either, and I wonder why: shouldn't this one be mentioned there?

vadz:
Applied as 626df737c936b6048742161eba7e99c499ec88ff.

Issue #26300, opened 2026-03-16
wxOSX wxToggleButton style regression(?)


ryancog:
#### Bug description:

The wxToggleButton now displays the same as a normal push button.

~~I assume this is a consequence of d6d9b70, I haven't looked at it deeply yet, but it's the only commit that sticks out.~~ EDIT: Looking more closely, this commit seems unlikely to have changed anything. I'm not sure in that case, and don't have an excellent routine to bisect things, so I suppose I'll defer.

#### Expected vs observed behaviour:

Earlier (v3.3.0) the wxToggleButton has a distinct, albeit dated style. While this old toggle button style is dated and rarely seen, it is still expected. Moreso at least than the push button style.

~~I'll struggle to find the PR or issue now, but~~
Found it: https://github.com/wxWidgets/wxWidgets/issues/22140

I know this was brought up previously and changed to what is seen in the v3.3.0 screenshot, which is why I call it a regression in the title.

v3.3.0:

<img width="454" height="120" alt="Image" src="https://github.com/user-attachments/assets/71ee4ceb-8a91-4294-95b8-a48bcdb90a30" />

v3.3.2:

<img width="414" height="118" alt="Image" src="https://github.com/user-attachments/assets/e52ac386-71e6-4fb6-89d4-40952e9c8889" />

### Platform and version information

- wxWidgets version  you use: v3.3.2
- wxWidgets port  you use: wxOSX
- OS  and its version: macOS Sequoia 15.7.1

vadz:
@csomor I'd like to ask for help with this one too because I don't even know which style is the correct one to use here (or maybe the correct one is something else?).

csomor:
@ryancog could you please test with the PR #26308, thanks. It mixes the look and the behavior, if this would not be acceptable we would revert to the rather outdated look

ryancog:
@csomor I have built the widgets sample on 3.3.0, 3.3.2, and your branch.

Interestingly, 3.3.2 actually looks correct normally. It looks like a pushbutton only when using a bitmap (which my application does)

On your branch, it appears as a push button regardless of using a bitmap.

On 3.3.0 it always appears as the old style.

It seems like reverting to the old style would be ideal, then.
The way that the pushbutton highlights when "toggled" makes it look like a default button, and of course a push button acting as a toggle is also unexpected.

3.3.0:

[...]

PR #26308, opened 2026-03-17
OSX Toggle Button Styling

csomor:
use one bottleneck procedure for ordinary and toggling buttons, see #26300

### DIFF
--- a/include/wx/osx/cocoa/private.h
+++ b/include/wx/osx/cocoa/private.h
@@ -50,7 +50,7 @@ wxBitmapBundle WXDLLIMPEXP_CORE wxOSXCreateSystemBitmapBundle(const wxString& id
 WXWindow WXDLLIMPEXP_CORE wxOSXGetMainWindow();
 WXWindow WXDLLIMPEXP_CORE wxOSXGetKeyWindow();
 WXImage WXDLLIMPEXP_CORE wxOSXGetNSImageFromNSCursor(const WXHCURSOR cursor);
-void WXDLLIMPEXP_CORE wxOSXSetBezelStyleFromBorderFlags(WX_NSButton v,
+void wxOSXSetBezelStyleFromBorderFlags(WX_NSButton v,
                              long style,
                              wxWindowID winid,
                              const wxString& label = wxString(),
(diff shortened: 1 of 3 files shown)


vadz:
Should this be merged or do you plan more changes here?

csomor:
> Should this be merged or do you plan more changes here?

wait a moment please, we have to decide whether to move forward, using a rounded style, or keeping the old look, and I want to document how to arrive at another style for a toggle button as well ...

src/osx/cocoa/button.mm
```diff
@@ -187,6 +199,7 @@ - (NSControlSize)controlSize;
 {
     NSButton* button = GetNSButton();
     [button setAlternateImage: wxOSXGetImageFromBundle(bitmap)];
+    NSLog(@"alternateImage height %d, normalImage height %d", (int)[button alternateImage].size.height, (int)[button image].size.height);
```
vadz: This should probably be removed?

include/wx/osx/cocoa/private.h
```diff
@@ -50,7 +50,12 @@ wxBitmapBundle WXDLLIMPEXP_CORE wxOSXCreateSystemBitmapBundle(const wxString& id
 WXWindow WXDLLIMPEXP_CORE wxOSXGetMainWindow();
 WXWindow WXDLLIMPEXP_CORE wxOSXGetKeyWindow();
 WXImage WXDLLIMPEXP_CORE wxOSXGetNSImageFromNSCursor(const WXHCURSOR cursor);
-
+void WXDLLIMPEXP_CORE wxOSXSetBezelStyleFromBorderFlags(WX_NSButton v,
```
vadz: Is `WXDLLIMPEXP_CORE` really needed here? I.e. is it really going to be used from outside the core library? I think it doesn't need to be exported from it.

vadz (commented):
Thanks, just to confirm: should this one be merged now or should we wait more for it to be tested?

include/wx/osx/cocoa/private.h
csomor: you are right, I'll change that

src/osx/cocoa/button.mm
csomor: yes, I'll do that

csomor:
> Thanks, just to confirm: should this one be merged now or should we wait more for it to be tested?

fine to merge for me, we need broad testing now, if we run into trouble, I'll re-add the last reverted commit, which makes things less beautiful

Issue #26301, opened 2026-03-16
wxOSX dynamic background color broken


ryancog:
### Description

The background color for windows on macOS does not dynamically update, and instead the color is static and has an (oddly) orange tint to it.

On macOS, the background color of windows (and groups/static boxes, etc.) is dynamic by default depending on the desktop wallpaper and the window's position on the desktop.

Moving to wxWidgets v3.3.2, this behavior is no longer working. Things were last working as expected for me on v3.3.0. I see there's been lots of changes to the wxOSX port over these versions (relating to both the unfortunate macOS 26 and some iOS work). Browsing through the commit history it seems 8ada351adf and 5d41b24173 are potential culprits.

Of course, in v3.3.0 (and normally) the dynamic background makes a screenshot not entirely useful, but perhaps it is to see the v3.3.2 static color(s) and the side-by-side.

v3.3.0 on left, v3.3.2 on right:

<img width="130" height="38" alt="Image" src="https://github.com/user-attachments/assets/6bd60cbb-9e0f-49e4-86b2-bfa4d37322e1" />

<img width="158" height="40" alt="Image" src="https://github.com/user-attachments/assets/45a36136-3f39-43eb-8d2d-039f2b6dafe4" />

It is much less noticeable in light mode (and the shadow muddies things a bit more), but still present.

### Platform and version information

- wxWidgets version  you use: 3.3.2
- wxWidgets port  you use: wxOSX
- OS  and its version: macOS Sequoia 15.7.1

vadz:
@csomor I hope you could look at this, TIA!

csomor:
@ryancog in order to help me, could you please pick a sample app which shows these things the best (perhaps widgets ?) and a 3.3.2 and 3.3.0 screenshot, and give me a step-by-step instruction to arrive at the same window content. Then I'll try to preserve as much as possible without ruining the Tahoe fixes. Thanks

ryancog:
Here I have the widgets demo with 3.3.2 and 3.3.0

Here's 3.3.2:
It does not change based on position or size.

<img width="2860" height="1610" alt="Image" src="https://github.com/user-attachments/assets/cc6f5ee9-5aec-47f5-b0ac-2055ec058841" />

And for 3.3.0, I've taken a few screenshots around my desktop. My wallpaper is predominantly blue, so there's a blue tinge on the window (and then it changes just depending on the specific are of the wallpaper, it also changes based on the window size/area of the wallpaper occluded)

[...]

PR #26309, opened 2026-03-18
OSX transparent background

csomor:
fixes regression, allow tlw backgrounds changing according to desktop background, see #26301

### DIFF
--- a/include/wx/osx/window.h
+++ b/include/wx/osx/window.h
@@ -126,6 +126,14 @@ public:
 
 #endif
 
+    virtual wxVisualAttributes GetDefaultAttributes() const
+    {
+        return GetClassDefaultAttributes(GetWindowVariant());
+    }
+
+    static wxVisualAttributes
+    GetClassDefaultAttributes(wxWindowVariant variant = wxWINDOW_VARIANT_NORMAL);
+
     // implementation from now on
     // --------------------------
 
--- a/include/wx/osx/window.h
+++ b/include/wx/osx/window.h
@@ -126,7 +126,7 @@ public:
 
 #endif
 
-    virtual wxVisualAttributes GetDefaultAttributes() const
+    virtual wxVisualAttributes GetDefaultAttributes() const override
     {
         return GetClassDefaultAttributes(GetWindowVariant());
     }
--- a/src/osx/window_osx.cpp
+++ b/src/osx/window_osx.cpp
@@ -1514,6 +1514,17 @@ wxWindowMac::AlwaysShowScrollbars(bool hflag, bool vflag)
         DoUpdateScrollbarVisibility();
 }
 
+/* static */ wxVisualAttributes
+wxWindowMac::GetClassDefaultAttributes(wxWindowVariant variant)
+{
+    // it is important to return valid values for all attributes from here,
+    // GetXXX() below rely on this
+    wxVisualAttributes attrs = wxWindowBase::GetClassDefaultAttributes(variant);
+    attrs.colBg = wxTransparentColour;
+
+    return attrs;
+}
+
 //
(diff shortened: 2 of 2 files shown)


vadz:
Stefan, should I merge this one or do you think it needs more testing (unlikely to happen as long as it's not merged, IME...)?

csomor:
> Stefan, should I merge this one or do you think it needs more testing (unlikely to happen as long as it's not merged, IME...)?

Hi Vadim, no sorry, #26301 already mentions some things that don't look right

vadz:
OK, I'll mark it as being a draft too then — please reset the draft status when it's ready to merge (or close it if it finally doesn't work). TIA!

ryancog:
@csomor For wxTextCtrl (and wxSpinCtrl, wxSearchCtrl), you're just missing an override attributes getter, since wxTextCtrl has its own override which uses `wxControlBase::GetCompositeControlsDefaultAttributes()`.

The following patch fixes the appearance for me:
```
diff --git a/src/common/ctrlcmn.cpp b/src/common/ctrlcmn.cpp
index 672139f3e8..3865ed8c3c 100644
--- a/src/common/ctrlcmn.cpp
+++ b/src/common/ctrlcmn.cpp
@@ -236,7 +236,11 @@ wxControlBase::GetCompositeControlsDefaultAttributes(wxWindowVariant WXUNUSED(va
     wxVisualAttributes attrs;
     attrs.font = wxSystemSettings::GetFont(wxSYS_DEFAULT_GUI_FONT);
     attrs.colFg = wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOWTEXT);
+#   ifdef __WXOSX__
+    attrs.colBg = wxTransparentColour;
+#   elif
     attrs.colBg = wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOW);
+#   endif

     return attrs;
 }
```

For some reason I don't understand, adding an override in the OSX `wxTextCtrl` in the same vein as what you've done in this PR branch for `wxWindowMac`/`wxWindow` doesn't seem to apply to the `wxSearchCtrl`... maybe I'm missing something though.

EDIT: It also didn't apply to `wxSpinCtrl`, but I realize that it does just use `wxTextCtrl`... I don't know why adding an override to `wx/osx/textctrl.h` wouldn't apply to `wxSpinCtrl` or `wxSearchCtrl`. Unless I had something weird/stale in my build. I'll retry this in a bit.

~~In any case, the `wxSpinCtrl` uses a generic implementation on wxOSX, (as does `wxOwnerDrawnComboBox` and `wxBitmapComboBox`) and so the appearance needs to be overridden there too, but these all use `wxControlBase::GetCompositeControlsDefaultAttributes()`, so it seems simpler to just set the transparent background there.~~

Beyond those, the only other abnormality I see is with `wxDatePickerCtrl`, `wxTimePickerCtrl` (those two I figure is a similar case), and `wxListCtrl` and friends, which I'll look into next.

Of course, there's the other generic controls I mentioned which have a opaque background on 3.3.0 also, which I want to look into as well...

ryancog:
As something of an aside, the `controlBackgroundColor` used for `wxSYS_COLOUR_WINDOW` and `wxSYS_COLOUR_LISTBOX` is noted in the [documentation](https://developer.apple.com/documentation/appkit/nscolor/controlbackgroundcolor?language=objc):

> Do not use this color for drawing. Instead, use an [NSVisualEffectView](doc://com.apple.documentation/documentation/appkit/nsvisualeffectview?language=occ) with the appropriate background material.

So it seems like maybe this is the wrong color to return in `wxSystemSettingsNative::GetColour()`?

vadz:
Merged as 565553ee98 (Make default background color transparent in wxOSX, 2026-03-18)

PR #26311, opened 2026-03-21
 Add Dark Mode Support for TaskDialog (wxMSW)

### LINKED ISSUE #26240: [MSW] [DarkMode] Enable Dark Mode support for native TaskDialog
## Description
This issue addresses the lack of dark mode support in TaskDialog-based components when `wxApp::MSWEnableDarkMode()` is enabled. Currently, the following dialogs appear with light backgrounds even when the application is in dark mode:
- wxMessageDialog
- wxRichMessageDialog
- wxProgressDialog  
- wxAboutBox (simple version)

This is a documented limitation (see [MSWEnableDarkMode documentation](https://docs.wxwidgets.org/latest/classwx_app.html#af8c93d7e3345e62a58325f3ab1d158d6)) because Windows doesn't natively provide dark TaskDialogs.

## Proposed Enhancement
I have developed a working solution that successfully enables dark mode for all TaskDialog-based dialogs on Windows. The implementation:
- **Integrates with the existing wxMSWDarkMode infrastructure**
- **Works on both Windows 10 and Windows 11**
- **Preserves all native dialog functionality**
- **Includes graceful fallback** when dark mode cannot be applied
- **Is opt-in** - only affects applications that enable dark mode

### Windows 11
https://github.com/user-attachments/assets/91d3117f-c5b8-4f0a-a78c-d87dc15e880e

### Windows 10
https://github.com/user-attachments/assets/c01e5e34-6e32-48b3-a7ce-3d191d4d51d8

## Testing
✅ Windows 11 (all recent builds)
✅ Windows 10 (versions 1809 through 22H2)  
✅ All TaskDialog-based wxWidgets components
✅ Light/Dark system theme switching
✅ Backward compatibility with non-dark mode apps

## Questions for the Team
1. Would you be interested in a PR that implements dark mode support for TaskDialog-based dialogs?
2. Are there any concerns about the approach that I should address before preparing a PR?
3. Would you prefer this to be enabled automatically when dark mode is active, or should it have a separate opt-in flag?

## Next Steps
I have a working implementation ready and can prepare a PR if the team agrees this enhancement would be valuable. The code is:
- **Non-intrusive** - only affects dark mode-enabled applications
- **Backward compatible** - existing behavior unchanged for non-dark mode apps
- **Well-tested** across Windows versions

---

This would resolve the previously closed issue #24973 by providing a practical solution that maintains native dialog behavior while adding dark mode support.

memoarfaa:
# Add Dark Mode Support for TaskDialog (wxMSW)

## Summary

This PR adds full dark mode theming support for the native Windows `TaskDialog` API as used by `wxMessageDialog` and `wxProgressDialog` on Windows. When the application enables dark mode via `wxApp::MSWEnableDarkMode()`, the TaskDialog and its child controls now render consistently with the rest of the application's dark theme instead of appearing with the default light system colours.
 

Fixes #26240
---

## Motivation

Previously, calling `wxApp::MSWEnableDarkMode()` successfully applied dark mode to all native `wxWidgets` controls and top-level windows, but the native `TaskDialog` (used by `wxMessageDialog` and `wxProgressDialog`) was left entirely unstyled. This created an obvious visual inconsistency: the rest of the application appeared dark while every message box and progress dialog retained bright white panels.

---

### `src/msw/darkmode.cpp`

This is the core of the implementation. The following were added to the existing `wxMSWDarkMode` namespace:

**Colour palette (`TDDarkCol` namespace)**
A dedicated set of `constexpr COLORREF` constants matching `wxDarkModeSettings::GetColour()` values for the primary panel, secondary panel, footnote area, separator, and all text roles (instruction, content, expando, verification, footnote, radio button).

**Per-dialog state (`TDPageState` struct)**
A `thread_local` map (`tls_tdStates`) stores rendering state per `HWND`, including:
- Open theme handles (`HTHEME`) for `TaskDialog`, `TaskDialogStyle`, and `Button` parts
- GDI brushes for each background zone
- A UIA-backed layout cache (`std::vector<TDLayoutElement>`) with bounding rects and automation IDs for each UI element
- Mouse interaction state (hover/press/tracking) for interactive controls
- Logical state (`isExpanded`, `isChecked`) mirrored from window properties

**UIA layout cache (`TDBuildLayoutCache` / `TDUpdateLayoutCache`)**
UI Automation is used to walk the TaskDialog child tree and build a cache of element bounding rects and automation IDs. This avoids hooking individual child windows for hit testing and allows the paint routine to overdraw text and glyphs without relying on private Win32 control classes.

**Paint routines**
- `TDPaintPixelSwap` — On Windows 10 (no native dark TaskDialog theme), performs a pixel-by-pixel colour substitution on the buffered surface, replacing the light `TaskDialog` panel colours with the dark palette.
- `TDPaintIcons` — Redraws stock icons (`TD_WARNING_ICON`, `TD_ERROR_ICON`, etc.) over the correct dark background brush using `SHGetStockIconInfo`.
- `TDPaintGlyphs` — Redraws the expand/collapse triangle (`TDLG_EXPANDOBUTTON`) and verification checkbox (`BP_CHECKBOX`) using themed parts with correct hot/pressed states.
- `TDPaintText` — Overdraw all text elements (main instruction, content, expando label, verification label, footnote, expanded footer) using `DrawThemeTextEx` with explicit `COLORREF` values from the dark palette.
- `TDPaintPage` — Orchestrates the above into a single double-buffered `BeginBufferedPaint` / `EndBufferedPaint` pass.

**Subclass procedures**
- `TDPageSubclassProc` — Installed on the inner DirectUI `TaskDialog` child window. Handles `WM_PAINT`, `WM_ERASEBKGND`, mouse messages for hover/press tracking, `WM_THEMECHANGED`, and `WM_DESTROY`.
- `TDCtrlContainerSubclassProc` — Installed on container panels to handle `WM_CTLCOLOR*` messages, setting the correct text and background colour for child static/edit controls.
- `TDRadioButtonSubclassProc` — Installed on individual radio button controls to render their text in dark colour using `DrawThemeTextEx`.

**Attachment/detachment helpers (`TDAttach` / `TDDetach`)**
- `TDAttach` enumerates the dialog's child windows via UIA, applies `SetWindowTheme` to each, installs the subclasses above, sets the class background brush, applies dark title bar (`EnableForTLW`), and broadcasts `WM_SYSCOLORCHANGE` / `WM_THEMECHANGED`.
- `TDDetach` removes all subclasses and restores default themes during `TDN_DESTROYED`.

**Public entry points added to `wxMSWDarkMode` namespace**
```cpp
void AllowForTaskDialog(HWND hwnd, const TASKDIALOGCONFIG* pCfg);
void RemoveFromTaskDialog(HWND hwnd);
```

**Windows 11 native dark theme detection (`TDHasNativeDarkTheme`)**
A one-time probe checks whether `DarkMode_DarkTheme::TaskDialog` is available and distinct from the standard `TaskDialog` theme. When available (Windows 11+), the pixel-swap and manual overdraw paths are skipped and the OS-supplied theme is used directly.

---

### `src/msw/msgdlg.cpp`

**`wxTaskDialogCallback`** (the `TASKDIALOGCONFIG::pfCallback`) was updated to call the new dark mode entry points at the appropriate lifecycle events:

| Notification | Action |
|---|---|
| `TDN_CREATED` | Calls `wxMSWDarkMode::AllowForTaskDialog(hwnd, pCfg)` to install dark mode after the full HWND tree is built but before the dialog is shown. Uses `TDN_CREATED` instead of the previous `TDN_DIALOG_CONSTRUCTED` because UIA cannot reliably walk child windows at the earlier event. |
| `TDN_EXPANDO_BUTTON_CLICKED` | Sets the `IsExpanded` window property so the subclassed page panel can read the state during `WM_PAINT` without an additional UIA call. |
| `TDN_DESTROYED` | Calls `wxMSWDarkMode::RemoveFromTaskDialog(hwnd)` to clean up all subclasses and GDI objects. |
| `WM_SETTINGCHANGE` | Re-applies or removes theming when the OS light/dark preference changes while the dialog is open. |

`lpCallbackData` is set to `&tdc` (the `TASKDIALOGCONFIG` pointer) so the callback can forward it to `AllowForTaskDialog` for icon and flag inspection.

---

### `src/msw/progdlg.cpp`

`wxProgressDialogTaskRunner::TaskDialogCallbackProc` was updated with the same `TDN_CREATED` / `TDN_DESTROYED` / `WM_SETTINGCHANGE` handling as `msgdlg.cpp`. Because `wxProgressDialog` does not carry a `TASKDIALOGCONFIG` pointer at the callback site, `AllowForTaskDialog` is called with `nullptr` for `pCfg`, which disables the icon-overdraw path (acceptable since progress dialogs rarely use stock icons).

---

## Platform & Version Compatibility

| Windows version | Behaviour |
|---|---|
| Windows 10 (1809+) | Full dark mode via pixel-swap + manual overdraw of text, glyphs, and icons. Dark title bar applied via `DwmSetWindowAttribute`. |
| Windows 11 | Uses native `DarkMode_DarkTheme::TaskDialog` theme where available; pixel-swap and overdraw paths are skipped. |
| Windows < 1809 | `wxMSWImpl::ShouldUseDarkMode()` returns `false`; all new code paths are bypassed entirely. |
| Dark mode disabled | `AllowForTaskDialog` returns immediately. No impact on normal (light mode) rendering. |

---

## Testing

Tested on:
- Windows 10 21H2 — pixel-swap path, all dialog variants (info, warning, error, shield icon, progress, expandable, verification checkbox, radio buttons, command links, footnote)
- Windows 11 24H2 — native dark theme path
- Light mode on both versions — no visual regression

To reproduce, call `wxApp::MSWEnableDarkMode()` in `OnInit()` and show a `wxMessageDialog` or `wxProgressDialog`.

---
Screenshot 2026-03-05 052827

Screenshot 2026-03-03 053105

## Related Issues / References

- reference implementation: [`DarkTaskDialog-Native`](https://github.com/memoarfaa/DarkTaskDialog-Native)


memoarfaa:
Hi @vadz 
Could you advise me on whether adding files to darkmode.cpp and darkmode.h is better, or creating separate files for dark mode taskdialog ?

include/wx/msw/private/darkmode.h
```diff
@@ -11,6 +11,7 @@
 #define _WX_MSW_PRIVATE_DARKMODE_H_
 
 #include "wx/settings.h"
+#include <commctrl.h>
```
vadz: Minor, but we usually do

```suggestion
#include "wx/msw/wrapcctl.h"
```

src/msw/progdlg.cpp
```diff
@@ -32,6 +32,7 @@
 
 #include "wx/msw/private/msgdlg.h"
 #include "wx/evtloop.h"
+#include <wx/msw/private/darkmode.h>
```
vadz, line 35: Even more minor, but:


```suggestion
#include "wx/msw/private/darkmode.h"
```

src/msw/progdlg.cpp
```diff
@@ -1143,30 +1144,59 @@ wxProgressDialogTaskRunner::TaskDialogCallbackProc
...
-            if ( !(sharedData->m_style & wxPD_CAN_ABORT) )
-                EnableCloseButtons(hwnd, false);
-            break;
-
-        case TDN_BUTTON_CLICKED:
+    case TDN_CREATED:
+        // TDN_CREATED fires after the full HWND hierarchy is built and
+        // before the window is shown — the right moment to install theming.
+        // (The old code used TDN_DIALOG_CONSTRUCTED which fires too early
+        // for UI Automation to walk the child windows reliably.)
```
vadz, line 1152: This was copy-pasted from `wxMessageDialog` but doesn't make sense here, this code never used `TDN_DIALOG_CONSTRUCTED`, so let's just remove it:


```suggestion
```

src/msw/msgdlg.cpp
```diff
@@ -584,13 +584,44 @@ namespace
...
+        break;
+
+    case TDN_DESTROYED:
+        // Clean up all subclasses and GDI/theme objects.
+        // Without this call, GDI handles and thread_local state leak
+        // when the dialog is destroyed.
+        wxMSWDarkMode::RemoveFromTaskDialog(hwnd);
+        break;
+
+    case WM_SETTINGCHANGE:
```
vadz, line 615: Do we actually get this here? [MSDN docs](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nc-commctrl-pftaskdialogcallback) don't mention it as a possible value.

src/msw/darkmode.cpp
```diff
@@ -44,6 +44,13 @@
 
 #include "wx/msw/private/darkmode.h"
 
+#include <uiautomation.h>
+#include <vssym32.h>
+#include <commctrl.h>
+#include <atlbase.h>    // CComPtr, CComBSTR
```
vadz: We don't use MSVC-specific ATL in our code, please use `wxCOMPtr` and `wxBasicString` instead.

src/msw/darkmode.cpp
```diff
@@ -125,14 +132,17 @@ wxDarkModeModule::ms_pfnDwmSetWindowAttribute = (DwmSetWindowAttribute_t)-1;
 #include "wx/msw/private/custompaint.h"
 #include "wx/msw/private/menu.h"
 #include "wx/msw/private/metrics.h"
-
 #include <memory>
 
 #if wxUSE_LOG_TRACE
 static const char* TRACE_DARKMODE = "msw-darkmode";
 #endif // wxUSE_LOG_TRACE
+#include <atlwin.h>
```
vadz: Also please remove/replace this.

src/msw/darkmode.cpp
```diff
@@ -164,6 +174,781 @@ bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
...
+{
+    HTHEME hTD = nullptr; // TaskDialog panel + glyph parts
+    HTHEME hTDS = nullptr; // TaskDialogStyle / text fonts
+    HTHEME hButton = nullptr; // Button (checkbox glyph)
+    bool   isDark = false;   // native dark theme available
+    bool   themesOk = false;
+
+    HBRUSH brPrimary = nullptr;
+    HBRUSH brSecondary = nullptr;
+    HBRUSH brFootnote = nullptr;
```
vadz, line 214: And you definitely could use `AutoHBRUSH` for those, to avoid having to manually delete them.

src/msw/darkmode.cpp
```diff
@@ -164,6 +174,781 @@ bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
...
+    std::wstring name;        // visible text (used for overdraw)
+    LONG         legacyState = 0; // STATE_SYSTEM_* flags
+};
+
+// All per-dialog rendering state, stored in a thread_local map.
+struct TDPageState
+{
+    HTHEME hTD = nullptr; // TaskDialog panel + glyph parts
+    HTHEME hTDS = nullptr; // TaskDialogStyle / text fonts
+    HTHEME hButton = nullptr; // Button (checkbox glyph)
```
vadz: You could probably use `wxUxThemeHandle` for those instead of closing them manually.

src/msw/darkmode.cpp
```diff
@@ -164,6 +174,781 @@ bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
...
+        if (hButton) { CloseThemeData(hButton); hButton = nullptr; }
+        themesOk = false;
+    }
+    void DestroyBrushes()
+    {
+        if (brPrimary) { DeleteObject(brPrimary);   brPrimary = nullptr; }
+        if (brSecondary) { DeleteObject(brSecondary); brSecondary = nullptr; }
+        if (brFootnote) { DeleteObject(brFootnote);  brFootnote = nullptr; }
+    }
+    void Destroy() { CloseThemes(); DestroyBrushes(); elements.clear(); elemsOk = false; }
```
vadz, line 261: This should be the dtor, as it's only called before destroying the object anyhow. But if the raw handles are replaced with RAII helpers, then this could be removed entirely.

src/msw/darkmode.cpp
```diff
@@ -164,6 +174,781 @@ bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
...
+// TaskDialog theme helpers
+// ============================================================================
+
+// Cached probe: does the OS have a native dark TaskDialog theme (Win11+)?
+static bool TDHasNativeDarkTheme()
+{
+    static int s_cached = -1;
+    if (s_cached == -1)
+    {
+        HTHEME hD = OpenThemeData(nullptr, L"DarkMode_DarkTheme::TaskDialog");
```
vadz: This should use `wxUxThemeHandle` too.

src/msw/darkmode.cpp
```diff
@@ -164,6 +174,781 @@ bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
...
+    }
+}
+
+// ============================================================================
+// Icon loading
+// ============================================================================
+
+static HICON TDLoadStockIcon(const TASKDIALOGCONFIG* cfg, bool isMain)
+{
+    if (!cfg) return nullptr;
```
vadz: I'm not going to leave comments about this everywhere but it would be nice to reformat the code to follow the usual wx style, i.e.


```suggestion
    if ( !cfg )
        return nullptr;
```

vadz (commented):
Thanks a lot for all this work, it's pretty impressive — and the idea of using UI Automation for this is great.

This mostly looks fine to me, although I admit I didn't look into all the details, the only thing which definitely needs to be done is to replace ATL classes with wx ones, but hopefully it shouldn't be very difficult.

One other annoying thing is that it would be nice to avoid having to link with `uiautomation.lib`, as this breaks any existing makefiles for the applications using static wx libs with MinGW (for MSVC we could use `#pragma comment(lib)`...). What exactly do we need from it? If it's just the UUID, I'd simply hardcode it instead.

Concerning the question of whether it should all go to this file: I think it would be better to create a separate `darktaskdialog.cpp` or something like this.

src/msw/darkmode.cpp
memoarfaa: If style look good now.

src/msw/darkmode.cpp
memoarfaa: removed

src/msw/darkmode.cpp
memoarfaa: If use wxCOMPtr and wxBasicString look good now?

src/msw/progdlg.cpp
memoarfaa: This not copy-pasted from wxMessageDialog from wxMessageDialog.
it's my old comment for don't use AllowForTaskDialog at TDN_DIALOG_CONSTRUCTED it's need TDN_CREATE

src/msw/darkmode.cpp
```diff
@@ -164,6 +167,902 @@ bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
...
+                BSTR bId; pChild->get_CurrentAutomationId(&bId);
+                const std::wstring id(bId ? static_cast<LPCWSTR>(bId) : L"");
+                HWND hP = GetParent(hBtn);
+
+                if (ct == UIA_ProgressBarControlTypeId)
+                {
+                    HTHEME hCE = OpenThemeData(nullptr, L"DarkMode_CopyEngine::Progress");
+                    HTHEME hBase = OpenThemeData(nullptr, L"Progress");
+                    bool hasCE = (hCE && hCE != hBase);
+                    if (hCE) CloseThemeData(hCE); if (hBase) CloseThemeData(hBase);
```
vadz: This statement provokes gcc error in one of the CI builds. As mentioned before, this could be avoided entirely by using `wxUxThemeHandle` instead of manual open/close theme data calls.

src/msw/darkmode.cpp
```diff
@@ -24,6 +24,7 @@
 
 // for compilers that support precompilation, includes "wx.h".
 #include "wx/wxprec.h"
+#include "wx/module.h"
```
vadz, line 27: Was this moved intentionally? Not really a problem, of course, but just looks weird.

src/msw/darkmode.cpp
```diff
@@ -164,6 +167,902 @@ bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
...
+// ============================================================================
+
+static LRESULT CALLBACK TDPageSubclassProc(
+    HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam,
+    UINT_PTR uId, DWORD_PTR)
+{
+    switch (msg)
+    {
+    case WM_ERASEBKGND:
+        return 1;
```
vadz, line 772: Minor, but this should be

```suggestion
        return TRUE;
```

src/msw/darkmode.cpp
```diff
@@ -164,6 +167,902 @@ bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
...
+    case WM_CTLCOLORBTN:
+    case WM_CTLCOLORDLG:
+    case WM_CTLCOLORSCROLLBAR:
+    case WM_CTLCOLORSTATIC:
+    {
+        HDC hdc = reinterpret_cast<HDC>(wParam);
+        COLORREF bg = TDDarkCol::kSecondary;
+        if (dwRef) { LOGBRUSH lb = {}; GetObject(reinterpret_cast<HBRUSH>(dwRef), sizeof(lb), &lb); if (lb.lbStyle == BS_SOLID) bg = lb.lbColor; }
+        SetBkColor(hdc, bg); SetTextColor(hdc, TDDarkCol::kTextNormal);
+        return reinterpret_cast<LRESULT>(dwRef ? reinterpret_cast<HBRUSH>(dwRef) : CreateSolidBrush(TDDarkCol::kSecondary));
```
vadz, line 864: Isn't this brush leaked?

src/msw/darkmode.cpp
```diff
@@ -164,6 +167,902 @@ bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
...
+    bool isExpanded = false;
+    bool isChecked = false;
+    bool defExpanded = false;
+    bool defChecked = false;
+
+    const TASKDIALOGCONFIG* pCfg = nullptr;
+
+    void CloseThemes()
+    {
+        if (hTD) { CloseThemeData(hTD);     hTD = nullptr; }
```
vadz: It would be really nice to avoid putting more than one statement per line, this improves readability a lot, in general, e.g.


```suggestion
        if (hTD)
        {
            CloseThemeData(hTD);
            hTD = nullptr;
        }
```

vadz (commented):
Thanks for the updates, but there are still some build problems. They look easy to fix (`WINAPI` missing in one place, just fix formatting — or, preferably, switch to using `wxUxThemeHandle` — in another) and I'll try to review this once again and merge as soon as this is done. Or I could just fix them myself if you prefer, please let me know.

vadz:
Just wanted to check: do you plan to make more changes here or should I merge this (I'll make some changes myself if I do it)?

MaartenBent:
I (finally) tried this, and it looks great. One of Vadz remarks is about leaking a brush, and this (or other leaks) are indeed happening. After showing the about dialog in the minimal sample and closing the app, Visual Studio reports multiple memory leaks.

On both Windows 10 and 11 the progress dialog has a white bar. The screenshot in the reference implementation has a dark bar. When I run the reference implementation (on Windows 10) it is also white. Is this a bug?

Screenshot 2026-06-01 233203

vadz:
> After showing the about dialog in the minimal sample and closing the app, Visual Studio reports multiple memory leaks.

FWIW I do plan to fix this and merge this PR soon, i.e. before 3.3.3.

> On both Windows 10 and 11 the progress dialog has a white bar.

I think this is indeed a bug, but it's still worth merging even without fixing it.

MaartenBent:
I was using a Windows 11 24h2, it might only work in newer Windows 11 that has `DarkMode_DarkTheme` or better `DarkMode_CopyEngine` themes. I'll try that later.

src/msw/msgdlg.cpp
vadz: I've tested and this code is never executed when changing system theme, so I'm removing it.

vadz:
@MaartenBent I don't see any leaks when showing the "About" dialog in the "dialogs" sample and ASAN doesn't report any problems, how can I reproduce this?

I do get weird messages like this in the debugger output

```
mincore\com\oleaut32\dispatch\ups.cpp(2122)\OLEAUT32.dll!00007FFF0964A726: (caller: 00007FFF09649B39) ReturnHr(1) tid(36dc) 8002801D Library not registered.
```

when instantiating `IUIAutomation`. They seem to be harmless but still suspicious and might indicate that we do something wrong, but I don't see what (the TID in the message above is that of the main thread, FWIW).

MaartenBent:
I have Windows 10 in light mode and add `MSWEnableDarkMode(DarkMode_Always);` to `MyApp::OnInit()` in the minimal sample to enable dark mode. Then start it (visual studio 2022, debug build), open about, close the app, and it reports leaks. Same in the dialogs sample.
```
Detected memory leaks!
Dumping objects ->
{43903} normal block at 0x00000163597855D0, 128 bytes long.
 Data: <  zYc     zYc   > 10 07 7A 59 63 01 00 00 10 07 7A 59 63 01 00 00 
{43902} normal block at 0x000001635713B4E0, 16 bytes long.
 Data: <  uYc           > A0 D6 75 59 63 01 00 00 00 00 00 00 00 00 00 00 
{43901} normal block at 0x00000163597A0710, 136 bytes long.
 Data: <  zYc     zYc   > 10 07 7A 59 63 01 00 00 10 07 7A 59 63 01 00 00 
{43900} normal block at 0x000001635713AC20, 16 bytes long.
 Data: <  uYc           > 88 D6 75 59 63 01 00 00 00 00 00 00 00 00 00 00 
{43516} normal block at 0x000001635971D670, 128 bytes long.
 Data: <0 nYc   0 nYc   > 30 06 6E 59 63 01 00 00 30 06 6E 59 63 01 00 00 
{43515} normal block at 0x0000016357145500, 16 bytes long.
 Data: <  rYc           > 90 92 72 59 63 01 00 00 00 00 00 00 00 00 00 00 
{43514} normal block at 0x00000163596E0630, 136 bytes long.
 Data: <0 nYc   0 nYc   > 30 06 6E 59 63 01 00 00 30 06 6E 59 63 01 00 00 
{43513} normal block at 0x00000163571454B0, 16 bytes long.
 Data: <x rYc           > 78 92 72 59 63 01 00 00 00 00 00 00 00 00 00 00 
{43509} normal block at 0x00000163571A9190, 128 bytes long.
 Data: <  nYc     nYc   > 20 0F 6E 59 63 01 00 00 20 0F 6E 59 63 01 00 00 
{43508} normal block at 0x000001635713D420, 16 bytes long.
 Data: < ,qYc           > E0 2C 71 59 63 01 00 00 00 00 00 00 00 00 00 00 
{43507} normal block at 0x00000163596E0F20, 136 bytes long.
 Data: <  nYc     nYc   > 20 0F 6E 59 63 01 00 00 20 0F 6E 59 63 01 00 00 
{43506} normal block at 0x000001635713D4C0, 16 bytes long.
 Data: < ,qYc           > C8 2C 71 59 63 01 00 00 00 00 00 00 00 00 00 00 
{43502} normal block at 0x00000163571A9E50, 128 bytes long.
 Data: <@ nYc   @ nYc   > 40 17 6E 59 63 01 00 00 40 17 6E 59 63 01 00 00 
{43501} normal block at 0x0000016357140E90, 16 bytes long.
 Data: < YpYc           > A0 59 70 59 63 01 00 00 00 00 00 00 00 00 00 00 
{43500} normal block at 0x00000163596E1740, 136 bytes long.
 Data: <@ nYc   @ nYc   > 40 17 6E 59 63 01 00 00 40 17 6E 59 63 01 00 00 
{43499} normal block at 0x0000016357140D00, 16 bytes long.
 Data: < YpYc           > 88 59 70 59 63 01 00 00 00 00 00 00 00 00 00 00 
{43495} normal block at 0x00000163571A81D0, 128 bytes long.
 Data: <P nYc   P nYc   > 50 0E 6E 59 63 01 00 00 50 0E 6E 59 63 01 00 00 
{43494} normal block at 0x0000016357140CB0, 16 bytes long.
 Data: <  oYc           > E0 96 6F 59 63 01 00 00 00 00 00 00 00 00 00 00 
{43493} normal block at 0x00000163596E0E50, 136 bytes long.
 Data: <P nYc   P nYc   > 50 0E 6E 59 63 01 00 00 50 0E 6E 59 63 01 00 00 
{43492} normal block at 0x0000016357140EE0, 16 bytes long.
 Data: <  oYc           > C8 96 6F 59 63 01 00 00 00 00 00 00 00 00 00 00 
{43488} normal block at 0x00000163571A84D0, 128 bytes long.
 Data: <  nYc     nYc   > F0 02 6E 59 63 01 00 00 F0 02 6E 59 63 01 00 00 
{43487} normal block at 0x0000016357140C10, 16 bytes long.
 Data: < hoYc           > C0 68 6F 59 63 01 00 00 00 00 00 00 00 00 00 00 
{43486} normal block at 0x00000163596E02F0, 136 bytes long.
 Data: <  nYc     nYc   > F0 02 6E 59 63 01 00 00 F0 02 6E 59 63 01 00 00 
{43485} normal block at 0x0000016357140C60, 16 bytes long.
 Data: < hoYc           > A8 68 6F 59 63 01 00 00 00 00 00 00 00 00 00 00 
{43481} normal block at 0x00000163571A8410, 128 bytes long.
 Data: <  nYc     nYc   > 20 02 6E 59 63 01 00 00 20 02 6E 59 63 01 00 00 
{43480} normal block at 0x00000163571408F0, 16 bytes long.
 Data: <   Xc           > 10 B6 BA 58 63 01 00 00 00 00 00 00 00 00 00 00 
{43479} normal block at 0x00000163596E0220, 136 bytes long.
 Data: <  nYc     nYc   > 20 02 6E 59 63 01 00 00 20 02 6E 59 63 01 00 00 
{43478} normal block at 0x00000163571408A0, 16 bytes long.
 Data: <   Xc           > F8 B5 BA 58 63 01 00 00 00 00 00 00 00 00 00 00 
Object dump complete.
The program '[15852] minimal.exe' has exited with code 0 (0x0).
```

MaartenBent:
It might be the static `tls_tdAutomation` object that is reported as being leaked.

src/msw/darkmode.cpp
```diff
@@ -164,6 +166,948 @@ bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
...
+
+    COLORREF srcPri = RGB(255, 255, 255), srcSec = RGB(240, 240, 240);
+    COLORREF srcSep = RGB(128, 128, 128), srcSp2 = RGB(223, 223, 223);
+
+    wxUxThemeHandle hL (wxUxThemeHandle ::NewAtStdDPI(L"TaskDialog"));
+    if (hL)
+    {
+        GetThemeColor(hL, TDLG_PRIMARYPANEL, 0, TMT_FILLCOLOR, &srcPri);
+        GetThemeColor(hL, TDLG_SECONDARYPANEL, 0, TMT_FILLCOLOR, &srcSec);
+        GetThemeColor(hL, TDLG_FOOTNOTESEPARATOR, 0, TMT_FILLCOLOR, &srcSep);
```
vadz, line 555: This seems useless: under Windows 10 I'm getting `0x80070490` (a.k.a. `HRESULT_FROM_WIN32(ERROR_NOT_FOUND)`) for these 2 calls, i.e. the "TaskDialog" theme doesn't define this, and under Windows 11 this code is not executed at all because it has native dark theme support.

vadz:
I didn't realize ASAN in MSVC doesn't show memory leaks :-( After disabling it, I do see them, sorry for the stupid question.

src/msw/darkmode.cpp
```diff
@@ -164,6 +166,948 @@ bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
...
+        if (brPrimary) { DeleteObject(brPrimary);   brPrimary = nullptr; }
+        if (brSecondary) { DeleteObject(brSecondary); brSecondary = nullptr; }
+        if (brFootnote) { DeleteObject(brFootnote);  brFootnote = nullptr; }
+    }
+    void Destroy() { CloseThemes(); DestroyBrushes(); elements.clear(); elemsOk = false; }
+};
+
+// Thread-local state map and UIA singleton
+static thread_local std::unordered_map<HWND, TDPageState> tls_tdStates;
+static thread_local wxCOMPtr<IUIAutomation>                 tls_tdAutomation;
```
vadz, line 266: Why does this have to be thread-local? It seems to be used from the main (GUI) thread only.

vadz:
> It might be the static `tls_tdAutomation` object that is reported as being leaked.

It's `tls_tdStates` actually. Will fix.

src/msw/darkmode.cpp
vadz: After checking and rechecking, I couldn't find any accesses to these globals from other threads, so I'm changing this. Please let me know if I'm missing something.

src/msw/darkmode.cpp
```diff
@@ -164,6 +166,948 @@ bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
...
+    const wchar_t* mainClass = s.isDark
+        ? L"DarkMode_Explorer::TaskDialog"
+        : L"TaskDialog";
+    const wchar_t* btnClass = s.isDark
+        ? L"DarkMode_Explorer::Button" : L"Button";
+    // For TaskDialog style, likely same as mainClass; could be different.
+    const wchar_t* mainClassDark = s.isDark ? L"DarkMode_Explorer::TaskDialog" : nullptr; // or nullptr if not needed
+
+    resetTheme(s.hTD, mainClass, mainClassDark);
+    resetTheme(s.hTDS, mainClass, mainClassDark);
```
vadz, line 363: Another thing I can't understand: `hTDS` seems to be always exactly the same as `hTD`, what's the point of having it?

I'm going to just drop it...

vadz:
I'm closing this one in favour of #26570 which is closer to be ready to be merged, but it still has leaks, as explained there. Any help with finding and fixing them would be welcome, but so would be testing more generally because I could have broken something with my changes, even though they were not supposed to have any material effect.

Thanks again for contributing all this code!

PR #26313, opened 2026-03-24
Prevent debug log messages on windows

mcorino:
Postpone dynamic init with FromDIP() till after window has been created.

### DIFF
--- a/src/propgrid/propgrid.cpp
+++ b/src/propgrid/propgrid.cpp
@@ -410,14 +410,14 @@ void wxPropertyGrid::Init1()
     m_doubleBuffer = nullptr;
 
 #ifndef wxPG_ICON_WIDTH
-    m_iconWidth = FromDIP(11);
-    m_iconHeight = FromDIP(11);
+    m_iconWidth = 11;
+    m_iconHeight = 11;
 #else
-    m_iconWidth = FromDIP(wxPG_ICON_WIDTH);
-    m_iconHeight = FromDIP(wxPG_ICON_WIDTH);
+    m_iconWidth = wxPG_ICON_WIDTH;
+    m_iconHeight = wxPG_ICON_WIDTH;
 #endif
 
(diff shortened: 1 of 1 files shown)


mcorino:
@vadz @MaartenBent 
Ever since working with the 3.3.2 release I've been bothered by some annoying debug log messages anytime a `wxPropertyGrid` is created on Windows.
I finally tracked this down to some changes in commits 942fcf8 and 098c7f0 which introduced calls to `FromDIP` before the window is created.

This change here partially reverts those commits (only the code in `Init1()`) and postpones the dynamic initialisation bit using `FromDIP ` to the second init stage in `Init2`.
Afaict this has no negative effects (but please check) and it gets rid of the annoying debug logs.

vadz:
Thanks, we should probably apply this to fix the problem, but I'm not very happy with doing initialization in 2 places, I think we should just remove it from `Init1()`.

In fact, I think the initialization here should be redone in a quite different way:

1. Initialize members with literal values (e.g. `0`, `nullptr` etc) in their declarations and get rid of `Init1()` completely.
2. Initialize members which need some window (including `m_iconWidth` etc) in `Create()` using the parent window DPI before creating the window.
3. Initialize stuff that needs the window to be already created after creating it in `Create()`.

This would, obviously, result in much more extensive changes but would make much more sense and be easier to maintain in the future IMO.

mcorino:
I agree wholehartedly. I'm just not familiar enough with all ins and outs of wxPropertyGrid to be confident to make such extensive changes.
I just tracked this issue down because it annoyed the hell out of me ;-) and left the initialisation in Init1 in place because "that's how it was" although I could not see a real use for it.

vadz:
I've convinced myself that we could remove the initialization in `Init1()`, please let me know if I broke anything by doing this.

mcorino:
Nice work. Debug logs are gone and everything still seems to work just fine afaict. Thanks.

PR #26316, opened 2026-03-25
Try running wxMSW/Univ builds using Debian Sid

vadz:
Somehow wine64 package doesn't seem to exist in Debian Testing any more.

### DIFF
--- a/.github/workflows/ci_msw_cross.yml
+++ b/.github/workflows/ci_msw_cross.yml
@@ -82,7 +82,7 @@ jobs:
             debian_release: stable
             triplet: i686-w64-mingw32
           - name: wxMSW/Univ
-            debian_release: testing
+            debian_release: unstable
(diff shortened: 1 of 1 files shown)


swt2c:
> Somehow wine64 package doesn't seem to exist in Debian Testing any more.

Yeah, it got kicked out (temporarily, this happens from time to time in Debian during development when RC bugs get reported):
https://tracker.debian.org/news/1734243/wine-removed-from-testing/

vadz:
I wonder if there is a way to be notified when it is added back? It's probably better to stay on testing, using Sid in the CI builds will probably result in more problems, on average...

swt2c:
> I wonder if there is a way to be notified when it is added back? It's probably better to stay on testing, using Sid in the CI builds will probably result in more problems, on average...

If you go to the package tracker page:
https://tracker.debian.org/pkg/wine

There is a subscribe button in the upper right where I believe you should get an email when it migrates back to testing.

PR #26319, opened 2026-03-26
wxScrolled<>:  reduce public API

wsu-cb:
Reduce the number of functions declared public by giving internal classes friend access

### DIFF
--- a/include/wx/scrolwin.h
+++ b/include/wx/scrolwin.h
@@ -67,10 +67,11 @@ enum wxScrollbarVisibility
 // wxVarScrollHelperBase in wx/vscroll.h without duplicating its code.
 class WXDLLIMPEXP_CORE wxAnyScrollHelperBase
 {
-public:
+protected:
     explicit wxAnyScrollHelperBase(wxWindow* win);
     virtual ~wxAnyScrollHelperBase() = default;
 
+public:
     // Disable use of keyboard keys for scrolling. By default cursor movement
     // keys (including Home, End, Page Up and Down) are used to scroll the
     // window appropriately. If the derived class uses these keys for something
@@ -108,6 +109,7 @@ public:
     wxWindow *GetTargetWindow() const { return m_targetWindow; }
 
 
+private:
     // The methods called from the window event handlers.
     void HandleOnChar(wxKeyEvent& event);
     void HandleOnPaint(wxPaintEvent& event);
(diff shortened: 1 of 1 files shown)


vadz:
Should this be merged or do you still plan more changes here?

wsu-cb:
> Should this be merged or do you still plan more changes here?

I am waiting for a review before declaring it ready.

vadz:
Ah, this is a catch-22 because I typically wait for a PR to exit the draft status before reviewing it (which is how it's supposed to work, I think, as implied by the "Ready for review" button label). Anyhow, I'll look at it in a moment, thanks.

include/wx/scrolwin.h
```diff
@@ -161,13 +167,16 @@ class WXDLLIMPEXP_CORE wxScrollHelperBase : public wxAnyScrollHelperBase
     // Forbid autoscrolling when the mouse is outside the window
     void DisableAutoScrollOutside();
 
+private:
```
vadz: Minor, but I'd rather move this part to the end of the class declaration instead of having public/private/public scopes.

include/wx/scrolwin.h
```diff
@@ -196,9 +204,11 @@ class WXDLLIMPEXP_CORE wxScrollHelperBase : public wxAnyScrollHelperBase
         DoShowScrollbars(horz, vert);
     }
 
+private:
```
vadz: I think `IsScrollbarShown()` should remain public, at least I don't see why would we want to hide it, it's a possibly useful function and it could be already used by the applications code.

include/wx/scrolwin.h
```diff
@@ -254,15 +264,19 @@ class WXDLLIMPEXP_CORE wxScrollHelperBase : public wxAnyScrollHelperBase
         return p2;
     }
 
+private:
```
vadz: Again, minor, but I'd rather move them to a private section in the end.

include/wx/scrolwin.h
```diff
@@ -430,7 +448,8 @@ public:                                                                       \
     virtual void DoSetVirtualSize(int x, int y) override                    \
         { ScrollDoSetVirtualSize(x, y); }                                     \
     virtual wxSize GetBestVirtualSize() const override                      \
-        { return ScrollGetBestVirtualSize(); }
+        { return ScrollGetBestVirtualSize(); }                              \
+private:
```
vadz: I'd rather not do this, hiding access change inside a macro is not very developer-friendly, it's better to put `private:` explicitly after calling it if necessary.

vadz (commented):
Thanks, this mostly looks good, but I think a couple of places may need to be changed.

wsu-cb:
@vadz This is ready for a re-review.

vadz:
Thanks, will push soon.

PR #26323, opened 2026-03-28
Fix minimal MSW static-build

tanzislam:
I needed these changes to be able to build WxWidgets with these commands in a MSYS2 MinGW64 window:
```sh
$ mkdir build-debug
$ cd build-debug
$ ../configure --enable-debug --disable-shared --disable-pic --enable-threads --enable-exceptions --without-subdirs --disable-tests --disable-all-features --disable-accessibility --enable-intl --enable-file --enable-log --enable-msgdlg --enable-wxdib --disable-unsafe_conv_in_wxstring --enable-regkey --enable-image --enable-uxtheme --enable-controls
$ make -j4
```

then build the following minimal program in a sibling directory of WxWidgets with this command: `g++ $(../wxWidgets/build-debug/wx-config --cxxflags) test.cpp $(../wxWidgets/build-debug/wx-config --libs) && ./a.exe`
```cpp
#include <wx/frame.h>
#include <wx/app.h>

{
  public:
    MyFrame(const wxString & title)
        : wxFrame(NULL, wxID_ANY, title)
    {}

  private:
    wxDECLARE_EVENT_TABLE();
};

wxBEGIN_EVENT_TABLE(MyFrame, wxFrame)
wxEND_EVENT_TABLE()

class MyApp : public wxApp
{
  public:
    bool OnInit() override
    {
        (new MyFrame("My first WxWidgets App"))->Show(true);
        return true;
    }
};

wxDECLARE_APP(MyApp);
wxIMPLEMENT_APP(MyApp);
```

### DIFF
--- a/src/msw/frame.cpp
+++ b/src/msw/frame.cpp
@@ -913,11 +913,13 @@ WXLRESULT wxFrame::MSWWindowProc(WXUINT message, WXWPARAM wParam, WXLPARAM lPara
     WXLRESULT rc = 0;
     bool processed = false;
 
+#if wxUSE_MENUBAR
     if ( GetMenuBar() &&
           HandleMenuMessage(&rc, this, message, wParam, lParam) )
     {
         return rc;
     }
+#endif // wxUSE_MENUBAR
 
     switch ( message )
     {
@@ -949,6 +951,7 @@ WXLRESULT wxFrame::MSWWindowProc(WXUINT message, WXWPARAM wParam, WXLPARAM lPara
             }
             break;
 
+#if wxUSE_MENUBAR
         case WM_INITMENUPOPUP:
         case WM_UNINITMENUPOPUP:
             // We get these messages from the menu bar even if the menu is
@@ -964,6 +967,7 @@ WXLRESULT wxFrame::MSWWindowProc(WXUINT message, WXWPARAM wParam, WXLPARAM lPara
                 }
             }
             break;
+#endif // wxUSE_MENUBAR
 
         case WM_QUERYDRAGICON:
             {
--- a/src/msw/msgdlg.cpp
+++ b/src/msw/msgdlg.cpp
@@ -121,10 +121,12 @@ wxMessageDialog::HookFunction(int code, WXWPARAM wParam, WXLPARAM lParam)
         // too big to fit the display
         wnd->ReplaceStaticWithEdit();
 
+#ifdef wxHAS_ANY_BUTTON
         // update the labels if necessary: we need to do it before centering
         // the dialog as this can change its size
         if ( wnd->HasCustomLabels() )
             wnd->AdjustButtonLabels();
+#endif // wxHAS_ANY_BUTTON
 
         // centre the message box on its parent if requested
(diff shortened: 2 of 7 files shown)


src/msw/frame.cpp
```diff
@@ -913,11 +913,15 @@ WXLRESULT wxFrame::MSWWindowProc(WXUINT message, WXWPARAM wParam, WXLPARAM lPara
     WXLRESULT rc = 0;
     bool processed = false;
 
+#if wxUSE_MENUS
```
vadz: We shouldn't need this one, `wxUSE_MENUBAR` requires `wxUSE_MENUS`, so if the former is enabled, the latter must be too.

src/msw/frame.cpp
```diff
@@ -949,6 +953,8 @@ WXLRESULT wxFrame::MSWWindowProc(WXUINT message, WXWPARAM wParam, WXLPARAM lPara
             }
             break;
 
+#if wxUSE_MENUS
```
vadz: Same comment as above.

src/msw/msgdlg.cpp
```diff
@@ -121,10 +121,12 @@ wxMessageDialog::HookFunction(int code, WXWPARAM wParam, WXLPARAM lParam)
         // too big to fit the display
         wnd->ReplaceStaticWithEdit();
 
+#ifdef wxUSE_ANY_BUTTON
```
vadz: This symbol is not defined anywhere, so this is wrong as it just always disables this code. Not sure what was the intention here, I don't see why the code inside this `#ifdef` wouldn't compile.

src/msw/msgdlg.cpp
```diff
@@ -262,6 +264,7 @@ void wxMessageDialog::ReplaceStaticWithEdit()
     }
 }
 
+#ifdef wxUSE_ANY_BUTTON
```
vadz: Same comment as above.

vadz (changes requested):
Thanks, mostly this looks good but please review the `wxUSE_ANY_BUTTON` check.

src/msw/frame.cpp
tanzislam: Thanks for explaining, fixed.

src/msw/frame.cpp
tanzislam: Fixed.

src/msw/msgdlg.cpp
tanzislam: My mistake, I meant to use `wxHAS_ANY_BUTTON` -- because that symbol is used to guard the contents of `src/msw/anybutton.cpp` where `wxMSWButton::GetFittingSize()` is defined. Fixed.

src/msw/msgdlg.cpp
tanzislam: Fixed.

vadz:
Thanks for the updates! I'll merge this soon, except for the changes to `src/common/framecmn.cpp`: looking at it, it seems clear that `gs_windowDisablers` shouldn't be defined inside `#if wxUSE_MENUS` check, it has nothing to do with the menus, so I'll just make sure it's always defined.

PR #26329, opened 2026-03-31
MSW: Refresh tabs after changing tabs programmatically in dark mode

RobertRoeb:
- This did not occur in dark mode, leaving the tabs a wrong state
  - Fixes #26281

### DIFF
--- a/src/msw/notebook.cpp
+++ b/src/msw/notebook.cpp
@@ -354,8 +354,8 @@ int wxNotebook::SetSelection(size_t nPage)
 
             (void)TabCtrl_SetCurSel(GetHwnd(), nPage);
 
-            /// MSW does not redraw the tab in darkmode
-            if (wxSystemSettings::GetAppearance().IsDark())
+            /// MSW does not redraw the tab in dark mode
+            if ( wxMSWDarkMode::IsActive() )
                 Refresh();
 
             SendPageChangedEvent(selectionOld, nPage);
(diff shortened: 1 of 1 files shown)


src/msw/notebook.cpp
```diff
@@ -354,6 +354,10 @@ int wxNotebook::SetSelection(size_t nPage)
 
             (void)TabCtrl_SetCurSel(GetHwnd(), nPage);
 
+            /// MSW does not redraw the tab in darkmode
+            if (wxSystemSettings::GetAppearance().IsDark())
```
vadz: This almost certainly should be `wxMSWDarkMode::IsActive()`, as we only want to do this when our dark mode support is used, not just because the system is using dark mode.

Could you please check that this works?

RobertRoeb: Yes, indeed. I believe the other ports work differently - wxSystemSettings::GetAppearance().IsDark() returns true if the app is dark, no matter what the system is so assumed this to be the same on MSW. Will check again if I have time.

RobertRoeb:
Changed to using wxMSWDarkMode::IsActive()

vadz:
Thanks for checking this, I'm going to merge this soon but in a slightly modified form because while applying it I've noticed that we were already calling `Refresh()` after calling `UpdateSelection()` in `wxNotebook::MSWOnNotify()` and this PR adds 2 more such calls, so it seems logical to just call `Refresh()` from `UpdateSelection()` itself.

PR #26332, opened 2026-03-31
Fix wxMSW build errors when wxUSE_REGKEY=0

stahta01:
Tested with

```
# Testing done under MSys2 UCRT64 enviroment
cd "C:\devel\repos\wxwidgets_libs\wxWidgets"
rm -fR cmake_build_msw_shared_unicode
mkdir -p cmake_build_msw_shared_unicode
        cmake \
          -DwxBUILD_TOOLKIT="msw" \
          -DBUILD_SHARED_LIBS="ON" \
          -DwxBUILD_SHARED="ON" \
          -DwxUSE_REGKEY="OFF" -DwxUSE_MIMETYPE="OFF" -DwxUSE_WEBVIEW="OFF" \
          -S . \
          -B cmake_build_msw_shared_unicode
cmake --build cmake_build_msw_shared_unicode
```


stahta01:
I did not know what to do with the MIMETYPE and WEBVIEW build errors.

vadz:
Looks good to me, thanks.

> I did not know what to do with the MIMETYPE and WEBVIEW build errors.

MIME types can't be meaningfully supported without registry access, so I think `wxUSE_MIMETYPE` should be set to 0 in this case. I am not sure that is the "WEBVIEW build error"?

stahta01:
Part of the error was this; I am building again to get full error. Edit: My PC does not have IE so that might be the cause.
`wxWidgets/src/msw/webview_ie.cpp:66:13: error: 'wxRegKey' was not declared in this scope`

stahta01:
Building with [-DwxUSE_REGKEY="OFF" -DwxUSE_MIMETYPE="OFF" -DwxUSE_WEBVIEW_IE="OFF"] built okay

The below built okay
`wxUSE_WEBVIEW:  ON       (enable wxWebview with Edge)`

lanurmi:
Out of curiosity, is there some real-world advantage in building with `wxUSE_REGKEY` disabled?

vadz:
I thought you would want to add the logic for disabling the components that don't compile without it to `wx/chkconf.h`, do you plan to do it? If not, I can merge this as is.

stahta01:
I plan to do a single PR that does that in the future once I find all the things that need done. I have a few other related things I am testing to make wxWidgets to build and run under MSys2 /usr/bin Cygwin like environment.

vadz:
FWIW I am pretty sure that wx supports MSYS2 just fine if you simply use the default build options.

I also don't understand why do you want to disable `wxRegKey` and while I don't want to prevent you from fixing build problems due to it, this is definitely a very low priority.

stahta01:
> FWIW I am pretty sure that wx supports MSYS2 just fine if you simply use the default build options.
> 
> I also don't understand why do you want to disable `wxRegKey` and while I don't want to prevent you from fixing build problems due to it, this is definitely a very low priority.

The MSys2 has either MINGW environments or the CYGWIN like environment that I am trying to build for right now.
And I am finding some corner cases that I verify exists under the UCRT64 MINGW enviroment.

stahta01:
I have decided to start the work in this PR to do the checks; might be a while before I figure out how to do it. It been over 5 years since I looked at those files.

stahta01:
I found #14514 issue and think I can add a fix to that to this PR. Do it in this PR or in a separate PR after this one?

I added the needed changes to this PR; if needed I can remove them.

vadz:
> I added the needed changes to this PR; if needed I can remove them.

This is fine, please feel free to add any other fixes for building with `wxUSE_XXX=0` here too. Just don't forget to remove the draft status when this is ready.

stahta01:
Everything but a single task has finished marking as ready because that appears to be taking a long time to start running.

include/wx/config.h
```diff
@@ -20,7 +20,7 @@
 
 // under Windows we prefer to use the native implementation but can be forced
 // to use the file-based one
-#if defined(__WINDOWS__) && wxUSE_CONFIG_NATIVE
+#if defined(__WINDOWS__) && wxUSE_CONFIG_NATIVE && wxUSE_REGKEY
```
vadz: Is this still needed after the changes in `wx/msw/chkconf.h`? AFAICS it shouldn't be.

samples/regtest/regtest.cpp
```diff
@@ -25,7 +25,7 @@
 #include "wx/imaglist.h"
 #include "wx/tokenzr.h"
 
-#if wxUSE_CONFIG_NATIVE && defined( __WINDOWS__ )
+#if wxUSE_CONFIG_NATIVE && defined( __WINDOWS__ ) && wxUSE_REGKEY
```
vadz: Same question as for the change `wx/config.h`.

samples/typetest/typetest.h
```diff
@@ -14,10 +14,22 @@
...
 {
 public:
-    MyApp() { m_textCtrl = nullptr; m_mimeDatabase = nullptr; }
+    MyApp()
+    {
+        m_textCtrl = nullptr;
+#if wxUSE_MIMETYPE
+        m_mimeDatabase = nullptr;
+#endif // wxUSE_MIMETYPE
+    }
```
vadz: Ideal would be to replace this with

```cpp
MyApp() = default;
```

and initialize the variables in their declarations, this would reduce the amount of `#if`s.

vadz (approved):
Mostly looks good, thanks, but I wonder if a couple of the changes are really necessary now.

samples/typetest/typetest.h
stahta01: I am a good C programmer but a bad C++ programmer so what you posted is not something I really understand.

samples/regtest/regtest.cpp
stahta01: I will test and see if it works without it

include/wx/config.h
stahta01: I will test and see if it works without it

stahta01: It tested okay on my local PC so pushed the changes

samples/regtest/regtest.cpp
stahta01: It tested okay on my local PC so pushed the changes, this one failed on the cross compiler once maybe it will work this time. It has always worked on my local PC.

stahta01:
> Out of curiosity, is there some real-world advantage in building with `wxUSE_REGKEY` disabled?

I am building wxWidgets for the MSYS2 Cygwin like environment. At first, it shut off `wxUSE_REGKEY` and other things till I got most of them fixed. And I decided that `wxUSE_REGKEY` being off might be a good idea. I have gotten both wxGTK and wxMSW to build. But minimal only works with wxMSW because my gtk3 appears to be broken.

Tim S.

stahta01:
I have no idea what to do with "samples/typetest/typetest.h"

src/propgrid/props.cpp
```diff
@@ -2055,7 +2055,7 @@ bool wxFileProperty::DisplayEditorDialog(wxPropertyGrid* pg, wxVariant& value)
         m_dlgTitle.empty() ? _("Choose a file") : m_dlgTitle,
         m_initialPath.empty() ? path : m_initialPath,
         file,
-        m_wildcard.empty() ? wxGetTranslation(wxALL_FILES) : m_wildcard,
+        m_wildcard.empty() ? wxString(wxGetTranslation(wxALL_FILES)) : m_wildcard,
```
MaartenBent, line 2058: Maybe we should let `wxGetTranslation` always return a `wxString` instead of the input type / `TString`.

Since the return type changed from `const wxString&` to `wxString` in #26196, this comment above the functions seems obsolete and there is no need to return `TString` in this and the functions below it.

https://github.com/wxWidgets/wxWidgets/blob/c456ccb28f96f536774896b312aca1aa9ebb5b0a/include/wx/translation.h#L497-L504

samples/typetest/typetest.h
vadz:
```suggestion
    MyApp() = default;
```

samples/typetest/typetest.h
```diff
@@ -35,7 +47,9 @@ class MyApp: public wxApp
 
 private:
     wxTextCtrl* m_textCtrl;
```
vadz:
```suggestion
    wxTextCtrl* m_textCtrl = nullptr;
```

samples/typetest/typetest.h
```diff
@@ -35,7 +47,9 @@ class MyApp: public wxApp
...
+#if wxUSE_MIMETYPE
     wxMimeTypesManager *m_mimeDatabase;
```
vadz:
```suggestion
    wxMimeTypesManager* m_mimeDatabase = nullptr;
```

vadz (commented):
I've left suggestions that can be applied to do what I meant in the typetest sample.

src/propgrid/props.cpp
stahta01: I really do not think this is a question I can or should answer

vadz: @MaartenBent I agree that the comment doesn't make sense and should be removed, but I'm less sure about the return type change: presumably, the idea behind setting `wxUSE_INTL=0` is to cut down on code size/speed and it might be better to avoid potentially unnecessary conversions to `wxString` in this case?

FWIW here we should probably use `wxGetTranslation(wxString(wxALL_FILES))` which would be optimal in any case, but it's really minor.

vadz:
Applied now with minor changes, thanks.

PR #26325, opened 2026-03-29
CMake improvements

MaartenBent:
Some install improvements:
* Map the correct built-in static libraries for MinSizeRel and RelWithDebInfo configurations
* Add `base_only` target
* Use relative symlinks

And some more improvements/fixes/refactoring.

PR #26338, opened 2026-04-03
Update functions.cmake

juliocesarperezfranco:
Fixed a little error when trying to install **"Release"** build while the option **wxBUILD_INSTALL_PDB** is ON.

### DIFF
--- a/build/cmake/functions.cmake
+++ b/build/cmake/functions.cmake
@@ -553,7 +553,7 @@ macro(wx_add_library name)
         )
 
         if(wxBUILD_SHARED AND MSVC AND wxBUILD_INSTALL_PDB)
-            wx_install(FILES $<TARGET_PDB_FILE:${name}> DESTINATION "${runtime_dir}")
+            wx_install(FILES $<TARGET_PDB_FILE:${name}> DESTINATION "${runtime_dir}" CONFIGURATIONS Debug RelWithDebInfo)
         endif()
 
         wx_target_enable_precomp(${name} "${wxSOURCE_DIR}/include/wx/wxprec.h")


vadz:
Thanks!

@MaartenBent Does this look good to you?

MaartenBent:
PDB files in Release builds can be useful as well, they give file / line numbers instead of memory addresses.
 And they are also available when downloading wxWidgets binaries from the website (even though those are not built with CMake).
Why do you want to disable this?

And if you really need this, I then prefer to make `wxBUILD_INSTALL_PDB` and 3-way option, `OFF`, `ON`, `DEBUG_ONLY`.

vadz:
Oops, sorry, I thought the fix was for the problem of not being able to install PDB files because they were not generated in the release build. If they are generated (and I agree that they should be), then I don't understand what does this actually fix.

@juliocesarperezfranco Could you please explain what was the error exactly?

juliocesarperezfranco:
I'm sorry, I was supposed to push the changes only on the mirror repository i am currently working on, not the official one, I got confused.

I am testing a set of configuration combinations for building a group of projects on Windows using MSVC and analyze their behaviour.

Someone already advised me to use different build directories for each configuration with their own individual settings because this seems like a visual studio only thing.

This specific issue appeared to me when testing a configuration with the options **wxBUILD_INSTALL_PDB=ON** and **wxBUILD_STRIPPED_RELEASE:BOOL=ON** together.
.
When setting only the first one, even the release build generates pdb files so its fine, but when adding the second option, the installation step fails because the pdb files are not generated.

This is the configuration command-line:
```batch
cmake -S C:/home/Workspaces/build_tests/sources/wxWidgets/v3.3.2 -B C:/home/Workspaces/build_tests/binaries/wxWidgets/v3.3.2/O24/msvc/19.44.35225/x64/multi/shared -G "Visual Studio 17 2022" -A x64 -DwxBUILD_INSTALL_PDB:BOOL="1" -DwxBUILD_STRIPPED_RELEASE:BOOL="1" ...
```

This is the installation log
[InstallRelease.log](https://github.com/user-attachments/files/26477151/InstallRelease.log)

In the line 667:
```log
 CMake Error at libs/base/cmake_install.cmake:55 (file):
           file INSTALL cannot find
           "C:/home/Workspaces/build_tests/binaries/wxWidgets/v3.3.2/O24/msvc/19.44.35225/x64/multi/shared/lib/vc_x64_dll/wxbase332u_build_test_vc_x64_smed.pdb":
           No error.
           libs/cmake_install.cmake:37 (include)
           cmake_install.cmake:42 (include)
```

MaartenBent:
I haven't tested that specific combination, but if it indeed fails with `wxBUILD_STRIPPED_RELEASE` we could change it to something like:
```cmake
        if(wxBUILD_SHARED AND MSVC AND wxBUILD_INSTALL_PDB)
            if(wxBUILD_STRIPPED_RELEASE)
                wx_install(FILES $<TARGET_PDB_FILE:${name}> DESTINATION "${runtime_dir}" CONFIGURATIONS Debug RelWithDebInfo)
            else()
                wx_install(FILES $<TARGET_PDB_FILE:${name}> DESTINATION "${runtime_dir}")
            endif()
        endif()
```

I'll test it later and if it fixes things I can include the fix in #26325

PR #26341, opened 2026-04-03
fix eglDestroySurface argument order

florolf:
eglDestroySurface takes the display first and the surface second rather than the other way around. Since both EGLDisplay and EGLSurface are just typedef'd to void* this doesn't raise a compiler warning. This currently doesn't crash at runtime because Mesa validates the display pointer against a list of known good values and fails gracefully, but it still causes a resource leak.

I came across this while looking into #26340. This is not the cause, but it still seems worth fixing.

### DIFF
--- a/src/unix/glegl.cpp
+++ b/src/unix/glegl.cpp
@@ -611,7 +611,7 @@ void wxGLCanvasEGL::OnRealized()
     {
         if ( m_surface != EGL_NO_SURFACE )
         {
-            eglDestroySurface(m_surface, m_display);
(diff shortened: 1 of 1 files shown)


vadz:
Oops, thanks for finding and fixing this, merged locally and will push soon.

I'm a bit surprised that this is not related to #26340 because 23ccdb23c8 (Improve and document wxGLCanvas::CreateSurface(), 2023-03-21) which added this call speaks about this code path being taken after reparenting, but I didn't have time to look at the other bug in details yet.

Thanks again for the fix!

PR #26342, opened 2026-04-03
Fix `wxStyledTextCtrl::Get/SetUseAntiAliasing()`

### LINKED ISSUE #26339: `wxStyledTextCtrl::SetUseAntiAliasing()` is totally broken
`wxStyledTextCtrl::SetUseAntiAliasing()` is broken and does either nothing, or the opposite of what it is supposed do.

I ran into this while working on #26277. 

## Some History

I have done some code archeology to figure out how we got to this state. Here is what I found:

<details>

- e0e5663fe4c8771d7e5d0c212fcf3f6cd9b1a37f (*"Allow an extra flag to be given to Font::Create", 2004-02-11*) added `extraFontFlag` to Scintilla as a dummy boolean field that platform implementations can use how they want. This change was later merged into Scintilla upstream in https://github.com/wxWidgets/scintilla/commit/e635868056bec7da5a8b0357f8692665420bf71a and https://github.com/wxWidgets/scintilla/commit/f8c49b2cfcd9bff44095bb892993c24da9e33bb4 (*"Patch from Robin Dunn to pass an extra flag down to font rendering to allow a platform-specific tweak.", 2004-03-29*).

- d1558f3d532c2e1c15e385d17c6200069f72714b (*"Use extraFontFlag to specify if anti-aliased fonts should be used. Use GetPartialTextExtents.", 2004-02-11*) added an option to disable anti-aliasing. This was apparently motivated by performance problems with anti-aliasing on Mac. The `extraFontFlag` field was used to store the setting, with the following meaning: `true` = AA enabled, `false` = AA disabled. `GetUseAntiAliasing()` and `SetUseAntiAliasing()` methods were added to read and modify the field.

- 8462a84b2caab30265823ebbdbdfc7cff91486e6 (*"Removed wxFont::Set/GetNoAntiAliasing() implementations.", 2009-10-09*) removed `wxFont::SetNoAntiAliasing()` and with that, the functionality of `SetUseAntiAliasing` was lost. After this change, `extraFontFlag` does nothing in the WX Scintilla platform, although `GetUseAntiAliasing()` and `SetUseAntiAliasing()` still read and write the value.

- In Scintilla upstream, https://github.com/wxWidgets/scintilla/commit/edd30534041e9b6328c2fce99b55ec84fe6e5afe (*"Added setting for font quality to allow application to choose anti-aliased, non-anti-aliased or lcd-optimized text.", 2009-11-02*) implemented the ability to configure anti-aliasing. It made use of the existing `extraFontFlag` to store the setting, but in a way incompatible with the previous implementation in WX platform. The field was changed from bool to int, with the following values:
    ```c++
    #define SC_EFF_QUALITY_DEFAULT 0
    #define SC_EFF_QUALITY_NON_ANTIALIASED 1
    #define SC_EFF_QUALITY_ANTIALIASED 2
    #define SC_EFF_QUALITY_LCD_OPTIMIZED 3
    ```

    The `SCI_SETFONTQUALITY` and `SCI_GETFONTQUALITY` messages were added for setting and getting the value of the field.

- 9e96e16fe0bc7af32d329c965ad110a3e5be5ad8 (*"Apply patch (plus some additional changes) upgrading Scintilla to version 2.03.", 2010-03-30*) updated Scintilla, importing the upstream change. WX platform's use of `extraFontFlag`, which now conflicts with Scintilla upsream code, was not updated.

    The methods `SetFontQuality()` and `GetFontQuality()`, generated from Scintilla's interface definitions for the new window messages, are added by this commit.

- b936bfe85efaf0e179ec6ce071853934aecdf43a (*"Add Direct2D support to wxSTC", 2018-01-25*) imported some drawing code form Scintilla win32 platform into Scintilla WX platform. This new code actually makes use of `extraFontFlag` (which from the removal of `wxFont::SetNoAntiAliasing()` up to this point had no effect), interpreting it as one of the `EFF_QUALITY_*` constants.

</details>

## Where We Are Now

As of now, there are two pairs of methods for controlling anti-aliasing:
 - `SetUseAntiAliasing`/`GetUseAntiAliasing` - legacy wxWidgets-specific interface, only supports a simple yes/no toggle, broken
 - `SetFontQuality`/`GetFontQuality` - new interface based on upstream `SCI_SETFONTQUALITY`/`SCI_GETFONTQUALITY` messages

The legacy interface has been broken since 2009 (see history above) and currently does the following:
 - `SetUseAntiAliasing(false)` sets font quality to `EFF_QUALITY_DEFAULT`
 - `SetUseAntiAliasing(true)` sets font quality to `EFF_QUALITY_NON_ANTIALIASED`
 - `GetUseAntiAliasing()` returns true is font quality is set to anything other than `EFF_QUALITY_DEFAULT`

On most platforms, neither `SetUseAntiAliasing` nor `SetFontQuality` has any effect, because the implementation in the Scintilla WX platform is missing, but it does have effect on wxMSW when DirectWrite is enabled.

## Proposed Solutions

I believe we should deprecate `SetUseAntiAliasing`/`GetUseAntiAliasing` - it is a wxWidgets specific interface, has been broken for more than 16 years, and `SetFontQuality`/`GetFontQuality` is a straightforward and less broken replacement.

In addition, the implementation has to be changed. However, it is unclear how to map the `true`/`false` value onto the `EFF_QUALITY_*` constants. We could define some mapping, but there is also another alternative: we could just say that the methods have been broken for so long that there is no point in fixing them now, and make them do nothing (and document that).

After we agree on what the solution should be, I'll do a PR.

filip-hejsek:
Fixes #26339.

To be resolved before merging:
- [x] Improve documentation formatting
- [x] Decide on whether `SetUseAntiAliasing(true)` enables grayscale or subpixel antialiasing, or uses OS defaults (which might be no antialiasing)

### DIFF
--- a/interface/wx/stc/stc.h
+++ b/interface/wx/stc/stc.h
@@ -8026,15 +8026,22 @@ public:
     bool DoDropText(long x, long y, const wxString& data);
 
     /**
-       Specify whether anti-aliased fonts should be used.
+       Specify whether anti-aliasing should be used when rendering text.
 
-      This will have no effect on some platforms, but on some (wxMac for
-      example) can greatly improve performance.
+       If @p useAA is true, sets @ref SetFontQuality() "font quality" to @c wxSTC_EFF_QUALITY_DEFAULT.
+
+       If @p useAA is false, sets @ref SetFontQuality() "font quality" to @c wxSTC_EFF_QUALITY_NON_ANTIALIASED.
+
+       Prefer using @ref SetFontQuality() directly in new code.
     */
     void SetUseAntiAliasing(bool useAA);
 
     /**
-       Returns the current UseAntiAliasing setting.
+       Returns whether anti-aliasing is enabled by @ref SetFontQuality() "font quality" setting.
+
+       This function returns true if GetFontQuality() is something other than @c wxSTC_EFF_QUALITY_NON_ANTIALIASED.
+
+       Prefer using @ref GetFontQuality() directly in new code.
     */
     bool GetUseAntiAliasing();
 
--- a/src/stc/ScintillaWX.cpp
+++ b/src/stc/ScintillaWX.cpp
@@ -264,7 +264,6 @@ void ScintillaWX::Initialise() {
     dropTarget->SetScintilla(this);
     stc->SetDropTarget(dropTarget);
 #endif // wxUSE_DRAG_AND_DROP
-    vs.extraFontFlag = true;   // UseAntiAliasing
 
     // Set up default OS X key mappings. Remember that SCI_CTRL stands for
     // "Cmd" key here, as elsewhere in wx API, while SCI_ALT is the "Option"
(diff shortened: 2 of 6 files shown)


vadz:
Just to confirm: this looks good to me and I'll merge it soon. Thanks again!

PR #26344, opened 2026-04-04
Add way to get generated SVG content from wxSVGFileDC

Blake-Madden:
Currently, if you pass empty filename string to either of `wxSVGFileDC`'s CTORs then nothing happens. Now, the content will be stored in a buffer that you can get via the new `GetSVGDocument()` function (regardless of whether you save to a file or not).

I was wanting to do some post hoc editing on the content and save it myself. This is more elegant than writing to a temp file and doing surgery on that.

Also, fixes an issue with `IsOK()` being wrong if bitmaps were processed successfully.


vadz (approved):
Thanks, looks good and I'll merge it.

On a somewhat related note, I wonder if we shouldn't also add a `Save()` member function which would allow to check for the error when writing the output file because doing it in dtor doesn't allow reporting any errors.

Issue #26266, opened 2026-03-05
wxDataViewTreeCtrl mirrored in RTL mode on Windows


stevecor:
#### Bug description:

The wxDataViewTreeCtrl control is rendered backwards/mirrored when in right-to-left layout mode on Windows. The problem appeared in 3.2.9.

The incorrect appearance:
<img width="740" height="242" alt="Image" src="https://github.com/user-attachments/assets/826a4f15-4200-4bed-9e03-7f4ad851dea3" />

Expected appearance using 3.2.8:
<img width="739" height="236" alt="Image" src="https://github.com/user-attachments/assets/cd985801-99e7-4826-968b-2e89b086ad4a" />

#### To Reproduce:

1. On Windows 11, unpack wxWidgets-3.2.10.tar.bz2
2. Run Visual Studio 2026
3. Open build/wx_vc18.slnx and build Debug x64
4. Open samples/dataview/dataview_vc9.vcproj and build Debug x64
5. Run the program
6. Click on the tab labelled wxDataViewTreeCtrl
7. Use the menu command File Toggle layout direction
8. The control has the incorrect appearance as shown above in this bug report

### Platform and version information
Windows 11 25H2 x64
wxWidgets 3.2.10 wxMSW

vadz:
This is really strange, I don't see anything obviously responsible for this in the list of commits between 3.2.8 and 3.2.9. The only commits touching wxDVC are 9de810e218 (Improve padding when using HiDPI in generic wxDataViewCtrl, 2025-10-15), 47041891cf (Avoid collapsing wxDVC node if already done by event handler, 2025-07-14) and eec195c0ec (Implement GetMainWindowOfCompositeControl for wxDataViewCtrl, 2025-05-07) but they don't seem to have anything to do with this.

If you can run git-bisect to find the exact commit which broke this, it would be most helpful.

AliKet:
I am curious to know which commit has broken RTL layout for wxDVC in wx 3.2.9; however, the following patch should fix this issue in 3.2.9 as well as in master:

```C++
diff --git a/src/msw/dc.cpp b/src/msw/dc.cpp
index 7ae8c2a21e..b24b5a3c2c 100644
--- a/src/msw/dc.cpp
+++ b/src/msw/dc.cpp
@@ -2380,6 +2380,10 @@ bool wxMSWDCImpl::DoStretchBlit(wxCoord xdest, wxCoord ydest,
[...]

PR #26346, opened 2026-04-04
Ensure that `wxDC::DeviceToLogical{X,Y}()` and `DeviceToLogical()` etc are always consistent

vadz:
This wasn't the case before, resulting in confusing bugs such as #26266.

Also clean up the code by removing unnecessary macros/functions.


AliKet:
Great! thanks for your work on this.

PR #26347, opened 2026-04-04
Add Save function to wxSVGFileDC

Blake-Madden:
Also, add validity checks for pens and brushes, update docs.

I have it so that once you call `Save`, the DTOR will skip calling it. I'm torn on that design choice, so let me know if you disagree.

### DIFF
--- a/src/common/dcsvg.cpp
+++ b/src/common/dcsvg.cpp
@@ -602,7 +602,7 @@ bool wxSVGFileDCImpl::Save()
 
     wxFile file(m_filename, wxFile::write);
     if ( file.IsOpened() )
-        m_saved = file.Write(GetSVGDocument(), wxConvUTF8);
+        m_saved = file.Write(GetSVGDocument(), wxConvUTF8) && file.Close();
 
     if ( !m_saved )
         m_writeError = true;
--- a/src/common/dcsvg.cpp
+++ b/src/common/dcsvg.cpp
@@ -137,9 +137,6 @@ wxString GetBrushFill(const wxColour& c, int style = wxBRUSHSTYLE_SOLID)
 
 wxString GetPenPattern(const wxPen& pen)
 {
-    if ( !pen.IsOk() )
-        return wxString();
-
     wxString s;
 
     // The length of the dashes and gaps have a constant factor.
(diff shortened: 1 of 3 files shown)


src/common/dcsvg.cpp
```diff
@@ -580,14 +595,24 @@ wxString wxSVGFileDCImpl::GetSVGDocument() const
...
 }
 
+bool wxSVGFileDCImpl::Save()
+{
+    if ( m_saved || m_filename.empty() )
+        return m_saved;
+
+    wxFile file(m_filename, wxFile::write);
+    if ( file.IsOpened() )
+        m_saved = file.Write(GetSVGDocument(), wxConvUTF8);
```
vadz: We should also check `Close()` return code as typically this is what would fail if e.g. there is not enough space remaining on disk.

src/common/dcsvg.cpp
```diff
@@ -137,6 +137,9 @@ wxString GetBrushFill(const wxColour& c, int style = wxBRUSHSTYLE_SOLID)
 
 wxString GetPenPattern(const wxPen& pen)
 {
+    if ( !pen.IsOk() )
```
vadz: Is it normal for the pen to be invalid? I.e. should we silently ignore it or fail with an assert?

vadz (commented):
Thanks for adding this!

src/common/dcsvg.cpp
Blake-Madden: I've gotten into the habit of silently failing with invalid colors and brushes and pens in my code. A client may set a pen to wxNullPen to indicate not using it, so seems reasonable to me to treat it like it isn't there (i.e., transparent). Asserts seem a bit harsh to me IMO.

src/common/dcsvg.cpp
Blake-Madden: Got it.

src/common/dcsvg.cpp
vadz: But doesn't using an invalid pen/etc result in an assert right now, i.e. without the changes here? I'm almost sure they do, which means that this doesn't actually happen and I don't think it's a good idea to start allowing it, why would we want to do it?

Blake-Madden: Yes, they assert currently before these changes. You can ignore the validity commit if you prefer the asserts.

Blake-Madden: I rolled these back

PR #26350, opened 2026-04-05
Don't impose arbitrary maximum length on picker controls text

vadz:
This code was there ever since ec376c8fd9 (added
wx{Colour|File|Dir|Font}PickerCtrl (patch 1472329 by Francesco), 2006-05-31), but didn't make any sense because it should be possible to enter arbitrarily long strings in these controls.

Additionally, the maximum length was increased for wxFilePickerCtrl after creation which meant that that its initial contents was not shown in it, but the same string could be shown later, which was even more confusing.

Just stop imposing any length constraints.

See #7863.

Closes #26314.


vadz:
Applied as 932004d7225c3b31f484d452325d74109856048b

PR #26353, opened 2026-04-05
wxOSX Cache Alignment Rect Insets

ryancog:
Partially addresses #26318

The first commit da08c5761342 is the interesting one. As noted, it's unclear if there's ever a time where the insets would be invalidated and would need to be refetched, but the Apple documentation is sparse on how the usual Auto Layout system would query it. It doesn't seem like something that would ever change, but worth noting.

The second and third commits are related, but I guess it depends on opinions if they should be included or not.

The second commit 25e46b903075 adds `MacGetBorderSize()` following the logic that the `MacGet*Border...` functions are almost never called on their own, and in fact `GetLayoutInset()` always returns all the values, so it's (superficially... it ultimately doesn't matter much for performance and is moreso semantic I guess) "wasteful" in the current setup w/o a way to grab all of the values, as would be more natural given the underlying API.

The third commit 7b2307e2c521 uses the new function where appropriate in wxCocoa code and, IMO, cleans things up a bit.

I've added `wxMacBorderSize` as it doesn't seem like there's an already-existing suitable structure, and 25e46b903075 also removed `m_macTopLeftInset` and `m_macBottomRightInset` from `wxWindowMac` as those are never used, and seem like they were intended to be used to store these inset values. But such storage makes more sense in the Impl class.

### DIFF
--- a/include/wx/osx/cocoa/private.h
+++ b/include/wx/osx/cocoa/private.h
@@ -268,6 +268,11 @@ protected:
     // events, don't resend them
     bool m_hasEditor;
 
+    mutable int m_insetLeft;
+    mutable int m_insetRight;
+    mutable int m_insetTop;
+    mutable int m_insetBottom;
+
     friend class wxWidgetCocoaNativeKeyDownSuspender;
 
     wxDECLARE_DYNAMIC_CLASS_NO_COPY(wxWidgetCocoaImpl);
--- a/include/wx/osx/cocoa/private.h
+++ b/include/wx/osx/cocoa/private.h
@@ -113,7 +113,10 @@ public :
     virtual void        GetPosition( int &x, int &y ) const override;
     virtual void        GetSize( int &width, int &height ) const override;
     virtual void        SetControlSize( wxWindowVariant variant ) override;
+
     virtual void        GetLayoutInset(int &left , int &top , int &right, int &bottom) const override;
+    virtual void        InvalidateLayoutInset() const override;
+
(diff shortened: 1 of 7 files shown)


vadz:
@csomor Not sure if you noticed this one but I just wanted to say that I don't know anything about this stuff, so it would be great if you could please review this. TIA!

csomor:
@vadz I did, coincidentally I'm right now reading through the changes and reflect on invalidation and iOS implications ... thanks for reminding me, better double than never

csomor (approved):
Thanks, @ryancog. This looks fine to me. One addition I'd like to ask you for is an explicit method to invalidate the cached border structure, so that we don't have to expose that m_insetLeft has to be set to -1 for this. I don't know yet all the moments when this might be needed, but eg changing the native control variant `wxWidgetCocoaImpl::SetControlSize` would be a place where I'd call it just to be sure.

ryancog:
@csomor I added `InvalidateLayoutInset()` to the wxOSX `wxWidgetImpl` base and added a proper implementation in `wxWidgetCocoaImpl`

Lmk if the change in `wxWidgetCocoaImpl::SetControlSize()` matches what you were thinking.

csomor:
> @csomor I added `InvalidateLayoutInset()` to the wxOSX `wxWidgetImpl` base and added a proper implementation in `wxWidgetCocoaImpl`
> 
> Lmk if the change in `wxWidgetCocoaImpl::SetControlSize()` matches what you were thinking.

@ryancog Thanks a lot, that looks fine to me, @vadz if you don't see other fixes needed, it's ready to be merged

PR #26354, opened 2026-04-05
Use appropriate locale for display and region settings in wxMSW

utelle:
On Windows the default locale instantiated by `wxUILocale::UseDefault()` referenced the locale configured for regional formatting in the Windows settings. This could lead to unintended behaviour (like using the wrong layout direction), if the locale for the Windows display language differs from the locale for regional formatting.

This has been fixed by referencing the locale corresponding to the display language resp the regional formatting depending on the locale property in question.

Additionally, the German translation of the `internat` sample was updated.


utelle:
AFAICT the failing checks have nothing to do with my changes.

vadz:
Yes, we really need to stop using Ubuntu 18.04 for the CI jobs, its repositories are unusably flaky... I've restarted the failed CI jobs, let's hope they fare better this time.

include/wx/uilocale.h
```diff
@@ -268,11 +271,12 @@ class WXDLLIMPEXP_BASE wxUILocale
 private:
     // This ctor is private and exists only for implementation reasons.
     // It takes ownership of the provided pointer.
-    explicit wxUILocale(wxUILocaleImpl* impl = nullptr) : m_impl(impl) { }
+    explicit wxUILocale(wxUILocaleImpl* impl = nullptr) : m_impl(impl), m_isDefault(false) { }
 
     static wxUILocale ms_current;
 
     wxUILocaleImpl* m_impl;
+    bool m_isDefault;
```
vadz: Minor, but I'd prefer to do it like this:

```suggestion
    bool m_isDefault = false;
```

and remove the initialization above.

src/common/uilocale.cpp
```diff
@@ -854,6 +855,7 @@ bool wxUILocale::UseLocaleName(const wxString& localeName)
 
     impl->Use();
     ms_current = wxUILocale(impl);
+    ms_current.m_isDefault = false;
```
vadz, line 858: This is unnecessary?

src/msw/uilocale.cpp
```diff
@@ -173,14 +171,23 @@ WXDLLIMPEXP_BASE wxString wxGetMSWDateTimeFormat(wxLocaleInfo index)
         return wxString{};
     }
 
+    // Determine the name of the locale to be used for querying the date/time format
+    // Use a nullptr in case of the system default locale
```
vadz: OK, this works because `LOCALE_NAME_USER_DEFAULT` is defined as `NULL`, but it would still be better/more clear to use `LOCALE_NAME_USER_DEFAULT` as `name` value.

src/msw/uilocale.cpp
```diff
@@ -774,14 +786,35 @@ class wxUILocaleImplName : public wxUILocaleImpl
...
+    wxString DoGetInfoDisplay(LCTYPE lctype) const
+    {
+        wchar_t buf[256];
+        if ( !::GetLocaleInfoEx(m_nameDisplay, lctype, buf, WXSIZEOF(buf)) )
+        {
+            wxLogLastError(wxT("GetLocaleInfoEx"));
+            return wxString();
+        }
+
+        return buf;
```
vadz: I'm not sure if it's really better to duplicate `DoGetInfo()` (which is pretty short, of course, but still) than to add an extra parameter for it. I.e. maybe it would be better to use `DoGetInfo(m_nameDisplay, lctype)` instead of `DoGetInfoDisplay(lctype)` etc?

src/msw/uilocale.cpp
```diff
@@ -820,7 +853,8 @@ class wxUILocaleImplName : public wxUILocaleImpl
         return wxLocaleNumberFormatting(groupSeparator, grouping, decimalSeparator, fractionalDigits);
     }
 
-    const wchar_t* const m_name;
+    const wchar_t* const m_nameRegion;  // Name of locale for regional formatting
+    const wchar_t*       m_nameDisplay; // Name of locale for display language
```
vadz, line 850: I think we need to explain in the comment here that it is usually equal to `m_nameRegion` and only different from it when using default locale.

vadz (approved):
Thanks! I didn't test this yet but it looks like it should work fine, let's merge it (maybe after minor tweaks if you'd like to address the comments below) to see what we're missing...

include/wx/uilocale.h
utelle: Done.

src/common/uilocale.cpp
utelle: Although `wxUILocale::UseDefault()` and `wxUILocale:UseLocaleName()` are meant to be called only once, it is not forbidden or prevented. And one older application of mine - still using `wxLocale` - allows to change the UI language (resp locale) on the fly. In such a case it is necessary to reset `m_isDefault`. So, yes, I think it is necessary.

src/msw/uilocale.cpp
utelle: Done.

src/msw/uilocale.cpp
utelle: I applied the proposed change.

src/msw/uilocale.cpp
utelle: Done.

vadz:
Great, thanks, I'll merge and push this soon.

PR #26355, opened 2026-04-05
Escape titles in wxSVGFileDC

Blake-Madden:
Titles with `&` and friends currently cause broken SVG; this fixes it.


vadz:
Thanks, will merge soon.

PR #26362, opened 2026-04-07
Various small fixes to wxSVGFileDC

Blake-Madden:


src/common/dcsvg.cpp
```diff
@@ -204,7 +204,7 @@ wxString GetPenStyle(const wxPen& pen)
 {
     wxString penStyle;
 
-    penStyle += wxString::Format(wxS("stroke-width=\"%d\""), pen.GetWidth());
+    penStyle += wxString::Format(wxS("stroke-width=\"%d\""), wxMax(0, pen.GetWidth()));
```
vadz: Sorry, I don't think this is the right place to do this. Negative pen widths are not allowed, we need to detect this in `wxPen::SetWidth()` and assert there, not allow silently using them as invisible pens.

vadz (commented):
Thanks, the other commits look good but I'd rather not apply the first one.

src/common/dcsvg.cpp
Blake-Madden: OK, just rolled that one back.

PR #26364, opened 2026-04-07
Reduce number of checks in CMake and more cleanup

vadz:
This started as an attempt to reduce the number of checks uselessly done in CMake but, as usual, transformed into something more with cleanup of some adjacent areas. Note that in addition to removing and skipping some checks, this PR also _adds_ some checks that were (wrongly) not done by CMake before.

I'm not sure if my `_if_not_linux` hack is the best way to do what I want, but I'm pretty sure I want it, it's a bit ridiculous to check for the functions available since 20+ years on every run. Of course, this is the case for configure too but it makes a bit more sense to preserve support for legacy systems there.

@MaartenBent Please let me know if you have any objections/proposals for improvement, otherwise I'll merge this if/when the CI jobs pass.


MaartenBent:
I think the CMake part looks fine, but I have to admit I did not check every single change in `setup.cmake`.

PR #26366, opened 2026-04-09
Change wxSVGFileDC to create unique clip/gradient IDs across multiple outputs

Blake-Madden:
Instead of resetting IDs to zero with every `wxSVGFileDC` construction, have it remember the IDs that were already used.

This is useful if you need to combine multiple output SVGs. In my situation, I am ouputting many SVGs from different graph renderings, extracting the body from them, and then combining them into a larger SVG report. (Perhaps this is a use case for others as well.) With the current behaviour, the IDs for my clip rects are not always unique, which then causes rendering failures.  This PR fixes this.

### DIFF
--- a/include/wx/dcsvg.h
+++ b/include/wx/dcsvg.h
@@ -289,10 +289,10 @@ private:
 
     // Unique ID for every clipping graphics group: this is simply always
     // incremented in each SetClippingRegion() call.
-    size_t m_clipUniqueId;
+    static size_t m_clipUniqueId;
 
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, will push soon.

BTW, one small thing: could you please use the blank line between the subject and the body in your commit messages, as it should be done (see [the classic guidelines](https://cbea.ms/git-commit/))? Otherwise I always have to edit them before pushing. TIA!

PR #26367, opened 2026-04-09
OSX: Add value to EVT_DATAVIEW_ITEM_EDITING_DONE

RobertRoeb:
- The value only seems to be reported in OSXCellChanged so defer sending the event until then (unless it was cancelled).
  - This was recently reported as a bug, but I cannot find the report anymore

### DIFF
--- a/include/wx/osx/dvrenderer.h
+++ b/include/wx/osx/dvrenderer.h
@@ -81,7 +81,7 @@ public:
 #if wxOSX_USE_COCOA
     bool m_callEditingDoneOnCellChange = false;
     void OSXCallEditingDoneOnCellChange() { m_callEditingDoneOnCellChange = true; }
-    void OSXSendEditingDoneEvent( const wxDataViewItem &item, const wxVariant &value );
+    void OSXSendEditingDoneEventIfPending( const wxDataViewItem &item, const wxVariant &value );
 
     // called when a value was edited by user
     virtual void OSXOnCellChanged(NSObject *value,
--- a/include/wx/osx/dvrenderer.h
+++ b/include/wx/osx/dvrenderer.h
@@ -79,7 +79,7 @@ public:
     void OSXUpdateAlignment();
 
 #if wxOSX_USE_COCOA
-    bool m_callEditingDoneOnCellChange;
+    bool m_callEditingDoneOnCellChange = false;
     void OSXCallEditingDoneOnCellChange() { m_callEditingDoneOnCellChange = true; }
     void OSXSendEditingDoneEvent( const wxDataViewItem &item, const wxVariant &value );
 
--- a/include/wx/osx/dvrenderer.h
+++ b/include/wx/osx/dvrenderer.h
@@ -79,6 +79,10 @@ public:
     void OSXUpdateAlignment();
 
 #if wxOSX_USE_COCOA
+    bool m_callEditingDoneOnCellChange;
+    void OSXCallEditingDoneOnCellChange() { m_callEditingDoneOnCellChange = true; }
+    void OSXSendEditingDoneEvent( const wxDataViewItem &item, const wxVariant &value );
+
     // called when a value was edited by user
     virtual void OSXOnCellChanged(NSObject *value,
                                   const wxDataViewItem& item,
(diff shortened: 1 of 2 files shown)


src/osx/cocoa/dataview.mm
```diff
@@ -1831,7 +1831,7 @@ -(void) outlineViewSelectionDidChange:(NSNotification*)notification
     dvc->GetEventHandler()->ProcessEvent(event);
 }
 
--(void) sendEditingDoneEvent:(BOOL)isCancelled
+-(void) sendEditingDoneEvent:(BOOL)isCancelled withNotification:(NSNotification*)notification
```
vadz: The newly added `notification` parameter doesn't seem to be used, am I missing something?

include/wx/osx/dvrenderer.h
```diff
@@ -79,6 +79,10 @@ class WXDLLIMPEXP_ADV wxDataViewRenderer : public wxDataViewRendererBase
     void OSXUpdateAlignment();
 
 #if wxOSX_USE_COCOA
+    bool m_callEditingDoneOnCellChange;
```
vadz: Minor but it's better to initialize members in their declarations in the new code:
```suggestion
    bool m_callEditingDoneOnCellChange = false;
```

src/osx/cocoa/dataview.mm
```diff
@@ -2704,7 +2710,8 @@ void UpdateWithRow(int row)
     : wxDataViewRendererBase(varianttype, mode, align),
       m_alignment(align),
       m_mode(mode),
-      m_NativeDataPtr(nullptr)
+      m_NativeDataPtr(nullptr),
+      m_callEditingDoneOnCellChange(false)
```
vadz:
```suggestion
      m_NativeDataPtr(nullptr)
```

src/osx/cocoa/dataview.mm
```diff
@@ -2790,13 +2797,29 @@ void UpdateWithRow(int row)
...
+{
+    if (m_callEditingDoneOnCellChange)
+    {
+        wxDataViewColumn *column = GetOwner();
+        wxDataViewCtrl *dvc = column->GetOwner();
+        wxDataViewEvent event(wxEVT_DATAVIEW_ITEM_EDITING_DONE, dvc, column, item);
+        event.SetValue( value );
+        dvc->GetEventHandler()->ProcessEvent(event);
+
+        m_callEditingDoneOnCellChange = false;
```
vadz: Call me paranoid, but I'd rather do it as the first statement in the `if` branch, before dispatching the event — just in case it could somehow result in reentering this function.

include/wx/osx/dvrenderer.h
```diff
@@ -79,6 +79,10 @@ class WXDLLIMPEXP_ADV wxDataViewRenderer : public wxDataViewRendererBase
...
+    void OSXCallEditingDoneOnCellChange() { m_callEditingDoneOnCellChange = true; }
+    void OSXSendEditingDoneEvent( const wxDataViewItem &item, const wxVariant &value );
```
vadz: I'd like to either rename this function or at least add a comment explaining that it's not unconditional, i.e.
```suggestion
    void OSXSendEditingDoneEventIfPending( const wxDataViewItem &item, const wxVariant &value );
```

vadz (commented):
Thanks for the fix, I think this bug was recently reported by @hwiesmann on the mailing list.

If there is no other way to fix it, I guess we should do it like this even if this likely means that the order of events may be different in wxOSX and the other ports.

I also wonder if we really need to do this for anything but string values, it looks like they're the only ones for which we want to have the new value in the handler. And if it were done like this, the changes could be made much smaller by calling `OSXSendEditingDoneEvent()` from `-outlineView:outlineView:setObjectValue:forTableColumn:byItem:` itself.

src/osx/cocoa/dataview.mm
RobertRoeb: Ah right, I tried to get the data from the notification, which is what a post on SO suggested, but that didn't work and I forgot to revert

PR #26370, opened 2026-04-10
Stop scaling icons loaded from resources in wxMSW

### LINKED ISSUE #26369: Using `wxImageList` in `wxListCtrl` broken in wxMSW in high DPI
I've found this while testing the fix for #26349, which was done by replacing the current listctrl sample files (`.cpp` and `.h` only) with the ones from v3.1.5, and while I don't care much about the deprecated `wxImageList` class, this seems to be serious: even for DPI-unaware programs (i.e. not using our manifest) its use is broken in 3.2 and later, the small icons are not displayed correctly at all:

Image

Debugging this shows that this happens because the icon loaded from resources is actually 32x32 and not 16x16 as expected. I think this happens because `LoadIcon()` only loads images of the standard size — but how could this work before if this is the reason?

In any case, it would be nice to make this work at least as well/bad as before, i.e. upscaling the icons, instead of breaking any code using `wxImageList` completely in high DPI.

vadz:
This is not backwards compatible but looks like the right thing to do because old behaviour was inconsistent: using LoadIcon() meant that we always loaded icons of the given (system default size) which meant that some icons were upscaled while the others were not, e.g. in a system using 200% scaling 16x16 icons were loaded as 32x32 ones, but so were 32x32 icons too.

Use LoadImage() to load the image in its actual size, where "actual" means the size of the first icon for icons with multiple bitmaps. This might not be the best thing to do and maybe we should load the default icon size if the icon provides it, which could be done in a way similar to what we do in MSW-specific wxIconBundle::AddIcon().

Not scaling the icons means that they always use scale of 1, which makes things behave correctly for DPI-unaware applications that were broken before, which seems better even if it breaks DPI-aware applications: but those should be using wxBitmapBundle, which still works correctly.

Closes #26369.

-----

I'd really like to get some testing/feedback for this one because it's clearly backwards incompatible. But this seems the right thing to do — or, at least, using `LoadIcon()` as we did before is very clearly a wrong thing to do, for the reasons explained above.

Please let me know what does this break and let's see if we can repair it without going to the old behaviour. TIA!

### DIFF
--- a/src/msw/gdiimage.cpp
+++ b/src/msw/gdiimage.cpp
@@ -606,49 +606,42 @@ bool wxICOResourceHandler::LoadIcon(wxIcon *icon,
     HICON hicon;
 
     // do we need the icon of the specific size or would any icon do?
-    bool hasSize = desiredWidth != -1 || desiredHeight != -1;
+    const bool hasSize = desiredWidth != -1 && desiredHeight != -1;
 
-    wxASSERT_MSG( !hasSize || (desiredWidth != -1 && desiredHeight != -1),
-                  wxT("width and height should be either both -1 or not") );
+    if ( !hasSize )
+    {
+        wxASSERT_MSG( desiredWidth == -1 && desiredHeight == -1,
+                      wxT("width and height should be either both -1 or not") );
+
+        // LoadImage() interprets 0 as meaning "use the actual size".
+        desiredWidth =
+        desiredHeight = 0;
+    }
 
     // try to load the icon from this program first to allow overriding the
     // standard icons (although why one would want to do it considering that
     // we already have wxApp::GetStdIcon() is unclear)
+    hicon = (HICON)::LoadImage(wxGetInstance(), name.t_str(), IMAGE_ICON,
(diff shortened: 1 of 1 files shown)


vadz:
Well, no comments since 2 weeks, so I'll merge this and see what it breaks.

vadz:

MaartenBent:
After the second fix ([b45b02c](https://github.com/wxWidgets/wxWidgets/commit/b45b02c22747ad5e4bd9fc5de9f6afa88121531b)), I get debug asserts in `wxIcon::InitFromHICON` with `wxDEBUG_LEVEL=2`. Because width and height are `0` instead of the determined icon size.

`wxICOResourceHandler::LoadIcon` does not determine the correct size anymore. I think `size` should always be determined, and `scale` only when `iconMayBeScaled`.

This happens in every sample, by `SetIcon(wxICON(sample));`.

vadz:
Oops, thanks for noticing this, fixed now in 42c12b2d8fae842b454a9af18f6c0539b3699b0f

PR #1208, opened 2019-01-31
wxSlider and widgets sample improvements

MaartenBent:
I tried to fix some DPI issues for #334, but it resulted in this PR.

- Document `wxSL_BOTH` style of `wxSlider`.
- Centre labels above and below vertical `wxSlider`. Improve label position for specific combinations of `wxSlider` styles.
- Improve `wxSlider` in high DPI. Make thumb and tick size DPI dependent.
- Better fix layout of `wxSlider` labels after changing value range. Commit 0b821ad did not (always) trigger a call to `DoMoveWindow`. See [#17093](https://trac.wxwidgets.org/ticket/17093).
- Fix `wxSlider::SetThumbLength()` on `wxMSW`. It requires `TBS_FIXEDLENGTH` style. Use `GetThumbLength()` instead of the arbitrary defined `THUMB` size.
- Improve guards in button pages of widgets sample.
- Use `wxSizerFlags` for slider page in widgets sample.
- Some fixes for `wxSlider` in widgets sample
  - Correctly save the slider value before recreating the slider.
  - Layout the entire page instead of only the slider sizer.
  - Update the text controls after recreating the slider.
  - Always enable 'both sides' checkbox. It has additional effects in `wxMSW`.
- Demonstrate `wxSlider::SetSelection()` in widgets sample.

PR #26372, opened 2026-04-12
Set wxSlider thumb size to match its initial size in wxMSW

vadz:
The changes of 8f3b87dce7 (Fix wxSlider::SetThumbLength() on wxMSW, 2019-01-31) broke the behaviour of wxSlider controls created with a fixed height as their thumb didn't fit it any longer and left no space for ticks at all.

Work around this by setting the initial thumb size ourselves if a fixed control size is given. Also use the fixed default thumb size when calculating the best size as the value of GetThumbLength() could vary depending on how the control was created.

It might be better to revert the use of TBS_FIXEDLENGTH and just leave with the default thumb size in any DPI.

See #1208.

----

@MaartenBent I had to change this because existing code using fixed sizes for its controls got completely wrecked by this change. This is the minimal fix I can see for it, but as I mentioned above, I'm not sure why we can't leave the control manage its thumb size itself, it looks fine to me in 200% DPI (under Windows 10), do you remember by chance which problem did the original change fix? FWIW I don't think we care about `SetThumbLength()` one way or the other, nobody uses it anyhow.

### DIFF
--- a/src/msw/slider.cpp
+++ b/src/msw/slider.cpp
@@ -50,8 +50,10 @@ enum
 // the gaps between the slider and the labels, in pixels
 const int HGAP = 5;
 const int VGAP = 4;
-// this value is arbitrary:
-const int TICK = 8;
+// these values are arbitrary:
+const int THUMB = 20;
+const int THUMB_BORDER = 2;
+const int TICK = 4;
 
 } // anonymous namespace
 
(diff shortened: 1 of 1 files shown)


MaartenBent:
I don't remember why I added it, I think `SetThumbLength` just didn't work.
I'll try to test these changes in some DPI configurations.

vadz:
`SetThumbLength()` doesn't work without `TBS_FIXEDLENGTH`, this is true. But I don't know why do we want/need to call it when the native control seems to adjust the thumb length to the control size on its own.

In any case, we need to do something here because this makes `wxSlider`s created with fixed size in old code unusable with the current wx version (while they worked fine before).

vadz:
@MaartenBent Can we merge this for now and improve it later if there any problems with this? I need to apply this patch all the time to test the application I'm working on as it's UI is just completely broken without it.

MaartenBent:
Seems to still work fine in the Widgets sample, so go ahead and merge 👍

vadz:

PR #26379, opened 2026-04-14
Add accessibility support to wxSVGFileDC

Blake-Madden:
Adds an ARIA group builder and wrapper for drawing commands within SVG output.

### DIFF
--- a/include/wx/dcsvg.h
+++ b/include/wx/dcsvg.h
@@ -17,6 +17,7 @@
 #include "wx/string.h"
 #include "wx/filename.h"
 #include "wx/dc.h"
+#include "wx/arrstr.h"
 
 #include <memory>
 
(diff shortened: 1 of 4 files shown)


include/wx/dcsvg.h
```diff
@@ -35,6 +36,36 @@ enum wxSVGShapeRenderingMode
...
+    wxSVGAttributes& Add(const wxString& name, const wxString& value);
+
+    // Returns the attributes as a string of name="value" pairs, each prefixed with a space.
+    wxString GetAsString() const;
+
+    bool IsEmpty() const { return m_names.empty(); }
+
+private:
+    wxArrayString m_names;
+    wxArrayString m_values;
```
vadz: I guess it doesn't really matter for the number of attributes we're going to have, but it looks like this ought have been `map<wxString, wxString>` instead.

vadz (approved):
Thanks, this looks really nice!

Please let me know if you'd like to replace the string arrays with a map or prefer to leave it like this.

include/wx/dcsvg.h
MaartenBent: According to the docs these should not be used anymore, in favour of `std::vector`. In this case `std::map` would be even better.

samples/svg/svgtest.cpp
```diff
@@ -282,15 +291,36 @@ void DrawOnDC(wxDC& dc, const int index)
...
+            wxSVGFileDC* const svgDC = wxDynamicCast(&dc, wxSVGFileDC);
+
             // draw standard shapes
             dc.SetBrush(*wxCYAN_BRUSH);
             dc.SetPen(*wxRED_PEN);
+            if ( svgDC )
+                svgDC->BeginAccessibleGroup(wxSVGAttributes().AriaLabel("Cyan rectangle with red border"));
             dc.DrawRectangle(dc.FromDIP(10), dc.FromDIP(10), dc.FromDIP(100), dc.FromDIP(70));
+            if ( svgDC )
+                svgDC->EndAccessibleGroup();
```
MaartenBent, line 312: Maybe you can use some sort of RAII (is this the name?) so the constructor opens the group an the destructor closes it.
```c++
{
    wxSVGAccessibilityGroup ag(svgDC, label, value);
}

include/wx/dcsvg.h
Blake-Madden: Done, thanks for the suggestion.

Blake-Madden: Done, thanks for the suggestion.

samples/svg/svgtest.cpp
Blake-Madden: Added `wxSVGAccessibleGroup` to the sample.

PR #26381, opened 2026-04-14
Don't show hidden TLWs in wxMSW when parent is restored

### LINKED ISSUE #23997: The hidden wxFrame is shown automatically and can not be hidden after restoring from minimize
### Description
See  [https://github.com/wxWidgets/Phoenix/issues/1606](url) for detailed problem description and reproduction.
I can confirm that this is a bug in the wxWidgets core, not in the wxPython binding, as I'm using C++.

- OS  and its version: Windows 10 22H2

vadz:
The default Windows behaviour is to show the children of a hidden window when this window is restored from the taskbar, but we don't want this to happen as this is rather unexpected and, worse, breaks the relationship between the internal state of the window (m_isShown) and its actual visibility.

We could fix the latter problem by calling Show() explicitly ourselves, but it seems better to just not show the children.

Closes #23997.

### DIFF
--- a/src/msw/nonownedwnd.cpp
+++ b/src/msw/nonownedwnd.cpp
@@ -255,6 +255,14 @@ WXLRESULT wxNonOwnedWindow::MSWWindowProc(WXUINT message, WXWPARAM wParam, WXLPA
                                             wxRectFromRECT(*prcNewWindow));
             }
             break;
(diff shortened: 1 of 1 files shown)


vadz:
Applied as the commit above.

PR #26382, opened 2026-04-15
Slightly simplify and make more robust CMake version parsing

vadz:
Check that we parsed all version components and avoid doing 2 regex matches for every component when we can just use CMAKE_MATCH_1 directly.

----

@MaartenBent This is very minor, but I was just implementing something similar in my own code and I was wondering why did wx do it in the more complicated way. I couldn't find any answer (`CMAKE_MATCH_1` seems to be available since always), so I've decided to change it and ask you if you see anything wrong with it, so that I could fix my own code if necessary. And if not, I think this is slightly better, isn't it?


MaartenBent:
I don't know why, it has been there since CMake support was added.

Personally I would use a macro instead of a function, and then get rid of `PARENT_SCOPE`.

vadz:
Thanks, I'll change it to the macro — I tend to avoid them for hygienic reasons, but you're right that it should probably be fine for such a simple case.

vadz:
Applied in the commit above.

PR #26383, opened 2026-04-15
Honour `WX_APPNAME_DATA_DIR` environment variable in all ports

vadz:
Telling the program where to find its data/resource files can be useful under Windows too, so return the value of this variable if it is defined there too instead of always return the directory where the executable is located.


vadz:
Applied as the commit above.

PR #26390, opened 2026-04-17
Update interface for recent changes

mcorino:
missing changes from #26117

### DIFF
--- a/interface/wx/ribbon/page.h
+++ b/interface/wx/ribbon/page.h
@@ -46,7 +46,7 @@ public:
     wxRibbonPage(wxRibbonBar* parent,
                 wxWindowID id = wxID_ANY,
                 const wxString& label = wxEmptyString,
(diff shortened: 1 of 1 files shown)


mcorino:
@vadz @Randalphwa The updates for #26117 introduced some interface changes for which no doc updates were created.

mcorino:
I probably did not capture all missing doc updates but these were triggered by wxRuby3 build failures.

vadz:
Thanks, I'll apply this.

It would be really great to have a CI job doing what this part of building wxRuby does to avoid breaking your build all the time...

mcorino:
Ah, yes that would be nice. Unfortunately, currently that would require building all of wxRuby3 as the wxRuby3 project does on a weekly basis. And even that will not catch all omissions because added interface methods missing from the docs will not trigger any problems until some user tries to use them in code.

PR #26392, opened 2026-04-18
Don't use /MP option when not using MSVS CMake generator

vadz:
It is useless in this case as make/Ninja generators only compile a single file at once anyhow and can also be harmful when using clang-cl compiler which gives warnings about unsupported option.

---

@MaartenBent Please let me know if you have any objections or see a better way to do this, but I need some way to fix it to avoid getting warnings about unrecognized option for every source file when using clang-cl.


MaartenBent:
I think this is the best way to do it, as using `MSVC` alone is not enough.

vadz:
Applied in the commit above.

PR #26395, opened 2026-04-19
Add wxGraphicsContext support to wxSVGFileDC

Blake-Madden:
By calling `GetGraphicsContext()` on a `wxSVGFileDC`, you can now access a `wxSVGGraphicsContext` (derived from `wxGraphicsContext`) which supports:
- Cubic and quadratic Bézier curves via `wxGraphicsPath`
- Linear and radial gradient brushes
- Semi-transparent colors and alpha blending
- Arbitrary affine transformations

Basically, the `wxSVGGraphicsContext*` writes to its parent `wxSVGFileDC`'s SVG buffer as you issue drawing commands to it.

I added a demonstration to the SVG sample (basic stuff, but shows off the features).

The following images are of my output SVG. I'm able to get some nice effects, such as a sunflower:

image

Or a Chernoff face:

image

I believe the full graphics context API is available in here now (at least the parts that are analogous to SVG).


samples/svg/svgtest.cpp
```diff
@@ -547,6 +580,261 @@ void DrawOnDC(wxDC& dc, const int index)
...
+                break;
+
+            // Cyan rectangle with red border.
+            gc->SetPen(*wxRED_PEN);
+            gc->SetBrush(*wxCYAN_BRUSH);
+            gc->DrawRectangle(dc.FromDIP(10), dc.FromDIP(10),
+                              dc.FromDIP(100), dc.FromDIP(70));
+
+            // Transparent rounded rectangle.
+            gc->SetBrush(wxBrush("DARK ORCHID", wxBRUSHSTYLE_TRANSPARENT));
```
vadz: Just curious: why give a colour to a transparent brush?

include/wx/dcsvg.h
```diff
@@ -186,11 +199,38 @@ class WXDLLIMPEXP_CORE wxSVGFileDCImpl : public wxDCImpl
     // BeginAccessibleGroup() call.
     void EndAccessibleGroup();
 
+    // Open a group with the given opacity.
+    void BeginLayer(double opacity);
+
+    // Close the group opened by the most recent BeginLayer() call.
+    void EndLayer();
+
```
vadz, line 204: Should we have a RAII helper calling `{Begin,End}Layer()` in its ctor/dtor as usual?

include/wx/dcsvg.h
```diff
@@ -186,11 +199,38 @@ class WXDLLIMPEXP_CORE wxSVGFileDCImpl : public wxDCImpl
...
+    void SetCompositionMode(wxCompositionMode mode);
+#endif
+
     wxString GetSVGDocument() const;
 
     bool Save();
 
 private:
+
+    // String helpers used by both the DC and the SVG graphics context.
```
vadz: An alternative to putting them here would be to put them in `include/wx/private/svg.h` (maybe in some `wxSVG` namespace to avoid name clashes).

There is no real problem with doing it like you did, but having them in a completely private header would reduce rebuilds when changing anything here.

vadz (commented):
Thanks, this definitely looks very nice and useful!

I didn't have the time to look at the new code yet (but I will), however I do have to ask: why use such strange API instead of adding `wxSVGFileGC` deriving from `wxGraphicsContext` as I'd naively expect? Is it significantly simpler to implement in this way?

samples/svg/svgtest.cpp
Blake-Madden: Good point, just uses `wxTRANSPARENT_BRUSH` now.

Blake-Madden:
> why use such strange API instead of adding `wxSVGFileGC` deriving from `wxGraphicsContext` as I'd naively expect? Is it significantly simpler to implement in this way?

My main motivation was to add graphics context rendering to an existing `wxSVGFileGC`, I hadn't thought of a stand alone `wxSVGGraphicsContext` handling everything by itself. Should I try to explore that?

The other thing is I wanted to piggyback off of `wxSVGFileGC`'s plumbing, as it does most of the core SVG text building. This way, I just overload or add any special graphics context-like functions and use a parent `wxSVGFileGC` to do most of the content management.

include/wx/dcsvg.h
Blake-Madden: Done.

vadz:
> > why use such strange API instead of adding `wxSVGFileGC` deriving from `wxGraphicsContext` as I'd naively expect? Is it significantly simpler to implement in this way?
> 
> My main motivation was to add graphics context rendering to an existing `wxSVGFileGC`, I hadn't thought of a stand alone `wxSVGGraphicsContext` handling everything by itself. Should I try to explore that?

I think it would be a more natural API from the user point of view, so if it's not too complicated to implement it like this, I think it would be preferable.

From the implementation point of view, I believe it might make sense to refactor the code to add some `wxSVGFileWriter` (or maybe just `wxSVGWriter` if we might generalize it to other kinds of output) in `include/wx/private/svg.h` and use it from both `wxSVGFileDC` and `wxSVGFileGC`.

include/wx/dcsvg.h
Blake-Madden: done

Blake-Madden:
@vadz Fulfills the expected API now and "plumbing" has been abstracted.

include/wx/dcsvg.h
```diff
@@ -18,6 +18,10 @@
 #include "wx/filename.h"
 #include "wx/dc.h"
 
+#if wxUSE_GRAPHICS_CONTEXT
+#include "wx/graphics.h"
+#endif
+
```
vadz: This could/should be replaced by `wxGraphisContext` forward declaration.

samples/svg/svgtest.cpp
```diff
@@ -265,6 +302,258 @@ bool MyPage::OnSave(const wxString& filename)
 namespace
 {
 
+#if wxUSE_GRAPHICS_CONTEXT
+// Shared content for the two SVGGraphicsContext demo pages: each page acquires
+// the GC differently, but draws the same shapes.
+void DrawSVGGraphicsContextDemo(wxGraphicsContext* gc, wxDC& dc)
```
vadz: Minor, but why do we pass both `gc` and `dc` here? And why is the former passed by pointer while the latter by reference? Seems inconsistent...

include/wx/svggc.h
```diff
@@ -0,0 +1,233 @@
...
+    wxFont m_currentFont;
+    wxColour m_currentTextColour;
+
+    // Running current-transform-matrix plus save-state stack.
+    wxGraphicsMatrix m_transform;
+
+    struct State
+    {
+        wxGraphicsMatrix transform;
+        size_t clipNestingLevel;
```
vadz: I'd use `int`, this is not a size of anything.

include/wx/svggc.h
```diff
@@ -0,0 +1,233 @@
...
+    // matrix is the identity.
+    wxString GetCurrentTransformAttr() const;
+
+    // Push the path's local-coordinate bounding box through the current
+    // transform and feed the four corners to the parent DC's bounding box,
+    // so wxSVGFileDC::MaxX/MaxY include GC-rendered geometry.
+    void AccumulatePathBounds(wxGraphicsPathData* data);
+
+    wxSVGFileDCImpl* m_impl;
+    wxSVGWriter* m_writer;
```
vadz: Could these made `const` (as `T* const`, not `const T*`)? I think they should never change during the lifetime of the object.

src/common/dcsvg.cpp
```diff
@@ -563,8 +234,104 @@ wxString wxSVGAttributes::GetAsString() const
 
 wxIMPLEMENT_ABSTRACT_CLASS(wxSVGFileDCImpl, wxDCImpl);
 
-size_t wxSVGFileDCImpl::m_clipUniqueId = 0;
-size_t wxSVGFileDCImpl::m_gradientUniqueId = 0;
+size_t wxSVGWriter::ms_clipUniqueId = 0;
+size_t wxSVGWriter::ms_gradientUniqueId = 0;
+
+// ----------------------------------------------------------
+// wxSVGFileDCImpl - string helpers (static)
```
vadz, line 241: Could this be kept above, to reduce the diff? I guess it would still show nicely with `git --color-moved` but not in GitHub web UI.

src/common/svggc.cpp
```diff
@@ -0,0 +1,1469 @@
...
+                constexpr wxDouble twoPi = 2.0 * M_PI;
+                if ( std::abs(sweep) >= twoPi - 1e-9 )
+                {
+                    sweep = seg.clockwise ? twoPi : -twoPi;
+                }
+                else
+                {
+                    if ( seg.clockwise )
+                    {
+                        while ( sweep <= 0 ) sweep += twoPi;
```
vadz: I hesitated to comment on this, but it's written on 2 lines below and it would be nice to split these loops on 2 lines too, both for consistency and readability.

src/common/svggc.cpp
```diff
@@ -0,0 +1,1469 @@
...
+// ----------------------------------------------------------------------------
+
+wxIMPLEMENT_DYNAMIC_CLASS(wxSVGGraphicsRenderer, wxGraphicsRenderer);
+
+wxGraphicsRenderer* wxSVGGraphicsRenderer::Get()
+{
+    static wxSVGGraphicsRenderer s_renderer;
+    return &s_renderer;
+}
+
```
vadz, line 846: I'm not sure how useful is this, what can be done with the renderer returned from here?

include/wx/svggc.h
```diff
@@ -0,0 +1,233 @@
...
+class WXDLLIMPEXP_FWD_CORE wxSVGWriter;
+
+// Singleton renderer producing wxSVGGraphicsContext and its supporting
+// handle/data objects. This renderer does not know how to create contexts
+// from arbitrary DCs or windows - the SVG graphics context is only ever
+// constructed indirectly via wxSVGFileDCImpl::GetGraphicsContext().
+class WXDLLIMPEXP_CORE wxSVGGraphicsRenderer : public wxGraphicsRenderer
+{
+public:
+    static wxGraphicsRenderer* Get();
```
vadz, line 32: Maybe this should take the path of the output file (and/or some `wxSVGFile` struct containing the path, size, DPI and title, and reused here and in `wxSVGFileDC` ctor)?

vadz (commented):
Thanks, the API is much nicer now!

But I wonder if we could make it even more convenient by allowing to create `wxSVGGC` directly, without passing by `wxSVGDC`? In the sample we already have the latter anyhow, of course, but I'd expect application code to not need it when using `wxSVGGC`, so it looks like it could be useful to add a way to create it directly?

Also, I didn't look at the changes to the existing `dcsvg.cpp` carefully because it's difficult to do in the web UI, I'll do it later locally, but please let me know if there were any real changes that I should pay special attention to.

Thanks again!

include/wx/svggc.h
Blake-Madden: done

Blake-Madden: Same is true for the various depth levels. They were `size_t` before, but I can make them `int` for consistency.

include/wx/svggc.h
Blake-Madden: done

include/wx/dcsvg.h
Blake-Madden: done

Blake-Madden:
> But I wonder if we could make it even more convenient by allowing to create `wxSVGGC` directly, without passing by `wxSVGDC`? In the sample we already have the latter anyhow, of course, but I'd expect application code to not need it when using `wxSVGGC`, so it looks like it could be useful to add a way to create it directly?

I was mimicking the current `wxGraphicsContext` API. If you are wanting a `wxSVGGraphicsContext` that can also run on its own without taking a `wxSVGFileDC` I can explore that, but this will take a new CTOR or Create method that has no analog in `wxGraphicsContext`. I'll try adding a CTOR that takes the same params as `wxSVGFileDC`'s CTOR, based on your comments.

src/common/svggc.cpp
vadz: Sorry, this comment was misguided/misplaced, please ignore it (not deleting it just to avoid confusion, but please consider it deleted).

include/wx/svggc.h
vadz: This is also wrong, sorry, I'll explain what I meant in another comment.

vadz:
> > But I wonder if we could make it even more convenient by allowing to create `wxSVGGC` directly, without passing by `wxSVGDC`? In the sample we already have the latter anyhow, of course, but I'd expect application code to not need it when using `wxSVGGC`, so it looks like it could be useful to add a way to create it directly?
> 
> I was mimicking the current `wxGraphicsContext` API. If you are wanting a `wxSVGGraphicsContext` that can also run on its own without taking a `wxSVGFileDC` I can explore that, but this will take a new CTOR or Create method that has no analog in `wxGraphicsContext`. I'll try adding a CTOR that takes the same params as `wxSVGFileDC`'s CTOR, based on your comments.

This case doesn't fit nicely the `wxGraphicsContext`/`wxGraphicsRenderer` design, where the renderer is supposed to be able to create contexts for all the different DCs, which define the rendering target. In our case we have a single target, so I think it makes sense to have some

```cpp
wxGraphicsContext* CreateContext(const wxSVGOutput&);
```

in `wxSVGGraphicsRenderer` which would create the context associated with the given output, and the corresponding

```cpp
wxSVGGraphicsContext(const wxSVGOutput&);
```

in `wxSVGGraphicsContext`.

Right now `wxSVGFileDC` is used instead of the putative `wxSVGOutput` which kind of works, but is a bit weird: why should we need to create a `wxDC`-derived object just to create a `wxGC`? It would be better to just pass the same parameters as you pass when creating `wxSVGFileDC`, i.e.

```cpp
                const wxString& filename,
                int width = 320,
                int height = 240,
                double dpi = wxSVG_DEFAULT_DPI,
                const wxString& title = wxString()
```

directly to `wxSVGGC` ctor IMHO.

samples/svg/svgtest.cpp
Blake-Madden: Just takes a wxGraphicsContext now.

src/common/svggc.cpp
Blake-Madden: Done, fixed in a few places

src/common/svggc.cpp
Blake-Madden: Just used internally by other objects. I removed from docs and will be implemented "standalone" API separately.

Blake-Madden:
> Right now `wxSVGFileDC` is used instead of the putative `wxSVGOutput` which kind of works, but is a bit weird: why should we need to create a `wxDC`-derived object just to create a `wxGC`? It would be better to just pass the same parameters as you pass when creating `wxSVGFileDC`, i.e.
> 
> ```c++
>                 const wxString& filename,
>                 int width = 320,
>                 int height = 240,
>                 double dpi = wxSVG_DEFAULT_DPI,
>                 const wxString& title = wxString()
> ```
> 
> directly to `wxSVGGC` ctor IMHO.

Yes, working on that now. There will be a multifaceted API, where you can:

- **create a standalone `wxSVGGraphicsContext` with the same CTORs that `wxSVGFileDC` has,**
- create a `wxSVGGraphicsContext` and connect your own `wxSVGFileDC` into it,
- or create a `wxSVGFileDC` and access the `wxSVGGraphicsContext` within it.
 
The latter two are for drop-in compatibility with the rest of the DC/GraphicsContext APIs. (This plugs write into my existing code, I don't have to make any changes to get fancy GC rendering in by SVG.)

The first one is a new way of doing things that is much simpler, as you have requested.

Blake-Madden:
> Also, I didn't look at the changes to the existing `dcsvg.cpp` carefully because it's difficult to do in the web UI, I'll do it later locally, but please let me know if there were any real changes that I should pay special attention to.

Not really, but a bit of moving code around. I abstracted a lot of the string building logic into the common writer class so that the DC and GC can share the same writing logic.

src/common/dcsvg.cpp
Blake-Madden: I've reorged a bit of content in here into the writer class. Sorry, I'm not sure where you want this to go.

Blake-Madden:
OK, now you can do something like:

```
wxSVGGraphicsContext gc("/users/joe/output.svg", 800, 700, 72, "My title");
gc.SetBitmapHandler(new wxSVGBitmapEmbedHandler());

// amazing drawing code
// ...

gc.Save();
```

interface/wx/svggc.h
```diff
@@ -0,0 +1,508 @@
...
+    access to advanced vector features not available in the standard
+    @ref wxDC API, such as cubic and quadratic Bézier curves, gradient
+    brushes, and alpha blending.
+
+    This context is used when using @ref wxGraphicsContext on a @ref wxSVGFileDC.
+
+    Example of use:
+    @code
+    wxSVGFileDC dc("output.svg", 800, 600);
+    wxGraphicsContext* gc = dc.GetGraphicsContext();
```
vadz: Maybe the example should show creating `gc` directly from the output file name etc? I think it's more natural for a code without any compatibility constraints (i.e. I don't see why would you create `wxSVGFileDC` in the new code).

interface/wx/svggc.h
```diff
@@ -0,0 +1,508 @@
...
+{
+public:
+    /**
+        Constructor.
+
+        The DC must outlive the context.
+
+        @since 3.3.3
+    */
+    wxSVGGraphicsContext(wxSVGFileDC& dc);
```
vadz:
```suggestion
    explicit wxSVGGraphicsContext(wxSVGFileDC& dc);
```

include/wx/svggc.h
```diff
@@ -0,0 +1,298 @@
...
+// the shared SVG buffer owned by wxSVGFileDCImpl. Constructed only via
+// wxSVGFileDCImpl::GetGraphicsContext(); state (pen/brush/font) is mirrored
+// back onto the owning DC so NewGraphicsIfNeeded() can keep the output
+// well-formed regardless of whether drawing came through the DC or the GC.
+class WXDLLIMPEXP_CORE wxSVGGraphicsContext : public wxGraphicsContext
+{
+public:
+    explicit wxSVGGraphicsContext(wxSVGFileDCImpl* impl);
+    explicit wxSVGGraphicsContext(wxSVGFileDC& dc);
+    wxSVGGraphicsContext(const wxString& filename,
```
vadz:
```suggestion
    explicit wxSVGGraphicsContext(const wxString& filename,
```

(this should have been done for `wxSVGFileDC` ctor(s) too, of course, but at least let us do it for the new class).

include/wx/svggc.h
```diff
@@ -0,0 +1,298 @@
...
+                         int width = 320,
+                         int height = 240,
+                         double dpi = wxSVG_DEFAULT_DPI,
+                         const wxString& title = wxString());
+    wxSVGGraphicsContext(const wxSize& size,
```
vadz:
```suggestion
    explicit wxSVGGraphicsContext(const wxSize& size,
```

interface/wx/svggc.h
```diff
@@ -0,0 +1,508 @@
...
+
+    /**
+        Constructor for creating an SVG file directly.
+
+        @since 3.3.3
+    */
+    wxSVGGraphicsContext(const wxString& filename,
```
vadz:
```suggestion
    explicit wxSVGGraphicsContext(const wxString& filename,
```

interface/wx/svggc.h
```diff
@@ -0,0 +1,508 @@
...
+                         int height = 240,
+                         double dpi = wxSVG_DEFAULT_DPI,
+                         const wxString& title = wxString());
+
+    /**
+        Constructor for creating an SVG file directly.
+
+        @since 3.3.3
+    */
+    wxSVGGraphicsContext(const wxSize& size,
```
vadz, line 65:
```suggestion
    explicit wxSVGGraphicsContext(const wxSize& size,
```

interface/wx/svggc.h
```diff
@@ -0,0 +1,508 @@
...
+*/
+class wxSVGGraphicsContext : public wxGraphicsContext
+{
+public:
+    /**
+        Constructor.
+
+        The DC must outlive the context.
+
+        @since 3.3.3
```
vadz, line 45: There is no need to use `@since 3.3.3` for all member functions when the class itself has it IMO.

interface/wx/svggc.h
```diff
@@ -0,0 +1,508 @@
...
+        settings) of the context on a stack.
+        Multiple balanced calls to PushState() and PopState() can be nested.
+    */
+    virtual void PushState() override;
+
+    /**
+        Pop the last state from the stack and restore the previous state.
+    */
+    virtual void PopState() override;
+    ///@}
```
vadz: I'm not sure we really need to repeat the documentation for the base class functions here, I'd leave only SVG-specific documentation here.

interface/wx/svggc.h
```diff
@@ -0,0 +1,508 @@
...
+                                             wxDouble w, wxDouble h) override;
+    ///@}
+
+    /**
+        Returns the name of the renderer, which is always "svg".
+    */
+    virtual wxString GetName() const override;
+
+    /**
+        Returns the version of the renderer.
```
vadz: Not really clear what does this mean for SVG?

src/common/dcsvg.cpp
vadz: What I meant is that this looks like the same code that is marked as removed above when viewing diff in GitHub. If we could avoid moving it, i.e. put `NumStr()` just under `SVG_DPI` definition, the diff should become much smaller/easier to read.

src/common/dcsvg.cpp
```diff
@@ -1379,24 +1124,47 @@ void wxSVGFileDCImpl::DoGetTextExtent(const wxString& string,
...
+    if ( scale != 1.0 && scale != 0.0 )
+    {
+        // SVG is DPI-independent, so we want text metrics at the default (96) DPI.
+        // Text does not scale linearly (at least on MSW), so scale the font size,
+        // rather than dividing the returned extent.
+        font.SetFractionalPointSize(font.GetFractionalPointSize() / scale);
+    }
+
+#if wxUSE_GRAPHICS_CONTEXT
+    // Measure through a wxGCDC so widths come from the platform's modern font engine
```
vadz, line 1143: `GetTextExtent()` must return results consistent with `DrawText()` and it's not really obvious why is wxGC more consistent with it than wxDC. It would be nice to explain this if you know the reason.

src/common/svggc.cpp
```diff
@@ -0,0 +1,1536 @@
...
+                Extend(s.x, s.y);
+                break;
+            case wxSVGPathSegment::QuadCurveSegment:
+                m->TransformPoint(&s.x1, &s.y1);
+                m->TransformPoint(&s.x, &s.y);
+                Extend(s.x1, s.y1);
+                Extend(s.x, s.y);
+                break;
+            case wxSVGPathSegment::ArcSegment:
+                break;
```
vadz, line 428: This should be unreachable, right?

vadz (approved):
Thanks, there is a lot of code here and I still couldn't really read all of it, but superficially everything looks good. It would be really great to have at least some unit tests for this, i.e. use the new class and verify that the output corresponds to the expected one — it would be obviously too difficult to cover all SVG features in a test suite, but at least some basic tests checking that this works as expected would be very nice.

The only other thing is that wxGraphics API, with its context/renderer split is not really appropriate for this kind of things, but I don't think there is anything we can do about it and have no choice but to provide all these dummy `CreateContext()` overloads always returning `nullptr` in the renderer. Let me cc @csomor just in case I'm missing some better way this could be done.

If there are no other comments, I'll merge it soon, with just the minor `explicit` additions below.

Thanks again!

interface/wx/svggc.h
Blake-Madden: The SVG specification that it targets. Updating the docs now...

interface/wx/svggc.h
Blake-Madden: Removing redundant docs...

src/common/svggc.cpp
Blake-Madden: I'll add a `wxFAIL_MSG`

src/common/dcsvg.cpp
Blake-Madden: From my testing, this produces more accurate text sizing (e.g., labels would fit inside of measured rects correctly, while using `wxDC` I would have text going outside of them). It was fairly noticeable, especially at smaller scales.

interface/wx/svggc.h
Blake-Madden: done

vadz:
Thanks for the latest fixes! I've squashed all commits with some minor changes (moving things around to minimize the diff, restoring the probably inadvertently deleted comments and reintroducing `write()` helper to minimize the diff even more) and will push them soon.

src/common/dcsvg.cpp
csomor: Sorry, I needed some time to look at this and to build things on both sides MSW and OSX. The resulting file on macOS does not have the correct text boundaries; things look different on the screen sizewise, the font appears to be smaller, thus when its width is used, the text looks very compressed when rendering the file. The text is smaller on screen, somehow looks to me like the old 96 vs 72 DPI problem. But I haven't found the time to dive into it yet.

Conceptually, the code generation would belong in the renderer, and the SVG content would belong to the context, but it's perfectly ok that things are piped through wxSVGFileDCImp right now.

vadz: > Sorry, I needed some time to look at this and to build things on both sides MSW and OSX. The resulting file on macOS does not have the correct text boundaries; things look different on the screen sizewise, the font appears to be smaller, thus when its width is used, the text looks very compressed when rendering the file. The text is smaller on screen, somehow looks to me like the old 96 vs 72 DPI problem. But I haven't found the time to dive into it yet.

Oops, thanks for noticing this. @Blake-Madden Are you able to test under Mac (and hopefully fix this there, although  I'm not sure if it's a regression or not)?

> Conceptually, the code generation would belong in the renderer, and the SVG content would belong to the context, but it's perfectly ok that things are piped through wxSVGFileDCImp right now.

It's mostly all these overloads of pure virtual `CreateContext()` in `wxGraphicsRenderer` that seem problematic to me because I don't see how can they make sense for SVG renderer which can be only ever be used with a single type of DC and nothing else. And OTOH the existing static `wxGraphicsContext::Create()` overloads are fine, but should we really add an overload for `wxSVGFileDC` not that we have a SVG renderer? This just doesn't seem right...

csomor: > It's mostly all these overloads of pure virtual `CreateContext()` in `wxGraphicsRenderer` that seem problematic to me because I don't see how can they make sense for SVG renderer which can be only ever be used with a single type of DC and nothing else. And OTOH the existing static `wxGraphicsContext::Create()` overloads are fine, but should we really add an overload for `wxSVGFileDC` not that we have a SVG renderer? This just doesn't seem right...

No it definitely does not, I'll think about how this could be solved in a more elegant way

Blake-Madden: In defense of the API, I don't have to create a `wxSVGGraphicsRenderer` anywhere in my codebase to actually use it. My graphics system deals with `wxGCDC` throughout the API and accesses its `wxGraphicsRenderer` when I need to do more advanced drawing. When I export as SVG, I just pass a `wxSVGFileDC` to my system and under the hood it gets the `wxSVGGraphicsRenderer` implicitly and does all the fancy drawing that was being done before now. I had to make zero changes to my program to benefit from `wxSVGGraphicsRenderer`. (Before, my calls to getting the `wxGraphicsRenderer` from the DC would fallback to a more basic rendering system if it returned `nullptr`, now it just works).

I agree that the (contractual) overloads are weird, but it is nice having `wxSVGGraphicsRenderer` derived from `wxGraphicsRenderer` so that works seamlessly with the rest of the API.

Blake-Madden: > Oops, thanks for noticing this. @Blake-Madden Are you able to test under Mac (and hopefully fix this there, although I'm not sure if it's a regression or not)?

That's where I was originally working on this, but I'll take another look. My experience was the SVG text output was sometimes wider than it appeared in the UI and would go out of its bounding boxes.

csomor:
To add a post with the differences

SVG graphics context on Mac

SVGDemoMac

SVG graphics context on Win

SVGDemoWin

SVG File created on Mac

SVG graphics context (direct) Mac

SVG File created on Win

SVG graphics context (direct) Win

MaartenBent:
To me it looks like the result of `wxSVGFileDCImpl::DoGetTextExtent` is incorrect now, the width is too small resulting in an incorrect `textLength` attribute and squashed text.

For `Drawn via wxGraphicsContext`, the textLength is 161 instead of 215. Which is a factor 1.33 (or 0.75) different. The same as 72/96. So maybe font scaling is not needed anymore (on macOS, what about Cairo?). Or it is scaled twice. Or you need to scale with `gcdc.GetPPI()` instead of `memDc.GetPPI()`.

When I worked on the text metric stuff in `wxSVGFileDC`, my goal was to get the same SVG output on all 3 platforms.
So in my opinion the font size itself is correct in the macOS SVG images. Even though the font in the app looks smaller. 
See also #25723 where I posted screenshots from the text-page of the SVG sample, and mention that macOS renders a smaller font.

MaartenBent:
Looks like the font is not scaled at all, because scale is `0`. Which happens because it is dividing two integers before casting it to a double:
https://github.com/wxWidgets/wxWidgets/blob/bbad4b8eeb8f36e524a0859374fff4ae07ec0d49/src/common/dcsvg.cpp#L1389-L1390

It used to be:
https://github.com/wxWidgets/wxWidgets/blob/bbad4b8eeb8f36e524a0859374fff4ae07ec0d49/src/common/dcsvg.cpp#L371-L373

csomor:
@MaartenBent I can confirm that this fixes the file created, it now looks the same as the windows version

SVG graphics context (direct)2

vadz:
Thanks for finding this, I'll push the fix soon if @Blake-Madden doesn't mind.

I've also realize that I've mistakenly committed this as myself instead of using the proper commit author, really sorry about this, but unfortunately it's too late to change it now :-(

Blake-Madden:
> Thanks for finding this, I'll push the fix soon if @Blake-Madden doesn't mind.

Of course, and sorry for the mix up.


> I've also realize that I've mistakenly committed this as myself instead of using the proper commit author, really sorry about this, but unfortunately it's too late to change it now :-(

You own that code now 😉

csomor:
@MaartenBent Now that the file is consistent, I still have the problem of the difference when `DrawSVGGraphicsContextDemo` is painting on the native graphics context (where the native device-independent Point is 1/72 of an inch on macOS, while SVG is using 1/96):

a) The rendering of the entire page is smaller
b) The font used is even smaller; you see it merely touches the red circle

a) The line `const double scale = dc.FromDIP(100) / 100.0`; does not help us here, as we are already on device-independent drawing, but on Mac this would have to be multiplied by 96.0/72.0 in order to have the same real sizes. 

b) Still, this would not help with the font size of 12 points, because SVG is also there based on a 96 DPI, and macOS on 72, which means that the font size itself would have to be multiplied as well. So, instead of adjusting at the time of measurement, should we use a different point size when creating the real font size? kepping the one used in the file still the same of course

I could do a PR, but I'd need some guidance on how to best approach this in the sense of your code. Thanks

csomor:
PS: perhaps we'd need the 'point-scaling' information somewhere in the wxGraphicsContext as it's needed in each of these two situations, @vadz perhaps you have some ideas on this as well, the basic problem is, that the font on macOS is per default calculated on the basis of 72 of an inch, if you scale the graphics context, you scale everything, so the relative error when compared to a font with an 96 of an inch point everywhere else remains

vadz:
Sorry, I don't understand the problem fully :-( Does an SVG containing text in, say, 10pt, font appear differently under Mac? This would be very surprising, so I guess this is not it. Perhaps we're just rendering it incorrectly in the sample?

csomor:
The SVG File now is correct, same appearance since @MaartenBent reintroduced the proper scaling of the measurement. But the 2 SVG rendering tabs in the sample show a font too small. The same factor that is used for measuring, would have to be applied when rendering the font. I'll try to come up with a precise description of the source of the problem.

csomor:
If I add on mac, to our sample file the following:
```
<text x="150" y="180" font-size="12pt">Text of 12 Pt</text>
<text x="20" y="180" font-size="16px"> Text of 16 Px</text>
```
Then the two text elements are rendered as the same font size. Thus, SVG rendering assumes that 1 pt = 1.33 px, corresponding to a 96 dpi renderer for the 'native' SVG of 72 dpi. Now, when the same svgdemo code is rendered natively on Mac on a graphics context, where 1 pt = 1 logical pixel, the font is rendered too small. So in order to receive the same relative text size, I'd have to scale the font size from 12 to 16 ...

MaartenBent:
> a) The rendering of the entire page is smaller

In my opinion this is not really a problem, it is just a macOS quirk. `wxGraphicsContext` drawings still looks the same, just a bit smaller.

> b) The font used is even smaller; you see it merely touches the red circle

This is a more visible problem, because the same drawing code (with text) looks different on macOS compared to Windows/Linux.
I suppose you could try to scale the font (point size) before drawing so it matches the size at 96dpi. But there are lots of side effects to consider. Like how to deal with the textmetric functions. And `GetFont`/`Get(Fractional)PointSize` should still return the correct values, unmodified/scaled.
And `wxDC`/`wxGraphicsContext` might also be used for drawing UI elements (at least on Windows it is), where maybe you do not want scaling. Because then it looks different compared to native UI text.

PR #26397, opened 2026-04-19
Fix positioning images in generic wx{List,Tree}Ctrl in high DPI

### LINKED ISSUE #26391: wxListCtrl in Report mode overlaps text and images when using wxBitmapBundle with display scaling.
### Description

Using the following code which adds `wxBitmapBundle` with different configurations as images, the text and the images overlap.

Reproduces on KDE/Wayland with display scaling set to 125% (downscaled 200%).
Probably happens on macOS as well.

Code: [listtest.cpp.txt](https://github.com/user-attachments/files/26838183/listtest.cpp.txt)

Image

- wxWidgets version  you use: wx 3.3.3 (master)
- OS  and its version: Manjaro
  
  + Desktop environment : KDE
  + Current theme:  Breath

vadz:
The implementation of GetImageLogicalSize() overload taking the image index returned correct answer for this individual image, but was wrong when used in the control with multiple images with different logical sizes. This resulted in the actual bitmap being drawn, as returned by GetImageBitmapFor(), being of different size than the size returned by this function, as GetImageBitmapFor() uses the best ("consensus") size for all images, which is also returned by GetImageLogicalSize() overload not taking the image.

Solve this by simply getting rid of the overload taking the image and replacing all existing calls to it with the calls to the other overload. This is backwards-incompatible, but this function was only added in 3.3 and removing it is a lesser evil than either leaving a function which can only be misused, or keeping the unused image index parameter which would be confusing.

Closes #26391.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -251,6 +251,9 @@ Changes in behaviour which may result in build errors
   configurations using EGL is not available any longer as exposing it doesn't
   make sense when wxGLCanvas may use either EGL or GLX.
 
+- wxWithImages::GetImageLogicalSize() overload taking the icon index didn't
+  make sense and was removed, please use the other overload instead if needed.
+
 
 3.3.3: (released 2026-??-??)
 ----------------------------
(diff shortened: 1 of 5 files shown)


vadz:

PR #26398, opened 2026-04-20
wxMSW: A better (IMHO) RTL fixes for wx{Auto}BufferedDC

AliKet:
Also fix custom drawing where `wxDC::DeviceToLogical()` and related functions need to be used.

### DIFF
--- a/src/msw/dc.cpp
+++ b/src/msw/dc.cpp
@@ -475,27 +475,18 @@ void wxMSWDCImpl::DoSetClippingRegion(wxCoord x, wxCoord y, wxCoord w, wxCoord h
     // 4 corners of the rectangle to create a polygonal clipping region
     // in device coordinates.
     POINT rect[4];
-    {
-        // Forcing LTR layout to calculate _rect_, otherwise the region created
-        // by CreatePolygonRgn() will not be correct in RTL layout.
-        const auto oldLayoutDir = GetLayoutDirection();
-        SetLayoutDirection(wxLayout_LeftToRight);
-
-        wxPoint p = LogicalToDevice(x, y);
-        rect[0].x = p.x;
-        rect[0].y = p.y;
-        p = LogicalToDevice(x + w, y);
-        rect[1].x = p.x;
-        rect[1].y = p.y;
-        p = LogicalToDevice(x + w, y + h);
-        rect[2].x = p.x;
-        rect[2].y = p.y;
-        p = LogicalToDevice(x, y + h);
-        rect[3].x = p.x;
-        rect[3].y = p.y;
-
-        SetLayoutDirection(oldLayoutDir);
-    }
+    wxPoint p = LogicalToDevice(x, y);
+    rect[0].x = p.x;
+    rect[0].y = p.y;
+    p = LogicalToDevice(x + w, y);
+    rect[1].x = p.x;
+    rect[1].y = p.y;
+    p = LogicalToDevice(x + w, y + h);
+    rect[2].x = p.x;
+    rect[2].y = p.y;
+    p = LogicalToDevice(x, y + h);
+    rect[3].x = p.x;
+    rect[3].y = p.y;
 
     HRGN hrgn = ::CreatePolygonRgn(rect, WXSIZEOF(rect), WINDING);
     if ( !hrgn )
--- a/src/msw/dc.cpp
+++ b/src/msw/dc.cpp
@@ -2158,8 +2158,7 @@ bool wxMSWDCImpl::DoStretchBlit(wxCoord xdest, wxCoord ydest,
     xdest += XLOG2DEV(0);
     ydest += YLOG2DEV(0);
 
-    // Adjust for RTL layout to fix problems during scrolling, see #26266.
-    const int xsrcOrig = GetLayoutDirection() == wxLayout_LeftToRight ? xsrc : -xsrc;
+    const int xsrcOrig = xsrc;
     const int ysrcOrig = ysrc;
 
     // This does the same thing as XLOG2DEV() but for the source DC.
@@ -2392,6 +2391,22 @@ bool wxMSWDCImpl::DoStretchBlit(wxCoord xdest, wxCoord ydest,
                 {
                     // reflect ysrc
                     ysrc = hDIB - (ysrc + srcHeight);
+                }
+
+                if ( GetLayoutDirection() == wxLayout_RightToLeft )
+                {
+                    // Unlike BitBlt() and StretchBlt(), StretchDIBits() doesn't
+                    // honor the LAYOUT_RTL flag set on the DC, so we have to apply
+                    // mirroring ourselves (see SetLayout() documentation in MSDN).
+                    const LONG wDIB = ds.dsBmih.biWidth;
+                    if ( wDIB > 0 )
+                    {
+                        // reflect wsrc
+                        xsrc = wDIB - (xsrc + srcWidth);
+
+                        xdest += dstWidth;
+                        dstWidth = -dstWidth;
+                    }
                 }
 
                 if ( ::StretchDIBits(GetHdc(),
--- a/src/msw/dc.cpp
+++ b/src/msw/dc.cpp
@@ -69,7 +69,7 @@ wxIMPLEMENT_ABSTRACT_CLASS(wxMSWDCImpl, wxDCImpl);
 
 // The device space in Win32 GDI measures 2^27*2^27 , so we use 2^27-1 as the
 // maximal possible view port extent.
-static const int VIEWPORT_EXTENT = 134217727;
+static const int VIEWPORT_EXTENT = 134217726;
 
 // ROPs which don't have standard names (see "Ternary Raster Operations" in the
 // MSDN docs for how this and other numbers in wxDC::Blit() are obtained)
--- a/src/msw/dc.cpp
+++ b/src/msw/dc.cpp
@@ -38,6 +38,7 @@
 #include "wx/msw/dc.h"
 
 #include "wx/scopedarray.h"
+#include "wx/scopeguard.h"
 #include "wx/sysopt.h"
 
 #ifdef wxHAS_RAW_BITMAP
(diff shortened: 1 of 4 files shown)


AliKet:
I don't have a real explanation of why we need this commit da9c92e to fix tests failures in `OneDevRegionRTL`, Just some weired rounding errors? I'm guessing...

AliKet:
Micro-optimisation:
```C++
diff --git a/src/msw/dc.cpp b/src/msw/dc.cpp
index dc56fa3e53..dda510a168 100644
--- a/src/msw/dc.cpp
+++ b/src/msw/dc.cpp
@@ -2379,10 +2379,12 @@ bool wxMSWDCImpl::DoStretchBlit(wxCoord xdest, wxCoord ydest,
                 // automatically as it doesn't even work with the source HDC.
                 // So do this manually to ensure that the coordinates are
                 // interpreted in the same way here as in all the other cases.
-                xsrc = source->LogicalToDeviceX(xsrcOrig);
-                ysrc = source->LogicalToDeviceY(ysrcOrig);
-                srcWidth = source->LogicalToDeviceXRel(srcWidth);
-                srcHeight = source->LogicalToDeviceYRel(srcHeight);
+                const auto devPos = implSrc->LogicalToDevice(xsrcOrig, ysrcOrig);
+                const auto devSize = implSrc->LogicalToDeviceRel(srcWidth, srcHeight);
+                xsrc = devPos.x;
+                ysrc = devPos.y;
+                srcWidth = devSize.x;
+                srcHeight = devSize.y;
 
                 // Figure out what co-ordinate system we're supposed to specify
                 // ysrc in.

```

`LogicalToDevice{X,Y}[Rel]()` now internally call `SetLayout()` to force LTR before calling `LPtoDP()` as a pre-condition and restore the old layout when the function returns as a post-condition. So there will be 8 calls to `SetLayout()` with the old code, but 4 with the patch above. I didn't try to profile this to see whether the patch makes any difference, but I still prefer to apply it, though.

AliKet:
@vslavik I would be grateful if you could test this and see whether it breaks something I’m not aware of. Thanks in advance!

src/msw/dc.cpp
```diff
@@ -68,7 +69,7 @@ wxIMPLEMENT_ABSTRACT_CLASS(wxMSWDCImpl, wxDCImpl);
 
 // The device space in Win32 GDI measures 2^27*2^27 , so we use 2^27-1 as the
 // maximal possible view port extent.
-static const int VIEWPORT_EXTENT = 134217727;
+static const int VIEWPORT_EXTENT = 134217726;
```
vadz: I'm not sure why did this change, but the comment above needs to be updated as it's now `2^27-2` and not `-1`. Ideal, of course, would be to also explain why is an extra `-1` needed.

src/msw/dc.cpp
```diff
@@ -101,6 +102,19 @@ static const int VIEWPORT_EXTENT = 134217727;
...
+// ----------------------------------------------------------------------------
+// macro to temporarily disable RTL layout if already set on the device context
+// ----------------------------------------------------------------------------
+
+// Mainly used with the LogicalToDevice{Rel}()/DeviceToLogical{Rel}() functions
+// to force them to return unmirrored coordinates if the LAYOUT_RTL flag is set
+// on the device context. This avoids double-mirroring problems when the result
+// is passed to GDI drawing functions. And also to be consistent with wxGTK3.
+
+#define wxSCOPED_DC_RTL_DISABLER(hdc)              \
```
vadz: I'd prefer to replace this macro with `wxScopedRTLDisabler` class that would be created with the `HDC` to disable RTL layout for.

src/msw/dc.cpp
```diff
@@ -2381,6 +2393,22 @@ bool wxMSWDCImpl::DoStretchBlit(wxCoord xdest, wxCoord ydest,
                     ysrc = hDIB - (ysrc + srcHeight);
                 }
 
+                if ( GetLayoutDirection() == wxLayout_RightToLeft )
+                {
+                    // Unlike BitBlt() and StretchBlt(), StretchDIBits() doesn't
+                    // honor the LAYOUT_RTL flag set on the DC, so we have to apply
+                    // mirroring ourselves (see SetLayout() documentation in MSDN).
+                    const LONG wDIB = ds.dsBmih.biWidth;
+                    if ( wDIB > 0 )
```
vadz, line 2425: Can it really be 0 here? It might be clearer/simpler to return early from this function if the width (or height) is 0.

tests/graphics/clippingbox.cpp
```diff
@@ -1030,7 +1030,7 @@ static void OneDevRegionRTL(wxDC& dc, const wxBitmap& bmp, bool useTransformMatr
 
     // Setting one clipping region in device coordinates
     // inside transformed DC area.
-    const int x = 11;
+    const int x = 10;
```
vadz: Curious why changing this still works for wxGTK?

vadz (commented):
Sorry, I'm still very confused by this stuff. It would be great to have an explanation in a comment somewhere about how RTL works in wxMSW, i.e. which coordinates take into account and which don't. Right now I don't understand why is is it correct to disable RTL when converting between device and logical, hopefully it would become clearer with a comment like this...

vslavik:
> @vslavik I would be grateful if you could test this and see whether it breaks something I’m not aware of. Thanks in advance!

Sorry, I wanted to, but ran out of time and won’t be able to until May 12-ish.

vadz:
I'm still not sure if it this is correct, it seems wrong to disable RTL when doing device/logical conversions. But if this is how the other ports work and if there is no good way to fix this otherwise, we could still merge this...

@vslavik Please let me know what do you think when you get back.

vadz:
@AliKet What (if anything) has changed since the previous version after the last force push?

AliKet:
> @AliKet What (if anything) has changed since the previous version after the last force push?

Nothing interesting, just rebased on master and updating the last commit:
```C++
> git diff HEAD~1 HEAD

diff --git a/tests/graphics/clippingbox.cpp b/tests/graphics/clippingbox.cpp
index 802f62d29c..bd11ab4871 100644
--- a/tests/graphics/clippingbox.cpp
+++ b/tests/graphics/clippingbox.cpp
@@ -1057,15 +1057,9 @@ static void OneDevRegionRTL(wxDC& dc, const wxBitmap& bmp, bool useTransformMatr
     dc.Clear();
     // right physical edge becomes left logical edge in a mirrored DC.
     const int x2 = s_dcSize.x - (x + w);
-    // In a mirrored DC, the device origin (0, 0) is always at the top left
-    // of the DC under wxMSW, but under wxGTK3 is at the top right.
-#if defined(__WXGTK3__) || defined(__WXQT__)
+
     wxPoint pos = dc.DeviceToLogical(x, y);
     wxSize dim = dc.DeviceToLogicalRel(w, h);
-#else
-    wxPoint pos = dc.DeviceToLogical(s_dcSize.x-x, y);
-    wxSize dim = dc.DeviceToLogicalRel(-w, h);
-#endif

     CheckClipBox(dc, bmp,
                  pos.x, pos.y, dim.x, dim.y,

```

vadz:
Thanks!

I'd still like to have some explanation of how RTL works (i.e. what is mirrored and what is not) somewhere, but I'm ready to merge this anyhow.

Pinging @vslavik one last time just in case.

AliKet:
@vadz let me know if this commit [886240d](https://github.com/wxWidgets/wxWidgets/pull/26398/commits/886240dfb0a95a8d6633fd8d598d6536ad2ba591) should be extracted to its own PR?

vslavik:
I couldn’t reproduce any of the previous RTL issues in Poedit with this PR (or with `master`).

Like @vadz, I don’t really understand why this works, but I’m similarly confused by any device-vs-logical coordinates handling code, so this is a very humbling PR for me to look at...

In any case, explicit `wxScopedRTLDisabler` is definitely an improvement w.r.t. understandability of the code.

AliKet:
> I couldn’t reproduce any of the previous RTL issues in Poedit with this PR (or with `master`).
> 

Thanks a lot for testing this, Glad it worked for you.

> Like @vadz, I don’t really understand why this works, but I’m similarly confused by any device-vs-logical coordinates handling code, so this is a very humbling PR for me to look at...
> 

This **ChatGPT** answer about _LPtoDP/DPtoLP returning mirrored coordinates in RTL_ may help you understand why disabling RTL when calling these functions is the right thing to do to get correct results:

[gdi_rtl_lptodp_explanation.pdf](https://github.com/user-attachments/files/27978349/gdi_rtl_lptodp_explanation.pdf)

Or in other words: Coordinates passed to **GDI** functions must not be mirrored, as these functions will mirror them anyway, resulting in a double-mirror problem.

> In any case, explicit `wxScopedRTLDi
sabler` is definitely an improvement w.r.t. understandability of the code.

aliket83:
> 
> Or in other words: Coordinates passed to **GDI** functions must not be mirrored, as these functions will mirror them anyway, resulting in a double-mirror problem.
> 

Sorry, I mean those functions expecting logical units.

vadz:
> > I couldn’t reproduce any of the previous RTL issues in Poedit with this PR (or with `master`).
> 
> Thanks a lot for testing this, Glad it worked for you.

Yes, thanks for testing and thanks for all these changes, of course. I'm going to merge them now.
 
> > Like @vadz, I don’t really understand why this works, but I’m similarly confused by any device-vs-logical coordinates handling code, so this is a very humbling PR for me to look at...
> 
> This **ChatGPT** answer about _LPtoDP/DPtoLP returning mirrored coordinates in RTL_ may help you understand why disabling RTL when calling these functions is the right thing to do to get correct results:
> 
> [gdi_rtl_lptodp_explanation.pdf](https://github.com/user-attachments/files/27978349/gdi_rtl_lptodp_explanation.pdf)

Thanks, if this is correct (and it looks like you checked that it is), this indeed explains it and I had no idea this was true :-(
 
> Or in other words: Coordinates passed to **GDI** functions must not be mirrored, as these functions will mirror them anyway, resulting in a double-mirror problem.

I feel like this should be mentioned somewhere, but for now I'll put it into the commit message, at least.

Issue #26378, opened 2026-04-14
With Windows high contrast mode, selected text in wxDataViewCtrl is not readable


stevecor:
On Windows, selecting a contrast theme results in the selected text in the wxDataViewCtrl control drawn using the same or similar color as the selected item background color. There is little or no contrast between these colors, making the text either invisible or nearly invisible.

To repro:

1. Run the dataview sample. Selected text looks fine
2. In Windows Settings, go to Accessibility: Contrast themes
3. Select "Aquatic" and click Apply
4. Return to the dataview sample and see that the selected text is not visible

![Image](https://github.com/user-attachments/assets/80ef9820-572b-415a-ae12-df338187f274)

The text color should be the same as in the Windows File Explorer:

<img width="828" height="247" alt="Image" src="https://github.com/user-attachments/assets/31805d1a-fa7e-4d8c-bf80-fa7f3ec046c1" />

Windows 11 25H2, wxWidgets 3.3.2

vadz:
Could you please run the "drawing" sample and go to its "Screen » System colours" menu item and check the values of the listbox-related colours for this theme (or just make a screenshot of this screen)?

stevecor:
Here are the system colors for all four high contrast themes:

Aquatic:
<img width="800" height="790" alt="Image" src="https://github.com/user-attachments/assets/81222b49-eb7d-4a59-966c-328037be1670" />

Desert:
<img width="800" height="772" alt="Image" src="https://github.com/user-attachments/assets/b682bc99-f086-41a4-8e02-d79bea5b3d66" />

Dusk:
<img width="800" height="783" alt="Image" src="https://github.com/user-attachments/assets/4c2fc29b-eb1a-4c04-9476-f88fc2fcce6f" />

Night Sky:
<img width="800" height="779" alt="Image" src="https://github.com/user-attachments/assets/29c44e92-3341-4829-8f8a-be01101c36a8" />

stevecor:
For high contrast mode, I think the proper selected text color should be wxSYS_COLOUR_HIGHLIGHTTEXT rather than wxSYS_COLOUR_LISTBOXTEXT. wxRendererGeneric uses wxSYS_COLOUR_HIGHLIGHTTEXT.

However, wxSYS_COLOUR_HIGHLIGHTTEXT does not look right in regular non-contrast mode:

[...]

PR #26399, opened 2026-04-20
Fix wxRendererXP::DrawItemText() text color for selected item when in dark mode

stevecor:
The fix for #26378 inadvertently broke dark mode. The proposed solution is to check for dark mode, and in that case, get the color wxSYS_COLOUR_HIGHLIGHTTEXT.

### DIFF
--- a/src/msw/renderer.cpp
+++ b/src/msw/renderer.cpp
@@ -956,7 +956,7 @@ void wxRendererXP::DrawItemText(wxWindow* win,
                                 int flags,
                                 wxEllipsizeMode ellipsizeMode)
 {
-    wxUxThemeHandle hTheme(win, L"EXPLORER::LISTVIEW;LISTVIEW");
+    wxUxThemeHandle hTheme(win, L"EXPLORER::LISTVIEW;LISTVIEW", L"DarkMode::LISTVIEW");
 
     const int itemState = GetListItemState(flags);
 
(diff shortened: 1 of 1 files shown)


vadz:
The canonical way to test for dark mode in wxMSW code is to use `wxMSWDarkMode::IsActive()` but beyond this, I wonder if this could be fixed by using an alternate theme in dark mode here. E.g. could something like this work?

```diff
diff --git a/src/msw/renderer.cpp b/src/msw/renderer.cpp
index 101fcba4be..2cc01c3f5d 100644
--- a/src/msw/renderer.cpp
+++ b/src/msw/renderer.cpp
@@ -956,7 +956,7 @@ void wxRendererXP::DrawItemText(wxWindow* win,
                                 int flags,
                                 wxEllipsizeMode ellipsizeMode)
 {
-    wxUxThemeHandle hTheme(win, L"EXPLORER::LISTVIEW;LISTVIEW");
+    wxUxThemeHandle hTheme(win, L"EXPLORER::LISTVIEW;LISTVIEW", L"ItemsView");
 
     const int itemState = GetListItemState(flags);
 
```

(sorry, I didn't have time to test this myself yet).

stevecor:
Yes that seems to be the correct fix. The dark mode alternate theme name "DarkMode::LISTVIEW" occurs in a related function in this source, so I used that. Testing looks good.

vadz:

PR #26401, opened 2026-04-20
wxDC::LogicalToDevice() and sister functions now return unmirrored coordinates in RTL under wxQt

AliKet:


vadz:
I think this is independent of #26398 and can be merged without waiting for it, right?

AliKet:
> I think this is independent of #26398 and can be merged without waiting for it, right?

Yes, thanks in advance.

PR #26406, opened 2026-04-24
Missing doc updates for updated GetMinimisedIcon(Bundle) methods

mcorino:


mcorino:
@Randalphwa @vadz Another doc miss.

interface/wx/ribbon/panel.h
```diff
@@ -190,8 +190,8 @@ class wxRibbonPanel : public wxRibbonControl
         Get the bitmap to be used in place of the panel children when it is
         minimised.
     */
-    wxBitmap& GetMinimisedIcon();
-    const wxBitmap& GetMinimisedIcon() const;
+    wxBitmap GetMinimisedIcon();
+    const wxBitmap& GetMinimisedIconBundle() const;
```
vadz, line 194: This should be

```suggestion
    const wxBitmapBundle& GetMinimisedIconBundle() const;
```

I'll merge it with this fix a bit later.

PR #26409, opened 2026-04-26
Fix Retina display issues with ribbon

Blake-Madden:
In macOS, there were two issues:

- The area drawn around regular-sized buttons was twice the size it should be.
Screenshot 2026-04-25 at 12 52 33 PM
- Smaller sized buttons would display the icon at the wrong (too large) size.
Screenshot 2026-04-25 at 12 53 28 PM

This PR fixes the size calculation of the button area, bitmap scaling, and implements some dead code with small button sizing.

### DIFF
--- a/src/ribbon/buttonbar.cpp
+++ b/src/ribbon/buttonbar.cpp
@@ -519,18 +519,12 @@ void wxRibbonButtonBar::FetchButtonSizeInfo(wxRibbonButtonBarButtonBase* button,
         wxRibbonButtonBarButtonState size, wxReadOnlyDC& dc)
 {
     wxRibbonButtonBarButtonSizeInfo& info = button->sizes[size];
+    const double scale = GetDPIScaleFactor() / GetContentScaleFactor();
+    const wxSize bmpSizeLarge = m_bitmap_size_large * scale;
+    const wxSize bmpSizeSmall = m_bitmap_size_small * scale;
+
     if(m_art)
     {
-#ifdef __WXOSX__
-        // macOS DC is in logical units (points); use base sizes directly
-        const wxSize& bmpSizeLarge = m_bitmap_size_large;
-        const wxSize& bmpSizeSmall = m_bitmap_size_small;
-#else
-        // Windows DC is in physical pixels; scale bitmap sizes to match
-        const double scale = GetDPIScaleFactor();
-        const wxSize bmpSizeLarge = m_bitmap_size_large * scale;
-        const wxSize bmpSizeSmall = m_bitmap_size_small * scale;
-#endif
         info.is_supported = m_art->GetButtonBarButtonSize(dc, this,
             button->kind, size, button->label, button->text_min_width[size],
             bmpSizeLarge, bmpSizeSmall, &info.size,
@@ -950,11 +944,8 @@ void wxRibbonButtonBar::OnPaint(wxPaintEvent& WXUNUSED(evt))
     wxRibbonButtonBarLayout* layout = m_layouts.Item(m_current_layout);
 
     // Calculate DPI-scaled bitmap sizes (must match layout calculation)
-#ifdef __WXOSX__
-    const double scale = GetContentScaleFactor();
-#else
     const double scale = GetDPIScaleFactor();
-#endif
+    const double contentScale = GetContentScaleFactor();
     const wxSize scaledLarge = m_bitmap_size_large * scale;
     const wxSize scaledSmall = m_bitmap_size_small * scale;
 
@@ -978,10 +969,8 @@ void wxRibbonButtonBar::OnPaint(wxPaintEvent& WXUNUSED(evt))
                 const wxBitmapBundle& bundle = disabled ?
                     m_bundlesLargeDisabled[idx] : m_bundlesLarge[idx];
                 bitmap = bundle.GetBitmap(scaledLarge);
-#ifdef __WXOSX__
                 if ( bitmap.IsOk() )
-                    bitmap.SetScaleFactor(scale);
-#endif
+                    bitmap.SetScaleFactor(contentScale);
             }
         }
 
@@ -993,10 +982,8 @@ void wxRibbonButtonBar::OnPaint(wxPaintEvent& WXUNUSED(evt))
                 const wxBitmapBundle& bundle = disabled ?
                     m_bundlesSmallDisabled[idx] : m_bundlesSmall[idx];
                 bitmap_small = bundle.GetBitmap(scaledSmall);
-#ifdef __WXOSX__
                 if ( bitmap_small.IsOk() )
-                    bitmap_small.SetScaleFactor(scale);
-#endif
+                    bitmap_small.SetScaleFactor(contentScale);
             }
         }
 
--- a/src/ribbon/buttonbar.cpp
+++ b/src/ribbon/buttonbar.cpp
@@ -1184,14 +1184,12 @@ void wxRibbonButtonBar::MakeLayouts()
                               wxRIBBON_BUTTONBAR_BUTTON_MEDIUM);
         }
 
-        // TODO: small buttons are not implemented yet in
-        //       art_msw.cpp:2581 and will be invisible
-        /*iLast = btn_count;
+        iLast = btn_count;
         while(iLast-- > 0)
         {
             TryCollapseLayout(m_layouts.Last(), iLast, &iLast,
                               wxRIBBON_BUTTONBAR_BUTTON_SMALL);
-        }*/
+        }
     }
 }
 
--- a/src/ribbon/buttonbar.cpp
+++ b/src/ribbon/buttonbar.cpp
@@ -521,14 +521,19 @@ void wxRibbonButtonBar::FetchButtonSizeInfo(wxRibbonButtonBarButtonBase* button,
     wxRibbonButtonBarButtonSizeInfo& info = button->sizes[size];
     if(m_art)
     {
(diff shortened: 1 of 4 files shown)


src/ribbon/buttonbar.cpp
```diff
@@ -945,7 +950,11 @@ void wxRibbonButtonBar::OnPaint(wxPaintEvent& WXUNUSED(evt))
     wxRibbonButtonBarLayout* layout = m_layouts.Item(m_current_layout);
 
     // Calculate DPI-scaled bitmap sizes (must match layout calculation)
+#ifdef __WXOSX__
+    const double scale = GetContentScaleFactor();
+#else
```
vadz: This shouldn't be necessary at all, `GetDPIScaleFactor()` returns the same thing as `GetContentScaleFactor()` in wxOSX.

src/ribbon/buttonbar.cpp
```diff
@@ -969,6 +978,10 @@ void wxRibbonButtonBar::OnPaint(wxPaintEvent& WXUNUSED(evt))
                 const wxBitmapBundle& bundle = disabled ?
                     m_bundlesLargeDisabled[idx] : m_bundlesLarge[idx];
                 bitmap = bundle.GetBitmap(scaledLarge);
+#ifdef __WXOSX__
+                if ( bitmap.IsOk() )
+                    bitmap.SetScaleFactor(scale);
+#endif
```
vadz: This is very suspicious. First of all because if we need to check for anything, it should be `wxHAS_DPI_INDEPENDENT_PIXELS` and not `__WXOSX__` as wxGTK must be affected by the same problem. Second, and more importantly, because the bitmap returned from `wxBitmapBundle` must already have the correct scale factor.

src/ribbon/buttonbar.cpp
```diff
@@ -521,14 +521,19 @@ void wxRibbonButtonBar::FetchButtonSizeInfo(wxRibbonButtonBarButtonBase* button,
     wxRibbonButtonBarButtonSizeInfo& info = button->sizes[size];
     if(m_art)
     {
-        // Scale base bitmap sizes by current DPI for layout calculation
+#ifdef __WXOSX__
+        // macOS DC is in logical units (points); use base sizes directly
+        const wxSize& bmpSizeLarge = m_bitmap_size_large;
+        const wxSize& bmpSizeSmall = m_bitmap_size_small;
+#else
```
vadz: This is not necessary, it's enough to use `GetContentScaleFactor()` (which will return 1 under Windows) instead of `GetDPIScaleFactor()` below.

vadz (changes requested):
Sorry, I think changes here are not right. There should be really no need for any preprocessor checks, wx API provides all the necessary functions for translating between different kinds of pixels and `wxBitmapBundle` should already take care of the bitmap scale (if it really doesn't, it's a bad bug which needs to be fixed in it).

src/ribbon/buttonbar.cpp
Blake-Madden: `const double scale = GetDPIScaleFactor() / GetContentScaleFactor();` worked for me on both MSW and macOS

src/ribbon/buttonbar.cpp
Blake-Madden: I remove the guards, but I still need to call `SetScaleFactor` or the icons don't go down to 16x16 when the ribbon is small on macOS.

Blake-Madden:
circlecli error is from an unrelated Lexilla warning

src/ribbon/buttonbar.cpp
```diff
@@ -519,16 +519,15 @@ void wxRibbonButtonBar::FetchButtonSizeInfo(wxRibbonButtonBarButtonBase* button,
         wxRibbonButtonBarButtonState size, wxReadOnlyDC& dc)
 {
     wxRibbonButtonBarButtonSizeInfo& info = button->sizes[size];
+    const double scale = GetDPIScaleFactor() / GetContentScaleFactor();
+    const wxSize bmpSizeLarge = m_bitmap_size_large * scale;
+    const wxSize bmpSizeSmall = m_bitmap_size_small * scale;
```
vadz: I think this is identical to much more comprehensible

```suggestion
    const wxSize bmpSizeLarge = FromDIP(m_bitmap_size_large);
    const wxSize bmpSizeSmall = FromDIP(m_bitmap_size_small);
```

This should work as long as `m_bitmap_size_xxx` are indeed expressed in DIPs. Are they?

src/ribbon/buttonbar.cpp
```diff
@@ -969,6 +969,8 @@ void wxRibbonButtonBar::OnPaint(wxPaintEvent& WXUNUSED(evt))
                 const wxBitmapBundle& bundle = disabled ?
                     m_bundlesLargeDisabled[idx] : m_bundlesLarge[idx];
                 bitmap = bundle.GetBitmap(scaledLarge);
+                if ( bitmap.IsOk() )
+                    bitmap.SetScaleFactor(contentScale);
```
vadz: Sorry but this just can't be correct. `wxBitmapBundle` returns the bitmap with the scale factor consistent with the requested size, changing it later can only break things.

If you pass the correct size to `GetBitmap()` there should be no need for this.

src/ribbon/buttonbar.cpp
```diff
@@ -946,6 +945,7 @@ void wxRibbonButtonBar::OnPaint(wxPaintEvent& WXUNUSED(evt))
 
     // Calculate DPI-scaled bitmap sizes (must match layout calculation)
     const double scale = GetDPIScaleFactor();
+    const double contentScale = GetContentScaleFactor();
     const wxSize scaledLarge = m_bitmap_size_large * scale;
     const wxSize scaledSmall = m_bitmap_size_small * scale;
```
vadz: This is weird because it's _not_ the same as `FromDIP()`. But I think this just means this is wrong and this is why you have to work around this bug with `SetScaleFactor()` call below.

src/ribbon/buttonbar.cpp
```diff
@@ -980,6 +982,8 @@ void wxRibbonButtonBar::OnPaint(wxPaintEvent& WXUNUSED(evt))
                 const wxBitmapBundle& bundle = disabled ?
                     m_bundlesSmallDisabled[idx] : m_bundlesSmall[idx];
                 bitmap_small = bundle.GetBitmap(scaledSmall);
+                if ( bitmap_small.IsOk() )
+                    bitmap_small.SetScaleFactor(contentScale);
```
vadz: Same comment as above.

vadz (changes requested):
Sorry for the delay, but after looking at the latest version I am still almost sure it is wrong, I hope you can find the correct solution by just setting the sizes correctly (which means expressing them in either physical or DPI-independent pixels, as necessary).

src/ribbon/buttonbar.cpp
Blake-Madden: Should I try a different approach? Sorry, I'm just struggling with this one. This small icon issue was introduced when the `wxBitmapBundle` support was added, but I can't trace it to what specifically is causing the issue. This was the cleanest fix I could come up with that worked on all platforms, but I'm having difficulty following the interactions going on in here.

vadz: I am not sure what the issue is, so I can't say anything useful but the basic principles are:

1. It should be never necessary to use neither `GetDPIScaleFactor()` nor `GetContentScaleFactor()` explicitly, using `{To,From}DIP()` should be always sufficient.
2. `wxBitmapBundle::GetBitmap(size)` works, i.e. it returns the bitmap of size *and* scale factor compatible with physical size (so when you ask for 32x32 bitmap you can get back a 32x32 bitmap with scale factor 1 or 64x64 bitmap with scale factor 2), and so it is wrong to override the scale factor of the returned bitmap.

Giving this, I strongly suspect that the sizes of icons are not computed correctly.

Blake-Madden: OK, I reworked it, still looks correct on Retina and HiDPI Windows.

src/ribbon/buttonbar.cpp
```diff
@@ -944,19 +945,18 @@ void wxRibbonButtonBar::OnPaint(wxPaintEvent& WXUNUSED(evt))
...
+    // wxBitmapBundle::GetBitmap() expects a size in physical pixels, so convert
+    // the stored DIP sizes all the way to physical pixels (ToPhys() is a no-op
+    // under MSW, while FromDIP() is a no-op under the ports using DPI-independent
+    // pixels, so this gives the right result everywhere).
+    const wxSize scaledLarge = ToPhys(FromDIP(m_bitmap_size_large));
+    const wxSize scaledSmall = ToPhys(FromDIP(m_bitmap_size_small));
 
     for ( auto& button : layout->buttons )
     {
         wxRibbonButtonBarButtonBase* base = button.base;
```
vadz, line 957: As you can see from the [diagram here](https://docs.wxwidgets.org/latest/overview_high_dpi.html#high_dpi_pixel_conversions), the new code is exactly equivalent to the old one, so it shouldn't be necessary to change this.

vadz (approved):
Thanks, this looks good to me now, but I'd rather not apply the change that I commented below. If you don't see any problems with this, I can merge it.

src/ribbon/buttonbar.cpp
Blake-Madden: sure, sounds good. thanks.

PR #26411, opened 2026-04-27
wxOSX: Don’t crash in wxNativeFontInfo::FromString()

vslavik:
This PR fixes crash when deserializing font info from string representation. This happens if the font face is missing on the current system; I don’t know if it can happen under other circumstances too.

Split into two commits:

1. 9c18a70ab7aee205a7b3b0fc0ae8a3ba6cf260c2 — this one is clear, no code should crash, period. This results in *some* font being created, bug degraded (no face, fallback traits).

2. 9ad3f86964da17313fdda99dde2cd4324d39367f — treats this failure as deserialization failure, so that the missing font is detectable, but I’m not sure if that’s the intention there. E.g. wxGTK seems to accept such fonts unless I’m mistaken...

Applies to 3.2 as-is.

### DIFF
--- a/src/osx/carbon/font.cpp
+++ b/src/osx/carbon/font.cpp
@@ -887,9 +887,10 @@ void wxNativeFontInfo::InitFromFontDescriptor(CTFontDescriptorRef desc)
 
     // determine approximate family
 
-    CTFontSymbolicTraits symbolicTraits;
+    CTFontSymbolicTraits symbolicTraits = 0;
     wxCFDictionaryRef traits((CFDictionaryRef)CTFontDescriptorCopyAttribute(desc, kCTFontTraitsAttribute));
-    traits.GetValue(kCTFontSymbolicTrait).GetValue((int32_t*)&symbolicTraits, 0);
+    if (traits)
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, the crash definitely needs to be fixed but it looks like both wxMSW and wxGTK accept anything syntactically valid in `FromString()`, so I think it would be better to do the same thing in wxOSX. Unless we really, really need to detect unavailable face names (and do what in this case?)?

vslavik:
> and do what in this case?

I hit this code when loading user-customized font preference and just revert to the default (which won’t be the same as default-constructed wxFont, I suspect) on failure. But it’s such an extremely rare situation that all I *really* care about is not crashing there.

IOW, I’m perfectly fine with dropping the [9ad3f86](https://github.com/wxWidgets/wxWidgets/pull/26411/commits/9ad3f86964da17313fdda99dde2cd4324d39367f) part.

vadz:
I've applied the first commit to master and also update the documentation to explain how this works and will backport both commits to 3.2 too.

Thanks again!

Issue #26404, opened 2026-04-23
Dark mode colors do not match Windows 11


stevecor:
Some color values in wxDarkModeSettings::GetColour() do not match what I see with supplied apps in Windows 11 25H2, uncustomized dark mode.

To determine color values, I used the Snipping Tool to capture an actual color shown, pasted into Paint, used the "Color picker" tool, and clicked the "Edit colors" button to see the RGB value.

- wxSYS_COLOUR_INFOBK. In File Explorer, hover over a toolbar button. The tooltip background color is 0x2d2d2d, not 0x202020.
- Text colors. Text colors are pure white, not 0xe0e0e0. Before sampling, maximize text size by going to Settings: Accessibility: Text size and set to max. In File Explorer, all text is white. The Notepad menu text is white. To get a proper color, sample the center of horizontal or vertical letter strokes.
- wxSYS_COLOUR_HIGHLIGHT. In File Explorer, select an item. The background color is 0x505050, not 0x9e5315.

Proposed changes:

```
diff --git a/src/msw/darkmode.cpp b/src/msw/darkmode.cpp
index 9ff28b6fe8..39218e6ad5 100644
--- a/src/msw/darkmode.cpp
+++ b/src/msw/darkmode.cpp
@@ -312,11 +312,13 @@ wxColour wxDarkModeSettings::GetColour(wxSystemColour index)

         case wxSYS_COLOUR_ACTIVECAPTION:
         case wxSYS_COLOUR_APPWORKSPACE:
-        case wxSYS_COLOUR_INFOBK:
         case wxSYS_COLOUR_LISTBOX:
         case wxSYS_COLOUR_WINDOW:
             return wxColour(0x202020);

+        case wxSYS_COLOUR_INFOBK:
+            return wxColour(0x2d2d2d);
+
         case wxSYS_COLOUR_BTNTEXT:
         case wxSYS_COLOUR_CAPTIONTEXT:
         case wxSYS_COLOUR_HIGHLIGHTTEXT:
@@ -325,7 +327,7 @@ wxColour wxDarkModeSettings::GetColour(wxSystemColour index)
         case wxSYS_COLOUR_LISTBOXTEXT:
         case wxSYS_COLOUR_MENUTEXT:
         case wxSYS_COLOUR_WINDOWTEXT:
-            return wxColour(0xe0e0e0);
+            return *wxWHITE;

         case wxSYS_COLOUR_HOTLIGHT:
             return wxColour(0xe48435);
[...]

PR #26412, opened 2026-04-28
Update some color values for Windows dark mode

stevecor:
Update some color values in wxDarkModeSettings::GetColour() to better match the latest Windows version. See #26404

### DIFF
--- a/interface/wx/settings.h
+++ b/interface/wx/settings.h
@@ -62,8 +62,8 @@ enum wxSystemColour
     wxSYS_COLOUR_ACTIVEBORDER,        //!< Active window border colour.
     wxSYS_COLOUR_INACTIVEBORDER,      //!< Inactive window border colour.
     wxSYS_COLOUR_APPWORKSPACE,        //!< Background colour for MDI applications.
-    wxSYS_COLOUR_HIGHLIGHT,           //!< Colour of item(s) selected in a control.
-    wxSYS_COLOUR_HIGHLIGHTTEXT,       //!< Colour of the text of item(s) selected in a control.
+    wxSYS_COLOUR_HIGHLIGHT,           //!< Background colour of a selected item or selected text in a control.
+    wxSYS_COLOUR_HIGHLIGHTTEXT,       //!< Colour of selected text.
     wxSYS_COLOUR_BTNFACE,             //!< Face shading colour on push buttons.
     wxSYS_COLOUR_BTNSHADOW,           //!< Edge shading colour on push buttons.
     wxSYS_COLOUR_GRAYTEXT,            //!< Colour of greyed (disabled) text.
--- a/src/msw/darkmode.cpp
+++ b/src/msw/darkmode.cpp
@@ -312,10 +312,13 @@ wxColour wxDarkModeSettings::GetColour(wxSystemColour index)
 
         case wxSYS_COLOUR_ACTIVECAPTION:
         case wxSYS_COLOUR_APPWORKSPACE:
-        case wxSYS_COLOUR_INFOBK:
         case wxSYS_COLOUR_LISTBOX:
         case wxSYS_COLOUR_WINDOW:
             return wxColour(0x202020);
+
+        // Tooltip background of File Explorer toolbar button and file, averaged
+        case wxSYS_COLOUR_INFOBK:
+            return wxColour(0x2a2a2a);
 
         case wxSYS_COLOUR_BTNTEXT:
         case wxSYS_COLOUR_CAPTIONTEXT:
@@ -325,7 +328,7 @@ wxColour wxDarkModeSettings::GetColour(wxSystemColour index)
         case wxSYS_COLOUR_LISTBOXTEXT:
         case wxSYS_COLOUR_MENUTEXT:
         case wxSYS_COLOUR_WINDOWTEXT:
-            return wxColour(0xe0e0e0);
+            return *wxWHITE;
 
         case wxSYS_COLOUR_HOTLIGHT:
             return wxColour(0xe48435);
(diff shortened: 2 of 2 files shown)


stevecor:
The check failure for ci/circleci: build-fedora is unrelated.

src/msw/darkmode.cpp
```diff
@@ -346,8 +350,8 @@ wxColour wxDarkModeSettings::GetColour(wxSystemColour index)
             return wxColour(0x626262);
 
         case wxSYS_COLOUR_HIGHLIGHT:
-        case wxSYS_COLOUR_MENUHILIGHT:
-            return wxColour(0x9e5315);
+            // Selected text background in File Open dialog
+            return wxColour(0xd47800);
```
vadz, line 354: Under Windows 10 it's `#0078d7` (i.e. difference of 3 for blue component) which is again annoyingly minor, but I'd rather take it into account than not.

src/msw/darkmode.cpp
```diff
@@ -312,10 +312,14 @@ wxColour wxDarkModeSettings::GetColour(wxSystemColour index)
...
-        case wxSYS_COLOUR_INFOBK:
         case wxSYS_COLOUR_LISTBOX:
         case wxSYS_COLOUR_WINDOW:
-            return wxColour(0x202020);
+            // File Explorer background
+            return wxColour(0x191919);
+
+        case wxSYS_COLOUR_INFOBK:
+            // Tooltip background of File Explorer toolbar button and file, averaged
+            return wxColour(0x2a2a2a);
```
vadz, line 322: This one is `#2b2b2b` for me under Windows 10 and while the difference is really minor, I think it makes sense to return exactly the same colour as the native Explorer uses just in case something is capable of noticing the difference.

src/msw/darkmode.cpp
```diff
@@ -312,10 +312,14 @@ wxColour wxDarkModeSettings::GetColour(wxSystemColour index)
 
         case wxSYS_COLOUR_ACTIVECAPTION:
         case wxSYS_COLOUR_APPWORKSPACE:
-        case wxSYS_COLOUR_INFOBK:
         case wxSYS_COLOUR_LISTBOX:
         case wxSYS_COLOUR_WINDOW:
-            return wxColour(0x202020);
+            // File Explorer background
+            return wxColour(0x191919);
```
vadz, line 318: I double checked and it's really `#202020` under Windows 10...

interface/wx/settings.h
```diff
@@ -62,8 +62,8 @@ enum wxSystemColour
     wxSYS_COLOUR_ACTIVEBORDER,        //!< Active window border colour.
     wxSYS_COLOUR_INACTIVEBORDER,      //!< Inactive window border colour.
     wxSYS_COLOUR_APPWORKSPACE,        //!< Background colour for MDI applications.
-    wxSYS_COLOUR_HIGHLIGHT,           //!< Colour of item(s) selected in a control.
-    wxSYS_COLOUR_HIGHLIGHTTEXT,       //!< Colour of the text of item(s) selected in a control.
+    wxSYS_COLOUR_HIGHLIGHT,           //!< Background colour of a selected item or selected text in a control.
```
vadz, line 65: The trouble is that these colours are completely different in Windows dark mode, so I'm really not sure it makes sense to update the documentation like this.

vadz (commented):
Thanks, I'm going to push the changes to the colours with Windows 10 adjustments, but I think we need to change how `HIGHLIGHT` is used instead of changing its documentation as it will still be wrong.

src/msw/darkmode.cpp
stevecor: I see in Windows 10, the right pane is 0x202020 but the left pane is 0x191919. It's ambiguous. In Windows 11, both panes are 0x191919. Windows 11 breaks the tie.

vadz: The trouble is that these colours are easily distinguishable, i.e. I do see the difference between the tree and the list in Windows 10 Explorer window and I don't have the best eye acuity (to put it mildly). We probably need to return different colours from `wx{Tree,List}Ctrl::GetClassDefaultAttributes()`...

stevecor: There really is no right or wrong color here. The background in Windows 11 Notepad is 0x272727. In Visual Studio 2026, it's 0x1c1c1c. The color is a preference. I think we should keep the code simple and just return a single value, not checking the Windows version. Whatever value you want is fine with me.

vadz: It's crazy that there is no standard colour for the window background. But let's keep the current code for now and see if anybody else complains.

src/msw/darkmode.cpp
stevecor: I made a mistake, let's keep 0x2b2b2b and delete the comment. I attempted to average two colors, one for the toolbar (A) and the other for hovering over a file (B). But I see now pressing the PrintScreen key causes (B) to fade out and I got an incorrect color value.

src/msw/darkmode.cpp
paulcor: > Under Windows 10

Windows 10 is EOL.

lanurmi: > > Under Windows 10
> 
> Windows 10 is EOL.

It is a well known fact that the wxWidgets project supports Microsoft products for at least 5 years longer than Microsoft itself does. Besides, Windows 10 is still within the easily obtainable extended support program.

src/msw/darkmode.cpp
stevecor: On Windows 11, the toolbar tooltip background is a dithered pattern with pixels ranging from 0x2b2b2b to 0x303030. Since we now have specific separate colors for Windows 10 and Windows 11, I think the Windows 11 color should be the average brightness of this pattern which I measured at 0x2e2e2e.

src/msw/darkmode.cpp
vadz: > > Under Windows 10
> 
> Windows 10 is EOL.

Yes, but many people still use it and it doesn't cost much to keep the code for it, so I don't see any real harm in doing it.

src/msw/darkmode.cpp
vadz: Thanks, done in dbce2cf0d1b0b73f194b9bf6bc8f080666d66abe

PR #26415, opened 2026-04-28
Support imperial units for page setup dialog

LegalizeAdulthood:
If you set the page size (`wxPageSize`) to an imperial page size before launching the dialog, the margin units will be inches not millimeters.


LegalizeAdulthood:
This change only affects msw

vadz:
Thanks, will merge soon.

PR #26416, opened 2026-04-29
Remove redundant Visual Studio optimization settings

stevecor:
The three optimization settings listed below occur in all the files build/msw/*.vcxproj for the Release configurations, but are redundant because they are implied by the setting /O2 (`<Optimization>MaxSpeed</Optimization>`) which is also in all these files.

```
/Ob2    <InlineFunctionExpansion>AnySuitable</InlineFunctionExpansion>
/Oi     <IntrinsicFunctions>true</IntrinsicFunctions>
/Ot     <FavorSizeOrSpeed>Speed</FavorSizeOrSpeed>
```

These three options interfere with changing the optimization from /O2 to /O1. In order to change from /O2 to /O1 requires making four changes instead of one.

These options and their relationship is documented at https://learn.microsoft.com/en-us/cpp/build/reference/o1-o2-minimize-size-maximize-speed.


vadz:
Looks good, thanks, will push soon.

PR #26417, opened 2026-04-29
Update Lexilla submodule to fix build with gcc 16 -Werror

vadz:
Just suppress a new warning in gcc 16.


vadz:
Oops, it turns out that even gcc 16 doesn't have `__has_warning` (for some reason I decided it had it, but it was (my own, can't even blame this one on a LLM) hallucination). We'll have to do it the hard way...

paulcor:
Since you are modifying the source anyway, is there a reason you don't want to just fix the problem by removing the unused variable, which would work for any compiler, rather than trying to suppress the warning in a somewhat fragile way?

vadz:
The warning was suppressed like this upstream, I'm trying to avoid changing the code itself because this will almost surely result in conflicts later. The changes here may well result in conflicts too, but should hopefully be simpler to resolve.

vadz:
Replaced by #26418 as there are warnings not just in Lexilla.

Issue #26403, opened 2026-04-23
wxQt build does not honor --enable-privatefonts


AtesComp:
### Build System Used

I build wxWidgets and/or my application using:

- [X] CMake
- [X] configure
- [ ] mingw32-make with makefile.gcc
- [ ] MSBuild (Microsoft Visual Studio solution file)
- [ ] nmake with makefile.vc
- [ ] Xcode

### Description
On Linux, when building `wxQt`, the `configure` script does not honor `--enable-privatefonts`.

### Platform and version information

- wxWidgets version you are building: 3.3.3
- wxWidgets port you are building: wxQt
- OS and its version: Kubuntu 25.10
- Compiler being used: gcc 
- Non-default compiler options, if any: none

### Start for Possible Fix:

Edit the "configure" script:
* After the first `if test "$wxUSE_PRIVATE_FONTS" = "yes"; then` line:
  Change line `if test "$wxUSE_GTK" = 1; then`
  to `if test "$wxUSE_GTK" = 1 -o "$wxUSE_QT" = 1; then`

Edit "include/wx/chkconf.h":
* Change `#   if !defined(__WXMSW__) && !defined(__WXGTK__) && !defined(__WXOSX__)`
   to `#   if !defined(__WXMSW__) && !defined(__WXGTK__) && !defined(__WXQT__) && !defined(__WXOSX__)`

I've tested a build with the above changes and it complete without error. However, its missing all the Pango hookups present in wxGtk:
* font.cpp
* font.h
* private.h and
* the private subdirectory code

so the calls are undefined in wxQt.

AtesComp:
The wxQt arch should likely directly use:
`QFontDatabase.addApplicationFont(const QString &)`
as a call in its to be created `wxFontBase::AddPrivateFont(const wxString &)` wrapper.

`QFontDatabase.addApplicationFont(const QString &)` was introduced in Qt 4.2.

This mitigates replicating the wxGtk code using Pango.

vadz:
Support for this is indeed not implemented in wxQt, please consider [making a PR](https://www.wxwidgets.org/develop/how-to-submit-patches/) if you know how to implement it.

TIA!

AtesComp:
Yes, I'm looking into doing a PR.

[...]

PR #26419, opened 2026-04-30
wxQt Private Font

AtesComp:
Add wxQt support for --enable-privatefonts (wxUSE_PRIVATE_FONTS)
* Add Qt based wxFontBase AddPrivateFont()
* Update configuration files for automated wxQt Private Fonts
  * configure.ac
  * chkconfig.h
  * options.cmake


AtesComp:
Closes #26403

include/wx/qt/font.h
```diff
@@ -55,11 +55,12 @@ class WXDLLIMPEXP_CORE wxFont : public wxFontBase
     virtual wxSize GetPixelSize() const override;
     virtual wxFontStyle GetStyle() const override;
     virtual int GetNumericWeight() const override;
-    virtual bool GetUnderlined() const override;
     virtual wxString GetFaceName() const override;
+    virtual bool GetUnderlined() const override;
+    virtual bool GetStrikethrough() const override;
```
vadz: It's preferable to avoid mixing insignificant changes like this one (just moving the function declarations) with the real changes. If you really want to move these functions, you can do it in a separate commit/PR, although some explanation about why they should be moved should be provided even in this case.

include/wx/qt/font.h
```diff
@@ -55,11 +55,12 @@ class WXDLLIMPEXP_CORE wxFont : public wxFontBase
...
     virtual wxFontEncoding GetEncoding() const override;
     virtual const wxNativeFontInfo *GetNativeFontInfo() const override;
-    virtual bool GetStrikethrough() const override;
+    virtual bool IsFixedWidth() const override;
```
vadz: This doesn't seem to be needed at all, why add it when it just calls the base class version?

src/qt/font.cpp
```diff
@@ -1,22 +1,32 @@
 /////////////////////////////////////////////////////////////////////////////
 // Name:        src/qt/font.cpp
-// Author:      Peter Most, Mariano Reingart, Javier Torres
+// Purpose:     wxFont for wxQt
+// Author:      Peter Most, Mariano Reingart, Javier Torres, Keven Ates
```
vadz: We can keep this change but we typically avoid modifying these comments with every modification of the file, `git log` or `git shortlog` already provides the information about the authors of the file.

src/qt/font.cpp
```diff
@@ -42,116 +52,120 @@ static QFont::StyleHint ConvertFontFamily(wxFontFamily family)
...
-        case wxFONTFAMILY_SWISS:
-            return QFont::SansSerif;
-
-        case wxFONTFAMILY_MODERN:
-            return QFont::TypeWriter;
-
-        case wxFONTFAMILY_TELETYPE:
-            return QFont::TypeWriter;
-
+        case wxFONTFAMILY_DEFAULT:      return QFont::AnyStyle;
```
vadz: Sorry but changes like this are really not welcome, especially when mixed with significant changes.

If you use automatic code formatter you need to turn it off for wx sources.

src/qt/font.cpp
```diff
@@ -42,116 +52,120 @@ static QFont::StyleHint ConvertFontFamily(wxFontFamily family)
...
     return QFont::AnyStyle;
 }
 
-// Helper of ConvertFontWeight() and GetNumericWeight(): if a value lies in
-// ]fromMin, fromMax] interval, then map it to [toMin, toMax] interval linearly
-// and return true, otherwise return false and don't modify it.
+// Helper of ConvertFontWeight() and GetNumericWeight()
+//      If a value lies in (fromMin, fromMax] interval, then
+//          map it to [toMin, toMax] interval linearly and return true
+//      Otherwise don't modify it and return false.
```
vadz, line 85: Same as above.

src/qt/font.cpp
```diff
@@ -42,116 +52,120 @@ static QFont::StyleHint ConvertFontFamily(wxFontFamily family)
...
+                     wxQFont_ExtraBold,  wxQFont_Black) ||
+         TryToMap(w, wxFONTWEIGHT_HEAVY,      wxFONTWEIGHT_EXTRAHEAVY,
+                     wxQFont_Black,      99) )
     {
         return w;
     }
-
     wxFAIL_MSG("invalid wxFont weight");
-
     return wxQFont_Normal;
```
vadz: Ditto.

src/qt/font.cpp
```diff
@@ -42,116 +52,120 @@ static QFont::StyleHint ConvertFontFamily(wxFontFamily family)
...
-            m_nativeFontInfo.SetFaceName(info.GetFaceName());
-        else
-            m_nativeFontInfo.SetFamily(info.GetFamily());
-
-        if ( info.IsUsingSizeInPixels() )
-            m_nativeFontInfo.SetPixelSize(info.GetPixelSize());
-        else
-            m_nativeFontInfo.SetSizeOrDefault(info.GetFractionalPointSize());
+    // main ctor
+    wxFontRefData(const wxFontInfo& info);
```
vadz: Again, this doesn't seem to be related to the topic of this PR.

src/qt/font.cpp
```diff
@@ -604,22 +630,48 @@ void wxNativeFontInfo::SetEncoding(wxFontEncoding WXUNUSED(encoding))
...
 {
-    return FromString(s);
+    return FromString(strUserFontInfo);
 }
 
 wxString wxNativeFontInfo::ToUserString() const
 {
     return ToString();
 }
+
```
vadz, line 652: This is the only change to this file which seems really needed, right?

src/qt/font.cpp
```diff
@@ -604,22 +630,48 @@ void wxNativeFontInfo::SetEncoding(wxFontEncoding WXUNUSED(encoding))
...
+#include "wx/fontenum.h"
+
+#include <QtGui/QFontDatabase>
+
+bool wxFontBase::AddPrivateFont(const wxString& strFilename)
+{
+    int iFontID =
+        QFontDatabase::addApplicationFont( wxQtConvertString(strFilename) );
+    if (iFontID == -1) return false;
```
vadz, line 667: While reformatting existing code should be turned off, new code should ideally conform to wx coding style which notably prohibits single line compound statements, i.e. this should be


```suggestion
    if (iFontID == -1)
        return false;
```

configure
```diff
@@ -39770,7 +39770,7 @@ $as_echo "yes" >&6; }
 
 fi
         fi
-    elif test "$wxUSE_MAC" = 1 -o "$wxUSE_MSW" = 1; then
+    elif test "$wxUSE_QT" = 1 -o "$wxUSE_MAC" = 1 -o "$wxUSE_MSW" = 1; then
```
vadz, line 39773: This file is generated and shouldn't be modified directly, please modify `configure.ac` and follow the instructions in `build/tools/autoconf/README.md` for updating configure itself.

vadz (commented):
Thanks for your contribution, but in its current state I would commit just the last change to `font.cpp` and the change to `wx/chkconf.h`.

I can update `configure.ac` and `configure` myself when applying, but it would be worth checking CMake build system too, please do it if you can.

include/wx/qt/font.h
AtesComp: I compared to the gtk version and made these kind of changes for parity. It confusing to see similar code vary when trying to add consistency.

include/wx/qt/font.h
AtesComp: IsFixedWidth() was missing. Again, parity with gtk.

src/qt/font.cpp
AtesComp: Again, parity with gtk. Log are what that are, but coders read code. My name does not need to be there if it is truly an issue.

src/qt/font.cpp
AtesComp: I was looking for readability when comparing to gtk, so the condensed code helped. It can be removed.

src/qt/font.cpp
AtesComp: The original text was not clear and it's "math" symbolism was incorrect.

src/qt/font.cpp
AtesComp: It can be removed.

src/qt/font.cpp
AtesComp: Again, added parity with gtk.

src/qt/font.cpp
AtesComp: No, also parity and other added method.

src/qt/font.cpp
AtesComp: Easily changed.

configure
AtesComp: Will do.

include/wx/qt/font.h
vadz: I don't think "parity" is that important, but if you think it is, you can make such changes except that:

1. Please make them in separate commits/PRs clearly stating that they don't make any real changes.
2. In this particular case it is obviously better to remove the unnecessary function from wxGTK rather than add one to wxQt.

AtesComp: Reviewing Qt, I should change it to use the native checker with:
`return QFontInfo(font).fixedPitch()`
I should likely check all the other related methods and make them native as well.

I'll back this out and make those changes in another PR.

configure
MaartenBent: To enable it by default when using CMake, you can change
https://github.com/wxWidgets/wxWidgets/blob/52f9219bf4b730d382f7221896f6ff2b76f7ab4f/build/cmake/options.cmake#L473
to
```cmake
if(WIN32 OR APPLE OR WXQT)
```

AtesComp: Thanks! I'll add it to the updated PR.

AtesComp: I made the change to the `configure.ac`. I think I understand why the Docker image process is use to generate a new `configure` script. How does the `configure` script get pulled from the Docker image to the repository?

vadz: The "Using Pre-Made Image" section has the command for updating `configure`.

AtesComp: Got it working. It wasn't entirely clear that the docker image was pushing back the `configure` script. I see it is somewhat implied in the text.

src/qt/font.cpp
AtesComp: This change includes the `wxFontRefData(const wxString& strNativeFontInfo)` that pulls the font info from a native Qt font string. I'll revert and add to a new PR for native Qt font management.

include/wx/qt/font.h
AtesComp: Reverted.

include/wx/qt/font.h
AtesComp: Reverted.

src/qt/font.cpp
AtesComp: Reverted.

src/qt/font.cpp
AtesComp: I left the `Purpose` statement as it is uniform with other files.

src/qt/font.cpp
AtesComp: I left this change to correct the "math" symbolism and make the comment readable as pseudocode.

src/qt/font.cpp
AtesComp: Reverted.

src/qt/font.cpp
AtesComp: Reverted.

src/qt/font.cpp
AtesComp: This is the main change for this PR.

src/qt/font.cpp
AtesComp: Reverted.

src/qt/font.cpp
vadz: Sorry, "incorrect" is in the eye of beholder, `]` is a perfectly cromulent way of denoting an open interval end in many countries. In any case, I'd really like to emphasize that while such changes could be done, in principle, they *must* be done in separate commits and not intermixed with significant changes. So I'll omit this one when merging.

PR #26420, opened 2026-05-01
Add system color change event handler to dataview control

stevecor:
Add wxDataViewCtrl::OnSysColourChanged() so that wxDataViewCtrl redraws properly after system colors change on Windows. This works when changing high contrast themes and dark mode themes.

### DIFF
--- a/include/wx/generic/dataview.h
+++ b/include/wx/generic/dataview.h
@@ -362,6 +362,8 @@ public:     // utility functions not part of the API
 
     virtual void OnInternalIdle() override;
 
+    void OnSysColourChanged(wxSysColourChangedEvent& event);
+
 #if wxUSE_ACCESSIBILITY
     virtual wxAccessible* CreateAccessible() override;
 #endif // wxUSE_ACCESSIBILITY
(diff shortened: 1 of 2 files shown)


vadz (approved):
Thanks, will merge soon.

paulcor:
I'm curious, does MSW really not arrange for all windows to be repainted when the theme changes? If not, then maybe we should implement a more general fix for this. It really seems like it shouldn't be up to every single window to individually have to say "yeah, repaint me when the colors change, duh".

vadz:
MSW doesn't do it, to the best of my knowledge, it just sends `WM_SYSCOLORCHANGE` or `WM_SETTINGCHANGE` to all TLWs and I don't think `DefWindowProc()` does anything when getting them. We could, of course, call `Refresh()` by default in wxMSW if the resulting `wxSysColourChangedEvent` is not handled, but for most controls this is not going to be enough because they typically cache colours/pens/brushes and need to recreate them.

Still, I don't see any real drawbacks in refreshing by default, so let's do it?

Issue #25755, opened 2025-09-02
MSW: Window background colors are overridden by default/don't update with dark mode switch


willco007:
#### Bug description:

On Windows, `wxFrame` is setting the background color of windows to `wxSYS_COLOUR_APPWORKSPACE` on Create() [here](https://github.com/wxWidgets/wxWidgets/blob/170c38631458dcef648c31b9d98dcc20a8c57f9c/src/msw/frame.cpp#L126).  This is overriding Windows default color OOB, I wonder why this is being done. To remove it I call `SetBackgroundColour(wxNullColor)` after every wxFrame creation.

Secondly, when toggling Dark/Light mode on Windows, the window background isn't being updated as expected with or without a background color set.

#### Expected vs observed behaviour:

1) windows background color is wrong OOB on Windows
2) when switching from dark to light mode, the background of wxFrame isn't updated

#### To Reproduce:

Create a default wxFrame window on Windows and show it/toggle dark/light mode

### Platform and version information

- wxWidgets version you use: 3.3.1
- wxWidgets port you use: wxMSW
- OS and its version: Windows 11

vadz:
I think the default background colour is correct (for the light mode). And in any case, this is what it was for ~30 years and so I really don't think we should change it.

However we should update it when the mode changes, of course. But please note that currently this is not at all the only thing that is not updated and some of them are rather difficult to update dynamically.

willco007:
I'm pretty sure the background color is incorrect in light mode - or this particular color is no longer used. It's a dark gray while the various OS windows are light gray. If you clear the color (using SetBackgroundColour(wxNullColor)) you get the default background color that matches the OS.

willco007:
Here are a couple screenshots comparing the File Explorer Properties dialog (a win32 app) to the WX default color (disregard the incorrect UI elements).

<img width="1338" height="382" alt="Image" src="https://github.com/user-attachments/assets/04ba5dd9-5a93-4caa-aba6-7171a4eb49ca" />

<img width="1414" height="498" alt="Image" src="https://github.com/user-attachments/assets/dc054f3d-ade6-4532-aedf-8b84041b4020" />

vadz:
The default background colour is what Windows uses (used?) for document-oriented applications, not the dialogs — which use a different colour by default (and always did).

willco007:
I figured that is the reason it is dark. However, shouldn't it only be using that color if you're making a MDI parent frame, not a regular wxFrame? 

[...]

PR #26421, opened 2026-05-01
MSW: Fix wxFrame::OnSysColourChanged()

stevecor:
This change fixes wxFrame so that it properly responds to switching between high contrast themes as well as dark mode. See #25755 

A new member is added to to wxWindowBase named m_hasUserBgCol to track whether the background was set by the user. This serves the purpose that m_hasBgCol apparently once did. Currently, m_hasBgCol is set whenever the background color is set, even by SetOwnBackgroundColour(). Therefore m_hasBgCol cannot determine whether the background color should be changed in OnSysColourChanged().

### DIFF
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -2457,10 +2457,13 @@ void wxAuiManager::LayoutAddPane(wxSizer* cont,
         }
     }
 
-    if (min_size != wxDefaultSize)
-    {
-        sizer_item->SetMinSize(min_size);
-    }
+    // We need to reset any previous set min size to allow decreasing the pane
+    // size by dragging the sash between it and other panes, so always set it
+    // to something, even if it's not specified.
+    if (min_size == wxDefaultSize)
+        min_size = wxSize(1, 1);
+
+    sizer_item->SetMinSize(min_size);
 
 
     // add the vertical sizer (caption, pane window) to the


stevecor:
Check failures are unrelated

vadz:
We really, really shouldn't add `m_hasUserBgCol` because, as you say, this is exactly what `m_hasBgCol` is already supposed to do: it should be set it if, and only if, the window uses a custom background colour.

I didn't have time to look at why/how exactly did this get broken, but if you already do know it, please share it and/or make a PR fixing it. But, again, we shouldn't just work around the brokenness.

stevecor:
Changing the behavior of m_hasBgCol seems pretty risky. The current behavior was set 22 years ago, in f8ff87ed6da. This member is returned by a public function UseBgCol() whose documentation is consistent with the current behavior. Nearly all the references to these names in wxWidgets were introduced with the current behavior in place. The comment describing m_hasBgCol is now more wrong than the usage of this member.

I have not been able to come up with a fix to wxFrame::OnSysColourChanged() without adding a new member. Because the background color is always set by wxFrame::Create(), the decision tested in OnSysColourChanged() is always false.

```
void wxFrame::OnSysColourChanged(wxSysColourChangedEvent& event)
{
    // Don't override the colour explicitly set by the user, if any.
    if ( !UseBgCol() )
    {
        SetOwnBackgroundColour(wxSystemSettings::GetColour(wxSYS_COLOUR_APPWORKSPACE));

```
The wxFrame class needs a way to detect whether the background color changed after wxFrame::Create(). One alternative is to add a new member in wxFrame rather than wxWindow, to keep the change more isolated to where it is needed.

vadz:
> Because the background color is always set by wxFrame::Create(), the decision tested in OnSysColourChanged() is always false.

I think this is the part which is wrong. We shouldn't call `SetOwnBackgroundColour()` in `Create()` but should override `GetDefaultAttributes()` to return the right colour.

Am I missing some reason this wouldn't work (I didn't test it yet)?

PR #26422, opened 2026-05-02
Fix setting wxListCtrl column widths from wxEVT_SIZE in wxMSW

### LINKED ISSUE #26394: wxListCtrl items become invisible after calling SetColumnWidth and resizing.
### Description

1. Add the following to `listtest.cpp`:
```cpp
        m_listCtrl->Bind(wxEVT_SIZE,
            [this](wxSizeEvent& aEvt)
            {
                const int width = std::max(m_listCtrl->GetClientSize().GetWidth() - 2, 20);
                m_listCtrl->SetColumnWidth(0, width);

                aEvt.Skip();
            });
```

2. Decrease the frame's height until the vertical scrollbar appears
3. Scroll down the list
4. Increase the frame's height until the vertical scrollbar disappears
5. The rendering is now broken (items only redraw when hovered)

https://github.com/user-attachments/assets/c8e8eff0-f9b1-4dcd-a987-2c0595b6ca91

- wxWidgets version  you use: wx 3.3.3 (master)
- OS  and its version: Windows 11 Pro 25H2

---

KiCad report: https://gitlab.com/kicad/code/kicad/-/work_items/23986

vadz:
Calling SetColumnWidth() from wxEVT_SIZE handler could break the native control refresh logic and result in it not being refreshed at all any more.

Work around this by delaying actually setting the column width until the resizing is done.

Closes #26394.

### DIFF
--- a/src/msw/listctrl.cpp
+++ b/src/msw/listctrl.cpp
@@ -845,6 +845,24 @@ bool wxListCtrl::SetColumnWidth(int col, int width)
     else if ( width == wxLIST_AUTOSIZE_USEHEADER)
         width = LVSCW_AUTOSIZE_USEHEADER;
 
+    if ( m_inResize )
+    {
+        // Changing column width while handling WM_SIZE seems to be break
+        // something inside the native control, resulting in it not redrawing
+        // at all any longer in some circumstances, see #26394, so defer the
+        // call until resizing is done.
+        CallAfter([this, col, width]()
+        {
+            // This is a recursive call but it shouldn't recurse for ever (or,
+            // normally, more than once) as m_inResize should be reset "soon".
+            SetColumnWidth(col, width);
+        });
+
+        // Optimistically assume setting the column width later will succeed:
+        // better this than returning a bogus failure.
+        return true;
+    }
+
     if ( !ListView_SetColumnWidth(GetHwnd(), col, width) )
         return false;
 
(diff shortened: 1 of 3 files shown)


vadz:
This can't be backported to 3.2 as is because it adds a new member variable, but it could be done with some extra effort — please let me know if we really need it there.

src/msw/listctrl.cpp
```diff
@@ -845,6 +845,24 @@ bool wxListCtrl::SetColumnWidth(int col, int width)
     else if ( width == wxLIST_AUTOSIZE_USEHEADER)
         width = LVSCW_AUTOSIZE_USEHEADER;
 
+    if ( m_inResize )
+    {
+        // Changing column width while handling WM_SIZE seems to be break
+        // something inside the native control, resulting in it not redrawing
+        // at all any longer in some circumstances, see #26394, so defer the
+        // call until resizing is done.
+        CallAfter([this, col, width]()
```
dsa-t, line 854: I think it might be expected that calling `GetColumnWidth` right after `SetColumnWidth` returns the same value?

dsa-t:
> This can't be backported to 3.2 as is because it adds a new member variable, but it could be done with some extra effort — please let me know if we really need it there.

That's ok, KiCad uses wx 3.3 on Windows anyway

src/msw/listctrl.cpp
vadz: Maybe, but OTOH would it really be good to lie about the actual column width in `GetColumnWidth()`? If the code uses it for anything (e.g. find in which column the mouse pointer currently is), it risks getting wrong results, which seems even worse.

I think I'm just going to document that `GetColumnWidth()` will eventually return the correct result but it might not happen immediately. What do you think?

PR #26424, opened 2026-05-04
MSW: Repaint all windows upon setting change

stevecor:
Per feedback in #26420, this change adds a Refresh() into wxWindowMSW::OnSysColourChanged() so that all Windows are repainted upon theme change.

This change supersedes #26420. Those changes are no longer needed.

### DIFF
--- a/include/wx/generic/dataview.h
+++ b/include/wx/generic/dataview.h
@@ -362,8 +362,6 @@ public:     // utility functions not part of the API
 
     virtual void OnInternalIdle() override;
 
-    void OnSysColourChanged(wxSysColourChangedEvent& event);
-
 #if wxUSE_ACCESSIBILITY
     virtual wxAccessible* CreateAccessible() override;
 #endif // wxUSE_ACCESSIBILITY
--- a/src/generic/datavgen.cpp
+++ b/src/generic/datavgen.cpp
@@ -5683,8 +5683,6 @@ bool wxDataViewCtrl::Create(wxWindow *parent,
 
     EnableSystemThemeByDefault();
 
-    Bind(wxEVT_SYS_COLOUR_CHANGED, &wxDataViewCtrl::OnSysColourChanged, this);
-
 #if wxUSE_ACCESSIBILITY
     wxAccessible::NotifyEvent(wxACC_EVENT_OBJECT_CREATE, this, wxOBJID_CLIENT, wxACC_SELF);
 #endif // wxUSE_ACCESSIBILITY
@@ -6266,12 +6264,6 @@ void wxDataViewCtrl::OnInternalIdle()
 
     if ( m_colsDirty )
         UpdateColWidths();
-}
-
-void wxDataViewCtrl::OnSysColourChanged(wxSysColourChangedEvent& event)
-{
-    Refresh();
-    event.Skip();
 }
 
 int wxDataViewCtrl::GetColumnPosition( const wxDataViewColumn *column ) const
--- a/src/msw/mdi.cpp
+++ b/src/msw/mdi.cpp
@@ -522,10 +522,7 @@ void wxMDIParentFrame::OnIconized(wxIconizeEvent& event)
 void wxMDIParentFrame::OnSysColourChanged(wxSysColourChangedEvent& event)
 {
     if ( m_clientWindow )
-    {
         m_clientWindow->SetBackgroundColour(wxSystemSettings::GetColour(wxSYS_COLOUR_APPWORKSPACE));
-        m_clientWindow->Refresh();
-    }
 
     event.Skip();
 }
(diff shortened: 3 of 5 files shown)


paulcor:
A `wxEvent` handler can be blocked by a user handler not calling `Skip()`, I think it would be safer to do it something like the following, although maybe the current code should not be done from `EVT_SYS_COLOUR_CHANGED` either...

```diff
diff --git a/src/msw/window.cpp b/src/msw/window.cpp
index 9632bd1a1c..60fa39aa96 100644
--- a/src/msw/window.cpp
+++ b/src/msw/window.cpp
@@ -3548,6 +3548,8 @@ wxWindowMSW::MSWHandleMessage(WXLRESULT *result,
         case WM_SYSCOLORCHANGE:
             // the return value for this message is ignored
             processed = HandleSysColorChange();
+            if (IsTopLevel())
+                Refresh();
             break;
 
         case WM_DISPLAYCHANGE:
@@ -3564,6 +3566,8 @@ wxWindowMSW::MSWHandleMessage(WXLRESULT *result,
 
         case WM_SETTINGCHANGE:
             processed = HandleSettingChange(wParam, lParam);
+            if (IsTopLevel())
+                Refresh();
             break;
 
         case WM_QUERYNEWPALETTE:
```

stevecor:
I find that IsTopLevel() is always false in that context. I added a Refresh() upon WM_SETTINGCHANGE.

vadz:
I don't understand why would `IsTopLevel()` be always false, but then I don't understand why would we only want to call `Refresh()` when it's true, isn't the idea to refresh all the windows?

Although thinking more about it, we probably don't need to refresh the standard controls, i.e. those for which `IsOfStandardClass()` returns true, as those should already refresh themselves correctly — and if they don't, calling `Refresh()` on them is probably not going to help anyhow.

Also, it looks like in the case which motivated all this, i.e. switching between light and dark modes, we're going to refresh all the windows twice: first from `HandleSysColorChange()` and then from `HandleSettingChange()` itself. It's not the end of the world, but it would still be better to avoid this.

paulcor:
> but then I don't understand why would we only want to call `Refresh()` when it's true, isn't the idea to refresh all the windows?

`Refresh()` documentation says it refreshes all children recursively.

stevecor:
After experimenting changing between contrast and dark themes I find we only need to Refresh() upon WM_SETTINGCHANGE. Since Refresh() uses RDW_ALLCHILDREN, we only need to do the top level window.

vadz:
Sorry, I wanted to test this and related stuff a bit more, but never found time for it, hence the delay... I just wanted to ask, before returning to this (hopefully this weekend) if the changes in #26454 are related to this one or not?

stevecor:
> Sorry, I wanted to test this and related stuff a bit more, but never found time for it, hence the delay... I just wanted to ask, before returning to this (hopefully this weekend) if the changes in #26454 are related to this one or not?

Not related.

stevecor:
How to test the effect of this PR:

1. Add SetAppearance(Appearance::System) to the dataview sample.

```
diff --git a/samples/dataview/dataview.cpp b/samples/dataview/dataview.cpp
index b80c9270a1..c081da88f0 100644
--- a/samples/dataview/dataview.cpp
+++ b/samples/dataview/dataview.cpp
@@ -381,6 +381,7 @@ wxIMPLEMENT_APP(MyApp);

 bool MyApp::OnInit()
 {
+    SetAppearance(Appearance::System);
     if ( !wxApp::OnInit() )
         return false;

```
2. Comment out the code below and run the sample in light mode. That code is the fix.

```
diff --git a/src/msw/window.cpp b/src/msw/window.cpp
index e0ce689250..be093e682b 100644
--- a/src/msw/window.cpp
+++ b/src/msw/window.cpp
@@ -5236,8 +5236,8 @@ bool wxWindowMSW::HandleSettingChange(WXWPARAM wParam, WXLPARAM lParam)
         node = node->GetNext();
     }

-    if ( IsTopLevel() )
-        Refresh();
+//    if ( IsTopLevel() )
+//        Refresh();

     // let the system handle it
     return false;

```
3. Switch from light mode to dark mode. The only part of the app redrawn is in the non-client area, such as the menu bar.

4. Restore the fix that was commented out in step 2 and run again in light mode.

5. Switch from light mode to dark mode. You can see that a repaint occurs in the client area because some of the windows turn dark - the dataview control and the log window.

Notes:
a. This PR is not meant to completely address responding to dark/light theme change. It is simply a common sense improvement that happens to be enough for some controls to respond to color changes properly.

b. All the changes other than the two-line fix mentioned above are to remove Refresh() calls from individual system color change handlers that are no longer needed.

stevecor:

vadz:
Thanks and sorry again for the delay with applying this. I've now added commits from here to #26503, so I'm going to close this one — please let me know if you see anything wrong with the way I applied this (and your other PRs) in the one I opened.

Here I'd just like to say that it looks like we're going to be refreshing all windows completely unnecessarily in many cases (as mentioned in the comment I added), but I couldn't find a simple and reliable way to determine whether we need to refresh or not, so it still seems better to do it rather not to. I hope there are no system settings changing often enough for this to become a problem.

PR #26425, opened 2026-05-04
Update non-client colors upon theme change

stevecor:
This change updates the colors in the top level window title bar, menu bar, etc, with dark/light theme change. Upon the "ImmersiveColorSet" setting change, call GetDwmSetWindowAttribute() with DWMWA_USE_IMMERSIVE_DARK_MODE to update the new dark/light mode.

### DIFF
--- a/include/wx/msw/private/darkmode.h
+++ b/include/wx/msw/private/darkmode.h
@@ -20,8 +20,8 @@ namespace wxMSWDarkMode
 WXDLLIMPEXP_CORE
 bool IsActive();
 
-// Enable dark mode for the given TLW if appropriate.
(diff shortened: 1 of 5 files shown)


vadz (approved):
Thanks, this is a nice addition!

Other than minor style issues (we always put `{` on its own line...), I wonder if we shouldn't rename `EnableForTLW()` to something else, e.g. maybe `ConfigureTLW()`, because its current name seems a bit confusing if we can now use to turn dark mode support off.

Do you (or anybody else) have better suggestions for the name?

vadz:
Also part of #26503 now, thanks!

PR #26427, opened 2026-05-05
MSW: Fix wxFrame system color change

stevecor:
I accidentally closed #26421, this continues that work.

These changes fix the wxFrame system color change by setting colors in an overridden GetDefaultAttributes() rather than by the Create() calling SetOwnBackgroundColour().

This required deleting a check on m_hasBgCol that was preventing erasure of the background using our default color.

### DIFF
--- a/src/msw/window.cpp
+++ b/src/msw/window.cpp
@@ -5604,16 +5604,8 @@ wxWindowMSW::MSWGetBgBrushForChild(WXHDC hDC, wxWindowMSW *child)
         return hbrush;
     }
 
-    // Otherwise see if we have a custom background colour.
-    if ( m_hasBgCol )
-    {
-        wxBrush *
-            brush = wxTheBrushList->FindOrCreateBrush(GetBackgroundColour());
-
-        return (WXHBRUSH)GetHbrushOf(*brush);
-    }
-
-    return 0;
+    wxBrush *brush = wxTheBrushList->FindOrCreateBrush(GetBackgroundColour());
+    return (WXHBRUSH)GetHbrushOf(*brush);
 }
 
 WXHBRUSH wxWindowMSW::MSWGetBgBrush(WXHDC hDC)
--- a/include/wx/msw/frame.h
+++ b/include/wx/msw/frame.h
@@ -150,6 +150,8 @@ protected:
 
     virtual bool IsMDIChild() const { return false; }
 
+    virtual wxVisualAttributes GetDefaultAttributes() const override;
+
     // get default (wxWidgets) icon for the frame
     virtual WXHICON GetDefaultIcon() const;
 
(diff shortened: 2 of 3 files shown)


lanurmi:
> I accidentally closed #26421, this continues that work.

Does the `Reopen pull request` button not do what one would expect?

stevecor:
> > I accidentally closed #26421, this continues that work.
> 
> Does the `Reopen pull request` button not do what one would expect?

That button was disabled.

stevecor:

vadz:
This is a much better way of doing this, thanks! Also part of #26503 now, so closing.

src/msw/window.cpp
```diff
@@ -5604,16 +5604,8 @@ wxWindowMSW::MSWGetBgBrushForChild(WXHDC hDC, wxWindowMSW *child)
...
-    if ( m_hasBgCol )
-    {
-        wxBrush *
-            brush = wxTheBrushList->FindOrCreateBrush(GetBackgroundColour());
-
-        return (WXHBRUSH)GetHbrushOf(*brush);
-    }
-
-    return 0;
+    wxBrush *brush = wxTheBrushList->FindOrCreateBrush(GetBackgroundColour());
```
vadz, line 5607: Sorry, this change is wrong and I had to revert it in #26503 because it always used the parent window background, breaking the background of partially transparent controls.

Can you please explain why was it necessary to do this and what problem does reverting this cause?

stevecor: If wxFrame does not set its own background color, the member m_hasBgCol is always false, which prevents this function from returning a brush with the proper color. Without this change, the wxFrame background erasing is broken by the other changes. So please remove all the changes for this PR from #26503 . This will need some reconsideration. I also realize now there is a similar problem with MDI frames that ought to be addressed at the same time. This should remain a separate PR.

vadz: Do you have a reproducer for the problem of frame background erasing? I do like the change here, so I'd like to make this work.

stevecor: Run this program:

```
#include <wx/wx.h>
class App : public wxApp {
public:
    virtual bool OnInit();
};
wxIMPLEMENT_APP(App);
bool App::OnInit() {
    auto *frame = new wxFrame(nullptr, wxID_ANY, "App");
    frame->Show(true);
    return true;
}
```
The intended background is wxSYS_COLOUR_APPWORKSPACE which is dark gray. The actual background is light gray (wxSYS_COLOUR_BTNFACE?).

image
image

vadz: The simplest possible fix I see is 7e89064439 (Fix frame background colour after the parent commit, 2026-05-24) which I've just pushed to #26503. Do you see anything wrong with it?

stevecor: Looks good to me. I think this is OK for all top level windows, including dialogs.

PR #26429, opened 2026-05-06
Add click event to tree ctrl

szpeak:
In many scenarios, the event of left clicking on a tree item  is required. This modification:
1. Defined the wxEVT_TREE_ITEM_LEFT_CLICK event;
2. Added a few lines of code in the mouse event handler to issue a wxEVT_TREE_ITEM_LEFT_CLICK event for the situation where the previous left button press and the current left button release result in the same item for both events.


vadz:
Sorry, I don't know what happened here but something went very wrong here, there are unrelated changes and commits here and also, I think, missing changes because the code can't compile in this state. Please rebase your branch on the latest master, squash the commits that shouldn't be preserved in the final version and force push.

Also, you will need to modify the documentation (files under `interface`) and, in particular, explain how handling this event interacts with selection, e.g. whether handling it (without skipping) prevents the item being clicked from being selected.

PR #26430, opened 2026-05-06
Fix text rendering in RTL layout when using Direct2D backend

AliKet:
**Before:**
Direct2D_RTL_Bad

**After:**
Direct2D_RTL_Ok


vadz:
Thanks for fixing this!

PR #26431, opened 2026-05-07
Upgrade to Doxygen 1.15.0

### LINKED ISSUE #25603: Many Doxygen warnings in `interface/wx/stc/stc.h`
There are currently 138 warnings of the form

```
interface/wx/stc/stc.h:5439: warning: unable to resolve link to 'wxStyledTextCtrl::wxSTC_ANNOTATION_HIDDEN' for \link command
```

in this file and I'm not sure what to do about them, so for now I'm just going to ignore them when reporting warnings in `docs/doxygen/regen.sh` as otherwise any new warnings go unnoticed. But it would be nice to actually fix this, of course.

tierra:
While Doxygen 1.17.0 is the latest release, I've chosen to use 1.15.0 here for similar reasons we've stuck with older releases before. Doxygen 1.15.0 is the version bundled with Ubuntu 26.04 LTS, and also the current version slated for Debian Forky (maybe that changes before release though?). It is also the latest version of Doxygen that I have any kind of confirmation that Doxygen Awesome CSS is updated to be compatible with (changes in newer Doxygen may break DA styles in DA's latest release used here).

The GitHub docs update workflow manually installs Doxygen 1.15.0 from it's binary release hosted on GitHub, but only because there isn't a `ubuntu-26.04` GitHub Actions image yet. Once that's out, we can switch back to using the official distro package again. I still haven't tested the workflow changes, not sure how to do that without merging, but will followup with fixes if necessary.

Since we're using Doxygen Awesome, even though Doxygen has wildly changed the default style of docs in 1.14.0 (though I still don't really like their new style, DA is still a big improvement), our docs still look nearly identical with this update.

Other changes as detailed in commit:
- Upgrade Doxyfile configuration for 1.15.0
- Upgrade Doxygen Awesome to v2.4.2 (compatible up to Doxygen 1.15.0)
- Upgrade GitHub workflow Doxygen version to 1.15.0
- Tweak main navigation button spacing to match previous style
- Fix search box icon styles
- Fix class member definition group styles (missed by DA)
- Replace backticks used as apostrophe/quote in docs
- Fix remaining Doxygen warnings when processed with 1.15.0
- Bump required Doxygen version to 1.15.0

Closes: #25603

### DIFF
--- a/.github/workflows/docs_update.yml
+++ b/.github/workflows/docs_update.yml
@@ -23,7 +23,7 @@ permissions:
 
 jobs:
   update:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     name: Update Online Documentation
 
     steps:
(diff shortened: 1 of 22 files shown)


tierra:
> I still haven't tested the workflow changes, not sure how to do that without merging, but will followup with fixes if necessary.

Nevermind, I see that the job was run with this PR (without uploading), and appears to be working as expected.

vadz:
Thanks, I've tested it locally and it looks good, so merging.

PR #26432, opened 2026-05-08
Add left click to treectrl

szpeak:
In many scenarios, the event of left clicking on a tree item is required. This modification:

Defined the wxEVT_TREE_ITEM_LEFT_CLICK event;
Added a few lines of code in the mouse event handler to issue a wxEVT_TREE_ITEM_LEFT_CLICK event for the situation where the previous left button press and the current left button release result in the same item for both events.


src/generic/treectlg.cpp
```diff
@@ -3593,6 +3595,23 @@ void wxGenericTreeCtrl::OnMouse( wxMouseEvent &event )
...
+    }
+    else
+    {
+        if (event.LeftUp() && m_isLastMouseEventLeftDown)
+        {
+            m_isLastMouseEventLeftDown = false;
+            if (underMouse != nullptr && underMouse == m_underMouse)
+            {
+                wxTreeEvent* pEvent = new wxTreeEvent(wxEVT_TREE_ITEM_LEFT_CLICK, this, m_underMouse);
+                GetEventHandler()->QueueEvent(pEvent);
```
vadz, line 3610: Why do you do it like this instead of just calling `ProcessWindowEvent()`?

src/generic/treectlg.cpp
```diff
@@ -3593,6 +3595,23 @@ void wxGenericTreeCtrl::OnMouse( wxMouseEvent &event )
...
     }
+    if (event.LeftDown())
+    {
+        m_isLastMouseEventLeftDown = true;
+    }
+    else
+    {
+        if (event.LeftUp() && m_isLastMouseEventLeftDown)
+        {
+            m_isLastMouseEventLeftDown = false;
```
vadz, line 3606: Minor, but this could/should be done just right after `else` above instead of being done twice, here and below.

.gitignore
```diff
@@ -457,3 +457,4 @@
 
 /build/cmake/output/
 CMakeUserPresets.json
+/.vs
```
vadz, line 460: Sorry, this is unrelated to this PR, please remove it.

interface/wx/treectrl.h
```diff
@@ -142,6 +142,9 @@
     @event{EVT_TREE_ITEM_EXPANDING(id, func)}
           The item is being expanded. This can be prevented by calling Veto().
           Processes a @c wxEVT_TREE_ITEM_EXPANDING event type.
+    @event{EVT_TREE_ITEM_LEFT_CLICK(id, func)}
+          The user has clicked the item with the left mouse button.
+          Processes a @c wxEVT_TREE_ITEM_LEFT_CLICK event type.
```
vadz, line 147: Could you please add this:

```suggestion
          Processes a @c wxEVT_TREE_ITEM_LEFT_CLICK event type.
          This even is available since wxWidgets 3.3.3.
```

interface/wx/treectrl.h
```diff
@@ -142,6 +142,9 @@
     @event{EVT_TREE_ITEM_EXPANDING(id, func)}
           The item is being expanded. This can be prevented by calling Veto().
           Processes a @c wxEVT_TREE_ITEM_EXPANDING event type.
+    @event{EVT_TREE_ITEM_LEFT_CLICK(id, func)}
```
vadz, line 145: Please also add this in other places where other events are documented in this file, search for e.g. `EVT_TREE_ITEM_EXPANDING` to find them.

interface/wx/treectrl.h
```diff
@@ -142,6 +142,9 @@
...
+          The user has clicked the item with the left mouse button.
```
vadz, line 146: I think it would be also useful to mention that the item is always valid for the events of this type.

vadz (commented):
Thanks, this looks good but I don't understand why does the generic version send the event asynchronously, could you please explain or change this?

And also the documentation should be improved a bit more.

TIA!

PR #26434, opened 2026-05-08
Remove otherwise unused variables which only exist to be passed to GetTextExtent()

paulcor:


vadz:
FWIW as long as we're modifying these calls, I'd prefer to switch to the overload returning `wxSize` to avoid ugly output parameters.

PR #26436, opened 2026-05-09
Documentation - maximum one function is disconnected from an event handler

hwiesmann:
Calling Disconnect(int id = wxID_ANY, wxEventType eventType = wxEVT_NULL, wxObjectEventFunction function = nullptr,
                                 wxObject* userData = nullptr, wxEvtHandler* eventSink = nullptr) will only disconnect one function from the event handler even if several functions match the specified parameters.
This behaviour is now documented.

### DIFF
--- a/interface/wx/event.h
+++ b/interface/wx/event.h
@@ -840,6 +840,9 @@ public:
 
         This overload takes the additional @a id parameter.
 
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for improving this, I will merge it soon.

PR #26440, opened 2026-05-10
Avoid out-of-bounds palette write in wxIFFDecoder::ConvertToImage

### LINKED ISSUE #26437: Avoid out-of-bounds palette write in wxIFFDecoder::ConvertToImage on malformed BMHD transparent index
#### Bug description:

The IFF `BMHD` chunk's transparent-colour field is read as a 16-bit big-endian word and stored unclamped in `m_image->transparent`. In `wxIFFDecoder::ConvertToImage`, that value is used as a palette index without checking it against the CMAP-derived colour count, writing three bytes at offset `3 * transparent` past the palette buffer.

The unsafe site is `src/common/imagiff.cpp:161`:

```cpp
int transparent = GetTransparentColour();   // line 145, sourced from BMHD bmhd_transcol
...
if (transparent != -1)
{
    ...
    pal[3 * transparent + 0] = 255;          // line 161 — OOB write
    pal[3 * transparent + 1] = 0;
    pal[3 * transparent + 2] = 255;
}
```

`transparent` originates from `iff_getword` at `imagiff.cpp:403`, returning 0..65535. The write value is the fixed triplet `(255, 0, 255)`, so an attacker controls only the offset, but the offset is fully attacker-controlled.

Found by an ASAN+UBSAN libFuzzer harness against `wxIFFHandler::LoadFile`.

Loading an IFF whose BMHD `transcol` is outside `[0, colors)` should either skip the transparent-mask application or reject the file. Currently it performs a 3-byte heap-buffer-overflow write at an attacker-chosen offset.

#### Stack trace:

```
==1==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7b9699c2e430
WRITE of size 1 at 0x7b9699c2e430 thread T0
    #0 wxIFFDecoder::ConvertToImage(wxImage*) const  src/common/imagiff.cpp:161:34
    #1 wxIFFHandler::LoadFile(wxImage*, wxInputStream&, bool, int)  src/common/imagiff.cpp:759:17
    #2 LLVMFuzzerTestOneInput  tests/fuzz/iff.cpp:17:7
SUMMARY: AddressSanitizer: heap-buffer-overflow src/common/imagiff.cpp:161:34
```

Drop the following as `tests/fuzz/iff.cpp` next to the existing `tests/fuzz/runner.cpp`:

```cpp
#include "wx/log.h"
#include "wx/mstream.h"
#include "wx/imagiff.h"

{
    wxLogNull noLog;
    wxImage img;
    wxIFFHandler h;
    h.LoadFile(&img, mis, false);
    return 0;
}
```

50-byte PoC `F1-iff-oob-write.iff` (attach to the issue; hex below):

```
46 4f 52 4d 00 00 00 2e 49 4c 42 4d 42 4d 48 44
00 00 00 14 00 01 00 01 00 00 00 00 01 00 00 00
40 00 00 00 00 01 00 01 42 4f 44 59 00 00 00 02
00 00
```

The BMHD `transcol` field at offset 0x20 is `0x4000` (16384); the CMAP chunk is absent so `colors` stays at 0.

#### To Reproduce:

1. Check out master at the current HEAD and apply the harness above to `tests/fuzz/iff.cpp`.
2. Save the hex above as `F1-iff-oob-write.iff`. 
3. Build:

   ```
   clang++ -g -fsanitize=address,undefined tests/fuzz/{iff,runner}.cpp \
       `wx-config --cxxflags --libs core,base` -o repro_iff
   ```
4. Run `./repro_iff F1-iff-oob-write.iff` and observe the ASAN trace above.

### Suggested fix

Gate the palette write on `transparent < colors`:

```cpp
if (transparent >= 0 && transparent < colors)
{
    ...
    pal[3 * transparent + 0] = 255;
    ...
}
```

I have a patch ready with this fix and a regression `.iff` under `tests/image/`. Happy to send a PR.

- wxWidgets version: master at HEAD `f57b17ba5f63db53dd1b465646ade105fee22302`. The same line/pattern is also present at tag `v3.3.2` (line 161) and at `v3.2.0` (line 161).
- wxWidgets port: wxGTK (reproduced in the OSS-Fuzz container; the bug is in `wxBase`/`wxCore` parsing code and is port-independent).

MarkLee131:
Fix https://github.com/wxWidgets/wxWidgets/issues/26437:
The transparent colour index from the IFF BMHD chunk is a 16-bit value stored unclamped in m_image->transparent. Using it as a palette index without checking it against the CMAP-derived colour count writes 3 bytes at an attacker-controlled offset past the palette buffer.

Validate the index against the actual palette size before applying the magenta mask.

### DIFF
--- a/src/common/imagiff.cpp
+++ b/src/common/imagiff.cpp
@@ -146,7 +146,7 @@ bool wxIFFDecoder::ConvertToImage(wxImage *image) const
     long i;
 
     // set transparent colour mask
-    if (transparent != -1)
+    if (transparent >= 0 && transparent < colors)
     {
         for (i = 0; i < colors; i++)
         {
--- a/src/common/imagiff.cpp
+++ b/src/common/imagiff.cpp
@@ -146,8 +146,11 @@ bool wxIFFDecoder::ConvertToImage(wxImage *image) const
     long i;
 
     // set transparent colour mask
-    if (transparent >= 0 && transparent < colors)
+    if (transparent != -1)
     {
+        if (transparent < 0 || transparent >= colors)
+            return false;
+
         for (i = 0; i < colors; i++)
         {
             if ((pal[3 * i + 0] == 255) &&
--- a/tests/image/image.cpp
+++ b/tests/image/image.cpp
@@ -1363,6 +1363,31 @@ TEST_CASE_METHOD(ImageHandlersInit, "wxImage::BadXPM", "[image][xpm][error]")
 
 #endif // wxUSE_XPM
 
+#if wxUSE_IFF
+
+TEST_CASE_METHOD(ImageHandlersInit, "wxImage::BadIFF", "[image][iff][error]")
+{
+    // A FORM/ILBM file whose BMHD transparent-colour field is 0x4000 while
+    // no CMAP chunk is present, so the palette colour count stays at 0. The
+    // transparent index used to be applied to the palette without a bounds
+    // check, writing past the end of the palette buffer.
+    static const unsigned char data[] =
+    {
+        0x46,0x4f,0x52,0x4d,0x00,0x00,0x00,0x2e,
+        0x49,0x4c,0x42,0x4d,0x42,0x4d,0x48,0x44,
+        0x00,0x00,0x00,0x14,0x00,0x01,0x00,0x01,
+        0x00,0x00,0x00,0x00,0x01,0x00,0x00,0x00,
+        0x40,0x00,0x00,0x00,0x00,0x01,0x00,0x01,
+        0x42,0x4f,0x44,0x59,0x00,0x00,0x00,0x02,
+        0x00,0x00,
+    };
+    wxMemoryInputStream mis(data, WXSIZEOF(data));
+    wxImage img;
+    REQUIRE( !img.LoadFile(mis, wxBITMAP_TYPE_IFF) );
+}
+
+#endif // wxUSE_IFF
+
 TEST_CASE_METHOD(ImageHandlersInit, "wxImage::DibPadding", "[image]")
 {
     /*
--- a/tests/image/image.cpp
+++ b/tests/image/image.cpp
@@ -1368,9 +1368,10 @@ TEST_CASE_METHOD(ImageHandlersInit, "wxImage::BadXPM", "[image][xpm][error]")
 TEST_CASE_METHOD(ImageHandlersInit, "wxImage::BadIFF", "[image][iff][error]")
 {
     // A FORM/ILBM file whose BMHD transparent-colour field is 0x4000 while
-    // no CMAP chunk is present, so the palette colour count stays at 0. The
+    // the palette only has 2 entries (1 bitplane and no CMAP chunk). The
     // transparent index used to be applied to the palette without a bounds
-    // check, writing past the end of the palette buffer.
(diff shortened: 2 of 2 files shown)


vadz:
Thanks for the PR (and the detailed description of the bug in the corresponding issue)!

I'm not sure how does the test checking that loading an invalid image doesn't succeed works currently: don't we just ignore the invalid palette indices? I.e. could it be that loading fails but for unrelated reasons?

I don't know if it's better to ignore invalid values as this PR does or refuse to load the file using them outright. The best decision probably depends on the application, so maybe in the ideal case it should be a `wxImage` option for this handler, with the default being to reject invalid images but also a way of enabling it if necessary. However I'm saying this just in case you'd like to improve this further, I'm happy with just not writing beyond the end of the buffer and the only real question is about the test above.

Thanks again!

vadz:
In fact, the test does fail in some wxMSW builds (not sure why only some and not all), so there is definitely something wrong here. Should we apply

```diff
diff --git a/src/common/imagiff.cpp b/src/common/imagiff.cpp
index e7e8df8a80..fcc84b0c95 100644
--- a/src/common/imagiff.cpp
+++ b/src/common/imagiff.cpp
@@ -148,6 +148,9 @@ bool wxIFFDecoder::ConvertToImage(wxImage *image) const
     // set transparent colour mask
     if (transparent != -1)
     {
+        if (transparent < 0 || transparent >= colors)
+            return false;
+
         for (i = 0; i < colors; i++)
         {
             if ((pal[3 * i + 0] == 255) &&
```

to fix this instead?

vadz:
@MarkLee131 Will you look at the test failure/why does it fail or supposed to fail in the first place or should I?

MarkLee131:
Hi @vadz. Sorry, I am on leave recently. I can handle this issue next week if it is not late 😂　please let me know your preference

vadz:
I don't think IFF is so popular that fixing it is urgent, so next week would certainly be fine. Thanks in advance!

MarkLee131:
Hi @vadz, back from leave. Thanks again for the patch suggestion and for the patient timeline.

I've pushed a follow-up commit that applies exactly the diff you proposed: in wxIFFDecoder::ConvertToImage, when transparent != -1 but the index is out of bounds, we now return false instead of silently ignoring it.

I think this also explains the flaky wxMSW failures: with the previous "silently ignore" version, LoadFile would actually succeed on this malformed input, so the REQUIRE(!img.LoadFile(...)) only passed as a side effect of memory corruption from the unrelated overrun perturbing later allocation/state — which is exactly the "passing for unrelated reasons" concern you raised in your first review. With the explicit return false, the test now fails the load deterministically and for the right reason. I've also tweaked the test comment accordingly. Sorry for the earlier inaccuracy.

Please let me know if you'd prefer the two commits squashed into one before merging.

MarkLee131:
Rebased onto master to resolve the conflict with the recently-landed BadPCX/BadXPM tests (appended BadIFF after them).

vadz:
Thanks for the update! I'll merge this as is, i.e. without squashing, to allow optionally  returning to the first version if we ever decide to implement the idea about a "lax" mode for the image handlers.

vadz:
Merged and added to #26447, so will be part of the next 3.2 release too.

Issue #26408, opened 2026-04-25
Creating a child with disabled parent doesn't visually disable the child


ryancog:
### Description

Whenever a parent window is created, disabled, and then child windows are added, the child windows do not appear disabled.

The child windows are logically disabled. E.g. a button cannot be pressed.

I figure this is only an issue for platforms w/o `wxHAS_NATIVE_ENABLED_MANAGEMENT`. Afaics in wxWindowBase::CreateBase, `SetParent()` is called, which doesn't do the same enable updating (or child handling? I don't know enough here to understand if/why that's okay) that `Reparent()` does.

#### Expected vs observed behaviour:

Newly-created child windows should appear disabled.

#### Patch or snippet allowing to reproduce the problem:

```
// wxWidgets Disabled Parent

#include <wx/app.h>
#include <wx/frame.h>
#include <wx/panel.h>
#include <wx/sizer.h>
#include <wx/button.h>

class Frame : public wxFrame {
public:
    Frame();

private:
    void rebuild();

    wxPanel *mPanel;
    wxSizer *mInnerSizer;
};

class Sample : public wxApp {
public:
    bool OnInit() override {
        auto *frame{new Frame};
        frame->Show();
        return true;
    }
};

// NOLINTNEXTLINE(misc-use-internal-linkage)
wxIMPLEMENT_APP(Sample);

Frame::Frame() : wxFrame(nullptr, wxID_ANY, "Sample") {
    auto *panel{new wxPanel(this)};
    auto *sizer{new wxBoxSizer(wxVERTICAL)};

    mPanel = new wxPanel(panel);

    auto *button{new wxButton(panel, wxID_ANY, "Rebuild")};
    button->Bind(wxEVT_BUTTON, [this](wxCommandEvent&) { rebuild(); });
    sizer->Add(button, 0, wxEXPAND | wxALL, 10);

    mInnerSizer = new wxBoxSizer(wxVERTICAL);
    mPanel->SetSizer(mInnerSizer);

    sizer->Add(mPanel, 1, wxEXPAND | wxALL, 10);
    panel->SetSizer(sizer);

    rebuild();

    mPanel->Disable();
}

void Frame::rebuild() {
    mInnerSizer->Clear(true);

    mInnerSizer->Add(new wxButton(mPanel, wxID_ANY, "Button 2"), 0, wxEXPAND);

    Fit();
    SetMinSize(GetSize());

    mPanel->Layout();
}
```

#### To Reproduce:

[...]

PR #26445, opened 2026-05-10
Ensure child is visually disabled on creation if appropriate (MSW and OSX)

ryancog:
Addresses #26408 for macOS.

I've not yet looked to see if it's a problem on MSW, the other platform which doesn't have `wxHAS_NATIVE_ENABLED_MANAGEMENT`, but it's separate anyways.

Afaics, especially in light of the check for IsShown() immediately above, this would be the proper place to put this. It seems to make things work correctly for me.

### DIFF
--- a/src/msw/window.cpp
+++ b/src/msw/window.cpp
@@ -1497,7 +1497,10 @@ WXDWORD wxWindowMSW::MSWGetStyle(long flags, WXDWORD *exstyle) const
     // wxTopLevelWindow) should remove WS_CHILD in their MSWGetStyle()
     WXDWORD style = WS_CHILD;
 
-    if ( !IsThisEnabled() )
+    // For creation, check not only the enable state of this window, but also
+    // its parent hierarchy when setting this style, as any disabled parent
+    // logically disables this window.
+    if ( !IsEnabled() )
         style |= WS_DISABLED;
 
     // using this flag results in very significant reduction in flicker,
--- a/src/osx/cocoa/window.mm
+++ b/src/osx/cocoa/window.mm
@@ -2737,6 +2737,12 @@ wxWidgetImpl( peer, flags )
     if ( !peer->IsShown() )
         SetVisibility(false);
 
+    // If the widget is created with a disabled parent, the widget will be
+    // logically disabled as well. Make sure its appearance is made consistent
+    // if this is the case.
+    if ( !peer->IsEnabled() )
+        Enable(false);
+
     if ( IsUserPane() )
         ClipsToBounds(true);
 }


ryancog:
Had a stale branch, updated now.

vadz:
Thanks, this looks correct to me (ideal would be to have a unit test for this, of course...), but let's wait a bit to see if Stefan has anything to say about it.

csomor:
@ryancog Thanks for your report and patch. 
This is not the native behavior on macOS, where parent NSViews don't disable children - while NSBox does. So before adapting, I'm not really sure whether I understood your first sentence correctly - could you please verify that this is already the current behavior on wxMSW and wxGTK ? Thanks

csomor (commented):
@ryancog Thanks a lot, the place to put the fix is correct, and I'm almost sure that the other platforms already behave like this, but I'd appreciate if you'd check first. If yes, then I'm fine with merging.

ryancog:
@csomor Right, wxWidgets does disabling hierarchically through all `wxWindow`s.

This isn't native behavior on MSW or macOS, but afaics it's more of a developer convenience than a platform guideline, that an entire section of UI can be disabled at once.

This sort of hierarchical disabling is already the behavior on macOS w/ wxWidgets, it's just that in this edge case the native UI isn't visually drawn as disabled, so while it cannot be interacted with (wxWidgets has code to suppress/ignore events from the controls while `!IsEnabled()`) it looks like it can be.

Normally, if you were to call `Enable()` on a parent, it would affect the state of the children also. `wxWindowBase::NotifyWindowOnEnableChange()` is the primary place for this logic.

I.e. this PR doesn't change the application/wxWidgets behavior, it just makes the UI reflect the enable state of wxWidgets in an edge case where it didn't prior.

Hopefully that makes more sense? Lmk if not.

EDIT: The fact that wxWidgets ignores events for a `!IsEnabled()` control is what I mean by "logically disabled." I believe what controls that in my example is that `-[wxNSView hitTest:]` returns `nil` when the wx peer is `!IsEnabled()`, but there's other similar checks for `IsEnabled()`.

csomor:
@ryancog Thanks, sorry, I was not clear. I know about the overall disabling we do. I just wanted to ask you to verify with the sample of your bug report that the same code under MSW and GTK already does show the control as disabled.

ryancog:
@vadz For unit testing, this would be platform specific afaics, right? I know OSX has `MacIsReallyEnabled()`, but I don't know if the other platforms have something similar. (If they do, it doesn't follow the `IsReally` naming w/ a quick search).

If there are other platform-specific functions you can point me to, or if you just want a case for macOS here, I can put something together. I figure `tests/controls/windowtest.cpp`'s `"Window::Enable"` is the right place.

ryancog:
@csomor Oh, I see.

Well, there's a reason I'm only working on macOS right now :sweat_smile:. My other computer I use for Linux and Windows died overnight, and I'm currently waiting on a replacement.

I can't test that right now, at least. I was planning on keeping the issue open and getting around to confirming that those other platforms either did work correctly or figuring out how to implement a similar fix for them once I have access to those OSs again.

I guess my question would be, without intending to be rude despite the usual tone of the phrase, "what difference does it make?" If they didn't behave like this, it seems like they should. It doesn't make sense to me that the controls could appear enabled but not respond to any user input.

And, if they don't, it seems like that would be better addressed in a separate PR? I'm self taught and not terribly experienced though, so it's likely I misunderstand your perspective, intent, or best practices around this sort of thing.

I apologize for the long-winded response.

ryancog:
Or is it that you are not sure if the controls would be logically disabled when added to a disabled parent under other platforms?

I suppose I took for granted that that behavior would be consistent, since it seems most correct/sensible to me.

csomor:
> Or is it that you are not sure if the controls would be logically disabled when added to a disabled parent under other platforms?
> 
> I suppose I took for granted that that behavior would be consistent, since it seems most correct/sensible to me.

Exactly, I didn't realize that you don't have the other platforms ready to test. I'm concerned about introducing behavioral differences that others may already have written workarounds for. @vadz would you know this ?

vadz:
Well, this is why I wanted to have a unit test... But I am not sure I understand what exactly should it (or I) test, could you (meaning Ryan) please describe what do I need to do?

FWIW I think that the value returned by `wxWindow::IsThisEnabled()` should always be the same as the state of the window at Cocoa level.

csomor:
> FWIW I think that the value returned by `wxWindow::IsThisEnabled()` should always be the same as the state of the window at Cocoa level.

I agree, I just think that some time in the last 28 years, I had the situation that a certain native container control could not be enabled at all, so the native state always returned false, even though in the wx' sense it was enabled. But unfortunately, I have no recollection of what it was ...

ryancog:
@vadz In the associated issue, I provide a simple sample program.

Essentially, the flow is, on current master:

1. Parent Created
2. Child Created (child UI enabled by default)
3. Parent Disabled (child disabled, child UI disabled)
4. Child Destroyed
5. Child Re-created (child disabled by `IsEnabled()` finding parent, child UI enabled by default)

Where, when the child is re-created, the child UI (i.e. `wxWidgetCocoaImpl` on macOS, I don't actually know very well how the other platforms handle this) should be synced to `IsEnabled()`. On macOS at least, currently this syncing happens if the child catches a call to `NotifyWindowOnEnableChange`. That is, if it was already existing at the time when the enable state was changed.

There are test cases for 1-3) (that don't check the UI drawing, but I know from usage this behaves well on the other platforms), aside from it probably being pretty widely used, so I'm confident this works on GTK and MSW as well.

In the second half (4-5), this is what this PR addresses, where I believe the concern is from @csomor, and where the testing would be wanted:

Whenever a child is created with a disabled parent, obviously `IsEnabled()` will immediately return `false`, but it's unclear the exact behavior this results in. So, in the particular case where the child is re-created (in my sample, after pressing "Rebuild"):

- Like on the OSX port, do the other ports *not* receive events when `IsEnabled()` is false? On wxOSX master, in this case the view is (incorrectly, I agree) drawn as enabled, but still wxWidgets (IMO the correct half of the behavior) doesn't send events because `IsEnabled()` is `false`.

- Unlike the OSX port, do the other ports correctly handle the UI being drawn as disabled after this re-creation? I.e. is the control visually disabled after "Rebuild" is pressed?

Those seem like the two questions. Correct me if I'm wrong @csomor.

ryancog:
I should clarify that "re-created" is something of a misnomer in my explanation. I simply mean creating a child after the parent has been disabled.

The "Re" bit comes from the example, but isn't necessary for what I think are obvious reasons. If more children were added w/o deleting old ones, the same behavior would be seen--new children wouldn't appear right.

vadz:
Could you please create at least an interactive test for this? I.e. a small window with buttons that I could press and then check that the state corresponds to the expected one? I don't have time for anything else right now, but I could at least run such example under both Windows and Linux.

ryancog:
Of course, one moment.

ryancog:
I believe this should work.

Radio buttons to enable/disable the children's parent, button to add a child, and each child button will display a message box "Pressed" when clicked (if an event reaches them, of course).

So disabling the box, and then adding children would replicate the situation, testing to see if a) the child button appears disabled and b) if clicking it causes the message box.

[text.cpp](https://github.com/user-attachments/files/27808643/text.cpp)

vadz:
We have a discrepancy between wxMSW and wxGTK: under the latter the child added when "Disabled" is selected appears disabled and is disabled, while under the former it is actually disabled (clicking on it doesn't do anything) but appears enabled, see the corresponding screenshots:

wxGTK: image

wxMSW: image

I think wxGTK behaviour is correct and wxMSW is wrong, we need to check whether the parent is disabled when creating (or reparenting!) a child under it.

ryancog:
That makes sense, at least. I agree wxGTK behavior is correct and aligns with the changes in this PR for macOS.

When reparenting a child this check already happens (`wxWindowBase::Reparent`):
```
    // We need to notify window (and its subwindows) if by changing the parent
    // we also change our enabled/disabled status.
    const bool newEnabledState = IsEnabled();
    if ( newEnabledState != oldEnabledState )
    {
        NotifyWindowOnEnableChange(newEnabledState);
    }
```

So it seems it's only a problem for newly-created windows.

I will look into how to implement a similar fix for MSW.

csomor:
@ryancog @vadz thanks a lot, that's exactly what I feared might be the case, and it's great if you can help iron it out.

ryancog:
@vadz Curious, this was almost accounted for.

MSWGetStyle() checked `IsThisEnabled()` though, rather than `IsEnabled()`. Afaics the fact that MSWGetStyle() even sets the `WS_DISABLED` flags is only for creation, because, as I'm sure you already know, when it's otherwise called (e.g. `MSWUpdateStyle`), these "special" (enable/visible) flags are cleared since they require special API calls in any case besides creation.

I've changed `MSWGetStyle` accordingly. I looked through where and how it's used and don't see anything that should mean that's a problem.

If you could, lmk if that makes the sample behave correctly.

vadz:
Yes, this works as expected under MSW too now, thanks! I'll merge this soon.

PR #26448, opened 2026-05-12
wxMSW: Fix drawing vertical lines with pen width > 1 in RTL layout

AliKet:
In the following pictures, there are four (04) vertical lines (blue & red) in the middle drawn with pen width  == 1 and 2. Blue lines are drawn by calling `wxDrawLine()` function internally. Whereas red lines are drawn by  `wxDrawHVLine()` function.
I removed `m_pen.GetWidth() <= 1 ||` check from this line https://github.com/wxWidgets/wxWidgets/blob/4ed05f7b47d3bd3ac01c4b0d867bece20fe083ff/src/msw/dc.cpp#L710 to expose the bug only. The code I use to draw these line ():
```C++
    // Put this in MyCanvas::DrawDefault() in drawing sample
    for ( int i = 0; i < 2; i++ )
    {
        int dx = 400 + 3*(i);
        int pw = i+1; // pen width

        wxPen pen(*wxBLUE, pw);
        dc.SetPen(pen);

        dc.DrawLine(dx, 0, dx, 80); // calls wxDrawLine() internally

        pen = wxPen(*wxRED, pw);
        pen.SetCap(wxCAP_BUTT);
        dc.SetPen(pen);

        dc.DrawLine(dx, 80, dx, 160); // calls wxDrawHVLine() internally

        dc.SetPen(wxNullPen);
    }
```
### The results:

**In LTR layout:**

drawline

**In RTL layout (with this PR):**

drawline_rtl_ok

**In RTL layout (before):**

drawline_rtl_bad

This PR also fixes a bug in `wxGrid` where the vertical frozen border was not drawn in RTL layout:

**Before:**

grid_frozen_rtl_bad

**After (I changed the grid lines colour to grey):**

grid_frozen_rtl_ok

### DIFF
--- a/src/msw/dc.cpp
+++ b/src/msw/dc.cpp
@@ -718,7 +718,18 @@ void wxMSWDCImpl::DoDrawLine(wxCoord x1, wxCoord y1, wxCoord x2, wxCoord y2)
     }
     else
     {
-        wxDrawLine(GetHdc(), XLOG2DEV(x1), YLOG2DEV(y1), XLOG2DEV(x2), YLOG2DEV(y2));
+        int dx = 0;
+
+        // In RTL layout, we need this adjustment because vertical lines drawn
+        // with pen width > 1 are incorrectly shifted to the left by one pixel.
+        if ( x1 == x2 && y1 != y2 &&
+             m_pen.GetWidth() > 1 &&
+             GetLayoutDirection() == wxLayout_RightToLeft )
+        {
+            dx = -1;
+        }
+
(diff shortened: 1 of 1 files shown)


AliKet:
This is unrelated to #26398

vadz (approved):
Thanks, it's really strange that the shift is 1 independently of the pen width (I would find it understandable if it were `width - 1`), but if this is what we need to do to make this work, let's do it.

I'll merge this soon, thanks again.

PR #26452, opened 2026-05-13
Recommend adding wxWidgets.natvis to one's project

lanurmi:
The primary documented way of using a `.natvis`, in the Microsoft documentation linked to, is to add it to one's project, not to copy it under `%USERPROFILE%`.


vadz:
Good point, thanks.

PR #26454, opened 2026-05-13
MSW: Eliminate redundant system color change events

stevecor:
The handling for WM_SETTINGCHANGE is modified to eliminate redundant wxSysColourChangedEvent events. Prior to this change, the handling for WM_SETTINGCHANGE with the "ImmersiveColorSet" argument introduced an additional system color change event in addition to handling the message normally. This doubled the number of events per window. Since WM_SETTINGCHANGE is applied recursively to child windows, this resulted in an exponential increase in the number of events generated, doubling with each level of child windows. The solution is to handle WM_SETTINGCHANGE with "ImmersiveColorSet" as a system color change only - as if it were WM_SYSCOLORCHANGE.

Also, the wxFrame::OnSysColourChanged() is fixed so that it does explicitly send a wxSysColourChangedEvent to the status bar. The status bar is assumed to be a descendant of the frame and so already gets the event. Now the status bar is handled the same as the toolbar.

### DIFF
--- a/src/msw/frame.cpp
+++ b/src/msw/frame.cpp
@@ -482,15 +482,6 @@ void wxFrame::OnSysColourChanged(wxSysColourChangedEvent& event)
         Refresh();
     }
 
-#if wxUSE_STATUSBAR
-    if ( m_frameStatusBar )
-    {
-        wxSysColourChangedEvent event2;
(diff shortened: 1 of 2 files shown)


vadz:
Thanks! Also applied in #26503 (as 2 separate commits), so closing.

PR #26455, opened 2026-05-14
Do not reuse the old memory DC after the wxOverlay has been reset

AliKet:
Closes #26251


vadz:
Replaced by 358d1b7d3f (Fix reusing wxOverlay after resetting it in wxMSW, 2026-05-15).

Issue #26428, opened 2026-05-05
WebP support in 3.3.x breaks linkage in static builds


hatelamers:
### Build System Used

I build wxWidgets and/or my application using:

- [ x] CMake
- [ ] configure
- [ ] mingw32-make with makefile.gcc
- [ ] MSBuild (Microsoft Visual Studio solution file)
- [ ] nmake with makefile.vc
- [ ] Xcode

### Description

When 3.3.x wxWidgets is configured with `-DCMAKE_BUILD_TYPE=Debug -DwxBUILD_SHARED=OFF -DwxBUILD_USE_STATIC_RUNTIME=ON` it builds allright, however application using it cannot be succefully linked anymore (used to work with 3.2.x versions).

First of all `wxweb` ignores `wxBUILD_USE_STATIC_RUNTIME` setting and tries to link against a dynamic RT lib, hense producing library conflicts while linking
```log
LINK : warning LNK4098: defaultlib 'MSVCRTD' conflicts with use of other libs; use /NODEFAULTLIB:library 
LINK : warning LNK4286: symbol '_wassert' defined in 'libucrtd.lib(assert.obj)' is imported by 'wxsharpyuvd.lib(sharpyuv_csp.obj)' 
LINK : warning LNK4286: symbol '_wassert' defined in 'libucrtd.lib(assert.obj)' is imported by 'wxsharpyuvd.lib(sharpyuv_dsp.obj)' 
LINK : warning LNK4217: symbol '_wassert' defined in 'libucrtd.lib(assert.obj)' is imported by 'wxsharpyuvd.lib(sharpyuv_gamma.obj)' in function 'Log10f' 
LINK : warning LNK4217: symbol '_wassert' defined in 'libucrtd.lib(assert.obj)' is imported by 'wxwebpd.lib(huffman_encode_utils.obj)' in function 'VP8LCreateHuffmanTree'
wxsharpyuvd.lib(sharpyuv_gamma.obj) : error LNK2019: unresolved external symbol __imp_expf referenced in function ToLinearHlg 
wxsharpyuvd.lib(sharpyuv_gamma.obj) : error LNK2019: unresolved external symbol __imp_logf referenced in function FromLinearHlg 
```
Only after supplying additional `-DCMAKE_MSVC_RUNTIME_LIBRARY="MultiThreaded$<$<CONFIG:Debug>:Debug>"` argument, the conflicts go away. None of other wxWidgets components behaves this way.

[...]

PR #26456, opened 2026-05-14
CMake: Fix libwebp build properties

MaartenBent:
Allow to call `wx_add_builtin_library` for external targets. and use it for the libwebp targets.

See #26428


vadz:
Thanks, to be honest my eyes glazed over all the changes here, but I'll just merge them and see how it goes.

Will anything remain in #26428 after applying this PR?

PR #26459, opened 2026-05-14
wxQt: Fix wxRenderer::DrawCheckBox() clipping in RTL layout after rec…

AliKet:
…ent changes

This should have been part of #26401


vadz:
Thanks, will merge soon.

PR #26460, opened 2026-05-14
wxMSW: Add runtime checks for Windows 8+ Pointer API functions

tomay3000:
To support legacy Windows versions, this commit ensures that `GetPointerType()` and `GetPointerPenInfo()` are only invoked if available at runtime. This prevents "Procedure entry point not found" errors on Windows 7.

This adds the missing runtime checks to fix the compatibility issue introduced in 3732496 (Add support for generating wxStylusEvent for graphical tablets, 2026-02-02).

### DIFF
--- a/src/msw/window.cpp
+++ b/src/msw/window.cpp
@@ -6412,19 +6412,17 @@ bool wxWindowMSW::HandlePointer(WXUINT message, WXWPARAM wParam, WXLPARAM lParam
     typedef BOOL (WINAPI *GetPointerPenInfo_t)(UINT32 pointerId, POINTER_PEN_INFO *penInfo);
     static GetPointerType_t s_pfnGetPointerType = nullptr;
     static GetPointerPenInfo_t s_pfnGetPointerPenInfo = nullptr;
-    static bool s_initDone = false;
 
-    if ( !s_initDone )
+    if ( !s_pfnGetPointerType )
     {
         wxLoadedDLL dllUser32("user32.dll");
         wxDL_INIT_FUNC(s_pfn, GetPointerType, dllUser32);
         wxDL_INIT_FUNC(s_pfn, GetPointerPenInfo, dllUser32);
-        s_initDone = true;
     }
 
     const UINT32 pointerId = GET_POINTERID_WPARAM(wParam);
     POINTER_INPUT_TYPE pType = 0;
-    if ( s_pfnGetPointerType && s_pfnGetPointerType(pointerId, &pType) )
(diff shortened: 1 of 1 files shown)


vadz (commented):
Thanks, I'm fine with maintaining Windows 7 compatibility as long as people are interested in it, but I'd like to do it with as little code as possible.

And AFAICS this function (`wxWindow::HandlePointer()`) should be only called under Windows 8+, right? So the function lookup should always succeed and thus the code could be simplified a bit by removing `s_initDone` and simply checking if the function pointer(s) is/are null as they shouldn't be after it's called the first time. And we also don't need to check for them below AFAICS.

tomay3000:
> Thanks, I'm fine with maintaining Windows 7 compatibility as long as people are interested in it, but I'd like to do it with as little code as possible.

Thank you.
I am pretty sure there are many people still using Windows 7, including me, my friends, the old PCs I am still running legacy software on...
Sure, as little code as possible is enough for me, at least it should run.

> And AFAICS this function (`wxWindow::HandlePointer()`) should be only called under Windows 8+, right?

Right.

> So the function lookup should always succeed and thus the code could be simplified a bit by removing `s_initDone` and simply checking if the function pointer(s) is/are null as they shouldn't be after it's called the first time.

I agree with you.
What i did the first time is just tried to follow wxWidgets coding conventions as much as possible.
This file : `src/msw/window.cpp` is full of `s_initDone` (May be it is old code!).

> And we also don't need to check for them below AFAICS.

I don't trust Windows API `GetProcAddress()` to always succeed, do you!
Doing so, wxWidgets could explode in our faces anytime.

a better additional check would be:
```
    if ( !s_pfnGetPointerType || !s_pfnGetPointerPenInfo )
        return false;
```

Correct me if am wrong.
TIA.

Anyway I have updated the code as you suggested.

vadz:
> I don't trust Windows API `GetProcAddress()` to always succeed, do you!

I do trust it to succeed under Windows 8+ where these functions are always present. I've never seen `GetProcAddress()` fail for an existing function and I'd extremely surprised if it ever did this.

> a better additional check would be:
> 
> ```
>     if ( !s_pfnGetPointerType || !s_pfnGetPointerPenInfo )
>         return false;
> ```

We can add this, but I'm fine with applying this PR just as it is (I assume it does work for you under Windows 7, I can't easily test it there any more). 

> Anyway I have updated the code as you suggested.

Thanks!

tomay3000:
> (I assume it does work for you under Windows 7, I can't easily test it there any more).

Confirmed. All samples are running perfectly on Windows 7 with the latest commit.

Issue #26457, opened 2026-05-14
wxAnimation::Load is very slow


BazaarAcc32:
I'm loading a webp file into wxAnimation. The file size is almost 11MB. The webp file contains an animation and consists of 234 frames. However, the wxAnimation::Load method takes about 5 seconds to complete. For example, Google Chrome/Edge loads and displays the same image in about a second.

I'm using the following image: https://colinbendell.github.io/webperf/animated-gif-decode/2.webp

Platform: Windows 11
wxWidgets Version: 3.3.0
Build: Release
Compiler: MSVC, from Visual Studio 2022

Here's the test code

```cpp
#include <wx/wx.h>
#include <wx/webrequest.h>
#include <wx/stopwatch.h>
#include <wx/mstream.h>
#include <wx/animate.h>

class MyFrame : public wxFrame
{
public:
    MyFrame()
        : wxFrame(nullptr,
            wxID_ANY,
            "wxWidgets WebP Animation Timing",
            wxDefaultPosition,
            wxSize(600, 400))
    {
        Bind(wxEVT_WEBREQUEST_STATE,
            &MyFrame::OnWebRequestState,
            this);

        StartDownload();
    }

private:
    void StartDownload()
    {
        m_stopwatch.Start();

        wxWebSession session = wxWebSession::New();

        m_request = session.CreateRequest(
            this,
            "https://colinbendell.github.io/webperf/animated-gif-decode/2.webp"
        );

        m_request.Start();
    }

    void OnWebRequestState(wxWebRequestEvent& evt)
    {
        switch (evt.GetState())
        {
        case wxWebRequest::State_Completed:
        {
            long downloadMs = m_stopwatch.Time();

            wxInputStream* stream = m_request.GetResponse().GetStream();
            if (!stream)
            {
                wxMessageBox("Failed to get response stream", "Error");
                return;
            }

            wxStopWatch decodeWatch;

            wxAnimation animation;
            bool ok = animation.Load(*stream, wxANIMATION_TYPE_WEBP);

            long decodeMs = decodeWatch.Time();

            if (!ok)
            {
                wxMessageBox("Failed to load animation", "Error");
                return;
            }

            wxString text;
            text.Printf(
                "Download time: %ld ms\n"
                "wxAnimation load time: %ld ms, frame count: %ld",
                downloadMs,
                decodeMs,
                animation.GetFrameCount()
            );

            wxMessageBox(text, "Timing Result");

            Close();
            break;
        }

[...]

PR #26461, opened 2026-05-14
Improve speed of reading webp into memory

MaartenBent:
Read the input stream directly into a buffer instead of using a `wxMemoryOutputStream`.

Increase default temporary buffer size of streams. Reading large files using a small buffer causes significant delays.

See #26457

### DIFF
--- a/include/wx/stream.h
+++ b/include/wx/stream.h
@@ -134,6 +134,10 @@ public:
     // when EOF is reached or an error occurs
     wxInputStream& Read(wxOutputStream& streamOut);
 
+    // copy the entire contents of this stream into buffer, stopping only
+    // when EOF is reached or an error occurs
+    bool Read(std::vector<wxUint8>& buffer);
+
 
     // status functions
     // ----------------
--- a/interface/wx/stream.h
+++ b/interface/wx/stream.h
@@ -649,6 +649,17 @@ public:
     wxInputStream& Read(wxOutputStream& stream_out);
 
     /**
+        Read all data from the stream and stores it in the specified buffer.
+        Data is read until the end of the input stream, or until an error occurs.
+        The buffer will be resized to the number of bytes read.
+
+        @return Returns true if the entire stream is read.
+
+        @since 3.3.3
+    */
+    bool Read(std::vector<wxUint8>& buffer);
+
+    /**
         Reads exactly the specified number of bytes into the buffer.
 
         Returns @true only if the entire amount of data was read, otherwise
--- a/src/common/imagwebp.cpp
+++ b/src/common/imagwebp.cpp
@@ -34,8 +34,6 @@ wxIMPLEMENT_DYNAMIC_CLASS(wxWEBPHandler, wxImageHandler);
 
 #if wxUSE_STREAMS
 
-#include "wx/mstream.h"
-
 #include <memory>
 #include <functional>
 
@@ -103,21 +101,20 @@ bool DecodeWebPDataIntoImage(wxImage* image, WebPData* webp_data, bool verbose)
     return true;
 }
 
-WebPDemuxerPtr CreateDemuxer(wxMemoryOutputStream& mos, bool verbose = false)
+WebPDemuxerPtr CreateDemuxer(wxInputStream& stream, std::vector<uint8_t>& buffer, bool verbose = false)
 {
-    wxStreamBuffer* mosb = mos.GetOutputStreamBuffer();
-    if (mosb == nullptr)
+    if (!stream.Read(buffer))
     {
         if (verbose)
         {
-            wxLogError(_("WebP: Allocating stream buffer failed."));
+            wxLogError(_("WebP: Error reading data."));
         }
         return nullptr;
     }
 
     WebPData webp_data{};
-    webp_data.bytes = reinterpret_cast<uint8_t*>(mosb->GetBufferStart());
-    webp_data.size = mosb->GetBufferSize();
+    webp_data.bytes = buffer.data();
+    webp_data.size = buffer.size();
 
     WebPDemuxerPtr demux(
         WebPDemux(&webp_data),
@@ -157,9 +154,8 @@ bool wxWEBPHandler::LoadFile(wxImage* image, wxInputStream& stream, bool verbose
         // be a sub-frame, smaller than the full image. Its size is known, but
         // the x_offset and y_offset can not be queried, so we can't
         // reconstruct the full-size image.
-        wxMemoryOutputStream mos;
-        stream.Read(mos);
-        WebPDemuxerPtr demux = CreateDemuxer(mos, verbose);
+        std::vector<uint8_t> buffer;
+        WebPDemuxerPtr demux = CreateDemuxer(stream, buffer, verbose);
         if (demux != nullptr)
         {
             WebPIterator iter;
@@ -191,21 +187,19 @@ bool wxWEBPHandler::LoadFile(wxImage* image, wxInputStream& stream, bool verbose
 
 bool wxWEBPHandler::LoadAnimation(std::vector<wxWebPAnimationFrame>& frames, wxInputStream& stream, bool verbose)
 {
-    wxMemoryOutputStream mos;
-    stream.Read(mos);
-    wxStreamBuffer* mosb = mos.GetOutputStreamBuffer();
-    if (mosb == nullptr)
+    std::vector<uint8_t> buffer;
+    if (!stream.Read(buffer))
     {
         if (verbose)
         {
-            wxLogError(_("WebP: Allocating stream buffer failed."));
+            wxLogError(_("WebP: Error reading data."));
         }
         return false;
     }
 
     WebPData webp_data{};
-    webp_data.bytes = reinterpret_cast<uint8_t*>(mosb->GetBufferStart());
-    webp_data.size = mosb->GetBufferSize();
+    webp_data.bytes = buffer.data();
+    webp_data.size = buffer.size();
 
     WebPAnimDecoderOptions dec_options;
     WebPAnimDecoderOptionsInit(&dec_options);
@@ -305,9 +299,8 @@ bool wxWEBPHandler::SaveFile(wxImage* image, wxOutputStream& stream, bool)
 int wxWEBPHandler::DoGetImageCount(wxInputStream& stream)
 {
     int frame_count = 0;
-    wxMemoryOutputStream mos;
-    stream.Read(mos);
-    WebPDemuxerPtr demux = CreateDemuxer(mos);
+    std::vector<uint8_t> buffer;
+    WebPDemuxerPtr demux = CreateDemuxer(stream, buffer);
     if (demux != nullptr)
     {
         frame_count = WebPDemuxGetI(demux.get(), WEBP_FF_FRAME_COUNT);
--- a/src/common/stream.cpp
+++ b/src/common/stream.cpp
@@ -915,6 +915,43 @@ wxInputStream& wxInputStream::Read(wxOutputStream& stream_out)
     return *this;
 }
 
+bool wxInputStream::Read(std::vector<wxUint8>& buffer)
+{
+    const size_t stream_size = IsOk() ? GetSize() : 0;
+    size_t lastcount = 0;
+
+    if ( stream_size > 0 )
+    {
+        buffer.resize(stream_size);
+        lastcount = Read(buffer.data(), stream_size).LastRead();
+
+        // Read one more byte to set EOF
+        wxUint8 temp;
+        Read(&temp, 1);
+    }
+    else if ( IsOk() )
+    {
+        const size_t read_size = BUF_TEMP_SIZE * 8;
+        for ( ;; )
+        {
+            if ( buffer.size() < (lastcount + read_size) )
+                buffer.resize(lastcount + read_size);
+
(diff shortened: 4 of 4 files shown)


vadz:
Thanks but, just in case you didn't notice it, this seems to have broken WebP unit tests.

src/common/stream.cpp
```diff
@@ -448,16 +448,16 @@ size_t wxStreamBuffer::Read(wxStreamBuffer *dbuf)
 {
     wxCHECK_MSG( m_mode != write, 0, wxT("can't read from this buffer") );
 
-    char buf[BUF_TEMP_SIZE];
+    std::vector<char> buf(BUF_TEMP_SIZE);
```
PBfordev: This changes the buffer allocation from stack to heap. The function can be called in a loop. Doing something like this usually was not recommended (but it was back then, when the heap was fragmentation-prone and smaller).

I understand that this is probably because of drastically increasing the buffer size. Perhaps a buffer size of the 16 kB would be a sweet spot between stack memory consumption and reading speed? I don't think wxWidgets has a stream benchmark, where this would be easy to test...

MaartenBent: Yes, I mentioned in the commit message that I changed it to use the heap. To prevent a bigger stack allocation from causing downstream issues. I was not aware of these loop recommendations.

I could make `std::vector<char> buf` a member of the class so it needs to be allocated only once?

src/common/imagwebp.cpp
```diff
@@ -103,21 +103,46 @@ bool DecodeWebPDataIntoImage(wxImage* image, WebPData* webp_data, bool verbose)
...
+
+        wxStreamBuffer* mosb = mos.GetOutputStreamBuffer();
+        size = mosb->GetBufferSize();
+        mosb->Seek(0, wxFromStart);
+
+        buffer.resize(size);
+        size_t readSize = mosb->Read(buffer.data(), size);
+        success = readSize == size;
+    }
+    else {
```
vadz: Trivial:
```suggestion
    else
    {
```

src/common/imagwebp.cpp
```diff
@@ -103,21 +103,46 @@ bool DecodeWebPDataIntoImage(wxImage* image, WebPData* webp_data, bool verbose)
...
+
+    size_t size = stream.GetSize();
+    if (size == 0)
     {
-        if (verbose)
-        {
-            wxLogError(_("WebP: Allocating stream buffer failed."));
-        }
-        return nullptr;
+        // size unknown (for example zip stream), load via wxMemoryOutputStream
```
vadz: Shouldn't we have a helper function in `wxInputStream` itself doing this, i.e. reading all data into a buffer in the optimal way? This doesn't seem WebP-specific.

src/common/stream.cpp
vadz: FWIW I don't think allocating 64KiB on the stack is a problem for any of the platforms supported by wxWidgets, so IMHO we don't need to change this buffer to a vector at all.

vadz (commented):
Thanks for improving this but I agree that reallocating the buffer for each call to `Read()` may not be ideal (although I suspect not a big deal either, considering how little all this code is optimized for speed anyhow). I would just increase the size of the existing buffer and do something else only if anybody runs into a problem due to this (which seems very unlikely).

src/common/stream.cpp
MaartenBent: Ok, I've reverted the changes to only increasing the buffer size.

src/common/imagwebp.cpp
MaartenBent: Added it. I hope the API is fine like this.
I kept using the `wxMemoryOutputStream`. I could also replace it with the contents of `wxInputStream::Read(wxOutputStream& stream_out)`, but write to (and keep resizing) the vector instead. Any preference?

MaartenBent: Actually, I will replace it. Because both cases (size known/unknown) can use the same code then.

include/wx/stream.h
```diff
@@ -134,6 +134,10 @@ class WXDLLIMPEXP_BASE wxInputStream : public wxStreamBase
     // when EOF is reached or an error occurs
     wxInputStream& Read(wxOutputStream& streamOut);
 
+    // copy the entire contents of this stream into buffer, stopping only
+    // when EOF is reached or an error occurs
+    bool Read(std::vector<char>& buffer);
```
vadz: I wonder if we're not going to have problems with people (possibly including ourselves) wanting to use this with `unsigned char` or `uint8_t` or `std::byte`... But I don't see any nice way to generalize it, we probably don't want to make the entire function template and extracting a type-independent implementation of it would require using some `VectorHolder` class with virtual functions wrapping `data()` and `resize()` which would be quite ugly.

So I guess for now this is the best we can do, except that perhaps we should use `unsigned char`?

interface/wx/stream.h
```diff
@@ -648,6 +648,15 @@ class wxInputStream : public wxStreamBase
     */
     wxInputStream& Read(wxOutputStream& stream_out);
 
+    /**
+        Read all data from the stream and stores it in the specified buffer.
+        Data is read until the end of the input stream, or until an error occurs.
+        The buffer will be resized to the number of bytes read.
+
```
vadz, line 655: It would be good to document that it returns false if nothing was read, even if no error occurred which is not obvious but which is how the code seems to behave (at least if I'm reading the code correctly).

src/common/stream.cpp
```diff
@@ -915,6 +915,41 @@ wxInputStream& wxInputStream::Read(wxOutputStream& stream_out)
...
+
+            lastcount += bytes_read;
+        }
+    }
+
+    m_lastcount = lastcount;
+
+    buffer.resize(m_lastcount);
+
+    return !buffer.empty();
```
vadz: Actually I don't think it's the correct return value. In the single read case it should return `lastcount == stream_size` and in the other one it should `return false` instead of `break` (possibly after setting `m_lastcount`).

vadz (commented):
Thanks, but I think there is a problem with the new `Read()` overload return value, please let me know if I'm missing something.

include/wx/stream.h
MaartenBent: I tried `uint8_t` initially (because `libwebp` uses it) but that didn't work everywhere without extra includes.
I'm using `char` because it is already used by other functions in the classes in `stream.h`, while `unsigned` is not used at all.

vadz: I still think `char` is a bad choice for the "byte" type due to the possibility of introducing difficult to notice bugs when using it and I don't think it's used anywhere in th public API (`char` itself is, but not `char*`), so I'd still consider using `unsigned char` (a.k.a. `wxUint8` a.k.a. `wxByte`) but I'm leaving the choice to you.

src/common/stream.cpp
MaartenBent: I think it should return `true` if the entire input stream is written into the buffer. I'll update the code.

The second case will always `break`, it keeps reading until nothing is available. In this case I'll return `true` when no error happened.

include/wx/stream.h
MaartenBent: I changed it to `wxUint8` now.

src/common/stream.cpp
```diff
@@ -915,6 +915,43 @@ wxInputStream& wxInputStream::Read(wxOutputStream& stream_out)
     return *this;
 }
 
+bool wxInputStream::Read(std::vector<wxUint8>& buffer)
+{
+    size_t stream_size = GetSize();
+    size_t lastcount = 0;
+    char buf[BUF_TEMP_SIZE];
```
vadz: I'm sorry, I didn't think of it before somehow, but why do we have this temp buffer at all? I.e. why not just read directly into the vector buffer? We could call `resize()` before `Read()` and then `resize()` it again if we read less than `CHUNK_SIZE` bytes (which should happen just once at the end).

This would avoid any worries about the stack overflow (and we could possibly use much bigger buffer if it's warranted) and would save a `memcpy()` which is definitely very cheap but, still, why do it if we don't have to?

src/common/stream.cpp
```diff
@@ -915,6 +915,43 @@ wxInputStream& wxInputStream::Read(wxOutputStream& stream_out)
...
+
+            lastcount += bytes_read;
+        }
+    }
+
+    m_lastcount = lastcount;
+
+    buffer.resize(m_lastcount);
+
+    return Eof();
```
vadz, line 952: I'm sorry, I never know how do our own stream classes work, so maybe it's a stupid question, but what does this return if there was an error on the stream? Does `Eof()` return false (as I'd expect this function to) in this case?

vadz (commented):
Thanks for the update, the API looks perfectly fine now but I wonder about whether we could improve the implementation a bit now. Please let me know if you think I'm nitpicking here.

src/common/stream.cpp
MaartenBent: My reason for using the temporary  `char buf` is to reduce the number of calls to `resize()`. Because there will always be 1 unnecessary `resize()` call before the last `Read()`, when `LastRead()` returns `0`.
And when you're unlucky and the buffer capacity is too small, and has to reallocate and copy the entire buffer.

But maybe these are micro-optimizations and shorter code is preferable.

src/common/stream.cpp
MaartenBent: Yes. `Eof()`returns the result of `GetLastError() == wxSTREAM_EOF`. So end-of-file is considered a specific type of error.

When a different error occurs, `GetLastError()` will return `wxSTREAM_WRITE_ERROR` or `wxSTREAM_READ_ERROR`, and then `Eof()` will  return false.

src/common/stream.cpp
vadz: Wouldn't this only happen when the data size is a multiple of buffer size, i.e. pretty rarely?

MaartenBent: I think it will be quite rarely. It is probably `std` implementation dependent, with msvc I've seen the allocated memory/capacity be doubled every time the size exceeds the capacity.

These are just estimates, but I think it works a bit like this. You have a 1023kb input file/stream. While reading 64kb chunks and resizing the vector, it keeps doubling the capacity when necessary, to 256kb, 512kb, 1024kb.
After reading the last 63kb chunk, it will do one more read to detect the end-of-file and return `0`.
But before that last read, you `resize` to 1023+64=1087kb, which exceeds the 1024kb capacity. The capacity will be increased to 2048kb, which needs a new memory allocation and possibly copying of the existing data.

So I guess it could happen when the data size is between 2<sup>n</sup>-BUF_TEMP_SIZE and 2<sup>n</sup>.

On the other hand, reading into the vector directly instead of using the temp buffer would allow to read much bigger chunks at once. So I'm fine with either solution.

vadz: We could keep a (relatively small, maybe just original 4KB) buffer on stack but only use it for the read after the one that would return less than the (large) chunk size used when reading into the vector. This should be as good as we can do without any heroic efforts, AFAICS.

But this would make the code more complicated for a relatively small gain in the relatively rare case of using unseekable streams, so I don't know if it's worth it, probably not.

I'll let you choose whatever you prefer -- just please let me know if you're going to make more changes or if this should be applied as is. TIA!

paulcor:
Using the test program below, and the webp file from #26457, on a Linux system with an optimized wxGTK3 build, using the current master I get an average-of-5 time of 1222ms. With these changes, I get... 1222ms. No measurable improvement. Am I missing something?

```cpp
#include <wx/app.h>
#include <wx/image.h>
#include <wx/wfstream.h>
#include <wx/stopwatch.h>

class MyApp : public wxApp
{
    bool OnInit() override
    {
        wxWEBPHandler h;
        wxFileInputStream fis("2.webp");
        std::vector<wxWebPAnimationFrame> frames;

        wxStopWatch stopWatch;
        stopWatch.Start();
        if (h.LoadAnimation(frames, fis))
            wxLogMessage("%ld ms", stopWatch.Time());

        return false;
    }
};
wxIMPLEMENT_APP(MyApp);
```

PBfordev:
> Using the test program below, and the webp file from #26457, on a Linux system with an optimized wxGTK3 build, using the current master I get an average-of-5 time of 1222ms. 

You must either have a very fast computer or there is a huge difference between Windows and Linux.

Testing with and old Windows 10 PC desktop (i5-7600, baseline 3.50 GHz) and using your code (compiled with MSVS 2022 64-bit Release build), it takes about 6 seconds, i.e., the same as with my almost similar code from #26457 but using `wxAnimation::LoadFile()`.

MaartenBent:
CI errors seem unrelated.

> Am I missing something?

Tried it in a Ubuntu VM and there the differences are also neglectable. Maybe this problem is specific to Windows or MSVC.

src/common/stream.cpp
MaartenBent: Changed it to read directly into the vector, with bigger chunks. Ready to merge.

PBfordev:
> > Am I missing something?
> 
> Maybe this problem is specific to Windows or MSVC.

I have tried also GCC 15.2 from MSYS2 UCRT64 package with self-built wxWidgets using built-in libraries (not in MSYS2 shell). The time was comparable to that of MSVC, about 5 s for both (different PC than in my previous post, CPU Ryzen 5 2600);

Issue #26458, opened 2026-05-14
Sample projects in CMake-generated MSVS solution have MBCS character set


PBfordev:
In the master (and 3.3.2, did not test older 3.3), the samples project in the CMake-generated MSVS solution have character set to _Use Multi-Byte Character Set_ instead of the expected _Use Unicode Character Set_. 

<img width="853" height="377" alt="Image" src="https://github.com/user-attachments/assets/347fe0c8-389f-4654-9347-b055ed3ac3f7" />

<img width="490" height="484" alt="Image" src="https://github.com/user-attachments/assets/415783e2-9cd4-4192-b952-bc8da7cf1075" />

I did not change any defaults, the solution was generated with just
```bat
cmake -G "Visual Studio 18 2026" -DwxBUILD_SAMPLES=SOME -S ../wxWidgets-PB -B .
```

However, the libraries' charset is set to Unicode, the sample reports it is built as Unicode, so this is probably just a cosmetic issue.

I used CMake 4.3.2, so this may be related to https://cmake.org/cmake/help/latest/policy/CMP0204.html:
> In CMake 4.1 and below, [Visual Studio Generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#visual-studio-generators) compile sources in with _MBCS, _UNICODE or _SBCS defined due to behavior of Visual Studio itself. The preprocessor definition is not modeled by CMake and is therefore not added by other generators, such as [Ninja](https://cmake.org/cmake/help/latest/generator/Ninja.html#generator:Ninja).

> **CMake 4.2 and above, when targeting the MSVC ABI, prefer to compile sources with _MBCS** defined by all generators unless another charset preprocessor definition is found (_UNICODE or _SBCS). This policy provides compatibility with projects that have not been updated to be aware of the definition. Its setting is recorded by each target as it is created, and affects compilation of sources in that target.

where the samples may not have `_UNICODE` defined? For some reason, the charset is Unicode in wxWidgets 3.2.10 for samples, but this may be due to Unicode-related in changes in wxWidgets 3.3...

MaartenBent:
This is the result of #25965. `UNICDE`/`_UNICODE` is still defined when including wxWidgets headers (via `platform.h`), but it is not inherited as compile option when linking with the libraries. As a result the IDE doesn't know about it, in that sense it is just a cosmetic problem.

[...]

PR #26462, opened 2026-05-14
CMake: Add UNICODE define to all targets

MaartenBent:
So IDEs like Visual Studio show the correct character set.

Add an option to override the character set and use it in the win32 API tests.

See #26458


vadz:
Thanks for fixing this!

PR #26465, opened 2026-05-15
Update Poedit links in documentation

vslavik:
Update links to Poedit in the docs, plus adjacent things things such as the name or current path to xgettext.


vadz:
Thanks, will also merge into both master and 3.2 soon.

PR #26473, opened 2026-05-17
Fixes for `wxUSE_XXX` in CMake

vadz:
@MaartenBent I'd be grateful for your review of this, TIA!

### DIFF
--- a/build/cmake/init.cmake
+++ b/build/cmake/init.cmake
@@ -429,16 +429,13 @@ if(wxUSE_GUI)
         set(wxUSE_ENH_METAFILE ON)
     endif()
 
-    # Direct2D check
-    if(WIN32 AND wxUSE_GRAPHICS_DIRECT2D)
+    # Direct2D check: skip it for MSVC which is known to always have it
+    if(WIN32 AND wxUSE_GRAPHICS_DIRECT2D AND NOT MSVC)
         check_include_file(d2d1.h HAVE_D2D1_H)
         if (NOT HAVE_D2D1_H)
             wx_option_force_value(wxUSE_GRAPHICS_DIRECT2D OFF)
         endif()
     endif()
-     if(MSVC) # match setup.h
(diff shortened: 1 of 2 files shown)


MaartenBent:
`build/cmake/setup.h.in` is a copy of `./setup.h.in`, which also has the double defines. Not sure if this part is auto-generated or just copied.

Decoupling `wxUSE_GRAPHICS_DIRECT2D` from `wxUSE_GRAPHICS_CONTEXT` seems fine to me. Originally it tried to match the defines in the visual studio setup.h file. Just make sure the code correctly handles the case when `wxUSE_GRAPHICS_CONTEXT` is off but `wxUSE_GRAPHICS_DIRECT2D` is on.

vadz:
> `build/cmake/setup.h.in` is a copy of `./setup.h.in`, which also has the double defines. Not sure if this part is auto-generated or just copied.

Ugh, sorry for stupidly forgetting about this. These checks come from `setup_inc.h` so I need to either modify it or improve the generation script. This needs more time than I have right now :-(
 
> Decoupling `wxUSE_GRAPHICS_DIRECT2D` from `wxUSE_GRAPHICS_CONTEXT` seems fine to me. Originally it tried to match the defines in the visual studio setup.h file. Just make sure the code correctly handles the case when `wxUSE_GRAPHICS_CONTEXT` is off but `wxUSE_GRAPHICS_DIRECT2D` is on.

Is this a valid build configuration? I don't think so but OTOH `include/wx/msw/chkconf.h` doesn't check for this, so maybe it is...

MaartenBent:
`wxUSE_GRAPHICS_DIRECT2D=1` and `wxUSE_GRAPHICS_CONTEXT=0` is not a valid build option.
So if you want to continue with these changes, a check needs to be added to `chkconf.h`, or CMake should disable `wxUSE_GRAPHICS_DIRECT2D` when `wxUSE_GRAPHICS_CONTEXT` is disabled.

PR #26474, opened 2026-05-17
Switch image handling in HTML from wxImage to wxBitmapBundle

Randalphwa:
Primary reason is to allow SVG images in HTML pages, with the added bonus of multi-resolution png files being bundled together for better scaling on high DPI displays.

I also added a toolbar to the html/test sample just to make it easier to move forward/back including disabling/enabling the buttons depending on whether any history was available.


src/html/m_image.cpp
```diff
@@ -393,18 +393,45 @@ wxHtmlImageCell::wxHtmlImageCell(const wxHtmlTag& tag,
...
+
+                // SVG image path
+                if ( loc.Matches("*.svg") || loc.Matches("*.SVG") )
+                {
+#ifdef wxHAS_SVG
+                    // Read the entire stream into a buffer for SVG
+                    wxMemoryBuffer svgBuf;
+                    for ( ;; )
+                    {
+                        char tmp[4096];
```
vadz: We have just seen in #26461 that using 64KiB buffer is much faster than using 4KiB, so I think we should change the size here too. Or maybe just use `Read(std::vector<char>& buffer)` overload once the other PR is merged.

src/html/m_image.cpp
```diff
@@ -393,18 +393,45 @@ wxHtmlImageCell::wxHtmlImageCell(const wxHtmlTag& tag,
...
+                        s->Read(tmp, WXSIZEOF(tmp));
+                        const size_t n = s->LastRead();
+                        if ( n == 0 )
+                            break;
+                        svgBuf.AppendData(tmp, n);
+                    }
+                    wxBitmapBundle svgBundle = wxBitmapBundle::FromSVG(
```
vadz: Or maybe add `FromSVG()` overload taking a stream?

vadz (commented):
Looks good, thanks!

I'd just like to avoid duplicating the code reading SVGs from stream, could you please do this? TIA!

src/html/m_image.cpp
Randalphwa: I'm going to wait until #26461 gets checked in. With SVG in an HTML file, I'm concerned about a decompression bomb blowing up the caller, so if ReadStream doesn't have a buffer size limit setting, then I wouldn't risk it for SVG inside HTML. We don't currently support .svgz format, but that could definitely be supported in the future (and probably should be) -- and that would be the biggest risk for a decompression bomb with an svgz file from an untrusted source.

vadz: To handle this we'd need to add something like `maxSize = 0` (meaning none by default) parameter to the `Read()` overload being added there and probably also provide some way of setting it for SVGs explicitly as I don't think we can choose one max value fitting all use cases.

Randalphwa: Hmm, I need to do a deeper dive into this. A compression bomb can also be in webp as well -- I gather it's something Google is trying to work around. I *think* for an SVG the only way to get a compression bomb is via .svgz -- in that case, we would presumably be calling 'wxZlibInputStream(wxInputStream* stream)' -- which would mean that trying to read the entire stream at once could blow up the app unless there was a limit with an error return if that limit was succeeded. For SVG that would probably be around 10 MB as the largest valid SVG -- no idea what webp, jpeg, png, tiff, etc. are -- Tiff I know can easily be over 1GB in size and still be valid.

vadz:
I'm marking this as draft because I think you still plan to change it — please let me know if this was wrong or reset the draft status when it's done. TIA!

Randalphwa:
Still a draft -- I need to update documentation, and fix the Check Mixed EOL test that doesn't realize a .svgz is a binary file. However, the updated htmltest is now able to display svg_test.htm which has both a regular and compressed .svg/.svgz file referenced in the HTML file, so support for .svgz is definitely working, as well as the new FromSVG that takes an input stream per @vadz recommendation (which made implementing support for .svgz trivial).

vadz (approved):
Thanks, all this looks good to me but, just to confirm before merging: the removed Mac-specific code is really not needed any longer, i.e. high DPI bitmaps are still used under Mac even without it?

Also, a minor point but I think the new code doesn't compile if `wxUSE_ZLIB && !wxUSE_STREAMS`, but, again, this is very minor and I'm not sure if this build configuration is supported even now.

And another very minor point is that we use both `s.Lower().EndsWith(".svgz")` and `s.Matches("*.gif") || s.Matches("*.GIF")` which just seems gratuitously inconsistent, I'd standardize on one or the other (or, alternatively, `wxFileName(s).GetExt().Lower() == "foo"` which is the most correct, even also the most cumbersome).

Randalphwa:
Since everything is now going through our wxBundle code, there should no longer be any reason to special case Mac -- but I'll check that to be sure.

It may not be supported, but I agree with you that adding the !wxUSE_STREAMS is both correct, and avoids a potential maintenance issue down the road.

I also agree that 'wxFileName(s).GetExt().Lower()' would be preferred -- replacing the *gif handling and using the same system -- it's a bit more processing, but this isn't in a loop, so it won't have any realistic impact on performance.

Randalphwa:
> Thanks, all this looks good to me but, just to confirm before merging: the removed Mac-specific code is really not needed any longer, i.e. high DPI bitmaps are still used under Mac even without it?

I had AI look into the history of the Mac code -- the following is it's respons:

---

**History — why Mac was special-cased.** The code was added deliberately by Julian Smart in May 2017:
- `32355f7172` — *"Implemented @2 HiDPI images for wxHTML on Mac"*
- `d3687e7da3` — *"Fixed suffix which should be @2x"* (same day)

At the time there was no `wxBitmapBundle`, so `wxHtmlImageCell` had to manually probe for a sibling `name@2x.ext` file, open it, and build a `wxBitmap` with scale factor 2 (`m_bitmap = new wxBitmap(img, -1, scaleHDPI)`), keeping the logical size at half the physical size. That's the entire purpose of the removed block in the IMG tag handler plus the `scaleHDPI` parameter

The new code path is:
```cpp
wxImage image(*s, wxBITMAP_TYPE_ANY);       // s = stream from m_WParser->OpenURL()
SetImage(wxBitmapBundle(image));
```
That is a **single-bitmap** bundle built from whatever URL `OpenURL()` resolved — i.e. `foo.png`. There is no `@2x` probe anywhere else (I grepped `src/html`, `src/common`, `src/osx` — nothing). Note that `wxBitmapBundle::FromFiles()` *does* implement the `@2x`/`_2x`/`2.0x` lookup, but wxHTML can't use it because it loads images as streams through `wxFileSystem` (possibly remote), not as local paths.

**Answer to Vadim:** High-DPI *rendering* still works — on a Retina Mac the DC scale factor is still applied and the bundle returns a bitmap for the requested size. But the specific behavior the removed code implemented (`@2x` **source file** selection) is genuinely gone: a page referencing `logo.png` will no longer pick up `logo@2x.png`, so that image is drawn upscaled from the 1x source (blurrier on Retina). A single scale-1 bitmap cannot render crisply on a 2x display regardless of the draw path.

So it's not strictly "not needed any longer" — it's an **image-quality regression** for pages relying on the (undocumented) `@2x` convention, not a functional break.

Options:
- **(a) Restore it, bundle-aware** — keep the macOS probe in the IMG handler and build a 2-entry bundle (`1x` + `@2x`) instead of a lone scaled bitmap. Preserves the feature; keeps a Mac `#if`.
- **(b) Accept the regression** — consider this intentional: for raster HTML images the bundle is built from the single resolved file, and the `@2x` naming convention is dropped. (SVG is unaffected — it's resolution-independent.)

vadz:
Thanks, I think we can live with this regression as people are supposed to use bundles to get proper high DPI support, but this probably still needs to be documented (unless we consider it to be too niche?).

I have to say that I've completely forgotten about this PR because it's marked as "draft" and so doesn't appear in my usual view, but I think it would be good to merge it before 3.3.4. What exactly remains to be done here before this could happen?

Randalphwa:
I was originally hoping to tie this into the Filtering SVG code I tried to add to lunasvg -- but that PR never got accepted on lunasvg's side, so I'm going to take a different approach for that, and this one definitely needs to be in the upcoming release or I'll break my own product which has been using this for months now...  

AI seems to think this was never documented -- which would definitely make it niche.

I resolved all the merge conflicts with the PRs from @LegalizeAdulthood  when I rebased with master, and fixed the other items you mentioned -- that's ready to check in unless you want me to document the change to Mac Retina displays -- though I'm not sure where if wasn't documented in the first place. Then I need @LegalizeAdulthood to verify I didn't break any of his code changes in m_layout.cpp, then we're good to go.

interface/wx/html/htmlwin.h
```diff
@@ -123,10 +123,18 @@ class wxHtmlWindowInterface
...
           consider using wxWebView instead.
 
-    wxHtmlWindow uses the wxImage class for displaying images, so you need to
-    initialize the handlers for any image formats you use before loading a page.
+    wxHtmlWindow uses the wxBitmapBundle class for displaying images, so you
+    need to initialize the handlers for any raster formats you use before
+    loading a page.
     See ::wxInitAllImageHandlers and wxImage::AddHandler.
 
+    New to wxWidgets 3.3.3 is the ability to load vector graphics in SVG format
```
vadz:
```suggestion
    New to wxWidgets 3.3.4 is the ability to load vector graphics in SVG format
```

interface/wx/bmpbndl.h
```diff
@@ -311,16 +311,36 @@ class wxBitmapBundle
...
         that function documentation for more information about SVG support.
 
-        @param path Path to the SVG file. Notice that it should a local file,
-            not an URL.
+        @param stream The input stream to read SVG data from.
+        @param sizeDef The default size to return from GetDefaultSize() for
+            this bundle. If an empty wxSize or wxDefaultSize is provided,
+            the size of the SVG will be the default size.
+
+        @since 3.3.3
```
vadz:
```suggestion
        @since 3.3.4
```

interface/wx/bmpbndl.h
```diff
@@ -311,16 +311,36 @@ class wxBitmapBundle
...
+    /// @overload
+    static wxBitmapBundle FromSVG(wxInputStream& stream, const wxSize& sizeDef);
+
+    /**
+        Create a bundle from the SVG image loaded from the given file.
+
+        This function loads the SVG data from the given @a path and calls
+        FromSVG() with it. If the filename has a .svgz extension, then it
+        will first be decompressed before converting it into a bundle (new
+        since wxWidgets 3.3.3). As this function is just a wrapper for
```
vadz:
```suggestion
        since wxWidgets 3.3.4). As this function is just a wrapper for
```

src/html/m_image.cpp
```diff
@@ -393,18 +397,46 @@ wxHtmlImageCell::wxHtmlImageCell(const wxHtmlTag& tag,
...
+                    if ( svgBundle.IsOk() )
+                    {
+                        SetImage(svgBundle);
+                        readImg = false;
+                    }
+                }
+#endif // wxUSE_ZLIB
+
+                // SVG image path
+                if ( loc.Matches("*.svg") || loc.Matches("*.SVG") )
```
vadz: This is inconsistent with the test above for some reason, should be


```suggestion
                if ( loc.Lower().Matches("*.svg") )
```

src/html/m_image.cpp
```diff
@@ -393,18 +397,46 @@ wxHtmlImageCell::wxHtmlImageCell(const wxHtmlTag& tag,
...
+                        SetImage(svgBundle);
+                        readImg = false;
+                    }
+                }
+#endif // wxHAS_SVG
+
+#if wxUSE_GIF && wxUSE_TIMER
+                if ( readImg && m_windowIface &&
+                     (loc.Matches(wxT("*.gif")) ||
+                      loc.Matches(wxT("*.GIF"))) )
```
vadz:
```suggestion
                     loc.Lower.Matches(wxT("*.gif")) )
```

src/html/m_image.cpp
```diff
@@ -548,8 +570,9 @@ void wxHtmlImageCell::Layout(int w)
 
         m_Width = w*m_bmpW/100;
 
-        if (!m_bmpHpresent && m_bitmap != nullptr)
-            m_Height = m_bitmap->GetLogicalHeight()*m_Width/m_bitmap->GetLogicalWidth();
+        // Make certain to avoid a division by zero if the bitmap has no width.
+        if (!m_bmpHpresent && m_bitmapBundle.IsOk() && m_bitmapBundle.GetDefaultSize().GetWidth() != 0)
+            m_Height = m_bitmapBundle.GetDefaultSize().GetHeight()*m_Width/m_bitmapBundle.GetDefaultSize().GetWidth();
         else
             m_Height = static_cast<int>(m_scale*m_bmpH);
```
vadz: Maybe simplify it like this:

```suggestion
        wxSize bmpSize;
        if (!m_bmpHpresent && m_bitmapBundle.IsOk())
            bmpSize = m_bitmapBundle.GetDefaultSize();
        if (bmpSize.GetWidth())
            m_Height = bmpSize.GetHeight()*m_Width/bmpSize.GetWidth();
        else
```

?

(I'd probably also use `x` and `y` instead of `Get{Width,Height}()`, but this is subjective).

src/html/m_layout.cpp
```diff
@@ -327,9 +332,35 @@ TAG_HANDLER_BEGIN(BODY, "BODY")
...
+                        wxBitmapBundle svgBundle =
+                            wxBitmapBundle::FromSVG(zlibStream, wxDefaultSize);
+                        if ( svgBundle.IsOk() )
+                            winIface->SetHTMLBackgroundImage(svgBundle);
+                    }
+                    else
+#endif // wxUSE_ZLIB
+
+                    // SVG background image path
+                    if ( loc.Matches("*.svg") || loc.Matches("*.SVG") )
```
vadz: Also

```suggestion
                    if ( loc.Lower().Matches("*.svg") )
```

but the need to make the same edit in 2 files naturally raises the question of whether we should extract this code in some private function and reuse it from both of them?

vadz (commented):
Please resolve the remaining conflict and remove the "Draft" status when it's ready (or let me know if I should do it myself).

Thanks!

LegalizeAdulthood:
> I was originally hoping to tie this into the Filtering SVG code I tried to add to lunasvg -- but that PR never got accepted on lunasvg's side, so I'm going to take a different approach for that, and this one definitely needs to be in the upcoming release or I'll break my own product which has been using this for months now... 
> 
> AI seems to think this was never documented -- which would definitely make it niche.
> 
> I resolved all the merge conflicts with the PRs from @LegalizeAdulthood when I rebased with master, and fixed the other items you mentioned -- that's ready to check in unless you want me to document the change to Mac Retina displays -- though I'm not sure where if wasn't documented in the first place. Then I need @LegalizeAdulthood to verify I didn't break any of his code changes in m_layout.cpp, then we're good to go.

My general workflow is "first person to commit wins" `:)` rebasing my changes on master is easy to do, so don't wait for me unless you want to explicitly have your changes on top of mine, such as when you want my changes underneath your changes.

LegalizeAdulthood:
I don't see anything pushed to the PR branch that indicates a conflict resolution with any changes I've made recently, so there doesn't appear to be anything for me to do here?

Randalphwa:
I got another merge conflict with the upstream m_layout.cpp that I have to resolve first before anything shows up...

Randalphwa:
I haven't looked at every one of these build failures, but those that I have were due to nanosvg failures which I didn't touch in this PR.

vadz:
[This one](https://app.circleci.com/pipelines/github/wxWidgets/wxWidgets/7022/workflows/bc00bb9e-74b5-473a-bdb8-ea90e0219c3e/jobs/6910/parallel-runs/0/steps/0-107) looks like a real error.

vadz:
Actually I have no idea what happened to `src/generic/bmpsvg.cpp` in the latest push, but it looks like a mistake, please check.

vadz:
There are now a lot of whitespace-only changes that hadn't been there before... Would it be possible to undo them please?

And the last commit should be squashed with a previous one, but I can do this myself if necessary.

Issue #26472, opened 2026-05-17
Fails to build on macOS < 10.12


mohd-akram:
### Build System Used

I build wxWidgets and/or my application using:

- [ ] CMake
- [x] configure
- [ ] mingw32-make with makefile.gcc
- [ ] MSBuild (Microsoft Visual Studio solution file)
- [ ] nmake with makefile.vc
- [ ] Xcode

### Description
wxWidgets uses some constants that are not available before 10.12. I get the following error when building on 10.10 and 10.11:

```
../src/osx/carbon/utilscocoa.mm:295:39: error: use of undeclared identifier 'NSCompositingOperationSourceOver'
        NSCompositingOperation mode = NSCompositingOperationSourceOver;
                                      ^
../src/osx/carbon/utilscocoa.mm:299:24: error: use of undeclared identifier 'NSCompositingOperationClear'
                mode = NSCompositingOperationClear;
                       ^
../src/osx/carbon/utilscocoa.mm:302:24: error: use of undeclared identifier 'NSCompositingOperationCopy'
                mode = NSCompositingOperationCopy;
                       ^
../src/osx/carbon/utilscocoa.mm:305:24: error: use of undeclared identifier 'NSCompositingOperationSourceOver'
                mode = NSCompositingOperationSourceOver;
                       ^
../src/osx/carbon/utilscocoa.mm:308:24: error: use of undeclared identifier 'NSCompositingOperationSourceIn'
                mode = NSCompositingOperationSourceIn;
                       ^
../src/osx/carbon/utilscocoa.mm:311:24: error: use of undeclared identifier 'NSCompositingOperationSourceOut'
                mode = NSCompositingOperationSourceOut;
                       ^
../src/osx/carbon/utilscocoa.mm:314:24: error: use of undeclared identifier 'NSCompositingOperationSourceAtop'
                mode = NSCompositingOperationSourceAtop;
                       ^
../src/osx/carbon/utilscocoa.mm:317:24: error: use of undeclared identifier 'NSCompositingOperationDestinationOver'; did you mean 'NSCompositeDestinationOver'?
                mode = NSCompositingOperationDestinationOver;
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                       NSCompositeDestinationOver
/System/Library/Frameworks/AppKit.framework/Headers/NSGraphics.h:27:5: note: 'NSCompositeDestinationOver' declared here
    NSCompositeDestinationOver, 
[...]

PR #26476, opened 2026-05-18
OSX 10.10 compatiblity for compositing operators

csomor:
see #26472


csomor:
@vadz if you're fine with this approach - moving the `#define`s to a private header, then I'll merge this

vadz:
Sure, why not. It just would be better to squash merge this to avoid touching `src/osx/carbon/utilscocoa.mm` needlessly. Please let me know if you'll do it or if I should. TIA!

csomor:
> Sure, why not. It just would be better to squash merge this to avoid touching `src/osx/carbon/utilscocoa.mm` needlessly. Please let me know if you'll do it or if I should. TIA!

Sure, I also thought about doing it that way, Thanks!

vadz:
Thanks! I'll cherry pick it to 3.2 to fix the original issue.

PR #26478, opened 2026-05-18
wxWebRequestSync Finalize() response (ensure data file is flushed)

### LINKED ISSUE #26477: wxWebRequestSync does not flush data file
#### Bug description:

Whenever wxWebRequestSync::Execute is called, if the data storage is set to be a file, the file is buffered, and the writes are never flushed before control goes back to the application, leading to getting (potentially) incomplete data when then reading the file.

The file is flushed and its whole contents appear when reading.

#### More information.

Reading through WinHTTP, cURL, and URLSession backends, it doesn't seem like any of them handle this properly. Afaics they should call `m_response->Finalize()` like `wxWebRequestImpl::SetState()` (and from `wxWebRequestImpl::HandleResult()`) does whenever the state is one of the three "end" states. (`State_Completed`, `State_Failed`, and `State_Cancelled`).

`Finalize()` is private to the `wxWebResponseImpl` though...

It seems correct that none of these implementations call `SetState()`, as it handles things for async mode (notably assuming things happening on the main thread, and `ProcessStateEvent()` deleting the file, assuming the event handler used it if it wanted).

It seems `Finalize()` should just be made public (rather than private and marking `wxWebRequestImpl` a friend), and called by each of these implementations' `Execute()`.

I'm happy to submit a PR if that sounds reasonable.

- wxWidgets version  you use: 3.3.3
- OS  and its version: macOS 15

ryancog:
Closes #26477

### DIFF
--- a/include/wx/private/webrequest.h
+++ b/include/wx/private/webrequest.h
@@ -270,6 +270,8 @@ public:
 
     void ReportDataReceived(size_t sizeReceived);
 
+    void Finalize();
+
 protected:
     wxWebRequestImpl& m_request;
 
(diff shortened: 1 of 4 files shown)


vadz:
Looks good, thanks! I'll merge it soon.

PR #26479, opened 2026-05-19
Add getters to wxSVGAttributes, add unit tests

Blake-Madden:
I was wanting to mulipurpose `wxSVGAttributes` in another project and noticed that I forgot the getters, sorry. Adding these also allows for unit tests now.


vadz:

Issue #26480, opened 2026-05-19
Non-resizable frame issues on wxGTK


hamishmb:
### Description
If a frame is non-resizable (eg has any style with a non-resizable border such as SIMPLE_BORDER), and SetClientSize() is called inside the construction while also specifying a frame size in the constructor call, the window has a 1x1 pixel size.

Previously discussed here, I just forgo to report it for ages: https://forums.wxwidgets.org/viewtopic.php?f=29&t=47747

#### Bug description:
As above.

#### Expected vs observed behaviour:

The window always has the desired size.

#### Stack trace:

N/A

#### Patch or snippet allowing to reproduce the problem:

See attached wxPython snippet.

[demo.py](https://github.com/user-attachments/files/28023395/demo.py)

#### To Reproduce:

Steps to reproduce the behaviour, please make them as detailed as possible.
For example:

1. Run the file and see the frame displays correctly.
2. Uncomment the SetClientSize line and observe that it now displays as 1x1 pixel.

### Platform and version information

- wxWidgets version  you use: 3.2.8
- wxWidgets port  you use: wxGTK
- OS  and its version: LMDE 7 (Debian 13)
  
  + GTK version:  3.24.49
  + Which GDK  backend is used: X11
  + Desktop environment : Cinnamon
  + Current theme:  N/A

May also affect other platforms but I haven't tested this.

vadz:
This is not X11-specific and still present in master because I can reproduce it under Sway using the following C++ translation of the reproducer:

```cpp
#include <wx/wx.h>

class InitialWindow : public wxFrame
{
public:
    InitialWindow()
        : wxFrame(nullptr, wxID_ANY, "Test",
                  wxDefaultPosition, wxSize(600, 420),
                  wxSIMPLE_BORDER)
    {
        SetClientSize(600, 420);
    }
};

class Test : public wxApp
{
public:
    bool OnInit() override
    {
        (new InitialWindow())->Show();
        return true;
    }
};

wxIMPLEMENT_APP(Test);
```

PR #26481, opened 2026-05-19
Fix non-resizable TLW size on GTK when client size is set to initial size

paulcor:
There are a number of ways to fix this, hopefully this one is pretty safe.

See #26480


vadz:
While this commit applies cleanly to 3.2, it doesn't fix the problem there, do you know what else is missing from that branch by chance?

paulcor:
Hmm. Works for me with 3.2, both X11 and Wayland.

vadz:
Sorry for wasting your time, I didn't rebuild the right library :-( I can confirm that it works for me too, I'll add it to the backports PR.

Issue #22192, opened 2022-03-11
Linux: wxAboutDialogInfo::SetIcon() can't use a hiDPI image 


willco007:
**Describe the bug**
When running on linux, you can't use a hiDPI image in the about dialog. 

**Expected vs observed behaviour**
SetIcon() takes a wxIcon and can't scale it based on DPI, it just doubles the size.  A possible solution would be to take an wxIconBundle and use the correct icon for the scale factor.

**Patch or snippet allowing to reproduce the problem**

On a hiDPI display, attempt to put a hi res icon into the about dialog on linux like so:

wxAboutDialogInfo aboutInfo;
aboutInfo.SetIcon(icon);

Skip this step if the problem can be reproduced in one of the samples without any changes.

**To Reproduce**
See above

**Platform and version information**
 - wxWidgets version you use: 3.1.5
 - wxWidgets port you use: wxGTK
 - OS and its version: Fedora 35
 - For wxGTK only:
   + GTK version: 3.24.31
   + Which GDK backend is used: Wayland

vadz:
@paulcor This doesn't seem GTK-specific, I just forgot about this class and didn't change its `SetIcon()` to take `wxBitmapBundle` yet.

I'll ask @kosh543 to look at this as this should be similar to the other recent changes.

paulcor:
Using `wxBitmapBundle` won't help for GTK, as `GtkAboutDialog` only accepts a `GdkPixbuf`. I don't think there is a way to use a HiDPI image with that.

vadz:
Sorry if I'm missing something, but couldn't we extract/create a `GdkPixbuf` of appropriate size and pass it to `gtk_about_dialog_set_logo()`?

In any case, I think we need to change the API (and implement support for high DPI images in the generic version) in 3.2.0, even if we can't fix this wxGTK right because perhaps we'll be able to do it later and at least like this it ought to be possible without breaking the ABI.

paulcor:
Of course you can make a `GdkPixbuf` any size you want, but it will be drawn *at that size*, it has no scaling information.

vadz:
[...]

PR #26484, opened 2026-05-20
Add support for wxBitmapBundle to wxAboutBox

paulcor:
See #22192

### DIFF
--- a/interface/wx/aboutdlg.h
+++ b/interface/wx/aboutdlg.h
@@ -179,6 +179,10 @@ public:
 
     /**
        Returns the icon set by SetIcon().
+
+       @param window
+           If non-null, this is used to help select an icon based on window DPI.
+           New since wxWidgets 3.3.3.
     */
     wxIcon GetIcon(wxWindow* window = nullptr) const;
 
--- a/interface/wx/aboutdlg.h
+++ b/interface/wx/aboutdlg.h
@@ -180,7 +180,7 @@ public:
     /**
        Returns the icon set by SetIcon().
     */
-    wxIcon GetIcon() const;
+    wxIcon GetIcon(wxWindow* window = nullptr) const;
 
     /**
         Set the icon to be shown in the dialog. By default the icon of the main frame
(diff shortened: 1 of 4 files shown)


interface/wx/aboutdlg.h
```diff
@@ -180,7 +180,7 @@ class wxAboutDialogInfo
     /**
        Returns the icon set by SetIcon().
     */
-    wxIcon GetIcon() const;
+    wxIcon GetIcon(wxWindow* window = nullptr) const;
```
vadz, line 187: We should document that `window` parameter is new in 3.3.3 and explain that it needs to be provided to get the icon at appropriate DPI.

src/gtk/aboutdlg.cpp
```diff
@@ -172,7 +172,7 @@ void wxAboutBox(const wxAboutDialogInfo& info, wxWindow* parent)
         gtk_about_dialog_set_license(dlg, nullptr);
 
     static wxIcon s_icon;
```
vadz, line 174: Not related to this PR, but I wonder how this works: this static object is destroyed when the display connection is already shut down, so why are we not getting any errors due to this? It must leak, at the very least...

vadz (commented):
Thanks for implementing this!

src/gtk/aboutdlg.cpp
paulcor: In the usual case, the `wxIcon` is `UnRef()`-ed in the "response" handler when the user closes the dialog, so the static dtor does nothing.

It is possible to close the app while the dialog is still up, and in that case the "response" event doesn't fire, so the static dtor will end up running the `wxBitmapRefData` dtor. But it doesn't seem to matter. `GdkPixbuf` doesn't care about the display, and `cairo_surface_t` doesn't seem to either. So there doesn't seem to be any error, and I don't think there is a leak. Also, `valgrind` says the entire `GtkAboutDialog` leaks in that case. Since no one has complained about that in 20 years, and there doesn't seem to be anything we can do about it anyway, I'm disinclined to be very concerned.

PR #26489, opened 2026-05-20
OSX restore local resource access to ArtProvider

csomor:
The former implementation (before #25860) also accessed app-local resources, this is now reinstantiated


vslavik:
Thanks! I confirm the fix works.

vadz:
I'll merge it then, unless Stefan prefers to do it himself/does it first.

PR #26494, opened 2026-05-21
MSW: Dark mode initialization

stevecor:
- One of the undocumented functions we use was changed in Windows 10 1903. Therefore the minimum Windows version for enabling dark mode is updated from v1809 to v1903. This issue is shown at https://github.com/ysc3839/win32-darkmode/blob/master/win32-darkmode/DarkMode.h - see the second occurrence of "g_buildNumber < 18362". I verified this issue by running a test on v1809 and v1903.

- The pointers to the undocumented functions are initialized so that they are never null. When the functions are not available, the pointers refer to stubs. This eliminates any possibility of dereferencing a null pointer. This simplifies future planned changes to allow dark mode switching after initialization.
 
- The declaration of SetPreferredAppMode() is updated to match the referenced source of information. The return type and argument type are changed from DWORD to PreferredAppMode.

### DIFF
--- a/src/msw/darkmode.cpp
+++ b/src/msw/darkmode.cpp
@@ -151,8 +151,10 @@ PreferredAppMode gs_appMode = AppMode_Default;
 template <typename T>
 bool TryLoadByOrd(T& func, const wxDynamicLibrary& lib, int ordinal)
 {
-    func = (T)::GetProcAddress(lib.GetLibHandle(), MAKEINTRESOURCEA(ordinal));
-    if ( !func )
+    T tmp = (T)::GetProcAddress(lib.GetLibHandle(), MAKEINTRESOURCEA(ordinal));
+    if ( tmp != nullptr )
+        func = tmp;
+    else
     {
         wxLogTrace(TRACE_DARKMODE,
                    "Required function with ordinal %d not found", ordinal);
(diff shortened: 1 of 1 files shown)


stevecor:

vadz:
Thanks, I've applied the version-related change as part of #26503, but I'm not sure about the stub functions idea, why do we want to do this? I'd rather keep function pointers as null if dark mode is not supported, at least this provides a simple way to check whether it is supported or not.

If we really, really want to be able to always call them, I'd prefer to make them real functions delegating to the Windows function pointer if it could be loaded and returning the fall back value if not.

stevecor:
> Thanks, I've applied the version-related change as part of #26503, but I'm not sure about the stub functions idea, why do we want to do this? I'd rather keep function pointers as null if dark mode is not supported, at least this provides a simple way to check whether it is supported or not.
> 
> If we really, really want to be able to always call them, I'd prefer to make them real functions delegating to the Windows function pointer if it could be loaded and returning the fall back value if not.

Just applying the version check is good.

I am working on improving switching between dark and light modes. I was planning to do this in small relatively independent steps. One step is to improve the way we call the undocumented functions so that it is obviously guaranteed that a null pointer dereference cannot occur, without having to consider the associated logic or state. For example, I plan to add calls upon system color change events.

I agree making real functions that delegate to function pointers is a better idea than the stubs. Do you want to hold all these changes for that? Please give me some guidance on whether you prefer smaller incremental changes or bigger all-in-one PRs.

vadz:
It is better to make smaller changes as big PRs may be difficult to review, but ideally each PR should still make sense in isolation, so (again, ideally) it would be best to avoid making a PR which just changes things in order to facilitate future PRs without any other rationale.

However in case of doubt, multiple too small PRs are still better than a single huge one.

TIA!

vadz:
I'll close this one for now then, with part of it applied due to the merge of the other PR, and part to be redone later.

PR #26495, opened 2026-05-22
Avoid deprecation warnings from latest MSVS 2026 in wxPG headers

vadz:
Apparently the compiler resolves the calls in the deprecated overloads as calls to the same overload in the base class instead of calling the non-deprecated overload in this one.

Work around this by using explicit "this->".

### DIFF
--- a/include/wx/propgrid/advprops.h
+++ b/include/wx/propgrid/advprops.h
@@ -210,7 +210,7 @@ public:
     virtual int GetCustomColourIndex() const;
 
 #if WXWIN_COMPATIBILITY_3_2
-    wxDEPRECATED_MSG("use ValueToString with 'flags' argument as wxPGPropValFormatFlags")
+    wxDEPRECATED_BUT_USED_INTERNALLY_MSG("use ValueToString with 'flags' argument as wxPGPropValFormatFlags")
     virtual wxString ValueToString(wxVariant& value, int flags) const override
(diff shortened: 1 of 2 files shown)


stevecor:
This Microsoft compiler behavior is intentional. Starting with compiler 19.51 / Visual Studio 18.6, overriding a deprecated virtual function issues C4996, even if there is no call.  An additional source of confusion in this case is that both the overriding and overridden functions have [[deprecated(msg)]] with the same msg argument. The message shown is not from the source location indicated but rather from the declaration in the base class.

I don't think using wxDEPRECATED_BUT_USED_INTERNALLY_MSG is the answer because this issue will start happening to end users when they upgrade their compiler version.

See also the reply from Microsoft here: https://developercommunity.visualstudio.com/t/MSVC-1451-toolset-emits-C4996-warning-f/11083608

vadz:
OK, thanks, this actually makes sense (even if the error message does not).

So we need to stop overriding it, but just removing the override would hide the base class overload. Could you please check, if you have this compiler, if it also gives the warning for a using declaration of the function? TIA!

stevecor:
Replacing the overriding functions with a using-declaration seems to solve the problem. There is no C4996 upon the using-declaration.

vadz:
Thanks, I've created #26502 doing it in this way.

PR #26496, opened 2026-05-22
Reject BMP RLE absolute runs that overrun the current scanline

dxbjavid:
### Bug

In `src/common/imagbmp.cpp` the absolute-mode branches of both the BI_RLE4 (around line 763) and BI_RLE8 (around line 870) decode paths inside `LoadBMPData()` increment `column` for each decoded pixel without checking it against `width`. The image buffer is sized `width * height * 3` and `poffset` is computed as

    line * width * 3 + column * 3

so an absolute escape with a count larger than `width - column` keeps writing 3 bytes per pixel through the rest of the row and into adjacent rows, eventually overrunning the heap buffer on the last decoded scanline (for the common bottom-up BMP that corresponds to `row == 0`, `line == height - 1`).

The neighbouring encoded-mode loops at `imagbmp.cpp:798` (RLE4) and `imagbmp.cpp:896` (RLE8) already clamp the run with `&& column < width`, and the delta-mode escapes at `imagbmp.cpp:758` and `imagbmp.cpp:863` reject out-of-range row offsets with `return false`. The two absolute-mode loops are the only RLE paths left without bounds checking.

### Reproducer

A minimal 4x4 8bpp RLE BMP with a single absolute escape `0x00 0x64` (count 100) followed by 100 colour-index bytes triggers an out-of-bounds write the very first row of decoded pixels: `column` is incremented 100 times for a width of 4, so 96 pixels (288 bytes) are written past the right edge of the scanline and through the end of the heap buffer.

The same pattern reproduces under `BI_RLE4`, where the count is also in pixels.

### Fix

Reject the file at the start of each absolute-mode block when `column + absolute > width`, matching the existing "return false on malformed input" style used by the surrounding delta and palette-index checks. The fix is local to the callee; no caller changes are needed.

### Test

`tests/image/image.cpp` gains a `SECTION("RLE absolute run past right edge")` under the existing `wxImage::BMP` test case, loading the 4x4/absolute-100 BMP described above and expecting `LoadFile` to return false.

### Build

`make wxcore` builds clean with the change applied (verified locally with CMake).


vadz:
Thanks a lot for fixing this! Will merge soon and also cherry pick to 3.2.

PR #26497, opened 2026-05-22
Validate IFF BMHD fields to prevent pixel-buffer overflow

jmestwa-coder:
## Bug

`wxIFFDecoder::ReadIFF()` in `src/common/imagiff.cpp` parses the BMHD chunk and stores width/height/bitplanes without any bounds check, then allocates the pixel buffer with signed-int multiplication:

```cpp
m_image->p = new byte[bmhd_width * bmhd_height * 3];
```

With `bmhd_width = 21849` and `bmhd_height = 65535` (both legal 16-bit BMHD values), the product is `4,295,622,645`. The `int * int * int` expression overflows and wraps down to `655,349`, so only ~640 KiB is allocated. The BODY decode loop a little further down (`src/common/imagiff.cpp` around the `ILBM_NORMAL`/`ILBM_EHB` branch) then writes `3 * bmhd_width` bytes per scanline, so a BODY chunk containing 10 lineskips' worth of zeros (`lineskip = 2732` for this width, total 27,320 bytes) is enough to overrun the heap allocation.

Two further malformed-input cases land in the same place:

* `bmhd_width == 0` makes `lineskip = ((bmhd_width + 15) >> 4) << 1 == 0`, so the subsequent `chunkLen / (lineskip * bmhd_bitplanes)` divides by zero.
* `bmhd_bitplanes == 0` does the same, and additionally interacts badly with `1 << bmhd_bitplanes` later on; values above ~30 push that shift into undefined-behaviour territory.

## Reproducer

The new `wxImage::BadIFFBmhdOverflow` test case in `tests/image/image.cpp` constructs the malformed IFF in memory:

```
FORM 27360
  ILBM
  BMHD 20    width=21849 height=65535 nPlanes=1 compression=0
  BODY 27320 (zeros)
```

Without the fix, running the test under the address sanitiser reports a heap-buffer-overflow at the write loop in `wxIFFDecoder::ReadIFF()`; under a non-sanitised build `LoadFile()` returns `true` (silently scribbling over adjacent heap memory), so the `REQUIRE(!img.LoadFile(...))` assertion fails. With the fix the file is rejected at BMHD parse time and the test passes.

## Fix

Reject malformed BMHD chunks at parse time. The new check is local to the BMHD branch in `wxIFFDecoder::ReadIFF()` and bails via the same `break` the rest of the parser uses:

```cpp
if (bmhd_width <= 0 || bmhd_height <= 0
        || bmhd_bitplanes <= 0 || bmhd_bitplanes > 32
        || static_cast<wxUint64>(bmhd_width)
            * static_cast<wxUint64>(bmhd_height)
                > static_cast<wxUint64>(INT_MAX) / 3) {
    break;
}
```

The `INT_MAX / 3` cap makes the existing `new byte[bmhd_width * bmhd_height * 3]` allocation safe without having to widen any of the surrounding int arithmetic; the cap of 32 bitplanes is well above the largest format the decoder actually handles (24-bit ILBM). `INT_MAX` comes from `<limits.h>`, which the file did not already include, so the patch adds the include alongside the existing `<stdlib.h>` / `<string.h>`.

## Test evidence

```
$ cmake --build . --target test_gui
$ ./lib/test_gui.app/Contents/MacOS/test_gui "[iff]"
All tests passed (2 assertions in 2 test cases)

$ ./lib/test_gui.app/Contents/MacOS/test_gui "[image]"
All tests passed (13077 assertions in 37 test cases)
```

The full `[image]` suite (37 cases / 13,077 assertions) continues to pass, including the existing `wxImage::BadIFF`, `wxImage::PNM` and BMP malformed-image tests.

### DIFF
--- a/src/common/imagiff.cpp
+++ b/src/common/imagiff.cpp
@@ -29,6 +29,7 @@
     #include "wx/palette.h"
 #endif // wxUSE_PALETTE
 
+#include <limits.h>     // for INT_MAX
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for finding and fixing this! Will merge soon and backport to 3.2 branch.

PR #26499, opened 2026-05-22
Stop reading past wxCharBuffer end on unterminated XPM quote

dxbjavid:
The quote-stripping loop in `wxXPMDecoder::ReadFile()` sets `p = q + 1`
after the strncpy. If the closing `"` is missing, `q` stops at the
buffer terminator, so `p` ends up one past it and the next outer
for-loop iteration's `p++` then dereferences two bytes off the end of
the wxCharBuffer. Mirror the existing handling of an unterminated
`/*`-comment and break out of the loop when `*q == '\0'`.


vadz:
Thanks for finding and fixing this! Will merge soon.

PR #26500, opened 2026-05-22
Fix user scaling in SVG sample

### LINKED ISSUE #26396: User scaling not working in wxSVGFileDC
This can be assigned to me, I'm looking at it now, but have another SVG PR pending at the moment.

In the SVG sample, save the "Scaled Text" tab as SVG and open it in a browser. Only a quarter of the content is visible. I believe it has to do with user scaling vs. logical units not being used in the drawing commands.

Blake-Madden:
Closes #26396


vadz:
Thanks for the fix, will merge soon.

PR #26501, opened 2026-05-23
Reject GIF files with LZW minimum code size > 11

dxbjavid:
The minimum code size byte that follows the local colour table in
`wxGIFDecoder::LoadGIF()` is only checked for `<= 0`. `dgif()` sizes
`ab_prefix`/`ab_tail` with `allocSize = 4096 + 1`, so a value of 12
starts `ab_free` at 4098 and the first alphabet update at
`gifdecod.cpp:457` writes one entry past the end of both arrays. The
existing `wxASSERT(ab_free < allocSize)` already flagged this in debug
builds. The new `wxImage::BadGIFLZWMinCodeSize` test feeds a 37-byte
2x1 GIF with code size 12 to `LoadFile` and asserts it is rejected.

### DIFF
--- a/src/common/gifdecod.cpp
+++ b/src/common/gifdecod.cpp
@@ -859,7 +859,11 @@ wxGIFErrorCode wxGIFDecoder::LoadGIF(wxInputStream& stream)
 
                 // get initial code size from first byte in raster data
                 bits = stream.GetC();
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, I'll apply this soon and will also fix the memory leak that the new test exposed.

PR #26504, opened 2026-05-24
Add & fix SVG Graphics Context unit tests

### LINKED ISSUE #26486: Add unit tests for new wxSvgGraphicsContext

Blake-Madden:
Basically, a total overhaul of the `wxSVGFileDC` unit tests. I was getting failing tests with `test_drawing` and the image test case would only work on Windows. I did a rewrite from CppUnit to Catch2 and removed the external file dependencies. I also removed the "plugin" test interface in here. It seems to have never been used and is orphaned in numerous places. I removed it from here; let me know if you want me to remove the remaining traces of this.

Closes #26486


vadz:
Just checking: was this closed intentionally?

Blake-Madden:
Yes, I was getting a bunch of warnings as errors from CppUnit in other files in some of the build configurations here. No idea why though, was coming from files I hadn't touched. Catch2'ifying the `test_drawing` target broke something somehow, so I'll either have to rethink this (or finish removing CppUnit from everyone else first).

vadz:
Looking at the first failing build, the errors are due to `wxBrush` being incomplete, almost certainly due to the fact that `wx/brush.h` was indirectly included before but needs to be explicitly included now.

vadz:
I'm not sure about the "plugin" tests but I'd definitely prefer to remove it one independent commit before making the SVG-related changes. Could you do it like this please?

Blake-Madden:
> I'm not sure about the "plugin" tests but I'd definitely prefer to remove it one independent commit before making the SVG-related changes. Could you do it like this please?

See #26568

vadz:
This has conflicts now, do you plan on resolving them or should I (try to) do it?

Blake-Madden:
I'll take a look at it tonight

Blake-Madden:
Removed all CppUnit and old image comparison content. Tests pass for me now, although it doesn't appear that our CI here is running `test_drawing`. Not sure if that is wanted or not, but probably just a matter of adding it to the matrix in `ci.yml`.

src/common/dcsvg.cpp
```diff
@@ -1525,11 +1526,12 @@ void wxSVGFileDCImpl::DoStartNewGraphics()
     if ( penStroke.empty() )
         penStroke = GetPenStroke(m_pen.GetColour(), m_pen.GetStyle());
 
-    s = wxString::Format(wxS("<g %s %s %s %s transform=\"translate(%d %d) scale(%s %s)\">\n"),
+    s = wxString::Format(wxS("<g %s %s %s %s %s transform=\"translate(%d %d) scale(%s %s)\">\n"),
```
vadz, line 1528: I'm not sure how this works, but it looks like we might be outputting `shape-rendering="auto"` here unnecessarily, shouldn't we only add this if the rendering mode is different from default?

vadz (commented):
There are a lot of whitespace only and otherwise insignificant (e.g. additions of `wxASCII_STR`, removal of comments, ...) changes here so it's really difficult to see what has changed, i.e. what exactly has been fixed?

We do need to run these tests if we want to preserve/keep them going forward, so we should do it in at least some CI jobs, could you please add running them there?

src/common/dcsvg.cpp
Blake-Madden: fixed

Blake-Madden:
I added `test_drawing` to the CI runs.

Sorry, this was more of a total rewrite than changing things. I don't think it can be meaningfully compared to the old test harness, but rather reviewed as all new code.

PR #26505, opened 2026-05-24
Stop reading past dataend in truncated IFF BODY decode

dxbjavid:
wxIFFDecoder::ReadIFF() at imagiff.cpp:471 sets `chunkLen = dataend - dataptr` on the truncated BODY path, which is 8 too large because bodyptr lives 8 bytes after dataptr. The non-RLE decode loop below then reads up to 8 bytes past the end of databuf; clamp chunkLen to `dataend - bodyptr` instead.


vadz:
Thanks! Tested and confirmed, will merge soon and backport to 3.2 too.

PR #26506, opened 2026-05-24
validate data chunk header room in wxSound::LoadWAV

dxbjavid:
Noticed LoadWAV in src/unix/sound.cpp checks that the LIST chunk
length fits in the buffer but not that the 8-byte "data" chunk
header still has room afterwards. A 44-byte WAV with a LIST chunk
of length 0 lands data_offset at exactly length, and the next
memcmp(&data[data_offset], "data", 4) reads 4 bytes past the heap
buffer (also the memcpy of the data-chunk size four lines later).


vadz:
Looks good, thanks!

PR #26507, opened 2026-05-25
Reject too-short ZIP64 extra field in wxZipEntry::LoadExtraInfo

dxbjavid:
Noticed wxZipEntry::LoadExtraInfo() calls wxZipHeader::Read64() up to three
times on a wxZipHeader of length min(fieldLen, 28). Read64() doesn't
bounds-check m_pos against m_size, so a short ZIP64 extra field returns
uninitialised bytes from the header's 64-byte stack-allocated m_data and
they end up in the entry's m_Size / m_CompressedSize / m_Offset. Reject the
entry when fieldLen is below the requested 64-bit total.


vadz:
Thanks for the fix and the test!

PR #26508, opened 2026-05-25
MSW Cleanup

Blake-Madden:
A few buffer fixes and extra guardrails in MSW code

### DIFF
--- a/src/msw/darkmode.cpp
+++ b/src/msw/darkmode.cpp
@@ -682,7 +682,7 @@ HandleMenuMessage(WXLRESULT* result,
                 WinStruct<MENUITEMINFO> mii;
                 mii.fMask = MIIM_STRING;
                 mii.dwTypeData = buf;
-                mii.cch = sizeof(buf) - 1;
+                mii.cch = WXSIZEOF(buf);
 
                 // Note that we need to use the iPosition field of the
                 // undocumented struct here because DRAWITEMSTRUCT::itemID is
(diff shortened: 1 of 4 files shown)


vadz:
Nice, thanks, will merge in a moment.

I wonder if we shouldn't introduce a `wxStrlcpy()` overload for `char[]`/`wchar_t[]`, as this would avoid the need to pass the buffer size to it implicitly, what do you think?

Blake-Madden:
> I wonder if we shouldn't introduce a `wxStrlcpy()` overload for `char[]`/`wchar_t[]`, as this would avoid the need to pass the buffer size to it implicitly, what do you think?

Seems like a really big change, and I'm not sure how often people are calling these low level functions. I'm OK with keeping this as is myself. But if you want to add that to the API, I'd be happy to apply it to various places.

vadz:
Sorry, maybe I didn't explain what I proposed well enough because I don't think it's a big change at all: I just wanted to add an extra overload (well, 2, for different kinds of characters) taking 2 parameters, without touching the existing ones. E.g. something like

```cpp
    template <size_t N>
    inline size_t wxStrlcpy(char (&dest)[N], const char *src)
    {
        return wxStrlcpy(dest, src, N);
    }
```

Blake-Madden:
> Sorry, maybe I didn't explain what I proposed well enough because I don't think it's a big change at all: I just wanted to add an extra overload (well, 2, for different kinds of characters) taking 2 parameters, without touching the existing ones. E.g. something like
> 
> ```c++
>     template <size_t N>
>     inline size_t wxStrlcpy(char (&dest)[N], const char *src)
>     {
>         return wxStrlcpy(dest, src, N);
>     }
> ```

I see. Yes, that would be pretty cool :)

PR #26509, opened 2026-05-25
MSW: Remove wxUSE_UXTHEME

stevecor:
Remove configuration option wxUSE_UXTHEME. It has not been possible to build with this option disabled for 4 years. It appears there has been little effort to support this option since then. For 8 years, there has been a duplicate definition of wxUxThemeIsActive(), causing a linker warning.


src/msw/anybutton.cpp
```diff
@@ -52,30 +52,6 @@
...
-            BUTTON_IMAGELIST_ALIGN_BOTTOM
-        };
-
-        struct BUTTON_IMAGELIST
-        {
-            HIMAGELIST himl;
-            RECT margin;
-            UINT uAlign;
-        };
-    #endif
```
vadz, line 76: Removal of those seems to be unrelated to `wxUSE_UXTHEME` itself. It's probably fine to remove them anyhow, but I'll split it into a separate commit when applying.

vadz (approved):
Thanks for the cleanup! I agree that it's not needed since a long time and will merge this soon.

src/msw/anybutton.cpp
stevecor: Those names exist at least since the Vista SDK, the oldest I have handy.

PR #26511, opened 2026-05-26
validate 4-bit BMP palette index in LoadBMPData

dxbjavid:
The 8-bit decode path in LoadBMPData already rejects a pixel value that isn't below ncolors before indexing cmap, but the three 4-bit branches index cmap with the raw nibble (0 to 15) and never compare it against the palette size. A BMP that sets biClrUsed below 16 yet stores higher nibble values then reads past the cmap array; with biClrUsed = 1 the index can reach 15 against a one-entry palette. Apply the same ncolors check the 8-bit path uses.


vadz:
Looks good, thanks again!

PR #26512, opened 2026-05-26
Add wxSizer::DetachItem()

ryancog:
Added the `wxSizer::DetachItem()` function described in the linked topic, added documentation for it, and updated `Detach(int)`'s documentation as mentioned.

https://groups.google.com/g/wx-users/c/5mlHwIREgTU/m/nBtGfAU6AwAJ

### DIFF
--- a/interface/wx/sizer.h
+++ b/interface/wx/sizer.h
@@ -440,15 +440,29 @@ public:
     virtual bool Detach(wxSizer* sizer);
 
     /**
-        Detach an item at position @a index from the sizer without destroying it.
+        Detach the child sizer or window in item at position @a index without
+        destroying it.
 
         This method does not cause any layout or resizing to take place, call Layout()
         to update the layout "on screen" after detaching a child from the sizer.
+
         Returns @true if the child item was found and detached, @false otherwise.
 
         @see Remove()
     */
     virtual bool Detach(int index);
+
+    /**
+        Detach the item at position @a index without destroying it.
+
+        This method does not cause any layout or resizing to take place, call Layout()
+        to update the layout "on screen" after detaching a child from the sizer.
+
+        Returns the item if it was found and detached, @nullptr otherwise.
+
+        @see Remove(), Add()
+     */
+    virtual wxSizerItem *DetachItem(int index);
 
     /**
         Tell the sizer to resize the @a window so that its client area matches the
--- a/include/wx/sizer.h
+++ b/include/wx/sizer.h
@@ -617,6 +617,8 @@ public:
     virtual bool Detach( wxSizer *sizer );
     virtual bool Detach( int index );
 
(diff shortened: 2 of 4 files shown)


ryancog:
Had a feeling that C++ syntax was too new, updated.

vadz:
Thanks, but it would be really nice to have a unit test for the new function, checking that it works correctly even in the build with ASAN. Could you please add something?

ryancog:
@vadz I looked but didn’t see any unit tests for sizer stuff. Maybe I didn’t look hard enough, but if there aren’t any, I’m not sure what all you’d want tests to do with it.

Something like an add/detach/add with a sizer, spacer, and window?

vadz:
Hmm, there is `tests/sizers/boxsizer.cpp`, what's wrong with adding a test case there?

ryancog:
@vadz Evidently I looked poorly! :sweat_smile:

Is something like this case reasonable?

interface/wx/sizer.h
```diff
@@ -440,16 +440,30 @@ class wxSizer : public wxObject
...
+        Detach the item at position @a index without destroying it.
+
+        This method does not cause any layout or resizing to take place, call Layout()
+        to update the layout "on screen" after detaching a child from the sizer.
+
+        Returns the item if it was found and detached, @nullptr otherwise.
+
+        @see Remove(), Add()
+     */
+    virtual wxSizerItem *DetachItem(int index);
```
PBfordev, line 465:
```suggestion
    virtual wxSizerItem *DetachItem(size_t index);
```

The type is different from `Detach()`, for some reason both `int` and `size_t` are used for index in the `wxSizer` API?

vadz: It's a mess :-( I would prefer to standardize on `int`, but many of the functions using `index` are virtual so changing their parameter type would require adding compatibility overloads to allow the existing code overriding these functions to compile and this just doesn't seem to be worth it.

And most of the functions use `size_t index`, so I guess it's better to use it for the new function too for consistency.

vadz:

interface/wx/sizer.h
ryancog: @vadz, right, I figured since the Detach function that takes an `int index` casts it to `size_t` anyways, and some of the "fundamental" operations like `Insert` take a `size_t`, that `size_t` was the correct choice.

@PBfordev though I didn't realize I had a typo in the interface file (Since I copied from `Detach()`), thanks for bringing attention to it, in any case.

~~`DetachItem()` does take a `size_t`. Could you fix that @vadz? I can open another PR but I figure it's a quick fix. Sorry about that.~~

I see you already did in the actual commit, thanks :)

vadz: I did fix this in the committed version.

PR #26513, opened 2026-05-26
avoid integer overflow in message catalog StringAtOfs() bounds check

dxbjavid:
StringAtOfs() reads ofsString and nLen from the .mo file and checks ofsString + nLen against the data length, but both are 32-bit so the sum wraps and a translated entry declaring nLen 0xffffffff passes the check; FillHash() then scans past the end of the catalog. Compare each value against the length separately. The added test feeds such a catalog through wxMsgCatalog::CreateFromData().

### DIFF
--- a/src/common/translation.cpp
+++ b/src/common/translation.cpp
@@ -873,7 +873,7 @@ private:
         const wxMsgTableEntry * const ent = pTable + n;
 
         // this check could fail for a corrupt message catalog
-        size_t32 ofsString = Swap(ent->ofsString);
+        wxULongLong_t ofsString = Swap(ent->ofsString);
         if ( ofsString + Swap(ent->nLen) > m_data.length())
         {
             return nullptr;
--- a/tests/intl/intltest.cpp
+++ b/tests/intl/intltest.cpp
@@ -18,6 +18,7 @@
 #endif // WX_PRECOMP
 
 #include "wx/intl.h"
+#include "wx/translation.h"
 #include "wx/uilocale.h"
 #include "wx/scopeguard.h"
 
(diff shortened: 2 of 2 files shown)


vadz:
There seem to be some real issues in the CI builds, could you please have a look at them and fix them? TIA!

dxbjavid:
I checked the CI failures locally. The failing jobs appear to be unrelated network/webrequest tests. I rebuilt wxWidgets, ran the locale/message catalog tests (including the new malformed catalog coverage), and they pass successfully on my side.

vadz:
Please look more carefully at the failing checks above, the ASAN builds fail due to an actual error and MSW cross-builds fail due to a compilation warning (which may or not be spurious, I didn't check, but has to be avoided anyhow) and don't have anything to do with webrequest.

dxbjavid:
Thanks, I checked the ASAN and MSW cross-build failures more carefully. The issue was caused by the bounds-check arithmetic still triggering analyzer/warning complaints, so I updated the validation logic to avoid mixed-width arithmetic and problematic subtraction patterns and pushed an updated patch. I also verified the updated code with an ASAN build locally.

paulcor:
Simpler fix:
```diff
diff --git a/src/common/translation.cpp b/src/common/translation.cpp
index c5b3889949..80c027da2d 100644
--- a/src/common/translation.cpp
+++ b/src/common/translation.cpp
@@ -873,7 +873,7 @@ private:
         const wxMsgTableEntry * const ent = pTable + n;
 
         // this check could fail for a corrupt message catalog
-        size_t32 ofsString = Swap(ent->ofsString);
+        wxULongLong_t ofsString = Swap(ent->ofsString);
         if ( ofsString + Swap(ent->nLen) > m_data.length())
         {
             return nullptr;
```

dxbjavid:
Both CI failures came from the test, not the fix: the `mo` array was one byte too small, and wxCharTypeBuffer copies len+1 bytes, so the ASAN and the MSW -Werror=array-bounds errors were both reading one past the end. Gave the backing array the extra byte.

Also took @paulcor's simpler approach for the fix itself, just computing the offset as 64-bit so the addition can't wrap, and squashed it down to one commit. CI is green locally now.

vadz:
Thanks for the update, looks good now — will merge soon.

PR #26517, opened 2026-05-27
Fix out-of-bounds table read in wxMBConvUTF7::ToWChar

dxbjavid:
In wxMBConvUTF7::ToWChar the byte after '+' is looked up as utf7unb64[(unsigned)*src] at strconv.cpp:705, so with signed char a byte >= 0x80 sign-extends to a ~4 GiB index into the 256-entry table. Mask it with (unsigned char), like the cc lookup just above already does.


vadz:
Good find, thanks! Will merge soon.

PR #26518, opened 2026-05-27
fix one-byte over-read in IFF decomprle replicate packet

dxbjavid:
Noticed `decomprle()` in imagiff.cpp reads a replicate packet's data byte after only checking that one source byte remains, though it consumes two: the control byte plus the data byte. A run-length encoded BODY ending in a lone replicate control byte then reads one byte past the heap-allocated input. Require two bytes left, matching the literal-run check just above.

The added test feeds a 1x8 compressed ILBM whose BODY is a single 0xFF byte; under ASAN it reports a heap-buffer-overflow read without the fix.


vadz:
Great, thanks for fixing this!

PR #26519, opened 2026-05-27
fix width*chars_per_pixel overflow in wxXPMDecoder::ReadData

dxbjavid:
Noticed wxXPMDecoder::ReadData() checks each image line with `strlen(xpmImgLine) < width*chars_per_pixel`, but the product is 32-bit unsigned while the per-pixel index uses size_t. A header like `68174085 1 1 63` wraps the product to 59, so a one-pixel line passes the check and the key-reading loop runs off the end of the buffer.

The added test feeds such an XPM and confirms it's now rejected; under ASAN it shows the over-read without the fix.


vadz:
Thanks for fixing this! The fix looks correct and I'll merge it soon, but I had to add `/* XPM */` to the test to make it trigger the problem, I'm not sure how could it work for you without it.

PR #26520, opened 2026-05-27
wxURI::IsReference Update Description

ryancog:
The documentation currently describes the inverse of the name of the function and what it actually does: `return !HasScheme() || !HasServer();`.

This PR flips `true` and `false` to make the documentation match the behavior.


vadz:

PR #26521, opened 2026-05-28
validate decoded pixel index against ncolours in wxGIFDecoder::ConvertToImage

dxbjavid:
wxGIFDecoder::ConvertToImage() in src/common/gifdecod.cpp reads pal[3*(*src) + ...] for each decoded pixel, but per-frame pal is malloc'd at a fixed 768 bytes and only 3*ncolours of them are populated from the file. A malformed GIF whose LZW minimum code size implies more literals than its declared colour table (e.g. min code size 3 with a 2-entry global colour table) can emit literal codes >= ncolours, and those reads then return uninitialised heap bytes into the wxImage RGB data. The patch validates each pixel against ncolours up front and rejects the frame, matching the pattern from #26511 and the recent imagiff fix.

### DIFF
--- a/src/common/gifdecod.cpp
+++ b/src/common/gifdecod.cpp
@@ -154,10 +154,13 @@ bool wxGIFDecoder::ConvertToImage(unsigned int frame, wxImage *image) const
     // Reject decoded pixels that reference palette entries past the
     // loaded portion of the per-frame palette buffer: GIF files where
     // the LZW minimum code size implies a larger alphabet than the
-    // declared colour table can emit literal codes >= ncolours, and
-    // pimg->pal is malloc'd at a fixed 768 bytes with only 3*ncolours
-    // of them populated from the file, so the pal[3*(*src) + ...] reads
-    // below would otherwise yield uninitialised heap bytes.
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, but I don't see any error from ASAN even without the fix, shouldn't the new test case provoke one?

dxbjavid:
You're right, ASAN stays quiet here. The pixel index is a single byte, so the worst case is 255 and 3*255+2 is 767, which still lands inside the 768-byte pal buffer, so there's no out-of-bounds read. The actual problem is that only 3*ncolours of those bytes get populated from the file (the rest of the local pal[768] is uninitialised stack and gets memcpy'd into pimg->pal as-is), so a pixel index >= ncolours pulls uninitialised bytes into the output image. That's an MSAN-class uninitialised read rather than something ASAN catches, so the test guards against leaking that memory into the decoded pixels, not against a crash.

I've reworded the commit and the code/test comments to say that clearly, since the old note about ASAN was misleading.

vadz:
Makes sense, thanks! Will merge soon.

PR #26522, opened 2026-05-28
fix OOB write in wxVsnprintf and ConvertStringToBuf when size==0

dxbjavid:
The trailing \`str[size - 1] = 0\` in the char-overload of wxVsnprintf()
in src/common/wxcrt.cpp:608 runs unconditionally, so calling
wxSnprintf(buf, 0, "...") writes to str[SIZE_MAX]. The same off-by-one
exists in both ConvertStringToBuf() helpers, whose else branch does
memcpy(out, buf, outsize-1) with outsize == 0. Guard each on a non-zero
size, matching what the wchar_t wxVsnprintf already does just below.

### DIFF
--- a/src/common/wxcrt.cpp
+++ b/src/common/wxcrt.cpp
@@ -516,7 +516,7 @@ int ConvertStringToBuf(const wxString& s, char *out, size_t outsize)
     {
         memcpy(out, buf, len+1);
     }
-    else // not enough space
+    else if ( outsize > 0 ) // not enough space
     {
         memcpy(out, buf, outsize-1);
         out[outsize-1] = '\0';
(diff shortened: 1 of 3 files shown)


vadz:
Thanks, the fix looks correct to me, but somehow the new test fails when using MinGW. I don't understand what's wrong here but this needs to be fixed somehow...

dxbjavid:
Tracked it down. On MinGW we don't use the system vsnprintf but our own wxDoVsnprintf() in src/common/wxprintf.cpp (because of positional parameter support), and that one doesn't handle lenMax == 0 either. With an empty buffer there are no specifiers to process, so it skips to the trailing wxCopyStrWithPercents() which returns 0, then does `lenCur += 0 - 1` underflowing lenCur to SIZE_MAX, and immediately reads/writes buf[lenCur]. On 32-bit that wraps to buf - sizeof(CharType), i.e. one element *before* the buffer, which is why the wide wxSnprintf(bufw, 0, ...) call ended up zeroing the last two bytes of the adjacent bufa. glibc/macOS use the system implementation so they never hit it.

Added an early `if (lenMax == 0) return -1;` at the top of wxDoVsnprintf(). Test should pass on MinGW now.

vadz:
> Tracked it down.

Oh, wow, thanks!

PR #26524, opened 2026-05-28
reject GIF frames with zero width or height

dxbjavid:
wxGIFDecoder::LoadGIF in src/common/gifdecod.cpp reads each frame's width and height from the image descriptor and computes size = w * h before malloc(size). The subsequent-frames branch above rejects zero dimensions, but the first frame of an animation and every GIF87a frame reach the allocation unchecked. When either dimension is zero, the resulting empty allocation is still written to by the dgif() decode loop as soon as the LZW stream emits a literal code. The same check is now applied to all frames before the allocation. A new wxImage::BadGIFZeroFrameSize test feeds a minimal 35-byte GIF87a with a 1x0 frame and confirms the loader rejects it instead of overflowing the heap.

### DIFF
--- a/src/common/gifdecod.cpp
+++ b/src/common/gifdecod.cpp
@@ -836,6 +836,14 @@ wxGIFErrorCode wxGIFDecoder::LoadGIF(wxInputStream& stream)
                     }
                 }
 
+                // Reject frames with zero width or height: malloc(0) below
+                // returns an empty allocation that the dgif() decode loop
+                // still writes into. The subsequent-frames check above
+                // rejects zero size in animations but the first frame and
+                // the non-animated GIF87a path both reach here unchecked.
+                if (pimg->w == 0 || pimg->h == 0)
+                    return wxGIF_INVFORMAT;
+
                 interl = ((buf[8] & 0x40)? 1 : 0);
                 size = pimg->w * pimg->h;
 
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, the fix here also looks correct to me, but I can't reproduce the problem (dereferencing the invalid pointer to an empty allocation): the test just fails for me with

```
tests/image/image.cpp:1443: FAILED:
  REQUIRE( !img.LoadFile(mis, wxBITMAP_TYPE_GIF) )
  must create the image before setting its type in SetType() at src/common/image.cpp:1973
```

without any other errors. So does this pointer actually gets used in this case?

dxbjavid:
Yes, the pointer does get written. With this input dgif() reads the CLEAR code, then a literal code 0, and at the `(img->p)[x + (y * (img->w))] = ...` line it writes that byte to `img->p[0]` even though `size` was 0 and the malloc was malloc(0).

The reason it's not visible for you is that malloc(0) hands back a small usable block on most allocators, so the single stray byte doesn't corrupt anything you'd notice and the decode returns OK. I only caught it under ASan, which flags the write past the 0-byte allocation. On a normal build the frame decodes "successfully" as 1x0 and the first thing that actually fails is ConvertToImage doing Create(1, 0), which is the SetType assert you're hitting.

So it's a real but quiet OOB write; rejecting the frame before the allocation avoids both the write and the later assert.

vadz:
Weirdly, neither ASAN nor valgrind detect it here, although looking at the code I do see that

```cpp
            (img->p)[x + (y * (img->w))] = (char) stack[pos];
```

line writes beyond the end of the buffer. I'll apply this but I wonder what am I missing here.

Another thing which bothers me is that it seems it would be better to rewrite the loop in `dgif()` so that `x < img->w` is always true, whether `w` is 0 or not. But this doesn't seem that simple to do.

PR #26525, opened 2026-05-28
fix data-size bound check past LIST chunk in wxSound::LoadWAV

dxbjavid:
The data-size check in wxSound::LoadWAV compares ul against length minus the fixed offset FMT_INDEX + uiSize + 16, which equals data_offset + 8 only when no LIST chunk is present. When a LIST chunk is present data_offset has already been advanced past it, so a crafted WAV can pass this check with ul larger than the bytes that actually follow the data-chunk header, and the OSS/SDL playback paths then read past m_dataWithHeader. Use data_offset directly so the same constraint holds either way.

### DIFF
--- a/src/common/gifdecod.cpp
+++ b/src/common/gifdecod.cpp
@@ -836,6 +836,14 @@ wxGIFErrorCode wxGIFDecoder::LoadGIF(wxInputStream& stream)
                     }
                 }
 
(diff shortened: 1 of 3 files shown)


vadz:
Thanks, this looks correct to me, so I'll apply it (only the second commit, the first one is included in another PR) even without the test.

PR #26526, opened 2026-05-28
fix off-by-one buffer overflow in wxPathOnly()

dxbjavid:
ASan reports a 1-byte global-buffer-overflow on `static wxChar buf[_MAXPATHLEN]` in src/common/filefn.cpp:278, and the same overflow on the stack `wxChar buf[_MAXPATHLEN]` at :327.

The length check rejected only when `i = strlen(path) - 1 >= _MAXPATHLEN`, so a path of length exactly `_MAXPATHLEN` passed and `wxStrcpy()` wrote `_MAXPATHLEN + 1` wxChars (including the NUL) into a `_MAXPATHLEN`-element buffer.

### DIFF
--- a/src/common/filefn.cpp
+++ b/src/common/filefn.cpp
@@ -261,105 +261,10 @@ wxString wxFileNameFromPath (const wxString& path)
     return wxFileName(path).GetFullName();
 }
 
-// Return just the directory, or nullptr if no directory
-wxChar *
-wxPathOnly (wxChar *path)
-{
-    if (path && *path)
-    {
-        static wxChar buf[_MAXPATHLEN];
-
-        int l = wxStrlen(path);
-        int i = l - 1;
-        if ( l >= _MAXPATHLEN )
-            return nullptr;
-
-        // Local copy
-        wxStrcpy (buf, path);
-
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, we clearly have a problem here and the fix looks correct, but I'd prefer to

1. Get rid of `wxPathOnly(wxChar *path)` overload which is not used nor declared/exported.
2. Rewrite the other overload using `wxFileName`.

as there is really no need to keep this code at all. Please let me know if you'd like to do this, otherwise I'll do it myself a bit later. Thanks!

dxbjavid:
Makes sense, done. Dropped the wxChar* overload (nothing uses it and it's not exported) and the wxString one is now just wxFileName(path).GetPath(), matching wxFileNameFromPath right above it. Kept the long-path test as a regression guard.

vadz:
Sorry, I should have mentioned that it's not that simple (otherwise I would have just done it immediately) as the function has peculiar behaviour for paths with drive letters under Windows which we need to preserve for compatibility.

I've created #26537 doing this now, which replaces this PR, so I'm closing it.

PR #26528, opened 2026-05-28
fix off-by-one buffer overflow in deepCopyHostent name copy

dxbjavid:
The name copy in deepCopyHostent() checks `len > size` but then writes buffer[len], so a hostname whose length equals the buffer size overflows it by one byte. deepCopyServent() just below already uses the correct `len >= size`.

ASan on a reduced reproducer of the name copy:

    heap-buffer-overflow WRITE of size 1
    0 bytes after 16-byte region ... in copy_name


vadz:
Looks good, thanks!

PR #26530, opened 2026-05-29
fix integer overflow in tar extended header bounds check

dxbjavid:
Found while fuzzing the tar reader. recSize in wxTarInputStream::ReadExtendedHeader() comes from the record's decimal byte-count field, so a pax header declaring a length close to SIZE_MAX makes recPos + recSize wrap below len and pass the check. The following pRec[recSize - 1] then reads far past the buffer, which ASAN flags as a heap-buffer-overflow. Comparing recSize against len - recPos avoids the wrap. Test added in tests/archive/tartest.cpp.


vadz:
Thanks for the fix and the test, will merge soon.

PR #26531, opened 2026-05-29
fix out-of-bounds read on trailing % in ExpandCommand

dxbjavid:
Found this fuzzing wxFileType::ExpandCommand with mailcap-style strings. A command ending in a bare '%' makes the loop advance pc onto the terminating NUL inside the switch, then the for-loop increment steps one byte past the end of the c_str() buffer and the next condition check reads it (ASAN reports a heap-buffer-overflow read). If that byte is non-zero the loop keeps going and appends adjacent memory into the expanded command. Handle the trailing '%' before the increment so it is just copied verbatim.


vadz:
Thanks, I am applying the fix in a slightly different form which seems a bit more clear tyo me.

PR #26532, opened 2026-05-29
fix palette buffer overflow in wxGIFHandler_GetPalette

dxbjavid:
Found by round-tripping XPM files through GIF save: an XPM can declare more than 256 colours (xpmdecod.cpp), so the wxImage palette can exceed the fixed 256-entry wxRGB pal[] buffer that wxGIFHandler_GetPalette() fills, overflowing the stack. The over-large count is also returned and reused by WritePalette() later, reading past the same buffer. Cap palCount at 256, which is all a GIF can hold.

### DIFF
--- a/src/common/imaggif.cpp
+++ b/src/common/imaggif.cpp
@@ -554,7 +554,7 @@ bool wxGIFHandler_GetPalette(const wxImage& image,
     }
 
     const wxPalette& palette = image.GetPalette();
-    const int palCount = palette.GetColoursCount();
+    int palCount = palette.GetColoursCount();
 
     // The caller's pal[] buffer only has room for the 256 entries a GIF can
     // hold. A wxImage palette may be larger, e.g. when the image was loaded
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for finding this, but I'm not sure it makes sense to truncate the palette size like this because what happens to the pixels using the palette entries > 256 in this case?

I think it would be better to fail saving the image, after logging a message, wouldn't it?

dxbjavid:
Good point. Simply truncating the palette avoids the overflow, but it doesn't handle images that actually use palette entries beyond the GIF limit, so the resulting file could be incorrect.

I'll update the patch to reject saving images with more than 256 palette entries and log an error message instead. Since GIF only supports up to 256 colours, failing early seems safer than silently producing an invalid or corrupted output.

dxbjavid:
Updated. It now logs an error and fails the save when the palette has more than 256 entries, before anything is written to pal[], rather than truncating. Test checks that SaveFile() returns false in that case.

PR #26536, opened 2026-05-31
fix stack buffer overflow in TGA DecodeRLE for oversized bpp

dxbjavid:
DecodeRLE() reads each RLE run's pixel into a 4-byte stack buffer with stream.Read(buf, pixelSize), but pixelSize is bpp/8 from an unchecked header byte. An RLE TGA (image type 9/10/11) declaring bpp >= 40 makes pixelSize 5..31 and overruns buf; ReadTGA() only rejects unsupported depths in the switch that runs after decoding, so the read happens first. A 1x1 RLE-RGB file with bpp=64 smashes the stack. Reject pixelSize larger than buf at the top of DecodeRLE, which is also where the unsupported depth would be rejected anyway.

### DIFF
--- a/src/common/imagtga.cpp
+++ b/src/common/imagtga.cpp
@@ -106,6 +106,12 @@ int DecodeRLE(unsigned char* imageData, unsigned long imageSize,
     unsigned int length;
     unsigned char buf[4];
 
+    // A pixel is at most 4 bytes (32 bpp); a larger pixelSize comes from a
+    // corrupt bpp field in the header and would overflow buf when reading an
+    // RLE run below.
+    if ( pixelSize > (short)sizeof(buf) )
+        return wxTGA_INVFORMAT;
+
     while (outputLength < imageSize)
     {
         int ch = stream.GetC();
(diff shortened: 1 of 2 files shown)


vadz:
The fix looks good but the reproducer doesn't work for me and it's simple to see why: `wxTGAHandler::DoCanRead()` [check that bpp is valid](https://github.com/wxWidgets/wxWidgets/blob/b627852868459c8722b9b04533d51bdb987bf3a7/src/common/imagtga.cpp#L959), so any file with bpp=64 is rejected outright.

I'm not sure if it's worth adding a check to `DecodeRLE()` if it can't even be tested.

dxbjavid:
You're right. wxTGAHandler::LoadFile() calls CanRead() before ReadTGA(), and DoCanRead() already rejects any bpp outside 8/16/24/32, so DecodeRLE() never sees an oversized pixelSize from a real file. The guard is unreachable in practice and there's no way to write a test that actually hits it, so I'll close this one. Thanks for looking.

PR #26539, opened 2026-05-31
Fix Android base (no-gui) compilation

phantomhugo:
Hello, this fixes the android compilation with no GUI with a straight configure. I know android is not officially supported but I see no harm in allowing easy use of wxBase classes. For anyone who wants to test the commands I used were:

```
export NDK=~/Android/Sdk/ndk/29.0.14206865
export HOST_TAG=linux-x86_64
export TOOLCHAIN=$NDK/toolchains/llvm/prebuilt/$HOST_TAG
export AR=$TOOLCHAIN/bin/llvm-ar
export AS=$TOOLCHAIN/bin/llvm-as
export CC=$TOOLCHAIN/bin/x86_64-linux-android35-clang
export CXX=$TOOLCHAIN/bin/x86_64-linux-android35-clang++
export LD=$TOOLCHAIN/bin/ld.lld
export RANLIB=$TOOLCHAIN/bin/llvm-ranlib
export STRIP=$TOOLCHAIN/bin/llvm-strip
CXX_FLAGS=-fno-keep-inline-dllexport
./configure --disable-gui --with-zlib=builtin --with-regex=builtin --enable-intl --prefix=${NDK}/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/x86_64-linux-android/35/ --host x86_64-linux-android
make -j12
make install
```

I tested targeting android x64 and arm64. I did my best to avoid messing with other platforms but let me know any extra fix that is needed.


vadz:
Thanks, any work on wxAndroid is welcome but I am not sure that deleting the existing files is a step in right direction as they could be used for compilation under Windows and I just don't see why do they need to be removed.

I also have big reservations about `-include unistd.h`, I'd expect zlib to compile without hacks like this under Android and I hope there is some better way.

Also, as you can see, this broke a few CI builds and this needs to be fixed, of course.

Speaking of CI jobs, if we can make wxAndroid build at all, it would be very useful to add a CI job building it because this is the only way to prevent it from getting broken again in the future.

phantomhugo:
Hello @vadz thank you for taking the time to review this. I was trying to make the compilation as easy as possible, but you are right, the forced inclusion of unistd is a hack and the same results can be achieved with export CFLAGS="-include unistd.h" or using an external zlib, so I reverted that specific change.

Regarding the deleted files, those are all related with android and I think are not needed on modern NDKs as all those values can be achieved via running configure or cmake, and even in one of those files the content was ignored via preprocessor when building wxQt for android. If you still think is better to avoid the deletion I can restore the files and just remove the lines that use it.

I have fixed the issues on the CI/CD by regenerating the configure script with autoconf 2.69.

vadz:
> Regarding the deleted files, those are all related with android and I think are not needed on modern NDKs as all those values can be achieved via running configure or cmake

The `setup.h` files are supposed to be used for build systems without configuration step, e.g. when using MSVS or Xcode. Now I'm not really sure if wxAndroid can be built without configure/CMake but I think there was a way to do it using MSVS in the past. So for me it would be easier to leave these files alone and just avoid including them when using configure or CMake if necessary.

> I have fixed the issues on the CI/CD by regenerating the configure script with autoconf 2.69.

It's probably too late, but we have instructions in `build/tools/autoconf/README.md` about configure regeneration.

phantomhugo:
> The `setup.h` files are supposed to be used for build systems without configuration step, e.g. when using MSVS or Xcode. Now I'm not really sure if wxAndroid can be built without configure/CMake but I think there was a way to do it using MSVS in the past. So for me it would be easier to leave these files alone and just avoid including them when using configure or CMake if necessary.

Done.

> It's probably too late, but we have instructions in `build/tools/autoconf/README.md` about configure regeneration.

Not too late, I'm sorry I miss that. I have regenerated the configure script with the instructions given.

Let me know if you have any other concern or suggestion.

include/wx/platform.h
```diff
@@ -141,11 +141,6 @@
 
 #endif /* (__WXGTK__ || __WXQT__) && __WINDOWS__ */
 
-#ifdef __ANDROID__
```
vadz, line 144: Oh, sorry, I do agree that it doesn't make sense to keep the `config_android.h` file if it's not included from anywhere any more. I thought we'd modify the condition here in some way, e.g.

```cpp
#if defined(__ANDROID__) && !defined(wxHAVE_SETUP_H)
```

(`wxHAVE_SETUP_H` is not currently defined, but we could easily define it).

vadz (approved):
Sorry for change of mind, but after searching more carefully and looking at 1b4bff8262c70299a1c9f20548ec3043209616dc which was the only commit that affected this, I finally agree that `wx/android/setup.h` and `config_android.h` are not needed, so I'll finally remove them just as you did.

I'll merge this soon, thanks.

PR #26541, opened 2026-06-01
fix out-of-bounds read on trailing backslash in wxRegExImpl::Replace

dxbjavid:
wxRegExImpl::Replace() scans replacement.c_str() and does `*++p` after a backslash. When the replacement ends in a lone backslash, that reads the terminating NUL, the else branch appends it, and the loop's `p++` then steps one byte past the NUL so the `*p` condition reads out of bounds (ASan: heap-buffer-overflow read, 1 byte past the buffer; if the byte is non-zero it keeps scanning and copies adjacent memory into the result). Reachable through the public Replace()/ReplaceAll(). Keep a trailing backslash verbatim and stop before the increment. Test added in tests/regex/wxregextest.cpp.


vadz:
Thanks for finding and fixing this one, will merge soon.

PR #26542, opened 2026-06-01
wxOSX perform cleanup when applicationWillTerminate: is called

ryancog:
In the handling of `OSXOnWillTerminate()`, the TLW is `Close()`d, but not deleted (the pending cleanup never occurs), additionally, `OnExit()` is never called.

When clicking "Quit" from the Dock, Cocoa ends up in `-[NSApplication terminate:]`, which calls `exit(0)` more or less immediately after sending `applicationWillTerminate:`, and the documentation reads:
> Don’t bother to put final cleanup code in your app’s main() function—it will never be executed. If cleanup is necessary, perform that cleanup in the delegate’s [applicationWillTerminate:](doc://com.apple.documentation/documentation/appkit/nsapplicationdelegate/applicationwillterminate(_:)?language=occ) method.

So then, this PR changes wxOSX's `wxApp::OnEndSession()` to do cleanup similarly to the wxMSW port, making sure that `OnExit()` and all other cleanup is performed (part of that cleanup is deleting TLWs, so I've removed the `Close()`).

### DIFF
--- a/src/osx/cocoa/utils.mm
+++ b/src/osx/cocoa/utils.mm
@@ -238,6 +238,17 @@ void wxBell()
 - (void)applicationWillTerminate:(NSNotification *)application {
     wxUnusedVar(application);
     wxTheApp->OSXOnWillTerminate();
+
+    // Cocoa will `exit(0)` soon after returning from
+    // `applicationWillTerminate:`, without another opportunity for cleanup, so
+    // do it here.
+
+    // Ignore the return code. It's probably better to let Cocoa do the rest of
+    // its cleanup like normal rather than calling `exit()` here, and the
+    // return code doesn't mean much, if anything, for macOS usually.
+    (void)wxTheApp->CallOnExit();
+
+    wxEntryCleanup();
 }
 
 - (BOOL)applicationShouldTerminateAfterLastWindowClosed:(NSApplication *)sender


vadz (approved):
Thanks, assuming the comment is correct about there being no better place to do this, this looks the right thing to do to me.

Of course, as always, I'd prefer @csomor's confirmation before merging this if possible.

csomor:
Actually, the sequence of calling all the app callbacks was once preserved. I'll have to investigate.

csomor:
Ok, some flows are not the way they used to be, the system does not emulate a quit-app apple event anymore, but still putting these things in `wxApp::OnEndSession` seems too general to me, I'd rather add 

```
    wxTheApp->CallOnExit();
    wxEntryCleanup();
```
 at the end of `- (void)applicationWillTerminate:` directly, since for this we are given the contract, that `exit` will soon be called

ryancog:
@csomor Would you still want to keep the TLW `Close` in `OnEndSession()`? I’m not sure what problem is solved having that there…

csomor:
@ryancog yes, please leave it in there, although the LTW should be closed from within OnQueryEndSession, if there is a veto from the TWL it won't. So I'd still like to leave it in there in case there is an overrule and the system still calls `applicationWillTerminate`, I prefer to be defensive here.

ryancog:
Alright, moved to that to `applicationWillTerminate:`, then.

vadz:
@csomor Is this ok to merge now?

csomor:
> @csomor Is this ok to merge now?

@vadz yes, thanks a lot @ryancog

vadz:
Will push to master in a moment, but I think this might be also worth applying to 3.2, right?

csomor:
> Will push to master in a moment, but I think this might be also worth applying to 3.2, right?

yes, definitely, thanks

vadz:
Oops, backporting it to 3.2 didn't work because it doesn't have `CallOnExit()` — and the commit adding it can't be backported because it adds a virtual function, so I'm backing the backport out.

If somebody can create a version of this change appropriate for 3.2 before 3.2.11 release, it would be great, but I won't be able to do it, sorry.

csomor:
@vadz the additional methods used in `CallOnExit` do not exist yet in 3.2, so I'd suggest doing the same as `wxApp::OnEndSession` does for MSW in 3.2

just call `wxApp->OnExit()` instead of `wxApp-> CallOnExit()` there ? and leave the `wxEntryCleanup()` unchanged

vadz:
`CallOnExit()` was introduce to solve a problem with `OnExit()` being called with some TLWs still existing (among other things), and we'd run into this problem here, so I think we need to at least close all of them before calling it. But this really needs to be tested, which is why I can't/don't have time to do it.

csomor:
> `CallOnExit()` was introduce to solve a problem with `OnExit()` being called with some TLWs still existing (among other things), and we'd run into this problem here, so I think we need to at least close all of them before calling it. But this really needs to be tested, which is why I can't/don't have time to do it.

ok, thanks for the explanation, but right now this would not be called at all under 3.2, not under msw either, would it? I'd think the problem is present under 3.2 anyway 

So one way could be using the low-level methods to replace `wxApp-> CallOnExit` with

```
    wxApp->DeletePendingObjects();

    // and any remaining TLWs
    wxApp->DeleteAllTLWs();
    
    wxApp->OnExit()
    
    wxApp->CleanUp()
```

PR #26543, opened 2026-06-01
fix out-of-bounds read on trailing % in wxDateTime::Format

dxbjavid:
Fuzzing date format strings turned this up. Format() does `(*++p)` when it hits a '%', so a format ending in a lone '%' advances the iterator onto end(), and then the for-loop's own ++p steps one past it; since end()+1 != end() the loop re-enters and reads `*p` past the terminating NUL. It asserts in debug builds via the existing `case 0`, but over-reads the format buffer in release. Advance once and output the trailing '%' verbatim then stop, which is what the `case 0` comment already intended. ParseFormat() already handles this by returning false.

### DIFF
--- a/src/common/datetimefmt.cpp
+++ b/src/common/datetimefmt.cpp
@@ -405,8 +405,17 @@ wxString wxDateTime::Format(const wxString& formatp, const TimeZone& tz) const
             continue;
         }
 
+        // advance to the format specifier following the '%': a trailing '%'
+        // is not a valid format, so output it verbatim and stop here instead
+        // of letting the loop's own ++p move the iterator past the end
+        if ( ++p == format.end() )
+        {
+            res += wxT('%');
+            break;
+        }
+
         // set the default format
-        switch ( (*++p).GetValue() )
+        switch ( (*p).GetValue() )
         {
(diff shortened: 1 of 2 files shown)


vadz:
Thanks! The fix looks correct but MSVC standard library doesn't like invalid format strings, apparently, making the CI jobs using it fail. The simplest is to just skip/disable this test when using it, could you please do it?

dxbjavid:
Done, wrapped the new check in #ifndef _MSC_VER so it's skipped with the MSVC CRT. The fix itself is unaffected.

vadz:
Sorry, I should have thought about this, but MSVC CRT is also used with MinGW, so the test needs to be excluded for it too.

dxbjavid:
Good point, MinGW links the MS CRT too. Changed the guard to `!defined(_MSC_VER) && !defined(__MINGW32__)` so it's skipped there as well.

vadz:
Thanks for fixing the CI jobs! Thinking more about this, maybe it would be more useful to assert in our own implementation of this function too, considering how using such format strings can't be the right thing to do, knowing they won't work when using MSVC CRT.

But I'll leave this for another PR later and will just (squash) merge this one for now, thanks again.

PR #26547, opened 2026-06-01
MSW: Dark mode theme for wxDataViewCtrl item buttons

stevecor:
Add dark mode theme class for collapse/expand tree item buttons. This fix is for the wxDataViewCtrl when switching between dark and light themes. The problem solved is that the button colors fail to update and so they disappear.


vadz:
Thanks, will merge soon.

PR #26548, opened 2026-06-01
fix out-of-bounds read on truncated UTF-8 in wxUString::assignFromUTF8

dxbjavid:
The length-counting pass in the NUL-terminated assignFromUTF8 advances p by the lead byte's UTF-8 sequence length without checking those bytes are really there. A string ending in a truncated multibyte sequence (e.g. a lone 0xC3 before the NUL) skips p past the terminating NUL, so the while(*p) test then reads past the end of the buffer. The assignFromUTF8(str, n) overload below already guards this via utf8_pos + len > n; this just adds the equivalent check for the terminated form. Reachable through wxUString::FromUTF8(const char*).


vadz:
Thanks, will merge soon.

We really should have some unit tests for this class...

PR #26552, opened 2026-06-02
wxOSX: Add wxDF_UNICODETEXT flag for text sample

ssrlive:
Based on my testing, when copying text from Terminal and TextEdit, the logic here triggers the `wxDF_TEXT` condition. My OS is Sonoma 14.8.7.

### DIFF
--- a/samples/text/text.cpp
+++ b/samples/text/text.cpp
@@ -190,7 +190,7 @@ public:
     {
         wxClipboardLocker lockClip;
 
-        event.Enable( wxTheClipboard->IsSupported(wxDF_TEXT) );
+        event.Enable( wxTheClipboard->IsSupported(wxDF_TEXT) || wxTheClipboard->IsSupported(wxDF_UNICODETEXT) );
     }
 
     void OnUpdateCopyToClipboard(wxUpdateUIEvent& event)


vadz:
Thanks!

@csomor are you fine with merging this, i.e. is it correct to get `wxDF_TEXT` here or should this be fixed instead?

csomor:
@vadz `wxDF_TEXT` is already handled by the default handler in front of the switch statement, the problem with the proposed extension is that it's too broad when compared to the `wxDF_TEXT` interpretation within wx, ie you might get text with the wrong encoding information. We should use `wxDF_UNICODETEXT` as the preferred option everywhere nowadays IMHO.

`wxDF_TEXT` used to mean "platform" encoding on mac, but the terminal eg offers the following types

```
[0]	__NSCFConstantString *	"public.rtf"	0x000000020764b070
[1]	__NSCFConstantString *	"NeXT Rich Text Format v1.0 pasteboard type"	0x0000000207640ed0
[2]	__NSCFConstantString *	"public.utf8-plain-text"	0x00000002075b8a30
[3]	__NSCFConstantString *	"NSStringPboardType"	0x000000020763d5d0
[4]	__NSCFConstantString *	"public.utf16-external-plain-text"	0x000000020767a270
```

so with the patch `HasData` responds true, because there is `public.rtf `available, but the wxTextDataObject works properly because it automatically uses `wxDF_UNICODETEXT` as well resolving to the available utf data. If it used `wxDF_TEXT` only, then it would not get platform encoded text but `rtf`.

ssrlive:
Which example can I use for testing? I’d like to verify this on macOS. @csomor

csomor:
@ssrlive the clipboard example does not exist as sample by itself anymore, but the dnd sample has some code. 

(As I said, to make things work correctly, you should change the `wxDF_TEXT` occurrences to `wxDF_UNICODE` there, when dealing with text data is available on the clipboard)

If you want to see what gets returned 'natively' as `wxDF_TEXT` you cannot use the `wxTextDataObject` itself, as it always checks unicode text as well, you'd have to retrieve the data manually, eg you could add this in the `DnDFrame::OnPaste` right in front of the `wxTextDataObject text;` line 

```
    wxCustomDataObject data(wxDF_TEXT);
    if ( wxTheClipboard->GetData(data) ) {
        wxLogError("Text retrieved from clipboard");
    } else {
        wxLogError("Can't get text data from the clipboard");
    }
```

This is what you get:

ssrlive:
Based on my observations regarding the [text](https://github.com/wxWidgets/wxWidgets/blob/master/samples/text/text.cpp) example, the way the clipboard content is evaluated differs from how the native context menus of operating system controls evaluate it.

Here is a comparison image showing what I observed after copying text selected in the terminal.

When I apply my patch, the "Clipboard/paste" menu item becomes enabled.


Screenshot 2026-06-06 at 10 39 13
Screenshot 2026-06-06 at 10 37 08

ssrlive:
But when i change this line 
https://github.com/wxWidgets/wxWidgets/blob/14687e2b0a6af13ed03135c8a0a7d9d8d5a24895/samples/text/text.cpp#L193

to `event.Enable( wxTheClipboard->IsSupported(wxDF_UNICODETEXT) );` , everthing become fine.

So, is it enough to change just this one line?

Of course, this conclusion holds true for macOS, but I don't know if it applies to other OSs.

csomor:
> But when i change this line
> 
> https://github.com/wxWidgets/wxWidgets/blob/14687e2b0a6af13ed03135c8a0a7d9d8d5a24895/samples/text/text.cpp#L193
> 
> to `event.Enable( wxTheClipboard->IsSupported(wxDF_UNICODETEXT) );` , everthing become fine.
> 
> So, is it enough to change just this one line?

Exactly, that's what I meant. 

> Of course, this conclusion holds true for macOS, but I don't know if it applies to other OSs.

If you want to be safe for all systems, you could OR two `wxTheClipboard->IsSupported` asking for `wxDF_TEXT` and `wxDF_UNICODETEXT`,

```
...( wxTheClipboard->IsSupported(wxDF_TEXT) || wxTheClipboard->IsSupported(wxDF_UNICODETEXT) )

```

ssrlive:
Hi @vadz , I have finished this PR.

vadz:
Sorry, I don't agree with this. wxOSX should provide `wxDF_TEXT`, applications shouldn't have to use `wxDF_UNICODETEXT` which only exists for compatibility reasons (we needed it when we still had a non-Unicode build variant). There is no meaningful distinction between them any more, the application gets a `wxString` with the data on the clipboard in both cases, so why should it bother with supporting it.

@csomor Why can't wxOSX provide data in `wxDF_TEXT` format?

csomor:
@vadz On the binary level, `wxDF_TEXT` and `wxDF_UNICODETEXT` are not the same. `wxDF_TEXT` has the native encoding `com.apple.traditional-mac-plain-text` corresponding to `MacRoman`. So if there is no data with MacRoman encoding available we cannot read it into a data object expecting `wxDF_TEXT`.

The encoding of `wxDF_UNICODETEXT` meanwhile, can ingest anything

```
#ifdef wxNEEDS_UTF8_FOR_TEXT_DATAOBJ
            f = kUTTypeUTF8PlainText;
#elif defined(wxNEEDS_UTF16_FOR_TEXT_DATAOBJ)
            f = kUTTypeUTF16PlainText;
#else
#error 
```

vadz:
This looks like an implementation detail that we should hide from the application. I.e. whether the actual data is in Mac Roman or UTF-8 or UTF-16, I think we should still pretend that both `wxDF_TEXT` *and* `wxDF_UNICODETEXT` are available (the latter for compatibility only) and return `wxString` with the Unicode text obtained from raw data by converting it if necessary.

csomor:
Right now, wxTextDataObject does its job, dealing with both perfectly. So that's the thing that should be used anyway. 

But IMHO we must stay true to the low level, if we say wxDF_TEXT is available, then we must be able to provide the data in this format, byte-wise. So to me the only way to deliver this in all situations would be to change the interpretation of wxDF_TEXT from MacRoman encoding to UTF8.

vadz:
> Right now, wxTextDataObject does its job, dealing with both perfectly. So that's the thing that should be used anyway.

Yes, certainly, but `wxClipboard::IsSupported()` requires a `wxDataFormat`.
 
So the simplest fix I see is to just return `true` from `IsSupported(wxDF_TEXT)` when the actual format on clipboard is `wxDF_UNICODETEXT`. AFAICS this shouldn't break anything because the only thing the application may do after calling this function is to use `wxTextDataObject` to retrieve the data and this will work.

We would need to also include `wxDF_TEXT` in `wxClipboardEvent` but I'm not even sure if this is supported/generated by wxOSX.

csomor:
> > Right now, wxTextDataObject does its job, dealing with both perfectly. So that's the thing that should be used anyway.
> 
> Yes, certainly, but `wxClipboard::IsSupported()` requires a `wxDataFormat`.

Yes, that's why I also thought about adding an overload accepting a data object.

> So the simplest fix I see is to just return `true` from `IsSupported(wxDF_TEXT)` when the actual format on clipboard is `wxDF_UNICODETEXT`. AFAICS this shouldn't break anything because the only thing the application may do after calling this function is to use `wxTextDataObject` to retrieve the data and this will work.

If we only do this in this method, not generally within the data objects, I agree.
Actually, as `GetData` may use any data object, as in my sample from above, this still could be a problem, but I guess everybody uses wxTextDataObject for this.

> We would need to also include `wxDF_TEXT` in `wxClipboardEvent` but I'm not even sure if this is supported/generated by wxOSX.

Not I'm not generating it, there's no native notification.

csomor:
so this would be along your intention

```
bool wxClipboard::IsSupported(const wxDataFormat &dataFormat)
{
    wxLogTrace(TRACE_CLIPBOARD, wxT("Checking if format %s is available"),
               dataFormat.GetId().c_str());

    if ( m_data )
        return m_data->IsSupported( dataFormat ) ||
            (dataFormat.GetType() == wxDF_TEXT && m_data->IsSupported( wxDF_UNICODETEXT ));

    return wxOSXPasteboard::GetGeneralClipboard()->IsSupported(dataFormat) ||
        (dataFormat.GetType() == wxDF_TEXT && wxOSXPasteboard::GetGeneralClipboard()->IsSupported(wxDF_UNICODETEXT));
}

```

vadz:
Conceptually, yes, but the first change shouldn't be necessary, `wxTextDataObject::IsSupported()` should already return true for both formats.

And I think it might make more sense to apply something like this instead of the second one:

```diff
diff --git a/src/osx/carbon/dataobj.cpp b/src/osx/carbon/dataobj.cpp
index 423da3784a..4c9348daeb 100644
--- a/src/osx/carbon/dataobj.cpp
+++ b/src/osx/carbon/dataobj.cpp
@@ -156,6 +156,13 @@ void wxDataFormat::DoAddSupportedTypes(CFMutableArrayRef cfarray, bool forSettin
     else
     {
         CFArrayAppendValue(cfarray, GetFormatForType(m_type));
+
+        // Text and Unicode text are the same thing at wx level.
+        if ( m_type == wxDF_TEXT )
+            CFArrayAppendValue(cfarray, kUTTypeUTF8PlainText);
+        else if ( m_type == wxDF_UNICODETEXT )
+            CFArrayAppendValue(cfarray, kUTTypeTraditionalMacText);
+
         if ( forSetting )
         {
             // add additional accepted types which we are ready to accept and can
```

csomor:
But then we have the problem that we accept something as wxDF_TEXT in UTF8, which cannot be rendered as wxDF_TEXT byte-wise. A smiley in utf8 does not have a counterpart in wxDF_TEXT, which, as I said, currently means MacRoman encoding. I don't want any ambiguity about what wxDF_TEXT is at the lower level. 

So the only way out, when we want to solve it at the individual format level, I see, would be to have wxDF_TEXT mean utf8 instead of MacRoman.

And we don't need the second alias, MacRoman is translated by the system into Unicode encoding on demand.

vadz:
> But then we have the problem that we accept something as wxDF_TEXT in UTF8, which cannot be rendered as wxDF_TEXT byte-wise.

Yes, indeed. This being said, I still don't see any realistic use cases when this would be a problem for the application.

> So the only way out, when we want to solve it at the individual format level, I see, would be to have wxDF_TEXT mean utf8 instead of MacRoman.

I don't see any problems with this either.

vadz:
This is replaced by #26565.

PR #26554, opened 2026-06-02
fix rowspan/colspan integer overflow in wxHtmlTableCell::AddCell

dxbjavid:
AddCell() reads COLSPAN and ROWSPAN from the markup into ints with no upper bound, then uses r + rowspan and c + colspan to grow the cell table and to index ypos[] in Layout(). A value near INT_MAX, e.g. <td rowspan="2147483647"> in the second row, overflows the addition so the `r + rowspan > m_NumRows` growth check is bypassed and Layout() writes past the end of ypos[]. Clamp colspan and rowspan to the limits from the HTML spec (1000 and 65534).


vadz:
Thanks for fixing this!

dxbjavid:
Glad it helped - happy to contribute.

PR #26556, opened 2026-06-02
Update GitHub Actions to be Node v24 compatible

swt2c:
Node v20 has gone EOL and will be removed from GitHub Actions soon. This migrates all the GitHub Actions to use versions that are Node 24 compatible.

The only change of real substance is moving to the step-security fork of gha-setup-vsdevenv which seems probably better than wx maintaining its own fork.


vadz:
Thanks for taking care of this!

We need to do this in 3.2 too, will try to cherry pick this there.

Issue #26515, opened 2026-05-26
cmake doesn't install WebView, MediaCtrl headers unconditionally


swt2c:
When building with cmake, the WebView and MediaCtrl headers (and possibly others) are not installed, for example, on Linux if the webkitgtk libraries are not found.  The webview.h header compiles fine without WebView support, and the autoconf build installs it unconditionally, so it would be nice if cmake matched autoconf and always installed this header.

I haven't checked whether the same thing happens with wxGLCanvas, but would be nice to fix those as well, if the same thing applies.

barracuda156:
Are we certain that in this case CMake does the wrong thing and not autoconf? It is not uncommon for build systems to check for specific headers in order to determine supported optional features. In this scenario having webview headers (or gstreamer ones, or GL) may lead to incorrect behavior for a dependent.

P. S. I understand that regardless of correctness per se there is also that some software may already rely on a wrong behavior, and then fixing the latter will break the former.

MaartenBent:
I'll try to fix this. 

wxWidgets has `wxUSE_XXX` checks to determine supported features, there should be no need to check for header existence and you should be able to always include them.

vadz:
Thanks Maarten!

FWIW I think that while, in theory, both approaches (always install everything or install the header only if the corresponding feature is available) could be considered valid, in practice we've always installed all headers and we should continue doing it like this unless we have a very good reason to change it.

PR #26562, opened 2026-06-06
CMake: Always install common headers

MaartenBent:
See #26515

And fix building when some options are disabled.


vadz (approved):
Thanks for all the improvements! Will merge soon.

PR #26563, opened 2026-06-06
Nullify WebView (Edge) implemenation pointer, add pointer check

Blake-Madden:
Null out `m_impl` in case there are any stray pending events.

Also, check that pointer in size event handler.

Also, minor comment fixes.


vadz:
Sorry, this doesn't look like the correct fix: just calling `OnSize()` on an already destroyed instance of `wxWebViewEdge` is already undefined behaviour, so we must prevent this from happening. Probably the correct fix is to call `Unbind(wxEVT_SIZE)` in dtor.

Also, if you really need to change comments (I'd rather leave them be), please do it in a separate commit and not in the same one making some non-trivial changes. TIA!

PR #26565, opened 2026-06-06
OSX: make wxDF_TEXT be in UTF8 encoding

csomor:
see #26552

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -140,6 +140,9 @@ Changes in behaviour not resulting in compilation errors
   already didn't do it in wxMSW. Call Show() explicitly before Raise() if
   necessary.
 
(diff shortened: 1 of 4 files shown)


vadz:
Stefan, should we document this as an incompatible change and, if so, how would you explain it, or does it have no real consequences on the application code?

csomor:
@vadz yes, I'll have to add the docs, it has consequences if the code explicitly retrieves `wxDF_TEXT` and expects it to be in `MacRoman` encoding. No real consequences if `wxTextDataObject` is used, though.

vadz:
Thanks, should this be merged now?

FWIW I don't think it's worth to provide the rationale for this change in the change log, it would be more useful to explain how the code should be updated (or at least which code even needs updating).

vadz:
Thanks for the update(s)! I think it's good to merge it now, so I'll do it.

PR #26566, opened 2026-06-06
Fix artifacting issue with Edge WebView

Blake-Madden:
When an Edge WebView is hidden and then reshown, artifacts from the old window will appear in the `wxWebView`. This fixes it by calling `put_IsVisible` when the control (or its parent) is shown or hidden and also forces a repaint by updating the painting rect (this is the necessary part and the closest I could come to emulating an async refresh).

Here is an example of what I see with a `wxWebView` in a composite control inside of a split window when I select a different window and come back:

image

What it should look like:

image

Here is a reference to parent visibility issues with WebView2:

https://github.com/MicrosoftEdge/WebView2Feedback/issues/1094

and maybe related:

https://groups.google.com/g/wx-dev/c/NoVbVMODDB4

P.S.

I have to use `CallAfter` here because of the async nature of WebView.

### DIFF
--- a/src/msw/webview_edge.cpp
+++ b/src/msw/webview_edge.cpp
@@ -1133,7 +1133,6 @@ wxWebViewEdge::~wxWebViewEdge()
     if (topLevelParent)
         topLevelParent->Unbind(wxEVT_ICONIZE, &wxWebViewEdge::OnTopLevelParentIconized, this);
     Unbind(wxEVT_SHOW, &wxWebViewEdge::OnShow, this);
-    GetParent()->Unbind(wxEVT_SHOW, &wxWebViewEdge::OnShow, this);
     delete m_impl;
 }
 
@@ -1162,7 +1161,6 @@ bool wxWebViewEdge::Create(wxWindow* parent,
     if (topLevelParent)
         topLevelParent->Bind(wxEVT_ICONIZE, &wxWebViewEdge::OnTopLevelParentIconized, this);
     Bind(wxEVT_SHOW, &wxWebViewEdge::OnShow, this);
-    GetParent()->Bind(wxEVT_SHOW, &wxWebViewEdge::OnShow, this);
 
     LoadURL(url);
     return true;
--- a/src/msw/webview_edge.cpp
+++ b/src/msw/webview_edge.cpp
@@ -1161,8 +1161,7 @@ bool wxWebViewEdge::Create(wxWindow* parent,
     wxWindow* topLevelParent = wxGetTopLevelParent(this);
     if (topLevelParent)
         topLevelParent->Bind(wxEVT_ICONIZE, &wxWebViewEdge::OnTopLevelParentIconized, this);
-    if (GetParent())
-        GetParent()->Bind(wxEVT_SHOW, &wxWebViewEdge::OnParentShow, this);
+    GetParent()->Bind(wxEVT_SHOW, &wxWebViewEdge::OnParentShow, this);
 
     LoadURL(url);
     return true;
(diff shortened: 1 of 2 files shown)


src/msw/webview_edge.cpp
```diff
@@ -1157,6 +1161,8 @@ bool wxWebViewEdge::Create(wxWindow* parent,
     wxWindow* topLevelParent = wxGetTopLevelParent(this);
     if (topLevelParent)
         topLevelParent->Bind(wxEVT_ICONIZE, &wxWebViewEdge::OnTopLevelParentIconized, this);
+    if (GetParent())
```
vadz: I think this is unnecessary as it should always have a parent, it shouldn't be possible to create it as a TLW.

src/msw/webview_edge.cpp
```diff
@@ -1182,6 +1188,27 @@ void wxWebViewEdge::OnTopLevelParentIconized(wxIconizeEvent& event)
...
 
+void wxWebViewEdge::OnParentShow(wxShowEvent& event)
+{
+    if ( m_impl && m_impl->m_webViewController )
+    {
+        m_impl->m_webViewController->put_IsVisible(event.IsShown());
+        // Force a refresh by refreshing its paint area
+        if ( event.IsShown() )
+        {
+            // put_Bounds is a no-op when the rect is unchanged, so collapse to
```
vadz, line 1197: There is really no other way to force it to repaint?

vadz (commented):
This looks a bit ugly, but if this is the only way to fix the problem, let's do it...

src/msw/webview_edge.cpp
Blake-Madden: Got, will remove. I tend to be overly pedantic.

src/msw/webview_edge.cpp
Blake-Madden: To my understanding, there is no "invalidate and refresh" mechanism that we can use because the rendering is taking place in a DirectComposition visual, not the HWND, so traditional repaint request methods aren't available:

https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/windowed-vs-visual-hosting

https://github.com/unoplatform/uno/issues/22694

And the only other solutions are full reloads or complicated JScript:

https://weblog.west-wind.com/posts/2026/Feb/04/Reliably-Refreshing-the-WebView2-Control

Setting the drawing rect to {0,0} and then telling the control to get its drawing rect from the client forces it to see that its rendering area changed and should repaint. This seems to be the closest "invalidate and refresh" method that I can find and works well on Windows 11.

Blake-Madden:
I was experiencing this artifacting issue when just hiding/showing `wxWebView` from a splitter window (no composite control) also, so updated this to handle the controls show event also.

PBfordev:
So something have changed since the referenced wxTrac issue from 2020, where the problem was that the deep-nest child windows may not get the show event? 
https://github.com/wxWidgets/wxWidgets/issues/18851#issuecomment-1011247975

vadz:
Sorry, I'm really confused now: why do we need to handle `wxEVT_SHOW` from both this window and the parent? Or, to put it differently: if we need to handle it for this window, why do we need to handle it for the parent too, when we should get this event anyhow when the parent is hidden/shown?

Blake-Madden:
> Sorry, I'm really confused now: why do we need to handle `wxEVT_SHOW` from both this window and the parent? Or, to put it differently: if we need to handle it for this window, why do we need to handle it for the parent too, when we should get this event anyhow when the parent is hidden/shown?

Fair enough, I removed the parent binding.

PR #26567, opened 2026-06-07
Apply  wx coding standards to ribbon source

Blake-Madden:
Name variables correctly (camel case, members prefixed with `m_`).
Fix spacing and other formatting.


vadz (commented):
I'm always uncomfortable applying changes like this because there is always a risk of breaking something without any really good reason to do it.

For example, I've randomly checked if anybody [uses `m_primary_scheme_colour`](https://github.com/search?q=m_primary_scheme_colour+language%3AC%2B%2B+-path%3Awx%2Fribbon%2Fart.h+-path%3Aart_msw.cpp&type=code) and it seems like CodeLite does. So I think we need to leave the protected variables alone...

PR #26568, opened 2026-06-07
Remove plugin interface from test_drawing

Blake-Madden:
This was never used and doesn't seem like it ever will be.

Refer to #26504


vadz:
If we remove this, shouldn't we remove `test_drawing` entirely? It seems like it's not really useful without this plugin (but I could be wrong...).

Let me cc @aasselin-en just in case (I think it's the account of the original author of this code, sorry in advance if I tagged a wrong person).

Blake-Madden:
It fails on MSW for me because it is doing a pixel comparison, not sure if this ever worked there. Yeah, I could just remove this whole thing and move SVG GC testing into the test_gui instead.

aasselin:
I'll check soon and tell you. I remember that i developed years ago some additional images formats.

aasselin:
HI @vadz , I see nothing I would have touched in this area, so no problem for me.

PR #26570, opened 2026-06-08
Add dark mode support to wxMSW task dialog-based dialogs

vadz:
This is #26311 with cleanup and some fixes.

Unfortunately there are still ressource leaks remaining under Windows 10: every time a message box is shown, 3 GDI handles (and 1 USER one) are leaked. I can't figure out what they are: Application Verifier doesn't show anything at all and [GDIView](https://www.nirsoft.net/utils/gdi_handles.html) shows the total number of handles growing while each column remains stable. It also shows the total number increasing by 6, unlike the task manager which only shows 3 leaked handles.

P.S. There are no leaks under Windows 11, but this is not really surprising, a lot of drawing code is skipped entirely there as native dark mode support is available.

### DIFF
--- a/include/wx/msw/private/taskdlg.h
+++ b/include/wx/msw/private/taskdlg.h
@@ -1,0 +1,56 @@
+///////////////////////////////////////////////////////////////////////////////
+// Name:        wx/msw/private/taskdlg.h
+// Purpose:     Dark mode support for native TaskDialog.
+// Author:      Mohmed Abdel-Fattah
+// Created:     2026-06-07
+// Copyright:   (c) 2026 wxWidgets development team
+// Licence:     wxWindows licence
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef _WX_MSW_PRIVATE_TASKDLG_H_
+#define _WX_MSW_PRIVATE_TASKDLG_H_
+
+#include "wx/msw/wrapcctl.h"
+
+namespace wxMSWDarkMode
+{
+
+// ----------------------------------------------------------------------------
+// TaskDialog dark mode helpers
+// ----------------------------------------------------------------------------
+//
+// These two functions bracket the lifetime of dark-mode theming for any
+// TaskDialog-based dialog (wxMessageDialog, wxRichMessageDialog,
+// wxProgressDialog, wxAboutBox).
+//
+// Typical usage inside a PFTASKDIALOGCALLBACK:
+//
+//   case TDN_CREATED:
+//       wxMSWDarkMode::ApplyToTaskDialog(hwnd, pCfg);
+//       break;
+//   case TDN_DESTROYED:
+//       wxMSWDarkMode::RemoveFromTaskDialog(hwnd);
+//       break;
+//
+// Both functions are no-ops when IsActive() returns false, so no guard is
+// needed at the call site.
+
+// Apply dark mode theming to an existing TaskDialog window.
+//
+// Must be called from TDN_CREATED (not TDN_DIALOG_CONSTRUCTED) so that the
+// complete child-window hierarchy exists and UI Automation can walk it.
+//
+// pCfg may be nullptr; when provided it enables correct icon overdraw on
+// Windows 10 where there is no native dark TaskDialog theme.
+void AllowForTaskDialog(HWND hwnd, const TASKDIALOGCONFIG* pCfg = nullptr);
+
+// Remove all subclasses and free all GDI/theme resources installed by
+// ApplyToTaskDialog().
+//
+// Must be called from TDN_DESTROYED or WM_DESTROY.
+// Safe to call even if ApplyToTaskDialog() was never called for this HWND.
+void RemoveFromTaskDialog(HWND hwnd);
+
+} // namespace wxMSWDarkMode
+
+#endif // _WX_MSW_PRIVATE_TASKDLG_H_
--- a/include/wx/msw/private/taskdlg.h
+++ b/include/wx/msw/private/taskdlg.h
@@ -1,7 +1,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 // Name:        wx/msw/private/taskdlg.h
 // Purpose:     Dark mode support for native TaskDialog.
-// Author:      Mohmed Abdel-Fattah
+// Author:      Mohmed Abdel-Fattah (memoarfaa)
 // Created:     2026-06-07
 // Copyright:   (c) 2026 wxWidgets development team
 // Licence:     wxWindows licence
--- a/src/msw/taskdlg.cpp
+++ b/src/msw/taskdlg.cpp
@@ -1257,12 +1257,18 @@ void TDAttach(HWND hwndTD, const TASKDIALOGCONFIG* pCfg)
     if ( native )
         wxMSWDarkMode::AllowForWindow(hwndTD, L"DarkMode_Explorer", nullptr);
 
-    SetWindowSubclassIfNeeded(hwndTD, TDCtrlContainerSubclassProc, kTDMainSubclassId);
+    HBRUSH nb = GetSolidBrush(TDDarkCol::kPrimary);
+
+    SetWindowSubclassIfNeeded
+    (
+        hwndTD,
+        TDCtrlContainerSubclassProc,
+        kTDMainSubclassId,
+        [nb]() { return reinterpret_cast<DWORD_PTR>(nb); }
+    );
 
     // Dark title bar
     wxMSWDarkMode::ConfigureTLW(hwndTD);
-    HBRUSH nb = GetSolidBrush(TDDarkCol::kPrimary);
-    ::SetClassLongPtr(hwndTD, GCLP_HBRBACKGROUND, reinterpret_cast<LONG_PTR>(nb));
     ::EnumChildWindows(hwndTD, TDEnumSysColorProc, 0);
     ::SendMessage(hwndTD, WM_THEMECHANGED, 0, 0);
 }
--- a/src/msw/taskdlg.cpp
+++ b/src/msw/taskdlg.cpp
@@ -118,11 +118,6 @@ struct TDPageState
 
     const TASKDIALOGCONFIG* pCfg = nullptr;
 
-    void CloseThemes()
-    {
-        themesOk = false;
-    }
-
     using TDStateMap = std::unordered_map<HWND, TDPageState>;
 
     static TDStateMap& GetAll()
@@ -233,7 +228,6 @@ static bool TDHasNativeDarkTheme()
 
 static void TDRefreshThemes(HWND hwnd, TDPageState& s)
 {
-    s.CloseThemes();
     s.isDark = TDHasNativeDarkTheme();
     const int dpi = wxGetWindowDPI(hwnd).x;
 
@@ -715,7 +709,8 @@ static LRESULT CALLBACK TDPageSubclassProc(
     case WM_THEMECHANGED:
     {
         TDPageState& s = TDPageState::Get(hwnd);
-        s.CloseThemes(); s.elemsOk = false;
+        s.themesOk = false;
+        s.elemsOk = false;
         InvalidateRect(hwnd, nullptr, FALSE);
         break;
     }
--- a/src/msw/taskdlg.cpp
+++ b/src/msw/taskdlg.cpp
@@ -24,6 +24,7 @@
 #endif
 
 #ifndef WX_PRECOMP
+    #include "wx/brush.h"
 #endif // WX_PRECOMP
 
 #include "wx/msw/private/taskdlg.h"
--- a/src/msw/taskdlg.cpp
+++ b/src/msw/taskdlg.cpp
@@ -605,7 +605,8 @@ static void TDPaintText(HDC hdc, const TDPageState& s)
 
         if (!native)
         {
-            DTTOPTS opts = { sizeof(opts) }; opts.dwFlags = DTT_COMPOSITED | DTT_TEXTCOLOR;
+            WinStructWordSize<DTTOPTS> opts;
+            opts.dwFlags = DTT_COMPOSITED | DTT_TEXTCOLOR;
             opts.crText = TDGetTextColour(s, part);
             FillRect(hdc, &rcT, brBg);
             ::DrawThemeTextEx(s.hTD, hdc, part, 0, el.name.c_str(), -1, dtF, &rcT, &opts);
@@ -780,7 +781,8 @@ static LRESULT CALLBACK TDRadioButtonSubclassProc(
         GetWindowTextW(hwnd, text, static_cast<int>(std::size(text)));
         auto gs = hBtn.GetTrueSize(BP_RADIOBUTTON, RBS_UNCHECKEDNORMAL);
         RECT rcT = { gs.x + 2,0,rcC.right,rcC.bottom };
-        DTTOPTS opts = { sizeof(opts) };
+
+        WinStructWordSize<DTTOPTS> opts;
         opts.dwFlags = DTT_COMPOSITED | DTT_TEXTCOLOR;
         opts.crText = TDDarkCol::kTextNormal;
         LOGFONT lf = {};
--- a/src/msw/taskdlg.cpp
+++ b/src/msw/taskdlg.cpp
@@ -68,6 +68,19 @@ namespace TDDarkCol
 namespace
 {
 
+// Helper function create an HBRUSH from a brush stored in wxTheBrushList.
+// This allows not to recreate the brushes for the same colour and also ensures
+// that the brushes are eventually deleted.
+HBRUSH GetSolidBrush(COLORREF rgb)
+{
+    const wxColour col = wxRGBToColour(rgb);
+    wxBrush* const brush = wxTheBrushList->FindOrCreateBrush(col);
+    if ( !brush )
+        return 0;
+
+    return GetHbrushOf(*brush);
+}
+
 // Helper function to get an HWND from IUIAutomationElement.
 HWND GetHWNDFromElement(IUIAutomationElement* element)
 {
@@ -861,13 +874,12 @@ TDCtrlContainerSubclassProc(HWND hwnd,
                 ::SetTextColor(hdc, TDDarkCol::kTextNormal);
 
                 if ( !hbr )
-                    hbr = ::CreateSolidBrush(TDDarkCol::kSecondary);
+                    hbr = GetSolidBrush(TDDarkCol::kSecondary);
 
                 return reinterpret_cast<LRESULT>(hbr);
             }
 
         case WM_DESTROY:
-            ::DeleteObject(hbr);
             ::RemoveWindowSubclass(hwnd, TDCtrlContainerSubclassProc, uId);
             break;
     }
@@ -1003,14 +1015,12 @@ void TDSubclassContainer(HWND hwndParent, COLORREF bg)
     if ( !hwndParent )
         return;
 
-    // The brush created here is passed to the subclass procedure as reference
-    // data and is deleted in the WM_DESTROY handler there.
     SetWindowSubclassIfNeeded
     (
         hwndParent,
         TDCtrlContainerSubclassProc,
         kTDCtrlSubclassId,
-        [bg]() { return ::CreateSolidBrush(bg); }
+        [bg]() { return GetSolidBrush(bg); }
     );
 }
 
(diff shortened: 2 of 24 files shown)


vadz:
I was looking for the leaks in a wrong place, we've been actually leaking `HICON`s, of all things... AFAICS this is fixed now, neither GDIView nor Process Explorer show any handles growth when repeatedly opening and closing a message box.

There is still the fact that a lot of resources are allocated the first time this happens, but I think we can live with it, so this should be ready for merge if nobody sees any big problems with it.

include/wx/msw/private/taskdlg.h
```diff
@@ -0,0 +1,56 @@
+///////////////////////////////////////////////////////////////////////////////
+// Name:        wx/msw/private/taskdlg.h
+// Purpose:     Dark mode support for native TaskDialog.
+// Author:      Mohmed Abdel-Fattah
```
memoarfaa: @vadz, thank you so much for completing this PR. I was sick for a while and couldn't work on it, but I'm fine now – though I have a lot of backlog because of that. I really appreciate you taking over, fixing the leaks, and doing all the cleanup.

Thank you also for adding my name as the original author in include/wx/msw/private/taskdlg.h – I noticed that and I'm grateful.

If it's not too much trouble, could you also add my GitHub nickname there? Something like:

cpp
// Author:      memoarfaa (Mohmed Abdel-Fattah)
That way people can more easily associate the contribution with my GitHub profile. Thanks again for all your hard work – the dark mode support is a great addition to wxWidgets!

memoarfaa:
@vadz – thank you so much for completing this PR. I was sick for a while and couldn't work on it, but I'm fine now – though I have a lot of backlog because of that. I really appreciate you taking over and fixing the leaks and cleanup.

> I'm closing this one in favour of #26570 which is closer to be ready to be merged, but it still has leaks, as explained there. Any help with finding and fixing them would be welcome, but so would be testing more generally because I could have broken something with my changes, even though they were not supposed to have any material effect.
> 
> Thanks again for contributing all this code!
> — https://github.com/wxWidgets/wxWidgets/pull/26311#issuecomment-4644638845

Regarding your comment about possibly having broken something – unfortunately, I think there may be a regression. After testing the PR, I noticed two related issues that affect the dark theme after a task dialog closes. Please consider these as observations (not fully confirmed root causes).

---

### ❓ Issue 1: The application's menu and background brush appear to be destroyed after the dialog closes

After opening and closing a task dialog (e.g., a message box), the dark theme on the main application's menu bar disappears (white/corrupted background).

**Possible cause (not fully tested):**  
The brush might be stored in `dwRef` of `TDCtrlContainerSubclassProc`, and when the task dialog closes, that brush handle could become invalid. If the brush came from the global `wxTheBrushList`, invalidating it could break the dark theme for the whole app.

**Suggested change (optional, not tested):**  
Try *not* storing the brush in `dwRef`. Instead, call `GetSolidBrush()` directly inside the `WM_CTLCOLOR*` handlers.

```cpp
// In TDSubclassContainer
SetWindowSubclassIfNeeded(hwndParent, TDCtrlContainerSubclassProc, kTDCtrlSubclassId, 0);

// In TDCtrlContainerSubclassProc WM_CTLCOLORSTATIC handler
return reinterpret_cast<LRESULT>(GetSolidBrush(TDDarkCol::kSecondary));
```

### ❓ Issue 2: The task dialog's button container background may also be affected
A similar corruption might occur inside the task dialog itself. The same possible cause and suggested fix apply as above.

https://github.com/user-attachments/assets/f69f2879-146e-4bc8-ba67-d9f85b64a3ee


### 🚀 Suggestion for better performance (optional – only if you need more performance)
Currently s.elemsOk = false is set inside WM_PAINT, forcing a full UI Automation tree walk on every repaint. For better performance, this should be moved out of WM_PAINT and triggered only when actual state changes (expando click, verification toggle, text/icon update).

Step 1 – Define custom reflection messages (add at the top of taskdlg.cpp):

```cpp
static constexpr UINT WM_TD_REFLECT_EXPANDO = WM_USER + 5001;
static constexpr UINT WM_TD_REFLECT_CHECK   = WM_USER + 5002;
static constexpr UINT WM_TD_REFLECT_UPDATE  = WM_USER + 5003; // text/icon changes
```
Step 2 – In wxTaskDialogCallback, send these messages:

```cpp
case TDN_EXPANDO_BUTTON_CLICKED:
    ::SendMessageW(hwnd, WM_TD_REFLECT_EXPANDO, wParam, 0);
    break;

case TDN_VERIFICATION_CLICKED:
    ::SendMessageW(hwnd, WM_TD_REFLECT_CHECK, wParam, 0);
    break;

case TDM_SET_ELEMENT_TEXT:   // 1132
case TDM_UPDATE_ELEMENT_TEXT: // 1138
case TDM_UPDATE_ICON:         // 1140
    ::SendMessageW(hwnd, WM_TD_REFLECT_UPDATE, 0, 0);
    break;
```
Step 3 – In TDPageSubclassProc, handle the messages and invalidate the cache:

```cpp
case WM_TD_REFLECT_EXPANDO:
    {
        TDPageState& s = TDPageState::Get(hwnd);
        s.isExpanded = (wParam != 0);
        s.elemsOk = false;  // Invalidate cache only here, not in WM_PAINT
        ::InvalidateRect(hwnd, nullptr, FALSE);
    }
    return 0;

case WM_TD_REFLECT_CHECK:
    {
        TDPageState& s = TDPageState::Get(hwnd);
        s.isChecked = (wParam != 0);
        s.elemsOk = false;
        ::InvalidateRect(hwnd, nullptr, FALSE);
    }
    return 0;

case WM_TD_REFLECT_UPDATE:
    {
        TDPageState& s = TDPageState::Get(hwnd);
        s.elemsOk = false;
        ::InvalidateRect(hwnd, nullptr, FALSE);
    }
    return 0;
```
Step 4 – Remove s.elemsOk = false from WM_PAINT – the paint handler should only call TDUpdateLayoutCache() without forcing a rebuild.

This change eliminates unnecessary UI Automation walks during mouse moves, hover effects, and simple redraws, improving performance especially for progress dialogs and animations.

### ℹ️ COM initialization warning (not a code issue)

> @MaartenBent I don't see any leaks when showing the "About" dialog in the "dialogs" sample and ASAN doesn't report any problems, how can I reproduce this?
> 
> I do get weird messages like this in the debugger output
> 
> ```
> mincore\com\oleaut32\dispatch\ups.cpp(2122)\OLEAUT32.dll!00007FFF0964A726: (caller: 00007FFF09649B39) ReturnHr(1) tid(36dc) 8002801D Library not registered.
> ```
> 
> when instantiating `IUIAutomation`. They seem to be harmless but still suspicious and might indicate that we do something wrong, but I don't see what (the TID in the message above is that of the main thread, FWIW).


https://github.com/wxWidgets/wxWidgets/pull/26311#issuecomment-4642740971

I saw the same 8002801D warning in my development environment. I remember that after running regsvr32 oleaut32.dll once (as administrator), the warning went away. So this is not a bug in the PR, just a note for anyone who sees it.

include/wx/msw/private/taskdlg.h
vadz: > I was sick for a while and couldn't work on it

Sorry to hear this!

> but I'm fine now 

And glad to hear that you're better!

> Thank you also for adding my name as the original author in include/wx/msw/private/taskdlg.h – I noticed that and I'm grateful.

This is perfectly normal, thank you for writing all this code!

> If it's not too much trouble, could you also add my GitHub nickname there? 

I've modified these lines to read

```cpp
// Author:      Mohmed Abdel-Fattah (memoarfaa)
```

in both files.

I'll push this soon, when I read/deal with the other problems.

vadz:
> Issue 1: The application's menu and background brush appear to be destroyed after the dialog closes

Oops, thanks a lot for noticing this. I think the real problem is that we set our brush as the class `HBRUSH` — which means it will be deleted by the class. I think I've fixed this now.

> Issue 2: The task dialog's button container background may also be affected

I think it's the same issue and so also fixed, but please let me know if you still see this (somehow I can't play the video right now, it looks like there is a problem with this on GitHub...).

> Suggestion for better performance (optional – only if you need more performance)

I think this could be indeed worth doing it, but I won't do it now — please open new PRs once this one is merged if you can implement this. TIA!

> COM initialization warning (not a code issue)
> 
> > I do get weird messages like this in the debugger output
> > ```
> > mincore\com\oleaut32\dispatch\ups.cpp(2122)\OLEAUT32.dll!00007FFF0964A726: (caller: 00007FFF09649B39) ReturnHr(1) tid(36dc) 8002801D Library not registered.
> > ```
> 
> I saw the same 8002801D warning in my development environment. I remember that after running regsvr32 oleaut32.dll once (as administrator), the warning went away. So this is not a bug in the PR, just a note for anyone who sees it.

I don't think it should be necessary to register `oleaut32.dll` — this is a system DLL which really should be already registered (and I'm pretty sure it is, although I'm too lazy to check this). But I don't see what are we doing wrong and this doesn't seem to result in any actual problems, so I'm leaving this for now. But I'm still almost sure it's a bug in our code.

vadz:
Please let me know if anybody else is going to test this in near future, otherwise I'll merge it soon to get as much testing as possible before 3.3.3.

src/msw/taskdlg.cpp
```diff
@@ -0,0 +1,1283 @@
...
+        TDLayoutElement info;
+        pChild->get_CurrentBoundingRectangle(&info.rect);
+        ::ScreenToClient(hwnd, reinterpret_cast<POINT*>(&info.rect.left));
+        ::ScreenToClient(hwnd, reinterpret_cast<POINT*>(&info.rect.right));
+
+        info.automationId = GetCurrentAutomationId(pChild);
+
+        BSTR b;
+        if ( pChild->get_CurrentName(&b) == S_OK )
+            info.name = b;
```
memoarfaa, line 416: progress dialog is not work becuase of this line get_CurrentName "p" can be null it throw Access violation reading location replace it with 
```
 BSTR b;
 if (pChild->get_CurrentName(&b) == S_OK)
 {
     if (b)
     {
         info.name = b;
         ::SysFreeString(b);
     }
 }
```

src/msw/taskdlg.cpp
```diff
@@ -0,0 +1,1283 @@
...
+}
+
+// Another helper to get the automation ID of an element.
+std::wstring GetCurrentAutomationId(IUIAutomationElement* element)
+{
+    std::wstring result;
+
+    BSTR bstr;
+    if ( element->get_CurrentAutomationId(&bstr) == S_OK )
+        result = bstr;
```
memoarfaa, line 141: here also bstr can be null 
``` 
BSTR bstr;
if (element->get_CurrentAutomationId(&bstr) == S_OK)
{
    if (bstr)
    {
        result = bstr;
        ::SysFreeString(bstr);
    }
}
```

memoarfaa (changes requested):
Fix progress dialog Access violation reading location at start up.

src/msw/taskdlg.cpp
```diff
@@ -0,0 +1,1283 @@
...
+
+    ::EnumChildWindows(hwndTD, TDEnumAttachProc, reinterpret_cast<LPARAM>(&data));
+
+    if ( !data.found )
+        return;
+
+    if ( native )
+        wxMSWDarkMode::AllowForWindow(hwndTD, L"DarkMode_Explorer", nullptr);
+
+    SetWindowSubclassIfNeeded(hwndTD, TDCtrlContainerSubclassProc, kTDMainSubclassId);
```
memoarfaa: this shoud add dark Background Brush for main dialog becuse it's Appear at progress dialog

```
 COLORREF bg = TDDarkCol::kPrimary;
 SetWindowSubclassIfNeeded
 (
     hwndTD,
     TDCtrlContainerSubclassProc,
     kTDMainSubclassId,
     [bg]() { return reinterpret_cast<DWORD_PTR>(GetSolidBrush(bg)); }
 );
```


Image

src/msw/taskdlg.cpp
vadz: Oops, thanks for noticing this, I forgot to retest with the progress dialog after making the changes. Fixed in 6132c1705977df7dc5ba43aba47cc1408e120285 now.

src/msw/taskdlg.cpp
vadz: Fixed in the same commit too, thanks.

src/msw/taskdlg.cpp
vadz: Fixed in aeef45946c5eb4f912b04ef0fae3bf54c7dc1919 where I also removed setting the class background brush because this doesn't seem to be necessary and it could have resulted in the same problem we've seen before.

Thanks again for testing this dialog too!

src/msw/taskdlg.cpp
```diff
@@ -0,0 +1,1321 @@
...
+
+        WinStructWordSize<DTTOPTS> opts;
+        if ( !native )
+        {
+            opts.dwFlags = DTT_COMPOSITED | DTT_TEXTCOLOR;
+            opts.crText = TDGetTextColour(s, part);
+
+            ::FillRect(hdc, &rcText, brBg);
+        }
+
```
memoarfaa, line 755: Debug Alert dialog need draw Background of Expand Area.
```
  if (!native)
 {
     opts.dwFlags = DTT_COMPOSITED | DTT_TEXTCOLOR;
     opts.crText = TDGetTextColour(s, part);

     ::FillRect(hdc, &rcText, brBg);
 }
 else
 {
    // Debug Alert dialog Expand Footer Area need this.
     if (part == TDLG_EXPANDEDFOOTERAREA)
     {
         DTBGOPTS opts = {sizeof(opts)};
         opts.dwFlags = DTBG_OMITBORDER;
         DrawThemeBackgroundEx(s.hTD, hdc, TDLG_SECONDARYPANEL, 0, &el.rect, &opts);
     }
 }
 ::DrawThemeTextEx(s.hTD, hdc, part, 0, el.name.c_str(), -1, dtF, &rcText, &opts);
```

src/msw/taskdlg.cpp
```diff
@@ -0,0 +1,1321 @@
...
+                isFn && !TDHasNativeDarkTheme()
+                    ? TDDarkCol::kFootnote
+                    : TDDarkCol::kPrimary
+            );
+        }
+
+        return TRUE;
+    }
+
+    // Main TaskPage (DirectUI "TaskDialog" class)
```
memoarfaa, line 1206: Set Debug Alert dialog Scrollbar dark Theme

```
 if (wcscmp(cls, L"CCVScrollBar") == 0 || wcscmp(cls, L"CCHScrollBar") == 0)
 {
     if (const HWND hScroll = GetHWNDFromElement(pEl))
     {
         wxMSWDarkMode::SetTheme(hScroll, L"DarkMode_Explorer");
     }
     return TRUE;
 }
```

memoarfaa (changes requested):
> 
> Thanks again for testing this dialog too!

latest dialog to finish this  is Debug Alert dialog need draw Background and set Scrollbar dark Theme 
to see it with this image use dialog simple ->Message Dialog then choose both Yes and Ok Buttons

* ### Before 

Image

* ###  After

Image

src/msw/taskdlg.cpp
vadz: I don't actually see this problem here (10.0.26200.8655), but I've applied this patch (committed under your name as I applied it almost unchanged), thanks.

src/msw/taskdlg.cpp
vadz: Applied too, thanks (I do see this one).

vadz:
Changed pushed, thanks again! For the record, I've used the following diff to test (both to make showing the dialog simpler and because the default max stack of 20 was too little to show scrollbar on my systems):

```diff
diff --git a/src/common/appbase.cpp b/src/common/appbase.cpp
index 1cd9b35243..4417ce87a3 100644
--- a/src/common/appbase.cpp
+++ b/src/common/appbase.cpp
@@ -1065,7 +1065,7 @@ wxString wxAppTraitsBase::GetAssertStackTrace()
             // don't show more than maxLines or we could get a dialog too tall
             // to be shown on screen: 20 should be ok everywhere as even with
             // 15 pixel high characters it is still only 300 pixels...
-            if ( m_numFrames++ > 20 )
+            if ( m_numFrames++ > 200 )
                 return;

             m_stackTrace << wxString::Format(wxT("[%02u] "), m_numFrames);
diff --git a/samples/dialogs/dialogs.cpp b/samples/dialogs/dialogs.cpp
index a984f468fb..3c646037a9 100644
--- a/samples/dialogs/dialogs.cpp
+++ b/samples/dialogs/dialogs.cpp
@@ -1095,8 +1095,16 @@ void MyFrame::MessageBoxWindowModalClosed(wxWindowModalDialogEvent& event)
     delete dialog;
 }

+void recurse(int n)
+{
+    wxCHECK_RET(n > 0, "n should be non-negative");
+
+    recurse(n - 1);
+}
+
 void MyFrame::MessageBoxDialog(wxCommandEvent& WXUNUSED(event))
 {
+    recurse(40); return;
     TestMessageBoxDialog dlg(this);
     dlg.Create();
     dlg.ShowModal();
```

PBfordev:
If the task dialog now supports dark mode, we should probably remove this from the docs:
https://github.com/wxWidgets/wxWidgets/blob/aef89c0e875d67738b44f5557a55737f74df434f/interface/wx/app.h#L1436-L1440

src/msw/taskdlg.cpp
memoarfaa: @vadz 
> I don't actually see this problem here (10.0.26200.8655), but I've applied this patch (committed under your name as I applied it almost unchanged), thanks.

### Note  : 
It's actually not an OS version issue, but rather a DPI scaling issue. To reproduce the expanded footer area paint bug and see the scrollbar, you need to change your screen display settings to a higher DPI layout, such as 125% or 150%.We need to draw over the old text, and DrawThemeBackgroundEx is required here to properly erase the old text from the background. Without it, due to the tiny margin differences at higher DPIs, the old text doesn't get cleared and remains visible underneath.


* 96 dpi image from same system

Screenshot 2026-06-12 192411

vadz: I see, thanks. I did test this in 200% scaling, but you probably need to use fractional scaling to see the problem, which explains why I didn't.

vadz:
> If the task dialog now supports dark mode, we should probably remove this from the docs:

Yes indeed, thanks for the reminder, done (locally).

memoarfaa:
> If the task dialog now supports dark mode, we should probably remove this from the docs:
> 
> https://github.com/wxWidgets/wxWidgets/blob/aef89c0e875d67738b44f5557a55737f74df434f/interface/wx/app.h#L1436-L1440

@vadz 
@PBfordev Thanks for pointing that out!According to the official documentation for[ MSWEnableDarkMode](https://docs.wxwidgets.org/latest/classwx_app.html#af8c93d7e3345e62a58325f3ab1d158d6) 

``` 
Known limitations of dark mode support include:
 * Anything based on TaskDialog() Win32 API doesn't support dark mode... [Fixed in this PR]
 * The following dialogs wrapping common windows dialogs don't support dark mode: wxColourDialog, wxFindReplaceDialog, wxFontDialog, wxPageSetupDialog, wxPrintDialog.
* wxTimePickerCtrl, wxDatePickerCtrl and wxCalendarCtrl don't support dark mode and use the same (light) background as by default in it.
* Toolbar items for which wxToolBar::SetDropdownMenu() was called don't draw the menu drop-down correctly, making it almost invisible.
```

🗓️ wxCalendarCtrl can be  fixed!using just a one-liner code As you can see in the attached screenshot .
Screenshot 2026-06-12 195529

* #### I plan to work on fixing the toolbar dropdown issue and the common dialogs listed in the documentation in my free time. To help me get started on the toolbar dropdown bug, could you guide me on how to reproduce it? If there is an existing sample or a specific setup I should use to see it, please let me know!

PBfordev:
@memoarfaa, thanks for the work on this (and other things). IMO, this significantly improves the dark mode experience on Windows.

>  To help me get started on the toolbar dropdown bug

I think that may have been fixed without updating the docs by https://github.com/wxWidgets/wxWidgets/commit/c301af42f923a9997e99bf8d3b8f00df667f88f9

memoarfaa:
> @memoarfaa, thanks for the work on this (and other things). IMO, this significantly improves the dark mode experience on Windows.
> 
> > To help me get started on the toolbar dropdown bug
> 
> I think that may have been fixed without updating the docs by [c301af4](https://github.com/wxWidgets/wxWidgets/commit/c301af42f923a9997e99bf8d3b8f00df667f88f9)

for toolbars there is new better themes availble now at 
windows 11 24H2/25H2 Build 26220.6772 and newer the "DrakMode_DrakTheme". 

Screenshot 2026-06-12 215123

vadz:
I'm going to merge this, but please don't hesitate to open new PRs for further improvements. TIA!

Issue #26529, opened 2026-05-29
Text alignment in wxSearchCtrl (MSW, Dark Mode)


BazaarAcc32:
When I used wxSearchCtrl, I noticed that the text inside it was vertically centered in light mode, but in dark mode, the vertical alignment not centered.

It would be nice if the text inside wxSearchCtrl was also vertically centered in dark mode.

OS: Windows 11
wxWidgets version: 3.3.0

Example code:
```cpp
#include <wx/wx.h>
#include <wx/srchctrl.h>

class MyFrame : public wxFrame
{
public:
    MyFrame()
        : wxFrame(nullptr, wxID_ANY, "wxSearchCtrl Example",
            wxDefaultPosition, wxSize(400, 200))
    {
        wxBoxSizer* sizer = new wxBoxSizer(wxVERTICAL);

        wxSearchCtrl* search = new wxSearchCtrl(this, wxID_ANY);
        search->ShowSearchButton(true);
        search->SetMargins(wxPoint(10, 10)); // does not work for vertical direction

        sizer->Add(search, 0, wxEXPAND);

        SetSizer(sizer);
    }
};

class MyApp : public wxApp
{
public:
    bool OnInit() override
    {
        MSWEnableDarkMode(wxApp::DarkMode_Always);
        MyFrame* frame = new MyFrame();
        frame->Show();
        return true;
    }
};

wxIMPLEMENT_APP(MyApp);
```

**UPDATE** Same problem with wxTextCtrl

vadz:
The text is centered just fine for me by default, but it is not when `SetMargins()` is used. I think this is a pretty useless function, so my advice would be to just stop using it.

Just in case, here is how this looks for me (with `SetMargins()` call, removing it fixes the problem):

<img width="230" height="85" alt="Image" src="https://github.com/user-attachments/assets/0660f1e6-b017-41a3-b66c-555d9996b31d" />

BazaarAcc32:
@vadz If I remove SetMargins(), problem still exists in dark mode. Here is how this looks for me:

<img width="383" height="184" alt="Image" src="https://github.com/user-attachments/assets/d758c945-1a93-451f-a1cb-a8cdfc41bf6a" />

In light mode everything is OK

vadz:
Can you show light mode side by side (please prefer to paste your screenshots directly here instead of using external links, I've edited the screenshot above to do this)?

BazaarAcc32:
[...]

PR #26571, opened 2026-06-08
MSW: Dark mode borders

stevecor:
For Windows dark mode, draw themed borders instead of translating wxBORDER_THEME to wxBORDER_SIMPLE. Borders have the same thickness between light mode and dark mode for consistent alignment and positioning. See #26529.

In dark mode, the border styles wxBORDER_STATIC, wxBORDER_RAISED and wxBORDER_SUNKEN are sometimes drawn by the system using light mode colours. So draw these borders  similarly to themed borders, but using the proper width. There is no attempt to mimic a raised or sunken look.

The WM_NCCALCSIZE and WM_NCPAINT message handling is simplified and corrected for calculating the border width.

### DIFF
--- a/src/msw/window.cpp
+++ b/src/msw/window.cpp
@@ -1466,14 +1466,6 @@ wxBorder wxWindowMSW::DoTranslateBorder(wxBorder border) const
 {
     if (border == wxBORDER_THEME)
     {
-        // In dark mode the standard sunken border is too bright, so prefer
-        // using a simple(r) and darker border instead.
-        //
-        // And themed borders don't look good either in dark mode, so don't
-        // use them in it.
-        if ( wxMSWDarkMode::IsActive() )
-            return wxBORDER_SIMPLE;
-
         if (CanApplyThemeBorder())
         {
             if ( wxUxThemeIsActive() )
@@ -3820,6 +3812,8 @@ wxWindowMSW::MSWHandleMessage(WXLRESULT *result,
         // If we want the default themed border then we need to draw it ourselves
         case WM_NCCALCSIZE:
             {
+                // The default handling for this message is proper for all
+                // border styles except wxBORDER_THEME.
                 if (DoTranslateBorder(GetBorder()) == wxBORDER_THEME)
                 {
                     // first ask the widget to calculate the border size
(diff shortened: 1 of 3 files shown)


src/msw/window.cpp
```diff
@@ -3839,93 +3833,65 @@ wxWindowMSW::MSWHandleMessage(WXLRESULT *result,
...
 
+                    // Draw the theme border and background.
+
+                    // The EDIT theme gives a good general purpose border in light mode.
+                    // There does not seem to be a dark mode EDIT theme that looks good.
+                    // The ListView theme below looks good in dark mode.
+                    wxUxThemeHandle hTheme(this, L"EDIT", L"DarkMode_DarkTheme::ListView");
+                    // The part and state we use are the same values for EDIT and ListView.
+                    static_assert(EP_EDITTEXT == LVP_LISTITEM);
+                    static_assert(ETS_NORMAL == LISS_NORMAL);
```
vadz: Need to cast to int to avoid warnings and probably should provide a message to avoid requiring C++14.

src/msw/window.cpp
```diff
@@ -3839,93 +3833,65 @@ wxWindowMSW::MSWHandleMessage(WXLRESULT *result,
...
+                    InflateRect(&rcClient, -borderWidth, -borderWidth);
                     ::ExcludeClipRect(GetHdcOf(*impl), rcClient.left, rcClient.top,
                                       rcClient.right, rcClient.bottom);
 
+                    // Draw the theme border and background.
+
+                    // The EDIT theme gives a good general purpose border in light mode.
+                    // There does not seem to be a dark mode EDIT theme that looks good.
+                    // The ListView theme below looks good in dark mode.
+                    wxUxThemeHandle hTheme(this, L"EDIT", L"DarkMode_DarkTheme::ListView");
```
vadz: I wonder if it could be worth caching this theme to avoid opening/closing it all the time. I have no idea how expensive this is, so maybe it's not worth it, but OTOH keeping it open (as a global var) should be pretty simple.

src/msw/window.cpp
```diff
@@ -3839,93 +3833,65 @@ wxWindowMSW::MSWHandleMessage(WXLRESULT *result,
...
-                    InflateRect(&rcClient, -1, -1);
-                    if (wParam)
-                        csparam->rgrc[0] = rcClient;
-                    else
-                        *((RECT*)lParam) = rcClient;
-
-                    // WVR_REDRAW triggers a bug whereby child windows are moved up and left,
-                    // so don't use.
-                    // rc.result = WVR_REDRAW;
+                    const auto borderWidth = GetWindowBorderSize().x / 2;
```
vadz: I wonder if it could make sense to add some `MSWGetBorderThickness()` returning just this, i.e. a single int, not multiplied by 2, unlike `GetWindowBorderSize()` which returns a `wxSize` multiplied by it.

The idea being is that `/ 2` begs the question of what happens when the value is odd, which actually can never happen, but this is not really obvious.

src/msw/window.cpp
```diff
@@ -3839,93 +3833,65 @@ wxWindowMSW::MSWHandleMessage(WXLRESULT *result,
...
+                        drawBorder = true;
+                        break;
+                    case wxBORDER_STATIC:
+                    case wxBORDER_RAISED:
+                    case wxBORDER_SUNKEN:
+                        // In dark mode, explicitly draw these border styles because
+                        // the default drawing uses light mode colours.
+                        drawBorder = wxMSWDarkMode::IsActive();
+                        break;
+                }
```
vadz, line 3863: I'd add the cases for the other values (`SIMPLE` and `NONE`) just to make it clear that we do handle them here.

vadz (commented):
Thanks, I didn't test this yet but the changes look good to me.

src/msw/window.cpp
stevecor: I timed this with wxStopWatch and got just 11µs for 242 calls.

vadz:
Thanks for the update, will merge soon.

Issue #26569, opened 2026-06-07
wxMSW CheckBox dark mode apperance inconsistent


ryancog:
@stevecor With your recent PR #26516 it seems the `wxCheckBox` isn't quite functioning right with respect to appearance.

If you start the widgets sample (Adding `MSWEnableDarkMode()` to `OnInit()`) with Windows in dark mode, you can see the checkbox text has a dark/black text color.

<img width="131" height="45" alt="Image" src="https://github.com/user-attachments/assets/dd840447-691d-44aa-9b13-8b1a196a83ba" />

If you toggle to light mode, then toggle back, the appearance is correct:

<img width="131" height="45" alt="Image" src="https://github.com/user-attachments/assets/a5ca3a06-7737-4213-ae3c-0fc0bee98da8" />

If you click "Reset" to rebuild the control, it goes back to having the dark text color.

I'm not familiar with the details of your PR, but I guess something isn't being considered on initial creation here. Hopefully it's trivial.

### Platform and version information

- wxWidgets version  you use: master
- wxWidgets port  you use: wxMSW
- OS  and its version: Windows 10

ryancog:
Looks like this is an issue for RadioBox too

<img width="108" height="39" alt="Image" src="https://github.com/user-attachments/assets/c5c8ae68-2bc3-4dc0-a258-ef46611d195f" />

This is on the slider page. Interestingly, it starts looking okay, but once you check "Show value label" to enable the box, the text becomes dark colored. Going back to disabled doesn't fix it (i.e. it stays dark).

Disabling the box on the Radio page (with the Widget menu option) though doesn't seem to do it, but it looks like the foreground and background colors are explicitly set when that occurs.

stevecor:
I am not able to reproduce this on Windows 10.0.19045. What version of Windows 10 are you using?

ryancog:
@stevecor 

Edition	Windows 10 Pro
Version	22H2
Installed on	‎10/‎23/‎2025
OS Build	19045.6466

So I guess the same version. That's odd

stevecor:
Here are the steps I followed. Can describe if this is right?

1. Put system in dark mode
2. Add MSWEnableDarkMode(DarkMode_Auto) and run
3. Click on "Slider"
4. Click on "Show value label"

<img width="172" height="151" alt="Image" src="https://github.com/user-attachments/assets/0487e0e5-f30b-467d-bf87-1a81a0043de0" />

[...]

PR #26572, opened 2026-06-08
MSW: Dark mode switching for wxCheckBox and wxRadioButton

stevecor:
MSWSetDarkOrLightMode() always sets the text color, for Windows 10. See #26569.

@ryancog I was not able to reproduce one of your two symptoms. Please test this fix.

### DIFF
--- a/include/wx/msw/ownerdrawnbutton.h
+++ b/include/wx/msw/ownerdrawnbutton.h
@@ -39,6 +39,9 @@ protected:
     // pointers to this class (and protected dtor enforces this).
     ~wxMSWOwnerDrawnButtonBase() = default;
 
+    // Make the control owner drawn or reset it to normal style.
+    void MSWMakeOwnerDrawn(bool ownerDrawn);
+
     // Make the control owner drawn if necessary to implement support for the
     // given foreground colour.
     void MSWMakeOwnerDrawnIfNecessary(const wxColour& colFg);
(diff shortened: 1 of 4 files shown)


ryancog:
This does seem to fix things for the checkbox, but I still (unsurprisingly) see the issue with the `wxRadioBox`.

vadz:
Thanks, this fixes the problem, but I don't understand why returning `wxSYS_COLOUR_BTNTEXT` from `wxControl::GetClassDefaultAttributes()` is not enough. Have you already looked into this and know why is it so by chance? If not, I'd like to do it.

ryancog:
Perhaps it's unrelated, and certainly an aside, but I wonder if it's at all related to returning `wxTransparentColor` in `wxControl` not working for making all the OSX controls background transparent (for the tinted effect, some controls would retain the opaque background even when it seemed they should've pulled from `wxControl`'s attires).

There were at least a couple controls where the attributes didn't apply, and I couldn't see any obvious place that would've caused it so. I believe `wxDateTimePickerCtrl` was one that was odd to me, since if I recall correctly it inherits from `wxControl` relatively plainly.

Perhaps there's something (subtle?) in code that's not port-specific...

stevecor:
> Thanks, this fixes the problem, but I don't understand why returning `wxSYS_COLOUR_BTNTEXT` from `wxControl::GetClassDefaultAttributes()` is not enough. Have you already looked into this and know why is it so by chance? If not, I'd like to do it.

The reason we were using SetForegroundColour() is not to change the color but to put the control into owner-draw mode. I've made MSWMakeOwnerDrawn() accessible and called that instead.

vadz:
Ah, this makes much more sense, thanks.

But now I have another question: should we switch off `BS_OWNERDRAW` when switching to light mode? Normally there should be no need to leave the checkbox owner drawn if its colours haven't been changed.

vadz:
Looks perfect, thanks!

PR #26575, opened 2026-06-08
MSW: Fix dark mode wxRadioBox disabled text color

stevecor:
When a wxRadioBox is enabled or disabled, send WM_THEMECHANGED to force an update of the text color. See #26569.

This problem appeared in 56bd4a629f66422f9f72662946870e1c4ae819ec, which was to stop double buffering. Apparently, this exposed a quirk in the Windows painting/theming.

### DIFF
--- a/include/wx/msw/statbox.h
+++ b/include/wx/msw/statbox.h
@@ -78,8 +78,6 @@ public:
 protected:
     virtual wxWindowList GetCompositeWindowParts() const override;
 
-    virtual void MSWSetDarkOrLightMode(SetMode setmode) override;
-
     // return the region with all the windows inside this static box excluded
     WXHRGN MSWGetRegionWithoutChildren();
 
--- a/src/msw/radiobox.cpp
+++ b/src/msw/radiobox.cpp
@@ -35,6 +35,7 @@
     #include "wx/tooltip.h"
 #endif // wxUSE_TOOLTIPS
 
+#include "wx/msw/private/darkmode.h"
 #include "wx/msw/wrapwin.h"
 
 // TODO: wxCONSTRUCTOR
@@ -406,6 +407,9 @@ bool wxRadioBox::Enable(bool enable)
 
     for ( auto button : m_radioButtons )
         button->Enable(enable);
+
+    if ( wxMSWDarkMode::IsActive() )
+        ::SendMessage(m_hWnd, WM_THEMECHANGED, 0, 0);
 
     return true;
 }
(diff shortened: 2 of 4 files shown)


vadz:
Sorry, I don't understand why is this done for `wxRadioBox` when it looks like the problem is at `wxStaticBox` level. Moreover, as we paint the static box entirely ourselves, it should be possible to just fix the bug in our code instead of triggering any extra messages.

If you haven't looked at this code yet, I'll try to do it soon.

ryancog:
Fwiw, it looks like the text being dark/black isn't the only anomaly. The border becomes a much brighter color (appears relatively white) vs how it's normally drawn in dark mode.

Image

You can see it a little in my screenshot in the linked issue (look at the color of the box immediately above), but I didn't notice it before since I was focused on the text.

stevecor:
The painting is fixed so that we no longer need to use SetForegroundColour().

Part of the problem was not with the painting itself but that no paint event occurred after enabling/disabling the control. I think all windows should get a Refresh() upon enable/disable. So that change is at a low level in wxWindowMSW.

This PR fixes the wxRadioBox only. This fix exposed an existing problem with wxRadioButton. That will be fixed in #26572, as it is a similar issue and requires some of the changes there.

vadz:
Thanks, this makes much more sense, but I am very surprised that `::EnableWindow()` doesn't result in a `WM_PAINT` — are you sure about this?

FWIW we could call `Refresh()` in `wxStaticBox::MSWWindowProc()` on getting `WM_ENABLE` if the problem is specific to `wxStaticBox`, but even this shouldn't be normally necessary...

stevecor:
I do believe WM_PAINT is not guaranteed when a window's enable status changes. The documentation for 
EnableWindow() mentions a couple resulting messages, but not WM_PAINT. The documentation of WM_PAINT indicates painting only occurs upon invalidation, no mention of status changes. Realistically the issue probably only affects owner-draw windows.

My concern with sprinkling Refresh() calls only where we think they are needed is that we might miss something and have another surprise. It's a good bet that when a window is disabled or enabled, it changes appearance and there is going to be a paint event anyway. Multiple invalidations do little harm - there is only one resulting paint event.

vadz:
My concern is that some big window which does _not_ change appearance may start flickering now when its parent is enabled/disabled, for example.

vadz:
Thank you, will merge in a moment (even keeping "color" spelling in the commit message, at a great personal cost :-).

PR #26576, opened 2026-06-10
Correct translation of "Del" key in some languages

taler21:
Use the key name that is used by Windows itself (for example, on its on-screen keyboard as well as an accelerator in its notepad).

While I am not a native speaker of any of these languages, I think Windows and its applications are a good and reliable source for which keyboard key name should be used in each language.

For the record, the keyboard key names used by Windows were determined under Windows 10.

### DIFF
--- a/locale/da.po
+++ b/locale/da.po
@@ -226,7 +226,7 @@ msgstr "Slet"
 #: ../src/common/accelcmn.cpp:48
 msgctxt "keyboard key"
 msgid "Del"
-msgstr "Slet"
+msgstr "Del"
 
 #: ../src/common/accelcmn.cpp:49
 msgctxt "keyboard key"
--- a/locale/es.po
+++ b/locale/es.po
@@ -221,7 +221,7 @@ msgstr "Eliminar"
 #: ../src/common/accelcmn.cpp:48
 msgctxt "keyboard key"
 msgid "Del"
-msgstr "Eliminar"
+msgstr "Supr"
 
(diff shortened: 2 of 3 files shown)


locale/da.po
```diff
@@ -226,7 +226,7 @@ msgstr "Slet"
 #: ../src/common/accelcmn.cpp:48
 msgctxt "keyboard key"
 msgid "Del"
-msgstr "Slet"
+msgstr "Del"
```
vadz, line 229: I'm a bit worried to change this because it looks like "Slet" was used by actual Danish speakers, see e.g. cb81c2e914 (Danish translations update from Bue Vester-Andersen., 2015-07-21). Are we sure that Danish keyboards use "Del"?

vadz (commented):
I don't speak any of these languages either, but while Spanish fix looks good to me and I have no idea about Hungarian, I'm not sure we want to apply the Danish one.

Any Danes reading this?

locale/da.po
vadz: FWIW https://support.apple.com/da-dk/102650 uses "Slet" for backspace.

taler21: "Slet" is the correct translation of "Delete", but the key on the keyboard is labelled "Del".
I asked a Danish client and he confirmed that the key on the keyboard is labelled "Del".

A screenshot of Windows' Notepad on a Danish Windows:

Notepad-Win10-Danish

vadz: Thanks for taking time to check this!

PR #26577, opened 2026-06-10
reject out-of-range parent index in LoadCachedBook

dxbjavid:
LoadCachedBook reads each index entry's parent back-reference straight from the .cached file and uses it as m_index[m_index.size() - parentShift] with no check, so a crafted cached file (which can sit inside a .htb help archive opened through AddBook) makes the subtraction wrap and leaves the parent pointer outside the array; that pointer is later dereferenced when the index is sorted at the end of AddBookParam. The change rejects the file when parentShift is negative or larger than the number of index entries loaded so far, matching the existing version and flags checks just above. There's a small test under tests/html that feeds such a cached stream and confirms it is now refused rather than reading out of bounds.

### DIFF
--- a/build/cmake/tests/gui/CMakeLists.txt
+++ b/build/cmake/tests/gui/CMakeLists.txt
@@ -98,6 +98,7 @@ set(TEST_GUI_SRC
     image/image.cpp
     image/imagwebp.cpp
     image/rawbmp.cpp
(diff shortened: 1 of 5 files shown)


vadz:
Nice find, thanks, will merge soon.

dxbjavid:

PR #26578, opened 2026-06-10
OSX macOS 27 adaptions

csomor:
Collect changes needed to compile on macOS 27 with Xcode 27


vadz:
@csomor Stefan, should this be merged? Or is it still a draft?

csomor:
> @csomor Stefan, should this be merged? Or is it still a draft?

Thanks @vadz, I thought I'd wait till the second beta (of macOS 27), and collect what comes up until then.

csomor:
@vadz ok, for Xcode based builds the minimum deployment is indeed macOS 12 Monterey, while command line builds still accept `-with-macosx-version-min=11.00` I don't know whether we'd have some logic somewhere that could adjust the `10.10` minimum version to `11.00` when building against macOS SDK 27 with `__MAC_OS_X_VERSION_MAX_ALLOWED` from `Availability.h` being at least `270000`

vadz:
There is some code in configure setting min version from the SDK deployment version, see https://github.com/wxWidgets/wxWidgets/blob/0a6b26ea435f69fb8ea9fd764c1e06cdc148587c/configure.ac#L1314-L1337

I wonder what does it output for you on the new system after "Checking for SDK deployment version"?

csomor:
> I wonder what does it output for you on the new system after "Checking for SDK deployment version"?

if I give a `-with-macosx-sdk=yes` then the `DefaultProperties` of `SDKSettings` are read, but unfortunately this is an array having all possible deployment targets exposed, and the current code picks the last ie highest

vadz:
> > I wonder what does it output for you on the new system after "Checking for SDK deployment version"?
> 
> if I give a `-with-macosx-sdk=yes` then the `DefaultProperties` of `SDKSettings` are read, but unfortunately this is an array having all possible deployment targets exposed, and the current code picks the last ie highest

Assuming the targets are always sorted, we could use `sed 1,DEPLOYMENT_TARGET_SUGGESTED_VALUES/d | head -n 1` to get the first of them, would this be the right thing to use?

csomor:
the `defaults read ... DefaultProperties` gives us

```
{
    "AD_HOC_CODE_SIGNING_ALLOWED" = YES;
    "CODE_SIGNING_REQUIRED" = YES;
    "CODE_SIGN_ENTITLEMENTS" = "";
    "CODE_SIGN_IDENTITY" = "Apple Development";
    "DEFAULT_COMPILER" = "com.apple.compilers.llvm.clang.1_0";
    "DEPLOYMENT_TARGET_SUGGESTED_VALUES" =     (
        "12.0",
        "12.2",
        "12.3",
        "12.4",
        "13.0",
        "13.1",
        "13.2",
        "13.3",
        "13.4",
        "13.5",
        "14.0",
        "14.1",
        "14.2",
        "14.3",
        "14.4",
        "14.5",
        "14.6",
        "15.0",
        "15.1",
        "15.2",
        "15.3",
        "15.4",
        "15.5",
        "15.6",
        "26.0",
        "26.1",
        "26.2",
        "26.3",
        "26.4",
        "26.5",
        "26.6",
        "27.0"
    );
    "ENTITLEMENTS_DESTINATION" = Signature;
    "ENTITLEMENTS_REQUIRED" = YES;
    "IOS_UNZIPPERED_TWIN_PREFIX_PATH" = "/System/iOSSupport";
    "KASAN_CFLAGS_CLASSIC" = "-DKASAN=1 -DKASAN_CLASSIC=1 -fsanitize=address -mllvm -asan-globals-live-support -mllvm -asan-force-dynamic-shadow";
    "KASAN_CFLAGS_TBI" = "-DKASAN=1 -DKASAN_TBI=1 -fsanitize=kernel-hwaddress -mllvm -hwasan-recover=0 -mllvm -hwasan-instrument-atomics=0 -mllvm -hwasan-instrument-stack=1 -mllvm -hwasan-generate-tags-with-calls=1 -mllvm -hwasan-instrument-with-calls=1 -mllvm -hwasan-use-short-granules=0 -mllvm -hwasan-memory-access-callback-prefix=__asan_";
    "KASAN_DEFAULT_CFLAGS" = "$(KASAN_CFLAGS_CLASSIC)";
    "KASAN_DEFAULT_CFLAGS[arch=arm64]" = "$(KASAN_CFLAGS_TBI)";
    "KASAN_DEFAULT_CFLAGS[arch=arm64e]" = "$(KASAN_CFLAGS_TBI)";
    "MACOSX_DEPLOYMENT_TARGET" = "27.0";
    "PLATFORM_NAME" = macosx;
    "SWIFT_ENABLE_EXPLICIT_MODULES" = "$(SWIFT_USE_INTEGRATED_DRIVER)";
    "TAPI_USE_SRCROOT" = YES;
    "TAPI_VERIFY_MODE" = Pedantic;
    "TEST_FRAMEWORK_SEARCH_PATHS" = "$(inherited) $(PLATFORM_DIR)/Developer/Library/Frameworks";
    "TEST_LIBRARY_SEARCH_PATHS" = "$(inherited) $(PLATFORM_DIR)/Developer/usr/lib";
    "_SWIFT_EXPLICIT_MODULES_ALLOW_CXX_INTEROP" = "$(SWIFT_ENABLE_EXPLICIT_MODULES)";
}
```

addiing 

```
| grep '^ *"\{0,1\}MACOSX_DEPLOYMENT_TARGET"\{0,1\} *= *"\{0,1\}[^"]*"\{0,1\}; *$' | sed 's/^ *"\{0,1\}MACOSX_DEPLOYMENT_TARGET"\{0,1\} *= *"\{0,1\}\([^"]*\)"\{0,1\} *; *$/\1/'
```

gives the 27.0, but I don't manage to replace the sed from above with your line correctly. Also: when we do this don't we revert something that under some system once worked correctly ?

vadz:
I don't know what we're even trying to do here, so I'm not sure what to answer.

I thought that we wanted the earliest version in `DEPLOYMENT_TARGET_SUGGESTED_VALUES` (which would be "12.0" in the example above), and this would be given by deleting all lines up to and including this one and then taking the next line. But maybe this is not what we want? If so, what is the goal?

csomor:
> I don't know what we're even trying to do here, so I'm not sure what to answer.
> 
> I thought that we wanted the earliest version in `DEPLOYMENT_TARGET_SUGGESTED_VALUES` (which would be "12.0" in the example above), and this would be given by deleting all lines up to and including this one and then taking the next line. But maybe this is not what we want? If so, what is the goal?

That is what I want, but I don't manage to construct the proper line to use within the configure script from your hint

vadz:
Sorry, I made a typo in my reply (missing `/`). I've now pushed the modified configure, does this do what is needed?

csomor:
@vadz thanks, I fear there is now some whitespace after the `-version-min=` where it should not be:

```
checking SDK deployment version...         12.0
checking if C compiler (gcc -isysroot /Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk -mmacosx-version-min=        12.0) works with SDK/version options... configure: error: in `/Users/cs/
```

and you've also removed the 'old' part that deals with systems where the key is `buildSettings`, I think we should let that one in, as I assume its for older SDKs

vadz:
> @vadz thanks, I fear there is now some whitespace after the `-version-min=` where it should not be:

Ugh, sorry, fixed.
 
> and you've also removed the 'old' part that deals with systems where the key is `buildSettings`, I think we should let that one in, as I assume its for older SDKs

As I wrote in the commit message, I don't think any still supported Xcode needs this, it was used by very old (like Xcode 1.x) versions from what I could find.

csomor:
> As I wrote in the commit message, I don't think any still supported Xcode needs this, it was used by very old (like Xcode 1.x) versions from what I could find.

sorry, my bad, I only read the title, that part works perfectly now, thanks a lot

I've  now looked at the different SDKs down to the lowest 10.11 we support in 3.3

Starting from macOS 10.14 SDK there is a key in `DefaultProperties` named `DEPLOYMENT_TARGET_SUGGESTED_VALUES`, 

Below that eg in 10.11 there is no `buildSettings` anymore, you are right, but in `DefaultProperties` there already is the single `MACOSX_DEPLOYMENT_TARGET` key whose value corresponds to the SDK version. So this would be the former extraction command line, that could deal with that, but only if there is no `DEPLOYMENT_TARGET_SUGGESTED_VALUES`

vadz:
So you think we need a fallback on `MACOSX_DEPLOYMENT_TARGET` if `DEPLOYMENT_TARGET_SUGGESTED_VALUES` is not present, is that right?

csomor:
Yes, the "old" one should be taken in that case (happening in SDK macOS 10.11 - 10.13)

vadz:
Hopefully this should be complete then?

I've also golfed it a bit as we don't need to run multiple commands when a single one will do.

csomor:
> Hopefully this should be complete then?
> 
> I've also golfed it a bit as we don't need to run multiple commands when a single one will do.

something went wrong, 

`WARNING: Could not determine deployment target from SDKSettings.plist`

and I don't manage to see what the problem is, the variable `defaultProperties` is filled correctly and if I execute the line 

```
MACOSX_SDK_PLIST_VERSION_MIN=`echo $defaultProperties | sed -n -e '1,/DEPLOYMENT_TARGET_SUGGESTED_VALUES/d; s/^ *//; s/"//g; s/,//p; q'`
```

in the shell everything is calculated correctly, but in the configure script `MACOSX_SDK_PLIST_VERSION_MIN` remains empty ...

vadz:
Sorry, I stupidly forgot to quote :-( Should really work now.

csomor:
Thanks a lot, configure works perfectly, build is running now, I'll report

csomor:
ok, builds fine, sample runs

vadz:
Thanks for testing, I'll merge it then.

PR #26579, opened 2026-06-11
OSS-Fuzz: Add new fuzzer targets tar operation

arthurscchan:
This PR creates a new OSS-Fuzz fuzzer targeting tar operation, mirroring the zip fuzzer.


vadz:
Looks useful, thanks! Will merge soon.

PR #26581, opened 2026-06-12
MSW: Preserve wxToolBar custom colour upon dark mode switch

stevecor:
Avoid overwriting wxToolBar custom colours upon switching into or out of dark mode.


vadz:
Hmm, why did we do this in the first place if removing it doesn't create any problems?

Anyhow, will merge soon, thanks.

PR #26582, opened 2026-06-13
validate background colour index in wxGIFDecoder::LoadGIF

dxbjavid:
The background colour index in a GIF's logical screen descriptor is read straight from the file and used to index the global colour table, but only the first 3*global_ncolors bytes of the 768-byte palette buffer are read from the stream. A file that declares a small global table together with a larger background index therefore makes LoadGIF take the background colour from the uninitialised tail of the buffer, which then leaks out through GetBackgroundColour(). This only uses the index when it falls within the colour table and otherwise leaves the background unset, as already happens for a GIF that doesn't specify one.

### DIFF
--- a/src/common/gifdecod.cpp
+++ b/src/common/gifdecod.cpp
@@ -699,9 +699,18 @@ wxGIFErrorCode wxGIFDecoder::LoadGIF(wxInputStream& stream)
             return wxGIF_INVFORMAT;
         }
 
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, not sure if it's better to ignore this silently or return `wxGIF_INVFORMAT`, but let's do it like this for now.

dxbjavid:

PR #26583, opened 2026-06-13
Add PrintToPDF to webview (Edge and macOS)

Blake-Madden:
Adds parity to Qt's webview engine.


Blake-Madden:
From my tests, printing `wxwidgets.org` from the sample looks excellent on macOS, but rather "meh" with Edge on MSW and WebKit2 on Linux (layout looks clunky, large image is stretched out and clipped). Seems like Edge and WebKit2 printing just isn't all that great, I don't see any options to make it look nicer.

Also, printing as landscape with WebKit (GTK) is (sporadically) broken, but that's a WebKit issue (https://bugs.webkit.org/show_bug.cgi?id=286614).

src/msw/webview_edge.cpp
```diff
@@ -1472,6 +1472,98 @@ void wxWebViewEdge::Print(const wxPrintData& printData, int flags)
...
+        return false;
+
+    const wxString filePathCopy = filePath;
+    HRESULT hr = webView7->PrintToPdf(
+        filePath.wc_str(),
+        nullptr,
+        Callback<ICoreWebView2PrintToPdfCompletedHandler>(
+            [this, filePathCopy](HRESULT errorCode, BOOL isSuccessful) -> HRESULT
+            {
+                wxWebViewEvent event(wxEVT_WEBVIEW_PDF_SAVED, GetId(), filePathCopy, wxString());
```
vadz: It's better to allocate this event on heap and use `QueueEvent()`. Or, alternatively, use `CallAfter()` to generate this event directly from the main thread.

src/msw/webview_edge.cpp
```diff
@@ -1472,6 +1472,98 @@ void wxWebViewEdge::Print(const wxPrintData& printData, int flags)
...
+        double widthInches = paperSizeTenthsMM.x / 254.0;
+        double heightInches = paperSizeTenthsMM.y / 254.0;
+        printSettings->put_PageWidth(widthInches);
+        printSettings->put_PageHeight(heightInches);
+    }
+
+    printSettings->put_ShouldPrintHeaderAndFooter(FALSE);
+
+    const wxString filePathCopy = filePath;
+    hr = webView7->PrintToPdf(
```
vadz: Could we extract this code in a helper function taking (possibly null) print settings to avoid duplicating the lambda?

vadz (commented):
Thanks, looks good! Just 2 minor comments that it would be nice to address before merging.

vadz:
Oh, and clang CI failure looks to be a real problem, apparently we are going to have to provide the UUID of `ICoreWebView2_7` ourselves.

src/msw/webview_edge.cpp
Blake-Madden: Got it, using QueueEvent now.

src/msw/webview_edge.cpp
Blake-Madden: Done.

Blake-Madden:
> Oh, and clang CI failure looks to be a real problem, apparently we are going to have to provide the UUID of `ICoreWebView2_7` ourselves.

Done.

src/gtk/webview_webkit2.cpp
```diff
@@ -1525,6 +1527,99 @@ void wxWebViewWebKit::Print(const wxPrintData& printData, int WXUNUSED(flags))
...
+        return false;
+
+    WebKitPrintOperation* printop = webkit_print_operation_new(m_web_view);
+
+    wxGtkObject<GtkPrintSettings> settings(gtk_print_settings_new());
+    // Do NOT translate this; this is the name of the "printer" that GTK looks up
+    gtk_print_settings_set_printer(settings, "Print to File");
+    gtk_print_settings_set(settings, GTK_PRINT_SETTINGS_OUTPUT_FILE_FORMAT, "pdf");
+    gtk_print_settings_set(settings, GTK_PRINT_SETTINGS_OUTPUT_URI, uri);
+    g_free(uri);
```
vadz: Could/should use `wxGtkString uri` instead.

samples/webview/webview.cpp
```diff
@@ -1753,6 +1773,59 @@ void WebFrame::OnPrintWithSettings(wxCommandEvent& WXUNUSED(evt))
...
+{
+    wxArrayString paperChoices;
+    paperChoices.Add("Letter (Portrait)");
+    paperChoices.Add("Letter (Landscape)");
+    paperChoices.Add("A4 (Portrait)");
+    paperChoices.Add("A4 (Landscape)");
+    paperChoices.Add("Legal (Portrait)");
+    paperChoices.Add("Legal (Landscape)");
+
+    int sel = wxGetSingleChoiceIndex(
```
vadz: This is really just in case you'd like to improve it, but IMHO it would make for a nicer demo if the paper choice combobox was part of the file dialog itself (to which it can be added using `SetCustomizeHook()`).

src/gtk/webview_webkit2.cpp
```diff
@@ -1525,6 +1527,99 @@ void wxWebViewWebKit::Print(const wxPrintData& printData, int WXUNUSED(flags))
...
+
+static void wxDoHandlePDFResult(gpointer user_data, int success)
+{
+    wxWebViewGtkPDFData* data = static_cast<wxWebViewGtkPDFData*>(user_data);
+    wxWebViewEvent event(wxEVT_WEBVIEW_PDF_SAVED, data->webView->GetId(),
+                         data->filePath, wxString());
+    event.SetInt(success);
+    event.SetEventObject(data->webView);
+    data->webView->HandleWindowEvent(event);
+    g_object_unref(data->printop);
```
vadz: This should be arguably done in `wxWebViewGtkPDFData` dtor (which should be added).

src/gtk/webview_webkit2.cpp
```diff
@@ -1525,6 +1527,99 @@ void wxWebViewWebKit::Print(const wxPrintData& printData, int WXUNUSED(flags))
 }
 #endif // wxUSE_PRINTING_ARCHITECTURE
 
+struct wxWebViewGtkPDFData
```
vadz, line 1531: This struct should define a ctor to prevent it from being created with uninitialized fields.

src/gtk/webview_webkit2.cpp
```diff
@@ -1525,6 +1527,99 @@ void wxWebViewWebKit::Print(const wxPrintData& printData, int WXUNUSED(flags))
...
+        return false;
+
+    WebKitPrintOperation* printop = webkit_print_operation_new(m_web_view);
+
+    wxGtkObject<GtkPrintSettings> settings(gtk_print_settings_new());
+    // Do NOT translate this; this is the name of the "printer" that GTK looks up
+    gtk_print_settings_set_printer(settings, "Print to File");
+    gtk_print_settings_set(settings, GTK_PRINT_SETTINGS_OUTPUT_FILE_FORMAT, "pdf");
+    gtk_print_settings_set(settings, GTK_PRINT_SETTINGS_OUTPUT_URI, uri);
+    g_free(uri);
```
vadz: Same as above.

src/gtk/webview_webkit2.cpp
```diff
@@ -1525,6 +1527,99 @@ void wxWebViewWebKit::Print(const wxPrintData& printData, int WXUNUSED(flags))
...
+
+    wxApplyPrintData(printop, settings, printData);
+
+    webkit_print_operation_set_print_settings(printop, settings);
+
+    wxWebViewGtkPDFData* data = new wxWebViewGtkPDFData{this, filePath, printop};
+    g_signal_connect(printop, "finished", G_CALLBACK(wxgtk_webview_webkit_pdf_finished), data);
+    g_signal_connect(printop, "failed", G_CALLBACK(wxgtk_webview_webkit_pdf_failed), data);
+
+    webkit_print_operation_print(printop);
```
vadz: What's going to happen if `wxWebViewWebKit` is destroyed (e.g. because the window containing it is closed) before the operation is finished? The pointer will probably leak, which may be not the worst thing, but I'm not sure if the code wouldn't crash, which would be worse.

Can a print operation be cancelled? If so, we should ideally do it from the dtor if one is in progress.

src/msw/webview_edge.cpp
```diff
@@ -1472,6 +1472,80 @@ void wxWebViewEdge::Print(const wxPrintData& printData, int flags)
 }
 #endif // wxUSE_PRINTING_ARCHITECTURE
 
+bool wxWebViewEdge::DoCallPrintToPdf(const wxString& filePath, ICoreWebView2PrintSettings* printSettings)
+{
+    wxCOMPtr<ICoreWebView2_7> webView7;
+    if (FAILED(m_impl->m_webView->QueryInterface(IID_PPV_ARGS(&webView7))))
+        return false;
```
vadz, line 1481: Should we log some error here? It could be confusing to the user if the "Print to PDF" menu item simply did nothing, without any explanation.

vadz (commented):
Thanks, this looks good but GTK version could be tidied up a bit and maybe made safer.

src/msw/webview_edge.cpp
Blake-Madden: done

samples/webview/webview.cpp
Blake-Madden: done

src/gtk/webview_webkit2.cpp
Blake-Madden: done

src/gtk/webview_webkit2.cpp
Blake-Madden: done

src/gtk/webview_webkit2.cpp
Blake-Madden: done

src/gtk/webview_webkit2.cpp
Blake-Madden: done

src/gtk/webview_webkit2.cpp
Blake-Madden: On my OpenSuse at least, I don't have a change to cancel, it performs the PDF save asynchronously. But I've wrapped this in a `wxWeakRef` just in case.

Blake-Madden:
The Qt failure doesn't appear to be me:

```
tests\controls\gridtest.cpp(759): FAILED:
  CHECK( colsize.GetCount() == 1 )
with expansion:
  0 == 1
```

vadz:
Yes, it's due to

```
  Timed out waiting for mouse release to be processed
```

which is not related to the changes here.

I'll merge this soon, thanks for the updates!

PR #26584, opened 2026-06-13
MSW: Preserve wxListCtrl custom colours upon dark mode switch

stevecor:
Avoid overwriting custom colours upon switching into or out of dark mode.

The text colour is now the same thing as the foreground colour, as with wxGenericListCtrl.


vadz:
Nice, thanks!

PR #26586, opened 2026-06-13
Support pen and brush styles in wxSVGGraphicsContext

MaartenBent:
Enable saving as SVG in the drawing sample using graphics context, and enable saving the graphics page as SVG.

Add missing pen and brush styles that `wxSVGFileDC` does support but `wxSVGGraphicsContext` did not.
Prevent SVG patterns with the same ID from being added multiple times, which would not pass SVG validation.

Fix `wxOverlayDC` in the drawing sample when `wxGraphicsContext` is disabled.

@Blake-Madden If possible, please check I didn't break anything. Thanks!

### DIFF
--- a/include/wx/private/svg.h
+++ b/include/wx/private/svg.h
@@ -26,6 +26,7 @@
 #endif
 
 #include <memory>
+#include <set>
 
 class WXDLLIMPEXP_FWD_CORE wxBitmap;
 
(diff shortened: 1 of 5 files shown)


vadz (approved):
This looks good to me, FWIW, thanks!

PR #26587, opened 2026-06-13
Disable MinGW thread local workaround in wxTranslations

MaartenBent:
This has been fixed in MSYS2 MinGW.

Fixes #26195

### DIFF
--- a/src/common/translation.cpp
+++ b/src/common/translation.cpp
@@ -1472,8 +1472,14 @@ using UntranslatedStrings = std::unordered_set<wxString>;
     in global variables outside of this class and only relying on the dtor to
     be executed when any thread (not necessarily created by wxWidgets) exits to
     ensure that we always perform the required cleanup.
+
+    thread_local seems to work correctly in MSYS2 MinGW as of May 2026,
+    see https://www.msys2.org/news/#2026-05-11-native-thread-local-storage-tls-with-gcc-16
+    Around that same time the major version was incremented to 15, so use that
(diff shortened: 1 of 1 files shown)


oltolm:
Wouldn't it be better to check GCC version? It was introduced with GCC 16.1.

vadz:
I _think_ this was a MinGW-w64 problem and not a gcc one, so I believe the current PR is correct, but I'm not completely sure. As it seems highly unlikely that anybody would use MinGW-w64 < 15 with gcc > 16 or vice versa, this is probably theoretical, but the check for MinGW-w64 version looks more correct.

Does anybody see any actual problems with it?

MaartenBent:
The [news article](https://www.msys2.org/news/#2026-05-11-native-thread-local-storage-tls-with-gcc-16) mentions that this also affects Clang (in the GCC environments, so mingw32/mingw64/ucrt64, not clang64).

[GCC 16](https://gcc.gnu.org/gcc-16/changes.html#windows) added support for native TLS. This is a configuration option when building the compiler. And MSYS2 has to enable this when building the compilers in their environments (see https://github.com/msys2/MINGW-packages/issues/29272).

So it does not only depend on the GCC version, but also on the MinGW version. It needs to be a version where native TLS was enabled for the compiler. And the easiest is to check  `__MINGW64_VERSION_MAJOR`.

Doing configure-time compiler checks to see if they use native tls or emutls does not seems necessary to me (if at all possible) for a workaround that was only added recently.

PR #26589, opened 2026-06-15
docs: add minimal wxFrame construction example (beginner-friendly)

ErenyMic:
## What & Why

New contributors often struggle with wxFrame's constructor signature.
This PR adds a minimal, well-commented sample that clarifies each parameter.

## Changes

- Added `samples/frameparams/frameparams.cpp`
- Added `samples/frameparams/CMakeLists.txt`
- No existing source, header or build-system files modified

## Notes for Reviewers

- Pure sample — zero risk to ABI or runtime behaviour
- Follows existing wxWidgets coding style (based on `samples/minimal/`)
- Happy to adjust placement or style if needed

## Parameter reference

| Parameter | Recommended default | Purpose |
|-----------|-------------------|---------|
| `parent`  | `nullptr`         | Top-level window |
| `id`      | `wxID_ANY`        | Auto-assigned ID |
| `title`   | your string       | Title bar text |
| `pos`     | `wxDefaultPosition` | Platform default |
| `size`    | `wxSize(w,h)`     | Initial size |
| `style`   | `wxDEFAULT_FRAME_STYLE` | Standard frame |
| `name`    | `"MyFrame"`       | Internal name |

### DIFF
--- a/samples/minimal/frameparams/frameparams.cpp
+++ b/samples/minimal/frameparams/frameparams.cpp
@@ -1,0 +1,45 @@
+#include <wx/wx.h>
+
+// wxApp is the application object. It controls app startup and shutdown.
+class MyApp : public wxApp
+{
+public:
+    // OnInit() runs once when the app starts.
+    // We create and show the main window here.
+    bool OnInit() override
+    {
+        // Create a top-level window (wxFrame).
+        // Why these parameters?
+        // - nullptr: no parent -> this is a main window.
+        // - wxID_ANY: let wxWidgets choose an internal ID.
+        // - "Hello wxWidgets": text shown in the title bar.
+        // - wxDefaultPosition: let the OS choose initial position.
+        // - wxSize(640, 480): initial width and height.
+        // - wxDEFAULT_FRAME_STYLE: normal frame buttons/border.
+        // - "MyFrame": internal name (useful for framework internals/debugging).
+        auto* frame = new wxFrame(
+            nullptr,
+            wxID_ANY,
+            "Hello wxWidgets",
+            wxDefaultPosition,
+            wxSize(640, 480),
+            wxDEFAULT_FRAME_STYLE,
+            "MyFrame"
+        );
+
+        // Add a small status bar at the bottom of the window.
+        frame->CreateStatusBar();
+        // Show a short message in the status bar.
+        frame->SetStatusText("Ready.");
+        // Put the window in the center of the screen.
+        frame->Centre();
+        // Make the window visible. Without this, the user sees nothing.
+        frame->Show(true);
+
+        // Return true so the app continues running.
+        return true;
+    }
+};
+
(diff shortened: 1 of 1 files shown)


PBfordev:
> New contributors often struggle with wxFrame's constructor signature.

Just curious: Where have you encountered those struggling new contributors (and to what they contribute)? I have been active in the wxWidgets user community for 15+ years and I can't remember seeing anything like that.

ErenyMic:
Hi @PBfordev, thanks for your reply!

To answer your question: the "struggling beginners" I mentioned are actually me and my classmates. We are students at a technical high school, and we recently had to use wxWidgets for a programming test in our software development class (POS). We've only been learning wxWidgets for half a semester.

Prior to this, we mainly worked with the standard C++ and Python documentation. During our classes and the test preparation, almost everyone in my class (about 18 students) found the wxWidgets documentation quite difficult to navigate compared to what we were used to. In fact, many of us ended up writing our own simplified cheat sheets and custom documentation—very similar to the format I proposed—just to understand things like the wxFrame constructor.

My suggestion comes directly from this real-world classroom experience. Since future students at our school will also be learning wxWidgets, I thought that making the documentation a bit more approachable for absolute beginners (perhaps orienting it a bit closer to the style of Python/C++ docs) would lower the barrier to entry.

I totally understand that for someone with 15+ years of experience this isn't an issue at all! I just wanted to share the perspective of a class trying to learn the framework from scratch. Thanks for taking the time to look at my suggestion.

> > New contributors often struggle with wxFrame's constructor signature.
> 
> Just curious: Where have you encountered those struggling new contributors (and to what they contribute)? I have been active in the wxWidgets user community for 15+ years and I can't remember seeing anything like that.

PBfordev:
IMO, there is already more than enough samples. 

Looking at the wxFrame ctor
>wxFrame (wxWindow *parent, wxWindowID id, const wxString &title, const wxPoint &pos=wxDefaultPosition, const wxSize &size=wxDefaultSize, long style=wxDEFAULT_FRAME_STYLE)

are you really saying that you and your classmates cannot understand that
1. `wxWindow *parent` is the window parent of the frame,
2. `wxWindowID id` is the frame id,
3. `wxString &title` is the frame title,
4. similarly for `position`, `size`, and `style`.

I understand that learning how to use an extensive library could be challenging, in particular if the people also have little experience programming, and may be struggling just with the language constructs. However, I don't think your solution is really going to solve anything. 

BTW, adding a sample is a bit more complex than just adding a .cpp file, see https://github.com/wxWidgets/wxWidgets/blob/master/docs/contributing/how-to-add-new-sample.md

BTW, I am not a library maintainer, just a (loud) user.

vadz:
Sorry, but this is highly redundant with the existing minimal sample and it makes no sense to add another sample doing the same thing. If something can be improved in the minimal sample, please try to do it there.

PR #26590, opened 2026-06-15
#add samples/image/image_basic.cpp

Roman-Designer:
This pull request adds a new minimal sample called "simpleimage".

The existing image sample in wxWidgets is very comprehensive and demonstrates
many formats and advanced operations. While this is useful, it can be
overwhelming for beginners who only want to learn how to load and display a
single image.

The goal is to offer a beginner-friendly entry point before exploring the
full image sample.

### DIFF
--- a/samples/image/image_basic.cpp
+++ b/samples/image/image_basic.cpp
@@ -1,0 +1,70 @@
+
+// beginner-friendly image sample
+
+#include <wx/wx.h>
+#include <wx/dcbuffer.h>
+
+class ImagePanel : public wxPanel
+{
+public:
+    ImagePanel(wxWindow* parent, const wxString& file)
+        : wxPanel(parent)
+    {
+
+        if (!m_image.LoadFile(file))
+        {
+            wxLogError("Could not load image: %s", file);
+        }
+        else
+        {
+            m_bitmap = wxBitmap(m_image);
+        }
+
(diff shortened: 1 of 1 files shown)


lanurmi:
For a beginner-friendly sample, it has remarkably few instructional comments in it, i.e. none at all.

PBfordev:
> For a beginner-friendly sample, it has remarkably few instructional comments in it, i.e. none at all.

Sorry to be blunt, but unfortunately, there are more issues:
1. `wxImage::LoadFile()` already reports an error on failure, there is no need to report the error again by ourselves.
2. The image path is the current directory dependent which is a bad practice which should not be promoted.
3. There is no reason for the `wxImage` instance to be a class member, it should be a local variable in the constructor (or better yet, don't use `wxImage` at all and use just `wxBitmap::LoadFile()`).

vadz:
Thanks for your attempt to help, but unfortunately I don't think this is acceptable because it adds some completely undiscoverable and not even compilable/buildable snippet which will at best be ignored and, realistically, will rot away pretty soon.

I believe the right way to show some simple code is to include the most important parts of it in the documentation and refer to the named part of the image sample (e.g. "See `LoadUserImage()` function in the image sample for the fully working version.").

But we definitely don't want to add code exactly duplicating what is already there, this is not going to help anybody.

PR #26591, opened 2026-06-15
new beginner friendly documentation for statusbar

Gtschoj:
In the samples under the status bar, I have created a beginner-friendly version for creating a status bar, including inserting text, as the old version is very complicated and difficult to understand without prior knowledge. This makes it very difficult to get started.


vadz:
This looks very similar to #26590 and I'm closing it for the same reasons, please see my comment there.

PR #26592, opened 2026-06-15
webview: add a minimal local HTML RunScript sample

kiDra94:
This PR adds a small wxWebView sample that loads a local HTML file and calls JavaScript from C++ with RunScript().

The goal of this sample is to make wxWebView easier to try out quickly, especially for developers who are new to GUI development with C++ and wxWidgets. A very small, copy-paste-friendly example is often much easier to learn from than a larger demo project, so this sample is intentionally kept minimal and focused.

The idea was inspired by the way C++ reference material often includes small method-level examples that can be copied, compiled, and tested immediately. This sample follows the same approach: it is simple to build, easy to understand, and useful as a starting point for experimenting with wxWebView.

The PR also adds a short documentation reference in webview.h pointing to the sample.

### DIFF
--- a/interface/wx/webview.h
+++ b/interface/wx/webview.h
@@ -1512,7 +1512,7 @@ public:
 
     /**
         Runs the given JavaScript code.
-        
+
         See @ref wxWebViewScriptResult for information about the script result and
         errors.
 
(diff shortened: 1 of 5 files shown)


vadz (changes requested):
This may be a good idea but I'm afraid it's not acceptable in its current form, we really don't want to add extra samples with their own ad hoc build system. Please integrate this into the main webview sample, e.g. add an item to its menu allowing to open this page.

vadz:
Please see my comment in #26590 about the right way to do this (include snippet in the docs, point to the full version in the existing sample).

PR #26593, opened 2026-06-15
Clarify RequestUserAttention() behaviour under wxGTK

song2617:
While using RequestUserAttention() on a Linux/wxGTK setup, I found that the call had no visible effect and produced no error, which made it hard to tell whether the function was failing or my code was wrong. It turned out the desktop environment simply did not support the urgency hint. The current documentation states the function is implemented "for wxGTK with task bars supporting it", but does not make clear that it silently does nothing when the environment lacks this support. This small documentation change clarifies that behaviour to help users avoid the same confusion.

### DIFF
--- a/interface/wx/toplevel.h
+++ b/interface/wx/toplevel.h
@@ -388,6 +388,12 @@ public:
         the window icon in the taskbar, and for wxGTK with task bars
         supporting it.
 
+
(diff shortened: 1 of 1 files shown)


vadz:
I agree that the documentation could be more explicit, but I think that this should be merged with the previous paragraph instead of being just added after it.

PR #26594, opened 2026-06-15
Remove unused function wxDialogBase::OnSysColourChanged()

stevecor:
This function became unreferenced in 2158f4d7986.


vadz:
Thanks for timely cleanup (not even 20 years later!).

PR #26595, opened 2026-06-15
avoid uninitialised palette read in wxXPMDecoder::ReadData

dxbjavid:
wxXPMDecoder::ReadData allocates the palette r/g/b arrays from the colour count given in the XPM header but fills them by walking the colour map, so a malformed file that reuses the same key on more than one colour line collapses those lines into a single map entry and leaves the tail of the arrays uninitialised. wxPalette() then copies that uninitialised memory into the palette later returned by GetPalette(), and in debug builds the wxASSERT(i == colors_cnt) at the end fires. The fix sizes the arrays and the palette by the number of distinct colours actually in the map. I added a small test loading such an XPM, which trips the assertion before the change and loads with a one-colour palette after it.

### DIFF
--- a/src/common/xpmdecod.cpp
+++ b/src/common/xpmdecod.cpp
@@ -819,9 +819,15 @@ wxImage wxXPMDecoder::ReadData(const char* const* xpm_data)
         }
     }
 #if wxUSE_PALETTE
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for the fix and the explanation!

dxbjavid:

PR #26596, opened 2026-06-15
Preserve custom colours upon dark mode switch

stevecor:
For wxNotebook and wxStaticText, preserve custom colours upon dark mode switch.

Not sure why wxStaticText::MSWSetDarkOrLightMode() was ever added. The control seems to respond fine to dark mode switching without it.

### DIFF
--- a/include/wx/msw/stattext.h
+++ b/include/wx/msw/stattext.h
@@ -57,8 +57,6 @@ protected:
                                   WXWPARAM wParam,
                                   WXLPARAM lParam) override;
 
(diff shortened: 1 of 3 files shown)


vadz:
I am almost certain that it was needed at some point, but maybe we don't support the very old Windows 10 versions that required it, so let's merge this.

Thanks!

vadz:
BTW, I usually put the contents of your PR descriptions in the commit message but this can't be done when it's a merge, i.e. when there are multiple commits. Please consider putting your (helpful) explanations of the changes not just in the PR but also as commit message body too. TIA!

Issue #26588, opened 2026-06-14
Fix issue found by OSS-Fuzz in `wxTarInputStream`


vadz:
The attached file is supposed to make the code run out of memory but I didn't have time to test it yet.

@arthurscchan Please let me know if you'd like to look at this.

[min.tar.gz](https://github.com/user-attachments/files/28930062/min.tar.gz)

arthurscchan:
Yes. I could take a look of it.

PR #26597, opened 2026-06-15
OSS-Fuzz: Add new fuzzer targets fileconf

arthurscchan:
This PR creates a new OSS-Fuzz fuzzer targets fileconf widgets.


vadz:
Thanks, but this will have to wait until #26588 is fixed, I don't want to add more potential fuzzing errors before dealing with the existing one.

PR #26600, opened 2026-06-16
treat truncated reads as errors in wxDataInputStream

dxbjavid:
Reading from a truncated or corrupt stream returned values built from uninitialised memory. ReadString() read a length prefix, allocated a buffer of that size and then decoded the full declared length even when fewer bytes were actually read, so the unread (uninitialised) tail of the temporary buffer leaked into the returned string. The fixed size readers like Read32() have the same problem: on a short read they return whatever was already in the local variable on the stack.

Following the review, this now checks that all requested bytes could actually be read before using them. If not, the input stream is marked as being in error (so the caller can detect the failure via IsOk()) and an empty string or zero value is returned instead of silently wrong data. This is applied to all the Read() calls in wxDataInputStream, not just ReadString().

The tests feed a string whose length prefix exceeds the bytes present and a 32 bit value with only two bytes available, and check the result is empty/zero and that the stream is no longer Ok.

### DIFF
--- a/include/wx/datstrm.h
+++ b/include/wx/datstrm.h
@@ -112,6 +112,14 @@ public:
 protected:
     wxInputStream *m_input;
 
+private:
+    // Try to read exactly the given number of bytes into the provided buffer.
+    //
+    // Return true if all of them could be read or mark the input stream as
+    // being in error and return false if fewer bytes than requested are
+    // available, as this means the data is truncated and can't be used.
+    bool ReadBytes(void *buffer, size_t size);
+
     wxDECLARE_NO_COPY_CLASS(wxDataInputStream);
 };
 
--- a/src/common/datstrm.cpp
+++ b/src/common/datstrm.cpp
@@ -78,9 +78,21 @@ wxDataInputStream::wxDataInputStream(wxInputStream& s, const wxMBConv& conv)
 {
 }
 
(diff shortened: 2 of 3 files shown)


vadz:
Thanks for finding the problem, but I'm not sure if this is the right fix, the stream really _should_ have the requested bytes and I think the right thing to do is to return en empty string and mark the stream as having an error.

This should also be done for all the other calls to `Read()` in `wxDataInputStream`, i.e. reading an integer may also fail if the stream is truncated and it should be an error, not a silently wrong result.

Could you please change this PR to do it like this?

dxbjavid:
Makes sense, a silently wrong value is worse than just returning the truncated decode. I've reworked it: there's now a small ReadBytes() helper that reads the requested number of bytes and, if it gets fewer, marks the input stream with wxSTREAM_READ_ERROR and bails out. ReadString() returns an empty string in that case and the fixed size readers (Read8/16/32/64 and the array overloads) return zero, so a truncated stream is reported through IsOk() rather than producing data built from uninitialised memory.

The test now checks both a truncated string and a truncated Read32(): the result is empty/zero and the stream is no longer Ok afterwards.

vadz:
Very nice, thanks!

dxbjavid:
Thanks for the review and merge! Appreciated

PR #26601, opened 2026-06-16
MSW: Dark mode switching for wxTreeCtrl

stevecor:
Update wxTreeCtrl to respond to system colour changes, including switching into and out of dark mode.


vadz:
Looks good, thanks!

PR #26602, opened 2026-06-16
Replace wxTopLevelWindow::GeometrySerializer with GeometryStore

vadz:
The old class used incorrect const qualifiers on its member functions: saving function typically modifies the object being saved into, and so must not be const, while loading function should be const, but the old class reversed this.

Do it right this time in the new GeometryStore class and keep the old class for compatibility only.


vadz:
Merged in the commit above.

PR #26603, opened 2026-06-16
MSW: Dark mode switching for wxWindow

stevecor:
Default window background erasing responds to dark mode switching for wxWindow and derived windows such as wxPanel. This is done by adapting the default window erasing for dark mode in the background erase handling rather than when the window class is registered.

### DIFF
--- a/include/wx/msw/mdi.h
+++ b/include/wx/msw/mdi.h
@@ -226,6 +226,8 @@ public:
 
     void OnIdle(wxIdleEvent& event);
 
+    virtual wxVisualAttributes GetDefaultAttributes() const override;
+
 protected:
     virtual void DoGetScreenPosition(int *x, int *y) const override;
     virtual void DoGetPosition(int *x, int *y) const override;
--- a/src/msw/app.cpp
+++ b/src/msw/app.cpp
@@ -564,15 +564,6 @@ const wxChar *wxApp::GetRegisteredClassName(const wxChar *name,
             return gs_regClassesInfo[n].GetRequestedName(flags);
     }
 
-    // In dark mode, use the dark background brush instead of specified colour
-    // which would result in light background.
-    HBRUSH hbrBackground;
-    if ( wxMSWDarkMode::IsActive() )
-        hbrBackground = wxMSWDarkMode::GetBackgroundBrush();
-    else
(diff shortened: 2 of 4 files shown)


vadz (commented):
It's not clear to me if the other code using `GetRegisteredClassName()` doesn't need to be updated, see `wxDataViewMainWindow`, `wxMDIChildFrame` and `wxMDIFrame`, (there are also `wxGLCanvas` and `Scintilla`, but their background is probably always painted over anyhow).

Could we use `SetClassLongPtr(GCL_HBRBACKGROUND)` when switching between modes instead?

stevecor:
None of the calls to `GetRegisteredClassName()` need to be modified. This change only affects calls that specify a background colour. Most of the calls you mentioned specify the colour as -1, so there is no effect with those.

In `wxMDIChildFrame` the background colour is specified as `wxSYS_COLOUR_WINDOW`. I find that this change does in fact fix the dark mode switching for MDI client windows for the background erasing of the client area, as intended.

Yes we could instead use `SetClassLongPtr(GCL_HBRBACKGROUND)` when switching modes. ~~A disadvantage to this approach is the colour index information would be lost. We would have to use a fixed index such as `wxSYS_COLOUR_WINDOW` or `wxSYS_COLOUR_BTNFACE`, which would ignore whatever was originally specified. That might be a bigger issue switching to light mode than to dark mode. Maybe we could store that information somewhere, but that's starting to get more complicated. The proposed changes preserve the colour index.~~ We can use GetDefaultAttributes() to get the background colour.

vadz:
> None of the calls to `GetRegisteredClassName()` need to be modified. This change only affects calls that specify a background colour. Most of the calls you mentioned specify the colour as -1, so there is no effect with those.

Good point, sorry for missing this.
 
> Yes we could instead use `SetClassLongPtr(GCL_HBRBACKGROUND)` when switching modes.

I have a slight preference for this because this should minimize differences in behaviour between light and dark mode, which is an advantage in my eyes.

A potential disadvantage is that it might actually be a bit more complicated, as we'd ideally want to do it only once after the theme switch and I'm not sure if we have a way to do this. OTOH having a virtual function/event called/sent only once per theme change in/to `wxApp` could be a useful thing anyhow, and if we could implement this, we could use it to change the brush for all the currently registered classes.

vadz:

Issue #26557, opened 2026-06-03
AUI pane becomes huge on DPI change if a partial minimum size is specified


taler21:
For example, if an AUI pane has a minimum width but no minimum height specified, and it has a scroll bar because its content cannot be fully displayed at its current size, the pane becomes huge when the DPI changed.
It seems as if the size of the pane grows to such an extent that its entire content would be visible, even if the size extends far beyond the screen.

I would expect the current size to be retained, merely scaled to the changed DPI.

The issue can also be reproduced in the aui sample with the following changes.
```patch
diff -r -u -p a/samples/aui/auidemo.cpp b/samples/aui/auidemo.cpp
--- a/samples/aui/auidemo.cpp	2026-06-03 08:31:53.061465300 +0200
+++ b/samples/aui/auidemo.cpp	2026-06-03 08:32:40.014698200 +0200
@@ -1044,7 +1044,7 @@ MyFrame::MyFrame(wxWindow* parent,
 
     m_mgr.AddPane(CreateTreeCtrl(), wxAuiPaneInfo().
                   Name("tree").Caption("Tree Pane").
-                  Left().Layer(1).Position(1).
+                  Left().Layer(1).Position(1).MinSize(FromDIP(wxSize(200,-1))).
                   CloseButton(true).MaximizeButton(true).MinimizeButton().
                   IconMin(wxArtProvider::GetBitmapBundle(wxART_CUT, wxART_MENU)));
 
@@ -2573,6 +2573,9 @@ wxTreeCtrl* MyFrame::CreateTreeCtrl()
         tree->AppendItem(id, "Subitem 3", 1);
         tree->AppendItem(id, "Subitem 4", 1);
         tree->AppendItem(id, "Subitem 5", 1);
+        for (int j = 6; j <= 20; ++j)
+            tree->AppendItem(id, wxString::Format("Subitem %d", j), 1);
+        tree->Expand(id);
     }
 
 

```

### Platform and version information

- wxWidgets version  you use: master (39c73b42f5882fc0eef568eda077ee20385c7bbd)
- wxWidgets port  you use: wxMSW
- OS  and its version: Windows 11

vadz:
I've finally looked at this and it's indeed due to `wxTreeCtrl` returning a huge best size. So the questions we need to answer are:

1. Is it correct to return such size from its `DoGetBestSize()` at all or should it be limited to something reasonable?
[...]

PR #26604, opened 2026-06-16
Fixes for behaviour after DPI change in wxMSW

vadz:
See #26557.

### DIFF
--- a/include/wx/sizer.h
+++ b/include/wx/sizer.h
@@ -748,6 +748,15 @@ public:
     virtual bool InformFirstDirection( int WXUNUSED(direction), int WXUNUSED(size), int WXUNUSED(availableOtherDir) )
         { return false; }
 
+    // Update stored dimensions expressed in logical pixels on DPI change.
+    // This is only needed on the platforms where logical pixels differ from
+    // the physical ones, e.g MSW.
+    //
+    // This is an internal function, only called by wxWidgets itself.
+#ifndef wxHAS_DPI_INDEPENDENT_PIXELS
+    virtual void UpdateOnDPIChange(wxSize oldDPI, wxSize newDPI);
+#endif // !wxHAS_DPI_INDEPENDENT_PIXELS
+
 protected:
     wxSize              m_size;
     wxSize              m_minSize;
--- a/include/wx/sizer.h
+++ b/include/wx/sizer.h
@@ -831,6 +831,10 @@ public:
     // (for internal use only)
     int CalcRowsCols(int& rows, int& cols) const;
 
+#ifndef wxHAS_DPI_INDEPENDENT_PIXELS
+    virtual void UpdateOnDPIChange(wxSize oldDPI, wxSize newDPI) override;
+#endif // !wxHAS_DPI_INDEPENDENT_PIXELS
+
 protected:
     // the number of rows/columns in the sizer, if 0 then it is determined
     // dynamically depending on the total number of items
--- a/src/aui/framemanager.cpp
+++ b/src/aui/framemanager.cpp
@@ -2504,8 +2504,7 @@ void wxAuiManager::LayoutAddPane(wxSizer* cont,
     // We need to reset any previous set min size to allow decreasing the pane
     // size by dragging the sash between it and other panes, so always set it
     // to something, even if it's not specified.
-    if (min_size == wxDefaultSize)
-        min_size = wxSize(1, 1);
+    min_size.IncTo(wxSize(1, 1));
 
     sizer_item->SetMinSize(min_size);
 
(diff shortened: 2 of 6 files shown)


vadz:
I'm going to merge this soon if there are no objections/comments from @taler21 or anybody else because this does improve things, even if it doesn't solve all the problems.

taler21:
This PR significantly improves things. Many thanks!

vadz:
Thanks for testing!

taler21:
Unfortunately, commit 8386a98409d0964d4aa62cba9a69c4991b8c3235 (`Restrict the new size after DPI change to the display size, 2026-06-16`) introduced a regression for non-resizable dialogs.
After changing the DPI, the dialog buttons are no longer visible because the dialog becomes too small.

This can also be seen in the dialogs sample, for instance with the 'Credentials dialog', which can be opend by pressing `Ctrl+Shift+C`.
After changing the DPI scaling from 200% to 100%, the dialog becomes too small.
CredentialEntryDialog-after-DPI-change

vadz:
Oops, thanks for noticing and sorry for a stupid mistake: I don't know why, but I've switched from window to client size in this code and we shouldn't do this, of course.

This should fix it and I'll push it soon:

```diff
diff --git a/src/msw/nonownedwnd.cpp b/src/msw/nonownedwnd.cpp
index c08e6f8358..635858876e 100644
--- a/src/msw/nonownedwnd.cpp
+++ b/src/msw/nonownedwnd.cpp
@@ -319,7 +319,7 @@ bool wxNonOwnedWindow::HandleDPIChange(const wxSize& newDPI, const wxRect& newRe
         wxRect actualNewRect = newRect;
         if ( wxSizer* sizer = GetSizer() )
         {
-            const wxSize minSize = sizer->ComputeFittingClientSize(this);
+            const wxSize minSize = sizer->ComputeFittingWindowSize(this);
             wxSize diff = minSize - newRect.GetSize();
 
             // We don't want to shrink the window as if the user had increased
```

taler21:
Commit 75284bb8f05ab3fe8a87ea560788af4a31ec2068 (`Ensure the window origin is visible after DPI change, 2026-06-16`) has an unwanted side effect.

When moving a window from one monitor to the right to another monitor with a different DPI scaling, it causes an offset between window position and mouse position.
This is because, at the time of the DPI change, the origin of the window is still on the previous monitor. And the added code therefore moves the window (unwantedly in this case) to the next monitor.

taler21:
I have also noticed a strange behavior under Windows 11 (Version 25H2, Build 26200.8655), which is not reproducible under Windows 10.
Depending on the mouse position when dragging the window, i.e. whether the mouse is in the left or right half of the window, `HandleDPIChange()` is either called as expected, i.e. as soon as the majority (center) of the window is on the target screen, or even before.

For this reason, `wxDisplay(this).GetClientArea()` added in 75284bb8f05ab3fe8a87ea560788af4a31ec2068 (`Ensure the window origin is visible after DPI change, 2026-06-16`) sometimes detects the wrong/previous monitor under Windows 11, instead of the expected/targeted one.

This can be reproduced and visualized using the following code and procedure.
```patch
diff -r -u -p a/src/msw/nonownedwnd.cpp b/src/msw/nonownedwnd.cpp
--- a/src/msw/nonownedwnd.cpp	2026-06-30 08:57:47.074200300 +0200
+++ b/src/msw/nonownedwnd.cpp	2026-06-30 08:59:05.556618900 +0200
@@ -342,6 +342,7 @@ bool wxNonOwnedWindow::HandleDPIChange(c
 
             // However still ensure that the window origin is visible on its
             // display, we don't move to move it out of visible space.
+            wxLogDebug( "DPI scale factor = %3.02f", wxDisplay(this).GetScaleFactor() );
             const wxRect screenRect = wxDisplay(this).GetClientArea();
             if ( actualNewRect.x < screenRect.x )
                 actualNewRect.x = screenRect.x;

```

Scenarios with unexpected behavior:
1. Move a window from a monitor with 200% DPI scaling to the right to a monitor with 100% DPI scaling, with the mouse pointer on the right half of the window.
2. Move a window from a monitor with 100% DPI scaling to the left to a monitor with 200% DPI scaling, with the mouse pointer on the right half of the window.

In both cases, `HandleDPIChange()` is apparently already being called even though the center of the window (`this`) is still on the previous monitor.
As a result, the previous monitor will be detected by `wxDisplay(this)`.

If the mouse is pointing to the other half of the window while dragging it, it works as expected in both scenarios.

I don't know if this is a bug in Windows 11 or a Windows 11-specific issue in wxWidgets. Therefore, I wanted to report what I have noticed.

vadz:
Thanks for reporting this!

Unfortunately I don't know how to determine the new monitor if we can't rely on the window center being (already) on it.

We could revert that commit, I think, it was mostly meant to be a workaround for the case when the window became too big to fit on screen. Alternatively, perhaps we could check whether the window is visible on at least some monitor, not necessarily the current one? But this risks being difficult to get right and we'd still need to decide to which monitor to move it if the check fails.

Issue #26605, opened 2026-06-16
[MSW] [DARK MODE] Current wxNotebook in master is broken


eranif:
Using the latest git master, `wxNotebook` are using white background even when app is forced to dark mode.
My current setup is: Windows 11 Light background, however, CodeLite (built with latest git master) is setup for "Force Dark".

BEFORE the recent pull from git, it looked like this:

<img width="2150" height="1491" alt="Image" src="https://github.com/user-attachments/assets/5a6943b6-aa46-4185-a086-87d3d8ac5ce5" />

With recent pull it looks like this:

<img width="2178" height="1566" alt="Image" src="https://github.com/user-attachments/assets/d0ecfb39-a3df-4acc-adf4-207d50db523a" />

You can easily reproduce this with the `notebook.exe` sample:

Just add `SetAppearance(Appearance::Dark)` in the App class and you will see it.

Windows 11, Clang 21/ MSYS2

Thanks

eranif:
FYI: I am running `git bisect` to locate the offending commit, will report it here once I find it.

eranif:
According to `git bisect` this is the commit: https://github.com/wxWidgets/wxWidgets/commit/f8363719d46ffd10b200e5efeedffb83aae9671f

Confirmed: reverting this commit fixes the issue.

vadz:
Thanks, I can indeed see it in the sample.

@stevecor Are you already looking at this? I've seen your (now deleted?) comment in #26603, but it indeed doesn't fix the issue. I think it's not setting the background in the commit linked above which broke this, there may be better ways to fix it than restoring this call, but I'm not sure.

stevecor:
Yes, I'm investigating.

Issue #26599, opened 2026-06-16
MSW dark mode: add dark mode support for legacy common dialogs (wxColourDialog, wxFontDialog, wxPageSetupDialog, wxPrintDialog, wxFindReplaceDialog)


memoarfaa:
### Summary

The `MSWEnableDarkMode()` documentation explicitly lists five dialogs as unsupported for dark mode:

- `wxColourDialog`
- `wxFindReplaceDialog`
- `wxFontDialog`
- `wxPageSetupDialog`
- `wxPrintDialog`

These all wrap legacy Win32 Common Dialog procedures (`ChooseColor`, `FindText`/`ReplaceText`, `ChooseFont`, `PageSetupDlg`, `PrintDlg`/`PrintDlgEx`). Windows itself does not dark-theme these dialogs, so they render with a light background even when the entire application is in dark mode — creating a jarring visual break for end users.

### Proposed solution

I have implemented two new internal functions in the `wxMSWDarkMode` namespace that bracket the lifetime of dark-mode theming for any legacy common dialog window:

```cpp
namespace wxMSWDarkMode
{
    void AllowForSystemDialog(HWND hwnd);
    void RemoveFromSystemDialog(HWND hwnd);
}
```

`AllowForSystemDialog()` is called from `WM_INITDIALOG` inside each dialog's existing hook procedure (e.g. `lpfnHook` for `CHOOSECOLOR`, `lpfnHook` for `FINDREPLACE`, etc.). It applies dark-mode theming to the dialog frame and recursively to all child controls via `EnumChildWindows`.

`RemoveFromSystemDialog()` is called from `WM_NCDESTROY` and cleans up all subclasses, GDI resources, and theme state installed by the first call. Both functions are no-ops when `IsActive()` returns false, so no guard is needed at call sites.

**Hook registration mapping:**

| wxWidgets class | Win32 structure | Hook field | Flag |
|---|---|---|---|
| `wxColourDialog` | `CHOOSECOLOR` | `lpfnHook` | `CC_ENABLEHOOK` |
| `wxFindReplaceDialog` | `FINDREPLACE` | `lpfnHook` | `FR_ENABLEHOOK` |
| `wxFontDialog` | `CHOOSEFONT` | `lpfnHook` | `CF_ENABLEHOOK` |
| `wxPageSetupDialog` | `PAGESETUPDLG` | `lpfnPageSetupHook` | `PSD_ENABLEPAGESETUPHOOK` |
[...]

PR #26606, opened 2026-06-16
MSW: Dark mode wxNotebook background erasing

stevecor:
Use dark mode themed background brush for proper background erasing and update background brush upon dark mode switching. See #26605.


vadz:
This doesn't solve the problem under Windows 10, probably because `DarkMode_DarkTheme::Tab` theme doesn't exist there.

P.S. According to #26599 this theme is only available since Windows 11 26100.6772.

vadz:
Thanks, this works for me so I'll merge this.

PR #26607, opened 2026-06-17
Bug: OSS-Fuzz: Deny an unreasonably large header to avoid OOM

arthurscchan:
This PR adds a guard to deny an unreasonably large header to avoid possible OOM in `wxTarInputStream` processing, discovered by OSS-Fuzz fuzzer (#26588).

### DIFF
--- a/src/common/tarstrm.cpp
+++ b/src/common/tarstrm.cpp
@@ -622,6 +622,9 @@ void wxTarEntry::SetMode(int mode)
 /////////////////////////////////////////////////////////////////////////////
 // Input stream
 
+// Default cap of 10 MiB; can be changed with SetMaxExtendedHeaderSize().
+size_t wxTarInputStream::sm_maxExtendedHeaderSize = 10 * 1024 * 1024;
+
 wxTarInputStream::wxTarInputStream(wxInputStream& stream,
                                    wxMBConv& conv /*=wxConvLocal*/)
  :  wxArchiveInputStream(stream, conv)
@@ -909,16 +912,16 @@ bool wxTarInputStream::ReadExtendedHeader(wxTarHeaderRecords*& recs)
     if (!recs)
         recs = new wxTarHeaderRecords;
 
-    // round length up to a whole number of blocks
     size_t len = m_hdr->GetOctal(TAR_SIZE);
 
-    // An extended header just holds metadata records and is meant to be small.
-    static const size_t MAX_EXTENDED_HEADER = 1024 * 1024;
-    if (len > MAX_EXTENDED_HEADER) {
-        wxLogWarning(_("invalid data in extended tar header"));
+    // reject an unreasonably large extended header to avoid excessive allocation
+    if (len > GetMaxExtendedHeaderSize())
+    {
+        wxLogError(_("Extended tar header size %zu is greater than maximum allowed."), len);
         return false;
     }
 
+    // round length up to a whole number of blocks
     size_t size = RoundUpSize(len);
 
     // read in the whole header since it should be small
--- a/src/common/tarstrm.cpp
+++ b/src/common/tarstrm.cpp
@@ -911,6 +911,14 @@ bool wxTarInputStream::ReadExtendedHeader(wxTarHeaderRecords*& recs)
 
     // round length up to a whole number of blocks
     size_t len = m_hdr->GetOctal(TAR_SIZE);
+
+    // An extended header just holds metadata records and is meant to be small.
+    static const size_t MAX_EXTENDED_HEADER = 1024 * 1024;
+    if (len > MAX_EXTENDED_HEADER) {
+        wxLogWarning(_("invalid data in extended tar header"));
+        return false;
+    }
+
     size_t size = RoundUpSize(len);
 
     // read in the whole header since it should be small
--- a/include/wx/tarstrm.h
+++ b/include/wx/tarstrm.h
@@ -166,6 +166,10 @@ public:
     wxFileOffset GetLength() const override      { return m_size; }
     bool IsSeekable() const override { return m_parent_i_stream->IsSeekable(); }
 
+    // Limit on the extended header size, to avoid excessive allocation.
+    static size_t GetMaxExtendedHeaderSize() { return sm_maxExtendedHeaderSize; }
+    static void SetMaxExtendedHeaderSize(size_t size) { sm_maxExtendedHeaderSize = size; }
+
 protected:
     size_t OnSysRead(void *buffer, size_t size) override;
     wxFileOffset OnSysTell() const override      { return m_pos; }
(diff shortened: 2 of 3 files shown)


vadz:
Thanks, I don't know enough about (GNU?) TAR format to be sure if this limitation can be a problem in practice. Does anybody have any example tar files using extended header? I've tried a few of tarballs I had lying around but none of them uses it at all, so I couldn't see what could its reasonable size be like.

Maybe bump it up to 10MiB? This should be enough and still shouldn't be a problem to allocate on any modern system. We could also add a static `wxTarInputStream` function to set this.

arthurscchan:
I don't think we care about legitimate tar files. But since the widget itself is meant to process tar stream and thus it be could possible to have an adversary just pass in a random stream claimed to be tar file (not necessary to be a real tar) but with extremely large extended header and cause the OOM.

vadz:
Err, we do care about legitimate tar files. In case I wasn't clear, my concern was that we could now start rejecting valid (legitimate) tar files which have extended header bigger than 1MiB. I don't know how likely is this to happen but it doesn't seem completely impossible, something like a Boost tarball can contain thousands of files with very long names.

arthurscchan:
er... true. I am now double checking about the standard for tar ball extended headers. Sorry for the rush conclusion.

imciner2:
I looked in the POSIX 1.2001 standard, and there is nothing stating how long an extended header can be - it is basically left as a free field in terms of length.

vadz:
Thanks Ian, this confirms my suspicion that there is no universally acceptable limit here. But I still think adding `wxTarInputStream::SetMaxExtendedHeaderSize()` and setting it to 10MiB by default would be a reasonable thing to do.

arthurscchan:
Thanks for confirming. Do you want me help do the changes in this PR following the 10MB assumption you mentioned?

wh11204:
I just want to remark that the last modification time field in the given header (bytes 136 to 147, Unix timestamp in octal) contains a '@', obviously not a valid octal digit. I do not know if the code uses this field at all, but it could possibly crash.

vadz:
> Thanks for confirming. Do you want me help do the changes in this PR following the 10MB assumption you mentioned?

I am still not sure what to do about this, but the best idea I have remains to

1. Restrict this to 10MiB (or any other arbitrary but pretty high number).
2. Add a static function allowing to change this limit if the application ever really needs it.

If you could please do it, it would be great, of course. And if you can think of something better to do, it would be even better :-)

vadz:
> I just want to remark that the last modification time field in the given header (bytes 136 to 147, Unix timestamp in octal) contains a '@', obviously not a valid octal digit. I do not know if the code uses this field at all, but it could possibly crash.

Sorry, I hardly know this code at all but if you think it's a bug, please open a separate issue for it, I don't think it's related to anything being discussed here (other than in the sense that the fuzzer should be able to find this bug as well).

src/common/tarstrm.cpp
```diff
@@ -909,8 +912,26 @@ bool wxTarInputStream::ReadExtendedHeader(wxTarHeaderRecords*& recs)
...
+    // read the declared size as 64-bit before narrowing, to avoid truncation
+    const wxTarNumber declared = m_hdr->GetOctal(TAR_SIZE);
+
+    // reject an unreasonably large extended header to avoid excessive allocation
+    if (declared < 0 || declared > (wxTarNumber)GetMaxExtendedHeaderSize()) {
+        wxLogWarning(_("invalid data in extended tar header"));
+        return false;
+    }
+
+    size_t len = (size_t)declared;
```
vadz: If we did this above, we could compare it with `GetMax...Size()` without a cast (`< 0` would become unnecessary).

src/common/tarstrm.cpp
```diff
@@ -909,8 +912,26 @@ bool wxTarInputStream::ReadExtendedHeader(wxTarHeaderRecords*& recs)
     if (!recs)
         recs = new wxTarHeaderRecords;
 
+    // read the declared size as 64-bit before narrowing, to avoid truncation
+    const wxTarNumber declared = m_hdr->GetOctal(TAR_SIZE);
+
+    // reject an unreasonably large extended header to avoid excessive allocation
+    if (declared < 0 || declared > (wxTarNumber)GetMaxExtendedHeaderSize()) {
```
vadz: Minor, but in wx coding style braces should be on their own line:

```suggestion
    if (declared < 0 || declared > (wxTarNumber)GetMaxExtendedHeaderSize())
    {
```

arthurscchan:
I have updated the logic, now the logic rejects extended headers larger than a configurable cap (default 10 MiB, settable via wxTarInputStream::SetMaxExtendedHeaderSize()) instead of allocating the declared size up front. I also adds a check of size capped to 64-bit value before narrowing to size_t, and the allocation is clamped
to the bytes actually available on streams.

src/common/tarstrm.cpp
```diff
@@ -909,8 +912,26 @@ bool wxTarInputStream::ReadExtendedHeader(wxTarHeaderRecords*& recs)
...
+        wxLogWarning(_("invalid data in extended tar header"));
```
vadz: This is an error and not a warning as the stream can't be used and I think it would be worth giving more details about the problem, i.e.:

```suggestion
        wxLogError(_("Extended tar header size %zu is greater than maximum allowed."), len);
```

src/common/tarstrm.cpp
```diff
@@ -909,8 +912,26 @@ bool wxTarInputStream::ReadExtendedHeader(wxTarHeaderRecords*& recs)
...
+        return false;
+    }
+
+    size_t len = (size_t)declared;
+
+    // don't allocate more than the stream can actually provide
```
vadz: Sorry, why do we need this? Allocating up to 10MiB shouldn't be a huge problem and we hopefully get an error if we read less data than expected (if we don't, we should fix this).

And if we really want to check for this, we should return an error if there is not enough data in the file as this means that it is corrupted, not allocate less.

vadz (changes requested):
Thanks, this looks good but we need to document the new functions in `interface/wx/tarstrm.h`, could you please do this (please use `@since 3.3.3` in their documentation to indicate that they're new)? TIA!

src/common/tarstrm.cpp
arthurscchan: Changed.

src/common/tarstrm.cpp
arthurscchan: Changed.

src/common/tarstrm.cpp
arthurscchan: True, it is more like defensive in my original though. But I do agree it is not needed. Removed.

src/common/tarstrm.cpp
arthurscchan: Fixed.

vadz:
Sorry for the delay! This looks good now and I'll merge it in a moment, thanks!

PR #26608, opened 2026-06-17
MSW: Fix dark mode switching for wxStatusBar

stevecor:
When switching to light mode, always use the theme name L"Explorer" rather than using MSWGetDarkModeSupport(). The theme name and id needed for dark mode does not always work for light mode. For example, the wxStatusBar combination name=nullptr id=L"ExplorerStatusBar" results in an un-themed status bar with strange internal borders and missing gripper.


vadz:
Thanks, will merge too soon.

Issue #26482, opened 2026-05-19
Search seems to be broken in the online manual


fishnet37222:
This issue is being opened from this [thread](https://forums.wxwidgets.org/viewtopic.php?p=231398#p231398).

I've noticed some issues when trying to use the search functionality of the online manual.

When I select one of the 3.2 versions, I get this:
<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/1e8ecccd-e595-459d-b30b-3dfdfe423132" />

When I select one of the 3.3 versions, I get this instead:
<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/d64021df-3671-4da2-9f42-c94d16288020" />

vadz:
I can confirm that [3.3.2 search](https://docs.wxwidgets.org/3.3.2/search.php?query=lzma) is broken for me too. @tierra Would you have any idea by chance? I don't understand how could Doxygen update affect old documentation, generated before we did it... Do they cache different versions of the same CSS or something like this?

But [latest search](https://docs.wxwidgets.org/latest/search.php?query=lzma) is even worse, it simply doesn't show anything at all because I get status 500 in response to this query.

tierra:
For the layout issue: The CSS hasn't changed at all since it was uploaded. I believe it has always had this weird layout bug. It depends heavily on how wide your browser window is (responsive breakpoints), and also how many search results (and thus how tall the results are) your search has. I just think 4k monitors weren't much of a thing back when this combo of Doxygen + Doxygen Awesome was released. If someone has a quick and simple CSS fix to improve it, I'm open to uploading that to fix it up on old versions. However, this still appears to be a problem even with Doxygen 1.15.0 and latest version of DA on trunk docs (when search works anyway).

The 500 on latest: I tried replacing that `search_functions.php` script with the latest, and it has fixed all the older versions of the manual. But I can't figure out why even the latest Doxygen 1.15.0 is uploading the older incompatible version. It was fixed in 1.9.7. I did replace that file on the server, but new actions builds always just rsync over the older broken version regardless. I don't even know how those builds have the old version, as Doxygen 1.15.0 release is the only version ever installed in those containers.

vadz:
This has been fixed by removing our old custom search function, although I can't find commit at the moment, thanks to everybody involved.

PR #26610, opened 2026-06-18
Fix PHP 8 incompatibilities in Doxygen search backend

anhnong90:
Hi! I noticed that the search on the wxWidgets documentation site currently returns a 500 error. After checking the code and testing locally with PHP 8.3, I found several small incompatibilities that may cause the backend search to fail. This PR updates `custom_search_functions.php`, so it works cleanly on PHP 8.3.

## Summary of changes
* **Replaced deprecated curly-brace string offsets:** Changed `$word{0}` to `$word[0]`.
* **Replaced removed `each()`:** Swapped with `foreach`.
* **Added safety checks around `fgetc()`:** Prevents `ord(false)` errors during file reads.
* **Added guards around `substr()`:** Prevents errors when `$w` may be empty.
* **Improved robustness of `readString()` and `readInt()`:** Handles unexpected EOF (End-of-File) gracefully.
* **Updated `readHeader()`:** Safely handles truncated or invalid headers.

## Reason for the update
The documentation server appears to run PHP 8.x, where several constructs used in the original Doxygen-generated script are no longer supported. These updates preserve the original behavior while ensuring compatibility with modern PHP versions.

## Testing
* Verified functionality locally using PHP 8.3.
* Confirmed backend search results load correctly.
* No warnings or fatal errors.
* Autocomplete (JS-based) remains unchanged.

**Fixes:** #26482


vadz:
Thanks for looking at this! I know nothing about PHP, but changes seem reasonable to me. cc'ing @tierra just in case.

Have you been able to confirm that with these fixes the search works (e.g. by deploying it locally)?

anhnong90:
Thanks for the review!

Yes, I tested the updated script locally using `PHP 8.3`. 
After generating the documentation and serving the `out/html` directory with a local PHP server, the search backend returns valid responses. The endpoint works without errors, and the search data loads correctly (e.g. `all_1d.js` and related files). Search results behave as expected.

tierra:
I knew this was all fixed in upstream Doxygen, and I upgraded copies in all older versions of docs using the upstream version of the file, and also attempted to upgrade the version in latest, but was baffled about how it just would break again on the next docs update, as you can see in a [previous discussion of this](https://github.com/wxWidgets/wxWidgets/issues/26482#issuecomment-4492737663).

What I didn't realize is that we had our own custom search functions file that was being copied in over the upstream version. Thanks for spotting this and pointing it out.

As far as I can tell, we only added our own override for that file to work around a bug in Doxygen < 1.8.19. If that's true, I believe this file can be removed entirely, which will then pull in the upstream maintained version that already has the PHP 8.x fixes. Want to try that out instead @anhnong90 ?

tierra:
To be clear, basically just means reverting cd4d9011b76b42fa9c212e665b029d4db3192f4e

anhnong90:
Thank you for the clarification! @tierra 
I suspected the same issue — something in the build pipeline was overwriting the fixed PHP search backend and likely causing the 500 error. I’ve removed custom_search_functions.php and the copy step, so CI now uses the upstream Doxygen version as intended.

tierra:
This looks perfect, thanks.

Do you usually squash and merge these @vadz ?

vadz:
I'll indeed squash and merge it, thanks again for fixing this!

PR #26613, opened 2026-06-18
MSW: Only use DarkMode_DarkTheme if available

stevecor:
Check windows version before using theme "DarkMode_DarkTheme" and use alternative theme if it is not available.


vadz:
Thanks for fixing this!

PR #26614, opened 2026-06-19
Avoid using very large wxDC origin offsets with wxGenericListCtrl

paulcor:
Using the wxDC origin to implement scrolled drawing can require very large origin offsets, which can exceed Cairo coordinate limits (2^23), and therefore currently wxGTK3 limits as well. Managing the scroll offsets turns out to be pretty simple in this case, avoiding any coordinate limitations.
See #25365

### DIFF
--- a/include/wx/generic/private/listctrl.h
+++ b/include/wx/generic/private/listctrl.h
@@ -275,13 +275,12 @@ public:
     }
 
     // draw the line on the given DC in icon/list mode
-    void Draw( wxDC *dc, bool current );
+    void Draw( wxDC *dc, const wxPoint& origin, bool current );
 
     // the same in report mode: it needs more parameters as we don't store
     // everything in the item in report mode
     void DrawInReportMode( wxDC *dc,
                            const wxRect& rect,
-                           const wxRect& rectHL,
                            bool highlighted,
                            bool current,
                            bool checked );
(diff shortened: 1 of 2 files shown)


vadz (commented):
Thanks! This is good and useful, as it solves a real problem, but I still wonder if we could do this inside `wxDC` itself, i.e. handle the offset ourselves instead of relying on Cairo for this?

paulcor:
> but I still wonder if we could do this inside `wxDC` itself

It looks pretty unappealing to me. I think we would have to do the whole matrix transform ourselves for every point and size in every function in `wxGCDC`.

vadz:
> > but I still wonder if we could do this inside `wxDC` itself
> 
> It looks pretty unappealing to me. I think we would have to do the whole matrix transform ourselves for every point and size in every function in `wxGCDC`.

Yes, true. It would, however, be simple to do it if the transformation is restricted to the origin shift (well, maybe axis scaling could be taken into account as well). This would still require modifying all the functions, but it's such an important use case (because this bug affects not just this control, but probably `wxGrid`, `wxDVC`, `wxVListBox` as well) that it might still be worth doing it.

paulcor:
Well, regardless of whether it would be terrible or merely awful, I don't think a `wxGCDC` solution is going to happen before 3.3.3.

vadz:
Sure, this should be merged to fix the problem. But I don't know if it's more or less awful to do the same thing in all the other affected classes (which still wouldn't fix it in the user code) or do it once in `wxDC`.

PR #26615, opened 2026-06-19
Allow the Named Return Value Optimization to work for some easy cases

paulcor:


src/common/filefn.cpp
```diff
@@ -200,11 +200,13 @@ wxString wxPathList::FindValidPath (const wxString& file) const
 
 wxString wxPathList::FindAbsoluteValidPath (const wxString& file) const
 {
+    wxString buf;
+
     wxString f = FindValidPath(file);
     if ( f.empty() || wxIsAbsolutePath(f) )
-        return f;
+        return buf;
```
vadz, line 207: This seems wrong, it returns empty string if `f` is an absolute path, which must break this function.

vadz (commented):
Sorry, I'm too tired to review all the changes right now, I'll try to do it later but seeing how this:

1. Makes the code less clear (you need to track the variable value in your head instead of seeing what the function returns from the code).
2. May actually introduce bugs.

I wonder if we really want to do this, i.e. if there are any measurable improvements compensating the above? I don't think performance improvements could be noticeable, maybe the code size is smaller? Or what is the rationale for these changes?

paulcor:
I'm just going to back off to the most obvious ones.

PR #26555, opened 2026-06-02
3.2 backports 202606

vadz:

PR #26617, opened 2026-06-21
Add wxGrid::GetFrozenRowLabelWindow() and GetFrozenColLabelWindow().

jmoraleda:
These accessors for the frozen row and column label sub-windows were missing from the public API even though the equivalent data cell windows (GetFrozenRowGridWindow/GetFrozenColGridWindow) are already public. Without them it is impossible to bind events (e.g. EVT_MOTION) to the frozen label strips from outside the wxGrid implementation.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -487,6 +487,7 @@ All (GUI):
 - Add wxApp::SetAppearance() to explicitly set dark or light mode (#24461).
 - Add wxAuiManager::{Save,Load}Layout() (#24235).
 - Add wxGrid::CopySelection() (Blake Madden, #24124).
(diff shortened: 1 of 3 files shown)


vadz:
Thanks, will merge soon (with a minor fix in the interface file).

jmoraleda:
Thank you @vadz. And thank you for fixing the col/row documentation swap in the interface file. Do you plan to add this to the 3.2 branch as well? It would be convenient for me. Thank you.

vadz:
Doing this in 3.2 requires adding some extra checks etc but I've done it in 80a009db93c5fefc45255a010b597adb9f47578e, part of #26555.

PR #26620, opened 2026-06-22
Don't write unnecessary auto render value to SVG output

Blake-Madden:
I'll get this fix in separately while I work on SVG unit tests.

### DIFF
--- a/src/common/dcsvg.cpp
+++ b/src/common/dcsvg.cpp
@@ -203,6 +203,8 @@ wxString GetPenPattern(const wxPen& pen)
         // these penstyles do not need a pattern.
         break;
     }
+    if (!s.empty())
+        s.Prepend(wxS(" "));
     return s;
 }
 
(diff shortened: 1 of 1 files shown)


vadz:
Note that this will still result in an extra space (due to `"%s "` being present in the format string). Not sure if it's worth fixing, but it would be tidier if we avoided it.

Blake-Madden:
> Note that this will still result in an extra space (due to `"%s "` being present in the format string). Not sure if it's worth fixing, but it would be tidier if we avoided it.

Fixed

vadz (approved):
Thanks, this is difficult to review and I'm not sure if any changes here would be detected by our tests, but assuming you tested it and there are no missing spaces anywhere, it looks good and I'll merge it.

I do wonder if we could use some better/safer way of building strings with SVG attributes than `wxString::Format("%s")`, e.g. something like `AttributeBuilder::Add("name", possibleEmptyValue).Add(...).ToString()`. But we would definitely want to have tests detecting breakage first if we ever decide to change this code like this.

PR #26624, opened 2026-06-23
reject truncated reads in ReadPCX to avoid uninitialised data

dxbjavid:
ReadPCX reads the 128-byte header, the uncompressed scanline buffer and the trailing 768-byte palette of an 8-bit image with stream.Read but never checks how many bytes actually came back, so a truncated pcx leaves the unread tail of those buffers uninitialised and that data then ends up in the decoded image and in the wxImage palette returned by GetPalette. A crafted file that stops just after the 0x0c palette marker is enough to copy uninitialised stack memory into the loaded image. This rejects the file when any of those reads comes up short, in the same way the loader already bails out on other malformed input, and adds a test feeding such a truncated palette.

### DIFF
--- a/src/common/imagpcx.cpp
+++ b/src/common/imagpcx.cpp
@@ -173,6 +173,8 @@ int ReadPCX(wxImage *image, wxInputStream& stream)
     // be at least 5 or higher for 8 bit and 24 bit images).
 
     stream.Read(hdr, 128);
+    if (stream.LastRead() != 128)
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, the fix looks good to me, so I'll merge it, but I'm a bit surprised that I don't get any errors neither from ASAN nor from Valgrind with the new test even without the fix. Shouldn't they complain about reading uninitialized memory? I must be missing something here because they don't...

PR #26626, opened 2026-06-23
Add DisableGPUOptimization() to wxWebView

Blake-Madden:


Blake-Madden:
I saw `CalculateRefreshOptimizations()` elsewhere in the wx API, so I went with the American "Optimizations" to be consistent.

PBfordev:
Firstly, should this not be a static method, since it is not instance-specific?

Secondly, setting the environment variable may collide with the user doing the same for their purposes with other arguments...

Perhaps it would be better to use `wxWebViewConfiguration` for this; and with Edge, `ICoreWebView2EnvironmentOptions::put_AdditionalBrowserArguments()`?

BTW, why using _Optimization_ instead of _Acceleration_ in the method name, these are not the same?

vadz:
Thanks, but I agree with all PBfordev's remarks:

1. This should be a static function.
2. It should be called `DisableGPUAcceleration()`.
3. Ideal would be to put this into `wxWebViewConfiguration` instead of adding a separate function for this.

Thanks!

Blake-Madden:
Now static, renamed, and in `wxWebViewConfiguration`. I didn't try `ICoreWebView2EnvironmentOptions::put_AdditionalBrowserArguments()` to keep it simple and so that I don't have to rely on a `wxWebViewConfiguration` existing before calling this. MSDN recommends setting this via the environment also.

src/common/webview.cpp
```diff
@@ -490,6 +490,18 @@ wxWebView::GetBackendVersionInfo(const wxString& backend,
         return wxVersionInfo();
 }
 
+// static
+bool wxWebViewConfiguration::DisableGPUAcceleration()
+{
+#if defined(__WXGTK__)
+    return wxSetEnv("WEBKIT_DISABLE_DMABUF_RENDERER", "1");
+#elif defined(__WXMSW__)
+    return wxSetEnv("WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS", "--disable-gpu");
```
vadz: I see a potential problem with doing it like this: what if this variable had already been set? I think we need to check for this and append `" --disable-gpu"` to it (note the space) to its value then.

Blake-Madden: Got it

vadz:
Thanks, will merge soon (with just one small tweak).

PR #26627, opened 2026-06-24
Fix dark mode ribbon under Linux (KDE)

Blake-Madden:
Under KDE, the ribbon takes a blue hue because its system gray has a blue tint in it that gets amplified here. This balances this to look more like how dark mode is under macOS and MSW.


Blake-Madden:
Before (KDE, OpenSuse and Linux Mint):

Screenshot from 2026-06-24 08-47-05

After:

Screenshot from 2026-06-24 08-58-06

Windows 11 and macOS look the same as before.

vadz:
Thanks, I don't see any difference here but I agree that the new screenshot looks better, so I'll merge this soon.

PR #26629, opened 2026-06-24
MSW: Dark mode switching for non-rich read-only multiline wxTextCtrl

stevecor:
For dark mode switching, update the background colour for non-rich read-only multiline `wxTextCtrl` controls. The foreground colour is already updated.

The background colour is controlled by assigning `m_backgroundColour` rather than calling `SetBackgroundColour()`. This leaves the `m_hasBgCol` member untouched so that it distinguishes custom colours set by the user which should not be overwritten, from colours set internally which can be overwritten.

Some checks on the value of `hasBgCol` are changed to `m_backgroundColour.IsOK()`. When `m_backgroundColour.IsOK()` is true, there is either a custom colour or an internally set colour.

### DIFF
--- a/src/msw/control.cpp
+++ b/src/msw/control.cpp
@@ -360,7 +360,7 @@ WXHBRUSH wxControl::DoMSWControlColor(WXHDC pDC, wxColour colBg, WXHWND hWnd)
     // finally also set the background colour for text drawing: without this,
     // the text in an edit control is drawn using the default background even
     // if we return a valid brush
-    if ( colBg.IsOk() || m_hasBgCol )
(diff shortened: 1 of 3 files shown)


stevecor:
This also fixes the dark mode background of `wxMDIClientWindow`. This class already sets it's background colour by assigning `m_backgroundColour` rather than calling `SetBackgroundColour()`.

vadz (approved):
Thanks, I've looked at this for a long time but I still can't think of anything that would be broken by this[*], so I'm going to merge it with maybe just a minor comment update.

[*] Which doesn't mean that nothing will — I would actually be surprised if nothing really breaks, but we have to apply this change to see what it is going to be.

Issue #26630, opened 2026-06-25
MSW: Dark mode border flickers white on WM_NCPAINT


memoarfaa:
### Description
When dark mode is active and a window has a themed border (wxBORDER_THEME,
wxBORDER_SUNKEN, etc.), the NC area briefly flashes the system light-mode border
before the dark custom border is drawn. This is visible as a white/light border flicker
on any resize, move-over, or partial NC invalidation.

https://github.com/user-attachments/assets/70a73e63-b6cd-4abf-8d82-2d91a4f9fdea

 wxWidgets port : wxMSW.

vadz:
This should have been fixed by 013f249f93159d74ab1e524f16bd04b3b14336ad which was part of #26646.

PR #26631, opened 2026-06-25
MSW: Minimize dark mode border flicker

stevecor:
These changes address flickering borders in dark mode, primarily with the non-rich `wxTextCtrl`. The `wxListBox` border also flickers, but is less noticeable. See #26630.

The flicker occurs when the mouse moves over the border or scroll bar. The problem might not be reproducible on all systems, perhaps depending on graphics drivers. The cause is that these controls generate many unnecessary `WM_NCPAINT` messages. In our `WM_NCPAINT` handler, the the light mode border gets drawn and the dark mode border over that.

The solution is to use the `DarkMode_DarkTheme` theme when available. With this theme, these controls draw a good dark border for most border styles, including the default style. In this case, we do not draw over the border. A new member is introduced to the `MSWDarkModeSupport` struct to indicate whether we need the custom border drawing.

When `DarkMode_DarkTheme` is not available, and for the border styles that flicker even with `DarkMode_DarkTheme`, convert the border style to `wxBORDER_SIMPLE`, which never flickers.

### DIFF
--- a/include/wx/msw/listbox.h
+++ b/include/wx/msw/listbox.h
@@ -135,6 +135,7 @@ public:
     bool MSWCommand(WXUINT param, WXWORD id) override;
     WXDWORD MSWGetStyle(long style, WXDWORD *exstyle) const override;
     void MSWGetDarkModeSupport(MSWDarkModeSupport& support) const override;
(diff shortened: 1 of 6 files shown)


stevecor:
PR #26635 appears to be a better solution.

PR #26633, opened 2026-06-25
Handle system colour change in wxInfoBarGeneric

stevecor:
Move colour settings from `Create()` into a new function `UpdateColours()`, which gets called from `Create()` and also upon system colour change.


vadz:
Thanks, LGTM.

